PR #16, opened 2026-04-06
support zig and powershell

avadevs:
Hello there, 

I stumbled across your project and I really like the idea. So I tried it out tonight. Sadly I need Zig and Powershell support, because thats what my projects currently use.

I have implemented the support for both languages in this PR. If you find the time to take a look and give me feedback, I would appreciate that a lot.


safishamsi:
Shipped in 9d998cc. I implemented both languages directly so there is no merge conflict with recent changes to extract.py. The approach matches exactly what you had. Thanks for the contribution.

PR #28, opened 2026-04-07
fix: add explicit encoding="utf-8" to all pathlib read/write calls

Will-hxw:
On Windows systems with GBK as the default encoding (common on Chinese Windows installations), pathlib's read_text() and write_text() fail with UnicodeDecodeError when the file contains UTF-8 characters. This affected all 16 read_text/write_text calls across install, uninstall, claude_install, claude_uninstall, and benchmark commands.

Fix: pass explicit encoding="utf-8" to every read_text() and write_text() call, and to json.loads(read_text()) calls that read config files.


safishamsi:
Already shipped in v0.3.4 with the same fix. The encoding='utf-8' is now on all 16 read_text/write_text calls. Closing since it's covered — appreciate you putting this together.

PR #33, opened 2026-04-07
Add follow_symlinks support to file discovery

mcharrel:
## Summary
- Adds opt-in `follow_symlinks` parameter to `detect()` and `collect_files()` so symlinked files and directories are no longer invisible to graphify
- Implements cycle detection that prevents infinite recursion from circular symlinks while still allowing legitimate sibling symlinks (checks if symlink target is an ancestor of current path)
- Updates `watch.py` to use `collect_files()` with the new parameter for consistency

## Test plan
- [x] Added 3 tests for `detect()`: symlinked directory, symlinked file, circular symlink protection
- [x] Added 2 tests for `collect_files()`: symlinked directory, circular symlink protection
- [x] All 359 existing tests pass

### DIFF
--- a/graphify/detect.py
+++ b/graphify/detect.py
@@ -293,7 +293,7 @@ def _is_ignored(path: Path, root: Path, patterns: list[str]) -> bool:
     return False
 
 
-def detect(root: Path) -> dict:
+def detect(root: Path, *, follow_symlinks: bool = False) -> dict:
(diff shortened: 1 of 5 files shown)


safishamsi:
Applied manually in v0.3.9. Kept the symlink support and cycle detection. The collect_files refactor from tuple of globs to set of extensions was also a nice cleanup. Thanks!

PR #44, opened 2026-04-07
fix: typo in SECURITY.md regarding MCP server

cosiato:
In Security.md, there is a typo on the word studio (written 'stdio')


safishamsi:
Thanks for the PR! Actually `stdio` is correct here - graphify's MCP server uses stdio transport (standard input/output), not a studio server. Closing to avoid introducing a typo.

PR #54, opened 2026-04-08
fix: PreToolUse hook output + add wiki CLI command

DarkSlaayer:
## Summary

- **Fix Claude Code PreToolUse hook**: The current hook uses plain `echo` which goes to the debug log and is never seen by the AI model. Changed to structured JSON output with `hookSpecificOutput.additionalContext`, which is the correct way to surface messages to Claude Code via hooks.
- **Add `graphify wiki` CLI command**: The `to_wiki()` function in `wiki.py` was implemented but never wired up to the CLI. This adds `graphify wiki [--graph path]` which generates a Wikipedia-style wiki from `graph.json`, reading community labels from `.graphify_labels.json` when available. Output goes to `graphify-out/wiki/` with `index.md` as the entry point.

## Test plan

- [ ] Run `graphify claude install` on a fresh project and verify `.claude/settings.json` hook uses JSON format
- [ ] Run a Glob or Grep tool in Claude Code and verify the hook message appears as `PreToolUse:Grep hook additional context`
- [ ] Run `graphify wiki` in a project with `graphify-out/graph.json` and verify wiki is generated
- [ ] Run `graphify wiki` with `.graphify_labels.json` present and verify community names are used
- [ ] Run `graphify wiki --graph custom/path/graph.json` and verify custom path works
- [ ] Run `graphify --help` and verify wiki command appears in help text

### DIFF
--- a/graphify/__main__.py
+++ b/graphify/__main__.py
@@ -30,8 +30,11 @@ _SETTINGS_HOOK = {
             "type": "command",
             "command": (
                 "[ -f graphify-out/graph.json ] && "
-                "echo 'graphify: Knowledge graph exists. Read graphify-out/GRAPH_REPORT.md "
-                "for god nodes and community structure before searching raw files.' || true"
+                "printf '{\"hookSpecificOutput\":{\"hookEventName\":\"PreToolUse\","
+                "\"permissionDecision\":\"allow\","
+                "\"additionalContext\":\"graphify: Knowledge graph exists. "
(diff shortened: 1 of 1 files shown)


safishamsi:
Thanks for catching both of these! The hook JSON format fix was merged in v0.3.13 (#83). The `graphify wiki` CLI command is a good idea — filing it as a separate feature request so it doesn't get lost.

PR #59, opened 2026-04-08
refactor(extract): split 2046-line god module into focused subpackage

darion-yaphet:
- Extract ExtractionContext class to eliminate 5x duplicated add_node/add_edge/call-graph-pass/clean-edges boilerplate
- Split by responsibility: _base (infrastructure), _imports (handlers), _configs (LanguageConfig instances), _generic (AST extractor), _go/_rust/_zig/_powershell (custom extractors), _python_extra (rationale + cross-file resolution), _dispatch (public API)
- Every file stays under 300 lines with single responsibility
- Full backward compatibility via __init__.py re-exports
- All 349 tests pass with no behavior changes


safishamsi:
The refactor is well-executed and the architectural concerns you're addressing are real — extract.py is getting large. We're deliberately deferring this split until the language set stabilizes (we're adding Dart, COBOL etc. soon and the split boundary will shift). When we do it, we'll use a similar structure to what you've proposed here. Closing for now but your design is a good reference.

PR #61, opened 2026-04-08
feat: add --output-dir flag to specify custom output directory

AveshLutchman:
## Summary

- Add `--output-dir <dir>` flag to the `/graphify` skill pipeline — all outputs (graph.json, GRAPH_REPORT.md, graph.html, cache/, etc.) go to the specified directory instead of the hardcoded `graphify-out`
- Add `output_dir` parameter to all cache functions in `cache.py` (backward-compatible, falls back to `root/graphify-out/cache` when not given)
- Add `output_dir` parameter and `--output-dir` CLI flag to `watch.py`; also fix a bug where the file-event guards hardcoded `"graphify-out"` and would have caused the watcher to loop on custom output dirs

## How it works

In `skill.md`, `${GRAPHIFY_OUT}` is used as a substitution placeholder (same pattern as the existing `INPUT_PATH`). Claude substitutes the user-provided value — or the default `graphify-out` — before executing each bash block.

```
/graphify <path> --output-dir my-outputs    # all outputs → my-outputs/
/graphify <path>                            # unchanged — outputs → graphify-out/
```

## Test Plan

- [x] `uv run pytest tests/test_cache.py` — 8/8 pass (includes 2 new tests for custom `output_dir`)
- [x] `uv run pytest tests/test_watch.py` — 9/9 pass (includes 1 new test for custom `--output-dir`)
- [x] `uv run pytest tests/` — 370/370 pass (no regressions)
- [x] Verify default behavior unchanged: omitting `--output-dir` still writes to `graphify-out/`


taitruong:
Hi, we're using Graphify with OpenClaw in a Docker container where the workspace and output directories are separate bind-mounts. The output always goes to ./graphify-out/ relative to CWD, which causes issues when the agent's working directory differs from where we want the output.

We currently work around this with a symlink (raw/graphify-out/ → /home/node/graphify-out/), but an --output-dir flag would be the clean solution.

Our use case: OpenClaw bot receives files via Telegram, saves to /home/node/raw/, runs /graphify /home/node/raw, but output needs to go to /home/node/graphify-out/ (separate bind-mount for the graph viewer).

Was there a specific reason this PR was closed/abandoned? Would a new PR for this feature be welcome?

Issue #57, opened 2026-04-08
Feature Request: Add Support for Google Gemini


ThangNguyenTan:
Graphify currently doesn't support Google Gemini as an LLM provider, which limits users who rely on or prefer Gemini over other available models.
Add support for Google Gemini (e.g. `gemini-3.1-pro`, `gemini-3.0-flash`) so users can plug it in as their model of choice when generating graphs.

Gemini has a well-documented API and an official SDK (`google-generativeai`), so integration should be fairly straightforward. 

Thanks for the great project! 🙌

safishamsi:
Gemini isn't directly supported yet as a model provider — graphify currently uses the Anthropic SDK for semantic extraction. Adding Gemini would require a second extraction backend.

That said, if you run graphify purely for AST/structural extraction (code files only, no docs/papers/images), no LLM is used at all — the output is deterministic and free.

Gemini support is tracked as part of the v0.4.0 milestone (#7) which will add a model-agnostic extraction interface.

PR #62, opened 2026-04-08
Gemini Cli support

chenxinhu:


safishamsi:
We just closed the Gemini feature request (#57) — Gemini support requires a second extraction backend and we're keeping the LLM interface tight for now. This will be revisited as part of the model-agnostic v0.4.0 milestone (#7). Thanks for the contribution.

PR #67, opened 2026-04-08
feat: expose build command for direct graph generation without LLM integration

ian000:
### What this PR does
Currently, the `graphify` CLI acts primarily as a skill wrapper for AI assistants. The underlying deterministic AST extraction (via tree-sitter) is incredibly powerful, but it wasn't easily accessible as a standalone command to simply build the `graphify-out` directory directly from a terminal without triggering the LLM/Claude flow.

This PR exposes the internal build process so that users can run:
```bash
graphify .
# or
graphify build .
```
This will detect the files, extract the AST, build the NetworkX graph, cluster communities, and generate the `graph.json` and `GRAPH_REPORT.md` artifacts locally, fully bypassing the LLM subagents. 

### Why is this useful?
1. **Zero-LLM Pipeline**: Allows external tools and CLIs to leverage Graphify's AST engine purely for structural code analysis.
2. **Offline Mode**: Enables generation of the knowledge graph in entirely air-gapped environments without any API keys.
3. **CI/CD Integration**: The graph can be generated as a build artifact in pipelines before any LLM review steps.

### Changes made
- Added a `build_graph` orchestrator function in `__main__.py` that composes the existing modules (`detect`, `extract`, `build_from_json`, `cluster`, `analyze`, `report`, `export`).
- Added CLI routing for `cmd == "."` or `cmd == "build"` to trigger this flow.

### DIFF
--- a/graphify/__main__.py
+++ b/graphify/__main__.py
@@ -299,6 +299,59 @@ def claude_uninstall(project_dir: Path | None = None) -> None:
     _uninstall_claude_hook(project_dir or Path("."))
 
 
+def build_graph(target_dir: Path, generate_html: bool = True, cluster_only: bool = False) -> None:
+    from graphify.detect import detect
+    from graphify.extract import collect_files, extract
+    from graphify.build import build_from_json
+    from graphify.cluster import cluster, score_all
+    from graphify.analyze import god_nodes, surprising_connections, suggest_questions
+    from graphify.report import generate
+    from graphify.export import to_json
+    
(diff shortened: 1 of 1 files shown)


safishamsi:
The `graphify build .` UX is the right direction — running the AST pipeline directly from the CLI without needing the skill wrapper is genuinely useful. The routing via `Path(cmd).is_dir()` is a bit fragile and the HTML generation path isn't wired through. We'll implement this properly as a dedicated `graphify build` subcommand. Thanks for the concept.

PR #70, opened 2026-04-08
[codex] Fix graphify compatibility and export hardening

missflash:
## What changed

This PR fixes a set of compatibility and hardening issues discovered while running the graphify pipeline inside Codex and reviewing the resulting local changes.

### Compatibility fixes
- added a small NetworkX node-link compatibility shim so both `edges=` and `link=` keyword styles work across 3.x variants
- updated clustering to pass only supported `louvain_communities()` kwargs for the installed NetworkX version
- updated JSON export and benchmark loading paths to handle the current node-link API without breaking older call sites

### Shared configuration cleanup
- introduced `graphify/constants.py` as the single source of truth for file-extension sets
- switched `detect.py`, `analyze.py`, and `watch.py` to consume the shared constants instead of maintaining duplicated extension tables

### Export and security hardening
- hardened the HTML graph info panel by escaping dynamic content before injecting it into `innerHTML`
- removed inline neighbor `onclick` handlers and replaced them with delegated click handling to avoid broken quoting and improve safety
- tightened Neo4j label / relation sanitization with allowlist-based filtering plus safe fallbacks
- added a post-connect peer IP check to reduce DNS rebinding risk during fetches
- kept DNS lookup failures as connectivity errors rather than turning them into URL-safety failures
- extended `sanitize_label()` with an optional `html_safe` mode for direct HTML contexts

### Codex integration
- added `AGENTS.md` guidance so Codex sessions can rebuild and consult the graph consistently inside this repository

## Why it changed

The graphify pipeline hit real runtime failures in this environment due to NetworkX API differences:
- `louvain_communities(..., max_level=...)` was not accepted by the installed version
- `node_link_data()` / `node_link_graph()` expected `link=` instead of `edges=`

While validating the generated outputs, the review also found two concrete issues:
- the HTML neighbor links could break because quoted node IDs were embedded directly inside inline handlers
- DNS resolution failures were being treated like security violations instead of normal network failures

This PR addresses those failures and folds in the related hardening work so the pipeline remains usable on current dependencies.

## Impact

- graph generation now runs successfully on the current local environment
- benchmark and export paths work with the installed NetworkX version
- the HTML graph viewer is safer and its neighbor navigation is no longer vulnerable to broken inline quoting
- extension handling is easier to maintain because file-type lists live in one place
- fetch security keeps the rebinding defense without breaking ordinary DNS/network error handling

## Validation

Ran:

```bash
/opt/anaconda3/envs/missflash/bin/python -m pytest -q tests/test_cluster.py tests/test_export.py tests/test_benchmark.py tests/test_security.py tests/test_watch.py tests/test_detect.py
```

Result:
- `79 passed in 0.46s`

Additional manual checks:
- generated `graph.html` and confirmed inline neighbor `onclick` handlers were removed
- confirmed delegated click handling was present in the info panel script
- verified `validate_url("https://example.com/")` still succeeds

## Root cause

The main root cause was dependency/API drift against the installed NetworkX version, combined with a couple of hardening changes that needed follow-up fixes after review:
- compatibility assumptions around `louvain_communities()` and node-link helpers were too narrow
- the first HTML escaping pass fixed XSS exposure but introduced broken neighbor-link quoting
- the first DNS validation change over-classified temporary name-resolution failures as security errors

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1,0 +1,8 @@
+## graphify
+
+This project has a graphify knowledge graph at graphify-out/.
+
+Rules:
+- Before answering architecture or codebase questions, read graphify-out/GRAPH_REPORT.md for god nodes and community structure
+- If graphify-out/wiki/index.md exists, navigate it instead of reading raw files
+- After modifying code files in this session, run `python3 -c "from graphify.watch import _rebuild_code; from pathlib import Path; _rebuild_code(Path('.'))"` to keep the graph current
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1,8 +1,0 @@
-## graphify
-
-This project has a graphify knowledge graph at graphify-out/.
-
-Rules:
-- Before answering architecture or codebase questions, read graphify-out/GRAPH_REPORT.md for god nodes and community structure
-- If graphify-out/wiki/index.md exists, navigate it instead of reading raw files
-- After modifying code files in this session, run `python3 -c "from graphify.watch import _rebuild_code; from pathlib import Path; _rebuild_code(Path('.'))"` to keep the graph current
--- a/graphify/__init__.py
+++ b/graphify/__init__.py
@@ -1,5 +1,39 @@
 """graphify - extract · build · cluster · analyze · report."""
 
+from __future__ import annotations
+
+import inspect
+
+from networkx.readwrite import json_graph
+
+
+def _patch_networkx_node_link_compat() -> None:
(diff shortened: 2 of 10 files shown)


safishamsi:
Really solid work — the NetworkX compat shim, `constants.py`, and `louvain_communities` kwargs inspection are all improvements over the current code. Unfortunately this conflicts with changes we landed today (cluster.py, detect.py). We're cherry-picking the constants.py and NetworkX shim parts into v3 directly. Thanks for the detailed PR.

PR #77, opened 2026-04-08
fix(update): prune graph nodes when source files are deleted (#51)

### LINKED ISSUE #51: --update should remove graph nodes when source files are deleted
# Problem
When running /graphify <path> --update, deleted files are detected (stored in deleted_files from detect_incremental()), but their corresponding nodes and edges are not removed from graph.json. This leaves "ghost nodes" in the graph — nodes that reference files that no longer exist.

# Current behavior
detect_incremental() correctly identifies deleted files:
deleted_files = [f for f in manifest if f not in current_files]
However, the --update workflow only processes new_files and unchanged_files. The deleted_files list is computed but never acted upon.

# Expected behavior
When a source file is deleted from the corpus, running --update should:
1. Remove all nodes whose source_file matches the deleted file
2. Remove all edges connected to those removed nodes
3. Update graph.json and GRAPH_REPORT.md accordingly
This keeps the knowledge graph in sync with the actual file system state.

# Proposed solution
Add a --prune flag (or always-on behavior during --update):
/graphify . --update        # --prune is default behavior
/graphify . --update --no-prune  # skip pruning (backward compatible)
Implementation would add a prune_deleted_files(deleted_files, G) step in the --update merge workflow, removing nodes where node.get('source_file') matches any deleted file path.

mehdiShariati:
`detect_incremental()` already computes `deleted_files` (manifest paths that no longer exist on disk). The `--update` merge path merged new extractions into the existing graph but never removed nodes tied to deleted sources, so `graph.json` could retain **stale nodes and edges**—reported in [#51](https://github.com/safishamsi/graphify/issues/51).

### What changed

**1. Library: `prune_nodes_for_deleted_files` (`graphify/build.py`)**

- Selects nodes whose `source_file` matches an entry in `deleted_files`, comparing paths via a normalized key (`normpath` + `normcase`) so Windows and POSIX spellings align with manifest strings.
- Skips nodes with no `source_file` (e.g. external/stdlib-style stubs).
- Removes selected nodes; incident edges drop with the nodes (undirected `Graph` semantics).
- Updates `G.graph["hyperedges"]`: drop hyperedges anchored to a deleted file; otherwise strip removed node ids; drop entries with fewer than two remaining members (avoids dangling references in exports/report).

**2. Skills: `--update` flow (`skill*.md`)**

- **Delete-only updates** no longer early-exit when `new_total == 0` if `deleted_files` is non-empty; writes an empty extraction artifact so the merge step still runs.
- **Merge:** load `hyperedges` from existing `graph.json` into the in-memory graph (parity with prior `to_json` output), **prune** using `deleted_files` from `.graphify_incremental.json`, then merge the new extraction and continue with the existing pipeline (cluster / report / export unchanged).

**3. Tests**

- `tests/test_build.py`: regression coverage for incremental detection, pruning, merge ordering, path normalization, and hyperedge behavior.
- `tests/test_hooks.py`: relax executable-bit assertions on **Windows** (POSIX execute bits are not reflected in `st_mode` the same way); still assert executable semantics on non-Windows. Unrelated to #51 but restores a green suite on Windows CI/dev machines.

### How to verify

```bash
pip install -e ".[mcp,pdf,watch]"
pytest tests/ -q
```

### Scope / follow-ups

- **`--update --no-prune`** is **out of scope** here; pruning is **always applied** in the updated skill merge. If maintainers want an opt-out, that should be a separate change (CLI flag and/or skill parameter).

---
Closes #51

### DIFF
--- a/graphify/build.py
+++ b/graphify/build.py
@@ -21,9 +21,68 @@
 #    before any graph construction happens.
 #
 from __future__ import annotations
+import os
 import sys
 import networkx as nx
 from .validate import validate_extraction
+
+
+def _canonical_file_key(path: str) -> str:
+    """Normalize paths for comparing manifest paths to node source_file (Windows-safe)."""
+    return os.path.normcase(os.path.normpath(path))
+
+
+def prune_nodes_for_deleted_files(G: nx.Graph, deleted_files: list[str]) -> int:
+    """Remove nodes tied to deleted source files, plus incident edges.
+
+    Matches ``node['source_file']`` to paths in *deleted_files* using normalized
+    path comparison. Nodes without ``source_file`` are kept (e.g. some inferred
+    aggregates). Hyperedges that belong to a deleted file are dropped; remaining
+    hyperedges have removed node ids stripped, and hyperedges with fewer than two
+    nodes left are dropped.
+
+    Returns the number of nodes removed.
+    """
+    if not deleted_files:
+        return 0
+
+    deleted_keys = {_canonical_file_key(p) for p in deleted_files}
+    to_remove: list = []
+    for nid, data in G.nodes(data=True):
+        sf = (data.get("source_file") or "").strip()
+        if not sf:
+            continue
+        if _canonical_file_key(sf) in deleted_keys:
+            to_remove.append(nid)
+
+    removed_ids = set(to_remove)
+    G.remove_nodes_from(to_remove)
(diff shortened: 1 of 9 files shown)


safishamsi:
Thanks for the thorough fix — this was merged by us in v0.3.14 (#51) about an hour before your PR arrived. The approach is very similar. Really appreciated the test coverage in your version.

PR #82, opened 2026-04-08
Fix unreliable node click detection in HTML viz

roach88:
## Summary

- **Fix**: vis-network's `click` event sometimes returns empty `params.nodes` even when clicking directly on a node. Replaced with a hover-tracked approach — `hoverNode`/`blurNode` track which node the cursor is over, a DOM click listener on the container selects the tracked node, and the original `network.on('click')` is kept as a fallback.
- **Test**: Added `tests/visual_test_click.py` — a visual browser test that generates a 20-node mixed graph (code, document, paper, image, rationale) with a debug overlay showing hover/click events in real time. All existing `test_export.py` tests pass.

## How it works

The hover detection path in vis-network is more reliable than click params. The fix layers:
1. `hoverNode`/`blurNode` events track which node the cursor is over
2. A plain DOM `click` listener on the container selects the tracked node
3. The original `network.on('click')` remains as a fallback for edge cases

## Test plan

- [x] All 13 `test_export.py` tests pass
- [x] Visual test generates HTML with debug overlay covering all node types
- [ ] Manual verification: hover → click → inspect panel populates reliably
- [ ] Manual verification: small (low-degree) nodes clickable
- [ ] Manual verification: clicking empty canvas resets info panel

### DIFF
--- a/graphify/export.py
+++ b/graphify/export.py
@@ -183,9 +183,28 @@ function focusNode(nodeId) {{
   showInfo(nodeId);
 }}
 
+// Track hovered node — hover detection is more reliable than click params
+let hoveredNodeId = null;
+network.on('hoverNode', params => {{
+  hoveredNodeId = params.node;
+  container.style.cursor = 'pointer';
+}});
+network.on('blurNode', () => {{
+  hoveredNodeId = null;
+  container.style.cursor = 'default';
+}});
+container.addEventListener('click', () => {{
+  if (hoveredNodeId !== null) {{
+    showInfo(hoveredNodeId);
+    network.selectNodes([hoveredNodeId]);
+  }}
+}});
 network.on('click', params => {{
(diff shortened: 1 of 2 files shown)


safishamsi:
The hover-track fix for vis-network click detection is correct and worth merging. One ask: `tests/visual_test_click.py` generates an HTML file and opens a browser — it's not a pytest-runnable unit test, so it doesn't belong in `tests/`. Could you either remove it or move it to `scripts/` or `tools/`? Once that's done we'll merge.

safishamsi:
Applied in v0.3.18. The hover tracking approach works much better than relying on vis.js click params. The hoveredNodeId variable now tracks the active node via hoverNode and blurNode events, and the DOM click handler uses that to trigger showInfo. Small and dense nodes that were previously unclickable now work reliably. Thanks for the fix.

Issue #52, opened 2026-04-08
Large-Scale App Project Analysis


sunrain520:
<img width="787" height="165" alt="Image" src="https://github.com/user-attachments/assets/a04877af-bb0d-4043-b52b-206c43782567" />

<img width="730" height="776" alt="Image" src="https://github.com/user-attachments/assets/1e71a9af-0ad8-4fe6-a4b1-18ead4c24401" />

Issue 1: tree-sitter Version Incompatibility Causes Full Pipeline Crash                                                                                                                                                
                                                                                                                                                                                                                         
  Severity: P0 — Blocker                                                                                                                                                                                                 
                                                                                                                                                                                                                         
  Symptom: extract() crashes inside _resolve_cross_file_imports():                                                                                                                                                       
  TypeError: Language.__init__() missing 1 required positional argument: 'name'                                                                                                                                          
                                                                                                                                                                                                                         
  Root Cause: graphify's pinned tree-sitter==0.20.1 (old API) is incompatible with newer language bindings (e.g. tree-sitter-python, tree-sitter-swift). The new Language() constructor requires two arguments           
  (language_capsule, name), while the old version only needed one.                                                                                                                                                       
                                                                                                                                                                                                                         
  Impact: The entire AST extraction pipeline fails — 9,389 code files produce zero output.                                                                                                                               
                                                                     
  Suggested Fix:                                                                                                                                                                                                         
  - Pin compatible tree-sitter version ranges in pyproject.toml      
  - Add a version check at the extract() entry point with a clear upgrade message instead of a cryptic TypeError                                                                                                         
                                                                                                                
  ---                                                                                                                                                                                                                    
  Issue 2: Python-Only Logic Crash Blocks Non-Python Projects                                                                                                                                                            
                                                                                                                                                                                                                         
  Severity: P0 — Design Flaw                                                                                                                                                                                             
                                                                                                                                                                                                                         
  Symptom: _resolve_cross_file_imports() is a Python-exclusive feature (uses Python parser for cross-file import resolution). When it crashes, the entire extract() function fails — including Swift, ObjC, C, and all   
  other languages.                                                                                                                                                                                                       
                                                                                                                                                                                                                         
  Root Cause: In extract(), lines 76–77:                                                                                                                                                                                 
  cross_file_edges = _resolve_cross_file_imports(py_results, py_paths)
  all_edges.extend(cross_file_edges)                                                                                                                                                                                     
[...]

PR #89, opened 2026-04-08
fix: pin tree-sitter>=0.23.0 with clear version error

sebastianbreguel:
## Summary
Users with `tree-sitter<0.23.0` hit a cryptic `TypeError: Language.__init__() missing 1 required positional argument: 'name'` because the codebase uses the Language API v2 (`Language(lang_fn())`) but `pyproject.toml` had no minimum version pin.

Pins `tree-sitter>=0.23.0` in `pyproject.toml` and adds a `_check_tree_sitter_version()` guard at the top of `extract()` that catches stale environments and gives an actionable error message instead of a deep traceback.

## Behavior
- **Before:** `TypeError` crash deep in the AST pipeline, no indication of what's wrong
- **After:** `RuntimeError: tree-sitter X.Y.Z is too old. graphify requires tree-sitter >= 0.23.0. Run: pip install --upgrade tree-sitter`

## Diff
- `pyproject.toml`: `"tree-sitter"` → `"tree-sitter>=0.23.0"` (+1/-1)
- `graphify/extract.py`: `_check_tree_sitter_version()` function + call in `extract()` (+19)

Total: 2 files, +20/-1

Addresses #52 (Issue 1)

### DIFF
--- a/graphify/extract.py
+++ b/graphify/extract.py
@@ -2334,6 +2334,24 @@ def extract_elixir(path: Path) -> dict:
 
 # ── Main extract and collect_files ────────────────────────────────────────────
 
+
+def _check_tree_sitter_version() -> None:
+    """Raise a clear error if tree-sitter is too old for the new Language API."""
+    try:
+        from tree_sitter import LANGUAGE_VERSION
+    except ImportError:
+        raise ImportError(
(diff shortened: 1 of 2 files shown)


safishamsi:
Merged into v3 — tree-sitter pin bumped to >=0.23.0 and _check_tree_sitter_version() added to extract(). Shipping in v0.3.20.

PR #93, opened 2026-04-08
Fix #92: preserve .graphify_python across invocations

### LINKED ISSUE #92: Skill falls back to system Python when .graphify_python is missing
## Bug

When graphify is installed via `pipx`, the skill's Step 1 writes the correct Python interpreter path to `graphify-out/.graphify_python`. All subsequent steps use `$(cat graphify-out/.graphify_python)` to run against the pipx venv where `networkx` and other dependencies live.

However, Step 9 cleans up `.graphify_python` along with the other temp files. This means any follow-up invocation (`--update`, `--cluster-only`, `query`, `path`, `explain`) that doesn't re-run Step 1 will fall back to `$(which python3)` — the system Python — which doesn't have networkx installed:

```
ModuleNotFoundError: No module named 'networkx'
```

## Repro

1. Install via `pipx install graphifyy`
2. Run `/graphify .` (works fine — Step 1 resolves the pipx interpreter)
3. Run `/graphify query "something"` (fails — `.graphify_python` was cleaned up, falls back to system Python)

## Suggested fix

Every skill entry point (not just the full pipeline) should resolve the correct interpreter at the top. Either:

1. Don't clean up `.graphify_python` in Step 9 — it's tiny and useful across invocations
2. Or add the interpreter detection snippet to the start of every subcommand (query, path, explain, --update, --cluster-only)

Option 1 is simpler and covers all cases.

## Environment

- macOS (Homebrew Python 3.14, externally managed)
- graphify installed via `pipx install graphifyy` (v0.3.15)
- Claude Code CLI

azizur100389:
## Summary
- Stop deleting `.graphify_python` during Step 9 cleanup so follow-up commands (`query`, `path`, `explain`, `--update`, `--cluster-only`) continue to use the correct pipx interpreter instead of falling back to system Python
- Fix `mkdir` ordering in `skill.md` — create `graphify-out/` before writing `.graphify_python` into it
- Add an interpreter guard before subcommand sections that re-resolves the Python path if `.graphify_python` is missing (e.g. user manually deleted `graphify-out/`)
- Applied across all 7 skill files (skill.md, skill-codex.md, skill-claw.md, skill-droid.md, skill-opencode.md, skill-trae.md, skill-windows.md)

Closes #92

## Test plan
- [x] Bash syntax validation on interpreter guard snippet
- [x] Verified `.graphify_python` is no longer in any cleanup line across all skill files
- [x] No regressions in existing tests

### DIFF
--- a/graphify/skill-claw.md
+++ b/graphify/skill-claw.md
@@ -594,7 +594,7 @@ cost_path.write_text(json.dumps(cost, indent=2))
 print(f'This run: {input_tok:,} input tokens, {output_tok:,} output tokens')
 print(f'All time: {cost[\"total_input_tokens\"]:,} input, {cost[\"total_output_tokens\"]:,} output ({len(cost[\"runs\"])} runs)')
 "
-rm -f .graphify_detect.json .graphify_extract.json .graphify_ast.json .graphify_semantic.json .graphify_analysis.json .graphify_labels.json .graphify_python
+rm -f .graphify_detect.json .graphify_extract.json .graphify_ast.json .graphify_semantic.json .graphify_analysis.json .graphify_labels.json
 rm -f graphify-out/.needs_update 2>/dev/null || true
 ```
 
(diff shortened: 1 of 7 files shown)


safishamsi:
Applied in v0.3.18. Three changes went in: mkdir -p graphify-out now runs before writing .graphify_python so the first run on a fresh directory no longer fails, .graphify_python is no longer deleted in the Step 9 cleanup across all skill files, and an interpreter guard section was added to skill.md for cases where graphify-out gets deleted before a follow-up command. Thanks for tracking this down.

PR #102, opened 2026-04-09
Fix packaged skill coverage for install targets

missflash:
## Behavior before

  - CLI accepted newer install targets
  - package metadata did not fully include all required skill files
  - packaged installs for some supported platforms were at risk of failing with missing-file errors

  ## Behavior after

  - all supported install targets have matching packaged skill assets
  - tests now cover the full supported install matrix exposed by the CLI
  - future regressions are more likely to be caught during test runs

  ## Validation

  Verified with:

  - `python3 -m compileall graphify tests`
  - direct `install(platform=...)` execution for:
    - `claude`
    - `codex`
    - `opencode`
    - `claw`
    - `droid`
    - `trae`
    - `trae-cn`
    - `windows`

  `pytest` was not available in the current environment, so full test-suite execution was not run here.

  ## Scope

  This PR intentionally avoids broader refactors and only fixes the packaging/test gap needed to make supported
  install targets consistent with shipped artifacts.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -57,4 +57,4 @@ where = ["."]
 include = ["graphify*"]
 
 [tool.setuptools.package-data]
-graphify = ["skill.md", "skill-codex.md", "skill-opencode.md", "skill-claw.md", "skill-windows.md", "skill-droid.md"]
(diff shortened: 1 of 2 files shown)


safishamsi:
Applied in v0.3.18. skill-trae.md is now included in pyproject.toml package-data alongside the other skill files. The test suite also has four new install tests covering droid, trae, trae-cn, and windows, and test_all_skill_files_exist_in_package now checks all seven skill files. Thanks for catching this.

PR #109, opened 2026-04-09
fix: extension set drift in analyze.py and watch.py; remove dead build_graph

xiaolai:
## Problem

Three modules maintain independent copies of the supported-file-extension sets. They have drifted apart as new languages were added.

### 1. `analyze.py:_file_category()` misclassifies 15 of 30 supported languages

`analyze.py:112` defined its own `_CODE_EXTENSIONS` with only 15 entries:

```python
# analyze.py (before)
_CODE_EXTENSIONS = {"py", "ts", "tsx", "js", "go", "rs", "java", "rb", "cpp", "c", "h", "cs", "kt", "scala", "php"}
```

The canonical set in `detect.py:20` has 30 entries. Missing from `analyze.py`:

> `.jsx`, `.swift`, `.lua`, `.zig`, `.ps1`, `.ex`, `.exs`, `.m`, `.mm`, `.jl`, `.cc`, `.cxx`, `.hpp`, `.kts`, `.toc`

**Effect:** `_file_category()` classifies Swift, Lua, Zig, Julia, PowerShell, Elixir, Objective-C, and JSX files as `"doc"` instead of `"code"`. This inflates surprise scores for cross-file edges involving those languages — they get a +2 "crosses file types (doc ↔ code)" bonus from `_surprise_score()` that they shouldn't.

There was also a format mismatch: `detect.py` uses dotted extensions (`.py`), while `analyze.py` used bare extensions (`py`). The lookup worked before only because both sides were consistent within themselves. This PR normalizes `_file_category()` to prepend `.` before lookup.

### 2. `watch.py` ignores newer languages in `--watch` mode

`watch.py:8-18` maintained two hardcoded sets (`_WATCHED_EXTENSIONS`, `_CODE_EXTENSIONS`) that were also behind `detect.py`. File saves in Lua, Zig, PowerShell, Elixir, Objective-C, Julia, JSX, and TSX would not trigger auto-rebuild during `--watch`.

### 3. `cluster.py:build_graph()` is dead code

`cluster.py:55-70` defines a `build_graph(nodes, edges)` function that:
- Is never imported by any module or test (`grep` confirms)
- Is not exported via `__init__.py` (which exports `build_from_json` from `build.py`)
- Duplicates the logic in `build.build_from_json()`, including the `_src`/`_tgt` direction-preservation pattern

## Fix

- **analyze.py**: Delete the 4 local extension sets. Import `CODE_EXTENSIONS`, `DOC_EXTENSIONS`, `PAPER_EXTENSIONS`, `IMAGE_EXTENSIONS` from `detect.py`. Fix `_file_category()` to use dotted extension format.
- **watch.py**: Delete the 2 local extension sets. Derive `_WATCHED_EXTENSIONS` and `_CODE_EXTENSIONS` from `detect.py` imports.
- **cluster.py**: Delete the unused `build_graph()` function.
- **test_analyze.py**: Add 8 assertions to `test_file_category()` covering the languages that were previously misclassified (swift, lua, zig, ps1, ex, jsx, jl, m).

## Verification

- 375/375 tests pass (the 7 `test_security.py` failures in my environment are pre-existing — caused by a local proxy resolving public hostnames to `198.18.x.x` private IPs, unrelated to this change)
- `git grep` confirms no remaining imports of `cluster.build_graph` anywhere in the codebase
- Net change: **+17 lines, −39 lines** across 4 files

## Test plan

- [x] `pytest tests/test_analyze.py` — all pass, including new `test_file_category` assertions
- [x] `pytest tests/test_cluster.py` — all pass after `build_graph` removal
- [x] `pytest tests/test_watch.py` — all pass with derived extension sets
- [x] Full suite `pytest tests/` — 375 passed

### DIFF
--- a/graphify/analyze.py
+++ b/graphify/analyze.py
@@ -1,6 +1,7 @@
 """Graph analysis: god nodes (most connected), surprising connections (cross-community), suggested questions."""
 from __future__ import annotations
 import networkx as nx
+from .detect import CODE_EXTENSIONS, DOC_EXTENSIONS, PAPER_EXTENSIONS, IMAGE_EXTENSIONS
 
 
 def _node_community_map(communities: dict[int, list[str]]) -> dict[str, int]:
@@ -109,19 +110,13 @@ def _is_concept_node(G: nx.Graph, node_id: str) -> bool:
     return False
 
 
-_CODE_EXTENSIONS = {"py", "ts", "tsx", "js", "go", "rs", "java", "rb", "cpp", "c", "h", "cs", "kt", "scala", "php"}
-_DOC_EXTENSIONS = {"md", "txt", "rst"}
-_PAPER_EXTENSIONS = {"pdf"}
-_IMAGE_EXTENSIONS = {"png", "jpg", "jpeg", "webp", "gif", "svg"}
-
-
 def _file_category(path: str) -> str:
-    ext = path.rsplit(".", 1)[-1].lower() if "." in path else ""
-    if ext in _CODE_EXTENSIONS:
+    ext = ("." + path.rsplit(".", 1)[-1].lower()) if "." in path else ""
+    if ext in CODE_EXTENSIONS:
         return "code"
-    if ext in _PAPER_EXTENSIONS:
+    if ext in PAPER_EXTENSIONS:
         return "paper"
-    if ext in _IMAGE_EXTENSIONS:
+    if ext in IMAGE_EXTENSIONS:
         return "image"
     return "doc"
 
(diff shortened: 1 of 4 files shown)


safishamsi:
Applied in v0.3.18. analyze.py and watch.py now import CODE_EXTENSIONS, DOC_EXTENSIONS, PAPER_EXTENSIONS, and IMAGE_EXTENSIONS directly from detect.py instead of keeping their own copies. The dead build_graph() function in cluster.py has been removed. Eight new test cases were added to test_file_category covering the languages that were previously misclassified. Thanks for the clean fix.

PR #125, opened 2026-04-09
feat: add --directed flag to preserve edge direction (source→target)

freezingflamecode:
Undirected graphs store edges in arbitrary order, causing 60% of arrows in HTML visualization to point the wrong way. This adds an opt-in --directed flag that builds a DiGraph, preserving call/import direction.

Default remains undirected for backward compatibility. Louvain/Leiden community detection converts to undirected internally as required.

Includes worked example with A/B test evidence (worked/directed-graph/).


freezingflamecode:
@safishamsi rebased with the latest as suggested, pls review

safishamsi:
Fixed in 0.3.25. You can now pass --directed to get a directed graph that preserves edge direction. The cluster step handles it automatically by converting to undirected internally for community detection. Upgrade with pip install --upgrade graphifyy

PR #127, opened 2026-04-09
fix: AST call edges use correct confidence and directed graph

Minidoracat:
## Summary

- **Directed graph by default**: Changed `nx.Graph()` to `nx.DiGraph()` in `build.py` and `cluster.py`, preserving call edge direction (caller→callee). Louvain/Leiden community detection converts to undirected internally as required.
- **Correct confidence for AST call edges**: Tree-sitter-resolved function calls were marked `INFERRED/0.8` — changed to `EXTRACTED/1.0` across all language extractors (generic, Go, Rust, Zig, Obj-C, Java, Elixir). Cross-file class-level `uses` edges remain `INFERRED` as intended.
- **Adapted downstream consumers**: Updated `analyze.graph_diff` edge key logic, `cluster.cohesion_score` denominator, and type annotations throughout.

### Before
- `extractItems() --calls--> getTable()` produced **bidirectional** edges (phantom reverse edge)
- Utility functions like `cached()` and `getTable()` got inflated degree (~2x), ranking as god nodes
- All AST call edges marked `INFERRED` with weight `0.8`

### After
- Call edges are unidirectional (caller→callee only)
- God node rankings reflect actual connectivity
- AST call edges correctly marked `EXTRACTED` with weight `1.0`

## Relation to #125

This PR overlaps with #125 (`--directed` flag) in the graph construction layer. The key difference:

- **#125**: Adds an opt-in `--directed` flag, default remains undirected
- **This PR**: Makes DiGraph the default (breaking change), and also fixes the INFERRED confidence bug which #125 does not address

If both are merged, the `--directed` flag from #125 could be repurposed as `--undirected` for backward compatibility. Merging this PR first would require #125 to rebase and remove overlapping changes.

## Test plan

- [x] 52 relevant tests pass (test_build, test_cluster, test_analyze, test_confidence, test_export)
- [x] Full test suite: same pass/fail count as before (257 passed, 105 pre-existing failures unchanged)
- [x] No new test failures introduced

### DIFF
--- a/graphify/analyze.py
+++ b/graphify/analyze.py
@@ -444,7 +444,7 @@ def suggest_questions(
     return questions[:top_n]
 
 
-def graph_diff(G_old: nx.Graph, G_new: nx.Graph) -> dict:
+def graph_diff(G_old: nx.Graph | nx.DiGraph, G_new: nx.Graph | nx.DiGraph) -> dict:
     """Compare two graph snapshots and return what changed.
 
     Returns:
@@ -471,7 +471,9 @@ def graph_diff(G_old: nx.Graph, G_new: nx.Graph) -> dict:
         for n in removed_node_ids
     ]
 
-    def edge_key(G: nx.Graph, u: str, v: str, data: dict) -> tuple:
+    def edge_key(G: nx.Graph | nx.DiGraph, u: str, v: str, data: dict) -> tuple:
+        if G.is_directed():
+            return (u, v, data.get("relation", ""))
         return (min(u, v), max(u, v), data.get("relation", ""))
 
     old_edge_keys = {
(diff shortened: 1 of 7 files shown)


safishamsi:
Cherry-picked the confidence fix in the latest commit. AST call edges are now EXTRACTED/1.0 across all 7 language extractors — tree-sitter resolves these directly from source so INFERRED was wrong. Cross-file class-level uses edges remain INFERRED as intended. The DiGraph change was left out for now to avoid breaking graph.json compatibility, but is tracked separately via PR #125. Thanks for flagging both issues.

safishamsi:
Merged into v3 — all AST call edges now use confidence=EXTRACTED, weight=1.0. Shipping in v0.3.20.

PR #153, opened 2026-04-09
[codex] Fix codex install idempotency

DarkSlaayer:
## Summary
- make `graphify codex install` continue past an existing `AGENTS.md` marker so reruns still install the Codex PreToolUse hook
- preserve idempotency by keeping the `AGENTS.md` section single-instance while letting Codex and OpenCode heal partial installs
- extend install tests to cover fresh Codex installs, repeated installs, and the partial-install recovery path

## Root cause
`_agents_install()` returned as soon as it saw the graphify marker in `AGENTS.md`. For Codex, that meant a partial install could never resume into `_install_codex_hook()`, leaving `.codex/hooks.json` missing forever after the first failure.

## Impact
`graphify codex install` is now resumable after a partial success. A rerun will create a missing `.codex/hooks.json` without duplicating the `AGENTS.md` section or the hook entry.

## Validation
- `uv run --with pytest pytest tests/test_install.py -q`
- CLI verification in fresh temp directories for:
  - fresh install creates `AGENTS.md` and `.codex/hooks.json`
  - partial install rerun creates the missing hook
  - repeated install keeps one graphify section and one hook entry

### DIFF
--- a/graphify/__main__.py
+++ b/graphify/__main__.py
@@ -343,18 +343,24 @@ def _uninstall_codex_hook(project_dir: Path) -> None:
 def _agents_install(project_dir: Path, platform: str) -> None:
     """Write the graphify section to the local AGENTS.md (Codex/OpenCode/OpenClaw)."""
     target = (project_dir or Path(".")) / "AGENTS.md"
+    agents_already_configured = False
 
     if target.exists():
         content = target.read_text(encoding="utf-8")
         if _AGENTS_MD_MARKER in content:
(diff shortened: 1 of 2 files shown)


safishamsi:
Thanks for catching this! The codex install idempotency fix shipped in 0.3.24. Re-runs now correctly install the hook even when AGENTS.md already has the graphify section.

safishamsi:
Implemented from our end in 0.3.24. Thanks for the contribution!

PR #163, opened 2026-04-09
fix: catch exceptions in MCP call_tool handler

Koushik-Salammagari:
## Summary

- Wrap the `call_tool()` handler dispatch in a `try/except` so any exception raised by a tool handler is caught and returned as a `TextContent` error message instead of propagating out and crashing the MCP stdio loop.
- Add 4 unit tests covering: unknown tool, handler exception, valid handler, and no-propagation guarantee.

## Problem

Previously, if a registered tool handler raised an exception, `call_tool()` would propagate it uncaught. In an MCP stdio server, this crashes the server process rather than returning a graceful error to the client.

## Test plan

- [x] `pytest tests/test_serve.py -q` — all tests pass (including 4 new error-handling tests)
- [x] No regressions in full suite

### DIFF
--- a/graphify/serve.py
+++ b/graphify/serve.py
@@ -309,7 +309,10 @@ def serve(graph_path: str = "graphify-out/graph.json") -> None:
         handler = _handlers.get(name)
         if not handler:
             return [types.TextContent(type="text", text=f"Unknown tool: {name}")]
-        return [types.TextContent(type="text", text=handler(arguments))]
+        try:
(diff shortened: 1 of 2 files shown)


safishamsi:
Fixed in 0.3.25. The MCP call_tool handler now catches exceptions and returns an error message to the client instead of crashing the server process. Upgrade with pip install --upgrade graphifyy

PR #168, opened 2026-04-09
Fix .graphifyignore discovery to search parent directories

codeOfDUlbrich:
## Summary
- `.graphifyignore` is now discovered by walking upward from the scan root through parent directories, matching `.gitignore` behavior
- Stops at the git repo boundary (`.git` directory) so patterns don't leak across repositories
- Patterns from all ancestor `.graphifyignore` files are merged with the local one

## Test plan
- [x] Existing `.graphifyignore` tests still pass
- [x] New test: parent's ignore file applies to subdirectory scans
- [x] New test: patterns from local and ancestor files are merged
- [x] New test: upward search stops at `.git` boundary
- [x] New test: ignore file at repo root is included when scanning a subdir

### DIFF
--- a/graphify/detect.py
+++ b/graphify/detect.py
@@ -250,21 +250,34 @@ def _is_noise_dir(part: str) -> bool:
 
 
 def _load_graphifyignore(root: Path) -> list[str]:
-    """Read .graphifyignore from root and return a list of patterns.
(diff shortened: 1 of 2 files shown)


safishamsi:
Fixed in 0.3.25. graphifyignore now searches parent directories up to the git repo root, the same way gitignore works. Upgrade with pip install --upgrade graphifyy

PR #190, opened 2026-04-10
feat(extract): add node_type and modifiers to AST-extracted nodes

ksaketh19:
## Summary

Adds `node_type` and `modifiers` fields to every node produced by the AST extractors. These fields were previously absent, making it impossible to filter nodes by kind (e.g. _give me all exported interfaces_) without reading source files.

**New node fields (both optional, written only when non-null):**
- `node_type`: `"class"` | `"interface"` | `"function"` | `"method"` | `"type"` | `"enum"` | `"variable"`
- `modifiers`: `["export", "public", "private", "protected", "static", "abstract", "async", "readonly", ...]`

### `_extract_generic` (JS/TS/Python/Java/C/C++/Ruby/C#/Kotlin/Scala/PHP/Lua/Swift and all other tree-sitter languages)
- Class walk: detects `node_type` from tree-sitter node type (`interface_declaration` → `"interface"`, `class_declaration` → `"class"`, etc.); collects `export`, `abstract`, and accessibility modifiers
- Function/method walk: sets `node_type="method"` when inside a class, `"function"` otherwise; collects `export`, `async`, `static`, accessibility, and `override` modifiers
- Arrow functions in `_js_extra_walk`: marks `export` when parent is `export_statement`
- Added `interface_declaration` to `_TS_CONFIG.class_types` so TypeScript interfaces are extracted as first-class nodes

### `extract_go` (custom walk)
- `function_declaration` → `node_type="function"`; uppercase first letter → `modifiers=["export"]`
- `method_declaration` → `node_type="method"`; receiver type node → `node_type="class"`; uppercase names → `modifiers=["export"]`
- `type_declaration`/`type_spec` → `node_type="interface"` for `interface_type` bodies, `node_type="class"` for `struct_type` bodies, `node_type="type"` otherwise; uppercase name → `modifiers=["export"]`

## Backward compatibility

Fields are only written when non-null — existing consumers of `graph.json` that don't use these fields are unaffected.

## Example output

```json
{ "id": "skyflow_IInsertRecord", "label": "IInsertRecord", "node_type": "interface", "modifiers": ["export"], "source_file": "src/utils/common/index.ts", "source_location": "L42" }
{ "id": "skyflow_SkyflowClient", "label": "SkyflowClient", "node_type": "class", "modifiers": ["export"], "source_file": "src/index.ts", "source_location": "L12" }
{ "id": "skyflow_insert", "label": "insert()", "node_type": "method", "modifiers": ["public", "async"], "source_file": "src/index.ts", "source_location": "L45" }
```

### DIFF
--- a/graphify/extract.py
+++ b/graphify/extract.py
@@ -37,16 +37,21 @@ def extract_python(path: Path) -> dict:
     edges: list[dict] = []
     seen_ids: set[str] = set()
 
-    def add_node(nid: str, label: str, line: int) -> None:
+    def add_node(nid: str, label: str, line: int, node_type: str | None = None, modifiers: list | None = None) -> None:
         if nid not in seen_ids:
             seen_ids.add(nid)
-            nodes.append({
+            entry: dict = {
                 "id": nid,
                 "label": label,
                 "file_type": "code",
                 "source_file": str_path,
                 "source_location": f"L{line}",
-            })
+            }
+            if node_type is not None:
+                entry["node_type"] = node_type
+            if modifiers:
+                entry["modifiers"] = modifiers
+            nodes.append(entry)
 
     def add_edge(src: str, tgt: str, relation: str, line: int) -> None:
         # Only add edge if both endpoints exist or src is the file node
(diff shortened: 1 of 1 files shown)


graphify/extract.py
```diff
@@ -1966,16 +2083,21 @@ def extract_php(path: Path) -> dict:
...
                 "file_type": "code",
                 "source_file": str_path,
                 "source_location": f"L{line}",
-            })
+            }
+            if node_type is not None:
+                entry["node_type"] = node_type
+            if modifiers:
+                entry["modifiers"] = modifiers
+            nodes.append(entry)
```
Pranava29, line 2100: this piece of code looks repeated several times across the extract.py. any way to make it common?

PR #204, opened 2026-04-11
fix: use utf-8 encoding and LF newlines for all text file IO

NYCU-Chung:
## Summary

Python's `open()` / `Path.write_text()` / `Path.read_text()` default to the
system locale encoding on Windows (CP1252 / CP932 / CP950), causing
`UnicodeEncodeError` when graph labels contain CJK characters or emojis.
This breaks file writes *after* LLM extraction is complete, discarding the
paid Claude API work from pass 3.

Also forces LF line endings for git hook installation, extending the fix
direction of commit `210243f` ("fix hook reinstall, CRLF labels") to the
remaining hook write sites.

## Concrete failure modes

1. **GRAPH_REPORT.md / wiki `.md` / Cypher `.txt` writes crash on CJK Windows**
   Unlike `graph.json` (which benefits from `json.dumps`'s default
   `ensure_ascii=True` escaping), these files write raw CJK bytes. Under
   CP932 (JP) / CP950 (zh-TW) / CP949 (KR) locales, the final write step
   raises `UnicodeEncodeError` and destroys the extraction run.

2. **Git hook CRLF breaks `sh` interpreter**
   `Path.write_text()` without `newline="\n"` converts `\n` → `\r\n` on
   Windows, producing a `#!/bin/sh\r` shebang. Git Bash / WSL then reports
   `bad interpreter: No such file or directory` when the hook fires. This
   is the same class of bug as `210243f`.

3. **`read_text()` on UTF-8 files written by other platforms**
   Reading `graph.json` / cache files / manifests created on Linux or
   macOS can fail on a Windows system with a non-UTF-8 default locale,
   even if the file itself is valid UTF-8.

## Changes

All text file IO in `graphify/` now explicitly passes `encoding="utf-8"`:

| File | Locations |
|---|---|
| `graphify/export.py` | `to_json`, `to_cypher`, `to_obsidian` |
| `graphify/cache.py` | `save_cached`, `load_cached` |
| `graphify/detect.py` | paper detection, word count, `.graphifyignore` load, manifest read/write |
| `graphify/watch.py` | `GRAPH_REPORT.md`, flag file |
| `graphify/wiki.py` | 3 wiki article write sites |
| `graphify/hooks.py` | install / uninstall / status — **also adds `newline="\n"` on hook writes to force LF** |
| `graphify/serve.py` | `graph.json` read on startup |
| `graphify/benchmark.py` | `graph.json` read |

No logic changes, no function signature changes, no behavior changes.
Binary-mode IO (`"rb"` / `"wb"`) is untouched. `newline="\n"` is only
applied to git hook script writes — markdown / JSON outputs keep default
newline handling to match existing file conventions.

## Tests

New file: `tests/test_encoding_roundtrip.py` (5 tests):

- `test_write_read_text_cjk_roundtrip_primitive` — baseline IO pattern
  with CJK + emoji + `ensure_ascii=False` JSON payloads
- `test_cache_save_load_uses_utf8` — exercises
  `graphify.cache.save_cached` / `load_cached` with a CJK label
- `test_to_json_open_pattern_cjk` — reproduces the fixed `to_json`
  `open()` pattern with `ensure_ascii=False` to force UTF-8 bytes into
  the output
- `test_hook_install_lf_only` — asserts
  `graphify.hooks._install_hook` produces no `\r\n` on fresh install
- `test_hook_append_lf_only` — same assertion when appending to an
  existing hook

The hook CRLF tests were manually verified to **fail on the un-patched
`v4` HEAD** on Windows (raw bytes contain `\r\n`) and **pass after the
fix** — so they function as real regression guards rather than
tautologies.

All 5 tests pass on Windows Python 3.10.

## Out of scope (intentionally)

- `os.chmod(0o755)` in hook install is still a no-op on Windows — that's
  a separate cross-platform concern and not addressed here.
- `json.dumps(..., ensure_ascii=False)` for human-readable `graph.json`
  output is a product decision, not touched.
- Performance / parallelism of the extract loop is out of scope.

## Related

- `210243f` — "fix hook reinstall, CRLF labels" (prior fix in the same
  bug class; this PR sweeps the remaining sites)

### DIFF
--- a/graphify/benchmark.py
+++ b/graphify/benchmark.py
@@ -75,7 +75,7 @@ def run_benchmark(
 
     Returns dict with: corpus_tokens, avg_query_tokens, reduction_ratio, per_question
     """
-    data = json.loads(Path(graph_path).read_text())
+    data = json.loads(Path(graph_path).read_text(encoding="utf-8"))
     try:
         G = json_graph.node_link_graph(data, edges="links")
     except TypeError:
--- a/graphify/cache.py
+++ b/graphify/cache.py
@@ -55,7 +55,7 @@ def load_cached(path: Path, root: Path = Path(".")) -> dict | None:
     if not entry.exists():
         return None
     try:
-        return json.loads(entry.read_text())
+        return json.loads(entry.read_text(encoding="utf-8"))
     except (json.JSONDecodeError, OSError):
         return None
 
@@ -70,7 +70,7 @@ def save_cached(path: Path, result: dict, root: Path = Path(".")) -> None:
     entry = cache_dir(root) / f"{h}.json"
     tmp = entry.with_suffix(".tmp")
     try:
-        tmp.write_text(json.dumps(result))
+        tmp.write_text(json.dumps(result), encoding="utf-8")
         os.replace(tmp, entry)
     except Exception:
         tmp.unlink(missing_ok=True)
--- a/graphify/detect.py
+++ b/graphify/detect.py
@@ -69,7 +69,7 @@ def _looks_like_paper(path: Path) -> bool:
     """Heuristic: does this text file read like an academic paper?"""
     try:
         # Only scan first 3000 chars for speed
-        text = path.read_text(errors="ignore")[:3000]
+        text = path.read_text(encoding="utf-8", errors="ignore")[:3000]
         hits = sum(1 for pattern in _PAPER_SIGNALS if pattern.search(text))
         return hits >= _PAPER_SIGNAL_THRESHOLD
     except Exception:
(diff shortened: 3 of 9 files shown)


pedrolbacelar (approved):
Confirmed this fixes the exact same issue I hit on Windows 11 with Python 3.11 — `UnicodeEncodeError: 'charmap' codec can't encode character` when GRAPH_REPORT.md contains Unicode arrows (→). Had to work around it with `PYTHONUTF8=1` on every invocation. This PR is the proper fix. LGTM.

safishamsi:
Cherry-picked into v0.4.2 (encoding + hook CRLF fix) — thank you!

PR #212, opened 2026-04-11
fix: accept 'links' key from NetworkX <=3.1 alongside 'edges'

mrsyb:
NetworkX changed its node_link_data() default edge key from 'links' (<=3.1) to 'edges' (>=3.2). Any data serialised with an older NetworkX version — or any serialiser that follows the original JSON Graph Format spec — uses 'links', causing graphify to silently produce graphs with zero edges.

Changes:
- validate.py: fall back to data['links'] when data['edges'] is absent
- build.py: normalise 'links' -> 'edges' at the top of build_from_json()

Both fixes are backward-compatible: data already using 'edges' is unaffected.

Tests added:
- test_build_from_json_with_links_key  (old-style 'links' key)
- test_build_from_json_with_edges_key  (new-style 'edges' key, current NX)
- test_links_key_accepted_as_edges     (validate accepts 'links')
- test_edges_key_takes_priority_over_links  (priority order is correct)

### DIFF
--- a/graphify/build.py
+++ b/graphify/build.py
@@ -32,6 +32,9 @@ def build_from_json(extraction: dict, *, directed: bool = False) -> nx.Graph:
     directed=True produces a DiGraph that preserves edge direction (source→target).
     directed=False (default) produces an undirected Graph for backward compatibility.
     """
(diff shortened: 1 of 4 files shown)


safishamsi:
Cherry-picked into v0.4.2 — thank you!

PR #218, opened 2026-04-11
fix: normalize edge keys and prevent to_html zero division

### LINKED ISSUE #216: build_from_json crashes on edges with 'from'/'to' keys instead of 'source'/'target'
## Bug

`build_from_json` in `graphify/build.py` crashes with `KeyError: 'source'` when edges use alternate key names like `from`/`to` instead of the expected `source`/`target`.

## How to reproduce

When using LLM subagents (particularly smaller/faster models) for semantic extraction, they sometimes produce edges with `{"from": "node_a", "to": "node_b", ...}` instead of `{"source": "node_a", "target": "node_b", ...}`. This is a reasonable variation that the schema doesn't enforce at the subagent level.

When these edges reach `build_from_json`, line 45 does:

```python
src, tgt = edge["source"], edge["target"]
```

This raises `KeyError` and crashes the entire graph build, losing all extraction work (including paid LLM tokens).

## Expected behavior

`build_from_json` should normalize common alternate key names (`from` → `source`, `to` → `target`) before processing, and skip edges that are truly malformed (missing both variants).

## Impact

- Edges with valid relationships are silently lost or cause crashes
- Smaller models (Sonnet, Haiku) are more likely to produce alternate key names
- The crash happens _after_ the expensive LLM extraction, so the token cost is wasted

## Suggested fix

```python
for edge in extraction.get("edges", []):
    if "from" in edge and "source" not in edge:
        edge["source"] = edge.pop("from")
    if "to" in edge and "target" not in edge:
        edge["target"] = edge.pop("to")
    if "source" not in edge or "target" not in edge:
        continue
    src, tgt = edge["source"], edge["target"]
```

I have a PR ready for this: #217 (pending)

### LINKED ISSUE #217: to_html crashes with ZeroDivisionError on graphs with zero edges
## Bug

`to_html` in `graphify/export.py` crashes with `ZeroDivisionError` when the graph has nodes but zero edges.

## How to reproduce

Build a graph where all semantic edges were filtered out (e.g. malformed extraction) but AST nodes exist. The graph has nodes with degree 0.

```python
from graphify.export import to_html

# Graph with nodes but no edges
G = nx.Graph()
G.add_node("a", label="A")
G.add_node("b", label="B")
communities = {0: ["a", "b"]}
to_html(G, communities, "out.html")  # ZeroDivisionError
```

## Root cause

Line 349:
```python
max_deg = max(degree.values()) if degree else 1
```

When all nodes have degree 0, `max(degree.values())` returns 0 (not falsy), so the `if degree else 1` guard doesn't trigger. Then line 358:

```python
size = 10 + 30 * (deg / max_deg)  # division by zero
```

## Suggested fix

```python
max_deg = max(degree.values()) if degree else 1
if max_deg == 0:
    max_deg = 1
```

I have a PR ready for this: #217 (pending)

pedrolbacelar:
## Summary

Two small fixes for issues encountered when running graphify with LLM subagents on multi-repo codebases:

- **`build_from_json`**: normalize alternate edge key names (`from`/`to` → `source`/`target`) before processing. LLM subagents (especially smaller models like Sonnet) sometimes produce edges with `from`/`to` instead of the expected `source`/`target`. Previously these edges caused a `KeyError` crash or were silently lost. Now they are normalized transparently, and truly malformed edges (missing both variants) are skipped.

- **`to_html`**: guard against `ZeroDivisionError` when `max_deg` is 0. This happens on graphs where all nodes are isolated (no edges), e.g. very small repos or when all semantic edges were malformed. The node size calculation `deg / max_deg` crashed.

## Files changed

| File | Change |
|---|---|
| `graphify/build.py` | Normalize `from`/`to` → `source`/`target`, skip malformed edges |
| `graphify/export.py` | Guard `max_deg == 0` before division |

## How I found these

While indexing a multi-repo project (13 repos, ~4400 nodes total), I dispatched Sonnet subagents for semantic extraction. Two repos produced edges with `from`/`to` keys, causing `build_from_json` to crash. On one very small repo (5 code files, no docs), all edges were lost, leaving a 0-edge graph that then crashed `to_html`.

## Testing

- Verified both fixes resolve the original crashes
- No behavior change for well-formed inputs (the normalization only triggers when `source`/`target` are absent)
- No logic changes, no new dependencies

Closes #216
Closes #217

### DIFF
--- a/graphify/build.py
+++ b/graphify/build.py
@@ -42,6 +42,13 @@ def build_from_json(extraction: dict, *, directed: bool = False) -> nx.Graph:
         G.add_node(node["id"], **{k: v for k, v in node.items() if k != "id"})
     node_set = set(G.nodes())
     for edge in extraction.get("edges", []):
+        # Normalize alternate key names that LLM subagents sometimes produce
+        if "from" in edge and "source" not in edge:
+            edge["source"] = edge.pop("from")
+        if "to" in edge and "target" not in edge:
+            edge["target"] = edge.pop("to")
+        if "source" not in edge or "target" not in edge:
+            continue  # skip malformed edges
         src, tgt = edge["source"], edge["target"]
         if src not in node_set or tgt not in node_set:
             continue  # skip edges to external/stdlib nodes - expected, not an error
--- a/graphify/export.py
+++ b/graphify/export.py
@@ -347,6 +347,8 @@ def to_html(
     node_community = _node_community_map(communities)
     degree = dict(G.degree())
     max_deg = max(degree.values()) if degree else 1
+    if max_deg == 0:
+        max_deg = 1  # avoid division by zero on graphs with no edges
 
     # Build nodes list for vis.js
     vis_nodes = []


safishamsi:
Both fixes (#216 edge normalization, #217 ZeroDivisionError) were applied directly in v0.4.2 — thank you for the detailed write-up!

PR #223, opened 2026-04-11
fix(hooks): drop stale CODE_EXTS gate from post-commit hook

### LINKED ISSUE #222: post-commit hook silently skips rebuild on .tsx/.jsx and other valid code files (stale CODE_EXTS gate)
## Summary

The `post-commit` hook installed by `graphify hook install` (in [`graphify/hooks.py`](https://github.com/safishamsi/graphify/blob/v4/graphify/hooks.py)) gates the rebuild on a hardcoded `CODE_EXTS` allowlist that has drifted from the authoritative `CODE_EXTENSIONS` set in [`graphify/detect.py`](https://github.com/safishamsi/graphify/blob/v4/graphify/detect.py#L21). The result: commits that change valid, supported code files exit silently with status 0 and the graph is never rebuilt — so users see a stale `graph.json` and assume the hook isn't working.

## Repro

1. `graphify .` in any project (build the initial graph).
2. `graphify hook install`.
3. Commit a change to a `.tsx`, `.jsx`, `.lua`, `.ex`, `.ps1`, or `.jl` file.
4. Inspect `graphify-out/graph.json` mtime — unchanged. No log, no error, the hook just exited 0.

Same outcome for any markdown/config-only commit on a doc repo, even though `graphify` supports doc graphs and `watch.py` has a `_notify_only` fallback that the hook never reaches.

## Root cause

In `graphify/hooks.py` the embedded post-commit script defines its own `CODE_EXTS`:

```python
CODE_EXTS = {
    '.py', '.ts', '.js', '.go', '.rs', '.java', '.cpp', '.c', '.rb', '.swift',
    '.kt', '.cs', '.scala', '.php', '.cc', '.cxx', '.hpp', '.h', '.kts',
}
```

But `graphify/detect.py` (the source of truth used by `_rebuild_code`) ships:

```python
CODE_EXTENSIONS = {'.py', '.ts', '.js', '.jsx', '.tsx', '.go', '.rs', '.java', '.cpp', '.cc', '.cxx', '.c', '.h', '.hpp', '.rb', '.swift', '.kt', '.kts', '.cs', '.scala', '.php', '.lua', '.toc', '.zig', '.ps1', '.ex', '.exs', '.m', '.mm', '.jl'}
```

11 extensions missing from the hook's copy: `.jsx .tsx .lua .toc .zig .ps1 .ex .exs .m .mm .jl`. Worse, `_rebuild_code` calls `detect()` internally which rescans the whole project from scratch using the authoritative set anyway — so the hook's filter doesn't even *control* what gets rebuilt. It only gates *whether* the rebuild fires. The whole filter block is dead weight that introduces a stale-duplicate bug class.

## Impact

- React/Next/Vite repos (mostly `.tsx/.jsx`) — hook never fires.
- Lua, Elixir, PowerShell, Julia, Objective-C repos — hook never fires.
- Doc-heavy repos (skills, knowledge bases, READMEs) — hook never fires.
- Silent failure: no log line, no error, exit 0. Users discover the staleness only when query results disagree with the code.

## Proposed fix

Drop the filter entirely. The rebuild is cheap (~2s for 400 files, no LLM cost), `detect()` already handles "no code files found" cleanly, and removing the filter eliminates the duplication bug class permanently. PR incoming.

Bonus: log success/failure to a file under `graphify-out/` so silent failures become visible if anything else goes wrong.

ukanhaupa:
## Summary

- Fixes #222
- Removes the hardcoded `CODE_EXTS` allowlist from the embedded `post-commit` hook script in `graphify/hooks.py`. The list had drifted from the authoritative `CODE_EXTENSIONS` set in `graphify/detect.py` (missing `.tsx .jsx .lua .toc .zig .ps1 .ex .exs .m .mm .jl`), so commits touching only those files exited 0 silently and never rebuilt the graph.
- Routes hook output to `graphify-out/.hook.log` so any future failure stays visible instead of vanishing into git's hook stderr.

## Why drop the filter, not just sync it

`_rebuild_code` already calls `detect()` internally, which rescans the project from scratch using the authoritative `CODE_EXTENSIONS` set. So the hook's filter only ever gated *whether* the rebuild fires, not *what* gets rebuilt — the work it claimed to save was an illusion. Keeping any hardcoded copy here just re-introduces the same drift bug class the next time `detect.py` adds a language.

The rebuild itself is cheap (~2s for ~400 files in my test repos, no LLM cost), so running it on every commit is harmless. The simpler model — "if anything changed, rebuild" — is also easier to reason about and document.

## Test plan

Manually verified on a clean test repo:

- [x] Init repo, install patched hook via `graphify.hooks.install()`.
- [x] Commit a `.tsx` file (the canonical bug case from #222) → hook fires, `graphify-out/graph.json` and `GRAPH_REPORT.md` updated, log line written.
- [x] Commit a markdown-only change → hook fires, rebuild succeeds, log line written.
- [x] Confirmed `graphify-out/.hook.log` accumulates entries across commits with `[graphify hook] <iso-timestamp> - rebuilding...` headers.

```
$ cat graphify-out/.hook.log

[graphify hook] 2026-04-11T23:52:44 - rebuilding...
[graphify watch] Rebuilt: 3 nodes, 1 edges, 2 communities
[graphify watch] graph.json and GRAPH_REPORT.md updated in graphify-out

[graphify hook] 2026-04-11T23:52:51 - rebuilding...
[graphify watch] Rebuilt: 3 nodes, 1 edges, 2 communities
[graphify watch] graph.json and GRAPH_REPORT.md updated in graphify-out
```

## Notes for reviewers

- Switched the inline `python -c` block to use *single*-quoted shell strings so we can drop a layer of escaping. The embedded Python uses double quotes for the `Path(".")` literal, which is now safe.
- The `_PYTHON_DETECT` block is unchanged.
- `_CHECKOUT_SCRIPT` is intentionally not touched in this PR. It already calls `_rebuild_code` directly without an extension gate, so it doesn't have the same bug. Happy to align its logging with the new `.hook.log` pattern in a follow-up if you'd like.

### DIFF
--- a/graphify/hooks.py
+++ b/graphify/hooks.py
@@ -32,8 +32,16 @@ fi
 
 _HOOK_SCRIPT = """\
 # graphify-hook-start
-# Auto-rebuilds the knowledge graph after each commit (code files only, no LLM needed).
+# Auto-rebuilds the knowledge graph after each commit (no LLM needed).
 # Installed by: graphify hook install
+#
+# Why no extension filter: _rebuild_code internally calls detect() which
+# rescans the project from scratch using the authoritative CODE_EXTENSIONS
+# set in graphify/detect.py. A hook-side filter would only gate WHETHER the
+# rebuild fires, not WHAT gets rebuilt — and any hardcoded copy here drifts
+# from the real allowlist (.tsx/.jsx/.lua/.ex/.jl/... were missed in the past).
+# The rebuild is cheap (~2s, no LLM), so just always run it. Output is
+# appended to graphify-out/.hook.log so silent failures stay visible.
 
 CHANGED=$(git diff --name-only HEAD~1 HEAD 2>/dev/null || git diff --name-only HEAD 2>/dev/null)
 if [ -z "$CHANGED" ]; then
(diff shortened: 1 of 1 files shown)


safishamsi:
The CODE_EXTS removal was applied directly in v0.4.2. We kept the simpler approach (rebuild on any change) but didn't add the .hook.log file — keeping hooks minimal. Thanks!

PR #227, opened 2026-04-11
fix: filter blank stdin lines to prevent MCP JSONRPCMessage parse errors

### LINKED ISSUE #201: Error running MCP service
```bash

Received exception from stream: 1 validation error for JSONRPCMessage
  Invalid JSON: EOF while parsing a value at line 2 column 0 [type=json_invalid, input_value='\n', input_type=str]
    For further information visit https://errors.pydantic.dev/2.12/v/json_invalid
{"method":"notifications/message","params":{"level":"error","logger":"mcp.server.exception_handler","data":"Internal Server Error"},"jsonrpc":"2.0"}
```

yetval:
Fixes #201.

Some MCP clients send blank lines between JSON messages. The MCP stdio transport tries to parse every line as a `JSONRPCMessage`, so a bare `\n` triggers a Pydantic `ValidationError`.

`_filter_blank_stdin()` installs an OS-level pipe that relays stdin while silently dropping blank lines. Because it operates at the file-descriptor level, it is transparent to anyio regardless of how it wraps stdin.

### DIFF
--- a/graphify/serve.py
+++ b/graphify/serve.py
@@ -108,6 +108,29 @@ def _find_node(G: nx.Graph, label: str) -> list[str]:
             if term in d.get("label", "").lower() or term == nid.lower()]
 
 
+def _filter_blank_stdin() -> None:
+    import os
+    import threading
+
+    r_fd, w_fd = os.pipe()
+    saved_fd = os.dup(sys.stdin.fileno())
+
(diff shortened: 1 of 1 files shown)


safishamsi:
Applied from our end in v0.4.5 — thank you for the fix!

PR #235, opened 2026-04-12
feat(php): capture static property access as uses_static_prop edges

### LINKED ISSUE #234: PHP extractor misses scoped_property_access_expression (static property access)
## Summary

The PHP extractor wires up `function_call_expression` and `member_call_expression` into `calls` edges, but it never visits `scoped_property_access_expression`. That means references to static properties such as `DefaultPalette::$primary` or `Config::$default` do not produce any graph edges, so classes that expose configuration or theme defaults through public static properties disappear from the graph even when they are read heavily.

## Current behavior

Minimal reproducer:

```php
<?php

namespace App\Theme;

class DefaultPalette
{
    public static string $primary = '#3366ff';
}

class ColorResolver
{
    public function primary(): string
    {
        return DefaultPalette::$primary;
    }
}
```

Running `extract_php()` on this file returns both classes, the method, the static property declaration and the usual `contains`/`method` edges. No edge connects `ColorResolver` to `DefaultPalette`.

## Expected behavior

The reading function or method should be linked to the class that owns the static property, with a relation label that is distinct from `calls`. For example `uses_static_prop` keeps the intent obvious without polluting `calls`, which is reserved for callable invocations.

## Proposed solution

Extend `LanguageConfig` with an optional `static_prop_types` frozenset and populate it in `_PHP_CONFIG` with `scoped_property_access_expression`. In the `walk_calls` dispatcher, handle the new node type by reading its class reference (via the `scope` field with a positional name fallback, since tree-sitter-php has exposed the class name under different field names across versions) and adding a `uses_static_prop` edge when the class resolves in `label_to_nid`.

The change is additive. Languages that do not set `static_prop_types` see no behavioral difference. I already have a patch with tests ready.

## Notes

This is the last of three related issues for the PHP extractor. The other two are filed separately for `class_constant_access_expression` and `scoped_call_expression`. All three share the same shape, so each can be reviewed and merged independently.

erhanurgun:
Fixes #234

## What

The PHP AST extractor now recognizes `scoped_property_access_expression` nodes and produces a new `uses_static_prop` edge from the enclosing function or method to the class that owns the referenced static property. Previously these reads were silently dropped, so classes that expose configuration or theme defaults through public static properties disappeared from the graph.

## Why

Current PHP handling only wires up `function_call_expression` and `member_call_expression`. A reference like `DefaultPalette::$primary` does not become an edge. On codebases that rely on static configuration holders this loses meaningful coupling.

Full details and reproducer are in issue #234.

## How

- `LanguageConfig` gains an optional `static_prop_types` frozenset with an empty default, so every other language is untouched.
- `_PHP_CONFIG` sets `static_prop_types=frozenset({"scoped_property_access_expression"})`.
- A new branch inside the existing `walk_calls` dispatcher handles the node. It reads the class reference via the `scope` field and falls back to the first named child, because tree-sitter-php has exposed the class name under slightly different field layouts across versions.
- When the class resolves through `label_to_nid`, a `uses_static_prop` edge is emitted. A dedicated `seen_static_ref_pairs` set prevents duplicate static edges while allowing `calls` and `uses_static_prop` to coexist for the same `(caller, target)` pair when legitimate.

## Tests

- `tests/fixtures/sample_php_static_prop.php` is a minimal fixture that declares `DefaultPalette` with two static properties and a `ColorResolver` class whose methods read both of them.
- `tests/test_languages.py` gains two tests: `test_php_finds_static_property_access` checks the relation is present, and `test_php_static_prop_target_is_holding_class` verifies the target is the class that holds the property.

## Backwards compatibility

The change is additive. Other language configs never populate `static_prop_types`. Existing PHP tests continue to pass against the unchanged `tests/fixtures/sample.php` fixture.

## Follow-ups

This is the last of three related patches for the PHP extractor. The other two cover `class_constant_access_expression` and `scoped_call_expression`, opened as separate focused PRs.

Marked as draft until you have a chance to weigh in on the approach and the new relation name. I am happy to rename or reshape.

### DIFF
--- a/graphify/extract.py
+++ b/graphify/extract.py
@@ -29,6 +29,7 @@ class LanguageConfig:
     function_types: frozenset = frozenset()
     import_types: frozenset = frozenset()
     call_types: frozenset = frozenset()
+    static_prop_types: frozenset = frozenset()
 
     # Name extraction
     name_field: str = "name"
@@ -536,6 +537,7 @@ _PHP_CONFIG = LanguageConfig(
     function_types=frozenset({"function_definition", "method_declaration"}),
     import_types=frozenset({"namespace_use_clause"}),
     call_types=frozenset({"function_call_expression", "member_call_expression"}),
+    static_prop_types=frozenset({"scoped_property_access_expression"}),
     call_function_field="function",
     call_accessor_node_types=frozenset({"member_call_expression"}),
     call_accessor_field="name",
@@ -849,6 +851,7 @@ def _extract_generic(path: Path, config: LanguageConfig) -> dict:
         label_to_nid[normalised.lower()] = n["id"]
 
     seen_call_pairs: set[tuple[str, str]] = set()
+    seen_static_ref_pairs: set[tuple[str, str, str]] = set()
 
     def walk_calls(node, caller_nid: str) -> None:
         if node.type in config.function_boundary_types:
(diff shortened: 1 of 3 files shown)


safishamsi:
Implemented in v0.4.9 — static property access (`Foo::$bar`) produces `uses_static_prop` edges. Upgrade: `pip install --upgrade graphifyy`

PR #237, opened 2026-04-12
feat(php): capture config() helper calls as uses_config edges

### LINKED ISSUE #236: PHP extractor drops config('key.path') helper calls and similar dot-key helpers
## Summary

Projects that resolve runtime configuration through a global helper function, the canonical PHP example is `config('foo.bar')`, lose all the associated coupling when they are graphified. The extractor sees `config(...)` as a plain `function_call_expression`, tries to map the name `config` through `label_to_nid` (which usually has nothing), and drops the edge. No relationship is produced between the caller and the configuration holder that owns the key.

## Current behavior

Minimal reproducer:

```php
<?php

namespace App\Support;

class Throttle
{
    public const int DEFAULT_PER_SECOND = 60;
}

class RateLimiter
{
    public function perSecond(): int
    {
        return (int) config('throttle.api.per_second', 60);
    }
}
```

Running `extract_php()` on this file returns both classes and the method, plus the usual `contains`/`method` edges. No edge connects `RateLimiter` to `Throttle` even though the method clearly reads `throttle.*` configuration.

## Expected behavior

When the extractor sees a helper call whose first argument is a dot-notation string literal, it should use the first segment of that string as the target lookup key. If the key resolves to an existing node (a class named `Throttle`, a file named `throttle.php`, or similar), an edge should be produced with a relation that identifies the helper (for example `uses_config`).

## Proposed solution

Extend `LanguageConfig` with an optional `helper_fn_names` frozenset (default empty). `_PHP_CONFIG` populates it with `config`. Inside the existing `walk_calls` dispatcher, when a `function_call_expression` matches a helper name, the first string literal argument is parsed. Its first dot-separated segment becomes the lookup key, and the extractor tries both the bare name and a `.php` suffix before giving up. When a target is found, a `uses_{helper}` edge is emitted.

The design is intentionally extensible. Adding more helpers later (for example `route()`, `view()`, `env()`, `asset()`) only needs a wider `helper_fn_names` frozenset and possibly helper-specific parsing rules.

The change is additive. Languages that never set `helper_fn_names` see no behavioral difference.

## Notes

This is the next in a small family of PHP extraction gaps I am filing as separate, focused issues. The others cover `class_constant_access_expression`, `scoped_call_expression`, and `scoped_property_access_expression`. I have a PR ready for this one with tests.

erhanurgun:
Fixes #236

## What

The PHP extractor now recognizes helper function calls whose first argument is a dot-notation string literal and produces a new `uses_{helper}` edge pointing from the caller to the target indicated by the first segment of that string. The first helper handled out of the box is `config()`, which emits `uses_config` edges.

## Why

Global helper calls like `config('throttle.api.per_second')` carry real coupling between the caller and the configuration holder that owns the key, but the graph loses that signal entirely. The extractor sees `config` as an unknown name and drops the edge.

Full details and reproducer are in issue #236.

## How

- `LanguageConfig` gains an optional `helper_fn_names` frozenset with an empty default, so every other language is untouched.
- `_PHP_CONFIG` sets `helper_fn_names=frozenset({"config"})`.
- A new branch inside `walk_calls` runs after the regular call dispatch. When the current node is a `function_call_expression` whose callee name matches one of `helper_fn_names`, the first `argument` child is walked and its first `string_content` payload is captured.
- The first dot-separated segment of that payload is used as the lookup key. The extractor tries both the bare name and a `.php` suffix against `label_to_nid` before giving up. When a target is resolved, a `uses_{callee_name}` edge is emitted (so `config('throttle.foo')` becomes `uses_config`).
- A dedicated `seen_helper_ref_pairs` set prevents duplicates and keeps this pathway independent from `seen_call_pairs`, which is scoped to `calls` edges.

The design is extensible. Adding more helpers later (for example `route()`, `view()`, `env()`, `asset()`) only requires widening `helper_fn_names` and, if needed, teaching the segment-resolution step about helper-specific targets.

## Tests

- `tests/fixtures/sample_php_config.php` declares a `Throttle` class and a `RateLimiter` class that calls `config('throttle.api.per_second', ...)` and `config('throttle.api.per_day', ...)`.
- `tests/test_languages.py` gains two tests: `test_php_finds_config_helper_call` verifies the relation is present, and `test_php_config_helper_target_matches_first_segment` checks the target is the `Throttle` class (first dot segment).

## Backwards compatibility

The change is additive. Other languages never set `helper_fn_names`. Existing PHP tests still pass against the unchanged `tests/fixtures/sample.php` fixture. Emitted edges use a new relation label, so any downstream tooling that filters on `calls` or the older relations is unaffected.

## Follow-ups

Related focused PRs are open or being prepared for `class_constant_access_expression`, `scoped_call_expression`, `scoped_property_access_expression`, service container `bind()` calls, `$listen` array mapping, and `.blade.php` file handling.

Marked as draft until you have a chance to weigh in on the approach, in particular the relation name `uses_config` and the extensibility shape of `helper_fn_names`.

### DIFF
--- a/graphify/extract.py
+++ b/graphify/extract.py
@@ -29,6 +29,7 @@ class LanguageConfig:
     function_types: frozenset = frozenset()
     import_types: frozenset = frozenset()
     call_types: frozenset = frozenset()
+    helper_fn_names: frozenset = frozenset()
 
     # Name extraction
     name_field: str = "name"
@@ -536,6 +537,7 @@ _PHP_CONFIG = LanguageConfig(
     function_types=frozenset({"function_definition", "method_declaration"}),
     import_types=frozenset({"namespace_use_clause"}),
     call_types=frozenset({"function_call_expression", "member_call_expression"}),
+    helper_fn_names=frozenset({"config"}),
     call_function_field="function",
     call_accessor_node_types=frozenset({"member_call_expression"}),
     call_accessor_field="name",
@@ -849,6 +851,7 @@ def _extract_generic(path: Path, config: LanguageConfig) -> dict:
         label_to_nid[normalised.lower()] = n["id"]
 
     seen_call_pairs: set[tuple[str, str]] = set()
+    seen_helper_ref_pairs: set[tuple[str, str, str]] = set()
 
     def walk_calls(node, caller_nid: str) -> None:
         if node.type in config.function_boundary_types:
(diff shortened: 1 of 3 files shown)


safishamsi:
Implemented in v0.4.9 — `config()` helper calls produce `uses_config` edges pointing to the first key segment. Upgrade: `pip install --upgrade graphifyy`

PR #239, opened 2026-04-12
feat(php): capture service container bindings as bound_to edges

### LINKED ISSUE #238: PHP extractor misses service container bind/singleton/scoped/instance calls
## Summary

Dependency injection container calls of the shape `$this->app->bind(FooContract::class, BarImpl::class)` (or `singleton`, `scoped`, `instance`) express the core of a project's runtime wiring, but the extractor does not produce any edge for them. Reviewers cannot see from the graph which implementation a contract resolves to.

## Current behavior

Minimal reproducer:

```php
<?php

namespace App\Providers;

class PaymentGateway {}
class StripeGateway {}

class AppServiceProvider
{
    public function register(): void
    {
        $this->app->bind(PaymentGateway::class, StripeGateway::class);
    }
}
```

Running `extract_php()` on this file returns all three classes, the method and the usual structural edges. No edge connects `PaymentGateway` to `StripeGateway`, even though at runtime the former always resolves to the latter.

## Expected behavior

When the walker sees a `member_call_expression` whose method name is one of `bind`, `singleton`, `scoped`, `instance`, it should look at the first two arguments. If both are `Foo::class` literals, an edge should be produced from the first (contract) to the second (implementation). The relation label should make the runtime meaning clear, for example `bound_to`.

## Proposed solution

Extend `LanguageConfig` with an optional `container_bind_methods` frozenset. `_PHP_CONFIG` populates it with `{bind, singleton, scoped, instance}`. The existing `walk_calls` dispatcher gains a new branch that parses the first two arguments of a matching call, pulls class names out of `class_constant_access_expression` nodes, and emits a `bound_to` edge when both resolve through `label_to_nid`.

The change stays additive. Languages that do not set `container_bind_methods` see no behavioral difference.

## Notes

This is part of a small family of PHP extraction gaps I am filing as separate, focused issues. The others cover static access (const, call, property), the `config()` helper, `$listen` array mapping and `.blade.php` handling. I have a PR ready for this one with tests.

erhanurgun:
Fixes #238

## What

The PHP AST extractor now recognizes service container binding calls like `$this->app->bind(FooContract::class, BarImpl::class)` and produces a new `bound_to` edge from the contract class to the implementation class. The same handling applies to `singleton`, `scoped`, and `instance`. Before this change there was no way to see a project's DI wiring in the graph.

## Why

Dependency injection container calls express which implementation a contract resolves to at runtime. Without these edges, the graph is silent about a whole layer of coupling that drives actual program behavior.

Full details and reproducer are in issue #238.

## How

- `LanguageConfig` gains an optional `container_bind_methods` frozenset with an empty default, so every other language is untouched.
- `_PHP_CONFIG` sets `container_bind_methods=frozenset({"bind", "singleton", "scoped", "instance"})`.
- A small helper, `_php_class_const_scope`, pulls the class name out of a `class_constant_access_expression`, handling both the `scope` field and the positional name fallback used by older tree-sitter-php layouts.
- A new branch inside the `walk_calls` dispatcher runs on `member_call_expression` nodes. When the callee name matches one of `container_bind_methods`, it walks the first two `argument` children, extracts a class name from each if the inner shape is `Foo::class`, and emits a `bound_to` edge from the contract node to the implementation node when both resolve through `label_to_nid`.
- A dedicated `seen_bind_pairs` set prevents duplicates and keeps this pathway independent from `seen_call_pairs`.

## Tests

- `tests/fixtures/sample_php_container.php` declares `PaymentGateway`, `StripeGateway`, `CashierGateway`, and an `AppServiceProvider` that issues both a `bind()` and a `singleton()` call.
- `tests/test_languages.py` gains two tests: `test_php_finds_container_bind` checks the relation is present, and `test_php_container_bind_links_contract_to_implementation` verifies the exact contract-to-implementation pairs appear.

## Backwards compatibility

The change is additive. Other language configs never populate `container_bind_methods`. Existing PHP tests still pass against the unchanged `tests/fixtures/sample.php` fixture. Closures and factory callbacks that do not pass two class literals are silently skipped.

## Follow-ups

Related focused PRs are open or being prepared for `class_constant_access_expression`, `scoped_call_expression`, `scoped_property_access_expression`, `config()`, `$listen` array mapping, and `.blade.php` handling.

Marked as draft until you weigh in on the approach, in particular the new relation name `bound_to` and the choice to emit a single contract-to-implementation edge instead of two provider-to-class edges.

### DIFF
--- a/graphify/extract.py
+++ b/graphify/extract.py
@@ -29,6 +29,7 @@ class LanguageConfig:
     function_types: frozenset = frozenset()
     import_types: frozenset = frozenset()
     call_types: frozenset = frozenset()
+    container_bind_methods: frozenset = frozenset()
 
     # Name extraction
     name_field: str = "name"
@@ -536,6 +537,7 @@ _PHP_CONFIG = LanguageConfig(
     function_types=frozenset({"function_definition", "method_declaration"}),
     import_types=frozenset({"namespace_use_clause"}),
     call_types=frozenset({"function_call_expression", "member_call_expression"}),
+    container_bind_methods=frozenset({"bind", "singleton", "scoped", "instance"}),
     call_function_field="function",
     call_accessor_node_types=frozenset({"member_call_expression"}),
     call_accessor_field="name",
@@ -849,6 +851,18 @@ def _extract_generic(path: Path, config: LanguageConfig) -> dict:
         label_to_nid[normalised.lower()] = n["id"]
 
     seen_call_pairs: set[tuple[str, str]] = set()
+    seen_bind_pairs: set[tuple[str, str, str]] = set()
+
+    def _php_class_const_scope(n) -> str | None:
+        scope = n.child_by_field_name("scope")
+        if scope is None:
+            for c in n.children:
+                if c.is_named and c.type in ("name", "qualified_name", "identifier"):
+                    scope = c
+                    break
+        if scope is None:
+            return None
+        return _read_text(scope, source)
 
     def walk_calls(node, caller_nid: str) -> None:
         if node.type in config.function_boundary_types:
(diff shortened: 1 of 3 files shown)


safishamsi:
Implemented in v0.4.9 — service container bindings (`bind`, `singleton`, `scoped`, `instance`) produce `bound_to` edges. Upgrade: `pip install --upgrade graphifyy`

PR #241, opened 2026-04-12
feat(php): capture $listen array mapping as listened_by edges

### LINKED ISSUE #240: PHP extractor ignores $listen style array mappings in class properties
## Summary

Event-driven PHP projects often declare their event to listener mappings in a static class property. The canonical form, seen in Laravel EventServiceProvider classes but broadly applicable, is `protected $listen = [Event::class => [Listener::class, ...]]`. These mappings drive runtime behavior, but the extractor walks past them silently and the relationships never reach the graph.

## Current behavior

Minimal reproducer:

```php
<?php

namespace App\Providers;

class UserRegistered {}
class SendWelcomeEmail {}
class NotifyAdmins {}

class EventServiceProvider
{
    protected $listen = [
        UserRegistered::class => [
            SendWelcomeEmail::class,
            NotifyAdmins::class,
        ],
    ];
}
```

Running `extract_php()` on this file returns all four classes and the property declaration, with the usual structural edges, but there is nothing that connects `UserRegistered` to either listener. The property array literal is completely opaque to the graph.

## Expected behavior

When the walker encounters a class property whose name matches a configured list (for example `listen`, `subscribe`), it should inspect the array literal and emit edges of the form `Event -> Listener` with a clear relation, for example `listened_by`. Every listener in the value array should produce its own edge.

## Proposed solution

Extend `LanguageConfig` with an optional `event_listener_properties` frozenset (default empty). `_PHP_CONFIG` populates it with `{listen, subscribe}`. The existing `walk()` recursion gains a branch that fires when a `property_declaration` sits inside a class and its property name is in the configured set. For each `event_class => [listener_class, ...]` entry, raw class names are recorded along with the source line in a pending list. After `walk_calls` runs (once `label_to_nid` is fully populated), a short pass resolves the pending pairs and emits `listened_by` edges for those that point to known nodes.

The change stays additive. Languages that do not populate `event_listener_properties` see no behavioral difference.

## Notes

This is part of a small family of PHP extraction gaps I am filing as separate, focused issues. The others cover static access (const, call, property), the `config()` helper, service container bindings, and `.blade.php` handling. I have a PR ready for this one with tests.

erhanurgun:
Fixes #240

## What

The PHP AST extractor now understands static class properties whose value is an event to listener mapping, and produces `listened_by` edges from each event class to each of its listener classes. Out of the box the property names `listen` and `subscribe` are recognized, which covers the common PHP patterns (Laravel `EventServiceProvider::$listen` and related event subscriber shapes).

## Why

Event-driven projects encode significant runtime behavior in these array literals, but the graph was blind to them. Without the edges, any query or analysis that looks at how events propagate through the system is missing a critical layer.

Full details and reproducer are in issue #240.

## How

- `LanguageConfig` gains an optional `event_listener_properties` frozenset with an empty default, so every other language is untouched.
- `_PHP_CONFIG` sets `event_listener_properties=frozenset({"listen", "subscribe"})`.
- The extractor keeps a `pending_listen_edges` list alongside the existing state. In the `walk` recursion, when a `property_declaration` sits inside a class and its property name matches the configured set, the walker inspects the array literal and, for every `Event::class => [Listener::class, ...]` entry, records raw class names (plus the source line of the listener entry).
- After the call-graph pass runs and `label_to_nid` is populated, a short post-pass resolves the pending pairs and emits `listened_by` edges for those that resolve to known nodes. A `seen_listen_pairs` set deduplicates.

## Tests

- `tests/fixtures/sample_php_listen.php` declares two events, three listeners, and an `EventServiceProvider` that maps them. One event has two listeners, the other has one.
- `tests/test_languages.py` gains two tests: `test_php_finds_listen_array_mapping` checks the relation is present, and `test_php_listen_array_covers_all_listeners` asserts every expected `(event, listener)` pair is present.

## Backwards compatibility

The change is additive. Other language configs never populate `event_listener_properties`. Existing PHP tests continue to pass against the unchanged `tests/fixtures/sample.php` fixture. Array shapes that do not fit the `Event::class => [Listener::class, ...]` pattern are silently skipped.

## Follow-ups

Related focused PRs are open or being prepared for `class_constant_access_expression`, `scoped_call_expression`, `scoped_property_access_expression`, `config()`, service container bindings, and `.blade.php` handling.

Marked as draft until you have a chance to weigh in on the approach. I am happy to rename the relation (for example `triggers` instead of `listened_by`) or reshape the configuration surface.

### DIFF
--- a/graphify/extract.py
+++ b/graphify/extract.py
@@ -29,6 +29,7 @@ class LanguageConfig:
     function_types: frozenset = frozenset()
     import_types: frozenset = frozenset()
     call_types: frozenset = frozenset()
+    event_listener_properties: frozenset = frozenset()
 
     # Name extraction
     name_field: str = "name"
@@ -536,6 +537,7 @@ _PHP_CONFIG = LanguageConfig(
     function_types=frozenset({"function_definition", "method_declaration"}),
     import_types=frozenset({"namespace_use_clause"}),
     call_types=frozenset({"function_call_expression", "member_call_expression"}),
+    event_listener_properties=frozenset({"listen", "subscribe"}),
     call_function_field="function",
     call_accessor_node_types=frozenset({"member_call_expression"}),
     call_accessor_field="name",
@@ -649,6 +651,7 @@ def _extract_generic(path: Path, config: LanguageConfig) -> dict:
     edges: list[dict] = []
     seen_ids: set[str] = set()
     function_bodies: list[tuple[str, object]] = []
+    pending_listen_edges: list[tuple[str, str, int]] = []
 
     def add_node(nid: str, label: str, line: int) -> None:
         if nid not in seen_ids:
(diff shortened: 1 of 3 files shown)


safishamsi:
Implemented in v0.4.9 — `$listen`/`$subscribe` event listener arrays produce `listened_by` edges. Upgrade: `pip install --upgrade graphifyy`

PR #243, opened 2026-04-12
feat(blade): regex-based extractor for .blade.php templates

### LINKED ISSUE #242: Blade templates (.blade.php) are invisible to the extractor
## Summary

Blade template files (`.blade.php`) carry a large share of the UI layer in many PHP projects, but they produce zero signal in the graph. The extractor does not recognize the extension at all, so every template ends up either skipped or routed through the PHP tree-sitter parser (which treats the directives as noise). On real Laravel projects this manifests as hundreds of isolated template nodes with no edges, which poisons community detection and dependency analysis.

## Current behavior

Minimal reproducer:

```blade
<div class="panel">
    <livewire:panel.listings :user="$user" />

    <flux:button wire:click="bulkActivate">Activate selected</flux:button>
    <flux:button wire:click="bulkDelete">Delete selected</flux:button>

    @include('partials.footer')
    @include('panel.components.pagination')
</div>
```

Running the pipeline on a folder that contains this file gives either nothing at all, because `.blade.php` is not in the collect set, or a malformed PHP parse if the file is misrouted through the PHP tree-sitter parser. In both cases the relationships hinted at by `wire:click`, `<livewire:...>` and `@include` are gone.

## Expected behavior

Even without a full Blade grammar, the extractor should recognize the three directive shapes that carry the most useful signal:

- `@include('path.to.partial')` - include edge to another template
- `<livewire:component.name />` - component usage edge
- `wire:click="methodName"` - method binding edge

Each match should produce an edge from the template file to a node that identifies the target (raw name is fine as a first step; a later resolver can map it to real nodes).

## Proposed solution

Add a lightweight `extract_blade(path)` helper that reads the source file and runs three regex passes over it. Every match becomes an edge from the file to a placeholder node that carries the raw target label. The placeholders stop the template from looking isolated and give a downstream resolver something to bind against later.

Teach the `extract()` dispatcher to check `path.name.endswith('.blade.php')` before its suffix lookup, because `Path.suffix` returns `.php` for `foo.blade.php` and we need to avoid routing templates through the tree-sitter-php parser. Extend `collect_files` and the filename walk so that `.blade.php` files are actually picked up.

Blade has no mature tree-sitter grammar, so regex is intentional. Scope is narrow on purpose: only the three most common directive shapes. More advanced Blade features (`@if`, `@foreach`, `@section`/`@yield`, custom directives) stay out of scope until there is demand.

## Notes

This is the last in a small family of PHP extraction gaps I am filing as separate, focused issues. The others cover static access (const, call, property), `config()`, service container bindings, and `$listen` arrays. I have a PR ready for this one with fixtures and tests.

erhanurgun:
Fixes #242

## What

A new `extract_blade()` reads `.blade.php` files and produces edges for the three Blade directive shapes that carry the most graph signal. The `extract()` dispatcher routes templates to it based on filename, which works around the fact that `Path.suffix` returns `.php` for `foo.blade.php`.

The three handled shapes:

- `@include('path.to.partial')` produces `includes`
- `<livewire:component.name />` produces `uses_component`
- `wire:click="methodName"` produces `wire_click`

## Why

Templates often make up a large share of a PHP project's surface area but they were silent in the graph. On a real Laravel project this left hundreds of template files as isolated nodes with no edges, which poisoned community detection and dependency analysis.

Full details and reproducer are in issue #242.

## How

- New `extract_blade(path)` helper alongside the other `extract_*` functions. It reads the source as UTF-8 (with `errors="replace"` to be defensive), creates a file node, and runs three compiled regexes (`_BLADE_INCLUDE_RE`, `_BLADE_LIVEWIRE_RE`, `_BLADE_WIRE_CLICK_RE`).
- Every match creates a placeholder node keyed by a stable id (so duplicate directives do not produce duplicate nodes) and an edge from the file node to the placeholder. The placeholder carries the raw target label. This is enough to stop the template looking isolated in the graph and gives a later resolver something to map to real nodes.
- The `extract()` dispatcher checks `path.name.endswith('.blade.php')` before its suffix lookup and routes matches to `extract_blade` instead of `extract_php`. This is the one piece that has to live in the dispatcher, because `Path.suffix` alone cannot distinguish `foo.blade.php` from `foo.php`.

Blade has no mature tree-sitter grammar, so regex is intentional. The scope is deliberately narrow: only the three most common directive shapes. More advanced Blade features (`@if`, `@foreach`, `@section`/`@yield`, custom directives) are out of scope for this PR and can follow later.

## Tests

- `tests/fixtures/sample.blade.php` exercises all three directive shapes.
- `tests/test_languages.py` gains four tests: `test_blade_no_error`, `test_blade_finds_include`, `test_blade_finds_livewire_component`, `test_blade_finds_wire_click`. Each verifies both the relation and the expected placeholder target labels.
- Full suite stays green: `pytest` reports 429 passed.

## Backwards compatibility

Additive. Projects without `.blade.php` files are unaffected. Existing PHP tests still pass against `tests/fixtures/sample.php`. The dispatcher change does not touch any non-Blade path.

## Follow-ups

Related focused PRs are open or being prepared for the PHP extractor: `class_constant_access_expression`, `scoped_call_expression`, `scoped_property_access_expression`, `config()`, service container bindings, and `$listen` array mapping.

Marked as draft until you have a chance to weigh in on the approach, in particular whether you would rather route Blade through a new `LanguageConfig` entry or keep it as its own bespoke extractor.

### DIFF
--- a/graphify/extract.py
+++ b/graphify/extract.py
@@ -1141,6 +1141,76 @@ def extract_php(path: Path) -> dict:
     return _extract_generic(path, _PHP_CONFIG)
 
 
+# Blade directive patterns. Conservative on purpose: only the three forms that
+# carry the most graph signal, not the full Blade syntax.
+_BLADE_INCLUDE_RE = re.compile(r"""@include\s*\(\s*['"]([^'"]+)['"]""")
+_BLADE_LIVEWIRE_RE = re.compile(r"<livewire:([\w\-\.]+)")
+_BLADE_WIRE_CLICK_RE = re.compile(r"""wire:click\s*=\s*["'](\w+)""")
+
+
+def extract_blade(path: Path) -> dict:
+    """Extract basic Blade template references: @include, <livewire:*>, wire:click.
+
+    Blade has no mature tree-sitter grammar, so this uses regex on three common
+    forms. Every match becomes an edge from the template file to a placeholder
+    node that carries the raw target name. Downstream passes can resolve those
+    names against real nodes; if nothing resolves, the placeholder still keeps
+    the template from looking isolated in the graph.
+    """
+    try:
+        source = path.read_text(encoding="utf-8", errors="replace")
+    except Exception as e:  # noqa: BLE001
+        return {"nodes": [], "edges": [], "error": str(e)}
+
+    str_path = str(path)
+    file_nid = _make_id(str(path))
+    stem = path.name
+
+    nodes: list[dict] = [{
+        "id": file_nid,
+        "label": stem,
+        "file_type": "code",
+        "source_file": str_path,
+        "source_location": "L1",
+    }]
+    edges: list[dict] = []
+    seen_targets: set[tuple[str, str]] = set()
+
+    def _line_for(offset: int) -> int:
+        return source.count("\n", 0, offset) + 1
+
+    def _add(target_label: str, relation: str, offset: int) -> None:
+        target_id = _make_id("blade", target_label)
+        key = (target_id, relation)
+        if key not in seen_targets:
+            seen_targets.add(key)
+            nodes.append({
+                "id": target_id,
+                "label": target_label,
+                "file_type": "code",
+                "source_file": "",
+                "source_location": "",
+            })
+        edges.append({
+            "source": file_nid,
+            "target": target_id,
+            "relation": relation,
+            "confidence": "EXTRACTED",
+            "source_file": str_path,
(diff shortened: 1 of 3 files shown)


safishamsi:
Implemented in v0.4.7 — `.blade.php` files are extracted via regex (`@include`, `<livewire:>`, `wire:click`). Upgrade: `pip install --upgrade graphifyy`

Issue #244, opened 2026-04-12
Windows: post-commit and post-checkout hooks fail with 'python3: command not found'


arkritico:
## Summary

On Windows, `graphify hook install` (v0.4.2) writes `post-commit` and `post-checkout` hooks that hardcode `python3` as their fallback interpreter. Windows Python installs use `python` (no `python3` shim), so every commit and checkout produces:

```
.git/hooks/post-commit: line 32: python3: command not found
```

The hook runs as a silent no-op, so it's not fatal, but the knowledge graph never rebuilds — defeating the purpose of the hook install.

## Root cause

The hooks try to sniff the Python interpreter from the `graphify` binary's shebang line:

```bash
GRAPHIFY_BIN=$(command -v graphify 2>/dev/null)
if [ -n "$GRAPHIFY_BIN" ]; then
    _SHEBANG=$(head -1 "$GRAPHIFY_BIN" | sed 's/^#![[:space:]]*//')
    ...
```

On Windows, `graphify` is a **`.exe` wrapper** (not a Python script with a shebang). `head -1` of an .exe returns binary garbage starting with `MZ`, the allowlist filter rejects it, and the hook falls through to the hardcoded `GRAPHIFY_PYTHON="python3"` fallback. Since Windows doesn't have `python3` on PATH, execution fails.

## Repro

1. Windows 10/11 with Python 3.x installed (so `python` is on PATH but `python3` is not)
2. `pip install graphifyy`
3. `cd some-repo && graphify hook install`
4. Make any commit
5. Observe: `python3: command not found` on stderr, no graph rebuild happens

## Fix

Both `post-commit` and `post-checkout` hooks need a cross-platform fallback chain. Suggested patch:

```sh
# Fall back: try python3, then python (Windows), else abort silently
if [ -z "$GRAPHIFY_PYTHON" ]; then
    if command -v python3 >/dev/null 2>&1 && python3 -c "import graphify" 2>/dev/null; then
        GRAPHIFY_PYTHON="python3"
    elif command -v python >/dev/null 2>&1 && python -c "import graphify" 2>/dev/null; then
        GRAPHIFY_PYTHON="python"
    else
        # No usable Python interpreter — hook is a no-op on this machine
        exit 0
    fi
fi
```

I've been running this patched version locally on a Windows wallnut project since 2026-04-12 without issues. Happy to submit a PR if it's useful.

## Environment

- OS: Windows 11 (build via git-bash shell from Git for Windows)
[...]

PR #246, opened 2026-04-12
fix: use cross-platform Python fallback chain in git hooks

saheersk:
#244 

  On Windows, graphify is installed as a .exe wrapper with no shebang,
  so the interpreter-sniffing logic always fell through to the hardcoded
  GRAPHIFY_PYTHON="python3" fallback. Windows Python installs expose
  python (not python3), so the hook silently exited with
  "python3: command not found" and never rebuilt the graph.

  Replace the three scattered python3 hardcodes with a single unified
  fallback block that probes python3 then python (each verified via
  `import graphify`), and exits 0 silently if neither is usable.


safishamsi:
Superseded — we shipped the same Python fallback chain (`python3` → `python` → exit 0) in v0.4.3. Thank you!

PR #268, opened 2026-04-12
feat: enhance Antigravity IDE rules with native MCP hints and skill metadata

Gennady87:
# Title
`feat: enhance Antigravity IDE rules with native MCP hints and skill metadata`

# Description

Hi Safi! Awesome work merging the Google Antigravity platform support in `0.4.6`. 

I am an Antigravity IDE user and noticed three minor adjustments that will make the integration feel completely native and autonomous within the IDE ecosystem. 

This PR builds on top of your implementation by hooking directly into Antigravity's tooling pipeline:

### 1. Injected YAML frontmatter into `SKILL.md`
Antigravity strictly requires YAML frontmatter (`name` and `description`) to register local files as agentic tools. I added a quick prepend step during installation so the IDE natively recognizes `graphify-manager` and knows when to trigger the manual fallback CLI checks.

### 2. Enhanced MCP-driven Agent Rules
The generated `.agent/rules/graphify.md` relied on classical file reading (which is perfect for Cursor). However, since Antigravity pairs heavily with your MCP server, I added an explicit `Rules` bullet pointing the agent to utilize `query_graph`, `get_node`, and `shortest_path` first. This forces the agent to traverse the graph logically instead of falling back to slow `grep` operations, drastically saving context tokens.

### 3. Print MCP Setup Instructions
Added a terminal print block that outputs the exact schema for `mcp_config.json` using `uv run ...`. This makes the "always-on" architecture step effortless for the user to configure.

Your `/graphify` workflow command remains untouched and works beautifully! Thanks for building such an incredible tool.

### DIFF
--- a/graphify/__main__.py
+++ b/graphify/__main__.py
@@ -314,6 +314,7 @@ This project has a graphify knowledge graph at graphify-out/.
 Rules:
 - Before answering architecture or codebase questions, read graphify-out/GRAPH_REPORT.md for god nodes and community structure
 - If graphify-out/wiki/index.md exists, navigate it instead of reading raw files
+- If the graphify MCP server is active, utilize tools like `query_graph`, `get_node`, and `shortest_path` for precise architecture navigation instead of falling back to `grep`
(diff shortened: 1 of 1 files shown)


safishamsi:
Fixed in v0.4.9.

PR #270, opened 2026-04-12
fix(watch): read 'links' key from graph.json instead of edges (#267)

### LINKED ISSUE #269: sem_edges reads wrong key from graph.json
## Problem
The edge merge logic in `watch.py` `_rebuild_code()` reads from `existing.get("edges", [])` but `graph.json` uses the key `"links"` (NetworkX node-link format). This means `sem_edges` is always an empty list and all semantic edges are silently dropped on every watcher rebuild.
This is the root cause behind #253 and #261. The fixes in 0.4.3, 0.4.4, and 0.4.5 added correct merge logic but all read from the wrong key, so the edge preservation never actually executes.
Node preservation works (reads `"nodes"`, correct key). Hyperedge preservation works (reads `"hyperedges"`, correct key). Only edge preservation is broken.
## Fix
One line in `watch.py`, line 46:
```python
# Before (broken)
sem_edges = [e for e in existing.get("edges", [])
# After (working)
sem_edges = [e for e in existing.get("links", existing.get("edges", []))
```
Falls back to `"edges"` for safety in case a future version changes the key.
## Verification
Tested locally by patching the installed 0.4.5 `watch.py`:
| Metric | Baseline | Unpatched watcher | Patched watcher |
|---|---|---|---|
| Total nodes | 587 | 588 | 588 |
| Total links | 534 | 396 | 531 |
| INFERRED | 49 | 13 | 47 |
| AMBIGUOUS | 2 | 0 | 2 |
| Edges touching docs | 138 | 3 | 138 |
| Doc INFERRED | 35 | 1 | 35 |
| Hyperedges | 22 | 22 | 22 |
Patched watcher preserves nearly all semantic edges. The small delta (534 vs 531 links, 49 vs 47 INFERRED) is from edge deduplication during rebuild, not data loss.

Joseph94m:
Closes #269

### DIFF
--- a/graphify/watch.py
+++ b/graphify/watch.py
@@ -43,7 +43,7 @@ def _rebuild_code(watch_path: Path, *, follow_symlinks: bool = False) -> bool:
                 existing = json.loads(existing_graph.read_text(encoding="utf-8"))
                 code_ids = {n["id"] for n in existing.get("nodes", []) if n.get("file_type") == "code"}
                 sem_nodes = [n for n in existing.get("nodes", []) if n.get("file_type") != "code"]
-                sem_edges = [e for e in existing.get("edges", [])
+                sem_edges = [e for e in existing.get("links", existing.get("edges", []))
                              if e.get("confidence") in ("INFERRED", "AMBIGUOUS")
                              or (e.get("source") not in code_ids and e.get("target") not in code_ids)]
                 result = {


Joseph94m:
@safishamsi it seems this was fixed in another PR. Closing PR

Issue #272, opened 2026-04-12
`graphify/SKILL.md` contains Claude-specific instructions and stale agent-tool naming (`wait` instead of `wait_agent` in one place).


3esmit:
References to "Claude" in all skill files, for example, in case of skill-aider.md:

https://github.com/safishamsi/graphify/blob/c713cf89fd042f6dd192795598d7ed8001686d6a/graphify/skill-aider.md?plain=1#L41
https://github.com/safishamsi/graphify/blob/c713cf89fd042f6dd192795598d7ed8001686d6a/graphify/skill-aider.md?plain=1#L160
https://github.com/safishamsi/graphify/blob/c713cf89fd042f6dd192795598d7ed8001686d6a/graphify/skill-aider.md?plain=1#L573
https://github.com/safishamsi/graphify/blob/c713cf89fd042f6dd192795598d7ed8001686d6a/graphify/skill-aider.md?plain=1#L575
https://github.com/safishamsi/graphify/blob/c713cf89fd042f6dd192795598d7ed8001686d6a/graphify/skill-aider.md?plain=1#L1120
https://github.com/safishamsi/graphify/blob/c713cf89fd042f6dd192795598d7ed8001686d6a/graphify/skill-aider.md?plain=1#L1162
https://github.com/safishamsi/graphify/blob/c713cf89fd042f6dd192795598d7ed8001686d6a/graphify/skill-aider.md?plain=1#L1167
https://github.com/safishamsi/graphify/blob/c713cf89fd042f6dd192795598d7ed8001686d6a/graphify/skill-aider.md?plain=1#L1173

This might confuse agents, and ideally the SKILL.md should be agnostic to specific brand of AI. 

All 3 flavors of skill-*.md contain 13 mentions to CLAUDE.

safishamsi:
Fixed in 0.4.8 — all platform skill files now use platform-agnostic wording.

3esmit:
The fix is partial. There still 11 references to Claude in the skill files.

safishamsi:
let me look into it tomorrow then

PR #275, opened 2026-04-12
Update skill-codex.md with correct Codex instructions.

### LINKED ISSUE #273: skill-codex.md: There is no `wait` tool in the available tool definitions, instead there is `wait_agent`.
The agent_tool.rs in codex defines the function as `wait_agent`, and there is no `wait` in the document. As the `wait` is surrounded by actual callable agent tools, I assume this `wait` is named incorrectly, and should be changed to `wait_agent`.

https://github.com/openai/codex/blob/a4d5112b37c1b012b761df66286e73b858205482/codex-rs/tools/src/agent_tool.rs#L205

This problem raised when I asked Codex to verify compatibility of the skill with Codex, where it pointed out the problem. However, I was not able to find any references in Codex official documentation, and this bug probably was caused by lack of documentation., 

This is the line:

https://github.com/safishamsi/graphify/blob/c713cf89fd042f6dd192795598d7ed8001686d6a/graphify/skill-codex.md?plain=1#L233

Codex was able to identify from the live Codex tool schema in this session, not from a public doc page.

3esmit:
This fix was auto-generated by Codex itself, and reviewed by me. 

The fix removes several mentions do Claude, and replace them by Codex equivalents. The agent tool funciton names are also fixed, and the MCP server configuration as well. 

This changes should make the skill fully compilant with Codex, make it more stable while working with subagents and MCP, and should save tokens by not having to reason about `Claude`.

Fixes partially #272 (only for Codex) and fixes #273

### DIFF
--- a/graphify/skill-codex.md
+++ b/graphify/skill-codex.md
@@ -1220,7 +1220,7 @@ Run once per project to make graphify always-on in Codex sessions:
 graphify codex install
 ```
 
-This writes a `## graphify` section to the local `AGETNTS.md` that instructs Codex to check the graph before answering codebase questions and rebuild it after code changes. No manual `/graphify` needed in future sessions.
+This writes a `## graphify` section to the local `AGENTS.md` that instructs Codex to check the graph before answering codebase questions and rebuild it after code changes. No manual `/graphify` needed in future sessions.
 
 ```bash
 graphify codex uninstall  # remove the section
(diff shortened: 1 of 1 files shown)


3esmit:
There still issues with the commands `/` vs `$` here, and the use of `.graphify_python` is causing problems with Codex.
The installer is using .agents for codex, which should be .codex. 
Also, the hooks are not available in Windows.

Tomorrow in my free time I will look more into Codex integration to fix all these issues.

safishamsi:
Implemented in v0.4.9 — `wait` → `wait_agent` in skill-codex.md. Upgrade: `pip install --upgrade graphifyy`

PR #282, opened 2026-04-12
Fix executable checks for hooks on Windows

### LINKED ISSUE #279: Windows hook tests assume POSIX executable bits
## Description
The hook installation tests currently assert POSIX executable mode bits on Windows-created files. In this environment, `graphify.hooks.install()` writes the hook files and calls `chmod(0o755)`, but the resulting `st_mode` still does not satisfy `0o111`, so the test suite fails even though the hook files exist and contain the expected content.

## Impact
Windows contributors cannot get a clean test run. The current failure masks whether hook installation is actually broken or whether the test is asserting a platform-specific filesystem detail that does not hold on Windows.

## Consequences
- CI or local validation will be noisy on Windows.
- Real regressions in hook content or install behavior can be obscured by a false-negative test.
- Cross-platform confidence in the hook subsystem is reduced.

## Affected By
- `graphify/hooks.py:120-130`
- `graphify/hooks.py:164-165`
- `tests/test_hooks.py:22-26`
- `tests/test_hooks.py:91-95`

## Evidence
- Repro command: `python -m pytest tests/ -q --tb=short`
- Observed failures:
  - `FAILED tests/test_hooks.py::test_install_is_executable`
  - `FAILED tests/test_hooks.py::test_install_post_checkout_is_executable`
- Observed assertion detail:
  - `AssertionError: assert (33206 & 73)`
- Summary: `2 failed, 423 passed in 2.86s`

## Recommended Fix
Make hook validation platform-aware. On Windows, assert that the hook file exists, contains the graphify marker, and is runnable by Git, instead of asserting POSIX execute bits. If executable semantics still matter for non-Windows platforms, keep the mode-bit assertion behind a platform guard.

3esmit:
## Summary

Fix #279 the Windows hook tests so they validate a Windows-meaningful invariant instead of relying on POSIX executable bits.

## What changed

- Keep the existing executable-bit assertions on non-Windows platforms
- On Windows, assert that installed git hooks start with `#!/bin/sh`
- Leave production hook behavior unchanged

## Why

`st_mode & 0o111` is not reliable on Windows filesystems, which caused the hook tests to fail even though the generated hooks were valid for Git for Windows.

This keeps the test intent intact:
- POSIX still verifies executable permissions
- Windows verifies the hook script shape Git for Windows expects

## Testing

- `python -m pytest tests\test_hooks.py -q --tb=short`
- `python -m pytest tests\ -q --tb=short`

### DIFF
--- a/tests/test_hooks.py
+++ b/tests/test_hooks.py
@@ -1,9 +1,9 @@
 """Tests for hooks.py - git hook install/uninstall."""
+import os
 import subprocess
 from pathlib import Path
 import pytest
 from graphify.hooks import install, uninstall, status, _HOOK_MARKER, _CHECKOUT_MARKER
-
 
 def _make_git_repo(tmp_path: Path) -> Path:
     subprocess.run(["git", "init", str(tmp_path)], check=True, capture_output=True)
@@ -23,7 +23,10 @@ def test_install_is_executable(tmp_path):
     repo = _make_git_repo(tmp_path)
     install(repo)
     hook = repo / ".git" / "hooks" / "post-commit"
-    assert hook.stat().st_mode & 0o111  # executable bit set
+    if os.name == "nt":
+        assert hook.read_text(encoding="utf-8").startswith("#!/bin/sh\n")
+    else:
+        assert hook.stat().st_mode & 0o111  # executable bit set
 
 
 def test_install_idempotent(tmp_path):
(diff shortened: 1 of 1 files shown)


safishamsi:
Implemented in v0.4.9 — hook tests now use platform-aware assertions (`os.name == 'nt'` guard) instead of POSIX executable bit checks. Upgrade: `pip install --upgrade graphifyy`

PR #291, opened 2026-04-13
fix: diacritic-insensitive query scoring and relevance-based output ordering

pieralukasz:
## Problem

Querying `graphify query "cwiczenia z lukami"` fails to match a node labeled **"Ćwiczenia z lukami – dlaczego są do bani"** because Python's `in` operator does exact byte comparison — `c` (U+0063) ≠ `ć` (U+0107).

Additionally, matched nodes appear at the bottom of results because output is sorted by node degree (connection count) rather than query relevance. A perfectly matching node with few connections gets buried behind high-degree but irrelevant neighbors.

This affects all languages with diacritics: Polish (ć, ź, ó, ą, ę), French (é, è, ê), Czech (ř, ž, č), German (ä, ö, ü), etc.

## Fix

**1. Diacritic-insensitive matching** — `_strip_diacritics()` using `unicodedata.normalize("NFKD")` strips combining characters before comparison. Both the query and node labels are normalized, so `cwiczenia` matches `ćwiczenia`. Zero new dependencies (unicodedata is stdlib).

**2. Relevance-based output ordering** — `_subgraph_to_text()` now accepts optional `node_scores` and sorts by (score, degree) instead of degree alone. Direct query matches always appear before traversal neighbors.

**3. Consistent diacritic support** — Applied to `_score_nodes`, `_find_node`, and refactored `_tool_get_node` to use `_find_node` instead of duplicating inline matching logic.

## Backward compatibility

- `node_scores` defaults to `None` → falls back to degree-based ordering
- Existing exact matches still work (diacritic normalization is additive)
- No new dependencies

## Tests

10 new tests in `test_serve.py` covering:
- `_strip_diacritics` with Polish, French, and ASCII inputs
- `_score_nodes` with ASCII query vs diacritic labels
- `_find_node` diacritic-insensitive lookup
- `_subgraph_to_text` relevance ordering and fallback behavior

### DIFF
--- a/graphify/__main__.py
+++ b/graphify/__main__.py
@@ -926,9 +926,10 @@ def main() -> None:
         if not scored:
             print("No matching nodes found.")
             sys.exit(0)
+        score_map = {nid: s for s, nid in scored}
         start = [nid for _, nid in scored[:5]]
         nodes, edges = (_dfs if use_dfs else _bfs)(G, start, depth=2)
-        print(_subgraph_to_text(G, nodes, edges, token_budget=budget))
+        print(_subgraph_to_text(G, nodes, edges, token_budget=budget, node_scores=score_map))
     elif cmd == "save-result":
         # graphify save-result --question Q --answer A --type T [--nodes N1 N2 ...]
         import argparse as _ap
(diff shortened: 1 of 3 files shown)


pieralukasz:
Superseded by #293 which moves normalization to build time instead of runtime.

PR #292, opened 2026-04-13
feat: add Dart/Flutter language support

XXX1694:
## Summary

Adds Dart/Flutter language support to graphify's AST extraction pipeline.

## What's extracted

- Classes, abstract classes, mixins, enums, extensions, typedefs
- Flutter-specific patterns: StatelessWidget, StatefulWidget, State<T>
- Relationships: `inherits`, `implements`, `uses` (mixins), `defines`, `has_method`
- Package imports (`package:flutter/...`) and Dart builtins (`dart:async`, etc.)
- Method calls between known functions in the same file

## Why regex, not tree-sitter

`tree-sitter-dart` does not exist on PyPI with the new Language API (>=0.23)
that graphify requires. The regex-based approach covers all Flutter-specific
patterns without adding external dependencies.

## Changes

- `graphify/extract.py` — added `extract_dart()` + registered `.dart` in dispatch table and `collect_files()`
- `graphify/detect.py` — added `.dart` to `CODE_EXTENSIONS`
- `tests/fixtures/sample.dart` — realistic Flutter fixture with widgets, mixins, enums
- `tests/test_languages.py` — 15 tests for Dart extraction
- `README.md` — updated language count (22 → 23) and extensions table

## Tests

15/15 passing

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,7 +8,7 @@
 [![Sponsor](https://img.shields.io/badge/sponsor-safishamsi-ea4aaa?logo=github-sponsors)](https://github.com/sponsors/safishamsi)
 [![LinkedIn](https://img.shields.io/badge/LinkedIn-Safi%20Shamsi-0077B5?logo=linkedin)](https://www.linkedin.com/in/safi-shamsi)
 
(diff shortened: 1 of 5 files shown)


safishamsi:
Fixed in v0.4.9.

PR #293, opened 2026-04-13
fix: add norm_label at build time for diacritic-insensitive search

pieralukasz:
## Problem

Queries without diacritics (e.g. `cwiczenia z lukami`) fail to match nodes with diacritics in their labels (e.g. `Ćwiczenia z lukami`). Python's `in` operator does exact byte comparison — `c` (U+0063) ≠ `ć` (U+0107). This affects all languages with diacritics: Polish, French, Czech, German, etc.

Additionally, output nodes are sorted by degree (connection count) rather than query relevance, so even when a node matches it can appear buried behind high-degree but irrelevant neighbors.

## Approach

**Build time (`export.py`):** `to_json()` now generates a `norm_label` field for every node — lowercase, diacritics stripped via `unicodedata.normalize("NFKD")`. This sits alongside the existing `community` field that's already added at export time.

**Query time (`serve.py`):** `_score_nodes()` and `_find_node()` match against `norm_label` instead of normalizing on every query. A runtime fallback (`_strip_diacritics`) handles old graphs that lack the field.

**Output ordering:** `_subgraph_to_text()` accepts optional `node_scores` and sorts by (relevance, degree) so direct matches appear before high-degree traversal neighbors.

**Consistency:** `_tool_get_node` refactored to use `_find_node` instead of duplicating inline matching.

## Backward compatibility

- Old graphs without `norm_label` still work — `_score_nodes` and `_find_node` fall back to runtime normalization
- `node_scores` defaults to `None` → preserves degree-based ordering
- No new dependencies (`unicodedata` is stdlib)

## Tests

15 new tests across `test_serve.py` and `test_export.py`:
- `_strip_diacritics` with Polish, French, ASCII
- `_score_nodes` / `_find_node` with and without `norm_label` (fallback path)
- `to_json` produces `norm_label` on every node
- `_subgraph_to_text` relevance ordering

### DIFF
--- a/graphify/__main__.py
+++ b/graphify/__main__.py
@@ -926,9 +926,10 @@ def main() -> None:
         if not scored:
             print("No matching nodes found.")
             sys.exit(0)
+        score_map = {nid: s for s, nid in scored}
         start = [nid for _, nid in scored[:5]]
         nodes, edges = (_dfs if use_dfs else _bfs)(G, start, depth=2)
-        print(_subgraph_to_text(G, nodes, edges, token_budget=budget))
+        print(_subgraph_to_text(G, nodes, edges, token_budget=budget, node_scores=score_map))
     elif cmd == "save-result":
         # graphify save-result --question Q --answer A --type T [--nodes N1 N2 ...]
         import argparse as _ap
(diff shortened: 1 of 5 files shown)


safishamsi:
Fixed in v0.4.9.

pieralukasz:
Thanks for including the diacritics fix in v0.4.9! One piece from #293 that didn't make it in is the output ordering — `_subgraph_to_text` still sorts by node degree, so direct query matches can get buried behind high-degree traversal neighbours. Opened #308, please review it.

Issue #298, opened 2026-04-13
AST extraction: cross-file call resolution is limited to Python imports only


cyxer000:
## Problem

`walk_calls` in `extract.py` resolves call targets using a **file-local** `label_to_nid` dict (L948-949):

```python
tgt_nid = label_to_nid.get(callee_name.lower())
```

This means any call to a function/class defined in another file is silently dropped. Cross-file relationships — which are theoretically fully resolvable from AST — are almost entirely lost.

### Current state

The only cross-file mechanism is `_resolve_cross_file_imports()` (L2039), which has significant limitations:

| Limitation | Detail |
|---|---|
| **Python-only** | L2652: `# Python only - uses Python parser internally` |
| **Import-level only** | Only resolves `from .module import Name`, does not trace actual call sites |
| **Coarse granularity** | Connects *all* classes in the importing file to the imported entity, regardless of whether they actually use it |
| **Stem-only module matching** | Matches by filename stem (L2083), ignores package paths like `foo.bar.baz` |

All other languages (JS/TS, Go, Rust, Java, C/C++, etc.) have **zero** cross-file resolution — calls to imported symbols produce no edges at all.

## Impact

For any multi-file codebase, the AST call graph is effectively incomplete. The semantic subagents (LLM) partially compensate, but:
- They are non-deterministic and token-expensive
- They miss structural relationships that AST could extract for free
- Code-only corpora skip semantic extraction entirely, so cross-file calls are completely lost

## Proposed approach

**Build a global symbol table in a two-pass architecture:**

1. **Pass 1 (existing):** Per-file extraction — collect all nodes (`label_to_nid`) and raw call names
2. **Pass 2 (new):** Merge all per-file symbol tables into a global `label_to_nid`, then re-resolve unresolved calls against the global table

This is the minimal change — `walk_calls` already extracts the callee name, it just fails to find a target because the lookup scope is too narrow. A global lookup table would immediately resolve cross-file calls for all languages without language-specific import handlers.

### Additional improvements (optional, can be separate PRs)

- Per-language import resolvers to map `import X from './foo'` to the actual file's symbols
- Package-path-aware module matching (not just stem)
- Narrowing `_resolve_cross_file_imports` granularity to only connect entities that actually reference the import

## Environment

- graphify version: installed via `pip install graphifyy`
- Python: 3.14
- OS: Windows 11

safishamsi:
[...]

PR #315, opened 2026-04-13
 fix: resolve cross-file call edges via global symbol table 

saheersk:
(all languages) #298 

   walk_calls in each per-file extractor only looked up callee names in a
   file-local label_to_nid dict, so any call to a symbol defined in another
   file was silently dropped. For multi-file codebases this left most call
   relationships missing from the graph.

   Two-pass fix:
   - Each extractor (extract_generic, Go, Rust, Zig, PowerShell, Elixir) now records unresolved calls — where a callee name was extracted but not found locally — in an unresolved_calls list returned with the result.
   - extract() builds a global label_to_nid from all nodes across all files after per-file extraction completes, then resolves the unresolved calls against it. Resolved edges are added as calls with confidence INFERRED and weight 0.9 to distinguish them from same-file EXTRACTED edges.

   Stale cache entries (written before this change, lacking unresolved_calls)
   are detected and transparently re-extracted so the global pass gets full data.

   Fixes cross-file call resolution for all 20+ supported languages without
   any language-specific import logic. Import-aware resolution remains a
   follow-up improvement for reducing false positives on common names.


saheersk:
Problem

  The global symbol table added in #[prev-pr] resolves cross-file calls for all languages, but it has a known weakness: name collisions. If two files define a function with the same name (e.g. process, init, run), the global table maps that name to whichever node was
  inserted first. A call in a third file may resolve to the wrong target.

  Proposed fix

  Use import statements to constrain which file a callee can come from, before falling back to the global table.

  Two-step resolution:

  1. Import map pass — before resolving calls, parse each file's import statements and build a per-file map: {imported_name → source_file}. For example:
    - Python: from utils import helper → helper comes from utils.py
    - JS/TS: import { helper } from './utils' → helper comes from utils.ts
    - Go: import "pkg/utils" → utils.Method comes from that package
    - Java: import com.example.Utils → Utils comes from that class file
  2. Constrained lookup — when resolving an unresolved call, first check if the callee name appears in the file's import map. If yes, only match nodes whose source_file matches the imported file. Fall back to the global table only if no import-constrained match exists.

  Impact

  - Eliminates false positives for common function names
  - Makes INFERRED edges nearly as reliable as EXTRACTED ones for well-structured codebases
  - No changes to AST parsing — works by post-processing the existing imports/imports_from edges already in the graph
  -

PR #367, opened 2026-04-15
Harden SSRF validation, fix path traversal in exports, prevent info disclosure

brodsky754:
## Summary

Security hardening across 5 files (39 insertions, 10 deletions), found via automated audit. All 433 existing tests pass unchanged.

- **SSRF validation (security.py)**: DNS resolution failures now raise `ValueError` (fail-closed) instead of silently passing. Expanded cloud metadata blocklist to cover AWS (`instance-data.ec2.internal`) and Azure (`metadata.azure.com`). IPv4-mapped IPv6 addresses (`::ffff:x.x.x.x`) are now normalized before private IP range checks.
- **Path traversal (wiki.py, export.py)**: Wiki and Obsidian vault exports now strip `..` from generated filenames, truncate to 200 chars, and enforce `resolve().relative_to()` containment — crafted node labels can no longer write files outside the output directory.
- **Info disclosure (serve.py)**: MCP tool error handler now returns a generic message instead of leaking `str(exc)` (which could expose file paths, library versions, or internal state). Full errors are logged to stderr.
- **Weak hash (transcribe.py)**: Replaced `hashlib.sha1` with `hashlib.sha256` for URL-based audio cache keys, consistent with `cache.py` which already uses SHA-256.

## Test plan

- [x] All 433 tests pass (`pytest tests/ -v` — 0 failures, 0 new failures)
- [x] Verified via independent review agent: all fixes confirmed in place
- [ ] Manual: confirm `graphify add <url>` with an unresolvable hostname now raises instead of passing through
- [ ] Manual: confirm Obsidian/wiki export with adversarial node labels stays contained

### DIFF
--- a/graphify/export.py
+++ b/graphify/export.py
@@ -479,11 +479,19 @@ def to_obsidian(
 
     # Map node_id → safe filename so wikilinks stay consistent.
     # Deduplicate: if two nodes produce the same filename, append a numeric suffix.
+    resolved_out = out.resolve()
+
     def safe_name(label: str) -> str:
         cleaned = re.sub(r'[\\/*?:"<>|#^[\]]', "", label.replace("\r\n", " ").replace("\r", " ").replace("\n", " ")).strip()
         # Strip trailing .md/.mdx/.markdown so "CLAUDE.md" doesn't become "CLAUDE.md.md"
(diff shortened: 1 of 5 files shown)


brodsky754:
Opened against wrong repo — intended for downstream project.

PR #370, opened 2026-04-15
perf: use igraph C backend for betweenness centrality (~100x speedup)

s-a-c:
## Problem

`nx.betweenness_centrality(G)` in `analyze.py` uses NetworkX's pure-Python O(V×E) implementation, which becomes a hard bottleneck on large codebases. On a 33K-file corpus (167K nodes, 202K edges), the `graphify update` command ran for **1 hour 43 minutes** and never completed — it was stuck in the BFS inner loop of betweenness centrality.

A second call site, `nx.edge_betweenness_centrality(G)` in `_cross_community_surprises()`, has the same issue.

## Solution

Replace both bare NetworkX calls with tiered helper functions:

1. **igraph (C backend)** — preferred path. Exact computation, ~30-100x faster than pure-Python NX. Requires `python-igraph` as an optional dependency.
2. **NetworkX sampled approximation** — fallback if igraph is unavailable. Uses `k=500` sampled source nodes for graphs >5K nodes (O(k×E) instead of O(V×E)).
3. **NetworkX exact** — for small graphs (<5K nodes) where it's fast enough.

The helpers (`_fast_betweenness_centrality`, `_fast_edge_betweenness_centrality`) are drop-in replacements — same return types, same normalization.

## Benchmark Results

| Graph Size | NetworkX (pure Python) | igraph (C) | Speedup |
|---|---|---|---|
| 5K nodes / 15K edges | 35.3s | 1.15s | ~30x |
| 20K nodes / 60K edges | est. >10min | 20.3s | ~30-50x |

## Real-world validation

| Corpus | Files | Nodes | Edges | Before | After |
|---|---|---|---|---|---|
| fam-a-lam | 33,004 | 167,161 | 201,912 | >1h43m (interrupted) | **30s** |
| lsk-13-teams | 23,755 | 287,186 | 306,318 | N/A | **1m 22s** |

## Changes

- Added `_nx_to_igraph()` — NX→igraph graph conversion utility
- Added `_fast_betweenness_centrality()` — tiered node betweenness
- Added `_fast_edge_betweenness_centrality()` — tiered edge betweenness
- Replaced call in `suggest_questions()` (line 363)
- Replaced call in `_cross_community_surprises()` (line 265)

## Notes

- `python-igraph` is an optional dependency — if not installed, the helpers gracefully degrade to the NetworkX fallback path
- No changes to public API or return value contracts
- Normalization between igraph and NetworkX output is handled in the helpers

---
[Warp conversation](https://app.warp.dev/conversation/78a656e2-6c3e-4270-9434-9afcb102e8af)

Co-Authored-By: Oz <oz-agent@warp.dev>


s-a-c:
Closing to recreate with a clean branch based on current main.

PR #372, opened 2026-04-15
fix: recognize .mjs and .ejs files in AST extractor

### LINKED ISSUE #365: .mjs and .ejs files not recognized by AST extractor
My entire codebase uses .mjs (ES modules with import/export) and .ejs (Express templates). Graphify skips all of them silently during graph building, reporting only 3 code files detected out of 75+ .mjs files.
.mjs is standard modern JavaScript — same language as .js, just with explicit ES module syntax. It should be treated identically to .js by the tree-sitter AST extractor.
.ejs is a widely used Node.js templating engine and should at minimum be processed as a doc via semantic extraction.
The result for my stack is that Graphify builds a graph of images and README files only, with zero architectural insight from the actual business logic layer.
Request: Add .mjs to the tree-sitter JS/TS extension list, and add .ejs to the semantic extraction doc pipeline.

jakeefr:
Fixes #365. Added `.mjs` and `.ejs` file extensions to the AST extractor so modern ES module files and EJS templates are properly parsed.

- `.mjs` added to `CODE_EXTENSIONS`, `_DISPATCH` (mapped to `extract_js`), and `collect_files` `_EXTENSIONS` — treated identically to `.js`
- `.ejs` added to `DOC_EXTENSIONS` for semantic extraction processing

I maintain [PRISM](https://github.com/jakeefr/prism), a post-session diagnostics tool for Claude Code — session health scoring and CLAUDE.md adherence analysis. Both tools help developers get deeper signal from their codebases and sessions.

### DIFF
--- a/graphify/detect.py
+++ b/graphify/detect.py
@@ -18,8 +18,8 @@ class FileType(str, Enum):
 
 _MANIFEST_PATH = "graphify-out/manifest.json"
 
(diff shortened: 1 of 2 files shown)


safishamsi:
Implemented in v0.4.15. .mjs and .ejs support added to the AST extractor. Thanks for the fix!

PR #373, opened 2026-04-15
fix: use polling observer by default on macOS watch

samqin123:
## Summary\n- use  by default on macOS watch mode for better stability\n- allow override via \n- document the behavior in README + README.zh-CN\n- add tests for observer mode parsing\n\n## Why\nOn macOS, the native watchdog observer can fail to start reliably in some environments with errors like . This patch makes watch mode more stable out of the box on macOS while still letting advanced users opt back into the native observer.\n\n## Testing\n- ............                                                             [100%]
12 passed in 0.01s


safishamsi:
Implemented in v0.4.15. macOS watch now uses PollingObserver by default. Thanks!

PR #375, opened 2026-04-15
fix: rename god_nodes 'edges' field to 'degree'

yangmeilly:
god_nodes() returned a dict with key 'edges' for the degree count, but all consumers (report, serve, wiki, skill scripts) referenced it as 'degree', causing KeyError at runtime.

detail info 
```
$ $(cat graphify-out/.graphify_python) -c "
import json
from pathlib import Path

analysis = json.loads(Path('graphify-out/.graphify_analysis.json').read_text())

# 只显示 GOD 节点和问题
print('=== GOD NODES ===')
for node in analysis['gods'][:10]:  # 前10个
    print(f\"  {node['label']} (degree={node['degree']})\")

print('\n=== SUGGESTED QUESTIONS ===')
for q in analysis['questions'][:5]:  # 前5个
    print(f\"  {q}\")

print('\n=== COMMUNITY COUNT ===')
print(f\"  Total communities: {len(analysis['communities'])}\")

# 查看前几个社区的规模
print('\n=== TOP COMMUNITIES BY SIZE ===')
sorted_comms = sorted(analysis['communities'].items(), key=lambda x: len(x[1]), reverse=True)[:10]
for cid, nodes in sorted_comms:
    print(f\"  Community {cid}: {len(nodes)} nodes\")
"
=== GOD NODES ===
  File "<string>", line 10, in <module>
KeyError: 'degree'
```
fix policy
 ```
 god_nodes() in analyze.py:51-55 returns nodes with the field named "edges", not "degree":

  result.append({
      "id": node_id,
      "label": G.nodes[node_id].get("label", node_id),
      "edges": deg,   # <-- field is "edges", not "degree"
  })

  The "degree" field at export.py:385 belongs to the vis.js node payload and is unrelated to the gods array written to .graphify_analysis.json.

  Fix options:
  1. Rename "edges" → "degree" in analyze.py:54 (preferred — degree is the standard graph theory term)
  2. Update consuming code to use node['edges'] instead of node['degree']
```

### DIFF
--- a/graphify/analyze.py
+++ b/graphify/analyze.py
@@ -51,7 +51,7 @@ def god_nodes(G: nx.Graph, top_n: int = 10) -> list[dict]:
         result.append({
             "id": node_id,
             "label": G.nodes[node_id].get("label", node_id),
-            "edges": deg,
+            "degree": deg,
         })
         if len(result) >= top_n:
             break
--- a/graphify/report.py
+++ b/graphify/report.py
@@ -72,7 +72,7 @@ def generate(
         "## God Nodes (most connected - your core abstractions)",
     ]
     for i, node in enumerate(god_node_list, 1):
-        lines.append(f"{i}. `{node['label']}` - {node['edges']} edges")
+        lines.append(f"{i}. `{node['label']}` - {node['degree']} edges")
 
     lines += ["", "## Surprising Connections (you probably didn't know these)"]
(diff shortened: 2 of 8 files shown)


safishamsi:
Implemented in v0.4.15. god_nodes() field renamed from edges to degree across all production code and tests. Thanks!

PR #382, opened 2026-04-15
Emit wiki/ on code-only rebuild so report wikilinks resolve

prkash1704:
## Summary

- `watch._rebuild_code` and the `cluster-only` CLI path regenerate `GRAPH_REPORT.md` + `graph.json` but never call `to_wiki()`.
- Consequence: the `[[_COMMUNITY_<name>]]` and `[[<god-node>]]` wikilinks the report emits dead-end until the user runs a separate wiki export. The AGENTS/CLAUDE.md guidance to *"navigate wiki/index.md instead of raw files"* resolves to nothing on incremental rebuilds — agents fall back to reading raw source, exactly what the graph is meant to avoid.
- This PR wires `to_wiki()` into both rebuild paths so `graphify-out/wiki/` stays in sync with the report on every tick.

`to_wiki()` already produces rich articles (key concepts with degree, cross-community relationships, source files, audit trail, god-node neighbors grouped by relation) — no changes needed there, just call it.

## Test plan

- [x] Ran `_rebuild_code(Path('.'))` against a real project (141 communities, 1791 nodes, 3321 edges) — `graphify-out/wiki/` populated with community + god-node articles + index.md, all non-empty
- [x] Verified `GRAPH_REPORT.md` wikilinks now resolve to populated files
- [ ] `cluster-only` path (not exercised end-to-end, but mirrors the same call shape)


prkash1704:
Follow-up commit on this branch: fixed a related dead-link issue in god-node articles. `_god_node_article` was emitting `[[<neighbor_label>]]` for every connection, but `to_wiki` only writes articles for community hubs + god nodes, so non-god neighbors had no landing page.

Now the god-node connection list only wikilinks neighbors that have their own article; everyone else gets `**<label>** (in [[<community>]])` so the reader still lands somewhere real.

Verified on a 1791-node / 141-community project: broken-link ratio dropped from **43.8% (272/621) → 0% (0/484)**. Happy to split into a separate PR if you'd prefer to review them independently.

PR #385, opened 2026-04-16
fix: escape colon in SKILL.md description for Kiro compatibility

masterArwel:
## Problem
  The \`description\` field in SKILL.md contains unquoted colons (\`:\`), which is a YAML special character. This causes YAML parsing to fail in Kiro, making the skill
  unrecognizable.

  ## Fix
  Wrap the description value in double quotes and replace the problematic colon with a dash.

  ## Before
  \`\`\`yaml
  description: ...three outputs: interactive HTML...
  \`\`\`

  ## After
  \`\`\`yaml
  description: \"...three outputs - interactive HTML...\"
  \`\`\`


safishamsi:
Fixed in v0.4.19 — `description` value in the Kiro skill YAML frontmatter is now quoted and colons replaced with dashes, preventing the parse error in Kiro's YAML loader.

Issue #387, opened 2026-04-16
.mjs files detected but produce 0 nodes during extraction (incomplete fix for #365)


shaun0927:
## Summary

Issue #365 was closed as fixed in v0.4.15, but the fix is incomplete. `.mjs` files were added to `CODE_EXTENSIONS` in `detect.py` so they appear in corpus metrics, but were not added to the `_DISPATCH` table in `extract.py`. The extractor silently skips them, producing zero nodes and edges.

Users who upgraded to v0.4.15 expecting #365 to be resolved will see `.mjs` files counted in the detection summary but get no architectural insight from them.

## Reproduction

```python
from pathlib import Path
from graphify.detect import CODE_EXTENSIONS
from graphify.extract import extract
import tempfile

tmpdir = Path(tempfile.mkdtemp())
mjs = tmpdir / "app.mjs"
mjs.write_text('export function hello() { return "world"; }')

print(".mjs in CODE_EXTENSIONS:", ".mjs" in CODE_EXTENSIONS)  # True

result = extract([mjs], cache_root=tmpdir)
print("Nodes:", len(result["nodes"]))  # 0 — should be >= 2
print("Edges:", len(result["edges"]))  # 0
```

**Output:**
```
.mjs in CODE_EXTENSIONS: True
Nodes: 0
Edges: 0
```

## Root cause

`detect.py:21` includes `.mjs` in `CODE_EXTENSIONS`:
```python
CODE_EXTENSIONS = {'.py', '.ts', '.js', ..., '.mjs', '.ejs', ...}
```

`extract.py:3062-3098` — the `_DISPATCH` table has no entry for `.mjs`:
```python
_DISPATCH = {
    ".py": extract_python,
    ".js": extract_js,
    # ... no .mjs entry
}
```

The extraction loop at line 3109 does `extractor = _DISPATCH.get(path.suffix)` which returns `None`, then `continue` at line 3111 silently skips the file.

PR #372 (which included the `_DISPATCH` addition) was closed without merge. The direct fix applied to v0.4.15 updated `detect.py` but missed `extract.py`.

## Fix

Add `".mjs": extract_js` to the `_DISPATCH` dict in `extract()`. `.mjs` is standard ES module JavaScript and should be parsed identically to `.js` by tree-sitter-javascript.

[...]

PR #389, opened 2026-04-16
fix: add .mjs to AST extractor dispatch table

shaun0927:
Partial fix for #387 (and #365).

`.mjs` was added to `CODE_EXTENSIONS` in `detect.py` (v0.4.15) but not to the `_DISPATCH` table in `extract()`, so `.mjs` files are detected and counted in corpus metrics but silently produce zero nodes/edges during extraction.

## Change

One line: `".mjs": extract_js` added to `_DISPATCH` in `graphify/extract.py`.

`.mjs` is standard ES module JavaScript — same language as `.js`, just with explicit module syntax. tree-sitter-javascript handles both identically.

## Testing

```python
from pathlib import Path
import tempfile

tmpdir = Path(tempfile.mkdtemp())
mjs = tmpdir / "app.mjs"
mjs.write_text("""
import { readFile } from "fs/promises";
export function parseConfig(path) {
    return readFile(path, "utf-8").then(JSON.parse);
}
export class AppServer {
    constructor(port) { this.port = port; }
    start() { console.log("started"); }
}
""")
result = extract([mjs], cache_root=tmpdir)
print(f"Nodes: {len(result['nodes'])}")  # 5 (was 0 before fix)
print(f"Edges: {len(result['edges'])}")  # 5 (was 0 before fix)
```

## Note on `.ejs`

`.ejs` has the same dispatch gap but EJS is an HTML templating engine, not pure JavaScript. It may be better suited to `DOC_EXTENSIONS` for semantic extraction. Left for separate discussion in #387.


shaun0927:
Closing to rebase onto main — the original branch was based on the v0.4.15 tag which diverges from main, producing a huge diff. Will reopen with a clean single-commit branch.

PR #392, opened 2026-04-16
fix: add missing sys import in watch.py

### LINKED ISSUE #386: graphify watch crashes with NameError on all platforms (v0.4.15 regression)
## Summary

`graphify watch <path>` crashes immediately with `NameError: name 'sys' is not defined` on every platform at v0.4.15. The macOS `PollingObserver` fix (#373) added a `sys.platform` check at `watch.py:150` but did not add `import sys` to the module imports.

## Reproduction

```bash
pip install graphifyy==0.4.15
python -c "from graphify.watch import watch; from pathlib import Path; watch(Path('.'))"
```

**Output:**
```
NameError: name 'sys' is not defined
```

Tested on macOS 15.4 / Python 3.12. The crash is platform-independent since `sys` is referenced before the platform check can execute.

## Root cause

`watch.py` imports (lines 2-8):
```python
from __future__ import annotations
import json
import time
from pathlib import Path
from graphify.detect import CODE_EXTENSIONS, ...
```

Line 150 references `sys.platform`:
```python
observer = PollingObserver() if sys.platform == "darwin" else Observer()
```

`sys` is never imported at module scope.

## Fix

Add `import sys` after line 3 in `graphify/watch.py`. One-line change, no side effects.

shaun0927:
Fixes #386.

`watch()` references `sys.platform` at line 150 to select `PollingObserver` on macOS, but `sys` was never imported. This causes `NameError` on all platforms when running `graphify watch`.

Introduced in v0.4.15 by #373 (macOS PollingObserver change).

## Change

One line: `import sys` added to module imports in `graphify/watch.py`.

## Testing

```bash
# Before fix:
python -c "from graphify.watch import watch; from pathlib import Path; watch(Path('.'))"
# NameError: name 'sys' is not defined

# After fix:
python -c "from graphify.watch import watch; from pathlib import Path; watch(Path('.'))"
# [graphify watch] Watching /path - press Ctrl+C to stop
```

### DIFF
--- a/graphify/watch.py
+++ b/graphify/watch.py
@@ -1,6 +1,7 @@
 # monitor a folder and auto-trigger --update when files change
 from __future__ import annotations
 import json
+import sys
 import time
 from pathlib import Path
 


shaun0927:
I re-checked the current upstream `v4` branch and the released `graphifyy==0.4.16` package.

This fix is already shipped upstream: `graphify/watch.py` now includes `import sys`, and the current release contains the same change as well.

Since the bug was real but the patch is already present in the maintained branch/release, I'm closing this PR as superseded.

PR #393, opened 2026-04-16
fix: add .mjs to AST extractor dispatch table

shaun0927:
Partial fix for #387 (and #365).

`.mjs` was added to `CODE_EXTENSIONS` in `detect.py` (v0.4.15) but not to the `_DISPATCH` table in `extract()`, so `.mjs` files are detected and counted in corpus metrics but silently produce zero nodes/edges during extraction.

## Change

One line: `".mjs": extract_js` added to `_DISPATCH` in `graphify/extract.py`.

`.mjs` is standard ES module JavaScript — same language as `.js`, just with explicit module syntax. tree-sitter-javascript handles both identically.

## Testing

```python
from pathlib import Path
import tempfile

tmpdir = Path(tempfile.mkdtemp())
mjs = tmpdir / "app.mjs"
mjs.write_text("""
import { readFile } from "fs/promises";
export function parseConfig(path) {
    return readFile(path, "utf-8").then(JSON.parse);
}
export class AppServer {
    constructor(port) { this.port = port; }
    start() { console.log("started"); }
}
""")
result = extract([mjs], cache_root=tmpdir)
print(f"Nodes: {len(result['nodes'])}")  # 5 (was 0 before fix)
print(f"Edges: {len(result['edges'])}")  # 5 (was 0 before fix)
```

## Note on .ejs

`.ejs` has the same dispatch gap but EJS is an HTML templating engine, not pure JavaScript. It may be better suited to `DOC_EXTENSIONS` for semantic extraction. Left for separate discussion in #387.

### DIFF
--- a/graphify/extract.py
+++ b/graphify/extract.py
@@ -3095,6 +3095,7 @@ def extract(paths: list[Path], cache_root: Path | None = None) -> dict:
         ".dart": extract_dart,
         ".v": extract_verilog,
         ".sv": extract_verilog,
+        ".mjs": extract_js,
     }
 
     total = len(paths)


vhsantos26:
Confirming impact on a real corpus: ran graphify on a Node.js/TypeScript project that uses `.mjs` extensions for all internal ESM imports (`'../../shared/logger.mjs'` style).

Current behavior (main):
- imports_ratio: 11.6% — and the project uses purely relative imports (no aliases)
- 45% of nodes are leaves, 7% isolated
- Expected ratio for a codebase this size should be 25%+

The gap is exactly what this PR addresses — `.mjs` is never mapped in the extension dispatch table, so the resolved target ID doesn't match the file node ID and every `.mjs` import edge is dropped.

One-line fix, clear correctness argument. Would be great to get this merged.

shaun0927:
Thanks @vhsantos26 for confirming the real-world impact.

I re-checked the current upstream `v4` branch and the released `graphifyy==0.4.16` wheel: both already contain the exact `.mjs` dispatch fix (`graphify/extract.py` now maps `".mjs"` to `extract_js`).

So the bug was real, but the maintained branch/release already includes the one-line fix. Closing this PR as superseded to avoid leaving a stale duplicate open.

PR #431, opened 2026-04-18
fix(extract): Go stdlib imports collide with same-named local files

whytiger3-sys:
Problem: Go file importing stdlib package (e.g. "context") produces imports_from edge pointing at a LOCAL file of the same basename, not a separate stdlib node.

Root cause in extract_go (extract.py:1949-1951):

    module_name = raw.split("/")[-1]   # "context" (stdlib)
    tgt_nid = _make_id(module_name)    # same id as context.go file_nid

Impact: In a Presto Go monorepo (~1000 files), this produced 115 false cycle-dependency pairs. Go compiler reported 0 cycles (go build clean); the graph showed 115. All were same-name file collisions.

Fix: Prefix Go import target id with go_pkg_ and use full import path:
- import "context" → go_pkg_context
- import "presto/internal/accounting" → go_pkg_presto_internal_accounting

Tests: Added tests/test_go_import_nodeid.py with 2 regression tests. All 20 existing extract tests still pass.

Happy to iterate on naming if you prefer a different convention.

(NOTE: This replaces my accidentally-filed PR #430 which I targeted at main instead of v4; apologies for the noise.)

### DIFF
--- a/graphify/extract.py
+++ b/graphify/extract.py
@@ -1947,15 +1947,20 @@ def extract_go(path: Path) -> dict:
                             path_node = spec.child_by_field_name("path")
                             if path_node:
                                 raw = _read_text(path_node, source).strip('"')
-                                module_name = raw.split("/")[-1]
-                                tgt_nid = _make_id(module_name)
+                                # Use the full import path as node id so stdlib
+                                # packages like "context" don't collide with
+                                # local files also named "context.go".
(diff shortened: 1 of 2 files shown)


safishamsi:
Solid find — 115 false cycle pairs is a real impact. Implemented the `go_pkg_` prefix fix from our end in v0.4.23 using the full import path as the node ID. Thanks for the clear root-cause analysis.

PR #432, opened 2026-04-18
fix(watch): rebuild continues when HTML viz exceeds MAX_NODES_FOR_VIZ

whytiger3-sys:
Problem: _rebuild_code calls to_html() unconditionally, which raises ValueError for graphs exceeding MAX_NODES_FOR_VIZ (5000). The entire rebuild fails and returns False, even though AST extraction, clustering, and report generation all succeeded. Users with medium-to-large codebases can't use _rebuild_code at all; no graph.json / GRAPH_REPORT.md is produced.

Real-world impact: In a Presto Go monorepo (~1000 files), _rebuild_code reliably produces 5574 nodes (above 5000). This made the git post-commit hook a no-op: every code change raised the HTML-viz error, graph.json never got written, AI tools read 9-day-old stale data.

Fix: Wrap to_html() in its own try/except. When it raises:
1. Log a warning (don't fail the rebuild)
2. Delete stale graph.html from prev smaller run
3. Continue returning True - core products (graph.json + GRAPH_REPORT.md) are already written

Tests: Added tests/test_watch_rebuild_large_graph.py with 2 tests. All existing tests pass.

Alternative considered: --no-viz parameter; rejected because it requires callers (post-commit hook, CI) to know the threshold. Caller-agnostic fix is cleaner.

### DIFF
--- a/graphify/watch.py
+++ b/graphify/watch.py
@@ -78,7 +78,22 @@ def _rebuild_code(watch_path: Path, *, follow_symlinks: bool = False) -> bool:
                           {"input": 0, "output": 0}, str(watch_path), suggested_questions=questions)
         (out / "GRAPH_REPORT.md").write_text(report, encoding="utf-8")
         to_json(G, communities, str(out / "graph.json"))
-        to_html(G, communities, str(out / "graph.html"), community_labels=labels or None)
+
+        # HTML viz is optional: it raises ValueError when the graph exceeds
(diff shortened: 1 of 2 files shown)


safishamsi:
Thanks for the detailed writeup and the real-world repro (Presto monorepo). Implemented the same fix from our end in v0.4.23 — `to_html` is now wrapped in its own try/except so `graph.json` and `GRAPH_REPORT.md` always land on large graphs.

PR #443, opened 2026-04-19
fix: skip structural-only communities in reports

RickwangF:
## Summary
- skip community sections that become empty after filtering structural/file-only nodes
- avoid rendering `Nodes (0):` entries in `GRAPH_REPORT.md`
- add a regression test for structural-only communities

## Test plan
- [x] Run `python3 -m pytest tests/test_report.py -q --tb=short`
- [x] Rebuild a real contact-module report and confirm `Nodes (0):` no longer appears


safishamsi:
Thanks! Implemented directly in v0.4.24 — the `if not real_nodes: continue` guard is in report.py.

PR #444, opened 2026-04-19
fix(cache): skip non-file source_file in save_semantic_cache

rashikkathuria:
## Summary

`save_semantic_cache` aborts a whole batch with `IsADirectoryError` when a subagent emits a node whose `source_file` happens to be a directory path instead of a file. Observed on a TS/NestJS corpus where a subagent attributed an abstract base to `src/common/database/abstracts/mongo/repositories` (the containing folder). `save_cached → file_hash → read_bytes` then crashes, so every other file in the batch fails to cache too.

This PR makes the save path tolerant:
- `save_cached(path, ...)` now no-ops when `path` is not a regular file.
- `save_semantic_cache(...)` uses `p.is_file()` instead of `p.exists()` so directories are silently skipped.

Real files in the same batch cache normally; the bogus entry is just dropped.

## Changes

- `graphify/cache.py` — 2 small defensive checks, docstring explaining the case.
- `tests/test_cache.py` — 3 new tests:
  - `test_save_cached_noop_on_directory`
  - `test_save_semantic_cache_skips_directory_source_file` (mixed batch: real file + directory)
  - `test_save_semantic_cache_skips_missing_file`

## Test plan

- [x] `pytest tests/test_cache.py` — 15/15 pass (12 existing + 3 new)
- [x] Reproduced original bug on a 288-file NestJS corpus (22-file chunk aborted mid-save)
- [x] Verified patched version caches all files in the same corpus

### DIFF
--- a/graphify/cache.py
+++ b/graphify/cache.py
@@ -73,8 +73,16 @@ def save_cached(path: Path, result: dict, root: Path = Path(".")) -> None:
 
     Stores as graphify-out/cache/{hash}.json where hash = SHA256 of current file contents.
     result should be a dict with 'nodes' and 'edges' lists.
+
+    No-ops if `path` is not a regular file. Subagent-produced semantic fragments
+    occasionally carry a directory path in `source_file` (e.g. an abstract base
+    whose "source" the model inferred as a folder). Skipping such entries keeps
+    save_semantic_cache from aborting mid-batch on a single malformed node.
     """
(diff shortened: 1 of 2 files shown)


safishamsi:
Thanks for the detailed fix and tests! Implemented directly in v0.4.24 — the `save_cached` directory guard and test coverage are in.

PR #472, opened 2026-04-21
fix: aggregated community view fallback for graphs >5000 nodes (fixes #447)

### LINKED ISSUE #1: v3: semantic query with embeddings
## Problem

## Goal

## Plan

**What changes:**

**New optional dependency:**
```
```

## Why this matters

### LINKED ISSUE #447: HTML viz hard-refuses at >5000 nodes without WebGL fallback or aggregated view
## Context
On an 8333-node graph, the skill pipeline prints:

```
Graph has 8333 nodes - too large for HTML viz. Use Obsidian vault instead.
```

and skips `graph.html` generation entirely. No fallback is offered beyond Obsidian (which requires a separate app) or the markdown wiki.

## Why this matters
The wiki is useful but is agent-oriented; for human exploration the interactive HTML graph is the primary artefact. Hitting the 5K ceiling on a mid-sized codebase (≤500 files) means visualisation is unavailable in exactly the situation where a knowledge graph is most valuable.

## Root cause
The ceiling appears to be hard-coded against vis.js DOM performance:

```python
if G.number_of_nodes() > 5000:
    print(f'Graph has {G.number_of_nodes()} nodes - too large for HTML viz.')
```

vis.js degrades past ~3–5K DOM nodes. But this is a renderer choice, not an inherent graph limit.

## Proposed fixes
Ranked by effort:

1. **Aggregated view** (cheapest): when above ceiling, emit an HTML that collapses each community to a meta-node, edges weighted by inter-community edge count. Click to expand into member nodes. Keeps vis.js.
2. **WebGL renderer option**: sigma.js or Cosmograph handle 50K+ nodes at 60fps. Ship as `--renderer webgl` flag.
3. **Explicit override**: `--force-html` for users who accept slow render.

Even #1 alone (aggregated view as default above ceiling) would eliminate the "no viz" outcome for mid-sized graphs.

## Repro
- 410 files → 8333 nodes → HTML refused.
- 5K ceiling is approximately 250-300 mid-complexity Python files.

Harsh-Codes-77:
## What this fixes

Closes #447.

When a codebase produces more than 5 000 graph nodes, graphify currently
prints `Graph has N nodes - too large for HTML viz` and skips
`graph.html` entirely. For a 400-file Python project this means the
primary human-facing output is simply absent — in exactly the situation
where a knowledge graph is most useful.

## What I changed

**File:** `graphify/skill.md` — Step 6 (HTML generation block)

Instead of stopping at the node limit, the code now:
1. Builds a **meta-graph** where each Leiden community becomes one node.
2. Sets the meta-node size proportional to how many members are in that community.
3. Counts inter-community edges and uses those counts as edge weights.
4. Calls the existing `to_html()` with the meta-graph — no changes needed
   to `export.py`.

The result is a fully interactive `graph.html` that shows community-level
structure. The user can still access full node detail via `graph.json` or
`--obsidian`.

## How to test it

1. Run graphify on a repo with 400+ Python files (or any corpus that
   produces >5000 nodes):
   
2. **Before this fix:** you see `Graph has 8333 nodes - too large for HTML viz`
   and no `graph.html` is created.

3. **After this fix:** `graph.html` is created with a community-level
   aggregated view. Opening it in a browser shows clickable meta-nodes,
   one per community. The terminal prints how many community nodes were
   generated.

## Approach chosen

This implements **Proposed fix #1** from the issue (aggregated view) —
the lowest-effort option that eliminates the hard refusal. It reuses the
existing `to_html()` function with no changes to `export.py`, keeping the
diff minimal and reviewable.

### DIFF
--- a/graphify/export.py
+++ b/graphify/export.py
@@ -344,12 +344,16 @@ def to_html(
     communities: dict[int, list[str]],
     output_path: str,
     community_labels: dict[int, str] | None = None,
+    member_counts: dict[int, int] | None = None,
 ) -> None:
     """Generate an interactive vis.js HTML visualization of the graph.
 
     Features: node size by degree, click-to-inspect panel, search box,
     community filter, physics clustering by community, confidence-styled edges.
     Raises ValueError if graph exceeds MAX_NODES_FOR_VIZ.
+
+    If member_counts is provided, node sizes and legend counts are based on
+    community member counts instead of graph degree.
     """
     if G.number_of_nodes() > MAX_NODES_FOR_VIZ:
         raise ValueError(
@@ -360,6 +364,7 @@ def to_html(
     node_community = _node_community_map(communities)
     degree = dict(G.degree())
     max_deg = max(degree.values(), default=1) or 1
+    max_mc = (max(member_counts.values(), default=1) or 1) if member_counts else 1
 
     # Build nodes list for vis.js
     vis_nodes = []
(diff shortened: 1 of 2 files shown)


safishamsi:
Implemented in v0.4.31 — `to_html()` now accepts `member_counts` and the aggregated community meta-graph is built in both `skill.md` and `skill-opencode.md` (the PR only updated skill.md). 441 tests passing.

PR #474, opened 2026-04-21
Allow Homebrew python@ hook paths in generated git hooks

### LINKED ISSUE #473: macOS: git hooks reject Homebrew python@3.11 shebang paths and silently skip graph rebuilds
## Summary

On macOS, the git hooks installed by `graphify hook install` can silently skip graph rebuilds when the `graphify` executable uses a Homebrew Python shebang such as:

```sh
#!/opt/homebrew/opt/python@3.11/bin/python3.11
```

The hook reads the shebang, sanitizes it through an allowlist, and rejects the path because `@` is not allowed. It then falls back to `python3`, which may point to the Xcode Command Line Tools Python instead of the interpreter where `graphify` is installed. In that case, `import graphify` fails and the hook exits successfully without rebuilding the graph.

## Root cause

The shared `_PYTHON_DETECT` constant in `graphify/hooks.py` uses this allowlist:

```sh
case "$GRAPHIFY_PYTHON" in
    *[!a-zA-Z0-9/_.-]*) GRAPHIFY_PYTHON="" ;;
esac
```

That pattern rejects valid Homebrew interpreter paths containing `@`, for example:

```text
/opt/homebrew/opt/python@3.11/bin/python3.11
```

After that, the fallback chain runs:

1. `GRAPHIFY_PYTHON` becomes empty
2. the hook tries `python3 -c "import graphify"`
3. on my machine `python3` is `/Library/Developer/CommandLineTools/usr/bin/python3`
4. that interpreter does not have `graphify` installed
5. the hook exits `0` without rebuilding the graph

## Repro

1. Install `graphifyy` into a Homebrew-managed Python environment whose executable path contains `@`, for example `python@3.11`
2. Confirm `graphify` has a shebang like:
   ```sh
   #!/opt/homebrew/opt/python@3.11/bin/python3.11
   ```
3. Run `graphify hook install` in a git repo that already has `graphify-out/graph.json`
4. Make a commit touching a code file, or switch branches
5. Observe that the hook does not rebuild the graph

## Actual behavior

- `post-commit` / `post-checkout` hooks are installed
- the shebang-derived interpreter path is discarded during sanitization
- fallback `python3` may not contain `graphify`
- the hook exits silently with no rebuild

## Expected behavior

A valid Homebrew Python path such as `/opt/homebrew/opt/python@3.11/bin/python3.11` should be accepted, and the hook should use that interpreter to import and run `graphify`.

## Suggested fix

Allow `@` in the interpreter path allowlist by updating the shared `_PYTHON_DETECT` constant in `graphify/hooks.py`:

```sh
case "$GRAPHIFY_PYTHON" in
    *[!a-zA-Z0-9/_.@-]*) GRAPHIFY_PYTHON="" ;;
esac
```

`_PYTHON_DETECT` is reused by both generated git hook scripts, so a single source change fixes both `post-commit` and `post-checkout`.

For users who already have the hooks installed, rerunning `graphify hook install` alone may not regenerate the existing hook files, because the installer currently treats them as already installed. They may need to run `graphify hook uninstall` followed by `graphify hook install`, unless the installer is also updated to replace existing graphify-managed hook sections.

## Environment

- OS: macOS
- package: `graphifyy 0.4.18`
- `graphify` executable: `~/Library/Python/3.11/bin/graphify`
- shebang: `#!/opt/homebrew/opt/python@3.11/bin/python3.11`
- fallback `python3`: `/Library/Developer/CommandLineTools/usr/bin/python3`

## Temporary workaround

Patch the generated git hooks, or the hook template, to allow `@` in the shebang-derived interpreter path.

ocsarpo:
## Summary
- allow `@` in the hook interpreter-path allowlist so Homebrew paths like `python@3.11` are not discarded
- add a regression test that verifies both generated hooks include the updated allowlist
- fix #473

## Root cause
The shared `_PYTHON_DETECT` snippet rejected valid shebang-derived interpreter paths containing `@`, which caused Homebrew Python paths to be cleared and forced the hook into the fallback interpreter path.

## Existing installs
Existing users can recover by rerunning `graphify hook uninstall && graphify hook install` to regenerate the installed hook scripts. It may also be possible to refresh existing hooks automatically, but that is not included in this PR.

## Validation
- `/tmp/graphify-pr-473/.venv/bin/python -m pytest tests/test_hooks.py -q`

### DIFF
--- a/graphify/hooks.py
+++ b/graphify/hooks.py
@@ -21,7 +21,7 @@ if [ -n "$GRAPHIFY_BIN" ]; then
     # Allowlist: only keep characters valid in a filesystem path to prevent
     # injection if the shebang contains shell metacharacters
     case "$GRAPHIFY_PYTHON" in
-        *[!a-zA-Z0-9/_.-]*) GRAPHIFY_PYTHON="" ;;
+        *[!a-zA-Z0-9/_.@-]*) GRAPHIFY_PYTHON="" ;;
     esac
     if [ -n "$GRAPHIFY_PYTHON" ] && ! "$GRAPHIFY_PYTHON" -c "import graphify" 2>/dev/null; then
         GRAPHIFY_PYTHON=""
--- a/tests/test_hooks.py
+++ b/tests/test_hooks.py
@@ -92,6 +92,17 @@ def test_install_creates_post_checkout_hook(tmp_path):
     assert _CHECKOUT_MARKER in hook.read_text()
 
 
+def test_install_allows_homebrew_python_paths(tmp_path):
+    repo = _make_git_repo(tmp_path)
+    install(repo)
+    expected_allowlist = '*[!a-zA-Z0-9/_.@-]*) GRAPHIFY_PYTHON="" ;;'
+    post_commit_hook = repo / ".git" / "hooks" / "post-commit"
+    post_checkout_hook = repo / ".git" / "hooks" / "post-checkout"
+
+    assert expected_allowlist in post_commit_hook.read_text()
+    assert expected_allowlist in post_checkout_hook.read_text()
+
+
 def test_install_post_checkout_is_executable(tmp_path):
     repo = _make_git_repo(tmp_path)
     install(repo)


safishamsi:
Thanks for the catch! Implemented directly in v0.4.24 — `@` added to the hooks.py allowlist.

Issue #71, opened 2026-04-08
Use OpenCode's `tool.execute.before` event instead of AGENTS.md


johnsaigle:
> Claude Code does two things: writes a CLAUDE.md section telling Claude to read graphify-out/GRAPH_REPORT.md before answering architecture questions, and installs a PreToolUse hook (settings.json) that fires before every Glob and Grep call. ...
> Codex, OpenCode, OpenClaw, Factory Droid write the same rules to AGENTS.md in your project root. These platforms don't support PreToolUse hooks, so AGENTS.md is the always-on mechanism.

I'm pretty sure this _can_ be done via OpenCode's plugins system. 

See:
* https://opencode.ai/docs/plugins/
* https://opencode.ai/docs/plugins/#tool-events specifically, which has a `tool.execute.before` event.

I think this can even be scoped to just the Glob and Grep calls in OpenCode via a opencode.json file.

safishamsi:
Confirmed — OpenCode's `tool.execute.before` plugin event is the right hook for this. The implementation requires a JavaScript plugin file (`.opencode/plugins/graphify.js`) and registration in `opencode.json`. We're tracking this for the next platform integration release. The AGENTS.md approach continues to work in the meantime.

safishamsi:
Implemented in the latest commit. Running graphify opencode install now does three things: writes the AGENTS.md section as before, writes .opencode/plugins/graphify.js with a tool.execute.before handler, and registers the plugin in opencode.json. The hook checks for graphify-out/graph.json before each bash call and prepends the graph reminder to the command output so OpenCode sees it in the tool result. Uninstall removes both the plugin file and the opencode.json entry. Thanks for flagging this.

PR #481, opened 2026-04-21
feat: add graph-query CLI rule to AGENTS.md / CLAUDE.md / GEMINI.md sections

tramzel-milc:
## Summary

Add a new rule to the three "`Rules:`" sections written to project-root instruction files (`_CLAUDE_MD_SECTION`, `_AGENTS_MD_SECTION`, `_GEMINI_MD_SECTION`) naming `graphify query`, `graphify path`, and `graphify explain` as preferred tools over `grep` for cross-module architecture questions. Mirrors the MCP-tool guidance that's already in `_ANTIGRAVITY_RULES`, extended with a parallel CLI-fallback line there too.

## Motivation

On platforms where graphify's proactive hook covers only `bash` (OpenCode, per #71) or where there's no hook at all (AGENTS.md-only platforms), agents default to `grep` when answering architecture questions — because the instructions don't name the graph-traversal commands.

Observed this on OpenCode with the question *"What is the exact relationship between JWT Token Authentication and One-CommonAPI Sequelize Models?"*:

1. Agent read `GRAPH_REPORT.md` (following rule 1 ✓)
2. Glob'd `graphify-out/wiki/index.md` (following rule 2, file existed ✓)
3. Went straight to `grep` across the repo

The agent never tried `graphify query "..."` / `graphify path "A" "B"` / `graphify explain "..."` because they weren't mentioned in `_AGENTS_MD_SECTION`. For that particular question (an inter-module path), `graphify path` would've been the ideal first step.

## The change

Four rule sections get one new line each:

**`_CLAUDE_MD_SECTION`, `_AGENTS_MD_SECTION`, `_GEMINI_MD_SECTION`** — add:
> `- For cross-module "how does X relate to Y" questions, prefer ` `graphify query "<question>"` `, ` `graphify path "<A>" "<B>"` `, or ` `graphify explain "<concept>"` ` over grep — these traverse the graph's EXTRACTED + INFERRED edges instead of scanning files`

**`_ANTIGRAVITY_RULES`** — already mentions MCP tools; add a parallel CLI fallback:
> `- If the MCP server is not active, the CLI equivalents are ` `graphify query "<question>"` `, ` `graphify path "<A>" "<B>"` `, and ` `graphify explain "<concept>"` ` — prefer these over grep for cross-module questions`

## Scope

- `_VSCODE_INSTRUCTIONS_SECTION` left alone — shorter, different structure; happy to also update if desired
- `_KIRO_STEERING` left alone — its single-paragraph form is already abstract ("navigate by graph structure")
- No test changes — `tests/test_install.py` doesn't assert section content

## Why now

After #71 added the OpenCode plugin, the remaining gap for OpenCode (and other AGENTS.md-only platforms like Codex / OpenClaw) is that agents don't know the graph-query CLI exists. The instructions file is the one surface that reaches every supported platform.

### DIFF
--- a/graphify/__main__.py
+++ b/graphify/__main__.py
@@ -191,6 +191,7 @@ This project has a graphify knowledge graph at graphify-out/.
 Rules:
 - Before answering architecture or codebase questions, read graphify-out/GRAPH_REPORT.md for god nodes and community structure
 - If graphify-out/wiki/index.md exists, navigate it instead of reading raw files
+- For cross-module "how does X relate to Y" questions, prefer `graphify query "<question>"`, `graphify path "<A>" "<B>"`, or `graphify explain "<concept>"` over grep — these traverse the graph's EXTRACTED + INFERRED edges instead of scanning files
 - After modifying code files in this session, run `graphify update .` to keep the graph current (AST-only, no API cost)
 """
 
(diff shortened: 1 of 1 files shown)


safishamsi:
Implemented in v0.4.25: added the graph-query CLI rule to _CLAUDE_MD_SECTION, _AGENTS_MD_SECTION, _GEMINI_MD_SECTION, and the MCP fallback line to _ANTIGRAVITY_RULES, exactly as proposed. Good catch — agents on AGENTS.md-only platforms had no signal that graphify query/path/explain existed.

PR #482, opened 2026-04-21
fix(extract): Java extends/implements + resolve cross-file imports

### LINKED ISSUE #435: Feature Request: Add Java Inheritance Relationship Support in Tree-sitter Parsing
### Description

Currently, the tool supports extracting method call relationships for Java code. However, inheritance relationships (e.g., extends, implements) are not captured, which limits the completeness of the code graph for Java projects.

In contrast, inheritance relationships are already supported for other languages such as Python and C#.

### Problem Statement

When parsing Java code, the absence of inheritance edges leads to:

- Incomplete dependency graphs
- Reduced accuracy in code understanding and analysis
- Inconsistency across language support

### Expected Behavior
The parser should:

- Detect extends relationships between classes
- Detect implements relationships for interfaces
- Add corresponding edges in the extracted graph
- 
### Suggested Approach
Enhance Tree-sitter Java parsing logic to capture:
- class_declaration → superclass
- class_declaration → interfaces
Map these to inheritance edges similar to existing implementations in Python and C#

Optionally distinguish edge types:
- INHERITS_FROM
- IMPLEMENTS

Goodday12:
Closes #435 (Java inheritance support). Also fixes a related Java-imports bug that would have blocked #435 from actually working end-to-end.

## Summary

Two issues in Java AST extraction that silently drop all Java-type edges from the graph. On any real Java corpus, the `structural_type` layer (imports / extends / implements) comes out **empty**.

### Issue 1 — Java inheritance was never extracted

The class handler inside `_extract_generic` has language-specific blocks for Python (`inherits`), Swift (`inherits`), and C# (`inherits`) — but **nothing for Java**, even though `tree-sitter-java` exposes the `superclass` / `interfaces` / `extends_interfaces` fields.

Result: for a Java corpus of 5,727 files, `extends` = 0 and `implements` = 0.

### Issue 2 — Java import edges had dangling targets

`_import_java` emits edges whose `target` is just the bare class name (`_make_id(module_name)` → e.g. `list` for `java.util.List`), while real node ids use `{stem}_{classname}`. Those edges never match a node id, so `build_from_json` silently drops every single one at line 75-76 of `build.py`.

Result: `imports` = 0 in the built graph, even though `_import_java` ran.

### `graphify/extract.py`

- **New `tree_sitter_java` block in the class handler** (alongside the existing Python / Swift / C# blocks). Emits:
  - `extends` edge for `class X extends Y` (via `node.child_by_field_name(\"superclass\")`)
  - `implements` edges for `class X implements A, B` (via `node.child_by_field_name(\"interfaces\") → super_interfaces → type_list`)
  - `extends` edges for `interface X extends A, B` (via the `extends_interfaces` child)
  - When the parent/interface is defined outside the current file, a synthetic external node is added (same pattern as the existing Python / Swift / C# inheritance blocks).

- **New `_resolve_cross_file_java_imports` helper** mirroring the existing `_resolve_cross_file_imports` (Python). Runs a second pass on Java corpora, builds a `{ClassName: [node_id, …]}` index across all nodes, re-parses each file's imports, and emits resolved `imports` edges pointing at real class nodes elsewhere in the corpus.
  - Wildcard imports (`a.b.*`) are skipped — the target is ambiguous.
  - `static` imports strip the `static` keyword and resolve the class name (not the method).
  - Stdlib / third-party names left unresolved produce no edge (they would be dropped anyway).

- **`extract()` now calls the Java resolver** after the existing Python one, in the same try/except style so an import failure in `tree_sitter_java` only logs a warning.

### Tests

All four new tests pass locally; the full upstream suite (**437 tests**) passes with no regressions.

- `tests/test_languages.py::test_java_finds_extends`
- `tests/test_languages.py::test_java_finds_implements`
- `tests/test_languages.py::test_java_finds_interface_extends`
- `tests/test_multilang.py::test_extract_resolves_java_cross_file_imports`

Two new fixtures: `sample_inherit.java` (single inheritance + multi-interface + interface-extends-interface) and `sample_imported.java` (defines a class imported cross-file).

## Validated on a real Java corpus

Spring / JPA / Lombok monorepo of **5,727 `.java` files** parsed with no errors:

| relation | before this PR | after this PR |
| --- | ---: | ---: |
| `extends` | **0** | **2,072** |
| `implements` | **0** | **1,285** |
| `imports` (internal) | **0** | **58,019** |

Before the fix, the entire `structural_type` layer of this corpus had **11 edges** (all from README.md via the LLM pass, none from AST). After the fix it has **61,376** real AST edges.

## Test plan

- [x] All 4 new tests pass against the patched `graphify/extract.py`
- [x] Full upstream `pytest tests/` — 437 passed, 0 failed
- [x] Validated on 5,727-file Java codebase: counts above
- [x] Cross-language regression check: no changes to non-Java code paths

## Notes

- `_import_java` is left in place. The edges it produces still reach `build_from_json` and still get dropped for stdlib / third-party imports — that is the expected behaviour for dangling externals. The new cross-file pass now re-resolves the internal ones correctly.
- `extends` / `implements` are used as distinct relation names (rather than the `inherits` used by Python / Swift / C#) because Java separates the two cleanly and the distinction is useful downstream. If you'd prefer to unify under `inherits`, happy to adjust.

### DIFF
--- a/graphify/extract.py
+++ b/graphify/extract.py
@@ -803,6 +803,52 @@ def _extract_generic(path: Path, config: LanguageConfig) -> dict:
                                         seen_ids.add(base_nid)
                                 add_edge(class_nid, base_nid, "inherits", line)
 
+            # Java-specific: extends (superclass) / implements (super_interfaces) / interface-extends
+            if config.ts_module == "tree_sitter_java":
+                def _emit_java_parent(base_name: str, rel: str, at_line: int) -> None:
+                    if not base_name:
+                        return
+                    base_nid = _make_id(stem, base_name)
+                    if base_nid not in seen_ids:
+                        base_nid = _make_id(base_name)
+                        if base_nid not in seen_ids:
+                            nodes.append({
+                                "id": base_nid,
+                                "label": base_name,
+                                "file_type": "code",
+                                "source_file": "",
+                                "source_location": "",
+                            })
+                            seen_ids.add(base_nid)
+                    add_edge(class_nid, base_nid, rel, at_line)
+
+                # class X extends Y  -> superclass field on class_declaration
+                sup = node.child_by_field_name("superclass")
+                if sup is not None:
+                    for sub in sup.children:
+                        if sub.type == "type_identifier":
+                            _emit_java_parent(_read_text(sub, source), "extends", line)
+                            break  # Java has single inheritance for classes
+
+                # class X implements A, B  -> interfaces field -> super_interfaces -> type_list
+                ifs = node.child_by_field_name("interfaces")
+                if ifs is not None:
+                    for sub in ifs.children:
+                        if sub.type == "type_list":
+                            for tid in sub.children:
+                                if tid.type == "type_identifier":
+                                    _emit_java_parent(_read_text(tid, source), "implements", line)
+
+                # interface X extends A, B  -> extends_interfaces child -> type_list
+                if t == "interface_declaration":
+                    for child in node.children:
+                        if child.type == "extends_interfaces":
+                            for sub in child.children:
+                                if sub.type == "type_list":
+                                    for tid in sub.children:
+                                        if tid.type == "type_identifier":
+                                            _emit_java_parent(_read_text(tid, source), "extends", line)
+
             # Find body and recurse
             body = _find_body(node, config)
             if body:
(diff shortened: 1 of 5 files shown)


WHALE88 (approved):

jaygandhi129 (approved):

safishamsi:
Implemented in v0.4.31 — Java `extends`/`implements`/interface-extends extraction added to the structural walker, plus `_resolve_cross_file_java_imports()` two-pass resolver. Both wired into `extract()` with exception guards.

PR #489, opened 2026-04-22
feat(extract): add Lean 4 AST support via tree-sitter-language-pack

Dewars30:
## Summary
- Adds `tree-sitter-language-pack>=1.6.2` dependency (provides 305+ languages including Lean 4 under the `"lean"` key)
- Implements hand-written `extract_lean()` in `graphify/extract.py` (the config-driven `LanguageConfig` pattern didn't fit Lean's grammar — every construct is wrapped in a generic `declaration` node discriminated by anonymous keyword tokens one level deeper)
- Maps Lean constructs:
  - `theorem` / `def` / `abbrev` / `instance` / `lemma` → Function (with `subtype` set to original keyword)
  - `structure` / `class` / `inductive` → Class
  - `namespace` → Module
  - `import` → Dependency edge with relation `imports`
- Registers `.lean` in `_DISPATCH` (extract.py) and `CODE_EXTENSIONS` (detect.py)
- 6 pytest cases covering all node types against a Fulcrum-Proofs-style fixture
- Tested on the Fulcrum-Proofs repo (Mathlib-adjacent Lean 4): produces ~115 Lean AST nodes from ~30 files, including 36 theorem nodes

## Test plan
- [x] `pytest tests/test_lean_extractor.py -v` — all 6 pass
- [x] End-to-end: `graphify update .` against a Lean-only directory produces expected nodes
- [x] Integration: `graphify update .` against a real Lean 4 repo (Fulcrum-Proofs) surfaces theorem/structure/def nodes with correct subtypes


Qodo-Free-For-OSS:
Hi, extract_lean() emits `imports` edges whose `target` is a raw module string, not a node id. build_from_json() drops edges whose endpoints are not in the node set, so Lean import edges will never make it into the built graph.

**How to fix:** Create import nodes or ids

> ### Issue description
> `extract_lean()` creates `imports` edges whose `target` is a raw module name string, but the graph builder keeps only edges that connect existing node IDs. As a result, Lean import edges disappear in the final graph.
> 
> ### Issue Context
> Other extractors either (a) normalize targets via `_make_id(...)` so they can match an existing node ID when present, or (b) create explicit nodes for import targets when they want imports preserved.
> 
> ### Fix Focus Areas
> - graphify/extract.py[1508-1517]
> - graphify/build.py[40-76]
> 
> ### What to change
> - Option A (recommended): create a node for each imported module (e.g., `tgt_nid = _make_id(module_name)`; `add_node(tgt_nid, module_name, line, node_type="Module")`) and emit `add_edge(file_nid, tgt_nid, "imports", line)`.
> - Option B: if you intentionally don’t want import target nodes, then change the build/validation expectations (not recommended because it affects the whole schema/pipeline).

---

Dewars30:
Follow-up fix pushed (2ab1c0e): extract_lean() now creates a Module node for each imported module via _make_id() and targets the edge against that node id. Before this, import edges pointed at raw dotted module strings, so build_from_json() dropped all of them at the node-set endpoint check. Verified on proofs/lean/Proofs.lean: 14 import edges now resolve against the node set (was 0/14) and land in the built graph. Same pattern other extractors already use (cf. _import_python).

PR #434, opened 2026-04-18
fix(watch): keep update outputs project-relative

takeItIzzy:
## Summary

Fixes #433.

`graphify update` should keep generated paths project-relative for code-only rebuilds, matching the rest of the pipeline.

## Root cause

`graphify.watch._rebuild_code()` resolved the requested path before detection and report generation. That changed a relative invocation such as `graphify update .` into an absolute rebuild root, which then leaked absolute paths into:

- `graphify-out/graph.json` `source_file` fields
- `graphify-out/GRAPH_REPORT.md` title and path-bearing sections
- `graphify-out/graph.html` title and node metadata
- `graphify-out/cache/*.json` AST cache entries

## What changed

- keep the cache root absolute, but run detection against the user-provided path so extractor inputs stay project-relative for normal `graphify update .` usage
- normalize merged rebuild payloads back to project-relative `source_file` values before report/JSON/HTML export
- render a non-absolute report root label for the common `.` case
- add watch rebuild regression tests for repo-root and subdirectory invocations

## Validation

- `.venv/bin/python -m pytest -q tests/test_watch.py`
- `.venv/bin/python -m pytest -q tests/test_pipeline.py`
- manual smoke test: `.venv/bin/python -m graphify update .` in a temporary repo, verifying no absolute project paths in `graph.json`, `GRAPH_REPORT.md`, `graph.html`, or `graphify-out/cache/*.json`

Issue #433, opened 2026-04-18
graphify update writes absolute paths into generated artifacts


takeItIzzy:
## Summary

`graphify update .` currently serializes absolute filesystem paths into the generated artifacts for code-only rebuilds.

The regression is specific to the `update` / `watch` rebuild path. Running the full `/graphify` workflow on a relative path still produces project-relative `source_file` values, but `graphify update .` does not.

## Affected outputs

- `graphify-out/graph.json` node and edge `source_file` fields
- `graphify-out/GRAPH_REPORT.md` report title and any path-bearing sections
- `graphify-out/graph.html` title and node metadata
- `graphify-out/cache/*.json` AST cache entries

## Reproduction

1. Create a small repo with `pkg/a.py`.
2. From the repo root, run `graphify update .`.
3. Inspect `graphify-out/graph.json` or `graphify-out/GRAPH_REPORT.md`.

## Actual behavior

Generated outputs contain absolute paths such as:

- `/Users/alice/project/pkg/a.py`
- `# Graph Report - /Users/alice/project (...)`

This makes committed graph artifacts machine-specific and leaks local filesystem layout into reports, HTML, and cache files.

## Expected behavior

`graphify update .` should preserve project-relative paths, matching the rest of the pipeline:

- `pkg/a.py` in `source_file`
- a non-absolute report title for the project root
- no absolute project paths in `graphify-out/cache/*.json`

## Root cause

`graphify.watch._rebuild_code()` resolves the input path before running detection and report generation:

- `watch_path = watch_path.resolve()`
- `detect(watch_path, ...)` therefore returns absolute file paths
- the AST extractors serialize those absolute paths directly into `source_file`
- `report.generate(..., str(watch_path), ...)` also receives an absolute root label

## Scope

This issue is about path serialization only. The large-graph HTML failure path was already handled separately in `v0.4.23`.

takeItIzzy:
Opened draft PR #434 with a fix and regression coverage for this path serialization bug.

PR #503, opened 2026-04-22
fix(extract): keep graph node ids relative

### LINKED ISSUE #502: graph.json edge endpoints still leak absolute-path-derived file ids
## Summary

This is a follow-up to #433.

The fix for `source_file` path serialization in #433 / #434 does not cover graph edge endpoint IDs. When `extract()` receives absolute paths, file node IDs are still derived from those absolute paths and `graphify-out/graph.json` exposes them through `links[].source`, `links[].target`, `links[]._src`, and `links[]._tgt`.

## Reproduction

1. Create two sibling files with the same basename, for example `pkg1/main.py` and `pkg2/main.py`.
2. Call `extract()` with absolute file paths, or run a graph build path that feeds absolute code paths into AST extraction.
3. Inspect `graphify-out/graph.json`.

## Actual behavior

Edge endpoint IDs contain machine-specific prefixes such as `users_alice_project_pkg1_main_py`.

## Expected behavior

File node IDs should be derived from project-relative paths, so graph JSON contains stable endpoint IDs like `pkg1_main_py` and `pkg2_main_py`.

## Root cause

`graphify.extract.extract()` infers a common root incorrectly by counting matching path segments even after divergence, and the extraction payload is not normalized before graph construction. That leaves absolute-path-derived file node IDs in edges even when `source_file` has already been relativized elsewhere.

## Scope

This is related to #433, but narrower: `source_file` may already be relative while `graph.json` edge endpoints still leak absolute-path-derived node IDs.

takeItIzzy:
## Summary

- normalize extracted `source_file` values relative to the inferred project root before graph construction
- remap file node IDs when they were derived from absolute paths so `graph.json` edge endpoints stay machine-independent
- add regression coverage for absolute-path inputs and exported `graph.json` edge endpoint fields

## Why

Issue #502 is the remaining path leak after #433 / #434: `source_file` may already be relative, but `links[].source`, `links[].target`, `links[]._src`, and `links[]._tgt` can still expose absolute-path-derived file node IDs.

Closes #502.

## Testing

- `/tmp/graphify-fix/venv/bin/python -m pytest tests/test_extract.py`
- `/tmp/graphify-fix/venv/bin/python -m pytest tests/test_build.py tests/test_export.py tests/test_cache.py tests/test_watch.py tests/test_multilang.py`

### DIFF
--- a/graphify/extract.py
+++ b/graphify/extract.py
@@ -18,6 +18,66 @@ def _make_id(*parts: str) -> str:
     return cleaned.strip("_").lower()
 
 
+def _normalize_source_path(path_str: str, root: Path | None) -> str:
+    """Return a stable display path relative to root when possible.
+
+    Graphify often receives absolute file paths from callers running on a
+    project root or worktree. Absolute paths leak machine-specific prefixes
+    into file node IDs, which then surface in graph.json edge endpoints.
+    """
+    if not path_str:
+        return path_str
+    path = Path(path_str)
+    if not path.is_absolute():
+        return path.as_posix()
+    candidates: list[Path] = []
+    if root is not None:
+        candidates.append(root.resolve())
+    candidates.append(Path.cwd().resolve())
+    resolved = path.resolve()
+    for base in candidates:
+        try:
+            return resolved.relative_to(base).as_posix()
+        except ValueError:
+            continue
+    return resolved.as_posix()
+
+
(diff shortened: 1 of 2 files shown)


safishamsi:
Thanks! The node ID relative-path fix was already shipped in v0.4.27 (issue #502). The common-root inference and ID remap post-processing were added to extract.py in that release. Closing as duplicate.

PR #476, opened 2026-04-21
fix: skip empty communities in GRAPH_REPORT.md

dsremo:
## Problem

On projects where the Leiden/Louvain clustering algorithm produces communities whose only members are file-level stub nodes (filtered by `_is_file_node`), the report was still writing empty community blocks like:

```
### Community 47 - "Community 47"
Cohesion: 1.0
Nodes (0): 
```

On a mid-size TypeScript codebase (168 files) this produced **80 empty community blocks out of 113 total**. On a larger Django backend (374 files) it produced **142 empty blocks out of 180 total**. These add zero useful signal and inflate the report significantly — wasting tokens when an LLM reads the report.

The same empty communities also appeared in:
- **Community Hubs navigation** — linking to community files that have no content
- **Knowledge Gaps section** — listed as "thin communities" when they are really just clustering artifacts with no real nodes at all

## Fix

- Compute `non_empty` once up front: communities that contain at least one non-file node
- Use `non_empty` count in the Summary line instead of raw `len(communities)`
- Skip communities with `not real_nodes` in the Communities section
- Only include communities with `0 < len(real_ns) < 3` in `thin_communities` (Knowledge Gaps), excluding zero-node artifacts
- Remove empty entries from the Community Hubs navigation

## Before / After

**Before** (113 communities reported, 80 empty):
```
## Summary
- 502 nodes · 514 edges · 113 communities detected

## Community Hubs (Navigation)
- [[_COMMUNITY_Community 0|Community 0]]
...113 entries including empties...

### Community 47 - "Community 47"
Cohesion: 1.0
Nodes (0): 
```

**After** (only 13 real communities shown):
```
## Summary
- 502 nodes · 514 edges · 13 communities detected

## Community Hubs (Navigation)
- [[_COMMUNITY_Community 0|Community 0]]
...13 entries, all with content...

(Community 47 not shown — empty)
```

PR #523, opened 2026-04-23
fix: restore empty-community filter lost in v2 report rewrite

dsremo:
## Summary

The empty-community filter originally shipped in **v0.4.25 (PR #476)** regressed during the v2 report rewrite. On mid-size codebases 70-80% of communities are clustering artifacts whose entire membership is file-level stubs (filtered by `_is_file_node`), and these are once again polluting the Summary count, the Communities section, and the Knowledge Gaps list.

This PR restores the three behaviours from v0.4.25 that are gone on current `main`:

1. **Summary** now reports `len(non_empty)` — communities that contain at least one non-stub node. Before: 287 on my aria-core graph. After: a much smaller, meaningful number.
2. **Communities section** skips entries whose `real_nodes` list is empty, dropping blank `Nodes (0):` blocks.
3. **Knowledge Gaps** (`thin_communities`) now keeps only communities with 1 or 2 real nodes, excluding zero-real-node artifacts that aren't "too small to be a meaningful cluster" warnings — they are not clusters of real nodes at all.

## Verification against current `main`

On the 14996-node aria-core graph, `graphify update src/` currently reports:
```
## Summary
- 14996 nodes · 38860 edges · 287 communities detected
## Communities
### Community 47 - "Community 47"
Cohesion: 1.0
Nodes (0):
```
…and dozens of identical empty `Community N` blocks. After this patch, the Summary count drops to the non-empty subset and the empty blocks are suppressed entirely.

## Tests

`tests/test_report_empty_communities.py` — 3 new tests:

- `test_summary_counts_only_non_empty_communities`
- `test_empty_community_not_rendered_in_communities_section`
- `test_stub_only_community_excluded_from_knowledge_gaps`

Each drives `generate()` with a 3-node graph and two communities (one with a real `class Authenticator` node, one with only file stubs `auth.py` + `models.py`) and asserts the stub-only community is suppressed in every downstream section. No graspologic dependency — runs on any dev box.

## History

- [PR #476](https://github.com/safishamsi/graphify/pull/476) — the original, closed with the v0.4.25 fix applied.
- This is a clean re-submission against current `main` after the v2 rewrite lost the filter.

### DIFF
--- a/graphify/report.py
+++ b/graphify/report.py
@@ -28,6 +28,16 @@ def generate(
     inf_scores = [d.get("confidence_score", 0.5) for _, _, d in inf_edges]
     inf_avg = round(sum(inf_scores) / len(inf_scores), 2) if inf_scores else None
 
+    # Communities whose members are all file-level stubs (filtered out by
+    # _is_file_node in the Communities section below) add pure noise — on
+    # mid-size codebases they accounted for 70-80% of communities and
+    # distorted the Summary count / Knowledge Gaps flags. The filter was
+    # shipped in v0.4.25 (PR #476) and regressed during the v2 report
+    # rewrite; this restores it.
+    from .analyze import _is_file_node as _ifn
+    non_empty = {cid: nodes for cid, nodes in communities.items()
+                 if any(not _ifn(G, n) for n in nodes)}
+
     lines = [
         f"# Graph Report - {root}  ({today})",
         "",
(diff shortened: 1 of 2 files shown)


dsremo:
Closing - my mistake, I targeted main (0.1.14, stale v1 line) instead of v4 (0.4.27, the active default branch). v4 already has all three behaviors I tried to restore (non_empty dict, skip-empty in Communities, and filtered thin_communities). No regression on the real development line. Sorry for the noise.

PR #542, opened 2026-04-25
Closes #541: GRAPHIFY_NO_VIZ + GRAPHIFY_VIZ_NODE_LIMIT env vars

### LINKED ISSUE #541: Expose --no-viz flag and/or GRAPHIFY_VIZ_NODE_LIMIT env var
**Labels:** hook, viz, large-repo

**Body:**

## Problem

`graphify hook install` writes a `.git/hooks/post-commit` whose `_HOOK_SCRIPT` calls `to_html()` unconditionally. On graphs with more than ~5,000 nodes, `to_html()` raises (the pyvis layout pass blows up). The hook fails after every commit on large repos.

We patched our local copy of the hook (2026-04-25) to gate the viz step:

```python
if G.number_of_nodes() <= 5000:
    to_html(...)
else:
    print('[graphify hook] HTML viz skipped (> 5000 nodes); graph.json + GRAPH_REPORT.md updated.')
```

But `graphify hook install` regenerates the script from upstream `_HOOK_SCRIPT`, clobbering the patch on every reinstall.

## Reproducer

1. Install the hook on any repo whose code-graph exceeds 5,000 nodes (a 50k-LOC Django project does it).
2. Make a commit.
3. Observe the hook fail during `to_html` rendering.

## Proposal

Any one of:

1. **Flag.** `graphify hook install --no-viz` — generates a hook that skips `to_html` entirely. Simplest, opt-in.
2. **Env var.** `GRAPHIFY_VIZ_NODE_LIMIT=5000` (default) — the generated hook reads the env var and skips `to_html` above the threshold, printing a warning. Zero-config for most users; tunable for large repos.
3. **Both.** Env var as default behavior, `--no-viz` as explicit opt-out.

Option 2 is the safest default — the hook stays useful on every repo size without manual patching, and CI consumers (who don't need the HTML) can set the limit to 0.

## Motivation

Keeps the post-commit hook trustworthy across repo sizes. Also unblocks downstream users who don't need the HTML viz at all (CI runners, headless dev boxes).

saxster:
Closes #541.

## Summary

Adds two env-var opt-outs to `_rebuild_code` in `graphify/watch.py` so users can skip the HTML viz step without patching the post-commit hook script:

- **`GRAPHIFY_NO_VIZ=1`** — unconditional skip; for CI runners and headless dev boxes that never open `graph.html`.
- **`GRAPHIFY_VIZ_NODE_LIMIT=N`** — soft cap; skip when node count exceeds N.

`GRAPHIFY_NO_VIZ` takes priority. Both default to off, preserving current behavior. The pre-existing `ValueError` catch around `to_html` stays as a backstop for unrelated rendering failures.

## Why

v0.5.0 catches `ValueError` from `to_html` so the hook no longer fails on >5000-node graphs — but it still pays the full `to_html` attempt before the guard rejects. On a 6000-node Django graph this was costing several seconds per commit. The env vars let the user short-circuit before the attempt.

## Implementation

New helper `_viz_skip_reason(node_count)` reads the two env vars and returns a reason string (or None). The `to_html` call site checks the reason first and skips with a printed message if set.

## Tests

7 new tests in `tests/test_watch.py`:
- defaults off (no env vars set)
- `GRAPHIFY_NO_VIZ` truthy values
- `GRAPHIFY_NO_VIZ` falsy values (`0`, `false`, `no`, empty, whitespace)
- `GRAPHIFY_VIZ_NODE_LIMIT` threshold (one below, equal, one above)
- invalid (non-int) `GRAPHIFY_VIZ_NODE_LIMIT` silently treated as unset
- `GRAPHIFY_NO_VIZ` priority over `GRAPHIFY_VIZ_NODE_LIMIT`

`pytest tests/test_watch.py` — 18/18 pass. Full suite — 440/447 (the 7 failures are pre-existing in v0.5.0 and unrelated to this change; verified by stashing the diff and re-running).

## Test plan

- [x] `pytest tests/test_watch.py -q`
- [x] `pytest -q` (full suite — same baseline failures, no new ones)
- [ ] Manual: `GRAPHIFY_NO_VIZ=1 graphify update .` shows skip message + no `graph.html` produced
- [ ] Manual: `GRAPHIFY_VIZ_NODE_LIMIT=10 graphify update .` on a >10-node repo shows skip message

### DIFF
--- a/graphify/watch.py
+++ b/graphify/watch.py
@@ -1,12 +1,39 @@
 # monitor a folder and auto-trigger --update when files change
 from __future__ import annotations
 import json
+import os
 import sys
 import time
 from pathlib import Path
 
 
 from graphify.detect import CODE_EXTENSIONS, DOC_EXTENSIONS, PAPER_EXTENSIONS, IMAGE_EXTENSIONS
+
+
+def _viz_skip_reason(node_count: int) -> str | None:
+    """Return a reason string if the HTML viz step should be skipped, else None.
+
+    Two opt-outs, in priority order:
+
+    - ``GRAPHIFY_NO_VIZ=1`` (or any truthy value): skip unconditionally.
+      For CI runners and headless dev boxes that never open graph.html.
+    - ``GRAPHIFY_VIZ_NODE_LIMIT=N``: skip if the graph has more than N
+      nodes. Defaults are off; callers who want a soft cap can set
+      this in their env. Useful when ``to_html`` is slow on large
+      graphs and the user knows they don't need the visualization.
+    """
+    no_viz = os.environ.get("GRAPHIFY_NO_VIZ", "").strip().lower()
+    if no_viz and no_viz not in ("0", "false", "no", ""):
+        return "GRAPHIFY_NO_VIZ is set"
+    raw_limit = os.environ.get("GRAPHIFY_VIZ_NODE_LIMIT", "").strip()
+    if raw_limit:
+        try:
+            limit = int(raw_limit)
+        except ValueError:
+            return None
+        if node_count > limit:
+            return f"GRAPHIFY_VIZ_NODE_LIMIT={limit} (graph has {node_count} nodes)"
+    return None
 
 _WATCHED_EXTENSIONS = CODE_EXTENSIONS | DOC_EXTENSIONS | PAPER_EXTENSIONS | IMAGE_EXTENSIONS
 _CODE_EXTENSIONS = CODE_EXTENSIONS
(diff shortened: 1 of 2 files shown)


Qodo-Free-For-OSS:
Hi, When viz is skipped, `_rebuild_code` unlinks a stale `graph.html` without handling `OSError`, so file permission/locking issues can make the entire rebuild fail even though core outputs were written. The new env-var skip path increases how often this cleanup branch runs, raising the likelihood of hitting this failure mode.

**How to fix:** Make stale delete best-effort

> ### Issue description
> `_rebuild_code()` deletes `graph.html` when viz is skipped. If the file is locked or permissions prevent deletion, `Path.unlink()` raises and the outer `except Exception` causes the whole rebuild to fail.
> 
> ### Issue Context
> The env-var early-skip path adds another route into the stale-delete branch, increasing exposure.
> 
> ### Fix Focus Areas
> - graphify/watch.py[141-155]
> 
> ### Suggested change
> Wrap `stale.unlink()` in a `try/except OSError` (or `Exception`) that logs a warning and continues, so core rebuild products are not marked failed due to cleanup.

---
*Found by [Qodo](https://github.com/marketplace/qodo-merge-pro-for-open-source) code review. FYI, Qodo is free for open-source.*

Issue #549, opened 2026-04-25
Feature request: post-commit hook should optionally rebuild the wiki


Beemgit:
## Feature request: post-commit hook should optionally rebuild the wiki

Today `graphify hook install` produces a hook that rebuilds
`graph.json` / `graph.html` / `GRAPH_REPORT.md` after each commit but
skips the wiki. Wiki drift is silent: agents read stale community
articles after every code change.

### Proposed

`graphify hook install --with-wiki` flag adds a wiki rebuild step to
the post-commit hook. Implementation is a small addition that calls
`graphify.wiki.to_wiki(G, communities, output_dir, ...)` from the
existing `graph.json` after the AST rebuild — no LLM cost, ~1-2 seconds
for a 1000-node graph.

### Workaround today

Users either run `graphify <path> --wiki` manually on a cadence, or
extend the hook by hand (which the next `graphify hook install` may
clobber). Both work, but staleness is silent and the manual rebuild
is easy to forget.

### Why it matters

Agents that follow the integration guidance (e.g. the `## graphify`
section that `graphify claude install` writes into `CLAUDE.md`) prefer
`graphify-out/wiki/index.md` over raw file reads. If the wiki is
stale relative to the AST graph, the agent navigates outdated
community structure for current code state.

### Suggested shape

```
graphify hook install                    # existing behavior, code-only
graphify hook install --with-wiki        # adds wiki rebuild
graphify hook install --without-wiki     # explicit opt-out (default)
```

Or a config knob in a `.graphify.toml` so projects can opt in once
without re-running `hook install`.

### Implementation sketch

The post-commit hook currently calls (roughly) `graphify update .`.
Adding a wiki step would be:

```python
from pathlib import Path
import json
from graphify.build import build_from_json
from graphify.cluster import score_all
from graphify.wiki import to_wiki
from graphify.analyze import god_nodes
from networkx.readwrite import json_graph

data = json.loads(Path('graphify-out/graph.json').read_text())
G = json_graph.node_link_graph(data, edges='links')
communities = {}
for nid, d in G.nodes(data=True):
    cid = d.get('community')
    if cid is not None:
        communities.setdefault(int(cid), []).append(nid)
to_wiki(G, communities, 'graphify-out/wiki',
        cohesion=score_all(G, communities), god_nodes_data=god_nodes(G))
```

Adds ~1-2s to commit time on a 1000-node graph; opt-in via flag keeps
default behavior unchanged.

Happy to PR if direction is acceptable.

Issue #229, opened 2026-04-11
[skill] --wiki flag has no implementation step: running after Step 9 cleanup silently loses community labels


C-Hung85:
**Problem**

  The `--wiki` flag is documented in the usage section but Step 6 of the skill only handles
  `--obsidian` and HTML. There is no wiki generation step.

  This creates two related failure modes:

  1. **Pipeline ordering bug:** Anyone who uses `--wiki` does so outside the documented flow —
     typically after Step 9, at which point `.graphify_labels.json` has already been deleted.
     `to_wiki()` receives `None` for `community_labels` and silently falls back to
     `[[Community 0]]`, `[[Community 1]]`, etc., discarding the labels set in Step 5.

  2. **Post-pipeline wiki regeneration is impossible:** Even if a user ran the full pipeline
     correctly (without `--wiki`), they cannot regenerate the wiki later with correct labels.
     Community labels are ephemeral — deleted in Step 9 — and not persisted anywhere.
     `graph.json` stores integer community IDs per node but not the human-readable names.

  **Expected behavior**

  Wiki generation should run in Step 6 (before Step 9 cleanup), with access to
  `.graphify_labels.json`. The `--wiki` flag should remain opt-in, consistent with
  `--obsidian`, `--svg`, and other optional outputs.

  **Two solution paths**

  *Path A — Fix the skill only:*
  Add a wiki generation block to Step 6, between the HTML block and Step 7, that reads
  `.graphify_labels.json` before Step 9 deletes it. Simple, minimal change.

  *Path B — Fix the root cause:*
  Persist community labels to `graphify-out/labels.json` as a first-class output artifact
  in Step 5 (alongside `graph.json` and `GRAPH_REPORT.md`). This also enables:
  - Post-pipeline wiki regeneration at any time
  - `--update` and `--cluster-only` reruns that can reuse existing labels without
    re-running the full labeling step

  **Note:** this is separate from #228 (wikilink format issue in `wiki.py`) which concerns
  the library output format. This issue is about the skill's pipeline ordering and label
  persistence.

safishamsi:
Implemented in v0.4.14.

```bash
pip install --upgrade graphifyy
```

PR #558, opened 2026-04-26
fix(wiki): clear stale articles before regenerating to prevent orphan accumulation

szsip239:
## Problem

`to_wiki()` writes a fresh set of community + god-node articles each call but never deletes old files from previous runs. Since community labels are LLM-generated and non-deterministic across rebuilds (per `skill.md` Step 5), the same conceptual community is often named differently each time, leaving its previous file as an orphan. After N rebuilds, `wiki/` contains roughly N times the active article count, with `index.md` only referencing the most recent run's labels.

## Repro

```bash
graphify <path> --wiki              # generates ~100 articles + index
graphify <path> --update --wiki     # generates ~100 NEW articles (LLM picks different names)
ls graphify-out/wiki/*.md | wc -l   # ~200 files, index references ~100
# Repeat → unbounded growth
```

Real-world: a knowledge corpus accumulated **822 wiki .md files over 5 rebuilds**, of which only 111 were referenced by `index.md` (710 orphans).

## Fix

Clear `*.md` files in the output directory at the start of `to_wiki()`. This is consistent with its existing fully-regenerative behavior — it always writes the full set of articles + index, never partial updates. Subdirectories and non-`.md` files are preserved (only top-level `.md` is touched), so any user-placed auxiliary assets survive.

```python
out.mkdir(parents=True, exist_ok=True)

# NEW: prevent orphan accumulation
for old_article in out.glob("*.md"):
    old_article.unlink()
```

## Tests

Two new regression tests in `tests/test_wiki.py`:

1. **`test_to_wiki_clears_stale_articles`** — calls `to_wiki()` twice with different community labels, asserts old files are gone and new files exist.
2. **`test_to_wiki_preserves_non_md_files`** — places PNG/JSON/subdirectory content in the wiki dir, asserts they survive cleanup.

All 17 tests in `test_wiki.py` pass (15 existing + 2 new).

## Compatibility

- No CLI / API change
- No behavioral change for first-time wiki generation (empty dir → no `.md` files to clear)
- Custom `.md` files placed at top level of `wiki/` (not a documented workflow) would be removed on next `--wiki`. Subdirectories and non-`.md` files are unaffected.

## Related work

Existing wiki improvements addressed adjacent concerns but not orphan cleanup:
- #229 — wiki labels persistence (`.graphify_labels.json`)
- #382 — emit wiki on code-only rebuild
- #549 — post-commit hook with wiki rebuild

### DIFF
--- a/graphify/wiki.py
+++ b/graphify/wiki.py
@@ -185,6 +185,15 @@ def to_wiki(
     out = Path(output_dir)
     out.mkdir(parents=True, exist_ok=True)
 
+    # Clear stale .md files from previous runs to prevent orphan accumulation.
+    # Community labels are LLM-generated (per skill.md Step 5) and non-deterministic
+    # across runs — the same conceptual community may be named differently each time
+    # (e.g. "AutoAgent Skills" → "AutoAgent Methodology"), leaving the previous file
+    # as an orphan. Since to_wiki() owns wiki/ entirely (always writes the full set),
+    # it can safely clear .md files at the start of each call.
+    for old_article in out.glob("*.md"):
+        old_article.unlink()
+
     labels = community_labels or {cid: f"Community {cid}" for cid in communities}
     cohesion = cohesion or {}
     god_nodes_data = god_nodes_data or []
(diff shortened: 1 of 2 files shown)


safishamsi:

PR #561, opened 2026-04-26
refactor: configurable home dir + unified pathspec ignore

kobor42:
- New paths.py: GRAPHIFY_HOME-configurable home dir (default `.graphify`), with auto_migrate() called from every `graphify install` flow so an upgrading user's `graphify-out/` gets renamed in lockstep with the CLAUDE.md / AGENTS.md / hooks / skill rewrites.
- detect.py: built-in noise as a prepended GitIgnoreSpec; 5 parallel pruning systems (_SKIP_DIRS, _SKIP_FILES, _is_noise_dir, dotfile prune, user gitignore) collapse into 1 chain. `_is_ignored` tri-state -> plain bool via `spec.check_file()`.
- Drop hardcoded "graphify-out" everywhere except `migrate-home`: security.py, hooks.py, watch.py, transcribe.py, __main__.py CLI defaults, all skill template strings, the Codex/Gemini hook dicts.
- Bug fix: watch.py event filter no longer infinite-loops on the new home dir (was hardcoded to `graphify-out`).
- 329 tests pass (was 267). +pathspec>=0.12 dep.

### DIFF
--- a/graphify/__main__.py
+++ b/graphify/__main__.py
@@ -7,11 +7,32 @@ import shutil
 import sys
 from pathlib import Path
 
+from . import paths as _paths
+
 try:
     from importlib.metadata import version as _pkg_version
     __version__ = _pkg_version("graphifyy")
 except Exception:
     __version__ = "unknown"
+
+
+# Word-boundary substitution so install/print embeds the active GRAPHIFY_HOME
+# in every host-config template (settings.json, CLAUDE.md, AGENTS.md, skills, ...).
+_GRAPHIFY_OUT_TOKEN = re.compile(r"\bgraphify-out\b")
+
+
+def _render(template: str) -> str:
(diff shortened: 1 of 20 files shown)


kobor42:
@safishamsi I was annoyed by the graphity-out folder and the json files put in the root directory.
Also I am building a dotnet blazorwasm projekt which generates typescript files, which are analysed by your graphily - and gets confused by it.

So I made a PR that fixes both problem.
1. Every file is generated under .graphify/ folder - can be changed with config. Migration script included.
2. .gitignore files are taken into account, and .graphifyignore file behaviour is extended:
2.1. Regular .gitignore is taken as .graphifyignore
2.2. .graphifyignore now supports standard .gitignore syntax (negation, pattern matching, etc.)
2.3. .graphifyignore file overrides (or extends) .gitignore rules.

PR #583, opened 2026-04-28
feat(detect): add hidden path allowlist

Bichalla:
## Summary
- Add `.graphifyinclude` loading to `detect()`.
- Keep hidden paths skipped by default, but traverse allowlisted hidden paths needed for curated docs.
- Keep sensitive-looking filenames hard-skipped even when allowlisted.
- Add regression tests for `.hermes/plans/**/*.md` inclusion and sensitive allowlist hard-skip.

## Verification
- `uv run --with pytest python -m pytest tests/test_detect.py -q` → 30 passed
- `uv run --with pytest python -m pytest -q` → 443 passed

## Context
This supports SSOT-safe project operating docs: hidden operating folders like `.hermes/plans/` can remain canonical while Graphify indexes only explicitly curated Markdown paths.

### DIFF
--- a/graphify/detect.py
+++ b/graphify/detect.py
@@ -263,13 +263,13 @@ def _is_noise_dir(part: str) -> bool:
     return False
 
 
-def _load_graphifyignore(root: Path) -> list[tuple[Path, str]]:
-    """Read .graphifyignore from root **and ancestor directories**.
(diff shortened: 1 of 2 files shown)


safishamsi:

PR #586, opened 2026-04-28
Replace html2text (GPL-3.0) with markdownify (MIT)

mrummuka:
## Summary

Swaps the optional HTML→Markdown converter used by URL ingestion from `html2text` (GPL-3.0) to `markdownify` (MIT). Aligns the `pdf` and `all` extras with the project's MIT license and removes a copyleft dependency that affected anyone redistributing or embedding graphify.

## Why

`html2text` is the only GPL-licensed dependency in the project. Its presence in the `pdf`/`all` extras creates copyleft obligations that conflict with the project's MIT license and impact downstream redistribution.

## Changes

- **`graphify/ingest.py`**: `_html_to_markdown()` now uses `markdownify` with `heading_style=ATX`, `bullets='-'`, and `strip=['img']`. Script/style blocks are pre-stripped (with content) via regex because markdownify's `strip=` removes tags but keeps their inner text — preventing CSS/JS from leaking into the emitted markdown (small security improvement).
- **`pyproject.toml`**: `html2text` → `markdownify` in both `pdf` and `all` extras.
- **`graphify/skill.md`** and **`skills/graphify/skill.md`**: user-facing description updated.
- **`CHANGELOG.md`**: Unreleased entry added.

Behaviour preserved:
- ATX headings (`# Title`)
- Inline links kept
- Images dropped (matches prior `ignore_images=True`)
- No body-width wrapping
- Regex-strip fallback path retained when markdownify is unavailable

## Tests

Adds `tests/test_html_to_markdown.py` with 12 tests:
- 9 conversion tests (paragraphs, ATX headings, links, image removal, script/style stripping, bullet lists, no wrapping, empty input, malformed HTML)
- Fallback path test that forces `ImportError`
- End-to-end `_fetch_webpage` integration test
- Regression guard that fails if `html2text` reappears in shipped code or `pyproject.toml`

```
Baseline: 293 passed
After:    305 passed (293 + 12 new), 0 failed, 0 regressions
```

## License posture

| Dep | Before | After |
|---|---|---|
| HTML→MD converter | `html2text` (GPL-3.0) | `markdownify` (MIT) |

Closes the only copyleft dependency in the project.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,14 @@
 # Changelog
 
+## Unreleased
+
+### Changed
+- Replace `html2text` (GPL-3.0) with `markdownify` (MIT) for HTML→Markdown
+  conversion in URL ingestion. Aligns the `pdf`/`all` extras with the
+  project's MIT license and removes a copyleft dependency that affected
+  anyone redistributing or embedding graphify. Fallback regex-strip path
+  unchanged.
+
 Full release notes with details on each version: [GitHub Releases](https://github.com/safishamsi/graphify/releases)
 
 ## 0.4.23 (2026-04-18)
--- a/graphify/ingest.py
+++ b/graphify/ingest.py
@@ -49,19 +49,21 @@ def _fetch_html(url: str) -> str:
 
 
 def _html_to_markdown(html: str, url: str) -> str:
-    """Convert HTML to clean markdown. Uses html2text if available, else basic strip."""
+    """Convert HTML to clean markdown. Uses markdownify if available, else basic strip."""
+    # Drop script/style content (markdownify's `strip` removes tags but keeps inner text).
+    cleaned = re.sub(r"<script[^>]*>.*?</script>", "", html, flags=re.DOTALL | re.IGNORECASE)
+    cleaned = re.sub(r"<style[^>]*>.*?</style>", "", cleaned, flags=re.DOTALL | re.IGNORECASE)
     try:
-        import html2text
-        h = html2text.HTML2Text()
-        h.ignore_links = False
-        h.ignore_images = True
-        h.body_width = 0
-        return h.handle(html)
+        from markdownify import markdownify as _md
+        return _md(
+            cleaned,
+            heading_style="ATX",
+            bullets="-",
+            strip=["img"],
+        ).strip()
     except ImportError:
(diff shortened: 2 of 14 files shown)


Qodo-Free-For-OSS:
Hi, tests/test_html_to_markdown.py asserts markdownify-specific output even when markdownify is not installed, so the test suite fails on a base install before exercising the intended regex-strip fallback behavior.

**How to fix:** Gate tests on markdownify

> ### Issue description
> The new tests in `tests/test_html_to_markdown.py` assume `markdownify` is installed and assert markdown-specific output. However, `markdownify` is only an optional dependency (`pdf`/`all` extras). When it isn’t installed, `_html_to_markdown()` falls back to regex stripping and returns plain text, causing these tests to fail on a base install.
> 
> ### Issue Context
> CI currently installs `-e ".[mcp,pdf,watch]"`, so CI passes, but developers/packagers running `pip install -e .` (or consumers running tests without extras) will see failures.
> 
> ### Fix Focus Areas
> Choose one:
> - Make `markdownify` a required dependency, OR
> - Mark markdownify-dependent tests with a conditional skip (e.g., `pytest.importorskip("markdownify")`) and keep only fallback assertions unconditionally, OR
> - Split tests into two groups: converter-present vs fallback.
> 
> - tests/test_html_to_markdown.py[13-69]
> - pyproject.toml[44-53]
> - .github/workflows/ci.yml[24-31]

---

mrummuka:
Thanks — fixed in b48a0a5. Used the skip-marker option: gated the 5 markdownify-shape assertions behind @requires_markdownify (using importlib.util.find_spec for a side-effect-free check), kept all fallback/regression tests unconditional, and added a test_fallback_basic_text_extraction smoke test so the regex-strip path gets end-to-end coverage on base installs.

Verified:

pip install -e ".[pdf]": 454 passed (full suite)
pip install -e .: 449 passed, 5 skipped, 0 failed (full suite)
Yes, please share the other issues — happy to address them in this PR.

safishamsi:
The requested changes have been implemented in 3fdae8f. `html2text` (GPL-3.0) replaced with `markdownify` (MIT) in both `ingest.py` and `pyproject.toml`. Script/style blocks pre-stripped before conversion. Shipped in v0.6.2.

PR #589, opened 2026-04-28
fix: cache.save_cached race + _clone_repo argument-injection hardening

Theprofitplatform:
Two small, independent hardening fixes — each in its own commit with a regression test.

## 1. cache.save_cached: per-writer unique tempfile

`save_cached` used `entry.with_suffix(\".tmp\")` for the staging path, which is a single shared filename per hash. Concurrent writers for the same hash (parallel extractors, retries) raced on that file — one writer's partial bytes could land in another's atomic \`os.replace\`, leaving a truncated/mixed cache entry on disk.

Switched to \`tempfile.mkstemp(dir=target_dir, prefix=f\"{h}.\", suffix=\".tmp\")\` so every writer gets its own staging file. The final \`os.replace\` is still atomic; only the final rename contends, and last-writer-wins is fine because two writers with the same hash are writing the same content.

Test: \`test_save_cached_concurrent_writers_no_collision\` fans out 8 threads writing the same hash, asserts the final entry is parseable, no orphan \`.tmp\` files remain, and no writer raised.

## 2. _clone_repo: argument-injection hardening

Two changes to the git invocation:

- Reject \`branch\` names starting with \`-\`. git can treat refspecs starting with \`-\` as options in some positions, so a value like \`--upload-pack=evilcmd\` passed via \`--branch\` could be reinterpreted by git for the next argument. Now fails fast with a clear error.
- Add a \`--\` separator before positional URL/dest arguments. The \`git_url\` is currently regex-locked to github.com URLs so immediate exposure is small, but the separator makes it impossible for future changes to the URL-building code to leak an option through this surface. Same guard added to the \`git pull origin <branch>\` path.

Test: \`tests/test_clone.py\`
- \`test_clone_rejects_branch_starting_with_dash\` — \`SystemExit\` raised before any \`subprocess.run\`
- \`test_clone_argv_uses_double_dash_separator\` — constructed argv contains a literal \`--\` before positional URL/dest

## Out of scope (filing separately)

These came up in the same review pass but need design discussion, not a 5-fix PR:

- DNS rebinding TOCTOU in \`security.validate_url\` / \`safe_fetch\` (resolves twice, attacker DNS can return public IP first and metadata IP second).
- \`yt-dlp\` SSRF bypass in \`transcribe.download_audio\` — the URL never goes through \`validate_url\`.

I'll open these as separate issues so they can be discussed independently.

## Test run

\`\`\`
$ pytest tests/test_cache.py tests/test_clone.py
12 passed (1 pre-existing failure in test_clear_cache that exists on v5 HEAD without my changes, due to cache_dir now returning .../cache/{kind}/)
\`\`\`

Co-Authored-By: Claude Opus 4.7 (1M context) <noreply@anthropic.com>

### DIFF
--- a/graphify/__main__.py
+++ b/graphify/__main__.py
@@ -943,6 +943,13 @@ def _clone_repo(url: str, branch: str | None = None, out_dir: Path | None = None
         sys.exit(1)
     owner, repo = m.group(1), m.group(2)
 
+    # Reject branch names that look like CLI flags. git treats refspecs starting
+    # with `-` as options in some positions, which would let a caller smuggle
+    # arguments through this surface (e.g. `--upload-pack=evilcmd`).
+    if branch is not None and branch.startswith("-"):
+        print(f"error: refusing branch name starting with '-': {branch!r}", file=sys.stderr)
+        sys.exit(1)
+
     if out_dir:
         dest = out_dir
     else:
@@ -952,7 +959,7 @@ def _clone_repo(url: str, branch: str | None = None, out_dir: Path | None = None
         print(f"Repo already cloned at {dest} — pulling latest...", flush=True)
         cmd = ["git", "-C", str(dest), "pull"]
         if branch:
-            cmd += ["origin", branch]
+            cmd += ["--", "origin", branch]
         result = _sp.run(cmd, capture_output=True, text=True)
         if result.returncode != 0:
             print(f"warning: git pull failed:\n{result.stderr}", file=sys.stderr)
@@ -962,7 +969,7 @@ def _clone_repo(url: str, branch: str | None = None, out_dir: Path | None = None
         cmd = ["git", "clone", "--depth", "1"]
         if branch:
             cmd += ["--branch", branch]
-        cmd += [git_url, str(dest)]
+        cmd += ["--", git_url, str(dest)]
         result = _sp.run(cmd, capture_output=True, text=True)
         if result.returncode != 0:
             print(f"error: git clone failed:\n{result.stderr}", file=sys.stderr)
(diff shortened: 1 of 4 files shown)


Qodo-Free-For-OSS:
Hi, `save_cached()` creates a raw file descriptor via `tempfile.mkstemp()` and then wraps it with `os.fdopen()`, but if `os.fdopen()` raises before owning the fd, the descriptor is leaked and the temp file may become undeletable (notably on Windows). This is a production-path resource leak that can accumulate in long-running runs or leave orphan `.tmp` files.

**How to fix:** Ensure fd closed on failure

> ### Issue description
> `graphify/cache.py::save_cached` uses `tempfile.mkstemp()` + `os.fdopen()`. If `os.fdopen()` raises before taking ownership of the fd, the fd leaks and the temp file may be left behind / undeletable.
> 
> ### Issue Context
> `mkstemp()` returns an already-open OS fd. It must be closed even when `fdopen()` fails.
> 
> ### Fix Focus Areas
> - graphify/cache.py[102-123]
> 
> ### Suggested fix
> Refactor to a pattern that guarantees closure:
> - Option A (preferred): use `tempfile.NamedTemporaryFile(mode="w", encoding="utf-8", dir=..., prefix=..., suffix=..., delete=False)` and write via the file object; capture `tmp = Path(f.name)`; close occurs via the context manager.
> - Option B: wrap `os.fdopen(fd, ...)` in an inner try/except and `os.close(fd)` if `fdopen` fails, then re-raise.
> 
> Keep the existing cleanup logic (`tmp.unlink(missing_ok=True)` on exceptions) and the atomic `os.replace` behavior.

---
*Found by [Qodo](https://github.com/marketplace/qodo-merge-pro-for-open-source). Free code review for open-source maintainers.*

safishamsi:
The requested changes have been implemented in 3fdae8f. Cache writer now uses `tempfile.mkstemp` for a unique tmp file per writer eliminating the race condition. Branch names starting with `-` are rejected and `--` separator added before positional git args. Shipped in v0.6.2.

PR #594, opened 2026-04-28
wiki: sanitize Windows-reserved characters in filenames

cod3warrior:
## Problem

On Windows, `to_wiki()` crashes with `OSError: [Errno 22] Invalid argument` when a node label contains any of `< > : " / \ | ? *` — these are characters Windows reserves in filenames.

Concrete repro: an Obsidian note titled `Czy Wii jeszcze ok?.md` becomes a community label, then `write_text()` fails because Windows rejects the question mark in the path.

The previous `_safe_filename` only handled three characters (`/`, space, `:`) which is enough on Linux/macOS but not on Windows. It also doesn't strip trailing dots or spaces, which Windows also reserves at the end of filename segments.

## Fix

Extended substitution to cover all Windows-reserved characters plus trailing dot/space, and added a 200-char length cap to stay well under MAX_PATH-segment limits.

```python
def _safe_filename(name: str) -> str:
    import re
    s = name.replace("/", "-").replace(" ", "_").replace(":", "-")
    s = re.sub(r'[<>:"/\\|?*]', '_', s)
    s = s.strip('. ')
    return s[:200] if s else 'unnamed'
```

Falls back to `'unnamed'` if sanitization produces an empty string, so we never try to create a file named `""` or `"."`.

## Test scenario

```python
# Vault with edge-case labels
import os
os.makedirs('test_vault', exist_ok=True)
open('test_vault/Czy Wii ok?.md', 'w', encoding='utf-8').write('Test')
open('test_vault/Some : ratio.md', 'w', encoding='utf-8').write('Test')
open('test_vault/Trailing dot.md.', 'w', encoding='utf-8').write('Test')

# Before this PR: OSError on Windows
# After this PR: clean wiki generation on all platforms
# graphify ./test_vault --wiki
```

## Verified on

A 1773-file Obsidian vault with mixed Polish/English content where ~12 community labels and several god node labels contained reserved characters. Generates 763 wiki articles cleanly on Windows 11.

### DIFF
--- a/graphify/wiki.py
+++ b/graphify/wiki.py
@@ -7,7 +7,18 @@ import networkx as nx
 
 
 def _safe_filename(name: str) -> str:
-    return name.replace("/", "-").replace(" ", "_").replace(":", "-")
+    """Make a label safe for use as a filename across platforms.
+
+    Substitutes characters that Windows reserves in filenames
+    (< > : " / \\ | ? *) and strips trailing dots/spaces, also reserved.
+    Falls back to 'unnamed' for empty results and caps length at 200
+    chars to stay well under common filesystem limits.
+    """
+    import re
+    s = name.replace("/", "-").replace(" ", "_").replace(":", "-")
+    s = re.sub(r'[<>:"/\\|?*]', '_', s)
+    s = s.strip('. ')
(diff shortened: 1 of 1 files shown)


safishamsi:

PR #604, opened 2026-04-29
fix: add 'concept' to VALID_FILE_TYPES

Arun96980:


Arun96980:
When running `graphify update` on projects with Markdown documentation, the semantic extractor often produces nodes with `file_type: "concept"`. However, `validate.py` currently only allows `{"code", "document", "paper", "image", "rationale"}`.



This PR adds `"concept"` to the `VALID_FILE_TYPES` set in `graphify/validate.py` to resolve these warnings.

PR #605, opened 2026-04-29
fix(detect): strip inline comments from .graphifyignore patterns

LEELADEVS:
## Summary

`_load_graphifyignore` only handled full-line comments. Inline comments like `vendor/  # legacy code` were loaded as a single pattern including the trailing whitespace and comment text, which `fnmatch` could not match against any real path → the entry silently did not ignore anything.

## Real-world impact

In our knowledge corpus (~640 files, gitignore-style `.graphifyignore` with inline-documented patterns), 52 + 12 files in two categories were marked as changed on every `--update` run because their patterns had inline comments. This caused unnecessary LLM re-extraction (~\$1-3 per run) and noisy diff output.

The fix lets users document patterns inline (a common gitignore convention even though git itself does not support it natively):

```
chitta/varta/    # daily CEO briefings, regenerated nightly
chitta/spanda/   # webhook event cache
```

## Behavior

| Input line | Before | After |
|---|---|---|
| `vendor/  # legacy code` | pattern = `vendor/  # legacy code` (matches nothing) | pattern = `vendor/` (matches `vendor/...`) |
| `*.log` (no inline comment) | pattern = `*.log` | pattern = `*.log` (unchanged) |
| `# full-line comment` | skipped | skipped (unchanged) |
| `path#with#hash.py` (no whitespace before `#`) | pattern = `path#with#hash.py` | pattern = `path#with#hash.py` (unchanged — gitignore semantics) |
| `file \# name.py` (escaped hash) | pattern = `file \# name.py` | pattern = `file # name.py` (literal `#` preserved) |

## Tests

Three new tests added to `tests/test_detect.py`:
- `test_graphifyignore_inline_comments_stripped` — main case
- `test_graphifyignore_inline_comment_requires_whitespace_before_hash` — `#` without whitespace stays literal
- `test_graphifyignore_escaped_hash_is_literal` — `\#` escape

All 9 `.graphifyignore` tests pass.

## Notes

- Backward compatible: patterns without inline comments behave identically.
- Docstring updated to describe the inline-comment + escape semantics as a documented gitignore extension.
- The whitespace-before-`#` requirement matches user intuition while preserving the rare case of `#` as a literal filename character.

### DIFF
--- a/graphify/detect.py
+++ b/graphify/detect.py
@@ -272,17 +272,41 @@ def _load_graphifyignore(root: Path) -> list[tuple[Path, str]]:
     parent directory still work when graphify is run on a subfolder.
 
     Walks upward from *root* towards the filesystem root, stopping at a
-    ``.git`` boundary. Lines starting with # are comments; blank lines ignored.
+    ``.git`` boundary.
+
+    Comment handling (gitignore extension):
+      * Lines starting with # are full-line comments — skipped.
+      * Inline comments (whitespace + hash to end of line) are stripped
+        from each pattern. This matches user intuition for documenting
+        patterns inline, e.g. `chitta/varta/  # daily briefings`.
+      * Use ``\\#`` to keep a literal hash in a pattern (rare).
+      * Blank lines after stripping are ignored.
     """
     patterns: list[tuple[Path, str]] = []
     current = root.resolve()
     while True:
         ignore_file = current / ".graphifyignore"
         if ignore_file.exists():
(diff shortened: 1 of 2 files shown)


safishamsi:
The requested changes have been implemented in 3fdae8f. `_parse_gitignore_line` now strips inline comments (whitespace + `#` suffix) while preserving `path#hash.py` literals and `\#` escapes. Shipped in v0.6.2.

Issue #533, opened 2026-04-24
codex not support additionalContext on PreToolUse


inkpark:
PreToolUse hook (failed)
  error: PreToolUse hook returned unsupported additionalContext

wlb632363:
same problem
graphifyy：0.5.0
codex-cli： 0.124.0

<img width="634" height="145" alt="Image" src="https://github.com/user-attachments/assets/026da9e6-74ed-43cc-ae41-d397da896c3c" />

bratnieks:
## Temporary fix for Codex CLI `unsupported additionalContext`

  If Codex CLI shows:

  ```text
  PreToolUse hook (failed)
  error: PreToolUse hook returned unsupported additionalContext

  Ask Codex to run this:

  Patch my local Graphify installation for the Codex CLI hook error.

  Replace all Graphify hook outputs using `additionalContext` with `systemMessage`.

  Run:

  which graphify || true
  python3 -c "import graphify; print(graphify.__file__)" 2>/dev/null || true
  .venv/bin/python -c "import graphify; print(graphify.__file__)" 2>/dev/null || true

  rg -n 'additionalContext|hookSpecificOutput|systemMessage' \
    ~/.local/lib/python*/site-packages/graphify/__main__.py \
    .venv/lib/python*/site-packages/graphify/__main__.py \
    2>/dev/null

  perl -0pi -e 's/additionalContext/systemMessage/g; s/"hookSpecificOutput":
  \{"hookEventName":"PreToolUse","systemMessage":/"systemMessage":/g; s/\}\}//g' \
    ~/.local/lib/python*/site-packages/graphify/__main__.py \
    .venv/lib/python*/site-packages/graphify/__main__.py

  graphify codex install || true
  .venv/bin/python -m graphify codex install || true

  rg -n 'additionalContext|hookSpecificOutput' \
    .codex/hooks.json \
    ~/.local/lib/python*/site-packages/graphify/__main__.py \
    .venv/lib/python*/site-packages/graphify/__main__.py \
    2>/dev/null

  pwd

  Expected final check:

  no matches for additionalContext or hookSpecificOutput

  Then restart Codex CLI.

takeItIzzy:
I can reproduce this on macOS with the current Codex hook behavior.

Environment:
- graphifyy: 0.4.32 (`graphify --version` is not supported, checked via Python package metadata)
- codex-cli: 0.125.0
- OS: macOS
- shell: zsh

Generated project hook:

```json
{
[...]

PR #614, opened 2026-04-29
Use UserPromptSubmit for Codex graph reminders

robertmonka:
## Summary
- switch Codex graph reminders from `PreToolUse` to `UserPromptSubmit`
- remove old Graphify Codex `PreToolUse` hooks during install/uninstall while preserving unrelated hooks
- update README wording for the Codex hook behavior
- add tests for the generated Codex hook and upgrade cleanup path

## Why
Codex rejects `additionalContext` from `PreToolUse` hooks, which caused the failure reported in #533.

For Graphify's graph reminder, `UserPromptSubmit` is also a better fit than `PreToolUse`: the assistant should see the `GRAPH_REPORT.md` reminder before it decides whether to search files, not after it has already selected a tool.

## Tests
- `python -m py_compile graphify/__main__.py tests/test_install.py tests/test_claude_md.py`
- `uv run --with pytest pytest tests/ -q --tb=short` (`443 passed`)

Addresses #533.


Qodo-Free-For-OSS:
Hi, _remove_graphify_codex_hooks() removes any Codex hook whose stringified JSON contains "graphify", which can delete unrelated user hooks that merely mention graphify. It also replaces a non-dict "hooks" value with an empty dict, potentially discarding user configuration when hooks.json is malformed but still recoverable.

**Severity:** remediation recommended | **Category:** correctness

**How to fix:** Match hooks structurally

> ### Issue description
> `_remove_graphify_codex_hooks()` currently deletes hooks by checking whether `'graphify'` appears in `str(h)`. This is too broad and can remove unrelated user hooks.
> 
> ### Issue Context
> Graphify should remove only hooks that it previously installed, while preserving unrelated hooks even if they contain the word `graphify`.
> 
> ### Fix Focus Areas
> - graphify/__main__.py[733-748]
>   - Replace substring filtering with a structural predicate that matches only graphify-installed hooks.
>   - Suggested predicate: match on hook dict shape and exact `command` content (e.g., contains `[ -f graphify-out/graph.json ] &&` and `graphify-out/GRAPH_REPORT.md`, and/or a unique marker string), and/or match the exact hook object you generate in `_CODEX_HOOK`.
>   - Avoid overwriting `existing["hooks"]` when it is not a dict; prefer leaving it untouched or normalizing only when you can do so without data loss.
> - tests/test_install.py[147-185]
>   - Add a regression test where an unrelated hook includes the string "graphify" but should *not* be removed (e.g., `command: "echo graphify for my own plugin"`).

---

PR #623, opened 2026-04-30
fix: disable thinking on Moonshot calls so kimi-k2.6 populates content

jasonm4130:
## Summary

- Disable Moonshot's `thinking` mode for kimi-k2.6 by passing `extra_body={"thinking": {"type": "disabled"}}` on the OpenAI-compatible call
- Gated on `"moonshot" in base_url` so other OpenAI-compatible providers (real OpenAI, Together, Groq) aren't affected — they'd 400 on the flag
- Add 2 mock-based tests covering the moonshot path and the non-moonshot path

## Why

`kimi-k2.6` is a reasoning model. Its OpenAI-compatible response splits across two fields:

```python
resp.choices[0].message.content            # the structured answer
resp.choices[0].message.reasoning_content  # chain-of-thought
```

With `max_completion_tokens=8192`, large chunks (the default `chunk_size=20` regularly produces 50k+ token user messages) cause reasoning to consume the entire output budget. `content` comes back empty, `_parse_llm_json("{}")` returns `{"nodes":[],...}`, and the chunk silently contributes nothing.

## Reproduction

162-file repo (~125k words, mixed code/docs/images), default settings:

| | input tokens | output tokens | nodes | edges |
|---|---|---|---|---|
| Before | 453,282 | 71,361 | 2 | 1 |
| After  | 446,890 | 40,523 | 138 | 160 |

The 71k output tokens before were almost entirely `reasoning_content` that graphify ignores. Empty `content`, empty graph, full bill.

## Test plan

- [x] `test_kimi_call_disables_thinking` — verifies the flag is sent on a moonshot URL
- [x] `test_non_moonshot_call_does_not_disable_thinking` — verifies the flag is not sent on `api.openai.com`
- [x] Full existing test suite (443 tests) passes with zero regressions

PR #625, opened 2026-04-30
feat(llm): pack chunks by token budget, parallelise, retry on truncation

jasonm4130:
Two commits, both improvements to `extract_corpus_parallel`. Reviewable independently.

## Summary

**Commit 1: token-budget chunking, parallelism, optional tiktoken**

- Replace `chunk_size=20` static packing with greedy `_pack_chunks_by_tokens(token_budget=60_000)`, grouped by parent directory
- Add `tiktoken` to the `[kimi]` extra; `_estimate_file_tokens` uses `cl100k_base` when available, falls back to chars/4 when not
- Run chunks via `ThreadPoolExecutor` capped at `max_concurrency=4`. `on_chunk_done(idx, total, result)` fires in completion order with the original submission `idx` so progress UIs work unchanged. `max_concurrency=1` skips the pool to preserve sequential semantics
- Catch per-chunk exceptions, log to stderr, continue. One bad chunk no longer aborts the run
- `token_budget=None` falls back to legacy `chunk_size`-based packing for backwards compatibility

**Commit 2: adaptive retry on `finish_reason == "length"`**

- Plumb `finish_reason` out of `_call_openai_compat` and `_call_claude` (Anthropic's `stop_reason == "max_tokens"` is normalised to `"length"`)
- Add `_extract_with_adaptive_retry`: when a chunk's response is truncated, split in half and recurse on each half. Recursion bounded by `max_retry_depth` (default 3)
- Single-file chunks that truncate can't recover — surface a warning rather than infinite-loop
- `extract_corpus_parallel` routes every chunk through the retry wrapper; recursive splits are invisible to callers (callback still fires once per top-level chunk with merged result)

## Why

`extract_corpus_parallel` had three issues that compounded on real corpora:

| # | Issue | Concrete failure |
|---|---|---|
| 1 | `chunk_size=20` static packing has unbounded per-chunk cost | A 162-file mixed code/docs/images repo (~125k words) packed unevenly. One PNG-heavy chunk hit **282k input tokens** and got 400'd by Moonshot's 262k context limit |
| 2 | Function name says "parallel" but body is a sequential `for` loop | Same 162-file repo took **~36 minutes wall-clock** |
| 3 | A single chunk raising aborts the whole run | Lose all preceding chunks' work to one transient API error |

After fixing 1-3, a fourth issue surfaces: chunks too dense to fit their JSON output in `max_completion_tokens=8192` are silently truncated and contribute nothing. Adding a hard `max_files_per_chunk` cap reintroduces the "tune a static constant" problem the chunking commit set out to fix. The `finish_reason` signal is what the API gives us — acting on it is the principled fix.

## Test plan

- [x] Packer: small-file packing into one chunk
- [x] Packer: starts new chunk when next file would exceed budget
- [x] Packer: groups files from same directory contiguously
- [x] Packer: oversized single file gets its own chunk
- [x] Packer: rejects non-positive budget
- [x] Tokenizer: uses tiktoken when available (mocked)
- [x] Tokenizer: falls back to chars/4 when tiktoken absent
- [x] Parallel: 4 chunks × 0.3s sleep finishes in <1s with `max_concurrency=4`
- [x] Sequential: `max_concurrency=1` preserves call order
- [x] Resilience: simulated chunk failure logs to stderr, other chunks still merge
- [x] Legacy: `token_budget=None` reverts to fixed-count chunking (45 files / `chunk_size=20` = `[20, 20, 5]`)
- [x] Token-budget default: 50 tiny files pack into 1 chunk
- [x] Adaptive retry: pass-through when `finish_reason="stop"`
- [x] Adaptive retry: single-level split on `finish_reason="length"`
- [x] Adaptive retry: recursive split when halves are still truncated (8 → 4+4 → 2+2+2+2)
- [x] Adaptive retry: `max_depth` bounds recursion (no infinite loop)
- [x] Adaptive retry: single-file truncation surfaces warning instead of recursing
- [x] Adaptive retry: integration with `extract_corpus_parallel` — `on_chunk_done` fires once per top-level chunk
- [x] Full existing test suite (459 tests total) passes with zero regressions

## Companion

#623 — kimi-k2.6 reasoning fix. Independent, can land in either order.

### DIFF
--- a/graphify/llm.py
+++ b/graphify/llm.py
@@ -9,7 +9,39 @@ import os
 import sys
 import time
 from collections.abc import Callable
+from concurrent.futures import ThreadPoolExecutor, as_completed
 from pathlib import Path
+
+# `_read_files` truncates each file at this many characters before joining into
+# the user message. Token estimates use the same cap so packing matches reality.
+_FILE_CHAR_CAP = 20_000
+# `_read_files` also wraps each file in a `=== {rel} ===\n...\n\n` separator;
+# this is roughly the per-file overhead in characters that the prompt adds.
+_PER_FILE_OVERHEAD_CHARS = 80
+# Coarse fallback used only when `tiktoken` is not installed. 1 token ≈ 4 chars
+# is the standard heuristic for English/code on BPE tokenizers.
+_CHARS_PER_TOKEN = 4
+
+
+def _get_tokenizer():
+    """Return a tiktoken encoder for accurate token counts, or None if tiktoken
+    is not installed. We use `cl100k_base` (GPT-4 / GPT-3.5-turbo) as a proxy:
+    Kimi-K2 ships a tiktoken-based tokenizer with very similar BPE behaviour,
+    and Claude's tokenizer has a comparable token-to-char ratio for prose/code.
+    Estimates only need to be within ~5%, not exact.
+    """
+    try:
+        import tiktoken
+    except ImportError:
+        return None
+    try:
+        return tiktoken.get_encoding("cl100k_base")
+    except Exception:  # network failure on first-use download, etc.
+        return None
+
+
+# Cached at import time. None if tiktoken is unavailable; consumers must handle.
+_TOKENIZER = _get_tokenizer()
 
 BACKENDS: dict[str, dict] = {
     "claude": {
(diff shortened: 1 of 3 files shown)


Qodo-Free-For-OSS:
Hi, extract_corpus_parallel() logs per-chunk exceptions and skips them, but the returned merged dict provides no structured indication that the run was partial. Callers cannot reliably detect missing chunks without scraping stderr.

**Severity:** remediation recommended | **Category:** reliability

**How to fix:** Return failure metadata to caller

> ### Issue description
> `extract_corpus_parallel()` continues after chunk errors but does not expose failures in the returned value.
> 
> ### Issue Context
> This is a behavioral change from aborting-on-error to best-effort. Best-effort is fine, but callers need a programmatic signal that results may be incomplete.
> 
> ### Fix Focus Areas
> - graphify/llm.py[349-446]
> 
> ### Recommended change
> Add structured failure reporting, e.g.:
> - Maintain `failed_chunks: list[dict]` with `{idx, error}` (and maybe `chunk_files`), and include it in the returned dict.
> - Optionally add a `fail_fast: bool = False` parameter to restore old semantics when desired.
> - Consider incrementing/returning `chunks_succeeded`/`chunks_failed` counts for easy UI reporting.

---
*Found by [Qodo](https://github.com/marketplace/qodo-merge-pro-for-open-source). Free code review for open-source maintainers.*

Qodo-Free-For-OSS:
Hi, When tiktoken is available, token-budget packing reads and tokenizes each file, and extraction then reads the same files again to build the prompt. This increases I/O and can noticeably slow startup on large corpora.

**Severity:** informational | **Category:** performance

**How to fix:** Cache content during estimation

> ### Issue description
> With tiktoken installed, files are read once to estimate tokens and again to build the prompt.
> 
> ### Issue Context
> This can add significant overhead for large corpora.
> 
> ### Fix Focus Areas
> - graphify/llm.py[208-269]
> - graphify/llm.py[80-93]
> 
> ### Suggested approaches
> - Cache per-file truncated content during packing (e.g., `dict[Path, str]`) and allow `extract_files_direct()` / `_read_files()` to accept pre-read content.
> - Or, estimate using `st_size` only as a fast heuristic even when tiktoken exists (but then accept less accurate packing).
> - Or, add a flag to disable tiktoken-based estimation when throughput matters.

---

jasonm4130:
Closing this — it targets the pre-1.0 `llm.py`, which has since been substantially rewritten (multi-backend extraction + content-hash per-file cache in `cache.py`). The token-budget packing / parallelisation / truncation-retry here no longer applies cleanly to the current code, and the incremental-update path it was meant to support has been redesigned upstream. Thanks for the project!

PR #635, opened 2026-04-30
v5: fix CI gap, silent failures, and fake success states

toplevelsolutions:
## Summary

Repo audit on `v5` surfaced a CI gap, several silent-failure paths, and a few APIs whose docstrings overpromised what the implementation delivered. This PR addresses 9 of them in atomic commits.

## What changed

| # | Fix | Why |
|---|---|---|
| 1 | CI triggers on `v5` + adds `dashboard`/`embeddings` extras | CI was listed for v1..v4+main only — no tests ran on `v5`. Dashboard/embedding tests also need `fastapi`/`numpy` installed. |
| 2 | Dashboard `assert _G is not None` → `HTTPException` | Asserts are stripped under `python -O`, leaving handlers to AttributeError on `None`. |
| 3 | Cypher direction enforced via `_src`/`_tgt` | Parser never recorded `->`; execution iterated NetworkX canonical order, so directional MATCH silently returned back-edges. |
| 4 | Embeddings: persist BoW vocab + drop zero-norm queries | BoW search rebuilt vocab per call (dim mismatch crashed search), and even when shapes lined up, no-overlap queries returned arbitrary top-k with `score=0.0`. |
| 5 | Plugin auto-discovery gated by `GRAPHIFY_ENABLE_PLUGINS=1` | `~/.graphify/plugins/*.py` was auto-`exec_module`'d on every run — code-execution sink, especially with the post-commit hook installed. |
| 6 | `.gitignore` `graph.db` and `_debug_ts.py` | Both leaked into the working tree as untracked dev artifacts. |
| 7 | `SQLiteStore` per-thread connections via `threading.local` | Cached one connection with `check_same_thread=False` and no lock — class doc claimed concurrent-safe, it wasn't. |
| 8 | `_diff_communities` prefers stored community attrs | Re-running Leiden/Louvain made `splits/merges` non-deterministic; two diffs of the same pair produced different reports. |
| 9 | Dashboard caches `EmbeddingIndex` across `/api/search` | Each request rebuilt the index — with sentence-transformers installed, that meant reloading the model every search. |

## Test plan

- [x] `pytest tests/ --ignore=tests/test_install.py` → 529 passed, 5 skipped
- [x] New regression tests added for items 3, 4, 5, 7, 8, 9 (direction skip, BoW no-overlap, plugin gate, threaded reads, stored-community diff, cached index)
- [x] Verified item 2 under `python -O` — handlers raise 503 instead of stripped-assert AttributeError
- [x] Verified item 4 BoW path explicitly with `_load_model` stubbed to None
- [ ] CI green on the new `v5` trigger

## Notes

Out of scope but flagged in the audit and worth a follow-up:
- `SQLiteStore.search_nodes` claims FTS but uses `LIKE '%q%'` — needs an FTS5 virtual table.
- `extract_incremental` has dead delta logic (`to_extract` computed and unused).
- `plugins.load_plugins` caches the registry forever; new plugin files dropped mid-session aren't picked up until process restart.
- `compress.expand_super_node` returns a lossy graph without signaling it.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -2,9 +2,9 @@ name: CI
 
 on:
   push:
-    branches: ["v1", "v2", "v3", "v4", "main"]
+    branches: ["v1", "v2", "v3", "v4", "v5", "main"]
   pull_request:
-    branches: ["v1", "v2", "v3", "v4", "main"]
+    branches: ["v1", "v2", "v3", "v4", "v5", "main"]
 
 jobs:
   test:
@@ -23,7 +23,7 @@ jobs:
 
       - name: Install dependencies
         run: |
-          pip install -e ".[mcp,pdf,watch]"
+          pip install -e ".[mcp,pdf,watch,dashboard,embeddings]"
           pip install pytest
 
       - name: Run tests
--- a/.gitignore
+++ b/.gitignore
@@ -27,3 +27,7 @@ scripts/run_k2_*.py
 scripts/llm.py
 scripts/benchmark_kimi*.json
 scripts/benchmark_kimi*.py
+
+# Dev artifacts that have leaked into the repo before
+graph.db
+_debug_ts.py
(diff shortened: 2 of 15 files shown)


toplevelsolutions:
Wrong target — closing. These commits should land in toplevelsolutions/Fathomark (private), not upstream graphify.

PR #637, opened 2026-05-01
filter cross-language INFERRED calls in Surprising Connections (#630)

### LINKED ISSUE #630: GRAPH_REPORT: filter or downgrade cross-language INFERRED calls in Surprising Connections
﻿## Summary

`GRAPH_REPORT.md` can promote impossible cross-language `INFERRED calls` edges in the “Surprising Connections” section. In mixed AST-only corpora, name-collision edges such as Python helper → TypeScript route handler are often false positives, but `analyze.py` currently ranks `INFERRED` edges as more surprising and does not check whether the source/target language pair is plausible.

## Current behavior

In `analyze.py`, cross-file surprises skip structural edges, then include any cross-file non-structural edge:

```py
# analyze.py:204-219
for u, v, data in G.edges(data=True):
    relation = data.get("relation", "")
    if relation in ("imports", "imports_from", "contains", "method"):
        continue
    ...
    if not u_source or not v_source or u_source == v_source:
        continue
    score, reasons = _surprise_score(...)
```

The scoring function gives uncertain edges a bonus:

```py
conf_bonus = {"AMBIGUOUS": 3, "INFERRED": 2, "EXTRACTED": 1}.get(conf, 1)
```

So an impossible inferred `calls` edge can be promoted above mundane real edges simply because it is inferred and cross-file.

## Reproducer pattern

A mixed repo with Python scripts and TypeScript application code has common names on both sides:

```py
# scripts/build.py
def normalize(): ...
```

```ts
// src/app/api/report/route.ts
export async function GET() { ... }
function normalize(input: string) { ... }
```

If cross-file resolver pollution creates an `INFERRED calls` edge between the Python node and a TypeScript node, `GRAPH_REPORT.md` can list it under “Surprising Connections” even though there is no import/runtime bridge between those languages in the corpus.

A downstream wrapper had to post-process `GRAPH_REPORT.md` to remove Python→TypeScript `INFERRED calls` entries from “Surprising Connections”; otherwise the report repeatedly surfaced false architecture questions.

## Expected behavior

For `relation == "calls"` and `confidence == "INFERRED"`, cross-language edges should be filtered or heavily downgraded unless there is explicit bridge evidence.

Examples of evidence that could keep an edge:

- an import/interop edge showing the languages are connected;
- generated bindings or FFI files in the same path cluster;
- a known framework bridge configured by the user.

Without evidence, a Python→TypeScript or TypeScript→Python inferred call is more likely resolver pollution than a useful surprise.

## Suggested fix

Add language-aware filtering or scoring in `_cross_file_surprises()` / `_surprise_score()`:

- derive language from `source_file` extension;
- for `INFERRED calls`, treat cross-language pairs as suspicious;
- either skip them in `GRAPH_REPORT.md`, downgrade their surprise score, or move them to a separate “possible resolver false positives” audit subsection;
- add a regression test with `.py` + `.ts` files sharing a function name and assert the impossible inferred call is not promoted as a top surprising connection.

This preserves the graph as an audit artifact while keeping the report’s first-pass navigation signal cleaner.

## Environment

- Package: `graphifyy 0.5.4`
- Corpus shape: mixed Python scripts + TypeScript app code, AST-only extraction

ashishkr96:
## Summary

Fixes #630.

`GRAPH_REPORT.md`'s "Surprising Connections" section can promote impossible cross-language `INFERRED calls` edges (e.g. a Python helper named `normalize` linked to a TypeScript handler with a colliding name) over real, mundane edges, because `_surprise_score` rewards `INFERRED` confidence and cross-file location without checking whether the source/target language pair is plausible.

This PR filters those edges in `_cross_file_surprises()` unless an explicit import bridge between the two source files provides interop evidence.

## Approach

- Added a `_LANG_ECOSYSTEM` map grouping freely-interoperable extensions: TS/JS family, JVM (Java/Kotlin/Scala), C/C++/ObjC. Same-ecosystem inferred calls remain untouched.
- Added `_bridged_file_pairs(G)` which collects source-file pairs connected by `imports` / `imports_from` edges — concrete evidence the two files actually interop (FFI, generated bindings, configured bridge, etc.).
- In `_cross_file_surprises`, skip an edge when **all** of the following hold:
  - `relation == "calls"`
  - `confidence == "INFERRED"`
  - the two source files are in different ecosystems
  - no import edge connects the two files

This preserves the underlying graph as an audit artifact (the edge is still there for tooling that wants it) while keeping the report's first-pass navigation signal cleaner. AMBIGUOUS edges, EXTRACTED edges, semantic-similarity edges, and same-ecosystem inferred calls are all unaffected.

## Test plan

- [x] New regression test: Python→TypeScript `INFERRED calls` with no bridge is filtered out of Surprising Connections.
- [x] New regression test: cross-language `INFERRED calls` with an `imports` edge between the source files is kept (interop is plausible).
- [x] New regression test: TS↔JS `INFERRED calls` (same ecosystem) is kept.
- [x] Full suite green: `444 passed`.

### DIFF
--- a/graphify/analyze.py
+++ b/graphify/analyze.py
@@ -123,6 +123,61 @@ def _file_category(path: str) -> str:
     return "doc"
 
 
+# Map a file extension to its language ecosystem. Files in the same ecosystem
+# can plausibly call each other directly (TS↔JS, Kotlin↔Java, ObjC↔C/C++).
+# Files in different ecosystems generally cannot — a Python→TypeScript `calls`
+# edge with no import bridge is almost always resolver pollution from a
+# shared function name.
+_LANG_ECOSYSTEM = {
+    ".ts": "js", ".tsx": "js", ".js": "js", ".jsx": "js",
+    ".mjs": "js", ".ejs": "js", ".vue": "js", ".svelte": "js",
+    ".java": "jvm", ".kt": "jvm", ".kts": "jvm", ".scala": "jvm",
+    ".c": "c", ".cpp": "c", ".cc": "c", ".cxx": "c",
+    ".h": "c", ".hpp": "c", ".m": "c", ".mm": "c",
+    ".py": "python", ".pyi": "python",
+    ".rs": "rust",
+    ".go": "go",
+    ".rb": "ruby",
+    ".swift": "swift",
+    ".cs": "csharp",
+    ".php": "php",
+    ".lua": "lua",
+    ".zig": "zig",
+    ".ps1": "powershell",
+    ".ex": "elixir", ".exs": "elixir",
+    ".jl": "julia",
+    ".dart": "dart",
+    ".v": "v", ".sv": "v",
+}
+
+
+def _file_ecosystem(path: str) -> str | None:
+    """Return the language ecosystem for a source file path, or None if not code."""
+    ext = ("." + path.rsplit(".", 1)[-1].lower()) if "." in path else ""
+    return _LANG_ECOSYSTEM.get(ext)
+
+
+def _is_cross_language_call(u_source: str, v_source: str) -> bool:
+    """True when both files are code in different language ecosystems."""
+    eu = _file_ecosystem(u_source)
+    ev = _file_ecosystem(v_source)
+    return eu is not None and ev is not None and eu != ev
+
+
+def _bridged_file_pairs(G: nx.Graph) -> set[frozenset[str]]:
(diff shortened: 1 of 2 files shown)


safishamsi:
Thanks — we addressed the same issue in v0.6.2 via a different approach (conf_bonus=0 for cross-language INFERRED calls in analyze.py). Closing as already fixed.

PR #638, opened 2026-05-01
boost exact-match seeds and stop hubs hijacking query output

ReverendJack3000:
## What

`_score_nodes` gets an `EXACT_MATCH_BONUS` for terms that equal a node's `norm_label` (or `norm_label` with trailing `()` stripped, since the AST extractor emits function labels as `foo()`), and `_subgraph_to_text` accepts an optional `seeds` arg so the rendered output emits seeds first in score order before the BFS expansion sorted by degree desc.

## Why

Single-token identifier queries — by far the most common usage — get buried under high-degree hub modules. Two issues compound:

1. `_score_nodes` only used substring matching, so the seed function and every sibling that mentioned the symbol tied at score 1; ties broke alphabetically on node id.
2. `_subgraph_to_text` always re-sorted the rendered nodes by `G.degree(n)` descending, so even when `_score_nodes` returned the right seed, the hub it expanded into (e.g. `app.js`, `controller.js`) always rendered first.

### Before

```
$ graphify query "pasteFromClipboard" --budget 600
NODE app.js [src=frontend/app.js loc=L1 community=3]
NODE now() [src=frontend/modules/core/perf.js loc=L1 community=1]
NODE .filter() [src=backend/tests/_fake_inventory.py loc=L80 community=6]
NODE draw() [src=frontend/app.js loc=L4047 community=3]
NODE FakeInventoryQuery [src=backend/tests/_fake_inventory.py loc=L18 community=1]
...
```

`pasteFromClipboard()` does not appear in the first 30 NODE lines on a ~6k-node graph.

### After

```
$ graphify query "pasteFromClipboard" --budget 600
NODE pasteFromClipboard() [src=frontend/app.js loc=L4847 community=3]
NODE app.js [src=frontend/app.js loc=L1 community=3]
NODE now() [src=frontend/modules/core/perf.js loc=L1 community=1]
...
```

## Scope of behavior change

- `seeds=None` in `_subgraph_to_text` preserves the previous degree-desc ordering exactly. The only call sites updated to pass `seeds=` are `_tool_query_graph` (MCP) and the `query` CLI handler in `__main__.py`. `_tool_shortest_path` and the `path` CLI use `_score_nodes` only for endpoint resolution and do not render via `_subgraph_to_text`, so they automatically benefit from the bonus without further changes.
- Multi-term / prose queries are unaffected because the bonus only fires on `term == norm_label` exactly. Substring scoring is untouched.
- `EXACT_MATCH_BONUS = 100.0` swamps any plausible substring sum (substring score caps at ~1.5/term).

## Tests

Six new cases in `tests/test_serve.py`:

- `test_score_nodes_exact_match_beats_substring` — exact match outranks substring distractors on the same graph.
- `test_score_nodes_exact_match_strips_function_parens` — `saveDiagram()` matches term `savediagram`.
- `test_score_nodes_exact_match_no_false_positive` — unrelated term still returns no matches.
- `test_subgraph_to_text_seeds_render_first` — seed before hub when `seeds=[...]`.
- `test_subgraph_to_text_no_seeds_preserves_legacy_order` — `seeds=None` matches no-arg call (back-compat guard) and hub still wins on degree.
- `test_query_pipeline_exact_match_ranks_above_hub` — end-to-end `_score_nodes -> _bfs -> _subgraph_to_text(seeds=...)`.

Full suite (447 tests) green.

### DIFF
--- a/graphify/__main__.py
+++ b/graphify/__main__.py
@@ -1223,7 +1223,7 @@ def main() -> None:
             sys.exit(0)
         start = [nid for _, nid in scored[:5]]
         nodes, edges = (_dfs if use_dfs else _bfs)(G, start, depth=2)
-        print(_subgraph_to_text(G, nodes, edges, token_budget=budget))
+        print(_subgraph_to_text(G, nodes, edges, token_budget=budget, seeds=start))
     elif cmd == "save-result":
         # graphify save-result --question Q --answer A --type T [--nodes N1 N2 ...]
         import argparse as _ap
--- a/graphify/serve.py
+++ b/graphify/serve.py
@@ -45,13 +45,25 @@ def _strip_diacritics(text: str) -> str:
     return "".join(c for c in nfkd if not unicodedata.combining(c))
 
 
+# Identifier-style queries (e.g. a function name) often tie at score 1 against
+# every hub module that mentions the symbol, then lose tie-breaks to high-degree
+# nodes during _subgraph_to_text rendering. The bonus lifts an exact-token match
+# above any combination of substring + source-file scoring so the actual symbol
+# always seeds the BFS and renders first.
+EXACT_MATCH_BONUS = 100.0
+
+
 def _score_nodes(G: nx.Graph, terms: list[str]) -> list[tuple[float, str]]:
     scored = []
     norm_terms = [_strip_diacritics(t).lower() for t in terms]
     for nid, data in G.nodes(data=True):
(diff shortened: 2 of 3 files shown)


safishamsi:
The requested changes have been implemented in 3fdae8f. Added `_EXACT_MATCH_BONUS = 100.0` to `_score_nodes` for exact label matches (strips trailing `()` for functions), and seeds now render first in `_subgraph_to_text` before the degree-sorted expansion. Shipped in v0.6.2.

Issue #479, opened 2026-04-21
Prevent partial-chunk overwrite in --update: pre-build node-count assertion + build_merge() helper


aturner71:
## Summary

`--update` (and any multi-session incremental workflow) has a silent data-loss failure mode: calling `build()` with a partial chunk list silently replaces the existing `graph.json` with a smaller graph. There is no warning, no abort, and the resulting graph looks plausible enough that the problem isn't obvious until you notice god-node degrees have collapsed.

Two library-level changes would eliminate the entire failure class.

---

### Part A — `build()` has replace semantics, not merge semantics

`build()` constructs a graph from scratch from the provided chunk list. It does **not** load and merge with existing `graph.json`. In a multi-session incremental workflow:

1. Session 1: base extraction (123 nodes) → passed directly to `build()` → `graph.json` written. The extraction is **not** persisted as a chunk file.
2. Session 2: `build()` called with glob `".graphify_chunk_0*.json"` → finds only new chunks (e.g. chunks 04, 05, 06) → builds a 55-node graph → **silently overwrites the 123-node graph.json**.

The output looks correct — nodes cluster, god nodes appear, GRAPH_REPORT.md generates — but 2/3 of the graph is gone.

### Part B — Compounding silent edge drop from `source_file` field-name drift

Manual chunk files using field name `"source"` instead of the canonical `"source_file"` cause `build()` to silently drop all edges from those nodes (a warning is emitted, but the build continues and succeeds). In our case this dropped 107 edges from two chunks, leaving 62 edges from one conforming chunk. The graph had 55 nodes / 62 edges vs the expected 178 / 287.

**Detection signal** (had we known to look): post-build god-node list showed `execution_ledger` at #1 with 8 edges. Expected `brain_orchestrator` at 21+. A node-count check would have caught it before the file was written.

---

## Proposed Fix 1: Pre-build node-count safety check

Before overwriting `graph.json`, assert that the new graph is not smaller than the existing one. If it is, abort with a clear error instead of silently overwriting.

```python
# In build_from_json() or to_json(), before writing:
existing_path = Path("graphify-out/graph.json")
if existing_path.exists():
    existing_data = json.loads(existing_path.read_text())
    existing_n = len(existing_data.get("nodes", []))
    new_n = G.number_of_nodes()
    if new_n < existing_n:
        raise ValueError(
            f"graphify: refusing to overwrite graph.json — new graph has {new_n} nodes "
            f"but existing graph has {existing_n}. "
[...]

PR #639, opened 2026-05-01
Add --force flag (and GRAPHIFY_FORCE env var) to graphify update + git hooks

faeton:
## Summary

The node-count safety check in `to_json` (added in #479) refuses to overwrite `graph.json` when the rebuild has fewer nodes. The warning text directs users to *\"Pass force=True to override\"* — but `force` was only reachable from the Python API, with **no CLI escape hatch**.

This is painful for legitimate shrinks: refactors that delete or rename packages produce a smaller graph by design. Without a CLI flag, users get stuck unless they monkey-patch `to_json`.

## Changes

- `watch._rebuild_code`: accept `force` kwarg, propagate to `to_json`.
- `__main__` update handler: parse `--force`; also honor `GRAPHIFY_FORCE` env var.
- Hooks (post-commit + post-checkout): honor `GRAPHIFY_FORCE` env var, so automated rebuilds can opt in without editing the hook script. (Useful when a refactor PR's commit triggers the post-commit hook.)
- Help text updated.

## Reproduction (before this PR)

```bash
# After a refactor that legitimately deletes packages:
$ graphify update .
[graphify] WARNING: new graph has 1125 nodes but existing graph.json has 1139.
  Refusing to overwrite — you may be missing chunk files from a previous session.
  Pass force=True to override.
Nothing to update or rebuild failed — check output above.

# No CLI flag accepted:
$ graphify update . --force
(same warning — flag silently ignored)

$ FORCE=1 graphify update .
(same warning)
```

The only escape was a Python monkey-patch:

```python
import graphify.export as ex
orig = ex.to_json
ex.to_json = lambda *a, **k: orig(*a, force=True)
```

## After this PR

```bash
$ graphify update . --force
[graphify watch] Rebuilt: 1125 nodes, 1591 edges, 90 communities
[graphify watch] graph.json, graph.html and GRAPH_REPORT.md updated in graphify-out

# or
$ GRAPHIFY_FORCE=1 graphify update .
(same — works)

# or for the post-commit hook:
$ GRAPHIFY_FORCE=1 git commit -m \"refactor: delete dead package\"
(rebuild succeeds, no manual intervention)
```

## Tested

Verified all three paths against a real corpus (200 files, ~187k words) on macOS / Python 3.12:

- No flag → still refuses (existing behavior preserved)
- `--force` → overwrites cleanly
- `GRAPHIFY_FORCE=1` → overwrites cleanly
- Help text shows the new option

## Test plan

- [ ] `graphify update . --force` overwrites a smaller graph
- [ ] `GRAPHIFY_FORCE=1 graphify update .` overwrites a smaller graph
- [ ] `graphify update .` (no flag) still refuses on shrink — safety preserved
- [ ] `GRAPHIFY_FORCE=1` honored by the post-commit hook script

## Notes

- Watch mode (`graphify watch`) intentionally **not** changed. The continuous file watcher is incremental by nature; forcing every rebuild would defeat the safety check entirely. If desired in a follow-up PR, it can read `GRAPHIFY_FORCE` at startup.
- Closes the gap referenced by the warning text in #479.

### DIFF
--- a/graphify/__main__.py
+++ b/graphify/__main__.py
@@ -1001,6 +1001,8 @@ def main() -> None:
         print("    --dir <path>            target directory (default: ./raw)")
         print("  watch <path>            watch a folder and rebuild the graph on code changes")
         print("  update <path>           re-extract code files and update the graph (no LLM needed)")
+        print("    --force                 overwrite graph.json even if the rebuild has fewer nodes")
+        print("                            (also: GRAPHIFY_FORCE=1 env var; use after refactors that delete code)")
         print("  cluster-only <path>     rerun clustering on an existing graph.json and regenerate report")
         print("  query \"<question>\"       BFS traversal of graph.json for a question")
         print("    --dfs                   use depth-first instead of breadth-first")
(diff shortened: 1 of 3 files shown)


safishamsi:
Thanks for this — implemented the same approach manually on v6 (ca48092). The --force flag and GRAPHIFY_FORCE env var are now live, including hook support. Closing as the change is in.

Issue #628, opened 2026-04-30
Feature: support gitignore-style ! re-include patterns in .graphifyignore


sdinaro:
﻿## Summary

`.graphifyignore` is currently exclusion-only. It is loaded as raw patterns and matched with `fnmatch`; ordered negation patterns such as `!src/lib/secrets.ts` are not supported.

That makes it hard to express the common gitignore-style policy: broadly exclude sensitive-looking or generated paths, then re-include a vetted file that is safe and important for the graph.

## Current behavior

In `graphifyy 0.5.4`, `detect.py` loads non-comment lines directly:

```py
# detect.py:281-285
for line in ignore_file.read_text(...).splitlines():
    line = line.strip()
    if line and not line.startswith("#"):
        patterns.append((current, line))
```

Then `_is_ignored()` strips slashes and applies `fnmatch` as a one-way exclusion:

```py
# detect.py:314-331
for anchor, pattern in patterns:
    p = pattern.strip("/")
    ...
    if _matches(rel, p):
        return True
```

There is no ordered include/exclude state, so a negated pattern is just another literal-ish pattern and cannot re-include a file.

## Reproducer

```text
# .graphifyignore
**/*secret*
!src/lib/secrets.ts
```

Expected: `src/lib/secrets.ts` is included if it is a deliberately vetted source file.

Actual: `.graphifyignore` cannot express this; consumers must either weaken the broad ignore or monkey-patch detection with a separate allowlist.

## Why this matters

Some projects have source files with names that look sensitive (`secrets.ts`, `token-store.ts`, etc.) but contain only SDK wrappers or typed accessors, not secret values. Those files are often important architecture nodes. The current behavior pushes consumers toward local wrappers because the native ignore file cannot encode “exclude broadly, then re-include this reviewed exception.”

## Suggested fix

Use `pathspec` with `gitwildmatch` semantics, or implement the same ordered negation behavior directly:

- evaluate patterns in file order;
- regular matches set ignored = true;
- `!` matches set ignored = false;
- preserve the existing ancestor `.graphifyignore` behavior;
- add tests for root and parent `.graphifyignore` files with both exclusion and re-inclusion.

## Environment

- Package: `graphifyy 0.5.4`
- Platform: Windows + Python 3.13 locally, but the behavior is platform-independent.

safishamsi:
[...]

PR #644, opened 2026-05-01
 [ENH] Refactor .graphifyignore engine to match .gitignore semantics

Deep-Axe:
This replaces the ```fnmatch```-based ignore logic with a robust matching engine that maintains functional parity to .gitignore. This resolves Issue #628 by supporting ordered negation (!).

### Features & Enhancements

1. Full GitWildMatch Semantics
The new engine moves beyond simple string matching to a regex-based compiler that supports the full Git specification:
- Ordered Negation (!): Implements "last-match-wins" logic, allowing users to broadly exclude paths and re-include vetted exceptions.
- Recursive Wildcards (`/`):** Correctly handles zero or more directory segments.
- Path Anchoring: Distinguishes between root-anchored patterns (/main.py) and unanchored global patterns (main.py).
- Directory-Only Matching: Respects trailing slashes (e.g., vendor/) to match only directories while leaving files with the same name included.
- Character Classes: Full support for bracket expressions ([abc], [!abc]) and POSIX classes ([[:alpha:]], [[:digit:]]).

2. High-Performance Traversal
Optimization of the discovery and extraction loops to handle repositories efficiently:
-  In-Place Pruning: Both detect.py and extract.py now prune dirnames in-place during os.walk. Massive ignored subtrees (like node_modules/ or vendor/) are skipped at the structural level rather than being traversed and filtered per-file.
- Redundant Stat Removal: The ignore checker now accepts an explicit is_dir flag. By passing is_dir=False for confirmed files, we eliminate unnecessary filesystem stat() calls during deep scans.

3. Bug Fixes & Alignment
-  Memory Re-indexing Fix: Resolved a conflict where ```graphify-out/memory/``` files were being incorrectly ignored. These are now explicitly exempted so the graph can always learn from past query results.
- Watcher Alignment: The file watcher (```watch.py```) now respects ```.graphifyignore```, preventing redundant graph rebuilds when ignored files change.
-  Whitespace Integrity: Ported Git's exact ```trim_trailing_spaces``` to handle backslash-escaped trailing spaces correctly.

Verification Results
- Conformance Suite: Added tests/test_ignore_rules.py covering anchoring, negation chains, and directory pruning.
- Unit Tests: Updated tests/test_detect.py to verify the new engine against standard ignore categories.
- Status: All 49 non-symlink tests pass successfully.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -17,6 +17,7 @@ graphify-out/
 .graphify_*.json
 .graphify_python
 .claude/
+.gemini/
 skills/
 docs/superpowers/
 .vscode/
--- a/README.md
+++ b/README.md
@@ -49,7 +49,7 @@ dist/
 *.generated.py
 ```
 
-Same syntax as `.gitignore`. You can keep a single `.graphifyignore` at your repo root — patterns work correctly even when graphify is run on a subfolder.
+Full gitignore semantics: `**` crosses directories, `!` negates (last pattern wins), trailing `/` matches directories only, leading `/` anchors to the root. Place it at your repo root — graphify walks up to the `.git` boundary to find it, so patterns apply correctly even when run on a subfolder.
 
 ## How it works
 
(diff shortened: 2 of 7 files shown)


safishamsi:
Thanks — .graphifyignore gitignore semantics were fixed in v0.6.1. Closing as already covered.

Deep-Axe:
no they aren't, you haven't implemented negation which was the core issue from #628, you're closing issues and PRs without properly reviewing them

PR #646, opened 2026-05-01
fix(detect): hermetic .graphifyignore discovery + correct gitignore   semantics

### LINKED ISSUE #643: Enhancement: Decouple .graphifyignore discovery from .git boundaries
The current ```_load_graphifyignore``` implementation relies exclusively on the presence of a ```.git/``` directory to stop its upward search for ignore rules.

### The Problem
Currently, ```graphify``` assumes that every project is managed by Git. If a user runs the tool on a directory that is not a Git repository (e.g., a simple folder of research papers or a project using Mercurial/SVN), the logic will:

1. Continue climbing the directory tree past the intended project folder.

2. Potentially reach the user's home directory or the filesystem root.

3. Accidentally pick up and apply ```.graphifyignore``` rules from unrelated projects found higher up in the hierarchy.

This makes the tool less predictable when used outside of a strictly Git-centric workflow.

Deep-Axe:
Refines 6b68de1 and 8a6306f. Closes #643. Related to #628. Adding on from #644 
 
 ### Features implemented:

- **No-VCS ceiling**: when no VCS root is found, the walk stops at the scan root, not the home directory. The home guard still allows leakage across sibling projects under a shared workspace (`~/work/project-alpha` picking up `~/work/.graphifyignore`). Stopping at the scan root eliminates that entirely.

- **Correct pattern ordering**: parent `.graphifyignore` patterns are evaluated outermost-first so inner (closer) patterns win via
  last-match-wins. The prior ordering let VCS-root patterns override a file's own `.graphifyignore`.

- **Escaped trailing spaces preserved**:  `line.strip()` silently drops `vendor\ ` patterns. Replaced with `rstrip()` + escape-awareness so `vendor\ ` correctly matches a path named `vendor ` (per gitignore spec).

- **`\]` escape**: `]` was missing from the backslash-escape set in `_pattern_to_regex`; `\]` was being converted to a path separator
  instead of a literal `]`.

- **Anchored pattern misfire**: a leading-slash pattern (e.g. `/vendor/`) in a parent `.graphifyignore` was incorrectly matching `src/vendor/` when `detect()` was called on `src/`. Added a guard so anchored patterns only evaluate relative to their own anchor directory, not the scan root.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -17,6 +17,7 @@ graphify-out/
 .graphify_*.json
 .graphify_python
 .claude/
+.gemini/
 skills/
 docs/superpowers/
 .vscode/
--- a/README.md
+++ b/README.md
@@ -49,7 +49,7 @@ dist/
 *.generated.py
 ```
 
-Same syntax as `.gitignore`. You can keep a single `.graphifyignore` at your repo root — patterns work correctly even when graphify is run on a subfolder.
+Full gitignore semantics: `**` crosses directories, `!` negates (last pattern wins), trailing `/` matches directories only, leading `/` anchors to the root. Place it at your repo root — graphify walks up to the `.git` boundary to find it, so patterns apply correctly even when run on a subfolder.
 
 ## How it works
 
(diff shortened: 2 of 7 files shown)


safishamsi:
Thanks for the thorough work — .graphifyignore semantics were fixed in v0.6.1 covering the same cases. Closing as already covered.

Deep-Axe:
the issue #643 is not addresses which is what PR #646 was meant to solve, like i has mentioned in my last comments from #643 i would encourage you to read the code you push out since your fixes do the opposite of your claim, you are returning True on the first match while iterating from the ceiling down, global rules will now always override local ones and you have closed this PR without properly reviewing it and just pushing out unverified AI code.

PR #648, opened 2026-05-01
feat: add VB.NET (.vb) language support via tree-sitter

rrangraj:
This commit adds full VB.NET language support to graphify, raising the supported language count from 25 to 26. The implementation follows the established LanguageConfig pattern used by all other tree-sitter-backed extractors.

New dependency:
- Adds optional extra [vbnet] backed by tree-sitter-vbnet (published to PyPI at https://pypi.org/project/tree-sitter-vbnet/0.1.0/). Install with: pip install graphifyy[vbnet]

graphify/detect.py:
- Added .vb to CODE_EXTENSIONS so VB.NET files are discovered during corpus ingestion and file-system watching.

graphify/extract.py:
- _import_vbnet(): import handler for imports_statement nodes; emits imports edges using the namespace_name child text.
- _vbnet_extra_walk(): extra-walk hook that intercepts namespace_block nodes, emits a namespace node, and recurses.
- _VBNET_CONFIG: full LanguageConfig covering class_block / module_block / structure_block / interface_block as class types; method_declaration / constructor_declaration / property_declaration as function types; invocation call nodes with target/member_access fields.
- VB.NET-specific branches in _extract_generic:
  * Class body: VB.NET has no wrapper body node; inherits and implements are named fields directly on the class_block. Emits separate inherits and implements edges for each base type, stripping generic arguments.
  * Constructor name: constructor_declaration carries no name field in the grammar; always resolves to New.
  * Function body: uses the declaration node itself as body sentinel so the call-graph pass can find invocations inside methods.
- extract_vbnet(path): public wrapper that delegates to _extract_generic.
- _DISPATCH['.vb']: routes .vb files to extract_vbnet.

pyproject.toml:
- Added vbnet = ['tree-sitter-vbnet'] optional dependency group.
- Added 'tree-sitter-vbnet' to the all extra.

tests/fixtures/sample.vb:
- New fixture file exercising: Imports statements, Namespace block, Interface, Class with Inherits + Implements, Module, Structure, Sub/Function/Property methods, and method calls.

tests/test_languages.py:
- Added 13 tests covering: no-error, class/interface/module/structure detection, method detection, imports relation, inherits edge, implements edge, and no-dangling-edges invariant.

README.md:
- Updated language count 25 to 26.
- Added VB.NET to language list and file-extension table.

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,7 +23,7 @@
 
 **An AI coding assistant skill.** Type `/graphify` in Claude Code, Codex, OpenCode, Cursor, Gemini CLI, GitHub Copilot CLI, VS Code Copilot Chat, Aider, OpenClaw, Factory Droid, Trae, Hermes, Kiro, or Google Antigravity - it reads your files, builds a knowledge graph, and gives you back structure you didn't know was there. Understand a codebase faster. Find the "why" behind architectural decisions.
 
-Fully multimodal. Drop in code, PDFs, markdown, screenshots, diagrams, whiteboard photos, images in other languages, or video and audio files - graphify extracts concepts and relationships from all of it and connects them into one graph. Videos are transcribed with Whisper using a domain-aware prompt derived from your corpus. YAML/YML files (Kubernetes, Kustomize, Helm, config) are indexed for semantic extraction. SQL files are AST-extracted deterministically — tables, views, functions, foreign keys, and FROM/JOIN relationships map directly into the graph with no LLM needed. 25 languages supported via tree-sitter AST (Python, JS, TS, Go, Rust, Java, C, C++, Ruby, C#, Kotlin, Scala, PHP, Swift, Lua, Zig, PowerShell, Elixir, Objective-C, Julia, Verilog, SystemVerilog, Vue, Svelte, Dart).
+Fully multimodal. Drop in code, PDFs, markdown, screenshots, diagrams, whiteboard photos, images in other languages, or video and audio files - graphify extracts concepts and relationships from all of it and connects them into one graph. Videos are transcribed with Whisper using a domain-aware prompt derived from your corpus. YAML/YML files (Kubernetes, Kustomize, Helm, config) are indexed for semantic extraction. SQL files are AST-extracted deterministically — tables, views, functions, foreign keys, and FROM/JOIN relationships map directly into the graph with no LLM needed. 26 languages supported via tree-sitter AST (Python, JS, TS, Go, Rust, Java, C, C++, Ruby, C#, Kotlin, Scala, PHP, Swift, Lua, Zig, PowerShell, Elixir, Objective-C, Julia, Verilog, SystemVerilog, Vue, Svelte, Dart, VB.NET).
 
 > Andrej Karpathy keeps a `/raw` folder where he drops papers, tweets, screenshots, and notes. graphify is the answer to that problem - 71.5x fewer tokens per query vs reading the raw files, persistent across sessions, honest about what it found vs guessed.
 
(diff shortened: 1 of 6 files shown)


safishamsi:
Great addition! The implementation looks clean and follows the established pattern well. One blocker: `tree-sitter-vbnet` is not yet available on PyPI (`pip index versions tree-sitter-vbnet` returns no distributions). Once the package is published we can merge this — please ping here when it's up.

ventouris:
Is this merged already? I installed the latest version and it says the vb.net is not supported

> The repo has xxx .vb files but graphify's detector doesn't recognise VB.NET.

A-PhK:
For anyone wondering why VB.NET support from this PR doesn't appear to be available anymore: it seems to have been subsequently removed due to the tree-sitter-vbnet dependency only providing Windows wheels, which prevents Linux/macOS support.

See the removal commit:
[https://github.com/Graphify-Labs/graphify/commit/1e848458f9df16fc961e01bf9b5ba0979d02ea12]

The original PR was indeed merged, but the VB.NET support was later removed.

ivanstanojkovski:
@safishamsi you asked here to be pinged once the dependency situation was
sorted out. It never was on that route — as @A-PhK notes above, the
tree-sitter-vbnet wheels stayed Windows-only, and the support was removed
in 1e84845.

We hit the same wall on our own codebase and took a different route, which
removes that blocker entirely. Posting the numbers here, and asking whether
this direction would be welcome before I write anything.

## Our starting point

We maintain a large VB.NET business application: 2,051 `.vb` files, about
95% of the codebase. With no VB support the graph shows us almost nothing.

We tried tree-sitter-vbnet first. On those 2,051 files:

- 1% of files parsed with no errors
- 43.5% of all bytes ended up inside ERROR nodes
- the largest files produced nothing at all

We then neutralised the four constructs it choked on most often. That
moved the error share from 43.5% to 43.4%, so this is not a small gap in
the grammar that could be patched.

## What we built instead

A small helper program built on Roslyn — `Microsoft.CodeAnalysis.VisualBasic`,
the VB compiler Microsoft ships. graphify runs it once per extraction. It
reads the `.vb` files and prints plain JSON: types, members, calls,
inheritance and event wiring. Python turns that JSON into nodes and edges
and keeps ownership of node ids.

On the same 2,051 files: **no parse errors, about 10 seconds.**

## Platforms

This is the part that matters for this thread. The helper targets .NET 8.
I built it for three platforms from one machine and checked what came out:

| target | needs .NET installed | size | files |
|---|---|---|---|
| linux-x64 | no | 77.5 MB | 1 |
| linux-x64 | yes | 13.8 MB | 8 |
| osx-arm64 | no | 89.5 MB | 190 |
| osx-arm64 | yes | 13.9 MB | 8 |
| win-x64 | no | 84.3 MB | 190 |
| win-x64 | yes | 13.9 MB | 8 |

I checked the file headers: the Linux output is a real ELF binary and the
macOS output is a real Mach-O binary. The single-file row uses
`PublishSingleFile`; I applied it to linux-x64 in this test, and it works
the same for the other two.

One thing I have not done: run the Linux build. I have no Linux machine
here, so I can show it builds but not that it runs. CI would settle that
in one step.

## How it could be shipped

- CI builds the binaries and attaches them to a release
- graphify uses the one for the current platform
- if there is no binary for that platform but `dotnet` is on PATH, build
  from source instead
- if neither is available, skip `.vb` files and print one line saying so

That way .NET is never a requirement, only used when it is there. ruff and
esbuild ship along similar lines.

## Questions before I open a PR

1. Is an external helper acceptable at all? It has a different shape from
   the tree-sitter extractors, because it is a separate process rather
   than a `LanguageConfig`.
2. Does the shipping plan above look right, or would you prefer something
   else?
3. Should CI install .NET so the VB tests run, or should those tests skip
   when it is missing?

Happy to go whichever way you prefer. If the answer is no, that is fine as
well — better to ask before sending a large PR.

PR #654, opened 2026-05-02
feat: implement parallel AST extraction with benchmarking support

hanzala-sohrab:
## Description
Parallelizes AST extraction in `extract()` using `ProcessPoolExecutor` to eliminate the CPU bottleneck on large repositories. 

`importlib.import_module` holds the Python GIL, so processes were chosen over threads to ensure true parallelism.

## Changes
- **Parallel Dispatch**: Refactored `extract()` to split cached vs. uncached files upfront. If `uncached >= 20` (to account for ~50ms process spawn overhead), extraction is dispatched to a process pool.
- **Pickle-safe Workers**: Moved the extractor dispatch dictionary (`_DISPATCH`) and the worker function (`_extract_single_file`) to the module level so they can be pickled by the executor.
- **API Continuity**: Added `parallel=True` and `max_workers=None` as keyword-only arguments to `extract()`. Existing callers are fully backward compatible. Defaults to `min(os.cpu_count() or 4, num_files, 8)` workers to prevent memory explosion.
- **Benchmark Script**: Added `tests/bench_extract.py` to measure sequential vs parallel time and validate identical node/edge counts.

## Performance
Measured via `uv run python tests/bench_extract.py .` on the graphify repo itself (84 files):
- **Sequential**: 0.77s
- **Parallel (8 workers)**: 0.46s
- **Speedup**: 1.66x

*Note: Speedup scales non-linearly with repo size. Expect 3-5x on 1k+ file repos where process spawn overhead is negligible relative to parsing time.*

## Testing
- Automated tests pass (`442 passed`) with zero regressions.
- Validated that parallel execution produces identical `nodes` and `edges` output (order-agnostic) as sequential execution.

### DIFF
--- a/graphify/extract.py
+++ b/graphify/extract.py
@@ -3417,7 +3417,167 @@ def _check_tree_sitter_version() -> None:
         )
 
 
-def extract(paths: list[Path], cache_root: Path | None = None) -> dict:
+_DISPATCH: dict[str, Any] = {
+    ".py": extract_python,
+    ".js": extract_js,
+    ".jsx": extract_js,
+    ".mjs": extract_js,
+    ".ts": extract_js,
+    ".tsx": extract_js,
+    ".go": extract_go,
+    ".rs": extract_rust,
+    ".java": extract_java,
+    ".c": extract_c,
+    ".h": extract_c,
+    ".cpp": extract_cpp,
+    ".cc": extract_cpp,
+    ".cxx": extract_cpp,
+    ".hpp": extract_cpp,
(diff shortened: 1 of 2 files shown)


safishamsi:
Parallel AST extraction is a solid idea but this PR is based on an older version (reports 442 tests, we're now at 449) and ProcessPoolExecutor pickling is fragile across the extractor dispatch table. If you rebase against v6 and all 449 tests pass we'd be happy to review again.

PR #668, opened 2026-05-02
Pin UTF-8 encoding for cache and manifest I/O

azizur100389:
## Summary

On Windows the default text codec is cp1252, so `Path.read_text()` / `Path.write_text()` raise `UnicodeDecodeError` / `UnicodeEncodeError` whenever the graphify cache or manifest holds non-ASCII content — for example:

- Chinese / Japanese / Korean identifier names extracted from source code
- File paths under directories with accented Latin or CJK characters
- Emoji or non-ASCII labels in semantic extraction output

Four call sites are affected:

- `cache.load_cached()` — `read_text(encoding="utf-8")` + catch `UnicodeDecodeError`
- `cache.save_cached()` — `write_text(encoding="utf-8")` with `ensure_ascii=False` so CJK/emoji stay readable UTF-8 instead of `\uXXXX` escapes
- `detect.load_manifest()` — `read_text(encoding="utf-8")`
- `detect.save_manifest()` — `write_text(encoding="utf-8")` with `ensure_ascii=False`

The repo already uses `encoding="utf-8"` in `__main__.py` and `export.py`; this extends the same convention to the file types most likely to hold user data.

## Test plan
- [x] New `tests/test_encoding.py` (13 tests) covers cache + manifest roundtrips with CJK, Japanese, Korean, accented Latin, emoji, and mixed-script payloads — including non-ASCII filenames
- [x] Verifies cache files on disk are valid UTF-8 JSON readable by external tools
- [x] No regressions in existing tests

### DIFF
--- a/graphify/cache.py
+++ b/graphify/cache.py
@@ -55,8 +55,8 @@ def load_cached(path: Path, root: Path = Path(".")) -> dict | None:
     if not entry.exists():
         return None
     try:
-        return json.loads(entry.read_text())
-    except (json.JSONDecodeError, OSError):
+        return json.loads(entry.read_text(encoding="utf-8"))
+    except (json.JSONDecodeError, OSError, UnicodeDecodeError):
         return None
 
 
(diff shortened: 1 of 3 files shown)


azizur100389:
Closing as redundant — checked the v6 default branch and `graphify/cache.py` already uses `encoding="utf-8"` (line 87, 95), and `detect.load_manifest`/`save_manifest` already use it too. This PR was opened against v3 before realizing v6 was the active branch. Sorry for the noise.

PR #670, opened 2026-05-02
Fix #525: document --wiki flag in remaining platform skill files

### LINKED ISSUE #525: Docs inconsistency: is --wiki supported outside Claude Code?
## Problem

`--wiki` is documented in the main Claude-oriented skill file, but it does not appear in the non-Claude platform skill files.

From what I can see in `v4`:

- `graphify/skill.md` documents `/graphify <path> --wiki` in the Usage section as `build agent-crawlable wiki (index.md + one article per community)`.
- The same file also has a dedicated `Step 6b - Wiki (only if --wiki flag)` section.
- But platform-specific skill files such as `graphify/skill-codex.md` and `graphify/skill-opencode.md` do not list `--wiki` in their Usage sections.
- The README presents graphify as working across multiple assistants/platforms, so this ends up being unclear from a user perspective.

## Question

Is `--wiki` currently intended to be:

1. supported on all platforms, and just missing from the non-Claude skill docs
2. supported only on Claude Code right now
3. supported on some subset of platforms only

## Why this is confusing

If I compare the docs, the current impression is that only the main `skill.md` exposes `--wiki`, while other platform skill variants do not. That makes it hard to tell whether:

- the flag is actually unsupported outside Claude Code
- the support exists but the platform-specific skill docs are stale/incomplete
- there are platform-specific caveats that are not documented

## Request

Could you clarify the support status of `--wiki` for platforms other than Claude Code, such as Codex / OpenCode / Cursor / Aider / OpenClaw / others?

If `--wiki` is supported more broadly, it would help to update the platform-specific skill files so they stay aligned with `graphify/skill.md`.

If it is intentionally Claude-only, it would help to document that explicitly in the main docs and/or platform-specific docs.

## References

- Main skill file with `--wiki`: `graphify/skill.md` on `v4`
- Platform-specific examples missing `--wiki`: `graphify/skill-codex.md`, `graphify/skill-opencode.md` on `v4`

azizur100389:
## Summary

Issue #525 reports that `--wiki` is documented in `skill.md` but missing from several platform skill files, leaving users on those platforms unable to discover or invoke the wiki export.

Audit of v6:

| Skill file | Has `--wiki` before this PR? |
|---|---|
| skill.md | yes |
| skill-windows.md | yes |
| skill-opencode.md | yes |
| skill-copilot.md | yes |
| **skill-codex.md** | **no** |
| **skill-claw.md** | **no** |
| **skill-droid.md** | **no** |
| **skill-trae.md** | **no** |
| **skill-aider.md** | **no** |

Each of the 5 missing files now gets:
- A Usage line: `/graphify <path> --wiki   # build agent-crawlable wiki...`
- A `Step 6b - Wiki (only if --wiki flag)` block, mirroring the canonical block in `skill.md` but adapted to each file's existing path conventions (these 5 use `.graphify_python` rather than `graphify-out/.graphify_python`).

The Step 6b code matches the live `to_wiki()` API: `to_wiki(G, communities, output_dir, community_labels, cohesion, god_nodes_data) -> int` (verified via `inspect.signature`).

Closes #525

## Test plan
- [x] Pure docs change in 5 markdown files; no Python changes, no new tests required
- [x] Verified `from graphify.wiki import to_wiki` and signature matches the skill code
- [x] All existing tests still pass (no new regressions)

### DIFF
--- a/graphify/skill-aider.md
+++ b/graphify/skill-aider.md
@@ -22,6 +22,7 @@ Turn any folder of files into a navigable knowledge graph with community detecti
 /graphify <path> --graphml                            # export graph.graphml (Gephi, yEd)
 /graphify <path> --neo4j                              # generate graphify-out/cypher.txt for Neo4j
 /graphify <path> --neo4j-push bolt://localhost:7687   # push directly to Neo4j
+/graphify <path> --wiki                               # build agent-crawlable wiki (index.md + one article per community)
 /graphify <path> --mcp                                # start MCP stdio server for agent access
 /graphify <path> --watch                              # watch folder, auto-rebuild on code changes (no LLM needed)
 /graphify add <url>                                   # fetch URL, save to ./raw, update graph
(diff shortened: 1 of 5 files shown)


bcdiaconu:
also missing from `skill-pi.md`

graphify/skill-aider.md
```diff
@@ -507,6 +508,36 @@ else:
...
+communities = {int(k): v for k, v in analysis['communities'].items()}
+cohesion = {int(k): v for k, v in analysis['cohesion'].items()}
+labels = {int(k): v for k, v in labels_raw.items()}
+gods = god_nodes(G)
+
+n = to_wiki(G, communities, 'graphify-out/wiki', community_labels=labels or None, cohesion=cohesion, god_nodes_data=gods)
+print(f'Wiki: {n} articles written to graphify-out/wiki/')
+print('  graphify-out/wiki/index.md  ->  agent entry point')
+"
+```
```
bcdiaconu, line 539: is it actually required to run the wiki this way or should `graphify export wiki` be used instead?

bcdiaconu: same question for all other files.

bcdiaconu (commented):
is this the correct way to generate wiki?

PR #96, opened 2026-04-08
Add HTML sampling for 5000+ node graphs and extraction progress logging (#52)

azizur100389:
## Summary
Addresses three sub-issues from #52 (Large-Scale App Project Analysis):

- **tree-sitter version mismatch (sub-issue 1):** `_extract_generic()` now catches `TypeError` from `Language()` API changes and returns a clear error message suggesting `pip install --upgrade tree-sitter`, instead of an opaque traceback
- **HTML visualization for large graphs (sub-issue 5):** Added `sample_for_viz()` that selects the top-N largest communities and highest-degree nodes per community. `to_html()` accepts `sample=True` to auto-sample instead of raising `ValueError` on graphs exceeding 5,000 nodes
- **Extraction progress feedback (sub-issue 8):** `extract()` now logs periodic progress (every 10% or 100 files) showing files processed, cached, skipped, and errors — users get feedback during large batch extractions instead of silence

Note: sub-issues 2 and 3 from #52 were already fixed in v3 (`_resolve_cross_file_imports` guard and `_ASSET_DIR_MARKERS` for xcassets).

## Test plan
- [x] 9 new tests covering all three fixes (`tests/test_issue52.py`)
- [x] All existing export, extract, and detect tests pass
- [x] Verified `sample_for_viz` produces subgraphs under 5,000 nodes
- [x] Verified progress logging fires only for batches >= 100 files

PR #671, opened 2026-05-02
Surface tree-sitter version-mismatch hint instead of bare TypeError

azizur100389:
## Summary

When a user has an older `tree-sitter` paired with a newer language binding (or vice versa), `Language()` raises `TypeError` with messages like *"missing 1 required positional argument: 'name'"*. The current catch-all `except Exception` in `_extract_generic()` stores that bare message in the per-file `error` field, leaving users no actionable signal.

Add a dedicated `TypeError` branch that returns the upgrade hint:

> tree-sitter version mismatch for `tree_sitter_python`: missing 1 required positional argument: 'name'. Try: `pip install --upgrade tree-sitter tree-sitter-languages`

Behavior is unchanged for all non-version-mismatch errors — the broader `Exception` handler still runs.

This PR replaces the old #96 (which targeted stale v3 and bundled three changes). The other two changes from #96 are no longer needed:
- Progress logging — already implemented in v6 (parallel/sequential extraction emit it).
- HTML sampling — overlaps with v6's `GRAPHIFY_VIZ_NODE_LIMIT` env-var approach, so leaving v6's design as-is.

## Test plan
- [x] New `tests/test_extract.py::test_extract_generic_surfaces_tree_sitter_version_mismatch_hint` — uses `monkeypatch.setitem(sys.modules, ...)` to inject a fake `tree_sitter` whose `Language()` raises `TypeError`, then asserts the returned error contains both `"tree-sitter version mismatch"` and `"pip install --upgrade"`.
- [x] All existing tests still pass; no behavior change for any path that doesn't hit `TypeError`.

### DIFF
--- a/graphify/extract.py
+++ b/graphify/extract.py
@@ -863,6 +863,15 @@ def _extract_generic(path: Path, config: LanguageConfig) -> dict:
         language = Language(lang_fn())
     except ImportError:
         return {"nodes": [], "edges": [], "error": f"{config.ts_module} not installed"}
+    except TypeError as e:
+        # tree-sitter version mismatch: old Language() expects (lib_path),
+        # new Language() expects (language_capsule, name). Surface a hint
+        # so users see the upgrade path instead of a bare TypeError.
+        hint = (
+            f"tree-sitter version mismatch for {config.ts_module}: {e}. "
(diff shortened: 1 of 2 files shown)


safishamsi:
Merged via PR #671. Shipped in v0.7.8.

PR #673, opened 2026-05-02
Add --with-wiki flag to 'graphify hook install' (#549)

### LINKED ISSUE #549: Feature request: post-commit hook should optionally rebuild the wiki

azizur100389:
## Summary

Issue #549 reports that the post-commit hook rebuilds `graph.json` / `graph.html` / `GRAPH_REPORT.md` but skips the wiki, so agents following the integration guidance (reading `graphify-out/wiki/index.md`) navigate stale community articles after every code change. The reporter's workaround was either running `graphify <path> --wiki` manually or hand-editing the hook (clobbered by the next `graphify hook install`).

This adds an opt-in `--with-wiki` flag:

```
graphify hook install               # unchanged - no wiki rebuild
graphify hook install --with-wiki   # post-commit hook now also rebuilds graphify-out/wiki/
```

Implementation matches the design in #549: no LLM cost, reuses the freshly-written `graph.json` plus `.graphify_labels.json` to call `wiki.to_wiki()`.

### Diff

- `graphify/hooks.py` — extracted `_HOOK_SCRIPT` into a `_build_hook_script(with_wiki: bool)` factory; `install()` takes a `with_wiki=False` kwarg
- `graphify/__main__.py` — parse `--with-wiki` for `hook install`; updated help text
- `tests/test_hooks.py` — 4 new tests

### Backward compatibility

Default behavior unchanged: `graphify hook install` produces the same script as before. A regression test (`test_build_hook_script_wiki_block_only_when_requested`) asserts byte-for-byte equivalence with the legacy `_HOOK_SCRIPT` constant when `with_wiki=False`. Existing installations don't need any action.

Closes #549

## Test plan

4 new tests:
- `test_install_default_omits_wiki_rebuild` — without flag, hook script doesn't import `to_wiki`
- `test_install_with_wiki_injects_wiki_rebuild` — with flag, hook contains the wiki block + AST rebuild + `--with-wiki` marker
- `test_install_with_wiki_then_uninstall_clean` — uninstall removes the entire graphify section, no wiki residue
- `test_build_hook_script_wiki_block_only_when_requested` — `_build_hook_script(with_wiki=False) == _HOOK_SCRIPT` byte-for-byte

All existing tests pass; no behavior change for non-opt-in users.

### DIFF
--- a/graphify/__main__.py
+++ b/graphify/__main__.py
@@ -1067,6 +1067,7 @@ def main() -> None:
         print("    --label NAME            project label in header")
         print("  benchmark [graph.json]  measure token reduction vs naive full-corpus approach")
         print("  hook install            install post-commit/post-checkout git hooks (all platforms)")
+        print("    --with-wiki             also rebuild the wiki on each commit (#549)")
         print("  hook uninstall          remove git hooks")
         print("  hook status             check if git hooks are installed")
         print("  gemini install          write GEMINI.md section + BeforeTool hook (Gemini CLI)")
@@ -1232,13 +1233,14 @@ def main() -> None:
         from graphify.hooks import install as hook_install, uninstall as hook_uninstall, status as hook_status
         subcmd = sys.argv[2] if len(sys.argv) > 2 else ""
         if subcmd == "install":
-            print(hook_install(Path(".")))
+            with_wiki = "--with-wiki" in sys.argv[3:]
+            print(hook_install(Path("."), with_wiki=with_wiki))
         elif subcmd == "uninstall":
             print(hook_uninstall(Path(".")))
         elif subcmd == "status":
             print(hook_status(Path(".")))
         else:
-            print("Usage: graphify hook [install|uninstall|status]", file=sys.stderr)
+            print("Usage: graphify hook [install [--with-wiki] | uninstall | status]", file=sys.stderr)
             sys.exit(1)
     elif cmd == "query":
         if len(sys.argv) < 3:
--- a/graphify/hooks.py
+++ b/graphify/hooks.py
@@ -42,10 +42,43 @@ if [ -z "$GRAPHIFY_PYTHON" ]; then
 fi
 """
 
-_HOOK_SCRIPT = """\
+_WIKI_REBUILD_PYTHON = """
+# Optional wiki rebuild (#549 - 'graphify hook install --with-wiki')
+try:
+    import json as _json
+    out_dir = Path('graphify-out')
+    graph_json = out_dir / 'graph.json'
+    if graph_json.exists():
+        from graphify.build import build_from_json
+        from graphify.cluster import cluster, score_all
+        from graphify.analyze import god_nodes
+        from graphify.wiki import to_wiki
+        graph_data = _json.loads(graph_json.read_text(encoding='utf-8'))
+        G2 = build_from_json(graph_data)
+        comms = cluster(G2)
+        cohesion = score_all(G2, comms)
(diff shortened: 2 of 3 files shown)


Qodo-Free-For-OSS:
Hi, `graphify hook install --with-wiki` does not take effect if the hook is already installed because `_install_hook()` returns "already installed" when it sees the marker and never rewrites the hook body. This makes it impossible to upgrade/downgrade an existing hook’s wiki behavior via reinstall.

**How to fix:** Rewrite existing marked hook

> ### Issue description
> Reinstalling hooks with a different `with_wiki` value does not update the already-installed hook because `_install_hook()` returns early when it finds the marker.
> 
> ### Issue Context
> Users will expect `graphify hook install --with-wiki` to enable wiki rebuild even if they previously installed hooks without it.
> 
> ### Fix Focus Areas
> - graphify/hooks.py[211-222]
> - graphify/hooks.py[246-265]
> 
> ### Implementation notes
> - Update `_install_hook()` to *replace* the existing graphify section between `# graphify-hook-start` and `# graphify-hook-end` when the marker exists (instead of returning early).
> - Alternatively, add a `force`/`overwrite` option and use it for `hook install --with-wiki` upgrades, but the best UX is: marker present => update in-place.
> - Ensure uninstall still removes the section correctly afterward.

---

Issue #712, opened 2026-05-04
extract_svelte() regex fallback stamps stub-node source_file to importer's path, corrupting target file metadata after #701


jippi:
## Summary

After #701 fixed the alias-resolution + synthetic-ID problems in `extract_svelte()`, dynamic_import edges now resolve to real file nodes correctly. But the regex fallback still creates a **stub node** for the import target with the wrong `source_file`: it stamps the *importer's* path instead of the *target's* path.

When `_extract_generic` later runs on the target file and emits its own canonical file node (same ID), `build_from_json` does last-write-wins on attributes (`G.add_node` overwrite). Depending on extraction order, the importer-supplied stub can clobber the correct `source_file` from the real extraction.

The `[contains]` edges (file → its functions) are still emitted correctly, but anything that reads `source_file` off a file node — display, "where is X defined", blast-radius, community summaries — sees the file as living inside its importer.

## Repro (real SvelteKit app, 1,873 files)

```python
# After `graphify update .` on graphifyy 0.7.5
import json
g = json.load(open('graphify-out/graph.json'))

for n in g['nodes']:
    if n['id'] == 'src_lib_components_library_manage_library_entry_popover_svelte':
        print(n['label'])         # $lib/components/library/manage-library-entry-popover.svelte
        print(n['source_file'])   # src/partials/Series/Grid/Card/series-card-library.svelte  ← WRONG
```

The popover lives at `src/lib/components/library/manage-library-entry-popover.svelte`. Its `source_file` is stamped to a *consumer* (`series-card-library.svelte`) that does `{#await import('$lib/components/library/manage-library-entry-popover.svelte')}`.

**Scope on this repo:** 16 svelte files affected (every dynamically-imported component I sampled). Examples:

| Component (label) | Stamped source_file (wrong) |
|---|---|
| `$lib/components/ui/confirm-delete-dialog/...svelte` | `src/routes/+layout.svelte` |
| `$lib/components/ui/sonner/sonner.svelte` | `src/routes/+layout.svelte` |
| `$lib/components/series/similar/similar-series-section.svelte` | `src/routes/(series)/[id=id]/(view)/(meta)/+page.svelte` |
| `$lib/components/widget/custom-rating.svelte` | `src/lib/components/search/search-form.svelte` |

## Root cause

`graphify/extract.py` in `extract_svelte()` (~lines 1797–1806):

```python
result.setdefault("nodes", []).append({
    "id": node_id, "label": raw,
[...]

Issue #701, opened 2026-05-04
extract_svelte() regex fallback: alias paths skipped + synthetic IDs don't match real file nodes, so 100% of generated edges are dropped by build


jippi:
Follow-up to #692 (and the v0.7.1 fix). The regex fallback was added at `extract.py:1699-1732`, and a unit-level call confirms it does emit `dynamic_import` edges. But none of them survive into `graph.json`. Two independent bugs:

## Bug A — alias paths excluded by design

`extract.py:1712-1715`:

```python
for m in _re.finditer(r"""import\(\s*['"]([^'"]+)['"]\s*\)""", src):
    raw = m.group(1)
    if not raw.startswith("."):
        continue   # ← skips $lib/…, $partials/…, ~/, @/, src/…, etc.
```

Only relative `./X` paths are kept. The original ticket explicitly listed both relative *and* aliased examples — and in real SvelteKit codebases the aliased form dominates.

## Bug B — synthetic node IDs use the **source** file's path, so edges never resolve

`extract.py:1711, 1716`:

```python
file_node_id = _make_id(path.stem, str(path))
node_id      = _make_id(raw,        str(path))   # path = the importer
```

But the real file nodes elsewhere in the graph are created with `_make_id(str(path))` where `path` is the **target** file (relative, no stem prefix). So:

```
real target node id    : src_partials_series_grid_card_series_card_library_svelte
regex-pass target id   : card_series_card_library_svelte_users_jippi_…_src_partials_series_grid_series_single_view_svelte
real source file id    : src_partials_series_grid_series_single_view_svelte
regex-pass source id   : series_single_view_users_jippi_…_src_partials_series_grid_series_single_view_svelte
```

`build_from_json` in `build.py:99-104` tries normalized fallback, then drops the edge:

```python
if src not in node_set or tgt not in node_set:
    continue  # skip edges to external/stdlib nodes
```

Both endpoints are phantom, so every `dynamic_import` edge from the regex pass is discarded. The phantom nodes themselves leak through (I see one labelled `./Card/series-card-library.svelte` orbiting nothing), but no edge connects to the real target file.

## Repro

```python
>>> from graphify.extract import extract_svelte
>>> r = extract_svelte(Path('src/partials/Series/Grid/series-single-view.svelte'))
>>> [e for e in r['edges'] if e['relation'] == 'dynamic_import']
[{'source': 'series_single_view_…', 'target': 'card_series_card_library_svelte_…', ...}]   # produced
```

[...]

Issue #700, opened 2026-05-04
_read_tsconfig_aliases() silently fails on tsconfig.json with comments — every JS/TS project hits this


jippi:
Follow-up to #691 (and the v0.7.1 fix). The `extends` chain is now followed correctly, but the underlying parser still uses `json.loads`, which raises on `//` line comments and trailing commas — i.e. on the de-facto format every TypeScript tool ships and every framework starter generates. The exception is swallowed by the bare `except Exception: return {}` at `extract.py:62-63`, so `_load_tsconfig_aliases()` returns `{}` and every aliased import falls into the bare-package branch, exactly as it did pre-#691.

## Repro (real project — 1,870-file SvelteKit app)

`tsconfig.json` — the only thing that matters is one trailing-line comment, generated as-is by `pnpm create svelte` / `npm create vite` / `nx` / NestJS / T3 / Astro:

```jsonc
{
  "extends": "./.svelte-kit/tsconfig.json",
  "compilerOptions": {
    "skipLibCheck": true, // only way to make svelte-check not include node_modules
    "strict": true
  }
}
```

`.svelte-kit/tsconfig.json` (auto-generated by `svelte-kit sync`, has the actual paths):

```json
{
  "compilerOptions": {
    "paths": {
      "$lib": ["../src/lib"],
      "$lib/*": ["../src/lib/*"],
      "$partials": ["../src/partials"],
      "$partials/*": ["../src/partials/*"]
    }
  }
}
```

**Observed:**

```python
>>> from graphify.extract import _load_tsconfig_aliases
>>> _load_tsconfig_aliases(Path('src/lib/server'))
{}
```

The silent `except` at `extract.py:60-63` masks `json.JSONDecodeError: Expecting property name enclosed in double quotes: line 22 column 25`.

**Expected** (what you get if `json.loads` is monkey-patched to strip `//` comments first):

```
$lib       -> /…/website/.svelte-kit/../src/lib
$partials  -> /…/website/.svelte-kit/../src/partials
$app/types -> /…/website/.svelte-kit/types/index.d.ts
```

So the extends-following logic from #691 *does* work — it's just unreachable in practice because the parent file fails to parse.

## How aliases are used in this codebase

These are exactly the imports the resolver needs to handle. All standard SvelteKit shapes:

```ts
// src/hooks.server.ts (TS, top-level alias usage)
import { connect_to_db, disconnect_from_db } from '$lib/server/database'
[...]

PR #714, opened 2026-05-04
fix(svelte): parse <script> blocks so static imports produce edges (#713)

### LINKED ISSUE #713: extract_svelte() misses static `import X from './foo.svelte'` — script-block content not sliced before tree-sitter JS
## Summary

After #701 + #712, `dynamic_import` edges from `.svelte` files now resolve correctly. But **static imports** (the common case — `import Foo from './foo.svelte'` in a `<script>` block) are still effectively invisible to the graph. The root cause is that `extract_svelte()` hands the *entire raw `.svelte` file* (template + style + script) to tree-sitter's JavaScript grammar via `_extract_generic`, and the grammar chokes on the surrounding markup before reaching the script-block imports.

Net effect: questions like *"which components are unreferenced?"* or *"who imports this component?"* can't be answered from the graph — almost all `.svelte → .svelte` import edges are missing.

## Use case

Dead-component detection on a real SvelteKit codebase. The graph is supposed to make this trivial: list `.svelte` file nodes with zero incoming edges, exclude `+page.svelte` / `+layout.svelte` / `+error.svelte`, ship the rest as a dead-code candidate list.

I tried this on a 1,873-file SvelteKit app and got **214 candidates**. Spot-checking 8 of them via grep:

| Candidate (graph says "unreferenced") | Reality |
|---|---|
| `series-card-image.svelte` | Imported by 3 sibling files |
| `highlights.svelte` | Imported by `app-sidebar.svelte` + stats `+page.svelte` |
| `SeriesGrid.svelte` | Imported by 3 route pages |
| `LibraryEntryDialog.svelte` | Imported by 2 sibling cards |
| `app-sidebar.svelte` | Imported by `+layout.svelte` + command palette |
| `umami.svelte` | Imported by `change-locale.svelte` + others |
| `service-worker.svelte` | Imported by `+layout.svelte` |
| `tag-popover.svelte` | Imported by 5 sibling files |

8/8 false positives. The list is unusable for its intended purpose.

## Reproducer

`src/partials/Series/Grid/series-entry-grid.svelte` — first 16 lines:

```svelte
<script lang="ts">
    import { page } from '$app/state'
    import { Button } from '$lib/components/ui/button'
    import { resolve_titles } from '$lib/frontend/helpers/resolve-titles'
    import { getLibrarySeries } from '$lib/remote.functions/library.remote'
    import { type FullLibraryEntrySchema } from '$lib/remote.functions/library.types'
    import { BookPlus } from '@lucide/svelte'
    import SeriesCardAnime from './Card/series-card-anime.svelte'
    import SeriesCardDescription from './Card/series-card-description.svelte'
    import SeriesCardImage from './Card/series-card-image.svelte'
    import SeriesCardLibrary from './Card/series-card-library.svelte'
    import SeriesCardMeta from './Card/series-card-meta.svelte'
    import SeriesCardPublishers from './Card/series-card-publishers.svelte'
    import SeriesCardRatings from './Card/series-card-ratings.svelte'
    import SeriesCardRecommendationReason from './Card/series-card-recommendation-reason.svelte'
    import SeriesCardStaff from './Card/series-card-staff.svelte'
```

Nine static `.svelte` imports plus several `.ts` imports. After `graphify update .` on 0.7.5:

```python
import json, re
def make_id(p): return re.sub(r"[^a-zA-Z0-9]+", "_", p).strip("_").lower()
fid = make_id('src/partials/Series/Grid/series-entry-grid.svelte')
out = [e for e in g['links'] if e.get('source') == fid]
print(len(out), "outgoing edges")
# → 1 outgoing edges  (the dynamic_import for manage-library-entry-popover.svelte)
```

Only **one** outgoing edge — the `dynamic_import` from #701's regex pass. The 9 static `.svelte` imports and 6 `.ts` imports are all dropped.

## Graph-wide impact

```
Edge relation       Total    Targeting .svelte
imports             1348     4
imports_from        1652     351
dynamic_import      97       97
```

`imports` edges almost never reach `.svelte` targets. The 351 `imports_from` edges that *do* land are presumably from `.ts` files importing `.svelte` (or successful corner cases) — but the dominant `.svelte → .svelte` static-import flow is missing.

## Root cause

`graphify/extract.py` `extract_svelte()`:

```python
def extract_svelte(path: Path) -> dict:
    result = _extract_generic(path, _JS_CONFIG)   # ← whole .svelte file fed to tree-sitter JS
    ...
    src = path.read_text(...)
    for m in _re.finditer(r"""import\(\s*['"]([^'"]+)['"]\s*\)""", src):
        ...   # only handles dynamic_import via regex
```

`_JS_CONFIG` uses `tree_sitter_javascript`, which doesn't understand Svelte's `<script>...</script>` / `<style>...</style>` markup. The parser produces a degraded/truncated AST and the import-handler walk misses the script-block contents. The regex pass that #701 added covers `{#await import(...)}` only.

Two additional gaps with the current approach (relevant to the fix):

- `<script lang="ts">` is parsed with `_JS_CONFIG`, not `_TS_CONFIG`. TS-specific syntax (type imports, interfaces, generics, `as const`) silently degrades the AST.
- `<script context="module">` (Svelte 4) and `<script module>` (Svelte 5) blocks are not seen at all.

## Suggested fix

Slice every `<script>` block out first, dispatch each to JS or TS config based on its `lang=` attribute, then run the existing #701 regex pass over the **full** source so markup-level dynamic imports keep working.

```python
import re

_SCRIPT_RE = re.compile(
    r"<script\b(?P<attrs>[^>]*)>(?P<body>.*?)</script\s*>",
    re.DOTALL | re.IGNORECASE,
)
_LANG_RE = re.compile(r"""\blang\s*=\s*['"]?([a-zA-Z]+)""", re.IGNORECASE)


def _script_config(attrs: str) -> "LanguageConfig":
    """Pick TS config for lang=\"ts\" / lang=\"typescript\", else JS."""
    m = _LANG_RE.search(attrs)
    lang = (m.group(1).lower() if m else "")
    if lang in ("ts", "typescript"):
        return _TS_CONFIG
    return _JS_CONFIG


def extract_svelte(path: Path) -> dict:
    src = path.read_text(encoding="utf-8", errors="replace")
    result = {"nodes": [], "edges": []}

    # Path 1 (NEW): slice <script> blocks. finditer naturally handles
    # multiple blocks (e.g. instance + module-level). Pick TS or JS config
    # per-block based on lang= attribute. Line offsets are computed so
    # source_location stays correct relative to the .svelte file.
    for m in _SCRIPT_RE.finditer(src):
        config = _script_config(m.group("attrs") or "")
        line_offset = src[:m.start("body")].count("\n")
        partial = _extract_generic_from_string(
            m.group("body"), path, config, line_offset=line_offset
        )
        result["nodes"].extend(partial["nodes"])
        result["edges"].extend(partial["edges"])

    # Path 2 (UNCHANGED from #701): regex over the FULL .svelte source so
    # markup-level dynamic imports like `{#await import('./X.svelte')}`
    # remain captured.
    for m in _re.finditer(r"""import\(\s*['\"]([^'\"]+)['\"]\s*\)""", src):
        ...   # existing #701 logic, unchanged
```

The trickiest piece is that `_extract_generic` currently takes a `Path` and calls `path.read_text()` itself, so it would need a small refactor to accept a pre-read source string + a virtual path for ID stamping + an optional `line_offset`. Once that's in place, both the AST extractor and the dynamic-import regex pass produce edges that share the same file_nid, and downstream queries ("where is X used?") finally work for `.svelte → .svelte` flows.

### Coverage matrix

| Script block flavor | Behavior today | After fix |
|---|---|---|
| `<script>` (Svelte default — JS) | partial / fails | parsed with JS grammar |
| `<script lang="ts">` | silently degrades | parsed with TS grammar |
| `<script context="module">` (Svelte 4) | not seen | parsed with JS grammar |
| `<script context="module" lang="ts">` (Svelte 4 + TS) | not seen | parsed with TS grammar |
| `<script module>` (Svelte 5) | not seen | parsed with JS grammar |
| `<script module lang="ts">` (Svelte 5 + TS) | not seen | parsed with TS grammar |
| Instance + module blocks together (any combo of langs) | inst. degrades; module not seen | both parsed, each with its own config |
| `.svelte.ts` (Svelte 5 rune-only TS file, no `<script>`) | works (dispatch → `extract_js` + `_TS_CONFIG`) | unchanged — out of scope |

### Compatibility with #701 (markup dynamic imports)

The regex pass that #701 introduced for `{#await import('./X.svelte')}` is **not** narrowed by this change. It still runs over the entire `.svelte` source string; nothing about markup-level lazy loading regresses.

| Import shape | Lives in | Caught by | Status |
|---|---|---|---|
| `import X from './X.svelte'` | `<script>` | JS/TS AST (path 1) | NEW — fixed |
| `import X from './X.svelte'` | `<script module>` | JS/TS AST (path 1) | NEW — fixed |
| `const X = await import('./X.svelte')` | either script block | regex (path 2) | Same as today |
| `{#await import('./X.svelte') then Mod}` | markup | regex (path 2) | Same as today |

## Related

- #700 (tsconfig JSONC parsing — fixed)
- #701 (regex fallback alias resolution + synthetic IDs — fixed)
- #712 (stub-node `source_file` stamped to importer — open)

This is the natural next item: with #701 + #712 resolved, the `dynamic_import` codepath is solid, but the much larger static-import codepath needs the script-block slice to actually run.

## Environment

- graphifyy 0.7.5
- Python 3.14
- Real SvelteKit 2.59 app — 1,873 files, 968 `.svelte` nodes

jippi:
Fixes #713. Builds on #700 + #701; aware of #712 (separate PR forthcoming).

Code written by Claude Code, but reviewed by me (I'm not super familiar with Python, but SWE with +20 years experience)

## Problem

`extract_svelte()` handed the entire raw `.svelte` file (template + style + script) to tree-sitter's JavaScript grammar via `_extract_generic`. The grammar choked on the surrounding markup before reaching script-block imports, so static `import X from './foo.svelte'` produced **zero** edges.

On a real 1,873-file SvelteKit codebase: only **4 of 1,348** `imports` edges targeted `.svelte` files. Queries like *"where is this component used?"* or dead-component detection were unanswerable from the graph.

Three additional gaps with the original approach:
- `<script lang=\"ts\">` was parsed with `_JS_CONFIG`, silently degrading on TS-specific syntax (type imports, generics, `as const`).
- `<script context=\"module\">` (Svelte 4) and `<script module>` (Svelte 5) were never parsed.
- The #701 regex pass only handled dynamic imports, not static.

## Approach

Slice `<script>` blocks out and parse each with the right grammar (JS or TS, picked from the `lang=` attribute), then run the existing #701 regex pass over the **full** `.svelte` source so markup-level dynamic imports keep working unchanged.

Implementation uses a **whitespace-mask** trick rather than a string-slice + line-offset refactor: non-script content is replaced with spaces while preserving newlines, so tree-sitter parses only the script body but `node.start_point[0]` line numbers remain aligned with the original `.svelte` file. `_extract_generic` gains an optional `source_bytes=` kwarg; **all other extractors are untouched**.

## Coverage matrix

| Script block flavor | Behavior before | After |
|---|---|---|
| `<script>` | partial / fails | parsed with JS grammar |
| `<script lang=\"ts\">` | silently degrades | parsed with TS grammar |
| `<script context=\"module\">` (Svelte 4) | not seen | parsed with JS grammar |
| `<script context=\"module\" lang=\"ts\">` | not seen | parsed with TS grammar |
| `<script module>` (Svelte 5) | not seen | parsed with JS grammar |
| `<script module lang=\"ts\">` | not seen | parsed with TS grammar |
| Instance + module together (any langs) | inst. degrades; module not seen | both parsed, each with own config |
| `.svelte.ts` (Svelte 5 rune-only file) | works (`extract_js` + `_TS_CONFIG`) | unchanged — out of scope |

## #701 regression guards

The regex pass for `{#await import('./X.svelte')}` runs unchanged over the full source. Verified by:
- `test_markup_dynamic_import_still_extracted`
- `test_script_block_dynamic_import_still_extracted`

## Tests (25 new, all passing)

`tests/test_svelte.py` covers every cell in the coverage matrix plus the load-bearing invariants:

- **Lang detection** — default, ts, typescript, mixed attrs, unknown lang fallback
- **Whitespace-mask invariants** — newline count preserved, only target lang kept, markup/style stripped, **byte count exactly equals original**, **kept content stays at identical byte offsets**
- **Static-import resolution** — the #713 reproducer: 9 `.svelte` imports all produce edges
- **TS lang support** — `import type`, generics (`<T,>`) don't block subsequent imports
- **Svelte 4 + 5 module blocks** — both syntaxes extracted
- **Mixed-lang configs** — module-TS + instance-JS, and reverse
- **#701 regression guards** — markup `{#await import(...)}` and script-body `await import(...)`
- **Shape invariants** — file node always created, no dangling edge sources, file_nid matches `_make_id(str(path))` (#701 invariant)
- **Line/byte preservation** — function on L7 reports L7; deeply offset script on L31→fn on L32 reports L32; module-block fn and instance-block fn each report their original `.svelte` line; **newline byte positions match exactly between original and masked source**

## Validation on real codebase

Re-ran `graphify update .` on a 1,873-file SvelteKit app:

| Metric | Before | After | Δ |
|---|---|---|---|
| Total edges | 12,096 | 15,603 | **+29%** |
| `imports` edges → `.svelte` | 4 | 170 | **+42×** |
| `imports_from` edges → `.svelte` | 351 | 1,538 | **+4.4×** |
| `dynamic_import` edges (#701 check) | 97 | 97 | **unchanged** |
| Unreferenced-component candidates | 214 | 6 | **−97%** |

Spot-checked all 6 remaining candidates: 5 are genuinely dead components, 1 is a barrel-reexport limitation (`export { default as X } from './x.svelte'` in `index.ts`) that's orthogonal to this fix and worth a follow-up issue.

## Test plan

- [x] `pytest tests/test_svelte.py` — 25/25 pass
- [x] `pytest tests/` — 568 pass, 7 pre-existing failures unrelated to this change (SQL parser, incremental manifest, macOS case-insensitive `.F90`/`.f90` collision)
- [x] Smoke test on a real 1,873-file SvelteKit project — `.svelte → .svelte` import edges now appear; previously-broken \"where is X used?\" queries return correct results

### DIFF
--- a/graphify/extract.py
+++ b/graphify/extract.py
@@ -930,8 +930,16 @@ _SWIFT_CONFIG = LanguageConfig(
 
 # ── Generic extractor ─────────────────────────────────────────────────────────
 
-def _extract_generic(path: Path, config: LanguageConfig) -> dict:
-    """Generic AST extractor driven by LanguageConfig."""
+def _extract_generic(path: Path, config: LanguageConfig,
+                     *, source_bytes: bytes | None = None) -> dict:
+    """Generic AST extractor driven by LanguageConfig.
+
+    If source_bytes is provided, parses that instead of reading from path.
+    The path is still used for ID stamping, source_file metadata, and stem
+    derivation. This is what extract_svelte() uses to feed pre-sliced
+    <script> block content to tree-sitter while preserving file-level
+    identity (#713).
+    """
     try:
         mod = importlib.import_module(config.ts_module)
         from tree_sitter import Language, Parser
@@ -949,7 +957,7 @@ def _extract_generic(path: Path, config: LanguageConfig) -> dict:
 
     try:
         parser = Parser(language)
-        source = path.read_bytes()
+        source = source_bytes if source_bytes is not None else path.read_bytes()
         tree = parser.parse(source)
         root = tree.root_node
     except Exception as e:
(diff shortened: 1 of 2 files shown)


jippi:

PR #715, opened 2026-05-04
fix(svelte): stamp dynamic-import stub source_file to target, not importer (#712)

### LINKED ISSUE #712: extract_svelte() regex fallback stamps stub-node source_file to importer's path, corrupting target file metadata after #701

jippi:
> Code written by Claude Code, but reviewed by me (I'm not super familiar with Python, but SWE with +20 years experience).

Fixes #712. Independent of #714 — branched off `v7`, applies cleanly with or without #714 merged first.

## Problem

`extract_svelte()`'s regex pass for `import('...')` resolved a target path (relative or via tsconfig alias) and created a stub node for it — but stamped the stub's `source_file` to `str(path)`, which is the **importer's** path, not the target's.

`build_from_json` does last-write-wins on node attributes (`G.add_node` overwrites). When the target file is later extracted by `_extract_generic` on its own, both nodes share the same id (`_make_id(str(target))`) and merge. Whether the stub's wrong `source_file` or the real correct one wins depends purely on file-iteration order — non-deterministic.

**Effect:** downstream tools that read `source_file` off file nodes (display, *"where is X defined"*, blast-radius queries, community summaries) see the file as living inside whichever component first imported it. On a real 1,873-file SvelteKit codebase, 16 `.svelte` file nodes were corrupted this way — every dynamically-imported component sampled.

## Fix

Thread the resolved target path through to `source_file`:
- **relative imports** → `str(resolved)`
- **alias imports** (`$lib/...`) → `str(resolved_alias)`
- **bare/scoped externals** → leave `source_file` empty (we genuinely don't know where `node_modules` live, and stamping the importer would lie)

The stub **edge**'s `source_file` is unchanged — that legitimately is the importer (where the import statement lives). Edge vs node distinction is preserved.

## Tests

7 new tests in `tests/test_svelte_dynamic_import_stub.py`:

| Test | What it guards |
|---|---|
| `test_relative_dynamic_import_stub_points_to_target` | Stub source_file == target's path |
| `test_relative_dynamic_import_stub_does_not_carry_importer_path` | The literal #712 bug |
| `test_alias_dynamic_import_stub_points_to_resolved_path` | `$lib/...` aliases resolve correctly |
| `test_two_importers_agree_on_target_source_file` | Two stubs of same target are consistent → merge-order independent |
| `test_target_extraction_does_not_clobber_correct_source_file` | End-to-end: stub's source_file matches real extraction's source_file |
| `test_bare_module_dynamic_import_does_not_corrupt_source_file` | External modules don't get importer-path stamped |
| `test_dynamic_import_edge_source_file_is_importer` | Edge source_file unchanged (over-correction guard) |

**6 of 7 fail against unpatched v7** — confirming they're real regression guards. The 7th is the over-correction check that intentionally documents what should NOT change (edge source_file remains the importer).

## Test plan

- [x] `pytest tests/test_svelte_dynamic_import_stub.py` — 7/7 pass
- [x] Confirmed 6/7 fail against unpatched v7
- [x] `pytest tests/` — 556 pass, 7 pre-existing failures unrelated to this change

### DIFF
--- a/graphify/extract.py
+++ b/graphify/extract.py
@@ -1758,6 +1758,11 @@ def extract_svelte(path: Path) -> dict:
             raw = m.group(1)
             if not raw:
                 continue
+            # target_source_file: where the IMPORTED file actually lives.
+            # Used as the stub node's source_file so build_from_json's
+            # last-write-wins merge agrees with the real file's extraction.
+            # Empty for bare/scoped externals where we don't know the path (#712).
+            target_source_file = ""
             if raw.startswith("."):
                 # Relative import - resolve to full path so IDs match file node IDs.
                 resolved = Path(os.path.normpath(path.parent / raw))
@@ -1767,6 +1772,7 @@ def extract_svelte(path: Path) -> dict:
                 elif resolved.suffix == ".jsx":
                     resolved = resolved.with_suffix(".tsx")
                 node_id = _make_id(str(resolved))
+                target_source_file = str(resolved)
             else:
                 # Check tsconfig.json path aliases (e.g. "$lib/" -> "src/lib/", "@/" -> "src/")
                 # before treating as external. Mirrors _import_js logic so SvelteKit alias
@@ -1779,6 +1785,7 @@ def extract_svelte(path: Path) -> dict:
                         break
                 if resolved_alias is not None:
                     node_id = _make_id(str(resolved_alias))
+                    target_source_file = str(resolved_alias)
                 else:
                     # Bare/scoped import (node_modules) - use last segment;
                     # build_from_json drops as external if no matching node exists.
@@ -1786,6 +1793,9 @@ def extract_svelte(path: Path) -> dict:
                     if not module_name:
                         continue
                     node_id = _make_id(module_name)
+                    # target_source_file stays "" — we genuinely don't know
+                    # where externals live, and stamping the importer's path
+                    # would corrupt the node's metadata (#712).
             if node_id in existing_ids:
                 # Edge target already a real node - just add the edge, don't add a node.
                 result.setdefault("edges", []).append({
@@ -1796,7 +1806,7 @@ def extract_svelte(path: Path) -> dict:
                 continue
             result.setdefault("nodes", []).append({
                 "id": node_id, "label": raw,
-                "file_type": "code", "source_file": str(path),
+                "file_type": "code", "source_file": target_source_file,
                 "confidence": "EXTRACTED",
             })
             result.setdefault("edges", []).append({
--- a/tests/test_svelte_dynamic_import_stub.py
+++ b/tests/test_svelte_dynamic_import_stub.py
@@ -1,0 +1,227 @@
+"""Tests for #712 — extract_svelte()'s regex pass for `import('...')` must
+stamp the stub node's source_file to the *imported* file's path, never to
+the *importer's* path.
+
+build_from_json does last-write-wins on node attributes when two extractors
+produce nodes with the same ID. Before #712, the importer-stamped stub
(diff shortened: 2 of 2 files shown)


jippi:

Issue #436, opened 2026-04-18
_is_sensitive false-positives on directory names containing "secret"/"token"/"password" as substrings


dannigee:
## Summary                                                                                                                                          
                                                                                                                                                      
  `graphify.detect._is_sensitive()` uses substring matching on the generic secrets pattern, causing false positives on any file path that contains    
  `secret`, `token`, `password`, `credential`, `passwd`, or `private_key` as a **substring** of a larger word. This silently drops legitimate source
  files from the graph with no warning to the user.                                                                                                   
                                                   
  ## Version                                                                                                                                          
            
  graphify 0.4.23 (graphifyy on PyPI), macOS 26.4, Python 3.13                                                                                        
                                                              
  ## Concrete failure case                     
                          
  My repo is named `daniel-ai-secretary` and contains an `AISecretary/` directory. Every file path through those directories matches `(secret)` inside
   "**secret**ary".                                                                                                                                   
                                                                                                                                                      
  Result on my repo:
  - Without workaround: `graphify update .` reports `0 code files found` and fails                                                                    
  - With `Path('.')` workaround: build succeeds but **7,023 nodes from `AISecretary/` are silently missing** because per-file checks still flag
  `AISecretary/Services/Foo.swift` as sensitive                                                                                                       
                                                                                                                                                      
  Other plausibly-affected names: any repo/dir containing `tokenizer`, `tokenize`, `password_manager`, `credentials_ui`, `secretariat`, `tokenized_*`,
   etc.                                                                                                                                               
                                                                                                                                                      
  ## Root cause                                                                                                                                       
                                                                                                                                                      
  `detect.py` line 36:                         
                                                                                                                                                      
  ```python                  
  re.compile(r'(credential|secret|passwd|password|token|private_key)', re.IGNORECASE),
                                                                                                                                                      
  No word boundaries. Matches secret inside Secretary, token inside tokenizer, password inside password_manager (the last is probably desired — but
  the substring match also catches safe collisions).                                                                                                  
                             
  Proposed fix                                 
                                                                                                                                                      
  Add \b word boundaries to that single pattern:
                                                                                                                                                      
  re.compile(r'\b(credential|secret|passwd|password|token|private_key)\b', re.IGNORECASE),
                                                                                                                                                      
  The other _SENSITIVE_PATTERNS entries (dotfiles, key extensions, id_rsa family, etc.) already have sufficient specificity and don't need changes.
                                                                                                                                                      
  Additional observations
[...]

PR #719, opened 2026-05-04
fix(detect): exempt source-code files from generic-keyword secret filter

### LINKED ISSUE #718: _is_sensitive() still drops legitimate source files: word-boundary fix from #436 wasn't applied + keyword filter shouldn't apply to source code extensions
Follow-up to #436. The fix in commit \`4738e88\` removed the full-path check (`str(path)` → directory names no longer trip the filter), but the **filename** keyword regex was left unchanged and the **word-boundary fix originally proposed by the original reporter was never applied**. Two failure classes remain.

## Two remaining failure modes

`detect.py` line 36:

```python
re.compile(r'(credential|secret|passwd|password|token|private_key)', re.IGNORECASE),
```

This is still substring-matched against \`path.name\`. Combined with the fact that it applies to **all** files (regardless of extension), it silently drops these from `detect()` and therefore from the entire graph:

| Filename | Substring matched | Code or secret? |
|---|---|---|
| `password-reset.ts` | `password` | Code (email-template helper) |
| `AuthOauthAccessToken.model.ts` | `token` | Code (Sequelize model) |
| `test.search-tokenizer.ts` | `token` (in `tokenizer`) | Code |
| `password_manager.py` | `password` | Code (likely) |
| `JwtTokenValidator.java` | `token` | Code |

These are real cases from a 1,873-file SvelteKit codebase. There are no actual secrets in any of these files — they're source code that happens to mention auth/token/password concepts in their filenames.

## Why #436's fix didn't fully resolve this

the original reporter's original proposal was:

```python
re.compile(r'\b(credential|secret|passwd|password|token|private_key)\b', re.IGNORECASE),
```

The actual fix in `4738e88` was to remove `or p.search(full)` from `_is_sensitive`, so directory paths no longer trip the filter. The word-boundary refinement to the regex itself was not included.

Word boundaries would help with some cases (`tokenizer.ts`, `AuthOauthAccessToken.model.ts`, `daniel-ai-secretary`) but **not** filenames where the keyword IS a standalone word (`password-reset.ts`, `access-token.ts`, `jwt-token-validator.ts`).

## Root insight

The keyword filter is the only generic pattern in `_SENSITIVE_PATTERNS`. Every other entry is highly specific:

- `\.env|envrc` → extension/dotfile
- `\.(pem|key|p12|pfx|cert|crt|der|p8)$` → cryptographic key extensions
- `id_rsa|id_dsa|id_ecdsa|id_ed25519` → SSH key names
- `\.netrc|\.pgpass|\.htpasswd` → exact credential-store names
- `aws_credentials|gcloud_credentials|service.account` → cloud credential file names

These specific patterns target **data files** that store raw secrets. The keyword pattern is over-broad because it applies to **all** files including source code. Source code is, by definition, code — not credential storage. If a project commits secrets into `.ts`/`.py`/etc., that's a different security problem and graphify is the wrong line of defense.

## Suggested fix

Two changes in `_is_sensitive`:

```python
def _is_sensitive(path: Path) -> bool:
    """Return True if this file likely contains secrets and should be skipped."""
    # Skip the generic keyword check for known source-code extensions —
    # source files are code, not credential storage. The dotfile/cert/SSH-key
    # patterns still apply (a file like 'foo.pem' is sensitive even if it
    # accidentally has a code-like name).
    name = path.name
    is_code = path.suffix.lower() in CODE_EXTENSIONS or _shebang_file_type(path) == FileType.CODE
    for pattern in _SENSITIVE_PATTERNS:
        if pattern is _GENERIC_KEYWORD_PATTERN and is_code:
            continue
        if pattern.search(name):
            return True
    return False
```

Plus restructure `_SENSITIVE_PATTERNS` so the keyword pattern is identifiable (named constant or split into its own variable), and add word boundaries to it for the cases where it does apply:

```python
_GENERIC_KEYWORD_PATTERN = re.compile(
    r'\b(credential|secret|passwd|password|token|private_key)\b',
    re.IGNORECASE,
)
_SENSITIVE_PATTERNS = [
    # ... extension/exact-name patterns unchanged ...
    _GENERIC_KEYWORD_PATTERN,
    # ...
]
```

Combined effect:
- `tokenizer.ts` — passes (word boundaries reject `\btoken\b` inside `tokenizer`)
- `daniel-ai-secretary/foo.ts` — passes (already fixed by #436)
- `password-reset.ts` — passes (code extension, keyword filter skipped)
- `JwtTokenValidator.java` — passes (code extension)
- `secrets.json` — still flagged (data file + word-bounded match on `secret`)
- `database-credentials.yml` — still flagged (data file + word-bounded match)
- `.env`, `.env.local`, `id_rsa`, `aws.pem` — still flagged (specific patterns unchanged)

## Optional additional observation

the original reporter also flagged that **users get no warning when files are dropped as sensitive**. The skipped count is in `detected['skipped_sensitive']` but never surfaced to stdout. On a repo where the heuristic misfires, this is silent data loss — exactly what made our case hard to diagnose (had to compare `collect_files()` output to the manifest to notice the gap). Worth a separate enhancement: log a one-line warning at end of `detect()` when N or N% of files are skipped as sensitive, so users can audit.

## Reproducer

```python
from pathlib import Path
import tempfile

with tempfile.TemporaryDirectory() as d:
    d = Path(d)
    for name in ('password-reset.ts', 'AuthOauthAccessToken.model.ts',
                 'access-token.ts', 'JwtTokenValidator.java'):
        f = d / name
        f.write_text('export const x = 1')
        print(f"{name}: sensitive={_is_sensitive(f)}")

# Output (current behaviour):
#   password-reset.ts: sensitive=True
#   AuthOauthAccessToken.model.ts: sensitive=True
#   access-token.ts: sensitive=True
#   JwtTokenValidator.java: sensitive=True
```

All four are source code; none should be flagged.

## Real-world impact

On a 1,873-file SvelteKit codebase: 3 files silently dropped, including a `Sequelize` model that's referenced 30+ times across the codebase. Larger projects with auth/payment/identity domains will hit dozens.

## Environment

- graphifyy 0.7.5
- Python 3.14
- macOS

jippi:
> Code written by Claude Code, reviewed by me (I'm not super familiar with Python, but SWE with +20 years experience).

Fixes #718. Independent of the other open PRs (#714, #715, #717) — applies cleanly off `v7`.

## Problem

The sensitive-file heuristic in `_is_sensitive` has a generic keyword pattern that substring-matches every filename regardless of extension:

```python
re.compile(r'(credential|secret|passwd|password|token|private_key)', re.IGNORECASE),
```

This silently drops legitimate source-code files from the graph when their name mentions auth concepts:

| Filename | Real content |
|---|---|
| `password-reset.ts` | Email-template helper |
| `AuthOauthAccessToken.model.ts` | Sequelize model |
| `test.search-tokenizer.ts` | Test fixture |
| `JwtTokenValidator.java` | Auth code |
| `password_manager.py` | Manager class |
| `access-token.svelte` | UI component |

There are no actual secrets in any of these files — they're source code that happens to mention auth/token/password concepts in their filenames. They get silently dropped with no warning, exactly the failure mode that made this hard to diagnose (had to compare `collect_files()` output to the manifest to notice the gap).

## Background — relationship to #436

#436 (closed by `4738e88`) reported the same class of bug. The fix in `4738e88` removed the full-path check (`or p.search(full)`), so directory paths no longer trip the filter. The original issue's proposed word-boundary refinement to the keyword regex itself was not included, and the broader question of whether the keyword pattern should apply to source-code files at all was not addressed.

## Fix

Two complementary refinements to `graphify/detect.py`:

1. **Code-extension exemption.** The keyword pattern is now pulled into a named constant `_SENSITIVE_KEYWORD_PATTERN` and skipped when the file's extension is in `CODE_EXTENSIONS`. Source files are CODE, not credential storage. The structural patterns (`.env*`, `.pem`, `id_rsa`, `.netrc`, `aws_credentials`, etc.) continue to apply to *all* files, so a hypothetical `foo.ts.pem` is still flagged correctly.

2. **Word boundaries with optional plural.** `\b(credential|secret|...)s?\b` so substring matches inside larger words (`tokenizer`, `secretary`, `passwordless`, `credentialing`, `AccessToken`) no longer false-positive on data files. The `s?` covers plurals so canonical secret-storage filenames like `secrets.json` and `database-credentials.yml` remain flagged.

```python
_SENSITIVE_KEYWORD_PATTERN = re.compile(
    r'\b(credential|secret|passwd|password|token|private_key)s?\b',
    re.IGNORECASE,
)

def _is_sensitive(path: Path) -> bool:
    name = path.name
    is_code = path.suffix.lower() in CODE_EXTENSIONS
    for pattern in _SENSITIVE_PATTERNS:
        if pattern is _SENSITIVE_KEYWORD_PATTERN and is_code:
            continue
        if pattern.search(name):
            return True
    return False
```

## Coverage matrix

| File | Before | After |
|---|---|---|
| `password-reset.ts` (code) | dropped | indexed |
| `AuthOauthAccessToken.model.ts` (code) | dropped | indexed |
| `tokenizer.ts` (code, substring match) | dropped | indexed |
| `JwtTokenValidator.java` (code) | dropped | indexed |
| `secrets.json` (data, plural keyword) | flagged | flagged |
| `database-credentials.yml` (data, plural) | flagged | flagged |
| `password.txt` (data) | flagged | flagged |
| `secretary-notes.txt` (data, substring) | dropped (false positive) | indexed |
| `tokenizer-config.json` (data, substring) | dropped (false positive) | indexed |
| `.env`, `.env.local`, `.envrc` | flagged | flagged |
| `*.pem`, `*.key`, `*.p12`, `*.cert` | flagged | flagged |
| `id_rsa`, `id_ed25519.pub` | flagged | flagged |
| `.netrc`, `.pgpass`, `.htpasswd` | flagged | flagged |
| `aws_credentials`, `gcloud_credentials.json` | flagged | flagged |
| `foo.ts.pem` (mixed) | flagged | flagged (`.pem` is structural) |

## Tests (26 new, all passing)

`tests/test_is_sensitive.py`:

- **Source-code exemption (8):** legitimate `.ts`/`.py`/`.js`/`.java`/`.svelte`/`.d.ts` files with auth-related names pass through.
- **Specific-pattern preservation (12):** `.env`, `.env.local`, `.pem`, `.p12`, `.cert`, `id_rsa`, `id_ed25519.pub`, `.netrc`, `.pgpass`, `.htpasswd`, `aws_credentials`, `gcloud_credentials.json` still flagged.
- **Data-file keyword still flagged (4):** `secrets.json`, `database-credentials.yml`, `password.txt`, `api-token.json`.
- **Word-boundary correctness on data files (4):** substring-but-not-real-word cases pass through.
- **Mixed structural cases (2):** code-extension exemption is keyword-pattern-only and doesn't override structural patterns like `.pem` or `.env`.

13 of 26 fail on unpatched `v7` — real regression guards. The other 13 verify pre-existing flag behavior is preserved.

## Validation on a real codebase

A 1,873-file SvelteKit app:

- 3 source files previously silently dropped (`password-reset.ts`, `AuthOauthAccessToken.model.ts`, `test.search-tokenizer.ts`) are now extracted and contribute their outgoing import edges.
- Isolated `.ts` file count: 32 → 29.
- Total nodes: +6 (the new file nodes plus their per-symbol nodes).

## Test plan

- [x] All 26 new tests pass
- [x] 13 of 26 fail on unpatched `v7` (regression-guard quality)
- [x] Full test suite: 575 pass, 7 pre-existing failures unrelated
- [x] Smoke test on real 1,873-file SvelteKit project — 3 previously-dropped files now in graph

### DIFF
--- a/graphify/detect.py
+++ b/graphify/detect.py
@@ -29,11 +29,30 @@ CORPUS_WARN_THRESHOLD = 50_000    # words - below this, warn "you may not need a
 CORPUS_UPPER_THRESHOLD = 500_000  # words - above this, warn about token cost
 FILE_COUNT_UPPER = 200             # files - above this, warn about token cost
 
-# Files that may contain secrets - skip silently
+# Files that may contain secrets - skip silently.
+#
+# Keyword pattern is named separately so _is_sensitive() can opt source-code
+# files out of it: source files are CODE, not credential storage, so something
+# like 'password-reset.ts' or 'AuthOauthAccessToken.model.ts' should not be
+# silently dropped from the graph just because the filename mentions auth
+# concepts. Word boundaries also narrow the match so 'tokenizer' / 'secretary'
+# (substrings inside larger words) don't trip the filter even on data files.
+# The other patterns target real credential storage by extension or exact name
+# and continue to apply to ALL files regardless of code/data classification.
+# Optional `s?` after each keyword catches plurals (`secrets.json`,
+# `database-credentials.yml`) without re-introducing the substring-match
+# false positives the word boundaries are there to prevent. Trailing word
+# chars after the optional `s` ('tokenizer', 'credentialing', 'secretary',
+# 'passwordless') still fail the closing \b and pass the filter.
+_SENSITIVE_KEYWORD_PATTERN = re.compile(
+    r'\b(credential|secret|passwd|password|token|private_key)s?\b',
+    re.IGNORECASE,
+)
+
 _SENSITIVE_PATTERNS = [
     re.compile(r'(^|[\\/])\.(env|envrc)(\.|$)', re.IGNORECASE),
     re.compile(r'\.(pem|key|p12|pfx|cert|crt|der|p8)$', re.IGNORECASE),
-    re.compile(r'(credential|secret|passwd|password|token|private_key)', re.IGNORECASE),
+    _SENSITIVE_KEYWORD_PATTERN,
     re.compile(r'(id_rsa|id_dsa|id_ecdsa|id_ed25519)(\.pub)?$'),
     re.compile(r'(\.netrc|\.pgpass|\.htpasswd)$', re.IGNORECASE),
     re.compile(r'(aws_credentials|gcloud_credentials|service.account)', re.IGNORECASE),
@@ -59,9 +78,23 @@ _PAPER_SIGNAL_THRESHOLD = 3  # need at least this many signals to call it a pape
 
 
 def _is_sensitive(path: Path) -> bool:
-    """Return True if this file likely contains secrets and should be skipped."""
+    """Return True if this file likely contains secrets and should be skipped.
+
+    The generic keyword pattern (password/token/secret/...) is suppressed for
+    known source-code extensions so legitimate code files with auth-related
+    names ('password-reset.ts', 'AuthOauthAccessToken.model.ts',
+    'access-token.svelte', etc.) aren't silently dropped from the graph.
+    Specific patterns targeting real credential storage (.env, .pem, id_rsa,
+    .netrc, etc.) keep applying to all files.
+    """
     name = path.name
-    return any(p.search(name) for p in _SENSITIVE_PATTERNS)
+    is_code = path.suffix.lower() in CODE_EXTENSIONS
+    for pattern in _SENSITIVE_PATTERNS:
+        if pattern is _SENSITIVE_KEYWORD_PATTERN and is_code:
+            continue
+        if pattern.search(name):
+            return True
+    return False
 
 
 def _looks_like_paper(path: Path) -> bool:
--- a/tests/test_is_sensitive.py
+++ b/tests/test_is_sensitive.py
@@ -1,0 +1,193 @@
+"""Tests for _is_sensitive(): the file-skip heuristic that drops files
+expected to contain secrets. Two correctness properties:
+
+  1. Source-code files are never dropped by the generic keyword filter,
+     because source IS code, not credential storage. A file like
+     'password-reset.ts' is a code module — graphify should index it.
+
+  2. Specific patterns (.env, .pem, id_rsa, .netrc, etc.) keep applying
+     to all files, so a config or credential file with one of those
+     names is still flagged regardless of code/data classification.
+"""
+
+from pathlib import Path
+
+from graphify.detect import _is_sensitive
+
+
+def _f(tmp_path: Path, name: str) -> Path:
+    p = tmp_path / name
+    p.parent.mkdir(parents=True, exist_ok=True)
+    p.write_text("placeholder", encoding="utf-8")
+    return p
+
+
+# ── Source code is never dropped by the keyword filter ──────────────────────
+
+
+def test_password_in_ts_filename_not_sensitive(tmp_path):
+    """Real reproducer: 'password-reset.ts' is a TS module that handles the
+    user-facing password-reset flow. It contains email-template code, not
+    secrets. Must not be dropped."""
+    assert not _is_sensitive(_f(tmp_path, "password-reset.ts"))
+
+
+def test_token_in_ts_model_filename_not_sensitive(tmp_path):
+    """Real reproducer: 'AuthOauthAccessToken.model.ts' is a Sequelize
+    model definition. Code, not credentials."""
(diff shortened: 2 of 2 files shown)


jippi:

PR #720, opened 2026-05-05
Improve generated graph HTML explorer

DevStudio2k25:
## Summary

- Redesigns the generated `graph.html` explorer with a compact canvas-style layout.
- Adds a tabbed sidebar for node info, hubs, and community controls.
- Adds confidence/evidence filters for `EXTRACTED`, `INFERRED`, and `AMBIGUOUS` edges.
- Improves large graph usability with an explore-first seed view, bounded expansion, cached node/edge lookups, and a bottom command/search bar.
- Adds dark/light theme support while preserving the existing embedded, no-server HTML export model.

## Why

The previous HTML view loads every node and sidebar control into one dense panel, which becomes hard to scan on larger code graphs. This update keeps the same data contract and export path while making the output easier to navigate and less visually noisy.

## Validation

- `python -m py_compile graphify/export.py`
- `python -m pytest tests/test_export.py tests/test_pipeline.py -q`

I also ran the full local suite after installing CI extras. It collected and ran, with unrelated local/environment failures from a stale installed skill version, the Windows case-collided `sample.F90` fixture in my checkout, and local DNS resolution for `arxiv.org`.

## Notes

This PR only changes `graphify/export.py`. No generated `graphify-out` artifacts are included.

### DIFF
--- a/graphify/export.py
+++ b/graphify/export.py
@@ -51,52 +51,38 @@ def _viz_node_limit() -> int:
 
 
 def _html_styles() -> str:
-    return """<style>
-  * { box-sizing: border-box; margin: 0; padding: 0; }
-  body { background: #0f0f1a; color: #e0e0e0; font-family: -apple-system, BlinkMacSystemFont, "Segoe UI", sans-serif; display: flex; height: 100vh; overflow: hidden; }
-  #graph { flex: 1; }
-  #sidebar { width: 280px; background: #1a1a2e; border-left: 1px solid #2a2a4e; display: flex; flex-direction: column; overflow: hidden; }
-  #search-wrap { padding: 12px; border-bottom: 1px solid #2a2a4e; }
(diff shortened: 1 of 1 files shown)


safishamsi:
The HTML explorer redesign looks great conceptually — the tabbed sidebar, command bar, and evidence filters are all useful additions.

However the diff has a UTF-8 encoding problem that blocks the merge. Your editor saved the file in latin-1, which corrupted every non-ASCII character in `export.py` — including em-dashes and arrows in functions like `to_obsidian`, `safe_name`, and `to_svg` that your PR doesn't intend to touch. Merging as-is would silently break unrelated exporters.

To fix: re-create the changes from a **fresh checkout of `v7`**, make sure your editor is set to UTF-8, and restrict the diff strictly to the HTML explorer sections (`_html_styles`, `_html_script`, `to_html`). The unrelated `to_obsidian` / `to_svg` context lines in the current diff show the corruption clearly.

Happy to merge once the encoding is clean — the feature itself is worth having.

PR #723, opened 2026-05-05
Add support for custom LLM integration

the-cybersapien:
This pull request adds support for using any custom OpenAI-compatible LLM as a backend for semantic extraction, making it easier and potentially cheaper to run Graphify with alternative LLM providers. 
The implementation of kimi-2.6 is extended to support any custom backend, such as openrouter, qwen etc. As long as the endpoint is OpenAI compatible, graphify will support it. 
Env vars are updated to be specific to Graphify, to ensure it does not clash with any other running service. 

The documentation is updated with clear setup instructions, and the codebase is extended to detect and configure a "custom" LLM backend using environment variables.

**LLM backend support and configuration:**

* Added documentation in `README.md` explaining how to use Kimi 2.6 via Moonshot or any custom OpenAI-compatible LLM, including required environment variables and example configuration for providers like Openrouter.
* Introduced a "custom" backend in `graphify/llm.py` allowing users to specify their own OpenAI-compatible endpoint, model, and optional temperature via environment variables.
* Updated the backend detection logic in `graphify/llm.py` to check for a custom backend (via `GRAPHIFY_LLM_BASE_URL`) after Kimi and before Claude, enabling seamless selection based on available API keys.

### DIFF
--- a/README.md
+++ b/README.md
@@ -226,6 +226,33 @@ The MCP server gives your assistant structured access: `query_graph`, `get_node`
 
 ---
 
+## Using Kimi or a custom "cheaper" LLM instead of Claude
+
+Graphify can use **Kimi 2.6** (Via Moonshot) or any other openai-compatible model for semantic extraction. Kimi, for example is roughly 3x cheaper than claude per token.
+
+### Option A: Kimi 2.6 via Moonshot (recommended)
+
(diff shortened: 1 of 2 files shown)


the-cybersapien:
Closed in favor of #1041

PR #752, opened 2026-05-06
Add optional Google Workspace shortcut export

nyldn:
## Summary

- add opt-in Google Workspace shortcut export for `.gdoc`, `.gsheet`, and `.gslides`
- wire `graphify extract <path> --google-workspace` and `GRAPHIFY_GOOGLE_WORKSPACE=1` through detection and incremental extraction
- export shortcuts to Markdown sidecars under `graphify-out/converted/` using the authenticated `gws` CLI, with Sheets rendered through the new `[google]` extra
- document setup and add no-network tests for shortcut parsing, conversion, and detection behavior

## Why this helps

- Google Workspace shortcut files are only local pointers, so Graphify can otherwise see the filenames but not the real Doc, Sheet, or Slide content behind them.
- This makes native Google Workspace material usable as first-class corpus input: extracted entities, relationships, communities, wiki pages, and reports can be based on actual document content rather than shortcut metadata.
- Teams that already work in shared Drive folders get a repeatable path instead of manually exporting every Google file before each graph refresh.
- The integration is deliberately opt-in, so existing local-only extraction stays unchanged and there is no surprise auth or network access unless `--google-workspace` or `GRAPHIFY_GOOGLE_WORKSPACE=1` is set.
- Exports are written as sidecars under `graphify-out/converted/`, keeping source files untouched while still fitting the existing detection and incremental extraction flow.

## Tests

- `python3 -m py_compile graphify/google_workspace.py graphify/detect.py graphify/__main__.py tests/test_google_workspace.py tests/test_detect.py`
- `uv run --with pytest pytest tests/test_google_workspace.py tests/test_detect.py tests/test_extract.py tests/test_incremental.py tests/test_pipeline.py -q`
- `uv run python -m graphify --help | rg -n "google-workspace|extract <path>|Google Workspace"`
- `git diff --cached --check`


safishamsi:
Merged via PR #752. Google Workspace .gdoc/.gsheet/.gslides export via gws CLI. Account email pseudonymized with SHA256 hash — no PII in graph or version control. Install with pip install graphifyy[google]. Shipped in v0.7.9.

PR #753, opened 2026-05-06
fix: extract CommonJS require() imports as EXTRACTED edges

bionicbutterfly13:
## Summary

The JS/TS extractor only handled ES `import` statements. CommonJS `require()` calls produced zero import edges, which forced the call-graph pass to mark every cross-file call in CJS Node.js codebases as INFERRED — even when the binding was a top-of-file destructured `const { x } = require('./mod')`.

This patch wires three patterns into `_js_extra_walk` via a new `_require_imports_js` helper:

- `const { foo, bar: alias } = require('./mod')` → `imports_from` + per-symbol `imports`
- `const mod = require('./mod')` → `imports_from`
- `const x = require('./mod').y` → `imports_from` + symbol edge for `y`

Path resolution (relative / tsconfig alias / bare module) is factored out of `_import_js` into `_resolve_js_import_target` so ES and CJS share it.

The arrow-function path inside `_js_extra_walk` is preserved unchanged (regression test included).

## Why this matters

Node.js codebases that haven't migrated to ESM (build tooling, Express apps, CLI scripts, older monorepos) currently produce graphs where the EXTRACTED/INFERRED audit signal is wrong: every cross-file call collapses to INFERRED because no import edge ever pinned the binding to a target file. The honesty contract degrades to "trust nothing," which defeats one of graphify's core value props.

Validated against a real 92-file CJS Symphony orchestrator: the 5 previously-INFERRED edges from `runExecute()` (`loadFoundation`, `validateDispatchConfig`, `fetchSymphonyIssues`, `listSymphonyWorktrees`, `workspacePathForIssue`) all resolve to top-of-file destructured requires that the new handler picks up.

## Test plan

- [x] 4 new tests in `tests/test_extract.py`:
  - `test_extract_js_destructured_require_imports_from`
  - `test_extract_js_destructured_require_named_symbols`
  - `test_extract_js_member_require_emits_property_symbol`
  - `test_extract_js_arrow_function_still_extracted` (regression)
- [x] All 26 existing tests in `test_extract.py` pass
- [x] Confidence guarantee preserved: every new edge is `EXTRACTED` (passes `test_structural_edges_are_extracted`)

## Notes

- Does not touch ES `import` handling — `_import_js` still runs unchanged for `import_statement` nodes
- Symbol-level edges only emit when the require target resolves to a file in the corpus (so destructured names of bare `require('lodash')` won't pollute the graph with phantom symbol edges)
- TypeScript `import =` syntax is intentionally out of scope (it parses differently and is rare)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@ Full release notes with details on each version: [GitHub Releases](https://githu
 ## 0.7.8 (2026-05-06)
 
 - Fix: CommonJS `require()` imports now extracted from JS/TS -- `const { foo } = require('./mod')`, `const m = require('./mod')`, and `const x = require('./mod').y` all emit EXTRACTED `imports_from` (and per-symbol `imports`) edges. Previously CJS-only Node.js codebases produced AST graphs missing every import edge, which downgraded all cross-file calls to INFERRED.
+- Fix: cross-file `calls` edges are now promoted from INFERRED to EXTRACTED when the caller's file has an explicit `imports` or `imports_from` edge to the callee. Previously every cross-file call was unconditionally INFERRED, even when a top-of-file `import` / `require` proved the binding. On a 92-file CJS Node.js corpus this promoted 88% of cross-file calls (104 of 118) to EXTRACTED.
 - Feat: Gemini and OpenAI backends -- `graphify extract ./docs --backend gemini` (GEMINI_API_KEY / GOOGLE_API_KEY) or `--backend openai` (OPENAI_API_KEY); `[gemini]` and `[openai]` extras added (#735)
 - Feat: Groovy and Spock support -- `.groovy` and `.gradle` extracted via tree-sitter-groovy; Spock spec files (`def "feature"()` syntax) handled via regex fallback (#732)
 - Feat: Luau support -- `.luau` (Roblox Luau) added to code extraction using the Lua tree-sitter parser (#745)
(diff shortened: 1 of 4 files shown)


bionicbutterfly13:
Pushed a follow-up commit (2dd6ee6) that closes the loop end-to-end.

The first commit makes CJS imports EXTRACTED, but I noticed downstream `calls` edges those imports backed were *still* marked INFERRED — the cross-file call resolver in `extract()` unconditionally emitted INFERRED regardless of import evidence. So the import fix alone didn't actually improve the EXTRACTED/INFERRED audit signal that consumers rely on.

The new commit promotes cross-file `calls` to EXTRACTED when the caller's file has an `imports` (symbol) or `imports_from` (module) edge to the callee. INFERRED stays the default for genuine guesses — name-match-only calls without import backing are still INFERRED.

One subtlety: the promotion logic runs after the file-id remap at line 4736, so node `source_file` paths must be relativized before computing file_nids — otherwise the caller's file_nid (absolute-path-derived) wouldn't match the imports_from edge source (already remapped to relative). Test included.

Validated on the same 92-file corpus: the 5 previously-INFERRED `runExecute()` edges all resolve to EXTRACTED, and 88% of cross-file calls in the corpus (104 of 118) promote — INFERRED is now reserved for heuristic guesses, which is the original audit-trail intent.

Two new tests cover both branches (`with_import_evidence`, `without_import_evidence`). All 33 tests pass.

safishamsi:
Merged via PR #753. CommonJS require() imports now extracted as EXTRACTED edges. Shipped in v0.7.8.

PR #758, opened 2026-05-06
Fix GRAPHIFY_OUT default for query/path/explain

### LINKED ISSUE #756: bug: query/path/explain commands ignore GRAPHIFY_OUT env var, hardcode 'graphify-out/graph.json'
## Summary

The `query`, `path`, and `explain` subcommands hardcode the graph path as `"graphify-out/graph.json"` instead of using the `_GRAPHIFY_OUT` variable that is already defined at the top of `__main__.py` and correctly reads the `GRAPHIFY_OUT` environment variable.

## Affected code

`graphify/__main__.py`

```python
# Line 19 — correctly reads env var:
_GRAPHIFY_OUT = os.environ.get("GRAPHIFY_OUT", "graphify-out")

# Lines 1295, 1380, 1428 — hardcoded, ignores _GRAPHIFY_OUT:
graph_path = "graphify-out/graph.json"   # ← bug
```

All three occurrences are in the same pattern:
- `cmd == "query"` (line 1295)
- `cmd == "path"` (line 1380)
- `cmd == "explain"` (line 1428)

## Reproduction

```bash
# Set custom output dir
GRAPHIFY_OUT=".graphify" graphify query "How does module X connect to Y?"
# error: graph file not found: /path/to/project/graphify-out/graph.json
#        ↑ ignores GRAPHIFY_OUT, always looks in hardcoded "graphify-out/"
```

Even though the graph lives in `.graphify/graph.json`, the command fails because it never consults `_GRAPHIFY_OUT`.

## Root cause

`_GRAPHIFY_OUT` is used in other commands (e.g. `cluster-only`, `tree`, `watch`) but was not applied to `query`, `path`, and `explain`.

## Fix

Replace the three hardcoded defaults with `f"{_GRAPHIFY_OUT}/graph.json"`:

```diff
-        graph_path = "graphify-out/graph.json"
+        graph_path = f"{_GRAPHIFY_OUT}/graph.json"
```

Apply to all three occurrences: lines 1295 (`query`), 1380 (`path`), 1428 (`explain`).

The `--graph` override already works correctly — this only affects the default path when `--graph` is not supplied.

## Expected behavior

```bash
GRAPHIFY_OUT=".graphify" graphify query "..."   # reads .graphify/graph.json ✓
GRAPHIFY_OUT="custom-out" graphify path "A" "B" # reads custom-out/graph.json ✓
graphify explain "Foo"                           # falls back to graphify-out/graph.json ✓
```

## Environment

- graphifyy version: latest (pip)
- Python 3.12
- Linux

nyldn:
Fixes #756.

The three graph lookup commands already accept `--graph`, but their default path still pointed at `graphify-out/graph.json` even when `GRAPHIFY_OUT` was set. That breaks shared-output and non-default output directory setups.

This switches the default path for `query`, `path`, and `explain` to the existing `GRAPHIFY_OUT` setting, while leaving explicit `--graph` overrides alone.

Tests:
- `python3 -m py_compile graphify/__main__.py tests/test_cli_export.py`
- `uv run --with pytest pytest tests/test_cli_export.py -q`

### DIFF
--- a/graphify/__main__.py
+++ b/graphify/__main__.py
@@ -19,6 +19,10 @@ except Exception:
 _GRAPHIFY_OUT = os.environ.get("GRAPHIFY_OUT", "graphify-out")
 
 
+def _default_graph_path() -> str:
+    return str(Path(_GRAPHIFY_OUT) / "graph.json")
+
+
 def _check_skill_version(skill_dst: Path) -> None:
     """Warn if the installed skill is from an older graphify version."""
     version_file = skill_dst.parent / ".graphify_version"
@@ -1292,7 +1296,7 @@ def main() -> None:
         question = sys.argv[2]
         use_dfs = "--dfs" in sys.argv
         budget = 2000
-        graph_path = "graphify-out/graph.json"
+        graph_path = _default_graph_path()
         context_filters: list[str] = []
         args = sys.argv[3:]
         i = 0
@@ -1377,7 +1381,7 @@ def main() -> None:
         import networkx as _nx
         source_label = sys.argv[2]
         target_label = sys.argv[3]
-        graph_path = "graphify-out/graph.json"
+        graph_path = _default_graph_path()
         args = sys.argv[4:]
         for i, a in enumerate(args):
             if a == "--graph" and i + 1 < len(args):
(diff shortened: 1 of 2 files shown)


safishamsi:
Merged via PR #758. GRAPHIFY_OUT env var now respected in query/path/explain commands. Shipped in v0.7.8.

PR #761, opened 2026-05-07
feat: add .qmd file extension support

nueshym:
This PR adds support for [Quarto](https://quarto.org) markdown (`.qmd`) files by routing them through the existing `extract_markdown` pipeline.

Since `.qmd` files are largely compatible with standard Markdown, this change only adds `.qmd` support at the detection and dispatch layers.

Changes included in this PR:

* Add `.qmd` to document file extension detection
* Update export logic to properly sanitize `.qmd` filenames
* Add `.qmd` extractor dispatch using the existing markdown extractor
* Update watch comments to include `.qmd` files

Quarto supports executable fenced code blocks using language labels wrapped in curly braces:

````markdown
```{python}
# some code
```
````

The only notable difference is that Quarto executable code blocks use labels like `{python}`, so `extract_markdown` produces `"code: {python}"` instead of `"code: python"`. I believe this should be fine since the label is only used for display and the code block itself is still extracted correctly.


safishamsi:
Merged via PR #761. .qmd files now route through extract_markdown, added to DOC_EXTENSIONS, _DISPATCH, and both to_obsidian/to_obsidian_canvas safe_name regexes. Shipped in v0.7.9.

PR #772, opened 2026-05-07
Harden Google Workspace gws exports

nyldn:
## Summary

Follow-up hardening for the Google Workspace shortcut export added in #752. That implementation is working well in a real Google Drive-backed project, and this carries over the part of that local workflow that mattered for `gws` path validation.

- run `gws drive files export` from the sidecar output directory
- pass a relative `-o` filename instead of an absolute output path
- avoid sending Drive `resourceKey` as an unsupported `files.export` query parameter
- add tests for the subprocess contract and params shape

## Why

Current `gws` validates `--output` against its current working directory. Passing an absolute temp path can fail when Graphify extracts a corpus outside the shell cwd, even if the destination is inside the corpus output tree. Running from the sidecar directory with a relative output filename matches the pattern that has been working in a larger local Graphify workspace.

Drive resource keys are header-based (`X-Goog-Drive-Resource-Keys`), and `gws drive files export` does not currently expose a custom-header flag. This keeps the params aligned with the Drive export schema (`fileId`, `mimeType`) rather than implying resource-key support through an ignored/unsupported query parameter.

## Validation

- `python3 -m py_compile graphify/google_workspace.py graphify/detect.py graphify/__main__.py tests/test_google_workspace.py tests/test_detect.py`
- `uv run --with pytest pytest tests/test_google_workspace.py tests/test_detect.py tests/test_incremental.py tests/test_cli_export.py -q`
- local `gws drive files export --dry-run` from an output directory with relative `-o out.txt`; this passed path validation and failed only at local OAuth reauth, as expected on this machine
- `uv run graphify update .`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ Full release notes with details on each version: [GitHub Releases](https://githu
 - Feat: TypeScript extraction parity -- interface, enum, type alias, and module-level const nodes extracted; new_expression emits calls edges; parity with Java/C# class_types (#708)
 - Feat: Quarto (`.qmd`) file support -- routed through existing Markdown extractor; Quarto executable code blocks (` ```{python} `) extracted as code nodes (#761)
 - Feat: optional Google Workspace shortcut export for headless extraction -- `graphify extract ./docs --google-workspace` converts `.gdoc`, `.gsheet`, and `.gslides` files into Markdown sidecars with the `gws` CLI before semantic extraction; account email pseudonymized via SHA256 hash; `[google]` extra adds Sheets table rendering support (#752)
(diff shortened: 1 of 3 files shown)


nyldn:
Thanks again for merging #752 and shipping the Google Workspace shortcut support so quickly.

This is a small follow-up from using the same flow in a real Drive-backed project. The main change is to run `gws` from the sidecar output directory and pass a relative `-o` filename, which matches the current `gws` path validation behavior. I also tightened the resource-key handling so Graphify does not pass it as an unsupported `files.export` query param.

CI is green on 3.10 and 3.12. No rush, just flagging this as a small hardening PR when you have a moment. cc @safishamsi

PR #774, opened 2026-05-07
fix(skills): rewrite YAML descriptions from pipeline-only to trigger-oriented

capyBearista:
## Summary

10 skill-*.md files currently have descriptions that only describe what graphify does (input→pipeline→output), not when agents should use it. This means skills never load proactively on codebase questions. Agents only invoke `/graphify` when the user explicitly asks for a knowledge graph but not when they ask, "how does the auth flow work?".

Changed to hybrid descriptions that retain the pipeline summary but add trigger conditions, matching the pattern already used by `skill.md` (presumably for Claude Code only) and `skill-kiro.md`.

## Changes

- `graphify/...`
  - `skill-aider.md`
  - `skill-claw.md`
  - `skill-codex.md`
  - `skill-copilot.md`
  - `skill-droid.md`
  - `skill-opencode.md`
  - `skill-pi.md`
  - `skill-trae.md`
  - `skill-vscode.md`
  - `skill-windows.md`
  YAML description rewritten from pipeline-only to trigger-oriented

### DIFF
--- a/graphify/skill-aider.md
+++ b/graphify/skill-aider.md
@@ -1,6 +1,6 @@
 ---
 name: graphify
-description: any input (code, docs, papers, images) → knowledge graph → clustered communities → HTML + JSON + audit report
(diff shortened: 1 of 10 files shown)


capyBearista:
P.S., `skill-kiro.md` seems to have a formatting issue, according to GitHub at least.
image

Issue #722, opened 2026-05-05
Question — manifest.json absolute paths + missing graphify-out/.gitignore: intended or oversight?


gnrlbzik:
### Observation 1 — manifest.json paths

`save_manifest()` in `graphify/detect.py` (verified on 0.7.5, lines 760-771) writes file paths as dict keys (`manifest[f] = {...}`), where `f` originates from `os.walk` over the absolute root and is never relativized. Result: `graphify-out/manifest.json` contains entries like:

```json
{
  "/Users/<name>/Projects/<repo>/packages/types/src/portfolio.ts": {
    "mtime": 1771971357.10,
    "hash": "3743c6e8a0c40024bccbaa7499cf547d"
  }
}
```

Note: `graph.json` `source_file` paths were relativized in 0.5.0, but the same change wasn't applied to manifest.json.

### Observation 2 — no default .gitignore shipped

`graphify-out/` is created on first run but contains a mix of:

- **Shareable artifacts** — `graph.json`, `GRAPH_REPORT.md`, `wiki/`, `cost.json`, `graph.html`
- **Per-developer state** — all of which contain absolute paths or per-machine values:
  - `manifest.json` — incremental detect cache (`detect.py:save_manifest`, absolute path keys + mtime + hash)
  - `.graphify_root` — saved scan-root (absolute path; `__main__.py:1563`, `watch.py:133`)
  - `.graphify_incremental.json` — incremental detect cache (absolute path keys)
  - `.graphify_python` — venv interpreter path (absolute)
  - `.graphify_detect.json` — transient detect output cache (absolute paths; `__main__.py:1880`)

Consumers must manually figure out which files are local-state vs. shareable, or end up tracking the wrong set. We hit this — `manifest.json` was committed and started leaking absolute `/Users/<name>/...` paths across the team.

### Questions

1. Is keeping absolute paths in `manifest.json` intended (e.g. cross-machine portability not a goal because it's local cache only)?
2. If absolute paths aren't intended, would you accept a PR that relativizes `manifest.json` keys against the root, matching `graph.json`'s 0.5.0 behavior?
3. Would you consider shipping a default `.gitignore` (or README guidance) under `graphify-out/` listing which files are local-state and shouldn't be committed? With five separate per-machine files, this is non-obvious without reading the source.

Happy to open a PR for either if intent is "oversight, fix accepted."

### Environment
- graphifyy `0.7.5`
- Python 3.13
[...]

PR #778, opened 2026-05-08
fix: use project-relative paths in manifest.json and GRAPH_REPORT.md

luw2007:
## Summary

- `manifest.json` keys now store project-relative paths instead of absolute paths (e.g. `src/main.py` instead of `/Users/foo/project/src/main.py`)
- `GRAPH_REPORT.md` title uses the directory name instead of the full absolute path
- `detect_incremental` is backwards-compatible: matches both legacy absolute keys and new relative keys

## Motivation

Absolute paths in output files leak local filesystem structure and break portability — the same graph output produces different content on different machines, making it impossible to commit `graphify-out/` to version control or share across environments.

## Test plan

- [x] `tests/test_detect.py` — all 33 tests pass (including incremental + symlink tests)
- [x] `tests/test_report.py`, `tests/test_build.py`, `tests/test_ingest.py` — all pass
- [x] No new ruff lint errors introduced
- [x] Backwards compatible: existing manifests with absolute keys still work via fallback lookup

### DIFF
--- a/graphify/__main__.py
+++ b/graphify/__main__.py
@@ -1606,7 +1606,7 @@ def main() -> None:
         _commit = _gh()
         report = generate(G, communities, cohesion, labels, gods, surprises,
                           {"warning": "cluster-only mode — file stats not available"},
-                          tokens, str(watch_path), suggested_questions=questions,
+                          tokens, watch_path.name or str(watch_path), suggested_questions=questions,
                           min_community_size=min_community_size, built_at_commit=_commit)
         (out / "GRAPH_REPORT.md").write_text(report, encoding="utf-8")
         to_json(G, communities, str(out / "graph.json"))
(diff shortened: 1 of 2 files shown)


safishamsi (changes requested):
Good direction — manifest relativization in `detect.py` and `__main__.py` is correct and the backwards-compat reading of legacy absolute keys is well handled. Two things need fixing.

## Required changes

**1. `watch.py:148` not updated**

`save_manifest(detected["files"])` is called with no `root` argument. The new fallback resolves `root` to `Path.cwd()`, which only works when CWD matches the watch path. Running `graphify watch /some/other/path` from a different directory will still write absolute manifest keys. Fix: pass `root=watch_root` and `manifest_path=str(out / "manifest.json")` explicitly.

**2. No tests for the new behaviour**

No test asserts that saved manifest keys are relative, and none exercises the legacy-absolute-key backwards-compat path. A small addition to `tests/test_detect.py` covering both cases would lock in the contract cheaply.

## Minor (non-blocking)
- `watch_path.name or str(watch_path)` at `__main__.py:1609` returns a literal `"."` for `Path(".")` — reusing `_report_root_label` from `watch.py` would be cleaner
- Issue #722 will remain partially open after this PR (`.graphify_root`, `.graphify_incremental.json`, and the `.gitignore` guidance are not addressed) — worth noting in the PR description so #722 isn't closed prematurely

gnrlbzik:
@safishamsi @luw2007 can this pr also introduce gitignore template per #722

sumanthkumarc:
+1, the main reason why the /graphify-out output artifacts cannot be committed right now.

PR #780, opened 2026-05-08
Deterministic extraction, deduplication hardening, portability, and expanded test coverage

hypnwtykvmpr:
This roll-up bundles several independent lines of work that touch extraction reliability, graph determinism, cross-machine portability, and defense against edge-case regressions.

## Deterministic extraction pipeline

- A source-location pre-pass in deduplication prevents entities sharing the same (source_file, source_location) from being treated as independent, with token-level label containment to avoid wrongly merging same-line symbols.
- A prune pipeline removes stale references to nodes that no longer exist in the working tree after incremental rebuilds.
- Deterministic output: manifest.json uses sort_keys; cluster sorting is stabilized so `graphify update` produces repeatable output on unchanged inputs.

## Portability between machines

- Manifest keys are stored root-relative instead of absolute, so graphify-out/ directories committed to git remain usable when cloned on a different machine.
- Legacy absolute keys in manifests are normalized at load time, including macOS /var to /private/var symlink resolution.

## Edge-case hardening

Eight defects found through systematic review of the build, dedup, detect, extract, watch, and symbol_resolution modules were resolved:

- Stale nodes from deleted or modified source files no longer survive incremental graph merges.
- Watch rebuilds now filter preserved nodes whose source files have been removed.
- build_from_json guards against collapsing parallel edges with different relation types between the same endpoints.
- Intra-file call resolution detects ambiguous duplicate labels and prefers scope-qualified labels over bare ones.
- Cross-file call resolvers now track (source, target, relation) triples instead of only (source, target) pairs.
- Source-location dedup uses token-level rather than character-level substring containment.
- Incremental detection no longer skips content re-hashing when a file's mtime hasn't changed within the second-resolution filesystem clock.
- Normalized endpoint ID collision detection with overlap threshold and suffix disambiguation.

## Additional file format support

- .json and .jsonc are now indexed as document types alongside .yaml and .yml for document extraction.

## Test coverage

- Added unit tests spanning cache, dedup, export, ingest, llm_backends, security, tree_html, cluster, detect, manifest, and watch modules.
- pytest.importorskip guards added for optional dependency imports to keep the suite green on GitHub CI.

**69 files changed, +21,176 / -725 lines**


safishamsi (changes requested):
Thanks for the large rollup — there are real improvements here (manifest portability, dedup pre-pass, Leiden seed, community sort stability). However the PR as submitted has a few blockers and is too large to merge safely as a single unit.

## Blockers

**1. Breaking API change in `deduplicate_entities()`**

The return type changed from `(nodes, edges)` to `(nodes, edges, hyperedges, stats)` without being flagged. Any external caller (downstream scripts, integrations, forks) will break silently. This needs to be documented as a breaking change, and all call sites in the repo verified.

**2. Edge collapse fix is incomplete**

The "first relation wins" logic in `build_from_json` uses ordered `(src, tgt)` pairs. For an undirected `nx.Graph`, a `B→A` edge arriving after `A→B` with a different relation will be silently overwritten by NetworkX — the new collision counter won't catch it. The fix doesn't achieve determinism without guaranteed upstream edge ordering.

**3. Fork is inaccessible (404)**

The source repo is currently returning 404, which prevents full-file context verification for the review. Please ensure the branch is publicly accessible.

**4. No CI**

No checks ran on this PR. All other merged PRs in this repo have CI green before merge.

## Recommended: split into focused PRs

At 21,176 lines in a single commit this exceeds review tooling limits. Each of these is independently mergeable and reviewable:

- Portability: `detect.py` / `cache.py` / `watch.py` relative-path machinery
- Dedup pre-pass: `_source_location_dedup` + `prune_graph_references` in `dedup.py`  
- Build determinism: `build_from_json` edge collapse + `sort_keys`
- New modules: `symbol_resolution.py`, `scip_ingest.py`, `semantic_facts.py`, `deterministic_docs.py` — each deserves its own PR with rationale

Happy to merge each piece once it comes in separately with CI green.

PR #781, opened 2026-05-08
feat(extract): add Pascal/Delphi and Lazarus IDE support

simeonbodurov:
## Summary

  Adds full knowledge-graph extraction for Pascal/Delphi codebases and
  Lazarus IDE project files. Developed with Claude Sonnet 4.6.

  ---

  ### 1 — Pascal/Delphi extractor (`extract_pascal`)

  New extractor for `.pas`, `.pp`, `.dpr`, `.dpk`, `.inc` files via
  [tree-sitter-pascal](https://github.com/Isopod/tree-sitter-pascal).

  **Extracted nodes:** file, unit/program/library, class/interface/helper,
  procedure/function implementations.

  **Extracted edges:**
  - `file → contains → module`
  - `module → imports → dependency` (uses clause, resolved to path-based IDs)
  - `class → inherits → base class / interface`
  - `class/module → contains/method → procedure or function`
  - `procedure → calls → procedure` (in-file call resolution, including bare
    `Reset;`-style calls without parentheses)

  **Key design — import resolution:** Uses clause targets are resolved to
  path-based node IDs by scanning all Pascal files under the project root
  (`_pascal_project_root` + `_pascal_resolve_unit` helpers with caching).
  Without this, bare unit names like `SysUtils` resolve to IDs that never
  match any file node, making the entire import graph invisible.

  **Optional dependency:** `pip install tree-sitter-pascal`
  If not installed, `extract_pascal` returns `{"nodes":[], "edges":[], "error":...}`
  and the rest of the pipeline is unaffected.

  ---

  ### 2 — Inherits-edge fix

  The original `declType`/`typeref` handler built `inherits` edge targets
  with `_make_id(_read(child))` — just the bare class name. But class nodes
  use `_make_id(stem, type_name)`, so targets never matched, making the
  entire class hierarchy invisible in the graph.

  Fix adds `_pascal_class_stem_cache` and `_pascal_resolve_class()`: strips
  the conventional `T`/`I` prefix, locates the defining file by stem lookup
  (same cache mechanism as `_pascal_resolve_unit`), and returns the correct
  `_make_id(file_stem, class_name)` ID. RTL/unresolvable bases (e.g.
  `TObject`) fall back to `_make_id(bare_name)` with an explicit stub node,
  following the same pattern as the Python extractor.

  Also removes the `break` that stopped after the first `typeref`, so all
  parents of multi-inheritance declarations are captured.

  ---

  ### 3 — Lazarus form extractor (`extract_lazarus_form`, `.lfm`)

  `.lfm` files are text-based UI component trees. The extractor parses
  `object Name: TClassName ... end` blocks into a containment graph and
  captures `OnXxx = HandlerName` event bindings as `references` edges
  (context: `"event"`) linking each component to its handler procedure.

  ---

  ### 4 — Lazarus package extractor (`extract_lazarus_package`, `.lpk`)

  `.lpk` files are XML package definitions. The extractor reads the package
  name, required package dependencies (`→ imports` edges), and listed unit
  files (`→ contains` edges). Unit names are resolved to path-based node IDs
  via `_pascal_resolve_unit` so they link to the same nodes produced by
  `extract_pascal` on `.pas` files.

  ---

  ### Files changed

  | File | Change |
  |---|---|
  | `graphify/extract.py` | `extract_pascal`, `extract_lazarus_form`, `extract_lazarus_package`, resolution helpers |
  | `graphify/detect.py` | Add `.pas .pp .dpr .dpk .inc .lfm .lpk` to `CODE_EXTENSIONS` |
  | `tests/test_pascal.py` | 28 new tests covering all three extractors |
  | `tests/fixtures/sample.pas` | Representative Delphi fixture |
  | `tests/fixtures/sample.lfm` | Lazarus form fixture |
  | `tests/fixtures/sample.lpk` | Lazarus package fixture |

  ---

  ### Prerequisites

  pip install tree-sitter-pascal

  Wheel for Windows (cp38 abi3): available at
  https://github.com/Isopod/tree-sitter-pascal

  Co-Authored-By: Claude Sonnet 4.6 <noreply@anthropic.com>

### DIFF
--- a/graphify/detect.py
+++ b/graphify/detect.py
@@ -24,7 +24,7 @@ class FileType(str, Enum):
 
 _MANIFEST_PATH = "graphify-out/manifest.json"
 
-CODE_EXTENSIONS = {'.py', '.ts', '.js', '.jsx', '.tsx', '.mjs', '.ejs', '.go', '.rs', '.java', '.groovy', '.gradle', '.cpp', '.cc', '.cxx', '.c', '.h', '.hpp', '.rb', '.swift', '.kt', '.kts', '.cs', '.scala', '.php', '.lua', '.luau', '.toc', '.zig', '.ps1', '.ex', '.exs', '.m', '.mm', '.jl', '.vue', '.svelte', '.dart', '.v', '.sv', '.sql', '.r', '.f', '.F', '.f90', '.F90', '.f95', '.F95', '.f03', '.F03', '.f08', '.F08', '.pas', '.pp', '.dpr', '.dpk', '.inc', '.lfm', '.lpk'}
+CODE_EXTENSIONS = {'.py', '.ts', '.js', '.jsx', '.tsx', '.mjs', '.ejs', '.go', '.rs', '.java', '.groovy', '.gradle', '.cpp', '.cc', '.cxx', '.c', '.h', '.hpp', '.rb', '.swift', '.kt', '.kts', '.cs', '.scala', '.php', '.lua', '.luau', '.toc', '.zig', '.ps1', '.ex', '.exs', '.m', '.mm', '.jl', '.vue', '.svelte', '.dart', '.v', '.sv', '.sql', '.r', '.f', '.F', '.f90', '.F90', '.f95', '.F95', '.f03', '.F03', '.f08', '.F08', '.pas', '.pp', '.dpr', '.dpk', '.inc', '.dfm', '.lfm', '.lpk'}
 DOC_EXTENSIONS = {'.md', '.mdx', '.qmd', '.txt', '.rst', '.html', '.yaml', '.yml'}
 PAPER_EXTENSIONS = {'.pdf'}
 IMAGE_EXTENSIONS = {'.png', '.jpg', '.jpeg', '.gif', '.webp', '.svg'}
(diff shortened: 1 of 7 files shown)


safishamsi (changes requested):
Great PR overall — follows extractor conventions, solid test coverage (28 tests), and the inherits-edge fix is a genuine improvement. Just one thing needed before merge.

## Required: add `.lpr` extension

The PR title promises Lazarus IDE support but `.lpr` (the Lazarus program file — the entry point of every Lazarus project, identical Pascal syntax to `.dpr`) is missing from both `CODE_EXTENSIONS` in `detect.py` and `_DISPATCH` in `extract.py`. This is a one-line fix in each file, mapping `.lpr` to `extract_pascal`.

`.lpi` (Lazarus project info, XML format) can be skipped or added in a follow-up — that's a separate parser.

## Minor (non-blocking)
- `rstrip("()")` strips any combination of `(` and `)` chars rather than the literal suffix `"()"` — `removesuffix("()")` would be more precise (though benign in practice since labels are always `"Name()"`)
- Module-level `_pascal_unit_cache` / `_pascal_class_stem_cache` persist across `extract()` calls in the same process — could leak stale data if files move between runs. Low priority but worth a note.

Once `.lpr` is added this is a clean merge.

safishamsi:
Merged - thank you for your first OSS contribution! Pascal/Delphi/Lazarus support is now in v7. Great to hear graphify is useful for navigating your Delphi project.

PR #782, opened 2026-05-08
Add Windows cmd install skill and migrate Codex hook to UserPromptSubmit

PorunC:
## Summary

- Add a dedicated Windows cmd.exe install platform via `graphify install --platform cmd`
- Add and package `graphify/skill-cmd.md` with cmd-compatible command snippets
- Clarify README Windows install options for PowerShell vs cmd.exe
- Update Windows PowerShell skill snippets to avoid fragile inline quoting
- Move Codex graph guidance hook registration from `PreToolUse` to `UserPromptSubmit`
- Clean up legacy graphify-generated `PreToolUse` hooks without deleting unrelated user hooks

## Why

Windows users need separate skill instructions for PowerShell and cmd.exe because their quoting and heredoc behavior differ. This PR adds a cmd-specific install target while keeping PowerShell behavior under the existing `windows` platform.

The Codex hook migration is an intentional behavior change: graph context is now injected through `UserPromptSubmit` instead of `PreToolUse`, matching current hook expectations and making the prompt-time context behavior visible for the changelog. Old graphify-generated `PreToolUse` hooks are removed safely without deleting unrelated user hooks.

## Testing

```bash
PYTHONPATH=/home/misaka/wiki/graphify pytest tests/test_install.py tests/test_hooks.py -q --basetemp=/home/misaka/pytest-graphify
```

Result:
```
57 passed
```

Also ran:
```
graphify update .
```


safishamsi (changes requested):
Thanks for the PR — the install wiring, packaging, and Codex hook refactor are all solid and well-tested. One correctness issue needs fixing before merge.

## Required change: backslash-in-path bug in `skill-cmd.md`

Every Python payload in the skill is wrapped in `python -c "exec('...')"`. On Windows, user-substituted paths frequently contain backslash-letter sequences that Python re-interprets as escape sequences:

- `C:\Users\bob` — `\U` triggers `SyntaxError` (8 hex digits required)
- `C:\notes\todo` — `\n` becomes newline, `\t` becomes tab
- `C:\foo\bar` — `\b` becomes backspace

The PowerShell skill avoids this entirely via `@'...'@` raw heredoc. The cmd skill needs equivalent protection — options:
1. Use raw-string Python sources: `r'C:\Users\bob'` in the substitution instructions
2. Add an explicit instruction telling the model to normalise backslashes to forward slashes before substitution
3. Wrap the payload differently (e.g. write to a temp file and `python <file>`)

The `cmd.exe execution rule` section mentions escaped newlines and quotes but doesn't address the user-substitution backslash hazard.

## Minor (non-blocking)
- `.graphify_python` is written in Step 1 but never referenced in later steps (same gap exists in `skill-windows.md` — fine to leave for a follow-up)
- The Codex hook migration (`PreToolUse` → `UserPromptSubmit`) is a behaviour change worth mentioning in the PR title or description so it's visible in the changelog

PorunC:
Thanks for the review.

I pushed the required cmd.exe backslash-path fix. The cmd skill now explicitly tells agents to normalize user-substituted filesystem paths to forward slashes before inserting them into any `python -c "exec(...)"` payload, so raw Windows paths like `C:\Users\bob` are not reinterpreted by Python escapes. I also added regression coverage for that instruction in `tests/test_install.py`.

I updated the PR title/description to make the Codex hook migration from `PreToolUse` to `UserPromptSubmit` visible for the changelog.

I did not address the `.graphify_python` follow-up in this PR. Agreed that Step 1 writes it but later steps do not reuse it; since the same gap exists in `skill-windows.md`, I will handle the Windows/cmd interpreter reuse cleanup in a separate follow-up PR.

PorunC:
Update: commit b999446cd1ff88493298765b3aff3c829533c0de fixes the `.graphify_python` follow-up noted in review. Step 1 writes the interpreter path, and the later steps now reference it instead of falling back to an implicit Python command.

PR #784, opened 2026-05-08
[codex] Add OpenAI Responses fallback for mini models

nyldn:
## Summary

- add OpenAI Responses API routing for GPT-5.x / Codex model names when users explicitly select them with `--backend openai --model ...` or `GRAPHIFY_OPENAI_MODEL`
- add opt-in OpenAI model fallback via `GRAPHIFY_OPENAI_FALLBACK_MODELS`, preserving the existing `gpt-4.1-mini` default and Chat Completions path
- document mini-model examples such as `gpt-5.5-mini` with fallback to `gpt-5.4-mini,gpt-4.1-mini`

## Scope / duplication check

- This does not change backend detection priority or the existing OpenAI default model.
- I checked open upstream issues and PRs for `openai codex mini fallback gpt-5.5 gpt-5.4`; no matching open item came back.

## Validation

- `uv run --with pytest pytest tests/test_llm_backends.py tests/test_ollama.py -q` -> 18 passed
- `python3 -m compileall graphify/llm.py graphify/__main__.py` -> passed
- `uv run graphify --help | sed -n '/extract <path>/,+8p'` -> verified help text
- `graphify update .` -> completed code graph refresh
- `uv run --with pytest pytest -q` -> 667 passed, 6 failed in existing Fortran/SQL parser tests unrelated to this backend change (`test_fortran_capital_F_parses_preprocessed`, SQL table/view/function/edge assertions)

### DIFF
--- a/README.md
+++ b/README.md
@@ -299,6 +299,8 @@ graphify antigravity install / uninstall
 graphify extract ./docs                        # headless LLM extraction for CI (no IDE needed)
 graphify extract ./docs --backend gemini       # explicit backend: gemini, kimi, claude, openai, ollama, or bedrock
 graphify extract ./docs --backend gemini --model gemini-3.1-pro-preview
+graphify extract ./docs --backend openai --model gpt-5.5-mini
+GRAPHIFY_OPENAI_FALLBACK_MODELS=gpt-5.4-mini,gpt-4.1-mini graphify extract ./docs --backend openai --model gpt-5.5-mini
 graphify extract ./docs --backend ollama       # local Ollama (set OLLAMA_BASE_URL / OLLAMA_MODEL)
 graphify extract ./docs --backend bedrock      # AWS Bedrock via IAM - no API key, uses AWS credential chain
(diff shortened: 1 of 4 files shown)


safishamsi (changes requested):
Two issues to fix before merging:

**1. `_call_llm` not updated for Responses API routing**
`extract_files_direct` routes GPT-5/codex-mini to `_call_openai_responses`, but `_call_llm` (used by the dedup tiebreaker path) still routes everything through `_call_openai_compat`. GPT-5 users hitting the dedup path will get API errors. `_call_llm` needs the same `if model in RESPONSES_API_MODELS` branch.

**2. Replace `assert last_exc is not None` with an explicit raise**
```python
# instead of:
assert last_exc is not None
raise last_exc

# use:
if last_exc is None:
    raise RuntimeError("retry loop exited without exception")
raise last_exc
```
`assert` is silently removed under `python -O`, which would cause an `UnboundLocalError` on `last_exc` in optimized builds.

Fix these two and CI passes → ready to merge.

PR #788, opened 2026-05-09
fix(windows): unblock pipeline on cp1252 console + missing __main__ guards

nauman73:
## Summary

Three small Windows-compatibility fixes that all surface during the same first `/graphify` run on a Windows host.

### `graphify/benchmark.py` — `print_benchmark` Unicode crash on cp1252
`print_benchmark()` unconditionally prints U+2500 (box drawing) and U+2192 (rightwards arrow). On the legacy Windows console (cp1252) the second `print(...)` call raises `UnicodeEncodeError` mid-output:

```
File "C:\...\graphify\benchmark.py", line 121, in print_benchmark
    print(f"{'─' * 50}")
UnicodeEncodeError: 'charmap' codec can't encode characters in position 0-49: character maps to <undefined>
```

Adds a `_safe(unicode, ascii_fallback)` helper that detects whether the active stdout encoding can carry the glyph and falls back to ASCII (`-`, `->`) otherwise. Two regression tests cover both branches plus a full `print_benchmark` run against a strict-cp1252 stream.

### `graphify/extract.py` — opaque `BrokenProcessPool` failure on Windows-spawn
`ProcessPoolExecutor` on Windows uses `spawn`, so worker subprocesses re-import the calling `__main__`. When the caller is `python -c "..."` (or any script without `if __name__ == "__main__":`), the workers recursively spawn themselves and the pool dies. The user-visible failure is a ~290-line traceback ending in `BrokenProcessPool` — the actual cause (`RuntimeError: An attempt has been made to start a new process before the current process has finished its bootstrapping phase.`) is buried.

`_extract_parallel` now:
- catches `BrokenProcessPool`,
- prints a one-line warning that names the `__main__`-guard idiom and points at `parallel=False`,
- returns `False` so the public `extract()` routes to the existing `_extract_sequential` path.

Two regression tests:
- `_extract_parallel` returns `False` (not raises) when the pool breaks, and the warning includes both `BrokenProcessPool` and `__main__`.
- `extract()` calls `_extract_sequential` when `_extract_parallel` returns `False`, and still produces real nodes.

### `graphify/skill-windows.md` — quote-escaping minefield
Every `python -c "..."` block (30 in total) is replaced with a literal here-string + run + delete pattern:

```powershell
@'
<python source>
'@ | Out-File -FilePath .graphify_step_<name>.py -Encoding utf8
python .graphify_step_<name>.py
Remove-Item -ErrorAction SilentlyContinue .graphify_step_<name>.py
```

Why: any double-quote inside the Python source previously had to be backslash-escaped for the shell, and PowerShell's parser ate them inconsistently — failing on f-strings like `f'AST: {len(result["nodes"])} nodes'`. The new form passes Python source to disk literally, so what the model writes is what Python sees. The AST step's script template also gets an explicit `if __name__ == "__main__":` guard so multi-core extraction works on Windows even before the runtime fallback above kicks in. All 31 resulting heredoc blocks parse cleanly under `ast.parse`.

## Test plan
- [x] `pytest tests/test_benchmark.py tests/test_extract.py` — 48 passed (4 new tests)
- [x] `pytest -q --ignore=tests/test_languages.py --ignore=tests/test_multilang.py --ignore=tests/test_ollama.py` — 492 passed, no regressions
- [x] `ast.parse` smoke check across all 31 heredoc Python blocks in `skill-windows.md`
- [x] Verified end-to-end against a real corpus (357 files, 538K words, 4947 nodes / 12352 edges) using a local install before opening this PR

## Notes
- No public API change. `_extract_parallel` returns `bool` instead of `None` but it's underscore-prefixed.
- The skill rewrite touches only `skill-windows.md`. `skill.md` (POSIX) is unchanged.
- Discovered while running `/graphify` on a 357-file Angular/TypeScript codebase; both bugs blocked the pipeline and required local workarounds.

### DIFF
--- a/graphify/benchmark.py
+++ b/graphify/benchmark.py
@@ -1,12 +1,32 @@
 """Token-reduction benchmark - measures how much context graphify saves vs naive full-corpus approach."""
 from __future__ import annotations
 import json
+import sys
 from pathlib import Path
 import networkx as nx
 from networkx.readwrite import json_graph
 
 
 _CHARS_PER_TOKEN = 4  # standard approximation
+
+
+def _safe(unicode_char: str, ascii_fallback: str) -> str:
+    """Return unicode_char if stdout can encode it, else ascii_fallback.
+
+    Windows consoles often default to cp1252 which cannot encode box-drawing
+    or arrow glyphs; printing them raises UnicodeEncodeError mid-output.
+    """
+    encoding = getattr(sys.stdout, "encoding", None) or ""
+    try:
+        unicode_char.encode(encoding)
+        return unicode_char
+    except (UnicodeEncodeError, LookupError):
+        return ascii_fallback
+
+
+def _hr(width: int = 50) -> str:
+    """Horizontal rule that survives non-UTF-8 stdout (e.g. Windows cp1252 console)."""
+    return _safe("─", "-") * width
 
 
 def _estimate_tokens(text: str) -> int:
@@ -118,8 +138,9 @@ def print_benchmark(result: dict) -> None:
         return
 
     print(f"\ngraphify token reduction benchmark")
-    print(f"{'─' * 50}")
-    print(f"  Corpus:          {result['corpus_words']:,} words → ~{result['corpus_tokens']:,} tokens (naive)")
+    print(_hr(50))
+    arrow = _safe("→", "->")
+    print(f"  Corpus:          {result['corpus_words']:,} words {arrow} ~{result['corpus_tokens']:,} tokens (naive)")
     print(f"  Graph:           {result['nodes']:,} nodes, {result['edges']:,} edges")
     print(f"  Avg query cost:  ~{result['avg_query_tokens']:,} tokens")
     print(f"  Reduction:       {result['reduction_ratio']}x fewer tokens per query")
--- a/graphify/extract.py
+++ b/graphify/extract.py
@@ -4660,8 +4660,14 @@ def _extract_parallel(
     effective_root: Path,
     max_workers: int | None,
     total_files: int,
-) -> None:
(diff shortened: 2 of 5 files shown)


nauman73:
**Follow-up commit (5b2a79c):** auto-redirect `graphify install --platform claude` to the `windows` profile on win32. The `claude` profile shipped `skill.md`; on Windows that POSIX skill is unusable because of `python -c "..."` quote-escaping. Both profiles target the same destination, so the redirect is transparent — opt out with `GRAPHIFY_NO_AUTO_WINDOWS=1` (e.g. when invoking the Windows-side `graphify` from inside WSL).

4 new regression tests cover win32 auto-redirect, win32 opt-out, non-win32 unaffected, and explicit `--platform windows` ignoring the opt-out env var. `tests/test_install.py` now: 44 passed (was 40). Full suite: 496 passed, no regressions.

nauman73:
Dropped the follow-up commit `5b2a79c` (auto-redirect `claude` profile to `windows` flavor on win32). On reflection, the `claude` and `windows` install profiles are a maintainer-design call and the auto-redirect prejudges it — happy to revisit if you'd like that behavior, but pulling it out keeps the PR tightly scoped to the runtime/skill bugs I actually hit.

Branch is now at `2370ecd` (the original commit). PR contents unchanged from there.

safishamsi (approved):
**MERGE** - correct, narrowly scoped fixes.

The encoding change is encoding-driven (not platform-driven) which is the right call - `chardet` detection + `errors='replace'` handles cp1252/latin-1 on any OS, not just Windows. The `__main__` guard placement at the `ProcessPoolExecutor` call site is exactly right for spawn-based multiprocessing. Exception narrowing is good hygiene.

Tests run on all CI hosts. Merging.

PR #789, opened 2026-05-09
Recover from context-window-exceeded API errors in adaptive retry

tmaeder:
## Summary

`_extract_with_adaptive_retry` already handles output truncation (`finish_reason="length"`) by splitting a chunk in half and recursing. It does not handle the other shape of context overflow: the API **rejecting the prompt outright** with a 400 because the input plus `max_completion_tokens` doesn't fit in the model's context window.

Today that case bubbles out as an `Exception` from `extract_files_direct`, the chunk fails with no output, and `extract_corpus_parallel` logs `[graphify] chunk N/M failed: …` — even though the two halves of that chunk would fit cleanly.

This is the dominant failure mode for local OpenAI-compatible backends (LM Studio, llama.cpp, vLLM, Ollama), where the default context window is 4K-32K and a 60K-token chunk packed for cloud Kimi/Claude blows past it the moment you point graphify at a local URL.

## What changed

- Added `_looks_like_context_exceeded(exc)` — heuristic substring match on common error messages (`\"context size\"`, `\"context_length_exceeded\"`, `\"n_keep\"`, `\"prompt is too long\"`, etc.). Works across SDKs without depending on a specific exception class.
- `_extract_with_adaptive_retry` now wraps the `extract_files_direct` call. On a context-exceeded exception it splits the chunk in half and recurses, identically to the `finish_reason="length"` path.
- Single-file overflow returns an empty fragment instead of raising, so the rest of the corpus keeps running. Unrelated errors (rate limit, auth, network) still propagate unchanged.
- 4 new tests in `tests/test_llm_backends.py` covering: matcher accuracy, the split-recurse-and-merge path, single-file give-up, and unrelated-error pass-through.

## Why

Reproduced against qwen3.5-9b on LM Studio (32K ctx) extracting a 215-file corpus. With current `v7` 11/12 chunks aborted with `Error code: 400 - {'error': 'Context size has been exceeded.'}`. With this change overflowing chunks self-heal by splitting until the halves fit; only chunks where a *single file* still overflows are dropped, with a clear stderr warning.

The matching `finish_reason="length"` path that this mirrors has been in the codebase since the adaptive-retry module was introduced — this just extends the same recovery to the symmetric overflow signal.

## Test plan

- [x] `pytest tests/test_llm_backends.py -q` — 12 passed (4 new + 8 existing)
- [x] Manually verified against LM Studio + qwen3.5-9b on a 215-file mixed code/docs corpus

## Risk

Low. The retry only fires when `extract_files_direct` raises *and* the message matches one of the overflow markers. Misses (false negatives) preserve current behaviour — the chunk fails as it does today. False positives just trigger an extra split-retry and waste one round-trip, with the same recovery semantics already in place for output truncation.

### DIFF
--- a/graphify/llm.py
+++ b/graphify/llm.py
@@ -449,6 +449,35 @@ def _pack_chunks_by_tokens(
     return chunks
 
 
+_CONTEXT_EXCEEDED_MARKERS = (
+    "context size",
+    "context length",
+    "context_length",
+    "context window",
+    "n_keep",
+    "exceeds the available",
+    "n_ctx",
+    "maximum context",
+    "too many tokens",
+    "prompt is too long",
+    "context_length_exceeded",
+)
+
+
+def _looks_like_context_exceeded(exc: BaseException) -> bool:
+    """Heuristically classify an exception as a context-window overflow.
+
+    Different backends raise different exception types and messages for the
+    same underlying problem ("the prompt + max_completion_tokens did not fit
+    in the model's context window"). We match on substrings of the stringified
+    exception so the retry layer can recover without depending on a specific
+    SDK class. False positives are cheap (we'll re-extract on halves and
+    likely recover); false negatives are expensive (chunk fails entirely).
+    """
+    msg = str(exc).lower()
+    return any(marker in msg for marker in _CONTEXT_EXCEEDED_MARKERS)
+
+
 def _extract_with_adaptive_retry(
     chunk: list[Path],
     backend: str,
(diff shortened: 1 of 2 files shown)


safishamsi (approved):
**MERGE** - well-scoped, production-safe context-window retry.

The bisection approach is clean: exponential bisect until the chunk fits, bounded by `max_depth=6` so it never loops forever. `_CONTEXT_EXCEEDED_MARKERS` is a good extensibility point.

One follow-up for a separate PR: add Bedrock (`"input is too long"`) and Gemini (`"input token count exceeds"`) to `_CONTEXT_EXCEEDED_MARKERS` so multi-backend users get the same retry behavior.

Tests look solid. Merging.

PR #805, opened 2026-05-10
feat: improve generated graph HTML explorer

DevStudio2k25:
## Why

The previous HTML view loads every node and sidebar control into one dense panel, which becomes hard to scan on larger code graphs. This update keeps the same data contract and export path while making the output easier to navigate and less visually noisy.

## What changed

All changes are strictly confined to `graphify/export.py` — specifically `_html_styles()`, `_html_script()`, and `to_html()`. No generated `graphify-out/` artifacts are included.

### 4 new interactive features:

1. **Focus Mode** — Click any node to dim all unrelated nodes to 0.15 opacity, direct neighbors stay bright. Click background to clear.
2. **Glass Tooltips** — Custom HTML tooltips replace browser defaults. Shows node label, community, source file, and edge confidence on hover. Follows cursor, stays within viewport.
3. **Community Focus Button** — Target icon next to each group in the Groups tab. Click to auto-zoom the camera to that cluster. Auto-unhides hidden communities.
4. **Edge Labels on Hover** — Edge relation and confidence shown in the glass tooltip on hover. Labels hidden by default to keep the graph clean.

## Encoding

All non-ASCII characters in the file have been replaced with ASCII equivalents (`-`, `->`). File is 100% ASCII-clean.

### DIFF
--- a/graphify/export.py
+++ b/graphify/export.py
@@ -51,44 +51,43 @@ def _viz_node_limit() -> int:
 
 
 def _html_styles() -> str:
-    return """<style>
-  * { box-sizing: border-box; margin: 0; padding: 0; }
-  body { background: #0f0f1a; color: #e0e0e0; font-family: -apple-system, BlinkMacSystemFont, "Segoe UI", sans-serif; display: flex; height: 100vh; overflow: hidden; }
-  #graph { flex: 1; }
-  #sidebar { width: 280px; background: #1a1a2e; border-left: 1px solid #2a2a4e; display: flex; flex-direction: column; overflow: hidden; }
-  #search-wrap { padding: 12px; border-bottom: 1px solid #2a2a4e; }
-  #search { width: 100%; background: #0f0f1a; border: 1px solid #3a3a5e; color: #e0e0e0; padding: 7px 10px; border-radius: 6px; font-size: 13px; outline: none; }
-  #search:focus { border-color: #4E79A7; }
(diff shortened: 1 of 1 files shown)


safishamsi:
The new glass UI looks great in concept, but there's a breaking bug that prevents the graph from rendering at all — the PR needs one fix before we can merge.

**Breaking bug: script references DOM elements that don't exist in the body template**
The new `_html_script()` references these IDs/classes in `init()` and `bindControls()`:
`#mode-exploration`, `#mode-full`, `#fit-btn`, `#reset-btn`, `#labels-btn`, `#theme-btn`, `#community-search`, `#tooltip`, `#hub-list`, `#mNodes`, `#mEdges`, `#mVisible`, `#cCount`, `.tabBtn`, `.evidence-cb`, `[data-panel]`

None of these IDs or classes are present in the `to_html()` body template in `export.py` — the body only has `#graph`, `#sidebar`, `#search`, `#search-results`, `#info-content`, `#legend`, `#stats`. So `mNodes` is `null` and `init()` throws immediately — the loading screen stays up, the graph never renders, and `#tooltip` calls fail on hover.

The PR only modified `_html_styles()` and `_html_script()` but did not update the body HTML in `to_html()`. Please add the missing sidebar/panel/control elements to the body template.

**Minor security note**
`onclick="focusCommunity('+esc(c.cid)+')"` uses HTML-escape for a value injected into a JS string context. Today `cid` is always an integer so it's fine, but please use a JS-safe escape (or template literals) to be safe for future label types.

Once the body HTML is updated and the script can find its elements, we'll merge. Thanks!

PR #806, opened 2026-05-10
Update sample.F90

rafaelfiguereod-stack:


safishamsi:
Thanks for the fix! We landed this ourselves in commit 53da143 — renamed to `sample_preprocessed.F90` to avoid the macOS case-insensitive filesystem collision. Closing since it's already in main.

PR #807, opened 2026-05-10
skill: prevent AST/semantic ghost duplicates in extraction merge

hybridindie:
## Summary

Two related fixes to `graphify/skill.md` that prevent the subagent extraction phase from producing ghost-duplicate nodes that split one entity across two communities.

## The bug

The subagent prompt currently shows only `filestem_entityname` as a placeholder, with no rule for how to derive `filestem`. In practice subagents tend to use the bare filename (e.g., `session_validatetoken`) while the AST extractor uses a path-based stem (e.g., `src_auth_session_validatetoken`). The Part C merge dedupes by exact string, so the two forms coexist as separate nodes.

Symptoms observed in the wild on a real ~1500-node Python project:
- `audit_auditlogger` AND `comfyui_mcp_audit_auditlogger`
- `client_comfyuiclient` AND `comfyui_mcp_client_comfyuiclient`
- 67 such pairs total

The semantic side's INFERRED edges land on the ghost; the AST side's structural edges land on the canonical. Louvain clusters them into separate communities, splitting what should be one entity into two and producing misleading "cross-cutting hub" communities.

## Fix 1 — explicit stem rules in the subagent prompt

Adds a node-ID-format section right before the `Output exactly this JSON` line. Three rules:

1. **Stem is the relative path including parent directories**, not just the filename.
2. **Drop leading underscores from entity names** (e.g., `_helper` → `helper`) — the AST extractor does this, so the subagent must too. Without this rule, a function `_closure_values` becomes `tests_test_foo__closure_values` (double underscore) on the semantic side and `tests_test_foo_closure_values` (single underscore) on the AST side.
3. **Module-level constants and aliases**: AST only extracts functions and classes, but if the subagent adds a node for a constant/alias, it should still use the path-based stem so it joins the right community.

## Fix 2 — prefix-aware alias rewrite in Part C merge

Even with the new prompt, a subagent could still produce a shorter stem. Added a defensive alias-rewrite pass: if a semantic node's id has an AST counterpart ending in `_<sem_id>`, the ghost id is rewritten to the canonical AST id, and `sem`-side edges are routed through the alias map so they land on the canonical node.

This is purely additive — graphs that already had clean ids see zero behavior change because `_alias_for` returns the input unchanged when no candidate matches.

## Test plan

- [x] Manual run on a real Python project (~74 source files, 1,484 node graph): 67 ghost pairs detected and merged in one pass; community count dropped from 206 → 201 (the 5 phantom communities collapsed into the right ones); previously-stranded `rationale_for` edges from saved-query memories now land on the AST-canonical nodes.
- [x] No-op verification: re-running the alias logic against a graph with no ghosts produced 0 rewrites.
- Did not add a unit test in this PR — the merge logic is inline in the skill markdown, executed via `python3 -c`, not a callable function. Happy to add a test if you'd like to extract it into `graphify/extract.py` or similar; that's a larger refactor I didn't want to bundle here.

## What this does not change

- AST extractor behavior — untouched.
- Existing graphs — won't be retroactively cleaned. Users with already-built graphs would need to re-run extraction or run a one-off cleanup script. Happy to PR that separately if useful.
- The `seen` set semantics — `if new_id not in seen` is preserved; only the id resolution changed.

### DIFF
--- a/graphify/skill.md
+++ b/graphify/skill.md
@@ -229,6 +229,21 @@ confidence_score rules:
   Reasonable but not certain: 0.6-0.7. Weak inference: 0.4-0.5.
 - AMBIGUOUS edges: score 0.1-0.3
 
+Node ID format: lowercase, only `[a-z0-9_]`, no dots or slashes. Format: `{stem}_{entity}` where:
+- `stem` is the **relative path including parent directories**, with the file extension dropped and all non-alphanumeric chars (including `/`) replaced by `_`. Do NOT use just the filename — include the parent directories so the stem matches what the AST extractor produces.
+- `entity` is the symbol name, lowercased with non-alphanumerics replaced by `_`. Drop leading underscores from the symbol name (e.g., `_helper` → `helper`) — the AST extractor does this, so you must too.
+
+Examples (all using the relative-path stem, leading underscores dropped):
+- `src/auth/session.py` + `ValidateToken` → `src_auth_session_validatetoken`
+- `tests/test_foo.py` + `_helper` → `tests_test_foo_helper`
+- `lib/utils/helpers.py` + `parse_url` → `lib_utils_helpers_parse_url`
+
+This is critical: AST and semantic extractors must agree on the stem. If you use just the filename (e.g., `session_validatetoken` instead of `src_auth_session_validatetoken`), your nodes will become orphans alongside the AST-extracted versions and the merge step will not be able to dedupe them, splitting one entity into two communities.
+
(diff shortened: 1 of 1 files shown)


safishamsi:
Thanks for digging into this! The stem-ID divergence between AST and semantic extraction is a real problem. We're tracking it but plan to address it at the `extract.py` / ID normalization layer rather than in the skill prompt, so the fix is consistent regardless of how extraction is invoked. Closing this PR but keeping the issue open for tracking.

safishamsi:
Thanks for tracking this down! The underlying bug is real - subagents were using just the filename stem instead of parent_dir+stem, causing ghost duplicates. However, the node ID format in this PR is also incorrect (PR claims full-path stem like `src_auth_session_validatetoken`, but the AST extractor uses only the immediate parent dir: `auth_session_validatetoken`). We fixed skill.md directly in commit 7d1c109 with the correct format. Closing this PR.

PR #812, opened 2026-05-11
fix(skill): route subagent chunk writes through /tmp to survive Claude Code sandbox

### LINKED ISSUE #808: Claude Code subagents can't write to graphify-out/ (skill.md): chunks silently lost during /graphify --update
## Summary

When `/graphify --update` dispatches `general-purpose` Claude Code subagents for semantic extraction (Step B2), each subagent successfully extracts the JSON but **cannot persist** it because the bundled `skill.md` instructs them to write to `graphify-out/.graphify_chunk_NN.json` — a path outside Claude Code's `general-purpose` subagent sandbox allowlist.

The subagent does the work, then fails with "Write tool is denied" for that path. Step B3 then can't find the chunks, the merge gets empty input, and docs/images are silently dropped from the graph.

## Reproducer

1. `uv tool install graphifyy` (v0.6.7 confirmed)
2. `graphify install --platform claude` in a project
3. In Claude Code, run `/graphify --update` against a project with mixed code + docs/images
4. Step B2 dispatches `general-purpose` subagents per `skill.md` lines 351–368
5. Each subagent attempts `Write` to `<project>/graphify-out/.graphify_chunk_NN.json`
6. **Observed:** subagent reports "Write tool is denied". It returns gracefully after burning ~3 min of inference per chunk.
7. Step B3 reads `graphify-out/.graphify_chunk_*.json` → empty → `.graphify_semantic_new.json` has 0 nodes added → docs/images effectively skipped from the update.

## Root cause

Claude Code's `general-purpose` subagent runs in a sandbox that does NOT inherit the orchestrator's filesystem allowlist. As of 2026-05 the writable paths for a typical project layout are:

- The specific app/component subdirs the user has open (e.g. `apps/mobile`, `src/`)
- `<project>/.claude/`
- `/tmp`

Notably **NOT** writable from a `general-purpose` subagent:
- `<project>/graphify-out/`
- The project root itself

The bundled `skill.md` assumes the orchestrator's permissions propagate. They don't.

## Cost data point

A `/graphify --update` on a ~400-file Expo monorepo on 2026-05-08 lost **~16 min and ~290k tokens** across two retry cycles before the issue was diagnosed by reading the subagent's error output. Both subagents reported "Write tool is denied" after completing the extraction work in memory.

## Proposed fix

Route subagent chunk writes through `/tmp/graphify_chunk_NN.json` — universally writable across every Claude Code sandbox configuration I've tested. The orchestrator's merge step copies them into `graphify-out/` before the existing glob+merge logic, so cache semantics + the rest of the pipeline are unchanged.

Four small edits to `graphify/skill.md` (1437 lines → 1437 lines, net):

1. **Step B2 prompt template (~line 354)** — replace "Output exactly this JSON" with explicit "Write to `/tmp/graphify_chunk_NN.json`" instruction + sandbox rationale comment.
2. **Step B3 success-signal check (~line 361)** — change path from `graphify-out/.graphify_chunk_NN.json` to `/tmp/graphify_chunk_NN.json`.
3. **Step B3 merge step (~line 374)** — insert 4 lines that `shutil.copy` `/tmp/graphify_chunk_*.json` → `graphify-out/` before the existing glob (200+ lines unchanged).
4. **Step 9 cleanup (~line 838)** — also `rm -f /tmp/graphify_chunk_*.json`.

PR coming next with the exact diff.

## Tradeoffs of `/tmp`

- Universally writable across every Claude Code sandbox I've seen
- Auto-cleared on reboot (cleanup-friendly)
- Theoretical concurrency collision: two `/graphify` runs from different projects could clash on filenames. Mitigation: subdirectory-per-run if it ever bites in practice (it hasn't for me across ~30 update runs).
- Doesn't survive reboot mid-run. Acceptable — `/graphify --update` completes in <10 min.

## Why not project-level `additionalDirectories`

Per-project busywork that doesn't help users running graphify across multiple projects, and would silently regress for any user who hasn't applied it. The `/tmp` path is portable and works everywhere by default.

## Other affected platforms?

The bundled package has 12 platform-specific `skill-*.md` files. Only `skill.md` (Claude Code default) is confirmed affected by me. The fix likely applies to other platforms with similar subagent sandbox models — needs verification per-platform. I can scope this PR to `skill.md` only and let other contributors validate the parallel files.

jaimesada-jpg:
## Summary

Fixes #808. Claude Code's `general-purpose` subagent is sandboxed away from `graphify-out/` at the project root, so the Step B2 chunk write silently fails with "Write tool is denied". The extraction work completes in memory but the chunk is lost; Step B3's merge then reads zero files, and `/graphify --update` drops docs/images from the graph without surfacing the error.

This PR routes chunk writes through `/tmp/graphify_chunk_NN.json` — universally writable across every Claude Code sandbox configuration — then has the orchestrator copy them into `graphify-out/` before the existing glob+merge logic runs. Cache semantics and the rest of the pipeline are unchanged.

## What changed (4 small edits to `graphify/skill.md`)

| # | Where | Change |
|---|---|---|
| 1 | Step B2 instruction (~line 345) | Replace `Output exactly this JSON (no other text):` with explicit `Write to /tmp/graphify_chunk_NN.json` instruction + sandbox rationale comment |
| 2 | Step B3 success-signal check (~line 352) | Change path from `graphify-out/.graphify_chunk_NN.json` → `/tmp/graphify_chunk_NN.json` |
| 3 | Step B3 merge block (~line 365) | Insert 4 lines that `shutil.copy` `/tmp/graphify_chunk_*.json` → `graphify-out/` before the existing glob (200+ lines unchanged) |
| 4 | Step 9 cleanup (~line 681) | Also `rm -f /tmp/graphify_chunk_*.json` |

Net diff: +15 / -4.

## Reproducer & validation

Reproducer is in #808. Validation here:

- Applied the same patch locally to `~/.claude/skills/graphify/SKILL.md` on 2026-05-08
- Ran `/graphify --update` on a ~400-file Expo monorepo with mixed code + docs + images
- **Before patch:** 2 subagents dispatched, both reported "Write tool is denied", chunks lost, ~16 min and ~290k tokens burned across two retries
- **After patch:** 2 subagents wrote to `/tmp` cleanly, orchestrator copied to `graphify-out/`, merge proceeded normally. Graph grew from 901 → 1057 nodes; 156 new doc/image-derived nodes
- No regression on subsequent `--update` runs (~6 runs to date)

## Tradeoffs of `/tmp`

- ✅ Universally writable across every Claude Code sandbox I've encountered
- ✅ Auto-cleared on reboot (cleanup-friendly)
- ⚠️ Theoretical concurrency collision: two `/graphify` runs from different projects could clash on filenames. Hasn't happened in practice; mitigation if it ever bites is `/tmp/graphify_<run_id>/chunk_NN.json`. Punted as YAGNI for now.
- ⚠️ Doesn't survive reboot mid-run. Acceptable — `/graphify --update` completes in <10 min.

## Scope

This PR only touches `graphify/skill.md` (Claude Code default). The package ships 12 platform-specific `skill-*.md` variants. Several likely share the same sandbox issue, but I haven't validated on other platforms — left for follow-up PRs by people running those stacks. The fix pattern is identical wherever the bug applies.

## Why not project-level `additionalDirectories`

Per-project busywork that doesn't help users running graphify across multiple projects, and silently regresses for anyone who hasn't applied it. The `/tmp` path is portable and works everywhere by default.

---

### DIFF
--- a/graphify/skill.md
+++ b/graphify/skill.md
@@ -342,16 +342,20 @@ confidence_score is REQUIRED on every edge - never omit it, never use 0.5 as a d
 
 Node ID format: lowercase, only `[a-z0-9_]`, no dots or slashes. Format: `{stem}_{entity}` where stem is the filename without extension and entity is the symbol name, both normalized (lowercase, non-alphanumeric chars replaced with `_`). Example: `src/auth/session.py` + `ValidateToken` → `session_validatetoken`. This must match the ID the AST extractor generates so cross-references between code and semantic nodes connect correctly. CRITICAL: never append chunk numbers, sequence numbers, or any suffix to an ID (no `_c1`, `_c2`, `_chunk2`, etc.). IDs must be deterministic from the label alone — the same entity must always produce the same ID regardless of which chunk processes it.
 
-Output exactly this JSON (no other text):
+**Output path:** Write the JSON to `/tmp/graphify_chunk_NN.json` (where NN is the chunk number, zero-padded — e.g. `/tmp/graphify_chunk_01.json`). The Claude Code general-purpose subagent's sandbox typically allows writes to `/tmp` but BLOCKS writes to project subdirectories like `graphify-out/`. Writing to `/tmp` is the portable path. The orchestrator (parent /graphify run) reads chunks from `/tmp/graphify_chunk_*.json` in Step B3.
+
+Use the Write tool with that exact `/tmp/graphify_chunk_NN.json` path. Then return a one-line confirmation to the orchestrator: "chunk_NN written: X nodes, Y edges, Z hyperedges". Nothing else.
+
+The JSON file content schema:
 {"nodes":[{"id":"session_validatetoken","label":"Human Readable Name","file_type":"code|document|paper|image|rationale","source_file":"relative/path","source_location":null,"source_url":null,"captured_at":null,"author":null,"contributor":null}],"edges":[{"source":"node_id","target":"node_id","relation":"calls|implements|references|cites|conceptually_related_to|shares_data_with|semantically_similar_to|rationale_for","confidence":"EXTRACTED|INFERRED|AMBIGUOUS","confidence_score":1.0,"source_file":"relative/path","source_location":null,"weight":1.0}],"hyperedges":[{"id":"snake_case_id","label":"Human Readable Label","nodes":["node_id1","node_id2","node_id3"],"relation":"participate_in|implement|form","confidence":"EXTRACTED|INFERRED","confidence_score":0.75,"source_file":"relative/path"}],"input_tokens":0,"output_tokens":0}
 ```
 
 **Step B3 - Collect, cache, and merge**
 
 Wait for all subagents. For each result:
-- Check that `graphify-out/.graphify_chunk_NN.json` exists on disk — this is the success signal
+- Check that `/tmp/graphify_chunk_NN.json` exists on disk — this is the success signal
 - If the file exists and contains valid JSON with `nodes` and `edges`, include it and save to cache
-- If the file is missing, the subagent was likely dispatched as read-only (Explore type) — print a warning: "chunk N missing from disk — subagent may have been read-only. Re-run with general-purpose agent." Do not silently skip.
+- If the file is missing, the subagent was either dispatched as read-only (Explore type) OR failed silently — print a warning: "chunk N missing from /tmp — subagent may have been read-only or hit a sandbox restriction. Re-run with general-purpose agent." Do not silently skip.
 - If a subagent failed or returned invalid JSON, print a warning and skip that chunk - do not abort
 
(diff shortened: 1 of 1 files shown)


jaimesada-jpg:
Thanks @safishamsi — saw v0.7.14 landed with the absolute-CHUNK_PATH approach. Closing this PR since #808 is addressed your way.

For posterity / future readers: my /tmp routing was a workaround for what I'd diagnosed as a Claude Code sandbox restriction on writes to `graphify-out/` (I was hitting `Write tool denied` errors from `general-purpose` subagents on a project under `/Users/jaimegsada/Code/...`). If the real root cause was undefined subagent cwd rather than sandbox blocking, your fix is cleaner. Will validate on my end after upgrading to v0.7.15; if I still hit denials I'll re-open with new evidence.

PR #816, opened 2026-05-11
Add Clojure extraction support

alendit:
## Summary

Adds tree-sitter backed Clojure .clj extraction to Graphify.

- extracts namespaces, require/import edges, common top-level definitions, protocol methods, and same-file calls
- wires .clj through detection, collection, watch, hooks, docs, and skill metadata
- adds fixture-backed regression coverage for bare requires and quoted/commented non-calls

## Validation

- .venv/bin/python3 -m pytest -q -> 302 passed, 2 warnings
- git diff --check -> clean

### DIFF
--- a/README.md
+++ b/README.md
@@ -194,7 +194,7 @@ To remove graphify from all platforms at once: `graphify uninstall` (add `--purg
 
 | Type | Extensions |
 |------|-----------|
-| Code (31 languages) | `.py .ts .js .jsx .tsx .mjs .go .rs .java .c .cpp .h .hpp .rb .cs .kt .scala .php .swift .lua .luau .zig .ps1 .ex .exs .m .mm .jl .vue .svelte .astro .groovy .gradle .dart .v .sv .sql .f .f90 .f95 .f03 .f08 .pas .pp .dpr .dpk .lpr .inc .dfm .lfm .lpk .sh .bash .json` |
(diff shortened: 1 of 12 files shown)


safishamsi:
Solid Clojure extractor — the namespace/defs/require/call extraction is well-structured and the quoted-form skipping is a nice touch. Two things to resolve before we merge:

**1. Merge conflict**
The PR currently has `mergeStateStatus: CONFLICTING` — it was branched from an older `v7` base. Please rebase onto current `main` (v0.7.19) and resolve any conflicts.

**2. Missing test helper (verify or add)**
Several of the new tests call `_edges_with_relation()` (e.g. `test_clojure_import_edges_have_import_context`, `test_clojure_call_edges_have_call_context`). Please confirm this helper exists in `tests/test_languages.py` — if not, add it. Without it those tests will `NameError` at collection time.

**Minor**
`label_to_nid` uses `label.lower()` which silently collapses Clojure's case-sensitive identifiers (`Foo` and `foo` resolve to the same target). Low impact for now, but worth a follow-up issue.

Once rebased and the test helper is confirmed, we'll merge. Thanks!

alendit:
Hi @safishamsi when you say "current `main`" do you mean the `main` branch? Because it sort of looks stale to me and v8 is now the default. Tbh I'm a bit confused by the branching in this project. My best reading is to actually rebase the change on the `v8` branch, I'll do that, tell we if it's wrong.

alendit:
Confirming that the full run passes

```
uv run --extra sql pytest tests/test_languages.py tests/test_detect.py tests/test_multilang.py tests/test_watch.py -> 254 passed
```

alendit:
Fixed the lower-case issue as suggested.

lucasrim:
Would love to have this feature added!

kagel:
Would also love to have this feature added!

alendit:
@safishamsi anything else missing for this to be merged?

falberto:
Would also love to have this feature added!

VictorCovalski:
Would also love to have this feature added!

PR #823, opened 2026-05-11
Fix Fortran fixture case-collision on macOS

FatahChan:
## Why this change
`tests/fixtures/sample.F90` and `tests/fixtures/sample.f90` differ only by case. On case-insensitive filesystems (default macOS), those paths collide to the same on-disk file, which causes persistent dirty state and fixture content churn even after undo/restore operations.

That collision made local workflows noisy and brittle (checkout/rebase/test runs can repeatedly re-dirty fixtures with no intentional code change).

## Summary
- replace the case-colliding fixture path `tests/fixtures/sample.F90` with `tests/fixtures/sample_preprocessed.F90`
- update the Fortran preprocessed fixture test to point to the new path
- keep fixture contents unchanged while avoiding repeated churn on case-insensitive filesystems

## Test plan
- [x] `uv run --with pytest pytest tests/test_languages.py::test_fortran_finds_module tests/test_languages.py::test_fortran_capital_F_parses_preprocessed`
- [x] `uv run --with pytest pytest tests/test_languages.py`


safishamsi:
Applied directly in commit `53da143` — thanks @FatahChan! Clean fix, no other references to `sample.F90` were missed.

Note: `test_fortran_capital_F_parses_preprocessed` still fails in our CI environment because `tree-sitter-fortran` isn't installed there — that's pre-existing and unrelated to this rename. The fix is correct on macOS where `sample.f90` and `sample.F90` collide on a case-insensitive filesystem.

PR #838, opened 2026-05-12
fix #837: populate .graphify_detect.json in --update flow

### LINKED ISSUE #837: bug: --update shrinks manifest.json to changed-subset only, causing next --update to flag every unchanged file as new
**Graphify version:** 0.4.11 (pipx via miniforge), macOS 14, Python 3.12

## Symptom

After a `/graphify <path>` (full rebuild), every subsequent `/graphify --update` reports a manifest that contains only the files seen in the *prior* `--update` run — not the cumulative corpus. The next `--update` then re-flags every other file in the corpus as "new", even though their content (and the existing `graph.json`) is unchanged.

Concrete reproducer in my corpus (a mixed wiki+code+raw archive, ~1028 files):

1. Full rebuild → `manifest.json` correctly contains all ~1028 files (Step 9 saves it from `.graphify_detect.json`).
2. Touch ~20 wiki entities; run `/graphify --update`. Behaves correctly for *this* run.
3. After the run, `manifest.json` now contains **24 entries** (only the files touched in step 2's input, since `--update`'s `.graphify_detect.json` is populated with `new_files` only).
4. Touch nothing; run `/graphify --update` again. `detect_incremental` reports **1008 new files** — every file in the corpus except the 24 from step 2.

The content-hash semantic cache rescues this from being catastrophic — it hits on ~890 of the 1008 files and only ~118 are re-extracted. But:

- The "new files" report is misleading and triggers re-extraction of any uncached doc.
- The deletion-detection logic in the merge step has the wrong "manifest universe" to compare against, so it would miss true deletions.
- `code_only` fast-path doesn't fire when 99 stale code files appear in `new_files`.
- It's the [#538](https://github.com/safishamsi/graphify/issues/538) bug, but symmetric: that one fixed manifest persistence on the full-rebuild path; the `--update` path now persists too little.

## Root cause

Two related defects in `skill.md` (installed/shipped Claude Code skill):

### 1. `--update` flow never populates `.graphify_detect.json`

The `## For --update` section only writes `.graphify_incremental.json`, but Steps 3A–9 (which the section instructs the agent to "run as normal") all read from `.graphify_detect.json`. The skill works at all only because each agent ends up writing this file ad-hoc — and the natural ad-hoc shape is `{'files': new_files}`, which leads to defect 2.

### 2. Step 9 `save_manifest(detect['files'])` shrinks the manifest

```python
# Step 9, current
detect = json.loads(Path('graphify-out/.graphify_detect.json').read_text())
save_manifest(detect['files'])
```

`save_manifest` (graphify/detect.py) iterates the provided file list and writes `{path: mtime}` to `manifest.json`, **overwriting** the previous manifest. If `detect['files']` is the `--update` subset (changed files only), every unchanged file's mtime is dropped from the manifest. Next `detect_incremental` sees `stored_mtime is None` for all of them → flags as new.

This is the same family as [#538](https://github.com/safishamsi/graphify/issues/538) (manifest not saved on full rebuild), but on the opposite path: here the manifest *is* saved on `--update`, but with the wrong (subset) data.

## Suggested fix

Minimal, two-touch patch to `graphify/skill.md`:

**(a)** In the `## For --update` section, after the `detect_incremental` block, add an explicit `.graphify_detect.json` population step that carries both keys:

```python
$(cat graphify-out/.graphify_python) -c "
import json
from pathlib import Path
r = json.loads(Path('graphify-out/.graphify_incremental.json').read_text())
Path('graphify-out/.graphify_detect.json').write_text(json.dumps({
    'files': r.get('new_files', {}),           # drives processing in Steps 3A-3C
    'all_files': r.get('files', {}),           # full corpus, used by Step 9 save_manifest
    'total_files': r.get('new_total', 0),
    'total_words': r.get('total_words', 0),
    'skipped_sensitive': r.get('skipped_sensitive', []),
    'needs_graph': True,
}))
"
```

**(b)** Change Step 9's `save_manifest(detect['files'])` to prefer `all_files` when present:

```python
# In --update mode, 'all_files' carries the full corpus; 'files' is the changed
# subset. Full-rebuild mode populates only 'files', so the fallback handles that.
save_manifest(detect.get('all_files') or detect['files'])
```

These changes are additive — full-rebuild flow is unaffected because it never writes `all_files`, and the fallback `or detect['files']` preserves its existing semantics.

A PR is being prepared.

## Observations / related

- Issue [#826](https://github.com/safishamsi/graphify/issues/826) (cache key path mismatch) is in the same operational area but distinct — the content-hash cache *was* masking some of this for me on subsequent runs.
- The defect mostly hides on code-only corpora because AST extraction is cheap and `code_only` would normally trigger the no-LLM fast-path. On mixed wiki/doc corpora it's much more visible: hundreds of doc files get flagged as "new" and the semantic cache becomes the only thing preventing a full re-extraction bill.
- I've already applied this patch locally and verified: `manifest.json` is restored to 1028 entries, and a subsequent `detect_incremental` reports `new_total: 0` against an unchanged corpus.

Happy to revise the PR shape (e.g., move the file-population logic into `detect.py` instead of the skill markdown if that's preferred upstream).

Adriftnote:
Fixes #837.

## What changes

Adds one bash block to `## For --update (incremental re-extraction)` in `graphify/skill.md` that derives `.graphify_detect.json` from `.graphify_incremental.json` right after `detect_incremental` runs.

Both keys carried:
- `files = new_files` — drives the *incremental* path: Step 3A AST extracts only changed code, Step 3B0 cache check is bounded to the changed set, etc.
- `all_files = files` (from `detect()`) — kept for any stage that wants the full corpus (Step 5 timing estimate header, future audits, etc.)

## Why

In v7 / 0.7.14, every step from 3A onward reads `.graphify_detect.json` unconditionally:

```bash
detect = json.loads(Path('graphify-out/.graphify_detect.json').read_text(...))
for f in detect.get('files', {}).get('code', []):
    ...
```

The `--update` flow currently only writes `.graphify_incremental.json`, so those reads either fail (no prior full-rebuild) or operate on stale corpus data from a prior run.

The downstream merge block (`save_manifest(incremental['files'])` at line 839) already handles manifest persistence correctly, so this PR doesn't need to touch Step 9 or the manifest path.

## Repro that prompted the issue

Mixed wiki+code+raw archive (~1028 files). After a clean full rebuild + a small `--update`, the next `--update` was reporting 1008 "new" files instead of 0. Tracing showed:
1. `.graphify_detect.json` was either absent or stale, and the agent ended up populating it ad-hoc with `{'files': new_files}` to make Step 3A run.
2. Then Step 9's `save_manifest(detect['files'])` shrunk the manifest to the changed subset — root cause of the false-positive cascade.

On v0.7.14 the new merge block sidesteps (2). This PR addresses (1) directly so the flow stops depending on ad-hoc rescue.

## Verification

After this patch, on my corpus:
- `manifest.json` retains all 1028 corpus entries across consecutive `--update` runs
- `detect_incremental` reports `new_total: 0` against an unchanged corpus (was reporting 1008 before)
- A subsequent `--update` after touching ~20 wiki files extracts only those 20 (verified via cache-hit count)

## Notes / alternatives

- Considered moving the `.graphify_detect.json` synthesis into `detect.py` (e.g. an `as_detect_dict()` helper). Kept it in `skill.md` to match the existing per-stage scratchpad convention, but happy to refactor if you'd prefer.
- Did not add a `save_manifest` defensive fallback at Step 9 — current `--update` flow never reaches Step 9 (Steps 4–8 only), and adding one would muddle the diff. Worth a follow-up if you want belt-and-suspenders for future flow changes.

### DIFF
--- a/graphify/skill.md
+++ b/graphify/skill.md
@@ -772,6 +772,24 @@ print(f'{new_total} new/changed file(s) to re-extract.')
 "
 ```
 
+Then populate `.graphify_detect.json` so Steps 3A–6 (which read it unconditionally) see the right state for an incremental run. `files` carries the changed subset (drives Step 3A AST + Step 3B0 cache check on only what changed); `all_files` carries the full corpus for any step that needs corpus-wide context (e.g. Step 5 timing estimate, report header):
+
+```bash
+$(cat graphify-out/.graphify_python) -c "
+import json
+from pathlib import Path
+r = json.loads(Path('graphify-out/.graphify_incremental.json').read_text(encoding=\"utf-8\"))
+Path('graphify-out/.graphify_detect.json').write_text(json.dumps({
+    'files': r.get('new_files', {}),
+    'all_files': r.get('files', {}),
+    'total_files': r.get('new_total', 0),
+    'total_words': r.get('total_words', 0),
+    'skipped_sensitive': r.get('skipped_sensitive', []),
+    'needs_graph': True,
+}, ensure_ascii=False), encoding=\"utf-8\")
+"
+```
+
 If new files exist, first check whether all changed files are code files:
 
 ```bash


safishamsi:
Hey @Adriftnote — thanks for the correct diagnosis and the fix! Your approach was right. We merged a slightly extended version of this directly in commit `822abd6`:

- Your populate step (writing both `files` and `all_files` to `.graphify_detect.json` after `detect_incremental`) — included as-is.
- One additional line we added on top: Step 9's `save_manifest(detect['files'])` was changed to `save_manifest(detect.get('all_files') or detect['files'])` as a defense-in-depth guard, since LLM agents can sometimes run Step 9 during `--update` even though the spec says not to. Without this, an agent that does execute Step 9 would still overwrite the manifest with just the delta.

Your PR was the right call — we just bundled the extra line in directly rather than merging separately. Really appreciate the detailed root-cause write-up and the local verification. 🙏

PR #416, opened 2026-04-17
feat: add HCL/Terraform structural extractor

mauricewittek:
Adds tree-sitter-based extraction for Terraform `.tf` and `.tfvars` files, producing block-level graph nodes and cross-file module dependency edges.

## What it does

- Extracts all 7 HCL block types (resource, data, module, variable, output, locals, provider) as graph nodes with stable namespaced IDs
- Creates containment edges (file → block) and module_source edges (module → referenced source)
- Resolves cross-file module-input edges (module call argument → child variable definition)
- Classifies .tf/.tfvars as code in detect.py and routes them through the HCL extractor in the dispatch loop
- Handles parse errors gracefully with structured diagnostics
- Enforces resource limits (5MB file size, 200k AST nodes)

I tried to implement this in small commits to make it easier to review commit-by-commit.

PR #841, opened 2026-05-12
feat: Terraform/HCL support

wootencl:
Add a Terraform extractor so .tf files participate in graphify's knowledge graph alongside the existing 50+ languages.

  What's new

  - Detection: .tf registered as a code extension in graphify/detect.py. .tfvars is intentionally excluded.
  - Extractor (extract_terraform in graphify/extract.py):
    - Emits nodes for resource, data, module, variable, output, and individual local.* entries inside locals { ... } blocks.
    - Records intra-file references (var.x, local.x, module.x, data.TYPE.NAME, aws_s3_bucket.foo, etc.) for resolution.
    - For modules/<name>/main.tf it emits a synthetic Terraform Module node keyed by directory, so callers using source = "../modules/<name>" get wired up to it.
  - Cross-file pass (_resolve_cross_file_terraform_refs): Terraform treats every .tf in a directory as one module, so references resolve against (parent_dir, ref) tuples — letting main.tf, variables.tf, and outputs.tf reference each other
   naturally. Remote module sources (anything containing ://) are skipped.
  - Tests: a sample.tf fixture plus three new tests in tests/test_extract.py covering single-file extraction, cross-file resolution across main.tf / variables.tf / outputs.tf, and local-module source = "../modules/x" linkage.

  Design notes

  - Intentionally regex-based and structural rather than using a full HCL parser — keeps the dependency footprint flat and matches the conservative pattern used by other extractors in this file.
  - The resource-reference regex requires an underscore in the type (e.g. aws_s3_bucket.foo) to keep false positives off ordinary foo.bar attribute access.
  - Block bodies are sliced by walking match offsets instead of brace counting — sufficient for reference extraction, since nested blocks are scanned with their parent.

---
**Disclaimer**: This was implemented by codex and cross reviewed by claude. Intend on reviewing more thoroughly soon but wanted to get the PR up for visibility to verify this is something acceptable upstream.


kedar-1:
Did you check for PR #416? Is it your PR simpler/better?

wootencl:
Ah whoops... seems I entirely missed that PR/feature. Suspect it's much more robust given this was a one shot from an LLM. Will close this PR in preference to that PR.

PR #843, opened 2026-05-12
fix: add missing file_type values to VALID_FILE_TYPES (closes #840)

### LINKED ISSUE #840: extract emits nodes with file_type values outside the validator's allowed enum, producing warnings on every subsequent update
## Symptom

After `graphify extract`, every subsequent `graphify update <path>` prints one warning line summarising N invalid-file_type nodes in the graph, with one node shown as the example:

```
[graphify] Extraction warning (N issues): Node <id> (id='...') has invalid file_type '<value>' - must be one of ['code', 'concept', 'document', 'image', 'paper', 'rationale']
```

## Repro

`graphifyy 0.7.15`, repo with mixed content (markdown ADRs, YAML pattern files, code).

After `graphify extract .`, inspect the output:

```python
python -c "import json, collections; d=json.load(open('graphify-out/graph.json')); print(collections.Counter(n.get('file_type') for n in d['nodes']).most_common())"
```

Yields invalid `file_type` values: `pattern`, `tech`, `data-source`, `constraint`, `markdown`, `tool`, `gotcha`.

## Impact

Nodes are still queryable via `graphify path` / `get_node` MCP tool and contribute to communities and PageRank — so the warning is cosmetic, not a functional break. But every CLI run emits the warning line.

Concrete count on a real repo (mixed code + ~750 YAML knowledge entries + ~80 ADRs):

- **80 of 71,569 nodes (~0.1%)** carry invalid file_type values
- Distribution: `pattern` (33), `tech` (23), `data-source` (13), `constraint` (4), `markdown` (4), `tool` (2), `gotcha` (1)

Example node returned by `get_node`:

```
Node: Layered Integration Architecture (Base + Adapters + Normalization)
  ID: pattern_layered_integration
  Source: docs/architecture/ADR/archive/ADR-014-integration-patterns.md
  Type: pattern   <- not in the allowed enum
  Community: 22
  Degree: 2
```

## Likely cause

Graphify's LLM-backed `extract` step allows the model to emit broader `file_type` values when extracting concept nodes from documents. The schema validator then rejects values outside the 6-value enum (`code`, `concept`, `document`, `image`, `paper`, `rationale`).

## Possible fixes

1. **Constrain the LLM extraction prompt** to the allowed enum (tightest fix, but may lose semantic nuance — `tech` and `constraint` are useful distinctions)
2. **Widen the validator's accepted enum** to include the LLM's broader vocabulary (most permissive)
3. **Post-process** invalid values to nearest allowed (`pattern` → `concept`, `constraint` → `concept`, `tech` → `concept`, `data-source` → `concept`, `markdown` → `document`, `tool` → `concept`, `gotcha` → `concept`)

Any of the three would silence the warning. (3) preserves backward compatibility for graphs that already have the broader vocabulary; (2) requires a schema-version bump.

## Environment

- `graphifyy` 0.7.15
- Python 3.13
- Windows 11

Happy to provide more sample output or test fixes if helpful.

Pratham-r05:
## Problem
`graphify extract` emits nodes with `file_type` values like `pattern`, 
`tech`, `constraint`, `markdown`, and `tool` that are not in 
`VALID_FILE_TYPES`. This causes a warning line on every subsequent 
`graphify update` run.

## Fix
Added the 5 missing values to `VALID_FILE_TYPES` in `graphify/validate.py`.
Same approach used previously to add `rationale` (see CHANGELOG).

## Tests
567 passed, 0 failures introduced. The 2 pre-existing failures 
(`test_fortran_capital_F_parses_preprocessed`, `test_sql_finds_tables`) 
are unrelated to this change.

Closes #840

### DIFF
--- a/graphify/validate.py
+++ b/graphify/validate.py
@@ -1,7 +1,7 @@
 # validate extraction JSON against the graphify schema before graph assembly
 from __future__ import annotations
 
-VALID_FILE_TYPES = {"code", "document", "paper", "image", "rationale", "concept"}
+VALID_FILE_TYPES = {"code", "document", "paper", "image", "rationale", "concept", "pattern", "tech", "constraint", "markdown", "tool"}
 VALID_CONFIDENCES = {"EXTRACTED", "INFERRED", "AMBIGUOUS"}
 REQUIRED_NODE_FIELDS = {"id", "label", "file_type", "source_file"}
 REQUIRED_EDGE_FIELDS = {"source", "target", "relation", "confidence", "source_file"}


safishamsi:
Hey @Pratham-r05 — thanks for catching this and sending a PR! We went a different direction though, so closing this one.

**Why we didn't merge:** Widening `VALID_FILE_TYPES` in `validate.py` to include `pattern`, `tech`, `constraint` etc. treats the symptom rather than the cause. Once those values are "valid", LLMs have permission to keep inventing arbitrary type names indefinitely — and the validator can never close the set.

**What we did instead (commit `822abd6`):**
1. **Fixed the root cause in the prompts** — `skill.md` and `llm.py` disagreed on the allowed enum (one was missing `concept`, the other was missing `rationale`). Both now explicitly state: *"file_type MUST be one of exactly these six values: code, document, paper, image, rationale, concept — any other value is invalid."*
2. **Added a synonym mapper in `build.py`** — a `_FILE_TYPE_SYNONYMS` dict that coerces known LLM synonyms (`markdown→document`, `pattern→concept`, `tool→code`, etc.) silently before validation, so existing graphs with these values don't warn on every update.

This keeps the validator's enum closed at 6 values while still handling legacy/external graphs gracefully. Thanks again for the report — the symptom description was spot-on.

Issue #760, opened 2026-05-06
AST extractor: 25 reversed calls edges + 3 under-classified cross-file calls (TS extractor)


jjdemon666:
# `graphifyy` v0.7.0 ΓÇö AST extractor emits wrong-direction `calls` edges + under-classifies cross-file calls

**Package:** [`graphifyy`](https://pypi.org/project/graphifyy/) v0.7.0
**Repo:** https://github.com/safishamsi/graphify
**Environment:** Windows 11, Python 3.14, graphifyy 0.7.0 site-packages install
**Reproducer project:** A React + TypeScript + Vite project (`forex-backtester`, ~144 nodes / 178 edges after `/graphify update`)

Two distinct bugs in the TypeScript AST extractor. Together they make ~15% of `calls` edges in a single small project either reversed in direction or under-classified in confidence.

---

## Bug 1 ΓÇö Intra-file `calls` edges have caller and callee swapped

When a function `A()` defined in file X calls another function `B()` also defined in file X, the extractor emits an edge `B --calls--> A` instead of `A --calls--> B`.

This directly violates the rule documented in graphify's own `SKILL.md`:

> Code files: when adding `calls` edges, source MUST be the caller (the function/class doing the calling), target MUST be the callee. Never reverse this direction.

#### `src/engine/MQL4Preprocessor.ts` (6 swaps)

| # | Reported (wrong) | Actual direction in source |
|---|---|---|
| 1 | `parseInputParam` -> `preprocessMQL4` | `preprocessMQL4` -> `parseInputParam` (called from inside `preprocessMQL4` body, ~L430) |
| 2 | `expandMultiDecl` -> `preprocessMQL4` | `preprocessMQL4` -> `expandMultiDecl` |
| 3 | `convertTypedDeclaration` -> `preprocessMQL4` | `preprocessMQL4` -> `convertTypedDeclaration` |
| 4 | `applyConstants` -> `preprocessMQL4` | `preprocessMQL4` -> `applyConstants` |
| 5 | `joinMultilineParens` -> `preprocessMQL4` | `preprocessMQL4` -> `joinMultilineParens` |
| 6 | `fixClassBodies` -> `preprocessMQL4` | `preprocessMQL4` -> `fixClassBodies` |

All six callees are invoked from inside the `preprocessMQL4` body.

#### `src/engine/BacktestEngine.ts` (1 swap)

| # | Reported (wrong) | Actual direction in source |
|---|---|---|
[...]

PR #851, opened 2026-05-13
fix(path): render arrows in original edge direction (#849)

### LINKED ISSUE #849: `graphify path` renders reversed arrows because graph is undirected (renderer bug, post-#760)
# `graphifyy` v0.7.16 — `graphify path` reverses edge arrows; symptom looks like #760 but root cause is the build/render layer

**Package:** `graphifyy` v0.7.16 (`pip install graphifyy`)
**Environment:** macOS, Python 3.14.4
**Reproducer:** A Sanity Studio plugin written in TypeScript (~33 files, 192 nodes, 416 edges after `graphify update .`)

This is **distinct from #760** (which fixed the extractor). The extractor now emits correct caller→callee directions — I verified this on my graph. The bug is at the **build/render** layer: the graph is built as **undirected**, then `graphify path` walks it in traversal order and renders every hop as `u --rel--> v` regardless of which way the underlying edge actually points.

---

## Reproducer

Source (real, two-file cross-file call):

```ts
// server/create-patch-handler.ts:28
const user = await validateSanitySession({ projectId: opts.projectId, credential });
```

So the true call direction is **`createPatchHandler --calls--> validateSanitySession`**.

Build & query:

```
$ graphify update . --force
[graphify watch] Rebuilt: 192 nodes, 416 edges, 10 communities

$ graphify path "validateSanitySession()" "createPatchHandler()"
Shortest path (1 hops):
  validateSanitySession() --calls [EXTRACTED]--> createPatchHandler()
```

That's backwards.

Sanity check that the extractor is fine — `graphify explain` renders the same edge correctly when the source node is the caller:

```
$ graphify explain "createEditHandler()"
Node: createEditHandler()
...
Connections (7):
  --> validateSanitySession() [calls] [EXTRACTED]    ← correct
```

And the edge in `graph.json` is stored correctly (caller in `source`, callee in `target`):

```json
{
  "relation": "calls",
  "source": "server_create_patch_handler_createpatchhandler",
  "target": "server_sanity_validate_session_validatesanitysession",
  "confidence": "EXTRACTED"
}
```

## Root cause

`graph.json` top-level: `"directed": false`. `build.py:74` defaults `build_from_json(... directed: bool = False)`, so the in-memory graph is `nx.Graph`, not `nx.DiGraph`.

In `__main__.py` around line 1549:

```python
path_nodes = _nx.shortest_path(G, src_nid, tgt_nid)   # walks undirected graph
...
for i in range(len(path_nodes) - 1):
    u, v = path_nodes[i], path_nodes[i + 1]
    edata = edge_data(G, u, v)                         # dict still has source/target
    rel = edata.get("relation", "")
    ...
    segments.append(f"--{rel}{conf_str}--> {G.nodes[v].get('label', v)}")
```

`edata["source"]` and `edata["target"]` are preserved on the edge dict but never consulted. The printer assumes traversal direction = edge direction.

## Suggested fix (renderer-side, minimal)

Consult the preserved `source`/`target` on the edge dict and pick the arrow accordingly:

```python
edata = edge_data(G, u, v)
rel = edata.get("relation", "")
if edata.get("source") == u:
    segments.append(f"--{rel}{conf_str}--> {G.nodes[v].get('label', v)}")
else:
    segments.append(f"<--{rel}{conf_str}-- {G.nodes[v].get('label', v)}")
```

Or — cleaner — load the graph as directed (`json_graph.node_link_graph(_raw, directed=True)`) and run `nx.shortest_path` on `G.to_undirected(as_view=True)` for the traversal, while keeping the directed `G` for `edge_data` lookups. That also fixes any other path-style commands that share the blind spot.

## Impact

The `path` command is exactly what an assistant relays verbatim when a user asks "how does X reach Y?" — wrong arrows assert a backwards call graph. Single most user-facing direction bug because of how `path` output is consumed.

spindle79:
Fixes #849.

## Summary

`graphify path` previously printed every hop as `u --rel--> v` based on traversal order from the undirected `shortest_path` walk, ignoring the actual stored direction of the edge between u and v. When a user queried `path B A` and the stored edge was A→B, the output asserted a backwards call graph — the single most user-facing direction bug because of how `path` output is relayed by assistants.

This is **distinct from #760** (extractor). Verified on the repro that the extractor emits correct caller→callee directions; the bug is purely at the build/render layer.

## Root cause

`graph.json` is serialized with `"directed": false` (build.py's default), so `nx.json_graph.node_link_graph` returns an undirected `nx.Graph`. The original direction is consumed into edge endpoint positions but the `--rel-->` arrow at `__main__.py:1565` was printed in traversal order regardless.

```json
// graph.json — link encodes caller→callee correctly
{ "relation": "calls",
  "source": "server_create_patch_handler_createpatchhandler",
  "target": "server_sanity_validate_session_validatesanitysession" }
```

## Fix

In the `cmd == "path"` branch of `__main__.py`:

- Force directed interpretation on read (`_raw = {**_raw, "directed": True}`) so the loaded `G` is a `DiGraph` that preserves edge direction.
- Run `shortest_path` on `G.to_undirected(as_view=True)` so a path is still found regardless of which direction the user supplied src/tgt in.
- For each hop, check `G.has_edge(u, v)` against the directed graph. Render `--rel-->` if the stored edge matches traversal direction, `<--rel--` if it points the other way.

The renderer for `explain` is structurally similar but, by always rendering edges as outbound from a *fixed* queried node, happens to get the right answer when the queried node is the caller and the wrong answer otherwise. Out of scope for this PR — happy to follow up if you'd like, but it warrants its own issue.

## Tests

`tests/test_path_cli.py` — two tests, both on a graph.json fixture whose on-disk shape matches what `graphify update` produces today (`directed: false`, `links[i].source` = caller):

1. `test_path_renders_forward_arrow_when_traversal_matches_edge_direction` — `path createPatchHandler validateSanitySession` → `createPatchHandler() --calls [EXTRACTED]--> validateSanitySession()`
2. `test_path_renders_reverse_arrow_when_traversal_opposes_edge_direction` — `path validateSanitySession createPatchHandler` → `validateSanitySession() <--calls [EXTRACTED]-- createPatchHandler()` (and asserts the old buggy forward output is NOT present)

## Risk

- Loading the JSON as directed when it was serialized undirected is safe: `links[].source`/`.target` are preserved verbatim by `node_link_graph`, and they always represent caller→callee per the existing build conventions (verified in `build.py` and corroborated by `explain` already relying on this for outbound edge iteration).
- `shortest_path` runs on the undirected view, so paths that previously resolved still resolve.
- Multigraph case is preserved — `node_link_graph` honors the `multigraph` flag independently, and `edge_data(G, u, v)` already handles `MultiDiGraph`.

## Test plan

- [x] Two new tests pass locally.
- [x] Full suite: 741 passed, same 11 baseline failures (SQL needs optional `tree-sitter-sql`; one Fortran preprocessed test; two Ollama env-var tests) — zero new failures.
- [x] Manual repro from #849 now renders correctly with both src/tgt orderings.

### DIFF
--- a/graphify/__main__.py
+++ b/graphify/__main__.py
@@ -1514,6 +1514,11 @@ def main() -> None:
         _raw = json.loads(gp.read_text(encoding="utf-8"))
         if "links" not in _raw and "edges" in _raw:
             _raw = dict(_raw, links=_raw["edges"])
+        # #849: the serialized link.source/link.target fields encode the original
+        # caller→callee (or import→importee) direction regardless of the top-level
+        # "directed" flag the build wrote. Force directed interpretation on read so
+        # the renderer can recover and display the correct arrow direction below.
+        _raw = {**_raw, "directed": True}
         try:
             G = json_graph.node_link_graph(_raw, edges="links")
         except TypeError:
@@ -1547,7 +1552,9 @@ def main() -> None:
                         file=sys.stderr,
                     )
         try:
-            path_nodes = _nx.shortest_path(G, src_nid, tgt_nid)
+            # Traverse on the undirected view so a path can be found regardless of
+            # which direction the user supplied src/tgt in.
+            path_nodes = _nx.shortest_path(G.to_undirected(as_view=True), src_nid, tgt_nid)
         except (_nx.NetworkXNoPath, _nx.NodeNotFound):
             print(f"No path found between '{source_label}' and '{target_label}'.")
             sys.exit(0)
@@ -1556,13 +1563,22 @@ def main() -> None:
         from graphify.build import edge_data
         for i in range(len(path_nodes) - 1):
             u, v = path_nodes[i], path_nodes[i + 1]
-            edata = edge_data(G, u, v)
+            # #849: render the arrow in the original stored edge direction.
+            if G.has_edge(u, v):
+                edata = edge_data(G, u, v)
+                forward = True
+            else:
+                edata = edge_data(G, v, u)
+                forward = False
             rel = edata.get("relation", "")
             conf = edata.get("confidence", "")
             conf_str = f" [{conf}]" if conf else ""
             if i == 0:
                 segments.append(G.nodes[u].get("label", u))
-            segments.append(f"--{rel}{conf_str}--> {G.nodes[v].get('label', v)}")
+            if forward:
+                segments.append(f"--{rel}{conf_str}--> {G.nodes[v].get('label', v)}")
+            else:
+                segments.append(f"<--{rel}{conf_str}-- {G.nodes[v].get('label', v)}")
         print(f"Shortest path ({hops} hops):\n  " + " ".join(segments))
 
     elif cmd == "explain":
--- a/tests/test_path_cli.py
+++ b/tests/test_path_cli.py
@@ -1,0 +1,92 @@
+"""Tests for `graphify path` arrow-direction rendering (regression for #849)."""
+from __future__ import annotations
+
+import json
+
+import networkx as nx
+from networkx.readwrite import json_graph
+
+import graphify.__main__ as mainmod
+
+
+def _write_undirected_graph(tmp_path):
+    """Write a graph.json with `directed: false` (the build's default) but link
+    source/target encoding caller→callee. This mirrors the on-disk shape that
+    ships from `graphify update`."""
+    G = nx.Graph()
+    G.add_node(
+        "create_patch_handler",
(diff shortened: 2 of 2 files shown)


safishamsi:
Thanks for the report! We shipped this fix ourselves in commit 7bb0919 (v0.7.17) — path arrow direction now correctly renders `-->` and `<--` based on actual edge direction in directed graphs. Closing since the fix is already in main.

PR #854, opened 2026-05-13
fix(explain): render arrows in original edge direction (#853)

### LINKED ISSUE #853: `graphify explain` renders inbound edges as outbound (twin of #849)
# `graphifyy` v0.7.16 — `graphify explain` renders inbound edges as outbound (structural twin of #849)

**Package:** `graphifyy` v0.7.16 (`pip install graphifyy`)
**Environment:** macOS, Python 3.14.4
**Related:** [#849](https://github.com/safishamsi/graphify/issues/849) — same root cause as `graphify path`, fixed there in PR [#851](https://github.com/safishamsi/graphify/pull/851).

`graphify explain` prints every neighbor as `--> X [relation]`, regardless of whether the underlying edge actually points out of the queried node or into it. For a queried *callee*, all its callers are rendered as outbound calls — asserting a backwards call graph.

I noted this as out-of-scope in [#851](https://github.com/safishamsi/graphify/pull/851); now confirmed in the wild on a real codebase.

---

## Reproducer

Same fixture as #849. Source (real, two-file cross-file call):

```ts
// server/create-patch-handler.ts:28
const user = await validateSanitySession({ projectId: opts.projectId, credential });
```

So the true direction is **`createPatchHandler --calls--> validateSanitySession`**. `validateSanitySession()` is a leaf — it has no outbound `calls` edges; everything that touches it is inbound.

Build & query:

```
$ graphify update . --force
[graphify watch] Rebuilt: 192 nodes, 416 edges, 10 communities

$ graphify explain "validateSanitySession()"
Node: validateSanitySession()
  ...
Connections (13):
  --> create-edit-handler.ts [imports] [EXTRACTED]
  --> createPatchHandler() [calls] [EXTRACTED]
  --> createEditHandler() [calls] [EXTRACTED]
  --> createAuthHandler() [calls] [EXTRACTED]
  ...
```

Every arrow should be `<--` — these are all *callers/importers* of `validateSanitySession`, not things it calls.

## Root cause

Same as #849 — the build serializes `directed: false`, `__main__.py`'s `cmd == "explain"` branch calls `json_graph.node_link_graph(_raw, edges="links")` and gets an undirected `nx.Graph`. Then:

```python
neighbors = list(G.neighbors(nid))   # all neighbors, direction lost
for nb in sorted(neighbors, ...)[:20]:
    edata = edge_data(G, nid, nb)
    ...
    print(f"  --> {G.nodes[nb].get('label', nb)} [{rel}] [{conf}]")
```

Hard-coded `-->`. The renderer assumes outbound because `G.neighbors()` on an undirected graph returns everything as if it were outbound from the queried node.

## Suggested fix (mirrors #851)

Force directed interpretation on read, then iterate successors and predecessors separately:

```python
_raw = {**_raw, "directed": True}
...
G = json_graph.node_link_graph(_raw, edges="links")
...
connections = []
for nb in G.successors(nid):
    connections.append(("out", nb, edge_data(G, nid, nb)))
for nb in G.predecessors(nid):
    connections.append(("in", nb, edge_data(G, nb, nid)))
# Sort by neighbor degree, render '-->' / '<--' per direction.
```

A neighbor that has both inbound *and* outbound edges to the queried node (mutual references) gets rendered twice — once per direction — which is the truthful directed representation.

## Impact

Same shape as #849 — `explain` output is what an assistant relays verbatim when a user asks "what is X / what touches X?". Wrong arrows assert a backwards graph.

PR with fix + tests incoming.

spindle79:
Fixes #853.

## Summary

Same structural bug as #849, same fix shape as PR #851 — applied to `graphify explain`. Previously every connection rendered as `--> X [relation]`, asserting a backwards call graph whenever the queried node was a callee.

## Reproducer

Real-world: `explain validateSanitySession()` printed `--> createPatchHandler() [calls]` even though the stored edge is `createPatchHandler --calls--> validateSanitySession` (verified in `graph.json` and by the file source). `validateSanitySession()` is a pure leaf utility — all 13 of its connections are inbound — yet every one rendered as outbound.

## Fix

Same as #851:

- Force directed interpretation on read (`_raw = {**_raw, "directed": True}`) so the loaded `G` is a `DiGraph` preserving edge direction.
- Iterate `G.successors(nid)` and `G.predecessors(nid)` separately, tagging each connection as `"out"` or `"in"`.
- Render `-->` for outbound, `<--` for inbound.

A neighbor that has both an inbound and outbound edge to the queried node (mutual references) is rendered twice — once per edge. That's the truthful directed representation; the previous undirected `G.neighbors()` would have silently collapsed it to one row.

## Tests

`tests/test_explain_cli.py` — two tests, both on a `graph.json` fixture whose on-disk shape matches what `graphify update` produces (`directed: false`, `links[i].source` = caller):

1. `test_explain_renders_callers_as_inbound_and_callees_as_outbound` — queried node is `validateSanitySession`. Its two callers (`createPatchHandler`, `createEditHandler`) must render as `<--`; its own callee (`stableStringify`) must render as `-->`. Asserts the old buggy `--> createPatchHandler() [calls]` is **not** in the output.
2. `test_explain_renders_pure_caller_as_outbound` — converse: querying `createPatchHandler` must show `validateSanitySession` as `-->` and no `<--` rows.

The fixture writes graph.json directly rather than via `nx.Graph().add_edge()` + `node_link_data()` — the latter emits links in *node-insertion order*, not edge-argument order, which can silently encode the wrong direction in small fixtures and mask the bug.

## Test plan

- [x] Two new tests pass locally.
- [x] Full suite: 737 passed, same 11 baseline failures (SQL needs optional `tree-sitter-sql`; one Fortran preprocessed test; two Ollama env-var tests) — zero new failures.
- [x] Manual repro: `explain validateSanitySession()` now renders all 13 connections as `<--`. Verified on the same Sanity Studio plugin reproducer used in #849/#851.

## Notes for the reviewer

- If #851 has already merged, this branch may need a trivial rebase (both touch adjacent regions of `__main__.py` for path/explain), but the two changes are independent.
- The directed-on-read interpretation is safe even for graphs serialized with `directed: false` because `links[].source`/`.target` are preserved verbatim by `node_link_graph` and always represent caller→callee per the existing build conventions — verified empirically on real graphs.

### DIFF
--- a/graphify/__main__.py
+++ b/graphify/__main__.py
@@ -1584,6 +1584,11 @@ def main() -> None:
         _raw = json.loads(gp.read_text(encoding="utf-8"))
         if "links" not in _raw and "edges" in _raw:
             _raw = dict(_raw, links=_raw["edges"])
+        # Same fix as #849 for `graphify path`: the serialized link.source/link.target
+        # fields encode caller→callee direction regardless of the top-level "directed"
+        # flag the build wrote. Force directed interpretation on read so the renderer
+        # can distinguish outbound (-->) from inbound (<--) connections.
+        _raw = {**_raw, "directed": True}
         try:
             G = json_graph.node_link_graph(_raw, edges="links")
         except TypeError:
@@ -1600,17 +1605,26 @@ def main() -> None:
         print(f"  Type:      {d.get('file_type', '')}")
         print(f"  Community: {d.get('community', '')}")
         print(f"  Degree:    {G.degree(nid)}")
-        neighbors = list(G.neighbors(nid))
-        if neighbors:
-            from graphify.build import edge_data
-            print(f"\nConnections ({len(neighbors)}):")
-            for nb in sorted(neighbors, key=lambda n: G.degree(n), reverse=True)[:20]:
-                edata = edge_data(G, nid, nb)
+        # Collect outbound and inbound edges separately so we render each arrow
+        # in the actual stored direction. A neighbor that has both an inbound and
+        # an outbound edge (mutual references) is rendered twice — once per edge —
+        # which is the truthful representation in a directed graph.
+        from graphify.build import edge_data
+        connections: list[tuple[str, str, dict]] = []  # (direction, neighbor_id, edge_data)
+        for nb in G.successors(nid):
+            connections.append(("out", nb, edge_data(G, nid, nb)))
+        for nb in G.predecessors(nid):
+            connections.append(("in", nb, edge_data(G, nb, nid)))
+        if connections:
+            print(f"\nConnections ({len(connections)}):")
+            connections.sort(key=lambda c: G.degree(c[1]), reverse=True)
+            for direction, nb, edata in connections[:20]:
                 rel = edata.get("relation", "")
                 conf = edata.get("confidence", "")
-                print(f"  --> {G.nodes[nb].get('label', nb)} [{rel}] [{conf}]")
-            if len(neighbors) > 20:
-                print(f"  ... and {len(neighbors) - 20} more")
+                arrow = "-->" if direction == "out" else "<--"
+                print(f"  {arrow} {G.nodes[nb].get('label', nb)} [{rel}] [{conf}]")
+            if len(connections) > 20:
+                print(f"  ... and {len(connections) - 20} more")
 
     elif cmd == "add":
         if len(sys.argv) < 3:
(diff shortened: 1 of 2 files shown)


safishamsi:
Thanks for the report! We shipped this fix ourselves in commit 7bb0919 (v0.7.17) — explain arrow direction now correctly shows `-->` for successors and `<--` for predecessors in directed graphs. Closing since the fix is already in main.

PR #856, opened 2026-05-13
feat(llm): add `--backend claude-cli` (routes through Claude Code CLI, no ANTHROPIC_API_KEY needed)

### LINKED ISSUE #855: Feature request: `--backend claude-cli` (route through Claude Code CLI, no `ANTHROPIC_API_KEY` needed)
# Feature request: `--backend claude-cli` — route through the local Claude Code CLI instead of requiring `ANTHROPIC_API_KEY`

**Package:** `graphifyy` v0.7.16
**Use case:** Claude Code Pro/Max subscribers running graphify's semantic pass on their own codebase.

## What

Add a `claude-cli` backend that shells out to the locally-installed `claude` CLI (Claude Code) via `claude -p --output-format json` instead of calling the Anthropic API directly.

## Why

`--backend claude` currently requires `ANTHROPIC_API_KEY` — pay-as-you-go API billing, separate from the Claude Code subscription. For Pro/Max subscribers, this means:

1. Provisioning a second Anthropic account (or a billing key on the same one) just to run graphify's one-off semantic pass.
2. Paying twice — once via the subscription (which they're already using for Claude Code itself) and once via API credit (for graphify).

The local `claude` CLI is already authenticated via the user's subscription OAuth and supports a non-interactive `-p` mode with `--output-format json` that returns a parseable envelope including the model result, usage stats, stop reason, and modelUsage map. It can serve as a drop-in transport for graphify's `_call_claude()` use case.

Verified on a real codebase: ran graphify's full semantic extract (167k input / 8k output tokens, Opus 4.7 1M context) entirely through `claude -p`, billed against my Max plan instead of API credit.

## Approach

Mirror the existing `claude` backend with a new `claude-cli` arm:

- New `BACKENDS["claude-cli"]` entry with `env_key: None` and zero pricing (calls bill against the plan, not the API — graphify's cost estimator treats this as $0).
- New `_call_claude_cli(user_message, max_tokens)` that:
  - Verifies `shutil.which("claude")` finds the binary.
  - Subprocess-calls `claude -p --output-format json --append-system-prompt <_EXTRACTION_SYSTEM>` with the user message piped on stdin.
  - Parses the JSON envelope: extracts `result` (the model output), sums `usage.input_tokens + cache_read_input_tokens + cache_creation_input_tokens` for the input token total, and maps `stop_reason == "max_tokens"` to graphify's `finish_reason == "length"`.
- New dispatch arm in `extract_files_direct` plus the lower `_call_llm` helper.
- New env-key carve-out in `__main__.py` extract validation: when `backend == "claude-cli"`, check `shutil.which("claude")` instead of an env variable. Mirrors ollama's local-loopback carve-out.

The user must have authenticated `claude` interactively at least once so the OAuth tokens are stored locally; after that, `-p` mode runs against the subscription without further prompts.

## Notes for the maintainer

- The current Claude Code CLI's `--bare` mode (which would strip the default system prompt) requires `ANTHROPIC_API_KEY` and would defeat the purpose, so we use `--append-system-prompt` instead. The base Claude Code prompt is appended, but the `_EXTRACTION_SYSTEM` prompt is strict enough that the model produces valid JSON anyway. Verified empirically — see PR for test fixture using a real Claude Code response envelope.
- Tests in the PR mock `subprocess.run` and `shutil.which` so the suite still runs on CI without the `claude` binary or network access.
- Performance is comparable to the API backend: the Claude Code CLI caches its system prompt server-side, so subsequent calls in a session benefit from the same cache discount as direct API calls.

PR with implementation + tests incoming.

spindle79:
Closes #855.

## Summary

Adds a `claude-cli` backend that shells out to the locally-installed `claude` CLI (Claude Code) via `claude -p --output-format json` instead of calling the Anthropic API directly. Lets Pro/Max subscribers run graphify's semantic pass against their plan instead of provisioning a separate `ANTHROPIC_API_KEY`.

Verified end-to-end on a real codebase: full semantic extract (167k in / 8k out tokens, Opus 4.7 1M context), billed against the user's Max plan, $0 marginal cost.

## Changes

- `graphify/llm.py`:
  - New `BACKENDS["claude-cli"]` entry — `env_key: None`, zero pricing (calls bill against the subscription, not API credit, so estimator should treat this as $0).
  - New `_call_claude_cli(user_message, max_tokens)`:
    - Verifies `shutil.which("claude")` finds the CLI.
    - Subprocess-calls `claude -p --output-format json --append-system-prompt <_EXTRACTION_SYSTEM>` with the user message piped on stdin.
    - Parses the JSON envelope (`result` → model output; `usage` → token counts; `stop_reason` → `finish_reason`; `modelUsage` → model id).
    - Reports total input tokens as `usage.input_tokens + cache_read_input_tokens + cache_creation_input_tokens` so accounting reflects what the model actually processed (the CLI's default system prompt is cached aggressively, so subsequent calls within a session see high cache_read counts).
  - New dispatch arm in `extract_files_direct` plus the lower `_call_llm` helper used by the dedup tiebreaker.
  - New env-key skip for `claude-cli` (mirrors the existing `bedrock` skip in the validation).

- `graphify/__main__.py`:
  - New env-key carve-out in the extract command's backend validation: when `backend == "claude-cli"`, check `shutil.which("claude")` instead of an env var. Same shape as the existing ollama local-loopback carve-out, with a helpful error message if the CLI isn't installed.

- `tests/test_claude_cli_backend.py`:
  - 7 tests, all mocking `subprocess.run` + `shutil.which` so the suite runs on CI without the `claude` binary or network.
  - Coverage: success path (parsed result + usage sums + model id), `stop_reason == "max_tokens"` → `finish_reason == "length"`, missing CLI raises, non-zero exit raises, unparseable JSON envelope raises, end-to-end through `extract_files_direct`, and pricing-is-zero contract.

## Why not `--bare` mode?

The CLI's `--bare` flag strips the default Claude Code system prompt but requires `ANTHROPIC_API_KEY` to authenticate — defeating the purpose. So this backend uses `--append-system-prompt` instead. The default Claude Code prompt gets appended to `_EXTRACTION_SYSTEM`, but the extraction prompt is strict enough that Claude produces valid JSON anyway. Empirically verified: ran the full graphify extract pipeline through this backend on a 1,083-node graph and got valid extraction output.

One side effect: `usage.input_tokens` for the first call in a session is large (the CLI's system prompt is part of the model input). Subsequent calls hit `cache_read_input_tokens` for that prefix and the fresh-input count drops to just the user message. This is normal Anthropic prompt caching behavior; the test fixture uses a realistic envelope showing this pattern.

## Test plan

- [x] Seven new tests pass locally.
- [x] Full suite: 742 passed, same 11 baseline failures (SQL needs optional `tree-sitter-sql`; one Fortran preprocessed test; two Ollama env-var tests) — zero new failures.
- [x] Manual end-to-end: `graphify extract . --backend claude-cli` on a real 336-code-file / 32-doc corpus completed in a single chunk, produced 1,083 graph nodes and 2,140 edges, $0 marginal cost on the user's Max plan.

## Notes for the reviewer

- I considered adding a `detect_backend()` arm that auto-selects `claude-cli` when the binary is present and no API keys are set. Skipped for this PR to keep behavior change minimal — `--backend claude-cli` must be explicit. Happy to add auto-detection in a follow-up if you'd prefer.
- The pricing dict is `{"input": 0.0, "output": 0.0}` because cost is plan-based, not per-token. If you'd rather track plan token *usage* (without dollarizing it) the estimator could grow a `plan_tokens` field — also a follow-up.

### DIFF
--- a/graphify/__main__.py
+++ b/graphify/__main__.py
@@ -2422,6 +2422,18 @@ def main() -> None:
                     host in ("localhost", "127.0.0.1", "::1")
                     or host.startswith("127.")
                 )
+            if backend == "claude-cli":
+                # Routes through the locally-installed Claude Code CLI which
+                # authenticates via its own subscription session — no API key.
+                import shutil as _shutil_check
+                allow_no_key = _shutil_check.which("claude") is not None
+                if not allow_no_key:
+                    print(
+                        "error: backend 'claude-cli' requires the `claude` CLI on $PATH "
+                        "(install Claude Code and run `claude` once to authenticate).",
+                        file=sys.stderr,
+                    )
+                    sys.exit(1)
             if not allow_no_key:
                 print(
                     f"error: backend '{backend}' requires {_format_backend_env_keys(backend)} to be set.",
--- a/graphify/llm.py
+++ b/graphify/llm.py
@@ -53,6 +53,20 @@ BACKENDS: dict[str, dict] = {
         "temperature": 0,
         "max_tokens": 16384,
     },
+    "claude-cli": {
+        # Routes through the locally-installed `claude` CLI (Claude Code) using
+        # `-p --output-format json`. Authenticates via the user's existing
+        # Pro/Max subscription instead of a separate ANTHROPIC_API_KEY — costs
+        # are billed to the plan, not pay-as-you-go API credit, so pricing
+        # below is zero for graphify's cost-estimator. Model selection is
+        # whatever the active Claude Code session is using; the `default_model`
+        # value is only for cost-tracking labels.
+        "default_model": "claude-code-plan",
+        "env_key": None,
+        "pricing": {"input": 0.0, "output": 0.0},
+        "temperature": 0,
+        "max_tokens": 16384,
+    },
     "kimi": {
         "base_url": "https://api.moonshot.ai/v1",
         "default_model": "kimi-k2.6",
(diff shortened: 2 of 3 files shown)


safishamsi:
Implemented in commit `258d260` — great PR @spindle7

Two small additions on top of your implementation before landing:
1. **`--no-session-persistence`** added to the subprocess args — prevents session state accumulating across sequential extraction chunks, no auth impact.
2. **`max_concurrency=1` guard** added to `extract_corpus_parallel` (mirrors the ollama guard) — parallel `claude -p` processes conflict over Claude Code session state. Can be opted out with `GRAPHIFY_CLAUDE_CLI_PARALLEL=1` for users who know what they're doing.

9/9 tests pass. End-to-end verified.

PR #875, opened 2026-05-15
feat(sql): extract FK relations, trigger bindings, and procedure/function table dependencies from Firebird SQL

simeonbodurov:
Summary

  This PR fixes three gaps in extract_sql() that made Firebird 5 SQL files (specifically large schemas like Speedy.sql at ~56k lines) produce a nearly empty knowledge graph. After this change, a 56k-line
  Firebird schema goes from ~26 extracted nodes to a full dependency graph with ~150 FK edges, ~440 trigger bindings, and hundreds of reads_from edges for stored procedures and functions.

  ---
  Problem 1 — FK relations were silently dropped

  Speedy.sql uses table-level FOREIGN KEY ... REFERENCES ... constraints declared inside the CREATE TABLE body, not inline per-column REFERENCES. The old code only iterated column_definition children and
  skipped everything else, so all 151 FK relationships were invisible to graphify.

  Root cause: The parser wraps table-level constraints in a constraints node (sibling of column_definition), which the old loop skipped with continue.

  Fix:
  - Added a second branch that handles constraints children containing constraint nodes, using the same keyword_references → object_reference pattern already used in the alter_table branch.
  - Added a regex fallback that scans each CREATE TABLE block's raw text for any REFERENCES still missed — needed because Firebird's COMPUTED BY columns produce ERROR nodes in the parse tree that cause the
  parser to drop the trailing constraints block entirely.
  - Added a global post-walk regex fallback that de-duplicates against already-emitted edges, catching anything the tree walk still misses.

  ---
  Problem 2 — Triggers were not bound to their tables

  444 Firebird procedures and triggers are absorbed as opaque fb_proc_or_trigger regex tokens (a single atomic parse node with no children). Only 2 create_trigger nodes survived with full structure. Neither
  case was handled — no trigger nodes were emitted at all.

  Fix:
  - Added a create_trigger branch: extracts trigger name (first object_reference after keyword_trigger) and table name (first object_reference after keyword_for), emits a triggers edge.
  - Added a fb_proc_or_trigger branch: regex-parses the raw token text to extract the object type (PROCEDURE / TRIGGER / FUNCTION) and name, then for triggers finds the FOR <tablename> clause and emits a
  triggers edge.
  - Added top-level dispatch for fb_proc_or_trigger, set_term, and declare_external_function statement types that were previously skipped by the statement-only dispatch loop.

  ---
  Problem 3 — Procedure/function → table dependencies not extracted

  _walk_from_refs() looks for from/join tree-sitter nodes, but fb_proc_or_trigger bodies have no children — the entire body is opaque text. So no reads_from edges were emitted for any of the 444 stored
  procedures and functions.

  Fix: In the fb_proc_or_trigger branch, regex-scan the body text for:
  - FROM <table>, JOIN <table>, INTO <table> → reads_from edge
  - UPDATE <table> → reads_from edge

  A _NON_TABLES exclusion set filters SQL keywords that appear after these tokens but are not table names (SELECT, WHERE, NULL, FIRST, SKIP, etc.). Deduplication prevents multiple edges to the same table from
  one object.

  ---
  Results on Speedy.sql (56k lines, Firebird 5)

  ┌───────────────────────┬────────┬──────────┐
  │        Metric         │ Before │  After   │
  ├───────────────────────┼────────┼──────────┤
  │ references edges (FK) │ 0      │ ~151     │
  ├───────────────────────┼────────┼──────────┤
  │ Trigger nodes         │ 0      │ ~444     │
  ├───────────────────────┼────────┼──────────┤
  │ triggers edges        │ 0      │ ~444     │
  ├───────────────────────┼────────┼──────────┤
  │ reads_from edges      │ ~0     │ hundreds │
  └───────────────────────┴────────┴──────────┘

  ---
  Files changed

  - graphify/extract.py — all changes inside extract_sql() / its inner walk() function. No API or interface changes.

### DIFF
--- a/graphify/extract.py
+++ b/graphify/extract.py
@@ -2681,20 +2681,53 @@ def extract_sql(path: Path) -> dict:
                 # Foreign key REFERENCES
                 for col in node.children:
                     if col.type == "column_definitions":
+                        has_error = any(cd.type == "ERROR" for cd in col.children)
+                        seen_refs: set[str] = set()
                         for cd in col.children:
-                            if cd.type != "column_definition":
-                                continue
-                            ref_name: str | None = None
-                            found_ref = False
-                            for cc in cd.children:
-                                if cc.type == "keyword_references":
-                                    found_ref = True
-                                elif found_ref and cc.type == "object_reference":
-                                    ref_name = _read(cc)
-                                    break
-                            if ref_name:
-                                ref_nid = _make_id(stem, ref_name)
-                                _add_edge(nid, ref_nid, "references", line)
+                            if cd.type == "column_definition":
+                                # Inline column-level REFERENCES
+                                ref_name: str | None = None
+                                found_ref = False
+                                for cc in cd.children:
+                                    if cc.type == "keyword_references":
+                                        found_ref = True
+                                    elif found_ref and cc.type == "object_reference":
+                                        ref_name = _read(cc)
+                                        break
+                                if ref_name:
+                                    ref_nid = _make_id(stem, ref_name)
+                                    _add_edge(nid, ref_nid, "references", line)
+                                    seen_refs.add(ref_name.lower())
+                            elif cd.type == "constraints":
+                                # Table-level constraints wrapper: FOREIGN KEY ... REFERENCES ...
+                                for constraint in cd.children:
+                                    if constraint.type != "constraint":
+                                        continue
+                                    ref_name = None
+                                    found_ref = False
+                                    for cc in constraint.children:
+                                        if cc.type == "keyword_references":
+                                            found_ref = True
+                                        elif found_ref and cc.type == "object_reference":
+                                            ref_name = _read(cc)
+                                            break
+                                    if ref_name:
+                                        ref_nid = table_nids.get(ref_name.lower()) or _make_id(stem, ref_name)
+                                        _add_edge(nid, ref_nid, "references", line)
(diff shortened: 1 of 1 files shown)


safishamsi:
Hey @simeonbodurov — really solid work here, thank you for the detailed write-up and the before/after numbers on Speedy.sql. The three fixes (table-level FK constraints, trigger bindings, procedure/function table deps) are exactly right and address a genuine gap in the extractor.

We did a thorough review before merging and caught two small issues we wanted to fix before shipping:

1. **`latin-1` decode in the global regex fallback** — the comment says "byte-stable" but that property isn't actually used anywhere. The side effect is that non-ASCII SQL identifiers get hashed differently than the tree walk (which uses UTF-8), producing phantom disconnected reference targets. We switched to `utf-8` with `errors="replace"`.

2. **Block-end regex missing `re.IGNORECASE`** — `re.search(r"\n(?:CREATE|SET\s+TERM|ALTER)\s", ...)` wouldn't match lowercase keywords, and a `CREATE TABLE` block could extend past its boundary and pick up stray `REFERENCES` from the next statement.

We also added `"lateral"` to `_NON_TABLES` and renamed the inner `seen_refs` to `seen_tbls` in the `fb_proc_or_trigger` branch for clarity.

All of this has been shipped in v0.8.4 (`299b6ba`). Closing this PR — your contribution is in the release, thank you!

Issue #872, opened 2026-05-15
deepseek api env support


heartacker:
thanks

safishamsi:
Implemented in v0.8.8 — DeepSeek is now a supported backend. Set `DEEPSEEK_API_KEY` and use `--backend deepseek`. Default model is `deepseek-v4-flash`.

PR #876, opened 2026-05-15
Two fixes for graph integrity after file deletions — the --update flow and the pruning pipeline.

dg-ye:
Changes

  - --update flow (#872): The early-exit guard (new_total == 0) previously skipped the deletion check entirely. When a user removed a file,
  ghost nodes and edges stayed in the graph until a full rebuild. Now deleted_files is also checked before the early exit, with an
  empty-extraction fallback so build_merge(prune_sources=...) can run even without new files to extract.
  - Edge pruning (#871): build_merge(prune_sources=...) now removes edges whose source_file matches a deleted source (previously only nodes were
   pruned, leaving dangling edges). Also normalizes relative/absolute path comparison between detect_incremental output and graph storage.
  - Community label key type: report.generate() and analyze.suggest_questions() now normalize community_labels keys to int at entry — JSON
  deserialization produces string keys ("0") but the rest of the pipeline uses int community IDs, causing .get() to fall through to the
  "Community N" default.

  Test plan

  - Verify graphify --update . after deleting a tracked file removes its nodes and edges
  - Verify graphify report and graphify analyze produce correct community labels with JSON-loaded community_labels

### DIFF
--- a/graphify/analyze.py
+++ b/graphify/analyze.py
@@ -383,6 +383,14 @@ def suggest_questions(
     questions = []
     node_community = _node_community_map(communities)
 
+    # Normalize community_labels keys to int — JSON deserialization produces
+    # string keys but the rest of the pipeline uses int community IDs.
+    if community_labels:
+        community_labels = {
+            int(k) if isinstance(k, str) else k: v
+            for k, v in community_labels.items()
+        }
+
     # 1. AMBIGUOUS edges → unresolved relationship questions
     for u, v, data in G.edges(data=True):
         if data.get("confidence") == "AMBIGUOUS":
--- a/graphify/build.py
+++ b/graphify/build.py
@@ -297,9 +297,14 @@ def build_merge(
 
     # Prune nodes from deleted source files
     if prune_sources:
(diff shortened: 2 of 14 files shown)


safishamsi:
Hey @iamgpa — great catch on all three issues, and the community label key normalization + skill file deletion-only fix were spot on.

We caught one correctness bug in the `build.py` change before merging: the `_prune_set` was built as `set(prune_sources) | {Path(s).name for s in prune_sources}`, which adds bare filenames (basenames) to the set. This creates false positive matches — if a project has `src/utils.py` and `lib/utils.py` and only one is deleted, edges from the surviving file also get pruned.

The fix is simpler than it looks: both `deleted_files` (from the manifest) and `source_file` in the graph are produced by the same `detect()` call on the same absolute root, so they're always in the same format and match directly. Just `set(prune_sources)` is correct — no basename expansion needed.

We also gave the Windows skill a proper PowerShell fallback block (`if (-not (Test-Path ...))`) instead of bash syntax for the empty-extraction step.

Everything has been shipped in v0.8.4 (`299b6ba`). Closing this PR — your contributions are in the release, thank you!

PR #884, opened 2026-05-15
feat: BYOND DreamMaker support (.dm/.dme/.dmi/.dmm/.dmf)

kanya-approve:
## Summary

Adds first-class graphify support for the full BYOND project surface, not just the DM language.

**Commit 1 — `.dm` / `.dme` (DreamMaker source)**

- `tree-sitter-dm` lives in `[project.optional-dependencies].dreammaker` so the base PyPI install of `graphifyy` stays clean (no git deps in `[project]`, no broken-sdist resolution failures). Mirrors the pattern used for `[sql]`.
- `tree-sitter-dm` 0.25.0's PyPI artifact is currently broken (sdist missing `scanner.c`, no wheels). `[tool.uv.sources]` overrides the resolution to git for uv users, so `uv sync --extra dreammaker` works today. Pip users with the extra are documented as needing a manual git install until 0.25.1 lands.
- DM source-language tests are guarded by `@pytest.mark.skipif(not _HAS_TREE_SITTER_DM, ...)` so the suite stays green when the grammar isn't installed.
- DM identity is path-based (`/datum/foo/proc/bar`), so the generic class-body walker doesn't fit. `extract_dm` is a custom walk handling `type_definition`, nested `type_proc_definition` / `type_proc_override`, top-level `proc_definition` / `proc_override` with optional `type_path` prefix, `preproc_include`, `call_expression` / `field_proc_expression`, and `new_expression` (→ `instantiates` edges).
- Same-name ambiguity (e.g. 8 path-distinct `f()` overrides in one file) is left unresolved rather than picked arbitrarily — matches the cross-file resolver's single-match rule. `..()` super-calls are skipped because resolving them needs corpus-wide type hierarchy unavailable at single-file extract time.
- Unresolved `#include` paths get `"external": true` on the edge so consumers reading raw extractions can skip them without re-deriving the check (build.py:160 already drops dangling targets at graph-build time).

**Commit 2 — `.dmi` / `.dmm` / `.dmf` (project assets, no grammar dep)**

- **`.dmi`** (icon sheets): PNG with a zTXt "Description" chunk holding BYOND state metadata. `extract_dmi` parses it with stdlib `struct`+`zlib`, emits one node per icon state name — the same names DM code references via `icon_state = "X"`.
- **`.dmm`** (map files): tile dictionary entries name the types composing each tile. `extract_dmm` parses the dictionary (skipping the grid), splits on top-level commas, strips `{var=val}` overrides, emits `uses` edges from map → types. Format reference: [`SpacemanDMM/crates/dmm-tools/src/dmm/read.rs`](https://github.com/SpaceManiac/SpacemanDMM/blob/master/crates/dmm-tools/src/dmm/read.rs). `tree-sitter-dm` claims `.dmm` support but errors on the `"key" = (...)` syntax in practice, hence the custom parser.
- **`.dmf`** (interface forms): hierarchical key-value text. `extract_dmf` emits window + elem nodes with the BYOND control type baked into the elem label (e.g. `elem "map" [MAP]`).

All three are pure stdlib — they don't need the `dreammaker` extra.

## Test plan

- [x] `uv sync --extra dreammaker && uv run --with pytest pytest tests/test_languages.py -k "dm or dmi or dmm or dmf"` — 26 pass
- [x] `uv sync && uv run --with pytest pytest tests/test_languages.py -k "dm or dmi or dmm or dmf"` — 13 DM tests skip cleanly via `@needs_dm`, 13 BYOND-format tests pass
- [x] `uv run pytest tests/` — 792 passing without the extra (8 pre-existing SQL failures unchanged)
- [x] Stress: extracted all 660 `.dm`/`.dme` files in [spacestation13/dm-test-suite](https://github.com/spacestation13/dm-test-suite) — 0 errors
- [x] Stress: all `.dmi` / `.dmf` files in dm-test-suite parse without errors
- [x] End-to-end: `uv run graphify update <dm-corpus>` produces a graph with correct type nodes (`/datum/object`), proc nodes (`/datum/object/New()`), and clustered communities
- [ ] CI green on Python 3.10 + 3.12 — DM tests will skip until either CI installs the `dreammaker` extra or upstream cuts 0.25.1

## ARCHITECTURE.md checklist

1. [x] Added `extract_dm`, `extract_dmi`, `extract_dmm`, `extract_dmf` in `graphify/extract.py`
2. [x] Registered `.dm`, `.dme`, `.dmi`, `.dmm`, `.dmf` in `_DISPATCH` + `collect_files`
3. [x] Added all five suffixes to `CODE_EXTENSIONS` in `detect.py` (watch.py picks them up automatically)
4. [x] Added `tree-sitter-dm` to `pyproject.toml` (as an optional `dreammaker` extra)
5. [x] Added fixtures (`sample.dm`, `sample.dmi`, `sample.dmm`, `sample.dmf`) and tests in `tests/test_languages.py`

## Related

- pytest as a uv dev dep is split out into #885.


safishamsi (commented):
Exciting to see DreamMaker support! Two blocking issues before this can land:

**1. Dangling include-edge targets**
When an `#include` points to a file not in the corpus (e.g. engine headers, external libs), the edge target node doesn't exist in the graph — this creates orphan edges that break graph traversal and dedup. Fix: check whether the target node ID exists before emitting the edge, or mark it with `"external": true` so downstream code can skip it.

Example guard:
```python
if target_id in known_node_ids:
    edges.append({source: src_id, target: target_id, ...})
```

**2. Git dependency blocks PyPI publishing**
`tree-sitter-dreammaker @ git+...` in the dependency list means we can't publish to PyPI (PyPI rejects git deps). Please either:
- Publish `tree-sitter-dreammaker` to PyPI and depend on the PyPI release, or
- Make it an optional extra (e.g. `graphifyy[dreammaker]`) with a try/except import guard in the extractor, same pattern as our other optional grammars

Happy to help with either approach — just let us know.

kanya-approve:
> Exciting to see DreamMaker support! Two blocking issues before this can land:
> 
> **1. Dangling include-edge targets** When an `#include` points to a file not in the corpus (e.g. engine headers, external libs), the edge target node doesn't exist in the graph — this creates orphan edges that break graph traversal and dedup. Fix: check whether the target node ID exists before emitting the edge, or mark it with `"external": true` so downstream code can skip it.
> 
> Example guard:
> 
> ```python
> if target_id in known_node_ids:
>     edges.append({source: src_id, target: target_id, ...})
> ```
> 
> **2. Git dependency blocks PyPI publishing** `tree-sitter-dreammaker @ git+...` in the dependency list means we can't publish to PyPI (PyPI rejects git deps). Please either:
> 
> * Publish `tree-sitter-dreammaker` to PyPI and depend on the PyPI release, or
> * Make it an optional extra (e.g. `graphifyy[dreammaker]`) with a try/except import guard in the extractor, same pattern as our other optional grammars
> 
> Happy to help with either approach — just let us know.

Made tree-sitter-dm an optional dep and labeling files outside of corpus as external now

safishamsi:
Implemented manually on v8 — the PR had merge conflicts with recent changes (uv migration, `.svh`, OpenCode path fix, cluster remap). All four extractors landed: `extract_dm` (tree-sitter-dm, covers `.dm`/`.dme`), `extract_dmi` (PNG tEXt/zTXt chunk parser for icon states), `extract_dmm` (tile dictionary parser, `uses` edges), `extract_dmf` (window/elem/control-type hierarchy). Fixture files + 26 tests added, all green. `tree-sitter-dm` added to `pyproject.toml` dependencies. Thanks for the thorough contribution!

PR #886, opened 2026-05-15
chore: commit uv.lock for reproducible installs

kanya-approve:
## Summary
- Removes `uv.lock` from `.gitignore` and commits the lockfile.
- graphify is distributed as a CLI (`uv tool install graphifyy` / `uvx graphifyy`), so committing the lockfile (application convention) fits better than ignoring it (library convention) — contributors and CI get the same resolved dependency tree.

## Why now
`uv.lock` was added to `.gitignore` in `d4b24d8` as part of a bulk "repo cleanup" with no per-change rationale. The current behavior makes every contributor's resolution non-deterministic.

## Test plan
- [x] `uv sync` succeeds on a fresh clone and produces the same resolved tree as the committed lockfile.


kanya-approve:
@safishamsi - why revert this? The message says it's a library, but I thought this was a cli tool with no intent to be used as a library?

https://github.com/safishamsi/graphify/commit/6d54b25d20e900af1dd38d5729ad3ba636c01b65

PR #889, opened 2026-05-16
Exit non-zero when all semantic-extraction chunks fail

jattree:
When `graphify extract --backend claude` runs on a machine where the `anthropic` package isn't installed (which it isn't by default - `pip install graphifyy` doesn't pull it in), every semantic-extraction chunk fails inside `extract_corpus_parallel`. The per-chunk failures print to stderr but the function still returns the (empty) merged accumulator, so `extract` proceeds to write an AST-only `graph.json` and exit 0. CI that checks exit status sees success.

Reproduces cleanly:

```
pip install graphifyy
graphify extract /some/repo --backend claude --out .
# stderr: [graphify] chunk N/M failed: Claude direct extraction requires the anthropic package. Run: pip install anthropic
# stdout: [graphify extract] wrote .../graph.json: NNN nodes, ...
# echo $?  => 0
```

The fix tracks per-chunk success via the existing `on_chunk_done` callback (which only fires after a chunk succeeds - `extract_corpus_parallel` swallows per-chunk exceptions and continues). If fresh extraction was requested (`uncached_paths` non-empty) and no chunks completed, abort with exit 1 before the merge / cluster / write phase, with a message that names the backend and points at the likely-missing SDK.

Same shape covers the other backends that don't pull their SDK as a hard dep (`openai`, `google-generativeai`, etc.). Cached-only runs are unaffected - `uncached_paths` is empty and the guard doesn't fire.

### Tests

`tests/test_extract_cli.py` patches `graphify.llm.extract_corpus_parallel` in-process and simulates:

- All chunks failed (callback never invoked): asserts `SystemExit.code == 1` and no `graph.json` written.
- One chunk succeeded (callback fires once): asserts exit 0 and `graph.json` exists.

Full suite: 821 pass, 11 pre-existing failures unchanged.

### DIFF
--- a/graphify/__main__.py
+++ b/graphify/__main__.py
@@ -2598,9 +2598,12 @@ def main() -> None:
 
                 # Minimal progress callback so the CLI is no longer silent
                 # during long local-inference runs (issue #792 addendum).
-                _total_chunks = {"n": 0}
+                # Also track per-chunk success so we can fail loudly when
+                # every chunk errors (e.g. missing backend SDK package).
+                _chunk_stats = {"total": 0, "succeeded": 0}
                 def _progress(idx: int, total: int, _result: dict) -> None:
-                    _total_chunks["n"] = total
+                    _chunk_stats["total"] = total
+                    _chunk_stats["succeeded"] += 1
                     print(
                         f"[graphify extract] chunk {idx + 1}/{total} done",
                         flush=True,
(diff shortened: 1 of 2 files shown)


safishamsi (commented):
This is a solid fix — the 'all chunks failed' threshold is correct, exit is placed before graph.json write, and the two tests cover the right cases. One thing to fix before we merge: the error message on line 2634 contains an em dash (`—`). Project style uses hyphens or colons instead. Please change:

```
({len(uncached_paths)} uncached files) — see per-chunk errors above
```
to:
```
({len(uncached_paths)} uncached files) - see per-chunk errors above
```

Fix that and we'll merge.

Issue #866, opened 2026-05-14
Add tree-sitter parsers for bash (.sh) and JSON (.json)


mattmillartech:
## Summary

Graphify currently doesn't index `.sh` (bash) or `.json` files at the AST level — these extensions aren't in the extension→extractor map in `graphify/extract.py`, and the corresponding `tree-sitter-bash` / `tree-sitter-json` packages aren't dependencies.

Result: graphs of repositories that include hook scripts, CI configs, MCP configs, package manifests, and other operationally-relevant files have no nodes for any of that content. `graphify get_node "scripts/hooks/my-hook.sh"` returns "not found" even when the file exists.

## Reproduction

In a fresh graphify venv:

```bash
# Verify parsers
pip list | grep -E "tree.?sitter"
# No tree_sitter_bash or tree_sitter_json

# Verify extension map
grep -E "\.(sh|json)" graphify/extract.py
# No .sh entry; only stray .json references for tsconfig.json path resolution

# Verify empirical
graphify update /path/to/repo-with-shell-scripts
graphify query "any-bash-symbol-in-the-repo"   # zero results
```

## Why this matters

For users on AI agent platforms (Claude Code, Codex, Cursor, etc.), a significant portion of the "operating environment" lives in shell + JSON:

- `.claude/settings.json` — hook registration, MCP servers, permissions
- `scripts/hooks/*.sh` — PreToolUse / PostToolUse / Stop hooks
- `.mcp.json` — MCP server configuration
- `package.json` — script entry points

These files contain real cross-references (a `settings.json` matcher points at a hook script; a hook script invokes another shell script; a `package.json` script references a `.sh` entry-point). Without bash/json indexing, queries like "how does PreToolUse on Bash get gated?" return zero hits — agents fall back to manual file walks.

## Proposed fix

1. Add `tree-sitter-bash` and `tree-sitter-json` to `pyproject.toml` dependencies.
2. Register the parsers in graphify's language map (whichever file holds the tree-sitter language registration).
3. Add extension→extractor entries:
   ```python
   ".sh": extract_bash,
   ".bash": extract_bash,
   ".json": extract_json,
   ```
4. Write `extract_bash` (functions + top-level commands) and `extract_json` (top-level keys + nested object refs) — both can be lightweight; the goal is file + symbol nodes, not deep semantic analysis.

## Workaround

For users who can't wait, a workspace-local monkey-patch:

```python
# scripts/persistence/graphify-bash-json-shim.py
[...]

PR #890, opened 2026-05-16
Reduce headline-output noise in god_nodes and surprising_connections

jattree:
Two ranking-policy changes after running graphify on a Go monorepo with both READMEs and JSON fixtures. Both reduce noise in the headline-output sections; both are conservative.

### 1. Extend cross-language INFERRED suppression to code ↔ doc

d14e8a7 added a suppression for INFERRED `calls`/`uses` edges between source files in different language families. `_cross_language()` consults `_LANG_FAMILY`, which only contains programming-language extensions, so the gate doesn't fire when one endpoint is a `.md` / `.txt` / `.rst` doc.

On a Go monorepo with READMEs, the top "Surprising Connections" came out like:

```
HandleVideoProcessing() --[calls]--> NewVideoProcessor    [INFERRED]
libs/transcoder/worker.go -> services/.../Readme.md
```

The LLM saw the symbol name in the README's prose and emitted a `calls` edge. Cross-dir (+2) and cross-community (+1) bonuses promoted these into the top-5, crowding out real structural surprises. Same failure shape as the rsl-siege-manager Python ↔ TypeScript case d14e8a7 addressed.

Apply the same `_suppress_structural` treatment when `file_category` is `doc` on at least one side. Preserves:

- `semantically_similar_to` (explicit LLM insight, per d14e8a7 doctrine)
- `EXTRACTED` edges (real references the doc literally contains)
- code ↔ paper (`.pdf` is `paper` category, a genuine cross-format signal)

`_file_category` falls back to `"doc"` for unknown extensions, so this also catches code ↔ unknown-extension INFERRED calls/uses. That matches the resolver-pollution profile and is covered by a test documenting the behavior.

### 2. Filter common JSON keys from god_nodes

The JSON extractor (#866) emits one node per JSON key, so common keys recurring across sibling records (`start`, `end`, `name`, `id`, `type`, `properties`, `value`, `key`, `data`, `items`, `title`, `description`, `version`) accumulate positional degree, not architectural meaning.

On the Go corpus, a `testhelpers/dates.json` file containing terms with `start`/`end` timestamps put both labels at degree 95 in god_nodes; `properties` and `name` from other JSON fixtures took two more top-10 slots. Four of the top-10 god_nodes were positional-key noise. After the filter, those slots are taken by real Go method abstractions (`ToLower`, `Process()`, `ZoneString()`, `List()`).

Gate is conjunctive: `source_file` must end in `.json` AND normalized label must be in the noise set. `start()` defined in a Go file still ranks; `ProductionEsConfig` in `elasticsearch.json` still ranks; only the generic-key + json combination is filtered. Case-insensitive on the label.

Set is intentionally narrow (13 labels). Not made configurable to avoid adding a knob for a ranking heuristic.

### Tests

10 new tests in `test_analyze.py`:

- code ↔ doc INFERRED `calls`/`uses` suppressed
- code ↔ doc EXTRACTED preserved (real references stay)
- code ↔ doc INFERRED `semantically_similar_to` preserved (d14e8a7 doctrine)
- code ↔ paper INFERRED `calls` preserved (genuine cross-format)
- code ↔ unknown-extension INFERRED suppressed (documents fallback behavior)
- god_nodes filters massive-degree JSON `start` while keeping real `DB()`
- god_nodes keeps `start` when defined in a `.go` source
- god_nodes keeps domain entity in a JSON source
- JSON-key filter is case-insensitive

Full suite: 821 pass, 11 pre-existing failures unchanged.

### Note on scope

Both changes touch `analyze.py` and address the same "headline output got noisy" failure mode. Happy to split into two PRs if preferred.

### DIFF
--- a/graphify/analyze.py
+++ b/graphify/analyze.py
@@ -65,17 +65,43 @@ def _is_file_node(G: nx.Graph, node_id: str) -> bool:
     return False
 
 
+# JSON key noise: the JSON extractor emits one node per key, so common
+# keys (start/end/name/properties/...) that recur across sibling records
+# in one file accumulate positional degree rather than architectural
+# meaning. Filter them out of god_nodes. Set is intentionally narrow.
+_JSON_NOISE_LABELS = frozenset({
+    "start", "end", "name", "id", "type", "properties", "value", "key",
+    "data", "items", "title", "description", "version",
+})
+
+
+def _is_json_key_node(G: nx.Graph, node_id: str) -> bool:
+    """True if this node is a common JSON key whose degree is positional noise."""
+    attrs = G.nodes[node_id]
+    src = (attrs.get("source_file") or "").lower()
+    if not src.endswith(".json"):
+        return False
+    label = (attrs.get("label") or "").strip().lower()
+    return label in _JSON_NOISE_LABELS
+
+
 def god_nodes(G: nx.Graph, top_n: int = 10) -> list[dict]:
     """Return the top_n most-connected real entities - the core abstractions.
 
     File-level hub nodes are excluded: they accumulate import/contains edges
     mechanically and don't represent meaningful architectural abstractions.
+
+    Common JSON keys (start, end, name, id, type, ...) extracted from .json
+    files are also excluded: their degree comes from being referenced by every
+    sibling record in the same file, not from being architecturally central.
     """
     degree = dict(G.degree())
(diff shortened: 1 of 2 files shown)


safishamsi:
Both changes implemented directly on v8 in commits aa541a6 and 766566f — no rebase needed. Full test coverage including all 10 cases from the PR description (semantically_similar_to preservation, unknown-extension fallback, case-insensitive JSON key filter). Closing as superseded.

PR #892, opened 2026-05-16
feat(bash): harden extractor — literal filtering, entrypoint nodes, broader skip list

hypnwtykvmpr:
Builds on #866 (tree-sitter bash + JSON extractors). Three correctness/security improvements to the bash extraction in `graphify/extract.py`:

## Changes

**1. `literal()` filtering (security)** — command names containing shell metacharacters (`$`, `` ` ``, `$(`, `<(`, `>`, `|`, `;`, `&`) are now rejected at parse time. Previously, command substitutions like `$(rm -rf /)` were tokenized as command names and could pollute the graph or trigger graph-resolution paths the extractor wasn't designed for. Strips matched outer quotes; returns `None` for anything that looks like a substitution, expansion, or compound construct rather than a plain command name.

**2. Entrypoint node** — every `.sh` file now produces both a `file` node and a `bash_entrypoint` node, with a `contains` edge between them. Top-level commands in the script attach to the entrypoint via a new top-level call walk, rather than being orphaned. Matches the entrypoint pattern other-language extractors use. Node metadata gains `language` + `kind` for downstream filtering.

**3. Broader skip list** — `_BASH_SKIP` (~30 entries) replaced with `_BASH_EXTERNAL_OR_BUILTIN_COMMANDS` (~85 entries), separating bash builtins/keywords (alias, declare, mapfile, etc.) from common external commands (awk, jq, curl, sed, etc.). Prevents external programs from being recorded as user-defined function calls when no matching function exists, reducing false-positive call edges in the graph.

## Implementation notes

- `walk_calls` now skips nested `function_definition` children so calls inside nested functions aren't double-counted at the enclosing scope.
- A separate top-level `walk_calls(root, entry_nid, ...)` pass attributes top-level command calls to the entrypoint node.
- `_BASH_NAME_RE`, `_BASH_SOURCE_COMMANDS`, `_BASH_EXTERNAL_OR_BUILTIN_COMMANDS` lifted to module-level inside `extract_bash` so they're declared once per call.

## Devtools (separate concern in same commit)

`pyproject.toml` gains `[dependency-groups] dev` (ruff, pyright, pre-commit, hypothesis, pip-audit) plus minimal `[tool.ruff]`, `[tool.ruff.lint]`, `[tool.pyright]` configs targeting py311. Bundled here rather than as a standalone PR since these are the tools that verify this change locally.

## Verification

- 821/821 tests pass (`pytest -p no:cacheprovider`)
- 10/10 bash-relevant tests pass (`pytest -k bash`)
- No new pyright errors at the bash extractor lines

## Diff

86 insertions, 16 deletions across 2 files.

### DIFF
--- a/graphify/extract.py
+++ b/graphify/extract.py
@@ -5740,11 +5740,12 @@ def extract_bash(path: Path) -> dict:
     function_bodies: list[tuple[str, Any]] = []
     defined_functions: set[str] = set()
 
-    def add_node(nid: str, label: str, line: int) -> None:
+    def add_node(nid: str, label: str, line: int, kind: str = "code") -> None:
         if nid and nid not in seen_ids:
             seen_ids.add(nid)
             nodes.append({"id": nid, "label": label, "file_type": "code",
-                          "source_file": str_path, "source_location": f"L{line}"})
+                          "source_file": str_path, "source_location": f"L{line}",
+                          "metadata": {"language": "bash", "kind": kind}})
 
     def add_edge(src: str, tgt: str, relation: str, line: int,
                  confidence: str = "EXTRACTED", weight: float = 1.0,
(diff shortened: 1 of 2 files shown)


hypnwtykvmpr:
Closing — internal review (Codex audit `.dox/2026-05-16-codex-audit-cycle-2.1-bash-commit.md`) found two blockers in this implementation: (1) the literal() filter doesn't detect command substitutions because `$(cmd)` exposes `cmd` as a child node with no metacharacters in its token, and (2) the broader skip list is checked before `name in defined_functions` so user-defined functions named like external commands are incorrectly suppressed. No tests cover the new behavior. Will re-open as a clean PR after fixing both issues and adding regression tests. Apologies for the noise.

PR #893, opened 2026-05-16
feat(bash): harden extractor — AST-ancestry-aware command detection, entrypoint nodes

hypnwtykvmpr:
Builds on #866 (tree-sitter bash + JSON extractors). Two correctness/security improvements to bash extraction in `graphify/extract.py`. (Supersedes my earlier #892 which had two blockers caught in internal review — fixed here with regression tests.)

## Changes

**1. AST-ancestry-aware command detection (security)** — token-level filtering misses constructs like `$(build)` because tree-sitter exposes `build` as a child node of `command_substitution` whose own token has no metacharacters. Added `is_inside_expansion(node)` helper that walks `node.parent` until it finds `command_substitution` or `process_substitution`. Used as a gate in both `walk` and `walk_calls`. Combined with a token-level `literal()` filter that rejects names containing shell metacharacters (`$`, backtick, `$(`, `<(`, redirections, pipes, sequencers). Prevents constructs like `$(rm -rf /)` from being tokenized as command names and polluting the graph.

**2. Entrypoint node** — every `.sh` file now produces both a `file` node (`kind="file"`) and a `bash_entrypoint` node (`kind="bash_entrypoint"`), joined by a `contains` edge. A separate top-level `walk_calls(root, entry_nid, ...)` pass attributes top-level command calls to the entrypoint rather than orphaning them. Matches the entrypoint pattern other-language extractors use. Node metadata gains `language` + `kind` for downstream filtering.

## Implementation notes

- `walk_calls` skips nested `function_definition` children so calls inside nested functions aren't double-counted at enclosing scope.
- Resolved-call resolution uses `defined_functions` as the only filter for call edges. User-defined functions named like external commands (`install`, `find`, `git`, etc.) are correctly recorded — an external-builtin skip list that I had in an earlier iteration was creating false negatives for shadowing functions and is intentionally not included. A skip list belongs with raw/unresolved call recording (potential future PR).
- `_BASH_SOURCE_COMMANDS = frozenset({"source", "."})` lifted to a constant inside `extract_bash`.

## Devtools (bundled)

`pyproject.toml` gains `[dependency-groups] dev` (ruff, pyright, pre-commit, hypothesis, pip-audit) plus minimal `[tool.ruff]`, `[tool.ruff.lint]`, `[tool.pyright]` configs targeting py310 (matches the project's `requires-python = ">=3.10"`). Bundled here rather than as a standalone PR — these are the tools that verify this change locally.

## Verification

```
$ pytest tests/test_extract.py -k bash --tb=short -q
15 passed, 43 deselected in 0.12s

$ pytest -p no:cacheprovider -q
826 passed, 2 warnings in 17s
```

5 new regression tests (in addition to the 10 existing bash tests):
- `test_extract_bash_rejects_command_substitution_as_call`
- `test_extract_bash_process_substitution_not_recorded`
- `test_extract_bash_shadowing_function_is_recorded`
- `test_extract_bash_creates_entrypoint_node`
- `test_extract_bash_top_level_call_attributes_to_entrypoint`

## Diff size

183 insertions, 17 deletions across 3 files (`graphify/extract.py`, `tests/test_extract.py`, `pyproject.toml`).

### DIFF
--- a/README.md
+++ b/README.md
@@ -98,6 +98,20 @@ graphify install
 
 That's it. Open your AI assistant and type `/graphify .`
 
+To install the assistant skill into the current repository instead of your user
+profile, add `--project`:
+
+```bash
+graphify install --project
+graphify install --project --platform codex
+```
+
+Project-scoped installs write under the current directory, for example
+`.claude/skills/graphify/SKILL.md` or `.agents/skills/graphify/SKILL.md`, and
+print a `git add` hint for files that can be committed.
+Per-platform commands that support project-scoped installs accept the same flag,
+for example `graphify claude install --project` or `graphify codex install --project`.
+
 > **PowerShell note:** Use `graphify .` not `/graphify .` — the leading slash is a path separator in PowerShell.
 
 > **`graphify: command not found`?** Use `uv tool install graphifyy` or `pipx install graphifyy` — both put the CLI on PATH automatically. With plain `pip`, add `~/.local/bin` (Linux) or `~/Library/Python/3.x/bin` (Mac) to your PATH, or run `python -m graphify`.
@@ -433,6 +447,7 @@ graphify install  # overwrites the skill file
 
 graphify uninstall                 # remove from all platforms in one shot
 graphify uninstall --purge         # also delete graphify-out/
+graphify uninstall --project --platform codex  # remove project-scoped install files only
 
 graphify hook install              # post-commit + post-checkout hooks
 graphify hook uninstall
--- a/graphify/__main__.py
+++ b/graphify/__main__.py
@@ -67,6 +67,110 @@ def _refresh_all_version_stamps() -> None:
         if skill_dst.exists():
             vf.write_text(__version__, encoding="utf-8")
 
+
+def _platform_skill_destination(platform_name: str, *, project: bool = False, project_dir: Path | None = None) -> Path:
+    """Return the skill destination for a platform and scope."""
+    if platform_name == "gemini":
+        if project:
+            return (project_dir or Path(".")) / ".gemini" / "skills" / "graphify" / "SKILL.md"
+        if platform.system() == "Windows":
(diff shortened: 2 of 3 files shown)


safishamsi (commented):
The core idea (AST-ancestry-aware filtering + entrypoint nodes) is sound and well-motivated. Three bugs need fixing before this can merge:

**Bug 1 — entrypoint node ID collision**
`_make_id(stem, "__script__")` normalises to `{stem}_script` because `_make_id` strips leading/trailing underscores from each part. Any bash function literally named `script` produces the same ID, causing the function node to be silently dropped and its edges to point at the entrypoint instead.

Fix: use a sentinel that survives `_make_id`'s normalization, e.g. pass a third part: `_make_id(stem, "entrypoint", "bash")`, or bypass `_make_id` entirely for the entrypoint and construct the ID directly.

**Bug 2 — nested function definitions disappear**
The structural `walk()` returns immediately on `function_definition` without recursing into bodies, so nested defs are never added to `function_bodies`. The PR then adds `if child.type == "function_definition": continue` to `walk_calls`, which also skips them in the call-resolution pass. Net result: nested function definitions produce no node, no `defines` edge, and no call edges — worse than pre-PR.

Fix: either have `walk()` recurse into function bodies to discover nested defs, or replace the `continue` in `walk_calls` with a recursive call that walks the nested body under its own scope.

**Bug 3 — scope misrepresentation**
The PR title says `feat(bash)` and the description claims 3 files / 183 insertions. The actual diff is 5 files / 641 lines — the `detect.py` shebang parser rewrite (+165) and `pyproject.toml` tooling config (+26) are substantial undisclosed additions. Please split those into separate PRs so each change can be reviewed on its own merits.

The `source`/`." shadowing inconsistency (same call site can now emit both a `calls` edge from `walk_calls` and an `imports` edge from `walk`) is a fourth issue worth addressing, but we'd consider that a follow-up if the three above are fixed.

Happy to help if you want to talk through any of these.

PR #894, opened 2026-05-16
feat(detect): parse macOS/BSD and GNU env(1) shebang option forms

hypnwtykvmpr:
Upstream's `_shebang_file_type` parses shebangs via `line[2:].split()` and only handles `#!/usr/bin/env <interp>` (single-arg env). Files using documented env(1) option forms are silently classified as non-code.

### POSIX/macOS short forms
- `env -S "python3 -u"` (split-args)
- `env -i bash` (no-operand flag)
- `env -u NAME python3` (option + required operand)
- `env -C DIR python3` (workdir + operand)
- `env -P PATH python3` (utilpath + operand)
- `env DEBUG=1 python3` (inline `NAME=value`)
- Clumped `-uNAME` / `-CDIR` / `-PPATH` / `-aARG`

### GNU coreutils long/compact forms
- `env --split-string='python3 -u'` (= operand)
- `env --split-string "python3 -u"` (separate operand)
- `env -Spython3 -u` (compact -S)
- `env -vSpython3 -u` (-v + compact -S)
- `env -a alias python3` and `env --argv0=alias python3`
- `env --unset NAME python3` and `--unset=NAME`
- `env --chdir DIR python3` and `--chdir=DIR`
- `env --default-signal[=SIG] / --ignore-signal[=SIG] / --block-signal[=SIG] python3`
- `env --ignore-environment / --null / --debug / --list-signal-handling python3`

### GNU `-S` payload contents (per the documented shebang synopsis)
Per the GNU coreutils manual, the shebang synopsis is:

    #!/usr/bin/env -[v]S[option]... [name=value]... command [args]...

So `-S` / `--split-string` payloads can carry env options and assignments BEFORE the interpreter:
- `env -S PYTHONPATH=/opt/custom python3` (assignment before interp)
- `env -S -i OLDUSER=\${USER} python3` (flag + assignment before interp)
- `env --split-string='-i python3 -u'` (long-form variant)

## Implementation

Added `_shebang_interpreter(path) -> str | None` using `shlex.split` plus a small `_env_command_args(args, *, allow_split=True)` parser with an explicit option table.

`-S` / `--split-string` payloads are tokenized then **recursively re-parsed as env args** with `allow_split=False`. The flag bounds recursion at depth 1 — a nested `-S` inside a `-S` payload becomes an unknown option and yields `None`, which prevents stack blowup on malicious or weird input.

**Unknown hyphen-prefixed options return `None`** rather than misclassifying their operand as the interpreter. This is the key safety property: we refuse to guess.

`_shebang_file_type` becomes a 4-line wrapper. Read buffer raised 128 → 256 bytes for longer env -S strings.

## Verification

```
$ pytest tests/test_detect.py -k shebang --tb=short -q
32 passed, 43 deselected in 0.16s

$ pytest -p no:cacheprovider -q
853 passed, 2 warnings in 25s
```

32 regression tests covering:
- Every documented env spelling (POSIX/macOS shorts + GNU longs + compact + `=` vs separate-operand variants)
- `-S` payload assignments and flags before the interpreter (GNU's documented shebang form)
- Nested split-string rejection (recursion bound)
- Failure modes: no shebang, unreadable file, missing operand, unknown option

## Diff size

458 insertions, 12 deletions across 2 files (`graphify/detect.py`, `tests/test_detect.py`).

### DIFF
--- a/graphify/detect.py
+++ b/graphify/detect.py
@@ -4,6 +4,7 @@ import fnmatch
 import json
 import os
 import re
+import shlex
 from enum import Enum
 from pathlib import Path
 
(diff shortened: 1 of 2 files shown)


hypnwtykvmpr:
Closing — filed from an unauthorized topic branch by mistake. Will re-file from origin v8 once both cycle 2.1 (already in #893) and this cycle 2.2 work are consolidated on v8.

Issue #907, opened 2026-05-17
hook install creates phantom directory on git < 2.31 — `--path-format=absolute` not recognized, output contaminates hooks path


itsnoproblem:
## Symptom

`graphify hook install` silently writes hooks into a phantom directory whose name contains an embedded newline, rather than into `.git/hooks/`. The real git hooks dir is never touched, so the hooks never fire. `graphify hook status` subsequently reports `not installed` even though the install command exited 0.

After running `graphify hook install` in my repo:

```console
$ find ./ -maxdepth 2 -name '--path-format*' -print | od -c | head -2
0000000    .   /   -   -   p   a   t   h   -   f   o   r   m   a   t   =
0000020    a   b   s   o   l   u   t   e  \n   .   g   i   t  \n

$ ls -R "./--path-format=absolute"$'\n'".git"
hooks
post-commit  post-checkout   # ← the graphify hooks, written into a dead dir
```

The directory name is literally `--path-format=absolute\n.git` (embedded `\n`).

## Root cause

This is a regression of the fix for #865 / #516. In `graphify/hooks.py:_hooks_dir`:

```python
# graphify/hooks.py:204
res = _sp.run(
    ["git", "-C", str(root), "rev-parse", "--path-format=absolute", "--git-path", "hooks"],
    capture_output=True, text=True,
)
if res.returncode == 0:
    d = Path(res.stdout.strip())
    d.mkdir(parents=True, exist_ok=True)
    return d
```

The `--path-format` option was added to `git rev-parse` in [git 2.31](https://github.com/git/git/blob/v2.31.0/Documentation/RelNotes/2.31.0.txt) (March 2021). On older git, it is *not* rejected with a non-zero exit — instead, it is interpreted as a literal rev/path argument and echoed to stdout, followed by the normal `--git-path hooks` output:

```console
$ git --version
git version 2.23.0

$ git rev-parse --path-format=absolute --git-path hooks; echo "exit=$?"
--path-format=absolute
.git/hooks
exit=0
```

`res.stdout.strip()` only strips leading/trailing whitespace, so the embedded `\n` survives:

```python
>>> Path("--path-format=absolute\n.git/hooks").mkdir(parents=True, exist_ok=True)
# creates ./--path-format=absolute<LF>.git/hooks  on POSIX (newline is a legal filename char)
```

[...]

PR #910, opened 2026-05-17
test(hooks): cover old git hook path output

balloon72:
## Summary
- Add regression coverage for relative hooks paths from `git rev-parse --git-path hooks`.
- Cover contaminated multiline output from older git versions.
- Cover absolute hooks paths returned by git.

## Tests
- `python -m pytest tests/test_hooks.py`

## Notes
Regression coverage for #907; the runtime fix is already present on `v8`.


safishamsi:

PR #911, opened 2026-05-17
fix(opencode): remove invalid general-purpose agent guidance

### LINKED ISSUE #825: OpenCode integration uses invalid `general-purpose` agent and blocks on large corpus prompt
## Summary

When using Graphify with OpenCode, the installed `SKILL.md` appears to contain instructions that are not fully compatible with OpenCode.

There are two related issues:

1. The skill instructs the model to use `subagent_type="general-purpose"`, but OpenCode does not have a `general-purpose` subagent type.
2. For large repositories, the skill stops and asks the user to choose a subfolder instead of continuing with a deterministic OpenCode-compatible policy.

This makes `/graphify .` fail or become interactive in OpenCode environments, especially in offline/self-hosted model setups.

## Environment

- Platform: OpenCode
- Graphify version: latest available version
- Install command:

```bash
graphify install --platform opencode
Issue 1: Invalid OpenCode subagent type
In OpenCode, this fails with:

Task Extract chunk 1 of 26
Error: Unknown agent type: general-purpose is not a valid agent type

balloon72:
## Summary
- Remove OpenCode guidance that referenced the unsupported `general-purpose` agent type.
- Replace missing chunk recovery with OpenCode @agent retry, smaller chunk, or serial fallback guidance.
- Avoid stopping large OpenCode runs for an interactive subfolder prompt.

## Tests
- `python -m pytest tests/test_install.py tests/test_install_strings.py`

## Notes
Closes #825

### DIFF
--- a/graphify/skill-opencode.md
+++ b/graphify/skill-opencode.md
@@ -110,7 +110,7 @@ Omit any category with 0 files from the summary.
 Then act on it:
 - If `total_files` is 0: stop with "No supported files found in [path]."
 - If `skipped_sensitive` is non-empty: mention file count skipped, not the file names.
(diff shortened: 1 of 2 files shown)


safishamsi:

PR #912, opened 2026-05-17
docs: clarify code-only semantic extraction

### LINKED ISSUE #836: Do code files need semantic extract?
https://github.com/safishamsi/graphify/blob/v7/docs/how-it-works.md:
'''
Pass 1 — Code structure (free, no API calls) Tree-sitter parses your code files and extracts classes, functions, imports, call graphs, and inline comments. This runs locally with no LLM involved. 25 languages supported. SQL files get special treatment: tables, views, foreign keys, and JOIN relationships are extracted deterministically.

Pass 3 — Docs, papers, images (Claude subagents, costs tokens) Claude runs in parallel over markdown, PDFs, images, and transcripts. Each subagent reads a batch of files and outputs a JSON fragment: nodes, edges, and any group relationships. The fragments are merged into a single graph. 
‘’‘

https://github.com/safishamsi/graphify/blob/v7/graphify/skill.md:
'''
Code files: focus on semantic edges AST cannot find (call relationships, 
shared data, arch patterns). Do not re-extract imports - AST already has those.
'''

does code file need semantic extract?
infact , all codes are sended to llm for semantic extract?

balloon72:
## Summary
- Clarify that code files are handled by the local Tree-sitter/deterministic extraction path.
- State that code-only corpora skip the semantic LLM pass.
- Keep semantic extraction scoped to docs, papers, images, and transcripts.

## Tests
- `python -m pytest tests/test_install_strings.py`

## Notes
Closes #836

### DIFF
--- a/docs/how-it-works.md
+++ b/docs/how-it-works.md
@@ -7,6 +7,8 @@ graphify processes your files in three passes:
 **Pass 1 — Code structure (free, no API calls)**
 Tree-sitter parses your code files and extracts classes, functions, imports, call graphs, and inline comments. This runs locally with no LLM involved. 25 languages supported. SQL files get special treatment: tables, views, foreign keys, and JOIN relationships are extracted deterministically.
 
+Code files are not sent to the LLM semantic extractor in the normal pipeline. If a corpus contains only code files, Pass 3 is skipped entirely; semantic extraction is reserved for docs, papers, images, and transcripts.
+
 **Pass 2 — Video and audio (local, no API calls)**
 Video and audio files are transcribed with faster-whisper. To focus the transcript on your domain, the transcription prompt is seeded with your top god nodes (the most-connected concepts in your code graph so far). Transcripts are cached — re-runs skip already-processed files.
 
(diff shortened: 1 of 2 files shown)


safishamsi:
Thanks for addressing #836 — the second sentence (code-only corpora skip the LLM pass entirely) is accurate and worth documenting. The first sentence needs a correction though.

> Code files are not sent to the LLM semantic extractor in the normal pipeline.

This is only true for **code-only corpora**. In mixed corpora (code + docs/papers/images), code files do go through the LLM — `skill.md` line 318 explicitly says: *"Code files: focus on semantic edges AST cannot find (call relationships, shared data, arch patterns)."* The current wording would mislead users with a mixed corpus.

Suggested replacement:

> In a code-only corpus, the LLM semantic pass is skipped entirely — Tree-sitter handles all extraction. In mixed corpora (code alongside docs, papers, or images), code files also go through the LLM, but only to extract semantic edges that AST cannot find (call relationships, shared data, architectural patterns) — never to re-extract imports or structure.

Update the test to match and this is good to go.

PR #913, opened 2026-05-17
fix(codex): keep graph-first guidance with dirty graph output

### LINKED ISSUE #860: Graphify Codex install does not reliably make the agent use Graphify first for orientation due to the graph files being dirty in workspace
## Description

I have Graphify installed for Codex with both `AGENTS.md` configured and the `.codex/hooks.json` `PreToolUse` hook registered. The install reports success and says Codex will check the knowledge graph before answering codebase questions and rebuild it after code changes.

In practice, the agent still skipped Graphify and went straight to targeted local `rg` / file inspection during a normal code-change task where Graphify should have been the first orientation step.

The main reason it gave was that `graphify-out/` files were dirty in the git worktree, even though those files are expected to change frequently and may often be dirty because of Graphify’s update flow and hooks.

The issue is that the current Codex integration does not make it reliable enough that the agent will still use Graphify first when `graphify-out/` is present but dirty.

## Observed Behavior

- `graphify codex install` reports successful setup
- `AGENTS.md` is configured
- `.codex/hooks.json` has the `PreToolUse` hook
- despite that, Codex may still bypass Graphify and inspect files directly first
- in this case, the agent treated dirty `graphify-out/` files as a reason not to use Graphify for orientation

## Expected Behavior

When `graphify-out/` exists, Codex should still reliably use Graphify as the default orientation step before broad search or multi-file inspection, even if `graphify-out/` is dirty.

Dirty graph artifacts should not be treated as a reason to skip Graphify unless the task is specifically about stale or incorrect graph output, or the user has said not to rely on it.

balloon72:
## Summary
- Make the Codex skill explicitly use Graphify query/explain/path for orientation when `graphify-out/` exists.
- Clarify that dirty graph artifacts are expected after hooks or incremental updates.
- Limit Graphify skipping to stale/incorrect graph tasks or explicit user direction.

## Tests
- `python -m pytest tests/test_install.py tests/test_install_strings.py`

## Notes
Closes #860

### DIFF
--- a/graphify/__main__.py
+++ b/graphify/__main__.py
@@ -311,6 +311,7 @@ When the user types `/graphify`, invoke the `skill` tool with `skill: "graphify"
 
 Rules:
 - For codebase questions, first run `graphify query "<question>"` when graphify-out/graph.json exists. Use `graphify path "<A>" "<B>"` for relationships and `graphify explain "<concept>"` for focused concepts. These return a scoped subgraph, usually much smaller than GRAPH_REPORT.md or raw grep output.
+- Dirty graphify-out/ files are expected after hooks or incremental updates; dirty graph files are not a reason to skip graphify. Only skip graphify if the task is about stale or incorrect graph output, or the user explicitly says not to use it.
 - If graphify-out/wiki/index.md exists, use it for broad navigation instead of raw source browsing.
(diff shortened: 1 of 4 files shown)


safishamsi:

PR #918, opened 2026-05-17
fix(detect): save_manifest drops untouched files on incremental runs

### LINKED ISSUE #917: save_manifest drops untouched files on incremental runs, forcing full re-extraction
## Summary

`save_manifest()` in `graphify/detect.py` rebuilds the manifest **only** from the `files` passed to it. On an incremental `extract`/`update` run that processes a small subset of the corpus, every other file's entry is dropped from the manifest.

The next `detect_incremental()` then sees those files as missing from the manifest and flags the **entire corpus** as new — forcing a full, expensive re-extraction.

## Where

`graphify/detect.py`, `save_manifest()` (currently ~line 843 on `v8`):

```python
existing = load_manifest(manifest_path)
manifest: dict[str, dict] = {}          # <-- starts empty
for file_list in files.values():
    for f in file_list:
        ...
        manifest[f] = entry             # <-- only this run's files land here
```

`existing` is loaded but used solely for per-file `prev` lookups; it never seeds the output `manifest`.

## Reproduce

1. Run a full build so the manifest has the whole corpus.
2. Run an incremental update touching only a few files.
3. Run `detect_incremental()` again — it reports the whole corpus as new.

## Fix

Seed `manifest` from the existing on-disk manifest (pruned to files that still exist) before the loop. Deleted files are still dropped, so deletions self-heal. PR attached.

dkauto2026-cyber:
## Summary

Fixes #917.

`save_manifest()` rebuilt the manifest only from the `files` passed to it, so an incremental `extract`/`update` run that processed a handful of files wiped every other file's entry. The next `detect_incremental()` then flagged the entire corpus as new, forcing a full re-extraction.

This seeds the output `manifest` from the existing on-disk manifest (pruned to files that still exist) before stamping this run's files.

## Test plan

- [ ] Full build → manifest contains whole corpus
- [ ] Incremental update touching a few files → manifest still contains the whole corpus, with touched files re-stamped
- [ ] Delete a file, run update → deleted file pruned from manifest
- [ ] `detect_incremental()` after an incremental run reports only changed files, not the whole corpus

### DIFF
--- a/graphify/detect.py
+++ b/graphify/detect.py
@@ -854,9 +854,18 @@ def save_manifest(
     kind="semantic" — written by `graphify extract` after semantic extraction.
                       Stamps semantic_hash; preserves existing ast_hash.
     kind="both"     — full pipeline: stamps both hashes (default).
+
+    The manifest is seeded from the existing on-disk manifest (pruned to files
+    that still exist) before stamping the files in this run. Without this, an
+    incremental run that only processed a handful of files would drop every
+    other file out of the manifest, so the next detect_incremental() would flag
+    the whole corpus as "new". Entries for deleted files are pruned, so
+    deletions still self-heal.
     """
     existing = load_manifest(manifest_path)
-    manifest: dict[str, dict] = {}
+    manifest: dict[str, dict] = {
+        f: e for f, e in existing.items() if Path(f).exists()
+    }
     for file_list in files.values():
         for f in file_list:
             try:


safishamsi:
Thanks for the fix — this is exactly right. We independently identified and shipped the same fix in v0.8.10 (commit `2d783e5`). Closing as the issue is already resolved on the `v8` branch. Really appreciate the thorough write-up and test plan.

PR #926, opened 2026-05-18
fix: add .ets (ArkTS) extension to CODE_EXTENSIONS

yyautumn330:
## Summary
- Adds `.ets` (ArkTS) to the `CODE_EXTENSIONS` set in `detect.py`
- ArkTS is the primary language for HarmonyOS/OpenHarmony application development
- `.ets` files are TypeScript-compatible — the tree-sitter TypeScript parser already handles them for AST extraction

## Problem
In OpenHarmony codebases (e.g., SceneBoard), `.ets` files make up ~90% of code. The detect module silently skips them, resulting in severely incomplete knowledge graphs.

**Before** (SceneBoard `product/phonebase` — 476 code files total):
- 0 `.ets` files detected
- 9 `.ts` files detected

**After**:
- 467 `.ets` files detected
- 9 `.ts` files detected

## Test Plan
- [x] `classify_file(Path("test.ets"))` returns `FileType.CODE`
- [x] `detect()` on OpenHarmony project correctly discovers all `.ets` source files
- [x] All other extensions continue to classify correctly

## Note
The tree-sitter TypeScript parser handles `.ets` files for AST extraction — the parser was never the bottleneck, only the detect module's file-type recognition.

### DIFF
--- a/graphify/detect.py
+++ b/graphify/detect.py
@@ -24,7 +24,7 @@ class FileType(str, Enum):
 
 _MANIFEST_PATH = "graphify-out/manifest.json"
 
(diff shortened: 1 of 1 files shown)


safishamsi:
Thanks for this — the detect change is correct and the impact numbers are real. However the PR description's claim that "the tree-sitter TypeScript parser already handles [.ets]" isn't accurate for this codebase.

The `_DISPATCH` table in `extract.py` has no `.ets` entry, so `_get_extractor()` returns `None` for `.ets` files and `_extract_single_file` returns empty nodes/edges. With only the detect change, users get `.ets` files counted in the corpus (good — the LLM semantic pass will see them) but zero structural extraction: no functions, classes, or imports. That's silently worse than the PR description implies.

Two things needed before this can merge:

1. **Add `.ets` to `_DISPATCH` in `extract.py`** — map it to the TypeScript extractor (or TSX if your ArkTS files use declarative `build() { Column() {...} }` UI syntax). The existing `extract_js` already handles `.ts` and `.tsx` via separate configs at lines 2054-2062.

2. **Add a fixture + test** — a `tests/fixtures/sample.ets` with at least one class/struct, one function, and one import, plus a test in `test_multilang.py` or `test_languages.py` asserting those nodes are extracted. There are currently zero `.ets` tests.

Happy to help if you get stuck on which config (`_TS_CONFIG` vs `_TSX_CONFIG`) fits ArkTS syntax best.

PR #927, opened 2026-05-18
fix(extract): resolve TypeScript wildcard path aliases (@*, @*/foo)

ghaiat-yoobic:
v7's tsconfig path-alias resolution drops the substitution wildcard, so paths like `"@*": ["features/*/src/"]` or `"@*/interfaces": ["features/*/src/interfaces.ts"]` never resolve to a real file. Common pattern on Nx/Angular monorepos that use a generic catch-all alongside specific aliases — `_load_tsconfig_aliases` and `_resolve_js_import_target` need to preserve and substitute the `*`.

## What was broken

`_read_tsconfig_aliases` strips trailing `/*` from both the alias key and the target, so:

```json
"@*": ["features/*/src/"]
"@*/interfaces": ["features/*/src/interfaces.ts"]
```

become `{ "@": "<repo>/features", "@*/interfaces": "<repo>/features/*/src/interfaces.ts" }` — the substitution group is lost for `@*`, and the resolver has no concept of wildcards anyway, so the second entry is dead code.

`_resolve_js_import_target` then matches by literal prefix only. `@communicate/documentv2` matches the bare `@` prefix, joins the rest, and produces `<repo>/features/communicate/documentv2`. No file there, falls through to basename `documentv2`. Every import that goes through a wildcard alias becomes a stub.

## Fix

1. **Preserve wildcards.** `_read_tsconfig_aliases` keeps both the alias key and the target verbatim when either contains `*`. Non-wildcard aliases still get the trailing-`/` strip.

2. **Wildcard substitution in the resolver.** `_resolve_js_import_target` splits a wildcard alias on the `*` into `(prefix, suffix)`, matches the import as `raw.startswith(prefix) and raw.endswith(suffix)` with at least one captured char, then substitutes the captured segment into the target's `*`.

3. **Specificity sorting.** Aliases are sorted before matching per the TypeScript path-mapping spec: exact (non-wildcard) entries win over wildcard entries; among wildcards, the longer non-wildcard prefix wins. `@common/integration/*` beats `@*` when both could match.

## Sub-bug fixed too: phantom-node ids on alias resolution

Even when alias resolution found the right file, `_make_id(str(resolved))` was computing a node id from the absolute path (`/Users/.../features/foo/bar.ts`) while the file extractor uses `_make_id(str(path))` against the project-relative path it was given. The two ids never matched, so every alias-resolved import pointed at a phantom node.

New `_resolved_path_to_nid` helper relativizes the resolved Path to `Path.cwd()` before computing the node id, falling back to the absolute form only for paths outside cwd.

## Measured impact

Same Nx/Angular monorepo (7,759 TS files, ~50 tsconfig path entries including a `@*` catch-all and three `@*/foo` patterns):

| | edges (AST) | edges (final graph after dedup) |
|---|---|---|
| v7 baseline | 164,041 | 99,871 |
| with this fix | **193,970** (+30K) | **107,369** (+7.5K) |

Specific failing case from triage on the source repo:

```typescript
// features/cross/agent/src/context/context-providers.ts
import { FileChipComponent, FilesComboListFormComponent } from '@communicate/documentv2';
```

The repo's tsconfig.base.json defines `@communicate/chat/*`, `@communicate/document/*`, `@communicate/feed/*`, plus a catch-all `@*: features/*/src/`. The `documentv2` import has no specific alias and relies on the wildcard.

- Before: importing file had **zero** `imports_from` edges, only internal `contains` edges. The `@communicate/documentv2` import collapsed to a bare `documentv2` basename node disconnected from the real file.
- After: 1 `imports_from` edge to `features_communicate_documentv2_src_index_ts` at L5, plus a symbol-level `imports` edge to `FileChipComponent`. 29 total edges now target the documentv2 index node.

Bare/scoped imports (npm packages, unmatched aliases) keep the existing basename fallback unchanged.

## Test plan

I didn't add tests in this PR — the existing `tests/` directory has no coverage for `_read_tsconfig_aliases` or `_resolve_js_import_target` that I could extend. Happy to add a fixture + tests as a follow-up if you'd like, or attach them to this PR before merge.

The three behaviours that need tests:
- Wildcard alias substitution: `@*` → `features/*/src/` resolves `@foo/bar` to `features/foo/bar/src/`
- Specificity ordering: `@common/integration/*` beats `@*` for `@common/integration/foo`
- Non-wildcard aliases keep working unchanged

The fix is also straightforward to verify against the spec at https://www.typescriptlang.org/docs/handbook/module-resolution.html#path-mapping.

### DIFF
--- a/graphify/extract.py
+++ b/graphify/extract.py
@@ -126,9 +126,15 @@ def _read_tsconfig_aliases(tsconfig: Path, base_dir: Path, seen: set) -> dict[st
     for alias, targets in paths.items():
         if not targets:
             continue
-        alias_prefix = alias.rstrip("/*")
-        target_base = targets[0].rstrip("/*")
-        aliases[alias_prefix] = str(base_dir / target_base)
+        target = targets[0]
+        # Preserve the wildcard in both alias and target so the resolver can
+        # substitute the captured segment back in.  TypeScript's path mapping
+        # spec uses '*' as a single capture group; keeping it lets us tell
+        # 'wildcard alias' from 'exact alias' at resolve time.
+        if "*" in alias or "*" in target:
+            aliases[alias] = str(base_dir / target) if target else str(base_dir)
+        else:
+            aliases[alias.rstrip("/")] = str(base_dir / target.rstrip("/"))
 
     return aliases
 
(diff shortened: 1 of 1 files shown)


safishamsi:
Strong work on the wildcard alias resolution — the measured impact (+30K AST edges on a real Nx monorepo) confirms this is a real problem. A few asks before merge:\n1. **Rebase onto v8** — this is based on `v7` and v8 has substantial `extract.py` changes including the MCP extractor\n2. **Split out the barrel re-export work** — it's a significant undocumented feature (barrel resolution, `_reexport_target_path`, 16-deep cycle guard, named/star/namespace maps) that deserves its own PR and review surface\n3. **Tests** — the three behaviors you called out (wildcard substitution, specificity sort, non-wildcard regression) need tests; the barrel path needs even more coverage (named, star, cycle, max-depth)\n\nThe wildcard alias fix + `_resolved_path_to_nid` helper on their own would be very mergeable. Happy to review a scoped version.

oleksii-tumanov:
Would you mind if I prepare a scoped `v8` follow-up with the wildcard alias fix and requested tests, leaving barrel re-exports out? I’ll reference #927 as the original work.

safishamsi:
Feel free

oleksii-tumanov:
Following up on the requested split, I reworked the original two areas for the current `v8` pipeline:

- #1544 covers wildcard path aliases, with substitution, specificity, and regression tests.
- #1552 covers the remaining namespace re-export binding, with cycle and long-chain coverage around the existing `v8` barrel resolver.

Both reference #927 as the original work. If both scoped PRs are accepted, #927 can be closed as superseded.

oleksii-tumanov:
this PR can be closed. Changes already merged.

PR #930, opened 2026-05-18
fix(watch): honor .graphifyignore patterns

### LINKED ISSUE #928: watch.py bypasses .graphifyignore; recursive watch on busy volumes causes CPU saturation from FSEvents firehose
### Summary

`graphify watch <path>` schedules a recursive `watchdog.observers.Observer()` watcher and then filters events in the Python handler using **only** three hardcoded rules:

1. `path.suffix.lower() in _WATCHED_EXTENSIONS` (code/doc/paper/image)
2. `not any(part.startswith(".") for part in path.parts)`
3. `"graphify-out" not in path.parts`

The handler does **not** consult `.graphifyignore`, even though the project has a working `_load_graphifyignore()` + `_is_ignored()` pair in `detect.py` that is used by `update`, `extract`, and the post-commit hook path.

This means: when `graphify watch` is pointed at a directory that lives on a high-FS-event-volume root (Time Machine snapshot volume, Spotlight-indexed volume, Docker/Colima-shared mount, sync staging directory), the Python event handler is woken for every file event in the tree before discarding it. On a quiescent project tree this is fine. On a busy shared mount, this saturates a CPU core.

I have hit this pattern twice on the same host roughly a month apart:
- First occurrence: `graphify watch` over a project tree that shared an APFS volume with a Colima virtio mount; sustained 1-min load avg in the low 20s while the watcher was active.
- Second occurrence: several `graphify watch` instances over different subdirectories of the same APFS volume; sustained load avg 25+ for several days until the watchers were disabled.

The workaround is "don't run graphify watch on shared/busy volumes at all," which is unsatisfying because the watcher would be useful on the *project subdirectories* — it's the recursive enumeration of every event in a shared root that's the problem.

### Why filing this as a bug rather than an enhancement

Two reasons:
- `.graphifyignore` is a documented and supported mechanism (see issues #861, #882, #870, #869, all of which treat it as the canonical filter). Users reasonably expect it to be honored everywhere the tool walks the filesystem, including the watcher.
- The fix is small (~10 lines in `watch.py`) and uses code that already ships in the package.

Happy to relabel as an enhancement if maintainer prefers.

### Affected version

- pipx package `graphifyy` 0.4.1
- CLI reports `graphify 0.7.15`
- Python 3.14.3

### Environment

- macOS, Apple Silicon
- File systems involved: APFS with Time Machine local snapshots enabled, Spotlight-indexed; Colima virtio-shared mount over the same volume root.
- The pattern should reproduce on any `recursive=True` watch over a tree that has a steady background rate of FS events from external sources.

### Reproduction

1. On macOS, install `graphifyy` via pipx.
2. Create a `.graphifyignore` in a project root with broad patterns:
   ```
   *
   !src/**
   ```
   (i.e. ignore everything except `src/`)
3. Run `graphify watch <project_root>`.
4. In a separate terminal, generate FS events outside `src/` at a steady rate. Easiest reproducer: run a TM local snapshot (`tmutil localsnapshot`), or `touch` 1000 files in a subdir not under `src/`, or rsync into a sibling directory.
5. Observe `python3` CPU consumption in `top`. It will spike well past what `.graphifyignore` would suggest is necessary.

Expected: handler skips events under ignored paths, CPU near zero.
Actual: handler wakes for every event, only filters at the suffix / `.`-prefix / `graphify-out` stage.

### Code citation

`src/graphify/watch.py` (current `main` as of v0.7.15):

```python
class Handler(FileSystemEventHandler):
    def on_any_event(self, event):
        nonlocal last_trigger, pending
        if event.is_directory:
            return
        path = Path(event.src_path)
        if path.suffix.lower() not in _WATCHED_EXTENSIONS:
            return
        if any(part.startswith(".") for part in path.parts):
            return
        if "graphify-out" in path.parts:
            return
        last_trigger = time.monotonic()
        pending = True
        changed.add(path)
```

vs. `detect.py:288` (the existing infrastructure that's not being used here):

```python
def _is_ignored(path: Path, root: Path, patterns: list[str]) -> bool:
    """Return True if path matches any .graphifyignore pattern."""
    ...
```

### Suggested fix

Two-line change in `watch.py`, leveraging existing `detect` helpers. Load patterns once at `watch()` startup; check in the event handler. Sketch:

```python
def watch(watch_path: Path, debounce: float = 3.0) -> None:
    from graphify.detect import _load_graphifyignore, _is_ignored
    ignore_patterns = _load_graphifyignore(watch_path)
    ...
    class Handler(FileSystemEventHandler):
        def on_any_event(self, event):
            ...
            if ignore_patterns and _is_ignored(path, watch_path, ignore_patterns):
                return
            ...
```

(Caveat: this loads `.graphifyignore` once at startup, so adding patterns mid-session requires a watcher restart. That seems acceptable; if not, a stat-based reload check is a one-liner.)

### Adjacent enhancements that would also help

If you're already in this code path, two small additions would unblock most "watcher unsafe on this volume" scenarios:

1. **Expose `watchdog.observers.polling.PollingObserver` via a `--poll` flag.** The default `Observer()` resolves to FSEvents on macOS, which delivers events for the whole tree regardless of where the consumer is interested. `PollingObserver(timeout=<interval>)` walks the tree on a schedule instead, trading I/O for predictable cost. Useful escape hatch for volumes where FSEvents delivery is itself the cost driver. ~3 lines.
2. **Allow multiple `watch_path` arguments** (e.g. `graphify watch src/ tests/ docs/`) so users can avoid `recursive=True` on a root containing busy siblings. Currently the only way to scope is run multiple graphify processes. ~5 lines.

Either of those would also resolve the use case, but the `.graphifyignore` fix is the most principled because it uses an already-supported mechanism.

### Workaround in use

Disabled the affected `graphify watch` instances. Still using `graphify update <path>` on a daily schedule — that gives ~90% of the watcher's value at zero background cost. Probably worth a note in README under "macOS / shared volumes" until a release lands.

### Willingness to PR

Happy to send a PR for the `.graphifyignore`-honoring change if useful (~10 lines + test). Less sure about the polling/multi-path additions since those are more API-surface decisions.

YuanHanzhong:
Summary
- Load .graphifyignore patterns when watch mode starts.
- Reuse the existing ignore matcher before scheduling rebuilds for file events.
- Add watch-path filtering tests for ignored and allowed files.

Testing
- uv run --with pytest pytest tests/test_watch.py -q
- git diff --check

Fixes #928


safishamsi:
Thanks for the thorough PR — the intent is exactly right and your tests are well-written. The fix was actually already shipped in **v0.8.12** as part of #928.

Current `watch.py` (lines 659-679) already does what this PR proposes:
- `.graphifyignore` patterns loaded once at startup via `_load_graphifyignore(watch_root_for_ignore)`
- `_is_ignored()` checked first, before the extension/dotfile filters, so broad patterns like `node_modules/` short-circuit cheaply

Your PR branch is also significantly behind `v8` (120k line diff from accumulated upstream commits), which made it un-mergeable directly.

The fix is live — upgrade with `pip install --upgrade graphifyy` or `uv tool upgrade graphifyy`. Thanks again for the investigation.

Issue #811, opened 2026-05-11
Chinese/Unicode identifiers cause silent node deduplication — only 8 of 202 functions captured


quinnwen:
### Problem

The `_make_id()` function in `extract.py` (line 35) uses `re.sub(r"[^a-zA-Z0-9]+", "_", combined)` which strips all non-ASCII characters from node IDs. When processing files with Chinese (or other Unicode) function names, all non-ASCII characters are removed, leaving empty strings or just digits as identifiers. This causes massive silent deduplication — the graph ends up missing the vast majority of functions.

### Impact

On a Lua file with Chinese function names (17,000 lines, 828KB):

| Function Name | Generated ID (broken) |
|---|---|
| `道具处理类:初始化` | `""` (empty string) |
| `道具处理类:附灵洗练1` | `"1"` |
| `道具处理类:数据处理` | `""` (duplicate of above!) |
| `道具处理类:附灵洗练2` | `"2"` |

- Only **8 of 202** functions were captured in the knowledge graph.
- **194 functions were silently deduplicated** due to ID collisions.
- Tree-sitter correctly parses ALL functions — the bug is purely in graphify's ID generation layer.

### Root Cause

Two locations use the same ASCII-only regex pattern:

1. **`graphify/extract.py` line 35** — `_make_id()` uses `re.sub(r"[^a-zA-Z0-9]+", "_", combined)`
2. **`graphify/build.py` line 38** — `_normalize_id()` uses the same `[^a-zA-Z0-9]+` pattern

Both strip every non-ASCII character, causing Chinese, Japanese, Korean, Arabic, and other Unicode identifiers to collapse into empty strings or bare digits.

### Suggested Fix

Change the regex from `[^a-zA-Z0-9]+` to `[^\w]+` with `flags=re.UNICODE`:

```python
# extract.py _make_id():
cleaned = re.sub(r"[^\w]+", "_", combined, flags=re.UNICODE)

# build.py _normalize_id():
cleaned = re.sub(r"[^\w]+", "_", s, flags=re.UNICODE)
```

This preserves Unicode word characters (CJK ideographs, accented Latin, etc.) while still replacing punctuation and whitespace with underscores.

### Before/After Comparison

| Function Name | Before (broken) | After (fixed) |
|---|---|---|
| `道具处理类:初始化` | `""` | `"道具处理类_初始化"` |
| `道具处理类:附灵洗练1` | `"1"` | `"道具处理类_附灵洗练1"` |
[...]

PR #937, opened 2026-05-19
fix: CJK/Unicode labels silently skipped in _norm/_norm_label dedup (follow-up to #811)

szgnewGh:
## Summary

Follow-up to **#811**: the original fix correctly changed `_make_id()` and `_normalize_id()` to use `[^\w]+` with `re.UNICODE`, but **two more functions** still use ASCII-only regex patterns, causing CJK/Unicode labels to be silently stripped:

### Affected functions

| Function | Before (broken) | After (fixed) |
|----------|-----------------|---------------|
| `dedup.py` `_norm()` | `[^a-z0-9]+` | `[^\w]+` + `re.UNICODE` + `casefold()` |
| `build.py` `_norm_label()` | `[^a-z0-9 ]` | `[^\w ]` + `re.UNICODE` + `casefold()` |

### Impact

When processing repos with Chinese/Japanese/Korean identifiers, these two functions strip all Unicode characters from labels before dedup comparison. This causes:
- **False dedup merges**: different Chinese labels like `道具处理类_初始化` and `道具处理类_数据处理` both normalize to empty strings, getting merged into a single node
- **Silent data loss**: CJK nodes that should be distinct entities are collapsed into one

### What changed

- `dedup.py:17` — `lower()` → `casefold()`, `[^a-z0-9]+` → `[^\w]+` with `flag=re.UNICODE`
- `build.py:230` — `lower()` → `casefold()`, `[^a-z0-9 ]` → `[^\w ]` with `flag=re.UNICODE`

Using `casefold()` instead of `lower()` for better Unicode case folding (e.g. German ß → ss).

### References

- Original issue: #811
- The upstream fixed `extract.py:_make_id()` and `build.py:_normalize_id()` but missed these two dedup functions.

### DIFF
--- a/graphify/build.py
+++ b/graphify/build.py
@@ -228,8 +228,8 @@ def build(
 
 
 def _norm_label(label: str) -> str:
-    """Canonical dedup key — lowercase, alphanumeric only."""
-    return re.sub(r"[^a-z0-9 ]", "", label.lower()).strip()
+    """Canonical dedup key — Unicode-aware, preserves CJK/word characters."""
+    return re.sub(r"[^\w ]", "", label.casefold(), flags=re.UNICODE).strip()
 
 
 def deduplicate_by_label(nodes: list[dict], edges: list[dict]) -> tuple[list[dict], list[dict]]:
--- a/graphify/build.py
+++ b/graphify/build.py
@@ -229,7 +229,8 @@ def build(
 
 def _norm_label(label: str) -> str:
     """Canonical dedup key — Unicode-aware, preserves CJK/word characters."""
-    return re.sub(r"[^\w ]", "", label.casefold(), flags=re.UNICODE).strip()
+    label = unicodedata.normalize("NFKC", label)
+    return re.sub(r"[\W_ ]+", " ", label.casefold(), flags=re.UNICODE).strip()
 
 
 def deduplicate_by_label(nodes: list[dict], edges: list[dict]) -> tuple[list[dict], list[dict]]:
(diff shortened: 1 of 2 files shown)


safishamsi:
Thanks for the careful follow-up to #811 — the CJK false-dedup is real and the direction is right. Two issues need fixing before this can merge:

**1. Underscore regression (breaks cross-extractor dedup)**

`\w` includes `_`. The old pattern stripped underscores from dedup keys:

```python
# Old: "my_function" → "my function"
re.sub(r"[^a-z0-9 ]", "", "my_function".lower()) == "my function"

# New: "my_function" → "my_function"
re.sub(r"[^\w ]", "", "my_function".casefold(), flags=re.UNICODE) == "my_function"
```

This breaks merges between LLM-generated labels (`"my function"`) and AST labels (`"my_function"`) — exactly the cross-extractor reconciliation these functions exist to do. Fix: use `[\W_]+` to drop underscores while gaining Unicode:

```python
# _norm (dedup.py) — collapse runs to space
return re.sub(r"[\W_]+", " ", label.casefold(), flags=re.UNICODE).strip()

# _norm_label (build.py) — strip entirely
return re.sub(r"[\W_ ]+", " ", label.casefold(), flags=re.UNICODE).strip()
```

**2. Missing NFKC normalization (parity with #811)**

`_make_id()` and `_normalize_id()` (the functions fixed in #811) both apply `unicodedata.normalize("NFKC", ...)` before the regex. This PR omits it, so `"café"` (NFC) and `"café"` (NFD) still produce different dedup keys — CJK labels from two different extractors won't merge if their Unicode form differs.

Add this to both functions:

```python
import unicodedata
label = unicodedata.normalize("NFKC", label)
```

Fix both and this is ready to merge.

PR #938, opened 2026-05-19
Added LM Studio as a supported backend

telno:
Added LM Studio backend parameter and definition to BACKENDS


telno:
Closed due to updated pull available

Issue #919, opened 2026-05-17
Feature request: CLI flags to handle super-hub-driven low-cohesion communities


neo1777:
Thanks for graphify — I've been using it on a React/Firebase PWA codebase and it's been really useful for surfacing structural patterns I wouldn't have caught manually.

Related: #830 — god-node transit in BFS/DFS. The two issues address the same underlying phenomenon (high-degree hubs degrading downstream analysis), applied here to clustering rather than path-finding.

## Context

Running graphify on a ~1500-node React/TypeScript graph, the output flagged Community 0 (86 nodes) with cohesion 0.05. Manual inspection showed the community is not a semantic neighborhood: it bundles six distinct sub-families (UI primitives, page roots, modal family, layout/nav, PWA bootstrap, RBAC) that have been aggregated together by the clustering algorithm rather than by topology that reflects the actual codebase structure.

The aggregation traces to two "super-hub" nodes:

| Node | Degree | Breakdown |
|---|---|---|
| `src/components/ui/index.tsx` | 44 | 14 external imports + 30 contained symbols (barrel re-export module) |
| `useRBAC()` hook | 32 | 13 imports + 18 call sites across consumer modules |

Both hubs are adjacent to nodes from every sub-family at once, which forces Leiden to attract them all into the same basin under uniform edge weights.

## Why the existing second-pass mitigation doesn't catch this

I noticed the comment at `cluster.py:124-125` showing that you've already addressed a related class of cluster bloating (doc-hub bridging, e.g. `CLAUDE.md` connected to everything) with the cohesion-based second-pass split. Two reasons it doesn't cover the case described here:

1. **The cohesion threshold doesn't trigger.** The actual cohesion of Community 0 is `178 / 3655 = 0.0487`. `cohesion_score()` at `cluster.py:166` returns `round(actual / possible, 2)`, producing `0.05`. The threshold check at `cluster.py:128` (`cohesion_score(G, nodes) < _COHESION_SPLIT_THRESHOLD`) becomes `0.05 < 0.05 → False`, so the community falls through unchanged. (See *Additional context* below — this looks like a small latent bug worth a separate one-line fix.)

[...]

PR #942, opened 2026-05-19
fix(cluster-only): ensure output directory exists before writing report (#934)

### LINKED ISSUE #934: cluster-only crashes with FileNotFoundError if graphify-out/ doesn't exist before invocation
Surfaced while testing v0.8.11 on #919. One-line fix; opening as a standalone issue rather than commenting on the closed #919 thread.

## Reproducer

```
$ rm -rf graphify-out
$ graphify cluster-only . --graph some/other/path/graph.json --exclude-hubs 99
Loading existing graph...
Graph: 1501 nodes, 2200 edges
Re-clustering...
  ...
  File ".../graphify/__main__.py", line 1827, in main
    (out / "GRAPH_REPORT.md").write_text(report, encoding="utf-8")
  ...
FileNotFoundError: [Errno 2] No such file or directory: 'graphify-out/GRAPH_REPORT.md'
```

## Root cause

On `v8` HEAD (`__main__.py` SHA `895f626`, line 1827):

```python
(out / "GRAPH_REPORT.md").write_text(report, encoding="utf-8")
```

`out` is dereferenced without first being ensured to exist. The pattern `out.mkdir(parents=True, exist_ok=True)` is used correctly elsewhere in the same file at 20 different sites — including `__main__.py:2935` where the very same identifier `out` is created before being written to. So this looks like an asymmetry between the `extract` codepath (creates `out` first) and the `cluster-only` codepath (assumes `out` already exists), rather than a design choice.

The crash is reachable any time a user renames or deletes `graphify-out/` before re-running `cluster-only` — a natural workflow for archiving a previous run before re-clustering with new flags (the scenario that surfaced it). Related precedent: #422 (`cluster-only` skipping detect-phase setup) showed the same general pattern.

## Suggested fix

One line, right before the first write into `out`:

```python
out.mkdir(parents=True, exist_ok=True)
(out / "GRAPH_REPORT.md").write_text(report, encoding="utf-8")
```

## Workaround in the meantime

```
mkdir -p graphify-out && graphify cluster-only ...
```

## Environment

- `graphifyy` 0.8.11 (installed via pipx); line still present on `v8` HEAD (v0.8.13, same SHA `895f626`)
- Python 3.12, Linux Mint
- Reproduced on the same React/Firebase PWA codebase as #919

Happy to send a PR if useful.

neo1777:
Fixes #934.

## What

Adds `out.mkdir(parents=True, exist_ok=True)` before the first write into `out` in the `cluster-only` codepath (`graphify/__main__.py` line 1827 on `v8` HEAD).

## Why

`cluster-only` crashes with `FileNotFoundError` when `graphify-out/` doesn't exist at invocation — a natural state when archiving a previous run before re-clustering with new flags. The `extract` codepath creates `out` explicitly (line 2935); `cluster-only` assumed it existed, which is the asymmetry causing the bug.

The same pattern (`mkdir(parents=True, exist_ok=True)`) is already used at ~20 sites in the same file, so this fix aligns `cluster-only` with the rest of the codebase.

## Manual verification

Reproduced the original crash and confirmed the fix on the same React/Firebase PWA codebase that surfaced #919 / #934 (full reproducer documented in #934).

Also ran the existing test suite locally with the fix applied:
- 945 tests passed
- The 8 failing SQL tests in `test_multilang.py` are pre-existing in my local environment (missing optional `tree-sitter-sql` dependency) and fail identically with the fix reverted, so unrelated to this change.

## Tests

I held off adding a regression test specific to `cluster-only` because the codepath is currently invoked through the manual argparse-style parser in `main()` rather than a directly callable function, and I wasn't sure of the preferred pattern. Happy to add an end-to-end subprocess test, or to extract the codepath into a callable first if you prefer — let me know what would fit best.

---

First-time contributor — happy to iterate on anything.

### DIFF
--- a/graphify/__main__.py
+++ b/graphify/__main__.py
@@ -1824,6 +1824,7 @@ def main() -> None:
                           {"warning": "cluster-only mode — file stats not available"},
                           tokens, str(watch_path), suggested_questions=questions,
                           min_community_size=min_community_size, built_at_commit=_commit)
+        out.mkdir(parents=True, exist_ok=True)
         (out / "GRAPH_REPORT.md").write_text(report, encoding="utf-8")
         from graphify.export import backup_if_protected as _backup
         _backup(out)


safishamsi:
Great catch and a correct fix — thank you for the thorough write-up and for reproducing it.

We implemented this directly with two small improvements over the PR:

1. **Hoisted the `mkdir` earlier** — moved it to right after `out = watch_path / "graphify-out"` (line 1810) so it runs before the `labels_path.exists()` read and any subsequent write, not just before the first write. This also ensures `backup_if_protected` always has a dir to operate on.

2. **Added the regression test** you offered to write — an end-to-end subprocess test that archives `graphify-out/`, confirms the dir is gone, runs `cluster-only --graph`, and asserts `GRAPH_REPORT.md` lands correctly.

The fix will be in the next release. First-time contributor badge well earned — come back anytime.

neo1777:
Thanks for the quick turnaround, and for laying out the changes so clearly. Hoisting the mkdir right after the output path is computed is the better call — it covers backup_if_protected and the labels read, not just the first write.
Glad the reproducer and the write-up were useful. I'd like to look into the third option from #919 — the edge-weight-mode idea (inverse-degree edge weighting) — since exclude-hubs and edge-weighting tackle the super-hub problem from different angles. I'll dig into it properly and open it as a real PR with tests if it holds up. Thanks again.

Issue #952, opened 2026-05-20
AST extractor produces ID collisions on identical filenames across folders, merging distinct functions into one node


yuritoledo:
## Bug

`graphify.extract` (AST mode) generates node IDs from `{filename_stem}_{symbol}`. In monorepos where many folders contain a same-named file (e.g. Next.js App Router with one `action-bar/index.tsx` per feature, or one `lib/atom.ts` per feature), every function with the same name across those files collapses into a single graph node. The union of their edges then creates phantom cross-module "bridges" that don't exist in source.

## Repro

In a Next.js repo with structure:

    app/(innerPages)/<feature>/action-bar/index.tsx   # each exports `function ActionBar()`
    app/(innerPages)/<feature>/lib/atom.ts            # each exports `useFeatureAtom`

Run `graphify .` on a code-only corpus.

Result: a single node `action_bar_index_actionbar` accumulates `contains` edges from ~16 distinct files and `calls` edges to atoms from ~10 unrelated features. `god_nodes()` reports it as the most connected abstraction. `surprising_connections()` flags fake cross-feature bridges.

## Expected

IDs should be unique per source file, e.g. derived from the full relative path (`innerpages_<feature>_action_bar_index_actionbar`) or filepath hash, so functions with identical names in different folders remain distinct nodes.

## Evidence

In one 552-file scan, 16 `action-bar/index.tsx` files merged into 1 node showing 30 edges and 0.167 betweenness — entirely an artifact.

## Suggested fix

Change ID generation in the AST extractor to use the relative source path (slashes → underscores) instead of just `Path(...).stem`. This is also what the Step 3 prompt instructs semantic subagents to do, so the AST extractor would just be matching its sibling.

## Version

graphifyy from PyPI, Python 3.x via pipx.

AbarnaaSree:
Hi! I'm interested in working on this issue. I'll first reproduce the bug, analyze the ID generation logic in the AST extractor, and submit a fix with regression tests. Please let me know if there's any preferred approach before I start.

Issue #949, opened 2026-05-20
graphify update --no-cluster picks wrong import target on macOS for fully-qualified imports when two modules share a basename


augustin-emburse:
### Observation

On a Python project with two `<name>.py` modules under different packages and a fully-qualified import of one of them, `graphify update . --no-cluster` produces different `graph.json` output on macOS vs Linux at the **same git SHA**, same graphifyy `0.8.13`, same Python 3.12.

Minimal layout (full repro below):

```
repro/
  pkg_a/__init__.py
  pkg_a/settings.py        # class Settings
  pkg_b/__init__.py
  pkg_b/settings.py        # class Settings
  main.py                  # from pkg_a.settings import Settings
```

The import `from pkg_a.settings import Settings` is unambiguous: it refers to `pkg_a/settings.py`. After v0.8.13's directory-qualified stem fix, both files become distinct nodes (`pkg_a_settings_settings` and `pkg_b_settings_settings`) — no node-ID collision.

When the same SHA is rebuilt on a Linux GitHub Actions runner and diffed against a macOS-built `graph.json` committed by a developer, edges originating in `pkg_a/*` flip their `target` between the two same-basename nodes:

```diff
 {
   "source": "main_f",
-  "target": "pkg_b_settings_settings",
+  "target": "pkg_a_settings_settings",
   "relation": "uses",
   "source_file": "main.py",
   ...
 },
```

The `+` (Linux) target is correct. The `−` (macOS) target is wrong — `pkg_a.settings.Settings` and `pkg_b.settings.Settings` are different symbols.

The pattern repeats for every edge that touches one of these same-basename modules: the resolver lands on whichever `<basename>.py` came first in the filesystem walk, ignoring the import's package prefix.

### Why this is unexpected after v0.8.13

The "Node ID collisions across same-named files in different directories" fix in 0.8.13 makes both files distinct nodes, but the import-target lookup appears to still match candidates by bare `<basename>` and pick the first hit. The package prefix on the import (`pkg_a.` vs `pkg_b.`) is not used to disambiguate.

Hypothesis: filesystem walk order from `os.scandir()` differs between APFS (macOS) and ext4 (Linux). Either order is fine in isolation, but it shouldn't affect the picked target for a fully-qualified import.

### Practical impact

[...]

Issue #572, opened 2026-04-27
Feature request: make graph queries context-aware across languages


TheFedaikin:
## What we want

  We want graphify queries to distinguish between edge kinds that are currently traversed together.

  Examples:
  - calls vs type/declaration references
  - imports vs executable relationships
  - field references vs other symbol mentions

  A query like `who calls X?` should prefer call-heavy results instead of mixing calls, imports, and declaration-style references.

  ## Why

  This improves query precision in larger or noisier graphs and makes edge disambiguation a graph-level, cross-language capability.

  Model boundary:
  - `relation` = semantic edge type
  - `context` = syntactic disambiguation where needed

  ## Proposal

  Add a cross-language `context` layer for edges where it improves query quality, and use it during query traversal.

  Initial contexts:
  - `call`
  - `import`
  - `field`

  Queries should support:
  - explicit filtering
  - small heuristic inference when the query clearly implies a context

  Examples:
  - `graphify query "extract" --context call`
  - `who calls extract` -> infer `call` when no explicit filter is provided

  ## Scope

  - add/standardize edge `context` where supported
  - support context-aware filtering in:
    - CLI `graphify query`
    - MCP `query_graph`
  - show active context filtering in query output

  ## Non-goals

  - do not move `inherits` / `implements` into `context`
  - do not replace `relation` with `context`
  - do not build broad NLP intent parsing
  - do not treat this as a fix for general `explain` / `path` disambiguation

  ## References #569

  This is related to #569, but narrower in scope.

  - `#569` is mainly about resolving the right node or candidate set in large, noisy graphs
  - this issue is about traversing the right edge kinds once the node is found

  In short:
  - `#569` = node resolution / disambiguation
  - this issue = edge-context-aware traversal

  ## Done when

  - queries can explicitly filter by context
  - obvious query phrasing can infer context when no filter is passed
  - output makes the active context visible
  - extraction/tests cover the touched languages

Issue #462, opened 2026-04-19
Support pnpm monorepos: resolve workspace packages to real files and cross-package type usage


sebazai:
In a **pnpm workspace** (and similar JS monorepos), shared code lives in packages such as `packages/types`, while apps live under `apps/*`. Imports use the **published package name** from `package.json` (for example `@workspace/types`), not relative file paths.

Graphify’s JS/TS import handling today effectively treats many non-relative specifiers like opaque external modules. In practice:

- **Scoped packages** are easy to mis-handle (for example `@workspace/types` must not collapse to a meaningless or colliding token such as `types`).
- **`imports_from` edges often do not attach to real file nodes** in the repository, so they can be dropped when building the graph and **structure between apps and shared packages is lost**.
- There is **no strong story for “this type from the shared package is used in the frontend (and backend)”** at graph level, which matters for architecture navigation and impact analysis.

## Example layout

`pnpm-workspace.yaml`:

```yaml
packages:
  - 'apps/*'
  - 'packages/*'
```

- `apps/backend` and `apps/frontend` depend on `packages/types`.
- `packages/types` is exposed as e.g. `@workspace/types` in its `package.json`.
- Code imports: `import type { SomeDto } from '@workspace/types'`.

## Why this matters

- The graph should show **dependencies between workspaces** (for example app → shared types package), not only a flat set of “external” import names.
- Ideally, **symbols** (types/interfaces) defined in shared packages can be related to **consumers** in other packages (similar in spirit to cross-file inferred edges for Python).

## Desired behavior (high level)

1. **Detect** a monorepo when `pnpm-workspace.yaml` (and optionally other workspace manifests) is present; discover package roots from workspace globs and each package’s `package.json` `name`.
2. **Resolve** `import … from '@scope/pkg'` (and reasonable subpaths) to **concrete entry files** under the repo so edges target **existing file nodes** (and optionally **package-level hub** nodes for navigation).
3. **Cache / rebuild correctness** when workspace layout or package names change.
4. **(Stretch / Phase 2)** Infer edges from named workspace imports to **exported types/interfaces** so questions like “where is this shared type used in the frontend?” are answerable from the graph.

cbartens:
## Data point from a 22-workspace pnpm monorepo, plus a workaround and one sharp edge

[...]

PR #953, opened 2026-05-20
Add v8 affected-symbol resolution for imports, aliases, and re-exports

TheFedaikin:
## Summary

  Adds graph-level affected-symbol resolution for imports, aliases, and re-exports on top of `v8`.

  This lets graphify answer “what is affected if this symbol changes?” across common indirection patterns instead of only direct declarations/usages.

  ## Why

  Large JS/TS and Python projects often route symbols through barrels, aliases, and re-export layers. Without resolving those relationships, affected analysis misses real downstream impact.

  This improves practical query precision for monorepos and framework-heavy codebases where the changed symbol is rarely consumed directly from its defining file.

  ## What changed

  - Added `graphify affected`.
  - Added generic affected-symbol support for import, alias, and re-export relationships.
  - Added JS/TS coverage for:
    - `export *`
    - `export { X as Y }`
    - inline import aliases
    - type/interface/abstract class/value exports
    - const aliases
    - imported symbol usage affecting local callers
  - Added Python import-resolution coverage.
  - Kept SQL tests optional when `tree_sitter_sql` is not installed.
 
 ## Related

  Related to https://github.com/safishamsi/graphify/issues/569 and the broader
  source-resolution/import-noise issues around large workspaces and ambiguous traversal.

  - #949: import target ambiguity when multiple modules share the same basename.
  - #952: AST node identity collisions for identical filenames across folders.
  - #462: resolving workspace/package imports in JS/TS monorepos.
  - #147: TypeScript path alias resolution.
  - #572: context-aware graph queries; this PR improves the source/import relationship data
  that query filtering can build on.

  ## Test Plan

  ```bash
  mise run pytest -- tests/ -q --tb=short

  Result:

  980 passed, 11 skipped in 9.12s

  Also refreshed the graph:

  uv run --python 3.12 graphify update .

  Result:

  No code-graph topology changes detected; outputs left untouched.

### DIFF
--- a/graphify/__main__.py
+++ b/graphify/__main__.py
@@ -1241,6 +1241,10 @@ def main() -> None:
         print("    --context C             explicit edge-context filter (repeatable)")
         print("    --budget N              cap output at N tokens (default 2000)")
         print("    --graph <path>          path to graph.json (default graphify-out/graph.json)")
+        print("  affected \"X\"             reverse traversal to find nodes impacted by X")
+        print("    --relation R            edge relation to traverse in reverse (repeatable)")
+        print("    --depth N               reverse traversal depth (default 2)")
+        print("    --graph <path>          path to graph.json (default graphify-out/graph.json)")
         print("  save-result             save a Q&A result to graphify-out/memory/ for graph feedback loop")
         print("    --question Q            the question asked")
         print("    --answer A              the answer to save")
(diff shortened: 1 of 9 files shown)


safishamsi:
This was cherry-picked onto v8 and shipped in 0.8.15. Closing — affected-symbol resolution for imports, aliases, and re-exports is live.

PR #960, opened 2026-05-21
feat: track JS/TS barrel re-exports as explicit graph edges

deXterbed:
## Summary

Adds explicit JS/TS barrel re-export edges so graphify can resolve symbol flow through barrel files on top of v8.

This makes relationships like consumer -> barrel export -> source export visible in the graph instead of stopping at the barrel file boundary.

## Why

Large TS/JS codebases commonly use barrel files with patterns like export { X } from './mod'.  
Without explicit re-export edges, graph analysis under-reports real dependencies and weakens refactor suggestions (dead/single-use exports, affected traversal, dependency impact).

## What changed

- Updated JS/TS extraction to detect export ... from re-export statements.
- Added re_exports relation emission for barrel symbol forwarding.
- Kept non-re-export export statements traversable so local declarations are still captured.
- Included re_exports in edge-cleaning allowlist so valid edges are preserved.
- Added tests and fixtures for:
  - named re-export edge creation
  - imports_from continuity to source module
  - re-export context tagging
  - confidence tagging
  - preservation of local exports in barrel files
  - no false re_exports for export { X } without from

## Related

- Related to the broader source-resolution/refactor-accuracy work for v8.
- Complements affected-symbol and import-resolution improvements discussed in #953.
- Supports follow-up refactor reporting work (dead/single-use export analysis).

## Test Plan

Run targeted extraction tests:

uv run --with pytest pytest tests/test_extract.py -k "barrel or reexport"

Result:

6 passed, 53 deselected

Optional full extract test file run:

uv run --with pytest pytest tests/test_extract.py -q --tb=short

### DIFF
--- a/graphify/extract.py
+++ b/graphify/extract.py
@@ -369,6 +369,17 @@ def _resolve_js_import_target(raw: str, str_path: str) -> "tuple[str, Path | Non
 
 
 def _import_js(node, source: bytes, file_nid: str, stem: str, edges: list, str_path: str) -> None:
+    is_reexport = node.type == "export_statement"
+    # Only handle export_statement if it has a `from` clause (re-export).
+    # Pure exports like `export const x = 1` or `export { localVar }` have no source module.
+    if is_reexport:
+        has_from = any(child.type == "from" or (_read_text(child, source) == "from") for child in node.children if child.type in ("from", "identifier"))
+        if not has_from:
+            # Check for string child (source path) as a more reliable indicator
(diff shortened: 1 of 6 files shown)


TheFedaikin:
I think it's covered in #953, unless it's not going to be merged. I don't think you should close it, especially if it gets merged first - it's just for awareness.

safishamsi:
This was cherry-picked onto v8 and shipped in 0.8.15. Closing — barrel re-export tracking is live.

PR #994, opened 2026-05-23
fix: ignore search punctuation in graph queries

### LINKED ISSUE #978: graphify query: punctuation in search terms breaks node matching
## Bug Description

graphify query, graphify path, and graphify explain all fail to match nodes when the search term contains punctuation (e.g. ?, !, ., ,, :).

## Root Cause

In serve.py line 309, _query_graph_text tokenizes the question by simply splitting on whitespace:

`python
terms = [t.lower() for t in question.split() if len(t) > 2]
`

This means a search for "authentication?" stays as "authentication?" and never matches a node with label "Authentication" or "auth_service".

The same issue affects _find_node (line 328), which is used by graphify path and graphify explain - if the user passes "Authentication?" as an argument, the ? prevents matching.

## Example

`ash
graphify query "how does the authentication work?"
# Searches for ["how", "does", "authentication?", "work?"]
# "authentication?" never matches node "Authentication" or "auth_service"
# Falls back to unrelated nodes matching shorter terms

graphify query "how does the authentication work"
# Works correctly - matches "Authentication", "auth_service" nodes
`

## Suggested Fix

In serve.py, strip punctuation from each term before matching. Change line 309 from:

`python
terms = [t.lower() for t in question.split() if len(t) > 2]
`

to:

`python
import re
terms = [re.sub(r'[^\w\s]', '', t).lower() for t in question.split() if len(t) > 2]
`

A similar cleanup should be applied in _find_node for the label parameter.

azizur100389:
Summary:
- Normalize graph query search tokens by stripping punctuation before scoring or node lookup.
- Add regressions for punctuated query, path/explain lookup helpers, and CLI query behavior.

Why:
- Fixes #978, where searches like "extract?" failed to match nodes labeled "extract".

Tests:
- PASS: python -m pytest tests\test_serve.py -k "punctuation or query_terms_strips or query_graph_text_ignores"
- FAIL: python -m pytest tests\test_query_cli.py -k punctuation (local NetworkX does not support the existing test fixture call node_link_data(..., edges="links"); fixture left unchanged per review)
- FAIL: python -m pytest (collection stops because local environment is missing optional dependency: ModuleNotFoundError: No module named 'datasketch' in tests/test_dedup.py)
- PASS: python -m graphify update . (graph rebuilt; graph.html skipped because graph has 5378 nodes, above the default 5000-node visualization limit)

Scope:
- Intentionally avoids unrelated refactors, dependency changes, and generated-file changes.

### DIFF
--- a/graphify/serve.py
+++ b/graphify/serve.py
@@ -2,6 +2,7 @@
 from __future__ import annotations
 import json
 import math
+import re
 import sys
 from pathlib import Path
 import networkx as nx
@@ -51,16 +52,18 @@ def _strip_diacritics(text: str) -> str:
     return "".join(c for c in nfkd if not unicodedata.combining(c))
 
 
+def _search_tokens(text: str) -> list[str]:
+    return re.findall(r"\w+", _strip_diacritics(str(text)).lower())
+
+
 def _query_terms(question: str) -> list[str]:
     """Split a query into searchable terms, filtering only short English terms."""
     terms: list[str] = []
     for raw in question.split():
-        term = raw.lower().strip()
-        if not term:
-            continue
-        is_english_only = all("a" <= ch <= "z" for ch in term)
-        if not is_english_only or len(term) > 2:
-            terms.append(term)
+        for term in _search_tokens(raw):
+            is_english_only = all("a" <= ch <= "z" for ch in term)
+            if not is_english_only or len(term) > 2:
+                terms.append(term)
     return terms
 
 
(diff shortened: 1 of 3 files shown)


safishamsi (commented):
Thanks for the PR! The core fix looks good, but the test file changes are unrelated to the PR's purpose (you've edited test_build.py to modify node_link_data tests that are unrelated to this PR's scope). Could you revert those test changes and keep only the changes relevant to this PR? That will make it easier to review and merge.

azizur100389:
Addressed, thanks. I reverted the unrelated node_link_data fixture compatibility edits and kept the PR scoped to punctuation normalization plus the focused regression tests. I also updated the PR test section to note that the local CLI regression is blocked by the existing NetworkX fixture call after reverting that change.

safishamsi:
Implemented in v8 (commit 80301a0). Added `_search_tokens` helper using `re.findall(r'\\w+', ...)` that strips punctuation before matching. Updated `_query_terms`, `_score_nodes`, and `_find_node` to use it. Non-Chinese Unicode scripts (Hangul, kana) are handled correctly since the token stripping avoids NFKD normalization on the stored terms. Three new tests (term stripping, `_find_node` punctuation, query-graph-text integration). All 1390 tests pass. Thanks!

PR #999, opened 2026-05-23
fix: prevent machine-specific Dart absolute paths from leaking into child node IDs

Overman775:
## Problem

`extract_dart` was building class and function node IDs from `str(path)` — the absolute local filesystem path — instead of `_file_stem(path)`, which is the convention used by every other extractor.

As a result, the same Dart symbol got a different node ID on every machine:

```
# Developer A
users_alice_projects_myapp_lib_auth_user_service_UserService

# Developer B — same file, different ID
users_bob_desktop_myapp_lib_auth_user_service_UserService
```

The `id_remap` post-processor (introduced in #502) remaps **file node IDs** to project-root-relative paths, but does not touch compound child node IDs — so class and function nodes permanently retained the absolute path of whoever ran `graphify` last.

This broke three things for any team sharing `graphify-out/` in version control:

- **Duplicate nodes** — every developer appends new orphaned nodes on incremental update instead of overwriting the existing ones.
- **Broken edges** — cross-file import and call edges point to non-existent IDs on other machines.
- **AI context degradation** — ghost nodes and dead edges cause incorrect code suggestions and hallucinated imports.

## ⚠️ Migration note

This is a breaking ID change for existing Dart graphs. Old absolute-path-derived node IDs will become orphans on the next incremental run. Rebuild from scratch after updating:

```bash
graphify . --force
```

### DIFF
--- a/graphify/extract.py
+++ b/graphify/extract.py
@@ -2716,6 +2716,8 @@ def extract_dart(path: Path) -> dict:
     except OSError:
         return {"error": f"cannot read {path}"}
 
+    # Use stem (not str(path)) for child IDs to keep them machine-independent.
+    stem = _file_stem(path)
     file_nid = _make_id(str(path))
     nodes = [{"id": file_nid, "label": path.name, "file_type": "code",
               "source_file": str(path), "source_location": None}]
@@ -2724,7 +2726,7 @@ def extract_dart(path: Path) -> dict:
 
     # Classes and mixins
     for m in re.finditer(r"^\s*(?:abstract\s+)?(?:class|mixin)\s+(\w+)", src, re.MULTILINE):
-        nid = _make_id(str(path), m.group(1))
+        nid = _make_id(stem, m.group(1))
         if nid not in defined:
             nodes.append({"id": nid, "label": m.group(1), "file_type": "code",
                           "source_file": str(path), "source_location": None})
@@ -2738,7 +2740,7 @@ def extract_dart(path: Path) -> dict:
         name = m.group(1)
         if name in {"if", "for", "while", "switch", "catch", "return"}:
             continue
-        nid = _make_id(str(path), name)
+        nid = _make_id(stem, name)
         if nid not in defined:
             nodes.append({"id": nid, "label": name, "file_type": "code",
                           "source_file": str(path), "source_location": None})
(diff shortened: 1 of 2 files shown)


safishamsi (commented):
Correct fix — Dart was the only remaining extractor embedding absolute paths in child node IDs, and switching to `_file_stem` matches every other extractor's convention. Two small cleanup items:

1. **Test assertion**: `assert 'tmp' not in nid` is a fragile negative check that would pass trivially on most CIs and miss a regression that swaps in a different non-portable identifier. Better to assert the positive expected value: `assert nid == _make_id('sample', 'MyClass')` (or equivalent based on the fixture).

2. **Scope wording**: the PR title says "all affected extractors" but Dart was the only one — every other extractor already uses `_file_stem`. Worth correcting the description so reviewers don't think there are other files touched.

3. **Migration note**: this is a format-breaking change for anyone with existing Dart entries in `graphify-out/` — old absolute-path IDs become orphans on next run. A one-line note in the PR body ("existing Dart graphs should be rebuilt with `--force`") would help users hit by this.

Otherwise solid — ready to merge after those tweaks.

Overman775:
> Correct fix — Dart was the only remaining extractor embedding absolute paths in child node IDs, and switching to `_file_stem` matches every other extractor's convention. Two small cleanup items:
> 
> 1. **Test assertion**: `assert 'tmp' not in nid` is a fragile negative check that would pass trivially on most CIs and miss a regression that swaps in a different non-portable identifier. Better to assert the positive expected value: `assert nid == _make_id('sample', 'MyClass')` (or equivalent based on the fixture).
> 2. **Scope wording**: the PR title says "all affected extractors" but Dart was the only one — every other extractor already uses `_file_stem`. Worth correcting the description so reviewers don't think there are other files touched.
> 3. **Migration note**: this is a format-breaking change for anyone with existing Dart entries in `graphify-out/` — old absolute-path IDs become orphans on next run. A one-line note in the PR body ("existing Dart graphs should be rebuilt with `--force`") would help users hit by this.
> 
> Otherwise solid — ready to merge after those tweaks.

Hey Sami, thanks for the feedback! 

I've updated the PR and squashed everything into one clean commit:

- Replaced the negative check in tests with a proper positive assertion using `_make_id` and `_file_stem`. Shortened the docstring too.
- Scoped the PR title/description strictly to Dart.
- Added a quick note in the description about rebuilding with `--force`.

Let me know if it looks good to merge now!

Overman775:
@safishamsi all done, please recheck )

PR #1002, opened 2026-05-24
fix: accept "edges" schema in affected and benchmark commands

SaarShai:
## Summary

`graphify affected` and `graphify benchmark` both crash on graphs produced by `graphify extract --no-cluster`:

```
$ graphify extract . --no-cluster
[graphify extract] wrote graphify-out/graph.json — N nodes, M edges (no clustering)

$ graphify affected SomeSymbol
error: could not load graph: 'links'

$ graphify benchmark graph.json
Traceback ... KeyError: 'links' in node_link_graph
```

Both commands call `json_graph.node_link_graph(raw, edges="links")` wrapped in `try/except TypeError`. The failing case throws a `KeyError` (the JSON has `edges`, not `links`), so the existing fallback never triggers.

The fix is to normalize the schema before the networkx call — the same idiom already used in `graphify/global_graph.py:34-35`:

```python
if "links" not in raw and "edges" in raw:
    raw = dict(raw, links=raw["edges"])
```

## Reproduction

```bash
# in any folder with code
graphify extract . --no-cluster
graphify affected AnySymbol       # was: error: could not load graph: 'links'
graphify benchmark graphify-out/graph.json   # was: KeyError 'links'
```

After this PR, both run cleanly on either schema.

## Test plan

Added regression tests that write a graph with the `edges` top-level key (mimicking `extract --no-cluster`) and assert each command succeeds:

- `tests/test_affected_cli.py::test_affected_cli_handles_edges_schema`
- `tests/test_benchmark.py::test_run_benchmark_handles_edges_schema`

Full suite: `1261 passed, 11 skipped` on this branch.

## Context

Found while integrating graphify into an external skill catalog. Full eval (token-reduction A/B, build-cost curve, quality + staleness probes) measured −93% tokens vs grep+read at parity evidence rate using `graphify explain`. Both `affected` and `benchmark` were on the original "don't recommend in agent flows" list because of this crash; this PR removes them from that list.

### DIFF
--- a/graphify/affected.py
+++ b/graphify/affected.py
@@ -145,6 +145,11 @@ def load_graph(path: Path) -> nx.Graph:
     from networkx.readwrite import json_graph
 
     raw = json.loads(path.read_text(encoding="utf-8"))
+    # Normalize schema: graphs written by `extract --no-cluster` store edges
+    # under the key "edges"; networkx node-link format uses "links". Accept
+    # either. Mirrors the idiom in graphify/global_graph.py.
+    if "links" not in raw and "edges" in raw:
+        raw = dict(raw, links=raw["edges"])
     try:
         return json_graph.node_link_graph(raw, edges="links")
     except TypeError:
--- a/graphify/benchmark.py
+++ b/graphify/benchmark.py
@@ -101,6 +101,11 @@ def run_benchmark(
     from graphify.security import check_graph_file_size_cap
     check_graph_file_size_cap(Path(graph_path))
     data = json.loads(Path(graph_path).read_text(encoding="utf-8"))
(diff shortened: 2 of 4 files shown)


SaarShai:

The fix is preserved on the fork as a single-purpose branch [`SaarShai/graphify:fix/affected-benchmark-edges-links-schema`](https://github.com/SaarShai/graphify/tree/fix/affected-benchmark-edges-links-schema) and as one of four layered fixes on [`SaarShai/graphify:token-economy-patches`](https://github.com/SaarShai/graphify/tree/token-economy-patches) (all four pass the full 1,270-test upstream suite).

Thanks for graphify.

PR #1003, opened 2026-05-24
fix(cluster-only): add --force flag + non-zero exit when overwrite refused

SaarShai:
## Summary

`graphify cluster-only` silently no-ops when the rebuilt graph has fewer nodes than the existing `graph.json`, but prints a misleading success message anyway. There's no CLI flag to override.

`graphify/__main__.py:2201` (pre-PR):
```python
to_json(G, communities, str(out / "graph.json"))   # returns False on refusal — return value ignored
# ...
print(f"Done — {len(communities)} communities. GRAPH_REPORT.md, graph.json and graph.html updated.")
```

When `to_json` hits its node-count safety guard, it returns `False` and prints a WARNING to stderr — but cluster-only doesn't check, then prints "graph.json updated" anyway. Users (and CI) get no exit-code signal.

This PR:
- Parses `--force` from sys.argv. Also honors `GRAPHIFY_FORCE=1`, matching the convention `graphify update` already uses.
- Threads `force=force` into the `to_json` call.
- When `to_json` returns False, prints a clear "NOT updated" message to stderr and `sys.exit(2)`.

## Reproduction

```bash
graphify extract . --no-cluster     # writes N nodes
graphify cluster-only .              # build_from_json drops a few isolates
                                     # → rebuild has N-k, refuses overwrite,
                                     # PRE-PR: printed "graph.json updated" (LIE), exit 0
                                     # POST-PR: prints "NOT updated", exit 2
graphify cluster-only . --force      # POST-PR: overrides, exit 0
```

## Test plan

Two new tests in `tests/test_cli_export.py`:

- `test_cluster_only_refuses_overwrite_on_node_drift_and_exits_nonzero` — inflates the on-disk JSON node count with duplicate-id entries (networkx dedupes on rebuild), asserts rc=2 and graph.json byte-for-byte unchanged.
- `test_cluster_only_force_flag_overrides_node_drift_guard` — same setup with `--force`, asserts rc=0 and the phantom duplicates are gone from the rebuilt graph.

Full suite: `1261 passed, 11 skipped` on this branch.

### DIFF
--- a/graphify/__main__.py
+++ b/graphify/__main__.py
@@ -2126,6 +2126,8 @@ def main() -> None:
         # Mirror the tree/export arg-parsing pattern: walk argv so flags and
         # the optional positional path can appear in any order (#724).
         no_viz = "--no-viz" in sys.argv
+        force = ("--force" in sys.argv
+                 or os.environ.get("GRAPHIFY_FORCE", "").lower() in ("1", "true", "yes"))
         _min_cs_arg = next((a for a in sys.argv if a.startswith("--min-community-size=")), None)
         min_community_size = int(_min_cs_arg.split("=")[1]) if _min_cs_arg else 3
         args = sys.argv[2:]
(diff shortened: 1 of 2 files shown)


SaarShai:

The fix is preserved on the fork as a single-purpose branch [`SaarShai/graphify:fix/cluster-only-force-flag`](https://github.com/SaarShai/graphify/tree/fix/cluster-only-force-flag) and as one of four layered fixes on [`SaarShai/graphify:token-economy-patches`](https://github.com/SaarShai/graphify/tree/token-economy-patches) (all four pass the full 1,270-test upstream suite).

Thanks for graphify.

Issue #595, opened 2026-04-28
Feature: native graph hygiene primitives (rename/merge-node/drop-edge/relabel-edge)


anupparikh:
## Problem

The graphify CLI exposes powerful build/cluster/query commands but has no first-class primitive for **graph hygiene** — the inevitable post-extraction cleanup of LLM-introduced noise:

- alias collapse (e.g. `Cipher Labs` ↔ `Cypher Labs`, `Moxie` ↔ `Moxi`)
- dropping wrong `semantically_similar_to` edges between distinct entities the model conflated
- relabeling edges (e.g. `semantically_similar_to` → `variant_of` when one node is an instance of another)
- bulk label rewrites by regex while preserving node IDs (so existing edges survive)

Today the only path is to write a one-off Python script that loads `graph.json` via `networkx.readwrite.json_graph.node_link_graph`, mutates, dumps, and then runs `graphify cluster-only` to regenerate the report. That works but:

1. Not portable across machines/users — every project ends up with its own ad-hoc cleanup script in `graphify-out/work/`.
2. No audit trail format like `ALIAS_CLEANUP.md` is enforced.
3. New users repeatedly rediscover the same pattern.

## Concrete use case

Just ran a 170-transcript Jan/Feb 2026 ingest (~2400 nodes). Sonnet/Opus subagents produced:
- 88 nodes labeled `Moxie` (canonical name in the corpus is `Moxi`)
- 40 nodes labeled `Cipher` / `Cipher Labs` (canonical: `Cypher Labs`)
- 174 `semantically_similar_to` edges, several flatly wrong (`NanoCellect ↔ Vellum`, `Singulator ↔ Moxi`, etc.)

I had to hand-roll `hygiene_pass.py` to fix it, then `graphify cluster-only` to regen the report.

## Proposed shape

```
graphify hygiene <rules.json>
```

Where `rules.json` is something like:

```json
{
  "label_rewrites": [
    {"pattern": "\\bMoxie\\b", "replacement": "Moxi", "flags": ""},
    {"pattern": "\\bCipher Lab(s)?\\b", "replacement": "Cypher Lab\\1", "flags": ""}
  ],
  "drop_edges": [
    {"relation": "semantically_similar_to", "between": ["NanoCellect", "Vellum (Roche spinout)"]},
    {"relation": "semantically_similar_to", "between": ["Singulator", "Moxi"]}
  ],
  "relabel_edges": [
[...]

PR #1004, opened 2026-05-24
fix(update): evict nodes from re-extracted files, not just by id-match

SaarShai:
## Summary

\`graphify update .\` is documented as the AST-only refresh after edits. In practice it's **additive**: a rename like \`def foo → def bar\` leaves the old \`foo\` node in the graph forever, alongside the new \`bar\` node. Same for outright deletions.

Reproduced:

\`\`\`bash
# initial state
echo 'def read_page(): pass' > app.py
graphify extract .
graphify explain read_page             # works

# edit: rename in-place
echo 'def read_page_RENAMED(): pass' > app.py
graphify update . --force
graphify explain read_page             # still works — BUG, function no longer exists
graphify explain read_page_RENAMED     # also works
# → graph now contains BOTH symbols. Agents querying \`read_page\` get
#   confident "yes it exists" answers about a deleted function.
\`\`\`

## Root cause

\`graphify/watch.py\` \`_rebuild_code\` preserves nodes from the previous graph and filters them by:

\`\`\`python
preserved_nodes = [
    n for n in existing.get("nodes", [])
    if n["id"] not in new_ast_ids
    and (not evict_sources or n.get("source_file") not in evict_sources)
]
\`\`\`

The \`evict_sources\` set is **only populated when \`changed_paths is not None\`** (the watcher path). The CLI path (\`graphify update .\`) calls \`_rebuild_code\` with \`changed_paths=None\`, so \`evict_sources\` stays empty — the filter degenerates to "drop only nodes whose id matches a new one". Since a renamed function gets a new node-id, the old node passes both filters and lives forever.

## Fix

Always populate \`evict_sources\` with every file in \`extract_targets\`. A file being re-extracted is by definition authoritative for its own nodes — anything previously preserved from that source path is stale.

Same eviction loop, just hoisted out of the \`if changed_paths is not None:\` guard.

## Test plan

Two regression tests in \`tests/test_watch.py\` exercising the CLI code path (no \`changed_paths\`):

- \`test_rebuild_code_full_corpus_evicts_renamed_symbols\` — write \`def read_page\`, rebuild, rename to \`def read_page_RENAMED\`, rebuild again, assert the old node is gone.
- \`test_rebuild_code_full_corpus_evicts_deleted_symbols\` — write two functions, rebuild, delete one, rebuild again, assert only the surviving function is in the graph.

Both fail on \`v8\` before this patch and pass after.

Full suite: \`1261 passed, 11 skipped\` on this branch.

## Related

#595 (broader graph-hygiene primitives — rename/merge-node/drop-edge). This PR is narrower: it fixes the implicit graph-hygiene contract of \`update\` rather than introducing new primitives.

### DIFF
--- a/graphify/watch.py
+++ b/graphify/watch.py
@@ -394,13 +394,22 @@ def _rebuild_code(
                 existing = json.loads(existing_graph.read_text(encoding="utf-8"))
                 existing_graph_data = existing
                 new_ast_ids = {n["id"] for n in result["nodes"]}
+                # Evict preserved nodes whose source_file was just re-extracted.
+                # Without this, a rename like `def foo → def bar` leaves the
+                # old `foo` node in the graph forever: its old node-id no
+                # longer appears in new_ast_ids, so the ID-only filter keeps
+                # it. The same applies to deleted symbols.
+                #
+                # Done unconditionally — both incremental (changed_paths
+                # supplied) and full-corpus (changed_paths is None) rebuilds
+                # need it. Previously the full-corpus path (used by `graphify
+                # update .`) left stale nodes behind on every rename/deletion.
                 evict_sources: set[str] = set(deleted_paths)
-                if changed_paths is not None:
-                    for p in extract_targets:
-                        try:
-                            evict_sources.add(str(p.relative_to(project_root)))
-                        except ValueError:
-                            evict_sources.add(str(p))
+                for p in extract_targets:
+                    try:
+                        evict_sources.add(str(p.relative_to(project_root)))
+                    except ValueError:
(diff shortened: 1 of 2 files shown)


SaarShai:

The fix is preserved on the fork as a single-purpose branch [`SaarShai/graphify:fix/update-evicts-stale-nodes`](https://github.com/SaarShai/graphify/tree/fix/update-evicts-stale-nodes) and as one of four layered fixes on [`SaarShai/graphify:token-economy-patches`](https://github.com/SaarShai/graphify/tree/token-economy-patches) (all four pass the full 1,270-test upstream suite).

Thanks for graphify.

PR #1010, opened 2026-05-24
fix: make graph output deterministic (stop graphify-out churn)

atarutin:
## Problem

`graphify-out/` (`graph.json`, `graph.html`, `GRAPH_REPORT.md`) regenerates **differently on every `graphify update`** even when no source file changed. The committed graph is therefore perpetually dirty, and the post-commit / post-checkout hooks fight every commit and branch switch.

There are **two independent nondeterminism sources**. This PR fixes both.

### 1. Edge direction flips (`build.py`)

`build_from_json` builds an **undirected** `nx.Graph` and stores the original direction in `_src` / `_tgt` edge attributes. When two edges collapse onto the same undirected node pair, **last-write-wins** — and because the edge iteration order isn't stable run-to-run, the serialized `_src`/`_tgt` flips (same edge appears `A→B` one run, `B→A` the next). In a real graph this accounted for roughly half the churn.

**Fix:** iterate edges in a deterministic order (`sorted` by `source, target, relation`) before the add loop, so the last-write outcome is stable.

### 2. Clustering depends on `PYTHONHASHSEED` (`hooks.py`)

The `networkx` Louvain fallback (used when `graspologic` isn't installed) iterates **string-keyed** node/neighbour sets internally, whose order is randomized per-process by Python's hash randomization. Even with `seed=42`, community assignments differ run-to-run, so every node's community id churns.

Empirically, pinning `PYTHONHASHSEED=0` makes clustering fully deterministic. The git hooks invoke Python directly, so the env var has to be set in the generated hook scripts.

**Fix:** export `PYTHONHASHSEED=0` in both `_HOOK_SCRIPT` and `_CHECKOUT_SCRIPT`.

## Verification

With both fixes applied, `graphify update` is idempotent: rebuilding an already-converged `graphify-out` reproduces `graph.json` and `GRAPH_REPORT.md` **byte-for-byte** across repeated runs. Verified on a ~3,770-node / ~12,900-edge real repository graph (stable at 49 communities across runs) where the working tree previously never stayed clean.

## Notes

- Fix 1 is a clean root-cause fix with no behavior change beyond stable ordering.
- Fix 2 is scoped to the generated hook scripts. For users who run `graphify update` manually outside the hooks, consider also setting `PYTHONHASHSEED=0` at CLI entry (e.g. a re-exec guard in `__main__.py`) — happy to add that here if you'd prefer.

### DIFF
--- a/graphify/build.py
+++ b/graphify/build.py
@@ -160,7 +160,18 @@ def build_from_json(extraction: dict, *, directed: bool = False, root: str | Pat
     # slightly different casing or punctuation than the AST extractor.
     # e.g. "Session_ValidateToken" maps to "session_validatetoken".
     norm_to_id: dict[str, str] = {_normalize_id(nid): nid for nid in node_set}
-    for edge in extraction.get("edges", []):
+    # Iterate edges in a deterministic order. The graph is undirected and stores
+    # direction in _src/_tgt; when two edges collapse onto the same node pair the
+    # last write wins, so an unstable iteration order flips _src/_tgt run-to-run
+    # and makes the serialized graph churn. Sorting fixes the last-write outcome.
+    for edge in sorted(
+        extraction.get("edges", []),
+        key=lambda e: (
+            str(e.get("source", e.get("from", ""))),
+            str(e.get("target", e.get("to", ""))),
+            str(e.get("relation", "")),
+        ),
+    ):
         if "source" not in edge and "from" in edge:
             edge["source"] = edge["from"]
         if "target" not in edge and "to" in edge:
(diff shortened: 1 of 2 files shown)


safishamsi (commented):
Both fixes are correct — edge sort is consistent with the same pattern already in `cluster.py`, and `PYTHONHASHSEED=0` in hooks is the right lever for Louvain hash-order sensitivity.

Two requests before merge:

1. **Non-hook gap**: `PYTHONHASHSEED=0` only helps runs triggered by the post-commit/post-checkout hooks. Manual `graphify update` invocations still churn. You mentioned adding a re-exec guard in `__main__.py` — that would complete the fix. Either add it here or open a follow-up issue so it doesn't get lost.

2. **Regression tests**: two trivial additions would lock this in — an idempotency test in `test_build.py` asserting `build_from_json` output is invariant to input-edge ordering, and an `assert 'PYTHONHASHSEED=0' in _HOOK_SCRIPT` check in `test_hooks.py`.

Also worth clarifying in the PR description whether `graph.html` is now stable — it's listed as a churning file in the problem statement but isn't mentioned in the verification claim.

PR #1014, opened 2026-05-25
Clarify Chinese query segmentation scope

### LINKED ISSUE #962: Is graphify will filter string that length < 2?
i Use this query 
```
graphify query "前端 dependency 依赖 install 安装 package manager 包管理器 项目约定"
```

But it can not find my docs.

I ask codex, it answer this:
```

我看了 graphify query 的实现，核心逻辑是：

  terms = [t.lower() for t in question.split() if len(t) > 2]

  也就是说它只按空格切词，并且过滤掉长度 <= 2 的词。

  所以你的 query 里：

  前端 dependency 依赖 install 安装 package manager 包管理器 项目约定

  实际参与检索的只有：

  dependency, install, package, manager, 包管理器, 项目约定

  前端、依赖、安装 都被过滤掉了。
```

### LINKED ISSUE #1013: feat: add CJK word segmentation for graph query
## Problem

`graphify query` does not segment CJK (Chinese/Japanese/Korean) text. Searching for "页面路由" treats it as one opaque string, so nodes with label "路由" are not found. Only searching for the exact substring "路由" works.

## Root Cause

`_query_terms()` in `serve.py` splits queries by whitespace only. CJK text has no spaces between words, so "页面路由" becomes a single term. The substring matching (`t in norm_label`) is one-directional — it checks if the term is inside the label, not the reverse.

## Proposed Solution

Add jieba-based CJK word segmentation to `_query_terms()`:
- "页面路由" → terms = ["页面", "路由", "页面路由"]
- Segmented terms enable substring matching ("路由" matches nodes with "路由" in label)
- Original term kept for exact matching ("页面路由" gets EXACT_MATCH_BONUS)
- Falls back to character bigrams when jieba is not installed
- jieba added as optional dependency: `pip install graphifyy[cjk]`

## MR

#1000

candy-Tong:
## Summary
- Rebase the query tokenization fix onto the latest `v8` branch (`68863a7`).
- Keep the scope explicitly Chinese-only by renaming the CJK helpers and optional extra to Chinese.
- Cache the optional `jieba` import at module level.
- Segment Chinese query terms only when `jieba` is available; when `jieba` is unavailable, keep the original query term unchanged.
- Use real `jieba` behavior in optional segmentation tests instead of mocking the tokenizer.

Fixes #962
Closes #1013

## Tests
- `uv run --with pytest pytest tests/test_serve.py tests/test_query_cli.py tests/test_benchmark.py` — 66 passed, 2 skipped
- `uv run --with pytest --with jieba pytest tests/test_serve.py -k "chinese or non_chinese"` — 5 passed, 1 skipped
- `graphify update .`

### DIFF
--- a/graphify/serve.py
+++ b/graphify/serve.py
@@ -9,6 +9,11 @@ from networkx.readwrite import json_graph
 from graphify.security import sanitize_label, check_graph_file_size_cap
 from graphify.build import edge_data
 
+try:
+    import jieba as _jieba  # type: ignore[import-untyped]
+except ImportError:
+    _jieba = None
+
 
 def _load_graph(graph_path: str) -> nx.Graph:
     try:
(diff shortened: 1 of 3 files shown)


safishamsi (commented):
Good contribution — CJK segmentation is a real gap and the jieba + bigram fallback design is sound. Two small things before merge: (1) `_has_cjk()` only covers U+4E00–U+9FFF (Chinese), so Japanese (Hiragana/Katakana) and Korean (Hangul) users get no segmentation — either broaden the ranges or rename to `_has_chinese` so the scope is honest; (2) jieba is imported on every `_segment_cjk` call — cache it at module level with a try/except guard. Could you fix those and open a fresh PR against v8?

safishamsi:
Good contribution — CJK segmentation is a real gap and the jieba + bigram fallback design is sound. Two small things before merge: (1) `_has_cjk()` only covers U+4E00–U+9FFF (Chinese), so Japanese (Hiragana/Katakana) and Korean (Hangul) users get no segmentation — either broaden the ranges or rename to `_has_chinese` so the scope is honest; (2) jieba is imported on every `_segment_cjk` call — cache it at module level with a try/except guard. Could you fix those and open a fresh PR against v8?

candy-Tong:
Okay, I'll revise it and resubmit the PR later.

safishamsi:
The issues raised here (rename `_has_cjk` → `_has_chinese`, cache jieba at module level) were addressed in #1026 which has already been merged and shipped in v0.8.19. Thanks @candy-Tong — no need to resubmit, this is now resolved.

PR #1024, opened 2026-05-26
Add nix flake

vkarasen:
This pull request adds native Nix flake support for the project, making it easier for users to install and use `graphify` via the Nix package manager. The main changes include introducing a `flake.nix` file to define the Nix build, updating the documentation to describe Nix-based installation methods, and listing Nix as an alternative installation option.

**Preparatory work**

We require a checked-in `uv.lock` to be able to make reproducible builds, so the first commit adds it and removes it from .gitignore

**Nix flake integration and documentation updates:**

* Added a new `flake.nix` file that defines how to build and package `graphify` using Nix flakes, including support for development shells and direct execution as a Nix flake app.
* Updated the `README.md` to list Nix as an alternative installation method alongside pipx and uv, and provided instructions for running or packaging `graphify` using Nix flakes.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -104,3 +104,22 @@ jobs:
       - name: pip-audit (dependency vulnerabilities)
         continue-on-error: true
         run: uv run --frozen pip-audit --strict
+  nix:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Install Nix
+        uses: cachix/install-nix-action@v27
+        with:
+          extra_nix_config: |
+            experimental-features = nix-command flakes
+
+      - name: Check flake
+        run: nix flake check --all-systems
+
+      - name: Build default package
+        run: nix build .#default
+
+      - name: Verify built binary runs
+        run: nix run .#default -- --help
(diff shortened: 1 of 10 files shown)


safishamsi (commented):
Nice addition — the flake design is idiomatic and the `numba`/`nuitka` overrides are the right call. Three things before merge: (1) typo in flake.nix description: `graaphify`; (2) the README NixOS snippet is broken — `${system}` is undefined in that scope and `environment.systemPackages` can't be set at the top level of `outputs`, which will confuse users; (3) no CI job for `nix flake check` or `nix build .#default`, so the flake can silently break on dependency updates. Could you fix those and open a fresh PR against v8?

vkarasen:
@safishamsi 

just pushed an update here. Adding the ci job became a little bit of a side-quest, some of the existing tests were misbehaving when running in a sandbox, so I did a few touches to fix that. Running `nix flake check` will also execute pytest, but skip the tests that need network.

If needed, one can still run pytest through the devshell via `nix develop . --command 'uv run pytest'`

vkarasen:
rebased on latest v8. I had to skip a few more tests which look up old commits because `.git/` is not available in a nix sandbox

vkarasen:
rebased on latest v8 @safishamsi

PR #1028, opened 2026-05-26
fix(cli): apply remap_communities_to_previous in cluster-only path

### LINKED ISSUE #1027: cluster-only skips overlap-based community remapping, so labels become misaligned after re-clustering
## Summary
`graphify cluster-only` re-runs Leiden clustering and then re-applies the existing
`.graphify_labels.json` by raw cid index, without invoking the
`remap_communities_to_previous` safety net that PR #822 introduced for the
`watch` / `update` paths. As a result, community labels attach to clusters whose
members are unrelated to the original label's meaning whenever the underlying
graph has changed between the labeling pass and the `cluster-only` invocation.

## Reproduce
```bash
# 1) Full build with Step 5 labeling
/graphify .                          # or any equivalent that writes .graphify_labels.json

# 2) Make any code change that alters cluster sizes

# 3) Re-extract code without clustering (the official #822 workflow)
graphify update --no-cluster .

# 4) Re-cluster
graphify cluster-only .

# 5) Inspect graphify-out/GRAPH_REPORT.md or graph.html
#    → community labels are attached to clusters whose actual members
#      are unrelated to the label's original meaning.
```

The mismatch is deterministic: `cluster()` reindexes communities by size-descending
after each run, and the prior labels file is read back verbatim by cid index in
`graphify/__main__.py:1714-1722`.

## Expected
After `cluster-only`, labels should follow the same overlap-based remapping that
PR #822 added to the watch/update paths — a label associated with a previous
community should attach to the new community whose node set most overlaps with
the old one.

## Actual
Labels in `.graphify_labels.json` are reapplied verbatim by cid index, producing
misalignment whenever the graph topology has changed between labeling and
re-clustering.

## Affected file
`graphify/__main__.py:1714-1722` — the `elif cmd == \"cluster-only\":` branch.

The neighbouring code path in `graphify/watch.py` (`_rebuild_code`, around
lines 465-468) calls `remap_communities_to_previous` correctly — `cluster-only`
is the only re-clustering entry point that omits it.

## Suggested fix
Mirror the working pattern from `watch.py`. Immediately after
`communities = cluster(G)` in the `cluster-only` branch, build the previous
node→community map from `_raw` (already loaded a few lines above) and apply
the remap:

```python
from graphify.cluster import remap_communities_to_previous

previous_node_community = {
    n[\"id\"]: n[\"community\"]
    for n in _raw.get(\"nodes\", [])
    if n.get(\"community\") is not None and n.get(\"id\") is not None
}
if previous_node_community:
    communities = remap_communities_to_previous(communities, previous_node_community)
```

This matches the design intent stated in PR #822:

> \"stabilize clustering output with deterministic partition input ordering,
>  seeded Leiden when supported, and overlap-based remapping of new communities
>  to prior IDs\"

It looks like `cluster-only` was missed when #822 was applied. If this was an
intentional omission (e.g. for a reason that does not apply to `watch`/`update`),
I'd love to understand the rationale before proposing a PR — happy to send one
mirroring the fix otherwise.

## Environment
- `graphifyy` 0.7.18 (PyPI)
- Python 3.12.3
- Linux (WSL2 / Ubuntu 24.04)

bgmbgm94:
## Summary
`cluster-only` re-runs Leiden clustering and then re-applies the existing
`.graphify_labels.json` by raw cid index, without invoking the
`remap_communities_to_previous` safety net that #822 added to the
`watch` / `update` paths. As a result, labels become misaligned with the
actual community contents whenever the graph has changed between the
labeling pass and the `cluster-only` invocation — including the official
`update --no-cluster` + `cluster-only` workflow introduced by #822.

Closes #1027.

## What changed
Mirror the existing pattern in `graphify/watch.py:_rebuild_code`:

- Build `previous_node_community` from `_raw` (already loaded a few lines
  above) using each node's `community` attribute.
- Call `remap_communities_to_previous(communities, previous_node_community)`
  immediately after `cluster(...)` so the existing labels file keeps
  tracking the same conceptual communities by node-overlap, not by raw
  cid index.

Diff is contained to the `elif cmd == "cluster-only":` branch in
`graphify/__main__.py` (one import addition, one block right after the
`cluster(...)` call).

## Test
Added `tests/test_cli_export.py::test_cluster_only_remaps_labels_to_previous_cids`.

The test tags every node with out-of-band sentinel community ids
(`4242` / `9999`), writes a labels file keyed on those ids, then runs
`graphify cluster-only`. With the fix, at least one sentinel cid still
appears in the final `graph.json` community attributes (= remap matched
by node overlap). Without the fix, Leiden renumbers to `0, 1, ...` and
the prior labels become orphaned.

Verified locally:

```
=== WITH fix ===
1 passed in 17.56s

=== WITHOUT fix (stashed source change) ===
FAILED tests/test_cli_export.py::test_cluster_only_remaps_labels_to_previous_cids
1 failed in 17.14s
```

Full suite passes:
```
pytest tests/ -q --ignore=tests/bench_extract.py
1317 passed in 58.20s
```

## Open question
The `remap_communities_to_previous` helper already exists in
`graphify.cluster` and is used correctly in `watch.py:_rebuild_code`.
#822's description explicitly mentions:

> "stabilize clustering output with deterministic partition input
>  ordering, seeded Leiden when supported, and overlap-based remapping
>  of new communities to prior IDs"

— but `cluster-only` was not updated alongside the watch/update path. The
neighbouring code paths re-applying labels (`watch.py:_rebuild_code`,
`update`) all go through remap; `cluster-only` is the only
re-clustering CLI entry point that omitted it.

Was there a design reason to deliberately keep `cluster-only` without
the remap (e.g. a use case where label inheritance is undesirable)?
If so, happy to revise this PR — for example, gate the remap behind an
opt-out flag, or expose a `--reset-labels` switch instead. Otherwise it
looks like a simple miss when #822 was applied and this PR just mirrors
the existing pattern.

## Refs
- Issue: #1027
- Stability work that introduced `remap_communities_to_previous`: #822

### DIFF
--- a/graphify/__main__.py
+++ b/graphify/__main__.py
@@ -2238,7 +2238,7 @@ def main() -> None:
             sys.exit(1)
         from networkx.readwrite import json_graph as _jg
         from graphify.build import build_from_json
-        from graphify.cluster import cluster, score_all
+        from graphify.cluster import cluster, score_all, remap_communities_to_previous
         from graphify.analyze import god_nodes, surprising_connections, suggest_questions
         from graphify.report import generate
         from graphify.export import to_json, to_html
@@ -2250,6 +2250,18 @@ def main() -> None:
         print(f"Graph: {G.number_of_nodes()} nodes, {G.number_of_edges()} edges")
         print("Re-clustering...")
         communities = cluster(G, resolution=co_resolution, exclude_hubs_percentile=co_exclude_hubs)
+        # Mirror the watch/update path (#822): map new cids to prior ones by
+        # node-overlap so the existing .graphify_labels.json keeps attaching
+        # to the same conceptual community after re-clustering. Without this,
+        # labels follow raw cid index and become misaligned whenever the
+        # graph has changed between labeling and cluster-only (#1027).
+        previous_node_community = {
+            n["id"]: n["community"]
+            for n in _raw.get("nodes", [])
+            if n.get("community") is not None and n.get("id") is not None
+        }
+        if previous_node_community:
+            communities = remap_communities_to_previous(communities, previous_node_community)
         cohesion = score_all(G, communities)
         gods = god_nodes(G)
         surprises = surprising_connections(G, communities)
--- a/tests/test_cli_export.py
+++ b/tests/test_cli_export.py
@@ -288,6 +288,58 @@ def test_cluster_only_creates_output_dir_when_missing(tmp_path):
     assert (tmp_path / "graphify-out" / "GRAPH_REPORT.md").exists()
 
 
+# Regression test for #1027 - cluster-only must remap labels via node overlap
+
+def test_cluster_only_remaps_labels_to_previous_cids(tmp_path):
+    """cluster-only must invoke remap_communities_to_previous so the existing
+    .graphify_labels.json keeps tracking the same conceptual communities after
+    re-clustering. Without the remap call, Leiden's size-descending cid order
+    re-applies labels by raw index and they silently misalign with cluster
+    contents (#1027). Mirror of the watch/update fix from #822.
+    """
+    out = _make_graph(tmp_path)
+    graph_json = out / "graph.json"
+    labels_json = out / ".graphify_labels.json"
+
+    # Tag every node with an out-of-band community id and write a labels file
+    # keyed on those ids. After cluster-only, at least one of those sentinel
+    # ids must survive in the labels file (= remap succeeded by node overlap).
+    # If the cluster-only branch skips remap, Leiden returns small ints
+    # (0, 1, ...) and the sentinel keys disappear entirely.
+    g = json.loads(graph_json.read_text(encoding="utf-8"))
+    nodes = g.get("nodes", [])
+    assert len(nodes) >= 4, "fixture must have enough nodes to form 2+ communities"
+    sentinel_a, sentinel_b = 4242, 9999
+    half = len(nodes) // 2
+    for i, n in enumerate(nodes):
+        n["community"] = sentinel_a if i < half else sentinel_b
+    graph_json.write_text(json.dumps(g), encoding="utf-8")
+    labels_json.write_text(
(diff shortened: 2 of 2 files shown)


bgmbgm94:
After re-reading `skill.md` more carefully, I want to flag that this PR may not be the right shape — and would love your guidance before you spend review time.

The `cluster-only` section in `skill.md` says:

> Skip Steps 1–3. Re-run clustering on the existing graph: `graphify cluster-only .`
> Then run **Steps 5–9 as normal** (label communities, generate viz, benchmark, clean up, report).

And Step 5 itself reads:

> ### Step 5 - Label communities
> Read `graphify-out/.graphify_analysis.json`. For each community key, look at its node labels and write a 2-5 word plain-language name…

That language reads to me as **"label from scratch using the new community contents"**, not "reuse the prior labels file". If that's correct, then the existing behavior of `cluster-only` reading back the old `.graphify_labels.json` at all (by raw cid index) is itself the divergence from intent — and the right fix might actually be to **drop / ignore the prior labels file** on `cluster-only` rather than to teach it to remap them.

If that interpretation matches what you had in mind, two alternatives look more consistent with `skill.md`'s intent than this PR:

- **B — drop labels on `cluster-only`**: clear or skip `.graphify_labels.json` in the cluster-only branch, leaving Step 5 to re-label. Smallest fix, matches the doc.
- **C — opt-in remap**: default to dropping labels (option B), but expose `--keep-labels` for users who explicitly want the watch-style overlap remap.

Apologies for not raising this in the original PR description — I was anchoring too hard on "watch already does this, mirror it." Could you let me know which direction you'd prefer?

1. **Land A as-is** (mirror watch — overlap remap by default on `cluster-only`)
2. **Replace A with B** (drop labels — closest to `skill.md` Step 5 intent)
3. **Replace A with C** (default-drop + `--keep-labels` opt-in)
4. Something else entirely

I'm happy to amend this PR in whichever direction you point at — or to close it and open a fresh one if you'd rather start from a clean slate.

PR #1030, opened 2026-05-26
Add --mode deep support to graphify extract CLI.

dobryakov:
Wire deep_mode through the LLM extraction pipeline so headless runs and Gemini fast-path skill flows match slash-command deep extraction behavior.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ Full release notes with details on each version: [GitHub Releases](https://githu
 - Fix: `backup_if_protected` no longer accumulates one folder per run — uses content-hash comparison to skip identical backups and overwrite in-place when content changes; one folder per day maximum
 - Feat: Devin CLI support — `graphify devin install/uninstall` installs the skill into Devin's `.devin/rules/` directory (#1020)
 - Fix: TypeScript 5.0 array-form `extends` in `tsconfig.json` now handled — `_read_tsconfig_aliases` normalizes `extends` to a list before iteration (#1017)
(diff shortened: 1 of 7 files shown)


safishamsi:
Good addition — threading `deep_mode` through the call chain is the right approach and the default-off design is correct. A few things to address before merge:

**Missing test coverage:** `_call_claude_cli` and `_call_bedrock` are touched by the PR but not covered in `tests/test_deep_mode.py`. Please add or parameterize tests for those two backends.

**Print fires too late:** The "deep mode enabled" confirmation print is inside `if uncached_paths:` — on a fully-cached run it never fires, so the user gets no feedback that `--mode deep` was recognized. Move it before the cache check.

**`_DEEP_EXTRACTION_SUFFIX` wording:** "latent couplings" is vague enough that different models will interpret it very differently, and more edges means more token output → more bisection retries → higher cost. Worth tightening the language and noting the cost tradeoff in the PR description/changelog.

**Minor:** the `--mode` flag isn't mentioned in the top-level `help` output alongside other extract flags — worth adding for discoverability.

Otherwise looks good. Fix the above and this is ready to merge.

dobryakov:
Addressed feedback in follow-up commit 1ccc75a:
- Added deep_mode coverage for claude-cli and bedrock paths in tests/test_deep_mode.py.
- Moved deep-mode confirmation print before semantic cache check, so fully-cached runs still show mode recognition.
- Tightened DEEP_MODE prompt wording toward concrete architectural signals.
- Added changelog note about deep mode potentially increasing token usage/cost and adaptive bisection retries.

safishamsi:
Implemented manually on v8 to avoid merge conflicts with recent changes (amp platform, OpenCode path fix, cluster-only remap). The `--mode deep` flag is now fully wired: parsing in `__main__.py`, `_DEEP_EXTRACTION_SUFFIX` / `_extraction_system(deep=)` in `llm.py`, and `deep_mode` threaded through all four LLM backends and `extract_corpus_parallel`. Thanks for the contribution!

PR #1034, opened 2026-05-26
feat: MCP config extractor (.mcp.json, claude_desktop_config.json)

adityachaudhary99:
## Why

Graphify already exposes itself **as** an MCP server via `graphify/serve.py` and the `graphify --mcp` flag. This PR adds the other direction — indexing MCP server configs **as** a corpus type, so an agent running graphify on a project can query its own configured MCP layer.

Concretely: a Claude Code / Cursor / OpenCode user has a `.mcp.json` (or `claude_desktop_config.json`) that defines the agent's actual tool surface — which servers it can talk to, what commands they run, what env vars they need. That surface is invisible to graphify today. After this PR, those configs are first-class nodes in the same graph as the user's code.

## What it does

Filename-routed extractor for the standard MCP config filenames:

- `.mcp.json` (Claude Code project config)
- `claude_desktop_config.json` (Claude Desktop)
- `mcp.json`, `mcp_servers.json` (generic / alternates)

`_get_extractor` checks filename via `is_mcp_config_path` **before** falling back to `_DISPATCH.get(path.suffix)`. Generic `.json` files continue to flow through `extract_json` unchanged.

### Schema

| Node kind | Scope | Notes |
|---|---|---|
| `mcp_config_file` | per-file | the config itself |
| `mcp_server` | per-config | one per entry under `mcpServers` |
| `mcp_command` | **global** | `npx`, `uvx`, `node`, ... — same command across two configs collapses to one node |
| `mcp_package` | **global** | `@modelcontextprotocol/server-filesystem`, `mcp-server-fetch`, ... — same |
| `env_var` | **global** | env var **names only**, never values |

| Relation | Direction | Meaning |
|---|---|---|
| `contains` | file → server | natural containment |
| `references` | server → command, server → package | runtime dependencies |
| `requires_env` | server → env_var | new relation; distinguishes env deps from generic refs |

The global-scope nodes are the value-add: with two configs in a corpus, the graph naturally surfaces "which MCP servers depend on this env var?" / "which configs use this package?" without any cross-file resolver — emergent from the shared IDs.

## Security guarantees

- **Env values are never read.** Only env var names. `env: {"GITHUB_PERSONAL_ACCESS_TOKEN": "ghp_..."}` produces a node labelled `GITHUB_PERSONAL_ACCESS_TOKEN`; the token itself is not opened, persisted, labelled, or surfaced anywhere in the result.
- **`args` are not persisted as nodes or edges.** Some servers embed local filesystem paths or secrets as positional args. The extractor parses `args` only for package detection and discards the rest.
- All labels pass through `sanitize_label` (control chars stripped, length capped).
- File size capped at 1 MiB to match `extract_json`.

## Sample output

Run on `tests/fixtures/sample.mcp.json` (4 servers — filesystem, fetch, github, time):

```
NODES (13)
  [mcp_config_file   ] label=sample.mcp.json
  [mcp_server        ] label=filesystem
  [mcp_command       ] label=npx
  [mcp_package       ] label=@modelcontextprotocol/server-filesystem
  [env_var           ] label=FILESYSTEM_ROOT
  [mcp_server        ] label=fetch
  [mcp_command       ] label=uvx
  [mcp_package       ] label=mcp-server-fetch
  [mcp_server        ] label=github
  [mcp_package       ] label=@modelcontextprotocol/server-github
  [env_var           ] label=GITHUB_PERSONAL_ACCESS_TOKEN
  [mcp_server        ] label=time
  [mcp_package       ] label=mcp-server-time

EDGES (14)
  sample.mcp.json   --contains    --> filesystem
  filesystem        --references  --> npx
  filesystem        --references  --> @modelcontextprotocol/server-filesystem
  filesystem        --requires_env--> FILESYSTEM_ROOT
  sample.mcp.json   --contains    --> fetch
  fetch             --references  --> uvx
  fetch             --references  --> mcp-server-fetch
  sample.mcp.json   --contains    --> github
  github            --references  --> npx
  github            --references  --> @modelcontextprotocol/server-github
  github            --requires_env--> GITHUB_PERSONAL_ACCESS_TOKEN
  sample.mcp.json   --contains    --> time
  time              --references  --> uvx
  time              --references  --> mcp-server-time
```

Note how `npx` collapses to one node despite being used by both `filesystem` and `github`, and `uvx` collapses across `fetch` and `time` — the cross-config dependency surface comes for free.

## Tests

29 tests, all passing locally. Coverage:

- Filename detection (recognition + rejection of generic `.json`)
- Happy-path emission of every server / command / package / env_var
- Package detection: npm scoped, npm with version stripping, pypi `mcp-server-*` / `*-mcp` patterns, `-y`/`--flag` skipping, unknown arg shapes
- Env-name-only persistence (asserts env values never appear in any node, edge, or metadata field)
- `args` not persisted as nodes (path-leak test)
- Cross-config emergent edges: same command + env var across two configs collapse to one node
- Per-directory server-node isolation (two configs in different dirs with same server name don't collide)
- Malformed JSON, oversize files, root-not-object, non-dict server entries, missing `mcpServers` key, alternate `{"mcp":{"servers":...}}` shape
- End-to-end dispatch routing via `_get_extractor`

## Design notes

- **Pure stdlib, no tree-sitter.** MCP configs are already structured JSON; running them through `tree-sitter-json` would just add ceremony without giving us more semantic information than `json.loads` produces.
- **`requires_env` is the only new relation.** I considered overloading `references`, but env vars are a categorically distinct runtime dependency (a missing one breaks the server at launch) and queries like "what env vars does my agent need?" benefit from a distinct relation. Happy to fold into `references` with a `context="env"` if you'd prefer.
- **Module lives at top level (`graphify/mcp_ingest.py`) rather than inside `extract.py`.** Matches the precedent set by `scip_ingest.py` and `google_workspace.py` — single-corpus concern, kept out of the already-large `extract.py`. Discussion #532 was a signal that integration code probably shouldn't accumulate in the central files.

## Out of scope (for follow-ups)

- Tool-level edges (`server → tool`) — would require either running each MCP server's `tools/list` RPC (non-deterministic, security-sensitive) or maintaining a curated lookup table for well-known packages. Not in this PR.
- Sidecar `tools.json` ingestion (some servers publish a static tool manifest alongside their package). Easy add later.
- MCP registry (`server.json` per the registry spec). Easy add later.

---

Closes nothing directly but addresses the asymmetry that surfaced in discussion #500 ("Copilot was falling back to text search instead of using the MCP-backed graph context") and the architecture conversation in #532. Happy to iterate on relation naming, node-kind metadata, or the global-vs-per-file scope choices if any of them clash with the v8 direction.

### DIFF
--- a/graphify/extract.py
+++ b/graphify/extract.py
@@ -10,6 +10,7 @@ from dataclasses import dataclass, field
 from pathlib import Path
 from typing import Callable, Any
 from .cache import load_cached, save_cached
+from .mcp_ingest import extract_mcp_config, is_mcp_config_path
 
 _RECURSION_LIMIT = 10_000
 
@@ -7914,6 +7915,11 @@ def _get_extractor(path: Path) -> Any | None:
     """Return the correct extractor function for a file, or None if unsupported."""
     if path.name.endswith(".blade.php"):
         return extract_blade
+    # MCP config files (.mcp.json, claude_desktop_config.json, ...) are routed
+    # by filename before generic .json dispatch so they get MCP-aware nodes
+    # (servers, commands, packages, env vars) instead of opaque JSON keys.
+    if is_mcp_config_path(path):
+        return extract_mcp_config
     return _DISPATCH.get(path.suffix)
 
 
--- a/graphify/mcp_ingest.py
+++ b/graphify/mcp_ingest.py
@@ -1,0 +1,392 @@
+"""mcp_ingest.py — Extract MCP (Model Context Protocol) server configuration files.
+
+Reads `.mcp.json` / `claude_desktop_config.json` / `mcp.json` / `mcp_servers.json`
+and turns the `mcpServers` map into Graphify nodes and edges.
+
+Symmetry with `serve.py`: Graphify exposes itself AS an MCP server. This module
+indexes MCP servers AS a corpus type, completing the loop — an agent that runs
+graphify with `--mcp` can now query its own configured MCP layer.
+
+Entry point:
+  extract_mcp_config(path: Path) -> dict[str, list[dict]]
+
+  Returns `{"nodes": [...], "edges": [...]}` compatible with Graphify's
+  extraction-result format. Returns `{"nodes": [...], "edges": [...], "error": "..."}`
+  when the file is malformed, too large, or has no `mcpServers` map — the empty
+  result keeps it indistinguishable from "no MCP config here" for downstream
+  callers.
+
+Detected filenames (case-sensitive, matched on basename):
+  - .mcp.json                       (Claude Code project config)
+  - claude_desktop_config.json      (Claude Desktop)
+  - mcp.json                        (generic / per-tool)
+  - mcp_servers.json                (alternate naming)
+
+Schema emitted:
+  Node kinds:
+    - file              the config file itself (label = filename)
+    - mcp_server        one per entry under mcpServers
+    - mcp_command       executable (npx, uvx, node, python, ...) — global ID
+    - mcp_package       npm / pypi package id parsed from args — global ID
+    - env_var           env variable NAME only — global ID. VALUES ARE NEVER READ.
+
+  Edge relations:
+    - contains          file -> mcp_server
+    - references        mcp_server -> mcp_command
+    - references        mcp_server -> mcp_package
+    - requires_env      mcp_server -> env_var   (new relation; distinguishes
+                                                  env dependencies from generic refs)
+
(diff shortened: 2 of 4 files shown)


safishamsi:
Merged manually after fixing two issues: server_count budget is now checked after the validity guard (invalid entries no longer consume capacity), and the misleading 'uv run' example in the `_PY_MCP_PKG_RE` docstring was corrected. Thanks @adityachaudhary99!

Issue #1007, opened 2026-05-24
Bug: Nodes and edges from deleted/moved files persist in graph — "already clean" false negative


sub4biz:
## Summary

When files are moved out of the scan root, graphify detects them as deleted but fails to remove their nodes and edges from the graph. Both `extract` and `update` report "already clean — no matching nodes or edges in graph", yet the stale nodes remain.

---

## Steps to Reproduce

**1. Create a minimal corpus:**

```bash
mkdir -p corpus/module-a corpus/module-b
```

`corpus/module-a/auth.py`:
```python
def login(user, password):
    return validate(user, password)
def validate(user, password):
    return user == "admin"
def logout(user):
    pass
```

`corpus/module-b/utils.py`:
```python
def format_date(date):
    return str(date)
def parse_date(s):
    return s.strip()
```

**2. Build initial graph:**
```bash
cd corpus
graphify extract .
# → graph.json: 7 nodes, 6 edges (4 from module-a, 3 from module-b)
```

**3. Move module-b out of scan root:**
```bash
mv module-b ../module-b-moved
```

**4. Re-run extraction:**
```bash
graphify extract .
```

**5. Alternatively — same result with update:**
```bash
graphify update .
```

---

## Expected Behavior

After steps 4 or 5 — graph updated to **4 nodes, 4 edges**, only `module-a` nodes remain. `module-b/utils.py`, `format_date()`, `parse_date()` and their edges removed.

---

## Actual Behavior

**`graphify extract .` output:**
```
[graphify extract] 0 code changed; 1 unchanged; 1 deleted
[graphify] 1 source file(s) deleted since last run — no matching nodes or edges in graph, already clean.
[graphify extract] wrote graph.json: 7 nodes, 6 edges   ← unchanged
```

**`graphify update .` output:**
```
[graphify watch] No code-graph topology changes detected; outputs left untouched.
```

In both cases graph still contains **7 nodes, 6 edges** including all 3 nodes from the deleted file:
```
module-b/utils.py  |  utils.py
module-b/utils.py  |  format_date()
module-b/utils.py  |  parse_date()
```

---

## Root Cause Hypothesis

Likely a path normalization mismatch: the manifest stores **absolute paths** (e.g. `D:\corpus\module-b\utils.py`) while node `source_file` attributes store **relative paths** (e.g. `module-b/utils.py`). The pruning lookup fails to match them, reports "already clean", and skips deletion.

---

## Impact

[...]

PR #1036, opened 2026-05-26
fix: evict stale nodes in _rebuild_code for graphify update (#1007)

sub4biz:
### Problem

`graphify update` (`_rebuild_code` with no `changed_paths`) silently preserved nodes
from deleted files. The fix in `build_merge` (eef623a) covered `graphify extract`
but `_rebuild_code` has its own eviction logic that was not touched.

Reproduction with v8 branch after eef623a:

```
graphify extract corpus        # → 7 nodes (module-a + module-b)
rm -r corpus/module-b
graphify update corpus         # → "No code-graph topology changes detected" ✗
                               #    graph.json still has 7 nodes
```

### Root cause

When `changed_paths is None` (plain `graphify update`), `evict_sources` is never
populated — deleted files are simply absent from the new extraction and end up in
`preserved_nodes` unchanged:

```python
# watch.py — before this fix
evict_sources: set[str] = set(deleted_paths)   # always empty when changed_paths is None
if changed_paths is not None:
    ...                                         # evict_sources never filled for plain update
preserved_nodes = [
    n for n in existing.get("nodes", [])
    if n["id"] not in new_ast_ids
    and (not evict_sources or n.get("source_file") not in evict_sources)  # guard is no-op
]
```

Additionally, `str(cand.relative_to(project_root))` produces backslash paths on
Windows (`module-b\utils.py`) while `graph.json` stores forward-slash paths
(`module-b/utils.py`) — causing eviction to silently fail even when
`changed_paths` is provided.

### Fix

**watch.py — three changes:**

1. `_relativize_source_files` (line 141): `str(...)` → `.as_posix()` —
   fixes forward-slash normalization for absolute-path conversion.

2. `deleted_paths` / `evict_sources` (lines 379, 415): `str(...)` →
   `_norm_source_file(str(p), root)` — same function used in `build_merge`,
   so prune keys are normalized identically to node `source_file` attrs.

3. `_rebuild_code` reconciliation: when `changed_paths is None`, iterate
   existing graph nodes and build `current_sources` — both sides normalized
   via `_norm_source_file` so prune keys and node `source_file` attrs are
   compared identically (mirrors `build_merge` fix in eef623a).
   Any code node whose normalized source is absent from disk is added to
   `evict_sources` and `deleted_paths` (to pass the shrink-guard).

### Test

`tests/test_watch.py::test_rebuild_code_evicts_nodes_from_deleted_files` —
builds a two-file corpus, deletes one file, calls `_rebuild_code` directly,
asserts the deleted file's nodes are gone and the surviving file's nodes remain.

Full test run: 47 passed, 7 skipped across `test_build.py` and `test_watch.py`
(skipped: POSIX fcntl + watchdog — not applicable on Windows).

### DIFF
--- a/graphify/watch.py
+++ b/graphify/watch.py
@@ -138,7 +138,7 @@ def _relativize_source_files(payload: dict, root: Path) -> None:
             if not source_path.is_absolute():
                 continue
             try:
-                item["source_file"] = str(source_path.resolve().relative_to(root))
+                item["source_file"] = source_path.resolve().relative_to(root).as_posix()
             except ValueError:
                 continue
 
@@ -376,7 +376,7 @@ def _rebuild_code(
                     # (e.g. .gitignore, vendored). Either way, evict any
                     # preserved nodes that still claim this source path.
                     try:
-                        deleted_paths.add(str(cand.relative_to(project_root)))
+                        deleted_paths.add(cand.relative_to(project_root).as_posix())
                     except ValueError:
                         deleted_paths.add(str(cand))
             if not wanted and not deleted_paths:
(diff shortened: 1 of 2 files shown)


safishamsi:
Implemented the `else:` reconciliation block directly in v8 (commit d1d5751) to avoid merge conflicts — v8 already had the `as_posix()` fixes in slightly different form. The final code uses `_norm_source_file` (aliased as `_nsf`) on both the eviction and current-sources sides, mirroring exactly what `build_merge` does. Also ported your test (`test_rebuild_code_evicts_nodes_from_deleted_files`) to `tests/test_watch.py` — all 1387 tests pass. Thanks for diagnosing the missing `else:` block!

PR #1041, opened 2026-05-27
feat(llm): add openrouter backend support with configurable API and model options

the-cybersapien:
This pull request adds support for OpenRouter as a new LLM backend in the `graphify/llm.py` configuration. The main change is the introduction of a configuration block for OpenRouter, allowing users to route requests through OpenRouter's OpenAI-compatible API proxy and select from a variety of models.

New LLM backend support:

* Added an `openrouter` configuration to the LLM backends, setting up default parameters, environment variable keys, and documentation for using OpenRouter's API proxy with a default model of `qwen/qwen3.6-flash`.…options

### DIFF
--- a/graphify/llm.py
+++ b/graphify/llm.py
@@ -115,6 +115,20 @@ BACKENDS: dict[str, dict] = {
         "temperature": 0,
         "max_tokens": 16384,
     },
+    "openrouter": {
+        # Routes through OpenRouter's OpenAI-compatible API proxy, which can connect to various backends including Kimi and Gemini. 
+        # Honours the same env var overrides as the underlying backend (e.g. GRAPHIFY_OPENROUTER_MODEL) but 
+        # with a separate API key (e.g. OPENROUTER_API_KEY).
+        # Default model is set to qwen/qwen3.6-flash as it's a strong open-source model that performs well on code tasks and 
+        # is widely accessible via OpenRouter, but users can override with any model supported by their OpenRouter account.
+        "base_url": "https://openrouter.ai/api/v1",
+        "default_model": "qwen/qwen3.6-flash",
+        "env_keys": ["OPENROUTER_API_KEY"],
+        "model_env_key": "GRAPHIFY_OPENROUTER_MODEL",
+        "pricing": {"input": 0.0, "output": 0.0},
+        "temperature": 0,
+        "max_tokens": 16384,
+    },
 }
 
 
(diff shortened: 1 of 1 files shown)


the-cybersapien:
Hello @safishamsi , I previously raised a PR #723 for this. We have several other backends supported now. Openrouter provides a flexible option for this use-case. 
I am currently using custom env vars with ollama backend, this would simplify it a lot.

safishamsi:
Thanks for this! OpenRouter backend is a great addition. A few things to fix before we can merge:

1. **Wrong default model**: `qwen/qwen3.6-flash` doesn't appear to be a valid OpenRouter model ID. The correct Qwen3 IDs on OpenRouter are `qwen/qwen3-8b`, `qwen/qwen3-30b-a3b`, `qwen/qwen3-235b-a22b`, etc. Could you double-check the default against https://openrouter.ai/models and pick one that actually resolves?

2. **Zero cost**: The 0.0 cost for free-tier models is correct behavior, but it should be documented in the docstring/comment so it doesn't look like a bug to future readers. Something like: `# Free-tier models return 0.0 cost — this is correct`.

3. **Misleading comment**: There's a comment that implies something about the request structure that doesn't match what the code actually does - worth cleaning up for clarity.

Once those are addressed this is ready to go!

the-cybersapien:
Hey @safishamsi ,

I've made some fixes. Responses point by point:
1. I am actively using `qwen/qwen3.6-flash` It's available here: [Openrouter link](https://openrouter.ai/qwen/qwen3.6-flash).
2. Updated cost to the default (qwen3.6-flash). 
3. Updated the comment now, I hope it's more clear!

safishamsi:
Good idea — OpenRouter slots cleanly into the existing `_call_openai_compat` pattern. Three things to fix:

1. **Fabricated model name**: `qwen/qwen3.6-flash` does not exist on OpenRouter. Pick a real, currently available model — `qwen/qwen-2.5-coder-32b-instruct` or `meta-llama/llama-3.3-70b-instruct` are solid cheap defaults. Check [openrouter.ai/models](https://openrouter.ai/models) and verify it's live before submitting.

2. **No tests**: Add at minimum a config-dict shape test — env key resolution (`OPENROUTER_API_KEY`), default model, and `GRAPHIFY_OPENROUTER_MODEL` override. See how `test_backends.py` handles the other backends.

3. **Trailing whitespace**: The three comment lines in the config dict end with trailing spaces — clean those up.

Fix those three and this is a clean merge.

garvae:
@safishamsi just leaving a 5 cents from me: this PR seems to add some needed improvements to the project, and it would be great to see this shipped to production =)

Also, thank you for continuing to develop and maintain Graphify - and thanks @the-cybersapien for putting this update.

PR #1044, opened 2026-05-27
feat: ReScript (.res, .resi) extraction

bjornj12:
> This contribution was AI-authored end-to-end (Claude Opus 4.7), reviewed and tested by a human (bjornj12) before submission. Happy to revise anything that doesn't match the style of the repo.

## Summary

Adds AST extraction for ReScript so `.res` and `.resi` files emit Function / Variable / Module / Type nodes and `imports` / `calls` / `references_type` edges. Today graphify's pipeline silently drops them — on a working ReScript codebase with over 100 files the current `v8` build produces 15 nodes total, all from `README.md`, with zero ReScript symbols. After this PR every `.res` / `.resi` file produces a file node plus a node per top-level entity, with `imports` / `calls` / `references_type` edges resolving cross-file when both endpoints are in the scan.

ReScript shops gain code-graph parity with TS/Java/Scala.

### Vendored binding (`graphify/_vendor/tree_sitter_rescript/` + `setup.py`)
No `tree-sitter-rescript` exists on PyPI. Upstream `rescript-lang/tree-sitter-rescript@v6.0.0` ships a complete Python binding (`bindings/python/.../binding.c` + `setup.py`) but it isn't published. Vendored the binding + grammar + LICENSE; `setup.py` compiles the C extension at install time with upstream's exact compile flags. `graphify/__init__.py` aliases the vendored package into `sys.modules["tree_sitter_rescript"]` so `_extract_generic`'s `importlib.import_module(...)` resolves with no per-language detour.

### `_RESCRIPT_CONFIG` + walker hooks (`graphify/extract.py`)
Follows the `LanguageConfig` pattern (Lua/Swift/Scala were the closest templates). `_rescript_extra_walk` (mirrors `_js_extra_walk` and `_csharp_extra_walk`) owns every entity-declaration form:
- `let_declaration` → Function or Variable based on whether `let_binding[body]` is a `function` node, or whether the type annotation is `function_type` (`.resi` signature). Tuple/record destructure emits one Variable per bound name.
- `module_declaration` → Module node; nested modules recurse into the definition block via `walk_fn` with `parent_class_nid` threaded through.
- `type_declaration` → Type node.
- `external_declaration` → Function (callable) when the annotation is `function_type`, Variable otherwise.
- **Function-locals are NOT registered.** Same convention as Python's `function_definition` (returns without descending) and JS's `_js_extra_walk` (module-scope only). Architecture view stays at module surface.

Call dispatch handles `f(x)` (`value_identifier`), `Foo.bar(x)` / `Belt.Array.some(...)` (`value_identifier_path` → last `value_identifier`), and pipe expressions (`arr->Belt.Array.some(...)`) picked up by generic recursion.

### `references_type` edges
Emitted when a `type_identifier_path` appears in a record field, variant/polyvariant arm payload, function signature, let-binding annotation, or external. Targets the leftmost module of nested paths (`Animal.Habitat.species` → `Animal`). Bare local types (`option`, `int`) parse as `type_identifier` not `type_identifier_path` so they emit no edges. Same-file self-references EXTRACTED; cross-file INFERRED until the multi-file resolver rewrites them.

Currently ReScript-only — Java/TS/Scala stay call-edge-only. ReScript codebases are type-driven enough that call edges alone produce a misleading dependency graph. Happy to follow up with the same treatment for the others in a separate PR.

### Cross-file resolver in `extract()`
Two `.res`/`.resi` passes: `open Foo`/`include Foo` import targets rewritten from bare module name to real file node id; `references_type` targets rewritten from `<module>.<type>` to real type-node ids when both sides are in scan. `build.py`'s node-existence filter drops the rest (third-party `Belt.list`/`React.element`/etc.) — same as Swift's `import UIKit` against a non-Apple-SDK scan.

### `graphify/detect.py`
Adds `.res`, `.resi` to `CODE_EXTENSIONS` next to `.swift`.

### Tests + fixture
Single canonical fixture `tests/fixtures/sample.res` (flat, matches every other language) covering every documented surface. 34 new ReScript tests under a `# ── ReScript ──` section in `tests/test_languages.py`, modelled on Scala. Cross-file scenarios use `tmp_path` inline sources, matching `test_cross_file_call_*` in `tests/test_extract.py`. Two "snapshot" tests assert the EXACT set of node labels and edge triples produced by `sample.res` as a forcing function against unintended drift.

### One shared-infra touch
`graphify/extract.py` (per-file `clean_edges` filter) adds `"references_type"` to the survives-unresolved tuple alongside `imports`/`imports_from`/`re_exports`. Same lifecycle. No other language emits `references_type` so the change is functionally ReScript-only.

## Why

Today graphify's pipeline silently drops `.res` / `.resi` files: on a working ReScript codebase with over 100 files the current `v8` build produces only 15 nodes — all from `README.md`, none from any ReScript source. After this PR every `.res` / `.resi` file produces a file node plus a node per top-level entity (types, modules, values, functions, externals, including nested-module members), with `imports` / `calls` / `references_type` edges resolving cross-file when both endpoints are in the scan.

Bundled fixture (`tests/fixtures/sample.res`): 23 nodes / 28 edges from a single file exercising the polyvariant + module + nested-let + function + method + type-reference surface — the qualitative shape any ReScript file produces.

## Test plan

- [x] `pytest tests/test_languages.py -k rescript -v` → 34/34 pass
- [x] `pytest tests/` → no new regressions vs. v8 baseline
- [x] `graphify update tests/fixtures/sample.res` → produces the snapshot-tested node/edge set
- [x] Full `graphify update` on a real-world ReScript codebase with over 100 files → file + symbol nodes per file, cross-file edges resolved as expected. Before this PR the same corpus produced 15 nodes total, all from `README.md`.
- [x] Fresh `pip install -e .` on macOS Apple Silicon (Python 3.14) compiles the vendored C extension successfully

## Risk

- **Adds a C extension to the build.** First one for graphify. Users without a C toolchain installing from sdist will fail; cibuildwheel would fix that — happy to follow up.
- **No behavior changes for any other language.** Only shared-infra touch is the one-line tuple addition in the per-file edge filter.
- **`references_type` is ReScript-only.** No cross-language interference.
- **Function-locals stay invisible** — matches Python and JS, no surprises for users coming from those.


safishamsi:
Impressive work — the extractor is thorough (pipe expressions, nested modules, `.resi` signatures, `references_type` edges) and PROGRESS.md gives a clear picture of the tradeoffs. Three blocking items before merge:

**Vendored C extension needs a policy decision**

Graphify has never vendored a compiled C extension. This adds a `setup.py`, a new `graphify._vendor` package, and a `sys.modules` alias in `graphify/__init__.py` — all of which affect every `import graphify` regardless of ReScript. The stated reason (no PyPI release) is valid, but the impact on the build pipeline is significant: sdist users now need a C toolchain, and the cibuildwheel CI to produce pre-built wheels for PyPI doesn't exist yet. Before this lands we need at least: (a) cibuildwheel CI added to the repo so the PyPI release actually ships wheels, and (b) confirmation that the `__init__.py` alias is safe in edge cases (partial installs, editable mode, multiple Python versions). If that's too much scope for this PR, an alternative is to ship the extractor as a soft-optional import (same as `tree-sitter-nix` above) with a clear error message pointing users to build the extension manually.

**`references_type` is a new edge relation type**

The existing edge schema has `relation` values like `calls`, `imports`, `defines`, `contains`. Adding `references_type` is a schema change that affects `build.py` (cross-language grouping), `serve.py` (MCP relation filtering), `callflow_html.py` (edge rendering), and any downstream tooling. Please: (1) add `references_type` to whatever relation enum/docs exist, (2) check that `build_from_json` and `callflow_html.py` handle the new relation without crashing (even if just a fallthrough), and (3) add at least one test in `test_serve.py` or `test_build.py` that exercises a graph containing `references_type` edges.

**Remove `PROGRESS.md` from the PR**

This file is a contributor's working notes, not repo documentation. It belongs in the PR description, not committed to the tree. Please drop it from the branch.

The extractor code itself is high quality and the decision to skip function-locals matches the Python/JS convention. Fixing the three items above will get this over the line.

bjornj12:
Thanks for the careful read — all three landed. Pushed 3 commits in response: 7e90a6a, ef83aef, 2621ad6.

**Vendored binding → soft-optional extra.** Fair call, this is the first C extension graphify would ship and it deserved its own conversation rather than riding in on a language PR. Went with your option (b).

Removed `setup.py`, the vendored source under `graphify/_vendor/`, the `sys.modules` alias in `__init__.py`, the `_vendor` packages from `[tool.setuptools]`, and `wheel` from `build-system.requires`. `pip install graphifyy` is back to pure-Python — no toolchain required.

Opt-in path:

```toml
rescript = ["tree-sitter-rescript @ git+https://github.com/rescript-lang/tree-sitter-rescript.git@v6.0.0"]
```

`[all]` deliberately not updated — didn't want it silently requiring a C toolchain. Happy to flip that if you'd rather.

Someone running `graphify update` on a `.res` file without the extra now gets a clear next step instead of a bare "not installed". I tucked a ReScript-specific hint into the `ImportError` branch in `_extract_generic`:

```
tree_sitter_rescript not installed. ReScript is opt-in (no PyPI release).
Install via: pip install "graphifyy[rescript]"
```

Tests go through `pytest.importorskip("tree_sitter_rescript")` via a helper that mirrors the existing `_extract_sql_or_skip` pattern in `test_multilang.py`. Without the extra installed, 34 of the rescript tests skip cleanly; with it they pass.

**`references_type` schema.** Yep — adding a new relation type without walking the schema was a miss on my side. Wired it into:

- `DEFAULT_AFFECTED_RELATIONS` in `affected.py` — `--affected` follows type refs by default now, which felt right for typed codebases. Easy to drop if you'd rather make it opt-in.
- `SEMANTIC_RELATIONS` in `extract.py`.
- `relation_label` in `callflow_html.py`, both en and zh — without this the diagram label was falling through to the auto `replace("_", " ")` path. Works, but accidental.
- `docs/how-it-works.md` relation example list.

For build coverage, added `test_build_from_json_preserves_references_type_relation` — round-trip test. Small in scope but it locks the current behaviour against accidental drops if relation-name filtering ever gets added to `build_from_json`.

I looked at `preferred_edges` and `edge_score` while I was in there — `references_type` falls through to the base scoring path without an explicit case, no crash. That felt like the right default for a non-call non-import relation, but happy to add explicit entries if you'd rather see it grouped with `contains` / `rationale_for`.

**PROGRESS.md** — removed. You're right, that was working notes; the engineering write-up lives in the PR description where it belongs.

---

Branch is at 10 commits now. Happy to squash to something tighter if you'd prefer a single feat commit + the review-response cleanup as a second.

PR #1062, opened 2026-05-28
Fix hollow-response loop in claude-cli backend

### LINKED ISSUE #1: v3: semantic query with embeddings

### LINKED ISSUE #2: v3: Codex and OpenClaw support
## Goal

## Approach

## Work needed

christophepub:
# Fix hollow-response loop in `claude-cli` backend

## Problem

When running `graphify extract <path> --backend claude-cli` against a multi-modal corpus (code + Markdown docs), ~30-50% of semantic chunks come back as "hollow responses" and trigger adaptive bisection. On a ~800-file repo this turned a ~15 min run into ~44 min and consumed 2-3× more `claude` subscription quota than necessary.

Symptoms in the log:
```
[graphify] LLM returned invalid JSON, skipping chunk: Expecting value: line 1 column 1 (char 0)
[graphify] chunk of 39 truncated at depth 0, splitting into halves of 19 and 20
```

The hollow-detection path works as designed — the issue is what causes Claude to return content that fails `json.loads`.

### 1. `_parse_llm_json` only strips markdown fences at offset 0

```python
if raw.startswith("```"):
    raw = raw.split("```", 2)[1]
    ...
```

Claude (and most chat models) frequently prepend a short preamble before the JSON, e.g.:

```
Here are the extracted entities:

```json
{"nodes": [...], "edges": [...]}
```
```

`raw.startswith("```")` returns `False`, the fence-stripping is skipped entirely, `json.loads("Here are the extracted entities:\n\n```json\n...")` fails, the chunk is dropped, the hollow detector re-routes it to the bisection path. Each bisected half is another `claude -p` call that may also come back with a preamble. Cost compounds.

### 2. `_call_claude_cli` uses `--append-system-prompt`

`--append-system-prompt` adds graphify's extraction prompt **on top of** Claude Code's default interactive-agent system prompt, which contains instructions like *"output text to communicate with the user"*, *"use markdown formatting"*, etc. These conflict with graphify's "return raw JSON only" instruction, and the default prompt wins about half the time — producing the preambles and markdown fences from issue #1.

The `claude` CLI exposes `--system-prompt` (replace) since at least 2.1.x, which is the right primitive for a headless extraction backend.

## Fix

Two complementary changes in `graphify/llm.py`:

**1. Robust JSON extraction in `_parse_llm_json`** — strips fences regardless of position, with a fallback that scans for the first balanced `{...}` object in the response. Handles preambles, trailing prose, and prose-wrapped JSON without fences.

**2. Switch `claude-cli` to `--system-prompt`** — eliminates the conflict at the source. Claude receives only graphify's extraction prompt and returns clean JSON on the first call. As a side benefit, cache-creation tokens per call drop ~19% (47k vs 58k in my measurements) because Claude Code's default system prompt is no longer materialized.

The two fixes are complementary: fix #2 dramatically reduces the rate of malformed responses; fix #1 keeps graphify robust against the residual cases (soft refusals, model confusion) and benefits every other backend too.

## Evidence

Test run on a 43-file `modes/` directory (Markdown docs):

| Metric | Before | After |
|---|---|---|
| Hollow responses | ~30-50% of chunks | **0** |
| Bisections triggered | several per run | **0** |
| Output tokens | inflated by preambles | clean |
| Cache-creation tokens / call | ~58k | ~47k (-19%) |
| Wall time on full ~800-file repo | ~44 min | ~15 min (estimated) |

Both fixes verified with isolated unit tests covering:
- preamble + fence (the primary bug)
- prose-wrapped JSON without fence
- raw JSON (regression check)
- model soft refusal (graceful empty return with diagnostic log)

## Trade-offs / risks

- `--system-prompt` replaces Claude Code's default prompt entirely. For the `-p` headless extraction use case this is desirable. Subscription auth is unaffected (verified via `is_error: False` test call).
- The balanced-brace scanner in strategy 2 is O(n) and only runs if both `json.loads(stripped)` and fence-stripping fail — no perf impact on the common path.
- `--no-session-persistence` is unchanged. A follow-up could explore session reuse to reclaim more cache budget, but that's orthogonal.

## Diff

Total: ~50 added lines, ~6 removed, in a single file (`graphify/llm.py`). Patch attached.


safishamsi:
Thanks for tracking down the claude-cli hollow-response issue — this is a real problem worth fixing. Unfortunately the PR is unreviewable in its current state: the diff shows 138,000+ additions because the branch is based off a stale `main` rather than `v8`. The actual fix is buried under a full re-fork of the repo.

Please:
1. Re-target the branch to `v8` (`git rebase v8` or open a fresh branch from `v8`)
2. Drop all unrelated files — the PR should only touch `graphify/llm.py`
3. Add the unit tests described in the PR body (preamble/fence handling, the four test cases)
4. Re-open when the diff is clean

Happy to merge once we can see what's actually changing.

christophepub:
Superseded by #1063 — re-targeted to v8 with clean diff and unit tests as requested. Apologies for the noisy first attempt.

PR #1065, opened 2026-05-28
fix: normalize file-level node IDs to stem-based format (fixes #1033)

### LINKED ISSUE #1033: AST and semantic extractors generate mismatched node IDs for the same file, splitting it into two disconnected nodes

adityachaudhary99:
## Problem
AST extractor file-level node IDs used _make_id(str(path)) (e.g., script_pipeline_step_py with extension). The semantic subagent uses {parent_dir}_{filename_without_ext} (e.g., script_pipeline_step). This mismatch caused the same file to appear as two unconnected nodes in the knowledge graph.

## Fix
- Added _file_node_id(path) helper -> _make_id(_file_stem(path)) matching subagent format
- Replaced all ~52 _make_id(str(path)) file-level ID calls across extractors, JS resolution, import targets, and extract() remap
- Changed extract() legacy ID remap fallback from rel = path to continue — files outside root are silently skipped instead of using stale absolute paths
- Added missing imports for extract_svelte and _resolve_js_module_path in test file

## Scope
Affected extractors: JS/TS, Python, Go, Rust, Zig, Fortran, Julia, Pascal, Bash, Astro, Svelte, Objective-C, SQL, C/C++/C#, Java, Kotlin, Scala, PHP, Groovy, Markdown, plus import resolution and Pascal unit maps.


safishamsi:
Thanks for tackling this @adityachaudhary99. I went with a different approach for #1033 and am going to close this one - want to explain why so it's useful feedback.

This PR converts the per-language extractor sites individually (`_make_id(str(path))` -> stem-based). The problem is that file IDs aren't only node IDs - they're also import/dependency **edge targets** in many resolvers (Python relative imports, the JS/TS resolver, C includes, Lua `require`, bash `source`). Converting the node format without converting every edge-target construction site in lockstep just relocates the orphan from "AST vs semantic" to "file node vs its own import edges."

Two concrete gaps in this PR:
- `_resolve_lua_import_target` (`extract.py:2025/2029`) still returns the old-format ID, so Lua `require` edges would point at a phantom after the file node moved.
- `symbol_resolution.py:_file_node_id_for_path` is untouched, so bash `source` edges orphan.

The fix I shipped (`c898dc6`) instead changes the single **remap chokepoint** in `extract()` - every file node ID and every edge endpoint already funnels through the `#502` relative-path remap, so converting that one pass moves the node and all referencing edges together, atomically, across all languages. No per-site churn, no orphan risk.

Appreciate the effort here - the diagnosis was right, it was the surface area of the fix that was the tricky part.

Issue #1075, opened 2026-05-29
Lua require() dependencies do not appear as file-level edges in the final graph


call-me-sean:
### Summary

It looks like Lua `require()` dependencies are not consistently preserved as file-level edges in the final graph.

Because of this, Lua files that are referenced via `require()` may not show the expected inbound dependency relationships.

### Expected behavior

When Lua source contains imports such as:

```lua
local mod = require("foo.bar")
require("foo.baz")
```

the final graph should include a file-level dependency edge between the importing file and the imported Lua file.

### Actual behavior

The final graph appears to keep other relationships such as structural containment and call edges, but Lua file-to-file dependency edges from `require()` are missing.

### Possible cause

From a quick look at the current Lua extraction path, this may be related to how `require()` targets are extracted and resolved before graph construction.

In particular, it seems possible that:

- only some `require()` forms are detected,
- and/or the extracted target does not always match the corresponding file node in the graph.

If so, the dependency edge may be dropped during graph build.

### Minimal example

`a.lua`
```lua
local b = require("pkg.b")
return b
```

`pkg/b.lua`
```lua
local M = {}
return M
```

Expected result:
- a file-level dependency edge from `a.lua` to `pkg/b.lua`

Actual result:
- the file-level dependency edge is missing

### Suggested direction

It may help to:

- support more Lua `require()` forms,
- resolve Lua module paths to actual project file nodes,
- and preserve those relationships as file-level edges in the final graph.

Thanks for taking a look.

call-me-sean:
Opened a fix PR for this:
- PR #1076: https://github.com/safishamsi/graphify/pull/1076

It includes the Lua `require()` file-level edge fix, follow-up path consistency fixes, and targeted validation notes.

safishamsi:
Fixed in 5642c1b — `_import_lua` was taking only the last dot-segment of the module path and using the raw string as the edge target, which never matched the file node ID. Replaced with `_resolve_lua_import_target` that converts dots to path separators, probes the filesystem (up 6 parent dirs, tries `.lua`/`.luau`/`init.lua` variants), and returns the correct `_make_id(str(resolved_path))`. Graceful fallback on unresolved imports.

call-me-sean:
Thanks for the quick fix, and for linking it here.

I re-tested this on our Lua corpus and on a smaller repro. The core issue in #1075 does look fixed: require("pkg.mod") is now resolved to a real file node instead of using only the last dot segment.

[...]

PR #1076, opened 2026-05-29
Fix Lua require file-level edge resolution

call-me-sean:
## Summary
- add a Lua-specific post-pass that resolves `require()` module strings to file nodes and emits `imports_from` edges
- support bare, nested, assigned, and chained `require()` forms by collecting calls from the Lua AST
- add focused tests covering Lua file-level import edge extraction

## Why
Lua `require()` dependencies were being detected inconsistently at the file graph level, which meant inbound file edges were missing from `graph.json` and downstream consumers could not reliably answer "who imports this file?"

## Validation
- `python3 -m py_compile graphify/extract.py tests/test_extract.py`
- targeted Lua validation script covering assigned, bare, nested, and chained `require()` forms
- verified on a real Lua corpus that the patched extractor emits file-level `imports_from` edges

## Notes
I did not run the full pytest suite in this environment because `pytest` is not installed here.

Refs #1075

### DIFF
--- a/graphify/extract.py
+++ b/graphify/extract.py
@@ -1219,6 +1219,99 @@ def _resolve_js_import_target(raw: str, str_path: str) -> "tuple[str, Path | Non
     return _make_id(module_name), None
 
 
+def _file_node_id(path: Path, root: Path | None = None) -> str:
+    """Return the graph node id for a file path, relative to root when possible."""
+    if root is not None:
+        try:
+            return _make_id(str(path.relative_to(root)))
+        except ValueError:
+            pass
+    return _make_id(str(path))
+
+
+def _iter_lua_requires(path: Path) -> list[tuple[str, int]]:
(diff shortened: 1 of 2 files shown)


call-me-sean:
Follow-up update after local review:

- preserve logical corpus paths during Lua `require()` resolution instead of eagerly using fully-resolved paths
- keep Lua post-pass `source_file` values consistent with the rest of extraction (`str(path)`)
- add a symlink-focused regression check so file-level `imports_from` edges still attach to the correct file nodes

Validation I ran locally:
- `python3 -m py_compile graphify/extract.py tests/test_extract.py`
- targeted Lua file-edge validation: OK
- targeted Lua symlink validation: OK
- real-corpus recheck on NearbyFeed: `imports_from = 775`

I did not run the full pytest suite here because `pytest` is not installed in this environment.

PR #1101, opened 2026-06-01
add inheritance edges to the dart extractor

Telomelonia:
## What

`extract_dart` only emitted `defines` (file→symbol) and `imports` (file→package) edges, so Dart graphs had **no symbol-to-symbol structure** — every class/mixin was a leaf hanging off its file. This parses the `extends` / `with` / `implements` clauses of class and mixin declarations and emits the corresponding edges, reusing relation names already in `SEMANTIC_RELATIONS`:

| Dart | edge relation |
|------|---------------|
| `extends Base` | `inherits` |
| `with Mixin` | `mixes_in` |
| `implements Iface` | `implements` |

Targets are the bare base-type name (cross-file class resolution is out of scope for the regex extractor, the same way `imports` targets a package by its raw string).

## Also

Fixed the docstring, which claimed it extracted **"calls"** — it never did.

## Design

- Clause parsing lives in a small `_dart_inheritance(header)` generator that yields `(relation, base_type)` pairs, so the call site stays a readable two-liner.
- Generic arguments are removed first by `_strip_generics`, which depth-counts `<...>` so nested generics like `BaseMapper<GFooItem, FooModelItem>` and `Node<Tree<int>>` collapse to the bare base type instead of leaving stray fragments — a plain `re.sub` can't track nesting.
- Node/edge creation uses local `add_node` / `add_edge` closures, matching the pattern the other extractors in this file already use.

## Verification

On the `material_3_demo` app from [`flutter/samples`](https://github.com/flutter/samples) (23 Dart files), inheritance edges originating from `.dart` files:

| | inheritance edges from Dart | malformed targets |
|---|---|---|
| before | 0 | 0 |
| after | 110 | 0 |

(The few inheritance edges the baseline already had came from non-Dart files via the existing extractors, not Dart.)

## Tests

Three new tests in `tests/test_extract.py` (synthetic sources, matching the existing Dart test style): basic `extends`/`with`/`implements`, multiple bases + generics, and a nested/multi-arg generic regression guard. Full `tests/test_extract.py` passes (79 tests).

### DIFF
--- a/graphify/extract.py
+++ b/graphify/extract.py
@@ -4089,8 +4089,51 @@ def extract_blade(path: Path) -> dict:
     return {"nodes": nodes, "edges": edges}
 
 
+def _strip_generics(text: str) -> str:
+    """Drop balanced `<...>` generic spans, including nested ones, so that
+    `Foo<A, Bar<B, C>>` collapses to `Foo`. A plain regex can't track nesting."""
+    out, depth = [], 0
+    for ch in text:
+        if ch == "<":
+            depth += 1
+        elif ch == ">":
+            depth = max(0, depth - 1)
+        elif depth == 0:
+            out.append(ch)
+    return "".join(out)
+
+
+# Dart class header, e.g. `Dog extends Animal with A, B implements C`. Clauses
+# are optional and always appear in this order; each captures a comma-separated
+# type list that `_dart_inheritance` splits.
+_DART_HEADER_RE = re.compile(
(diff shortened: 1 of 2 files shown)


Telomelonia:
@safishamsi 
Hey! Small PR here — adds Dart inheritance edges to the extractor, tests green locally. Mind kicking off CI / taking a look when you have a sec? 🙏

Overman775:
@Telomelonia same pr ;) https://github.com/safishamsi/graphify/pull/1098

Telomelonia:
@Overman775 Thanks a lot! this pr can be closed now

PR #1103, opened 2026-06-01
feat: Add live PostgreSQL introspection connector (resolves #1093)

### LINKED ISSUE #1093: Feature: connect directly to a live PostgreSQL instance (--postgres DSN) instead of dumping schema to .sql first
## Summary

Add an optional live-database source so users can point graphify at a running PostgreSQL instance (via a connection string / DSN) and have it auto-extract the schema, instead of manually running `pg_dump --schema-only` first and feeding the resulting `.sql` file.

## Motivation

Graphify already maps database schemas well: `extract_sql()` (extract.py:4251) parses tables, views, functions, and foreign-key relationships from `.sql` files via tree-sitter, and this is exactly what the v0.6.0 "map your entire backend in one graph" feature is about.

The friction is the manual dump step. Today the workflow is:

```bash
pg_dump --schema-only mydb > schema.sql
graphify extract ./schema.sql
```

For users who just want a graph of their live database, the dump-to-file round trip is an unnecessary step, and it goes stale the moment the schema changes. A direct connector would let them do:

```bash
graphify extract --postgres "postgresql://user:pass@host:5432/mydb"
```

and get a fresh graph of the current schema on demand.

## Proposed design

This should be a thin wrapper around the **existing** extraction path, not a new extractor:

1. New optional dependency group, e.g. `postgres = ["psycopg[binary]"]`.
2. A small introspection helper that connects, reads the schema from the catalog (`information_schema` / `pg_catalog`), and emits the same node/edge shape `extract_sql()` already produces - tables, views, functions, and FK `references` edges - so everything downstream (build, cluster, report, export) is unchanged.
3. Wire it as a `--postgres <dsn>` flag (or a `graphify extract` source variant) that feeds the introspected schema into the normal pipeline.

Reusing the `extract_sql` node/edge contract keeps this isolated: no changes to build/cluster/export, and the graph looks identical whether the schema came from a `.sql` dump or a live connection.

## Scope / non-goals

- **Schema only**, not data. Same as the current `.sql` path - this maps structure (tables, columns, FKs, views, functions), not rows.
- Read-only. The connector only introspects the catalog; it never writes.
- Postgres first. The same approach generalizes to other engines later (MySQL, etc.) but this issue is scoped to Postgres.
- Credentials handling: support a DSN string and standard `PG*` env vars (`PGHOST`, `PGUSER`, ...); document not committing secrets.

## Alternatives

- Keep the status quo (`pg_dump` then `graphify extract`). Works today, but stale and manual.

## Context

Came up from a user asking whether graphify could point directly at PostgreSQL (and VictoriaLogs). The `.sql`-file path covers the Postgres-schema case once dumped; this issue is about removing the dump step. (Live log/time-series ingestion like VictoriaLogs is a separate, out-of-scope domain.)

geektan123:
This PR implements direct PostgreSQL database introspection support via `graphify extract --postgres "postgresql://..."`, closing #1093.

### What it does:
- Adds a `--postgres` CLI argument (DSN string or PG* env vars) to extract the schema structure (tables, views, routines, foreign key relations) directly from a running PG instance.
- Avoids the manual `pg_dump` round-trip.
- Reuses `extract_sql` by converting catalog queries into standard SQL DDL strings in memory, meaning the downstream mapping behaves identically.
- Correctly skips LLM API requirements when only introspecting a database schema.
- Added comprehensive mocking unit tests (`tests/test_pg_introspect.py`) that run without needing a live PG instance.
- Fully documented in `README.md` and `ARCHITECTURE.md`.


safishamsi (changes requested):
Thanks for this @geektan123 - really nice first cut, and exactly the thin-wrapper approach #1093 was after: reconstruct DDL from the catalog and hand it to `extract_sql` so the whole downstream pipeline stays unchanged. I traced it end-to-end and the core works - FK edges resolve correctly (by table name, so the `(id INT)` stub is harmless), credentials are correctly kept out of `source_file`, the catalog queries are static/non-injectable, and the tests genuinely exercise the reconstruction → real tree-sitter parse rather than over-mocking. 

A few things to address before merge - two are silent-correctness bugs that the current tests don't catch (all fixtures use clean lowercase names + a single-column FK, so these paths are invisible):

## Must-fix

**1. Quote identifiers (silent table loss).** Schema/table/column names are interpolated raw into the DDL:
```python
ddl.append(f"CREATE TABLE {schema}.{name} (id INT);")
```
A table named `order` (reserved word), `user-data` (hyphen), `"MixedCase"` (case-sensitive), or anything with a space/`;`/quote produces DDL that tree-sitter-sql misparses - and the table (plus any FK touching it) is **silently dropped**, no error. Postgres allows almost any character in a quoted identifier, so this will bite real schemas.
- Fix: emit double-quoted identifiers, e.g. `"{schema}"."{name}"`, doubling any embedded `"`. Please double-check that `table_nids` keying in `extract_sql` stays consistent once names are quoted (the `.lower()` keying may need to account for the quotes).

**2. Composite (multi-column) foreign keys over-count.** The FK query returns one row per key column, and the code emits a separate single-column `ADD CONSTRAINT fk_{schema}_{table}_{col}` per row. A 2-column composite FK becomes two constraints → two duplicate `references` edges between the same table pair (there's no edge dedup), inflating edge counts, and the per-column constraint naming is semantically wrong.
- Fix: `GROUP BY` the constraint and aggregate columns into one `FOREIGN KEY (a, b) REFERENCES t (c, d)`, named by the constraint rather than the column.

**3. Tests for both of the above.** Add a fixture with a reserved-word/quoted/special-char table name (assert its node + FK survive) and a composite (2-column) FK (assert exactly one `references` edge). These are the highest-severity paths and currently untested.

## Should-fix (can be follow-up, but ideally here)

**4. Graceful connection errors.** `psycopg.connect(dsn or "")` throws a raw `OperationalError` (full stack trace) on a bad DSN / unreachable host / auth failure. Please catch it and surface a clean `error: could not connect to PostgreSQL: ...` with a non-zero exit, matching the CLI's existing error style. Be careful the error message doesn't echo the DSN/credentials.

**5. Function-body dollar-quote collision.** `CREATE FUNCTION ... AS $$ {body} $$` breaks if `routine_definition` itself contains `$$` → the function is dropped. Use a unique tag (e.g. `$gfx$ ... $gfx$`). Also `routine_definition` is often SQL-language (or NULL for C/internal functions), so hardcoding `LANGUAGE plpgsql` is sometimes wrong - consider reading `external_language` / `routine_body` from `information_schema.routines`.

## Note (not blocking)
The SQL extractor only produces object-level nodes (tables/views/functions), not column nodes - so columns aren't represented in the graph today. #1093 mentioned "columns", so worth a one-line note in the docs that this maps table/view/function structure + FK relationships, not individual columns.

Overall this is close - the design is right and the mechanism is sound. Fix the two quoting/composite-FK bugs (+ tests) and it's good to merge. Thanks again for picking this up so fast!

geektan123:
All five items from the review are addressed in d8f79c1. Also added the non-blocking docs note about column-level detail not being represented in the graph.

safishamsi:
Landed in 7467c1b.

**What it adds**: `graphify extract --postgres "postgresql://..."` introspects a live PostgreSQL database directly — no `pg_dump` roundtrip needed. Extracts tables, views, routines, and foreign key relations via `information_schema` in a `SERIALIZABLE READ ONLY` transaction. Credentials are sanitized from error messages. New `graphify[postgres]` optional extra (psycopg3).

5 mocked tests pass (no real PostgreSQL instance needed). 1910 passed, 0 failures.

Issue #450, opened 2026-04-19
Per-image chunk rule forces one subagent per image — 59 images ≈ 59 agents, each ~28K tokens for ~5 nodes output


harnill:
## Context
Skill says:

> Split into chunks of 20-25 files each. **Each image gets its own chunk (vision needs separate context).**

On a 410-file corpus with 59 images, this produces 59 image chunks + 16 non-image chunks = **75 total subagents**. Measured token cost per image chunk: ~28-30K total tokens (system + instructions + single image + output), yielding ~4-7 nodes each.

## Why this is wasteful
Vision models handle multiple images per request fine, up to the model's image-count cap. The per-chunk instruction payload (~2K tokens from the shared instructions file + tool-call boilerplate) is paid 59 times instead of ~10. Back-of-envelope: batching 6 images per chunk would cut image-chunk overhead from 59 × 30K = 1.77M tokens to 10 × (2K instructions + 6 × ~3K per image + output) ≈ 210K tokens. ~8× reduction for the image pass.

## Proposed fix
Change the rule to:

> Split non-image files into chunks of 20-25 files each. Split images into chunks of 4-6 images each (shared instruction context; separate image inputs within the chunk).

Plus: document in the subagent prompt that the output JSON must include one `source_file` per image, so multi-image chunks still produce per-image nodes correctly.

## Secondary observation
Image-chunk output was fairly thin across this test (average 5 nodes + 5 edges per image). If that's representative, an even more aggressive batch (8-10 images per chunk) would be fine given context-window headroom.

## Measurement
- 59 image chunks × ~28K tokens each = ~1.65M tokens for images alone
- Output produced: ~310 nodes, ~300 edges, 59 hyperedges
- Per-image avg duration: 33s
- Implied cost per node from the image pass: ~5K tokens/node

A batched version would trade that for ~600 tokens/node — in line with the non-image chunks.

PR #1110, opened 2026-06-02
feat(llm): stop reading images and PDFs as garbage in headless extract

### LINKED ISSUE #1109: Headless graphify extract reads images as garbage text instead of seeing them
> Fix proposed in #1110.

## What happens

Run `graphify extract` on a folder that has images in it and every image comes out broken. With `--backend claude-cli` the run fails outright:

```
[graphify] chunk 1/1 failed: claude -p exited 1:
[graphify] WARNING: 1/1 semantic chunk(s) failed
[graphify extract] error: all semantic chunks failed for backend 'claude-cli'
```

With the API backends (claude, openai, gemini, bedrock, ...) it doesn't crash, but the image nodes are meaningless: the model never actually saw the picture, so the labels are guessed from the filename at best.

## Why

`extract_files_direct` builds the prompt with `_read_files`, which does `path.read_text(encoding="utf-8", errors="replace")` on every semantic file. For a PNG/JPG/WEBP that turns the binary into replacement-character garbage. So:

- claude-cli gets a prompt full of invalid text and the CLI exits 1.
- the API backends get noise and emit a few hollow nodes.

The schema already has a `file_type: "image"` slot, and the extract command explicitly routes images into the semantic pass (`semantic_files = doc_files + paper_files + image_files`), but no backend ever puts the actual image in front of a vision model. Images are classified, then read as if they were text.

## Repro

```
mkdir -p /tmp/imgtest && cp any-logo.png /tmp/imgtest/
graphify extract /tmp/imgtest --backend claude-cli
```

## Expected

Raster images should be sent as real visual input to vision-capable backends, so the graph captures what each image depicts instead of garbage. Backends whose selected model has no vision should degrade to a plain reference node rather than crashing or emitting noise.

Related: #450 (per-image chunking on the skill path) and #181 (images inside PDFs not parsed) are downstream of the same gap on the headless extraction path.

yaanfpv:
## What this does

Headless `graphify extract` feeds every semantic file through `_read_files`, which does `path.read_text(encoding="utf-8", errors="replace")`. That is fine for markdown and code, but for a binary file it produces replacement-character garbage. Two file types hit this:

- **Images** (`.png .jpg .jpeg .gif .webp`) — turned into garbage text. Noise for the API backends, and an outright `exit 1` for claude-cli. The schema already has a `file_type: "image"` slot, but no backend ever put a real image in front of a model.
- **PDFs** (`.pdf`, classified as `paper`) — same thing. graphify already ships `extract_pdf_text` (pypdf), it just was never wired into the headless path.

This makes both contribute real content to the graph.

## How

**Images** are split out from the text files and rendered as actual visual input, the way each backend's API expects:

| backend | how images are sent |
|---|---|
| `claude` | base64 image blocks (Anthropic Messages) |
| `openai` / `gemini` | `image_url` data URIs |
| `kimi` | `image_url` data URIs (k2.6 is multimodal) |
| `bedrock` | Converse image blocks (raw bytes) |
| `claude-cli` | image paths opened with the CLI's Read tool (`--add-dir`) |

A per-backend `vision` capability flag gates the pixel path. Backends whose default model has no vision (DeepSeek, and Ollama unless you set `GRAPHIFY_OLLAMA_VISION=1` with a vision model) get a short text reference instead, so the image still becomes a graph node rather than breaking the run. `.svg` stays on the text side and is read as XML. Images pack at a flat token estimate and are capped at 20 per chunk, so a corpus with hundreds of images splits into right-sized requests instead of one oversized one.

**PDFs** are routed through the existing `extract_pdf_text` (pypdf) instead of `read_text`, so a PDF contributes its real text. A scanned PDF with no text layer extracts to empty and becomes a reference node rather than noise. (No new dependency; pypdf is already required.)

## Testing

`tests/test_image_vision.py` mocks every backend (no API keys, no network, no `claude` binary) and asserts the image data actually lands in each payload: the Anthropic base64 block, the OpenAI data URI, the Bedrock raw bytes, and the claude-cli `--add-dir` + Read instruction. It also covers the non-vision text-note fallback, the dispatch-level gating, the images-per-chunk cap, oversized/unreadable images degrading to a reference node, SVG staying text, and PDFs routing through pypdf rather than `read_text`.

- Full suite: `1539 passed, 24 skipped`. `ruff check` clean on the changed files.
- Verified end to end against real images with claude-cli: it produces image nodes by reading each image with its Read tool (a faster model such as `GRAPHIFY_CLAUDE_CLI_MODEL=haiku` keeps the agentic read-loop quick), and on a real corpus it picked up detail that is not in the filenames, so the model genuinely read the pixels.

No new dependencies. Backends without images, and text-only corpora, are byte-for-byte unchanged.

## Not in this PR

- **A keyless `codex-cli` backend** (image vision via `codex exec -i`) is split into a focused follow-up PR. It is a new subprocess-spawning backend with its own security surface, so it deserves its own review rather than riding along with this bug fix. The vision feature works across the six backends above without it.
- Video/audio files are not handled by `graphify extract` at all today (they are not in the semantic pass), so giving them a real input is a separate, larger change (audio transcription, or sampling frames for vision) rather than a one-line bug fix. Tracking that separately.

## Related

Closes #1109.

#450 (per-image chunking forces one subagent per image) and #181 (images inside PDFs not parsed) both stem from images never reaching a vision model on the headless path. This addresses the root cause for the `graphify extract` backends. Image-only / scanned PDFs would still benefit from page rasterization to vision, which could build on this.


safishamsi (changes requested):
This is genuinely well-engineered @yaanfpv - thanks for the thorough writeup and tests. I traced it against the code and the core is solid: text-only corpora are byte-identical (`test_no_images_is_byte_identical` confirms it), the per-backend payload shapes are correct (Anthropic base64 with real `media_type`, OpenAI/Kimi/Gemini data URIs, Bedrock Converse bytes), claude-cli's `--add-dir` scopes to each image's parent dir rather than `/` or the repo root, PDFs degrade safely to a reference node, and the codex sandbox flags (`--ephemeral --skip-git-repo-check --sandbox read-only`) are genuinely read-only. Nice work.

Two things before merge:

## Please split `codex-cli` into its own PR

The image/PDF bug fix and the new `codex-cli` backend are really two changes. `codex-cli` is a **new keyless subprocess backend** with its own security surface (spawning `codex exec`), and it's unrelated to the read-as-garbage bug except as one of the vision paths. Splitting it lets the bug fix - which is low-risk and broadly useful - merge quickly, and gives the new backend the focused review a subprocess-spawning backend deserves. The vision feature works across the existing backends (claude, openai, gemini, kimi, bedrock, claude-cli) without it.

## Harden the `-i` image paths (in the codex PR)

The `--` terminator protects the *prompt* but not the `-i` values:
```python
for img in image_paths:
    cli_args.extend(["-i", img])
cli_args.extend(["--", prompt])
```
Paths are likely absolute (so `-foo.png` -> `/.../-foo.png`, mostly safe), but a relative or crafted path beginning with `-` could be misparsed by Codex's arg parser and break the call. Please ensure each `-i` path is absolute (and a regular file under root - you already check `is_file()`), and add an adversarial test: a corpus image literally named `-foo.png` should still be passed safely (or rejected), not treated as a flag. The current `test_codex_cli_passes_image_flags_and_reads_output` doesn't cover this.

## Nice-to-have
- The `_call_llm` codex branch duplicates ~25 lines of the `_call_codex_cli` subprocess/tempfile logic - factor into a shared helper so the read-only flags can't drift in one place.
- CHANGELOG note: routing PDFs through `extract_pdf_text` changes their token footprint, which can shift chunk boundaries for neighboring files (i.e. existing graphs may differ). Intended, but worth calling out.

Net: the image/PDF fix is close to mergeable on its own - please pull `codex-cli` out, and the `-i` hardening + test go with it. Thanks again!

yaanfpv:
@safishamsi  Thanks. And I agree. That split makes total sense. Update in latest push: codex-cli is pulled out, so this PR is just the image/PDF vision fix across claude, openai, gemini, kimi, bedrock, and claude-cli. I also added the CHANGELOG note about PDF text extraction shifting chunk boundaries on re-extract.

I'll send codex-cli as its own PR shortly, with the -i hardening (each -i path resolved to an absolute path and confirmed to be a regular file under the corpus root, plus an adversarial test for an image literally named -foo.png) and a shared subprocess helper so the read-only flags can't drift between the two call sites. Thanks for the careful review.

yaanfpv:
I have already built the  codex-cli backend on a separate branch (with the -i absolute-path hardening, the -foo.png adversarial test, and a shared _run_codex_exec helper for the read-only flags). I'll rebase it onto v8 and open it as its own PR once this merges, so the diff stays codex-only. Cheers

safishamsi:
The `llm.py` side of the codex-cli pull-out is clean (impl + BACKENDS entry gone), but the split isn't quite complete — `__main__.py` and the README still reference the now-removed backend:

- `__main__.py`: the `elif backend == "codex-cli":` dispatch block (the `shutil.which("codex")` keyless gate + error message) and `codex-cli` in the `--backend ...` usage string are still added by this PR.
- `README.md`: still lists "the `codex` CLI binary (OpenAI Codex …)" as a backend.

As-is, `graphify extract --backend codex-cli` would pass the `__main__.py` gate, then fail at `extract_files_direct` with an unknown-backend error since the impl is gone. Could you move those three (the dispatch block, the usage-string entry, the README line) into the codex-cli PR too? Then this one is purely the image/PDF fix and good to merge. Thanks!

yaanfpv:
Done. Removed all three from this PR (the codex-cli dispatch gate, the usage-string entry, and the README backend line), so it is now purely the image and PDF fix. They belong with the codex-cli backend, which I'll open as its own PR once this merges. Thanks for the catch.

yaanfpv:
Anything I can do to help move this one along? The three codex-cli bits are pulled out as of 25c0dce, so it is purely the image and PDF fix now. Looks like the workflows are still waiting on a maintainer approve-to-run, so the checks haven't had a chance to go green yet. Happy to rebase or adjust if anything has drifted since. No rush, just flagging it's ready whenever you have a window.

safishamsi:
Landed in 7467c1b.

**What it fixes**: Headless `graphify extract` was feeding binary image and PDF files through `path.read_text(errors='replace')`, producing replacement-character garbage sent to the LLM.

**Fix**: Images now route through per-backend real vision payloads (base64/data-URI/bytes for claude/openai/bedrock, file path for claude-cli). Non-vision backends get `_strip_pixels` for graceful degradation — image becomes a graph node with a text reference rather than garbage. PDFs reuse the existing pypdf path. 5MB cap, 20-image chunk limit. No new dependencies.

1910 passed, 0 failures.

Issue #792, opened 2026-05-09
[Feature Request] Enhancing Local LLM Performance (Ollama) & High-Core CPU Scaling via graphify.yaml


RANGERVIII:
I’ve been stress-testing `graphify` on a high-end local workstation (Ryzen 9 9950X3D, RTX 5090). While the tool is conceptually brilliant, the current hardcoded limits and environment-variable-only configuration create significant bottlenecks for local-first workflows.

Current Bottlenecks Identified:

1. CPU Underutilization: In `extract.py`, `max_workers` is capped at 8. On a 32-thread CPU, this limits AST extraction performance by 400%.
2. Local API Timeouts: The default `token_budget` (60k) and `max_concurrency` (4) are ideal for Cloud APIs but cause `Request timed out` or VRAM OOMs on local Ollama/llama.cpp instances during semantic extraction.
3. Authentication Friction: The CLI blocks execution if `OLLAMA_API_KEY` is missing, even when targeting a local instance where no auth is required.

Proposed Solution: The `graphify.yaml` Configuration Layer
To support the growing segment of local power users, I suggest introducing an optional configuration file (`graphify.yaml`) to override defaults persistently.

Example `graphify.yaml` Structure:

```yaml
# Performance Tuning
extraction:
  max_workers: 32         # Fully utilize high-end CPUs (9950X3D)
  token_budget: 16384     # Optimized for 24GB+ VRAM (Gemma 4 31b)
  max_concurrency: 1      # Avoid local API queue congestion

# Backend Overrides
backend:
  name: "ollama"
  model: "gemma4:31b"
  base_url: "http://127.0.0.1:11434/v1"
  # Suggested: bypass API_KEY check if backend is Ollama and base_url is loopback

# Model Parameters
parameters:
  temperature: 0.3        # Deterministic graph generation
  seed: 42
  num_ctx: 65536

```

Why this is a "Win-Win":

* User Experience: Removes the need to re-export `set/export` commands in every terminal session.
* Scalability: `graphify` becomes the gold standard for high-performance local knowledge graph extraction.
* Maintainability: Decouples hardware-specific tuning from the core logic.

Verified Performance:
By manually patching these values in my local instance, I achieved **95% GPU saturation on an RTX 5090** and reduced AST extraction time significantly. I’m happy to provide more telemetry or contribute code snippets if this aligns with the project roadmap!

RANGERVIII:
**Addendum: UX, Observability & Hard Timeouts during Local Inference**

[...]

PR #1111, opened 2026-06-02
fix(llm): honour GRAPHIFY_API_TIMEOUT in the claude-cli backend, not just HTTP

### LINKED ISSUE #1112: claude-cli backend ignores GRAPHIFY_API_TIMEOUT and --api-timeout (hardcoded 600s)
Long extraction runs on the `claude-cli` backend get cut off at exactly ten minutes, and there's no way to raise the ceiling. Setting `GRAPHIFY_API_TIMEOUT` or passing `--api-timeout` has no effect on a `claude-cli` run, even though both take effect for the OpenAI-compatible HTTP backends.

The cause is that the timeout only reaches the HTTP client. The two `claude-cli` subprocess calls in `graphify/llm.py` hardcode `timeout=600`:

- `_call_claude_cli` (extraction)
- the `claude-cli` branch of `_call_llm` (community labeling)

Neither reads `GRAPHIFY_API_TIMEOUT`, so a dense chunk or a large 1M-context model that legitimately runs past ten minutes can't be given more headroom without editing the source. The env var was added for the API path in #792 and the README env-var table presents it as general, which makes the CLI gap easy to trip over.

Proposed fix in #1111: factor the parse the HTTP path already does into a small helper and route both `claude-cli` subprocess calls through it, so the same env var and the `--api-timeout` flag apply uniformly across backends. The 600s default is unchanged.

yaanfpv:
Closes #1112.

`GRAPHIFY_API_TIMEOUT`, and the `--api-timeout` flag that sets it, only ever reached the OpenAI-compatible HTTP client. The two `claude-cli` subprocess calls, extraction (`_call_claude_cli`) and community labeling (the `claude-cli` branch of `_call_llm`), hardcoded `timeout=600` and ignored the knob completely.

That bites on long `claude-cli` runs. A dense chunk or a large 1M-context model can legitimately run past ten minutes, and today there is no way to give it more headroom without editing the source. The HTTP backends already advertise the escape hatch (the env-var table documents it, and #792 wired it in); the CLI path just never picked it up.

### What changed
- Pulled the parse the HTTP path was doing inline into a small `_resolve_api_timeout()` helper, placed next to and mirroring the existing `_resolve_max_tokens()`.
- Routed all three call sites through it: the OpenAI client, `_call_claude_cli` (extract), and the `claude-cli` branch of `_call_llm` (labeling).
- Updated the README env-var row so it no longer describes the timeout as HTTP-only.

The HTTP backend behaves exactly as before (same env var, same 600s default, same parsing). The `claude-cli` extraction and labeling subprocesses now respect `GRAPHIFY_API_TIMEOUT`, and since `--api-timeout` simply exports that env, the flag now applies to `claude-cli` as well. Non-positive or unparseable values fall back to 600s, matching the HTTP path.

### Tests
Added to `tests/test_claude_cli_backend.py`: four unit tests for `_resolve_api_timeout` (default, env override, invalid string, non-positive) and two that assert the resolved value reaches `subprocess.run`'s `timeout=` argument for both the default and an env override. The file passes (19 tests) and `ruff check` is clean.

### DIFF
--- a/README.md
+++ b/README.md
@@ -361,7 +361,7 @@ These are only needed for **headless / CI extraction** (`graphify extract`). Whe
 | `AWS_*` / `~/.aws/credentials` | AWS Bedrock — standard credential chain | `--backend bedrock` (no API key, uses IAM) |
 | `GRAPHIFY_MAX_WORKERS` | AST parallelism thread count | optional — also `--max-workers` flag |
 | `GRAPHIFY_MAX_OUTPUT_TOKENS` | Raise output cap for dense corpora | optional — e.g. `32768` for large files |
-| `GRAPHIFY_API_TIMEOUT` | HTTP timeout in seconds (default: 600) | optional — also `--api-timeout` flag |
+| `GRAPHIFY_API_TIMEOUT` | Per-call timeout in seconds for every backend, HTTP and `claude-cli` alike (default: 600) | optional — also `--api-timeout` flag |
 | `GRAPHIFY_FORCE` | Force graph rebuild even with fewer nodes | optional — also `--force` flag |
 | `GRAPHIFY_GOOGLE_WORKSPACE` | Auto-enable Google Workspace export | optional — set to `1` |
 | `GRAPHIFY_TRIAGE_BACKEND` | Backend for `graphify prs --triage` | optional — auto-detected from available keys |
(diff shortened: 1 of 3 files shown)


safishamsi:
Implemented from our end in 75c4de7 with two additions over the PR: (1) the Anthropic SDK path (_call_anthropic) now also honours the timeout, so the README claim 'every backend' is accurate; (2) added a test for the _call_llm claude-cli branch separately from _call_claude_cli. Thanks for the diagnosis — your root cause analysis was exactly right.

PR #1113, opened 2026-06-02
feat: support ANTHROPIC_BASE_URL for custom API endpoints

jkorsvik:
## Problem

The Claude backend hardcodes `base_url` to `https://api.anthropic.com`, making it impossible to route requests through Anthropic-compatible API gateways or proxies.

This breaks usage behind:
- Corporate API gateways (our use case — a centralized gateway that routes to Bedrock)
- LiteLLM / other proxy servers
- Self-hosted Anthropic-compatible endpoints

## Fix

Read `ANTHROPIC_BASE_URL` from the environment and pass it to the `anthropic.Anthropic()` constructor. Falls back to `https://api.anthropic.com` when unset — fully backwards-compatible.

This matches the convention used by the [official Anthropic Python SDK](https://github.com/anthropics/anthropic-sdk-python#usage) itself, which respects this env var.

## Changes

`graphify/llm.py`: Two call sites patched:
1. `_call_claude()` — used for semantic extraction
2. `_call_llm()` claude backend — used for query/cluster

## Testing

Verified with a corporate API gateway (Anthropic Messages API compatible, JWT auth):
```bash
export ANTHROPIC_API_KEY=$JWT_TOKEN
export ANTHROPIC_BASE_URL=https://gateway.example.com
graphify . --update --backend claude --model claude-haiku-4-5-20251001
# → Successfully extracts via gateway
```

Without the env var set, behavior is identical to before (hits api.anthropic.com).

### DIFF
--- a/graphify/llm.py
+++ b/graphify/llm.py
@@ -119,6 +119,20 @@ BACKENDS: dict[str, dict] = {
 }
 
 
+_ANTHROPIC_DEFAULT_BASE_URL = "https://api.anthropic.com"
+
+
+def _anthropic_base_url() -> str:
+    """Return the Anthropic API base URL, respecting ANTHROPIC_BASE_URL env var.
+
+    Falls back to the default (https://api.anthropic.com) when the env var is
+    unset or empty.  This allows routing requests through corporate API gateways,
+    LiteLLM proxies, or other Anthropic-compatible endpoints.
+    """
+    url = os.environ.get("ANTHROPIC_BASE_URL", "").strip()
+    return url if url else _ANTHROPIC_DEFAULT_BASE_URL
+
+
 def _custom_providers_path(global_: bool = True) -> Path:
     if global_:
         return Path.home() / ".graphify" / "providers.json"
@@ -484,7 +498,10 @@ def _call_claude(api_key: str, model: str, user_message: str, max_tokens: int = 
             "Run: pip install anthropic"
         ) from exc
 
-    client = anthropic.Anthropic(api_key=api_key)
+    client = anthropic.Anthropic(
+        api_key=api_key,
+        base_url=_anthropic_base_url(),
+    )
     resp = client.messages.create(
         model=model,
         max_tokens=max_tokens,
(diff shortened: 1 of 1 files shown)


safishamsi:
Thanks for this — the corporate-gateway use case is totally legit, and routing Claude through a proxy is something we want to support.

One thing worth checking before we merge: the `anthropic` SDK already reads `ANTHROPIC_BASE_URL` from the environment on its own when `base_url` isn't passed, so `graphify ... --backend claude` *should* already route through your gateway with just `ANTHROPIC_BASE_URL` set — no code change. Could you confirm whether it works **without** the patch (just the env var)?

- If it works without the patch → the SDK's already handling it, and we can probably skip the change (or keep a trimmed version purely for explicitness/discoverability).
- If it *doesn't* work without the patch → that's a real gap and we'll merge this. Either way, good to know.

Asking because we'd rather not carry code that duplicates SDK behavior unless it's actually closing a gap. Appreciate you digging into this!

jkorsvik:
Closing — the SDK already handles this. No code change needed.

jkorsvik:
**More context on the failure mode** (for anyone else hitting this):

The reason it *appeared* broken was a shell export issue, not a code bug:

```bash
# This does NOT work — vars are set in current shell but NOT exported to child processes:
source .env
graphify . --update --backend claude
# → 'backend claude requires ANTHROPIC_API_KEY to be set'

# This WORKS — vars are exported and inherited by graphify's Python subprocess:
export $(grep -v '^#' .env | xargs)
graphify . --update --backend claude
# → success, routes through gateway
```

The confusing part: `echo $ANTHROPIC_API_KEY` shows the value in the shell, so you *think* it's set. But graphify runs as a subprocess (`~/.local/bin/graphify` → Python), and `source` without `export` doesn't propagate to child processes.

The gateway showed `bad_gateway` in its logs because:
1. The SDK *did* read `ANTHROPIC_BASE_URL` and routed to the gateway ✓
2. But `ANTHROPIC_API_KEY` wasn't in the process env, so graphify's own check (`_get_backend_api_key`) rejected it before ever calling the SDK
3. When we forced past that check by using `export`, it worked immediately

**Suggestion for graphify docs:** A note in the README like:
> ```bash
> # For proxy/gateway setups, export both vars:
> export ANTHROPIC_API_KEY=$YOUR_KEY
> export ANTHROPIC_BASE_URL=https://your-gateway.example.com
> graphify . --update --backend claude
> ```

This would save future users the same debugging loop.

PR #1118, opened 2026-06-02
fix: prune stale AST nodes on full re-extraction (#1116)

### LINKED ISSUE #1116: graphify update leaks stale nodes: symbols removed from surviving files are never pruned
## Summary

`graphify update` (and the post-checkout git hook) leak stale nodes: a **symbol removed from a file that still exists** is never pruned. Its node — and any inbound `calls`/`uses` edge — survives every incremental rebuild until a full clean rebuild (clearing `graph.json` + manifest). Reported in #1115.

## Root cause (confirmed in code)

The CLI `graphify update` (`__main__.py` ~2853) and the post-checkout hook (`hooks.py` ~166) call `watch._rebuild_code(...)` with **no `changed_paths`** → the **full-re-extraction branch**.

In that branch (`watch.py` ~526-545), `evict_sources` is populated **only for files that no longer exist on disk** (`norm not in current_sources`). A surviving file with a removed symbol is never added to the eviction set, so the stale node:
- is absent from the freshly re-extracted `new_ast_ids` (it's gone from the new AST output), **and**
- has a `source_file` that still exists (not in `evict_sources`),

…so it passes the `preserved_nodes` filter and survives. Its inbound edge survives too (both endpoints remain in `all_ids`).

The **post-commit** hook path is fine — it passes `changed_paths`, which adds changed files to the eviction set (`watch.py` ~523-525). So the bug only manifests on the full path (`graphify update` CLI, post-checkout).

## The subtle part — naive fix is unsafe

"Just evict every re-extracted file's nodes" would **silently delete semantic/LLM nodes**: the update pass is **AST-only** (semantic is not re-extracted), and semantic nodes carry a `source_file` pointing at a surviving code file. Blanket-evicting by `source_file` = semantic data loss.

## Safe fix

Evict by **AST-node identity**, not by file: in the full branch, drop a node when it is an **AST-class node whose id ∉ `new_ast_ids`**; keep genuine semantic nodes. Cleanest implementation: tag AST nodes with a provenance marker (e.g. `"_src": "ast"`) at extraction time so the watch filter is unambiguous instead of inferring AST-vs-semantic from extension/id heuristics. Once the stale node leaves `all_ids`, the existing edge filter (`watch.py` ~552-555) drops its dangling edge for free.

## Regression test

1. Fixture: `a.py` defines `foo()` + `bar()`; `b.py` defines `caller()` calling `a.foo`. Full extract → assert `foo`, `bar`, `caller` and the `calls` edge `caller→foo` exist.
2. Delete `foo` from `a.py` (keep `bar`); `b.py` unchanged.
3. Run the **full** path: `_rebuild_code(root)` with no `changed_paths` (what `graphify update` invokes).
4. Assert: `foo` and the `caller→foo` edge are **gone**; `bar` + `caller` survive; `b.py` intact.
5. Guard the unsafe-blanket-evict regression: a pre-seeded **semantic** node on the surviving `a.py` (id not in AST output) must **NOT** be dropped.

stijnswapped:
## Summary
`graphify update` and the post-checkout hook call `_rebuild_code` with no
`changed_paths` (full re-extraction). A symbol removed from a file that still
exists was never pruned — its node is gone from the fresh `new_ast_ids`, but its
`source_file` still exists on disk so it wasn't added to `evict_sources`. The
stale node (and its inbound call/use edge) survived every rebuild until a full
clean rebuild. Fixes #1116.

A naive "evict every re-extracted file's nodes by `source_file`" fix would
silently delete semantic/LLM nodes, which aren't re-extracted in the AST-only
update pass yet point at a surviving code file.

## Fix
Prune by node identity, not by file:
- Tag AST nodes with `_origin="ast"` at extraction time (`extract.py`).
- On a full rebuild, drop any AST-marked node missing from the fresh AST output
  (`watch.py`). Semantic nodes lack the marker and are preserved; the existing
  deleted-file reconciliation still evicts them when their file is gone. The
  marker round-trips through `graph.json`. Incremental rebuilds are unaffected.

## Test
New regression test in `test_watch.py`: removes `foo()` from a surviving file
and asserts the node + its `caller→foo` edge are pruned, while `bar()`,
`caller()`, and a pre-seeded semantic node on the same file survive. Verified it
fails without the fix.


safishamsi (changes requested):
Excellent fix @stijnswapped — this is exactly the safe approach (prune by AST-node identity via the `_origin="ast"` marker, never blanket-by-source-file). I verified it end to end:

- The marker is stamped at a **single choke point** at the end of `extract()` over `all_nodes`, after every node-producing pass — so coverage is complete (per-language extractors, `ensure_symbol_node` stubs, file nodes, disambiguation/rewire all flow through it). No AST node escapes tagging.
- **Semantic nodes are safe** — they enter only via `build_merge`, never through `extract()`, so they never get the marker → never wrongly dropped. The data-loss landmine is avoided.
- watch.py eviction is gated on the full-rebuild branch only; the dangling inbound edge is removed by the existing both-endpoints filter. The new test genuinely passes-with / fails-without the fix.

**One thing to add before merge — the first-update backward-compat case.**

An existing `graph.json` built before this PR has no `_origin` markers. On the **first** `graphify update` after upgrading, a symbol removed from a surviving file won't be dropped that cycle — its old node carries no marker, so the new drop-rule skips it. It self-heals on the *next* update (once the node has been re-stamped), so there's no data loss — but it's a one-cycle lag that's currently unstated and untested.

Could you:
1. Add a test seeding a **marker-less** AST node (simulating a pre-upgrade graph) on a surviving file, and assert the first-update behavior (documents the one-cycle lag explicitly), and
2. Add a one-line note about it in the PR/CHANGELOG so the behavior is on record.

**Nice-to-have (non-blocking):** `_origin` round-trips into `graph.json` (correct, needed for persistence) but also leaks as a node property into the Cypher/GraphML exports (`export.py` unfiltered prop passthrough). Cosmetic — worth stripping it there.

With the backward-compat test + note, this is good to merge. Great catch on the landmine and clean execution.

stijnswapped:
> Excellent fix @stijnswapped — this is exactly the safe approach (prune by AST-node identity via the `_origin="ast"` marker, never blanket-by-source-file). I verified it end to end:
> 
> * The marker is stamped at a **single choke point** at the end of `extract()` over `all_nodes`, after every node-producing pass — so coverage is complete (per-language extractors, `ensure_symbol_node` stubs, file nodes, disambiguation/rewire all flow through it). No AST node escapes tagging.
> * **Semantic nodes are safe** — they enter only via `build_merge`, never through `extract()`, so they never get the marker → never wrongly dropped. The data-loss landmine is avoided.
> * watch.py eviction is gated on the full-rebuild branch only; the dangling inbound edge is removed by the existing both-endpoints filter. The new test genuinely passes-with / fails-without the fix.
> 
> **One thing to add before merge — the first-update backward-compat case.**
> 
> An existing `graph.json` built before this PR has no `_origin` markers. On the **first** `graphify update` after upgrading, a symbol removed from a surviving file won't be dropped that cycle — its old node carries no marker, so the new drop-rule skips it. It self-heals on the _next_ update (once the node has been re-stamped), so there's no data loss — but it's a one-cycle lag that's currently unstated and untested.
> 
> Could you:
> 
> 1. Add a test seeding a **marker-less** AST node (simulating a pre-upgrade graph) on a surviving file, and assert the first-update behavior (documents the one-cycle lag explicitly), and
> 2. Add a one-line note about it in the PR/CHANGELOG so the behavior is on record.
> 
> **Nice-to-have (non-blocking):** `_origin` round-trips into `graph.json` (correct, needed for persistence) but also leaks as a node property into the Cypher/GraphML exports (`export.py` unfiltered prop passthrough). Cosmetic — worth stripping it there.
> 
> With the backward-compat test + note, this is good to merge. Great catch on the landmine and clean execution.



Thanks for the thorough review, really appreciate the careful end-to-end verification. Pushed a follow-up commit addressing all three points:

1. Backward-compat test (one-cycle lag). Added test_rebuild_code_preupgrade_marker_less_node_one_cycle_lag: it seeds a marker-less AST node (simulating a pre-PR graph.json) for a symbol removed from a surviving file, asserts it is not pruned on the first post-upgrade update — exactly the one-cycle lag you flagged, no data loss — and then asserts it self-heals once the node carries _origin="ast" (as a full re-extraction stamps every surviving symbol).

2. CHANGELOG note. Recorded the fix with an explicit backward-compat line: a graph.json built before this PR has no _origin markers, so on the first update after upgrading a removed symbol isn't pruned that one cycle; it self-heals on the next re-extraction.

3. _origin export leak (nice-to-have). Stripped it at the unfiltered prop passthrough in push_to_neo4j (nodes + edges) and to_graphml. Went slightly broader and excluded all internal _-prefixed markers, since the same passthrough was also leaking the _src/_tgt edge-direction markers from build.py. The filtered to_cypher path never leaked these. _origin still round-trips into graph.json for persistence.

All watch + export tests green. Let me know if you'd like the marker-strip narrowed to just _origin.

safishamsi:
Landed in 7467c1b.

**What it fixes**: `graphify update` never pruned symbols removed from files that still exist on disk — stale nodes survived every rebuild.

**Fix**: `extract()` now stamps every AST-extracted node with `_origin="ast"`. On a full rebuild, `_rebuild_code` drops any AST-marked node absent from the fresh output even when its source file survives. Semantic/LLM nodes are never touched. Backward-compat: pre-1118 graphs (no marker) get one free cycle then self-heal.

Smoke tested end-to-end: removing `foo()` from a surviving file → node gone after next rebuild; `bar()` stays. 1910 passed, 0 failures.

Issue #1114, opened 2026-06-02
Agent bypasses the graph by Read-ing files: PreToolUse hook only catches Bash grep, not the Read tool (help wanted)


safishamsi:
## Problem

graphify's core value prop is: once `graphify-out/graph.json` exists, the coding agent should answer codebase questions with `graphify query` / `path` / `explain` (a scoped subgraph) instead of grepping or reading raw files. Today that's enforced by a `PreToolUse` hook (`graphify/__main__.py`, `_SETTINGS_HOOK`) that fires before Bash search commands and nudges the agent toward `graphify query`:

```
case "$CMD" in
  *grep*|*rg\ *|*ripgrep*|*find\ *|*fd\ *|*ack\ *|*ag\ *)
    [ -f graphify-out/graph.json ] && echo '{...additionalContext: use `graphify query` instead...}' ;;
esac
```

**The gap:** it only matches **Bash search commands**. The most common way an agent actually avoids the graph is invisible to it:

- **The native `Read` tool** — an agent that answers "how does X work?" by reading 20-30 files one at a time never runs a Bash command, so the hook never fires and the graph is bypassed entirely.
- **Bash content reads** — `cat`, `head`, `tail`, `sed -n`, `awk` over many files dodge the search-tool match too.

So a graph can exist and still be ignored on exactly the questions it's meant to answer.

## Ask (help wanted)

Extend the enforcement so the agent is nudged toward `graphify query` when a graph exists and it's about to browse raw files, not just when it greps:

1. Add a `PreToolUse` hook on the **`Read` tool** (and `Glob`) that fires only when `graphify-out/graph.json` exists, injecting the same "query first" `additionalContext`.
2. Broaden the Bash matcher to include multi-file content reads (`cat`/`head`/`sed`/`awk`) — but only when **multiple paths / a glob** are involved, so single-file reads don't trigger.

## Design constraint: don't spam

Firing on *every* `Read` is noise that trains the agent to ignore the nudge. Scope it:

- Gate on `graphify-out/graph.json` existing (already the pattern).
- Prefer nudging when the Read/glob targets a **source or doc file** (not configs, not files already inside `graphify-out/`).
- Keep it a nudge (`additionalContext`), never a block — legitimate reads must still go through.

## Scope / why it's low-risk

- This is hook configuration in `graphify/__main__.py` + tests — it touches **no skill-file content**, so it can't break the extraction pipeline.
[...]

Issue #1106, opened 2026-06-01
Compress skill files via progressive disclosure to cut token/context cost (help wanted)


safishamsi:
## Problem

The skill files are large and get loaded into the agent's context on every `/graphify` invocation, burning tokens on material that isn't needed for the task at hand. Current sizes:

- `skill.md` — ~57 KB / 1,156 lines
- 14 platform variants (`skill-windows.md`, `skill-opencode.md`, `skill-devin.md`, `skill-copilot.md`, `skill-codex.md`, `skill-amp.md`, `skill-droid.md`, `skill-trae.md`, `skill-claw.md`, ...) each ~51–69 KB

Every one carries the full extraction spec, per-language node-ID rules, embedded Python snippets, and troubleshooting inline. On smaller-context agents this is a real squeeze (see also #984 "too much context"), and it inflates token cost for simple queries that never touch most of that content.

## Ask (help wanted)

Compress / restructure the skill files via **progressive disclosure** so the always-loaded core stays minimal:

1. Keep `skill.md` lean — the trigger, the high-level workflow, and routing. Target a fraction of the current size.
2. Move the heavy reference material into separate files loaded **on demand**:
   - the extraction JSON spec + node-ID format rules
   - per-language extraction details
   - platform-specific setup / troubleshooting
3. Move embedded Python snippets out of the always-loaded path into referenced script files.
4. Because the bulk is **duplicated across all 14 platform variants**, a shared-fragment / templated-generation approach would cut both context size *and* the maintenance burden of keeping 14 files in sync.

## Constraints

- Node-ID format rules and the extraction JSON contract must stay byte-accurate after the split — these are load-bearing (a wrong stem/format silently creates ghost-duplicate nodes). Don't paraphrase them during the move.
- Verify the on-demand references actually resolve on each platform's loader (Claude Code skills, Codex, OpenCode, etc.) before/after.

Related: #1045 (same underlying request from a user). Filing this as a tracked, scoped help-wanted task.

TPAteeq:
# Skill restructure proposal

Happy to take this on. Here's the design I propose:

1. **Shared-fragment generator.** Add `skill_src/` (a lean `core.md` body plus per-platform delta files for the frontmatter description, the `/graphify` vs `$graphify` token, and install paths) and a build script that generates all 14 `skill-*.md`. Generated files stay committed. CI regenerates and diffs to catch drift. This kills the 14-files-in-sync manually burden.

[...]

PR #1119, opened 2026-06-02
feat(skill): progressive-disclosure split for the Claude skill (#1106)

TPAteeq:
Addresses #1106 (Claude-first; follow-ups noted below). Implements the design agreed in the issue thread.

  ## What this does

  Splits the monolithic `skill.md` (~1,156 lines, loaded in full on every `/graphify`) into a lean `SKILL.md` +
  on-demand `reference/` files, generated from `skill.md` as the single source of truth.

  **Per-invocation win:** a `/graphify query` now loads `SKILL.md` (76 lines) + `reference/query.md` (~126) =
  **~202 lines vs 1,156 (~83% less always-on context)**.

  ## How

  - **`graphify/skill_build.py`** — a fence-aware heading slicer driven by a route-by-command manifest. Two output
  modes from one source: **split** (lean `SKILL.md` + `reference/{extract,update,query,add,watch}.md`) for 
  verified skill loaders, and **flatten** (a lossless single-file round-trip of `skill.md`) for the rest. `python 
  -m graphify.skill_build` regenerates; `--check` is a CI drift gate.
  - **Generated Claude tree committed** at `graphify/skill_claude/`.
  - **Install** — the `claude` platform uses the split tree; `_copy_skill_file` copies `SKILL.md` + `reference/` 
  with `SKILL.md` written **last** for crash-safety. Removal centralized in `_remove_skill_tree` (handles 
  `reference/`, scoped to graphify-owned names) and routed through `_remove_skill_file` + the 
  vscode/kiro/antigravity uninstalls. Packaged via `pyproject` `package-data`; confirmed it lands in the built 
  wheel.
  - **Byte-accuracy** — the node-ID rule and JSON extraction contract move as verbatim slices and are pinned 
  against `_file_node_id` in tests (not a frozen string). The skill `description` is preserved verbatim with a 
  test asserting it.

  ## Scope

  Claude is the **only** loader switched to split mode; every other platform keeps its existing single-file 
  behavior.

  ## Testing
  
  - Full suite green: **1,545 passed, 11 skipped**, incl. 17 new tests in `tests/test_skill_build.py`.
  - **Verified live in a Claude Code session:** a fresh `/graphify .` followed `reference/extract.md` and built a 
  real graph; `/graphify query` ran the constrained query-expansion procedure that lives **only** in 
  `reference/query.md` — confirming on-demand loading at runtime.

  ## Follow-ups (intentionally not in this PR)

  - Regenerate the other 12 flatten-mode platform files from `skill.md`. Also a **correctness fix**: 8 stale 
  variants carry ~49 bare-path refs (`Path('.graphify_detect.json')` → `FileNotFoundError`) that regenerating 
  wipes out. Kept separate to keep this PR Claude-only and reviewable.
  - #1114 (PreToolUse hook only catches Bash grep, not the Read tool) — separate track.

  ## Two pre-existing issues found while here (not changed)

  - `uninstall_all` (`graphify uninstall`) removes the *gemini* skill dir but not the *claude* skill dir (only 
  strips the CLAUDE.md section). Left as-is since changing uninstall scope is a behavior decision - flagging.
  - Headless `graphify extract <path>` validates an LLM backend upfront ( before checking corpus content ), so it errors '              
     no LLM API Key found' even for a code-only corpus that needs no LLM calls. the keyless AST-only path is 'graphify
     update <path> --no-cluster' ( or '--backend ollama' on loopback).


safishamsi (changes requested):
Really nice work @TPAteeq — this lands the design from the thread faithfully. I merged it onto current `v8` and verified the risky parts mechanically, all green:

- Full suite passes on merged `v8` (1,533 passed, 0 failed).
- Drift gate works: `python -m graphify.skill_build --check` → in sync.
- **Byte-accuracy confirmed** — the node-ID spec, JSON schema, and the full subagent prompt template are present **verbatim** in `reference/extract.md`; the `description` is identical to `skill.md`.
- Lean core = 76 lines (query path ≈ 202 vs 1,156, ~82% cut).
- Wheel ships all 6 files (`skill_claude/SKILL.md` + `reference/*.md`) — built and inspected.
- Install lays down the tree correctly; scope is genuinely claude-only.

**One change before merge — make uninstall remove the tree it now installs.**

Testing surfaced that `graphify claude uninstall` (and bare `graphify uninstall`) don't remove the claude skill — they only touch `CLAUDE.md`. I confirmed this is **pre-existing** (base `v8` leaves the single `SKILL.md` behind too), so the PR doesn't regress it — but it makes the orphan a **6-file `reference/` directory** instead of one file. And since this PR already builds `_remove_skill_tree`, the helper is right there but the actual claude uninstall path never calls it.

Could you wire the claude uninstall to call `_remove_skill_tree` (the gemini path already removes its skill dir — this just makes claude consistent)? That turns "uninstall orphans 6 files" into a clean removal, and the helper you wrote does exactly the right scoped thing (I verified it preserves a sibling user file). With that, this is good to merge.

**Two follow-ups you already flagged — agreed, keep them separate:**
- The 8 stale flatten-mode variants with the bare-path `FileNotFoundError` bugs — regenerating from `skill.md` fixes them; worth its own PR.
- The headless `graphify extract` upfront-backend-validation on code-only corpora — pre-existing, separate issue.

Thanks again — the byte-verbatim slicing + the flatten round-trip + the CI drift gate are exactly the safety net this needed.

TPAteeq:
Thanks for the thorough review and for verifying the risky parts — really appreciate it.
  
  Done in 609a98c. Wired the user-scope `claude_uninstall` through `_remove_skill_file` → `_remove_skill_tree`
  (first thing it does now, mirroring the gemini path), so `graphify claude uninstall` and bare `graphify
  uninstall` remove the full tree — `SKILL.md` + `reference/` + the version stamp — instead of orphaning the
  6-file directory. Project scope was already covered, and I passed `project=True` through that path so a project
  uninstall still never touches the user-global skill. The scoped removal also preserves a sibling user file
  (verified).
  
  Added a regression test for the user-scope path. Full suite green (1,547 passed) and the drift gate (`python -m
  graphify.skill_build --check`) passes.
  
  Also agreed on keeping the two follow-ups separate — happy to pick up the stale flatten-variant regeneration
  (the bare-path `FileNotFoundError` fix) as the next PR once this lands.
  
  Thanks again!

safishamsi (approved):
Verified the fix on current `v8` — `609a98c` does exactly the right thing:

- `graphify claude uninstall` (and bare `graphify uninstall`) now remove the full tree — `SKILL.md` + `reference/` + the version stamp — via `_remove_skill_tree`, mirroring the gemini path.
- Scoped removal confirmed: a sibling `user_notes.md` I dropped in the skill dir **survives** the uninstall — it only removes graphify-owned names, never a blind `rmtree`.
- Drift gate green (`skill_build --check` in sync), full suite 1534 passed.

That was the only blocker. This is clean — byte-verbatim slices, lossless flatten round-trip, preserved `description`, wheel ships the tree, claude-only scope. Approving. Thanks for the careful turnaround.

TPAteeq:
Thanks @safishamsi — appreciate the thorough review and the quick verification turnaround. Glad it landed
  cleanly.

  On next steps: regenerating the other platform variants from skill.md (which also fixes the bare-path
  FileNotFoundError bugs in the stale ones) was the natural follow-up, but #1121 already covers that all-platform
  rollout. Rather than overlap, I'll follow your lead — I can extend #1119's generator to the other hosts if you'd
  like, or leave it to #1121.

safishamsi:
Thank you @TPAteeq, this was the right Claude-first reference and shaped the approach. We ended up merging the fuller all-platform version in #1121, which extends the same progressive-disclosure design to 18 hosts and adds a single-source generator (`tools/skillgen/`) with drift guards. That is now released in **0.8.29**, so this PR is superseded. Really appreciate the work, it directly informed what shipped.

PR #1120, opened 2026-06-02
feat(hooks): close the Read-tool graph bypass (Read/Glob + multi-file reads)

### LINKED ISSUE #1114: Agent bypasses the graph by Read-ing files: PreToolUse hook only catches Bash grep, not the Read tool (help wanted)

yaanfpv:
## What this does

graphify's whole point is that once `graphify-out/graph.json` exists, the agent should answer codebase questions with `graphify query` / `path` / `explain` instead of grepping or reading raw files. Today the `PreToolUse` nudge (`_SETTINGS_HOOK`) only fires on Bash search (`grep`, `rg`, `find`, ...). The most common way an agent actually skips the graph slips right past it: reading 20 files one at a time with the `Read` tool, globbing, or `cat`-ing several files at once. So a graph can exist and still be ignored on exactly the questions it is built to answer.

This closes that gap on the Claude path and extends the always-on prose for the other platforms.

## How

- **New `Read|Glob` PreToolUse hook** (`_READ_SETTINGS_HOOK`). It fires only when `graphify-out/graph.json` exists, and only for a source or doc file (suffix allowlist); anything under `graphify-out/` is suppressed so reading the graph's own report never starts a "go read the graph" loop. It injects the same query-first `additionalContext` and never blocks. The parser is `python3` only (already a graphify dependency, no `jq`), the shell is POSIX, and every branch fails open.
- **Broadened the Bash matcher** to also cover multi-file content reads (`cat`, `head`, `tail`, `sed`, `awk`), but only when multiple paths or a glob are involved, so a single-file read stays silent and the nudge does not become spam. Redirects, heredocs, pipes, and reads under `graphify-out/` are excluded. The existing grep/search arm and its payload are byte-for-byte unchanged.
- **Idempotency:** both the install and uninstall dedup filters now recognize the new matcher, so reinstalling or upgrading does not stack duplicate Read hooks.
- **Non-Claude paths:** the always-on sections (CLAUDE.md, AGENTS.md, GEMINI.md, Cursor, Kiro, Antigravity, Devin) now name reading files alongside grep as the thing `graphify query` replaces. Codex stays prose-only, since Codex Desktop rejects `additionalContext` on `PreToolUse`, so its coverage comes from the AGENTS.md text.

## Testing

- New `tests/test_read_hook.py` runs the hook commands the way Claude Code does, via subprocess with crafted stdin JSON: silent without a graph, nudges with one, source/doc vs config vs `graphify-out/` targets, glob coverage, single-file-silent vs multi-file/glob-nudge, and a regression that grep still nudges after the broadening.
- `tests/test_claude_md.py` gains the settings.json shape checks, an idempotency-across-double-install assertion, and clean uninstall. The new hook is added to the issue-#580 install-string contract so it is held to the same "recommends `graphify query`" guarantee as every other surface.
- Full suite green (47 new tests). `ruff check` clean on the changed files.

## Scope / why it's low-risk

This is hook configuration in `graphify/__main__.py` plus tests and the always-on prose. It touches no skill-file content, so it cannot affect the extraction pipeline. It never blocks a tool call: every branch is a pure `additionalContext` nudge that fails open, so legitimate reads always go through.

## Related

Closes #1114.

### DIFF
--- a/graphify/__main__.py
+++ b/graphify/__main__.py
@@ -192,6 +192,8 @@ def _print_project_git_add_hint(paths: list[Path]) -> None:
 _SETTINGS_HOOK = {
     # Claude Code v2.1.117+ removed dedicated Grep/Glob tools; searches now go through Bash.
     # We match on Bash and inspect the command string to avoid firing on every shell call.
+    # The search arm nudges grep/rg/find/etc.; the content-read arm nudges multi-file
+    # cat/head/tail/sed/awk (single-file reads stay silent so the nudge does not become spam).
     "matcher": "Bash",
     "hooks": [
         {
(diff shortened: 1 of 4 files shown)


safishamsi:
Closing as superseded — this feature is already in v8 in a refined form. The `_READ_SETTINGS_HOOK` (Read|Glob PreToolUse hook), dedup/uninstall wiring, and `tests/test_read_hook.py` are all present on the current branch. Thanks for the contribution — this was implemented independently.

PR #1123, opened 2026-06-02
fix(extract): don't require an LLM API key for a code-only corpus (#1122)

### LINKED ISSUE #1122: extract requires an LLM API key for code-only corpora, even though code extraction makes no LLM calls
## Problem

`graphify extract <path>` refuses to run on a **code-only** corpus unless an LLM API key is set — even though code extraction is pure local AST (tree-sitter) and never calls a model. This blocks the common "just give me an AST graph, no LLM" use case from the headless CLI.

## Repro (all LLM keys unset)

```bash
mkdir -p /tmp/x/src
printf 'def login(u):\n    return validate(u)\n\ndef validate(u):\n    return True\n' > /tmp/x/src/auth.py

graphify extract /tmp/x/src
# error: no LLM API key found. Set GEMINI_API_KEY ... or pass --backend.
# exit 1, no graph.json produced
```

The AST-only path *does* work via a different command:

```bash
graphify update /tmp/x/src --no-cluster
# Re-extracting code files ... (no LLM needed) -> graph.json written
```

So the capability exists; `extract` just gates on a key it doesn't need.

## Cause

The `extract` command (`graphify/__main__.py`) resolves and validates an LLM backend **upfront** — it calls `detect_backend()` and exits with `no LLM API key found` if none is set — **before** detecting the corpus composition. A code-only corpus (no docs/papers/images) needs no model, but is blocked at that gate anyway.

## Expected

A code-only `extract` should run with no API key, producing `graph.json` from AST alone (matching `graphify update --no-cluster`). The key should only be required when there's actual semantic work — docs/papers/images, or `--dedup-llm`.

## Proposed fix

Defer backend resolution until after file detection and only require a key when `semantic_files` is non-empty or `--dedup-llm` is passed. Improve the error to name why a key is needed and note that code-only corpora need none.

TPAteeq:
Fixes #1122.
  
  `graphify extract <path>` validated an LLM backend upfront — before detecting corpus composition — so a
  code-only corpus (pure local AST, zero LLM calls) errored with `no LLM API key found` and exited 1 without
  writing a graph. 
  
  This defers backend resolution until after file detection and only requires a key when there's actual semantic
  work (docs/papers/images) or `--dedup-llm`. A code-only `extract` now runs keyless, matching `graphify update
  --no-cluster`. The error message now names why a key is needed and notes that code-only corpora need none.
  
  ## Testing
  - New regression tests: code-only `extract` succeeds with no key; a corpus with a doc still errors (exit 1) with
  the guidance message.
  - Full suite green; ruff clean on changed files.

### DIFF
--- a/graphify/__main__.py
+++ b/graphify/__main__.py
@@ -3548,80 +3548,6 @@ def main() -> None:
         if cli_max_workers is not None:
             os.environ["GRAPHIFY_MAX_WORKERS"] = str(cli_max_workers)
 
-        # Backend resolution. If user did not pass --backend, sniff env.
-        # If backend was explicitly requested, validate its key is present
-        # and surface a clear error early — don't let extract_corpus_parallel
-        # raise mid-run after we've spent time on AST extraction.
-        from graphify.llm import (
-            BACKENDS as _BACKENDS,
-            detect_backend as _detect_backend,
-            estimate_cost as _estimate_cost,
-            extract_corpus_parallel as _extract_corpus_parallel,
-            _format_backend_env_keys,
-            _get_backend_api_key,
-        )
-        if backend is None:
-            backend = _detect_backend()
-            if backend is None:
-                print(
-                    "error: no LLM API key found. Set GEMINI_API_KEY or GOOGLE_API_KEY "
-                    "(gemini), MOONSHOT_API_KEY (kimi), ANTHROPIC_API_KEY (claude), "
-                    "OPENAI_API_KEY (openai), DEEPSEEK_API_KEY (deepseek), "
-                    "or pass --backend.",
-                    file=sys.stderr,
-                )
-                sys.exit(1)
-        if backend not in _BACKENDS:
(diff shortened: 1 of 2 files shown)


safishamsi:
Landed in f85339b with one improvement: `_clear_backend_keys` in the tests now also clears `AWS_PROFILE`, `AWS_REGION`, `AWS_DEFAULT_REGION`, `OLLAMA_BASE_URL` — those are read by `detect_backend()` and could make the code-only test pass for the wrong reason on a machine with ambient AWS credentials or a local Ollama setup. The second test already patches `detect_backend` directly so it was fine; only the first needed the fuller clear.

Everything else applied as-is — the placement (after `semantic_files` population), the `needs_llm` condition, the `backend not in _BACKENDS` check outside `if needs_llm:`, and the ollama/bedrock/claude-cli special cases all look correct. 1849 passed, 0 failures.

PR #1129, opened 2026-06-03
feat(terraform): add Terraform/HCL support (#187)

### LINKED ISSUE #187: Terraform Support
Thank you very much for providing such a good tool. Could you please provide support for Terraform files? Our cloud infrastructure is managed through Terraform. Thank you! The suffix of the Terraform file is.tf

TPAteeq:
Closes #187.

#### What this does

Adds AST extraction for **Terraform / HCL** files (`.tf`, `.tfvars`, `.hcl`) via `tree-sitter-hcl`, turning an infrastructure repo into a real dependency graph. Before this, `.tf` files matched no extension set in `detect.py`, so they classified as `None` and were **skipped entirely** — graphify produced zero nodes/edges for Terraform. Now they become a structured graph.

#### Graph model

- **Nodes:** `resource`, `data` source, `module`, `variable`, `output`, `provider`, and per-key `locals`. (The `terraform {}` settings block is intentionally not a node.)
- **Edges:**
  - `contains` — file -> each block it defines
  - `references` — interpolation references, e.g. `aws_instance.web -> var.region`, `-> data.aws_ami.ubuntu`, `module.vpc -> local.cidr`, `output.ip -> aws_instance.web`
  - `depends_on` — explicit dependency edges

Labels use canonical Terraform addresses (`aws_instance.web`, `data.aws_ami.ubuntu`, `var.region`, `module.vpc`, `local.cidr`, `output.ip`).

#### Key design decision: directory-scoped node IDs

Terraform resources are **module (directory)-scoped**, not file-scoped — a resource defined in `main.tf` is referenced from sibling `.tf` files. Node IDs are therefore scoped by the parent directory rather than the file stem, so a definition in one file and a reference in another resolve to the same node when per-file extractions are merged. (Stem scoping, used by most language extractors, would split a definition from its references.)

#### Why AST (deterministic, free) is the right engine here

Extraction is byte-identical across runs, costs no tokens/API key, resolves references exactly from the HCL grammar (not guessed), and covers every resource/reference — all properties an LLM pass can't match for IaC.

#### Verified

- **Real repo** (a 9-file Terraform module): **25 nodes, 51 edges** (27 `contains`, 24 `references`), with cross-file resolution confirmed (`azurerm_network_interface.*` in one file -> `azurerm_resource_group.main` in another). Extraction is deterministic (two runs identical).
- **Tests** (`tests/test_terraform.py`, 7): block/edge model, `references`, `depends_on`, meta-head filtering (`count.index`/`each.key`/`path.module` are not emitted as refs), leading-comment handling, empty/comment-only files, and cross-file resolution through a real `build_from_json` merge.
- Full suite **1810 passed, 1 skipped** (no regressions); `ruff` clean; `uv sync --all-extras --frozen` consistent; `skillgen --check` unaffected.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 Full release notes with details on each version: [GitHub Releases](https://github.com/safishamsi/graphify/releases)
 
+## Unreleased
+
+- Feat: Terraform/HCL support (#187). `.tf`, `.tfvars`, and `.hcl` files are now AST-extracted via tree-sitter-hcl into a real infrastructure dependency graph: resources, data sources, modules, variables, outputs, providers, and locals become nodes, and interpolation references (`aws_instance.web -> var.region`, `-> data.aws_ami.ubuntu`, ...) plus explicit `depends_on` become edges. Node IDs are scoped by directory rather than file, so cross-file references (a resource defined in `main.tf` and used in `network.tf`) resolve when per-file extractions are merged.
+
 ## 0.8.30 (2026-06-03)
 
 - Fix: `graphify install --project --platform antigravity` now writes Antigravity's always-on layer (`.agents/rules/graphify.md` + `.agents/workflows/graphify.md`), not just the skill. The project-scoped path went through the skill-only branch and skipped them, even though the project uninstall removes them.
--- a/README.md
+++ b/README.md
@@ -225,7 +225,7 @@ To remove graphify from all platforms at once: `graphify uninstall` (add `--purg
 
 | Type | Extensions |
 |------|-----------|
-| Code (33 languages) | `.py .ts .js .jsx .tsx .mjs .go .rs .java .c .cpp .h .hpp .rb .cs .kt .scala .php .swift .lua .luau .zig .ps1 .ex .exs .m .mm .jl .vue .svelte .astro .groovy .gradle .dart .v .sv .svh .sql .f .f90 .f95 .f03 .f08 .pas .pp .dpr .dpk .lpr .inc .dfm .lfm .lpk .sh .bash .json .dm .dme .dmi .dmm .dmf .sln .csproj .fsproj .vbproj .razor .cshtml` (`.dm`/`.dme` AST extraction requires `uv tool install graphifyy[dm]`) |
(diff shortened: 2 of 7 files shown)


safishamsi (commented):
Great extractor — the directory-scoped node ID design is correct (verified: cross-module collision doesn't occur, `_make_id` variadic args work fine), the `.id` comparison for tree-sitter node identity is the right approach, the `depends_on` traversal reaches into the tuple correctly, and all 7 tests pass. Three things to fix before merge:

**1. Move `tree-sitter-hcl` to an optional extra (blocking)**

Currently placed in core `dependencies`, which means every graphify user installs it even if they never touch Terraform. The established precedent in this codebase is the opposite: `tree-sitter-dm` and `tree-sitter-sql` are both in `[project.optional-dependencies]`, even though their extensions ARE in `CODE_EXTENSIONS`. `extract_terraform` already has the correct graceful-degradation path on `ImportError`, so making it optional is low-effort:

```toml
[project.optional-dependencies]
terraform = ["tree-sitter-hcl"]
all = [..., "tree-sitter-hcl"]  # add here too
```

And add a note to the README row like the existing `(requires `uv tool install graphifyy[dm]`)` note. If the maintainer explicitly wants HCL in core (it's a pure wheel, no compile), that's a call to make explicitly — just needs sign-off.

**2. README count regression (blocking)**

The PR branch predates commit `fff1c98` which corrected the grammar count to "27 tree-sitter grammars." The PR's README change reintroduces both errors: "Code (34 languages)" instead of the current correct "Code (27 tree-sitter grammars)" (or 28 after HCL). Please rebase onto current `v8` and update the count using the "N tree-sitter grammars" phrasing, not "N languages." This will also surface as a merge conflict on that line.

**3. `.tfvars` silent no-op — add a test (should fix)**

`.tfvars` files contain only key=value assignments (no block structure), so `extract_terraform` extracts zero nodes from them — only the file node is emitted. The extension is registered and dispatched to this extractor, but the behavior is undocumented and untested. Please add a test asserting the current behavior:

```python
def test_tfvars_is_safe(tmp_path):
    r = extract_terraform(_write(tmp_path, "terraform.tfvars", 'region = "us-east-1"\nenv = "prod"\n'))
    assert r.get("error") is None
    assert len(r["nodes"]) == 1  # only the file node
```

If you want to extract tfvars assignments as `var.*` nodes (since they override variables), that would be a nice enhancement — but at minimum document the current behavior.

**Minor (not blocking):**
- Non-Terraform `.hcl` files (Packer, Vault, Nomad job specs) produce zero nodes silently — consistent with the rest of graphify's handling of unparseable content, so acceptable as-is.

safishamsi:
Landed in 200edec with the three fixes applied:
1. **`tree-sitter-hcl` moved to optional extra** — `graphify[terraform]` (not core), added to `all` extra, matching the `sql`/`dm` precedent. Graceful-degradation path on ImportError was already in place.
2. **README count preserved** — kept at '27 tree-sitter grammars' per the corrected count (HCL adds one grammar package but the README wording is now grammar-based not language-based).
3. **`.tfvars` test added** — `test_tfvars_key_value_is_safe` documents that .tfvars key=value assignments produce only the file node (no block structure to extract), which is the correct behaviour.

All 8 tests pass, full suite 1846 passed 0 failures. Great extractor — the directory-scoped ID design, tree-sitter .id comparison, and depends_on traversal are all correct. Thanks for the clean implementation!

PR #1136, opened 2026-06-05
feat: Add CodeBuddy platform support

studyzy:
[CodeBuddy](https://www.codebuddy.ai/) uses the same Agent tool and PreToolUse hook mechanism as Claude Code. 
* Installs skill to ~/.codebuddy/skills/, 
* writes CODEBUDDY.md section, 
* registers a Glob|Grep PreToolUse hook in .codebuddy/settings.json.

### DIFF
--- a/README.md
+++ b/README.md
@@ -126,7 +126,8 @@ for example `graphify claude install --project` or `graphify codex install --pro
 | Platform | Install command |
 |----------|----------------|
 | Claude Code (Linux/Mac) | `graphify install` |
-| Claude Code (Windows) | `graphify install --platform windows` |
+| Claude Code (Windows) | `graphify install` (auto-detected) or `graphify install --platform windows` |
+| CodeBuddy | `graphify install --platform codebuddy` |
 | Codex | `graphify install --platform codex` |
 | OpenCode | `graphify install --platform opencode` |
 | Kilo Code | `graphify install --platform kilo` |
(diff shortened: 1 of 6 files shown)


safishamsi (commented):
CodeBuddy support is welcome — the registry entry and skill-file routing look structurally correct. Three issues need to be resolved before merge:

**1. PreToolUse hook is missing (contradicts the PR description)**

The description says the install "registers a Glob|Grep PreToolUse hook in `.codebuddy/settings.json`" but the diff only shows CODEBUDDY.md installation. Without the hook, CodeBuddy users get no automatic graph-nudging before Glob/Grep calls — which is the main value of the platform integration. Please add a `_install_codebuddy_hook` function mirroring `_install_claude_hook` (`__main__.py:1740–1760`) with the appropriate matcher filter, and wire it from both the install and uninstall paths.

**2. Missing uninstall handler — CODEBUDDY.md will be orphaned**

Every other platform that writes a platform-specific file has a corresponding removal path in `_project_uninstall` and `uninstall_all`. CodeBuddy is not added to either. The fallback `else: _remove_skill_file(...)` only removes the SKILL.md — the CODEBUDDY.md written by the install block is never cleaned up. Please add an explicit uninstall branch (analogous to the `gemini_uninstall` or `_agents_uninstall` patterns).

**3. CODEBUDDY.md idempotency**

The install block uses a simple `if "graphify" in content` check rather than the marker-based `_replace_or_append_section` used for CLAUDE.md. This means re-running install after an upgrade won't update a stale section — it will just skip silently. Please use the same `_CLAUDE_MD_MARKER` / `_replace_or_append_section` mechanism so re-install is idempotent and upgradeable.

**Minor:** No platform-specific tests for CODEBUDDY.md write, hook, or uninstall. `test_install_roundtrip.py` will auto-cover the skill-file roundtrip once the registry entry lands, but the platform-specific file paths need their own test (see `test_devin.py` / `test_antigravity_install.py` as models).

studyzy:
I tested this PR by installing graphify into CodeBuddy Code on macOS. All three commands work correctly:

- `graphify install --platform codebuddy`
- `graphify codebuddy install`
- `graphify codebuddy uninstall`

Behavior is consistent with the Claude Code integration. The skill file lands in `~/.codebuddy/skills/graphify/`, the CODEBUDDY.md section is written correctly, and the PreToolUse hook is registered in `.codebuddy/settings.json`. Uninstall cleanly removes all of the above.

Tested with CodeBuddy Code on darwin/arm64. LGTM from a functional standpoint.

safishamsi:
Landed in 9e1ad42 
CodeBuddy support is live: `graphify codebuddy install` writes the skill to `~/.codebuddy/skills/graphify/SKILL.md`, registers the Bash + Read|Glob hooks in `.codebuddy/settings.json`, and writes the `CODEBUDDY.md` always-on section. Resolves the stale codex path conflict against current v8 on our end.

Two doc fixes applied: the matcher is `Bash` + `Read|Glob` (not `Glob|Grep` as the PR description said). Great contribution!

Issue #831, opened 2026-05-12
Antigravity's Graphify skill uses pip to install graphifyy instead of uv, even when uv is available and graphify was installed with uv


zenwav:
Graphify is installed system-wide using uv

#### Antigravity Build Information
```
Version: 1.118.1 (system setup)
Commit: 034f571df509819cc10b0c8129f66ef77a542f0e
Date: 2026-04-29T17:36:44+03:00
Electron: 39.8.8
ElectronBuildId: 13870025
Chromium: 142.0.7444.265
Node.js: 22.22.1
V8: 14.2.231.22-electron.0
OS: Windows_NT x64 10.0.19045
```
#### Graphify Version Information
```powershell-console
PS C:\> graphify -v
graphify 0.7.15
```
<img width="709" height="289" alt="Image" src="https://github.com/user-attachments/assets/1bce891e-3f9f-4764-aa05-faa7490145f7" />

safishamsi:
Fixed in v0.7.16 — Step 1 install fallback in skill.md and skill-windows.md now checks for `uv` first and runs `uv tool install --upgrade graphifyy` before falling back to pip.

zenwav:
@safishamsi, Latest update didn't resolve my issue.

<img width="818" height="653" alt="Image" src="https://github.com/user-attachments/assets/c17a73a2-d82a-42b0-b0bf-cb31f7c5495c" />

safishamsi:
Hey @zenwav — thanks for the screenshot, it made the root cause clear.

**What was wrong:** The skill was checking `python import graphify`, failing, installing via `uv tool install`, but then saving the path of bare `python` to `.graphify_python` — not the uv-managed Python where graphify actually lives. Every subsequent step then used the wrong interpreter and hit `ModuleNotFoundError`.

**What we fixed:** When uv or pipx installs graphify, `graphify.exe` and `python.exe` sit in the same `Scripts\` directory. The skill now uses `Get-Command graphify` to find that directory and picks up the co-located Python directly. We also replaced all bare `python` calls throughout the skill with `& (Get-Content graphify-out\.graphify_python)` so every step consistently uses the saved interpreter.

Pushed in commit 1ac3b96 on v8. To get the fix, reinstall the skill:

```
graphify install
```

Let me know if it works.

zenwav:
[...]

Issue #651, opened 2026-05-02
Codex: PreToolUse failed


rexkoh425:
Version 0.6.1
Python 3.14.0

**OS**

Windows

**Issue**

<img width="424" height="65" alt="Image" src="https://github.com/user-attachments/assets/c7eeae87-b3ef-4145-9592-126f1a60d57e" />

{
  "hooks": {
    "PreToolUse": [
      {
        "matcher": "Bash",
        "hooks": [
          {
            "type": "command",
            "command": "[ -f graphify-out/graph.json ] && echo '{\"hookSpecificOutput\":{\"hookEventName\":\"PreToolUse\",\"additionalContext\":\"graphify: Knowledge graph exists. Read graphify-out/GRAPH_REPORT.md for god nodes and community structure before searching raw files.\"}}' || true"
          }
        ]
      }
    ]
  }
}

**Instructions/commands applied so far**

1) pip install graphifyy && graphify install
2) graphify install --platform codex
3) /graphify .
4) graphify codex install
5) graphify hook install

Stef500:
Hello, same problem here.

safishamsi:
Fixed in v0.6.4. The `[ -f ]` bash check has been replaced with a cross-platform Python one-liner — re-run `graphify codex install` after upgrading to regenerate the hook.

rexkoh425:
#PreToolUse still fails after running upgrading to v0.6.4 and running "graphify codex install".

**OS:** Windows
**Virtual Enviroment**: Conda
**CLI** : Powershell
Python version: 3.14.0

Linking similar issue below

#522 


<img width="428" height="82" alt="Image" src="https://github.com/user-attachments/assets/b62fd5c9-d3c1-4892-bc81-cdd448ccc44e" />

@safishamsi

safishamsi:
I’ll look into it by evening then (GMT)

safishamsi:
Re-opening — the `python3 -c "..."` fix in v0.6.4 still fails on Windows Conda (no `python3` shim) and PowerShell (JSON brace/quote parsing). Root cause confirmed via #522.

**Real fix (in next release):** the hook command is now just `graphify hook-check` — a new shell-agnostic subcommand that prints the hookSpecificOutput JSON if `graph.json` exists and exits silently if not. No inline Python, no shell quoting, works on PowerShell, cmd.exe, macOS, and Linux.

After upgrading, re-run `graphify codex install` to regenerate the hook with the new command.

[...]

Issue #374, opened 2026-04-15
Not able to install in VScode


Mr-Five:
Hello, I am trying to install it in vscode but i am not able to do it. can someone please help me. 

<img width="823" height="1360" alt="Image" src="https://github.com/user-attachments/assets/52db71a8-62fd-45a6-a763-b687f857feb7" />

safishamsi:
Thanks for reporting. graphify is not a VS Code extension — it is an AI assistant skill that works through VS Code Copilot Chat.

v0.4.15 adds first-class VS Code Copilot Chat support. To set it up:

```bash
pip install --upgrade graphifyy
graphify vscode install
```

This installs the skill and writes `.github/copilot-instructions.md` to your project so Copilot Chat always has graph context. Then open the Copilot Chat panel in VS Code and type `/graphify`.

If you were trying to use it with GitHub Copilot CLI (the terminal tool) rather than the chat panel, use `graphify copilot install` instead.

If you are still seeing an error after upgrading, please share the exact error message and we can help.

lymons:
However, in version v0.4.18, running this CLI results in an error indicating that the vscode command is not available.

> graphify vscode install

(.venv) lymons@Mac docs % graphify vscode install
error: unknown command 'vscode'
Run 'graphify --help' for usage.

NishanthMolleti:
https://github.com/safishamsi/graphify/pull/411 here is the PR to fix this. @safishamsi

safishamsi:
This is fixed as of v0.4.19 — `graphify vscode install` is fully wired up in the published wheel. If you're still seeing `unknown command 'vscode'`, run `pip install --upgrade graphifyy` to pick up the fix.

safishamsi:
Confirmed fixed in v0.4.19 — `graphify vscode install` is fully wired up. Please upgrade: `pip install --upgrade graphifyy`

PR #1138, opened 2026-06-05
docs: add troubleshooting table for common install issues

hyls9527:
## Summary
- Add a compact troubleshooting table to the README Install section
- Covers the most common issues reported in issues: wrong pip path, missing PATH, PowerShell slash syntax, Codex PreToolUse hook failure

## Changes
- Added `### Troubleshooting` section after the existing install notes in README.md
- 5 rows covering the top recurring symptoms and their fixes

## Motivation
New users frequently hit the same issues and file duplicate issues. This adds a self-service troubleshooting section so they can resolve common problems without waiting for maintainer response.

Addresses recurring issues: #39, #178, #356, #374, #533, #651, #831

### DIFF
--- a/README.md
+++ b/README.md
@@ -121,6 +121,16 @@ for example `graphify claude install --project` or `graphify codex install --pro
 
 > **Git hooks and uv tool / pipx:** `graphify hook install` embeds the current interpreter path directly into the hook scripts at install time, so the post-commit hook fires correctly even in GUI git clients and CI runners where `~/.local/bin` is not on PATH. If you reinstall or upgrade graphify, re-run `graphify hook install` to refresh the embedded path.
 
+### Troubleshooting
+
+| Symptom | Fix |
+|---------|-----|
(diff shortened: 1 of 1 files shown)


safishamsi (commented):
Appreciate the intent — duplicate issues from common install failures are a real problem. Two things to address:

**1. This content already exists in the README**

The inline notes at lines 116–122 cover PowerShell slash, `~/.local/bin`, and `ModuleNotFoundError`. There's also an existing `## Troubleshooting` section (around line 391) that covers `command not found`, `python -m graphify`, and PowerShell. Adding a third table would create three separate places a user has to check. Please fold the new rows into the existing `## Troubleshooting` section instead.

**2. The temp-files row is stale**

PR #1139 fixes the orphaned temp-file bug in code, so a manual workaround row in the README will contradict fixed behaviour as soon as #1139 lands. Either drop this row entirely or hold it until it's clear #1139 won't be merged.

PR #1139, opened 2026-06-05
fix: clean up leftover temp files from crashed rebuilds

hyls9527:
## Summary
- Add cleanup of stale `.graph.tmp.json` files at the start of `_rebuild_code`
- Prevents orphaned temp files when a previous rebuild crashes or is killed

## Changes
- Added cleanup loop at the beginning of `_rebuild_code()` in `graphify/watch.py`
- Removes any `.graph.tmp.json` files found in `graphify-out/` before starting a new rebuild

## Motivation
When `graphify update` or a hook-driven rebuild crashes (e.g. OOM, timeout, user kill), the `.graph.tmp.json` file created during the rebuild can be left behind. This file can confuse subsequent rebuilds or be mistaken for stale project files. The cleanup ensures a clean slate on every rebuild.

Addresses reports of orphaned temp files: #831

### DIFF
--- a/graphify/watch.py
+++ b/graphify/watch.py
@@ -449,16 +449,6 @@ def _rebuild_code(
 
     watch_root = watch_path.resolve()
     project_root = Path.cwd().resolve() if not watch_path.is_absolute() else watch_root
-
-    # Clean up any leftover temp files from a previous crashed rebuild.
-    # These are created by _rebuild_code (e.g. .graph.tmp.json) and should
-    # be removed on success, but a crash/kill can leave them behind.
-    if out.exists():
-        for stale in out.glob(".graph.tmp.json"):
-            try:
-                stale.unlink()
-            except OSError:
-                pass
     report_root = _report_root_label(watch_path)
     try:
         from graphify.extract import extract, _get_extractor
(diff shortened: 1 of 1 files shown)


safishamsi (commented):
Good intent — orphaned temp files from crashed rebuilds are a real problem. Two issues to fix before merge:

**1. Race condition — cleanup must run inside the lock**

`_rebuild_code` recurses: the outer call acquires the flock, then re-enters with `acquire_lock=False` to run the actual body. The cleanup needs to be placed **inside the locked region** (after the lock is held, alongside the temp-file creation at line 673), not at the very top of `_rebuild_code` before the lock is acquired. As written, a concurrent rebuild process could have the temp file actively open for writing between lines 674–706 while a second invocation cleans it up — a narrow but real race that could corrupt an in-progress rebuild.

**2. Simplify: replace the glob loop with `unlink(missing_ok=True)`**

```python
# Current (works but roundabout):
if out.exists():
    for stale in out.glob(".graph.tmp.json"):
        try:
            stale.unlink()
        except OSError:
            pass

# Cleaner (also no longer needs the out.exists() guard):
try:
    (out / ".graph.tmp.json").unlink()
except OSError:
    pass
```

The glob with a literal filename is just a single-file existence check dressed up as a loop. The direct form is clearer and consistent with the `except OSError: pass` pattern already used elsewhere in `watch.py` (e.g. line 147).

PR #1140, opened 2026-06-05
docs: add MCP server diagnostic section to skill

hyls9527:
Add diagnostic section to skill.md for verifying MCP server status when queries return empty results.

### DIFF
--- a/graphify/skill.md
+++ b/graphify/skill.md
@@ -576,6 +576,14 @@ if [ ! -f graphify-out/.graphify_python ]; then
 fi
 ```
 
+## MCP server diagnostic
+
+If the user reports that `graphify query`, `graphify path`, or `graphify explain` returns empty or generic results (not grounded in the actual graph), the MCP stdio server may not be loaded. Suggest:
+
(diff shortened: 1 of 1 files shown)


safishamsi (commented):
The placement is sensible but there are two problems that need fixing before merge.

**1. `graphify --mcp` does not exist**

The suggested verification command will error. The MCP stdio server lives in `graphify/serve.py` and is launched as:
```
```
Please replace `graphify --mcp` with this.

**2. The diagnosis points the AI at the wrong root cause**

The `graphify query` CLI subcommand reads `graphify-out/graph.json` directly — it does not depend on the MCP stdio server being loaded. So if a user reports that `graphify query` returns empty or generic results, an unloaded MCP server is almost never the cause. The real causes are: (a) stale or missing graph (run `graphify update .`), (b) IDF seed mismatch on generic query terms, or (c) the graph not covering the files the user is asking about.

Framing MCP-server loading as the primary diagnostic sends the AI down a path that is both misleading and commands it to run a non-existent CLI flag. Please either restrict this section to MCP-specific workflows (i.e. when the assistant is configured to call MCP tools directly, not the CLI query path) or remove it and fold the root-cause guidance into the existing empty-results section.

PR #1141, opened 2026-06-05
fix(extract): recover <script> imports from .vue Single-File Components

bymle:
## Summary

`.vue` files were dispatched to `extract_js`, which feeds the **whole** Single-File Component to the JavaScript tree-sitter grammar. A `.vue` file is not valid JS — it is `<template>`/`<script>`/`<style>` — so the parser produces a top-level `ERROR` node and the `import_statement` nodes inside `<script>`/`<script setup>` are never reached. The imports are silently dropped, leaving `.vue` files as disconnected nodes in the knowledge graph.

This is the **same problem already solved for `.svelte`** (#713): the JS grammar fed a full `.svelte` file also produces an ERROR root, so `extract_svelte` recovers imports with a regex pass over each `<script>` block (plus dynamic `import()`). The maintainer awareness is already in the code — `_get_extractor`/symbol-resolution treats `.vue` as non-JS at `extract.py` (`path.suffix != ".vue"` guard) — but the main extraction path still fed the whole SFC to the JS grammar.

## Fix

Add `extract_vue`, which reuses the Svelte SFC recovery (`<script>`-block static imports + dynamic `import()`) since a Vue SFC has the identical structure, and route `.vue` to it.

## Verification

Added `tests/test_vue_extraction.py` (modeled on `tests/test_astro_extraction.py`): a `.vue` SFC whose `<script setup>` statically imports a sibling `.vue` component is extracted through the real dispatch (`_get_extractor`) and the import edge is asserted. Verified it fails before the fix (the import is dropped) and passes after. `tests/test_astro_extraction.py`, `tests/test_js_import_resolution.py`, and `tests/test_import_extension_resolution.py` remain green (the fix is additive — `extract_svelte` is unchanged); `ruff check` clean.

### DIFF
--- a/graphify/extract.py
+++ b/graphify/extract.py
@@ -3752,18 +3752,6 @@ def extract_svelte(path: Path) -> dict:
     return result
 
 
-def extract_vue(path: Path) -> dict:
-    """Extract imports from .vue Single-File Components.
-
-    A ``.vue`` SFC has the same ``<template>``/``<script>``/``<style>`` structure as a
-    ``.svelte`` file, so tree-sitter-javascript fed the whole file produces a top-level
-    ERROR node and the ``import_statement`` nodes inside ``<script>``/``<script setup>``
-    are never reached and silently dropped (#713). The Svelte recovery (static imports in
-    each ``<script>`` block plus dynamic ``import()``) applies identically, so reuse it.
-    """
-    return extract_svelte(path)
-
-
 def extract_astro(path: Path) -> dict:
     """Extract imports from .astro files: frontmatter (TS) + template regex fallback.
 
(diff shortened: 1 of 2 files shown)


safishamsi (commented):
Good bug fix — the diagnosis is correct and reusing `extract_svelte` is the right call since it already does regex rescue of `<script>` blocks rather than relying on tree-sitter.

Two things before merge:

**1. Drop the wrapper function — dispatch directly**

```python
# Instead of:
def extract_vue(path: Path) -> dict:
    return extract_svelte(path)
# ...
".vue": extract_vue,

# Prefer:
".vue": extract_svelte,
```

The wrapper adds a stack frame and a maintenance point with no benefit. If you anticipate Vue-specific behaviour diverging from Svelte in future, a comment at the dispatch entry is enough to signal that.

**2. Beef up the test coverage**

The current test only covers a `<script setup>` static import. Please add:
- A `dynamic import()` case (e.g. `const m = await import('./utils.js')` inside a `<script>`)
- A `lang="ts"` attribute case (TypeScript SFC — the regex matches `<script[^>]*>` so it should work, but worth asserting)

Also worth noting: `extract.py:7256` explicitly excludes `.vue` from the JS symbol-resolution bypass (`path.suffix != ".vue"` guard). This is intentional and unaffected by your change, but a one-line comment at the dispatch entry would help future readers understand why `.vue` and `.svelte` are routed identically here but differ elsewhere in the pipeline.

PR #1149, opened 2026-06-06
Add GitHub Actions workflow for Python application

FolatheDuckofDuckingburg:
This workflow installs Python dependencies, runs linting with flake8, and executes tests using pytest for the 'v8' branch.


safishamsi:
Closing — this is the unmodified GitHub 'Python application' starter template. It duplicates the existing `ci.yml` test job without installing deps via uv (the project's actual toolchain), so it would never run correctly. The security scanning it references is better handled by #1209, which was just merged.

FolatheDuckofDuckingburg:
Oh, okay.

PR #1160, opened 2026-06-06
fix: use .codex config dir for codex platform

matiasduartee:
Fixes an issue where the skill was incorrectly being copied to \.agents\ instead of \.codex\ for the codex platform. This aligns with the \_install_codex_hook\ which correctly uses \.codex/hooks.json\.

### DIFF
--- a/graphify/__main__.py
+++ b/graphify/__main__.py
@@ -415,7 +415,7 @@ _PLATFORM_CONFIG: dict[str, dict] = {
     },
     "codex": {
         "skill_file": "skill-codex.md",
-        "skill_dst": Path(".agents") / "skills" / "graphify" / "SKILL.md",
+        "skill_dst": Path(".codex") / "skills" / "graphify" / "SKILL.md",
         "claude_md": False,
         "skill_refs": "codex",
     },
--- a/tests/test_install.py
+++ b/tests/test_install.py
@@ -8,7 +8,7 @@ import pytest
 
 PLATFORMS = {
     "claude": (".claude/skills/graphify/SKILL.md",),
-    "codex": (".agents/skills/graphify/SKILL.md",),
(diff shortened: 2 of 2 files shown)


safishamsi (commented):
The fix is correct — Codex's hook already writes to `.codex/hooks.json`, so the skill should also live under `.codex/`. The change propagates cleanly through install and uninstall because both resolve the path dynamically via `_PLATFORM_CONFIG["codex"]["skill_dst"]` — no hardcoded paths to update there.

**However the PR is missing test updates and CI will fail.** `tests/test_install.py` has four assertions that still reference `.agents` for the codex platform — all need to be updated to `.codex`:

- Line 11: `"codex": (".agents/skills/graphify/SKILL.md",),` → `.codex/...`
- Line 44: `assert (tmp_path / ".agents" / "skills" / "graphify" / "SKILL.md").exists()` → `.codex/...`
- Lines 90 and 93: `test_install_project_codex_writes_skill_and_agents` asserts the skill under `.agents/`

Note: the Amp and Antigravity tests that legitimately use `.agents/` should NOT be changed — only the codex-specific assertions.

Add those four test fixes and this is good to merge.

matiasduartee:
Thanks for the quick review, @safishamsi 

You are completely right. I've just pushed a new commit updating those four test assertions to point to .codex/.

While at it, I also noticed that the test_codex_subcommand_project_install_and_uninstall_are_project_scoped and test_uninstall_project_removes_project_skill_only tests were also asserting against the .agents path for the Codex platform, so I took the liberty of proactively updating those as well to prevent the CI from failing on them.

The CI should be green now! Let me know if anything else is needed.

safishamsi:
Landed in f146be3. All codex test assertions updated (.agents → .codex). Applied via 3-way merge to avoid reverting the postgres feature that landed after your branch was cut. Thanks!

Issue #569, opened 2026-04-27
Feature request: scoped resolution for explain/query/path (monorepos, minified bundles, name collisions)


DevangML:
## Environment

- **graphifyy:** PyPI `graphifyy` (CLI `graphify`); version pinned at time of report e.g. 0.5.x
- **Python:** 3.10+
- **Corpus:** Large multi-root workspace: multiple **Frappe** benches (same app names under different roots), **frappe** / **erpnext** trees, and generated assets under **`sites/assets/`** (including `*.min.js`).
- **Command:** `graphify update .` from repository root

## Problem

For this corpus shape, the **defaults** for `explain`, `graphify query`, and `graphify path` are easy to misread:

1. **`explain "<label>"`** often attaches to the **wrong** node for short or overloaded names — e.g. symbols inside **minified** `sites/assets/**/*.min.js`, or a **test** class before the **`Document`** controller for the same name.
2. **`query`** (BFS) surfaces **high-degree framework hubs** (`get`, `get_doc`, `Document`, ERPNext-heavy nodes) before app-specific nodes unless **`--budget`** is very large.
3. **`path "A" "B"`** often returns **no path** for informal strings; disambiguation is needed when the **same relative path** exists under multiple bench roots.

`graph.json` itself still contains the right **per-file** nodes and many **EXTRACTED** edges; the pain is mostly **resolution / traversal ergonomics** and **signal-to-noise** at the CLI, not “missing data” in the graph.

## Workarounds (already supported — README)

These reduce noise **before** any new flags:

- **`.graphifyignore`** (gitignore-style) — e.g. `**/sites/assets/`, `**/*.min.js`; optionally exclude `frappe/` / `erpnext/` when analysis targets custom apps only.
- **Narrow scan root** — `graphify update <subdir>` (e.g. one `frappe-bench/apps/` tree).
- **`merge-graphs`** — combine smaller graphs when a single full-root graph is too hub-heavy.

This issue is for cases where teams still want a **wide** index or need **better disambiguation** without hand-editing `graph.json`.

## Proposal

| Idea | Rationale |
|------|-----------|
| **`explain --file <path> [--line N]`** and/or **top-k matches** (each with `source_file`, line, **node id**) | Avoids a single silent wrong match; matches how power users already use **stable ids** from `graph.json`. |
| **`query --seed-file <path>`** or **`--path-prefix <prefix>`** | Starts BFS from nodes under a subtree (one bench / one app), reducing hub flooding. |
[...]

PR #1165, opened 2026-06-07
fix: resolve path/query endpoints by full-label match in _score_nodes

wiz0floyd:
## Problem

`graphify path "A" "B"` (and the MCP `shortest_path` tool) resolve each endpoint with `_score_nodes` in `graphify/serve.py`. Its exact-match tier compares a **single query token** to the **whole node label**:

```python
for t in norm_terms:
    if t == norm_label or t == bare_label:   # one token vs the entire label
        score += _EXACT_MATCH_BONUS * w
```

For a multi-word query no single token ever equals the full label, so the exact bonus never fires. Every node sharing the query's token set scores identically; the tie breaks by arbitrary node-id sort, so `path` picks a wrong or disconnected endpoint and reports `No path found` (or returns a path to the wrong node) even when the two labels are one hop apart.

`explain` is unaffected because it uses `_find_node`, which joins the query into one string and matches the whole label (exact → prefix → substring). `_score_nodes` had no equivalent full-query tier.

### Repro

Against a real graph with the unique labels `MC Delta Stress — air EMC below wood filtered MC` and `UOCE: Dehumidifier Driver` (one `references` edge apart):

```
$ graphify path "MC Delta Stress — air EMC below wood filtered MC" "UOCE: Dehumidifier Driver"
No path found
```

`explain "MC Delta"` resolves the same node and shows the edge — so the graph is fine; the resolver is at fault.

## Fix

Add a full-query tier to `_score_nodes` that compares the joined query string to the whole label, complementing (not replacing) the existing per-token tiers:

- A query equal to the label gets a dominant bonus (`_EXACT_MATCH_BONUS * 10 * idf`) so a full-label exact match outranks any token-bag sum; an equal-prefix gets the scaled prefix bonus.
- The comparison is done against the label's **tokenized** form (`_search_tokens(label)`), not only the stored `norm_label`. A node's `norm_label` can still carry punctuation (`:`, `—`) that a `\w+`-tokenized query never equals, so comparing against the raw `norm_label` alone is a silent no-op for punctuated labels like `UOCE: Dehumidifier Driver`. The joined query is also matched against `norm_label`, `bare_label`, and the node id, so all existing forms still hit.
- The final sort gains a length tiebreak so a concise exact label beats a longer superset on equal score.

Net effect: `path`/`query` now resolve the same node `explain` already does.

```
$ graphify path "MC Delta Stress — air EMC below wood filtered MC" "UOCE: Dehumidifier Driver"
Shortest path (1 hops):
  MC Delta Stress — air EMC below wood filtered MC <--references [EXTRACTED]-- UOCE: Dehumidifier Driver
```

No new dependencies; `uv.lock` unchanged.

## Test

Adds `test_score_nodes_multiword_exact_label_outranks_superset` to `tests/test_serve.py`. It reproduces the real condition (a node whose `norm_label` retains punctuation) and asserts a multi-word exact label scores strictly above a longer superset and a decoy, resolving uniquely. `tests/test_serve.py` is green (53 passed).

## Notes

- **Not a duplicate of #879.** That PR adds a per-token exact bonus (`any(t == label for t in terms)`), which fixes only single-word labels and targets `main`; this targets `v8` and addresses the multi-word / full-query gap, which `v8`'s existing per-token exact tier still has. Adjacent context: #569 (scoped resolution).
- The `label_tokens` comparison adds one `_search_tokens` regex call per node per `_score_nodes` query — O(nodes), negligible at these graph sizes. A maintainer who wants it lazier could hoist it behind the cheaper `norm_label`/`bare_label` checks; happy to adjust.

### DIFF
--- a/graphify/serve.py
+++ b/graphify/serve.py
@@ -136,11 +136,38 @@ def _score_nodes(G: nx.Graph, terms: list[str]) -> list[tuple[float, str]]:
     scored = []
     norm_terms = [tok for t in terms for tok in _search_tokens(t)]
     idf = _compute_idf(G, norm_terms)
+    # Whole-query string for full-label matching (mirrors _find_node's `term`).
+    joined = " ".join(norm_terms)
+    # Weight the full-query bonus by the rarest constituent term so a specific
+    # multi-word label still outweighs common-token noise; floor at 1.0.
+    joined_w = max((idf.get(t, 1.0) for t in norm_terms), default=1.0)
     for nid, data in G.nodes(data=True):
         norm_label = data.get("norm_label") or _strip_diacritics(data.get("label") or "").lower()
         bare_label = norm_label.rstrip("()")
+        # Tokenized form of the label (punctuation stripped, same transform as the
+        # query). norm_label may still carry punctuation like ':' or '-', which a
+        # tokenized query can never equal; comparing token-joined forms on both
+        # sides makes "uoce: dehumidifier driver" match query "uoce dehumidifier
+        # driver".
+        label_tokens = " ".join(_search_tokens(data.get("label") or ""))
         source = (data.get("source_file") or "").lower()
         score = 0.0
+        # Full-query tier: a multi-word query that equals (or prefixes) the whole
+        # label must dominate the per-token bag-of-words sums below, so `path`/
+        # `query` resolve the same node `explain` does (via _find_node). Without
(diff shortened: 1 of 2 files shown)


safishamsi:
Landed in a8dbbe5. Full-query join added alongside per-token comparison in _score_nodes. O(1) per node, fixes multi-word endpoint resolution for graphify path and shortest_path. Thanks!

PR #1166, opened 2026-06-07
docs: add RFC for file-level node summaries

leonaburime-ucla:
## Summary

Adds an RFC-style doc for optional file-level node summaries aimed at AI coding-agent navigation.

The doc compares two storage options for community/maintainer feedback:

- inline optional `summary` attributes on file-level nodes in `graph.json`
- a sidecar `node-summaries.json` artifact keyed by node ID

It also proposes shared constraints for either implementation: file-level only in v1, deterministic/offline generation first, no default LLM/API calls, bounded summary length, and surfacing summaries through `graphify explain` and serve/MCP node lookup.

## Why

Agents can use graph structure to avoid reading a whole repo, but they may still need to inspect raw files just to understand what a file/node is responsible for. A compact summary could reduce repeated context reads while preserving graphify's local-first behavior.

## Tests

Not run; docs-only change.

### DIFF
--- a/docs/how-it-works.md
+++ b/docs/how-it-works.md
@@ -88,6 +88,9 @@ The output `graph.json` uses NetworkX's node-link format. Each node has:
 - `file_type` — `code`, `document`, `paper`, `image`, `rationale`
 - `source_file` — where it came from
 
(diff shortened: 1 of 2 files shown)


anupddas (approved):
Looks all okay.

Issue #920, opened 2026-05-17
Sensitive filename detection misses underscore-separated names (\b boundary)


chriscbradley76:
## Issue

`_SENSITIVE_PATTERNS` in `graphify/detect.py` uses `\b` word boundaries. Python's `\b` treats `_` as a word character (per `[A-Za-z0-9_]`), so the regex `\b(credential|secret|passwd|password|token|private_key)s?\b` does **not** match common secret filenames that use underscores:

- `api_token.txt`
- `session_token.txt`
- `secret_access_key.txt`
- `aws_credentials.txt`
- `private_key.pem` _(saved by `.pem` extension regex, but the wordboundary one misses it)_

## Reproduction

```python
import re
p = re.compile(r'\b(credential|secret|passwd|password|token|private_key)s?\b', re.IGNORECASE)
print(p.search('api_token.txt'))         # None — should match
print(p.search('session_token.txt'))     # None
print(p.search('secret_access_key.txt')) # None
```

## Real-world impact

In a 1100-file repo audit, `_is_sensitive` caught **0/101** files with token/key/secret patterns in their name. Most were legitimate code files (`iam_token_exchange.py`, `update_api_token.py`), but 5 were actual credentials under `.secrets/` directories that slipped through `detect()` filtering and would have been read by semantic extraction subagents.

Local mitigation via `.graphifyignore` works but requires every user to know about this gap.

## Proposed fix

Two-part:

**1. Replace `\b` with lookarounds that treat `_` as a separator:**

```python
re.compile(
    r'(?<![A-Za-z0-9])(credential|secret|passwd|password|token|private_key)s?(?![A-Za-z0-9])',
    re.IGNORECASE,
)
```

This still allows letter/digit boundaries (so `MyToken.java` matches) but also catches `_token_`, `.token.`, `-token-`, etc.

**2. Add directory-level filter (covers the credentials-by-location case):**

```python
_SENSITIVE_DIRS = {".secrets", "secrets", ".aws", ".gnupg", ".ssh"}

def _is_sensitive(path: Path) -> bool:
    if any(part in _SENSITIVE_DIRS for part in path.parts):
        return True
    return any(p.search(path.name) for p in _SENSITIVE_PATTERNS)
```

## Side-effect to consider

The stricter regex would now flag legitimate code files like `iam_token_exchange.py` as sensitive. Two options:

- Restrict filename-based filter to explicit credential file extensions only (`.pem`, `.key`, `.txt` under a sensitive dir)
[...]

PR #1169, opened 2026-06-07
fix: _is_sensitive silently drops topic notes (token-economics-of-recall.md flagged as a secret)

edudatsuj45:
## Problem

The generic keyword patterns in `_SENSITIVE_PATTERNS` flag any filename containing `token`/`secret`/`password`/`credential` as a standalone word — so prose documents whose descriptive slug merely *mentions* the topic are silently dropped from the graph:

| Filename | Current | Reality |
|---|---|---|
| `token-economics-of-recall.md` | skipped as sensitive | a note **about** LLM token costs |
| `password-policy-discussion.md` | skipped as sensitive | a design discussion |

This is the remaining failure class after the #436 → #718 → #920 lineage. Hit in the wild: an Obsidian-style memory vault where a note on token economics vanished from the graph with no visible warning (`skipped_sensitive` is returned but nothing surfaces it — exactly the silent-data-loss failure mode described in #718's closing observation).

## Fix

Split the two generic keyword patterns out of `_SENSITIVE_PATTERNS` and only count a match when the keyword is **load-bearing** in the filename:

- the keyword **ends the stem** (`api_token.txt`, `oauth_token.json`, `github-personal-access-token.txt`) — secret stores name their contents, and the content noun is the head of the compound, which comes last in English; **or**
- the stem has **≤ 2 words** (`token.txt`, `token_config.yaml`, `secret_handler.txt`).

A keyword buried mid-phrase in a ≥ 3-word slug is a topic word, not a credential store. The specific patterns (`.pem`/`.env`/`id_rsa`/`.netrc`/`aws_credentials`/`_SENSITIVE_DIRS`) are unchanged and still always apply.

The end-of-stem check runs **before** word counting, so multi-word keywords survive their own separator: `my_private_key.txt` is still flagged even though splitting on `_` would break `private_key` apart. Leading dots are stripped before stem extraction so dotfiles like `.token` keep their keyword.

## Behavior table

| Filename | Before | After |
|---|---|---|
| `token-economics-of-recall.md` | flagged ❌ | clean ✅ |
| `password-policy-discussion.md` | flagged ❌ | clean ✅ |
| `api_token.txt` / `oauth_token.json` (#920) | flagged | flagged |
| `token.txt` / `tokens.txt` / `.token` | flagged | flagged |
| `token_config.yaml` / `secret_handler.txt` (#920) | flagged | flagged |
| `github-personal-access-token.txt` | flagged | flagged |
| `my_private_key.txt` | flagged | flagged |
| `passwords.py` / `credentials.json` | flagged | flagged |
| `tokenizer.py` / `tokenize.py` (#718) | clean | clean |
| `.env` / `server.pem` / `id_rsa` / `.ssh/…` | flagged | flagged |

## Tests

All 13 existing `test_sensitive_*` contracts pass unchanged; adds 6 regression tests covering the topic-slug false positive plus the dotfile, plural, end-of-long-name, and multi-word-keyword edge cases.

```
tests/test_detect.py: 19 passed (sensitive subset)
tests/test_detect.py: 109 passed, 6 failed* , 1 skipped
* pre-existing WinError 1314 symlink-privilege failures on Windows without Developer Mode, unrelated to this change
```

### DIFF
--- a/graphify/detect.py
+++ b/graphify/detect.py
@@ -98,22 +98,59 @@ _SENSITIVE_DIRS = frozenset({
     ".ssh", ".gnupg", ".aws", ".gcloud", "secrets", ".secrets", "credentials",
 })
 
-# Files that may contain secrets - skip silently.
+# Files that may contain secrets - skip silently. These patterns are specific
+# (extensions, exact credential-store names) and always apply.
+_SENSITIVE_PATTERNS = [
+    re.compile(r'(^|[\\/])\.(env|envrc)(\.|$)', re.IGNORECASE),
+    re.compile(r'\.(pem|key|p12|pfx|cert|crt|der|p8)$', re.IGNORECASE),
+    re.compile(r'(id_rsa|id_dsa|id_ecdsa|id_ed25519)(\.pub)?$'),
+    re.compile(r'(\.netrc|\.pgpass|\.htpasswd)$', re.IGNORECASE),
+    re.compile(r'(aws_credentials|gcloud_credentials|service.account)', re.IGNORECASE),
+]
+
+# Generic keyword patterns - these only count when the keyword is LOAD-BEARING
+# in the filename (see _generic_keyword_hit), because a keyword buried mid-phrase
+# in a long descriptive slug names a topic, not a credential store:
+# "token-economics-of-recall.md" is a note ABOUT tokens; "api_token.txt" IS one.
 # Uses lookarounds instead of \b so underscore-prefixed names like api_token.txt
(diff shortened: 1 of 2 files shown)


safishamsi:
Landed in a8dbbe5. Generic keywords (token/secret/password) now only fire when the keyword ends the filename stem or the stem is ≤2 words. Specific patterns (.env/.pem/id_rsa etc.) remain unconditional. token-economics-of-recall.md passes through; tokens.txt is still caught. Thanks!

PR #1170, opened 2026-06-07
fix(hooks): detach background rebuild via Python instead of nohup (#1161)

### LINKED ISSUE #1161: post-commit/post-checkout hooks use `nohup`, which Git for Windows' shell lacks — rebuild fails silently
## Summary
  The git hooks installed by `graphify hook install` background the graph rebuild with `nohup ... &`. **Git for Windows' bundled shell (MSYS/MINGW64) does not ship `nohup`** (nor
  `setsid`), so the line dies with `nohup: command not found` and the rebuild never runs. Worse, it fails **silently**: `git commit` / `git pull` return success, no error surfaces to
  the user, and the knowledge graph just quietly goes stale.

  ## Environment
  - OS: Windows 11
  - Shell running the hooks: `C:\Program Files\Git\usr\bin\sh.exe` (Git for Windows, `MSYSTEM=MINGW64`)
  - `git version 2.42.0.windows.2`
  - graphify installed under conda Python; `graphify hook install` reported success and wrote `post-commit` + `post-checkout`.

  ## Steps to reproduce
  1. On Git for Windows, run `graphify hook install` in a repo with an existing `graphify-out/`.
  2. `git commit` (or pull a branch that fires `post-checkout`).
  3. Observe the rebuild log at `~/.cache/graphify-rebuild.log`.

  ## Expected
  The graph rebuilds in the background and `graph.json` / `GRAPH_REPORT.md` update.

  ## Actual
  The log contains:
  ```
  .git/hooks/post-commit: line 101: nohup: command not found
  ```
  No rebuild happens. The commit/pull exits 0, so the user gets no signal that the graph is now stale. The same flaw is in the `post-checkout` hook.

  ## Root cause
  The hook templates use:
  ```sh
  nohup "$GRAPHIFY_PYTHON" -c "...rebuild..." >> "$_GRAPHIFY_LOG" 2>&1 < /dev/null &
  disown 2>/dev/null || true
  ```
  `nohup` is a coreutils/util-linux tool that is absent from the minimal MSYS2 runtime Git for Windows bundles. `setsid` is also absent, so it isn't a drop-in substitute. There's no
  package manager (`pacman`) in a stock Git for Windows install to add it either.

  ## Suggested fix
  graphify already requires Python, so let Python do the detaching instead of `nohup`/`setsid` — this is cross-platform and dependency-free. A tiny launcher spawns the real rebuild as
  a fully detached process and returns immediately:

  ```python
  import os, sys, subprocess

  cmd = [sys.executable, "-c", REBUILD_SRC]
  kwargs = dict(stdout=open(LOG, "a", buffering=1),
                stderr=subprocess.STDOUT,
                stdin=subprocess.DEVNULL,
                cwd=os.getcwd(), close_fds=True)

  if os.name == "nt":
      DETACHED_PROCESS = 0x00000008
      CREATE_NEW_PROCESS_GROUP = 0x00000200
      CREATE_BREAKAWAY_FROM_JOB = 0x01000000
      base = DETACHED_PROCESS | CREATE_NEW_PROCESS_GROUP
      try:
          subprocess.Popen(cmd, creationflags=base | CREATE_BREAKAWAY_FROM_JOB, **kwargs)
      except OSError:
          subprocess.Popen(cmd, creationflags=base, **kwargs)
  else:
      subprocess.Popen(cmd, start_new_session=True, **kwargs)   # POSIX setsid equivalent
  ```

  The hook then just runs `"$GRAPHIFY_PYTHON" -c "$LAUNCHER"` — no `nohup`, no `&`, returns instantly, and the detached child survives the hook (and its shell) exiting.

JOhnsonKC201:
## Summary

Fixes #1161.

The git hooks installed by `graphify hook install` background the graph rebuild with `nohup "$GRAPHIFY_PYTHON" -c "..." &`. **Git for Windows' bundled MSYS/MINGW shell ships no `nohup`** (nor `setsid`), so that line dies with `nohup: command not found` and the rebuild never runs. Worse, it fails **silently** — `git commit` / `git pull` still return `0`, so the knowledge graph quietly goes stale with no signal to the user. Both `post-commit` and `post-checkout` were affected.

## Fix

graphify already requires Python, so let Python do the detaching — portable and dependency-free. A small launcher spawns the real rebuild as a fully detached child and returns immediately, so the hook never blocks the commit/checkout:

- **POSIX:** `subprocess.Popen(..., start_new_session=True)` (the `setsid` equivalent)
- **Windows:** `DETACHED_PROCESS | CREATE_NEW_PROCESS_GROUP`, adding `CREATE_BREAKAWAY_FROM_JOB` when the job object allows it (falls back gracefully if not)

This follows the approach suggested in the issue.

### Implementation notes

- The two rebuild bodies are factored into shared constants (`_REBUILD_BODY_COMMIT`, `_REBUILD_BODY_CHECKOUT`) and embedded verbatim into a single `_LAUNCHER_TEMPLATE` via `_detached_launch()`, so the detach logic lives in one place.
- The launcher + body are carried inside a shell double-quoted `-c "..."` argument, so they deliberately use **only single-quoted Python strings** — no `"`, `$`, backtick or backslash that the shell would interpret. There are regression tests guarding exactly this.
- Child output still streams to `~/.cache/graphify-rebuild.log`. The log path is now exported as `GRAPHIFY_REBUILD_LOG` so the path the hook echoes matches where the detached child actually writes.

## Testing

- **End-to-end on Git for Windows' shell** (the exact environment from the issue): the parent returns instantly with **no `nohup: command not found`**, and the detached child runs independently and writes its progress to the log.
- `sh -n` validates both fully-rendered hook scripts as syntactically correct POSIX sh.
- Added 8 regression tests in `tests/test_hooks.py`:
  - generated hooks contain no `nohup` / `setsid` / `disown`
  - both hooks use the cross-platform detach (`start_new_session` + Windows creation flags)
  - the embedded `-c` payload is shell-quote-safe
  - both the launcher and the rebuild body it re-executes parse as valid Python
  - end-to-end: the files written to `.git/hooks` are nohup-free
- `tests/test_hooks.py` is fully green (31 passed) under the CI matrix interpreter (Python 3.12, `uv run --frozen --all-extras`).

### Test plan
- [x] `uv run --frozen --all-extras pytest tests/test_hooks.py -q` → 31 passed
- [x] `ruff check graphify/hooks.py --select E9,F63,F7,F82` → clean
- [x] `sh -n` on both rendered hooks → OK
- [x] Manual end-to-end launch through Git for Windows `sh`

### DIFF
--- a/graphify/hooks.py
+++ b/graphify/hooks.py
@@ -74,6 +74,110 @@ if [ -z "$GRAPHIFY_PYTHON" ]; then
 fi
 """
 
+# The Python that the rebuild runs, shared by both hooks. Embedded verbatim into
+# the launcher below and re-executed in the detached child. Must not contain the
+# double-quote, $, backtick or backslash characters: it is carried inside a
+# shell double-quoted `-c "..."` argument (see _detached_launch).
+_REBUILD_BODY_COMMIT = """\
+import os, signal, sys
+from pathlib import Path
+
+changed_raw = os.environ.get('GRAPHIFY_CHANGED', '')
+changed = [Path(f.strip()) for f in changed_raw.strip().splitlines() if f.strip()]
+
+if not changed:
+    sys.exit(0)
+
+print(f'[graphify hook] {len(changed)} file(s) changed - rebuilding graph...')
+
+try:
+    from graphify.watch import _rebuild_code, _apply_resource_limits
+    _apply_resource_limits()
+    _timeout = int(os.environ.get('GRAPHIFY_REBUILD_TIMEOUT', '600'))
+    if _timeout > 0 and hasattr(signal, 'SIGALRM'):
+        signal.signal(signal.SIGALRM, lambda *_: (_ for _ in ()).throw(TimeoutError(f'graphify rebuild exceeded {_timeout}s')))
+        signal.alarm(_timeout)
+    _force = os.environ.get('GRAPHIFY_FORCE', '').lower() in ('1', 'true', 'yes')
+    _rebuild_code(Path('.'), changed_paths=changed, force=_force)
+except TimeoutError as exc:
+    print(f'[graphify hook] {exc}')
+    sys.exit(1)
+except Exception as exc:
+    print(f'[graphify hook] Rebuild failed: {exc}')
+    sys.exit(1)
+"""
+
+_REBUILD_BODY_CHECKOUT = """\
+from graphify.watch import _rebuild_code, _apply_resource_limits
+from pathlib import Path
+import os, signal, sys
+try:
+    _apply_resource_limits()
+    _timeout = int(os.environ.get('GRAPHIFY_REBUILD_TIMEOUT', '600'))
+    if _timeout > 0 and hasattr(signal, 'SIGALRM'):
+        signal.signal(signal.SIGALRM, lambda *_: (_ for _ in ()).throw(TimeoutError(f'graphify rebuild exceeded {_timeout}s')))
+        signal.alarm(_timeout)
+    _force = os.environ.get('GRAPHIFY_FORCE', '').lower() in ('1', 'true', 'yes')
+    # post-checkout: branch switch can touch arbitrary files; full rebuild path
+    # (no changed_paths) is correct here. The flock inside _rebuild_code still
+    # prevents pile-ups when commit + checkout fire back-to-back.
+    _rebuild_code(Path('.'), force=_force)
+except TimeoutError as exc:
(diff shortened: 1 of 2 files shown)


safishamsi:
Landed in a8dbbe5. Cross-platform Python detach replaces nohup — DETACHED_PROCESS|CREATE_NEW_PROCESS_GROUP on Windows, start_new_session=True on POSIX. Quoting-safe via argv list. Fixes #1161. Thanks!

PR #1171, opened 2026-06-07
feat: add Augment Code support

itshoax:
Adds graphify augment install for Augment Code (VS Code, JetBrains, Auggie CLI).
Augment reads all .augment/rules/*.md files at the start of every Agent/Chat session — no trigger needed. The skill is written there. .augment-guidelines gets an idempotent graphify section for workspace-level discoverability.

`graphify augment install` — writes .augment/rules/graphify.md + .augment-guidelines
`graphify augment uninstall` — removes both cleanly
`graphify install --platform augment` — same as above
`graphify uninstall` covers Augment in the global sweep


itshoax:
@safishamsi Please have a look

Issue #1079, opened 2026-05-29
Graphify installs to wrong skills folder for Google Antigravity


jlmanohar:
# Graphify installs to wrong directory for antigravity

According to Google Antigravity Agent skills docs [Antigravity Agent Skills](https://antigravity.google/docs/skills), the skills can live in the below folder 
<doc-template _ngcontent-ng-c307066731="" _nghost-ng-c237084581=""><h2 _ngcontent-ng-c237084581="" class="heading-7 template-content-heading" id="where-skills-live">Where skills live</h2></doc-template><doc-template _ngcontent-ng-c307066731="" _nghost-ng-c237084581=""><p _ngcontent-ng-c237084581="" class="caption template-content-paragraph">Antigravity supports two types of skills:</p></doc-template><doc-template _ngcontent-ng-c307066731="" _nghost-ng-c237084581=""><p _ngcontent-ng-c237084581="" class="caption template-content-paragraph">
Location | Scope
-- | --
<workspace-root>/.agents/skills/<skill-folder>/ | Workspace-specific
~/.gemini/config/skills/<skill-folder>/ | Global (all workspaces)

</p></doc-template>

When running graphify install antigravity, it installed it in my ~/agents/skills/graphify/SKILL.md

Graphify version : graphify 0.8.24

Code ran and output 
```
jlmanohar@fedora:~/Documents/projects/CashewAI$ graphify install antigravity
  skill installed  ->  /home/jlmanohar/.agents/skills/graphify/SKILL.md

Done. Open your AI coding assistant and type:

  /graphify .
```
FYI - It did not create a local .agents folder in my project workspace, nor do i see /graphify skill picked up in antigravity

# Suggested solution :

For Antigravity install, update the gemini config folder since thats where skills live for antigravity

safishamsi:
Fixed in 9985940. Three root causes addressed:

1. **Wrong global path** — `graphify install antigravity` was writing to `~/.agents/skills/graphify/SKILL.md` (home + relative path). Per the Antigravity docs, global skills live at `~/.gemini/config/skills/`. Now correctly installs there.

2. **Uninstall couldn't find the skill** — `_antigravity_uninstall` was reading the path from the config dict (relative `.agents/skills/`) so the global skill was silently left behind after uninstall. Fixed to use the same resolved path as install, with proper project vs global scope awareness.

[...]

PR #1185, opened 2026-06-07
fix: gemini platform global install now writes to ~/.gemini/config/skills/

matiasduartee:
@safishamsi 
Fixes the issue where `graphify install --platform gemini` writes to `~/.agents/` or `~/.gemini/skills/` instead of `~/.gemini/config/skills/` globally. This aligns it with the Agent Kit and Antigravity conventions, matching the fix introduced in #1079.

**Additional changes in this PR:**
- Updated `antigravity` and `antigravity-windows` platforms to exclusively use the global `~/.gemini/config/skills/` path (ignoring `--project`), since the Antigravity CLI, IDE, and v2.0 all share this single global directory now.
- Resolved a merge conflict in `tests/test_install.py` caused by recent upstream changes (`codebuddy`).
- Updated the unit tests for `antigravity` install/uninstall to reflect the new global-only path behavior.


matiasduartee:
Hi @safishamsi ! I've updated the PR to resolve the merge conflicts and fix the failing tests.

Here is a summary of the fixes applied:

Merge Conflicts: Resolved conflicts in graphify/__main__.py to properly integrate the new agent platform structures.
Bug Fix (UnboundLocalError): Fixed a scoping issue in the uninstall process that caused skill_dst to be referenced before assignment when uninstalling from a project directory.
Test Paths Adjustments: Updated test_install_roundtrip.py and test_install_references.py. The assertions were assuming all project installations would be scoped under the project directory, which caused false negatives for platforms like gemini and antigravity (since they are designed to always use the global config path).
Test Isolation: Added a mock for the %LOCALAPPDATA% environment variable during tests on Windows so that the hermes platform tests run safely inside the tmp_path without leaking into the host's actual AppData folder.
The CI is now fully green! Let me know if there's anything else needed for this to be merged.

matiasduartee:
@safishamsi could you take a look at this one when you get a chance? 🙏

**Context — this reflects a real change on Google's side.** Since the launch of **Antigravity 2.0**, the whole Gemini/Antigravity ecosystem consolidated its global config under `~/.gemini/config/`:

- **Antigravity IDE**, the **Antigravity CLI**, and the **Gemini CLI** all read global skills from `~/.gemini/config/skills/` — on every OS, for every workspace.
- The previous `~/.gemini/skills/` path is **no longer scanned**, so a skill installed there is silently ignored.

I confirmed this on a live install (Windows): `~/.gemini/config/skills/` exists and holds the actually-loaded skills (agents, workflows, `config.json` all live under `~/.gemini/config/` too), while `~/.gemini/skills/` does not exist at all.

**What this PR does**
- Route `gemini` (and `antigravity` / `antigravity-windows`) global installs to `~/.gemini/config/skills/graphify/SKILL.md`, matching Google's current layout.
- Guard `_antigravity_uninstall` with `if not project:` so a project-scope uninstall never deletes the shared global skill out from under other workspaces.

I've rebased/merged on top of the latest `v8` (including the `install.py` extraction refactor), so it's up to date and the conflict is resolved. The current install path in `main`/`v8` points at a directory Google no longer reads, so users on Antigravity 2.0 end up with a skill that never loads. Happy to adjust naming/structure if you'd prefer a different shape. Thanks!

PR #1187, opened 2026-06-07
Add Common Lisp support

fade:
## What

Adds Common Lisp as a supported language. graphify now extracts a structural
knowledge graph from `.lisp`, `.cl`, `.lsp`, and `.asd` sources — packages, classes,
functions, methods, macros, generic-function dispatch, struct/type/condition
definitions, custom definer forms, reader-conditional and wrapper-macro bodies, and
call edges.

The grammar (`tree-sitter-commonlisp`) is wired as an optional extra
(`graphifyy[commonlisp]`), matching the policy for other niche grammars (`sql`, `dm`).

## Why

Common Lisp is a conspicuous gap for a class of users — Lisp/AllegroGraph shops,
scientific and systems CL projects — who otherwise can't graph their codebases.
Common Lisp is homoiconic: every `defun`/`defclass`/`defpackage` parses to the same
`list_lit` node, so it can't be expressed through the declarative `LanguageConfig`
(which keys on tree-sitter node types). It needs a dedicated walker that classifies
forms by their definer symbol — provided here as a standalone `extract_commonlisp`,
registered in `_DISPATCH` alongside the other custom extractors (`extract_sql`,
`extract_markdown`).

## Credit

The extractor and its test suite originate from @jansaasman's earlier
#172, which was opened against the pre-rewrite line and never merged.
This PR rebases that work onto v8, brings it into conformity with the
current extraction schema and contribution conventions, and adds the
integration. Every commit carries their co-authorship.

## What's included

- `extract_commonlisp` walker + `detect.py`/`_DISPATCH` registration for `.lisp/.cl/.lsp/.asd`
- `tree-sitter-commonlisp` as an optional extra (kept out of core dependencies)
- 23 extraction tests in `tests/test_languages.py` + `tests/fixtures/sample.lisp`
- README supported-extensions table (28 → 29 grammars) and the `commonlisp` extras row

## Reviewer notes

- **Rationale nodes:** docstring-derived nodes are emitted with `file_type="rationale"`
  so the cross-file resolution filter excludes them, consistent with the Python extractor.
- **Grammar deprecation:** `tree-sitter-commonlisp` 0.4.1 (latest) returns an int pointer
  that `tree_sitter.Language` only accepts via a deprecated path; that single warning is
  suppressed at the construction site with an explanatory comment. Glad to switch to a
  capsule-based construction if/when the grammar ships one.
- **Suite:** language, cache, and detect suites pass (411 passed, 22 skipped locally).

🄯 Brian O'Reilly <fade@deepsky.com>, 2026


fade:
Closing this in favour of #2274, which supersedes it.

This branch had gone stale and conflicting against v8. The replacement is rebased onto
current v8, and the extractor now lives in `graphify/extractors/commonlisp.py` following
`extractors/MIGRATION.md` and #1212, rather than adding to `extract.py`. It also keeps the
lineage from #172 by @jansaasman explicit.

Nothing is lost by closing this one; #2274 is the same feature, brought up to date.

PR #1202, opened 2026-06-09
fix: affected/benchmark crash on native "edges" graphs (KeyError: links)

zenwindowcleaning:
## Problem

`graphify affected <node>` and `graphify benchmark` crash with:

```
error: could not load graph: 'links'
```

on every real `graphify-out/graph.json`. Reproduce on any extracted graph:

```bash
graphify extract ./some-repo --backend claude-cli --out /tmp/g
graphify affected "SomeSymbol" --graph /tmp/g/graphify-out/graph.json
# -> error: could not load graph: 'links'
graphify benchmark /tmp/g/graphify-out/graph.json
# -> KeyError: 'links'
```

## Root cause

`affected.py:load_graph` and `benchmark.py:run_benchmark` call:

```python
try:
    json_graph.node_link_graph(raw, edges="links")
except TypeError:
    json_graph.node_link_graph(raw)
```

The extractor serialises edges under the **`"edges"`** key. Only merged /
cross-repo graphs use `"links"` (written via `node_link_data`, whose default
edges key is `"links"`). On a native graph, the `edges="links"` attempt raises
**`KeyError`**, which the `except TypeError` clause does not catch — so the load
aborts instead of falling through to the working default.

## Fix

Widen the fallback to `except (TypeError, KeyError)`. Native `"edges"` graphs
now fall through to the networkx default; merged `"links"` graphs still load via
the fast path. Both formats verified.

## Why CI didn't catch it

`tests/test_affected_cli.py` builds its fixture with
`json_graph.node_link_data(graph, edges="links")`, producing a `"links"`-keyed
graph that happens to match the buggy reader — so the tests pass while every
real graph crashes.

Added `test_affected_cli_loads_native_edges_key_graph`, which writes the native
`"edges"` format the extractor actually emits. It **fails without this fix** with
the exact production error (`could not load graph: 'links'`) and passes with it.

```
$ pytest tests/test_affected_cli.py -q
....                                                                     [100%]
4 passed
```

### DIFF
--- a/graphify/affected.py
+++ b/graphify/affected.py
@@ -148,7 +148,11 @@ def load_graph(path: Path) -> nx.Graph:
     # Force directed so stored caller→callee direction survives the round-trip;
     # mirrors serve.py and __main__.py (#1174).
     raw = {**raw, "directed": True}
+    # Merged/cross-repo graphs serialise edges under "links" (node_link_data
+    # default); native extractor graphs use "edges". Try "links" first and fall
+    # back to the networkx default ("edges") — the KeyError raised on a native
+    # graph must be caught here, otherwise load fails on every real graph.json.
     try:
         return json_graph.node_link_graph(raw, edges="links")
-    except TypeError:
+    except (TypeError, KeyError):
         return json_graph.node_link_graph(raw)
--- a/graphify/benchmark.py
+++ b/graphify/benchmark.py
@@ -101,9 +101,12 @@ def run_benchmark(
     from graphify.security import check_graph_file_size_cap
     check_graph_file_size_cap(Path(graph_path))
     data = json.loads(Path(graph_path).read_text(encoding="utf-8"))
+    # Native extractor graphs store edges under "edges"; merged graphs use
+    # "links". Catch the KeyError from the "links" attempt so native graphs
+    # fall back to the networkx default instead of crashing.
     try:
         G = json_graph.node_link_graph(data, edges="links")
-    except TypeError:
+    except (TypeError, KeyError):
         G = json_graph.node_link_graph(data)
 
     if corpus_words is None:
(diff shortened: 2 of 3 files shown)


safishamsi (changes requested):
## Review: Request Changes

The core fix is correct — broadening `except TypeError` to `except (TypeError, KeyError)` in `affected.py` and `benchmark.py` resolves the crash. The `affected.py` regression test is solid. Two gaps before merging:

### 1. Identical unfixed sibling in `__main__.py:3403-3407`

The merge driver's local `_load_graph` has the exact same bug:

```python
try:
    return _jg.node_link_graph(data, edges="links"), data
except TypeError:
    return _jg.node_link_graph(data), data
```

This will crash with `KeyError: 'links'` when the git merge driver runs on a native-edges `graph.json`. Please fix here or explicitly scope it out with a follow-up issue.

### 2. No benchmark test

`benchmark.py` is fixed but untested. The existing `tests/test_benchmark.py` builds fixtures via `node_link_data(G, edges="links")` so it only exercises the `"links"`-format path and cannot catch a regression. Please add a benchmark test analogous to `test_affected_cli_loads_native_edges_key_graph`.

### Minor (non-blocking)
The rest of the codebase uses a pre-normalize pattern for this:
```python
if "links" not in data and "edges" in data:
    data = dict(data, links=data["edges"])
```
(see `serve.py:29-30`, `global_graph.py:34-35`, `__main__.py:2699-2700`). Aligning to this pattern is cleaner but not required.

safishamsi:
Closing in favour of #1280, which was merged. #1280 uses the established key-normalisation pattern (`raw = dict(raw, links=raw['edges'])`) already used in `__main__.py` and `serve.py`. The fallback approach here (`except (TypeError, KeyError)` then retry without `edges=`) still crashes on networkx 3.4.2 / Python <3.11 because the default `edges` param is also `'links'` in that version.

PR #1208, opened 2026-06-09
Pin tree-sitter dependency version ranges

nucleusjay:
All 27 `tree-sitter-*` dependencies in `pyproject.toml` are declared with no version specifier:

```toml
\"tree-sitter-python\",
\"tree-sitter-javascript\",
...
```

`uv.lock` covers `uv sync` users, but the README's primary install path (`pip install graphifyy` / `uv tool install graphifyy`) does not consult the lock and resolves whatever PyPI currently serves. A breaking minor bump in any of the 27 grammar packages -- and the 0.x ones break often -- can land in user installs without notice.

This PR adds explicit lower and upper bounds matching the current `uv.lock`:

- Lower bound = currently-locked version
- Upper bound = next minor for stable packages, next major for 0.x packages with frequent breakage

Patch updates still flow through automatically; incompatible major/minor jumps now require a deliberate bump.

## Test plan
- [ ] `uv sync --frozen` still works (no lock churn).
- [ ] `uv lock --check` passes (lock is consistent with the new constraints).
- [ ] Spot-check a fresh `pip install graphifyy` resolves cleanly.


safishamsi (changes requested):
Good call on this — unpinned tree-sitter grammars are a real silent-break risk on `pip install graphifyy` (which doesn't read `uv.lock`). I cross-checked all 26 bounds against the locked versions in `uv.lock` and they're all correct: every lower bound ≤ locked version, every upper bound > locked version. The width choices look sensibly tuned per-package too.

## One gap: 3 optional-extra grammars remain unpinned

The same silent-break risk exists for grammars in the optional extras block, which this PR doesn't touch:

- `tree-sitter-sql` (extra `sql`, line ~68) — no specifier
- `tree-sitter-dm` (extra `dm`, line ~72) — no specifier  
- `tree-sitter-hcl` (extra `terraform`, line ~73) — no specifier (note: it *is* pinned in the `dev` group already, so the locked version is available for reference)

A user who installs `pip install graphifyy[sql]` or `graphifyy[terraform]` hits the exact vulnerability this PR addresses. Please apply the same lower/upper bound scheme to these three and I'll approve.

PR #1209, opened 2026-06-09
Wire bandit and pip-audit into CI

nucleusjay:
`bandit`, `pip-audit`, and `safety` are already declared in the `[dependency-groups].dev` table but nothing in `.github/workflows/ci.yml` invokes them. So a new HIGH-severity bandit finding or a freshly-disclosed CVE in a pinned dep can land without anyone noticing until the next manual audit.

Adds a `security-scan` job that runs on every push and PR:

- `bandit -r graphify -ll` (HIGH-severity only -- LOW/MEDIUM is noisy without a tuned config)
- `pip-audit --strict` (against the committed lock via `uv sync`)

Marked `continue-on-error: true` for now so this doesn't block PRs on pre-existing findings. A follow-up should do the cleanup pass and flip the flag to blocking.

`safety` intentionally omitted -- it now requires a free-tier API key for its commercial backend, which is a setup burden for forks and CI. `pip-audit` covers the same ground via the PyPI JSON advisory feed and OSV.

## Test plan
- [ ] CI runs the new job and posts results.
- [ ] Confirm existing pyproject `[tool.bandit]` skip (`B404`) is honored.
- [ ] No churn to `uv.lock` from the new job.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -77,3 +77,30 @@ jobs:
         run: |
           uv run --frozen graphify --help
           uv run --frozen graphify install
+
+  security-scan:
+    # The dev deps already include bandit, pip-audit, and safety. Run them in
+    # CI so a new HIGH-severity finding or vulnerable dependency is caught on
+    # the PR that introduces it, rather than at the next manual audit.
+    # Non-blocking for now (continue-on-error) to avoid breaking CI on
+    # pre-existing findings; remove continue-on-error after the initial
+    # cleanup pass.
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v6
+
+      - name: Install uv
+        uses: astral-sh/setup-uv@v8.1.0
+        with:
+          python-version: "3.12"
+
+      - name: Install dependencies
+        run: uv sync --frozen
+
+      - name: bandit (static security analysis)
+        continue-on-error: true
+        run: uv run --frozen bandit -r graphify -ll
+
+      - name: pip-audit (dependency vulnerabilities)
+        continue-on-error: true
+        run: uv run --frozen pip-audit --strict


safishamsi (changes requested):
The concept is right — non-blocking rollout with `continue-on-error: true` is the correct approach given the current state of findings. But there are three bugs in the implementation that need fixing before this lands.

## Bug 1 — bandit ignores `[tool.bandit]` config

```yaml
# current
run: bandit -r graphify -ll

# fix
run: bandit -r graphify -c pyproject.toml -ll
```

bandit does **not** auto-discover `[tool.bandit]` from `pyproject.toml` — `-c` must be passed explicitly. Without it, the `skips = ["B404"]` setting is silently ignored and the job surfaces 10 B404 false positives on every run. I verified this locally: with `-c pyproject.toml`, B404 is correctly suppressed.

## Bug 2 — pip-audit misses 6 of 7 real CVEs

```yaml
# current
run: uv sync --frozen

# fix  
run: uv sync --all-extras --frozen
```

The `test` job already uses `--all-extras --frozen` — this job should match. Without it, pip-audit audits only the base environment and misses the `mcp` extra entirely, which is where the real findings live:

- **pyjwt** — 4 CVEs (in `mcp` extra)
- **starlette** — 2 CVEs (in `mcp` extra)

With the current setup, the job reports 1 finding (the `pip` bootstrap tool, not even a shipped dep). With `--all-extras`, it reports 7. The job as written gives false confidence.

## Bug 3 — severity flag mislabeled

The PR description says `-ll` = "HIGH-severity only" but per `bandit --help`, `-ll` is **MEDIUM and higher** (`-lll` is HIGH-only). The flag itself is probably fine (MEDIUM+ is a reasonable threshold), but the description needs correcting — or if HIGH-only was the actual intent, change to `-lll`.

---

Fix the three items above and this is good to merge. One heads-up: once `--all-extras` is in place, pip-audit will start surfacing the pyjwt/starlette CVEs on every run. Worth resolving those before flipping `continue-on-error` to `false` in a follow-up.

PR #1066, opened 2026-05-28
feat: extract markdown links as first-class edges (fixes #951)

adityachaudhary99:
## Problem
Markdown documents often contain rich cross-references through [links](path.md) and [[wikilinks]], but graphify ignored these as first-class graph relationships.

## What changed
- _resolve_markdown_link() — resolves [text](path.md) targets, handles anchors, extensionless refs, external URL filtering, title attribute stripping
- _resolve_markdown_wikilink() — resolves [[Page Name]] to existing .md files
- extract_markdown() — emits links_to edges with EXTRACTED confidence and link text as edge context
- Edge cases handled: ignores image links, excludes links inside fenced code blocks, strips [[page|display text]] pipe syntax
- links_to added to SEMANTIC_RELATIONS
- Markdown file-level IDs now use _file_node_id
- 8 new tests covering normal links, wikilinks, pipe syntax, images, code blocks, external URLs, title attributes

## Depends on
- #1033 (_file_node_id helper)

PR #1216, opened 2026-06-09
feat: extract markdown links as first-class edges

adityachaudhary99:
Re-implements the markdown link extraction from #1066 on top of current upstream/main, addressing chris-boson's review comment: *"we also need a change here so that we actually run the extraction on markdown files"*.

## What's in this PR

- `graphify/extract.py`: new `extract_markdown(path)` function + dispatch in `extract()` for `.md`/`.mdx`/`.qmd` + extension list update in `collect_files()`. Emits `links_to` (EXTRACTED) edges for resolved `[text](path)` and `[[wikilink]]` references. Filters images, external URLs, unresolved refs, and content inside fenced code blocks.
- `graphify/skill.md`: Part A orchestrator now passes `docs` markdown files alongside `code` files into `extract()`. **This is the actual fix for the review** — without it, the new extractor is never invoked.
- `tests/test_languages.py`: 9 new tests covering file node, link, wikilink, external URL filtering, unresolved link filtering, image link filtering, fenced code block filtering, the resolver unit test, and the high-level `extract()` dispatch.

## Why a new PR instead of fixing #1066

PR #1066's branch (`feat/markdown-link-edges`) is 36 commits behind `upstream/main` AND contains 4 separate PRs' worth of work bundled in (`#1065`/`#1033` file-level node IDs, `#1067` bare-name collisions, `#1069` CI extras, plus the actual #1066 markdown work). A clean rebase is hours of careful work. This PR is a focused, single-commit re-implementation of the review's ask on top of current `upstream/main`.

## Tests

- 9/9 new markdown tests pass
- 47/47 in `tests/test_languages.py` (38 pre-existing + 9 new) — zero regressions
- Pre-existing failures in `test_pipeline.py` and `test_report.py` are missing `graspologic`/`networkx` deps in the test env, unrelated to this change

## Closes

Addresses the review on #1066 from chris-boson.

```
feat: extract markdown links as first-class edges

Adds extract_markdown() to graphify/extract.py and routes .md/.mdx/.qmd
files through it from extract(). For each file, emits:
  - a file-level node (file_type=doc)
  - one links_to edge per resolved [text](path) and [[wikilink]] ref

Unresolved links, image links, external URLs, and content inside fenced
code blocks are all skipped so the graph only contains real, EXTRACTED
cross-references.

Also updates skill.md's Part A orchestrator to pass markdown files
alongside code files into extract(), so the new extractor is actually
invoked by the pipeline (the previous orchestrator only iterated
detect['files']['code']).
```

### DIFF
--- a/graphify/extract.py
+++ b/graphify/extract.py
@@ -14,6 +14,107 @@ def _make_id(*parts: str) -> str:
     return cleaned.strip("_").lower()
 
 
+# Patterns for markdown link / wikilink extraction. Kept conservative:
+#  - _MD_LINK_RE matches [text](target); we then filter out images (![...])
+#    and external URLs (http://, https://, mailto:, ftp://) in code.
+#  - _WIKILINK_RE matches [[Page Name]] and [[Page Name|Display Text]].
+_MD_LINK_RE = re.compile(r"(?<!\!)\[([^\]]+)\]\(([^)\s]+)(?:\s+\"[^\"]*\")?\)")
+_WIKILINK_RE = re.compile(r"\[\[([^\]|]+)(?:\|[^\]]+)?\]\]")
+
+
+def _resolve_md_link(target: str, source: Path) -> Path | None:
+    """Resolve a markdown link target to an existing file in the same directory,
+    trying common extensions. Returns None for external URLs or unresolvable refs.
+    """
+    if not target:
+        return None
+    if re.match(r"^[a-z]+://", target) or target.startswith(("mailto:", "tel:")):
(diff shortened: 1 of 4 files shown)


adityachaudhary99:
Re-triggered CI: the earlier `test (3.10)` failure was an upstream `onnxruntime` wheel-availability issue (no py3.10 wheel), since fixed on the base branch (onnxruntime is now gated to `python_full_version >= '3.11'` in the lockfile). This PR's diff is unchanged.

Issue #1147, opened 2026-06-06
AST extractor emits meaningless `uses` edges from builtin-type nodes (str/int/bool) at confidence 0.5 — inflates god-node rankings


janreview:
**Version:** graphifyy 0.8.32 (pip, Windows, Python 3.12)

**What happened**
The AST extractor creates per-file nodes for builtin type annotations (`str`, `int`, `bool`, `float`, also `MagicMock`) and links them with `uses` edges to user classes that merely co-occur in the same function signature. All these edges carry `confidence_score=0.5` — the exact value the extraction spec forbids as a default.

**Example (measured on a ~6,000-node graph of a Python repo)**
A class `PipelineLog` reported 60 edges / 37 INFERRED. On inspection, 14 of the INFERRED edges were of the form:

```
src_services_google_scout_py_str  -> services_pipeline_logger_pipelinelog  [uses, 0.5]
src_services_google_scout_py_int  -> services_pipeline_logger_pipelinelog  [uses, 0.5]
tests_test_find_contact_fallthrough_py_magicmock -> ... [uses, 0.5]
```

These come from signatures like `def f(place_id: str, domain: str, plog: PipelineLog) -> Optional[dict]`. A "`str` uses PipelineLog" edge is semantically meaningless.

**Impact**
- Degree counts inflate ~25% on affected classes; `str` (35 edges) and `JSONResponse` (50 edges, 48 INFERRED) appeared in the god-node top-10 of GRAPH_REPORT.md, displacing real abstractions.
- The report's suggested questions then point users at verifying these junk edges.

**Related (same family):** a type annotation at a *usage site* (`plog: PipelineLog` in another module) produced a ghost duplicate node (`pipelinelog`, source = the using file) separate from the real class node (`services_pipeline_logger_pipelinelog`), connected only by an INFERRED `uses` 0.5 edge.

**Suggested fix**
Skip `uses` edges where either endpoint is a builtin/typing name (str/int/float/bool/bytes/dict/list/Optional/Any/...), or at minimum exclude builtin nodes from degree/god-node statistics.

safishamsi:
Fixed in v0.8.33 (a380b34).

[...]

Issue #908, opened 2026-05-17
Rust workspaces: AST extractor stem-collides function names across crate boundaries (e.g. test helpers linking to unrelated functions in sibling crates)


akannk001:
## Environment

- **graphifyy:** 0.8.8 (CLI `graphify` 0.4.23, package up to date)
- **Python:** 3.13.x via `uv tool install`
- **Corpus:** Rust workspace with two crates under one root: a primary crate at `src/` and a workspace member at `tools/check-hot-path-unwrap/`. Both are Rust 1.95, edition 2024.
- **Command:** `/graphify .` (full pipeline, project-surface scope ~117 files)

## Problem

The AST extractor's cross-file call resolver matches unresolved callee names against any node label in the graph regardless of crate boundary. In a Rust workspace where two crates share a common helper-function name (e.g. `parse`, `new`, `from`, `eval`, `main`), this produces wrong INFERRED edges that link unrelated callsites in separate crates.

Concrete reproduction in cc2kiro (Rust adapter, two-crate workspace):

The workspace-member crate has a tiny test helper:

```rust
// tools/check-hot-path-unwrap/src/walker.rs:182
fn parse(src: &str) -> File {
    syn::parse_file(src).expect("parse")
}
```

It is called only by unit tests in the same file (10 internal `EXTRACTED` edges, all correct).

The graph also emits **5 spurious `INFERRED` edges** from `walker.rs::parse` to functions in completely separate files in the primary crate:

| graph claim | actual relationship |
|---|---|
| `walker.rs::parse → src/bin/cc2kiro-query.rs::main` | none — `main` is a CLI entry point, not a parser |
| `walker.rs::parse → src/client/auth.rs::needs_refresh_at` | none — date helper, never called from walker |
| `walker.rs::parse → src/client/auth.rs::dt` | none — same |
| `walker.rs::parse → src/client/auth.rs::format_rfc3339_drops_nanos` (test) | none — test in different crate |
| `walker.rs::parse → src/client/refresh.rs::now` | none — time helper |

The `parse(&str) -> File` helper takes a `&str` and returns `syn::File`. It cannot reach any of those functions across the crate boundary.

## Quantification

In a 1,415-node / 2,058-edge graph, 123 `INFERRED` `calls` edges total. Bare-name magnets that produce most of the cross-crate noise:

| label | INFERRED-call edge count |
|---|---|
| `.start()` | 32 |
| `app()` | 11 |
| `.from_str()` | 8 |
[...]

PR #1221, opened 2026-06-09
fix: blocklist Python stdlib and JS global names from noise nodes and INFERRED edges

arcticrobotops:
## Problem

Python annotation references and JS require-bindings to stdlib/runtime names become free-floating graph nodes. On a real ~25k-node graph, `Path` existed as **174 per-file nodes**, plus `Any` ×77, `os` ×74, `datetime` ×31 — and stdlib names held 3 of the top 5 god-node slots in GRAPH_REPORT.md, masking the project's actual hubs.

## Root cause

Three cooperating layers (traced on 0.8.36):

1. `_python_collect_type_refs` emits nodes for annotation identifiers like `def f(p: Path) -> Any`. `Any` is missing from `_PYTHON_TYPE_CONTAINERS`, and stdlib classes/modules are covered nowhere, so `ensure_named_node` materializes each as a node.
2. `_disambiguate_colliding_node_ids` sees the same bare id from N files and splits it into per-file ids — hence 174 `Path` nodes.
3. `_resolve_cross_file_imports` classifies those nodes as importing classes and emits spurious `Path --uses--> ProjectClass` INFERRED edges from every one.

JS path is separate: `_js_extra_walk` creates a per-file binding node for module-level `const fs = require('fs')`, and runtime globals (`setTimeout`, `fetch`, `require`) enter the unresolved-call queue where they attract cross-file INFERRED `calls` edges.

## Fix

Follows the existing precedents for this exact class of problem:

- **`_PY_STDLIB_BLOCKLIST`** applied in the annotation walker via a new `_python_type_ref_ok()` — the same layer as `_PYTHON_ANNOTATION_NOISE` (#1147), so noise is suppressed before node *or* edge creation.
- **`_JS_GLOBAL_BLOCKLIST`** suppresses the require-binding const node in `_js_extra_walk` (the `imports_from` edge from `_require_imports_js` is preserved).
- **`_UNRESOLVED_CALL_BLOCKLISTS`** gates raw-call queue insertion for Python/JS/TS — mirroring `_RUST_TRAIT_METHOD_BLOCKLIST` (#908): same-file EXTRACTED matches still resolve; only cross-file resolution is skipped.
- **`analyze.py`**: both sets folded into `_BUILTIN_NOISE_LABELS` (defense-in-depth, matching #1147) — this also covers graphs built from stale extraction caches, since cache keys are content-hash only.

Matching is on raw identifiers and extracted function labels carry a `()` suffix, so project functions are never shadowed (verified empirically).

**Tradeoff (same one the Rust blocklist accepts):** a project class literally named `Path`/`Counter`/`json` loses annotation-reference edges and cross-file call resolution. Same-file resolution and `contains`/`method` edges are unaffected. Lists deliberately exclude commonly project-named identifiers (`Event`, `Queue`, `Logger`, `Request`, `Response`, `Thread`, `Task`).

## Tests

- 6 new tests in `tests/test_extract.py`: node suppression, no INFERRED `uses` edges, project import/`uses` edge survival, Python and JS unresolved-queue gates, god_nodes guard.
- `tests/test_extract.py`: 82 passed. Related extraction/resolution suites: 369 passed, 22 skipped.
- Full suite: identical pass/fail profile as the unmodified base tree (the 21 failures are pre-existing environmental ones — missing optional grammars). Zero regressions.

## Release-note suggestion

Existing extraction caches still contain the noise nodes until source files change or the cache is cleared; the `analyze.py` guard covers god-node rankings in the meantime, but raw node counts shrink only after re-extraction.

### DIFF
--- a/graphify/analyze.py
+++ b/graphify/analyze.py
@@ -4,16 +4,20 @@ from pathlib import Path
 import networkx as nx
 
 from graphify.build import edge_data
+from graphify.extract import _JS_GLOBAL_BLOCKLIST, _PY_STDLIB_BLOCKLIST
 
 # Builtin/mock names that can appear as annotation-derived nodes in pre-existing
 # graphs. Excluded from god-node ranking so they don't displace real abstractions
-# even if they weren't filtered at extraction time (#1147).
+# even if they weren't filtered at extraction time (#1147). Extended with the
+# Python stdlib and JS global blocklists so stale per-file cache entries
+# (cache keys are content hashes, so old extractions survive upgrades) can't
+# push `Path`/`Any`/`os` reference nodes back into the ranking.
 _BUILTIN_NOISE_LABELS = frozenset({
     "str", "int", "float", "bool", "bytes", "bytearray", "complex", "object",
     "True", "False",
     "MagicMock", "Mock", "AsyncMock", "NonCallableMock",
     "NonCallableMagicMock", "PropertyMock", "patch", "sentinel",
-})
+}) | _PY_STDLIB_BLOCKLIST | _JS_GLOBAL_BLOCKLIST
 
 # Language families — extensions sharing a runtime can legitimately call each other
 _LANG_FAMILY: dict[str, str] = {
--- a/graphify/extract.py
+++ b/graphify/extract.py
@@ -42,6 +42,33 @@ _LANGUAGE_BUILTIN_GLOBALS: frozenset[str] = frozenset({
     "callable", "getattr", "setattr", "hasattr", "delattr", "vars", "dir",
 })
 
+# JS/Node runtime globals and Node core module names that appear as bare
+# identifier references but are never project-defined symbols. Module-level
+# `const fs = require('fs')` bindings otherwise become a per-file node
+# (`app_fs`, `app_path`, `app_os`, ...) in every file that requires the module,
+# and unresolved bare calls to runtime globals (setTimeout, fetch, require)
+# produce spurious cross-file INFERRED edges whenever a project symbol shares
+# the name. Counterpart to _PY_STDLIB_BLOCKLIST; same pattern as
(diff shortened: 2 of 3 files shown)


safishamsi (changes requested):
## Review: Request Changes

Overall this is a strong PR — correct layering (extraction-time blocking + god-node defense-in-depth for stale caches), the same-file resolution is correctly preserved (blocklist only gates the cross-file unresolved-call queue at `extract.py:3420`), and test coverage is solid. One blocking issue:

### Blocking: JS const-binding guard is over-broad

`extract.py:1874-1880` drops any const binding whose name is in `_JS_GLOBAL_BLOCKLIST` regardless of the RHS. The PR comment says this targets `const fs = require('fs')` and friends, but the code does not check that the value is a `require()` call. Project variables with these names are silently dropped:

```js
const path = computeRoute(req);   // dropped — not a require
const url  = new MyUrlBuilder().build();  // dropped
const crypto = makeCryptoHelper();        // dropped
```

Names like `path`, `url`, `crypto`, `stream`, `process`, `module`, `document`, `window` are plausible project variable names. The guard needs to verify the RHS is a `require()` call (or that an `imports_from` edge for this binding already exists) before dropping the node.

### Missing test for the false-positive case

`test_extract_js_core_module_require_binding_produces_no_node` only tests the intended-require path. Please add a test asserting that a JS const named `path` or `url` with a non-`require` RHS is **not** dropped from the graph.

### Non-blocking nits
- The two new fixture files in `FIXTURES` dir use `finally: unlink()` which leaks on process kill and collides under `pytest-xdist`. Prefer `tmp_path`.
- Minor redundant overlap with `_LANGUAGE_BUILTIN_GLOBALS` for `JSON`, `Math`, `Promise`, `console`, `TypeError` — harmless but worth a comment.

PR #1231, opened 2026-06-10
Fix cross-file inheritance and ghost type references in AST extractor

### LINKED ISSUE #1186: AST extractor misses cross-file inheritance and produces ghost-duplicate nodes for cross-file type references
# AST extractor misses cross-file inheritance and produces ghost-duplicate nodes for cross-file type references

**graphify version:** 0.8.35
**Python:** 3.x (uv tool install)
**Corpus:** 891 Python files, 14,050 nodes, 17,822 edges, AST-only extraction

## Summary

The AST extractor has a single root-cause gap that surfaces as two distinct bugs:

1. `class X(Base):` does not produce an `inherits/EXTRACTED` edge when `Base` is imported from another project file.
2. Type annotations and class-base references that cross a file boundary create duplicate "ghost" nodes with inconsistent ID formats instead of resolving to the canonical definition node.

Both stem from the same shortcoming: when AST sees a Name reference whose binding comes from an `import` of another project module, it does not resolve the reference back to the canonical node. Instead it emits a fresh node (often with a malformed ID) and/or downgrades the relation to `uses/INFERRED` at the import line.

## Repro

Any Python codebase with cross-file inheritance is enough. Concrete observed examples below come from `iwfm-py` (procedural codebase with one Strategy/Backend hierarchy):

```python
# iwfm/iwfm/hdf5/hdf_metadata_base.py
class HdfBackend(ABC):  # L64
    @abstractmethod
    def get_n_nodes(...): ...

# iwfm/iwfm/hdf5/hdf_metadata.py
from iwfm.iwfm.hdf5.hdf_metadata_base import HdfBackend, HdfMetadata  # L42
class HdfReader(HdfBackend):  # L47
    def get_n_nodes(self, ...) -> int: ...
    def get_metadata(self, ...) -> HdfMetadata: ...  # L284

# iwfm/iwfm/dll/dll_backend.py
from iwfm.iwfm.hdf5.hdf_metadata_base import HdfBackend, HdfMetadata  # L25
class DllBackend(HdfBackend):  # L29
    def get_metadata(self, ...) -> HdfMetadata: ...  # L261
```

## Bug 1 — Cross-file inheritance silently dropped

Greping the corpus for all `class X(Y):` declarations and comparing to graph `inherits` edges:

| Source declaration | Captured? | Why |
|---|---|---|
| `class SimulationFiles(_DictAccessMixin):` | ✓ | Same file |
| `class PreprocessorFiles(_DictAccessMixin):` | ✓ | Same file |
| `class GroundwaterFiles(_DictAccessMixin):` | ✓ | Same file |
| `class RootzoneFiles(_DictAccessMixin):` | ✓ | Same file |
| `class BackendNotAvailableError(HdfError):` | ✓ | Same file |
| `class DataSourceError(HdfError):` | ✓ | Same file |
| `class HdfReader(HdfBackend):` | **✗** | **Cross-file** |
| `class DllBackend(HdfBackend):` | **✗** | **Cross-file** |

Capture rate:
- Same-file internal bases: 6/6 (100%)
- Cross-file internal bases: 0/2 (0%)
- Stdlib bases (`ABC`, `Enum`, `Exception`, `str`, `object`): all captured

What the graph contains for the missed cases:

```
HdfReader  --uses/INFERRED--> HdfBackend  @ L42  (the import line, not L47 where class is declared)
DllBackend --uses/INFERRED--> HdfBackend  @ L25  (the import line, not L29)
```

Expected:

```
HdfReader  --inherits/EXTRACTED--> HdfBackend  @ L47
DllBackend --inherits/EXTRACTED--> HdfBackend  @ L29
```

### Impact on graph quality

Community detection has no way to bind a polymorphic family together without inheritance edges. In this corpus:

| Community | What it contains | Should be |
|---|---|---|
| C4 | `HdfBackend` ABC + tests, alone | Same family |
| C17 | `HdfReader` + tests, alone | Same family |
| C147 | `DllBackend` alone (no tests) | Same family |

Three communities for one Strategy/Backend abstraction. The HDF backend hierarchy is invisible to any clustering or `god_nodes` computation that relies on structural edges.

## Bug 2 — Ghost-duplicate nodes from cross-file type references

Cross-file references in type annotations and class bases produce duplicate nodes with malformed IDs instead of resolving to the canonical definition.

Four substantive ghosts observed in 14,050 nodes:

| Ghost ID | Canonical ID | Source of ghost | In-edges absorbed |
|---|---|---|---|
| `context` (bare) | n/a (refers to external `typer.Context`) | `def main(ctx: typer.Context, ...)` @ main.py:L90 | 1 |
| `userlevel` (bare) | `cli_context_userlevel` (20/5 edges) | `def configure_logging(user_level: UserLevel)` @ main.py:L67 | 2 |
| `iwfm_iwfm_dll_dll_backend_py_hdfmetadata` | `hdf5_hdf_metadata_base_hdfmetadata` (19/0 edges) | `def get_metadata(...) -> HdfMetadata:` @ dll_backend.py:L261 | 1 |
| `iwfm_iwfm_hdf5_hdf_metadata_py_hdfmetadata` | `hdf5_hdf_metadata_base_hdfmetadata` | `def get_metadata(...) -> HdfMetadata:` @ hdf_metadata.py:L284 | 1 |

The bare-name ghosts (`context`, `userlevel`) and stdlib bare names (`object`, `str`, `enum`, `exception`, `abc`) are present in the raw AST cache — not introduced during merge — so the bug is in the extractor.

### Two distinct ID-format conventions coexist

Skill spec mandates `{parent_dir}_{filename_stem}_{entity}` (e.g. `cli_main_userlevel`). Observed formats:

1. **Bare name** — e.g. `userlevel`, `context`. No prefix at all. Used when the resolver gives up on a name.
2. **Full-path-with-`_py_`** — e.g. `iwfm_iwfm_dll_dll_backend_py_hdfmetadata`. Includes full path from repo root, keeps `.py` extension as a literal `_py_` segment. Used for type annotations referencing cross-file imports.
3. **Clean spec** — e.g. `hdf5_hdf_metadata_base_hdfmetadata`. Used for entity definitions inside the file.

Both (1) and (2) produce nodes that don't match (3), so canonical and ghost coexist for the same logical symbol.

### Adjacent cosmetic issue

214 nodes have an `iwfm_iwfm_*` double-prefix because the codebase layout is `iwfm/iwfm/*.py` (package nested in repo root with same name). The ID generator joined two levels of parent path when it should have used only the immediate parent. These are mostly unique nodes — no ghost duplicates — but the format inconsistency obscures the spec.

## Root cause hypothesis

When AST encounters a `Name` node (in a class base, type annotation, or other typed context) whose binding comes from `from X import Y` where `X` is a project-internal module:

- Resolver doesn't follow the import to the canonical definition node.
- Resolver emits a fresh node with whatever ID the local context produces (bare name, or full path with `_py_` separator).
- The relation gets degraded — `inherits` becomes `uses/INFERRED` at the import line, type annotation becomes a `references/EXTRACTED` to the ghost.

Same-file references work because the resolver finds the definition in the current file's symbol table without needing import resolution. Stdlib references work because they bottom out at known external types that don't need canonical-node resolution.

## Suggested fix direction

A single pass over `from X import Y` statements that builds a name → canonical-node-ID map for project-internal imports, applied before emitting any edge involving an imported name. With that map:

- `class HdfReader(HdfBackend):` resolves `HdfBackend` to `hdf5_hdf_metadata_base_hdfbackend` and emits `inherits/EXTRACTED`.
- `def get_metadata(...) -> HdfMetadata:` resolves `HdfMetadata` to `hdf5_hdf_metadata_base_hdfmetadata` and emits `references/EXTRACTED` without a fresh ghost.
- Bare-name fallback (`userlevel`, `context`) reserved for truly unresolvable references (stdlib, third-party, builtins).

ID-format inconsistency (1) vs (2) vs (3) is a separate normalization fix — pick one and apply everywhere.

## Reproduction evidence

Audit Q&A from the iwfm-py corpus (saved to `graphify-out/memory/`):
- `query_20260607_193207_audit_of_75_inferred_edges_on_read_next_line_value.md`
- `query_20260607_193602_why_does_context___uses____userlevel_appear_as_sur.md`
- `query_20260607_193903_sweep_for_ghost_duplicate_nodes_across_14050_node.md`
- `query_20260607_194331_why_is_hdf_reader_vs_metadata_backend_split_across.md`
- `query_20260607_194503_sweep_for_missing_inheritance_edges_across_corpus.md`

Happy to provide the corpus or extracted graph.json for repro.

MananJain39:
## Summary
Fixes #1186

This PR fixes missing cross-file Python inheritance edges and unresolved type references by routing them through the deferred symbol-resolution pipeline instead of resolving them immediately during AST traversal.

## Changes
* **Python Superclasses:** Collects base classes from `class_definition` nodes as `_SymbolUseFact(relation="inherits")`. This restores proper `inherits` edges for project-internal cross-file bases instead of downgrading them to generic `uses` edges on import lines.
* **Type Annotations:** Collects parameter and return annotations from `function_definition` nodes as `_SymbolUseFact(relation="references")`. This prevents the creation of duplicate fallback ("ghost") nodes for imported types.

## Result
Cross-file references now successfully leverage the import-aware resolution post-pass, ensuring canonical edge generation and eliminating duplicate unresolved nodes.

### DIFF
--- a/graphify/extract.py
+++ b/graphify/extract.py
@@ -7820,6 +7820,35 @@ def _collect_python_symbol_resolution_facts(
                     )
                 )
 
+        # NEW: Collect heritage and annotation uses for classes/functions so
+        # they can be resolved via import/export facts in the post-pass.
+        # Walk top-level class definitions and collect base names and any
+        # annotation type names as _SymbolUseFact entries.
+        stem = _file_stem(path)
+        for node in _walk_python_tree(root_node):
+            if node.type == "class_definition":
+                # class name -> nid (file-scoped)
+
+                name_node = node.child_by_field_name("name")
+                if name_node is None:
+                    continue
+                class_name = source[name_node.start_byte:name_node.end_byte].decode("utf-8", errors="replace")
+                class_nid = _make_id(stem, class_name)
+                # Collect superclass references so they can be resolved
+                # against import/export facts during the symbol-resolution post-pass.
+
+                supers = node.child_by_field_name("superclasses")
+                if supers is not None:
+                    for s in supers.children:
+                        if s.type == "identifier":
+                            base_name = source[s.start_byte:s.end_byte].decode("utf-8", errors="replace")
+                            facts.uses.append(_SymbolUseFact(path, class_nid, base_name, "inherits", "type", s.start_point[0] + 1))
+                        elif s.type == "attribute":
+                            #qualified name like mod.Base -> tail
+                            raw = source[s.start_byte:s.end_byte].decode("utf-8", errors="replace") 
+                            tail = raw.rsplit(".", 1)[-1]
+                            facts.uses.append(_SymbolUseFact(path, class_nid, tail, "inherits", "type", s.start_point[0] + 1))
+
 
 def _augment_symbol_resolution_edges(
     paths: list[Path],
--- a/graphify/extract.py
+++ b/graphify/extract.py
@@ -7849,6 +7849,42 @@ def _collect_python_symbol_resolution_facts(
                             tail = raw.rsplit(".", 1)[-1]
                             facts.uses.append(_SymbolUseFact(path, class_nid, tail, "inherits", "type", s.start_point[0] + 1))
 
+            # Collect annotation type references for both classes and functions, which are handled the same way for import resolution purposes. Walk function definitions and class-level variable declarations; for any annotation node, collect identifier and attribute references as _SymbolUseFact entries with relation "references" and context "annotation".
+            if node.type == "function_definition":
+                fn_name_node = node.child_by_field_name("name")
+                if fn_name_node is None:
+                    continue
+                fn_name = source[fn_name_node.start_byte:fn_name_node.end_byte].decode("utf-8", errors = "replace")
+                fn_nid = _make_id(stem, fn_name)
+                # Parameters
+                params = node.child_by_field_name("parameters")
+                if params is not None:
+                    for p in params.children:
+                        if p.type in ("typed_parameter", "default_parameter"):
+                            type_node = p.child_by_field_name("type")
+                            if type_node is not None:
+                                # Collect simple identifiers and qualified attribute references (e.g. `mod.Type`) appearing in annotations as separate facts; the post-pass will attempt to resolve both against imports/exports.
+                                for c in type_node.children:
+                                    if c.type == "identifier":
+                                        tname = source[c.start_byte:c.end_byte].decode("utf-8", errors="replace")
+                                        facts.uses.append(_SymbolUseFact(path, fn_nid, tname, "references", "annotation", c.start_point[0] + 1))
+                                    elif c.type == "attribute":
+                                        raw = source[c.start_byte:c.end_byte].decode("utf-8", errors="replace")
+                                        tail = raw.rsplit(".", 1)[-1]
+                                        facts.uses.append(_SymbolUseFact(path, fn_nid, tail, "references", "annotation", c.start_point[0] + 1))
+
+            #return type
+            ret = node.child_by_field_name("return_type")
+            if ret is not None:
+                for c in ret.children:
+                    if c.type == "identifier":
+                        tname = source[c.start_byte:c.end_byte].decode("utf-8", errors="replace") 
+                        facts.uses.append(_SymbolUseFact(path, fn_nid, tname, "references", "annotation", c.start_point[0] + 1))
+                    elif c.type == "attribute":
+                        raw = source[c.start_byte:c.end_byte].decode("utf-8", errors="replace")
+                        tail = raw.rsplit(".", 1)[-1]
+                        facts.uses.append(_SymbolUseFact(path, fn_nid, tail, "references", "annotation", c.start_point[0] + 1))
+
 
 def _augment_symbol_resolution_edges(
     paths: list[Path],


safishamsi (changes requested):
## Review: Request Changes

### Root issue: incomplete fix — ghost node/edge persists

The `inherits` direction targets a real bug (ambiguous cross-file base classes), but the fix is incomplete. It adds a parallel canonical edge via the deferred resolution pipeline while leaving the original ghost node (empty `source_file`) and ghost edge created at `extract.py:2413-2423` in the graph. Because `(source, relation)` differs in target, `nx.DiGraph` in `build.py:300` does not collapse them — both survive. Issue #1186's duplicate-node symptom is not resolved.

**Required fix:** suppress the eager bare-fallback at `extract.py:2413-2423` for Python when the base is import-bound, rather than adding a parallel edge.

### Real defect: `fn_nid` scoping bug

The `#return type` block (diff ~lines 55-65) is outside the `if node.type == "function_definition"` guard but references `fn_nid`, which is only defined inside it. For any non-function node with a `return_type` field this raises `UnboundLocalError` or silently attaches return-type refs to the wrong function using a stale value from the previous iteration.

### Annotation half is redundant and uses wrong context

The existing path at `extract.py:2978-2997` (using `_python_collect_type_refs`) already handles cross-file annotation edges correctly and is covered by `test_python_import_resolution.py::test_python_parameter_return_and_generic_contexts`. The PR's hand-rolled walker:
- Emits `context="annotation"` which is not in `REFERENCE_CONTEXTS` (`extract.py:139-141`)
- Skips `_python_type_ref_ok` / `_PY_STDLIB_BLOCKLIST` checks
- Misses subscripted generics (`list[Payload]`, `Optional[X]`) that the recursive `_python_collect_type_refs` handles
- Misses `typed_default_parameter` node type

Recommend: drop the annotation half entirely and reuse the existing path, or route through `_python_type_ref_ok` with `context="parameter_type"`/`"return_type"`.

### No tests

No test exercises the ambiguous cross-file base scenario or confirms ghost node removal. Please add a test verifying: (1) the canonical `inherits` edge resolves to the correct imported class, and (2) no ghost node with empty `source_file` remains in the graph.

MananJain39:
Thanks for the detailed review @safishamsi  — you're right that the current change only adds the deferred canonical edge path without suppressing the original eager fallback emission, so the ghost node/edge can still survive in parallel.

I'll rework the inheritance fix to suppress the eager fallback path for import-bound Python base references rather than layering an additional edge on top.

You're also correct about the annotation half:

* the existing `_python_collect_type_refs` path already covers those cases more comprehensively,
* my added walker duplicates logic and misses several existing safeguards/contexts.

I'll drop the annotation portion and focus the PR on the missing cross-file inheritance resolution path only.

I'll also add regression coverage for:

1. canonical cross-file `inherits` edge resolution
2. absence of ghost nodes/edges after extraction

Issue #1072, opened 2026-05-29
[Windows] _call_claude_cli fails with WinError 2 — subprocess.run cannot execute claude.ps1 (proposed fix included)


dougwmorrow:
## Summary

On Windows, `--backend claude-cli` fails for every chunk with `[WinError 2] The system cannot find the file specified`, because `_call_claude_cli` in `graphify/llm.py` invokes `subprocess.run(["claude", ...])` and Windows' `CreateProcess` cannot directly execute the `claude.ps1` script that npm installs. The bare-name `claude` resolves via PATHEXT — and on systems where `.PS1` is listed in PATHEXT ahead of `.CMD`, Python `subprocess.run` resolves to the PowerShell script and crashes. `claude.cmd` (also installed by npm) IS directly executable by `CreateProcess` and works fine.

Result: `--backend claude-cli` is broken on Windows for a non-trivial fraction of installs. All N semantic chunks fail with the same `WinError 2`, then `graphify` aborts with `all semantic chunks failed for backend 'claude-cli'`.

## Environment

- OS: Windows 11 Home 10.0.26200
- Python: 3.12.12 (uv-bundled)
- Graphify: v0.8.24 (installed via `uv tool install git+https://github.com/safishamsi/graphify`)
- Claude Code: v2.1.142
- npm `claude` install layout (typical):
  - `C:\Users\<user>\AppData\Roaming\npm\claude` (shell shim, not Windows-executable)
  - `C:\Users\<user>\AppData\Roaming\npm\claude.cmd` (Windows batch wrapper, **subprocess-executable**)
  - `C:\Users\<user>\AppData\Roaming\npm\claude.ps1` (PowerShell script, **NOT subprocess-executable**)
- `shutil.which("claude")` returns `claude.ps1` on this machine (PATHEXT prefers `.PS1`).

## Reproduction

```powershell
# Prereqs: claude CLI installed + authenticated; graphify installed; ANTHROPIC_API_KEY NOT needed for claude-cli backend
$env:GRAPHIFY_CLAUDE_CLI_MODEL = "claude-sonnet-4-6"
mkdir _gtest
copy README.md _gtest\
graphify extract _gtest --backend claude-cli --max-concurrency 1
```

## Actual output

```
[graphify extract] scanning C:\...\_gtest
[graphify extract] found 0 code, 1 docs, 0 papers, 0 images
[graphify extract] semantic extraction on 1 files via claude-cli...
[...]

PR #1240, opened 2026-06-10
fix: resolve claude.cmd path in _call_llm claude-cli backend on Windows

setifismedia-stack:
## Bug

On Windows, `graphify label . --backend=claude-cli` fails silently: every batch dies with `[WinError 2] The system cannot find the file specified` and the `Community N` placeholders are kept.

Root cause: in `graphify/llm.py`, `_call_llm` (the `claude-cli` branch, ~line 1580) spawns the CLI with the bare name:

```python
subprocess.run(["claude", "-p", ...])
```

The npm-installed Claude Code CLI on Windows is `claude.cmd` / `claude.ps1` — there is no `claude.exe` — and `CreateProcess` cannot execute a bare name without an `.exe` extension.

This is the exact same bug that was already fixed in `_call_claude_cli` for issue #1072 (~line 868, which resolves via `shutil.which("claude.cmd")` first), but the fix was never applied to `_call_llm`, which has its own independent subprocess call (used by community labeling).

## Fix

Same resolution as #1072, applied to `_call_llm`:

```python
_claude = shutil.which("claude.cmd") or shutil.which("claude")
if _claude is None:
    raise RuntimeError("Claude Code CLI not found on $PATH")
proc = subprocess.run([_claude, "-p", ...])
```

On non-Windows platforms, `shutil.which("claude.cmd")` returns `None`, so behavior is unchanged.

## Verification

Patch applied locally on Windows 11 against an installed copy and verified functional: `graphify label . --backend=claude-cli` then named **227 communities successfully** (previously 0, all batches failing with WinError 2).

### DIFF
--- a/graphify/llm.py
+++ b/graphify/llm.py
@@ -1579,10 +1579,13 @@ def _call_llm(prompt: str, *, backend: str, max_tokens: int = 200) -> str:
 
     if backend == "claude-cli":
         import shutil, subprocess
-        if shutil.which("claude") is None:
+        # Prefer claude.cmd on Windows: CreateProcess cannot execute the bare
+        # name when shutil.which resolves to claude.ps1 (same as issue #1072).
+        _claude = shutil.which("claude.cmd") or shutil.which("claude")
+        if _claude is None:
             raise RuntimeError("Claude Code CLI not found on $PATH")
         proc = subprocess.run(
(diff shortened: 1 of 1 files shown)


safishamsi:
The `_call_llm` Windows claude.cmd fix is already on v8 HEAD (merged as part of b830862 which also fixes issues #1257, #1298, and #1299). This PR is superseded — closing.

PR #1242, opened 2026-06-10
feat: add GDScript (.gd) and HTML (.html/.htm) language extractors

jerryymkim:
## What

Two new language extractors, each following the ARCHITECTURE.md 5-step recipe:

**GDScript (`.gd`)** — custom tree-sitter walk extracting `class_name`, `extends` (both identifier and `extends \"res://...gd\"` forms → `inherits` edges, with `res://` resolved against the `project.godot` root), `signal` declarations, functions (incl. static), inner classes, and `preload()`/`load()` string literals → `imports` edges. Call-graph second pass: EXTRACTED for same-file targets, INFERRED for unknown/engine calls.

**HTML (`.html`/`.htm`)** — file node, `tag#id` element nodes (`contains`), local `<script src>` / `<link href>` → `imports` (http(s)/`//`/data:/mailto:/javascript: skipped per the external-ref convention), inline `<script>` parsed with `tree_sitter_javascript` (row-offset-corrected) so inline functions become real nodes with `calls` edges, and `getElementById('x')`/`querySelector('#x')` literals → INFERRED `uses` edges to matching id elements.

## Dependencies

- `tree-sitter-html` added as a core dependency (small, on PyPI like the other grammars).
- GDScript's grammar has no standalone PyPI package, so it uses `tree-sitter-language-pack` as an **optional extra** (`pip install graphifyy[gdscript]`, also in `all`) — mirroring the existing `sql`/`dm`/`terraform` extras. Without it, `.gd` files degrade gracefully to the verilog-style `{"error": "... not installed"}` dict.

## Doc-pass reconciliation

`.html` previously sat in `DOC_EXTENSIONS`; it's now moved to `CODE_EXTENSIONS` so AST extraction wins. `classify_file` checks CODE before DOC, so there is no double-processing path. `graphify-out/` HTML reports remain excluded via `_SKIP_DIRS`.

## Tests

- `tests/fixtures/sample.gd` + `sample.html`; 18 new tests in `test_languages.py` (GDScript tests behind a `skipif` on the optional extra, matching the `[dm]` pattern). 248/248 passing locally; no new failures in the full suite vs a clean v8 baseline on the same machine.
- Smoke-tested on a real 296-file Godot 4.6 project: 3,170 nodes / 15,162 edges, 0 errors; and on a 1,619-line single-file HTML app: 107 nodes / 354 edges with inline-JS function nodes and id-`uses` edges.

## Notes for review

- `uses` edges source from the file node rather than the enclosing inline function (regex pass over script text). Happy to scope them per-function if preferred.
- `@onready`/`@export` variable declarations are deliberately skipped as graph noise; preload scans still run over them.


jerryymkim:
Withdrawing — keeping this as personal work on my fork for now.

jerryymkim:
Reopening — decided to contribute this upstream after all.

PR #1256, opened 2026-06-11
fix: sync claude references/query.md with full vocab-expansion Step 0 (closes #1255)

### LINKED ISSUE #1255: fix: graphify install --project ships older references/query.md (103 lines, no Step 0 vocab expansion) vs graphify vscode install (249 lines)
## Summary

`graphify install --project` and `graphify vscode install` ship **different versions of `references/query.md`** in v0.8.37. The project-scoped install (used by Claude Code) gets an older, shorter file that is missing the critical Step 0 constrained vocabulary expansion. The vscode install gets the correct, full version.

This means Claude Code  -  the primary platform  -  silently runs inferior query logic compared to VS Code Copilot Chat.

## Environment

- macOS Apple Silicon
- graphifyy 0.8.37 (installed via `uv tool install graphifyy`)
- Claude Code (project-scoped skill via `graphify install --project`)
- VS Code Copilot Chat (global skill via `graphify vscode install`)

## Repro

```bash
uv tool install graphifyy
cd /path/to/any-project
graphify install --project       # Claude Code
graphify vscode install          # VS Code
wc -l .claude/skills/graphify/references/query.md        ~/.copilot/skills/graphify/references/query.md
```

## Actual

```
103 .claude/skills/graphify/references/query.md
249 /Users/<user>/.copilot/skills/graphify/references/query.md
```

The 103-line file shipped to Claude Code is missing:
- **Step 0  -  Constrained query expansion** (the section marked `REQUIRED before traversal`)
- Graph vocabulary extraction (the `import json, re` block that builds `.vocab.txt`)
- The token-selection constraints (pick only tokens present in graph vocab, no invented synonyms)
- The auditable expansion printout before traversal

Instead it has a naive fallback: `[t.lower() for t in question.split() if len(t) > 3]`  -  plain word split with no graph-vocabulary grounding.

## Expected

Both `graphify install --project` and `graphify vscode install` should ship the **same** `references/query.md`. The 249-line version with Step 0 vocab expansion is the correct one.

## Why it matters

The vocab expansion step is what makes `graphify query` useful for natural-language questions. Without it, the CLI latches onto the wrong start node (e.g. querying "how does chat streaming work" starts BFS from the `Chat()` UI component instead of `streamText`/`useChat`/`DataStreamHandler`) and returns 100+ noisy nodes with no useful signal.

The result is that Claude Code users  -  who are using the project-scoped install path  -  get systematically worse query results than VS Code Copilot users from the same graph, despite identical underlying data. This makes the tool appear broken when it is not.

## Verified on

Tested on a 757-file, 11 634-node TypeScript/Next.js repo (graphify 0.8.37). After manually copying the 249-line `query.md` from the VS Code skill location to the Claude Code skill location, query quality improved significantly for natural-language flow questions (e.g. auth, MCP integration, artifact creation).

burrows99:
## What

`graphify/skills/claude/references/query.md` was 103 lines — missing the **Step 0 Constrained query expansion** block that every other platform skill (vscode, codex, copilot, opencode, claw, kiro, amp, pi, trae, droid, kilo, windows) already ships at 249 lines.

Additionally, a cross-platform bug was found (via reviewer @uditstocks) and fixed across all 13 platform skills: Step 1's length filter used `len(t) > 3` (min 4 chars) while Step 0 extracts tokens with `len(t) >= 3` (min 3 chars), silently dropping valid 3-char tokens like `api`, `use`, `get`, `tok`, `auth` before BFS/DFS.

## Root cause

The `skills/claude/` directory contains a stale copy of `references/query.md` that predates the vocabulary-expansion rewrite. All other platform directories received the update; `claude/` did not.

The `len(t) > 3` vs `>= 3` mismatch existed in the source file and was copied to all platforms — making it cross-platform from the start.

## Fix

**Commit `a1ac4cb`** — sync `skills/claude/references/query.md` (103 lines → 249 lines):
```
before: 103 lines  graphify/skills/claude/references/query.md
after:  249 lines  graphify/skills/claude/references/query.md
```

**Commit `5643c7f`** — align Step 1 filter for `claude/` (`> 3` → `>= 3` on line 49)

**Commit `f485fc2`** — same `>= 3` fix applied to all 12 remaining platforms:
`amp, claw, codex, copilot, droid, kilo, kiro, opencode, pi, trae, vscode, windows`

See: https://github.com/safishamsi/graphify/pull/1256#issuecomment-4680592589

## Why it matters

Without Step 0, the `graphify query` CLI latches onto wrong start nodes for natural-language questions. The binary does no stemming or synonym matching — it relies entirely on the agent first expanding the question against actual graph vocabulary tokens.

Without the `>= 3` fix, tokens that Step 0 deliberately selected from graph vocab are silently discarded before traversal even begins.

Verified on a 757-file, 11 634-node TypeScript/Next.js repo: after applying this fix, natural-language query quality matches VS Code Copilot results from the identical graph.

Closes #1255

### DIFF
--- a/graphify/skills/claude/references/query.md
+++ b/graphify/skills/claude/references/query.md
@@ -1,6 +1,6 @@
 # graphify reference: query, path, explain
 
-Load this when the user asks a question against an existing graph, or runs `/graphify path` or `/graphify explain`. The core's query stub points here for the full traversal flow.
+Load this when the user asks a question against an existing graph, or runs `/graphify path` or `/graphify explain`. The core's query stub points here for the full traversal flow. These flows use the `graphify query` CLI when it is available and fall back to an inline NetworkX traversal otherwise.
 
 Two traversal modes - choose based on the question:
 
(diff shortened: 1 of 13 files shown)


graphify/skills/claude/references/query.md
```diff
@@ -9,62 +9,127 @@ Two traversal modes - choose based on the question:
...
+from networkx.readwrite import json_graph
+import networkx as nx
+from pathlib import Path
+
+data = json.loads(Path('graphify-out/graph.json').read_text())
+G = json_graph.node_link_graph(data, edges='links')
+
+question = 'QUESTION'
+mode = 'MODE'  # 'bfs' or 'dfs'
+terms = [t.lower() for t in question.split() if len(t) > 3]
```
uditstocks: # Bug: Step 0 vocab tokens silently dropped by Step 1's length filter

**Step 0 extracts tokens with a minimum length of 3:**

```
if 3 <= len(t) <= 30:
    vocab.add(t)
```
**But the inline traversal in Step 1 filters with:**
```
terms = [t.lower() for t in question.split() if len(t) > 3]
```

> len(t) > 3 means minimum 4 characters. So any 3-char token that Step 0 deliberately selected from the graph vocab **_(api, tok, auth, use, get)_** gets silently dropped here before the BFS/DFS even starts.

## **Suggested fix**:
Since the QUESTION passed into Step 1 is already the space-joined output of Step 0 (vocab-constrained, quality-filtered), the length check is redundant and harmful here:

```
terms = [t.lower() for t in question.split()]
```
If you want to keep some noise filtering, at minimum align with Step 0's threshold:
```
terms = [t.lower() for t in question.split() if len(t) >= 3]
```

uditstocks (commented):
Spotted a minor inconsistency between Step 0 and Step 1.
**inline comment added on line 49.**

graphify/skills/claude/references/query.md
burrows99: Fixed in 5643c7f — changed `len(t) > 3` to `len(t) >= 3` on line 49 to align Step 1 with Step 0's threshold. Thanks for catching this!

burrows99:
Also addressed the reviewer's observation cross-platform in f485fc2.

The `len(t) > 3` bug (Step 1 silently dropping 3-char tokens that Step 0 deliberately selected) was present in all 13 platform skills, not just `claude/`. Fixed in the same commit:

```
amp, claw, codex, copilot, droid, kilo, kiro, opencode, pi, trae, vscode, windows
```

All 13 platforms now use `len(t) >= 3` on line 49, consistent with Step 0's `if 3 <= len(t) <= 30` threshold.

LeanderBlume:
As far as I can tell, none (or very few, I haven't looked at all of them) of the shipped skills include this step 0. Looks like skillgen wasn't done correctly?

**Edit:** Nevermind, I figured out that none of the other platforms have it because they use the cli-inline query variant. Not sure why or why that one doesn't include this step 0 (which seems pretty important to get good results), but that's a different issue.

Issue #1145, opened 2026-06-06
bug: AST and semantic extractors generate different node IDs for the same symbol → god nodes duplicated, cluster cohesion degraded


lmngoldtrade-arch:
## Summary

On a single `/graphify <path>` run, the AST extractor and the semantic-extraction subagents produce **different node IDs for the same code symbol** because they disagree on two things:

1. **Stem derivation.** AST uses `<parent_dir>_<filename>` (e.g. `harness_runner` from `tests/harness/runner.ts`). The SKILL.md prompt tells semantic subagents to use just `<filename>` ("stem is the filename without extension"). Result: AST emits `harness_runner_skillevaluationrunner`, semantic emits `runner_skillevaluationrunner` for the same class.
2. **`source_file` prefix.** AST records paths relative to the scan root's parent (paths start `tests/...`). Semantic subagents record paths as given to them (paths start `skills/tests/...`). Same file, two different `source_file` values.

Because IDs differ, the Part C merge in SKILL.md (`if n['id'] not in seen`) treats them as separate nodes. Every god-node in a sufficiently-sized run ends up doubled: one AST copy with `source_location=L<n>` and one semantic copy with `source_location=None`. The two copies land in different Louvain communities (the AST copy lives in a tight functional cluster; the semantic copy gets pulled into a sprawling cross-references cluster), which inflates the community count and tanks cohesion scores for the affected clusters.

## Environment

- graphifyy via `uv tool install` on Windows 10 / Python 3.13 (git bash shell)
- Claude Code with the `graphify` skill installed via `graphify install --platform claude`
- Run command: `/graphify skills` on a vendored gstack `skills/` bundle (187 files, ~221k words)
- No `GEMINI_API_KEY` set → semantic extraction ran via parallel Claude general-purpose subagents per SKILL.md Step 3B

## Repro / Evidence

From the resulting `graphify-out/graph.json`, the same nine god-node symbols each appear twice:

```
"SkillEvaluationRunner" — 2 nodes:
    id=harness_runner_skillevaluationrunner          src=tests/harness/runner.ts          loc=L67   (AST)
    id=runner_skillevaluationrunner                  src=skills/tests/harness/runner.ts   loc=None  (semantic)
"CodeEvaluator" — 2 nodes:
    id=harness_evaluator_codeevaluator               src=tests/harness/evaluator.ts       loc=L40   (AST)
[...]

PR #1258, opened 2026-06-11
fix(build): skip ghost-merge when basename+label matches multiple AST nodes (#1257)

### LINKED ISSUE #1257: Ghost-merge picks an arbitrary AST node when two same-named files define the same symbol
## Problem

The #1145 ghost-merge in `graphify/build.py` (`build_from_json`) collapses semantic ghost nodes (no `source_location`) into AST nodes sharing the same `(source_file basename, label)` key:

```python
_loc_nodes: dict[tuple[str, str], str] = {}   # (basename, label) -> AST node id
...
if attrs.get("source_location"):
    _loc_nodes[(basename, label)] = nid
```

`_loc_nodes` is last-writer-wins. When two files with the same basename in different directories both define a symbol with the same label — e.g. `render` in `src/a/index.ts` and `src/b/index.ts`, a very common pattern with `index.ts`/`__init__.py`/`main.go` naming — the key is ambiguous, and the ghost is merged into whichever located node happened to be written last. Since iteration is over a `set`, the winner is arbitrary and can flip run-to-run, mis-pointing all of the ghost's edges to the wrong file's symbol and making the serialized graph churn.

Reproduction:

```python
ext = {
    "nodes": [
        {"id": "a_render", "label": "render", "file_type": "code",
         "source_file": "src/a/index.ts", "source_location": "L10"},
        {"id": "b_render", "label": "render", "file_type": "code",
         "source_file": "src/b/index.ts", "source_location": "L20"},
        {"id": "ghost_render", "label": "render", "file_type": "code",
         "source_file": "src/a/index.ts"},
        {"id": "caller", "label": "main", "file_type": "code",
         "source_file": "src/main.ts", "source_location": "L1"},
    ],
    "edges": [{"source": "caller", "target": "ghost_render", "relation": "calls",
               "confidence": "EXTRACTED", "source_file": "src/main.ts", "weight": 1.0}],
    "input_tokens": 0, "output_tokens": 0,
}
G = build_from_json(ext)
# ghost_render is merged into a_render OR b_render, depending on set iteration order
```

Related but distinct from #952, which is about the same kind of basename collision at the extractor level; this one is in the graph-build ghost-merge pass and persists even with correct extractor output.

## Proposed solution

Track keys shared by 2+ located nodes and skip the ghost-merge for those keys — only merge when the `(basename, label)` key identifies exactly one AST node. This mirrors the conservatism already used by `_rewire_unique_stub_nodes` in `extract.py` (`len(candidates) != 1: continue`).

## Acceptance criteria

- A ghost whose `(basename, label)` key matches 2+ located nodes is left intact, along with its edges.
- A ghost matching exactly one located node still merges as today.
- Regression test covering the two-`index.ts` collision case.

teemow:
Fixes #1257.

## Summary

- The #1145 ghost-merge pass in `build_from_json` keys located (AST) nodes by `(source_file basename, label)` in a last-writer-wins dict. When two files with the same basename in different directories both define a symbol with the same label (e.g. `render` in two `index.ts` files), the semantic ghost is merged into whichever located node was written last — and since iteration is over a `set`, the winner is arbitrary and can flip run-to-run, mis-pointing the ghost's edges and churning the serialized graph.
- Fix: track keys shared by 2+ located nodes and skip the ghost-merge for those keys. Ghosts only merge when the key identifies exactly one AST node — the same conservatism `_rewire_unique_stub_nodes` applies at the extractor level (`len(candidates) != 1: continue`).

Related but distinct from #952, which covers basename collisions at the extractor level; this pass mis-merges even with correct extractor output.

## Test

Regression tests in `tests/test_build.py`:

- `test_ghost_merge_skipped_on_basename_collision`: two `index.ts` files in different dirs, each with a located `render`, plus one ghost `render` — the ghost and its edge must survive untouched. Fails before the fix (ghost merged into an arbitrary node), passes after.
- `test_ghost_merge_unique_located_node_still_merges`: guards the existing happy path — a ghost with exactly one located counterpart still collapses and its edges re-point.

Full suite: 1985 passed, 27 skipped (`uv run --frozen pytest tests/ -q`).

### DIFF
--- a/graphify/build.py
+++ b/graphify/build.py
@@ -163,7 +163,12 @@ def build_from_json(extraction: dict, *, directed: bool = False, root: str | Pat
     # pairs that share (source_file basename, label) and collapse the semantic
     # copy into the AST copy so edges re-point to a single node.
     # Two passes: first collect all AST (located) nodes, then find ghosts.
+    # When 2+ located nodes share a key (same-named symbols in same-named
+    # files across directories, e.g. render in two index.ts), the key is
+    # ambiguous: merging the ghost would pick an arbitrary winner. Skip those
+    # keys entirely (same conservatism as _rewire_unique_stub_nodes).
     _loc_nodes: dict[tuple[str, str], str] = {}   # (basename, label) -> AST node id
+    _loc_collisions: set[tuple[str, str]] = set()  # keys shared by 2+ AST nodes
     _noloc_nodes: dict[tuple[str, str], str] = {}  # (basename, label) -> semantic node id
     for nid in node_set:
         attrs = G.nodes[nid]
@@ -173,7 +178,11 @@ def build_from_json(extraction: dict, *, directed: bool = False, root: str | Pat
         if not label or not basename:
             continue
         if attrs.get("source_location"):
-            _loc_nodes[(basename, label)] = nid
+            key = (basename, label)
+            if key in _loc_nodes:
+                _loc_collisions.add(key)
+            else:
+                _loc_nodes[key] = nid
     for nid in node_set:
         attrs = G.nodes[nid]
         label = str(attrs.get("label", "")).strip()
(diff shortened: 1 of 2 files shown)


safishamsi:
The ghost-merge collision bug (#1257) has been fixed directly in v8 at commit b830862. The fix was re-implemented against the current `_origin=="ast"` structure introduced by cce2673 (which this PR conflicted with), adding `_loc_collisions` tracking so ambiguous AST-vs-AST key collisions are skipped rather than merging into an arbitrary winner. Closing as superseded.

PR #1270, opened 2026-06-11
fix: emit symbol edges for default imports/exports

jevans-agent:
## Problem

JS/TS symbol resolution (`_collect_js_symbol_resolution_facts`) only handles **named** imports. A default-exported symbol — `export default class Foo {}` imported as `import Foo from './foo'` — produces only a file→file `imports_from` edge; the class node receives **no incoming symbol edge**.

On codebases that default-export most of their classes (NestJS-style services/helpers/models, factories, etc.) the effect is large:
- `graphify affected "<Class>"` and `graphify explain "<Class>"` report **no callers**, because they seed the class symbol whose only in-edge is `contains`.
- Those symbols look like isolated leaf nodes.

Concrete example from a ~960-file NestJS API: a service class with 18 real importers showed **0** incoming symbol edges. After this change it has direct `imports` edges from all 18 (plus `references` edges from call sites), and `affected`/`explain` work as expected.

## Fix

Entirely in fact collection — the exported-origin resolver is already generic:
- Record default imports with `imported_name="default"` (`_js_default_import_name`), including the mixed form `import Foo, { Bar } from './x'`.
- Register a `"default"` export for `export default <class|function|abstract class|identifier>` (`_js_default_export_name`).
- The existing `resolve_exported_origin` then wires the `imports` edge to the origin symbol, and — because the default binding now lives in the file's alias table — calls/refs through the local binding resolve too, **even when the binding is renamed** (`import Bar from './foo'; new Bar()`).

`export { X as default }` was already handled via the export-clause path. Anonymous defaults (`export default class {}`, `export default {...}`) have no resolvable symbol and stay file-level only.

## Tests

Adds four regression tests (default-export class, `export default <identifier>`, renamed-binding import, and a renamed-binding **call** that can only resolve via the default-import alias). Each fails on `v8` and passes with the fix. Full `pytest tests/` (1987 passed) + `ruff` green.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 Full release notes with details on each version: [GitHub Releases](https://github.com/safishamsi/graphify/releases)
 
+## Unreleased
+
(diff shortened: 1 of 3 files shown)


safishamsi:
Merged manually after rebasing onto v8 (CHANGELOG.md conflict resolved). Changes are in v8 @ 6dc23db.

Issue #1178, opened 2026-06-07
Bug Report — `graphify --update` performs a destructive, net-negative fuzzy node merge on an already-current code graph


MethCDN:
# Bug Report — `graphify --update` performs a destructive, net-negative fuzzy node merge on an already-current code graph 

## **Report Bug write by Opus 4.8**

- **Date observed:** 2026-06-07 (~14:05–14:15 local, GMT+2)
- **graphify package:** `graphifyy`, installed via `uv tool` (interpreter path redacted) 
- **Skill version:** `last version at this day/data/
- **Host:** Windows 11 (win32), PowerShell 7
- **IA Model: Opus 4.8 xhigh. | Claude version version 2.1.165 
- **Project type:** a code-only TypeScript/React (Electron/Vite-class) corpus
- **Severity:** High (silent data loss is *prevented* only by the `to_json` guard; if a user/agent passes `force=True`, 16% of the graph is destroyed)
- **Status:** Reproduced, root-caused, damage averted (graph left intact)

> **Note:** all project-specific names, file paths, source-file names, symbol/identifier names, dependency names, and commit hashes have been replaced with generic placeholders (`<PROJECT_ROOT>`, `<ClassA>`, `<functionA>()`, `<source-file>`, `<third-party-dep-N>`, `<COMMIT_HASH>`, …).
---

## TL;DR

Running the manual incremental update (`/graphify --update`) on a graph that the in-session auto-updater had **already brought current** caused the merge step to produce a graph with **fewer nodes than it started with**:

- Existing `graph.json`: **1515 nodes / 2435 edges**
- After `build_merge`: **1272 nodes / 2000 edges**
- **Net result: −243 nodes, −435 edges, +0 new nodes**

The 243 lost nodes were **distinct, real symbols** (third-party dependency nodes, a service class plus several of its methods, standalone functions, persistence modules, test fixtures), collapsed by `build_merge`'s **global fuzzy deduplication** (`Deduplicated 243 node(s): 4 exact, 234 fuzzy`). The new AST extraction contributed **zero** genuinely new nodes — all 476 re-extracted node IDs already existed in the graph — so the merge had nothing to add and the fuzzy pass was the *only* operation with any effect, and that effect was purely subtractive.

[...]

Issue #1152, opened 2026-06-06
--update: ghost nodes survive for symbols deleted from re-extracted (modified) files (grow-only merge never replaces per source_file)


janreview:
**Version:** graphifyy 0.8.33 · Windows 10 · Python 3.12

## Summary

Incremental update (`build_merge`) leaves **ghost nodes** for symbols that were deleted from files that still exist. `build_merge` is documented as "Never replaces - only grows (or prunes deleted-file nodes via prune_sources)" — which means a function removed from a *modified* file can never leave the graph through the documented `--update` flow, even though the file was just re-extracted by AST.

## Repro

1. Full graph build. `src/database/queries.py` contains `get_default_pricing()` → node `database_queries_get_default_pricing` exists.
2. Delete the function from the file (file itself stays). Re-run the incremental flow: `detect_incremental` correctly lists `queries.py` as changed; AST re-extracts it (the new extraction has **no** `get_default_pricing` node).
3. `build_merge([new_extraction], graph_path=...)` → the old node **survives** (grow-only merge keeps every existing node).

Measured on a real dead-code-removal commit: 12 functions deleted across 8 modified files → all 12 still present as nodes after `--update`, with their stale `calls` edges. For a tool whose report advertises dead-code-ish insights (god nodes, isolated nodes), the graph silently disagreeing with the codebase after an update is a trap.

## Why this is safe to fix for AST nodes

AST extraction is deterministic and **complete per file**: when a changed file is re-extracted, the new chunk contains the full current symbol set for that file. So for every `source_file` present in `new_chunks`, any *existing* node with the same `source_file` (and `file_type == "code"`) that is absent from the new chunk is provably stale and can be dropped before the grow-merge.

Semantic (doc/concept) nodes are trickier (a chunk may intentionally be partial), so replace-by-source-file could apply to AST/code nodes only, or be opt-in:

```python
build_merge(chunks, graph_path=..., replace_sources=[...changed code files...])
```

## Workaround used

Post-merge: collect node ids whose `source_file` is among the re-extracted code files and whose id is not in the new extraction, drop them + their edges. Verified 0 ghosts afterwards.

Related: filed alongside the `prune_sources` path-mismatch issue — both surfaced from the same `--update` run after a dead-code-removal commit.

ayushcodes10:
Already fixed on current `v8`.

[...]

PR #1285, opened 2026-06-12
fix: prune base graph before merging in build_merge so modified files keep their fresh nodes

### LINKED ISSUE #1283: build_merge(prune_sources=...) prunes AFTER merging — modified files passed per the bundled skill update templates lose their freshly re-extracted nodes
### Summary

`build_merge()` applies `prune_sources` **after** merging `new_chunks` into the base graph. Any node/edge whose `source_file` is in the prune set is removed from the **merged** graph — including the freshly re-extracted nodes that were just passed in `new_chunks`. So the documented incremental flow for a *modified* file (re-extract it, pass it in `prune_sources` to drop the stale copies) deletes the new extraction too: the changed file ends up with **zero nodes**.

This is not a hypothetical misuse: the bundled skill templates (`graphify/skills/*/references/update.md`, step "Also prune old nodes for re-extracted (changed) files before inserting fresh AST", added for #1178) explicitly build `prune = deleted + changed` and call:

```python
G = build_merge([new_extraction], graph_path='graphify-out/graph.json', prune_sources=prune)
```

Under the current ordering, every `--update` driven by those templates wipes the changed files' fresh nodes right after extracting them.

### Reproduction (verified on 0.8.22, 0.8.37, 0.8.38, and current `main`)

```python
import json, tempfile
from pathlib import Path
from graphify.build import build_merge

tmp = Path(tempfile.mkdtemp()); gp = tmp / "graph.json"
old = {"directed": False, "multigraph": False, "graph": {},
  "nodes": [
    {"id": "x_old", "label": "Old A", "file_type": "document", "source_file": "x.md"},
    {"id": "y_b", "label": "B", "file_type": "document", "source_file": "y.md"},
  ],
  "links": [{"source": "x_old", "target": "y_b", "relation": "references",
             "confidence": "EXTRACTED", "confidence_score": 1.0, "source_file": "x.md"}]}
gp.write_text(json.dumps(old))

# x.md was MODIFIED — fresh re-extraction:
chunk = {"nodes": [{"id": "x_new", "label": "New A", "file_type": "document", "source_file": "x.md"}],
         "edges": [{"source": "x_new", "target": "y_b", "relation": "references",
                    "confidence": "EXTRACTED", "confidence_score": 1.0, "source_file": "x.md"}]}

G = build_merge([chunk], graph_path=gp, prune_sources=["x.md"], dedup=False)
print(sorted(G.nodes()))   # ['y_b']  ← x_new was pruned together with x_old
```

Output: `[graphify] Pruned 2 node(s) from 1 deleted source file(s).` — the "2" includes the fresh node.

### Why this matters / interaction with existing issues

The API currently has no correct option for modified files:

- **Don't** pass changed files in `prune_sources` → stale symbols from previous versions of the file survive forever (grow-only merge; that's #1152).
- **Do** pass them (as the bundled skill templates instruct) → the fresh extraction is destroyed.

For context: the same stale-symbol leak in the `watch._rebuild_code` path (#1116) was fixed in #1118 by stamping `_origin="ast"` and dropping absent AST nodes on rebuild — but `build_merge` never got an equivalent per-file replace mechanism, which is why #1152 remains open and the skill templates fall back to `prune = deleted + changed`.

### Proposed fix

Prune the **base** graph (the nodes/edges loaded from `graph.json`) *before* merging `new_chunks`, instead of pruning the merged result. Then:

- modified files become an atomic per-file replace: stale copies dropped, fresh chunk kept — which also gives `--update` flows the mechanism to eliminate the #1152 ghosts;
- genuinely deleted files (no chunk) behave exactly as today;
- base edges from unchanged files that pointed at a pruned node become dangling and are already dropped inside `build()` (same outcome as `G.remove_nodes_from()`'s implicit incident-edge removal);
- the #1007 dual-form path matching (raw + `_norm_source_file`) is preserved, applied to the as-stored base JSON (including legacy `"source"`-keyed nodes, which only gain `source_file` during `build_from_json`).

**Fix submitted: PR #1285** — prune-before-merge with three regression tests (changed-file replace, #1152 ghost-drop, legacy `"source"`-keyed nodes); full test suite passes (2024 passed, 27 skipped).

van4oza:
Fixes #1283.

### Problem

`build_merge()` merged the base graph + `new_chunks` first, then pruned every node/edge whose `source_file` ∈ `prune_sources` from the **merged** graph. For *modified* files — re-extracted into `new_chunks` and passed in `prune_sources` to drop their stale copies — the prune deleted the fresh extraction too: the changed file ended up with zero nodes (`[graphify] Pruned 2 node(s) …` on a 1-node file).

The bundled skill update templates (`graphify/skills/*/references/update.md`) already instruct exactly that flow (`prune = deleted + changed`, added for #1178), so the shipped documentation triggers the data loss on every `--update` with modified files. Not passing changed files isn't an option either — that leaves stale-symbol ghosts (#1152). The API had no correct choice for modified files.

### Fix

Prune the **base** graph (the `existing_nodes` / `existing_edges` loaded from `graph.json`) *before* the merge:

- **Modified files become an atomic per-file replace** — stale copies dropped, fresh chunk nodes always survive. This also gives `--update` flows the mechanism to eliminate the #1152 ghosts (the skill templates' `deleted + changed` prune now does what its comment intends).
- **Genuinely deleted files behave exactly as before** (existing prune tests for #1007 absolute/Windows paths pass unchanged).
- Base edges from unchanged files that pointed at a pruned node become dangling and are dropped inside `build()` — same outcome as the old `G.remove_nodes_from()`'s implicit incident-edge removal (`build_from_json` only adds edges whose endpoints exist).
- The #1007 dual-form matching (raw + `_norm_source_file`) is preserved; since the base JSON is now checked as-stored (the old code compared against build-time-normalised attributes), the filter also falls back to the legacy `"source"` node key that `build_from_json` migrates — without the fallback, legacy-keyed nodes would have escaped the prune. Edges deliberately get no such fallback (their `source` is an endpoint id, not a path).
- The `#479` shrink-guard and all stderr messages are unchanged.

### Tests

Three new regression tests in `tests/test_build.py`:

- `test_build_merge_prune_changed_file_keeps_reextracted_nodes` — fresh node + fresh edge survive; stale node + stale edge pruned; unrelated file untouched (fails on the previous ordering).
- `test_build_merge_prune_changed_file_drops_removed_symbols` — a symbol deleted from a modified file does not survive as a ghost (#1152 scenario).
- `test_build_merge_prune_matches_legacy_source_keyed_nodes` — legacy `"source"`-keyed nodes are still pruned.

Full suite: **2024 passed, 27 skipped**.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,8 @@ Full release notes with details on each version: [GitHub Releases](https://githu
 
 ## Unreleased
 
+- Fix: `build_merge(prune_sources=...)` now prunes the base graph **before** merging new chunks. Pruning previously ran on the merged graph, so an `--update` that passed *modified* (re-extracted) files in `prune_sources` deleted their freshly extracted nodes too — the changed file ended up with zero nodes. With base-first pruning, passing changed files atomically replaces their nodes (stale copies dropped, fresh chunk kept), which also eliminates the stale-symbol ghosts of #1152; genuinely deleted files behave exactly as before. (#1283)
+
 ## 0.8.38 (2026-06-11)
 
 - Fix: LLM-generated `calls` edges now have correct direction. The extraction prompt previously never stated that `source` = caller and `target` = callee; the LLM systematically emitted callee→caller edges. An explicit direction rule was added to the prompt. Separately, ghost-node merge was extended to collapse LLM duplicate nodes (bare-stem IDs) onto AST canonical nodes (parent-qualified IDs) even when the LLM node carries a `source_location` — the old check only caught `source_location=None` ghosts. Post-fix annotation: `calls` precision 100% (n=6), overall INFERRED precision 94% (n=16).
--- a/graphify/build.py
+++ b/graphify/build.py
@@ -390,6 +390,12 @@ def build_merge(
     Never replaces - only grows (or prunes deleted-file nodes via prune_sources).
     Safe to call repeatedly: existing nodes and edges are preserved.
     root: if given, absolute source_file paths in new_chunks are made relative (#932).
+
+    prune_sources is applied to the BASE graph before the merge, so it is safe
+    to pass *modified* files together with their fresh re-extraction in
+    new_chunks: the stale copies are dropped and the fresh nodes survive
+    (atomic per-file replace). Genuinely deleted files (no chunk) are simply
+    removed, as before.
     """
     graph_path = Path(graph_path)
     if graph_path.exists():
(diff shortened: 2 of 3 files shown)


safishamsi:
Fixed by #1344 (merged to v8): `build_merge` now prunes a re-extracted file's prior nodes/edges *before* merging the fresh extraction, so modified files are replaced rather than accumulating stale edges. Same fix that closed #1283.

Issue #1243, opened 2026-06-10
Fuzzy dedup over-merges distinct entities that share a common prefix (JaroWinkler prefix-bonus; #1201 guard only covers strict prefix-extension)


Perf:
## Summary

`deduplicate_entities` (Pass 2, fuzzy/JaroWinkler) merges **distinct entities that share a common prefix but diverge in the middle or tail**. The recent prefix-extension guard (`dedup.py`, "Prefix-extension pairs … #1201") only blocks the special case where one label is a *strict* prefix-extension of the other (`_hi.startswith(_lo)`). It does **not** cover the much more common case where two labels share a long prefix and then differ — which JaroWinkler's prefix bonus still scores ≥ the 92 merge threshold.

The result is silent collapse of genuinely different nodes: two different npm packages, two different migration files, a schema vs. its `…List` sibling, FE vs. MOB docs, etc.

Reproduced on `graphifyy==0.8.36`. The v0.8.35 (#1201, prefix-extension) and v0.8.37 (identical same-named symbol cross-file) changelog fixes do not address this case.

## Self-contained repro

```python
from graphify.dedup import deduplicate_entities

cases = {
    "two different npm packages": [
        {"id": "a", "label": "@testing-library/jest-native",  "source_file": "package.json"},
        {"id": "b", "label": "@testing-library/react-native", "source_file": "package.json"},
    ],
    "two different migration snapshots": [
        {"id": "a", "label": "0000_snapshot.json", "source_file": "meta/0000_snapshot.json"},
        {"id": "b", "label": "0001_snapshot.json", "source_file": "meta/0001_snapshot.json"},
    ],
    "schema vs its List variant": [
        {"id": "a", "label": "AdminConnectedSourceSchema",     "source_file": "admin.ts"},
        {"id": "b", "label": "AdminConnectedSourceListSchema", "source_file": "admin.ts"},
    ],
    "FE vs MOB design doc": [
        {"id": "a", "label": "2026-05-17-phase4-fe-api-integration.md",  "source_file": "a.md"},
        {"id": "b", "label": "2026-05-17-phase4-mob-api-integration.md", "source_file": "b.md"},
    ],
}
for name, nodes in cases.items():
    comm = {n["id"]: 1 for n in nodes}
    out, _ = deduplicate_entities([dict(n) for n in nodes], [], communities=comm)
    print(("MERGED (bug)" if len(out) < len(nodes) else "kept distinct").ljust(14), name)
```

[...]

Issue #1284, opened 2026-06-12
Fuzzy dedup still merges distinct non-code nodes across files (near-identical labels) — rationale/document/concept lack the #1205 protection


van4oza:
### Summary

#1205 fixed label-based dedup for code symbols (`file_type == "code"` is now exempt from both exact and fuzzy passes), and the exact-label cross-file block (#1046/#895) works. But **non-code nodes — `rationale`, `document`, `concept` — get no equivalent protection**, and pass-2 fuzzy matching still merges *distinct* entities across files whenever their labels are merely near-identical (Jaro-Winkler above `_MERGE_THRESHOLD` but not exactly equal, so the cross-file same-label block never fires).

### Evidence (full merge audit, real graph)

Method: on an 11.7k-node graph built from a Django monorepo + its docs (graphifyy 0.8.37), I wrapped `graphify.dedup._UF.union` to log every union call, ran `deduplicate_entities`, and reconstructed every merge group:

- 145 groups: same normalized label, same `source_file` → benign duplicate collapse ✅
- 0 groups: same label across files → the #1046 guard holds ✅
- **49 groups with different labels — many destructive:**
  - **15 distinct Django `AppConfig` rationale nodes merged into 1.** Every app's `apps.py` docstring is the same boilerplate ("Django app config for `apps.<name>`. No business logic here…") differing only in the module path → JW ≈ 0.95+. Fifteen different files, one surviving node.
  - **Antonym metrics merged:** a success criterion label `40%+ <metric>` absorbed the failure criterion `<20% <metric>` from the same PRD — opposite meanings, high string similarity.
  - **Three distinct SQL DDL fence nodes → 1:** `code:sql (create table topics (` absorbed `…topic_items (` and `…topic_candidates (`.
  - **Parallel provider modules merged:** "Structured devtools facts…" (`providers/github.py`) ⇄ "Structured travel facts…" (`providers/places.py`) — same boilerplate shape, different adapters.
  - `VideoAnalyzer` ⇄ `FakeVideoAnalyzer` (interface vs test fake), `…(ADR 0013 D4)` ⇄ `…(ADR 0011 §D5)`, `Module layout (… VAN-375)` ⇄ `(… VAN-376)`, section headings `3.1 Product Goals` (doc A) ⇄ `1.1 Product Goals` (doc B).

[...]

Issue #1205, opened 2026-06-09
--update / build_merge dedup merges distinct same-named code symbols


Hammer7:
# `--update` / `build_merge` dedup merges distinct same-named code symbols

## Summary

`deduplicate_entities` (run by `build()`, hence by `build_merge` and the `--update`
flow) merges distinct **code** symbols that merely share a label. It keys code-node
identity on `(normalized_label, source_file)` and on fuzzy label similarity, which
collapses unrelated symbols.

On a real ~6,000-node code graph, a single `--update` dropped ~1,300 nodes —
distinct methods and sibling classes silently conflated into one. The full-build
path (`build_from_json`) is **unaffected** because it does not run
`deduplicate_entities`, so `--update` and a clean rebuild of the same corpus
diverge significantly.

## Environment

- graphifyy 0.8.35
- Python 3.14, Linux

## Root cause

Code-node identity is the node **ID** (which already encodes the qualified
path), not the label. Two passes in `deduplicate_entities` treat the label as
identity:

1. **Pass 1 (exact normalized label, same file).** AST emits method labels
   without the owning class (e.g. `.draw()`), so several distinct methods in one
   file normalize to the same key and are merged.
2. **Pass 2 (fuzzy, cross-file).** Functions/classes whose long names differ
   slightly (a trailing word, or two parallel backend implementations) fuzzy-match
   and merge. The existing equal-label cross-file guard only blocks *identical*
   labels, not similar ones. A code↔concept fuzzy match can also transitively
   union two distinct code symbols through a shared concept node.

Genuinely-duplicate code nodes (the *same* symbol re-extracted — e.g. existing
graph + new chunk during `build_merge`) are already collapsed by the exact-ID
pre-dedup (`seen_ids`), so code never needs label-based merging at all.

## Reproduction

```python
from graphify.dedup import deduplicate_entities

def code(nid, label, src):
    return {"id": nid, "label": label, "file_type": "code", "source_file": src}

# Pass 1: three distinct methods sharing a bare method-name label in one file
a = [code("m_basic_draw", ".draw()", "renderer.py"),
     code("m_fast_draw",  ".draw()", "renderer.py"),
     code("m_slow_draw",  ".draw()", "renderer.py")]

# Pass 2: two distinct functions, similar long names, different files
b = [code("dense_build_feature_matrix_dense",   "build_feature_matrix_dense",  "dense.py"),
[...]

Issue #1046, opened 2026-05-27
Pass 2 fuzzy dedup merges identical-label nodes across unrelated source_files (leaks #895's fix)


RobRipley:
# Pass 2 fuzzy dedup merges identical-label nodes across unrelated source files

## Summary

`dedup.py:deduplicate_entities()` Pass 1 was fixed in #895 to only merge nodes within the same `source_file`. Pass 2 (MinHash+Jaro-Winkler, `dedup.py` L195-254) still freely merges across files. When two distinct nodes have **identical** normalized labels in different files, they hit Pass 2 with JW score = 100 and silently pass through every guard (`_is_variant_pair`, `_short_label_blocked`) because those guards only block *similar-but-not-equal* cases.

Result: the fix for #895 leaks straight through Pass 2 for the most common collision pattern — identical method names across modules.

## Reproducer (real corpus: 605 files, ~880k words, mixed Rust/TS/Svelte)

After `/graphify --update` with the merged extraction, build_merge reported:

```
[graphify] Deduplicated 792 node(s) (253 exact, 475 fuzzy).
```

Of the 475 fuzzy merges, ~231 are cross-`source_file`. Spot-checking 200 of them, ~150 are false positives (different functions with the same name in different files). Concrete examples:

| Old node (`source_file`) | New node (`source_file`) | Issue |
|---|---|---|
| `state.borrow_from_vault()` (`src/rumi_protocol_backend/src/state.rs`) | `main.borrow_from_vault()` (`src/rumi_protocol_backend/src/main.rs`) | Same canister, state-layer method merged with main-layer endpoint wrapper |
| `analytics/storage/mod.from_bytes()` (`src/rumi_analytics/src/storage/mod.rs`) | `3pool/storage.from_bytes()` (`src/rumi_3pool/src/storage.rs`) | Different canisters, different `Storable::from_bytes` impls |
| `threePoolService.getSwapEventCount()` | `ammService.getSwapEventCount()` | Different frontend services calling different canisters |
| `icusd_ledger.did.d.ts:Transaction` | `rumi_amm/lib.rs:Transaction` | Same name, completely different types (Candid TS binding vs Rust struct) |
| `GuardPrincipal::new()` (`guard.rs`) | `ICP::new()` (`numeric.rs`) | Different types, same trait method name |
| `InterpolationMethod::default()` | `Mode::default()` | Different types, same trait method name |

[...]

PR #1286, opened 2026-06-12
fix: block fuzzy-dedup merges of numbered siblings and cross-file rationale boilerplate (#1284, proposals 1+2)

van4oza:
Implements proposals **1 + 2** from #1284. Submitted as a **draft** deliberately — #1284 lists three possible directions and the choice is yours; this PR makes the two least-controversial ones concrete and reviewable. Proposal 3 (restricting cross-file fuzzy merging to `concept` nodes) is intentionally omitted as the more opinionated call.

### Problem (from #1284)

#1205 exempted code symbols from label-based merging and #1046 blocks exact-label cross-file pairs, but non-code nodes (`rationale`/`document`/`concept`) with *near*-identical labels still fuzzy-merge across files. On a real 11.7k-node graph: 49 destructive different-label merge groups, including 15 distinct Django `AppConfig` docstring rationale nodes → 1, antonym metric labels (`40%+ X` absorbed `<20% X`), and ADR/issue-numbered sibling notes.

### Change

Two targeted guards in pass 2 (placed with the existing `_is_variant_pair` / `_short_label_blocked` / prefix-extension guards, and mirrored in `_llm_tiebreak`'s ambiguous-pair collection so decisively-distinct pairs never reach the LLM):

1. **`_numeric_tokens_differ`** — digit runs of the normalized labels compared as zero-padding-insensitive multisets (`"09"` ≡ `"9"`; string-normalized rather than `int()` so a pathological >4300-digit run can't hit Python's int-conversion limit). Any mismatch blocks the merge: `ADR 0011 §D5` vs `ADR 0013 D4`, `3.1 Product Goals` vs `1.1 Product Goals`, `block3` vs `block13`, `40%+ …` vs `<20% …` are numbered siblings, not duplicates. Digitless labels are unaffected.
2. **`_crossfile_rationale_blocked`** — `rationale` nodes are docstring-derived and as file-anchored as the code they describe (#1205's reasoning, one layer up). Label-based merging of a rationale node is allowed only within one file, so parallel modules' boilerplate ("Django app config for `apps.<name>`. No business logic…") no longer collapses; same-file rationale duplicates still merge.

Known false-negative trade-off (accepted): pairs like `OAuth2 flow` / `OAuth 2.0 flow` are now blocked. Given that this engine's observed failure mode is destructive merging, a missed merge is the cheaper error.

### Validation on the real graph that produced #1284

| | upstream `v8` | with this PR |
|---|---|---|
| different-label merge groups | **49** | **19** (−30) |
| benign same-label-same-file groups | 144 | 144 (unchanged) |
| nodes absorbed by fuzzy dedup | 249 | 195 |

All flagship destructive cases eliminated (15× AppConfig, antonym metrics, ADR 0034/0035, VAN-375/376, github/places providers, cross-doc numbered headings). The 19 survivors are predominantly legitimate same-file near-duplicates; the cross-file remainder is the class proposal 3 would address.

### Tests

4 new tests in `tests/test_dedup.py` (helper semantics incl. zero-padding; numbered-siblings blocked; cross-file rationale blocked; same-file rationale control still merges). Both guard tests verified to **fail** against the pre-patch `dedup.py`. Full suite: **2025 passed, 27 skipped**.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,8 @@ Full release notes with details on each version: [GitHub Releases](https://githu
 
 ## Unreleased
 
+- Fix: fuzzy dedup no longer merges numbered siblings or cross-file rationale boilerplate. Two new pass-2 guards (also mirrored in the `--dedup-llm` ambiguous-pair collection): (1) labels whose embedded digit runs differ as zero-padding-insensitive multisets (`ADR 0011` vs `ADR 0013`, `3.1 Product Goals` vs `1.1 Product Goals`, `block3` vs `block13`, `40%+ …` vs `<20% …`) are numbered/versioned siblings and never merge — zero-padding (`09` vs `9`) does not count as a difference; (2) `rationale` nodes are file-anchored like code (#1205's reasoning): parallel modules' near-identical docstring boilerplate ("Django app config for `apps.<name>`. No business logic…") no longer collapses across files, while same-file rationale duplicates still merge. On a real 11.7k-node graph these guards eliminated 30 of 49 destructive different-label merge groups, including a 15-node AppConfig collapse and antonym metric labels. (#1284)
+
 ## 0.8.38 (2026-06-11)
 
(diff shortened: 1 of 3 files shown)


safishamsi:
Thanks @van4oza - this is solid work and it's now incorporated on v8 in `e9f8dde`, closing #1284.

Your `_numeric_tokens_differ` guard is adopted as-is (the zero-padding-insensitive multiset comparison and the string-not-int() reasoning are exactly right), and your 4 tests are in the suite. I extended the second guard from `rationale`-only to `rationale` + `document` (`_crossfile_fileanchored_blocked`), since #1284's title lists both and document nodes are equally file-anchored (headings/positional content). Both guards are mirrored into `_llm_tiebreak` as in your PR.

The same commit also closes #1243 (the JaroWinkler prefix-bonus over-merge) by scoring cross-file long labels on plain Jaro - which subsumes the residual concept-node cases your proposal 3 would have addressed, without a blanket cross-file block.

Crediting you via `Co-Authored-By` on the commit. Closing this PR as incorporated. Much appreciated!

Issue #1299, opened 2026-06-13
graphify query/explain/path crash on unreadable ~/.claude/skills/graphify/.graphify_version


scp7:
## Summary

`graphify query`, `graphify explain`, and `graphify path` can fail during startup if the skill-version probe hits an unreadable `.graphify_version` file under a platform-specific skills directory.

That makes an advisory compatibility warning behave like a hard failure.

## Reproduction

1. Install `graphify` in an environment where `~/.claude/skills/graphify/.graphify_version` exists but is unreadable.
2. Run a normal command such as:

```bash
graphify query "how is graphify implemented?"
```

## Actual behavior

The command fails with `PermissionError` before it reaches the normal query path.

## Expected behavior

The command should continue running. If the version probe cannot be read, it should be treated as non-fatal and skipped.

## Why this matters

The version probe is only a staleness warning. It should never block normal graph queries when a platform-specific skill directory is inaccessible.

## Suggested fix

Catch unreadable probe paths during startup, or skip unreadable skill directories before comparing `.graphify_version`.

safishamsi:
Fixed in v8 at b830862.

PR #1300, opened 2026-06-13
fix: ignore unreadable skill version probes

scp7:
## Summary

Treat unreadable skill-version probe files as non-fatal during startup.

## Change

- Skip or ignore unreadable `.graphify_version` paths
- Preserve the warning when the file is readable and the version is stale
- Add a regression test for the permission-failure case

## Notes

This keeps the compatibility warning behavior without making normal commands fail in sandboxed or restricted environments.

### DIFF
--- a/graphify/__main__.py
+++ b/graphify/__main__.py
@@ -94,9 +94,16 @@ def _enforce_graph_size_cap_or_exit(gp: Path) -> None:
 def _check_skill_version(skill_dst: Path) -> None:
     """Warn if the installed skill is from an older graphify version."""
     version_file = skill_dst.parent / ".graphify_version"
-    if not version_file.exists():
+    try:
+        if not version_file.exists():
+            return
+    except OSError:
         return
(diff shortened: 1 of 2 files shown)


safishamsi:
The implementation is correct — `_check_skill_version` properly returns on `OSError`. However the new test `test_check_skill_version_ignores_permission_error` fails because it calls `_install()` (which prints install output), then asserts `out.out == ""` without flushing the capsys buffer first.

**One-line fix:** add `capsys.readouterr()` after `_install(tmp_path, platform)` and before calling `_check_skill_version(skill)`.

Please push the fix and we can merge.

safishamsi:
Issue #1299 has been fixed directly in v8 at commit b830862 — all three FS probes in `_check_skill_version` are now wrapped in `try/except OSError: return`. The test fix (adding `capsys.readouterr()` after `_install()`) is also included. Closing as superseded.

Issue #1104, opened 2026-06-01
Unable to install with uv tool install graphifyy


majacque:
Hello, I tried to install graphify using uv on Ubuntu 24.04.4 LTS.
I first ran ` curl -LsSf https://astral.sh/uv/install.sh | sh` and then `uv tool install graphifyy` like it's said in the README.
But the package can't be install due to a problem with `tree-sitter-dm` dependence.
Here is the output:
```Resolved 33 packages in 492ms
  × Failed to build 'tree-sitter-dm==0.25.1'
  ├─▶ The build backend returned an error
  ╰─▶ Call to 'setuptools.build_meta.build_wheel' failed (exit status: 1)

      [stdout]
      running bdist_wheel
      running build
      creating build/lib.linux-x86_64-cpython-312/tree_sitter_dm/queries
      creating build/lib.linux-x86_64-cpython-312/tree_sitter_dm/queries/vscode
      copying queries/vscode/injections.scm -> build/lib.linux-x86_64-cpython-312/tree_sitter_dm/queries/vscode
      copying queries/vscode/highlights.scm -> build/lib.linux-x86_64-cpython-312/tree_sitter_dm/queries/vscode
      copying queries/vscode/tags.scm -> build/lib.linux-x86_64-cpython-312/tree_sitter_dm/queries/vscode
      creating build/lib.linux-x86_64-cpython-312/tree_sitter_dm/queries/neovim
      copying queries/neovim/injections.scm -> build/lib.linux-x86_64-cpython-312/tree_sitter_dm/queries/neovim
      copying queries/neovim/highlights.scm -> build/lib.linux-x86_64-cpython-312/tree_sitter_dm/queries/neovim
      copying queries/neovim/tags.scm -> build/lib.linux-x86_64-cpython-312/tree_sitter_dm/queries/neovim
      copying queries/highlights.scm -> build/lib.linux-x86_64-cpython-312/tree_sitter_dm/queries
      copying queries/tags.scm -> build/lib.linux-x86_64-cpython-312/tree_sitter_dm/queries
      running build_py
      copying bindings/python/tree_sitter_dm/__init__.py -> build/lib.linux-x86_64-cpython-312/tree_sitter_dm
      running egg_info
      writing bindings/python/tree_sitter_dm.egg-info/PKG-INFO
      writing dependency_links to bindings/python/tree_sitter_dm.egg-info/dependency_links.txt
[...]

PR #1307, opened 2026-06-13
feat: add Haxe language support via tree-sitter-haxe

mallyskies:
Adds Haxe (.hx) as a supported language for AST extraction.

**What this does:**

- `detect.py`: registers `.hx` in `CODE_EXTENSIONS`
- `extract.py`: adds `extract_haxe()` which extracts classes, interfaces,
  enums, enum abstracts, typedefs, and functions from `.hx` files using the
  `tree-sitter-haxe` grammar
- `extract.py`: adds `_haxe_recover_scattered()` fallback for files where
  the grammar emits scattered tokens instead of proper declaration nodes
  (minified files, unsupported preprocessor patterns)
- `pyproject.toml`: no `haxe` extra — `tree-sitter-haxe` has no PyPI
  release, and PyPI rejects packages with a direct URL/VCS dependency in
  `Requires-Dist`, so declaring one here would block every future
  `graphifyy` release. `extract_haxe()` lazy-imports `tree_sitter_haxe`
  with a graceful `ImportError` guard (same pattern as `dm`/`terraform`),
  so this is a pure packaging change with no functional impact.

**Implementation notes:**

CR/CRLF line endings are normalized before parsing — the codebase being
tested against contains legacy files with `\r`-only Mac line endings which
would cause the `//` comment rule to run to EOF.

The fallback (`_haxe_recover_scattered`) handles three patterns the grammar
currently struggles with: bare `class`/`enum` tokens in ERROR nodes,
struct `typedef` bodies with optional fields, and `@deprecated`-prefixed
declarations that block declaration recognition.

**Tested against 6,978 Haxe source files with zero parse errors.**
Produces 73,419 nodes and 88,084 edges.

**Dependency:**

Requires `tree-sitter-haxe` — not on PyPI, so install the patched fork
directly:
```bash
pip install git+https://github.com/masquepublishing/tree-sitter-haxe.git


mallyskies:
**Note on parse quality**

This PR depends on `tree-sitter-haxe` from PyPI (currently v0.0.1). That version
has two known grammar bugs that affect `.hx` parsing:

- **Strings swallowing inline comments** — `"str" // comment` causes the comment
  to be consumed into the string token, corrupting the rest of the expression.
- **Member expressions are right-associative** — `a.b.c` is parsed as `a.(b.c)`
  instead of `(a.b).c`, producing incorrect AST structure for all chained
  property access.

Both are fixed in this pending PR to the grammar repo:
https://github.com/vantreeseba/tree-sitter-haxe/pull/67

**Update:** to correct the above — there's actually no PyPI release of
`tree-sitter-haxe` at all ("v0.0.1" was this repo's own unpublished package
metadata, not a PyPI version). Rather than wait on the upstream merge, this
PR now depends directly on
[our patched fork](https://github.com/masquepublishing/tree-sitter-haxe)
(branch `fix/grammar-issues-52-53`), which includes both fixes today — see
the updated PR description's Dependency section. Once
vantreeseba/tree-sitter-haxe#67 merges and a real PyPI release exists,
this PR will switch to a normal `haxe` extra and the fork can be retired.

safishamsi:
Thanks @mallyskies — the Haxe extractor itself is well-built (follows the `LanguageConfig`/dispatch pattern, graceful `ImportError` guard, sensible inherits-vs-implements split). Two blockers before it can merge:

1. **Dependency doesn't resolve.** `tree-sitter-haxe` returns 404 on PyPI — `pip install "graphifyy[haxe]"` / `uv sync` can't find it. Worse, it's added to the **`all`** extra, so it breaks `uv sync`/CI for *everyone* (dependency resolution fails before tests can even collect), contradicting the `#1104` note above the `dm`/`terraform` extras. Please publish the grammar to PyPI first and **pin** it (`tree-sitter-haxe>=x.y`), and at minimum keep it out of the `all` group until then.
2. **Rebase** — the branch is currently CONFLICTING with v8.

Nice-to-have: add `imports` and `calls` edge assertions (the fixture has both but no test checks them). Happy to merge once the dependency is real + pinned and the branch rebases clean.

mallyskies:
@safishamsi Thanks for your help and patience. I believe I've addressed both concerns:

1. Dependency resolution — removed the haxe extra from pyproject.toml entirely (rather than pinning) since the tree-sitter-haxe project I forked still has no PyPI release — the upstream grammar-fix PR (vantreeseba/tree-sitter-haxe#67) is still open. extract_haxe() already lazy-imports with a graceful fallback, so this is packaging-only. Manual install documented in the README and CHANGELOG. PR description updated to match.
2. Rebase — done; branch is on current v8 tip, no conflicts.

NOTE:  test_haxe_finds_imports/test_haxe_finds_calls (with real edge-label assertions) were already in the second commit, so the edge-assertion ask should be covered, but let me know if I didn't do that the way you want.

mallyskies:
@safishamsi — thanks for the review!

This was my first pass at Haxe support, and the grammar choice turned out to be the weak part. The tree-sitter grammar repo this PR depends on parsed only about 50% of our 5,372-file Haxe codebase cleanly; I forked it and spent five rounds of fixes getting that to 75%, and it stayed hard going — the remaining failures were structural rather than a list of bugs.

I've since moved to tong/tree-sitter-haxe, which models the language much more closely. It starts at 73% on the same corpus without any changes from us, and our fork is at 91% with six minor fixes on top (which I'm offering back upstream to tong). The extractor rewritten against it also produces materially better edges: parameterized supertypes, class fields, and abstract underlying/cast types all resolve now and produced nothing before.

I want more mileage on our own codebase before proposing it again, and your dependency blocker still stands — the grammar isn't on PyPI yet, and I'd rather publish and pin it myself or have tong do it than reopen the same objection. So this gets closed rather than left stale, and the replacement comes as a fresh PR: against current v8 it's 7 files and +830 lines, in graphify/extractors/haxe.py per extractors/MIGRATION.md, rather than the 100 files and +86k this branch has drifted to.

NOTE: The imports/calls assertions you asked for are in my new tests.

PR #1311, opened 2026-06-14
fix: handle truncated JSON and unescaped quotes in community labels

matiasduartee:
This PR adds a regex fallback in \_parse_label_response\ so that if the LLM output is truncated (Unterminated string) or contains unescaped quotes inside community names, the parser can salvage the valid name pairs instead of discarding the entire batch of communities.


safishamsi:
Thanks for this. The same idea shipped in `v8`: `_parse_label_response` now falls back to a regex that salvages the complete `"<cid>": "<name>"` pairs when a reply fails a strict `json.loads` (e.g. truncated mid-object), instead of dropping the whole batch to placeholders. Since the truncated-JSON case your PR targets is now covered, I am closing this as already implemented. Appreciate you flagging it.

Issue #1077, opened 2026-05-29
Extraction noise: scope-blind identifier collision + over-eager markdown fragment nodes


heitrix:
Running `/graphify --update` on a TS monorepo (~10.6k nodes, 23.5k edges, 629 communities). Two recurring extractor issues surfaced when answering the auto-generated "Suggested Questions" — both inflate the god-node list and the surprising-connection list with false positives.

## 1. Local `const`/`let` identifiers form phantom god-nodes

A test file with `const set = new Set(DOMAIN_EVENT_TYPES)` produces a node labeled simply `set`. Across many test files, identical bare identifiers (`set`, `sorted`, `getAttr`, …) collide on the same node ID and accumulate edges to unrelated communities. Result: `set` reports betweenness 0.113 across ~23 communities even though it's a one-line local in a single file.

**Repro signal:** "Suggested Questions" surfaces questions like _"Why does `set` connect Community 57 to Community 128 …"_ — the node turns out to be a local var, not an architectural bridge.

**Fix:** namespace local `const`/`let`/`var` identifiers by file or function scope (e.g. `events_types_test::set`) so they can't collide across files. Module-level / exported symbols should keep bare IDs.

## 2. Markdown fragmentation creates thousands of dead-end nodes

The doc extractor emits one node per fenced code block (`code:bash (# Repo root)`) and one per bare section heading (`Stack`, `Design Context`), each with a single `contains` edge to its parent doc. They never link to anything else and inflate the "weakly-connected nodes" warning. In my graph, 305 connected components total — giant = 7,439; next = 100; rest are mostly markdown fragments.

**Repro signal:** "Suggested Questions" surfaces _"What connects `Design Context`, `Stack`, `code:bash (# Repo root (Turborepo))` to the rest of the system?"_ The answer is "nothing, by design — they're markdown fragments."

**Fix options (pick one):**
- Drop fenced code-block and bare-heading nodes entirely (low value).
- Or: cross-link headings to code symbols when the heading text matches an exported symbol name.

## 3. Minor: `covers` edges between audit docs and "GAP" cells get tagged AMBIGUOUS

Lower priority. When a doc explicitly says _"not covered — accessed via getDirectDb()"_, the `covers` edge from doc → table gets confidence AMBIGUOUS because the cell text reads as negation. It's an explicit architectural decision, not low confidence. A `relation: intentionally_excludes` (or `documents_exclusion`) variant would let the audit flow report "documented gaps" separately from "uncertain coverage."

[...]

PR #1323, opened 2026-06-15
feat: extract JS/TS this.X=, exports.X=, prototype, class arrow fields, and function expressions

### LINKED ISSUE #1322: JS/TS AST: function symbols assigned via this.X=, exports.X=, prototype, class arrow fields, and function expressions are never extracted
## Summary

The JS/TS extractor captures top-level `function` declarations, top-level `const x = () => …` arrows, classes, and class-method shorthand — but silently **drops** several extremely common ways of defining a named function. On expression-style and CommonJS codebases this means the majority of callable symbols never become nodes (and so can never be call-edge endpoints).

## Forms that are missed

| Form | Example | Today |
|------|---------|-------|
| constructor-assigned method | `function UserDAO(db){ this.addUser = () => {} }` | ❌ |
| `this.x = function(){}` | inside any function body | ❌ |
| CommonJS export | `exports.foo = () => {}` / `module.exports.bar = fn` | ❌ |
| prototype method | `Foo.prototype.baz = function(){}` | ❌ |
| class arrow field | `class C { onClick = (e) => {} }` | ❌ |
| `const f = function(){}` | function expression (non-arrow) | ❌ |

The first one is the big one: the constructor-function-with-`this.method = () => {}` pattern is pervasive in pre-class JS (DAOs, route handlers, services). For such a file the extractor emits only the outer `function` node and none of its methods.

## Minimal repro

```js
// probe.js
function declTop(a){ return a; }              // ✅ captured
const arrowTop = (b) => b;                     // ✅ captured
const fnExprTop = function(c){ return c; };    // ❌ missed
exports.exportedArrow = (d) => d;              // ❌ missed
module.exports.modArrow = (e) => e;            // ❌ missed
Foo.prototype.protoFn = function(){};          // ❌ missed
class Foo { methodInClass(f){ return f; } arrowField = (g) => g; }  // method ✅, field ❌
function Outer(db){ this.thisMethod = (h) => h; this.thisFnExpr = function(i){ return i; }; }  // ❌ both missed
```

```python
from graphify.extract import extract_js
from pathlib import Path
print(sorted(n["label"] for n in extract_js(Path("probe.js"))["nodes"]))
# ['.methodInClass()', 'Foo', 'Outer()', 'arrowTop()', 'declTop()', 'probe.js']
# missing: fnExprTop, exportedArrow, modArrow, protoFn, arrowField, thisMethod, thisFnExpr
```

## Root cause

- `_js_extra_walk` only emits arrow declarations from **module-level** `lexical_declaration`, and only for `arrow_function` values (not `function_expression`).
- `assignment_expression` (the `this.X =` / `exports.X =` / `Foo.prototype.X =` forms) is not handled at all.
- Class `field_definition` / `public_field_definition` with a function value is not emitted as a method.
- The function-emission branch appends the body to `function_bodies` for the call pass and returns, so `this.X = …` assignments *inside* the body are never visited by the node-emitting walk.

## Proposed fix

Extend the JS/TS walk to emit these forms, scoped to avoid the phantom-god-node class (#1077):
- `exports.X` / `module.exports.X` → file-contained function
- `Foo.prototype.X` → method owned by `Foo`
- `this.X = fn` in a function/constructor body → method owned by the enclosing class (if any) else the function
- class arrow/function fields → method owned by the class
- `const f = function(){}` → file-contained function

Arbitrary `obj.x = fn` (object not `this`/`exports`/`module.exports`/`<Name>.prototype`) is deliberately **not** captured, to preserve the #1077 guard.

PR with tests to follow.

siddarthc:
Closes #1322.

## Problem

The JS/TS extractor captured top-level `function` declarations, top-level `const x = () => …` arrows, classes, and method shorthand — but silently dropped several common ways of defining a named function. On expression-style and CommonJS codebases the majority of callable symbols never became nodes (and could not be call-edge endpoints):

```js
function UserDAO(db){ this.addUser = () => {} }   // constructor-assigned method — missed
this.handler = function(){}                        // this.X function expression — missed
exports.foo = () => {}                             // CommonJS export — missed
module.exports.bar = fn                             // missed
Foo.prototype.baz = function(){}                   // prototype method — missed
class C { onClick = (e) => {} }                    // class arrow field — missed
const f = function(){}                             // function expression — missed
```

The constructor-function-with-`this.method = () => {}` pattern is pervasive in pre-class JS (DAOs, route handlers, services); such a file emitted only the outer `function` node and none of its methods.

## Fix

- `_js_member_assignment_target()` classifies an assignment LHS into `this` / `exports` / `module.exports` / `<Name>.prototype`. Anything else (an arbitrary `obj.x = fn`) returns `None` and is skipped — this preserves the module-level scope guard from #1077 (no phantom god-nodes from bare-named locals).
- `_js_extra_walk` now also handles module-level `exports.X` / `module.exports.X` / `Foo.prototype.X` assignments and class function-valued fields, and accepts `function_expression` alongside `arrow_function` in `lexical_declaration`.
- The function-emission branch scans the function/constructor body for `this.X = fn` direct assignments and emits them as methods owned by the enclosing class (if present) else the function. Recovered function bodies are appended to `function_bodies` so the call pass sees them.

Ownership / labels follow the existing conventions: file-contained functions get `name()` + a `contains` edge; methods get `.name()` + a `method` edge from their owner.

## Tests

6 new tests in `tests/test_extract.py` — one per recovered form plus a negative guard asserting arbitrary `obj.x = fn` is **not** captured. The existing arrow / TSX / require regression tests still pass.

Full suite: no regressions. (The 18 failures in `test_terraform` / `test_skillgen` / `test_install_references` reproduce on a clean checkout of this branch's base — missing tree-sitter grammar and unbuilt skill payload in the dev environment, unrelated to this change.)

## Repro before/after

```python
from graphify.extract import extract_js
from pathlib import Path
print(sorted(n["label"] for n in extract_js(Path("probe.js"))["nodes"]))
# before: ['.methodInClass()', 'Foo', 'Outer()', 'arrowTop()', 'declTop()', 'probe.js']
# after:  [+ '.arrowField()', '.thisFnExpr()', '.thisMethod()', 'exportedArrow()', 'fnExprTop()', 'modArrow()', ...]
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@ Full release notes with details on each version: [GitHub Releases](https://githu
 
 ## Unreleased
 
+- Feat: JS/TS AST now extracts function symbols defined via `this.X = () => {}` / `this.X = function(){}` (constructor-assigned methods), `exports.X = fn` / `module.exports.X = fn`, `Foo.prototype.X = fn`, class arrow/function fields (`class C { onClick = (e) => {} }`), and `const f = function(){}` function expressions. Previously only top-level `function` declarations, top-level `const x = () =>` arrows, classes, and method shorthand were captured, so the majority of callable symbols in constructor-style and CommonJS codebases (DAOs, route handlers, services) never became nodes and could not be call-edge endpoints. Arbitrary `obj.x = fn` is deliberately not captured, preserving the #1077 phantom-god-node guard (#1322).
 - Fix: `graphify query`, `graphify explain`, and MCP `query_graph`/`get_node` now show the human-readable community name (e.g. "FlashAttention Paper") instead of a blank or numeric ID after running `cluster-only`. `to_json` now accepts `community_labels` and embeds `community_name` on each node; read paths fall back to the numeric `community` field for backward compatibility with old graphs (#1305).
 - Fix: `graphify-mcp` and `python -m graphify.serve` now accept `--graph <path>` as an alias for the positional argument, consistent with every other graphify subcommand. Previously `--graph` raised "unrecognized arguments" (#1304).
 
--- a/graphify/extract.py
+++ b/graphify/extract.py
@@ -1726,10 +1726,117 @@ def _require_imports_js(node, source: bytes, file_nid: str, stem: str, edges: li
     return found
 
 
+# Node types whose value is a callable, for the JS/TS assignment / class-field
+# / function-expression forms below. Older tree-sitter-javascript grammars
+# label a function expression `function`; current ones use `function_expression`.
+_JS_FUNCTION_VALUE_TYPES = frozenset({"arrow_function", "function_expression", "function"})
+
+
+def _js_member_assignment_target(left, source: bytes):
+    """Classify the symbol an `assignment_expression` LHS defines when its RHS
+    is a function. Returns (kind, owner_name, member_name) or None.
+
+      this.foo = fn            → ("this",      None,  "foo")
+      exports.foo = fn         → ("exports",   None,  "foo")
+      module.exports.foo = fn  → ("exports",   None,  "foo")
+      Foo.prototype.bar = fn   → ("prototype", "Foo", "bar")
+
+    Any other shape (an arbitrary `obj.x = fn`) returns None and is skipped —
+    capturing those would reintroduce the bare-named / phantom-god-node class
+    of bug the module-level scope guard (#1077) exists to prevent.
+    """
+    if left is None or left.type != "member_expression":
+        return None
+    prop = left.child_by_field_name("property")
+    if prop is None:
+        return None
+    member_name = _read_text(prop, source)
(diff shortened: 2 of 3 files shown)


siddarthc:
Thanks for the quick merge! 🙏

For context on the motivation: this closes a real gap for tools that build a code graph over expression-style JS and CommonJS — constructor-assigned methods (`this.x = () => {}`), `exports.x` / `module.exports.x`, prototype methods, and class arrow fields. On those codebases the previous extractor surfaced only the top-level declaration, so the bulk of callable symbols (and any call edges to them) were invisible. On a NodeGoat-style DAO file, for example, the change takes a class from 1 captured symbol to all of its methods.

Whenever you next cut a release to PyPI, that'd let downstream consumers pick this up — no rush, just flagging interest. Happy to follow up if any edge cases surface.

PR #1330, opened 2026-06-16
fix: resolve Swift module imports to shared module nodes (#1327)

### LINKED ISSUE #1327: Swift `imports` edges are silently dropped from the final graph
A report authored by Claude on missing 1st-party module-level imports in Swift AST parsing:

[GraphifyModuleParsingIssue-repro.zip](https://github.com/user-attachments/files/28974604/GraphifyModuleParsingIssue-repro.zip)

# Swift `imports` edges are silently dropped from the final graph

## Summary

On a clean, from-scratch build, **100% of Swift `import` statements produce zero `imports` edges** in
`graphify-out/graph.json`. The AST extractor *does* emit the edges, but it emits them pointing at a
bare module-name target node (e.g. `corekit`) that it never creates. `build_from_json` then drops
every edge whose target isn't a known node — treating first-party module imports as
"external/stdlib" — so all of them vanish.

This contradicts the documented behavior
(<https://graphify.net/tree-sitter-ast-extraction.html>: *"import edges are module-level… extracted
without LLM"*). Other Swift relations (`calls`, `implements`, `inherits`, `references`, `method`,
`contains`) extract correctly — the gap is specifically module-level import resolution.

## Environment

- graphify **0.8.36** (installed via `uv tool`, Python 3.11)
- tree-sitter-swift
- macOS (Darwin 25.3.0)
- Clean, **from-scratch** AST extraction (no `graphify update`, no prior curated state)

## Minimal reproduction

A self-contained anonymized sample is attached. It is an Xcode app target plus a local SwiftPM
package with four cross-importing library modules, each exporting a `public` type named after its
module (mirrors a real monorepo).

```
GraphifyModuleParsingIssue/            # Xcode app target (imports FeatureAKit, FeatureBKit)
  GraphifyModuleParsingIssue/ContentView.swift
  GraphifyModuleParsingIssue/GraphifyModuleParsingIssueApp.swift
  Packages/
    Package.swift                      # 4 library products/targets
    Sources/CoreKit/CoreKit.swift
    Sources/NetworkingKit/NetworkingKit.swift   # import CoreKit
    Sources/FeatureAKit/FeatureAKit.swift       # import CoreKit, NetworkingKit
    Sources/FeatureBKit/FeatureBKit.swift       # import CoreKit, FeatureAKit
```

There are **7 internal `import <Module>` statements** across these files (plus `import SwiftUI` and
`import PackageDescription`). A correct graph should contain ~7 corresponding `imports` edges.

### Steps

1. Build the graph from scratch (AST extraction; this is a code-only corpus).
2. Run `check.py` (included) against `graphify-out/graph.json`.

### Expected

~7 `imports` edges (e.g. `FeatureAKit -> CoreKit`, `FeatureBKit -> FeatureAKit`, etc.).

### Actual

```
type=module nodes: 0 -> []
module dirs on disk: 4 -> ['CoreKit', 'FeatureAKit', 'FeatureBKit', 'NetworkingKit']
imports: 0 | references: 2 | inherits: 1 | implements: 1 | calls: 4
import edges:
```

Zero `imports` edges. Zero `type=module` nodes. Every other relation type extracted fine.

## Root cause

The edges *are* produced by the extractor, then dropped at graph-build time.

**1. The Swift import handler emits an edge to a module-name node it never creates.**
`_import_swift` in `graphify/extract.py` (0.8.36):

```python
def _import_swift(node, source, file_nid, stem, edges, str_path):
    for child in node.children:
        if child.type == "identifier":
            raw = _read_text(child, source)
            tgt_nid = _make_id(raw)        # e.g. "CoreKit" -> "corekit"
            edges.append({
                "source": file_nid,
                "target": tgt_nid,         # <-- no node with this id is ever appended
                "relation": "imports",
                ...
            })
            break
```

It appends an `imports` edge but never appends a corresponding node for the imported module.
(Contrast the JS/TS handlers, which resolve imports to real, existing file-path nodes.) So the
target id (`corekit`, `featureakit`, `swiftui`, …) is always dangling.

**2. `build_from_json` drops every dangling-target edge as "external/stdlib".**
`graphify/build.py`:

```python
if src not in node_set or tgt not in node_set:
    continue  # skip edges to external/stdlib nodes - expected, not an error
```

Because *every* Swift import target is dangling, this prunes 100% of `imports` edges — including
first-party, in-repo module imports.

### Evidence that the edges exist before the drop

Raw `extract()` output for this corpus contains **10 `imports` edges** (all 7 internal + `SwiftUI`
×2 + `PackageDescription`), each with a target id that does not exist as a node:

```
featureakit_featureakit -> corekit        (target exists as node: False)
featureakit_featureakit -> networkingkit  (target exists as node: False)
featurebkit_featurebkit -> corekit        (target exists as node: False)
featurebkit_featurebkit -> featureakit    (target exists as node: False)
networkingkit_networkingkit -> corekit    (target exists as node: False)
...
```

After `build_from_json` → `to_json`, the final graph has **0** `imports` edges.

### This is not monorepo-specific

Running extraction on **just the `Packages/` SwiftPM package** (graphify's documented happy path,
no app target) shows the same thing: raw `extract()` yields 6 `imports` edges, but after
`build_from_json` the final graph has **0**. The defect applies to any Swift corpus.

## Suggested fix

When the Swift import handler emits an `imports` edge, also emit a node for the imported module
(e.g. `type=module`, label = module name, id = `_make_id(module_name)`), the way other languages
anchor imports to a resolvable node. Optionally, parse `Package.swift` targets/products into
`type=module` nodes so first-party imports resolve to them while genuinely external modules
(`SwiftUI`, `Foundation`) can still be created-or-skipped deliberately rather than dropped by the
generic "dangling edge" pruning in `build_from_json`.

## Attached

- The full sample project (app + SwiftPM package).
- `check.py` — the assertion script that produced the output above.
- `graphify-out/graph.json` and `graphify-out/GRAPH_REPORT.md` from the from-scratch build.

duncan-daydream:
Fixes #1327.

## Problem

On a clean, from-scratch AST build, **100% of Swift `import` statements produced zero `imports` edges** in `graph.json`. `_import_swift` emitted an `imports` edge to a bare module-name id (e.g. `corekit`) that no node ever backed, so `build_from_json` pruned every Swift import edge as a dangling/external reference — including first-party, in-repo module imports. Zero `type=module` nodes were produced.

## Fix

Two coordinated changes in `graphify/extract.py`:

1. **Anchor imports to module nodes.** `_import_swift` now returns the imported modules as `(id, label)` pairs, and the generic extractor materializes a `type=module` node for each so the edge resolves. `add_node` gains an optional `node_type` kwarg to tag them.

2. **Don't split shared module nodes.** `_disambiguate_colliding_node_ids` exempts `type=module` nodes. The same module imported from N files yields N module-node dicts sharing one id; without the exemption the collision-splitter scattered them into N path-qualified duplicates instead of collapsing into a single shared node. Keeping one id lets `build_from_json` merge them, so reverse traversal ("what imports CoreKit?") works.

Other languages resolve imports to existing file-path nodes and are unaffected — their handlers return `None`, and their symbol nodes carry no `type=module` marker.

## Verification

On the reporter's repro project, via `graphify <path>` + the issue's `check.py`:

| | Before (0.8.36) | After |
|---|---|---|
| `type=module` nodes | 0 | 6 (4 first-party + SwiftUI, PackageDescription), deduplicated |
| `imports` edges | 0 | 10 (all 7 first-party + SwiftUI×2 + PackageDescription) |

All 7 expected first-party edges now resolve (`FeatureAKit → CoreKit`, `FeatureBKit → FeatureAKit`, `NetworkingKit → CoreKit`, etc.).

## Tests

- New `tests/test_swift_import_resolution.py`: import targets resolve to nodes, same-module-imported-from-multiple-files collapses to one shared node, and edges survive `build_from_json`.
- New regression cases in `tests/test_languages.py`.
- Full suite: **2035 passed, 28 skipped**.

### DIFF
--- a/graphify/extract.py
+++ b/graphify/extract.py
@@ -2107,7 +2107,16 @@ _LUA_CONFIG = LanguageConfig(
 )
 
 
-def _import_swift(node, source: bytes, file_nid: str, stem: str, edges: list, str_path: str) -> None:
+def _import_swift(node, source: bytes, file_nid: str, stem: str, edges: list, str_path: str) -> list[tuple[str, str]]:
+    """Emit module-level `imports` edges and report the imported modules.
+
+    A Swift `import CoreKit` names a module, not a file path, so — unlike the
+    file-resolving JS/TS handlers — there is no existing node for the edge to
+    point at. The returned (id, label) pairs let the extractor materialize a
+    `type=module` anchor node so the edge survives; without it build_from_json
+    prunes every Swift import edge as a dangling/external reference (#1327).
+    """
+    modules: list[tuple[str, str]] = []
     for child in node.children:
         if child.type == "identifier":
             raw = _read_text(child, source)
@@ -2122,7 +2131,9 @@ def _import_swift(node, source: bytes, file_nid: str, stem: str, edges: list, st
                 "source_location": f"L{node.start_point[0] + 1}",
                 "weight": 1.0,
             })
+            modules.append((tgt_nid, raw))
             break
+    return modules
 
 
 def _read_csharp_type_name(node, source: bytes) -> str | None:
@@ -2219,16 +2230,19 @@ def _extract_generic(path: Path, config: LanguageConfig) -> dict:
     if config.ts_module == "tree_sitter_swift":
         swift_protocol_names, swift_class_names = _swift_pre_scan(root, source)
 
-    def add_node(nid: str, label: str, line: int) -> None:
+    def add_node(nid: str, label: str, line: int, *, node_type: str | None = None) -> None:
         if nid not in seen_ids:
             seen_ids.add(nid)
-            nodes.append({
+            node = {
                 "id": nid,
                 "label": label,
                 "file_type": "code",
                 "source_file": str_path,
                 "source_location": f"L{line}",
-            })
+            }
+            if node_type:
+                node["type"] = node_type
+            nodes.append(node)
 
     def add_edge(src: str, tgt: str, relation: str, line: int,
                  confidence: str = "EXTRACTED", weight: float = 1.0,
@@ -2264,7 +2278,16 @@ def _extract_generic(path: Path, config: LanguageConfig) -> dict:
         # Import types
         if t in config.import_types:
             if config.import_handler:
-                config.import_handler(node, source, file_nid, stem, edges, str_path)
+                imported_modules = config.import_handler(node, source, file_nid, stem, edges, str_path)
+                # Module-level import handlers (Swift) name a module, not a file
+                # path, so there is no pre-existing node to anchor the edge to.
+                # They return (id, label) pairs for which we materialize a
+                # `type=module` node; otherwise build_from_json prunes every such
+                # import edge as a dangling/external reference (#1327).
+                if imported_modules:
+                    line = node.start_point[0] + 1
+                    for mod_nid, mod_label in imported_modules:
+                        add_node(mod_nid, mod_label, line, node_type="module")
             # For export_statement: only return (skip children) if it's a re-export
             # (has a `from` source). Otherwise fall through to walk children which may
             # contain function_declaration, class_declaration, etc.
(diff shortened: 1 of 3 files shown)


safishamsi:
Superseded: this fix's approach (synthesize a shared `type=module` node per Swift import + exempt module nodes from id-disambiguation so the same module collapses to one node) was adopted and shipped to v8 in commit 4fbb333, with your tests, crediting you in the commit and CHANGELOG. Closing as the change is merged. Thanks @duncan-daydream for the fix and the clear analysis.

Issue #1318, opened 2026-06-15
AST extraction generates inconsistent node IDs for same symbol at definition   vs. reference sites, causing edges to connect to orphan "shadow nodes" instead   of actual definitions


snoopyshi:
When extracting Java code, graphify's AST extractor generates node IDs using
  the {stem}_{entity} format where stem is derived from the immediate parent
  directory + filename without extension. This causes two problems:

  Problem 1 — ID collision between definition and reference sites:
  The same symbol AIResponseHandler gets two different IDs:
  - At definition (com/xx/ai/handler/AIResponseHandler.java):
  handler_airesponsehandler_airesponsehandler
  - At reference (com/xx/ai/service/DifyAiServiceImpl.java): airesponsehandler
  or
  xx_src_main_java_com_xx_ai_service_difyaiserviceimpl_java_airesponsehandler

  Problem 2 — Implements edge connects to wrong node:
  The implements edge is created correctly:
  handler_airesponseconsumerhandler_airesponseconsumerhandler implements
  airesponsehandler
  But airesponsehandler (community 191) is a reference-only shadow node with no
  outgoing edges to the actual interface definition. The real interface
  definition handler_airesponsehandler_airesponsehandler (community 1586) has
  zero external edges and is wrongly classified as an isolated community.

safishamsi:
Progress: cross-file type-reference resolution now covers **C#** too (PR #1466, on `v8`) — the C# resolver re-points dangling inherits/implements/references edges from shadow stubs to their real definitions, disambiguating same-named types across namespaces via `using` directives (Java was already covered). Go phantom duplicates were also fixed (#1500/#1501) and same-label imported stubs no longer falsely merge (#1462/#1479). Keeping this open as the cross-language umbrella — remaining: C# types in nested/multiple namespaces per file, and other languages without a dedicated resolver.

safishamsi:
Progress: **0.9.0** ships full repo-relative node IDs, which removes the same-relative-path file collisions and aligns the AST/LLM id rules (#1504/#1509/#1158). Combined with the C# and Go cross-file resolvers landed earlier, a lot of the def-vs-reference inconsistency is now resolved. Keeping this open as the umbrella for the remaining cross-file *type-reference* resolution work (languages without a dedicated resolver, nested-namespace cases).

kbrilla:
[...]

PR #1332, opened 2026-06-16
fix(extract): skip Python primitive/library type-annotation nodes

MasterKinjalk:
## Problem
Type annotations (`return_type`, `parameter_type`, `field`, `generic_arg`, base class) materialize Python language primitives and common library types — `str`, `int`, `Tensor`, `ndarray`, `Path`, `Module`, `np`, `torch` — as graph nodes. They carry no domain meaning and become spurious high-degree hubs that pollute community clustering and clutter the report's node list.

## Cause
`_LANGUAGE_BUILTIN_GLOBALS` already filtered these names at **call** sites, but not at **type-reference** sites, which route through `ensure_named_node`.

## Fix (scoped to Python — see review thread)
- Add `_PYTHON_TYPE_REF_PRIMITIVES` = `_LANGUAGE_BUILTIN_GLOBALS` ∪ {stdlib / typing / NumPy / PyTorch annotation types}, gated per language via `_TYPE_REF_PRIMITIVES_BY_LANG` / `_type_ref_primitives(ts_module)`.
- `_extract_generic` skips `add_node` for these names **only when the language is Python**. The node id is still returned, so reference edges still resolve (dangling-to-external, tolerated by `build.py`).
- The hand-written Julia / Fortran / Rust / PowerShell / Obj-C walkers no longer carry the guard — it only over-filtered foreign-named user types there.

## Why Python-only
The same bare names are ordinary user-defined types in other languages (a Rust `Response`, a Julia `Module`, a Go `Set`). A single flat membership test applied in every extractor silently dropped them. Languages that need to suppress their own primitives already do so at the collector, scoped to the language — Go via `_GO_PREDECLARED_TYPES`, Rust by skipping `primitive_type` nodes, Python scalars via `_PYTHON_ANNOTATION_NOISE` (#1147). This PR layers the library/annotation names on top, for Python only.

## Tests
`tests/test_type_ref_primitives.py` — per affected extractor (Python / Rust / Julia / Go): primitive dropped, a real same-named user type kept, edge still resolves, plus the cross-language over-filter regression.

## Verification
Extracting a sample with `Tensor` / `ndarray` / `int` annotations:
- **Before:** `Tensor`, `ndarray`, `int`, `np`, `torch` appear as nodes.
- **After:** only real nodes remain (`Net`, `.forward()`, `.to_np()`, `count()`); no primitive leakage.

Rebased onto `v8` (`9e0b876`). Full suite: no new failures vs the v8 base.

### DIFF
--- a/graphify/extract.py
+++ b/graphify/extract.py
@@ -43,6 +43,47 @@ _LANGUAGE_BUILTIN_GLOBALS: frozenset[str] = frozenset({
 })
 
 
+# Python primitive / library type names that leak in as nodes purely via type
+# annotations (return_type, parameter_type, field, generic_arg, base class)
+# rather than as real domain concepts. Skipping node creation for these keeps the
+# graph clean; the leftover reference edges are the dangling-to-external kind
+# build.py already tolerates. Reuses _LANGUAGE_BUILTIN_GLOBALS (the shared
+# call-site builtin filter) and layers stdlib / NumPy / PyTorch annotation types.
+#
+# Deliberately PYTHON-ONLY: these bare names are ordinary user-defined types in
+# other languages (a Rust `Response`, a Julia `Module`, a Go `Set`), so applying
+# one flat list to every extractor mis-drops real types (review #1332). Languages
+# that suppress their own primitives already do so at the collector, scoped to the
+# language — Go via _GO_PREDECLARED_TYPES, Rust by skipping `primitive_type`
+# nodes, Python scalars via _PYTHON_ANNOTATION_NOISE (#1147) — so this set only
+# layers the library/annotation type names on top, for Python.
+_PYTHON_TYPE_REF_PRIMITIVES: frozenset[str] = _LANGUAGE_BUILTIN_GLOBALS | frozenset({
+    # Python typing / stdlib annotation types
+    "Any", "Optional", "Union", "List", "Dict", "Set", "Tuple", "Sequence",
+    "Iterable", "Iterator", "Mapping", "Callable", "Type", "Literal", "Final",
+    "Annotated", "Path", "PathLike", "datetime", "timedelta", "Decimal",
+    "frozenset", "bytearray", "complex", "object", "None", "NoneType",
+    # NumPy
+    "ndarray", "np", "dtype", "NDArray", "ArrayLike",
+    # PyTorch
+    "Tensor", "Module", "torch", "device", "DataLoader", "Dataset",
+    "nn", "Parameter",
+})
+
+
+# Type-reference primitive filter keyed by tree-sitter module. Only languages with
+# an entry suppress annotation-only type names in ensure_named_node; every other
+# extractor keeps them (or filters its own predeclared types at the collector).
+_TYPE_REF_PRIMITIVES_BY_LANG: dict[str, frozenset[str]] = {
+    "tree_sitter_python": _PYTHON_TYPE_REF_PRIMITIVES,
+}
+
+
+def _type_ref_primitives(ts_module: str) -> frozenset[str]:
+    """Primitive/library type-ref names to skip as nodes for the given language."""
+    return _TYPE_REF_PRIMITIVES_BY_LANG.get(ts_module, frozenset())
+
+
 def _raise_recursion_limit() -> None:
     if sys.getrecursionlimit() < _RECURSION_LIMIT:
         sys.setrecursionlimit(_RECURSION_LIMIT)
@@ -2459,12 +2500,19 @@ def _extract_generic(path: Path, config: LanguageConfig) -> dict:
             edge["context"] = context
         edges.append(edge)
 
+    # Annotation-only primitive/library type names to skip as nodes, scoped to
+    # this language (empty for everything but Python — see _type_ref_primitives).
+    type_ref_primitives = _type_ref_primitives(config.ts_module)
+
     def ensure_named_node(name: str, line: int) -> str:
         nid = _make_id(stem, name)
         if nid in seen_ids:
             return nid
         nid = _make_id(name)
-        if nid not in seen_ids:
+        # Skip node creation for primitive / library type names — they pollute the
+        # graph as noise hubs. Return the id so reference edges still resolve
+        # (dangling-to-external, tolerated by build.py); just don't materialize the node.
+        if nid not in seen_ids and name not in type_ref_primitives:
             add_node(nid, name, line)
         return nid
 
--- a/tests/test_type_ref_primitives.py
+++ b/tests/test_type_ref_primitives.py
@@ -1,0 +1,156 @@
+"""Regression tests for type-reference primitive filtering (PR #1332 review).
+
+Behaviour contract:
+  * Python primitive / library type names that appear only in annotations
+    (return_type / parameter_type / field / generic_arg / base) are NOT
+    materialized as graph nodes — they are noise hubs, not domain concepts.
+  * A real, user-defined type whose NAME collides with a primitive/library type
+    in some *other* language must still be materialized, and its edges resolve.
+  * The filter is scoped per language: a name that is a primitive in Python
+    (`Tensor`, `Module`) is a legitimate user-defined type name in Rust/Julia/Go,
+    so a single flat global list applied to every extractor mis-drops real types
+    (the over-filtering #1332 review flagged). Languages that already filter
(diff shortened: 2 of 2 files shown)


safishamsi:
Thanks @MasterKinjalk — the idea (skip primitive/library type names so they don't become noise nodes) is sound, but a few things need addressing before this can merge:

1. **Inconsistent coverage.** The guard is added to 6 of the 7 `ensure_named_node` definitions but is **missing in the Go extractor** (`extract_go`), so Go type references will still materialize `int`/`string` etc. as nodes. Add it there or document why Go is excluded — the PR description says 'all language extractors', which isn't currently true.
2. **False-positive over-filtering.** The membership test is on the bare type name with no namespace/origin check, so a user-defined type named `Path`/`Set`/`Type`/`Module`/`Dataset`/`Parameter`/`device`/`File`/`object` is silently dropped as a node while its inheritance/reference edge still points at it — recreating the orphan/shadow-node problem #1318 just fixed. Please gate by language and/or trim names that plausibly collide with user-defined types.
3. **Scope.** The filter only touches the hand-written walkers, not `_extract_generic`, so the high-traffic typed languages (Java/Swift/TS/JS) aren't covered. Either extend it to `_extract_generic` or narrow the description to match.
4. **Tests.** This is exactly the kind of change that silently drops nodes — please add a regression test per affected extractor (primitive dropped, a real same-named user type kept, edge still resolves).

Also please rebase onto current `v8` (the branch predates #1318/#1327). Happy to re-review once these are addressed.

MasterKinjalk:
Thanks for the detailed review. Rebased onto current `v8` and reworked the approach after tracing how each extractor actually handles type references — addressed point by point below.

**Scope reworked to Python-only.** The root issue you flagged — a name-only membership test with no language/origin check — drove the redesign. The filter is no longer one flat set applied in every walker; it's keyed by language (`_TYPE_REF_PRIMITIVES_BY_LANG`) and populated for Python only.

**1. Go / consistency.** Go is intentionally excluded, with a reason rather than a guard: `_go_collect_type_refs` already filters Go's predeclared types via `_GO_PREDECLARED_TYPES`, so Go never materializes `int`/`string`/`error` as nodes. Adding the flat guard to Go would only over-filter user types (e.g. a Go type named `Set`). Same pattern already exists for Rust (`primitive_type` nodes are skipped) and Python scalars (`_PYTHON_ANNOTATION_NOISE`, #1147) — predeclared filtering lives at the per-language collector, which is exactly where Go does it.

**2. Over-filtering / shadow nodes.** Two findings:
- A user type **defined in the same file** is never dropped — `ensure_named_node` returns the stem-qualified id for in-file symbols and only gates the bare cross-file/external id. So the orphan/shadow-node problem #1318 fixed does not recur for same-file defs (regression test included).
- The genuine over-filtering was **cross-language**: the flat set carried JS globals (`Response`, `Request`, `File`) and PyTorch names (`Module`, `Tensor`), which mis-dropped identically-named Rust/Julia/etc. user types. Scoping to Python removes that. `test_rust_type_named_like_js_global_not_overfiltered` covers it (red before, green after).

**3. Scope / typed languages.** `_extract_generic` is now the only place the guard lives, so Python (routed through it) is covered. Java/Swift/TS/JS and the other typed languages **keep** their primitive type nodes on purpose — their extractor tests assert those refs resolve as nodes (PowerShell `string`/`void`, Julia `Float64`, C++ `string`/`vector`, Obj-C `NSString`). Narrowed the PR description to "Python" to match.

**4. Tests.** Added `tests/test_type_ref_primitives.py`: per extractor (Python / Rust / Julia / Go) — primitive dropped, real same-named user type kept, edge still resolves, plus the cross-language over-filter regression.

**5. Rebase.** Done — rebased onto `v8` (`9e0b876`), clean. #1318/#1327 are already in the base.

Full suite shows no new failures vs the v8 base. Happy to adjust scope if you'd prefer the filter extended to other languages' own primitives (that would mean updating those extractors' existing assertions, so I left them as-is).

Synvoya (commented):
Thanks for tackling this — the problem is real and correctly diagnosed. Library/annotation type names (`Tensor`, `ndarray`, `np`, `torch`, `Module`, `Path`, `datetime`) leaking into the graph as high-degree noise hubs is worth fixing, and the per-language scoping rationale in the description is exactly the right instinct (a flat global list *would* mis-drop a Rust `Response` or a Go `Set`). Two things before it can land, though — the patch as written can't apply to `v8`, and its mechanism collides with a fix that went in after this branch was cut.

### 1. The patch no longer applies to `v8`

`git apply --check` fails at `extract.py:43`. This was written against a pre-refactor `extract.py`: the hunk anchors on an `ensure_named_node` whose body is

```python
nid = _make_id(name)
if nid not in seen_ids:
    add_node(nid, name, line)
```

but on current `v8` (`extract.py:3504`) that function is now `nid = _make_id(stem, ".".join(namespace_stack), name)` and its body is a **sourceless-stub append**, not `add_node`. Needs a rebase before it can be assessed for merge.

### 2. Suppressing the node here would regress the #1402 phantom-duplicate fix

This is the blocking concern. On `v8`, `ensure_named_node` deliberately materializes a *sourceless* stub for cross-file type references (`extract.py:3510-3525`):

```python
# The name isn't defined in this file, so this is a cross-file reference ...
# Emit a SOURCELESS stub ... so the corpus-level rewire can collapse it onto
# the real definition. A sourced stub here ... blocks the rewire, which is
# the phantom-duplicate-node bug (#1402).
```

The PR's change is "don't create the node at all, just return the id." For an imported **real** type (`from .models import Dataset`, used only in an annotation), that removes the very stub the corpus rewire keys on — so instead of collapsing onto the real `Dataset` definition, the reference edge now points at an id with no node and no rewire target. That reintroduces #1402 for real imported types. The same sourceless-stub pattern is load-bearing at seven sites (`3516, 7690, 8007, 8270, 8625, 8989, 12274`), so a suppression at `ensure_named_node` fights all of them.

### 3. The scalar names in `_PYTHON_TYPE_REF_PRIMITIVES` are already filtered upstream

Python parameter/return/generic annotations don't reach `ensure_named_node` raw — they go through `_python_collect_type_refs` (`extract.py:651`), which already drops `_PYTHON_ANNOTATION_NOISE` (`str, int, float, bool, bytes, bytearray, complex, object, True, False`, at `641`) and `_PYTHON_TYPE_CONTAINERS` *before* the emit sites at `4384-4401`. So `int`, `str`, `bytearray`, `complex`, `object` in the new set can never fire — they're filtered a layer earlier. The genuinely-new coverage is only the library names (`Tensor`, `ndarray`, `np`, `torch`, `Module`, `Dataset`, `NDArray`, `Path`, `datetime`, …).

**Suggested direction:** drop the `ensure_named_node` change and instead extend the existing collector filter — add the library/annotation names to `_PYTHON_ANNOTATION_NOISE` (or a sibling `_PYTHON_LIBRARY_ANNOTATION_NOISE` frozenset unioned into the same guards at `651-698`). That filters the noise at the same point Python scalars are already filtered, keeps the fix Python-scoped without the `_TYPE_REF_PRIMITIVES_BY_LANG` indirection, and never suppresses a stub the #1402 rewire depends on — a name dropped at the collector never creates an edge in the first place, so there's no dangling target.

One caveat on that path: a name like `Path` or `Dataset` can legitimately be a user's own class. Filtering it at the collector means a real project-internal `class Dataset` referenced in an annotation would also lose that annotation edge. `_PYTHON_ANNOTATION_NOISE` already accepts that trade for scalars; worth a sentence on why it's acceptable for these library names too (they're overwhelmingly third-party), or gating on whether the name resolves to a local definition.

### 4. Tests assert behavior this diff doesn't change

`tests/test_type_ref_primitives.py` asserts Rust/Julia/Go outcomes, but the diff only adds a **Python** key to `_TYPE_REF_PRIMITIVES_BY_LANG` — every other language resolves to `frozenset()` and is untouched, so those cases test pre-existing collector behavior, not this change. The one test that exercises the new path (`test_python_primitive_type_refs_not_materialized`) is the one to keep; consider folding it into `tests/test_languages.py` to match the per-language convention the rest of the suite (and the merge-onto-`v8` flow) uses.

Net: correct problem, right scoping instinct, but (2) is a real regression risk and (1) blocks the merge. Moving the filter to the collector resolves both and shrinks the diff. Happy to re-review after a rebase.

PR #1333, opened 2026-06-16
fix: resolve package subpath exports

### LINKED ISSUE #1308: TypeScript workspace imports do not resolve package subpath exports
## Summary

Graphify's TypeScript/JavaScript workspace import resolver does not resolve package subpath imports through `package.json` `exports`.

For a workspace package import like:

```ts
import { value } from "@example/package-a/browser";
```

Graphify records an `imports_from` edge to a bare `browser` node instead of resolving the target to the exported source file from `packages/package-a/package.json`.

This breaks affected-file traversal and graph accuracy for monorepos that expose browser/server or other subpath entrypoints through package exports.

## Minimal repro

```text
repo/
  package.json
  packages/
    package-a/
      package.json
      src/browser.ts
  apps/
    package-b/
      src/consumer.ts
```

Root `package.json`:

```json
{
  "private": true,
  "workspaces": ["packages/*", "apps/*"]
}
```

`packages/package-a/package.json`:

```json
{
  "name": "@example/package-a",
  "exports": {
    "./browser": {
      "source": "./src/browser.ts",
      "import": "./dist/esm/browser.js",
      "require": "./dist/cjs/browser.js",
      "types": "./dist/types/browser.d.ts"
    }
  }
}
```

`packages/package-a/src/browser.ts`:

```ts
export const value = "ok";
```

`apps/package-b/src/consumer.ts`:

```ts
import { value } from "@example/package-a/browser";

export const consumerValue = value;
```

## Actual behavior

Graphify emits an import edge similar to:

```json
{
  "source": "src_consumer",
  "target": "browser",
  "relation": "imports_from",
  "source_file": "apps/package-b/src/consumer.ts"
}
```

## Expected behavior

Graphify should resolve the package subpath through `package.json` `exports` and emit the import edge to the exported file node, for example:

```json
{
  "source": "src_consumer",
  "target": "src_browser",
  "relation": "imports_from",
  "source_file": "apps/package-b/src/consumer.ts"
}
```

The exact node id may differ, but the target should be the workspace source file behind the export map, not the bare subpath segment.

## Why this matters

Many TypeScript monorepos use package subpath exports for public entrypoints, especially to separate browser-safe and server-only APIs:

```json
{
  "exports": {
    "./browser": { "source": "./src/browser.ts", "import": "./dist/browser.js" },
    "./server": { "source": "./src/server.ts", "import": "./dist/server.js" }
  }
}
```

When Graphify records these as `browser` or `server`, reverse traversal and affected-file analysis become either incomplete or noisy.

## Suspected cause

In the workspace package resolver, the subpath appears to be checked as a filesystem path before package export maps are considered.

Conceptually, resolving `@example/package-a/browser` does something equivalent to:

```python
package_dir / "browser"
```

When that path does not resolve to a concrete file, the resolver falls back to the bare subpath string rather than consulting:

```json
exports["./browser"]
```

## Suggested fix

For workspace package imports with a subpath:

1. Locate the package root from the workspace package name.
2. Read `package.json`.
3. Look up the subpath in `exports` using the `./subpath` key.
4. Choose the best source target when available.
5. Fall back to runtime/type targets if no source target exists.
6. Only then fall back to direct filesystem probing such as `package_dir / subpath`.

Suggested condition preference for source-oriented graph extraction:

```text
source -> import -> module -> default -> require -> types
```

For condition objects, Graphify could recursively walk nested condition maps and return the first target matching the preferred condition list.

For wildcard exports such as:

```json
{
  "exports": {
    "./utils/*": {
      "source": "./src/utils/*.ts",
      "import": "./dist/utils/*.js"
    }
  }
}
```

replace the `*` segment from the import subpath before resolving the target.

This would make Graphify match modern Node/TypeScript monorepo package boundaries while preserving the existing direct-file fallback for packages that do not define exports.

pkudinov:
## Summary

- resolve workspace package subpath imports through package.json `exports`
- support exact subpath exports such as `./browser`
- support wildcard subpath exports such as `./server/queries/*`
- prefer source-oriented conditions before runtime/type-only targets

Fixes #1308.

## Verification

- `uv run pytest tests/test_js_import_resolution.py`

### DIFF
--- a/graphify/extract.py
+++ b/graphify/extract.py
@@ -383,6 +383,61 @@ def _load_workspace_packages(start_dir: Path) -> dict[str, Path]:
     return packages
 
 
+_PACKAGE_EXPORT_CONDITION_PRIORITY = ("source", "import", "module", "default", "require", "types", "svelte")
+
+
+def _package_export_target_values(value: Any) -> list[str]:
+    if isinstance(value, str):
+        return [value]
+    if isinstance(value, list):
+        targets: list[str] = []
+        for item in value:
+            targets.extend(_package_export_target_values(item))
+        return targets
+    if not isinstance(value, dict):
+        return []
+
+    targets: list[str] = []
+    for key in _PACKAGE_EXPORT_CONDITION_PRIORITY:
+        targets.extend(_package_export_target_values(value.get(key)))
+    for key, child in value.items():
+        if key not in _PACKAGE_EXPORT_CONDITION_PRIORITY:
+            targets.extend(_package_export_target_values(child))
+    return targets
+
+
+def _package_exports_candidates(exports: Any, subpath: str) -> list[Path]:
+    if isinstance(exports, str):
+        return [Path(exports)] if not subpath else []
+    if not isinstance(exports, dict):
+        return []
+
+    requested = "." if not subpath else f"./{subpath}"
+    candidates: list[Path] = []
+
+    exact_export = exports.get(requested)
+    for target in _package_export_target_values(exact_export):
+        candidates.append(Path(target))
+
+    if subpath:
+        for export_key, export_value in exports.items():
+            if "*" not in export_key:
+                continue
+            prefix, suffix = export_key.split("*", 1)
+            if not requested.startswith(prefix) or (suffix and not requested.endswith(suffix)):
+                continue
+            replacement_end = len(requested) - len(suffix) if suffix else len(requested)
+            replacement = requested[len(prefix):replacement_end]
+            for target in _package_export_target_values(export_value):
+                candidates.append(Path(target.replace("*", replacement)))
+
+    if not subpath and not candidates:
+        for target in _package_export_target_values(exports):
+            candidates.append(Path(target))
+
+    return candidates
+
+
 def _package_entry_candidates(package_dir: Path, subpath: str) -> list[Path]:
     manifest = package_dir / "package.json"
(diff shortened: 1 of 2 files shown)


safishamsi:
Thanks @pkudinov — the `extract.py` change is correct and well-scoped: it handles exact subpath exports, conditional exports (`source`/`import`/`require`/`types`, source-first), and wildcard subpaths, and it's additive (export candidates are tried first, the `package_dir / subpath` fallback is preserved), so no regression risk to existing resolution.

The only blocker is a **rebase**: #1352 (npm/yarn workspaces) landed first, and both PRs added tests at the same anchor in `tests/test_js_import_resolution.py`, so this is now CONFLICTING. The conflict is **only in the test file** — `extract.py` auto-merges cleanly. Please rebase onto current `v8` and keep both sets of tests (trivial concatenation); then CI can run and I'll merge.

Optional follow-up nit: also handle array-form export targets (`"./x": ["./a.ts", "./b.ts"]`), which currently return no candidates.

pkudinov:
Rebased this PR onto current `v8` and resolved the test-file conflict by keeping both the npm/yarn workspace coverage and the pnpm package-subpath export coverage.

I also included the optional array-form export target handling for entries like `"./x": ["./a.ts", "./b.ts"]`.

Verification: `uv run pytest -q` passed with `2098 passed, 28 skipped`.

PR #1351, opened 2026-06-17
fix: NFC-normalize labels in resolve_seed for accented/CJK lookups (#1338)

### LINKED ISSUE #1338: Node lookup fails for accented labels on macOS (NFD vs NFC) and for full labels with punctuation
Two related node-resolution problems hit non-English corpora (Spanish, accents/`ñ`), especially on
macOS where the filesystem stores filenames in Unicode **NFD** while `graph.json` labels are **NFC**.

Bug 1 — `affected.resolve_seed()` doesn't normalize Unicode.
It compares with `.lower()` only. A query in NFD never equals an NFC label:

```python
# repro
import unicodedata, networkx as nx
from graphify.affected import resolve_seed
G = nx.Graph(); G.add_node("n1", label="Auditoría")          # NFC, as stored in graph.json
resolve_seed(G, unicodedata.normalize("NFC", "Auditoría"))   # -> "n1"  OK
resolve_seed(G, unicodedata.normalize("NFD", "Auditoría"))   # -> None  ✗  (same text, NFD)
```
On macOS, a label argument copied from a filename arrives as NFD → `affected` reports "not found".

Bug 2 — `serve._find_node()` doesn't resolve a full label that contains punctuation.
It is diacritic-insensitive per word (good), but the search `term` is tokenized (`_search_tokens`,
strips `—`, `/`, `()`…) while it's compared against `norm_label`, which keeps punctuation. So a user
who passes the *exact* node label can't resolve it:

```python
# label = "Skill /auditar — Auditoría inquisitiva de enlaces"
_find_node(G, "Auditoría")                                   # -> matches (substring)  OK
_find_node(G, "Skill /auditar — Auditoría inquisitiva de enlaces")  # -> []  ✗
```
This affects `explain`/`query`/`path` and the MCP serve layer.

Impact: any non-ASCII knowledge base; `explain`/`query`/`path`/`affected` and the MCP server.

Environment: graphifyy 0.8.37, macOS, Python 3.14.

CJdev232:
## What

Fixes #1338.

`resolve_seed` in `graphify/affected.py` compared queries and labels using `.lower()` only, without Unicode normalization. As a result, a query in NFD form never matched an NFC-stored label (or vice versa) — even though the two are visually identical.

This affects:
- **macOS users** — APFS returns filenames in NFD, so any source/label flowing through the filesystem can be NFD while labels extracted from source code are NFC.
- **Accented Latin** — `Auditoría` (NFC) vs `Auditori` + combining acute (NFD).
- **Japanese with dakuten / handakuten** — `が` is one codepoint in NFC, `か` + combining `゛` in NFD.
- **Korean Hangul** — every syllable decomposes; `한` (NFC) vs three jamo (NFD).
- **Chinese** is largely unaffected (single-codepoint Han characters) except for CJK compatibility cases.

## Fix

A small `_normalize()` helper applies `unicodedata.normalize("NFC", ...)`. It's invoked on both the query and every label/source_file value before comparison in all 4 comparison sites inside `resolve_seed`. Both sides normalized → no assumption about which form the graph happened to store.

Why NFC (and not NFKC): NFC handles the practical macOS + CJK cases without folding full-width/half-width Latin or CJK compatibility forms — keeping the lookup faithful to the user's intent.

## Test

New `tests/test_affected.py` adds a unit test using a Spanish label (`Auditoría`) where the query is the NFD form of the NFC-stored label. Confirmed:
- Without the fix: `assert None == 'n1'` (the bug)
- With the fix: passes

All 7 existing `tests/test_affected_cli.py` tests continue to pass — no regression to bare-name matching, callable decoration, or tie-handling.

## Notes

- Real-user case: I hit this on macOS with bilingual (中文 / 日本語) node labels — `graphify explain` and `graphify path` would silently return "no match" for labels that visibly existed in the graph.
- The same NFD-vs-NFC pattern likely affects `serve._find_node` (the second symptom mentioned in the issue, around punctuation tokenization). Happy to follow up in a separate PR if useful — keeping this one minimal.

### DIFF
--- a/graphify/affected.py
+++ b/graphify/affected.py
@@ -1,11 +1,22 @@
 from __future__ import annotations
 
+import unicodedata
 from collections import deque
 from dataclasses import dataclass
 from pathlib import Path
 from typing import Iterable
 
 import networkx as nx
+
+
+def _normalize(text: str) -> str:
+    """Return *text* in a canonical form so NFD and NFC compare equal.
+
+    macOS stores filenames in NFD; Python string literals and most input
+    methods produce NFC. Comparing the two without normalization silently
+    fails for accented Latin, Japanese with dakuten, and Korean Hangul.
+    """
+    return unicodedata.normalize("NFC", text)
 
 
 DEFAULT_AFFECTED_RELATIONS = (
@@ -52,11 +63,11 @@ def _bare_name(label: str) -> str:
 def resolve_seed(graph: nx.Graph, query: str) -> str | None:
     if query in graph:
         return query
-    query_lower = query.lower()
+    query_lower = _normalize(query).lower()
     exact_label_matches = [
         str(node_id)
         for node_id, data in graph.nodes(data=True)
-        if str(data.get("label", "")).lower() == query_lower
+        if _normalize(str(data.get("label", ""))).lower() == query_lower
     ]
     if len(exact_label_matches) == 1:
         return exact_label_matches[0]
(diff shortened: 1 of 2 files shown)


safishamsi:
Superseded by #1353, which fixes the same `resolve_seed` normalization (using `casefold()` + also normalizing `_bare_name`) and additionally fixes the `serve._find_node` tokenized-label lookup — together fully closing #1338. #1353 is merged with green CI. Thanks @CJdev232 for the fix and the clear repro; closing this as the change is covered.

PR #1355, opened 2026-06-17
Add Qoder CLI platform support

wenxiaoyu:
## Summary
    - Add Qoder CLI as the 20th supported platform
    - New `skill-qoder.md` with Qoder-compatible YAML frontmatter (`triggers:` array, `metadata.scope: core`)
    - `_qoder_install`/`_qoder_uninstall`: project-scope skill to `.qoder/skills/graphify/` + `AGENTS.md` section
    - Registered in `_PLATFORM_CONFIG`, `_platform_skill_destination`, `install()` redirect, `main()` dispatch, `_project_install`/`_project_uninstall`, and
    `uninstall_all()`
    - 8 new tests (86/86 passing, zero regressions)

    ## Usage
    \`\`\`bash
    graphify install --platform qoder   # or: graphify qoder install
    graphify qoder uninstall
    \`\`\`

    ## Test plan
    - [x] `pytest tests/test_install.py` — 86/86 passed
    - [x] `graphify qoder install` writes `.qoder/skills/graphify/SKILL.md` + `AGENTS.md`
    - [x] `graphify qoder uninstall` removes both
    - [x] Idempotent install (no duplicate AGENTS.md sections)
    - [x] Noop uninstall when not installed


safishamsi:
Thanks @wenxiaoyu — the install/uninstall wiring in `__main__.py` is clean and the package-data/README updates are correct. But the PR **bypasses the skillgen system**, which is how platforms are defined on `v8`:

- It hand-writes `graphify/skill-qoder.md` — the only `graphify/skill-*.md` in the repo *not* generated by skillgen. Every other one is rendered from `tools/skillgen/` fragments + `platforms.toml` and snapshot-guarded in `tools/skillgen/expected/`.
- There's **no `[platform.qoder]` entry in `tools/skillgen/platforms.toml`** and no `tools/skillgen/expected/` snapshot, so the hand-authored file will silently rot when shared fragments change. `skillgen-check` only passes because there's no orphan-file guard.

Please add qoder to `platforms.toml` (it reads AGENTS.md with no PreToolUse hook, so mirror the `amp`/`trae` `hooks_variant = "agents-md"` shape), delete the hand-written monolith, regenerate via `python -m tools.skillgen`, then `--bless` and commit the resulting `graphify/skills/qoder/...` + `expected/` artifacts. After that `skillgen --check` should render qoder *and* pass — that's the real gate.

PR #1358, opened 2026-06-17
feat: make per-file extraction char cap configurable via GRAPHIFY_FILE_CHAR_CAP

NikitaUshakovV:
## Problem

`_read_files` truncates each file at `_FILE_CHAR_CAP = 20_000` chars before joining into the extraction prompt (`llm.py`), and `_estimate_file_tokens` caps at the same value so packing matches. For corpora made of a few **large** files (e.g. concatenated chat/dialogue logs of 100-600 KB each):

1. Each file loses 80-97% of its content — only the first 20k chars are extracted.
2. Token-packing under-counts, so `_pack_chunks_by_tokens` puts everything into a single chunk (`chunk 1/1`), which then extracts almost no edges.

Symptom: `graphify extract <large-corpus> --backend claude-cli` produces a sparse graph (many nodes, ~0 edges) regardless of `--token-budget`, because every file is silently clipped to its first 20k chars.

## Fix

Make the cap configurable via `GRAPHIFY_FILE_CHAR_CAP` (default unchanged at `20000`). Large-file corpora opt in with e.g. `GRAPHIFY_FILE_CHAR_CAP=200000`; default cost/context behavior is unchanged. README env-var table updated.

## Test

- default -> `_FILE_CHAR_CAP == 20000`
- `GRAPHIFY_FILE_CHAR_CAP=200000` -> `_FILE_CHAR_CAP == 200000`

### DIFF
--- a/README.md
+++ b/README.md
@@ -417,6 +417,7 @@ These are only needed for **headless / CI extraction** (`graphify extract`). Whe
 | `AWS_*` / `~/.aws/credentials` | AWS Bedrock — standard credential chain | `--backend bedrock` (no API key, uses IAM) |
 | `GRAPHIFY_MAX_WORKERS` | AST parallelism thread count | optional — also `--max-workers` flag |
 | `GRAPHIFY_MAX_OUTPUT_TOKENS` | Raise output cap for dense corpora | optional — e.g. `32768` for large files |
+| `GRAPHIFY_FILE_CHAR_CAP` | Per-file character cap before truncation in semantic extraction (default `20000`). Raise it for corpora of large individual files (e.g. concatenated chat logs) so the tail of each file isn't silently dropped | optional — e.g. `200000` |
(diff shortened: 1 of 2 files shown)


safishamsi:
Thanks @NikitaUshakovV — good problem statement (large files silently clipped to 20k chars is a real issue), and the default is correctly preserved. But the implementation needs hardening before merge:

1. **Import-time crash.** The cap is parsed at module top-level with a bare `int(os.environ.get("GRAPHIFY_FILE_CHAR_CAP", "20000"))`. An empty (`export GRAPHIFY_FILE_CHAR_CAP=`) or non-numeric value raises `ValueError` at *import*, crashing **every** `graphify` subcommand, not just extract. `0` extracts nothing; a negative value silently drops the tail.
2. **Follow the existing pattern.** The repo already has this convention in `_resolve_max_tokens`/`_resolve_api_timeout`/`GRAPHIFY_MAX_WORKERS`: `.strip()`, `try/except ValueError`, a `> 0` lower bound, silent fallback to default. Please wrap this in a resolver function the same way (also makes it testable without `importlib.reload`).
3. **Tests.** The PR description lists assertions, but no test file is in the diff. Add: default, valid override, empty, garbage, zero/negative → default.

PR #1359, opened 2026-06-17
docs: canonical entity-derived IDs for non-code nodes (fix doc/dialogue graph fragmentation)

NikitaUshakovV:
## Problem

The extraction-subagent spec mandates node IDs of the form `{parent_dir}_{filename}_{entity}` so they match the AST extractor. For **code** that's correct. But **non-code** nodes — `concept`, `rationale`, `document`, `paper` (named concepts, people, products, design rationale) — recur across many files and chunks, and there's no AST id to match. File-stem-prefixed IDs therefore give the *same* concept a *different* ID in every file, so the chunks never merge.

Result on doc/paper/dialogue corpora: the graph fragments into ~one tiny connected component per chunk -> **empty god-node rankings** and **~zero cross-file edges**. Observed while building a graph over concatenated dialogue logs: 808 nodes / 718 edges, avg degree ~1.0, no god-nodes — purely because each chunk's `Claude`/`nikita`/project nodes got distinct file-stem IDs.

## Fix

Add an explicit rule: non-code nodes derive their ID from the **entity alone** — a stable typed canonical slug (`person_nikita`, `product_myown_academy`, `concept_dom_first`) — never file-stem-prefixed. Code keeps the AST-matching `{stem}_{entity}` rule unchanged. Rebuilding the same corpus with canonical IDs: 929 nodes / 1124 edges, avg degree ~2.4, god-nodes populated.

Updated both fragment sources (full + compact) and regenerated the 13 per-platform `extraction-spec.md` references via `tools/skillgen`.


safishamsi:
Thanks @NikitaUshakovV — verified this is docs-only and the guidance is **accurate**: the entity-derived (non-file-stem) IDs you document for `concept`/`document`/`paper` nodes match what the code already does (e.g. `_make_id("framework", fw)` in extract.py), and it's the LLM-extraction path that needed this, so merge-fragmentation is genuinely reduced. No conflict with #1318.

One required fix: the PR updated the published `skills/*/references/extraction-spec.md` and the `fragments/` sources but didn't re-bless the skillgen snapshots, so `skillgen-check` / `test_skillgen.py` fail (4 tests, 'expected/ snapshot out of date'). Please run `python -m tools.skillgen --bless` and commit the 13 changed `tools/skillgen/expected/...extraction-spec.md` files. After that single commit it's a clean merge.

(Minor: AST `rationale` nodes are file+line-prefixed by design — worth a one-line note that the entity-alone rule targets the LLM path, not those.)

PR #1360, opened 2026-06-17
fix: Dart extractor cleanup

Overman775:
## What changed

This PR makes Dart extraction stricter about parser noise. It keeps the current regex-based approach, but avoids creating graph nodes from syntax fragments that are not project entities.

The patch covers a few cases that showed up in real Flutter code:

- `part of 'package:...'` files now resolve through `pubspec.yaml` to the parent file under `lib/`
- generated/private names like `_` and `_$Foo` no longer become graph nodes
- getters and modifiers like `String get`, `bool get`, `static const`, `static final` are filtered out
- function types, records, constructor params, and factory params no longer create comma-split labels
- common Dart SDK types such as `FutureOr`, `Never`, `MapEntry`, `Iterator`, `Comparable`, `StackTrace`, `Exception` are filtered from noisy type references
- SDK targets from `extension on String/List/Map/Set/Iterable/Future/Stream` and SDK typedef targets are filtered as noise
- SDK `Iterable<T>` from an extension type `implements` clause is filtered as noise
- Dart 3 syntax is covered: records, patterns, class modifiers, extension types, enhanced enums, wildcard `_`
- Dart `with MyMixin` still emits `mixes_in`

This is not a Dart analyzer rewrite. It only removes false positives that showed up in real projects.

The SDK filter is intentionally conservative. It covers common `dart:core` and `dart:async` noise, but does not try to mirror the full Dart SDK. Graphify filters by short name here, so a very broad SDK list could hide project classes with the same names.

## Why

The old extractor sometimes treated syntax fragments as project entities. On a real Flutter monorepo I saw nodes like:

```text
_
String get
bool get
static const
String sku, int
mocks_mock_webview_platform_mock
```

Those nodes are not useful graph facts. They make traversal noisier because they look like real classes, methods, or types.

## Real project check

I also ran the extractor against a Flutter monorepo with `1552` Dart files. Extraction errors stayed at `0`.

Noise comparison:

| item | before | after |
| --- | ---: | ---: |
| `_` labels | 211 | 0 |
| `String get` labels | 114 | 0 |
| `bool get` labels | 39 | 0 |
| `static const/final` labels | 49 | 0 |
| comma-label nodes | 49 | 0 |
| fake file-scoped `Mock` node | 1 | 0 |
| SDK `String/List/Map/Iterable/Future` labels | 119 | 0 |

Most structural relations stayed stable:

| relation | before | after |
| --- | ---: | ---: |
| imports | 6066 | 6066 |
| exports | 364 | 364 |
| inherits | 687 | 687 |
| calls | 351 | 351 |
| navigates | 40 | 40 |
| extends | 283 | 265 |
| implements | 580 | 579 |

The two drops are intentional. `extends` drops because SDK targets from ordinary Dart extensions are now filtered; project extension targets still remain. `implements` drops because one SDK `Iterable<T>` edge from an extension type is now filtered.

## Migration note

If you already have a graph for a Dart or Flutter project, run a forced rebuild once after updating:

```bash
graphify . --force
```

Older `graphify-out/graph.json` files may already contain the parser-noise nodes removed here. A forced rebuild removes those stale nodes instead of carrying them through an incremental update.

## Notes

The new tests focus on regression cases, not every Dart grammar rule. They cover the bugs above plus Dart 3 syntax that previously had a good chance of creating fake graph nodes.

### DIFF
--- a/graphify/extractors/dart.py
+++ b/graphify/extractors/dart.py
@@ -30,6 +30,94 @@ def extract_dart(path: Path) -> dict:
         return token
     src_clean = comment_string_pattern.sub(_comment_replace, src)
 
+    # Dart part files can point to their parent library through package: URIs.
+    # Resolve those through the nearest pubspec.yaml so part declarations use
+    # the parent library stem instead of a machine-local part-file path.
+    def _find_package_root(start: Path) -> tuple[Path, str] | None:
+        for parent in [start, *start.parents]:
+            pubspec = parent / "pubspec.yaml"
+            if not pubspec.exists():
+                continue
+            try:
+                for line in pubspec.read_text(encoding="utf-8", errors="replace").splitlines():
+                    m = re.match(r"\s*name\s*:\s*([A-Za-z0-9_]+)\s*$", line)
+                    if m:
+                        return parent, m.group(1)
+            except OSError:
+                return None
+        return None
+
+    def _resolve_part_parent(parent_ref: str) -> Path:
+        if parent_ref.startswith("package:"):
+            # Dart part files can name their parent via package:pkg/lib-relative path.
+            package_ref = parent_ref[len("package:"):]
+            package_name, _, package_path = package_ref.partition("/")
+            package_root = _find_package_root(path.parent)
+            if package_root and package_root[1] == package_name and package_path:
+                return (package_root[0] / "lib" / package_path).resolve()
+            return (path.parent / Path(package_path or package_name).name).resolve()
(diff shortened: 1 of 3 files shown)


Overman775:
@safishamsi Hi! Please take a look at the PR )

Overman775:
@safishamsi Rebased on latest v8, resolved conflicts. Could you re-review when you get a chance? )

Overman775:
Up

Overman775:
Up

Issue #945, opened 2026-05-20
Feature: fall back to .gitignore when .graphifyignore is absent


danyz81:
## Proposal

When a project has no `.graphifyignore`, fall back to reading `.gitignore` automatically. Keep current behavior when `.graphifyignore` exists (explicit always wins).

## Why

Most projects already maintain `.gitignore` for the same directories that pollute graphify scans (logs, build artifacts, caches). Without a fallback, first-time users hit a wall on real codebases.

Related: #303 — an author there suggests copying `.gitignore` to `.graphifyignore` as a workaround, which confirms the equivalence is intuitive to users.

## Real-world example

Firefox extension repo. `.gitignore` already covers `logs/`, `tmp/`, `telegram-export/`.

| Run | total_files | total_words |
|---|---|---|
| Out of the box | 10,239 | 72,882,253 |
| After copying ignore rules to `.graphifyignore` | 163 | 308,159 |

Reduction: -98.4% files, -99.6% words. The `.graphifyignore` I had to create is a near-duplicate of `.gitignore`.

## Design notes

- Behavior: if `.graphifyignore` exists → use it (current behavior). Else if `.gitignore` exists → use it.
- Backward compatible.
- Opt-out: `--no-gitignore-fallback` flag or env var.
- Detection point: `_load_graphifyignore()` in `detect.py` (~line 464).

## Alternative

A note in the README's "Ignoring files" section recommending users start from `.gitignore`. Lower effort, partial fix.

safishamsi:
Fixed in v0.8.14. When no `.graphifyignore` exists in a directory, `.gitignore` patterns are now used as fallback. `.graphifyignore` still takes precedence when both exist. Regression tests added.

PR #1364, opened 2026-06-17
Merge .gitignore into .graphifyignore instead of replacing it (#1363)

### LINKED ISSUE #1363: .graphifyignore silently replaces .gitignore instead of merging — secrets ignored only in .gitignore get indexed into the graph
## Summary

When a directory contains a `.graphifyignore`, graphify **stops reading that directory's `.gitignore` entirely** — the two are *not* merged. A pattern present only in `.gitignore` is silently dropped, so files the user believed were ignored get pulled into the knowledge graph.

The README describes this as *"If both exist, `.graphifyignore` takes priority"*, which reads like **merge-with-precedence**. The actual behavior is **full replacement** of that directory's `.gitignore`. Those are very different, and the difference is a security footgun.

## Why this is a security problem

`graph.json` / `GRAPH_REPORT.md` / `graph.html` embed file contents and are routinely committed. A realistic, easy-to-hit sequence:

1. A project ignores secrets via `.gitignore` (e.g. `prod-dump.sql`, `customer-data.json`, `config/local.yaml`).
2. Someone later adds a `.graphifyignore` to exclude media (`*.png`, `*.pdf`, …) and **forgets to copy the `.gitignore` entries into it**.
3. graphify now ignores `.gitignore`, indexes the secret file into the graph.
4. The graph artifacts get committed/pushed — secret leaks to a (possibly public) repo.

Two sources of truth for "what to ignore," where adding the second one silently disables the first, is exactly the shape of mistake that leaks data.

The built-in sensitive-file detection (`_is_sensitive`: `.env`, `*.pem`, `*.key`, `id_rsa`, files with load-bearing `secret`/`token`/`password`/`credential` in the name) mitigates the *obvious* cases, but does **not** cover neutrally-named secrets like `prod-dump.sql`, `customer-data.json`, or `seed.yaml` — precisely the files people rely on `.gitignore` to exclude.

## Reproduction (verified on graphifyy 0.8.40)

```bash
T=$(mktemp -d)
printf 'build-artifact.md\n' > "$T/.gitignore"   # gitignore excludes this non-secret file
printf '# keep\n'  > "$T/keep.md"
printf '# art\n'   > "$T/build-artifact.md"
printf 'x\n'       > "$T/pic.png"

python -c "from graphify.detect import detect; from pathlib import Path; \
print(sorted(p.split('/')[-1] for p in detect(Path('$T'))['files'].get('document',[])))"
# -> ['keep.md']                      (build-artifact.md correctly excluded by .gitignore)

printf '*.png\n' > "$T/.graphifyignore"           # add a .graphifyignore that ONLY lists *.png
python -c "from graphify.detect import detect; from pathlib import Path; \
print(sorted(p.split('/')[-1] for p in detect(Path('$T'))['files'].get('document',[])))"
# -> ['build-artifact.md', 'keep.md']  (.gitignore is now IGNORED; build-artifact.md re-appears)
```

`build-artifact.md` reappears purely because a `.graphifyignore` exists in the same directory — even though that `.graphifyignore` says nothing about it.

## Root cause

`graphify/detect.py`, `_load_graphifyignore()`:

```python
# Prefer .graphifyignore; fall back to .gitignore so projects that already
# maintain a .gitignore get sensible defaults without duplicating it (#945).
ignore_file = d / ".graphifyignore"
if not ignore_file.exists():
    ignore_file = d / ".gitignore"
```

Per directory, it reads **one or the other**, never both.

## Expected behavior

`.gitignore` and `.graphifyignore` should be **merged**, not mutually exclusive — read `.gitignore` first, then append `.graphifyignore` so its patterns win on conflicts via the existing last-match-wins semantics. That matches the natural mental model ("graphify *additionally* ignores these") and removes the footgun: adding a `.graphifyignore` can only ever exclude *more*, never silently re-include something `.gitignore` excluded.

## Suggested fixes (in order of preference)

1. **Merge both files** (gitignore outer, graphifyignore inner/last-match-wins). Single change, eliminates the footgun.
2. **A global ignore** (e.g. `~/.config/graphify/ignore`, or honor git's `core.excludesFile`) so users don't have to duplicate media/secret patterns into every repo. There's currently no global-ignore mechanism at all.
3. If replacement semantics must stay: at minimum, **emit a loud warning** when a `.graphifyignore` shadows a non-empty sibling `.gitignore`, and **fix the README wording** — "takes priority" should read "replaces `.gitignore` in that directory (patterns are not merged)."

Happy to send a PR for option 1 if that direction is acceptable.

suenot:
Fixes #1363.

## Problem

Within a directory, `_load_graphifyignore` reads **either** `.graphifyignore` **or** `.gitignore`, never both:

```python
ignore_file = d / ".graphifyignore"
if not ignore_file.exists():
    ignore_file = d / ".gitignore"
```

So adding a `.graphifyignore` silently disables that directory's `.gitignore`. A file ignored **only** in `.gitignore` (a `prod-dump.sql`, `customer-data.json`, neutrally-named secret, …) then gets indexed into the graph — and can leak when `graph.json` / `GRAPH_REPORT.md` / `graph.html` are committed. The built-in sensitive-file detection catches `.env`/keys, but not neutrally-named secrets people rely on `.gitignore` to exclude.

The README's "`.graphifyignore` takes priority" reads like merge-with-precedence; the behavior is total replacement.

## Fix

Read **both** files and merge them per directory: `.gitignore` first, then `.graphifyignore` appended last so it still wins on direct conflicts via the existing last-match-wins semantics. Adding a `.graphifyignore` can now only ever exclude **more**, never silently re-include a path `.gitignore` excluded.

- The #945 fallback (honor `.gitignore` when no `.graphifyignore` exists) is unchanged.
- `.graphifyignore` precedence on conflicts is preserved (it's read last → wins), including `!` re-includes.

## Tests

- Rewrote `test_graphifyignore_takes_precedence_over_gitignore` → `test_graphifyignore_merges_with_gitignore`: both files' patterns apply; `.graphifyignore`'s `!negation` wins on conflict.
- Added `test_secret_ignored_only_in_gitignore_stays_excluded`: the #1363 scenario — a file excluded only by `.gitignore` stays out even when an unrelated `.graphifyignore` is present.
- `tests/test_detect.py`: **125 passed**.

### DIFF
--- a/graphify/detect.py
+++ b/graphify/detect.py
@@ -757,16 +757,20 @@ def _load_graphifyignore(root: Path) -> list[tuple[Path, str]]:
 
     patterns: list[tuple[Path, str]] = []
     for d in dirs:
-        # Prefer .graphifyignore; fall back to .gitignore so projects that already
-        # maintain a .gitignore get sensible defaults without duplicating it (#945).
-        ignore_file = d / ".graphifyignore"
-        if not ignore_file.exists():
-            ignore_file = d / ".gitignore"
-        if ignore_file.exists():
-            for raw in ignore_file.read_text(encoding="utf-8", errors="ignore").splitlines():
-                line = _parse_gitignore_line(raw)
-                if line:
-                    patterns.append((d, line))
+        # MERGE .gitignore + .graphifyignore — do NOT let one shadow the other.
+        # Read .gitignore first, then .graphifyignore, so .graphifyignore patterns
+        # are appended last and win on conflicts via last-match-wins. This keeps
+        # ".graphifyignore takes priority" while ensuring that adding a
+        # .graphifyignore can only ever exclude MORE — never silently re-include a
+        # path the .gitignore already excluded (which previously let secrets that
+        # were ignored only in .gitignore get indexed into the graph). (#1363)
+        for fname in (".gitignore", ".graphifyignore"):
+            ignore_file = d / fname
+            if ignore_file.exists():
+                for raw in ignore_file.read_text(encoding="utf-8", errors="ignore").splitlines():
+                    line = _parse_gitignore_line(raw)
+                    if line:
+                        patterns.append((d, line))
     return patterns
 
 
--- a/tests/test_detect.py
+++ b/tests/test_detect.py
@@ -922,19 +922,47 @@ def test_gitignore_fallback_when_no_graphifyignore(tmp_path):
     assert not any("generated" in f for f in code)
 
 
-def test_graphifyignore_takes_precedence_over_gitignore(tmp_path):
-    """When both exist, .graphifyignore is used and .gitignore is ignored (#945)."""
+def test_graphifyignore_merges_with_gitignore(tmp_path):
+    """When both exist they are MERGED, not mutually exclusive: .gitignore patterns
+    still apply, and .graphifyignore only wins on direct conflicts via last-match-wins.
+
+    Previously .graphifyignore fully shadowed .gitignore in the same directory, so a
+    file ignored only in .gitignore (e.g. a secret) would get indexed into the graph
+    once any .graphifyignore was added. (#1363)
+    """
     (tmp_path / ".git").mkdir()
-    # .gitignore would exclude main.py; .graphifyignore excludes only other.py
-    (tmp_path / ".gitignore").write_text("main.py\n")
-    (tmp_path / ".graphifyignore").write_text("other.py\n")
-    (tmp_path / "main.py").write_text("x = 1")
-    (tmp_path / "other.py").write_text("x = 2")
+    (tmp_path / ".gitignore").write_text("gitonly.py\nkeep.py\n")
+    # .graphifyignore adds its own pattern and re-includes keep.py via negation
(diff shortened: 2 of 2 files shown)


safishamsi:
Thanks @suenot - this is exactly the right fix, and your diagnosis was spot on.

We landed the identical fix independently in `b2a1722` (on v8, closing #1363) about an hour before I got to this PR: read both files per directory, `.gitignore` first and `.graphifyignore` last so it still wins on conflicts (including `!` negations) via last-match-wins, with the same security framing (a `.gitignore`-only secret like `prod-dump.sql` must stay excluded when a `.graphifyignore` is added). We even converged on the same test rewrite (`test_graphifyignore_takes_precedence_over_gitignore` → merge semantics) plus a secret-stays-excluded regression, and the README wording fix.

Since the fix is already on v8 I'm closing this PR as resolved - but full credit to you for the report-quality diagnosis and the correct approach. Apologies for the duplicated effort; I should have checked the open PRs before implementing. Really appreciate the contribution.

Issue #1361, opened 2026-06-17
/graphify --update: build_merge called without root= → prune_sources never matches, graph inflates on every update


bchan84x:
## Summary

The `/graphify <path> --update` skill runbook calls `build_merge(prune_sources=...)` **without passing `root=`**, so the abs/rel path normalization added for #1007 is bypassed. `prune_sources` (absolute paths) never matches the graph's relative `source_file` values, **nothing is pruned**, and the graph **accumulates duplicate/stale nodes on every incremental update**. The native `graphify update` CLI is unaffected — it passes `root`.

Version: graphifyy 0.8.38

## Root cause

`build_merge` only relativizes `prune_sources` when `root` is provided (`build.py`):

```python
_root_str = str(Path(root).resolve()) if root is not None else None
prune_set = set()
for p in prune_sources:
    prune_set.add(p)
    norm = _norm_source_file(p, _root_str)   # no-op when _root_str is None
    if norm:
        prune_set.add(norm)
to_remove = [n for n, d in G.nodes(data=True) if d.get("source_file") in prune_set]
```

`_norm_source_file(p, None)` returns `p` unchanged for absolute paths (the `if root and os.path.isabs(p)` guard fails). Graph nodes store **relative** `source_file`, so `d.get("source_file") in prune_set` never matches.

The shipped update runbook (`references/update.md`, ~line 104) calls it **without `root`**:

```python
G = build_merge(
    [new_extraction],
    graph_path='graphify-out/graph.json',
    prune_sources=prune,
)
```

and `prune` is built from `detect_incremental(...)`'s `new_files` / `deleted_files`, which are **absolute** paths — so the mismatch is guaranteed.

By contrast the native CLI (`__main__.py`, incremental branch) does it correctly:

```python
G = _build_merge([merged], graph_path=existing_graph_path,
                 prune_sources=deleted_files or None,
                 dedup=True, dedup_llm_backend=dedup_backend, root=target)
```

## Observed

Running `/graphify . --update` after editing 4 tracked docs:

```
[graphify] 4 source file(s) deleted since last run — no matching nodes or edges in graph, already clean.
[update] Merged: 737 nodes   # ← 4 changed files' OLD nodes survived; 28 new appended
```

Re-running `build_merge` with the same prune list relativized against root:

```
[graphify] Pruned 84 node(s) from 4 deleted source file(s).
[update] Merged: 653 nodes   # ← correct
```

[...]

PR #1366, opened 2026-06-17
fix(skill): pin extraction source_file + root the full build so --update stops duplicating

RelywOo:
## Problem

After #1344 (build_merge replace-on-re-extract) and #1361 (pass `root=` to `build_merge` in the `--update` runbook), the full build and an incremental `/graphify --update` could still emit **different `source_file` bases for the same file** (e.g. `docs/wiki/overview.md` from one run vs basename `overview.md` from another). Because `build_merge`'s replace keys on the `source_file` string, the mismatch made it miss the stored node, so re-extraction **accumulated duplicates** instead of replacing. Two gaps remained:

1. **`extraction-spec(.md/-compact.md)`** — the subagent's `source_file` slot was an unpinned `"relative/path"`, so it invented a base per run. The node id (derived from the same path) drifted too.
2. **`core.md`** — the full build called `build_from_json` **without `root=`**, so #1361's update-side `root=` had no matching base on the full-build side; the two halves never agreed.

## Fix (engine `build.py` unchanged)

- **extraction-spec(+compact):** pin `source_file` to the **verbatim `FILE_LIST` path** — no basename, no re-relativize, no separator change. `_norm_source_file(root)` then canonicalizes every run identically.
- **core.md:** pass `root='INPUT_PATH'` at **both** `build_from_json` sites (Step 4 export, Step 5 report), matching #1361's update-side root so full build and `--update` relativize to the same base.
- **update.md:** `prune_sources = deleted` only. Changed files are already replaced by `build_merge` (#1344); once `root=` aligns the bases, leaving `changed` in `prune_sources` would **delete the freshly re-extracted nodes**.

All skill artifacts regenerated via `python tools/skillgen/gen.py`.

## Test

Adds `test_build_merge_root_collapses_convention_drift`: a drifted relative base with no root leaves 3 nodes (stale + duplicate); the verbatim path + `root=` collapses to 1 canonical node and drops the stale node. `tests/test_build.py` — 35 passed.

Builds on #1344 and #1361 (same bug family).


safishamsi:
Merged to v8 in `897483f` — thank you @RelywOo, this is an excellent catch and a genuinely important fix.

Your diagnosis is exactly right, and it's more consequential than the PR framing suggests: I verified that after #1361 added `root=` to the `--update` `build_merge` call (shipped in 0.8.41), `build_merge`'s prune — which runs *after* the merge — began matching the freshly re-extracted nodes of **changed** files (still in `prune_sources`) and deleting them. Direct repro:

```
0.8.41 (prune = deleted + changed, root passed):  changed file's fresh node -> DELETED
this PR (prune = deleted only):                   changed file's fresh node -> preserved
```

So `--update` on a changed file was wiping that file's nodes in 0.8.41. Dropping `changed` from `prune_sources` (since #1344's replace-on-re-extract already reconciles changed files), passing `root=` to the full build's `build_from_json` at both sites, and pinning the extraction-spec `source_file` to the verbatim path all compose correctly.

Validation on merge: clean merge onto v8, `skillgen --check` passes (re-blessed the `expected/` snapshots, which were the only thing missing from the PR), full suite **2129 passed**, and your `test_build_merge_root_collapses_convention_drift` is in. Credited you via `Co-Authored-By`.

Because this is a regression in the current 0.8.41 release, I'll cut a prompt 0.8.42 with it. Really appreciate the rigorous writeup.

PR #1367, opened 2026-06-17
feat: Solidity AST extraction via tree-sitter-solidity

### LINKED ISSUE #1362: feat: add Solidity (.sol) AST extraction
I use graphify on web3 repos (Foundry/Hardhat) and `.sol` files get skipped entirely right now. They are not in `CODE_EXTENSIONS` and there is no entry in `_DISPATCH`, so contracts never show up in the graph.

## The gap

Graphify picks up code in two places:

1. `graphify/detect.py` — `CODE_EXTENSIONS` decides what counts as code
2. `graphify/extract.py` — `_DISPATCH` maps a suffix to an extractor

`.sol` is in neither. Solidity files never enter the AST pass. The fix is the same as every other language in `ARCHITECTURE.md`: register the extension, add an extractor, wire tests.

## Approach

Use `tree-sitter-solidity` from PyPI (https://pypi.org/project/tree-sitter-solidity/) with the existing `_extract_generic` path in `extract.py`. Pre-built wheels, works with graphify's `tree-sitter>=0.23`. Regex-only (like Apex `.cls`) would miss modifiers and fall over on real project files.

Relevant AST node types from a quick probe on sample Solidity:

- Declarations: `contract_declaration`, `interface_declaration`, `library_declaration`, `struct_declaration`, `enum_declaration`
- Members: `function_definition`, `constructor_definition`, `modifier_definition`, `event_definition`, `fallback_receive_definition`
- Structure: `import_directive`, `inheritance_specifier`, `using_directive`
- Behavior: `call_expression`, `new_expression`, `emit_statement`, `member_expression`

### Nodes

- `contract Foo` → node `Foo`, `contains` from file
- `interface IBar` → node `IBar`
- `library Lib` → node `Lib`
- `struct Point`, `enum Status` → nodes on file
- `function transfer()` → `.transfer()`, `method` edge on parent
- `constructor()`, `modifier onlyOwner`, `receive()` / `fallback()` → same pattern
- `event Transfer` → node `Transfer`, `contains` from contract

File-level node ID must use `_file_node_id()` / `_make_id(str(path))` so AST nodes line up with semantic subagent IDs (#1033).

### Edges

**Inheritance** (`contract Token is Base, IToken`):
- Pre-scan file/corpus for `interface_declaration` names
- Known interface parent → `implements`
- Contract/library parent → `inherits`
- `I` prefix fallback only when name is unknown; some codebases skip the prefix

**Imports**:
- `import "./interfaces/IERC20.sol"` → resolve relative to importing file, `imports` to real file node (like `_import_c`)
- `import {SafeMath, IERC721} from "..."` → stub per symbol
- `@openzeppelin/...` paths → stub by stem unless remapping resolves them

**Foundry remappings**: parse `remappings.txt` and `foundry.toml` `[profile.default.remappings]` so `@openzeppelin/contracts/...` can resolve to `lib/openzeppelin-contracts/contracts/...`. Without this, Foundry imports are orphan stubs.

**Calls**:
- `transfer()` → same-contract via `raw_calls` second pass
- `Lib.add()` → library receiver
- `new Token()` → `instantiates`
- `emit Transfer(...)` → `emits`
- Filter builtins (`require`, `assert`, `revert`, `msg`, `abi`, `block`, `tx`, etc.) so they do not become god nodes

**Using directives** (`using SafeMath for uint256`) → `uses` / `references` to library node.

Optional if cheap: state variable type refs, NatSpec → `rationale_for`, `foundry.toml` / `hardhat.config` as config nodes, ABI JSON in `artifacts/`.

## Edge cases

- Vendor code in `lib/`: do not hard-skip by default; document `.graphifyignore` for `lib/`, `out/`, `cache/`. `node_modules` is already skipped.
- Imports that only resolve with remappings (common in Foundry) — include remapping support in the initial PR, not a follow-up.
- Multiple contracts per file: one file node, multiple `contains` edges.
- `assembly { }`: partial parse is fine; `_safe_extract` catches failures.
- Solidity 0.4.x–0.8.x: test with `pragma solidity ^0.8.0`; add older-syntax fixture if gaps show up.
- Diamond / proxy patterns: AST alone cannot resolve facet routing or opaque delegatecall targets. Do not fake edges; mark `AMBIGUOUS` at most.
- Generated `out/` / `cache/`: recommend `.graphifyignore`; do not add to `_SKIP_DIRS` without maintainer buy-in.

## Implementation checklist

**Dependency**: `tree-sitter-solidity>=1.2,<2.0` in main `pyproject.toml` dependencies (has wheels on all major platforms).

**Files**:
- `graphify/detect.py` — add `'.sol'` to `CODE_EXTENSIONS`
- `graphify/extract.py` — `_SOLIDITY_CONFIG`, `_import_solidity()`, inheritance/emit hooks in `_extract_generic` walk, `extract_solidity()`, `".sol"` in `_DISPATCH`
- Skill `update.md` fragments / `tools/skillgen` — add `.sol` to `code_exts`
- `CHANGELOG.md`

**Tests**:
- `tests/fixtures/sample.sol` plus cross-file fixtures (`Base.sol`, `interfaces/IERC20.sol`)
- `tests/fixtures/foundry.toml` + `remappings.txt` if remapping support lands
- `tests/test_solidity.py` — dispatch, CODE_EXTENSIONS, entities, imports, inherits/implements, calls, emits, new, no dangling edges

**Validation**:
```bash
uv sync --all-extras
uv run pytest tests/test_solidity.py tests/test_detect.py -q
uv run graphify extract /path/to/foundry-project --no-cluster
```

## Open questions

1. Add `lib/` to `_SKIP_DIRS`? Leaning no; `.graphifyignore` is more flexible.
2. `[solidity]` optional extra vs main dependency? Main deps is simpler given wheels exist.
3. `foundry.toml` / `hardhat.config` parsing in same PR or fast follow-up?

## Prior art in repo

- `.psm1` (#1315): CODE_EXTENSIONS + _DISPATCH gap
- `.astro` (#850): extension + AST pass
- `.cls` Apex: regex when no PyPI grammar
- `.tf` HCL: tree-sitter optional extra

Solidity is closest to Java/C#: tree-sitter + `_extract_generic` + import handler + inheritance hook.

---

I would like to work on this. Please assign me (@guglxni).

guglxni:
This PR implements full Solidity `.sol` AST extraction via `tree-sitter-solidity`, closing #1362.

### What's new
- **Contracts, interfaces, libraries, structs, enums, functions** are indexed with full node metadata.
- **Import resolution** supports both relative imports and Foundry remappings (`remappings.txt` + `foundry.toml`).
- **Inheritance / implements** edges split correctly between `inherits` and `implements`.
- **Events, emit, new/instantiates** and **call-graph** edges (`calls`, `raw_calls`) are emitted.
- **Traversal guards** prevent path escapes via malicious relative imports or remapped imports.
- **Primitive type filter** uses a dual-layer approach (frozenset + regex) to avoid spurious `references` edges for sized types (`uint128`, `bytes32`, `fixed128x18`, etc.).

### Test coverage
22 tests covering:
- Foundry `remappings.txt` and `foundry.toml` remapping
- Multi-contract files, abstract contracts, interface-without-I-prefix
- Malformed / missing imports
- Primitive type filtering
- Relative path traversal guard
- Remapping-based traversal guard

### Docs
- Added `docs/solidity-support.md` with usage notes
- CHANGELOG entry under `## Unreleased`

### Skillgen
- Regenerated platform skill fragments across all 13 platforms for `.sol` extension support

Closes #1362


guglxni:
@safishamsi Pls review.

PR #1369, opened 2026-06-17
feat: intra-file markdown slicing and adaptive retry (1/3)

Const011:
## Context

Split from former [#1326](https://github.com/safishamsi/graphify/pull/1326) (closed) per review feedback — **part 1 of 3**. See also the follow-up PRs for incremental safety and cross-file stitch.

## Motivation

Graphify incremental `extract` is how we maintain a knowledge graph as a corpus grows file-by-file. In production we hit failures where large documents **silently overflow** the model context: semantic extraction appears to succeed (exit 0) but returns empty or invalid JSON — especially under token pressure (local Ollama, tight budgets, or long `.md` files).

Without fixing chunking, downstream merge can proceed as if extraction worked when it did not.

## Problem: silent truncation on large documents

**Symptom:** Semantic extraction on large `.md` files returned empty graphs or unparseable JSON while the CLI still exited 0.

**Root cause:** `_pack_chunks_by_tokens` treated each file as atomic. A single file larger than the chunk budget was sent whole; the model truncated output (`finish_reason="length"`) or returned JSON that failed to parse. The chunk was dropped or merged as empty with no hard failure.

## Fix

New module `graphify/file_slice.py`:

- Splits oversized **splittable** text (`.md`, `.mdx`, `.txt`, `.rst`) into `FileSlice` units at heading / paragraph boundaries before packing.
- Adaptive retry can **bisect** a slice when a chunk still overflows context (`split_unit_for_retry`, `split_chunk_for_retry`).
- `llm.py` orchestrates only: `expand_oversized_files()` → pack → extract; slice logic lives in `file_slice.py`.

This matches how full runs co-pack related files for cross-file edges, but fixes the case where **one file alone** exceeds the budget.

## Files

| File | Change |
|------|--------|
| `graphify/file_slice.py` | New — slicing + bisect helpers |
| `graphify/llm.py` | Wire `SemanticUnit` / slices into packing and adaptive retry |
| `tests/test_file_slice.py` | New |
| `tests/test_chunking.py` | Slice-aware chunking regressions |

**Not included here (other parts of #1326):** incremental abort-on-failure (PR 2), cross-file stitch (PR 3), OpenRouter/BYOK backend wiring (upstream #1273 covers env-based OpenAI URLs).

## Test plan

- [x] `pytest tests/test_file_slice.py tests/test_chunking.py tests/test_llm_parser.py tests/test_extract.py`
- [ ] Upstream CI on Linux

### DIFF
--- a/graphify/file_slice.py
+++ b/graphify/file_slice.py
@@ -1,0 +1,253 @@
+"""Intra-file slicing for semantic LLM extraction.
+
+Large markdown/text documents can exceed a model context window or graphify's
+per-chunk token budget. FileSlice represents a byte range within one source
+file; several slices still cache and increment under the parent file path.
+"""
+from __future__ import annotations
+
+import re
+from dataclasses import dataclass
+from pathlib import Path
+import sys
+
+# Document-like extensions eligible for intra-file splitting.
+_SPLITTABLE_SUFFIXES = frozenset({".md", ".mdx", ".txt", ".rst"})
+
+_CHARS_PER_TOKEN = 4
+_PER_FILE_OVERHEAD_CHARS = 160
+
+_HEADING_SPLIT = re.compile(r"(?=^#{1,6}\s)", re.MULTILINE)
+
+
+@dataclass(frozen=True)
+class FileSlice:
+    """A character range within a single on-disk file."""
+
+    path: Path
+    start_char: int
+    end_char: int
+    slice_index: int = 0
+    slice_count: int = 1
+
+    def source_location(self) -> str:
+        return f"chars:{self.start_char}-{self.end_char}"
+
+
+SemanticUnit = Path | FileSlice
+
+
+def is_file_slice(unit: SemanticUnit) -> bool:
+    return isinstance(unit, FileSlice)
+
+
+def unit_path(unit: SemanticUnit) -> Path:
+    return unit.path if is_file_slice(unit) else unit
+
+
+def is_splittable_text(path: Path) -> bool:
+    return path.suffix.lower() in _SPLITTABLE_SUFFIXES
+
+
+def read_unit_text(unit: SemanticUnit) -> str:
+    if is_file_slice(unit):
+        full = unit.path.read_text(encoding="utf-8", errors="replace")
+        return full[unit.start_char : unit.end_char]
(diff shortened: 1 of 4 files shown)


safishamsi:
Thank you @Const011 - the diagnosis here is excellent and the problem is real. We've implemented intra-file slicing on v8 in `4f539e7`, crediting your analysis.

We verified the underlying bug independently: `_read_files` caps every file at `_FILE_CHAR_CAP` (20,000 chars), so a document longer than that had everything past the cap silently dropped, and the adaptive-retry path "can't make one file smaller than itself." Your framing of the fix - slice at heading/paragraph boundaries before packing, keep the parent `source_file` so the graph isn't fragmented, and bisect a slice on retry - is exactly right, and our implementation follows that shape (`graphify/file_slice.py`).

We landed our own version rather than merging this PR for two reasons, both caught by existing tests against your branch:

1. **`_response_is_hollow` regression** - the added `and finish_reason != "stop"` guard (`llm.py`) disables hollow-response recovery, because hollow responses *always* carry `finish_reason="stop"` - that's the silent-truncation symptom itself. It fails `test_call_openai_compat_relabels_unparseable_json_as_length` and would re-introduce the exact "exit 0 but empty JSON" bug this PR targets.
2. **`_estimate_file_tokens` lost its `_is_vision_image` branch** - so it reads PNG bytes as text (fails `test_image_token_estimate_is_flat`).

Our version keeps those untouched, scopes slicing strictly to splittable text (`.md/.mdx/.markdown/.txt/.rst` - never code or PDFs), and adds 18 regression tests (full-coverage/contiguity invariant, the parent-`source_file` anti-fragmentation guarantee, slice bisection). Full suite green (2147 passed).

Closing this PR as implemented-from-the-diagnosis. Genuinely appreciate the clear writeup - it made the verification straightforward. (Notes on parts 2/3 coming on those PRs.)

Const011:
Thank you, @safishamsi, for incorporating the fix and for the thorough explanation of the implementation. We are incredibly grateful for the work you've put into graphify; it truly feels like the missing piece necessary to move AI toward true AGI. If you are open to it, we would love to continue contributing patches and feature proposals to ensure graphify optimized for documents written in natural language.

PR #1374, opened 2026-06-17
fix: resolve package subpath exports in import graph

pkudinov:
## Summary

This PR fixes Graphify import resolution for package subpath exports that point at source files in monorepos using `package.json` `exports` with subpath keys.

### What this changes

- In `graphify/extract.py`, resolves subpath export target strings through `_package_export_target_values`.
- Added array-form support for export targets so both `"./foo.ts"` and `["./foo.ts", "./foo.dev.ts"]` forms are handled recursively.
- Extended `tests/test_js_import_resolution.py` with a regression test covering pnpm workspace imports of package subpath exports.

### Why this matters for tokenbooks

`packages/accounting` uses package subpath exports (for browser/server split) and Graphify currently recorded the target as `browser` instead of resolving through `package.json` to a source file path.
That prevented accurate Jest test impact and could over-run package-level test selection.

### Verification

- `uv run pytest -q`
  - Result: `2098 passed, 28 skipped`.

### Notes

This also resolves the rebase conflict from prior comment context by preserving both npm/yarn and pnpm import scenarios in existing test coverage.

### DIFF
--- a/graphify/extract.py
+++ b/graphify/extract.py
@@ -383,6 +383,61 @@ def _load_workspace_packages(start_dir: Path) -> dict[str, Path]:
     return packages
 
 
+_PACKAGE_EXPORT_CONDITION_PRIORITY = ("source", "import", "module", "default", "require", "types", "svelte")
+
+
+def _package_export_target_values(value: Any) -> list[str]:
+    if isinstance(value, str):
+        return [value]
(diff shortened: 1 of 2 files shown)


pkudinov:
Closing this as a duplicate. I pushed the same rebased fix to the original PR: #1333.

PR #1375, opened 2026-06-18
fix(deps): bump vulnerable dependencies to patched versions

hypnwtykvmpr:
Dependabot flags several known-vulnerable dependencies in `uv.lock`. This bumps each to its first patched release. All are version-only changes (no code changes), and the full test suite passes.

| Dependency | From | To | Advisory |
|---|---|---|---|
| pypdf | 6.11.0 | 6.13.3 | GHSA-cj93-chg6-vgv8 / CVE-2026-48155 (layout-mode memory) and GHSA-248m-82v9-q6g6 / CVE-2026-48156 (zero-width xref runtime) — both fixed in 6.12.0 |
| yt-dlp | 2026.3.17 | 2026.6.9 | fixed in 2026.6.9 |
| pyjwt | 2.12.1 | 2.13.0 | fixed in 2.13.0 |
| cryptography | 48.0.0 | 49.0.0 | fixed in 48.0.1 |
| python-multipart | 0.0.28 | 0.0.32 | fixed in 0.0.30 |

For the two direct deps (`pypdf`, `yt-dlp`) I also added a lower-bound floor in `pyproject.toml`, so `pip install` / `uv tool install` users get the patched version, not just the dev lockfile. The remaining three are transitive and bumped in the lock.

Note: `nltk` (<= 3.9.4) is also flagged by Dependabot but has no fixed release yet, so it is intentionally left unchanged here.

Tested: full `pytest` suite passes with the bumps.


safishamsi:
Merged to v8 in `9e8fa2a` — thanks @hypnwtykvmpr. Validated all five bumps against real installs (pypdf 6.13.3 PDF path verified end-to-end; cryptography/pyjwt/python-multipart are transitive via `mcp`, unused in graphify source). One addition on merge: regenerated `uv.lock` so the project version reads 0.8.42 (the PR's lock still recorded 0.8.41, which would trip the lockfile-consistency gate); all five security versions intact. Full suite green.

Issue #550, opened 2026-04-25
Node ID collision when multiple files share the same filename (e.g. ASP.NET Razor Pages)


tetiz123:
## Problem

  In ASP.NET Razor Pages projects, every page follows the convention:

  Pages/Attendance/Index.cshtml.cs  → class IndexModel
  Pages/Approval/Index.cshtml.cs    → class IndexModel
  Pages/Overtime/Index.cshtml.cs    → class IndexModel
  Pages/RemoteWork/Index.cshtml.cs  → class IndexModel

  All four files have the same filename (Index.cshtml.cs) and the same class name (IndexModel).
  graphify merges them into a single node (index_cshtml_indexmodel), which makes the graph incorrect for navigation.

  ## Request

  Please add an option like --full-path-ids that includes the folder structure in node IDs:

  attendance_index_cshtml_indexmodel
  approval_index_cshtml_indexmodel
  overtime_index_cshtml_indexmodel
  remotework_index_cshtml_indexmodel

  This would also help other frameworks that use convention-based naming (e.g. Next.js page.tsx files).

safishamsi:
Fixed: added `_file_stem(path)` helper that returns `parent.stem` (e.g. `api.utils`) instead of bare `stem`. All 11 extractor `stem = path.stem` assignments replaced. Java cross-file resolution also fixed from `_make_id(path.stem)` to `_make_id(str(path))`.

PR #1378, opened 2026-06-18
Unify node-ID normalization into a single source of truth (#1)

danielnguyenfinhub:
extract._make_id and build._normalize_id each carried a copy of the same NFKC -> \w -> underscore-collapse -> casefold recipe, kept in sync only by mirrored docstrings. That duplication is the root of the recurring ID-drift bug class (#811, #550, #1033, #1104): the AST extractor, the LLM subagents, and the graph builder must all produce identical node IDs or a single entity splits into disconnected ghost nodes.

Move the recipe into a new graphify/ids.py (normalize_id + make_id). extract and build now both delegate to it, so they can no longer diverge. Behavior is unchanged — both copies were already identical.

Add tests/test_id_normalization_contract.py: parametrized regression cases for the cited issues, plus hypothesis property tests asserting _make_id(s) == _normalize_id(s) and idempotency. The shared-implementation guard fails if a future change re-forks the normalization.

Claude-Session: https://claude.ai/code/session_01UZZai31gG49fxtUf3onjN3


safishamsi:
Merged to v8 in `8883991` — thanks @danielnguyenfinhub, excellent refactor. I independently verified the safety claim: a differential test (42 single-string inputs + 11 multi-arg forms + a brute force over all strings ≤ length 4) confirmed the new `graphify.ids` reproduces both old recipes **exactly** — zero divergence, no ghost-node regression. (Minor note: the two old recipes weren't byte-identical — extract's had an extra per-part `strip('_.')` — but they're behaviorally equivalent on the single-string contract, and `make_id` preserves the multi-arg join.) On merge I **completed the migration**: `mcp_ingest._make_id` and `symbol_resolution._bash_make_id` were two more live copies of the exact recipe (the latter's 'avoid an import cycle' reason is moot now that `ids.py` is dependency-free), so they now delegate too, and the contract guard asserts all four producers resolve to the shared recipe. Hypothesis property tests run (not skipped). Full suite green (2218 passed).

PR #1379, opened 2026-06-18
FalkorDB backend (server + embedded Lite), replacing NetworkX

galshubeli:
## Replace NetworkX with FalkorDB (server **and** embedded Lite)

This migrates graphify's runtime graph backend from in-memory **NetworkX** to a cache-free **FalkorDB `GraphStore`**: the graph lives in the engine and is never materialized in Python. The existing NetworkX-shaped API (`G.nodes`, `G.edges`, `G.degree`, `G[u][v]`, …) is preserved as a thin facade, so the ~18 call-site modules were minimally touched, but every read is now a scoped/streamed Cypher query.

### Highlights
- **Cache-free `GraphStore`** — nx-compatible views backed by scoped Cypher; nothing is loaded into Python.
- **Server-side JS UDFs** for the algorithms FalkorDB lacks: Louvain, edge-betweenness, simple-cycles (deterministic — sorted iteration, no RNG).
- **FalkorDB Lite** — an embedded, in-process engine (`falkordblite`/`redislite`, no server) selectable via a `falkordb-lite://` URI or `GRAPHIFY_FALKORDB_LITE`; new `lite` extra (Python ≥ 3.12).
- **NetworkX fully removed** — zero runtime `networkx` imports; the test suite builds fixtures via a `MemGraph`-backed `nxcompat` shim.
- Query-path optimizations (truncate-before-fetch rendering, in-engine context filtering, cached hub threshold) and an audit removing whole-graph pulls from the interactive paths.
- CI runs a `falkordb/falkordb` service container; both `test (3.10)` and `test (3.12)` are green.

### Benchmark
Per repo, **mean over the four operations** (`query · path · explain · affected`), **3 repetitions**. Each cell is `mean ± run-to-run standard deviation`. **Bold = better (lower).**

<table>
<thead>
<tr><th rowspan="2">repo</th><th colspan="2">latency (ms)</th><th colspan="2">CPU (ms)</th><th colspan="2">memory (MB)</th></tr>
<tr><th>NetworkX</th><th>FalkorDB</th><th>NetworkX</th><th>FalkorDB</th><th>NetworkX</th><th>FalkorDB</th></tr>
</thead>
<tbody>
<tr><td>FalkorDB-repo (9.2k/26k)</td><td align="right">852 ± 8</td><td align="right"><b>152 ± 1</b></td><td align="right">851 ± 7</td><td align="right"><b>81 ± 1</b></td><td align="right">89 ± 0</td><td align="right"><b>35 ± 1</b></td></tr>
<tr><td>llama-stack (15k/35k)</td><td align="right">708 ± 6</td><td align="right"><b>78 ± 1</b></td><td align="right">709 ± 6</td><td align="right"><b>35 ± 0</b></td><td align="right">109 ± 0</td><td align="right"><b>34 ± 0</b></td></tr>
<tr><td>django (50k/142k)</td><td align="right">438 ± 3</td><td align="right"><b>409 ± 4</b></td><td align="right">438 ± 2</td><td align="right"><b>120 ± 2</b></td><td align="right">311 ± 2</td><td align="right"><b>37 ± 0</b></td></tr>
</tbody>
</table>
**Notes**
- **Memory & CPU are the host (Python) process.** FalkorDB offloads the graph + query compute to the engine, so the host stays flat (~34–37 MB) and CPU-light; NetworkX holds the whole graph in-process (88→311 MB, growing with size).
- These host figures are identical for the embedded **`falkordblite`** backend (Lite ≈ server on latency).
- Averages are lifted by `affected` (NetworkX runs a 0.4–3.2 s in-memory BFS there); the one op NetworkX wins is `query` on a hub node, where its in-RAM traversal beats engine round-trips (verified equivalent work: same seed/depth, ~11.3k nodes reached on django).


Azeem1985:
**Production datapoint at 1.2M nodes: this write-path shape is the one that works — plus one silent-failure edge worth closing**

We run graphify against FalkorDB (server, in Docker) at whole-machine scale in production: 215 repos merged into one graph, 1,211,189 nodes / ~1.15M edges, queried interactively. Getting there forced us through the same design decisions this PR makes, so two measurements and one question:

1. **Shared label + id index before writes is make-or-break, not an optimization.** Our first loader wrote nodes under per-type labels only, with no pre-created index: edge upserts (`MATCH` both endpoints by id, then `MERGE`) ran at ~2 edges/second — at that rate our 1.15M edges were days away. Creating the id index before any writes and stamping one shared label on every node made node batches ~60x faster and edges land in minutes. Your `:Entity` on every node plus `_ensure_schema()` running `CREATE INDEX FOR (n:Entity) ON (n.id)` at store init is exactly the shape we converged on independently.

2. **Host-memory flatness is the property that matters at this scale.** Loading a 1.87GB graph.json through NetworkX peaked at ~5.3GB RSS and was OOM-killed inside a 6GB-capped unit on our 7.6GB box. A cache-free store that never materializes the graph in Python is what makes a million-node graph usable on a small machine at all — the flat ~35MB host memory in your benchmark table matches what we see when the graph lives in the engine.

The question, from the failure mode we actually hit: `_ensure_schema()` swallows all exceptions. If index creation fails — a managed server that forbids it (the same case `_ensure_resultset_size()` already anticipates for `GRAPH.CONFIG`), or an older engine — a fresh bulk load silently degrades to unindexed `MATCH` per `UNWIND` chunk. At 1M+ nodes that presents as a multi-day hang with no error anywhere. Would you consider verifying the index actually exists after `_ensure_schema` (e.g. `CALL db.indexes()`) and emitting a loud warning when it doesn't?

(Also: the PR currently shows as conflicting with v8 after 0.9.50 — as a heavy production FalkorDB user of this tool we'd be glad to see it rebased and moving.)

PR #1380, opened 2026-06-18
fix(llm): adaptive split-and-retry for label_communities on parse failure (#1278)

### LINKED ISSUE #1278: `cluster-only` skips labeling batch on JSON parse error without retry or chunk split (inconsistent with `extract`)
## Version
0.8.38

## Steps to reproduce

Run `cluster-only` on a large graph (e.g. 121k nodes, 135 labeling batches):

```
graphify cluster-only <path>
```

## What happened

During community labeling, one batch failed with a JSON parse error:

```
[graphify label] batch N/M failed:
Expecting property name enclosed in double quotes: line 8 column 40 (char 242)
```

The error is logged but the batch is skipped without retry or chunk split. The final output says "Done" without indicating that some batches failed.

## Expected behavior

`extract` already handles this case correctly - when the LLM returns invalid JSON, it splits the chunk into halves and retries:

```
[graphify] LLM returned invalid JSON, skipping chunk (first 200 chars: ...)
[graphify] chunk of 10 truncated at depth 0, splitting into halves of 5 and 5
```

`cluster-only` labeling should apply the same split-and-retry logic for consistency and to avoid silently incomplete community names.

## Suggested fix

Reuse the existing invalid-JSON handler from the `extract` path in the `label`/`cluster-only` labeling loop.

## Environment

- OS: Windows 11
- graphify: 0.8.38
- Install method: `uv tool install graphifyy[gemini]`
- Graph size: 121,500 nodes, 13,409 communities, 135 labeling batches

CJdev232:
## What

Fixes #1278.

`label_communities` previously logged-and-skipped any batch where the LLM returned malformed JSON or a non-object payload. On large graphs (the reporter's case: 121k nodes / 135 batches), a single failed batch silently lost ~100 community names with no recovery and no warning beyond the per-batch stderr line.

This mirrors the recovery shape `_extract_with_adaptive_retry` already uses for the extract path: when a batch fails to parse, split it at the midpoint and retry each half recursively. Most transient LLM JSON glitches go away at the first or second split (smaller prompt → smaller output → less likely to truncate or mangle).

## Fix

- New helper `_label_batch_with_retry()` in `graphify/llm.py`. Catches `(json.JSONDecodeError, ValueError)` from `_parse_label_response`, splits `batch_cids` and `batch_lines` together at `mid = len(batch_cids) // 2`, recurses on each half with `depth + 1`. Recursion capped at `max_depth=3` (same as extract).
- Non-parse exceptions (network errors, missing API keys, programming bugs) propagate unchanged via bare `raise`. The retry net is intentionally narrow.
- `label_communities` is now a thin wrapper: builds slices, calls the helper, tracks `first_error` and `written` exactly as before. The outer graceful-degradation contract is preserved — when retries exhaust at the base case, the original parse error re-raises so `generate_community_labels` still falls back to placeholders rather than silently returning empty.

## Test

New `tests/test_label_retry.py` adds a unit test with a stateful mock that returns malformed JSON on the first call and clean JSON on subsequent (smaller) calls. Confirms:

- All 4 input communities end up labeled despite the initial failure
- `_call_llm` was called more than once (proving the retry actually happened)

All 15 existing tests in `tests/test_labeling.py` continue to pass — including `test_label_communities_malformed_raises` (the old contract is preserved) and `test_generate_community_labels_degrades_on_error` (graceful fallback still works when retries exhaust).

## Design choices worth flagging

- **Exception types caught** — only `(json.JSONDecodeError, ValueError)` (the two things `_parse_label_response` raises). Network errors / config errors are not retried because they're not the recoverable case the issue describes.
- **`max_depth=3`** for consistency with `_extract_with_adaptive_retry`. With `batch_size=100`, this allows up to 8 leaf batches of ~12-13 communities each before giving up. If real-world cases need deeper recursion, that's a follow-up tuning.
- **Re-raise at base case (not silent return)** so the existing `first_error` tracking in `label_communities` still fires. A partial-success edge case (left half succeeds, right half exhausts) loses the left half's labels — but in exchange, the existing graceful-degradation contract is preserved. The alternative (return empty dict at base case) breaks `test_label_communities_malformed_raises` semantics.

## Notes

- Reproducible context: the original issue was on a 121k-node corpus with `batch_size=100` and Gemini backend. I was unable to reproduce the LLM-side malformed-JSON nondeterminism directly; the test uses a stateful mock to simulate the documented failure mode.
- The fix touches only `llm.py` — no API surface changes.
- Ran the full test suite. One pre-existing failure unrelated to this change: `tests/test_cli_export.py::test_export_wiki_accepts_edges_only_graph_json` fails on unmodified v8 with `KeyError: 'links'`. Looks like the same `edges`-vs-`links` key pattern already fixed in `affected.py` (#1280); the wiki export path appears to have been missed. Happy to follow up in a separate PR if useful.

### DIFF
--- a/graphify/llm.py
+++ b/graphify/llm.py
@@ -2144,6 +2144,75 @@ def _parse_label_response(text: str, labeled_cids: list[int]) -> dict[int, str]:
     return out
 
 
+def _label_batch_with_retry(
+    batch_cids: list[int],
+    batch_lines: list[str],
+    *,
+    backend: str,
+    model: str | None,
+    depth: int = 0,
+    max_depth: int = 3,
+) -> dict[int, str]:
+    """Label a batch of communities, splitting in half and retrying on parse failure.
+
+    Mirrors `_extract_with_adaptive_retry`'s recovery shape for the labeling path
+    (#1278). When the LLM returns malformed JSON or a non-object payload, the
+    batch is split at the midpoint and each half is retried recursively. Recursion
+    is capped at ``max_depth`` to bound cost.
+
+    Returns ``{cid: name}`` for everything that could be labeled. Communities
+    that still couldn't be parsed at the depth/size base case are absent from
+    the result — callers treat them as placeholder. Any non-parse exception
+    (network, missing config, programming bug) propagates unchanged.
+    """
+    prompt = (
+        "You are naming clusters in a knowledge graph. For each community below, "
+        "return a concise 2-5 word plain-language name describing what it is about "
+        "(e.g. \"Order Management\", \"Payment Flow\", \"Auth Middleware\"). "
+        "Respond ONLY with a JSON object mapping the community id (as a string) to "
+        "its name - no prose, no markdown fences.\n\n" + "\n".join(batch_lines)
+    )
+    max_tokens = _resolve_max_tokens(min(64 + 24 * len(batch_cids), 8192))
+    call_kwargs: dict = {"backend": backend, "max_tokens": max_tokens}
+    if model is not None:
+        call_kwargs["model"] = model
+
+    try:
+        text = _call_llm(prompt, **call_kwargs)
+        return _parse_label_response(text, batch_cids)
+    except (json.JSONDecodeError, ValueError) as exc:
+        # YOUR TURN — implement the split-and-retry logic here.
+        #
+        # Two cases to handle:
+        #
+        # 1. BASE CASE: we can't or shouldn't split further.
+        #    - condition: len(batch_cids) <= 1  OR  depth >= max_depth
+        #    - action: print a warning to sys.stderr identifying which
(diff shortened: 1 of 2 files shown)


safishamsi:
Merged to v8 in `ab1e0ec` — thanks @CJdev232. The split-and-retry logic is correct and mirrors `_extract_with_adaptive_retry` cleanly; verified termination (single-element base case re-raises, caller skips that batch), split alignment, and that non-parse errors aren't retried. On merge I removed the leftover `# YOUR TURN —` scaffolding comment block and corrected the helper docstring (it said the base case returns empty, but it re-raises — which is what the caller's per-batch skip + all-fail-reraise contract expects). Full suite green. The partial-success label-loss on a fully-exhausted sub-branch is a known trade-off (it preserves the malformed-raises contract) — fine to revisit separately.

PR #1394, opened 2026-06-19
Remove faulty GitHub Actions Workflow for Python, fix: resolve explain-ambiguity, fix(ruby): resolve compact syntax mixin targets and prevent phantom concern hubs.

FolatheDuckofDuckingburg:
Resolves issue https://github.com/Graphify-Labs/graphify/issues/2302 where Ruby include targets declared with compact syntax (module Foo::Bar) never resolve, and extend ActiveSupport::Concern collides with any local module named Concern. Normalizes compact and nested class/module declarations during AST extraction, and resolves mixin references via lexical lookup.


FolatheDuckofDuckingburg (commented):
@FolatheDuckofDuckingburg approves (bc im not a maintainer)

FolatheDuckofDuckingburg (commented):
:D

safishamsi:
Thanks for this, and for the diagnosis of the Ruby mixin bug in particular.

The branch has diverged from `v8` through several upstream-sync merges, so its computed diff against `v8` is now empty and it cannot be merged as-is. Rather than lose the work, I recovered the Ruby fix: #2302 (compact/nested `module Foo::Bar` mixin resolution + the phantom `Concern` hub) is fixed and shipped in v0.9.32 (`graphifyy==0.9.32`), with you credited in the changelog and release notes. Your approach (canonicalizing compact and nested declarations to one label, keeping the full constant path for mixin callees, and the scoped lexical lookup that stops a qualified external name binding to a bare local module) is what landed, adapted to the current extractor layout with an added guard so nested-declared classes keep resolving as typed receivers.

The explain-ambiguity change was already shipped separately in v0.9.31 (#2305), and the GitHub Actions workflow removal is tracked in your #2264. Closing this one in favour of the shipped fix. Really appreciate the contribution.

PR #1410, opened 2026-06-20
fix(export): never emit punctuation-only Obsidian filenames (e.g. @.md)

### LINKED ISSUE #1409: to_obsidian: all-punctuation labels emit punctuation-only filenames (e.g. @.md) that break downstream slug/indexers (qmd handelize crash)
## Summary

`to_obsidian()`'s `safe_name()` can emit a note whose filename is **only punctuation** (e.g. `@.md`) when a node label is all-punctuation after the unsafe-char strip (e.g. a `@/*` node extracted from a `tsconfig.json` `paths` entry). Such a filename is valid on disk, but any downstream tool that re-slugs the stem on *word* characters reduces it to empty and fails.

Concretely, it crashes `qmd` (the Quick Markdown Search CLI) on `qmd update`, because its `handelize()` reduces `@` → `""` and raises:

```
Error: handelize: path "graphify-out/obsidian/@.md" has no valid filename content
    at handelize (.../@tobilu/qmd/dist/store.js:1493:15)
    at reindexCollection (.../store.js:969:22)
```

Since `qmd update` indexes collections in a batch, this **one generated file aborts the entire run** — every other collection on the machine stops being reindexed too. Found on `graphifyy==0.8.33`.

## Root cause

In `graphify/export.py`, `safe_name()` (used in both the node export and the community export of `to_obsidian`):

```python
def safe_name(label: str) -> str:
    cleaned = re.sub(r'[\\/*?:"<>|#^[\]]', "", label. ...).strip()
    cleaned = re.sub(r"\.(md|mdx|qmd|markdown)$", "", cleaned, flags=re.IGNORECASE)
    return _cap_filename(cleaned) if cleaned else "unnamed"   # <-- "@" is truthy, slips through
```

For label `@/*`: the unsafe-char strip removes `/` and `*`, leaving `@`. `@` is non-empty, so the `else "unnamed"` fallback never fires, and the stem becomes `@` → file `@.md`.

## Proposed fix

Require at least one *word* character; otherwise fall back (so an all-punctuation label can never produce a punctuation-only filename):

```python
    cleaned = re.sub(r"\.(md|mdx|qmd|markdown)$", "", cleaned, flags=re.IGNORECASE)
    # A stem of only punctuation ("@", "*", "#") survives the unsafe-char strip but
    # is empty once a downstream tool re-slugs on word chars (e.g. qmd's handelize()
    # → crash). Require at least one word char; else fall back.
    if not re.search(r"\w", cleaned, flags=re.UNICODE):
        return "unnamed"
    return _cap_filename(cleaned)
```

Applies to **both** occurrences of `safe_name` in `export.py`. The existing dedup logic already handles multiple `unnamed` collisions (`unnamed_1`, …).

## Repro

A node whose label is all punctuation after the unsafe-char strip (e.g. `@/*` from a `tsconfig.json` `paths` key) → `graphify-out/obsidian/@.md`. Other `@*.md` files with a word char (`@types/node` → `@typesnode.md`) are fine; only the bare-punctuation case breaks.

Happy to send a PR if the fix looks right.

Mylock51:
Fixes #1409.

### Problem
`to_obsidian()`'s `safe_name()` can produce a **punctuation-only filename** (e.g. `@.md`) when a node label is all-punctuation after the unsafe-char strip — e.g. a `@/*` node from a `tsconfig.json` `paths` key leaves a bare `@`, which is truthy and so slips past the `else "unnamed"` fallback.

The file is valid on disk but **empty once a downstream tool re-slugs the stem on word chars**. In particular it crashes `qmd update` (handelize() reduces `@` → `""` and raises), and because qmd indexes collections in a batch, that one generated file aborts reindexing for *every* collection on the machine.

### Fix
Require at least one `\w` char in the stem; otherwise fall back to `"unnamed"`. Applied to **both** `safe_name` occurrences (node export + community export). The existing dedup logic already disambiguates multiple `unnamed` collisions.

Behavior is unchanged for any label containing a word char (`@types/node` → `@typesnode.md` as before); only the pathological all-punctuation case changes.

### DIFF
--- a/graphify/export.py
+++ b/graphify/export.py
@@ -866,7 +866,14 @@ def to_obsidian(
         cleaned = re.sub(r'[\\/*?:"<>|#^[\]]', "", label.replace("\r\n", " ").replace("\r", " ").replace("\n", " ")).strip()
         # Strip trailing .md/.mdx/.markdown so "CLAUDE.md" doesn't become "CLAUDE.md.md"
         cleaned = re.sub(r"\.(md|mdx|qmd|markdown)$", "", cleaned, flags=re.IGNORECASE)
-        return _cap_filename(cleaned) if cleaned else "unnamed"
+        # A stem of only punctuation (e.g. "@", "*", "#") survives the unsafe-char
+        # strip above but is empty once a downstream tool re-slugs on word chars
+        # (e.g. qmd's handelize() reduces "@" -> "" and raises, aborting the whole
+        # `qmd update`). Require at least one word char; else fall back so we never
+        # emit a "@.md"-style filename. (#1409)
+        if not re.search(r"\w", cleaned, flags=re.UNICODE):
+            return "unnamed"
+        return _cap_filename(cleaned)
 
     node_filename: dict[str, str] = {}
     seen_names: dict[str, int] = {}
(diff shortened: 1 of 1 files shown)


safishamsi:
Merged into `v8` as 30cc4c0 (cherry-picked onto current v8), with a regression test added in 739230e. Verified the fix is multilingual-safe (`\w` is unicode-aware, so `café`/`日本語` labels are preserved) and the tests fail against the pre-fix `safe_name()`. Full suite green on 3.10 + 3.12; smoke confirms a `@/*` node now writes `unnamed.md`, never `@.md`. Thanks @Mylock51!

PR #1411, opened 2026-06-21
Add CUDA (.cu/.cuh) support via the C++ extractor

raylei50653:
## What

Adds first-class support for CUDA source files (`.cu`, `.cuh`).

CUDA is a C++ superset, so these files parse cleanly with `tree-sitter-cpp` (already a project dependency — no new packages). Wiring is two registrations:

- **`detect.py`** — add `.cu`/`.cuh` to `CODE_EXTENSIONS`, so they're detected and watched. `watch.py`'s `_WATCHED_EXTENSIONS` derives from `CODE_EXTENSIONS`, so file watching is covered automatically.
- **`extract.py`** — route `.cu`/`.cuh` to `extract_cpp` in `_DISPATCH`. This also makes `collect_files()` pick them up, since `_EXTENSIONS = set(_DISPATCH.keys())`.

Also updates the README extension table and the CHANGELOG.

## Tests

- New fixture `tests/fixtures/sample.cu` exercising a `__global__` kernel, a `__device__` function, a host function, a `struct`, includes, and host-to-host call edges.
- New CUDA cases in `tests/test_languages.py` (mirroring the existing C++ block) asserting: no extraction error, kernel/device function extraction, struct extraction, include edges, and host call edges.

All CUDA + existing C++ language tests pass (`414 passed` across `test_detect.py`, `test_languages.py`, `test_watch.py`).

The `__global__`/`__device__` qualifiers and `<<<grid, block>>>` kernel-launch syntax don't break the C++ grammar; host functions, structs, and includes are extracted as expected.

## Known limitation

Because the C++ grammar treats the `__global__` / `__device__` qualifiers as identifiers, they surface as their own graph nodes (one `__global__` node and one `__device__` node, globally deduped). This is cosmetic and low-impact. I left it unfiltered deliberately to keep the diff to pure registration and avoid touching the shared C++ extraction path — happy to add a qualifier stop-list in a follow-up if you'd prefer them suppressed.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@ Full release notes with details on each version: [GitHub Releases](https://githu
 
 ## Unreleased
 
+- Feat: CUDA (`.cu`/`.cuh`) source files are now extracted. CUDA is a C++ superset, so these files route through the existing C++ (`tree-sitter-cpp`) extractor — no new grammar dependency. `__global__`/`__device__` kernels, host functions, structs and `#include`s are captured, host call edges are inferred, and `<<<grid, block>>>` kernel-launch syntax parses without error. Detection and file-watching follow automatically since both derive their extension sets from the dispatch table / `CODE_EXTENSIONS` (#1411).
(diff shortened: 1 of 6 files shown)


safishamsi:
Merged into `v8` as dbce453 (cherry-picked; resolved a CHANGELOG conflict against the 0.8.45 release). CUDA routes through the C++ extractor with no new dep; the 5 added tests pass on 3.10 + 3.12. Thanks @raylei50653!

Issue #1413, opened 2026-06-21
bug: opencode plugin reminder string has unescaped backticks → bash command substitution silently runs 'graphify query "<question>"' on every search


WSHAPER:
## Summary

The opencode plugin template shipped by `graphify install --platform opencode` (v0.8.39) embeds the reminder string with backticks around `graphify query "<question>"`. Because the plugin prepends `echo "..." && <user-cmd>` to the user's bash command, those backticks are interpreted by bash as **command substitution** — every grep/rg/find command silently runs `graphify query "<question>"` and substitutes its output (`No matching nodes found.`) into the middle of the reminder text.

This (a) corrupts every search tool's output with graphify noise, and (b) actually invokes graphify on every search, costing a process spawn + a graph load each time.

## Environment

- **OS:** Linux
- **graphify version:** `0.8.39` (uv tool install)
- **opencode:** latest, global config at `~/.config/opencode/`
- **Shell:** bash

## Reproduction

1. Install: `graphify install --platform opencode` (then ensure the plugin is at `~/.config/opencode/.opencode/plugins/graphify.js` — see companion issue about the install path bug).
2. Build a graph: in any repo, run `/graphify .` so `graphify-out/graph.json` exists.
3. Restart opencode.
4. Ask the agent anything that triggers a bash search, e.g. "find all uses of prisma in backend".

## Actual behaviour

The plugin emits this exact line for the bash hook:

```js
// .opencode/plugins/graphify.js, line 16
output.args.command =
  'echo "[graphify] knowledge graph at graphify-out/. For focused questions, run \`graphify query \"<question>\"\` (scoped subgraph, usually much smaller than GRAPH_REPORT.md) instead of grepping raw files. Read GRAPH_REPORT.md only for broad architecture context." && ' +
  output.args.command;
```

The literal text `run \`graphify query "<question>"\`` contains backticks. When bash evaluates `echo "..."`, it spawns a subshell for the backticks, runs `graphify query "<question>"`, and substitutes stdout in place.

### Observed tool output (real session)

The user runs a `gh` command (contains "find"-like substrings or grep-like patterns):

```
[graphify] knowledge graph at graphify-out/. For focused questions, run No matching nodes found. (scoped subgraph, usually much smaller than GRAPH_REPORT.md) instead of grepping raw files. Read GRAPH_REPORT.md only for broad architecture context.
```

[...]

Issue #1040, opened 2026-05-27
`--project` flag creates non-discoverable skill path for OpenCode platform


storybite:
## Summary

When running `graphify install --platform opencode --project`, the `SKILL.md` is written to `./.config/opencode/skills/graphify/SKILL.md`, which is **not** a path that OpenCode discovers for project-scoped skills. As a result, OpenCode silently ignores the installed skill.

The `AGENTS.md`, `.opencode/plugins/graphify.js`, and `.opencode/opencode.json` files are written correctly, so the always-on hook behavior still works — but the skill itself is invisible to OpenCode's skill discovery.

## Environment

- **OS:** Windows 11
- **Shell:** PowerShell
- **graphify version:** 0.8.20 (installed via `uv tool install graphifyy`)
- **Project path:** `D:\workspace\vibe-java\loan-eval`

## Reproduction

```powershell
cd D:\workspace\vibe-java\loan-eval
graphify install --platform opencode --project
```

### Actual output

```
skill installed  ->  .config\opencode\skills\graphify\SKILL.md
graphify section written to D:\workspace\vibe-java\loan-eval\AGENTS.md
.opencode\plugins\graphify.js  ->  tool.execute.before hook written
.opencode\opencode.json  ->  plugin registered
Opencode will now check the knowledge graph before answering
codebase questions and rebuild it after code changes.
Project-scoped install. Add to version control:
  git add .config/ AGENTS.md .opencode/
```

Notice the first line: `.config\opencode\skills\graphify\SKILL.md` — a `.config/` directory is created **inside the project root**.

### Expected output

```
skill installed  ->  .opencode\skills\graphify\SKILL.md
```

The `SKILL.md` should live under the project's `.opencode/skills/` directory, alongside the plugin that's already being placed there correctly.

## Why this is a bug

According to the [OpenCode official documentation on Agent Skills](https://opencode.ai/docs/skills/), OpenCode searches **exactly these six locations** for skills:

- Project config: `.opencode/skills/<name>/SKILL.md`
- Global config: `~/.config/opencode/skills/<name>/SKILL.md`
- Project Claude-compatible: `.claude/skills/<name>/SKILL.md`
- Global Claude-compatible: `~/.claude/skills/<name>/SKILL.md`
- Project agent-compatible: `.agents/skills/<name>/SKILL.md`
- Global agent-compatible: `~/.agents/skills/<name>/SKILL.md`

[...]

PR #1414, opened 2026-06-21
fix(opencode): write global-install plugin to ~/.config/opencode/ instead of cwd

### LINKED ISSUE #1412: bug: 'graphify install --platform opencode' (global, no --project) writes plugin to ~/.opencode/ instead of ~/.config/opencode/.opencode/plugins/
## Summary

Running `graphify install --platform opencode` **without** `--project` (i.e. the documented global-install mode) writes the plugin to `~/.opencode/plugins/graphify.js` and creates a new `~/.opencode/opencode.json` next to it. Neither file is loaded by opencode — opencode's global config lives at `~/.config/opencode/opencode.json`, and plugins referenced as `.opencode/plugins/<name>.js` in that config are resolved relative to **that config's directory** (`~/.config/opencode/.opencode/plugins/`). The installer writes to the wrong tree.

The installer log claims success (`tool.execute.before hook written`, `plugin registered`), but the plugin silently never loads. The `~/.opencode/` directory opencode creates is the npm-style binary install dir (`~/.opencode/bin/opencode`, `~/.opencode/node_modules/`), not a config dir.

This is the **global-mode** sibling of #1040 (which was the `--project` variant of the same class of bug).

## Environment

- **OS:** Linux (Ubuntu / NixOS)
- **Shell:** bash
- **graphify version:** `0.8.39` (installed via `uv tool install graphifyy`)
- **opencode:** config at `~/.config/opencode/opencode.json` (the documented global config location per https://opencode.ai/config.json)
- **Run from:** `~` (home directory)

## Reproduction

```bash
cd ~
rm -rf .opencode/plugins/graphify.js .opencode/opencode.json \
       .config/opencode/.opencode/plugins/graphify.js
graphify install --platform opencode
```

### Actual output

```
  references       ->  /home/ikv/.config/opencode/skills/graphify/references
  skill installed  ->  /home/ikv/.config/opencode/skills/graphify/SKILL.md
  .opencode/plugins/graphify.js  ->  tool.execute.before hook written
  .opencode/opencode.json  ->  plugin registered
```

### Actual filesystem state after install

```
/home/ikv/.opencode/                                 ← npm-style binary install dir
├── bin/opencode
├── node_modules/
├── package.json
├── opencode.json                                   ← installer CREATED this (orphan)
│   (contents: {"$schema":"…","plugin":[".opencode/plugins/graphify.js"]})
└── plugins/
    └── graphify.js                                 ← installer wrote plugin HERE (orphan)

/home/ikv/.config/opencode/                         ← the REAL global config dir
├── opencode.json                                   ← user's actual config (untouched)
│   (already has "plugin": [".opencode/plugins/graphify.js"] from a prior install)
├── skills/graphify/SKILL.md                        ← correctly placed (skill DOES load)
└── .opencode/plugins/                              ← EMPTY — no graphify.js here
```

### Observed symptom

The skill loads correctly, but the plugin never fires its `tool.execute.before` hook, because opencode resolves `.opencode/plugins/graphify.js` against `~/.config/opencode/` (where the loaded config lives), not against `~/.opencode/`.

Running `graphify query "<question>"` confirms the skill is registered, but bash search commands (`grep`, `rg`, `find`) produce no reminder — the hook is dead.

### Workaround

After install, manually move the plugin to where opencode actually loads it from:

```bash
mkdir -p ~/.config/opencode/.opencode/plugins
cp ~/.opencode/plugins/graphify.js ~/.config/opencode/.opencode/plugins/graphify.js
# Optional: delete the orphans
rm -rf ~/.opencode/plugins ~/.opencode/opencode.json
```

## Expected behaviour

The installer should resolve the global opencode config dir via (in order of preference):

1. `$XDG_CONFIG_HOME/opencode` if `XDG_CONFIG_HOME` is set
2. `~/.config/opencode/` (the documented default — see https://opencode.ai/config.json and the `customize-opencode` skill shipped with opencode itself, which explicitly states: *"Global config — `~/.config/opencode/opencode.json` (NOT `~/.opencode/`)"*)

…and write the plugin to `<config_dir>/.opencode/plugins/graphify.js`, registering it in `<config_dir>/opencode.json` only if not already present.

`~/.opencode/` is opencode's binary/cache dir, not a config dir, and should not be touched by the installer.

## Related

- #1040 — `--project` variant of the same bug class (skill written to non-discoverable path)
- #378 — earlier "installation reports success but plugin/json/command missing" report
- The opencode docs at https://opencode.ai/config.json and the built-in `customize-opencode` skill both clearly distinguish `~/.config/opencode/` (config) from `~/.opencode/` (runtime/binary).

## Suggested fix

In the installer code path for `--platform opencode` global mode, replace the destination resolution (pseudocode):

```python
# Before (buggy)
config_dir = home / ".opencode"

# After (correct)
xdg = os.environ.get("XDG_CONFIG_HOME")
config_dir = Path(xdg) / "opencode" if xdg else home / ".config" / "opencode"
```

…and write `graphify.js` to `config_dir / ".opencode" / "plugins" / "graphify.js"`. The skill write path (`config_dir / "skills" / "graphify" / ...`) already does this correctly — only the plugin + `opencode.json` writes are wrong.

WSHAPER:
Fixes #1412.

## Summary

Running `graphify install --platform opencode` **without** `--project` (the documented global-install mode) wrote the plugin to `<cwd>/.opencode/plugins/graphify.js` and created a stray `<cwd>/.opencode/opencode.json`. Neither file is loaded by opencode — opencode resolves `.opencode/plugins/<name>.js` entries **relative to the declaring config's directory**, and the global config lives at `~/.config/opencode/opencode.json`, not cwd.

The installer log claimed success (`tool.execute.before hook written`, `plugin registered`), but the plugin silently never loaded. This is the global-mode sibling of #1040 (which was the `--project` variant of the same class of bug).

## Root cause

`graphify/__main__.py` line 697 (v0.8.44):

```python
if platform == "opencode":
    _install_opencode_plugin(project_dir if project else Path("."))
```

For global installs (`project=False`), this resolved the plugin destination against `Path(".")` — the current working directory. Run from `~`, the plugin landed at `~/.opencode/plugins/graphify.js`. Run from any other cwd, it landed wherever the user happened to be.

Meanwhile, the **skill** install path (line 148-151) already handled both modes correctly:

```python
if platform_name == "opencode":
    if project:
        return (project_dir or Path(".")) / ".opencode" / "skills" / "graphify" / "SKILL.md"
    return Path.home() / ".config" / "opencode" / "skills" / "graphify" / "SKILL.md"
```

The plugin install path simply didn't have the same branching — it always used `project_dir` regardless of mode.

## Fix

Thread `project: bool = True` into `_install_opencode_plugin` / `_uninstall_opencode_plugin`. When `project=False`, resolve base dir via `Path.home() / ".config" / "opencode"` (matching the existing skill-path logic, monkey-patchable in tests via `Path.home`). When `project=True` (default, used by `_agents_install` etc.), behaviour is unchanged.

Only the global-mode caller (`install()` at line 697) passes the new flag; the five other callers (`_agents_install`, `_agents_uninstall` × 3, `uninstall_all`) are project-scoped by design and use the default.

Also: the install/uninstall messages now print the resolved absolute path so users can see where the plugin actually landed (previously they saw the relative `.opencode/plugins/graphify.js` which gave no clue about which tree it was written to).

## Verification

Three new regression tests:

1. `test_install_opencode_global_writes_plugin_to_config_dir` — after global install, plugin exists at `~/.config/opencode/.opencode/plugins/graphify.js` and there is **no** stray `<cwd>/.opencode/` tree.
2. `test_install_opencode_global_registers_plugin_in_config` — registration goes into `~/.config/opencode/.opencode/opencode.json`, not a stray cwd-relative file.
3. `test_uninstall_opencode_global_removes_plugin` — global uninstall (`project=False`) removes the plugin from the correct location.

All 12 opencode install tests pass; no regressions in the broader install/roundtrip/upgrade/reference suites (156 tests pass).

## Workaround for users on already-released versions

```bash
mkdir -p ~/.config/opencode/.opencode/plugins
cp ~/.opencode/plugins/graphify.js ~/.config/opencode/.opencode/plugins/graphify.js
rm -rf ~/.opencode/plugins ~/.opencode/opencode.json
```

## Related

- #1040 — `--project` variant of the same bug class (project-mode skill path was wrong; this PR fixes the corresponding global-mode plugin path)
- #1413 — backtick command-substitution bug in the plugin template (filed same day; companion PR)
- The opencode docs at https://opencode.ai/config.json and the built-in `customize-opencode` skill both clearly distinguish `~/.config/opencode/` (config) from `~/.opencode/` (runtime/binary).

### DIFF
--- a/graphify/__main__.py
+++ b/graphify/__main__.py
@@ -694,7 +694,7 @@ def install(platform: str = "claude", *, project: bool = False, project_dir: Pat
             print(f"  CODEBUDDY.md     ->  created at {codebuddy_md}")
 
     if platform == "opencode":
-        _install_opencode_plugin(project_dir if project else Path("."))
+        _install_opencode_plugin(project_dir or Path("."), project=project)
 
     # Refresh version stamps in all other previously-installed skill dirs so
     # stale-version warnings don't fire for platforms not explicitly re-installed.
@@ -1404,14 +1404,38 @@ _OPENCODE_PLUGIN_PATH = Path(".opencode") / "plugins" / "graphify.js"
 _OPENCODE_CONFIG_PATH = Path(".opencode") / "opencode.json"
 
 
-def _install_opencode_plugin(project_dir: Path) -> None:
-    """Write graphify.js plugin and register it in opencode.json."""
-    plugin_file = project_dir / _OPENCODE_PLUGIN_PATH
+def _install_opencode_plugin(project_dir: Path, *, project: bool = True) -> None:
+    """Write graphify.js plugin and register it in opencode.json.
+
+    When ``project`` is True (default), the plugin lands at
+    ``<project_dir>/.opencode/plugins/graphify.js`` and is registered in
+    ``<project_dir>/.opencode/opencode.json`` — opencode's project-scoped
+    config location (one of three valid project-config paths opencode walks
+    up from cwd to find).
+
+    When ``project`` is False (global install), the plugin lands at
+    ``~/.config/opencode/.opencode/plugins/graphify.js`` (opencode resolves
+    a ``.opencode/plugins/<name>.js`` plugin entry relative to the
+    declaring config's directory) and is registered in
+    ``~/.config/opencode/opencode.json`` — opencode's ONLY global config
+    location. Note: the previous behaviour wrote both files into
+    ``<cwd>/.opencode/`` which only worked if cwd was the global config
+    dir. See issue #1412.
+    """
+    base = project_dir if project else (Path.home() / ".config" / "opencode")
+    plugin_file = base / _OPENCODE_PLUGIN_PATH
+    # Project mode writes the config inside .opencode/ (one of three valid
+    # project-config locations). Global mode must write to the user's real
+    # global config at <base>/opencode.json — <base>/.opencode/opencode.json
+    # is NOT a config location opencode loads.
+    if project:
+        config_file = base / _OPENCODE_CONFIG_PATH
+    else:
+        config_file = base / "opencode.json"
     plugin_file.parent.mkdir(parents=True, exist_ok=True)
     plugin_file.write_text(_OPENCODE_PLUGIN_JS, encoding="utf-8")
-    print(f"  {_OPENCODE_PLUGIN_PATH}  ->  tool.execute.before hook written")
+    print(f"  {plugin_file.resolve()}  ->  tool.execute.before hook written")
 
-    config_file = project_dir / _OPENCODE_CONFIG_PATH
     if config_file.exists():
         try:
             config = json.loads(config_file.read_text(encoding="utf-8"))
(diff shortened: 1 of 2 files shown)


safishamsi (changes requested):
The helper rewrite is sound, but this doesn't fix the reported bug for the **documented** command. There are two opencode install entry points:

- `install()` (`graphify install --platform opencode`) — patched here ✅
- `_agents_install()` (`graphify opencode install`, dispatch ~`__main__.py:2511`) — **not patched**. It still calls `_install_opencode_plugin(project_dir or Path("."))` with the default `project=True`, so `graphify opencode install` (no `--project`) still writes to `<cwd>/.opencode/` — exactly #1412.

Please thread `project=` through the `_agents_install` call site, and the uninstall sites (~1710/1719/1738/1873) which still default to project-scope (a global uninstall won't clean `~/.config/opencode/`). The added tests drive `install()` only — please add one driving `_agents_install` global so the primary command path is covered.

Note: #1415 edits the same region — whichever merges second will need a rebase.

Issue #1403, opened 2026-06-19
bug: `graphify install --platform hermes` installs skill to wrong directory on Windows


SHJordan:
## Bug Description

On Windows, running `graphify install --platform hermes` copies the skill to `~/.hermes/skills/` instead of `~/AppData/Local/hermes/skills/`. Hermes Agent on Windows looks for skills in `~/AppData/Local/hermes/skills/`, so the installed skill is never discovered.

## Steps to Reproduce

1. Install graphify: `uv tool install graphifyy`
2. Run: `graphify install --platform hermes`
3. Check where the skill was placed: `ls ~/.hermes/skills/graphify/` — it exists here
4. Check where Hermes expects it: `ls ~/AppData/Local/hermes/skills/graphify/` — not found

## Expected Behavior

`graphify install --platform hermes` should detect the Windows environment and install the skill to `~/AppData/Local/hermes/skills/` (which is `%LOCALAPPDATA%\hermes\skills\` in Windows terms).

## Actual Behavior

The skill is installed to `~/.hermes/skills/`, which is the Linux/macOS path. On Windows (using Git Bash / MSYS2), `~` resolves to `C:\Users\<username>`, so the skill ends up at `C:\Users\<username>\.hermes\skills\graphify\` — a path that Hermes Agent never scans.

## Environment

- OS: Windows 10
- Shell: Git Bash (MSYS2)
- Python: 3.11 (via uv)
- graphify version: 0.8.44
- Hermes Agent: latest

## Workaround

Manually copy the skill to the correct location:

```bash
cp -r ~/.hermes/skills/graphify ~/AppData/Local/hermes/skills/graphify
```

## Additional Context

The `~/.hermes/` directory does exist on Windows (Hermes creates it for config), but the **skills** directory specifically lives under `~/AppData/Local/hermes/` (the `%LOCALAPPDATA%` equivalent). The install script should check for the Hermes skills directory at:

- Linux/macOS: `~/.hermes/skills/`
- Windows: `~/AppData/Local/hermes/skills/` (or `%LOCALAPPDATA%\hermes\skills\`)

A simple platform check like `if [[ "$OSTYPE" == "msys" || "$OSTYPE" == "cygwin" || "$OSTYPE" == "win32" ]]` would handle this.

safishamsi:
[...]

PR #1416, opened 2026-06-21
fix: hermes install missing SKILL.md + python3 fallback on Windows

OmniaZ1:
### Bug 1: `graphify hermes install` skips SKILL.md installation

`graphify hermes install` (and `graphify codex install`, `graphify claw install`, etc.) only calls `_agents_install()` which writes the AGENTS.md section, but never calls `_copy_skill_file()`. This leaves the user with no SKILL.md and no `references/` sidecar — the skill is invisible to the host agent.

The generic `graphify install --platform hermes` works correctly because it goes through `install()` → `_copy_skill_file()`. The per-platform shortcuts bypass this.

**Fix:** Add `_copy_skill_file(cmd)` before `_agents_install()` and `_remove_skill_file(cmd)` before `_agents_uninstall()` in the per-platform CLI handler (`__main__.py` L2508-2514).

This affects all platforms in the group: aider, codex, opencode, claw, droid, trae, trae-cn, hermes.

### Bug 2: `python3` fallback fails on Windows

The skill-claw.md (used by Hermes, claw, aider, and others) falls back to `python3` when uv/pipx/shebang detection fails. On Windows, `python3` does not exist — the executable is `python`.

**Fix:** Replace bare `python3` fallbacks with a two-probe chain:
```bash
if command -v python3 >/dev/null 2>&1; then PYTHON=python3
elif command -v python >/dev/null 2>&1; then PYTHON=python
fi
```

Applied in two locations in skill-claw.md:
- L84: Main detection chain fallback
- L599/L601: Re-detection chain when `.graphify_python` is missing

### Backwards compatibility

Both fixes are fully backwards-compatible:
- Unix users see no change (python3 probe succeeds first)
- The skill install path now works identically to `graphify install --platform`

### Tested on

- Windows 11, Git Bash (MSYS), uv tool install, Python 3.11
- Verified: `graphify hermes install` now installs SKILL.md + references/ + AGENTS.md
- Verified: all 5 python3 references in installed SKILL.md are protected by `command -v` checks

### DIFF
--- a/graphify/__main__.py
+++ b/graphify/__main__.py
@@ -2506,11 +2506,13 @@ def main() -> None:
             if "--project" in sys.argv[3:]:
                 _project_install(cmd, Path("."))
             else:
+                _copy_skill_file(cmd)
                 _agents_install(Path("."), cmd)
         elif subcmd == "uninstall":
             if "--project" in sys.argv[3:]:
                 _project_uninstall(cmd, Path("."))
             else:
+                _remove_skill_file(cmd)
                 _agents_uninstall(Path("."), platform=cmd)
                 if cmd == "codex":
                     _uninstall_codex_hook(Path("."))
--- a/graphify/skill-claw.md
+++ b/graphify/skill-claw.md
@@ -81,8 +81,12 @@ if [ -z "$PYTHON" ] && [ -n "$GRAPHIFY_BIN" ]; then
         *) "$_SHEBANG" -c "import graphify" 2>/dev/null && PYTHON="$_SHEBANG" ;;
     esac
 fi
-# 3. Fall back to python3
-if [ -z "$PYTHON" ]; then PYTHON="python3"; fi
+# 3. Fall back to python3 (Unix) or python (Windows)
+if [ -z "$PYTHON" ]; then
+    if command -v python3 >/dev/null 2>&1; then PYTHON="python3"
+    elif command -v python >/dev/null 2>&1; then PYTHON="python"
+    fi
+fi
 if ! "$PYTHON" -c "import graphify" 2>/dev/null; then
     if command -v uv >/dev/null 2>&1; then
         uv tool install --upgrade graphifyy -q 2>&1 | tail -3
@@ -596,9 +600,15 @@ if [ ! -f graphify-out/.graphify_python ]; then
     GRAPHIFY_BIN=$(which graphify 2>/dev/null)
     if [ -n "$GRAPHIFY_BIN" ]; then
         PYTHON=$(head -1 "$GRAPHIFY_BIN" | tr -d '#!')
-        case "$PYTHON" in *[!a-zA-Z0-9/_.-]*) PYTHON="python3" ;; esac
+        case "$PYTHON" in *[!a-zA-Z0-9/_.-]*)
+            if command -v python3 >/dev/null 2>&1; then PYTHON="python3"
+            elif command -v python >/dev/null 2>&1; then PYTHON="python"
+            fi ;;
+        esac
     else
-        PYTHON="python3"
+        if command -v python3 >/dev/null 2>&1; then PYTHON="python3"
+        elif command -v python >/dev/null 2>&1; then PYTHON="python"
+        fi
     fi
     mkdir -p graphify-out
     "$PYTHON" -c "import sys; open('graphify-out/.graphify_python', 'w', encoding='utf-8').write(sys.executable)"


safishamsi (changes requested):
Thanks for catching this — the diagnosis is right: the AGENTS.md-family user-scope dispatch (`hermes`/`claw`/`droid`/`trae`/…) never called `_copy_skill_file`, so `SKILL.md` was never written (#1403). Two blockers before merge:

1. **Edits a generated file → fails CI.** `graphify/skill-claw.md` is generated by `tools/skillgen` from fragments, and the `skillgen-check` CI job byte-diffs the render against the committed files + `expected/`, so a hand-edit fails. The python3 fallback needs to go in the **fragment** (`tools/skillgen/fragments/shell/posix.md`, which also feeds `core/aider.md`/`core/devin.md`), then `python -m tools.skillgen` to regenerate **all** skill files and `--bless` the `expected/` snapshots. As-is the fallback only landed in 1 of the 13 generated skills.
2. **No test.** This is exactly what `tests/test_install*.py` covers — please add a regression test asserting `SKILL.md` exists after a user-scope `hermes install` (ideally claw/droid/trae too, since they share the dispatch).

The `__main__.py` fix itself (adding `_copy_skill_file`/`_remove_skill_file` to the dispatch, matching the `_amp_install` pattern) is correct — it just needs to go through skillgen + a test.

saman6384:
Still reproduces on the current release, ~3 months and many releases after this was first filed.

**Version:** `graphifyy` **0.9.64** (`uv tool install`), Linux, `graphify/install.py` at commit `b9cd957`.

```console
$ cd /tmp && graphify hermes install
graphify section written to /tmp/AGENTS.md

Hermes will now check the knowledge graph before answering
codebase questions and rebuild it after code changes.

$ ls ~/.hermes/skills/graphify
ls: cannot access '/home/<user>/.hermes/skills/graphify': No such file or directory
```

Exit code is `0` and nothing indicates a problem, so Hermes ends up with the AGENTS.md text that says to run `graphify query` but without the skill that documents the query flags and ambiguity handling.

### Root cause is unchanged at 0.9.64

The non-project branch (`install.py:2343-2349`) still calls `_agents_install` alone:

```python
elif cmd in ("aider", "codex", "opencode", "claw", "droid", "trae", "trae-cn", "hermes"):
    subcmd = sys.argv[2] if len(sys.argv) > 2 else ""
    if subcmd == "install":
        if "--project" in sys.argv[3:]:
            _project_install(cmd, Path("."))
        else:
            _agents_install(Path("."), cmd)      # <-- _copy_skill_file never called
```

so on this path the following stay unreachable for `hermes`:

- `_PLATFORM_CONFIG["hermes"]["skill_dst"]` (`install.py:458-462`) — `skill_file: skill-claw.md`, `skill_refs: claw`, `skill_dst: .hermes/skills/graphify/SKILL.md`
- the Hermes branch of `_skill_path` (`install.py:95-102`), including the Windows `%LOCALAPPDATA%\hermes\skills` handling added for #1403

`_project_install` (`install.py:1643-1645`) *does* call `_copy_skill_file(..., project=True)` for the same platform tuple, which is why `--project` works and the bare command does not.

The uninstall side is symmetric — `install.py:2350-2354` calls `_agents_uninstall` without `_remove_skill_file` — so `graphify hermes uninstall` also will not remove a skill the user installed by hand.

### Workaround verified on 0.9.64

The copy machinery itself is fine; only the dispatch branch is broken. Calling it directly installs the intended bundle (SKILL.md from `skill-claw.md` + the 8-file `claw` references sidecar + the `.graphify_version` stamp):

```bash
~/.local/share/uv/tools/graphifyy/bin/python3 -c \
  "from graphify.install import _copy_skill_file; _copy_skill_file('hermes')"
#   references       ->  ~/.hermes/skills/graphify/references
#   skill installed  ->  ~/.hermes/skills/graphify/SKILL.md
```

Confirmed loadable by Hermes afterwards (`~/.hermes/skills/graphify/SKILL.md` + `references/`, correct `name`/`description` frontmatter).

Also confirmed on 0.9.64, agreeing with the note above: `graphify install --platform hermes` **does** write the skill to `~/.hermes/skills/graphify/` correctly and writes **no** stray `AGENTS.md`. It additionally preserves a locally-modified `SKILL.md` as `SKILL.md.bak` before overwriting. So that documented path is a dependable fallback until this is fixed.

Scope note: this comment is only about the missing skill. The cwd-`AGENTS.md` half of the behavior is tracked in #2164, with the trust-boundary angle in #1846 and the help-text wording in #1849.

Issue #1390, opened 2026-06-19
Feature request: `--max-concurrency` and `--batch-size` for `cluster-only` / `label`


sub4biz:
### Environment

graphify v0.8.42 (graphifyy package)

### Problem

Community labeling (`cluster-only`, `label`) runs **sequentially** with a fixed batch size of 100 communities per LLM call. For large graphs this creates an unnecessary bottleneck: a graph with 19 094 communities requires **191 sequential API calls**, even when the configured backend supports thousands of concurrent requests (e.g. DeepSeek V4-Flash has a concurrency limit of 2500).

By contrast, `graphify extract` already exposes `--max-concurrency` and uses a `ThreadPoolExecutor` to parallelise semantic extraction chunks. The labeling path has no equivalent.

### Current behavior

`llm.py::label_communities` iterates batches sequentially:

```python
# llm.py
_LABEL_BATCH_SIZE = 100        # hardcoded; not exposed to CLI

for batch_idx in range(n_batches):
    ...
    text = _call_backend(...)  # one request at a time
    labels.update(parsed)
```

`cluster-only` / `label` CLI commands accept `--backend` and `--model` but no concurrency or batch-size flags.

### Proposed solution

1. **`--max-concurrency N`** — parallelise batch LLM calls via `ThreadPoolExecutor`, mirroring the implementation already used in `extract_corpus_parallel`. Default: 4 (safe for all backends).

2. **`--batch-size N`** — expose `_LABEL_BATCH_SIZE` as a CLI flag so users can tune prompt size vs. request count. Default: 100 (current behavior preserved).

Example usage:

```
graphify cluster-only ./work --backend=deepseek --model=deepseek-v4-flash --max-concurrency=32 --batch-size=200
```

With `--max-concurrency=32` and `--batch-size=200`: 96 batches running 32 at a time → ~3 rounds instead of 191 sequential calls.

### Implementation hint

`label_communities` already accepts `batch_size` as a keyword argument — it just isn't wired up to the CLI. Adding `max_concurrency` and replacing the `for` loop with a `ThreadPoolExecutor` fan-out (same pattern as `_extract_corpus_parallel`) should be a small, self-contained change.

safishamsi:
[...]

PR #1421, opened 2026-06-21
feat: add --missing-only flag to label command to skip already named communities

matiasduartee:
I am working on a massive codebase that has over 7,500 nodes. When running `graphify label .`, the LLM backend often struggles to label all 1,000+ communities perfectly in one go due to API rate limits or parsing errors, successfully labeling some while leaving others as `Community N`. The core issue is that running `graphify label .` again completely wipes out the perfectly valid community names generated in the previous run, forcing the LLM to start from scratch. Sometimes, this leads to communities that were correctly named in the first run reverting back to `Community N` in the second run, creating an endless loop of wasted tokens.

To solve this, I introduced the `--missing-only` flag. When passed, `graphify label .` will first load `.graphify_labels.json`, preserve all successfully named communities, and only query the LLM backend for communities that are missing or still have the `Community N` placeholder.

This ensures a robust, incremental labeling process for very large codebases and saves a massive amount of LLM tokens and API costs. Tested locally on my 7,500+ node graph with perfect success.

### DIFF
--- a/graphify/__main__.py
+++ b/graphify/__main__.py
@@ -2150,6 +2150,7 @@ def main() -> None:
         print("    --no-viz                skip graph.html generation (useful for >5000 node graphs / CI)")
         print("    --graph <path>          path to graph.json (default <path>/graphify-out/graph.json)")
         print("    --no-label              keep 'Community N' placeholders (skip LLM community naming)")
+        print("    --missing-only          only re-label communities that have 'Community N' placeholders")
         print("    --backend=<name>        backend to use for community naming (default: auto-detect)")
         print("    --model=<name>          model to use for community naming")
         print("  label <path>            (re)name communities with the configured LLM backend, regenerate report")
@@ -3146,6 +3147,7 @@ def main() -> None:
         # the optional positional path can appear in any order (#724).
         no_viz = "--no-viz" in sys.argv
         no_label = "--no-label" in sys.argv
+        missing_only = "--missing-only" in sys.argv
         _backend_arg = next((a for a in sys.argv if a.startswith("--backend=")), None)
         label_backend = _backend_arg.split("=", 1)[1] if _backend_arg else None
         _model_arg = next((a for a in sys.argv if a.startswith("--model=")), None)
@@ -3178,7 +3180,7 @@ def main() -> None:
                 co_exclude_hubs = float(args[i_arg + 1]); i_arg += 2
             elif a.startswith("--exclude-hubs="):
                 co_exclude_hubs = float(a.split("=", 1)[1]); i_arg += 1
-            elif a == "--no-viz" or a.startswith("--min-community-size="):
+            elif a == "--no-viz" or a == "--missing-only" or a.startswith("--min-community-size="):
                 i_arg += 1
             elif a.startswith("--"):
                 i_arg += 1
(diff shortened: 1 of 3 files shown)


matiasduartee:
I just tested this implementation on a large internal project with over 7,500 nodes and 1,000+ communities, and it worked flawlessly. 

Before this PR, whenever we ran `graphify label .` on a graph this size, the LLM backend would inevitably hit API rate limits or safety filters halfway through the process. Because the original command always labeled everything from scratch, a single failed batch would wipe out hundreds of perfectly good names generated in previous runs, replacing them all with `Community N` placeholders again. It was practically impossible to reach 100% coverage.

By using the `--missing-only` flag introduced here, we were able to incrementally label the graph. When we hit an API rate limit, we simply waited or swapped to a local model, and ran `graphify label . --missing-only` again. It successfully preserved our 700+ good labels and only processed the remaining `Community N` placeholders until we fully labeled all 1,027 communities. This flag makes the labeling process finally viable and robust for massive codebases!

matiasduartee:
As a follow-up: I just re-ran `graphify label . --missing-only` on the same 1,024 community graph after it was fully labeled. It instantly outputted `Labeling 0 missing communities...` and finished without sending a single token to the LLM. 

I want to emphasize that this implementation makes Graphify significantly more economical. In its current state, users are forced to burn massive amounts of tokens re-evaluating and re-naming perfectly good, established clusters on every single execution. 

By matching semantic similarities and keeping established names intact, we are saving >99% in token costs and API execution time for subsequent updates to any codebase!

safishamsi (changes requested):
The core `--missing-only` logic is correct (relabels only unnamed / `Community N` placeholders, preserves real labels, int keys throughout). Three things to address:

- **Unrelated change bundled in:** the `MAX_NODES_FOR_VIZ` 5000→15000 bump in `export.py` is unrelated to this flag — please split it into its own PR.
- **No tests** — please cover the merge/skip logic (existing named communities preserved; only placeholders relabeled).
- **Confusing interaction:** `--missing-only` only takes effect on `label` (force-relabel); for plain `cluster-only` with an existing labels file, control returns early and the flag is silently ignored. A line in the help text clarifying it pairs with `label` would help. Also a couple of trailing-whitespace lines to clean up.

matiasduartee:
@safishamsi Thanks for the review! I've just pushed the requested changes:

1. Reverted the \MAX_NODES_FOR_VIZ\ increase back to 5000 in \export.py\ (we cancelled the other PR, so this is no longer needed).
2. Added tests to cover the merge/ignore logic (\	est_label_cli_missing_only_preserves_existing\ in \	ests/test_labeling.py\) ensuring existing named communities are preserved and only placeholders are renamed.
3. Clarified in the help text that \--missing-only\ works with the \label\ command and removed the trailing whitespaces.

Let me know if there's anything else!

safishamsi:
Thanks @matiasduartee, and apologies for the overlap. The `label --missing-only` flag has landed on `v8` (383cabd) — the same feature you proposed here. Your branch had drifted ~47 commits behind v8 (conflicts in `__main__.py` and the labeling tests after the #1390 concurrency changes), so the version that merged was rebased on the current signature. Credited you in the commit for proposing it first. Closing as superseded — thank you.

PR #1425, opened 2026-06-22
feat: fix examples wrong path on chinese translation doc

lau1944:


safishamsi (changes requested):
Right fix — the zh-CN doc is at depth 2, so top-level `worked/` links need `../../`. But the middle row has a typo: `(../..worked/mixed-corpus/)` is missing a slash; it should be `../../worked/mixed-corpus/`. The other two rows are correct. One-char fix and this is good to merge.

lau1944 (commented):
@safishamsi  Could you review this PR and merge?

PR #1426, opened 2026-06-22
fix(path): resolve endpoints whole-string like explain, not per-token

Gianwie:
## Problem

`graphify path "<source>" "<target>"` resolves both endpoints with `_score_nodes(...)[0]` (`graphify/__main__.py`). `_score_nodes` tokenises the label and scores per token (EXACT 1000 / PREFIX 100 / SUBSTRING 1, × idf).

A multi-word / natural-language endpoint therefore never gets a whole-string exact hit. A single-token node that EXACT-matches one of the query tokens (1000 × idf) outscores the real multi-token node (prefix 100 + substring 1 + …), so `path` resolves to the wrong node and prints a spurious **"No path found"** for a pair that is actually a hop or two apart.

`explain` does not have this problem — it already resolves whole-string via `_find_node` (exact > prefix > substring).

### Repro

Graph: `src("alpha") --contains--> tgt("beta gamma delta")`, plus an isolated decoy node labelled `"delta"`.

```
$ graphify path "alpha" "beta gamma delta"
No path found between 'alpha' and 'beta gamma delta'.   # decoy "delta" wins the scorer, has no path
```

## Fix

Resolve each endpoint with `_find_node` first (the same exact > prefix > substring whole-string precedence `explain` uses), falling back to the per-token lexical scorer only when the whole-string resolver finds nothing — so partial/fuzzy endpoints keep working.

- Same-node ambiguity guard preserved.
- The "match was ambiguous" warning now fires only on the lexical-scorer fallback (a tiered `_find_node` hit is deterministic, so no false warning).

```
$ graphify path "alpha" "beta gamma delta"
Shortest path (1 hops):
  alpha --contains [EXTRACTED]--> beta gamma delta
```

## Tests

`tests/test_path_cli.py`:
- `test_path_resolves_multiword_endpoint_whole_string` — the repro above (fails before this change).
- `test_path_falls_back_to_lexical_scorer_when_no_whole_string_match` — partial endpoint still resolves via the scorer.

Existing arrow-direction tests (#849) still pass. `pytest tests/test_path_cli.py` → 4 passed.

### DIFF
--- a/graphify/__main__.py
+++ b/graphify/__main__.py
@@ -2843,7 +2843,7 @@ def main() -> None:
                 file=sys.stderr,
             )
             sys.exit(1)
-        from graphify.serve import _score_nodes
+        from graphify.serve import _score_nodes, _find_node
         from networkx.readwrite import json_graph
         import networkx as _nx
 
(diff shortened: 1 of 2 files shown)


safishamsi (changes requested):
The premise looks stale against current `v8`. `_score_nodes` already has a whole-query resolution tier (added in #1165, ~`serve.py:141-172`) specifically so `path`/`query` resolve the same node `explain` does — and the headline test added here passes on unpatched `v8`. Two concerns:

- **Ambiguity-guard regression:** suppressing the "ambiguous match" warning whenever `_find_node` returns a hit is unsafe — `_find_node` returns a list and can contain multiple equally-exact matches (e.g. two nodes both labeled `handler`), so this silently picks one by iteration order. At minimum only suppress when `len(_find) == 1`.
- **Inconsistency:** it changes only the CLI `path` handler, leaving the MCP `_tool_shortest_path` twin (~`serve.py:802`) on the old path. If the goal is "resolve like explain," the cleaner fix is to converge `_score_nodes`/`_find_node` into one resolver both call sites use.

Gianwie:
Thanks for the careful review — you're right on all three points. `_score_nodes` already gained the whole-query tier in #1165, so `path` resolves whole-string natively and the headline test passes on unpatched `v8`; the premise here is stale. The leftover ambiguity-guard change also has the bug you flagged (`_find_node` returns a list, so `[0]` silently picks one of several equally-exact matches). Closing as superseded by #1165.

PR #1427, opened 2026-06-22
feat(label-communities): add producer for .graphify_labels.json

Gianwie:
## Problem

`serve._load_community_labels`, `report.py` and `analyze.py` already **read** a `.graphify_labels.json` next to the graph (`{community_id: label}`) and fall back to `"Community <id>"` when it's absent:

```python
# serve.py
def _load_community_labels() -> dict[int, str]:
    labels_path = Path(graph_path).parent / ".graphify_labels.json"
    if labels_path.exists():
        return {int(k): v for k, v in json.loads(labels_path.read_text()).items()}
    return {cid: f"Community {cid}" for cid in communities}
```

But nothing in the toolchain ever **writes** that file — so the read side is effectively dead and every community renders as the opaque integer `Community 42` in `explain`, the MCP resources, the report and suggested questions.

## Fix

Add the missing producer: a `graphify label-communities [--graph path]` subcommand. Deterministic, no LLM, no new dependencies.

For each Louvain community it picks:
- the dominant **source area** (first path segment of `source_file`), and
- the strongest **representative phrase** — highest-degree `concept` node, else `document`/`rationale`, else any non-generic, non-`foo()` label

and writes `<area> · <phrase>` to `.graphify_labels.json` (the exact file/format the read side expects). Errors clearly if no node has a `community` attribute yet (run `extract`/`cluster` first).

```
$ graphify label-communities --graph graph.json
Labeled 1425 communities -> .../.graphify_labels.json
  community 0     ( 530 nodes)  billing · Payment gateway
  community 4     ( 206 nodes)  agent · LangGraph orchestration for infrastructure tasks
  ...
```

## Tests

`tests/test_label_communities_cli.py`:
- writes `.graphify_labels.json` with `<area> · <concept>` labels, int-parseable keys (the `_load_community_labels` contract);
- errors when the graph has no `community` attribute.

`pytest tests/test_label_communities_cli.py` → 2 passed.

### DIFF
--- a/graphify/__main__.py
+++ b/graphify/__main__.py
@@ -2996,6 +2996,101 @@ def main() -> None:
             nodes_returned=len(connections),
         )
 
+    elif cmd == "label-communities":
+        # Produce human-readable community labels. serve._load_community_labels,
+        # report.py and analyze.py already READ a `.graphify_labels.json` next to
+        # the graph ({community_id: label}) and fall back to "Community <id>" when
+        # it is absent — but nothing WROTE it. This is that missing producer:
+        # deterministic, no-LLM, no new deps. For each Louvain community it picks
+        # the dominant source area + the strongest representative phrase and
+        # writes "<area> · <phrase>".
+        import re as _re
+        from collections import Counter as _Counter, defaultdict as _defaultdict
+
+        graph_path = _default_graph_path()
+        args = sys.argv[2:]
+        for i, a in enumerate(args):
+            if a == "--graph" and i + 1 < len(args):
+                graph_path = args[i + 1]
+        gp = Path(graph_path).resolve()
+        if not gp.exists():
+            print(f"error: graph file not found: {gp}", file=sys.stderr)
+            sys.exit(1)
+        _enforce_graph_size_cap_or_exit(gp)
+        g = json.loads(gp.read_text(encoding="utf-8"))
+        nodes = g.get("nodes", [])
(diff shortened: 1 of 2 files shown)


safishamsi (changes requested):
Useful idea (offline / no-LLM labeling), but the framing and integration need work:

- The premise that "nothing writes `.graphify_labels.json`" isn't accurate — `cluster-only`/`label` already write it (~`__main__.py:3286`). This adds a **third** path writing the same file, with a different serialization (`indent=0` vs the existing compact write) and **without** the `remap_communities_to_previous` step or the `GRAPH_REPORT.md`/`graph.json` label refresh the existing path does. So labels from this command can drift from — or silently overwrite — the canonical ones.
- Suggested direction: fold the heuristic into the existing `cluster-only` path as a `--no-llm` / `--heuristic-labels` mode rather than a separate command, so there's a single producer that keeps cid-alignment and refreshes the report. The int-key contract tests you added are good and can move with it.

safishamsi:
Thanks for this. graphify already writes `.graphify_labels.json` next to the graph: the build/label CLI path emits it (with a `.graphify_labels.json.sig` membership-signature sidecar so stale labels can be detected after re-clustering), which is what `serve`/`report`/`analyze` read back. So the producer this PR adds is already present in `v8`, and I am closing it as already implemented. Appreciate you catching the read-without-write gap.

PR #1434, opened 2026-06-23
Add `graphify resolve-entities` CLI for post-build entity dedup

iretonsean:
## Summary

Adds a `graphify resolve-entities <path>` subcommand that runs the existing `dedup.deduplicate_entities()` pipeline against an already-built `graph.json`, without re-extracting from source.

## Motivation

Today, entity dedup is only called inside `build()` during the normal extraction pipeline. If a graph is produced through a custom build script (e.g. one that wires the AST + semantic extraction together manually) or imported from an external source, the dedup step is silently skipped — and there's no first-class way to apply it after the fact short of re-extracting the whole corpus.

I hit this on a ~7k-file CRM archive (49k nodes, 76% isolated). The fix turned out to be one line in a custom build script that forgot to call `deduplicate_entities()` — but the only recovery path was either an expensive full re-extract or hand-rolling the dedup call against `graph.json`. This subcommand makes that recovery a one-liner.

## What's in this PR

- New `elif cmd == "resolve-entities"` block in `__main__.py` that loads `graph.json`, calls `deduplicate_entities()`, optionally re-runs clustering, and writes back. Auto-detects whether the graph uses NetworkX `links` or raw `edges` shape.
- Flags mirror `cluster-only`:
  - `--graph <path>` — custom location
  - `--backend <name>` — optional LLM for ambiguous-pair resolution (75–92 JW zone)
  - `--dry-run` — report merges, do not write
  - `--skip-recluster` — keep current community assignments
- A `.bak` of `graph.json` is written before the rewrite
- Help banner + README + CHANGELOG entries
- 4 new tests in `tests/test_resolve_entities_cli.py` covering dry-run safety, the apply path (merge + backup), the no-op case, and missing-graph error handling

## Non-changes

- The `dedup.deduplicate_entities()` function itself is unchanged — same algorithm, same guards (cross-project still disabled, code-keyed-by-ID still skipped, same-source-file constraint preserved).
- No behavior change to existing commands.

## Test plan

- [x] `pytest tests/test_resolve_entities_cli.py` — 4/4 pass
- [x] `pytest tests/test_dedup.py tests/test_cluster.py` — no regressions
- [x] Manually exercised on the original use-case archive (49k → 9k nodes, all in-tree dedup logic, no LLM calls)
- [x] `graphify resolve-entities . --dry-run` exits cleanly with a before/after summary
- [x] `graphify resolve-entities .` writes `.bak`, applies, and triggers re-cluster
- [x] Missing `graph.json` produces a clear error and exits 1

https://claude.ai/code/session_01GPKqfJ8eUEps7HJ1JMn6p9


iretonsean:
Closing — not the right scope for our use case. Will keep the work private and pull from local scripts.

Issue #1417, opened 2026-06-21
graphify-out files use absolute paths breaking portability across machines and directories


shiv-source:
On macOS (Darwin 25.4.0), running `/graphify .` generates files in `graphify-out/` that contain hardcoded absolute paths, making the outputs non-portable.

### Affected files

**`manifest.json`** — All file keys are absolute paths:
```json
{
  "/Users/username/projects/my-project/backend/cmd/server/main.go": {
    "mtime": 1781639382.9082484,
    ...
  }
}
```
This makes `graphify --update` fail to match any cached files when the repo is cloned to a different directory or shared between machines.

**`.graphify_root`** — Stores the absolute scan root:
```
/Users/username/projects/my-project
```
Should store the relative path (e.g., `.`) and resolve against CWD at runtime.

**`graph.json`** — Some `source_file` values leak absolute paths (files at repo root level), while most are correctly relativized. Semantic extraction subagents receive absolute paths in `FILE_LIST` and write them verbatim.

**Cache directory** (`cache/semantic/`) — Likely uses absolute paths for cache keys, same as manifest.

### Expected behavior
All paths in `graphify-out/` outputs should be relative to the scan root, so the outputs are portable across clones, machines, and users.

### Steps to reproduce
```bash
git clone https://github.com/some-project
cd some-project
graphify .
# Observe manifest.json keys, .graphify_root, and graph.json source_file values
```

### Environment
- OS: macOS (Darwin 25.4.0)
- graphify: v0.8.44, installed via `uv tool install graphifyy`

safishamsi:
Fixed on `v8` (ships in the next release).

The portability-breaking paths are now relativized:

- **`manifest.json`** — the skill runbooks now pass `root=` to `save_manifest`, so keys are stored relative to the scan root (`1f42a2d`); the native `graphify extract`/`update` CLI already did this (#777). This is the headline fix: `graphify --update` now matches cached files after a clone/move instead of re-extracting the whole corpus.
- **`graph.json` `source_file`** — relativized for nodes/edges (pre-existing) and now for `hyperedges[]` too (`d168de9`, #1418).

Two of the sub-points are intentional, not bugs:

[...]

PR #1437, opened 2026-06-23
skill: graph-health gate (Step 4.5) + anchor caches on the scan root

bahcgscateringsa-design:
Two integrity improvements to the core skill runbook (edited in
`tools/skillgen/fragments/core/core.md`, artifacts + `expected/` regenerated via
`python -m tools.skillgen` / `--bless`). These are independent — happy to split,
drop, or cherry-pick either one (see note at the bottom).

## 1. Anchor the caches on the scan root (consistency fix)

The skill currently passes `cache_root=Path('.')` to AST extraction and lets
`check_semantic_cache` / `save_semantic_cache` default to `.` — i.e. it keys the
cache on the **cwd**. The CLI extract path keys on the **scan root**:
`ast_kwargs = {"cache_root": out_root}` and `check/save_semantic_cache(root=out_root)`,
with `out_root = out_dir or target`. `build_from_json()` / `save_manifest()` already
take `root='INPUT_PATH'` in the runbook (#1361).

This change makes the runbook match the CLI: both AST and semantic caches anchor on
`INPUT_PATH`. Without it, running graphify from a cwd ≠ scan root cold-misses the
cache (every file re-extracts) and can even split AST vs semantic cache locations.

## 2. Step 4.5 — read-only graph-health gate (proposal)

`graphify/diagnostics.py` (`diagnose_extraction` / `format_diagnostic_report`) is a
ready integrity checker that isn't wired into the default pipeline. This adds a
**read-only** Step 4.5 after build / before labeling that surfaces dangling /
missing / self-loop and same-endpoint-collapsed edges — the silent-corruption modes
of incremental updates and AST/LLM id mismatches. It **never aborts**; it just makes
an integrity issue visible (per the Honesty Rules) instead of letting users query a
silently-broken graph.

## Tests

- Full suite: `2229 passed, 28 skipped`.
- `tests/test_skillgen.py`: `53 passed` (check / audit-coverage / monolith-roundtrip /
  lean-core invariants all hold), `tests/test_cache.py`: `27 passed`.

## Note on scope

I know the skill is mid lean-core split. #1 is a straightforward CLI-parity fix and
stands alone. #2 adds a step to the core, which may be at odds with the lean-core
direction — totally fine to take only #1, or to treat #2 as a discussion. Tell me
how you'd like it sliced and I'll adjust.

### DIFF
--- a/graphify/skill-amp.md
+++ b/graphify/skill-amp.md
@@ -179,7 +179,7 @@ for f in detect.get('files', {}).get('code', []):
     code_files.extend(collect_files(Path(f)) if Path(f).is_dir() else [Path(f)])
 
 if code_files:
-    result = extract(code_files, cache_root=Path('.'))
+    result = extract(code_files, cache_root=Path('INPUT_PATH'))
     Path('graphify-out/.graphify_ast.json').write_text(json.dumps(result, indent=2, ensure_ascii=False), encoding=\"utf-8\")
     print(f'AST: {len(result[\"nodes\"])} nodes, {len(result[\"edges\"])} edges')
 else:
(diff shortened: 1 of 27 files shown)


safishamsi:
Merged into `v8` as e7ba16b (cherry-picked). Verified all five skillgen guards pass on the PR (check / audit-coverage / schema-singleton / monolith-roundtrip / always-on-roundtrip) — the fragment was edited and artifacts regenerated/blessed correctly, and since the PR branched off current v8 it already carries the #1417/#1419 core.md fixes (no conflict, no stale artifacts). The cache anchoring is consumer-side only (passes root= into the already-shipped check_semantic_cache/save_semantic_cache), no conflict with the graphify.paths work. Step 4.5 calls a real diagnose_extraction API and never aborts. Thanks @bahcgscateringsa-design!

PR #192, opened 2026-04-10
feat: Add Haskell language support (.hs, .lhs)

JYC11:
Summary
  - Add Haskell as a supported language with a custom tree-sitter extractor, following the pattern used by Go, Rust, and Julia
  - Extracts modules, imports, data types (with constructors), newtypes, type aliases, type classes, instances, functions, and call edges
  - Supports literate Haskell (.lhs) via bird-track preprocessing
  - Fix existing bug: collect_files() was missing .ex, .exs, .jl extensions despite having extractors for Elixir and Julia

  Changes
  - pyproject.toml — Add tree-sitter-haskell dependency
  - graphify/extract.py — Add extract_haskell() custom extractor (~160 lines), register .hs/.lhs in _DISPATCH, fix missing extensions in _EXTENSIONS
  - graphify/detect.py — Add .hs, .lhs to CODE_EXTENSIONS
  - tests/fixtures/sample.hs — Representative Haskell fixture covering all constructs
  - tests/fixtures/sample.lhs — Minimal literate Haskell fixture
  - tests/test_languages.py — 12 new tests (module, functions, data types, constructors, typeclass, imports, instances, calls, dangling edges, .lhs parsing, collect_files coverage)
  - tests/test_extract.py — Update test_collect_files_from_dir to include new extensions

  Test plan
  - All 12 new Haskell tests pass
  - Full test suite passes (422 tests, 0 failures)
  - Tested against a real-world Haskell codebase (215 files, 5058 nodes, 5812 edges, 0 errors)

PR #1439, opened 2026-06-23
feat(haskell): add tree-sitter-haskell language support

alexbiehl:
Adds Haskell (`.hs`) as a first-class extracted language, following the existing `LanguageConfig`/`_extract_generic` pattern (modeled on Kotlin).

## What's included
- **`tree-sitter-haskell>=0.23,<0.24`** as a core dependency (abi3 wheels exist for all platforms).
- **`_HASKELL_CONFIG`** — maps `data_type`/`newtype`/`type_synomym`/`class`/`instance` → containers, `function`/`bind` → functions, `import` → imports, `apply`/`variable`/`qualified` → calls.
- **`_import_haskell`** — reads the `import` node's `module` field (Haskell-specific; Kotlin uses `path`).
- **A Haskell branch in `walk_calls`** — Haskell application nests (`f x y` → `apply(apply(f, x), y)`), so it descends the leftmost `function` field for the direct callee, and also treats bare `variable`/`qualified` leaves as calls so higher-order uses (the `area` in `map area xs`) attribute to the caller.
- `extract_haskell`, `.hs` in `_DISPATCH`, `.hs` in `detect.CODE_EXTENSIONS`.
- Fixture `tests/fixtures/sample.hs` + 7 tests (skipif when the grammar is absent).

## Two grammar gotchas handled
1. In tree-sitter-haskell a function *type* (`Shape -> Double`) is **also** a `function` node (nested under `signature`). A naïve config emits a spurious `Double()` "function". Fixed by `name_fallback_child_types=()` so type-level nodes (which lack a `name` field) are skipped. Regression test included.
2. `type_synomym` is the grammar's real (misspelled) node-type string — kept verbatim, not "corrected".

## Verification
- 7 Haskell tests pass; full `test_languages.py` = 257 passed, 13 skipped (optional `tree-sitter-dm`), no regressions.
- Exercised on a large real Haskell codebase (~470 `.hs` files): clean parse, with data types, classes, functions, imports, and cross-function call edges (e.g. `updateS7TagMonitor → resolveUpdateDataType`).

### DIFF
--- a/graphify/detect.py
+++ b/graphify/detect.py
@@ -27,7 +27,7 @@ class FileType(str, Enum):
 
 _MANIFEST_PATH = str(out_path("manifest.json"))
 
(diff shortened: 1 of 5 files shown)


Mikolaj:
Related: https://github.com/Graphify-Labs/graphify/pull/192

PR #1442, opened 2026-06-23
fix(llm): enforce API timeout in secondary LLM dispatch path

DhruvTilva:
`_call_llm` creates OpenAI and Anthropic clients without passing the configured timeout, so requests in this path do not honor `GRAPHIFY_API_TIMEOUT`.

The primary extraction paths (`_call_openai_compat` and `_call_claude`) already pass `timeout=_resolve_api_timeout()`. This change applies the same timeout configuration to `_call_llm` so behavior is consistent across all LLM calls.

## Changes

```diff
- client = anthropic.Anthropic(api_key=key, base_url=cfg["base_url"])
+ client = anthropic.Anthropic(
+     api_key=key,
+     base_url=cfg["base_url"],
+     timeout=_resolve_api_timeout(),
+ )
```

## Verification
Using a local endpoint that intentionally delayed responses and setting `GRAPHIFY_API_TIMEOUT=1`:

**Before**
* Requests in `_call_llm` did not respect the configured timeout and waited for the stalled response.

**After**
* Requests fail with a timeout exception as expected once the configured timeout is exceeded.

## Testing
* `ruff check .`
* `pyright`
* `pytest`

All pass with this change.


safishamsi:
Landed on `v8` as 4e4935a with your authorship. The bug was real and worth fixing: `_call_llm` (the dedup tiebreaker path) built its Anthropic/OpenAI clients without `timeout`, so `GRAPHIFY_API_TIMEOUT` was silently ignored there while the primary extraction paths honored it.

Heads up on why I re-applied rather than merging directly: your branch had merged `v8` in the meantime, and that merge resolved the conflict in favor of the `max_retries`-bearing line, which dropped your one-line fix — `gh pr diff 1442` had gone empty. I re-applied it on current `v8` with `timeout=_resolve_api_timeout()` added alongside `max_retries` on both constructors, and added regression coverage for both `_call_llm` branches (the path was previously untested). Thanks for catching this.

DhruvTilva:
@safishamsi   thanks for the detailed explanation and for re-applying the fix appreciate it, and i M glad I could help catch the issue

Issue #1441, opened 2026-06-23
Self-improving "work memory": capture session outcomes/corrections + a `reflect` pass on top of the graph


safishamsi:
## Summary

Explore a **self-improving "work memory"** layer on top of graphify's knowledge graph — inspired by #1401 and Perplexity's "Brain" (a context graph of the *agent's work* that, on a schedule, reviews itself and "learns to do the work better": remembering what worked/failed, what corrections were made, and which sources were dead ends).

The interesting part: **graphify already ships most of the substrate** this needs. This issue proposes a concrete, scoped first slice so a contributor can pick it up. **PR / design input welcome.**

## What graphify already has (the substrate)

| Capability | Today in graphify |
|---|---|
| Context graph | the knowledge graph |
| Auto-loadable **LLM wiki** | `graphify export wiki` → `wiki/index.md` + per-community/per-node pages (the skill already tells agents to navigate it) |
| Provenance / "show your work" | every node & edge carries `source_file` / `source_location` |
| Incremental refresh | `graphify update`, `watch`, git hooks |
| Usage feedback loop | `save_query_result` / `graphify save-result` → `graphify-out/memory/` re-ingested into the graph on the next build (`graphify/ingest.py`) |

## The gap (what "self-improving memory" adds)

1. **Memory of the *work*, not just the corpus.** graphify graphs your code/docs; this would also capture *sessions, outcomes, and corrections* — "this query/answer was useful," "this source was a dead end," "the user corrected X."
2. **A scheduled *reflection* pass.** A `graphify reflect`-style step (cron/overnight) that reviews the accumulated work-memory + graph and synthesizes durable lessons back into the wiki/graph — the actual "self-improving" piece.
3. **A corrections/mistakes loop** — reweight or prune memory from dead ends and corrections so future retrieval improves.

## Layer split (why this is mostly thin glue + one new step)

graphify is the **memory substrate** (graph + wiki + provenance + feedback ingestion); the autonomous loop is largely **agent orchestration on top**. Much is feasible with today's primitives:

- **Feasible now:** record session outcomes/corrections as memory docs (extend `save-result` with an outcome/correction field), then cron a `graphify update` to fold them in.
- **The novel piece worth building:** a `graphify reflect` step that reads the work-memory + graph and writes a synthesized "lessons / preferred sources / known-dead-ends" wiki page the agent loads next session.

## Proposed first slice (scoped for a PR)

[...]

PR #1443, opened 2026-06-23
feat: work-memory outcomes + `graphify reflect` (#1441)

TPAteeq:
# feat: work-memory outcomes + `graphify reflect` (#1441)

First deterministic slice of the self-improving "work memory" loop proposed in #1441.

graphify already files Q&A results back into the graph (`graphify save-result` →
`graphify-out/memory/` → re-ingested on the next build). This PR adds the missing
two pieces from the issue's "proposed first slice": a way to record **how a saved
answer turned out**, and a `reflect` step that turns those outcomes into a lessons
doc an agent can load at the start of the next session.

Deterministic, no LLM. Bare `graphify save-result` and all existing behavior are
unchanged.

## What

- **`graphify save-result --outcome useful|dead_end|corrected [--correction TEXT]`** —
  optional work-memory signals attached to a saved Q&A.
- **`graphify reflect`** — scans `graphify-out/memory/`, aggregates the outcomes, and
  writes `graphify-out/reflections/LESSONS.md`:
  - **Preferred sources** — nodes that recurred in `useful` answers (start here).
  - **Known dead ends** — questions/sources marked `dead_end` (don't re-derive).
  - **Corrections** — answers the user corrected, and what the right answer was.
  - Grouped by community when a `graph.json` is present; a single flat section otherwise.

## How

- The outcome/correction are written to **both** places on purpose:
  - the memory-doc **frontmatter** (`outcome:` / `correction:`), so `reflect` can parse
    them deterministically without an LLM, and
  - an **`## Outcome` body section**, so the signal round-trips into the graph on the
    next semantic re-extraction (the existing re-ingestion is the extractor reading the
    memory-doc prose, per the `graphify-out/memory/` special-casing in `detect.py`).
- `reflect` is fully deterministic: docs are sorted by `(date, filename)`, sources ranked
  by `(count desc, name asc)`, community ties broken to the lexicographically-smallest
  label, and `Uncategorized` always sorts last — so the same `memory/` contents always
  produce byte-identical output.
- Frontmatter parsing reuses the repo's hand-rolled YAML convention (no PyYAML dep):
  `reflect._yaml_unescape` reverses `ingest._yaml_str`. Foreign/non-frontmatter `.md`
  files in `memory/` are skipped, never crashed on.

## Changes

- `graphify/ingest.py` — `save_query_result` gains `outcome`/`correction` params + a new
  `OUTCOMES` constant; `source_nodes` are now escaped via `_yaml_str` (they previously
  weren't, which would have produced invalid YAML for a node name containing `"`/`\`).
- `graphify/reflect.py` *(new)* — `parse_memory_doc`, `load_memory_docs`,
  `aggregate_lessons`, `render_lessons_md`, `reflect`.
- `graphify/__main__.py` — `reflect` dispatch + help; `--outcome`/`--correction` on
  `save-result`.
- `graphify/__init__.py` — lazy-import registration for `reflect` / `save_query_result`.
- `CHANGELOG.md`, `README.md` — entry + command-reference lines.

## Tests

- `tests/test_reflect.py` *(new)* — parsing round-trips (incl. backslash/newline/quoted
  node names and CRLF), aggregation, deterministic ordering / byte-stability, community
  plurality + tie-break, the issue's "second session benefits from the first" worked
  example, and the CLI surface (incl. cold-start with no `memory/`, bad-`--outcome`
  rejection, `--out`, and community grouping against a real `graph.json`).
- `tests/test_ingest.py` — outcome/correction frontmatter+body, and a backward-compat
  test proving a no-outcome result is byte-identical to before.
- Full suite **2314 passed, 2 skipped**; `python -m tools.skillgen --check` clean; ruff
  clean; pyright clean on the new files.

## Notes / scope

A few deliberate calls worth a maintainer's eyes — happy to change any of these:

1. **Output goes to `graphify-out/reflections/LESSONS.md`, not `wiki/_LESSONS.md`.**
   `graphify export wiki` unlinks every `wiki/*.md` on each run, so a lessons file written
   there would be clobbered on the next export. The issue explicitly allowed "a
   `reflections/` doc," so I used that. Easy to move if you'd prefer it under `wiki/`
   (would need `reflect` to run after `export wiki`, or to hook into `to_wiki`).
2. **Deterministic only.** The issue's optional "LLM-synthesis behind a backend flag" is
   intentionally deferred — the issue framed it as "deterministic first … LLM-synthesis
   optional." No half-wired flag.
3. **`reflect` is not wired into the skill runbook yet.** Agents won't auto-run it or
   auto-load `LESSONS.md`; that's the natural next slice (the orchestration layer the
   issue calls "largely agent orchestration on top"). This is also *why* `skillgen --check`
   stays green — no skill fragment references the new command, so nothing needs
   re-rendering.

Because of (2) and (3), this PR is the **first slice** of #1441, not a full close — the
autonomous/scheduled reflection and per-user memory remain out of scope per the issue.

## Checklist

- [x] `uv run pytest tests/ -q` green (2314 passed, 2 skipped)
- [x] `python -m tools.skillgen --check` clean
- [x] `ruff` clean
- [x] No version bump
- [x] Bare `save-result` / existing behavior unchanged (backward-compat test)


safishamsi:
Really solid first slice — deterministic, no-LLM, well-tested, and writing the lessons to `reflections/` (rather than back into `memory/`) is exactly the right call. Before we merge, a few refinements worth folding in. **Three of these fix real issues, not just polish:**

1. **Recency-aware scoring instead of raw frequency** *(real conflict bug)* — `_rank_sources` ranks by raw count, so the same node can show up as *both* a preferred source and a dead end, and a stale `useful` from months ago outranks a fresh `dead_end`. Score each source node as a signed, time-decayed value (`useful` positive, `dead_end`/`corrected` negative, with a half-life — default ~30d, exposed as a flag). Have it work over the doc list (which carries `date` + `outcome`) instead of a pre-summed `Counter`, and pass `now` in as a parameter so output stays byte-stable in tests.
2. **Corroboration threshold (k≥2)** *(real robustness issue)* — right now a single `save-result` mints a trusted lesson, and saves are agent-driven, so one wrong save propagates into the next session. Only promote a node to "preferred" once it's corroborated by ≥2 distinct results (default 2, configurable); render the rest as "tentative."
3. **Node-existence gate** *(stale-lesson fix)* — `reflect` already loads `graph.json`, so drop any `source_node` that's no longer in the current node set. A lesson pointing at deleted/renamed code shouldn't keep getting recommended. Nearly free given the graph is already in hand.
4. **"Contested" rendering** — when a node has both positive and negative signals, render it once under a "Contested" line with counts + dates and a recency-wins verdict, instead of silently appearing in both buckets.
5. **Two small guards** — reword the `LESSONS.md` header from "reuse what worked / skip what didn't" to something more cautious ("verify before relying; revisit dead ends if the code has changed"), and add a regression test that the `reflections/` output can never be globbed back into `load_memory_docs`.

All five stay deterministic / embedding-free and land in functions that already exist (`aggregate_lessons`, `_rank_sources`, `_render_bucket`, `render_lessons_md`, `reflect`). Nice work 🙏

TPAteeq:
Thanks for the detailed review. All five folded in (75f2eeb):

  - Recency-aware scoring — source nodes are now scored as a signed, time-decayed value (useful +, dead_end/corrected −) over the doc list instead of a raw count, with a configurable half-life (--half-life-days, default 30) and now passed in so output stays byte-stable. A fresh dead end outweighs a stale useful, and the same node can't land in both buckets anymore.
  - Corroboration (k≥2) — a node is only promoted to preferred once ≥--min-corroboration distinct results back it (default 2); the rest render as tentative. Counting is per-distinct-doc, so one save citing a node twice still counts once.
  - Node-existence gate — reflect already loads graph.json, so source nodes no longer in the node set are dropped before scoring.
  - Contested — mixed-signal nodes render once under a Contested line with the counts, the latest date, and a recency-wins verdict, instead of silently appearing in two places.
  - Guards — header reworded toward "verify before relying; revisit dead ends if the code has changed", and added a regression test that the reflections/output can never be globbed back into load_memory_docs (no frontmatter → skipped even if it lands in memory/).

  All deterministic / embedding-free. Full suite + skillgen-check green, with tests added for each (recency in both directions, distinct-doc corroboration, the gate end-to-end, the evenly-split verdict, negative-only isolation).

safishamsi:
The self-improving work-memory loop has shipped to `v8` (commit 89dd00f), reworked from this PR with recency-decayed signed scoring, a corroboration threshold, a label-matched node-existence gate, contested-conflict handling, and determinism hardening — verified end-to-end on real data. Closing this as superseded by the shipped version. Thanks for surfacing the idea from #1441.

PR #1447, opened 2026-06-24
Fix: validate_extraction crashes on non-hashable node id / edge endpoint

dschwartzi:
## Problem

`validate_extraction()` is documented to *return* a list of error strings ("empty list means valid"), but it raises `TypeError: unhashable type: 'list'` when a node `id` — or an edge `source`/`target` — is a non-hashable value such as a list. This happens in practice when an LLM extraction subagent emits malformed JSON like `{"id": ["foo", "bar"], ...}`.

Two crash sites:
- `node_ids = {n["id"] for ...}` — the set comprehension over node ids.
- `edge["source"] not in node_ids` / `edge["target"] not in node_ids` — the set-membership tests.

Because `build_from_json()` calls `validate_extraction()` at its start, the whole build aborts before any graph is produced — a single malformed node from a semantic pass kills an otherwise-complete extraction of a large corpus. `build_from_json()` itself would also raise (`G.add_node(<list>)`, and the `not in node_set` membership test) even if the validator were bypassed.

## Fix

- **`validate.py`**: collect `node_ids` during the node pass, adding only hashable ids; report a non-hashable id/endpoint as an error string instead of crashing. All existing error messages and the dangling-edge checks are preserved.
- **`build.py`**: in `build_from_json`, skip dict nodes with a missing/non-hashable id and edges with non-hashable endpoints (with a stderr warning), so a malformed entry is dropped rather than aborting the build. Non-dict nodes are deliberately left to raise as before, so `multigraph_diagnostics` still observes shape errors.

## Tests

- 3 new `tests/test_validate.py` cases: non-hashable node id reported, non-hashable edge endpoint reported, and valid ids still collected (dangling edge still flagged) when a sibling id is bad.
- 2 new `tests/test_build.py` cases: build skips non-hashable/missing node ids and non-hashable edge endpoints while keeping well-formed entries.

Full suite locally (`pytest tests/`): **2331 passed**, with the only 2 failures (`test_ollama::test_detect_backend_ollama`, `test_detect_backend_none_without_envvars`) being pre-existing and environmental — they fail identically on a pristine checkout because backend auto-detect resolves to an installed paid-API package. **This change adds 5 passing tests and introduces 0 new failures.** `ruff check` is clean on the changed files.

### DIFF
--- a/graphify/build.py
+++ b/graphify/build.py
@@ -182,8 +182,25 @@ def build_from_json(extraction: dict, *, directed: bool = False, root: str | Pat
         print(f"[graphify] Extraction warning ({len(real_errors)} issues): {real_errors[0]}", file=sys.stderr)
     G: nx.Graph = nx.DiGraph() if directed else nx.Graph()
     for node in extraction.get("nodes", []):
-        if "source_file" in node:
-            node["source_file"] = _norm_source_file(node["source_file"], _root)
+        # Skip dict nodes with a missing or non-hashable id (e.g. a list emitted
+        # by a buggy LLM extraction) so NetworkX add_node never raises
+        # TypeError: unhashable type. Non-dict nodes are deliberately left to
+        # raise as before, so callers that probe build for shape errors (e.g.
+        # the multigraph diagnostic) still observe the malformed shape.
+        if isinstance(node, dict):
+            if "id" not in node:
+                continue
+            try:
+                hash(node["id"])
+            except TypeError:
+                print(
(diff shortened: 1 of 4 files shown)


safishamsi:
Merged into `v8` via cherry-pick (commit 2c7cbb6, authorship preserved). Verified locally: full suite 2333 passed, 0 failures. Clean defensive fix — thanks for catching that a single malformed node would abort the whole build, and for leaving non-dict nodes to raise so the multigraph diagnostic still sees shape errors. 🙏

PR #1448, opened 2026-06-24
fix(serve): show community name in get_community MCP output

rmart1308:
﻿## What

`get_community` was the only MCP graph query tool still returning a bare numeric
community id. `get_node` (and the query-traversal output in `_subgraph_to_text`)
already read the `community_name` node attribute that `to_json` writes onto every
node, so an agent that had located a community by name â€” via `GRAPH_REPORT.md`,
`get_node`, or query output â€” had no way to confirm it through `get_community`.

This makes `get_community` consistent with the rest of the serve layer: it reads
the name from the community's member nodes (they all share it) and formats it into
the header, e.g.

```
Community 12 â€” Auth & Sessions (8 nodes):
  login() [auth/router.py]
  ...
```

It falls back to the bare id for graphs built before `community_name` was
persisted, and routes the value through `sanitize_label` like every other
LLM-derived field (F-010).

## Why

Small consistency/UX gap. Once you can see community names everywhere else, the
one tool whose whole job is "show me this community" shouldn't be the one that
hides them behind a number.

## Notes

- Follows the existing `get_node` precedent (inline read of `community_name`), so
  no new test was added â€” the tool closures aren't unit-tested directly in
  `tests/test_serve.py`, and the `community_name` attribute itself is produced and
  tested in the `to_json` / export path. Happy to extract a small module-level
  formatter and add a focused test if you'd prefer the change be directly covered.
- CHANGELOG `Unreleased` entry uses a `(#XXXX)` placeholder â€” replace with the PR
  number on merge.

ðŸ¤– Generated with [Claude Code](https://claude.com/claude-code)


safishamsi:
Landed on `v8` (commit f9ded63), authored to you. I added two things on top of your change: the name is skipped when it is only the `Community N` placeholder (written for unnamed communities), so the header doesn't read `Community 12 — Community 12`; and I pulled the formatting into a small module-level `_community_header()` with focused tests (named / placeholder / empty / sanitised), which also answers your note about the tool closures not being directly testable. Verified: serve/MCP suite green, full suite 2397 passed. Thanks for spotting the consistency gap and for the clean writeup. Closing as landed.

PR #1454, opened 2026-06-25
fix(ids): use repo-relative file node stems

### LINKED ISSUE #1158: Node IDs are context-dependent (_file_stem uses immediate parent of the passed path) → split & cross-repo-collided nodes
## Summary

`graphify.extract._file_stem()` derives a node-ID prefix from `path.parent.name` of whatever `Path` it is handed, so node identity is **not** a pure function of the file's repo-relative location. The same entity can get two IDs (split nodes), and different entities can get one ID (collisions). Both corrupt degree/betweenness and community structure.

### 1. Split nodes (when IDs come from more than one extractor)
The AST extractor receives **absolute** paths, so a repo-root file `foo.py` gets the **repo folder name** baked in: `<repofolder>_foo_<entity>`. Any other ID source that records a relative `source_file` / bare stem (an LLM/agent extraction step, a hand-authored fragment) produces `foo_<entity>`. They never join → the entity becomes two nodes with split degree. (`_pick_winner` in `dedup.py` already prefers the *shorter, un-prefixed* ID — i.e. the repo-prefixed form is effectively non-canonical.)

### 2. Cross-repo collisions (multi-repo workspace)
`_file_stem` qualifies with only the **immediate parent** dir. Running `graphify extract .` over a workspace of several repos collides same-named files: every service's `app/models.py` → one `app_models_*` node; `tests/conftest.py` from N services → shared `tests_conftest_*` nodes. Distinct entities silently merge, fabricating cross-service edges.

Real-world data point: a single `extract .` over a 7-repo workspace produced **29 such collisions** (e.g. `tests/conftest.py` merged across 6 services, `app/app.py` across 5). `merge-graphs`/`prefix_graph_for_global` *do* namespace correctly via `<repo>::<id>` + the `repo` attribute — but a combined `extract .` does not, and that's a common usage.

## Root cause

```python
def _file_stem(path: Path) -> str:
    parent = path.parent.name          # immediate parent of the *passed* path
    if parent and parent not in (".", ""):
        return f"{parent}.{path.stem}"
    return path.stem
```

Identity depends on CWD / absolute-vs-relative path / repo folder name, and disambiguates only one directory level (partial #550 — still collides for same-named files 2+ dirs deep).

## Invariant

A node's identity should be a pure function of **(repo-relative path, entity)** — stable across CWD, absolute-vs-relative paths, and the repo folder's name. The "which repo" dimension belongs in the `repo` attribute (which `deduplicate_entities` and `prefix_graph_for_global` already use), never in the ID.

## Proposed fix

1. Thread the repo/scan root into `_file_stem` and qualify by the **full repo-relative path** (also fixes deep same-name collisions):
   ```python
   def _file_stem(path, repo_root=None):
       if repo_root is not None:
           try:
               rel = path.resolve().relative_to(Path(repo_root).resolve())
               return ".".join(rel.with_suffix("").parts)
           except ValueError:
               pass
       parent = path.parent.name  # legacy fallback (unchanged)
       return f"{parent}.{path.stem}" if parent not in (".", "") else path.stem
   ```
2. Extract identity into one `graphify/ids.py` (`node_id(repo_root, path, entity)`) called by every ID-producing path, so non-AST sources can't drift.
3. Add a `validate.py` collision check (`canonical_id -> [ids]`) wired into `validate_extraction` / `build_from_json` so a regression fails loudly.
4. Backward-compat: this changes existing IDs → ship a one-time `graphify migrate-ids`. I have a working reference implementation (canonicalization + alias-assertion + selftest) and am happy to contribute it.

## Tests to add
- `canon` idempotency; deep-nesting (`a/b/c.py` vs `x/b/c.py` distinct); rename-invariance (same repo under two folder names → identical IDs); collision-hook flags a split and passes a clean graph.

Filing as an issue first since it's a backward-incompatible change to node IDs — happy to open a PR if this direction looks right.

pkudinov:
Fixes #1158.

## Problem

File-derived node IDs were based on the immediate parent directory plus file stem. That makes IDs depend on how a file path reaches the extractor and can also collide in deeper source trees, for example:

- `packages/a/src/index.py` -> `src_index`
- `packages/b/src/index.py` -> `src_index`

For symbol nodes, the same collapse can produce split nodes or false traversal between unrelated files.

## Change

- Add a shared `file_stem()` helper that uses the full repo-relative path without extension when the repo root is known, or when a relative path is already supplied.
- Keep the historical one-parent fallback only for absolute paths that cannot be related to a repo root yet, so the existing post-extraction remap path can still canonicalize them later.
- Route AST extraction, MCP ingest, and shell symbol-resolution ID construction through the same path-stem recipe.
- Update the extraction spec prompt text and generated skill references so semantic extraction follows the same contract.

## Verification

- `uv run pytest tests/test_file_node_id_spec.py -k 'deep_same_named_files or repo_root_name'` failed before the fix and passes after it.
- `uv run pytest tests/test_file_node_id_spec.py tests/test_extraction_spec_ids.py tests/test_id_normalization_contract.py`
- `uv run pytest tests/test_extract.py tests/test_js_import_resolution.py tests/test_python_import_resolution.py tests/test_mcp_ingest.py tests/test_build.py tests/test_validate.py`
- `uv run pytest`
- `uv run graphify update .`
- `git diff --check`


safishamsi:
Thanks @pkudinov — this proposed exactly the right fix (repo-relative file-node stems), and it shipped in **0.9.0**. The released change makes the full repo-relative path the node-ID stem across the AST extractor, the LLM prompt, and the extraction-spec (fixing #1504/#1509/#1158), plus a deterministic re-key in `build_from_json` so existing graphs and the unversioned semantic cache migrate without ghosts or a re-bill, and a legacy-graph warning. Closing as implemented — credit to you for proposing the repo-relative-stem direction. Thank you.

PR #1455, opened 2026-06-25
fix(packaging): include extractor subpackage

pkudinov:
## Problem

The current wheel declares only `packages = ["graphify"]`, but `graphify.extract` now imports `graphify.extractors.base`.

A source-tree test run passes because the package is importable from the checkout, but an installed git package can fail at runtime:

```text
ModuleNotFoundError: No module named 'graphify.extractors'
```

One minimal install-time repro is:

```bash
uvx --isolated --from git+https://github.com/safishamsi/graphify.git@v8 graphify extract . --no-cluster --out .
```

## Change

- Include `graphify.extractors` in the setuptools package list.
- Extend the wheel packaging guard so every committed `graphify/**/*.py` module must be present in the built wheel, not only the Markdown skill artifacts.

## Verification

- `uv run pytest tests/test_wheel_packaging.py`
- `uv run graphify update .`

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -107,7 +107,7 @@ dev = [
 package = true
 
 [tool.setuptools]
-packages = ["graphify"]
+packages = ["graphify", "graphify.extractors"]
 include-package-data = false
 
 [tool.setuptools.package-data]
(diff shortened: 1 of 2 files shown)


safishamsi:
Thanks, but this is already fixed and released. The missing `graphify.extractors` subpackage was added to `[tool.setuptools] packages` and shipped in **0.8.49** (0.8.48 was yanked for exactly this). Verified by installing the 0.8.49 wheel into a clean venv and confirming `import graphify.extractors` and `graphify extract` both work. Closing as already resolved, appreciate you catching it independently.

Issue #497, opened 2026-04-22
wiki.py: _safe_filename(label) collisions cause silent article overwrites


belyk1:
## Summary
`graphify/wiki.py` writes one Markdown file per community and per god node using `_safe_filename(label)` to derive the filename (lines 198, 206). There is no collision check. If two different communities (or a community and a god node, or two god nodes) produce the same `_safe_filename(...)` output, the second `write_text()` silently overwrites the first. No warning is emitted.

## Environment
- graphify installed from PyPI
- Python 3.14 on Windows 11

## Relevant code
`graphify/wiki.py`:
- Line 198: `(out / f"{_safe_filename(label)}.md").write_text(article)` — in the community-articles loop.
- Line 206: `(out / f"{_safe_filename(node_data['label'])}.md").write_text(article)` — in the god-node-articles loop.

Neither call guards against the filename already existing. Whichever iteration writes last wins.

## Repro

**Reproducing live in a real corpus right now.** Our knowledge graph of `C:\\AI` has 1,081 nodes and 64 labeled communities. Two distinct communities have both been assigned the label `ComfyUI Model Releases (Reddit Digest)` — same string, different community IDs. `_safe_filename(...)` therefore produces the same slug for both, and `to_wiki()` silently overwrites one community's article with the other's. `wiki/index.md` lists one of them; the other community's content is simply gone.

The duplicate labels in our case come from an algorithmic labeling step — a keyword heuristic in a user-side post-processing script that runs on top of graphify's community clustering (not upstream code). But the issue generalizes: any caller that passes `to_wiki()` a `community_labels` dict with duplicate values (whether produced by a heuristic, an LLM, or manual tagging) will hit the same silent overwrite. The library has no way to know the labels collide unless it checks.

**Minimal conceptual repro** (independent of our specific labeler): pass `to_wiki()` a `communities` dict plus a `community_labels={1: "Foo", 2: "Foo"}` argument. Observe: `wiki/` contains one `Foo.md` file, `index.md` contains two entries pointing at it, one community's article content is lost, no warning surfaces.

## Impact
- Silent data loss: the losing community/god-node's article is simply not present in the wiki.
- `wiki/index.md` may list the entry but link to a file whose content is actually from a different subject (the winner of the slug race).
- Installations that use the wiki as the primary navigation surface get misled — the index appears complete, but points at articles that don't match their headings.

## Suggested fix (maintainer's choice)
[...]

PR #1457, opened 2026-06-25
Fix case-only filename collisions silently overwriting Obsidian/Canvas/Wiki notes

### LINKED ISSUE #1453: to_obsidian()/to_canvas(): case-sensitive seen_names dedup silently overwrites files on case-insensitive filesystems
## Summary

Both `to_obsidian()` and `to_canvas()` in `graphify/export.py` deduplicate node-label-derived filenames using a case-sensitive `dict[str, int]` keyed on the literal sanitized name. macOS (APFS default) and Windows (NTFS default) are case-insensitive filesystems. Two node labels that differ only by case (e.g. "References" vs "references") are treated as non-colliding in the dedup map, so both get written without a numeric suffix — and the second `write_text()` silently overwrites the first **on disk**, with no error, warning, or any signal in the function's return value.

This is the same class of bug as #497 (`wiki.py`'s `_safe_filename` having no collision check at all), except here a collision check *does* exist and *does* work correctly for same-case duplicates — it just doesn't account for filesystem case-insensitivity, so it silently fails for a specific subset of collisions.

## Environment

- `graphifyy` 0.8.47 (current PyPI/pipx, confirmed via `pyproject.toml` on `main`)
- macOS / APFS (default, case-insensitive) — also reproducible on Windows/NTFS default config

## Relevant code

`graphify/export.py`:

`to_obsidian()`, lines ~879-886:
```python
seen_names: dict[str, int] = {}
...
if base in seen_names:
    seen_names[base] += 1
    node_filename[node_id] = f"{base}_{seen_names[base]}"
else:
    seen_names[base] = 0
```

`to_canvas()`, lines ~1134-1141 — identical pattern, duplicated:
```python
seen_names: dict[str, int] = {}
...
if base in seen_names:
    seen_names[base] += 1
    node_filenames[node_id] = f"{base}_{seen_names[base]}"
else:
    seen_names[base] = 0
```

Both check membership with the exact-case `base` string as the dict key, so `"References"` and `"references"` are distinct keys in Python even though they resolve to the same path on a case-insensitive filesystem.

## Repro

Build any graph where two distinct node labels differ only by case (common with auto-extracted identifiers/class names vs. lowercase doc references, e.g. a class `References` and a markdown section heading `references`). Run:
```bash
graphify .
graphify export obsidian
ls graphify-out/obsidian/*.md | wc -l
```
The count on disk will be lower than `G.number_of_nodes()` (plus community notes) by the number of case-only collisions, with no warning printed. Confirmed empirically on a 3,288-node real-world graph: the unmodified installed `to_obsidian()` produced 3,366 files against an expected 3,489 — **123 notes silently lost** to case-collision overwrites.

## Impact

Silent data loss, same severity class as #497: a node's note is simply gone from the vault, with nothing in stdout/stderr or the function's return value indicating it happened. Anyone running `export obsidian`/`export canvas` on the two most common desktop OSes (macOS, Windows) is exposed to this whenever their graph happens to contain case-only-distinct labels — which is plausible for any corpus mixing code identifiers and prose.

## Suggested fix

Key `seen_names` on the lowercased form while still emitting the original-case `base` in the filename, e.g.:
```python
seen_names: dict[str, int] = {}  # keyed on base.lower()
...
key = base.lower()
if key in seen_names:
    seen_names[key] += 1
    node_filename[node_id] = f"{base}_{seen_names[key]}"
else:
    seen_names[key] = 0
    node_filename[node_id] = base
```
This guarantees a numeric suffix on any collision that would actually collide on a real case-insensitive filesystem, matching the existing suffix behavior for same-case duplicates. Worth fixing in both `to_obsidian()` and `to_canvas()` since the logic is duplicated between them — and possibly extracting into one shared helper to avoid a third copy drifting out of sync in the future.

TPAteeq:
## Summary

Closes #1453.

Three exporters derive on-disk filenames from node/community labels and
dedup them, but keyed the dedup on the **exact-case** name. On
case-insensitive filesystems — macOS/APFS and Windows/NTFS, the two most
common desktop targets — two labels that differ only by case (e.g. a class
`References` and a markdown heading `references`) are distinct keys but
resolve to the **same path** on disk, so the second `write_text()` silently
overwrites the first: no numeric suffix, no warning, no signal in the
return value. Same data-loss class as #497.

The reporter confirmed it empirically: a 3,288-node graph wrote 3,366 files
against an expected 3,489 — 123 notes silently lost.

## Fix

- **`to_obsidian` / `to_canvas` (`graphify/export.py`):** the node-note
  dedup now keys a `used` set on `base.lower()` while still emitting the
  **original-case** filename, so a real collision gets a numeric suffix. The
  dedup was duplicated verbatim across both functions (the issue flagged the
  drift risk), so it's extracted into one shared `_dedup_node_filenames()`
  helper.
- **`to_wiki` (`graphify/wiki.py`):** `_unique_slug` had the same
  case-sensitive `slug in used_slugs` membership (the #497 dedup didn't fold
  case); it now folds case the same way.
- **Review-driven hardening (same PR):**
  - `to_obsidian`'s `_COMMUNITY_*.md` overview notes had **no dedup at all**,
    so two community labels differing only by case — or even identical after
    sanitizing — clobbered each other. One case-folded-deduped filename per
    community is now computed up front and used for both the write and the
    intra-vault `[[_COMMUNITY_...]]` cross-references so they stay consistent.
  - The generated suffix is itself re-checked, so a synthesized `base_1`
    can't silently overwrite a node whose literal label is already `base_1`
    (this collides on case-sensitive filesystems too).

## Testing

- Regression tests in `tests/test_export.py` and `tests/test_wiki.py`,
  robust on both filesystem types: each asserts every note lands on disk
  (catches the overwrite on case-insensitive FS) **and** that no two stems
  collide under `.lower()` (catches the missing suffix on case-sensitive FS).
  Coverage: obsidian/canvas case collision, literal-suffix collision,
  community-note collision, obsidian↔canvas filename agreement (the real CLI
  path calls them separately), wiki community↔community and god-node↔community
  collisions. Each new test verified red on the pre-fix code, green after.
- Full `test_export.py` / `test_wiki.py` plus the adjacent
  obsidian/canvas/CLI/pipeline suites pass (99 tests).
- Verified end-to-end in Obsidian: a graph with `References`/`references`
  and `Parser`/`parser`/`PARSER` produces five distinct notes
  (`References`, `references_1`, `Parser`, `parser_1`, `PARSER_2`) and a
  canvas whose cards each resolve to a real file.

## Out of scope (follow-ups)

- `report.py`'s `GRAPH_REPORT.md` emits its own `[[_COMMUNITY_...]]` hub
  links via a separate sanitizer; for a suffixed (case-colliding) community
  its link can dangle. Pre-existing cross-tool coupling — the vault's own
  notes and internal links are now correct and no longer lose data.
- `transcribe.py` writes `{stem}.txt` transcripts with no dedup, so files
  from different dirs with case-only-distinct stems can overwrite. Separate,
  pre-existing, not label-keyed — maybe worth its own issue.


safishamsi:
Landed on `v8` (commit 5d63aad), authored to you. Verified with a rigorous edge-case battery beyond the included tests: case-only collision (`References` + `references_1`), the `base_1` literal re-check (a synthesized `base_1` becomes `base_1_1` rather than clobbering a node literally named `base_1`), community-label case folding (`_COMMUNITY_API` + `_COMMUNITY_api_1`), and determinism across runs. Full suite 2404 passed, ruff + skillgen clean. Nicely reasoned fix — the shared helper and the suffix re-check are exactly right. Closing as landed.

PR #1458, opened 2026-06-25
feat(llm): add *_BASE_URL env var support for kimi, gemini, deepseek backends

jc2shile:
## What

Previously `kimi`, `gemini`, and `deepseek` backends had **hardcoded** API URLs:

| Backend | Hardcoded URL |
|---------|--------------|
| kimi | `https://api.moonshot.ai/v1` |
| gemini | `https://generativelanguage.googleapis.com/v1beta/openai/` |
| deepseek | `https://api.deepseek.com` |

Other backends (`openai`, `ollama`, `anthropic`) already support `*_BASE_URL` env var overrides.

This PR adds the same pattern:

| Env Var | Overrides |
|---------|-----------|
| `KIMI_BASE_URL` | Kimi/Moonshot endpoint |
| `GEMINI_BASE_URL` | Gemini endpoint |
| `DEEPSEEK_BASE_URL` | DeepSeek endpoint |

## Why

Users behind corporate proxies, LiteLLM gateways, or self-hosted LLM servers need to route **all** OpenAI-compatible backends through a single endpoint. Previously only `openai` and `ollama` supported this.

## Changes

- `graphify/llm.py`: Replace hardcoded `base_url` strings with `os.environ.get("X_BASE_URL", default)` for kimi, gemini, deepseek
- `tests/test_custom_base_url.py`: 6 new tests covering env var override + default fallback for each backend

## Testing

```
pytest tests/test_custom_base_url.py -v  # 6 passed
pytest tests/test_llm_backends.py -v     # 51 passed (19 skipped due to pre-existing Windows tmpdir PermissionError)
```


safishamsi:
Thanks @jc2shile — good call. The `kimi`, `gemini`, and `deepseek` backends hardcoded their `base_url`, so anyone behind an OpenAI-compatible proxy/gateway or running a self-hosted relay had no way to redirect them, even though `ollama` and `openai` already read `*_BASE_URL`.

Landed on `v8` in 68dba89 with your authorship preserved. The PR branch had picked up 624 unrelated files from a stale base, so I ported just the clean 16-line `llm.py` change onto current `v8` rather than merging the branch. Each backend now reads `KIMI_BASE_URL` / `GEMINI_BASE_URL` / `DEEPSEEK_BASE_URL` and falls back to its official default when unset, so nothing changes for anyone who doesn't set it. Added tests covering both the override and the default for all three backends.

Closing as merged-by-port. Thanks for the contribution.

PR #1459, opened 2026-06-25
Fix to_canvas: place node cards in the same sqrt(n)-column grid the group box is sized for

### LINKED ISSUE #1452: to_canvas(): group box sizing uses ceil(sqrt(n)) columns but node placement hardcodes 3
## Summary

`to_canvas()` in `graphify/export.py` computes each community's group-box width from a `ceil(sqrt(n))`-column grid, but the node-placement loop that fills that box hardcodes exactly 3 columns. For any community larger than ~9-12 members, the box is sized for a roughly-square grid while the actual node cards render as a narrow 3-wide column that wraps far past the box's intended height — producing a tall, mostly-empty box with a cramped 3-wide strip of nodes squeezed into one side.

## Environment

- `graphifyy` 0.8.47 (current PyPI/pipx, confirmed via `pyproject.toml` on `main`)
- macOS (but this is a pure layout-math bug, not platform-specific)

## Relevant code

`graphify/export.py`, `to_canvas()`:

Box sizing (line ~1169) explicitly targets a `ceil(sqrt(n))`-column layout:
```python
w = max(600, 220 * math.ceil(math.sqrt(n)) if n > 0 else 600)
h = max(400, 100 * math.ceil(n / 3) + 120 if n > 0 else 400)
```
(note: `h` here is already inconsistently using `/3`, not `sqrt(n)`, while `w` uses `sqrt(n)` — the two dimensions don't even agree with each other on an assumed column count.)

Node placement (line ~1238) hardcodes 3 columns regardless of `n`:
```python
col = m_idx % 3
row = m_idx // 3
nx_x = gx + 20 + col * (180 + 20)
nx_y = gy + 80 + row * (60 + 20)
```

## Repro

```bash
graphify .                 # any repo with a community of ~20+ members
graphify export obsidian
# open graphify-out/obsidian/graph.canvas in Obsidian's Canvas viewer
```

A community with, say, 25 members gets a box width sized for `ceil(sqrt(25)) = 5` columns (1100px wide), but the node cards are placed 3-per-row, occupying only ~600px of that 1100px width and wrapping into 9 rows instead of the 5 the box height was sized for. The visual result is a wide, half-empty box with a narrow, tall stack of cards on the left edge.

## Impact

Purely cosmetic (no data loss), but it makes the canvas effectively unreadable for any non-trivial community — which is the primary use case for `export obsidian`'s canvas output (visually browsing graph structure). Communities over ~12 members render as illegible, wrapped 3-wide strips inside oversized boxes.

## Suggested fix

Compute the column count once per community and reuse it for both box sizing and placement, e.g.:
```python
inner_cols = max(1, math.ceil(math.sqrt(n)))
...
col = m_idx % inner_cols
row = m_idx // inner_cols
```
and use that same `inner_cols` value in both the `w`/`h` box-sizing calc and the placement loop, rather than recomputing `sqrt(n)` in one place and hardcoding `3` in the other.

TPAteeq:
## Summary

Closes #1452.

`to_canvas()` sizes each community's group box for a `ceil(sqrt(n))`-column
grid (`w = 220 * ceil(sqrt(n))`), but the node-placement loop that fills the
box hardcoded exactly **3 columns** (`col = m_idx % 3`). So the cards span
only ~600px (3 cards wide) no matter how large the community, while the box
keeps widening with `sqrt(n)` — for any community larger than ~9 members the
result is a cramped 3-wide strip squeezed into the **left** of an over-wide,
mostly-empty box. (The height term was inconsistent in the other direction:
it used `ceil(n / 3)` while the width used `sqrt(n)`, so the two dimensions
didn't even agree on a column count — the height happened to track the
3-wide layout's rows, so it's the *width* that's left empty.)

Concretely, a 25-member community:

| | box size | card grid | result |
|---|---|---|---|
| before | 1100×1020 | 3 cols × 9 rows | ~600px strip in a 1100px-wide, half-empty box |
| after  | 1100×620  | 5 cols × 5 rows | cards fill the box in a ~square grid |

Purely cosmetic (no data loss), but it makes the canvas effectively
unreadable for any non-trivial community — the primary use case for the
canvas output is visually browsing graph structure.

## Fix

Compute the column count once per community
(`inner_cols = max(1, ceil(sqrt(n)))`) and reuse that single value for both
the box dimensions and the placement loop, instead of recomputing `sqrt(n)`
in one place and hardcoding `3` in the other:

- `w = max(600, 220 * inner_cols)`
- `h = max(400, 100 * ceil(n / inner_cols) + 120)` — now divides by the same
  `inner_cols` the width uses, so the two dimensions agree.
- `col = m_idx % inner_cols`, `row = m_idx // inner_cols`.

`inner_cols` floors at 1, so empty / single-node communities keep their
existing 600×400 minimum and there's no divide-by-zero. This matches the
reporter's suggested fix.

## Testing

- New regression test `test_to_canvas_node_grid_matches_box_columns` in
  `tests/test_export.py`, over a perfect-square (25 → 5×5) and a non-square
  (10 → 4 cols with a partial last row) member count. It pins the **column
  and row counts** (the load-bearing checks: distinct card x/y positions must
  equal `ceil(sqrt(n))` and `ceil(n / cols)`) and that every card sits inside
  its box. Both counts are red on the pre-fix code for both n
  (n=25: cols 3≠5, rows 9≠5), green after.
- Full `tests/test_export.py` (28) plus the canvas/obsidian-adjacent suites pass.
- Verified visually in Obsidian Canvas: a 25-node community now renders as a
  filled 5×5 grid inside a 1100×620 box instead of a 3-wide strip occupying
  ~600px of an 1100×1020 box.


safishamsi:
Thanks @TPAteeq — this is a real bug and the fix is right. The group box was sized for a `ceil(sqrt(n))`-column grid but the placement loop hardcoded 3 columns, so any community larger than ~9 members wrapped into a narrow 3-wide strip inside an over-wide, mostly-empty box.

Landed on `v8` in 7278e24 with your authorship preserved. I ported it onto current `v8` rather than merging the branch directly (it had drifted from the base), computing the column count once per community and reusing it for the box width, box height, and card placement so the cards fill the box. Added a regression test pinning the geometry (n=25 -> 5x5, n=10 -> 4 columns) and verified on a real canvas that every card lands inside its box.

Closing as merged-by-port. Appreciate the contribution.

PR #1460, opened 2026-06-25
feat: add XAML extraction support

MikeKatsoulakis:
Refs #1456.

## What

Adds basic WPF/XAML structural extraction without a new parser dependency.

This PR makes `.xaml` files first-class code inputs and extracts:

- the XAML file/root element
- named controls from `x:Name` / `Name`
- simple `{Binding ...}` references
- `x:Class` references
- event-handler references that resolve to methods on the matching `.xaml.cs` partial class

## Why

WPF projects keep important frontend structure in XAML, but graphify currently skips `.xaml` files. That means UI controls, bindings, view classes, and event handlers are invisible in the graph even though the matching C# code-behind is already extractable.

Resolving `x:Class` and event handlers gives the graph a useful bridge from the view markup to the C# partial class and methods without turning this into a full WPF analyzer.

## XAML coverage in this PR

Added now:

- `.xaml` file discovery and dispatch
- root element extraction (`Window`, `UserControl`, etc.)
- control nodes for elements with `x:Name` or `Name`
- control type references (`Button`, `TextBox`, `StackPanel`, etc.)
- simple `{Binding PropertyName}` references
- `x:Class="Namespace.Type"` links to the matching `.xaml.cs` partial class when present
- event-handler links from XAML attributes to methods on that partial class

Deferred deeper analysis:

- `StaticResource` / `DynamicResource` and resource dictionary resolution
- merged dictionaries, styles, templates, and control templates
- complex binding paths, converters, commands, and `DataContext` / ViewModel inference
- routed event bubbling semantics
- cross-file resolution between XAML views and separate ViewModel classes

## Changes

- `graphify/detect.py`: add `.xaml` to `CODE_EXTENSIONS`
- `graphify/extract.py`: add `extract_xaml()` and dispatch registration
- `tests/fixtures/sample.xaml` / `sample.xaml.cs`: add a small WPF fixture pair
- `tests/test_dotnet.py`: cover XAML class resolution, named controls, bindings, and event-handler links
- `tests/test_languages.py`: add the contribution-path XAML smoke test
- `README.md`: document `.xaml` in the supported file extensions table

## Testing

- `uv run pytest tests/test_dotnet.py tests/test_languages.py::test_xaml_finds_class_and_event_references -q`
- `uv run ruff check graphify tests --select E9,F63,F7,F82`

Full `uv run pytest tests/ -q` was attempted locally on Windows, but existing unrelated Windows/path and install/skillgen failures prevent a clean full-suite result on this machine.

### DIFF
--- a/README.md
+++ b/README.md
@@ -242,7 +242,7 @@ To remove graphify from all platforms at once: `graphify uninstall` (add `--purg
 
 | Type | Extensions |
 |------|-----------|
-| Code (36 tree-sitter grammars) | `.py .ts .js .jsx .tsx .mjs .go .rs .java .c .cpp .h .hpp .cu .cuh .rb .cs .kt .scala .php .swift .lua .luau .zig .ps1 .psm1 .ex .exs .m .mm .jl .vue .svelte .astro .groovy .gradle .dart .v .sv .svh .sql .f .f90 .f95 .f03 .f08 .pas .pp .dpr .dpk .lpr .inc .dfm .lfm .lpk .sh .bash .json .dm .dme .dmi .dmm .dmf .sln .slnx .csproj .fsproj .vbproj .razor .cshtml` (`.dm`/`.dme` requires `uv tool install graphifyy[dm]`; CUDA `.cu`/`.cuh` reuse the C++ grammar) |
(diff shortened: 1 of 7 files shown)


safishamsi:
Thanks @MikeKatsoulakis — this is a genuinely useful addition, and the code-behind bridge (resolving `x:Class` and event handlers to the matching `.xaml.cs` partial class) is the right idea. I especially liked that it scopes handler matching to the partial class's own `method` edges and reuses the existing XML safety guards (DOCTYPE/ENTITY rejection + size cap), so there's no new attack surface and no new parser dependency.

Landed on `v8` in 7dc5d96 with your authorship preserved.

One change I made on top before merging: tightened event-handler resolution so it can't create false edges. The original matched *any* attribute value against code-behind method names, so in real WPF a property like `Content="Save"` next to a business method `Save()`, or `Tag="..."`, would produce a spurious `event` edge (I could reproduce this). Resolution now requires three things: the attribute isn't a known free-form/identity property (`Content`, `Text`, `Tag`, `Title`, `ToolTip`, `Header`, …), the value is a bare identifier, and the matched method actually has the .NET event-handler signature `(object sender, …EventArgs e)` — read from the code-behind source, since the C# extractor doesn't record parameter lists on method nodes. Added regression tests for both false-positive cases.

Full suite (2417 passed) + skillgen guards green, and I verified the end-to-end bridge on the fixture (3 real events, `x:Class`, binding, all events resolving to code-behind methods).

The deferred items you listed (resource dictionaries, styles/templates, converters, ViewModel inference, cross-file view↔VM resolution) are good follow-ups if you want to keep going. Closing as merged-by-port — thanks for the contribution.

PR #1464, opened 2026-06-25
fix(hooks): match real file extension in Read|Glob hook (.astro miss, .json false-fire)

### LINKED ISSUE #1463: Read/Glob PreToolUse hook: substring ext match misses .astro and false-fires on .json
## Summary

The `Read|Glob` `PreToolUse` hook installed by `graphify claude install` (and `codebuddy install`) matches file extensions by **substring search** over the concatenated `file_path` + `pattern` + `path` string. This produces two opposite failures:

1. **False negative — `.astro` / `.vue` / `.svelte` never nudge.** These extensions aren't in the `exts` tuple, so on Astro/Vue/Svelte projects (where they're the primary source type) a `Read` or `Glob` of a component never triggers the graphify hint. Only the Bash/grep hook ends up covering those projects.
2. **False positive — `.json` matches `.js`.** The check is `any(e in s for e in exts)`, and the substring `.js` is contained in `.json`, so reading `package.json`, `tsconfig.json`, etc. fires the "run graphify first" nudge spuriously. More generally, any path containing a listed ext as a substring over-fires.

## Location

`graphify/__main__.py` → `_READ_SETTINGS_HOOK` (the inline `python3 -c` payload). Referenced by `_install_claude_hook` and the codebuddy installer, so both platforms inherit the bug. Reproduced on `graphifyy==0.8.49`.

The current payload reduces to:

```python
s = (file_path + ' ' + pattern + ' ' + path).lower().replace('\\','/')
hit = 'graphify-out/' not in s and any(e in s for e in exts)   # substring match
```

## Repro

```bash
# fires, but should be silent (.json contains .js)
echo '{"tool_input":{"file_path":"package.json"}}' | bash -c '<the installed Read|Glob hook command>'

# silent, but should fire (.astro not in exts)
echo '{"tool_input":{"pattern":"**/*.astro"}}'     | bash -c '<the installed Read|Glob hook command>'
```

## Proposed fix

Extract the real trailing extension per value (segment after the last `/`, then after the last `.`) instead of substring-searching, and add the framework extensions:

```python
import json,sys
d=json.load(sys.stdin); t=d.get('tool_input',d)
exts=('.py','.js','.ts','.tsx','.jsx','.astro','.vue','.svelte','.go','.rs','.java','.rb','.c','.h','.cpp','.hpp','.cc','.cs','.kt','.swift','.php','.scala','.lua','.sh','.md','.rst','.txt','.mdx')
vals=[str(t.get('file_path') or ''),str(t.get('pattern') or ''),str(t.get('path') or '')]
j=' '.join(vals).lower().replace(chr(92),'/')
tails=[('.'+x.rsplit('.',1)[-1]) for v in vals if v for x in [v.lower().replace(chr(92),'/').rsplit('/',1)[-1]] if '.' in x]
sys.stdout.write('1' if 'graphify-out/' not in j and any(tl in exts for tl in tails) else '')
```

`package.json` → tail `.json` (not in `exts`, silent); `**/*.astro` → tail `.astro` (matched, fires); the `graphify-out/` suppression is preserved and every branch still fails open.

## Verification (8/8 after fix)

| input | before | after | expected |
|---|---|---|---|
| Bash `grep -r …` | fire | fire | fire |
| Bash `npm run build` | silent | silent | silent |
| Read `src/x.ts` | fire | fire | fire |
| Glob `**/*.astro` | **silent (bug)** | fire | fire |
| Read `Hero.astro` | **silent (bug)** | fire | fire |
| Read `package.json` | **fire (bug)** | silent | silent |
| Read `graphify-out/GRAPH_REPORT.md` | silent | silent | silent |
| Read `netlify.toml` (no code ext) | silent | silent | silent |

## Environment

- `graphifyy` 0.8.49, Python 3.13, Windows 11.

Happy to open a PR with the above change if useful.

marketechniks:
Fixes #1463.

## Problem

The `Read|Glob` `PreToolUse` hook (`_READ_SETTINGS_HOOK` in `graphify/__main__.py`, used by both the Claude and CodeBuddy installers) decided whether to nudge by **substring-scanning** the joined `file_path` + `pattern` + `path` for known extensions:

```python
s = (file_path + ' ' + pattern + ' ' + path).lower().replace('\\','/')
hit = 'graphify-out/' not in s and any(e in s for e in exts)
```

Two opposite failures fall out of that:

- **False negative — `.astro` / `.vue` / `.svelte` never nudge.** They weren't in `exts`, so on Astro/Vue/Svelte projects (where they're the primary source type) reading or globbing a component never surfaced the graph. Only the Bash/grep hook covered those repos.
- **False positive — `.json` matches `.js`.** `'.js'` is a substring of `'.json'`, so reading `package.json`, `tsconfig.json`, etc. spuriously fired the "run graphify first" nudge. Any path containing a listed ext as a substring over-fires.

## Fix

Compare each value's **real trailing extension** (segment after the last `/`, then after the last `.`) against `exts`, and add the framework extensions:

```python
vals  = [file_path, pattern, path]
j     = ' '.join(vals).lower().replace('\\','/')
tails = [('.'+x.rsplit('.',1)[-1])
         for v in vals if v
         for x in [v.lower().replace('\\','/').rsplit('/',1)[-1]] if '.' in x]
hit   = 'graphify-out/' not in j and any(t in exts for t in tails)
```

`package.json` → tail `.json` (not in `exts`, silent); `**/*.astro` → tail `.astro` (matched, fires). The `graphify-out/` suppression and the fail-open behaviour are unchanged.

## Tests

Added regression tests in `tests/test_read_hook.py`:
- `test_nudges_on_framework_source` — `.astro` / `.vue` / `.svelte` reads nudge
- `test_astro_glob_nudges` — `**/*.astro` glob nudges
- `test_silent_on_json_config` — `package.json` / `tsconfig.json` / `data.geojson` stay silent

All existing cases still pass:

```
pytest tests/test_read_hook.py tests/test_install_strings.py tests/test_codebuddy.py
42 passed
```

### DIFF
--- a/graphify/__main__.py
+++ b/graphify/__main__.py
@@ -405,6 +405,10 @@ _READ_SETTINGS_HOOK = {
     # python3 (already a graphify dependency), the shell is POSIX, and every branch
     # fails open, so a legitimate read always goes through. Reading the graph's own
     # report under graphify-out/ is suppressed so it never starts a feedback loop.
+    # The extension test compares each value's real trailing extension (segment
+    # after the last '/' then after the last '.') against exts -- not a substring
+    # scan, which both missed framework files like .astro and false-matched .json
+    # against .js (the substring '.js' is inside '.json').
     "matcher": "Read|Glob",
     "hooks": [
         {
@@ -414,9 +418,11 @@ _READ_SETTINGS_HOOK = {
                 "import json,sys;"
                 "d=json.load(sys.stdin);"
                 "t=d.get('tool_input',d);"
-                "s=(str(t.get('file_path') or '')+' '+str(t.get('pattern') or '')+' '+str(t.get('path') or '')).lower().replace(chr(92),'/');"
-                "exts=('.py','.js','.ts','.tsx','.jsx','.go','.rs','.java','.rb','.c','.h','.cpp','.hpp','.cc','.cs','.kt','.swift','.php','.scala','.lua','.sh','.md','.rst','.txt','.mdx');"
-                "sys.stdout.write('1' if 'graphify-out/' not in s and any(e in s for e in exts) else '')\" 2>/dev/null || true); "
+                "exts=('.py','.js','.ts','.tsx','.jsx','.astro','.vue','.svelte','.go','.rs','.java','.rb','.c','.h','.cpp','.hpp','.cc','.cs','.kt','.swift','.php','.scala','.lua','.sh','.md','.rst','.txt','.mdx');"
+                "vals=[str(t.get('file_path') or ''),str(t.get('pattern') or ''),str(t.get('path') or '')];"
+                "j=' '.join(vals).lower().replace(chr(92),'/');"
+                "tails=[('.'+x.rsplit('.',1)[-1]) for v in vals if v for x in [v.lower().replace(chr(92),'/').rsplit('/',1)[-1]] if '.' in x];"
+                "sys.stdout.write('1' if 'graphify-out/' not in j and any(tl in exts for tl in tails) else '')\" 2>/dev/null || true); "
                 "if [ \"$HIT\" = 1 ] && [ -f graphify-out/graph.json ]; then "
                 r"""echo '{"hookSpecificOutput":{"hookEventName":"PreToolUse","additionalContext":"MANDATORY: graphify-out/graph.json exists. You MUST run graphify before reading source files. Use: `graphify query \"<question>\"` (scoped subgraph), `graphify explain \"<concept>\"`, or `graphify path \"<A>\" \"<B>\"`. Only read raw files after graphify has oriented you, or to modify/debug specific lines. This rule applies to subagents too — include it in every subagent prompt involving code exploration."}}'; """
                 "fi || true"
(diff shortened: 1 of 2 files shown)


safishamsi:
Thanks @marketechniks — clean diagnosis and the right fix. Substring-matching the path was the root of both failures (the `.js`⊂`.json` false-fire and the missing `.astro`/`.vue`/`.svelte`), and comparing the real trailing extension (last `/` segment, then last `.`) is exactly the correct fix — it even handles the directory-extension trap (`my.ts/file` stays silent) for free.

Landed on `v8` in 9b583a0 with your authorship preserved. I added three regression tests on top of yours to pin the trickiest parts of the segment-split logic (multi-dot `a.test.tsx`/`foo.min.js`, a Windows backslash path, and the directory-extension case).

Verified end-to-end by running the actual hook command the way Claude Code invokes it (`sh -c` + stdin JSON): `package.json`/`tsconfig.json`/`data.geojson` stay silent, `.astro`/`.vue`/`.svelte` and `**/*.astro` fire, normal source still fires, and `graphify-out/` stays suppressed. Full suite 2424 passed. Closing as merged-by-port — thanks for the contribution.

PR #1465, opened 2026-06-25
fix(wiki): portable relative markdown links so wiki navigation works outside Obsidian

### LINKED ISSUE #1444: Bug: `graphify export wiki` emits Obsidian `[[wikilinks]]` that break in every non-Obsidian renderer
Bug: `graphify export wiki` emits Obsidian `[[wikilinks]]` that break in every non-Obsidian renderer

## Summary
`graphify export wiki` writes `index.md` and article files using Obsidian-style
`[[Title]]` wikilinks, but writes the article **filenames** with spaces replaced
by underscores (and other chars sanitized). Obsidian resolves `[[Title]]` by note
title internally, so it works there — but in VS Code preview, GitHub, GitLab, or a
plain browser, `[[Domain Data Models]]` resolves literally to `Domain Data Models.md`,
which does not exist (the file is `Domain_Data_Models.md`). Result: nearly every
navigation link in the generated wiki opens an empty page.

## Reproduction
1. `graphify <any-corpus> --wiki` (or `graphify export wiki` on an existing graph)
2. Open `graphify-out/wiki/index.md` in VS Code preview / GitHub / a browser (not Obsidian)
3. Click any community or god-node link, e.g. `[[Domain Data Models]]`
4. Lands on an empty/non-existent page

## Expected
Links resolve in any markdown renderer, Obsidian included.

## Actual
- Link text:   `[[Domain Data Models]]`
- Actual file: `Domain_Data_Models.md`
- These reconcile only inside Obsidian.

In one real run: index + 38 articles contained 206 `[[...]]` links. 108 were
community/god-node navigation links (all broken outside Obsidian). The other 98
were node-level links such as `[[AwsHelper.py]]` and `[[.read_object_key_content_from_s3()]]`
— these point at nodes for which the exporter **never writes an article file at all**,
so they are dead even inside Obsidian.

## Two distinct defects
1. **Link/filename convention mismatch.** Emit standard relative markdown links
   `[Display](Domain_Data_Models.md)` (the exact filename, URL-encoded or `<...>`-wrapped
   if it contains spaces/parens/`&`). These work everywhere, Obsidian included.
   Alternatively, gate `[[...]]` output behind an `--obsidian`-style flag.
2. **Links to non-existent articles.** The exporter links to individual nodes
   (`[[AwsHelper.py]]`, `[[.method()]]`) but only generates article files for
   communities and god nodes. Either generate those article files, or render
   non-article nodes as plain text instead of dead links.

## Environment
- graphify (PyPI package `graphifyy`), installed via `uv tool`
- Robot Framework / Python test corpus, ~69 files
- Observed on Linux; renderer-agnostic (any non-Obsidian markdown viewer)

## Workaround in use
A post-export normalizer rewrites `[[Title]]` → `[Title](Filename.md)` using a
resolver built from each file's H1 title + filename stem (with `_`↔space variants),
wrapping targets containing spaces/`&`/parens in `<...>`. Links with no matching
article file are downgraded to plain text. Idempotent. Happy to contribute it as a PR.

TPAteeq:
Closes #1444.

`graphify export wiki` (`to_wiki`) wrote articles with Obsidian-style `[[Title]]`
wikilinks. Two defects made nearly every navigation link dead outside Obsidian:

1. **Convention mismatch.** `[[Title]]` resolves by note title only inside
   Obsidian. Everywhere else — VS Code preview, GitHub, GitLab, a plain browser —
   `[[Domain Data Models]]` resolves to a literal `Domain Data Models.md`, but the
   article file is `Domain_Data_Models.md` (the slug replaces spaces with
   underscores and substitutes reserved characters). The reporter measured one real
   run: index + 38 articles, 206 `[[...]]` links, 108 of them community/god-node
   navigation links — all broken outside Obsidian.
2. **Dead links.** God-node articles linked every neighbor (`[[AwsHelper.py]]`,
   `[[.read_object_key()]]`), but only communities and god nodes get article files —
   so those ~98 node-level links were dead even inside Obsidian.

## Fix

Emit standard relative markdown links `[display](slug.md)` instead of `[[...]]`,
with the target URL-encoded via `urllib.parse.quote` so spaces, `&`, parentheses,
and `#` survive intact in every CommonMark renderer and in Obsidian (which resolves
standard relative links and still shows them in graph view / backlinks). Links
whose label has no article render as plain text instead of a dangling link.

To make the link target always match the file actually written, `to_wiki` now
assigns every article its slug in a first pass (building a `label -> slug`
resolver) before rendering any body — so a link tracks the on-disk filename
including any numeric collision suffix (`parser_2.md`). Slug-assignment order
(communities, then god nodes) is unchanged, so the case-fold collision dedup keeps
producing identical filenames.

This is the portable-links option from the issue; I left the alternative
`--obsidian-style` flag out to keep the change focused — standard links already
work in Obsidian, so nothing is lost there, and a flag is easy to add later if
anyone wants `[[...]]` back.

## Testing

- Updated the existing `test_wiki.py` cases that asserted the old `[[...]]` form to
  assert the new `[display](slug.md)` form.
- New regression tests: no file contains `[[...]]`; every inline link target across
  the whole wiki resolves to a real file on disk; display text keeps the human
  label while the target is the underscored filename; special characters (space,
  `&`, `#`, parens) round-trip through URL-encoding to the real file; node-level
  links to article-less neighbors are plain text; and links to a collision-suffixed
  article target the suffixed slug. All red on the pre-fix code, green after.
- Full suite: 2423 passed, 2 skipped.

## Out of scope

Two pre-existing label-ambiguity edge cases (two communities with an *identical*
label collapse to one link target and orphan the second article; a community
literally labelled `index` collides with `index.md`) are not introduced or worsened
here — the old `[[...]]` form was equally ambiguous — so I left them for a separate
change to keep this PR scoped to the reported bug.


safishamsi:
Thanks @TPAteeq — this is the correct fix for #1444 (standard `[display](slug.md)` links, URL-encoded, resolving in every renderer). Landed on `v8` in ec5bb33 with your authorship. Minor maintainer edit: trimmed the CHANGELOG wording that said the target "always" matches the written file — two articles with a byte-identical label still keep the first slug (pre-existing, same as the old `[[label]]`), so I scoped the claim to the case-fold collision case. Verified end-to-end: no `[[..]]` emitted, links URL-encode and resolve to real files. Closing as merged-by-port.

PR #1468, opened 2026-06-25
fix(extract): parse .vue SFC <script> with the right grammar

papinto:
## Summary

`.vue` Single File Components were extracting almost nothing because the whole file was being parsed with the wrong grammar.

`.vue` is dispatched to `extract_js`, which picks a tree-sitter grammar by file suffix. Since `.vue` is neither `.ts` nor `.tsx`, it fell through to the JavaScript grammar, and the **entire** SFC — `<template>` markup, `<script>`, and `<style>` — was handed to that grammar. Markup isn't valid JS, so the parse produced a top-level ERROR node and no imports, symbols, or type references were recovered.

This adds a dedicated `extract_vue` that parses only the `<script>` block, with the grammar its `lang` implies.

## Impact

Measured on a 300-file random sample from a production Vue 3 + Vite + TypeScript app:

| Metric | Before | After |
|---|---|---|
| Files extracting nothing (0 edges) | 40% | **6%** |
| Files with no import edges | 69% | **8%** |
| Total edges | 2,834 | **8,882** (3.1×) |

The residual ~6% are genuinely template-only components (no `<script>`), which correctly produce just a file node.

## How it works

1. **Mask, don't strip.** Every character outside a `<script>` body — template, style, and the `<script …>`/`</script>` tags themselves — is replaced with a space, while newlines are preserved. The surviving script therefore sits at its original line numbers, and the blanked regions parse as empty. No line-offset bookkeeping required.
2. **Pick the grammar from `lang`.** `tsx` → TSX, `js`/`jsx` → JS, and `ts` or unspecified → TS. TypeScript is a strict superset of JavaScript, so defaulting an unannotated block to the TS grammar is lossless and matches the dominant Vue 3 + `<script setup>` convention.
3. **Recover dynamic imports.** A small regex pass picks up `import('...')` calls (e.g. `defineAsyncComponent(() => import('./X.vue'))`) that the AST pass does not emit as edges — the same rescue `extract_svelte`/`extract_astro` already do.

This goes a step further than the Svelte/Astro extractors, which regex-scrape import strings only: because Vue's component logic lives in `<script>`, parsing it with a real grammar recovers the full symbol and type-reference graph (classes, functions, typed props referencing imported types), not just imports.

`.vue` files also now join the JS cross-file symbol-resolution pass. They were previously excluded from it precisely because the whole-file parse failed; with a working per-file parse, an SFC's calls resolve to definitions in other files like any `.ts` file's would.

## Implementation notes

- `_extract_generic` gains an opt-in keyword-only `source_override: bytes | None = None`. When provided, it parses those bytes instead of reading the file, while still keying every node/edge off the real path. Default-off, so no existing caller changes behavior.
- The one-site `and path.suffix != ".vue"` guard in `_collect_js_symbol_resolution_facts` is removed. `.vue` stays in the JS-family suffix set used elsewhere for cache-bypass.

## Tests

`tests/test_vue_extraction.py` (11 cases): masking + line-number preservation, `<script setup lang="ts">` static imports, symbol extraction with correct lines, typed props referencing an imported type, dual `<script>` + `<script setup>` blocks, dynamic-import recovery, plain-JS blocks, template-only files, and end-to-end cross-file call resolution.

The JS/TS import-resolution, Svelte/Astro, and symbol-resolution suites remain green.

### DIFF
--- a/graphify/extract.py
+++ b/graphify/extract.py
@@ -2343,8 +2343,15 @@ _SWIFT_CONFIG = LanguageConfig(
 
 # ── Generic extractor ─────────────────────────────────────────────────────────
 
-def _extract_generic(path: Path, config: LanguageConfig) -> dict:
-    """Generic AST extractor driven by LanguageConfig."""
+def _extract_generic(
+    path: Path, config: LanguageConfig, *, source_override: bytes | None = None
+) -> dict:
+    """Generic AST extractor driven by LanguageConfig.
+
+    ``source_override`` parses the given bytes instead of reading ``path``, while
+    still keying nodes/edges off ``path``. Lets container formats (e.g. Vue SFCs)
+    mask the wrapper and parse just the embedded ``<script>``.
+    """
     try:
         mod = importlib.import_module(config.ts_module)
         from tree_sitter import Language, Parser
@@ -2371,7 +2378,7 @@ def _extract_generic(path: Path, config: LanguageConfig) -> dict:
 
     try:
         parser = Parser(language)
-        source = path.read_bytes()
+        source = path.read_bytes() if source_override is None else source_override
         tree = parser.parse(source)
         root = tree.root_node
     except Exception as e:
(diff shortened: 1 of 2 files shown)


MentorFilou:
The same issue was similarly addressed in #1141 . Sadly the author didnt respond for 3 weeks.

While yours seems to have better test coverage, @safishamsi suggested that reusing the svelte function is the right call, even asking to not create a distinct vue wrapper function! Consider reading their response on that PR to evaluate possibly changing this here.

Hoping to see this merged soon 👍

safishamsi:
Thanks @papinto — great catch and the right fix. Feeding the whole SFC to the JS grammar produced a top-level ERROR and dropped everything; masking the non-`<script>` regions with spaces (preserving newlines so line numbers stay accurate) and parsing the script with the `lang`-named grammar is exactly the correct approach, and wiring `.vue` into cross-file symbol resolution is a nice bonus.

Landed on `v8` in 349465b with your authorship preserved. One maintainer fix on top: the `<script>` open-tag scan now skips over quoted attribute values, so a `>` inside one — Vue 3.3+ generic components like `generic="T extends Record<string, unknown>"` — no longer ends the tag early and swallows the body. Added a regression test for it. Verified end-to-end: a `<script setup lang="ts">` SFC now recovers its imports and symbols with no ERROR node. Closing as merged-by-port — thanks for the contribution!

PR #1473, opened 2026-06-26
feat(extract): link XAML ViewModels and bindings

MikeKatsoulakis:
## What

Adds deeper WPF/XAML graph extraction on top of the initial XAML support.

This branch resolves XAML views to ViewModels from:

- explicit `<Window.DataContext><vm:MainViewModel /></Window.DataContext>` / equivalent `DataContext` elements
- design-time `d:DataContext="{d:DesignInstance Type=...}"`
- naming conventions like `SettingsView` -> `SettingsViewModel`
- Prism `ViewModelLocator.AutoWireViewModel="True"`

It also extracts richer binding references:

- binding paths: `{Binding User.Name}`, `{Binding Path=Order.Total}`
- commands: `Command="{Binding SaveCommand}"`
- converters: `Converter={StaticResource MoneyConverter}`
- direct `<Binding Path="..." Converter="..." />`

This adds support for common WPF MVVM patterns/frameworks:

- plain WPF/MVVM `DataContext` patterns
- design-time ViewModel hints
- naming-convention ViewModel lookup
- Prism AutoWire ViewModel conventions
- CommunityToolkit.Mvvm generated members:
  - `[ObservableProperty] private string userName;` -> inferred `UserName`
  - `[ObservableProperty]` on the line above a field -> inferred generated property
  - `[RelayCommand] private void Save()` -> inferred `SaveCommand`
  - `[RelayCommand] private async Task SaveAsync()` -> inferred `SaveCommand`
- EventToCommand-style command binding via `InvokeCommandAction Command="{Binding SaveCommand}"`

Inference is conservative:

- only links a ViewModel when exactly one matching C# class is found
- skips ambiguous ViewModel candidates
- marks convention/generated links as `INFERRED`
- does not infer commands from plain methods without `[RelayCommand]`
- respects `.gitignore` / `.graphifyignore`
- keeps ViewModel `.cs` scans inside the active extraction root

## Why

WPF apps often split UI behavior across `.xaml`, `.xaml.cs`, ViewModels, binding paths, commands, converters, and MVVM framework conventions. Without these links, graphify can see the files but misses the relationships that explain how UI flows into application code.

This makes WPF/XAML more useful in the graph:

- views connect to their real ViewModels
- buttons/commands connect to command members
- Toolkit-generated properties and commands are represented
- Prism AutoWire conventions are understood
- bindings and converters become searchable graph references

## Scope

This is WPF/XAML-oriented support.

WinUI, MAUI, Avalonia, and other XAML-based stacks may get some collateral benefit where their syntax overlaps with WPF/XAML, but they are not the focus of this PR or the previous WPF PR. I am not a WinUI/MAUI expert, so I would not want to claim direct support unless someone specifically asks for it and provides real examples.

If there are issues, bugs, or additions needed for further WPF support, please tag me: @MikeKatsoulakis.

## Tests

Added focused fixtures and tests for:

- explicit ViewModel `DataContext`
- design-time ViewModel `DataContext`
- naming inference
- Prism AutoWire true/false behavior
- ambiguous ViewModel candidates
- CommunityToolkit generated members, including same-line and next-line attributes
- EventToCommand command bindings
- binding paths, converters, and commands
- extraction ID remapping
- `.graphifyignore` behavior
- cache-root scan boundary behavior
- false-positive cases for Toolkit comments and non-event XAML attributes

Validated with:

```bash
uv run --frozen pytest tests/test_dotnet.py -q
uv run --frozen pytest tests/test_languages.py -k "sln or csproj or xaml or razor" -q
uv run --frozen ruff check graphify\extract.py tests\test_dotnet.py
git diff --check
```

Local full-suite note: `uv run --frozen pytest tests/ -q` still fails on this Windows checkout with existing platform/path/skillgen baseline failures; the targeted .NET/XAML checks above pass.

## Security Review

A security pass found that ViewModel lookup could scan `.cs` files outside the requested extraction boundary and repeat scans per XAML file. This branch fixes that by bounding lookup to the active extraction root, respecting ignore files, and caching the class scan during extraction.

No XML entity expansion, regex DoS, or graph-label injection issue was found.

### DIFF
--- a/graphify/extract.py
+++ b/graphify/extract.py
@@ -11060,6 +11060,84 @@ _XAML_NON_EVENT_ATTRS = frozenset({
 # markup, a path, or a sentence. Used to skip values like "{Binding ...}" or
 # free-form content before looking them up as code-behind methods.
 _XAML_IDENT_RE = re.compile(r"[A-Za-z_]\w*")
+_XAML_DESIGN_INSTANCE_TYPE_RE = re.compile(
+    r"\bType\s*=\s*(?:\{x:Type\s+)?(?P<type>[\w.:+]+)"
+)
+
+
+def _xaml_markup_extension(value: str) -> tuple[str, str] | None:
+    value = value.strip()
+    if not (value.startswith("{") and value.endswith("}")):
+        return None
+    inner = value[1:-1].strip()
+    if not inner or inner.startswith("}"):
+        return None
+    name, _, args = inner.partition(" ")
+    return name, args.strip()
+
+
+def _xaml_split_markup_args(args: str) -> list[str]:
+    parts: list[str] = []
+    start = 0
+    depth = 0
+    for idx, ch in enumerate(args):
+        if ch == "{":
+            depth += 1
+        elif ch == "}" and depth:
+            depth -= 1
+        elif ch == "," and depth == 0:
+            parts.append(args[start:idx].strip())
+            start = idx + 1
+    tail = args[start:].strip()
+    if tail:
+        parts.append(tail)
+    return parts
+
+
+def _xaml_static_resource_key(value: str) -> str | None:
+    markup = _xaml_markup_extension(value)
+    if not markup:
+        return None
+    name, args = markup
+    if name != "StaticResource":
+        return None
+    for part in _xaml_split_markup_args(args):
+        if "=" not in part:
+            return part.strip() or None
+        key, resource = part.split("=", 1)
+        if key.strip() == "ResourceKey":
+            return resource.strip() or None
+    return None
+
+
+def _xaml_binding_refs(value: str) -> tuple[str | None, str | None]:
+    markup = _xaml_markup_extension(value)
+    if not markup:
+        return None, None
+    name, args = markup
+    if name != "Binding":
+        return None, None
+
+    path_ref = None
+    converter_ref = None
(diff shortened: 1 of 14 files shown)


safishamsi:
Thanks @MikeKatsoulakis — strong follow-up to #1460. The ViewModel resolution is exactly the kind of careful work I hoped for: it resolves against actually-extracted C# class nodes and refuses to emit an edge when the name matches no class or an ambiguous 2+, with EXTRACTED for explicit DataContext and INFERRED for naming/Prism conventions. The binding-path/command/converter extraction and CommunityToolkit member synthesis are clean, and — importantly — the #1460 event-handler signature gate and the free-form-attribute denylist are preserved unchanged (both regression tests still pass).

Landed on `v8` in 905e0a7 with your authorship preserved (clean 3-way merge onto current v8). One maintainer fix on top: the CommunityToolkit member reader now reads the code-behind with `errors="replace"`, so a non-UTF8 ViewModel `.cs` can't raise `UnicodeDecodeError` and abort `extract_xaml` (every other reader in the module already does this). Added a regression test. Verified end-to-end: explicit `DataContext` resolves to the real VM (EXTRACTED) and the `View`->`ViewModel` convention resolves (INFERRED). Full suite 2451 passed. Closing as merged-by-port — thanks!

Issue #1402, opened 2026-06-19
Cross-file type-annotation refs create phantom duplicate nodes (sourced stub blocks _rewire_unique_stub_nodes)


ZedUserdesign:
## Summary

Cross-file type-annotation (and other bare-name) references create **phantom duplicate nodes** instead of resolving to the single canonical definition. A class defined once but referenced via type annotations in N other files appears as **1 + N nodes** in the graph, inflating node counts and god-node/centrality rankings, and splitting edges across duplicates.

Observed on `graphifyy==0.8.44` (AST/structural extraction, no LLM key set), Python, but the faulty code path is shared across all 6 language extractors.

## Minimal reproduction

```
pkg/thing.py
  class Thing:
      def run(self):
          return 1

pkg/a.py
  from pkg.thing import Thing
  def use_a(obj: Thing) -> Thing:
      return obj

pkg/b.py
  from pkg.thing import Thing
  def use_b(obj: Thing) -> Thing:
      return obj
```

```python
from graphify.extract import extract
from pathlib import Path
r = extract([Path('pkg/thing.py'), Path('pkg/a.py'), Path('pkg/b.py')], cache_root=Path('.'))
print(sorted(n['id'] for n in r['nodes'] if n['label'] == 'Thing'))
```

**Actual (0.8.44):**
```
['pkg_a_py_thing', 'pkg_b_py_thing', 'pkg_thing_thing']   # 3 nodes
```
Note `pkg_a_py_thing` / `pkg_b_py_thing` — the file path **with extension** baked into the id, the signature of the bug.

**Expected:**
```
['pkg_thing_thing']   # 1 canonical node; the annotation refs become edges to it
```

## Root cause

`ensure_named_node` (defined identically as a nested closure in all 6 per-language extractors) builds the bare cross-file reference node via `add_node`, which always stamps `source_file = str_path` (the **using** file):

```python
def ensure_named_node(name: str, line: int) -> str:
    nid = _make_id(stem, name)
    if nid in seen_ids:
        return nid
    nid = _make_id(name)
    if nid not in seen_ids:
        add_node(nid, name, line)   # <-- stamps source_file = using file
    return nid
```

A *sourced* stub looks like a real definition, which breaks the two passes that are supposed to clean it up:

[...]

PR #1479, opened 2026-06-27
fix(extract): disambiguate imported symbol stubs

### LINKED ISSUE #1462: Bug: AST extractor produces duplicate node IDs for same-name imported symbols (collision, not fragmentation)
**Graphify version:** v0.8.47
**Repository:** https://git.eisen5.eu/xyz00777/ansible-aom/src/branch/feat/nom-compact-renderer
**Project:** Python (166 source files, 11K+ LoC)

### What happens

The AST extractor creates nodes for imported symbols using `_make_id(symbol_name)` without source-file context. When multiple files import the same standard-library type, they all get the same node ID:

| Symbol | Occurrences | Import source |
|--------|-------------|---------------|
| `path` | 60 | `from pathlib import Path` |
| `any` | 16 | `from typing import Any` |
| `datetime` | 10 | `import datetime` |
| `monkeypatch` | 10 | `monkeypatch` fixture param |
| others | 1-6 each | various stdlib |

Total: **21 colliding IDs** across **122 nodes**.

### Reproducer

```bash
cd /opt/syncthing/sync/ncc1031/git/ansible-aom
uv run python3 -c "
import json
from collections import Counter
d = json.load(open('graphify-out/.graphify_graph.json'))
ids = [n['id'] for n in d['nodes']]
dupes = {k:v for k,v in Counter(ids).items() if v > 1}
print(f'{len(dupes)} duplicate IDs:')
for k,v in sorted(dupes.items(), key=lambda x:-x[1])[:10]:
    print(f'  {k!r}: {v} occurrences')
"
```

### This is NOT #1318

- **#1318** (fragmentation): One symbol → different IDs. A class `Foo` defined in `a.java` gets ID `a_foo`, but when referenced from `b.java` it gets `b_foo`. The entity is *split*.
- **Our bug** (collision): Different symbols → same ID. 60 `path` nodes all get `id: "path"`. Entities are *squashed*.

Opposite failure modes. Same root cause (IDs lack source context), but different symptoms.

### This IS a variant of #438

**#438** (FIXED, v0.4.x): "Node ID collisions when multiple files share the same stem" — files like `cli.py` in different dirs all got `cli_main`. Fix: added parent-directory prefix to file node IDs.

Our bug is **#438 for imported symbols instead of filenames**. The fix for #438 only touched file-level nodes; imported-symbol nodes still use bare `_make_id(name)`.

### Impact

- `networkx.Graph` treats 60 `path` nodes as one → single fake hub with degree 60+, distorts centrality
- `vis.DataSet` in HTML export silently drops 59 of 60 → edges point to phantom nodes
- Community detection inflates: the fake hub bridges unrelated communities that never actually touch

### Suggested fix

**Option A**: Disambiguate imported-symbol IDs by `source_file`, same way #438 fixed file nodes:

```python
_make_id(source_file_parent, source_file_stem, symbol_name)
```

So `src/parser.py` doing `from pathlib import Path` → `src_parser_path`, not `path`.

**Option B** (complementary): Add a noise filter for stdlib-only imports (like `_PYTHON_ANNOTATION_NOISE` in `extract.py`) so `pathlib.Path`, `typing.Any`, `datetime`, `os`, `sys`, `re`, `collections`, etc. are not extracted as nodes at all. These carry zero project-specific meaning and inflate every graph with no benefit.

### Workaround used

Post-processing in the HTML generator: detect colliding IDs, append `_1`, `_2` suffixes, rewire edges. Band-aid only; the `graph.json` still has the collision. The `graph.html` renders correctly with this fix.

jiangyq9:
## Summary
- preserve the referencing file for generic imported symbol stubs via an internal origin_file field
- use that origin when duplicate node IDs need source-based disambiguation
- add regression coverage for repeated imported Path type stubs across files

Fixes #1462

## Testing
- python -m pytest tests/test_extract.py -k "imported_type_stubs or cross_file_type_annotation_refs or duplicate_symbol_ids"

### DIFF
--- a/graphify/extract.py
+++ b/graphify/extract.py
@@ -2461,6 +2461,7 @@ def _extract_generic(
                 "file_type": "code",
                 "source_file": "",
                 "source_location": "",
+                "origin_file": str_path,
             })
         return nid
 
@@ -7517,6 +7518,13 @@ def _source_key(source_file: str, root: Path) -> str:
         return str(source_path)
 
 
+def _node_disambiguation_source_key(node: dict, root: Path) -> str:
+    source_file = str(node.get("source_file", ""))
+    if source_file:
+        return _source_key(source_file, root)
+    return _source_key(str(node.get("origin_file", "")), root)
+
+
 def _disambiguate_colliding_node_ids(
     nodes: list[dict],
     edges: list[dict],
@@ -7542,12 +7550,12 @@ def _disambiguate_colliding_node_ids(
     remap: dict[tuple[str, str], str] = {}
     ambiguous_ids: set[str] = set()
     for old_id, group in by_id.items():
-        source_keys = {_source_key(str(node.get("source_file", "")), root) for node in group}
+        source_keys = {_node_disambiguation_source_key(node, root) for node in group}
         if len(group) < 2 or len(source_keys) < 2:
             continue
         ambiguous_ids.add(old_id)
         for node in group:
-            source_key = _source_key(str(node.get("source_file", "")), root)
+            source_key = _node_disambiguation_source_key(node, root)
             if not source_key:
                 continue
             new_id = _make_id(source_key, old_id)
--- a/tests/test_extract.py
+++ b/tests/test_extract.py
@@ -123,6 +123,25 @@ def test_cross_file_type_annotation_refs_resolve_to_single_node(tmp_path):
     assert "_py" not in thing_nodes[0]["id"], thing_nodes[0]["id"]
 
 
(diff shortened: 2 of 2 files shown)


safishamsi:
Thanks @jiangyq9 — landed on `v8` in 6509d0c with your authorship. The change is surgical (an internal `origin_file` used only as a disambiguation key); I verified the key property that it does NOT block the #1402 rewire — a type with a real project definition still collapses to one node, while two unrelated imported uses with no def stay distinct. Squash-merged onto current v8 (the branch's stale base made the raw diff look like it reverted later features; the 3-way merge applied only the real change). Closing as merged-by-port.

PR #1450, opened 2026-06-24
feat: extract Metal (.metal) kernels into the graph

GoodOlClint:
## What

Adds Metal Shading Language (`.metal`) support to graphify by routing it through
the **existing C++ extractor** — the exact reuse pattern CUDA's `.cu`/`.cuh`
already follow. MSL is C++14, so `extract_cpp` (tree-sitter-cpp) handles it with
no new grammar. ERROR nodes around Metal-specific qualifiers (`kernel`,
`[[buffer(0)]]`) are expected and harmless, just as they are for CUDA's
`__global__`/`__device__`.

## Why

Before this change graphify classified `.metal` as non-code and skipped it
entirely, so MLX/Metal compute kernels never appeared in the graph. This lets
them join it like any other source.

## Approach

Mirror CUDA. No dedicated Metal grammar — deferred. ~3-line change across the
same three touchpoints CUDA uses.

## Edits (3 touchpoints)

1. `graphify/detect.py` — `'.metal'` added to `CODE_EXTENSIONS` (next to `.cu`/`.cuh`).
2. `graphify/extract.py` — `".metal": extract_cpp` in the extension→extractor dict.
3. `graphify/build.py` — `".metal": "cpp"` in `_LANG_FAMILY`, so Metal↔C++ count
   as one family for cross-language seam / import resolution.

Plus a contract test + `tests/fixtures/sample.metal` (a real MSL snippet: a
`struct` + a `dot3` function + a `saxpy` kernel).

## Evidence (before → after)

**Contract test** (`tests/test_languages.py`, mirrors the cpp/CUDA tests):
`extract_cpp(FIXTURES / "sample.metal")` — `error` absent, `Vec3` struct and
`dot3` function both appear in nodes. Symbols graphify produced nothing for before.

**E2e** — `GRAPHIFY_OUT=<tmp> graphify update <dir-with-.metal>` on a dir
containing one `kernel.metal`:

| | graph.json | metal-source nodes |
|---|---|---|
| **before** (origin/v8) | not created — `No code files found - nothing to rebuild` | **0** (skipped) |
| **after** (this branch) | `Rebuilt: 11 nodes, 10 edges, 4 communities` | **7** |

After-state metal nodes: `kernel.metal`, `Vec3`, `x`, `y`, `z`, `dot3()`, `saxpy()`.

**Test gate** — `uv run pytest` (full suite): **2351 passed, 28 skipped**
(includes 3 new `test_metal_*`). `ruff check` on changed files: clean.

## Notes

[...]

PR #1480, opened 2026-06-27
feat(extract): index Metal shader files

jiangyq9:
## Summary
- classify .metal files as code so they are included in scans and watches
- route .metal through the existing C++ extractor, matching the CUDA reuse pattern
- document Metal support and add a fixture-based regression test for a struct, helper function, and kernel

Fixes #1450

## Testing
- python -m pytest tests/test_languages.py -k "metal or cuda or cpp"

### DIFF
--- a/README.md
+++ b/README.md
@@ -242,7 +242,7 @@ To remove graphify from all platforms at once: `graphify uninstall` (add `--purg
 
 | Type | Extensions |
 |------|-----------|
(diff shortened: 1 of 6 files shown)


safishamsi:
Thanks @jiangyq9 — landed on `v8` in 0b33511 with your authorship. This was the more complete of two Metal PRs: it also closes the pre-existing `_LANG_FAMILY` gap for `.cu`/`.cuh`/`.metal` and updates the README, and merged clean against current v8. Credit to @GoodOlClint (#1450) for the original `.metal`→C++ approach and fixture. Closing as merged-by-port.

PR #1481, opened 2026-06-27
feat(label): add missing-only relabeling

jiangyq9:
## Summary
- add a graphify label --missing-only flag for incremental community naming
- preserve existing non-placeholder labels from .graphify_labels.json
- send only missing or Community N placeholder communities through the existing labeling backend
- cover the CLI path with a regression test

Fixes #1421

## Testing
- python -m pytest tests/test_labeling.py


safishamsi:
Thanks @jiangyq9 — landed on `v8` in 383cabd with your authorship. Written against the current label signature (post-#1390 concurrency/batch params) and merges clean. Credit to @matiasduartee (#1421) who proposed the same `--missing-only` flag first. Closing as merged-by-port.

PR #1482, opened 2026-06-27
fix(llm): force non-streaming OpenAI-compatible calls

### LINKED ISSUE #1223: OpenAI-compatible backends: allow forcing `stream:false` (gateways that stream on omission break the SDK)
## Problem
`_call_openai_compat` (graphify 0.8.33, `llm.py`) calls `client.chat.completions.create(**kwargs)` **without** setting `stream`. Most providers treat an omitted `stream` as non-streaming, so this is fine against OpenAI/Kimi/etc.

But some **OpenAI-compatible gateways default to streaming when `stream` is omitted** (returning an SSE `text/event-stream` body). The OpenAI Python SDK then tries to parse the SSE stream as a single `ChatCompletion` and the call fails with:

```
'str' object has no attribute 'choices'
```

Every semantic chunk fails, so `graphify extract --backend <such-provider>` produces no semantic nodes. We hit this against a self-hosted SGLang-fronting gateway: omitting `stream` returned SSE; sending `"stream": false` returned a normal `chat.completion` object.

## Why it matters
A growing number of on-prem / self-hosted inference gateways (used precisely to keep extraction local and free) sit in front of SGLang/vLLM and stream by default. There is currently no way to make graphify send `stream:false`, and custom-provider configs in `~/.graphify/providers.json` don't carry a `stream` (or arbitrary extra-body) field, so the only fix today is patching graphify.

## Suggested fix (any one)
1. **Send `stream=False` explicitly** in `_call_openai_compat` (the SDK already aggregates a non-streamed response; this is a safe default for an extraction call that needs the whole JSON).
2. Honour a `stream` (and/or generic `extra_body`) key in custom-provider configs / a `GRAPHIFY_OPENAI_STREAM=false` env var.

Option 1 is the smallest change and is correct for graphify's use case (it always needs the complete JSON object, never a stream).

## Repro
```bash
# provider points at a gateway that streams on omission
graphify extract . --backend <custom-openai-compatible-provider> --max-concurrency 1
# -> [graphify] chunk 1/1 failed: 'str' object has no attribute 'choices'
```

jiangyq9:
## Summary
- pass `stream: False` on OpenAI-compatible chat completion calls
- add regression coverage that captures the request kwargs

Fixes #1223

## Testing
- python -m pytest tests/test_llm_backends.py -k "openai_compat or ollama_extra_body or non_ollama_backend"

### DIFF
--- a/graphify/llm.py
+++ b/graphify/llm.py
@@ -912,6 +912,7 @@ def _call_openai_compat(
             {"role": "user", "content": _openai_content(user_message, images or [])},
         ],
         "max_completion_tokens": max_completion_tokens,
+        "stream": False,
     }
     if temperature is not None:
         kwargs["temperature"] = temperature
--- a/tests/test_llm_backends.py
+++ b/tests/test_llm_backends.py
@@ -533,6 +533,17 @@ def test_non_ollama_backend_gets_no_num_ctx_extra_body(monkeypatch):
     assert eb is None or "options" not in eb, "non-ollama backends must not get num_ctx injection"
 
 
+def test_openai_compat_forces_non_streaming_response(monkeypatch):
+    captured = _install_capturing_openai(monkeypatch)
+
+    llm._call_openai_compat(
+        "https://gateway.example/v1", "sk-test", "gpt-4.1-mini",
+        "u", temperature=0, max_completion_tokens=8192, backend="openai",
+    )
+
+    assert captured["stream"] is False
+
+
 # ---------------------------------------------------------------------------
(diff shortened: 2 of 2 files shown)


safishamsi:
Thanks @jiangyq9 — verified safe (nothing in llm.py ever consumes a stream; every result is read as `resp.choices[0]`). Landed on `v8` in 666093f with your authorship. Maintainer fix on top: applied the same `stream: False` to the second OpenAI-compatible call site (`_call_llm`, used by the `--dedup-llm` tiebreaker), which had the identical bug. Closing as merged-by-port.

Issue #1446, opened 2026-06-24
Python qualified class-method calls miss EXTRACTED edges


noamsiegel:
## Summary

On graphifyy v0.8.46, Python calls written as `ClassName.method(...)` do not produce an EXTRACTED `calls` edge from the calling method to the target class method, even when both methods are extracted with class-qualified node IDs.

This leaves impact/caller lookup incomplete for common Django/service-layer patterns such as a ViewSet action delegating to an action/service class.

## Environment

- graphifyy: v0.8.46
- OS: macOS Darwin arm64
- Corpus: large Python/Django monorepo
- Graph: 233,713 nodes / 577,904 edges / 11,277 communities
- Built commit: `9e33b711243fb5b041e1500621430839c9da3303`

## Source shape

Caller in `apps/hoa/api/admin_api/viewsets/customer_tasks.py`:

```python
class CustomerTaskAdminViewSet(...):
    @action(detail=True, methods=["post"])
    def approve_draft_with_charge(self, request, pk=None):
        ...
        result = CustomerTaskActions.approve_draft_with_charge(
            pk, request.data, request.user, task_file, charge_file
        )
```

Target in `apps/hoa/api/actions/customer_task.py`:

```python
class CustomerTaskActions:
    @staticmethod
    def approve_draft_with_charge(pk, data, user, task_file=None, charge_file=None):
        ...
        charge_data = CustomerTaskActions._prepare_charge_data(data, approved_violation)
```

## What Graphify extracts correctly

The target method now has a class-qualified node ID and method ownership:

```bash
graphify explain actions_customer_task_customertaskactions_approve_draft_with_charge --graph graphify-out/graph.json
```

Observed:

```text
Node: .approve_draft_with_charge()
ID: actions_customer_task_customertaskactions_approve_draft_with_charge
Source: apps/hoa/api/actions/customer_task.py L970
Incoming: CustomerTaskActions --method [EXTRACTED]
Outgoing: ._prepare_charge_data() --calls [EXTRACTED]
```

The callee inside the same class is also correct:

```bash
graphify explain actions_customer_task_customertaskactions_prepare_charge_data --graph graphify-out/graph.json
```

Observed incoming callers include:

```text
.approve_draft_with_charge() --calls [EXTRACTED]--> ._prepare_charge_data()
.save_charge_draft() --calls [EXTRACTED]--> ._prepare_charge_data()
[...]

Issue #1356, opened 2026-06-17
Cross-file class relationships are mostly undetected by AST extraction


JabberYQ:
## Problem

The AST extractor (tree-sitter) processes files in isolation and does not perform cross-file type inference. This causes a large class of inter-file relationships — arguably the most architecturally important ones — to be absent from the graph.

## Root Cause

Tree-sitter is a per-file parser. When analyzing `FileA.swift`, it can resolve types and method calls only for symbols defined within the same file. Any reference to a type defined in `FileB.swift` requires cross-file type inference, which tree-sitter does not do.

### Example 1: View → ViewModel (MVVM core relationship)

**Code:**
```swift
// SessionViewController.swift
class SessionViewController: UIViewController {
    let viewModel = SessionViewModel()   // ← constructor call to another file

    override func viewDidAppear(_ animated: Bool) {
        super.viewDidAppear(animated)
        viewModel.update()               // ← method call to another file
    }
}

// SessionViewModel.swift  
class SessionViewModel: NSObject {
    func update() { ... }
}
```

**Graph result:** Zero edges between `SessionViewController` and `SessionViewModel`. They sit in different communities (284 vs 34) with no connection.

**Why:**
- `viewModel.update()` — `var.method()` syntax requires inferring `viewModel`'s type (`SessionViewModel`), which is defined in another file. Tree-sitter cannot resolve this.
- `SessionViewModel()` — constructor syntax `TypeName()` should theoretically be detectable, but property initializers do not generate `calls` edges the way method-body constructors do.

### Example 2: Static method on cross-file class

**Code:**
```swift
// SettingViewController.swift (LaikeMineModule)
let alert = NLKAlertStandardViewController.alertView(
    withTitle: "...",
    isButtonHorizontalLayout: true,
    highlightButtonTitle: "...",
    ...
)

// NLKAlertStandardViewController+Helper.m (LaikeUIKit)
@implementation NLKAlertStandardViewController (Helper)
- (id)alertViewWithTitle:... { ... }
```

**Graph result:** No `calls` edge from `SettingViewController` to `NLKAlertStandardViewController`.

**Why:** `TypeName.staticMethod(...)` requires resolving `TypeName` first, then looking up `staticMethod` on it. `NLKAlertStandardViewController` is defined in another pod (`LaikeUIKit`), tree-sitter cannot resolve.

### Example 3: Singleton method chain

**Code:**
```swift
// LoginManager.swift
public func logout() {
[...]

PR #1499, opened 2026-06-27
feat(ruby): type-aware member-call resolution via a resolver framework

vamsipavanmahesh:
## The problem, by example

Take three ordinary Ruby files:

```ruby
# helper.rb
class Processor
  def run(items)
    items.map { |i| transform(i) }
  end
end

# main.rb
def process_all(items)
  p = Processor.new
  p.run(items)        # <-- which `run` does this call?
end
```

A human knows instantly: `p` is a `Processor`, so `p.run` calls `Processor#run`.

But today the extractor **throws the receiver away**. When it sees `p.run`, it keeps only the bare method name `run` and discards `p`. With no receiver, the only way left to resolve the call is to search the whole project for *anything* named `run`:

- Right now there's exactly one `Processor#run`, so it links to it — **by luck**, not by understanding.
- Add one unrelated file and the luck runs out:

```ruby
# worker.rb  — a totally separate class, unrelated to main.rb
class Worker
  def run(jobs)
    jobs.each { |j| j }
  end
end
```

Now there are **two** definitions named `run` (`Processor#run` and `Worker#run`). The name match is ambiguous, so the resolver gives up and **drops the edge entirely**. The exact same line `p.run(items)` went from resolved to unresolved because of a file it never references. In other shapes the bare-name match can also pick the **wrong** `run`.

This is endemic to Ruby, where method names like `run`, `call`, `perform`, and `save` are reused everywhere.

## The fix

Resolve `obj.method()` by the **type of the receiver**, not by globally-unique name:

1. **Keep the receiver.** Ruby's `call` node stores `receiver`/`method` as direct fields (the generic accessor model doesn't fit it), so a Ruby-specific branch now captures `p` from `p.run`.
2. **Infer the receiver's type** from the method's local bindings. `p = Processor.new` ⇒ `p` is a `Processor`. (Strict guard: if `p` is reassigned, or assigned to anything other than a single `Const.new`, no type is inferred.)
3. **Resolve by type.** `p` is a `Processor` and `Processor` owns `run` ⇒ a `calls` edge to `Processor#run`.

Result for the example above, **with `worker.rb` present**:

```
process_all  --calls-->  Processor        (from `Processor.new`)
process_all  --calls-->  Processor#run    (NOT Worker#run, and not dropped)
```

The edge survives the name collision and always points at the right method, because it never depended on the name being unique.

## How it's built

Two layers:

**1. A small resolver-registry framework (`graphify/resolver_registry.py`)**

The cross-file, language-specific resolution passes at the tail of `extract()` were hand-wired as suffix-gated `if <lang>_paths: try: _resolve_<lang>(...)` blocks. This formalizes that de-facto extension point into a registry: a language registers a `LanguageResolver`, and the driver runs each whose suffix appears in the corpus, in order, with per-pass fault isolation.

It's a **behavior-identical refactor** of the existing Swift (#1356) and Python (#1446) passes — they become registered entries, so the current suite already covers them. The framework lives in its own module so `extract.py` doesn't keep growing, and a new language plugs in with one `register()` call. Ruby is its first new consumer.

**2. Ruby type-aware resolution (`graphify/ruby_resolution.py`)**

- *Extraction* (Ruby-gated): capture the member-call receiver, and attach a `receiver_type` inferred from local `var = ClassName.new` bindings.
- *Resolution*: `Processor.new` ⇒ `calls` edge to the `Processor` class; typed `p.run` ⇒ `calls` edge to `Processor#run`. Both `EXTRACTED` (1.0), emitted only when the owning class and method are unambiguous (god-node guard).

## Precision over recall (by design)

Ruby member calls are now **excluded from name-based ("unique label") resolution** and resolved by type instead. This trades some low-confidence `INFERRED` guesses for precise `EXTRACTED` edges and **zero false positives**, matching the project's existing confidence model: a member call whose receiver type can't be proven now resolves to nothing rather than to a guessed (possibly wrong) target.

## Tests (TDD)

New tests in `tests/test_ruby_resolution.py` and `tests/test_language_resolvers.py`:

- `p.run` resolves to `Processor#run`;
- **survives a name collision** — adding an unrelated `Worker#run` keeps the edge on `Processor#run` (a name-match resolver drops it as ambiguous);
- a call on an **unknown-type receiver resolves to nothing** (no false positive);
- ambiguous binding (reassignment) ⇒ no type inferred;
- registry contract: suffix gating, in-order execution, fault isolation, in-place mutation.

Full suite green; no regressions. The AST cache is version-namespaced, so the new `receiver_type` field is cache-safe on release.

## Scope / follow-ups

Deliberately focused on type-aware member-call resolution. Natural next increments: `require`/`require_relative` file-import edges (a file-level mechanism that belongs with the bash-`source` resolver) and `include`/`extend` mixin edges.

### DIFF
--- a/graphify/extract.py
+++ b/graphify/extract.py
@@ -13,6 +13,12 @@ from typing import Any, Callable
 from .cache import load_cached, save_cached
 from .mcp_ingest import extract_mcp_config, is_mcp_config_path
 from .manifest_ingest import extract_package_manifest, is_package_manifest_path
+from .resolver_registry import (
+    LanguageResolver,
+    register as register_language_resolver,
+    run_language_resolvers,
+)
+from .ruby_resolution import resolve_ruby_member_calls
 
 # --- migrated to graphify/extractors/ (see graphify/extractors/MIGRATION.md) ---
 from graphify.extractors.base import (  # noqa: F401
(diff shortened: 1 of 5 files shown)


vamsipavanmahesh:
@safishamsi any feedback here?

safishamsi:
Merged into `v8` as e3e4198-range (commit 86ecb76), with your authorship preserved. Thanks for this — the resolver-registry framing is clean and the type-based receiver resolution is exactly the right call for Ruby.

Reviewed as a behavior-identical refactor of the existing Swift/Python passes (same suffix gating, order, fault isolation, in-place mutation), confirmed `receiver_type`/`member_receiver` never leak into the output graph, and ran the full suite (2530+ passing) plus an edge-case smoke: 5-file project where the typed `p.run` resolves to `Processor#run`, survives a colliding `Worker#run`, and unknown/reassigned receivers correctly drop to no edge.

One note for anyone reading later: this intentionally excludes Ruby member calls from name-based cross-file resolution, so a `var.method` whose receiver type can't be proven from a local `X.new` binding now resolves to nothing rather than by name-luck. That precision-over-recall tradeoff is the right one and is captured in the merge commit message.

PR #1501, opened 2026-06-27
fix(extract): Go ensure_named_node emits sourceless cross-file stub (#1402 gap)

### LINKED ISSUE #1500: Go AST extractor emits phantom duplicate nodes for cross-file type references
A Go type defined once but referenced via parameter/return/field types in N other
files produces **1+N** nodes — the extras carry the referencing file's path
(extension and all) baked into the id (`pkg_a_go_thing`). This is the same
phantom-duplicate class #1402 handled for the other extractors; the Go copy of
`ensure_named_node` still uses the older *sourced*-stub fallback, so its cross-file
references don't get picked up by `_rewire_unique_stub_nodes`.

Surfaced while running graphify over two production Go codebases — `sql.NullTime`
referenced across files showed up as 15 separate nodes and `PublicKey` as 33.

### Minimal repro
```
pkg/thing.go:  package pkg; type Thing struct{}; func (t Thing) Run() int { return 1 }
pkg/a.go:      package pkg; func UseA(obj Thing) Thing { return obj }
pkg/b.go:      package pkg; func UseB(obj Thing) Thing { return obj }
```
```python
from pathlib import Path
r = extract([Path('pkg/thing.go'), Path('pkg/a.go'), Path('pkg/b.go')], cache_root=Path('.'))
print(sorted(n['id'] for n in r['nodes'] if n['label'] == 'Thing'))
# actual:   ['pkg_a_go_thing', 'pkg_b_go_thing', 'pkg_thing']
# expected: ['pkg_thing']
```

### Fix
Make the Go `ensure_named_node` cross-file fallback emit a *sourceless* stub like the
other extractors, so the references resolve to the single canonical definition. With
that the repro yields one `pkg_thing`, and across the two corpora the phantom
type-ref nodes drop from 116→7 and 158→7 — the residual being external types
(`sql.NullTime`, no local definition) and same-named types defined in two packages,
which the other extractors leave alone by design.

TPAteeq:
Closes #1500. Follow-up to #1402 / `0aeda15`.

#1402 fixed the phantom-duplicate-node bug by making `ensure_named_node`'s cross-file
fallback emit a **sourceless** stub instead of a **sourced** one, so
`_rewire_unique_stub_nodes` collapses a type referenced in N files onto its single
definition (instead of 1+N nodes with the referencing file's path baked into the id,
`pkg_a_go_thing`). This extends that fix to the Go copy of `ensure_named_node`, which
still used the older **sourced**-stub fallback and so still produced the phantoms on `v8`.

## Change

`graphify/extract.py` — Go `ensure_named_node` cross-file fallback now appends a
sourceless stub (`source_file=""`, `source_location=""`) like the other extractors,
instead of calling `add_node(...)`. Go's package-scoped first check
(`_make_id(pkg_scope, name)`, which the others don't have) is preserved; only the
fallback branch changed. Adds a one-line `## Unreleased` CHANGELOG entry.

## Result

Minimal repro now yields one `pkg_thing` instead of `['pkg_a_go_thing',
'pkg_b_go_thing', 'pkg_thing']`. On two production Go corpora (AST-only):

| repo | phantom type-ref nodes | edges stranded on phantoms | total nodes |
|---|---|---|---|
| A (62 .go) | 116 → **7** | 503 → **297** | 1629 → 1428 |
| B (103 .go) | 158 → **7** | 230 → **82** | 1105 → 824 |

The residual 7 per repo are the cases the other extractors also leave alone —
mostly references to external types (the `database/sql` `Null*` family: a same-named
local constructor func, no local type), plus the occasional type genuinely defined
under the same name in two packages (conservative non-merge, #1257).

## Tests

- New `test_go_cross_file_type_refs_resolve_to_single_node` (mirrors the Python #1402
  test). Fails pre-fix (3 nodes: `pkg_thing`, `pkg_a_go_thing`, `pkg_b_go_thing`),
  passes post-fix.
- `tests/test_extract.py`, `tests/test_languages.py`, `tests/test_multilang.py`: 416 pass.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 Full release notes with details on each version: [GitHub Releases](https://github.com/safishamsi/graphify/releases)
 
+## Unreleased
+
+- Fix: the Go AST extractor no longer creates phantom duplicate nodes for cross-file type references — the Go copy of `ensure_named_node` still used the older sourced-stub fallback; it now emits a sourceless stub like the other extractors, extending the #1402 fix to Go (#1500).
+
 ## 0.8.50 (2026-06-27)
 
 - Feat: `graphify label --missing-only` relabels only communities that are unnamed or still hold a `Community N` placeholder, preserving existing non-placeholder labels from `.graphify_labels.json` (#1481, thanks @jiangyq9; supersedes #1421 by @matiasduartee, who proposed the same flag). Lets a large graph be relabeled incrementally without re-naming (and paying for) communities that already have good names.
--- a/graphify/extract.py
+++ b/graphify/extract.py
@@ -6472,7 +6472,21 @@ def extract_go(path: Path) -> dict:
             return nid
         nid = _make_id(name)
         if nid not in seen_ids:
-            add_node(nid, name, line)
+            # The name isn't declared in this file, so this is a cross-file reference
+            # (e.g. a type defined in another file of the package). Emit a SOURCELESS
+            # stub — like the inheritance-base path in the other extractors — so the
+            # corpus-level rewire can collapse it onto the real definition. A sourced
+            # stub here makes _disambiguate_colliding_node_ids bake the referencing
+            # file's path (with extension) into the id and blocks the rewire, which is
+            # the phantom-duplicate-node bug (#1402).
+            seen_ids.add(nid)
+            nodes.append({
+                "id": nid,
+                "label": name,
+                "file_type": "code",
+                "source_file": "",
+                "source_location": "",
+            })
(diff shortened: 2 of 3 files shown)


safishamsi:
Thanks @TPAteeq — correct and well-scoped. Go's `ensure_named_node` was emitting a sourced stub, unlike the other extractors, which baked the referencing file's path into the id and blocked the rewire (the phantom duplicates). Verified on a fixture: 3 `Thing` nodes on v8 → 1 with this fix. Landed on `v8` in 36b76ce with your authorship. Closing as merged-by-port.

PR #1502, opened 2026-06-27
V8 windows skill fix

antonioscarinci:
fix: Fixed some issues with the Windows skill and excessive token consumption

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -4,3 +4,19 @@
 worked/**/*.html linguist-vendored=true
 graphify-out/**/*.html linguist-vendored=true
 *.html linguist-detectable=false
+
+# Normalize line endings
+* text=auto
+
+# Scripts always get LF
+*.sh text eol=lf
+*.py text eol=lf
+
+# Windows batch files always get CRLF
+*.bat text eol=crlf
+*.cmd text eol=crlf
+
+# Binary files
+*.pdf binary
+*.png binary
+*.jpg binary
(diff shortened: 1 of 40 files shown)


safishamsi:
Thanks for this. I pulled the two clean, cross-platform fixes out of here and landed them on `v8` as 407a7f1 (with your authorship), since they're useful on every platform and shouldn't wait on the larger Windows rework:

- **`to_graphml` null-attr coercion** — `nx.write_graphml` raises `ValueError` on a `None` attribute value, so a node/edge with a null field crashed the export. Now coerced to `""`.
- **`save-result --answer-file`** — lets long/multiline answers come from a file instead of a fragile inline arg (Windows/PowerShell quoting). Exactly one of `--answer`/`--answer-file` is required.

Both shipped with regression tests and verified by smoke (GraphML with `None` on node+edge attrs exports to valid XML; a multiline+unicode answer round-trips through `--answer-file`).

Leaving this PR open for the rest, which needs rework before it can merge:

1. **`pyproject.toml` version** — the branch was cut from an old `v8` and this drags the version **backward** from 0.9.1 to 0.8.49. Please drop that line; releases are managed separately.
2. **`graphify/skill-windows.md` is a generated artifact** (rendered from `tools/skillgen/fragments/` by `python -m tools.skillgen`). It was hand-edited without touching the fragments, so `skillgen --check` fails and CI is red. The Windows skill changes need to be made in the fragments and regenerated so committed == rendered.
3. **`windows-scripts/*.py`** — `query_graph.py` and `build_graph.py` reimplement the existing `graphify query` / build pipeline. Prefer having the scripts shell out to the existing CLI commands rather than duplicating traversal/pipeline logic, which will drift over time.

Happy to re-review once those are addressed. Appreciate the Windows focus — it's a real gap.

PR #1503, opened 2026-06-27
Fix source_file path lookup for explain and affected

behavio1:
## Summary

This PR fixes full `source_file` path lookup for:

- `graphify explain "<source-file-path>"`
- `graphify affected "<source-file-path>"`

It also adds regression coverage for the case where a graph contains multiple
nodes from the same source file, such as a file-level node plus function nodes.

## Root cause

Graphs can contain several nodes with the same `source_file`. For example, a
route file can have a file-level `route.ts` node and a `GET()` handler node from
the same path.

Before this change:

- `explain` could miss full path queries because `_find_node()` tokenizes the
  query, while the trigram prefilter only indexed the raw `source_file`.
- `affected` could fail to resolve a full path because exact `source_file`
  lookup returned multiple candidate nodes and no preferred seed.

## What changed

- Include tokenized `source_file` text in the `serve.py` trigram search index.
- Let `_find_node()` match exact tokenized `source_file` paths.
- Prefer the file-level node when a full source path matches multiple nodes from
  the same file.
- Apply the same file-level preference in `affected.resolve_seed()`.
- Add regression tests for `explain`, `affected`, and `_find_node()`.

## Validation

```bash
uv run python -m py_compile graphify/serve.py graphify/affected.py
uv run ruff check graphify/serve.py graphify/affected.py tests/test_serve.py tests/test_affected_cli.py tests/test_explain_cli.py
uv run pytest tests/test_serve.py tests/test_affected_cli.py tests/test_explain_cli.py -q
env -u GEMINI_API_KEY -u GOOGLE_API_KEY -u GOOGLE_GENAI_API_KEY -u MOONSHOT_API_KEY -u ANTHROPIC_API_KEY -u OLLAMA_BASE_URL uv run pytest tests/ -q
```

Full test result with provider env vars unset:

```text
2430 passed, 28 skipped
```

### DIFF
--- a/graphify/affected.py
+++ b/graphify/affected.py
@@ -54,6 +54,41 @@ def _normalize_label(label: str) -> str:
     return unicodedata.normalize("NFC", label).casefold()
 
 
+def _prefer_file_node(
+    graph: nx.Graph,
+    node_ids: list[str],
+    query: str,
+) -> str | None:
+    """Return the file-level node when a source_file query matches many nodes."""
+    query_basename = _normalize_label(Path(query).name)
+    exact_file_nodes = [
+        node_id
+        for node_id in node_ids
+        if str(graph.nodes[node_id].get("source_location", "")) == "L1"
+        and _normalize_label(str(graph.nodes[node_id].get("label", ""))) == query_basename
+    ]
+    if len(exact_file_nodes) == 1:
+        return exact_file_nodes[0]
+
+    l1_nodes = [
+        node_id
+        for node_id in node_ids
+        if str(graph.nodes[node_id].get("source_location", "")) == "L1"
+    ]
+    if len(l1_nodes) == 1:
+        return l1_nodes[0]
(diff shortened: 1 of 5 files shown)


safishamsi:
Thanks @behavio1 — real bug, cleanly fixed, and the multi-node-per-source-file regression tests are exactly the right coverage. The file-node preference (L1 + basename match, with safe fallbacks) is sensible. Landed on `v8` in 1b99496 with your authorship. Two small maintainer fixes on top: aligned trailing-separator handling between `resolve_seed` (affected) and `_find_node` (explain) — affected previously returned None for a trailing-slash path that explain resolved — and corrected the stale `_find_node` docstring; added regression tests for both. Verified end-to-end: path query resolves to the file node, label lookups unaffected, nonexistent path graceful, ambiguous no-L1 returns None. Closing as merged-by-port.

PR #1507, opened 2026-06-28
fix(llm): add errors=replace to claude-cli subprocess.run to fix UnicodeDecodeError on Windows GBK systems

### LINKED ISSUE #1505: Windows + claude-cli backend crashes with UnicodeDecodeError when claude.cmd emits GBK/cp936 bytes
Issue 1 (main): Windows + claude-cli backend crashes with UnicodeDecodeError when claude.cmd emits GBK/cp936 bytes

  Title

  Windows + claude-cli backend: UnicodeDecodeError crashes reader thread when claude.cmd emits GBK/cp936 bytes                                          
  
  Environment

  - Windows 11 (Chinese system, ACP = cp936/GBK)
  - Python 3.13
  - graphify 0.8.50
  - --backend claude-cli (Claude Code CLI 2.x, claude.cmd)

  What happened

  Running graphify extract ./docs --backend claude-cli, the semantic extraction stage hangs the terminal. After Ctrl+C, the following is thrown:

  Exception in thread Thread-8 (_readerthread):
    File "...\Python313\Lib\subprocess.py", line 1611, in _readerthread
      buffer.append(fh.read())
                    ~~~~~~~^^
    File "<frozen codecs>", line 325, in decode
  UnicodeDecodeError: 'utf-8' codec can't decode byte 0xd6 in position 778: invalid continuation byte

  process.communicate() then blocks forever and the whole graphify process has to be killed with Ctrl+C, losing the result. The traceback points to graphify/llm.py:1157, 
  the subprocess.run(..., encoding="utf-8") call in _call_claude_cli.

  Root cause

  _call_claude_cli (graphify/llm.py:1157-1166) uses:

  proc = subprocess.run(
      cli_args, input=user_message, capture_output=True, text=True,
      encoding="utf-8",  # Force UTF-8 — prevents UnicodeEncodeError on Windows cp1252                                                                                    
      timeout=_resolve_api_timeout(), check=False, **_no_window_kwargs(),
  )

  The comment on llm.py:1162 shows the UTF-8 forcing was added to prevent a stdin-side UnicodeEncodeError on cp1252. However, encoding="utf-8" is also applied to
  stdout/stderr decoding, with errors defaulting to strict. On a Chinese Windows system, claude.cmd's stdout/stderr is emitted in the system ACP (cp936/GBK) whenever the 
  output contains CJK text (Chinese prose returned by the model, error messages, etc.). Forcing UTF-8 decoding then hits non-UTF-8 bytes (e.g. 0xd6, a typical first byte 
  of a GBK CJK character) → UnicodeDecodeError → the reader thread dies → communicate() blocks indefinitely → graphify hangs.

  In short: the comment guards the input-side encode, but not the output-side decode; the decode side crashes on GBK bytes.

  The same subprocess.run(..., encoding="utf-8") pattern exists in _call_llm (around llm.py:1910, the --dedup-llm claude-cli path) and has the identical bug.

  Suggested fix

  Add errors="replace" to the subprocess.run call so output decoding tolerates non-UTF-8 bytes instead of crashing:

    proc = subprocess.run(
        cli_args,
        input=user_message,
        capture_output=True,
        text=True,
        encoding="utf-8",  # Force UTF-8 — prevents UnicodeEncodeError on Windows cp1252
  +     errors="replace",  # Tolerate non-UTF-8 bytes (e.g. GBK/cp936 CJK output from claude.cmd on Windows) instead of crashing the reader thread                        
        timeout=_resolve_api_timeout(),
        check=False,
        **_no_window_kwargs(),
    )

  Apply the same one-liner to the _call_llm claude-cli path (llm.py:1910).

  Why this is safe

  - errors only affects decoding. The stdin encode remains utf-8 strict, and user_message is graphify-controlled text that is always UTF-8-encodable, so no encode error  
  is introduced.
  - With tolerant decoding, the JSON structure in stdout still parses fine (_parse_llm_json is unaffected; at worst a few CJK field values may contain replacement chars).
  - It makes no assumption about claude.cmd's output encoding (which varies with Windows ACP and the model's reply language). utf-8 + replace is the most general tolerant
  choice.

  Repro

  On a Chinese Windows environment, run graphify extract ./docs --backend claude-cli where the docs contain CJK text or the model returns CJK prose. Not always
  reproducible — depends on whether claude.cmd happens to emit non-UTF-8 bytes.

  ---
  Issue 2 (optional, separate): claude-cli backend ignores --system-prompt due to project/global CLAUDE.md steering

  Title

  claude-cli backend: claude -p loads project/global CLAUDE.md steering, ignores --system-prompt and returns prose instead of JSON                                        

  What happened

  With --backend claude-cli, the claude -p subprocess loads the cwd's CLAUDE.md/AGENTS.md plus the global ~/.claude/CLAUDE.md, enters a "project assistant standby" mode, 
  and replies with a natural-language greeting ("I've loaded the project context... tell me what you need") instead of the JSON graphify requests via --system-prompt.    
  This causes LLM returned invalid JSON, skipping chunk and triggers the adaptive-retry bisection loop.

  Root cause

  _call_claude_cli (llm.py:1142-1148) passes --system-prompt, which only replaces Claude Code's default coding-agent prompt. It does not suppress the loading of
  project-level / global steering files (CLAUDE.md/AGENTS.md), so the model's effective context still includes the steering and the --system-prompt extraction
  instructions get diluted/ignored.

  Suggested fix

  When spawning claude -p, point cwd at a throwaway empty directory and set CLAUDE_CONFIG_DIR to an empty directory in the child env, so no steering is loaded. Auth is   
  unaffected because the ANTHROPIC_* env vars are still inherited by the child. Clean up the temp dir in a try/finally:

    import tempfile
    _clean_cfg = tempfile.mkdtemp(prefix="graphify-cc-")
    _child_env = {**os.environ, "CLAUDE_CONFIG_DIR": _clean_cfg}
    try:
        proc = subprocess.run(
            cli_args,
            input=user_message,
            cwd=_clean_cfg,
            env=_child_env,
            capture_output=True, text=True,
            encoding="utf-8", errors="replace",
            timeout=_resolve_api_timeout(), check=False,
            **_no_window_kwargs(),
        )
    finally:
        shutil.rmtree(_clean_cfg, ignore_errors=True)

  Note: this is a more involved change because it touches the auth/steering semantics of the claude CLI, so it likely needs discussion. Filing separately from Issue 1    
  keeps the simple crash fix unblocked.

  ---
  File Issue 1 first — it's a clear-cut bug with a one-line fix. Issue 2 is more design-level and can be discussed separately.

nuthalapativarun:
## Summary

- `subprocess.run(..., encoding="utf-8")` defaults to `errors="strict"`, so when `claude.cmd` emits GBK/cp936-encoded bytes (CJK output on Chinese Windows), the reader thread raises `UnicodeDecodeError` and `communicate()` blocks indefinitely
- Adds `errors="replace"` to both `subprocess.run` call sites in `_call_claude_cli` (line ~1157) and the `_call_llm` claude-cli path (line ~1910)

## Changes

- `graphify/llm.py` — add `errors="replace"` to both `subprocess.run` calls in the claude-cli backend
- `tests/test_llm_backends.py` — two regression tests: one asserting `errors="replace"` is passed to `subprocess.run`, one verifying non-UTF-8 stderr is handled without raising

## Verification

```
uv run pytest tests/test_llm_backends.py -q
79 passed
```

Closes #1505

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -40,3 +40,7 @@ paper/
 
 # macOS Finder metadata
 .DS_Store
+
+# OSS contribution analysis files (local only)
+analyze-*.md
+tracking.md
--- a/graphify/llm.py
+++ b/graphify/llm.py
@@ -1160,6 +1160,7 @@ def _call_claude_cli(user_message: str, max_tokens: int = 8192, *, deep_mode: bo
         capture_output=True,
         text=True,
         encoding="utf-8",  # Force UTF-8 — prevents UnicodeEncodeError on Windows cp1252
+        errors="replace",  # Tolerate non-UTF-8 bytes (e.g. GBK/cp936 from claude.cmd on Chinese Windows)
         timeout=_resolve_api_timeout(),
         check=False,
         **_no_window_kwargs(),
@@ -1913,6 +1914,7 @@ def _call_llm(
             capture_output=True,
             text=True,
             encoding="utf-8",  # Force UTF-8 — prevents UnicodeEncodeError on Windows cp1252
+            errors="replace",  # Tolerate non-UTF-8 bytes (e.g. GBK/cp936 from claude.cmd on Chinese Windows)
             timeout=_resolve_api_timeout(),
             check=False,
             **_no_window_kwargs(),
--- a/tests/test_llm_backends.py
+++ b/tests/test_llm_backends.py
@@ -931,3 +931,55 @@ def test_base_url_defaults_without_env(backend, default):
         env=env, capture_output=True, text=True, check=True,
     )
     assert out.stdout.strip() == default
+
+
+# ---------------------------------------------------------------------------
+# #1505: claude-cli subprocess.run must use errors="replace" so non-UTF-8
+# bytes from claude.cmd on Chinese Windows (GBK/cp936) don't crash the reader
+# thread.
+# ---------------------------------------------------------------------------
+
+import json as _json
+
+
+def _make_cli_envelope(result_text: str) -> str:
+    """Return a minimal claude -p --output-format json envelope."""
+    return _json.dumps({"type": "result", "result": result_text, "usage": {}, "modelUsage": {}})
+
+
+def test_call_claude_cli_passes_errors_replace_to_subprocess():
+    """subprocess.run must be called with errors='replace' so non-UTF-8 output
+    bytes (e.g. GBK from claude.cmd on Chinese Windows) are tolerated instead
+    of crashing the reader thread with UnicodeDecodeError (#1505)."""
+    from unittest.mock import patch, MagicMock
+
+    valid_envelope = _make_cli_envelope('{"nodes":[],"edges":[],"hyperedges":[]}')
+    mock_proc = MagicMock()
+    mock_proc.returncode = 0
+    mock_proc.stdout = valid_envelope
+    mock_proc.stderr = ""
+
+    with patch("platform.system", return_value="Linux"), \
+         patch("shutil.which", return_value="/usr/bin/claude"), \
+         patch("subprocess.run", return_value=mock_proc) as mock_run:
+        llm._call_claude_cli("test prompt")
+
+    assert mock_run.call_args.kwargs.get("errors") == "replace", \
+        "subprocess.run missing errors='replace' — non-UTF-8 bytes will crash the reader thread"
+
+
+def test_call_claude_cli_tolerates_non_utf8_in_stderr():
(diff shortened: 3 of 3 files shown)


safishamsi:
Thanks @nuthalapativarun — reproduced the UnicodeDecodeError with real cp936 bytes; `errors="replace"` is the right knob and `capture_output=True` covers stdout+stderr. Both claude-cli subprocess sites patched. Landed on `v8` in 0e8d92c with your authorship (dropped an unrelated .gitignore change). Closing as merged-by-port.

PR #1515, opened 2026-06-28
fix(extract): add origin_file to cross-file stubs in the six dedicated extractors

TPAteeq:
Follow-up to #1462. `ensure_named_node`'s cross-file fallback emits a **sourceless**
stub (`source_file=""`) for a type referenced from another file. #1462 added an
internal `origin_file` field so that when there's **no single project definition** to
rewire onto, two files referencing the same bare type name get **distinct** stubs
(disambiguated by `origin_file`) instead of conflating into one shared bare-id node —
while keeping `source_file=""` so the #1402 rewire still collapses them onto a real
definition when one does exist.

That landed in only the **generic** extractor. The **six dedicated** extractors
(`extract_go`, `extract_rust`, `extract_julia`, `extract_fortran`,
`extract_powershell`, `extract_objc`) still lacked `origin_file`, so they kept
conflating.

## Change

Add `"origin_file": str_path` to the sourceless stub in those six extractors (one line
each), matching the generic one. Only the `ensure_named_node` imported-type-stub path
is touched — the same path #1462 changed.

## Result

Two files referencing the same bare type with no local definition:

| extractor | before | after |
|---|---|---|
| Python (generic) | 2 distinct stubs | 2 (unchanged) |
| Go / Rust / … (dedicated) | **1 conflated `widget`** | **2 distinct** (`a_use_a_go_widget`, `b_use_b_go_widget`) |

The #1402 rewire is unaffected (`source_file` stays empty): when a single real
definition exists, the stubs still collapse onto it.

## Tests

- New `test_go_imported_type_stubs_do_not_collide_across_source_files` (mirrors the
  Python #1462 test). Fails pre-fix (one conflated node), passes post-fix.
- `tests/test_extract.py`, `tests/test_languages.py`, `tests/test_multilang.py`: 420 pass.

## Out of scope

The generic extractor has other sourceless-stub sites (inheritance bases) that also
lack `origin_file`. #1462 didn't touch those either, and a shared base class is more
plausibly the same type (collapsing is more often correct there), so they're kept aside for a different PR.


safishamsi:
Thanks @TPAteeq — this completes the #1462 disambiguation across the six dedicated extractors (Julia/Fortran/Go/Rust/PowerShell/ObjC), which previously only the generic extractor had. Landed on `v8` in d177f04 with your authorship, alongside #1516 (which strips the field from output — they had to ship together since this widens `origin_file` to more languages). Verified: a 2-file Go corpus now yields 2 distinct `Widget` stubs AND graph.json carries no `origin_file`; the #1402 rewire onto a real definition is unchanged. Closing as merged-by-port — ships in 0.9.1.

Issue #932, opened 2026-05-18
Semantic extractors emit mixed absolute/relative source_file paths


cornwe19:
## Symptom

After running graphify on a docs corpus from the repo root with no explicit path argument (so graphify defaults to `.`), the resulting `graph.json` contains a roughly 50/50 mix of absolute and relative `source_file` values on nodes, links, and hyperedges.

Concrete numbers from one of our runs (2,587-node graph over a markdown/docs corpus):

| Element     | Absolute | Relative |
|-------------|---------:|---------:|
| nodes       |    1,292 |    1,295 |
| links       |    1,479 |        — |
| hyperedges  |       55 |        — |

Absolute samples (note: these are not paths we passed in — graphify was invoked from the repo root with no path arg, so the cwd is the only thing it knew about):

```
/Users/dennis/Documents/projects/documents/document-linking-investigation.md
/Users/dennis/Documents/projects/documents/CLAUDE.md
```

All affected nodes are `file_type: document` (`.md` files) — i.e., they go through the semantic-extraction subagent path, not the AST path. AST-extracted nodes appear to be consistently relative.

## Impact

Downstream consumers that key off `source_file` for exact-string matching break silently. For us, the MCP traversal seeder matches chunk paths (which are always relative) against `source_file`, so every absolute-path node/link/hyperedge becomes invisible to graph traversal — roughly half the graph in our case.

## Likely cause

The subagent prompt in the SKILL/README instructs each chunk worker to emit `\"source_file\":\"relative/path\"`, but with no enforcement the LLM is free to echo whatever it received in `FILE_LIST`. In practice ~half the time it does. The fix probably belongs in the extractor write path — normalize `source_file` to be relative to the corpus root before serializing into `graph.json` — rather than relying on the subagent to obey the instruction.

## Workaround we're using

A post-processing script that rewrites `graph.json` after every build, stripping the docs-root prefix. Effective but obviously a band-aid — happy to upstream if useful, but a guarantee at write time would be the right fix.

Let me know if a repro corpus would help.

cornwe19:
I think this is probably a duplicate of #777

Issue #555, opened 2026-04-25
graph.json stores absolute source_file paths, breaking portability across git worktrees and machines


lucasrcosta:
## Problem

`graph.json` stores `source_file` as absolute paths (e.g. `/Users/alice/projects/myapp/src/lib/auth.ts`). This makes the committed graph non-portable:

- **Git worktrees**: building the graph in one worktree produces paths that don't resolve in sibling worktrees (same repo, different directory)
- **Teammates**: cloning the repo gives a graph with paths pointing to the original author's machine
- **CI/CD**: any pipeline that queries the graph gets paths from wherever it was last built

The graph is designed to be committed alongside the code. Absolute paths defeat that entirely.

## Expected behavior

`source_file` should be stored relative to the project root (e.g. `src/lib/auth.ts`). Tools consuming the graph can resolve to absolute by joining with their own CWD, which is the standard convention for committed path references.

## Workaround

We're post-processing `graph.json` in the post-commit and post-checkout hooks to strip the build-time CWD prefix:

```python
cwd_prefix = str(Path('.').resolve()) + '/'
for node in graph_data['nodes']:
    sf = node.get('source_file') or ''
    if sf.startswith(cwd_prefix):
        node['source_file'] = sf[len(cwd_prefix):]
```

This works but shouldn't be necessary — the extraction step should write relative paths from the start.

## Environment

- graphifyy 0.5.0
- macOS, git worktree setup (main worktree + multiple linked worktrees via conductor)

safishamsi:
Fixed: `extract()` now relativizes `source_file` fields in all nodes and edges to the computed `root` before returning. This ensures paths are portable across machines for both full and incremental builds.

PR #1516, opened 2026-06-28
fix(extract): drop internal origin_file so it stops leaking into graph.json

TPAteeq:
Follow-up to #1462. `ensure_named_node`'s cross-file fallback tags each sourceless stub
with an internal `origin_file` field — the referencing file's **absolute** path — so
`_disambiguate_colliding_node_ids` can keep same-named cross-file references distinct.
That field is consumed only by the colliding-id pass during extraction, but it was never
dropped, so it shipped into `graph.json` verbatim as an absolute, machine-specific path,
violating the "no absolute paths in output" contract that already relativizes `source_file`
(#555, #932). No exporter caught it: `to_json` passes node fields through unchanged, and
`to_graphml` (and the Neo4j/FalkorDB pushers) strip only `_`-prefixed keys — `origin_file`
has no underscore.

## Change

Drop `origin_file` from the combined node list after the colliding-id pass consumes it,
right before the `_origin` tagging loop in `extract()` (one strip at the source — covers
every export surface and the cache-hit path; cannot re-enter via `build_from_json`).

## Why `_origin` is left alone

The original observation also named `_origin`. It is **intentionally kept**: the
incremental watcher reads `n.get("_origin") == "ast"` from the persisted `graph.json` to
evict stale AST nodes on a full rebuild (#1116, `watch.py`). Stripping it would reintroduce
that bug. It is the constant `"ast"` — not a path — so it is not an absolute-path leak.

## Result

| field on a cross-file stub | before | after |
|---|---|---|
| `origin_file` (abs path) | present, e.g. `/Users/you/proj/pkg/a.py` | **gone** |
| node id / `source_file` distinctness | distinct | distinct (unchanged) |
| `_origin` | `"ast"` | `"ast"` (unchanged) |

The colliding-id pass runs **before** the strip, so cross-file stubs stay distinct; the
per-file AST cache keeps its own `origin_file` copy for the cache-hit path.

## Tests

- New `test_origin_file_is_not_serialized_into_extract_output`: asserts no node carries
  `origin_file` (or any absolute sandbox path) while the two cross-file `Path` stubs stay
  distinct. Fails pre-fix, passes post-fix.
- Full suite: 2494 passed, 2 skipped.

## Out of scope

- The **per-file AST cache** still stores `origin_file` as an absolute path
  (`cache.save_cached` relativizes `source_file` per #777 but not `origin_file`). The cache
  is a private, content-hashed, version-namespaced artifact that never ships, and the strip
  above keeps it out of `graph.json` regardless — so this is a separate portability/hygiene
  follow-up, not part of this fix.
- For a same-named cross-file stub, the disambiguated node **id** is a slug derived from the
  referencing path; under a fully-resolved scan root that slug is root-relative. This is
  #1462's existing disambiguation behavior, unchanged here.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 Full release notes with details on each version: [GitHub Releases](https://github.com/safishamsi/graphify/releases)
 
+## Unreleased
+
+- Fix: the internal `origin_file` disambiguation field (#1462) is no longer serialized into graph.json, where it had shipped as an absolute, machine-specific path — it is dropped once the colliding-id pass consumes it, keeping output portable (#555, #932). `_origin` stays (the incremental watcher needs it, #1116).
+
 ## 0.8.51 (2026-06-28)
 
 - Fix: the Obsidian export (`--obsidian` / `to_obsidian`) no longer overwrites a user's own notes or `.obsidian/` config when pointed at an existing vault (#1506). It wrote one note per node straight into the target dir and unconditionally replaced `.obsidian/graph.json`, so `--obsidian-dir ~/my-vault` could clobber a same-named note (`Database.md`) and the user's graph-view settings — silently, no backup. graphify now records the files it owns in a `.graphify_obsidian_manifest.json` and refuses to overwrite any pre-existing file it didn't create (skipping it with one aggregated warning); a re-run still updates graphify's own notes. The default `graphify-out/obsidian` output is unchanged.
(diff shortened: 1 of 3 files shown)


safishamsi:
Thanks @TPAteeq — confirmed: `origin_file` was shipping into graph.json in 0.9.0 as an absolute machine-specific path (same portability class as #555/#932). Landed on `v8` in afa4ade with your authorship. The strip sits at the single `extract()` output chokepoint, after the colliding-id pass consumes the field, so all persist paths (clustered, `--no-cluster`, cluster-only) emit clean output and the #1462 disambiguation still works. Verified: fresh extract → zero `origin_file` in graph.json. Closing as merged-by-port — ships in 0.9.1.

PR #1530, opened 2026-06-29
Use host-generic graphify skill instructions

ari-mitophane:
## Purpose
Fix Graphify-generated install guidance so Codex/agent-facing surfaces do not tell hosts to invoke a literal `skill` tool with `skill: "graphify"`. The generated wording now stays host-generic and points users to the installed graphify skill or instructions.

## Implemented Delta
- Updated the always-on AGENTS source fragment, packaged `graphify/always_on/agents-md.md`, and expected skillgen snapshot to use host-generic `/graphify` guidance.
- Updated `_skill_registration()` so Claude/CodeBuddy registration snippets use the same host-generic wording.
- Replaced stale English code-block examples in translated README files.
- Added regression tests for AGENTS and registration helper wording.
- Tightened the always-on roundtrip test to allow only the exact approved agents-md sentence delta.
- Made skillgen git blob reads decode as UTF-8 explicitly on Windows.

## Current Behavior
`graphify codex install` and related generated/registration examples can surface wording that tells agents to invoke a literal `skill` tool with `skill: "graphify"`, which is host-specific and not valid in every environment.

## Desired Behavior
Generated install guidance should say that `/graphify` should use the installed graphify skill or instructions before doing anything else, without naming a literal tool API that may not exist.

## Key Interfaces / Modules
- `tools/skillgen/fragments/always-on/agents-md.md`: source AGENTS wording.
- `graphify/always_on/agents-md.md`: packaged always-on artifact.
- `tools/skillgen/expected/graphify__always_on__agents-md.md`: generated snapshot.
- `graphify/__main__.py`: `_skill_registration()` output and UTF-8 git blob reads.
- `tests/test_install_strings.py` and `tests/test_skillgen.py`: regression coverage.

## Decision Path
The bad text is produced by Graphify install surfaces, so the durable fix is upstream in the template/generator path rather than downstream edits in every consuming repository. The historical roundtrip guard remains strict: the test now proves the rendered AGENTS block equals the v8 baseline with only the exact approved sentence replacement.

## Discarded Candidates
- Downstream AGENTS edits were rejected because `graphify codex install` would regenerate the old wording.
- A file-level `agents-md` roundtrip exemption was rejected because it could hide unrelated drift.
- Broad translation rewrites were rejected because the stale translated README lines are English code-block examples.

## Verification
- `python -m pytest tests/test_skillgen.py::test_always_on_roundtrip_is_byte_faithful -q`
- `python -m pytest tests/test_install_strings.py -k "agents_section or skill_registration"`
- `python -m pytest tests/test_install.py -k agents_install`
- `python -m pytest tests/test_skillgen.py -k always_on`
- `python -m tools.skillgen --check`
- `rg -n 'invoke the Skill tool|invoke the `skill` tool|skill: "graphify"|`skill` tool' -S . --glob '!tests/**' --glob '!graphify-out/**' --glob '!.git/**'`
- `git diff --check`
- `graphify update .`

Noted but not caused by this PR: broader local pytest over `tests/test_install_strings.py tests/test_install.py tests/test_skillgen.py` has pre-existing Windows/source-checkout failures around pytest import/source state, Windows path separator expectations, and missing installed `graphifyy` package metadata causing `__version__` to be `unknown`.

## Failure Modes / Residual Risk
Line-ending warnings appear in this Windows checkout because global `core.autocrlf` is enabled. The committed diff is limited to the intended wording, test, and generator changes.

## Reviewer Guidance
Start with `tests/test_skillgen.py::test_always_on_roundtrip_is_byte_faithful` to check the intentionally narrow historical-baseline exception, then review the AGENTS fragment and `_skill_registration()` wording.

## Out of Scope
- Changing hook behavior.
- Reworking translated README prose beyond the stale English code-block sentence.
- Fixing unrelated Windows/source-checkout test failures.

## Related Issues
None.

### DIFF
--- a/docs/translations/README.ja-JP.md
+++ b/docs/translations/README.ja-JP.md
@@ -114,7 +114,7 @@ curl -fsSL https://raw.githubusercontent.com/safishamsi/graphify/v3/graphify/ski
 
 ```
 - **graphify** (`~/.claude/skills/graphify/SKILL.md`) - any input to knowledge graph. Trigger: `/graphify`
-When the user types `/graphify`, invoke the Skill tool with `skill: "graphify"` before doing anything else.
+When the user types `/graphify`, use the installed graphify skill or instructions before doing anything else.
 ```
 
 </details>
--- a/docs/translations/README.ko-KR.md
+++ b/docs/translations/README.ko-KR.md
@@ -150,7 +150,7 @@ curl -fsSL https://raw.githubusercontent.com/safishamsi/graphify/v3/graphify/ski
 
 ```
 - **graphify** (`~/.claude/skills/graphify/SKILL.md`) - any input to knowledge graph. Trigger: `/graphify`
-When the user types `/graphify`, invoke the Skill tool with `skill: "graphify"` before doing anything else.
+When the user types `/graphify`, use the installed graphify skill or instructions before doing anything else.
 ```
 
 </details>
--- a/docs/translations/README.zh-CN.md
+++ b/docs/translations/README.zh-CN.md
@@ -110,7 +110,7 @@ curl -fsSL https://raw.githubusercontent.com/safishamsi/graphify/v3/graphify/ski
 
 ```
 - **graphify** (`~/.claude/skills/graphify/SKILL.md`) - any input to knowledge graph. Trigger: `/graphify`
-When the user types `/graphify`, invoke the Skill tool with `skill: "graphify"` before doing anything else.
+When the user types `/graphify`, use the installed graphify skill or instructions before doing anything else.
 ```
 
 </details>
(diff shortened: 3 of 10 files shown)


safishamsi:
Merged into `v8` as e3e4198 with your authorship. The host-generic wording is the right fix — telling agents to invoke a literal `skill` tool with `skill: "graphify"` only made sense on Claude and was invalid elsewhere.

One thing I had to add before it could land: the change passed `skillgen --check` and the pytest mirror, but the authoritative `--always-on-roundtrip` CLI guard (a blocking CI step) was still red, because the PR only patched the pytest test to tolerate the mismatch and not the guard itself — so the two disagreed. Rather than weaken the guard, I added an explicit, reviewable `ALWAYS_ON_SANCTIONED_EDITS` registry: the guard now applies the approved old→new substitution to the frozen baseline before the byte compare, so this exact sentence is allowed while any *other* drift still fails. CLI guard and pytest test now agree and all five skillgen guards are green. Thanks for the clean, well-scoped PR.

PR #1534, opened 2026-06-29
fix(groovy): emit inherits/implements edges for extends/implements

Synvoya:
## Summary

The Groovy extractor silently drops **every** `extends` and `implements` relationship. `contains`, methods, imports, and calls all work — but class inheritance produces no edges.

## Root cause

tree-sitter-groovy exposes inheritance through the **same** grammar shape as tree-sitter-java:

```
class_declaration <superclass=superclass, interfaces=super_interfaces>
  superclass        -> extends + type_identifier
  super_interfaces  -> implements + type_list -> type_identifier
```

But the inheritance-emitting block in `_extract_generic` is gated on `config.ts_module == "tree_sitter_java"`. Groovy (`ts_module="tree_sitter_groovy"`) never enters it — it was the only class-based JVM language in the file without an inheritance handler (Java, Kotlin, Scala, C#, Swift, C++, PHP, Python all have one).

## Fix

Widen the gate to include `tree_sitter_groovy`. The existing `_emit_java_parent_type` / `_java_collect_type_refs` path consumes the identical node shapes verbatim — no new handler needed.

```diff
-            if config.ts_module == "tree_sitter_java":
+            if config.ts_module in ("tree_sitter_java", "tree_sitter_groovy"):
```

## Verification

- Added a base class + interface to the Groovy fixture and two regression tests (`extends` -> `inherits`, `implements` -> `implements`).
- `pytest tests/test_languages.py` -> **269 passed, 13 skipped** (was 267 + my 2 new); Java inheritance tests unaffected (the shared path is reused, not changed).

## Before / after

`class UserService extends Base implements Repo {}`
- before: only `contains` edges
- after: `inherits UserService->Base`, `implements UserService->Repo`

### DIFF
--- a/graphify/extract.py
+++ b/graphify/extract.py
@@ -2801,7 +2801,7 @@ def _extract_generic(
                                                  context="generic_arg")
 
             # Java-specific: extends (superclass) / implements (interfaces) / interface-extends
-            if config.ts_module == "tree_sitter_java":
+            if config.ts_module in ("tree_sitter_java", "tree_sitter_groovy"):
                 def _emit_java_parent(base_name: str, rel: str, at_line: int) -> None:
                     if not base_name:
                         return
--- a/tests/fixtures/sample.groovy
+++ b/tests/fixtures/sample.groovy
@@ -19,3 +19,17 @@ class SampleService {
         processor.reset()
     }
 }
+
+interface Resettable {
+    void reset()
+}
+
+class ExtendedService extends SampleService implements Resettable {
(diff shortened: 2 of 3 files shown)


safishamsi:
Merged into `v8` as `64a6093` (your authorship). Confirmed the Java handler consumes the Groovy `superclass`/`super_interfaces` node shapes verbatim — extends→inherits and implements→implements both emit, Java inheritance tests unaffected. Clean one-line gate reuse. Full suite green. Ships next release.

PR #1535, opened 2026-06-29
fix(ruby): emit inherits edge for class superclass

Synvoya:
## Summary

The Ruby extractor silently drops **every** class inheritance edge. `class Dog < Animal` produces `contains`/method/call edges but no `inherits` edge.

## Root cause

Ruby exposes the base class in the `superclass` field:

```
class <name=constant, superclass=superclass, body=body_statement>
  superclass -> '<' + constant "Animal"
```

The inheritance handler in `_extract_generic` has dedicated branches for java, kotlin, c#, scala, cpp, php, swift, and python — but **none for Ruby**, so the `superclass` field is never read.

## Fix

Add a Ruby branch that reads the `superclass` field and emits an `inherits` edge, handling both forms:
- bare constant: `class Dog < Animal` -> `Animal`
- qualified: `class Cat < M::Base` (`scope_resolution`) -> `Base`

It reuses the same synthesize-base-node-if-absent + `add_edge` pattern as the other language branches.

## Verification

- Added a subclass to the Ruby fixture and a regression test.
- `pytest tests/test_languages.py` -> **268 passed, 13 skipped** (existing Ruby tests unaffected).

## Before / after

`class Dog < Animal`
- before: only `contains`
- after: `inherits Dog->Animal`

### DIFF
--- a/graphify/extract.py
+++ b/graphify/extract.py
@@ -2752,6 +2752,37 @@ def _extract_generic(
                                             add_edge(class_nid, target, "references", line,
                                                      context="generic_arg")
 
+            # Ruby: `class Dog < Animal` puts the base class in the `superclass`
+            # field (a `<` token followed by a constant or scope_resolution).
+            # There was no Ruby branch, so every Ruby inherits edge was dropped.
+            if config.ts_module == "tree_sitter_ruby":
+                sup = node.child_by_field_name("superclass")
+                if sup is not None:
+                    base = ""
+                    for sub in sup.children:
+                        if sub.type == "constant":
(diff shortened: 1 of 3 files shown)


safishamsi:
Merged into `v8` as `a19b9e9` (your authorship). Verified: exactly one `ApiClient` node (no duplicate) and the `inherits` edge targets the real class node with its source_file intact — the synth-if-absent path correctly resolves to the existing definition. Handles bare and `M::Base` qualified forms. Full suite 2767 passed. Ships next release.

PR #1538, opened 2026-06-29
fix: prune orphaned semantic cache entries after extraction

### LINKED ISSUE #1527: Semantic cache is never pruned — orphan entries accumulate unbounded (AST cache is pruned, semantic isn't)
**Version:** graphifyy 0.8.51 · macOS · Python 3.14

## Summary

The semantic extraction cache (`graphify-out/cache/semantic/`) accumulates orphaned entries without bound. When a document changes, `extract` writes a new `{file_hash}.json` and leaves the old one behind forever. The AST cache has a cleanup pass (`_cleanup_stale_ast_entries`), but there is no equivalent for the semantic cache, and the only semantic-cache removal available is `clear_cache`, which wipes everything and forces a full cold re-extraction.

## Why it matters

This is invisible for users who gitignore the cache, but it's a real problem for the documented pattern of **committing the content-addressed semantic cache** (publish `graph.json` + cache, exclude `manifest.json`) so a fresh clone/CI gets warm rebuilds — see #769, which describes exactly that setup. The committed cache then grows indefinitely as docs are edited. Observed: 152 `cache/semantic/` entries for 124 live docs after a few editing rounds (28 orphans), with nothing to reclaim them short of a full wipe.

## Reproduction

1. `graphify extract . --backend <backend>` on a corpus with documents → note the `cache/semantic/` file count.
2. Edit one document; re-extract.
3. `cache/semantic/` now holds both the old and new `{hash}.json` for that document; the old entry is never removed.

## Root cause

In `cache.py`:
- `save_semantic_cache` writes one entry per file keyed by `file_hash(content + relpath)`.
- `_cleanup_stale_ast_entries` (called from the AST path) has no semantic counterpart.
- `clear_cache` is all-or-nothing (deletes `ast/`, `semantic/`, and legacy entries).

So a changed document's prior entry is orphaned, and nothing selectively reclaims it.

## Suggested fix

A selective semantic prune, analogous to the AST cleanup: given the current detected document set, delete `cache/semantic/*.json` whose stem isn't in the live hash set

```
{ file_hash(doc, root)
  for kind in ("document", "paper", "image", "video")
  for doc in detect(root)["files"].get(kind, []) }
```

It could run at the end of `extract` (the live set is already known there) or as a `graphify cache-prune` verb. The failure mode is benign — deleting a still-live entry just re-extracts that one document on the next run.

We implemented exactly this downstream using `detect()` + `cache.file_hash()` + `cache.cache_dir(root, "semantic")`, and it keeps the committed cache equal to the live document set. Happy to send a PR if that's welcome.

guyoron1:
## Summary

Closes #1527

The semantic extraction cache (`graphify-out/cache/semantic/`) accumulates orphaned entries when documents change. The AST cache has version-based cleanup, but the semantic cache had no equivalent — orphans grew unbounded.

## Changes

- **`graphify/cache.py`**: Add `cleanup_stale_semantic_entries()` — given the live document paths, compute their hashes and remove any `cache/semantic/*.json` entry not in the live set
- **`graphify/__main__.py`**: Call the cleanup at the end of semantic extraction, print count of pruned entries

## Design

Mirrors the existing `_cleanup_stale_ast_entries` concept:
- Computes live hashes from the detected document set (already known at the call site)
- Removes orphaned `{hash}.json` files whose hash no longer matches any live file
- Failure mode is benign — accidentally deleting a live entry just re-extracts on next run
- Best-effort: OSError during unlink is silently ignored (retried next run)

## Test plan

- [ ] Extract a corpus with documents → note `cache/semantic/` file count
- [ ] Edit one document, re-extract → old orphan should be removed, count stays at live doc count
- [ ] Delete a document, re-extract → its cache entry should be pruned
- [ ] Happy path: unchanged documents still served from cache (no re-extraction)

### DIFF
--- a/graphify/__main__.py
+++ b/graphify/__main__.py
@@ -4546,6 +4546,7 @@ def main() -> None:
         # Semantic extraction on docs/papers/images. Check cache first.
         from graphify.cache import (
             check_semantic_cache as _check_semantic_cache,
+            cleanup_stale_semantic_entries as _cleanup_stale_semantic,
             save_semantic_cache as _save_semantic_cache,
         )
         sem_result: dict = {
@@ -4636,6 +4637,13 @@ def main() -> None:
                 sem_result["hyperedges"].extend(fresh.get("hyperedges", []))
                 sem_result["input_tokens"] += fresh.get("input_tokens", 0)
                 sem_result["output_tokens"] += fresh.get("output_tokens", 0)
+        if semantic_files:
+            try:
+                pruned = _cleanup_stale_semantic(sem_paths_str, root=out_root)
+                if pruned:
+                    print(f"[graphify extract] pruned {pruned} orphaned semantic cache entries", file=sys.stderr)
+            except Exception:
+                pass
         stages.mark("semantic extract")
 
         pg_result: dict = {"nodes": [], "edges": []}
--- a/graphify/cache.py
+++ b/graphify/cache.py
@@ -474,3 +474,44 @@ def save_semantic_cache(
             save_cached(p, result, root, kind="semantic")
             saved += 1
     return saved
+
+
+def cleanup_stale_semantic_entries(
+    live_paths: list[str],
+    root: Path = Path("."),
+) -> int:
+    """Remove orphaned semantic cache entries not matching any live file.
+
+    When a document changes, ``save_semantic_cache`` writes a new
+    ``{hash}.json`` keyed on the updated content, but the old entry lingers
+    forever.  This mirrors the AST-side ``_cleanup_stale_ast_entries`` concept
+    for the semantic cache: given the current detected document set, delete
+    ``cache/semantic/*.json`` whose stem is not in the live hash set.
+
+    The failure mode is benign — deleting a still-live entry just re-extracts
+    that document on the next run (#1527).
+    """
+    sem_dir = cache_dir(root, "semantic")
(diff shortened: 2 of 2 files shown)


safishamsi:
Thank you for this — you correctly identified the asymmetry (AST cache swept, semantic cache never pruned) and the right shape of the fix (compute live hashes from the detected docs, remove `cache/semantic/*.json` not in the set, mirror the AST cleanup, benign failure mode). The fix for #1527 shipped on `v8` (7a9cda2) along these exact lines, and apologies for the parallel work.

One reason it went in as a separate implementation: the live set has to be computed from the **full** detected document set, not the per-run semantic set. In incremental mode (`__main__.py:4413-4430`) `semantic_files` is only the *changed/new* docs (`new_by_type`), while `files_by_type` is the full set. Pruning against `sem_paths_str` (derived from `semantic_files`) would, on an incremental run touching one doc, delete every *unchanged* doc's still-valid cache entry — defeating the cache and re-billing LLM extraction on the next run. The shipped version keys the live set off `files_by_type`, runs even when nothing changed (so deletion-orphans are swept too), and adds regression tests covering orphan removal, live-entry preservation, deleted files, and ast/tmp isolation.

Sorry again for not catching your PR first. Closing as fixed in `v8`; the credit for surfacing #1527 and the prune approach is yours.

Issue #1529, opened 2026-06-29
import edges silently dropped on Next.js codebase — short IDs vs full-path node IDs (regression of #256)


FahadBinHussain:
i'm hitting the same class of bug as #256 on v0.9.1 — all import edges are silently dropped because the edge target IDs don't match the file node IDs.

**repro**

- 216-code-file Next.js/TypeScript codebase (app router, `app/`, `components/`, `lib/`, `prisma/`, `scripts/`)
- `graphify .` with node_modules excluded via `extra_excludes`
- ast extraction produces 168 nodes and 260 edges
- after `build_from_json()`: 168 nodes, **0 edges** — every single edge dropped

**root cause (same as #256)**

edge targets use short names from import specifiers:
```json
{"source": "app_about_page_ts", "target": "about_page", "relation": "imports_from"}
{"source": "app_admin_articles_new_newarticleclient_ts", "target": "components_languageprovider_uselanguage", "relation": "imports_from"}
```

but file node IDs use full paths:
```json
{"id": "app_about_page_ts"}
{"id": "components_languageprovider_uselanguage_ts"}
```

`about_page` doesn't match `app_about_page_ts`. `components_languageprovider_uselanguage` doesn't match `components_languageprovider_uselanguage_ts`. so `build_from_json` drops all 260 edges as "external/stdlib".

the v0.4.3 fix resolved **relative imports** (`./foo.js` → full path), but these edge targets look like they're generated from import specifier stems without resolving to the full filesystem path. possibly bare imports or next.js alias patterns (`@/components/...`).

**what i tried**

- v0.8.51 → same issue
- upgraded to v0.9.1 → same issue
- post-build fuzzy match workaround (stem matching) recovered 110/260 edges

**environment**

- graphifyy 0.9.1
- windows 11
- next.js 15, typescript 5.9
- python 3.13 (uv tool install)

safishamsi:
Confirmed and fixed on `v8` (2133539). You were exactly right that it's a short-ID vs full-path-ID mismatch, and it was a regression from the 0.9.0 full-repo-relative node-ID migration.

[...]

PR #1541, opened 2026-06-29
fix(imports): resolve workspace subpath exports (#1308)

### LINKED ISSUE #1308: TypeScript workspace imports do not resolve package subpath exports

guyoron1:
## Summary
- Workspace imports with subpath exports (e.g. `import { value } from "@example/pkg-a/browser"`) now resolve through the `package.json` `exports` map instead of falling back to a bare path string
- Supports string values, condition objects (`source > import > module > default > require > svelte > types`), nested conditions, and wildcard patterns (`"./*": "./src/*.ts"`)
- Normalizes import node IDs to use `_file_stem` instead of full paths across JS/Svelte/Astro/Vue extractors

Closes #1308

## Test plan
- [x] 4 new tests: string subpath, condition object, wildcard, filesystem fallback
- [x] All 10 workspace tests pass
- [x] All 38 import extension resolution tests pass
- [x] Full 34-test JS import resolution suite passes — 0 regressions

### DIFF
--- a/graphify/extract.py
+++ b/graphify/extract.py
@@ -356,6 +356,26 @@ def _load_workspace_packages(start_dir: Path) -> dict[str, Path]:
     return packages
 
 
+_EXPORT_CONDITION_PRIORITY = (
+    "source", "import", "module", "default", "require", "svelte", "types",
+)
+
+
+def _resolve_export_target(value: Any) -> str | None:
+    if isinstance(value, str):
+        return value
+    if isinstance(value, dict):
+        for cond in _EXPORT_CONDITION_PRIORITY:
+            v = value.get(cond)
+            if isinstance(v, str):
+                return v
+            if isinstance(v, dict):
+                nested = _resolve_export_target(v)
+                if nested:
+                    return nested
+    return None
+
+
 def _package_entry_candidates(package_dir: Path, subpath: str) -> list[Path]:
     manifest = package_dir / "package.json"
     manifest_data: dict[str, Any] = {}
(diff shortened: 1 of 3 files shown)


safishamsi:
Thanks for this — the `exports`-map subpath resolver is a genuinely useful addition and it shipped on `v8` (e8dabad) with your authorship. It resolves `@scope/pkg/subpath` imports through the package's `exports` map (string values, condition objects, nested conditions, `./*` wildcards), which graphify couldn't do before.

I landed the resolver portion only, with two changes:
- Dropped the import-node-ID normalization (`_make_id(str(...))` → `_make_id(_file_stem(...))`) that the PR also included. Current `v8` already fixes the node-ID mismatch a different, more complete way — commit 2133539 (#1529), whose id-remap post-pass registers the absolute-resolved forms and handles the relative-input alias case. Keeping the `_file_stem` change on top regressed `test_alias_import_edge_resolves_with_relative_input_paths` (it changed the key the post-pass matches on, silently disabling the remap), so it was left out.
- Two hardening tweaks: `default` is now consulted LAST in the condition priority (Node's catch-all), and export targets that escape the package dir are rejected (a `"./x": "../../../etc/..."` value can no longer resolve to a file outside the package).

Brought your 4 subpath tests plus two more (path-escape, default-last). Full suite green (2551), and the #1529 test still passes. Closing as shipped — credit for #1308 is yours. Thank you.

Issue #1475, opened 2026-06-26
Bug: ObjC extractor — 4 bugs causing ~60% of relationships to be silently dropped


JabberYQ:
## Summary

Testing `/graphify .` on a pure Objective-C iOS project (12 files, ~1,125 words) revealed 6 bugs in the ObjC extractor that collectively cause ~60% of code-level relationships to be silently dropped.

Test repo structure:
```
Demo/
  AppDelegate.h/.m
  Product.h/.m
  ProductListViewController.h/.m
  ProductDetailViewController.h/.m
  AlertManager.h/.m
  main.m
```

Expected: `@interface` → `inherits` edges, `[self method]` → `calls` edges, `self.product.name` → property access edges, `addTarget:action:` → callback edges, `@property NSArray<X*> *` → `references` edges, `+classMethod` labeled correctly.
Actual: 37 nodes, 30 edges, 11 communities — but all `calls` edges are missing, .h files are extracted as C (0 ObjC edges), property accesses & target-action are invisible, generic property types are invisible, and class methods are mislabeled.

---

## Bug 1: `.h` files dispatched to C extractor instead of ObjC extractor

**Root cause:** `_DISPATCH` mapping at `extract.py:12210`:

```python
".h": extract_c,     # ← ObjC headers parsed as C!
".m": extract_objc,
".mm": extract_objc,
```

**Impact:** Every `.h` file in an ObjC project is parsed by `extract_c`, which doesn't understand `@interface`, `@property`, `@protocol`, or method declarations. The C extractor produces exactly 1 node (the file itself) and 0 edges.

**What's lost per `.h` file:**
- `@interface Foo : Bar` → no `inherits` edge
- `<Protocol>` conformance → no `implements` edge
- `@property (strong) SomeType *prop` → no `references` edge
- Method declarations → no nodes

**Verification:**
```python
# extract_c on AppDelegate.h → 1 node, 0 edges
# extract_objc on AppDelegate.h → 5 nodes, 5 edges
#   (AppDelegate, UIResponder, UIApplicationDelegate, UIWindow
#    + inherits/implements/references edges)
```

**Suggested fix:** In the dispatch logic, detect ObjC `.h` files (by `@interface` content heuristic, or when sibling `.m`/`.mm` files exist) and route them to `extract_objc`.

---

## Bug 2: `walk_calls` never matches — 0 `calls` edges for all ObjC projects

[...]

Issue #543, opened 2026-04-25
Cross-file INFERRED `calls` resolution lacks import evidence + ambiguous-candidate handling — short common names blow up god_nodes ranking


AlphaMinusOne:
## Summary

The current cross-file `calls` resolution (`extract.py` around L3345-L3374) builds a **global name map** keyed by `label.strip(\"()\").lstrip(\".\").lower()`, then for each unresolved call it adds an INFERRED `calls` edge to **any** node whose label matches that key. There's no import-evidence check, no multi-candidate disambiguation, and confidence is hard-coded to `0.8`.

For corpora with common short names (`log`, `execute`, `mutate`, `replace`, `find`, `optional`, `check`, `main`), this systematically inflates the degree of harmless local helpers until they become the top god_nodes — even though the true call graph never reaches them.

`god_nodes()` itself already filters file-level hubs / method stubs / isolated stubs (`_is_file_node`, `_is_concept_node`), but it does not consult edge confidence or import evidence, so polluted nodes pass straight through.

This is **not** the same as the calibration issue in #540 — recalibrating the `0.8` score won't change degree, and `god_nodes` ranks by raw degree.

## Reproducer (real corpus, 2026-04-25)

TypeScript ERP repo, ~3080 nodes, 4218 edges, refreshed via `/graphify-refresh`. Top 8 god_nodes:

| Node | Total degree | Cross-file INFERRED `calls` | Same-file edges | EXTRACTED edges |
|---|---|---|---|---|
| `log()` | 150 | **144** | 1 | 6 |
| `execute()` | 80 | **79** | 1 | 1 |
| `createRawMcpServer()` | 78 | **76** | 2 | 2 |
| `registerAllOriginalTools()` | 76 | **74** | 2 | 2 |
| `optional()` | 69 | **67** | 2 | 2 |
| `mutate()` | 66 | **64** | 2 | 2 |
| `replace()` | 49 | **46** | 3 | 3 |
| `find()` | 30 | **28** | 1 | 2 |

Concrete example: `log()` is a one-line arrow function at `openclaw/sync.mjs:99`:

```js
const log = (msg) => process.stdout.write(`[sync] ${msg}\n`);
```

It has no `export`. It cannot be called by `scripts/check-cipl-data.ts:check()` or any of the other 130 distinct files the graph claims call it. All 144 cross-file INFERRED edges are spurious — the global name map matched `log` against `console.log`-style usage in unrelated scripts.

[...]

PR #1543, opened 2026-06-29
fix(objc): resolve dot-syntax property access and @selector() calls (#1475)

guyoron1:
## Summary
- Add `field_expression` handler in `walk_calls` so dot-syntax property access (`self.name`, `self.product.name`) produces `calls` edges — previously invisible because the grammar emits `field_identifier`, not a message send
- Add `selector_expression` handler so `@selector(method)` compile-time references produce `calls` edges — previously unhandled node type
- Both use the same suffix-match resolution strategy as the existing `message_expression` handler

Closes the remaining two sub-bugs from #1475 (Bug 5: dot-syntax, Bug 6: @selector).

## Test plan
- [x] `test_objc_dot_syntax_property_calls_edge` — verifies `self.name` resolves to a calls edge
- [x] `test_objc_selector_expression_calls_edge` — verifies `@selector(fetch)` resolves to a calls edge
- [x] All 17 ObjC tests pass
- [x] Full test suite: 269 passed, 13 skipped, 0 failures

### DIFF
--- a/graphify/extract.py
+++ b/graphify/extract.py
@@ -9723,7 +9723,7 @@ def extract_objc(path: Path) -> dict:
     nodes: list[dict] = []
     edges: list[dict] = []
     seen_ids: set[str] = set()
-    method_bodies: list[tuple[str, Any]] = []
+    method_bodies: list[tuple[str, Any, str]] = []
 
     def add_node(nid: str, label: str, line: int) -> None:
         if nid not in seen_ids:
@@ -9937,7 +9937,7 @@ def extract_objc(path: Path) -> dict:
                 add_node(method_nid, f"{prefix}{method_name}", line)
                 add_edge(container, method_nid, "method", line)
                 if t == "method_definition":
-                    method_bodies.append((method_nid, node))
+                    method_bodies.append((method_nid, node, container))
             return
 
         for child in node.children:
@@ -9947,8 +9947,13 @@ def extract_objc(path: Path) -> dict:
 
     # Second pass: resolve calls inside method bodies
     all_method_nids = {n["id"] for n in nodes if n["id"] != file_nid}
+    class_method_nids: dict[str, set[str]] = {}
+    for m_nid, _, container_nid in method_bodies:
+        class_method_nids.setdefault(container_nid, set()).add(m_nid)
     seen_calls: set[tuple[str, str]] = set()
-    for caller_nid, body_node in method_bodies:
+    for caller_nid, body_node, container_nid in method_bodies:
+        sibling_nids = class_method_nids.get(container_nid, set())
+
         def walk_calls(n) -> None:
             if n.type == "message_expression":
                 # `[[Foo alloc] init]` is a message_expression whose method is the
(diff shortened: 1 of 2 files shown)


safishamsi (changes requested):
Thanks for tackling the two deferred ObjC follow-ups. A few changes needed before this can land.

**1. Drop the #1308 commit from this PR.** It re-includes the workspace `exports`-map work, but that already shipped on `v8` (e8dabad, from your #1541) — and the merged version has two things this copy is missing: a package-dir containment guard (so an `exports` target like `"./x": "../../../etc/..."` can't resolve outside the package) and `default` consulted last in the condition order. The duplicate is also what's causing all the merge conflicts here. Please rebase on current `v8` and reduce this PR to the ObjC changes only.

**2. The dot-syntax / `@selector` resolution needs a scoping guard — this is the blocker.** These two were deferred specifically because resolving an untyped name globally floods false edges. Right now the resolution suffix-matches the property/selector name against every method id, with no receiver-type restriction and no single-definition check. On a fixture with two classes that each declare a property named `name`, one `self.product.name` access fans out to *both* `-name` methods (plus the `product` class node) — four edges from one access. `@selector(doThing)` with two `doThing` methods resolves to both. That's the god-node inflation (#543/#1219) the deferral was meant to avoid.

Please resolve only to an unambiguous, in-scope definition:
- for `self.<prop>`, restrict to the current class's own declared properties / the known receiver type;
- for `@selector(...)`, emit only when exactly one method matches;
- skip emission entirely when 0 or >1 candidates match (and don't match property names against class node ids).

**3. Use `accesses` for dot-syntax, not `calls`.** The issue scopes `self.product.name` to a property-access edge; the PR currently emits `calls`.

**4. Add a fan-out regression test.** The current tests are happy-path (one uniquely-named method, `any(... in ...)` assertions) and wouldn't catch the flooding above. Please add a two-same-named-property fixture asserting the exact edge count (no fan-out), plus an unresolvable-property case asserting zero edges, and the same for a two-same-named `@selector`.

For reference, the alloc/init reference edges that just landed (#1475 on `v8`) use the unique-stub guard (`ensure_named_node` + `_rewire_unique_stub_nodes`, which only resolves when exactly one class of that name exists) — the same "resolve only when unambiguous" idea applies here. The new `walk_calls` branches otherwise coexist fine with that change, so once the guard and the rebase are in, this should be a clean ObjC-only PR. Thanks again.

guyoron1:
Hey !!!
Thank you so much  the detailed review! 
All four points are now addressed:

1. **Dropped #1308 commit** — rebased on current v8. The PR is now a single ObjC-only commit. 

2. **Scoping guard added** — the blocker is fixed:
   - `self.<prop>` (field_expression): resolution is restricted to **sibling methods within the same class** (via a `class_method_nids` index built from the method_bodies container). Emits only when **exactly 1** sibling matches; skips on 0 or >1.
   - `@selector(...)` (selector_expression): emits only when **exactly 1** method matches across the entire file; skips on 0 or >1.
   - Neither handler matches property names against class node IDs.

3. **`accesses` for dot-syntax** — field_expression edges use `relation: "accesses"` (not `calls`). @selector keeps `calls` with `context: "call"`.

4. **Fan-out regression tests added** — five new tests:
   - `test_objc_dot_syntax_property_accesses_edge` — happy path, exactly 1 accesses edge
   - `test_objc_dot_syntax_no_fanout_two_same_named_properties` — two classes with `-name`, asserts exactly 2 scoped accesses (one per class), no cross-class fan-out
   - `test_objc_dot_syntax_unresolvable_property_zero_edges` — `self.missing` → 0 edges
   - `test_objc_selector_expression_calls_edge` — unique `@selector(fetch)` → 1 calls edge
   - `test_objc_selector_no_fanout_two_same_named_methods` — `@selector(doThing)` with 2 classes having `-doThing` → 0 edges

All tests pass.

safishamsi:
Merged into `v8` (0792b41) with your authorship — thanks for the rework. The scoping (sibling-only for dot-syntax, unique-match for @selector) and the move to `accesses` are exactly right.

One change I made on top: the resolver matched the method node id by `endswith` (suffix substring), which had two bugs your fixtures didn't cover — `self.name` would falsely resolve to a sibling `-surname` (`...surname` ends with `name`), and a substring-colliding sibling would suppress the correct edge. Since a method id is `_make_id(container, name)`, I switched both branches to exact-id matching and added substring-collision regression tests (`-name`/`-surname`, `-doThing`/`-reallyDoThing`). Verified end-to-end. Ships next release.

PR #1554, opened 2026-06-30
fix(hooks): limit Windows hook rebuild workers

matiasduartee:
## Summary

This fixes a Windows/Git Bash/MSYS-specific failure mode in Graphify git hooks.

When the generated post-commit hook launches the detached background rebuild on Windows, the rebuild can reach AST extraction and spawn ProcessPoolExecutor children. In some Git-for-Windows/MSYS or GUI-client environments, the Python child process inherits fragile pipe handles and the rebuild reports a Windows startup failure similar to:

`	ext
error 2147942632 (0x800700e8) starting <python> -c from multiprocessing.spawn import spawn_main; spawn_main(...) --multiprocessing-fork
`

The commit itself can still succeed, but the hook-triggered graph rebuild becomes noisy or unreliable.

## Fix

- In generated post-commit and post-checkout hooks, detect Windows/MSYS via WINDIR or MSYSTEM.
- Default hook-triggered rebuilds to GRAPHIFY_MAX_WORKERS=1 in those environments.
- Preserve explicit user configuration: if GRAPHIFY_MAX_WORKERS is already set, the hook does not override it.
- Leave Linux/macOS behavior unchanged because those variables are absent.
- Keep the existing Windows-path rejection for POSIX/WSL core.hooksPath, but do not reject native Windows paths when running on Windows itself.

## Why this shape

The hook remains installed and keeps the graph updated after commits/checkouts. The change only avoids multiprocessing from the Git hook environment on Windows by default, which is the fragile part. Users who prefer parallel hook rebuilds can still opt back in by setting GRAPHIFY_MAX_WORKERS explicitly.

## Tests

`	ext
python -m pytest tests/test_hooks.py -q
# 42 passed

python -m pytest tests/test_extract.py -q -k BrokenProcessPool or windows_spawn or parallel
# 2 passed, 98 deselected

git diff --check
# clean
`

I also ran a smoke test in a temporary git repo: installed the Graphify hooks from this branch, verified the generated post-commit hook contains the Windows/MSYS worker default, made a real commit, and confirmed the detached rebuild completed successfully.

### DIFF
--- a/graphify/hooks.py
+++ b/graphify/hooks.py
@@ -1,6 +1,7 @@
 # git hook integration - install/uninstall graphify post-commit and post-checkout hooks
 from __future__ import annotations
 import configparser
+import os
 import re
 import sys
 from pathlib import Path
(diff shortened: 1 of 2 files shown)


safishamsi:
Merged into `v8` as `879c058` (your authorship). Both parts verified: hook scripts default GRAPHIFY_MAX_WORKERS=1 on Windows/MSYS (explicit value still wins), and the Windows-path guard correctly no-ops on native Windows where such paths are legitimate. Full suite green. Ships next release.

Issue #1316, opened 2026-06-14
Member calls via injected fields (`this.db.query()`) do not produce class→dependency `calls` edges (TS/NestJS)


kairostm-ia:
## Summary

When a TypeScript/NestJS service injects a dependency and invokes it through a member call (e.g. `this.db.query(...)`), the extractor does not emit a class-level `calls` edge from the service class to the injected dependency. The dependency relationship is only partially captured, which leaves the class-to-dependency link absent from `graph.json`.

## Environment

- **graphifyy:** 0.8.38
- **Python:** 3.12
- **Install method:** `uv tool install graphifyy`

## Expected behaviour

A NestJS service that injects `DatabaseService` in its constructor and calls `this.db.query(...)` should yield a class-level `calls` edge `MyService → DatabaseService` in `graph.json` (relation `calls`).

## Actual behaviour

A fresh extraction produces **zero** class→`DatabaseService` `calls` edges. The dependency is captured only indirectly:

- via an `imports` edge (file → `DatabaseService`), and
- via a `references` edge (constructor parameter type → `DatabaseService`).

The convenience class-level `calls` edge is missing, so consumers that rely on class-to-class call relationships (e.g. god-node analysis, impact analysis) do not see the service as a caller of its database dependency.

## Minimal reproduction

```ts
// database.service.ts
export class DatabaseService {
  query(sql: string) { /* ... */ }
}

// crm.service.ts
export class CrmService {
  constructor(private db: DatabaseService) {}
  async getOrders() {
    return this.db.query('SELECT 1');
  }
}
```

Run extraction and inspect the `links` array in `graph.json`. There is no edge with `source: crmservice`, `target: databaseservice`, `relation: calls`.

## Root cause (located)

In `extract.py`, within `walk_calls` (approximately L3298–3314), `calls` edges are constructed such that they:

1. use **`caller_nid` = the enclosing method node**, never the enclosing class node; and
2. resolve the call target by **bare callee name** (`"query"`), with no receiver-type resolution — so `this.db.query()` cannot be mapped back to the `DatabaseService` type.

Because of (2), the bare name `query` is generic and either fails to resolve or is ambiguous, and no edge to `DatabaseService` is produced at any level.

## Suggested fix

[...]

PR #1555, opened 2026-06-30
fix(extract): resolve this.field.method() calls via constructor injection types (#1316)

guyoron1:
## Summary

- Resolves `this.field.method()` calls (e.g., `this.repo.findById()`) to proper `calls` edges by leveraging TypeScript constructor parameter-property type annotations
- Adds `call_accessor_object_field` to JS/TS/TSX configs, enabling receiver capture in `walk_calls`
- Extends `walk_calls` to detect the `this.field.method()` AST pattern (nested `member_expression` with inner `object=this`), preventing false-positive in-file matches
- Builds a per-file type table from TS constructor parameter properties (parameters with `accessibility_modifier`), used by a new `_resolve_typescript_member_calls` cross-file resolver with god-node guard
- Adds `method_signature` to TS `function_types` so interface method declarations (e.g., `IUserRepository.findById()`) are extracted as nodes

## Test plan

- [x] `test_ts_constructor_injection_calls_edge` — two-file cross-file test verifying `getUser()` → `findById()` calls edge through constructor injection type resolution
- [x] `test_ts_this_field_receiver_not_same_file_collision` — verifies `this.db.query()` does NOT false-match a bare `query()` function in the same file
- [x] All 269 existing TS/JS tests pass (0 regressions)

### DIFF
--- a/graphify/extract.py
+++ b/graphify/extract.py
@@ -2147,6 +2147,7 @@ _JS_CONFIG = LanguageConfig(
     call_function_field="function",
     call_accessor_node_types=frozenset({"member_expression"}),
     call_accessor_field="property",
+    call_accessor_object_field="object",
     function_boundary_types=frozenset({"function_declaration", "arrow_function", "method_definition"}),
     import_handler=_import_js,
 )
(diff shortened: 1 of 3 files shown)


safishamsi:
Merged into `v8` (1801da0) with your authorship. Thanks — the implementation correctly reuses the single-definition god-node guard (same discipline as the Swift/Python/Ruby member-call resolvers), so `this.db.query()` resolves to the constructor-typed class only. I added the decisive regression tests it needed: a two-classes-same-method case (resolves to the injected type only, not the collision) and an ambiguous-type case (no edge). Verified end-to-end. Scope is constructor parameter-property injection (not separate `private db: Database;` field declarations) — a reasonable follow-up if wanted. Ships next release.

Issue #1547, opened 2026-06-29
Bug: paired-cpp creates islands, never showing how they connect with other classes.


c0dezer019:
# Cross-file `#include` import edges are silently dropped when a header has a paired `.cpp`

## Description

For C/C++ corpora, an `#include "foo.h"` produces an `imports` edge only when `foo.h` is **header-only**. If a sibling `foo.cpp` exists, graphify mints the header's node id in a different (doubled, translation-unit-suffixed) form than the one its `#include` resolver targets. The two ids never match, so the import edge has no valid endpoint and is dropped at build time.

The result: any header that has a paired implementation file appears as an isolated island — no incoming edges from the files that `#include` it — even though those include relationships exist in the source. The drop is silent (no warning, no error), and the graph health check does not flag it because the edge is discarded during build rather than left dangling.

## Minimal reproduction

Create this tree:

```
repro/
├── bar.h          # header-only (no bar.cpp)
├── widget.h       # has a paired widget.cpp
├── widget.cpp
└── consumer.cpp   # includes BOTH headers
```

**`repro/bar.h`**
```cpp
#ifndef BAR_H
#define BAR_H
struct Bar { int x; };
#endif
```

**`repro/widget.h`**
```cpp
#ifndef WIDGET_H
#define WIDGET_H
class Widget {
public:
    static int make();
};
#endif
```

**`repro/widget.cpp`**
```cpp
#include "widget.h"
int Widget::make() { return 42; }
```

**`repro/consumer.cpp`**
```cpp
#include "bar.h"
#include "widget.h"
int use() { return Widget::make(); }
```

Then build the graph:

```bash
graphify repro/
```

## Expected

`consumer.cpp` should have an `imports` edge to **both** `bar.h` and `widget.h`.

## Actual

`consumer.cpp` has an `imports` edge to `bar.h` only. The edge to `widget.h` is missing; `widget.h` + `widget.cpp` form an island with no incoming edges.

## Verification snippet

Run from the directory containing `graphify-out/` (edges are stored under the `links` key):

```python
import json, os, glob
g = json.load(open('graphify-out/graph.json'))
ids = set(n['id'] for n in g['nodes'])

# Reproduce the canonical id the #include resolver targets, and check it exists.
for h in sorted(glob.glob('repro/**/*.h', recursive=True)):
    base = h[:-2]                                   # strip ".h"
    has_cpp = os.path.exists(base + '.cpp')
[...]

PR #1562, opened 2026-06-30
Namespace-aware C# type resolution with lexical using-scope and qualified refs

TheFedaikin:
## Summary

  The C# cross-file resolver was sound but conservatively incomplete: it resolved
  the common case and *skipped* three reference classes it couldn't faithfully
  represent. This turns those three skips into real resolution while keeping the
  same hard guarantee — **never emit a wrong edge**: resolve only on exact, unique
  facts, otherwise leave the reference dangling.

  ## What changed

  **1. Namespace in the C# type-node id.** A C# type node's id was built from the
  file stem + simple name, so two same-named types in different namespaces of one
  file (`namespace A { class T } namespace B { class T }`) collapsed to one node
  and the second was dropped. The dotted namespace is now folded into the id at
  the three sites that construct or probe it (class declaration,
  `ensure_named_node`, base-list). `make_id` drops the empty namespace segment, so
  every non-C# id and every C# global-namespace id stays byte-identical — the
  change is opt-in for C#. The previous `ns_collision` detect-and-skip workaround
  is removed.

  **2. Lexical `using`-scope.** C# `using` scope is lexical (per declaration
  block), not per namespace name — `namespace A { using N; class Good : T {} }
  namespace A { class Bad : T {} }` rejects `Bad : T` even though both blocks are
  named `A`. The previous file-wide clamp, keyed on unique namespace strings, was
  unsound for that case. Each declaration node now carries a per-file lexical
  scope-id chain, and each `using` edge carries its `scope_kind` (file vs
  namespace) and `scope_id`. A `using` — namespace or alias — applies to a
  reference iff it is file-scoped or its scope id is in the reference's scope
  chain, matched within the same file. This is sound for sibling same-name blocks,
  nested blocks (a using flows inward), file-level usings, and aliases, and it
  recovers usings in multi-namespace files that were previously dropped. The
  scenarios were verified against `dotnet build`.

  **3. Qualified references.** Qualified references (`B.T`) were skipped and
  qualified generics (`A.B<C>`) emitted junk labels. The qualifier is now
  preserved on the edge (`metadata.ref_qualifier`; `ref_token` stays the simple
  lookup key), and a qualified `Q.T` resolves via an in-scope namespace alias
  (`using Q = X.Y`) or an exact known namespace, otherwise dangles. Generic args
  are stripped from qualified tails so `N.Box<int>` resolves to the real `N.Box`.
  An in-scope alias shadows a like-named namespace.

  ## Soundness

  Every emitted edge is exact and uniquely scoped; uncertainty dangles rather than
  guessing. The `ns_collision` removal and def-index completion (which expose the
  de-collapsed types) land in the same change as the lexical resolver (which makes

  13 new C# resolution tests cover id-distinctness, same-name-in-one-file
  resolution, the four lexical `using` cases, and qualifier resolution (generics,
  namespace aliases, and out-of-scope / shadowing alias edge cases). Three
  superseded tests were rewritten around the new behavior. The full graphify suite
  passes.

  ## Residual gaps (sound, deferred)

  - Partial qualifiers (`Core.T` meaning `Game.Core.T` via `using Game;`) —
    exact-match only; prefix composition is deferred.
  - Ambiguous same-name in-scope aliases dangle (innermost-shadows-outer is not
    modeled).
  - Nested-type identity and member-call resolution remain follow-ups.

### DIFF
--- a/graphify/extract.py
+++ b/graphify/extract.py
@@ -29,10 +29,14 @@ from graphify.extractors.base import (  # noqa: F401
     _read_text,
 )
 from graphify.extractors.blade import extract_blade  # noqa: F401
-from graphify.extractors.csharp import _resolve_csharp_type_references
+from graphify.extractors.csharp import (
+    _resolve_cross_file_csharp_imports,
+    _resolve_csharp_type_references,
+)
 from graphify.extractors.elixir import extract_elixir  # noqa: F401
 from graphify.extractors.razor import extract_razor  # noqa: F401
 from graphify.extractors.zig import extract_zig  # noqa: F401
+from graphify.security import sanitize_metadata
 
 _RECURSION_LIMIT = 10_000
 
@@ -74,6 +78,11 @@ def _file_node_id(rel_path: Path) -> str:
     ID semantic subagents generate, or AST and semantic extraction split a file
     into two disconnected ghost nodes (#1033)."""
     return _make_id(_file_stem(rel_path))
+
+
+def _csharp_namespace_id(dotted_name: str) -> str:
+    digest = hashlib.sha1(dotted_name.encode("utf-8")).hexdigest()[:16]
+    return f"csharp_namespace:{digest}"
 
 
 _TSCONFIG_ALIAS_CACHE: dict[str, dict[str, list[str]]] = {}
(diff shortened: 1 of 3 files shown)


safishamsi:
Merged into `v8` (with your authorship) — thanks, this is a clean advance on the #1466 resolver. Folding the namespace into the node id (dropping the multi-namespace detect-and-skip), the lexical per-block using-scope, and qualified-reference resolution all landed. I verified the false-edge discipline holds: `using A` resolves `Widget` to A.Widget only (not B), an ambiguous `using A; using B;` resolves to nothing rather than fanning out, qualified `B.Widget` resolves regardless of usings, and sibling-block using-scope stays isolated — no dangling edges.

One note: your branch predated the C++/ObjC cross-file member-call resolvers that landed on `v8` (#1547/#1556), so a straight merge was a deceptively-clean 3-way (it kept that newer code as 'theirs'). I reconciled it onto current `v8` and re-ran the full suite there to confirm both features coexist — all green. Ships next release.

PR #1565, opened 2026-06-30
feat(extract): capture indirect dispatch as indirect_call edges

sheik-hiiobd:
## Problem

`affected` (blast radius) is sound for direct calls but blind to **indirect dispatch**. A function passed *by name* as a call argument — `executor.submit(fn)`, `Thread(target=fn)`, `map(fn, xs)` — is a real dependency, but the callee-only scan in `walk_calls` never records it, so those callers are dropped from `affected`.

Minimal repro (0.9.3):

```python
def handler(x): return x * 2
def direct():         return handler(1)         # caught
def via_submit(pool): pool.submit(handler, 1)   # MISSED
```

`graphify affected handler` lists `direct()` but omits `via_submit()`. An under-counting blast radius reads complete while dropping exactly the callers where regressions hide.

## Change

- In `walk_calls` (Python), scan a call's `arguments` for resolvable function-name identifiers (positional + `keyword_argument` values) and emit a **distinct `indirect_call` relation**. Keeping it separate leaves strict `calls` queries precise while `affected`/blast-radius picks up the edge.
- Add `indirect_call` to `DEFAULT_AFFECTED_RELATIONS`.
- Dedup-safe (won't duplicate an existing direct `calls` edge); `confidence="INFERRED"`, `context="argument"`.

After: `affected handler` -> `direct`, `via_submit`, `via_thread`.

## Scope / limits (intentional)

- **Python only** for now; the same pattern extends to the other tree-sitter grammars.
- Covers the call-argument case (executors, threads, callbacks, `map`/`filter`). Dispatch via **dict literals** (`{k: fn}`), **`getattr`-by-string**, and **decorators** lives in other AST nodes and is left for a follow-up.
- Same-file resolution via the in-file `label_to_nid`; cross-file indirect refs can later reuse the existing `raw_calls` path.

## Tests

`tests/test_indirect_dispatch.py` — the indirect callers get `indirect_call` edges, the direct caller stays a `calls` edge (no relation regression), and `affected(handler)` now includes the dispatchers. Full suite green; `ruff` clean.

### DIFF
--- a/graphify/affected.py
+++ b/graphify/affected.py
@@ -11,6 +11,7 @@ import networkx as nx
 
 DEFAULT_AFFECTED_RELATIONS = (
     "calls",
+    "indirect_call",
     "references",
     "imports",
     "imports_from",
--- a/graphify/extract.py
+++ b/graphify/extract.py
@@ -4314,6 +4314,44 @@ def _extract_generic(
                         rc_entry["lang"] = "cpp"
                     raw_calls.append(rc_entry)
 
+            # Indirect dispatch: a function passed BY NAME as a call argument
+            # (executor.submit(fn), Thread(target=fn), map(fn, xs)) is a real dependency
+            # the callee-only scan above can't see. Emit it as a distinct `indirect_call`
+            # relation so strict `calls` queries stay precise while affected/blast-radius
+            # picks up the edge. Python only for now; dispatch via dict literals, getattr
+            # or decorators lives in other AST nodes and is left to a follow-up.
+            if config.ts_module == "tree_sitter_python":
+                args_node = node.child_by_field_name("arguments")
+                if args_node is not None:
+                    for arg in args_node.children:
+                        if arg.type == "identifier":
+                            ident = arg
+                        elif arg.type == "keyword_argument":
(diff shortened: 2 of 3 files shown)


safishamsi:
Merged into `v8` (with your authorship) — thanks, this is a genuinely useful gap to close: `affected`/blast-radius silently dropping callers that pass a function by name (`submit(fn)`, `Thread(target=fn)`, `map(fn, xs)`). The design is right — a distinct `indirect_call` INFERRED relation kept separate from `calls`, added to the affected set.

One hardening I added before merge: the guard was purely name-based (it emitted whenever an argument identifier matched any function label in the file), which produces a false edge for the idiomatic local/param-shadow case — `def via(pool, handler): pool.submit(handler)` (here `handler` is the parameter, not the module function), and for a data variable sharing a function's name (`process(config)`). I added: (1) emit only when the name resolves to a callable definition, and (2) reject when the name is shadowed by a parameter or local binding in the enclosing function. Added negative tests for param-shadow, local-shadow, and data-var alongside your positives. Verified: genuine `submit(handler)` still emits; the shadow cases emit nothing; full suite green. Ships next release.

Issue #1566, opened 2026-06-30
Extend indirect_call to remaining dispatch patterns (dict tables, assignments, other languages)


safishamsi:
### Context

#1565 plus the cross-file follow-up (commit `cf747ab`) capture a function passed by name as a **call argument**, same-file and cross-file (`submit(fn)`, `Thread(target=fn)`, `map(fn, xs)`). That covers the most common indirect-dispatch shape but not all of them. `graphify affected` (blast radius) still misses callers that reference a function through the patterns below.

All of these are detectable from the AST with the same guards already in place (resolves to a callable def, not shadowed by a param/local, single-definition cross-file), and all stay quarantined in the INFERRED `indirect_call` relation so strict `calls` queries are untouched.

### Proposed slices, ranked by value vs. false-edge risk

1. **Dispatch tables (collection literals).** A function referenced as a value inside a dict, list, tuple, or set literal:
   ```python
   ROUTES = {"create": create_user, "delete": delete_user}
   HANDLERS = [on_start, on_stop]
   ```
   Very common registry idiom. Same guards as the argument case. Low risk, highest payoff.

2. **Assignment and return references.** A function bound to a name and used later, or returned from a factory:
   ```python
   cb = handler          # cb() called later
   def make(): return handler
   ```
   Medium risk: the shadow logic has to be inverted carefully (here the local binding is the reference we want), so it must not re-introduce the false edges #1565 fixed.

3. **getattr by string literal.** `getattr(obj, "method")()` where the argument is a constant string. Literal-string case only. Niche, lower priority.

4. **Decorator-registry dispatch.** `@registry.register` style, where the decorated function becomes a callback. Framework-specific, higher false-edge risk, deferred unless requested.

5. **Other languages.** `indirect_call` is Python-only today. JS/TS is the next ecosystem for callback-passing (`arr.map(fn)`, `setTimeout(fn)`, `addEventListener("x", fn)`); the cross-file resolver already supports JS/TS direct calls, so it could reuse the same path.

### Plan

Start with slice 1 (dispatch tables), same test rigor: positive cases plus negatives for shadowing and non-callable values. Land it, then decide on 2 and 5. Slices 3 and 4 stay deferred unless a concrete request comes in.

### Non-goals

No dataflow analysis (resolve names, not values). No influence on `calls`. No change to how `affected` traverses, only what edges feed it.

safishamsi:
Slice 1 (dispatch tables) landed in `8288829` on `v8`.

[...]

PR #1567, opened 2026-06-30
feat(extract): capture dispatch-table indirect_call edges (#1566 slice 1)

sheik-hiiobd:
Closes slice 1 of #1566 (dispatch tables).

## What

`indirect_call` (#1565) covered a function passed as a **call argument**. This extends it to the registry idiom: a function referenced as a **value inside a collection literal**.

```python
ROUTES = {"create": create_user, "delete": delete_user}   # module-level dict
HOOKS  = [on_start, on_stop]                               # list / tuple / set
```

`graphify affected create_user` previously dropped the module/function that registers it; now it's in the blast radius.

## How

- Emit `indirect_call` (`context="collection"`, `INFERRED`) for callable-resolving identifiers that are **dict pair values** or **list/set/tuple elements**, at:
  - **module scope** — owner = the file node (the common top-level registry);
  - **inside a function body** — owner = the enclosing function.
- **Reuses the argument-case guards** (`cf747ab`): resolve only to a `callable_def_nids` target, and skip a name in the enclosing function's `local_bound_names` (param/local shadow). Cross-file values defer to the same single-definition resolver as the argument case.
- Strict `calls` is untouched; only `indirect_call` (already in `DEFAULT_AFFECTED_RELATIONS`) is affected.

## Scope / limits (intentional)

- **Python only**, matching the existing `indirect_call`.
- Covers all four literal kinds (dict/list/set/tuple) at module + function scope. **Class-body** attributes are deferred (they need a class owner, not the file) — a small follow-up.
- No dataflow: resolves names, not values (a module-level rebind isn't tracked — out of scope per the issue's non-goals).

## Tests

`tests/test_indirect_dispatch_collections.py` — positives (module dict+list, function-level, tuple+set, `affected`) **and** the negatives the issue asks for: param-shadow, local-shadow, and non-callable values each emit nothing. Full suite green; ruff clean.

### DIFF
--- a/graphify/extract.py
+++ b/graphify/extract.py
@@ -4180,6 +4180,60 @@ def _extract_generic(
             return None
         return _read_text(scope, source)
 
+    _PY_COLLECTION_TYPES = ("dictionary", "list", "set", "tuple")
+
+    def _collection_value_idents(coll):
+        # Function-value items of a collection literal: dict -> each pair's `value`;
+        # list/set/tuple -> each element. Identifiers only.
+        if coll.type == "dictionary":
+            for ch in coll.children:
+                if ch.type == "pair":
+                    v = ch.child_by_field_name("value")
+                    if v is not None and v.type == "identifier":
+                        yield v
+        else:
+            for ch in coll.children:
+                if ch.type == "identifier":
+                    yield ch
+
+    def _emit_collection_dispatch(coll, owner_nid: str, enclosing_locals) -> None:
+        # Dispatch tables: a function referenced as a value inside a collection literal
+        # (ROUTES = {"x": handler}, HOOKS = [on_start]). Same guards as the call-argument
+        # case -- skip a param/local shadow, resolve only to a callable def -- under the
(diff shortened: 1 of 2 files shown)


sheik-hiiobd:
Looks like you folded slice 1 into 8288829 with the shared Python+JS helper — nicer than mine (I'd duplicated the emit). Closing this in favour of that. Grabbing slice 2 (assignment/return refs) next.

safishamsi:
Closing the loop here for the record: the slice-1 dispatch-table work landed on `v8` in `8288829` (it was implemented in parallel just before this PR came in), so this is already shipped rather than dropped. Your approach here matched it — dict/list/set/tuple values at both module and function scope, on the same callable-target + param/local shadow guards, with the shadow and non-callable negatives up front.

Credit for the parallel implementation stands, and the next-release changelog covers the slice-1 work. Thanks for jumping on it — and slice 2 (#1569) is already merged with your authorship. Slices 3 (getattr-by-string) and 4 (decorator-registry) on #1566 remain if you want them.

PR #1575, opened 2026-07-01
feat(extract): capture getattr(obj, "name") indirect_call edges (#1566 slice 3)

sheik-hiiobd:
Closes slice 3 of #1566 (getattr-by-string reflective dispatch).

## What

`indirect_call` covered call arguments (#1565), dispatch tables (#1566 slice 1) and assignment/return refs (#1566 slice 2). This adds reflective dispatch by string literal:

```python
def dispatch(obj):
    fn = getattr(obj, "handler")     # string-literal attribute name
    return fn()

HANDLER = getattr(sys.modules[__name__], "handler")   # module-level reflective alias
```

`graphify affected handler` previously dropped these; now they're in the blast radius. Emitted at both function scope (owner = enclosing function) and module scope (owner = file node), context `"getattr"`, `INFERRED`.

## How

- A `getattr(obj, "name"[, default])` whose **name arg is a plain string literal** resolves to a callable def of that label and emits one `indirect_call`.
- **The scope rule is the inverse of the identifier paths.** A `getattr` string names an *attribute*, not a binding, so it is **never shadowed by a param/local** — `def via(handler): getattr(x, "handler")` still resolves to the module `handler`. Applying the identifier shadow guard here would be a false *negative*. The getattr path therefore resolves **without** the shadow guard, while the identifier paths keep it.
- To reuse the resolution logic without duplicating it, the shared resolve-and-emit core (callable-target-only, cross-file deferral, call-aware dedup) is lifted out of `_emit_indirect_ref` into `_emit_indirect_by_name`. The identifier wrapper (`_emit_indirect_ref`) keeps the type + shadow guards and delegates — **behavior-preserving**; the whole existing suite (all languages, slices 1/2) stays green.
- **Only a plain literal is resolvable.** A variable, an f-string (`f"on_{x}"`), a concatenation (`"on_" + x`), or any expression is dynamic → no edge. `obj.getattr(...)` (a method, not the builtin) and the 1-arg form are ignored.

## Scope / limits

- **Python** (where the slice was specced); JS/TS reflective access is a follow-up like the others.
- **Cross-file getattr** (`getattr(imported_mod, "fn")`) is handled by construction, not by a new path: an unresolved name defers to the same cross-file resolver as slices 1/2, which is context-agnostic and applies the global callable-target guard — so no separate cross-file test is added here.
- No dataflow: the reference is captured at the `getattr` site, not at later uses of the bound alias — per the issue's non-goals.

## Tests

`tests/test_indirect_dispatch_getattr.py` — positives (function-scope, 3-arg form, module-level alias, `affected` reachability, context/confidence) and the negatives that discriminate:
- **param-collision still emits** — a `getattr("handler")` whose name collides with a same-named parameter must *still* resolve (this would flip to a false negative under the identifier shadow guard);
- variable / f-string / concatenation name → nothing;
- non-callable name → nothing;
- method named `getattr` (not the builtin) → nothing.

Full suite green; ruff clean.

### DIFF
--- a/graphify/extract.py
+++ b/graphify/extract.py
@@ -4328,22 +4328,17 @@ def _extract_generic(
     # receiver_type so the cross-file pass resolves `var.method` by type (#ruby).
     ruby_var_types: dict[str, dict[str, str | None]] = {}
 
-    def _emit_indirect_ref(ident, scope_nid: str, enclosing_locals, context: str) -> None:
-        """A function referenced BY NAME — passed as a call argument, or listed as a
-        value in a dispatch table — is an indirect dependency of ``scope_nid``. Emit
-        it as a distinct INFERRED ``indirect_call`` (kept out of the precise ``calls``
-        relation) only when the name resolves to a real callable and is NOT shadowed
-        by a parameter / local binding. A callback defined in another file is deferred
-        to the cross-file resolver via an ``indirect`` raw_call carrying its context.
-        Language-agnostic; shared by the call-argument and dispatch-table capture
-        paths for Python and JS/TS (#1565, #1566).
+    def _emit_indirect_by_name(ident_name: str, loc_node, scope_nid: str,
+                               context: str) -> None:
+        """Resolve a name that is referenced AS A VALUE to a real callable def and emit
+        one INFERRED ``indirect_call`` edge — deferring an unknown / foreign name to the
+        cross-file resolver, which applies the single-definition god-node guard and the
+        GLOBAL callable-target check. The name is already extracted; scope filtering is
+        the CALLER's job: an identifier reference must reject param/local shadows (a bare
+        name IS a binding — see ``_emit_indirect_ref``), whereas a ``getattr(obj, "x")``
+        string names an ATTRIBUTE and is never shadowed by a local, so that path passes
+        the name straight through. ``loc_node`` supplies the source line.
         """
-        if ident is None or ident.type not in ("identifier", "shorthand_property_identifier"):
-            return
-        ident_name = _read_text(ident, source)
-        # shadowing: a param / local binding names a local value, not the module fn
-        if ident_name in enclosing_locals or ident_name in ("self", "cls"):
-            return
         ref_nid = label_to_nid.get(ident_name)
         # Defer to the cross-file resolver when the name is not defined in this file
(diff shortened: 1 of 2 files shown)


safishamsi (approved):
Reviewed and tested locally — approving. Clean continuation of the arc.

**What I checked**
- The refactor is behavior-preserving: lifting the resolve-and-emit core into `_emit_indirect_by_name` while `_emit_indirect_ref` keeps the type + shadow guards and delegates. The whole existing suite (slices 1/2/5, all languages) stays green, which is the proof.
- The scope-rule inversion is correct: a `getattr(obj, "x")` string is an attribute name, never shadowed by a param/local, so bypassing the shadow guard is right — applying it would be a false negative. Your `test_..._not_shadowed_by_param` pins exactly that.
- `_getattr_ref_name` is appropriately strict: builtin `getattr` only (not `obj.getattr`), >=2 positional args, plain string literal only.

**Adversarial cases I ran beyond the PR's tests, all correct:**
- variable name, f-string, `"on_"+x` concatenation, empty string, 1-arg form, `obj.getattr(...)` method -> no edge
- inline 2-arg call `getattr(o,"h")()` -> edge
- 3-arg with a callable fallback `getattr(obj,"handler",fb)` -> both the string (`handler`, context getattr) AND the fallback (`fb`, context argument) resolve, each via the right path
- keyword form `getattr(obj, name="handler")` -> no edge (unhandled, fine)

**One caveat worth noting** (not blocking): resolving the string to a callable by bare label means `getattr(obj, "handler")` attributes to a module function `handler` even when the attribute is really a method on `obj`'s class. That's a genuine precision tradeoff, but it's consistent with how the whole `indirect_call` feature name-matches, it's INFERRED and confined to `affected` (recall-biased — over-inclusion is the accepted default there), and the single-definition god-node guard still applies cross-file. Fine as-is.

Merging into `v8`.

safishamsi:
Merged into `v8` as `8fdbf50` (cherry-picked with your authorship). Verified clean on current `v8`: full suite 2756 passed, plus a combined smoke where all four indirect contexts (collection, assignment, argument, getattr) coexist in one file — the getattr string `"handler"` resolves even when `handler` is a shadowed param, while the identifier arg path correctly suppresses it. All the adversarial negatives (variable/f-string/concat/1-arg/method/empty-string) hold. On the next-release changelog crediting you. Nice work on the scope-inversion reasoning.

PR #1576, opened 2026-07-01
feat(cluster): deterministic hub community labels (readable without an LLM)

sheik-hiiobd:
## What

Community labels fall back to `Community N` whenever no LLM backend is configured, which makes the report and its Suggested Questions unreadable:

> why does `log_action` connect Community 70 to Community 129, 4, 133, 5…

Add `label_communities_by_hub`: name each community after its **highest-degree member** — the structural hub — so the report reads `log_action` / `auth` with zero LLM cost.

## How

- New `label_communities_by_hub(G, communities)` in `cluster.py`: for each community, pick the highest-degree member (ties broken by node id for run-to-run stability), use its label as the community name (`()` stripped). A community whose members are all absent from `G` keeps the `Community {cid}` placeholder.
- Wired as the **default base** at both label-building sites — the `label`/standalone path in `__main__` and the watch/detect rebuild in `watch.py`.
- A configured LLM naming pass still runs and **overrides** the hub labels with richer names. Its no-backend placeholder fallback is **guarded** (`v != f"Community {cid}"`) so it can't clobber the deterministic labels.

## Why this direction

The hub is a good zero-cost proxy for "what is this cluster about": the most-connected node is usually the subsystem's entry point or shared utility. It's deterministic (no model, no tokens, no run-to-run drift), so a plain `graphify` run is readable out of the box, and `--no-label` still yields bare `Community N` for anyone who wants it.

## Scope

- Default (no backend) and watch/detect rebuild paths. LLM naming is unchanged where configured.
- `--no-label` semantics preserved (explicit opt-out → bare placeholders).

## Tests

`tests/test_community_hub_labels.py` — highest-degree hub wins; deterministic id tie-break (order-independent); absent-members fall back to `Community N`; node without a `label` attr uses its id; multiple communities each get their own hub; a real community is never left as a placeholder. Full suite green; ruff clean.

### DIFF
--- a/graphify/__main__.py
+++ b/graphify/__main__.py
@@ -3559,17 +3559,20 @@ def main() -> None:
         else:
             # No labels file yet (or `graphify label` forced a refresh). When run
             # standalone there is no orchestrating agent to do skill.md Step 5, so
-            # auto-name communities with the configured backend rather than leave
-            # "Community N" (#1097). Degrades to placeholders if no backend/on error.
+            # auto-name communities rather than leave "Community N" (#1097).
+            from graphify.cluster import label_communities_by_hub
             from graphify.llm import generate_community_labels
             print("Labeling communities...")
-            # The final labels (LLM or placeholder fallback) are persisted to
-            # .graphify_labels.json by the unconditional write below.
+            # Deterministic, LLM-free base labels: name each community after its
+            # highest-degree hub, so the report is readable even with no backend
+            # (previously bare "Community N"). A configured LLM backend overrides these
+            # with richer names below; its no-backend placeholder fallback does NOT.
+            hub_labels = label_communities_by_hub(G, communities)
             label_communities_input = communities
-            labels = {}
+            labels = dict(hub_labels)
             if missing_only:
                 labels = {
-                    cid: existing_labels.get(cid, f"Community {cid}")
+                    cid: existing_labels.get(cid, hub_labels[cid])
                     for cid in communities
                 }
                 label_communities_input = {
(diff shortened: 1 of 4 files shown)


safishamsi (approved):
Reviewed and tested locally — approving. Genuinely useful: a plain `graphify` run with no backend now reads `auth` / `log_action` instead of `Community 70`.

**What I checked**
- `label_communities_by_hub` is deterministic: highest degree wins, ties break by node id ascending. Confirmed order-independence.
- Integration is sound at both sites. In `__main__`, `labels = dict(hub_labels)` as the base and the LLM override guarded by `v != f"Community {cid}"` means a no-backend LLM fallback can't clobber the hub labels; a real LLM name still wins. In `watch.py`, only communities missing a saved label get hub names — saved labels are untouched.
- No relabel regression: "missing" is decided by presence in `.graphify_labels.json`, not by the `Community N` pattern, and `graphify label` (force) still overrides hub labels with LLM names. So persisting a readable hub name instead of `Community N` is equivalent behavior, just readable. Full suite green.

**Two small observations** (non-blocking):
- Two communities can end up with the same hub label when their hubs share a name (e.g. two `handler` in different files) — I confirmed `{0: 'handler', 1: 'handler'}`. Not incorrect, just mildly ambiguous in the report; the LLM path disambiguates when configured.
- The hub can occasionally be a generic shared utility (a logger, `__init__`), giving a slightly off name — inherent to the heuristic and still far better than `Community N`.

Nice, self-contained, well-tested. Merging into `v8`.

safishamsi:
Merged into `v8` as `1aab291` (cherry-picked with your authorship). Verified: full suite green, deterministic ordering confirmed, and the LLM-override guard + file-presence relabel semantics mean no regression — a readable hub name simply replaces the bare `Community N` default. Real quality-of-life win for no-backend runs. On the next-release changelog crediting you.

PR #1577, opened 2026-07-01
fix(elixir): expand multi-alias brace form into imports edges

Synvoya:
## Summary

Elixir's multi-alias brace form emits **no** imports edges:

```elixir
alias Foo.{Bar, Baz}   # -> should import Foo.Bar and Foo.Baz; emitted nothing
```

Single aliases, `import`, `require`, `use` all work — only the brace expansion was dropped.

## Root cause

tree-sitter-elixir represents `alias Foo.{Bar, Baz}` as a `dot` node holding the base `alias` (`Foo`) plus a trailing `tuple` of member aliases (`Bar`, `Baz`):

```
arguments
  dot
    alias "Foo"
    tuple ( alias "Bar", alias "Baz" )
```

The import handler used `_get_alias_text`, which only matches a bare `alias` child — so the `dot`/`tuple` form returned `None` and no edge was emitted.

## Fix

Add `_get_alias_modules`, which expands the brace form to `["Foo.Bar", "Foo.Baz"]` and leaves the single form (`alias Foo.Bar` -> `["Foo.Bar"]`) unchanged. The import handler now emits one edge per expanded module.

## Verification

- `alias Foo.{Bar, Baz}` -> 2 imports edges (Foo.Bar, Foo.Baz); single/import/require/use unchanged.
- Added a fixture line + regression test.
- `pytest tests/test_languages.py` -> **300 passed, 13 skipped** (8 elixir, incl. the new test).

### DIFF
--- a/graphify/extractors/elixir.py
+++ b/graphify/extractors/elixir.py
@@ -58,6 +58,35 @@ def extract_elixir(path: Path) -> dict:
                 return source[child.start_byte:child.end_byte].decode("utf-8", errors="replace")
         return None
 
+    def _get_alias_modules(node) -> list[str]:
+        """Every module named by an alias/import/require/use argument.
+
+        Handles the single form (``alias Foo.Bar`` -> ``["Foo.Bar"]``) and the
+        multi-alias brace form (``alias Foo.{Bar, Baz}`` ->
+        ``["Foo.Bar", "Foo.Baz"]``), which the grammar represents as a ``dot``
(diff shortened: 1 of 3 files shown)


safishamsi:
Merged into `v8` as `f2ea6a6` (your authorship). Verified: single `alias`/`import` still work (preserved via the dot-without-tuple fallback), and `alias App.Schemas.{User, Post, Comment}` expands to one edge per member. Full suite 2774. Ships next release — thanks!

PR #1578, opened 2026-07-01
fix(fortran): emit calls edges for function invocations

Synvoya:
## Summary

Fortran function calls (`y = f(x)`) emit **no** `calls` edge. Only `call sub(...)` (subroutine) calls were captured.

## Root cause

`walk_calls` handled `subroutine_call` but not `call_expression`, which is how tree-sitter-fortran represents a function invocation:

```
assignment_statement
  identifier "x"
  call_expression        <-- unhandled
    identifier "compute"
```

So every function-to-function call was dropped.

## Fix

Handle `call_expression`. Fortran uses the **same** `name(...)` syntax for array indexing (`arr(i)`), so — unlike `subroutine_call` — the callee is resolved against procedures **defined in the file** (`target_nid in seen_ids`) before emitting. Array accesses whose name isn't a defined procedure produce no edge, so they can't fabricate spurious `calls`.

## Verification

- `f = g()` -> calls(f, g); `call sub()` still works; `arr(3)` array access -> **no** edge.
- Added a function + caller to the fixture and a regression test.
- `pytest tests/test_languages.py` -> **300 passed, 13 skipped** (13 fortran).
- `ruff check` -> clean.

### DIFF
--- a/graphify/extract.py
+++ b/graphify/extract.py
@@ -7503,6 +7503,19 @@ def extract_fortran(path: Path) -> dict:
                 target_nid = _make_id(stem, callee)
                 add_edge(scope_nid, target_nid, "calls", node.start_point[0] + 1,
                          confidence="EXTRACTED", context="call")
+        # x = compute(args) — function invocations are `call_expression`, which
+        # shares Fortran's `name(...)` syntax with array indexing. Only emit a
+        # call edge when the callee resolves to a procedure defined in this file
+        # (an array variable produces no matching node), so array accesses can't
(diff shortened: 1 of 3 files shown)


safishamsi:
Merged into `v8` as `b8f41c7` (your authorship). The array-indexing guard is the right call and it holds — adversarial smoke confirmed `z = dbl(a)` emits a call while `z = arr(3)` (same `name(...)` syntax) does not. Full suite 2774. Ships next release.

PR #1579, opened 2026-07-01
fix(rust): emit references edges for enum variant field types

Synvoya:
## Summary

Rust enum variant payload types emit **no** `references` edges:

```rust
enum Event {
    Click(Logger),            // -> should reference Logger; emitted nothing
    Resize { size: Dim },     // -> should reference Dim; emitted nothing
}
```

Struct field type references work fine — only enums were affected.

## Root cause

`struct_item` and `trait_item` have type-reference handlers, but `enum_item` has none. Variant field types nest one level deeper than struct fields:

```
enum_item
  enum_variant_list
    enum_variant
      ordered_field_declaration_list   (tuple variant)  -> type_identifier
      field_declaration_list           (struct variant) -> field_declaration -> type
```

That path was never traversed.

## Fix

Add an `enum_item` branch that walks both variant forms and emits a `references` edge from the enum to each field type, reusing the same `_rust_collect_type_refs` path as structs.

## Verification

- Tuple + struct variant refs now emit; struct field refs unchanged.
- Added an enum to the fixture and a regression test.
- `pytest tests/test_multilang.py tests/test_languages.py` -> **337 passed, 22 skipped**.
- `ruff check` -> clean.

### DIFF
--- a/graphify/extract.py
+++ b/graphify/extract.py
@@ -8143,6 +8143,45 @@ def extract_rust(path: Path) -> dict:
                                 if tgt != item_nid:
                                     add_edge(item_nid, tgt, "references",
                                              field.start_point[0] + 1, context=ctx)
+                if t == "enum_item":
+                    # Variant payload types nest under enum_variant_list ->
+                    # enum_variant -> ordered_field_declaration_list (tuple variant,
+                    # `Click(Logger)`) | field_declaration_list (struct variant,
+                    # `Resize { size: Dim }`). Neither was traversed, so every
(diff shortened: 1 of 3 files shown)


safishamsi:
Merged into `v8` as `674184d` (your authorship). Confirmed both tuple (`Click(Logger)`) and struct (`Resize { size: Dim }`) variants emit references, and a unit variant (`Quit`) adds nothing spurious. Mirrors the struct handler cleanly. Full suite 2774. Ships next release.

PR #1580, opened 2026-07-01
fix(julia): emit imports for qualified, relative & scoped-selected forms

Synvoya:
## Summary

Julia imports beyond the bare form emit **no** edge:

```julia
using Base.Threads              # qualified  -> dropped
using ..Sibling                 # relative   -> dropped
import Base.Threads: nthreads   # scoped pkg -> pointed at `nthreads`, not the module
using Foo                       # simple     -> OK
```

## Root cause

The `using_statement`/`import_statement` handler only matched bare `identifier` children. tree-sitter-julia represents qualified paths as `scoped_identifier`, relative paths as `import_path`, and the package inside a `selected_import` may itself be a `scoped_identifier` — none of which were handled.

## Fix

Resolve the module name from `identifier` / `scoped_identifier` / `import_path` in all three positions (top-level child and selected-import package).

## Verification

- All 3 dropped forms now emit; `using Foo` / `import Bar: baz` unchanged.
- Added fixture lines + a regression test.
- `pytest tests/test_languages.py tests/test_multilang.py` passes; `ruff` clean.

### DIFF
--- a/graphify/extract.py
+++ b/graphify/extract.py
@@ -7281,19 +7281,39 @@ def extract_julia(path: Path) -> dict:
         # Using / Import
         if t in ("using_statement", "import_statement"):
             line = node.start_point[0] + 1
+
+            def _julia_mod_name(n):
+                # identifier (`Foo`), scoped_identifier (`Base.Threads`), or
+                # import_path (relative `..Sibling`) -> the module name. Only bare
+                # identifiers were handled, so qualified/relative imports — and the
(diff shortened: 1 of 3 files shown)


safishamsi:
Merged into `v8` as `984a6a8` (your authorship). Verified bare/qualified/relative/scoped-selected forms all emit and the single/selected forms are unchanged. Full suite 2777. Thanks!

PR #1583, opened 2026-07-01
fix(systemverilog): emit field references for qualified class properties

Synvoya:
## Summary

SystemVerilog class properties declared with a leading qualifier (`rand`, `local`, `protected`, …) have their type references silently dropped from the graph. The class-body field extractor in `_augment_systemverilog_semantics` matched only unqualified `<type> <name>;` declarations.

## Symptom

```systemverilog
class Packet;
  Payload   m_unqualified;   // edge emitted: Packet -> Payload
  rand Config    m_qualified; // DROPPED
  local Header   m_local;     // DROPPED
  protected Meta m_meta;      // DROPPED
endclass
```

Only `Packet -> Payload` (`references`, `field`) is emitted; the `rand`/`local`/`protected` field type references never appear.

## Root cause

In `graphify/extract.py`, the class-body field loop in `_augment_systemverilog_semantics`:

```python
for field in re.finditer(r"^\s*([A-Za-z_]\w*(?:\s*#\s*\([^;]+?\))?)\s+\w+\s*;", body_without_functions, re.MULTILINE):
```

The pattern is `^\s* <type> \s+ <name> ;`. `^\s*` consumes only leading whitespace, so a qualified declaration like `rand Config m_cfg;` presents three tokens (`rand`, `Config`, `m_cfg`) to a two-token shape: the regex binds `rand` as the type and `Config` as the name, then the trailing `\s*;` fails (there is ` m_cfg;` left), so the whole line fails to match and the field is skipped.

## Fix

Consume optional leading class-property qualifiers before the type token:

```python
r"^\s*(?:(?:rand|randc|local|protected|static|const|automatic|var)\s+)*([A-Za-z_]\w*(?:\s*#\s*\([^;]+?\))?)\s+\w+\s*;"
```

The qualifier group is `*` (zero-or-more), so unqualified fields keep their exact current behavior and the type/name capture is unchanged. The listed qualifiers are reserved SV keywords, so there is no risk of consuming a real type name.

## Verification

- New regression test `test_systemverilog_qualified_field_references` — fails before the fix, passes after — covering `rand`- and `protected`-qualified fields.
- `tests/test_languages.py` + `tests/test_multilang.py`: **340 passed / 22 skipped**.
- `ruff check graphify/extract.py tests/test_languages.py` clean.

Before → after on the example above:

```
before: references[field] = {(Packet, Payload)}
after:  references[field] = {(Packet, Payload), (Packet, Config),
                             (Packet, Header), (Packet, Meta)}
```

### DIFF
--- a/graphify/extract.py
+++ b/graphify/extract.py
@@ -6637,7 +6637,11 @@ def _augment_systemverilog_semantics(
             body,
             flags=re.DOTALL,
         )
-        for field in re.finditer(r"^\s*([A-Za-z_]\w*(?:\s*#\s*\([^;]+?\))?)\s+\w+\s*;", body_without_functions, re.MULTILINE):
+        # Optional leading class-property qualifiers (rand/local/protected/etc.)
+        # must be consumed: otherwise a qualified field like `rand Config x;`
+        # (three tokens) fails the `<type> <name>;` shape and its type reference
+        # is silently dropped.
+        for field in re.finditer(r"^\s*(?:(?:rand|randc|local|protected|static|const|automatic|var)\s+)*([A-Za-z_]\w*(?:\s*#\s*\([^;]+?\))?)\s+\w+\s*;", body_without_functions, re.MULTILINE):
             # Count to the start of the type token (group 1), not the match
             # start: `^\s*` consumes the leading newline(s), so field.start()
             # would resolve to the class's line instead of the field's.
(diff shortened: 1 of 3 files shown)


safishamsi:
Merged into `v8` as `297075c` (your authorship). The offset stays correct — it already keyed on group(1)/start(1), so the non-capturing qualifier prefix leaves the type token intact. Verified rand/protected/local-static qualified fields resolve and unqualified fields still work. Full suite 2777.

PR #1587, opened 2026-07-01
fix(scala): emit field type references for var declarations

Synvoya:
## Summary

Scala mutable fields (`var x: T`) have their type reference silently dropped from the graph — only `val` fields are handled.

## Root cause

In `graphify/extract.py`, the Scala field handler guards on `t == "val_definition"`. A `var` field parses as `var_definition`, which is never matched, so its type reference is skipped. `val_definition` and `var_definition` are structurally identical — both expose a `type` field consumed by `child_by_field_name("type")`.

## Fix

Widen the guard to `t in ("val_definition", "var_definition")`. Generics/tuples flow through `_scala_collect_type_refs` unchanged; the `val` path is untouched.

## Verification

- Added `var fallback: BaseClient` to the Scala fixture + `test_scala_var_definition_field_context` (fails before the fix, passes after).
- All 11 Scala tests pass; `ruff` clean.

Before → after: `HttpClient → BaseClient` (`references`, `field`) is now emitted for the `var` field.

### DIFF
--- a/graphify/extract.py
+++ b/graphify/extract.py
@@ -3868,7 +3868,7 @@ def _extract_generic(
             return
 
         if (config.ts_module == "tree_sitter_scala"
-                and t == "val_definition"
+                and t in ("val_definition", "var_definition")
                 and parent_class_nid):
             type_node = node.child_by_field_name("type")
             if type_node is not None:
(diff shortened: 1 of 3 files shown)


safishamsi:
Merged as `67b4525` (your authorship). `var` now mirrors `val` field-type refs. Full suite 2788.

PR #1588, opened 2026-07-01
fix(powershell): emit inherits/implements edges for class base types

Synvoya:
## Summary

PowerShell class inheritance is dropped: `class Dog : Animal` never emits the `Dog → Animal` inherits edge, so derived classes become isolated nodes.

## Root cause

In `graphify/extract.py`, the `class_statement` handler reads only the first `simple_name` child (the class name) and never inspects the base type(s) after the `:` token. Multi-base declarations (`class Dog : Animal, IBark`) drop the interfaces too.

## Fix

Walk the `class_statement` children; once the `:` token is seen, treat each following `simple_name` as a base type. PowerShell has no syntactic base-vs-interface split, so — matching the existing C# convention — the first base is emitted as `inherits` and the rest as `implements` (resolved via `ensure_named_node`).

## Verification

- Added a `Shape` base + `Circle : Shape` pair to the fixture and `test_powershell_class_base_type_emits_inherits_edge` (fails before the fix, passes after).

Before → after: `Circle → Shape` (`inherits`) is now emitted.

### DIFF
--- a/graphify/extract.py
+++ b/graphify/extract.py
@@ -8495,6 +8495,22 @@ def extract_powershell(path: Path) -> dict:
                 class_nid = _make_id(stem, class_name)
                 add_node(class_nid, class_name, line)
                 add_edge(file_nid, class_nid, "contains", line)
+                # Base type(s) after ':'. PowerShell has no syntactic base vs
+                # interface split, so (matching the C# convention) treat the
+                # first base as the superclass (inherits) and the rest as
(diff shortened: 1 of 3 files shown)


safishamsi:
Merged as `a129ff2` (your authorship). Verified multi-base first=inherits / rest=implements and single-base=inherits. Full suite 2788.

PR #1589, opened 2026-07-01
fix(objc): emit implements edge for protocol-to-protocol adoption

Synvoya:
## Summary

Objective-C protocol-to-protocol adoption is dropped: `@protocol Derived <Base>` never emits the `Derived → Base` edge — even though the same extractor emits `implements` for class adoption `@interface Foo <Proto>`.

## Root cause

In `graphify/extract.py`, the `protocol_declaration` handler walks children only for methods and ignores the `protocol_reference_list` node holding the adopted protocols. That node type is distinct from the `parameterized_arguments` node used by `@interface` adoption — so the interface path emits `implements` but the protocol path drops it.

## Fix

In `protocol_declaration`, iterate the `protocol_reference_list` children and emit an `implements` edge for each adopted protocol (a direct `identifier` child), mirroring the `@interface` adoption behaviour.

## Verification

- Added `@protocol Base` + `@protocol Derived <Base>` to the fixture and `test_objc_protocol_adopts_protocol` (fails before the fix, passes after).
- AST-confirmed the adopted-protocol name is a direct `identifier` child of `protocol_reference_list`.

Before → after: `<Derived> → <Base>` (`implements`) is now emitted.

### DIFF
--- a/graphify/extract.py
+++ b/graphify/extract.py
@@ -11609,6 +11609,18 @@ def extract_objc(path: Path) -> dict:
                 proto_nid = _make_id(stem, name)
                 add_node(proto_nid, f"<{name}>", line)
                 add_edge(file_nid, proto_nid, "contains", line)
+                # Adopted protocols: `@protocol Derived <Base, Other>`. These
+                # nest under a protocol_reference_list node (distinct from the
+                # parameterized_arguments node used by @interface adoption), so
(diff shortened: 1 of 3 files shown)


safishamsi:
Merged as `cd3a376` (your authorship). Verified protocol-to-protocol adoption emits implements and a lonely protocol emits nothing. Full suite 2788.

PR #1590, opened 2026-07-01
fix(php): emit type references for promoted constructor properties

Synvoya:
## Summary

PHP 8 constructor property promotion drops type references. In `class Service { public function __construct(private Repo $repo) {} }`, the promoted `Repo` — which is both a parameter type and a class field — is entirely absent from the graph, while a non-promoted param emits `parameter_type` normally.

## Root cause

In `graphify/extract.py`, the PHP constructor param loop filters `if p.type != "simple_parameter": continue`. A promoted parameter parses as `property_promotion_parameter`, so it is skipped — dropping both the `parameter_type` edge and the class `field` edge.

## Fix

Accept `property_promotion_parameter` alongside `simple_parameter` (its type sits in the same direct named-child shape, so existing extraction emits `parameter_type` unchanged). Because a promoted param also declares a real class field, additionally emit a `field`-context reference on the class — mirroring the `property_declaration` handler, and preserving `generic_arg` role handling for promoted generic collections.

## Verification

- Added a promoted-property constructor to the fixture and `test_php_constructor_property_promotion_contexts` (fails before the fix, passes after).

Before → after: for `private Repo $repo`, both `__construct → Repo` (`parameter_type`) and `Service → Repo` (`field`) are now emitted.

### DIFF
--- a/graphify/extract.py
+++ b/graphify/extract.py
@@ -4069,8 +4069,14 @@ def _extract_generic(
                         break
                 if params_container is not None:
                     for p in params_container.children:
-                        if p.type != "simple_parameter":
+                        # PHP 8 constructor property promotion (`__construct(private
+                        # Repo $repo)`) parses the promoted param as
+                        # property_promotion_parameter, not simple_parameter. Its
+                        # type sits in the same direct named child shape, so accept
+                        # both here; a promoted param is additionally a class field.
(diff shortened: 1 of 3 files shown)


safishamsi:
Merged as `51f805e` (your authorship). Promoted ctor property emits the type reference as a class field. Full suite 2788.

PR #1591, opened 2026-07-01
fix(csharp): emit type references for properties

Synvoya:
## Summary

C# auto-properties drop their type references. `public Widget Main { get; set; }` and `public List<Widget> Items { get; set; }` emit no `references(field)` edges — so a class whose state is declared as properties (idiomatic modern C#) has no field edges at all.

## Root cause

In `graphify/extract.py`, the class-body walker has a C# `field_declaration` handler but no `property_declaration` handler (the existing property_declaration branches are guarded on the PHP/Kotlin/Swift `ts_module`, so none fire for `tree_sitter_c_sharp`).

## Fix

Add a C# `property_declaration` branch mirroring the field handler. A property exposes its type on the node directly (no `variable_declaration` wrapper), so read `child_by_field_name("type")` and run `_csharp_collect_type_refs` (like the Java/PHP/Kotlin siblings) — yielding both the container field ref and the generic_arg ref, with the same `ref_token`/`qualified`/`ref_qualifier` metadata as the field path.

## Verification

- Added `Owner`/`Workers` properties to the fixture and `test_csharp_property_type_references_have_field_context` (fails before the fix, passes after).

Before → after: `DataProcessor → Processor` (`field`), `DataProcessor → List` (`field`), and `DataProcessor → Processor` (`generic_arg`) are now emitted.

Note: fully-qualified generic names (`System.Collections.Generic.List<T>`) still drop the type argument — a pre-existing limitation of `_csharp_collect_type_refs`'s `qualified_name` branch, out of scope for this change.

### DIFF
--- a/graphify/extract.py
+++ b/graphify/extract.py
@@ -3786,6 +3786,35 @@ def _extract_generic(
                          "references", line, context="field", metadata=metadata)
             return
 
+        if (config.ts_module == "tree_sitter_c_sharp"
+                and t == "property_declaration"
+                and parent_class_nid):
+            # C# auto-properties (`public Widget Main { get; set; }`) are the
+            # idiomatic way to declare state, yet only field_declaration was
+            # handled — so property types produced no references edge. Unlike a
+            # field, a property exposes its type on the node directly (no
+            # variable_declaration wrapper), so read it straight off the `type`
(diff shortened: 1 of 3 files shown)


safishamsi:
Merged as `bb5e519` (your authorship). Auto-property type + generic-arg refs, mirroring field_declaration. Full suite 2788.

PR #1592, opened 2026-07-01
fix(cpp): emit generic_arg references for base-class template arguments

Synvoya:
## Summary

C++ base-class template arguments are dropped. `class Car : public Base<Dep>` emits `Car → Base` (inherits) but never `Car → Dep` — even though the tool emits these for fields, method params/returns, and for Java base classes.

## Root cause

In `graphify/extract.py`, the C++ inheritance handler's `template_type` branch reads only the base's `name` (`Base`) and never descends into its `template_argument_list`.

## Fix

After emitting the `inherits` edge, collect the base's template arguments via `_cpp_collect_type_refs` (generic=True) and emit a `generic_arg` reference for each — mirroring Java's `_emit_java_parent_type`. `_cpp_collect_type_refs` already handles nested/qualified args, so `Base<std::vector<Dep>>` is covered.

## Verification

- Added a `Connection<T>` base + `PooledClient : Connection<HttpClient>` to the fixture and `test_cpp_generic_parents_include_type_argument_references` (fails before, passes after).

Before → after: `PooledClient → HttpClient` (`references`, `generic_arg`) now emitted; `inherits` unchanged.

### DIFF
--- a/graphify/extract.py
+++ b/graphify/extract.py
@@ -3668,6 +3668,7 @@ def _extract_generic(
                         continue
                     for sub in child.children:
                         base = ""
+                        template_args_node = None
                         if sub.type == "type_identifier":
                             base = _read_text(sub, source)
                         elif sub.type == "qualified_identifier":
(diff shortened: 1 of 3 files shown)


safishamsi:
Merged as `21bcb43` (your authorship). Verified `Base<Dep>` emits inherits+generic_arg, and a plain non-template base stays inherits-only (no spurious generic). Full suite 2788.

PR #1593, opened 2026-07-01
fix(swift): emit references for enum associated-value types

Synvoya:
## Summary

Swift enum associated-value types are dropped. `case started(Session)` emits the `case_of` edge but never `Event → Session` — the associated payload types are invisible to the graph.

## Root cause

In `graphify/extract.py`, the Swift `enum_entry` handler iterates the entry's children only for the case name (`simple_identifier`) and never descends into the sibling `enum_type_parameters` node, which holds the associated-value types (`user_type → type_identifier`).

## Fix

After emitting `case_of`, descend into each `enum_type_parameters` child, collect its type refs via `_swift_collect_type_refs`, and emit a `references` edge from the enum to each type (context `type`, or `generic_arg` for generic roles). Cases with no associated value emit no reference.

## Verification

- Added `case failed(Config)` to the fixture's `NetworkError` enum and `test_swift_enum_associated_value_type_emits_references` (fails before, passes after).

Before → after: `NetworkError → Config` (`references`, `type`) now emitted.

### DIFF
--- a/graphify/extract.py
+++ b/graphify/extract.py
@@ -2598,16 +2598,37 @@ def _csharp_extra_walk(node, source: bytes, file_nid: str, stem: str, str_path: 
 
 def _swift_extra_walk(node, source: bytes, file_nid: str, stem: str, str_path: str,
                       nodes: list, edges: list, seen_ids: set, function_bodies: list,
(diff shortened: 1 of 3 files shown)


safishamsi:
Merged as `ad70152` (your authorship). Associated-value types emit references; call-site signature update looks correct. Full suite 2788.

Issue #897, opened 2026-05-16
graphify query: seed scoring returns low-relevance nodes, causing multi-call token cascade


safishamsi:
## Background

PR #891 fixed the "upfront big-file read" tax — the install was pointing assistants at `GRAPH_REPORT.md` before every question, costing 12-25K tokens upfront. That's resolved.

@diwanmail raised a separate but related problem in a comment on #891 that deserves its own tracking:

> The problem I see is that the query CLI returns a default limit of 2000 tokens, and many times it does not include all the relevant nodes in the response. The response gets cut due to the 2000-token budget, and many of the returned nodes are not meaningful. Because of this, Claude Code ends up doing additional reasoning, making multiple query and explain calls, which further fills the context (another ~2K tokens per query command).

## The actual failure mode

`graphify query` returns a 2K-budget BFS subgraph, but when seed scoring selects low-relevance nodes the assistant doesn't get what it needs on the first call. Its only recourse is to fire follow-up `query` and `explain` calls — each costing another ~2K tokens. Three or four retries on a bad seed = 6-8K of low-signal context, which produces the same surface symptom as the old report-first bug: graphify installed is slower and more expensive than vanilla grep.

This is distinct from #891's root cause — the assistant never reads `GRAPH_REPORT.md` in this failure mode. The cost comes from retrieval quality, not from a fixed upfront read.

## What needs fixing

- **Seed scoring relevance** — the ranking that picks which nodes to return for a natural-language query needs auditing. Likely suspects: term-match weights, centrality/PageRank contribution, type-filter bias (files vs. symbols vs. rationale nodes).
- **Truncation signal** — when the BFS subgraph is cut by the 2K budget, the caller has no way to know. A `truncated: true` flag (or similar) would let the assistant extend selectively rather than re-querying from scratch.
- **Relevance benchmark** — the existing benchmark measures token cost, not result quality. A precision@k test against hand-labelled "where is X defined" / "what calls Y" queries would catch scoring regressions before they ship.

## Repro request

@diwanmail — if you're able to share any of the following it would help a lot in diagnosing the scoring issue:

- The specific query you were running
- Rough corpus shape (language mix, file count)
- Which nodes `graphify query` actually returned (or a sample)
- Whether the problem is consistent across queries or only on certain types (e.g. symbol lookups vs. architectural questions)

Even a rough description of the corpus and query type is enough to narrow down where scoring is going wrong.

diwanmail:
Sure, I will try to provide details.

[...]

PR #1596, opened 2026-07-01
fix: guarantee per-term BFS seed diversity in query (fixes #1445)

### LINKED ISSUE #1445: BFS path drift on vague Chinese queries: 40-node corpus, 12 unrelated nodes returned for "rate_limit_check 怎么用？"
## Summary

`graphify query` returns semantically unrelated nodes when a Chinese-language query is vague (doesn't name a specific symbol). This was reproduced on a 716-node real-world corpus. The same query *does* work in English when phrased around a named symbol.

## Environment

- `graphifyy` v0.8.46 (CLI)
- Backend: `claude` (Anthropic API)
- Corpus: 40 files (37 .py + 3 .md) of a real Python utility library
- Resulting graph: 716 nodes / 1159 edges / 94 communities, 3 hyperedges

## Reproduction

The query was "**rate_limit_check 怎么用？**" (How do I use rate_limit_check?).

The expected behavior: return nodes related to `rate_limit_check` (the function in `src/moving_average.py`).

The actual behavior: the BFS started from these two unrelated start nodes:
```
spec Section 4: 纯函数，不修改 text, 不依赖外部状态, 多次调用结果一致.
JWT 认证模块 — 修复 JWT Token 永久有效安全漏洞 (kb: ks128001).  正确做法：使用 exp claim 设置过期时间，默认 1
```

…and then returned 12 nodes, none of which were about `rate_limit_check`:

```
NODE test_caesar_cipher.py [src=tests/test_caesar_cipher.py]
NODE generate_token() [src=src/jwt_auth.py]
NODE verify_token() [src=src/jwt_auth.py]
NODE TestPureFunction [src=tests/test_caesar_cipher.py]
NODE jwt_auth.py [src=src/jwt_auth.py]
NODE _b64url_encode() [src=src/jwt_auth.py]
NODE .test_repeated_calls_return_same_result() [src=tests/test_caesar_cipher.py]
NODE .test_does_not_depend_on_call_count() [src=tests/test_caesar_cipher.py]
NODE .test_does_not_mutate_input() [src=tests/test_caesar_cipher.py]
NODE _b64url_decode() [src=src/jwt_auth.py]
NODE JWT 认证模块 — 修复 JWT Token 永久有效安全漏洞 ...
NODE spec Section 4: 纯函数，不修改 text...
```

For comparison, the *English* equivalent query `How is rate_limit_check implemented?` returned **0 nodes** ("No matching nodes found"). And the *specific* query `moving_average 实现` (Chinese) returned 34 correct nodes including the function and all its test methods. So the problem is specific to vague Chinese queries that don't name a target symbol.

## What works (for contrast)

| Query | Result |
|---|---|
| `moving_average 实现` (Chinese, names symbol) | 34 nodes, all correct |
| `moving_average implementation` (English) | 34 nodes, all correct |
| `How does JWT authentication work?` (English) | 19 nodes, all correct |
| `JWT 鉴权怎么实现的？` (Chinese, names symbol) | 19 nodes, all correct |
| `What known issues are documented?` (English) | 29 nodes, all correct |
| `rate_limit_check 怎么用？` (Chinese, vague) | 12 nodes, **all wrong** |
| `滑动窗口限流器` (Chinese, vague concept) | 3 nodes, partial match to median_filter |
| `How is rate_limit_check implemented?` (English, vague) | 0 nodes |

The pattern: **queries that name a specific symbol (function name, class name) work in both Chinese and English. Queries that describe a concept without naming a symbol fail or get wrong content.**

## Hypothesis

The BFS seed selection appears to be doing a similarity match between the query embedding and node embeddings, and the matches are dominated by a small set of "concept-dense" nodes (pure function conventions, JWT auth lessons) when the query is vague. The target `rate_limit_check` has 17 edges (top-9 god node) and is in the graph, but it didn't make it into the BFS neighborhood.

This might be:
- A query-side embedding issue (Chinese "rate_limit_check 怎么用" doesn't embed close to "rate_limit_check function" nodes)
- A node-side issue (rate_limit_check node label doesn't include the function's Chinese description)
- A BFS seed-selection issue (top-K similar nodes are not actually similar to the query intent)

## Suggestion

Some investigation directions (not a fix, just where I'd look):
1. Add a "Chinese concept index" to nodes — extract Chinese descriptions from docstrings and link to the function node
2. Increase the BFS seed count for queries with low-confidence top-1 match
3. Surface the seed nodes and let the user filter, rather than silently picking top-K

## Context

I evaluated graphify on two corpora:
- **Small (9 docs synthetic)**: I initially reported "Chinese queries don't work" — but this was a false negative due to embedding sparsity at that scale.
- **Real (40 files, 716 nodes)**: Chinese queries work for *symbol-named* queries but fail for *concept-described* queries. The above reproduction is on the 40-file corpus.

The full evaluation report is here (Chinese + English, with all 10 query outputs):
- https://github.com/kongshan001/spec-driven-eval/tree/main/graphify-evaluation

Happy to provide additional reproduction data or run more queries if useful.

nokternol:
## Summary

Fixes #1445. `_pick_seeds`' `gap_ratio` cutoff discards any BFS-seed candidate
scoring below 20% of the top score. On a multi-term natural-language query,
one term's incidental **exact** label match on a node that is otherwise
unrelated to the query's intent (e.g. a common word that also happens to be
used as a field name or identifier elsewhere in the corpus) scores ~1000x
higher than any **substring** match on the query's other, actually-relevant
terms (`_EXACT_MATCH_BONUS` vs `_SUBSTRING_MATCH_BONUS`). The cutoff then
silently discards every one of those substring-tier candidates as seeds, so
the traversal only ever explores the neighborhood of the one unrelated exact
match — and `query` returns confidently-wrong results with no signal that
anything went wrong.

This is the same shape as #1445's reproduction: a vague query that doesn't
name a target symbol seeds from unrelated "concept-dense" nodes instead, even
though the target node is present in the graph and would have matched on a
weaker (substring) tier.

## Fix

`_pick_seeds` now optionally accepts the graph and the tokenized query terms.
When supplied, it guarantees at least one seed per distinct query term that
has *any* match at all, so one term's incidental collision cannot starve out
the others. Ties within a term are broken by node degree (structural
centrality), so an isolated incidental match doesn't out-rank a real,
well-connected hub for that term.

Both new parameters default to `None`; existing callers that don't pass them
see byte-identical behavior — verified by a new test
(`test_pick_seeds_without_diversity_args_is_unchanged`) alongside the existing
`#897` test suite, which is unmodified and still passing.

## Testing

- Added `test_pick_seeds_diversity_recovers_starved_term`, reproducing the
  exact failure shape from #1445 with a minimal graph (one node that's an
  exact match for an unrelated term, one node that's a substring match for
  the actually-relevant term via an edge from a third node) — confirms the
  target is discarded without the fix and recovered with it.
- Full suite: `74 passed` (`uv run --group dev pytest tests/test_serve.py -q`).
- `ruff check` clean on both changed files.

## Scope note

While investigating this, I found (and locally patched, but am **not**
including here) two related but separable issues on the same code path,
happy to open as follow-ups if useful:

1. A hub-degree threshold in `_bfs` (`max(50, p99_degree)`) that, once seed
   diversity is fixed and BFS starts from more seeds, lets shared
   infrastructure nodes below that floor flood the output. The right general
   fix isn't obvious (a repo-specific floor is a blunt hack), so I left it out
   of this PR rather than guess at the right heuristic.
2. A `keywords`-style node metadata field so enrichment can attach
   paraphrase/alias surface beyond a bare identifier label — this is a real
   feature addition (plus a matching change to the trigram prefilter in
   `_node_search_text`, which silently excludes any field not indexed there),
   not a bug fix, so it felt out of scope for a PR against this issue
   specifically.

Happy to adjust the approach if you'd prefer a different fix shape (e.g.
tuning `gap_ratio` per-term instead of a hard per-term guarantee, or a cap on
total diversity seeds for very long queries).

### DIFF
--- a/graphify/serve.py
+++ b/graphify/serve.py
@@ -326,13 +326,36 @@ def _score_nodes(G: nx.Graph, terms: list[str]) -> list[tuple[float, str]]:
     return scored
 
 
-def _pick_seeds(scored: list[tuple[float, str]], max_k: int = 3, gap_ratio: float = 0.2) -> list[str]:
+def _pick_seeds(
+    scored: list[tuple[float, str]],
+    max_k: int = 3,
+    gap_ratio: float = 0.2,
+    *,
+    G: "nx.Graph | None" = None,
+    terms: list[str] | None = None,
+) -> list[str]:
     """Select BFS seed nodes, stopping when score drops too far below the top.
 
     Prevents high-frequency noise terms (error, exception) from stealing seed
     slots from a dominant identifier match. When FooBarService scores 1000 and
     error nodes score 1.0, only FooBarService is seeded — the score gap is 99.9%
     which is well above the 20% threshold that would allow additional seeds.
+
+    That same gap_ratio cutoff has a failure mode on multi-term natural-language
+    queries: if one term happens to hit an EXACT label match on a node that is
+    otherwise unrelated to the query's intent (e.g. a common word that is also
+    used as an unrelated identifier or field name elsewhere in the corpus), it
+    can outscore every SUBSTRING match on the query's other, actually-relevant
+    terms by ~1000x (see `_EXACT_MATCH_BONUS` vs. `_SUBSTRING_MATCH_BONUS`).
+    The 20%-gap cutoff then silently discards all of those substring-tier
+    seeds, so the BFS traversal only ever explores the neighborhood of the one
+    unrelated exact match — see #1445.
+
+    When `G` and `terms` are supplied, this guarantees at least one seed per
+    distinct query term that has any match at all, so one term's incidental
+    collision cannot starve out the others. Ties within a term are broken by
+    graph degree (structural centrality), so an isolated incidental match
+    doesn't out-rank a real, well-connected hub for that term.
     """
     if not scored:
         return []
@@ -342,6 +365,18 @@ def _pick_seeds(scored: list[tuple[float, str]], max_k: int = 3, gap_ratio: floa
         if seeds and score < top_score * gap_ratio:
             break
         seeds.append(nid)
+
+    if G is not None and terms:
+        norm_terms = sorted({tok for t in terms for tok in _search_tokens(t)})
+        for term in norm_terms:
+            term_scored = _score_nodes(G, [term])
+            if not term_scored:
+                continue
+            best_score = term_scored[0][0]
+            tied = [nid for s, nid in term_scored if s == best_score]
+            best_nid = max(tied, key=lambda n: G.degree(n)) if len(tied) > 1 else term_scored[0][1]
+            if best_nid not in seeds:
+                seeds.append(best_nid)
     return seeds
 
 
@@ -580,7 +615,7 @@ def _query_graph_text(
 ) -> str:
     terms = _query_terms(question)
     scored = _score_nodes(G, terms)
-    start_nodes = _pick_seeds(scored)
+    start_nodes = _pick_seeds(scored, G=G, terms=terms)
     if not start_nodes:
         return "No matching nodes found."
     resolved_filters, filter_source = _resolve_context_filters(question, context_filters)
--- a/tests/test_serve.py
+++ b/tests/test_serve.py
@@ -634,6 +634,42 @@ def test_pick_seeds_respects_max_k():
     assert len(seeds) == 3
 
 
+def test_pick_seeds_without_diversity_args_is_unchanged():
+    """G/terms are optional and default to None: existing callers see identical
+    behavior to before this change."""
+    scored = [(1000.0, "fbs"), (1.0, "err1"), (0.9, "err2")]
+    assert _pick_seeds(scored) == ["fbs"]
+
+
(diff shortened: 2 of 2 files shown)


mark-butterworth-lookinglocal:
Sorry, I'll get Claude to redo this PR tonight... I'm not sure how the patch got so messed up.

safishamsi:
Merged into `v8` as `d56ee83` (your authorship). Verified with a #1445-shaped graph: a multi-term query that used to collapse to a single incidental leaf (`['home']`) now seeds the whole relevant cluster (`['home','alert','dash','grid','kpi']`). Backward-compatible (G/terms are optional kwargs), and the degree tie-break is a good call so an isolated incidental match doesn't out-rank a real hub. Also partially addresses the seed-hijack in #1602 (this is the seed-selection half; the scoring/term-coverage half is still open there). Full suite 2855. Ships next release — thanks.

PR #1597, opened 2026-07-01
fix(serve): drop question/filler stopwords from query terms

Edge4711:
## Problem

`graphify query "<question>"` tokenises the whole question and seeds BFS on every word. Natural-language scaffolding words therefore compete with content words for seed slots — and often win, because a one-word term that prefix-matches the first word of a prose label scores at the 100× prefix tier.

Concrete case, querying a graph of a Python project:

```
graphify query "how does the frontier cache work"
```

`_query_terms` returns `['how', 'does', 'the', 'frontier', 'cache', 'work']`, and the top-scored seeds are:

| node label | score | why it matched |
|---|---|---|
| `How much less pension … the` | 572 | `how` prefix + `the` substring |
| `Working order` / `Working Principles` | 560 | `work` prefix-matches "Working" |
| `Does correcting gold's …` | 489 | `does` prefix |
| `frontier_cache.py` (the actual answer) | 396 | swamped |

BFS then seeds from pension/《Working Principles》 nodes — the wrong part of the graph entirely.

With the scaffolding words removed, the same scorer is perfect: `['frontier', 'cache']` seeds `Frontier Cache — Invalidation Contract` at **4650**, then `frontier_cache.py`, `FRONTIER_CACHE.md`.

## Fix

Drop a set of English question/filler words from the query terms in `_query_terms` so content words drive seeding, with a fallback to the unfiltered terms when a query is entirely stopwords (`"how does it work"` still seeds on something rather than nothing).

Scope guardrails:

- **Query-side only.** The filter runs in `_query_terms`, never on node text, so a symbol literally named `work` stays findable via `explain`/`path`.
- `work`/`works`/`working` are included because "how does X work" / "how X works" is the single most common question phrasing, and (as above) they prefix-match prose labels.
- Chinese/CJK segmentation is untouched — the stopword set is ASCII English only.

## Tests

- Updated `test_query_terms_strips_search_punctuation`, which pinned `"what"` as a kept term (`what calls extract?` → `['calls', 'extract']`; punctuation stripping still asserted).
- Added `test_query_terms_drops_question_stopwords` (the motivating case) and `test_query_terms_all_stopwords_falls_back_to_unfiltered`.
- Full suite green locally: **2745 passed, 28 skipped**.

## Notes

This is the minimal, high-signal fix for the presenting symptom. A complementary idea — folding each node's `community_name` into the scorer so a question can seed on a community label even when no node label matches — is deliberately **out of scope** here; it's orthogonal and much lower-impact (it does nothing for the case above once the scaffolding words are gone). Happy to follow up separately if that's of interest.

### DIFF
--- a/graphify/serve.py
+++ b/graphify/serve.py
@@ -107,8 +107,29 @@ def _is_searchable(term: str) -> bool:
     return True
 
 
+# English question/filler words dropped from query terms so content words drive
+# BFS seeding. Without this, "how does the frontier cache work" seeds on "how"/
+# "the"/"work" (which prefix-match prose labels like "Working Principles" at 100x)
+# instead of "frontier"/"cache", and lands in the wrong part of the graph. Applied
+# to query terms only — node text is never filtered, so a symbol literally named
+# `work` stays findable via explain/path. `work`/`works`/`working` are included
+# because "how does X work" / "how X works" is the most common question phrasing.
+_QUERY_STOPWORDS = frozenset({
+    "how", "what", "why", "when", "where", "which", "who", "whom", "whose",
+    "does", "did", "is", "are", "was", "were", "be", "been", "being",
+    "can", "could", "should", "would", "will", "shall", "may", "might", "must",
+    "has", "have", "had", "the", "and", "but", "not", "for", "from", "with",
+    "without", "into", "onto", "off", "that", "this", "these", "those", "there",
+    "here", "its", "their", "them", "they", "about", "any", "all", "some",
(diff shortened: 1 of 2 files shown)


safishamsi:
Merged into `v8` as `6e97088` (your authorship). Nice — this also directly mitigates #1602 (generic question words hijacking BFS seeds). Query-terms-only (symbols named `work` stay findable via explain/path), with the all-stopwords fallback so "how does it work" still seeds. Tests green, full suite 2800. Ships next release.

PR #1599, opened 2026-07-01
feat: extract rationale comments + ADR/RFC doc references from JS/TS

niltonmourafilho-arch:
## Summary

Parity with `_extract_python_rationale`: Python files get rationale nodes from docstrings and `# NOTE:`-style comments, but JS/TS comments were discarded entirely. This PR adds a post-pass to `extract_js` that:

1. **Rationale comments** — `// NOTE:`, `// WHY:`, `// HACK:` etc. (plus block-comment `* NOTE:` variants) become `rationale` nodes with `rationale_for` edges, matching the existing Python behavior.
2. **Doc references** — `ADR-NNNN` / `RFC NNNN` citations found in comments become `doc_ref` nodes with `cites` edges from the file node.

## Why

The doc_ref pass is the natural join point between code and design docs in mixed corpora. Teams conventionally cite ADR ids in TS file headers, but today those citations produce **zero edges**, so code↔ADR connections never form in the graph even when the citation discipline exists in the codebase.

Tested on a real mixed corpus (Flutter/Supabase monorepo, ~163 TS files + 40 ADRs): a single `router.ts` yields 10 ADR citations that previously produced no edges. With this patch they become direct `cites` edges, closing the code↔ADR gap without any LLM cost (pure line scan, same cost profile as the Python rationale pass).

## Design notes

- Spellings normalized (`ADR-11` / `ADR 0011` → `ADR-0011`) so references to the same document collapse to one node; deduped per file.
- String literals excluded — only comment-shaped lines (`//`, `/*`, `*`) are scanned, so `const s = "ADR-0099"` produces nothing.
- Conservative token set (ADR + RFC only) to avoid noise; easy to extend.

## Test plan

- [x] 5 new tests in `tests/test_rationale.py` (line comment, block comment, multi-ref, normalization/dedup, string-literal exclusion)
- [x] `tests/test_rationale.py` 18/18 pass
- [x] `tests/test_extract.py` + `tests/test_build.py` + `tests/test_languages.py`: failure list identical before/after patch (5 pre-existing Windows symlink/path failures, unrelated)

### DIFF
--- a/graphify/extract.py
+++ b/graphify/extract.py
@@ -5304,7 +5304,114 @@ def extract_js(path: Path) -> dict:
         config = _TS_CONFIG
     else:
         config = _JS_CONFIG
-    return _extract_generic(path, config)
+    result = _extract_generic(path, config)
+    if "error" not in result:
+        _extract_js_rationale(path, result)
+    return result
+
+
+# ── JS/TS rationale + doc-reference extraction ────────────────────────────────
+#
+# Parity with _extract_python_rationale: Python files get rationale nodes from
+# docstrings and `# NOTE:`-style comments, but JS/TS comments were discarded
+# entirely. That silently drops two high-value signals in mixed corpora:
+#   1. rationale comments (`// NOTE:`, `// WHY:`, ...) — same as Python;
+#   2. architecture-decision references (`ADR-0011`, `RFC 793`) that teams
+#      conventionally cite in file/function headers. These are the natural
+#      join points between code and design docs in the same graph — without
+#      them, code<->ADR edges never form even when the code cites the ADR.
+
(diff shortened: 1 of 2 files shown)


safishamsi:
Merged into v8 as `6d3a6f1`, authorship preserved (ships in 0.9.7). Thanks @niltonmourafilho-arch. Verified: on v8 a JS/TS file with `// NOTE:` / `// WHY:` and `ADR-NNNN`/`RFC NNNN` produced zero rationale/doc_ref nodes; with the patch they become `rationale` nodes (`rationale_for` edges) and `doc_ref` nodes (`cites` edges), matching the Python pass, and a file with no such comments stays unaffected. Your 18 tests plus the full suite pass.

PR #1601, opened 2026-07-01
perf(hooks): eliminate multi-second foreground stalls before the detached launch

TNRealtorApp:
## Problem

On Windows (and any machine with a cold import cache or AV-scanned site-packages), `git commit` stalls for 10-40+ seconds inside the graphify post-commit hook **before** the #1161 detached launcher even starts:

1. **Interpreter probes import the full package.** Each probe runs the equivalent of `python -c 'import graphify'`, which executes the whole package import — measured **13s per probe** (cold) on a Windows 11 dev box — and up to four probes can run synchronously per commit.
2. **The shebang probe reads a binary.** Git for Windows' `command -v graphify` can return the launcher path **without** its `.exe` suffix, so the `*.exe)` guard misses and `head -1` reads a PE binary: the shell prints `warning: command substitution: ignored null byte in input` on every commit, and the garbage 'shebang' always fails the allowlist and falls through to the slowest fallbacks (including the Microsoft Store `python3` alias).
3. **Redundant `git rev-parse`.** git already exports `GIT_DIR` to hooks; the unconditional re-derive costs another ~1.3s on machines where every git exec is AV-scanned.

## Fix

- Probes use `importlib.util.find_spec('graphify')` — locates the package without executing it, so each probe costs interpreter startup only. A broken install under the selected interpreter still fails loudly in the rebuild log (same as today). Probe order and fallback semantics are unchanged.
- The Windows pip layout is resolved directly: `Scripts/graphify(.exe)` → sibling `..\python.exe` (or `.\python.exe` for venv Scripts dirs) — no binary parsing, works with or without the `.exe` suffix.
- The remaining POSIX shebang read strips NUL bytes first (`head -c 256 | tr -d '\000' | head -n 1`).
- `GIT_DIR=${GIT_DIR:-$(git rev-parse --git-dir 2>/dev/null)}` in both hooks (rev-parse remains the manual-invocation fallback).

## Measured (Windows 11, Python 3.13, real repo)

| | before | after |
|---|---|---|
| hook foreground (cold) | 26s+ (2 x 13s import probes) | ~1.4s (one find_spec probe) |
| null-byte warnings | every commit | none |

## Tests

`tests/test_hooks.py`: 47 passed (4 new — find_spec probes, NUL-safe shebang read, sibling python.exe resolution, GIT_DIR reuse; plus the `_launcher_payload` docstring updated to match).

### DIFF
--- a/graphify/hooks.py
+++ b/graphify/hooks.py
@@ -21,9 +21,18 @@ _PYTHON_DETECT = """\
 # Detect the correct Python interpreter (handles uv tool, pipx, venv, system installs).
 # _PINNED was recorded at hook-install time; tried first so the hook works even
 # when the graphify launcher is not on PATH (common in GUI clients and CI).
+#
+# Probes check availability with importlib.util.find_spec instead of importing
+# the package: a probe that imports graphify wholesale executes the full package
+# import (10s+ cold on machines with AV-scanned or large site-packages) and used
+# to run up to FOUR times synchronously, stalling every commit before the
+# detached launch even started. find_spec locates the package without executing
+# it, so each probe costs interpreter startup only. The detached rebuild still
+# fails loudly in the log if the package is broken under that interpreter.
+_GFY_PROBE="import importlib.util, sys; sys.exit(0 if importlib.util.find_spec('graphify') else 1)"
 GRAPHIFY_PYTHON=""
 _PINNED='__PINNED_PYTHON__'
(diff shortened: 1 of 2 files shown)


safishamsi:
Merged into `v8` as `1256d65` (your authorship). Verified: probes now use `find_spec` (no full package import), NUL-safe shebang read, Windows sibling-python resolution, and GIT_DIR reuse — 4 new tests + full suite 2800 green. The find_spec/detached tradeoff is sound (foreground picks the interpreter; the detached rebuild still logs a broken install). Genuinely nice perf work for Windows/cold-cache hooks. Ships next release.

PR #1603, opened 2026-07-02
support java

ZongYue99:


safishamsi:
Merged the code + tests into v8 as `92edf78`, authorship preserved (ships in 0.9.7). Thanks @NydiaChung. Verified: on v8 a Java file emitted `references` edges to `String`/`List`/`Optional`/`Integer` (stdlib noise) alongside real user types; with `_JAVA_BUILTIN_TYPES` those are suppressed while `List<Item>` still keeps the `Item` edge. 25 java tests + full suite pass. Note: I applied only the extract.py + test changes and wrote a fresh changelog entry, since the PR's CHANGELOG/uv.lock had drifted from the current base.

PR #1607, opened 2026-07-02
Treat .mts/.cts (TypeScript module extensions) as TypeScript

ashmitg:
## Problem

`.mts` (ESM) and `.cts` (CommonJS) — the TypeScript counterparts of the already-supported `.mjs`/`.cjs` — were handled in **neither** graphify's code-extension detection **nor** its TypeScript grammar path. Two compounding effects:

1. **Skipped entirely** — absent from `CODE_EXTENSIONS`, so `.mts`/`.cts` files were classified as non-code and never extracted.
2. **Under-parsed even once routed to the JS extractor** — the grammar selector picks the TypeScript grammar only for literal `.ts`/`.tsx`, so `.mts`/`.cts` fell back to the **plain JavaScript grammar**, which silently drops every `type` / `interface` / type-alias.

Real impact: on a TypeScript repo whose tooling/config layer is `.mts` (files run under `tsx`/`ts-node`), those files were dropped from the graph — and where detected, their type surface vanished.

## Fix — treat `.mts`/`.cts` as TypeScript everywhere `.ts` is handled

**Detection** (add alongside `.mjs`/`.cjs`/`.ts`):
| File | Set |
|---|---|
| `detect.py` | `CODE_EXTENSIONS` — the code/non-code gate |
| `analyze.py` | `_LANG_FAMILY` → `"js"` |
| `extract.py` | `_JS_RESOLVE_EXTS`, `_JS_CACHE_BYPASS_SUFFIXES` |
| `build.py` | per-node language map → `"js"` |

**TypeScript grammar** (`.mts`/`.cts` are non-JSX → the plain TypeScript grammar, like `.ts`):
| Site | Change |
|---|---|
| `extract_js()` grammar selector | `.mts`/`.cts` → `_TS_CONFIG` |
| `_DISPATCH` | `.mts`/`.cts` → `extract_js` |
| `use_ts` tree-sitter selector | include `.mts`/`.cts` |
| `typescript_member_calls` resolver | include `.mts`/`.cts` |

Same `tree-sitter-typescript` grammar as `.ts` → **no new dependency**.

## Result (verified on a real 47 KB `.mts` file)

`.mts`/`.cts` now extract **identically to `.ts`** — 36 nodes each, vs 22 under the JS grammar. The 14 recovered nodes are exactly the TS `type`/`interface`/type-alias declarations the JS grammar was dropping.

## Tests

`tests/test_typescript_module_extensions.py`:
- membership regression locks for the four extension sets + the `_DISPATCH` routing, and
- end-to-end parity: a `.mts`/`.cts` source with a `type` alias + `interface` captures those TS-only nodes and produces the **same node set** as the equivalent `.ts`.

Verified locally (Python 3.12, `uv run --frozen`): `ruff` clean · new tests 6/6 · `skillgen --check` OK (134 artifacts) · broader JS/TS + detection + multilang suites **578 passed, 24 skipped, 0 regressions**.

## Note

`.cjs` is in the `build.py`/`analyze.py` JS maps but **not** in `detect.CODE_EXTENSIONS` (only `.mjs` is), so `.cjs`-only files may be detected as non-code too. Kept this PR scoped to `.mts`/`.cts`; happy to fold `.cjs` in if you'd like.


safishamsi:
Merged into `v8` as `1226c34` (your authorship). Thorough — hits all six extension sets (detection, language family, cache-bypass, resolve-exts, resolver registry, dispatch) with locks on each. Verified `.mts`/`.cts` route to the TS grammar and `.ts` is unaffected. Full suite 2841. Ships next release.

PR #1613, opened 2026-07-02
Contain symlinked extraction inputs

Tok6Flow0:
## Summary
- make symlink directory following explicit opt-in instead of auto-enabled
- require resolved corpus paths to stay under the scan root before detection, AST collection, and LLM/image extraction reads
- skip out-of-root symlink targets while preserving in-root symlink support

## Security impact
A corpus can contain symlinks that resolve outside the selected scan root. Previously, direct symlinked directories could be followed automatically and symlinked files could be read by default. If semantic extraction is enabled, those outside-root file contents can enter the extraction prompt or image payload. This patch treats the scan root as a containment boundary before reading or sending file contents.

## Tests
- `python -m pytest tests/test_detect.py tests/test_image_vision.py tests/test_extract.py -q`
  - 256 passed

### DIFF
--- a/graphify/detect.py
+++ b/graphify/detect.py
@@ -991,14 +991,11 @@ def _could_contain_included_path(path: Path, root: Path, patterns: list[tuple[Pa
 
 
 def _auto_follow_symlinks(root: Path) -> bool:
-    """Auto-detect: ``True`` if ``root`` has any direct symlinked child.
+    """Return whether ``root`` has any direct symlinked child.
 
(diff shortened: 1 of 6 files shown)


safishamsi:
Merged into `v8` as `009a98b` (your authorship). Verified the containment: an in-corpus symlink whose target resolves outside the scan root is skipped at detection (and the AST/LLM read boundaries), while in-root files extract normally and symlink-following stays explicit opt-in. Good security hardening. Full suite 2841. Ships next release.

PR #1615, opened 2026-07-02
fix(extract): four TS/JS extractor gaps — generators, namespace containers, decorators, import-equals

papinto:
## Summary

Four independent gaps in the TS/JS extractor, found while running graphify against a large TypeScript front end. One commit per fix so they can be reviewed — or cherry-picked — individually. Each commit carries its own regression tests, and none of them changes existing behavior for code that was already extracted (the ESM import path is regression-guarded explicitly).

### 1. Generator functions were not nodes (`0bb8895`)

The declaration form `function* g()` parses as `generator_function_declaration`, which was absent from the JS/TS `function_types`, so it produced no node; the expression form `const h = function*(){}` parses as `generator_function`, which was absent from the JS function-value types. Generator *methods* (`*gen()` in a class) were already covered via `method_definition`.

The fix adds `generator_function_declaration` to the JS/TS `function_types` and `function_boundary_types` (so a generator is a node and its body's calls attribute to it, parity with `function_declaration`), and `generator_function` to the JS function-value types (so the const-assigned form is captured like `function_expression`).

### 2. TS namespace / module containers were not nodes (`d4418fc`)

`namespace Foo {}` (`internal_module`) and `module Bar {}` / ambient `declare module "pkg" {}` (`module`) were not in any type set and had no extra-walk, so the container produced no node — members were still reached by the default recurse, but the namespace itself was invisible and its members lost their namespace context.

The fix adds `_ts_extra_walk` (dispatched for TypeScript after `_js_extra_walk`, mirroring the C# extra-walk): it emits a container node plus a file→container `contains` edge and recurses the body, leaving members file-contained exactly as before. An `is_named` guard skips the anonymous `module` keyword token, which shares the `module` type string.

### 3. Decorators produced no edges (`71632b2`)

`@Component`, `@Injectable`, `@Input`, `@Inject`, `@Entity`, … — the `decorator` node kind was never walked, so decorators produced no edge at all. This is framework-critical signal (Angular, NestJS, Vue class components, TypeORM): decorators are often the primary statement of what a class is and does.

The fix emits a `references` edge (context=`"decorator"`) from the decorated entity to the decorator symbol: class decorators → the class (both `@Deco class C` and `@Deco export class C`, plus stacked decorators), method decorators → the method node, field/accessor decorators → the class (the field is not a graph node), parameter decorators (`@Inject(T)`) → the enclosing method/constructor. The symbol is the head identifier (`@Injectable`, the function of `@Component({...})`, or the property of `@ns.Component()`), routed through `ensure_named_node` so out-of-corpus decorators become sourceless stubs consistent with type references.

### 4. The TS import-equals form was invisible (`31aed9a`)

`import x = require("./m")` produced no edge at all: tree-sitter parses it as an `import_statement` whose module string sits inside an `import_require_clause`, not as a direct child of the statement, so the direct-child string scan in `_import_js` never found it — while the equivalent ESM `import * as x from "./m"` was captured.

The fix restructures the scan to locate the module string in either position and emits the `imports_from` edge from the single shared path. Relative paths, tsconfig aliases, and bare modules all resolve through the same `_resolve_js_import_target` as ESM, giving the import-equals form exact parity with a namespace import (verified by a require-vs-ESM parity test). Plain JS is unaffected (its grammar has no `import_require_clause`), and the pure alias form `import A = B.C` is deliberately out of scope — it has no module string and models an intra-code alias, not an import.

## Testing

- 25 new tests across four files: `test_ts_generators.py` (5), `test_ts_namespace.py` (6), `test_ts_decorators.py` (9), `test_ts_import_require.py` (5).
- Full relevant suites green on top of `v8` @ `d89ec68`: `test_languages`, `test_js_import_resolution`, `test_symbol_resolution`, `test_language_resolvers`, `test_ts_inheritance`, `test_vue_extraction` + the four new files — **451 passed, 13 skipped**.
- Each fix was validated against real-world code from a large production TypeScript/Vue codebase before being reduced to the minimal test fixtures here.

🤖 Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/graphify/extract.py
+++ b/graphify/extract.py
@@ -1787,12 +1787,23 @@ def _import_js(node, source: bytes, file_nid: str, stem: str, edges: list, str_p
                 return
 
     resolved_path: "Path | None" = None
+    module_string = None
     for child in node.children:
         if child.type == "string":
-            raw = _read_text(child, source).strip("'\"` ")
-            resolved = _resolve_js_import_target(raw, str_path)
-            if resolved is None:
-                break
+            module_string = child
+            break
+        if child.type == "import_require_clause":
+            # TS import-equals form: `import x = require("./m")`. The module
+            # string sits inside the clause, not on the import_statement
+            # itself, so the direct-child scan above never sees it.
+            module_string = next(
+                (sub for sub in child.children if sub.type == "string"), None
+            )
+            break
+    if module_string is not None:
+        raw = _read_text(module_string, source).strip("'\"` ")
+        resolved = _resolve_js_import_target(raw, str_path)
+        if resolved is not None:
             tgt_nid, resolved_path = resolved
             edges.append({
                 "source": file_nid,
@@ -1804,7 +1815,6 @@ def _import_js(node, source: bytes, file_nid: str, stem: str, edges: list, str_p
                 "source_location": f"L{node.start_point[0] + 1}",
                 "weight": 1.0,
             })
-            break
 
     # Emit symbol-level edges for named imports/re-exports from local/aliased files.
     # e.g. `import { Foo, type Bar } from './bar'` → file → Foo, file → Bar (EXTRACTED)
--- a/graphify/extract.py
+++ b/graphify/extract.py
@@ -3357,6 +3357,12 @@ def _extract_generic(
             callable_def_nids.add(class_nid)  # a class is callable (constructor)
             add_edge(file_nid, class_nid, "contains", line)
 
+            # TS/JS decorators on the class and its members (@Component, @Injectable,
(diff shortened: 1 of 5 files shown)


safishamsi:
Merged into `v8` (your authorship) — all four commits: generators (`09aeb97`), namespace/module containers (`869aaf7`), decorators (`3540416`), import-equals (`9811def`). Verified each: `function*` decl + const-generator register; `namespace`/`module`/`declare module` become nodes (members still contained); `import x = require()` emits an import edge alongside normal imports. Full suite 2841. Ships next release. Nice, well-scoped batch.

PR #1616, opened 2026-07-02
feat: add `pascal` optional extra for tree-sitter-pascal

vinicius-l-machado:
extract_pascal() already imports tree-sitter-pascal for AST-quality extraction and falls back to a regex extractor when it is absent (#781), but the grammar was not declared anywhere in the package metadata, so it was never installed and the AST path never ran out of the box.

Declare a `pascal` extra (and add it to `all`) so users can opt into the AST extractor with `uv tool install "graphifyy[pascal]"`. tree-sitter-pascal publishes prebuilt wheels for every platform (win/macOS/Linux), so unlike the `dm` extra it needs no C toolchain.

On a mid-size Delphi codebase the AST path yields notably more accurate relationship edges than the regex fallback (calls and inherits both up ~25%). README extras table and uv.lock updated accordingly.


safishamsi:
Merged into v8 as `7d463c9`, authorship preserved (ships in 0.9.7). Thanks @vinicius-l-machado. Verified the `pascal` extra is declared (and in `[all]`), the regex fallback still works when the grammar is absent, and the extractor resolves the AST path when installed. One note: your uv.lock had drifted from the current base, so I regenerated it (`uv lock`) and confirmed the diff is scoped to tree-sitter-pascal only and `uv lock --check` passes, so CI's --frozen sync stays green.

PR #1617, opened 2026-07-02
fix: persist cluster-only analysis sidecar

### LINKED ISSUE #1610: cluster-only doesn't persist .graphify_analysis.json, causing export html to silently report "Single community"
## Bug: `graphify cluster-only` doesn't leave `.graphify_analysis.json` behind, so a subsequent `graphify export html` silently skips with "Single community"

**Version:** graphifyy 0.8.13 (installed via `uv tool install`)

### Steps to reproduce

1. Run the full pipeline once on a corpus large enough to exceed the HTML viz node limit (>5000 nodes), so `graphify-out/graph.json` and `graphify-out/GRAPH_REPORT.md` exist.
2. Manually edit `graphify-out/graph.json` (e.g. merge two duplicate nodes) so the graph no longer matches the last full extraction.
3. Run `graphify cluster-only .` to re-cluster the edited graph. It reports success (e.g. `Done — 387 communities. GRAPH_REPORT.md and graph.json updated.`) and, in my case, also printed `Skipped graph.html: ... too large for HTML viz (limit: 5000)`.
4. Run `graphify export html` to get the aggregated community view for the oversized graph.

### Expected

`graphify export html` builds the aggregated community meta-graph (one node per community, as it does when run right after the initial full pipeline) and writes `graph.html`.

### Actual

```
Graph has 5558 nodes (above 5000 limit). Building aggregated community view...
Single community - aggregated view not useful. Skipping graph.html.
```

No error, no indication anything is wrong — it just silently produces a 0-node/1-node-looking result and skips the file.

### Root cause

`export html`'s CLI handler reads community assignments from a specific file, not from `graph.json`'s own per-node `community` attribute:

```python
# graphify/__main__.py, ~line 2119
analysis_path = Path(_GRAPHIFY_OUT) / ".graphify_analysis.json"
...
# ~line 2240-2242
communities: dict[int, list[str]] = {}
if analysis_path.exists():
    _an = json.loads(analysis_path.read_text(encoding="utf-8"))
    communities = {int(k): v for k, v in _an.get("communities", {}).items()}
```

`graphify cluster-only` never writes `.graphify_analysis.json` — it computes communities internally and writes only `graph.json` (with per-node `community` fields) and `GRAPH_REPORT.md`, then presumably cleans up its own intermediates. So after a `cluster-only` run, `.graphify_analysis.json` doesn't exist, `communities` ends up `{}`, and `to_html`'s aggregation path (`graphify/export.py`, `to_html()`, the `node_limit is not None` branch) iterates zero communities:

```python
meta = _nx.Graph()
for cid, members in communities.items():   # communities == {} here
    meta.add_node(...)
...
if meta.number_of_nodes() <= 1:
    print("Single community - aggregated view not useful. Skipping graph.html.")
    return
```

— hence the misleading "Single community" message on a graph that actually has hundreds of real communities (still recoverable from `graph.json`'s per-node `community` field).

### Suggested fix

Either:
- Have `cluster-only` also (re)write `.graphify_analysis.json` (communities/cohesion/gods/surprises) alongside `graph.json` and `GRAPH_REPORT.md`, matching what the full pipeline leaves behind at the equivalent step, or
- Have `export html`'s CLI handler fall back to deriving `communities` from `graph.json`'s per-node `community` field when `.graphify_analysis.json` is missing, rather than defaulting to `{}`.

Either fix would also make the "Single community" message accurate again — right now it fires even when the graph clearly has many communities, which is confusing to debug from the CLI output alone (I only found the cause by reading `graphify/export.py` and `graphify/__main__.py` source directly).

### Workaround

Manually reconstruct and write `graphify-out/.graphify_analysis.json` with `{"communities": ..., "cohesion": ..., "gods": ..., "surprises": ...}` (derived from `graph.json`'s per-node `community` field) before calling `graphify export html`.

sanmaxdev:
## Summary
- Persist `.graphify_analysis.json` during `graphify cluster-only` / `graphify label` runs.
- Add a regression test that verifies the refreshed sidecar matches the communities written to `graph.json`.

## Testing
- [x] `uv run --frozen pytest tests/test_cli_export.py -q --tb=short`
- [x] `uv run --frozen pytest tests/test_cli_export.py tests/test_serve.py -q --tb=short`
- [x] `uv run --frozen pytest tests/ -q --tb=short`
- [x] `uv run --frozen ruff check graphify/__main__.py tests/test_cli_export.py`
- [x] `uv run --frozen python -m tools.skillgen --check`
- [x] `uv run --frozen python -m tools.skillgen --audit-coverage`
- [x] `uv run --frozen python -m tools.skillgen --schema-singleton`
- [x] `uv run --frozen python -m tools.skillgen --monolith-roundtrip`
- [x] `uv run --frozen python -m tools.skillgen --always-on-roundtrip`
- [x] `uv run --frozen graphify --help`
- [x] `uv run --frozen graphify install`
- [x] `git diff --check`

Closes #1610

### DIFF
--- a/graphify/__main__.py
+++ b/graphify/__main__.py
@@ -3655,6 +3655,17 @@ def main() -> None:
         stages.mark("report")
         from graphify.export import backup_if_protected as _backup
         _backup(out)
+        analysis = {
+            "communities": {str(k): v for k, v in communities.items()},
+            "cohesion": {str(k): v for k, v in cohesion.items()},
+            "gods": gods,
+            "surprises": surprises,
+            "questions": questions,
+        }
+        (out / ".graphify_analysis.json").write_text(
+            json.dumps(analysis, indent=2, ensure_ascii=False),
+            encoding="utf-8",
+        )
         to_json(G, communities, str(out / "graph.json"), community_labels=labels)
         labels_path.write_text(json.dumps({str(k): v for k, v in labels.items()}, ensure_ascii=False), encoding="utf-8")
         # Membership signatures beside the labels so a later cluster-only can detect
--- a/tests/test_cli_export.py
+++ b/tests/test_cli_export.py
@@ -323,6 +323,36 @@ def test_cluster_only_creates_output_dir_when_missing(tmp_path):
 
 # Regression test for #1027 - cluster-only must remap labels via node overlap
 
+def test_cluster_only_persists_analysis_sidecar(tmp_path):
+    """cluster-only must refresh .graphify_analysis.json alongside graph.json.
+
+    Downstream export commands use the sidecar for community membership and
+    should not see stale or missing community analysis after a recluster.
+    """
+    out = _make_graph(tmp_path)
+    analysis_path = out / ".graphify_analysis.json"
+    analysis_path.unlink()
+
+    r = _run(["cluster-only", ".", "--no-viz"], tmp_path)
+    assert r.returncode == 0, r.stderr
+    assert analysis_path.exists()
+
+    analysis = json.loads(analysis_path.read_text(encoding="utf-8"))
+    assert analysis["communities"]
+    assert analysis["cohesion"]
+    assert "gods" in analysis
+    assert "surprises" in analysis
+    assert "questions" in analysis
+
+    graph = json.loads((out / "graph.json").read_text(encoding="utf-8"))
+    graph_cids = {
+        str(node["community"])
+        for node in graph.get("nodes", [])
+        if node.get("community") is not None
+    }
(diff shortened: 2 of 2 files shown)


safishamsi:
Merged into `v8` as `62f49ba` (your authorship) — this closes #1610. Verified cluster-only now writes `.graphify_analysis.json` with communities/cohesion/gods/surprises/questions, matching the full extract path's shape, so a later `export html` no longer reports "Single community". Full suite 2841 green. Ships next release. Thanks!

Issue #1609, opened 2026-07-02
C#: method calls on a typed receiver (`field/param/local.Method()`) are not emitted as `calls` edges


JensD-git:
## Summary
For C#, invocations of the form `receiver.Method(...)` where `receiver` is a strongly-typed field, property, parameter, or local variable do **not** produce a `calls` edge to the receiver type's method — even though the receiver's type is already known to graphify (a `references` edge to that type is emitted). Only unqualified / `this.` calls and  static-import calls are captured.

The type information needed to resolve these calls is present; it just isn't linked to the invocations. Swift and Python already have member-call resolution (`_resolve_swift_member_calls`, `_resolve_python_member_calls`); C# has none.

## Minimal reproduction
```csharp
// File: Sample.cs
public class Server
{
    public bool Save() => true;
}

public class Repository
{
    private Server _server = new Server();          // typed field

    public bool Commit()
    {
        Log();                                       // (A) unqualified -> edge IS emitted
        return _server.Save();                       // (B) member call  -> edge is MISSING
    }

    private void Log() { }
}
```

**Expected:** two `calls` edges — `Repository.Commit -> Repository.Log` (A) and `Repository.Commit -> Server.Save` (B).

**Actual:** only (A) is emitted. (B) is dropped. A `references` edge `Repository -> Server` *is* present, so the type is known.

The same happens when the receiver is a **method parameter** or a **local variable**:
```csharp
public static bool CopyToServer(Server server)     // parameter
{
    return server.Save();                            // edge MISSING
}

public bool Run()
{
    Server s = new Server();                         // local (also: var s = new Server();)
    return s.Save();                                 // edge MISSING
}
```

## Why it matters
In codebases where classes delegate through typed member/parameter objects (wrappers, service layers, transpiled code), this is not an edge case — it is the majority of the real call graph. In one C# project we measured a single wrapper class with **119** distinct `field.Method(...)` delegation call sites resolving to **0** `calls` edges (the receiver type was known via a `references` edge), and a service class with **35** `parameter.Method(...)` sites → **0** `calls` edges (20 `references` edges to the receiver type present). Path/blast-radius/`query` results across those boundaries are effectively blind.

[...]

PR #1620, opened 2026-07-03
Namespace-aware C# type resolution and member-call receiver inference

TheFedaikin:
## Summary

Adds two C# capabilities and splits the C# extractor out of the 16.6k-line
`graphify/extract.py` into dedicated modules:

1. **Cross-file type resolution** — `class`/`interface`/`enum`/`struct`/`record`
   references resolve to their real definitions across files, namespace- and
   `using`-aware, with lexical using-scope and qualified references.
2. **Member-call receiver resolution** — `receiver.Method()` calls become real
   `calls` edges by inferring the receiver's type, under a strict
   **never-wrong-edge bar**: any ambiguity is skipped, never guessed.

**This supersedes #1609** (see below).

**This builds on and improves upon `#1609`** (see below).

## Relationship to `#1609`

#1609 ("C# receiver-typed member-call resolution") introduced the first C#
receiver-typed resolver: a file-wide, first-binding-wins name → type table with
`this` / capitalized-type / receiver-type resolution, inline in `extract.py`.
**This PR builds on that foundation and improves upon it** with a modular, tiered
implementation — it keeps every `#1609` case working (including the
`_server.Save()` lowercase-field ambiguity) and adds more receiver tiers,
method-return `var` inference, lexical (rather than file-wide) scoping with
shadow/poison handling, and the module split. `#1609`'s inline resolver is folded
into the modular version here, and its cases are carried into the 61-test suite
alongside the new tiers.

## Cross-file type resolution

- Type declarations carry their **enclosing namespace** (block, nested, and
  file-scoped `namespace N;`) and a lexical scope chain.
- References resolve **namespace- and `using`-aware**: file-scoped `using`,
  namespace-scoped `using`, `global using`, and the global namespace.
- **Qualified references** (`N.Type`) and unqualified references both resolve to
  the real definition, including cross-file inheritance (`class Weapon : Damage`).
- **Collision disambiguation** — when the same type name is defined in multiple
  namespaces, a reference binds only to the **unique** candidate reachable from the
  referrer's usings; genuinely ambiguous references stay dangling rather than bind
  to an arbitrary definition.

## Member-call receiver resolution

`obj.M()` resolves to the real method when the receiver's type is known or safely
inferable:

- **Declared locals** — `Service x = …; x.M()`
- **Parameters, fields, properties, `this.field`** — including inherited members
  through a single resolvable base chain
- **`this.M()` / implicit-enclosing** — `M()` inside a type
- **Static / dotted** — `Type.M()`, `Ns.Type.M()`
- **Records** — positional parameters
- **Method-return `var`** — `var x = Get(); x.M()`, inferred from the callee's
  unique **bare** return type resolved in the **callee's** namespace context
  (emitted as `INFERRED`)

Everything else is **skipped, not guessed**: overloaded RHS, non-bare returns
(`T[]`, `T?`, `N.T`, `Box<T>`, `ref`, tuples, type parameters, `void`),
reassigned/redeclared or shadowed locals, cross-file name collisions without a
unique resolution, extension methods, and chained receivers (`a.b.c.M()`).

## Modular C# extractor (no behavior change from the split)

- New `extractors/csharp_extract.py` — per-file C# extraction (binding/type-table/
  shadow model, type references, imports).
- New `extractors/csharp_resolve.py` — the member-call resolver.
- `extractors/csharp.py` — cross-file name resolver + corpus passes.
- `LanguageConfig` → `extractors/base.py`; C#-only logic lifted out of the shared
  `_extract_generic` core into fact-returning helpers.
- `extract.py` re-exports every moved name (import surface unchanged); the split
  itself is byte-identical extractor output, verified by a per-entry-point
  node/edge snapshot.

## Docs

- `extractors/MIGRATION.md` updated to record the C# module layout and the
  "split the language helpers + cross-file resolver, leave the config-driven
  `extract_<lang>` entry point in `extract.py`" pattern — a middle path for
  config-driven languages the doctrine previously said had to wait for the shared
  `_extract_generic` batch.

## Testing

- Full suite: **2873 passed, 0 failed** (on current upstream).
- `tests/test_csharp_member_calls.py`: 61 receiver-resolution + never-wrong-edge
  tests. `tests/test_csharp_type_resolution.py`: cross-file resolution.
- The never-wrong-edge guards are mutation-verified — removing any one guard
  (overload, RHS allow-list, scope-shadow stop, callee-context resolution) fails
  a test.


TheFedaikin:
Hey! 
@safishamsi can I get a review on this? Repo is moving very fast and I am merging a lot of conflicts since this is a big effort.

TheFedaikin:
Updated with current `C#` work and superseeded the new testing cases for method chaining. This branch also fixes the flaky `test_label_communities_batches_when_over_batch_size` by ensuring we are comparing the batches sorted because they can be done in any order in the implementation.

safishamsi:
Thanks @TheFedaikin, and sorry for the long review cycle. The core value of this PR shipped in v0.9.27: C# member calls on a typed receiver now resolve namespace-aware (a class name used in two namespaces resolves via using/scope instead of bailing), with base./this.field receivers, inherited-member lookup through the inherits chain, and shadow-poisoning so a local shadowing a field of a different type no longer produces a wrong edge. Your approach is credited in the commit and the changelog ("adapted from #1620 by @TheFedaikin").

I reworked it onto current v8 as a focused change rather than merging this PR directly: it had grown to ~4,900 lines and gone CONFLICTING against v8, it deleted the #1609 regression tests, and it bundled the module split plus unrelated test changes. Two of its behaviors I deliberately held for now: the module-file split, and emitting a references edge to the receiver type when the method isn't found (kept the existing skip-when-not-found semantics). If you'd like to land the module split or that fallback as separate, focused PRs, I'd be glad to review them. Closing this one as adapted. https://github.com/Graphify-Labs/graphify/releases/tag/v0.9.27

TheFedaikin:
@safishamsi thank you very much!

PR #1624, opened 2026-07-03
Stamp graph.json with generated_at, warn on stale graphify query

edgestack-ai:
## Summary
- `export.to_json` (the single write chokepoint every build/update/watch path funnels through) now stamps `generated_at` (ISO UTC) into `graph.json`'s top level, alongside the existing `built_at_commit`.
- `graphify query` compares that stamp against the last commit time of the repo the graph indexes (resolved via the same `.graphify_root` sidecar convention `build.py::_infer_merge_root` already uses) and prints one warning line to stderr when the graph is older than the repo's latest commit.
- Missing/unreadable stamps (graphs built by an older graphify version) warn too, telling the caller to regenerate.
- Outside a git repo, or when git isn't available, the check is a silent no-op — it never raises, and query output/exit code is otherwise unchanged.

## Test plan
- [x] `uv run ruff check graphify/export.py graphify/__main__.py` — clean
- [x] `uv run pytest tests/test_export.py tests/test_query_cli.py tests/test_cli_export.py` — 73 passed
- [x] `uv run pytest tests/` — 2769 passed / 27 pre-existing failures (all Windows path-separator flakiness in unrelated tests, reproduced identically on `v8` HEAD without this change)
- [x] Manual end-to-end verify: built a tiny demo repo + graph.json, confirmed a fresh graph produces no warning, confirmed advancing the demo repo's history (new commit after the stamp) produces `[graphify] warning: graph.json was generated ... but the indexed repo's last commit is ... - graph is stale, run \`graphify .\` (or \`graphify update\`) to refresh` on stderr, confirmed a stamp-less graph produces a "no stamp, regenerate" warning, and confirmed no crash/warning outside a git repo.

### DIFF
--- a/graphify/__main__.py
+++ b/graphify/__main__.py
@@ -2898,6 +2898,13 @@ def main() -> None:
             if "links" not in _raw and "edges" in _raw:
                 _raw = dict(_raw, links=_raw["edges"])
             try:
+                from graphify.export import check_staleness
+                _stale_warning = check_staleness(_raw, gp)
+                if _stale_warning:
+                    print(_stale_warning, file=sys.stderr)
+            except Exception:
+                pass  # staleness check is diagnostic-only, never blocks a query
+            try:
                 G = json_graph.node_link_graph(_raw, edges="links")
             except TypeError:
                 G = json_graph.node_link_graph(_raw)
--- a/graphify/__main__.py
+++ b/graphify/__main__.py
@@ -3867,7 +3867,7 @@ def main() -> None:
             except TypeError:
                 return _jg.node_link_graph(data), data
         try:
-            G_cur, _ = _load_graph(_current_path)
+            G_cur, _data_cur = _load_graph(_current_path)
             G_oth, _ = _load_graph(_other_path)
         except Exception as exc:
             print(f"[graphify merge-driver] error loading graphs: {exc}", file=sys.stderr)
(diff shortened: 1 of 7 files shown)


itskaism:
Independent confirmation that this is still reproducible on the current release, with numbers — posting as evidence rather than a competing patch, since #1624 already implements the fix.

**Still unfixed in 0.9.33** (PyPI, installed fresh today). A `graph.json` built by `graphify extract . --code-only --no-cluster` carries **no timestamp of any kind**:

```python
>>> d = json.load(open("graphify-out/graph.json"))
>>> [k for k in d if not isinstance(d[k], list)]
['input_tokens', 'output_tokens']
>>> d.get("generated_at"), d.get("built_at_commit")
(None, None)
```

So neither the stamp this PR adds nor the pre-existing `built_at_commit` it mentions is present in a released build — the consumer side has nothing to check even if it wanted to.

**Why the silent-staleness failure mode is the dangerous one.** I graphed a 6,130-file Python repo (122,114 nodes / 299,925 edges, 83s). Queries against it are confident and well-formed. Nothing in the output distinguishes a graph built 30 seconds ago from one built last month against since-deleted files — no warning, no metadata, no exit-code difference. A wrong answer that looks identical to a right one is worse than an error, because the agent consuming it has no signal to fall back to source.

As a stopgap I check mtime externally before trusting any query:

```python
g_mtime = Path("graphify-out/graph.json").stat().st_mtime
newer = [p for p in repo.rglob("*") if p.suffix in CODE_EXTS and p.stat().st_mtime > g_mtime]
```

That works, but it is strictly worse than what #1624 does: mtime is not commit time, it false-positives on a bare `touch` (verified), and every consumer has to reimplement the walk. A stamp inside the artifact is the right place for this — `generated_at` written at the single `export.to_json` chokepoint is exactly the minimal fix.

One observation on the check itself, offered in case it is useful: comparing against *last commit time* misses uncommitted working-tree edits, which is the common case during active development — the graph is stale the moment you edit a file, not when you commit it. A cheap addition would be to also compare against the newest mtime among indexed files (the file list is already in the graph), so dirty-tree staleness is caught too. Not a blocker for this PR — the commit-time check alone covers the reported case.

Environment: graphifyy 0.9.33 via `uv tool install`, Linux (WSL2), Python 3.11.

PR #1644, opened 2026-07-04
fix(kotlin): emit `implements` edge for interface delegation (`by`)

Synvoya:
## What it fixes

Kotlin interface delegation — `class Foo : Bar by baz` — emitted **no** edge, so the delegated interface `Bar` was silently dropped from the graph.

The `delegation_specifier` loop only handled `constructor_invocation` (→ `inherits`) and a bare `user_type` (→ `implements`) child. The `by` form produces an `explicit_delegation` node that *wraps* the `user_type`, so neither branch fired and no edge was emitted. Delegation is idiomatic Kotlin (the whole point of `by`).

## Fix

Add an `explicit_delegation` branch in that loop: take its wrapped `user_type` child and emit the `implements` edge, leaving the existing generic type-argument recovery intact.

```kotlin
class LoggingList<T>(inner: MutableList<T>) : MutableList<T> by inner
```
now emits `implements: LoggingList -> MutableList` (and `references: LoggingList -> T` for the generic arg).

## Test

- `tests/fixtures/sample.kt`: added the delegation example above.
- `tests/test_languages.py`: `test_kotlin_interface_delegation_emits_implements` asserts the `implements` edge.
- `pytest tests/test_languages.py -k kotlin` → 9 passed (negative control without the fix fails on the missing edge).

### DIFF
--- a/graphify/extract.py
+++ b/graphify/extract.py
@@ -3648,6 +3648,17 @@ def _extract_generic(
                             if sub.type == "user_type":
                                 user_type_node = sub
                                 break
+                            # `class Foo : Bar by baz` wraps the delegated
+                            # interface `Bar` in an `explicit_delegation`
+                            # node; grab its first `user_type` descendant so
+                            # the implements edge (and generic-arg recovery)
+                            # still fire.
+                            if sub.type == "explicit_delegation":
+                                for inner in sub.children:
+                                    if inner.type == "user_type":
+                                        user_type_node = inner
+                                        break
+                                break
                         if user_type_node is None:
                             continue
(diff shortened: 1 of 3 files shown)


safishamsi:
Merged into v8 as `9b04022`, with your authorship preserved (ships in 0.9.6). Thanks @Synvoya.

Nice catch on the `explicit_delegation` wrapper — the `by` form fell between the `constructor_invocation` and bare `user_type` branches so the edge was dropped. Unwrapping to the inner `user_type` keeps generic-argument recovery working, and `implements` (not `inherits`) is the right relation for delegation. Fixture + test made review easy.

PR #1645, opened 2026-07-04
fix(apex): emit `extends` edges for interface multiple inheritance

Synvoya:
## What it fixes

An Apex `interface X extends A, B` emitted only a `contains` edge — the parent interface list was dropped from the graph.

`iface_re` captures the `extends` parent list in group 2, but the interface handler read only group 1 (the interface name); group 2 was never used. (The class branch already splits its parent list and emits an edge per parent.)

## Fix

After the interface node / `contains` edge, iterate the captured parent list and emit one `extends` edge per parent, mirroring the class branch's node resolution (`extends` matches the keyword Apex interfaces use for inheritance; the class branch already emits `extends`). Apex interfaces support multiple inheritance.

```apex
public interface PaymentProcessor extends Processor, Auditable { void process(); }
```
now emits:
```
extends: PaymentProcessor -> Processor
extends: PaymentProcessor -> Auditable
```

## Test

- `tests/test_languages.py`: `test_apex_interface_extends` asserts both parents appear as inheritance edges.
- `pytest tests/test_languages.py -k apex` → 13 passed (negative control without the fix fails with an empty inheritance set).

### DIFF
--- a/graphify/extract.py
+++ b/graphify/extract.py
@@ -6297,6 +6297,16 @@ def extract_apex(path: Path) -> dict:
             add_node(iface_nid, iface_name, lineno)
             add_edge(file_nid if current_class_nid is None else current_class_nid,
                      iface_nid, "contains", lineno)
+            if im.group(2):
+                for parent in im.group(2).split(","):
+                    parent = parent.strip()
+                    if parent:
+                        parent_nid = _make_id(stem, parent)
+                        if parent_nid not in seen_ids:
+                            parent_nid = _make_id(parent)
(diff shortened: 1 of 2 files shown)


safishamsi:
Merged into v8 as `53c769d`, with your authorship preserved (ships in 0.9.6). Thanks @Synvoya.

Clean fix — the interface handler was reading only group 1 while group 2 already captured the parent list, and your addition mirrors the class branch's parent resolution exactly (same stem/bare `_make_id` fallback and `INFERRED` confidence). The negative-control note on the test was appreciated.

Issue #1226, opened 2026-06-10
detect() creates duplicate converted .md files on macOS due to NFC/NFD Unicode path normalization inconsistency, causing --update to re-extract all   Office files on every run


hellix98:
---
  Summary

  On macOS, running /graphify <path> --update repeatedly re-extracts already-processed Office files (.docx, .xlsx) even when no content has changed. Each
  run creates new .md sidecar files with different hash suffixes in <path>/graphify-out/converted/, making --update treat them as new files and triggering a
  full re-extraction every time.

  ---
  Environment

  - graphifyy version: 0.5.4
  - Python: 3.13.9
  - OS: macOS 15.6.1 Sequoia, APFS filesystem

  ---
  Steps to Reproduce

  # 1. Initial build with a folder containing .docx files
  /graphify raw

  # 2. Run --update without changing any files
  /graphify raw --update

  # 3. Run --update again
  /graphify raw --update

  Expected: Step 3 prints "No files changed since last run. Nothing to update."
  Actual: Every run treats existing .docx files as new and re-extracts all of them.

  ---
  Root Cause

  The issue is in convert_office_file() in graphify/detect.py:

  # current code
  name_hash = hashlib.sha256(str(path.resolve()).encode()).hexdigest()[:8]
  out_path = out_dir / f"{path.stem}_{name_hash}.md"

  On macOS (HFS+/APFS), path.resolve() returns paths in different Unicode normalization forms depending on how the Path object was constructed:

┌──────────────────────┬───────── ┬───────┐
 │             How the path was created                          │     Unicode form         │ Example hash    │
├──────────────────────┼──────────┬───────┤
 │ Via root.rglob('*') — OS filesystem enumeration  │ NFD (macOS default)  │ df70b1b3         │
├──────────────────────┼──────────┼───────┤
 │ Via Path('/absolute/path') — direct construction  │ NFC (Python default) │ a54fe208          │
└──────────────────────┴──────────┴───────┘

  detect() enumerates files via rglob(), so it uses NFD paths. However, the normalization form of str(path.resolve()) can vary between runs depending on the
  Python version, interpreter context, or OS state. The same .docx file can therefore produce a different hash on every run.

  Observed result — duplicate sidecars accumulate:

  raw/graphify-out/converted/
    report_b2b2ed3e.md   ← run 1
    report_df70b1b3.md   ← run 2  (same source file, different hash)
    report_a54fe208.md   ← run 3  (same source file, yet another hash)

[...]

PR #1660, opened 2026-07-04
fix(detect): detect Office source edits and stop re-parsing unchanged files (#1649, #1656)

### LINKED ISSUE #1649: Office files (.docx/.xlsx): modified sources never re-enter --update (sidecar keyed on path hash, source content never checked)
**Version:** graphifyy 0.9.5 (Windows 11, docs-heavy corpus: ~6800 nodes, 76% documents, ~800 files incl. many .docx/.xlsx)

## Problem

A `.docx`/`.xlsx` that is **modified after its first conversion never re-enters an `--update`**.

- `convert_office_file` (detect.py:604) names the sidecar `{stem}_{sha256(NFC path)[:8]}.md` — a hash of the **path**, not the content — and early-returns if the sidecar already exists (detect.py:636) without ever looking at the source file.
- `detect_incremental` (detect.py:1389+) compares mtime/hash of the **sidecar**, never of the Office source.

## Repro

1. Convert a .docx (first extract) → sidecar created.
2. Edit the .docx.
3. Run `detect_incremental` → file reported "unchanged"; the stale sidecar is what gets (not) re-extracted. Forever.

On our corpus this forces a manual pre-update pass that purges stale sidecars by comparing `source.st_mtime > sidecar.st_mtime` — workable, but any user with a living Office corpus silently gets a frozen graph otherwise.

## Proposal

Store the md5 of the SOURCE file in the sidecar (header comment) or in the manifest entry, and re-convert when it differs. Minimal alternative: regenerate the sidecar when `source.st_mtime > sidecar.st_mtime`.

Related (closed, different angles): #1226 (NFC/NFD duplicate sidecars), #861 (.graphifyignore bypass for sidecars).

### LINKED ISSUE #1656: Perf: cache word_count in manifest — detect_incremental re-parses every PDF/docx on each run
**Version:** graphifyy 0.9.5, corpus of ~800 files (many PDF/docx).

## Problem

`detect_incremental` calls the full `detect()`, which re-parses **all** PDFs/docx just to `count_words` (detect.py:645) on every invocation — ~3 minutes on our corpus when only 5 files actually changed. The expensive part isn't hashing, it's re-opening every Office/PDF binary for a word count that hasn't changed.

## Proposal

Store `word_count` in the manifest entry and recount only files whose hash changed. Turns incremental detection from O(corpus) parsing into O(changed).

TPAteeq:
Fixes #1649
Fixes #1656

## Summary

`#1649` and `#1656` both live in `convert_office_file` and pull in opposite
directions, but a single change — a **source-content fingerprint** — resolves both.

- **#1649 (Office edits never re-enter `--update`):** the sidecar name was keyed on a
  hash of the source *path*, never its *content*, and an early-return on
  `out_path.exists()` (the #1226 no-churn fix) meant a modified `.docx`/`.xlsx` was
  never re-converted. The graph silently froze on the stale version, with no error.
- **#1656 (re-parses every file every run):** `convert_office_file` fully *parsed* the
  binary before that early-return, so Office files were re-parsed on every run even when
  unchanged; PDFs were likewise re-parsed by `count_words` → `extract_pdf_text` on every
  run, with no caching.

## The single-fingerprint approach

`convert_office_file` now fingerprints the **source** by its **raw bytes** (`md5`).
Hashing the raw bytes does *not* unzip/parse the OOXML container, so it stays cheap —
that is the whole point. The fingerprint is recorded in the sidecar header:

```
<!-- converted from report.docx | source-md5: 3b1e… -->
```

On a later run:

- **Sidecar exists and the stored fingerprint matches the source** → return it
  **without parsing** (satisfies #1656) and **without rewriting** (so an unchanged
  source never churns its mtime, preserving the #1226 guarantee).
- **Sidecar missing, or the fingerprint differs (or is absent — a legacy sidecar)** →
  re-parse and rewrite. The rewritten sidecar's new mtime/md5 then flows through
  `detect_incremental` automatically, so the edit re-enters `--update` (satisfies #1649).
  No other file needs to change for change-detection to pick it up.

This threads the needle that a naïve "always re-convert" would miss: it neither freezes
edits (#1649) nor re-parses unchanged sources (#1656/#1226).

## PDF / word-count re-parse (#1656)

Word counts are now **cached in the manifest entry** and reused for unchanged files, so
PDFs are no longer re-parsed for their word count on every incremental run:

- `save_manifest` seeds/refreshes each entry's `word_count`, **reusing the previous
  value whenever the content hash is unchanged** and only (re)parsing a genuinely
  new/changed file (video files are skipped, mirroring `detect()`).
- `detect()` accepts the previous run's per-file counts and reuses the cached count for
  any file whose mtime is unchanged instead of re-parsing.
- `detect_incremental` loads the manifest first, hands those cached counts down to
  `detect()`, and proceeds as before.

`total_words` stays correct in every path (unchanged files contribute their cached
count; changed files are recomputed), so the `benchmark` command's `corpus_words`
keeps working — including via the skill's `--update` flow, which propagates
`total_words` into `.graphify_detect.json`.

Backward compatible: legacy manifests without `word_count` load unchanged (the existing
`_normalise_entry` dict passthrough preserves the new field, and old entries simply
recompute once on the next run).

## Tests

Added to `tests/test_detect.py`:

- Edited vs. unchanged `.docx` **and** `.xlsx` via `convert_office_file` — asserts the
  converter re-parses after an edit and is *not* re-invoked when the source is unchanged.
- Legacy (pre-fingerprint) sidecar is upgraded once, then reused.
- End-to-end: an edited Office source re-enters `detect_incremental().new_files`, while
  an unchanged one stays in `unchanged_files`.
- An unchanged PDF is **not** re-parsed (`extract_pdf_text` not called again) on a second
  incremental run.

## Test results

```
uv run pytest tests/test_detect.py tests/test_incremental.py tests/test_office_limits.py tests/test_manifest_ingest.py
# 153 passed
```

Full suite is green except one pre-existing, environment-only failure
(`test_extract.py::test_collect_files_skips_hidden`) caused solely by running inside a
`.claude/worktrees/…` checkout — it fails identically on the base commit with this
branch's changes stashed and is unrelated to this diff.


safishamsi:
Thanks @TPAteeq, and sorry for the overlap. Both halves of this (Office source-edit detection for #1649 and skipping the re-parse of unchanged files for #1656) already landed on v8 in `f917494`: `convert_office_file` re-converts when the source is newer than its sidecar, and word counts are memoized in the content-hash stat index so unchanged PDFs/docx are not re-parsed. Closing as already-resolved, but the approach here matched ours and was appreciated.

Issue #1657, opened 2026-07-04
Minor Windows/report papercuts: mojibake in manifest paths, stdout encoding, Import Cycles section on doc corpora


Ns2384-star:
**Version:** graphifyy 0.9.5, Windows 11. Three small independent issues, grouped to avoid noise — happy to split on request.

## 1. Mojibake in manifest keys and GRAPH_REPORT for accented Windows paths

Accented file paths end up as `Confirm�`, `ExpÃ©rience` in manifest.json keys and GRAPH_REPORT.md. Path encoding should be normalized (UTF-8) at write time. Same class as #904/#1505 (closed) but on the file-path side.

## 2. `query` needs manual `PYTHONIOENCODING=utf-8` on Windows

Without it, query output crashes/mangles on non-ASCII (cp1252 console). One-liner: `sys.stdout.reconfigure(encoding="utf-8")` in the CLI entry point.

## 3. "Import Cycles" report section is meaningless for documentary corpora

On a 76%-documents corpus the section is pure noise. Condition it on the presence of code nodes.

safishamsi:
Thanks for grouping these — all three are addressed as of `54825b6` (on v8, ships in 0.9.7).

3. **Import Cycles on doc corpora** — fixed: the section is now emitted only when the graph contains code (code nodes or import edges), so a documents-only corpus no longer shows the noise.

1. **Mojibake in manifest/report** and 2. **stdout encoding** — these are already handled on the current branch (they postdate 0.9.5): `manifest.json` and `GRAPH_REPORT.md` are both written with `encoding="utf-8"`, and the CLI `main()` reconfigures stdout/stderr to UTF-8 with `errors="replace"` (the #992 fix). If you still see mojibake in the manifest keys on 0.9.7, it would most likely be a viewer opening a UTF-8 file as cp1252 rather than a write-time bug — please reopen with a sample path and we'll dig in.

PR #1670, opened 2026-07-04
fix(#1657): resolve three independent Windows papercuts

raman118:
###  1. The Accented Path Mojibake Massacre

  • The Annoyance: If you had files named with beautiful accents (like  Confirmé.md  or  Expérience.py ), Windows would choke on them. They’d get written
  into  manifest.json  and  GRAPH_REPORT.md  looking like total garbage ( Confirm  or  ExpÃ©rience ).
  • The Root Cause: Standard Python  Path.read_text()  and  .write_text()  calls without explicit encoding fall back to the system locale default. On
  Windows, that means  cp1252  (Windows-1252). When UTF-8 characters were passed through, Windows decoded them using  cp1252 , causing a classic encoding
  leak.
  • The Fix: We went into  load_manifest  and  save_manifest  inside  graphify/detect.py  and wrapped the read/write boundaries in clean  open(...,
  encoding="utf-8")  blocks. Crucially, we also added  ensure_ascii=False  to  json.dump() , forcing the JSON writer to save raw UTF-8 characters to the
  file instead of converting them into escaped ASCII sequences ( \u00e9 ).
  • The Safety Net: We added  test_manifest_and_report_preserve_accented_paths  to check that accented names round-trip perfectly byte-for-byte in both
  the manifest and the generated markdown report.
 
  ###  2. The Windows Console Encoding Meltdown

  • The Annoyance: Running  graphify query  on Windows without manually setting  PYTHONIOENCODING=utf-8  would lead to ugly console crashes (
  UnicodeEncodeError  /  UnicodeDecodeError ) whenever non-ASCII characters were returned in the output.
  • The Root Cause: The default Windows command console doesn’t use UTF-8 by default. When Python tried to push non-ASCII characters to standard output,
  it threw a fit.
  • The Fix: At the absolute top of the CLI entry point ( graphify/__main__.py ), before any output gets printed, we added a reconfiguration guard:
    if hasattr(sys.stdout, "reconfigure"):
        sys.stdout.reconfigure(encoding="utf-8", errors="replace")
    This forces both  stdout  and  stderr  to speak UTF-8, using replacement characters if something is truly mangled, without breaking piped/redirected
  output.
  • The Safety Net: We added  test_query_cli_non_ascii_no_encoding_error  in  test_query_cli.py . It spawns a subprocess running the query CLI without
  PYTHONIOENCODING  to simulate the raw Windows environment, proving it no longer crashes.
  
  ###  3. The Import Cycles Noise Pollution

  • The Annoyance: When running graphify on a corpus composed mostly of documents (like a folder full of markdown papers), the report would print a
  section called  ## Import Cycles . Because import cycles are a code-only concept, this was pure noise, usually returning "None detected" or parsing
  markdown paths incorrectly.
  • The Root Cause: The report generator was blindly appending the Import Cycles section without checking whether the corpus actually contained code.
  • The Fix: In  graphify/report.py , we added a quick scan to count the number of nodes in the graph that represent code. If  has_code_nodes  is false,
  we skip building the  ## Import Cycles  section entirely.
  • The Safety Net: We wrote two new tests:  test_report_omits_import_cycles_on_document_only_corpus  and
  test_report_includes_import_cycles_on_code_heavy_corpus . Now we explicitly assert that the cycles section is omitted on markdown/document-only trees,
  while remaining perfectly intact on code-heavy repos.

### DIFF
--- a/graphify/__main__.py
+++ b/graphify/__main__.py
@@ -1,13 +1,24 @@
 """graphify CLI - `graphify install` sets up the Claude Code skill."""
 
 from __future__ import annotations
+import sys
+if hasattr(sys.stdout, "reconfigure"):
+    try:
+        sys.stdout.reconfigure(encoding="utf-8", errors="replace")
+    except Exception:
+        pass
+if hasattr(sys.stderr, "reconfigure"):
+    try:
+        sys.stderr.reconfigure(encoding="utf-8", errors="replace")
+    except Exception:
+        pass
+
 import functools
 import json
 import os
 import platform
 import re
 import shutil
-import sys
 from pathlib import Path
 
 try:
--- a/graphify/detect.py
+++ b/graphify/detect.py
@@ -1266,7 +1266,8 @@ def load_manifest(
     to :func:`save_manifest`) remains readable.
     """
     try:
-        raw = json.loads(Path(manifest_path).read_text(encoding="utf-8"))
+        with open(manifest_path, "r", encoding="utf-8") as f:
+            raw = json.load(f)
     except Exception:
         return {}
     if root is None or not isinstance(raw, dict):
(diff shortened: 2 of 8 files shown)


safishamsi:
Thanks @raman118. The Import Cycles gate for documents-only corpora landed on v8 in `54825b6` (#1657), and the other two items in that issue were already in place: manifest.json and GRAPH_REPORT.md are written with `encoding="utf-8"`, and `main()` already reconfigures stdout/stderr to UTF-8 with `errors="replace"`. So this is fully covered on v8 already. Closing as already-resolved.

PR #1674, opened 2026-07-04
Fix case-sensitive file suffix filtering silently skipping capitalized/mixed-case extensions (#1671)

### LINKED ISSUE #1671: [Bug] Case-sensitive file suffix filtering in collect_files() silently skips capitalized/mixed-case extensions
**Version:** graphifyy 0.9.6, Windows / Cross-platform.

### Description
In `graphify/extract.py`, the new `collect_files()` implementation uses `p.suffix in _EXTENSIONS` to check if a file should be collected for extraction.
Because `_EXTENSIONS` contains lowercase suffixes (like `.py`, `.js`), any file with a capitalized or mixed-case extension (such as `app.PY` or `index.JS`) is silently skipped.

On case-insensitive filesystems (such as Windows and macOS), files with capitalized extensions are valid and should be collected and processed. Previously, the legacy `_legacy_collect_files` implementation used `target.rglob(f"*{ext}")` which matched these extensions case-insensitively on Windows. The optimization in #1261 introduced this regression.

### Reproduction
Create a repository with a Python file named `app.PY` (capitalized extension) and run `collect_files` or run `graphify extract .`.

```python
import tempfile
from pathlib import Path

with tempfile.TemporaryDirectory() as tmpdir:
    tmp_path = Path(tmpdir)
    py_file = tmp_path / "app.PY"
    py_file.write_text("print('hello')", encoding="utf-8")
    
    collected = collect_files(tmp_path)
    print("Collected:", collected) # Returns [] instead of [app.PY]
```

### Expected Behavior
Files with capitalized or mixed-case extensions (like `.PY` or `.JS`) should be collected and parsed.

### Root Cause & Location
In [graphify/extract.py](file:///C:/Users/rcho3/Desktop/open-source/graphify/graphify/extract.py#L16617-L16637):
```python
if p.suffix in _EXTENSIONS and not _ignored(p) and _resolves_under_root(p, containment_root):
```
This performs a case-sensitive check against `_EXTENSIONS`. Suffix checks should be case-insensitive (e.g., using `p.suffix.lower()`).

raman118:
This pull request fixes an issue where files with capitalized or mixed-case extensions (such as `.PY`, `.JS`, or `.TS`) were silently skipped during file discovery.

#### Problem
In extract.py, the collect_files function checks file suffixes against _EXTENSIONS (derived from the keys of the _DISPATCH mapping). Because _DISPATCH only maps lowercase extensions for most languages, files with uppercase or mixed-case extensions were excluded from discovery. Additionally, _get_extractor would fail to resolve an extractor function for such files even if they were manually passed to it.

#### Solution
- Modified collect_files to match files if their exact suffix or their lowercase suffix is in _EXTENSIONS.
- Modified _get_extractor to check _DISPATCH using a lowercase fallback if the exact suffix casing is not registered. This ensures that language-specific case distinctions (such as Fortran .f vs .F) are still respected, while typical languages (like Python, JavaScript, and TypeScript) support uppercase extensions. Also updated the .blade.php check to be case-insensitive.
- Modified _legacy_collect_files in test_extract.py to enforce case-sensitivity on p.suffix == ext to prevent duplicate matching on Windows.
- Added a new regression test test_case_insensitive_suffix_filtering to verify that uppercase extensions (.PY, .JS, .Ts) are discovered, mapped to their correct extractors, and successfully parsed.

Fixes #1671

### DIFF
--- a/graphify/extract.py
+++ b/graphify/extract.py
@@ -15839,7 +15839,7 @@ def _is_cpp_header(path: Path) -> bool:
 
 def _get_extractor(path: Path) -> Any | None:
     """Return the correct extractor function for a file, or None if unsupported."""
-    if path.name.endswith(".blade.php"):
+    if path.name.lower().endswith(".blade.php"):
         return extract_blade
     # MCP config files (.mcp.json, claude_desktop_config.json, ...) are routed
     # by filename before generic .json dispatch so they get MCP-aware nodes
@@ -15855,14 +15855,17 @@ def _get_extractor(path: Path) -> Any | None:
     # (the suffix map sends `.h` to extract_c, which can't read @interface etc.).
     # ObjC sniffing has priority over the C++ sniff: an Objective-C++ header can
     # contain both `@interface` and inline C++ (`::`), and it must parse as ObjC.
-    if path.suffix == ".h":
+    suffix = path.suffix
+    if suffix not in _DISPATCH and suffix.lower() in _DISPATCH:
+        suffix = suffix.lower()
+    if suffix == ".h":
         if _is_objc_header(path):
             return extract_objc
         # A C++ class header routed to extract_c loses the class entirely (the C
         # grammar has no class_specifier). Reroute to extract_cpp (#1547).
         if _is_cpp_header(path):
             return extract_cpp
-    return _DISPATCH.get(path.suffix)
+    return _DISPATCH.get(suffix)
 
 
 def _safe_extract_with_xaml_root(extractor, path: Path, root: Path) -> dict:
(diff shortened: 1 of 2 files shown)


safishamsi:
Merged into v8 as `aa1bbda`, authorship preserved (ships in 0.9.7). Thanks @raman118. Suffix matching now falls back to the lowercased form in both `collect_files` and `_get_extractor`, so `App.PY`/`script.JS`/`Lib.Ts` are picked up while unsupported extensions still skip. Verified mixed-case dispatch (.PY/.JS/.Ts/.RB/.PHP) and that .xyz is excluded.

Issue #1672, opened 2026-07-04
[Bug] introspect_postgres() mangles virtual URI path with backslashes on Windows


raman118:
**Version:** graphifyy 0.9.6, Windows.

### Description
In `graphify/pg_introspect.py`, the `introspect_postgres()` function connects to a PostgreSQL database, retrieves the schema metadata, and formats it as in-memory SQL DDL, which is then parsed by `extract_sql()`.

To serve as a source_file identifier, `introspect_postgres()` instantiates a virtual URI path using `Path`:
```python
virtual_path = Path(f"postgresql://{host}/{dbname}")
```
On Windows, `Path` parses URI strings as filesystem paths and normalizes them using backslashes (`postgresql:\myhost\mydb`).
In `graphify/extract.py`, `extract_sql(path: Path)` converts this to a string using `str(path)` (retaining backslashes) and populates the `source_file` attribute on nodes and edges with this mangled string.

This causes:
1. Inconsistent and invalid `source_file` values (e.g., `postgresql:\myhost\mydb` instead of `postgresql:/myhost/mydb`).
2. Test failures in `tests/test_pg_introspect.py::test_pg_introspect_success` on Windows due to the backslashes.

### Reproduction
On Windows, run the test suite for pg_introspect:
```bash
python -m pytest tests/test_pg_introspect.py
```
This fails with:
```
E           AssertionError: assert 'postgresql:\\myhost\\mydb' == 'postgresql:/myhost/mydb'
E             
E             - postgresql:/myhost/mydb
E             ?            ^      ^
E             + postgresql:\myhost\mydb
E             ?            ^      ^
```

### Expected Behavior
The URI path `postgresql://host/dbname` should always use forward slashes (`/`), even on Windows.

### Root Cause & Location
1. In [graphify/pg_introspect.py](file:///C:/Users/rcho3/Desktop/open-source/graphify/graphify/pg_introspect.py#L138):
```python
virtual_path = Path(f"postgresql://{host}/{dbname}")
```
Instantiating `Path` on a URI string is incorrect on Windows.

2. In [graphify/extract.py](file:///C:/Users/rcho3/Desktop/open-source/graphify/graphify/extract.py#L7366):
```python
str_path = str(path)
```
This should use `path.as_posix()` if `path` is a Path instance, or replace backslashes with forward slashes for URI/POSIX normalization.

safishamsi:
[...]

PR #1675, opened 2026-07-04
Fix virtual Postgres source_file URI backslash normalization on Windows (#1672)

raman118:
### Summary
This PR resolves issue #1672, where `introspect_postgres()` produces invalid `source_file` URIs on Windows (e.g., `postgresql:\myhost\mydb` instead of `postgresql:/myhost/mydb`). 

### Cause
In `graphify/pg_introspect.py`, the `virtual_path` was constructed using `pathlib.Path(f"postgresql://{host}/{dbname}")`. On Windows, this instantiates a `WindowsPath` object, which normalizes the path separators to backslashes (`\`). Because the path is a synthetic URI rather than a local file path, it should be OS-independent and retain forward slashes.

### Solution
- Imported and utilized `pathlib.PurePosixPath` to construct `virtual_path`. Since `PurePosixPath` uses POSIX path semantics regardless of the host operating system, the path string uses forward slashes on all platforms.
- Added the regression test `test_pg_introspect_uri_forward_slashes` in `tests/test_pg_introspect.py` to assert that the generated virtual paths always contain forward slashes and no backslashes, even when run on Windows.
- Successfully verified that all postgresql introspection tests pass on Windows.
- Executed `graphify update .` to rebuild the AST knowledge graph.

### DIFF
--- a/graphify/pg_introspect.py
+++ b/graphify/pg_introspect.py
@@ -1,5 +1,5 @@
 from __future__ import annotations
-from pathlib import Path
+from pathlib import Path, PurePosixPath
 from graphify.extract import extract_sql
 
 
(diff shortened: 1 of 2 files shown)


safishamsi:
Merged into v8 as `5ffa921`, authorship preserved (ships in 0.9.7). Thanks @raman118. `PurePosixPath` keeps the virtual `postgresql://host/db` URI forward-slashed on Windows; verified no backslashes in the emitted `source_file`.

PR #1679, opened 2026-07-05
fix(extract): don't report deferred import() as a file cycle

### LINKED ISSUE #1241: Typescript, deferred imports incorrectly detected as cycles
Versions tested: 0.8.35, 0.8.37

```tsx
// modal.ts
import {hideModal} from './actions';

export function Modal(): void {
  hideModal();
}
```
```tsx
// actions.ts
export function showModal(): void {
  void import('./modal').then((m) => m.Modal());
}

export function hideModal(): void {}
```

```python
from graphify.analyze import find_import_cycles

def test_deferred_import_is_not_a_cycle():
    result = extract(
        ["modal.ts", "actions.ts"]
    )
    graph = build_from_json(result)
    cycles = find_import_cycles(graph)

    assert cycles == [], (
        "deferred import() was counted as a static import edge and reported "
        f"as a circular dependency: {cycles}"
    )

test_deferred_import_is_not_a_cycle()
```

I suspect this kind of pattern is what led to the cycling in https://github.com/safishamsi/graphify/issues/1196, so it might be worth keeping track of whether the import is deferred while computing cycles.

Synvoya:
## What it fixes

A deferred `import('./x')` was emitted by `_dynamic_import_js` (`graphify/extract.py`) as a plain `imports_from` edge — indistinguishable from a static import. `find_import_cycles` (`graphify/analyze.py`) counts every `imports_from` / `re_exports` edge as a hard file-cycle edge, so two files that reference each other via **one static import + one dynamic `import()`** were reported as a circular dependency (#1241).

## Fix

Keep the deferred edge as `imports_from` — the dependency stays visible in the graph — but tag it `"deferred": True`, and skip deferred edges in `find_import_cycles`.

This is deliberately *not* a retag to `dynamic_import`: the deferred edge's source is a symbol node (the enclosing function) and its target isn't a materialized file node, so the valid-id prune passes (`extract.py:5380` and the sibling passes) only spare it while its relation is `imports_from` / `re_exports`. Retagging would drop the edge entirely and lose the dependency; the `deferred` marker keeps it.

```ts
// actions.ts
export async function lazy() { const m = await import("./modal"); return m.openModal(); }
// modal.ts
import { doThing } from "./actions";
```

- **Before:** `find_import_cycles` → `[{'cycle': ['actions.ts', 'modal.ts'], 'length': 2, ...}]`
- **After:** `[]`, and the deferred dependency edge is still present (`imports_from`, `deferred=True`).

## Test

`tests/test_js_import_resolution.py::test_ts_dynamic_import_does_not_create_phantom_cycle` asserts the deferred edge stays (`imports_from` + `deferred`), the real static import is unaffected, and `find_import_cycles` returns no cycle. Full `test_js_import_resolution.py` + `test_analyze.py` pass (99 passed); reverting the fix fails the new test (negative control).

Closes #1241

### DIFF
--- a/graphify/analyze.py
+++ b/graphify/analyze.py
@@ -663,6 +663,11 @@ def find_import_cycles(
         if rel not in ("imports_from", "re_exports"):
             continue
 
+        # Deferred `import(...)` edges are real dependencies but do not form a
+        # hard file-level cycle, so they are excluded from cycle detection (#1241).
+        if data.get("deferred"):
+            continue
+
         src_file_attr = data.get("source_file", "")
         if not isinstance(src_file_attr, str) or not src_file_attr:
             continue
--- a/graphify/extract.py
+++ b/graphify/extract.py
@@ -1922,8 +1922,13 @@ def _dynamic_import_js(node, source: bytes, caller_nid: str, str_path: str, edge
             edges.append({
                 "source": caller_nid,
                 "target": tgt_nid,
+                # A deferred `import(...)` is a real dependency, so keep it as an
+                # `imports_from` edge (visible in the graph) but mark it `deferred`
+                # so find_import_cycles does not treat it as a static import and
+                # report a phantom file cycle (#1241).
                 "relation": "imports_from",
                 "context": "import",
+                "deferred": True,
                 "confidence": "EXTRACTED",
                 "source_file": str_path,
                 "source_location": f"L{node.start_point[0] + 1}",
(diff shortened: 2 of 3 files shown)


safishamsi:
Merged into v8 as `94392de`, authorship preserved (ships in 0.9.7). Thanks @Synvoya. Nice fix. Verified the deferred edge stays in the graph and, importantly, that a genuine static A<->B cycle is still reported (no over-suppression).

PR #1680, opened 2026-07-05
fix(extract): route extensionless shebang scripts to their AST extractor

Stashub:
## Problem

`detect.classify_file` already labels extensionless files with a bash/python/node/... shebang as CODE (via `_shebang_interpreter`), but `extract._get_extractor` dispatches purely on `path.suffix`. An extensionless CLI entry point — `devctl`, `manage`, `gradlew`-style wrappers — is therefore detected as code and then **silently contributes zero nodes** to the graph, and its doc-referenced symbols stay dangling stub IDs.

Found in the wild: a bash CLI `devctl` (`#!/usr/bin/env bash`, no extension) — the main executable of the corpus — was missing from the graph entirely while the 5 other code files extracted fine; the health diagnostic showed dangling-endpoint edges pointing at its never-created node IDs.

## Fix

In `_get_extractor`, resolve extensionless files through the same `detect._shebang_interpreter` and a new `_SHEBANG_DISPATCH` map, so extract honors the same signal detect already trusts. Only interpreters with a real extractor are mapped (python/bash-family/node/ruby/lua/php/julia); detect's wider set (perl, fish, tcsh, Rscript) stays unmapped and skipped rather than being mis-parsed by a wrong grammar.

## Tests

- `test_extensionless_shebang_via_dispatch` — bash & python3 shebangs, incl. the `env -S` split-args form
- `test_extensionless_without_usable_shebang_stays_unsupported` — plain text and perl stay `None`
- `test_extract_extensionless_bash_cli_end_to_end` — node IDs follow the path-stem scheme, so doc-created stub IDs merge with the real code nodes

`pytest tests/test_extract.py tests/test_detect.py` — 235 passed.


safishamsi:
Closing in favor of #1683, which is the same fix in a clean 2-file diff. This PR's diff is ~223k lines (it appears to include built artifacts / generated output), so #1683 is the one we merged (as `2ab0867`). Thanks for the fix either way.

PR #1683, opened 2026-07-05
fix(extract): route extensionless shebang scripts to their AST extractor

Stashub:
## Problem

`detect.classify_file` already labels extensionless files with a bash/python/node/… shebang as CODE via `_shebang_interpreter`, but `_get_extractor` dispatches purely on the path suffix. So a CLI entry point like `devctl` or `manage` is detected as code and then **silently contributes zero nodes** to the graph — its doc-referenced symbols stay dangling stubs.

Found while graphing a devops repo whose main artifact is an extensionless bash CLI: the file most referenced by the docs produced no AST nodes at all.

## Fix

Resolve extensionless files through the same `_shebang_interpreter` and a new `_SHEBANG_DISPATCH` map, so extraction honors the same signal as detection.

Only interpreters with a real extractor are mapped (python / bash-family / node / ruby / lua / php / julia). detect's wider set (perl, fish, tcsh, Rscript) stays unmapped and skipped rather than being mis-parsed by a wrong grammar.

## Tests

- `test_extensionless_shebang_via_dispatch` — bash/python/`env -S` shebangs resolve to their extractors
- `test_extensionless_without_usable_shebang_stays_unsupported` — no shebang / no-extractor interpreters (perl) still return `None`
- `test_extract_extensionless_bash_cli_end_to_end` — a shebang-only bash CLI contributes nodes with the same ID scheme as a `.sh` file, so doc-created stub IDs merge

`pytest tests/test_extract.py tests/test_detect.py` — 236 passed.

### DIFF
--- a/graphify/extract.py
+++ b/graphify/extract.py
@@ -15821,6 +15821,30 @@ _DISPATCH: dict[str, Any] = {
 }
 
 
+# Extensionless executables (CLI entry points like `devctl` or `manage`) carry
+# their language in the shebang, not the suffix. detect.classify_file already
+# routes them to the CODE path via _shebang_interpreter; _get_extractor must
+# honor the same signal or these files are classified as code and then silently
+# dropped by extraction. Only interpreters with a real extractor are mapped —
+# detect's wider set (perl, fish, tcsh, Rscript) stays unmapped and skipped.
+_SHEBANG_DISPATCH: dict[str, Any] = {
+    "python": extract_python,
+    "python2": extract_python,
+    "python3": extract_python,
+    "bash": extract_bash,
+    "sh": extract_bash,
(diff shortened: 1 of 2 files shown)


safishamsi:
Merged into v8 as `2ab0867`, authorship preserved (ships in 0.9.7). Thanks @Stashub. Verified end to end: reproduced the bug on v8 (an extensionless `#!/usr/bin/env bash` CLI is classified as code by detect but `_get_extractor` returned None, so it contributed 0 nodes), confirmed the fix routes it via `_SHEBANG_DISPATCH` (now extracts), and checked the edges: `env -S` form resolves, python/node/ruby map correctly, and perl / no-shebang / plain text still return None (skipped, not mis-parsed). Full suite green.

PR #1703, opened 2026-07-06
Fix rebuilds from deleted hook cwd

FranciscoJSBarragan:
## Summary

Fix detached rebuilds that inherit a deleted transient working directory before graphify can create `graphify-out/.pending_changes` or acquire the rebuild lock.

## Problem

Post-commit/post-checkout hook rebuilds can run in a background process after the launcher CWD has disappeared. In that state, relative path setup such as `graphify-out` queue/lock creation can raise:

```text
FileNotFoundError: [Errno 2] No such file or directory
```

This happens before the main rebuild body reaches the existing path/root handling.

## Changes

- Add `_stabilize_rebuild_cwd()` and call it at the start of `_rebuild_code()` before `_queue_pending()` / `_rebuild_lock()`.
- Recover by chdir'ing to `GRAPHIFY_REPO_ROOT` when hooks provide it.
- Return `False` with a clear message if the CWD is gone and no repo-root fallback is available.
- Add regression tests for both the clean-failure and repo-root recovery cases.

## Validation

- `.venv/bin/pytest tests/test_watch.py -q` → `54 passed, 2 skipped`
- `.venv/bin/ruff check graphify/watch.py tests/test_watch.py` → passed
- `.venv/bin/python -m py_compile graphify/watch.py tests/test_watch.py` → passed
- `env -u PYTHONPATH -u PYTHONHOME PYTHONHASHSEED=0 .venv/bin/pytest -q` → `2904 passed, 30 skipped`

Note: running the full suite without `PYTHONHASHSEED=0` exposed an unrelated ordering-sensitive labeling assertion (`[100, 50, 100]` vs `[100, 100, 50]`). With the deterministic hash seed already used by graphify hooks, the full suite passes.

### DIFF
--- a/graphify/watch.py
+++ b/graphify/watch.py
@@ -657,6 +657,37 @@ def _json_text(data: dict) -> str:
     return json.dumps(data, indent=2, ensure_ascii=False) + "\n"
 
 
+def _stabilize_rebuild_cwd(watch_path: Path) -> bool:
+    """Ensure relative rebuild paths have a usable CWD before queue/lock setup.
+
+    Detached git hooks can inherit a transient working directory that is deleted
+    before the background rebuild starts. In that state Path.cwd(),
+    Path('.').resolve(), and relative graphify-out mkdirs raise FileNotFoundError
+    before the normal rebuild error handling can run. Hooks that know the repo
+    root export GRAPHIFY_REPO_ROOT so the rebuild can recover by chdir'ing there.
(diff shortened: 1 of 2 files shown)


safishamsi:
Merged into `v8` (e5044c3), authorship preserved. Reviewed, full suite green (3078 passed). Thanks for the fix!

PR #1707, opened 2026-07-07
fix(extract): disambiguate unresolved base-class stubs by file

mallyskies:
## What's wrong
When a class extends a base class that isn't defined in the same file, graphify creates a placeholder "stub" node for that unknown base class. `ensure_named_node()` — the helper that creates these stubs — tags each one with `origin_file`, so a later pass (`_disambiguate_colliding_node_ids`) can tell "file A's reference to `Base`" apart from "file B's reference to `Base`".
Five places in the extractor build this same kind of stub *inline*, bypassing `ensure_named_node()` entirely — and none of them set `origin_file`. Without that tag, every file's reference to the same-named base class gets merged into one shared stub instead of staying separate. That shared stub can then collide with an unrelated real class somewhere else in the codebase that happens to have the same name.
Affected inheritance handlers:
- C++ `base_class_clause`
- Python class inheritance
- Kotlin delegation-specifier inheritance/conformance
- Ruby `class Foo < Base`
- Ruby `Class.new(Super)`
## The fix
Four of the five sit directly inside `_extract_generic`, where `ensure_named_node()` is already available — those now just call it instead of duplicating its logic.
The fifth (Ruby's `Class.new(Super)`) lives in a separate helper, `_ruby_extra_walk()`, that doesn't have `ensure_named_node()` in scope. Rather than change that helper's signature, it gets the same `origin_file` tag added directly to its own stub dict.
A sixth, similar occurrence exists in `extract_apex()`, a fully separate regex-based extractor with no `ensure_named_node()` of its own. Left out here to keep this PR scoped to the shared `_extract_generic` path and its one directly-related helper.
## How we found this
We hit a concrete case: ~163 different C++ files each had a class extending `mTimer`, a real, shared C++ base class. Because none of those references were disambiguated by file, they all collapsed onto one stub — and that stub's label happened to exact-case-match an unrelated class also named `MTimer` in a different language entirely. Result: 218 wrong `inherits` edges from unrelated C++ server code into that unrelated class.
## Testing
Added a regression test: two different C++ files, each inheriting from the same undefined base class, must produce two distinct stub nodes — not one shared one. It fails on `v8` (both files share one `'base'` id) and passes with this fix.
Full `tests/test_extract.py` suite: 94 passing (was 93; +1 new test), same 19 pre-existing failures before and after this change (unrelated — missing optional tree-sitter grammar packages in this environment).

### DIFF
--- a/graphify/extract.py
+++ b/graphify/extract.py
@@ -3424,10 +3424,16 @@ def _ruby_extra_walk(node, source: bytes, file_nid: str, stem: str, str_path: st
                         if base_nid not in seen_ids:
                             base_nid = _make_id(base)
                             if base_nid not in seen_ids:
+                                # origin_file lets _disambiguate_colliding_node_ids
+                                # tell this file's unresolved reference apart from
+                                # another file's same-named one, instead of every
+                                # file's stub collapsing onto one shared bare id
+                                # (see ensure_named_node(), which sets the same
+                                # field for this exact reason).
                                 nodes.append({
                                     "id": base_nid, "label": base,
                                     "file_type": "code", "source_file": "",
-                                    "source_location": "",
+                                    "source_location": "", "origin_file": str_path,
                                 })
                                 seen_ids.add(base_nid)
                         add_edge(class_nid, base_nid, "inherits", line)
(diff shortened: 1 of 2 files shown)


safishamsi:
Merged into `v8` (9557bf6), authorship preserved. Reviewed, full suite green (3078 passed). Thanks for the fix!

PR #1714, opened 2026-07-07
docs(security): clarify stdio-only claim now that HTTP transport exists

Thizeidler:
## Summary
- `SECURITY.md` states graphify "does not run a network listener" and that the MCP server "communicates over stdio only".
- `README.md` documents an opt-in `--transport http` mode (`python -m graphify.serve ... --transport http --host 0.0.0.0 --api-key ...`) for sharing one graphify server across a team, which does open a real network listener.
- This updates the claim to describe the actual default (stdio, no listener) plus the documented opt-in HTTP exception, so the security doc stays accurate.

## Test plan
- Docs-only change (single bullet in `SECURITY.md`); no code paths affected.
- Confirmed the referenced HTTP transport flags (`--transport`, `--host`, `--api-key`) exist in `README.md` (lines ~443-476).

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -42,7 +42,7 @@ graphify is a **local development tool**. It runs as a Claude Code skill and opt
 
 ### What graphify does NOT do
 
-- Does not run a network listener (MCP server communicates over stdio only)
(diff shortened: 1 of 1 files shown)


safishamsi:
Merged into `v8` (0ce8ea0), authorship preserved. Reviewed, full suite green (3078 passed). Thanks for the fix!

Issue #1659, opened 2026-07-04
TypeScript: unresolved calls bound to same-named exports in unimported packages → phantom cross-package `calls` edges (conf 0.8)


leonaburime-ucla:
## Summary

graphify's call resolver attributes an **unresolved** call to a same-named `export` in an **unrelated package that was never imported**, producing a phantom cross-package `calls` edge at `confidence 0.8`. On a monorepo this fabricates dependency relationships that don't exist.

## Minimal repro

Two packages, no import between them:

`pkg-a/src/index.ts` (imports nothing, defines `validate` nowhere):
```ts
declare function validate(x: number): boolean; // ambient/unresolved locally
export function run(x: number): boolean {
  return validate(x); // pkg-b is never imported
}
```

`pkg-b/src/index.ts`:
```ts
export function validate(name: string): boolean { return name.length > 0; }
```

Run:
```
graphify update <dir> --no-cluster
```

**Observed:** a cross-package edge `pkg-a.run() --calls--> pkg-b.validate()` (confidence 0.8).
**Expected:** no cross-package edge — `pkg-a` has no import path to `pkg-b`, so the call cannot resolve there. It should stay unresolved (or ambient) rather than being bound to an unrelated package's export by name alone.

## Real-world impact

On a ~14-package TypeScript monorepo (Apache-2.0, `open-design`), this produced edges like `platform → registry-protocol` and `sidecar → registry-protocol` even though `platform` and `sidecar` import **nothing** (verified against `package.json` and every `import` statement). `registry-protocol` exports many generically-named symbols (`*Schema`, etc.), so ambiguous/unresolved calls across the repo collapsed onto it — it appeared "depended on" by 14 packages spuriously. A reader trusting the graph would conclude a dependency direction that does not exist; only cross-checking against source caught it.

## Likely root cause

When a call has no local definition and no matching import, the resolver appears to fall back to any same-named exported symbol repo-wide, and emits the edge at 0.8 rather than leaving it unresolved. The fallback isn't gated on an actual import/module-resolution path existing between the two files.

## Suggestion

Gate cross-file/cross-package `calls` attribution on there being a real import (or module-resolution) path from caller to callee; when the only candidate lives in a package with no such path, leave the call unresolved (or emit it as `ambiguous`/`inferred` at markedly lower confidence) instead of a 0.8 `calls` edge. A `diagnose`-style warning for "cross-package edge with no supporting import" would also help users spot these.

## Related
[...]

PR #1718, opened 2026-07-07
fix(extract): never bind a cross-file call to a definition in another language family

edinaldoof:
## Problem

The cross-file call resolver matches raw-call callees against a repo-wide label index with **no language check**. In a repo that mixes a web app with a native Android app (a common layout: Next.js/React + a Kotlin WebView shell), this manufactures phantom cross-language call edges — which the extraction spec itself forbids (*"`calls` edges MUST stay within one language: a Python function cannot `calls` a JS/TS/Go/Rust/Java symbol and vice versa — cross-language call edges are phantom artifacts, never emit them"*). The local AST resolver never enforced that rule.

Minimal repro:

```tsx
// web/Upcoming.tsx
declare function register(cb: () => void): void;
export function UpcomingPanel() {
  register(refreshHeading);   // callback passed by name
  return null;
}
```

```kotlin
// android/HeadingSensorBridge.kt
class HeadingSensorBridge {
    fun refreshHeading() { println("native sensor") }
}
```

Current output:

```
UpcomingPanel --indirect_call--> .refreshHeading() [INFERRED 0.8]
```

A TSX component "calling" a Kotlin method. The #1659 import-evidence gate doesn't apply (it covers JS/TS **direct** calls only), and the indirect path's callable-target guard passes because the Kotlin method *is* a real callable — just in the wrong language.

Direct calls from non-JS/TS callers have the same hole with no gate at all:

```python
# py/worker.py
def process():
    return refreshHeading()   # binds to the Kotlin fun
```

```
process() --calls--> .refreshHeading() [INFERRED 0.8]
```

Beyond the wrong edge itself, the damage compounds downstream: on the codebase where I hit this, the phantom edge bridged the Android community and the web-navigation community, inflating the Kotlin node's betweenness centrality enough that GRAPH_REPORT.md surfaced it as a top suggested question ("Why does this node connect community X to communities Y, Z?") — an artifact presented as an insight.

## Fix

Filter resolution candidates by **language interop family** right after the label-index lookup, before the single-candidate/import-evidence logic runs. Families are grouped by *real* interop so legitimate cross-language resolution keeps working:

- `jsts`: .js/.jsx/.mjs/.cjs/.ts/.tsx/.mts/.cts + .vue/.svelte/.astro (one module graph)
- `jvm`: .java/.kt/.kts/.scala/.groovy/.gradle (JVM interop — a Kotlin call to a Java method still resolves)
- `native`: .c/.h/.cpp/.cc/.cxx/.hpp/.cu/.cuh/.metal/.m/.mm/.swift (shared headers, ObjC/C++ mixing, Swift↔ObjC bridging)
- single-language families for python, go, rust, ruby, php, dotnet, lua, zig, elixir, julia, dart, shell, powershell

Two deliberate permissive defaults, preserving existing behavior everywhere the guard has no opinion:

- candidates whose family is unknown (no `source_file`, non-code nodes) are never filtered
- callers with an unmapped extension skip the guard entirely

This covers both the `indirect_call` path and the direct-call path (the same candidates list feeds both), and can only ever *narrow* candidates — so an ambiguous 2-candidate name split across languages now collapses to the right-language definition instead of hitting the tie-breakers.

## Tests

`tests/test_cross_language_call_resolution.py` (mirrors the `test_phantom_cross_package_call.py` structure):

- `test_tsx_callback_does_not_bind_to_kotlin_method` — the repro above
- `test_python_call_does_not_bind_to_kotlin_function` — direct-call path for non-JS/TS callers
- `test_same_language_callback_still_resolves` — positive control: TS callback with import evidence keeps its INFERRED `indirect_call`
- `test_jvm_interop_kotlin_call_to_java_still_resolves` — positive control: Kotlin→Java keeps resolving (verified this passes on `main` too, so the family grouping changes nothing for JVM interop)

Full suite: 2986 passed; the handful of failures on my machine (`test_skillgen`, `test_hooks`, `test_ollama_retry_cap`, `test_detect::test_graphifyignore_hermetic_without_vcs`) fail identically on a clean checkout of `main` — environment-related, untouched by this change.

### DIFF
--- a/graphify/extract.py
+++ b/graphify/extract.py
@@ -9828,6 +9828,52 @@ def _lang_is_case_insensitive(source_file: object) -> bool:
     return Path(str(source_file)).suffix.lower() in _CASE_INSENSITIVE_EXTS
 
 
+# Language interop families for cross-file call resolution. A call in one language
+# can never bind by name to a definition in another family — a TSX component does
+# not invoke a Kotlin method, and a Python function does not invoke a Java one.
+# Families are grouped by REAL interop so legitimate cross-language resolution
+# keeps working: Kotlin/Java/Scala/Groovy share the JVM, C/C++/Objective-C/CUDA
+# share headers and symbols (Swift bridges to Objective-C), and JS/TS variants
+# (plus Vue/Svelte/Astro SFC script blocks) compile into one module graph.
+# Extensions absent from this map (docs, configs, unknown languages) resolve to
+# no family and are never filtered — same permissive default as before.
+_LANG_FAMILY_BY_EXT: dict[str, str] = {
+    # JS/TS module graph (SFCs embed JS/TS)
+    ".js": "jsts", ".jsx": "jsts", ".mjs": "jsts", ".cjs": "jsts",
+    ".ts": "jsts", ".tsx": "jsts", ".mts": "jsts", ".cts": "jsts",
+    ".vue": "jsts", ".svelte": "jsts", ".astro": "jsts",
+    # JVM interop
+    ".java": "jvm", ".kt": "jvm", ".kts": "jvm",
+    ".scala": "jvm", ".groovy": "jvm", ".gradle": "jvm",
+    # C-family: shared headers, Objective-C/C++ mix, Swift↔ObjC bridging
+    ".c": "native", ".h": "native", ".cpp": "native", ".cc": "native",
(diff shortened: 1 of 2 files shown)


safishamsi:
Merged into `v8` (bf7fa50), authorship preserved. Reviewed, full suite green (3078 passed). Thanks for the fix!

Issue #1700, opened 2026-07-06
Kotlin/Java: enum constants are not extracted as nodes — "where is ErrorCode.X used" is unanswerable


jerryliurui:
### Environment

- graphify 0.9.7 (PyPI `graphifyy`), macOS, Python 3.10
- Extraction invoked directly via `graphify.extract.extract(files, parallel=False)`

### Summary

Enum **constants** are not extracted as nodes in either Kotlin or Java. The enum class itself gets a node, but its members (`ChatType.SYSTEM`, `ErrorCode.GAME_DONE`, …) do not exist anywhere in the graph, so queries like "where is this error code used" / "who branches on ChatType.SYSTEM" are unanswerable — the constant name simply isn't in the node table.

### Reproducer (3 files)

`ChatType.kt`
```kotlin
package demo

enum class ChatType {
    NORMAL,
    GROUP,
    SYSTEM
}
```

`TypeCheck.kt`
```kotlin
package demo

class TypeCheck {
    fun isSystem(type: ChatType): Boolean {
        return type == ChatType.SYSTEM
    }
}
```

`ErrorCode.java`
```java
package demo;

public enum ErrorCode {
    OK(0),
    GAME_DONE(1001);

    private final int code;

    ErrorCode(int code) {
        this.code = code;
    }
}
```

### Current behavior (0.9.7)

Complete node list for the three files:

```
chattype_kt_chattype        (file)
chattype_chattype           (enum class — no children besides the file)
typecheck                   (file)
typecheck_typecheck         (class)
typecheck_typecheck_issystem (method)
boolean
errorcode                   (file)
errorcode_errorcode         (enum class)
```

`NORMAL` / `GROUP` / `SYSTEM` / `OK` / `GAME_DONE` appear nowhere. `TypeCheck.isSystem` gets a `references` edge to the `ChatType` class, but the fact that it specifically tests `SYSTEM` is lost. (Aside: the Java enum also emits its `errorcode --contains--> errorcode_errorcode` edge twice.)

### Expected

Enum constants extracted as child nodes of the enum class (analogous to methods), e.g.:

```
chattype_chattype --constant--> chattype_chattype_system
typecheck_typecheck_issystem --references--> chattype_chattype_system
```

Even without the usage edge, just having the constant as a searchable node would let downstream tooling find the definition and fall back to text search for usages.

### Real-world impact

[...]

PR #1719, opened 2026-07-07
fix(extract): emit Java enum constants as nodes with case_of edges

ivanzhl:
Addresses #1700 (Java only)

## Problem
Java enum constants aren't extracted as nodes. The enum *type* node is created,
but its members (`ErrorCode.GAME_DONE`) don't exist in the graph, so
"where is this constant defined/used" can't find them. `enum_declaration` is in
Java's class_types, so the walker recurses into the enum body, but
`enum_constant` matches no handled type and falls through, dropping the constant.

## Fix
Add `_java_extra_walk` (dispatched for `tree_sitter_java`) that emits each
`enum_constant` as a child node with a `case_of` edge to the enum, mirroring the
existing Swift `enum_entry` handling.

Before:
    NODE  ErrorCode
    EDGE  ErrorCode.java --contains--> ErrorCode

 After:
    NODE  ErrorCode
    NODE  OK
    NODE  GAME_DONE
    EDGE  ErrorCode.java --contains--> ErrorCode
    EDGE  ErrorCode      --case_of-->  OK
    EDGE  ErrorCode      --case_of-->  GAME_DONE

## Note for reviewer
For a constant with an anonymous body (`MONDAY { void greet(){} }`), the handler
descends into the body and attaches its methods to the constant node rather than
letting them fall to file scope. Without descending, those methods would be dropped. Side effect: it de-collides a
constant's method from a same-named enum method. 

## Out of scope
- Kotlin enum entries — same root gap, not addressed here; can follow up separately.
- Usage edges (`isSystem() --references--> SYSTEM`) — definitions only.
- Pre-existing duplicate `file --contains--> enum` edge (constructor shares the enum's name) — unrelated, untouched.

### DIFF
--- a/graphify/extract.py
+++ b/graphify/extract.py
@@ -2933,6 +2933,32 @@ def _swift_extra_walk(node, source: bytes, file_nid: str, stem: str, str_path: s
     return False
 
 
+# ── Java extra walk for enum constants ───────────────────────────────────────
+
+def _java_extra_walk(node, source: bytes, file_nid: str, stem: str, str_path: str,
+                     nodes: list, edges: list, seen_ids: set, function_bodies: list,
+                     parent_class_nid: str | None, add_node_fn, add_edge_fn,
+                     walk_fn) -> bool:
+    """Handle enum_constant for Java. Returns True if handled."""
+    if node.type == "enum_constant" and parent_class_nid:
+        name_node = node.child_by_field_name("name")
+        if name_node is None:
+            return True
+        const_name = _read_text(name_node, source)
+        line = node.start_point[0] + 1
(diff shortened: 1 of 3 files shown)


safishamsi:
Merged into `v8` (cf36d10), authorship preserved — thanks @ivanzhl. Clean fix, mirrors the existing `_ruby_extra_walk`/`_swift_extra_walk` pattern. One note from review: the anonymous-body case (`MONDAY { void greet(){} }`) emits the constant + `case_of` edge correctly, but the body's methods didn't come through in my probe — not a regression (they weren't captured before), just flagging as a follow-up. This closes the Java half of #1700; the Kotlin half stays open.

PR #1720, opened 2026-07-07
Feat/sql usage

maludb-ed:


safishamsi:
Thanks for the substantial work here, @maludb-ed. After reading through it as a maintainer, I don't think this one can go into upstream core, and I want to be clear about why so it's not discouraging.

As the PR description itself notes, this is the diff of the **MaluDb fork** — it adds a MaluDb export path (persisting the graph into a running MaluDb API server via bearer tokens, linking code to a tenant's data-model graph with `resolve_external`). Upstream `graphifyy` can't carry that: it couples core extraction to an external product and a running server that most users don't have, and the `extract.py` integration (~90 lines) threads that coupling through the hot path. That belongs in your fork / a plugin, which is exactly where you've got it — and the README correctly tells people to install from the fork for the exporter. So nothing here is wasted; it just lives downstream.

There *is* a piece I'd genuinely welcome upstream if you want to split it out: **`sql_evidence.py`** — detecting and parsing SQL embedded in application-code string literals (`looks_like_sql` / `sanitize_sql` / `parse_sql_statement`). That's a general capability with no MaluDb dependency, and graphify already has SQL schema nodes (`extract_sql`) it could link application code to ("which files write to `orders`?") entirely within a single graph. A focused PR that adds just that — SQL-in-code evidence linking FILE nodes to graphify's own table nodes, no external server, no MaluDb namespaces — would be a great addition and I'd review it happily. `pg_introspect.py` (live Postgres catalog enrichment) could also stand alone behind the existing `[postgres]` extra if you want.

Closing this as fork-specific, but the door's open for the decoupled SQL-in-code detection. Thanks again.

PR #1721, opened 2026-07-07
refactor(extract): move extract_terraform to extractors/terraform.py (verbatim)

Cekaru:
Part of the `extract.py` split tracked in #1212. Ports the Terraform/HCL extractor into its own per-language module.

## What changed
- Moved `extract_terraform` and its private `_TF_META_HEADS` constant **verbatim** into `graphify/extractors/terraform.py`.
- `extract.py` keeps the facade re-export (`from graphify.extractors.terraform import extract_terraform  # noqa: F401`), so every existing importer (`__main__.py`, `watch.py`, tests, etc.) is unchanged.
- Registered in `extractors/__init__.LANGUAGE_EXTRACTORS` (alphabetical).
- Added `tests/test_extractors_registry.py` with module-import / facade-identity / registry-identity checks.
- Updated `extractors/MIGRATION.md` status table.

## Behavior preservation
- `git diff --color-moved` confirms the function + constant are **byte-identical** (verified locally with `diff`).
- Net effect on `extract.py`: **-178 lines, +1 import line** — a pure move.
- Helper classification: `_TF_META_HEADS` had 0 remaining uses after the move → private → moved with the function (per MIGRATION.md).

## Tests
- `test_terraform.py` (8) + `test_extractors_registry.py` (1) pass.
- `test_extract.py` (114) pass.
- No test edits outside the new registry test.


safishamsi:
Thanks @Cekaru. The Terraform move itself already landed on `v8` as part of the larger #1737 decomposition — `graphify/extractors/terraform.py` exists with `extract_terraform` (+ `_TF_META_HEADS`) moved verbatim, `extract.py` keeps the facade re-export, and `extractors/LANGUAGE_EXTRACTORS` registers it. That's why this PR now shows as conflicting: the code change is a no-op against current `v8`.

The part that *wasn't* already covered is the regression test you added — there was no facade/registry-identity guard on `v8`. I salvaged it in 9c27a52 (authored to you) and generalized it: instead of one assertion per language, it sweeps the whole `LANGUAGE_EXTRACTORS` registry and asserts `graphify.extract` re-exports the same object and the registry maps to it, plus your concrete terraform anchor. So a future move that forgets a facade re-export now fails loudly.

Closing since the move is already in and the test is merged. Appreciate the clean, verified-move approach — exactly the right pattern for the split.

PR #1722, opened 2026-07-07
test(security): cover GRAPHIFY_MAX_GRAPH_BYTES parsing; clarify unit docstring

Cekaru:
## Problem
`_max_graph_file_bytes()` in `graphify/security.py` resolves the graph-load memory-bomb cap and parses the `GRAPHIFY_MAX_GRAPH_BYTES` env var (plain bytes, `MB`/`GB` suffix, and several fallbacks). Despite the branching logic, it had **no behavioral tests** — only the default constant (`_MAX_GRAPH_FILE_BYTES`) was asserted.

Its docstring also described the suffix as *"decimal multipliers of 1024"*, which is self-contradictory (decimal would be 1000).

## Change
- Add tests covering every branch: unset/blank → default, plain integer, `MB`/`GB` suffix (binary), case-insensitive + space-tolerant suffixes (`3gb`, `5 GB`), unparseable → default (`1.5GB`, `0x10`, `640KB`), and non-positive → default (`0`, `-1`, `-4GB`).
- Clarify the docstring: `MB`/`GB` are treated as **binary** (MiB/GiB). The tests lock this in.

## Notes
- **No behavior change** — tests document existing behavior; only the docstring wording changed.
- `tests/test_security.py` passes (59 tests).

### DIFF
--- a/graphify/security.py
+++ b/graphify/security.py
@@ -38,7 +38,8 @@ def _max_graph_file_bytes() -> int:
     Honors the ``GRAPHIFY_MAX_GRAPH_BYTES`` environment variable so users with
     large codebases can raise the limit without editing source. The value may
     be plain bytes (``671088640``) or carry an ``MB`` / ``GB`` suffix
-    (``640MB``, ``2GB`` — case-insensitive, decimal multipliers of 1024).
+    (``640MB``, ``2GB`` — case-insensitive, binary multipliers: ``MB`` is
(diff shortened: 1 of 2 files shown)


safishamsi:
Merged into `v8` (6c33e3a), authorship preserved. Reviewed, full suite green (3078 passed). Thanks for the fix!

PR #1724, opened 2026-07-08
fix: scale per-term exact/prefix score tiers by squared term coverage (#1602)

### LINKED ISSUE #1602: query: single generic term exact-matching a short leaf label hijacks seed selection in multi-term queries
## Summary

In a multi-term query, a single generic term that exactly equals a short, unrelated leaf node's label receives the `_EXACT_MATCH_BONUS` (1000×IDF) and outscores every genuinely relevant node by ~10×. `_pick_seeds` (gap_ratio 0.2) then drops all other candidates, so the traversal starts from one irrelevant leaf and returns a tiny dead-end subgraph.

Observed on `graphifyy` 0.8.49.

## Repro (real graph, Phoenix/Elixir codebase, ~5,400 nodes)

Query:

```
graphify query "dashboard home page LiveView content KPI cards alerts today grid"
```

The graph contains a rich cluster for the dashboard feature (`IrisWeb.DashboardLive`, `kpi_card()`, `alert_row()`, plus design docs whose labels match 3 of the query terms). Expected seeds somewhere in that cluster.

Actual `_score_nodes` output (top of the ranking):

```
6037.89  home()                          <- "home" == bare label -> exact tier
 680.74  grid_view()                     <- "grid" prefix tier
 621.57  Home / Today Dashboard (Iris)   <- 3 substring hits
 616.01  Home Today Mock
```

`_pick_seeds` keeps only `home()` because every other node is below 20% of the top score. `home()` is a 2-neighbor leaf (a controller action), so BFS depth 2 returns 3 nodes and the answer collapses:

```
Traversal: BFS depth=2 | Start: ['home()'] | 3 nodes found
```

The failure generalizes: any generic English word in a concept query that happens to equal (or prefix) a short function name hijacks the seed. Same graph, `"dashboard liveview kpi alerts today grid"` seeds at `grid_view()` / `gridlines()` for the same reason, while `"dashboard KPI alerts"` (no colliding generic words) seeds correctly.

## Why this feels like a scorer bug rather than user error

The exact-match tier is clearly right for identifier queries (`graphify query "FooBarService"`). But in a 10-term query, an exact hit on **one** term says much less than substring hits on **three** terms. The current scoring has no notion of term coverage, so one lucky collision dominates, and the aggressive seed gap then removes all diversity.

## Possible directions

- Weight the exact/prefix bonuses by the fraction of query terms the node matches, so a 1-of-10 exact hit can't bury a 3-of-10 multi-term match.
- Or relax `_pick_seeds` when the top seed's score comes from a single term of a many-term query — keep at least one seed from the best multi-term-coverage node.
- Or treat very short labels (`home`, `grid`) as low-information exact matches, e.g. dampen the exact bonus by label-token IDF.

Happy to provide more score traces if useful.

fkhawajagh:
Fixes #1602. Follows up on the discussion there; builds on the per-term seed guarantee from #1596 (`d56ee83`) and fixes the scoring half that was left open.

## Root cause

In a multi-term query, a single generic term that exactly equals a short leaf label (`home` vs a `home()` controller action, `list` vs a `list()` function) receives `_EXACT_MATCH_BONUS x IDF` and outscores every node matching several of the query's terms by roughly 10x. `_pick_seeds`' 20% gap cutoff then discards the genuinely relevant candidates. #1596 guarantees each term still contributes a seed, which stops the starvation; but nothing in `_score_nodes` rewarded matching *more* of the query, so the collision still owned the top of the ranking (and `path`-style consumers of `scored[0]` have no per-term backstop).

## The change

`_score_nodes` now scales the per-term exact/prefix tier contributions by squared term coverage:

```
score += tiered * (matched_terms / total_terms) ** 2
```

Design decisions, each of which came out of testing against a real graph (below):

1. **Squared, not linear.** The exact tier is 10x the prefix tier; at linear coverage a 1-of-10-terms exact hit still edges out a 3-of-10-terms prefix+substring match. The regression test pins this coupling: retuning either tier constant re-fails it.
2. **Coverage counts label hits only (exact/prefix/substring).** Source-path hits still score, but do not count toward coverage. Without this, a colliding leaf that lives in the target's own directory (the common case) wins back most of its exact tier via path fragments: on the regression scenario the leaf scored 122.5 vs the target's 66.3 when path hits counted, vs 30.9 / 40.1 when they do not.
3. **Substring and source bonuses stay unscaled**, as does the full-query tier (a full-label match already implies full coverage).
4. **Query tokens are deduped, order-preserving**, as `_pick_seeds` already does. A repeated query word previously double-added every tier; under coverage scaling it would also inflate the matched ratio, quadratically restoring the collision (a query with `list` twice re-ranked the `list()` leaf above the intended target, 154.0 vs 84).
5. **Single-term and full-coverage queries are unchanged** (coverage == 1), so identifier lookups keep exact-match dominance. A test pins the exact pre-change score.

## Results on a real graph

Production graph: 5,700 nodes / 7,899 edges (Phoenix/Elixir codebase plus design docs; the graph both #1602 repros came from). Methodology: run `_query_terms -> _score_nodes -> _pick_seeds` (the `_query_graph_text` flow) on base `v8` (`8a608d1`, which already includes #1596) and on this branch, and diff the top of the ranking and the seed set.

### Identifier queries: byte-identical

`ClientLive.Index` and `DashboardLive` produce identical rankings and identical seeds before and after. Example (`ClientLive.Index`):

| rank | before | after |
|---|---|---|
| 1 | 60028.71 ClientLive.Index | 60028.71 ClientLive.Index |
| 2 | 60026.57 ClientLive.Index (doc) | 60026.57 ClientLive.Index (doc) |
| 3 | 594.28 ClientLive.Form | 148.57 ClientLive.Form |

(Partial-coverage neighbors like `ClientLive.Form` scale down as intended; rank order and seeds are unchanged.)

### Natural-language query: seeds byte-identical

`dashboard KPI alerts` seeds `['Dashboard COI Alerts Integration', 'dashboard_live.ex', 'Dashboard (/) Rebuild Implementation Plan', 'Placement-location seam for overlap alerts', 'kpi.ex']` both before and after.

### Repro 1: `dashboard home page LiveView content KPI cards alerts today grid`

Before (v8 with #1596):

```
6087.92  home()                                   <- 1-term exact collision, rank 1
 703.97  grid_view()
 703.97  gridline_size_pct()
 626.73  Home / Today Dashboard (Iris)            <- intended cluster, below gap window
 621.12  Home Today Mock
seeds: home() first; cluster recovered only via the per-term guarantee
```

After:

```
  72.94  Home / Today Dashboard (Iris)            <- intended cluster, rank 1
  71.92  Dashboard (/) Rebuild to Home-Today Mock (Design)
  63.23  home()                                   <- collision dampened 6087.9 -> 63.2 (96x)
  36.91  Home Today Mock
seeds: ['Home / Today Dashboard (Iris)', 'Dashboard (/) Rebuild to Home-Today Mock (Design)', 'home()', ...]
```

The relevant cluster is now inside the gap window on its own merit, not only via the per-term guarantee, and it also becomes the top-ranked answer for `scored[0]` consumers.

### Repro 2: `ClientLive.Index clients list columns`

Before: three `list()` nodes at 4871.72 vs `ClientLive.Index` at 600.72 (8.1x, outside the 20% gap window; seeded only via the per-term guarantee).

After: `list()` drops to 194.87 (25x reduction) and `ClientLive.Index` at 101.52 is back inside the gap window. `list()` still ranks first (it is a legitimate 1-of-5 exact match), but its dominance collapses from 8.1x to 1.9x and the intended identifier is seeded through the gap window itself.

## Known residual (why #1596 stays load-bearing)

A relevant node matched only via interior substrings (camelCase mid-label terms) still scores below a dampened collision: flat substring scores are ~1 x IDF per term vs `exact/n^2` for the collision. The per-term seed guarantee from #1596 covers exactly this class, which is now documented in `_pick_seeds`' docstring alongside the coverage note. A symmetric coverage boost for the flat tier would be a larger behavioral change than this PR wants to take on; happy to explore it separately if you think it is worth it.

One more trade-off worth naming: on very long queries (10+ content terms after stopword filtering), a node matching exactly one term via prefix (`100/n^2`) can score below a node matching one term via substring (flat 1.0). Both are weak candidates for such queries, and the per-term guarantee still seeds each term's best node (single-term coverage == 1), but it is a place where the eval could be extended if you want stricter tier monotonicity.

## Tests

- New regression test reproducing the second #1602 repro: the query contains the target's literal identifier, yet three same-directory `list()` leaves outrank it on base `v8`. Fails on `v8`, passes with this change, with a 30% score margin so IDF drift will not flake it.
- New no-change test pinning the exact full-coverage score (`query "FooBarService"` class), so identifier dominance is provably untouched.
- `tests/test_serve.py`: 78/78 pass. `ruff check`: clean. The full suite's only failures on this machine are environment-specific (local provider env vars, XAML grammar) and reproduce identically on the unpatched base.

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,15 +15,11 @@
 </div>
 
 <p align="center">
-  <a href="https://www.ycombinator.com/companies/graphify"><img src="https://img.shields.io/badge/Y%20Combinator-S26-F0652F?style=flat&logo=ycombinator&logoColor=white" alt="YC S26"/></a>
-  <a href="https://discord.gg/598Ad9zQZ"><img src="https://img.shields.io/badge/Discord-Join-5865F2?style=flat&logo=discord&logoColor=white" alt="Discord"/></a>
-  <a href="https://safishamsi.gumroad.com/l/qetvlo"><img src="https://img.shields.io/badge/Book-The%20Memory%20Layer-2ea44f?style=flat&logo=gitbook&logoColor=white" alt="The Memory Layer"/></a>
-  <a href="https://github.com/safishamsi/graphify/actions/workflows/ci.yml"><img src="https://github.com/safishamsi/graphify/actions/workflows/ci.yml/badge.svg?branch=v8" alt="CI"/></a>
   <a href="https://pypi.org/project/graphifyy/"><img src="https://img.shields.io/pypi/v/graphifyy" alt="PyPI"/></a>
   <a href="https://pepy.tech/project/graphifyy"><img src="https://img.shields.io/pepy/dt/graphifyy?color=blue&label=downloads" alt="Downloads"/></a>
-  <a href="https://github.com/sponsors/safishamsi"><img src="https://img.shields.io/badge/sponsor-safishamsi-ea4aaa?logo=github-sponsors" alt="Sponsor"/></a>
-  <a href="https://www.linkedin.com/company/graphify-labs"><img src="https://img.shields.io/badge/LinkedIn-Graphify%20Labs-0077B5?logo=linkedin" alt="LinkedIn"/></a>
-  <a href="https://x.com/graphifyy"><img src="https://img.shields.io/badge/X-graphifyy-000000?logo=x&logoColor=white" alt="X"/></a>
+  <a href="https://github.com/safishamsi/graphify/actions/workflows/ci.yml"><img src="https://github.com/safishamsi/graphify/actions/workflows/ci.yml/badge.svg?branch=v8" alt="CI"/></a>
+  <a href="https://discord.gg/598Ad9zQZ"><img src="https://img.shields.io/badge/Discord-Join-5865F2?style=flat&logo=discord&logoColor=white" alt="Discord"/></a>
+  <a href="https://www.ycombinator.com/companies/graphify"><img src="https://img.shields.io/badge/Y%20Combinator-S26-F0652F?style=flat&logo=ycombinator&logoColor=white" alt="YC S26"/></a>
 </p>
 
 Type `/graphify` in your AI coding assistant and it maps your entire project (code, docs, PDFs, images, videos) into a **knowledge graph** you can **query instead of grepping** through files.
@@ -757,6 +753,7 @@ graphify label ./my-project --backend=openai --model gpt-4o   # force a specific
 - [How it works](docs/how-it-works.md) — the extraction pipeline, community detection, confidence scoring, benchmarks
 - [ARCHITECTURE.md](ARCHITECTURE.md) — module breakdown, how to add a language
 - [Optional integrations](docs/docker-mcp-sqlite.md) — Docker MCP Toolkit + SQLite
+- [The Memory Layer](https://safishamsi.gumroad.com/l/qetvlo) — the book on the ideas behind graphify, the architecture end to end
 
 ---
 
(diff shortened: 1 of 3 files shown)


safishamsi:
Merged the scoring fix on v8 in 2f9deae, authored to you. This is a well-reasoned change and the writeup made it easy to validate.

I verified it independently before landing:
- The two regression tests plus the full `test_serve.py` suite pass (80 tests).
- Edge cases I added on my side all hold: empty/whitespace queries don't divide by zero (the `if tiered:` guard covers the zero-term case), duplicate query words are deduped so `list list` scores identically to `list`, single-term/full-coverage queries are byte-identical to pre-change (coverage == 1), and the #1602 repro (`ClientLive.Index clients list columns`) now ranks the target at rank 0 instead of the `list()` collision.
- Full suite green (3094 passed / 3 skipped).

One note: I dropped the README badge reshuffle that was bundled in — that part is unrelated to #1602 and touches the project's own branding layout, so I kept the PR to just the scoring fix and its tests. Thanks for the thorough real-graph methodology in the description; credit is yours.

PR #1736, opened 2026-07-08
fix: uv interpreter detection falls back to graphify-less system python (#1735)

### LINKED ISSUE #1735: Interpreter-detection in SKILL.md silently falls back to system python3 (uv tool run graphifyy python is invalid)
## What happened

Installed via the documented path:

```
uv tool install graphifyy
graphify install --platform claude
```

Then ran `/graphify .` on a repo. Step 1 of the shipped `SKILL.md` resolves the interpreter with:

```bash
_UV_PY=$(uv tool run graphifyy python -c "import sys; print(sys.executable)" 2>/dev/null)
if [ -n "$_UV_PY" ]; then PYTHON="$_UV_PY"; fi
...
if [ -z "$PYTHON" ]; then PYTHON="python3"; fi
```

`uv tool run graphifyy python -c "..."` does not do what's intended. Since the package name (`graphifyy`) differs from the installed executable name (`graphify`), uv interprets `python` as an *argument* to a command named after the package, not as "run the `python` executable inside the `graphifyy` environment":

```
$ uv tool run graphifyy python -c "import sys; print(sys.executable)"
An executable named `graphifyy` is not provided by package `graphifyy`.
The following executables are available:
- graphify
- graphify-mcp

Use `uv tool run --from graphifyy <EXECUTABLE-NAME>` instead.
```

Because the script redirects stderr to `/dev/null`, this failure is invisible. `_UV_PY` ends up empty, `PYTHON` falls through to bare `python3`, which resolves to the system interpreter (Homebrew python3.14 on macOS) — not the `graphifyy` tool venv. `graphify-out/.graphify_python` then gets written with a path that can't `import graphify`:

```
$ /opt/homebrew/opt/python@3.14/bin/python3.14 -c "import graphify"
ModuleNotFoundError: No module named 'graphify'
```

Every subsequent step in the skill that shells out via `$(cat graphify-out/.graphify_python)` breaks immediately (detect, extract, build all fail with `ModuleNotFoundError`).

## Environment

- macOS (Darwin), uv 0.11.17
- `uv tool install graphifyy` (fresh install, no prior tool state)
- `which graphify` → `~/.local/bin/graphify`, shebang → `~/.local/share/uv/tools/graphifyy/bin/python` (the correct interpreter — this one works)

## Expected

Step 1's uv-detection branch should either:
- use `uv tool run --from graphifyy python -c "..."` (the syntax uv's own error message suggests), or
- skip straight to the shebang-parsing fallback (branch 2), which already works correctly and is more robust for asymmetric package/executable names.

## Suggested fix

In `SKILL.md` Step 1, replace:
```bash
_UV_PY=$(uv tool run graphifyy python -c "import sys; print(sys.executable)" 2>/dev/null)
```
with:
```bash
_UV_PY=$(uv tool run --from graphifyy python -c "import sys; print(sys.executable)" 2>/dev/null)
```

Happy to send a PR if that's useful — this bit me on a first-time install so it's likely hitting other users silently (the `2>/dev/null` masks it completely; you just get confusing `ModuleNotFoundError`s three steps later with no clue why).

mohammedMsgm:
Fixes #1735

## Bug

Step 1 of the shipped `SKILL.md` (and its per-platform generated variants) resolves the interpreter with:

```bash
_UV_PY=$(uv tool run graphifyy python -c "import sys; print(sys.executable)" 2>/dev/null)
```

Since the installed package name (`graphifyy`) differs from its executable name (`graphify`), uv treats `python` as an argument to a command literally named `graphifyy` rather than "run the `python` inside the `graphifyy` env", and errors:

```
$ uv tool run graphifyy python -c "import sys; print(sys.executable)"
An executable named `graphifyy` is not provided by package `graphifyy`.
The following executables are available:
- graphify
- graphify-mcp

Use `uv tool run --from graphifyy <EXECUTABLE-NAME>` instead.
```

The trailing `2>/dev/null` swallows this, so `_UV_PY` comes back empty, `PYTHON` falls through to a bare `python3` (the Homebrew/system interpreter, not the tool venv), and every subsequent step that shells out via the recorded interpreter fails with `ModuleNotFoundError: No module named 'graphify'`. Hit this on a fresh `uv tool install graphifyy && graphify install --platform claude` on macOS.

## Fix

- `tools/skillgen/fragments/core/aider.md`, `tools/skillgen/fragments/core/devin.md`, `tools/skillgen/fragments/shell/posix.md`: `uv tool run graphifyy python` → `uv tool run --from graphifyy python` (the syntax uv's own error message recommends).
- Regenerated every committed artifact and `expected/` output via `python -m tools.skillgen --bless`.
- Added `_is_uv_tool_run_from_fix_line` to `_SANCTIONED_MONOLITH_DIFFS` so `--monolith-roundtrip` recognizes the change on the aider/devin monoliths.
- CHANGELOG entry.

## Verification

```
$ uv tool run --from graphifyy python -c "import sys; print(sys.executable)"
/Users/.../uv/tools/graphifyy/bin/python
$ "$(uv tool run --from graphifyy python -c 'import sys; print(sys.executable)')" -c "import graphify; print('import OK')"
import OK
```

```
$ python -m tools.skillgen --check
$ python -m tools.skillgen --audit-coverage
audit-coverage OK: every per-host v8 heading single-homes in that host's render.
$ python -m tools.skillgen --schema-singleton
schema-singleton OK: the file_type enum is the six-value superset everywhere.
$ python -m tools.skillgen --monolith-roundtrip
monolith-roundtrip OK: each monolith matches v8 modulo the enum unification.
$ python -m tools.skillgen --always-on-roundtrip
always-on-roundtrip OK: each always_on/*.md reproduces its former constant byte for byte.
$ pytest tests/test_skillgen.py -q
57 passed in 1.86s
```

Diff is otherwise mechanical (fragment source edit → regenerated artifacts), no manual edits to generated files.


safishamsi:
Thanks for catching this and for the fix, @mohammedMsgm. The root cause you identified is exactly right: `uv tool run graphifyy python` is invalid because the `graphifyy` package exposes its executable as `graphify`, so uv looked for a `graphifyy` command that doesn't exist, and `2>/dev/null` hid the error, dropping the probe onto a graphify-less system python.

The diff here spans ~227k lines against `main` (it looks like a lot of generated/build artifacts got swept in, and v8 is our default branch), so I couldn't merge it directly. I've applied your fix by hand on `v8` in ea8b804, authored to you:

- Corrected the POSIX interpreter probe to `uv tool run --from graphifyy python -c ...` in the skillgen source (`tools/skillgen/fragments/shell/posix.md` plus the aider/devin monoliths) and regenerated all `skill*.md`.
- Added a sanctioned monolith-diff predicate and re-blessed the `expected/` snapshots so the drift guard stays green.
- The PowerShell path was already correct (it probes the venv `python.exe` directly), so no change was needed there.

Full suite passes (3092 passed / 3 skipped). Closing in favor of the direct commit; credit is yours. Thanks again.

PR #1737, opened 2026-07-08
refactor: decompose extract.py and __main__.py into focused modules

TPAteeq:
Splits the two largest modules into cohesive, single-responsibility modules —
**verbatim moves only**, every original import path preserved via re-exports, and
the full suite unchanged (3036 passed, 29 skipped) after every commit.

## Why

`extract.py` (17,054 LOC) and `__main__.py` (5,368 LOC) were by far the largest
files in the package and mixed several independent subsystems each. This is pure
decomposition: no behavior change, no renames, no reformatting.

## Size impact

| File | before | after |
|---|---|---|
| `graphify/extract.py` | 17,054 | **4,740** (−72%) |
| `graphify/__main__.py` | 5,368 | **673** (−87%) |
| `graphify/export.py` | 1,671 | **962** (−42%) |

## What moved (one subsystem per commit)

**extract.py → `graphify/extractors/`** (continues the `MIGRATION.md` / #1212 effort):
- `engine.py` — the `_extract_generic` tree-sitter engine + its 67-function closure and 10 private type-tables (~4.1k LOC).
- `resolution.py` — the cross-file symbol/import resolution passes (60 fns).
- `models.py` — shared types (`LanguageConfig`, the `_Symbol*Fact` dataclasses) + two shared caches.
- 23 bespoke language extractors: dart, rust, go, powershell, fortran, sql, dm, bash, apex, terraform, sln, pascal_forms, json_config, verilog, markdown, pascal, objc, julia.

**__main__.py**:
- `install.py` — the whole install/uninstall subsystem **and** the install/platform CLI dispatch (`dispatch_install_cli`).
- `cli.py` — every non-install subcommand (`dispatch_command`).

**export.py → `graphify/exporters/`**: `html.py`, `graphdb.py`, `base.py`.

## Why it's safe

- **Verbatim.** Each move was validated by AST closure-privacy analysis and byte-identity checks; nothing was rewritten.
- **No import cycles.** Strict layering `extract → engine → resolution → models/base`; `__main__ → cli`; the one back-edge (the `graphify <path>` redirect's recursive `main()`) uses a lazy import.
- **Nothing to update downstream.** `extract.py` / `__main__.py` / `export.py` re-export every moved name, so `from graphify.extract import extract_dart`, `from graphify.__main__ import claude_install`, etc. all still resolve (object identity preserved). No test or importer changed.
[...]

PR #1738, opened 2026-07-08
fix(extract): extract Kotlin enum entries as nodes with case_of edges

ivanzhl:
Addresses #1700 (Kotlin half; Java was #1719)

## Problem
Kotlin enum entries aren't extracted as nodes. The enum type node is created, but
its members (`ChatType.SYSTEM`) don't exist in the graph. Unlike Java, the walker
never even descends into the enum body: a Kotlin enum's body is an
`enum_class_body`, which wasn't in the Kotlin config's `body_fallback_child_types`,
so `_find_body` returned `None` and the entries were never visited.

## Fix
Two parts, mirroring the Swift handling:
1. Add `enum_class_body` to `_KOTLIN_CONFIG.body_fallback_child_types` so the
   walker descends into the enum body.
2. Add `_kotlin_extra_walk` (dispatched for `tree_sitter_kotlin`) that emits each
   `enum_entry` as a child node with a `case_of` edge to the enum.

Before:
    NODE ChatType
    EDGE ChatType.kt --contains--> ChatType
After:
    NODE ChatType / NORMAL / GROUP / SYSTEM
    EDGE ChatType --case_of--> NORMAL / GROUP / SYSTEM


safishamsi:
Merged into `v8` (00dd978), authorship preserved — thanks @ivanzhl, this closes the Kotlin half of #1700.

Small note: the #1737 refactor landed just before this, so the walk engine moved from `extract.py` into `graphify/extractors/engine.py` (`_KOTLIN_CONFIG` stayed in `extract.py`). I re-applied your exact change onto the new layout — `_kotlin_extra_walk` + the tree_sitter_kotlin dispatch in engine.py, `enum_class_body` in the config — so the diff is functionally identical to your PR. Verified: NORMAL/GROUP/SYSTEM now emit with `case_of` edges to ChatType; full suite green (3092). Ported your fixture + test too.

PR #1739, opened 2026-07-08
fix(extract): scope Pascal/Delphi call resolution + resolve inherited calls across files

richtext:
## Summary

Two related fixes to Pascal/Delphi `calls` edge resolution, found while evaluating graphify against a large real-world Delphi/MTM-framework codebase:

1. **Scope call resolution to class + inherits chain (same file).** Both `extract_pascal` (tree-sitter) and `_extract_pascal_regex` (fallback) resolved every call via a single file-wide `{method_name_lower: node_id}` dict with no class scoping. Two unrelated classes declaring a same-named method (a common pattern — property accessors, generated COM/TLB wrapper classes) silently collapsed onto whichever declaration was inserted last, producing wrong cross-class `calls` edges. Now resolves via: own class → ancestor chain (`inherits`) → file-level free function → unambiguous file-wide match: ambiguous at every level → no edge, rather than a guess (same "god-node guard" principle as `resolve_ruby_member_calls`).

2. **Resolve calls to methods inherited across file boundaries.** Real Delphi/MTM-style code very commonly splits a class across two files (a generated base class + a manual descendant that extends it in a separate unit — e.g. Sistec's `Th0Xxx`/`Th5Xxx` pattern). A call from the descendant to a method it inherits from the base fell outside any one file's own scope. Adds `graphify/pascal_resolution.py`, a corpus-wide post-extraction resolver (registered via `resolver_registry`, same mechanism as the Ruby resolver) that walks the `inherits` chain across files using the full merged graph. Also fixes a related bug surfaced while testing this: both extractors, on resolving a base class to another file, still added a duplicate stub node carrying the *referencing* file's `source_file` instead of the base class's own — colliding with the real node under cross-file id disambiguation and producing two different ids for what should be one class.

## Verification

- 46 tests added (`tests/test_pascal_call_scoping.py`, `tests/test_pascal_resolution.py`), covering both extractors, own-class resolution, the cross-class collision regression (both directions), ancestor-chain resolution, and cross-file resolution (own-class call still resolves, cross-file inherited call resolves, an unrelated same-named method in a third file is not crossed into, resolver registration).
- Confirmed these tests fail without the corresponding fix and pass with it (checked independently for both commits).
- No regressions: existing Pascal/Delphi/Lazarus suite and the full test suite pass unchanged (same pre-existing unrelated failures in Terraform/Ollama/image-vision/install areas, none touched by this change).
- Re-validated against a real ~1200-file Delphi/MTM-framework codebase (not included here): eliminated ~30% of cross-class false-positive `calls` edges on a COM/TLB wrapper module with 3 near-identical classes sharing method names, and correctly resolved cross-file inherited calls on a generated-base/manual-descendant module pair.

## Test plan

- [x] `pytest tests/test_pascal.py tests/test_pascal_call_scoping.py tests/test_pascal_resolution.py` — all pass
- [x] Full suite (`pytest -q`) — no new failures
- [x] Verified new tests fail on pre-fix code (both commits, independently)

### DIFF
--- a/graphify/cache.py
+++ b/graphify/cache.py
@@ -276,7 +276,11 @@ def _relativize_source_files_in(payload: dict, root: Path) -> None:
         root_resolved = Path(root).resolve()
     except OSError:
         return
-    for bucket in ("nodes", "edges", "hyperedges"):
+    # raw_calls (#: Pascal/Delphi cross-file inherited-call resolution) carries
+    # source_file the same way nodes/edges/hyperedges do, so it needs the same
+    # portable-path treatment for cache entries to round-trip correctly across
+    # machines/checkout directories.
+    for bucket in ("nodes", "edges", "hyperedges", "raw_calls"):
         for item in payload.get(bucket, []):
             if not isinstance(item, dict):
                 continue
@@ -307,7 +311,7 @@ def _absolutize_source_files_in(payload: dict, root: Path) -> None:
         root_resolved = Path(root).resolve()
     except OSError:
         return
-    for bucket in ("nodes", "edges", "hyperedges"):
+    for bucket in ("nodes", "edges", "hyperedges", "raw_calls"):
         for item in payload.get(bucket, []):
             if not isinstance(item, dict):
                 continue
@@ -400,7 +404,7 @@ def save_cached(path: Path, result: dict, root: Path = Path("."), kind: str = "a
     # source_file field's original absolute form. Mutating the input here would
     # silently break those remaps on the first extraction pass.
     on_disk = result
-    if isinstance(result, dict) and any(result.get(k) for k in ("nodes", "edges", "hyperedges")):
+    if isinstance(result, dict) and any(result.get(k) for k in ("nodes", "edges", "hyperedges", "raw_calls")):
         import copy as _copy
         on_disk = _copy.deepcopy(result)
         _relativize_source_files_in(on_disk, root)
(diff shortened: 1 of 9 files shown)


safishamsi:
Merged on v8 in d89efbf, authored to you. This is a well-engineered pair of fixes and the writeup made the design intent clear.

Because the #1737 module refactor landed on v8 after you opened this, the PR conflicted (the Pascal extractors moved from `extract.py` to `graphify/extractors/pascal.py`). I re-applied it to the new layout: the extractor changes went into `graphify/extractors/pascal.py`, the resolver registration and import stayed in `extract.py`, and `graphify/pascal_resolution.py` + `cache.py` + the fixtures/tests applied unchanged.

One adaptation your PR predated: on current v8 the tree-sitter Pascal extractor emits a `method` edge for both the interface declaration and the implementation, so the same `method_nid` reaches the cross-file resolver twice. Your god-node guard (`len(candidates) == 1`) then read that as ambiguity and resolved every inherited call to nothing (both `test_pascal_resolution.py` cases failed). I made the resolver count distinct method nids instead of edge multiplicity — duplicate edges to the same method aren't a collision. With that, all 50 Pascal tests pass and the guard still correctly rejects genuine same-name-different-class collisions.

Verified: `tests/test_pascal.py`, `tests/test_pascal_call_scoping.py`, `tests/test_pascal_resolution.py` all green, and the full suite passes (3106 passed / 3 skipped). Credit is yours. Thanks for the god-node-guard framing and the cross-file resolver design — both were faithful to how the Ruby resolver already works.

Issue #1716, opened 2026-07-07
feat: Add support for Qt C++ and QML language parsing via Tree-sitter


lkdigitalworks-53:
### **Is your feature request related to a problem? Please describe.**
Currently, Graphify does not support Qt (C++) and the QML (Qt Modeling Language) ecosystem when parsing codebases. When pointing Graphify to a modern desktop, mobile, or embedded UI application built using the Qt framework, it cannot extract deterministic AST subgraphs for `.qml` files. Furthermore, traditional C++ parsers often struggle to cleanly resolve custom meta-object system mechanics (like signals, slots, and properties) unique to Qt development, leading to missing structural dependencies, broken cross-file links, and unmapped architectural hierarchies.

### **Describe the solution you'd like**
I would like Graphify to add support for Qt and QML languages in its deterministic AST extraction layer (Pass 1). 
Specifically:
1. **QML Grammar Support:** Integrate the `tree-sitter-qml` parser to read, extract, and index UI components, properties, inline JavaScript functionalities, signals, and imports inside `.qml` files.
2. **Cross-Language Bindings:** Enable Graphify to build dependency edges between backend C++ logic (classes derived from `QObject`, registered via `qmlRegisterType` or `QML_ELEMENT`) and frontend QML visual layers.
3. **Qt Macros Handling:** Ensure the core C/C++ parsing pass successfully recognizes and surfaces connections stemming from Qt-specific macros (`Q_OBJECT`, `Q_PROPERTY`, `slots:`, `signals:`).

### **Describe alternatives you've considered**
- **LLM-Inference Only (Pass 3):** Relying solely on LLM fallback logic to piece together how QML and C++ link up. However, this relies on `INFERRED` or `AMBIGUOUS` confidence tags rather than a 100% deterministic, local `EXTRACTED` map, driving up token usage and lowering graph accuracy on complex codebases.
- **Pre-processing/Renaming:** Changing `.qml` file extensions to `.js` or `.json` to trick existing parsers, which fails completely because QML's structural layout is distinct.

### **Additional context**
- **Tree-sitter Parser for QML:** There are open-source tree-sitter implementations available (such as `tree-sitter-qml`) that can be integrated into Graphify’s core multi-language pipeline alongside its existing 33 supported languages.
- This feature would significantly benefit developers building cross-platform consumer apps, automotive interfaces, flight management systems, or desktop software leveraging the Qt framework who want to utilize toolkits like Claude Code or Cursor with a complete repository map.

[...]

Issue #29, opened 2026-04-07
UnicodeDecodeError on Windows (GBK locale) during graphify install


Will-hxw:
## Bug Description
Running `graphify install` on Windows with GBK default encoding fails with `UnicodeDecodeError`.

## Steps to Reproduce
1. `pip install graphifyy`
2. `graphify install`
3. Crash:
UnicodeDecodeError: 'gbk' codec can't decode byte 0x80 in position 44

## Root Cause
All 16 `read_text()`/`write_text()` calls in `graphify/__main__.py` omit the `encoding` parameter. Windows GBK locale defaults to GBK instead of UTF-8.

## Proposed Fix
Add `encoding="utf-8"` to every `read_text()` and `write_text()` call. PR already prepared: https://github.com/Will-hxw/graphify/pull/1

safishamsi:
Fixed in v0.3.4. All 16 `read_text()` and `write_text()` calls in `__main__.py` now pass `encoding="utf-8"` explicitly. Windows GBK/ANSI locales will no longer crash on `graphify install`.

```
pip install --upgrade graphifyy
```

PR #1748, opened 2026-07-09
feat(extract): add QML support and Qt C++/QML bridge (#1716)

Hamza-Masuadi:
## Summary

Adds Qt/QML support to graphify's deterministic AST extraction layer, covering the full scope of #1716:

1. **QML grammar support** — new `tree-sitter-qmljs` extractor for `.qml` files
2. **Cross-language C++ ↔ QML bindings** — `QML_NAMED_ELEMENT` / `qmlRegisterType` aliases resolve onto real C++ class nodes
3. **Qt macro handling** — `signals:`, `slots:`, `Q_PROPERTY`, and bare `Q_OBJECT`/`Q_GADGET` parsing fixes

### QML extractor (`graphify/extractors/qml.py`)

Walks the QML object tree and models:

- Real objects vs grouped bindings (uppercase type vs lowercase property group)
- File root as a reusable component named after the file stem
- `id:` bindings, properties, functions, signals, inline components, enums
- `Behavior on <prop> { ... }` and `ui_object_array` children
- Same-file references/calls (with member-call receiver validation via `func_owner`)
- Cross-file component instantiation via sourceless stubs + existing stub-rewire
- Import statements as `imports_from` edges

Wired into `_DISPATCH`, `LANGUAGE_EXTRACTORS`, `CODE_EXTENSIONS`, and `_LANG_FAMILY`.

### C++ / Qt bridge (`extractors/engine.py` + `extract.py`)

- `signals:` / `slots:` bodyless prototypes become tagged `signal` / `slot` nodes
- `Q_PROPERTY(type name READ … WRITE … NOTIFY …)` emits a `property` node plus accessor edges
- `QML_NAMED_ELEMENT("Alias")` and `qmlRegisterType<T>(…, "Alias")` feed `_resolve_cpp_qml_aliases`
- Bare `Q_OBJECT`/`Q_GADGET` get a semicolon preprocess so following `signals:`/`slots:` sections parse cleanly
- Macro noise (`Q_OBJECT`, `QML_NAMED_ELEMENT`, …) is suppressed as field nodes

### `tree-sitter-qmljs` dependency

`tree-sitter-qmljs` is not on PyPI, so `pip install graphifyy[qml]` cannot resolve the extra on its own (and PyPI rejects git-URL deps in published metadata). Left out of the `all` extra for the same reason.

- Users: `pip install git+https://github.com/yuja/tree-sitter-qmljs` first
- Dev/CI: pinned to `yuja/tree-sitter-qmljs@565872b` (scanner.c fix from #29 + regenerated sources)
- Asked upstream for a PyPI release: yuja/tree-sitter-qmljs#30

## Test plan

- [x] 17 unit tests in `tests/test_qml.py` (components, bindings, Behavior, ids, stubs, builtins, member-call disambiguation, bare refs, arrays, imports, inline components, missing grammar)
- [x] 9 unit tests in `tests/test_cpp_qml_bridge.py` (named-element alias, bare QML_ELEMENT, ambiguous alias guard, qmlRegisterType, signals, slots, Q_PROPERTY accessors, macro-noise suppression, build survival)
- [x] Related C++/QML suites green; ruff clean on touched files
- [x] Smoke-tested against a real Qt/QML app (~450 `.qml` files): zero extraction crashes; signal nodes and alias bridging verified


Hamza-Masuadi:
Update on the `tree-sitter-qmljs` dependency (the `qml` extra used by `extract_qml()`):

- The Python binding had a build bug (missing `src/scanner.c`), which blocked this extra from working at all. Filed and got the fix merged upstream: yuja/tree-sitter-qmljs#29 (merged as `073d361e`, with generated sources refreshed at `565872b`).
- `pyproject.toml`'s `qml` dev dependency now pins directly to `yuja/tree-sitter-qmljs@565872b` — verified it builds cleanly and all QML tests pass against it.
- `tree-sitter-qmljs` has never been published to PyPI, so `pip install graphifyy[qml]` can't resolve this extra on its own (and it's excluded from the `all` extra for the same reason — see the `pyproject.toml` comment for details). Filed yuja/tree-sitter-qmljs#30 requesting a PyPI release to fix that properly.
- Until that lands, this PR ships as-is: QML support works, but users need to `pip install git+https://github.com/yuja/tree-sitter-qmljs` by hand before installing `graphifyy[qml]`.

Hamza-Masuadi:
Update: `tree-sitter-qmljs` is now on PyPI 🎉

yuja published it as [`tree-sitter-qmljs` 0.3.1](https://pypi.org/project/tree-sitter-qmljs/) (2026-07-11) in response to yuja/tree-sitter-qmljs#30. This resolves the git-pin caveat from my earlier comment.

Changed in the latest push (`41b4030`):
- `qml` extra and dev group now use `tree-sitter-qmljs>=0.3.1` (dropped the `@ git+…@565872b` pin).
- Added `tree-sitter-qmljs>=0.3.1` to the `all` extra, since it now resolves cleanly from PyPI.
- Regenerated `uv.lock` (`565872b5` → `0.3.1`).

`pip install graphifyy[qml]` now works without any manual git install. One note: the package ships as an sdist only, so a C toolchain is still needed to compile the scanner on install. All QML + Qt C++/QML bridge tests pass against the PyPI build.

Issue #1689, opened 2026-07-05
R (.r/.R) listed in CODE_EXTENSIONS but has no AST extractor — silently drops all R files with zero warning


Sneakypeat:
## Summary

`.r`/`.R` is included in `detect.py`'s `CODE_EXTENSIONS` set, so R files are classified as `FileType.CODE` and counted in the scan summary (e.g. `found 24 code, ...`). But `extract.py`'s `_DISPATCH` dict — the table that maps extensions to an actual AST extractor function — has no `.r`/`.R` entry at all (it covers ~45 languages: Python, JS/TS, Go, Rust, Java, Ruby, C/C++, Swift, Kotlin, Scala, PHP, Lua, Zig, PowerShell, Elixir, Objective-C, Julia, Fortran, Vue, Svelte, Astro, Dart, Verilog, SQL, Pascal/Delphi, DragonFly BASIC, .NET project files, etc. — but not R).

## Reproduction

Run `graphify extract .` on a project that's mostly `.R` scripts (e.g. an R/renv analysis project with ~17 `.R` files and a handful of `.py` helper scripts).

- Scan step reports something like `found 24 code, 0 docs, 1 papers, 0 images` — the R files are counted as code.
- AST extraction runs on all 24 "code" files.
- The resulting `graph.json` only contains nodes/edges from the non-R files (in my case, 7 small Python scripts). All ~17 `.R` files contribute **zero** nodes and **zero** edges.
- No warning, error, or log line indicates this. The CLI reports success (`wrote graph.json: N nodes, M edges, K communities`) as if the whole codebase was mapped.

## Root cause

In `extract.py`:

```python
def _get_extractor(path):
    ...
    return _DISPATCH.get(path.suffix)
```

For a `.R` file, `_DISPATCH.get(".R")` returns `None` (no key for `.r` or `.R`). The caller handles this by silently returning an empty result:

```python
extractor = _get_extractor(path)
if extractor is None:
    return idx, {"nodes": [], "edges": []}
```

Meanwhile `detect.py::classify_file` lowercases the extension before checking `CODE_EXTENSIONS` (`ext = path.suffix.lower()`), so `.R` files *do* get classified as `FileType.CODE` and counted in scan stats — creating a mismatch between "files graphify says it found as code" and "files graphify can actually parse."

## Impact

[...]

Issue #1666, opened 2026-07-04
Ruby: stable subset of files yields zero nodes in full-repo runs (0.9.6) — each extracts fine in isolation


krishnateja7:
First: thanks for the fast turnaround on #1634 and #1640 — both verified working on our repo after upgrading to 0.9.6. Plain modules and `Struct.new`/`Data.define` constants now get real nodes (63 new `Result`-struct nodes here), constant-receiver edges land, and our Rails-truth sidecar's file-node fallback counter went 99 → 0 exactly as predicted. Roughly 1,000 edge-pairs our sidecar used to supply are now emitted natively. Great releases.

While re-measuring on 0.9.6 we found two new issues (filing separately; they may share a root cause in parallel extraction). This one is the more impactful of the two.

## Symptom

In full-repo runs of `graphify update .`, a **stable subset of Ruby files yields ZERO nodes** — not even file nodes. In our repo it is exactly 5 of ~700 `app/**/*.rb` files, and the set is **identical across consecutive runs**.

Detection method:

```ruby
covered = graph["nodes"].map { |n| n["source_file"] }.to_set
Dir["app/**/*.rb"].reject { |f| covered.include?(f) }
# => same 5 files, every run
```

## Each file extracts perfectly in isolation

Copy any of the 5 files alone (or into a small fresh probe corpus) and run `graphify update .` there: full extraction — class/module node, method nodes, file node all present, stable across reruns.

## Failure is batch-context-dependent, not content-dependent

- A 3-file probe corpus containing exactly three of the failing files: 1 extracted fine, 2 produced zero nodes.
- The same 2 files later extracted fine in a different 4-file corpus (original content, unmodified).
- Minimal probes of every construct the files use (`extend self`, `begin/rescue/retry`, namespaced superclass `class X < Devise::PasswordsController`, def-less AR model body, endless methods) all extract fine.
- Stripping comments from two of the files made them extract in a batch where the originals didn't — but the originals also extract fine in other batches, so comment content is not the trigger either. The only consistent predictor is which other files are being processed alongside.

File shapes for reference (no shared decisive construct): one Devise-inheriting namespaced controller (~30 lines), one ActiveRecord model with a kwargs `def self.` upsert + `begin/rescue/retry` (~70 lines, comment-heavy), three `extend self` modules with long leading comment blocks.

## Hypothesis

[...]

PR #1750, opened 2026-07-09
fix(extract): warn when files are skipped due to a missing optional dependency (#1745)

### LINKED ISSUE #1745: .sql files silently extract 0 nodes when the [sql] extra is missing
## Summary
When the `[sql]` extra is not installed, `.sql` files are counted as code and scanned, but silently produce **0 nodes** — no warning, no error. The graph looks successfully built while the entire SQL corpus is missing.

## Environment
- graphifyy 0.9.7 (pipx), Python 3.14.5, macOS (arm64)

## Reproduction
```bash
pipx install graphifyy            # note: no [sql] extra
graphify extract <dir-with-213-sql-files> --no-cluster
```
Output reports `found 227 code, 0 docs, ...` and writes a graph — but it contains only the non-SQL nodes (in our case a handful of TypeScript files). The per-file zero-node warning (#1666-style) listed only 2 JSON files; the 213 `.sql` files vanished without a trace.

After `pipx install --force "graphifyy[sql]"` the exact same command extracts tables, FK references and triggers correctly (single test file: 12 nodes / 35 edges).

## Expected
Either a hard gate like the existing LLM-key check ("A code-only corpus needs no key") — e.g. *"N .sql files need the [sql] extra: pip install 'graphifyy[sql]'"* — or at minimum, the .sql files should appear in the zero-node warning list.

## Impact
Users reasonably believe their database corpus is indexed when it isn't; we lost an evaluation pass to this before finding the extra by searching the README.

rithyKabir:
Fixes #1745.

## Problem

When the `[sql]` extra isn't installed, `extract_sql` returns an `"error"` result for every `.sql` file — and nothing consumes it. Both existing safety nets deliberately miss this case:

- the **#1666** zero-node warning skips any result carrying an `error`
- the **#1689** no-extractor warning only covers extensions with no dispatch entry, and `.sql` has one

So the graph builds "successfully" while the entire SQL schema silently contributes nothing. The same mechanism affects the other optional grammars that hard-fail on import: `.tf`/`.tfvars`/`.hcl` (`[terraform]`) and `.dm`/`.dme` (`[dm]`). Pascal is unaffected — it falls back to its regex extractor.

## Fix

`extract()` now surfaces these files right after the #1689 block, grouped by extension, naming the extra that restores the language:

```
warning: 2 .sql file(s) contributed nothing to the graph because a dependency is missing: tree_sitter_sql not installed. Install it with: pip install "graphifyy[sql]" (#1745)
```

A new `_EXTRA_FOR_EXTENSION` map lives next to `_DISPATCH` so the two stay in sight of each other. For a missing *core* grammar (broken env), the extension isn't in the map and the extractor's own error is printed verbatim with no extra hint.

## Tests

Mirrors the #1689 regression tests: one simulates the missing dependency (`monkeypatch.setitem(sys.modules, "tree_sitter_sql", None)`) and asserts the warning fires while a sibling `.py` file still extracts normally; one asserts no warning when the extra is installed. Full `tests/test_extract.py` passes (117/117), ruff clean.

### DIFF
--- a/graphify/extract.py
+++ b/graphify/extract.py
@@ -3742,6 +3742,20 @@ _DISPATCH: dict[str, Any] = {
 }
 
 
+# Extensions whose extractor depends on an optional-dependency extra
+# (pyproject [project.optional-dependencies]) and hard-fails without it,
+# rather than falling back like Pascal does. Used by the #1745 warning in
+# extract() to tell the user which extra restores the language.
+_EXTRA_FOR_EXTENSION = {
+    ".sql": "sql",
+    ".tf": "terraform",
+    ".tfvars": "terraform",
+    ".hcl": "terraform",
+    ".dm": "dm",
+    ".dme": "dm",
+}
+
+
 # Extensionless executables (CLI entry points like `devctl` or `manage`) carry
 # their language in the shebang, not the suffix. detect.classify_file already
 # routes them to the CODE path via _shebang_interpreter; _get_extractor must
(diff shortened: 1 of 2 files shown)


safishamsi:
Merged on `v8` in 8b7ffc5, authored to you. Clean, well-scoped fix.

I verified the diagnosis and the fix before landing:
- Confirmed `extract_sql` does set `error: "tree_sitter_sql not installed..."` and returns zero nodes when the grammar is absent, so the file falls through both existing warnings exactly as you describe — #1666 skips errored results, #1689 only covers no-extractor files. So the SQL corpus built "successfully" while contributing nothing.
- Your two tests pass (forced-missing grammar → warning fires with the `[sql]` extra named, Python still extracts; no false warning when installed), and the full suite is green (3120).
- Good call gating the `_EXTRA_FOR_EXTENSION` map behind an actual "not installed" error — it can't mislabel a language that has a working fallback (e.g. Pascal), since those never emit that error.

Thanks for the precise repro and the minimal surface.

PR #1761, opened 2026-07-10
fix(pg_introspect): read FKs from pg_constraint so read-only roles get 'references' edges (#1746)

### LINKED ISSUE #1746: --postgres introspection emits no FK 'references' edges (only contains/reads_from)
## Summary
Live PostgreSQL introspection (`--postgres`) produces **no foreign-key `references` edges** — only `contains` and `reads_from`. The README/docs state tables, views, functions **+ FK relationships** are mapped.

## Environment
- graphifyy 0.9.7 (pipx, with `[sql]` extra + injected `psycopg[binary]`), Python 3.14.5, macOS (arm64)
- PostgreSQL 15 (Supabase-hosted), connected via `PG*` env vars (`--postgres ""`)

## Reproduction
```bash
export PGHOST=… PGPORT=5432 PGDATABASE=postgres PGUSER=… PGPASSWORD=…
graphify extract . --postgres "" --out /tmp/pg-graph --no-cluster
```
Result on a schema with ~180 tables and several hundred FK constraints:
- `[graphify extract] PostgreSQL: 460 nodes, 706 edges` → written graph: 460 nodes / 642 edges
- Edge relation counts in graph.json: `{'contains': 459, 'reads_from': 183}` — **zero `references`**
- Concrete example: a table with ~15 inbound FKs shows only `db --contains-->` plus 3 `function --reads_from-->` edges.

For comparison, file-based extraction of the same schema's migration `.sql` files *does* produce `references` edges between tables — so the gap appears specific to the live-introspection path (`pg_introspect.py`).

The function→table `reads_from` edges are genuinely useful, by the way — happy to share redacted edge stats if helpful.

rithyKabir:
Fixes #1746.

## Root cause

`introspect_postgres` reads foreign keys from `information_schema.referential_constraints` — but per the [PostgreSQL docs](https://www.postgresql.org/docs/15/infoschema-referential-constraints.html):

> Only those constraints are shown for which the current user has write access to the referencing table (by way of being the owner or having some privilege other than `SELECT`).

A read-only introspection role — exactly what hosted providers like Supabase hand out, and what `pg_introspect`'s own `READ ONLY` transaction encourages — therefore gets **zero FK rows**, while tables/views/routines all still appear (SELECT is enough for those views). The graph then builds "successfully" with `contains`/`reads_from` edges but not a single `references` edge, matching the reported counts (459 contains + 183 reads_from, 0 references). This is also why file-based extraction of the same schema's migrations *does* produce `references` edges — the gap is specific to the live-introspection path.

Reproduced live on PostgreSQL 16 with a SELECT-only role: `referential_constraints` → 0 rows; `pg_catalog.pg_constraint` (`contype='f'`) → all 3.

## Fix

Query `pg_catalog.pg_constraint` instead — system catalogs are not privilege-filtered. The query returns the same 7-column shape, so DDL generation is untouched.

This also fixes a latent second bug: PostgreSQL constraint names are only unique **per table**, and the old name-based `key_column_usage` joins cross-matched same-named constraints on sibling tables. Verified live: two tables each with a constraint named `fk_owner` produced `columns = {user_id,user_id,user_id,user_id}` (malformed FK DDL) even for a privileged user. `pg_constraint` keys by oid, and `unnest(conkey) WITH ORDINALITY` preserves composite-FK column order.

## Verification

- Live PostgreSQL 16 (Docker), schema with a single-column FK, a composite FK, and same-named `fk_owner` constraints on two sibling tables, introspected as a SELECT-only role: all 3 `references` edges emitted end-to-end, composite FK produces exactly one edge, both `fk_owner` FKs attributed to the correct tables.
- New regression test asserts the FK query no longer touches `referential_constraints` and still yields the `references` edge; the test mock now records executed queries.
- `tests/test_pg_introspect.py` 7/7, plus `test_extract.py` + `test_multilang.py` (172 total) green; ruff clean.

### DIFF
--- a/graphify/pg_introspect.py
+++ b/graphify/pg_introspect.py
@@ -58,33 +58,42 @@ def introspect_postgres(dsn: str | None = None) -> dict:
             """)
             routines = cur.fetchall()
 
-            # 4. Query foreign keys — grouped by constraint to handle composites
+            # 4. Query foreign keys — grouped by constraint to handle composites.
+            # Read pg_catalog.pg_constraint, NOT information_schema.referential_
+            # constraints: that view only shows constraints where the current
+            # user has WRITE access to the referencing table (owner or a
+            # privilege other than SELECT), so a read-only introspection role
+            # sees zero FK rows while tables/views/routines all appear — the
+            # graph then silently loses every 'references' edge (#1746).
+            # pg_constraint is not privilege-filtered. It also keys constraints
+            # by oid rather than by name (constraint names are only unique per
+            # table, so the old name-based key_column_usage joins could
+            # cross-match same-named constraints on sibling tables).
             cur.execute("""
                 SELECT
-                    tc.constraint_name,
-                    kcu1.table_schema,
-                    kcu1.table_name,
-                    ARRAY_AGG(kcu1.column_name ORDER BY kcu1.ordinal_position) AS columns,
-                    kcu2.table_schema AS foreign_table_schema,
-                    kcu2.table_name AS foreign_table_name,
-                    ARRAY_AGG(kcu2.column_name ORDER BY kcu2.ordinal_position) AS foreign_columns
-                FROM
-                    information_schema.table_constraints AS tc
-                    JOIN information_schema.referential_constraints AS rc
-                      ON tc.constraint_name = rc.constraint_name
-                      AND tc.table_schema = rc.constraint_schema
-                    JOIN information_schema.key_column_usage AS kcu1
-                      ON tc.constraint_name = kcu1.constraint_name
-                      AND tc.table_schema = kcu1.table_schema
-                    JOIN information_schema.key_column_usage AS kcu2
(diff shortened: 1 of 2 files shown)


safishamsi:
Merged on v8, authored to you. The root cause is exactly right: `information_schema.referential_constraints` is privilege-filtered to write-access roles, so a read-only introspection role sees tables/views (SELECT suffices) but zero FKs — silent loss of every `references` edge. Switching to `pg_catalog.pg_constraint` (world-readable) is the correct fix, and keying by oid instead of constraint name also closes the same-named-constraint-on-sibling-tables cross-match in the old joins. `UNNEST ... WITH ORDINALITY` correctly preserves composite-FK column order. Mock tests pass (7) and full suite green (3139). I couldn't run it against a live Postgres here, but the catalog query is standard and the mock coverage + review cover it. Thanks for the sharp diagnosis.

PR #1777, opened 2026-07-10
feat(sql): extract table-to-table lineage from INSERT/CTAS/MERGE/UPDATE (#1572)

### LINKED ISSUE #1572: Feature: extract table-to-table dependencies from INSERT/CTAS/MERGE, not just CREATE VIEW
## What I'm trying to do

Build a table-to-table dependency (lineage) graph from a SQL/ETL codebase — "table A is populated from tables B and C".

## Current behavior (v0.9.3, `extract_sql` in `extract.py`)

`.sql` is parsed via `tree-sitter-sql` and produces good schema structure, but dataflow edges are only walked (`_walk_from_refs`) inside `create_view`, `create_function`, and `create_procedure`. As a result:

- `CREATE TABLE … AS SELECT …` (CTAS) → emits the table node but **no** `references` edges to the source tables in its `SELECT`.
- `INSERT INTO tgt SELECT … FROM src …` → produces **nothing** (no target node, no source edges).
- `MERGE INTO … USING …` / `UPDATE … FROM …` → not handled.

On a real ETL repo (~700 `.sql` files, mostly `INSERT … SELECT` bodies), only ~16% of files yielded any node, and essentially zero source→target lineage edges.

## Requested feature

Extend `extract_sql` to emit `references`/`reads_from` edges from the FROM/JOIN tables to the write target for:

- `INSERT INTO <tgt> … SELECT … FROM/JOIN <src…>`
- `CREATE TABLE <tgt> AS SELECT … FROM/JOIN <src…>` (CTAS)
- `MERGE INTO <tgt> USING <src>` and `UPDATE <tgt> … FROM <src>`

i.e. reuse the existing `_walk_from_refs` logic for these statement types so a lineage graph can be built directly.

## Nice-to-have

- Register `.ddl` as a SQL/code extension (some projects keep `CREATE TABLE` schemas in `.ddl` files, currently ignored).
- Optional hook to normalize templated identifiers (e.g. `{project}.{dataset}.tbl` from Python `.format()`/Jinja) so placeholder table names resolve instead of becoming literal node labels.

Happy to help test against a real ETL corpus. Thanks for graphify!

rithyKabir:
Closes #1572.

## Problem

Dataflow edges were only walked inside `create_view` / `create_function` / `create_procedure`, so ETL corpora — mostly `INSERT INTO … SELECT` bodies — produced schema structure but essentially zero source→target lineage (the issue measured ~16% of 700 files yielding any node and no lineage edges).

## What this adds

`extract_sql` now emits `<target> --reads_from--> <source>` edges, reusing the existing `_walk_from_refs` traversal, for all four requested statement types:

- **`INSERT INTO tgt … SELECT … FROM/JOIN src`** — the target node is materialized (so plain `VALUES` inserts now at least produce a node, per the issue's "no target node" complaint), and sources come from the nested `from`/`join` walk. `VALUES` inserts have no `from` node, so they can't create false lineage.
- **`CREATE TABLE tgt AS SELECT …`** (CTAS) — the SELECT lives in a `create_query` child of `create_table`; walked exactly like `create_view` already is.
- **`MERGE INTO tgt USING src`** — tree-sitter-sql has no dedicated merge node: the keywords sit directly under `statement`, so this is handled at the statement loop. Only *direct* children are scanned for the target/source `object_reference`s (nested ON-clause fields never surface there), and `USING (SELECT … FROM src)` subqueries are covered by the nested `from` walk.
- **`UPDATE tgt SET … FROM src`** — the target is the `relation` directly under `update`, so `_walk_from_refs` (which only looks under `from`/`join`) never sees the target itself.

`_walk_from_refs` now skips self-loops: `INSERT INTO t SELECT … FROM t` (dedup/backfill patterns) reads its own target and previously would have emitted a `t → t` edge.

Also includes the issue's nice-to-have: **`.ddl` is registered as a SQL extension** — dispatch table, `CODE_EXTENSIONS`, the case-insensitive-identifier set, and the #1745 missing-extra warning map.

## Tests

8 new tests over a realistic ETL fixture (`tests/fixtures/sample_dml_lineage.sql`) covering each statement type plus the two negative cases (no false lineage from `VALUES`, no self-loop from self-referential backfill) and a no-dangling-edges check. `tests/test_multilang.py` 56/56; `test_extract.py` + `test_pg_introspect.py` green; ruff clean. Full-suite failures on my machine (4× `test_ollama_retry_cap`, 1 order-flaky labeling test) reproduce identically on clean `v8` — pre-existing and unrelated.

The remaining nice-to-have from the issue (normalizing templated identifiers like `{project}.{dataset}.tbl`) is not included — it needs a design decision about placeholder handling and feels like its own change.


febryo-sc:
A data point for why this matters beyond lineage, from a codebase that keeps every query in its
own `.sql` file (loaded by name from a single loader module, ~220 statement files plus ~50
migrations).

With `tree-sitter-sql` installed, `extract_sql` produced 312 nodes — but **259 of the 270 `.sql`
file nodes had degree 0**. The DDL path works (migrations yield table nodes and FK `references`
edges); the statement files yield a bare file node, because a file whose entire content is a
`SELECT` / `UPDATE` / `INSERT` has no `create_*` node for the walker to hit. In a
SQL-in-files layout that is the majority of the corpus, so the language looks extracted while
the part carrying the actual contracts is inert.

Table references from DML — the lineage this issue proposes — is exactly what would connect
them: each statement node gaining edges to the tables it reads and writes would take those 259
isolated nodes into the graph proper.

For what it is worth, the caller-side half is recoverable outside graphify: scanning the loader
module's `key: load('file.sql')` map and the `SQL.<key>` call sites gave 586 statement↔caller
edges deterministically, no grammar needed. It is the *inside* of the statement — which tables
and columns it touches — that only a parser can supply, which is the gap here.

PR #1778, opened 2026-07-10
fix(bash): link executed scripts across files

### LINKED ISSUE #1756: Bash extractor: `bash x.sh` and `./x.sh` produce no cross-file edge (only `source`/`.` do)
## Summary

`extract_bash` creates a cross-file edge only when a command is `source` or `.`. The two most common ways one shell script runs another — `bash other.sh` and `./other.sh` — produce no edge at all. In any repo where scripts invoke each other by execution rather than sourcing, every script ends up as an isolated pair of nodes (the file node plus its `<name> script` entrypoint), and the call topology is missing from the graph.

## Version

Reproduced on **0.9.11** (also present in 0.9.7). Python 3.14.

## Reproduction

Four files in an empty directory:

```bash
# b.sh
#!/bin/bash
echo hi

# exec_call.sh
#!/bin/bash
bash ./b.sh

# bare_call.sh
#!/bin/bash
./b.sh

# source_call.sh
#!/bin/bash
source ./b.sh
```

```python
from pathlib import Path
from graphify.extractors.bash import extract_bash

for f in ["exec_call.sh", "bare_call.sh", "source_call.sh"]:
    r = extract_bash(Path(f))
    rels = sorted({e.get("relation") or e.get("type") for e in r["edges"]})
    print(f, rels)
```

## Actual

```
exec_call.sh     ['contains']
bare_call.sh     ['contains']
source_call.sh   ['contains', 'imports_from']
```

## Expected

`exec_call.sh` and `bare_call.sh` should each yield a cross-file edge to `b.sh` — something like `invokes`, alongside the existing `imports_from` for `source`.

## Cause

`graphify/extractors/bash.py:64`

```python
_BASH_SOURCE_COMMANDS = frozenset({"source", "."})
```

The `command` branch (~line 164) checks `cmd in _BASH_SOURCE_COMMANDS` and returns otherwise. An exec-position script path is never considered.

## Suggested fix

In the same `command` branch, additionally treat as an invocation:

1. a `command_name` whose literal ends in `.sh` (covers `./b.sh`, `/abs/path/b.sh`);
2. the first word argument of `bash`, `sh`, `zsh`, `ksh`, `dash` (covers `bash ./b.sh`).

Resolve the path the same way `source` already does, and keep the existing `resolved.exists()` guard — it preserves the traversal protection noted in the B-1 comment, since an edge is only emitted for a file that is actually present in the tree.

Paths that cannot be resolved on disk are out of scope here; skipping them silently, as `source` already does, is fine.

## Impact

Any shell-heavy repository. `source` is the minority idiom for invoking a sibling script; `bash x.sh` and `./x.sh` are the common ones, and today both are invisible to the graph.

balloon72:
## Summary

Link Bash scripts that execute sibling scripts directly or through a common shell runner.

## Root cause

The Bash extractor only treated `source` and `.` as cross-file commands. Direct `./script.sh` calls and commands such as `bash ./script.sh` were ignored, leaving script call topology disconnected.

## Changes

- Emit extracted `calls` edges from the current entrypoint or function to the target script entrypoint.
- Resolve literal `.sh` paths for direct execution and `bash`, `sh`, `zsh`, `ksh`, and `dash` runners.
- Skip dynamic, missing, and user-function-shadowed invocations.
- Cover absolute and relative inputs with regression tests.

## Validation

- `python -m pytest tests/test_extract.py -k "script_invocation" -q` (`6 passed`)
- `python -m pytest tests/test_extract.py -q -k "not collect_files_follows_symlinked_directory and not collect_files_skips_out_of_root_symlinked_directory and not collect_files_skips_out_of_root_symlinked_file_by_default and not collect_files_handles_circular_symlinks"` (`119 passed, 1 skipped`)
- `python -m py_compile graphify/extractors/bash.py tests/test_extract.py`
- `python -m graphify update .`
- `git diff --check origin/v8...HEAD`

The full local suite was attempted but could not complete because Windows Store Python repeatedly crashed in existing `ProcessPoolExecutor` pipeline tests. GitHub CI remains the full-suite check.

Fixes #1756

### DIFF
--- a/graphify/extractors/bash.py
+++ b/graphify/extractors/bash.py
@@ -62,6 +62,7 @@ def extract_bash(path: Path) -> dict:
     add_edge(file_nid, entry_nid, "contains", 1)
 
     _BASH_SOURCE_COMMANDS = frozenset({"source", "."})
+    _BASH_SCRIPT_RUNNERS = frozenset({"bash", "sh", "zsh", "ksh", "dash"})
     # Parent node types that mean a contained command is part of a substitution
     # or expansion, not a real function call. Token-level filtering misses
     # these because `$(build)` exposes `build` as a child command whose name
@@ -161,11 +162,11 @@ def extract_bash(path: Path) -> dict:
                 cmd_name_node = node.children[0]
             if cmd_name_node:
                 cmd = literal(cmd_name_node)
+                args = [c for c in node.children
+                        if c.type in ("word", "string", "concatenation")
+                        and c != cmd_name_node]
                 if cmd in _BASH_SOURCE_COMMANDS and cmd not in defined_functions:
                     # find the path argument (first word after command name)
-                    args = [c for c in node.children
-                            if c.type in ("word", "string", "concatenation")
-                            and c != cmd_name_node]
                     if args:
                         raw = _read_text(args[0], source).strip().strip("'\"")
                         line = node.start_point[0] + 1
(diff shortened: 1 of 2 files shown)


safishamsi:
Merged on v8, authored to you. Verified end-to-end (not just the unit level): on the four-file repro from #1756, both `bash ./b.sh` and `./b.sh` now emit `script_invocation` calls edges to b.sh's entry node, and — importantly — those targets are real nodes in the built graph, so they survive build rather than dangling and being silently dropped. Good defensive touches: the on-disk `is_file()` check (no phantom edges) and the shadowed-function skip. 26 bash tests pass, full suite 3153. Thanks.

Issue #1768, opened 2026-07-10
suggest_questions() isolated-node count (757) disagrees with report.py Knowledge Gaps count (245) — inflated by rationale/JSON-leaf/TS-Props nodes


Riaan180390:
## Summary

`suggest_questions()`'s "isolated/weakly-connected nodes" question overcounts by ~3x compared to the report's own "Knowledge Gaps" section, because it's missing filters that `report.py` already applies. The inflated number makes a healthy graph look like it has a major documentation/connectivity problem when it doesn't.

## Where

- `graphify/analyze.py`, `suggest_questions()`, the "Isolated or weakly-connected nodes" block (~line 504):
  ```python
  isolated = [
      n for n in G.nodes()
      if G.degree(n) <= 1 and not _is_file_node(G, n) and not _is_concept_node(G, n)
  ]
  ```
- `graphify/report.py`'s "Knowledge Gaps" section (~line 237) does the same computation but adds one more filter:
  ```python
  isolated = [
      n for n in G.nodes()
      if G.degree(n) <= 1
      and not _is_file_node(G, n)
      and not _is_concept_node(G, n)
      and G.nodes[n].get("file_type") != "rationale"
  ]
  ```

## Repro

On a 2213-node / 4862-edge real-world graph (Python + TS/React monorepo), the two computations disagree:

- `suggest_questions()` count (no rationale exclusion): **757**
- `report.py` "Knowledge Gaps" count (with rationale exclusion): **245**

Both appear in the *same generated report* (`GRAPH_REPORT.md`) — the Suggested Questions section says "757 weakly-connected nodes found — possible documentation gaps," while the Knowledge Gaps section a few paragraphs earlier says 245, for what's presented as the same concept. That's an internal inconsistency a reader will notice.

## Root cause of the inflation

I broke down the 757 by `file_type`:

| file_type | count | % | is it a real gap? |
|---|---:|---:|---|
| `rationale` | 512 | 68% | **No** — these are auto-extracted docstring/comment nodes attached to their parent symbol by exactly one edge (e.g. a node whose label is the first line of a function's docstring). Degree=1 is the *intended* shape for these, not a missing edge. |
[...]

PR #1780, opened 2026-07-10
fix(analyze): exclude rationale nodes from gap questions

balloon72:
## Summary

Keep the isolated-node count in suggested questions consistent with the report's Knowledge Gaps section.

## Root cause

`report.py` excludes `rationale` nodes from knowledge gaps because their expected shape is a single edge to the symbol they explain. `suggest_questions()` applied the other report filters but omitted that exclusion, so the same graph could show two different isolated-node counts.

## Changes

- Apply the existing `file_type != "rationale"` filter in `suggest_questions()`.
- Add a regression test that keeps a normal isolated node while excluding the rationale node from the count and question text.

This intentionally does not change cohesion scoring or add JSON/TypeScript naming heuristics.

## Validation

- `python -m pytest tests/test_analyze.py tests/test_report.py tests/test_cluster.py -q` (`74 passed`)
- `python -m py_compile graphify/analyze.py tests/test_analyze.py`
- `python -m graphify update .`
- `git diff --check origin/v8...HEAD`

The full local suite was not repeated because Windows Store Python crashed in existing `ProcessPoolExecutor` pipeline tests during the earlier full-suite attempt. GitHub CI remains the full-suite check.

Addresses the count mismatch in #1768.

### DIFF
--- a/graphify/analyze.py
+++ b/graphify/analyze.py
@@ -504,7 +504,10 @@ def suggest_questions(
     # 4. Isolated or weakly-connected nodes → exploration questions
     isolated = [
         n for n in G.nodes()
-        if G.degree(n) <= 1 and not _is_file_node(G, n) and not _is_concept_node(G, n)
+        if G.degree(n) <= 1
+        and not _is_file_node(G, n)
+        and not _is_concept_node(G, n)
+        and G.nodes[n].get("file_type") != "rationale"
     ]
     if isolated:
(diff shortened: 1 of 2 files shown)


safishamsi:
Merged on v8, authored to you. Confirmed parity: report.py's Knowledge Gaps filter (degree<=1, not file, not concept, file_type != rationale) now matches suggest_questions() exactly, so the two numbers in the same report agree. 49 analyze tests pass. Thanks — clean consistency fix.

PR #1785, opened 2026-07-11
fix(path): prefer full-token label matches when resolving path endpoints

CJNA:
## Problem

`graphify path` resolves each endpoint by committing to `_score_nodes()[0]`. The full-query bonus tier in `_score_nodes` only fires when the query **equals or prefixes** a label, so a query that is a token *subset* of the intended label gets no bonus, and a node prefix-matching a single rare token can out-score it on IDF alone.

Real-world repro (27k-node graph):

```
$ graphify path "Reject-everything judge" "Drift guard"
warning: source match was ambiguous (top score 82.5292, runner-up 82.5292)
No path found between 'Reject-everything judge' and 'Drift guard'.
```

The head of the scored list was an exact **3-way tie** of disconnected doc-island leaves ("Rejection Summary", "Rejected Alternatives", "Rejected / not surfaced (with reason)" — each matching only the `reject` token as a prefix), while the intended node "Degenerate Reject-Everything Judge" — which contains **every** query token — ranked **94th** at 1/24th the head score. The path anchored on a degree-1 leaf in another component → false "No path found".

## Fix

`_pick_scored_endpoint(G, scored, query)` in `serve.py`: scan the score-ordered list and take the **first candidate whose label contains every query token**; fall back to `scored[0]` when none does. When the head already full-matches (the common case — including every whole-query-tier hit), the pick is exactly `scored[0]`, so existing resolutions are unchanged.

Wired into both endpoint resolvers:
- `cli.py` `path` command
- `serve.py` `_tool_shortest_path` (MCP)

Also: the close-runner-up "ambiguous" warning is now emitted only when the picked endpoint *is* the raw score head — when the full-token override chose a different node, the head's margin didn't decide anything and the warning was misleading.

## After

```
$ graphify path "Reject-everything judge" "Drift guard"
Shortest path (3 hops):
  Degenerate Reject-Everything Judge --semantically_similar_to [INFERRED]--> Vacuous-Pass Guard <--references [EXTRACTED]-- Anchored, fail-closed detection ... <--semantically_similar_to [INFERRED]-- Drift guard
```

## Tests

- `test_endpoint_prefers_full_token_match` — minimal deterministic graph reproducing the misranking (decoy "Rejection Summary" out-scores the full match via IDF; verified the decoy genuinely wins under current scoring before the fix); asserts the path resolves through the full-match node.
- `test_endpoint_falls_back_to_score_head` — query with no full-token candidate behaves byte-identically to the old `scored[0]` pick.
- Full `tests/test_serve.py` + `tests/test_path_cli.py`: **84 passed**. `ruff check` clean on all touched files.

### DIFF
--- a/graphify/cli.py
+++ b/graphify/cli.py
@@ -602,7 +602,7 @@ def dispatch_command(cmd: str) -> None:
                 file=sys.stderr,
             )
             sys.exit(1)
-        from graphify.serve import _score_nodes
+        from graphify.serve import _pick_scored_endpoint, _score_nodes
         from networkx.readwrite import json_graph
         import networkx as _nx
 
@@ -635,7 +635,8 @@ def dispatch_command(cmd: str) -> None:
         if not tgt_scored:
             print(f"No node matching '{target_label}' found.", file=sys.stderr)
             sys.exit(1)
-        src_nid, tgt_nid = src_scored[0][1], tgt_scored[0][1]
+        src_nid = _pick_scored_endpoint(G, src_scored, source_label)
+        tgt_nid = _pick_scored_endpoint(G, tgt_scored, target_label)
         # Ambiguity guard: when both queries resolve to the same node, the
         # shortest path is trivially zero hops, which is almost never what the
         # caller wanted (see bug #828).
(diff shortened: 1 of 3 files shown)


safishamsi:
Merged on v8, authored to you. Verified on a synthetic version of your repro: with distractors prefix-matching only the `reject` token out-scoring the intended node, `path` used to anchor on `reject filler 0` (score head); `_pick_scored_endpoint` now correctly picks `Degenerate Reject-Everything Judge` (the full-token match), and a head-full-match query is unchanged (returns scored[0]). Nice call scoping the ambiguity warning to only fire when the head is what got picked. 80 serve tests + full suite (3156) green. Thanks.

PR #1790, opened 2026-07-11
fix(extract): emit inherits edge for JavaScript class extends

Synvoya:
## Bug

Plain JavaScript class heritage was dropped: `class Dog extends Animal {}` emitted no `inherits` edge, while the equivalent TypeScript did.

## Root cause

The JavaScript grammar stores `extends` and its identifier directly under `class_heritage`. TypeScript wraps the base in an `extends_clause`. The walker only handled the wrapped shape.

## Fix

When a `class_heritage` node has no `extends_clause` or `implements_clause`, treat the heritage node itself as the JavaScript extends clause. The guard preserves the existing TypeScript path and prevents duplicate edges.

The implementation remains precision-first: direct same-file and imported identifiers resolve; runtime expressions such as `extends mixin(Animal)` are not converted into fabricated static inheritance. Qualified member expressions require namespace-aware resolution and remain outside this focused fix.

## Validation

- Added same-file JavaScript inheritance coverage.
- Added imported JavaScript inheritance coverage.
- Added a dynamic-base negative control.
- Existing TypeScript interface, generic, imported and implements cases remain covered.
- `tests/test_ts_inheritance.py`: 10 passed.
- Ruff and `git diff --check`: passed.

### DIFF
--- a/graphify/extractors/resolution.py
+++ b/graphify/extractors/resolution.py
@@ -1363,19 +1363,34 @@ def _ts_walk_class_members(class_node, source: bytes, path: Path, class_nid: str
     line = class_node.start_point[0] + 1
     for child in class_node.children:
         if child.type == "class_heritage":
+            saw_clause = False
             for clause in child.children:
                 if clause.type == "extends_clause":
+                    saw_clause = True
                     for name in _ts_heritage_clause_entries(clause, source):
                         facts.uses.append(
                             _SymbolUseFact(path, class_nid, name, "inherits", "type",
                                            clause.start_point[0] + 1)
                         )
                 elif clause.type == "implements_clause":
+                    saw_clause = True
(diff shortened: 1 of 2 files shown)


Synvoya:
Follow-up audit against current v8 found that JavaScript member-expression parents such as class Dog extends ns.Animal were still dropped. The branch now handles that AST shape and adds a regression test. Current validation: tests/test_ts_inheritance.py 9 passed; Ruff and git diff --check passed.

Synvoya:
Precision follow-up: I removed the member-expression tail normalization mentioned in my previous comment. Treating ns.Animal as bare Animal can falsely bind an unrelated same-file class without namespace-aware resolution. The final branch stays focused on direct JavaScript heritage identifiers, adds imported-class coverage and a runtime-expression negative control, and is squashed to one commit.

safishamsi:

PR #1791, opened 2026-07-11
fix(php): capture interface/enum/trait heritage as class-like nodes

Synvoya:
## Bug

PHP interfaces, enums, and traits produce **no node** and **no heritage edges**. `interface B extends A`, `enum E implements I`, and `trait T { use U; }` are all silently dropped.

### Root cause

`_PHP_CONFIG.class_types` only contained `class_declaration`, so the generic extractor never treated `interface_declaration`, `enum_declaration`, or `trait_declaration` nodes as class-like. They never became nodes, and the PHP heritage handler (which runs per class-like node) never fired for them.

## Minimal repro

```php
<?php
interface Identifiable {}
interface Nameable extends Identifiable {}
enum Suit: string implements Nameable {
    case Hearts = 'H';
    public function label(): string { return 'x'; }
}
trait Named {}
trait Greets { use Named; }
```

Before: no `Nameable`/`Suit`/`Greets` nodes, no `inherits`/`implements`/`mixes_in` edges.

## Fix

- Add `interface_declaration`, `enum_declaration`, and `trait_declaration` to `_PHP_CONFIG.class_types`.
- Enums wrap members in an `enum_declaration_list` (not a `declaration_list`), so add it to `body_fallback_child_types` so enum methods/cases are walked.

After the fix the repro yields:
- `Nameable --inherits--> Identifiable`
- `Suit --implements--> Nameable`
- `Greets --mixes_in--> Named`
- `Suit --method--> label` (enum body now walked)

## Testing

- Added `test_php_interface_enum_trait_heritage` to `tests/test_languages.py`, mirroring the existing `test_php_splits_inherits_implements_mixes_in` assertion style. Fails before, passes after.
- `tests/test_languages.py`: 315 passed, 13 skipped (was 314 + this new test).

### DIFF
--- a/graphify/extract.py
+++ b/graphify/extract.py
@@ -796,7 +796,14 @@ _SCALA_CONFIG = LanguageConfig(
 _PHP_CONFIG = LanguageConfig(
     ts_module="tree_sitter_php",
     ts_language_fn="language_php",
-    class_types=frozenset({"class_declaration"}),
+    # interfaces/enums/traits are class-like containers whose heritage
+    # (extends/implements) and members must be captured, same as classes.
+    class_types=frozenset({
+        "class_declaration",
+        "interface_declaration",
+        "enum_declaration",
+        "trait_declaration",
+    }),
     function_types=frozenset({"function_definition", "method_declaration"}),
     import_types=frozenset({"namespace_use_clause"}),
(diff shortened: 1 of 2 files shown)


Lautaroscu:
good job

safishamsi:

PR #1792, opened 2026-07-11
fix(scala): capture trait declarations as class-like nodes

Synvoya:
## Bug

Scala `trait` declarations produce **no node** and **no heritage edges**. A `trait` that `extends`/`with`-mixes other types loses all of that structure.

### Root cause

`_SCALA_CONFIG.class_types` contained only `class_definition` and `object_definition`. The tree-sitter-scala grammar parses a trait as a `trait_definition` node, which was therefore never treated as class-like — so it never became a node and the Scala heritage handler never ran for it.

## Minimal repro

```scala
trait Greeter extends Base with Logging { def greet(): Unit }
```

Before: no `Greeter` node, no `inherits`/`mixes_in` edges.

## Fix

Add `trait_definition` to `_SCALA_CONFIG.class_types`. The existing Scala `extends_clause` handler already splits the first base (`inherits`) from subsequent `with` mixins (`mixes_in`), so once the trait becomes a class-like node the edges are emitted correctly.

After the fix:
- `Greeter --inherits--> Base`
- `Greeter --mixes_in--> Logging`

## Testing

- Added `test_scala_trait_definition_heritage` to `tests/test_languages.py`, mirroring the existing `test_scala_splits_inherits_and_mixes_in`. Fails before, passes after.
- `tests/test_languages.py`: 315 passed, 13 skipped.

### DIFF
--- a/graphify/extract.py
+++ b/graphify/extract.py
@@ -780,7 +780,9 @@ _KOTLIN_CONFIG = LanguageConfig(
 
 _SCALA_CONFIG = LanguageConfig(
     ts_module="tree_sitter_scala",
-    class_types=frozenset({"class_definition", "object_definition"}),
+    # traits are class-like containers with their own heritage (extends / with),
+    # so they need a node and the heritage walk just like classes and objects.
+    class_types=frozenset({"class_definition", "object_definition", "trait_definition"}),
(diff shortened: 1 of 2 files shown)


safishamsi:

PR #1793, opened 2026-07-11
fix(kotlin): resolve qualified supertype to its tail type name

Synvoya:
## Bug

A Kotlin class with a **qualified** supertype resolves its heritage edge to the package root instead of the type. `class Foo : com.example.Base()` emits `Foo --inherits--> com` instead of `Foo --inherits--> Base`, collapsing every qualified base/interface onto a bogus `com` node.

### Root cause

`_kotlin_user_type_name` returned the **first** identifier child of a `user_type` node. The tree-sitter-kotlin grammar stores a qualified name as flat identifier children:

```
user_type "com.example.Base"
  identifier "com"
  .
  identifier "example"
  .
  identifier "Base"
```

So the walker returned `com`.

## Minimal repro

```kotlin
class Foo : com.example.Base(), com.example.Iface
```

Before: `inherits {(Foo, com)}`, `implements {(Foo, com)}`.

## Fix

Scan the direct children and keep the **last** identifier/`type_identifier` segment. Type arguments live in a separate `type_arguments` child (not an identifier), so generics like `com.ex.Container<String>` correctly resolve to `Container`. This mirrors the existing C++ qualified-base handling that uses the unqualified tail. Simple (`Base`) and delegated (`Bar by baz`) supertypes are unaffected — a single-segment `user_type` still returns that segment.

After the fix:
- `Foo --inherits--> Base`
- `Foo --implements--> Iface`

## Testing

- Added `test_kotlin_qualified_supertype_uses_tail_name` to `tests/test_languages.py`. Fails before, passes after.
- `tests/test_languages.py`: 315 passed, 13 skipped — the existing Kotlin simple-supertype and interface-delegation tests still pass.

### DIFF
--- a/graphify/extractors/engine.py
+++ b/graphify/extractors/engine.py
@@ -608,22 +608,31 @@ def _php_method_return_type_node(method_node):
     return None
 
 def _kotlin_user_type_name(user_type_node, source: bytes) -> str | None:
-    """Return the head identifier text from a Kotlin user_type node (without generics)."""
+    """Return the tail identifier text from a Kotlin user_type node (without generics).
+
+    A qualified supertype like `com.example.Base` lists its segments as flat
+    `identifier` children (`com`, `example`, `Base`) separated by `.` tokens, so
+    the real type is the LAST segment, not the first — returning the head yielded
+    the package root (`com`). Type arguments live in a separate `type_arguments`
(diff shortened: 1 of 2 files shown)


safishamsi:

PR #1794, opened 2026-07-11
fix(scala): emit heritage edges for qualified extends/with clauses

Synvoya:
## Bug

Scala qualified parents were either dropped or retained as package-qualified labels. For example, `class Foo extends pkg.Base with other.Trait` should emit `Foo -> Base` and `Foo -> Trait`.

## Root cause

The heritage walker handled `type_identifier` and a narrow `generic_type` shape. Qualified parents parse as `stable_type_identifier`; qualified generic parents wrap that node inside `generic_type`.

## Fix

Normalize heritage nodes through one bounded helper:

- plain `type_identifier` returns its name;
- `stable_type_identifier` returns its final type segment;
- `generic_type` recursively normalizes its base type.

Only direct heritage children are considered, so package segments, generic arguments and constructor arguments cannot become parent edges. The first parent remains `inherits`; subsequent `with` parents remain `mixes_in`.

## Validation

- Covers simple qualified parents.
- Covers qualified generic parents.
- Covers multi-segment packages and constructor arguments.
- Verifies package qualifiers and generic arguments are not emitted as parents.
- `tests/test_languages.py`: 319 passed, 13 optional-DM skips.
- Ruff and `git diff --check`: passed.

### DIFF
--- a/graphify/extractors/engine.py
+++ b/graphify/extractors/engine.py
@@ -2797,18 +2797,39 @@ def _extract_generic(
                             break
                 if extend is not None:
                     bases: list[tuple[str, int]] = []
-                    for c in extend.children:
-                        if c.type == "type_identifier":
-                            bases.append((_read_text(c, source), c.start_point[0] + 1))
-                        elif c.type == "generic_type":
-                            base = c.child_by_field_name("type")
+
+                    def scala_base_name(type_node) -> str | None:
+                        if type_node.type == "type_identifier":
+                            return _read_text(type_node, source)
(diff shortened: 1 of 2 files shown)


Synvoya:
Follow-up audit against current v8 found that qualified generic parents such as pkg.Base[Int] still retained the package prefix through the generic_type path. The branch now normalizes both plain and generic qualified parents to the tail type name and extends the regression test. Current validation: tests/test_languages.py 319 passed, 13 skipped; Ruff and git diff --check passed.

safishamsi:

Issue #350, opened 2026-04-14
[Bug] Running `/graphify .` on a large repo triggers subdirectory prompt, causing AST cache to be written to `<subdir>/graphify-out/cache/` instead of `./graphify-out/cache/`


xiaoyuze88:
### Description

When graphify's built-in large-corpus flow prompts the user to select a subfolder (e.g. `src/`), it internally re-runs `detect()` and `extract()` on that subdirectory. The AST extraction cache is then silently written to `<subdir>/graphify-out/cache/` while `check_semantic_cache()` always reads from `./graphify-out/cache/`. The two paths never overlap, so **every subsequent run is a full cache miss** — all AST extraction is repeated from scratch even though nothing changed.

### Version

`graphifyy 0.4.13`

### Steps to Reproduce

> **Note: this bug is triggered by graphify's own built-in large-corpus flow, not by manually passing a subdirectory path.**

1. Run `/graphify .` on a project with more than 200 files
2. graphify detects the large corpus and — per its own built-in flow — asks the user to pick a subfolder to limit scope, presenting options like `src/`, `src/business/`, etc.
3. User selects `src`
4. graphify internally re-runs `detect()` and `extract()` on `src/`

After the run completes, two separate cache directories exist:

```
./graphify-out/cache/          ← empty (where check_semantic_cache looks)
./src/graphify-out/cache/      ← 829 files (where extract() actually wrote)
```

**This means the bug is triggered by graphify's own recommended workflow for large repos.** Any project large enough to hit the "pick a subfolder" prompt will silently produce a broken cache on every run.

### Root Cause

`extract()` infers `root` from the common prefix of the input file paths:

```python
# graphify/extract.py ~line 3006
common_len = sum(
    1 for i in range(min(len(p.parts) for p in paths))
    if len({p.parts[i] for p in paths}) == 1
)
root = Path(*paths[0].parts[:common_len]) if common_len else Path(".")
```

When all input files are under `src/`, this infers `root = src`. Cache entries are then written to `src/graphify-out/cache/`.

However, `check_semantic_cache()` and `save_semantic_cache()` both default to `root=Path(".")`, so they always read from and write to `./graphify-out/cache/`.

The two functions use **different roots independently** — there is no shared root value passed through the call chain.

### Impact

- Cache is never hit on the second run, wasting tokens on re-extraction
- A stray `<subdir>/graphify-out/` directory is created inside the user's source tree
[...]

Issue #1747, opened 2026-07-09
graphify-out/ written to corpus/CWD despite --out / --graph pointing elsewhere


bbqboogiedwonsen:
## Summary
Output artifacts land in the corpus root / CWD even when `--out <dir>` (or `--graph <path>`) points elsewhere. Two distinct cases observed on 0.9.7:

## Case 1 — `extract --out` still creates `graphify-out/` inside the corpus
```bash
cd <repoA>
graphify extract . --out /tmp/scratch/tenant --no-cluster
```
The requested output is written to `/tmp/scratch/tenant/graphify-out/` ✔ — but a stray `graphify-out/` also appears in the corpus root (observed containing `cache/` and partial content; also reproduced on a run that aborted at the no-LLM-key gate, which left `<corpus>/graphify-out/` behind despite `--out` pointing elsewhere).

## Case 2 — `cluster-only --graph <elsewhere>` writes results to CWD instead
```bash
cd <repoA>
graphify cluster-only . --no-label --graph /tmp/scratch/opexone/graphify-out/graph.json
```
`GRAPH_REPORT.md`, `.graphify_analysis.json`, `.graphify_labels.json` and the **clustered** `graph.json` were all written to `./graphify-out/` (CWD); the `--graph` target file remained un-clustered (community count 0). I.e. the flag selects the *input*, but outputs ignore its location.

## Environment
- graphifyy 0.9.7 (pipx), Python 3.14.5, macOS (arm64)

## Expected
All outputs follow `--out` / live next to `--graph`. Impact: repo pollution and accidental commits — we had to gitignore `graphify-out/` in every corpus repo as a workaround.

safishamsi:
Fixed on v8 in c5db9ff (0.9.12). Both cases, verified with regression tests:

**Case 1** (`extract --out` leaving a stray `graphify-out/` in the corpus): the graph itself already went to `--out`, but `detect()`'s word-count/stat-index cache used the scan root. It now honors `--out` via a threaded `cache_root`, so the corpus is left untouched. Confirmed: `extract <corpus> --out <dir> --code-only` writes graph **and** cache under `<dir>/graphify-out/`, with no `graphify-out/` created in the corpus.

[...]

PR #1802, opened 2026-07-11
fix(extract): default the AST cache location to CWD, not the analyzed source tree (#1774)

### LINKED ISSUE #1774: extract() writes AST cache into the analyzed source tree, bypassing CWD (pollutes read-only corpora)
## Summary
`graphify.extract.extract()` writes its per-file AST cache into a
`graphify-out/cache/` directory placed **inside the analyzed source tree**,
not under the current working directory. When graphify is run to analyze a
corpus you don't want to mutate (a read-only knowledge base, a git working
tree, someone else's repo), this silently creates files inside that corpus,
defeating any attempt to isolate outputs via CWD.

## Version
- `graphifyy==0.1.14` (installed), source @ `91f4d120b630ee35c79bf3c75ccd186870a808f9`
- Python 3.12

## Reproduction
```python
from pathlib import Path

# CWD is /tmp/work  (where I want all outputs)
extract([Path("/data/corpus/a.py"), Path("/data/corpus/b.py")])

# Observed: /data/corpus/graphify-out/cache/<hash>.json   <-- pollutes source
# Expected: cache under CWD (/tmp/work/graphify-out/cache/) or a caller-provided dir
```

## Root cause
`extract()` takes only `paths` and has no cache-location parameter, so it
derives a `root` from the **common parent of the input files** and passes
that to the cache layer:

- `extract.py:2256` `def extract(paths: list[Path]) -> dict:`  (no cache-root arg)
- `extract.py:2266-2279` `root` = common parent of `paths` (falls back to `.`)
- `extract.py:2285 / 2291` `load_cached(path, root)` / `save_cached(path, result, root)`
- `cache.py:14-18` `cache_dir(root)` -> `root / "graphify-out" / "cache"` (mkdir + write)

Note `cache_dir`'s own default is `Path(".")` (CWD), which is the sane
behavior - but `extract()` overrides it with the source-tree root, so the
cache follows the *input* location instead of the *output* location.

## Impact
- Read-only / analyze-only runs mutate the target corpus.
- In a git working tree this leaves untracked files (dirty status, risk of
  accidental commit).
- Recurs on every re-run; output-isolation via CWD does not prevent it.

## Suggested fix (backward-compatible)
Give `extract()` an optional cache location and default it to CWD rather
than the input tree:

```python
def extract(paths: list[Path], cache_root: Path | None = None) -> dict:
    root = cache_root if cache_root is not None else Path(".")
    ...
```

This keeps caching (the perf win of skipping unchanged files) while letting
callers keep all writes under CWD. The same `root` coupling exists in the
semantic-cache helpers (`cache.check_semantic_cache` / `save_semantic_cache`),
which would benefit from the same treatment.

SimiSips:
Fixes #1774.

## Problem

`extract()` called without `cache_root` derives the cache root from the inferred common parent of the input files — i.e. the analyzed source tree. Analyzing a corpus outside CWD (a read-only knowledge base, someone else's checkout) silently creates `graphify-out/cache/` inside that corpus, defeating output isolation via CWD. Reproduced on current `v8` (0.9.13). The CLI and watcher are unaffected — they already pass `cache_root` explicitly (added in 01fb51b for #350) — so this only bites library callers and `python -m graphify.extract`.

## Fix

One line: `effective_root` (the cache location) now defaults to `Path('.')` when `cache_root` is `None`, matching `cache_dir()`'s own default. An explicit `cache_root` still wins.

Deliberately **not** the patch suggested in the issue (defaulting `root` itself to CWD): `root` also anchors node-id relativization (`path.relative_to(root)`) and `_augment_symbol_resolution_edges()`, and pointing it at CWD would degrade ids for corpora outside CWD. Only the cache side changes; id/symbol anchoring is untouched.

## Notes

- `file_hash()` keys include the path relative to the cache root, so existing source-tree caches of **library callers** go cold once and re-extract (CLI/watch caches are unaffected). No correctness impact.
- Cache entries for out-of-root files keep absolute `source_file` paths — an already-handled pass-through in `_relativize_source_files_in` / `_absolutize_source_files_in`.

## Tests

- New `tests/test_extract_cache_location.py`: default cache lands in CWD and the corpus stays untouched; the CWD cache round-trips on a second run; explicit `cache_root` still wins.
- Full suite: `uv run pytest tests/ -q` — **3162 passed, 3 skipped** on macOS (Python 3.12).

### DIFF
--- a/graphify/extract.py
+++ b/graphify/extract.py
@@ -4245,9 +4245,12 @@ def extract(
 
     Args:
         paths: files to extract from
-        cache_root: explicit root for graphify-out/cache/ (overrides the
-            inferred common path prefix). Pass Path('.') when running on a
-            subdirectory so the cache stays at ./graphify-out/cache/.
+        cache_root: explicit root for graphify-out/cache/. Defaults to the
+            current working directory — the cache is an output, so it must
+            not land inside the analyzed source tree (#1774). The inferred
+            common path prefix of ``paths`` still anchors node ids and
+            symbol resolution; it just no longer decides where the cache
+            is written.
         parallel: if True and there are >= _PARALLEL_THRESHOLD uncached files,
             use ProcessPoolExecutor for multi-core extraction.
         max_workers: max subprocess count. Defaults to cpu_count (or the
@@ -4281,7 +4284,10 @@ def extract(
         root = cache_root
     root = root.resolve()
 
-    effective_root = cache_root or root
+    # #1774: the cache is an output, so it defaults to CWD — never the
+    # analyzed source tree. `root` (inferred above) still anchors node ids
+    # and symbol resolution; only the cache location diverges from it.
+    effective_root = cache_root if cache_root is not None else Path(".")
     total = len(paths)
 
     # Phase 1: separate cached hits from uncached work
--- a/tests/test_extract_cache_location.py
+++ b/tests/test_extract_cache_location.py
@@ -1,0 +1,64 @@
+"""#1774 — extract() must never write its AST cache into the analyzed source tree.
+
+The cache is an output. When no cache_root is given it used to default to the
+inferred common parent of the input files — the source tree — so analyzing a
+read-only corpus (someone else's repo, a knowledge base) silently created
+graphify-out/cache/ inside it. It now defaults to the current working
+directory, matching cache_dir()'s own default; an explicit cache_root still
+wins.
+"""
+from __future__ import annotations
+
+from pathlib import Path
+
+import graphify.extract as ex
+from graphify.cache import load_cached
+
+
+def _make_corpus(base: Path) -> Path:
+    corpus = base / "corpus"
+    corpus.mkdir()
+    (corpus / "a.py").write_text("class Base:\n    def hello(self):\n        return 1\n")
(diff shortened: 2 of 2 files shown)


safishamsi:
Thanks @SimiSips — you had the right diagnosis and the right default (the cache is an output; it belongs at CWD, not inside the analyzed tree). Landed in `c3a42a3` (ships in 0.9.14), crediting you, and I kept your CWD-default and the two location tests.

I extended it because `effective_root` was doing three jobs behind one name: cache-dir location, `file_hash` key anchoring, and the XAML/C# project-scan boundary. Pointing that single value at CWD fixes the location but, for a corpus that lives outside CWD, makes `file_hash` fall back to absolute keys (so shared/CI cache reuse silently breaks) and makes the XAML `.cs` scan run over CWD instead of the corpus. So I split the two concepts: `load_cached`/`save_cached` take a new `cache_root` for *where the dir lives*, while `root` (the inferred common parent) still anchors keys, node ids and the XAML boundary. The parallel worker tuple carries both.

Net: your behavior for the in-CWD case is preserved, and out-of-CWD corpora keep portable relative keys. I added a regression test asserting keys stay relative for a corpus outside CWD (the property the one-liner would have lost). Also mirrors the location/anchor split that #1747 introduced for the stat index. Appreciate the fix.

PR #1806, opened 2026-07-11
hooks: skip rebuild in linked git worktrees

Claude-Madera:
## Problem

The `post-commit` and `post-checkout` hooks rebuild `graphify-out/` on every commit and branch switch. Inside a **linked worktree** (`git worktree add`) this misbehaves:

- The canonical `graphify-out/` belongs to the **primary checkout**; rebuilding it from every worktree is wasteful.
- Deploy/CI flows that create a worktree, switch/checkout, then `git clean` it, race the detached background rebuild against the cleanup — producing:

  ```
  failed to remove graphify-out/: Directory not empty
  ```

This surfaced in a deploy "cockpit" that pins release worktrees of three repos: every pin fired a pointless full rebuild and intermittently failed the `git clean`.

## Fix

Guard both hooks: a linked worktree has `git-dir != git-common-dir`, so short-circuit before launching the rebuild.

Key correctness detail: the compare uses **absolute** paths —

```sh
_GFY_GITDIR=$(git rev-parse --absolute-git-dir 2>/dev/null)
_GFY_COMMONDIR=$(cd "$(git rev-parse --git-common-dir 2>/dev/null)" 2>/dev/null && pwd)
if [ -n "$_GFY_COMMONDIR" ] && [ "$_GFY_GITDIR" != "$_GFY_COMMONDIR" ]; then
    exit 0
fi
```

Git exports `GIT_DIR` to hooks and it is sometimes absolute; comparing a raw `$GIT_DIR` against a relative `--git-common-dir` (`.git`) would false-positive and wrongly skip the **primary** checkout. Resolving both to absolute paths avoids that. Factored into a shared `_WORKTREE_GUARD` constant, mirroring the existing `_PYTHON_DETECT` pattern.

## Tests

- `test_hooks_skip_linked_worktrees` — both generated scripts contain the guard exactly once and use the absolute-path compare.
- `test_worktree_guard_skips_only_linked_worktree` — end-to-end against a real `git worktree`: the guard runs through on the primary checkout and exits early in a linked worktree.

Both pass; no previously-passing tests affected.

### DIFF
--- a/graphify/hooks.py
+++ b/graphify/hooks.py
@@ -14,7 +14,7 @@ _CHECKOUT_MARKER_END = "# graphify-checkout-hook-end"
 # __PINNED_PYTHON__ is replaced at install time with the absolute path of the
 # Python interpreter that ran `graphify hook install`.  For uv-tool and pipx
 # installs the interpreter lives inside an isolated venv, so the launcher on
-# PATH is the only entry point — and GUI git clients / CI runners often have a
+# PATH is the only entry point -- and GUI git clients / CI runners often have a
 # minimal PATH that omits ~/.local/bin.  Pinning sys.executable at install time
 # makes the hook work regardless of PATH at git-trigger time.
 _PYTHON_DETECT = """\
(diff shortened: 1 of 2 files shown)


safishamsi:
Thanks for this, @Claude-Madera — the worktree diagnosis and especially the absolute-path detail (raw `$GIT_DIR` can be absolute while `--git-common-dir` is the relative `.git`, which would false-positive the primary checkout) were exactly right, and I've carried that guard into `5f57c46` (shipping in 0.9.14).

I landed it directly rather than merging so the same commit could also close the two related gaps from #1809 that this PR didn't cover: `post-checkout` not honoring `GRAPHIFY_SKIP_HOOK`, and the `.git/info/exclude` bloat in #1810. A couple of small deltas from your version: the guard normalizes both paths with `cd ... && pwd` instead of `git rev-parse --absolute-git-dir` (works on older git without the newer flag), and I left the existing comment em-dashes alone to keep the diff focused on behavior. Credited you in the commit trailer. Really appreciate the PR.

Issue #1807, opened 2026-07-11
CLI exits 255 (unhandled BrokenPipeError) when output is truncated by head / Select-Object-First


varuntej07:
_**TL;DR:** piping graphify query into anything that reads only the first few lines makes graphify crash with exit code 255 after the reader disconnects, so scripts and CI treat successful queries as failures — it should treat an early-disconnecting reader as success and exit 0._

## Summary
When a downstream consumer closes graphify's stdout early, the CLI crashes with an unhandled write-to-closed-pipe error and exits 255 after emitting partial output. Truncating a CLI's output is routine (`head`, PowerShell's `Select-Object -First N`, `sed q`), and the conventional behavior is to treat a closed pipe as "the reader has what it needs," not as failure.

## Repro
graphifyy 0.9.12 (uv tool install), Windows 11 + PowerShell 5.1, and the equivalent on Unix:

     graphify query "anything" | Select-Object -First 2
     $LASTEXITCODE     # -> -1 (raw exit code 255); $? is False

     # POSIX:
     graphify query "anything" | head -n 2; echo $?    # -> 255 (or 1)

**Actual**: partial output prints, process exits 255 ($LASTEXITCODE reports -1).
**Expected**: exit 0 (or the SIGPIPE 141 convention on POSIX).
**_Note_**: the truncated output prints normally and the shell shows no error - the only symptom is the exit code, so you must check `$LASTEXITCODE` / `$?` (or `echo $?` on POSIX) to see it. Reproduces regardless of output size.

Without truncation, `graphify query` exits 0 with all output on stdout, so this only bites pipelines - but it breaks any script or CI wrapper that both trims output and checks exit codes, and it makes agent harnesses (Claude Code etc.) render successful query results as command failures.

## Suggested fix
Wrap the console entry point (`graphify/__main__.py: main()`) in the standard CPython pattern (https://docs.python.org/3/library/signal.html#note-on-sigpipe):

    def main() -> None:
        try:
            _main()          # current main() body, renamed
        except BrokenPipeError:
            # Reader closed the pipe (head, Select-Object -First N): they
            # have what they need. Redirect stdout to devnull so the
            # interpreter's shutdown flush doesn't raise a second time.
            devnull = os.open(os.devnull, os.O_WRONLY)
            os.dup2(devnull, sys.stdout.fileno())
            sys.exit(0)

[...]

PR #1811, opened 2026-07-11
fix: exit 0 when a downstream reader closes stdout early

varuntej07:
## Problem

Piping graphify into a reader that stops early (`head`, PowerShell
`Select-Object -First N`, a pager that quits) closes stdout before we finish
writing. The interpreter's shutdown flush then hits the closed pipe and crashes:

```
Exception ignored on flushing sys.stdout:
OSError: [Errno 22] Invalid argument
```

The process exits nonzero (120 on POSIX). A downstream `head` closing the pipe
is normal, so graphify should exit 0 the way `foo | head` does.

## Fix

Wrap the CLI entry point: catch `BrokenPipeError` and `OSError` with `EINVAL` or
`EPIPE` (Windows reports a closed pipe as `EINVAL`), redirect stdout to devnull
so the shutdown flush stays quiet, and exit 0. Unrelated `OSError`s still
propagate, so real failures such as a missing `graph.json` are not masked.

## Verification

A reader that reads one line then closes the pipe:

```
before: returncode=120, "Exception ignored on flushing sys.stdout"
after:  returncode=0,   clean
```

Adds `tests/test_broken_pipe_exit.py` (5 tests covering the exit-code contract,
including that non-pipe errors still propagate). Existing CLI suites pass.

### DIFF
--- a/graphify/__main__.py
+++ b/graphify/__main__.py
@@ -1,6 +1,7 @@
 """graphify CLI - `graphify install` sets up the Claude Code skill."""
 
 from __future__ import annotations
+import errno
 import functools
 import json
 import os
@@ -444,7 +445,7 @@ _CODEX_HOOK = {
 
 
 
-def main() -> None:
+def _main() -> None:
     for _stream in (sys.stdout, sys.stderr):
         if _stream is not None and hasattr(_stream, "reconfigure"):
             try:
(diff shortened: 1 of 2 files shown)


safishamsi:
Thanks @varuntej07 — this shipped as part of the #1807 fix in 0.9.14. Your explicit in-guard `sys.stdout.flush()` was the important nuance my first pass missed (block-buffered piped stdout only flushing at shutdown, outside the guard), so I adopted it in `a46eee4` and credited you. Also handled the Windows `OSError(EINVAL)` form per your note. Really appreciate the well-reasoned PR.

varuntej07:
Thanks for the review and the credit @safishamsi. Glad the explicit flush helped close that edge case cleanly. Appreciate you building Graphify, keep pushing! More tasks, fewer tokens!!

PR #1830, opened 2026-07-12
fix: to_graphml crashes on dict/list-valued graph, node, and edge attributes

hofmockel:
## Summary

`nx.write_graphml` only accepts scalar attribute values. `to_graphml` already coerces `None -> ""` (fixing a prior `ValueError`, #1502), but a `dict` or `list` value still raises:

```
TypeError: GraphML does not support type <class 'dict'> as data values.
TypeError: GraphML does not support type <class 'list'> as data values.
```

This hits real graphs in practice, not just synthetic edge cases:
- a per-node `"metadata"` dict attribute
- `attach_hyperedges()`'s graph-level `"hyperedges"` list (`G.graph["hyperedges"]`)

Both crash `graphify export graphml` outright on any graph that uses them — confirmed against a real ~2,300-node/4,300-edge project graph.

## Fix

Extends the existing scrub step in `to_graphml` (the one that already coerces `None -> ""`) to also JSON-serialize any dict/list-valued attribute — across graph-level, node, and edge scopes — before calling `nx.write_graphml`.

## Test plan

- Added `test_to_graphml_tolerates_dict_and_list_attribute_values` (`tests/test_export.py`), mirroring the existing `test_to_graphml_tolerates_none_attribute_values` test — injects a node `metadata` dict, an edge `list` attribute, and a graph-level `hyperedges` list, asserts `to_graphml` doesn't raise.
- Ran full `tests/test_export.py` (all passing) and the full suite (`57 failed, 2891 passed, 165 skipped` — the 57 failures are pre-existing, all in `test_read_hook.py`/`test_reflect.py`/`test_search_hook.py`/`test_terraform.py`, unrelated to `export.py`/GraphML and present before this change).
- Verified against a real project graph that previously crashed on export — now produces a valid file that round-trips through `networkx.read_graphml`.

### DIFF
--- a/graphify/export.py
+++ b/graphify/export.py
@@ -912,14 +912,24 @@ def to_graphml(
         for k in [k for k in attrs if k.startswith("_")]:
             del attrs[k]
     # nx.write_graphml raises ValueError on None attribute values; replace with "".
+    # It also only accepts scalar attribute values, so dict/list-valued
+    # attributes (e.g. per-node "metadata", graph-level "hyperedges") must be
+    # serialized to a JSON string first, or the writer raises TypeError.
+    def _scrub(val):
+        if val is None:
+            return ""
+        if isinstance(val, (dict, list)):
+            return json.dumps(val)
+        return val
+
+    for key, val in list(H.graph.items()):
+        H.graph[key] = _scrub(val)
     for node_id in H.nodes():
         for key, val in list(H.nodes[node_id].items()):
-            if val is None:
-                H.nodes[node_id][key] = ""
+            H.nodes[node_id][key] = _scrub(val)
     for u, v in H.edges():
         for key, val in list(H.edges[u, v].items()):
-            if val is None:
-                H.edges[u, v][key] = ""
+            H.edges[u, v][key] = _scrub(val)
     nx.write_graphml(H, output_path)
 
 
--- a/tests/test_export.py
+++ b/tests/test_export.py
@@ -99,6 +99,25 @@ def test_to_graphml_tolerates_none_attribute_values():
         content = out.read_text()
         assert "<graphml" in content
 
+def test_to_graphml_tolerates_dict_and_list_attribute_values():
+    """nx.write_graphml only accepts scalar attribute values and raises
+    TypeError on a dict or list — e.g. a node's "metadata" dict, or a
+    graph-level "hyperedges" list (attach_hyperedges()). to_graphml must
+    JSON-serialize these instead of passing them through unmodified."""
+    G = make_graph()
+    communities = cluster(G)
+    a_node = next(iter(G.nodes()))
+    G.nodes[a_node]["metadata"] = {"kind": "file"}
+    if G.number_of_edges():
+        u, v = next(iter(G.edges()))
+        G.edges[u, v]["tags"] = ["a", "b"]
+    G.graph["hyperedges"] = [{"nodes": [a_node], "label": "x"}]
+    with tempfile.TemporaryDirectory() as tmp:
+        out = Path(tmp) / "graph.graphml"
+        to_graphml(G, communities, str(out))  # must not raise
+        content = out.read_text()
+        assert "<graphml" in content
+
 def test_to_html_creates_file():
     G = make_graph()
     communities = cluster(G)


fyzanshaik (changes requested):
I reproduced issue #1831 against this PR and came to this conclusion: the patch fixes GraphML serialization for regular `Graph`/`DiGraph` inputs, but `to_graphml()` still crashes when the input is a `MultiGraph` or `MultiDiGraph`. I believe the edge-scrubbing loop needs to handle NetworkX keyed edges before this is safe to merge.

- The original issue is fixed for regular graphs: dict/list values at graph, node, and edge scope are converted to JSON strings.

- `tests/test_export.py` passes on this branch: 45 tests passed.

- I reproduced a remaining failure using a `MultiDiGraph` with node metadata and a list-valued edge attribute. The export fails before `nx.write_graphml()` runs:

```text
ValueError: not enough values to unpack (expected 3, got 2)
```

- The failure comes from accessing `H.edges[u, v]`. That lookup is valid for `Graph`/`DiGraph`, but a multigraph edge is identified by `(u, v, key)`.

- This appears in scope because Graphify has explicit `MultiGraph`/`MultiDiGraph` compatibility and node-link loading support. Issue #1831 asks for dict/list serialization across edge scope, which currently remains incomplete for this supported graph shape.

- Please add a `MultiDiGraph` regression containing a dict/list-valued edge attribute. The test should read the generated GraphML back and assert that the edge attribute contains the expected JSON string, so it verifies preservation rather than only file creation.

- Non-blocking: `_scrub` could be renamed to `_normalize_graphml_value` or `_to_graphml_attribute_value` for clarity.

I left the suggested implementation directly on the affected edge loop.

graphify/export.py
```diff
@@ -912,14 +912,24 @@ def to_graphml(
...
+        return val
+
+    for key, val in list(H.graph.items()):
+        H.graph[key] = _scrub(val)
     for node_id in H.nodes():
         for key, val in list(H.nodes[node_id].items()):
-            if val is None:
-                H.nodes[node_id][key] = ""
+            H.nodes[node_id][key] = _scrub(val)
     for u, v in H.edges():
```
fyzanshaik, line 930: This lookup crashes for `MultiGraph`/`MultiDiGraph` because `H.edges[u, v]` requires an edge key. Iterating the mutable attribute dictionaries directly works for both regular and keyed graphs:

```suggestion
    for _, _, attrs in H.edges(data=True):
        for key, value in list(attrs.items()):
            attrs[key] = _scrub(value)
```

Please cover this with a `MultiDiGraph` edge containing a dict/list attribute and verify the value after a GraphML round trip.

fyzanshaik:
Suggested test cases for this change:

- `Graph`: graph-level `hyperedges` list, node-level `metadata` dict, and edge-level list attribute export without raising.

- `DiGraph`: the same values export successfully and preserve edge direction after `nx.read_graphml()`.

- `MultiGraph`: two parallel edges with different dict/list attributes both export; neither edge is skipped or overwritten.

- `MultiDiGraph`: two keyed, directed parallel edges export successfully and remain distinct after reading the GraphML. This is the regression for the current failure.

- Round trip: read the output with `nx.read_graphml()`, run `json.loads()` on each serialized dict/list value, and assert it equals the original value. Checking only that `<graphml` exists would not catch dropped or overwritten attributes.

- Empty containers: `{}` and `[]` serialize to valid JSON strings and round-trip correctly.

- Nested JSON values: a value such as `{"items": [1, {"enabled": true}]}` serializes and round-trips without structure loss.

- Existing null handling: graph, node, and edge `None` attributes should still become `""`; this prevents the current null-attribute behavior from regressing.

- Input immutability: after `to_graphml()`, the original graph still contains dict/list objects rather than the JSON strings written through the copied graph.

- Output validity: a successful export produces a non-empty file that `nx.read_graphml()` can parse.

The minimum merge-blocking set is `Graph`, `MultiDiGraph` with parallel edges, round-trip value assertions, and the existing `None` regression.

safishamsi:
Thanks @hofmockel — landed in `a46eee4` (0.9.14). I extended your approach slightly: coercion also covers the graph-level scope (where the `hyperedges` list lives) and preserves native int/float/bool rather than stringifying everything, plus an atomic temp-file write for the 0-byte-file case you called out in #1831. Credited you in the commit. Appreciate the PR and the mirrored test.

Issue #1766, opened 2026-07-10
query: homonymous generic symbols (e.g. many GET() route handlers) all get seeded, flooding results — related to #1303


jsilrecavo:
**Version:** graphifyy 0.9.11 (uv tool install), macOS. Same behavior on 0.9.9 and 0.9.11 (post-#1602).

**Symptom:** on a Next.js repo (~2.3k nodes, ~250 communities), a natural-language query about one specific endpoint seeds on four *different* `GET()` nodes from unrelated route files (every Next.js route handler is a function literally named `GET`; the query contained the word "get"), plus one stray lexical match on an unrelated infra node (the query contained "endpoint"). BFS depth-2 then returns 252 nodes, ~90% unrelated — the genuinely relevant community IS found via the other, specific seeds, but it's buried.

This is adjacent to #1303 (tied _EXACT_MATCH_BONUS scores aren't degree/community-diversified) and survives the #1602 fix — the difference here is many distinct nodes sharing one generic name, which is the norm for convention-named exports (GET/POST/default/handler/index).

**Suggested direction:**
- Penalize (or dedupe to one representative) seed candidates whose label is shared by many nodes — high label-multiplicity signals a framework convention, not a discriminator.
- Qualify generic-name seeds by co-occurrence (path/community affinity with the other seeds) before committing their BFS neighborhoods.
- A per-seed cap or relevance-ranked output would bound the blast radius of one bad seed.

safishamsi:
Fixed in `6ca8604` (ships in 0.9.14). BFS seed selection now dedups by normalized label, so a swarm of homonymous generic symbols (many `GET`/`POST` handlers, a repeated `handler`) contributes at most one representative seed instead of consuming every slot — while each distinct query term still gets a guaranteed seed. The fix is confined to seed selection; the shared node scorer used by `shortest_path`/`explain` is untouched. Thanks to @devcool20 for the PR (#1832).

PR #1832, opened 2026-07-12
fix(query): prevent generic symbol query flooding via multiplicity penalty and seed deduplication

devcool20:
### Description
This PR resolves **Issue #1766**, which causes query results to be flooded when codebases contain multiple generic/homonymous names (such as Next.js route handlers named `GET` or `POST`, or framework entrypoints like `index` or `handler`). 

Previously, when a query matched a generic word (e.g., `"get"`), multiple distinct nodes sharing the same generic label (e.g., `GET()`) would receive the full `_EXACT_MATCH_BONUS` and get selected as query seeds. Running a BFS traversal from multiple generic nodes ended up pulling in massive portions of unrelated code, burying the actual specific target.

---

### Proposed Changes

1. **Label Multiplicity Penalty** (`graphify/serve.py`):
   - Computes and caches a map of normalized label counts (`_label_counts_cache`) on the graph `G`.
   - Divides the final matching score of any node by its label multiplicity (`score /= multiplicity`), ensuring generic framework convention names cannot easily overpower specific identifiers.

2. **Seed Label Deduplication** (`graphify/serve.py`):
   - Adds a `seen_labels` tracking set in `_pick_seeds` to filter out duplicate labels, guaranteeing that only a single representative node gets seeded for any given homonymous label.

---

### Validation & Testing

- **Scenario Mock Test**: Verified using a Next.js route handler mock graph. Before the fix, querying `"get users"` seeded `users_model`, `api_users_GET`, and `api_analytics_GET` (flooding the BFS traversal). After the fix, the query correctly seeds only `users_model` and a single representative `GET` node (`api_analytics_GET`).
- **Regression Testing**: Ran the full serve test suite (`pytest tests/test_serve.py`). All **80 unit tests passed successfully** with 0 regressions.

### DIFF
--- a/graphify/serve.py
+++ b/graphify/serve.py
@@ -285,6 +285,14 @@ def _trigram_candidates(G: nx.Graph, needles: list[str], *, guard_frac: float = 
 
 def _score_nodes(G: nx.Graph, terms: list[str]) -> list[tuple[float, str]]:
     scored = []
+    # Compute label counts for multiplicity check (cached on G)
+    label_counts = G.graph.setdefault("_label_counts_cache", {})
+    if not label_counts:
+        for _, data in G.nodes(data=True):
+            lbl = (data.get("norm_label") or _strip_diacritics(data.get("label") or "")).lower()
+            if lbl:
+                label_counts[lbl] = label_counts.get(lbl, 0) + 1
+
     # Dedupe tokens, order-preserving (as _pick_seeds already does): a repeated
     # query word must not double-count every tier, and with coverage scaling
     # below it would also inflate the matched-term ratio (#1602).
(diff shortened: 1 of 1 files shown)


safishamsi:
Thanks @devcool20 — the seed-deduplication idea is the right fix and I've landed it in `6ca8604` (ships in 0.9.14), crediting you.

I kept the seed dedup (with two tweaks: the dedup key is normalized so `GET`/`Get`/`get` collapse together, and the per-term guarantee loop honors the same cap), but dropped the second mechanism — the per-label multiplicity penalty (`score /= multiplicity`). That penalty lives in `_score_nodes`, which `shortest_path` and `explain` also use to resolve endpoints, so dividing scores there silently reweights path/explain endpoint selection (out of scope for #1766 and able to flip which node a path anchors on). The dedup alone bounds the flood. I also added the regression tests the PR was missing. Appreciate the fix.

Issue #1810, opened 2026-07-11
graphify ignores `.git/info/exclude` — indexes git-worktree dirs, causing massive graph bloat


cdahl86-cyber:
## Summary

graphify builds its ignore set only from `.gitignore` and `.graphifyignore` (`detect.py`, `_load_graphifyignore` iterates exactly `(".gitignore", ".graphifyignore")`). It does **not** read `.git/info/exclude` — the standard place git records git-worktree paths and local-only excludes. As a result graphify walks into git-worktree directories nested under the repo (each a full second copy of the codebase) and the graph explodes.

## Repro

1. A repo with git worktrees checked out under it, e.g. `worktrees/foo`, excluded via `.git/info/exclude` (not `.gitignore`).
2. `graphify update .` (or the `graphify hook install` auto-rebuild hooks) walks into `worktrees/*` and indexes every copy.

In my case (5 worktrees) the graph went from **~9,400 nodes / 10 MB** to **~210,000 nodes / 311 MB** — ~77% duplicate worktree nodes, near the 512 MB `graph.json` cap — and it was regenerated on every commit by the auto-rebuild hooks.

## Version

graphifyy 0.9.12

## Suggested fix

Include `.git/info/exclude` in the ignore set (or, more robustly, defer to `git check-ignore` semantics when inside a git repo). At minimum, document that only `.gitignore` / `.graphifyignore` are honored, so users know to add worktree dirs there.

## Workaround

Add the worktree directory to `.gitignore` or `.graphifyignore`.

safishamsi:
Fixed in `5f57c46` (ships in 0.9.14). Detection now reads `.git/info/exclude` in addition to `.gitignore`/`.graphifyignore`, loaded at lowest precedence (below every per-directory ignore file, matching git — so a nearer `!` re-include still wins). The linked-worktree/submodule case (where `.git` is a file pointing at a gitdir) is resolved to the shared common git dir, so `info/exclude` is found there too. Verified against a repo whose worktree dir was excluded only via `info/exclude`: the worktree copy is no longer indexed, real source is kept. Thanks for the precise diagnosis and the before/after node counts, @cdahl86-cyber.

PR #1847, opened 2026-07-13
fix(detect): honor nested .gitignore/.graphifyignore files below the scan root

### LINKED ISSUE #1206: BUG: graphiphy does not respect nested .gitignore files
I asked claude code (with graphify setup done) to ensure graphiphy respects all .gitignore files, including nested files. And there is claude code answer:

```
Key finding: graphify only reads the ignore file in the scan-root's ancestor chain — i.e. just the root .gitignore when scanning from the repo root. It does NOT read nested .gitignore files.
```

Could you please fix this issue?

Mohak-Agrawal:
## Summary

Closes #1206.

`detect()` only read `.gitignore`/`.graphifyignore` files in the scan root and its **ancestor** directories (up to the nearest VCS root), loaded once before the walk began. A `.gitignore` sitting in a **descendant** directory — e.g. `vendor/sub/.gitignore` — was never read at all, so files/dirs it was meant to exclude leaked into the graph. Real `git` (and every other gitignore-aware tool) honors `.gitignore` at every directory level, not just the ancestor chain.

- Extracted the per-directory read+parse logic (previously inlined in `_load_graphifyignore`'s ancestor loop) into a shared `_load_dir_own_ignore()` helper — no behavior change to the existing ancestor-chain path.
- `detect()`'s `os.walk` loop now calls `_load_dir_own_ignore(dp)` for every directory it visits (except the scan root, already covered), appending patterns *before* that directory's children are pruned — so a nested ignore file governs its own subtree with git's usual closer-file-wins precedence.

## Test plan

- [x] 3 new regression tests in `tests/test_detect.py`: nested file exclude, nested directory prune (walk never descends into it), nested `!` negation overriding a broader root-level rule
- [x] All 6 pre-existing `.graphifyignore`/ancestor-chain tests re-verified — no regressions
- [x] Manually exercised all 9 tests end-to-end against real temp directories (this sandbox's Python has no `pytest` installed) — all pass; would appreciate a CI run to confirm under the project's actual test runner

### DIFF
--- a/graphify/detect.py
+++ b/graphify/detect.py
@@ -832,6 +832,33 @@ def _git_info_exclude(vcs_root: Path) -> Path | None:
     return exclude if exclude.is_file() else None
 
 
+def _load_dir_own_ignore(d: Path) -> list[tuple[Path, str]]:
+    """Read .gitignore/.graphifyignore directly inside *d* (not its ancestors).
+
+    Merges .gitignore and .graphifyignore for this one directory (#1363):
+    .gitignore is read first and .graphifyignore last, so .graphifyignore
+    patterns (including `!` negations) win on conflict via last-match-wins;
+    adding a .graphifyignore can only ever exclude MORE, never re-include a
+    .gitignore-excluded file (#945 kept: a dir with only a .gitignore still
+    gets sensible defaults).
+
+    Shared by `_load_graphifyignore` (ancestor chain, loaded once before the
+    scan) and the live os.walk loop in `detect()` (called per-directory as
+    each descendant is visited), so nested ignore files *below* the scan
+    root are honored too — previously only the scan root and its ancestors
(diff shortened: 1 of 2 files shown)


safishamsi:
Landed on v8 (`8a5287a`, your commit preserved) plus a test follow-up (`7ca43ab`) — ships in 0.9.15. Thanks @Mohak-Agrawal. The fix is correct and composes cleanly with the `.git/info/exclude` handling from #1810: nested patterns are read during the walk and anchored to their directory, so a nearer `!` re-include still wins over both a root `.gitignore` and `info/exclude` (verified). I tightened the negation test to use `.py` files (deterministic `code` bucket instead of a fuzzy document/unclassified union) and added a composition test pinning the three-source precedence.

Issue #1851, opened 2026-07-13
dedup ID-collision warning misdiagnoses cause, overstates data loss, and misses the silent drop (dedup.py:224-246)


bchan84x:
## Summary

`deduplicate_entities`' pre-dedup pass (`graphify/dedup.py:224-246`) warns about ID collisions with a message that (a) asserts a cause that is frequently wrong, (b) claims data loss where usually none occurs, and (c) is skipped entirely for the one case where a node really is silently dropped. The survivor is also picked by insertion order rather than by which file defines the entity.

Version: graphifyy 0.9.11, run via `/graphify . --update` (Claude Code skill; semantic extraction by host-agent subagents, no Gemini key).

## What I saw

```
[graphify] WARNING: node 'agents_make_batch_fixtures_make_batch_fixtures' from
'/…/available/diagnose-issue/SKILL.md' collides with node from 'agents/make-batch-fixtures.md'
— the second node will be dropped. This is a cross-chunk ID collision caused by two files with
the same name in different directories. To avoid data loss, run 'graphify extract' per subfolder
and merge with 'graphify merge-graphs'.
```

Three extracted nodes shared the id `agents_make_batch_fixtures_make_batch_fixtures`:

| Source file | Label | Fate |
| --- | --- | --- |
| `agents/make-batch-fixtures.md` | `make-batch-fixtures agent` | kept |
| `agents/make-batch-fixtures.md` | `make-batch-fixtures helper agent` | dropped, **no warning** (same `source_file`) |
| `available/diagnose-issue/SKILL.md` | `make-batch-fixtures` | dropped, warned |

All three describe the same entity: one is the agent's own defining doc, one is a duplicate label the extractor produced for that same doc, one is a cross-reference from a doc that mentions the agent. Edges are keyed by node id, so all 3 edges survived and the post-build health check reported 0 dangling / 0 missing / 0 collapsed edges out of 663. Nothing was actually lost.

## Problems

1. **The stated cause is often wrong.** The message hardcodes *"two files with the same name in different directories."* Here the two files have different names (`SKILL.md` vs `make-batch-fixtures.md`) in different directories. The real cause is that the node id is `<slug(source_file)>_<slug(label)>`, and any doc that *references* an entity can emit a node id derived from the referenced file's path — so a cross-reference node and the entity's own node collide by construction. That is dedup working as intended, not a filename clash.

[...]

PR #1852, opened 2026-07-13
fix(dedup): let the defining file win an ID collision, and warn only about real loss

bchan84x:
Fixes the ID-collision handling described in #1851.

## The problem

A node ID is `<slug(source_path)>_<slug(entity)>`, so a doc that merely *references* an entity mints that entity's own ID and collides with the entity's defining node **by construction**. The pre-dedup pass in `deduplicate_entities` handled that badly:

- **Survivor was insertion-order.** `seen_ids` kept whichever node arrived first, so chunk order decided whether an entity kept its own attributes or a passing cross-reference's thinner label.
- **The warning asserted a wrong cause.** It hardcoded *"two files with the same name in different directories"* — in the reported case the two files had different names, and nothing was lost (edges are keyed by ID and rewired to the survivor).
- **The genuinely lossy drop was silent.** The warning was gated on `existing_sf != new_sf`, so two same-ID nodes from the *same* file — the extractor emitting two different labels — were dropped with no output at all.

## The change

`graphify/dedup.py` only.

1. **Definer wins.** When several nodes share an ID, the survivor is the one whose `source_file` is the file the ID encodes; otherwise first-seen, as before. `_id_prefixes()` accepts any trailing slice of the path, so absolute paths, repo-relative paths, and pre-#1504 bare-filename-stem IDs all resolve.
2. **Report in proportion to what the drop costs.**
   - A cross-reference folding into the node it references → **silent** (nothing is lost).
   - Same file, two labels for one ID → `[graphify] note:` naming the discarded label (this was the previously-silent case).
   - Two different files both encoding the ID → still **WARNING**, now stating the ID-scheme cause and keeping the `extract` per-subfolder + `merge-graphs` advice. This is the #1504 case, and its existing test still passes unchanged.

## Tests

New tests in `tests/test_dedup.py`: definer-wins in both insertion orders, reference collisions stay silent and keep their edges, absolute `source_file` paths still resolve, same-file relabels emit the note, plus a unit test for `_defines_id`. Both existing collision tests (`test_cross_chunk_id_collision_emits_warning`, `test_same_id_same_source_file_no_warning`) are untouched and green.

`pytest tests/test_dedup.py` → 40 passed. `ruff check` clean.

## Notes

- CHANGELOG.md deliberately not touched — the entries there look maintainer-written at release time. Happy to add one if you'd prefer.
- Two local failure clusters are environmental, not from this change: `test_falkordb_integration.py` (no FalkorDB module in my local Redis) and `test_skillgen.py` (`git show` of baseline blobs, which a `--depth 1` clone lacks; CI uses `fetch-depth: 0`).

Refs #1851

### DIFF
--- a/graphify/dedup.py
+++ b/graphify/dedup.py
@@ -187,6 +187,76 @@ def _is_code(node: dict) -> bool:
     return node.get("file_type") == "code"
 
 
+# ── ID collisions ─────────────────────────────────────────────────────────────
+
+_ID_SEGMENT = re.compile(r"[^a-z0-9]+")
+_EXTENSION = re.compile(r"\.[^./]+$")
+
+
+def _id_prefixes(source_file: str) -> set[str]:
+    """The ID prefixes a node extracted from ``source_file`` may legitimately mint.
+
+    An ID is ``<path>_<entity>``, where the path is the extension-stripped source
+    path, each segment slugified and joined with ``_``. Every trailing slice of the
+    path counts as a prefix: the stored path may be absolute or repo-relative, and
+    graphs built under the pre-#1504 scheme keyed off the bare filename stem.
+    """
+    stem = _EXTENSION.sub("", source_file.replace("\\", "/"))
+    segments = [s for s in (_ID_SEGMENT.sub("_", p.casefold()).strip("_")
+                            for p in stem.split("/")) if s]
+    return {"_".join(segments[i:]) for i in range(len(segments))}
+
+
+def _defines_id(node: dict) -> bool:
+    """True when the node's own source_file is the file its ID encodes.
+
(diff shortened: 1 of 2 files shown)


rickgonzalez:
A scan of this PR to explain it - using a tool I'm working on. Hope that it helps: https://www.lenzon.ai/viewer/cmrju75nq0011odzhizb1hp1p?voice=google-chirp3

safishamsi:
Landed on v8 as 70bc9ca plus follow-up fd54f0e, ships in 0.9.16. Thanks @bchan84x - kept your tiered reporting and definer-wins heuristic; added a deterministic order-independent tiebreak and bare-file-node definer recognition.

PR #1862, opened 2026-07-13
fix(detect): legacy-manifest branch must `!=`-compare mtime to catch reverse jumps (#1859)

### LINKED ISSUE #1859: detect_incremental legacy-manifest branch misses backwards-moving mtimes (git checkout, tar restore)
### Summary

`detect_incremental` uses `current_mtime > stored` (a strict greater-than) on the legacy manifest schema and does no content-hash verification. Files whose mtime moved **backwards** — after `git checkout <older-commit>`, `tar -xf` restore, `rsync --times`, or a manual timestamp adjustment — are silently treated as unchanged and never re-extracted. The graph then reflects the *newer* content while the corpus on disk holds the older content.

The parallel dict-schema branch handles this correctly (`!=` + hash-verify). The legacy-schema branch does not.

### Where

`graphify/detect.py:1583-1610`:

```python
# Legacy manifest: plain float value — treat as ast_hash only
if isinstance(stored, (int, float)):
    changed = stored is None or current_mtime > stored
elif isinstance(stored, dict):
    # …
    if stored_mtime is None or current_mtime != stored_mtime:
        # mtime bumped — verify with content hash before re-extracting
        changed = _md5_file(Path(f)) != stored_hash
    else:
        changed = False
```

Two issues in the float branch:

1. **`>` misses backwards mtime motion.** `git checkout <older-commit>` restores older file timestamps; those files' `current_mtime < stored` and the check reports `changed = False`. The stored graph is now stale relative to what's on disk. The dict branch's `!=` catches this.
2. **`stored is None` is dead code.** The enclosing `isinstance(stored, (int, float))` already excludes `None`.

The legacy branch also skips the `_md5_file` verification the dict branch performs, so even when the float mtime *does* differ it re-extracts without checking whether the AST hash actually changed — the reverse cost of the same asymmetry.

### Reproduction

1. Build a graph from a clean checkout at commit `A`.
2. Copy the manifest into legacy form (single float mtime per file — the shape written by graphify prior to the dict-schema migration).
3. `git checkout <older-commit-B>` where at least one tracked file's content differs and its mtime is older than at A.
4. Run `graphify update` (which calls `detect_incremental`).
5. Observe: `new_files` does not contain the reverted file; its cached AST from A is retained. The graph now describes commit A's version of a file whose bytes are commit B.

### Suggested fix

Align the float branch with the dict branch: compare with `!=` and content-hash-verify before re-extracting:

```python
if isinstance(stored, (int, float)):
    if current_mtime != stored:
        # mtime moved (either direction) — verify with content hash
        changed = _md5_file(Path(f)) != ""    # legacy schema stored no hash;
        # re-extract on any mtime change since we have no hash to compare against
        changed = True if current_mtime != stored else False
    else:
        changed = False
```

Or, cleaner — normalise legacy floats up-front (already done in the dict branch for the `{mtime, hash}` shape) so a plain float is treated as `{"mtime": stored, "ast_hash": "", "semantic_hash": ""}` and falls through to the dict logic, which then correctly re-extracts (empty stored_hash → `changed = True`) and rewrites the entry in the new schema on the next save.

Happy to open a PR with either shape and a regression test that emulates a backwards-mtime checkout.

thejesh23:
## Summary

- Replaces `current_mtime > stored` with `current_mtime != stored` in the legacy-float branch of `detect_incremental`, so a file whose mtime moved **backwards** — after `git checkout <older-commit>`, `tar -xf`, or `rsync --times` — is treated as changed and re-extracted. The dict branch has always used `!=` and hash-verified; this brings the legacy branch in line.
- Also removes the dead `stored is None or` clause, since the enclosing `isinstance(stored, (int, float))` already excludes `None`.
- Adds two regression tests in `tests/test_detect.py`:
  - `test_detect_incremental_legacy_float_reextracts_on_backwards_mtime` — writes a legacy manifest with `mtime = current + 3600` (i.e. from the future, as would happen after checking out an older revision) and asserts the file lands in `new_files`. Fails on the pre-fix `>` comparator, passes now.
  - `test_detect_incremental_legacy_float_skips_when_mtime_matches` — non-regression: same mtime still skips.

The legacy branch stores no hash (only a float), so there is nothing to content-verify against; on any mtime delta we re-extract, and the next `save_manifest` promotes the entry into the dict schema, where the `!=`/hash-verify flow already exists.

Closes #1859.

## Test plan

- [x] `uv run --frozen pytest tests/test_detect.py -v -k "legacy_float or dict_valued_mtime"` → 3 passed.
- [x] `uv run --frozen pytest tests/ -q` → 3195 passed, 3 skipped.
- [x] `uv run --frozen pre-commit run --files graphify/detect.py tests/test_detect.py` → skillgen + ruff green.

### DIFF
--- a/graphify/detect.py
+++ b/graphify/detect.py
@@ -1581,9 +1581,15 @@ def detect_incremental(
             except Exception:
                 current_mtime = 0
 
-            # Legacy manifest: plain float value — treat as ast_hash only
+            # Legacy manifest: plain float value stores only mtime.
+            # Compare with `!=` so backwards mtime motion (git checkout of an
+            # older commit, tarball restore, rsync --times) still triggers a
+            # re-extract; the previous `>` silently kept the stale cache and
+            # the graph drifted from disk (#1859). No stored hash means we
+            # cannot verify content — any mtime delta forces a re-extract,
+            # and the next save promotes the entry into the dict schema.
             if isinstance(stored, (int, float)):
-                changed = stored is None or current_mtime > stored
+                changed = current_mtime != stored
             elif isinstance(stored, dict):
                 # Normalise legacy {mtime, hash} to new schema
                 if "hash" in stored and "ast_hash" not in stored:
--- a/tests/test_detect.py
+++ b/tests/test_detect.py
@@ -369,6 +369,67 @@ def test_detect_incremental_survives_dict_valued_mtime(tmp_path, monkeypatch):
     assert not any("mod.py" in f for f in result["unchanged_files"]["code"])
 
 
+def test_detect_incremental_legacy_float_reextracts_on_backwards_mtime(tmp_path, monkeypatch):
+    """Legacy float manifests must re-extract when mtime moves BACKWARDS (#1859).
+
+    Pre-fix the legacy branch used `current_mtime > stored`, which silently kept
+    the cached entry after operations that restore older mtimes: `git checkout`
+    of an older commit, `tar -xf` restore, or `rsync --times`. The graph then
+    reflected the newer content while disk held the older content. The dict
+    branch has always used `!=`; this test pins the legacy branch to the same
+    contract.
+    """
+    import json
+
+    monkeypatch.chdir(tmp_path)
+
+    src = tmp_path / "mod.py"
+    src.write_text("def old_content():\n    return 1\n", encoding="utf-8")
+    current_mtime = os.stat(src).st_mtime
+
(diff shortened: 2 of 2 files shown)


safishamsi:
Landed on v8 as e32be08, ships in 0.9.16. Thanks @thejesh23 - the compare matches the dict-schema branch and catches backwards mtime motion.

PR #1864, opened 2026-07-13
fix(llm): surface claude-cli errors reported on stdout

davetothek:
The `claude` CLI reports usage-limit and auth errors on **stdout** with an
empty stderr while exiting non-zero. `_call_claude_cli` / `_call_llm` only read
`proc.stderr`, so these failures surfaced as a bare `claude -p exited 1:` with
no indication of the actual cause — an unrecoverable usage limit looked
identical to any other non-zero exit.

Fall back to the stdout tail when stderr is empty, so the surfaced error carries
the real message. Adds `tests/test_llm_errors.py` covering both paths (prefer
stderr when present; fall back to stdout otherwise).


davetothek:
Superseded by #2555. Closing

PR #1868, opened 2026-07-13
fix(watch): preserve semantic edges of re-extracted sources (#1865)

### LINKED ISSUE #1865: Full `graphify update` silently evicts LLM semantic edges from re-extracted Markdown/doc sources (edge eviction is provenance-blind)
## Summary

A full `graphify update` (e.g. bare `graphify update .`, or the post-checkout hook) silently **evicts LLM-extracted semantic edges** (`semantically_similar_to`, `conceptually_related_to`, etc.) whose `source_file` is a document that *also* has an AST extractor — most notably Markdown (`.md`, `.mdx`, `.qmd`). The concept **nodes** survive but the semantic **edges** connecting them are dropped, leaving orphaned nodes.

This contradicts the stated intent in the code itself. `_reconcile_existing_graph` in `watch.py` comments that "edges from unchanged or **semantic** sources survive", and node preservation is deliberately provenance-aware ("Semantic nodes lack the AST origin marker and remain preserved"). Edge preservation is **not** provenance-aware, so on a full rebuild the intent is not met.

## Environment

- graphify (graphifyy) **0.9.14**, Python 3.13, macOS.

## Root cause

In `graphify/watch.py`, `_reconcile_existing_graph()`:

- **Node eviction is provenance-aware.** A node is only evicted if it is an AST-origin node (`node.get("_origin") == "ast"`) owned by a re-extracted source. Semantic nodes (no `_origin == "ast"` marker) are preserved.
- **Edge eviction is provenance-blind.** An edge is evicted purely by `source_file` membership: `is_evicted(edge, edge_evicted_source_identities)`, where on a full rebuild `edge_evicted_source_identities` includes the identity of **every re-extracted source** (`rebuilt_source_identities` = all `extract_targets`).

`extract_targets` on a full rebuild includes every code file **and every document file that has an AST extractor** — the loop that appends `.md`/`.mdx`/`.qmd` to `code_files`. So a semantic edge whose `source_file` is a Markdown doc is treated as "owned by a re-extracted source" and evicted, even though the AST pass that re-extracted the file does not (and cannot) regenerate that semantic edge — it was produced by the LLM pass, which `graphify update` does not run.

The AST pass legitimately owns and replaces a file's *AST* edges. It should not own that file's *semantic* edges. Node reconciliation already draws this distinction via `_origin`; edge reconciliation does not.

## Reproducer

Two Markdown docs describing the same idea with no structural link between them, so a real extraction emits a `semantically_similar_to` edge:

`auth.md`
```
# Token Validation
The access layer verifies a bearer token before granting a request. It checks the
token signature against the signing key, confirms the token has not expired, and
rejects any token whose audience claim does not match this service.
```

`login.md`
```
# Session Verification
When a login session is presented, the gateway decides whether it is still
trustworthy. It recomputes the session's signature from the secret, ensures the
session's expiry timestamp is in the future, and denies a session issued for a
different application.
```

1. AST-extract both, run a real semantic extraction (via `/graphify --mode deep` or `graphify extract`), and build `graph.json`. The stored edge is:
   ```json
   {"source": "auth_token_validation", "target": "login_session_verification",
    "relation": "semantically_similar_to", "confidence": "INFERRED", "source_file": "auth.md"}
   ```
   → **1 semantic edge present.**

2. Run a full rebuild:
   ```
   graphify update .
   ```

3. Inspect `graph.json`:
   - Concept nodes `auth_token_validation` and `login_session_verification`: **still present**.
   - `semantically_similar_to` edges: **0** — evicted.

The graph *grew* overall (more AST nodes/edges), so the `#479` shrink guard does not trigger — the loss is silent.

## Expected

A full `graphify update` (AST-only) should preserve semantic edges owned by sources it re-extracts, the same way it already preserves semantic *nodes*. Only the AST-tier edges of a re-extracted source should be replaced.

## Suggested fix

Make edge eviction provenance-aware, symmetric with node eviction: when reconciling a re-extracted source, evict only that source's **AST-tier** edges (e.g. gate the edge eviction on an `_origin`/tier marker, or only evict edges of a re-extracted source when a freshly extracted edge for that source is available to replace it). Semantic-tier edges of a re-extracted source should be preserved until a semantic re-extraction supersedes them — exactly the contract the surrounding comments already describe.

## Workaround (for reference)

Re-union the committed semantic graph after every full `update` (via `graphify merge-driver`), which restores the evicted edges. This works but requires carrying a separate semantic snapshot and re-applying it everywhere a full rebuild runs.

xor-xe:
Fixes #1865

**What was broken:** A full `graphify update` (and the post-checkout hook) silently evicted LLM-extracted semantic edges (`semantically_similar_to`, …) whose `source_file` is a document that also has an AST extractor (`.md`, `.mdx`, `.qmd`). Concept nodes survived; the edges connecting them were dropped, and since the graph grew overall the shrink guard never fired.

**Root cause:** `_reconcile_existing_graph` preserves nodes provenance-aware (`_origin == "ast"`, #1116) but evicted edges purely by `source_file` membership in the re-extracted set. On a full rebuild that set includes every doc with an AST extractor, so semantic edges were treated as owned by an AST pass that cannot regenerate them.

**Change:** AST-extracted edges now carry the same `_origin: "ast"` marker as nodes, and rebuild-driven edge eviction only claims that tier — a re-extraction replaces a source's AST edges while its semantic edges survive until a semantic re-extraction supersedes them. Deletion-driven eviction stays provenance-blind, so edges of deleted or excluded sources are still purged. `graph.json` already persists `_`-prefixed edge attrs (`_src`/`_tgt`) and `to_graphml` already strips them, so the marker stays internal.

**Migration note:** edges from graphs built before this change lack the marker; a stale AST edge from a file changed exactly once across the version boundary can linger until its source is deleted — the same trade-off the #1116 node marker made. Regenerating the graph clears it.

**Testing:** new regression test (full + incremental) reproducing the issue's scenario — fails on current `v8`, passes here — plus an assertion that stale AST-tier edges of a re-extracted source are still evicted. Full suite: 3195 passed, 3 skipped. Verified end-to-end with the issue's CLI repro.

### DIFF
--- a/graphify/extract.py
+++ b/graphify/extract.py
@@ -4874,9 +4874,14 @@ def extract(
     # Tag AST provenance so the incremental watch rebuild can distinguish
     # AST-extracted nodes from semantic/LLM nodes. On a full re-extraction
     # the watcher drops any AST-marked node missing from the fresh output
-    # even when its source file still exists (#1116).
+    # even when its source file still exists (#1116). Edges carry the same
+    # marker so edge eviction can be tier-scoped: re-extracting a source
+    # replaces its AST edges without evicting the semantic edges the AST
+    # pass cannot regenerate (#1865).
     for n in all_nodes:
         n["_origin"] = "ast"
+    for e in all_edges:
+        e["_origin"] = "ast"
 
     return {
         "nodes": all_nodes,
--- a/graphify/watch.py
+++ b/graphify/watch.py
@@ -408,11 +408,11 @@ def _reconcile_existing_graph(
         }
         node_evicted_source_identities = set(deleted_source_identities)
         hyperedge_evicted_source_identities = set(deleted_source_identities)
+        # Deletion evicts edges regardless of tier; re-extraction only owns a
+        # source's AST-tier edges (checked per-edge below, #1865).
+        edge_evicted_source_identities = set(deleted_source_identities)
         if not full_rebuild:
             node_evicted_source_identities.update(rebuilt_source_identities)
-        edge_evicted_source_identities = (
-            node_evicted_source_identities | rebuilt_source_identities
-        )
 
         # Reconcile every rebuild against the current watched corpus. Hook change
         # lists can contain only a rename destination, so explicit paths alone
@@ -485,14 +485,22 @@ def _reconcile_existing_graph(
         ]
         all_ids = new_ast_ids | {node["id"] for node in preserved_nodes}
 
-        # Edges are owned by source_file. Re-extraction must replace an owner's
-        # previous edges, while edges from unchanged or semantic sources survive.
+        # Edges are owned by source_file, but ownership is tier-scoped: the AST
+        # pass replaces a re-extracted source's AST edges, while that source's
+        # semantic/LLM edges — which the AST pass cannot regenerate — survive
+        # until a semantic re-extraction supersedes them. Same provenance rule
+        # the node reconciliation above applies via _origin (#1865). Deletion
+        # eviction stays provenance-blind.
         preserved_edges = [
             edge
             for edge in existing.get("links", existing.get("edges", []))
             if edge.get("source") in all_ids
             and edge.get("target") in all_ids
             and not source_paths.is_evicted(edge, edge_evicted_source_identities)
+            and not (
+                edge.get("_origin") == "ast"
+                and source_paths.is_evicted(edge, rebuilt_source_identities)
+            )
         ]
 
         new_hyperedge_ids = {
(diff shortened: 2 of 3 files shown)


rickgonzalez:
An explanation of this PR using Lenzon - hope that it helps explain this clearly. https://www.lenzon.ai/viewer/cmrjtm9ky000codzhxj2cbext?voice=google-chirp3

safishamsi:
Landed on v8 as 4ac5f07, ships in 0.9.16. Thanks @xor-xe - tier-scoped edge eviction via the _origin marker.

PR #1874, opened 2026-07-14
feat: add Solidity language extractor

babyblueviper1:
## Summary

Adds a Solidity extractor (`graphify/extractors/solidity.py`) over `tree-sitter-solidity`, following the existing per-language extractor pattern in this package (closest analog: `rust.py`).

Extracts: contracts, interfaces, libraries, functions/constructors/modifiers, structs, enums, events, inheritance (`is A, B`), imports (named `{A, B as C}`, namespace `* as X`, and bare), and same-file call resolution — including modifier invocations, which are a Solidity-specific call site with no explicit call syntax.

## Scope (disclosed honestly, not oversold)

Call resolution here is by same-file method-**name** matching, not receiver-type checking — matches rust.py's `label_to_nid` tradeoff exactly. This gets the common cases right (a storage variable typed as a locally-defined interface, `Library.staticMethod()`, `InterfaceName(addr)` cast-calls, modifier invocations) but:
- can mismatch if two unrelated types in the same file happen to share a method name
- never resolves a call to something this file doesn't itself define (an imported library/interface whose body lives elsewhere) — those are emitted via `raw_calls` for a possible future cross-file pass, same as several other extractors already do

Full type-qualified receiver resolution (the kind the larger OOP-language blocks in `extract.py` do for Java/C#/TS) is out of scope for this first cut — happy to follow up if that's wanted.

## Testing

- 14 new tests in `tests/test_multilang.py` + `tests/fixtures/sample.sol`, covering nodes, inherits, method/contains, import context, same-file call resolution, modifier-invocation calls, confidence, dangling-edge safety, and the missing-file error path. All pass.
- `pytest tests/ -q` on the full suite: no new failures introduced (confirmed by running the same set of failing tests against the clean `v8` HEAD before my change — same ~24 pre-existing failures on unrelated modules with missing optional deps/fixtures, e.g. `tree-sitter-hcl` not installed, `test_skillgen.py` audit baselines, one pre-existing test-ordering flake in `test_cache.py` unrelated to any language extractor).
- Ran the extractor against 16 real, non-vendored `.sol` files from a live project (contracts + Foundry tests) — zero extraction errors, and a real internal helper-function call chain (`validationRequest() → _authorized() → _agentOwner()`) resolved correctly end to end.
- `ruff check --select E9,F63,F7,F82` clean on the new/touched files.

## Checklist (per ARCHITECTURE.md § Adding a new language extractor)

- [x] `extract_solidity(path: Path) -> dict` following the existing pattern
- [x] Registered `.sol` in `extract.py`'s `_DISPATCH`
- [x] Added `.sol` to `CODE_EXTENSIONS` in `detect.py` (covers `watch.py`'s `_WATCHED_EXTENSIONS`, which derives from it — no separate edit needed there)
- [x] Added `tree-sitter-solidity>=1.0,<2.0` to `pyproject.toml`
- [x] Fixture + tests in `tests/test_multilang.py`


symulacr:
verified working on 0.9.63 — two new files plus four one-line edits, hand-rebased cleanly

PR #1876, opened 2026-07-14
fix(extract): filter Kotlin builtin/stdlib types from references graph

kebwlmbhee:
## Problem
`_java_collect_type_refs` filters out JDK builtin types via `_JAVA_BUILTIN_TYPES` before emitting a graph node/edge, and the same pattern exists for Python (`_PYTHON_ANNOTATION_NOISE`) and Go (`_GO_PREDECLARED_TYPES`). `_kotlin_collect_type_refs` has no equivalent — every `Boolean`, `String`, `Int`, `List`, ... language-level type annotation in a `.kt` file becomes a real graph node/edge.

In practice this means two unrelated Kotlin files that share nothing but `Boolean`/`String`/`Int` type annotations get treated as connected, and `cluster-only` groups them into the same community — a false positive with no real code coupling behind it. Confirmed by checking actual `import` statements between files in a real project: zero relation beyond one legitimate case.

## Fix
Add `_KOTLIN_BUILTIN_TYPES` mirroring `_JAVA_BUILTIN_TYPES` (kotlin.* scalars, collections, and throwables — deliberately **not** Android/JVM framework types like `Context`/`View`/`Bundle`, staying consistent with how `_JAVA_BUILTIN_TYPES` doesn't filter framework-specific types either), and check it plus `_JAVA_BUILTIN_TYPES` (Kotlin freely references `java.util.Date`/`UUID`/etc) at the three call sites in `_kotlin_collect_type_refs` that previously appended unconditionally.

Deliberately **excludes `"Result"`** from the filter even though `kotlin.Result` exists — it collides with the very common user-defined sealed-class name. Confirmed against this repo's own `tests/fixtures/sample.kt`, which defines its own `class Result<T>`; an earlier version of this patch that included `"Result"` broke `test_kotlin_parameter_return_generic_and_field_contexts`.

## Evidence
Real-world Android/Kotlin project (~400 source files), full `graphify . --code-only` + `cluster-only . --no-label`:

| | Before | After |
|---|---|---|
| Nodes | 3693 | 3239 (-12%) |
| Edges | 6481 | 5221 (-19%) |
| Communities | 269 | 271 |

Concretely: a utility file (`ScreenExt.kt`, purely `Int`/`Float` typed screen-dimension helpers) went from being merged into a 5-file grab-bag community (connected only via shared `Boolean`/`Int` type nodes) to splitting cleanly into its own community.

## Testing
- Added `test_kotlin_builtin_types_not_emitted_as_references` and `test_kotlin_user_types_still_emit_references` to `tests/test_languages.py` (the latter guards specifically against the `Result` collision above).
- Full suite: `3142 passed, 31 skipped` — the 5 failures on a clean checkout of `v8` (`test_collect_files_skips_hidden`, 4x `test_ollama_retry_cap.py` needing the optional `openai` package) are pre-existing/environmental and unaffected by this change (verified by running the same tests against `v8` with this commit stashed out).

## Scope note
This intentionally does not touch Android/JVM framework types (`Context`, `View`, `Bundle`, `SharedPreferences`, `JSONObject`, ...). Those still create some false-hub clustering in Android projects specifically; filtering them is a separate, more debatable scope-expansion (you'd lose "who touches `SharedPreferences`" as a real signal) and isn't part of this PR.

### DIFF
--- a/graphify/extractors/engine.py
+++ b/graphify/extractors/engine.py
@@ -607,6 +607,35 @@ def _php_method_return_type_node(method_node):
                 return c
     return None
 
+# Kotlin stdlib scalar/collection/core types that appear constantly as type
+# annotations but carry no useful semantic meaning as graph nodes (mirrors
+# _JAVA_BUILTIN_TYPES / _PYTHON_ANNOTATION_NOISE / _GO_PREDECLARED_TYPES).
+# Kotlin compiles to the JVM and freely references java.* types too, so this
+# is combined with _JAVA_BUILTIN_TYPES at the call site rather than duplicated.
+_KOTLIN_BUILTIN_TYPES = frozenset({
+    # kotlin — scalars & core
+    "Any", "Unit", "Nothing", "Boolean", "Byte", "Short", "Int", "Long",
+    "Float", "Double", "Char", "String", "CharSequence", "Number",
+    "Comparable", "Enum", "Annotation", "Pair", "Triple", "Lazy",
+    "Function",
+    # kotlin — throwables
+    "Throwable", "Exception", "RuntimeException", "Error",
+    "IllegalArgumentException", "IllegalStateException", "NullPointerException",
+    "IndexOutOfBoundsException", "ClassCastException", "NumberFormatException",
+    "ArithmeticException", "UnsupportedOperationException",
+    "NoSuchElementException", "ConcurrentModificationException",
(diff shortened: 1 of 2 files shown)


safishamsi:
Landed on v8 as 5c0a04c, ships in 0.9.16. Thanks @kebwlmbhee - matches Java/Python/Go builtin filtering.

PR #1881, opened 2026-07-14
fix(extract): recognize uppercase TypeScript extensions

xkam7ar:
## Summary

Normalize a file suffix once before grammar selection so uppercase TypeScript-family extensions (`.TS`, `.TSX`, `.MTS`, and `.CTS`) use the same TypeScript grammars as their lowercase forms.

Previously, `extract_js()` recognized uppercase JavaScript extensions through a lowercase check but compared the TypeScript branches against the original case-sensitive suffix. TypeScript-only declarations could therefore be omitted.

## Tests

- Regression failed before the source change on `.TS`
- Module-extension tests: **7 passed**
- Relevant extraction suite: **127 passed, 1 skipped**
- Ruff on both changed files: **passed**
- Required offline `graphify update .`: **passed** with no tracked generated changes
- `git diff --check`: **passed**

### DIFF
--- a/graphify/extract.py
+++ b/graphify/extract.py
@@ -1051,9 +1051,10 @@ def extract_python(path: Path) -> dict:
 
 def extract_js(path: Path) -> dict:
     """Extract classes, functions, arrow functions, and imports from a .js/.ts/.tsx/.mts/.cts file."""
-    if path.suffix == ".tsx":
+    suffix = path.suffix.lower()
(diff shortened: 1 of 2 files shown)


safishamsi:
Landed on v8 as 8361a81, ships in 0.9.16. Thanks @xkam7ar - uppercase TS extensions now use the TS grammar.

PR #1882, opened 2026-07-14
fix(llm): unwrap FileSlice via unit_path in the checkpoint allowlist (#1870)

### LINKED ISSUE #1870: Incremental semantic cache checkpoint fails on sliced docs: _checkpoint_chunk reads FileSlice.rel, which does not exist
## Summary

During `graphify extract`, the per-chunk incremental semantic cache checkpoint fails for every chunk that contains an oversized (sliced) document, printing:

```
[graphify] incremental cache checkpoint failed: argument should be a str or an os.PathLike object where __fspath__ returns a str, not 'FileSlice'
```

Extraction completes and produces a valid graph — but the incremental cache the checkpoint exists to provide is silently defeated for those chunks, so a re-run (or a run resumed after a crash/kill/rate-limit) re-bills them instead of reusing cached results.

## Environment

- graphify (graphifyy) **0.9.14**, Python 3.13, macOS.
- Observed on `graphify extract . --backend claude-cli` over a 70-document corpus (5 chunks): chunks 1–4 each printed the failure; chunk 5 (no sliced doc) did not.

## Root cause

`llm.py`, `_checkpoint_chunk` builds the `allowed_source_files` list as:

```python
# A FileSlice reports its file via `.rel`; a bare Path is the relative source_file itself.
allowed = [getattr(item, "rel", None) or item for item in chunk]
```

But `FileSlice` (`file_slice.py`) is a `@dataclass(frozen=True)` whose path field is **`.path`** — it has **no `.rel` attribute** (`[a for a in dir(FileSlice) if not a.startswith('__')]` is empty of `rel`). So for every `FileSlice`, `getattr(item, "rel", None)` returns `None`, and `or item` leaks the `FileSlice` object itself into `allowed`.

`save_semantic_cache(..., allowed_source_files=allowed)` then does, in `resolved_source_path`:

```python
path = Path(value)   # value is a FileSlice -> TypeError
```

(`cache.py`), which raises `argument should be a str or an os.PathLike object ... not 'FileSlice'`. The `except Exception` in `_checkpoint_chunk` catches it and prints the best-effort warning, so extraction continues but the chunk is never checkpointed.

The comment at the call site referencing `.rel` is stale relative to the `FileSlice` dataclass.

## Reproducer

`graphify extract . --backend <any> --out <dir>` on a corpus containing at least one document large enough to be split into `FileSlice` units (#1369). Every chunk that includes a slice prints the checkpoint failure.

## Suggested fix

Resolve a `FileSlice`'s source path via its actual attribute (`.path`) rather than the non-existent `.rel` — e.g. `getattr(item, "path", None) or item`, or unwrap `FileSlice` explicitly before building `allowed`. Whatever attribute genuinely carries the relative source_file should be used; `.rel` does not exist on the current dataclass.

HerenderKumar:
Fixes #1870

`_checkpoint_chunk` resolved each unit's source file with `getattr(item, "rel", None) or item`, but `FileSlice` carries its file in `.path` — it has no `.rel` — so for every sliced document the `FileSlice` object itself leaked into `allowed_source_files`. `save_semantic_cache` then raised on `Path(FileSlice)`, the best-effort guard downgraded that to the `incremental cache checkpoint failed` warning, and every chunk containing a slice silently lost its checkpoint — so a resumed run re-billed those chunks instead of reusing the cache.

- `llm.py`: build the allowlist with `unit_path(...)` — `file_slice.py`'s canonical unwrap (parent `Path` for a slice, bare `Path` unchanged), already imported and used elsewhere in the module — and fix the stale `.rel` comment
- `tests/test_chunking.py`: regression test that drives `extract_corpus_parallel` over a corpus whose document slices into 7 units and asserts the checkpoint fires with real paths; on current `v8` it fails with the exact `FileSlice` leak from the report

Since a slice's `.path` is the same value the file would have had as a bare `Path`, the allowlist entries for sliced files now match unsliced behavior exactly.

### DIFF
--- a/graphify/llm.py
+++ b/graphify/llm.py
@@ -1918,9 +1918,9 @@ def extract_corpus_parallel(
             # (#1757). The model can attribute a node's source_file to another
             # corpus file; without this bound, that stray node would clobber the
             # other file's complete cache entry (or, with merge_existing, pollute
-            # it). A FileSlice reports its file via `.rel`; a bare Path is the
-            # relative source_file itself.
-            allowed = [getattr(item, "rel", None) or item for item in chunk]
+            # it). unit_path unwraps a FileSlice to its parent file (#1870); a
+            # bare Path is the relative source_file itself.
+            allowed = [unit_path(item) for item in chunk]
             _scs(
                 result.get("nodes", []),
                 result.get("edges", []),
--- a/tests/test_chunking.py
+++ b/tests/test_chunking.py
@@ -203,6 +203,43 @@ def test_corpus_parallel_sequential_when_max_concurrency_is_one(tmp_path):
     assert call_order == [("f0.py",), ("f1.py",), ("f2.py",)]
 
 
+# ---- Incremental cache checkpoint --------------------------------------------
+
+def test_checkpoint_allowlist_unwraps_file_slices(tmp_path, monkeypatch):
+    """#1870: a chunk containing FileSlice units must checkpoint with the
+    slice's parent path in allowed_source_files — leaking the FileSlice
+    object itself made save_semantic_cache raise, silently skipping the
+    checkpoint for every chunk with a sliced document."""
+    from graphify import llm
+    from graphify.llm import extract_corpus_parallel
+
+    big = tmp_path / "big.md"
+    big.write_text("# heading\n" + "paragraph line\n" * 40)
+
+    captured = {}
+
+    def _capture_cache(*args, **kwargs):
+        captured.update(kwargs)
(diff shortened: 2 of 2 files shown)


safishamsi:
Thanks @HerenderKumar - same fix as #1870, already landed as cfc7cf2 (ships 0.9.16). Closing as superseded.

HerenderKumar:
Makes sense — glad it's fixed and shipping in 0.9.16. Thanks for taking a look!

PR #1884, opened 2026-07-14
fix(dedup): correct id-collision warning and prefer defining file

### LINKED ISSUE #1851: dedup ID-collision warning misdiagnoses cause, overstates data loss, and misses the silent drop (dedup.py:224-246)

SinghAman21:
### Fix the id-collision warning in dedup (wrong cause, overstated loss, silent drop)

Fixes #1851

I spent a while chasing "data loss" that never happened because of this warning,
so here's a fix. The pre-dedup pass in `dedup.py` has three problems:

- **The stated cause is usually wrong.** It blames "two files with the same name
  in different directories," but node ids are `<source-path>_<label>`, so any doc
  that *references* an entity mints the same id as the entity's own node. A
  cross-reference colliding with the definition is dedup working as intended, not
  a filename clash.
- **It overstates the damage.** When the colliding nodes are the same entity,
  edges rewire to the survivor and nothing detaches — only the dropped node's
  label/description is lost. The message tells you to re-run extraction per
  subfolder + `merge-graphs` to avoid a loss that didn't occur.
- **It's silent on the one real drop.** The warning was gated on the source files
  differing, so two same-id nodes from the *same* file (extractor emitting two
  labels for one entity) get dropped with no output at all — which is exactly the
  case where a description disappears unseen.

Also, the survivor was just "first one seen," so which node won depended on chunk
order.

**What changed**
- Survivor selection now prefers the node whose own `source_file` is the one
  encoded in the id (`_node_defines_id`), falling back to first-seen order. The
  authoritative node wins regardless of merge order.
- The warning now fires on actual information loss — a materially different
  label/description, or a genuine cross-file collision — and stays quiet on
  redundant same-entity copies. The previously-silent same-file case is now
  surfaced.
- Reworded the message to name the real id-scheme cause, and it only suggests the
  per-subfolder extract + `merge-graphs` workaround for genuine cross-file
  collisions.

**Testing**
- Updated the two existing collision tests and added two more (defining-file
  tiebreak, and the same-file-different-content warning) — full `test_dedup.py`
  passes.
- Ran the issue's exact 3-node scenario end to end: the defining doc survives,
  the same-file duplicate now warns without pushing re-extraction, the
  cross-reference warns with the workaround, and an identical same-file copy
  stays silent.

### DIFF
--- a/graphify/dedup.py
+++ b/graphify/dedup.py
@@ -9,8 +9,10 @@ import re
 import sys
 import unicodedata
 from collections import defaultdict
+from pathlib import Path
 
 from graphify._minhash import MinHash, MinHashLSH
+from graphify.ids import make_id, normalize_id
 from rapidfuzz.distance import Jaro, JaroWinkler
 
 
@@ -187,6 +189,42 @@ def _is_code(node: dict) -> bool:
     return node.get("file_type") == "code"
 
 
+def _node_defines_id(node: dict, nid: str) -> bool:
+    """True when ``node``'s own ``source_file`` is the file encoded in ``nid``.
+
+    Non-AST node ids are ``make_id(_file_stem(source_file), <label>)``, so the
+    entity's *defining* node has an id equal to its file stem (the file node) or
+    beginning with ``stem + "_"`` (a symbol in that file). A node that merely
+    *references* the entity was extracted from a different file, so its own
+    source_file yields a different stem and this returns False — which is how we
+    tell the authoritative node from a cross-reference that collided onto the
+    same id (#1851)."""
+    sf = node.get("source_file")
+    if not sf:
+        return False
+    from graphify.extractors.base import _file_stem  # local: avoid import cost at module load
+
+    try:
+        stem = make_id(_file_stem(Path(str(sf))))
+    except ValueError:
+        return False
+    if not stem:
+        return False
+    norm_nid = normalize_id(nid)
+    return norm_nid == stem or norm_nid.startswith(stem + "_")
+
+
+def _materially_differs(a: dict, b: dict) -> bool:
+    """True when two same-id nodes carry different content worth surfacing.
+
+    Compares the normalized label and the description: if either differs, keeping
+    one and dropping the other silently discards information. Same-entity copies
+    (identical label and description) return False so the merge stays quiet."""
+    if _norm(a.get("label")) != _norm(b.get("label")):
+        return True
+    return (a.get("description") or "").strip() != (b.get("description") or "").strip()
+
+
 # ── main entry point ──────────────────────────────────────────────────────────
 
 def deduplicate_entities(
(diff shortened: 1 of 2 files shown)


safishamsi:
Thanks @SinghAman21. Went with #1852 here (it also silences the harmless cross-reference collision and handles absolute source_file paths); landed in 0.9.16 with a deterministic tiebreak on top.

Issue #1880, opened 2026-07-14
Regression in 0.9.15: `graphify update` emits 0 nodes + refuses to overwrite when existing graph.json has no chunk cache (worked in ≤0.9.14)


anshulsao:
## Regression in 0.9.15: `graphify update` emits 0 nodes and refuses to overwrite when an existing `graph.json` has no chunk cache

`graphify update <src> --no-cluster`, run against an existing `graphify-out/graph.json` whose per-file **chunk cache is not present** alongside it, now produces:

```
[graphify] WARNING: new graph has 0 nodes but existing graph.json has <N>.
Refusing to overwrite — you may be missing chunk files from a previous session. Pass --force to override.
```

In **≤0.9.14** the exact same invocation rebuilt the graph from source and merged successfully (>0 nodes). **0.9.15** produces 0 nodes and refuses.

### Reproduction (controlled bisect)
Same source tree + same seeded `graph.json` (a real ~24k-node graph, no chunk cache written beside it) — **only the graphify version varied**:

| graphify | result |
|---|---|
| 0.9.13 | ✅ full graph built |
| 0.9.14 | ✅ full graph built |
| 0.9.15 | ❌ `new graph has 0 nodes … Refusing to overwrite` |

### Expected
When the chunk cache is absent but a `graph.json` exists, `update` should fall back to a full rebuild from source (as ≤0.9.14 did) instead of emitting 0 nodes and refusing — or gate the refusal so this legitimate "seed graph.json without chunk cache" case isn't a hard failure.

### Impact
Incremental-update workflows that seed from a prior publish's `graph.json` (+ label sidecars) **without** carrying the chunk cache — a common pattern — break on 0.9.15 with no code change on their side, since the tool is typically installed unpinned (latest).

### Version boundary
- Last good: **0.9.14** (uploaded 2026-07-13 09:39 UTC)
- First bad: **0.9.15** (uploaded 2026-07-13 22:20 UTC)

safishamsi:
Fixed in 0.9.16 (fb4d452 + regression test 40eae3c). Same root cause as #1873/#1887: a nested broad .gitignore zeroed update's re-scan. With subtree scoping fixed, update sees the real files again.

PR #1885, opened 2026-07-14
fix(detect): scope nested ignore patterns to their own subtree

### LINKED ISSUE #1873: Nested .gitignore rules apply outside their own subtree, zeroing the scan
## Summary
`detect()` loads nested ignore files during the walk (`_load_dir_own_ignore(dp)`, ref #1206) and appends their patterns to the shared `ignore_patterns` list — but `_is_ignored` then applies those patterns to paths **outside the loading directory's subtree**. Git scopes a nested .gitignore to its own directory tree; graphify applies it globally.

## Repro
```
mkdir -p repo/tooldir repo/src
echo '*' > repo/tooldir/.gitignore        # common pattern in tool-managed dirs
echo 'x = 1' > repo/src/app.py
cd repo && python -c "
from pathlib import Path
print(detect(Path('.'))['total_files'])   # 0 — expected: 1 (src/app.py)
"
```

Because the walk visits `tooldir` (alphabetically) before `src`, its `*` pattern is in the list when `src/app.py` is evaluated → every file in the repo is ignored and detect returns an empty corpus with no warning.

Hit in the wild: a Claude Code plugin state dir (`.superpowers/sdd/.gitignore` containing `*`) silently zeroed a whole-repo scan (v0.9.15).

## Direct evidence
```python
from graphify.detect import _load_dir_own_ignore, _is_ignored
pats = _load_dir_own_ignore(root / 'tooldir')    # [(<root>/tooldir, '*')]
_is_ignored(root / 'src/app.py', root, pats)      # True — should be False
```

## Expected
A pattern loaded from `dir/.gitignore` should only match paths under `dir/` (the tuple already carries the base dir — the matcher just needs to enforce it).

## Workaround
Root `.graphifyignore` excluding the offending directory (so the walk prunes it before loading its ignore file).

Aside, minor: `**/name` patterns from the global git ignore also fail to match their own target (`_is_ignored(root/'.claude/settings.local.json', ...)` is False for pattern `**/.claude/settings.local.json`) — opposite direction, may share a root cause in the matcher.

Alwyn93:
## Summary

Fixes #1873 — nested `.gitignore`/`.graphifyignore` patterns applied outside their own subtree, which can zero an entire scan.

`_is_ignored()` matched **non-anchored** patterns against the path relative to the scan **root** first, so patterns from a nested ignore file leaked out of their directory. In-the-wild trigger: the `hypothesis` library auto-writes `.hypothesis/.gitignore` containing a bare `*` — with nested-ignore support (8a5287a, #1206) that single file ignored the whole repository and `detect()` returned **0 files**.

## Fix

Per gitignore semantics, patterns from `A/.gitignore` apply only to paths **under `A`**. The patch matches every pattern against the path relative to its own anchor:

- target outside the pattern's anchor → pattern cannot match (`continue`)
- anchor directory itself is exempt (`rel_anchor == "."`) — an ignore file governs its directory's contents, not the directory
- anchored (`/`-prefixed) and non-anchored patterns unified on the same anchor-relative path; last-match-wins and `!` negation behavior unchanged

## Tests

- `tests/test_detect.py`: **143 passed** — including the existing nested-ignore, nested-negation-beats-root, and `git info/exclude` precedence tests
- Two new regressions:
  - `test_nested_gitignore_star_does_not_ignore_outside_its_dir` — the `.hypothesis` scenario; fails on `v8` HEAD (`total_files == 0`), passes with the fix
  - `test_nested_gitignore_patterns_still_apply_inside_their_dir` — guard that nested ignore files still work within their subtree
- Full-suite note: the 15 failures in `tests/test_skillgen.py` (11) and `tests/test_ollama_retry_cap.py` (4) are pre-existing at `v8` HEAD (verified on a clean checkout) and unrelated to this change.

Also verified on a real repo: a scan that returned 0 files at HEAD returns 204/230 with the fix.

## Related

- Same root cause as reported in #1873 (see my confirming comment there with the diagnosis).
- Separate but adjacent: `graphify update` / `--watch` re-run detection without the CLI `extra_excludes`, so a workaround via CLI excludes doesn't survive watch mode — happy to file/fix separately if useful.

### DIFF
--- a/graphify/detect.py
+++ b/graphify/detect.py
@@ -955,25 +955,18 @@ def _is_ignored(
             if not p:
                 continue
 
+            # gitignore semantics: patterns from A/.gitignore apply ONLY to paths
+            # under A. Matching non-anchored patterns against root-relative paths
+            # let e.g. .hypothesis/.gitignore's bare "*" ignore the ENTIRE repo
+            # (detect() returned 0 files). The anchor dir itself is exempt — an
+            # ignore file governs its directory's contents, not the directory.
             matched = False
-            if anchored:
-                try:
-                    rel_anchor = str(target.relative_to(anchor)).replace(os.sep, "/")
-                    matched = _matches(rel_anchor, p, anchored=True)
-                except ValueError:
-                    pass
-            else:
-                try:
-                    rel = str(target.relative_to(root)).replace(os.sep, "/")
-                    matched = _matches(rel, p, anchored=False)
-                except ValueError:
-                    pass
-                if not matched and anchor != root:
-                    try:
-                        rel_anchor = str(target.relative_to(anchor)).replace(os.sep, "/")
-                        matched = _matches(rel_anchor, p, anchored=False)
-                    except ValueError:
-                        pass
+            try:
+                rel_anchor = str(target.relative_to(anchor)).replace(os.sep, "/")
+            except ValueError:
+                continue  # target outside this pattern's anchor: cannot match
+            if rel_anchor != ".":
+                matched = _matches(rel_anchor, p, anchored=anchored)
 
             if matched:
                 result = not negated  # last match wins; ! flips to un-ignore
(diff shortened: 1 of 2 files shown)


roslinmahmud (approved):

safishamsi:
Landed on v8 as fb4d452, ships in 0.9.16. Thanks @Alwyn93 - subtree-scoping fix was exactly right and also resolved the #1880 update-emits-0-nodes symptom.

PR #1912, opened 2026-07-15
fix: treat .cjs (explicit CommonJS) as a code extension

Kookwater:
## Problem

`.cjs` is half-registered. The language maps in `build.py` and `extract.py` already route it to the JS grammar:

```python
# build.py
".js": "js", ".mjs": "js", ".cjs": "js", ...
# extract.py
".js": "jsts", ".jsx": "jsts", ".mjs": "jsts", ".cjs": "jsts",
```

…but the extension is missing from `CODE_EXTENSIONS` in `detect.py`, the extractor `_DISPATCH`, `_LANG_FAMILY` in `analyze.py`, and the JS resolution/cache suffix sets. So `.cjs` sources are **silently skipped** during a build — classified as non-code and never handed to the JS extractor:

```
[graphify extract] 7 file(s) not classified (no supported extension or shebang),
skipped: ..., main.cjs, preload.cjs
```

`.cjs` is not exotic: it's the standard extension for Electron main/preload scripts and for CommonJS escape hatches inside `"type": "module"` packages.

## Fix

Add `.cjs` alongside `.mjs` in the six lists that gate/route JS files:

| File | List |
|---|---|
| `graphify/detect.py` | `CODE_EXTENSIONS` |
| `graphify/extract.py` | `_DISPATCH` (`.cjs` → `extract_js`) |
| `graphify/analyze.py` | `_LANG_FAMILY` |
| `graphify/extractors/models.py` | `_JS_CACHE_BYPASS_SUFFIXES` |
| `graphify/extractors/resolution.py` | `_JS_RESOLVE_EXTS` |
| `graphify/cli.py` | `_HOOK_SOURCE_EXTS` |

This is the same shape of gap (and fix) as `.mts`/`.cts` in #1607.

## Tests

`tests/test_cjs_module_extension.py` mirrors `tests/test_typescript_module_extensions.py`: regression locks for each extension set, plus an end-to-end `extract_js` run proving a representative CommonJS source (`require()` imports, class, function, `module.exports`) parses identically as `.cjs` and `.js`.

```
tests/test_cjs_module_extension.py  6 passed
tests/test_typescript_module_extensions.py + tests/test_detect.py  153 passed
tests/test_js_import_resolution.py + tests/test_import_extension_resolution.py  89 passed
```

## Real-world impact

On an Electron app whose ~2000-line `main.cjs` backend was invisible to the graph, rebuilding after this fix went from **201 → 294 nodes and 256 → 441 edges** — the main-process file alone contributed 91 nodes.


safishamsi:
Landed on v8 in 2851f31 (ships in 0.9.17). Thanks @Kookwater - .cjs is now in CODE_EXTENSIONS and the dispatch map, routed to the JS grammar; verified across all six sets.

PR #1914, opened 2026-07-15
fix: prune scan-excluded files from manifest during full rebuild

### LINKED ISSUE #1908: save_manifest retains scan-excluded files, causing permanent false deleted_files
## Bug

`save_manifest()` retains a manifest entry when its path still exists on disk, even if the file is no longer returned by the current `detect()` scan. `detect_incremental()` then treats that same entry as deleted because it compares the manifest against the scan result.

This produces a permanent false-positive deletion on every incremental run for generated or ignored files.

## Observed with

- graphifyy `0.9.16`
- Windows 11 / Python 3.11
- Full AST rebuild via `graphify update .`

## Reproduction

1. Run Graphify on a repository so `graphify-out/manifest.json` contains a supported file.
2. Keep that file physically present, but make it excluded from Graphify's current scan (for example through the repository's ignore rules / generated-file handling).
3. Run a full AST rebuild:

   ```bash
   graphify update .
   ```

4. Inspect the results:

   - `detect(root)` does not return the file;
   - `save_manifest(detected["files"], kind="ast", root=root)` retains its old manifest row because `Path(f).exists()` is true;
   - `detect_incremental(root, kind="ast")` reports the row in `deleted_files` because it is absent from the current scan.

In my case, both `apps/web/next-env.d.ts` and `.impeccable/hook.cache.json` existed on disk but were excluded from the scan. They remained in `manifest.json`, so every AST check reported two deletions until those two stale rows were removed.

## Expected behavior

After a full-root scan/rebuild, the manifest should contain exactly the scanned corpus (plus any explicitly supported out-of-root entries). A physically present but scan-excluded file must not remain as a stale manifest entry.

## Suggested fix

Keep the current subset-preserving behavior for partial incremental callers, but add an explicit full-scan pruning mode, e.g.:

```python
save_manifest(files, kind="ast", root=project_root, prune_to_scan=True)
```

When `prune_to_scan=True`, prune existing entries that are absent from the supplied full `detected["files"]` corpus instead of deciding retention solely with `Path(f).exists()`. Use that mode from `graphify.watch._rebuild_code`, which already has the complete result of `detect(watch_path)`.

The default should remain subset-preserving so callers that pass only changed files do not erase unrelated manifest rows.

## Regression tests

- A scan-excluded file that still exists on disk is pruned during a full rebuild.
- A physically missing file is pruned.
- A partial/subset manifest save still preserves untouched entries.

## Impact

The graph itself can be valid, but the daily incremental status never reaches clean (`deleted_files != 0`) and repeatedly triggers needless rebuild work.

kimdzhekhon:
## Summary
- `save_manifest()` only dropped a manifest entry when its file was physically gone from disk, so a file still present but excluded from the current `detect()` scan (newly ignored, generated, etc.) stayed in the manifest forever.
- `detect_incremental()` then reported that entry in `deleted_files` on every subsequent run — a permanent false positive.
- Adds `save_manifest(prune_to_scan=True)`: when set, entries absent from the current full-scan corpus are pruned even if the file still exists. Default (`False`) is unchanged, so partial/incremental callers passing a subset of files still preserve untouched entries (#917).
- Wired into all three `save_manifest` call sites in `watch._rebuild_code`, which already holds the complete `detect(watch_path)` result for the watched root.

Fixes #1908.

(Re-opened from a fresh fork; identical content to the earlier #1913, which was closed to purge a stray commit-author email from fork history.)

## Test plan
- [x] Added `test_save_manifest_prune_to_scan_drops_scan_excluded_existing_file` — reproduces the reported bug (physically-present, scan-excluded file is pruned)
- [x] Added `test_save_manifest_prune_to_scan_still_drops_missing_files` — existing missing-file pruning still works under the new flag
- [x] Added `test_save_manifest_without_prune_to_scan_preserves_subset_entries` — default behavior (#917) unchanged
- [x] `uv run pytest tests/test_detect.py tests/test_watch.py -q` — all pass
- [x] `uv run pytest tests/ -q` — full suite: 3189 passed (5 pre-existing failures unrelated to this change, missing optional `openai` dependency in this environment)

### DIFF
--- a/graphify/detect.py
+++ b/graphify/detect.py
@@ -1466,6 +1466,7 @@ def save_manifest(
     *,
     kind: str = "both",
     root: Path | None = None,
+    prune_to_scan: bool = False,
 ) -> None:
     """Save current file mtimes + content hashes for change detection.
 
@@ -1481,6 +1482,13 @@ def save_manifest(
     machines and checkout locations (#777). Out-of-root entries are written
     as absolute so they continue to round-trip on the saving machine.
     When ``root`` is None the legacy absolute-keyed format is preserved.
+
+    ``prune_to_scan`` (default False) — when True, ``files`` is treated as the
+    complete corpus of a full-root scan: existing manifest entries absent
+    from it are pruned even if the underlying file still exists on disk (a
+    physically-present-but-scan-excluded file, e.g. now ignored or
+    generated). Leave this False for callers passing only a changed subset,
+    so unrelated manifest rows are preserved (#917).
     """
     existing = load_manifest(manifest_path, root=root)
 
@@ -1493,14 +1501,21 @@ def save_manifest(
             return entry
         return None
 
+    all_files = [f for file_list in files.values() for f in file_list]
+    scanned = set(all_files) if prune_to_scan else None
+
     # Seed from the existing manifest so incremental callers passing a subset
     # of files don't silently erase entries for untouched files (#917).
     # Prune entries whose file no longer exists on disk — those are genuine
-    # deletions that detect_incremental() should treat as gone.
+    # deletions that detect_incremental() should treat as gone. When
+    # ``prune_to_scan`` is set, also prune entries absent from the current
+    # full scan even if the file still physically exists.
     manifest: dict[str, dict] = {}
     for f, entry in existing.items():
         normalised = _normalise_entry(entry)
         if normalised is None:
+            continue
+        if scanned is not None and f not in scanned:
             continue
         try:
             if Path(f).exists():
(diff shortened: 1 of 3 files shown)


safishamsi:
Thanks @kimdzhekhon - this is the same fix I landed on v8 in 75cf56b (ships in 0.9.17): save_manifest now prunes scan-excluded-but-present files during a full rebuild (pruning against the full detect corpus, never out-of-root entries), and detect_incremental distinguishes truly-deleted files from excluded-but-present ones so they are no longer reported as deleted. Closing as superseded; appreciate the fix and the clear diagnosis.

Issue #1895, opened 2026-07-14
deep mode: nodes with out-of-scope source_file are warned about but still written to graph.json


Nyxx0x:
**Version:** 0.9.15

Deep-mode extraction created nodes whose `source_file` points at files that `.graphifyignore` excludes and that were never dispatched to the extractor. The LLM saw *references* to those files inside documents it did read, and set `source_file` to the referent rather than to the containing document. The node therefore asserts a provenance that never existed.

graphify partially detects this. Per offending file it emits:

```
RuntimeWarning: semantic cache skipped out-of-scope source_file '<path>';
the file was not dispatched for extraction
```

and correctly keeps the entry out of the cache — **but the node is still written to `graph.json`.** So the warning implies the situation was handled, while the graph carries the fabricated provenance.

### Observed

3 warnings, and exactly 3 corresponding out-of-scope nodes in the final graph (paths under `scripts/` and `spine/attestations/`, excluded by the ignore file's globs). The warning count was a reliable detector; the graph was not actually protected by it.

### Expected

The same out-of-scope check that guards the cache write should also run before the graph write — dropping the node, re-attributing `source_file` to the containing document, or at minimum flagging it in the node so downstream consumers can filter it.

Related: #1757 fixed the cache-overwrite consequence of this same `source_file` mis-attribution. The graph-write consequence appears to remain.

safishamsi:
Fixed on v8 in 0cacd70 (will ship in 0.9.17). A node whose source_file was never dispatched is now dropped from the merged graph, not just skipped from the cache - closing the graph-side gap of the #1757 cache guard.

PR #1916, opened 2026-07-15
fix(cache): prune edges/hyperedges referencing never-written node groups

obbax:
`save_semantic_cache()` groups nodes/edges/hyperedges by `source_file`
and writes one cache entry per file, but two classes of node groups are
never written at all:

1. **Out-of-scope** (#1757 guard): the group's file exists but was not
   in `allowed_source_files` — skipped with a warning.
2. **Ghost paths**: the group's `source_file` does not resolve to a
   real file (a hallucinated path) — silently dropped by the
   `if p.is_file()` main-loop condition, which has no else branch.

In both cases, an edge or hyperedge attributed to a *different*,
allowed file could still reference one of those never-written node
ids. That reference survived into the allowed file's cache entry with
no reachable endpoint anywhere in the cache — a dangling edge on every
`check_semantic_cache()` read (observed in the wild as 428
dangling-endpoint edges reported by `diagnose_extraction`).

Fix: when a scope is given, pre-scan the node ids of every group the
write loop will skip — `not p.is_file() or p not in allowed_paths`,
mirroring the loop's full skip condition — and prune any edge/hyperedge
in an allowed group that references one of them. This mirrors the
merged-graph out-of-scope filter already applied to `graph.json` in
`llm.py` (#1895), which covered the in-memory result but not this
cache-level write path. Callers without `allowed_source_files` are
unaffected.

`tests/test_cache.py` adds two regressions: an edge targeting a
dropped out-of-scope node, and an edge targeting a ghost-path node
(`source_file` that does not exist on disk). Both fail without the
fix. Full suite: 3172 passed, 31 skipped, failure set byte-identical
to the pre-existing environment-local baseline.


safishamsi:
The underlying bug is fixed directly on v8 in d916c61 (ships in 0.9.17), rather than merging this PR (its diff is very large / includes a vendored tree). save_semantic_cache now prunes edges/hyperedges referencing skipped (out-of-scope or ghost) node groups at write time, and build_from_json validates hyperedge members against the node set so a dangling hyperedge can't reach graph.json. Thanks @obbax for the diagnosis.

Issue #1889, opened 2026-07-14
perf(serve): eliminate repeated per-term graph scoring in _query_graph_text


Sirhan1:
## Summary

`_query_graph_text()` currently performs **T+1 graph-scoring passes** for a query with **T** unique terms:

- one combined `_score_nodes()` call to produce the ranked results;
- one additional `_score_nodes([term])` call per unique token inside `_pick_seeds()` to guarantee per-term seed coverage (introduced in #1596).

This means every query rescans the graph once for the combined query and then once again for every distinct search term.

## Proposal

Replace the current T+1 scoring flow with a single candidate-scoring traversal.

Introduce a private helper (for example `_score_query()`) that computes both:

- the existing combined ranking;
- the best matching node for each normalized query term.

`_pick_seeds()` would then consume the precomputed per-term winners instead of rescoring the graph once per token.

The goal is to preserve existing behaviour while eliminating the redundant scoring passes.

## Why this should be equivalent

The combined trigram candidate set (`norm_terms + [joined]`) is a superset of every individual per-token candidate set.

Therefore every node that could receive a non-zero singleton score is already visited during the combined scoring traversal, allowing the per-term winners to be computed without additional graph scans.

The intention is to preserve:

- combined ranking
- per-term seed selection
- existing tie-breaking behaviour
- existing seed deduplication
- existing trigram prefilter behaviour

while changing only the implementation.

## Prototype results

I implemented a prototype to validate the approach.

Across synthetic graphs ranging from **1k to 100k nodes**, together with Graphify's own ~11k node graph, I observed:

| Scenario | Improvement |
|----------|------------:|
| 1k–100k synthetic graphs | **1.4×–2.4× faster** |
| 100k graph (3-term query) | **~41% lower latency** |
| 100k graph (10-term query) | **~35% lower latency** |
| 11k real graph | **1.8×–2.0× faster** |

Single-term queries also improve because the current implementation still performs two scoring passes.

## Scope

This would be an internal refactor only.

No intended changes to:

- CLI
- MCP
- graph format
- public API
- query behaviour

## Current PR

https://github.com/Graphify-Labs/graphify/pull/1918

safishamsi:
[...]

PR #1918, opened 2026-07-15
perf(serve): collapse T+1 per-term scoring passes into one traversal

### LINKED ISSUE #1445: BFS path drift on vague Chinese queries: 40-node corpus, 12 unrelated nodes returned for "rate_limit_check 怎么用？"

Sirhan1:
## Summary

`_query_graph_text()` currently performs **T+1 graph-scoring passes** for a query with T unique terms:

1. one combined `_score_nodes()` call to produce the ranked results;
2. one additional `_score_nodes([term])` call per unique token, inside `_pick_seeds()`, to guarantee per-term seed coverage (introduced in #1596, fixes #1445).

This means every query rescans the graph once for the combined query and then once again for every distinct search term.

## Proposal

Replace the current T+1 scoring flow with a **single candidate-scoring traversal**.

Introduce a private helper (for example `_score_query()`) that computes both:

- the existing combined ranking;
- the best matching node for each normalized query term.

`_pick_seeds()` would then consume the precomputed per-term winners instead of rescoring the graph once per token.

The goal is to preserve existing behaviour while eliminating the redundant scoring passes.

## Why this should be equivalent

The combined trigram candidate set (`norm_terms + [joined]`) is a superset of every individual per-token candidate set.

Therefore every node that could receive a non-zero singleton score is already visited during the combined scoring traversal, allowing the per-term winners to be computed without additional graph scans.

The intention is to preserve:

- combined ranking
- per-term seed selection
- existing tie-breaking behaviour
- existing seed deduplication
- existing trigram prefilter behaviour

while changing only the implementation.

## Benchmark results

I implemented a prototype to validate the approach.

Across synthetic graphs ranging from 1k to 100k nodes, together with Graphify's own ~11k node graph, I observed:

| Scenario | Improvement |
| --- | --- |
| 1k–100k synthetic graphs | 1.4×–2.4× faster |
| 100k graph (3-term query) | ~41% lower latency |
| 100k graph (10-term query) | ~35% lower latency |
| 11k real graph | 1.8×–2.0× faster |

Single-term queries also improve because the current implementation still performs two scoring passes.

## Scope

This is an **internal refactor only**.

No intended changes to:

- CLI
- MCP
- graph format
- public API
- query behaviour

## Verification

- Tests: 3221 passed, 3 skipped
- Ruff clean
- Graphify graph updated

Refs #1889

### DIFF
--- a/graphify/serve.py
+++ b/graphify/serve.py
@@ -6,6 +6,7 @@ import re
 import sys
 from array import array
 from pathlib import Path
+from typing import NamedTuple
 import networkx as nx
 from networkx.readwrite import json_graph
 from graphify.security import sanitize_label, check_graph_file_size_cap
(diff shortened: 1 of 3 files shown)


safishamsi:
Landed on v8 in ff999a9 plus a small rebase fixup c0875cc (ships in 0.9.17). Thanks @Sirhan1 - verified behavior-preserving via a differential fuzz against the old per-term scoring, ~1.3-1.4x faster and query-length-independent. The fixup only updated the #1900 stopword test to your new _score_query/best_seed_by_term API (that test post-dated your branch) and added the bench file's trailing newline.

PR #1919, opened 2026-07-15
fix(detect): prune scan-excluded files from manifest on full rebuild

### LINKED ISSUE #1908: save_manifest retains scan-excluded files, causing permanent false deleted_files

Yashasm18:
Fixes #1908.

save_manifest kept a manifest entry as long as the file still existed on
disk, even when the current scan no longer returned it (ignore rules,
generated-file handling). detect_incremental then compared the manifest
against the scan and flagged those entries as deleted — permanently, on
every incremental run.

Added a prune_to_scan flag to save_manifest: when True, an entry only
survives if it's both on disk and present in the current files corpus.
Wired it into watch._rebuild_code's three full-rebuild call sites, which
already have the complete detect() result. Default stays False
(subset-preserving), so incremental callers passing only changed files
still don't erase untouched entries (#917).

Added the three regression tests the issue specified:
- scan-excluded-but-on-disk file is pruned under prune_to_scan=True
- physically deleted file is pruned under prune_to_scan=True
- a subset save under the default still preserves untouched entries


safishamsi:
Thanks @Yashasm18 - this is the same fix I landed on v8 in 75cf56b (ships in 0.9.17): save_manifest prunes scan-excluded-but-present files on a full rebuild (against the full detect corpus, never out-of-root entries) and detect_incremental distinguishes deleted from excluded-but-present. Closing as superseded; appreciate the fix.

Issue #1924, opened 2026-07-15
Incremental rebuild is non-idempotent: preserved node's ID gains its parent-dir prefix on every run


carlos-reyes-le-paliscot:
## Summary

A node whose source file sits in a dot-directory (here: `.claude/CLAUDE.md`, the pointer file `graphify install --project` itself writes) has its ID re-prefixed with the parent directory segment on **every** incremental rebuild, so `graph.json` never converges to a stable byte state.

## Repro

graphifyy 0.9.15, macOS, Python 3.11 (uv tool). Corpus ~370 files with `.claude/CLAUDE.md` indexed and a committed graph:

```
PYTHONHASHSEED=0 graphify . --update --no-viz --code-only   # run N
PYTHONHASHSEED=0 graphify . --update --no-viz --code-only   # run N+1
diff runN/graph.json runN+1/graph.json
```

Observed — the same node, one more prefix per run (everything else byte-identical):

```
<   "id": "claude_claude_claude_claude_claude_claude_claude_graphify_trigger"
>   "id": "claude_claude_claude_claude_claude_claude_claude_claude_graphify_trigger"
```

Node metadata: `label: "Graphify Trigger Pointer"`, `source_file: ".claude/CLAUDE.md"`, `file_type: "document"`. Logs report `0 re-extracted, 0 deleted` on both runs — the mutation happens in the preserved-node path, not extraction.

## Expected

Preserved-node ID normalization should be idempotent: re-running an incremental rebuild over an unchanged corpus must reproduce identical IDs (and ideally identical bytes).

## Workaround

Exclude the file via `.graphifyignore` (it's graphify's own 4-line install pointer, no graph value):

```
.claude/CLAUDE.md
```

After exclusion + one rebuild, consecutive runs are byte-identical.

Related: #1915 (same repo/setup). Our workaround commit: https://github.com/carlos-reyes-le-paliscot/esg-bowler-chart/pull/167

🤖 Reported via [Claude Code](https://claude.com/claude-code)

safishamsi:
[...]

Issue #1917, opened 2026-07-15
`_semantic_id_remap` lacks the `graph_has_legacy_ids` idempotency guard: node IDs accrete a stem prefix every build, defeating the same_topology/no_change short-circuits


oraphadev:
## Environment

- Package: `graphifyy` (PyPI) — installed via `uv tool install graphifyy`
- Version: 0.9.4 (confirmed still present in 0.9.5 — see below)
- Python: 3.14
- OS: macOS (Darwin), but the bug is platform-independent (pure string logic)

## Summary

`_semantic_id_remap()` in `graphify/build.py` re-derives non-AST node IDs from `source_file` on every `build()`/`build_merge()` call, to reconcile pre-migration short IDs with the canonical full-path-stem form (#1504/#1509). It has no "already canonical" guard before matching against the legacy stem forms in `_old_file_stems()`. When a file's canonical stem contains its own legacy stem as a literal prefix — which happens whenever the immediate parent directory name equals the file's own stem, e.g. `.claude/CLAUDE.md` → canonical stem `claude_claude`, legacy stem `claude` — every already-migrated ID still matches the legacy branch (`"claude_claude_X".startswith("claude_")` is true), so the function re-derives it and **prepends another copy of the stem**. The ID grows by one `claude_` segment every single build/update, unbounded, forever.

This is a regression gap left by #1504/#1509: the sibling function `graph_has_legacy_ids` (same file, used by the read-only query/serve path to *detect* this exact drift) already carries the correct guard:

```python
if norm == new_stem or norm.startswith(new_stem + "_"):
    checked += 1
```

but `_semantic_id_remap` — the function that actually *performs* the remap during build — never got the equivalent check.

## Reproduction

Deterministic, no LLM/network involved — calls `_semantic_id_remap` directly with a single semantic-origin node and applies the returned remap to itself repeatedly, simulating N consecutive `graphify update` runs on an unchanged file:

```python
from graphify.build import _semantic_id_remap

def one_build_pass(node_id: str, source_file: str) -> str:
    nodes = [{
        "id": node_id,
        "source_file": source_file,
        "label": "trigger",
        # no "_origin" key => treated as non-AST ("semantic"), which is the
        # code path _semantic_id_remap operates on
    }]
    remap = _semantic_id_remap(nodes, root=None)
    return remap.get(node_id, node_id)

node_id = "claude_graphify_trigger"          # legacy short id (stem "claude")
[...]

PR #1926, opened 2026-07-15
fix(build): make semantic id-remap idempotent for dot-dir files (#1924)

SinghAman21:
Fix non-idempotent node IDs for files in dot-directories (#1924)

Running an incremental rebuild twice over an unchanged corpus should produce an identical graph. For files under a dot-directory it doesn't — .claude/CLAUDE.md's node ID gains one claude_ every run and never converges (claude_claude_graphify_trigger → claude_claude_claude_graphify_trigger → …). Logs show 0 re-extracted, so it's the preserved-node path, not extraction.

Cause: _semantic_id_remap() rewrites legacy IDs to the canonical full-path stem. For a dot-dir file the canonical stem (claude_claude) contains the shorter legacy stem (claude) as a prefix. The guard only skipped old_stem == new_stem, so an already-canonical ID still matched the claude_ prefix, lost one segment, and got the full stem prepended back — +1 each run.

Fix: skip the remap when the ID is already canonical (== new_stem or starts with new_stem + "_"), the same guard graph_has_legacy_ids() already uses. Legitimate legacy migration still runs once.

Testing: added tests/test_semantic_id_remap_idempotent.py (stable across reruns, legacy forms still migrate once); build/merge/watch/id suites pass.


safishamsi:
Thanks for this. The same dot-dir id-accretion bug (#1924) was already fixed on v8 by #1917 (19c496d): the idempotency guard in _semantic_id_remap skips an id that already carries its canonical stem, so a file like .claude/CLAUDE.md migrates once and then stays stable. Verified against the #1924 repro. Closing as already-resolved rather than merging a second fix for the same path — much appreciated regardless.

SinghAman21:
hey @safishamsi  thanks for your explanation. 
could you suggest me which issue i should work on next?

Issue #1854, opened 2026-07-13
`--postgres`: foreign keys are dropped, and the schema is disconnected from the code


qt1bo:
## Summary

Two problems when importing a live DB into a code graph via `--postgres` (graphify 0.9.14):

1. **Bug — FK relationships are lost.** The help says `--postgres` maps "tables, views,
   functions **+ FK relationships**", but the graph gets **0 FK edges**. My DB declares
   **207** `FOREIGN KEY` constraints; `graph.json` contains **0** `table → table` edges.
2. **Feature gap — the schema is an island.** Imported tables form a star (one DB hub
   → tables via `contains`) with **0 edges to the code that maps them** (ORM entities,
   repositories). The schema can't take part in any code↔DB traversal.

## Environment

| Component | Version |
|---|---|
| graphify | 0.9.14 |
| PostgreSQL server | 17.2 |
| Doctrine ORM | 2.20.9 |
| Doctrine DBAL | 3.10.4 |
| DoctrineBundle | 2.13.3 |
| Symfony | 5.4.45 |
| Backend | claude-cli (claude-opus-4-8) |

## Repro & evidence

```bash
graphify extract . --postgres "postgresql://<user>:<pass>@localhost:5432/<db>"
# log: PostgreSQL: 152 nodes, 163 edges
```

Live DB vs. `graph.json`:

| | DB (real) | graph.json |
|---|---|---|
| tables + views | 148 + 3 | 148 + 3 ✅ |
| FOREIGN KEY constraints | **207** | **0** ❌ |
| `table → table` edges | 207 expected | 0 |
| DB→code edges | — | 0 |

FK count verified with:

```sql
SELECT constraint_type, count(*) FROM information_schema.table_constraints
WHERE table_schema='public' GROUP BY constraint_type;
--  FOREIGN KEY | 207
```

The only DB edges written are 151 `contains` (hub → object). Note the log's `163 edges`
matches neither the 151 written nor the 207 real FKs.

## Expected / request

1. Emit FK edges from `pg_constraint` / `information_schema.referential_constraints` as
   `table → table` edges.
2. Provide a way to link tables to code. For ORM stacks the table name is in the source
   (e.g. Doctrine `#[ORM\Table(name: 'x')]`), so `entity → table` edges could be matched
   deterministically — or just document the DB node ID scheme so users can inject bridges.

## Minor note

DB nodes carry `file_type: "code"` and no `kind`/`_origin`; the only way to tell a table
[...]

PR #1927, opened 2026-07-15
fix(pg_introspect): emit FK DDL before function stubs so unparseable routines can't drop references edges (#1854)

sekmur:
## Problem

The dropped-FK half of #1854 (closed as fixed-by-#1746, but reproducible on a fresh 0.9.16 — `graphify/pg_introspect.py` and `graphify/extractors/sql.py` are byte-identical from v0.9.14 through v0.9.16).

`introspect_postgres()` reconstructs one synthetic DDL document in the order *tables → views → functions → FKs* and hands it to `extract_sql()` in a single tree-sitter parse. A C-language routine's `routine_definition` is just the C symbol name, so its stub:

```sql
CREATE FUNCTION "public"."levenshtein"() RETURNS void AS $gfx$ levenshtein $gfx$ LANGUAGE c;
```

is unparseable, and tree-sitter's error recovery consumes the statements *after* it. With the FK `ALTER TABLE`s emitted last, they're what gets eaten — so any DB with a common extension installed (uuid-ossp, pgcrypto, pg_trgm, fuzzystrmatch, unaccent, PostGIS, …) silently loses its `references` edges. This is distinct from #1746 (FK *reading* privileges): the failure is downstream of a correct catalog read, in the parse.

## Fix

Emit the FK block before the routine DDL — a pure block move. FK statements only reference tables, which are emitted first, so the order is always safe; an unparseable stub can now only damage what follows it (the remaining stubs, which contribute no edges from their bodies anyway).

Measured on a real DB (35 tables / 41 FKs / 41 extension routines from fuzzystrmatch + pg_trgm): **1/41** references edges before, **41/41** after.

## Test

`test_pg_introspect_fk_edges_survive_unparseable_function_stubs` — 7 tables, 6 FKs, one `LANGUAGE c` routine + one plpgsql routine, asserts all 6 `references` edges survive. Fails on current v8 (0 or partial edges, statement-mix-dependent), passes with the fix. The existing baseline test never caught this because its mock routines only use `SQL`/`PLPGSQL` languages, never `C`.

Longer term this path may not need a parse at all — `introspect_postgres()` already holds tables and FK pairs as structured rows from `pg_catalog`, so emitting nodes/edges directly (instead of round-tripping through synthetic DDL) would make this failure class impossible. This PR is the minimal ordering fix; happy to discuss the direct-emission approach separately.

### DIFF
--- a/graphify/pg_introspect.py
+++ b/graphify/pg_introspect.py
@@ -113,6 +113,22 @@ def introspect_postgres(dsn: str | None = None) -> dict:
         else:
             ddl.append(f"CREATE VIEW {_quote_ident(schema)}.{_quote_ident(name)} AS SELECT 1;")
 
+    # FK edges — one ALTER TABLE per constraint (handles composite FKs correctly).
+    # Emitted BEFORE the function DDL: routine bodies the grammar can't parse
+    # (notably C-language extension functions, whose "body" is just the C symbol
+    # name) put tree-sitter into error recovery that consumes the statements
+    # after them — with FKs last, every 'references' edge is silently lost on
+    # any DB with a common extension installed (#1854). FK statements only
+    # reference tables, which are emitted first, so this order is always safe.
+    for constraint_name, t_schema, t_name, cols, r_schema, r_name, r_cols in fks:
+        col_list = ", ".join(_quote_ident(c) for c in cols)
+        ref_col_list = ", ".join(_quote_ident(c) for c in r_cols)
+        ddl.append(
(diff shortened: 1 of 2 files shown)


safishamsi:
Landed on v8 in 5840d0b (ships in 0.9.18), cherry-picked so your authorship is preserved. Thanks @sekmur — emitting the FK ALTER TABLE statements before the function stubs stops an unparseable routine (e.g. a C-language function) from swallowing the trailing FK statements into a tree-sitter ERROR region. Verified: the table-to-table references edges survive an unparseable routine in the same synthetic DDL.

PR #1836, opened 2026-07-12
feat(gdscript): add .gd extraction support (#697)

TPAteeq:
Closes #697

  ## Summary

  - Add a dedicated tree-sitter GDScript extractor for `.gd` files.
  - Wire GDScript into detection, dispatch, extractor registry, language-family guards, and analysis.
  - Add the optional `graphifyy[gdscript]` dependency using a tree-sitter-compatible language-pack release.
  - Document GDScript support and installation.

  ## Extraction coverage

  - Named and implicit script classes
  - Functions, members, constants, signals, enums, and inner classes
  - Identifier and `res://` inheritance
  - Typed member references
  - `Type.new()` instantiations
  - Same-script calls
  - `preload()` and `load()` dependencies
  - Cross-file type resolution

  Godot built-ins are filtered to avoid noisy graph hubs, ambiguous receiver calls are not guessed, and unresolved type stubs use a reserved namespace to prevent
  collisions with file nodes.

  ## Validation

  - `uv run pytest -q` — 3,119 passed, 30 skipped
  - Ruff passed
  - Pyright passed
  - Package build passed
  - `graphify update .` completed successfully

PR #1929, opened 2026-07-15
feat(extractors): add GDScript (.gd) + Godot scene/resource (.tscn/.t…

ozdemirsarman:
# Add GDScript + Godot scene/resource extraction
 
Adds first-class support for **Godot Engine** projects: GDScript (`.gd`) AST
extraction plus a grammar-free extractor for Godot's scene/resource/project
text formats (`.tscn`, `.tres`, `project.godot`). This closes the gap where a
Godot codebase currently contributes almost nothing to the graph.
 
## What's new
 
**GDScript (`.gd`) — `graphify/extractors/gdscript.py`** (tree-sitter-gdscript)
 
| GDScript | Graph |
|---|---|
| `class_name Foo` / inner `class Foo` | class node + `defines` |
| `extends Bar` / `extends "res://x.gd"` | `extends` edge (res:// resolved) |
| `func f(): …` | function node + `defines` |
| `foo()` / `obj.method()` in a body | `calls` edges (local funcs resolved) |
| `Autoload.method()` | `calls` edge resolved to the function in the autoload's script (autoload map read from `project.godot`) |
| `signal s(args)` | signal node + `declares` |
| `emit_signal("s")` / `s.emit()` | `emits` edge |
| `s.connect(handler)` | `connects` edge (handler resolved to local func) |
| `preload("res://y.gd")` / `load(...)` | `imports` edge (res:// resolved) |
 
**Godot scenes/resources/project — `graphify/extractors/godot_scene.py`** (no grammar; line parser)
 
| Godot | Graph |
|---|---|
| `[ext_resource type="Script" path="res://x.gd"]` | `attaches_script` edge |
| `[ext_resource type="PackedScene" path="res://y.tscn"]` | `instances` edge |
| `script = ExtResource("id")` on a node | node-level `attaches_script` |
| `[connection signal="s" from="A" to="." method="m"]` | `connects` edge, **resolved to the target script's function node** |
| `project.godot [autoload]` | global singleton node + `autoload` + `script` edges |
| `project.godot run/main_scene` | `main_scene` edge |
 
The scene `[connection]` → function resolution is the payoff Godot devs can't
get from GDScript-only tooling: signal wires defined in `.tscn` files link
straight to the handler `func` in the `.gd` file, because both sides mint the
same `_make_id(script_stem, method)` node id.
 
## Wiring (mirrors every other language)
 
* `graphify/extractors/gdscript.py`, `graphify/extractors/godot_scene.py` (new)
* `detect.py`: `.gd .tscn .tres .godot` added to `CODE_EXTENSIONS`
* `extract.py`: facade re-exports + `_DISPATCH` entries
* `extractors/__init__.py`: `LANGUAGE_EXTRACTORS` registry entries
* `pyproject.toml`: optional `godot = ["tree-sitter-gdscript"]` extra
* `README.md`: language table updated
## Graceful degradation
 
Both extractors are safe without the grammar. `.gd` returns a bare file node
when tree-sitter-gdscript is absent (same pattern as `sql`/`pascal`), and the
scene extractor needs no grammar at all — so scene/autoload/signal-wire edges
always work.
 
## Tests
 
`tests/test_gdscript.py` (skipped when the grammar is absent) and
`tests/test_godot_scene.py` (always runs) cover every relation above, plus the
registry-identity sweep in `test_extractors_registry.py` auto-covers the two new
registry entries. All fixtures are synthetic; no third-party project code.
 
## Open question for maintainers — grammar packaging
 
`tree-sitter-gdscript` (PrestonKnopp, MIT) is **not yet published to PyPI** as a
standalone wheel, though its repo ships working Python bindings and builds from
source cleanly. The `.gd` loader therefore tries `tree_sitter_gdscript` first
and falls back to a grammar from `tree-sitter-language-pack`. I'm happy to (a)
publish `tree-sitter-gdscript` to PyPI with proper attribution, or (b) switch
the `godot` extra to whatever dependency you prefer — let me know. The `.tscn`
half needs no grammar regardless.


luevano:
I think you can use [`tree-sitter-language-pack>=1.12.5,<1.13`](https://github.com/xberg-io/tree-sitter-language-pack) like #1836 does. He just didn't include `.tscn` or `.tres` extractors. This language pack contains [PrestonKnopp/tree-sitter-gdscript](https://github.com/PrestonKnopp/tree-sitter-gdscript) and [PrestonKnopp/tree-sitter-godot-resource](https://github.com/PrestonKnopp/tree-sitter-godot-resource), could you update your PR to be based on this?

ozdemirsarman:
Done — switched the godot extra to tree-sitter-language-pack>=1.12.5,<1.13 (matching #1836). Both grammars now come from the pack: .gd uses gdscript, and I also moved the .tscn/.tres/project.godot scene extractor onto the bundled godot_resource grammar (with a dependency-free line-parser fallback when the extra isn't installed). The pack only needs tree-sitter>=0.23, so it stays within the repo's <0.26 pin. Tests pass.

luevano:
Thanks for the quick update! seems to be usable as is now.

Looks like you have claude working on this, if that's the case: any thoughts on implementing the other file changes as with #1836?

* `graphify/extractors/MIGRATION.md`
* `graphify/analyze.py`
* `graphify/build.py`

I'd also consider the following nits (coming from #1835 as well):

* keep the comment on `pyproject.toml` out of it and here in the PR, or a smaller 1 liner.
* add fixtures for the tests instead of having them in-line.
* `README.md` could also have similar treatment (modifications around line 260 and 331), but one for gdscript and one for godot resource.

The naming of the "godot scene" I think should be changed to "godot resource", meaning a bit of changes in the code itself to reference this.

ozdemirsarman:
Thanks for the review! Addressed everything in `de0f6b2`:

- **`analyze.py` / `build.py` / `extractors/MIGRATION.md`** — registered `.gd` → `gdscript` and `.tscn`/`.tres`/`project.godot` → `godot_resource` in all three, matching #1836.
- **Renamed to `godot_resource`** — module `godot_scene.py` → `godot_resource.py`, `extract_godot_scene` → `extract_godot_resource`, the registry key, and all wiring (`extract.py`, `extractors/__init__.py`). No stale `godot_scene` references remain.
- **`pyproject.toml`** — trimmed the comment to a single line.
- **Test fixtures** — moved the inline `.gd`/`.tscn`/`project.godot` bodies into `tests/fixtures/godot/{gdscript,resource}/`, and renamed the test file to `test_godot_resource.py`.
- **README** — added a `godot` extras row (~L260), and split the Godot file types out of the big "Code" row into their own **GDScript** and **Godot resources** rows (~L331), same treatment as the Apex / Terraform rows.

Full suite green locally. Let me know if you'd like the `godot_resource` grammar wired into anything else.

Issue #1894, opened 2026-07-14
extract: semantic cache key ignores extraction mode, and extract has no --force, so --mode deep over a warm cache is a silent no-op


Nyxx0x:
**Version:** 0.9.15 (uv tool install of `graphifyy`, macOS, Python 3.14)

The semantic cache key is `sha256(file_content + "\x00" + relative_path)` (`cache.py`, `file_hash`, ~lines 162–210). It contains no extraction mode, model, or reasoning effort. `check_semantic_cache()` is called unconditionally from the extract path, and `graphify extract` on 0.9.15 exposes no `--force` and no `--no-cache` flag. (A `--force` exists on `update`; #1523 suggests one existed on `extract` at some point, but 0.9.15's extract parser does not accept it.)

### Repro

1. `graphify extract .` at default mode, to completion — cache is now warm.
2. `graphify extract . --mode deep`

### Actual

The log prints `deep mode enabled: richer semantic extraction`, and then every unchanged file hits the cache. The resulting graph is equivalent to the shallow run, wearing a "deep" label. On a 119-file corpus, zero files were re-extracted — the mode flag applied to nothing.

This is quiet: the run exits 0 and the "deep mode enabled" line reads like confirmation. The only signal that anything is wrong is the absence of a `semantic extraction on N files` line with a real N.

### Expected

Either the extraction mode (and ideally the model / reasoning effort) participates in the cache key, or `extract` grows a `--force` / `--no-cache` escape hatch.

Today the only workaround is `rm -rf graphify-out/cache/semantic/`, which is undocumented and easy to get wrong — users also need to know to *keep* `cache/ast/` (deterministic tree-sitter output, unaffected by mode) and `cache/stat-index.json`.

Related: #1252 raised the same content-only-key staleness for the AST cache across upgrades. This is the semantic-cache / extraction-mode variant of it.

safishamsi:
[...]

Issue #1252, opened 2026-06-11
AST cache survives graphify upgrades: content-only keys keep serving stale extractor output


teemow:
## Problem

The extraction cache (`graphify/cache.py`) keys entries on `SHA256(file content + relative path)` only. For the **AST** cache this is incomplete: the cached value is the output of graphify's own extractor code, so it is only valid for the version that produced it. After upgrading graphify, any extractor fix (new node kinds, corrected edge resolution, parser bug fixes) is silently masked for every unchanged file — `load_cached` keeps returning the pre-fix result because neither the file content nor its path changed.

The only escape hatches today are `graphify extract --force` or manually deleting `graphify-out/cache/`, and nothing tells the user they need to do that after an upgrade.

## Reproduction

```python
from pathlib import Path
from graphify.cache import save_cached, load_cached

root = Path("repo")
f = root / "mod.py"

# graphify 0.8.0 extracts and caches
save_cached(f, {"nodes": [{"id": "old-extractor-output"}], "edges": []}, root=root)

# ...pip install -U graphifyy (0.8.1 ships an extractor fix)...

# 0.8.1 still serves the 0.8.0 result for the unchanged file
print(load_cached(f, root=root))  # {'nodes': [{'id': 'old-extractor-output'}], ...}
```

## Proposed solution

Namespace the **AST** cache directory by package version — `cache/ast/v{version}/` — and remove entries left behind by other versions (sibling `v*/` dirs and unversioned `cache/ast/*.json`) on first use, so the cache does not grow one full copy per release. The legacy flat-`cache/` fallback in `load_cached` (pre-0.5.3 migration path) should also stop being consulted: those entries are by definition from an older extractor.

The **semantic** cache must stay untouched: its entries are produced by the LLM from file contents, and invalidating them on every patch release would re-bill extraction for unchanged files.

## Acceptance criteria

- An AST entry written under version X is not returned after a version bump, and stale version directories are cleaned up.
- Semantic cache entries survive version bumps and the AST cleanup.
- Regression tests in `tests/test_cache.py`; existing suite stays green.

PR #1942, opened 2026-07-16
fix(cache): key semantic cache on the extraction prompt (#1939)

### LINKED ISSUE #1939: Semantic cache replays stale extractions after upgrades: key has no prompt/skill-version component
**Version:** 0.9.16 (uv tool install of `graphifyy`, macOS)

While setting graphify up for a team repo I noticed our graph was mixing extraction vintages: nodes shaped by an older release's extraction prompt sitting next to nodes from the current one, in the same graph.json, with nothing flagging it.

The mechanism: the semantic cache keys entries on `sha256(file content + "\x00" + relative path)` and nothing else (`file_hash`, `cache.py` ~lines 162-210 in 0.9.16). There is no component for the version of the extraction prompt that produced the entry. After an upgrade, every unchanged file is a cache hit and replays whatever the older prompt extracted. The run exits 0, cost.json looks cheap, and the graph quietly carries two prompt generations.

What we saw concretely: `graphify-out/cache/semantic/` held 880 entries, 356 written by an older install and 524 by 0.9.16. A rebuild we expected to be cold re-extracted only 6 of 512 doc files; the other 506 replayed the old vintage. The only way we found to get a single-vintage graph was deleting the cache.

I know the unversioned semantic cache is deliberate: #1252 versioned the AST cache and explicitly left semantic alone so a patch release doesn't re-bill LLM extraction, and that reasoning is sound. The gap is that the extraction prompt itself changes across releases (we hit this through the bundled Claude skill's prompt), and when it does, the cache has no way to notice. #1894 added mode namespacing for 0.9.17, but there is still no prompt/version dimension in the key.

A fingerprint of the extraction prompt in the key, or a prompt-hash-namespaced directory like the AST cache's `v{version}` layout, seems like it would keep both properties: entries survive releases that don't touch the prompt, and invalidate only when it actually changed. Even a warning that N cache entries predate the installed version would help a lot, since today there is no signal at all.

Our workaround is wiping `graphify-out/` entirely before rebuilding after an upgrade. It works, but it forfeits the whole cache (AST included) and is easy to forget.

Thanks for the tool, and for how quickly #1894 moved.

SinghAman21:
# fix(cache): key the semantic cache on the extraction prompt

Fixes #1939

## What was the issue

The semantic cache keyed entries on `sha256(file content + "\x00" + relative path)` and nothing else. Nothing in the key described *which extraction prompt produced the entry*.

So when a release changed the prompt, every unchanged file was still a cache hit and replayed whatever the older prompt had extracted. Nothing flagged it: the run exited 0, `cost.json` looked cheap, and the resulting `graph.json` quietly carried two prompt generations side by side — nodes shaped by an old prompt sitting next to nodes from the current one.

The reporter hit this through the bundled skill's prompt: their `graphify-out/cache/semantic/` held 880 entries across two vintages, and a rebuild they expected to be cold re-extracted only 6 of 512 docs — the other 506 replayed the old prompt. Deleting the entire cache was the only workaround they found, which forfeits the AST cache too.

This was a gap, not an oversight. #1252 versioned the AST cache and deliberately left the semantic cache unversioned: its entries are LLM output, and invalidating them on every release would re-bill extraction for unchanged files. That reasoning is sound, and any fix has to preserve it. #1894 later added mode namespacing (`semantic` vs `semantic-deep`), but still no prompt dimension.

## Approach

Fingerprint the extraction prompt and namespace entries by it, rather than by package version.

This keeps both properties the issue asks for: entries **survive releases that don't touch the prompt** (no re-bill on a patch release — #1252's constraint holds), and **invalidate only when the prompt actually changed**.

Entries live under `cache/semantic/p{fingerprint}/`, mirroring the AST cache's `v{version}/` layout, so the two caches now read as one idea with two different validity keys — version for code-produced entries, prompt for LLM-produced ones.

Two decisions worth reviewer attention:

**Pre-existing entries are served, not dropped.** They predate fingerprinting, so their vintage is unknowable. Ignoring them would be strictly correct but would re-bill a whole corpus on upgrade (506 docs for the reporter). Instead they still hit, and `check_semantic_cache` warns with the count — turning the "no signal at all" the issue describes into a visible one. `--force` re-extracts them. The honest limitation: such an entry keeps serving its unknown vintage until its content changes or someone forces it. The transition is gradual, not immediate.

**Old-fingerprint entries are pruned by liveness only**, never swept wholesale the way `_cleanup_stale_ast_entries` sweeps old AST versions. Two hosts with different prompts (the verbose vs compact `extraction-spec.md`) can share one `graphify-out/`; a wholesale sweep would have each run delete the other's entries and re-bill on every alternation. Liveness keeps the total bounded by live docs × prompts seen.

## Changes

**`graphify/cache.py`** — the core.
- `prompt_fingerprint(prompt)` accepts prompt text or a `Path` to a prompt file. Normalizes line endings and trailing whitespace before hashing, so a CRLF checkout isn't mistaken for a prompt change (which would re-bill every Windows run).
- `cache_dir(..., prompt_fp=)` adds the `p{fingerprint}/` subdir for semantic kinds. `prompt_fp=None` yields the historical flat layout, so existing callers are byte-identical.
- `load_cached` / `save_cached` / `check_semantic_cache` / `save_semantic_cache` take `prompt=` (text) and `prompt_file=` (path). These are separate parameters on purpose: the skill callers are markdown snippets an agent copies with a path substituted in, and passing that path as `prompt` would hash the path string — a fingerprint that is stable, plausible, and tracks nothing. A silent wrong fingerprint is the exact failure class being fixed.
- `merge_existing` reads with `allow_legacy=False`: fusing a pre-fingerprint entry into a current-vintage write would mix two prompts inside one entry and then attest the result to a prompt that produced half of it.
- `prune_semantic_cache` / `clear_cache` / `cached_files` glob recursively, so fingerprinted entries can't become unprunable orphans (the #1527 failure mode).
- Prompt-file fingerprints are memoized on a `(path, size, mtime_ns)` signature — `check_semantic_cache` resolves the prompt once per file, so a 500-doc run would otherwise re-read and re-hash the same spec 500 times.

**`graphify/llm.py`** — the per-chunk checkpoint stamps entries with `_extraction_system(deep=deep_mode)`, the prompt every backend actually sends.

**`graphify/cli.py`** — `extract` passes the same prompt to both the cache read and the write (a mismatch would strand writes where the next read won't look). `cache-check` gains `--prompt-file`.

**`tools/skillgen/fragments/core/core.md`** + 14 rendered skills — Step B0/B3 pass `prompt_file='SPEC_PATH'`, the `references/extraction-spec.md` the subagents are handed, with instructions for substituting it. Regenerated via `python -m tools.skillgen`; `--check` is clean. The two monolith skills (aider, devin) inline their prompt rather than shipping a spec sidecar and stay on the unfingerprinted path — called out as a follow-up.

**Tests** — 12 new, covering: prompt change invalidates; unchanged prompt still hits; CRLF insensitivity; prompt × mode compose independently; legacy entries served with the count warning; fingerprinted entry beats legacy; `merge_existing` won't fuse vintages; prune/clear reach the subdirs; unreadable `prompt_file` warns and falls back; an edited spec defeats the memo; `cache-check --prompt-file`; and a skillgen guard asserting every split host passes `prompt_file` to **both** calls (they live ~80 lines apart in the rendered body, so adding it to one and not the other would silently disable the cache rather than fail loudly).

A few existing tests asserted the flat layout (`glob("*.json")`, unprompted `load_cached`) and were updated to the nested layout — behavior change, not a weakened assertion.

## Why

The alternative fixes each break something the issue explicitly wants kept:

- **Version the semantic cache like the AST cache** — re-bills LLM extraction on every patch release. This is exactly what #1252 rejected, and the issue agrees that reasoning is sound.
- **Warn only** — the issue offers this as a fallback ("even a warning ... would help a lot"), but it leaves the graph mixed. Warning is the right treatment for entries we *can't* attribute; it's not a fix for the ones we can.
- **Wipe `graphify-out/` on upgrade** — the reporter's current workaround. Forfeits the AST cache, and is easy to forget.

Fingerprinting is the only option that invalidates precisely when correctness demands and never otherwise.

## Verification

- Full suite: **3277 passed**. The only failures are 4 pre-existing ones in `tests/test_ollama_retry_cap.py`, confirmed failing identically on a clean tree (`openai` optional extra absent from a stock dev venv — unrelated, documented separately).
- `python -m tools.skillgen --check`: clean, 134 artifacts match.
- End-to-end through the real CLI, reproducing the reporter's scenario:
  - unchanged prompt → `Cache: 1 hit, 0 miss` (no re-bill)
  - prompt changed by an "upgrade" → `Cache: 0 hit, 1 miss` (re-extract, no stale replay)
  - pre-fingerprint cache → `Cache: 3 hit, 0 miss` plus `3 semantic cache entries predate extraction-prompt fingerprinting...`, confirmed visible under default warning filters
  - layout on disk: `cache/semantic/pa55707ed205d/{hash}.json`

## Conclusion

An upgrade that changes the extraction prompt now re-extracts affected files instead of silently mixing vintages, and an upgrade that doesn't touch the prompt still costs nothing — the property #1252 protected. Users upgrading with an existing cache keep their entries and get a count of how many are of unknown vintage, so the silent case the issue reports is now either correct or visible.

**Known follow-ups, deliberately not in this PR:**
- The monolith skills (aider, devin) inline their prompt and stay unfingerprinted.
- Separately, no bundled skill passes `mode` either, so a skill-driven `--mode deep` still writes into the plain namespace — the gap #1894's changelog already anticipates as skill-side follow-up.
- A stale installed SKILL.md (older than the package) will fingerprint against the spec it actually loaded, which is correct; the existing `.graphify_version` stale-skill warning remains the signal for that.


safishamsi:
Landed on v8 in 0d018b4 (ships in 0.9.18), cherry-picked so your authorship is preserved in the history. Thanks @SinghAman21 — this keys the semantic cache on a fingerprint of the extraction prompt, so an upgrade that changes the prompt invalidates only the entries it should, while releases that leave the prompt untouched keep their cache. Verified end to end: same prompt hits, a changed prompt re-extracts instead of replaying stale output, and the fingerprint is CRLF-insensitive. Credited in the changelog alongside @HunterMcGrew for the report.

Issue #1899, opened 2026-07-14
Absolute paths still land in node id + source_file on 0.9.15: out-of-root reference targets and degenerate-symbol id fallback (residual of #1789)


tparnell:
## Summary

Follow-up to #1789 (fixed in 0.9.13 for `.sln`/`.csproj` *file* node ids, and in 0.9.14 for virtual solution-folder ids — thanks for both). On **0.9.15** two code paths still emit machine-specific absolute paths into `graph.json`:

- **Variant A — out-of-root reference targets.** A node created for a file *outside the scan root* (e.g. the target of a `<ProjectReference Include="..\Core\Core.csproj" />` when only `WebApi/` is scanned) gets an **absolute `source_file` AND an absolute-path-derived `id`**. The relativization added for #1789 appears to rely on `relative_to(root)`, which can't apply to out-of-root paths, so both fields fall back to absolute.
- **Variant B — degenerate symbol nodes from in-root files.** Some symbol nodes get a correctly **relativized `source_file` but an absolute-path-derived `id`**. Two triggers observed: a JSON file using the tsconfig `"//"` comment-key convention (node label `//`), and a minified vendor bundle (node label `$()` from `moment.min.js`). Symbol-id derivation seems to fall back to `_make_id(str(absolute_path))` when no usable symbol/module name is available.

Consequences are the same as #1789: the committed `graph.json` leaks the local username / checkout path, ids differ per machine and per checkout location, so `graphify update` from another clone re-creates these nodes (diff churn, non-idempotent versioned graphs).

## Environment

- graphify (`graphifyy`) **0.9.15** (installed via `uv tool`), Python 3.12.9
- macOS 26.5 (also expected on Linux/Windows — the fallback logic is OS-agnostic)
- Plain AST build (`_origin: "ast"` on all affected nodes) — no LLM/deep mode involved, so this is distinct from #1895.

## Minimal repro (4 files, no .NET SDK, no API key)

```bash
#!/usr/bin/env bash
set -euo pipefail
R="$(mktemp -d)"
mkdir -p "$R/Core" "$R/WebApi"

cat > "$R/Core/Core.csproj" <<'EOF'
<Project Sdk="Microsoft.NET.Sdk">
  <PropertyGroup><TargetFramework>net8.0</TargetFramework></PropertyGroup>
</Project>
EOF

cat > "$R/WebApi/WebApi.csproj" <<'EOF'
<Project Sdk="Microsoft.NET.Sdk">
[...]

PR #1947, opened 2026-07-16
fix(extract): anchor source_file on the scan root, not the --out dir

### LINKED ISSUE #1941: `source_file` reduced to bare basename for all nodes in 0.9.16 (was relative path in 0.9.13) — breaks resolution against a code root
### Summary

Between `graphifyy` **0.9.13** and **0.9.16**, the `source_file` field on nodes changed from a **path relative to the scan root** to a **bare filename** (basename only) — for *every* node, not just the cases mentioned in the release notes. This removes the only field that carried the file's directory, so a `graph.json` can no longer be resolved back to files on disk by joining `source_file` onto the scan/code root.

I'm not sure whether this is intended. The 0.9.14/0.9.16 notes describe making paths **portable/relative** for narrow cases (VS solution-folder nodes → "now relative"; out-of-root reference targets and degenerate symbols → "now portable"), but a *relative* path is already free of absolute-path/username leaks, so the wholesale reduction to basename seems to go beyond what's documented. Hence this report/question.

### Environment

- `graphifyy` 0.9.13 (baseline) vs 0.9.16 (observed change)
- Windows 11, Python 3.14
- Language: C# (nodes are `_origin: ast`, so this is AST-derived and independent of the semantic backend)

### Steps to reproduce

1. Extract any C# tree that has files in nested subdirectories:
   `graphify extract <ROOT> --out <OUT>`
2. Inspect `source_file` on the nodes in the resulting `graph.json`.
3. Compare the same extraction under 0.9.13 and 0.9.16.

### Expected

`source_file` is a path relative to the scan root (as in 0.9.13), e.g. keeping the subdirectory so the file is locatable via `<root>/<source_file>`.

### Actual

`source_file` is the bare filename only; the directory component is gone.

**Same file, both versions (example):**

| version | `source_file` |
|---|---|
| 0.9.13 | `src/Data/Database/RepositoryTests/OrderRepositoryTests.cs` |
| 0.9.16 | `OrderRepositoryTests.cs` |

**Aggregate over one representative graph (~120k `source_file` occurrences):**

| version | with a path separator (`/`) | basename only |
|---|---|---|
| 0.9.13 | 117,897 | 7,335 |
| 0.9.16 | **0** | 119,341 |

**No other field carries the path in 0.9.16.** Node keys are `[id, label, source_file, source_location, file_type, community, norm_label, type, metadata]`; `source_location` is a line only (e.g. `"L14"`), and a scan of *every* string value on *every* node finds 0 nodes containing a directory separator in any `.cs`-bearing value.

The full directory does survive, but only **slugified and lossy**, inside the type node `id`, e.g.:

```
id = c_..._src_data_database_repositorytests_orderrepositorytests_...
```

i.e. lowercased with `.`, `/`, `\`, `-`, spaces all collapsed to `_` (and truncated). Because those distinct characters all map to `_`, this can't be reversed to a filesystem path reliably.

### Impact

Any downstream tooling that resolves nodes back to source by joining `source_file` onto the scan/code root breaks: 0% of files resolve after the change. The lossy `id` slug is not a viable substitute.

### Questions

1. Is the basename-only `source_file` for **in-root** files intended, or an over-broad side effect of the out-of-root / degenerate-symbol leak fixes?
2. If intended: is there (or could there be) a field carrying the **portable relative path** (as the 0.9.14 "now relative" direction suggests), so a `graph.json` remains resolvable against its root without parsing the `id` slug?

SinghAman21:
Closes #1941.

@JensD-git's report is accurate, and the answer to their Q1 is "over-broad side effect" — this isn't intended behavior.

What happens

--out passes the output dir into extract() as cache_root, purely to relocate the cache (we don't want a graphify-out/ appearing inside a project that asked for external output). But cache_root also anchored path relativization. So with --out, root becomes the output dir, every scanned file fails relative_to(root), and the whole corpus falls into the out-of-root fallback added in #1899 — which then trips its updepth > 3 walk-up guard and collapses the path to a basename.

That guard was written for a stray ProjectReference pointing outside the corpus. Nothing scoped it to "this is actually an external reference," so once the anchor was wrong it fired on everything. On the reporter's Windows setup an --out on a different drive hits the cross-drive branch instead and basenames unconditionally — which is exactly why they saw 0 of ~120k source_files keep a separator rather than some partial mix.

The damage is total because nothing else carries the path: origin_file is stripped before serialization (#1516), the export has no file table or scan root, and the id slug is lossy (., /, \, -, spaces all collapse to _). So 0% of nodes resolve, and their conclusion that the slug isn't a viable substitute is right.

The fix

extract() takes an explicit root anchor for source_file/ids/symbol resolution, and the CLI pins it to the scan root regardless of where the cache lives.

Two things convinced me this was the shape rather than tuning the updepth heuristic:

- build() was already being called with root=target (cli.py:2853, 2856). The build layer always expected scan-root-relative paths — extract was the only component keying off --out. This makes the two agree instead of papering over the disagreement.
- #1774 already established that cache location and the key/id anchor are separate concerns and decoupled them for the cache dir. The root = cache_root line was the leftover half of that job.

cache_root keeps its fallback-anchor role, so anything passing the scan root as cache_root (watch, the no---out path, ~50 tests where tmp_path is both) behaves identically. Cache location and #1899's out-of-root portability are untouched.

Verification

Reproduced at the API level first — same file, same call, only the anchor differing:

OLD (cache_root anchors): ['order_repository_tests.py']
NEW (explicit root):      ['src/Data/Database/RepositoryTests/order_repository_tests.py']

Then end-to-end through the real CLI with a far-away --out: both files keep their directories, both resolve against the scan root, and no absolute paths or usernames leak — so #1899 doesn't regress. Regression test covers the exact updepth > 3 layout that triggered it, and asserts the point of the field (that root / source_file is a real file), not just the string.

Full suite: 3266 passed. The 4 failures in test_ollama_retry_cap.py are pre-existing on a clean v8 — unrelated to this change.

Existing broken graphs self-heal on the next extract.

Not in scope

Below the walk-up threshold, _portable_out_of_root_sf still returns things like ../../Users/<name>/proj/src/Foo.cs — which embeds the username #1899 set out to remove. Genuinely out-of-root, so it's a separate bug from this one; happy to file it if we want it tracked.


safishamsi:
Landed on v8 (ships in 0.9.18), cherry-picked to preserve your authorship. Thanks @SinghAman21 — extract() now takes an explicit root anchor pinned to the scan root, so --out no longer collapses every source_file to a bare basename. Verified a repo-relative source_file with --out outside the project. One note adjusted in the changelog: a graph already written with basenamed paths does not self-heal on an unchanged corpus (the stale nodes prune out) — it needs one graphify extract --force, or a re-extract after a file change.

PR #1949, opened 2026-07-16
fix(llm): downgrade unverifiable code-typed semantic nodes to UNVERIFIED

TPAteeq:
### Problem
The semantic (LLM) extraction pass only ever runs on documents, papers, and images — code files are handled by the deterministic AST engine and are never sent to the model (`cli.py` builds the semantic worklist as `semantic_files = doc_files + paper_files + image_files`). So when the model returns a node tagged `file_type="code"`, that symbol was surfaced from *inside* a document — a name in a fenced code block, an API referenced in a paper, a symbol mentioned in prose. Today such a node enters `graph.json` with **no check that the symbol actually occurs in the source the model read**. The existing `_out_of_scope` filter (#1895) only rejects a node whose `source_file` resolves to a real file that was *not* dispatched in the batch; a fabricated symbol attributed to a file that *was* dispatched sails straight through it. That directly undercuts the "grounded in the source" guarantee: a plausible-looking but invented symbol becomes a permanent, first-class graph node.

### Fix
`extract_files_direct` now performs an evidence check on its own output before returning. For every `file_type=="code"` node whose `source_file` resolves to one of the files dispatched in *this* call, it verifies that at least one identifier token from the node's label (≥3 chars, case-insensitive) appears as a substring of that file's source bytes — the exact bytes the model saw, read through the same `_read_files` path and capped at `_FILE_CHAR_CAP`. If none of the label's identifiers occur, the node's `confidence` is downgraded to `"UNVERIFIED"` rather than dropped, and the number downgraded is reported to stderr so the signal is visible during extraction (and left on the node in `graph.json` for programmatic inspection).

The check is precision-first and deliberately conservative: concept/document nodes, nodes with no `source_file`, nodes attributed to a file not dispatched in this call (left to `_out_of_scope`), and labels with no checkable identifier are all untouched. The whole pass is best-effort — a single unreadable file (e.g. a malformed PDF) is skipped rather than aborting the chunk, and any unexpected error disables the check without failing extraction.

### Tests
`tests/test_evidence_binding.py`: fabricated symbol downgraded; real and qualified/prettified labels (`PaymentProcessor.charge_card()`) not false-flagged; concept, document, and no-source nodes untouched; symbols attributed to an undispatched file left to `_out_of_scope`; short/uncheckable labels skipped; existing lower confidence never clobbered; `FileSlice` input; absolute `source_file`; stderr surfacing; and that a downgraded node still passes validation. Full test suite green (3295 passed).

### DIFF
--- a/graphify/llm.py
+++ b/graphify/llm.py
@@ -555,6 +555,112 @@ def _read_files(units: "list[Path | FileSlice]", root: Path) -> str:
     return "\n\n".join(parts)
 
 
+# ── Semantic evidence-binding ─────────────────────────────────────────────────
+# The semantic (LLM) extraction runs on documents/papers/images — code files are
+# handled by the deterministic AST engine and never reach the model. So a
+# ``file_type == "code"`` node here is a symbol the model surfaced from WITHIN a
+# document (a name in a fenced code block, an API referenced in a paper). Verify
+# that such a symbol actually occurs in the source bytes the model was shown; a
+# node the model asserts with no evidence in its source is a likely fabrication.
+# `_out_of_scope` (#1895) only rejects a node attributed to a real file that was
+# NOT dispatched; a fabricated symbol attributed to a file that WAS dispatched
+# slips through it. This closes that intra-file gap with a lenient substring
+# check and DOWNGRADES (never drops) an unverifiable node's confidence to
+# UNVERIFIED, surfaced by the caller (stderr) and left on the node in graph.json.
+# Short tokens (len < 3) are ignored: they match too readily to be evidence and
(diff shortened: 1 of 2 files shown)


safishamsi:
Landed on v8 (ships in 0.9.18), cherry-picked to preserve your authorship. Thanks @TPAteeq — code-typed semantic nodes with no evidence in the source are now flagged rather than presented as read facts. Follow-up added: moved the flag off the confidence key (whose validated vocabulary is edge-only) onto a dedicated verification="unverified" node field, made it check the node id as well as the label (fewer false flags on prettified labels), skipped the source re-read when a result has no code node, and wired a consumer so it is not a dead field (graphify diagnose reports the count).

Issue #933, opened 2026-05-18
Files in failed semantic chunks are permanently skipped on re-run


sub4biz:
### Description

When a semantic extraction chunk fails (e.g. 429 rate limit), graphify marks
the files in that chunk as processed in `manifest.json` anyway. On the next run,
incremental mode reports "0 files changed" and skips them permanently — those
files never make it into the graph.

### Steps to reproduce

1. Run `graphify extract` with Gemini free tier (TPM=250K/min):
   `graphify extract . --backend gemini --token-budget 100000 --max-concurrency 1`
2. Chunks 1 and 2 succeed (~100K tokens each, 200K total within the minute window)
3. Chunk 3 hits 429 — cumulative 300K tokens exceeds TPM=250K limit
4. Re-run the same command
5. Output: incremental mode, 0 files to re-extract
6. Files from chunk 3 onwards are absent from the graph — silently

### Expected behavior

Files whose chunk failed should remain `pending` in manifest.
Re-running should build new chunks only from pending files.

### Root cause

`manifest.json` records file state (mtime/hash) per file, not chunk-level
success/failure. There is no way to distinguish "file was successfully
semantically extracted" from "file was grouped into a chunk that later failed".

### Environment

- graphify 0.8.8
- Backend: `gemini` (gemini-3-flash), `--max-concurrency 1`, `--token-budget 100000`
- Some chunks succeeded, others failed with 429 (TPM=250K/min limit exceeded)

### Suggested fix

Write chunk results to manifest only after the chunk completes successfully.
Files whose chunk failed keep their previous manifest state and are included
in the next run's chunk planning.

safishamsi:
Fixed in 0.8.13 — `save_manifest` now only stamps `semantic_hash` for files that appear in extraction output. Files whose chunk failed keep an empty hash and are re-queued on the next run.

PR #1950, opened 2026-07-16
fix(cache): don't promote truncated LLM chunks to the semantic cache as complete

TPAteeq:
### Problem
When an LLM chunk is truncated (`finish_reason="length"`) and can't be recovered by splitting — or the adaptive-retry recursion hits its depth cap — the extractor keeps the partial node set it managed to parse. That partial result was then checkpointed and written to the content-hash semantic cache and manifest-stamped as a *complete* extraction. Because the cache is keyed by file content, the incomplete node set is served on every subsequent run **forever** — until the file's content changes or the user passes `--force`. Half a document's symbols silently masquerade as the whole, with no persisted signal that anything was missing.

### Fix
A truncated give-up result is tagged with an internal `_partial` marker on each of its items. `save_semantic_cache` stamps the affected file's cache entry `partial: True` (detected either from that marker or from an explicit `partial_source_files` argument), and `load_cached` treats a partial entry as a cache **miss**, so the file is re-dispatched and retried instead of replayed. To make that re-dispatch actually happen on the common path, the partial file is also left **unstamped in the manifest** — mirroring the existing failed-chunk handling (#933) — so `detect_incremental` re-queues it on the next incremental `update`/`extract`, not only on a full/`--force`/content-change run.

A file split into slices across chunks is handled correctly: the `merge_existing` checkpoint peeks at a partial previous entry (`load_cached(allow_partial=True)`) so it accumulates all slices and keeps the entry partial, rather than a later clean slice silently dropping a truncated one and promoting the half-file to complete. The whole thing self-heals — a later complete extraction overwrites the same content-hash key with a non-partial entry — and the internal marker is stripped after the final save so it never leaks into `graph.json`.

### Tests
`tests/test_partial_cache.py` and additions to `tests/test_chunking.py`: partial entry is a load miss; explicit `partial_source_files`; self-heal on complete re-extraction; sliced-file accumulation stays partial without losing a slice; manifest excludes partial files; the marker helpers; and an end-to-end truncated `extract_corpus_parallel` checkpoint that reloads as a miss. Full test suite green (3297 passed).

### DIFF
--- a/graphify/cache.py
+++ b/graphify/cache.py
@@ -465,7 +465,8 @@ def cache_dir(root: Path = Path("."), kind: str = "ast",
 def load_cached(path: Path, root: Path = Path("."), kind: str = "ast",
                 cache_root: Path | None = None, prompt: "str | Path | None" = None,
                 prompt_file: "str | Path | None" = None,
-                allow_legacy: bool = True) -> dict | None:
+                allow_legacy: bool = True,
+                allow_partial: bool = False) -> dict | None:
     """Return cached extraction for this file if hash matches, else None.
 
     Cache key: SHA256 of file contents.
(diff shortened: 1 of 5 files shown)


safishamsi:
Landed on v8 (ships in 0.9.18), cherry-picked to preserve your authorship. Thanks @TPAteeq — truncated chunks are no longer cached/stamped as complete. Follow-up added for the empty-parse case: a mid-JSON cut parses to zero items, so the give-up sites now record the chunk own files independently of parsed items (propagated through the merges), the cache stamps a partial file even with no items, and a later clean slice cannot re-promote a partial entry. That closes the sliced-document scenario end to end.

PR #1951, opened 2026-07-16
fix(extract): don't force-write a partial graph over a complete one

TPAteeq:
### Problem
A full `graphify extract` writes the final graph with `to_json(..., force=True)`, which bypasses the `#479` shrink guard in its entirety (the guard is gated behind `if not force and existing_path.exists()`). Forcing is correct for a clean build that legitimately shrinks — fuzzy-dedup collapse, deleted code — but it means a build whose extraction was **incomplete** silently overwrites a good, larger graph with a smaller partial one. That happens when an AST extractor pass crashes (the `except` swallows it and continues) or when some, but not all, semantic chunks fail (only the *all*-failed case aborted before). A docs-heavy corpus with a few rate-limited chunks is exactly the case where a complete graph gets clobbered by a partial rebuild with no warning.

### Fix
The extract path now tracks a `_extraction_incomplete` flag, set when the AST pass raises, when the semantic pass raises, or when `succeeded < total` chunks completed. At the final write it computes `force = cli_allow_partial or not _extraction_incomplete`: a clean build still forces (legit shrinks proceed), but an incomplete build falls back to the shrink guard, so a smaller partial graph is **refused** rather than written. On refusal the command prints a clear error and exits non-zero *before* the manifest, marker, and analysis sidecars are written — so the manifest is never stamped for a graph we declined to write, and a later retry re-attempts cleanly. A new `--allow-partial` flag restores `force=True` for the cases where overwriting with a partial graph is intentional.

**Both write paths are guarded.** The clustered path gets this from `to_json`'s #479 check; the `--no-cluster` raw-dump path (which never calls `to_json`) gets the same check via a new `existing_graph_node_count` helper that reads the existing graph's node count under the same size cap and refuses an incomplete shrink before the raw write.

### Known limitation (documented in code)
The fallback reuses `to_json`'s coarse node-count guard rather than the source-aware `_check_shrink` that `watch`/`update` use. So on an incremental run, a legitimate deletion that coincides with an unrelated transient chunk failure can be refused — always recoverable by re-running or passing `--allow-partial` (the good graph is preserved and the manifest is not stamped).

### Tests
`tests/test_incomplete_build_guard.py`: clustered path — partial+shrink → refuse + exit 1 + no manifest, partial-but-not-shrinking → writes, `--allow-partial`, clean build keeps force; `--no-cluster` path — partial+shrink refused with the existing graph left intact, and `--allow-partial` overwrites; plus a unit test for `existing_graph_node_count`. Full test suite green (3290 passed).

### DIFF
--- a/graphify/cli.py
+++ b/graphify/cli.py
@@ -2762,7 +2762,10 @@ def dispatch_command(cmd: str) -> None:
             # across modes (#1317; node dedup also collapses shared Swift module
             # anchors emitted per importing file, #1327).
             from graphify.build import dedupe_edges as _dedupe_edges, dedupe_nodes as _dedupe_nodes
-            from graphify.export import backup_if_protected as _backup
+            from graphify.export import (
+                backup_if_protected as _backup,
+                existing_graph_node_count as _existing_graph_node_count,
+            )
             if (
                 incremental_mode
                 and not code_files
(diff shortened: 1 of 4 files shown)


safishamsi:
Landed on v8 (ships in 0.9.18), cherry-picked to preserve your authorship. Thanks @TPAteeq — an incomplete extraction (crashed pass, some-chunks-failed, or a walk that could not enumerate the corpus) no longer force-writes a partial graph over a complete one, on both the clustered and --no-cluster paths. Follow-up added: the --no-cluster guard now fails closed on a present-but-unparseable existing graph (a corrupt/mid-write file could be hiding a complete graph), and detect() walk_errors count as incomplete.

PR #1952, opened 2026-07-16
fix(io): write graph.json and manifest.json atomically

TPAteeq:
### Problem
`graph.json` (the clustered `to_json` write, the `--no-cluster` raw dump, and the merge driver) and `manifest.json` (`save_manifest`) were all written with a direct `open()`/`write_text`. A crash, `SIGKILL`/Ctrl-C, OOM, or a disk-full condition partway through the write leaves a **truncated, unparseable file** in place of the previous good one — and the very next `graphify` invocation (or `detect_incremental` reading the manifest) then fails to load it. For the graph and the manifest, which gate every subsequent run, a partial write is a self-inflicted outage.

### Fix
Add `write_text_atomic` / `write_json_atomic` to `graphify.paths` and route the four writers through them. The helpers write to a temp file in the **same directory** and then `os.replace` it into place — an atomic rename on one filesystem, so the destination is never observed half-written and a failed write leaves the previous file untouched (the temp is cleaned up). Details that make it safe in the real environments graphify runs in:

- JSON is **streamed** into the temp via `json.dump` rather than materialized as one giant string first (matters for the very large graphs graphify can produce).
- The destination's existing file mode is **preserved** so an atomic replace never silently tightens a `0644` output to `mkstemp`'s `0600` (new files get the umask default).
- A **symlinked destination** is resolved first so the write goes *through* the link to its target rather than replacing the link with a regular file (the shared-output / worktree setups this module documents keep working).
- A Windows `os.replace` `PermissionError` (destination briefly locked by antivirus or an open reader) falls back to copy-then-delete — matching the atomic writer already in `graphify.cache`.

This is not a power-loss durability guarantee (there is no `fsync`, consistent with the rest of the codebase); it protects concurrent readers and the process-kill / ENOSPC cases, not an OS/hardware crash.

### Tests
`tests/test_atomic_writes.py`: write leaves no temp behind; a failed `os.replace` preserves the original and cleans the temp; existing mode preserved and new-file mode follows umask (not `0600`); symlink write-through; streaming `write_json_atomic` round-trip; and `to_json` / `save_manifest` leave no `.tmp` leftover. Full test suite green (3291 passed).

### DIFF
--- a/graphify/cli.py
+++ b/graphify/cli.py
@@ -246,7 +246,8 @@ def _prune_graph_json_sources(graph_path: Path, stale_sources: list[str]) -> int
         data["hyperedges"] = kept_hyper
     from graphify.export import backup_if_protected as _backup
     _backup(graph_path.parent)
-    graph_path.write_text(json.dumps(data, indent=2), encoding="utf-8")
+    from graphify.paths import write_json_atomic
+    write_json_atomic(graph_path, data, indent=2)
     return n_removed
 
 
@@ -1615,7 +1616,8 @@ def dispatch_command(cmd: str) -> None:
             out_data = _jg.node_link_data(merged, edges="links")
         except TypeError:
             out_data = _jg.node_link_data(merged)
-        Path(_current_path).write_text(json.dumps(out_data, indent=2), encoding="utf-8")
+        from graphify.paths import write_json_atomic
+        write_json_atomic(_current_path, out_data, indent=2)
         sys.exit(0)
 
     elif cmd == "merge-graphs":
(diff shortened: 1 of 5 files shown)


safishamsi:
Landed on v8 (ships in 0.9.18), cherry-picked to preserve your authorship. Thanks @TPAteeq — graph.json/manifest.json now write to a temp file and os.replace into place (through a symlink, so shared-store setups keep working). Follow-up added: routed the writers the change had missed (the --no-cluster dump, merge-graphs/merge-chunks/merge-semantic, the analysis/labels sidecars, and the global graph/manifest), added an ensure_ascii flag so the raw-UTF-8 writers keep byte-for-byte output, and a Windows PermissionError-fallback test.

PR #1953, opened 2026-07-16
fix(merge-chunks): validate untrusted subagent chunk JSON before merging

TPAteeq:
### Problem
`graphify merge-chunks` concatenates agent-written `.graphify_chunk_*.json` files into one graph with only a JSON-decode guard around each file. Those chunks are untrusted output from semantic subagents, so an oversized or runaway payload, or a crafted node/edge id containing a path traversal (`../../etc/passwd`), flowed straight into the merged graph with nothing stopping it. The project already has a security-capping validator for exactly this threat model (#825 — memory exhaustion and chunk-directory escape via crafted IDs), but it was wired only into the skill merge path, not into the in-process `merge-chunks` command that feeds `build_from_json`.

### Fix
Route each chunk through `load_validated_semantic_fragment`, which stats the file size **before** reading it (so a multi-gigabyte chunk can't blow up memory at read time), parses the JSON, and validates the byte/count caps plus the node/edge/hyperedge id charset that blocks path separators and `..`. An invalid chunk is skipped with a warning (filter semantics — one bad chunk never aborts the merge). This validation is deliberately **not** wired into `build_from_json` / `load_graph_json`: those must keep loading valid, pre-existing graphs whose AST node ids predate the stricter semantic-id charset, and gating them would reject legitimate graphs.

While wiring this in, review caught that the shared validator was **too strict** in two ways that would have silently dropped whole legitimate chunks, so both were relaxed (a relaxation only — never a new rejection):

- `file_type` is no longer gated at all, because `build_from_json` already coerces any value via `_FILE_TYPE_SYNONYMS` (unknown → `concept`, #840) and common LLM outputs like `markdown`/`tool`/`framework` would otherwise fail validation wholesale.
- The id charset now allows Unicode word characters, because build's `normalize_id` deliberately preserves CJK/Cyrillic/accented-Latin identifiers (the explicit path-separator/`..` check still blocks directory escape).

The stale module docstring, which named the wrong skills, was corrected.

### Tests
`tests/test_merge_chunks_validation.py` and an update to `tests/test_semantic_cleanup.py`: path-escape id rejected; malformed shape rejected; valid chunks merge; a `file_type: "markdown"` / `"tool"` chunk now merges (regression guard for the data-loss blocker); a Unicode-id chunk merges; and the validator relaxation plus the still-active path-escape guard at the unit level. Full test suite green (3290 passed).

### DIFF
--- a/graphify/cli.py
+++ b/graphify/cli.py
@@ -3040,11 +3040,23 @@ def dispatch_command(cmd: str) -> None:
             chunk_files.extend(sorted(expanded) if expanded else [arg])
         merged: dict = {"nodes": [], "edges": [], "hyperedges": [], "input_tokens": 0, "output_tokens": 0}
         seen_ids: set[str] = set()
+        # These chunk files are untrusted subagent output. load_validated_...
+        # stats the file size BEFORE reading it (so a multi-GB chunk can't blow up
+        # memory), parses the JSON, and validates the security caps + the node/
+        # edge id charset that blocks path traversal (#825) — the same enforcement
+        # the skill merge path applies. A bad chunk is skipped with a warning
+        # (filter semantics; never abort). Deliberately NOT wired into
+        # build_from_json/load_graph_json, which must keep loading valid
+        # pre-existing graphs. file_type is left to build's coercion (#840).
+        from graphify.semantic_cleanup import load_validated_semantic_fragment
         for cf in chunk_files:
-            try:
-                chunk = json.loads(Path(cf).read_text(encoding="utf-8"))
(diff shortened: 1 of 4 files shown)


safishamsi:
Landed on v8 (ships in 0.9.18), cherry-picked so your authorship is preserved. Thanks @TPAteeq — merge-chunks now validates each untrusted chunk (reusing the #825 fragment validator) and skips a bad one with a warning instead of aborting the whole merge or passing a path-escape id through. Small follow-up added on top: non-numeric token counts are coerced so they cannot TypeError the merge after other chunks have merged.

PR #1955, opened 2026-07-16
fix(detect): spare genuine source from the Stage 1 sensitive-dir drop (#1943)

### LINKED ISSUE #1943: detect: Stage 1 sensitive-dir check silently drops legitimate source under secrets/ and credentials/ (follow-up to #1225)
### Summary

The Stage 1 sensitive-**directory** check in `detect.py` silently drops *every* file beneath a directory named `secrets/`, `.secrets/`, or `credentials/` — including legitimate programming-language source that just happens to live in a package with that name. The files vanish from the graph with no trace and no way to override.

This is the same false-positive class as #1225 (which fixed the Stage 3 *keyword* heuristic), but at the directory level, which #1225 did not touch.

### Repro

```
repo/
  internal/secrets/vault.go        # real Go source package
  app/services/credentials/manager.py  # real Python module
  secrets/db.json                  # actual credential store
```

`internal/secrets/vault.go` and `credentials/manager.py` are dropped from the graph. A quick check:

```python
from pathlib import Path
_is_sensitive(Path("internal/secrets/vault.go"))          # True  (wrong — this is source)
_is_sensitive(Path("app/services/credentials/manager.py"))# True  (wrong — this is source)
_is_sensitive(Path("secrets/db.json"))                    # True  (correct — real secret store)
```

`secrets/` and `credentials/` are just as often real source packages (Go `internal/secrets`, a `credentials/` service module) as credential stores, so pruning them wholesale by name loses legitimate source.

### Expected

Genuine programming-language source under an ambiguous `secrets/`/`credentials/` dir should be graphed, exactly the way Stage 3 already refuses to drop a `.py`/`.rb` source file on a keyword hit (#1225). Secret-prone data/config files (`.env`, `.json`, `.yaml`, key material) inside those dirs, and dedicated credential-store dirs (`.ssh`, `.gnupg`, `.aws`, `.gcloud`), must stay flagged.

### Proposed fix

Give Stage 1 the same source-code carve-out Stage 3 has: split `_SENSITIVE_DIRS` into dedicated credential stores (always skipped) and ambiguous bare-name dirs (skip everything except genuine source, via the existing `classify_file(...) == CODE and ext not in _SECRET_PRONE_DATA_EXTS` predicate). Credential stores keep their protection and `.graphifyinclude` still cannot rescue them.

alphanury:
Fixes #1943.

Stage 1 of `_is_sensitive` dropped every file beneath a directory named `secrets/`, `.secrets/`, or `credentials/` — including genuine source packages (Go `internal/secrets`, a `credentials/` service module). Same false-positive class as #1666, which already gave the Stage 3 keyword heuristic a source-code carve-out; this applies that carve-out at the directory level, as proposed in the issue.

## Change

- Split `_SENSITIVE_DIRS`:
  - `_CREDENTIAL_STORE_DIRS` (`.ssh`, `.gnupg`, `.aws`, `.gcloud`): everything beneath stays dropped, no carve-out — a `.py` inside `~/.ssh` is key tooling, not a package.
  - `_AMBIGUOUS_SENSITIVE_DIRS` (`secrets`, `.secrets`, `credentials`): dropped **except** genuine programming-language source.
- The existing Stage 3 predicate is extracted into `_is_graphable_source()` and shared by both stages, so the two carve-outs can't drift.
- Rescued source **falls through** to Stages 2–3 rather than short-circuiting to "safe": `secrets/service_account.py` (Stage 2 pattern) and `credentials/id_rsa` stay dropped.
- `parts[:-1]` parent-only semantics unchanged — #920's root-level-file guarantee still holds.

## Behavior (issue repro)

| path | before | after |
|---|---|---|
| `internal/secrets/vault.go` | dropped | **graphed** |
| `app/services/credentials/manager.py` | dropped | **graphed** |
| `secrets/db.json` | dropped | dropped |
| `.secrets/token.yaml` | dropped | dropped |
| `.ssh/config`, `.aws/credentials`, `.gnupg/helper.py` | dropped | dropped |

`.graphifyinclude` still cannot rescue credential stores (untouched here; #1921 tracks the include interplay for Stage 3).

## Safety notes

- Unreadable key material is safe on the new path: for extensionless files `classify_file` falls back to a shebang peek, and `_shebang_interpreter` swallows `OSError` — a `0600` file under `secrets/` classifies `None` → stays dropped, no crash.
- The shared predicate inherits Stage 3's stance that infra-config extensions (`.tf`, `.tfvars`, `.hcl`) classify as source. If that should tighten, adding them to `_SECRET_PRONE_DATA_EXTS` now covers both stages in one place.

## Tests

- 4 new tests: the issue's exact repro cases plus three guards (data/config under ambiguous dirs stay flagged; credential stores keep no-carve-out, source-classified files included; rescued source is still screened by Stages 2–3). Repro test confirmed red before the fix.
- `tests/test_detect.py`: 158 passed. `test_security` / `test_paths` / `test_extract` / `test_analyze` / `test_cache` / `test_export`: 379 passed, 1 skipped.


safishamsi:
Landed on v8 (ships in 0.9.18), cherry-picked to preserve your authorship. Thanks @alphanury — genuine source under an ambiguous secrets/ or credentials/ dir is spared while dedicated stores still drop everything. Follow-up added on top: .tfvars is now in the secret-prone extension set, so Terraform values files stay out of the graph (they routinely hold real secrets) while .tf/.hcl source stays graphable.

Issue #1890, opened 2026-07-14
extract: LLM-omitted documents vanish silently — no dispatch/return reconciliation, no anti-regression guard (53% of docs lost even with gpt-5)


AndreLFSMartins:
## Summary

`graphify extract` dispatches every detected document to the LLM but **never verifies that a node came back for each one**. Documents the model silently omits simply vanish from the graph — no warning, no count mismatch, no error. And unlike `update`, the `extract` path has **no anti-regression guard**, so the shrunken graph overwrites the good one unconditionally.

Coverage degrades with model strength, but **is never complete, even with a frontier model**.

## Measurement

Repo: `council` (Python + heavy docs). `detect()` classifies **33 code + 265 document** files. Same repo, same command, three runs:

| build | nodes | edges | files covered | `.md` covered (of 265 docs) |
|---|---|---|---|---|
| existing graph (built via the Claude Code skill) | 3,085 | 3,673 | 229 | **166** |
| `extract --backend openai --model gpt-4.1-mini` | 1,679 | 2,455 | 118 | **47** |
| `extract --backend openai --model gpt-5` | 2,135 | 3,167 | 276 | **125** |

So with `gpt-5` — the strongest model available — **140 of 265 detected documents (53%) produced no node at all**. The run log is clean: zero `chunk ... failed`, zero `invalid JSON`, zero truncation retries. The model simply didn't emit nodes for them, and graphify accepted that silently.

The dropped files are ordinary tracked sources — `README.md`, `CLAUDE.md`, `AGENTS.md`, `PLAN.md`, `docs/objectives.md`. All exist on disk; only 2 of the missing set are gitignored. Code coverage is unaffected (AST is deterministic), which is what makes this hard to notice: the graph still answers code queries correctly while half the documentation is gone.

## Why this is a bug and not just "use a better model"

1. **No dispatch/return reconciliation.** graphify knows it sent N documents to the LLM. It knows it got M nodes back. It never compares N to M. A file that goes in and produces nothing should at minimum be logged.
2. **No guard on the `extract` path.** The guard that protects `update` —
   `WARNING: new graph has N nodes but existing graph.json has M. Refusing to overwrite.`
   — does not fire here. `extract` overwrites regardless of how much coverage was lost.
3. **The failure is invisible and cumulative.** A user rebuilding a docs-heavy repo loses half their document graph and has no signal that it happened.

## Suggested fixes

[...]

PR #1956, opened 2026-07-16
fix(detect): clear stale semantic_hash for dispatched-but-omitted files (#1948)

### LINKED ISSUE #1948: save_manifest seeds stale semantic_hash from prior manifest, masking LLM-omitted docs — the re-run promised by #1890 never happens
### Summary

When the model omits documents from an extraction response, the #1890 warning promises:

> The model returned a response but omitted them; a re-run will retry them.

and `_stamped_manifest_files` (cli.py, #933) filters those files out of the dict passed to `save_manifest` so their `semantic_hash` stays empty and `detect_incremental` re-queues them.

That mechanism only works for files that have **never** been stamped. If a file was successfully extracted in *any previous run*, `save_manifest` seeds the new manifest from the existing one on disk and the file's **old** `semantic_hash` survives the omission. On the next incremental run the content hash still matches that inherited stamp, so the file is reported `unchanged` and is never retried. The #1890 promise is silently broken for exactly the runs where it matters most (`--force` re-extractions over an existing manifest).

Observed on 0.9.17 (also present at v8 HEAD, 0.9.18 dev): a `graphify extract . --backend gemini --force` over a 651-doc Obsidian vault ended with

```
[graphify] WARNING: 110/651 dispatched file(s) produced no nodes and are absent from the graph: … The model returned a response but omitted them; a re-run will retry them.
```

and the immediately following incremental run reported

```
0 docs changed; 683 unchanged; 0 re-extracted
```

Spot-checking one omitted file (`AI/sessions/2026-06-12-….md`, 0 nodes in graph.json) confirmed its `graphify-out/manifest.json` entry still carried the `semantic_hash` from an earlier successful run.

### Root cause

`save_manifest` (graphify/detect.py) is merge-by-seed, not replace:

1. **Seed loop** (detect.py ~1557–1569): every row of the *existing* manifest whose file is still on disk and still in the scan corpus is copied into the new manifest **verbatim**, stale `semantic_hash` included. (This seeding is intentional and correct for subset-saving callers, #917.)
2. **Stamp loop** (detect.py ~1578–1593): only iterates over the `files` dict the caller passed in. Files removed by `_stamped_manifest_files` (cli.py:57, called at cli.py ~2729) are not in it, so their seeded row is never overwritten *or cleared*.
3. `detect_incremental(kind="semantic")` then compares the file's current content hash against the inherited `semantic_hash`. The content hasn't changed since the old successful extraction, so it matches → `unchanged` → never re-queued.

So the #933/#1890 "leave it unstamped so it gets retried" contract holds only for first-ever extractions. Once a file has a stamp, every later omission (model drops it from a response, failed chunk on a `--force` run, …) is permanently masked.

### Reproduction (deterministic, no LLM needed)

Simulates run 1 = successful extraction, run 2 = `--force` re-run where the model omits the doc (so `_stamped_manifest_files` filtered it out of the files dict), then checks whether the promised retry happens:

```python
import json, sys, tempfile
from pathlib import Path
from graphify.detect import save_manifest, detect_incremental

tmp = Path(tempfile.mkdtemp()).resolve()   # resolve(): avoid macOS /var symlink aliasing
root = tmp / "vault"; root.mkdir()
doc = root / "doc.md"; doc.write_text("stable content that never changes\n")
manifest_path = str(tmp / "graphify-out" / "manifest.json")
corpus = {str(doc)}

# run 1: doc.md extracted successfully -> stamped
save_manifest({"document": [str(doc)]}, manifest_path, kind="both", root=root, scan_corpus=corpus)

# run 2 (--force): model omits doc.md -> _stamped_manifest_files drops it from the dict
save_manifest({"document": []}, manifest_path, kind="both", root=root, scan_corpus=corpus)

entry = next(iter(json.loads(Path(manifest_path).read_text()).values()))
print("semantic_hash after omission:", repr(entry["semantic_hash"]))   # expected: ''

inc = detect_incremental(root, manifest_path, kind="semantic")
requeued = [f for fl in inc["new_files"].values() for f in fl]
print("re-queued on next run:", requeued)                              # expected: [doc.md]
```

Output on v8 HEAD:

```
semantic_hash after omission: 'c561585056bce0a38112036a060bac99'
re-queued on next run: []
```

Expected: `semantic_hash` empty after run 2 and `doc.md` re-queued on the next incremental run, per the #1890 warning text and the #933 design.

### Suggested fix direction

The seed loop can't just stop preserving stamps — that would break subset-saving callers (#917). The caller needs a way to say "these files were dispatched but produced no output; clear their semantic stamp":

- `extract` already knows the exact set — `merged["uncovered_files"]` is computed for the #1890 warning (graphify/llm.py ~2081), and `_stamped_manifest_files` re-derives the same membership.
- Add e.g. a `clear_semantic: set[str] | None` parameter to `save_manifest`; seeded rows matching it (after the same root-resolution as #1897) get `semantic_hash = ""` instead of the inherited value. Full-pipeline callers in cli.py pass the uncovered/unstamped set; #917 subset callers keep passing nothing and are unaffected.

### Workaround

Manually blank `semantic_hash` in `graphify-out/manifest.json` for every source file with 0 nodes in graph.json, then run incremental extract — the 110 omitted docs were picked up and retried correctly after that.

### Environment

- graphify 0.9.17 (bug also present on v8 HEAD / 0.9.18 dev cycle)
- macOS (Darwin 25.6.0), Python 3.x
- backend: gemini (but the masking is backend-independent; only the omission rate varies)

alphanury:
Fixes #1948.

`save_manifest`'s seed loop copied every on-disk manifest row verbatim for files still alive on disk, with no way to distinguish "never touched this run" from "dispatched this run but omitted from the result". So a doc whose chunk the LLM dropped on a `--force` re-run kept its stale `semantic_hash` from the earlier successful run, `detect_incremental(kind="semantic")` reported it unchanged, and the #1890 re-run promise silently never happened — exactly the failure this issue describes, and what the manual "blank the hash by hand" workaround was compensating for.

## Change

- `save_manifest()` (graphify/detect.py) gains a keyword-only `clear_semantic` set: for any file in it, the seed loop forces `semantic_hash` to `""` instead of inheriting the prior run's value. Defaults to `None` (no-op) — every existing caller and the #917/#1908 seed/prune behaviors are unchanged (guard tests pass unmodified).
- `cli.py` derives the set as **`semantic_files` minus `_stamped_manifest_files()`'s result** — i.e. *dispatched this run but not stamped, regardless of why*. `semantic_files` is what was actually sent to the backend (narrowed by the incremental gate and `--code-only`, widened by deep mode), so untouched live files are never in the set and a partial incremental run cannot blank the rest of the corpus's stamps.
- The set is threaded through all three `_save_manifest` call sites.

## Review note (why the derivation matters)

An earlier draft of this branch derived the set from `files_by_type` (the full live corpus). Adversarial review before pushing caught that this wipes `semantic_hash` for every unchanged file on any partial incremental run — forcing a full-corpus re-extraction on the next run, which would have been worse than the bug. The committed version derives from the dispatched subset and adds an end-to-end regression test for exactly that scenario.

## Tests

- Unit: `test_save_manifest_clear_semantic_erases_stale_hash_for_omitted_file` (red before the detect.py change, green after), plus the seven pre-existing seed/prune guards, unmodified.
- End-to-end (new): `test_incremental_partial_run_preserves_untouched_semantic_hash` — full-scan run stamps two docs, a second incremental run dispatches only the changed one (asserted via the captured dispatch list), and the untouched doc's `semantic_hash` must survive in `manifest.json`. Red on the `files_by_type` derivation, green on the committed one.
- Suites: `test_extract_cli.py` + `test_extract_code_only_cli.py` + `test_detect.py` + `test_build.py` → 226 passed. Full suite: 3269 passed; the 15 failures in `test_ollama_retry_cap.py`/`test_skillgen.py` are pre-existing on a clean v8 checkout (verified byte-identical before/after).


safishamsi:
Landed on v8 (ships in 0.9.18), cherry-picked to preserve your authorship. Thanks @alphanury — save_manifest no longer seeds a stale semantic_hash for a dispatched-but-omitted file, so the #1890 re-run actually fires. Rebased onto current v8 where it composes with #1950: a truncated (partial) file is also dropped from the stamped set and therefore cleared/re-queued. Added a regression test pinning that #1948 x #1950 interaction.

Issue #1968, opened 2026-07-17
export obsidian: node notes are bare (frontmatter + Connections only) — wire each note to its source content


MBasithC:
## Context

Docs-only corpus: 167 markdown pages (the official Claude Code docs, ~650K words) → 1,490 nodes / 1,916 edges / 149 communities, exported with the Obsidian exporter. graphifyy 0.9.16 on Windows.

## Problem

Every exported node note contains only:

- frontmatter (`source_file`, `type`, `community`, tags)
- the H1 title
- `## Connections` (wikilinked edges)
- a tag line

There is no substance from the source in the note itself. The graph structure is excellent, but browsing the vault every note reads as a bare stub — a human (or an agent hopping between notes) has to leave the vault to learn anything about the node. `source_file` exists in frontmatter but is a dead string, not a link to anything inside the vault.

## Workaround we run on top of every export

A post-processing pass that:

1. Copies each source doc into the vault under `sources/` as `_src_<flattened-path>.md` (collision-safe flattened names, boilerplate stripped, frontmatter added).
2. Inserts a callout after each node note's H1:

```markdown
> [!info] Source doc
> [[_src_en_best-practices|en/best-practices.md]] - [read on docs site](https://code.claude.com/docs/en/best-practices)
```

Result: 1,490/1,490 notes wired, 0 unresolved links — and the vault becomes genuinely usable standalone. But the exporter rewrites the notes on every re-export, so the pass has to be re-run each time.

## Feature request

An opt-in flag on the Obsidian export, e.g. `--with-sources`, that:

- copies source files into the vault (`sources/`, collision-safe names) and turns each note's `source_file` frontmatter into a real wikilink/callout to that source note;
- for code corpora, the analog would be embedding the node's defining snippet (or linking a per-file source note).

Even just "make `source_file` a resolvable link" would remove most of the pain. Happy to share the full post-processing script if useful.

PR #1973, opened 2026-07-17
feat(export): add --with-sources to wire Obsidian notes to source content (#1968)

SinghAman21:
Wire Obsidian node notes to their source content (--with-sources) — #1968

Why

Right now every note in an exported Obsidian vault is a bare stub: frontmatter, the H1 title, and a ## Connections list of wikilinks. The source_file field exists in frontmatter but it's just a dead string — it doesn't link to anything. So anyone browsing the vault (or an agent hopping between notes) has to leave the vault entirely to learn what a node actually is. The graph structure is great, but the notes carry no substance.

Users were working around this with a post-processing script that copies sources in and rewires every note — but the exporter overwrites the notes on each re-export, so they had to re-run the script every single time.

What

A new opt-in flag on the Obsidian export:

graphify export obsidian --with-sources

With it, the vault becomes usable standalone:
- Every distinct source_file is copied into the vault under sources/ as a real note (_src_<flattened-path>.md).
- Each node note gets a callout right under its title linking to that source:
- Optional --source-root PATH to override where source paths are resolved from disk.

How

- graphify/export.py — to_obsidian gains with_sources / source_root params plus four small helpers:
  - resolve each source_file (stored relative to the scan root) to a real file on disk;
  - flatten nested/Windows paths into one collision-safe note name;
  - copy markdown/text verbatim, and embed code/other files in a syntax-highlighted fence;
  - write one deduped note per distinct source, then insert the callout after each node's H1.
- graphify/cli.py — parses the two new flags and defaults resolution to [scan root, out dir] (the out-dir fallback keeps older pre-0.9.17 graphs working).

Behavior chosen deliberately:
- Source notes are graphify-owned, so the existing manifest/prune logic cleans them up automatically — a re-export without the flag removes them and the vault returns to plain notes.
- A node whose source can't be found on disk (or exceeds a 2 MB cap) simply gets no callout rather than a dangling link.
- Nodes sharing a source_file all link to the same single source note.

Conclusion

This solves the issue's core ask — turn source_file into a resolvable link and bring source content into the vault — without any manual post-processing. It's fully opt-in, so default exports are unchanged, and it plays cleanly with the existing re-export/prune machinery. Covered by a new test file (tests/test_obsidian_with_sources.py, 5 tests) with the existing export suite (91 tests) still green.


SinghAman21:
@safishamsi it'd would be great if you review this PR and lmk any improvements

SinghAman21:
@safishamsi it will be great if you could review this, else PR will become stale and conflicts would occur thanks.

SinghAman21:
@safishamsi :)

PR #1976, opened 2026-07-17
fix(resolution): repoint cross-extension re-exports without leaking an absolute path (#1814)

### LINKED ISSUE #1814: Phantom self-cycles: same-basename cross-extension re-exports (X.ts → ./X.mjs) collapse to one node
**Version observed:** graphify (PyPI `graphifyy`) 0.9.12

### Summary
When a source file re-exports from a *same-basename* sibling with a different extension — e.g. `foo.ts` doing `export { … } from "./foo.mjs"` — graphify records a false **self-loop** (`foo.ts → foo.ts`) and reports it as a 1-file import cycle. The extension is stripped during edge resolution, so the re-export target `./foo.mjs` is matched back to the `foo.ts` node instead of a distinct `foo.mjs` node.

### Repro
```
mkdir demo && cd demo
# foo.mjs — hand-written plain-ESM runtime
printf 'export const N = 1;\n' > foo.mjs
# foo.ts — typed wrapper that re-exports the runtime
printf 'export { N } from "./foo.mjs";\n' > foo.ts
graphify update .
```
Then inspect:
```
graphify explain "foo.ts" --graph graphify-out/graph.json
```

### Expected
Two nodes (`foo.ts`, `foo.mjs`) with one directed edge `foo.ts —re_exports→ foo.mjs`. No cycle.

### Actual
One node whose id is the extension-stripped basename, with a self-edge:
```
--> foo.ts [re_exports]
<-- foo.ts [re_exports]
```
and `foo.ts` is listed under **Import Cycles** in `GRAPH_REPORT.md`. In the graph JSON the edge is `relation: "re_exports"` with `source == target` and a node id like `…foo_ts_…foo` (the `.ts` node id concatenated with the extension-stripped target).

### Root cause (best guess)
Node identity for an import/re-export target is keyed by path with the extension stripped, so `foo.ts` and `foo.mjs` resolve to the same node id. The manifest *does* index `.mjs`/`.d.mts` as separate files — it's specifically the edge/import resolver's basename matching that collapses them.

### Why the obvious workaround doesn't fit
This pattern is a legitimate, common one: a hand-written `.mjs` plain-ESM runtime (so it's importable by Node/CI without a TS build step) plus a thin typed `.ts` wrapper (`export … from "./X.mjs"`) and a hand-written `.d.mts`. So:
- Excluding `dist/` / `build/` does nothing — the files aren't build output.
- Excluding `**/*.d.mts` doesn't remove the loop (the loop is `.ts → .mjs`).
- Excluding `**/*.mjs` would delete real runtime source.

### Suggested fix
Key nodes by full path *including* extension so `foo.ts` and `foo.mjs` are distinct nodes (then `foo.ts → foo.mjs` is a correct 2-node edge and the phantom cycle disappears). Equivalently: when a re-export specifier carries an explicit extension that resolves to a different physical file, don't strip it back to the same-basename sibling.

### Impact
Over-reports import cycles for any codebase using the ESM-runtime + typed-wrapper convention (7 phantom cycles surfaced in a single app), which can mislead a cycle-cleanup pass into "fixing" non-problems.

alphanury:
Fixes #1814.

## Summary

A same-basename cross-extension re-export — a typed `.ts` wrapper re-exporting a hand-written `.mjs` runtime, `export { N } from "./foo.mjs"` — produced a phantom self-cycle: `foo.ts → foo.ts`, reported as a 1-file import cycle in `GRAPH_REPORT.md`.

## Root cause

`foo.ts` and `foo.mjs` collapse onto one base file id (the id stem drops the extension). `_disambiguate_colliding_node_ids` correctly salts the two *file nodes* apart (`foo_ts_foo` / `foo_mjs_foo`), but the re-export *edge* keyed its target salt by the **importer's own** `source_file`. So the `./foo.mjs` target was salted back onto `foo.ts` — a `source == target` self-loop. The residual was purely at the edge layer; node ids were already correct.

## Fix

When an import/re-export edge is emitted, the emitter now stamps the **resolved target file** on the edge as a transient `target_file` hint. During disambiguation, an `imports` / `imports_from` / `re_exports` edge keys its target salt by that file instead of the importer's, so the salt lands on the real sibling node and the phantom cycle disappears. This generalizes the C/ObjC `.h`-sibling carve-out from #1475 to every language and to `re_exports`.

The hint has no downstream reader and holds an absolute path, so it must never persist into `graph.json`:
- it is `pop`ed the moment disambiguation consumes it (both the colliding and the no-collision early-exit paths);
- the graph serializer drops any `target_file` as a backstop, so an incremental `build_merge` that re-serializes base edges without re-running disambiguation can't leak one either.

`graph.json` stays deterministic and byte-identical across checkout locations. Node ids are unchanged, so no forced rebuild. (The extension-aware-id alternative was rejected: it would rewrite every file and symbol id and force a full-rebuild migration in lockstep with the skill/validation id spec, #1033.)

One caveat, called out in the CHANGELOG: a graph written by a *pre-fix* build still records the stale self-loop, and because `graphify update` only re-extracts changed files, an unchanged wrapper keeps that edge until it is next edited or a `--force` full rebuild runs.

## Testing

New `tests/test_cross_extension_reexport_self_cycle.py` (10 cases): the reported `.ts`/`.mjs` repro, static + dynamic + `require` import forms, star and namespace re-exports, the no-collision path (target salt untouched), and a serializer check that no `target_file` survives into the persisted graph.

- New file: 10 passed.
- Full suite: 3363 passed, 33 skipped. The only failures (4, in `tests/test_ollama_retry_cap.py`) are pre-existing and unrelated — they fail identically on a clean `v8` checkout (an ollama SDK module missing from the test env).


safishamsi:
Landed on v8 (ships in 0.9.19) in d56b70a, cherry-picked to preserve your authorship. Thanks @alphanury for the fix and @Greg-Moskalenko for the report — a same-basename cross-extension re-export (foo.ts re-exporting ./foo.mjs) no longer manufactures a phantom self-cycle; the re-export edge target is salted by the resolved sibling file, no absolute path leaks, node ids unchanged.

PR #1979, opened 2026-07-17
fix: add --no-gitignore extraction opt-out

### LINKED ISSUE #1971: Allow extracting git-ignored code: add `--no-gitignore` opt-out
### Summary

Since ignore files *below* the scan root are read live during the walk (`#1206`) and `.gitignore` is merged into the ignore logic (`#1363`), `graphify extract` now honors every `.gitignore` it encounters anywhere under the scan root. That is usually correct — but there is **no way to turn it off**, and it silently removes source files that are deliberately git-ignored yet are exactly what we want in the graph.

Concretely: a large body of **transpiled/generated C# code** lives in `generated/` directories that the project's `.gitignore` excludes (generated output is conventionally not committed). That code *is* the behavior we analyze. On a version that only read the scan root's own ignore files this code was extracted; after upgrading, all of it disappears — no error, no warning, the files simply don't appear in the graph, so the loss is easy to miss.

### Environment

- `graphifyy`: a version that did **not** read nested `.gitignore` vs. current that **does** (0.9.13 → 0.9.18 in our case)
- Windows 11, Python 3.14, C# corpus (nodes are `_origin: ast`)

### Minimal reproduction (verified on 0.9.18)

```sh
mkdir -p root/proj/deep/generated
printf 'namespace N { public class Real { public void M(){} } }\n' > root/proj/Real.cs
printf 'namespace N { public class Keep { public void M(){} } }\n' > root/proj/deep/Keep.cs
printf 'namespace N { public class Gen  { public void M(){} } }\n' > root/proj/deep/generated/Gen.cs
printf 'generated/\n' > root/proj/.gitignore        # nested, BELOW the scan root

graphify extract root --out out                     # exits 0, no warning
```

Inspect `source_file` on the nodes in `out/graphify-out/graph.json`.

**Observed (actual):** only `proj/Real.cs` and `proj/deep/Keep.cs` are present.
`proj/deep/generated/Gen.cs` is **absent** — silently pruned because `proj/.gitignore` excludes `generated/`.

**Desired:** a supported way to include `Gen.cs` despite the `.gitignore` exclusion (see request below).

On a real transpiled graph the aggregate impact is large: distinct `.cs` files 1200 → 936 (−264, **all** under `generated/`); AST nodes −13%; edges −29%.

### What we ruled out while diagnosing

- **Not the parser:** identical `tree-sitter` (0.25.2) / `tree-sitter-c-sharp` (0.23.5); the dropped files parse fine in isolation (`extract_csharp(path)` returns 45 nodes, no error).
- **Not dedup, not a file-size cap, not `_SKIP_DIRS`** (there is no `generated` literal in the skip set).
- **Confirmed mechanism:** `detect._load_dir_own_ignore(proj)` reads `proj/.gitignore` during the walk and returns the `generated/` pattern, which prunes the directory. This function/behavior is only present in the newer version.

### Only workaround we could find (verified on 0.9.18)

Place a `.graphifyignore` containing **both** `!generated/` and `!generated/**` **in the same directory as** the offending `.gitignore` (`root/proj/.graphifyignore`). Re-running the repro then includes `proj/deep/generated/Gen.cs`.

- This works because per-directory `.graphifyignore` is read *after* `.gitignore` (`_load_dir_own_ignore`), so the `!` negation wins via last-match-wins and the directory is not pruned.
- A re-include placed at the **scan root** does **not** work: the nested `.gitignore` is loaded later during the walk and wins. So the escape hatch must be co-located with *each* offending `.gitignore`, which is non-obvious and awkward to automate.

### Questions / requests

1. Is there an existing flag or config to disable honoring `.gitignore` (and `$GIT_DIR/info/exclude`) during extraction? We could not find one.
2. If not, would you consider a `--no-gitignore` / `--ignore-vcs-ignores` opt-out? Honoring `.gitignore` is a sensible default, but generated/transpiled code that is git-ignored yet analysis-relevant is a real case (build output, codegen, transpiler targets), and there is currently no clean escape hatch.
3. If the intended answer is "use a co-located `.graphifyignore` re-include", could the precedence (a nested `.gitignore` beats a scan-root `.graphifyignore`; only a co-located `!` wins) be documented? It is currently surprising.

Either way, this project is great, and I'd like to thank everyone who's actively contributing to it. Have a great weekend. <3

Mzt00:
## Summary

Adds `graphify extract <path> --no-gitignore` for generated or transpiled source that is excluded from Git but still belongs in the graph. Fixes #1971

With the flag, Graphify:

- skips root, ancestor, and nested `.gitignore` files;
- skips `$GIT_DIR/info/exclude`;
- still prioritizes `.graphifyignore` explicit `--exclude` patterns, noise filters, and sensitive-file checks; and
- persists the setting for later update, watch, and hook rebuilds.

Default extraction behavior is unchanged.

## Implementation

The detector APIs now accept a keyword-only `gitignore: bool = True`. The extract CLI passes `False` for both full and incremental scans when `--no-gitignore` is present.

The setting is stored in `graphify-out/.graphify_build.json`. Rebuild and watch paths read it before rescanning or filtering filesystem events, preventing generated files from disappearing after the initial extraction. Existing build configs without the key default to `True`.

### Why `.graphifyignore` remains active

The staged loader change separates Graphify-owned rules from VCS-owned rules. With `gitignore=False`, `_load_dir_own_ignore()` loads only `.graphifyignore`; `_load_graphifyignore()` skips `.git/info/exclude` but still loads `.graphifyignore` at ancestor, root, and nested-directory anchors. `detect()` and watch/rebuild callers pass the same setting through every scan path.

The end-to-end test confirms this behavior: Git-ignored `Gen.cs` and `Local.cs` are included, while `.graphifyignore`-excluded `Hidden.cs` remains absent.

## Behavior

| Rule source | Default | `--no-gitignore` |
|---|---:|---:|
| `.gitignore` | Applied | Skipped |
| `.git/info/exclude` | Applied | Skipped |
| `.graphifyignore` | Applied | Applied |
| `--exclude` | Applied | Applied |
| Noise and sensitive-file filters | Applied | Applied |

## Tests and validation

- End-to-end extraction verifies VCS-ignored C# is included while `.graphifyignore` remains active.
- Rebuild coverage verifies the persisted setting continues to include generated code.
- Existing tests guard default Git-ignore behavior and persisted explicit exclusions.
- Ruff passed for all changed Python files.
- Focused regression set: **7 passed**.

### DIFF
--- a/README.md
+++ b/README.md
@@ -396,6 +396,8 @@ Create a `.graphifyignore` in your project root — same syntax as `.gitignore`,
 
 **`.gitignore` is respected automatically.** graphify reads the `.gitignore` in each directory. If a `.graphifyignore` is also present, the two are **merged** — `.graphifyignore` patterns are evaluated last, so they win on conflicts (including `!` negations). Adding a `.graphifyignore` only ever excludes more; it never re-includes a file your `.gitignore` already excluded. Subdirectory scoping works the same way as git — an ignore file only affects its own subtree.
 
+Pass `--no-gitignore` to `graphify extract` when git-ignored generated or transpiled code belongs in the graph. This disables `.gitignore` and `.git/info/exclude`; `.graphifyignore` still applies.
+
 ```
 # .graphifyignore
 node_modules/
@@ -714,6 +716,7 @@ graphify extract ./docs --token-budget 30000   # smaller semantic chunks for loc
 graphify extract ./docs --max-concurrency 2    # fewer parallel LLM calls (useful for local inference)
 graphify extract ./docs --api-timeout 900      # longer HTTP timeout for slow local models (default 600s)
 graphify extract ./docs --google-workspace     # export .gdoc/.gsheet/.gslides via gws before extraction
+graphify extract ./src --no-gitignore          # include git-ignored source; still honor .graphifyignore
 graphify extract ./docs --mode deep            # richer semantic extraction via extended system prompt
 graphify extract ./docs --no-cluster           # raw extraction only, skip clustering
 graphify extract ./docs --timing               # print per-stage wall-clock timings to stderr (also works on cluster-only)
--- a/graphify/__main__.py
+++ b/graphify/__main__.py
@@ -605,6 +605,7 @@ def _run_cli() -> None:
         print("    --api-timeout S         per-request timeout in seconds for the LLM client (default: 600)")
         print("    --out DIR               output dir (default: <path>); writes <DIR>/graphify-out/")
         print("    --google-workspace      export .gdoc/.gsheet/.gslides shortcuts via gws before extraction")
+        print("    --no-gitignore         ignore .gitignore and .git/info/exclude (prioritizes .graphifyignore)")
         print("    --no-cluster            skip clustering, write raw extraction only")
         print("    --code-only             index code (local AST, no API key) and skip doc/paper/image files")
         print("    --postgres DSN          extract schema from a live PostgreSQL database")
--- a/graphify/cli.py
+++ b/graphify/cli.py
@@ -2122,6 +2122,7 @@ def dispatch_command(cmd: str) -> None:
             print(
                 "Usage: graphify extract <path> [--backend gemini|kimi|claude|openai|deepseek|ollama] "
                 "[--model M] [--mode deep] [--out DIR] [--google-workspace] [--no-cluster] "
+                "[--no-gitignore] "
                 "[--max-workers N] [--token-budget N] [--max-concurrency N] "
                 "[--api-timeout S] [--postgres DSN] [--cargo] [--allow-partial] [--timing]",
                 file=sys.stderr,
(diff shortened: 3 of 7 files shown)


safishamsi:
Landed on v8 (ships in 0.9.19) in f17e2c5, cherry-picked to preserve your authorship. Thanks @mzt006 — --no-gitignore disables only VCS ignore rules while keeping .graphifyignore and the secret/sensitive screens. Follow-up added on top (caa6f9e): the setting now persists and is not clobbered back on by a later flag-less graphify extract, so git-ignored code does not silently disappear (the #1971 complaint).

safishamsi:
Credit correction: this fix is by @Mzt00 (I earlier mistyped @mzt006, an unrelated user — apologies to them). Report was @JensD-git. Fixed in the changelog and 0.9.19 release notes.

PR #1980, opened 2026-07-17
fix(prs): decode gh/git/claude output as UTF-8, not the Windows cp1252 codec

LukeTheoJohnson:
## Problem

On Windows, `graphify prs` crashes against any repo that has non-ASCII PR metadata. `prs.py` reads `gh`, `git`, and `claude` output through `subprocess.run(..., text=True)` with no explicit `encoding=`, so stdout is decoded with the locale codec, which is cp1252 on a default Windows install. `gh` emits UTF-8 JSON whose PR titles and author logins routinely contain non-Latin1 bytes (emoji, or the Persian title in this repo's own PR #1281), and cp1252 has no mapping for bytes such as 0x81.

The failure is indirect. The capture reader thread raises `UnicodeDecodeError`, `subprocess.run` swallows that and returns `stdout=None`, and the caller then dies one line later:

- `_gh` runs `json.loads(None)` and raises `TypeError`
- `fetch_pr_files` / `fetch_worktrees` run `None.splitlines()` and raise `AttributeError`

Neither is in `_gh`'s `except (TimeoutExpired, JSONDecodeError, FileNotFoundError)`, so the command aborts, and a spurious reader-thread traceback is printed to stderr as well.

## Root cause

`text=True` without `encoding=` uses `locale.getpreferredencoding()` for decoding (cp1252 on Windows). CI runs on `ubuntu-latest` only, where the locale is UTF-8, so this path is never exercised upstream.

## Fix

Pass `encoding="utf-8", errors="replace"` to all five subprocess reads in `prs.py`. This is the decode-side counterpart of #1505, which applied the same change to the `llm.py` claude-cli subprocess; `prs.py` was not touched then.

## Reproduction

Windows 11, CPython 3.13, cp1252 locale:

```python
>>> from graphify import prs
>>> prs._gh("pr", "view", "1281", "--repo", "safishamsi/graphify", "--json", "title")
# before: TypeError: the JSON object must be str, bytes or bytearray, not NoneType
# after:  {'title': 'docs: add Persian (فارسی) translation and language selector link'}
```

## Tests

`tests/test_prs.py::TestSubprocessOutputEncoding` asserts `encoding="utf-8"` on each of the five calls, mirroring the existing `tests/test_charmap_encoding.py`. They are red without the fix (the four behavioural asserts fail with `assert None == 'utf-8'`). A guard test confirms the fixture bytes are genuinely cp1252-undecodable, so the tests exercise the real failure surface. A behavioural decode test can't be cross-platform here, since Linux CI decodes UTF-8 regardless, so the tests assert the contract instead.

## Scope

Five subprocess calls in `prs.py`; no behaviour change on Linux or macOS, where UTF-8 is already what they decode with. Other modules have their own `text=True` calls but decode ASCII-only output (git hashes, refs), so I left them out to keep this focused. Happy to harden those in a follow-up if you'd like.

### DIFF
--- a/graphify/prs.py
+++ b/graphify/prs.py
@@ -142,7 +142,10 @@ def _gh(*args: str) -> list | dict | None:
     try:
         result = subprocess.run(
             ["gh", *args],
-            capture_output=True, text=True, timeout=30
+            # Decode gh's output as UTF-8, not the Windows cp1252 locale codec: gh
+            # emits UTF-8 JSON with non-Latin1 titles/logins (emoji, فارسی), and the
+            # default text=True decode crashes on those (#1505 fixed the same in llm).
+            capture_output=True, text=True, encoding="utf-8", errors="replace", timeout=30
         )
         if result.returncode != 0:
             return None
@@ -164,7 +167,7 @@ def _detect_default_branch(repo: str | None = None) -> str:
     try:
         result = subprocess.run(
             ["git", "symbolic-ref", "refs/remotes/origin/HEAD"],
-            capture_output=True, text=True, timeout=5
+            capture_output=True, text=True, encoding="utf-8", errors="replace", timeout=5
         )
         if result.returncode == 0:
             # refs/remotes/origin/main → main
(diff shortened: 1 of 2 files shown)


safishamsi:
Landed on v8 (ships in 0.9.19) in a86666a, cherry-picked to preserve your authorship. Thanks @Luke J — all gh/git/claude subprocess reads in prs.py now decode as UTF-8 with errors=replace (matching #1505), fixing the Windows cp1252 mojibake/crash on emoji and non-ASCII PR titles/diffs.

PR #1985, opened 2026-07-18
fix: indirect JS/TS call resolution for dispatch tables and callbacks

Tefa-hup:
# Ready-to-submit GitHub issue (or PR description) for Graphify-Labs/graphify

Repo: https://github.com/Graphify-Labs/graphify (redirects from https://github.com/safishamsi/graphify)
Diff is against `main` @ `8cee776cee7d5dc3279085797bde2d2dcce21e36` (2026-07-17).

I don't have push/fork access to this repo from my GitHub account, so I could not open
this as an issue or PR directly. Steps to submit it yourself:

1. Fork https://github.com/Graphify-Labs/graphify on GitHub (web UI button, top right).
2. Clone your fork locally, or apply the patch file directly to a fresh clone:
   `git apply graphify-js-indirect-call-fix.diff` (from the repo root — the diff already
   has correct paths for `graphify/extractors/engine.py` and `tests/test_indirect_dispatch.py`).
3. Commit, push to your fork, open a PR against `Graphify-Labs/graphify:main` using the
   title/body below (or open it as an issue first if you'd rather report it without a PR).

---

## Title

JS/TS AST extractor: two false-positive `indirect_call` edge sources (for-of destructuring, `module.exports = {...}`)

## Body

Found while auditing a `graphify` run against a real Node.js codebase: several `indirect_call` edges in the graph did not correspond to any real call in the source. Root-caused to two distinct bugs in `graphify/extractors/engine.py`, both in the JS/TS indirect-dispatch resolution path.

### Bug 1 — `for (const [a, b] of x)` loop variables aren't recognized as local bindings

`_js_local_bound_names`'s `walk()` (around L1151) only special-cases `variable_declarator` nodes to collect bound names. But a `for_in_statement`'s loop variable(s) live directly in its `left` field — never wrapped in a `variable_declarator`:

```
for_in_statement
  kind: const
  left: array_pattern            <- not a variable_declarator!
  operator: of
  right: call_expression
  body: statement_block
```

So a destructured loop var like `count` in `for (const [file, count] of Object.entries(x))` is never added to the function's local-shadow set. A later bare read of it (`if (count >= 3)`) then gets treated as a reference to a same-named module-level function, and if a same-named callable exists ANYWHERE in the corpus, the cross-file resolver (`extract.py` ~L4905-4927) manufactures a phantom `indirect_call` edge to it — confidence_score 0.8, i.e. presented as fairly confident.

Concretely: a file with `for (const [file, count] of editCounts) { if (count >= 3) {...} }` and an unrelated `JsonFileBackend.count()` method defined elsewhere in the repo produced a graph edge claiming the first function calls the second. It doesn't — there's no textual reference between them at all.

### Bug 2 — `module.exports = { a, b, c }` misidentified as a dispatch table

The module-level dispatch-table scanner (`_scan_js_module_dispatch`, ~L4386) treats *any* top-level `object`/`array` literal as a potential handler-registry, via `_js_dispatch_value_idents`. The standard CommonJS export idiom `module.exports = { init, consolidate, stats };` has exactly that shape (an object of bare/shorthand identifiers), so **every plain CommonJS module** hits this and gets a spurious `indirect_call` edge from the file node to each of its own exported functions — misrepresenting an export list as a runtime dispatch table.

### Fix

Two changes, both scoped to `graphify/extractors/engine.py` (see attached diff):

1. In `_js_local_bound_names.walk()`, add a branch for `for_in_statement` that pulls identifiers out of its `left` field via the existing `_js_collect_pattern_idents` helper (handles both destructured and bare-identifier loop vars).
2. In `_scan_js_module_dispatch`, detect when an `assignment_expression`'s LHS is the **whole** exports surface (`module.exports` or bare `exports` — not `module.exports.NAME` / `exports.NAME`, which can legitimately hold a real dispatch table under a named property) and skip dispatch-scanning that object directly, while still walking its children so a genuine nested table under a named property is still found.

### Verification

- Re-ran `extract()` on the four files that surfaced this in a real repo: `indirect_call` edges went from 9 (all false positives) to 0, with **zero change** in node count and only those 9 phantom edges removed from the edge count.
- Added 5 regression tests to `tests/test_indirect_dispatch.py` (for-of destructured var, for-in bare var, `module.exports = {...}` exemption, bare `exports = {...}` exemption, and a control test confirming `module.exports.NAME = {...}` / named sub-property assignment still resolves as a real dispatch table — the exemption is intentionally narrow).
- `tests/test_indirect_dispatch.py`: 31/31 pass with the patch.
- Broader JS/TS/dispatch/export-related subset across the whole suite: 374 passed, 4 skipped (pre-existing skips, unrelated).
- Full suite: 35 pre-existing failures (in `test_obsidian_filename_cap`, `test_ollama_retry_cap`, `test_skillgen`, `test_watch`) reproduce identically with the patch reverted (confirmed via `git stash`) — unrelated to this change, likely Windows/environment-specific.

---

## Files in this folder

- `graphify-js-indirect-call-fix.diff` — the patch (engine.py fix + new tests), apply with `git apply` from repo root.
- `engine_patched.py` — the full post-patch `graphify/extractors/engine.py` for reference/diffing.

### DIFF
--- a/graphify/extractors/engine.py
+++ b/graphify/extractors/engine.py
@@ -1156,6 +1156,15 @@ def _js_local_bound_names(func_node, source: bytes) -> set[str]:
                 name = c.child_by_field_name("name")
                 if name is not None:
                     _js_collect_pattern_idents(name, source, bound)
+            elif c.type == "for_in_statement":
+                # `for (const [a, b] of x)` / `for (const k in obj)`: the loop
+                # variable(s) sit in the `left` field directly, never wrapped in
+                # a variable_declarator, so the branch above misses them. A bare
+                # read of the loop var later on (`if (count >= 3)`) was then
+                # mistaken for a reference to a same-named module function.
+                left = c.child_by_field_name("left")
+                if left is not None:
+                    _js_collect_pattern_idents(left, source, bound)
             walk(c)
 
     body = func_node.child_by_field_name("body")
@@ -4374,9 +4383,48 @@ def _extract_generic(
     elif config.ts_module in ("tree_sitter_javascript", "tree_sitter_typescript"):
         js_module_bound = _js_module_bound_names(root, source)
 
+        def _is_whole_exports_target(node) -> bool:
+            # `module.exports` or bare `exports` — the LHS that names the
+            # module's ENTIRE export surface, as opposed to `exports.NAME` /
+            # `module.exports.NAME` which assigns one named export and could
+            # legitimately hold a real dispatch/handler table.
+            if node.type == "identifier":
+                return _read_text(node, source) == "exports"
+            if node.type == "member_expression":
+                obj = node.child_by_field_name("object")
+                prop = node.child_by_field_name("property")
+                if obj is None or prop is None:
+                    return False
+                return (
+                    _read_text(obj, source) == "module"
+                    and _read_text(prop, source) == "exports"
+                )
+            return False
+
         def _scan_js_module_dispatch(n) -> None:
             if n.type in _JS_SCOPE_BOUNDARY:
                 return  # function / class bodies are walked separately
+            if n.type == "assignment_expression":
+                left = n.child_by_field_name("left")
+                right = n.child_by_field_name("right")
+                if (
+                    left is not None and right is not None
+                    and right.type == "object"
+                    and _is_whole_exports_target(left)
+                ):
+                    # `module.exports = { init, consolidate, stats }`: this object
+                    # lists the module's own already-defined functions purely to
+                    # export them — it is not a dispatch/handler-registry table.
+                    # Every plain CommonJS module hit this and manufactured a
+                    # bogus indirect_call edge for each exported function. Skip
+                    # dispatch-scanning the object itself but keep walking its
+                    # children so a genuine nested table under a named property
+                    # (`module.exports = { routes: { get: a, post: b } }`) is
+                    # still found.
+                    _scan_js_module_dispatch(left)
+                    for c in right.children:
+                        _scan_js_module_dispatch(c)
+                    return
             if n.type in ("object", "array"):
                 for ident in _js_dispatch_value_idents(n):
                     _emit_indirect_ref(ident, file_nid, js_module_bound, "collection")
--- a/graphify/manifest_ingest.py
+++ b/graphify/manifest_ingest.py
@@ -142,29 +142,43 @@ def _parse_apm(text: str) -> dict | None:
 
 
 def _parse_apm_fallback(text: str) -> dict | None:
-    """Minimal line parser for apm.yml when PyYAML is unavailable: a top-level
-    ``name:`` plus a simple ``dependencies:`` block (list items or a name map)."""
+    """Minimal line parser for apm.yml when PyYAML is unavailable."""
     name = None
+    version = None
     deps: list[str] = []
     in_deps = False
+
     for line in text.splitlines():
         if not in_deps:
             m = re.match(r'^name:\s*["\']?([^"\'\s#]+)', line)
             if m:
                 name = m.group(1)
                 continue
-        if re.match(r'^dependencies:\s*$', line):
-            in_deps = True
-            continue
-        if in_deps:
-            dm = (re.match(r'^\s*-\s*["\']?([^"\'\s#:]+)', line)
-                  or re.match(r'^\s{2,}([A-Za-z0-9._/@-]+)\s*:', line))
+
+            m = re.match(r'^version:\s*["\']?([^"\'\s#]+)', line)
+            if m:
+                version = m.group(1)
+                continue
+
+            if re.match(r'^dependencies:\s*$', line):
+                in_deps = True
+                continue
+
+        else:
+            dm = (
+                re.match(r'^\s*-\s*["\']?([^"\'\s#:]+)', line)
+                or re.match(r'^\s{2,}([A-Za-z0-9.*/@-]+)\s*:', line)
+            )
             if dm:
                 deps.append(dm.group(1))
-            elif re.match(r'^\S', line):  # next top-level key ends the block
+            elif re.match(r'^\S', line):
                 in_deps = False
-    return {"name": name, "version": None, "deps": deps} if name else None
 
+    return {
+        "name": name,
+        "version": version,
+        "deps": deps,
+    } if name else None
 
 def _pep508_name(spec: str) -> str:
     """`requests>=2.0` -> `requests`; `pkg[extra]==1; python_version<'3.9'` -> `pkg`."""


imagineers-tyler:
Independent confirmation of the `for_in_statement` half of this PR, plus one uncovered sibling case.

## Still reproduces on the latest release

Ran against a clean checkout of **v0.9.35** (and 0.9.33 / 0.9.34 from PyPI). `_js_local_bound_names` is byte-identical across all of them — `grep -c "for_in_statement\|catch_clause" graphify/extractors/engine.py` → `0` in every version.

```js
// a.js
export function checkAll(pausedSeen, liveIds) {
  for (const k of pausedSeen.keys()) if (!liveIds.has(k)) pausedSeen.delete(k);
}
```

```js
// vendor.min.js
```

```python
from pathlib import Path
r = extract([Path('a.js'), Path('vendor.min.js')], cache_root=Path('.'))
```

| variant | `indirect_call` |
|---|---|
| v0.9.35, `for (const k of …)` | **1** — `a_checkall -> vendor_min_k`, INFERRED 0.8, `source_location L2` |
| v0.9.35, same file but `const k = 1;` (control) | 0 |
| v0.9.35 + this PR's first hunk | 0 |

Note this fires on a **non-destructuring** loop head too, not only `for (const [a, b] of x)` — the PR title says "for-of destructuring" but the fix correctly covers both, since `_js_collect_pattern_idents` handles the bare-identifier case.

## Impact on a real repo

A 1,188-node graph of a Node.js project that vendors `fullcalendar.global.min.js`:

- the bundle contributed **829 nodes (70%)** and exactly **one** edge into application code — the phantom above;
- that single false edge made `k()` the graph's top bridge (betweenness **0.393**) and pushed all ten god nodes to minified symbols (`ti`, `p()`, `ks`, …);
- two further phantoms of the same shape elsewhere in the repo: `for (const re of [/…/])`, and a `.then(fn, fn)` argument;
- after excluding the bundle: 359 nodes, real god nodes (`getApiContext()`, `logLine()`, `main()`), mean INFERRED confidence 0.61 → 0.78.

`.graphifyignore` works around it, but any corpus containing minified or generated JS hits this.

## One case this PR does not cover: `catch_clause`

`catch (e)` binds through `catch_clause.parameter`, which is likewise never a `variable_declarator`:

```js
// a.js
export function run(handlers) {
  try { boom(); } catch (k) { handlers.get(k); }
}
```

Same `vendor.min.js` as above → on v0.9.35:

```
indirect_call | a_run -> vendor_min_k | INFERRED 0.8 | source_location L2
```

Two more lines in the same `walk()`, right after the `for_in_statement` branch:

```python
            elif c.type == "catch_clause":
                param = c.child_by_field_name("parameter")
                if param is not None:
                    _js_collect_pattern_idents(param, source, bound)
```

Verified: with this added, the `catch` repro emits 0 edges and the `for-of` repro stays at 0.

Happy to open that as a separate PR if you'd rather keep this one scoped, or it can be folded in here — whichever is easier to land.

imagineers-tyler:
Follow-up: swept the JS binding forms against `v8` @ `9f25a3a` (v0.9.35) to see how much surface this PR's `for_in_statement` branch actually covers. Each form binds a name `k` inside a function and passes it on as a call argument, with a minified `vendor.min.js` supplying `function k(a){...}` as the collision target.

| form | v0.9.35 | + this PR | + this PR & #2517 |
|---|---|---|---|
| `for (const k of xs)` | leak | ok | ok |
| `for (let k in o)` | leak | ok | ok |
| `lbl: for (const k of xs)` | leak | ok | ok |
| `for await (const k of xs)` | leak | ok | ok |
| `catch (k)` | leak | leak | ok |
| `catch ({ k })` | leak | leak | ok |
| `for (let k = 0; ...)` | ok | ok | ok |
| `const k = 1` | ok | ok | ok |
| `const { k } = o` | ok | ok | ok |
| `let [k] = arr` | ok | ok | ok |
| parameter `k` | ok | ok | ok |
| arrow parameter `k` | ok | ok | ok |
| `catch` inside an untracked arrow | ok | ok | ok |
| `catch` inside a nested function decl | ok | ok | ok |
| block-scoped rebind | ok | ok | ok |

Two things worth noting for review:

**1. This branch covers more than the title suggests.** `for await (const k of xs)` and a labeled `lbl: for (const k of xs)` both parse to `for_in_statement` — the await token and the label wrapper do not change the node type, and `walk()` already recurses through `labeled_statement`. So the single `left`-field branch here closes all four of the for-family leaks, not just destructuring. Verified:

```
async function f(xs){ for await (const k of xs) g(k); }
  for_in_statement fields=kind,left,right,body

function f(xs){ lbl: for (const k of xs) g(k); }
  labeled_statement fields=body,label
    for_in_statement fields=kind,left,right,body
```

**2. With this PR plus #2517 (`catch_clause`), all 15 forms come back clean** — the node-type whitelist in `_js_local_bound_names` appears complete for value bindings inside a function body. The two branches are adjacent in the same `walk()` and do not conflict; #2517 deliberately leaves the `for_in_statement` path to this PR.

Repro script, if useful:

```python
import os, tempfile
from pathlib import Path

VENDOR = "var Lib=function(){function k(a){return a}return{k:k}}();\n"
FORMS = {
    "for (const k of xs)":  "export function run(h,xs){ for(const k of xs) h.get(k); }\n",
    "for await":            "export async function run(h,xs){ for await (const k of xs) h.get(k); }\n",
    "labeled for-of":       "export function run(h,xs){ lbl: for(const k of xs){ h.get(k); break lbl; } }\n",
    "catch (k)":            "export function run(h){ try{boom();}catch(k){ h.get(k); } }\n",
}
for name, src in FORMS.items():
    with tempfile.TemporaryDirectory() as td:
        Path(td, "vendor.min.js").write_text(VENDOR)
        Path(td, "a.js").write_text(src)
        old = os.getcwd()
        try:
            os.chdir(td)
            r = extract([Path("a.js"), Path("vendor.min.js")], cache_root=Path(".c"), parallel=False)
        finally:
            os.chdir(old)
        leak = [e for e in r["edges"]
                if e["relation"] == "indirect_call" and e["target"].endswith("_k")]
        print(("LEAK " if leak else "ok   ") + name)
```

imagineers-tyler:
Checked against the current head `7b5350d3`. Everything below was run, not inferred — the
repro script is at the end so you can reproduce the tables directly. I applied this PR's
`for_in_statement` hunk verbatim onto a pristine engine rather than testing the branch in
place, since it is based on `8cee776` while `v8` is now at `09a34ad`; that also means the
line numbers I quote are from current `v8`, not from the diff.

One process note before the substance: the diff currently touches only
`graphify/extractors/engine.py` and `graphify/manifest_ingest.py` — there are no test
changes, although the PR description says five regression tests were added to
`tests/test_indirect_dispatch.py` and that 31/31 pass. Whichever fix shape you settle on,
it is worth committing coverage for the four cases below (`const` referenced inside vs after
the loop, hoisted `var`, object-destructuring default, kind-less head), because all four pass
the existing suite either way — which is exactly why the regression in the third row went
unnoticed.

**First, a correction to my own earlier comment on this PR.** Above, I suggested folding
`catch (e)` into `_js_local_bound_names` right next to the `for_in_statement` branch. That
suggestion was wrong and I would ask you not to take it. The reviewer bot on #2517 rejected
exactly that shape a day later — a function-wide set suppresses a legitimate by-name
reference *outside* the binding's block — and I reworked #2517 onto the per-subtree
`extra_locals` channel because of it.

The same objection applies to the `for_in_statement` branch here. The fabrication it fixes
is real and I measured it independently; the issue is only which channel the binding goes into.

## The bug this PR fixes is real

Measured on 29 JS/TS corpora (6,041 files, 2,584 `indirect_call` edges on `v8`): loop-binding
fabrications account for **172 edges**, and removing them recovers nothing — strictly
subtractive, `+0/-172`. 74 of those sites were checked individually against the surrounding
source; every one was a genuine loop variable resolving to an unrelated same-named callable.

From my own project, the case that sent me here:

```js
// src/index.js:L299
for (const k of pausedSeen.keys()) if (!liveIds.has(k)) pausedSeen.delete(k);
```

`k` resolved against a vendored `fullcalendar.global.min.js` bundle's private `k`. Minified
bundles define a private function for nearly every letter, so one-letter loop variables land
constantly.

## Scope boundary — measured

`_js_local_bound_names` is function-wide, so the binding shadows the entire function rather
than the loop. Applying this PR's `for_in` hunk verbatim to a pristine `v8` engine:

| case | `v8` | `v8` + this PR | correct |
|---|---|---|---|
| `for (const k of …)`, `k` referenced **inside** the loop | EMIT | **SUPPRESSED** | SUPPRESSED — this PR's fix, works |
| `for (var k of …)`, `k` referenced **after** the loop | EMIT | **SUPPRESSED** | SUPPRESSED — `var` hoists, still the local |
| `for (const k of …)`, `k` referenced **after** the loop | EMIT | **SUPPRESSED** | **EMIT** — `const` is block-scoped |

The third row is a regression: a legitimate edge that exists at HEAD is deleted.

```js
function k(x){ return x; }
export function run(p) {
  for (const k of [1]) void k;
  p.submit(k);          // module-level k — edge exists on v8, destroyed by this PR
}
```

## Both directions matter, so `var` decides the channel

The fix is not simply "move it to `extra_locals`". That direction is wrong for `var`, which
hoists out of the loop and *is* still the local afterwards — row 2 above would break. The
head's `kind` has to pick the channel:

| head | scope | channel |
|---|---|---|
| `for (let/const x of …)` | block | `extra_locals`, loop subtree only |
| `for (var x of …)` | function | `_js_local_bound_names`, function-wide |

```python
def _js_forin_bound_names(node, source: bytes, want=("let", "const")) -> frozenset:
    """`for (const x of xs)` / `for (let [a, b] of xs)` — the loop binding.

    Covers `for await` and labeled loops: all parse as one `for_in_statement`.
    """
    if node.type != "for_in_statement":
        return frozenset()
    kind = node.child_by_field_name("kind")
    if kind is None or _read_text(kind, source) not in want:
        return frozenset()
    left = node.child_by_field_name("left")
    if left is None:
        return frozenset()
    bound: set = set()
    _js_collect_pattern_idents(left, source, bound)
    return frozenset(bound)
```

`let`/`const` fold into the subtree in `walk_calls`. #2517 adds a `catch_clause` fold through
the same `extra_locals` channel — that PR is still open, so these would need rebasing around
each other rather than landing as-written; the two branches don't otherwise interact:

```python
if node.type == "for_in_statement":
    looped = _js_forin_bound_names(node, source)
    if looped:
        extra_locals = extra_locals | looped
```

`var` goes to the function-wide set, inside `_js_local_bound_names`'s `walk`:

```python
if c.type == "for_in_statement":
    # `for (var x of xs)` hoists x to function scope — still the local AFTER the loop.
    bound.update(_js_forin_bound_names(c, source, want=("var",)))
```

With that split, all three rows above come out correct.

## Prerequisite: object-destructuring defaults

Routing loop heads through `_js_collect_pattern_idents` surfaces a pre-existing bug in that
helper. It special-cases `assignment_pattern`, but tree-sitter emits **`object_assignment_pattern`**
for the object form, so the default *value* is collected as a binding:

| case | `v8` | `v8` + this PR | correct |
|---|---|---|---|
| parameter `[ r = k ]` (array form) | EMIT | EMIT | EMIT |
| parameter `{ q = k }` (object form) | **SUPPRESSED** | **SUPPRESSED** | EMIT |
| `for (const { q = k } of xs)` | EMIT | **SUPPRESSED** | EMIT |

```js
function k(x){ return x; }
export function run({ q = k }, p) { p.submit(k); }   // only q is bound; k is the default value
```

Row 2 is already wrong on `v8` today, independent of this PR. Row 3 is this PR extending it
to loop heads. One line fixes both:

```python
if t == "assignment_pattern" or t == "object_assignment_pattern":
```

Honest caveat: this is a correctness fix, not an impact fix. Across the same 6,041 files it
changes **zero** edges — the pattern simply does not occur in that corpus.

## One judgment call, flagged rather than asserted

Because the hunk collects `left` regardless of `kind`, a kind-less head also shadows:

```js
for (k of [1]) void k;   // reassigns an existing binding, creates no new one
p.submit(k);             // v8: EMIT   ·   this PR: SUPPRESSED
```

Same for `for ([k] of xs)`. `for (obj.k of xs)` is unaffected in both. I do not think this
one is clear-cut — whether reassignment should suppress is a dataflow question rather than a
scope question — so I am flagging it for you to decide rather than calling it a defect. The
snippet above skips it (`kind is None` returns early); if you would rather keep the current
behaviour, that is a one-word change.

## Verification

These are my own measurements on my machine, so treat them as such — the tables above are the
part that matters and the repro below reproduces them in a few seconds. Method is stated so
each is re-runnable.

- 19 binding forms shadow correctly, 0 leaks: `for await`, labeled, `for-in`, array/object/
  renamed/nested/rest destructuring heads, nested loops reusing a name, loops inside untracked
  arrows and IIFEs, class/static/getter/object-literal methods, generators, loops inside catch
  clauses, `.ts`/`.tsx`.
- `227 passed, 1 skipped` — `pytest tests/test_indirect_call_*.py tests/test_indirect_dispatch*.py
  tests/test_extract.py tests/test_cross_language_call_resolution.py tests/test_node_id_canonical.py`.
- Full suite: no new failures. The failing set is identical with the change on and off (mine is
  15 pre-existing, in `test_ollama_retry_cap` and `test_skillgen`; `test_labeling`'s batching test
  is flaky for me on an unmodified checkout too).
- Disabled, the output is byte-identical to baseline: sha256 over sorted node ids + `src|dst|rel`
  triples, 52-file JS corpus → `1072 nodes / 3023 edges`, same digest on pristine `v8` and with
  the change compiled in but inactive. Enabled on that corpus: `3021 edges`.
- No measurable cost on minified input. Same corpus, which contains a 282,117-byte bundle whose
  longest line is 281,990 bytes: `0.71s / 103 MB` → `0.67s / 102 MB`.

The corpus totals (`6,041 files`, `2,584 edges`, `+0/-172`) are from a sweep over 29 JS/TS
projects here — I can't hand you those trees, so that figure is mine to stand behind rather than
something you can re-derive. The per-case tables are the part that stands on its own.

## Repro

```python
import os, pathlib, tempfile

CASES = {
    "const loop, ref AFTER  [want EMIT]":
        "function k(x){ return x; }\nexport function run(p){ for (const k of [1]) void k; p.submit(k); }\n",
    "const loop, ref INSIDE [want SUPPRESS]":
        "function k(x){ return x; }\nexport function run(p){ for (const k of [1]) { p.submit(k); } }\n",
    "var loop,   ref AFTER  [want SUPPRESS]":
        "function k(x){ return x; }\nexport function run(p){ for (var k of [1]) void k; p.submit(k); }\n",
    "param { q = k }        [want EMIT]":
        "function k(x){ return x; }\nexport function run({ q = k }, p){ p.submit(k); }\n",
}
for name, src in CASES.items():
    d = pathlib.Path(tempfile.mkdtemp()); (d / "src").mkdir()
    (d / "src" / "a.js").write_text(src)
    cwd = os.getcwd()
    try:
        os.chdir(d)
        r = extract([d / "src" / "a.js"], cache_root=d / ".c", parallel=False)
    finally:
        os.chdir(cwd)
    nid = {n["label"].rstrip("()"): n["id"] for n in r["nodes"]}
    ind = {(e["source"], e["target"]) for e in r["edges"] if e["relation"] == "indirect_call"}
    print(("EMIT      " if (nid.get("run"), nid.get("k")) in ind else "SUPPRESSED"), name)
```

Happy to open this as a separate PR against `v8` if you would rather not rework this one, or
hand it over as a patch here — whichever is less friction. Given this PR now also carries
`manifest_ingest.py` changes, a separately scoped PR for the loop-binding fix may be the
easier one to land; entirely your call.

For the record on scope: I also looked at fixing the same class of corruption at the node-id
level (`_make_id` collides — every class `constructor` in a file, and same-named methods
across classes, share one `local_bound_names` entry, so a bare assignment lets the last
definition parsed clobber the earlier one's parameters). Merging those entries is **not** a
valid fix — I tried it and it destroys genuine edges the same way a function-wide set does.
That is a separate problem and I am not proposing it here.

PR #1988, opened 2026-07-18
fix(install): emit hook exe paths with forward slashes so Git Bash does not strip them

### LINKED ISSUE #1987: PreToolUse hooks fail on Windows with "command not found", claude install writes backslash exe paths that Git Bash strips
**TL;DR:** On Windows, graphify claude install writes the PreToolUse hook commands using the executable's native backslash path (C:\Users\...\graphify.EXE). Claude Code runs hooks through Git Bash, which treats **\** as an escape character and strips them, so the path collapses to C:Usersvarn...graphify.EXE and every invocation fails with command not found. All graph guards, search and read, are silently dead on Windows.

**How I found this:**

While digging into the Grep matcher issue (#1986 ), I noticed my Claude Code terminal was quietly printing hook errors on almost every tool call. They weren't the guard choosing not to fire, they were the guard's command failing to execute at all. Once I read the actual path in the error, the backslashes were simply gone.

**Summary**

`_claude_pretooluse_hooks()` builds each hook command as `f"{exe} hook-guard ..."`, where exe is the resolved executable path. On Windows that path uses backslashes, and it lands in `.claude/settings.json` as `C:\\Users\\...\\graphify.EXE`. Claude Code on Windows executes PreToolUse hook commands through Git Bash (/usr/bin/bash, visible in the error). Bash interprets each **\** as an escape, so **\U \v \. \b** all collapse to their next character and the path is destroyed before the shell ever looks for the binary. The result is not a miss-fire, it's a hard failure on every guarded tool call.

**Evidence (real 0.9.18 session, Windows)**

```
PreToolUse:Bash hook error
Failed with non-blocking status code: /usr/bin/bash: line 1: C:Usersvarun.localbingraphify.EXE: command not found
PreToolUse:Read hook error
Failed with non-blocking status code: /usr/bin/bash: line 1: C:Usersvarun.localbingraphify.EXE: command not found
```

Note the path in the error: every backslash from `C:\Users\varun\.local\bin\graphify.EXE` is gone.

**The exact spot**

```
graphify/install.py, _claude_pretooluse_hooks():

command: f"{exe} hook-guard search"   # exe is a native Windows path: C:\Users\...\graphify.EXE

exe is stored verbatim, so on Windows the emitted settings.json is:

{ "matcher": "Bash",
  "hooks": [{ "type": "command",
              "command": "C:\\Users\\varun\\.local\\bin\\graphify.EXE hook-guard search" }] }
```

Both the search hook and the **Read|Glob** read hook use the same exe, so both die identically.

**Repro**

graphifyy 0.9.18 (uv tool install), Claude Code CLI on Windows:

1. graphify claude install  (in any dir)
2. Inspect .claude/settings.json -> hook commands use C:\\Users\\...\\graphify.EXE
3. In Claude Code, trigger any Bash or Read tool call.
4. The terminal prints: /usr/bin/bash: line 1: C:Users...graphify.EXE: command not found

**Actual**: every PreToolUse hook fails to execute on Windows; no graph guard ever runs.
**Expected**: the hook command runs and the graphify nudge fires.

Why bash, and why forward slashes fix it?

A manually corrected block using forward slashes, `C:/Users/varun/.local/bin/graphify.EXE`, runs cleanly in the same session. Windows accepts **/** as a path separator, and bash leaves **/** untouched, so a forward-slash path is correct on every platform. Backslashes only ever break Windows.

**Suggested fix**

Normalize exe to a forward-slash path before interpolating it into the hook command, e.g. Path(exe).as_posix() (or str(exe).replace("\\", "/")). One change covers both the search and read hooks since they share exe.

_Note for reproduction: this is Windows-only. On macOS/Linux the resolved path is already /-native, so graphify claude install there emits a working command and the bug will not reproduce locally. The evidence above is from a real Windows 0.9.18 install._

varuntej07:
On Windows, `graphify claude install` wrote the PreToolUse hook command as a raw backslash path (quoted only when it contained a space). Claude Code runs command-type hooks through Git Bash by default on Windows ("Defaults to `bash`, or to `powershell` on Windows when Git Bash isn't installed"), and bash strips the unquoted backslashes, so `C:\Users\me\graphify.EXE hook-guard search` becomes `C:Usersmegraphify.EXE: command not found`. Every graph guard silently fails.

Fix: normalize the resolved exe to forward slashes in `_resolve_graphify_exe()`, which feeds the Claude/CodeBuddy, Gemini, and Codex hook emitters. Forward slashes work under Git Bash, cmd.exe, and PowerShell, and the replacement is a no-op on POSIX. Added a cross-platform regression test that monkeypatches a backslash path and asserts the emitted command has none.

Closes #1987

### DIFF
--- a/graphify/install.py
+++ b/graphify/install.py
@@ -1286,23 +1286,31 @@ def _uninstall_opencode_plugin(project_dir: Path) -> None:
         config_file.write_text(json.dumps(config, indent=2), encoding="utf-8")
         print(f"  {_OPENCODE_CONFIG_PATH}  ->  plugin deregistered")
 def _resolve_graphify_exe() -> str:
-    """Return the absolute path to the graphify executable.
+    """Return the absolute path to the graphify executable, with forward slashes.
 
     Falls back to bare 'graphify' if resolution fails. Using an absolute path
     ensures the hook works in environments where the venv Scripts/ directory is
     not on PATH (e.g. VS Code Codex extension on Windows).
+
+    The path is normalized to forward slashes so it survives every shell that
+    runs the hook command. On Windows, Claude Code runs command-type hooks
+    through Git Bash by default, where an unquoted backslash is an escape
+    character: a raw ``C:\\Users\\me\\graphify.EXE`` collapses to
+    ``C:Usersmegraphify.EXE: command not found`` and the guard silently fails.
+    Forward slashes are accepted by Git Bash, cmd.exe, and PowerShell alike, and
+    ``.replace`` is a no-op on POSIX where paths already use forward slashes.
     """
     import shutil
     found = shutil.which("graphify")
-    if found:
-        return found
-    # Derive from sys.executable: same Scripts/ (Windows) or bin/ (Unix) dir
-    scripts_dir = Path(sys.executable).parent
-    for name in ("graphify.exe", "graphify"):
-        candidate = scripts_dir / name
-        if candidate.exists():
-            return str(candidate)
-    return "graphify"
+    if not found:
+        # Derive from sys.executable: same Scripts/ (Windows) or bin/ (Unix) dir
+        scripts_dir = Path(sys.executable).parent
+        for name in ("graphify.exe", "graphify"):
+            candidate = scripts_dir / name
+            if candidate.exists():
+                found = str(candidate)
+                break
+    return (found or "graphify").replace("\\", "/")
 def _install_codex_hook(project_dir: Path) -> None:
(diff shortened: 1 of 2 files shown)


safishamsi:
Landed on v8 and shipped in 0.9.20 (faa0ac2), cherry-picked to preserve your authorship. Thanks @varuntej07 — hook exe paths are normalized to forward slashes at the single choke point (_resolve_graphify_exe), so Git Bash no longer strips them; covers all emitters and preserves quoting + the --strict suffix.

PR #1998, opened 2026-07-18
Index Nix sources and accept external extraction

caniko:
## Summary

- Add structural Nix extraction to the existing code indexing path.
- Add a format-agnostic `--external-extraction` envelope for integrations that own their source parsing.
- Preserve normal Graphify building, clustering, incremental replacement, and stale-source pruning for external contributions.
- Keep source-format adapters outside the Graphify repository.

## Validation

- 168 detection tests passed.
- 3 external-ingress replacement, validation, and pruning tests passed.
- 8 focused extraction tests passed.
- `ruff check` passed for the changed Python files.

This PR contains no source-format-specific adapter; integrations provide the generic extraction envelope instead.

PR #1997, opened 2026-07-18
Fix extraction graph normalization and provider install

caniko:
This is slice 3/5 of the work previously gathered in #1787. It is independent of the Nix/ACP series.

Fixes OpenCode plugin registration, normalizes the no-cluster graph through endpoint reconciliation, and suppresses the zero-node warning for intentionally skipped JSON data files.

Validation:
- 453 focused tests passed: `tests/test_install.py`, `tests/test_watch.py`, `tests/test_extract.py`, and `tests/test_detect.py`

This can be reviewed or merged independently of the stacked Nix slices.

PR #1996, opened 2026-07-18
Add Graphify NixOS module and PostgreSQL extraction

caniko:
This is slice 2/5 of the work previously gathered in #1787, stacked on #1995.

Adds the Graphify NixOS service module, PostgreSQL extraction support, and the service/job separation fix. This branch includes slice 1 so it remains buildable; review the incremental commits after #1995.

Validation:
- `nix flake check path:. --accept-flake-config --print-build-logs`

Merge order: #1995, then this slice.

PR #1995, opened 2026-07-18
Add reproducible Nix packaging and CI

caniko:
This is slice 1/5 of the work previously gathered in #1787.

Adds the reproducible uv2nix flake, package/app/devShell outputs, hermetic pytest check, simit-scoped CI, generated pre-commit wiring, and sandbox-safe test fixtures.

Validation:
- `nix flake check path:. --accept-flake-config --print-build-logs`
- focused sandbox test coverage included in the flake check

The later slices are intentionally based on this foundation: #1787 will be superseded once the split series is available.

PR #1999, opened 2026-07-18
Route semantic extraction through generic ACP

caniko:
This is slice 5/5 of the work previously gathered in #1787, stacked on #1995, #1996, #1997, and #1998.

The branch includes the preceding slices so it is testable as a complete tree; review the final incremental commits `9b5c7a8` and `01ee9fb` after those predecessors.

Routes semantic extraction through the official ACP Python SDK and a provider-neutral adapter command, keeps `codex-cli` as a compatibility alias, removes direct `codex exec` transport, and moves model/read-only/session settings to ACP configuration.

Hardening included:
- ACP adapters receive a minimal environment by default; provider-specific variables require explicit `GRAPHIFY_CHILD_ENV_ALLOWLIST` entries.
- The deprecated `codex-cli` alias is normalized before community-labeling concurrency guards, so all ACP paths share the default serial session policy.

Validation:
- `nix flake check path:. --accept-flake-config --print-build-logs` — 3386 passed, 51 skipped; all checks passed
- 120 focused ACP, labeling, and LLM-backend tests passed
- Graph diagnostics: 11935 nodes, 20687 edges, 0 missing endpoints, 0 dangling endpoints, 0 duplicate edges

Merge order: slices #1995 and #1996, plus independent slices #1997 and #1998, then this slice.


JenZaAI:
Hi — we reviewed this ACP / `codex-cli` subscription-backed extraction path from a security angle and ran a small controlled evaluation on a documentation corpus.

First, thank you for adding this. The ACP/Codex subscription route is exactly the kind of backend needed by users who have a ChatGPT/Codex subscription but not an `OPENAI_API_KEY`.

## Security concern: inherited child environment

One concern we noticed: the ACP child-agent path appears to inherit the parent process environment:

- `graphify/acp.py` uses `os.environ.copy()` before spawning the ACP adapter process.

Because this backend runs an autonomous agent CLI over untrusted repository/corpus text, read-only filesystem mode is not a complete boundary if the child process receives unrelated env secrets. A prompt-injection payload in markdown/source could potentially cause the agent to inspect env vars or use available shell/network tooling, depending on the CLI sandbox behavior.

We locally tested a small mitigation against the earlier superseded PR: construct an allowlisted child env for ACP/Codex subprocesses, keeping only process plumbing/auth-location vars such as `HOME`, `PATH`, locale/temp vars, `NO_BROWSER`, `CODEX_HOME`, and XDG dirs, plus explicit operator additions via something like `GRAPHIFY_CHILD_ENV_ALLOWLIST`. We also added unit tests that secret-like vars such as `OPENAI_API_KEY`, `GITHUB_TOKEN`, `ANTHROPIC_API_KEY`, and `AWS_SECRET_ACCESS_KEY` are not forwarded by default.

Focused tests still passed after that local patch:

```text
uv run --extra acp pytest tests/test_acp_backend.py tests/test_llm_backends.py -q
93 passed
```

There may be a better upstream design, but I wanted to flag the inherited-env issue before this backend becomes a recommended no-API-key/subscription path.

## Evaluation note

We also ran a stricter baseline-vs-Graphify check on a documentation corpus using 5 isolated sequential workers per phase. Graphify extraction/mapping was treated as setup and excluded from worker performance.

Result: Graphify/Codex-full was usable and non-misleading as a navigation layer, but it did not improve worker-time or token efficiency on this docs-heavy corpus. Normal source search was already strong. This may be different on a code-heavy project where call/file relationships matter more.

## Smaller correctness note

Community-labeling serialization appears to guard ACP in some places; it may be worth confirming the deprecated `codex-cli` alias inherits the same serial/concurrency behavior to avoid subscription burst issues.

Happy to share the local patch/diff if useful.

—
Posted by [JenZaAI](https://github.com/JenZaAI) via [bSmart AI workflow](https://github.com/JenZaAS/bSmart), approved by [JenZAAS](https://github.com/JenZaAS).

caniko:
Good call on the framing. This PR is about using the standardized ACP path for the deprecated `codex-cli` backend, not making Graphify itself more efficient or secure. I treated the corpus evaluation as useful context rather than an acceptance criterion.

I incorporated the concrete pointers in `01ee9fb`: ACP child processes now get a minimal allowlisted environment, with `GRAPHIFY_CHILD_ENV_ALLOWLIST` for provider-specific variables, and `codex-cli` is normalized before community-labeling concurrency guards so it inherits ACP’s default serial policy. The adapter remains neutral; `CODEX_HOME` is part of the default auth-location set needed by the current Codex adapter.

Focused tests and the full flake check pass: 3386 passed, 51 skipped.

JenZaAI:
Thanks for the clear follow-up and for incorporating the allowlisted child environment and codex-cli normalization. That addresses the main security and concurrency concerns we identified. I appreciate the careful scope distinction between the ACP integration and Graphify’s own performance.

—
Posted by JenZaAI via bSmart AI workflow, approved by JenZAAS.

Issue #1989, opened 2026-07-18
`file_hash` stat-index memo omits the root anchor: same (file, root) call returns different digests depending on call order


tomipetteri-ops:
## Summary

`cache.file_hash(path, root)` computes
`SHA256(content + b"\x00" + <path relative to root, lowercased>)`, falling back
to the **absolute** path when the file is not under `root`. Its stat-index
fastpath, however, memoizes by `(absolute path, size, mtime_ns)` — **without the
root** — and returns the memoized digest for every later call:

```python
entry = _stat_index.get(abs_key)
if (entry and entry.get("hash") is not None
        and entry.get("size") == st.st_size
        and entry.get("mtime_ns") == st.st_mtime_ns):
    return entry["hash"]          # <- root of THIS call is ignored
```

So within a process (or across processes sharing a persisted
`stat-index.json`), **whichever root was used first wins**: the same
`file_hash(f, root)` call yields different digests depending on call order.
This is an independent correctness bug in the memo key; it does not need
`--out` to exist, but in practice it surfaces there:

- In an `extract --out` run, an early caller (change detection) hashes corpus
  files with a root that does not contain them → the **absolute-path fallback**
  digest gets memoized → all later cache reads/writes in the run inherit
  absolute-path-salted keys.
- Absolute-salted keys defeat the function's documented intent ("Using a
  relative path ... makes cache entries portable across machines and checkout
  directories") — the resulting cache entries are machine-specific. This may
  explain or contribute to the observation in #1012 that "File hashes are different in
  (1) and (2) for the same file", and thematically adjacent to #1964
  (absolute-path keys break portability on the manifest side).

To be clear about scope: the `--out` cache breakage itself is fully explained by
two companion reports (#1990 — checkpoints saved under the corpus instead of
`--out`, and #1991 — the final semantic-cache save silently writing 0
entries). This memo bug additionally makes key derivation order-dependent,
which obscures diagnosis and breaks key portability/stability guarantees.

## Environment

- graphifyy 0.9.18 (PyPI, `uv tool install`), Python 3.14 (uv-managed), Windows 11
- no LLM backend involved — pure `graphify.cache` behavior

## Reproduction (standalone, no LLM)

Run the script below with the graphify environment's Python. It runs two fresh
child processes with opposite call orders:

```text
[rel_first] first=8b1b7beae984 second=8b1b7beae984 memo_returned_first_for_both=True
[...]

PR #2002, opened 2026-07-18
fix(cache): anchor semantic cache writes to cache_root so --out round-trips (#1990, #1991)

### LINKED ISSUE #1990: `graphify extract --out` saves recovery checkpoints in the wrong directory
## Summary

When `graphify extract` is run with `--out <dir>`, completed semantic chunks are
saved in the source corpus instead of the requested output directory:

```text
written to: <corpus>/graphify-out/cache/semantic/
read from:  <out>/graphify-out/cache/semantic/
```

Because Graphify writes and reads from different directories, it cannot find
these recovery checkpoints later.

This has two user-visible effects:

1. Graphify creates an unwanted `graphify-out/` directory inside the analyzed
   source tree even though the user selected another output directory.
2. If extraction is interrupted, the next recovery attempt cannot reuse the
   completed chunks. Those chunks must be extracted and billed again.

The AST cache already keeps these two concepts separate: `root` identifies the
source corpus and `cache_root` selects where cache files are stored. Semantic
checkpoints do not currently have that separation. Their single `root` value is
used for both purposes, which places the files in the corpus.

## Environment

- graphifyy 0.9.18 (PyPI, `uv tool install`), Python 3.14 (uv-managed), Windows 11
- backend: claude-cli (Sonnet), but the mechanism is backend-independent
- observed identically in two separate corpora/runs (8-file corpus; 1-file sliced corpus)

## Reproduction

```bash
graphify extract <corpus> --backend claude-cli --token-budget 15000 --out <scratch>
# after a successful run:
ls <scratch>/graphify-out/cache/semantic/pf*/   # EMPTY
ls <corpus>/graphify-out/cache/semantic/pf*/    # checkpoint entries are HERE
```

To exercise recovery rather than the normal unchanged-manifest no-op, interrupt
the first build after at least one chunk checkpoint (or remove its incomplete
output state while retaining the checkpoint). The recovery run reports cache
misses and redispatches the completed work because it looks under `<scratch>`.
An ordinary unchanged incremental run may simply preserve its existing graph
without consulting the semantic cache, so it is not a valid recovery repro.

## Mechanism (0.9.18 source)

- `graphify/llm.py`, `extract_corpus_parallel` → `_checkpoint_chunk`:
  `save_semantic_cache(..., root=root, ...)` where `root` is the extract
  `root` kwarg (= `target`, wired in cli.py `corpus_kwargs = {"root": target}`).
  No `cache_root` is passed — `save_cached` therefore derives the cache
  directory from `root`, i.e. from the corpus.
- `graphify/cli.py` (extract): reader is
  `_check_semantic_cache(sem_paths_str, root=out_root, ...)` — location anchored
  to `--out`.

## Expected

Checkpoint entries land under the same cache root the reader consults
(`out_root`), and no `graphify-out/` is created inside the analyzed corpus
(same contract #1774 established for the AST cache).

## Suggested fix

Add a `cache_root` parameter to `save_semantic_cache` (mirroring
`load_cached`'s existing `cache_root`), plumb it from cli.py through
`extract_corpus_parallel` into `_checkpoint_chunk`, and pass
`cache_root=out_root` alongside `root=target` — the same split the AST path
already uses. Add a regression test that runs with an external `--out`, captures
a completed per-chunk checkpoint, and proves a recovery read finds it without
creating `<corpus>/graphify-out/`.

## Related issues

- #1747 (closed) — previously reported unwanted cache directories in the source tree
- #1774 (closed) — fixed the corresponding output-directory handling for the AST cache
- #1012 (open) — reports multiple `graphify-out/cache` directories
- #1870 / #1892 (closed) — earlier checkpoint bugs in the same function

Two companion reports cover related problems found during the same test:

- The final semantic-cache save also mishandles `--out`: it checks source paths
  relative to the output directory and consequently saves no cache entries.
  (#1991)
- The `file_hash` optimization caches one hash per file without accounting for
  the root directory used to calculate it. Hash results can therefore depend on
  which root was used first. (#1989)

### LINKED ISSUE #1991: With --out, the final semantic-cache save resolves relative source_file paths against out_root and silently writes 0 entries
## Summary

At the end of a successful extraction, cli.py persists the fresh semantic
results with `_save_semantic_cache(..., root=out_root, ...)`. Inside
`save_semantic_cache`, each result group's `source_file` (commonly a
corpus-relative path such as `report.md`) is resolved against that same `root`
— i.e. against the `--out` directory:

```python
def resolved_source_path(value):
    path = Path(value)
    if not path.is_absolute():
        path = root_path / path        # root_path = out_root here
    ...
# write loop:
p = resolved_source_path(fpath)
if p.is_file():                        # <out>/report.md does not exist -> False
    ...                                # group silently skipped, no warning
```

`<out>/report.md` does not exist, `p.is_file()` is False, and the group is
skipped **without any warning** (the out-of-scope warning only covers the
`allowed_source_files` branch, not the not-a-file branch). Net effect with
`--out` pointing outside the corpus: **the semantic cache the reader would
consult is never populated at all**. The loss becomes visible whenever semantic
files must be reconstructed or redispatched, such as recovery from an
interrupted first build or rebuilding without the prior graph/manifest. An
ordinary unchanged incremental run may instead preserve the existing graph
without consulting the cache, so it does not necessarily re-bill. Combined with
#1990 (checkpoints written to the corpus root), the semantic cache round-trip
needed for reconstruction and recovery is fully broken in the `--out`
configuration, invisibly: the successful run itself builds the graph from the
in-memory merge.

This is a write-side sibling of #826 (0.7.13: absolute-vs-relative
`source_file` mismatch broke cache *lookup*); here the wrong *anchor root*
breaks the cache *write*.

## Environment

- graphifyy 0.9.18 (PyPI, `uv tool install`), Python 3.14 (uv-managed), Windows 11
- backend: claude-cli (Sonnet); mechanism is backend-independent

## Reproduction

```bash
graphify extract <corpus> --backend claude-cli --token-budget 15000 --out <scratch>
# run succeeds, graph.json written, log shows NO cache warnings, then:
find <scratch>/graphify-out/cache/semantic -name '*.json'   # -> nothing
```

Observed identically on two corpora (8 markdown files; 1 sliced 54k-char file).
The per-file `stat-index.json` under `<scratch>` **does** get 1 entry per doc,
which makes the empty semantic cache easy to miss.

## Expected

Either the fresh entries are written under `out_root` keyed/anchored correctly
(resolve `source_file` against the corpus `target`, write under `cache_root=out_root`
— the split the AST path already uses), or at minimum a loud warning when every
group is skipped ("0 of N semantic result groups written — source_file paths do
not resolve under root").

## Suggested fix

Same plumbing as #1990: give `save_semantic_cache` separate `root`
(source-key/`source_file` anchor = corpus `target`) and `cache_root`
(cache location = `out_root`) parameters; cli.py's final save passes both.
Add a warning when a group is dropped because `resolved_source_path(...)` is
not a file. Add a separate regression test for the final-save call site: after
a successful extraction with external `--out`, assert that the out-root semantic
cache contains every dispatched document and that a fresh cache-only replay
returns hits without an LLM call. This is distinct from the interrupted-
checkpoint recovery test in #1990 even though both use the same
root/cache_root split.

## Related

- #826 (closed) — read-side variant of the same anchor mismatch (0.7.13)
- #1757 (closed) — the `allowed_source_files` guard this save already uses;
  its warning path does not cover the not-a-file skip
- Companion reports: #1990 (checkpoint location), #1989 (`file_hash` memo)

mdshzb04:
## Summary

With `graphify extract --out <dir>`, the semantic cache write and read sides disagreed on both **location** and **key anchoring**, breaking the cache round-trip in two ways:

- **Checkpoints (#1990):** `_checkpoint_chunk` called `save_semantic_cache` with only `root=target`, so per-chunk recovery checkpoints landed under `<corpus>/graphify-out/` while the reader consulted `<out>/graphify-out/`. This created an unwanted `graphify-out/` inside the analyzed source tree, and an interrupted run re-extracted (and re-billed) already-completed chunks.
- **Final save (#1991):** cli.py passed `root=out_root`, so corpus-relative `source_file` paths (e.g. `report.md`) resolved against the `--out` directory, failed `p.is_file()`, and every result group was **silently** skipped — the cache the reader would consult was never populated at all.

## Fix

Follows the `root`/`cache_root` split the AST cache already uses (#1774), as suggested in both issues:

- `save_semantic_cache` and `check_semantic_cache` gain a `cache_root` parameter mirroring `load_cached`/`save_cached`: `root` stays the source-key anchor (content-hash keys, `source_file` resolution/relativization), `cache_root` selects where the cache directory lives. Omitting it keeps `root` for both — existing callers are byte-identical.
- `extract_corpus_parallel` plumbs `cache_root` through to `_checkpoint_chunk`.
- cli.py extract passes `root=target, cache_root=out_root` at all three sites: the cache read, the checkpoint path (via `corpus_kwargs`), and the final save.
- The orphan-prune sweep's live hashes are re-anchored to `target` (with `cache_root=out_root` for the stat-index) — hash keys anchored to `out_root` would mismatch every fresh entry under `--out` and sweep the whole cache as orphaned.
- `save_semantic_cache` now emits a `RuntimeWarning` when **every** result group is dropped because its `source_file` does not resolve to a real file — surfacing the silent-0-writes failure mode #1991 asked for.

## Testing

New `tests/test_semantic_cache_out_root.py` (7 tests):

- checkpoint written with `cache_root` lands under `<out>`, not the corpus, and no `<corpus>/graphify-out/` is created (#1990)
- a checkpoint written via the `merge_existing` path is found by `check_semantic_cache` using the same `root`/`cache_root` split (recovery scenario, #1990)
- the final-save call shape (`root=corpus, cache_root=out`, corpus-relative `source_file`) writes 1 entry instead of 0 (#1991)
- the old broken call shape (`root=out_root`) now warns instead of silently writing 0 entries (#1991)
- omitting `cache_root` preserves the historical single-root behaviour
- `extract_corpus_parallel` exposes `cache_root`

Runs:

- `python -m pytest tests/test_semantic_cache_out_root.py -v` — 7 passed
- `python -m pytest tests/test_cache.py tests/test_chunking.py tests/test_partial_cache.py tests/test_extract_cli.py tests/test_incomplete_build_guard.py -q` — 134 passed, 1 skipped
- `python -m pytest tests/ -q` — 3226 passed, 167 skipped; the 13 failures (test_terraform.py, test_ollama_retry_cap.py, test_manifest_ingest.py apm, test_install_references.py wheel) are pre-existing on a clean checkout of `v8` in this environment (verified via `git stash`) and unrelated to this change
- `ruff check` on all touched files — passed

Fixes #1990
Fixes #1991

### DIFF
--- a/graphify/cache.py
+++ b/graphify/cache.py
@@ -722,6 +722,7 @@ def check_semantic_cache(
     mode: str | None = None,
     prompt: "str | Path | None" = None,
     prompt_file: "str | Path | None" = None,
+    cache_root: "Path | None" = None,
 ) -> tuple[list[dict], list[dict], list[dict], list[str]]:
     """Check semantic extraction cache for a list of absolute file paths.
 
@@ -744,6 +745,12 @@ def check_semantic_cache(
     vintage is unknowable and dropping them would re-bill a whole corpus — but
     a warning reports how many were served. Omitting ``prompt`` keeps the
     historical behavior for existing callers.
+
+    ``cache_root`` decouples *where* the cache is read from the key-anchor
+    ``root``, mirroring :func:`load_cached` and :func:`save_semantic_cache`
+    (#1774 / #1990). With ``--out``, pass the corpus as ``root`` (so content-hash
+    keys and relative-path resolution stay anchored to the source tree) and the
+    output directory as ``cache_root``. Omitting it keeps ``root`` for both.
     """
     global _legacy_semantic_hits
     kind = "semantic" if mode is None else f"semantic-{mode}"
@@ -757,7 +764,8 @@ def check_semantic_cache(
         p = Path(fpath)
         if not p.is_absolute():
             p = Path(root) / p
-        result = load_cached(p, root, kind=kind, prompt=prompt, prompt_file=prompt_file)
+        result = load_cached(p, root, kind=kind, cache_root=cache_root,
+                             prompt=prompt, prompt_file=prompt_file)
         if result is not None:
             cached_nodes.extend(result.get("nodes", []))
             cached_edges.extend(result.get("edges", []))
@@ -808,6 +816,7 @@ def save_semantic_cache(
     prompt: "str | Path | None" = None,
     prompt_file: "str | Path | None" = None,
     partial_source_files: Iterable[str | Path] | None = None,
+    cache_root: "Path | None" = None,
 ) -> int:
     """Save semantic extraction results to cache, keyed by source_file.
 
@@ -845,6 +854,16 @@ def save_semantic_cache(
     replaying them (#1939). Pass the same prompt here as to
     :func:`check_semantic_cache`, or the write lands in a namespace the next
     read won't consult.
+
+    ``cache_root`` decouples *where* the cache directory is written from the
+    source-key anchor ``root`` — mirroring the same split that :func:`load_cached`
+    and :func:`save_cached` already expose (#1774). When given, cache files land
+    under ``cache_root`` while ``source_file`` paths are still resolved and
+    relativized against ``root``. When omitted, ``root`` is used for both
+    purposes (unchanged behaviour for existing callers). This fixes checkpoints
+    and the final save going to the corpus tree instead of ``--out`` (#1990,
+    #1991).
+
     Returns the number of files cached.
     """
     from collections import defaultdict
@@ -952,6 +971,7 @@ def save_semantic_cache(
                 ]
 
     saved = 0
+    skipped_not_file = 0
     for fpath, result in by_file.items():
         p = resolved_source_path(fpath)
         if p.is_file():
@@ -975,9 +995,9 @@ def save_semantic_cache(
                 # markers ride through, so is_partial below re-detects it) rather
                 # than a later clean slice silently replacing it and promoting the
                 # half-file to complete.
-                prev = load_cached(p, root, kind=kind, prompt=prompt,
-                                   prompt_file=prompt_file, allow_legacy=False,
-                                   allow_partial=True)
+                prev = load_cached(p, root, kind=kind, cache_root=cache_root,
+                                   prompt=prompt, prompt_file=prompt_file,
+                                   allow_legacy=False, allow_partial=True)
                 _prev_partial = bool(prev.get("partial")) if prev else False
                 if prev:
                     result = {
@@ -1002,6 +1022,19 @@ def save_semantic_cache(
             )
             if is_partial:
                 result = {**result, "partial": True}
-            save_cached(p, result, root, kind=kind, prompt=prompt, prompt_file=prompt_file)
+            save_cached(p, result, root, kind=kind, cache_root=cache_root,
+                        prompt=prompt, prompt_file=prompt_file)
             saved += 1
+        else:
+            skipped_not_file += 1
+    if skipped_not_file and skipped_not_file == len(by_file):
+        warnings.warn(
+            f"save_semantic_cache: all {skipped_not_file} source_file group(s) were "
+            "skipped because their paths do not resolve to real files. This usually "
+            "means ``root`` is anchored to the wrong directory (e.g. the --out "
+            "directory instead of the corpus root). Pass the corpus directory as "
+            "``root`` and the output directory as ``cache_root`` (#1991).",
+            RuntimeWarning,
+            stacklevel=2,
+        )
     return saved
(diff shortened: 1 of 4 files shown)


safishamsi:
Landed on v8 and shipped in 0.9.20 (0816630), cherry-picked to preserve your authorship. Thanks @mdshzb04 — semantic cache entries now key on the scan root while the cache directory sits at cache_root/out_root, so --out round-trips (final save writes entries, checkpoints land in the right dir). Composes with the #1989 salt-keying and #1939 fingerprint namespacing.

Issue #1840, opened 2026-07-13
`hook-guard`: Read/Glob PreToolUse nudge has no project-membership or graph-freshness gate — fires identically on out-of-project files and stale graphs


chaosanderos:
**Version:** reproduced on 0.9.10 through 0.9.13 (current `v8` HEAD); `_run_hook_guard` is unchanged across that range.

**Problem:** `graphify/__main__.py::_run_hook_guard` (`read` branch, ~line 1676 on 0.9.10; moved to `graphify/cli.py` on `v8`) fires `_READ_NUDGE` whenever (a) the tool call's `file_path`/`pattern`/`path` has a trailing extension in `_HOOK_SOURCE_EXTS` (which includes `.md`/`.rst`/`.txt`/`.mdx`), (b) the path isn't under `graphify-out/`, and (c) `graphify-out/graph.json` exists in the cwd. Two real-world consequences observed in production use (Claude Code, 2026-07 audit sessions):

1. Reading a file with **no relationship to the project the graph indexes** — a global plugin-cache `SKILL.md`, another repo's config — fires the same "MANDATORY: run graphify before reading source files" `additionalContext` as reading the project's own `.py` source, because the guard only inspects the path string's extension, never whether the path is plausibly inside the tree `graphify-out/graph.json` was built from.
2. A graph that is arbitrarily stale (no cron/CI rebuild; `graphify update` is run by hand) gates identically to a freshly-rebuilt one — the guard calls only `.is_file()` on `graph.json`, never compares its age against the tracked tree.

Both make the nudge fire on a majority of reads in a typical audit/ops/docs-heavy session (measured: ~100% of reads were on non-application-code targets in one such session), training the agent to habitually override "MANDATORY" language — which also blunts the signal on the minority of reads where the nudge is actually correct.

**Proposed smallest fix:**

1. Resolve the candidate path and skip the nudge when it isn't under the graph's project root (reuse the existing `.graphify_root`/`GRAPHIFY_REPO_ROOT` root-resolution convention already used by `_rebuild_code` and friends elsewhere in `__main__.py`/`build.py`).
2. Read the `generated_at` stamp that PR #1624 already proposes adding to `graph.json`, and soften the nudge language (e.g. drop "MANDATORY", note staleness) when the stamp is older than the newest mtime in the tracked corpus, or than `git status`'s latest relevant change — reusing #1624's plumbing instead of adding a second staleness mechanism.
[...]

PR #2003, opened 2026-07-18
fix(install): fire the search nudge on Claude Code's Grep tool, not just Bash (#1986)

### LINKED ISSUE #1986: Grep searches never trigger the graphify-first nudge in Claude Code (CLI): PreToolUse search guard only matches Bash
**TL;DR:** graphify claude install registers the PreToolUse "search" guard with matcher "Bash" only, but current Claude Code searches code content through its dedicated Grep tool, not Bash. So the guard never fires on Grep, and the agent greps the codebase without ever being nudged to run graphify query first, silently bypassing the knowledge graph.

**How I found this**

Whenever I use Codex, I watch it reach for graphify on its own before it starts exploring. With Claude Code I never once saw it do that, even though everything was installed and configured correctly. That gap is what sent me digging, and it turned out not to be a Claude quirk but a hole in the guard that graphify installs.

**Summary**

The Claude Code search-nudge hook was changed from matcher **Glob|Grep** to Bash in ee1df22 ("Fix PreToolUse hook for Claude Code v2.1.117+"), on the stated premise that "Grep and Glob tools removed in CC v2.1.117; searches now go through Bash." That premise does not hold for current Claude Code. Grep and Glob are still first-class tools, and _content search is routed to the dedicated Grep tool, not through Bash._

The result is a regression: the search guard matches a tool the agent no longer uses for search, and never fires on the tool it does use, so the graphify-first nudge is silently dead for the most common exploration path.

**_I am Claude Code in this session. I have separate tool schemas for Bash, Grep, Glob, and Read, and I invoked the Grep tool repeatedly during this very investigation. It ran ripgrep and returned matches without ever going through Bash. My Bash tool's own description even says: "IMPORTANT: Avoid using this tool to run find, grep... use the appropriate dedicated tool."_**                  - Claude Code

So current Claude Code actively routes content search to Grep and away from Bash, the exact opposite of the assumption behind the "Bash"-only matcher.

The exact line

```
graphify/install.py:300-305, _claude_pretooluse_hooks():

return [
    {"matcher": "Bash",
     "hooks": [{"type": "command", "command": f"{exe} hook-guard search"}]},
    {"matcher": "Read|Glob",
     "hooks": [{"type": "command", "command": f"{exe} hook-guard read"}]},
]

```

Claude Code's matcher is a regex tested against the tool name. 

"Bash" matches the tool Bash and does not match Grep. Grep appears in neither matcher, so a Grep call is matched by no graphify hook and the guard never runs. The read guard is fine. Glob is wired into **"Read|Glob"** and a pattern like ****/*.py** fires the read nudge, so among the native tools, only Grep is orphaned.

**Repro**

```
graphifyy 0.9.18 (uv tool install), any Claude Code with the dedicated Grep tool:

1. graphify . (or any repo with graphify-out/graph.json)
2. graphify claude install
3. Inspect .claude/settings.json. The search hook is {"matcher": "Bash", ...}, no Grep.
4. In Claude Code (CLI), ask a question that triggers a content search. The agent uses the Grep tool, no additionalContext nudge is injected, and graphify query is never suggested.
```

**Actual**: Grep calls bypass the guard entirely and the graph is silently skipped.
**Expected**: a content search via Grep gets the same "run graphify query first" nudge that a Bash rg/grep gets.

**Suggested fix**

This is not a clean one-liner, and I want to flag that up front so the PR actually fires rather than just widening the matcher:

1. Matcher (necessary, not sufficient), install.py:301: "matcher": "Bash" becomes "matcher": "Bash|Grep".
2. Guard body (also required). The kind == "search" branch at graphify/cli.py:130-135 only inspects t.get("command", ""). The Grep tool's tool_input has no command, it carries pattern, path, and glob. So with only the matcher change, Grep calls reach the guard, read an empty command, and still never nudge. The branch needs to treat a Grep-shaped input (has pattern, no command) as a search.
3. Idempotency. Add the new matcher string to the four dedup tuples (install.py:1630, 1653, 1804, 1818), or the fix itself will append a duplicate block on upgrade instead of replacing in place.

PR should fold a regression test that feeds a Grep-shaped {"tool_input": {"pattern": "...", "path": "."}} into hook-guard search and asserts the nudge. tests/test_search_hook.py currently only exercises the Bash command shape, which is why this slipped through.

mdshzb04:
## Summary

The Claude Code search-guard matcher was narrowed from `Glob|Grep` to `Bash` in ee1df22 on the premise that dedicated search tools were removed in CC v2.1.117+ and searches go through Bash. That premise doesn't hold for current Claude Code: content search is routed through the first-class **Grep** tool (Bash's own tool description steers the agent away from shell grep), so the graphify-first nudge never fired on the agent's primary exploration path and the graph was silently bypassed.

## Fix — the three parts the issue called out

1. **Matcher** (`graphify/install.py`, `_claude_pretooluse_hooks`): `"Bash"` → `"Bash|Grep"`. Glob already fires the read nudge via `"Read|Glob"`, so Grep was the only orphaned tool.
2. **Guard body** (`graphify/cli.py`, `_run_hook_guard` search branch): the guard only inspected `tool_input.command`, which a Grep call doesn't carry (it has `pattern`/`path`/`glob`). With the matcher change alone, Grep calls would reach the guard, read an empty command, and still never nudge. A Grep-shaped input (`pattern` present, no `command`) is now treated as a search — it *is* one by definition — and nudges whenever a fresh graph exists. The Bash token-matching path is byte-identical, and the Grep shape requires an *absent* `command`, so a non-search Bash call carrying a stray `pattern` key stays silent (covered by a test).
3. **Idempotency**: `"Bash|Grep"` added to the four install/uninstall dedup filter tuples (claude install/uninstall + codebuddy install/uninstall), so upgrading replaces the stale `"Bash"` hook in place instead of appending a duplicate. Verified by seeding a pre-fix `settings.json` with the old `"Bash"` hook and running the installer: exactly 2 graphify hooks remain (`Bash|Grep` + `Read|Glob`), no dupes.

The nudge stays nudge-only (never `permissionDecision`), preserving the strict-mode contract that search is never blocked.

## Testing

New regression tests in `tests/test_search_hook.py` feed Grep-shaped `{"tool_input": {"pattern": ..., "path": ...}}` through `hook-guard search` (the exact gap the issue flagged — the file previously only exercised the Bash `command` shape):

- Grep-shaped inputs nudge when a graph exists (pattern-only, pattern+path, pattern+glob variants)
- silent without a graph
- nudge payload is valid PreToolUse JSON
- never blocks (no `permissionDecision`/`deny`)
- a non-search Bash command with a stray `pattern` key does **not** nudge

Existing matcher assertions updated in `test_search_hook.py`, `test_install.py`, `test_claude_md.py`, `test_codebuddy.py`, `test_hook_strict.py`.

Runs:

- `python -m pytest tests/test_search_hook.py tests/test_hook_guard.py tests/test_hooks.py tests/test_hook_strict.py tests/test_read_hook.py tests/test_codebuddy.py tests/test_claude_md.py tests/test_install.py tests/test_install_upgrade.py tests/test_gemini_hook.py -q` — 397 passed
- `python -m pytest tests/ -q` — 3224 passed, 167 skipped; the 13 failures (terraform grammar, ollama SDK, apm manifest, wheel build) are pre-existing on a clean `v8` checkout in this environment
- `ruff check` on touched files — passed

Fixes #1986

### DIFF
--- a/graphify/cli.py
+++ b/graphify/cli.py
@@ -444,11 +444,18 @@ def _run_hook_guard(kind: str, strict: bool = False) -> None:
     try:
         if kind == "search":
             cmd_str = str(t.get("command", "") or "")
-            # Same set the old `case` matched: *grep*, *ripgrep*, and rg/find/fd/
+            # Two input shapes reach this guard (matcher "Bash|Grep", #1986):
+            # the Bash tool carries `command`, while Claude Code's dedicated
+            # Grep tool carries `pattern` (plus optional path/glob) and no
+            # command — a Grep call IS a content search by definition, so it
+            # nudges whenever a graph exists. For Bash, keep matching the same
+            # set the old `case` matched: *grep*, *ripgrep*, and rg/find/fd/
             # ack/ag as a token (name followed by a space). Nudge-only, even in
             # strict mode — see the docstring.
-            if any(tok in cmd_str for tok in ("grep", "ripgrep", "rg ", "find ", "fd ", "ack ", "ag ")) \
-                    and out_path("graph.json").is_file():
+            is_grep_tool = not cmd_str and bool(t.get("pattern"))
+            is_bash_search = any(tok in cmd_str for tok in (
+                "grep", "ripgrep", "rg ", "find ", "fd ", "ack ", "ag "))
+            if (is_grep_tool or is_bash_search) and out_path("graph.json").is_file():
                 sys.stdout.write(_SEARCH_NUDGE)
         elif kind == "read":
             vals = [str(t.get("file_path") or ""), str(t.get("pattern") or ""), str(t.get("path") or "")]
--- a/graphify/install.py
+++ b/graphify/install.py
@@ -290,9 +290,12 @@ def _claude_pretooluse_hooks(strict: bool = False) -> "list[dict]":
 
     The command invokes `graphify hook-guard <search|read>` via the absolute exe
     path (`_resolve_graphify_exe`), so it parses under sh, cmd.exe and PowerShell
-    alike — this is the #522 fix, and mirrors the codex hook. Matchers stay "Bash"
-    and "Read|Glob" and the command always contains "graphify", so the existing
-    install/uninstall filters find and replace both old bash hooks and these.
+    alike — this is the #522 fix, and mirrors the codex hook. Matchers are
+    "Bash|Grep" and "Read|Glob" and the command always contains "graphify", so the
+    existing install/uninstall filters find and replace both old bash hooks and
+    these. "Grep" is in the search matcher because current Claude Code routes
+    content search through its dedicated Grep tool, not Bash (#1986) — a
+    Bash-only matcher never fired on the agent's primary search path.
 
     When ``strict`` is set, the read hook carries ``--strict`` so it blocks the
     first raw read per session (Claude Code only). The ``GRAPHIFY_HOOK_STRICT`` env
@@ -303,7 +306,7 @@ def _claude_pretooluse_hooks(strict: bool = False) -> "list[dict]":
         exe = f'"{exe}"'
     read_cmd = f"{exe} hook-guard read" + (" --strict" if strict else "")
     return [
-        {"matcher": "Bash",
+        {"matcher": "Bash|Grep",
          "hooks": [{"type": "command", "command": f"{exe} hook-guard search"}]},
         {"matcher": "Read|Glob",
          "hooks": [{"type": "command", "command": read_cmd}]},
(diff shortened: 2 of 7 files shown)


safishamsi:
Landed on v8 and shipped in 0.9.20 (0224bca), cherry-picked to preserve your authorship. Thanks @mdshzb04 — the PreToolUse search matcher is now Bash|Grep and the guard recognizes Claude Code Grep tool_input, so grep-via-Grep gets the graphify-first nudge. Stays nudge-only; composes with the strict hook and #1840 gating.

Issue #1744, opened 2026-07-09
# Bug: Same-named classes across modules silently overwrite each other in the graph


aviciot:
**Graphify version**: 0.9.9

---

Hey team ðŸ‘‹

We're running Graphify on a large multi-module Maven project (15 modules, ~700 Java files) and hit a frustrating silent bug,

## What happened

We have two completely separate classes both named `FinancialEntryValidator` â€” one in `payment-adaptor` and one in `core-adaptor`. Different packages, different logic, different purpose.

```
my-project/
  payment-adaptor/src/main/java/com/example/payment/service/FinancialEntryValidator.java
  core-adaptor/src/main/java/com/example/core/service/FinancialEntryValidator.java
```

When Graphify builds the graph, both get the same short node ID (`financialentryvalidator`). Since `G.add_node()` is idempotent, whichever gets processed last silently overwrites the other. The losing class completely disappears from the graph â€” no node, no edges, nothing.

There's no warning, no error, no indication anything went wrong.
 The graph looks complete.- But when we asked our LLM "does the payment module validate currency against ISO 4217?" it answered confidently: **no validation exists** :(

The real answer is the opposite â€” the check is right there in the class that got overwritten:

```java
// payment-adaptor - FinancialEntryValidator.java
private static final Set<String> ISO_CURRENCIES =
    Currency.getAvailableCurrencies().stream()
        .map(Currency::getCurrencyCode)
        .collect(Collectors.toSet());

if (!ISO_CURRENCIES.contains(currency)) {
    validationErrors.add("Currency [" + currency + "] is invalid ISO 4217 alpha-3 code");
}
```

Silent wrong answers are worse than no answers.

## Minimal reproduction

```python
import networkx as nx

G = nx.DiGraph()

# Module A processes first â€” real node with full source_file
G.add_node("financialentryvalidator",
    label="FinancialEntryValidator",
    source_file="payment-adaptor/.../FinancialEntryValidator.java"
)

# Module B processes second â€” stub node overwrites the real one
G.add_node("financialentryvalidator",
    label="FinancialEntryValidator",
    source_file=""   # stub â€” empty
)

# Real node is gone
print(G.nodes["financialentryvalidator"]["source_file"])  # "" â€” overwritten
```

## What we did as a workaround

We added a post-processing step after `build_merge` that detects stub nodes (empty `source_file`) colliding with real nodes and redirects their edges to the correct real node. It resolved about half of the stubs in one repo . 
[...]

Issue #15, opened 2026-04-06
Is this open source?


ShaggyDude:
Is this open source?
Thanks,

safishamsi:
yeap!

PR #2016, opened 2026-07-19
Emit calls edges for top-level function calls, with file node as caller

### LINKED ISSUE #1972: Calls and constant reads in top-level / script context never get a caller node, so they emit no edges (uniform across languages)
**Setup:** 0.9.17, uv, macOS, same Rails repo as #1634 / #1640 / #1666 / #1784.

First: the #1784 fix works exactly as specified. All 44 of our tracked `.rake` files now extract (81 nodes), helper classes and methods inside them get nodes, and their cross-file calls resolve. Thank you for the fast turnaround, and for applying the full seven-gate map.

Re-running our recall grading after the upgrade surfaced the next layer down. It is not a Ruby routing bug (it behaves identically in every language we probed), so filing it separately as a design/feature request.

**The property:** only calls made inside a definition context (a method, function, or class body) get a caller node. Code at the top level of a file, or inside a DSL block, has no enclosing definition, so its calls and constant references attach to nothing: no edge is emitted at all. The file node exists (and carries `contains` edges to any definitions), but nothing binds the file's top-level behavior to the things it uses.

**Repro** (five files, three languages):

```ruby
# widget.rb
class Widget
  def self.tally
    42
  end
end

# incontext.rb  (control: same call, inside a method)
class Runner
  def self.go
    Widget.tally
  end
end

# toplevel.rb
Widget.tally

# blockcall.rb  (the Rake shape: plain Ruby, call inside a DSL block)
namespace :ops do
  task fix: :environment do
    Widget.tally
  end
end
```

Plus the same top-level pattern in Python (`import widget; widget.tally()`) and JS (`import { tally } from './widget.js'; tally()`).

`graphify extract . --code-only` yields 14 nodes, 7 edges. The control resolves perfectly: `incontext_runner_go -[calls]-> widget_widget_tally`. The other four callers (`toplevel.rb`, `blockcall.rb`, `toplevel.py`, `toplevel.js`) each get a file node and zero outgoing edges. The call site is byte-identical to the control; only the missing enclosing definition differs.

**Why it matters on a real repo:** rake task bodies are almost entirely `task do ... end` blocks. In a 13-entity blast-radius grading on our Rails repo (`affected --depth 1` against comment-filtered grep ground truth, same method as #1784), 79 true dependents were rake files. Exactly 9 of them have edges today, the ones whose reference happens to sit inside a helper method: 11.4% recall, against 95.8% on `.rb` files. The same shape covers Rails initializers, `db/seeds.rb`, migration-adjacent backfill scripts, Python module-level wiring, and JS entry points. These are precisely the files a blast-radius query most needs to surface: ops and data-repair scripts that break silently when a class they use changes.

**Proposal:** when a call or constant reference has no enclosing definition, bind it to the FILE node as the caller. The file node already exists and already carries `contains`, so this adds outgoing `calls` / `references` edges to a node that is already the natural home for script-level behavior, and `affected` becomes complete for script-shaped code. Two softer landings if the full version feels too aggressive:

- gate it behind a flag or config key, since it changes what `affected` returns on every corpus (in our view for the better, but it is a semantics change), or
- start Ruby-only, where the pain concentrates (the Rake DSL, initializers, seeds), and extend later.

Happy to re-run our census + recall grading on whichever release this lands in, as before.

Rishet11:
Fixes #1972

## Problem

Calls made at a file's top level (outside any def) have no caller node, so they emit no `calls` edges. A function invoked only from module scope looks dead in the graph.

## Fix

In `graphify/extractors/engine.py`, after the per-definition walk, walk the file root once more with the file node as caller. `walk_calls` already returns at every definition boundary, so nothing already covered gets re-emitted.

Two scoping decisions:

- A `_toplevel_calls_only` flag limits the extra walk to direct `calls`. Indirect dispatch and PHP-family references have their own passes; emitting them here would duplicate work with empty shadow context.
- Java is skipped. It has no top-level statements, field initializers are walked elsewhere, and a root walk would re-create the phantom stubs #1744 removed.

## Test

3 new tests: top-level Python call, top-level JS call, and an in-function regression guard. Full suite shows no new failures against the v8 baseline.

One known behavior to flag: class-body statements now get attributed to the file (relevant for Swift field initializers). No test covers that yet, so I left it alone for now.

Still fairly new to this codebase, so if I have the approach or the scoping wrong anywhere, tell me and I will fix it up.

### DIFF
--- a/graphify/extractors/engine.py
+++ b/graphify/extractors/engine.py
@@ -3840,6 +3840,12 @@ def _extract_generic(
         for body_id, (method_node, class_nid) in csharp_method_scopes.items()
     }
 
+    # #1972: during the top-level root walk, emit ONLY direct `calls` edges.
+    # Module-level indirect dispatch already has a dedicated pass below with
+    # correct module-scope shadow filtering, so a root-walk emission would be a
+    # duplicate carrying wrong (empty) shadow context.
+    _toplevel_calls_only = False
+
     def _emit_indirect_by_name(ident_name: str, loc_node, scope_nid: str,
                                context: str) -> None:
         """Resolve a name that is referenced AS A VALUE to a real callable def and emit
@@ -3851,6 +3857,8 @@ def _extract_generic(
         string names an ATTRIBUTE and is never shadowed by a local, so that path passes
         the name straight through. ``loc_node`` supplies the source line.
         """
+        if _toplevel_calls_only:
+            return
         ref_nid = label_to_nid.get(ident_name)
         # Defer to the cross-file resolver when the name is not defined in this file
         # (`from .h import fn`), or resolves to an import-surfaced FOREIGN symbol whose
@@ -4401,7 +4409,7 @@ def _extract_generic(
                             _emit_indirect_ref(arg, caller_nid, enclosing_locals, "argument")
 
             # Helper function calls: config('foo.bar') → uses_config edge to "foo"
-            if (callee_name and callee_name in config.helper_fn_names):
+            if (not _toplevel_calls_only and callee_name and callee_name in config.helper_fn_names):
                 args_node = node.child_by_field_name("arguments")
                 first_key: str | None = None
                 if args_node:
@@ -4439,7 +4447,8 @@ def _extract_generic(
                             })
 
             # Service container bindings: $this->app->bind(Foo::class, Bar::class)
-            if (node.type == "member_call_expression"
+            if (not _toplevel_calls_only
+                    and node.type == "member_call_expression"
                     and callee_name
                     and callee_name in config.container_bind_methods):
                 args_node = node.child_by_field_name("arguments")
@@ -4477,7 +4486,7 @@ def _extract_generic(
                             })
 
         # Static property access: Foo::$bar → uses_static_prop edge
-        if node.type in config.static_prop_types:
+        if node.type in config.static_prop_types and not _toplevel_calls_only:
             scope_node = node.child_by_field_name("scope")
             if scope_node is None:
                 for child in node.children:
@@ -4504,7 +4513,8 @@ def _extract_generic(
                         })
 
         # PHP class constant access: Foo::BAR → references_constant edge
-        if config.ts_module == "tree_sitter_php" and node.type == "class_constant_access_expression":
+        if (config.ts_module == "tree_sitter_php" and not _toplevel_calls_only
+                and node.type == "class_constant_access_expression"):
             class_name = _php_class_const_scope(node)
             if class_name:
                 tgt_nid = label_to_nid_ci.get(class_name.lower())
@@ -4594,6 +4604,26 @@ def _extract_generic(
             caller_nid,
             receiver_types_by_body.get(id(body_node)),
         )
+
+    # Top-level / script-context calls have no enclosing definition, so they
+    # never entered function_bodies and got no caller at all (#1972). walk_calls
+    # already returns without descending at every config.function_boundary_types
+    # node, so walking from the file's root with file_nid as caller picks up
+    # calls outside every tracked def — it cannot double-emit anything already
+    # walked via the function_bodies loop above. During this walk we emit ONLY
+    # direct `calls` (via _toplevel_calls_only), because module-level indirect
+    # dispatch and references have their own dedicated passes.
+    #
+    # Java is skipped: it has no top-level executable statements, its field
+    # initializers are already walked via initializer_nodes with the owning
+    # class as caller, and a root walk would only defer them as file-attributed
+    # raw_calls that resurrect the ambiguous phantom stubs #1744 removes.
+    if config.ts_module != "tree_sitter_java":
+        _toplevel_calls_only = True
+        try:
+            walk_calls(root, file_nid, java_receiver_types.get(id(root)))
+        finally:
+            _toplevel_calls_only = False
 
     # #1356: walk property/field initializers (collected above). walk_calls
     # self-guards against re-entering function bodies and dedups via
--- a/graphify/extractors/engine.py
+++ b/graphify/extractors/engine.py
@@ -4617,7 +4617,7 @@ def _extract_generic(
     # Java is skipped: it has no top-level executable statements, its field
     # initializers are already walked via initializer_nodes with the owning
     # class as caller, and a root walk would only defer them as file-attributed
-    # raw_calls that resurrect the ambiguous phantom stubs #1744 removes.
+    # raw_calls that resurrect the ambiguous phantom stubs #1744 removed.
     if config.ts_module != "tree_sitter_java":
         _toplevel_calls_only = True
         try:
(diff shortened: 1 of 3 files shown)


Rishet11 (commented):
fixed the required changes

krishnateja7:
@Rishet11 Thanks for picking this up, this is exactly the shape we hoped for: file node as caller, with the walk returning at definition boundaries so nothing double-emits.

One coverage ask from the corpus that motivated #1972: our worst case is Ruby rake files, where every call lives either at the file root or inside a `task :name do ... end` block. Those blocks are not definitions, so the root walk should attribute their inner calls to the file node, but there is no Ruby test in the PR to lock that in. Could you add two: a plain top-level Ruby call, and a call inside a method-with-block (the rake-task shape)? `.rake` extraction landed in 0.9.13 and this fix is its main consumer.

On scoping: agreed that limiting the extra walk to direct calls is right for this PR. For Ruby, constant reads are missing everywhere, not just at top level, so that half of #1972 is bigger than this change anyway. Fine by us if it stays tracked on the issue after this merges.

Offer: we run a recall census against a real Rails corpus (same one behind our earlier graphify reports). Happy to run it against a build of your branch and report before/after edge counts if useful.

Rishet11:
Added both tests: test_ruby_toplevel_call_gets_file_caller and test_ruby_rake_task_block_call_gets_file_caller (the task :name do ... end shape), each asserting the file node is the caller, and both are green...
 They lock in the enclosing-definition-context property from #1972 for Ruby top-level and DSL-block call sites.

Rishet11:
@krishnateja7 gentle bump on this one. The two Ruby tests you asked for are pushed and green (top-level call plus the `task :name do ... end` block shape), so it should be ready for another look whenever you have time. No rush.

krishnateja7:
@Rishet11 Took another look, and thanks for adding both Ruby tests so quickly. Before signing off I ran the census I offered: your branch (head `0e561e4`) against its merge-base (`edec9ea`), both installed from git, each run with `graphify update .` on the same real Rails corpus (~1,800 tracked files, `PYTHONHASHSEED=0`).

Census result:

| build | edges | `calls` edges | `calls` with a file-node caller |
|---|---|---|---|
| merge-base | 28,178 | 6,898 | 0 |
| PR branch | 28,211 | 6,916 | 18 (15 `.rb`, 2 `.ts`, 1 `.rake`) |

For the rake case that motivated #1972, this corpus has ~195 file-level call sites we track out-of-band today; the PR surfaced 1 of them. That sent me to a minimal repro, and I think the real issue is upstream of the tests:

**The new edges do not survive the CLI pipeline, even for the exact content of your green test.** Clean directory, one file, the same content as `test_ruby_rake_task_block_call_gets_file_caller`:

```ruby
# build.rake
def compile
  1
end

task :build do
  compile()
end
```

`graphify update .` on the PR build produces only the `contains` edge; there is no `calls` edge sourced from the file node. Same result for a cross-file constant call inside a task block (`task :t do B.y end`) and for a bare top-level `B.y` in a `.rb` file. A control in the same project (the identical `B.y` call moved inside a method body) emits its `calls` edge fine, so calls extraction itself is healthy; it is specifically the file-caller edges that go missing. Padding the project past 60 files changes nothing, so it is not a small-project fast path.

Since your tests are green while the CLI does not produce the edge, the tests and the shipped pipeline are exercising different layers, and something between the extractor walk and the exported `graph.json` drops or never produces the file-sourced `calls` edges (same genus as #15, where edges existed in storage but disappeared from export). I did not chase the exact divergence point, but the repro is deterministic: any of the shapes above, `graphify update .`, then grep `graph.json` for a `calls` edge whose source is the file node.

Suggested next step: reproduce through the CLI rather than pytest, find where the file-sourced edges get lost, and consider one end-to-end test that runs the real build against a temp directory and asserts on the exported `graph.json`, since that is the surface users consume. Happy to re-run the corpus census on a new push; once the pipeline gap closes I would expect the rake number to jump from 1 into the hundreds here, which is the outcome #1972 is really after.

Rishet11:
@krishnateja7 Thank you for running the census — that was decisive, and you were right that the tests and the pipeline were exercising different layers. Reproduced your minimal repro exactly, and the cause is upstream of where I was testing.

**Root cause.** The built graph is an undirected `nx.Graph` (`build.py:673`; both `watch.py` and `cli.py` call `build_from_json` without `directed=True`). A node **pair** therefore holds exactly one edge, and the last write wins — `build.py:836-838` says so in a comment. Edges are inserted sorted by `(source, target, relation)`, and `"calls"` sorts before both `"contains"` and `"imports"`. Every top-level callee is a symbol its file already contains (same file) or imports (cross-file), so the file-sourced `calls` edge was written first and then overwritten by the structural edge on the identical pair.

The discriminating check, on your rake repro:

```
$ graphify update .
contains | build -> build_compile

contains | build -> build_compile
calls    | build -> build_compile      <-- survives here
```

`--no-cluster` skips `build_from_json` and uses the relation-aware `dedupe_edges`, so the edge lives. That is why your control (the call moved into a method) worked: its caller is a function node, so there is no pair collision. It also explains the 18 survivors — those are pairs that happened to carry no `contains`/`imports` edge.

**This means file-as-caller cannot work here**, and I want to be upfront that it changes what #1972 asked for. A `file → symbol` `calls` edge is unrepresentable in this graph model whenever the file also contains or imports that symbol, which is essentially always. Rather than change the graph model, I followed the existing precedent: the bash extractor has always attributed top-level calls to a synthetic `<file>__entry` node (`bash.py:136-139`, `:443`) for exactly this reason. Your repro now gives:

```
contains | build -> build_rake__entry
calls    | build_rake__entry -> build_compile
```

If you or @safishamsi would rather have the file node literally, say so — it needs `build_from_json` to become relation-aware per pair, which touches clustering, path queries, explain and export, so I did not bundle it here.

**On your suggestion for an end-to-end test:** added `tests/test_toplevel_calls_e2e.py`, which runs the real `graphify extract --code-only` against a temp repo and asserts on the exported `graph.json`. All three cases fail against the previous implementation. That layer is the reason this shipped broken, so it seemed like the right place to lock it down.

Also covered: a cross-file top-level call, a guard that files without a top-level call gain no node, and a corpus containing a symbol literally named `py_entry` — `normalize_id` collapses underscore runs, so that would otherwise share a normalized id with `<stem>_py__entry`.

Full suite unchanged against base (4 pre-existing env-var failures both before and after), ruff and skillgen clean. The branch is also rebased onto current `v8`, so the commits are a clean cherry-pick.

The census offer would be very welcome on this push — I would expect the rake number to move well off 1.

Rishet11:
Picking up the one review comment on this PR that was still unaddressed — thanks for flagging it, and sorry it sat.

The inline note on `_toplevel_calls_only` was right: the comment claimed the root walk emits "ONLY direct `calls` edges", but the flag only short-circuits module-scope indirect dispatch. Other relations `walk_calls` can emit by different routes — a JS/TS dynamic `import()` producing `imports_from` being the example given — are not affected by it. Reworded in `5c28462` to say what the flag actually does, and to state that narrower scope explicitly.

Comment-only change, no behavior touched. `ruff` clean.

The other inline note (the "#1744 removes" tense fix) no longer applies — that line is gone from the current branch after the rework.

PR #2017, opened 2026-07-19
fix(serve): reorder except clauses so the corrupted-graph message is reachable

### LINKED ISSUE #2005: `serve.py`: the "graph.json is corrupted" recovery message can never be printed (except-clause ordering)
`SECURITY.md` lists this under the threat table:

> | Corrupted graph.json | `_load_graph()` in `serve.py` wraps `json.JSONDecodeError` and prints a clear recovery message instead of crashing. |

The handler exists, but it is unreachable. In `graphify/serve.py`:

```python
    except (ValueError, FileNotFoundError) as exc:
        print(f"error: {exc}", file=sys.stderr)
        sys.exit(1)
    except json.JSONDecodeError as exc:
        print(f"error: graph.json is corrupted ({exc}). Re-run /graphify to rebuild.", file=sys.stderr)
        sys.exit(1)
```

`json.JSONDecodeError` subclasses `ValueError`, so the first clause always wins:

```python
>>> import json
>>> issubclass(json.JSONDecodeError, ValueError)
True
```

**Observed behaviour** with a truncated `graph.json`: users get the bare

```
error: Expecting value: line 1 column 1 (char 0)
```

instead of the intended `graph.json is corrupted (...). Re-run /graphify to rebuild.` This matters more than a cosmetic message, because the atomic-write and shrink-guard work in 0.9.18 explicitly treats a truncated `graph.json` as a state users need to recover from — and the recovery hint is exactly what never reaches them.

## Suggested fix

Move the `json.JSONDecodeError` clause above the `(ValueError, FileNotFoundError)` clause.

## How this was found

Static read of `graphify/serve.py` plus a local check of the exception hierarchy. No graphify process was run against a real corpus.

kimdzhekhon:
## Summary
- `json.JSONDecodeError` subclasses `ValueError`, so `_load_graph()`'s `except (ValueError, FileNotFoundError)` clause always matched first — the `except json.JSONDecodeError` clause with the "graph.json is corrupted (...). Re-run /graphify to rebuild." recovery message was unreachable dead code.
- Users hitting a truncated/corrupted `graph.json` got the bare underlying `json.JSONDecodeError` message instead, contradicting what `SECURITY.md` documents for this exact threat.
- Moved the `json.JSONDecodeError` clause first so it actually catches. Audited the rest of `serve.py`'s exception handlers for the same narrower-after-broader ordering bug — found no other instance.

Fixes #2005.

## Test plan
- [x] Added `test_load_graph_corrupted_json_prints_recovery_message` — writes invalid JSON, asserts the recovery message (not the bare decode error) is printed
- [x] `uv run pytest tests/test_serve.py -q -k load_graph` — 6 passed
- [x] `uv run pytest tests/ -q` — full suite: 3412 passed (5 pre-existing failures unrelated to this change, missing optional `openai` dependency in this environment)

### DIFF
--- a/graphify/serve.py
+++ b/graphify/serve.py
@@ -55,11 +55,11 @@ def _load_graph(graph_path: str) -> nx.Graph:
         except Exception:
             G.graph["_learning_overlay"] = {}
         return G
-    except (ValueError, FileNotFoundError) as exc:
-        print(f"error: {exc}", file=sys.stderr)
-        sys.exit(1)
     except json.JSONDecodeError as exc:
         print(f"error: graph.json is corrupted ({exc}). Re-run /graphify to rebuild.", file=sys.stderr)
+        sys.exit(1)
+    except (ValueError, FileNotFoundError) as exc:
+        print(f"error: {exc}", file=sys.stderr)
         sys.exit(1)
 
 
--- a/tests/test_serve.py
+++ b/tests/test_serve.py
@@ -612,6 +612,19 @@ def test_load_graph_corrupted_json_prints_recovery_message(tmp_path, capsys):
     assert "Re-run /graphify to rebuild" in err
 
 
+def test_load_graph_generic_value_error_message_unchanged(tmp_path, capsys):
+    """A non-decode ValueError (e.g. a non-.json path) must still print the
+    generic error, not the corrupted-graph hint — pins the except-clause
+    order from #2005 so a future refactor can't collapse them back."""
+    p = tmp_path / "graph.txt"
+    p.write_text("not a graph")
+    with pytest.raises(SystemExit):
+        _load_graph(str(p))
+    err = capsys.readouterr().err
+    assert "must be a .json file" in err
+    assert "corrupted" not in err
+
+
 def test_load_graph_rejects_oversized_file(monkeypatch, tmp_path, capsys):
     # #F4: oversized graph.json must fail fast (SystemExit) with a clear error.
     G = _make_graph()
--- a/tests/test_serve.py
+++ b/tests/test_serve.py
@@ -599,6 +599,19 @@ def test_load_graph_missing_file(tmp_path):
         _load_graph(str(graphify_dir / "nonexistent.json"))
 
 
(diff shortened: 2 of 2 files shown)


Job28703 (approved):
Independently reproduced the fail-then-pass: checked out `v8` base (unfixed) with the new test applied — `test_load_graph_corrupted_json_prints_recovery_message` fails exactly as expected (bare `JSONDecodeError` message instead of the recovery hint), confirming `json.JSONDecodeError` is indeed swallowed by the broader `(ValueError, FileNotFoundError)` clause ahead of it. With the reorder applied, the same test passes (6 passed on `test_serve.py -k load_graph`). Also spot-checked the rest of `serve.py`'s ~20 except clauses — no other narrower-after-broader ordering issue. Clean, well-scoped fix. Thanks!

HerenderKumar (commented):
Came here from #2005 — I had claimed it on the issue a few hours before this PR went up and had the same fix built locally, so I can vouch for the approach: `json.JSONDecodeError` subclasses `ValueError`, so moving its clause first is exactly right. I ran the same change against the full serve suites locally (117 passed, 1 skipped).

One suggestion: a small guard test asserting that a non-decode `ValueError` (e.g. a non-`.json` path) still prints the generic `error: ...` message rather than the corrupted-graph hint — it pins the clause order so a future refactor can't "simplify" it back:

```python
def test_load_graph_generic_value_error_message_unchanged(tmp_path, capsys):
    p = tmp_path / "graph.txt"
    p.write_text("not a graph")
    with pytest.raises(SystemExit):
        _load_graph(str(p))
    err = capsys.readouterr().err
    assert "must be a .json file" in err
    assert "corrupted" not in err
```

LGTM either way.

kimdzhekhon:
Good catch — added that exact guard test. Thanks for vouching for the approach and for validating against the full serve suite.

safishamsi:
Thanks @kimdzhekhon — cherry-picked (both the fix and your clause-order regression test) and shipped in v0.9.22. Released to PyPI: https://github.com/Graphify-Labs/graphify/releases/tag/v0.9.22

PR #2019, opened 2026-07-19
fix(llm): fall back to OLLAMA_HOST when OLLAMA_BASE_URL is unset

### LINKED ISSUE #1940: Env var `OLLAMA_BASE_URL` is inconsistent with actual Ollama-held env var name
`Graphfy` uses `OLLAMA_BASE_URL` to configure a custom API target here:

https://github.com/Graphify-Labs/graphify/blob/cb96bdaa0c367bec8d5c5aee5d7c9ebb727e9780/graphify/llm.py#L86

However, Ollama defines the `OLLAMA_HOST` env var and doesn't actually have a `OLLAMA_BASE_URL` concept at all:

[Ollama Documentation How Do I Configure Ollama Server](https://docs.ollama.com/faq#how-do-i-configure-ollama-server)

---

Expected behavior:
Graphify should re-use the Ollama env vars so that custom configuration is automatically picked up.

kimdzhekhon:
## Summary
- Ollama's own server has no `OLLAMA_BASE_URL` concept — its documented config var is `OLLAMA_HOST` ([Ollama FAQ](https://docs.ollama.com/faq#how-do-i-configure-ollama-server)). graphify only ever read `OLLAMA_BASE_URL`, so a user who configured Ollama the way its own docs describe got silently routed to the `localhost:11434` default instead.
- Adds `_resolve_ollama_base_url(default)`: `OLLAMA_BASE_URL` still wins when set (unchanged behavior — this mirrors graphify's existing per-backend `*_BASE_URL` override convention, e.g. `KIMI_BASE_URL`, `GEMINI_BASE_URL`, left untouched). Otherwise falls back to `OLLAMA_HOST`, normalized into an OpenAI-compatible URL: adds `http://` if no scheme is present, ensures a trailing `/v1` (no duplication if already present).
- Wired into all 3 read sites: the `BACKENDS["ollama"]["base_url"]` dict-literal default and the two `ollama_url = os.environ.get("OLLAMA_BASE_URL", ...)` call sites.

Fixes #1940.

## Test plan
- [x] Added 4 tests in `tests/test_llm_backends.py`: `OLLAMA_BASE_URL` takes priority over `OLLAMA_HOST`; scheme-less `OLLAMA_HOST` gets `http://` + `/v1`; an already-normalized `OLLAMA_HOST` (scheme + `/v1`) passes through unchanged; neither set → falls back to caller's default
- [x] `uv run pytest tests/test_llm_backends.py -q -k ollama` — 13 passed
- [x] `uv run pytest tests/ -q` — full suite: 3415 passed (5 pre-existing failures unrelated to this change, missing optional `openai` dependency in this environment)

### DIFF
--- a/graphify/llm.py
+++ b/graphify/llm.py
@@ -56,6 +56,22 @@ def _get_tokenizer():
 # Cached at import time. None if tiktoken is unavailable; consumers must handle.
 _TOKENIZER = _get_tokenizer()
 
+
+def _resolve_ollama_base_url(default: str) -> str:
+    ollama_base_url = os.environ.get("OLLAMA_BASE_URL")
+    if ollama_base_url is not None:
+        return ollama_base_url
+    ollama_host = os.environ.get("OLLAMA_HOST")
+    if ollama_host is None:
+        return default
+    if not ollama_host.startswith(("http://", "https://")):
+        ollama_host = f"http://{ollama_host}"
+    ollama_host = ollama_host.rstrip("/")
+    if not ollama_host.endswith("/v1"):
+        ollama_host = f"{ollama_host}/v1"
+    return ollama_host
+
+
 BACKENDS: dict[str, dict] = {
     "claude": {
         # ANTHROPIC_BASE_URL points the backend at any Anthropic-compatible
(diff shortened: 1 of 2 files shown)


safishamsi:
Thanks @kimdzhekhon — shipped in v0.9.21, with a follow-up so the fallback also drives backend auto-detection and defaults the port to 11434. Released to PyPI: https://github.com/Graphify-Labs/graphify/releases/tag/v0.9.21

PR #2024, opened 2026-07-19
fix(hook): nudge when wiki/ predates graph.json

### LINKED ISSUE #2006: `wiki/` is only ever written by the explicit `export wiki` subcommand, yet `AGENTS.md` tells agents to navigate it instead of reading source
### The wiring

`to_wiki` has exactly one call site:

- `graphify/cli.py:2188-2199` — inside the `export wiki` subcommand.

It is not reached from any other path:

- `graphify/watch.py` contains no reference to `wiki` at all.
- `graphify update` does not touch `wiki/`.
- The post-commit/rebuild hook path does not touch `wiki/`.

So once `graphify-out/wiki/` exists, it is frozen until someone manually re-runs `graphify export wiki`.

### The instruction

Meanwhile the always-loaded agent instructions state it unconditionally. From the repo's own `AGENTS.md`:

> - If `graphify-out/wiki/index.md` exists, navigate it instead of reading raw files

The same unconditional sentence appears in `graphify/always_on/agents-md.md` and `graphify/always_on/antigravity-rules.md` (the latter carries `trigger: always_on`).

### Why this is inconsistent rather than just unfortunate

`graph.json` is handled carefully — `cli.py:496-512` gates on mtime plus a `needs_update` sentinel, and 0.9.19 went further by softening the PreToolUse hook to a non-mandatory nudge when the graph is stale *for the target file*. That is a good design. `wiki/` gets none of it, while carrying a stronger instruction ("instead of reading raw files").

Two signals suggest this is an oversight rather than a decision:

1. `graphify/always_on/vscode-instructions.md` **does** include an escape hatch — "…or the graph is missing or stale" — but `antigravity-rules.md` and `AGENTS.md` do not. Users get different protection depending on which assistant they installed.
2. Everything else in the codebase treats staleness as a first-class concern (`_cleanup_stale_ast_entries`, the `#1939` salt-keyed cache miss, the `#1908` excluded-but-alive row drop, the incremental re-extract logic).

### Practical impact

An agent following the instruction answers architecture questions from a wiki that may be months out of date, with no staleness signal anywhere in the loop — the failure is silent and looks like a confident correct answer.

## Suggested fix (either is fine)

- Wire `to_wiki` into the `update`/`watch`/hook rebuild paths (it is AST-derived, so no API cost), **or**
- Add the same staleness condition to `AGENTS.md` and the `always_on/` variants that `vscode-instructions.md` already has, and have the wiki index record the build stamp it was generated from.

## How this was found

Static read of `cli.py`, `watch.py`, `wiki.py`, `AGENTS.md`, and `graphify/always_on/*`. Call-site enumeration by grep across `graphify/`. No runtime reproduction.

Job28703:
Fixes #2006

## Problem

`to_wiki` has exactly one call site (`export wiki` subcommand). It's never reached from `update`, `watch`, or the rebuild hook path, so `graphify-out/wiki/` freezes the moment it's generated. Meanwhile `AGENTS.md` and the `always_on/` prose variants tell agents to navigate the wiki unconditionally instead of reading source — an agent can silently answer architecture questions from a wiki that's months stale, with no signal anywhere in the loop.

`graph.json` already has this kind of staleness handling: the PreToolUse hook in `cli.py` compares `graph.json`'s mtime against the target file's mtime (plus a `needs_update` sentinel) and softens to a nudge instead of blocking. `wiki/` had no equivalent.

## Fix

- Extend the existing `_run_hook_guard("read", ...)` staleness check: when the read target is under `wiki/`, compare `wiki/index.md`'s mtime against `graph.json`'s. If wiki is older, emit a new `_READ_NUDGE_WIKI_STALE` nudge pointing the agent back to `graphify query`/source, and to `graphify export wiki` to refresh it.
- Update `AGENTS.md`, `always_on/agents-md.md`, and `always_on/antigravity-rules.md` — all three told agents to trust the wiki unconditionally — to reference this new signal instead.

Docs-only fix, no change to when/how `wiki/` is generated (that's the bigger option the issue also names — wiring `to_wiki` into `update`/`watch` — left for a separate PR if wanted).

## Test plan

- [x] Added 3 tests to `tests/test_hook_guard.py`: nudges when wiki predates graph.json, silent when wiki is fresh, silent (fail-open) when graph.json doesn't exist
- [x] Reproduced fail-then-pass by hand: reverted just the `cli.py` change (keeping the new tests) — `test_read_wiki_nudges_when_stale` fails as expected (`assert 'predates the current graph.json' in ''`). Restored the fix — passes.
- [x] `pytest tests/test_hook_guard.py tests/test_hook_strict.py -q` — 89 passed, no regressions
- [x] `pytest tests/ -q` — full suite: 3239 passed, 167 skipped, 13 pre-existing failures unrelated to this change (`tree_sitter_hcl` not installed, ollama SDK version mismatch, wheel-packaging test — all in `test_terraform.py`/`test_ollama_retry_cap.py`/`test_manifest_ingest.py`/`test_install_references.py`, none touching `cli.py` or hook logic)


Job28703:
Pushed a follow-up commit addressing the review comment above: both `_READ_NUDGE_STALE` and `_READ_NUDGE_WIKI_STALE` are now functions parameterized by the actual configured output dir name instead of hardcoding `graphify-out/`. Added 2 tests covering a custom `GRAPHIFY_OUT_NAME`.

Also caught and fixed a self-inflicted issue while doing that: my first commit hand-edited the *generated* `always_on/agents-md.md` and `antigravity-rules.md` directly instead of their fragment sources under `tools/skillgen/fragments/`, which `test_always_on_files_are_guarded_by_check` correctly flags as drift. Fixed by editing the fragments, regenerating, and registering the wording change in `gen.ALWAYS_ON_SANCTIONED_EDITS` (same mechanism as the existing #1530 entry) so the byte-for-byte roundtrip guard treats it as an intentional edit.

Full suite is back to the same 13 pre-existing, unrelated failures (tree_sitter_hcl/ollama/manifest/wheel-packaging) with all 3241 other tests passing.

PR #2034, opened 2026-07-19
fix(extract): resolve alias re-exports and imports through barrels to the defining symbol (#1983)

### LINKED ISSUE #1983: Alias-based named re-exports emit dangling absolute-path symbol targets
### Context

On current `v8` (`8cee776c`, graphifyy 0.9.18), a named re-export through a tsconfig path alias emits a symbol-level `re_exports` edge whose target contains the checkout path and does not exist in the graph. The edge is later dropped, so symbol-level provenance through the barrel is lost.

This is the separate follow-up requested in the maintainer comment on #1967. Open PR #1976 still reproduces this case at its current head; it fixes cross-extension file-node disambiguation rather than alias-derived symbol IDs.

### Reproducer

`tsconfig.json`:

```json
{"compilerOptions":{"baseUrl":".","paths":{"@/*":["src/*"]}}}
```

`src/lib/utils.ts`:

```ts
export function formatDate() { return "ok" }
```

`src/lib/index.ts`:

```ts
export { formatDate } from "@/lib/utils"
```

Run extraction with relative inputs from the project root:

```python
from pathlib import Path

result = extract(sorted(Path("src").rglob("*.ts")), cache_root=Path("."))
node_ids = {node["id"] for node in result["nodes"]}
print([
    (edge["target"], edge["target"] in node_ids)
    for edge in result["edges"]
    if edge.get("relation") == "re_exports"
])
```

### Current behavior

Path-sanitized relevant output:

```text
("<checkout>_src_lib_utils_formatdate", false)
("src_lib_utils", true)
```

The AST extractor is sufficient to reproduce this; no semantic cache entry is involved.

### Expected

The symbol-level edge should target the existing canonical symbol `src_lib_utils_formatdate`, while the file-level `re_exports` edge remains unchanged. No target ID should contain a checkout-specific prefix.

HerenderKumar:
## Summary

Follow-up to #1983. e5a4b51 fixed the issue's direct reproducer, but its rewrite learns old→canonical symbol forms only from symbols a file **defines**. A barrel defines nothing — so a named re-export or import that resolves *through* a barrel still synthesizes an absolute-prefixed target no rewrite ever learns: the checkout path leaks into the edge target and the edge dangles.

Verified on current `v8` head:

```text
re_exports: src_barrel2 -> <checkout>_src_lib_index_formatdate  (dangling)
imports:    src_consumer -> <checkout>_src_lib_index_formatdate  (dangling)
```

for `export { formatDate } from '@/lib'` / `import { formatDate } from '@/lib'` where `src/lib/index.ts` is itself a barrel re-exporting from `@/lib/utils`.

## Fix

Three parts, extending existing machinery rather than adding a new pass:

- `_import_js` stamps `target_file` on symbol-level `re_exports`/`imports` edges — the same transient hint the file-level edge has carried since #1814 (no downstream reader; stripped at build).
- The id-remap post-pass gains a repoint step after the candidates rewrite: decompose a still-unowned target into (canonical file stem, symbol) via the stamp, follow the barrel's own already-canonical `re_exports` edge to the defining symbol — iterating one hop per pass so multi-hop barrel chains resolve — and, when no chain leads to a real node, canonicalize the prefix anyway so a checkout-specific path never survives in an edge target.
- The redundant-twin dedup now excludes `target_file` from edge identity, so an alias edge (stamped) still matches the canonical twin the shared resolver emits (unstamped).

## Testing

- 3 new tests: single-hop barrel re-export resolves to the defining symbol; two-hop chain resolves; no `re_exports`/`imports` target contains the checkout prefix (the issue's core "Expected" line).
- Red-green verified: with the production change reverted, exactly the 3 new tests fail; restored, `tests/test_js_import_resolution.py` passes 61/61 — including the e5a4b51 owned-symbol guard tests (`test_alias_import_does_not_remap_an_owned_symbol_id`, `test_alias_import_preserves_owned_same_line_symbol_edge`).
- Regression sweep: failure sets for `tests/test_extract.py` + `tests/test_charmap_encoding.py` are name-for-name identical between this branch and clean `origin/v8` in my environment (31 pre-existing failures from missing grammar packs on both sides).

Fixes #1983.

### DIFF
--- a/graphify/extract.py
+++ b/graphify/extract.py
@@ -359,6 +359,12 @@ def _import_js(node, source: bytes, file_nid: str, stem: str, edges: list, str_p
                                     "source_file": str_path,
                                     "source_location": f"L{line}",
                                     "weight": 1.0,
+                                    # Which file this symbol target was synthesized
+                                    # from, so the id-remap post-pass can repoint a
+                                    # target the candidates rewrite never learns —
+                                    # a barrel defines no symbols (#1983). Transient,
+                                    # stripped at build like the #1814 stamp.
+                                    "target_file": str(resolved_path),
                                 })
         else:
             # Handle: import { Foo, type Bar } from './bar'
@@ -380,6 +386,8 @@ def _import_js(node, source: bytes, file_nid: str, stem: str, edges: list, str_p
                                             "source_file": str_path,
                                             "source_location": f"L{line}",
                                             "weight": 1.0,
+                                            # See the re_exports stamp above (#1983).
+                                            "target_file": str(resolved_path),
                                         })
 
 
@@ -4520,6 +4528,11 @@ def extract(
     # would otherwise orphan (#1529). Stored as a list so the symbol-prefix remap
     # below can try both (identical forms collapse to one — a no-op).
     prefix_remap: dict[Path, list[tuple[str, str]]] = {}
+    # Canonical stem plus every prefix form a file's symbol ids may appear
+    # under, keyed by resolved path — consumed by the target_file-guided
+    # barrel repoint below (#1983). Unlike prefix_remap this records ALL
+    # in-root files, not just those whose prefix changed.
+    stem_forms: dict[Path, tuple[str, list[str]]] = {}
     for path in paths:
         old_id = _make_id(str(path))
         try:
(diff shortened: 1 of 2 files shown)


safishamsi:
Thanks @HerenderKumar — shipped in v0.9.21, with a follow-up to leave a genuinely ambiguous barrel re-export unresolved rather than guess a target. Released to PyPI: https://github.com/Graphify-Labs/graphify/releases/tag/v0.9.21

PR #2038, opened 2026-07-19
Fix collapsed import self-loops

### LINKED ISSUE #2037: Bare import aliases can collapse into file-level self-loops
## Summary

Graph construction can turn an unresolved bare import target into a self-loop when the target matches a legacy alias for the importing file.

Examples seen with Graphify 0.9.20:

- Python stdlib: `src/contracting/stdlib/builtins.py` containing `import builtins`
- Python package: `playground/services/contracting.py` containing `from contracting import constants`
- Rust parent module imports such as `use crate::fixture::{...}` inside tests in `fixture.rs`
- Rust external paths containing the current basename, such as `...::poseidon::{...}` in `poseidon.rs`

The AST extraction initially emits a bare target such as `builtins`, `contracting`, or `poseidon`. During `build_from_json()`, the legacy-ID alias index can resolve that target to the importing file node, yielding an `imports` or `imports_from` edge whose source and target are identical.

## Expected behavior

A file-level import or re-export that resolves to its own source should not be materialized. It carries no connectivity at file-node granularity and is reported as unhealthy by `graphify diagnose multigraph`.

Other meaningful self-edges, especially recursive `calls`, should remain intact.

## Proposed fix

After endpoint resolution in `build_from_json()`, discard only self-loops whose relation is one of:

- `imports`
- `imports_from`
- `re_exports`

A regression suite should cover the Python and Rust cases above and verify that recursive call self-loops are preserved.

## Validation

The proposed patch removes all seven observed import self-loops across two real repositories. Their rebuilt diagnostics report zero self-loops, dangling endpoints, missing endpoints, or endpoint-collapse groups. The full Graphify test suite passes: 3,478 passed, 3 skipped.

Endogen:
## Summary

- drop file-level `imports`, `imports_from`, and `re_exports` edges when endpoint resolution collapses them into self-loops
- cover Python stdlib/package basename collisions and Rust parent/external module cases
- verify that meaningful recursive `calls` self-loops remain intact

Fixes #2037.

## Root cause

AST extraction can emit unresolved bare import targets such as `builtins`, `contracting`, or `poseidon`. The legacy-ID alias index in `build_from_json()` may then resolve that bare target to the importing file's own canonical node. Materializing the resolved edge creates a file-level self-loop that carries no connectivity and is flagged by `graphify diagnose multigraph`.

The filter is intentionally relation-specific. It removes only collapsed import/re-export edges and preserves other self-edges, including recursive calls.

## Impact

Rebuilt graphs no longer contain false import self-loops caused by basename/legacy-alias resolution. This eliminates diagnostic noise without changing application code or hiding real recursive program structure.

## Validation

- `uv run pytest tests/ -q` — 3,478 passed, 3 skipped
- `uv run ruff check .` — passed
- focused build/extraction suite — 148 passed
- rebuilt two affected real-world graphs — zero self-loops, dangling endpoints, missing endpoints, or endpoint-collapse groups

### DIFF
--- a/graphify/build.py
+++ b/graphify/build.py
@@ -767,6 +767,16 @@ def build_from_json(extraction: dict, *, directed: bool = False, root: str | Pat
                 # (unknown ext, e.g. a manifest) is never mistaken for a phantom.
                 if src_fam is not None and tgt_fam is not None and src_fam != tgt_fam:
                     continue
+        # A file-level import or re-export cannot carry useful connectivity when
+        # both endpoints resolve to the same node.  This most often happens when
+        # the target is an unresolved bare module name (``builtins``, ``poseidon``)
+        # that the legacy-ID alias index above mistakes for the importing file's
+        # own old stem.  It also covers a nested module importing its parent file:
+        # at file-node granularity that relationship necessarily collapses.  Keep
+        # other self-edges, notably recursive ``calls``, because those are real
+        # program structure rather than import-resolution artifacts.
+        if src == tgt and _edge_rel in ("imports", "imports_from", "re_exports"):
+            continue
         # Preserve original edge direction - undirected graphs lose it otherwise,
         # causing display functions to show edges backwards.
         attrs["_src"] = src
--- a/tests/test_import_self_loops.py
+++ b/tests/test_import_self_loops.py
@@ -1,0 +1,107 @@
+from pathlib import Path
+
+import pytest
+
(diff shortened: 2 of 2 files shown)


safishamsi:
Thanks @Endogen — shipped in v0.9.21. Released to PyPI: https://github.com/Graphify-Labs/graphify/releases/tag/v0.9.21

Issue #1377, opened 2026-06-18
Microsoft Agent Package Manager: Parent Package Nodes Duplicated


rcollette:
[Microsoft Agent Package Manager (APM)](https://microsoft.github.io/apm/) behaves in a similar manner to other package managers, allowing packages to declare dependencies on other packages.  Below are the apm.yml files for a parent core language agnostic coding standards package and a dependent language specific package.  In some cases a child package will show linkage to the core package.  In other cases, the core package is represented as a duplicate node in the graph.

Core package apm.yml file

```yml
name: coding-standards-core
version: 1.0.4
description: General code standards and clean code evaluation standards applicable to all languages.
```

Package with dependency on core apm.yml file

```yml
name: coding-standards-csharp
version: 1.1.0
description:
  C# and .NET coding standards including services, repositories, HTTP clients, cross-cutting concerns, and unit testing.
dependencies:
  apm:
    - MyPrivateGithubOrg/apm-internal-marketplace/packages/coding-standards-core
```

Notably, the first two path segments are the github org and repo and the remaining path portion is the path within the repo.

The directed graph was constructed using GPT 5.4 and root cause analyzed using the same

---

**Root cause: the extractor has no canonical rule for APM package dependencies, so `coding-standards-core` gets materialized as whatever local node a chunk happened to invent first.** That fragments one real dependency target into multiple graph nodes, and some package edges end up pointing at different stand-ins.

In the fresh graph there are already **three** different core-related nodes:
- `coding-standards-core package` from `packages/coding-standards-core/apm.yml`
- `coding-standards-core` from `packages/coding-standards-python/apm.yml`
- `coding-standards-core dependency` from `packages/coding-standards-python-fastapi/SKILL.md`

The key tell is that some edges point to a node whose ID is anchored to the **wrong source file**. For example, several Go standards docs reference target ID `coding_standards_python_apm_coding_standards_core`, which is a node sourced from **Python’s** `apm.yml`, not from Go. That means the semantic subagent created one shared “coding-standards-core” concept inside a mixed chunk and anchored it to an arbitrary file in that chunk, then reused that ID for unrelated files.

So the failure is:

1. **Dependencies are being extracted semantically from YAML/docs instead of structurally canonicalized** to the actual package node in `packages/coding-standards-core/apm.yml`.
2. **Node IDs are file-anchored by design** (`{parent}_{file}_{entity}`), which is fine for file-local concepts but wrong for a global package identity.
[...]

PR #2043, opened 2026-07-19
feat(extract): add Kubernetes/ArgoCD/Helm-values YAML extractor

stevenpollack:
## Problem

graphify has no YAML extractor. Kubernetes manifests, ArgoCD GitOps config and Helm values all fall through to the LLM **document** path — expensive, and what comes back is inferred prose relationships rather than a real deployment graph.

On a Helm + ArgoCD monorepo the practical effect is that the deployment topology is unqueryable. I hit this trying to answer "how does this service reach prod?" — every file holding the answer was invisible to the graph.

## What this adds

`extract_k8s()` parses the YAML that genuinely *is* YAML and emits:

`contains` · `in_namespace` · `references` · `deploys_to` · `deploys_chart` · `generates_from` · `produces`

Three decisions worth reviewing:

**1. Global `<kind>.<name>` addressing.** `extract_terraform` scopes IDs by directory because Terraform resources are module-scoped. Kubernetes isn't: an ApplicationSet's `project: platform` means the AppProject named `platform` *wherever it is declared*. Global scoping is what lets those cross-file edges survive the per-file extraction merge. Pinned by `test_cross_file_reference_resolves_by_global_address`.

**2. References matched by key NAME, not fixed spec paths.** An ApplicationSet nests an entire Application spec under `spec.template.spec`. A depth-agnostic recursive walk keyed on field names picks those up with zero ApplicationSet-specific code — that's most of why this is ~290 lines and not 600.

**3. Detection by CONTENT, not extension.** `.yaml` **stays** in `DOC_EXTENSIONS`; only files carrying both `apiVersion:` and `kind:` are rerouted to the AST path. Ordinary YAML (CI configs, docs data, front-matter) keeps going to the LLM path exactly as before. This mirrors the existing `is_package_manifest_path()` precedent (#1377).

Security note: the sniff runs *after* `_is_sensitive()` in `detect()`, so a `secrets.yaml` is dropped before it can reach the AST path. I verified that ordering rather than assuming it.

## Deliberately out of scope

Helm **templates** (`charts/*/templates/*.yaml`). Go template directives make them invalid YAML — measured on a real chart, **0 of 46 template files parse**, while 100% of the manifests and values files do. They degrade to an empty result plus an error string rather than raising.

Documented in the module docstring, so the graph doesn't overstate what it knows:

- A git generator yields `generates_from` to the **glob** it scans, not the files that glob matches. Resolving it would mean globbing the filesystem from inside a single-file extractor, breaking per-file caching and determinism.
- Registry-style YAML with neither `apiVersion` nor `kind` (an ArgoCD fleet entry is often just `app: <name>`) is indistinguishable from arbitrary config, so it stays a document.
- Same-named objects across clusters/namespaces collapse into one node.

## Validation

Real Helm + ArgoCD monorepo, 31 manifests: **0 parse errors**, 94-node graph end-to-end through `detect` → `extract` → `build_from_json`.

The `generates_from` edges alone answer "which services can reach prod at all":

```
ApplicationSet.fleet-apps-dev             -> path.gitops/apps/*.yaml
ApplicationSet.fleet-apps-dev             -> path.gitops/apps-dev/*.yaml
ApplicationSet.fleet-apps-prod            -> path.gitops/apps/*.yaml
ApplicationSet.fleet-apps-staging         -> path.gitops/apps/*.yaml
ApplicationSet.fleet-apps-under-migration -> path.gitops/apps-under-migration/*.yaml
```

Two ApplicationSets read different registries — which is exactly why one service deploys to dev but never to prod. That was previously invisible.

## Dependency

`pyyaml` as the optional `[k8s]` extra, matching the `sql` / `terraform` / `pascal` pattern. Without it, affected files hit the standard missing-extra warning (`_EXTRA_FOR_EXTENSION`) instead of failing the run.

## Tests

14 new tests in `tests/test_k8s.py` (+ a registry-identity test following the `test_<lang>_migrated` convention), covering each relation, ApplicationSet nesting, multi-document files, real line numbers, graceful Helm-template failure, and both directions of the detection sniff. Guarded by `pytest.importorskip("yaml")`.

Full suite: **3424 passed**. Four failures in `test_ollama_retry_cap.py` (`ModuleNotFoundError: No module named 'openai'`) reproduce identically on `v8` without this change — pre-existing and unrelated.

### DIFF
--- a/graphify/detect.py
+++ b/graphify/detect.py
@@ -442,6 +442,14 @@ def classify_file(path: Path) -> FileType | None:
     from graphify.manifest_ingest import is_package_manifest_path
     if is_package_manifest_path(path):
         return FileType.CODE
+    # Kubernetes / ArgoCD / External-Secrets manifests are deterministic YAML, so
+    # they take the AST path too. Sniffed by CONTENT (apiVersion + kind), never by
+    # extension: .yaml stays a DOC extension so ordinary YAML (CI configs, docs
+    # data) keeps going to the LLM path. Runs after _is_sensitive() in detect(),
+    # so a secrets.yaml is dropped before it can reach this branch.
+    from graphify.extractors.k8s import is_k8s_manifest_path
+    if is_k8s_manifest_path(path):
+        return FileType.CODE
     # Compound extensions must be checked before simple suffix lookup
     if path.name.lower().endswith(".blade.php"):
         return FileType.CODE
--- a/graphify/extract.py
+++ b/graphify/extract.py
@@ -43,6 +43,7 @@ from graphify.extractors.elixir import extract_elixir  # noqa: F401
 from graphify.extractors.fortran import _cpp_preprocess, extract_fortran  # noqa: F401
 from graphify.extractors.go import extract_go  # noqa: F401
 from graphify.extractors.json_config import extract_json  # noqa: F401
+from graphify.extractors.k8s import extract_k8s  # noqa: F401
 from graphify.extractors.markdown import extract_markdown  # noqa: F401
 from graphify.extractors.pascal_forms import extract_delphi_form, extract_lazarus_form  # noqa: F401
 from graphify.extractors.powershell import extract_powershell, extract_powershell_manifest  # noqa: F401
@@ -3923,6 +3924,10 @@ _DISPATCH: dict[str, Any] = {
     ".tf": extract_terraform,
     ".tfvars": extract_terraform,
     ".hcl": extract_terraform,
+    # Only reached for YAML that detect.classify_file sniffed as a k8s manifest
+    # (apiVersion + kind); ordinary .yaml stays a document on the LLM path.
+    ".yaml": extract_k8s,
+    ".yml": extract_k8s,
     ".dm": extract_dm,
     ".dme": extract_dm,
     ".dmi": extract_dmi,
@@ -3952,6 +3957,8 @@ _EXTRA_FOR_EXTENSION = {
     ".hcl": "terraform",
     ".dm": "dm",
     ".dme": "dm",
+    ".yaml": "k8s",
+    ".yml": "k8s",
 }
 
 
--- a/graphify/extractors/__init__.py
+++ b/graphify/extractors/__init__.py
@@ -20,6 +20,7 @@ from graphify.extractors.fortran import extract_fortran
 from graphify.extractors.go import extract_go
 from graphify.extractors.json_config import extract_json
 from graphify.extractors.julia import extract_julia
+from graphify.extractors.k8s import extract_k8s
 from graphify.extractors.markdown import extract_markdown
 from graphify.extractors.objc import extract_objc
 from graphify.extractors.pascal import extract_pascal
@@ -48,6 +49,7 @@ LANGUAGE_EXTRACTORS: dict[str, Callable[[Path], dict]] = {
     "go": extract_go,
     "json": extract_json,
     "julia": extract_julia,
+    "k8s": extract_k8s,
     "lazarus_form": extract_lazarus_form,
     "markdown": extract_markdown,
     "objc": extract_objc,
--- a/graphify/extractors/k8s.py
+++ b/graphify/extractors/k8s.py
@@ -1,0 +1,284 @@
+"""Kubernetes / Helm-values / ArgoCD-GitOps YAML extractor.
+
+Covers the manifests that are VALID YAML: plain k8s objects, ArgoCD
+Application/ApplicationSet/AppProject, External Secrets, and Helm `values*.yaml`.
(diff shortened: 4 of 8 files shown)


wrightbuilt:
Offering a second real-world corpus for this, if it is useful for validation.

I run graphify on a private NixOS + k3s infrastructure monorepo (AST-only refresh, `graphifyy==0.9.26`, committed graph). Measuring the committed `graph.json` against `manifest.json`:

| extension | files | nodes |
|---|---|---:|
| `.md` | 370 | 5981 |
| `.ts` | 239 | 1835 |
| `.json` | 42 | 414 |
| `.sh` | 34 | 174 |
| `.py` | 6 | 110 |
| **`.yaml` / `.yml`** | **515** | **0** |

So ~45% of the repo's source files are invisible to the graph, and the missing half is precisely the deployment topology this PR targets. The failure matches your description exactly: the files are classified as documents, land in `manifest.json`, and contribute nothing on a deterministic run.

Corpus shape, in case it is useful as a test target — 464 YAML files under `k8s/`:

| kind | count |
|---|---:|
| NetworkPolicy | 192 |
| Service | 95 |
| Deployment | 82 |
| IngressRoute (Traefik CRD) | 65 |
| ExternalSecret (ESO CRD) | 60 |
| PersistentVolumeClaim | 46 |
| Kustomization | 45 |
| ConfigMap | 44 |
| PrometheusRule | 38 |
| **Application (ArgoCD)** | **37** |
| Namespace | 34 |
| CronJob / Job | 30 |

Two things about this corpus that may exercise paths a Helm-centric repo would not:

1. **ArgoCD app-of-apps, no Helm.** 37 `Application` resources, only 2 `chart:` references — everything else is `path:`-based Kustomize/plain manifests. So the `deploys_to` / `contains` edges matter here much more than `deploys_chart`, and your global `<kind>.<name>` addressing decision is what would make the app-of-apps parent/child edges resolve across files. That seems worth a test case independent of Helm.
2. **Heavy CRD usage.** IngressRoute, ExternalSecret, Probe, ServiceMonitor and PrometheusRule together are ~190 files. Detection by content rather than extension should handle these, but they are not core-API kinds, so it is worth knowing whether the recursive name-keyed walk picks up their references or quietly skips them.

The practical consequence on our side is that "which Application owns this workload, and what does it reach" is currently unanswerable from the graph, which is the same question you hit. Happy to run a build against this corpus and report node/edge counts or failures if that would help move the PR — just say what output would be useful.

No pressure intended on timing; mainly recording that the gap has more than one reporter and that there is a non-Helm GitOps corpus available to test against.

wrightbuilt:
Following up on the corpus offer above — the branch is now `CONFLICTING` against main and has not been rebased since 2026-07-19, so I wanted to check whether it is still live before spending time on it.

Two things I can do, if either is useful:

1. **Rebase onto current main and report the result** — happy to hand back a patch rather than push anywhere.
2. **Run the extractor against the 464-file ArgoCD/Kustomize corpus** described above and report node/edge counts, parse failures, and specifically whether the recursive name-keyed walk picks up the CRD-heavy files (IngressRoute, ExternalSecret, PrometheusRule, ServiceMonitor — roughly 190 files between them).

Since I posted, I confirmed the gap is unchanged on `0.9.32`: `.yaml`/`.yml` sit in `DOC_EXTENSIONS` (`detect.py:32`) and `_get_extractor()` returns `None` for both, so a deterministic AST-only run indexes none of them — 0 nodes from 515 files in our case.

Your decision 3 — keep `.yaml` in `DOC_EXTENSIONS` and reroute by content sniff on `apiVersion:` + `kind:` — is what makes this adoptable for us without changing the cost profile of the rest of the corpus. Worth saying explicitly, since it is the design choice a reviewer is most likely to push back on.

No pressure on timing; mainly that it would be a shame for the branch to rot if it is only a rebase away.

stevenpollack:
Hey Scott,

I ended up abandoning graphify for a more strategic use of LSPs. You may want to actually run a comparison using your corpus: this branch of graphify vs. Helm-ls + schema-annotated yaml.

wrightbuilt:
Thanks for the straight answer — that is genuinely more useful than a rebase would have been.

We may well run that comparison. Our corpus is a reasonable test for it: 464 YAML files, ArgoCD app-of-apps with almost no Helm, so it leans on cross-file `Application` → workload resolution rather than chart values — which is roughly the case where a schema-aware LSP and a graph should diverge most. If we do run it, I will post the results here, since anyone else landing on this PR is asking the same question you were.

Leaving the branch alone for now rather than carrying it. Appreciate you writing it up as carefully as you did — the module docstring on what it deliberately does *not* resolve is what made it possible to evaluate without running it.

stevenpollack:
Please do report back, I'd love to see what the results are.

PR #2061, opened 2026-07-20
fix(hooks): skip post-checkout rebuild when HEAD doesn't move

### LINKED ISSUE #2060: post-checkout hook rebuilds on branch creation even when PREV_HEAD == NEW_HEAD
The post-checkout hook installed by `graphify hook install` triggers a full
background rebuild on any branch switch (`$3 == 1`), including
`git checkout -b <new-branch>`. In that case `$1` (PREV_HEAD) and `$2`
(NEW_HEAD) are identical — no file in the tree actually changed — yet the
hook still launches `_rebuild_code`, which then produces reordering/community
churn in graph.json unrelated to any real code change.

In a shared repo this generates recurring, purely cosmetic diffs on
graphify-out/* every time someone starts a new branch, which teams end up
discarding via `git stash` rather than committing.

Suggested fix: early-exit when `$1 = $2`, e.g. right after the existing
`BRANCH_SWITCH != 1` check in the generated hook template:

    if [ "$PREV_HEAD" = "$NEW_HEAD" ]; then
        exit 0
    fi

Version: graphifyy 0.9.16 (hook template written by `graphify hook install`)

vlecourt:
## Summary
- `git checkout -b <new-branch>` off the current HEAD sets `BRANCH_SWITCH=1` even though `PREV_HEAD`/`NEW_HEAD` are identical, so the post-checkout hook launched a full graph rebuild for a switch with no underlying code change.
- Since community detection isn't fully deterministic across runs, that produced pure reordering churn in `graphify-out/graph.json`, which teams ended up discarding via `git stash`.
- Added an early-exit guard right after the existing `BRANCH_SWITCH != 1` check: skip the rebuild when `PREV_HEAD` equals `NEW_HEAD`.

Fixes #2060

## Test plan
- [x] `uv run pytest tests/test_hooks.py -q` — 61 passed (3 new tests covering the guard, static + end-to-end)
- [x] `uv run ruff check graphify/hooks.py tests/test_hooks.py`
- [x] `uv run python -m tools.skillgen --check`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@ Full release notes with details on each version: [GitHub Releases](https://githu
 ## 0.9.22 (unreleased)
 
 - Fix: a node whose `source_file` is a URL/virtual scheme (`gdoc://`, `s3://`, `http://`, ...) is no longer evicted on the second `graphify update` (follow-up to #2051). The #2051 disk-absence sweep guarded such sources with a literal `"://"` check, but write-side path normalization collapses the double slash (`gdoc://x` becomes `gdoc:/x`), so the guard missed the node on the next run and dropped it into the disk-absence eviction branch. The scheme is now matched tolerantly (and a Windows drive letter like `C:/` is not misread as remote).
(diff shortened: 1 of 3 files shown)


zeriss:
🎉 Awesome first PR!

Really nice improvement. I have a feeling this is going to become part of our daily workflow. Thanks for contributing this—great job!

Issue #1809, opened 2026-07-11
git hooks: post-checkout ignores `GRAPHIFY_SKIP_HOOK`; post-commit builds a rogue graph in worktrees


cdahl86-cyber:
## Summary

Two small gaps in the generated `post-commit` / `post-checkout` git hooks (from `graphify hook install`):

**(a) `post-checkout` ignores `GRAPHIFY_SKIP_HOOK`.** `post-commit` early-exits on `GRAPHIFY_SKIP_HOOK=1` (`hooks.py:269`), but the `post-checkout` template has no such check. So `GRAPHIFY_SKIP_HOOK=1` suppresses commit-triggered rebuilds but **not** branch-switch rebuilds — surprising when scripting around the hooks (e.g. batch branch switches, or tooling that sets the var).

**(b) `post-commit` builds a rogue graph inside worktrees.** With `core.hooksPath` shared across git worktrees, a commit in any worktree fires these hooks. `post-checkout` guards on `[ -d graphify-out ]` and exits when it's absent, but `post-commit` has **no** such guard — so in a worktree with no local `graphify-out/`, it runs `graphify update` with `cwd=worktree` and writes a fresh delta-only `graph.json` into that worktree, which the user never asked for.

## Version

graphifyy 0.9.12

## Suggested fix

In the hook templates (`hooks.py`):

- Add to the **post-checkout** template: `[ "${GRAPHIFY_SKIP_HOOK:-0}" = "1" ] && exit 0`
- Add to the **post-commit** template: `[ -d "graphify-out" ] || exit 0` (matching post-checkout)

Editing the generated `.git/hooks/*` directly works, but is regenerated on reinstall/upgrade, so the fix belongs in the templates.

safishamsi:
Fixed in `5f57c46` (ships in 0.9.14). Both gaps addressed in the hook templates: (a) `post-checkout` now honors `GRAPHIFY_SKIP_HOOK` like `post-commit`; (b) both hooks short-circuit inside a linked worktree. For (b) I went with a worktree check (git-dir != git-common-dir) rather than the `[ -d graphify-out ]` guard you suggested — it targets the actual condition (a linked worktree) instead of the proxy (missing output dir), so it also skips the wasteful rebuild in a worktree that happens to have its own `graphify-out/`. The compare resolves both paths to absolute so the primary checkout isn't false-positived. Verified end-to-end against a real `git worktree`. Thanks @cdahl86-cyber.

PR #2064, opened 2026-07-20
fix(cli): resolve the graph from the primary checkout in linked worktrees (#2008)

### LINKED ISSUE #2008: Auto-resolve graph location from git worktrees (git-common-dir)
**Problem.** The graph (`graphify-out/`) lives in the main checkout. Running `graphify explain` from a git worktree fails hard ("graph file not found"), so every worktree-based workflow needs a wrapper like:

    (cd "$(git rev-parse --path-format=absolute --git-common-dir)/.." && graphify explain "<symbol>")

With AI-agent workflows (Claude Code etc.) increasingly running every task in its own worktree, this is a growing friction point: the workaround has to be carried in project rules and breaks when the cwd disappears mid-session.

**Proposal.** On startup, when no `graphify-out/` is found in the current root, detect a worktree via `git rev-parse --git-common-dir` and fall back to the main checkout's graph (read-only). A `--graph-dir` override would also help.

**Environment.** graphifyy 0.8.36, macOS, ~1.8k-file PHP/JS repo, graph built code-only (no LLM pass).

HerenderKumar:
## Summary

Fixes #2008. The read commands (`query` / `explain` / `path` / `affected`) resolve their default graph to `<GRAPHIFY_OUT>/graph.json` relative to CWD. Run from a linked git worktree — increasingly the default for AI-agent task isolation (Claude Code etc.) — that path does not exist, so the command fails hard with `Graph file not found` and callers have to wrap every invocation in a `git rev-parse --git-common-dir` shim.

## Fix

`cli.py::_default_graph_path()` — used only by the read commands; writers resolve through `graphify.paths` — gains a worktree fallback: when the local graph is absent **and** CWD is inside a linked worktree (`--git-dir` != `--git-common-dir`, the same absolute compare the commit/post-checkout hooks already use for #1809), resolve to the primary checkout's graph.

Read-only and surgical by construction:

- writers (`extract` / `build`) never touch this helper — a write is never redirected;
- an explicit `--graph` still overrides;
- an absolute `GRAPHIFY_OUT` short-circuits (already shared across worktrees);
- an absent primary graph keeps the original "run extract" error;
- the two `git rev-parse` probes run only on the cold path (no local graph), so the common case is untouched.

## Testing

New `tests/test_worktree_graph_path.py` (real `git worktree`, no grammar deps):

- resolves the primary checkout's graph from a linked worktree (red→green verified against `origin/v8`);
- a worktree that built its own graph keeps using it (local wins);
- the primary checkout is unaffected (local default preserved);
- outside a git repo the resolver never raises.

`tests/test_explain_cli.py` + `tests/test_affected_cli.py` pass unchanged.

### DIFF
--- a/graphify/cli.py
+++ b/graphify/cli.py
@@ -82,7 +82,54 @@ _GEMINI_NUDGE_TEXT = (
 
 
 def _default_graph_path() -> str:
-    return str(Path(_GRAPHIFY_OUT) / "graph.json")
+    """Default graph.json for the read commands (query/explain/path/affected).
+
+    Normally ``<GRAPHIFY_OUT>/graph.json`` relative to CWD. When run from a linked
+    git worktree that has no graph of its own, fall back to the primary checkout's
+    graph — worktrees share one build and agent workflows increasingly run each
+    task in its own worktree (#2008). An explicit ``--graph`` still overrides this,
+    and writers (extract/build) resolve through graphify.paths, so a write is never
+    redirected.
+    """
+    local = Path(_GRAPHIFY_OUT) / "graph.json"
+    if local.is_file():
+        return str(local)
+    shared = _worktree_primary_graph()
+    if shared is not None:
+        return shared
+    return str(local)
+
+
+def _worktree_primary_graph() -> str | None:
+    """The primary checkout's ``graph.json`` when CWD is inside a linked git
+    worktree, else ``None``.
+
+    A linked worktree has ``--git-dir`` != the shared ``--git-common-dir`` (whose
+    parent is the primary checkout) — the same absolute-path compare the commit and
+    post-checkout hooks use to skip worktrees (#1809). Returns ``None`` when git is
+    unavailable, we are not in a linked worktree, ``GRAPHIFY_OUT`` is absolute (an
+    absolute override already points every worktree at one graph), or the primary
+    checkout has no graph yet.
+    """
+    if os.path.isabs(_GRAPHIFY_OUT):
+        return None
+    import subprocess
+    try:
+        git_dir = subprocess.run(["git", "rev-parse", "--git-dir"],
+                                 capture_output=True, text=True, timeout=3)
+        common = subprocess.run(["git", "rev-parse", "--git-common-dir"],
+                                capture_output=True, text=True, timeout=3)
+    except (OSError, subprocess.SubprocessError):
+        return None
+    if git_dir.returncode != 0 or common.returncode != 0:
+        return None
+    cwd = os.getcwd()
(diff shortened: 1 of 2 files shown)


graphify/cli.py
```diff
@@ -82,7 +82,54 @@
...
+    local = Path(_GRAPHIFY_OUT) / "graph.json"
+    if local.is_file():
+        return str(local)
+    shared = _worktree_primary_graph()
+    if shared is not None:
+        return shared
+    return str(local)
+
+
+def _worktree_primary_graph() -> str | None:
```
IskandarovFirdavs, line 103: Nice helper function. Would it be worth adding a short docstring explaining why graph.json is the default output path?

tests/test_worktree_graph_path.py
```diff
@@ -0,0 +1,86 @@
+"""#2008: read commands resolve the graph from the primary checkout when they
+are run from a linked git worktree that has no ``graphify-out/`` of its own.
+
+Agent workflows (Claude Code etc.) increasingly run every task in its own
+worktree, where ``graphify explain``/``query``/``path``/``affected`` used to fail
```
IskandarovFirdavs, line 5: Using Path here improves readability. Is there any scenario where _GRAPHIFY_OUT could be missing or invalid? If so, should that be handled explicitly?

graphify/cli.py
HerenderKumar: Thanks !!! Actually both new functions already carry docstrings, If I mention specifically _default_graph_path (lines 85–93) and _worktree_primary_graph (lines 104–113). _default_graph_path explains why <GRAPHIFY_OUT>/graph.json is the default and how the worktree fallback kicks in, and _worktree_primary_graph documents the git-dir vs. git-common-dir check and every case where it returns None. Btw I am happy to extend further if any part seems less explanatory.

tests/test_worktree_graph_path.py
HerenderKumar: Thanks !!! Agreed on Path. On _GRAPHIFY_OUT — it isn't a value that gets passed in while the program runs. It's a fixed setting loaded once at the top of the file (line 14), so it's always a valid string (defaults to graphify-out). If it were ever wrong, the program wouldn't even start — it'd fail right at import, not inside this function. The only tricky case is when it's set to an absolute path, and that's already handled at line 114 (the os.path.isabs check). So there's nothing extra to handle here. That said, I'm happy to add a check if you'd still like one just to be safe.

jecanore:
Merged this on top of v8 into a local dev build alongside a couple other open staleness fixes (#2858, #3108) — needed working worktree resolution for something I'm doing and this was the cleanest fix in flight. Applied clean, no conflicts. Thanks for tackling this one.

Issue #2055, opened 2026-07-20
Scala qualified/dotted type names (`stable_type_identifier`) are invisible to structural extraction — dropped from `references` edges, and from `inherits`/`mixes_in` edges too


sub4biz:
## Summary

Any type referenced by a dotted, package-qualified name — `cats.data.EitherT`, `cats.effect.IO`,
`cats.effect.IOApp` — parses to a distinct `stable_type_identifier` node, which neither the type-reference
walker (`_scala_collect_type_refs`) nor the `extends_clause` base-detection code recognizes. The unqualified
form of the same name (`EitherT`, `IO`, `IOApp`) already works correctly everywhere. This one root cause has
three separate, verified manifestations: a qualified generic base is silently dropped from `references`
edges, a bare qualified type used as a type argument is silently dropped too, and — most severe — a qualified
`extends` base either loses its `inherits` edge entirely or (if it has type arguments) produces one to a
wrongly-keyed node that can never unify with the real class/object it points to.

## Manifestation 1 & 2: qualified types are silently dropped from `references` edges

Isolated from an unrelated, separately-filed gap about file-scope (top-level) definitions losing
`parent_class_nid` entirely, by nesting inside a real class here so `parent_class_nid` is genuinely present:

```scala
class Holder:
  val stack: cats.data.EitherT[cats.effect.IO, String, Int] = null
```

Running this through the actual installed extractor
(`graphify.extractors.engine._extract_generic` + `graphify.extract._SCALA_CONFIG`, called directly):

```
NODE Holder
NODE String
NODE Int
EDGE <file> -[contains]-> Holder
EDGE Holder -[references]-> String   (generic_arg)
EDGE Holder -[references]-> Int      (generic_arg)
```

`String`/`Int` (the plain generic arguments) resolve correctly. `EitherT` (the generic base) and `IO` (a
qualified type argument with no generics of its own) are both **completely absent** — not a single edge or
node for either.

```
generic_type
  stable_type_identifier   <- base: "cats.data.EitherT" — the qualified generic base
  type_arguments
    stable_type_identifier <- "cats.effect.IO" — bare qualified type argument
    type_identifier "String"
    type_identifier "A"
```

`_scala_collect_type_refs`'s `generic_type` handling only accepts a bare `type_identifier` as the base:

```python
if t == "generic_type":
    base = node.child_by_field_name("type")
    if base is None:
        for c in node.children:
            if c.type == "type_identifier":
                base = c
                break
[...]

Issue #2054, opened 2026-07-20
Scala top-level (file-scope) `val`/`var` definitions lose all type references — the same `parent_class_nid` reset already fixed once for Python (#1050)


sub4biz:
## Summary

Scala 3's headline "no boilerplate wrapper" feature — writing `val`/`var`/`given`/`def` directly at file
scope, with no enclosing `object`/`class`/`trait` — silently drops every type reference a top-level `val`/
`var` makes. Confirmed end-to-end with the real extractor, not just by reading the dispatch code: a top-level
`val greeting: String = ...` produces zero node and zero edge, while the exact same definition nested one
level inside any container resolves correctly. Functions are unaffected, since function parameter/return-type
references anchor to the function's own node id, not to an enclosing container. The root cause is a generic
mechanism in the walker's default recursion path — the same mechanism that already caused and was fixed once
for a Python-specific case.

## The gap, demonstrated end-to-end with the real extractor

```scala
import cats.effect.{Async, IO}

def greet(name: String): String = s"Hello, $name"

val greeting: String = greet("World")

given Ordering[Int] = Ordering.Int

def program[F[_]: Async]: F[Unit] = Async[F].unit
```

Running this through the actual installed extractor (`graphify.extractors.engine._extract_generic` with
`graphify.extract._SCALA_CONFIG`, called directly — not reimplemented) produces:

```
NODE greet()
NODE String
NODE program()
NODE F
NODE Unit
EDGE <file> -[contains]-> greet()
EDGE greet() -[references]-> String   (parameter_type)
EDGE greet() -[references]-> String   (return_type)
EDGE <file> -[contains]-> program()
EDGE program() -[references]-> F      (return_type)
EDGE program() -[references]-> Unit   (generic_arg)
```

`greet`/`program` — both top-level functions — get complete, correct node and reference output. `greeting`
(the top-level `val`) and the top-level `given` produce **no node and no edge of any kind** — not even the
`String` reference `greeting`'s type annotation makes shows up anywhere in the output.

## Root cause: the same `parent_class_nid` reset already documented as a bug source once, for Python

The Scala-specific `val`/`var` reference-extraction branch requires `parent_class_nid` to be truthy:

```python
if (config.ts_module == "tree_sitter_scala"
        and t in ("val_definition", "var_definition")
        and parent_class_nid):
    type_node = node.child_by_field_name("type")
    ...
```

The main walker's final fallback, reached whenever a node type isn't handled by anything more specific,
[...]

Issue #2042, opened 2026-07-19
Scala structural extraction misses trait/given/extension/enum containers AND all abstract method declarations (`function_declaration`) — 5 of 8 relevant node types missing from `_SCALA_CONFIG`


sub4biz:
## Summary

`_SCALA_CONFIG` in `extract.py` defines which tree-sitter node types count as a "class" or a "function"
boundary for Scala. It only lists:

```python
class_types=frozenset({"class_definition", "object_definition"})
function_types=frozenset({"function_definition"})
function_boundary_types=frozenset({"function_definition"})
```

The installed `tree_sitter_scala` grammar (0.26.0) parses `trait`, `given`, `extension`, and `enum` as their
own distinct top-level node types (`trait_definition`, `given_definition`, `extension_definition`,
`enum_definition`), and splits methods into `function_definition` (has a body) vs. `function_declaration`
(signature only — abstract methods, trait members without a default implementation). None of the four
container types, and neither `function_declaration`, appear anywhere in `_SCALA_CONFIG`. Confirmed by
dumping the grammar's own parse tree directly (see Reproduction) — this is not read off `tags.scm`/
`highlights.scm` alone.

Since `walk()`'s class-node branch, `contains`/`method` edge creation, and `parent_class_nid` propagation to
children all gate on `t in config.class_types` / `t in config.function_types`, none of these five node
shapes are recognized:

- A `trait_definition`, `given_definition`, `extension_definition`, or `enum_definition` node falls into the
  generic "default: recurse" branch (`engine.py:3569-3571`), which explicitly **resets
  `parent_class_nid=None`** before recursing into children. Their members are walked as if they were bare
  top-level code.
- A `function_declaration` (abstract method, no `=` body) never matches `function_types`/
  `function_boundary_types` at all, **in any container** — including inside an ordinarily well-handled
  `class`/`object`/`trait` — so it produces no node under any circumstance.

`trait` is Scala's primary unit of interface/typeclass abstraction, and `given`/`extension` are the idiomatic
Scala 3 replacement for `implicit`. Any codebase built around typeclass-style design (the dominant pattern in
the cats/cats-effect ecosystem — `Functor[F[_]]`, `Concurrent[F]`, `MonadError`, custom domain typeclasses,
`given` instances providing them) has its central abstraction layer almost entirely absent from the
structural graph. There is no LLM/semantic fallback for code files, so none of this is backfilled.

## Root cause

[...]

PR #2066, opened 2026-07-20
fix(scala): walk type_definition right-hand sides for type references

### LINKED ISSUE #2049: Scala `type_definition` right-hand sides are never scanned for type references — a one-line dispatch fix covers type aliases and `opaque type`, `match type` needs one more small addition
## Summary

`type_definition` (plain type aliases, `opaque type`, `match type`) never has its right-hand side walked for
type references at all, in any context — not just "no node for the alias itself" (an already-accepted,
narrower tradeoff shared with `val`/`given` bindings), but zero `references` edges of any kind, in contrast to
`val_definition`/`var_definition`, which do get their type annotation walked when inside a class. The fix for
plain aliases and `opaque type` is a one-line dispatch widening that reuses the existing `val`/`var` code
path verbatim. `match type` needs one more, independent addition on top: the type-reference walker's own
recursion whitelist doesn't know about `match_type`/`type_case_clause` nodes, so even a correctly-dispatched
call on a match type's right-hand side would still resolve nothing.

## The asymmetry with `val`/`var`

```scala
trait Repo:
  type Elem[X] = X match
    case String => Char
    case Array[t] => t

  opaque type Id = Long

  type Alias = List[Int]
```

```
trait_definition
  identifier "Repo"
  template_body
    type_definition            <- Elem: ['type','type_identifier','type_parameters','=','match_type']
    type_definition            <- Id:   ['opaque_modifier','type','type_identifier','=','type_identifier']
    type_definition            <- Alias:['type','type_identifier','=','generic_type']
```

All three sit at the exact same sibling level inside `template_body` that a `val_definition` would occupy.
The existing Scala-specific dispatch only fires for the latter:

```python
if (config.ts_module == "tree_sitter_scala"
        and t in ("val_definition", "var_definition")
        and parent_class_nid):
    type_node = node.child_by_field_name("type")
    if type_node is not None:
        line = node.start_point[0] + 1
        refs: list[tuple[str, str]] = []
        _scala_collect_type_refs(type_node, source, False, refs)
        for ref_name, role in refs:
            ctx = "generic_arg" if role == "generic_arg" else "field"
            target_nid = ensure_named_node(ref_name, line)
            if target_nid != parent_class_nid:
                add_edge(parent_class_nid, target_nid, "references", line, context=ctx)
```

`type_definition` is not in the `t in (...)` check, so `type Alias = List[Int]` loses the `List`/`Int`
references completely — a `val x: List[Int] = ...` in the same position would keep them.

## The fix is a one-line widening, not a new code path — `type_definition` already has a `"type"` field

The grammar assigns field names to `type_definition`'s children exactly the way it does for `val_definition`
— including a `"type"` field for the right-hand side, the same field name the existing `val`/`var` code above
already reads with `node.child_by_field_name("type")`. Enumerating the actual field assigned to each child,
for both RHS-bearing definitions and RHS-less abstract type members (`type Foo`, `type Bar <: Serializable`,
`type Baz >: Nothing <: AnyRef` — deferred type members with no `=` at all, legal and common in traits):

```scala
trait Repo:
  type Foo
  type Bar <: Serializable
  type Baz >: Nothing <: AnyRef
```

```
Elem: 'type_identifier' <- field "name", 'type_parameters' <- field "type_parameters", 'match_type' <- field "type"
Id:   'type_identifier' <- field "name", 'type_identifier' (RHS) <- field "type"
Alias:'type_identifier' <- field "name", 'generic_type' <- field "type"
Foo:  'type_identifier' <- field "name"                      (no "type" field at all)
Bar:  'type_identifier' <- field "name", 'upper_bound' <- field "bound"   (not "type")
Baz:  'type_identifier' <- field "name", 'lower_bound'/'upper_bound' <- field "bound"
```

Calling `child_by_field_name("type")` directly on all six confirms this without any special-casing:

```
Elem -> match_type      Foo -> None
Id   -> type_identifier  Bar -> None
Alias -> generic_type    Baz -> None
```

The three abstract members correctly return `None` — their RHS-equivalent (the bound) lives under field
`"bound"`, not `"type"`, so the exact same one-line call already used for `val`/`var` needs no additional
logic to skip them safely. No manual token-scanning or positional-child heuristic is needed for this part —
reusing `node.child_by_field_name("type")` verbatim, just with `"type_definition"` added to the `t in (...)`
check, is the complete fix for aliases and `opaque type`.

## Confirming the walker handles the RHS shapes it now reaches

```
Elem  (rhs type: match_type)      -> refs = []
Id    (rhs type: type_identifier) -> refs = [('Long', 'type')]
Alias (rhs type: generic_type)    -> refs = [('List', 'type'), ('Int', 'generic_arg')]
```

`Id` and `Alias` resolve correctly the moment `_scala_collect_type_refs` is called on them — for these two,
the one-line dispatch widening above is the complete fix. `Elem` (the `match type`) comes back empty even
when the walker is called on it directly, which rules out the dispatch fix alone as complete for match
types.

## The second, independent bug: `match_type`/`type_case_clause` aren't in the walker's recursion whitelist

```python
def _scala_collect_type_refs(node, source: bytes, generic: bool, out: list[tuple[str, str]]) -> None:
    ...
    if t in ("compound_type", "infix_type", "function_type", "tuple_type",
             "annotated_type", "projected_type"):
        for c in node.children:
            if c.is_named:
                _scala_collect_type_refs(c, source, generic, out)
```

`match_type` and `type_case_clause` are absent from this whitelist, so even a correctly-dispatched call on a
`match type`'s right-hand side recurses into nothing and returns an empty list — confirmed above (`Elem`
returns `[]`). This is independent of the dispatch fix: widening only the `t in (...)` check would make
`Alias`/`Id` resolve but `Elem` would still silently produce zero references.

## Proposed diff — both layers together

```python
# 1. dispatch: one-line widening, reuses the existing val/var code path verbatim —
#    child_by_field_name("type") already resolves correctly for type_definition,
#    including returning None for RHS-less abstract type members.
if (config.ts_module == "tree_sitter_scala"
        and t in ("val_definition", "var_definition", "type_definition")
        and parent_class_nid):
    type_node = node.child_by_field_name("type")
    ...  # unchanged from here — same ensure_named_node/add_edge calls already in place

# 2. walker: teach _scala_collect_type_refs about match_type/type_case_clause
if t in ("match_type", "type_case_clause"):
    for c in node.children:
        if c.is_named:
            _scala_collect_type_refs(c, source, generic, out)
```

**Verified against a real parsed tree, running both patches together (not just reasoned through)** — all six
`type_definition`s from the two shapes above (three with a RHS, three abstract members with none), walked
with the widened dispatch and patched walker:

```
Elem  -> [('X', 'type'), ('String', 'type'), ('Char', 'type'), ('Array', 'type'), ('t', 'generic_arg'), ('t', 'type')]
Id    -> [('Long', 'type')]
Alias -> [('List', 'type'), ('Int', 'generic_arg')]
Foo   -> []   (abstract member, no "type" field — correctly produces no references, not an error)
Bar   -> []   (abstract member with only a "bound" field — also correctly out of scope here)
Baz   -> []   (abstract member with two "bound" fields — same)
```

All three RHS-bearing definitions now resolve, and all three RHS-less abstract members correctly stay empty
instead of producing garbage. `X`/`t` (the match type's own type-parameter/binder names, not external types)
are also picked up — the same kind of reference a generic type parameter used anywhere else in this extractor
already produces (resolves through `ensure_named_node`'s bare-stub fallback), not a new class of noise
introduced by this fix.

## What this deliberately does not propose

- Making `type_definition` a node/container of its own — that's the already-covered, separate `val`/`given`
  value-binding-invisibility tradeoff. This proposal is scoped to reference-edge extraction only: the binding
  itself stays invisible, exactly like `val`, but the type references it makes are no longer silently
  dropped.
- Any change to `class_types` (trait/enum container support) — independent of this finding.
- Extracting anything from abstract type members' bounds (`type Bar <: Serializable`) — their reference
  (`Serializable`) lives under a different field (`"bound"`, not `"type"`) and is out of scope here; flagged
  as a possible narrow follow-up, not folded into this proposal to keep the diff minimal and precedent-driven.

## Environment

- graphifyy 0.9.20, tree_sitter_scala 0.26.0
- All six `type_definition` shapes (three RHS-bearing, three abstract/RHS-less) were dumped directly against
  the installed grammar (`tree_sitter.Parser` + `tree_sitter_scala.language()`), checked for absence of
  `ERROR`/`is_missing` nodes. Field assignments were read via the grammar's own `field_name_for_id`/
  `field_name_for_child` introspection, not guessed from node shape. The current-behavior and
  patched-behavior traces above were both produced by running the installed `_scala_collect_type_refs`
  (imported from the installed package) and a manually-patched copy against the same real parsed nodes, not
  inferred from reading the source.

mdshzb04:
Fixes #2049

`type_definition` nodes (plain aliases, `opaque type`, match types) sit at the exact same `template_body` sibling level as `val_definition`/`var_definition`, but the Scala-specific dispatch in `_extract_generic` only fired for the latter — so alias right-hand sides produced zero `references` edges of any kind, as reported.

## Changes

Two small edits in `graphify/extractors/engine.py`:

1. **Widen the dispatch** to include `type_definition`. Verified against tree-sitter-scala that the RHS lives under the same `type` field a val/var annotation uses (for plain aliases, `opaque type`, and match types alike), so the existing walk is reused verbatim — exactly the one-line widening the issue suggested.
2. **Teach `_scala_collect_type_refs` about `match_type` / `type_case_clause`**. As the issue predicted, the dispatch alone isn't enough for match types: the walker's recursion whitelist didn't know these node types, so a correctly-dispatched match-type RHS still resolved nothing. Adding them to the existing recursion tuple lets case patterns and results resolve.

## Before / after

```scala
class Repo {
  type Alias = List[Int]
  opaque type Id = Long
  type Elem[X] = X match { case String => Char; case Array[t] => t }
}
```

Before: `Repo` has no outgoing `references` edges.
After: `references` edges to `List`, `Int`, `Long`, `String`, `Char`, `Array`, `t`, `X` — matching what an equivalent `val` annotation already produced.

## Tests

Extended `tests/fixtures/sample.scala` with a `TypeAliases` class covering all three shapes, plus four new tests asserting the alias (`Map`/`HttpClient`), opaque (`Long`), and match-type case (`Option`/`ListBuffer`) references. Full run: `tests/test_languages.py` + `tests/test_builtin_global_type_refs.py` → **324 passed, 13 skipped**.

Out of scope (kept separate deliberately): file-scope/trait containers (#2042, #2054) and qualified `stable_type_identifier` names (#2055) — this PR only closes the `type_definition` dispatch gap so it composes cleanly with fixes for those.

### DIFF
--- a/graphify/extractors/engine.py
+++ b/graphify/extractors/engine.py
@@ -3075,12 +3075,12 @@ def _extract_generic(
                 type_table[prop_name] = prop_type
             return
 
+        # type_definition covers plain aliases (`type Alias = List[Int]`),
+        # `opaque type`, and match types — the right-hand side sits under the
+        # same `type` field a val/var annotation uses, so the walk is shared.
         if (config.ts_module == "tree_sitter_scala"
                 and t in ("val_definition", "var_definition", "type_definition")
                 and parent_class_nid):
-            # type_definition covers plain aliases (`type Alias = List[Int]`),
-            # `opaque type`, and match types — the right-hand side sits under the
-            # same `type` field a val/var annotation uses, so the walk is shared.
             type_node = node.child_by_field_name("type")
             if type_node is not None:
                 line = node.start_point[0] + 1
--- a/graphify/extractors/engine.py
+++ b/graphify/extractors/engine.py
@@ -969,7 +969,8 @@ def _scala_collect_type_refs(node, source: bytes, generic: bool, out: list[tuple
                         _scala_collect_type_refs(arg, source, True, out)
         return
     if t in ("compound_type", "infix_type", "function_type", "tuple_type",
-             "annotated_type", "projected_type"):
+             "annotated_type", "projected_type", "match_type",
+             "type_case_clause"):
         for c in node.children:
             if c.is_named:
                 _scala_collect_type_refs(c, source, generic, out)
@@ -3075,8 +3076,11 @@ def _extract_generic(
             return
 
         if (config.ts_module == "tree_sitter_scala"
-                and t in ("val_definition", "var_definition")
+                and t in ("val_definition", "var_definition", "type_definition")
                 and parent_class_nid):
+            # type_definition covers plain aliases (`type Alias = List[Int]`),
+            # `opaque type`, and match types — the right-hand side sits under the
+            # same `type` field a val/var annotation uses, so the walk is shared.
             type_node = node.child_by_field_name("type")
             if type_node is not None:
                 line = node.start_point[0] + 1
--- a/tests/test_languages.py
+++ b/tests/test_languages.py
@@ -765,6 +765,24 @@ def test_scala_method_return_type_context():
     assert ("create", "HttpClient") in _edge_labels(r, "references", "return_type")
 
 
+def test_scala_type_alias_rhs_references():
+    r = extract_scala(FIXTURES / "sample.scala")
+    assert ("TypeAliases", "Map") in _edge_labels(r, "references", "field")
+    assert ("TypeAliases", "HttpClient") in _edge_labels(r, "references", "generic_arg")
+
+
+def test_scala_opaque_type_rhs_references():
+    r = extract_scala(FIXTURES / "sample.scala")
+    assert ("TypeAliases", "Long") in _edge_labels(r, "references", "field")
+
+
+def test_scala_match_type_case_references():
+    r = extract_scala(FIXTURES / "sample.scala")
+    labels = _edge_labels(r, "references", "field")
+    assert ("TypeAliases", "Option") in labels
+    assert ("TypeAliases", "ListBuffer") in labels
+
+
 def test_scala_call_edges_have_call_context():
     r = extract_scala(FIXTURES / "sample.scala")
     call_edges = _edges_with_relation(r, "calls")
--- a/tests/fixtures/sample.scala
+++ b/tests/fixtures/sample.scala
@@ -27,3 +27,12 @@ object HttpClientFactory {
     new HttpClient(Config(baseUrl, 30))
   }
 }
+
+class TypeAliases {
+  type Routes = Map[String, HttpClient]
+  opaque type ClientId = Long
+  type Unwrap[X] = X match {
+    case Option[t] => t
+    case ListBuffer[t] => t
+  }
+}


graphify/extractors/engine.py
```diff
@@ -969,7 +969,8 @@ def _scala_collect_type_refs(node, source: bytes, generic: bool, out: list[tuple
                         _scala_collect_type_refs(arg, source, True, out)
         return
```
IskandarovFirdavs, line 970: Looks good overall. Could you briefly explain why True is passed here instead of using the existing generic variable? It would help future readers understand the intent.

graphify/extractors/engine.py
```diff
@@ -3075,8 +3076,11 @@ def _emit_java_parent_type(type_node, rel: str, at_line: int) -> None:
             return
```
IskandarovFirdavs, line 3076: This change seems reasonable. Would adding a short comment above this condition make the special handling for Scala type references easier to understand?

tests/test_languages.py
```diff
@@ -765,6 +765,24 @@ def test_scala_method_return_type_context():
...
+    assert ("TypeAliases", "Long") in _edge_labels(r, "references", "field")
+
+
+def test_scala_match_type_case_references():
+    r = extract_scala(FIXTURES / "sample.scala")
+    labels = _edge_labels(r, "references", "field")
+    assert ("TypeAliases", "Option") in labels
+    assert ("TypeAliases", "ListBuffer") in labels
+
+
```
IskandarovFirdavs, line 785: Nice improvement. Have you tested this against nested or more complex Scala type definitions to ensure there are no regressions?

graphify/extractors/engine.py
mdshzb04: This line is pre-existing and only appears as context in this diff. It’s where the walker descends into the children of a type_arguments node, so everything below that point is in generic-argument position regardless of the outer generic value. Wrapper nodes instead pass generic through unchanged since they don't change that context.

graphify/extractors/engine.py
mdshzb04: Good point. There’s already a comment just inside the condition explaining the type_definition handling, but I agree it would be clearer above the condition. I’ll move it there.

tests/test_languages.py
mdshzb04: Yes. The committed fixture covers nested type arguments with Map[String, HttpClient] and match-type cases with type-variable binding. I also tested deeper nesting locally with Map[String, List[Option[HttpClient]]] and a match type over Either[Throwable, Vector[Int]]; references resolved correctly at each depth. The full relevant test suite also passes with 324 passed and 13 skipped. Happy to add the deeper-nesting case to the fixture if you'd like it covered explicitly.

Issue #1734, opened 2026-07-08
extract requires an LLM API key even if I’m not interested in doc extraction


ottens:
Extract fails on a repo that contains non-code without access to an LLM. I’d like it to just index the code and ignore the rest. The workaround is to add a graphifyignore, but since it is not clear what is and what is not considered code, creating that list is a very onerous trial and error. Especially since the list of what is not code is much bigger than what is code and the ignore file is supposed to define what is not code. Adding a simple override or a partial failure like ‘x code files extracted, with y non code files failed due to no llm api key present’ might be a more friendly failure? Or happy to run it with a —code-only flag or something to that effect.

safishamsi:
Fixed on `v8` (next release) — added the `--code-only` flag you suggested.

`graphify extract . --code-only` indexes the code via local AST (no API key needed) and skips the doc/paper/image files entirely, reporting what it skipped:
```
[graphify extract] --code-only: skipping 12 non-code file(s) (9 docs, 2 papers, 1 images) — no LLM extraction
[graphify extract] found 340 code, 0 docs, 0 papers, 0 images
```
No more hand-crafting a .graphifyignore of everything non-code. And if you forget the flag on a mixed repo with no key, the error now points you straight at it ("Or pass --code-only to index just the code ..."). Regression tests cover both the flag succeeding keyless and the error hint. Thanks for the clear write-up — the trial-and-error-ignore pain was exactly the right thing to call out.

PR #2075, opened 2026-07-21
docs(cli): surface --code-only in the extract usage and README (#2071)

### LINKED ISSUE #2071: --code-only is missing from --help and the README
`--code-only` is a genuinely useful flag and it is currently discoverable only by triggering an error. It appears in neither the README nor `graphify --help`.

I evaluated graphify for a documentation-heavy repository where the docs contain infrastructure hostnames that must not leave the machine, so "can this run without any network call" was the deciding question. The README says code is parsed locally with tree-sitter and that docs go through an LLM, but it does not say whether the LLM path can be turned off. `graphify --help` lists `--watch`, `--wiki`, `--no-label`, `--no-viz` and others, and no code-only option. On that evidence I concluded no such flag existed and was about to drop the tool.

It exists. I found it by running with no API key set:

```
error: no LLM API key found (2 doc/paper/image file(s) need semantic extraction).
Set GEMINI_API_KEY or ... or pass --backend. A code-only corpus needs no key.
Or pass --code-only to index just the code (local AST, no key) and skip the non-code files.
```

That error message is excellent. It names the file count, lists the options, and gives the escape hatch. The problem is that you only see it if you run without a key, and someone evaluating the tool on a privacy constraint will often decide before that point.

`--code-only` worked exactly as described. 34 Python files, no key set, no network, 927 nodes and 2,016 edges.

Two suggestions, both small:

1. Add `--code-only` to `graphify --help`, alongside `--no-label` and `--no-viz`.
2. Add one line to the README's docs table, something like "pass `--code-only` to skip all non-code files and run with no API key and no network calls".

Worth saying that the underlying behaviour is already right. Refusing to run rather than silently falling back to a remote call is the correct default, and grepping the source turned up no telemetry endpoint. This is a documentation gap in front of a good design, not a design problem.

HerenderKumar:
Fixes #2071.

`--code-only` has been in the top-level `graphify --help` since #1734, but two other surfaces still never mention it:

- The `graphify extract` usage string (printed on an empty/bad invocation) lists `--no-cluster`, `--no-gitignore`, `--postgres`, `--cargo`… but **not** `--code-only`.
- The README documents the *concept* ("a code-only corpus requires no API key") but never the actual **flag**.

Someone evaluating graphify under a "can this run with no network call?" constraint reads the extract usage / README first, so they can reasonably conclude the flag doesn't exist — which is exactly what the reporter almost did before finding it by triggering the no-key error.

## Changes
- `graphify/cli.py` — add `[--code-only]` to the extract usage string.
- `README.md` — name `--code-only` in the Privacy section, and add it to the command reference.
- `tests/test_extract_code_only_cli.py` — new `test_extract_usage_advertises_code_only` asserting the usage advertises the flag.

## Verification
- `tests/test_extract_code_only_cli.py` — 9 passed (incl. the new test)
- `tests/test_extract_cli.py` — 25 passed

### DIFF
--- a/README.md
+++ b/README.md
@@ -530,7 +530,7 @@ These are only needed for **headless / CI extraction** (`graphify extract`). Whe
 
 ## Privacy
 
-- **Code files** — processed locally via tree-sitter. Nothing leaves your machine. A code-only corpus requires no API key — `graphify extract` runs fully offline.
+- **Code files** — processed locally via tree-sitter. Nothing leaves your machine. A code-only corpus requires no API key — `graphify extract` runs fully offline. On a mixed repo, add `--code-only` to index just the code and skip the docs/PDFs/images that would otherwise need an LLM.
 - **Video / audio** — transcribed locally with faster-whisper. Nothing leaves your machine.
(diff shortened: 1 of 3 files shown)


safishamsi:
Thanks @HerenderKumar. This shipped in v0.9.24. Your commit was cherry-picked onto v8 (bb8f5bd) to keep your authorship, with a follow-up clarifying that --code-only is an extract flag rather than a skill flag. Closed-unmerged here, but it is in the release: https://github.com/Graphify-Labs/graphify/releases/tag/v0.9.24

Issue #829, opened 2026-05-12
`build_from_json(directed=True)` is not reachable from the CLI, and `serve.py` BFS ignores edge direction


lynn0315:
**graphify version**: 0.7.15

---

## Symptom

Call graphs are inherently directed (`caller → callee`), but `graphify path` / `query` traverse both directions indistinguishably. `graphify path "A" "B"` can return a "valid" reverse path even when no forward dependency chain exists.

---

## Suspected root cause

Three places combine to make directed mode effectively unreachable:

1. **`build.py:74` — `build_from_json` already supports `directed=True`** (docstring: *"directed=False (default) produces an undirected Graph for backward compatibility"*).

2. **No `--directed` flag on the CLI** — `__main__.py:1670-1671` only reads the `directed` field from an existing `graph.json`. `update` / `extract` cannot pass the option through on a fresh build:
   ```python
   _directed = bool(_raw.get("directed", False))
   G = build_from_json(_raw, directed=_directed)
   ```

3. **`serve.py:147` `_dfs` (and `_bfs` at 131) iterate `G.neighbors(node)`** — on a `DiGraph`, this returns both successors and predecessors, so even a directed graph is traversed in both directions.

---

## Reproducer

Every `graph.json` produced by `graphify update .` has `"directed": false`. `graphify path "Caller" "Callee"` and `graphify path "Callee" "Caller"` return symmetric paths.

---

## Possible directions

- Expose `--directed` on `update` / `extract` / `init` (forward to the existing `build.py` option)
- In `serve.py`, branch on `G.is_directed()` and use `successors` for `DiGraph`
- Optionally, a query-time `--reverse` flag (predecessors traversal) for "who calls X?" queries

`cluster.py` already calls `to_undirected()` before Louvain/Leiden, so community detection is unaffected.

---

## Environment

Production Android browser fork (Chromium-derived), Java + Kotlin, 19,668 nodes / 50,881 edges (0.7.15). Full environment in the companion issue (substring `_find_node`, https://github.com/safishamsi/graphify/issues/828).

---

## Impact

`graphify path` / `query` cannot reliably answer "is there a forward dependency chain from A to B?" — the most common path-search question. The directed code path today is half-implemented: build supports it, but serve and CLI don't.

safishamsi:
[...]

PR #2080, opened 2026-07-21
fix: preserve calls edge direction in graphify query output

Yyunozor:
## Summary

`graphify query` can print a `calls` edge backwards — `callee --calls--> caller` — depending on which endpoint the query term matches. The graph on disk is correct; only the query CLI's text output is wrong.

## Reproduction

```
alpha.py:
    from beta import process_payload
    def run_pipeline(payload):
        return process_payload(payload)   # alpha calls beta

beta.py:
    def process_payload(payload):
        return {"status": "ok", "payload": payload}
```

```
$ graphify query "run_pipeline"      # seeded on the CALLER
EDGE run_pipeline() --calls [EXTRACTED context=call]--> process_payload()   # correct

$ graphify query "process_payload"   # seeded on the CALLEE
EDGE process_payload() --calls [EXTRACTED context=call]--> run_pipeline()   # backwards
```

`graph.json` has the correct `source`/`target` for this edge in both runs.

## Root cause

`query` loads the graph as an undirected `nx.Graph` — deliberately, so BFS/DFS can reach both callers and callees of the seed — but `_subgraph_to_text` renders each edge as `EDGE {u} --> {v}` in traversal visit order, which on an undirected graph isn't the stored `(source, target)`. Seeding on the callee visits the callee first, so the edge prints backwards.

Forcing `directed=True` on load (what `path` and `explain` do) isn't the right fix here: on a `DiGraph`, `G.neighbors()` returns successors only, so a query seeded on a leaf/sink node finds zero neighbors instead of its callers — a recall regression. #829 fixed this bug class for `path`, `explain` and the MCP tools, but the `query` CLI kept the backwards rendering.

## Fix

Same `_src`/`_tgt` pattern `build.py` already uses when direction has to survive undirected storage:

- `graphify/cli.py`: stash each link's on-disk `source`/`target` on its edge data as `_src`/`_tgt`.
- `graphify/serve.py` (`_subgraph_to_text`): render `EDGE` lines from `_src`/`_tgt` when present, falling back to `(u, v)`.

Traversal is unchanged (same node counts per seed before and after); no schema change; `path`, `explain` and the MCP tools untouched.

## Tests / validation

Two regression tests in `tests/test_query_cli.py` seed the same `calls` edge from both endpoints; the callee-seeded one fails on the prior code with the exact symptom above.

```
uv run pytest tests/ -q
3523 passed, 3 skipped, 0 failed
```

Parent commit (`abff1b1`), same environment: 3521 passed, 0 failed — the delta is exactly the 2 tests this PR adds. `ruff check` clean.

### DIFF
--- a/graphify/cli.py
+++ b/graphify/cli.py
@@ -818,6 +818,21 @@ def dispatch_command(cmd: str) -> None:
             _raw = _json.loads(gp.read_text(encoding="utf-8"))
             if "links" not in _raw and "edges" in _raw:
                 _raw = dict(_raw, links=_raw["edges"])
+            # `query` deliberately keeps the graph undirected (unlike `path` /
+            # `explain`, which force directed=True): BFS/DFS here must explore
+            # both callers and callees of the seed node to build useful
+            # context, and forcing a DiGraph would make G.neighbors() return
+            # successors only, silently dropping every caller-side result for
+            # a seed with no outgoing edges. Direction is instead preserved
+            # per-edge below (mirrors graphify/build.py's _src/_tgt pattern)
+            # so the *rendering* stays correct without narrowing traversal.
+            _raw = dict(
+                _raw,
+                links=[
+                    {**link, "_src": link.get("source"), "_tgt": link.get("target")}
+                    for link in _raw.get("links", [])
+                ],
+            )
             try:
                 G = json_graph.node_link_graph(_raw, edges="links")
             except TypeError:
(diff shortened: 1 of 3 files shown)


Yyunozor:
Landed directly on v8 as 4ace951, with the extra guard and tests in 53d0f88 — closing this as superseded. Thanks for the quick review and the hardening!

PR #2085, opened 2026-07-21
docs(readme): troubleshoot an older graphifyy shadowing `uv run` (#1540)

### LINKED ISSUE #1540: UX: `uv run --with graphifyy python -m graphify` silently loads an older system-installed version instead of the uv-managed one
## Summary

When a user has an older version of `graphifyy` installed in their system Python alongside a newer version managed by uv tools, `uv run --with graphifyy python -m graphify` silently loads the older one. There is no error — everything appears to work — but `OPENAI_BASE_URL` and other env-var overrides are ignored, causing requests to go to `api.openai.com` instead of the configured endpoint.

## Environment

- OS: Windows 11, Python 3.13 (system)
- graphifyy 0.9.1 (installed via `uv tool install graphifyy`)
- graphifyy 0.8.13 (older version installed in system Python via `pip install graphifyy`)

## Reproduction

1. Install `graphifyy` via uv tools: `uv tool install graphifyy`
2. Have an older version of `graphifyy` installed in system Python (e.g. `pip install graphifyy` at some point in the past). Note: `pip show graphify` reports "not found" — pip registers it under the package name `graphifyy`, not the module name.
3. Set env vars: `OPENAI_BASE_URL=https://api.groq.com/openai/v1` and `OPENAI_API_KEY=<groq_key>`
4. Run: `uv run --with graphifyy python -m graphify . --backend openai --token-budget 800 --out out/`

**Expected:** requests go to `api.groq.com`  
**Actual:** requests go to `api.openai.com` → 401 "Incorrect API key"

## Root cause

`uv run` uses the system Python and does not fully isolate from its site-packages. The `--with graphifyy` flag adds graphifyy to the ephemeral environment, but if system Python already has `graphify` in its site-packages, Python finds it first (earlier in `sys.path`).

Confirm with:
```python
import graphify.llm as llm
print(llm.__file__)
# Wrong: C:\Users\...\AppData\Local\Programs\Python\Python313\Lib\site-packages\graphify\llm.py
# Right: C:\Users\...\AppData\Roaming\uv\tools\graphifyy\Lib\site-packages\graphify\llm.py
```

The loaded module is graphifyy 0.8.13 — an older version that does not honor `OPENAI_BASE_URL`. Confirmed: `BACKENDS["openai"]["base_url"]` is always `https://api.openai.com/v1` regardless of the env var.

## Why it's silent

- graphify runs without error — AST extraction on code files succeeds
- Semantic extraction fails with 401 (wrong endpoint + wrong key type)
- graphify warns "chunk N/N failed" but exits with code 0
- The extracted graph contains only AST nodes (code files), zero semantic nodes from docs
- The user sees no obvious error and may assume the run succeeded

There is one partial signal already present: graphify prints a version mismatch warning when the skill file and the loaded package differ:

```
warning: skill is from graphify 0.9.1, package is 0.8.13
```

This is the actual fingerprint of the problem — the skill was written for 0.9.1 but the running module is 0.8.13 from the system Python install. The warning is correct, but a user seeing it is likely to dismiss it as a stale skill rather than as evidence that a different installation was loaded.

## Workaround

Use the installed executable directly instead of `uv run --with graphifyy`:

```bash
# Instead of:
uv run --with graphifyy python -m graphify . --backend openai ...

# Use (as documented in README) — uvx names the source package explicitly:
uvx --from graphifyy graphify extract . --backend openai ...
```

For DeepSeek backends (`--backend deepseek`), the bug was not observed in testing — DeepSeek extractions succeeded even when loaded via `uv run`. The exact mechanism was not confirmed.

## Relationship to existing documentation

The package author has documented a related case in the README:

> **Running with `uvx` / `uv tool run` instead of installing?** Name the package, not the command: `uvx --from graphifyy graphify install`. Plain `uvx graphify …` fails (`No solution found … no versions of graphify`) because `uv tool run` reads the first word as a *package*, and the package is `graphifyy` — the `graphify` command lives inside it.

The documented case (`uvx graphify` → package not found) **fails loudly** — the user gets an explicit error and can fix it.

The case in this issue is different and **fails silently**:
- `uv run --with graphifyy python -m graphify` — no error, graphify runs and produces output
- The wrong `graphify` module is loaded from system Python site-packages
- `OPENAI_BASE_URL` is ignored, requests go to `api.openai.com`
- The only symptom is a 401 that looks like a key problem, not a routing problem

The existing documentation does not cover this scenario.

## Suggested fix

Add a path check to `__main__.py` or the CLI entry point:

```python
import graphify
import os
import sys
from importlib.metadata import distribution

try:
    dist = distribution("graphifyy")
    expected = os.path.normpath(str(dist.locate_file("graphify")))
    loaded = os.path.normpath(os.path.dirname(graphify.__file__))
    if expected != loaded:
        print(
            f"error: loaded graphify from {loaded!r}, "
            f"but graphifyy {dist.version} is at {expected!r}\n"
            "Use 'uvx --from graphifyy graphify' instead of "
            "'uv run --with graphifyy python -m graphify'.",
            file=sys.stderr,
        )
        sys.exit(1)
except Exception:
    pass
```

- Compares where the active `distribution("graphifyy")` expects its files against where the loaded module actually came from — catches both stale pip installs and unrelated same-named packages
- Uses `importlib.metadata` (stdlib since 3.9), no magic sentinels
- The error message shows both paths, making the issue immediately diagnosable

**Side note:** while looking at `BACKENDS` in `llm.py`, `OPENAI_BASE_URL` is currently read once at module import time (`os.environ.get(...)` in the dict literal). For CLI use this is harmless — env vars are always set before process start. For library/programmatic usage where env vars might be set after import, reading lazily at call time would be safer.

---

HerenderKumar:
Fixes #1540.

`uv run --with graphifyy python -m graphify` runs the **system** interpreter, so if an older `graphifyy` is also installed in system site-packages, Python imports *that* copy first and `--with graphifyy` doesn't override it. It runs with no error, but you get the old version's behavior — e.g. `OPENAI_BASE_URL` is silently ignored, requests hit the default endpoint, and you get a 401 that looks like a bad key.

The tool already prints the real fingerprint — `warning: skill is from graphify <newer>, package is <older>` — but it's easy to read that as a stale skill rather than as evidence a different install was loaded.

This adds a Troubleshooting entry (next to the existing uvx-resolution one) that:
- names the symptom,
- shows how to confirm which module actually loaded (`python -c "import graphify; print(graphify.__file__)"`),
- and points at the fix (`uvx --from graphifyy graphify …`, or `pip uninstall graphifyy` to drop the stale copy).

Docs-only, purely additive (+11 lines).

### DIFF
--- a/README.md
+++ b/README.md
@@ -550,6 +550,17 @@ The CLI is installed but its bin directory isn't on your shell's `PATH`. Pick th
 **`uvx graphify …` or `uv tool run graphify …` fails to resolve `graphify`**
 The PyPI package is `graphifyy`; `graphify` is only the command it provides. `uv tool run` treats the first word as a *package name*, so it looks for a package called `graphify` and reports `No solution found … no versions of graphify`. Name the package explicitly: `uvx --from graphifyy graphify install` (same as `uv tool run --from graphifyy graphify install`). Or `uv tool install graphifyy` once and then call `graphify` directly.
 
+**`uv run --with graphifyy python -m graphify` silently runs an older install**
+`uv run` uses your *system* Python, so if an older `graphifyy` also lives there (e.g. a past `pip install graphifyy`), Python can find that copy first on `sys.path` and `--with graphifyy` won't override it. It runs with no error, but you get the *old* version's behavior — e.g. env overrides like `OPENAI_BASE_URL` are silently ignored, so requests hit the default endpoint and fail with a 401 that looks like a bad key. The fingerprint is a `warning: skill is from graphify <newer>, package is <older>` line — that means a different install was loaded, not just a stale skill. Check which copy actually loaded:
+```bash
+python -c "import graphify; print(graphify.__file__)"
+```
+Then run the installed command directly (it uses the uv-managed copy), or drop the stale system copy:
+```bash
+uvx --from graphifyy graphify extract . --backend openai   # names the package explicitly
+pip uninstall graphifyy                                    # or remove the old system install
+```
+
 **`python -m graphify` works but `graphify` command doesn't**
 Your shell's `PATH` doesn't include the bin directory the command was installed to. Prefer `uv tool install` / `pipx install` over plain `pip`, then run `uv tool update-shell` / `pipx ensurepath` and open a new terminal (see the install notes above).
 


safishamsi:
Thanks @HerenderKumar. This shipped in v0.9.24. Your commit was cherry-picked onto v8 (abb85cc) to preserve your authorship. Closed-unmerged here, but the note is in the release: https://github.com/Graphify-Labs/graphify/releases/tag/v0.9.24

Issue #1829, opened 2026-07-12
Node IDs collide for files sharing a basename (e.g. layout.py vs layout.ts)


shuff57:
### Summary

`graphify.extract` derives every node ID from `path.stem` (the extension-less basename). Two source files that share a basename therefore collapse into a single graph node — regardless of language or directory. This silently cross-wires unrelated code: edges from one file attach to another file's symbols.

### Environment

- graphify (PyPI `graphifyy`), current release as of 2026-07
- Python 3.14, Windows; the bug is language-agnostic, not OS-specific

### Reproduce

Two files sharing the basename `layout`:

- `a/layout.py` — a Python class `Layout`
- `b/layout.ts` — an unrelated TypeScript module

```python
from graphify.extract import extract
from pathlib import Path
r = extract([Path("a/layout.py"), Path("b/layout.ts")])
ids = [n["id"] for n in r["nodes"]]
print("collisions:", len(ids) - len(set(ids)))   # > 0
```

**Observed:** both files map to the file node ID `layout`. The TS file's `contains` edges point at the Python `Layout` class/methods, and `imports_from` edges make the TS file appear to import Python modules.

**Expected:** each file gets a distinct node ID; edges stay within their own file.

### Root cause

`extract.py`:

```python
def _make_id(*parts: str) -> str:
    combined = "_".join(p.strip("_.") for p in parts if p)
    cleaned = re.sub(r"[^a-zA-Z0-9]+", "_", combined)
    return cleaned.strip("_").lower()

# ...in each per-file extractor:
stem = path.stem            # no extension, no path -> collides
file_nid = _make_id(stem)
```

`_make_id` has no path/extension awareness, so `stem` is the only disambiguator and it isn't unique across files.

### Suggested fix

Add a path-aware stem and use it wherever a *file-scoped* ID is built:

```python
def _stem_id(path: Path) -> str:
    import hashlib
    rel = str(path).replace("\\", "/")
    h = hashlib.sha1(rel.encode("utf-8")).hexdigest()[:6]
    suffix = path.suffix.lstrip(".") or "noext"
    return f"{path.stem}_{suffix}_{h}"   # layout.py -> layout_py_acdcd1
```

Replace `stem = path.stem` -> `stem = _stem_id(path)` in every per-file extractor (and the Go extractor), and update the Python import-resolver's file-node exclusion check to `_make_id(_stem_id(path))`.

[...]

PR #2087, opened 2026-07-21
fix(cli): make graphify path deterministic across runs (#2074)

### LINKED ISSUE #2074: graphify path: non-deterministic shortest-path route + calls label not backed by any stored edge
**Summary**

`graphify path "<A>" "<B>"` returns a different shortest path on repeated, identical invocations against the same unmodified `graph.json` — and one of the observed paths displayed a `--calls-->` edge label that does not correspond to any stored edge with `relation: "calls"` in the graph data (there were zero `calls`-relation edges in the entire graph).

**Steps to reproduce**

1. Build a graph with `/graphify` (code + docs + papers corpus, ~490 nodes / ~1300 edges after build).
2. Run: `graphify path "<node_id_A>" "<node_id_B>"` where the shortest path is 4 hops with more than one route of equal length available.
3. Run the exact same command again (no changes to `graph.json` in between).

**Observed**

Run 1:
```
Shortest path (4 hops):
  <A> --rationale_for [EXTRACTED]--> ProductArgument <--references [EXTRACTED]-- ShuffleArgument --calls [EXTRACTED]--> ShuffleConsistencyArgument <--rationale_for [EXTRACTED]-- <B>
```

Run 2 (same command, same graph.json, no edits in between):
```
Shortest path (4 hops):
  <A> --rationale_for [EXTRACTED]--> ProductArgument --shares_data_with [EXTRACTED]--> VectorCommitment <--references [EXTRACTED]-- MultiExpArgumentSqrt <--rationale_for [EXTRACTED]-- <B>
```

Two different routes, both length 4, returned on identical repeated calls.

Additionally, when I went looking in `graph.json` for the edge that produced the `--calls-->` label from Run 1 (to fix what I initially assumed was a bad EXTRACTED edge), I found:
- Zero edges anywhere in the graph with `"relation": "calls"`.
- Zero direct edges of *any* relation between the two class-level nodes (`ShuffleArgument`, `ShuffleConsistencyArgument`) shown adjacent in that path.

So the `--calls-->` shown in the CLI path output does not appear to come verbatim from a stored edge's `relation` field — it looks like either a display-layer synthesized/fallback label, or the path traversal is silently passing through an intermediate node (e.g. a shared/ambiguous generic node like `.prove()` or `.verify()` that appears at multiple source locations) whose real relation got mislabeled or collapsed in the printed output.

**Why this matters**

Users (and the agent driving `/graphify`) reasonably treat `EXTRACTED`-confidence output as ground truth per the skill's own Honesty Rules ("never invent an edge"). A shortest-path command that (a) isn't deterministic between identical calls, and (b) can print a relation label that doesn't trace back to any real stored edge, undermines that guarantee specifically for the `path` command's output — someone could act on `--calls-->` as fact when no such edge exists.

**Environment**
- graphifyy 0.9.21
- Windows 11, Git Bash, Python via `uv tool`
- Graph: ~490 nodes, ~1300 edges, 21 communities, undirected

**Suggested fix directions**
- Make `path`'s tie-breaking among equal-length shortest paths deterministic (e.g. stable sort by node id).
- When printing an edge in the path, read the relation directly from the stored edge object rather than any synthesized/fallback label, and if multiple parallel edges exist between the same two nodes, either print all relations or pick deterministically (not silently pick one that doesn't match what's stored).

HerenderKumar:
Fixes #2074.

`graphify path A B` returned a **different equal-length route on identical, repeated calls** against an unchanged `graph.json`.

## Root cause
`path` loads the graph via `node_link_graph`, and `graph.json` carries no `multigraph` key, so it comes back as a **`MultiDiGraph`**. The route is then computed with `nx.shortest_path`, which returns whichever one of the equal-length shortest paths the adjacency-iteration order reaches first — and Python's per-process hash randomization shuffles that order. Same graph, a different route each run.

Reproduced on a diamond graph (`a→b→d`, `a→c→d`): `graphify path a d` flips between routes purely by `PYTHONHASHSEED`, no edits in between.

## Fix
Select the **canonical** shortest path instead: the lexicographically smallest by node-id sequence among `all_shortest_paths`. Deterministic regardless of hash seed (matches the reporter's stable-sort-by-node-id suggestion). With the route stable, the relation rendered for each hop is stable too.

## Test
`test_path_is_deterministic_across_hash_seeds` spawns `graphify path` across 12 hash seeds and asserts one identical route. Red-green verified: it fails on the old code (routes vary) and passes after the fix.

## Out of scope
The reported phantom `--calls-->` (a relation that doesn't match the two labeled nodes) is a label-collision artifact — `path` prints node **labels** but traverses node **IDs**, and distinct IDs can share a label — i.e. issue #1829. This PR fixes the non-determinism; the label disambiguation belongs with #1829.

### DIFF
--- a/graphify/cli.py
+++ b/graphify/cli.py
@@ -1130,7 +1130,17 @@ def dispatch_command(cmd: str) -> None:
                         file=sys.stderr,
                     )
         try:
-            path_nodes = _nx.shortest_path(G.to_undirected(as_view=True), src_nid, tgt_nid)
+            # Deterministic route: among equal-length shortest paths, pick the
+            # canonical one (lexicographically smallest by node-id sequence) so
+            # identical calls on an unchanged graph always return the same path.
+            # nx.shortest_path alone tie-breaks by adjacency-iteration order, which
+            # Python's per-process hash randomization shuffles — same graph.json,
+            # a different equal-length route each run (#2074).
+            path_nodes = min(
+                _nx.all_shortest_paths(
+                    G.to_undirected(as_view=True), src_nid, tgt_nid),
+                key=lambda p: [str(n) for n in p],
+            )
         except (_nx.NetworkXNoPath, _nx.NodeNotFound):
             print(f"No path found between '{source_label}' and '{target_label}'.")
             sys.exit(0)
--- a/tests/test_path_cli.py
+++ b/tests/test_path_cli.py
@@ -1,6 +1,9 @@
 """Regression tests for `graphify path` arrow direction (#849)."""
 from __future__ import annotations
 import json
+import os
+import subprocess
+import sys
 import networkx as nx
 import pytest
 from networkx.readwrite import json_graph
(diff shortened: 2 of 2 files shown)


safishamsi:
Thanks @HerenderKumar — #2074 was fixed in v0.9.23: `graphify path` (and the MCP shortest_path tool) now traverse a sorted, materialized undirected graph so the route is deterministic across runs regardless of hash seed, plus the hop labels now show the edge's real relation. Landed in parallel while triaging the batch, so closing this as resolved. https://github.com/Graphify-Labs/graphify/releases/tag/v0.9.23

Issue #2082, opened 2026-07-21
Python: `from pkg import mod as alias` drops every downstream `alias.func()` call edge — unaliased form resolves fine


Kyo14363:
## Summary

For Python, `from pkg import mod as alias` records the file-level `imports_from` edge, but the alias binding is not tracked — so every downstream `alias.func()` call fails to resolve and produces no `calls` edge. The unaliased form `from pkg import mod` followed by `mod.func()` resolves correctly.

Net effect: the callee keeps only its `contains` edge from its own file and looks like dead code to `affected`, while the file-level import edge is present and makes the graph look complete.

This is independent of #2072 (scan root) — it reproduces with the scan root at the package root — and independent of `try:` nesting.

## Reproduction

```
pkg/
  __init__.py      (empty)
  gate.py
  c_alias.py
  c_tryalias.py
  c_tryplain.py
```

`pkg/gate.py`:

```python
def validate(rows):
    return bool(rows)
```

`pkg/c_alias.py` — module-level, aliased:

```python
from pkg import gate as m_gate


def run_plain_alias():
    return m_gate.validate([1])
```

`pkg/c_tryalias.py` — try-block, aliased:

```python
try:
    from pkg import gate as t_gate
except Exception:
    t_gate = None


def run_try_alias():
    return t_gate.validate([1])
```

`pkg/c_tryplain.py` — try-block, **not** aliased (control):

```python
try:
    from pkg import gate
except Exception:
    gate = None


def run_try_plain():
    return gate.validate([1])
```

Run from the package parent:

```bash
graphify extract . --code-only --out ./out
```

## Result

| file | import form | `imports_from` (file level) | `calls` → `gate.validate` |
|---|---|---|---|
| `c_alias.py` | `from pkg import gate as m_gate` | present | **missing** |
| `c_tryalias.py` | `try: … as t_gate` | present | **missing** |
| `c_tryplain.py` | `try: from pkg import gate` | present | present |

Full edge list (9 nodes, 8 edges):

```
pkg_c_alias           --contains     --> pkg_c_alias_run_plain_alias      [EXTRACTED]
pkg_c_alias           --imports_from --> pkg_gate                         [EXTRACTED]
pkg_c_tryalias        --contains     --> pkg_c_tryalias_run_try_alias     [EXTRACTED]
pkg_c_tryalias        --imports_from --> pkg_gate                         [EXTRACTED]
[...]

Issue #1883, opened 2026-07-14
Python: qualified calls (module.function()) produce no calls edge; bare-name calls are captured


Victyco-Rom:
## Summary

In Python, calls made through a module alias (`import mathlib; mathlib.compute(x)`) produce **no `calls` edge**, while bare-name calls (`from mathlib import helper; helper(x)`) are captured correctly. Since `import module` + `module.function()` is the style recommended by most Python style guides, this leaves the function-level call graph almost empty on codebases that follow it, and `query`/`affected`/`explain` confidently report that a function has no callers when it has many.

Verified on **v0.9.12 and v0.9.15** (macOS 13, Python 3.10+). AST mode, `extract --backend claude`, and `extract --mode deep` all produce the same result for code files (code extraction appears to be AST-only by design, so no backend/mode works around this).

## Minimal reproduction

Three files in an empty directory:

```python
# mathlib.py
def compute(x):
    return x * 2

def helper(x):
    return x + 1
```

```python
# caller_qualified.py
import mathlib

def use_qualified(x):
    return mathlib.compute(x)
```

```python
# caller_bare.py
from mathlib import helper

def use_bare(x):
    return helper(x)
```

```bash
graphify update . --no-cluster
```

## Actual edges (graph.json, 7 nodes / 8 edges)

```
caller_bare.py      --imports_from[EXTRACTED]--> mathlib.py
caller_bare.py      --imports[EXTRACTED]-->      helper()
use_bare()          --calls[EXTRACTED]-->        helper()      ✅
caller_qualified.py --imports[EXTRACTED]-->      mathlib.py
(no calls edge from use_qualified() to compute())               ❌
```

## Expected

`use_qualified() --calls--> compute()` (EXTRACTED, or at least INFERRED). The receiver `mathlib` is an unambiguous module alias imported in the same file, so this subset is statically resolvable from the per-file symbol table without whole-program analysis.

## Impact

On a real project (~70 Python files consistently using `import module` + `module.function()`):

- `graphify explain "compute_targets()"` → degree 3, zero incoming `calls` edges, while grep finds 4 real call sites in other modules.
- `graphify affected "compute_targets()"` → "No affected nodes found."

Module-level `imports_from` edges are correct and `affected` at file granularity works well — the gap is only function-level call edges for qualified calls.

Happy to provide more details or test a fix.

Dron09Shrivastava:
[...]

Issue #1146, opened 2026-06-06
Package-form imports (`from package import module`) are not resolved to import edges — creates disconnected test-file islands


janreview:
**Version:** graphifyy 0.8.32 (pip, Windows, Python 3.12)

**What happened**
The AST extractor resolves `from src.services.ai_blog_enricher import X` to an import edge, but NOT the package-form `from src.services import ai_blog_enricher`. Files using the package form get zero edges to the module they import.

**Example (measured)**
```python
# tests/test_ai_blog_enricher.py line 7
from src.services import ai_blog_enricher
```
Result: the test file's 141 nodes form a fully disconnected component with **0 edges** to `src/services/ai_blog_enricher.py` (60 nodes) — despite the test calling `ai_blog_enricher.analyze_article(...)` etc. throughout.

**Impact**
On a ~6,000-node graph of a Python repo: 76 connected components; 799 nodes outside the giant component, of which 524 (66%) were test files using this import style (`from src.services import X` is a common pattern for tests that monkeypatch module attributes). GRAPH_REPORT.md then flags them as "possible documentation gaps or missing edges", which sends users hunting for gaps that are actually extractor blind spots.

**Suggested fix**
When resolving `from <pkg> import <name>`, check whether `<pkg>/<name>.py` (or `<pkg>/<name>/__init__.py`) exists in the corpus and emit the same import edge as for the dotted form.

safishamsi:
Fixed in v0.8.33 (a380b34).

`from pkg import submod` now resolves to a file-level `imports_from` edge when `pkg/submod.py` or `pkg/submod/__init__.py` exists on disk. The fix lives in `_collect_python_symbol_resolution_facts` — when the resolved target is a package (`__init__.py`), each imported name is checked as a potential submodule file. Matched submodules are stored in a new `module_imports` field and emitted as `imports_from` edges in `_apply_symbol_resolution_facts` using canonical stem-based IDs.

Upgrade: `uv tool upgrade graphifyy`

janreview:
Confirmed fixed in v0.8.33 on the same repo that produced the report: components 76->61, nodes outside the giant component 799->339 (the remainder are genuinely standalone archive/scripts), and the 141-node test island now has its imports_from edge to the service. The builtin `uses`@0.5 edges from #1147 are gone too (0 remaining; `str` dropped out of the god-node top-10). Thanks for the fast turnaround!

safishamsi:
Haha always fun working on weekends

[...]

PR #2093, opened 2026-07-21
fix: resolve calls edges through an aliased Python import (#2082)

Yyunozor:
## Summary

`from pkg import mod as alias` records the file-level `imports_from` edge correctly, but every downstream `alias.func()` call is dropped — no `calls` edge is produced, so the callee looks like dead code even though the import graph looks complete. `import pkg.mod as alias` regresses the same way. The unaliased form resolves fine.

## Reproduction

```
pkg/gate.py:
    def validate(rows):
        return bool(rows)

pkg/c_alias.py:
    from pkg import gate as m_gate
    def run_plain_alias():
        return m_gate.validate([1])
```

```
```

`c_alias.py --imports_from--> pkg/gate.py` is present; `run_plain_alias --calls--> validate` is missing. Swapping out the alias (`from pkg import gate`, `gate.validate(...)`) makes the edge appear.

## Root cause

The cross-file member-call resolver's module arm (`_resolve_python_member_calls` in `extract.py`, #1883) matches a call receiver against the imported module's own file stem. `mod.func()` resolves because the receiver equals the stem; `alias.func()` never does, so the arm bails. The local alias binding is parsed correctly in two places already (`_python_imported_names` in `extractors/resolution.py`, and the `aliased_import` branch of `_import_python`) but discarded before it reaches the edges the resolver reads.

## Fix

Thread the alias through as a `local_alias` field on the `imports`/`imports_from` edge (same transient-hint pattern as `target_file`, #1814, popped once consumed so it never reaches graph.json):

- `_import_python` stamps the alias from `import pkg.mod as alias`.
- `_SymbolResolutionFacts.module_imports` gains a `local_name` slot so the `from pkg import submod [as alias]` submodule path (#1146) carries the binding through.
- The module arm's receiver match now accepts the tracked alias in addition to the module's real stem, keyed per (importing file, target module).

## Known limitation

Two different aliases for the same module in one file only resolve the last one registered (one alias slot per importing-file/target-module pair) — not a regression, since the parent resolved neither, and left out of scope here.

## Validation

Five regression tests in `tests/test_extract.py`: the issue's shape (`from pkg import gate as m_gate`), its try/except-guarded variant, the adjacent `import mathlib as m` and dotted `import pkg.gate as g_alias` forms, and the relative `from . import gate as r_gate` form. All fail on the parent commit with the missing-edge symptom, pass after the fix; one also asserts `local_alias` never reaches the returned edges.

```
uv run pytest tests/ -q
3554 passed, 3 skipped, 0 failed
```

Parent commit, same environment: 3549 passed — delta is exactly these 5 tests. `ruff check` clean. #2080's calls-edge-direction regression tests pass unchanged.

### DIFF
--- a/graphify/build.py
+++ b/graphify/build.py
@@ -802,6 +802,11 @@ def build_from_json(extraction: dict, *, directed: bool = False, root: str | Pat
         # dropping it here as well keeps a pre-fix graph's stale absolute hint
         # from surviving an incremental build_merge, which re-serializes base
         # edges through here without re-running disambiguation.
+        # `local_alias` is the same shape of transient hint (#2082): it exists only
+        # for the module arm of _resolve_python_member_calls to match an aliased
+        # import receiver, and extract() already drops it once that pass has run.
+        # Dropping it here too covers a stale pre-fix graph re-serialized through
+        # an incremental build_merge, same rationale as target_file above.
         # Sanitize numeric edge fields (#1960): an explicit ``"weight": null`` in
         # the extraction JSON survives ``.get("weight", 1.0)`` (the key is present,
         # so the default never applies) and reaches Louvain/Leiden as None,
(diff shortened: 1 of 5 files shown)


Yyunozor:
Landed on v8 as cc70085, with the follow-up coverage in 8c5f585 — closing as superseded. Thanks!

Issue #2076, opened 2026-07-21
claude-cli backend returns an agentic summary instead of JSON on Claude Code 2.1.216 (the llm.py:1376 workaround has regressed)


sortakool:
The `claude-cli` backend produces an empty graph on Claude Code **2.1.216**. The model *does* the extraction and then reports it conversationally, so `_response_is_hollow` reads truncation, bisects, and converges on nothing.

## Repro

```bash
graphify extract ./corpus --backend claude-cli --out ./out \
  --max-concurrency 1 --token-budget 12000 --api-timeout 900
```

Two markdown documents. Output:

```
[graphify] LLM returned invalid JSON, skipping chunk (first 200 chars:
  Knowledge graph extracted and delivered — 21 nodes, 20 edges, 3 hyperedges
  from the Codex↔NIM probe document.)
[graphify] claude-cli returned a hollow response; treating as truncation so
  adaptive retry can bisect the chunk.
[graphify] single-file chunk ... truncated at max_completion_tokens
[graphify extract] graph is empty — extraction produced no nodes.
rc=1
```

The two semantic cache entries written are `{"partial": true, "nodes": [], "edges": []}`.

Note the response text: **the extraction succeeded** — 21 nodes / 20 edges / 3 hyperedges — and was then summarised rather than returned. The work is done and discarded.

## This is the failure `llm.py:1376-1391` already describes

That comment documents exactly this class, and states the fix:

> Putting the full extraction schema plus an explicit imperative in the user turn — and dropping `--system-prompt` — makes the CLI emit the JSON object directly. […] **verified against Claude Code 2.1.197**

We run **2.1.216**. The workaround appears to have regressed somewhere in between.

## Isolation — four arms

| # | hypothesis | probe | result |
|---|---|---|---|
| 1 | the caller repo's `CLAUDE.md`/`AGENTS.md` context (~107 KB) makes it behave agentically | re-ran in a clean directory outside any repo | **refuted** — fails identically (`Graph fragment extracted and delivered.`) |
| 2 | a user-level `~/.claude/CLAUDE.md` leaks in everywhere | checked | **refuted** — file does not exist, so the clean room was genuinely clean |
| 3 | `claude -p --output-format json` cannot return raw JSON | direct minimal prompt | **refuted** — `result` is exactly `{"nodes":[{"id":"a","label":"A"}],"edges":[]}` |
| 4 | graphify's prompt construction | by elimination | **confirmed** |

[...]

PR #2095, opened 2026-07-22
fix(llm): pin claude-cli output to a JSON schema when supported (#2076)

Yyunozor:
## Summary

The `claude-cli` backend puts the extraction schema in the user turn and trusts the model to emit raw JSON (the `llm.py:1376` workaround, "verified against Claude Code 2.1.197"). On newer Claude Code the model treats that prompt as an agentic task and *reports* the result in prose instead — `Knowledge graph extracted — 21 nodes, 20 edges…`. That parses to zero nodes, `_response_is_hollow` reads truncation, and adaptive-retry bisects the chunk without ever converging, so the graph comes out empty (#2076).

## Why this is version-robust, and how it was checked

We run Claude Code as this backend's harness, so the failing component is directly testable here. On **2.1.207** the current user-turn workaround still returns JSON — the empty-graph break does not reproduce there, so the regression is version-gated, consistent with the 2.1.216 report. But the failure mode is easy to force and to fix on a live CLI: a prompt that invites prose makes `claude -p --output-format json` put a conversational summary in `result`, while adding `--json-schema` makes the *same* call return the object regardless of framing (also surfaced in a `structured_output` envelope field). Structured output is a structural guarantee independent of the model's discretion, so it survives the prompt-behaviour drift that broke the prompt-shape workaround.

## Fix

Pass `--json-schema` — a schema pinning the top-level `{nodes, edges}` shape graphify consumes, item internals left loose so valid extractions are never rejected — when the CLI advertises the flag, probed once via `claude --help` and cached. Older CLIs that predate the flag keep the user-turn prompt as a fallback, unchanged. The `result` envelope still carries the JSON string, so the `_parse_llm_json` path is untouched.

The issue's second suggestion — routing a successful-sounding prose response away from the bisect path — is complementary and left out of scope; with structured output the prose case no longer arises on supported CLIs.

## Validation

`tests/test_claude_cli_backend.py`: 28 passed, 5 new (flag present when supported, absent when not, the `--help` capability probe, safe fallback on probe error). Full suite: 3317 passed on the branch vs 3312 on the parent `e32c9f4`, same environment — delta is exactly these 5 tests. (13 failures in unrelated suites — terraform/ollama/manifest/install-references — are pre-existing, identical on the parent, from optional deps absent locally.) `ruff check` clean.

### DIFF
--- a/graphify/llm.py
+++ b/graphify/llm.py
@@ -1337,6 +1337,62 @@ def _claude_cli_envelope(stdout: str) -> dict:
     return envelope
 
 
+# A JSON Schema pinning the top-level shape graphify consumes. Passed to
+# `claude -p --json-schema` (structured output) so the CLI CONSTRAINS the model
+# to emit the object directly instead of relying on it CHOOSING to honour a
+# "raw JSON only" instruction in the prompt. Item internals stay loose so a
+# valid extraction is never rejected; the `result` envelope field still carries
+# the JSON string, so the parse path is unchanged. See #2076.
+_EXTRACTION_JSON_SCHEMA = json.dumps(
+    {
+        "type": "object",
+        "properties": {
+            "nodes": {"type": "array", "items": {"type": "object"}},
+            "edges": {"type": "array", "items": {"type": "object"}},
+            "hyperedges": {"type": "array", "items": {"type": "object"}},
+        },
+        "required": ["nodes", "edges"],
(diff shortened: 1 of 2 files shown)


Yyunozor:
Landed on v8 as 7116a6f, extended by 16315f1 — closing as superseded. Thanks!

PR #2102, opened 2026-07-22
fix(dedup): preserve same-source node attributes

### LINKED ISSUE #2091: Default `dedup=True` discards a same-ID node's attributes instead of merging them — LLM semantic summaries are silently lost
## Summary

`build()`'s docstring promises a key-wise union for same-ID nodes:

> Extractions are merged in order. For nodes with the same ID, the last
> extraction's attributes win (NetworkX `add_node` overwrites). Pass AST
> results before semantic […]

But on the default path `deduplicate_entities()` runs *first* and collapses
same-ID nodes by keeping one whole dict and discarding the other. Any attribute
the loser had and the winner did not — `summary`, `confidence_score`,
`references` — is gone before NetworkX ever sees it.

## Why it matters

This is exactly the AST↔semantic reconciliation the pipeline is built to
perform. `_semantic_id_remap` exists specifically to force the AST node and the
LLM node onto the same ID so they merge; when they do, the LLM's contribution is
thrown away and only the AST attributes survive.

The user has already paid for those summaries in API credits. The loss is
silent — the only output is a `note:` line about the *label*, which does not
mention that `summary` and `confidence_score` were dropped.

`dedup=True` is not an opt-in: it is hard-coded on the CLI path
(`cli.py:3230`, `cli.py:3235`), so this is what every `graphify extract` run
with a semantic pass does.

## Reproduction

```python

AST = {"nodes": [{"id": "src_auth_login", "label": "login", "file_type": "code",
                  "source_file": "src/auth.py", "_origin": "ast",
                  "source_location": "L42"}], "edges": []}
SEM = {"nodes": [{"id": "src_auth_login", "label": "User login handler",
                  "file_type": "code", "source_file": "src/auth.py",
                  "summary": "Authenticates a user.",
                  "confidence_score": 0.9}], "edges": []}

for d in (False, True):
    n = dict(build([AST, SEM], directed=True, dedup=d).nodes["src_auth_login"])
    print("dedup=%-5s label=%-22r summary=%r score=%r"
          % (d, n.get("label"), n.get("summary"), n.get("confidence_score")))
```

Actual:

```
[graphify] note: node 'src_auth_login' was extracted twice from 'src/auth.py' under
different labels — keeping 'login', dropping 'User login handler'.
dedup=False label='User login handler'   summary='Authenticates a user.' score=0.9
dedup=True  label='login'                summary=None                    score=None
```

Expected: `dedup=True` yields at least the attribute set that `dedup=False`
yields, with the survivor's values winning on genuine conflicts. Instead
`summary` and `confidence_score` are dropped entirely.

## Root cause

`graphify/dedup.py:321-338` — the survivor replaces the incumbent wholesale
rather than gap-filling from it:

```python
incumbent = seen_ids.get(nid)
if incumbent is None:
    seen_ids[nid] = node
elif _collision_rank(node) < _collision_rank(incumbent):
    seen_ids[nid] = node          # <-- whole-dict replace, incumbent's keys lost
    dropped[nid].append(incumbent)
else:
    dropped[nid].append(node)     # <-- node's keys lost
```

The comment immediately above (`dedup.py:318-320`) states the premise that makes
this look safe:

> […] a same-entity merge costs nothing and stays quiet.

That premise is false whenever the two records carry disjoint attribute sets,
which is the normal AST-plus-semantic case.

## Suggested fix

Gap-fill instead of replacing, leaving the existing `dropped` /
`_report_id_collision` bookkeeping untouched:

```python
if incumbent is None:
    seen_ids[nid] = dict(node)
elif _collision_rank(node) < _collision_rank(incumbent):
    merged = dict(node)
    for k, v in incumbent.items():
        merged.setdefault(k, v)
    seen_ids[nid] = merged
    dropped[nid].append(incumbent)
else:
    for k, v in node.items():
        incumbent.setdefault(k, v)
    dropped[nid].append(node)
```

Conflicting keys still resolve to the survivor's value, so the collision
reporting and the `_collision_rank` ordering semantics are unchanged.

## Test coverage

`tests/test_dedup.py` asserts on `len(result_nodes)`, `label`, `source_file`,
and stderr text. No test asserts the *absence* of extra keys, and none covers a
same-ID pair with disjoint attributes, which is why this survived.
`test_defining_file_wins_over_referencing_file` and
`test_same_file_relabel_is_noted` are unaffected by the change above.

## Environment

- graphify `v8` @ `6bbc9d1` (pyproject `version = "0.9.22"`)
- Python 3.12, Linux

Triaged by : @SinghAman21 and @claude

Synvoya:
## Summary

- retain non-conflicting attributes when exact-ID nodes from the same source file collapse
- preserve the deterministic collision survivor for conflicting values
- keep cross-file references and genuine slugged-path ID collisions isolated
- document the dedup-enabled build contract

Fixes #2091.

## Root cause

The exact-ID pre-dedup pass selected one complete node dictionary and discarded the other. That made deterministic collision ranking work, but it also removed complementary AST/semantic data such as `summary` and `confidence_score` before NetworkX could merge attributes.

The fix gap-fills only missing keys from same-source duplicates. It does not import attributes from another source file, so the existing #1504 collision boundary remains intact.

## Impact

Default `dedup=True` builds now retain AST precision (`source_location`, `_origin`) and semantic enrichment (`summary`, `confidence_score`) on the surviving node. Existing survivor values still win genuine conflicts.

## Validation

- reproduced #2091 on current `v8` before the fix: `summary=None`, `confidence_score=None`
- `pytest tests/test_dedup.py -q --tb=short` on Python 3.10: 48 passed
- `pytest tests/ -q --tb=short` on Python 3.12: 3553 passed, 3 skipped
- `ruff check graphify/dedup.py graphify/build.py tests/test_dedup.py`
- all five `tools.skillgen` CI checks
- `graphify update .`

### DIFF
--- a/graphify/build.py
+++ b/graphify/build.py
@@ -952,10 +952,11 @@ def build(
         ambiguous pairs in the 75–92 Jaro-Winkler score zone.
     root: if given, absolute source_file paths are made relative to root (#932).
 
-    Extractions are merged in order. For nodes with the same ID, the last
-    extraction's attributes win (NetworkX add_node overwrites). Pass AST
-    results before semantic results so semantic labels take precedence, or
-    reverse the order if you prefer AST source_location precision to win.
+    With dedup disabled, extractions are merged in order and the last node's
+    attributes win (NetworkX add_node overwrites). With dedup enabled, nodes
+    sharing an ID use a deterministic survivor and retain missing attributes
+    from duplicate records of the same source entity. Genuine cross-file ID
+    collisions remain isolated and are reported.
     """
     from graphify.dedup import deduplicate_entities
     combined: dict = {"nodes": [], "edges": [], "hyperedges": [], "input_tokens": 0, "output_tokens": 0}
--- a/graphify/dedup.py
+++ b/graphify/dedup.py
@@ -244,14 +244,35 @@ def _collision_rank(node: dict) -> tuple:
     )
 
 
+def _same_source_entity(survivor: dict, duplicate: dict) -> bool:
+    """True when exact-ID records came from the same source file.
+
+    Exact IDs can also collide across files through references or slugged-path
+    ambiguity (#1504).  Keep those records isolated rather than importing
+    attributes whose provenance belongs to another file.
+    """
+    keep_file = survivor.get("source_file") or ""
+    lose_file = duplicate.get("source_file") or ""
+    return keep_file == lose_file
+
+
+def _merge_missing_attributes(survivor: dict, duplicate: dict) -> dict:
+    """Keep survivor values while retaining non-conflicting duplicate data."""
+    merged = dict(survivor)
+    for key, value in duplicate.items():
+        merged.setdefault(key, value)
+    return merged
+
+
 def _report_id_collision(nid: str, survivor: dict, losers: list[dict]) -> None:
     """Report an ID collision in proportion to what dropping the loser actually costs.
 
-    Cross-reference to a defining node: same entity, edges are keyed by ID and rewire
-    to the survivor — nothing is lost, so say nothing. Same file, different labels: the
-    extractor emitted two labels for one entity and one is discarded — note it. Two
-    files that both encode this ID: they are distinct entities and one is genuinely
-    lost — warn, and point at the extraction split that keeps them apart (#1504).
+    Cross-reference to a defining node: the structural entity and its edges survive;
+    foreign-file attributes stay isolated, so no collision warning is needed. Same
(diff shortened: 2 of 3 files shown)


safishamsi:
Thanks @Synvoya. This shipped in v0.9.24. Your commit was cherry-picked onto v8 (5a4b207) to preserve your authorship, with a follow-up making the gap-fill order-independent and gating it on source_file. Closed-unmerged here, but it is in the release: https://github.com/Graphify-Labs/graphify/releases/tag/v0.9.24

PR #2104, opened 2026-07-22
# fix(dedup): merge same-ID node attributes instead of discarding them (#2091)

SinghAman21:
## What

On the default `dedup=True` path, when two nodes share an ID the pre-dedup step
kept one whole node dict and threw the other away. In the AST↔semantic case
(where `_semantic_id_remap` forces the AST node and the LLM node onto one ID so
they merge), the LLM's `summary` / `confidence_score` / `references` were
silently dropped — even though `build()`'s docstring promises a key-wise union.

The user already paid API credits for those summaries, and the only output was a
`note:` about the *label*, never mentioning the lost attributes.

## How

The survivor now **gap-fills** from the loser instead of replacing it wholesale:

- Only keys the survivor lacks are copied over (`setdefault`).
- Conflicting keys (`label`, `source_file`) still resolve to the survivor, so
  `_collision_rank` ordering and the collision reporting are unchanged.
- The first insert now copies (`dict(node)`) so caller-owned node dicts are never
  mutated.

Before: `dedup=True` → `summary=None`. After: `dedup=True` → `summary` preserved,
survivor's label still wins the conflict.

## Files

- `graphify/dedup.py` — gap-fill merge in the same-ID pre-dedup loop.
- `tests/test_dedup.py` — 3 tests: gap-fill preserves disjoint attrs,
  survivor wins on conflict, input nodes are not mutated.
- `CHANGELOG.md` — entry under `0.9.24 (unreleased)`.

## Reference

- Issue #2091
- Root cause: `graphify/dedup.py` same-ID collapse loop.

## Conclusion

`dedup=True` now yields at least the attribute set `dedup=False` does, with the
survivor winning genuine conflicts. No public API change; full suite green
(pre-existing unrelated `test_ollama_retry_cap.py` failures aside).


SinghAman21:
issue is already addressed and fixed the bug. so i'm closing this PR.
Thanks

PR #406, opened 2026-04-16
feat: full Dart language + Flutter ecosystem support

dacianf:
## Summary

Replaces the regex-based Dart extractor with a full tree-sitter AST implementation that brings Dart to parity with the mature Go/Rust/Elixir extractors and adds Flutter-specific structural analysis.

- **Pure Dart core** — classes (sealed/final/base/abstract), mixins, extensions, enums, typedefs, functions, imports, inheritance graph, call-graph. Uses `tree-sitter-dart-orchard` for deterministic AST parsing. Works for server-side/CLI Dart projects with zero Flutter overhead.
- **Flutter layer** (conditional on `package:flutter` imports) — widget classification (stateless/stateful/state/inherited), StatefulWidget↔State pairing, nested widget composition from `build()` methods, InheritedWidget dependency detection, Navigator API navigation edges.
- **Framework layer** (conditional on framework imports) — Riverpod `ref.watch`/`ref.read` → `watches_provider`/`reads_provider` edges, BLoC `on<Event>` → `handles_event` edges, GoRouter route definitions + `context.go`/`context.push` → `navigates_to` edges. Follows the same pattern as the existing PHP/Laravel `bound_to`/`listened_by` edges.
- **Ecosystem file parsers** — `pubspec.yaml` (package dependency graph), `melos.yaml` (monorepo workspaces), `.arb` (localization message keys).
- **Generated file exclusion** — `.freezed.dart`, `.g.dart`, `.gr.dart`, `.gen.dart`, `.mocks.dart` skipped at detect phase. `.dart_tool/` added to skip dirs.

### Architecture

Three-layer design where each layer is conditional and additive:

```
.dart file → extract.py orchestrator
                │
    ┌───────────┼───────────┐
    ▼           ▼           ▼
 Pure Dart   Flutter    Frameworks
 (always)    (if Flutter  (if framework
              imports)     imports)
```

Pure Dart server projects get full structural extraction (21 nodes, 40 edges on the test fixture) with zero Flutter overhead. The existing regex extractor produced 15 nodes and 14 edges with only 2 edge types (`defines`, `imports`) and no line numbers.

### Changes

| File | What |
|------|------|
| `graphify/extract_dart.py` (new) | Core Dart AST extractor |
| `graphify/extract_flutter.py` (new) | Flutter widget/navigation layer |
| `graphify/extract_dart_frameworks.py` (new) | Framework detection (Riverpod/BLoC/GoRouter) |
[...]

PR #2114, opened 2026-07-22
feat(extract): tree-sitter-based Dart extraction with source_location line anchors

cornwe19:
## What

Dart is currently the one mainstream language whose extractor emits `source_location: None` on every node and edge — `extract_dart` is a ~520-line regex extractor predating the generic tree-sitter framework, so Dart symbols can't be jumped to or read by line range downstream, unlike Java/Kotlin/Swift/etc.

This PR makes tree-sitter the primary Dart path, using the community [`tree-sitter-dart`](https://pypi.org/project/tree-sitter-dart/) grammar as a new optional `[dart]` extra (prebuilt wheels for win/macOS/Linux, x86+arm):

- **AST-driven with line anchors**: declarations (classes, mixins, enums, extension types, extensions, typedefs), imports/exports, `part of` redirection, annotations, and functions/factory constructors all come from the AST with `L{n}` anchors on nodes and edges.
- **Proven heuristics retained**: the Flutter-framework sweeps (Bloc events/emissions, Riverpod providers, navigation routes) and Dart 3 pattern-destructured variables — which the 0.1.x grammar still parses with ERROR nodes — stay on the existing regexes, now line-anchored via match offsets and scoped by the AST.
- **Regex fallback preserved verbatim**: without the extra installed, extraction falls back to the previous implementation unchanged (`_extract_dart_regex`) — the same optional-grammar pattern as `extract_pascal` (#781). No new required dependency.

## Fixes along the way

The AST also closes two regex false-positive classes:
- Declarations inside string literals (e.g. lint-test fixtures with code in `'''...'''`) no longer produce phantom class/function nodes.
- Multi-line named/optional parameter lists (`required String id,` sits at the 0–2-space indent the variable regex scans) no longer leak parameters as variable nodes, including a case that produced labels containing newlines.

## Validation

- All existing `tests/test_dart.py` expectations pass **on both paths** (grammar installed and absent); new tests cover anchor stamping, `part of` redirection with anchors, the extension family (named/anonymous/generic), named/optional parameters, factory vs named constructors, and generators.
- Ran against a large production Flutter codebase (2,780 `.dart` files): **100% of locally-defined nodes and all edges anchored**, 100% anchor spot-check (2,547 sampled symbols found on their anchored line), node counts within 0.5% of the regex baseline (the delta is the string-literal phantoms the AST correctly drops), and the tree-sitter path ran slightly faster than the regex path (~5s vs ~5s for the whole repo, same order).
- Error recovery verified: files the grammar can't fully parse still extract (localized ERROR nodes), and any walker exception falls back to the regex extractor per file, so a grammar gap can never cost a file entirely.

`uv.lock` regenerated with `uv lock` (adds `tree-sitter-dart` to the `all`/`dart` extras; the surrounding marker churn is current uv re-normalizing environment markers).


cornwe19:
@safishamsi FYI - this is a more recent attempt at what it looks like #406 was attempting back in April. Would love to see either one of them merged when possible.

Issue #2107, opened 2026-07-22
Zero-node warning (#1666) fires on data .json that #1224 skips by design, so it always asks users to report intended behaviour


dmitryvostryakov:
## Summary

The zero-node warning added for #1666 fires on data-shaped `.json` files that `extract_json` skips **on purpose** per #1224. Because the skip is deliberate and deterministic, the warning's own escalation advice — *"A re-run will retry them (empties are no longer cached); if it persists, please report the file(s)"* — is guaranteed to trigger: it never stops persisting, so it reliably instructs users to report a settled design decision.

I hit this on a real repo and got as far as drafting the report before reading `extractors/json_config.py` and finding the behaviour is intended. That is the failure mode: the warning converts an intentional skip into an apparently-actionable bug.

## Reproduction

Any repo containing data JSON with no manifest-recognition key. On graphify 0.9.20:

```
warning: 8 source file(s) produced zero nodes and are absent from the graph:
stateful_corpus.json, stateless_corpus.json, underwrite_payload.json,
alfa-overview.json, alfa-underwriting.postman_collection.json (+3 more).
A re-run will retry them (empties are no longer cached); if it persists,
please report the file(s) (#1666).
```

Re-running `graphify update .` reproduces the identical list — as it must, since nothing about these files will ever change the outcome.

## Why these files are skipped (working as designed)

`extract_json` recognises a config/manifest by filename or a top-level key probe (`dependencies` / `extends` / `$ref` / `$schema` / `compilerOptions`), and deliberately skips everything else — per its docstring, AST-walking data JSON "produced hundreds of orphan key-nodes and duplicate communities that swamped real structure (#1224)".

Confirmed none of the 8 carries a probe key:

| file | top-level keys | outcome |
|---|---|---|
| `benchmarks/stateful_corpus.json` | `_comment`, `now`, `scenarios`, `seed` | skip |
| `benchmarks/stateless_corpus.json` | `_comment`, `scenarios` | skip |
| `benchmarks/underwrite_payload.json` | `coverage`, `flight`, `live_context`, `sale` | skip |
| `deploy/grafana/dashboards/alfa-overview.json` | `annotations`, `panels`, `refresh`, `schemaVersion`, `tags` | skip |
| `portal/src/locales/{en,ru}.json` | `common`, `controls`, `coverage`, `cta`, `dashboard` | skip |
[...]

PR #2118, opened 2026-07-23
fix: honor target filesystem filename limits

### LINKED ISSUE #2109: _cap_filename's hardcoded limit=200 assumes NAME_MAX=255; breaks on eCryptfs (real limit can be ~143 bytes)
## Summary

`graphify export obsidian` still crashes with `OSError: [Errno 36] File name too long` on Linux systems where the export target lives on an eCryptfs-encrypted mount, even after the `_cap_filename` fix from #1094. That fix's `limit=200` assumes the standard 255-byte `NAME_MAX`, but eCryptfs's ciphertext filename encoding reduces the *effective* limit well below 255 — 143 bytes on my system.

## Root cause

`_cap_filename(s, limit=200)` in `graphify/export.py` hardcodes `limit=200` as "safely under 255". That's correct for standard filesystems (ext4, APFS, NTFS), but eCryptfs (offered by Ubuntu's installer as "Encrypt my home folder", still a mainstream option) reports a much lower `NAME_MAX` because of its filename-encryption encoding overhead:

```
$ stat -f .
  ID: ...  Longnombre: 143     Tipo: ecryptfs
$ python3 -c "import os; print(os.pathconf('.', 'PC_NAME_MAX'))"
143
```

A node label of ~155-170 bytes (well under the 200-byte cap `_cap_filename` allows) still produces a filename that exceeds this filesystem's real 143-byte limit. `_owned_write`'s `target.exists()` check raises `OSError: [Errno 36] File name too long` before any truncation logic gets a chance to help, because the cap itself was already too generous for this filesystem.

## Reproduce

1. On a Linux machine with an eCryptfs-encrypted home directory (or any filesystem with a lower-than-255 `NAME_MAX` — check with `stat -f <path>`), point the export at a source doc with a heading/label ~155-170 bytes long, e.g.:
   ```
   ### [arxiv] A Multi-Year Survey of Something With A Genuinely Long Title That Exceeds Common Encrypted Filesystem Name Limits Even Though It Fits Comfortably Under Two Hundred Bytes
   ```
2. `graphify update <dir>`
3. `graphify export obsidian <out>`
4. Crash: `OSError: [Errno 36] File name too long: '.../A Multi-Year Survey ... Bytes.md'`

## Suggested fix

`_cap_filename` (and the sibling in `wiki.py`, if it makes the same assumption) could query the target filesystem's real limit at runtime instead of hardcoding 200:

```python
import os

def _effective_name_max(out_dir: str, fallback: int = 255) -> int:
    try:
        return os.pathconf(out_dir, "PC_NAME_MAX")
    except (OSError, ValueError, AttributeError):
        return fallback  # Windows / unsupported platform
```

...and derive the cap from that (e.g. `_effective_name_max(out_dir) - 9` to leave room for the hash suffix), computed once per export run and threaded through to `_cap_filename`.

## Environment

- graphify (PyPI `graphifyy`) v0.9.11
- Ubuntu 24.04, Linux 7.0 kernel
- `$HOME` on eCryptfs (`ecryptfs_cipher=aes, ecryptfs_key_bytes=16`) — a standard Ubuntu "encrypt my home folder" install option, not an exotic setup
- `stat -f` reports `Longnombre: 143` (`PC_NAME_MAX=143`) on this mount vs the 255 `_cap_filename` assumes

Related: #1094 fixed the *unbounded* case; this is the *wrong bound* case — the cap needs to be filesystem-aware, not just present. Happy to test a fix if useful.

Synvoya:
Fixes #2109.

## Problem

Export filename capping assumes the conventional 255-byte `NAME_MAX`. On eCryptfs and other filesystems with a lower runtime limit (143 bytes in the report), valid 155–200 byte labels can still fail with `ENAMETOOLONG`. The wiki exporter also caps characters rather than UTF-8 bytes.

## Change

- Query `PC_NAME_MAX` once for each export target, with a portable 255-byte fallback
- Derive a conservative UTF-8 stem budget that reserves `.md`, generated prefixes, and collision suffixes
- Share deterministic byte-aware truncation across Obsidian, Canvas, and wiki exports
- Preserve the existing 200-byte behavior on conventional filesystems

## Regression coverage

The new tests simulate `NAME_MAX=143`. Before the fix, generated components measured 184 bytes for Obsidian and 173 bytes for Canvas/wiki. They now verify:

- every generated component is at most 143 bytes
- duplicate long labels remain case-insensitively unique
- truncated wiki links still resolve to files on disk

## Validation

- Python 3.10: `3611 passed, 3 skipped`
- Python 3.12: `3611 passed, 3 skipped`
- Focused export/wiki suite: `88 passed`
- All five `tools.skillgen` integrity checks pass
- Ruff passes on all changed files
- `graphify update .` completed: 11,968 nodes / 21,176 edges

The reporter offered to verify on real eCryptfs hardware; the regression uses a controlled `pathconf` result because the local filesystem reports the conventional limit.

### DIFF
--- a/graphify/export.py
+++ b/graphify/export.py
@@ -13,6 +13,7 @@ from datetime import date
 from pathlib import Path
 import networkx as nx
 from networkx.readwrite import json_graph
+from graphify.paths import cap_filename_stem, filename_stem_limit
 from graphify.security import sanitize_label
 from graphify.analyze import _node_community_map
 from graphify.build import edge_data
@@ -413,20 +414,14 @@ generate_html = to_html
 
 
 def _cap_filename(s: str, limit: int = 200) -> str:
-    """Cap a filename stem to ``limit`` UTF-8 bytes so it stays under the 255-byte
-    filesystem limit even after the ``.md`` extension and dedup suffix are added
-    (#1094). The cap is on BYTES, not chars, because a label of multibyte
-    characters (CJK, accented) can exceed 255 bytes well under 255 chars. When
-    truncation happens, an 8-char hash of the full label is appended so two
-    distinct labels sharing a long prefix produce distinct, deterministic
-    filenames instead of colliding."""
-    b = s.encode("utf-8")
-    if len(b) <= limit:
-        return s
-    digest = hashlib.sha1(s.encode("utf-8")).hexdigest()[:8]  # nosec - not security
-    keep = limit - 9  # "_" + 8 hex chars
-    truncated = b[:keep].decode("utf-8", "ignore")  # "ignore" drops a split trailing char
-    return f"{truncated}_{digest}"
+    """Cap a filename stem to ``limit`` UTF-8 bytes (#1094).
+
+    Callers derive ``limit`` from the target filesystem and reserve the extension
+    and dedup suffix. The cap is on bytes, not characters, so multibyte labels
+    remain within that budget. Truncated names retain a deterministic digest to
+    keep distinct long labels from colliding.
+    """
+    return cap_filename_stem(s, limit)
 
 
 def _dedup_node_filenames(G: nx.Graph, safe_name) -> dict[str, str]:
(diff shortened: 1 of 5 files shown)


Synvoya:
@safishamsi, when you have review bandwidth, could you please prioritize this small issue-backed set: #2118 fixes target-filesystem filename limits (#2109), #2117 suppresses false warnings for intentionally skipped files (#2107), and #1818 prevents Go type-set constraints from becoming false embeds. All three are ready for review, mergeable, and fully green. Happy to adjust or rebase anything you prefer.

PR #2120, opened 2026-07-23
refactor(build): share the graph-json loader and prefixed-merge across multi-repo paths

StuartMVG:
## What

Pure refactor, no behavior change: `merge-graphs` and the global graph now share one persisted-graph loader (`build.load_graph_json`) and one prefixed-merge with external-node dedup (`build.merge_prefixed_into`).

Before this, the load-a-graph.json dance (size cap, legacy `edges`→`links` normalization, type coercion) existed in three copies — the global-graph load, `global_add`'s source load, and `merge-graphs` — and the external-dedup/remap/self-loop-skip merge lived only inside `global_add`. The copies had already drifted slightly (`merge-graphs` didn't apply the size cap consistently).

## Why now

This is the first of a 3-PR stack: the cluster-graphs feature (#2121) composes member graphs through exactly these two helpers, so extracting them first keeps that PR reviewable and guarantees the three multi-repo paths (merge-graphs, global graph, clusters) can't drift again.

One observable improvement (not a behavior change to valid inputs): `merge-graphs` on a malformed JSON input now prints `error: ...` and exits 1 instead of an unhandled traceback.

## Stack

1. **→ this PR** — shared graph-json loader
2. #2121 — cluster graphs (multi-repo linked graphs)
3. #2122 — multigraph mode (parallel relations)

## Testing

- Full suite passes (3,608 passed, 3 skipped).
- Covered by the existing `test_merge_graphs_cli.py` / global-graph tests; no test changes needed.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #2121, opened 2026-07-23
feat(cluster): cluster graphs — link multiple repos into one connected graph

StuartMVG:
## What

`graphify cluster` — link several repos' graphs into one connected graph with **real cross-repo edges**, for ecosystems where the coupling lives in contracts a single-repo scan can't see: service A calls service B's HTTP API via an env-var URL, two repos share a database table, a wire-format type file is copy-mirrored between a client and a worker. (`merge-graphs` and `global` union graphs side by side; a cluster also *connects* them.)

- A committable `cluster.json` names members by **git URL** (paths resolve per machine via a gitignored local override, the spec's path hint, or origin-remote sibling auto-discovery) and declares links: `api_call`, `shared_resource`, `mirrored_file`, `depends_on`, `references` — each optionally `direction: "both"`, which materializes a real reverse edge.
- `cluster build` composes members under `tag::` namespaces into a **directed** graph and resolves links into `EXTRACTED` edges in a standard `graphify-out/graph.json`, so `query`/`path`/`explain`/`affected`/`export` work unchanged. Node selectors are `{repo, file|label|id}` — users never write raw graph ids. `cluster check` dry-runs resolution (exit 1 on errors) for CI.
- `auto_links.packages` connects a member's manifest dependencies to their unique cross-repo provider (package nodes gain normalized `package_key`/`dependency_keys`); ambiguous providers are warned and skipped, declared links take precedence.
- **Member back-references:** `cluster build` writes a portable `cluster-ref.json` into each member's `graphify-out/` (opt out with `--no-refs`; `cluster remove` cleans up its own entry). Inside a member repo, `query`/`path`/`explain`/`affected` accept `--cluster [NAME]`, no-match failures note the membership (CLI + MCP), and the search-nudge hook + installed skills surface it to assistants. All passive surfaces are fail-open; marker fields are sanitized before reaching assistant-facing output (the marker travels with clones and is untrusted).
- `affected` traverses the new `calls_api`/`mirrors` relations by default so impact analysis crosses repo boundaries (`depends_on` deliberately excluded — it predates clusters).

## Design notes reviewers will want

- **Direction is load-bearing and tested.** Members load with their stored source/target order and compose into a DiGraph — an undirected round-trip re-emits endpoints by node insertion order and silently flips caller/callee (the #760 failure mode this codebase has hit before). Regression test pins the exact flip case (callee listed before caller).
- **Failure paths are actionable and pre-write.** Corrupt/missing member graphs, empty clusters, self-composition (cluster dir as its own member), and same-name cluster collisions (URL-tracked and URL-less) all fail with remediation text *before* any output is written.
- **`graph_mode` is validated but restricted to `"simple"`** in this PR; PR 3 widens it to `"multi"`. The spec key, manifest stamp, and status line ship now so existing clusters won't force-rebuild when PR 3 lands. One transitional test asserts `"multi"` is rejected; PR 3 replaces it.
- Writing markers into member repos' worktrees is opt-out (`--no-refs`) and fail-soft; the cluster's own `graphify-out/` stays gitignored (each machine composes its own graph) while member markers are committable and travel.

## Stack

1. #2120 — shared graph-json loader (base of this branch)
2. **→ this PR** — cluster graphs
3. #2122 — multigraph mode (parallel relations)

Until PR 1 merges, this PR's diff includes PR 1's one refactor commit; I'll rebase onto `v8` as the stack lands.

## Testing

- Full suite: 3,717 passed, 3 skipped; skillgen drift validators clean.
- ~180 new tests across `test_cluster_{spec,build,cli,links,refs}.py` and `test_affected_cluster_relations.py`: spec validation, compose, selector resolution (including dotfiles, LLM-labeled graphs, and member-order-independent externals), link direction round-trips, marker lifecycle (write/backfill/conflict/removal/corruption), `--cluster` end-to-end on all four query surfaces, hook/MCP hints, and sanitization of hostile marker fields.


StuartMVG:
Closing in favor of https://github.com/Graphify-Labs/graphify/pull/2134 — same content, consolidated into a single PR at the author's preference.

PR #2122, opened 2026-07-23
feat: multigraph mode — preserve parallel relations end to end

StuartMVG:
## What

Multigraph mode: `graphify extract --multigraph` (and `graph_mode: "multi"` clusters) builds keyed `MultiDiGraph`s on stable content-derived edge keys, so **parallel relations between the same node pair survive end to end** — today a pair that both `calls` and `references` silently collapses to one edge at build time.

- **Build/merge:** `build_from_json`/`build`/`build_merge` gain a `multigraph` param; incremental merges infer the existing graph's format; the format is **sticky** across rebuilds including `--force`, and `--no-multigraph` is the explicit downgrade path (warns that parallels collapse). Conversion in either direction bypasses the no-change early exit and carries the existing graph forward — an earlier draft of conversion could replace an unchanged graph with an empty one; that's pinned by tests now.
- **Watch/global:** watch rebuilds detect a multigraph on disk and rebuild through the keyed path; `global add` promotes when either side is keyed.
- **Query surfaces:** `query`/MCP render one EDGE line per parallel edge, `get_neighbors` iterates `edge_datas` (a `relation_filter` can no longer return empty for relations that exist), `explain` lists every parallel connection, and `affected` groups hits per source reporting all matched relations — while keeping the honest def-line fallback when an edge carries no location.
- **Analyze/export:** `graph_diff` uses occurrence-level identity on multigraphs only; GraphML/Cypher/Canvas exports keep parallel edges (Cypher `MERGE`s on `graphify_key` stay idempotent; canvas edge ids are index-qualified).
- Simple graphs are untouched: the multigraph edge-sort tiebreak is gated so the simple-path build keeps upstream's exact sort and tie behavior.

The first commit is a standalone one-liner (peelable on request): watch's rebuild now passes `root=project_root` to `build_from_json`, aligning watch-produced graphs with `graphify build`'s root-relative `source_file` paths (#932).

## Stack

1. #2120 — shared graph-json loader
2. #2121 — cluster graphs (base of this branch)
3. **→ this PR** — multigraph mode

Until PRs 1–2 merge, this PR's diff includes their commits; I'll rebase onto `v8` as the stack lands. Note: this PR replaces PR 2's transitional `graph_mode` validation test (which asserted `"multi"` is rejected) with the final round-trip test — the "removed test" in the diff is that swap, not lost coverage.

## Testing

- Full suite: 3,736 passed, 3 skipped; skillgen validators clean.
- New `test_multigraph_build.py` plus multigraph tests in the extract-CLI, analyze, cluster, and affected suites: stable keys across rebuilds, incremental format inference, conversion both directions (content preserved, `--force` stickiness, downgrade warning), mixed graph-mode composition, parallel-relation surfacing on every query surface, and exporter round-trips.


StuartMVG:
Closing in favor of https://github.com/Graphify-Labs/graphify/pull/2134 — same content, consolidated into a single PR at the author's preference.

Issue #1595, opened 2026-07-01
Cross-service edges: bridge client ↔ backend across the network boundary (API-contract matching)


sheik-hiiobd:
**Proposing a direction — keen on your read before any code.** graphify is excellent at blast-radius *within* a service; this is about extending `affected` across the client/backend **network** boundary. Would this fit graphify's scope? Happy to prototype a first adapter pair if so.

# Cross-service edges — bridge client ↔ backend across the network boundary

**Type:** feature proposal / design discussion (not a bug — current behavior is correct).

## Problem

graphify maps **code** edges — calls, imports, references. In a full-stack repo (a mobile/web client + an HTTP backend), the two sides share no code; they talk **over the network**. So the graph draws them as **two disconnected islands with zero edges crossing** — even though they are one system and tightly coupled.

## Evidence

A full-stack monorepo (a mobile client + an HTTP backend), **~2,249 nodes, zero links crossing the client/backend boundary**. That's *correct* — there is no code reference between them — but the system's real coupling is **invisible** to every graph query. `graphify affected <backend endpoint>` cannot answer "which client screens break if I change this response shape," which is exactly the blast-radius question the tool is best at *within* a service.

## Root cause

The coupling is an **API contract** — an HTTP `(method, path)` — not a code reference. A static code-graph has no node for "the wire," so there is nothing to link. This is a known wall for every code-graph tool on distributed / full-stack systems; it is not a defect in resolution (the integrity self-check is clean, and rightly so).

## Proposed direction — synthetic cross-service edges via API-contract matching

Bridge the islands by matching **client call sites** to **server route handlers** on the contract, and emitting a distinct edge — never a `calls` edge (there is no code call), but a new relation, e.g. `calls_endpoint` / `serves` (INFERRED, context `"api"`):

1. **Extract client call sites** → `(method, path-template)` from the HTTP client idioms per framework (Retrofit/OkHttp annotations, `fetch`/`axios`, `URLSession`, …).
2. **Extract server routes** → `(method, path-template)` from route declarations (`@app.get("/users/{id}")`, `@GetMapping`, `router.get(...)`, …).
3. **Match** on `(method, normalized path-template)` → emit `client-call-site → server-handler`.

Framework-specific, so it fits graphify's existing shape — **per-framework adapters**, mirroring the per-language tree-sitter extractors. Start with **one client + one server framework** (whatever pair is most common) and grow the adapter set.

### Guards (the accuracy story, learned from the indirect_call work)

[...]

Issue #1177, opened 2026-06-07
how to integrate with multi repo project


alireza-salehi:
what is the best practice to integrate it with vscode  when working on a project with following structure and always opening vscode in project-root:

project-root (no git repo)
--docs             git repo
--backend       git repo
--modules
----module1   git repo
----module2  git repo
--frontend      git repo

i need to ask a question and it checks backend, frontend, modules... automatically. 

also is it possible to integrate context-mode? is ther any integration with graphify? or combine benefit?

FolatheDuckofDuckingburg:
Working with a nested sub-repository setup in VS Code (where the project-root is not a git repository, but multiple sub-directories inside it are) can confuse standard AI extensions. Because they typically look for a single .git folder at the root to map the workspace, they often miss changes inside your backend, frontend, or modules/module1 repositories.

To get cross-repo checking working seamlessly, combine context-mode, and integrate Graphify, you can follow this best-practice blueprint.

1. Setting Up the VS Code Workspace
Since your project root isn't a git repo, VS Code needs to be told explicitly that this is a Multi-Root Workspace. This allows your AI extensions (like Copilot Chat or Cursor) to treat each sub-repo as an active context boundary.

Open your project-root folder in VS Code.

Go to File > Save Workspace As... and save a file named project.code-workspace in your root folder.

Edit that file to look like this:

JSON


{
  "folders": [
    { "name": "Root", "path": "." },
    { "name": "Backend", "path": "backend" },
    { "name": "Frontend", "path": "frontend" },
    { "name": "Docs", "path": "docs" },
    { "name": "Module-1", "path": "modules/module1" },
    { "name": "Module-2", "path": "modules/module2" }
  ],
  "settings": {
    "git.autoRepositoryDetection": "subfolders"
  }
}
💡 Why this works: Setting "git.autoRepositoryDetection": "subfolders" forces VS Code to track every inner git repo simultaneously. When you ask a question, your AI assistant will naturally parse paths across folders.

2. Integrating Graphify (The Context Reducer)
Graphify turns your fragmented directories into an interconnected, queryable knowledge graph. It bridges the gap between your docs repo, your frontend, and your backend by extracting AST trees and semantic relationships.

Because Graphify reads paths recursively, you can run it directly from your non-git project-root to index the entire cluster.

Setup Steps:
Install Graphify globally using uv or pipx (use the package name graphifyy with a double 'y' on PyPI):

Bash


uv tool install graphifyy
[...]

Issue #1174, opened 2026-06-07
`affected` is direction-blind on default (undirected) graphs — `load_graph` missed the #829/#849 forced-DiGraph fix


odedd:
**graphify version**: 0.8.33 (uv tool, Linux, Python 3.12)

## Symptom

`graphify affected` is direction-blind on graphs produced by the default build path (`graphify update`, which serializes `"directed": false`). Depending on NetworkX's arbitrary edge-iteration orientation it either:

- **misses true callers** (returns `No affected nodes found.` for nodes with verified incoming `calls` edges in graph.json), or
- **reports callees as "affected"** (false positives — reverse-dependency analysis returning forward dependencies).

## Repro (any repo)

```bash
graphify update . --force        # default build → graph.json has "directed": false

# pick any calls edge from the graph; its JSON source/target keys are faithful:
jq -r '[.links[] | select(.relation=="calls")][0]' graphify-out/graph.json
# e.g. {"source": "bin_preflight" is called BY "deploy_preflight_preflight", ...}

graphify affected "<target-node-id>" --relation calls --depth 1
```

Observed on my graph (26,963 nodes): `affected bin_preflight` returned one true caller **plus** `bin/pre_deploy_tests` — which is a *callee* of `bin_preflight` (edge source=`bin_preflight`, target=`bin_pre_deploy_tests` in graph.json). On other nodes the same command returned nothing despite verified incoming edges.

**Control experiment** — flip only the flag, same data:

```bash
jq '.directed = true' graphify-out/graph.json > /tmp/graph-directed.json
graphify affected "<target-node-id>" --relation calls --depth 1 --graph /tmp/graph-directed.json
# → correct: only the true caller(s)
```

## Root cause

`affected.py::load_graph` round-trips through `json_graph.node_link_graph(raw)`, which honors `"directed": false` and returns an undirected `nx.Graph`:

1. `nx.Graph` has no `in_edges`, so `affected_nodes` falls into the fallback generator `... if target == current` over `graph.edges(data=True)`.
2. On an undirected graph, NetworkX reports each edge once with **adjacency-insertion orientation**, not the JSON's source/target — so the `target == current` test matches an arbitrary subset of edges: true callers are dropped and callees can match instead.

[...]

Issue #1061, opened 2026-05-28
v0.8.22 regression of #760: fresh /graphify build round-trips correct AST output into reversed calls edges (build_from_json -> to_json undirected round-trip)


peteratx14:
# `graphifyy` v0.8.22 — fresh `/graphify <path>` round-trips correct AST output into reversed-direction `calls` edges (regression of #760, distinct entry point from #801/#849)

**Package:** `graphifyy` v0.8.22 (uv tool install)
**Skill:** bundled skill.md as shipped
**Environment:** Windows 11, Python 3.12, fresh `/graphify packages/db` on a Next.js / Prisma monorepo subfolder
**Related (closed):** #760 (extractor fix), #801 (skill `--update` merge bypass), #849 (`graphify path` undirected render)

## Summary

Raw `graphify.extract.extract()` correctly emits `calls` edges as `(caller_id → callee_id)`. But the fresh-build pipeline that the bundled skill walks — `build_from_json(extraction)` → `to_json(G, communities, 'graphify-out/graph.json')` — round-trips through `nx.Graph` (undirected) and writes back `graph.json` with **caller and callee swapped on 7 of 8 `calls` edges** in a 77-node graph. Only edges where the callee identifier is nested inside another call expression (`outer(callee(...))`) survive with correct direction.

This is **not** the `--update` path #801 fixed and **not** the renderer-only #849 bug. The wrong direction is persisted into `graph.json` itself, on first build, with no merge step involved.

## Reproduction

```
uv tool install graphifyy==0.8.22
mkdir repro; cd repro
# Drop in any TS project with >=1 caller->callee pair, e.g. the snippet below
/graphify .
```

Minimal TS reproducer (4 functions, 8 call sites):

```ts
// seed.ts
import { seedDefaults, provisionOrganizationRoles } from "./seed-defaults";
import { seedCommsConversations } from "./seed-comms-conversations";
async function main() {
  await seedDefaults(prisma);              // expect: main --calls--> seedDefaults
  await provisionOrganizationRoles(org.id, prisma);
  await seedCommsConversations(prisma, org.id, parkId);
}
main();
```

```ts
// db-for.ts
import { orgScopeExtension } from "./org-scope-extension";
export function dbFor(session) {
  return db.$extends(orgScopeExtension(() => orgId));  // nested-as-argument
}
```

```ts
// org-scope-extension.ts
function buildOrgFilter(...) {}
function mergeWhere(...) {}
[...]

PR #2134, opened 2026-07-23
feat(cluster): cluster graphs — link multiple repos into one connected graph (+ multigraph mode)

StuartMVG:
## Summary

This PR adds two related graph capabilities:

- **Cluster graphs:** compose multiple repositories into one directed graph with explicit cross-repository edges.
- **Multigraph mode:** preserve parallel relations between the same pair of nodes throughout extraction, analysis, queries, and exports.

It also makes member repositories cluster-aware, so existing `query`, `path`, `explain`, and `affected` commands can discover and use their cluster graph.

Unlike `merge-graphs` and the global graph, which union repositories side by side, a cluster graph explicitly models connections such as API calls, shared resources, mirrored files, dependencies, and references.

## Quick start

Each member repository must have an existing graph:

```bash
cd ~/work/frontend
graphify extract .

cd ~/work/backend
graphify extract .
```

Create and build the cluster:

```bash
graphify cluster init ~/clusters/my-stack --name my-stack
cd ~/clusters/my-stack

graphify cluster add ~/work/frontend --as frontend
graphify cluster add ~/work/backend --as backend
```

Declare links in `cluster.json`:

```json
{
  "links": [
    {
      "type": "api_call",
      "name": "events-api",
      "from": {
        "repo": "frontend",
        "file": "src/lib/events-client.ts"
      },
      "to": {
        "repo": "backend",
        "file": "src/routes/events.ts"
      }
    }
  ]
}
```

Then validate and build:

```bash
graphify cluster check
graphify cluster build
```

The resulting `graphify-out/graph.json` works with existing commands:

```bash
graphify query "how are events submitted?"
graphify affected "src/routes/events.ts"
graphify path "events-client" "events-route"
```

From inside a member repository:

```bash
graphify query "how are events submitted?" --cluster
graphify query "..." --cluster my-stack
```

## Cluster graphs

A committable `cluster.json` defines members and cross-repository links.

Git URLs are used as portable member identities when available. Checkouts resolve per machine through:

1. A gitignored `cluster.local.json` override
2. A path hint in the spec
3. Configured search roots
4. Origin-matched sibling auto-discovery

URL-less repositories remain supported through path hints and local overrides.

Members are composed under `tag::` namespaces into a directed graph. Link selectors use `{repo, file|label|id}`:

- `file` suffix-matches `source_file`
- `label` matches exactly, then case-insensitively
- `id` accepts a member-local node ID

Users do not need to write the namespace-prefixed IDs found in the composed graph.

Supported links include:

- `api_call`
- `shared_resource`
- `mirrored_file`
- `depends_on`
- `references`

`direction: "both"` materializes a real reverse edge rather than storing direction as metadata only.

`auto_links.packages` connects manifest dependencies to their unique provider in another member repository. Ambiguous providers are reported and skipped, and declared links take precedence.

`cluster check` performs the same spec, member, and selector validation as a build without writing output. It exits nonzero on errors, making it suitable for CI.

## Member awareness

`cluster build` writes a portable `graphify-out/cluster-ref.json` marker into each member repository. This can be disabled with `--no-refs`, and `cluster remove` removes only its own membership entry.

The marker contains no absolute paths and can travel with a clone. It enables:

- `query`, `path`, `explain`, and `affected` with `--cluster [NAME]`
- Membership hints when a local lookup has no result
- Cluster awareness in MCP responses, installed skills, and the search-nudge hook
- Multiple cluster memberships per repository

Missing or corrupt markers never break ordinary queries or hooks.

Markers are treated as untrusted input. Automatically injected hook context uses static guidance plus a bounds-checked member count; user-invoked diagnostics sanitize and bound marker-controlled text.

Writing markers into member worktrees is intentionally opt-out. `cluster init` separately adds the cluster directory's generated `graphify-out/` to `.gitignore`.

## Multigraph mode

```bash
graphify extract . --multigraph
```

Multigraph builds use keyed `MultiDiGraph` edges so parallel relations between the same nodes survive through:

- Initial and incremental builds
- Watch rebuilds
- Cluster and global composition
- Community detection
- `affected`
- Graph diff
- CLI and MCP queries
- GraphML, Cypher, and Canvas exports

Edge keys are stable and content-derived. Cypher exports include `graphify_key` in `MERGE` operations to remain idempotent.

The graph format is sticky across rebuilds, including `--force`. Converting back requires an explicit command:

```bash
graphify extract . --no-multigraph
```

That conversion warns that parallel relations will collapse. Conversion in either direction carries the existing graph forward.

Simple graphs retain their existing edge order. A content tiebreak is applied only where multiple simple edges have the same `(source, target, relation)` identity, preventing run-to-run changes in which duplicate survives.

Community detection aggregates parallel edge weights before running algorithms that require a simple graph.

## Impact analysis

`affected` now traverses the cluster-specific `calls_api` and `mirrors` relations by default.

`depends_on` remains opt-in to avoid changing the established default traversal behavior for existing single-repository graphs.

Parallel matching relations are all surfaced when running against a multigraph.

## Failure behavior

Validation occurs before writing cluster output. Actionable errors are produced for:

- Missing or corrupt member graphs
- Empty clusters
- A cluster attempting to include itself
- Conflicting cluster names
- Invalid selectors or link directions
- Simple-mode links that would overwrite another relation on the same node pair
- Missing CLI flag values and unknown options

External-library selectors and community IDs are normalized across members so results do not depend on member order or accidentally merge unrelated communities.

Directed composition is load-bearing: converting member graphs through an undirected intermediate can reverse caller/callee endpoints based on node insertion order. Regression coverage pins this behavior.

## Review map

| Area | Primary files | Review focus |
| --- | --- | --- |
| Shared graph composition | `graphify/build.py`, `graphify/global_graph.py` | Graph loading and namespace-prefixed merging |
| Cluster specification and CLI | `graphify/cluster_graph.py`, `graphify/cluster_cli.py` | Validation, member resolution, and commands |
| Cluster composition | `graphify/cluster_graph.py` | Directed composition, selectors, links, and auto-links |
| Community detection | `graphify/cluster.py` (unrelated to cluster specs, despite the name) | Parallel-edge weight aggregation only |
| Member awareness | `graphify/cluster_ref.py`, `graphify/cli.py`, `graphify/serve.py` | Marker lifecycle, `--cluster`, hints, and sanitization |
| Multigraph semantics | `graphify/build.py`, `graphify/analyze.py`, `graphify/affected.py` | Stable keys, conversion, and parallel-edge handling |
| Export behavior | `graphify/export.py` | GraphML, Cypher, and Canvas round-trips |
| Package auto-linking | `graphify/manifest_ingest.py` | Normalized package and dependency keys |
| Assistant integration | `graphify/skill*.md`, `tools/skillgen/` | Generated skill text and drift validation |
| Regression coverage | `tests/test_cluster*.py`, `tests/test_multigraph_build.py` | End-to-end and failure-path behavior |

The first commit centralizes graph JSON loading and namespace-prefixed merging for `merge-graphs`, the global graph, and clusters. It is intended to preserve successful-path behavior while normalizing malformed-input handling into clean CLI errors.

## Additional fix

`watch` rebuilds now pass `root=project_root` to `build_from_json`, aligning watch-generated `source_file` paths with those produced by `graphify build` (extends the #932 path normalization, which previously covered only the build/extract path, to watch rebuilds).

This fix is independent of clusters and can be split into a separate PR if maintainers prefer stricter scope.

## Related issues and discussions

- **Addresses #1177:** that issue asks how to query a project composed of separate backend, frontend, documentation, and module repositories from one workspace. A cluster provides the unified graph and query surface for that layout. Automatic repository discovery and the issue's separate context-mode integration question remain out of scope, so this PR does not close it.
- **Partially addresses #1595:** declared `api_call` links connect client call sites to backend handlers, and `affected` traverses those links across repository boundaries. The issue's proposed automatic framework-specific route extraction, API-contract matching, and dangling-contract detection remain future work.
- **Implements the linked-project portion of #425:** clusters preserve per-project graphs while adding a composed graph and explicit cross-project connections. Conversation ingestion, remote source support, compact reports, and automatic workspace indexing from that broader proposal remain out of scope.
- **Implements the core linked-workspace concept from [Discussion #117](https://github.com/Graphify-Labs/graphify/discussions/117):** independently built corpora can be registered as members, queried together, and connected through declared links. The discussion's proposed automatic semantic `connect` operation is not included.
- **Builds on [Discussion #528](https://github.com/Graphify-Labs/graphify/discussions/528):** that release introduced `merge-graphs` as a namespaced cross-repository union. Cluster graphs retain namespacing while adding explicit edges between members.

Direction-preservation coverage follows the regression history in #760, #1061, and #1174. Those issues are already closed; they are cited as design and regression context rather than as issues resolved by this PR.

## Testing

At `4ade974`:

- **3,755 passed, 3 skipped**
- 135 new test functions, plus parametrized cases
- Skill-generation drift checks pass:
  - `--check`
  - `--audit-coverage`
  - `--schema-singleton`
  - `--monolith-roundtrip`
  - `--always-on-roundtrip`

Coverage includes spec validation, directed composition, selector resolution, member-order independence, bidirectional links, marker lifecycle and sanitization, all four `--cluster` query surfaces, MCP and hook hints, stable multigraph keys, format conversion in both directions, mixed-mode composition, affected traversal, and exporter round-trips.

## Related PRs

Supersedes the earlier stacked drafts #2120, #2121, and #2122, consolidating the same work into a single review.


kitsupanic:
Additional real-world validation for the multigraph mode from an Expo / React Native TypeScript corpus on **graphifyy 0.9.25**.

A fresh code-only extraction reports exactly two same-endpoint relation collisions:

1. `app_add_scan -> src_components_scanscreen`
   - `imports_from` at `app/add/scan.tsx:L1`
   - `re_exports` at `app/add/scan.tsx:L3`

2. `src_components_notificationlifecycle_notificationlifecycle -> src_components_notificationlifecycle_opennotificationdestination`
   - `calls` at `src/components/NotificationLifecycle.tsx:L45`
   - `indirect_call` at `src/components/NotificationLifecycle.tsx:L63`

Both relations are semantically useful; a simple graph retains only one edge per pair. This corpus is a compact additional regression case for `--multigraph`, query/explain output, and stable round-tripping. I can test the PR branch against the full corpus if useful.

StuartMVG:
> Additional real-world validation for the multigraph mode from an Expo / React Native TypeScript corpus on **graphifyy 0.9.25**.
> 
> A fresh code-only extraction reports exactly two same-endpoint relation collisions:
> 
> 1. `app_add_scan -> src_components_scanscreen`
>    
>    * `imports_from` at `app/add/scan.tsx:L1`
>    * `re_exports` at `app/add/scan.tsx:L3`
> 2. `src_components_notificationlifecycle_notificationlifecycle -> src_components_notificationlifecycle_opennotificationdestination`
>    
>    * `calls` at `src/components/NotificationLifecycle.tsx:L45`
>    * `indirect_call` at `src/components/NotificationLifecycle.tsx:L63`
> 
> Both relations are semantically useful; a simple graph retains only one edge per pair. This corpus is a compact additional regression case for `--multigraph`, query/explain output, and stable round-tripping. I can test the PR branch against the full corpus if useful.

Thank you. These are excellent real-world examples of the exact information-loss case --multigraph is intended to solve. Both relation pairs map to distinct keyed edges on this branch, and the query/explain paths display parallel relations independently.

I’ve added a focused explain regression test based on your calls + indirect_call example. It verifies that both edges are rendered with their respective relation sites (L45 and L63).

If you’re willing to test the PR branch against the full corpus, that would be very useful. Please test at 36ecb25 so the results are comparable. The most valuable checks would be:

- the resulting graph has "multigraph": true, and both relation pairs appear as separately keyed parallel edges;
- query and explain display both relations with their respective source sites;
- an unchanged rebuild and --force preserve the multigraph format, edge keys, and edge count;
- after a harmless edit to one affected file and re-extraction, neither relation is duplicated nor collapsed.

No need to share the corpus. The output snippets and before/after edge counts would be enough. Thanks again for the concrete case.

kitsupanic:
Tested the exact requested commit, **36ecb25**, in an isolated environment against the full Expo / React Native TypeScript corpus.

## Passes

- Output is `"multigraph": true`: **556 nodes / 1,143 keyed edges**.
- Both real relation pairs are present as separately keyed edges:
  - `imports_from` at `app/add/scan.tsx:L1`
  - `re_exports` at `app/add/scan.tsx:L3`
  - `calls` at `src/components/NotificationLifecycle.tsx:L63`
  - `indirect_call` at `src/components/NotificationLifecycle.tsx:L45`
- `query` and `explain` render both relations independently with those source sites.
- An unchanged rebuild leaves the output byte-identical.
- A full `--force` rebuild preserves the multigraph flag, all keys, the 1,143 edge count, and this sorted edge-signature SHA-256:
  `6909eab1809b69383d02fdd594bd592b6996f6b152d547215d01f6622c646a46`
- Focused upstream tests pass: **39 passed** across `test_multigraph_build.py`, `test_explain_cli.py`, and `test_extract_code_only_cli.py`.

Correction to my earlier comment: the actual corpus has `calls` at **L63** and `indirect_call` at **L45** (I previously swapped the sites).

## Harmless-edit check exposes an incremental re-extraction defect

The multigraph keying itself works, but ordinary incremental extraction drops edges owned by the changed TSX file. I reduced it to this public-safe two-file fixture:

```tsx
// app/add/scan.tsx
import { ScanScreen } from '../../src/components/ScanScreen';
export default ScanScreen;
```

```tsx
// src/components/ScanScreen.tsx
export function ScanScreen() {
  return null;
}
```

```bash
graphify extract . --code-only --multigraph
# 3 nodes, 4 edges; imports_from + re_exports are both present

printf '\n// harmless edit\n' >> app/add/scan.tsx
# 3 nodes, 1 edge; all three scan.tsx import/re-export edges are gone

graphify extract . --code-only --force
# 3 nodes, 4 edges; both parallel relations are restored
```

On the full corpus the same edit changed **1,143 → 1,140** edges and removed the `app_add_scan -> src_components_scanscreen` pair. Editing `NotificationLifecycle.tsx` retained its same-file `calls` + `indirect_call` pair, but changed **1,143 → 1,134** by dropping nine imports/calls from that file. With `--no-cluster`, the incremental run was more severe: the existing 556-node / 1,143-edge graph was replaced by a 1-node / 0-edge graph.

So the requested multigraph checks pass for initial, unchanged, query/explain, and full forced rebuilds. The final harmless-edit check is blocked by what looks like a changed-file endpoint normalization/carry-forward problem in incremental extraction rather than a multigraph key collision. Happy to file this as a separate issue if that is cleaner for the PR scope.

StuartMVG:
> Tested the exact requested commit, **[36ecb25](https://github.com/Graphify-Labs/graphify/commit/36ecb25656a9770b1bc7c3dbf20aef8aeeb12fed)**, in an isolated environment against the full Expo / React Native TypeScript corpus.
> 
> ## Passes
> * Output is `"multigraph": true`: **556 nodes / 1,143 keyed edges**.
> * Both real relation pairs are present as separately keyed edges:
>   
>   * `imports_from` at `app/add/scan.tsx:L1`
>   * `re_exports` at `app/add/scan.tsx:L3`
>   * `calls` at `src/components/NotificationLifecycle.tsx:L63`
>   * `indirect_call` at `src/components/NotificationLifecycle.tsx:L45`
> * `query` and `explain` render both relations independently with those source sites.
> * An unchanged rebuild leaves the output byte-identical.
> * A full `--force` rebuild preserves the multigraph flag, all keys, the 1,143 edge count, and this sorted edge-signature SHA-256:
>   `6909eab1809b69383d02fdd594bd592b6996f6b152d547215d01f6622c646a46`
> * Focused upstream tests pass: **39 passed** across `test_multigraph_build.py`, `test_explain_cli.py`, and `test_extract_code_only_cli.py`.
> 
> Correction to my earlier comment: the actual corpus has `calls` at **L63** and `indirect_call` at **L45** (I previously swapped the sites).
> 
> ## Harmless-edit check exposes an incremental re-extraction defect
> The multigraph keying itself works, but ordinary incremental extraction drops edges owned by the changed TSX file. I reduced it to this public-safe two-file fixture:
> 
> ```tsx
> // app/add/scan.tsx
> import { ScanScreen } from '../../src/components/ScanScreen';
> export default ScanScreen;
> ```
> 
> ```tsx
> // src/components/ScanScreen.tsx
> export function ScanScreen() {
>   return null;
> }
> ```
> 
> ```shell
> graphify extract . --code-only --multigraph
> # 3 nodes, 4 edges; imports_from + re_exports are both present
> 
> printf '\n// harmless edit\n' >> app/add/scan.tsx
> graphify extract . --code-only
> # 3 nodes, 1 edge; all three scan.tsx import/re-export edges are gone
> 
> graphify extract . --code-only --force
> # 3 nodes, 4 edges; both parallel relations are restored
> ```
> 
> On the full corpus the same edit changed **1,143 → 1,140** edges and removed the `app_add_scan -> src_components_scanscreen` pair. Editing `NotificationLifecycle.tsx` retained its same-file `calls` + `indirect_call` pair, but changed **1,143 → 1,134** by dropping nine imports/calls from that file. With `--no-cluster`, the incremental run was more severe: the existing 556-node / 1,143-edge graph was replaced by a 1-node / 0-edge graph.
> 
> So the requested multigraph checks pass for initial, unchanged, query/explain, and full forced rebuilds. The final harmless-edit check is blocked by what looks like a changed-file endpoint normalization/carry-forward problem in incremental extraction rather than a multigraph key collision. Happy to file this as a separate issue if that is cleaner for the PR scope.

Thank you! This is exceptionally thorough, and the `--force` edge-signature SHA is a nice touch.

Confirming the passes on my side: the multigraph keying, parallel-edge preservation, query/explain rendering, unchanged-rebuild stability, and forced-rebuild key/count stability are all behaving as intended. Noted on the L45/L63 correction.

## The incremental defect reproduces — and it is pre-existing on `v8`

I reproduced your harmless-edit case independently on an unrelated ~4,600-node TypeScript/Next.js corpus (789 files), and A/B'd the PR branch against `upstream/v8` (`2fa6cd3`) using two identical clones, the same single-line comment edit, and the same invocation on both sides:

```
                                initial            after edit (incr)   --force
------------------------------  -----------------  -----------------  -----------------
PR 36ecb25   simple             13,001 (20)        12,995 (14)        13,001 (20)
upstream/v8  simple             13,001 (20)        12,995 (14)        13,001 (20)

PR 36ecb25   --no-cluster       4,586n / 15,340e   9n / 22e           4,586n / 15,340e
upstream/v8  --no-cluster       4,586n / 15,340e   9n / 22e           4,586n / 15,340e

PR 36ecb25   --multigraph       13,240 (20)        13,234 (14)        13,240 (20)

edges (edges owned by the edited file);  n = nodes, e = edges
```

The numbers are identical between the PR branch and `v8` in every scenario, so the defect predates this PR and is unaffected by it. It also reproduces with no cluster and no multigraph involved, which matches your read: this is changed-file carry-forward in incremental extraction, not a multigraph keying problem. Consistent with that, `--force` fully restores in every case.

Your `--no-cluster` observation is the more serious half, and it reproduces at the same magnitude I saw (4,586 nodes → 9): the incremental run writes only the changed file's extraction over the existing graph. The clustered path degrades gracefully by comparison — it loses only the edges owned by the changed file.

So please do file it as a separate issue; that keeps this PR's scope honest, and the bug deserves its own visibility rather than being buried in a feature thread. I would suggest leading with the `--no-cluster` case given the severity, and your two-file fixture is a perfect minimal repro.

## One incidental data point in favor of `--multigraph`

On that same corpus, the multigraph build preserved **13,240 edges vs 13,001** for the simple build — 239 real parallel relations (~1.8%) that a simple graph silently collapses, at a scale beyond your two pairs. Useful corroboration that the collapse you found is not an edge case.

The `explain` regression pinning your `calls` + `indirect_call` shape (each parallel edge rendering with its own source site) is in `36ecb25`, the commit you tested — so that is already covered by the 39 passing tests you ran.

Thanks again; this is the most useful kind of PR feedback.

StuartMVG:
Following up on @kitsupanic's validation above: they filed the incremental-extraction defect as #2169, and it was fixed in v0.9.27. Thank you again. It was a great find, and I'm glad it has been addressed.

This branch is now merged with `upstream/v8` at 0.9.27 (`f5a3592`), so it carries that fix and is current with main.

## #2169's repro verified on this branch

Running the exact minimal fixture from the issue against the merged branch, in all three modes:

```
                          initial     after harmless edit   --force
------------------------  ----------  --------------------  ----------
--no-cluster              3n / 4e     3n / 4e               3n / 4e
clustered (default)       3n / 3e     3n / 3e               3n / 3e
--no-cluster --multigraph 3n / 4e     3n / 4e               3n / 4e
```

Stable across the incremental run in every mode, with no dangling edges and no absolute-path-shaped IDs — so both halves of the issue (the overwrite and the endpoint-ID normalization) hold on the multigraph path too, not just the simple one.

## Merge notes

Two conflicts, both resolved in favor of upstream:

- **`cli.py`** — #2169's `build.merge_raw_extraction` supersedes and generalizes the multigraph-conversion carry-forward this branch had (it covers every incremental run, not just format conversions), so upstream's block wins. Format conversion still bypasses the no-change early exit and now inherits the general carry-forward.
- **`build.py`** — upstream refactored `build_merge`'s existing-graph read into `_load_existing_graph()`. That helper now also returns the stored `multigraph` flag, so this branch's format inference reads it from the same load instead of parsing `graph.json` a second time.

Full suite after the merge: **3,913 passed, 3 skipped**; skillgen drift validators clean.

## One residual, unrelated to this PR

On a larger corpus (~4,600 nodes, TypeScript), the clustered path still drops a few entries after a harmless edit — 4,613 nodes / 13,029 edges to 4,611 / 13,024, restored by `--force`. I measured identical counts on plain `upstream/v8` (`f5a3592`) and on this branch — same corpus, same edit, same invocation — so it is pre-existing and unaffected by these changes; #2169's fix addressed the `--no-cluster` overwrite rather than this clustered remainder. Happy to open a separate issue with the details if that is useful.

StuartMVG:
@safishamsi - I bundled too much into this PR. At 6,054 additions across 67 files, there is a lot to evaluate because the PR contains three separable concerns at once: cluster graphs, multigraph mode, and a handful of prep refactors and standalone fixes. Splitting them would make each decision independently reviewable.

I'd like to split it. Proposed ladder, three tracks:

**Track A - prep + fixes (independent, no opinion on clusters required):**

| PR | Contents |
|---|---|
| A1 | shared graph-JSON loader (`load_graph_json`) and prefixed-merge helper (`merge_prefixed_into`) in `build.py`; `global_graph.py` and merge-graphs shrink onto them. Retires three divergent copies of the #738/#2261/#2309/#2484 handling |
| A2 | bug fix: `--graph=PATH` is silently ignored by `query`/`path`/`explain` (only `affected` parses the `=` form), and a valueless `--graph` is silently dropped; one shared option parser replaces four copy-pasted loops and rejects missing values consistently |
| A3 | small fixes: watch rebuilds drop `root` (#932); pruning leaves hyperedges whose member nodes were removed; `_partition` accepts non-finite/negative edge weights that `build_from_json` already repairs |

**Track B - multigraph mode (stacked on A1):** B1 keyed MultiDiGraph build, library-only, gated on the `multigraph_compat` probe already on v8; B2 makes the affected consumers parallel-edge-aware (`graph_diff`, partition weights, `affected`, Cypher/GraphML/Canvas exports, MCP neighbors, explain); B3 lands the `extract --multigraph` flag last, so it's only reachable once downstream output is correct.

**Track C - cluster graphs (six rungs):** C1 spec + member resolution + `init/add/remove/locate` (the "do I want to own this format?" review, isolated from the implementation); C2 `build`/`check`/`status` composition under `tag::` namespaces; C3 declared cross-repo links; C4 normalized package identities + package auto-linking; C5 member back-references + `--cluster`; C6 assistant-facing hints/hook/MCP/skill awareness (~130 real lines plus 32 regenerated artifacts, nearly half of this PR's apparent file count).

Each rung will be independently mergeable: it will compile and pass the full suite on its declared base, with reserved-but-unimplemented spec features rejected explicitly rather than half-working. If twelve PRs is more process than you want, the same cut works as three (prep+fixes / multigraph / clusters), though the cluster PR stays large in that shape.

If this works for you, I'll open A1-A3 first (they're standalone improvements regardless of what happens to the rest). I'll convert this PR to a draft that tracks the children as a checklist, and keep it as the integration reference until every rung lands or is explicitly declined.

Happy to adjust the cut - the main thing I want is for each PR to be small enough that saying yes or no to it is easy.

durmazoguzhan:
A data point from the neighbouring case, since you mention API calls and shared resources as the link types.

I hit a cross-repo join that needs no declaration at all: two .NET services on a message bus, where the producer names the message type in one repo and the consumer implements `IConsumer<T>` on the same type in the other. Both sides already name the type, so the link can be derived rather than written down. With ids being repo-prefixed it currently arrives as two unconnected nodes.

Measured on that pair before trusting it: 1440 and 262 sourced type declarations, 7 pairs agreeing on both namespace and name, and all 7 are the shared contract classes with nothing else matching. Dropping the namespace from the key pairs each service's own `Settings` immediately, so the namespace is what makes it safe.

I filed that as #3007 with a patch in #3008 against `merge-graphs`. It is a small pass, and if cluster graphs land it would belong in `cluster build` too rather than only in the union path. Nothing there conflicts with the declared links here, they cover different cases: yours handles connections with no shared symbol, this one derives the ones that do have one.

Also, splitting this PR sounds right to me as a reader. The cluster spec and multigraph mode were hard to evaluate together.

StuartMVG:
Following up on the split proposal above: **Track A is now extracted and open as #3022**, ready for review.

It carries the prep refactor and the standalone correctness fixes - the slice that needs no opinion on cluster graphs or multigraph mode:

- **A1** - shared `load_graph_json` / `merge_prefixed_into` in `build.py`; `merge-graphs` and the global graph shrink onto them, retiring three drifted copies of the #738/#2261/#2309/#2484 handling.
- **A2** - `--graph=PATH` was silently ignored by `query`/`path`/`explain` (only `affected` parsed the `=` form), and a valueless `--graph` was dropped or gave a less-useful error; one shared parser now keeps all four surfaces honest, with characterization tests pinning the old space-form behavior before the refactor was applied.
- **A3** - three small fixes: watch rebuilds now pass the project root (#932), stale-source pruning no longer lets a hyperedge outlive its members (both serialized slots), and community detection repairs invalid edge weights instead of feeding them to Leiden/Louvain.

Numbers for #3022: 11 files, 813 added lines of which 482 are regression tests; full suite green (5,088 passed / 11 skipped, one pre-existing flake unchanged from base); skillgen artifacts clean.

Next slices, in dependency order once A lands:

1. **C1** - the cluster spec + member resolution + spec-management CLI. Independent of A1; this is the "do I want to own this format?" review, isolated from all implementation.
2. **B1** - keyed multigraph build, library-only, gated on the `multigraph_compat` probe already on v8. Consumes A1's loader.
3. Then the stacks: B2/B3 (multigraph consumers, then the flag) and C2-C6 (composition, links, auto-linking, member refs + `--cluster`, assistant awareness).

I'll keep this PR as the integration reference and convert it to a draft with a child-PR checklist once the next slice opens. Happy to reorder or resize any of it - and @durmazoguzhan's #3007 derived shared-type links will get wired into cluster composition in the C-track, where they complement the declared links here.

durmazoguzhan:
Thanks for the mention, and glad the derived-links framing was useful.

One thing that might save you a step in the C-track: #3008 landed on `v8` and shipped in 0.9.49, so it is not a pending PR you need to integrate. You can build straight on top of it.

Two details that matter if you wire it into cluster composition:

- `link_shared_type_declarations(merged)` runs on the already-merged graph, right after `compose` in the merge-graphs path. The equivalent hook in `cluster build` is wherever the per-repo graphs become one, and it needs the repo-prefixed ids to already exist, since the pairing is what tells it which two nodes are the same declaration.
- The join key is namespace + name, and both ends must have a source file. Dropping either half is what makes it unsafe: without the namespace, every service's own `Settings` pairs with every other one; without the source-file check, a phantom stub pairs with a real declaration. There is a test pinning the namespace half (`test_cross_repo_shared_types.py`), and safishamsi added one pinning that the key is deliberately not structural, so two drifted copies of the same contract still link.

On the split itself: from the outside, #3022 reads much easier than #2134 did, and the A/B/C ordering makes sense to me. FWIW the four small PRs I sent yesterday landed within a few hours each, so I think the slicing is going to pay off for you.

Happy to look at the C-track slice when it opens if another pair of eyes is useful.

PR #2139, opened 2026-07-23
fix(extract): resolve bash source edges built from ${VAR} paths (#2079)

### LINKED ISSUE #2079: Bash: source "${VAR}/lib.sh" emits an unresolvable edge — sourced libraries appear orphaned
## Summary

`extract_bash` handles `source` / `.` includes, but only when the path argument is a
literal starting with `.` or `/`. The very common idiom of sourcing through a
variable-built path:

```bash
BENCH_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
source "${BENCH_DIR}/lib/gpu-discover.sh"
source "${BENCH_DIR}/lib/gpu-control.sh"
```

produces an edge whose target ID contains the literal `${BENCH_DIR}` text. That ID
matches no real file node, so the edge is reported as dangling by the Step-4.5
health check and dropped at export. Net effect: shared shell libraries look
completely disconnected from their consumers, community detection places them in
separate communities, and the report flags them as "weakly-connected nodes /
possible documentation gaps" when the repo is actually fine.

## Reproduction

Repo layout:

```
bench/ollama-bench-all.sh      # source "${BENCH_DIR}/lib/gpu-discover.sh" (L22)
bench/ollama-bench-power.sh    # source "${BENCH_DIR}/lib/gpu-control.sh"  (L33)
bench/lib/gpu-discover.sh
bench/lib/gpu-control.sh
```

Build the graph, then:

```
$ graphify path "ollama-bench-all.sh" "gpu-control.sh"
# → no path found (expected: 1-hop imports_from)
```

The Step-4.5 diagnostic reports the dropped edges as `dangling-endpoint edges`
(they were 4 of the 38 dangling edges in our build).

## Root cause

In `extractors/bash.py`, the source-command handler branches on the raw path text:

```python
raw = _read_text(args[0], source).strip().strip("'\"")
if raw.startswith((".", "/")):
    resolved = (path.parent / raw).resolve()   # literal paths: resolved + existence-checked
    ...
else:
    tgt_nid = _make_id(raw)                    # "${BENCH_DIR}/lib/gpu-discover.sh" → dead ID
    add_edge(file_nid, tgt_nid, "imports", line, context="import")
```

A `${VAR}`-prefixed path takes the `else` branch and `_make_id()` bakes the
unexpanded variable text into the node ID, which can never match the AST-generated
ID of the actual file.

The hardening in #892 / #893 deliberately treats expansions and substitutions as
noise to be filtered (token-level `literal()` returning `None`,
`is_inside_expansion()` gating) — which is correct for command detection, but as a
side effect leaves the most robust sourcing idiom invisible to extraction. As far
as I can tell this side effect has not been reported before.

## Why repos can't work around it in code

Rewriting `source "${BENCH_DIR}/lib/x.sh"` as a literal `source ./lib/x.sh` would
extract fine — but bash resolves a relative `source` against the runtime *cwd*,
not the script's location, so the rewrite breaks the script when invoked from any
other directory. The `dirname "${BASH_SOURCE[0]}"` idiom exists precisely to make
sourcing cwd-independent; repos shouldn't have to trade runtime robustness for
graph edges.

## Suggested fix

When the source argument contains a `${...}` (or `$VAR`) expansion, don't emit the
dead ID. Instead:

1. Strip the variable segment(s) and take the remaining path suffix
   (`lib/gpu-discover.sh`).
2. Match that suffix against files in the scan root.
3. Unique match → emit `imports_from` to the real file node, marked `INFERRED`
   (the expansion can't be proven statically).
   Multiple matches → `AMBIGUOUS`, or skip.
   No match → skip (current behavior, minus the dead ID).

This mirrors what #1778 did for executed scripts, and is the same problem class as
#155 (tsconfig path aliases in TS imports). The emit-an-ID-that-matches-nothing
failure mode was also fixed for json_config in #1764. If #1871 (`graphify curate`)
lands, ambiguous cases could alternatively be deferred to a curate entry instead
of guessing.

## Workaround we're using

A post-build script that re-injects the ground-truth edges into `graph.json`
(tagged `_origin: manual_source_fix`) and re-runs `graphify export html`. It works,
but `build_merge`'s replace-per-source semantics drop the manual edges whenever the
sourcing files are re-extracted (or on a full rebuild), so it has to be re-run
after updates — exactly the kind of correction #1871 aims to make durable.

## Environment

- graphifyy 0.9.22 (uv tool install), Python 3.11, Linux
- Corpus: mixed bash/python repo, code-only AST path (no LLM extraction involved)

HerenderKumar:
Fixes #2079

When a bash script sources a file through a variable path like:

```bash
source "${BENCH_DIR}/lib/gpu-discover.sh"
```

graphify was putting the literal `${BENCH_DIR}` text into the edge target, so the edge pointed at nothing, got flagged as dangling, and was dropped. The end result was that shared shell libraries looked orphaned and got split off into their own communities.

This adds a small branch for variable-built source paths:

- strip the leading `${VAR}` / `$VAR` part,
- resolve the rest against the script's own folder (which is exactly what the common `dirname "${BASH_SOURCE[0]}"` idiom makes the variable point to),
- only emit the edge (as INFERRED) when the file actually exists on disk — otherwise skip it instead of writing a dead id.

Plain `source ./x.sh` and bare-name sources behave exactly as before.

I checked it end-to-end: building the repro repo now links the sourcing script straight to the real library node, with zero dangling edges.

Tests: two new cases in `tests/test_extract.py` — the `${VAR}` idiom resolves to the real file, and an unmatched `${VAR}` path emits no edge. Full extract + symbol-resolution suites pass (182 passed, 1 skipped).

### DIFF
--- a/graphify/extractors/bash.py
+++ b/graphify/extractors/bash.py
@@ -1,10 +1,32 @@
 """Bash extractor. Moved verbatim from graphify/extract.py."""
 from __future__ import annotations
 
-
+import re
 from pathlib import Path
 from typing import Any
+
 from graphify.extractors.base import _file_stem, _make_id, _read_text
+
+
+# Leading `${VAR}` / `$VAR` expansion segment(s) of a `source` path argument. The
+# canonical `BENCH_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"` idiom makes
+# such a variable resolve to the script's own directory, so the literal suffix that
+# follows (`lib/x.sh`) can be resolved against the sourcing file's own dir (#2079).
+_BASH_LEADING_EXPANSION = re.compile(
+    r"^(?:(?:\$\{[^}]*\}|\$[A-Za-z_][A-Za-z0-9_]*)/?)+"
+)
+
+
+def _bash_source_suffix(raw: str) -> str | None:
+    """Return the literal path suffix of a variable-built `source` argument, or
+    None when the remainder is empty, still holds an expansion, or escapes upward
+    with ``..``.  ``"${DIR}/lib/x.sh"`` -> ``"lib/x.sh"``."""
+    suffix = _BASH_LEADING_EXPANSION.sub("", raw, count=1).lstrip("/")
+    if not suffix or "$" in suffix:
+        return None
+    if ".." in suffix.split("/"):
+        return None
+    return suffix
 
 
 def extract_bash(path: Path) -> dict:
@@ -180,6 +202,23 @@ def extract_bash(path: Path) -> dict:
                                 tgt_nid = _make_id(str(resolved))
                                 add_edge(file_nid, tgt_nid, "imports_from", line,
                                          context="import")
+                        elif "$" in raw:
+                            # Variable-built path, e.g. the ubiquitous
+                            # `source "${BENCH_DIR}/lib/x.sh"` idiom. The raw text
+                            # bakes the unexpanded `${VAR}` into the id, which
+                            # matches no node and is dropped as a dangling edge
+                            # (#2079). Strip the leading expansion(s) and resolve
+                            # the literal suffix against the script's own dir;
+                            # emit INFERRED (the expansion can't be proven
+                            # statically) only when it resolves to a real file,
+                            # never a dead id.
+                            suffix = _bash_source_suffix(raw)
+                            if suffix:
+                                resolved = (path.parent / suffix).resolve()
+                                if resolved.is_file():
+                                    add_edge(file_nid, _make_id(str(resolved)),
+                                             "imports_from", line,
+                                             confidence="INFERRED", context="import")
                         else:
                             tgt_nid = _make_id(raw)
                             if tgt_nid:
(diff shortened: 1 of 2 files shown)


HerenderKumar:
Quick note on scope: this PR fixes the **source-edge** half of the bash sourcing problem — `source "${VAR}/lib.sh"` now resolves to the real file instead of a dead id.

It deliberately leaves a second, related thing alone: a call to a function that's **defined in a sourced file** still doesn't get a `calls` edge — even with a plain `source ./lib.sh` where the source edge already resolves. That's a separate gap in the bash call extractor/resolver, so I've filed it on its own as #2141 to keep this PR small and focused.

safishamsi:
Thanks @HerenderKumar. Shipped in v0.9.26 (cherry-picked to v8 as 0019fc4 to preserve your authorship). During the merge with #2157 I added one integration line so calls into a ${VAR}-sourced library also resolve, not just the source edge. Closed-unmerged here, but it is in the release: https://github.com/Graphify-Labs/graphify/releases/tag/v0.9.26

PR #2149, opened 2026-07-24
fix(extract,analyze): filter Swift/Foundation/SwiftUI builtins from resolution and god-node ranking (#2147)

### LINKED ISSUE #2147: Swift/Foundation/SwiftUI builtins missing from _LANGUAGE_BUILTIN_GLOBALS and _BUILTIN_NOISE_LABELS (false god-nodes, phantom bridges)
## Summary

`_LANGUAGE_BUILTIN_GLOBALS` (`graphify/extractors/base.py:13`) and `_BUILTIN_NOISE_LABELS` (`graphify/analyze.py:11`) exist specifically to stop language built-ins (`String`, `Promise`, `str`, `dict`, ...) from becoming false "god nodes" that create phantom bridges between unrelated parts of a graph (per the comment referencing #726: *"Without this filter they become god-nodes accumulating spurious edges from every call site"*).

Both lists cover **JavaScript/TypeScript and Python only**. Swift/Foundation/SwiftUI types are entirely absent from both, and there is no `graphify/extractors/swift.py` — Swift extraction still lives in the pre-modular `extract.py` (5,372 lines) rather than the newer per-language `extractors/` layout used by Go, Rust, C#, Dart, Pascal, Objective-C, etc.

## Repro

Ran `/graphify` (full pipeline: AST + LLM-vision semantic extraction) on a ~360k-word Swift codebase (macOS + iOS audio player, SwiftUI + Foundation + AVFoundation). Results:

**God Nodes report** surfaced framework types, not architecture:
```
1. Foundation   - 105 edges
2. NSLock       - 102 edges
3. NirvanaAudioStreamerIOS - 94 edges
4. View         - 60 edges
5. NAMNode      - 51 edges
6. Data         - 49 edges
```
`Foundation`, `NSLock`, `View`, `Data` are Swift/Foundation/SwiftUI built-ins, not project abstractions — exactly the class of noise `_BUILTIN_NOISE_LABELS` is supposed to filter, but it only lists Python names (`str`, `MagicMock`, `typing`, ...).

**`graphify path` between two genuinely unrelated symbols** (`DesignSystem`, a SwiftUI style helper, and `playFLACViaAudioUnit()`, an audio-streaming method) returned a "shortest path":
```
DesignSystem <--contains-- DesignSystem.swift --contains--> Theme --implements--> Sendable <--references-- .playFLACViaAudioUnit()
```
The entire "connection" is two unrelated symbols both touching the `Sendable` protocol. If `Sendable` (and `Codable`, `Equatable`, `Identifiable`, `Hashable`, `AnyObject`, etc.) were filtered the way `Promise`/`Map`/`dict`/`str` are, this phantom bridge would not exist.

## Where it lives

- `graphify/extractors/base.py:13-32` — `_LANGUAGE_BUILTIN_GLOBALS`, used at `extract.py:2434` (member-call receiver resolution) and `extract.py:5038` (call-target resolution). Zero Swift/Obj-C entries.
- `graphify/analyze.py:11-21` — `_BUILTIN_NOISE_LABELS`, used in `god_nodes()` (`analyze.py:113`). Zero Swift entries.
- No `graphify/extractors/swift.py` exists at all (checked against `go.py`, `rust.py`, `dart.py`, `csharp.py`, `pascal.py`, `objc.py`, `elixir.py`, `apex.py`, etc., which do exist).

## Suggested fix

1. Add a Swift/Objective-C section to `_LANGUAGE_BUILTIN_GLOBALS`, at minimum: `Sendable`, `Codable`, `Decodable`, `Encodable`, `Equatable`, `Hashable`, `Identifiable`, `Comparable`, `CustomStringConvertible`, `AnyObject`, `Error`, `LocalizedError`, plus common Foundation/SwiftUI value types (`String`, `Int`, `Double`, `Float`, `Bool`, `Data`, `URL`, `Date`, `View`, `Color`, `Font`) that show up as constructor/coercion or protocol-conformance targets from every file.
2. Extend `_BUILTIN_NOISE_LABELS` with the same set so `god_nodes()` doesn't rank them.
3. Longer term: migrate Swift extraction out of `extract.py` into `graphify/extractors/swift.py` to match the modular per-language pattern already used for Go/Rust/C#/Dart/Pascal/Obj-C, so language-specific builtin/noise lists live next to the extractor that needs them instead of one shared global list.

Happy to share the reproduction corpus details or test further if useful.

MasterFede5:
Fixes #2147.

## Problem

`_LANGUAGE_BUILTIN_GLOBALS` (`graphify/extractors/base.py`) and `_BUILTIN_NOISE_LABELS` (`graphify/analyze.py`) cover only JS/TS and Python, so on Swift codebases:

- Framework symbols (`Foundation`, `NSLock`, `View`, `Data`, ...) dominate the god-nodes report instead of project abstractions (the issue reports `Foundation` at 105 edges and `NSLock` at 102 on a ~360k-word SwiftUI/AVFoundation corpus).
- `_resolve_swift_member_calls` had no builtin guard at all, so a receiver typed with a builtin (`let payload: Data`) could bind cross-file to a same-named user symbol — the same phantom-edge shape #1726 fixed for TypeScript ("the cross-file CALL resolver already skips these globals; do the same here").

## Changes

- **`graphify/extractors/base.py`** — add Swift stdlib value-type initializers (`Int`, `Double`, `Data`, `UUID`, ...), conformance protocols (`Sendable`, `Codable`, `Equatable`, `Hashable`, `Identifiable`, ...), common Foundation types (`NSLock`, `DispatchQueue`, `JSONDecoder`, ...), and SwiftUI `View`/`Color`/`Font` to `_LANGUAGE_BUILTIN_GLOBALS`. `String`/`Date`/`URL`/`Error` were already present via the ECMAScript section.
- **`graphify/analyze.py`** — add the same set plus framework module names (`Foundation`, `SwiftUI`, `UIKit`, `AppKit`, `Combine`) to `_BUILTIN_NOISE_LABELS` so `god_nodes()` stops ranking them.
- **`graphify/extract.py`** — `_resolve_swift_member_calls` now skips builtin receiver types, mirroring the #1726 guard the TS/Python member-call resolvers already have.

## Scope notes

- The shared-list tradeoff (a user class literally named `Data`/`View` in another language loses receiver-type resolution) is the same one already accepted for `Request`/`Response`/`Error`/`File` in the existing list; the guard only affects resolution *to user symbols*, never node extraction itself.
- The issue's item 3 (migrating Swift extraction out of `extract.py` into `graphify/extractors/swift.py` so per-language builtin lists can live with their extractor) is intentionally left as follow-up.
- Conformance-edge emission (`implements → Sendable` stubs from `inheritance_specifier`) is untouched — filtering those changes graph content for every Swift repo and mirrors what Python does today with `inherits → Exception`, so it felt like a maintainer call. The god-node symptom those stubs cause is addressed by the `analyze.py` filter.

## Tests

`tests/test_swift_builtin_noise.py`:
- Parametrized `god_nodes()` exclusion for `Foundation`, `SwiftUI`, `NSLock`, `Data`, `View`, `Sendable`, `Codable`, `DispatchQueue`, `Color` (mirrors the npm dep-block test shape).
- Extraction-level regression: builtin-typed receiver (`let payload: Data`) does not bind to a user `class Data` in another file.
- Guard is a no-op for genuine user types: `engine.play()` on a user-typed property still resolves cross-file (#1356 shape).

Targeted run: `tests/test_swift_builtin_noise.py tests/test_builtin_global_type_refs.py tests/test_analyze.py tests/test_god_nodes_cli.py tests/test_swift_cross_file_calls.py` → 73 passed. Full suite on Windows: 3366 passed / 51 failed — the identical 51 failures occur on a clean `v8` checkout in the same environment (missing optional deps / Windows-specific), i.e. no regressions from this change.

### DIFF
--- a/graphify/analyze.py
+++ b/graphify/analyze.py
@@ -18,6 +18,14 @@ _BUILTIN_NOISE_LABELS = frozenset({
     "Callable", "Type", "ClassVar", "Final", "Literal", "Protocol",
     "Counter", "defaultdict", "OrderedDict", "datetime", "Enum",
     "os", "sys", "re", "json", "io", "abc", "typing",
+    # Swift / Foundation / SwiftUI framework symbols and module imports that
+    # otherwise dominate god-node rankings on Swift codebases (#2147)
+    "Foundation", "SwiftUI", "UIKit", "AppKit", "Combine",
+    "String", "Int", "Double", "Float", "Bool", "Data", "URL", "Date", "UUID",
+    "Sendable", "Codable", "Decodable", "Encodable", "Equatable", "Hashable",
+    "Identifiable", "Comparable", "AnyObject", "Error", "LocalizedError",
+    "NSObject", "NSString", "NSError", "NSLock",
+    "View", "Color", "Font", "DispatchQueue",
 })
 
 # Language families — extensions sharing a runtime can legitimately call each other
--- a/graphify/extract.py
+++ b/graphify/extract.py
@@ -2191,6 +2191,12 @@ def _resolve_swift_member_calls(
             type_qualified = False
         if not type_name:
             continue
+        # A builtin receiver type (Data, NSLock, DispatchQueue, ...) must not
+        # resolve to a same-named user symbol — the cross-file CALL resolver and
+        # the TS/Python member-call resolvers already skip these globals (#1726);
+        # do the same for Swift (#2147).
+        if type_name in _LANGUAGE_BUILTIN_GLOBALS:
+            continue
         type_defs = type_def_nids.get(_key(type_name), [])
         if len(type_defs) != 1:  # ambiguous or absent -> bail (god-node guard)
             continue
--- a/graphify/extractors/base.py
+++ b/graphify/extractors/base.py
@@ -29,6 +29,25 @@ _LANGUAGE_BUILTIN_GLOBALS: frozenset[str] = frozenset({
     "print", "open", "isinstance", "type", "super", "sorted", "reversed",
     "any", "all", "abs", "round", "next", "iter", "hash", "id", "repr",
     "callable", "getattr", "setattr", "hasattr", "delattr", "vars", "dir",
+    # Swift standard library / Foundation / SwiftUI (#2147). Value-type
+    # initializers (Data(x), Int(x), UUID()) and protocol conformance targets
+    # appear from virtually every file of a Swift codebase, exactly like the
+    # ECMAScript constructors above. String/Date/URL/Error are already listed.
+    "Int", "Int8", "Int16", "Int32", "Int64",
+    "UInt", "UInt8", "UInt16", "UInt32", "UInt64",
+    "Double", "Float", "Bool", "Character",
+    "Sendable", "Codable", "Decodable", "Encodable", "Equatable", "Hashable",
+    "Identifiable", "Comparable", "CaseIterable", "RawRepresentable",
+    "CustomStringConvertible", "CustomDebugStringConvertible", "AnyObject",
+    "LocalizedError",
+    "Data", "UUID", "Decimal", "Calendar", "Locale", "TimeZone", "Bundle",
+    "IndexPath", "IndexSet", "NotificationCenter", "UserDefaults",
+    "FileManager", "URLSession", "URLRequest", "URLComponents",
+    "JSONDecoder", "JSONEncoder", "DateFormatter", "NumberFormatter",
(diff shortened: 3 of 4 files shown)


safishamsi:
Thanks @MasterFede5. Shipped in v0.9.27 (cherry-picked to v8). Closed-unmerged here, but it's in the release: https://github.com/Graphify-Labs/graphify/releases/tag/v0.9.27

PR #2157, opened 2026-07-24
fix(extract): resolve bash calls into sourced-file functions (#2141)

### LINKED ISSUE #2141: Bash: calls to functions defined in a sourced file never get call edges
## Summary

When a Bash script calls a function that's defined in a **sourced** file, graphify never records the call edge. The function definitions and the source/import edge are all indexed fine, but the caller shows zero outgoing calls to the sourced function — so shell scripts look disconnected from the libraries they use.

This is separate from #2079. #2079 is about the *source* edge (`source "${VAR}/lib.sh"` pointing at nothing). This one is about the *call* edge, and it happens even with a plain `source ./lib.sh` where the source edge resolves fine.

## Reproduction (mainline v8)

`b.sh`
```bash
#!/usr/bin/env bash
b_func() { echo ok; }
```

`a.sh`
```bash
#!/usr/bin/env bash
source ./b.sh
main() { b_func; }
```

```
graphify .
```

In `graph.json`:

- `imports_from: a.sh -> b.sh` ✅ (source edge resolves)
- `calls: []` ❌ (expected `main -> b_func`)

`b_func` has no incoming call edge, so `graphify path main b_func` and `graphify callers b_func` come up empty even though both functions are indexed.

## Root cause

- `extract_bash` (`graphify/extractors/bash.py`) only records a call edge when the callee is defined in the **same** file (`name in defined_functions`). A call to a function from a sourced file is silently dropped, and the extractor returns just `{"nodes", "edges"}` — it never emits the `raw_calls` / `bash_sources` that other languages use for cross-file resolution.
- There's already a resolver for exactly this — `resolve_bash_source_edges` (`graphify/symbol_resolution.py:404`), with unit tests — but it has no call site anywhere in `graphify/`. It's defined and tested but not wired into the extract pipeline, and it has no `raw_calls` / `bash_sources` to work from anyway.

## Suggested fix

Have `extract_bash` emit `raw_calls` for calls whose callee isn't defined in the current file (plus `bash_sources` for the source edges), then run them through `resolve_bash_source_edges` in the pipeline so cross-file calls to sourced functions become real `calls` edges.

## Acceptance

- [ ] `main -> b_func` call edge exists for the repro above.
- [ ] Works for both `source ./b.sh` and the `dirname "${BASH_SOURCE[0]}"` / `${VAR}` idiom (the source-edge half is #2079).
- [ ] `graphify callers b_func` and `graphify path main b_func` resolve.
- [ ] A call to a genuinely undefined/external command still produces no edge.

## Prior art

Aziron's atlas hit the same thing and split it into these two exact halves — source resolution and the intra-function call — in aziron-ai/atlas#13, which also references #2079.

**Environment:** graphify 0.9.25 (v8), macOS arm64.

HerenderKumar:
Closes #2141.

## Problem

`extract_bash` only records a call edge when the callee is defined in the **same** file (`name in defined_functions`). A call to a function from a `source`d file is silently dropped, so shell scripts look disconnected from the libraries they use — `graphify callers b_func` and `graphify path main b_func` both come up empty even though both functions are indexed.

The resolver for exactly this — `resolve_bash_source_edges` (`graphify/symbol_resolution.py`) — already exists with unit tests, but had **no call site** anywhere in `graphify/`, and no `raw_calls` / `bash_sources` to work from.

## Fix

- **`extractors/bash.py`** now emits the two buckets cross-file resolution needs:
  - `bash_sources` — the files this script `source`s (gated on existence, like the existing `imports_from` edge, so crafted traversal paths never enter the resolver's data).
  - `raw_calls` — calls whose callee isn't defined locally (candidate calls into a sourced library), stamped `language: "bash"`.
- **`extract.py`** runs `resolve_bash_source_edges` after the id-remap passes (so caller/function node ids are final) and de-dupes the source edge the extractor already emitted via `existing_edges`.
- Bash `raw_calls` are **excluded from the generic global-name resolver**. That resolver matches by name across the whole corpus and would emit an INFERRED phantom edge from a bash call to any same-named function in an *unsourced* file — including calls to external commands that merely share a name with some function elsewhere. Resolution is therefore scoped strictly to the `source` relationship.

## Acceptance

- [x] `main -> b_func` call edge exists for the issue's repro.
- [x] `graphify callers b_func` / `graphify path main b_func` resolve (the `calls` edge now exists).
- [x] A call to a genuinely undefined/external command still produces no edge — verified even when a same-named function exists in an *unsourced* file.
- [~] `source ./b.sh` is fully handled. The `dirname "${BASH_SOURCE[0]}"` / `${VAR}` idiom is the source-edge half tracked by #2079 (PR #2139); once that source edge resolves, the call half composes by having that branch also record a `bash_sources` entry.

## Testing

Four tests added in `tests/test_extract.py` (unit: extractor emits `raw_calls`/`bash_sources`; integration: the call edge resolves end-to-end; guards: no duplicate source edge, and an external/unsourced same-named command stays unlinked). Red-green verified (the two core tests fail without the production change). Full suite: 3383 passed, with only pre-existing missing-dependency failures unrelated to this change. Also confirmed resolution survives the incremental-cache path.

### DIFF
--- a/graphify/extract.py
+++ b/graphify/extract.py
@@ -133,6 +133,8 @@ from graphify.extractors.resolution import (  # noqa: E402,F401
     _workspace_globs,
 )
 
+from graphify.symbol_resolution import resolve_bash_source_edges  # noqa: E402
+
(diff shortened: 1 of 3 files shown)


safishamsi:
Thanks @HerenderKumar. Shipped in v0.9.26 (cherry-picked to v8 as bdcae25 to keep your authorship). Closed-unmerged here, but it is in the release: https://github.com/Graphify-Labs/graphify/releases/tag/v0.9.26

PR #2159, opened 2026-07-24
feat(wiki): preserve hand-authored custom pages across export

TomaszWu14:
## Problem

`graphify export wiki` deletes **every** `wiki/*.md` at the start of each run:

```python
for old_article in out.glob("*.md"):
    old_article.unlink()
```

This exists to clear orphaned community articles (LLM-generated labels are non-deterministic, so a renamed community would otherwise leave its old file behind). But it also deletes **any page a user added by hand** — custom analysis, notes, topology deep-dives — with no way to keep them. Re-running the export silently wipes them, and there's no opt-out.

## Fix

`to_wiki` now tracks the files it generates in a `.graphify_wiki_files.json` manifest and, on re-export, deletes **only its own prior output** (files it generated last run but not this run — e.g. renamed communities). Any file graphify never wrote is left untouched.

Preserved pages are also auto-listed in a **Custom Pages** section of `index.md`, titled by their first markdown `# H1` (falling back to a de-slugged filename), so they stay reachable from the agent entry point instead of becoming orphans.

**Migration safety:** the first run after upgrade has no manifest, so it deletes nothing — existing custom pages survive. Self-generated orphans are cleaned up from the next run onward.

## Behavior

- Community/god-node articles regenerate exactly as before (byte-identical on an unchanged graph).
- A hand-authored `my-analysis.md` now survives re-export and appears under `## Custom Pages` in the index.
- Renaming a community still removes its stale article (verified by test).

## Tests

Adds 5 tests to `tests/test_wiki.py`:
- `test_to_wiki_preserves_custom_pages`
- `test_index_lists_custom_pages_by_h1`
- `test_to_wiki_removes_only_its_own_orphans`
- `test_to_wiki_writes_manifest`
- `test_custom_page_without_h1_falls_back_to_filename`

`pytest tests/test_wiki.py` → **35 passed** (30 existing + 5 new).

## Notes

- `import json` moved to module level (was needed by the new manifest read/write).
- No public API change; `to_wiki`'s signature is unchanged. `_index_md` gains an optional `custom_pages` argument.

### DIFF
--- a/graphify/wiki.py
+++ b/graphify/wiki.py
@@ -4,6 +4,7 @@ from __future__ import annotations
 from collections import Counter
 from pathlib import Path
 from urllib.parse import quote
+import json
 import networkx as nx
 
 from graphify.build import edge_data
@@ -166,6 +167,18 @@ def _god_node_article(G: nx.Graph, nid: str, labels: dict[int, str], node_commun
     return "\n".join(lines)
 
 
+def _custom_title(path: Path) -> str:
+    """Link title for a preserved custom page: its first markdown H1, else a
+    de-slugged filename."""
+    try:
+        for line in path.read_text(encoding="utf-8").splitlines():
+            if line.startswith("# "):
+                return line[2:].strip()
+    except Exception:
+        pass
+    return path.stem.replace("-", " ").replace("_", " ").title()
+
+
 def _index_md(
     communities: dict[int, list[str]],
     labels: dict[int, str],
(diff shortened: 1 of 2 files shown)


TomaszWu14:
Friendly heads-up: this PR is a first-time contribution, so the CI workflows are sitting in `action_required` pending maintainer approval. Whenever a maintainer has a moment, could you approve the workflow run so the checks can execute? Happy to address anything they surface. Thanks for the great tool! 🙏

PR #2162, opened 2026-07-24
skill: mandatory-ID gate for re-extracted docs on --update

AirRocker:
## Problem

Semantic re-extraction of a doc that already has nodes in `graph.json` is non-deterministic. Running `--update` twice on the same large project `CLAUDE.md` — changed by one line between runs — returned **75 nodes on one run and 49 on the next**, silently dropping an entire documented section.

Nothing errored. The loss went straight into `build_merge` and would only have surfaced much later as a question the graph could no longer answer. It was caught solely because the `to_json` shrink guard (#479) refused the write.

The failure mode is quiet by construction: the extraction "succeeds", the chunk is valid JSON, and the merge is happy. Only a node-count comparison against the prior build reveals it.

## Fix

Adds a required three-step gate to the `--update` flow, applied to any changed doc/paper that already has nodes in the graph:

1. **Snapshot** that file's existing node IDs from `graph.json` before extracting.
2. **Pass them to the subagent as mandatory must-include IDs**, with the baseline node count as an explicit target ("the prior build extracted N nodes; land at ~N — under 90% means you under-extracted").
3. **Hard-gate the merge** — refuse unless node count ≥ 90% of baseline *and* zero mandatory IDs are missing. On failure, re-dispatch naming the missing IDs; two consecutive failures stop and report rather than merging a regression.

It also restates that the shrink guard is the authoritative backstop: never `force=True` past it without diffing the old and new node sets and being able to name why each removed node is legitimately gone (file deleted, section removed, ID renamed with a verified replacement). ID churn on a re-extracted file is a legitimate shrink; a missing section is not.

Reusing the prior IDs has a second benefit beyond completeness: it suppresses gratuitous ID churn on re-extraction (`claude_weather_proxy_worker` → `claude_workers_weather_proxy`), which orphans saved queries and inflates the merge diff for no semantic gain.

## Relationship to #3203 / 0.9.53

0.9.53 ships a per-source, count-level guard on the `graphify extract` path. `build_merge` now flags any re-extracted source whose semantic node count strictly drops, and the extract CLI arms the shrink guard on that flag: the write is refused and the file's manifest stamp is withheld, so the file re-dispatches on the next run. I re-ran my #3203 repro against 0.9.53 and it does exactly that ([verification comment](https://github.com/Graphify-Labs/graphify/issues/3203#issuecomment-5479876567)).

So what this gate still covers is the residual, and there are two parts to it:

**1. The default skill `--update` path.** Only `graphify extract` acts on the new flag (`_handle_unverified_semantic_shrink` in `cli.py`). The skill's `--update` flow calls `build_merge` directly (`tools/skillgen/fragments/references/shared/update.md`) and never reads the flag. That call also passes no `dedup`, so #2497's identity diff is skipped on this path too (it is gated `if had_graph and not dedup:` in `build.py`). On a skill `--update`, an under-extracted doc is at best caught afterwards, by the whole-graph `to_json` count check (#479). That check can tell you something shrank, but not what, and it can't see the loss at all when growth elsewhere in the same run offsets it.

**2. The count-neutral identity shape.** When a re-extract returns the same number of nodes for a file under new IDs, a per-source count check passes it by construction. Upstream's own `test_build_merge_equal_count_different_ids_not_flagged` marks that shape as by design. On 0.9.53 it still writes: exit 0, 22 → 22 nodes, and the same 9 edges dropped as in the original repro. The mandatory-ID gate compares identity rather than count, so on the skill path that shape gets re-dispatched instead of merged.

In short, 0.9.53 guards strict shrinks on `graphify extract`. This gate covers the skill `--update` path, where that guard doesn't arm, and the count-neutral shape, which no count check can see.

## Evidence

- The 75 → 49 collapse above, caught only by the shrink guard.
- After adopting the gate: **three consecutive regression-free doc updates** on the same project.
- First run with the gate codified: **111/111 mandatory IDs present, node count held exactly**, merge clean, health check clean.
- A second, worse collapse (−26 nodes, dropping an entire documented API-contract section) was caught by the gate before merge and fixed by re-dispatching with the ID list — the graph never regressed.

## Scope

Markdown only — no code or CLI changes.

- Hand-edited: `tools/skillgen/fragments/references/shared/update.md` (the gate) and `tools/skillgen/fragments/core/core.md` (a pointer from the `--update` section).
- Everything else in the diff is regenerated via `python -m tools.skillgen` + `--bless`.

All five skillgen guards pass locally: `--check`, `--audit-coverage`, `--schema-singleton`, `--monolith-roundtrip`, `--always-on-roundtrip`.

## Note on CI

`tests/test_labeling.py::test_label_communities_batches_when_over_batch_size` may show red. It is **pre-existing and unrelated to this PR** — re-confirmed on an unmodified `v8` checkout at 50556ba, where it reproduces with:

```
pytest tests/test_skillgen.py tests/test_labeling.py
```

Cause: `label_communities` runs batches in a thread pool (`graphify/llm.py`, `workers = max(1, min(max_concurrency, n_batches))`), and for `backend="gemini"` concurrency stays > 1. The test appends to `calls` from inside the worker, so it records *completion* order but asserts *dispatch* order (`[100, 100, 50]`); under a different thread schedule it observes `[100, 50, 100]`. It passes when its file runs alone, which is why it usually goes unnoticed.

Happy to split that into its own issue/PR if useful — I left it untouched here to keep this diff to instructions only.

### DIFF
--- a/graphify/skill-agents.md
+++ b/graphify/skill-agents.md
@@ -675,6 +675,8 @@ fi
 
 Both are non-default subcommands. `--update` re-extracts only new or changed files; `--cluster-only` reruns clustering on the existing graph. See `references/update.md` for both flows.
 
+> **Re-extracting an already-graphed doc requires the mandatory-ID gate** (`references/update.md` → "REQUIRED: mandatory-ID gate for re-extracted docs"): snapshot that file's existing node IDs, pass them to the subagent as must-include, and refuse the merge if the node count drops below 90% of baseline or any ID is missing. Semantic re-extraction is non-deterministic and drops whole sections without it.
+
 ---
 
 ## For /graphify query
--- a/graphify/skill-amp.md
+++ b/graphify/skill-amp.md
@@ -675,6 +675,8 @@ fi
 
 Both are non-default subcommands. `--update` re-extracts only new or changed files; `--cluster-only` reruns clustering on the existing graph. See `references/update.md` for both flows.
 
+> **Re-extracting an already-graphed doc requires the mandatory-ID gate** (`references/update.md` → "REQUIRED: mandatory-ID gate for re-extracted docs"): snapshot that file's existing node IDs, pass them to the subagent as must-include, and refuse the merge if the node count drops below 90% of baseline or any ID is missing. Semantic re-extraction is non-deterministic and drops whole sections without it.
+
 ---
 
 ## For /graphify query
--- a/graphify/skill-claw.md
+++ b/graphify/skill-claw.md
@@ -678,6 +678,8 @@ fi
 
 Both are non-default subcommands. `--update` re-extracts only new or changed files; `--cluster-only` reruns clustering on the existing graph. See `references/update.md` for both flows.
 
+> **Re-extracting an already-graphed doc requires the mandatory-ID gate** (`references/update.md` → "REQUIRED: mandatory-ID gate for re-extracted docs"): snapshot that file's existing node IDs, pass them to the subagent as must-include, and refuse the merge if the node count drops below 90% of baseline or any ID is missing. Semantic re-extraction is non-deterministic and drops whole sections without it.
+
 ---
 
 ## For /graphify query
--- a/graphify/skill-codex.md
+++ b/graphify/skill-codex.md
@@ -675,6 +675,8 @@ fi
 
 Both are non-default subcommands. `--update` re-extracts only new or changed files; `--cluster-only` reruns clustering on the existing graph. See `references/update.md` for both flows.
 
+> **Re-extracting an already-graphed doc requires the mandatory-ID gate** (`references/update.md` → "REQUIRED: mandatory-ID gate for re-extracted docs"): snapshot that file's existing node IDs, pass them to the subagent as must-include, and refuse the merge if the node count drops below 90% of baseline or any ID is missing. Semantic re-extraction is non-deterministic and drops whole sections without it.
+
 ---
 
 ## For /graphify query
--- a/graphify/skill-copilot.md
+++ b/graphify/skill-copilot.md
@@ -678,6 +678,8 @@ fi
 
 Both are non-default subcommands. `--update` re-extracts only new or changed files; `--cluster-only` reruns clustering on the existing graph. See `references/update.md` for both flows.
 
+> **Re-extracting an already-graphed doc requires the mandatory-ID gate** (`references/update.md` → "REQUIRED: mandatory-ID gate for re-extracted docs"): snapshot that file's existing node IDs, pass them to the subagent as must-include, and refuse the merge if the node count drops below 90% of baseline or any ID is missing. Semantic re-extraction is non-deterministic and drops whole sections without it.
+
 ---
 
 ## For /graphify query
--- a/graphify/skill-droid.md
+++ b/graphify/skill-droid.md
@@ -675,6 +675,8 @@ fi
 
 Both are non-default subcommands. `--update` re-extracts only new or changed files; `--cluster-only` reruns clustering on the existing graph. See `references/update.md` for both flows.
 
+> **Re-extracting an already-graphed doc requires the mandatory-ID gate** (`references/update.md` → "REQUIRED: mandatory-ID gate for re-extracted docs"): snapshot that file's existing node IDs, pass them to the subagent as must-include, and refuse the merge if the node count drops below 90% of baseline or any ID is missing. Semantic re-extraction is non-deterministic and drops whole sections without it.
+
 ---
 
 ## For /graphify query
--- a/graphify/skill-kilo.md
+++ b/graphify/skill-kilo.md
@@ -678,6 +678,8 @@ fi
 
 Both are non-default subcommands. `--update` re-extracts only new or changed files; `--cluster-only` reruns clustering on the existing graph. See `references/update.md` for both flows.
 
+> **Re-extracting an already-graphed doc requires the mandatory-ID gate** (`references/update.md` → "REQUIRED: mandatory-ID gate for re-extracted docs"): snapshot that file's existing node IDs, pass them to the subagent as must-include, and refuse the merge if the node count drops below 90% of baseline or any ID is missing. Semantic re-extraction is non-deterministic and drops whole sections without it.
+
 ---
 
 ## For /graphify query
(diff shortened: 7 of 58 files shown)


AirRocker:
Rebased onto current `v8` (50556ba) — clean, no conflicts, and the diff is unchanged at 58 files / +1943.

Verified locally on the rebased branch: all five skillgen guards pass (`--check`, `--audit-coverage`, `--schema-singleton`, `--monolith-roundtrip`, `--always-on-roundtrip`), and `pytest tests/` gives 4218 passed / 3 skipped. The one failure is the pre-existing `test_label_communities_batches_when_over_batch_size` flake described in the PR body — it reproduces the same way on an unmodified `v8` checkout, so it isn't coming from this branch. The Actions run currently shows `action_required`, so CI is waiting on a workflow approval rather than on anything in the branch.

I've also added a short section to the description on how this sits alongside #2497, since that landed after the PR was opened and is clearly the better version of the count check. The reason I think it's still worth having both: the revived guard is gated `if had_graph and not dedup:` (`build.py:1823`) while `dedup` defaults to `True` (`build.py:1553`), so on a default `--update` the identity-level check is skipped by design, and a silent doc-node loss is still only caught afterwards by the `to_json` count comparison — which can tell you something shrank, but not which nodes went or whether it was legitimate ID churn.

Very happy to rework the shape if a lighter touch would fit the project better — an opt-in flag, a different default, or a config knob would all be fine by me. The protection is what I care about, not the mechanism. No rush on any of this.

AirRocker:
Rebased onto current `v8` (adf814f). It applied cleanly and the diff is unchanged at 58 files / +1943. All five skillgen guards pass locally, including `--check` on the Windows skill, which upstream had touched since the last rebase. `pytest tests/` is fully green run the way CI runs it (5538 passed / 13 skipped), and the `label_communities` flake noted in the description didn't trigger this time.

I've also updated the description for the post-0.9.53 picture. The count-level guard now ships on the extract CLI path, so this gate's remaining scope is the skill `--update` path plus the count-neutral ID-churn shape.

The fresh push needs a workflow approval before CI runs, whenever that's convenient. Genuinely no rush. I'm still happy to reshape this however suits the project, whether that's an opt-in flag, a different default, or a config knob.

AirRocker:
Thanks for this — the review earned its keep. All four distinct findings are real, and three of them I could reproduce against a live graph before touching a word of the text. Fixed in `a9b253d`, force-pushed over `7992600`.

### 1. Single shared state file for every re-extracted document (`…/update.md:84`, and the same finding on the `amp` variant)

Real, and the worst of the four, because it is the one that fails **open**. `.graphify_must_ids.txt` was a fixed path, so in a delta with two changed docs the second snapshot overwrote the first, and that doc went through the merge ungated while the gate still printed `GATE: PASS` — a silent hole in the one procedure whose whole job is to fail closed. Reproduced on a 3,545-node graph: snapshot doc A (2 mandatory IDs), snapshot doc B (27), and A's IDs are simply gone from disk.

One clarification on the framing, because it changes what the fix has to be: this isn't a concurrency race. Nothing writes these files concurrently — the extraction subagents write `CHUNK_PATH`, while the snapshot and the gate both run in the host session, sequentially. It is a plain sequential overwrite, which is why keying the state file by target is sufficient and no locking is needed.

**Fix:** state files are keyed by target (`.graphify_must_<slug>.json`), with a one-time clear of stale snapshots before the first target of a run.

### 2 + 3. Suffix collisions from `endswith(TARGET)` (`…:86` and `…:87`)

Real — and the sharpest of the three to demonstrate, because it needs no operator error at all. `endswith` is not anchored to a path boundary, so a `source_file` that is a bare root-level name is a suffix of every same-named file deeper in the tree. On one of our real corpora, `TARGET = 'README.md'` — which is exactly how the root README appears in `source_file`, so there is nothing to mistype — also matched `workers/README.md`:

```
gate (endswith) baseline : 22 nodes
true (exact)   baseline :  7 nodes
FOREIGN nodes pulled in  : 15  (from workers/README.md)
```

The consequence is a hard FAIL on correct output: a *perfect* re-extraction of `README.md` lands at 31.8% of the inflated baseline with 15 mandatory IDs reported missing. And the remedy the procedure then prescribes makes it worse — "re-dispatch naming the missing IDs" tells the subagent to emit another file's nodes under this file's `source_file`, which replace-on-re-extract would then attribute to the wrong file.

**Fix:** match `source_file` exactly. Where no exact match exists but suffix candidates do, the snapshot prints the candidates and stops rather than guessing.

### 4. Gate reads only `.graphify_chunk_01.json` (`…:105`)

Real. One correction to the mechanism, because it shaped the fix: a document never splits across chunks. Step 3B chunks by *file* — 20–25 files per chunk, and each image gets its own — so one file always lands in exactly one chunk. Multiple chunks come from a delta with many semantic files, and then the target doc can sit in `_02` while the gate reads `_01`, sees none of its mandatory IDs, and fails a clean extraction. Not hypothetical: a recent `--update` on our side dispatched two doc subagents on a 14-file delta.

**Fix:** the gate now unions every `.graphify_chunk_*.json` and evaluates every snapshot in one pass. That also closed a latent over-count nobody flagged — the old `len(d['nodes'])` was the whole chunk, so for any chunk holding more than one file the 90% shrink check was measured against the wrong number. It now counts only the nodes attributed to each target, longest-suffix wins, so `workers/README.md` nodes are not counted against `README.md`.

### Verification

Two-chunk delta with the target in `_02` now passes with correct per-doc counts (7/7 and 15/15), and two negative controls confirm the gate can still fail: drop one mandatory node → `FAIL` with the ID named; drop one baseline hyperedge → `FAIL`. All five skillgen guards pass. `pytest tests/` is 5412 passed / 95 skipped — the only failures are the four `test_ollama_retry_cap.py` tests that need the uninstalled `openai` extra and reproduce identically on a pristine checkout, plus one `scales_linearly` timing test that passes in isolation.

Worth saying plainly: this gate text runs in production for us, across a six-project fleet, in a hardened superset of this fragment. So these are not edits to a proposal — the identical three fixes went into the live copy in the same pass, which is also where the reproductions above came from.

CI is showing `action_required` on the new SHA, so it is waiting on a workflow approval whenever that is convenient. No rush on any of it, and as before I am happy to reshape the whole thing if a lighter touch would suit the project better.

Issue #1050, opened 2026-05-27
AST: @property/@staticmethod methods get class-unqualified node ids but class-qualified rationale_for edge targets → docstring nodes orphaned at build


p-gueguen:
## Summary

The Python AST extractor generates **inconsistent node ids** for methods decorated with `@property` / `@staticmethod`: the method **node** gets a *class-unqualified* id, but the `rationale_for` **edge** that links the method's docstring/comment targets a *class-qualified* id. The two ids don't match, so `build_from_json` treats the edge as dangling and drops it (the `src not in node_set or tgt not in node_set: continue` branch). The result is **orphaned (degree-0) `rationale` nodes** — every docstring/comment on a decorated method becomes a disconnected node in the graph.

The same class-qualified-vs-not mismatch can also drop `calls` / `references` edges that point *into* decorated methods, so those methods are under-connected in general.

## Minimal reproduction

`repro.py`:
```python
class Bar:
    @property
    def baz(self) -> int:
        """Return the baz value."""
        return 1

    def normal(self) -> int:
        """A normal instance method."""
        return 3
```

```python
from pathlib import Path
from graphify.extract import extract
from graphify.build import build_from_json

ast = extract([Path("repro.py")])

# method NODE ids:
#   baz (@property)      -> id "..._baz"          label "baz()"      (NO class segment, NO leading dot)
#   normal (instance)    -> id "..._bar_normal"   label ".normal()"  (class segment present)

# rationale_for EDGE targets:
#   baz docstring    -> target "..._bar_baz"     -> target_in_nodes = False  (class-qualified, does not match node)
#   normal docstring -> target "..._bar_normal"  -> target_in_nodes = True

G = build_from_json(ast)
# baz's docstring rationale node  -> degree 0  (ISOLATED)
# normal's docstring rationale node -> degree 1 (connected)
```

Observed output:
```
NODE ids:
  id=..._baz           label='baz()'
  id=..._bar_normal    label='.normal()'
rationale_for edges:
  target=..._bar_baz       target_in_nodes=False   <-- dropped at build
  target=..._bar_normal    target_in_nodes=True
after build:
  rationale ..._rationale_4   degree=0   'Return the baz value.'   <-- orphaned
  rationale ..._rationale_13  degree=1   'A normal instance method.'
```

## Root cause

[...]

PR #2175, opened 2026-07-25
fix(extract): emit references edges for Python decorators (#2154)

### LINKED ISSUE #2154: Python: applying a decorator creates no edge — `affected <decorator>` returns "No affected nodes found" for every function it wraps
## Summary

Applying a decorator (`@my_decorator`) creates **no edge** between the decorated function and the decorator. The decorated function looks independent of the thing that wraps it, so reverse-impact queries miss it.

## Reproduction

```
pkg/deco.py
pkg/consumer.py
```

```python
# pkg/deco.py
def my_decorator(func):
    def wrapper(*a, **k):
        return func(*a, **k)
    return wrapper
```

```python
# pkg/consumer.py
from pkg.deco import my_decorator

@my_decorator
def business_logic(x):
    return x * 2
```

`graphify extract . --code-only --force` on **0.9.25** — full edge list:

```
consumer.py --contains-----> business_logic()
consumer.py --imports_from--> deco.py
consumer.py --imports-------> my_decorator()
deco.py     --contains-----> my_decorator()
```

There is **no** `business_logic() -> my_decorator()` edge. The only link is the file-level import.

```
$ graphify affected "my_decorator" --depth 2
- consumer.py [imports] pkg/consumer.py:L1
```

`business_logic()` — the thing whose behaviour actually changes if the decorator changes — is not reported.

## Why it matters

This is the standard "framework entrypoint" pattern. In our fleet, a shared package exposes its whole public API as two decorators:

```python
# md_dataset/process.py
def md_py(func): ...      # -> get_file_manager() -> load_data(...)
def md_upload(func): ...  # -> get_file_manager().save_tables(...)
```

Six downstream repos consume the package **only** through `@md_py` / `@md_upload`. Because decoration creates no edge, reverse traversal from the decorator finds none of the functions it wraps, and:

```
$ graphify affected md_py --depth 2
No affected nodes found.
```

That is a **false negative on a "what breaks?" question** — cheap, fast, confident, and wrong. It is more dangerous than no answer, because an empty result reads as "nothing depends on this."

## Suggested fix

When a `decorated_definition` / decorator list is present, emit an edge from the decorated symbol to the decorator (e.g. `decorated_by`, or reuse `uses` / `calls`). Python is the clearest case; the same applies to TS/JS decorators.

Possibly related: #1566 (extending `indirect_call` to remaining dispatch patterns) — decoration is arguably another dispatch pattern.

## Environment

graphifyy 0.9.25 · macOS 15 (darwin 25.5.0) · Python 3.9 · `--code-only`

Rishet11:
Closes #2154

Python now emits the same decorator edges TS/JS already emits.

## Problem

Applying a Python decorator emitted no edge to the decorator symbol, so `affected <decorator>` answered "No affected nodes found" for every function it wraps — a silent false negative on reverse-impact queries.

`decorated_definition` was handled as a transparent wrapper so a decorated method keeps its class-qualified node id (#1050). Its `decorator` children were walked, but no branch consumed them:

```python
if t == "decorated_definition":
    for child in node.children:
        walk(child, parent_class_nid=parent_class_nid)
    return
```

TS/JS already handled this via `_ts_emit_decorator_edges`. Python was never wired up.

## Fix

A `_python_decorator_name` helper placed directly above its twin `_ts_decorator_name`, differing only in grammar node names (Python `call`/`attribute` vs TS `call_expression`/`member_expression`), plus emission inside the existing `decorated_definition` branch.

- Same shape as TS: a `references` edge with `context="decorator"` from the decorated function/class to each decorator symbol. **No new edge type, so no schema, HTML exporter, MCP server, or `affected` change.**
- Owner ids reuse the definition branches' own formulas verbatim (`_make_id(stem, ns, name)` for classes, `_make_id(parent_class_nid, name)` for methods, `_make_id(stem, name)` for module-level functions), so the edge lands on the node the walk is about to create.
- Targets go through `ensure_named_node`, so an imported decorator becomes a sourceless stub the corpus rewire collapses onto its real definition — this is what makes cross-file `affected` work.
- Gated on `config.ts_module == "tree_sitter_python"`; `decorated_definition` is a Python-only grammar node.
- The `#1050` `parent_class_nid` propagation is untouched, with a regression test pinning it.

Covered forms: `@deco`, `@deco(arg)`, `@mod.deco(arg)` (targets the symbol, not the module alias, matching `_ts_decorator_name`), stacked decorators, decorated classes, and decorated methods. Names that `normalize_id()` empties are skipped, matching the `#1899` guard in the definition branches, so no edge can dangle.

## Verification

End-to-end on the issue's repro, before:

```
Affected nodes for my_decorator()
- consumer.py [imports] pkg/consumer.py:L1
```

after:

```
Affected nodes for my_decorator()
- consumer.py [imports] pkg/consumer.py:L1
- business_logic() [references] pkg/consumer.py:L4
```

| Check | Result |
|---|---|
| New tests before the fix | 7 of 9 fail (real negative controls) |
| New tests after | 9 passed |
| `tests/test_ts_decorators.py` + Python extraction subsets | 61 passed |
| Full suite, base `66d8110` | 15 failed, 3620 passed |
| Full suite, this branch | 15 failed, 3629 passed |
| Failure-set diff | identical, no new failures |
| `ruff check` | passed |
| `python -m tools.skillgen --check` | OK, 134 artifacts match |

The 15 pre-existing failures are unrelated (`test_ollama.py` backend-detection and similar) and identical before and after; measured with `-p no:randomly` in the same env.

## Known limitation

A comment placed between `@` and the decorator expression (`@ # note` newline `deco`) drops that one edge. It is a missed edge, never an incorrect one, and does not affect decorators written in normal style. Left out of scope to keep the diff surgical; happy to fold in a fix if you would prefer it here.

### DIFF
--- a/graphify/extractors/engine.py
+++ b/graphify/extractors/engine.py
@@ -3581,6 +3581,41 @@ def _extract_generic(
         # node orphaned (#1050). Treat decorated_definition as a transparent
         # wrapper so parent_class_nid propagates to the real function node.
         if t == "decorated_definition":
+            # Applying a decorator emitted no edge to the decorator symbol, so
+            # `affected <decorator>` reported nothing for the functions it wraps
+            # (#2154). Emit the same shape TS/JS already emits in
+            # `_ts_emit_decorator_edges`: a `references` edge (context=
+            # "decorator") from the decorated function/class to each decorator,
+            # resolved via ensure_named_node so an imported decorator becomes a
+            # sourceless stub the corpus rewire collapses onto its definition.
+            # The owner ids mirror the definition branches below/above verbatim,
+            # so the edge lands on the node the walk is about to create.
+            if config.ts_module == "tree_sitter_python":
+                inner = node.child_by_field_name("definition")
+                inner_name = None
+                if inner is not None:
+                    name_node = inner.child_by_field_name("name")
+                    inner_name = _read_text(name_node, source) if name_node else None
+                # A name that normalizes to nothing is skipped by the definition
+                # branches (#1899), so an edge to it would dangle.
+                if inner_name and normalize_id(inner_name):
+                    if inner.type in config.class_types:
+                        owner_nid = _make_id(stem, ".".join(namespace_stack), inner_name)
+                    elif parent_class_nid:
+                        owner_nid = _make_id(parent_class_nid, inner_name)
+                    else:
+                        owner_nid = _make_id(stem, inner_name)
+                    for child in node.children:
+                        if child.type != "decorator":
+                            continue
+                        deco_name = _python_decorator_name(child, source)
+                        if not deco_name:
+                            continue
+                        deco_line = child.start_point[0] + 1
+                        target = ensure_named_node(deco_name, deco_line)
+                        if target != owner_nid:
+                            add_edge(owner_nid, target, "references", deco_line,
+                                     context="decorator")
             for child in node.children:
                 walk(child, parent_class_nid=parent_class_nid)
             return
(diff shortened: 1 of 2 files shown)


safishamsi:
Thanks @Rishet11. Shipped in v0.9.27 (cherry-picked to v8 to preserve your authorship), with a small follow-up excluding builtin/stdlib decorators from the reference edges. Closed-unmerged here, but it's in the release: https://github.com/Graphify-Labs/graphify/releases/tag/v0.9.27

PR #2176, opened 2026-07-25
fix(resolution): honor jsconfig.json and compilerOptions.baseUrl (#2153)

### LINKED ISSUE #2153: JS: jsconfig.json baseUrl not honoured — non-relative imports (static AND dynamic) produce no edge, orphaning every module in a Rails/webpacker app
## Summary

JS imports written against `jsconfig.json`'s `compilerOptions.baseUrl` produce **no edge**. Relative specifiers in the same file resolve fine. This affects **static and dynamic** imports equally — it is not a dynamic-`import()` issue.

Graphify already *parses* `jsconfig.json` (it emits `compilerOptions` / `baseUrl` nodes), it just doesn't use it for module resolution.

Related but distinct: #147 asks for **TypeScript** `tsconfig` `paths`. This is plain **JS** + `jsconfig` `baseUrl`, which is the default layout for Rails/webpacker and Vue SPAs.

## Reproduction

```
gfjs/
  jsconfig.json                       {"compilerOptions": {"baseUrl": "app/javascript"}}
  app/javascript/mods/Local.js        export default class Local {}
  app/javascript/mods/Widget.js       export default class Widget {}
  app/javascript/mods/Registrar.js
```

```js
// Registrar.js
import Local  from './Local.js';        // relative
import Widget from 'mods/Widget.js';    // baseUrl-relative

export default class Registrar {
  static L = Local;
  static W = Widget;
  static async lazy() { return import('mods/Widget.js'); }
}
```

`graphify extract . --code-only --force` on **0.9.25**:

```
Registrar.js     --imports_from --> Local.js
Registrar.js     --imports      --> Local
Registrar.js     --contains     --> Registrar
Registrar        --method       --> .lazy()
Widget.js        --contains     --> Widget      <-- no inbound edge at all
jsconfig.json    --contains     --> compilerOptions
compilerOptions  --contains     --> baseUrl     <-- config IS parsed
```

| specifier | edge created? |
|---|---|
| `'./Local.js'` relative | ✅ |
| `'mods/Widget.js'` baseUrl static | ❌ |
| `import('mods/Widget.js')` baseUrl dynamic | ❌ |

A separate matrix (no config) confirms relative works with **and** without the extension, so the variable is *non-relative*, not the extension.

## Why it matters

In a real Rails + webpacker app (~17k nodes, `jsconfig.json` with `baseUrl: app/javascript`), **every dashboard module and instruction is an orphan node**. The registrar does:

```js
// DashboardModuleRegistrar.js:58
import('workspaces/modules/registry/modules/EvosepSystemSuitabilityModule'),
```

and `EvosepSystemSuitabilityModule.js` has **0 inbound edges**. So:

```
$ graphify affected "EvosepSystemSuitabilityModule.js" --depth 2
No affected nodes found.
```

…even though the registrar is a real dependency. On a "what will break if I change this?" question, a confident empty result is the worst possible failure mode — plain `grep` finds it in one call. This silently affects every module/instruction in the app.

## Suggested fix

Honour `jsconfig.json` / `tsconfig.json` `compilerOptions.baseUrl` (and `paths`) when resolving JS specifiers, same as #147 proposes for TS. Falling back to "try `<baseUrl>/<specifier>`" would cover the common case.

## Environment

graphifyy 0.9.25 · macOS 15 (darwin 25.5.0) · Python 3.9 · `--code-only`

Rishet11:
Closes #2153

## Problem

Non-relative imports produced no edge in a Rails/webpacker project, so every module under `baseUrl` was orphaned and `affected` answered nothing.

There are two defects, and the first is not quite the one the issue describes. `baseUrl` **is** read today, at `resolution.py:146-147`:

```python
base_url = compiler_options.get("baseUrl") or "."
paths_base = base_dir / base_url
```

but only as the base that `paths` targets resolve against. `_read_tsconfig_aliases` returns *only* `paths` entries, so a config declaring `baseUrl` and **no `paths`** — exactly the webpacker layout — yields an empty alias map and every bare specifier dies.

The second is as reported: `_load_tsconfig_aliases` probed only `tsconfig.json`, never `jsconfig.json`, even though `json_config.py` already indexes `jsconfig.json`. That is why the config's own nodes appear in the graph while resolution ignores the file.

## Fix

- `_find_js_config` walks up probing `tsconfig.json` then `jsconfig.json`, tsconfig winning within a directory — matching tsc and editors, which consult `jsconfig.json` only when there is no `tsconfig.json`.
- `_load_tsconfig_base_url` exposes the resolved baseUrl directory, cached per config path.
- `_resolve_tsconfig_alias` gains `base_url=None`, used as a resolution root of **last resort**: only when no declared alias matches. Threaded through static resolution and the three regex-rescue dynamic-import paths (Svelte, Astro, TS/TSX), since the issue reports both.

Two deliberate design choices worth flagging for review:

**baseUrl is not modelled as a synthesized `"*"` alias.** That was the tempting one-line version, and it is wrong: `"*"` scores `(1, 0)` in `_match_tsconfig_alias`, while a declared non-wildcard directory-prefix alias scores `(2, -len(prefix))`. Since `_resolve_tsconfig_alias` picks the *smallest* tuple, `(1, 0) < (2, -4)` means an implicit `"*"` would silently shadow a declared alias like `"@lib": ["src/lib"]` and regress #1269. Keeping baseUrl out of the specificity contest entirely is what makes existing precedence provably untouched. `test_declared_directory_prefix_alias_still_wins` pins this.

**The fallback returns a candidate only when it is a real file on disk.** The declared-alias path deliberately returns a non-resolving first candidate to preserve phantom/external-edge behavior; doing that for baseUrl would turn `import React from 'react'` into a fabricated `<baseUrl>/react` edge. `test_external_package_not_fabricated_under_baseurl` pins this.

## Verification

End-to-end on the reported layout (`jsconfig.json` with `"baseUrl": "app/javascript"`, `packs/dashboard.js` importing `'mods/Widget.js'`).

Before — `dashboard.js` has no outbound edges at all:

```
edges into Widget: 1
   app_javascript_mods_widget -> app_javascript_mods_widget_widget ( contains )
--- what dashboard points at instead:
(nothing)
```

After:

```
edges into Widget: 3
   app_javascript_mods_widget -> app_javascript_mods_widget_widget ( contains )
   app_javascript_packs_dashboard -> app_javascript_mods_widget ( re_exports )
   app_javascript_packs_dashboard -> app_javascript_mods_widget_widget ( imports )
```

| Check | Result |
|---|---|
| New tests before the fix | 7 of 11 fail (real negative controls) |
| New tests after | 11 passed |
| Full suite, base `66d8110` | 15 failed, 3620 passed |
| Full suite, this branch | 15 failed, 3631 passed |
| Failure-set diff | identical, no new failures |
| `ruff check` | passed |
| `python -m tools.skillgen --check` | OK, 134 artifacts match |

The 15 pre-existing failures are unrelated (`test_ollama.py` backend detection and similar), identical before and after, measured with `-p no:randomly` in the same env.

Four of the eleven tests pass both before and after by design — they are preservation guards: declared `paths` and directory-prefix aliases are not shadowed, relative imports are untouched, an absent `baseUrl` changes nothing, and an external package is not fabricated.

## Known gap

`baseUrl` inherited through an `extends` chain is not followed — `_load_tsconfig_base_url` reads it only from the directly-found config, whereas alias loading does follow `extends`. In that case the fallback simply does not apply, degrading to today's behavior (unresolved import) rather than producing a wrong edge. I left it out to keep this diff focused on the reported case, but it is an asymmetry and I'm happy to follow the `extends` chain for `baseUrl` here if you'd prefer it in one PR.

### DIFF
--- a/graphify/extract.py
+++ b/graphify/extract.py
@@ -91,6 +91,7 @@ from graphify.extractors.resolution import (  # noqa: E402,F401
     _js_source_path,
     _js_top_level_function_bodies,
     _load_tsconfig_aliases,
+    _load_tsconfig_base_url,
     _load_workspace_packages,
     _match_tsconfig_alias,
     _merge_decl_def_classes,
@@ -1285,6 +1286,7 @@ def extract_svelte(path: Path) -> dict:
         # endpoint is a phantom node and build_from_json drops the edge (#701).
         file_node_id = _make_id(str(path))
         aliases = _load_tsconfig_aliases(path.parent)
+        base_url = _load_tsconfig_base_url(path.parent)
         for m in _re.finditer(r"""import\(\s*['"]([^'"]+)['"]\s*\)""", src):
             raw = m.group(1)
             if not raw:
@@ -1302,7 +1304,7 @@ def extract_svelte(path: Path) -> dict:
                 # Check tsconfig.json path aliases (e.g. "$lib/" -> "src/lib/", "@/" -> "src/")
                 # before treating as external. Mirrors _import_js logic so SvelteKit alias
                 # imports resolve to the same file node IDs the extractor creates (#701).
-                resolved_alias = _resolve_tsconfig_alias(raw, aliases)
+                resolved_alias = _resolve_tsconfig_alias(raw, aliases, base_url=base_url)
                 if resolved_alias is not None:
                     resolved_alias = _resolve_js_module_path(resolved_alias)
                     node_id = _make_id(str(resolved_alias))
@@ -1360,7 +1362,7 @@ def extract_svelte(path: Path) -> dict:
                     node_id = _make_id(str(resolved))
                     stub_source_file = str(resolved)
                 else:
-                    resolved_alias = _resolve_tsconfig_alias(raw, aliases)
+                    resolved_alias = _resolve_tsconfig_alias(raw, aliases, base_url=base_url)
                     if resolved_alias is not None:
                         node_id = _make_id(str(resolved_alias))
                         stub_source_file = str(resolved_alias)
@@ -1413,6 +1415,7 @@ def extract_astro(path: Path) -> dict:
         existing_ids = {n["id"] for n in result.get("nodes", [])}
         file_node_id = _make_id(str(path))
         aliases = _load_tsconfig_aliases(path.parent)
+        base_url = _load_tsconfig_base_url(path.parent)
         # Dynamic imports anywhere in the file: `import('./X.astro')` is legal in
         # frontmatter setup code and inside expression slots.
         for m in _re.finditer(r"""import\(\s*['"]([^'"]+)['"]\s*\)""", src):
@@ -1425,7 +1428,7 @@ def extract_astro(path: Path) -> dict:
                 node_id = _make_id(str(resolved))
                 stub_source_file = str(resolved)
             else:
-                resolved_alias = _resolve_tsconfig_alias(raw, aliases)
+                resolved_alias = _resolve_tsconfig_alias(raw, aliases, base_url=base_url)
                 if resolved_alias is not None:
                     resolved_alias = _resolve_js_module_path(resolved_alias)
                     node_id = _make_id(str(resolved_alias))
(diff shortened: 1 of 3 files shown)


safishamsi:
Thanks @Rishet11. Shipped in v0.9.27 (cherry-picked to v8 to keep your authorship). Closed-unmerged here, but it's in the release: https://github.com/Graphify-Labs/graphify/releases/tag/v0.9.27

PR #2178, opened 2026-07-25
fix: classify repeated edge occurrences

hsneves:
Depends on #2177. The new work in this stacked PR is commit `fb77da2`.

## Summary
- preserve portable source occurrence evidence for Python type references and calls
- normalize edges after endpoint remapping, separating legitimate repeated occurrences from producer duplicates
- classify zero-node extraction outcomes and treat unrecognized JSON data as intentional skips
- expose occurrence evidence in query/path/explain, HTML, GraphML, Cypher, Neo4j, and FalkorDB without collapsing parallel edges
- extend diagnostics and generated skills with the new classifications

## Validation
- `uv run ruff check .`
- `uv run python -m tools.skillgen --schema-singleton`
- `uv run python -m tools.skillgen --monolith-roundtrip`
- `uv run pytest -q` (3607 passed, 33 skipped)
- real monorepo AST baseline: 457 duplicate occurrences reconciled as 418 legitimate + 39 producer duplicates; 14 intentional JSON skips; zero unclassified duplicates


hsneves:
Follow-up validation on the real monorepo exposed and fixed a Python 3.10 packaging gap: `tomli` was development-only, so installed CLI environments could not parse `pyproject.toml`. Commit `5430a55` moves it to runtime dependencies. Final validation: wheel build succeeded; full suite is now 3608 passed / 33 skipped; the installed Python 3.10 CLI materialized both manifests and the final corpus has zero unexpected-empty files.

Issue #522, opened 2026-04-23
codex install writes bash-only PreToolUse hook that fails on Windows PowerShell


Userchenentao5:
﻿## Summary
`graphify codex install` writes a bash-style `PreToolUse` command into `.codex/hooks.json`. On Windows, Codex executes that hook under PowerShell, so the hook fails before every matching tool invocation with `hook exited with code 1`.

## Affected environment
- graphifyy: 0.4.25 (installed locally)
- OS: Windows
- Shell used by Codex CLI: PowerShell

## Generated hook
```json
{
  "hooks": {
    "PreToolUse": [
      {
        "matcher": "Bash",
        "hooks": [
          {
            "type": "command",
            "command": "[ -f graphify-out/graph.json ] && echo '{\"hookSpecificOutput\":{\"hookEventName\":\"PreToolUse\",\"additionalContext\":\"graphify: Knowledge graph exists. Read graphify-out/GRAPH_REPORT.md for god nodes and community structure before searching raw files.\"}}' || true"
          }
        ]
      }
    ]
  }
}
```

## Steps to reproduce
1. On Windows, run `graphify codex install` in a repo.
2. Ensure `graphify-out/graph.json` exists.
3. Let Codex invoke a matching shell/tool command in that repo.

## Actual result
The hook fails with exit code 1 before command execution.

PowerShell parses the generated command as invalid syntax, for example:
- `Missing type name after '['`
- `The token '&&' is not a valid statement separator in this version`
- `The token '||' is not a valid statement separator in this version`

## Expected result
`graphify codex install` should generate a hook command that works in the shell Codex uses on Windows, or emit a shell-agnostic command.

## Notes
A local workaround is to replace the generated command with a PowerShell-safe equivalent such as:

```powershell
if (Test-Path 'graphify-out/graph.json') {
  Write-Output '{"hookSpecificOutput":{"hookEventName":"PreToolUse","additionalContext":"graphify: Knowledge graph exists. Read graphify-out/GRAPH_REPORT.md for god nodes and community structure before searching raw files."}}'
}
```

I am intentionally omitting local file paths, usernames, and screenshots for privacy.

rexkoh425:
Please update to v0.6.8 and try. Thanks.

1) pip install --upgrade graphifyy
2) graphify codex install

#651

safishamsi:
Fixed on `v8` (f7911fd), shipping in the next release.

[...]

PR #2185, opened 2026-07-25
fix: document the Codex PreToolUse hook as an intentional no-op

### LINKED ISSUE #2165: codex install registers `graphify hook-check`, which is not a command in 0.9.25 (silent no-op)
## Affected version

- `graphifyy` 0.9.25 (uv tool install)
- Windows 11, Codex CLI

## Summary

`graphify codex install` registers a `PreToolUse` hook that invokes `graphify hook-check`. That subcommand does not exist in 0.9.25. The result is a silent no-op that costs a process spawn on every matching tool call, and gives the user a hook they believe is enforcing graph usage when it is doing nothing.

## What gets written

`~/.codex/hooks.json`:

```json
{
  "matcher": "Bash",
  "hooks": [
    { "type": "command", "command": "c:/users/<user>/.local/bin/graphify.EXE hook-check" }
  ]
}
```

## Evidence it is not a real command

`graphify --help` lists the hook-related commands as:

```
hook install            install post-commit/post-checkout git hooks (all platforms)
hook uninstall          remove git hooks
hook status             check if git hooks are installed
check-update <path>     check needs_update flag and notify if semantic re-extraction is pending (cron-safe)
```

`hook-check` appears nowhere in `--help`. Invoking it directly, both outside and inside a repo with a valid `graphify-out/graph.json`:

```console
$ graphify hook-check
$ echo $?
0
```

No output, exit 0 — indistinguishable from success, so nothing surfaces the misconfiguration.

## Expected

Either:

1. the installer registers whatever the current equivalent is (`check-update`?), or
2. the hook entry is dropped if there is no longer a per-tool-call check, or
3. `hook-check` is restored as an alias.

In all cases an unrecognized subcommand should exit non-zero rather than 0, so a stale hook fails loudly instead of silently.

## Impact

Low severity, but the failure mode is bad: the user is told "Codex will now check the knowledge graph before answering codebase questions" while the mechanism that would enforce it is inert. I only found it by grepping `--help` after the hook never appeared to do anything.

Souptik96:
Fixes #2165

## What I found (the report's diagnosis is not quite right)

`hook-check` **is** a real command, and its no-op behavior is deliberate — not a stale
subcommand falling through. Verified against `v8` (0.9.26):

- `graphify/cli.py` dispatches it explicitly, with the reason inline:
  *"Codex Desktop rejects hookSpecificOutput.additionalContext on PreToolUse. Keep this as a
  cross-platform no-op so installed hooks never break Bash tool calls. Graph guidance reaches
  the agent via AGENTS.md / skill instead."*
- `graphify/__main__.py` lists `hook-check` in `_silent_cmds` ("runs on every editor tool use
  and must be silent"), and CHANGELOG has a past fix specifically for its silence.
- `__main__.py` also states the split directly: *"Claude Code accepts additionalContext on
  PreToolUse (Codex Desktop does not — that path stays a no-op via `hook-check`)."*

Two of the report's other points also don't reproduce on 0.9.26:

```console
$ graphify totally-bogus-subcommand
error: unknown command 'totally-bogus-subcommand'
Run 'graphify --help' for usage.
$ echo $?      # -> 1
```

So an unrecognized subcommand **already** exits non-zero; `hook-check` returning instantly
with exit 0 is it working as designed, not falling through. And because
`hook-check` never reads stdin, it can't hang the way `hook-guard` does — so the missing
`timeout` field isn't a hazard on this entry.

**I therefore did not repoint the Codex installer at `hook-guard`.** That's what the issue
suggests, but `hook-guard` emits `additionalContext`, which is exactly what Codex Desktop
rejects — it would reintroduce the #522-class breakage this no-op exists to avoid.

## The real problem: the docs and the installer over-promise

What legitimately misleads users (and produced this report) is that both the README and the
installer's own output describe the Codex hook as if it enforced graph usage:

- **README** claimed the Codex hook "fires before every Bash tool call, same always-on
  mechanism as Claude Code" — it is not the same mechanism; on Codex the hook is inert and
  `AGENTS.md` does the work. Now stated accurately, with the reason.
- **`_install_codex_hook`** printed `PreToolUse hook registered (... hook-check)` with no hint
  the entry is intentionally inert. It now says so.

## Plus the guard the issue implicitly asks for

New test `test_codex_hook_command_is_a_real_cli_subcommand` reads the command back out of the
generated `.codex/hooks.json` and asserts its subcommand is one the CLI actually dispatches.
If a hook command is ever renamed out from under an installer, the suite fails instead of
shipping a permanently dead hook.

Verified it actually catches that (temporarily pointed the installer at `hook-check-BOGUS`):

```
E   AssertionError: codex hook registers 'hook-check-BOGUS', which the CLI does not dispatch
    (#2165). Known commands: ['add', 'affected', ..., 'hook', 'hook-check', 'hook-guard', ...]
```

## Testing

```console
$ uv run pytest tests/test_install.py::test_codex_hook_command_is_a_real_cli_subcommand -q
1 passed, 1 warning in 0.46s

$ uv run pytest tests/test_install.py tests/test_install_references.py tests/test_install_roundtrip.py tests/test_hooks.py -q
12 failed, 207 passed, 1 skipped, 1 warning in 24.09s

$ uv run ruff check --config pyproject.toml graphify/install.py tests/test_install.py
All checks passed!

```

Real installer output after the change:

```
  .codex/hooks.json  ->  PreToolUse hook registered (…/graphify.EXE hook-check - intentional
  no-op; Codex Desktop rejects additionalContext on PreToolUse, so graph guidance comes from
  AGENTS.md)
```

**On those 12 failures — they are pre-existing on Windows, not caused by this PR.** My
baseline `uv run pytest tests/ -q` on unmodified `v8` (66d8110) gives **45 failed, 3578
passed, 15 skipped**, of which exactly 12 are in these four files (`test_install.py` 2,
`test_install_references.py` 1, `test_install_roundtrip.py` 1, `test_hooks.py` 8) — the same
12, mostly `WinError 2/32` and POSIX path-separator assertions. I did not touch them, since
they're outside this issue's scope. I don't have a Linux/macOS box to confirm they're green
there, so please read that count in that light.

### DIFF
--- a/README.md
+++ b/README.md
@@ -307,7 +307,7 @@ This writes a small config file that tells your assistant to consult the knowled
 
 **CodeBuddy** does the same two things as Claude Code: writes a `CODEBUDDY.md` section telling CodeBuddy to read `graphify-out/GRAPH_REPORT.md` before answering architecture questions, and installs `PreToolUse` hooks (`.codebuddy/settings.json`) that fire before Bash search commands and file reads, nudging toward `graphify query` instead.
 
-**Codex** writes to `AGENTS.md` and also installs a `PreToolUse` hook in `.codex/hooks.json` that fires before every Bash tool call, same always-on mechanism as Claude Code.
+**Codex** writes to `AGENTS.md`, which is what actually carries the always-on graph guidance on this platform. `graphify codex install` also registers a `PreToolUse` hook in `.codex/hooks.json` (`graphify hook-check`), but that entry is deliberately a **no-op**: Codex Desktop rejects `hookSpecificOutput.additionalContext` on `PreToolUse`, so emitting a nudge there would break Bash tool calls. Unlike Claude Code — where the hook (`graphify hook-guard`) does the nudging — on Codex the hook fires and intentionally does nothing, and `AGENTS.md` is the always-on mechanism.
 
 **Kilo Code** installs the Graphify skill to `~/.config/kilo/skills/graphify/SKILL.md` and a native `/graphify` command to `~/.config/kilo/command/graphify.md`. `graphify kilo install` also writes `AGENTS.md` plus a native `tool.execute.before` plugin (`.kilo/plugins/graphify.js` + `.kilo/kilo.json` or `.kilo/kilo.jsonc` registration) so Kilo gets the same always-on graph reminder behavior through native `.kilo` config.
 
--- a/graphify/install.py
+++ b/graphify/install.py
@@ -1388,7 +1388,11 @@ def _install_codex_hook(project_dir: Path) -> None:
     existing["hooks"]["PreToolUse"] = [h for h in pre_tool if "graphify" not in str(h)]
     existing["hooks"]["PreToolUse"].extend(hook_entry["hooks"]["PreToolUse"])
     hooks_path.write_text(json.dumps(existing, indent=2), encoding="utf-8")
-    print(f"  .codex/hooks.json  ->  PreToolUse hook registered ({graphify_exe} hook-check)")
+    print(
+        f"  .codex/hooks.json  ->  PreToolUse hook registered ({graphify_exe} hook-check"
+        " - intentional no-op; Codex Desktop rejects additionalContext on PreToolUse,"
+        " so graph guidance comes from AGENTS.md)"
+    )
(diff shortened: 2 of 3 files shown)


safishamsi:

Issue #2126, opened 2026-07-23
post-commit hook's interpreter allowlist rejects backslashes → silent failure on Windows paths (despite the "covers Windows paths" comment)


mhintz1980:
## Affected version

- `graphifyy` 0.9.5 (uv tool install)
- File: `graphify/hooks.py`
  - runtime shell pattern: **line 45** of the generated `_PYTHON_DETECT` block (probe 2, `.graphify_python`)
  - the analogous probe-3 pattern at **line 83** is the same class without the `:`
  - install-time Python regex + comment: **lines 502 & 505**
- Shell: `/bin/sh` (the hook's shebang), also reproduced under `bash`

## Summary

The post-commit git hook's interpreter-detection allowlist is supposed to admit Windows paths (`C:\Users\...`), and the inline comment at install time explicitly claims it does:

> `# The allowlist includes ':' and '\' so Windows paths (C:\...) are accepted.` (hooks.py line 502)

At **install time** this is true — line 505 uses Python `re`, where `\` in a character class is a literal backslash, so the pinned path is accepted.

At **runtime** (when git fires the hook) this is false — line 45 uses a POSIX shell `case` glob, where `\-` inside `[...]` is parsed as "backslash-escaping the hyphen" (making `-` a literal hyphen, not a range operator). The backslash character itself is never added to the allowed set. Net effect: any Windows path containing `\` is rejected by probe 2 (`.graphify_python`) and probe 3 (shebang), and the hook falls through to the "could not locate a Python" warning.

The same character sequence (`[!a-zA-Z0-9/_.@:\-]`) has opposite meaning in the two regex dialects — that is the bug.

## Reproduce

1. Install graphify on Windows (uv tool): `uv tool install graphifyy`
2. From a WSL/Linux shell on a shared repo (or any cross-host scenario), run `graphify hook install`. The `_PINNED` placeholder gets a Linux path (`/home/.../python3`) — this is how the mismatch hides: the Linux pinned path is valid, so probe 1 succeeds and the bug in probe 2 is never exercised on the Linux side.
3. On Windows, from Git-Bash-on-Windows, the `_PINNED` Linux path does not exist. The hook falls to probe 2, which reads `graphify-out/.graphify_python`. On a Windows install that file contains the backslash form: `C:\Users\<user>\AppData\Roaming\uv\tools\graphifyy\Scripts\python.exe`
[...]

PR #2186, opened 2026-07-25
fix: pin interpreters whose path contains a space

### LINKED ISSUE #2166: hook install emits empty _PINNED on Windows uv installs, so every post-commit rebuild silently no-ops
## Affected version

- `graphifyy` 0.9.25 (uv tool install, installed **from Windows**)
- Windows 11, Git for Windows (`/bin/sh`)

## Summary

`graphify hook install` on a Windows `uv tool` install emits a post-commit hook with an **empty** `_PINNED`:

```sh
_PINNED=''
```

With `_PINNED` empty, all four interpreter probes fail on this setup, so every commit prints the "could not locate a Python" warning and exits 0. The graph silently never rebuilds.

This is **not** #2126. That issue covers the runtime allowlist rejecting backslashes in probes 2 and 3, and explicitly assumes "a Windows-only repo re-installed from Windows would pin a `C:\...` path into `_PINNED`, so probe 1 would succeed". On my Windows-only install, `hook install` run from Windows recorded nothing at all — probe 1 never gets the chance to succeed. Fixing #2126 alone would not fix this.

## Why all four probes fail (uv tool layout)

| # | Probe | Result |
|---|---|---|
| 1 | `_PINNED` | empty — never recorded |
| 2 | `graphify-out/.graphify_python` | file not created by `hook install` |
| 3a | `$(dirname $(command -v graphify))/../python.exe` | `~/.local/python.exe` — does not exist |
| 3b | `$(dirname $(command -v graphify))/python.exe` | `~/.local/bin/python.exe` — does not exist |
| 3c | shebang parse | skipped, launcher is `.exe` |
| 4 | `python` / `python3` on PATH | present, but cannot import graphify |

`uv` puts the launcher shim in `~/.local/bin/` and the actual environment in `%APPDATA%\uv\tools\graphifyy\`. Nothing in probes 3a/3b resolves across that gap, and the system `python` is unrelated (Anaconda here).

Verified:

```console
$ python  -c "import importlib.util,sys; sys.exit(0 if importlib.util.find_spec('graphify') else 1)"; echo $?
1
$ python3 -c "import importlib.util,sys; sys.exit(0 if importlib.util.find_spec('graphify') else 1)"; echo $?
1
$ ls ~/.local/bin/python.exe ~/.local/python.exe
ls: cannot access ...: No such file or directory
$ ls "$APPDATA/uv/tools/graphifyy/Scripts/python.exe"
...Scripts/python.exe        # the one that works
```

## Reproduce

1. Windows, `uv tool install graphifyy`
2. In any repo: `graphify hook install`
3. `grep '^_PINNED=' .git/hooks/post-commit` -> `_PINNED=''`
4. Commit twice (the first exits early on the `HEAD~1` check). Second commit prints:

```
[graphify hook] could not locate a Python with graphify installed. Add the graphify bin dir to PATH or re-run 'graphify hook install' from the env where graphify lives.
```

The advice in that message does not help: re-running `hook install` from the env where graphify lives is exactly what produced the empty `_PINNED`.

## Impact

Silent and total — the post-commit rebuild never runs, so graphs go stale with no signal. I installed hooks across 91 repos before noticing that not one of them could execute.

## Suggested fix

At install time, pin `sys.executable` of the interpreter running the installer — it is by definition an interpreter that can import graphify:

```python
_PINNED = shlex.quote(sys.executable.replace("\\", "/"))
```

Forward slashes keep it compatible with the probe-2/3 allowlist discussed in #2126. Writing `graphify-out/.graphify_python` at `hook install` time (rather than only from the skill/CLI) would also give probe 2 something to find.

## Workaround

Write the interpreter path into `graphify-out/.graphify_python` manually:

```
%APPDATA%\uv\tools\graphifyy\Scripts\python.exe
```

## Related

#2126 (same subsystem, different root cause — runtime allowlist vs. install-time recording).

Souptik96:
Fixes #2166

## Root cause (a bit narrower than the report)

The empty `_PINNED` comes from the install-time allowlist in `_pinned_python()`, not from the
uv-tool layout. It accepted `[a-zA-Z0-9/_.@:\-]` — no **space** — and rejects the path
wholesale on any non-matching character, so `sys.executable` under a profile whose name
contains a space records nothing.

Verified on Windows against `v8` (0.9.26), stubbing `sys.executable`:

```
PINNED   'C:\Users\devuser\AppData\Roaming\uv\tools\graphifyy\Scripts\python.exe'  -> pinned
EMPTY!!  'C:\Users\First Last\AppData\Roaming\uv\tools\graphifyy\Scripts\python.exe' -> ''
PINNED   '/home/u/.local/share/uv/tools/graphifyy/bin/python'                      -> pinned
EMPTY!!  'C:\Users\evil\py.exe; rm -rf /'                                          -> ''   (correct)
EMPTY!!  "C:\Users\ev'il\python.exe"                                               -> ''   (correct)
```

So a **space-free** Windows uv path pins fine — which is why this reads as uv-layout-specific.
What actually breaks it is a space anywhere in the interpreter path: `C:\Users\First Last\...`,
or the equally common `C:\Program Files\Python312\python.exe`. Once `_PINNED` is empty, probe 1
is skipped and (per the report's own table) probes 2–4 can't recover on a uv install, so every
commit warns and no-ops.

That also means the issue's other suggestion — writing `graphify-out/.graphify_python` at hook
install time — isn't needed to fix this; probe 1 works once it's actually populated. I left it
out to keep the change minimal.

## Fix

- **`_pinned_python()`: allow a plain space.** This is safe because every consumer already
  quotes the value — the hook embeds `_PINNED='<path>'` (single-quoted) and dereferences
  `"$_PINNED"`, so a space can neither split a word nor start a command. `$`, backtick, `;`,
  `'` and `"` remain rejected, so the injection guard is unchanged.
- **`_register_merge_driver()`: quote the interpreter.** It interpolated the path *unquoted*
  into `merge.graphify.driver`, which git runs through a shell — a spaced path would split into
  two words and the driver would never run. Now double-quoted, which is safe precisely because
  the allowlist keeps `$`/backticks out.

I deliberately did **not** convert the pinned path to forward slashes (the issue suggests
`sys.executable.replace("\\", "/")`). Probe 1 quotes `"$_PINNED"` so backslashes are fine there,
`#2126` already established that a pinned `C:\...` path is the expected shape, and rewriting it
would churn the existing `test_install_embeds_pinned_interpreter` assertion for no gain.

## Testing

New tests in `tests/test_hooks.py`:

```console
$ uv run pytest tests/test_hooks.py -q -k "pinned_python_accepts_paths_containing_spaces or pinned_python_still_rejects or merge_driver_quotes_interpreter or install_pins_interpreter_path_with_spaces"
11 passed, 67 deselected, 1 warning in 1.29s
```

Confirmed they actually catch the bug — same tests with `graphify/hooks.py` reverted to `v8`:

```
FAILED tests/test_hooks.py::test_pinned_python_accepts_paths_containing_spaces[C:\Users\First Last\...\python.exe]
FAILED tests/test_hooks.py::test_pinned_python_accepts_paths_containing_spaces[C:\Program Files\Python312\python.exe]
FAILED tests/test_hooks.py::test_pinned_python_accepts_paths_containing_spaces[/home/first last/...]
FAILED tests/test_hooks.py::test_merge_driver_quotes_interpreter_with_spaces - interpreter not quoted in merge driver: 'graphify merge-dri...'
FAILED tests/test_hooks.py::test_install_pins_interpreter_path_with_spaces - post-commit did not pin the spaced interpreter
5 failed, 73 deselected
```

Affected files plus both pre-commit gates:

```console
$ uv run pytest tests/test_hooks.py tests/test_hook_strict.py tests/test_install.py -q
11 failed, 172 passed, 1 warning in 19.40s

$ uv run ruff check --config pyproject.toml graphify/hooks.py tests/test_hooks.py
All checks passed!

```

**Those 11 failures are pre-existing on Windows, not from this PR.** Baseline `uv run pytest
tests/ -q` on unmodified `v8` (66d8110) is **45 failed, 3578 passed, 15 skipped**, of which
exactly 11 live in these three files (`test_hooks.py` 8, `test_hook_strict.py` 1,
`test_install.py` 2) — the same 11, and most of the `test_hooks.py` ones are the
`bash`-dependent allowlist tests plus `WinError` cases. I have no Linux/macOS box to confirm
they're green there, so please read that number in that light.

### DIFF
--- a/graphify/hooks.py
+++ b/graphify/hooks.py
@@ -503,10 +503,17 @@ def _pinned_python() -> str:
     that is not a valid plain filesystem path character, preventing $(...),
     backtick, double-quote, semicolon, etc. from being injected into generated
     shell scripts or the merge-driver command line. The allowlist includes ':'
-    and '\\' so Windows paths (C:\\...) are accepted. An empty return means
-    callers must fall back to the `graphify` launcher on PATH — safe degradation.
+    and '\\' so Windows paths (C:\\...) are accepted, and a plain space so
+    Windows profile paths (C:\\Users\\First Last\\...) are too — a space cannot
+    start a substitution or a new command, and every consumer quotes the value:
+    the hook scripts embed it as '$_PINNED' (single-quoted, then referenced as
+    "$_PINNED") and _register_merge_driver double-quotes it (#2166). Before that
+    a space rejected the whole path, so hooks installed under any Windows user
+    whose profile name contains a space silently pinned nothing. An empty return
+    means callers must fall back to the `graphify` launcher on PATH — safe
+    degradation.
     """
-    if re.search(r"[^a-zA-Z0-9/_.@:\\-]", sys.executable):
+    if re.search(r"[^a-zA-Z0-9/_.@: \\-]", sys.executable):
         return ""
     return sys.executable
 
@@ -551,7 +558,12 @@ def _register_merge_driver(root: Path) -> str:
     import subprocess as _sp
     pinned = _pinned_python()
     if pinned:
-        driver = f"{pinned} -m graphify merge-driver %O %A %B"
+        # Double-quoted: the allowlist in _pinned_python() permits a space (Windows
+        # profile paths), and git runs this driver string through a shell, so an
+        # unquoted "C:\\Users\\First Last\\...\\python.exe" would split into two
+        # words and the driver would never run (#2166). The same allowlist keeps
+        # '$' and backticks out, so double quotes cannot introduce expansion.
+        driver = f'"{pinned}" -m graphify merge-driver %O %A %B'
     else:
         driver = "graphify merge-driver %O %A %B"
     try:
--- a/tests/test_hooks.py
+++ b/tests/test_hooks.py
@@ -676,3 +676,82 @@ def test_uninstall_removes_merge_driver_keeps_other_attrs(tmp_path):
     content = (repo / ".gitattributes").read_text(encoding="utf-8")
     assert "*.png binary" in content
     assert "merge=graphify" not in content
+
+
+@pytest.mark.parametrize("exe", [
+    r"C:\Users\First Last\AppData\Roaming\uv\tools\graphifyy\Scripts\python.exe",
+    r"C:\Program Files\Python312\python.exe",
+    "/home/first last/.local/share/uv/tools/graphifyy/bin/python",
+])
+def test_pinned_python_accepts_paths_containing_spaces(exe, monkeypatch):
+    """#2166: a space must not empty the pin.
+
+    The install-time allowlist had no space, so `sys.executable` under any Windows
+    profile whose name contains one (`C:\\Users\\First Last\\...`, or the very common
+    `C:\\Program Files\\...`) was rejected wholesale and the hook shipped `_PINNED=''`.
+    Every interpreter probe then failed and each commit no-op'd with the "could not
+    locate a Python" warning, so the graph never rebuilt.
+    """
+    import sys as _sys
+
+    from graphify.hooks import _pinned_python
+
+    monkeypatch.setattr(_sys, "executable", exe)
+    assert _pinned_python() == exe, "a path containing a space must still be pinned"
+
+
+@pytest.mark.parametrize("exe", [
+    r"C:\Users\evil\python.exe; rm -rf /",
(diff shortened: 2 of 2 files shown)


safishamsi:

Issue #1910, opened 2026-07-15
ERROR nodes on PostgreSQL/PLpgSQL functions


DutchGerman:
The sql extractor does not extract functions from PostgreSQL/PLpgSQL files when `tree-sitter-sql` parses the statement as an `ERROR` node.

In our repo, sql files are included in `graph.json`, but important PostgreSQL functions are still missing. 

Example:

```sql
CREATE FUNCTION exposed.important_function(foo text, bar text, OUT fuzz text)
LANGUAGE plpgsql AS $$
BEGIN
  PERFORM api.do_something_important();
  fuzz := return_something_important();
END$$;
````

`tree-sitter-sql` seems to parse this as an ERROR node.

It would be nice if we could get [`tree-sitter-postgres` ](https://github.com/gmr/tree-sitter-postgres) support

Sirhan1:
Hey I'd like to take a look at this if nobody else is already working on it. I'll investigate why these PostgreSQL/PLpgSQL functions are ending up as ERROR nodes and see what the best approach to supporting them would be.

safishamsi:
Fixed on v8 in 49df466 (ships in 0.9.17). The SQL extractor recovers CREATE FUNCTION/PROCEDURE names from tree-sitter ERROR nodes (PL/pgSQL bodies with OUT params, tagged dollar-quotes, PERFORM/:= parse as ERROR), keeping the function node while leaving the unparseable body opaque; surrounding statements are unaffected.

PR #2188, opened 2026-07-25
fix: recover every declared SQL routine from unparseable PL/pgSQL

### LINKED ISSUE #2180: SQL extractor silently drops CREATE FUNCTION with PL/pgSQL-only bodies (PERFORM / := still dropped after #1910)
## Summary

The SQL extractor still silently drops `CREATE FUNCTION` definitions whose PL/pgSQL body contains PL/pgSQL-only statements. On a real (private) PostgreSQL schema this loses **57 of 186 functions (31%)** from a single file, with **no warning and exit code 0** — so a consumer cannot tell the graph is incomplete.

This looks like an incomplete fix for #1910 ("ERROR nodes on PostgreSQL/PLpgSQL functions", closed by `49df466`, shipped 0.9.17). That fix's closing note says the extractor now recovers names from ERROR nodes for "PL/pgSQL bodies with OUT params, tagged dollar-quotes, PERFORM/:= parse as ERROR". **`PERFORM` and `:=` still drop in 0.9.26** — see the matrix below.

**Version:** graphify 0.9.26 (`graphify update <path> --no-cluster`), macOS (darwin 25.5.0), Node 24.16.

## Minimal reproduction

Two files, identical except for one statement in the body. Both are valid PostgreSQL, both `RETURNS "void"`.

`migrations/20260101000001_dropped.sql`:

```sql
CREATE OR REPLACE FUNCTION "public"."probe_no_return"("p_id" "uuid") RETURNS "void"
    LANGUAGE "plpgsql" SECURITY DEFINER
    SET "search_path" TO 'public'
    AS $$
BEGIN
  RAISE EXCEPTION 'insufficient_privilege' USING ERRCODE = '42501';
END;
$$;
```

`migrations/20260101000002_extracted.sql`:

```sql
CREATE OR REPLACE FUNCTION "public"."probe_with_return"("p_id" "uuid") RETURNS "void"
    LANGUAGE "plpgsql" SECURITY DEFINER
    SET "search_path" TO 'public'
    AS $$
BEGIN
  RETURN;
END;
$$;
```

Then:

```
[graphify watch] Rebuilt (no clustering): 3 nodes, 1 edges
```

**Expected:** 4 nodes (2 file nodes + 2 function nodes).
**Actual:** 3 nodes — only `"public"."probe_with_return"()` is present. `probe_no_return` is absent, with no diagnostic.

## Which body statements decide it

Same 7-line function each time, `RETURNS "void"`, varying only the single body statement:

| Body statement | Result |
|---|---|
| `RETURN;` | extracted |
| `INSERT INTO public.t (a) VALUES (1);` | extracted |
| `UPDATE public.t SET a = 1;` | extracted |
| `NULL;` | **dropped** |
| `RAISE EXCEPTION 'boom';` | **dropped** |
| `RAISE NOTICE 'hi';` | **dropped** |
| `PERFORM public.other_fn();` | **dropped** |
| `x := 1;` | **dropped** |
| `IF true THEN RAISE EXCEPTION 'x'; END IF;` | **dropped** |

Bodies parseable as plain SQL survive; bodies using PL/pgSQL-only constructs are dropped. `PERFORM` and `:=` are the two cases #1910's fix specifically claims to handle.

## Impact at scale

Measured on one generated schema file: ~35,000 lines, **186** `CREATE OR REPLACE FUNCTION` declarations, single schema, no overloads.

- **186 declared → 129 extracted → 57 silently missing (30.6%).**
- Of the 57 missing, **44** have bodies using PL/pgSQL-only constructs (`RAISE`/`PERFORM`/`:=`/`IF…THEN`/`DECLARE`/`LOOP`/bare `NULL;`).
- The predicate is **not** clean at that scale, though: 76 functions that also use those constructs extract fine, and 13 of the missing use none of them. So the single-statement repro above is deterministic, but full-file behaviour likely also depends on the statement mix and where tree-sitter's error recovery resyncs. I did not narrow that further.

Ruled out as the discriminator (each checked across all 186 declarations): position in file / truncation, dollar-quote tag (`$$` vs `$_$`), `RETURNS` type, `SET search_path` shape, argument count, identifier length, body length, gap to the next declaration, redefinition in a later file, node-id collisions, and quote-escaping (`''`) inside a trailing `COMMENT ON FUNCTION`.

## Why this matters more than the coverage number

The failure is **silent**. Exit code is 0, no ERROR node is surfaced, and the file node is created normally — so the graph looks healthy while a third of the schema's callable surface is missing.

That makes a negative result untrustworthy: "graphify has no node for function X" cannot be distinguished from "function X does not exist". We have had to tell our agents that **a hit is trustworthy but a miss is not proof of absence**, and to fall back to `grep`/`pg_proc` for any database-object question — which removes much of the graph's value for the SQL layer specifically.

Same shape of complaint as #1689 (R files silently dropped with zero warning).

## Suggested resolution (in preference order)

1. Extend the #1910 ERROR-node name recovery to cover these bodies — at minimum `RAISE`, `PERFORM`, `:=`, `IF…THEN`, and bare `NULL;`, which are ordinary PL/pgSQL rather than edge cases.
2. Failing that, **make the drop loud**: emit a per-file warning with a count and the offsets it could not recover (e.g. `extracted N of M CREATE FUNCTION statements in <file>`). A visible partial result is far more useful than a silent one, and lets CI gate on it.
3. Longer term, the `tree-sitter-postgres` grammar suggested in #1910 addresses the class rather than the symptom.

Happy to test a patch against the 186-function schema — it is a good stress case, and I can report the before/after extraction count.

Souptik96:
Fixes #2180

## Root cause: three parse shapes, only one was recovered

#1910 added ERROR-node name recovery, but PL/pgSQL breaks the parse in more than one
shape and the recovery only handled the first:

1. **The whole `CREATE` lands in one `ERROR` node** — recovered before this PR.
2. **The statement is shredded into loose top-level tokens.** For a body using `PERFORM`
   or `:=`, tree-sitter emits `keyword_create`, `keyword_or`, `keyword_replace`,
   `keyword_function`, `object_reference`, `function_arguments`, … `keyword_begin`, and then
   an `ERROR` node containing **only the offending body line**. Dumping the tree for the
   fixture in this PR:

   ```
   keyword_function     line  27  FUNCTION
   object_reference     line  27  "public"."perform_fn"
   ...
   keyword_begin        line  28  BEGIN
   ERROR                line  29  PERFORM "public"."raise_notice_fn"();
   ```

   No `ERROR` node contains any `CREATE` text, so scanning ERROR nodes finds nothing.
   **This is why `PERFORM` and `:=` still dropped after #1910.**
3. **The routine name is a quoted identifier.** The recovery matched a bare `[\w$.]+`,
   which stops dead at the leading quote, so `CREATE OR REPLACE FUNCTION "public"."fn"(...)`
   matched nothing at all. Generated dumps quote every identifier, so entire files
   recovered zero routines.

Confirmed shape 3 directly against `v8` with your repro — the same PL/pgSQL body recovers
under an *unquoted* name and drops under a quoted one:

```
20260101000001_dropped.sql:   1 node(s) -> ['20260101000001_dropped.sql']                      # dropped
20260101000002_extracted.sql: 2 node(s) -> [..., '"public"."probe_with_return"()']
20260101000003_unquoted.sql:  2 node(s) -> [..., 'public.probe_unquoted()']                    # recovered!
```

That's the missing piece in the issue's analysis: the body decides whether you hit the
error path at all, but the **quoting and the shredding shape** decide whether recovery can
see the routine. It also explains why the predicate wasn't clean at 186 functions —
unquoted or fully-parsed routines came through, quoted-and-shredded ones didn't.

## Fix

Option 1 from the issue, implemented the way this extractor already handles the same class
of problem. There's an existing *global regex fallback* after the tree walk for
`REFERENCES` edges ("catch any REFERENCES missed due to ERROR nodes in the parse tree"); I
mirrored it for routines: scan the raw source for every
`CREATE [OR REPLACE] FUNCTION|PROCEDURE`, with name parts accepting bare **or** double-quoted
identifiers, and emit any routine the tree walk missed. `_add_node` already dedupes by node
id, so routines recovered from the tree aren't emitted twice.

This is shape-agnostic — it fixes all three cases above, so a future grammar quirk can't
silently drop a declared routine again. After the fix your repro gives the expected 4 nodes
(2 file + 2 function).

I did **not** implement option 2 (the loud per-file "extracted N of M" warning). It's a good
idea, but a raw-text count can't tell a real declaration from `CREATE FUNCTION` inside a
comment or a string literal, so it would need its own care to avoid warning on every clean
file. Happy to add it as a follow-up if you want it — and with this fix the count should now
be N == M anyway.

## Testing

New fixture `tests/fixtures/sample_plpgsql_quoted.sql`: generated-style DDL where every
identifier is quoted and the bodies cover your matrix — `RAISE EXCEPTION`, `RAISE NOTICE`,
`PERFORM`, `:=`, `IF..THEN`, bare `NULL;`, plus a `CREATE PROCEDURE` — with tables before and
after.

```console
$ uv run pytest tests/test_multilang.py -q -k "quoted_plpgsql"
2 passed, 50 deselected, 1 warning in 0.38s

$ uv run pytest tests/test_multilang.py tests/test_languages.py tests/test_pg_introspect.py -q
391 passed, 1 warning in 2.08s
```

Confirmed the tests catch the bug — same tests with `graphify/extractors/sql.py` reverted to `v8`:

```
FAILED tests/test_multilang.py::test_sql_quoted_plpgsql_routines_are_recovered
  - AssertionError: perform_fn dropped from a quoted-DDL file (#2180)
```

Full suite (this touches an extractor, so I ran everything):

```console
$ uv run pytest tests/ -q
45 failed, 3580 passed, 15 skipped, 4 warnings in 407.51s
```

```console
$ uv run ruff check --config pyproject.toml graphify/extractors/sql.py tests/test_multilang.py
All checks passed!

```

**Those 45 failures are pre-existing on Windows, not from this PR.** My baseline on
unmodified `v8` (66d8110) is **45 failed, 3578 passed, 15 skipped** — the same 45, and passed
moves 3578 -> 3580, exactly the two tests added here. The pre-existing ones are Windows-specific
(`WinError 2/32`, POSIX path-separator assertions, `bash`-dependent tests); none are in
`test_multilang.py`. I have no Linux/macOS box to confirm they're green there.

If you'd like, I'm happy to have the before/after count checked against your 186-function
schema — that's a better stress case than any fixture I can write.

### DIFF
--- a/graphify/extractors/sql.py
+++ b/graphify/extractors/sql.py
@@ -200,9 +200,16 @@ def extract_sql(path: Path, content: str | bytes | None = None) -> dict:
             # several statements, so scan for every CREATE in it. We deliberately
             # do not scan the body for FROM/JOIN references: PL/pgSQL loop
             # variables and locals would produce junk reads_from targets.
+            #
+            # Each name part is either a bare identifier or a double-quoted
+            # (delimited) one, so schema-qualified generated DDL such as
+            # CREATE OR REPLACE FUNCTION "public"."fn"(...) is recovered too.
+            # A bare [\w$.]+ stops dead at the leading quote, which silently
+            # dropped every quoted PL/pgSQL routine (#2180).
             text = _read(node)
             for m in re.finditer(
-                r"CREATE\s+(?:OR\s+REPLACE\s+)?(?:FUNCTION|PROCEDURE)\s+([\w$.]+)",
+                r"CREATE\s+(?:OR\s+REPLACE\s+)?(?:FUNCTION|PROCEDURE)\s+"
+                r"((?:\"[^\"\n]+\"|[\w$]+)(?:\s*\.\s*(?:\"[^\"\n]+\"|[\w$]+))*)",
                 text, re.IGNORECASE,
             ):
                 name = m.group(1)
@@ -291,5 +298,26 @@ def extract_sql(path: Path, content: str | bytes | None = None) -> dict:
             if (tbl_nid, ref_nid) not in emitted:
                 _add_edge(tbl_nid, ref_nid, "references", tbl_line)
                 emitted.add((tbl_nid, ref_nid))
+
+    # Global regex fallback for routines (#2180). PL/pgSQL bodies break the parse
+    # in more than one shape, and only the first was recovered before:
+    #   1. the whole CREATE lands in one ERROR node          -> handled in walk()
+    #   2. the statement is shredded into loose top-level tokens
+    #      (keyword_create/keyword_function/object_reference/... ) and the ERROR
+    #      node holds only the offending body line, e.g. `PERFORM x();` or
+    #      `x := 1;` -- so no CREATE text is inside any ERROR node at all
+    #   3. the name is a quoted identifier ("public"."fn"), which a bare
+    #      [\w$.]+ pattern cannot match
+    # Shapes 2 and 3 silently dropped the routine: no node, no warning, exit 0.
+    # Scanning the raw source catches all three, and _add_node dedupes by id so
+    # routines already recovered from the tree are not emitted twice.
+    for m in re.finditer(
+        r"CREATE\s+(?:OR\s+REPLACE\s+)?(?:FUNCTION|PROCEDURE)\s+"
+        r"((?:\"[^\"\n]+\"|[\w$]+)(?:\s*\.\s*(?:\"[^\"\n]+\"|[\w$]+))*)",
+        src_text, re.IGNORECASE,
+    ):
+        fn_name = m.group(1)
+        fn_line = src_text[: m.start()].count("\n") + 1
+        _add_node(_make_id(stem, fn_name), f"{fn_name}()", fn_line)
 
     return {"nodes": nodes, "edges": edges}
--- a/tests/fixtures/sample_plpgsql_quoted.sql
+++ b/tests/fixtures/sample_plpgsql_quoted.sql
@@ -1,0 +1,62 @@
+-- Generated-style PostgreSQL DDL (#2180): every identifier is double-quoted and
+-- each body uses a PL/pgSQL-only statement, so tree-sitter-sql parses these as
+-- ERROR nodes. Name recovery is the only path that can see them, and a bare
+-- [\w$.]+ name pattern stops at the leading quote -- which silently dropped
+-- every routine in files shaped like this.
+
+CREATE TABLE "public"."accounts" (
+    "id" INT PRIMARY KEY,
+    "name" TEXT
+);
+
+CREATE OR REPLACE FUNCTION "public"."raise_exception_fn"("p_id" "uuid") RETURNS "void"
+    LANGUAGE "plpgsql" SECURITY DEFINER
+    SET "search_path" TO 'public'
+    AS $$
+BEGIN
+  RAISE EXCEPTION 'insufficient_privilege' USING ERRCODE = '42501';
+END;
+$$;
+
+CREATE OR REPLACE FUNCTION "public"."raise_notice_fn"() RETURNS "void" LANGUAGE "plpgsql" AS $$
+BEGIN
+  RAISE NOTICE 'hi';
+END;
+$$;
+
+CREATE OR REPLACE FUNCTION "public"."perform_fn"() RETURNS "void" LANGUAGE "plpgsql" AS $$
+BEGIN
+  PERFORM "public"."raise_notice_fn"();
+END;
+$$;
+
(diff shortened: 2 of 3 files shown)


safishamsi:
Thanks @Souptik96. Shipped in v0.9.27 (cherry-picked to v8), with a follow-up gating the raw-text recovery on a failed parse so a clean file can't fabricate routines from commented-out DDL. Closed-unmerged here, but it's in the release: https://github.com/Graphify-Labs/graphify/releases/tag/v0.9.27

PR #2192, opened 2026-07-25
fix: resolve bash sources for extensionless libs and bare names

### LINKED ISSUE #2171: Bash source resolution: extensionless shebang scripts and bare `source lib.sh` are not resolved
Follow-up to #2141 / #2157 (which resolves calls into `source`d-file functions). Two coverage gaps remain, both verified during review of #2157. Neither is a regression; each is a miss, not a fabrication.

1. **Extensionless shebang scripts.** `_SHEBANG_DISPATCH` routes `#!/usr/bin/env bash` extensionless files to `extract_bash`, but the source-resolution filter selects bash results by suffix (`p.suffix in (".sh", ".bash")`). A sourced extensionless lib gets its functions indexed but calls into it never resolve.
   - Fix direction: select bash results by shape (a node whose `metadata.language == "bash"`) instead of by suffix.
   - Test: extensionless `mylib` with a bash shebang, sourced by `a.sh`; assert the `calls` edge into `mylib`'s function.

2. **Bare `source lib.sh`** (no `./` prefix). Only the `raw.startswith((".", "/"))` branch records `bash_sources`; a bare name falls through to the `imports` fallback and never resolves, even though the file usually sits next to the script.
   - Fix direction: in the else branch, if `(path.parent / raw).is_file()`, also append a `bash_sources` entry. Low priority (bash resolves bare names via `$PATH` at runtime).

Transitive sources (a→b→c, a calls c's function) are intentionally left unresolved (conservative).

Souptik96:
Fixes #2171

Both gaps confirmed against `v8` (0.9.26) and fixed along the directions in the issue.

## 1. Extensionless shebang scripts

`_SHEBANG_DISPATCH` routes a `#!/usr/bin/env bash` file with no extension to
`extract_bash`, so its functions *are* indexed — but the cross-file source pass picked
participants by filename suffix:

```python
sh_paths = [p for p in paths if p.suffix in (".sh", ".bash")]
```

so a sourced extensionless lib was excluded and calls into it never bound.

Now selected by shape as well: the bash extractor tags every node it emits with
`metadata.language == "bash"`. I kept the suffix check alongside it so an empty `.sh` file
— which has no nodes to inspect — still participates.

## 2. Bare `source lib.sh`

Only the `raw.startswith((".", "/"))` branch recorded a `bash_sources` entry; a bare name
fell through to the opaque `imports` fallback, so neither the source edge nor calls into the
lib resolved. The `else` branch now binds a sibling of that name when one exists.

The existence gate from the `./`-prefixed branch carries over, so this stays a miss-fix and
not a fabrication: a bare name resolving to no sibling keeps the old `imports` edge and
records no `bash_sources` entry. `is_file()` is wrapped against `OSError` so a name that's
invalid for the platform degrades instead of raising.

Transitive sources (a→b→c) remain unresolved, as the issue specifies.

## Testing

Three new tests in `tests/test_extract.py` — the two the issue asks for, plus a guard that
the bare-name branch still fabricates nothing:

```console
$ uv run pytest tests/test_extract.py -q -k "extensionless_sourced_lib or bare_source_name_resolves or bare_source_missing_file"
3 passed, 149 deselected, 1 warning in 2.82s
```

Confirmed they catch the bugs — same tests with `graphify/extract.py` and
`graphify/extractors/bash.py` reverted to `v8`:

```
FAILED tests/test_extract.py::test_extract_bash_call_into_extensionless_sourced_lib_resolves - AssertionError: []
FAILED tests/test_extract.py::test_extract_bash_bare_source_name_resolves_to_sibling - AssertionError: []
2 failed, 150 deselected
```

Affected suites and both pre-commit gates:

```console
$ uv run pytest tests/test_extract.py tests/test_symbol_resolution.py tests/test_languages.py -q
4 failed, 521 passed, 1 warning in 24.15s

$ uv run ruff check --config pyproject.toml graphify/extractors/bash.py graphify/extract.py tests/test_extract.py
All checks passed!

```

**Those 4 failures are pre-existing on Windows, not from this PR** — they're exactly the 4
`test_extract.py` failures in my baseline on unmodified `v8` (66d8110), all symlink-related
(`test_collect_files_*`, e.g. `OSError: [WinError 1314] A required privilege is not held by
the client`). Full-suite baseline there is 45 failed / 3578 passed / 15 skipped. No
Linux/macOS box here to confirm they're green, so please read that in context.

### DIFF
--- a/graphify/extract.py
+++ b/graphify/extract.py
@@ -4947,9 +4947,34 @@ def extract(
     # same-named function in an unsourced file. Runs after the id-remap passes
     # above so caller_nids and function node ids are final; dedups the source
     # edge the extractor already emitted via existing_edges.
-    sh_paths = [p for p in paths if p.suffix in (".sh", ".bash")]
-    if sh_paths:
-        sh_results = [r for r, p in zip(per_file, paths) if p.suffix in (".sh", ".bash")]
+    # Selecting by filename suffix alone missed extensionless scripts:
+    # _SHEBANG_DISPATCH routes a `#!/usr/bin/env bash` file with no extension to
+    # extract_bash, so its functions get indexed, but a suffix-only filter left it
+    # out of this pass and calls into it never resolved (#2171). Select by shape
+    # too — the bash extractor tags every node it emits with
+    # metadata.language == "bash" — while keeping the suffix check so an empty
+    # .sh file (no nodes to inspect) still participates.
+    def _looks_like_bash(result: object) -> bool:
+        if not isinstance(result, dict):
+            return False
+        nodes = result.get("nodes")
+        if not isinstance(nodes, list):
+            return False
+        for n in nodes:
+            if not isinstance(n, dict):
+                continue
+            md = n.get("metadata")
+            if isinstance(md, dict) and md.get("language") == "bash":
+                return True
+        return False
+
+    sh_pairs = [
+        (r, p) for r, p in zip(per_file, paths)
+        if p.suffix in (".sh", ".bash") or _looks_like_bash(r)
+    ]
+    if sh_pairs:
+        sh_results = [r for r, _ in sh_pairs]
+        sh_paths = [p for _, p in sh_pairs]
         try:
             all_edges.extend(
                 resolve_bash_source_edges(sh_results, sh_paths, root, existing_edges=all_edges)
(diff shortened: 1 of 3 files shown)


safishamsi:
Thanks @Souptik96. Shipped in v0.9.27 (cherry-picked to v8), with a follow-up marking the bare-name binding INFERRED (it resolves via $PATH at runtime). Closed-unmerged here, but it's in the release: https://github.com/Graphify-Labs/graphify/releases/tag/v0.9.27

PR #2220, opened 2026-07-26
fix(swift): extract computed & observed properties (#2181)

### LINKED ISSUE #2181: Swift: computed properties emit no node — erases the entire SwiftUI view layer (var body is a computed property)
**Version:** graphify 0.9.25
**Language:** Swift (`_SWIFT_CONFIG`, `extract.py:984`)

## Summary

`_SWIFT_CONFIG` treats only `function_declaration` / `init_declaration` / `deinit_declaration` / `subscript_declaration` as functions. A Swift *computed property* parses as `property_declaration` → `computed_property`, which appears in none of `function_types`, `function_boundary_types`, or `body_fallback_child_types`.

Consequence: a computed property gets **no node**, and its contents are **never walked as a body**. In SwiftUI, `var body: some View { … }` is a computed property — so for a SwiftUI codebase this silently erases the entire view layer.

The graph still looks healthy afterwards (types, funcs, and imports are all present), which is what makes this hard to notice: `query` and `explain` return confident, well-formed results that simply omit the UI, with no signal that anything is missing.

## Reproduction

```swift
struct PlayerScrubber: View {
    var body: some View {
        VStack { Slider(value: .constant(0), in: 0...1) }
        .contextMenu {
            Button { mode.toggle() } label: { Label("Scrub Chapter", systemImage: "book") }
        }
    }
    var toggled: Int { 1 }
    func realFunction() -> Int { 2 }
}
```

tree-sitter-swift parse (abridged):

```
class_declaration              L1  struct PlayerScrubber: View {
  class_body                   L1  {
    property_declaration       L2  var body: some View {
      computed_property        L2  {          <-- not a recognised body type
    property_declaration       L9  var toggled: Int { 1 }
      computed_property        L9  { 1 }      <-- same
    function_declaration       L10 func realFunction() -> Int { 2 }
      function_body            L10 { 2 }      <-- recognised
```

Resulting graph: nodes for `PlayerScrubber` and `.realFunction()`. Nothing for `body` or `toggled`.

Observed on a real ~6.8k-node / 17k-edge SwiftUI project: a 108-line view file yields exactly two nodes — the file and the `struct`. Every `Slider`, `.contextMenu`, and `Button` inside `body` is invisible. In that same file a `static func` on a sibling enum extracts correctly, so the omission is specifically computed properties, not the file or the type.

Practical symptom that led me here: `explain "<Type>"` works well, but no query can answer "where is this control wired into the UI?" — because the answer lives inside `body` and `body` does not exist in the graph.

## Why the obvious fix is wrong

Adding `property_declaration` to `function_types` would also match **stored** properties (`var x = 5`), which are not functions and have no body. On a medium codebase that mints thousands of junk nodes and leaves the graph worse than before.

## Suggested fix

Emit a function-like node only when a `property_declaration` has a `computed_property` child, and treat that `computed_property` as the body/scope boundary.

That condition is on tree shape, so it doesn't fit the declarative `LanguageConfig` — but there's existing precedent for exactly this shape in `_swift_extra_walk`, which already special-cases Swift enum cases.

Worth handling in the same pass, since both also parse as `computed_property`:

- `willSet` / `didSet` observers on stored properties
- explicit `get { } set { }` accessor blocks

## Impact

Any SwiftUI codebase, and any Swift codebase that puts logic in computed properties. Swift support is otherwise strong here — extension merging, receiver-typed member-call resolution, and module-aware imports all work well, which is why this one gap is easy to miss.

## Related

Distinct from #2147, but they compound on SwiftUI projects. On the same corpus, `View` (degree 210), `Foundation` (264), `SwiftUI` (111) and `RhapsodeKit` (275) act as universal connectors, so `path "A" "B"` between two symbols in the *same* feature returned a five-hop route through `View`, an unrelated screen type, and `Bool`. #2147 covers why those hubs exist; this issue covers why the real intermediate nodes — the view bodies that would otherwise carry the connection — are missing. Fixing one without the other leaves SwiftUI traversal weak either way.

## Happy to send a PR

If you can confirm the fix belongs in `_swift_extra_walk` rather than somewhere else, I'm glad to implement it. I have a 6.8k-node SwiftUI project currently exhibiting the bug, so I can report before/after node and edge counts on a real corpus rather than only a synthetic fixture — useful given this will shift node counts for every Swift repo and presumably move some snapshot tests.

ozdemirsarman:
Closes #2181.

## Problem
Swift's extractor only treats `function_declaration`, `init_declaration`,
`deinit_declaration` and `subscript_declaration` as callables
(`_SWIFT_CONFIG.function_types`). Computed properties parse as
`property_declaration → computed_property → statements`, which no branch
handled — so they produced **no node** and their bodies were **never walked**.

For SwiftUI this is severe: `var body: some View { … }` is a computed property,
so an entire view file collapses to just the file + struct nodes and every UI
element and call inside `body` disappears. `willSet`/`didSet` observers have the
same shape and lost their body calls too.

### Before (repro)
```swift
struct PlayerScrubber: View {
    var body: some View { VStack { doTap() } }  // dropped
    var toggled: Int { 1 }                       // dropped
    func doTap() {}                              // extracted
}
```
Only `.doTap()` was emitted; `body`, `toggled`, and the `doTap()` call inside
`body` were all missing.

## Fix
In the Swift `property_declaration` branch (which already collects type-annotation
references), after the existing handling: if the property has a `computed_property`
or `willset_didset_block` child, emit a function-like member node (`.name`,
`method` edge to the enclosing type) and defer its body to the call-walk via
`function_bodies` — the same mechanism methods use. This is additive: the
existing `references`/field edges from the property's type are unchanged.

Stored properties have no body child, so they are untouched (no regression).

### After
`.body` and `.toggled` are emitted, and the `doTap()` call inside `body` now
produces a `calls` edge from `.body` → `.doTap()`, so the view hierarchy is
traversed.

## Tests
Adds `tests/test_swift_computed_properties.py`:
- computed property emits a node and its body is walked (call captured);
- stored property is **not** emitted as a member but still yields its type
  `references` edge (regression guard);
- `willSet`/`didSet` observer body is walked.

Full language test suite (`test_languages.py`, 324 tests) and the existing Swift
tests pass with no changes.

### DIFF
--- a/graphify/extractors/engine.py
+++ b/graphify/extractors/engine.py
@@ -3101,6 +3101,22 @@ def _extract_generic(
             prop_name = _swift_property_name(node, source)
             if prop_name and prop_type:
                 type_table[prop_name] = prop_type
+            # #2181: a computed property (`var body: some View { … }`) or an
+            # observed one (`willSet`/`didSet`) carries a body that the branches
+            # above never emitted — so the property node AND every call inside it
+            # were dropped. For SwiftUI this erases the whole view layer, since
+            # `body` is a computed property. Emit a function-like member node and
+            # defer its body to the call-walk via function_bodies (mirroring how
+            # methods register their bodies). Stored properties have no such body
+            # child, so their behaviour is unchanged (no regression).
+            comp_bodies = [c for c in node.children
+                           if c.type in ("computed_property", "willset_didset_block")]
+            if comp_bodies and prop_name:
+                prop_nid = _make_id(parent_class_nid, prop_name)
+                add_node(prop_nid, f".{prop_name}", line)
+                add_edge(parent_class_nid, prop_nid, "method", line)
+                for body_block in comp_bodies:
+                    function_bodies.append((prop_nid, body_block))
             return
 
         if (config.ts_module == "tree_sitter_scala"
--- a/tests/test_swift_computed_properties.py
+++ b/tests/test_swift_computed_properties.py
@@ -1,0 +1,86 @@
+"""Regression tests for #2181.
+
+Swift computed properties (`var body: some View { … }`) and observed
+properties (`willSet` / `didSet`) carry a body. Before the fix the Swift
+extractor only recognised function/init/deinit/subscript as callables, so
+those properties produced no node and their bodies were never walked -- which
+for SwiftUI erased the entire view layer (`body` is a computed property).
+"""
+from __future__ import annotations
+
+import tempfile
+import unittest
+from pathlib import Path
+
+from graphify.extract import extract_swift
+
+
+def _labels(result):
+    return [n["label"] for n in result["nodes"]]
+
+
+def _rel(result, relation):
+    return [e for e in result["edges"] if e["relation"] == relation]
+
+
+class TestSwiftComputedProperties(unittest.TestCase):
+    def _extract(self, src: str) -> dict:
+        with tempfile.TemporaryDirectory() as d:
+            p = Path(d) / "View.swift"
+            p.write_text(src, encoding="utf-8")
+            return extract_swift(p)
+
+    def test_computed_property_emits_node_and_walks_body(self):
+        r = self._extract(
+            "struct PlayerScrubber: View {\n"
+            "    var body: some View {\n"
+            "        VStack { doTap() }\n"
+            "    }\n"
+            "    var toggled: Int { 1 }\n"
+            "    func doTap() {}\n"
+            "}\n"
+        )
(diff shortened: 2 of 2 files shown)


safishamsi:
Thanks @ozdemirsarman. Shipped in v0.9.28 (cherry-picked to v8 to preserve your authorship). Closed-unmerged here, but it's in the release: https://github.com/Graphify-Labs/graphify/releases/tag/v0.9.28

PR #2232, opened 2026-07-27
fix(detect): let .env.example-style templates through Stage 2

### LINKED ISSUE #2184: detect(): Stage 2 .env sensitive-pattern regex has no exemption for template suffixes (.env.example, .env.sample, .env.template, .env.dist)
## Summary
`detect._SENSITIVE_PATTERNS`'s `.env`/`.envrc` regex:

```python
re.compile(r'(^|[\\/])\.(env|envrc)(\.|$)', re.IGNORECASE)
```

matches any `.env`-prefixed filename followed by `.` or end-of-string — so it flags `.env.example`, `.env.sample`, `.env.template`, `.env.dist`, `.envrc.example`, etc. identically to real secret files like `.env`, `.env.local`, `.env.production`. There's no carve-out for the well-known "safe to commit" template-suffix convention.

## Observed (graphifyy 0.9.25)
`graphify-init` on a repo containing a template `.env.example` (documented example vars only, no real secrets) reports it under `skipped_sensitive`, so it's silently absent from the graph.

```python
>>> import re
>>> pat = re.compile(r'(^|[\\/])\.(env|envrc)(\.|$)', re.IGNORECASE)
>>> [(n, bool(pat.search(n))) for n in
...  ['.env', '.env.example', '.env.sample', '.env.template', '.env.local', '.envrc']]
[('.env', True), ('.env.example', True), ('.env.sample', True),
 ('.env.template', True), ('.env.local', True), ('.envrc', True)]
```

## No current escape hatch
- `.graphifyignore` `!` negation only affects `_is_ignored()`; `_is_sensitive()` never consults `ignore_patterns`, so it can't rescue a Stage 2 hit.
- `.graphifyinclude` is dead (#2112) — current code just warns and points at `.graphifyignore` negation, which (per above) doesn't actually work here.
- `--exclude` feeds the same `_is_ignored` path as `.graphifyignore`.
- Only workaround is renaming the file outside the regex's reach, which reshapes the user's repo instead of fixing the tool (same objection as #1225).

## Why this differs from the #1225 "no broad override" stance
PR #1921's discussion intentionally keeps Stage 2 patterns (`.env`, key files, secrets dirs) hard-excluded so a user's *broad allowlist glob* can't accidentally pull in a real secret file. This request is narrower: a *specific, principled suffix carve-out* for a naming convention whose entire purpose is "safe to commit," not a user-controlled allowlist. It doesn't reopen the risk #1921 is guarding against.

## Suggested fix
Exempt filenames ending in a known template suffix ahead of the Stage 2 check:

```python
_ENV_TEMPLATE_SUFFIXES = ('.example', '.sample', '.template', '.dist')
if not (re.search(r'\.(env|envrc)\.', name, re.IGNORECASE)
        and name.lower().endswith(_ENV_TEMPLATE_SUFFIXES)):
    if any(p.search(name) for p in _SENSITIVE_PATTERNS):
        return True
```

Happy to open a PR if this shape looks right to a maintainer.

SyedFahad7:
## Summary
Fixes #2184.

Stage 2's `.env` / `.envrc` regex also matched `.env.example` (and `.sample` / `.template` / `.dist`), so committed dotenv templates got dumped into `skipped_sensitive` and never made it into the graph. Real files like `.env.local` should still be blocked — just not the usual placeholder naming.

Carve-out is suffix + basename only; doesn't reopen a broad Stage 2 allowlist (#1921).

## Test plan
- [ ] `.env.example` / `.env.sample` are indexed (not in `skipped_sensitive`)
- [ ] `.env`, `.env.local`, `.env.production` still skipped
- [ ] `uv run pytest tests/test_detect.py -q -k sensitive`

### DIFF
--- a/graphify/detect.py
+++ b/graphify/detect.py
@@ -129,6 +129,19 @@ _SENSITIVE_PATTERNS = [
     # google/oauth2/service_account.py and prose like aws_credentials_rotation.md.
 ]
 
+# Committed dotenv / envrc templates — placeholders only, not live secrets.
+# Stage 2's `.env.` regex otherwise treats these like `.env.local` (#2184).
+_ENV_TEMPLATE_SUFFIXES = (".example", ".sample", ".template", ".dist")
+
+
+def _is_env_template(name: str) -> bool:
+    """True for `.env.example` / `.envrc.sample` style committed templates (#2184)."""
+    lower = name.lower()
+    if not lower.endswith(_ENV_TEMPLATE_SUFFIXES):
+        return False
+    # Basename must still be an .env* / .envrc* file (not e.g. secrets.example).
+    return bool(re.match(r"\.(env|envrc)\.", lower))
+
 # Generic keyword patterns - these only count when the keyword is LOAD-BEARING
 # in the filename (see _generic_keyword_hit), because a keyword buried mid-phrase
 # in a long descriptive slug names a topic, not a credential store:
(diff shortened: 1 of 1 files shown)


safishamsi:
Thanks @SyedFahad7. Shipped in v0.9.29 (cherry-picked to v8), with the regression test added as a follow-up. Closed-unmerged here, but it's in the release: https://github.com/Graphify-Labs/graphify/releases/tag/v0.9.29

PR #2245, opened 2026-07-27
fix(scala): dispatch self-type annotations to requires edges (#2052)

### LINKED ISSUE #2052: Scala self-type annotations (`self: Logging with Database =>`) are completely invisible to structural extraction — a dispatch-only fix, reusing the existing type-reference walker unchanged
## Summary

A trait's self-type annotation — `self: Logging with Database =>`, the cake-pattern idiom for declaring
"this trait can only be mixed into something that also provides `Logging`/`Database`" — produces a dedicated
`self_type` node that is never handled anywhere in the Scala extractor: not a single `references`, `inherits`,
or `mixes_in` edge is produced for the types it names, in any context. Unlike the `match type` finding in
[graphify#2049](https://github.com/Graphify-Labs/graphify/issues/2049), this needs no change to the
type-reference walker itself — `_scala_collect_type_refs` already handles every shape a self-type's type
position can take.
The fix is a single new dispatch branch.

## The gap

```scala
trait Logging
trait Database

trait UserService:
  self: Logging with Database =>

  def findUser(id: Int): Unit = ()
```

```
trait_definition
  identifier "UserService"
  template_body
    self_type
      identifier "self"
      compound_type
        type_identifier "Logging"
        with
        type_identifier "Database"
      =>
    function_definition
      identifier "findUser"
```

`self_type` sits at the same sibling level inside `template_body` as `val_definition`/`type_definition` would.
Grepping the whole extractor for `self_type` turns up zero matches — this node type is never dispatched on
anywhere. It's also not reachable by the existing `extends_clause`/`class_parameters` scan (the code that
produces `inherits`/`mixes_in` edges), because that scan walks the direct children of the
`trait_definition`/`class_definition` node itself — `self_type` is one level deeper, inside `template_body`,
so it was never in that scan's reach even incidentally:

```python
# existing extends_clause scan — iterates node.children where `node` IS the
# trait_definition/class_definition, never descends into template_body:
if config.ts_module == "tree_sitter_scala":
    extend = node.child_by_field_name("extend")
    if extend is None:
        for c in node.children:
            if c.type == "extends_clause":
                extend = c
                break
    ...
```

The practical effect: `Logging with Database` here is a real, declared structural dependency of
`UserService` — exactly the kind of typeclass/mixin wiring information cats-effect-style code relies on the
graph to surface — and it produces nothing at all.

## Isolating the type node — no field names, but `.is_named` filtering is exact

`self_type`'s children carry no field names (`field_name_for_child` returns `None` for every child, checked
directly), so the type has to be found positionally. Filtering by `.is_named` isolates exactly two entries in
the common case — the binder identifier first, the type node second:

```scala
trait Repository[A]:
  this: HasConfig =>

  def load(id: Int): A
```

```
self_type
  identifier "this"
  type_identifier "HasConfig"
```

Named children (punctuation `:`/`=>` are unnamed, confirmed via `.is_named`): `['identifier', 'type_identifier']`
for this shape, `['identifier', 'compound_type']` for the `Logging with Database` shape above — the type node
is reliably `named[1]` whenever a type is present.

**Legal self-type shape with no type at all**, checked so this isn't left as an assumption — Scala allows
naming `this` for use in nested classes without imposing any type constraint:

```scala
class Outer:
  self =>

  def name: String = "outer"
```

```
self_type
  identifier "self"
```

Only one named child here (no `:`, no type) — the "type node is `named[1]`" rule needs `len(named) >= 2` as a
guard, which correctly yields "no type node" for this shape rather than misreading the binder identifier as
a type.

## The walker already handles every real type shape — confirmed by direct invocation, no changes needed

Running the installed `_scala_collect_type_refs` (imported from the installed package, not reimplemented)
directly against each self-type's isolated type node:

```
Logging with Database (compound_type)  -> refs = [('Logging', 'type'), ('Database', 'type')]
HasConfig (type_identifier)            -> refs = [('HasConfig', 'type')]
```

Both resolve correctly with zero modification to the walker — `compound_type` is already in its recursion
whitelist (added independently of this proposal), and a bare `type_identifier` is the walker's base case.
Unlike the `match_type` situation in
[graphify#2049](https://github.com/Graphify-Labs/graphify/issues/2049), there is no second, independent bug
layer here — this is purely a missing dispatch branch.

**Structural refinement in a self-type** (`self: Logging { def extra: Int } =>`, another idiomatic
cake-pattern shape) checked too, not left as an untested assumption:

```scala
self: Logging { def extra: Int } =>
```

```
compound_type
  type_identifier "Logging"
  refinement
    function_declaration "extra"
```

Still a `compound_type` at the top level, so the same isolation logic applies unchanged —
`_scala_collect_type_refs` on it returns `[('Logging', 'type')]`. The refinement body's own member signature
(`extra: Int`) isn't picked up as a reference — checked this is consistent with the rest of the extractor, not
just assumed: a `val x: Logging { def extra: Int } = ...` produces the exact same `refs = [('Logging', 'type')]`
through the existing, unrelated `val_definition` branch. The refinement body is unscanned everywhere in this
extractor, not a limitation specific to self-type support.

## Proposed diff

```python
if (config.ts_module == "tree_sitter_scala"
        and t == "self_type"
        and parent_class_nid):
    named = [c for c in node.children if c.is_named]
    type_node = named[1] if len(named) >= 2 else None
    if type_node is not None:
        line = node.start_point[0] + 1
        refs: list[tuple[str, str]] = []
        _scala_collect_type_refs(type_node, source, False, refs)
        for ref_name, role in refs:
            target_nid = ensure_named_node(ref_name, line)
            if target_nid != parent_class_nid:
                add_edge(parent_class_nid, target_nid, "requires", line)
```

**Verified against a real parsed tree (not just reasoned through)** — running this exact logic against all
four shapes above:

```
UserService  -> requires Logging, requires Database
Repository   -> requires HasConfig
Outer        -> (no self_type type node — self =>, correctly produces nothing)
UserService2 -> requires Logging   (the structural-refinement shape)
```

Also checked that this doesn't interact with the existing `extends_clause` handling when both are present on
the same trait — `trait UserService extends Base: self: Logging => ...` resolves `extends_clause` to `Base`
(via the existing, unmodified `inherits` edge) and `self_type` to `Logging` (via this proposal's new
`requires` edge) independently, with no conflict between the two code paths.

## Open question: relation label

Proposing `"requires"` as a new dedicated label rather than reusing generic `"references"` — a self-type is a
structural precondition ("this can only be mixed into something providing X"), a different relationship than
`inherits`/`mixes_in` (is-a) or the generic `"references"` used for val/type-alias RHS mentions. Precedent in
this extractor is split both ways (`inherits`/`mixes_in`/`derives` are dedicated labels; most type mentions
reuse `"references"` with a `context` tag) — same open question already left to the maintainer for `export`
in [graphify#2048](https://github.com/Graphify-Labs/graphify/issues/2048); the maintainer's call here too.

## What this deliberately does not propose

- Any node/container for the self-type annotation itself — there is no meaningful "self-type node" to create;
  this is reference-edge extraction only, same scoping as the `val`/`type_definition` proposals.
- Scanning a self-type's structural refinement body (`{ def extra: Int }`) for its own member references —
  checked above that the base type still resolves correctly through it, but the refinement body's members
  aren't walked, consistent with every other refinement-type context in this extractor, not a new limitation.

## Environment

- graphifyy 0.9.20, tree_sitter_scala 0.26.0
- All four shapes (multi-type via `with`, single-type, no-type, structural refinement) were dumped directly
  against the installed grammar, checked for absence of `ERROR`/`is_missing` nodes. Field-name absence was
  confirmed via `field_name_for_child`, not assumed. The walker output was produced by running the installed
  `_scala_collect_type_refs` directly against real parsed nodes, not inferred from reading the source.

Yyunozor:
Fixes #2052

## Bug

`self_type` (`self: Logging with Database =>`, `this: T =>`) was never dispatched on anywhere in the Scala extractor, so a trait/class's structural precondition on its enclosing type produced zero edges, in any context.

## Fix

One new dispatch branch in `graphify/extractors/engine.py`, next to `val_definition`/`var_definition`. `self_type`'s children carry no field names, so the type node is found positionally: binder identifier first, type (when present) second. `self =>` binds a name with no type, so `len(named) < 2` correctly yields no type node instead of misreading the binder. `_scala_collect_type_refs` already handles every shape the type position can take (`type_identifier`, `compound_type` for `with`, refinement bodies) — reused unchanged, matching the issue's proposed diff. Edges use a new `requires` relation, following the issue's own proposal; the label stays your call per its open question.

Also added `requires` to `DEFAULT_AFFECTED_RELATIONS` (`graphify/affected.py`), mirroring how `indirect_call` was wired in when introduced, so `graphify affected` follows it like `inherits`/`mixes_in`/`embeds`. Verified end-to-end with `graphify update .` + `graphify affected` on a toy project, not just direct extractor calls.

## Known limitation

Self-types on `trait` bodies — this issue's own cake-pattern example — don't yet produce edges. Tracing the dispatch directly showed `trait_definition` isn't in `_SCALA_CONFIG.class_types` (only `class_definition`/`object_definition` are), so traits aren't class-like containers at all yet: `parent_class_nid` stays `None` inside a trait body today (`extends_clause` heritage is equally absent there, same cause). This dispatch is container-agnostic, so it will automatically cover trait self-types once trait recognition lands — already proposed separately in #1792 — no further change needed here. `class`/`object` self-types already work today (tested below).

## Tests

New `tests/test_scala_self_type.py` (8 tests, own fixture — avoids colliding with other open Scala PRs' shared-fixture edits): single type, `with`-compound, structural refinement (base type only, matching how refinement bodies are unscanned elsewhere), binder-only `self =>` (no edge), a plain class without a self-type (no spurious edge), coexistence with an unrelated `extends`, edge shape (no `context`), and an `affected`-traversal consumer test.

Full suite identical on branch and bare `v8`: same 6 pre-existing environment-dependent failures (missing `openai` extra, one wheel-build test) both sides, 8 net new passes. Ruff clean.

### DIFF
--- a/graphify/affected.py
+++ b/graphify/affected.py
@@ -22,6 +22,7 @@ DEFAULT_AFFECTED_RELATIONS = (
     "uses",
     "mixes_in",
     "embeds",
+    "requires",
 )
 
 
--- a/graphify/extractors/engine.py
+++ b/graphify/extractors/engine.py
@@ -3135,6 +3135,31 @@ def _extract_generic(
                                  line, context=ctx)
             # fall through so any call expressions in the initializer get walked
 
+        # Scala: `self: Logging with Database =>` (or `this: T =>`) declares a
+        # structural precondition on the enclosing type, not a mixin/reference.
+        # self_type carries no field names, so the type node is found
+        # positionally: the binder identifier is named[0], the type (when
+        # present) is named[1]. `self =>` binds a name with no type at all, so
+        # len(named) < 2 correctly yields no type node rather than misreading
+        # the binder as a type. _scala_collect_type_refs already handles every
+        # shape a self-type's type position can take (type_identifier,
+        # compound_type for `with`, refinement bodies via compound_type) --
+        # reused unchanged.
+        if (config.ts_module == "tree_sitter_scala"
+                and t == "self_type"
+                and parent_class_nid):
+            named = [c for c in node.children if c.is_named]
+            type_node = named[1] if len(named) >= 2 else None
+            if type_node is not None:
+                line = node.start_point[0] + 1
+                refs: list[tuple[str, str]] = []
+                _scala_collect_type_refs(type_node, source, False, refs)
+                for ref_name, role in refs:
+                    target_nid = ensure_named_node(ref_name, line)
+                    if target_nid != parent_class_nid:
+                        add_edge(parent_class_nid, target_nid, "requires", line)
+            return
+
         if (config.ts_module == "tree_sitter_cpp"
                 and t == "field_declaration"
                 and parent_class_nid):
--- a/tests/test_scala_self_type.py
+++ b/tests/test_scala_self_type.py
@@ -1,0 +1,141 @@
+"""Scala self-type annotations (`self: Logging with Database =>`, `this: T =>`).
+
+A trait/class's self-type declares a structural precondition on the enclosing
+type ("this can only be mixed into something also providing X") -- the
+cake-pattern idiom. `self_type` was never dispatched on anywhere in the
+extractor, so it produced no edges at all, in any context (#2052). These
+tests pin the new dispatch branch: every real type shape a self-type can
+carry, the negative case where no type is present, that an ordinary class
+without a self-type never gains a spurious edge, that the new `requires`
+relation coexists cleanly with an unrelated `extends` on the same class, and
+that `affected` (blast radius) now traverses it.
+"""
+from pathlib import Path
+
+import networkx as nx
+
+from graphify.affected import affected_nodes
+from graphify.extract import extract_scala
+
+SRC = '''\
+trait Loggable
+trait Database
+trait HasConfig
+trait BaseTrait
+
+class SimpleSelfType {
+  self: HasConfig =>
+
+  def configure(): Unit = ()
+}
+
+class CompoundSelfType {
+  self: Loggable with Database =>
+
+  def run(): Unit = ()
+}
+
+class RefinedSelfType {
+  self: Loggable { def extra: Int } =>
+
+  def refined(): Unit = ()
+}
+
+class BinderOnlySelfType {
+  self =>
+
+  def noop(): Unit = ()
+}
+
+class NoSelfType {
+  def plain(): Unit = ()
+}
+
+class SelfTypeWithExtends extends BaseTrait {
+  self: Loggable =>
+
+  def combined(): Unit = ()
+}
+'''
+
+
+def _build(tmp_path):
+    (tmp_path / "SelfTypes.scala").write_text(SRC)
+    r = extract_scala(tmp_path / "SelfTypes.scala")
+    nid = {n["label"]: n["id"] for n in r["nodes"]}
+    return r, nid
+
+
+def _rels(r, relation):
+    return {(e["source"], e["target"]) for e in r["edges"] if e["relation"] == relation}
+
+
+def test_self_type_simple_type_emits_requires(tmp_path):
+    r, nid = _build(tmp_path)
+    requires = _rels(r, "requires")
+    assert (nid["SimpleSelfType"], nid["HasConfig"]) in requires
+
+
+def test_self_type_compound_with_emits_requires_for_each_member(tmp_path):
+    r, nid = _build(tmp_path)
+    requires = _rels(r, "requires")
+    assert (nid["CompoundSelfType"], nid["Loggable"]) in requires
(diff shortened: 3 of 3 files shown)


safishamsi:
Thanks @Yyunozor. Shipped in v0.9.29 (cherry-picked to v8). Note: the requires-edge emission fires for self-types in a class/object; the trait cake-pattern case (#2052) additionally needs trait-as-class recognition, tracked in #2247. Closed-unmerged here, but it's in the release: https://github.com/Graphify-Labs/graphify/releases/tag/v0.9.29

Issue #2206, opened 2026-07-25
Extraction: rationale node labels are a raw 80-character slice of the docstring, producing truncated mid-word titles and filenames ending in `..md`


irvinolave-alt:
Found on **graphifyy 0.9.26** (uv tool install, Python 3.11, Linux), running `graphify <dir> --backend claude-cli`. Unlike #2204 and #2205 this one is not export-specific — it is in the extractor.

### Summary

`_extract_python_rationale` / `_extract_js_rationale` build the node label by slicing the raw docstring to 80 characters **before** normalizing whitespace. Three consequences:

1. the slice lands mid-word (`... solo dev`, `... GET articles/consulta).      Si`);
2. the docstring's newline + indentation survives as a run of literal spaces inside the label, because the slice happens before the `\n` → ` ` replacement collapses anything;
3. a docstring ending in `.` at the slice boundary yields an Obsidian filename ending in `..md`.

These labels are not export-only — they are the `label` in `graph.json`, so they also show up in `GRAPH_REPORT.md`, in `graphify query` output, and in `explain`.

### Where

`graphify/extract.py:1075` (Python) and `graphify/extract.py:1212` (JS/TS) — identical line:

```python
label = text[:80].replace("\r\n", " ").replace("\r", " ").replace("\n", " ").strip()
```

The slice is applied to `text`, before the replacements. Reversing the order would fix (2) but not (1) or (3).

### Reproduction

```python
def list_products(...):
    """Lista/busca productos del catálogo de The Best (GET /articles/consulta).
       Si no se pasa consulta, devuelve el catálogo completo.
    """
```

produces, in `graph.json`:

```json
{"label": "Listabusca productos del catálogo de The Best (GET articlesconsulta).      Si",
 "file_type": "rationale", "source_file": "mcp-thebest/server.py",
 "source_location": "L187", "_origin": "ast"}
```

and, in the Obsidian vault, the file:

```
Listabusca productos del catálogo de The Best (GET articlesconsulta).      Si.md
```

Twelve of my fourteen rationale nodes were affected. A second real example, where the truncation cuts mid-sentence and the trailing period makes a double-dot filename:

```
000005 - TIENDA PIURA' - '000005'..md
```

### Expected

A rationale label that reads as a title: whitespace collapsed, cut on a word boundary, no trailing punctuation artifact.

### Suggested fix

Normalize first, then slice on a word boundary:

```python
[...]

Issue #2205, opened 2026-07-25
Obsidian export: nodes whose label starts with `.` produce notes Obsidian never displays


irvinolave-alt:
Found on **graphifyy 0.9.26** (uv tool install, Python 3.11, Linux), running `graphify <dir> --backend claude-cli` followed by `graphify export obsidian --graph graphify-out/graph.json --labels graphify-out/.graphify_labels.json --dir <vault>`. Sibling of #2204, independent fix.

### Summary

`safe_name()` strips filesystem-unsafe characters but leaves a leading `.` intact, so a node for a dotfile is written as `.env.md`. Obsidian hides every file whose name begins with a dot: the note does not appear in the file explorer, in Quick Switcher, or in graph view. It is written, counted in the "N notes" total, and referenced by `[[wikilinks]]` from visible notes — those links render as unresolved.

### Where

`graphify/export.py:500` (and the parallel copy at `:801`)

```python
def safe_name(label: str) -> str:
    cleaned = re.sub(r'[\\/*?:"<>|#^[\]]', "", label...)
    ...
```

The character class does not cover a leading `.`.

### Reproduction

Corpus containing any dotfile that becomes a node — `.env`, `.env.example`, `.gitignore`, `.dockerignore`, `.eslintrc`, `.github/workflows/*.yml`. My run produced:

```
.env (credenciales TB).md
.env.example.md
```

Both invisible in Obsidian. `.env.example.md` is the target of a `references` edge from `README.md`, so that link shows as unresolved in the visible note.

### Expected

Every note the exporter writes is reachable in the Obsidian UI.

### Suggested fix

Strip or escape a leading dot in `safe_name`, e.g. `cleaned = cleaned.lstrip(".")` with a fallback if that empties the stem, or a `dotfile-` prefix to keep the name recognizable. Either way the H1 and `source_file:` frontmatter still carry the true filename, so nothing is lost. Whatever the choice, `_dedup_node_filenames` will handle a collision with a non-dotted sibling.

PR #2246, opened 2026-07-27
fix(extract): normalize whitespace before truncating rationale labels (#2206)

Yyunozor:
## Summary

`_extract_python_rationale` / `_extract_js_rationale` build rationale node labels by slicing the raw docstring/comment text to 80 characters *before* normalizing whitespace. This can cut a word in half, leave literal spaces where a newline + indentation used to be, and — when the cut lands on a `.` — produce an Obsidian filename ending in `..md`. Not export-only: it's the `label` in `graph.json`, so the artifacts show up in `GRAPH_REPORT.md`, `graphify query`, and `explain` too.

## Reproduction

```python
def reconcile_batches():
    """This routine reconciles pending settlement batches nightly because upstream feeds arrive unordered and out of sequence."""
```

Before: label ends `"...because upstream feed"` (80 chars, "feeds" sliced mid-word). A docstring whose 80th char lands on a sentence-ending `.` exports as `...and caches..md`.

## Root cause

Both `_add_rationale` closures in `graphify/extract.py` do:
```python
label = text[:80].replace("\r\n", " ").replace("\r", " ").replace("\n", " ").strip()
```
The slice runs before whitespace normalization, so it can land inside a word or a newline's collapsed space run, and `.strip()` only trims surrounding whitespace, never a trailing period.

## Fix

Both sites now call a shared `_shorten_rationale_label`:
```python
label = textwrap.shorten(text, width=width, placeholder="…")
if label in ("", "…"):
    flat = " ".join(text.split())
    label = flat if len(flat) <= width else flat[: width - 1] + "…"
```
`textwrap.shorten` collapses whitespace, then drops whole words from the end (never mid-word), adding the placeholder only when it truncates — a short label is unchanged. The fallback guards a `shorten` quirk: a first word alone >= 80 chars (e.g. one long URL opening a comment) collapses to a bare `"…"`, worse than the parent's raw slice. The fallback restores content.

## Known limitation

`shorten` leaves an untruncated label's own trailing punctuation alone, so a short rationale legitimately ending in `.` still hits a separate, pre-existing gap: `export.py`'s two `safe_name()` (~L500, ~L801) strip `.md/.mdx/.qmd/.markdown` but not a bare `.`. Different file, out of scope here — a follow-up would strip a trailing `.` in `safe_name`, as it already strips a leading one (#2205).

## Validation

Nine new tests in `tests/test_rationale.py`, split across Python and JS/TS: word-boundary truncation, whitespace-run collapse, no-trailing-period, long-token fallback, short-docstring non-regression. All fail on parent with the reported symptom, pass after the fix.

```
python -m pytest tests/ -q
```
Delta is exactly these 9 tests; skip set identical both sides. Cross-checked on Python 3.10. `ruff check` and all five `tools.skillgen` guards clean; `graphify update` smoke run OK.


safishamsi:
Thanks @Yyunozor. Shipped in v0.9.29 (cherry-picked to v8). Closed-unmerged here, but it's in the release: https://github.com/Graphify-Labs/graphify/releases/tag/v0.9.29

PR #2247, opened 2026-07-27
fix(scala): recognize trait as a class-like node

Yyunozor:
## Bug

Scala `trait` declarations are invisible to structural extraction: no node
for the trait, and its `extends`/`with` heritage is never evaluated.

Root cause: `_SCALA_CONFIG.class_types` (`graphify/extract.py`) only listed
`class_definition`/`object_definition`. tree-sitter-scala parses `trait Foo`
as its own kind, `trait_definition`, so it never entered `walk()`'s
class-node branch: no node was created, and members fell to the generic
recurse fallback, which resets `parent_class_nid` to `None` (misattributed
to file scope). `trait_definition` reuses the `template_body` shape already
in `body_fallback_child_types` -- confirmed against the grammar's own parse
tree -- so the existing heritage/field-reference code picks it up unchanged
once dispatched, no other config or extractor change needed.

## Changed since the last review

Thanks for the detailed review. Taking the split option: this version drops
the id-collision guard entirely rather than repairing it. `graphify/extractors/engine.py`
is now byte-identical to `v8`; the diff is the one-line `class_types`
addition plus tests.

The cross-file regression you traced (`_rewire_unique_stub_nodes` bailing
once a salted id makes a label non-unique, so a class+companion reference
from another file lands on a dangling stub) no longer applies -- there's no
salting left to trigger it. `test_cross_file_reference_to_a_companion_pair_still_resolves_to_the_real_node`
pins this directly: a `case class Point`/`object Point` pair referenced from
a second file resolves to the real node, byte-for-byte the same as on `v8`
before this diff.

## Completes the self-type case from #2052

Following your note when closing #2245: the requires-edge emission is gated
on `parent_class_nid`, so a self-type on a trait never reaches it. Verified
against current `v8` (0.9.29): `trait Cake { self: Loggable => }` emits no
edge before this diff, and `requires Cake -> Loggable` after it. Since the
cake pattern puts self-types on traits, this is the half of #2052 that
dispatch alone could not reach.

## Known limitation

Recognizing traits as class-like means a same-named trait and companion
object (`trait Foo` + `object Foo`) now collide onto one node the same way
a same-named class and companion object already do on `v8` today (single
node, source-order-first definition keeps the id, both definitions' members
attach to it) -- confirmed by direct comparison, not assumed: this is the
same pre-existing shape extended to a third node kind, not a new failure
mode. Left untouched here; a project-wide kind-aware id scheme would be a
natural follow-up, out of scope for this diff.

## Testing

`tests/test_scala_trait_classlike.py`: node creation, heritage onto real
(non-stub) nodes, member ownership, a distinct-name trait/object
coexistence check, an `affected` traversal through a trait, and the
cross-file companion non-regression pin above. Rebased onto current `v8`
(0.9.29): full suite 3847 -> 3854 passed, 3 skipped, 0 failed on Python
3.10 and 3.12 -- delta exactly these tests, no pre-existing failures on
either side. Ruff clean.

### DIFF
--- a/graphify/extract.py
+++ b/graphify/extract.py
@@ -905,7 +905,16 @@ _KOTLIN_CONFIG = LanguageConfig(
 
 _SCALA_CONFIG = LanguageConfig(
     ts_module="tree_sitter_scala",
-    class_types=frozenset({"class_definition", "object_definition"}),
+    # trait_definition is its own tree-sitter-scala node kind, distinct from
+    # class_definition/object_definition, so a `trait` never entered the class
+    # branch below: no node for the trait itself, and every member under it
+    # walked with parent_class_nid reset to None by the generic recurse
+    # fallback (fields/methods misattributed to file scope, extends_clause
+    # heritage never evaluated). Traits use the same template_body shape
+    # already declared in body_fallback_child_types, so no other config
+    # change is needed for the existing class-node/heritage/field-reference
+    # handling to pick them up.
+    class_types=frozenset({"class_definition", "object_definition", "trait_definition"}),
     function_types=frozenset({"function_definition"}),
     import_types=frozenset({"import_declaration"}),
     call_types=frozenset({"call_expression"}),
--- a/tests/test_scala_trait_classlike.py
+++ b/tests/test_scala_trait_classlike.py
@@ -1,0 +1,199 @@
+"""Regression tests: Scala `trait` recognized as a class-like container.
+
+`trait_definition` is tree-sitter-scala's own node kind for `trait Foo { ... }`,
+distinct from `class_definition`/`object_definition`. It was absent from
+`_SCALA_CONFIG.class_types`, so `walk()`'s class-node branch never matched a
+trait: no node represented the trait itself, its `extends_clause` heritage was
+never evaluated (no `inherits`/`mixes_in` edges), and every member underneath
+it fell to the generic "default: recurse" branch, which resets
+`parent_class_nid` to `None` -- so a trait's own `val`/`def` surfaced (if at
+all) as if it belonged to the file, not the trait. A trait uses the same
+`template_body` shape already declared in `body_fallback_child_types`, so
+recognizing it needs no new handler: the existing heritage/field-reference
+code that already runs for `class_definition`/`object_definition` picks it up
+unchanged.
+
+Known limitation, not covered here: a same-named trait and companion object
+(`trait Foo` + `object Foo`) now collide onto one node the same way a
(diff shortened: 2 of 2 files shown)


safishamsi (changes requested):
Thanks @Yyunozor. The trait-as-class-like recognition is exactly right (one-line config addition; traits reuse template_body, existing class/object handling untouched), and the tests are thorough.

The blocker is the same-name kind-collision guard. Salting the second same-named definition's id (`class_nid = _make_id(class_nid, t)`) is fine within a file, but it regresses cross-file resolution on Scala's most common idiom (a class/trait plus its companion object). `_rewire_unique_stub_nodes` only collapses a sourceless stub onto a real definition when the label is unique; with the salted id there are now two real candidates for the label, so the rewire bails and every cross-file reference to a class+companion (or trait+companion) pair lands on a dangling sourceless stub instead of the real node. Reproduced with a two-file corpus (`case class Point` + `object Point`, referenced from another file): on this PR the reference target is a `('Point','')` stub, where v8 resolves to the real node. For a tool whose core value is `affected`, that's a functional regression, and the PR's tests only pin the same-file case.

Requested change: keep the trait recognition; fix the rewire side so a kind-split companion pair still resolves — when all same-label real candidates come from one source file and one holds the plain (unsalted) id, rewire the stub to that plain-id holder instead of bailing on non-uniqueness. Alternatively, split the trait recognition (which I'd merge as-is) from the collision guard into separate PRs. Happy to re-review once the cross-file rewire is covered (a two-file class+companion reference test would pin it).

Yyunozor:
Thanks for the detailed review — took the split option.

The collision guard is gone entirely: `graphify/extractors/engine.py` is byte-identical to `v8`, and the diff is now the one-line `class_types` addition plus tests.

Added the two-file `class`+companion reference test you asked for — it fails on the previous commit (reference lands on the `('Point','')` stub) and passes here. Rebased onto 0.9.29: full suite 3847 → 3854 passed, 0 failed, on Python 3.10 and 3.12.

Yyunozor:
Closing this one — the trait recognition landed via #1792 in 0.9.53, so
there's nothing left to review here.

I rebased this branch on current `v8` to check: with that line already in,
the only remaining diff is the test file, including a two-file
class+companion cross-file case pinning the resolution you flagged in
review. That specific case (a reference from a third file to a
`class`/`object` companion pair) isn't in the current suite. Happy to open
it as a small standalone tests-only PR if that's useful — no pressure
either way.

Thanks for catching the collision-guard regression back in July.

Issue #2241, opened 2026-07-27
False indirect_call edges (INFERRED 0.8): call arguments bind as callees to an unrelated single-letter test helper across files


regiusti:
## Summary

When a function makes a call **through a function-valued variable** (`campoFn(r)`, `c.get(r)`, `pred(r)`) and an **argument** of that call is an identifier that shares its name with an unrelated symbol elsewhere in the repo, graphify emits an `indirect_call` edge from the calling function **to that unrelated symbol**, with confidence 0.8.

Single-letter identifiers make this explode: a **file-local one-letter test helper** (`const r = ...` inside `*.test.ts`) became the target of fabricated `indirect_call` edges from every function in the repo that passes a variable named `r` to a callback.

## Version / environment

- graphifyy **0.9.28** (PyPI), installed via `uv tool install graphifyy==0.9.28`
- Windows 11, Python 3.13
- `graphify . --code-only` (pure AST path, no LLM)

## Minimal repro (2 files)

`src/round.test.ts`:

```ts
import { describe, it, expect } from 'vitest';

// Local one-letter rounding helper, private to this test file.
const r = (v: number) => Math.round(v * 100) / 100;

describe('rounding', () => {
  it('rounds to 2 decimals', () => {
    expect(r(1.005)).toBe(1.01);
  });
});
```

`src/report.ts`:

```ts
type Row = { t: string; v: number };
type ColDef = { label: string; get: (row: Row) => number };

// None of these functions reference the test helper `r` from round.test.ts.
// The only `r` below is a loop variable / arrow parameter used as an ARGUMENT
// to calls made through function-valued variables (dispatch).

export function serieTemporal(
  rows: Row[],
  bucketFn: (t: string) => string,
  campoFn: (row: Row) => number,
): Map<string, number[]> {
  const buckets = new Map<string, number[]>();
  for (const r of rows) {
    const k = bucketFn(r.t);
    const arr = buckets.get(k) ?? [];
    arr.push(campoFn(r));
    buckets.set(k, arr);
  }
  return buckets;
}

export function buildSheet(rows: Row[], cols: ColDef[]): Record<string, number>[] {
  return rows.map((r) => {
    const o: Record<string, number> = {};
    for (const c of cols) o[c.label] = c.get(r);
    return o;
  });
}

export function filterRows(rows: Row[], pred: (row: Row) => boolean): Row[] {
  const out: Row[] = [];
  for (const r of rows) if (pred(r)) out.push(r);
  return out;
[...]

Issue #1630, opened 2026-07-03
TypeScript/JS: method calls on a typed receiver that is a local `new` binding or a closure-captured parameter are not emitted as `calls` edges (v0.9.5)


DanielC000:
### Summary

On TypeScript/JavaScript, a member call `x.method()` does not produce a `calls` edge to the receiver type's method when `x`'s type comes from **(A)** a local `const x = new Foo()` binding, or **(B)** a **closure over a typed parameter** (`function f(x: Foo) { return () => x.method() }`) — even though the receiver's type is statically known. No caller edge is built in either case, so `graphify affected <method>` reports "No affected nodes found" for a method that is, in fact, called, and it silently under-reports for refactor-impact ("what breaks if I change this method?").

This is the TypeScript/JS analog of **#1609** (C#: typed `field/param/local` receiver calls not emitted). The TS `this.field.method()` variant was already fixed in **#1316** (closed) / v0.9.3's constructor-injected-dependency resolution — but that fix does not extend to a local `new` binding or a closure-captured parameter, which this issue covers. (The existing TS/JS `_resolve_*_member_calls` path handles `this.`/unqualified/static-import calls, same gap #1609 describes for C#.)

### Related issues
- **#1609** (open) — same pattern class for C# (`field/param/local` typed receiver). This is its TS/JS sibling.
- **#1316** (closed/completed) — TS `this.db.query()` injected-field member calls; fixed the `this.field` case only.

### Minimal worked example (3 files)

`src/svc.ts`
```ts
export class Svc {
  doThing(): number { return 1; }
}
```

`src/direct.ts` — Pattern A: local variable from `new`
```ts
import { Svc } from "./svc";
const s = new Svc();
export function usesDirect(): number {
  return s.doThing();   // expected: usesDirect --calls--> Svc.doThing
}
```

`src/closure.ts` — Pattern B: closure over a typed parameter
```ts
import { Svc } from "./svc";
export function register(svc: Svc): () => number {
  return () => svc.doThing();   // expected: register (or its closure) --calls--> Svc.doThing
}
```

### Actual (graphify 0.9.5)

```
$ graphify affected "doThing" ...
Affected nodes for .doThing()
Depth: 2
No affected nodes found.

$ graphify explain "doThing" ...
Node: .doThing()   Degree: 1
Connections (1):
  <-- Svc [method] [EXTRACTED]
```

Only the container `Svc --method--> doThing` edge exists; neither `usesDirect` nor `register` is linked.

[...]

PR #2248, opened 2026-07-27
fix(extract): shadow untracked JS/TS closure params from indirect_call args (#2241)

Yyunozor:
## Summary

Fixes one root cause of #2241: when a function calls through an inline/untracked JS or TS closure (`rows.map((r) => c.get(r))`), and a call argument inside it shares a name with an unrelated symbol elsewhere in the repo, graphify emits a fabricated `indirect_call` edge (confidence 0.8) to that symbol. Single-letter identifiers make this common.

## Reproduction

```ts
// round.test.ts
const r = (v) => Math.round(v);
// report.ts
export function buildSheet(rows, cols) {
  return rows.map((r) => cols.map((c) => c.get(r)));
}
```
`buildSheet` gets a fabricated `indirect_call` edge to `round.test.ts`'s `r`, though `r` is the arrow's own parameter.

## Root cause

`walk_calls` flattens an inline/untracked closure (not separately tracked in `function_bodies`) onto the enclosing named function's `caller_nid`, so calls inside it resolve as if made by that function (#1630). The closure's own parameters/locals were never folded into the shadow set guarding `indirect_call` argument resolution — a name locally bound one scope down read as an unresolved reference instead of a shadow.

## Fix

Thread an `extra_locals` set through `walk_calls`'s recursion. Entering an untracked closure folds its own bindings (via the existing `_js_local_bound_names`) into `extra_locals` for its subtree only; deeper untracked closures compound the same way. All shadow-set reads in `walk_calls` union in `extra_locals`, applying uniformly to the argument, collection, and assignment/return capture paths sharing this guard, not just the one that surfaced it. Tracked closures (const-assigned arrows, methods) are unaffected — they already get their own `caller_nid` and correctly-scoped shadow set.

## Known limitation

A `for (const x of xs)` loop variable not wrapped in a `variable_declarator` is a separate, pre-existing gap in the same shadow computation — two of #2241's three reported functions (`serieTemporal`, `filterRows`) hit that gap, not this one. #1985 already fixes it (its diff removes those two edges but leaves `buildSheet`'s untouched, and vice versa here). Not duplicated, to avoid overlap.

## Validation

New `tests/test_indirect_call_nested_closure_shadow.py` (6 tests): the reported shape, a genuine unshadowed reference in the same closure shape, two-level nesting, a tracked const-arrow regression guard, `affected` exclusion, warm-cache stability. All 6 fail on parent, pass here.

Full suite, Python 3.10 and 3.12: 13 pre-existing failures identical both sides (unrelated env gaps — missing `tree-sitter-hcl`, ollama/apm tests), delta exactly +6. `ruff check` clean. Adversarial pass: tracked-outer/untracked-inner nesting, destructured closure params, sibling closures (one shadowed, one genuine) — all correct. Live `graphify update .` + `graphify affected` smoke confirms the fabricated edge and its blast-radius entry are gone.

### DIFF
--- a/graphify/extractors/engine.py
+++ b/graphify/extractors/engine.py
@@ -3852,6 +3852,7 @@ def _extract_generic(
         node,
         caller_nid: str,
         java_types: dict[str, str] | None = None,
+        extra_locals: frozenset[str] = frozenset(),
     ) -> None:
         if node.type in config.function_boundary_types:
             # JS/TS: an inline/returned closure not separately tracked in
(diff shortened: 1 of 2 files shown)


safishamsi:
Thanks @Yyunozor. Shipped in v0.9.29 (cherry-picked to v8). Closed-unmerged here, but it's in the release: https://github.com/Graphify-Labs/graphify/releases/tag/v0.9.29

Issue #2243, opened 2026-07-27
0.9.28: machine-slug node-ids still leak into links endpoints (residual of #1899, now in edges not nodes)


vanjos:
Follow-up to #1899 (thanks for fixing node `id` / `source_file` relativization in 0.9.13-0.9.15). On 0.9.28 the same absolute-path-derived slug still leaks, but now into edge endpoints (`links[].source` / `links[].target`) instead of node ids.

## What happens

A `links` entry's `source`/`target` is a machine-specific slug: the absolute scan path, lowercased with every run of non-alphanumerics collapsed to `_`, plus a trailing `_cs`/`_py`. For example `users_alice_code_myrepo_src_pkg_thing_cs` instead of the node id it should reference (`thing`). The node itself has the correct relativized id. Only the edge carries the slug.

## Root cause (best guess)

#1899 relativized the node ids, but the id references already written into `links` endpoints aren't updated in the same pass, so edges keep the pre-relativization slug and point at a string that no longer matches any node's id.

Evidence: most of these slug endpoints do correspond to a real node once you slugify that node's `source_file` the same way and match. The node exists with a clean id, the edge just references its stale slug. A minority resolve to no node at all (references to files outside the scan root) and are genuine orphans.

## Why it matters

The slug is derived from the absolute scan path, so the same source tree built on two different machines (or two different checkout paths) produces a different `graph.json`. The `links` differ even though the node id sets are byte-for-byte identical. This breaks build reproducibility and cross-machine caching. We commit graphs to git, so every rebuild shows a spurious diff, but the underlying non-determinism is general.

## Scale

About 2,442 such endpoints in a real 79k-node / 232k-link C# graph. Across two machines the node id sets were identical and only the link endpoints differed.

## Expected / suggested fix

Apply the #1899 relativization to `links` `source`/`target` as well. Either:

- update edge endpoints when the referenced node's id is relativized, or
- normalize endpoints at serialization time: resolve the path-derived slug back to the canonical node id (match by the node's slugified `source_file`), and drop the edge if it resolves to no node.

## Repro

Build a graph where a file is referenced from outside the scan root (for example a `<ProjectReference Include="..\Core\Core.csproj" />` when only `WebApi/` is scanned), then inspect `links` for `source`/`target` values that begin with a path-derived slug (`<abs-path-slug>_...`) rather than a bare node id.

safishamsi:
[...]

Issue #2231, opened 2026-07-27
indirect_call edges from module-top-level code keep absolute-path-derived source id (id_remap skips raw_calls)


phhktb:
## Summary

Running the full extraction pipeline on a ~420-file JS/TS project, `graphify.diagnostics.diagnose_extraction` reports 586 dangling-endpoint edges. I traced the `indirect_call` subset (58 edges, all with a missing **source**) to a concrete bug in the id-remap pass in `extract.py`.

## Confirmed bug: `id_remap` is applied to `all_nodes`/`all_edges` but never to `all_raw_calls`

- `extractors/engine.py:2342` creates a whole-file "module" node: `file_nid = _make_id(str(path))`, i.e. an id derived directly from the (often absolute) filesystem path.
- That `file_nid` is used as `scope_nid`/`caller_nid` for `indirect_call` edges emitted from **module-top-level code** (calls not inside any function), e.g. via `_emit_indirect_ref(ident, file_nid, js_module_bound, "collection")` around `engine.py:4468-4502`. These go through `raw_calls` (`all_raw_calls`) for cross-file resolution.
- `extract.py:4661-4753` builds an `id_remap` dict that maps exactly this absolute-path-derived id (`_make_id(str(path))` / `_make_id(str(path.resolve()))`) to the canonical relative form `_file_node_id(rel)` — this is the fix for #502.
- `extract.py:4754-4762` applies `id_remap` to `all_nodes` and `all_edges` only:
  ```python
  if id_remap:
      for n in all_nodes:
          if n.get("id") in id_remap:
              n["id"] = id_remap[n["id"]]
      for e in all_edges:
          if e.get("source") in id_remap:
              e["source"] = id_remap[e["source"]]
          if e.get("target") in id_remap:
              e["target"] = id_remap[e["target"]]
  ```
  **`all_raw_calls` is never touched by this loop.**
- The *other* remap (`sym_remap`, built from `prefix_remap`) does patch `raw_calls[caller_nid]` (`extract.py:4820-4823`), but `sym_remap` is keyed off `_file_node_id(path)` (the relative-form prefix), which never matches the absolute-form `file_nid`. So neither remap pass fixes `raw_calls[caller_nid]` for this id shape.
- Net effect: when a `raw_calls` entry with `caller_nid = file_nid` (absolute-derived) is later resolved into a real `indirect_call` edge, its `source` keeps the stale absolute-path id — which matches no node in `all_nodes` (those *were* remapped), so it dangles.

[...]

PR #2250, opened 2026-07-27
fix(extract): canonicalize out-of-root import/include edge targets (#2243)

Yyunozor:
## Summary

Follow-up to #1899, which gave a NODE minted from an absolute out-of-root path a portable `ext_`-namespaced id, matching the node's own id against `_make_id(str(its source_file))`. Some cross-file resolvers (Python relative imports, C/C++/ObjC quoted `#include`) only ever emit an EDGE for an import target, no node — so when that target is outside the scan root, the belt-and-braces pass has nothing to learn the old→new id from, and the edge keeps the raw `_make_id(str(absolute_path))` slug forever. The scan path, including the OS username, ends up in `links[].source`/`target` and differs between machines/checkouts even though the node id sets are identical.

`_import_c` also never stamped the transient `target_file` hint (#1814/#2169) its Python/JS siblings already use for this kind of canonicalization.

## Root cause

`extract.py`'s `id_remap` pass already walks `target_file`-stamped edges to canonicalize in-root-but-unscanned targets, but explicitly skipped the out-of-root branch ("leave its ids alone"), deferring to a later pass that only fires for nodes.

## Fix

- `_import_c` now stamps `target_file` on a resolved `#include`, mirroring `_import_python`/`_import_js`.
- `id_remap`'s out-of-root branch now mints the same portable `ext_` id an out-of-root NODE already gets (reusing `_portable_out_of_root_sf`, moved next to `id_remap` so both share it), so an edge with no node of its own is covered too. A target missing on disk still stays dangling, unchanged.

## Reproduction

```c
// app/main.c, scan root = app/, lib/ outside root
#include "../lib/foo.h"
```
Before: `links[].target` is a raw absolute-path slug, different on every checkout. After: `ext_lib_foo_h`, identical from two independently-built, differently-nested checkouts of the same tree.

## Validation

Four tests in `tests/test_extract.py`: out-of-root `#include` gets the portable id (transient hint never leaks); two differently-nested checkouts of the same corpus give a byte-identical target; an in-root, same-batch include still resolves to the real node's id (regression guard); the equivalent out-of-root Python relative import is fixed too, confirming the gap is in the shared path, not `_import_c`-specific.

```
pytest tests/ -q
```
py3.12: branch 3765/4 skipped, parent 3761/4 → delta +4, exact. py3.10: branch 3764/1 failed/4 skipped, parent 3760/1/4 → delta +4, exact; the 1 failure (`test_manifest_ingest.py::test_apm_parses_name_and_deps`) is identical both sides, pre-existing. `ruff check .` clean both sides; all five `tools.skillgen` guards and a `graphify update` smoke run pass.

## Known limitation

Covers every current `target_file`-stamping resolver (Python, C/C++/ObjC, JS/TS/Svelte/Astro/Vue). A resolver minting a path-derived target without stamping `target_file` — none do today — would still leak; this closes the shared-mechanism gap, not a per-language allowlist.

### DIFF
--- a/graphify/extract.py
+++ b/graphify/extract.py
@@ -550,6 +550,11 @@ def _import_c(node, source: bytes, file_nid: str, stem: str, edges: list, str_pa
                         "source_file": str_path,
                         "source_location": f"L{node.start_point[0] + 1}",
                         "weight": 1.0,
+                        # Stamp the resolved target, mirroring _import_python (#1814):
+                        # without it, an include whose header lives outside this
+                        # batch's paths keeps the raw absolute-path id no later pass
+                        # ever learns to relativize (#2243).
+                        "target_file": str(resolved),
                     })
                     break
             module_name = raw.split("/")[-1].split(".")[0]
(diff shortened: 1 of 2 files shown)


safishamsi:
Thanks @Yyunozor. Shipped in v0.9.29 (cherry-picked to v8) as part of closing the absolute-path node-id leak class (#2231/#2243), with a general backstop added on top. Closed-unmerged here, but it's in the release: https://github.com/Graphify-Labs/graphify/releases/tag/v0.9.29

PR #2271, opened 2026-07-28
fix: preserve edge direction in merge-graphs

### LINKED ISSUE #2261: merge-graphs silently rewires import edges to the importing file's own node
## Summary

`graphify merge-graphs` silently rewires import edges: the target is replaced by the **importing file's own node**. The edge count is preserved, so nothing looks wrong — but the graph now claims a file imports itself instead of the module it actually imports.

Impact: `graph.json` cannot be trusted for "does A depend on B?" after a merge. In a real monorepo merge (2 subprojects, 9423 nodes) this rewired 6 of 23 import edges on a single file I spot-checked, including the one that mattered for the question I was asking. Because the edge count is unchanged and no warning is printed, the loss is invisible.

## Repro

Six files, two projects.

`back/src/empresa.js`
```js
'use strict';
function makeGetEmpresaId(deps) { return async (req) => req.user.empresaId; }
module.exports = { makeGetEmpresaId };
```

`back/src/collections.js`
```js
'use strict';
module.exports = { atendimentos: 'atendimentos' };
```

`back/src/logger.js`
```js
'use strict';
module.exports = { info(msg) { console.log(msg); } };
```

`back/src/rota.js`
```js
'use strict';
const { makeGetEmpresaId } = require('./empresa');
const C = require('./collections');
const logger = require('./logger');

function createRouter(deps) {
  const getEmpresaId = makeGetEmpresaId(deps);
  return async function handler(req, res) {
    const empresaId = await getEmpresaId(req);
    logger.info('consultando ' + C.atendimentos);
    return res.json({ empresaId, col: C.atendimentos });
  };
}
module.exports = { createRouter };
```

`front/src/app.js`
```js
'use strict';
const { helper } = require('./util');
module.exports = { run: () => helper() };
```

`front/src/util.js`
```js
'use strict';
module.exports = { helper: () => 42 };
```

Then:

```bash
graphify extract ./back  --code-only
graphify extract ./front --code-only
graphify merge-graphs back/graphify-out/graph.json front/graphify-out/graph.json --out merged.json
```

## Result

Edges leaving `src/rota.js`, before vs after the merge:

| relation | line | before merge | after merge |
|---|---|---|---|
| `imports_from` | L2 | `empresa.js` | **`rota.js`** |
| `imports` | L2 | `makeGetEmpresaId()` | **`rota.js`** |
| `imports_from` | L3 | `collections.js` | **`rota.js`** |
| `imports_from` | L4 | `logger.js` | **`rota.js`** |
| `calls` | L7 | `makeGetEmpresaId()` | **`createRouter()`** |

The per-repo `graph.json` files are correct — the corruption is introduced by `merge-graphs`. The target nodes still exist in the merged graph (`back::src_empresa`, `back::src_collections`, …); they are simply no longer what the edges point at.

Note the last row: a `calls` edge was rewired to a *different function in the same file*, not just to the file node. So this is not only a file-node fallback.

## Expected

Either the edges keep pointing at their original targets, or `merge-graphs` fails loudly. Silently preserving the edge count while changing what the edges mean is the worst case — `graphify query` then answers questions confidently from a graph that no longer matches the source.

## Notes

- Reproduces with `--code-only` on both sides (pure AST, no LLM involved).
- Not a label-collision issue: in the real-world case only 1 of the 4 rewired modules had a name that existed in the other project; `empresa.js`, `asyncHandler.js` and `conformidadeRegras.js` were unique across both.
- I did not dig into the id-remapping code, so I have no root-cause claim — only the reproducible symptom.

macOS 15 (arm64), Python 3.13, graphify installed via `uv tool install graphifyy`.

graphify 0.9.29.

hopstreax:
## Summary

Fixes #2261.

`graphify merge-graphs` could silently rewrite the direction of relationship edges when merging persisted `graph.json` files. Although the merged graph preserved the correct node and edge counts, directional relationships such as `imports` and `imports_from` could be reversed, producing incorrect dependency information (for example, a file appearing to import itself instead of the intended module).

This change preserves the original edge direction throughout the merge process while keeping the merged graph format unchanged (`directed: false`), matching the existing `merge-graphs` contract.

## Root Cause

Persisted `graph.json` files are stored as undirected graphs (`"directed": false`). During `merge-graphs`, they were loaded into an undirected `nx.Graph`, which does not preserve edge orientation. When the merged graph was serialized back to JSON, NetworkX emitted edge endpoints according to its internal ordering, causing some directional relationships to be written with swapped `source` and `target` values.

## Solution

* Preserve directional edge metadata during the merge workflow.
* Update `prefix_graph_for_global()` to remap directional edge metadata alongside prefixed node IDs, ensuring internal consistency after relabeling.
* Restore the original edge direction before writing the merged graph back to disk.
* Document the helper's behavior to reflect the updated contract.

## Tests

Added regression coverage for Issue #2261 by verifying that:

* import edge direction is preserved after `merge-graphs`,
* repo prefixes are still applied correctly,
* node and edge counts remain unchanged,
* import edges are not rewritten into self-imports or reversed relationships.

Also added a focused unit test for `prefix_graph_for_global()` to verify that directional edge metadata is correctly updated when node IDs are prefixed.

### Validation

Ran:

```bash
uv run pytest tests/test_merge_graphs_cli.py -v
uv run pytest tests/test_global_graph.py -v
```

Both test suites pass successfully.

### DIFF
--- a/graphify/build.py
+++ b/graphify/build.py
@@ -1450,15 +1450,20 @@ def prefix_graph_for_global(G: nx.Graph, repo_tag: str) -> nx.Graph:
     """Return a copy of G with all node IDs prefixed with repo_tag::.
 
     Labels are preserved unchanged (for display). A 'local_id' attribute
-    is added to each node so the original ID can be recovered. Edges are
-    rewritten to match the new prefixed IDs. The 'repo' attribute is set
-    on every node.
+    is added to each node so the original ID can be recovered. Edges and
+    their directional attributes (_src/_tgt) are rewritten to match the new
+    prefixed IDs. The 'repo' attribute is set on every node.
     """
     relabel = {n: f"{repo_tag}::{n}" for n in G.nodes}
     H = nx.relabel_nodes(G, relabel, copy=True)
     for node, data in H.nodes(data=True):
         data["repo"] = repo_tag
         data.setdefault("local_id", node.split("::", 1)[1])
+    for u, v, data in H.edges(data=True):
+        if "_src" in data and data["_src"] in relabel:
+            data["_src"] = relabel[data["_src"]]
+        if "_tgt" in data and data["_tgt"] in relabel:
+            data["_tgt"] = relabel[data["_tgt"]]
     return H
 
 
--- a/graphify/cli.py
+++ b/graphify/cli.py
@@ -2093,6 +2093,15 @@ def dispatch_command(cmd: str) -> None:
             # via node_link_data but older runs may have used "edges" (#738).
             if "links" not in data and "edges" in data:
                 data = dict(data, links=data["edges"])
+            # Preserve stored edge direction across undirected node_link_graph (#2261).
+            # Mirrors cli.py's query pattern and export.py's _src/_tgt restoration.
+            data = dict(
+                data,
+                links=[
+                    {**link, "_src": link.get("source"), "_tgt": link.get("target")}
+                    for link in data.get("links", [])
+                ],
+            )
             try:
                 G = _jg.node_link_graph(data, edges="links")
             except TypeError:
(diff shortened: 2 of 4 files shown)


safishamsi:
Thanks @hopstreax. Shipped in v0.9.30 (cherry-picked to v8). Closed-unmerged here, but it's in the release: https://github.com/Graphify-Labs/graphify/releases/tag/v0.9.30

PR #2272, opened 2026-07-28
fix(detect): honor .graphifyignore for the memory tree (fixes #2267)

### LINKED ISSUE #2267: graphify-out/memory/ cannot be excluded from the corpus — .graphifyignore is bypassed for it and there is no opt-out
# `graphify-out/memory/` cannot be excluded from the corpus — `.graphifyignore` is bypassed for it and there is no opt-out

**Version:** graphifyy 0.9.26
**Environment:** Python 3.12.13, Linux x86_64 (glibc 2.31)

## Summary

Answers written by `graphify save-result` land in `graphify-out/memory/*.md` and are then re-detected as corpus `document` files on the next run. That feedback loop is deliberate and useful in general, but there is currently **no way to turn it off** — not via `.graphifyignore`, not via `.gitignore`, not via any environment variable.

The result is that a repo's graph starts absorbing its own Q&A output, and node counts grow for reasons unrelated to the codebase. For projects where the graph is meant to describe only the source tree, that's unwanted and currently unavoidable short of deleting the directory.

## Why `.graphifyignore` doesn't reach it

`graphify-out` is in `_SKIP_DIRS`, but `detect()` re-adds the memory subdirectory as a second scan root and then explicitly exempts that tree from the ignore check:

```python
# detect.py:1221-1225
# Always include graphify-out/memory/ - query results filed back into the graph
memory_dir = root / GRAPHIFY_OUT / "memory"
scan_paths = [root]
if memory_dir.exists():
    scan_paths.append(memory_dir)

# detect.py:1249, 1260 — hidden/noise filtering skipped for the memory tree
in_memory_tree = memory_dir.exists() and str(scan_root).startswith(str(memory_dir))
if not in_memory_tree:
    ...

# detect.py:1317, 1322 — ignore patterns skipped for the memory tree
in_memory = memory_dir.exists() and str(p).startswith(str(memory_dir))
if not in_memory and _is_ignored(p, root, ignore_patterns, _cache=ignore_cache):
    continue
```

So a user-authored pattern is matched against every path *except* the one they're trying to exclude. Writing `graphify-out/memory/` into `.graphifyignore` silently does nothing, which is worse than an error — the config looks applied.

I also checked every `GRAPHIFY_*` environment variable the package reads (45 of them); none gates this.

## Reproduce

```bash
cd <any repo with a built graph>
graphify query "something"            # writes graphify-out/memory/query_*.md
printf 'graphify-out/memory/\n' > .graphifyignore

python -c "
from pathlib import Path
d = detect(Path('.'))
print('total_files:', d['total_files'])
print('memory hits:', [f for c in d['files'] for f in d['files'][c] if 'graphify-out' in f])
"
```

The memory files are still listed and still counted. Removing the directory is the only thing that changes the outcome — on a 69-file repo, `total_files` went 69 → 71 after two saved answers, and back to 69 only when `graphify-out/memory/` was moved away (the inclusion is gated purely on `memory_dir.exists()`).

## Requested behaviour

Any one of these would be enough:

1. **Honor `.graphifyignore` for the memory tree** — apply `_is_ignored` to it like every other path. Most discoverable, since users will reach for the ignore file first.
2. **A CLI flag / env var**, e.g. `--no-memory-corpus` / `GRAPHIFY_NO_MEMORY_CORPUS=1`.
3. **A config key** in whatever per-project settings graphify already reads.

Option 1 seems most consistent with the rest of the ignore handling, and keeps the default (memory included) unchanged for everyone who wants the feedback loop.

## Lower-priority note

Because the tree also bypasses the hidden/noise filtering at `detect.py:1260`, anything a user or tool drops into `graphify-out/memory/` is picked up unconditionally. Scoping the exemption to `query_*.md` / `path_query_*.md` / `explain_*.md` — the files `save-result` actually writes — would tighten that without affecting the intended behaviour.

Munawarx:
## Problem
`.graphifyignore` patterns are silently bypassed for `graphify-out/memory/`, so users cannot exclude their own saved Q&A output from the corpus. The tree gets re-ingested on every run and node counts grow for reasons unrelated to the source tree.

## Root cause
In `graphify/detect.py` the ignore check is guarded:
```python
if not in_memory and _is_ignored(p, root, ignore_patterns, _cache=ignore_cache):
    ignored.append(str(p)); continue
```
The `not in_memory` condition means user ignore patterns never apply to the memory tree. This is issue #2267, Option 1 (the maintainer-friendly fix the reporter requested).

## Solution
Remove the `not in_memory` guard so `_is_ignored` applies to the memory tree like every other path. Default behavior is unchanged: memory is still ingested unless explicitly ignored.

As a side benefit, the hidden/noise filter (which also keyed off `in_memory`) now correctly applies to the memory tree, so stray dotfiles dropped there are no longer unconditionally picked up.

## Testing
- Added `test_graphifyignore_excludes_memory_tree` in `tests/test_detect.py` (matches existing style, uses `tmp_path`).
- Verified the test FAILS on the unfixed code and PASSES with the fix.
- Regression: with no `.graphifyignore`, memory is still ingested (default loop intact).
- Edge cases checked: parent pattern `graphify-out/`, nested `proj/graphify-out/memory/`, and negation patterns behave correctly.

## Expected impact
Users can finally scope the graph to the source tree only. Removes silent config-ignored behavior.

Fixes #2267

### DIFF
--- a/graphify/detect.py
+++ b/graphify/detect.py
@@ -1334,7 +1334,7 @@ def detect(root: Path, *, follow_symlinks: bool | None = None, google_workspace:
             # Skip files inside our own converted/ dir (avoid re-processing sidecars)
             if str(p).startswith(str(converted_dir)):
                 continue
-        if not in_memory and _is_ignored(p, root, ignore_patterns, _cache=ignore_cache):
+        if _is_ignored(p, root, ignore_patterns, _cache=ignore_cache):
             ignored.append(str(p))
             continue
         if not _resolves_under_root(p, root):
--- a/graphify/detect.py
+++ b/graphify/detect.py
@@ -965,7 +965,15 @@ def _git_info_exclude(vcs_root: Path) -> Path | None:
     return exclude if exclude.is_file() else None
 
 
-def _load_dir_own_ignore(d: Path, *, gitignore: bool = True) -> list[tuple[Path, str]]:
+_IGNORE_SOURCE_GITIGNORE = "gitignore"
+_IGNORE_SOURCE_GRAPHIFYIGNORE = "graphifyignore"
+_IGNORE_SOURCE_INFO_EXCLUDE = "info_exclude"
+_IGNORE_SOURCE_CLI_EXCLUDE = "cli_exclude"
+
+
+def _load_dir_own_ignore(
+    d: Path, *, gitignore: bool = True
+) -> list[tuple[Path, str, str]]:
     """Read .gitignore/.graphifyignore directly inside *d* (not its ancestors).
 
     Merges .gitignore and .graphifyignore for this one directory (#1363):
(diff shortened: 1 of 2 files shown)


safishamsi (changes requested):
Thanks @Munawarx — the gap is real (user ignore patterns never reach graphify-out/memory/). But the one-line change over-reaches: `ignore_patterns` includes `.gitignore` and `$GIT_DIR/info/exclude`, not just `.graphifyignore`. Since graphify-out/ is generated output and is commonly gitignored, dropping the guard silently disables the memory round-trip by default for those repos — I verified `_is_ignored(<root>/graphify-out/memory/query_1.md, ...)` returns True with `graphify-out/` in .gitignore, which contradicts the 'default unchanged' claim. Please scope this to `.graphifyignore`-sourced patterns only (track pattern provenance and apply just those to the memory tree), and add a regression test that a gitignored graphify-out/ still keeps memory included. Happy to re-review.

Munawarx:
Thanks @safishamsi for the detailed review. I've reworked the fix to track pattern provenance exactly as you described.

**What changed:**

Each ignore pattern now carries a **source tag** (3rd tuple element): `'gitignore'`, `'graphifyignore'`, `'info_exclude'`, or `'cli_exclude'`.

- `_load_dir_own_ignore`: tags each pattern by its origin file (`.gitignore` vs `.graphifyignore`)
- `_load_graphifyignore`: tags `/info/exclude` entries and propagates per-directory tags
- `_is_ignored`: new optional `sources` parameter filters by provenance
- `detect()` main loop: memory-tree files call `_is_ignored` with `sources={'graphifyignore'}` — **only** `.graphifyignore`-sourced patterns are evaluated for memory

This means a gitignored `graphify-out/` no longer drops memory files, while `.graphifyignore` patterns still apply to the memory tree as before.

**Regression tests added:**

1. `test_gitignored_graphify_out_keeps_memory_included` — `graphify-out/` in `.gitignore` → memory files stay included (the exact scenario you verified was broken)
2. `test_info_exclude_does_not_drop_memory` — same protection for `/info/exclude`
3. `test_graphifyignore_excludes_memory_tree` — user's `.graphifyignore` still excludes memory tree

All 3 new tests pass; no regressions in the existing suite (225 passed). Happy to address anything else.

Munawarx:
Also addressed the cache collision flagged by the Graphify review bot.

**Issue:** `_is_ignored` cached results keyed by path alone. Since the same `_cache` dict is shared across calls in one scan, a non-memory file under `graphify-out/` (evaluated with all sources including `.gitignore`) would poison the cache — when a memory file's ancestor walk hit that cached entry, it returned `True` and wrongly dropped the memory file.

**Fix:** Cache key is now `(path, frozenset(sources))` when a source filter is active. Evaluations under different source filters never collide.

**Test added:** `test_memory_cache_not_poisoned_by_non_memory_eval` — places a non-memory `output.json` + memory `query_1.md` under the same gitignored `graphify-out/` and verifies the memory file survives.

All tests pass: 226 passed, 0 new failures.

PR #21, opened 2026-04-07
Add Nimble web intelligence for URL extraction and topic search

ilchemla:
## Why

graphify's `/graphify add` command currently uses Python's built-in `urllib` for fetching URLs — no JS rendering, no anti-bot handling, and basic HTML-to-markdown conversion. Modern websites increasingly rely on client-side rendering, which means `urllib` often returns empty or incomplete content.

[Nimble](https://nimbleway.com) is a web intelligence platform that turns any public webpage into clean, structured data. Their Extract API handles JS rendering, stealth unblocking, and returns clean markdown — all in a single API call. Their Search API lets AI agents search the live web and retrieve precise, full-page content in real time.

This PR integrates Nimble as an optional dependency, so graphify users can get significantly better web extraction out of the box.

## What changed

- **Nimble Extract** — `/graphify add <url>` now uses Nimble for real-time HTML/markdown extraction with stealth unblocking and JS rendering when `nimble_python` is installed (`pip install graphifyy[nimble]`). The built-in urllib path is used as an alternative when Nimble is not installed.
- **Nimble Search** — New `/graphify ingest-topic "<topic>"` command. Searches the live web via Nimble Search (general focus, deep mode) and saves results as graphify-ready markdown files with full page content — not just snippets.
- Added `nimble = ["nimble_python"]` optional dependency in `pyproject.toml`
- Updated all 4 skill files (Claude, Codex, OpenCode, OpenClaw) with the new `ingest-topic` command
- Updated README with web ingestion section, usage examples, and API key setup
- Unit tests (mocked) and live integration tests (auto-skipped without API key)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #2274, opened 2026-07-28
Common Lisp support, rebased onto v8 and moved into extractors/

fade:
## The ask

Do you want Common Lisp as a supported language in graphify? A yes or a no is all I need.
If the answer is no, I will close this and stop. Either answer is genuinely useful to us, and
thank you for the attention it takes to give one.

## Where it comes from

This is not a fresh implementation. It descends directly from **#172 by @jansaasman**, opened
on 9 April, and the history still carries his commits as its base:

- `Add Common Lisp support (language #21)`, 9 April
- `Common Lisp: walk wrapper macros, reader conditionals, more definer forms`, 10 April
- `Fix edge direction lost in JSON, Cypher, HTML, and Canvas exports`, 10 April

That work established the extraction approach: which forms count as definitions, how to walk
wrapper macros so definitions nested inside them are not lost, and how reader conditionals are
handled. Those decisions are his and they are still the ones in force here. What has been added
on top is consolidation rather than redesign:

- typing the rationale nodes so they stop leaking into unrelated node types
- wiring the file extensions into detection
- declaring the grammar as an optional extra rather than a hard dependency
- porting and extending the extraction suite, and silencing a deprecation warning the grammar
  emits
- listing the language in the README tables

If Common Lisp lands, the credit for the design belongs upstream of this PR.

## How it has tracked the codebase since April

graphify has moved a long way in the four months since #172 was opened, and this branch has been
rebased forward twice rather than left to rot:

1. first onto the v4 line, shortly after the original PR
2. then onto v8
3. and again onto current v8 for this PR, across roughly 640 commits of upstream movement,
   spanning the 0.8.35 to 0.9.29 releases

The rebases were not mechanical. Each one surfaced real conflicts where the two sides had grown
into the same place: the code extension set had gained `.mts`, `.cts`, CUDA, Metal and others
alongside the four Lisp ones; the optional extras table had gained `pascal`; the README language
row had grown from 29 grammars to 36. In every case the upstream side was taken and the Lisp
addition re-applied on top, so nothing of yours was reverted by the rebase.

## Reorganised to match where extract.py is going

The most significant change since #172 is structural, and it is the reason this PR looks
different from that one.

`extractors/MIGRATION.md` states that `graphify/extract.py` is being split into the
`graphify/extractors/` package, one language per PR, tracking #1212. Twenty seven language
modules already live there. #172 predates that split and added its extractor to `extract.py`,
which was correct at the time and is now against the grain.

So the extractor has been moved to `graphify/extractors/commonlisp.py`, following the playbook in
MIGRATION.md rather than inventing a shape:

- a self contained module importing only from `graphify.extractors.base`, never from
  `graphify.extract`, so the dependency direction the package requires is preserved
- a facade re-export from `graphify.extract`, kept alphabetical
- an entry in `LANGUAGE_EXTRACTORS` in `extractors/__init__.py`
- no test relocation, per MIGRATION.md's rule that a migration edits no tests outside
  `test_extractors_registry.py`. The registry sweep there now covers `commonlisp` automatically.

The effect on the file being emptied is the point: **`extract.py` grows by five lines rather than
by five hundred.** The extractor is 492 lines and all of it lands in the new package.

If a *new* language should go somewhere other than `extractors/`, say so and I will move it. The
migration document is written for relocating existing extractors and does not state where new
ones belong, so this is an inference from direction of travel, not a claim about your intent.

## What it is

- `.lisp`, `.cl`, `.lsp` and `.asd`, via `tree-sitter-commonlisp`.
- Nodes for packages, classes, structs, conditions, functions, generic functions, methods,
  macros and variables; `contains`, `calls`, `inherits`, `specializes` and `imports` edges.
- The grammar is a `commonlisp` optional extra, so a default `uv tool install graphifyy` is
  unchanged, matching how `dm`, `terraform` and `pascal` already behave.
- 23 tests in `tests/test_languages.py` against `tests/fixtures/sample.lisp`, each skipped when
  the grammar is absent, in the same style as the existing `dm` guard.
- 8 files, +814 / -4, of which 492 lines are the one new module.
- Run against a clean checkout of current v8 on the same machine: no test fails here that does
  not also fail on v8 unpatched.

⚠ One thing to flag rather than let you discover: `uv.lock` has no `tree-sitter-commonlisp`
entry, while the other optional grammars do. CI installs with `--frozen`, so the grammar will not
be present there and the Lisp tests will skip rather than run. They will not fail, but they will
also not be proving anything on your side until the lock is regenerated. I have not touched the
lock file, since hand editing it seemed worse than naming the gap.

## Why a ruling rather than a review

Two Common Lisp PRs are open, #172 since April and this one since June, and neither has drawn a
comment. I do not read that as neglect: your recently merged PRs are outside contributions landing
within a day or two. It reads instead like these two are asking a different kind of question. Not
"is this code correct" but "does graphify want Common Lisp, and another optional grammar
dependency, at all". That is a scope call, it is yours, and one word settles it.

We maintain a fork in the meantime, so a no costs us nothing and is a perfectly good answer. What
does cost something is a third rebase against silence.


27dru27-dru (approved):

safishamsi:
Landed in v0.9.46, just published to PyPI. Cherry-picked onto `v8` with your authorship preserved. graphify now has Common Lisp `.lisp`/`.cl`/`.lsp`/`.asd` support via the optional `[commonlisp]` extra. I added a small follow-up so `:use`/`require` import targets are sourceless stubs (no dangling import edges; cross-file package resolution works). Thanks @fade — really nice extractor.

PR #2275, opened 2026-07-28
feat(svelte): real Svelte 5 extractor (props, runes, functions, snippets, component usage)

zendoj:
## Problem

`extract_svelte()` calls `_extract_generic(path, _JS_CONFIG)`, which feeds the **whole `.svelte` file** to the tree-sitter-javascript parser. Svelte markup is not valid JavaScript, so the parse returns a top-level `ERROR` node, the walk never reaches an `import_statement` or a declaration, and the only thing recovered is what the regex rescue pass finds — the imports.

The consequence is that a Svelte component contributes essentially nothing to the graph. On a 233-component SvelteKit app I measured **2.79 nodes per `.svelte` file, with a median of 1** — the typical component produced only its own file node, and the mean was lifted by import stubs rather than by symbols.

Concretely, a 280-line component with 12 props, 2 `$derived`, a function and 2 snippets became **1 node**. And because no symbol nodes existed, `graphify affected` on a function that four components call answered:

```
No affected nodes found.
```

which reads exactly like dead code.

This mirrors what `extract_vue` already does for `.vue` (mask the non-script regions, parse the script with the TypeScript grammar) — `.svelte` just never got the same treatment.

## Approach

- **A region scanner, not a regex.** `<script>`/`<style>` are raw-text elements, so their content genuinely ends at the first closing tag; the hard part is finding the *opening* tag without being fooled by `<!-- <script> -->` or a `>` inside a quoted attribute value.
- Script blocks are parsed with **tree-sitter-typescript at their true byte offset**, so reported line numbers point into the original file.
- **Nodes:** the component, `$props()` props, `$state`/`$derived` bindings, every function including `const f = () => {}`, `{#snippet}` snippets, exported constants, local types.
- **Component usage** in the template becomes relation **`uses`** with context `renders`. Deliberately not `renders`: `DEFAULT_AFFECTED_RELATIONS` does not contain it, so a `renders` edge would be invisible to `graphify affected`.
- **Cross-file calls bind to the symbol, not just the file.** A bare callee name can't be resolved corpus-wide — the app I measured has eleven distinct `formatOre` definitions — but the import statement names the exact module, and symbol ids are reconstructible as `make_id(<file stem>, <name>)`. A wrong guess dangles and is dropped at build time; the `imports_from` edge still records the dependency.
- Aliased imports map alias → module; `$lib`/tsconfig aliases and workspace packages go through the existing resolution helpers.
- Dynamic imports are still recovered from raw source, including markup-only forms like `{#await import('./X.svelte')}` that no JS parser can see.

### Deliberately *not* nodes

Node count is not the goal, and a node per `const x = 1` makes the graph noisier rather than more navigable. Excluded: scalar locals, loop variables, closures nested inside other functions, and anonymous `$effect` callbacks — though an effect's body is still swept for calls, attributed to the component, so what it *does* still reaches the graph.

Rune variants are an **attribute, not a node kind**: `$state` and `$state.raw` are one kind with the exact spelling kept in `rune`, so the graph doesn't fragment across spellings.

## Results

Same repo, same command, 233 `.svelte` files:

| | before | after | |
|---|---:|---:|---|
| nodes per file (mean) | 2.79 | **16.51** | 5.9× |
| nodes per file (median) | 1 | **12** | 12× |
| edges touching a `.svelte` node | 2 698 | **11 206** | 4.2× |
| the 280-line component above | 1 | **18** | |

The `affected` query that previously returned nothing now returns four exact call sites with correct line numbers, each verified against the source.

### Quality

Because node count is gameable, the extractor is scored against a **second, independent oracle** (regex + brace matching, sharing no code with the tree-sitter path). 100 randomly sampled files:

| category | agreed | extractor-only | oracle-only | precision | recall |
|---|---:|---:|---:|---:|---:|
| prop | 220 | 2 | 0 | 99.1% | 100% |
| rune | 954 | 3 | 0 | 99.7% | 100% |
| function | 398 | 5 | 0 | 98.8% | 100% |
| snippet | 94 | 0 | 0 | 100% | 100% |
| component usage | 1 105 | 0 | 1 | 100% | 99.9% |
| **total** | **2 771** | **10** | **1** | **99.6%** | **100%** |

Every remaining disagreement was checked by hand against source: all are limitations of the *oracle* (multi-declarator statements, arrow functions with return-type annotations, `async function`) or the extractor's deliberate choice to skip nested closures. None were extractor errors.

## Incidental fix

`make_id` casefolds, so `type State` and `let state` in the same file produce the **same id**, and the second declaration was silently swallowed by the seen-ids guard — a lost symbol, not a merge. This isn't Svelte-specific; it's reachable from any extractor that declares both a type and a value with names differing only in case. Colliding declarations are now disambiguated by kind.

## Tests

Adds `tests/test_svelte_extraction.py` — 13 cases covering region scanning (including the commented-out-`<script>` trap), props/runes/functions/snippets, renamed props (`class: className` exposes `class`), rune variants sharing one kind, the `{#if i < items.length}` less-than false positive, barrel siblings (`Alert` / `AlertTitle` resolving to one module but being two usages), the id collision, symbol-level call binding, markup-only dynamic imports, and the noise exclusions.

Full suite, same machine:

- `v8` baseline: **43 failed, 3599 passed, 166 skipped**
- this branch: **43 failed, 3612 passed, 166 skipped**

Identical failure sets — verified by diffing the two `FAILED` lists, not just the totals. The 43 are pre-existing and unrelated (terraform grammar not installed, `serve_http`, `skillgen`, ollama). The +13 are the new tests.

## Known limits

- Dynamic components (`<svelte:component this={X}>`, `<Foo />` where `Foo` is a variable) are emitted with `confidence: INFERRED` and context `renders_dynamic`.
- Template call detection resolves an identifier only when it is declared or imported in the same file; unknown names are dropped rather than guessed.
- Parsing the script with tree-sitter rather than `svelte/compiler` is a deliberate trade: it keeps extraction dependency-free and in-process, at the cost of a hand-written template pass. The measured template-edge precision (100%) is what justified staying with it.

### DIFF
--- a/graphify/extract.py
+++ b/graphify/extract.py
@@ -52,6 +52,7 @@ from graphify.extractors.razor import extract_razor  # noqa: F401
 from graphify.extractors.rust import extract_rust  # noqa: F401
 from graphify.extractors.sln import extract_sln  # noqa: F401
 from graphify.extractors.sql import extract_sql  # noqa: F401
+from graphify.extractors.svelte import extract_svelte  # noqa: F401
 from graphify.extractors.terraform import extract_terraform  # noqa: F401
 from graphify.extractors.verilog import extract_verilog  # noqa: F401
 from graphify.extractors.zig import extract_zig  # noqa: F401
@@ -1396,62 +1397,6 @@ def _emit_rescued_import(
     })
     result.setdefault("edges", []).append(edge)
     existing_ids.add(node_id)
-
-
-def extract_svelte(path: Path) -> dict:
-    """Extract imports from .svelte files: script-block via JS AST + template regex fallback.
-
-    Tree-sitter only sees the <script> block. Svelte template syntax like
-    {#await import('./X.svelte')} lives in the markup layer and is invisible
-    to the JS parser, so a regex pass covers those dynamic imports.
-    """
-    result = _extract_generic(path, _JS_CONFIG)
-    try:
-        import re as _re
-        src = path.read_text(encoding="utf-8", errors="replace")
-        existing_ids = {n["id"] for n in result.get("nodes", [])}
-        # Source file node ID must match the one _extract_generic creates:
-        # _make_id(str(path)) - single arg, no stem prefix. Otherwise the source
-        # endpoint is a phantom node and build_from_json drops the edge (#701).
-        file_node_id = _make_id(str(path))
-        aliases = _load_tsconfig_aliases(path.parent)
-        base_url = _load_tsconfig_base_url(path.parent)
-        for m in _re.finditer(r"""import\(\s*['"]([^'"]+)['"]\s*\)""", src):
-            raw = m.group(1)
-            if not raw:
-                continue
-            # Resolution + emit shared with the static pass below: relative
-            # paths and tsconfig aliases probe real on-disk extensions (#716,
-            # #701), and a target that IS a real file emits an edge stamped
-            # with target_file instead of an absolute-id ghost stub (#2195).
-            _emit_rescued_import(
-                result, existing_ids, file_node_id, path, raw,
-                "dynamic_import", aliases, base_url,
-            )
-        # Static imports inside <script> blocks. The JS tree-sitter parser fed
-        # the full .svelte file produces a top-level ERROR node (HTML markup
-        # is not valid JS), so import_statement nodes are never reached and
-        # static imports are silently dropped (#713). Regex over each script
-        # body recovers them.
-        script_re = _re.compile(
-            r"<script\b[^>]*>([\s\S]*?)</script\s*>", _re.IGNORECASE
-        )
-        static_import_re = _re.compile(
-            r"""import\s+(?:[^'"`;]+?\s+from\s+)?['"]([^'"]+)['"]"""
-        )
-        for script_match in script_re.finditer(src):
-            script_body = script_match.group(1)
-            for m in static_import_re.finditer(script_body):
-                raw = m.group(1)
-                if not raw:
-                    continue
-                _emit_rescued_import(
-                    result, existing_ids, file_node_id, path, raw,
-                    "imports_from", aliases, base_url,
-                )
-    except Exception:
-        pass
-    return result
 
 
 def extract_astro(path: Path) -> dict:
--- a/graphify/extractors/svelte.py
+++ b/graphify/extractors/svelte.py
@@ -241,6 +241,11 @@ def _prop_entries(node, source: bytes) -> list[tuple[str, str]]:
             key = c.child_by_field_name("key")
             value = c.child_by_field_name("value")
             public = _read_text(key, source) if key is not None else None
+            if public:
+                # A hyphenated prop must be written as a quoted key
+                # (`"data-slot": slot`). The prop is named data-slot; the quotes
+                # are syntax, not part of the name.
+                public = public.strip("\"'`")
             locals_ = _pattern_names(value, source)
             if public:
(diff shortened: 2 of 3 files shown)


zendoj:
Follow-up pushed (`2c6741b`). I re-scored the extractor against the independent oracle on the **full corpus of both apps** — 455 components, not a sample — and that turned up three defects the 100-file sample had missed. All are fixed, with regression tests.

**The one worth calling out:** I had made the template brace-depth tracker string- and comment-aware, on the theory that it would be more correct. It measured *worse*, and the failure mode is nasty — a single miscount pins the depth above zero, so every remaining component in the file looks like it lives inside an expression and is silently dropped. Real templates break it constantly:

- an apostrophe in markup text — `Don't`
- an apostrophe inside a comment within an expression — `class={cn('x', // the CRM's own scope`
- a regex literal whose escaped slashes read as a line comment — `replace(/^https?:\/\//, '')`

That last one cost a 737-line component **all 30** of its component-usage edges.

Counting braces plainly can only be fooled by a brace inside a string, which costs one local misjudgement instead of a whole file. Over those 455 components, plain counting balanced in **every single file**; the lexer did not. So it now counts plainly, with a self-check that drops the depth filter entirely when braces do not balance — an unparseable file should lose a little precision, never all of its components.

The other two: a self-import is a *recursive* component (`import Self from './ChainTree.svelte'` used as `<Self />`) and now emits the self-edge rather than being discarded; and quoted destructure keys keep only the name, so `"data-slot": slot` yields a prop called `data-slot`.

Result on both corpora, full sets:

| corpus | components | symbols | precision | recall |
|---|---:|---:|---:|---:|
| app A | 233 | 6 079 | 100.0% | 100.0% |
| app B (held out) | 222 | 1 492 | 100.0% | 100.0% |

App B was never inspected while developing the extractor — it is what surfaced the regex-literal bug.

Test suite is unchanged against `v8`: I diffed the two `FAILED` lists rather than comparing totals, and they are identical (42 pre-existing failures — terraform grammar not installed, `serve_http`, `skillgen`, ollama). 17 new Svelte tests.

zendoj:
One more fix pushed (`efb509e`), found by a second verifier rather than by the oracle.

`precision.py` scores declarations and component-usage edges, which left the **call** edges unmeasured — so "100% precision" would have been quietly narrower than it sounded. I wrote a different kind of check for those: for every `calls` edge, the cited line must actually mention the callee, and a cross-file target must really declare the symbol.

It immediately found that call sites were reported at **script-block lines rather than file lines**. The declaration walk already corrected for the block offset; the call sweep did not. So in any component whose instance `<script>` sits below a `<script module>` block, every call site pointed several lines too high — at code with nothing to do with the edge. Wrong line numbers are worse than missing ones in a navigation tool.

Verification surface now, across both apps:

| surface | count | how | result |
|---|---:|---|---|
| props, runes, functions, snippets, component usage | 7 571 | independent oracle | 100% |
| `calls` edges | 2 314 | cited line mentions callee; cross-file target declares it | 100% |
| `renders_dynamic` (`INFERRED`) | 5 | read by hand, all five | all genuinely dynamic |

The 5 inferred edges are 1.7‰ of component edges — `$derived(role.icon)`, `$derived(ICONS[...])`, and a lazily-loaded `$state<Component>`. Runtime values that cannot be resolved statically, which is why they carry `INFERRED`.

Test suite still matches `v8` exactly: I diff the two `FAILED` lists rather than compare totals — 42 pre-existing failures, no regressions, 18 new Svelte tests.

Issue #2277, opened 2026-07-28
(BUG) mcp MAJOR version bump 2.0.0 breaks graphify MCP server


grabear:
The mcp library just bumped their library to 2.0.0, which breaks the MCP server on fresh installs:  https://github.com/modelcontextprotocol/python-sdk/releases

safishamsi:
Fixed in v0.9.30. The mcp and all extras now require mcp>=1,<2 (resolving to 1.29.0), so a fresh install no longer pulls the breaking 2.0.0. Verified end-to-end from the published wheel: graphify.serve and mcp.types.AnyUrl import cleanly. Porting to the mcp 2.x API (to lift the cap) is tracked in #2308. https://github.com/Graphify-Labs/graphify/releases/tag/v0.9.30

PR #2278, opened 2026-07-28
fix(dependencies): restrict mcp version to <=1.27.1 to prevent breaki…

grabear:
…ng changes in mcp 2.0.0


grabear:
Closes https://github.com/Graphify-Labs/graphify/issues/2277

safishamsi:
Thanks @grabear — you flagged the right problem (unpinned mcp resolving to the breaking 2.0.0). We shipped the cap in 0.9.30, but as `mcp>=1,<2` rather than `<=1.27.1`: the latter excludes 1.28/1.29, and 1.29.0 is confirmed working, so it resolves there now. We also capped the `all` extra (still bare `mcp` here) and kept the uv.lock delta to just the mcp/starlette lines. Closing since the fix is in; appreciate the catch. Porting serve.py to the mcp 2.x API is tracked as a follow-up.

PR #2283, opened 2026-07-29
fix(llm): honor GRAPHIFY_API_TIMEOUT in the bedrock backend

### LINKED ISSUE #2284: bedrock backend ignores GRAPHIFY_API_TIMEOUT (botocore 60s default read timeout)
## Summary

`GRAPHIFY_API_TIMEOUT` (and the `--api-timeout` flag that sets it) has no effect on the `bedrock` backend. A long generation is cut off at botocore's 60s default read timeout and fails with `Read timeout on endpoint URL`, no matter how high the timeout is set.

## Reproduction

```
GRAPHIFY_API_TIMEOUT=1800 \
GRAPHIFY_BEDROCK_MODEL=global.anthropic.claude-opus-5 \
GRAPHIFY_LLM_TEMPERATURE=none \
AWS_PROFILE=<your-profile> \
graphify .
...
[graphify] chunk 2/3 failed: Read timeout on endpoint URL: "https://bedrock-runtime.us-east-1.amazonaws.com/model/global.anthropic.claude-opus-5/converse"
[graphify] chunk 1/3 failed: Read timeout on endpoint URL: "https://bedrock-runtime.us-east-1.amazonaws.com/model/global.anthropic.claude-opus-5/converse"
[graphify] WARNING: 2/3 semantic chunk(s) failed — see errors above. Partial results returned.
```

Large doc chunks on a slow opus-class model routinely generate for longer than 60s. Converse is non-streaming, so the whole response has to finish before the first byte arrives; small chunks squeak in under 60s while large ones time out — even with `GRAPHIFY_API_TIMEOUT` raised well past the generation time.

## Root cause

Both `bedrock-runtime` clients are constructed with no botocore config: the primary extraction path (`_call_bedrock`) and the secondary dispatch path (the `bedrock` branch of `_call_llm`). With no `config=`, boto3 uses botocore's default `read_timeout` of 60s and never sees `GRAPHIFY_API_TIMEOUT`. The env var only ever reached the OpenAI/Anthropic HTTP clients and, later, the claude-cli subprocess — `_resolve_api_timeout()` is simply never applied to the boto3 clients.

The README env-var table advertises `GRAPHIFY_API_TIMEOUT` as general, which makes the Bedrock gap easy to trip over.

## Suggested fix

Build both Bedrock clients with an explicit `botocore.config.Config` that wires `read_timeout` to `_resolve_api_timeout()`, mirroring how the other backends already consume it. This is the same class of gap that was closed for the `claude-cli` subprocess (#1112 / #1111) and the secondary LLM dispatch path (#1442); Bedrock is the last cloud backend still ignoring the knob.

Not a duplicate of #1112 (that issue is the `claude-cli` subprocess hardcoding `timeout=600`) or #1442 (the OpenAI/Anthropic clients in the secondary dispatch path missing `timeout=`) — both are non-Bedrock paths, and neither touches boto3 client construction, which has its own separate `read_timeout` default that no env var reaches.

zhiyanliu:
## Summary

Both `bedrock-runtime` clients — the primary extraction path (`_call_bedrock`) and the secondary dispatch path (the `bedrock` branch of `_call_llm`) — are built with no botocore config, so Converse uses botocore's 60s default read timeout and ignores `GRAPHIFY_API_TIMEOUT` / `--api-timeout` entirely. A long opus-class generation then dies with `Read timeout on endpoint URL` regardless of how high the timeout is set.

This wires both client constructions through a `botocore.config.Config` that sets `read_timeout=_resolve_api_timeout()` (default 600s), a 10s `connect_timeout`, and `retries` from `_resolve_max_retries()` in adaptive mode.

This mirrors the fixes that closed the same gap for the `claude-cli` subprocess (#1112 / #1111) and the secondary LLM dispatch path (#1442). Bedrock was the last cloud backend still ignoring the knob, and the README env-var table already advertised the timeout as general — so the gap was easy to trip over.

## Root cause

`session.client("bedrock-runtime")` was called with no `config=`, at both sites. botocore's default `read_timeout` is 60s, so a Converse call that generates for longer is cut off at 60s with a `Read timeout on endpoint URL` error — the env var never reaches boto3.

## What changed

- `_call_bedrock` and the `_call_llm` bedrock branch now build the client with `botocore.config.Config(read_timeout=_resolve_api_timeout(), connect_timeout=10, retries=…)`.
- README env-var row updated to include Bedrock.
- `_fake_boto3` test fixture registers `botocore.config` and captures the client config; added coverage asserting the timeout is wired (override + default) alongside the existing image-bytes test.

## Verification

Reproduced against a real Bedrock model (`global.anthropic.claude-opus-5`): before, large doc chunks failed with `Read timeout on endpoint URL` under a raised `GRAPHIFY_API_TIMEOUT`; after, the value reaches boto3's `read_timeout` and the chunks complete. Added a test that intercepts client construction and asserts `read_timeout == GRAPHIFY_API_TIMEOUT` (and the 600s default when unset). `ruff check` clean on the changed files; the full image-vision suite passes.

Happy to revisit the `connect_timeout` / retry-mode choices if you'd prefer to match a different convention.

Fixes #2284

### DIFF
--- a/README.md
+++ b/README.md
@@ -513,7 +513,7 @@ These are only needed for **headless / CI extraction** (`graphify extract`). Whe
 | `AWS_*` / `~/.aws/credentials` | AWS Bedrock — standard credential chain | `--backend bedrock` (no API key, uses IAM) |
 | `GRAPHIFY_MAX_WORKERS` | AST parallelism thread count | optional — also `--max-workers` flag |
 | `GRAPHIFY_MAX_OUTPUT_TOKENS` | Raise output cap for dense corpora | optional — e.g. `32768` for large files |
-| `GRAPHIFY_API_TIMEOUT` | Per-call timeout in seconds for HTTP, claude-cli, and Anthropic SDK backends (default: 600) | optional — also `--api-timeout` flag |
+| `GRAPHIFY_API_TIMEOUT` | Per-call timeout in seconds for HTTP, claude-cli, Anthropic SDK, and Bedrock backends (default: 600) | optional — also `--api-timeout` flag |
 | `GRAPHIFY_MAX_RETRIES` | How many times to retry a rate-limited (429) request before giving up (default: 6; honors `Retry-After`) | optional — raise for strict per-org limits (e.g. kimi); `0` disables |
 | `GRAPHIFY_FORCE` | Force graph rebuild even with fewer nodes | optional — also `--force` flag |
 | `GRAPHIFY_GOOGLE_WORKSPACE` | Auto-enable Google Workspace export | optional — set to `1` |
--- a/graphify/llm.py
+++ b/graphify/llm.py
@@ -1609,6 +1609,7 @@ def _call_bedrock(model: str, user_message: str, max_tokens: int = 8192, *, deep
     """Call AWS Bedrock via boto3 Converse API using the standard AWS credential chain."""
     try:
         import boto3
+        import botocore.config
         import botocore.exceptions
     except ImportError as exc:
         raise ImportError(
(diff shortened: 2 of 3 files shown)


zhiyanliu:
Thanks for the review. On the double-retry-layering note — I looked at how the application-level retry actually classifies failures, and I don't think the two layers meaningfully stack here, but wanted to lay out the evidence so a maintainer can decide.

The adaptive retry in `_extract_with_adaptive_retry` only re-runs a chunk when `_looks_like_context_exceeded` returns true; every other exception is re-raised. That helper matches on context-window markers (`context length`, `too many tokens`, `prompt is too long`, and similar). A boto3 `Read timeout on endpoint URL` doesn't match any of them, so a read timeout is never retried by the application layer — it surfaces as a failed chunk, which is exactly the behavior in the issue. The botocore `max_attempts` and the application loop therefore act on largely disjoint failure classes: botocore retries throttling/5xx/connection errors, while the application layer only splits on context overflow and re-raises the rest.

The retry policy on the Bedrock client also mirrors the existing cloud backends rather than introducing a new pattern — the OpenAI, Anthropic, and Azure clients each already pass `max_retries=_resolve_max_retries()` to their SDK's built-in retry, on top of the same application loop. This change brings Bedrock in line with them; if the layering is a concern it's a pre-existing property shared by all four backends, not something this PR adds.

Happy to drop the `retries` block (or the `adaptive` mode) and keep only the read/connect timeouts if you'd prefer the Bedrock client not to carry an SDK-level retry policy.

safishamsi:
Thanks @zhiyanliu. Shipped in v0.9.30 (cherry-picked to v8), with a follow-up mapping GRAPHIFY_MAX_RETRIES to botocore max_attempts + 1 (max_attempts counts the initial call). Closed-unmerged here, but it's in the release: https://github.com/Graphify-Labs/graphify/releases/tag/v0.9.30

Issue #2199, opened 2026-07-25
stat-index.json keyed by absolute path and never pruned: 0% cache hits for any root that didn't write it (11,614 entries for 1,130 files across 4 roots)


aeyeopsdev:
### Summary

`graphify-out/cache/stat-index.json` is keyed by **absolute** path and never prunes entries for roots that are no longer in play. When the same corpus is reached through more than one absolute path — a second mount of the same share, a CI/snapshot clone under `/tmp`, a moved checkout — the index accumulates one full set of entries per root, and a run from a root that isn't the one that wrote the index gets **zero cache hits**: every file re-extracts, every run, forever.

### Observed

A ~1,130-file repo, graphify 0.9.17, on macOS. Every `graphify update .` reported all files uncached:

```
AST extraction: 1131/1131 uncached files (100%) [16 workers]
```

The stat index had grown to **11,614 entries for ~1,130 files** — four roots, three of them dead:

| root | entries |
|---|---|
| `/System/Volumes/Data/mnt/dev/mm/<repo>` | 3198 |
| `/mnt/j/dev/mm/<repo>` (same share, second mount path) | 3062 |
| `/tmp/platform_snap.QSy9DT/repo` (throwaway clone, gone) | 2677 |
| `/tmp/platform_snap2.kaYXPI/repo` (throwaway clone, gone) | 2677 |

The index that happened to be committed was written from the `/mnt/j` root, so on the machine running from `/System/Volumes/Data/...` nothing matched.

Deleting `stat-index.json` restored caching immediately — the next run reported `3181 files cached/unchanged, 0 re-extracted`. It regrows the same way.

### Why this is worth fixing rather than documenting

- The failure is **silent**: output says "uncached", which reads as "these files changed", not "your cache is keyed to a path you aren't at". Nothing points at the index.
- The cost scales with corpus size — here it was a full re-extract of 1,130 files on every run, plus a full rewrite of a 40 MB `graph.json`.
- `graphify-out/` is documented as commit-worthy, so a cache index written on one machine travels to machines with a different absolute root, where it is guaranteed to miss.
- Dead roots are never reclaimed. Two of the four here were `/tmp` clones that had not existed for days.

`manifest.json` already stores **relative** keys, and the incremental reference explicitly calls this out for the manifest (#1417: "manifest keys stay relative to the scan root — portable across clones/machines, so `--update` keeps matching cached files instead of missing every one after a move"). The stat index appears to have missed that same treatment.

### Suggested fix

[...]

PR #2285, opened 2026-07-29
fix(cache): re-anchor cached ids so a warm hit can't replay another root (#2257)

### LINKED ISSUE #2257: AST cache replays node ids minted under the original root after a corpus move/clone
Surfaced while closing the node-id leak class (#2231/#2243). Distinct vector, own issue.

**Problem.** The per-file AST cache stores extracted nodes/edges keyed by content + relative path (namespaced by extractor version). When the same corpus is reached under a *different absolute root* (a clone, a moved checkout, a second mount) and the cache is carried along, a warm cache hit replays nodes/edges whose ids were minted from the ORIGINAL absolute root. The id-remap / final-canonicalization passes in extract() cannot fix these, because those ids no longer match any current-path form (the current run never sees the original root), so a stale machine-slug id can survive into graph.json on a cache hit.

This is separate from #2231/#2243 (which fix producers on a cold run) and from #2199 (stat-index portability, already fixed) — it's the *AST result cache* replaying pre-canonicalization ids.

**Fix direction.** Canonicalize node ids / edge endpoints when replaying a cached entry (re-anchor any id/endpoint minted from the entry's stored absolute source_file to the current root), or store cache entries already canonicalized (root-relative ids) so a replay is portable by construction. Mirror the manifest.json / stat-index.json portability approach (#2199).

**Test.** Extract a corpus under root A (populate cache), copy the tree + graphify-out to root B, extract under B with the warm cache, assert no node id / edge endpoint contains the root-A slug and ids match a cold B extraction.

Kaushik2003:
Fixes #2257.

## Problem

Extractors mint node ids from the path **string** they are handed (`_make_id(str(path))`, `_file_node_id(path)`), so an AST cache entry written under root A embeds A's slug in every id and edge endpoint. `save_cached` relativized only `source_file` — never the ids.

Because `extract()`'s id-remap / final-canonicalization passes key their rewrites off the **current** run's paths, an A-derived id matches no key on a warm hit under root B (a clone, a moved checkout, a second mount), so the stale machine slug survives into `graph.json`.

Reproduced before touching any code — the same corpus, warm vs cold under root B:

```
run B (WARM cache): ['tmp_..._a_dir_slug_aaa_corpus_app_py',
                     'tmp_..._a_dir_slug_aaa_corpus_pkg_mod_base', ...]
run B (COLD):       ['app', 'app_run', 'pkg_mod', 'pkg_mod_base', 'pkg_mod_base_hello']
warm == cold: False
```

This is distinct from #2231/#2243 (which fix producers on a cold run) and from #2199 (stat-index portability) — it is the AST result cache replaying pre-canonicalization ids.

## Fix

Store entries root-anchored, re-anchor on read — the same store-portable / re-anchor-on-load contract `source_file` (#777) and the stat index (#2199) already use:

- `_relativize_ids_in` (in `save_cached`, after the `source_file` pass) replaces the root's contribution with a `$graphify-root$` marker.
- `_absolutize_ids_in` (in `load_cached`) restores what the current run's extractor would mint — the pre-remap form every downstream pass expects, so a replay reproduces a cold run exactly and **no other pass changed**.

Details worth reviewing:

- The anchor is **derived per entry** rather than assumed equal to the scan root (`normalize_id` distributes over path joins), so a symlinked root (macOS `/tmp` → `/private/tmp`) or relative inputs decompose exactly.
- Only **absolute** root spellings may anchor an id. A relative `root` ("src") is a strip candidate but never a restore candidate, so admitting it would rewrite an already-canonical `src_utils_foo` into `<abs-root-slug>_utils_foo` on the `save_semantic_cache` path (which forwards `root` unresolved). Covered by a regression test.
- The walk covers the **whole payload** rather than a bucket list, because the id form is self-identifying (a long casefolded path slug). Hand-enumeration was already incomplete: this also reaches `raw_calls[].caller_nid`, `swift_extensions[].nid`, `edges[].target_file`, `bash_sources[].source_file` and `*_type_table.path` — the last three being resolution inputs that would otherwise still point at root A and diverge warm-vs-cold. Values only, never dict keys.
- `save_cached`'s deepcopy is now **unconditional**. The old truthiness gate skipped it for a payload whose only content lives outside `nodes`/`edges` (empty `nodes` beside a populated `bash_sources`), which would have let the transform mutate the caller's dict — and `extract()`'s `id_remap` is keyed on the absolute form, so that would break cold runs.

A cache entry on disk now looks like:

```json
{ "id": "$graphify-root$_run_sh__entry",
  "source_file": "run.sh",
  "target_file": "$graphify-root$/lib/common.sh" }
```

## Tests

`tests/test_cache.py`:

- `test_warm_cache_from_another_root_does_not_leak_that_root` — the issue's scenario. Extract a python/C/bash/markdown corpus under root A, copy the tree **and** `graphify-out` to root B, extract under B on the warm cache. Asserts the run is genuinely warm (zero extractor invocations), no node id or edge endpoint carries A's slug, the on-disk entries hold neither A's slug (case-insensitive — ids are casefolded, paths are not) nor an absolute path, a cold run still yields canonical ids, and warm ≡ cold.
  - The fixture deliberately avoids JS/TS: those suffixes are in `_JS_CACHE_BYPASS_SUFFIXES` and never cached, which would make every warm assertion vacuous.
- `test_cached_ids_round_trip_under_the_same_root` — the stored form must restore to the exact absolute-derived id, or a same-root warm hit would break `extract()`'s remap.
- `test_relative_root_does_not_reanchor_an_already_canonical_id` — the strip/restore asymmetry above.

**Full suite: 3787 passed.** The 4 `test_ollama_retry_cap.py` failures are pre-existing in my environment (`openai` not installed) — confirmed by stashing this branch and re-running.

Additionally verified out-of-band on graphify's own 334-file tree (216 cached files, 0 warm misses, warm graph identical to cold at 3067 nodes / 5865 edges), across the default layout, `--out` (decoupled `cache_root`), a symlinked root, and the parallel process pool.

## Note for the maintainer

Pre-fix entries carry no marker and their content hash never changes, so they **cannot self-heal**. They are swept when the release bumps the version, since AST entries live under `cache/ast/v{version}/` — so the version bump is part of this fix landing, not incidental to it. Per repo convention this PR does not touch `CHANGELOG.md` or the version itself.

### DIFF
--- a/graphify/cache.py
+++ b/graphify/cache.py
@@ -514,6 +514,202 @@ def _normalize_source_file_value(src: "str | Path", root_resolved: Path) -> str:
     return rel.replace(os.sep, "/")
 
 
+# Storage marker standing in for the absolute root a cached id/path was minted
+# under (#2257). Extractors mint node ids from the path STRING they are handed
+# (``_make_id(str(path))``, ``_file_node_id(path)``), so a cache entry written
+# under root A embeds A's slug in every id and edge endpoint. Those are only
+# rewritten to the canonical root-relative form by extract()'s whole-graph
+# id-remap, which keys its rewrites off the CURRENT run's paths — so on a warm
+# hit under root B (a clone, a moved checkout, a second mount) the stored ids
+# match no key and A's machine slug survives into graph.json. Entries are
+# therefore stored root-anchored: the root's contribution is replaced by this
+# marker on write and re-anchored to the current root on read, so a replay is
+# portable by construction and reproduces exactly what a cold run under the
+# current root would have minted (the pre-remap form every downstream pass in
+# extract() expects). Same store-portable/re-anchor-on-load contract as
+# ``source_file`` (#777) and the stat index (#2199). Neither ``$`` nor ``-`` can
+# occur in a normalized id (``normalize_id`` drops every non-word character), and
+# no plausible source literal — a shell ``$root``, a template ``${root}`` — opens
+# with this exact token, so the marker cannot collide with extractor output.
+_ROOT_MARKER = "$graphify-root$"
+
+
+def _id_anchor(path_str: str, rel_str: str) -> str:
+    """Return the id-slug prefix ``path_str`` contributes above ``rel_str``.
+
+    ``_make_id`` normalizes a whole path string, and normalization distributes
+    over path joins (every separator run collapses to one ``_``), so an id
+    minted from an absolute path decomposes exactly as
+    ``normalize_id(root) + "_" + normalize_id(rel)``. Recovering the root half
+    from the file's own two spellings — rather than assuming it equals the
+    scan root — keeps the decomposition exact when the extractor was handed an
+    unresolved or relative path (a symlinked root such as macOS ``/tmp`` ->
+    ``/private/tmp``, or inputs given relative to CWD).
+
+    Returns "" when ``path_str`` contributes no prefix (it already IS the
+    relative form) or when the two spellings disagree about the tail — both
+    mean "nothing to re-anchor", which leaves the entry untouched.
+    """
+    from graphify.ids import normalize_id  # ids imports only re/unicodedata: no cycle
+
+    full = normalize_id(path_str)
+    tail = normalize_id(rel_str)
+    if not tail or full == tail:
+        return ""
+    suffix = "_" + tail
+    return full[: -len(suffix)] if full.endswith(suffix) else ""
+
+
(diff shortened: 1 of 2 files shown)


safishamsi:
Thanks @Kaushik2003. Shipped in v0.9.30 (cherry-picked to v8 to preserve authorship) — closes #2257. Reproduced-before-fix, digest-stable, matches the manifest/stat-index portability contract. Closed-unmerged here, but it's in the release: https://github.com/Graphify-Labs/graphify/releases/tag/v0.9.30

PR #2288, opened 2026-07-29
fix(llm): read the first text block of a bedrock Converse response

### LINKED ISSUE #2287: bedrock backend reads only the first Converse content block, so reasoning models look like truncation
## Summary

The bedrock backend reads only the first content block of a Converse response. Reasoning-capable models place a `reasoningContent` block before the text block, so graphify receives no text at all — every call parses to zero nodes, gets classified as a hollow/truncated response, and the chunk is bisected until the recursion cap. The run reports partial results or `produced no nodes` while the model was in fact answering correctly and being billed for it.

## Root cause

Converse returns `output.message.content` as a **list** of content blocks, and the specification does not guarantee that a text block is first — `reasoningContent`, `toolUse`, and other block types may precede it. Both bedrock call sites index position 0:

```python
text = resp.get("output", {}).get("message", {}).get("content", [{}])[0].get("text", "{}")
```

- `_call_bedrock` — primary semantic extraction.
- the `backend == "bedrock"` branch of `_call_llm` — secondary dispatch.

When block 0 is a reasoning block the `.get("text", …)` default is returned, so `_parse_llm_json` sees `{}` → zero nodes → `_response_is_hollow` is true → `finish_reason` is rewritten to `length` → the adaptive retry bisects. Splitting cannot help, because the position assumption fails identically at every chunk size.

## Reproduction

Probing Converse directly with `global.anthropic.claude-opus-5`:

```
stopReason: end_turn
num content blocks: 2
  block[0] keys=['reasoningContent']
  block[1] keys=['text']
      text[:80]='{"nodes":[{"id":"a","label":"A"}],"edges":[],"hyperedges":[]}'

what content[0].get("text","{}") yields:
'{}'
```

`stopReason` is `end_turn` — the model completed normally. The truncation diagnosis is an artifact of reading the wrong block.

Extraction log for the same corpus (48 documents), before:

```
[graphify] bedrock returned a hollow response; treating as truncation so adaptive retry can bisect the chunk.
[graphify] chunk of 29 truncated at depth 0, splitting into halves of 14 and 15
[graphify] chunk of 4 still truncated at recursion depth 3 (max 3) — partial result kept (not cached as complete)
[graphify] WARNING: 17/48 dispatched file(s) produced no nodes and are absent from the graph
```

A distinguishing detail: raising `GRAPHIFY_MAX_OUTPUT_TOKENS` changes nothing, because output length was never the constraint.

## Suggested fix

Select the first content block that actually carries non-empty text rather than assuming its position, at both call sites. This keys on the block's shape, not on a model name, so it holds for any model and for block types added to Converse later. A response whose first block is already text — every non-reasoning model today — is unaffected.

## Result after the change

Same corpus and command, only the block selection changed:

| | before | after |
|---|---|---|
| `hollow response` lines | on essentially every call | none |
| bisection to `recursion depth 3 (max 3)` | repeatedly | none |
| files reported as producing no nodes | 17 of 48 | none |
| output tokens | 217,538 | 53,274 |
| estimated cost | $4.09 | $1.44 |
| graph | 2,080 nodes / 4,516 edges | 2,126 nodes / 4,580 edges |

The cost and output-token drop is the wasted bisection disappearing. Document-type nodes rose from 43 to 66 and concept nodes from 98 to 130, so the additional coverage is real rather than a re-labelling.

## Not a duplicate

- #1621 is DeepSeek running with thinking enabled and leaking chain-of-thought prose *into the JSON text* on the OpenAI-compatible path — a content-contamination problem where the text is read but polluted. Here the text block is never read at all, and the payload itself is valid.
- #1192 concerns the `reasoning_effort` / thinking request parameters conflicting at the API level, not response parsing.
- #1758 is the Claude Code subagent output-token ceiling in the skill-driven chunking path, a different dispatch mechanism.
- #1110 wired images *into* the Converse request; this is about reading the response.
- #1372 is an output-token cap producing genuinely truncated JSON; in that case `stopReason` reflects the limit and raising the cap helps, neither of which holds here.

zhiyanliu:
## Summary

Both bedrock call sites read only the first content block of a Converse response. Converse returns `output.message.content` as a **list** and does not promise a text block is first — reasoning-capable models emit `reasoningContent` ahead of the answer, and `toolUse` or future block types can precede it too. For those models the hard-coded index returned the `.get("text", …)` default on every call, so `_parse_llm_json` saw an empty object, `_response_is_hollow` reported a hollow result, `finish_reason` was rewritten to `length`, and the adaptive retry bisected the chunk.

Splitting cannot converge, because the position assumption fails identically at every chunk size. `stopReason` on those responses is `end_turn` — the model answered correctly and was billed for it.

Fixes #2287.

## What changed

Selection now keys on the block's shape rather than its position, via one helper used by `_call_bedrock` (primary extraction) and the `backend == "bedrock"` branch of `_call_llm` (secondary dispatch). A response whose first block is already text — every non-reasoning model today — is unaffected.

The helper is deliberately model-agnostic: it asks whether a block carries non-empty text, never which model produced it, so it also covers block types Converse may add later.

## Verification

Probing Converse directly with a reasoning model shows the shape the old code could not read:

```
stopReason: end_turn
num content blocks: 2
  block[0] keys=['reasoningContent']
  block[1] keys=['text']
content[0].get("text","{}") -> '{}'
```

Re-running extraction on a 48-document corpus, changing only the block selection:

| | before | after |
|---|---|---|
| `hollow response` lines | on essentially every call | none |
| bisection to `recursion depth 3 (max 3)` | repeatedly | none |
| files reported as producing no nodes | 17 of 48 | none |
| output tokens | 217,538 | 53,274 |
| graph | 2,080 nodes / 4,516 edges | 2,126 nodes / 4,580 edges |

Document-type nodes rose from 43 to 66 and concept nodes from 98 to 130, so the extra coverage is real rather than a re-labelling. The output-token drop is the wasted bisection disappearing.

## Tests

Added coverage for the block-selection contract: a single text block behaves exactly as before; a leading `reasoningContent`, `toolUse`, unknown future block type, or whitespace-only text block is skipped in favour of the real one; an absent text block, empty list, non-list content, and missing `output`/`message` all fall back to the caller's default; non-dict blocks and non-string `text` values are tolerated without raising. One end-to-end test drives `_call_bedrock` with a reasoning-model response and asserts the result is neither empty nor reclassified as truncation — it fails if the call sites are reverted to indexing position 0.

`pytest tests/` shows no new failures against `v8` (the pre-existing `test_security` / `test_ollama_retry_cap` / `test_labeling` failures in my environment are unrelated — local DNS and network sandboxing — and are identical with and without this change). `ruff check` is clean on both touched files.

Happy to reshape the helper or move it next to the other bedrock helpers if you'd prefer a different placement.

### DIFF
--- a/graphify/llm.py
+++ b/graphify/llm.py
@@ -1044,6 +1044,30 @@ def _parse_llm_json(raw: str) -> dict:
     return {"nodes": [], "edges": [], "hyperedges": []}
 
 
+def _bedrock_response_text(resp: dict, default: str = "") -> str:
+    """Return the first Converse content block that carries text.
+
+    Converse returns ``output.message.content`` as a list of blocks, and the
+    API does not promise a text block is first: reasoning-capable models emit a
+    ``reasoningContent`` block ahead of the answer, and ``toolUse`` or future
+    block types can precede it too. Indexing position 0 therefore yields no text
+    at all for those models, which reads downstream as a hollow response, gets
+    reclassified as truncation, and sends the chunk into bisection that cannot
+    converge. Select on the block's shape instead of its position so this holds
+    for any model; a response whose first block is already text is unaffected.
+    """
+    content = resp.get("output", {}).get("message", {}).get("content", [])
+    if not isinstance(content, list):
+        return default
+    for block in content:
+        if not isinstance(block, dict):
+            continue
+        text = block.get("text")
+        if isinstance(text, str) and text.strip():
+            return text
+    return default
+
+
 def _response_is_hollow(raw_content: str | None, parsed: dict) -> bool:
     """Detect a successful HTTP response that yielded no usable extraction.
 
@@ -1632,7 +1656,7 @@ def _call_bedrock(model: str, user_message: str, max_tokens: int = 8192, *, deep
         msg = exc.response["Error"]["Message"]
         raise RuntimeError(f"Bedrock API error ({code}): {msg}") from exc
 
-    text = resp.get("output", {}).get("message", {}).get("content", [{}])[0].get("text", "{}")
+    text = _bedrock_response_text(resp, default="{}")
     result = _parse_llm_json(text)
     usage = resp.get("usage", {})
     result["input_tokens"] = usage.get("inputTokens", 0)
@@ -2579,7 +2603,7 @@ def _call_llm(
         bu = resp.get("usage") or {}
         if bu:
             _rec(bu.get("inputTokens", 0), bu.get("outputTokens", 0))
-        return resp.get("output", {}).get("message", {}).get("content", [{}])[0].get("text", "")
+        return _bedrock_response_text(resp, default="")
 
     if backend == "azure":
         endpoint = os.environ.get("AZURE_OPENAI_ENDPOINT", "").strip()
--- a/tests/test_image_vision.py
+++ b/tests/test_image_vision.py
@@ -330,6 +330,85 @@ def test_call_bedrock_sends_raw_image_bytes(tmp_path, monkeypatch):
     assert img_block["image"]["source"]["bytes"] == _PNG_BYTES
 
 
+# ── Converse content-block selection ─────────────────────────────────────────
+# Converse returns a LIST of blocks and does not promise text is first: a
+# reasoning-capable model emits reasoningContent ahead of the answer. Selection
+# must key on block shape, not position, or those models yield no text at all.
+
+def _bedrock_resp(blocks: list) -> dict:
+    return {"output": {"message": {"content": blocks}}}
+
+
+def test_bedrock_response_text_single_text_block_unchanged():
+    resp = _bedrock_resp([{"text": _NODE_JSON}])
+    assert llm._bedrock_response_text(resp) == _NODE_JSON
+
+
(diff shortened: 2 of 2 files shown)


zhiyanliu:
Noted on the coupling advisory — flagging that I read it and am leaving the shape as-is, so a reviewer doesn't have to wonder.

The `+1 callee` on `_call_bedrock` and `_call_llm` is the new helper itself. Before this change the same expression was inlined verbatim at both sites, differing only in the default (`"{}"` vs `""`), so extracting it trades one duplicated expression for one shared function. The metric reads that as added coupling, but the alternative — fixing the block selection twice in place — leaves the duplication and the next Converse-shape change has to be made in two places again.

Inlining the loop at both sites would keep the callee counts flat and I'd rather not: that reintroduces exactly the duplication that let the two sites drift in the first place (one already defaulted differently). Happy to inline it if you'd prefer to hold the counts flat, or to move the helper next to the other bedrock helpers if placement is the concern.

safishamsi:
Thanks @zhiyanliu. Shipped in v0.9.30 (cherry-picked to v8). Closed-unmerged here, but it's in the release: https://github.com/Graphify-Labs/graphify/releases/tag/v0.9.30

PR #2290, opened 2026-07-29
chore(agents): refresh the dogfood AGENTS.md and pin it to the packaged block

PathGao:
## Summary

The repo's own `AGENTS.md` carried an older revision of the always-on block than what `graphify install` writes for users from `graphify/always_on/agents-md.md`: it still told agents to read `GRAPH_REPORT.md` first and to navigate `wiki/` unconditionally, while the packaged block moved to query-first rules and the "if wiki exists" soft-guard. (The unconditional wiki wording is the same surface #2006 flags.)

Two changes:

1. **Refresh**: `AGENTS.md` is regenerated by applying the real splice (`_replace_or_append_section` + `_always_on("agents-md")`) — byte-for-byte what an install would produce.
2. **Pin**: a new test runs that same splice against the checked-in file and requires a no-op, so any future edit to the packaged block forces the dogfood copy along in the same PR. Verified it fails against the pre-refresh file and passes after.

No behavior change for users — the packaged block is untouched; only the repo's own copy catches up.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1,8 +1,12 @@
 ## graphify
 
-This project has a graphify knowledge graph at graphify-out/.
+This project has a knowledge graph at graphify-out/ with god nodes, community structure, and cross-file relationships.
+
+When the user types `/graphify`, use the installed graphify skill or instructions before doing anything else.
 
 Rules:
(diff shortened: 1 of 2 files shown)


PathGao:
Closing — I'm no longer working on this project, so I won't be following up on this PR. Feel free to pick up anything useful from it.

Issue #726, opened 2026-05-05
Surprising Connections still shows cross-language false positives (INFERRED edges between unrelated source files)


slideshow-dingo:
## Problem

After the fix for issue #630 (filter cross-language INFERRED calls), surprising connections still shows clearly incorrect edges. Example from our graph:

```
- `update()` --calls--> `string()`  [INFERRED]
  skills/openclaw-acp/src/commands/profile.ts → claude-code-sourcemap/cli.mjs
- `balance()` --calls--> `string()`  [INFERRED]
  skills/openclaw-acp/src/commands/wallet.ts → claude-code-sourcemap/cli.mjs
- `formatPrice()` --calls--> `string()`  [INFERRED]
  skills/openclaw-acp/src/commands/browse.ts → claude-code-sourcemap/cli.mjs
```

The source files (profile.ts, wallet.ts, browse.ts) are in `skills/openclaw-acp/src/commands/` but the target is `claude-code-sourcemap/cli.mjs` — a completely unrelated repository. This is not a "surprising connection" — it is a noise from semantic INFERRED edge generation picking up the word `string` (a TypeScript type annotation in one file, a JavaScript type in another) and incorrectly inferring a cross-repo call relationship.

## Expected behavior

Surprising connections should filter out INFERRED edges where:
1. Source and target are from different Git repositories (different remote origins)
2. Source and target have no shared path prefix (e.g. both under `skills/openclaw-acp/`)

## Environment

graphifyy 0.7.5, Python 3.11, 10,006 node graph

otyeung:
Opened #916 which is a partial fix specifically for the `String()` / `Number()` / `Boolean()` god-node pattern in JS/TS codebases.

Reproduced this exact symptom on a 3848-node Next.js + TypeScript codebase: `String()` accumulated 134 spurious edges across 30+ communities (`docker/opencli/...` → `src/components/data-table/...`, identical to the report here). After the patch: 1 edge (file containment), zero cross-community noise. Total spurious edges removed: 286.

The PR mirrors the architecture of merged #908 — name-blocklist filtering for ambiguous cross-corpus names. Blocklist is 89 entries spanning JS/TS ECMAScript builtins, Browser/Node globals, and Python builtins.

This does **not** address the broader cross-repository filtering you mentioned (different Git origins) — that would need a separate change. But it removes the loudest source of cross-language false positives in JS/TS codebases.

All 122 existing tests pass.

safishamsi:
[...]

PR #2296, opened 2026-07-29
fix(go): filter Go predeclared functions from call-target resolution

PathGao:
## Summary

`_LANGUAGE_BUILTIN_GLOBALS` covers JS/TS, Python and Swift (#726, #2147) but never covered Go — while `graphify/extractors/go.py:358` already consults that table when resolving a callee. Since the Go resolver looks the callee up by **bare name**, an unexported method that happens to share a builtin's name absorbs every builtin call in the repository. This adds Go's predeclared functions to the table; no resolver logic changes.

## Evidence (real corpus, 8.9k nodes / 26k edges, Go + TypeScript)

The bug was found while graphing a real project. A single method —

```go
// internal/web/service/metric_history.go:133
func (h *metricHistory) append(metric string, t time.Time, v float64) { ... }
```

— collected **330 phantom inbound `calls` edges** from that project's ordinary `append(slice, x)` calls, spread across the whole repo. The knock-on effect was worse than the noise: it invented **twelve `database` → `service` edges**, i.e. a reported layering violation that does not exist in the source. That is exactly the phantom-bridge shape #726/#2147 fixed for other languages.

| | before | after |
|---|---|---|
| edges into builtin-named Go nodes | 333 | 8 |
| of those, **cross-file (phantom)** | 325 | **0** |
| `database` → `service` false violations | 12 | **0** |
| user's `append` method node | present | present (unchanged) |

The 8 remaining edges are all same-file and legitimate: `metricHistory --method--> .append()`, plus four genuine same-file calls to that method.

## Scope decision: functions yes, types no

Only the 14 predeclared **functions** are added (`append`, `cap`, `clear`, `close`, `complex`, `copy`, `delete`, `imag`, `make`, `new`, `panic`, `println`, `real`, `recover` — `len`, `min`, `max`, `print` were already in the table via other languages).

Predeclared **types** (`string`, `int64`, `byte`, `error`, …) are deliberately left out even though Go conversions are call-shaped: they produced **zero** phantom edges on the corpus, and listing them would suppress genuine constructor-like calls. Evidence-driven minimum, not a blanket sweep.

## Trade-off

A genuine cross-file call to a user function named exactly like a builtin is now skipped. That is the same trade-off #2147 accepted for `Data`/`String`, and the asymmetry is stark here: builtin `append` appears hundreds of times in any Go project, a package-level function named `append` cannot exist without shadowing the builtin.

## Tests

`tests/test_go_builtin_call_targets.py`, three cases:

1. builtin `append()` in one file does not bind to a same-named method in another — **fails without the fix, passes with it** (verified by reverting only `base.py`);
2. the user's `append` method still exists as a node (the filter must not delete symbols);
3. a genuine package-level cross-file call still resolves (guard is a no-op for real symbols).

Case 3 uses a package-level call on purpose: the Go resolver intentionally skips receiver method calls for lack of import evidence, so that shape would prove nothing about this filter.

Full suite: no new failures (the pre-existing reds are missing optional deps — terraform grammar, skillgen/ollama network — red on `v8` too).

### DIFF
--- a/graphify/extract.py
+++ b/graphify/extract.py
@@ -43,7 +43,7 @@ from graphify.extractors.dart import extract_dart  # noqa: F401
 from graphify.extractors.dm import extract_dm, extract_dmf, extract_dmi, extract_dmm  # noqa: F401
 from graphify.extractors.elixir import extract_elixir  # noqa: F401
 from graphify.extractors.fortran import _cpp_preprocess, extract_fortran  # noqa: F401
-from graphify.extractors.go import extract_go  # noqa: F401
+from graphify.extractors.go import _GO_PREDECLARED_FUNCS, extract_go  # noqa: F401
 from graphify.extractors.json_config import extract_json  # noqa: F401
 from graphify.extractors.markdown import extract_markdown  # noqa: F401
 from graphify.extractors.pascal_forms import extract_delphi_form, extract_lazarus_form  # noqa: F401
@@ -5321,6 +5321,13 @@ def extract(
         # to external commands that merely share a name with a function elsewhere
         # in the corpus — exactly what #2141 must not do.
         if rc.get("language") == "bash":
+            continue
+        # A Go predeclared function is never a cross-file call: the extractor
+        # already drops bare `append(s, x)` (extractors/go.py), so this is the
+        # backstop for Go raw_calls minted on any other path. Language-gated
+        # rather than folded into _LANGUAGE_BUILTIN_GLOBALS because `new`,
+        # `close` and `delete` are ordinary method names elsewhere (#2296).
+        if rc.get("language") == "go" and callee in _GO_PREDECLARED_FUNCS:
             continue
         # Exact-case match first (case is semantic). Fold only when the CALLING
         # file's language is case-insensitive, and only against the folded index of
--- a/graphify/extractors/base.py
+++ b/graphify/extractors/base.py
@@ -48,6 +48,16 @@ _LANGUAGE_BUILTIN_GLOBALS: frozenset[str] = frozenset({
     "NSObject", "NSString", "NSError", "NSLock", "NSAttributedString",
     "DispatchQueue", "DispatchGroup", "OperationQueue", "RunLoop",
     "View", "Color", "Font",
+    # Go predeclared functions. The Go resolver looks a callee up by bare name,
+    # so an unexported method that happens to share a builtin's name (e.g.
+    # `func (h *history) append(...)`) absorbs every builtin call in the repo:
+    # on a 8.9k-node Go codebase one such method collected 330 phantom
+    # inbound edges, inventing a database -> service layering violation.
+    # Builtin *types* are deliberately not listed: Go conversions are
+    # call-shaped too, but they produced no phantom edges on that corpus, and
+    # listing them would suppress genuine constructor-like calls.
+    "append", "cap", "clear", "close", "complex", "copy", "delete", "imag",
+    "make", "new", "panic", "println", "real", "recover",
 })
 
 
--- a/graphify/extractors/base.py
+++ b/graphify/extractors/base.py
@@ -48,16 +48,6 @@ _LANGUAGE_BUILTIN_GLOBALS: frozenset[str] = frozenset({
     "NSObject", "NSString", "NSError", "NSLock", "NSAttributedString",
     "DispatchQueue", "DispatchGroup", "OperationQueue", "RunLoop",
     "View", "Color", "Font",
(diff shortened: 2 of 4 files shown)


safishamsi (changes requested):
Thanks @PathGao — the diagnosis and test scaffolding are solid (bare `append(x)` phantom-binding cross-file is a real bug). But the set you added to is `_LANGUAGE_BUILTIN_GLOBALS`, which is shared across ~11 languages through the engine and the cross-file pass — adding `append`/`make`/`copy`/`new`/`close`/`delete` changes resolution for all of them, and it wraps the in-file EXTRACTED match too. Concretely it kills every in-file Rust `Type::new()` edge (`new` normalizes to the same token; Rust deliberately keeps its own name blocklist in the cross-file branch only), and drops genuine same-file Go `h.append(v)` selector calls. Please move this to a Go-local `_GO_PREDECLARED_FUNCS` set applied only when the callee node type is a bare identifier (mirroring Rust's language-local pattern), and gate Go `raw_calls` in the shared pass (raw_calls already carry `language`, like the bash gate). Keep the tests — happy to re-review once it's Go-scoped.

PathGao:
Thanks @safishamsi — you were right on both counts, and I reproduced them before changing anything. Pushed `60f0500`.

**Confirmed regressions.** Against `upstream/v8` both edges exist; on the previous revision both are gone:

| case | v8 | previous rev |
|---|---|---|
| in-file Rust `build()` → `Widget::new(3)` | 1 edge | 0 |
| in-file Go `record()` → `h.append(v)` | 1 edge | 0 |

The mechanism is exactly as you described: the `if callee_name and callee_name not in _LANGUAGE_BUILTIN_GLOBALS` guard wraps both the in-file EXTRACTED branch and the `elif` that feeds `raw_calls`, so a name in the shared set is unreachable in either. Rust's `scoped_identifier` handler takes the `name` field, which is the bare token `new`.

**What changed.**

1. `_GO_PREDECLARED_FUNCS` now lives in `extractors/go.py` next to `_GO_PREDECLARED_TYPES`, mirroring `_RUST_TRAIT_METHOD_BLOCKLIST`. The Go block is out of `_LANGUAGE_BUILTIN_GLOBALS` — `base.py` is back to its v8 content.
2. It fires only when `func_node.type == "identifier"`. `selector_expression` callees (`h.append(v)`, `pkg.Delete(x)`) never reach it. The check sits before both branches, because the in-file branch needs it too: a sibling function in the *same* file as the shadowing method binds through `label_to_nid` without ever producing a `raw_call`, so a cross-file-only gate would leave that edge behind.
3. Go `raw_calls` now carry `"language": "go"`, and the shared pass gates on it right after the bash gate — a backstop for Go `raw_calls` minted on any other path.

**One judgement call to flag:** the set is now the Go spec's predeclared function list *in full*, so it picks up `len`, `max`, `min`, `print` that the previous revision omitted. Being Go-local and bare-identifier-only is what makes that safe — those four carry the same shadowing hazard as `append`, and a spec boundary seemed easier to reason about than a hand-picked subset. Happy to trim it back to the original ten if you'd rather keep the delta minimal.

**Remeasured** on the same 466-file Go corpus (3x-ui), counting inbound edges to nodes whose label is a predeclared name:

```
upstream v8    16904 edges    334 inbound to `append`
previous rev   16571 edges      1 inbound   ← 332 phantom gone, genuine call gone too
this rev       16572 edges      2 inbound   ← 332 phantom gone, genuine call kept
```

That restored edge is the one your review predicted: `systemMetrics.append(metric, ...)` at `internal/web/service/metric_history.go:366` → `(*metricHistory).append`. The other surviving edge is the struct→method declaration edge.

**Tests.** The original three are unchanged. Three more added:

- `test_builtin_append_does_not_bind_in_file` — covers the branch a cross-file-only gate would miss.
- `test_go_selector_call_to_shadowing_method_survives` — `h.append(v)` must resolve.
- `test_rust_in_file_type_new_edge_survives` — the cross-language guard.

The last two fail on the previous revision and pass on this one. Full suite is green locally apart from 13 pre-existing failures from optional deps missing in my venv (`tree-sitter-hcl`, `openai`); they fail identically on the unmodified branch. `ruff check` clean.

safishamsi:
Thanks @PathGao. Shipped in v0.9.31 (cherry-picked to v8) — the rework to a Go-local set + bare-identifier gate + raw_calls gate is exactly right (Rust ::new() in-file edges guarded). The broader _LANGUAGE_BUILTIN_GLOBALS design concern you raised in #2313 stays open as the tracked follow-up. Closed-unmerged here, but it's in the release: https://github.com/Graphify-Labs/graphify/releases/tag/v0.9.31

Issue #2211, opened 2026-07-26
Incremental rebuild drops cross-file references edges between Markdown docs; GRAPHIFY_FORCE does not help in the hook (follow-up to #1018)


Jolex56:
Summary

An incremental rebuild re-extracts a changed Markdown file but does not re-establish that file's cross-file references edges to unchanged neighbour documents. The links are still literally present in the text — only the graph loses them. graphify update . --force restores all of them, so the extraction itself is fine; the incremental merge is what drops them.

Because the post-commit hook rebuilds incrementally after every commit that touches a graph-tracked file — including .md files — the doc half of the graph erodes a little with every documentation commit, silently.

Measurement (graphifyy 0.9.26)

A commit editing four Markdown files (CLAUDE.md, docs/FEATURE-INVENTORY.md, docs/features/overlay.md, docs/features/configuration.md) in a 44-file repo:

before:  408 nodes / 693 edges
after:   409 nodes / 676 edges     <- 18 edges lost, 1 node gained

All 18 lost edges are references edges whose source is one of the four edited files, and whose target is an unchanged document:

CLAUDE.md            --references--> diagnostics.md, input.md, output.md,
                                     pythonsidecar.md, recognition.md, setup.md   (6)
FEATURE-INVENTORY.md --references--> diagnostics.md, input.md, output.md,
                                     recognition.md, setup.md, history.md         (6)
overlay.md           --references--> diagnostics.md, output.md, pythonsidecar.md,
                                     recognition.md, setup.md                     (5)
configuration.md     --references--> output.md                                    (1)

No code node and no edge outside those four source files was affected.

The links were never removed from the text. Counted with grep in the post-rebuild working tree:

CLAUDE.md            -> docs/features/setup.md         10 occurrences
CLAUDE.md            -> docs/features/diagnostics.md    2 occurrences
FEATURE-INVENTORY.md -> features/overlay.md            28 occurrences
FEATURE-INVENTORY.md -> features/recognition.md        16 occurrences
(etc. — every dropped target still linked)

Running graphify update . --force immediately afterwards, with no further file changes:

409 nodes / 695 edges   -> all 18 edges restored (plus 2 new ones for content added in the same commit)

No cost.json was written, so this is the plain AST path — the edges are structurally extractable, they are simply not re-linked on the incremental path.

GRAPHIFY_FORCE=1 does not mitigate this in the hook

[...]

Issue #2169, opened 2026-07-24
Incremental --no-cluster extraction overwrites the full graph with the changed-file subset


kitsupanic:
## Summary

After a successful full `graphify extract`, editing one code file and rerunning `graphify extract . --code-only --no-cluster` overwrites `graphify-out/graph.json` with only the changed file’s incremental extraction.

Unchanged nodes and edges disappear, and retained edges can point to absent endpoint nodes. `--force` restores the full graph.

This reproduces on upstream `v8` (`2fa6cd3`) and PR #2134 commit `36ecb25`. The PR author independently confirmed the same behavior on an unrelated 789-file TypeScript/Next.js corpus: https://github.com/Graphify-Labs/graphify/pull/2134#issuecomment-5074078028

## Environment

- `graphifyy 0.9.25`
- macOS
- TypeScript / TSX
- code-only extraction
- reproduced with and without multigraph; the minimal case below uses the default simple/raw `--no-cluster` path

## Minimal reproduction

```text
.
├── app/add/scan.tsx
└── src/components/ScanScreen.tsx
```

```tsx
// app/add/scan.tsx
import { ScanScreen } from '../../src/components/ScanScreen';

export default ScanScreen;
```

```tsx
// src/components/ScanScreen.tsx
export function ScanScreen() {
  return null;
}
```

```bash
graphify extract . --code-only --no-cluster
# 3 nodes, 4 edges:
# imports_from, imports, contains, re_exports

printf '\n// harmless edit\n' >> app/add/scan.tsx

graphify extract . --code-only --no-cluster
# incremental scan: 1 changed, 1 unchanged
# 1 node, 3 edges

graphify extract . --code-only --no-cluster --force
# restored: 3 nodes, 4 edges
```

## Observed result

After the incremental run:

- only `app_add_scan` remains;
- the unchanged `ScanScreen.tsx` file/function nodes are gone;
- its `contains` edge is gone;
- the three retained import/re-export edges target absent, absolute-path-shaped IDs.

Representative missing targets:

```text
private_tmp_..._src_components_scanscreen_tsx
private_tmp_..._src_components_scanscreen_scanscreen
```

The output is therefore both incomplete and structurally invalid.

## Larger-corpus confirmation

On an Expo/React Native TypeScript corpus:

- clustered simple: 1,143 → 1,140 edges after editing `app/add/scan.tsx`;
- clustered edit of another TSX file: 1,143 → 1,134 edges;
[...]

PR #2301, opened 2026-07-29
fix: resolve bare in-repo Python module imports

### LINKED ISSUE #2280: Python AST import edges to in-repo modules emit bare module names, silently dropping real dependencies
**Version:** first observed on graphifyy 0.9.17, re-verified still present on 0.9.29
**Platform:** Windows 11, Python 3.11

## Summary

The Python AST extractor emits import edges whose endpoint is the bare module name. For external modules (`os`, `math`, `yaml`) that is correct: they were never nodes, and the edge is dropped at build time. But for modules inside the scanned repo, the node exists under its fully-qualified id, the bare-name endpoint matches nothing, and the edge is silently dropped. A real intra-repo dependency vanishes from the graph.

## Reproduction

A repo with `ha/pyscript/solar_coordinator.py` importing sibling modules:

```python
import solar_allocator
import ev_actuator
import hws_intent
```

Extraction produces nodes like `ha_pyscript_modules_solar_allocator`, and edges whose targets are `solar_allocator` / `ev_actuator` / `hws_intent`. Those targets match no node id, so `build_from_json` drops all of them. Re-tested on 0.9.29: all three bare endpoints still appear.

## Measured impact (clean extraction, ~3.1k-node graph)

49 distinct dangling endpoint names total:
- 25 external (stdlib/third-party), which is correct behaviour, not a defect
- 4 in-repo module names, silently losing 7 real edges. The importing file's own module structure is missing from the graph
- 20 from a separate JS id defect (reported separately)

## Suggested fix (validated locally)

We wrote a post-pass resolver and learned two things the hard way:

1. Matching by id suffix alone fabricates dependencies. Our first attempt bound `solar_allocator` to an unrelated node whose id happened to end in `_solar_allocator`. The reliable predicate is: candidate node's `Path(source_file).stem == module_name`.
2. Node ids legitimately repeat across extraction passes (AST and semantic reuse the same id with different `source_file` values), so the index must be id to *list* of nodes, qualifying if any entry has the matching stem. A dict keyed by id keeps the last writer and can shadow the real module node with a doc-mention node.

Exactly one qualifying candidate wins; zero or several stays unresolved. Conservative on purpose: a fabricated edge is worse than a missing one.

Related: #949 (same-basename import-target ambiguity in `update`), #2191 (diagnostics counting external refs as dangling; splitting the dangling count into external vs in-repo would have surfaced this class immediately).

hopstreax:
## Summary

Fixes #2280 by resolving bare in-repo Python module imports to their canonical file nodes during graph normalization.

Python AST extraction can emit bare module names (for example, `import solar_allocator`) as import edge targets. Since `build_from_json()` only resolves canonical node IDs, these bare targets remained unresolved, causing valid dependency edges to be silently dropped even when the corresponding Python file existed in the repository.

This change extends `_repoint_python_package_imports()` to perform a conservative second-pass rewrite for unresolved bare Python imports. It builds an index of canonical Python file nodes keyed by module stem and rewrites a bare import only when it uniquely identifies a single in-repo Python file. Existing package-qualified alias resolution continues to take precedence, while ambiguous matches and external imports are intentionally left unresolved to avoid introducing incorrect dependency edges. The rewrite is applied only to Python import edges, preventing cross-language rewrites.

## Tests

Added regression tests covering:

* Resolution of bare in-repo imports to canonical file nodes.
* External imports remaining unresolved and therefore not creating dependency edges.
* Ambiguous bare imports (multiple files with the same module stem) remaining unresolved to prevent false-positive rewrites.
* Mixed import syntaxes (bare, package-qualified, relative, and external) continuing to behave correctly together.
* Existing regression coverage for scan-root-independent Python import resolution and protection against cross-language alias rewrites remains intact.

This approach preserves existing behavior for package-qualified imports while restoring missing dependency edges for legitimate bare in-repo Python imports without introducing ambiguity or cross-language regressions.

### DIFF
--- a/graphify/extract.py
+++ b/graphify/extract.py
@@ -239,8 +239,36 @@ def _repoint_python_package_imports(paths, all_nodes, all_edges, root) -> None:
         for a, fs in alias_to_files.items()
         if len(fs) == 1 and a not in node_ids
     }
-    if not alias_map:
+
+    # Index canonical Python file nodes by stem for bare import resolution (#2280).
+    # Maps module stem id -> set of canonical file node IDs.
+    stem_to_file_nodes: dict[str, set[str]] = {}
+    for n in all_nodes:
+        if not isinstance(n, dict) or n.get("file_type") != "code":
+            continue
+        if n.get("_callable") or n.get("_callable_class"):
+            continue
+        sf = str(n.get("source_file", ""))
+        if not sf.lower().endswith((".py", ".pyi")):
+            continue
+        label = n.get("label", "")
+        try:
+            sf_name = Path(sf).name
+        except Exception:
+            continue
+        if label != sf_name:
+            continue
+        nid = n.get("id")
+        if not nid:
+            continue
+        stem = Path(sf).stem
+        if stem and stem != "__init__":
+            stem_id = _make_id(stem)
+            stem_to_file_nodes.setdefault(stem_id, set()).add(nid)
+
+    if not alias_map and not stem_to_file_nodes:
         return
+
     for e in all_edges:
         # Only repoint edges emitted from a Python file: a non-Python import edge
         # (e.g. C# `using Pkg.Mod;`, Java/Go dotted imports) can have a dangling
@@ -252,8 +280,14 @@ def _repoint_python_package_imports(paths, all_nodes, all_edges, root) -> None:
             and str(e.get("source_file", "")).lower().endswith((".py", ".pyi"))
         ):
             tgt = e.get("target")
-            if tgt in alias_map:
+            if not tgt:
+                continue
+            if alias_map and tgt in alias_map:
                 e["target"] = alias_map[tgt]
+            elif stem_to_file_nodes and tgt in stem_to_file_nodes:
+                candidates = stem_to_file_nodes[tgt]
+                if len(candidates) == 1:
+                    e["target"] = next(iter(candidates))
 
 
 SEMANTIC_RELATIONS = frozenset({
--- a/graphify/extract.py
+++ b/graphify/extract.py
@@ -240,33 +240,7 @@ def _repoint_python_package_imports(paths, all_nodes, all_edges, root) -> None:
         if len(fs) == 1 and a not in node_ids
     }
 
-    # Index canonical Python file nodes by stem for bare import resolution (#2280).
-    # Maps module stem id -> set of canonical file node IDs.
-    stem_to_file_nodes: dict[str, set[str]] = {}
-    for n in all_nodes:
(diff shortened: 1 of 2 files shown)


safishamsi (changes requested):
Thanks @hopstreax — the gap is real (bare in-repo `import pkg.mod` emits an unresolved bare-module target). But as written the fix fabricates wrong edges, which is the one thing we can't ship. The stem index is corpus-wide with no stdlib or importability gate, so I reproduced this end-to-end on the branch:

```
mypkg/logging.py     # repo's own helper
mypkg/app.py         # `import logging`  (the stdlib)
```
-> `mypkg_app --imports--> mypkg_logging`. Under Python 3 absolute imports, `import logging` inside a package is the stdlib, not `mypkg/logging.py`; any repo with a uniquely-named `logging.py`/`types.py`/`queue.py` gets every stdlib import of that name rewired to it (same for `from logging import ...`). The external-import test (`requests`/`numpy`) doesn't catch it because there's no in-repo stem collision.

Second issue: uniqueness is judged against the current extraction batch (`all_nodes`), so an incremental `update` where only one of two same-stem files is in the batch resolves what a full scan leaves ambiguous — the #2211/#2213 incremental-parity class.

Requested changes: (a) exclude `sys.stdlib_module_names`; (b) gate on importability — resolve only when the candidate is a sibling of the importer's directory or top-level under the scan root (sys.path semantics), which still fixes the #2280 sibling case; and ideally stamp `target_file` (existence-gated, like the relative-import branch) so the #2169 remap keeps incremental parity rather than resolving off the batch. The structure and self-loop guard are otherwise good — happy to re-review once stdlib/importability are gated.

hopstreax:
Thanks for the review! I've addressed the requested changes:

 * Switched bare import resolution to use `_resolve_python_module_path()` instead of filename-stem matching, so it now follows Python's importability rules.
 * Excluded standard library modules from this resolution using `sys.stdlib_module_names`.
 * Added `target_file` to resolved bare-import edges so incremental extraction follows the same canonicalization path as full extraction (#2169).
 * Removed the old stem-based bare import repointing logic from `_repoint_python_package_imports()`, which now remains responsible only for repository-wide package alias rewriting.
 * Updated the regression test to verify that `import utils` resolves to the importable sibling (`src/utils.py`) rather than a non-importable duplicate.
 * Added a clarifying comment documenting why only single-segment imports are resolved during extraction, while dotted imports continue to be handled by `_repoint_python_package_imports()` to preserve repository-wide ambiguity detection.

 I reran the relevant import-resolution test suite after the changes, and all tests are passing.

Issue #2032, opened 2026-07-19
File nodes labelled by basename (index.ts) break name lookup and blind discovery in directory-per-entrypoint repos


grumpypops:
## Summary

In repos that use a directory-per-entry-point convention, many distinct, business-critical files share the same basename, so basename-labelled file nodes collide and cannot be disambiguated by name. This makes `explain "<name>"` and free-text discovery fail for exactly the highest-value paths. The graph is genuinely useful once you already know the target node, but discovering the right node from a question does not work for these files.

## Environment

- Graphify version: [run `graphify --version` and paste]
- Extraction: code-only, local, installed as graphifyy[sql]
- Repo: TypeScript, Supabase Edge Functions

## Repro

Supabase Edge Functions put one entry file per function directory:

    supabase/functions/
      process-order/index.ts
      send-receipt/index.ts
      webhook-handler/index.ts
      verify-signup/index.ts
      ...dozens more

Every one of these is a separate, load-bearing entry point, but they all share the basename index.ts.

Two failures follow:

1. `graphify explain "process-order"` does not resolve, because the file-level node is labelled index.ts, not by its directory or path. The meaningful name (the function/directory) is not a node label you can look up.

2. Free-text `query` (for example "what handles order processing") picks poor seeds for these directory/function names, so blind discovery does not reliably surface the correct entry points.

Net effect: known-node browsing (explain/path/neighbourhood from a node you already found) works well. Discovering the right node from a business question fails for the paths that matter most.

## Suggested fixes

1. Path-aware node labels so basename collisions disambiguate, for example process-order/index.ts or functions/process-order/index.ts rather than bare index.ts.
2. Allow explain/lookup by path or by enclosing directory name, not only by basename.
3. Improve seed selection so directory and function names are candidate seeds for free-text queries.

## Generalises beyond Supabase

The same collision hits any convention with repeated basenames: Next.js page.tsx / route.ts, Rust mod.rs, Python __init__.py. Directory-aware labelling would help all of them.

safishamsi:
Fixed in v0.9.22: file nodes that share a basename now get a shortest-unique directory-qualified label (e.g. `process-order/index.ts`), so `explain`/discovery can resolve them; unique basenames stay bare and node ids/edges are unchanged. https://github.com/Graphify-Labs/graphify/releases/tag/v0.9.22

PR #2305, opened 2026-07-29
fix: explain no longer resolves an ambiguous name to an arbitrary file

0bLoM:
## The bug

`graphify explain "<name>"` answers an ambiguous name by taking `_find_node(G, label)[0]`. `_find_node` ranks matches into `source_exact + exact + prefix + substring`, but never signals that a tie was broken — so when the winning tier holds several nodes from **different files**, the command prints one of them as unqualified fact.

In a monorepo this is the common case, not an edge case: every workspace has its own `MetricsPort` / `index.ts` / `metrics.port.ts`, and they all land in the same `exact` tier separated only by `G.nodes()` iteration order.

### Reproduction

A graph with two `MetricsPort` symbols in different files:

```
$ graphify explain "MetricsPort" --graph graph.json
Node: MetricsPort
  ID:        chat_metrics_port
  Source:    services/chat/src/application/ports/metrics.port.ts
```

Swap only the order of the two nodes in the JSON — nothing else — and the same query answers with the other file, just as confidently:

```
Node: MetricsPort
  ID:        scraping_metrics_port
  Source:    services/scraping/src/application/ports/metrics.port.ts
```

The sibling commands already handle this. On the identical graph:

```
$ graphify affected "MetricsPort" --graph graph.json
No unique node match for MetricsPort

$ graphify path "MetricsPort" "trackScrapeEvent" --graph graph.json
warning: source match was ambiguous (top score 9320.28, runner-up 9320.28)
```

So `affected` declines and `path` warns; only `explain` — and the MCP `get_neighbors` tool, which shares the matcher — silently guessed. A wrong answer that looks authoritative is worse than an error, especially for an agent consuming the output.

This is distinct from #2032: `_disambiguate_file_node_labels` relabels colliding *file* nodes, which does not reach symbol-label collisions.

## The change

- Split `_find_node` into `_find_node_tiers` (matching logic **unchanged**, tiers exposed) plus a thin flattening wrapper, so `_find_node`'s contract and ordering are byte-identical for existing callers.
- Add `find_node_ambiguity(G, label)` — returns one representative per distinct `source_file` when the **winning tier** is split across files, else `[]`.
- CLI `explain` lists the candidates with their node ids and exits 1; MCP `get_neighbors` returns the same listing.

Matches *within* one file (a file node plus its members) remain ordinary precedence and resolve exactly as before.

```
$ graphify explain "MetricsPort" --graph graph.json
Ambiguous: 'MetricsPort' matches 2 nodes in different files.
  services/chat/src/application/ports/metrics.port.ts
    id: chat_metrics_port
  services/scraping/src/application/ports/metrics.port.ts
    id: scraping_metrics_port
Retry with the repo-relative path or the full node id.
```

## Tests

Three added to `tests/test_explain_cli.py`:

- `test_explain_ambiguous_label_lists_every_candidate`
- `test_explain_ambiguous_answer_does_not_depend_on_node_order` — encodes the bug directly: the answer must not change when only node order changes
- `test_explain_matches_within_one_file_are_not_ambiguous` — regression guard

Verified RED/GREEN: the first two **fail** against unpatched `cli.py` (the failure output shows `Node: MetricsPort / ID: chat_metrics_port` being reported as fact) and pass with the change. The third passes both ways by design.

Full suite, same environment, before vs after: **identical failure sets** — nothing broken, nothing incidentally fixed. (There are pre-existing failures in `test_skillgen.py` and `test_terraform.py` in my environment, from a shallow clone with no `origin/v8` to diff against and a missing grammar; they are unrelated and present on a clean checkout too.)

## Note for maintainers

MCP `get_node` (`serve.py`, `_tool_get_node`) has the same defect class via a *third*, independent ad-hoc matcher (`label in (d.get('label') or '').lower()`, then `matches[0]`). I left it out to keep this diff reviewable — happy to fold in a fix if you'd prefer it here.

### DIFF
--- a/graphify/cli.py
+++ b/graphify/cli.py
@@ -1289,7 +1289,7 @@ def dispatch_command(cmd: str) -> None:
         if len(sys.argv) < 3:
             print('Usage: graphify explain "<node>" [--graph path]', file=sys.stderr)
             sys.exit(1)
-        from graphify.serve import _find_node
+        from graphify.serve import _find_node, find_node_ambiguity
         from networkx.readwrite import json_graph
 
         label = sys.argv[2]
@@ -1316,6 +1316,14 @@ def dispatch_command(cmd: str) -> None:
         if not matches:
             print(f"No node matching '{label}' found.")
             sys.exit(0)
+        rivals = find_node_ambiguity(G, label)
+        if rivals:
+            print(f"Ambiguous: '{label}' matches {len(rivals)} nodes in different files.")
+            for rival in rivals:
+                print(f"  {G.nodes[rival].get('source_file') or rival}")
+                print(f"    id: {rival}")
+            print("Retry with the repo-relative path or the full node id.")
+            sys.exit(1)
         nid = matches[0]
         d = G.nodes[nid]
         print(f"Node: {d.get('label', nid)}")
--- a/graphify/serve.py
+++ b/graphify/serve.py
@@ -985,12 +985,15 @@ def _query_graph_text(
     return header + _subgraph_to_text(traversal_graph, nodes, edges, token_budget, seeds=start_nodes)
 
 
-def _find_node(G: nx.Graph, label: str) -> list[str]:
-    """Return node IDs whose label or ID matches the search term (diacritic-insensitive).
+def _find_node_tiers(
+    G: nx.Graph, label: str
+) -> tuple[list[str], list[str], list[str], list[str]]:
+    """Return match tiers in precedence order: (source_exact, exact, prefix, substring).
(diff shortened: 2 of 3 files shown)


safishamsi:
Thanks @0bLoM. Shipped in v0.9.31 (cherry-picked to v8). Follow-up: the MCP get_node matcher has the same arbitrary-pick defect — happy to take the follow-up you offered. Closed-unmerged here, but it's in the release: https://github.com/Graphify-Labs/graphify/releases/tag/v0.9.31

PR #2310, opened 2026-07-29
Persist validated curriculum Graphify skill and installer

rosstimo:
AI had too much permission


rosstimo:
Sorry, AI got away from me.

Issue #2299, opened 2026-07-29
C#: file-scoped receiver typing drops true `calls` edges in unrelated methods (regression in 0.9.27)


JensD-git:
**Regression window:** introduced in **0.9.27** (`#1609` namespace-aware member-call resolution, adapted from `#1620`).
**Last unaffected version:** 0.9.26. Still present in 0.9.29.
**Language:** C#. **Platform:** Windows 11, Python 3.12, `graphifyy` from PyPI, `tree-sitter-c-sharp 0.23.5` (identical parser in both versions tested).

## Summary

`_resolve_csharp_member_calls` types a receiver from a **per-file** `name -> Type` table and poisons the
entry when the same name is bound to conflicting/untypable expressions. Because the table is keyed per
**file** rather than per **method/block scope**, a single binding that the extractor cannot type — e.g.
`var x = list[i];` in one method — removes the `calls` edges for *every* use of that name in the file,
including methods where the receiver is an explicitly typed parameter and C# scoping is unambiguous.

The docstring describes the poisoning as protecting against "a shadowing local of a different type".
That intent is right, but C# shadowing is scoped: two methods that each declare their own local named
`item` do not shadow one another, so no poisoning should occur between them.

## Minimal reproducer

Two files, one namespace, no external dependencies:

`Item.cs`
```csharp
namespace Demo { public class Item { public void Handle() { } } }
```

`Runner.cs`
```csharp
using System.Collections.Generic;

namespace Demo
{
    public class Runner
    {
        public void RunOne(Item item)          // explicitly typed parameter
        {
            item.Handle();                     // EXPECTED: calls Item.Handle
        }

        public void RunIndexed(List<Item> items, int i)
        {
            var item = items[i];               // element type not inferred by the extractor
            item.Handle();                     // EXPECTED: calls Item.Handle
        }
    }
}
```

```
graphify extract ./src --out ./out --code-only
```

| Variant | 0.9.26 `calls` edges | 0.9.29 `calls` edges |
|---|---|---|
| **V1** — only `RunOne` | `RunOne -> Handle` | `RunOne -> Handle` |
| **V2** — `RunOne` + `RunIndexed` (both use the name `item`) | `RunOne -> Handle`, `RunIndexed -> Handle` | **none** |
| **V3** — as V2, but the second local is named `other` | `RunOne -> Handle` | `RunOne -> Handle` |
[...]

PR #2322, opened 2026-07-30
Kotlin: cross-file import + receiver-typed member-call resolution

### LINKED ISSUE #2: v3: Codex and OpenClaw support

### LINKED ISSUE #3: Wroked out examples missing graph.html
The worked out examples are appreciated, but they are missing the `graph.html`s. This file seems like the most valuable to the human, so requesting these examples to be added

sigarone:
## Summary

Kotlin extraction currently produces almost no usable cross-file structure -- measured on a real ~1170-file Kotlin/Android production module:

| | before | after |
|---|---|---|
| cross-file `calls` edges | 2576 | 7543+ |
| ...of which EXTRACTED confidence | 7 | 1600+ (200x+) |
| resolved `imports` edges | 691 | ~2900 explicit + ~34.7k same-package `imports_from` |

This PR bundles five fixes, all discovered while verifying against real production Android/Kotlin code (not just synthetic cases) -- each one closes a gap that made an otherwise-correctly-typed receiver still fail to resolve.

## 1. Import resolution was silently broken

`_import_kotlin`'s edge target never matches a real node id -- it builds `target = _make_id(module_name)` from the bare imported name, but Kotlin symbol ids are file/class-qualified (e.g. `crypto_messagecrypto_messagecrypto`, not `messagecrypto`). `build.py`'s dangling-edge filter silently drops the mismatch. On the measured corpus: 14616 source `import` lines -> 691 surviving edges.

**Fix:** new `graphify/extractors/kotlin.py` -- `_resolve_cross_file_kotlin_imports`, same two-pass strategy as `_resolve_cross_file_java_imports` (global name index, re-parse per file, resolve `import a.b.C`), plus a same-package pass Java doesn't need -- Kotlin requires no `import` for same-package visibility, so every same-package file pair gets one file-to-file `imports_from` edge (not one edge per symbol -- that's the exact granularity `_has_import_evidence`'s module-import check already consumes, ~3x smaller than symbol-level for identical promotion power).

## 2. Kotlin member calls never carried a receiver

The call-walk branch for `tree_sitter_kotlin` sets `is_member_call = True` for `recv.method()` but never captures `recv` into `member_receiver` -- unlike the Swift/Java/C#/C++/ObjC/TS branches right next to it. The shared cross-file pass's `if rc.get("is_member_call"): continue` then drops every Kotlin member call with no receiver-typed resolver to pick them back up.

**Fix:** receiver capture in the call-walk (mirrors Swift's `#1356`) -- only a direct identifier or `this` receiver; a chained receiver (`a.b().c()`) is left untyped rather than guessed, same policy as everywhere else. A Kotlin `type_table` (property declarations -- explicit type or constructor-call/static-factory-call inference -- primary-constructor `val`/`var` parameters, function-body locals) is threaded out as `kotlin_type_table`, and `_resolve_swift_member_calls` is extended to also read it -- its receiver-typing algorithm doesn't care which language produced the table, so no new resolver function, just a second lookup and `.kt`/`.kts` added to the registration's suffix set.

## 3. Companion-object members had no class ownership at all

`companion object { fun getInstance() }` -- the standard Kotlin/Android singleton/factory idiom -- fell through the generic walk's default recurse, which resets `parent_class_nid`. This meant `getInstance()` got extracted with **no `method` edge from the outer class whatsoever**, so even a correctly receiver-typed call to `Foo.getInstance()` could never resolve past the class itself (a `references` edge, not `calls`).

**Fix:** treat `companion_object` as a transparent wrapper -- but recurse into its own **body's children** (mirroring exactly how the `class_types` branch recurses via `_find_body`/`body.children`), not into `companion_object`'s direct children, which include the nested `class_body` node itself and would otherwise still hit the default-recurse reset one level down. (Caught this by testing an isolated companion-object fixture after the receiver-typing fix above resolved everything *except* this pattern -- worth flagging in review since it's a subtle two-level indirection.)

## 4. Constructor-default-value and `by lazy {}` calls were invisible

Two extremely common Kotlin idioms hold their initializer call in a place the existing initializer-walk (which only looks at direct `call_expression`/`navigation_expression` siblings) doesn't see:
- `class Foo(private val bar: Bar = Bar.getInstance())` -- a call as a **primary-constructor default parameter value**.
- `private val bar: Bar by lazy { Bar.getInstance() }` -- a call inside a **property delegate**'s lambda body.

**Fix:** both get queued into `initializer_nodes` the same way Swift's `#1356` property initializers already are -- `walk_calls` naturally recurses through either node on its own, no special unwrapping needed.

## Verification

- Synthetic fixture covering all paths (explicit cross-file import, primary-constructor param typing, explicit-typed property, constructor/static-factory-inferred untyped property, function-body local, companion-object singleton via constructor-default AND via `by lazy`) -- all resolve EXTRACTED/INFERRED as expected, reproduced bit-identically from a from-scratch checkout (clean clone -> patch -> build -> same edges every time).
- Ground-truthed against real production code:
  - `MessageCrypto.encrypt()` (line 140) calling `vault.listKeys()` resolves to the correct target with the correct source line.
  - `BCryptoSecureStorage.getInstance()` (a companion-object singleton) is now correctly `method`-owned by `BCryptoSecureStorage`, and a real cross-file caller (`VoicePrintVault`, via a primary-constructor default-value call) now resolves it as `calls`/EXTRACTED -- this exact call site was completely invisible to the graph before fix #3+#4 combined.
- Zero dangling edges introduced by either new Kotlin pass, checked against the final built graph.
- No cross-language contamination risk: `type_table` is already shared per-file infra (Swift/C++ both write into it, one language per file in practice); the resolver's suffix-gate change only adds `.kt`/`.kts` to when the pass *runs*, doesn't change Swift-only corpora behavior at all (empty `kotlin_type_table` contributes nothing).

## Test plan

- [ ] `pytest` full suite (I ran manual functional checks against a from-scratch checkout + real production code, not the project's test harness -- happy to add/run whatever's expected, let me know the convention)
- [x] Byte-compiles clean (`py_compile` on all three touched files)
- [x] Fresh checkout smoke test (synthetic fixture, all paths, all correct, reproduced identically across two independent clean clones)
- [x] Real-corpus before/after measurement (numbers above)
- [x] Ground-truth spot checks against multiple real call sites, including one that needed fixes #2, #3, and #4 together to resolve

Happy to split into smaller PRs if that's easier to review -- the five pieces share the same `type_table`/`initializer_nodes` infra and were each found by testing the *previous* fix against real code and hitting the next gap, so bundling them together is also how they were verified together.

### DIFF
--- a/graphify/extract.py
+++ b/graphify/extract.py
@@ -45,6 +45,7 @@ from graphify.extractors.elixir import extract_elixir  # noqa: F401
 from graphify.extractors.fortran import _cpp_preprocess, extract_fortran  # noqa: F401
 from graphify.extractors.go import extract_go  # noqa: F401
 from graphify.extractors.json_config import extract_json  # noqa: F401
+from graphify.extractors.kotlin import _resolve_cross_file_kotlin_imports  # kotlin-cross-file
 from graphify.extractors.markdown import extract_markdown  # noqa: F401
 from graphify.extractors.pascal_forms import extract_delphi_form, extract_lazarus_form  # noqa: F401
 from graphify.extractors.powershell import extract_powershell, extract_powershell_manifest  # noqa: F401
@@ -2087,8 +2088,8 @@ def _resolve_swift_member_calls(
     all_nodes: list[dict],
     all_edges: list[dict],
 ) -> None:
-    """Resolve cross-file Swift member calls (``recv.method()``) to the real
-    definition of the receiver's type (#1356).
+    """Resolve cross-file receiver-typed member calls (``recv.method()``) to the
+    real definition of the receiver's type (#1356).
 
     The shared cross-file call pass drops every ``is_member_call`` because a bare
     method name (``update``) collides across the corpus and inflates god-nodes
@@ -2098,7 +2099,17 @@ def _resolve_swift_member_calls(
     exactly one definition. A type-qualified call (``Type.staticMethod()``) is
     EXTRACTED (the type is named explicitly in source); an instance call typed via
     local inference (``obj.method()``) is INFERRED. The shared-pass member-call drop
-    stays intact: this is purely additive and fires only on receiver-typed Swift calls.
+    stays intact: this is purely additive and fires only on receiver-typed calls.
+
+    Despite the name, this pass is language-agnostic once a
+    ``(file path -> {name: TypeName})`` table exists for a file: Kotlin reuses it
+    via its own ``kotlin_type_table`` (same shape, populated by the Kotlin
+    property/class-parameter/local-var branches in engine.py) instead of
+    duplicating this whole resolver -- the receiver-typing logic below doesn't
+    care which language produced the table. The resolver is registered for both
+    ``.swift`` and ``.kt``/``.kts`` suffixes (see ``register_language_resolver``
+    below) so it activates for a Kotlin-only corpus with zero Swift files too.
+    # kotlin-cross-file
 
     Must run after id-disambiguation so node ids and caller_nids are final.
     """
@@ -2107,6 +2118,9 @@ def _resolve_swift_member_calls(
         tt = result.get("swift_type_table")
         if tt and tt.get("path"):
             type_table_by_file[tt["path"]] = tt.get("table", {})
+        ktt = result.get("kotlin_type_table")
+        if ktt and ktt.get("path"):
+            type_table_by_file[ktt["path"]] = ktt.get("table", {})
     if not type_table_by_file:
         return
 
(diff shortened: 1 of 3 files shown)


safishamsi (changes requested):
Thanks @sigarone — this is well above a drive-by; every grammar assumption checks out against the pinned tree-sitter-kotlin 1.1.0, the guards follow the Swift/Java precedents, and the companion-object handling is correct. Three things before it can land:

1. Rebase. It's CONFLICTING against v8 now — the 0.9.31 C# per-method receiver work (#2299) rewrote the engine.py type-table export region. The rebase is mechanical (re-add the kotlin_type_table elif after cpp_type_table), extract.py resolves clean.

2. Per-method typing, not per-file. `_kotlin_local_var_types` pools every method's locals + class properties + ctor params into one per-file dict, last-write-wins. `val conn = DbConnection()` in method A and `val conn = HttpCache()` in method B means B overwrites A and `conn.close()` in A emits a WRONG edge. We moved C# to per-method typing with poison-on-conflict for exactly this in 0.9.31 (#2299). Minimum ask: poison a name on conflicting/untypable rebind (delete the entry) rather than overwrite, so a shadowing local of a different type produces no edge instead of a wrong one. That's the no-wrong-edge bar.

3. Tests. No tests ship with this; repo convention is a per-language suite (see tests/test_csharp_member_calls.py, test_swift_import_resolution.py). Please add test_kotlin_member_calls.py + test_kotlin_import_resolution.py covering explicit import, same-package imports_from, ctor-param/property/local receiver typing, companion-object resolution, ambiguity bail, and a builtin-receiver negative (e.g. Log.d / list.map must not bind to a user symbol).

Follow-ups (don't hold the merge): a Kotlin/Android builtin-receiver set (Log, Intent, Context, Toast, Handler, Integer, System — the capitalized-receiver EXTRACTED path only checks _LANGUAGE_BUILTIN_GLOBALS today); gate the shared member-call resolver by source_file suffix so a Kotlin+Java corpus doesn't cross-process; and a cap on the O(N^2) same-package imports_from edges.

Note: the Kotlin anonymous-object members part (#2347) shipped separately in 0.9.32, so you can drop that slice if present. Happy to re-review once rebased with the per-method poison + tests.

PR #2328, opened 2026-07-30
perf(dedup): remove O(nodes x components) scan from remap construction

stupidprogrammer4:
deduplicate_entities() rebuilt its remap table by re-scanning the whole unique_nodes list once per merged union-find component, costing O(nodes x components) -- roughly 137M membership checks on a 50k-node corpus with 2.5k merged components, about a quarter of dedup wall-clock.

Build an id -> (position, node) index once and slice each component out of it instead: O(members log members) per component.

The index carries the enumeration position, not just the node, because _pick_winner() selects via min(), which returns the first minimum -- so ties (equal chunk-suffix status and equal id length) are resolved by list order. Sorting the group by id instead changed 179 of 2358 survivors on a tie-heavy corpus. Sorting by position reproduces unique_nodes order exactly, leaving survivors and edges byte-identical.

Benchmark, median of 5 runs:

    nodes    before     after   speedup
     5000     2.69s     2.51s     1.07x
    20000    10.43s     9.00s     1.16x
    50000    34.19s    25.07s     1.36x

Isolated remap loop at 50k nodes / 2.5k components: 8.671s -> 0.030s (289x). Merge counts and output node counts are identical at every scale.

This does not address the pre-existing arrival-order dependence of the component structure itself, which originates in the fuzzy-merge loop's uf.find() short-circuit and is out of scope here.

### DIFF
--- a/graphify/dedup.py
+++ b/graphify/dedup.py
@@ -574,10 +574,26 @@ def deduplicate_entities(
     components = uf.components()
     remap: dict[str, str] = {}
 
+    # id -> (position, node), built once. Previously each component re-scanned
+    # the whole unique_nodes list, making remap construction O(nodes x
+    # components) — 31% of dedup wall-clock on a 50k-node corpus.
+    # The position is carried so group_nodes keeps unique_nodes order: _pick_winner
+    # resolves ties (equal chunk-suffix status and equal id length) via min(),
+    # which returns the first minimum, so reordering here would silently change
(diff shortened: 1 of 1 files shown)


safishamsi:
Shipped in v0.9.32: the O(nodes x components) scan is gone from remap construction, with identical results (#2328). Credited in the release notes. Thanks @stupidprogrammer4!

Issue #803, opened 2026-05-10
dedup: KeyError 'source' in deduplicate_entities when merging extraction into existing graph.json


Zottelstrike:
## Summary

`graphify extract` crashes during deduplication when merging newly-extracted nodes/edges into an existing `graph.json`. Reproducible across multiple Ollama models on different hosts.

## Stack trace

```
File "/usr/local/lib/python3.13/site-packages/graphify/dedup.py", line 236, in deduplicate_entities
    e["source"] = remap.get(e["source"], e["source"])
                            ~^^^^^^^^^^
KeyError: 'source'
```

## Reproduction

1. Bootstrap graphify with a small markdown vault → existing `graph.json` has nodes + links (NetworkX-style: each link has a `source` key).
2. Add new markdown files with YAML frontmatter to the vault.
3. Run `graphify extract /vault --backend ollama --model qwen2.5:7b --out /data`.
4. Extraction completes (cache fills under `graphify-out/cache/semantic/`), but the dedup step crashes with the trace above.

## What I verified

- All 22 cached LLM responses under `cache/semantic/` contain `nodes` + `edges` + `hyperedges` arrays. Every `edges[i]` has the keys `[confidence, confidence_score, relation, source, source_file, source_location, target, weight]` — none missing `source`. Verified with:
  ```
  find … cache/semantic -name '*.json' -exec cat {} \; \
    | jq -s '[.[].edges[] | select(has("source") | not)]'
  ```
  → empty array.
- The pre-existing `graph.json` has 38 `links` and all of them have `source` too.
- Reproducible with both `qwen2.5:7b` and `qwen3.6:35b` on separate Ollama hosts (so it's not a model-output shape issue).

## Hypothesis

Somewhere between `combined["edges"]` construction and `deduplicate_entities()` (around `graphify/dedup.py:226–240`), an edge with no `source` key is being included. Two candidates:

- **Hyperedge → edge promotion.** If hyperedges are being flattened into the edges list during merge and the flattening doesn't always emit a `source` key, that would match.
- **NetworkX-format `links` → edge format mismatch.** The pre-existing `graph.json` stores edges under `links` (NetworkX convention) with `source`/`target` as node-IDs. If those are being loaded back into the dedup pipeline as "edges" but with a slightly different shape (e.g. integer IDs vs string keys, or stripped during a normalization step), one of them might lose `source` before the `remap.get` line.

[...]

PR #2337, opened 2026-07-31
fix(build): coerce non-string node ids so a numeric id cannot abort the build

### LINKED ISSUE #2326: dedup._score() crashes with TypeError when LLM emits non-string node id (int)
## Summary

`graphify extract --backend ollama` crashes with `TypeError: expected string or bytes-like object, got 'int'` during the final dedup/build step when the LLM backend emits a node with a non-string (integer) `id` field. `dedup.py`'s `_score()` calls a regex `.search()` directly on `n["id"]` with no type check/coercion.

## Traceback

```
  File "<frozen runpy>", line 198, in _run_module_as_main
  File "<frozen runpy>", line 88, in _run_code
  File "graphify.exe\__main__.py", line 10, in <module>
    sys.exit(main())
  File ".../graphify/__main__.py", line 466, in main
    _run_cli()
  File ".../graphify/__main__.py", line 708, in _run_cli
    dispatch_command(cmd)
  File ".../graphify/cli.py", line 3199, in dispatch_command
    G = _build_merge([merged], ..., root=target)
  File ".../graphify/build.py", line 1023, in build_merge
    G = build(all_chunks, directed=directed, dedup=dedup, dedup_llm_backend=dedup_llm_backend, root=root)
  File ".../graphify/build.py", line 864, in build
    combined["nodes"], combined["edges"] = deduplicate_entities(
        combined["nodes"], combined["edges"], communities={},
        dedup_llm_backend=dedup_llm_backend,
    )
  File ".../graphify/dedup.py", line 487, in deduplicate_entities
    winner = _pick_winner([node, neighbor])
  File ".../graphify/dedup.py", line 561, in _pick_winner
    return min(nodes, key=_score)
  File ".../graphify/dedup.py", line 558, in _score
    has_suffix = bool(_CHUNK_SUFFIX.search(n["id"]))
TypeError: expected string or bytes-like object, got 'int'
```

## Root cause

For at least some image-file chunks, the LLM backend returned a node with a bare integer `id` instead of a string, e.g. (found in the semantic cache):

```json
{
  "id": 10,
  "file_type": "image",
  "source_file": "Android-Cookbook-Examples-master/RecyclerViewDemo/app/src/main/res/mipmap-xhdpi/ic_launcher.png",
  "label": "Launcher Icon (XHDPI)"
}
```

```json
{
  "id": 12,
  "file_type": "image",
  "source_file": "<another mipmap-density variant of the same launcher icon>"
}
```

Both confirmed occurrences are Android launcher-icon images at different mipmap densities. Both instances crashed the entire extraction run (a large, multi-hour, otherwise-successful pass over a 10k+ file corpus) at the very last step, after all AST + semantic extraction had already completed and been cached.

## Repro context

- `graphify extract . --backend ollama --model qwen2.5-coder:3b --max-concurrency 1`
- Corpus: ~10,600 code files + ~800 docs + ~4,600 images (large mixed reference/sample-code corpus, not a single project)
- Windows, graphify installed via `uv tool install`

## Suggested fix

In `dedup.py`'s `_score()` (and anywhere else a node `id` is assumed to be a string), coerce with `str(n["id"])` before matching, or validate/normalize `id` to string immediately after parsing LLM output, before it reaches the dedup/build pipeline. Coercing at ingestion is probably more robust since other downstream code likely makes the same string assumption about `id`.

## Workaround used

Excluded image extensions via `.graphifyignore` (`*.png`, `*.jpg`, etc.) for this corpus and re-ran — unblocks the crash, at the cost of losing image nodes (acceptable for our use case, images are low-value nodes for a code-reference corpus anyway).

Rishet11:
Closes #2326

## The bug

A backend can emit `{"id": 10}` where the schema says `{"id": "10"}`. Every id consumer downstream assumes `str`, so one numeric id aborted the whole run at the final merge step, after the full extraction had already been paid for.

## Three crash sites, not one

While reproducing I found the reported traceback is only one of three:

| Case | Dies at |
|---|---|
| Numeric-id node lands in a duplicate group | `dedup.py:635` `_CHUNK_SUFFIX.search(n["id"])` → `expected string or bytes-like object, got 'int'` (the reported traceback) |
| Numeric-id node lands in a fuzzy-match group | `build.py` `sorted(node_set)` → `'<' not supported between instances of 'str' and 'int'` |
| **Single numeric-id node, nothing to dedup at all** | same `sorted(node_set)` |

That third row is why I did **not** take the one-line fix the traceback points at. A `str()` cast inside `_score` leaves the most common case still crashing, because a lone node never reaches `_pick_winner`.

## The fix

New `_coerce_non_string_ids(extraction)`, called at two sites:

- `build()`, immediately before `deduplicate_entities`, which keys on id
- `build_from_json()`, the direct entry used by `cli.py`, `watch.py` and `diagnostics.py`, which never goes through `build()`

Both are required: removing either reintroduces one of the crash sites. The nested call on the `build()` path is an idempotent no-op.

Edge endpoints and hyperedge members are coerced alongside the nodes on purpose. Coercing node ids alone would renumber `10` to `"10"` and leave every edge pointing at the vanished `10` — trading a loud crash for a silently disconnected graph. The legacy `from`/`to` aliases are included because dedup reads them directly (#803).

Only bool-free numeric scalars are coerced. A `None`, list or dict id is left for `validate_extraction` to report, since `str(None) == "None"` would fabricate an id that no edge references.

## Verification

- 13 new tests in `tests/test_non_string_node_ids.py`; **12 fail with the fix reverted**, so they genuinely bite
- Full suite on base `4fe1109`: 4 failed, 3895 passed, 3 skipped
- Full suite with this change: 4 failed, **3908** passed, 3 skipped — identical failure set (the 4 are pre-existing backend-detection tests reading real env vars), and +13 is exactly the new tests
- `ruff check .` clean; `python -m tools.skillgen --check`, `--audit-coverage`, `--schema-singleton` all OK
- End-to-end on the reporter's entry point `build_merge` with `{"id": 10}`: nodes `['10', 'b']`, edge `('10', 'b')` — node kept, edge intact

## Known limitation

If a backend emitted both `10` and `"10"` as separate nodes, they now merge, last-writer-wins. That matches existing behavior for two nodes already sharing a string id, so it is not a new failure mode, but it is worth knowing. Numeric hyperedge `id` fields are not coerced either — only hyperedge *members*, which are node references.

I could not reproduce against a live ollama backend (not available locally); I reproduced by feeding the builder the same fragment shape ollama produces, which is where the crash lives.

### DIFF
--- a/graphify/build.py
+++ b/graphify/build.py
@@ -157,6 +157,62 @@ def _fold_edge_aliases(edge: dict) -> None:
         edge["confidence"] = "INFERRED"
 
 
+def _coerce_id(value: object) -> object:
+    """Return a str for a numeric id, else the value unchanged.
+
+    ``bool`` is excluded despite subclassing ``int``: an id of ``True`` is not a
+    number the model meant to name a node, and ``"True"`` would invent a label.
+    """
+    if isinstance(value, bool) or not isinstance(value, (int, float)):
+        return value
+    return str(value)
+
+
+def _coerce_non_string_ids(extraction: dict) -> None:
+    """Coerce numeric node ids and edge/hyperedge references to str, in place (#2326).
+
+    A backend can emit ``{"id": 10}`` where the schema says ``{"id": "10"}``.
+    Every id consumer downstream assumes ``str``, so one int id aborted the build
+    in three places: ``_pick_winner``'s ``_CHUNK_SUFFIX.search(n["id"])`` raised
+    ``TypeError: expected string or bytes-like object``, and ``build_from_json``'s
+    ``sorted(node_set)`` raised ``'<' not supported between instances of 'str'
+    and 'int'`` — the latter for a lone node with nothing to dedup at all.
+    Coercing keeps the node and its edges rather than dropping either, which is
+    the same tolerate-and-heal treatment loose backend output already gets at the
+    parse chokepoint (#1631) and in the alias folds (#2194).
+
+    Endpoints and hyperedge members are coerced with the nodes, not after: a
+    node-only coercion would renumber ``10`` to ``"10"`` and leave every edge
+    pointing at the vanished ``10``, trading a loud crash for a silently
+    disconnected graph. The legacy ``from``/``to`` endpoint aliases are included
+    because dedup reads them directly (#803).
+
+    Runs in BOTH ``build`` (before dedup, which keys on id) and
+    ``build_from_json`` (the direct entry that reloads a persisted graph), for
+    the same two-site reason as the ``_fold_node_aliases`` fold (#2194). It is
+    idempotent, so the nested call on the ``build`` path is a no-op.
+
+    Non-numeric non-str ids (``None``, lists, dicts) are left alone for
+    ``validate_extraction`` to report: ``str(None) == "None"`` would fabricate a
+    node id that no edge references.
+    """
+    for node in extraction.get("nodes") or ():
+        if isinstance(node, dict) and "id" in node:
+            node["id"] = _coerce_id(node["id"])
+    for edge in extraction.get("edges") or ():
+        if not isinstance(edge, dict):
+            continue
+        for key in ("source", "target", "from", "to"):
+            if key in edge:
+                edge[key] = _coerce_id(edge[key])
+    for he in extraction.get("hyperedges") or ():
+        if not isinstance(he, dict):
+            continue
+        members = he.get("nodes")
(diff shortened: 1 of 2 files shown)


safishamsi:
Shipped in v0.9.32: non-string node ids from LLM fragments are coerced so a numeric id can no longer abort the build (#2326). Thanks @Rishet11!

PR #2338, opened 2026-07-31
fix(query): render every edge between visited nodes, not just traversal edges

### LINKED ISSUE #2323: `graphify query` omits edges between seed nodes and other visited nodes
## Summary

`graphify query` returns the relevant nodes but can omit real edges between them. This occurs when both endpoints are selected as seed nodes, and more generally for cross-edges between nodes that have already been visited.

Tested against Graphify `0.9.30` and the current `v8` branch at `ecfcd16`.

## Minimal reproduction

```python
# app.py
from pricing import discounted_total

def checkout(subtotal: float, discount_rate: float) -> float:
    return discounted_total(subtotal, discount_rate)
```

```python
# pricing.py
def discounted_total(subtotal: float, discount_rate: float) -> float:
    return subtotal * (1 - discount_rate)
```

```bash
graphify query "How does checkout calculate the final total?"
```

`graphify-out/graph.json` correctly contains a `checkout() --calls--> discounted_total()` edge. Both functions are selected as query seeds and both nodes appear in the output, but the `calls` edge is missing.

## Root cause

`_bfs()` and `_dfs()` in `graphify/serve.py` append an edge to `edges_seen` only when that edge discovers an unvisited neighbor:

```python
if neighbor not in visited:
    next_frontier.add(neighbor)
    edges_seen.append((n, neighbor))
```

Because all seed nodes are marked visited before traversal begins, an edge between two seeds is never recorded. The same behavior omits non-tree/cross-edges between other already-visited nodes. The rendered result is therefore a traversal tree rather than the induced subgraph described by the returned node set.

## Expected behavior

Every graph edge whose endpoints are both in the visited node set should be eligible for rendering, while preserving traversal-edge order and avoiding duplicates.

With that behavior, the reproduction includes:

```text
EDGE checkout() --calls [EXTRACTED context=call]--> discounted_total() at=app.py:L5
```

## Proposed fix

After BFS/DFS determines the visited node set, retain the traversal edges and append missing edges from `G.edges()` whose endpoints are both visited. Deduplicate using ordered endpoint pairs for directed graphs and unordered endpoint pairs for undirected graphs.

Validated locally with regression coverage for:

- an edge directly between two seed nodes;
- a cross-edge in a triangle where all three nodes are visited;
- the original CLI reproduction.

Relevant upstream tests passed:

```text
tests/test_serve.py + tests/test_query_cli.py
133 passed
```

## Disclosure

This report was prepared with OpenAI Codex at the user's request after reproducing the issue on a synthetic fixture. No private repository data is included.

Rishet11:
Closes #2323

Thanks for the precise root cause — it held up. Both traversals recorded an edge only when it discovered an *unvisited* neighbour, so what `query` rendered was a traversal tree, not the induced subgraph over the node set it reports.

## Before / after on your repro

Your two files, `graphify extract . --code-only`, then `graphify query "How does checkout calculate the final total?"`:

```text
# BEFORE — 3 edges, the calls edge missing
NODE checkout() [src=app.py loc=L3 community=0]
NODE discounted_total() [src=pricing.py loc=L1 community=0]
EDGE app.py --imports [EXTRACTED context=import]--> discounted_total() at=app.py:L1
EDGE pricing.py --contains [EXTRACTED]--> discounted_total() at=pricing.py:L1
EDGE app.py --contains [EXTRACTED]--> checkout() at=app.py:L3

# AFTER — 5 edges
...
EDGE app.py --imports_from [EXTRACTED context=import]--> pricing.py at=app.py:L1
EDGE checkout() --calls [EXTRACTED context=call]--> discounted_total() at=app.py:L4
```

## One correction to the report

`_dfs` does **not** share the defect in the general case. It appends on push rather than on visit, so it already captured seed-to-seed and ordinary cross-edges — I searched every connected four- and five-node graph exhaustively and found no miss. Its only gap is an edge between two **non-seed hubs**: the hub guard means neither endpoint is ever expanded, so neither records it. That case has its own test rather than a `_dfs` test that would have passed trivially.

## The fix

`_complete_induced_edges`, called at the end of both traversals.

- **Placed inside the traversals, not at the call site.** `_bfs`/`_dfs` receive `traversal_graph` (serve.py:1054-1055), so the completion pass is bound to the context-filtered graph by construction and cannot resurrect a relation the user filtered out. At the call site that would be a thing to remember.
- **Scans edges incident to the visited set**, not all of `G.edges()`, so cost tracks the subgraph. Bounded by O(2E). A visited hub is rescanned in full; that is unavoidable, since a hub-to-hub edge is precisely the case `_dfs` misses.
- **Dedup keys on the ordered pair when `G.is_directed()`, the unordered pair otherwise.** On a DiGraph `u->v` and `v->u` are genuinely distinct (mutual recursion, circular imports) and unordered dedup would drop a real edge. The MCP path loads every graph with `directed: True`, so this matters there.
- **Self-loops skipped.** A recursive function carries one, but no traversal ever recorded one, and surfacing them is a separate output change from the missing edges reported here.
- Traversal edges keep their discovery order; completions are appended after.

## Verification

- 11 new tests in `tests/test_query_induced_edges.py`; **9 fail with the fix reverted**
- Full suite on base `4fe1109`: 4 failed, 3895 passed, 3 skipped. With this change: 4 failed, **3906** passed, 3 skipped — **identical failure sets** (the 4 are pre-existing backend-detection tests reading real env vars)
- `tests/test_serve.py:476`, the only exact-equality assertion on `edges_seen`, passes untouched
- `ruff check .` clean; `python -m tools.skillgen --check`, `--audit-coverage`, `--schema-singleton` all OK

## Known limitations

- The MCP `query_graph` tool and its HTTP surface share this traversal, so their output gains edges too. Intentional; covered by a directed-graph test.
- Parallel edges on a multigraph still collapse to one rendered line — the renderer already showed only the first (serve.py:944 pre-change). Unchanged here.
- `_bfs`'s own traversal order is hash-seed sensitive (`frontier` is a plain set). Pre-existing and untouched; the appended completions are sorted, so only the tail is deterministic.
- `benchmark.py:50-71` carries a hand-inlined copy of the same traversal-plus-render logic. It is token estimation only and not in the render path, so I left it alone to keep the diff traceable to this issue — flagging it in case you want it mirrored.

## Property check, and one pre-existing oddity it surfaced

To confirm this holds generally and not just on the reported shape, I compared the returned edge list against the mathematically correct induced subgraph over 8,000 randomised traversals (2-9 nodes, directed and undirected, random densities, random seeds and depths, both `_bfs` and `_dfs`):

| | missing induced edges |
|---|---|
| base `4fe1109` | 34,547 |
| this branch | **0** |

That run also surfaced something pre-existing and unrelated to this change, which I did not touch: `_dfs` places 2,568 pairs in `edges_seen` whose endpoints were never visited. It appends on push, and a pushed node can then be dropped by the `d > depth` cut-off, so the pair stays in the list while the node never enters `visited`. The count is byte-identical before and after this PR, and `_subgraph_to_text`'s `if u in nodes and v in nodes` guard means none of them ever render — so it is invisible junk in the return value rather than a bug in output. Mentioning it only because the property test made it visible; happy to open a separate issue if you would like it cleaned up.

### DIFF
--- a/graphify/serve.py
+++ b/graphify/serve.py
@@ -819,6 +819,56 @@ def _filter_graph_by_context(G: nx.Graph, context_filters: list[str] | None) -> 
     return H
 
 
+def _complete_induced_edges(G: nx.Graph, visited: set[str], edges_seen: list[tuple]) -> None:
+    """Append edges between visited nodes that the traversal never recorded (#2323).
+
+    Both traversals only record an edge that *discovers* an unvisited neighbour,
+    so what they return is a traversal tree, not the induced subgraph over the
+    nodes they return. `_bfs` marks every seed visited up front, so an edge
+    between two seeds can never be recorded — the reported symptom, where both
+    endpoints render and the edge between them does not. It drops ordinary
+    cross-edges for the same reason. `_dfs` appends on push rather than on
+    visit, so it already captured those; its one gap is an edge between two
+    non-seed hubs, since neither endpoint is ever expanded.
+
+    Scans only edges incident to `visited`, so cost tracks the subgraph rather
+    than the whole graph, bounded by O(2E) overall. A visited hub is rescanned
+    in full even though the traversal deliberately did not expand it — that is
+    unavoidable, since a hub-to-hub edge is exactly the case `_dfs` misses.
+    `G` here is the context-filtered `traversal_graph` (see
+    `_query_graph_text`), so a filtered-out relation cannot reappear.
+
+    Self-loops are skipped. A recursive function legitimately carries one, but
+    neither traversal has ever recorded one (`n` is always already visited when
+    its own self-loop is examined), and surfacing them is a separate output
+    change from the missing edges reported here.
+
+    Dedup keys on the ordered pair for directed graphs and the unordered pair
+    otherwise: on a DiGraph `u->v` and `v->u` are genuinely distinct edges
+    (mutual recursion, circular imports), and collapsing them would drop a real
+    one. On a multigraph parallel edges collapse to one entry, matching the
+    renderer, which already shows only the first (`_subgraph_to_text`).
+
+    Traversal edges keep their discovery order; completions are appended after.
+    """
+    directed = G.is_directed()
+
+    def _key(u: str, v: str):
+        return (u, v) if directed else frozenset((u, v))
+
+    seen = {_key(u, v) for u, v in edges_seen}
+    # sorted() so the appended order can't shift run-to-run with CPython's
+    # per-process string-hash seed, the same reason the renderer sorts (#1753).
+    for u, v in G.edges(sorted(visited)):
+        if u == v or v not in visited:
+            continue
+        key = _key(u, v)
+        if key in seen:
+            continue
+        seen.add(key)
+        edges_seen.append((u, v))
+
+
 def _bfs(G: nx.Graph, start_nodes: list[str], depth: int) -> tuple[set[str], list[tuple]]:
     # Compute hub threshold: nodes above this degree are not expanded as transit.
     # p99 of degree distribution, floored at 50 to avoid over-blocking small graphs.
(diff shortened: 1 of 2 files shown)


safishamsi:
Shipped in v0.9.32: `graphify query` now renders every edge between visited nodes, not just the traversal-tree edges (#2323). Thanks @Rishet11!

PR #2235, opened 2026-07-27
feat(export): 3D WebGL renderer for graph.html behind --viz 3d

lopezonchain:
- New WebGL renderer (3d-force-graph, SRI-pinned like vis-network), selected with --viz 2d|3d on the main pipeline, cluster-only/label and export html, or with GRAPHIFY_VIZ_MODE.
- to_html() now builds a renderer-agnostic view model both renderers consume. 2d stays the default and its output is byte-identical, guarded by a test.
- 3D navigation: search, click-to-inspect, community filters, a "Show" control isolating 1-6 hops around the selection, an opt-in name overlay and keyboard shortcuts. Neighbour links keep the escaped data-nid listener from #1838.
- Skill flag list gains --viz 3d via the core fragment; the aider/devin monoliths stay frozen against their pinned v8 blob.

Issue #175, opened 2026-04-10
offline support


kongshan001:
# Offline Visualization Modification (vis-network Inlining)
## Problem
The `graph.html` file loads `vis-network` via external CDN:
`<script src="https://unpkg.com/...">`
The visualization page cannot be opened on intranet / offline devices.

## Solution
Modify `export.py` in the `graphify` package to inline `vis-network.min.js` (644 KB) directly into the HTML output, generating a fully self-contained, offline-ready single file.

## Modified File
`<site-packages>/graphify/export.py`

## Step 1: Download the Library
Download `vis-network.min.js` and place it in the package directory:
`<site-packages>/graphify/_vendor_vis-network.min.js` (644 KB)

## Step 2: Add New Inlining Function
Add the `_vis_network_inline()` function:
```python
def _vis_network_inline() -> str:
    """Return vis-network.min.js content for offline-capable HTML."""
    vendor = Path(__file__).parent / "_vendor_vis-network.min.js"
    if vendor.exists():
        return vendor.read_text(encoding="utf-8")
    # fallback: CDN reference if vendor file missing
    return 'document.write("<script src=\\"https://unpkg.com/vis-network/standalone/umd/vis-network.min.js\\"></script>")'
```

## Step 3: Replace External Script Reference
Update the HTML template:

### Before
```html
<script src="https://unpkg.com/vis-network/standalone/umd/vis-network.min.js"></script>
```

### After
```html
<script>{_vis_network_inline()}</script>
```

## Effect Comparison
| Item                | Before Modification       | After Modification        |
|---------------------|---------------------------|---------------------------|
| External dependency | Requires unpkg.com        | None                      |
| File size           | ~203 KB                   | ~829 KB (+626 KB inlined JS) |
| Offline usable      | No                        | Yes                       |
| Fallback mechanism  | None                      | Automatically falls back to CDN if vendor file is missing |

## Notes
- After modification, regenerate the HTML by running `/graphify .` or calling `to_html()` to apply the offline version.
- Running `pip install --upgrade graphify` will overwrite `export.py`; the patch must be reapplied after upgrade.
- To upgrade the vis-network version, simply replace `_vendor_vis-network.min.js` with the updated file.

bolwage777:
Still reproduces on **graphify v0.9.29** (via the Claude Code skill flow). `graphify export html` writes `graph.html` with a hard-coded external tag:

```html
[...]

PR #2344, opened 2026-07-31
feat(export): add fully offline 3D HTML renderer

AmeerJ97:
## Summary

Add an opt-in, natively rendered WebGL view to the existing HTML export:

```bash
graphify export html --viz 3d
```

The default remains `--viz 2d`. Both renderers consume the same sanitized view model, so community labels, colors, node sizing, edge direction/confidence, learning status, and aggregated community views stay consistent.

Unlike the current 2D CDN page and earlier 3D proposals, the generated 3D document embeds the exact reviewed renderer bundle. It can be copied to an offline or restricted machine and opened directly in a browser without a server or network request.

## Why this shape

A renderer selector is preferable to a standalone `graph3d` command or a one-off boolean flag:

- it preserves `graphify export html` as the canonical export pipeline;
- `2d` remains backwards-compatible and byte-identical by default;
- future renderers can extend the same mode boundary;
- node-limit aggregation cannot accidentally fall back to a different renderer;
- `/graphify`, `cluster-only`, `label`, and explicit HTML export can share the same selection semantics.

## Renderer and offline packaging

This uses `3d-force-graph` 1.80.0, whose UMD distribution bundles Three.js/WebGL, d3-force-3d, and its browser runtime dependencies.

The reviewed minified asset is packaged under `graphify/vendor/` and embedded in the output with `importlib.resources`:

- exact version: `1.80.0`;
- exact asset size: 1,313,897 bytes;
- SHA-384: `63b6c2d8f04abbcba3c6dbe8e7e67ad4e786752551cc5b185812b88b97672ff53a6850d3a1d93ad6a3949137c89dac52`;
- upstream MIT license and transitive runtime notices are included;
- wheel and sdist package-data declarations are tested by a clean wheel install;
- literal closing-script sequences are neutralized before embedding.

This intentionally adds no Python runtime dependency and no JavaScript build toolchain to Graphify.

## 3D interaction model

The 3D view provides:

- orbit, pan, zoom, reset, and freeze controls;
- node search and click-to-inspect details;
- delegated neighbor navigation without inline event handlers;
- community legend filtering and isolate/toggle shortcuts;
- selection-centered 1–6 hop focus;
- opt-in bounded node-name overlays;
- hyperedge star expansion with clickable hub nodes;
- learning-status coloring;
- a clear recovery message directing users to `--viz 2d` when WebGL is unavailable.

Community names come from `.graphify_labels.json`. Colors remain deterministic by community ID, while node scale follows degree or aggregated member count, matching the existing 2D semantics.

## Large-graph safety and performance

This does not remove Graphify's existing safety boundary. The default 5,000-node limit and community aggregation behavior are preserved for both renderers.

The force layout uses:

- a bounded many-body distance;
- weak center gravity for disconnected components;
- bounded warmup and cooldown ticks;
- graph-size-adaptive warmup (150 / 90 / 50 ticks);
- arrow-geometry degradation only above 4,000 edges;
- a 48-label normal budget and 150-label narrowed-view budget;
- no animation-frame loop while labels are disabled.

A reproducible synthetic 5,000-node path-graph check in headless Chrome with SwiftShader and networking disabled improved from 17.85 seconds with a constant 150-tick warmup to 7.96 seconds with the adaptive policy. This is a guardrail measurement, not a universal throughput claim; force-layout cost still depends heavily on graph density and browser/GPU.

## Security and failure handling

- Graph data continues through `js_safe`, including `</script>` boundary protection.
- Tooltip content is escaped once rather than double-escaped.
- Neighbor selection uses escaped `data-nid` values and a delegated listener, with no inline `onclick` interpolation.
- The 3D page has no external `<script src>`, stylesheet, font, or texture dependency.
- Missing and invalid `--viz` values now fail explicitly instead of silently producing a 2D page.
- WebGL capability is checked before renderer initialization and produces an actionable in-page fallback message.

## Compatibility

- Python remains `>=3.10`; the implementation uses Python 3.10-compatible typing and `importlib.resources`.
- Existing callers receive the 2D renderer by default.
- A regression test verifies default output equals explicit `mode="2d"` output.
- `GRAPHIFY_VIZ_MODE=3d` is supported, with an explicit function/CLI argument taking precedence.
- Community aggregation propagates the requested renderer mode.

## Verification

- full suite on Python 3.10: **4,566 passed, 7 skipped**\n- full suite on Python 3.12: **4,566 passed, 7 skipped**\n- focused 3D/CLI suite: **81 passed**
- `uv run --frozen ruff check .`
- `uv run --frozen pre-commit run --all-files`
- targeted Pyright for the changed exporter/tests: **0 errors**
- `uv build --sdist --wheel`
- `twine check` on wheel and sdist: **passed**
- clean wheel install generated an offline 3D page successfully
- headless Chrome with DNS mapped to failure rendered a Three.js canvas
- headless Chrome with GPU disabled rendered the WebGL recovery guidance
- Opengrep: no applicable feature finding (one inapplicable Python-3.7 compatibility note; Graphify requires 3.10+)
- Open Code Review: zero comments across all reviewed production files

Bandit and pip-audit retain pre-existing repository/lockfile findings; this change adds no Python dependency and introduced no new finding in the changed source.

## Prior work and coordination

This is intended as a consolidation/successor rather than unrelated duplicate work:

- It preserves the feature commit from #2235 with @lopezonchain as its original author, rebased onto current `v8`.
- It builds on that renderer-neutral model while replacing the CDN with a packaged offline asset, tightening CLI errors and XSS behavior, adding WebGL failure handling, and bounding large-graph startup work.
- It also acknowledges the alternative approaches in #2170 and #225.
- It is related to the offline-export need in #175, but does **not** claim to close that issue because the default 2D vis-network export remains CDN-backed.

Maintainers can use this PR as the consolidation point or ask me to narrow/restructure it around #2235; authorship of the foundational renderer commit has deliberately been preserved.

## Reviewer notes

The generated skill artifacts account for most of the file count: the human-edited source is `tools/skillgen/fragments/core/core.md`, and the checked artifacts were regenerated through the repository's normal skillgen workflow. The core implementation surface is limited to the CLI, shared HTML model, new 3D renderer, vendored browser asset/notices, tests, and documentation.


AmeerJ97:
Rebased on current `v8` and addressed the Graphify advisory in 997ac56: both CLI paths now trim/lower validated `--viz` values before passing them to `to_html`.

Verified with the locked `uv` environment:
- focused CLI/3D-export coverage: 81 passed
- complete `tests/` corpus in bounded CI-equivalent batches (all passed)
- `python -m tools.skillgen --check` and Ruff both pass
- AST-only `graphify update .` refreshed the local code graph

The PR is mergeable. The repository CI workflow for this fork currently shows `action_required` before any jobs start; could a maintainer approve or re-run it, then review this updated head?

AmeerJ97:
Rebased this consolidation PR onto current `v8` at `0738af3` (`0.9.45`); GitHub now reports the head as mergeable.

The only semantic conflict was in `graphify/exporters/html.py`. The shared 2D/3D model now preserves upstream’s portable `_html_document_title()` behavior, and the changelog entry is under the current `0.9.45 (unreleased)` section.

Current-head verification:
- full suite, Python 3.10: **4,566 passed, 7 skipped**
- full suite, Python 3.12: **4,566 passed, 7 skipped**
- focused 3D/CLI suite: **81 passed**
- all five skillgen/parity validators and Ruff: passed
- targeted Pyright: **0 errors**
- wheel + sdist build, Twine checks, packaged vendor assets, and project-scoped install smoke test: passed
- AST-only `graphify update .`: completed

#2559 is complementary: it adds an offline-assets option for the existing exporters, while this PR’s distinct scope is the optional renderer-neutral `--viz 3d` path and its interaction model.

Could a maintainer approve the fork CI run (it again stopped at `action_required` with zero jobs) and advise whether #2344 should remain the consolidation point for #2235/#2170/#225, or whether the project prefers a different 3D/WebGL direction?

Issue #2081, opened 2026-07-21
built_at_commit is read from CWD, not the extraction target — stamps graphs with an unrelated repo's commit (and fabricates one for non-git targets)


Kyo14363:
## Summary

`built_at_commit` is read from the **current working directory's** git repo, not from the **target path** being extracted. When you run `graphify extract <target>` from anywhere other than inside `<target>`, the graph is stamped with an unrelated commit.

The worst variant: if the target is **not a git repository at all**, the graph still gets stamped with the CWD repo's commit — fabricated provenance rather than an absent field.

This directly undermines the `## Graph Freshness` section that `GRAPH_REPORT.md` emits, which instructs the reader:

> - Built from commit: `<hash>`
> - Run `git rev-parse HEAD` and compare to check if the graph is stale.

Following that instruction compares the target repo's HEAD against a foreign repo's commit. The comparison silently always mismatches (or, worse, coincidentally matches), so staleness detection cannot be trusted.

## Reproduction

Minimal, two throwaway repos:

```bash
mkdir -p /tmp/repro/repo-A /tmp/repro/repo-B

cd /tmp/repro/repo-A
git init -q -b main && echo "print('A')" > a.py
git add -A && git commit -qm "repo A commit"

cd /tmp/repro/repo-B
git init -q -b main && echo "def f(): pass" > b.py
git add -A && git commit -qm "repo B commit"

git -C /tmp/repro/repo-A rev-parse HEAD   # cwd repo
git -C /tmp/repro/repo-B rev-parse HEAD   # target repo

# run from repo-A, targeting repo-B
cd /tmp/repro/repo-A
graphify extract /tmp/repro/repo-B --code-only --out /tmp/repro/out

python -c "import json;print(json.load(open('/tmp/repro/out/graphify-out/graph.json'))['built_at_commit'])"
```

**Actual** — records repo-A's commit:

```
repo-A HEAD (cwd)        : cdf257124f9cb3f28bdf5d98a1d2f189760551ba
repo-B HEAD (target)     : 05aa698cc0c5306c1e27a720b5f7be2effaa08d3
built_at_commit recorded : cdf257124f9cb3f28bdf5d98a1d2f189760551ba   ← repo-A
```

**Expected** — `05aa698c…` (the target's HEAD).

### Non-git target (worse)

```bash
mkdir -p /tmp/repro/repo-C-nogit
echo "def g(): pass" > /tmp/repro/repo-C-nogit/c.py

[...]

PR #2345, opened 2026-07-31
fix(watch): write manifest.json to the target's graphify-out (#2316)

### LINKED ISSUE #2316: update writes manifest.json to the CWD instead of the target — cross-project contamination, and the target gets no manifest
## Summary

`graphify update <target>` writes `manifest.json` into **the current working directory's** `graphify-out/`, not the target's — while every other artifact (`graph.json`, `graph.html`, `GRAPH_REPORT.md`, `cache/`, `.graphify_root`) goes to the target correctly.

Two things break at once:

1. **The CWD project gets a foreign manifest.** Its `graphify-out/manifest.json` now describes another project's files. Any tool that reads the manifest as "what the indexer saw for this project" gets a confidently wrong answer.
2. **The target ends up with no manifest at all**, so its incremental detection has nothing to compare against on the next run.

Running `graphify update` over a list of N projects from one script therefore leaves N-1 manifests wrong and one project's `graphify-out/` holding the last-processed project's file list.

This is the same root cause family as #2081 (`built_at_commit` read from CWD instead of the extraction target), but a different field and a different code path — #2081 is `extract`, this is the `watch`-based `update` path.

## Reproduction

```bash
mkdir -p /tmp/repro/projA /tmp/repro/projB
printf 'def alpha_only_in_A():\n    return 1\n' > /tmp/repro/projA/a_file.py
printf 'def beta_only_in_B():\n    return 2\n'  > /tmp/repro/projB/b_file.py
for d in projA projB; do (cd /tmp/repro/$d && git init -q . && git add -A && git commit -qm init); done

# run from projA, targeting projB
cd /tmp/repro/projA
graphify update /tmp/repro/projB

ls /tmp/repro/projA/graphify-out/
ls /tmp/repro/projB/graphify-out/
cat /tmp/repro/projA/graphify-out/manifest.json
```

**Actual** (graphify 0.9.13):

```
/tmp/repro/projA/graphify-out/:   manifest.json          <-- only this, and it is projB's
/tmp/repro/projB/graphify-out/:   cache/ graph.json graph.html GRAPH_REPORT.md
                                  .graphify_labels.json .graphify_root
                                  (no manifest.json)

$ cat /tmp/repro/projA/graphify-out/manifest.json
{
  "b_file.py": {                 <-- projB's file, in projA's output dir
    "mtime": 1785360486.2794108,
    "ast_hash": "259d778f11c851feb2ab5285781ebcd3",
    "semantic_hash": ""
  }
}
```

**Expected** — `manifest.json` lands next to the graph it describes, in `projB/graphify-out/`, and `projA` is not touched.

## Root cause

`detect.py:28` freezes the manifest location at **import time**, from a relative path:

```python
# graphify/paths.py:23
GRAPHIFY_OUT = os.environ.get("GRAPHIFY_OUT", "graphify-out")   # relative by default

# graphify/paths.py:224
def out_path(*parts: str) -> Path:
    return Path(GRAPHIFY_OUT, *parts)

# graphify/detect.py:28
_MANIFEST_PATH = str(out_path("manifest.json"))   # module-level -> resolves against the CWD
```

`save_manifest`, `load_manifest` and `detect_incremental` all default `manifest_path=_MANIFEST_PATH`.

`cli.py` does the right thing and passes the path explicitly (e.g. `cli.py:2472`):

```python
_save_manifest(_manifest_files, manifest_path=str(manifest_path), kind="both", root=target)
```

But the `watch` rebuild path does not — `watch.py:970`, `1009`, `1087`:

```python
save_manifest(detected["files"], kind="ast", root=project_root)
```

It passes `root=project_root`, so the correct root is already in hand, but lets the *file location* fall back to the CWD-frozen constant. The correct output dir is also already in scope as `out` — used two lines earlier:

```python
(out / ".graphify_root").write_text(str(watch_path), encoding="utf-8")

try:
    from graphify.detect import save_manifest
    save_manifest(detected["files"], kind="ast", root=project_root)
except Exception:
    pass
```

## Suggested fix (verified)

One line at each of the three `watch.py` call sites:

```python
save_manifest(detected["files"], manifest_path=str(out / "manifest.json"), kind="ast", root=project_root)
```

Applied locally against 0.9.13 and re-ran the reproduction above:

| | `projA/graphify-out` (CWD) | `projB/graphify-out/manifest.json` (target) |
|---|---|---|
| before | `manifest.json`, containing `b_file.py` | absent |
| after | not created at all | present, containing `b_file.py` |

## Side note

The three call sites are wrapped in `except Exception: pass`. A manifest write that fails for any other reason is also silent, so incremental detection would quietly degrade to a full rescan with nothing in the output to say why. Not the bug reported here, but it is why this one was invisible for so long — worth a log line even if the swallow stays.

## Workaround

`cd` into the target before calling `update` (or set the process CWD to the target per project when driving `update` from a script). With CWD == target, the frozen relative path happens to resolve to the right place.

## Environment

- graphify 0.9.13
- Windows 11, Python 3.10 (uv tool install), also reproducible with a POSIX shell

Rishet11:
Closes #2316

`graphify update <target>` wrote `manifest.json` into the process CWD's `graphify-out/` while every other artifact went to the target. Reproduced on `v8` `4fe1109` (0.9.31), and the reproduction turned up **two more defects in the same function** that the report does not cover. All three are fixed here; `graphify/watch.py` is the only file touched.

## Root cause

`_rebuild_code` computes the correct output dir at `watch.py:897`:

```python
out = watch_path / _GRAPHIFY_OUT
```

and routes every artifact through it, including one line above each manifest write (`out / ".graphify_root"`, `out / "needs_update"`). The three `save_manifest(...)` calls omitted `manifest_path=`, so they fell back to `detect._MANIFEST_PATH` (`detect.py:29`) — a module-import-time constant built from the relative `out_path("manifest.json")` (`paths.py:288`), resolved against the process CWD. `cli.py:2815` is the reference-correct wiring and passes it explicitly at `2853/3404/3510/3657`.

## Three defects

**1. Location** — as reported. The reporter's line numbers (970/1009/1087, from 0.9.13) have drifted to **1232/1275/1421** at `4fe1109`.

**2. It is data loss on the invoking project, not just a stray file.** `save_manifest` read-modify-writes the same path (`detect.py:1615`) and full-scan callers pass `scan_corpus` (#1908 pruning), so the invoking project's own rows are pruned as out-of-corpus and overwritten:

```
$ cd projA && graphify update .
$ cat projA/graphify-out/manifest.json
{"a_file.py": {...}}

$ graphify update /tmp/repro/projB          # from projA
$ cat projA/graphify-out/manifest.json
{"b_file.py": {...}}                        # a_file.py destroyed
```

Both projects then need a full rescan, not only the target.

**3. A relative target stored absolute keys.** `root=` followed `project_root`, which is CWD-derived for a relative invocation, so `graphify update ../projB` put the target's files out-of-root and `_to_relative_for_storage` kept them absolute — breaking the #777/#1964 portability guarantee even once the file is in the right directory:

```
$ cd projA && graphify update ../projB
{"/private/tmp/repro/projB/b_file.py": {...}}     # not "b_file.py"
```

Fixed by passing `root=watch_root` at those three calls. **`project_root` itself is deliberately left alone** — the persisted graph rehomes `source_file` against it when the user switches between absolute and relative invocation styles (`tests/test_watch.py:1389`, `:1428`). The manifest and the persisted graph are different artifacts with legitimately different anchors; the manifest's consumer is `detect_incremental(root=target)`.

**4. `built_at_commit` came from the wrong repo.** `_git_head()` ran `git rev-parse HEAD` with no `cwd=`, so it read the caller's repository:

```
projB HEAD:          66dc5e6d2fab2ca4e623a1a70ee4a4ea67828388
projA HEAD:          011fc671d2336d3c56c2df5376243096bc70ed46
projB graph.json:    "built_at_commit": "011fc671..."     # projA's
```

This is #2081 in the watch path — wrong provenance data rather than a misplaced file. One `cwd=` parameter.

## Before / after, the reporter's exact repro

| | before | after |
|---|---|---|
| `projA/graphify-out/` | `manifest.json`, containing projB's files | not created at all |
| `projA` pre-existing manifest | overwritten, own rows pruned | byte-identical |
| `projB/graphify-out/manifest.json` | absent | present, key `b_file.py` |
| relative-target key | `/private/tmp/.../projB/b_file.py` | `b_file.py` |
| `projB` `built_at_commit` | projA's HEAD | projB's HEAD |

## Verification

- `tests/test_watch_manifest_location.py`, 7 tests, **all 7 red at `4fe1109` and green with the fix** (verified by reverting only `watch.py`). They cover all three write sites: clustered, `--no-cluster`, and the unchanged-topology early return. The early-return test asserts on the `"No code-graph topology changes detected"` line so it cannot silently drift onto another branch and keep passing.
- Full suite `-p no:randomly`: base **4 failed / 3895 passed / 3 skipped** → **4 failed / 3902 passed / 3 skipped**, failure-set diff **empty**. The 4 are pre-existing backend-detection tests that read real env vars.
- `ruff check .` clean. `tools.skillgen --check` (134 artifacts), `--audit-coverage`, `--schema-singleton` all clean.

## Known gaps, stated openly

- The three call sites remain wrapped in bare `except Exception: pass`, so a manifest write that fails for any other reason is still silent. The reporter raises this; it is a different change and I left it alone rather than widen the diff. Happy to add a log line here if you want it.
- `tools/skillgen/fragments/references/shared/update.md:163` has the same missing `manifest_path` in the agent runbook (`save_manifest(..., root='INPUT_PATH')`). Same bug class, different artifact, and it needs regenerated skill markdown — not folded in here.
- `export.py:173` has its own `_git_head()` with the same no-`cwd=` shape, used by the `extract` path. Untouched.
- `_report_root_label` (`watch.py:268`) falls back to `Path.cwd().name` for a relative path. Cosmetic label only, left as is.

### DIFF
--- a/graphify/watch.py
+++ b/graphify/watch.py
@@ -242,11 +242,21 @@ def _apply_resource_limits() -> None:
         pass
 
 
-def _git_head() -> str | None:
-    """Return current git HEAD commit hash, or None outside a repo."""
+def _git_head(cwd: Path | str | None = None) -> str | None:
+    """Return current git HEAD commit hash, or None outside a repo.
+
+    ``cwd`` selects the repository to ask (#2316). Without it the command
+    inherits the caller's working directory, so `graphify update <target>`
+    stamped the *invoking* repo's commit into the target's graph.json — the
+    same CWD-anchoring mistake as the manifest path, but writing wrong
+    provenance rather than a misplaced file.
+    """
     import subprocess as _sp
     try:
-        r = _sp.run(["git", "rev-parse", "HEAD"], capture_output=True, text=True, timeout=3)
+        r = _sp.run(
+            ["git", "rev-parse", "HEAD"], capture_output=True, text=True, timeout=3,
+            cwd=str(cwd) if cwd is not None else None,
+        )
         return r.stdout.strip() if r.returncode == 0 else None
     except Exception:
         return None
@@ -947,6 +957,10 @@ def _rebuild_code(
             return ok
 
     watch_root = watch_path.resolve()
+    # project_root stays CWD-anchored for a relative invocation on purpose: the
+    # persisted graph rehomes source_file across invocation styles against it
+    # (tests/test_watch.py:1389, :1428). The manifest is a different artifact
+    # with a different anchor — see the save_manifest calls below.
     project_root = Path.cwd().resolve() if not watch_path.is_absolute() else watch_root
     report_root = _report_root_label(watch_path)
     try:
@@ -1114,7 +1128,7 @@ def _rebuild_code(
             # below still counts the doc as a rebuilt source.
             extract_targets = [p for p in code_files if p not in semantic_doc_files]
 
-        commit = _git_head()
+        commit = _git_head(cwd=watch_root)
         result = extract(extract_targets, cache_root=watch_root) if extract_targets else {
             "nodes": [], "edges": [], "hyperedges": [],
             "input_tokens": 0, "output_tokens": 0,
@@ -1230,7 +1244,8 @@ def _rebuild_code(
                 # scan but still exist on disk (newly excluded) are pruned
                 # instead of surviving as phantom "deleted" entries (#1908).
                 save_manifest(
-                    detected["files"], kind="ast", root=project_root,
+                    detected["files"], manifest_path=str(out / "manifest.json"),
+                    kind="ast", root=watch_root,
                     scan_corpus={f for _fl in detected["files"].values() for f in _fl},
                 )
             except Exception:
@@ -1273,7 +1288,8 @@ def _rebuild_code(
                     from graphify.detect import save_manifest
                     # Full-scan save: prune excluded-but-alive rows (#1908).
                     save_manifest(
-                        detected["files"], kind="ast", root=project_root,
+                        detected["files"], manifest_path=str(out / "manifest.json"),
+                        kind="ast", root=watch_root,
                         scan_corpus={f for _fl in detected["files"].values() for f in _fl},
                     )
                 except Exception:
@@ -1419,7 +1435,8 @@ def _rebuild_code(
             from graphify.detect import save_manifest
             # Full-scan save: prune excluded-but-alive rows (#1908).
             save_manifest(
-                detected["files"], kind="ast", root=project_root,
+                detected["files"], manifest_path=str(out / "manifest.json"),
+                kind="ast", root=watch_root,
                 scan_corpus={f for _fl in detected["files"].values() for f in _fl},
             )
         except Exception:
--- a/tests/test_watch_manifest_location.py
+++ b/tests/test_watch_manifest_location.py
@@ -1,0 +1,237 @@
+"""#2316: `graphify update <target>` must write manifest.json into the TARGET's
+graphify-out/, not the process CWD's.
+
+`_rebuild_code` computes ``out = watch_path / _GRAPHIFY_OUT`` and routes every
+other artifact through it, but the three ``save_manifest`` calls omitted
+``manifest_path=`` and so fell through to ``detect._MANIFEST_PATH`` — a
+module-import-time constant resolved against the CWD. Running `update` on
+project B from project A therefore wrote B's manifest into A's output dir and
+left B with none.
+
+Because ``save_manifest`` also *reads* that path and full-scan callers pass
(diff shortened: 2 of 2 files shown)


safishamsi:
Shipped in v0.9.32: `manifest.json` is now written to the target graphify-out instead of the CWD (#2316). Thanks @Rishet11!

Issue #801, opened 2026-05-10
skill.md --update inline merge bypasses fixed build_merge() — direction-inversion bug from #760 still hits Claude Code skill workflow


twipixel:
**Package:** [`graphifyy`](https://pypi.org/project/graphifyy/) v0.7.13
**Skill version installed:** 0.7.8 (warning printed by CLI)
**Environment:** macOS 24.6.0, Python 3.11 (uv tool install of `graphifyy`)
**Related (closed):** #760

## Summary

Issue #760 fixed the `build_merge()` direction-inversion bug in `build.py`. The fix is correctly used by the CLI `graphify update <path>` command (`__main__.py:2608` calls `_build_merge`).

However, the bundled **`skill.md`** template still ships a hand-rolled merge block in its `--update` section that **never calls `build_merge()`**. It uses the exact `node_link_graph()` round-trip pattern that #760 identified as the root cause. As a result, any AI coding assistant that follows the skill instructions (Claude Code, Codex, Cursor, etc.) re-introduces the very bug #760 closed.

## Evidence

`graphify-0.7.13/skill.md` line 788 (the `--update` flow):

```python
existing_data = json.loads(Path('graphify-out/graph.json').read_text())
G_existing = json_graph.node_link_graph(existing_data, edges='links')   # round-trip — direction lost
...
G_existing.update(G_new)
to_json(G_existing, communities, 'graphify-out/graph.json')             # write back inverted
```

Compare with the fix in `build.py:235`:

```python
def build_merge(...):
    ...
    # Read JSON directly instead of going through node_link_graph().
```

The skill never imports `build_merge`.

## Reproduction

1. Build a TypeScript project graph with a small helper imported by many call sites — e.g. an `axios` wrapper called `requestGet` imported in 100+ API files.
2. Run `/graphify --update` (Claude Code skill path), which executes the inline merge code from `skill.md`.
3. Inspect `graph.json` for edges where `source = requestGet`. They will all be of the form `requestGet --calls--> getXxx()` despite `requestGet` being a callee, not a caller, in every actual source file.

In one real codebase:
- `requestGet` (5-line axios wrapper) — graph reports **105 outgoing `INFERRED 0.8` `calls` edges**.
- `grep` ground truth: **171 files import `requestGet`**, **178 actual call sites**.
- All 105 edges point the wrong way; another 73 actual call sites are missing entirely.
[...]

PR #2350, opened 2026-07-31
fix(build): inherit persisted directed flag on incremental rebuild (#2342)

### LINKED ISSUE #2342: build_merge() defaults to directed=False, so `graphify update` silently downgrades a directed graph to undirected — betweenness becomes meaningless with no warning
## Summary

`graphify update` (incremental mode) calls `build_merge()` without passing `directed=`, which defaults to `False`. Since the merged graph is then written back with `to_json(..., force=True)`, **a graph built with `--directed` becomes undirected after any routine refresh**, and nothing in stdout or the report says so.

The saved `directed` flag round-trips correctly everywhere else — `cluster-only` inherits it (`cli.py:1681`, `build_from_json(_raw, directed=_directed)`), and `#801` already went to the trouble of preserving per-edge `source`/`target` through the merge. It's specifically the graph *type* that is dropped.

Verified against **v1.0.0** (current default branch) and reproduced on 0.8.27.

## Why it matters

Betweenness centrality is only meaningful on a directed graph. Undirected, a leaf utility that everything imports becomes a fake bridge: every pair of its importers appears two hops apart "through" it, though no control flow passes through it at all.

On a real 4,607-node / 10,760-edge polyglot graph, a logging helper with `in_degree=176, out_degree=0` — a pure sink — ranked as the **#1 god node at 0.090 betweenness**. Rebuilt directed, it scores exactly **0.0000**, along with three other high-degree sinks that had been occupying most of the top five. The genuine bridges were invisible until those four dropped out.

So the failure mode isn't "the graph looks a bit different" — it's that `## God Nodes` and the betweenness-derived `## Suggested Questions` both silently start reporting ubiquity instead of architecture, and there is no signal that the graph changed type.

## Reproduction

Minimal, no corpus required:

```python
import json
from pathlib import Path
from graphify.build import build_from_json, build_merge
from graphify.cluster import cluster

Path('graphify-out').mkdir(exist_ok=True)
extraction = {
    'nodes': [
        {'id': 'a_caller', 'label': 'caller()', 'file_type': 'code', 'source_file': 'a.py'},
        {'id': 'a_callee', 'label': 'callee()', 'file_type': 'code', 'source_file': 'a.py'},
        {'id': 'a_leaf',   'label': 'leaf()',   'file_type': 'code', 'source_file': 'a.py'},
    ],
    'edges': [
        {'source': 'a_caller', 'target': 'a_callee', 'relation': 'calls',
         'confidence': 'EXTRACTED', 'confidence_score': 1.0, 'source_file': 'a.py'},
        {'source': 'a_callee', 'target': 'a_leaf', 'relation': 'calls',
         'confidence': 'EXTRACTED', 'confidence_score': 1.0, 'source_file': 'a.py'},
    ],
    'hyperedges': [],
}

G = build_from_json(extraction, directed=True)
to_json(G, cluster(G), 'graphify-out/graph.json')
print('saved directed:', json.loads(Path('graphify-out/graph.json').read_text())['directed'])

# what `graphify update` does
G2 = build_merge([{'nodes': [], 'edges': [], 'hyperedges': []}],
                 graph_path='graphify-out/graph.json', prune_sources=None)
to_json(G2, cluster(G2), 'graphify-out/graph.json', force=True)
print('after update  :', json.loads(Path('graphify-out/graph.json').read_text())['directed'])
```

```
saved directed: True
after update  : False
```

Passing `directed=True` to that same `build_merge` call returns `is_directed()=True`, confirming the plumbing works and only the default is at fault.

## Location

`graphify/cli.py:3547` (v1.0.0) — `directed=` is absent:

```python
G = _build_merge(
    [merged],
    graph_path=existing_graph_path,
    prune_sources=_prune_sources or None,
    dedup=True,
    dedup_llm_backend=dedup_backend,
    root=target,
)
```

`graphify/build.py` — `def build_merge(..., *, directed: bool = False, ...)`.

Same path is reached by `graphify hook install` (post-commit) and `graphify --watch`, so on a directed graph those revert it on every commit / file change.

## Suggested fix

The robust version is to make `build_merge` inherit by default, since it already reads `graph_path` and so is the only place that knows the answer:

```python
def build_merge(..., *, directed: bool | None = None, ...):
    # directed=None -> inherit from the existing graph; explicit True/False still wins.
    if directed is None:
        directed = bool(data.get("directed", False)) if graph_path.exists() else False
```

That fixes every caller at once and makes "an existing directed graph stays directed" the default everywhere, which seems like the least surprising behaviour.

The minimal alternative is to thread it at the call site in `cli.py`, reading the flag off `existing_graph_path` before the merge.

Either way it would help to print the resolved mode (`[graphify update] directed=True`) so a future downgrade is visible in the log rather than silent.

## Related

- #1574 — `build_merge` / `--update` drops the existing graph's hyperedges. Same shape: graph-level state lost on merge while nodes/edges survive.
- #801 — preserved per-edge direction through `build_merge`; this is the remaining half (the graph type itself).
- #1606 — `merge-graphs` on mixed directed/undirected inputs.

MANOJ21K:
Fixes #2342.

## Problem

An incremental refresh silently rebuilds a `--directed` graph as undirected, and
writes it back with `force=True`. Nothing in stdout or the report says the graph
changed type.

Betweenness centrality is only meaningful on a directed graph. Undirected, a leaf
utility that everything imports becomes a fake bridge: every pair of its importers
sits two hops apart "through" it, though no control flow passes through it at all.
So `## God Nodes` and the betweenness-derived `## Suggested Questions` quietly start
reporting ubiquity instead of architecture.

## Root cause

Two paths drop the persisted graph type. Both re-serialize with `force=True`, so the
downgrade is permanent:

- `build.py::build_merge()` took `directed: bool = False`, and `cli.py`'s incremental
  extract calls it without `directed=` (`_build_merge([merged], graph_path=..., ...)`).
- `watch.py::_rebuild_code()` — the function behind `graphify update` and the
  post-commit hook — called `build_from_json(result)` with no `directed=` at all.

The flag already round-trips correctly everywhere else. `cluster-only` inherits it
(`cli.py`, `build_from_json(_raw, directed=_directed)`), and #801 went to the trouble
of preserving per-edge `source`/`target` through the merge. It is specifically the
graph *type* that was dropped.

## Fix

Inherit the persisted flag instead of defaulting it.

- `_load_existing_graph()` already parses the raw JSON dict for the merge, so it now
  returns `directed` alongside `(nodes, edges, hyperedges)` — no second read of a
  graph.json that may be up to the 512 MiB cap.
- `build_merge(directed=...)` becomes `bool | None`, defaulting to `None` = "inherit
  from the graph on disk". An explicit `True`/`False` still wins, so existing callers
  are unaffected. With no graph on disk there is nothing to inherit and the historical
  `False` default stands.
- `_rebuild_code()` passes `directed=` from `existing_graph_data`, which it has already
  loaded and uses two lines later for the topology compare.

## Reproduction

The script from the issue, before and after:

```
$ python repro.py          # before
saved directed: True
after update  : False

$ python repro.py          # after
saved directed: True
after update  : True
```

## Tests

Two regression tests, both failing on clean `v8` (`assert G2.is_directed()` → `assert False`,
`assert rebuilt["directed"] is True` → `assert False is True`), plus two guards that pin the
behaviour this must *not* change:

- `tests/test_build.py::test_build_merge_inherits_persisted_directed_flag` — merge an
  empty chunk set into a directed graph; asserts the graph stays a `DiGraph`, that the
  re-written JSON keeps `"directed": true`, and that the sink's betweenness is `0.0`
  (the symptom the issue is actually about).
- `tests/test_build.py::test_build_merge_directed_flag_still_forceable` — explicit
  `directed=True`/`False` still overrides the inherited value.
- `tests/test_build.py::test_build_merge_defaults_undirected_without_existing_graph` —
  pins the no-existing-graph default.
- `tests/test_watch.py::test_rebuild_code_preserves_directed_graph` — full
  `_rebuild_code` round trip: rebuild a directed graph after a corpus change and assert
  the persisted flag survives.

`uv run pytest tests/ -q` → 3903 passed, 3 skipped (baseline on `v8` is 3899 passed,
3 skipped — the delta is exactly these four tests, no regressions).

## Out of scope

`extract --no-cluster` / `update --no-cluster` raw-dumps a graph payload that has no
`directed` key at all, so that path loses the flag by a different mechanism (the raw
dict is assembled from the fresh extraction, not from the loaded graph). It needs its
own decision about carrying the key forward through `merge_raw_extraction`, and
touching it here would drag the shrink-guard's `same_graph` comparison into this diff.
Happy to open it as a separate issue if useful.

### DIFF
--- a/graphify/build.py
+++ b/graphify/build.py
@@ -1124,9 +1124,14 @@ def deduplicate_by_label(nodes: list[dict], edges: list[dict]) -> tuple[list[dic
     return deduped_nodes, deduped_edges
 
 
-def _load_existing_graph(graph_path: Path) -> "tuple[list, list, list] | None":
-    """Load (nodes, edges, hyperedges) from an existing graph.json for an
-    incremental merge, accepting both the ``links`` and ``edges`` spellings.
+def _load_existing_graph(graph_path: Path) -> "tuple[list, list, list, bool] | None":
+    """Load (nodes, edges, hyperedges, directed) from an existing graph.json for
+    an incremental merge, accepting both the ``links`` and ``edges`` spellings.
+
+    ``directed`` is the persisted graph type, so an incremental merge can rebuild
+    the same kind of graph instead of silently downgrading a directed graph to an
+    undirected one (#2342). Missing key reads as False, matching the
+    ``cluster-only`` round-trip in cli.py.
 
     Reads the JSON directly instead of going through node_link_graph().
     The latter rebuilds an undirected nx.Graph and then enumerating
@@ -1156,6 +1161,7 @@ def _load_existing_graph(graph_path: Path) -> "tuple[list, list, list] | None":
         list(data.get("nodes", [])),
         list(data.get(links_key, [])),
         list(data.get("hyperedges", [])),
+        bool(data.get("directed", False)),
     )
 
 
@@ -1193,7 +1199,7 @@ def merge_raw_extraction(
     loaded = _load_existing_graph(graph_path)
     if loaded is None:
         return new
-    existing_nodes, existing_edges, existing_hyperedges = loaded
+    existing_nodes, existing_edges, existing_hyperedges, _existing_directed = loaded
 
     _eff_root = (
         str(Path(root).resolve()) if root is not None
@@ -1260,7 +1266,7 @@ def build_merge(
     graph_path: str | Path | None = None,
     prune_sources: list[str] | None = None,
     *,
-    directed: bool = False,
+    directed: bool | None = None,
     dedup: bool = True,
     dedup_llm_backend: str | None = None,
     root: str | Path | None = None,
@@ -1273,17 +1279,23 @@ def build_merge(
     preserved unchanged; deleted files are removed via prune_sources.
     Safe to call repeatedly.
     root: if given, absolute source_file paths in new_chunks are made relative (#932).
+    directed: None (default) INHERITS the loaded graph's type, so a routine
+    incremental refresh can't silently rebuild a directed graph as undirected and
+    turn betweenness into a ubiquity score (#2342). Pass True/False to force it.
     """
     graph_path = Path(graph_path if graph_path is not None else _default_graph_json())
     _loaded = _load_existing_graph(graph_path)
     if _loaded is not None:
-        existing_nodes, existing_edges, existing_hyperedges = _loaded
+        existing_nodes, existing_edges, existing_hyperedges, existing_directed = _loaded
         had_graph = True
     else:
         existing_nodes = []
         existing_edges = []
         existing_hyperedges = []
+        existing_directed = False
         had_graph = False
+    if directed is None:
+        directed = existing_directed
 
     # Effective root for relativizing absolute source_file / prune paths back to the
     # stored relative source_file keys. When the caller passes root we use it;
--- a/graphify/watch.py
+++ b/graphify/watch.py
@@ -1258,7 +1258,11 @@ def _rebuild_code(
             "total_words": detected.get("total_words", 0),
         }
 
-        G = build_from_json(result)
+        # Inherit the persisted graph type. Rebuilding a graph that was written
+        # with --directed as an undirected one silently breaks betweenness — a
+        # pure sink every module imports scores as the top god node — and nothing
+        # in the report says the graph changed type (#2342).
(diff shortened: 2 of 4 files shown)


safishamsi:
Closing as superseded. The directed-flag inheritance for #2342 shipped in v0.9.32 via #2354. Thanks for the fix, @MANOJ21K.

Issue #2058, opened 2026-07-20
Silent data loss: env/*_env directories pruned as false-positive venvs, with zero trace in any report bucket


oz-pressler-sndk:
## Summary

`detect.py`'s directory-noise heuristic (`_SKIP_DIRS` / `_is_noise_dir`) treats any directory literally named `env` (or ending in `_env`) as a Python virtualenv and silently prunes it during the walk — with **zero trace** in any reportable bucket (`ignored`, `unclassified`, or any warning). This causes large, silent, undetectable data loss for any codebase that uses `env`/`*_env` as a real source directory name — a naming convention common in hardware/ASIC verification environments (UVM testbenches), and plausibly in other domains too.

## Root cause

```python
# graphify/detect.py
_SKIP_DIRS = {
    "venv", ".venv", "env", ".env",
    ...
}

def _is_noise_dir(part: str, parent: "Path | None" = None) -> bool:
    if part in _SKIP_DIRS:
        return True
    ...
    if part.endswith("_venv") or part.endswith("_env"):
        return True
    ...
```

This check is applied unconditionally to every subdirectory encountered during the `os.walk` traversal (`detect()`, around the `kept_dirs`/`dirnames[:]` pruning loop), **before** the `_is_ignored()` / `.graphifyignore` negation logic runs. Because of this ordering:

1. A `.graphifyignore` negation (`!env/`) **cannot** rescue the directory — pruning happens earlier and unconditionally.
2. Pruned directories are never added to `ignored`, `unclassified`, or any other bucket in the `detect()` return value — there is no way to detect the loss from the tool's own output. The only symptom is a lower file/node count than expected, with nothing pointing at why.
3. The exact-name check (`env`) applies to *every* subdirectory the walker descends into, not just files reached directly — so even pointing `detect()`/`extract` directly at a parent directory (e.g. `src_env/`) does not protect a nested `env/` subdirectory from being pruned.
4. The suffix check (`_env`) additionally prunes any directory whose *own* name ends in `_env` (e.g. `src_env/`) the moment it's reached as a subdirectory from a parent scan — even though it's not literally named `env`.

## Reproduction

Any repo with this structure:
```
repo/
  src_env/          # <- pruned entirely when scanned from repo root (matches *_env)
    env/            # <- pruned even if src_env/ itself is the scan root (exact match)
      my_class.py
    other_dir/
      also_real.py  # <- also lost, since the whole src_env/ subtree is pruned from the root scan
```

```python
[...]

Issue #870, opened 2026-05-14
coverage/*.html should be filtered by directory, not extension


akanoodles:
# `coverage/*.html` should be filtered by directory, not extension

## Repo
`safishamsi/graphify` — https://github.com/safishamsi/graphify

## Summary

The existing noise filter excludes the `coverage/` directory at file-collection time, so .ts / .js source files inside `coverage/` (rare, but they happen if someone ships a coverage-aware util there) get correctly skipped. However, the *generated HTML files* produced by Vitest / Istanbul / nyc — `coverage/lcov-report/<source-path>.html` — are categorised as `document` and sent to the semantic-extraction subagents. There's typically one HTML file per source file under the report tree, so a project with 800 source files generates ~828 noise documents.

Concrete numbers from one repo on 2026-05-14:
- 1,277 documents detected in total
- 828 of those (65%) were `coverage/lcov-report/<x>.html`
- The 828 files are mechanically-generated, contain only the coverage report markup for an existing source file, and produce no useful semantic information when extracted

After manual filter:
- Document count dropped from 1,277 → 449
- Semantic subagent dispatch dropped from ~64 → 37
- Net token saving on a single corpus build: ~27 subagents × ~100K tokens = ~2.7M tokens

## Repro

Run `/graphify .` against any Vitest-based monorepo that has run tests recently (`coverage/` directory present). Inspect `.graphify_detect.json`:

```bash
$(cat graphify-out/.graphify_python) -c "
import json
from pathlib import Path
d = json.loads(Path('graphify-out/.graphify_detect.json').read_text())
docs = d.get('files', {}).get('document', [])
in_coverage = sum(1 for f in docs if '/coverage/' in f or '/lcov-report/' in f)
print(f'{in_coverage} of {len(docs)} documents are in coverage/')
"
```

You'll see most of them are in coverage/.

## Suggested fix

The detect step should apply the directory-noise filter to *all* file categories, not just `code`. Today's behaviour (inferred from observed results — I haven't read your detect.py) appears to be:

```python
# Current behaviour (inferred)
if category == 'code' and any(noise_dir in path for noise_dir in NOISE_DIRS):
    skip
```

What it should be:

```python
# Suggested
if any(noise_dir in path for noise_dir in NOISE_DIRS):
    skip   # regardless of category
```

This is a one-line conceptual fix though it may require adjusting the
`detect()` function's per-category collection logic.

## Workaround (today)

[...]

PR #2351, opened 2026-07-31
fix(detect): gate coverage/ pruning on report artefacts (#2339)

### LINKED ISSUE #2339: coverage/ in _SKIP_DIRS silently drops real Python packages named 'coverage' (same class as #1666 snapshots / #2058 env)
### Summary

`"coverage"` is an unconditional entry in `graphify/detect.py::_SKIP_DIRS`, and `_is_noise_dir` matches directory names at **any depth**. In a Python repo where `coverage` is a legitimate package name, this silently drops real source from the graph — no warning, no entry in `skipped_sensitive`, nothing in the report.

In our repo (~2,700 tracked code files) it removed an entire 5-module production package (`auditor_toolkit/assurance/coverage/`, ~51 KB) that the rest of the codebase imports from. The graph looked complete: the package's *dependents* are present, so queries return plausible neighbours while the package itself simply doesn't exist as nodes. We only found it by diffing `git ls-files` against the distinct `source_file` values in `graph.json`.

### Repro

```python
from pathlib import Path
collect_files(Path("auditor_toolkit/assurance/coverage"))   # -> []
```

The directory holds `__init__.py`, `impact.py`, `inventory.py`, `mapping.py`, `summary.py`. Nothing is gitignored (`git check-ignore` reports none of them), and `.graphifyignore` doesn't mention them.

### Why this is the same bug class you've already fixed twice

The comment on the entry is `# Vitest/Istanbul/nyc HTML reports (#870)` — a JS-ecosystem artefact directory. But the pruning is unconditional and language-agnostic, so it also removes Python/Go/Rust packages that happen to be called `coverage`.

This is exactly the shape of two fixes already in `detect.py`:

- **#1666** — a bare `snapshots` dir is a Jest artefact *only* when it actually holds snapshot files or sits under a JS test root; elsewhere it's a real code namespace, so pruning by name "silently dropped legitimate source from the graph".
- **#2058** — `env`/`.env`/`*_env` is pruned only on real virtualenv evidence (`pyvenv.cfg`, `bin/activate`, `lib/python*`, `conda-meta/`), because otherwise it "silently drops legitimate source with no trace".

`coverage` hasn't been given the same treatment. The existing `_has_venv_markers` / `_JS_SNAPSHOT_TEST_ROOTS` gating helpers are the precedent.

### Suggested fix

Gate `coverage` (and probably `lcov-report`) on actual coverage-artefact evidence rather than the name alone. A generated coverage-report dir is easy to positively identify — e.g. any of `index.html` + `base.css`/`prettify.js` (Istanbul), `lcov.info`, `coverage-final.json`, `clover.xml`, `coverage.xml` — whereas a source package contains `__init__.py` / `*.py` / `*.go` / `mod.rs` and no report artefacts.

A cheaper variant that would also have caught this: treat a directory containing an `__init__.py` (or any recognised source-manifest file) as source regardless of name.

### Workaround note for other users

There is currently **no include-override**: `.graphifyignore` can only ever exclude more, and `.graphifyinclude` was removed in #2112. So a repo hitting this has to rename the package or patch `_SKIP_DIRS` — neither is great when the name is part of a public API.

### Lower-priority observation from the same audit

While diffing coverage we also hit a divergence worth flagging separately: `_parse_gitignore_line` strips inline `#` comments, but **git does not** — git only treats `#` as a comment at line start, so `*_backup_*.py  # temp scripts` is a literal pattern matching nothing for git, while graphify parses it as `*_backup_*.py` and excludes real files. Not necessarily wrong (the stricter reading is arguably more useful), but it means graphify's corpus can be a strict subset of what git tracks in a way that's invisible unless you go looking. Happy to open that as its own issue if useful.

---

Environment: `graphifyy` via pipx, Python 3.12, Linux. Repo is ~2,700 code files + ~2,100 docs, graph ~60k nodes / 125k edges.

MANOJ21K:
Fixes #2339.

## Problem

`"coverage"` is an unconditional entry in `detect.py::_SKIP_DIRS`, and `_is_noise_dir`
matches directory names at any depth. In a repo where `coverage` is a legitimate package
name, the whole package is dropped from the graph — no warning, no `skipped_sensitive`
entry, nothing in the report.

The failure is quiet in the worst way: the package's *dependents* are still there, so
queries return plausible neighbours while the package itself simply doesn't exist as
nodes. The reporter only found it by diffing `git ls-files` against the distinct
`source_file` values in `graph.json`.

## Same bug class as #1666 and #2058

The comment on the entry is `# Vitest/Istanbul/nyc HTML reports (#870)` — a JS-ecosystem
artefact directory — but the pruning is unconditional and language-agnostic.

`detect.py` has already been fixed twice for exactly this shape:

- **#1666** — a bare `snapshots` dir is a Jest artefact *only* when it holds `.snap`
  files or sits under a JS test root; elsewhere it's a real code namespace.
- **#2058** — `env`/`.env`/`*_env` is pruned only on real virtualenv evidence
  (`pyvenv.cfg`, `bin/activate`, `lib/python*`, `conda-meta/`).

`coverage` hadn't been given the same treatment. This PR applies the existing pattern
rather than inventing a new mechanism.

## Fix

- `coverage` moves out of the unconditional `_SKIP_DIRS` set into an evidence-gated
  branch in `_is_noise_dir`, backed by a new `_has_coverage_artifacts()` helper written
  to match `_has_venv_markers()` (same `OSError` guard, same "cannot verify → keep a
  possibly-real code dir" contract when no parent is passed).
- Evidence is a file a coverage tool actually writes: `lcov.info`, `coverage-final.json`,
  `coverage-summary.json`, `clover.xml`, `coverage.xml`, `cobertura-coverage.xml`,
  `jacoco.xml`, `.coverage`, `index.html` — or an `lcov-report/` / `html-report/`
  subtree. That covers lcov, nyc/Istanbul, coverage.py, JaCoCo and Cobertura.
- **`lcov-report` stays unconditional.** It has no false-positive class — no package is
  ever named that — so gating it would only add filesystem probes for no benefit. Only
  the ambiguous bare name is gated, mirroring how `__snapshots__` stayed unconditional
  in #1666 while `snapshots` was gated.

## Reproduction

The reporter's repro, before and after (`auditor_toolkit/assurance/coverage/` holding
`__init__.py`, `impact.py`, `inventory.py`, `mapping.py`, `summary.py`, with a genuine
Istanbul report at `webapp/coverage/` in the same tree):

```
# before
direct target : []
via repo walk : []

# after
direct target : ['__init__.py', 'impact.py', 'inventory.py', 'mapping.py', 'summary.py']
via repo walk : ['auditor_toolkit/assurance/coverage/__init__.py', ... 'summary.py']
```

The generated `webapp/coverage/` report is still pruned in the "after" walk — its
`prettify.js` never enters the corpus — so #870 is not regressed.

## Tests

`tests/test_detect.py`, following the `test_detect_keeps_snapshots_code_namespace`
precedent. Three of them fail on clean `v8`:

- `test_detect_keeps_coverage_code_namespace` — a `coverage/` package with real modules
  survives a `detect()` scan.
- `test_collect_files_keeps_coverage_code_namespace` — the issue as reported, via
  `collect_files`, both as the walk target and through the repo root, with a real report
  dir alongside that must still be skipped.
- `test_is_noise_dir_coverage_is_evidence_gated` — the gate directly: name alone isn't
  enough, artefact evidence prunes, no-parent keeps, `lcov-report` still unconditional.
- `test_detect_skips_coverage_dir_by_lcov_info` — guard: a single artefact file with no
  `lcov-report/` subtree is still enough to prune.

The existing `test_detect_skips_coverage_dir` (#870) passes unchanged.

`uv run pytest tests/ -q` → 3903 passed, 3 skipped (baseline on `v8` is 3899 passed,
3 skipped — the delta is exactly these four tests, no regressions).

## Noted but not fixed here

The reporter also flagged that `_parse_gitignore_line` strips inline `#` comments while
git only treats `#` as a comment at line start, so `*_backup_*.py  # temp scripts`
excludes real files for graphify but matches nothing for git. That's a separate
divergence in a different function and belongs in its own issue/PR.

### DIFF
--- a/graphify/detect.py
+++ b/graphify/detect.py
@@ -795,7 +795,9 @@ _SKIP_DIRS = {
     ".tox", ".nox", ".eggs", "*.egg-info",  # nox is tox's successor, same .nox/ venv shape (#1804)
     "graphify-out", GRAPHIFY_OUT_NAME,  # never treat own output as source input (#524); honour GRAPHIFY_OUT (#1423)
     # Coverage/test-artefact dirs — generated, never architecturally meaningful
-    "coverage", "lcov-report",              # Vitest/Istanbul/nyc HTML reports (#870)
+    "lcov-report",                          # Vitest/Istanbul/nyc HTML reports (#870);
+                                            # bare "coverage" is gated on report
+                                            # artefacts below (#2339)
     "visual-tests", "visual-test",          # Playwright/visual-regression bundles (#869)
     "__snapshots__",                        # Jest/Vitest snapshot dir (unambiguous)
     "storybook-static",                     # Storybook production build output
@@ -823,6 +825,39 @@ _SKIP_FILES = {
 # silently dropped legitimate source from the graph (#1666). "__snapshots__" stays
 # unconditionally pruned above; only the ambiguous bare name is gated here.
 _JS_SNAPSHOT_TEST_ROOTS = frozenset({"__tests__", "__test__"})
+
+# Files a coverage tool writes into its own output dir. Any one of them is proof
+# the directory is generated: lcov (lcov.info), nyc/Istanbul (coverage-final.json,
+# clover.xml, the lcov-report/ subtree), coverage.py (coverage.xml, .coverage),
+# JaCoCo/Cobertura (jacoco.xml, cobertura-coverage.xml).
+_COVERAGE_ARTIFACT_FILES = frozenset({
+    "lcov.info", "coverage-final.json", "coverage-summary.json",
+    "clover.xml", "coverage.xml", "cobertura-coverage.xml", "jacoco.xml",
+    ".coverage", "index.html",
+})
+_COVERAGE_ARTIFACT_DIRS = frozenset({"lcov-report", "html-report"})
+
+
+def _has_coverage_artifacts(d: "Path") -> bool:
+    """True only when *d* holds files a coverage tool actually generated.
+
+    ``coverage`` is a legitimate package name (a Python package, a Go/Rust module,
+    a domain namespace), so pruning it by name alone silently drops real source —
+    an entire 5-module package in #2339, with its dependents left in the graph so
+    queries still returned plausible neighbours. Prune it only on real evidence,
+    mirroring the ``snapshots``/``env`` gating (#1666/#2058): a coverage report
+    file, or an Istanbul/lcov HTML report subtree.
+    """
+    try:
+        for name in _COVERAGE_ARTIFACT_FILES:
+            if (d / name).is_file():
+                return True
+        for name in _COVERAGE_ARTIFACT_DIRS:
+            if (d / name).is_dir():
+                return True
+    except OSError:
+        pass
+    return False
 
 
 def _has_venv_markers(d: "Path") -> bool:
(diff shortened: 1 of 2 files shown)


safishamsi:
Shipped in v0.9.32: the `coverage/` prune is now gated on coverage-report artefacts, so a real package named `coverage/` is kept (#2339). Thanks @MANOJ21K!

Issue #1392, opened 2026-06-19
graphify claude install (v0.8.41): 26 validated bugs in the generated Claude Code skill (interpreter, --directed, shrink-guard, cleanup ordering)


illiakaliuzhnyi-preply:
## Bug bundle: `graphify claude install` skill (v0.8.41) — 26 validated issues in the generated Claude Code skill

The auto-generated Claude Code skill produced by `graphify claude install` (pinned `.graphify_version = 0.8.41`) ships several reproducible bugs in `SKILL.md` and the `references/*.md` runbooks. We vendored this snapshot into a monorepo and an automated reviewer (plus manual checks) surfaced the issues below. **Each was verified against the installed `graphify` v0.8.41 Python source** (`build.py`, `export.py`, etc.), not just read off the markdown.

Reporting upstream because the skill is regenerated by `graphify claude install`, so any downstream edit to the vendored copy is overwritten on the next install/upgrade — the fix has to live in the generator.

Happy to open individual issues or send a PR against the skill template if that's easier; grouping here so the pattern is visible.

### A. Interpreter resolution — generated commands call system `python3`, which can't import `graphify` under `uv tool install`
1. **SKILL.md interpreter bootstrap** — `uv tool run graphifyy python` selects no package and falls through to system `python3`. The skill should resolve and persist the tool's interpreter (e.g. read the shebang of the installed `graphify` bin) — we use `$(cat graphify-out/.graphify_python)`.
2. **references/add-watch.md `--watch`** — runs bare `python3 -m graphify.watch`; not importable under `uv tool install`. Must use the resolved interpreter.
3. **references/exports.md `--mcp`** — `python3 -m graphify.serve` and the generated Claude Desktop config `command` both point at system python; not importable. Must use the resolved interpreter / its absolute path.

### B. Environment variables set but never exported — child Python reads `os.environ` and never sees the override
4. **references/transcribe.md** — `GRAPHIFY_WHISPER_MODEL=base` set as an unexported shell var; transcriber silently uses the default model.
5. **references/transcribe.md** — Step 1 sets `WHISPER_PROMPT` but the transcriber reads `GRAPHIFY_WHISPER_PROMPT`; domain hint silently ignored (name mismatch + missing `export`).

### C. `--directed` not propagated — directed builds silently produce undirected graphs
6. **SKILL.md Steps 4/5** — `build_from_json(extraction)` omits `directed=`, so `--directed` builds undirected.
[...]

PR #2354, opened 2026-07-31
fix(build): inherit the on-disk directed flag in build_merge (#2342)

### LINKED ISSUE #2342: build_merge() defaults to directed=False, so `graphify update` silently downgrades a directed graph to undirected — betweenness becomes meaningless with no warning

Rishet11:
Closes #2342

## The bug

`build_merge(..., *, directed: bool = False, ...)` (`graphify/build.py:1263`) took direction from the caller only, and rebuilt with that default at `build.py:1327` (`G = build(all_chunks, directed=directed, ...)`), discarding the direction of the graph it had just loaded from `graph_path`. `cli.py:3547` is the only call site in the tree and passes no `directed=`, so every incremental `graphify update` through the merge path inherited `False`.

```
built directed:        True
on-disk directed flag: True
after build_merge:     False     <- returned graph is undirected
on-disk after:         True
```

One correction to the report worth stating: `build_merge` does not itself rewrite the flag on disk (that last line stayed `true`), so the downgrade materializes when the caller serializes the returned graph. The returned object is what had to be fixed.

## The fix

- `_load_existing_graph` now returns `(nodes, edges, hyperedges, directed)`, reading the flag out of the JSON it already parses — no second file read.
- `build_merge`s parameter becomes `directed: bool | None = None`. `None` inherits the on-disk flag when a graph exists, and falls back to `False` when there is none, so a brand-new graph behaves exactly as before. An explicit `True`/`False` from a caller still wins over the on-disk flag.
- `merge_raw_extraction`s unpack updated for the new arity; it never builds an `nx.Graph`, so it discards the flag and its behavior is unchanged.

I grepped for a second copy of this defect, since that is usually the situation in this repo. There is only one `build_merge` call site, and `cli.py:1680` already reads `_raw.get("directed", False)` correctly before calling `build_from_json`, so there was nothing else to fix here.

## Verification

- 6 new tests in `tests/test_build.py`: directed on disk with no `directed=` inherits True; undirected stays False; no existing graph file falls back to False; explicit `directed=False` against a directed on-disk graph still yields False; explicit `directed=True` against an undirected one yields True; and an `a -> b` calls edge survives the round trip without becoming readable as `b -> a`.
- Red before the fix: stashing only `build.py`, the two behavioral tests fail (`inherits_directed_flag_from_disk`, `directed_edge_direction_survives_round_trip`). The other four pass either way by design — they are regression guards for behavior that must not change, and I would rather say that than imply all six were red.
- `tests/test_build.py` + `tests/test_build_merge_hyperedges_and_prune.py`: 74 passed.
- Full suite: 3899 passed / 3 skipped / 4 failed. The same 4 fail with `build.py` stashed back to its pre-fix state (env-dependent backend-detection tests in `test_ollama.py` and `test_llm_backends.py`), so the failure set is unchanged — verified by re-running, not assumed.

A fresh-eyes review pass additionally confirmed: both `_load_existing_graph` unpack sites in live code are updated (`watch.py:474` only does an `is None` check, so it is unaffected), no test unpacks it directly, there is no path where a missing file yields `True` or where `existing_directed` is unbound, and `dedupe_edges` already keys on `(source, target, relation)` — direction-sensitive before this change — so letting a graph stay directed does not alter dedup behavior.

### DIFF
--- a/graphify/build.py
+++ b/graphify/build.py
@@ -1124,9 +1124,9 @@ def deduplicate_by_label(nodes: list[dict], edges: list[dict]) -> tuple[list[dic
     return deduped_nodes, deduped_edges
 
 
-def _load_existing_graph(graph_path: Path) -> "tuple[list, list, list] | None":
-    """Load (nodes, edges, hyperedges) from an existing graph.json for an
-    incremental merge, accepting both the ``links`` and ``edges`` spellings.
+def _load_existing_graph(graph_path: Path) -> "tuple[list, list, list, bool] | None":
+    """Load (nodes, edges, hyperedges, directed) from an existing graph.json for
+    an incremental merge, accepting both the ``links`` and ``edges`` spellings.
 
     Reads the JSON directly instead of going through node_link_graph().
     The latter rebuilds an undirected nx.Graph and then enumerating
@@ -1156,6 +1156,7 @@ def _load_existing_graph(graph_path: Path) -> "tuple[list, list, list] | None":
         list(data.get("nodes", [])),
         list(data.get(links_key, [])),
         list(data.get("hyperedges", [])),
+        bool(data.get("directed", False)),
     )
 
 
@@ -1193,7 +1194,7 @@ def merge_raw_extraction(
     loaded = _load_existing_graph(graph_path)
     if loaded is None:
         return new
-    existing_nodes, existing_edges, existing_hyperedges = loaded
+    existing_nodes, existing_edges, existing_hyperedges, _ = loaded
 
     _eff_root = (
         str(Path(root).resolve()) if root is not None
@@ -1260,7 +1261,7 @@ def build_merge(
     graph_path: str | Path | None = None,
     prune_sources: list[str] | None = None,
     *,
-    directed: bool = False,
+    directed: bool | None = None,
     dedup: bool = True,
     dedup_llm_backend: str | None = None,
     root: str | Path | None = None,
@@ -1273,17 +1274,25 @@ def build_merge(
     preserved unchanged; deleted files are removed via prune_sources.
     Safe to call repeatedly.
     root: if given, absolute source_file paths in new_chunks are made relative (#932).
+    directed: if None (default), honor the on-disk graph's own ``directed`` flag
+    when one exists, so an incremental merge can't silently flip a directed
+    graph undirected (#2342). Falls back to False when there is no existing
+    graph to inherit from. An explicit True/False always overrides the on-disk
+    flag.
     """
     graph_path = Path(graph_path if graph_path is not None else _default_graph_json())
     _loaded = _load_existing_graph(graph_path)
     if _loaded is not None:
-        existing_nodes, existing_edges, existing_hyperedges = _loaded
+        existing_nodes, existing_edges, existing_hyperedges, existing_directed = _loaded
         had_graph = True
     else:
         existing_nodes = []
         existing_edges = []
         existing_hyperedges = []
+        existing_directed = False
         had_graph = False
+    if directed is None:
+        directed = existing_directed if had_graph else False
 
     # Effective root for relativizing absolute source_file / prune paths back to the
     # stored relative source_file keys. When the caller passes root we use it;
--- a/graphify/watch.py
+++ b/graphify/watch.py
@@ -675,6 +675,13 @@ def _node_community_map(graph_data: dict) -> dict[str, int]:
 def _canonical_graph_for_compare(graph_data: dict) -> dict:
     canonical = dict(graph_data)
     canonical.pop("built_at_commit", None)
+    # A missing "directed" key means the same thing as "directed": false
+    # everywhere else in the codebase (#2342's --no-cluster path only started
+    # writing the key once it began inheriting it from the existing graph).
+    # Normalise so an old graph.json without the key doesn't register as
+    # "changed" against a freshly-written candidate that now carries
+    # "directed": false explicitly.
+    canonical["directed"] = bool(canonical.get("directed", False))
     for key in ("nodes", "links", "edges", "hyperedges"):
         if key in canonical and isinstance(canonical[key], list):
             canonical[key] = sorted(
@@ -1178,6 +1185,10 @@ def _rebuild_code(
                 **{k: v for k, v in result.items() if k not in ("edges", "nodes")},
                 "nodes": _dedupe_nodes(result.get("nodes", [])),
                 "links": _dedupe_edges(result.get("edges", [])),
+                # Inherit the existing graph's directed flag (#2342) so
+                # `graphify update --no-cluster` can't silently drop it -
+                # `result` (the raw merged extraction) never carries one.
+                "directed": bool((existing_graph_data or {}).get("directed", False)),
             }
             candidate_graph_text = _json_text(candidate_graph_data)
             same_graph = False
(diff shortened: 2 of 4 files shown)


Rishet11:
Important correction to this PR, found by testing it through the real CLI rather than only at the unit level. Pushed as `640e52a`.

**My original fix was incomplete, and it missed the symptom in the issue title.**

`build_merge` was the right diagnosis for the path reached by `graphify <path>`, and that part is verified — I instrumented it and ran the CLI on an existing directed graph:

```
[PROBE] build_merge had_graph=True existing_directed=True resolved=True
AFTER main-path build, directed = True
```

But when I ran `graphify update <dir>` with the same instrumentation, **`build_merge` was never called at all** — zero probe output — and graph.json still went from `directed: true` to `false`. `graphify update` rebuilds through `_rebuild_code` in `graphify/watch.py`, which is an entirely separate path. So the PR as it stood would have closed #2342 while leaving the reported behavior broken.

Two sites there, both now fixed:

- **`watch.py:1261`** (clustered) — `build_from_json(result)` with no `directed=`, so it took the `directed=False` default. `existing_graph_data` was already in scope.
- **`watch.py:1177`** (`--no-cluster`) — `candidate_graph_data` was spread from the raw merged extraction, which never carries a `directed` key, so the written file lost the flag entirely rather than merely defaulting it.

Both inherit from `existing_graph_data` and fall back to `False` when there is no existing graph, matching the `build_merge` semantics. `_canonical_graph_for_compare` now normalizes a missing `directed` key to `False`, so an older graph.json without the key does not read as "changed" against a candidate that writes it explicitly — without that, the first no-cluster update after this change would look like a spurious rebuild.

**On the two tests, because the first version of them was not trustworthy.** They originally edited a function body (`return 1` → `return 2`), which does not change topology — so the `same_topology` check short-circuits before graph.json is rewritten, the seeded flag survives untouched, and the assertion passes *without the fix running*. That also made them order-dependent: they passed 6/6 in isolation but failed when run with the rest of the file. Both now add a new function and a new call edge, and assert the node count actually grew before checking the flag, so a run that did not rewrite the graph fails loudly instead of passing quietly.

Evidence:

- both new tests red with `watch.py` stashed, green with it
- `tests/test_watch.py`: **90 passed**, four consecutive whole-file runs — the configuration the flakiness appeared in
- full suite: 3903 passed / 3 skipped / 4 failed, the same 4 env-dependent backend-detection failures as a clean `v8` tree here
- `ruff` clean; `python -m tools.skillgen --check` → `check OK: 134 artifact(s)`
- end to end, with the rewrite confirmed rather than assumed:

```
BEFORE: directed=True   nodes=3
[graphify watch] Rebuilt: 4 nodes, 5 edges, 1 communities
AFTER : directed=True   nodes=4        (the new function is present)
```

One clarification to the original body while I am here: I wrote that `cli.py:3547` is the only call site "in the tree". That is true of production Python code, but `tools/skillgen/fragments/references/shared/update.md:112` also documents calling `build_merge` from the skill pipeline. It passes `directed=IS_DIRECTED` explicitly, so it is unaffected by the default change — and its own comment cites #1392 for this same bug class, which suggests the flag has been lost on more than one path before.

safishamsi:
Shipped in v0.9.32 (`graphifyy==0.9.32` on PyPI). The on-disk `directed` flag is now inherited on both the clustered and `--no-cluster` rebuild paths (#2342). Credited in the release notes. Thanks @Rishet11!

PR #2357, opened 2026-07-31
fix(export): compare obsidian manifest ownership case-aware on case-insensitive filesystems (#2282)

### LINKED ISSUE #2282: False 'pre-existing file' warning on case-insensitive filesystems (macOS)
On macOS (case-insensitive APFS), `graphify export obsidian` reports notes it created as pre-existing user files:
Neither file exists on disk. The reported names never reach the filesystem.

**Cause:** the graph contains labels that differ only by case. In this repo, 14 such pairs: `AGORA`/`agora`, `Deploy`/`deploy`, `LEAD`/`Lead`, `GET()`/`get()`, `Row`/`row`, `STATUS`/`Status`, and 8 more. On a case-insensitive filesystem `AGORA.md` and `agora.md` are the same path. The export writes one, then the "did I create this file?" check for the other compares by exact name, fails to match the `_1`-disambiguated sibling, and classifies its own output as a user note.

**Impact:** cosmetic here. I verified all 28 labels across the 14 colliding pairs do get a note via the `_1` suffix, so nothing was lost. But the check is unsound: a real user note whose name collides by case with a node label could be misclassified in the other direction.

**Suggested fix:** normalise with `os.path.normcase` (or compare `Path.resolve()`) when deciding whether an output path was created by graphify, and detect case-insensitivity of the target filesystem rather than assuming exact-name equality.

**Environment:** macOS, APFS (case-insensitive), graphifyy installed via `uv tool`, corpus ~930 files / 5359 nodes.

Rishet11:
Closes #2282

## The bug is worse than a false warning — a node loses its note

`_owned_write` (`graphify/export.py:519`):

```python
if target.exists() and rel_name not in _owned:
```

`target.exists()` consults the real filesystem, case-**in**sensitive on APFS/NTFS. `rel_name not in _owned` is an exact-string lookup against the JSON manifest (`export.py:507-511`). A note graphify itself wrote as `AGORA.md` therefore looked pre-existing once a later run computed `agora.md`.

```
run1 nodes: a="AGORA"
  dir:      [AGORA.md, _COMMUNITY_Community 0.md]
  manifest: [AGORA.md, _COMMUNITY_Community 0.md, .obsidian/graph.json]

run2 nodes: b="agora", a="AGORA"
  [graphify] pruned 1 note(s) for nodes no longer in the graph
  [graphify] WARNING: skipped 1 pre-existing file(s) graphify did not create ...: agora.md
  dir:      [AGORA_1.md, _COMMUNITY_Community 0.md]
```

Two nodes, one note. `agora.md` was refused, and the stale-prune (`export.py:766`, `stale = _owned - _written - _skipped`) then **deleted** `AGORA.md` because that name was in neither set. So the reported symptom was cosmetic but the consequence was not.

## The fix

Two independent causes, both addressed:

1. **Case-aware ownership, not blanket lowercasing.** A `.lower()` everywhere would be wrong on ext4, where `Agora.md` and `agora.md` are genuinely two files graphify must keep. So `_detect_case_insensitive_fs(out)` probes once per `to_obsidian` call by writing a sentinel and checking whether its case-flipped name exists, cleans up after itself, and assumes case-**sensitive** on `OSError` — the conservative default that preserves today’s Linux behavior. Ownership lookups and the stale-prune are keyed through that; the manifest itself still stores real filenames, so it stays valid for case-sensitive filesystems and future runs.
2. **Stable filenames.** `_dedup_node_filenames` (`export.py:463-482`) recomputed `_N` suffixes from graph iteration order, so a node’s filename could drift between runs — that drift is what made the collision fire. It now iterates in sorted node order.

The protection itself is unchanged: a genuinely user-authored file with a colliding name is still skipped and still warned about, and there is a test for that.

I checked `export.py`, `callflow_html.py`, `exporters/html.py` and `report.py` for other copies of this pattern — the owned-file manifest exists only in `to_obsidian`, so unlike most bugs in this area there is no second site. `to_canvas` is untouched.

## Known limitation I am not fixing here, stated plainly

Sorted iteration stabilizes suffix assignment for an **unchanged** node set. It does not make filenames stable when the node set churns. Adding a node whose id sorts earlier than an existing node with a colliding label hands it the un-suffixed name and bumps the other to `_1`:

```
run1 (node zzz, label "Agora"):            Agora.md   -> zzz
run2 (add node aaa, same label "Agora"):   Agora.md   -> aaa
                                            Agora_1.md -> zzz
```

Both notes exist and no content is lost, but the filename-to-node identity shifts, which matters for a vault where the user has their own links pointing at `Agora.md`. That is a property of deriving filenames from labels and predates this change (before it, the same churn additionally deleted a note). Closing it properly means persisting a node-id-to-filename mapping in the manifest, which is a bigger change than #2282 asks for — happy to do it as a follow-up if you want it.

## Verification

- 58 tests pass in `tests/test_export.py`, including new ones for: the same node id whose label changes case between runs (the true fold repro — no warning, note survives, content is graphify’s); a genuinely user-authored colliding file still skipped and warned; filename stability across runs; and the fold applying only when the probe reports case-insensitive.
- Non-vacuity proven by mutation, not by assertion counting. Replacing the keying with the raw name makes both fold tests fail; reverting makes them pass. I checked this because an earlier version of these tests passed even with the fold disabled — they were pinning only the sort-order change, and I retitled the one that genuinely covers only that so it does not overclaim.
- Full suite: 3901 passed / 3 skipped / 4 failed, and the same 4 fail on a clean `v8` tree here (1 in `test_llm_backends.py` from a missing `openai` module, 3 env-dependent ollama backend-detection tests). Failure set unchanged.
- Pre-existing tests at `tests/test_export.py:511`, `:558`, `:596` still pass.

A fresh-eyes review pass confirmed the probe’s mkdir ordering, cleanup, and `OSError` fallback are correct, that the sentinel cannot collide with a note name or leak into the manifest, that every ownership set is keyed consistently, and that the manifest is still written with real filenames. It is also what surfaced both the weak tests and the churn limitation above.

### DIFF
--- a/graphify/export.py
+++ b/graphify/export.py
@@ -470,7 +470,14 @@ def _dedup_node_filenames(G: nx.Graph, safe_name) -> dict[str, str]:
     silently overwrites a node whose literal label is already "base_1"."""
     node_filenames: dict[str, str] = {}
     used: set[str] = set()
-    for node_id, data in G.nodes(data=True):
+    # #2282: iterate in a stable order (sorted node id) rather than graph iteration
+    # order. nx.Graph iteration order isn't guaranteed stable across process runs
+    # for the same node set, which let a node's `_N` suffix drift between runs of
+    # to_obsidian and defeat the manifest-based ownership check in _owned_write
+    # (a node's filename changing run-to-run makes its old note look orphaned and
+    # its new name look "pre-existing").
+    for node_id in sorted(G.nodes(), key=str):
+        data = G.nodes[node_id]
         base = safe_name(data.get("label", node_id))
         candidate = base
         n = 1
@@ -480,6 +487,26 @@ def _dedup_node_filenames(G: nx.Graph, safe_name) -> dict[str, str]:
         used.add(candidate.lower())
         node_filenames[node_id] = candidate
     return node_filenames
+
+
+def _detect_case_insensitive_fs(out: Path) -> bool:
+    """#2282: probe whether `out` sits on a case-insensitive filesystem
+    (macOS/APFS, Windows/NTFS) by writing a sentinel file and checking whether its
+    case-flipped name also `.exists()`. Assume case-SENSITIVE (the conservative
+    default that preserves today's Linux/ext4 behavior) if the probe can't run for
+    any filesystem reason. Module-level so tests can monkeypatch it to force a
+    result without needing a real filesystem of the opposite kind."""
+    probe = out / ".graphify_case_probe.tmp"
+    flipped = out / ".graphify_CASE_probe.tmp"
+    try:
+        out.mkdir(parents=True, exist_ok=True)
+        probe.write_text("", encoding="utf-8")
+        try:
+            return flipped.exists()
+        finally:
+            probe.unlink(missing_ok=True)
+    except OSError:
+        return False
 
 
 def to_obsidian(
@@ -512,11 +539,27 @@ def to_obsidian(
     _written: list[str] = []
     _skipped: list[str] = []
 
+    # #2282: target.exists() consults the real filesystem, which is
+    # case-INsensitive on APFS/NTFS, while a plain `rel_name in _owned` check is an
+    # exact-string lookup. On such a filesystem a node's note written as
+    # "AGORA.md" in one run looks like someone else's pre-existing file when a
+    # later run computes "agora.md" for the same node - it gets skipped, and then
+    # deleted by the stale-prune below because it's in neither _written nor
+    # _skipped. Probed once per call (not per file) and only applied when the
+    # filesystem is actually case-insensitive, so ext4/Linux keeps its existing
+    # case-sensitive behavior of treating "Agora.md" and "agora.md" as distinct.
+    _case_insensitive = _detect_case_insensitive_fs(out)
+
+    def _own_key(rel_name: str) -> str:
+        return rel_name.lower() if _case_insensitive else rel_name
+
+    _owned_keys = {_own_key(f) for f in _owned}
+
     def _owned_write(rel_name: str, content: str) -> bool:
         """Write a graphify-owned file, refusing to overwrite a pre-existing file
         graphify didn't create. Returns True if written."""
         target = out / rel_name
-        if target.exists() and rel_name not in _owned:
+        if target.exists() and _own_key(rel_name) not in _owned_keys:
             _skipped.append(rel_name)
             return False
         target.parent.mkdir(parents=True, exist_ok=True)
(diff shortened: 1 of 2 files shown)


Rishet11:
CI caught a real problem in my own tests, now fixed in `4df69c2`.

The production change was fine, but four of my new tests baked in macOS/APFS behavior. On the Linux runners `AGORA.md` and `agora.md` genuinely coexist, so the collision they asserted never happens — and two of them monkeypatched the case-sensitivity probe to `True` on a case-sensitive filesystem, which is incoherent: it tells the code to fold two names that really are two files.

They now derive their expectations from the actual filesystem instead of assuming one:

- the two-run label-case test forces the probe to the *real* measured value and asserts the invariant that holds everywhere (no false "pre-existing" warning, the node keeps a note, the note is graphify's);
- the user-file-protection test now creates the pre-existing file under the exact name graphify computes, so the collision is real on every filesystem and the test is no longer platform-dependent at all — it should have been written that way from the start;
- the case-sensitive-only and case-insensitive-only scenarios are `skipif`-gated on the real probe, so each runs where it is meaningful and skips with a reason where it is not.

Verified on both, not just reasoned about: I built a real case-sensitive APFS volume with `hdiutil` and ran the export suite with `TMPDIR` pointed at it. `57 passed, 1 skipped` on the case-sensitive volume and `57 passed, 1 skipped` on native APFS, with the skip flipping to the opposite test in each case, which is the check that the gating is doing something rather than quietly disabling coverage. `ruff` clean.

No `xfail`, nothing deleted, and no assertion loosened to make CI green — each test still has a concrete mutation to `export.py` that breaks it.

Rishet11:
Small correction to a number in my PR body: it says "58 tests pass in `tests/test_export.py`". The accurate figure is **57 passed, 1 skipped** (58 collected).

The skip is by design and came in with the follow-up commit `4df69c2`: the case-sensitive-only and case-insensitive-only scenarios are `skipif`-gated on the real filesystem, so exactly one of the pair skips on any given machine. The body was written before that commit and I did not go back and fix the number.

For completeness, measured both ways just now:

```
native APFS (case-insensitive):   57 passed, 1 skipped
  SKIPPED tests/test_export.py:754: scenario only exists on a case-sensitive filesystem

case-sensitive APFS volume:       57 passed, 1 skipped
  SKIPPED tests/test_export.py:786: fold is only meaningful on a case-insensitive filesystem
```

The skip flipping to the opposite test between the two runs is the check that the gating selects coverage rather than quietly removing it.

PR #2364, opened 2026-08-01
fix(go): extract package-level var/const as nodes (#2360)

### LINKED ISSUE #2360: Go: package-level var/const declarations are not extracted as nodes (skews query vocabulary in mixed Go/TS repos)
### Environment

- graphifyy 0.9.31 (PyPI, `uv tool install`), macOS Apple Silicon, Python 3.11
- Extraction invoked directly via `graphify.extract.extract(files, parallel=False)`
- Also reproduced end-to-end on a ~1,300-file Go + TypeScript repo (20.7k nodes)

### Summary

The Go extractor does not emit nodes for **package-level `var` or `const` declarations**. `graphify/extractors/go.py`'s `walk()` matches exactly four declaration types:

```python
if t == "function_declaration":   # L209
if t == "method_declaration":     # L223
if t == "type_declaration":       # L256
if t == "import_declaration":     # L325
```

There is no `var_declaration` or `const_declaration` branch, so Go constants and package-level data are absent from the node table entirely. Questions like "where is `ReasonDisallowedObject` defined", "what is in `DefaultBlacklist`", or "who references `CheckQRStatus`" are unanswerable, because the identifier is not in the graph at any confidence level.

This is the same bug class as #1700 (Kotlin/Java enum constants), which was accepted and fixed, applied to Go.

### Reproducer (2 files)

`policy.go`
```go
package demo

var DefaultBlacklist = []string{"vape", "box cutter", "power bank"}

const ReasonDisallowedObject = "DISALLOWED_OBJECT"

const (
	CheckQRStatus         = "qr_status"
	CheckDisallowedObject = "disallowed_object"
)

func Blacklist() []string { return DefaultBlacklist }
```

`use.go`
```go
package demo

func Classify(obj string) string {
	for _, b := range DefaultBlacklist {
		if b == obj {
			return ReasonDisallowedObject
		}
	}
	return CheckQRStatus
}
```

### Current behavior (0.9.31)

Complete node list for both files:

```
policy.go      <- policy.go L1
Blacklist()    <- policy.go L13
use.go         <- use.go L1
Classify()     <- use.go L3
```

`DefaultBlacklist`, `ReasonDisallowedObject`, `CheckQRStatus` and `CheckDisallowedObject` produce no nodes, and the `Classify() -> DefaultBlacklist` reference edge is therefore also absent.

### Expected behavior

Package-level `var` and `const` identifiers get nodes (as Go's own `go/doc` and `gopls` treat them: package-level declared identifiers), with `contains` edges from the file node and `references` edges from functions that use them. Matching the treatment of `type_declaration` would be consistent.

### Why this matters beyond one symbol

Go codebases conventionally express enums, sentinel values, error codes, status strings and policy data as package-level `const` blocks and `var` slices/maps. In our repo, every screening reason code (`DISALLOWED_OBJECT`, `QR_UNREGISTERED`, ...) and every policy list is a Go `const`/`var`, so none of that domain vocabulary exists in the graph.

That has a second-order effect on query quality. The Step 0 constrained vocabulary expansion in `references/query.md` builds `.vocab.txt` from **node labels**. Because Go consts are absent but the TypeScript/JS path (via `engine.py`, which does handle `lexical_declaration`/`variable_declaration`) indexes frontend constants fine, the vocabulary is systematically skewed toward the frontend. In our case, asking about a domain term whose only indexed occurrence was a frontend display constant seeded traversal from UI labels and returned noise, while the backend logic that actually implements it was unreachable through query expansion.

So in mixed Go + TypeScript repos this does not just hide symbols, it biases query seeding away from the backend.

### Suggested fix

Add `var_declaration` / `const_declaration` handling to `walk()` in `graphify/extractors/go.py`, iterating `var_spec` / `const_spec` children and emitting one node per declared name (mirroring how `type_declaration` iterates `type_spec`). Happy to open a PR if the approach looks right.

HaoChiBao:
## Summary
- Emit nodes and `contains` edges for Go package-level `var`/`const` declarations (including parenthesized blocks and multi-name specs), matching how `type_declaration` is handled.
- Add in-file `references` edges when functions use those symbols, and resolve unique-label cross-file uses as `references` (not `calls`) via an opt-in `relation`/`context` on `raw_calls`.
- Fixes #2360 — without this, Go domain constants/policy data are missing from the graph and vocabulary seeding skews toward languages that already index consts.

## Test plan
- [x] `pytest tests/test_go_package_level_vars.py` (issue repro: nodes, blank `_`, in-file + cross-file refs)
- [x] `pytest tests/test_go_builtin_call_targets.py tests/test_multilang.py -k go`
- [x] Broader extract/symbol suite still green (92 passed)
- [ ] Maintainer: spot-check a mixed Go/TS repo that `affected`/`query` can see package-level consts


HaoChiBao:
## Reviewer notes

Happy to adjust anything — writing this up so the change is easy to evaluate.

### Why this exists
`extract_go` only walked `function_declaration` / `method_declaration` / `type_declaration` / `import_declaration`. Package-level `var` / `const` never became nodes, so:
- Go enums / sentinels / policy constants were invisible to `query` / vocab seeding
- Mixed Go+TS repos skewed toward frontend constants (TS already indexes `lexical_declaration`)

Same bug class as the accepted Kotlin/Java enum-constant fix (#1700), applied to Go. Issue has a minimal 2-file repro.

### What changed (3 files)

**1. `graphify/extractors/go.py` (main fix)**
- New `var_declaration` / `const_declaration` branch in `walk()`, mirroring `type_declaration` → `type_spec`
- One node per declared name (`var_spec` / `const_spec`, including parenthesized blocks and `a, b = 1, 2`)
- IDs use `_make_id(pkg_scope, name)` like types, so same-package symbols share a stable id across files
- Skips blank `_`
- Function bodies still return early from `walk()`, so function-local `var`/`const` are **not** indexed (module-level only)
- In-file uses → `references` (`EXTRACTED`) via `package_level_nids`
- Cross-file uses → `raw_calls` with `relation: references` / `context: identifier`, resolved by the existing unique-label pass (same god-node guard as calls)

**2. `graphify/extract.py` (tiny, opt-in)**
- Shared cross-file pass used to hardcode `relation: calls` / `context: call`
- Now: `rc.get(relation) or calls` (same for `context`)
- **No other extractor sets `relation` on `raw_calls` today**, so this is a no-op for every other language

**3. `tests/test_go_package_level_vars.py`**
- Covers the #2360 repro end-to-end: nodes + `contains`, multi-name specs, blank `_`, in-file refs, cross-file `Classify → DefaultBlacklist` / consts

### Blast radius
| Surface | Impact |
|---|---|
| Non-Go languages | Unchanged (opt-in `relation` defaults to `calls`) |
| Existing Go calls / builtins | Untouched — call path + `_GO_PREDECLARED_FUNCS` still as before |
| Nested locals | Not emitted (walk never enters function bodies for decls) |
| Cross-file refs | Only unique labels; ambiguous names still skipped |

Residual tradeoff (same class as existing cross-file call resolution): a local identifier that uniquely matches a package-level symbol name elsewhere can get an `INFERRED` `references` edge. Happy to tighten (e.g. exported-only / same-package heuristic) if you prefer.

### Suggested review path
1. `graphify/extractors/go.py` — the new `var_declaration`/`const_declaration` block (~30 lines), then the `identifier` arm in `walk_calls`
2. `graphify/extract.py` — the 5-line `relation`/`context` default
3. Skim `tests/test_go_package_level_vars.py::test_go_cross_file_const_reference_edge` — that's the issue repro

```bash
pytest tests/test_go_package_level_vars.py tests/test_go_builtin_call_targets.py -q
```

Thanks for taking a look.

PR #2369, opened 2026-08-01
fix(extract): resolve Python cross-file type references instead of ghosting them (#2363)

### LINKED ISSUE #2363: Python AST extraction creates duplicate 'ghost' nodes for cross-file class references (dataclasses, type annotations)
## Summary

Running the graphify skill's Python AST extraction (`graphify.extract.extract_python`) on a corpus where a dataclass/class is defined in one file and referenced (imported + used in a type annotation, e.g. a function return type or another dataclass's field type) from a different file produces **two separate nodes with the same label** — one at the defining file's correct node ID, and one "ghost" duplicate anchored at the referencing file, instead of a single node with an `imports`/`references` edge pointing at it.

This inflates degree/centrality metrics (a class can show up as a "god node" partly due to its own duplicate's edges being counted separately) and fragments community detection (each ghost duplicate can end up isolated in its own thin community).

## Reproduction

A minimal repro, mirroring what I hit:

```python
# pkg/a/base.py
from dataclasses import dataclass

@dataclass
class PricePoint:
    value: float

# pkg/b/consumer.py
from pkg.a.base import PricePoint

def latest(rows: list[PricePoint]) -> PricePoint | None:
    ...
```

Run graphify's extraction over both files. Expected: one node for `PricePoint` (anchored at `pkg/a/base.py`), with an `imports`/`uses` edge from `consumer.py`'s file node (and/or from `latest()`) to it.

Observed (real corpus, not the minimal repro above — I didn't isolate a standalone reproduction script, but the pattern below is exactly this shape): I got **5 separate nodes all labeled `PricePoint`** across a 5-file corpus where one file defines it and 4 others import + reference it in type annotations:

```
agri_baseline_pricepoint                                  <- correct, at the defining file
signal_intelligence_connections_prices_py_pricepoint      <- ghost duplicate
signal_intelligence_fronts_agri_deviation_py_pricepoint   <- ghost duplicate
signal_intelligence_fronts_upstream_watch_signals_py_pricepoint <- ghost duplicate
signal_intelligence_signal_resolution_py_pricepoint       <- ghost duplicate
```

The 4 ghost duplicates all share one wrong-ID pattern: `signal_intelligence_<full_relative_path_with_underscores>_py_pricepoint` — i.e. the **full path from the project root** gets baked into the ID, while the correct/defining node instead uses the short `{immediate_parent_dir}_{filename_stem}_{entity}` scheme (`agri_baseline_pricepoint`). These are two different ID-generation schemes colliding.

I found **7 more instances of the same pattern** in the same run, but with the *other* malformed variant — a **bare, unprefixed lowercase name** instead of a fully-qualified one:

```
BaselineResult -> agri_baseline_baselineresult (correct)  vs  baselineresult (ghost, from a different file's type annotation)
GapClass       -> ledger_quality_gapclass (correct)       vs  gapclass (ghost, referenced from a file in a different top-level subdirectory)
DayRecord      -> agri_ingest_dayrecord (correct)          vs  dayrecord (ghost)
DeviationRecord-> agri_deviation_deviationrecord (correct) vs  deviationrecord (ghost)
GateResult     -> ledger_risk_gateresult (correct)         vs  gateresult (ghost)
ChokepointDay  -> upstream_watch_ingest_chokepointday (correct) vs  chokepointday (ghost)
PotashMonth    -> upstream_watch_ingest_potashmonth (correct)   vs  potashmonth (ghost)
```

All 11 ghost duplicates were verified against source: every one of these classes is defined exactly once (confirmed via `grep -n "^class "`), and the "extra" nodes are 100% attributable to a different file's `import` + type-annotation usage, never a genuine second definition.

## What I ruled out

I confirmed this is **not** a semantic-subagent/LLM extraction issue: I re-ran the semantic (Claude subagent) extraction pass with an explicit canonical-ID glossary telling subagents exactly which ID to use for each of these 24 classes, and the corrected subagent output had **zero** duplicate nodes for any of them. I then diffed the subagents' new output against `.graphify_ast.json` (the structural, non-LLM extraction) and found all 11 duplicate IDs originate purely from the AST pass, confirming the bug is in `extract_python`'s (or the shared cross-file resolution pass around `resolve_exported_origin`/`facts.uses` in `extract.py`) handling of Python — not the LLM layer.

## Suspected area

`graphify/extract.py`:
- `_python_collect_type_refs` (~line 477) walks a Python type annotation and appends `(name, "type")` — just the raw identifier text, not yet resolved to a file.
- The cross-file resolution path around `resolve_exported_origin` / `facts.uses` / `local_aliases_by_file` (~line 6874 onward) looks primarily written for JS/TS ES-module semantics (named exports, star exports) and may not correctly resolve a Python `from X import Y` alias back to `Y`'s defining-file node ID in all cases — falling back to a synthesized ID (either the referencing file's own full-path-qualified stem, or no prefix at all) instead of failing over to a "no edge" or a proper lookup.

I haven't isolated the exact branch producing each of the two different malformed-ID shapes (full-path vs bare), so treat the above as a strong lead rather than a confirmed line number.

## Suggested fix

- When a Python type annotation reference (`_python_collect_type_refs` output) can't be resolved to an existing symbol node via the file's own `from X import Y` aliases, either (a) resolve it properly against the corpus-wide symbol table keyed by `(defining_file, class_name)`, or (b) don't synthesize a new node ID at all — drop the edge rather than create a duplicate node with a fabricated ID.
- Add a post-extraction consistency check: if two nodes share an identical `label` and one of them has zero real definition-site metadata (no `contains` edge from a file node, no docstring, etc.), that's a strong signal one of them is a ghost duplicate worth auto-merging or at least flagging in `GRAPH_REPORT.md`'s "Knowledge Gaps" section.

## Environment

- `graphifyy==0.8.28` (via `uv tool install graphifyy`)
- macOS (Darwin 25.6.0, arm64), Python 3.14 (the tool's own venv)
- Corpus: 20 Python files + 1 markdown doc, real project code (not synthetic), all files under one repo root

Happy to share the full `graphify-out/graph.json` (before my manual post-processing merge) if useful — didn't attach it here since it's from a private-ish personal project, but I can redact/trim it to just the relevant nodes/edges if that helps debug.

Rishet11:
Closes #2363

## What I found

The reported duplication is one defect with two id shapes, plus a second, separate defect that the issue title also names. Both are Python-only gaps against behavior other languages already have.

I first measured the reporter's layout with a single definition of `PricePoint` and four referring files: that comes out clean on `v8` head, one node and four correct edges. The missing ingredient for the ghosts is a **second same-named definition** elsewhere in the corpus. With that, both reported shapes reproduce, and which one you get depends only on how many files refer to it:

```
ambiguous + 1 referring file
   pricepoint                                                     <- bare ghost

ambiguous + 4 referring files
   signal_intelligence_connections_prices_py_pricepoint           <- path-qualified ghosts
   signal_intelligence_fronts_agri_deviation_py_pricepoint
   signal_intelligence_fronts_upstream_watch_signals_py_pricepoint
   signal_intelligence_signal_resolution_py_pricepoint
```

Those are the four ids from the issue, character for character. The mechanism: each referring file mints its own sourceless stub with the bare id `pricepoint`; when there are several, they collide on that id but carry different `origin_file`s, so `_disambiguate_colliding_node_ids` salts each into a path-qualified id **before** `_rewire_unique_stub_nodes` runs. So the "two different ID-generation schemes colliding" the issue describes is really one scheme observed at two corpus sizes.

**1. Class-body field annotations were never collected.** `_python_collect_type_refs` had exactly two call sites, function parameters and return types. Nothing walked a `class_definition` body, so the dataclass shape named in the issue title emitted no edge at all — a dropped edge, not a ghost:

```python
@dataclass
class Quote:
    point: PricePoint    # before: no edge, no node
```

Java and TS both handle their equivalent today. Measured on identical corpora:

```
JAVA  b_quote_quote -> a_pricepoint_pricepoint  references field
TS    b_quote_quote -> a_base_pricepoint        references field
PY    (nothing)
```

**2. An ambiguous class name leaves the reference stuck on a shadow stub.** `_rewire_unique_stub_nodes` bails when the label is not globally unique, so with two modules defining `PricePoint` the sourceless stub survives and the edge stays on it — which is exactly the bare-lowercase ghost variant the reporter hit 7 times:

```
3 PricePoint nodes:
  pkg_a_base_pricepoint   src=base.py
  pkg_z_other_pricepoint  src=other.py
  pricepoint              src=''        <- ghost
  pkg_b_c_f -> pricepoint  references parameter_type
```

`from pkg.a.base import PricePoint` names the defining module exactly. Java resolves this with `_resolve_java_type_references`, PHP and C# with their own resolvers. Python had none — `_collect_python_symbol_resolution_facts` collects `imports` and `calls`, never type references. Java on the identical corpus gets it right:

```
2 PricePoint nodes, no ghost
  b_svc_svc_take -> a_pricepoint_pricepoint  references parameter_type
```

## The fix

Both changes follow existing precedent rather than introducing anything new — no new edge type, no schema, exporter or `affected` changes.

- `engine.py`: a Python class-body walk emitting `references` with context `field` / `generic_arg`, mirroring the Java record-component branch, through the same `_semantic_reference_edge` emitter the parameter/return path already uses (it validates the context against `REFERENCE_CONTEXTS`).
- `resolution.py` + `extract.py`: `_resolve_python_type_references`, mirroring `_resolve_java_type_references`, keyed on the local binding so `import X as Y` resolves. Runs after `_rewire_unique_stub_nodes`, so it only handles the ambiguous remainder.

Two deliberate narrowings, both measured rather than assumed:

- The repoint set omits `imports`, which Java and C# both carry. A Python file-level import edge is resolved upstream and never lands on a bare-name stub — verified against ambiguous, aliased and unresolvable-module corpora — so including it would have been an untested no-op.
- The resolver early-returns unless shadow stubs exist, so its parse pass is only paid on a corpus that actually has unresolved references.

## Before / after, real CLI

`graphify extract . --code-only` on one definition of `PricePoint` in `pkg/a/base.py`, a second in `pkg/z/other.py`, and a `Quote` dataclass importing the first:

Before — the field reference is absent entirely, and `Quote` is isolated:
```
E pkg_a_base -> pkg_a_base_pricepoint contains
E pkg_b_quote -> pkg_a_base_pricepoint imports
E pkg_b_quote -> pkg_b_quote_quote contains
E pkg_z_other -> pkg_z_other_pricepoint contains
```

After — the edge exists and binds to the imported definition, not the other same-named one:
```
E pkg_b_quote_quote -> pkg_a_base_pricepoint uses field
```

## Verification

- 11 tests in `tests/test_python_type_references.py`. Checked out against base, 8 fail and 3 pass — the 3 are the pins and negative controls (the no-import bail-out, the plain `x = 1` attribute, the already-working unambiguous case), which are supposed to be green both sides.
- One of them drives the real CLI and asserts on the written `graph.json`, since a unit test on `extract()` can pass while the shipped pipeline drops the result.
- The fix was staged in halves to confirm each test is load-bearing: the engine change alone flips the field tests, the resolver flips the ambiguity tests.
- Full suite: 4 failed / 3906 passed / 3 skipped. All 4 reproduce with the change reverted (`test_ollama.py` x3 and `test_llm_backends.py`, all reading real env vars), so the failure-set diff is empty.
- `ruff` clean. `python -m tools.skillgen --check` → `check OK: 134 artifact(s)`.

## Known limitations

- A reference whose name is ambiguous and which has **no** import naming its definition is still left on the stub. Guessing one of two equally-plausible classes would invent an edge; there is a test pinning this.
- Star imports (`from pkg.a.base import *`) are not used for disambiguation.
- Wide-corpus perf: the resolver re-parses Python files, like the Java and C# resolvers do. Gated on stubs existing, so a clean corpus pays nothing.

## Relationship to #1231

#1231 targets this same area and has been conflicting since June. I did not build on it — it routes annotations through `_SymbolUseFact`, which is a larger redesign of which layer owns Python type references, and it emits `context="annotation"`, which is not a member of `REFERENCE_CONTEXTS`. Happy to close this if you would rather revive that one.

### DIFF
--- a/graphify/extract.py
+++ b/graphify/extract.py
@@ -118,6 +118,7 @@ from graphify.extractors.resolution import (  # noqa: E402,F401
     _resolve_export_target,
     _resolve_java_type_references,
     _resolve_php_type_references,
+    _resolve_python_type_references,
     _resolve_js_import_path,
     _resolve_js_import_target,
     _resolve_js_module_path,
@@ -5118,6 +5119,13 @@ def extract(
         except Exception as exc:
             import logging
             logging.getLogger(__name__).warning("Cross-file import resolution failed, skipping: %s", exc)
+        # Re-point type-reference edges the bare-name rewire left on shadow stubs
+        # because the label was ambiguous, using the file's own imports (#2363).
+        try:
+            _resolve_python_type_references(paths, root, all_nodes, all_edges)
+        except Exception as exc:
+            import logging
+            logging.getLogger(__name__).warning("Python type-reference resolution failed, skipping: %s", exc)
 
     # Cross-file Java import resolution
     java_paths = [p for p in paths if p.suffix == ".java"]
--- a/graphify/extractors/engine.py
+++ b/graphify/extractors/engine.py
@@ -2510,6 +2510,39 @@ def _extract_generic(
                             base_nid = ensure_named_node(base, line)
                             add_edge(class_nid, base_nid, "inherits", line)
 
+                # Class-body field annotations (`point: PricePoint`), which is the
+                # dataclass shape (#2363). _python_collect_type_refs was only ever
+                # reached from function parameters and return types, so a field type
+                # produced no edge at all — while Java (record components / field
+                # declarations) and TS (`public_field_definition`) both emit
+                # `references`/context="field" for the identical shape. An assignment
+                # without a `type` child is a plain value binding (`plain = 3`), not a
+                # type reference, so it is skipped.
+                body_node = node.child_by_field_name("body")
+                if body_node is not None:
+                    for stmt in body_node.children:
+                        if stmt.type != "expression_statement":
+                            continue
+                        for assign in stmt.children:
+                            if assign.type != "assignment":
+                                continue
+                            type_node = assign.child_by_field_name("type")
+                            if type_node is None:
+                                continue
+                            field_refs: list[tuple[str, str]] = []
+                            _python_collect_type_refs(type_node, source, False, field_refs)
+                            field_line = assign.start_point[0] + 1
+                            for ref_name, role in field_refs:
+                                ctx = "generic_arg" if role == "generic_arg" else "field"
+                                target_nid = ensure_named_node(ref_name, field_line)
+                                if target_nid != class_nid:
+                                    # Same emitter as the parameter/return-type path
+                                    # below, which validates ctx against
+                                    # REFERENCE_CONTEXTS rather than trusting it.
+                                    edges.append(_semantic_reference_edge(
+                                        class_nid, target_nid, ctx, str_path, field_line
+                                    ))
+
             # Swift-specific: conformance / inheritance
             if config.ts_module == "tree_sitter_swift":
                 swift_kind = _swift_declaration_keyword(node) if t == "class_declaration" else "protocol"
--- a/graphify/extractors/resolution.py
+++ b/graphify/extractors/resolution.py
@@ -2252,6 +2252,113 @@ def _resolve_cross_file_java_imports(
 
     return new_edges
 
+def _resolve_python_type_references(
+    paths: list[Path],
+    root: Path,
+    all_nodes: list[dict],
+    all_edges: list[dict],
+) -> None:
+    """Re-point dangling Python type-reference edges at the imported definition.
+
+    Python type annotations resolve by bare name and fall back to a sourceless
+    "shadow" stub (``ensure_named_node``). ``_rewire_unique_stub_nodes`` repairs
+    that only when the label is globally unique; when two modules define the same
+    class name it bails, so the reference stays stuck on the shadow node and both
+    real definitions look unreferenced (#2363). A ``from pkg.a.base import
+    PricePoint`` names the defining module exactly, so it disambiguates where
+    bare-name matching cannot.
+
+    Mirrors ``_resolve_java_type_references`` and the C# resolver: a re-parse
+    pass, here via ``_collect_python_symbol_resolution_facts``, which the JS/TS
+    augment pass also calls earlier in the run. The shadow-stub check below runs
+    first precisely so that second parse is paid only on a corpus that actually
+    has unresolved stubs, which is the abnormal case.
+
+    Mutates ``all_nodes``/``all_edges`` in place. Runs after id-disambiguation so
+    target ids are final, and after ``_rewire_unique_stub_nodes`` so it only has
+    to handle the ambiguous remainder.
+    """
+    py_paths = [path for path in paths if path.suffix == ".py"]
+    if not py_paths:
+        return
+
+    # Sourceless shadow stubs with a type-like label — the only repoint targets.
+    stub_label: dict[str, str] = {
+        node["id"]: node.get("label", "")
+        for node in all_nodes
+        if node.get("id")
+        and not node.get("source_file")
+        and node.get("label", "")[:1].isupper()
+    }
+    if not stub_label:
+        return
+
+    facts = _SymbolResolutionFacts()
+    _collect_python_symbol_resolution_facts(py_paths, root, facts)
+    if not facts.imports:
+        return
+
+    # (defining file, symbol label) -> definition node id.
+    def_by_file_label: dict[tuple[str, str], str] = {}
+    for node in all_nodes:
+        source_file = str(node.get("source_file", ""))
+        label = str(node.get("label", ""))
+        nid = node.get("id")
+        if not (source_file and label and isinstance(nid, str) and nid):
+            continue
+        if not _is_type_like_definition(node):
+            continue
+        def_by_file_label.setdefault((_source_key(source_file, root), label), nid)
+
+    # importing file -> local binding -> definition node id. Keyed on the LOCAL
+    # name so `from pkg.a.base import PricePoint as PP` binds `PP`.
+    alias_by_file: dict[str, dict[str, str]] = {}
+    for imp in facts.imports:
+        if imp.target_path is None or not imp.local_name:
+            continue
+        resolved = def_by_file_label.get(
+            (_source_key(str(imp.target_path), root), imp.imported_name)
+        )
+        if resolved:
+            alias_by_file.setdefault(
+                _source_key(str(imp.file_path), root), {}
+            )[imp.local_name] = resolved
+    if not alias_by_file:
+        return
+
+    # Deliberately NARROWER than the Java/C# repoint sets, which also carry
+    # `imports`: a Python file-level `imports` edge is resolved to the real
+    # definition upstream and never lands on a bare-name stub, so including it
+    # here would be an untested no-op. Verified against ambiguous, aliased, and
+    # unresolvable-module corpora.
+    REPOINT_RELATIONS = {"references", "inherits", "implements"}
+    repointed_from: set[str] = set()
+    for edge in all_edges:
+        if edge.get("relation") not in REPOINT_RELATIONS:
+            continue
+        target = edge.get("target")
+        label = stub_label.get(target)
+        if not label:
+            continue
+        ref_file = _source_key(str(edge.get("source_file", "")), root)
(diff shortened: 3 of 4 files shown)


Rishet11:
Self-review follow-up in c8b3780, correcting two things in what I first pushed.

**The performance note in the PR body was wrong.** I wrote that the resolver's parse pass is "only paid on a corpus that actually has unresolved references" and that "a clean corpus pays nothing". That is not true. Every stdlib or third-party annotation leaves a sourceless stub — this one-file corpus produces two:

```python
import sqlite3
from pathlib import Path

def f(c: sqlite3.Connection, p: Path) -> None: ...
```
```
sourceless stubs: ['connection', 'path']
```

So ordinary code trips the gate and most real corpora do pay the pass, measured at ~15ms/file (1.18s across this repo's own 80 files under `graphify/`). Still bounded and still the same shape as the Java and C# resolvers, which re-parse unconditionally, but I should not have described it as rare.

I tried tightening the gate to "a stub label names a real definition in the corpus", which does shut it on the case above. I reverted it: the stub for `from X import Y as Z` carries the local alias `Z`, no definition is labeled `Z`, so the tighter gate silently disabled alias resolution. My own aliased-import test caught it. Both the real cost and the rejected alternative are now recorded in the docstring rather than left as a comment that reads better than the truth.

**`inherits` was in the repoint set with no test covering it.** Superclasses reach `ensure_named_node` from a different call site than type annotations, so none of the annotation cases exercised that path. Added `test_ambiguous_superclass_resolves_through_the_import`. It passes as written, so it pins existing behavior rather than fixing anything — but shipping an untested branch I had asserted was covered was the gap.

Now 10 tests. Suite 4 failed / 3905 passed, same four env-dependent failures that reproduce with the change reverted; ruff and `skillgen --check` clean.

One thing still open from the issue, which I do not want to paper over: I could not reproduce the **path-qualified** ghost variant (`signal_intelligence_<path>_py_pricepoint`) at all, so this PR does not explain where those four came from. My best guess was that per-file stubs collide and get salted apart by `_disambiguate_colliding_node_ids` before the rewire can collapse them, but every corpus I built that should trigger that came out clean, so I am not claiming it. @SatishRockzz's redacted `graph.json` would settle it. If it turns out to be a live third defect, it wants its own issue rather than being folded in here.

Rishet11:
Closing the loop on the one thing I said I could not explain. **The path-qualified ghosts reproduce, and they share the bare stub's root cause** — 271b130 adds the fixture, PR body updated.

The ingredient I was missing is a **second same-named definition** elsewhere in the corpus. With one, everything resolves cleanly, which is why my earlier repros came out empty. Add a `vendor/legacy.py` that also defines `PricePoint`, and the shape depends only on how many files refer to it:

```
ambiguous + 1 referring file
   pricepoint

ambiguous + 4 referring files
   signal_intelligence_connections_prices_py_pricepoint
   signal_intelligence_fronts_agri_deviation_py_pricepoint
   signal_intelligence_fronts_upstream_watch_signals_py_pricepoint
   signal_intelligence_signal_resolution_py_pricepoint
```

Those are @SatishRockzz's four ids character for character. Each referring file mints its own sourceless stub with the bare id `pricepoint`; with several they collide on that id but carry different `origin_file`s, so `_disambiguate_colliding_node_ids` salts each into a path-qualified id *before* `_rewire_unique_stub_nodes` can collapse them. So the "two different ID-generation schemes colliding" is one scheme seen at two corpus sizes, and the 4-vs-7 split in the report is just which classes had multiple referrers.

That also predicts something checkable in the reporter's corpus: each of those 11 classes should have a second definition of the same name somewhere — a vendored copy, a `tests/` fixture, a compat shim. The issue says each was confirmed defined once via `grep -n "^class "`, which would miss a second definition written as `class PricePoint(Base):` on an indented line, inside `if TYPE_CHECKING:`, or in a file that grep pass did not cover.

No behavior change from this commit — the resolver already handled the case, so it is coverage for a shape I previously could not assert. 11 tests now; 8 fail at base, the 3 that pass are the pins and negative controls. Suite 4 failed / 3906 passed, empty failure-set diff; ruff and skillgen clean. Also fixed the docstring wording Copilot flagged.

PR #2370, opened 2026-08-01
fix(detect): do not prune scan root when explicitly passed even if gitignored (#2358)

### LINKED ISSUE #2358: detect(): scan root itself matched by an ancestor .gitignore prunes nested files but not top-level ones (near-total silent data loss)
## Summary

`detect()` walks up from the scan root to find an ancestor `.git`/`.gitignore` even when the scan root **itself** is an ignored path (e.g. `corpus/` is deliberately listed in the enclosing repo's `.gitignore` because it holds private snapshot data that shouldn't be committed). The result is inconsistent and near-total pruning: files directly inside the scan root survive, but anything nested one level deeper or more is silently dropped — with no warning.

This is distinct from #1873/#1887 (a nested `.gitignore` inside the scan root leaking its patterns outside its own subtree). Here there's no nested `.gitignore` at all — the *only* ignore file involved is the top-level one in the parent repo, and the bug is that it still gets applied to a directory that was explicitly passed as the scan root, in a way that behaves inconsistently by depth.

## Repro

```bash
mkdir -p repo/corpus/subdir && cd repo
git init -q
printf 'corpus/\n' > .gitignore
printf '# top-level file\n' > corpus/manifest.md
printf '# nested file\n' > corpus/subdir/notes.md
printf '# nested file 2\n' > corpus/subdir/more.md

python3 -c "
from pathlib import Path
result = detect(Path('corpus'))
print('total_files:', result.get('total_files'))   # 1 — expected: 3
for cat, files in result.get('files', {}).items():
    if files:
        print(cat, files)
"
# document ['.../repo/corpus/manifest.md']
# corpus/subdir/notes.md and corpus/subdir/more.md are missing — no warning, no skip count.
```

## Expected

Either:
- `detect()` should not honor an ancestor `.gitignore` pattern that matches the scan root path itself (you explicitly pointed at this directory — that's a stronger signal than an inherited ignore rule), or at minimum
- it should behave consistently: currently top-level files in the scan root survive while nested ones don't, which is confusing and looks like partial success rather than a total-exclusion bug.

If exclude-by-design is intended, `detect()` should at least surface a `skipped_gitignored` count (or similar) so a near-empty result isn't silently mistaken for "this folder has almost no supported files."

## Impact hit in the wild

Real project: a Graphify knowledge-graph automation snapshots private source files into `corpus/` inside its own git repo, with `corpus/` (and `graphify-out/`) in `.gitignore` on purpose (so the snapshot/output never gets committed). `detect(Path('corpus'))` returns `total_files: 1` (a stray `.json` at the corpus root) instead of the real ~230 files. The project's own manifest-based snapshotter (independent of `detect()`) confirms 230 files are actually present.

This also breaks `graphify update <corpus>` on the same corpus (not just `detect()` directly) — the incremental updater's internal scan hits the same pruning, decided that ~207 previously-known files "left the scan corpus," and tried to shrink the graph from 1537 nodes down to 212 before the shrink-guard (correctly) refused the write. Had the shrink-guard not existed, this would have silently destroyed most of a curated knowledge graph.

## Environment

- graphify (graphifyy) 0.9.28
- Python 3.14.6
- macOS 24.6.0 (Darwin), x86_64
- Installed via `uv tool install graphifyy`

## Workaround

None found short of moving the scan root outside any git repo that ignores it, which isn't viable when the corpus needs to live in the automation's own repo (versioned tooling, gitignored data).

shard-c6:
Fixes #2358

## Problem
When the scan root itself is matched by an ancestor `.gitignore` 
(e.g. `corpus/` in the parent repo's `.gitignore`), `detect()` 
silently prunes all nested files while top-level files survive.

## Fix
In `_eval()`, when the scan root is a descendant of the pattern's 
anchor directory, rebase the relative path to the scan root so 
unanchored patterns only match path components *below* the scan 
root — not the scan root's own name.

## Testing
- Added `test_ancestor_gitignore_matching_scan_root_does_not_prune_nested`
- Full `test_detect.py` suite: 227 passed, 0 failed

### DIFF
--- a/graphify/detect.py
+++ b/graphify/detect.py
@@ -1127,6 +1127,19 @@ def _is_ignored(
                 rel_anchor = str(target.relative_to(anchor)).replace(os.sep, "/")
             except ValueError:
                 continue  # target outside this pattern's anchor: cannot match
+
+            # #2358: when the scan root is a descendant of the anchor (e.g. the
+            # user passed a gitignored `corpus/` directory as the scan target),
+            # rebase the relative path so that only the portion *below* the scan
+            # root is matched.  Without this, unanchored patterns match the scan
+            # root's own directory name in the prefix, silently pruning nested
+            # content while leaving top-level files unaffected.
+            if root != anchor:
+                try:
+                    rel_anchor = str(target.relative_to(root)).replace(os.sep, "/")
+                except ValueError:
+                    pass  # target outside root — keep anchor-relative path
+
             if rel_anchor != ".":
                 matched = _matches(rel_anchor, p, path_relative=path_relative)
                 if matched and directory_only and not target.is_dir():
--- a/tests/test_detect.py
+++ b/tests/test_detect.py
@@ -379,6 +379,36 @@ def test_gitignore_nested_negation_overrides_broader_root_rule(tmp_path):
     assert not any(f.endswith("other.py") for f in code)
 
 
+def test_ancestor_gitignore_matching_scan_root_does_not_prune_nested(tmp_path):
+    """When the scan root itself is listed in an ancestor .gitignore, nested
+    files must not be silently pruned (#2358).  The user explicitly pointed
+    detect() at this directory — that is a stronger signal than an inherited
+    ignore rule.  Before the fix, top-level files survived but anything one
+    level deeper was silently dropped."""
+    repo = tmp_path / "repo"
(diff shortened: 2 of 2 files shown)


shard-c6 (commented):
Fixes #2358 with proper reproducibility of the issue.

Issue #2334, opened 2026-07-31
_origin provenance marker is only stamped by current extract path — legacy AST nodes are misread as semantic


SungKyeol:
## Summary

`_origin: "ast"` is the marker several code paths use to decide whether a node came from AST or
semantic extraction. But it is only stamped by the current extract path (`graphify/extract.py:5433`),
so graphs carrying nodes written by older versions have a large population of **AST nodes with no
`_origin`**. Provenance-aware logic then misclassifies them as semantic.

## Evidence

Real repo, graph built up across 0.4.x → 0.7.10 → 0.9.27:

- `document` nodes total: **12,243**
- with `_origin == "ast"`: **783**
- with no `_origin` at all: **11,460**

All 12,243 are structurally identical — `file_type=document`, `source_location="L<line>"` — i.e. all
of them are markdown-extractor output (`extractors/markdown.py` emits file + heading nodes with
`L<line>`). Semantic extraction emits `source_location: null` per the extraction spec. So the true
AST population is 12,243 and only 6% of it is marked.

## Why it matters

Consumers treat missing `_origin` as "semantic":

- `graphify/watch.py:568-583` — `preserved_nodes` keeps anything without `_origin=="ast"`
- `graphify/watch.py:601` — same for edges
- `graphify/build.py:401`, `:724` — AST-wins-on-collision resolution

In this repo the misclassification happens to fail *safe* (legacy AST nodes are preserved rather than
replaced), but it is unintentional, and it makes the marker unusable as the basis for the tier-scoped
merge fix proposed in #2333. A migration that infers `_origin` where it is missing, or a fallback that
treats `source_location =~ ^L\d` as AST, would make provenance reliable.

I used the `source_location` fallback locally and it partitioned the graph correctly where `_origin`
did not (scenario C vs B in #2333: +2,639 nodes vs -8,290).

## Suggested fix

Either (a) backfill `_origin` on load for nodes matching the AST shape, or (b) make the provenance
predicate a helper that falls back to `source_location` when `_origin` is absent, and use that helper
everywhere instead of the raw `== "ast"` comparison.

## Environment

graphifyy 0.9.27, Python 3.11, Windows 11.

safishamsi:
Fixed in v0.9.32. The `_origin` provenance marker is backfilled on load, so graphs written before this change get tier provenance without a rebuild. Part of the tier-aware merge fix (#2333/#2336). Thanks.

Issue #2333, opened 2026-07-31
build_merge: replace-on-re-extract is provenance-blind — merging a semantic extraction deletes the AST layer for the same files


SungKyeol:
## Summary

`build_merge()`'s replace-on-re-extract drops a source file's **entire** prior contribution, without
distinguishing which extraction tier produced it. AST and semantic extraction are two independent
producers for the same file, so merging a semantic re-extraction deletes that file's AST layer (and
vice versa for any merge-based AST path).

`watch.py::_rebuild_code` already gets this right — it scopes ownership by `_origin` (#1865, "Semantic
nodes lack the AST origin marker and remain preserved"). `build_merge` has no equivalent guard.

## Location

`graphify/build.py:1320-1325` (0.9.27):

```python
def _kept(item: dict) -> bool:
    sf = item.get("source_file")
    return sf not in new_sources and _norm_source_file(sf, _replace_root) not in new_sources
existing_nodes = [n for n in existing_nodes if _kept(n)]
existing_edges = [e for e in existing_edges if _kept(e)]
```

Neither `_origin` nor `source_location` is consulted.

## Measured impact

Real repo, 0.9.27. Base graph 26,548 nodes / 47,564 edges. Merging 1,045 cached semantic chunks
covering 1,043 source files:

| scenario | nodes | edges | `document` | `concept` |
|---|---|---|---|---|
| before | 26,548 | 47,564 | 12,243 | 21 |
| **A — current `build_merge`** | **17,616 (-8,932)** | 39,983 | **1,327 (-10,916)** | 618 |
| B — preserve `_origin=="ast"` | 18,258 (-8,290) | 40,613 | 1,969 (-10,274) | 618 |
| **C — preserve `source_location=~^L\d`** | **29,187 (+2,639)** | **51,933 (+4,369)** | 12,932 (+689) | 604 |

Scenario A silently destroys the document structure layer (every markdown heading node) in exchange
for the semantic nodes. This is the documented `--update` path, so it is reachable from the standard
skill runbook.

Note the `to_json` shrink-guard (#479) does **not** catch this when `dedup=True` (the default) —
shrinkage is treated as intentional there.

## Suggested fix

Scope the replacement to the tier being re-extracted: build `new_sources` per tier and drop only
existing items of that tier, e.g. gate `_kept` on `_origin`/tier rather than `source_file` alone.
Semantic re-extraction should replace semantic contributions and leave AST ones intact.

## Repro

[...]

PR #2387, opened 2026-08-02
fix(build_merge): re-extracted node must not lose to its own stale copy

kellinkquinn-commits:
## The bug

A re-extracted node can be silently replaced by the **stale version it was meant to correct**.

Minimal repro (added as a regression test): a legacy unstamped node for `notes.md` whose `source_location` is `"L38-41"`, then a semantic re-extract of the same id with a corrected, longer label. Before this change the graph keeps the old label.

```
[graphify] note: node 'notes_rank' was extracted twice from 'notes.md' under
different labels — keeping 'Unreconciled: book-to-rank labels off by one between
C-23 and C-27', dropping 'Resolved: the C-23/C-27 book-to-rank off-by-one, closed
by C-45 mapping to Technical Ranks 2/3/4'.
```

## Two things combine

**1. The tier guess.** `_is_ast_tier`'s shape fallback reads any unstamped item with an `L<line>` `source_location` as AST-tier. Semantic extractors do emit that shape for markdown — `build.py`'s own dedup path already guards against it, in a comment that says so explicitly:

> Strict `_origin` check on purpose — NOT `_is_ast_tier` (#2334) ... Those may carry drifted `'L<line>'` source_locations

So a legacy **semantic** node can be read as AST-tier. The tier-scoped replace from #2333/#2336 then does not drop it on a semantic re-extract, and it survives alongside its own fresh version.

**2. The tie-break.** With both present, `_collision_rank` picks the survivor. Definer-ness and `source_file` tie, so it decides on `len(label)` and keeps the **shorter** one. Freshness is not a factor, so a re-extract that expands a claim loses to the text it was correcting.

The failure is silent — node counts unchanged, no dangling edges, and the only signal is a note that reads like routine dedup.

## Why not fix the shape test

`.md` is in the AST extractor registry (`extract.py`: `".md": extract_markdown`), so a doc legitimately carries both tiers, and `_write_two_tier_graph` in the existing tests encodes exactly that. A semantic node with a line-range location is therefore indistinguishable from an AST one by shape alone. Tightening the heuristic would just move the misclassification.

## The fix

Key on identity instead of shape: a base node whose id is re-emitted by a new chunk for the **same** `source_file` is that node's own freshly-extracted version, never a coexisting other-tier node. Drop it whatever tier each side is read as.

Tier scoping still protects the other tier's *distinct* nodes, which is what #2333 was about — an AST heading layer has different ids from the semantic concepts, so it survives untouched. The COEXIST tests are unmodified and still pass.

## Testing

Full suite on Windows / py3.11:

| | failed | passed |
|---|---|---|
| pristine `00efd6e` | 49 | 3681 |
| with this patch | 49 | 3682 |

Identical failure sets — no regressions; the extra pass is the new test. The 49 are pre-existing environment failures on this platform (`skillgen`, `terraform`, `ollama_retry_cap`, `uninstall_scope`, `watch`), unrelated to this path.

Also verified end-to-end against the real 1451-node graph where I hit this: with the patch the re-extract produces the corrected label and **zero** collision notes, because the replace now fires and the collision never happens.

## How I ran into it

A knowledge-base corpus where `wiki/` pages carry adjudicated claims. A re-extract of one page was supposed to update "Unreconciled: ..." to "Resolved: ... closed by C-45"; the graph kept the stale text. Nothing downstream flagged it — the node count was right and no edge dangled. In a corpus whose whole point is that adjudicated answers supersede earlier ones, a silent revert to superseded text is about the worst available failure mode, which is why I chased it to the tie-break rather than just working around it locally.

### DIFF
--- a/graphify/build.py
+++ b/graphify/build.py
@@ -1454,9 +1454,45 @@ def build_merge(
             if norm:
                 tier_sources.add(norm)
     new_sources: set[str] = new_ast_sources | new_sem_sources
-    if new_sources:
+
+    # Tier-scoped replace is not sufficient on its own: a base node whose id is
+    # re-emitted by a new chunk for the SAME source_file is that node's own
+    # freshly-extracted version, never a coexisting other-tier node. Tier scoping
+    # exists to protect the other tier's DISTINCT nodes (#2333), so it must not
+    # keep a stale copy of an id the re-extract just regenerated.
+    #
+    # It can, because _is_ast_tier's shape fallback is a guess: a legacy unstamped
+    # SEMANTIC node whose source_location happens to look like 'L<line>' is read as
+    # AST-tier, so a semantic re-extract does not replace it. It then collides with
+    # its own fresh version, and _collision_rank — which ties here on definer-ness
+    # and source_file — decides on len(label), keeping the SHORTER one. Freshness
+    # is not a factor, so a re-extract that expands a claim loses to the stale text
+    # it was meant to correct. Silent: node counts stay correct, no edge dangles.
+    #
+    # Shape cannot be made reliable here (semantic extractors do emit 'L<line>'
+    # for markdown, and .md is in the AST registry, so a doc legitimately carries
+    # both tiers), so key on identity instead: same id + same file = same entity.
+    new_node_keys: set[tuple[str, str]] = set()
+    for ch in new_chunks:
+        for n in ch.get("nodes", []):
+            nid, sf = n.get("id"), n.get("source_file")
+            if not nid or not sf:
+                continue
+            new_node_keys.add((nid, sf))
+            norm = _norm_source_file(sf, _replace_root)
+            if norm:
+                new_node_keys.add((nid, norm))
+
+    if new_sources or new_node_keys:
         def _kept(item: dict) -> bool:
             sf = item.get("source_file")
+            nid = item.get("id")
+            if nid and sf:
+                if (nid, sf) in new_node_keys:
+                    return False
+                norm = _norm_source_file(sf, _replace_root)
+                if norm and (nid, norm) in new_node_keys:
+                    return False
             own = new_ast_sources if _is_ast_tier(item) else new_sem_sources
             return sf not in own and _norm_source_file(sf, _replace_root) not in own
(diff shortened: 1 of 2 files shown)


yeochul-jeon:
Confirming this on 0.9.48 with an additional failure mode the issue does not
yet cover: the tier mismatch can be created by `build_merge` itself, at load
time, on a graph that never had `_origin` written.

`_load_existing_graph` (build.py:1493/:1496) backfills unstamped items via the
`_is_ast_tier` shape fallback, and `_AST_LOC_RE = ^L\d` (build.py:41) means any
stored `source_location` like `"L73-82"` is classified `ast`. A semantic
extractor that emits a line range — which nothing validates against — therefore
produces a node that a later semantic re-extract cannot replace, because
`_dropped` scopes replacement per tier. Both records reach dedup and
`_collision_rank` (dedup.py:365) decides on `len(label)`, so the fresh longer
label loses to the stale shorter one. The only signal is one stderr `note:`
line from `_report_id_collision`.

Four-combination repro on 0.9.48 (existing node in graph.json / newly extracted
chunk -> survivor):

| existing            | new chunk                | survivor |
|---------------------|--------------------------|----------|
| no `_origin`, L73-82 | no `_origin`, L73-82     | new      |
| no `_origin`, L73-82 | `_origin: semantic`      | **old**  |
| no `_origin`, L73-82 | `_origin: semantic`, loc=null | **old** |
| `_origin: semantic`  | `_origin: semantic`      | new      |

Note row 2 and 3: stamping only the NEW chunk — the obvious local mitigation —
makes things worse, because the existing node is backfilled to `ast` on load.
Any fix has to migrate stored items too, not just newly extracted ones.

Two things that would help independently of the ranking change in this PR:
1. `validate.py` currently accepts a semantic chunk with an `^L\d`
   `source_location` and accepts a missing `_origin`. Rejecting (or at least
   warning on) either would turn a silent label swap into a loud failure.
2. The shipped `skills/*/references/extraction-spec.md` does not mention
   `_origin` at all, so no semantic extractor emits it. Documenting it there
   would let existing installs opt in before the engine change lands.

Issue #981, opened 2026-05-22
Feature: configurable base URL for OpenAI and Anthropic backends (custom providers)


TommyC81:
## Summary

The OpenAI and Anthropic backends hardcode their `base_url`, making it impossible to use OpenAI-compatible or Anthropic-compatible API providers (Z.ai, OpenRouter, Together, Groq, Azure OpenAI, etc.) without modifying source.

The `ollama` backend already supports this via `OLLAMA_BASE_URL` — the same pattern should apply to other backends.

## Related

- **#959** — Same request for `OPENAI_BASE_URL` to support local models (vLLM etc.)
- **#723** — Open PR adding a `custom` backend via `GRAPHIFY_LLM_BASE_URL` env var (40 additions, not yet merged). This implements the Option B approach below for OpenAI-compatible providers.
- **#935** — Open PR adding OpenRouter as a dedicated backend

This issue expands on #959 by also covering the Anthropic backend and proposing two complementary approaches.

## Current Behavior

`graphify/llm.py` `BACKENDS` dict:

```python
"openai": {
    "base_url": "https://api.openai.com/v1",   # hardcoded
    "env_key": "OPENAI_API_KEY",
    ...
},
"claude": {
    "base_url": "https://api.anthropic.com",    # hardcoded
    "env_key": "ANTHROPIC_API_KEY",
    ...
},
"ollama": {
    "base_url": os.environ.get("OLLAMA_BASE_URL", "http://localhost:11434/v1"),  # ✓ configurable
    ...
},
```

At runtime, `_call_openai_compat` receives the base_url from the dict directly (line ~578):

```python
return _call_openai_compat(
    cfg["base_url"],  # always the hardcoded value
    ...
)
```

And `_call_claude` creates the Anthropic client with no base_url at all (line ~397):

```python
client = anthropic.Anthropic(api_key=api_key)  # defaults to api.anthropic.com
```

### Option A: env var overrides per backend (minimal change)

Add `GRAPHIFY_OPENAI_BASE_URL` and `GRAPHIFY_ANTHROPIC_BASE_URL` env var support, following the existing `OLLAMA_BASE_URL` pattern:

```python
"openai": {
    "base_url": os.environ.get("GRAPHIFY_OPENAI_BASE_URL", "https://api.openai.com/v1"),
    ...
},
"claude": {
    "base_url": os.environ.get("GRAPHIFY_ANTHROPIC_BASE_URL", "https://api.anthropic.com"),
    ...
},
```

[...]

Issue #976, opened 2026-05-22
Feature: Add copilot-cli backend for semantic extraction (like claude-cli)


kaovilai:
## Feature idea

Add a `copilot-cli` backend that shells out to the GitHub Copilot CLI for semantic extraction, mirroring how the existing `claude-cli` backend shells out to `claude -p --output-format json`.

## Why this helps

- **No API key needed.** Just like `claude-cli` uses the user's existing Pro/Max subscription, `copilot-cli` would use the user's existing Copilot subscription. No `OPENAI_API_KEY`, no `GEMINI_API_KEY`, no token management.
- **Works in GitHub Agentic Workflows (gh-aw).** In gh-aw, the Copilot CLI is already installed and authenticated as the agent engine. A `copilot-cli` backend would let `graphify extract` work out of the box with zero additional configuration — the CLI binary is right there.
- **Precedent exists.** The `claude-cli` backend already proves this pattern works. It routes through a locally-installed CLI, avoids API key friction, and bills to the user's existing plan.

## Proposed implementation

Add a `"copilot-cli"` entry to the `BACKENDS` dict in `graphify/llm.py`, following the same pattern as `claude-cli`:

```python
"copilot-cli": {
    # Routes through the locally-installed GitHub Copilot CLI.
    # Authenticates via the user's existing Copilot subscription.
    # No API key required — billed to the Copilot plan.
    "default_model": "copilot-cli",
    "pricing": {"input": 0.0, "output": 0.0},
    "temperature": 0,
    "max_tokens": 16384,
}
```

Implementation would need a `_call_copilot_cli()` function similar to `_call_claude_cli()`, invoking the Copilot CLI binary with the extraction prompt and parsing the JSON response.

Auto-detection: if `copilot` (or `github-copilot`) binary is found on `$PATH` and no other backend key is set, auto-select `copilot-cli`.

## Usage

```bash
# Locally (Copilot CLI installed and authenticated)
graphify extract . --backend copilot-cli

# In gh-aw (Copilot CLI is the agent engine, already authenticated)
graphify extract . --backend copilot-cli
```

## How claude-cli works today (for reference)

The `claude-cli` backend calls:
```
claude -p --output-format json "<extraction prompt>"
```

A `copilot-cli` backend would need the equivalent invocation for the Copilot CLI. The exact flags depend on how the Copilot CLI accepts non-interactive prompts — this may need research into the Copilot CLI's headless/non-interactive mode.

## Related

[...]

Issue #975, opened 2026-05-22
Feature: Add GitHub Models backend for semantic extraction


kaovilai:
## Feature idea

Add [GitHub Models](https://docs.github.com/en/github-models) as a backend for `graphify extract`. GitHub Models exposes an OpenAI-compatible inference API at `https://models.inference.ai.azure.com`.

## Why this helps

- **CI/CD environments already have the token.** In GitHub Agentic Workflows (gh-aw), the agent environment has a `COPILOT_GITHUB_TOKEN` (a fine-grained PAT with Copilot Requests permission) that can authenticate against GitHub Models. No additional secret setup is needed — the token is already there, graphify just can't use it.
- **Zero extra cost** for users with GitHub Copilot Pro/Business/Enterprise — model access is included in the subscription.
- **OpenAI-compatible** — the implementation can reuse `_call_openai_compat()` from `llm.py`, making this a small addition to the `BACKENDS` dict.

## Important: token requirements

Not all GitHub tokens can access GitHub Models. The required token is a **fine-grained Personal Access Token** with **Account permissions > Copilot Requests = Read**. This is distinct from:

- `GITHUB_TOKEN` (auto-generated in Actions) — repo-scoped, does **not** have Copilot/model access
- `GH_TOKEN` (from `gh auth login`) — may or may not have model access depending on scopes

The correct env var to check is one explicitly configured for model inference. Auto-detecting from `GITHUB_TOKEN` would silently fail in most CI environments.

## Proposed implementation

Add a `"github-models"` entry to the `BACKENDS` dict in `graphify/llm.py`:

```python
"github-models": {
    "base_url": "https://models.inference.ai.azure.com",
    "default_model": "gpt-4.1-mini",
    "env_keys": ["GRAPHIFY_GITHUB_MODELS_TOKEN", "COPILOT_GITHUB_TOKEN"],
    "model_env_key": "GRAPHIFY_GITHUB_MODEL",
    "pricing": {"input": 0.0, "output": 0.0},  # included in Copilot subscription
    "temperature": 0,
    "max_tokens": 16384,
}
```

Note: `GRAPHIFY_GITHUB_MODELS_TOKEN` as the primary env key avoids clashing with or accidentally consuming the general-purpose `GITHUB_TOKEN`. Users explicitly opt in to GitHub Models by setting this variable.

## Usage

```bash
# In gh-aw (COPILOT_GITHUB_TOKEN is already injected by the runtime)
graphify extract . --backend github-models

[...]

PR #2392, opened 2026-08-02
feat: add GitHub Copilot SDK backend with CLI fallback

nimide:
## Summary

Add GitHub Copilot as an explicit headless LLM backend for Graphify through two transports:

- `copilot-sdk` — preferred structured integration using the official Python SDK and a persistent Copilot runtime.
- `copilot-cli` — standalone one-shot subprocess backend and automatic compatibility fallback.

The integration covers semantic extraction, LLM-assisted deduplication, community naming, PR triage, usage reporting, model selection, timeout handling, and raster-image attachments through the SDK.

Related: #959, #975, #976, #981  
Prior implementation discussion: #1041

## Motivation

Graphify already supports API-backed providers and a local CLI transport. Copilot users should also be able to run Graphify's semantic pass through their existing Copilot entitlement and authenticated GitHub environment without Graphify parsing or persisting OAuth credentials.

The SDK is the preferred transport because it provides structured sessions, responses, image attachments, and a long-lived runtime. The CLI adapter remains valuable for diagnostics, Python 3.10 compatibility, managed environments that do not install the Python SDK, and SDK/runtime incompatibilities.

### New `copilot-cli` backend

- Invokes the official `copilot` executable with the prompt on standard input.
- Uses silent output mode and capability-detects optional hardening/model flags.
- Runs from an empty temporary working directory.
- Disables or denies tool classes and interactive behavior when supported by the installed CLI.
- Inherits normal Copilot authentication, proxy, certificate, and GitHub Enterprise host configuration.
- Runs serially by default; parallel dispatch is an explicit opt-in.

### New `copilot-sdk` backend

- Lazily imports `github-copilot-sdk>=1.0.7,<2` on Python 3.11+.
- Reuses one SDK client/runtime per Graphify process while creating a fresh isolated session for every Graphify request.
- Uses `mode="empty"`, registers no tools or MCP servers, rejects permission requests, and disables memory, infinite sessions, configuration discovery, and remote sessions.
- Uses temporary working and state directories instead of exposing the analyzed repository as an agent workspace.
- Disconnects and permanently deletes each session after the response is received.
- Supports raster images through SDK file attachments.
- Uses a system-installed Copilot CLI by default; an SDK-managed runtime is an explicit opt-in.
- Falls back to `copilot-cli` on SDK import, startup, transport, timeout, session, response, or cleanup failures unless `GRAPHIFY_COPILOT_SDK_FALLBACK=0` is set.
- Treats strict SDK-only operation as an explicit policy choice; because a timeout can be ambiguous, deployments that must avoid a possible replay can disable fallback.

### Integration, packaging, and portability

- Adds both backends to extraction, deduplication, labeling, and PR-triage dispatch.
- Adds backend/model/timeout help and configuration documentation.
- Adds a `copilot` optional dependency extra while retaining Graphify's Python 3.10 floor.
- Updates `uv.lock`, including the Python 3.11 marker for `github-copilot-sdk==1.0.7`.
- Updates generated skill references and their expected artifacts.
- Normalizes model-facing corpus paths to POSIX-style separators so prompts and `source_file` values are stable across Windows and POSIX systems.
- Makes symlink-boundary tests skip only when Windows explicitly denies symlink creation, while retaining the security coverage on platforms that support it.
- Adds focused tests for lifecycle, isolation, fallback, model precedence, enterprise-host propagation, images, errors, and adjacent Graphify integrations.

## Usage

Development install:

```bash
uv sync --extra copilot --frozen
```

Use an existing GitHub.com login, or authenticate to a GitHub Enterprise Cloud data-residency host:

```bash
copilot login --host https://example.ghe.com
export COPILOT_GH_HOST=example.ghe.com
```

Run the preferred SDK backend:

```bash
graphify extract ./docs --backend copilot-sdk --model auto
```

Select the CLI transport directly:

```bash
graphify extract ./docs --backend copilot-cli --model auto
```

Require SDK success and disable fallback:

```bash
export GRAPHIFY_COPILOT_SDK_FALLBACK=0
```

## Compatibility

- Graphify continues to support Python 3.10.
- The optional Python SDK dependency is installed only on Python 3.11+.
- On Python 3.10, selecting `copilot-sdk` can use the direct CLI fallback; `copilot-cli` remains directly selectable.
- Neither Copilot backend is auto-selected merely because a package or executable is installed. Corpus content is sent through Copilot only after the user explicitly chooses one of these backends.

## Security and privacy behavior

- Graphify does not read or store Copilot's persisted OAuth credential.
- SDK sessions have no registered Graphify tools or MCP servers, and every permission request is rejected.
- Temporary session state is removed and SDK session deletion is verified before accepting a successful response.
- Enterprise host, proxy, certificate, and authentication environment settings are inherited rather than hard-coded.
- Selecting either backend transmits semantic-pass content to the Copilot service associated with the authenticated account; code-only AST extraction remains local as before.

## Validation

Focused and adjacent regression suite:

```text
369 passed in 21.66s
```

Additional checks:

```text
python -m compileall -q graphify tests tools
python -m tools.skillgen --check
python -m tools.skillgen --schema-singleton
  schema-singleton OK
uv export --extra copilot --frozen --no-dev --no-hashes
  github-copilot-sdk==1.0.7 on Python >=3.11
git diff --check
```

A bounded extraction-CLI comparison produced the same 20 environment-related failures on the unchanged source snapshot and the feature branch. Those tests require tree-sitter packages that were unavailable in the offline validation environment, and several spawn an isolated interpreter where the checkout was not installed. No Copilot backend test failed.

A live Copilot request was not made from the validation environment because it had no authenticated Copilot account or approved test corpus. SDK and CLI behavior was exercised with controlled protocol/subprocess fakes.

## Checklist

- [x] Changelog and user documentation updated.
- [x] Generic GitHub Enterprise setup documented without an organization-specific hostname.
- [x] Optional dependency and frozen lockfile updated.
- [x] Generated skill artifacts regenerated and checked.
- [x] New backend and integration tests added.
- [x] Existing adjacent LLM/backend tests pass.
- [x] Model-facing paths are normalized and covered by a Windows-path regression test.
- [x] No credentials or tokens committed.


c-mongan:
## Live Copilot SDK validation and hardening evidence

Reference implementation and regression tests: https://github.com/c-mongan/graphify/commit/6cc91279d415de335df5b2bf4be8c6ad966db80e

I tested the Python Copilot SDK backend against a real multi-project Graphify corpus using `gpt-5.6-luna`.

### Live result

- Hardened 32-document pilot: 112 semantic nodes, 98 relationships
- 27/32 source documents represented
- 96,787 input tokens, 21,626 output tokens
- Copilot usage signal: 5.0
- Empty mode, temporary base/work/config storage, no tools/MCP/memory/discovery
- Session disconnect + permanent `delete_session`

### Bugs reproduced during live testing

1. `assistant.message` could terminate collection before `assistant.usage`, producing `0` input tokens and no Copilot usage signal despite real output.
2. History polling could replace complete callback usage with an incomplete history snapshot.
3. `assistant.message.output_tokens` could be double-counted when the authoritative usage event arrived later.
4. `session.get_events()` was not bounded by the request deadline.
5. A valid final message arriving just before the deadline could be discarded while waiting for late usage.
6. `mode="empty"` requires an explicit `base_directory`/session filesystem.
7. `disconnect()` preserves session state; permanent cleanup requires `delete_session(session_id)`.
8. Models sometimes shortened wrapper paths, breaking semantic cache attribution. Repair must be limited to one uniquely provable dispatched path and must reject absolute, URL-like, drive-prefixed, or `..` paths.
9. Generic session errors made live failures impossible to diagnose safely; SDK `error_type` and `error_code` can be surfaced without exposing message/stack/prompt data.

### TDD coverage added locally

- callback usage survives incomplete history
- usage arriving after final message is retained
- message token fallback is not double-counted
- history polling obeys the request timeout
- a valid final message wins over usage settling at the deadline
- empty mode uses isolated base storage
- success/failure cleanup permanently deletes sessions with bounded operations
- safe session error type/code diagnostics
- unique shortened paths repair across nodes/edges/hyperedges
- ambiguous/unsafe parent paths are not repaired
- extraction prompt requires literal wrapper-path copying
- opt-in live E2E requires nonzero input/output/Copilot usage

### Local gates

- Focused suite: 48 passed
- Ruff: passed
- compileall: passed
- skill generation: 134 artifacts matched
- lockfile and diff checks: passed
- Full clean-environment suite: 4,916 passed, 12 skipped

This evidence is intended to help PR #2392 rather than compete with it. The most useful deltas appear to be late-usage accounting, bounded history polling, deadline best-effort response handling, exact provenance repair, and safe live diagnostics.

PR #2398, opened 2026-08-02
fix(skill): share code extension source

### LINKED ISSUE #2227: skillgen: code_only check hard-codes 31 of 94 CODE_EXTENSIONS, so code-only updates pay for a semantic pass
## Summary

The `code_only` check in the `--update` runbook hard-codes a set of code extensions that has drifted far from `graphify.detect.CODE_EXTENSIONS`. On 0.9.22 the literal covers **31 of 94** extensions. The consequence is a silent token cost: an incremental update that touched *only* code is classified `code_only: False` and takes the full semantic path, dispatching LLM subagents over a changed-file set that has nothing semantic in it.

## Where

Source of truth is the shared fragment:

- `tools/skillgen/fragments/references/shared/update.md`

which propagates to every host variant (`graphify/skills/{pi,copilot,opencode,kilo,...}/references/update.md`) and to the `tools/skillgen/expected/*` fixtures — 36 code-search hits for the literal in total.

```python
code_exts = {'.py','.ts','.js','.go','.rs','.java','.cpp','.c','.rb','.swift','.kt','.cs','.scala','.php','.cc','.cxx','.hpp','.h','.kts','.lua','.toc','.f','.F','.f90','.F90','.f95','.F95','.f03','.F03','.f08','.F08'}
...
code_only = all(Path(f).suffix.lower() in code_exts for f in all_changed)
```

## What is missing

The literal is a strict subset of `CODE_EXTENSIONS` — nothing in it is stale, it simply never grew. The 68 missing entries include extensions that **do** have an AST extractor shipped in `graphify/extractors/`:

`.sql` `.ps1` `.psm1` `.psd1` `.jl` `.zig` `.tf` `.tfvars` `.hcl` `.pas` `.dpr` `.ex` `.exs` `.dart` `.groovy` `.cls` `.trigger` `.sv` `.v` `.m` `.mm` `.tsx` `.jsx` `.mjs` `.cjs` `.mts` `.cts` `.vue` `.svelte` `.astro` `.razor` `.cshtml` `.sh` `.bash` `.json` `.sln` …

## Reproduction

`--update` on a repo where the only changed files are three `.sql` files:

```
code_only: False        # expected True
```

`.sql` is in `CODE_EXTENSIONS`, is dispatched to `extractors/sql.py`, and extracted deterministically by Step 3A — so the semantic pass that this misclassification triggers has nothing to do. On my run the AST step produced 12 nodes / 9 edges for those files; the semantic dispatch would have added zero and cost a subagent batch.

## Suggested fix

Read the list from the installed package instead of restating it:

```python
try:
    from graphify.detect import CODE_EXTENSIONS as code_exts
except ImportError:
    code_exts = {...}   # existing literal as fallback
```

Then regenerate `tools/skillgen/expected/*`. I am running this patched locally against 0.9.22 and the three cases that flip (`.sql`-only, `.ps1`-only, `.R`-only) all behave correctly; mixed code+doc and doc-only are unchanged.

## One caveat worth deciding on

`CODE_EXTENSIONS` is not exactly the right predicate — it is *broader* than "has an extractor". `.r` is the clearest case: it is in `CODE_EXTENSIONS` but has no AST dispatch (#1689), so after this fix an R-only update becomes `code_only: True` and correctly skips the semantic pass, but then produces **zero nodes** on the AST path. The net outcome for R is unchanged (it yields nothing either way, since code-typed files are never queued for semantic extraction), so this fix does not regress anything — but it does mean the truly correct predicate is "extensions with a dispatchable extractor", i.e. gated on `_get_extractor(...)` rather than on the extension table. That distinction is the same one #1689 is about, and it might be worth exposing as a single source of truth both places can import.

zrh805:
## Summary
- make every shipped `--update` runbook import `graphify.detect.CODE_EXTENSIONS`
- keep split-platform and Aider/Devin runbooks aligned as new extractors are added
- guard the generated output and explicitly sanction the monolith baseline update

## Verification
- `uv run --frozen pytest tests/test_skillgen.py`
- `uv run --frozen python -m tools.skillgen --check`
- `uv run --frozen python -m tools.skillgen --audit-coverage`
- `uv run --frozen python -m tools.skillgen --schema-singleton`
- `uv run --frozen python -m tools.skillgen --monolith-roundtrip`
- `uv run --frozen python -m tools.skillgen --always-on-roundtrip`
- `uv run --frozen ruff check tools/skillgen/gen.py tests/test_skillgen.py`

Fixes #2227

### DIFF
--- a/graphify/skill-aider.md
+++ b/graphify/skill-aider.md
@@ -777,13 +777,13 @@ If new files exist, first check whether all changed files are code files:
 ```bash
 $(cat graphify-out/.graphify_python) -c "
 import json
+from graphify.detect import CODE_EXTENSIONS
 from pathlib import Path
 
 result = json.loads(open('.graphify_incremental.json').read()) if Path('.graphify_incremental.json').exists() else {}
-code_exts = {'.py','.ts','.js','.go','.rs','.java','.cpp','.c','.rb','.swift','.kt','.cs','.scala','.php','.cc','.cxx','.hpp','.h','.kts'}
 new_files = result.get('new_files', {})
 all_changed = [f for files in new_files.values() for f in files]
-code_only = all(Path(f).suffix.lower() in code_exts for f in all_changed)
+code_only = all(Path(f).suffix.lower() in CODE_EXTENSIONS for f in all_changed)
 print('code_only:', code_only)
 "
 ```
--- a/graphify/skill-devin.md
+++ b/graphify/skill-devin.md
@@ -916,13 +916,13 @@ If new files exist, first check whether all changed files are code files:
 ```bash
 $(cat graphify-out/.graphify_python) -c "
 import json
+from graphify.detect import CODE_EXTENSIONS
 from pathlib import Path
 
 result = json.loads(open('graphify-out/.graphify_incremental.json').read()) if Path('graphify-out/.graphify_incremental.json').exists() else {}
-code_exts = {'.py','.ts','.js','.go','.rs','.java','.cpp','.c','.rb','.swift','.kt','.cs','.scala','.php','.cc','.cxx','.hpp','.h','.kts','.lua','.toc'}
 new_files = result.get('new_files', {})
 all_changed = [f for files in new_files.values() for f in files]
-code_only = all(Path(f).suffix.lower() in code_exts for f in all_changed)
+code_only = all(Path(f).suffix.lower() in CODE_EXTENSIONS for f in all_changed)
 print('code_only:', code_only)
 "
 ```
--- a/graphify/skills/agents/references/update.md
+++ b/graphify/skills/agents/references/update.md
@@ -50,13 +50,13 @@ If new files exist, first check whether all changed files are code files:
 ```bash
 $(cat graphify-out/.graphify_python) -c "
 import json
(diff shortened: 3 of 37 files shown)


zrh805:
Follow-up on the automated review note: this is intentional and uses the same classification source as the runtime extractor. Every affected snippet runs through `graphify-out/.graphify_python`, so `from graphify.detect import CODE_EXTENSIONS` resolves against the installed graphify package that performs Step 3A. The predicate remains `Path(f).suffix.lower() in CODE_EXTENSIONS`; it is not platform-specific.

Validation included `tests/test_skillgen.py` (all shipped Aider, Devin, and split-platform runbooks import and use that shared constant), plus a direct classification check: `.sql`, `.ps1`, and `.R` are code-only while a mixed `.sql` + `.md` set is not. This prevents the existing literals from silently lagging new extractor support.

zrh805:
Follow-up: pushed `db34a83`, which turns the classification evidence into a regression test. It now asserts that `.sql`, `.ps1`, and `.R` are code-only through the exact `Path(...).suffix.lower() in CODE_EXTENSIONS` predicate shipped by every runbook, while a mixed `.sql` + `.md` batch remains non-code-only.

PR #2403, opened 2026-08-02
feat(query): exclude test paths from default seeds

### LINKED ISSUE #2381: query seeds can be hijacked by test/doc rationale nodes — make seed-path exclusion configurable with tests/ as the default
## Problem

`graphify query`'s seed selection (`_score_query` → `_pick_seeds`) matches query terms against node labels, and rationale/docstring nodes sourced from test files carry generic prose labels that substring-match common words. On multi-term natural-language queries these test nodes out-seed the production symbols the user actually asked about, so the BFS explores test neighborhoods and the 2000-token budget fills with low-signal nodes — the failure mode from #897.

**Repro corpus:** a Python repo with tests in both `tests/` and `mcp-svc/tests/`.

```
graphify query "how does the crawl engine handle deduplication of pages?"
```

After stopword filtering, terms are `['crawl', 'engine', 'handle', 'deduplication', 'pages']`. Seeds chosen:

```
crawl()                                    mcp-svc/mcp_server.py         (exact tier)
CrawlEngine                                agent-svc/agent/crawler.py     (relevant)
CrawlEngine calls store.increment_completed() ...   tests/integration/test_crawler.py
Deduplication Strategy                     docs/adr/0038-crawl-engine.md
Engines responded but returned no results with valid URLs.  tests/service/test_searxng_client.py
Handler                                    mcp-svc/tests/test_client.py
Pages with >70% link lines and no paragraphs fail.     tests/service/test_quality.py
```

Three of seven seeds come from test files, via the per-term guarantee (#1445): "engine" → `Engines responded...`, "handle" → `Handler`, "pages" → `Pages with >70% link lines...` are the *singleton winners* for their terms. The per-term guarantee re-injects them even though they are pure noise for this question, and the truncation notice reports 793 nodes, so the caller still can't see `DedupManager` without follow-up calls.

## Proposal

Add a configurable **seed-path exclusion** applied at seed selection only, with test directories as the default:

1. Patterns are gitignore-style globs matched against root-relative `source_file` (e.g. `tests/integration/test_crawler.py`), reusing the existing ignore-matcher in `detect.py` rather than a new matching engine.
2. Default patterns: `tests/`, `**/tests/**` — the second covers nested test dirs (`mcp-svc/tests/`, `src/foo/tests/`). `docs/` is intentionally *not* in the default; doc nodes are often the intended answer (ADR headings).
3. Scope is **seed selection only**: graph construction, `explain`, `path`, and the rendered subgraph traversal are untouched. A test node excluded from seeding remains findable via `explain "TestDedupManager"` and still renders if the BFS reaches it from a real seed (which is usually useful context, e.g. `TestCrawlEngineDedup` exercising `DedupManager`).
4. Explicit configuration replaces the default; an empty list disables filtering entirely (opt-out preserves current behavior).

## Config surface (following existing precedent)

- **CLI:** `graphify query "..." --seed-ignore <pattern>` (repeatable), alongside `--context` / `--budget`.
- **Env:** `GRAPHIFY_QUERY_IGNORE_PATTERNS` (comma-separated), matching the `GRAPHIFY_OUT` / `GRAPHIFY_MAX_CONTEXTS` pattern.
- **MCP `query_graph` tool:** `seed_ignore_patterns` argument, alongside `context_filter` / `token_budget` in `_tool_query_graph` (serve.py ~1494).
- **Precedence:** CLI flag > env var > default.

## Implementation sketch

The filter belongs in exactly one place because of how #1445 interacts with it:

- **`_score_query` (serve.py ~433):** skip excluded nodes in the `node_iter` loop (the trigram candidate iteration). This removes them from *both* the ranked list and `best_seed_by_term` in one pass.
- **`_pick_seeds` (serve.py ~627):** no change needed — but must not re-filter independently, because filtering only the ranked loop would let the per-term guarantee re-inject an excluded winner at serve.py ~701.
- **`_query_graph_text` (serve.py ~1085):** new keyword arg `seed_ignore_patterns: list[str] | None`, forwarded to `_score_query`; CLI parse and MCP tool forward it.

## Edge cases

1. **Test-focused queries must still work.** If filtering leaves zero seeds while the unfiltered set was non-empty, retry unfiltered once (optionally: skip the fallback when a query term contains `test`). A user asking "how do tests verify dedup" should get test nodes.
2. **Per-term guarantee semantics:** a term whose only matches are excluded contributes no seed rather than an excluded one. The fallback covers the all-excluded case.
3. **`explain` / `path` / `_find_node` are byte-identical** — the filter never reaches them.
4. `source_file` is already root-relative (verified: `tests/unit/...`, `mcp-svc/tests/...`), so no path-normalization work is needed.

## Acceptance criteria

- The repro query seeds on `CrawlEngine` / `DedupManager` instead of test-rationale nodes, without `--context` or vocabulary tricks.
- `--seed-ignore ''` (or `GRAPHIFY_QUERY_IGNORE_PATTERNS=`) restores today's behavior exactly.
- A test-focused query ("how does test_crawler assert dedup") still returns test nodes via the fallback.
- `graphify query` / `explain` / `path` / MCP `query_graph` agree; `explain` and `path` outputs are unchanged.

## Related

- #897 seed scoring returns low-relevance nodes (the token-cascade symptom)
- #1445 per-term seed guarantee (the re-injection point this feature must respect)
- #1602 term-coverage scaling (already dampens exact-tier collisions; this addresses the remaining substring-tier noise)
- #1766 seed label dedup

zrh805:
## Summary
- exclude `tests/` and nested test paths only while choosing initial query seeds, so test rationale nodes no longer displace production symbols
- add repeatable `--seed-ignore`, `GRAPHIFY_QUERY_IGNORE_PATTERNS`, and the MCP `seed_ignore_patterns` argument; explicit values replace the defaults and an empty value opts out
- preserve reachable test context and retry unfiltered only when every match was excluded

Fixes #2381

## Validation
- `uv run --frozen pytest tests/test_serve.py tests/test_query_cli.py` (140 passed)
- `uv run --frozen --extra mcp pytest -q tests/test_serve_http.py` (28 passed)
- `uv run --frozen ruff check graphify tests`
- `uv run --frozen graphify update .`

### DIFF
--- a/graphify/__main__.py
+++ b/graphify/__main__.py
@@ -556,6 +556,8 @@ def _run_cli() -> None:
         print("  query \"<question>\"       BFS traversal of graph.json for a question")
         print("    --dfs                   use depth-first instead of breadth-first")
         print("    --context C             explicit edge-context filter (repeatable)")
+        print("    --seed-ignore P         exclude matching paths from initial seeds (repeatable)")
+        print("                           defaults to tests/; GRAPHIFY_QUERY_IGNORE_PATTERNS overrides (empty disables)")
         print("    --budget N              cap output at N tokens (default 2000)")
         print("    --graph <path>          path to graph.json (default graphify-out/graph.json)")
         print("  affected \"X\"             reverse traversal to find nodes impacted by X")
--- a/graphify/cli.py
+++ b/graphify/cli.py
@@ -851,7 +851,7 @@ def dispatch_command(cmd: str) -> None:
             sys.exit(1)
     elif cmd == "query":
         if len(sys.argv) < 3:
-            print("Usage: graphify query \"<question>\" [--dfs] [--context C] [--budget N] [--graph path]", file=sys.stderr)
+            print("Usage: graphify query \"<question>\" [--dfs] [--context C] [--seed-ignore P] [--budget N] [--graph path]", file=sys.stderr)
             sys.exit(1)
         from graphify.serve import _query_graph_text
         from graphify.security import sanitize_label
@@ -863,6 +863,7 @@ def dispatch_command(cmd: str) -> None:
         budget = 2000
         graph_path = _default_graph_path()
         context_filters: list[str] = []
+        seed_ignore_patterns: list[str] | None = None
         args = sys.argv[3:]
         i = 0
         while i < len(args):
@@ -885,6 +886,16 @@ def dispatch_command(cmd: str) -> None:
                 i += 2
             elif args[i].startswith("--context="):
                 context_filters.append(args[i].split("=", 1)[1])
+                i += 1
+            elif args[i] == "--seed-ignore" and i + 1 < len(args):
+                if seed_ignore_patterns is None:
+                    seed_ignore_patterns = []
+                seed_ignore_patterns.append(args[i + 1])
+                i += 2
+            elif args[i].startswith("--seed-ignore="):
+                if seed_ignore_patterns is None:
+                    seed_ignore_patterns = []
+                seed_ignore_patterns.append(args[i].split("=", 1)[1])
                 i += 1
             elif args[i] == "--graph" and i + 1 < len(args):
                 graph_path = args[i + 1]
@@ -956,6 +967,7 @@ def dispatch_command(cmd: str) -> None:
             depth=2,
             token_budget=budget,
             context_filters=context_filters,
+            seed_ignore_patterns=seed_ignore_patterns,
         )
         querylog.log_query(
             kind="query",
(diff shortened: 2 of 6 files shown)


zrh805:
Addressed the review concern in d44252b. Seed-ignore normalization now treats a scalar string as one glob and safely ignores other unexpected shapes, rather than iterating an unvalidated value. Added a regression test; `tests/test_serve.py` + `tests/test_query_cli.py` pass (141), MCP HTTP tests pass (28), and `ruff check graphify tests` is clean.

PR #2415, opened 2026-08-03
fix(markdown): resolve [[wikilinks]] vault-wide, not sibling-only

### LINKED ISSUE #2413: Markdown [[wikilinks]] resolve sibling-only, dropping every cross-folder link in note-type vaults
### Summary

`[[wikilinks]]` between Markdown docs are resolved **sibling-only**, so in a vault
that organizes notes into folders (Concepts/, Entities/, Guides/, Maps/ ...) every
cross-folder link is dropped from the graph. Obsidian resolves `[[name]]` against
the whole vault by filename, not relative to the linking note's directory.

On a real 499-file Obsidian vault this dropped **1,030 of 2,323 authored links (44%)**.

### Where

`graphify/extractors/markdown.py` → `_resolve_markdown_link()`:

```python
candidate = Path(target)
if not candidate.is_absolute():
    candidate = source_dir / candidate     # sibling-only
return Path(os.path.normpath(str(candidate)))
```

`add_link()` then existence-gates the `target_file` stamp:

```python
target_file = None
try:
    if resolved.is_file():
        target_file = str(resolved)
except OSError:
    pass
```

So when the sibling guess misses, no stamp is written, the #2169 remap pass has
nothing to canonicalize, and the edge keeps an absolute-path-derived id that
matches no node — `build_from_json` drops it. This is the same failure shape as
#2211 (fixed in v0.9.28 for the incremental case), but here the target is not in
the sibling directory at all, so the stamp is never produced in the first place.

### Reproduce

```
vault/
  Concepts/alpha.md      # contains: [[beta]]
  Entities/beta.md
```

```python
res = extract([src, tgt], cache_root=root, root=root, parallel=False)
node_ids = {n["id"] for n in res["nodes"]}
refs = [e for e in res["edges"] if e["relation"] == "references"]
[e for e in refs if e["target"] in node_ids]     # -> []  (expected: 1 edge)
```

Observed target id: `..._entities_beta_md` built from the *sibling* guess
`Concepts/beta.md`, which does not exist.

### Measured impact (real vault, 499 notes, v0.9.32)

| | before | after fix | authored baseline |
|---|---:|---:|---:|
| valid `references` | 1,293 | 2,323 | 2,339 |
| recall | 55.3% | **99.3%** | 100% |
| graph edges | 7,623 | 8,470 | — |

The residual 16 are links to files that genuinely do not exist in the vault, and
correctly stay dangling.

### Why this matters beyond one vault

Note-type folders are the default layout for several popular Obsidian setups
(PARA, Zettelkasten with type folders, LYT/MOC). For those vaults the wikilinks
*are* the curated ground-truth relationships — the point raised in #295 — so
losing half of them silently makes the graph substantially sparser than what the
author already encoded, with no warning.

### Proposed fix

Keep sibling resolution first (it is correct for path-qualified links and cheap),
and add a corpus-level fallback for wikilinks only:

1. `markdown.py` — when the sibling guess missed *and* the link came from
   `_MD_WIKILINK_RE`, stamp the bare name (`wikilink_name`) on the edge.
   Path-qualified `[text](./other.md)` links never get the stamp, so relative-link
   semantics are unchanged.
2. `extract()` — in the existing post-pass (the first point that knows every
   in-root file), build a `stem -> [rel path]` index and repoint stamped edges.
   Ambiguity resolves deterministically: same folder as the linking note first
   (mirrors Obsidian), then shallowest path, then lexicographic.

The stamp is transient and popped in the same pass, so `graph.json` is unchanged
in shape. No public interface changes.

Happy to send a PR — I have the patch and tests ready (59 lines across the two
files, 4 new tests, full suite green against the same baseline as `v8` HEAD).

Phantomn:
Fixes #2413.

## Problem

`[[wikilinks]]` were resolved sibling-only. Obsidian resolves `[[name]]` against
the whole vault by filename, so in a vault organized by note type (`Concepts/`,
`Entities/`, `Guides/`, `Maps/`) every cross-folder link missed: no `target_file`
stamp was written, the #2169 remap pass had nothing to canonicalize, and
`build_from_json` dropped the edge.

Same failure shape as #2211 (fixed in v0.9.28 for the incremental case), but here
the target is not in the sibling directory at all, so the stamp is never produced
in the first place.

## Approach

Sibling resolution still runs first and still wins. Only a wikilink whose sibling
guess **missed** is stamped with its bare name; `extract()`'s post-pass — the
first point that knows every in-root file — builds a `stem -> [rel path]` index
and repoints those edges.

- **Path-qualified links are untouched.** `[text](./other.md)` is relative by
  spec, never gets the stamp, and never enters the new code path. Covered by a
  dedicated non-regression test.
- **Deterministic ambiguity resolution**: the linking note's own folder first
  (mirrors Obsidian), then shallowest path, then lexicographic.
- **Transient stamp**: `wikilink_name` is popped in the same pass, so `graph.json`
  shape is unchanged.
- **Broken links still dangle**: a `[[name]]` pointing at a file that exists
  nowhere behaves exactly as before.
- No public interface changes.

## Measured impact

Real Obsidian vault, 499 notes, against `v8` HEAD (0.9.32):

| | before | after | authored baseline |
|---|---:|---:|---:|
| valid `references` | 1,293 | **2,323** | 2,339 |
| recall | 55.3% | **99.3%** | 100% |
| graph edges | 7,623 | 8,470 | — |

The residual 16 are links to files that genuinely do not exist in the vault.

## Tests

4 new cases in `tests/test_languages.py`:

- `test_markdown_wikilink_stamps_name_when_sibling_missing` — stamp appears only
  for the non-sibling target; the sibling one resolves normally
- `test_markdown_path_qualified_link_is_not_stamped` — relative-link
  non-regression
- `test_extract_resolves_wikilinks_across_folders` — cross-folder link resolves,
  a genuinely missing target still dangles, no stamp leaks to the caller
- `test_extract_wikilink_prefers_same_folder_on_ambiguity` — same-stem tie-break

Full suite is green against the same baseline as `v8` HEAD: `3,891 passed` with
the identical 16 pre-existing `test_skillgen.py` failures present before and after
this change (they come from a shallow clone lacking the skill sources, unrelated
to this patch).


Phantomn:
**Gentle bump for review** 🙏

This PR has been open since Aug 3 with all automated checks green:
- ✅ **Graphify** — no coupling regressions or blocking issues
- ⚪ **Graphify Formal Verification** — NEUTRAL (advisory; abstained on two `Path`-typed params, outside the synthesizable set — not a failure)

It's `MERGEABLE` with no conflicts, but has no human review yet, so it can't move forward. Could a maintainer take a look when you have a moment? Happy to address any feedback. Thanks!

PR #2422, opened 2026-08-03
fix(hooks): skip post-checkout rebuild when PREV_HEAD equals NEW_HEAD (#2421)

### LINKED ISSUE #2421: post-checkout: PREV_HEAD/NEW_HEAD are assigned but never read, so `git checkout -b` full-rebuilds an unchanged tree
**Version:** graphifyy 0.9.32 · macOS 15.6 (darwin 25.5) · Python 3.10

## Summary

`_CHECKOUT_SCRIPT` assigns `PREV_HEAD=$1` and `NEW_HEAD=$2` and then never reads either one. As a result `git checkout -b <new>` with no start point — where git passes the *same* sha as `$1` and `$2` and sets `$3=1` — launches a full background rebuild of a tree that did not move.

This is the second half of the suggestion in #1809. The `GRAPHIFY_SKIP_HOOK` half landed (thank you — verified working at 0.9.32, see below); the same-head short-circuit did not, and the two now-unused variables are the tell that it was intended.

## Evidence

`graphify/hooks.py`, inside `_CHECKOUT_SCRIPT`:

```sh
PREV_HEAD=$1
NEW_HEAD=$2
BRANCH_SWITCH=$3
```

`grep` for either name across the rest of the template returns nothing — they are assigned and dropped. Every other guard in the block (`BRANCH_SWITCH != 1`, missing `graphify-out/`, rebase/merge/cherry-pick, `GRAPHIFY_SKIP_HOOK`, the worktree check) is read; these two are not.

Reproduce on a repo with a built graph:

```console
$ graphify hook install
$ git checkout -b throwaway
[graphify] Branch switched - launching background rebuild (log: ~/.cache/graphify-rebuild.log)
```

`git checkout -b` with no start point creates a branch at the current commit. `$1` and `$2` are both that commit, the working tree is byte-identical before and after, and a full rebuild starts anyway.

For contrast, the `GRAPHIFY_SKIP_HOOK` guard from #1809 works correctly at 0.9.32 — same repo, same session:

```console
$ GRAPHIFY_SKIP_HOOK=1 git checkout other-branch
Switched to branch 'other-branch'          # silent, as intended
$ git checkout other-branch                 # without the flag, for contrast
[graphify] Branch switched - launching background rebuild (...)
```

## Why it matters

`post-checkout` runs `_REBUILD_BODY_CHECKOUT`, which is a *full* rebuild with no `changed_paths` — the most expensive thing graphify does. `git checkout -b` is one of the most common git commands there is, and on a branch-per-task workflow this fires on essentially every new branch, each time rebuilding a graph that is already correct.

It also lands seconds-to-minutes after the git command returns, mutating `graphify-out/` under whatever runs next — which is the same class of surprise #1809 was about.

## Suggested fix

One line, next to the existing guards:

```sh
# A no-op checkout (e.g. `git checkout -b` with no start point) reports a branch
# switch but leaves the tree unchanged — nothing to rebuild.
[ "$PREV_HEAD" = "$NEW_HEAD" ] && exit 0
```

That also gives `PREV_HEAD`/`NEW_HEAD` a reader, so the assignments stop being dead.

nothariharan:
Fixes #2421

## Problem

`_CHECKOUT_SCRIPT` in `graphify/hooks.py` assigns `PREV_HEAD=` and `NEW_HEAD=` but never reads either variable.

`git checkout -b <new>` with no start point reports a branch switch (`BRANCH_SWITCH=1`) while leaving the tree byte-identical — git passes the **same** SHA as `` and ``. That still launched `_REBUILD_BODY_CHECKOUT`, a full background rebuild (no `changed_paths`), mutating `graphify-out/` seconds later.

The `GRAPHIFY_SKIP_HOOK` half of #1809 already landed; the same-head short-circuit did not. The unused assignments were the tell.

## Fix

After the existing branch-switch guard, short-circuit when the heads match:

```sh
# A no-op checkout (e.g. `git checkout -b` with no start point) reports a
# branch switch but leaves the tree unchanged — nothing to rebuild (#2421).
[ "$PREV_HEAD" = "$NEW_HEAD" ] && exit 0
```

Real branch switches (`PREV_HEAD != NEW_HEAD`) are unchanged. Rebase/merge/cherry-pick, missing `graphify-out/`, worktree, and `GRAPHIFY_SKIP_HOOK` guards are untouched.

## Testing

- Added `test_checkout_hook_skips_same_head_noop` asserting the short-circuit is present in `_CHECKOUT_SCRIPT`
- `uv run pytest tests/test_hooks.py::test_checkout_hook_skips_same_head_noop tests/test_hooks.py::test_hooks_honor_skip_env -q` → passed
- `python -m tools.skillgen --check` → OK (no skill artifacts touched)

## Notes

CI runs on Ubuntu. Local Windows full-suite noise (path separators / `geteuid` / env backends) is unrelated to this one-line hook-template change.

### DIFF
--- a/graphify/hooks.py
+++ b/graphify/hooks.py
@@ -355,6 +355,10 @@ if [ "$BRANCH_SWITCH" != "1" ]; then
     exit 0
 fi
 
+# A no-op checkout (e.g. `git checkout -b` with no start point) reports a
+# branch switch but leaves the tree unchanged ΓÇö nothing to rebuild (#2421).
+[ "$PREV_HEAD" = "$NEW_HEAD" ] && exit 0
+
 # Only run if graphify-out/ exists (graph has been built before)
 if [ ! -d "graphify-out" ]; then
     exit 0
--- a/tests/test_hooks.py
+++ b/tests/test_hooks.py
@@ -530,6 +530,15 @@ def test_hooks_honor_skip_env(name, script):
     )
 
 
+def test_checkout_hook_skips_same_head_noop():
+    """`git checkout -b` with no start point passes identical PREV/NEW heads.
+    Rebuild must short-circuit ΓÇö PREV_HEAD/NEW_HEAD were previously assigned
+    and never read (#2421 / leftover from #1809)."""
+    assert "PREV_HEAD=$1" in _CHECKOUT_SCRIPT
+    assert "NEW_HEAD=$2" in _CHECKOUT_SCRIPT
+    assert '[ "$PREV_HEAD" = "$NEW_HEAD" ] && exit 0' in _CHECKOUT_SCRIPT
+
+
 @pytest.mark.parametrize("name,script", _HOOK_SCRIPTS)
 def test_hooks_skip_linked_worktrees(name, script):
     """Both hooks must short-circuit in a linked worktree (git-dir != common-dir),


safishamsi:
Landed in v0.9.46, just published to PyPI. Cherry-picked onto `v8` with your authorship preserved in the commit, so both the change and the credit are in the history. Thanks @nothariharan. Closing since it is now released.

PR #2427, opened 2026-08-03
fix(hooks): refresh managed sections safely

rafaelMep:
## Summary

- refresh existing Graphify-managed hook sections instead of appending duplicate blocks
- preserve third-party hook content before and after managed sections
- run managed hook bodies in a subshell so an internal `exit 0` cannot suppress later hook content
- reject malformed Graphify marker layouts without modifying the hook file

## Testing

- focused hook regression tests: 7 passed
- Ruff and Python compilation checks passed

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@ Full release notes with details on each version: [GitHub Releases](https://githu
 
 ## 0.9.50 (2026-08-25)
 
+- Fix: `graphify hook install` now rejects malformed or duplicated managed-section markers without writing, and scopes Graphify's generated shell body so an early exit cannot suppress unrelated hook content that follows it.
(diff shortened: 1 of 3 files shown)


rafaelMep:
Thanks for the follow-up review. I inspected both advisories.

Regarding the `marker_end` finding: `_validated_hook_content()` guarantees that both markers occur exactly once and in the correct order before returning. The boolean expression short-circuits, so `content.index(marker_end)` is never evaluated when the end marker is absent. `_install_hook()` then uses the same validated content, so its `index()` call has that presence guarantee.

Regarding compatibility: rejecting malformed Graphify markers is intentional. Previously, malformed layouts could be silently treated as already installed, leaving an ambiguous or stale managed section. The new behavior fails before writing and preserves both hook files unchanged.

The atomicity case is covered by `test_install_validates_both_hooks_before_writing`, and the focused hook suite passes locally: 103 passed, 1 skipped, with 2 unrelated Windows/MSYS tests deselected.

No additional code change appears necessary for these advisories.

PR #2431, opened 2026-08-03
feat(affected): surface per-edge confidence in affected output (#2352)

### LINKED ISSUE #2352: Expose per-edge confidence in `affected` output (+ `affected --json` mode)
### Summary

`graphify affected "X"` is excellent for reverse blast-radius, but its text output omits the per-edge **confidence** tag that already exists in `graph.json` (`EXTRACTED` / `INFERRED` / `AMBIGUOUS`). Tools that turn blast-radius into *assertions* — e.g. a code reviewer flagging "you forgot to update caller Z" — need to gate on trustworthy edges, and can't today.

### Current output (graphify 0.9.30)

```
- applyToolsAction() [calls] cli/internal/cli/appactions.go:L165
```

The **relation** is shown; the **confidence** is not.

### Request

1. Include each edge's confidence inline, e.g. `- applyToolsAction() [calls, EXTRACTED] cli/internal/cli/appactions.go:L165`.
2. Add an `affected --json` mode mirroring `god-nodes --json`, emitting `{id, label, relation, confidence, confidence_score, source_file, source_location}` per affected node.

### Why

It lets a consumer floor a finding's confidence by the weakest edge on the path and drop `AMBIGUOUS`-only leads — turning `affected` from a "read these candidates" accelerator into a safe *finding source*. Without it, blast-radius findings can't be trusted for correctness-critical uses like review.

Verified against graphify 0.9.30.

Mustaqeem66:
Closes #2352 (text output). See "Scope" below for the `--json` half.

## Problem

`affected` walks edges in reverse and renders each hit as `- label [relation] file:line`. Every edge is *required* to carry a `confidence` (`validate.REQUIRED_EDGE_FIELDS`, one of `EXTRACTED` / `INFERRED` / `AMBIGUOUS`), but the reverse walk dropped it on the floor.

The result: a high-confidence extracted call and a speculative inferred edge render identically. On a wide blast radius you cannot tell which hits are facts and which are guesses without opening `graph.json` and matching edges up by hand — which defeats the point of the command.

## Change

`affected_nodes()` now reads `confidence` and `confidence_score` from **the same edge dict** that already supplies `relation` and the call-site location. That matters: a node can be reached by more than one edge, so the confidence reported has to belong to the edge whose relation actually passed the filter — not whichever edge happened to be stored first. Piggy-backing on the existing lookup makes that invariant structural rather than something a future edit can quietly break.

```
Affected nodes for applyToolsAction()
Relations: calls, indirect_call, references, ...
Depth: 2
- run() [calls, EXTRACTED] cli/run.go:L165
- maybe() [uses, INFERRED] cli/maybe.go:L7
```

Two deliberate calls:

- **An edge with no recorded confidence renders exactly as before** (`- X() [calls] app.py:L4`) rather than gaining an `UNKNOWN` placeholder. Hand-built and older graphs stay clean, and the tag always means something when it appears.
- **`confidence_score` is coerced defensively.** `bool` is rejected explicitly, because bools are ints in Python and a naive `float()` would turn a mis-typed flag into a `1.0` score. Unparseable values degrade to `None` instead of raising mid-walk; numeric strings are still read.

I also added `affected_records()` — the structured per-node rows (`id`, `label`, `depth`, `relation`, `confidence`, `confidence_score`, `source_file`, `source_location`). It shares the `_hit_source()` helper with the text renderer, so the two views resolve the call-site-vs-def-line fallback identically and cannot drift.

## Scope

Issue #2352 asks for two things. This PR fully delivers the first (inline confidence in text output) and provides the complete data layer for the second.

I stopped short of adding the `affected --json` flag itself because the argument parsing lives in `cli.py`, and I did not want to fold an unrelated CLI-surface change into a change that is otherwise contained in `affected.py` and reviewable on its own. With `affected_records()` in place the flag is a small, mechanical follow-up. **Happy to add it to this PR if you'd rather land the feature in one go — just say the word and I'll push it.**

## Compatibility

- `AffectedHit`'s new fields default to `None` and are appended after the existing ones, so positional construction in existing callers and tests is unaffected.
- No existing test asserts the exact contents of the `[...]` tag (`tests/test_affected_cli.py` uses substring assertions like `assert "calls" in out`), so those continue to pass.
- The `#BUG1` call-site behaviour is untouched: hits still report the relation SITE when the edge stored one, and honestly fall back to the node's own definition line when it did not.
- Python 3.10 compatible; no dependency or lockfile changes.

## Tests

New `tests/test_affected_confidence.py` (9 cases) covering: inline tagging of `EXTRACTED` / `INFERRED`; unchanged rendering for confidence-less edges; confidence resolving from the correct edge when parallel edges join the same pair (via `MultiDiGraph`); `AffectedHit` defaults; record shape and JSON round-trip; relation/depth filter behaviour; empty result for an unresolvable seed; and score coercion including the `True`-is-an-int trap.

### DIFF
--- a/graphify/affected.py
+++ b/graphify/affected.py
@@ -36,6 +36,12 @@ class AffectedHit:
     # existing constructors/tests working; None falls back to the node's def line.
     via_file: "str | None" = None
     via_location: "str | None" = None
+    # The traversed edge's confidence, taken from the same edge dict as
+    # via_relation (#2352). `confidence` is a required edge field
+    # (validate.VALID_CONFIDENCES), but hand-built graphs may omit it, so both
+    # default to None and every consumer treats that as "unknown".
+    via_confidence: "str | None" = None
+    via_confidence_score: "float | None" = None
 
 
 def _node_label(graph: nx.Graph, node_id: str) -> str:
@@ -49,6 +55,20 @@ def _format_location(data: dict) -> str:
     if source_location:
         return f"{source_file}:{source_location}"
     return str(source_file)
+
+
+def _coerce_score(value: object) -> "float | None":
+    """An edge's confidence_score as a float, or None when absent/unparseable.
+
+    Bools are rejected explicitly: they are ints in Python, so `float(True)`
+    would silently turn a mis-typed flag into a 1.0 score.
+    """
+    if value is None or isinstance(value, bool):
+        return None
+    try:
+        return float(value)  # type: ignore[arg-type]
+    except (TypeError, ValueError):
+        return None
 
 
 def _bare_name(label: str) -> str:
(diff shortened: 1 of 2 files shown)


Mustaqeem66 (commented):
check the latest commit

Issue #2438, opened 2026-08-03
Incremental update drops indirect_call edges to unchanged callable targets


aryanbonigala:
## Summary

A full build resolves an imported callback passed to a recognized dispatcher into an `indirect_call` edge. Rebuilding only the caller drops that edge when the callable target's file is unchanged.

## Reproduction

`b.py`:

```python
def handler():
    return 1
```

`a.py`:

```python
from b import handler


def run(pool):
    return pool.submit(handler)
```

1. Full build: the `run -> handler` `indirect_call` edge is present.
2. Edit only `a.py` (keeping the callback reference) and rebuild incrementally with `changed_paths=[a.py]`.
3. The `indirect_call` edge is gone. It only comes back after a full rebuild.

## Why the #2406 fix does not cover this

The shared direct-call resolution context added in #2406 supplies persisted node identity (id, label, source_file, type) to the shared **direct** call pass. `indirect_call` needs something that context cannot carry:

* Indirect-call resolution gates its target on the `_callable` and `_callable_class` markers, which say "this symbol is actually invocable" rather than "a symbol with this name exists".
* Those markers are **intentionally stripped** before `graph.json` is written, so they are absent from any persisted node an incremental rebuild could load.
* Inferring callability from the persisted **label** instead would reintroduce the same-named data-symbol false positives that #1566 and #2137 removed — a module-level constant named `handler` would once again look like a callable target.

So the guard correctly refuses to bind, and the edge is dropped rather than mis-resolved.

## Possible directions

* Persist the callable markers (or a narrower, explicitly-versioned "is invocable" flag) in a way that is safe to trust across rebuilds.
* Or give the indirect-call resolver a separate conservative context that carries provenance strong enough to distinguish a callable definition from a same-named data symbol.

Either way the callability signal must be **derived at extraction time and carried forward**, never re-inferred from labels.

## Constraints any fix must keep

* **Removing the callback reference must still remove the old edge.** Preserving prior `indirect_call` edges out of changed files is not an acceptable workaround.
* Same-named data symbols must never resolve as callable targets (the #1566 / #2137 regression guard).

## Tracking

`tests/test_watch.py::test_incremental_rebuild_drops_indirect_call_to_unchanged_target` is an `xfail(strict=True)` covering exactly this reproduction. Closing this issue should make it pass and the marker should then be deleted.

Related to #2406 (which fixes the shared direct-call case only, and does not address this one).

aryanbonigala:
[...]

Issue #2437, opened 2026-08-03
Incremental update drops member-call edges to unchanged files


aryanbonigala:
## Summary

A full build resolves a cross-file **member** call such as `Service.ping()`. Rebuilding only the caller drops that `calls` edge when the target file is unchanged.

## Reproduction

`B.ts`:

```ts
export class Service {
  ping(): number { return 1; }
}
```

`A.ts`:

```ts
import { Service } from "./B";

export function run(): number {
  const service = new Service();
  return service.ping();
}
```

1. Full build: the `run() -> ping()` `calls` edge is present.
2. Edit only `A.ts` (keeping the member call) and rebuild incrementally with `changed_paths=[A.ts]`.
3. The `calls` edge is gone. It only comes back after a full rebuild.

## Why the #2406 fix does not cover this

The shared direct-call resolution context added in #2406 widens the label/file indexes of the **shared direct** cross-file call pass only. That is enough for a plain `shared()`, but not for member calls:

* Member calls are resolved by the **language-specific** resolvers, not the shared pass.
* Those resolvers depend on transient per-file artifacts — notably `ts_type_table` (and the equivalent per-language type/receiver tables) — plus the callee file's `contains` / `method` edges, to know that `service` is a `Service` and that `Service` owns `ping`.
* None of those artifacts are stored in `graph.json`. An incremental rebuild only has them for the files it just parsed, so the unchanged callee's type information simply does not exist in that run.

Passing persisted nodes into those resolvers therefore cannot fix this the way it fixed the direct-call case — the missing input is the type table, not the node list.

## Possible directions

* Persist a safe, minimal subset of resolver metadata (e.g. a per-file receiver/type map) alongside the graph, versioned so a stale or absent map degrades to today's behavior.
* Or reconstruct the needed metadata for just the referenced target files, without reparsing the full corpus.

## Constraints any fix must keep

* **Stale calls must still disappear.** If the member call is removed from the changed caller, the old edge must be evicted — preserving old outgoing edges from changed files is not an acceptable workaround.
* **Ambiguity must continue to fail closed.** When the receiver's type cannot be determined unambiguously, no edge should be emitted, matching current conservative behavior.

## Tracking

`tests/test_watch.py::test_incremental_rebuild_drops_member_call_to_unchanged_target` is an `xfail(strict=True)` covering exactly this reproduction. Closing this issue should make it pass and the marker should then be deleted.

Related to #2406 (which fixes the shared direct-call case only, and does not address this one).

[...]

Issue #2406, opened 2026-08-03
Incremental update loses cross-file edges from changed→unchanged files (batch-scoped import resolution + replace-on-re-extract)


Unreal4598:
**Version:** graphifyy 0.9.19 · Windows 11 · Python 3.14 · TS/TSX corpus (~540 code files)

## Symptom

After an incremental update (re-extracting only changed files, then `build_merge`), the graph silently loses every cross-file edge from a **changed** file to an **unchanged** file. In our repo a utility imported by ~every component (`v2/src/lib/cn.ts`, byte-identical `import { cn } from "@/lib/cn"` across files) dropped from ~100 edges to 4, and its neighborhood became a disconnected island. ~610 edges were lost in one update.

The failure signature is exact and makes the cause easy to confirm:

| edge | outcome |
|---|---|
| changed → changed (`case-detail.tsx → text.ts`) | survives |
| unchanged → unchanged (`Kbd.tsx → cn.ts`) | survives |
| **changed → unchanged** (`case-detail.tsx → cn.ts`) | **deleted** |

## Root cause

1. `extract()` resolves cross-file imports only against the batch it is given (`_augment_symbol_resolution_edges(paths, ...)`). When the update batch contains only changed files, imports of unchanged files resolve to nothing — those edges are simply never produced.
2. `build_merge`'s replace-on-re-extract then drops the changed file's **old** (complete) edge set and replaces it with the new (gutted) one.

Each behavior is individually reasonable; combined they guarantee edge loss on every incremental update of a codebase.

## Minimal repro

```python
from pathlib import Path
from graphify.extract import extract
from graphify.build import build_merge
# A.ts imports { b } from "./B"
r_full = extract([Path("A.ts"), Path("B.ts")], cache_root=None)   # A→B edge present
r_incr = extract([Path("A.ts")], cache_root=None)                 # A→B edge absent (B not in batch)
# build_merge([r_incr], graph_path=...) replaces A's edges → A→B edge deleted from graph.json
```

## Suggested fix

Since AST extraction is deterministic and cheap (~30 s for 500 files, no LLM), the update path could always run structural extraction over the **full corpus** and be incremental only for semantic (LLM) extraction. Alternatively, `_augment_symbol_resolution_edges` could resolve against the existing `graph.json` node index in addition to the batch.

## Secondary finding: dangling edges to never-created node ids

In a full, clean extraction of the same corpus, ~356 edge endpoints reference node ids that no node pass ever creates:

[...]

Issue #2137, opened 2026-07-23
indirect_call callable guard admits classes → false edges for select(Model), except tuples, and getattr literals


marciolprado:
**Version:** graphify 0.9.25 (uv tool install), Python 3.11, macOS

## Summary

The `indirect_call` callable guard admits any node marked `_callable`, which includes **class definitions**. In Python a class is callable (`Foo()` constructs), so the guard passes for every class — but the overwhelmingly common way a class name appears as an argument is as a *descriptor/metadata value* that is never invoked: `select(Model)`, `db.get(Model, id)`, `except (ErrA, ErrB)`.

The result is a large volume of INFERRED `indirect_call` edges asserting a call relationship that does not exist.

The code comment already states the correct semantics:

> `# ... Stays INFERRED even with import evidence: the name is referenced as a value here, not invoked.`

The intent is genuine indirect dispatch — a callback passed by name that *will* be invoked later (`pool.submit(fn)`, dispatch tables). That intent is right; the callable check just can't separate "passed to be called later" from "passed as a value that is never called."

## Where

`graphify/extract.py`

```python
# L4969 — callable_nids includes classes, not just functions
callable_nids = {n["id"] for n in all_nodes if n.get("_callable")}

# L5148 — the guard
if tgt != caller and (caller, tgt) not in call_like_pairs and tgt in callable_nids:
    all_edges.append({..., "relation": "indirect_call", "confidence_score": 0.8, ...})
```

## Three false-positive shapes

Observed on a ~5,500-node Python/TypeScript codebase. All three contexts (`argument`, `collection`, `getattr`) produce them.

### 1. `context="argument"` — ORM model classes (highest volume)

```python
def list_articles(db: Session, ...):
    query = select(KbArticle)          # KbArticle is never called
```

```python
def get_article_analysis(cls, db: Session, article_id: int, ...):
    article = db.get(KbArticle, article_id)   # nor here
```

Both emit `indirect_call → KbArticle`. `KbArticle` is a SQLAlchemy declarative model; `select()`/`db.get()`/`db.query()` take the class as a *table descriptor*. It is never invoked at these sites.

That repo has **602** occurrences of `select(X` / `db.get(X` / `db.query(X` with a capitalized `X`. Any ORM-heavy project (SQLAlchemy, Django, SQLModel, Peewee) hits this on essentially every query.

### 2. `context="collection"` — exception classes in `except` tuples

```python
[...]

PR #2439, opened 2026-08-03
fix(watch): preserve direct calls to unchanged files during incremental updates

aryanbonigala:
## Summary

Incremental rebuilds previously re-extracted only the changed files. The shared direct-call resolver therefore lacked the unchanged target nodes it needed to bind a plain `foo()` to its definition in another file.

Reconciliation correctly removed the changed source's old AST edges — that part was never wrong — but with the callee absent from the extraction batch, the missing direct call could not be regenerated. The result: a cross-file `calls` edge that a full build resolves silently disappears after an incremental update, and only returns on the next full rebuild. (The file-level `imports` edge survived, because the JS/Python symbol-resolution pass reads the import target off disk rather than off the node list — which is why the loss was easy to miss.)

This patch supplies a **read-only AST context** containing unchanged, live, in-corpus nodes exclusively to the shared direct-call resolver:

* Fresh changed-file nodes remain authoritative — on any id collision the freshly extracted node wins.
* Context nodes are **not parsed, not returned, not persisted, and not treated as newly extracted**. They extend the resolver's label/file indexes and nothing else; `all_nodes` is untouched and `raw_calls` still come solely from the re-extracted `paths`.
* Context is scoped to AST-tier nodes only, excludes any file in this run's extract targets, excludes deleted files, and excludes sources no longer in the scanned corpus — so a renamed or removed file can never remain a resolver target.
* Removed calls, removed imports, removed targets, and deleted files still remove their old edges. Preserving stale outgoing edges was explicitly not the approach, and there are tests asserting the edge disappears when the call is deleted.
* If the persisted graph is unreadable or exceeds the size cap, resolution falls back to the changed batch alone (pre-existing behavior).

## Scope

This is intentionally a narrow fix.

* **Fixed:** the exact direct-call reproduction in #2406, for both TypeScript and Python.
* **Not fixed:** language-specific **member calls** (`obj.method()`) and **`indirect_call`** edges. Both require resolver metadata that `graph.json` does not persist — per-file type tables plus the callee's `contains`/`method` edges for member calls, and the `_callable`/`_callable_class` markers (stripped before write) for indirect calls. Inferring either from persisted labels would reintroduce the same-named data-symbol false positives that #1566 and #2137 removed. These remain tracked in the two follow-up issues below.
* The PR deliberately avoids a graph-schema change and avoids a full-corpus reparse. The incremental contract holds: only changed files are extracted, and there is a test asserting exactly that.

Addresses #2406
Related to #2437
Related to #2438

## Testing

All runs on `uv run --frozen`, with the OpenAI extra synced via `uv sync --frozen --group dev --extra openai`. `uv.lock` is unchanged.

| Command | Result |
| --- | --- |
| `pytest tests/test_watch.py -q` | `97 passed, 2 skipped, 2 xfailed` |
| `pytest tests/test_js_import_resolution.py tests/test_extract.py -q` | `222 passed, 1 skipped` |
| `pytest tests/test_ollama_retry_cap.py -q` | `4 passed` |
| `pytest -q` (full suite) | `3915 passed, 35 skipped, 2 xfailed` |
| `ruff check graphify tests` | `All checks passed!` |
| `git diff --check` | clean (no output, exit 0) |

The two `xfail(strict=True)` tests are the admitted scope gaps above, and each corresponds to one of the newly filed follow-up issues:

* `test_incremental_rebuild_drops_member_call_to_unchanged_target` → #2437
* `test_incremental_rebuild_drops_indirect_call_to_unchanged_target` → #2438

They are strict on purpose: whoever closes either issue will be told to delete the marker rather than leave it silently stale.

New passing coverage added alongside them:

* TypeScript and Python cross-file direct call preserved across an incremental rebuild.
* Edge is removed when the call is removed (no stale-edge workaround).
* Edge is not resolved when the import evidence is gone.
* Fresh nodes win when the target file also changed.
* Deleted target file is excluded from the resolution context.
* Only changed files are extracted (incremental contract).
* Repeat incremental rebuilds are idempotent, and match a from-scratch full build.

### DIFF
--- a/graphify/extract.py
+++ b/graphify/extract.py
@@ -4591,6 +4591,7 @@ def extract(
     root: Path | None = None,
     parallel: bool = True,
     max_workers: int | None = None,
+    direct_call_resolution_context_nodes: list[dict] | None = None,
 ) -> dict:
     """Extract AST nodes and edges from a list of code files.
 
@@ -4613,6 +4614,27 @@ def extract(
             use ProcessPoolExecutor for multi-core extraction.
         max_workers: max subprocess count. Defaults to cpu_count (or the
             value of GRAPHIFY_MAX_WORKERS if set), bounded by len(uncached_work).
+        direct_call_resolution_context_nodes: read-only AST nodes from files
+            that are NOT being extracted this run (an incremental rebuild's
+            unchanged corpus, #2406). They extend the label/file indexes of the
+            SHARED DIRECT cross-file call pass only, so a changed caller can
+            still bind a plain `foo()` to an unchanged callee. They are never
+            parsed, mutated, or returned — only edges sourced by `paths` are
+            emitted.
+
+            Deliberately NOT covered (each needs data this list cannot carry,
+            so an incremental rebuild still drops these edge families until the
+            next full rebuild):
+              * member calls (`obj.method()`, #2437) — `run_language_resolvers` below
+                resolves them from `per_file` type tables and the `contains`/
+                `method` edges of the callee's file. Neither survives in
+                graph.json, so the context would have to be reparsed.
+              * `indirect_call` (#2438) — its target guard reads the `_callable` /
+                `_callable_class` markers, which are stripped before the graph
+                is persisted (see the `n.pop` calls near the end of this
+                function). Inferring callability from a persisted label would
+                reintroduce the same-named-data-symbol false positives #1566
+                and #2137 removed.
     """
     paths = [Path(p) for p in paths]
     anchor_root = Path(root) if root is not None else None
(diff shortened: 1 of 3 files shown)


safishamsi:
Shipped in v0.9.33 (`graphifyy==0.9.33` on PyPI). Your approach (widening the resolver context to the persisted unchanged-file nodes, AST-tier only, excluding extract targets and deleted/out-of-corpus sources, failing open on an unreadable graph) is the base of the fix, and I extended it to also cover member calls (#2437) and `indirect_call` (#2438) by threading the unchanged `contains`/`method` edges and persisting the `_callable` markers. Credited in the release notes. Thanks @aryanbonigala, this was a really clean starting point.

PR #2441, opened 2026-08-03
fix(build): ignore global MCP IDs in legacy detection

### LINKED ISSUE #2408: `graph_has_legacy_ids` false-positives on any repo with a nested `.mcp.json`, nagging users to rebuild forever
## Summary

A graph built by 0.9.32 is immediately reported by 0.9.32 as using the "pre-#1504 node-ID scheme". Every `graphify query` / `graphify-mcp` call prints:

```
[graphify] note: this graph uses the pre-#1504 node-ID scheme; rebuild with `graphify extract --force` to get path-qualified IDs (fixes same-name-file collisions).
```

The rebuild it recommends does not clear the warning, because there is nothing to migrate â€” the IDs are already correct. `graph_has_legacy_ids()` is misclassifying the intentionally global-ID nodes emitted by `mcp_ingest`.

Trigger condition: a `.mcp.json` anywhere **below** the repo root (a root-level `.mcp.json` does not trigger it â€” see "Why nested only" below).

## Environment

- graphifyy 0.9.32 (PyPI), Python 3.13, Windows 11
- Not platform-specific; the logic is pure path/string handling

## Reproduction

```bash
mkdir -p /tmp/gfx-repro/sub && cd /tmp/gfx-repro
cat > sub/.mcp.json <<'EOF'
{
  "mcpServers": {
    "cloud-run": {
      "command": "npx",
      "args": ["-y", "@google-cloud/cloud-run-mcp"]
    }
  }
}
EOF
printf 'def hello():\n    return 1\n' > main.py

graphify extract . --force --code-only
graphify query "hello"
```

**Actual:** the `pre-#1504` note is printed against a graph written seconds earlier by the same version.

**Expected:** no note. The graph is on the current scheme.

## Root cause

`mcp_ingest` emits four nodes for that file. Two are stem-scoped, two are deliberately global:

| id | `mcp_kind` | `source_location` |
|---|---|---|
| `sub_mcp` | `mcp_config_file` | `L1` |
| `sub_mcp_mcp_server_cloud_run` | `mcp_server` | `L1` |
| `mcp_command_npx` | `mcp_command` | `L1` |
| `mcp_package_google_cloud_cloud_run_mcp` | `mcp_package` | `L1` |

The last two are correct by design â€” `mcp_ingest.py:49-55` documents that `mcp_command`, `mcp_package`, and `env_var` use global IDs with no per-file stem prefix, specifically so the same package across two configs collapses into one shared node.

`graph_has_legacy_ids` (`build.py:536`) filters candidates with `source_location == "L1"`, on the stated assumption that only file-level nodes carry an unambiguous file-stem ID. That assumption does not hold here: `mcp_ingest` hardcodes `line=1` for **every** node it emits (`mcp_ingest.py:139, 189, 197, 210, 218, 233, 241, 258, 266` â€” the comment at :189 notes JSON has no line info without a second parser pass). So all four nodes above are inspected as if they were file nodes.

Then for `sub/.mcp.json`:

```
make_id(_file_stem(rel))  ->  'sub_mcp'          # canonical
_old_file_stems(rel)      ->  ['sub_mcp', 'mcp'] # includes the bare filename stem
```

`mcp_command_npx` does not start with `sub_mcp`, but it does start with `mcp` + `_`, which matches the old bare-stem form. `graph_has_legacy_ids` returns `True` on the first match and the whole graph is condemned.

The docstring already anticipates exactly this class of collision â€” it explains that symbol nodes are skipped to avoid Go's `_make_id(pkg_dir, name)` producing something that looks like an old file stem. The `mcp_*` global nodes are the same hazard arriving through a filter that doesn't catch them.

## Why nested only

At the repo root, `_file_stem('.mcp.json')` is `mcp` and `_old_file_stems` is `['mcp']`; the `old != new_stem` guard at `build.py:572` short-circuits. The false positive needs a parent directory to make the canonical stem (`sub_mcp`) differ from the bare one (`mcp`).

## Suggested fix

Skip nodes that use global IDs, since their ID is by construction not derived from `source_file`:

```python
GLOBAL_ID_KINDS = {"mcp_command", "mcp_package", "env_var"}
...
if (node.get("metadata") or {}).get("mcp_kind") in GLOBAL_ID_KINDS:
    continue
```

Verified against both a fresh repro graph and a real 1396-node project graph â€” `graph_has_legacy_ids` goes `True -> False`, with genuine detection otherwise unchanged.

That is the minimal fix. The durable one is to stop using `source_location == "L1"` as the file-node predicate: `mcp_ingest` gives everything `L1`, so any future extractor that does the same will reintroduce this. An explicit marker on file-level nodes (or reusing `mcp_kind == "mcp_config_file"`) would make the filter mean what its docstring says.

## Impact

Cosmetic but persistent and self-defeating: the warning goes to stderr on every read-only `query`/`explain`/MCP tool call, and it instructs users to run `extract --force` â€” which for anyone without `--code-only` re-dispatches the full semantic pass to a paid LLM backend. Users can pay real money chasing a warning that will still be there afterward.

aryanbonigala:
### Summary

`graph_has_legacy_ids()` uses `source_location == "L1"` as a cheap proxy for "this node is file-level, so its ID is unambiguously the file stem." MCP ingestion breaks that assumption: JSON exposes no line information, so `mcp_ingest` stamps **every** node it emits with line 1 — including node kinds whose IDs are deliberately *global* rather than derived from `source_file`.

`mcp_command`, `mcp_package`, and `env_var` are global by design: `mcp_command_npx` is meant to be the same node across every config file that invokes `npx`, so its ID never carries a file stem.

For a nested config such as `sub/.mcp.json`, the canonical stem is `sub_mcp` while `_old_file_stems()` also yields the bare pre-#1504 form `mcp`. A perfectly valid, freshly generated `mcp_command_npx` therefore matches the old `mcp_` stem prefix but not the canonical one, and the heuristic reports the graph as legacy. Because the IDs are correct, rebuilding does not clear it — the warning is permanent.

A root-level `.mcp.json` never tripped this: there `mcp` *is* the canonical stem, so the same IDs hit the canonical branch first.

The fix excludes only the three documented global MCP kinds from the heuristic, via a named `_MCP_GLOBAL_ID_KINDS` set and a `_has_global_id()` predicate that reads `metadata.mcp_kind` defensively (missing or non-dict metadata is not exempted and does not raise). File-scoped MCP nodes (`mcp_config_file`, `mcp_server`) are stem-derived and remain fully subject to legacy detection, as do genuine pre-#1504 IDs. Callers, warning wording, and the graph schema are unchanged.

Fixes #2408

### Testing

Focused legacy/MCP tests:

```
$ uv run --frozen pytest tests/test_build.py -k "legacy or mcp" -q
12 passed, 62 deselected, 1 warning in 0.24s
```

The three new assertions fail on the base commit (both `mcp_`-prefixed global kinds and the nested end-to-end case) and pass with the fix.

Related suites:

```
$ uv run --frozen pytest tests/test_build.py tests/test_mcp_ingest.py \
    tests/test_id_normalization_contract.py tests/test_extract.py -q
297 passed, 1 skipped, 1 warning in 1.57s
```

Full suite:

```
$ uv run --frozen pytest -q
3914 passed, 35 skipped, 3 warnings in 46.66s
```

Lint and whitespace:

```
$ uv run --frozen ruff check graphify tests
All checks passed!

$ git diff --check
(clean)
```

`uv.lock` is unchanged (`git diff -- uv.lock` is empty).

The end-to-end regression (`test_fresh_mcp_graph_is_not_flagged_legacy`) is parametrized over **nested `sub/.mcp.json` and root-level `.mcp.json`**, and drives real extraction through `graphify.extract.extract()` rather than a hand-built node dict — it asserts `mcp_command_npx` is actually present before asserting the graph is not flagged. A companion test keeps `mcp_config_file` / `mcp_server` nodes and malformed-metadata nodes under detection so the exemption cannot widen silently.

### DIFF
--- a/graphify/build.py
+++ b/graphify/build.py
@@ -533,6 +533,22 @@ def _semantic_id_remap(nodes: list, root: str | None) -> dict:
     return remap
 
 
+# MCP node kinds whose ID is GLOBAL by design — deliberately shared across every
+# config file that mentions them (`mcp_command_npx`, `mcp_package_...`,
+# `env_var_...`), so it is not derived from ``source_file`` at all (#2408). The
+# file-scoped kinds (`mcp_config_file`, `mcp_server`) ARE stem-derived and stay
+# subject to legacy detection.
+_MCP_GLOBAL_ID_KINDS = frozenset({"mcp_command", "mcp_package", "env_var"})
+
+
+def _has_global_id(node: dict) -> bool:
+    """Whether ``node``'s ID is global by construction rather than file-derived."""
+    meta = node.get("metadata")
+    if not isinstance(meta, dict):
+        return False
+    return meta.get("mcp_kind") in _MCP_GLOBAL_ID_KINDS
+
+
 def graph_has_legacy_ids(nodes: list, root: str | Path | None = None, sample: int = 300) -> bool:
     """Whether a loaded graph still uses pre-#1504 node IDs (parent-dir / filename
     stem) rather than the full repo-relative path. Read-only consumers (query,
@@ -542,8 +558,10 @@ def graph_has_legacy_ids(nodes: list, root: str | Path | None = None, sample: in
     inspected, because their ID is unambiguously the file stem. Symbol nodes are
     skipped — some extractors scope a symbol by package/directory (Go's
     ``_make_id(pkg_dir, name)`` → ``sub_thing``), which can coincide with an old
-    file-stem form and would otherwise false-positive. Returns True as soon as one
-    file node's ID matches an OLD stem form but not the canonical full-path form."""
+    file-stem form and would otherwise false-positive. Nodes whose ID is global by
+    construction (see ``_MCP_GLOBAL_ID_KINDS``) are skipped for the same reason.
+    Returns True as soon as one file node's ID matches an OLD stem form but not the
+    canonical full-path form."""
     from graphify.extractors.base import _file_stem
     _r = str(root) if root is not None else None
     checked = 0
@@ -552,6 +570,14 @@ def graph_has_legacy_ids(nodes: list, root: str | Path | None = None, sample: in
             continue
         if str(node.get("source_location") or "") != "L1":
             continue  # only file-level nodes carry an unambiguous file-stem ID
+        if _has_global_id(node):
+            # #2408: MCP ingest stamps every node it emits with line 1 (JSON has no
+            # line info), so globally-scoped nodes slip past the L1 proxy for
+            # "file-level". For `sub/.mcp.json` the old bare stem is `mcp` while the
+            # canonical stem is `sub_mcp`, so a perfectly valid `mcp_command_npx`
+            # reads as a legacy `mcp_`-prefixed id and warns on every fresh build.
+            # (A root-level `.mcp.json` never tripped it: there `mcp` IS canonical.)
+            continue
         nid = node.get("id")
         sf = node.get("source_file")
         if not nid or not isinstance(nid, str) or not sf:
--- a/tests/test_build.py
+++ b/tests/test_build.py
@@ -1,6 +1,7 @@
 import json
 from pathlib import Path
 import networkx as nx
+import pytest
 from networkx.readwrite import json_graph
 from graphify.build import build_from_json, build, build_merge, edge_data, edge_datas, dedupe_edges, dedupe_nodes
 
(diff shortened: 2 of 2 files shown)


safishamsi:
Shipped in v0.9.42 (`graphifyy==0.9.42` on PyPI) (#2408). Global MCP ids excluded from legacy-id detection. Credited in the release notes. Thanks @aryanbonigala!

Issue #334, opened 2026-04-14
Hyperedge polygons render offset to lower-right on HiDPI (Retina) displays


nhkimmm:
# Hyperedge polygons render offset to lower-right on HiDPI (Retina) displays

## Summary

In `graphify/export.py`, the `drawHyperedges` script injected into `graph.html` re-applies the network view transform on top of a ctx that is already transformed by vis-network, and uses `canvas.width / 2` as a center anchor. On HiDPI (Retina, `devicePixelRatio = 2`) displays this pushes hyperedge convex-hull polygons and labels into the lower-right area of the canvas, visually detached from the nodes they should surround. Node and edge rendering are unaffected.

## Environment

- graphifyy: 0.4.10 (installed via `pipx`); same buggy code present on `v4` branch HEAD as of 2026-04-14.
- OS: macOS 15 (Darwin 25.2), Retina display, `devicePixelRatio = 2`
- Browser: Chrome + Safari (both reproduce)
- Python: 3.14.3
- Corpus: ~11 markdown docs producing 3 hyperedges spanning 3–5 nodes each.

## Reproduction

1. Run graphify on any corpus that produces at least one hyperedge.
2. Open `graphify-out/graph.html` on a Retina display.
3. Hyperedge polygons and labels render as disconnected shapes in the lower-right area of the viewport. Referenced nodes themselves render correctly in their force-directed cluster.

Screenshot from my reproduction (text labels redacted — I will attach it in a follow-up comment):

- Left/center: force-directed cluster of a markdown corpus. Nodes and pairwise edges render correctly.
- Lower-right, visibly detached: three purple convex-hull shapes (the hyperedges). Each hyperedge references 3–5 nodes that all live in the main cluster, but the hull polygons render at a completely unrelated canvas position.

## Root cause — two bugs stacked

Looking at `_hyperedge_script` in `graphify/export.py`:

```js
function drawHyperedges() {
    const canvas = network.canvas.frame.canvas;
    const ctx = canvas.getContext('2d');
    hyperedges.forEach(h => {
        const positions = h.nodes.map(nid => network.getPositions([nid])[nid]).filter(p => p !== undefined);
        if (positions.length < 2) return;
        // ...
        const scale = network.getScale();
        const offset = network.getViewPosition();
        const toCanvas = (p) => ({
            x: (p.x - offset.x) * scale + canvas.width / 2,
            y: (p.y - offset.y) * scale + canvas.height / 2
        });
[...]

PR #2450, opened 2026-08-04
fix(html): trace hyperedge perimeter in convex-hull order, not member order

### LINKED ISSUE #2449: Hyperedge polygons are traced in `h.nodes` array order, so the filled region can self-intersect
`_hyperedge_script` in `graphify/exporters/html.py` maps `h.nodes` straight to positions
and traces that array in order:

```js
// html.py:78, 88-96 (v8 @ HEAD, and 0.9.32 on PyPI)
const positions = h.nodes.map(nid => network.getPositions([nid])[nid]).filter(...);
...
// Centroid and expanded hull in network coordinates
const cx = positions.reduce((s, p) => s + p.x, 0) / positions.length;
const cy = positions.reduce((s, p) => s + p.y, 0) / positions.length;
const expanded = positions.map(p => ({
    x: cx + (p.x - cx) * 1.15,
    y: cy + (p.y - cy) * 1.15
}));
ctx.moveTo(expanded[0].x, expanded[0].y);
expanded.slice(1).forEach(p => ctx.lineTo(p.x, p.y));   // array order
ctx.closePath();
ctx.fill();
```

When the members' layout positions are not already in angular order around the centroid,
the traced path crosses itself and `fill()` renders overlapping wedges rather than one
region.

The 1.15x expansion does not mitigate this. It is a positive homothety about the centroid,
so it preserves segment intersections: `(-1,-1) -> (1,1) -> (-1,1) -> (1,-1)` is a bow-tie
before and after scaling.

Nor is it ruled out by small member counts. In the graph I hit this on, member counts were
`{4: 1, 5: 8, 6: 3, 7: 3}` across 15 hyperedges — no triangles, which are the only sizes
that cannot self-intersect.

It is also not a rare corner. Taking 20,000 random point sets of 4-7 members and tracing
them in arbitrary order, **75.3% self-intersect** after the 1.15x expansion. Members
landing in angular order is the exception, not the rule.

## Before I go further: is the array order intentional?

I want to flag the counter-argument, because it may make this a wontfix.

`h.nodes` order is not random. Several emitted arrays do follow their flow labels:

- `cagra_build_durable_or_nothing_flow`: backend corpus -> daemon build -> atomic persistence
- `artifact_lifecycle_flow`: persist -> reload -> validate -> REINDEX
- `snapshot_warmup_restore_flow`: snapshot -> warmup -> restore

So if the perimeter is meant to trace that sequence, tracing in array order is the intent
and this issue is invalid.

Against that reading, three things in the code suggest hyperedges are modelled as *sets*,
not sequences:

- The extraction prompt (`llm.py:475`) asks for nodes that "participate together in a
  shared concept, flow, or pattern" but never specifies an ordering, and its own examples
  are order-agnostic: "all classes implementing one protocol", "all functions in one auth
  flow **even if they don't all call each other**", "all concepts from a paper section
  forming one coherent idea".
- The schema's `relation` vocabulary is `participate_in | implement | form` — membership
  verbs. There is no sequence-typed relation. In my graph the split is `participate_in` 10,
  `form` 5.
- `callflow_html.py:1754-1772` renders each hyperedge as a bulleted list headed
  "`<relation>` — N **participants**". It preserves array order, but presents membership,
  not order.

And the polygon itself has no way to carry a sequence: `globalAlpha` 0.12 fill, 0.4 stroke,
one centroid label, no arrowheads or direction cues. A viewer cannot recover
`corpus -> build -> persist` from a filled translucent blob.

For what it is worth, #334 — the previous rendering fix in this same function — describes
these shapes as "hyperedge **convex-hull** polygons", which reads like hull semantics were
the intent there too.

That said, whether the ordering is meant to be load-bearing is your call, not mine. If it
is, feel free to close this.

## What I observed

On a 2013-node / 3928-edge graph with 15 hyperedges: before changing anything, several
overlapping blades crossed the canvas; after applying an angular sort, one elongated
region remained and the rest rendered as compact filled areas.

To be precise about the evidence: that is a visual before/after observation, not a
measurement. `graph.json` stores no coordinates — vis.js generates them at runtime — so I
cannot supply a coordinate fixture that reproduces the exact rendering.

The one region that stayed elongated is a data property, not a tracing artifact. Measuring
undirected graph distance between each hyperedge's members gives max finite distances of
`inf, 6, 3, 3, 3, 4, 3, 3, 3, 2, 3, 2, 5, 3, 4` — 14 of 15 are connected within 6 hops,
while `cagra_build_durable_or_nothing_flow` has a member (`architecture_fail_closed`) in a
different connected component. Its members genuinely sit at opposite ends of the layout.
(Hop distance does not prove layout distance; it is corroboration, not proof.)

## Possible fix, with its limits

Sorting the vertices by angle around the centroid before tracing:

```js
expanded.sort((a, b) => Math.atan2(a.y - cy, a.x - cx) - Math.atan2(b.y - cy, b.x - cx));
```

Sorting before or after the 1.15x expansion is equivalent, since the homothety preserves
each point's angle.

This is an ordinary-case mitigation, not a guarantee:

- It produces a *simple* polygon, not a *convex* one. It is not a convex-hull algorithm,
  despite what the line-88 comment says.
- No guard for coincident positions, equal angles, NaN coordinates, or a centroid landing
  on a vertex (`Math.atan2(0, 0)` ties at zero).
- Degenerate collinear sets such as `(-2,0), (-1,0), (1,0), (2,0)` still yield a
  zero-area overlapping path.

So I went with a convex hull instead (Andrew's monotone chain, ~12 lines). It matches the
line-88 comment and #334's wording, and it degrades cleanly on the cases above: collinear
and duplicate points collapse to the extremes rather than producing a zero-area path, and
there is no `atan2(0, 0)` tie to worry about.

PR attached. If you would rather have the one-line angular sort, or if the array order is
intentional and this should be closed, just say so and I will drop it.

## Reproduction path

Note that this code is not reached by headless `graphify extract`, which stops at graph
JSON and analysis (`cli.py:3685`). It renders via `cluster-only` (`cli.py:1891`),
`export html` (`cli.py:2450`), and watch mode (`watch.py:1487`).

## Environment

- graphifyy 0.9.32, installed via `uv tool install`; PyPI reports 0.9.32 as latest
- Same code confirmed in branch `v8` (repo default branch) at HEAD via the GitHub
  contents API
- macOS 15 (Darwin 25.5.0), Python 3.12
- 2013 nodes, below `MAX_NODES_FOR_VIZ`, so no community aggregation: the emitted
  `graph.html` carries all 2013 raw nodes and all 83 hyperedge members resolve against
  them. (Checked because of #1005.)

Searched existing issues for `hyperedge`, `polygon`, `convex hull`, `self-intersect`,
`shaded region`, `afterDrawing`, and `atan2`. The nearest neighbours are all distinct:

- #334 (closed, fixed 0.4.13) — HiDPI coordinate offset in this same function.
- #1005 (closed, fixed 0.8.18) — hyperedges dropped from the aggregated meta-graph above
  the viz node limit.
- #2361 (open) — optional WebGL renderer past the 5000-node aggregation ceiling; additive
  rather than a replacement for the vis-network path, so this still applies.

I did not find an existing report of the tracing order itself.

ysys143:
Fixes #2449.

## What

`_hyperedge_script` traced the hyperedge ring by walking `h.nodes` in array
order. That self-intersects whenever the layout does not happen to place members
in angular order around the centroid, so `fill()` paints crossed wedges instead
of one region. The 1.15x expansion cannot mask it — a positive homothety
preserves segment intersections.

This computes Andrew's monotone chain over the member positions and expands the
hull instead of the raw array.

## Why a hull rather than an angular sort

A one-line `atan2` sort also removes the crossings, and I tried that first. I
went with the hull because it matches the existing `// Centroid and expanded
hull` comment and #334's "convex-hull polygons" wording, and because it degrades
better on degenerate member sets: collinear and duplicate points collapse to the
extremes rather than producing a zero-area crossed path, and there is no
`atan2(0, 0)` tie when the centroid lands on a vertex.

Happy to swap it for the one-liner if you prefer the smaller diff.

## Scale of the defect

Over 20,000 random 4-7 member point sets (the sizes real hyperedges carry — mine
were `{4: 1, 5: 8, 6: 3, 7: 3}`), **75.3% of arbitrary orderings self-intersect**
after expansion. This was the common case, not a corner.

## Tests

Two tests in `tests/test_export.py`:

- `test_hyperedge_perimeter_uses_convex_hull_not_member_order` — guards against
  reverting to raw member order.
- `test_hyperedge_convex_hull_js_is_geometrically_sound` — executes the emitted
  JS in node and asserts the ring is simple, convex, counter-clockwise, and
  encloses every member, across 2,000 random sets plus the degenerate cases and
  the bow-tie ordering this exists for. Skipped when node is unavailable.

Both fail on the unpatched exporter and pass with the change.

`tests/` is otherwise unaffected: 3894 passed. The 15 failures in
`test_ollama_retry_cap.py` and `test_skillgen.py` reproduce identically on a
clean `v8` checkout without this change.

## Open question

#2449 raises it too: if `h.nodes` order is meant to be load-bearing — several
emitted arrays do follow their flow labels — then tracing in array order is the
intent and this should be closed. My reading is that the schema treats
hyperedges as sets (`participate_in | implement | form`, no sequence relation)
and that a translucent fill with no direction cues cannot convey a sequence
anyway, but that call is yours.

### DIFF
--- a/graphify/exporters/html.py
+++ b/graphify/exporters/html.py
@@ -73,6 +73,27 @@ def _hyperedge_script(hyperedges_json: str) -> str:
 const hyperedges = {hyperedges_json};
 // afterDrawing passes ctx already transformed to network coordinate space.
 // Draw node positions raw — no manual pan/zoom/DPR math needed.
+
+// Andrew's monotone chain. Returns the hull in counter-clockwise order, which
+// is what the perimeter must be traced in. Collinear and duplicate points
+// collapse to the extremes, so degenerate member sets render as a segment
+// rather than a zero-area crossed path.
+function convexHull(pts) {{
+    const p = pts.slice().sort((a, b) => (a.x - b.x) || (a.y - b.y));
+    if (p.length < 3) return p;
+    const cross = (o, a, b) => (a.x - o.x) * (b.y - o.y) - (a.y - o.y) * (b.x - o.x);
+    const build = seq => {{
+        const out = [];
+        for (const q of seq) {{
+            while (out.length >= 2 && cross(out[out.length - 2], out[out.length - 1], q) <= 0) out.pop();
+            out.push(q);
+        }}
+        out.pop();
+        return out;
+    }};
+    const hull = build(p).concat(build(p.slice().reverse()));
+    return hull.length >= 3 ? hull : p;
+}}
 network.on('afterDrawing', function(ctx) {{
     hyperedges.forEach(h => {{
         const positions = h.nodes
@@ -85,10 +106,14 @@ network.on('afterDrawing', function(ctx) {{
         ctx.strokeStyle = '#6366f1';
         ctx.lineWidth = 2;
         ctx.beginPath();
-        // Centroid and expanded hull in network coordinates
+        // Centroid and expanded hull in network coordinates.
+        // The perimeter must follow hull order, not h.nodes order: tracing the
+        // raw member order self-intersects whenever the layout does not happen
+        // to place members in angular order, filling as crossed wedges.
         const cx = positions.reduce((s, p) => s + p.x, 0) / positions.length;
         const cy = positions.reduce((s, p) => s + p.y, 0) / positions.length;
-        const expanded = positions.map(p => ({{
+        const hull = convexHull(positions);
+        const expanded = hull.map(p => ({{
             x: cx + (p.x - cx) * 1.15,
             y: cy + (p.y - cy) * 1.15
         }}));
--- a/tests/test_export.py
+++ b/tests/test_export.py
@@ -831,3 +831,81 @@ def test_existing_graph_node_count(tmp_path):
     assert existing_graph_node_count(p) is MALFORMED_GRAPH  # structurally wrong -> fail closed
     p.write_text('{"nodes": [{"id": "a"}, {"id": "b"}], "links": []}', encoding="utf-8")
     assert existing_graph_node_count(p) == 2               # valid
+
+
+def test_hyperedge_perimeter_uses_convex_hull_not_member_order():
+    """The hyperedge polygon must be traced in hull order. Tracing `h.nodes`
+    array order self-intersects whenever the layout does not place members in
+    angular order, so `fill()` paints crossed wedges instead of one region."""
+    from graphify.exporters.html import _hyperedge_script
+    script = _hyperedge_script("[]")
+    assert "function convexHull(pts)" in script
+    assert "const hull = convexHull(positions);" in script
+    # the traced ring must derive from the hull, never from raw member order
+    assert "const expanded = hull.map(" in script
+    assert "const expanded = positions.map(" not in script
+
+
+def test_hyperedge_convex_hull_js_is_geometrically_sound():
+    """Execute the emitted convexHull in node: the perimeter must be simple
+    (no self-intersection), convex, and contain every member point."""
+    import shutil
+    import subprocess
+    node = shutil.which("node")
+    if node is None:
+        import pytest
+        pytest.skip("node not available")
+    from graphify.exporters.html import _hyperedge_script
+    m = re.search(r"function convexHull\(pts\) \{.*?\n\}", _hyperedge_script("[]"), re.S)
+    assert m, "convexHull not found in emitted script"
+    harness = m.group(0) + r"""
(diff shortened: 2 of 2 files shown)


safishamsi:
Shipped in v0.9.42 (`graphifyy==0.9.42` on PyPI) (#2449). Hyperedge perimeter traced in convex-hull order. Credited in the release notes. Thanks @ysys143!

PR #2458, opened 2026-08-04
fix(extract): fail on AST extraction errors

### LINKED ISSUE #2445: graphify extract substitutes an empty AST result on any exception, so a total extraction failure looks like an empty corpus
## Summary

In `graphify extract`, a failed AST extraction is caught by a blanket `except Exception` and replaced with an empty result. The command then continues and completes normally, so a total extraction failure is indistinguishable from a corpus that genuinely has no code.

Version: **graphifyy 0.8.36**.

## Code

`__main__.py:4154-4158`:

```python
try:
    ast_result = _ast_extract(code_files, **ast_kwargs)
except Exception as exc:
    print(f"[graphify extract] AST extraction failed: {exc}", file=sys.stderr)
    ast_result = {"nodes": [], "edges": [], "input_tokens": 0, "output_tokens": 0}
```

The message goes to stderr, but the command's exit status and subsequent behaviour are unchanged: downstream steps consume `{"nodes": [], "edges": []}` as though extraction had succeeded and found nothing.

## Why it matters

Combined with #2444 (where `extract()` can itself return zero nodes without raising at all), there is no signal — neither an exception nor an exit code — that distinguishes:

- a corpus with no code files,
- an extraction where every worker died,
- an extraction that raised.

All three end as an empty AST contribution and a successful-looking run. For an incremental update this is worse than a crash, because the graph silently stops reflecting the code.

## Suggested fix

Either let the exception propagate, or keep the catch but exit non-zero — and in any case, don't substitute a well-formed empty result for an error. If the swallow is deliberate for robustness in batch runs, a `--strict` flag (or making the empty substitution conditional on `code_files` being empty) would let callers opt into failing loudly.

Related: I hit this alongside a per-future `BrokenProcessPool` swallow in `_extract_parallel`, filed separately.

Baziar:
## Summary

- Exit with status code `1` when AST extraction raises an exception.
- Stop replacing failed extraction with a successful-looking empty AST result.
- Prevent downstream processing and `graph.json` creation after a total AST failure.
- Add regression coverage for the failure path.

Corpora with no code files remain unaffected because AST extraction is only attempted when `code_files` is non-empty.

## Tests

- `186 passed, 1 skipped`
- Ruff checks passed.
- `graphify update .` completed successfully.

Fixes #2445

### DIFF
--- a/graphify/cli.py
+++ b/graphify/cli.py
@@ -3096,8 +3096,7 @@ def dispatch_command(cmd: str) -> None:
                 ast_result = _ast_extract(code_files, **ast_kwargs)
             except Exception as exc:
                 print(f"[graphify extract] AST extraction failed: {exc}", file=sys.stderr)
-                ast_result = {"nodes": [], "edges": [], "input_tokens": 0, "output_tokens": 0}
-                _extraction_incomplete = True  # the whole AST pass was lost
+                sys.exit(1)
         stages.mark("AST extract")
 
         # Semantic extraction on docs/papers/images. Check cache first.
--- a/tests/test_extract_cli.py
+++ b/tests/test_extract_cli.py
@@ -1,6 +1,7 @@
 """Tests for `graphify extract` CLI dispatch path in graphify.__main__."""
 from __future__ import annotations
 
+import importlib
 import os
 
 import pytest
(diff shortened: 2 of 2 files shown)


safishamsi:
Shipped in v0.9.33. The direction (do not let a whole-pass AST failure masquerade as an empty graph) is what landed, gated so a fresh-build failure exits non-zero while `--allow-partial` still opts into a best-effort partial graph (keeping the partial-continuation path). Credited in the release notes. Thanks @Baziar.

PR #2461, opened 2026-08-04
fix(extract): recover from per-future BrokenProcessPool

### LINKED ISSUE #2444: BrokenProcessPool swallowed per-future, defeating the sequential fallback: extract() returns 0 nodes and reports success (Windows)
## Summary

`_extract_parallel`'s per-future `except Exception` swallows `BrokenProcessPool`, so the deliberate `BrokenProcessPool` handler below it — and the sequential fallback it exists to trigger — never fire in the Windows-spawn case its own comment describes. The result is that `extract()` returns **zero nodes and reports success**.

Version: **graphifyy 0.8.36**, Windows 11, CPython 3.11.12.

## The mechanism

`extract.py`, inside `_extract_parallel`:

```python
for future in concurrent.futures.as_completed(futures):
    try:
        idx, result = future.result()
        per_file[idx] = result
    except Exception as exc:                      # <-- :11218 catches BrokenProcessPool
        idx = futures[future]
        print(f"  warning: worker failed for {work_items[idx][1]}: {exc}", file=sys.stderr)
    ...
except concurrent.futures.process.BrokenProcessPool:   # <-- :11234 never reached
    # On Windows (spawn start method) the worker subprocesses re-import the
    # caller's __main__ ... Fall back to in-process sequential extraction
    print("  warning: parallel extraction failed (BrokenProcessPool); falling back to sequential...")
    return False
return True                                        # <-- :11253 "the pool ran to completion"
```

When the pool dies after submission, `future.result()` raises `BrokenProcessPool` **per future**. The inner `except Exception` catches each one, prints a per-file warning, and swallows it. The `with` block then exits normally, `:11234` is never entered, and the function returns `True`. The caller's `if not ran_parallel:` fallback is therefore skipped.

`:11234` is not dead in general — `pool.submit` at `:11212` is inside the same `try`, so a pool that breaks *during submission* does reach it. But it does not fire for the case the comment is specifically about.

## Reproduction (Windows)

```python
# repro.py -- deliberately NO `if __name__ == "__main__":` guard
from pathlib import Path
result = extract([Path(p) for p in my_25_py_files], cache_root=Path('.'))
print(f"AST: {len(result['nodes'])} nodes, {len(result['edges'])} edges")
```

Needs ≥20 uncached files so the parallel threshold is crossed. Observed:

```
  warning: worker failed for ...\a.py: A process in the process pool was terminated abruptly...
  ... (one line per file) ...
AST: 0 nodes, 0 edges
```

**Exit code 0.** Expected either the sequential fallback, or a non-zero exit.

The underlying worker failure is the ordinary Windows-spawn one:

```
File "...\multiprocessing\spawn.py", line 297, in _fixup_main_from_path
    main_content = runpy.run_path(main_path, ...)
File "...\repro.py", line 15, in <module>
    result = extract(code_files, cache_root=Path('.'))
File "...\graphify\extract.py", line 11364, in extract
    ran_parallel = _extract_parallel(
```

`python -c "..."` is unaffected, because `__main__` then has no `__file__` and `_fixup_main_from_path` never runs.

## Why it matters

This is a silent-wrong failure rather than a loud one. In our case a `--update` over 56 changed code files produced `AST: 0 nodes, 0 edges` and exit 0; the pipeline would have gone on to merge, cluster, and re-export, printing a successful completion with the code half of the graph missing. It was caught only because a human noticed the number looked wrong.

## Suggested fix

Re-raise (or flag) `BrokenProcessPool` from the inner handler so the outer one can do its job:

```python
except concurrent.futures.process.BrokenProcessPool:
    raise
except Exception as exc:
    ...
```

Separately, it may be worth having `_extract_parallel` return `False` if *no* future produced a result, since "every worker failed" and "the pool ran to completion" should not be the same return value.

Happy to open a PR if the approach looks right.

Baziar:
## Summary

- Re-raise `BrokenProcessPool` from the per-future exception handler.
- Allow the existing pool-level handler to return `False`.
- Retry the uncached AST batch sequentially instead of returning an empty result.
- Preserve the existing behavior for ordinary per-file worker failures.
- Add regression coverage for `BrokenProcessPool` raised by `future.result()`.

## Why

`future.result()` can raise `BrokenProcessPool` after successful task submission. The generic per-future exception handler previously swallowed that exception, causing `_extract_parallel()` to report success and preventing the sequential fallback.

The result could therefore contain zero AST nodes while appearing successful.

## Tests

- `161 passed, 1 skipped`
- Ruff checks passed.
- `graphify update .` completed successfully.

Fixes #2444

### DIFF
--- a/graphify/extract.py
+++ b/graphify/extract.py
@@ -4502,6 +4502,10 @@ def _extract_parallel(
                 try:
                     idx, result = future.result()
                     per_file[idx] = result
+                except concurrent.futures.process.BrokenProcessPool:
+                    # Let the pool-level handler below signal the caller to
+                    # retry the entire uncached batch sequentially.
+                    raise
                 except Exception as exc:
                     pos = futures[future]
                     print(
--- a/tests/test_extract.py
+++ b/tests/test_extract.py
@@ -1593,6 +1593,56 @@ def test_extract_parallel_returns_false_on_broken_pool(tmp_path, monkeypatch, ca
     assert "__main__" in out, "warning must hint at the Windows __main__ guard idiom"
 
 
+def test_extract_falls_back_when_worker_future_breaks_pool(
+    tmp_path, monkeypatch, capsys
+):
+    """A BrokenProcessPool from future.result() must trigger sequential fallback."""
+    from concurrent.futures.process import BrokenProcessPool
+    import concurrent.futures
+    from graphify import extract as extract_mod
+
+    class BrokenFuture:
+        def result(self):
+            raise BrokenProcessPool("simulated worker termination")
+
+    class FakePool:
+        def __init__(self, *args, **kwargs):
+            pass
+
+        def __enter__(self):
+            return self
+
+        def __exit__(self, *args):
+            return False
+
+        def submit(self, *args, **kwargs):
+            return BrokenFuture()
+
+    monkeypatch.setattr(concurrent.futures, "ProcessPoolExecutor", FakePool)
+    monkeypatch.setattr(
+        concurrent.futures,
+        "as_completed",
+        lambda futures: iter(futures),
+    )
+
+    sequential_calls = 0
+    real_sequential = extract_mod._extract_sequential
+
+    def wrapped_sequential(*args, **kwargs):
+        nonlocal sequential_calls
+        sequential_calls += 1
+        return real_sequential(*args, **kwargs)
+
+    monkeypatch.setattr(extract_mod, "_extract_sequential", wrapped_sequential)
+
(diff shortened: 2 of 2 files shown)


safishamsi:
Shipped in v0.9.33. The core of your fix (re-raising `BrokenProcessPool` so the sequential fallback engages) is what landed, with a couple of additions: the fallback re-extracts only the not-yet-completed files, and a worker-failed file is retried sequentially and error-marked rather than left as a silent empty. Credited in the release notes. Thanks @Baziar.

Issue #2011, opened 2026-07-18
Private vulnerability reporting returns HTTP 500 — requesting a working channel for two security-relevant SECURITY.md mismatches


Job28703:
Opening this as a channel request rather than a report — there are no technical details below.

## What I have

While reading the codebase I ended up checking `SECURITY.md`'s threat table line by line against the implementation. Several entries no longer match the code. Three of them are documentation-only and I've filed them publicly as #2007. **Two of them are security-relevant and I don't think they belong in a public issue**, so I'd like to send those privately.

## The problem: neither channel in `SECURITY.md` currently works for an external reporter

`SECURITY.md` says:

> Report security issues via GitHub's private vulnerability reporting, or email the maintainer directly.

**Private vulnerability reporting** is enabled on the repo — `GET /repos/Graphify-Labs/graphify/private-vulnerability-reporting` returns `{"enabled": true}` — but submitting through it fails:

```
POST /repos/Graphify-Labs/graphify/security-advisories/reports
→ HTTP 500, empty body
```

Consistently, across four attempts. I checked that this isn't on my end:

| Request | Result |
|---|---|
| `GET /user` (same token) | 200 — auth is fine |
| `POST …/reports` with empty body | 422 — validation layer is alive |
| `POST …/reports` missing `description` | 422 — field validation works |
| `POST …/reports` to a nonexistent repo | 404 — routing works |
| `POST …/reports` with a valid payload | **500, empty body** |

So it fails only after validation passes, at the point of actually creating the report. `x-accepted-oauth-scopes` comes back empty, so it isn't a token-scope issue either. GitHub request IDs, if you want to raise it with Support: `FA35:F7A54:77BAD1:89E3E4:6A5BB24F`, `F9FC:1ECD68:771358:891DBD:6A5BB113`.

**Emailing the maintainer directly** — `SECURITY.md` offers this as the alternative but doesn't include an address anywhere in the file. (I'd rather not use an address scraped from commit metadata for something like this without being invited to.)

## What I'm asking

Any one of these works for me:

1. Confirm the web advisory form (`/security/advisories/new`) accepts reports — the REST API and the web UI are different paths, so it may well be fine and this is purely an API-side fault.
2. Add a contact address to `SECURITY.md` for when private reporting is unavailable.
3. Tell me you'd rather have them as regular issues, and I'll file them with the detail level you're comfortable with.

[...]

PR #2463, opened 2026-08-04
fix(extract): skip files that are not regular files (a FIFO hangs the run forever)

itskaism:
A repository can contain named pipes, sockets and device nodes, and
`graphify clone <github-url>` exists to point the extractor at trees the
operator did not write. Any of those carrying a source suffix is currently
admitted to the corpus and handed to a reader.

## The bug

A FIFO named `pipe.py` hangs the entire run:

```
$ graphify update <repo> --no-cluster
(no output; killed at 180s)
```

`open()` on a named pipe with no writer **blocks forever and never raises**,
so the `try`/`except` around every reader cannot help. Confirmed at kernel
level on the stuck worker process:

```
/proc/<pid>/wchan   -> wait_for_partner
/proc/<pid>/syscall -> 257  (openat)
```

`watch` mode re-runs the same walk, so the hang is silent and permanent.

Bisected against six other hostile shapes — only this one hangs:

| shape in the scanned repo | result |
|---|---|
| **FIFO named `.py`** | **hangs (>180s, no output)** |
| symlink to `/etc/passwd` | ok |
| self-referential symlink loop | ok |
| 5000-deep nested expression | ok |
| invalid syntax | ok |
| newline in filename | ok |

A unix socket fails differently — `ENXIO` on open — for the same underlying
reason: it is not a regular file.

## Why the check goes here

I first patched the reader I found, re-ran, still hung; found a second reader,
patched it, still hung; found a third, patched it, **still hung**. Four
independent readers surfaced in sequence (`collect_files`, `detect.count_words`,
`extractors/engine.py`, `extractors/resolution.py`), and a sweep counts 500+
file-read call sites across the extractors. Point-patching cannot converge.

So the predicate sits at the one place where a path is admitted to the corpus,
alongside the existing `_resolves_under_root` and `_is_sensitive` rules, and
records the skip in `skipped_sensitive` like every other rejection.

`os.stat` follows symlinks on purpose: a link *pointing at* a FIFO blocks
exactly like the FIFO does. A path that cannot be stat'ed is skipped rather
than raising.

## Verification

End to end, same repo, 60s budget:

| tree | exit | artefacts |
|---|---|---|
| v8 (`00efd6e`) | **124 — hang** | 0 |
| with this change | 0 (<1s) | graph.json + manifest.json |

A normal repository produces a byte-identical graph — 5 nodes / 5 edges on
both trees, so this is not filtering anything that used to be indexed.

Full suite on this branch: **18 failed, 3758 passed**. Baseline on a pristine
`v8` worktree: **19 failed, 3748 passed**. Diffing the FAILED lists shows
**zero new failures** (the pre-existing ones are in `test_terraform.py`,
`test_ollama_retry_cap.py`, `test_cache.py` and friends, unrelated to this
change).

The 9 new tests fail on `v8` without the fix — collection errors on the
missing symbol — so they are not vacuous. They cover: regular file accepted,
FIFO rejected, unix socket rejected, directory named `.py` rejected, symlink
to a regular file accepted, symlink to a FIFO rejected, broken symlink and
missing path rejected without raising, char device rejected.

## Note

I could not find a working private channel for this. `SECURITY.md` says to use
GitHub private vulnerability reporting, but #2011 (open since 18 July, no
response) reports it returns HTTP 500 on submit, and `GET /security-advisories`
is empty. Since the impact is availability-only, needs a hostile or unusual
repository, and the fix is public anyway, a normal PR seemed right — happy to
move it if you would prefer otherwise.

### DIFF
--- a/graphify/detect.py
+++ b/graphify/detect.py
@@ -5,6 +5,7 @@ import json
 import os
 import re
 import shlex
+import stat
 from concurrent.futures import ThreadPoolExecutor
 from enum import Enum
 from functools import lru_cache
@@ -779,10 +780,35 @@ def count_words(path: Path) -> int:
             return len(docx_to_markdown(path).split())
         if ext == ".xlsx":
             return len(xlsx_to_markdown(path).split())
+        # Only regular files may be opened. A repository can contain named
+        # pipes, sockets and device nodes, and `clone <github-url>` exists to
+        # point the scan at trees the operator did not write. open() on a FIFO
+        # with no writer BLOCKS FOREVER — it never raises, so the except below
+        # cannot help, and `graphify update` hangs with no output. os.stat
+        # follows symlinks on purpose: a link pointing at a FIFO blocks exactly
+        # like the FIFO itself.
+        if not stat.S_ISREG(os.stat(_os_path(path)).st_mode):
+            return 0
         with open(_os_path(path), encoding="utf-8", errors="ignore") as f:
             return len(f.read().split())
     except Exception:
         return 0
+
+
+def _is_regular_file(path: Path) -> bool:
+    """True only for regular files (symlinks followed).
+
+    Named pipes, sockets and device nodes must never reach a reader:
+    ``open()`` on a FIFO with no writer blocks forever and never raises, so a
+    single ``pipe.py`` in a scanned repository hangs the run with no output.
+    A symlink is resolved deliberately — a link pointing at a FIFO blocks
+    exactly like the FIFO itself. A path that cannot be stat'ed is treated as
+    not readable rather than raising.
+    """
+    try:
+        return stat.S_ISREG(os.stat(path).st_mode)
+    except OSError:
+        return False
 
 
 # Directory names to always skip - venvs, caches, build artifacts, deps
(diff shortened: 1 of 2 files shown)


itskaism:
Re-verified against **v0.9.33** (released 2026-08-05, after this PR was opened) — still reproducible, and the admission gate still has no regular-file check.

**Empirical reproduction with a control**, on a clean `uv tool install graphifyy` (0.9.33):

```
control repo (main.py only)        -> exit 0, graph.json written, 3 nodes
same repo + mkfifo pipe.py         -> exit 124 (timed out, no output)
```

Same corpus, same command (`graphify extract <dir> --code-only --no-cluster`), one FIFO apart. The control is what makes this meaningful: it isn't a slow scan, it's an unbounded block.

**Why the gate is still the right place**, confirmed on the shipped 0.9.33 source (`site-packages/graphify/detect.py`, 1932 lines):

```
_is_regular_file   -> absent
stat.S_ISREG       -> absent
import stat        -> absent
```

The admission loop already filters two other classes at exactly this point:

```
1399: skipped_sensitive.append(str(p) + " [symlink target outside scan root]")
1402: skipped_sensitive.append(str(p))          # _is_sensitive
```

so a third `[not a regular file]` sits naturally alongside them, and `count_words` (still unguarded — `S_ISREG` absent from its body in 0.9.33) stops being reachable with a blocking path.

**On why per-reader patching cannot work here** — I tried that first and it failed four times. Patching `collect_files` alone still hung; then `detect.count_words`; then `extractors/engine.py` `Path.read_bytes`; then `extractors/resolution.py`. A grep counted **522 file-read call sites** across the package. Every one of them is already wrapped in `try/except Exception`, which is useless for this bug: `open()` on a FIFO with no writer **blocks rather than raises**. Verified at kernel level on the stuck child — `/proc/<pid>/wchan` = `wait_for_partner`, `/proc/<pid>/syscall` = `257` (openat). One gate, not 522 readers.

Finding the readers at all needed a `sitecustomize.py` shim on `PYTHONPATH` hooking `builtins.open`, `os.open`, `io.open`, `Path.read_text` and `Path.read_bytes`, because the blocking read happens in a **worker child** — in-process tracing of the parent showed nothing.

**Class coverage**, both symlink postures (`follow_symlinks` False and True): FIFO blocks forever, unix socket raises ENXIO; char devices and directories named `*.py` were already filtered upstream of this. After the gate: 0/6 non-regular shapes reach a parser, `control.py` still collected, 3/3 runs finish in 0–1s where unpatched hangs at 124.

Severity is the shipped `graphify clone <github-url>` path, whose whole purpose is pointing the extractor at a tree the operator did not write, plus `watch` mode where the hang is silent and permanent.

Happy to rebase if the branch has drifted.

safishamsi:
Shipped in v0.9.42 (`graphifyy==0.9.42` on PyPI). Non-regular files (FIFO/device) are skipped so they can't hang extraction. Credited in the release notes. Thanks @itskaism!

PR #2464, opened 2026-08-04
fix(cache): a same-size rewrite in one filesystem tick reused a stale digest

itskaism:
`file_hash` memoises on `(size, mtime_ns)` to skip re-reading unchanged files.
That assumes a content change always moves one of the two. It doesn't — an
edit that keeps the file the same length and lands inside a single filesystem
timestamp tick is invisible to `stat`, so the previous digest is served:

```python
f.write_text("# A\n\nContent A.\n"); h1 = file_hash(f, root)
f.write_text("# B\n\nContent B.\n"); h2 = file_hash(f, root)
h1 == h2   # different documents, identical digest
```

Everything keyed on that digest then treats the new content as the old: the
semantic cache serves stale nodes, incremental extraction skips the file, and
`prune_semantic_cache` cannot see the orphan it should collect.

## How it surfaced

`tests/test_cache.py::test_semantic_prune_removes_orphan_entries` and
`::test_semantic_prune_sweeps_both_namespaces_against_same_live_set` fail on
`v8` today — `prune_semantic_cache` returns 0 where 1 is expected. The prune
logic is correct; there was only ever one cache entry, because the two writes
produced the same hash.

The realistic trigger is `watch` mode and any fast edit loop, and it gets
worse on filesystems that round mtime to whole seconds (HFS+, some network
mounts, containers with reduced clock resolution) where the collision window is
a full second rather than a few milliseconds.

## The fix

The memo now records the moment it was taken (`seen_ns`) and is distrusted only
while the file's mtime sits inside that same tick — precisely the window where a
later same-size write can hide behind an unchanged mtime.

Granularity is derived from the timestamp rather than assumed: a whole-second
mtime means the filesystem cannot separate writes within that second, so the
window widens; genuine nanosecond precision keeps it at a few milliseconds and
ordinary work keeps the fastpath.

Entries restored from disk have `seen_ns` cleared at load time. They come from a
completed run, so the ambiguity cannot apply to them — which is what keeps a
corpus relocated with `copy2` 100% warm, with no extra reads and no re-billed
extraction.

## Verification

| case | before | after |
|---|---|---|
| same-size rewrite in one tick | same digest | digest changes |
| `prune_semantic_cache` orphan | 0 pruned | 1 pruned |
| aged, untouched file | fastpath | fastpath (0 reads) |
| corpus moved with `copy2` | warm | warm (0 reads) |
| file grows | invalidates | invalidates |

Full suite on this branch: **3908 passed**, 33 skipped. On a pristine `v8`
worktree with the same interpreter and extras installed: **5 failed**. The
stat-index portability tests (`test_cache_hits_survive_corpus_move`,
`test_out_of_root_key_round_trips_absolute`) pass unchanged — those were the
constraint that ruled out the simpler "distrust any recent mtime" approach.

The 5 new tests fail on `v8` without the fix.

## Note on a related flake

While verifying this I found
`test_extract_cli.py::test_incremental_partial_run_preserves_untouched_semantic_hash`
fails **non-deterministically on `v8` itself** (about 2 runs in 6, unpatched) —
the manifest shows two files sharing one `mtime` with `ast_hash == semantic_hash`.
Same underlying class of problem, different code path. Not addressed here; happy
to open it separately if useful.

### DIFF
--- a/graphify/cache.py
+++ b/graphify/cache.py
@@ -7,6 +7,7 @@ import json
 import os
 import re
 import tempfile
+import time
 import warnings
 from collections.abc import Iterable
 from pathlib import Path
@@ -74,6 +75,14 @@ def _cleanup_stale_ast_entries(ast_base: Path, current_dir: Path) -> None:
 # actually changed. Entries live under cache/semantic/p{fingerprint}/ when the
 # caller supplies its prompt; callers that don't keep the historical flat layout.
 _PROMPT_FP_LEN = 12
+
+#: How close a memo's own timestamp may sit to the file's mtime before the
+#: (size, mtime_ns) fastpath stops proving content identity. The coarse value
+#: covers filesystems that round mtime to whole seconds (HFS+, some network
+#: mounts); the sub-second value applies when real nanosecond precision is
+#: reported, where only writes a few milliseconds apart can share a tick.
+_MTIME_GRANULARITY_NS = 2_000_000_000
+_MTIME_SUBSECOND_NS = 50_000_000
 
 # Count of pre-fingerprint (flat-layout) entries served this process, so
 # check_semantic_cache can report N to the user (#1939).
@@ -257,6 +266,13 @@ def _ensure_stat_index(root: Path, cache_root: "Path | None" = None) -> None:
                 for k, v in raw.items():
                     if not isinstance(k, str):
                         continue
+                    # Entries read back from disk were written by a COMPLETED
+                    # run, so the same-tick ambiguity that makes an in-process
+                    # memo untrustworthy cannot apply: any later write would
+                    # have to move size or mtime. Clearing seen_ns marks them
+                    # safe and keeps a moved/cloned corpus 100% warm.
+                    if isinstance(v, dict):
+                        v.pop("seen_ns", None)
                     if Path(k).is_absolute():
                         # Legacy/out-of-anchor key: pass through, but never
                         # clobber a re-anchored relative (new-format) entry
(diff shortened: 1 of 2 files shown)


itskaism:
Closing in favour of #2385, which reported the same defect two days earlier and fixes it more thoroughly.

I found this independently and opened it without searching the tracker for prior art first — my mistake, and the reason this is a duplicate. #2385 is the one to keep.

Comparing the two on the merits, not on filing order:

- **Same defect.** Both address a same-size rewrite inside one filesystem tick reusing a stale digest, both in `graphify/cache.py`.
- **@kavioavio's guard is the more general one.** `_stat_is_stable()` gates the fastpath on how old the timestamp is (with a ctime branch for restored artefacts carrying a future mtime). Mine keyed off a per-entry `seen_ns` marker, which only covers memos written by the *current* process — it cannot help a file that was already volatile when the index was loaded.
- **It also separates hash and word-count volatility**, which mine did not; a refreshed hash promoting a stale sibling count is a real failure mode I missed.

One detail from my patch that may be worth folding in, offered as a note rather than a competing change: entries re-read from disk in `_ensure_stat_index` were written by a *completed* run, so the same-tick ambiguity cannot apply to them — any later write has to move size or mtime. Treating them as immediately trustworthy keeps a moved or cloned corpus 100% warm on the first run after load. #2385 doesn't touch `_ensure_stat_index`, so if `_stat_is_stable()` is consulted for those entries too, a freshly restored tree could re-hash every file whose mtime happens to be recent. That may well be the intended trade-off — flagging it in case it isn't.

Sorry for the noise, @kavioavio — nice catch on the independent volatility tracking.

PR #2465, opened 2026-08-04
fix(manifest): apm.yml fallback parser dropped the package version

itskaism:
`_parse_apm` returns `name`, `version` and `deps` when PyYAML is installed.
PyYAML is not a hard dependency, so on any machine without it the line-based
fallback runs instead — and it hardcoded `"version": None`, so the package node
silently lost its version.

```python
>>> _parse_apm_fallback("name: my-pkg\nversion: 1.2.3\ndependencies:\n  - dep-a\n")
{'name': 'my-pkg', 'version': None, 'deps': ['dep-a']}
```

The two parsers are meant to agree; only this field diverged.

This is what makes
`tests/test_manifest_ingest.py::test_apm_parses_name_and_deps` fail with
`KeyError: 'version'` on a PyYAML-less install — the node builder drops the key
entirely when the value is `None`, so the failure surfaces one layer away from
the cause.

## The fix

The fallback reads a top-level `version:` exactly the way it already reads
`name:`, unwrapping quotes and stopping at a trailing comment. A `version:`
nested inside the `dependencies:` block still belongs to a dependency and is
not mistaken for the package version.

## Verification

7 new tests covering: version returned, name and deps unaffected, quoted
version unwrapped, trailing comment stripped, manifest without a version still
parses, a dependency's nested version not picked up, and a manifest with no
name still rejected.

3 of them fail on `v8` without the fix. `tests/test_manifest_ingest.py` goes
from 1 failed / 7 passed to **8 passed**.

### DIFF
--- a/graphify/manifest_ingest.py
+++ b/graphify/manifest_ingest.py
@@ -143,8 +143,10 @@ def _parse_apm(text: str) -> dict | None:
 
 def _parse_apm_fallback(text: str) -> dict | None:
     """Minimal line parser for apm.yml when PyYAML is unavailable: a top-level
-    ``name:`` plus a simple ``dependencies:`` block (list items or a name map)."""
+    ``name:``/``version:`` plus a simple ``dependencies:`` block (list items or
+    a name map)."""
     name = None
+    version = None
     deps: list[str] = []
     in_deps = False
     for line in text.splitlines():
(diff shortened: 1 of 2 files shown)


itskaism:
Prior art, flagged for the maintainer's benefit: **#1812** by @maikramer reported this same defect on 12 July, three weeks before this PR. I opened this without searching the tracker first — that was my mistake, and @maikramer has priority on the finding.

I'm leaving this one open rather than closing it, because #1812 is currently not landable in its present shape: its branch appears to have drifted out of sync, so the diff carries **730 files / +230,725 −5,128** (`.dockerignore`, `.github/FUNDING.yml`, `ci.yml` and the rest of the tree alongside the actual fix) and GitHub reports it as `dirty`. It hasn't been updated since the day it was opened.

The one-line fix inside it is correct and is the same change as here:

```python
# _parse_apm_fallback previously hardcoded "version": None
m = re.match(r'^version:\s*["\']?([^"\'\s#]+)', line)
```

This PR is that change scoped to 2 files (+58 −2) with a regression test, so it can be applied directly. **If @maikramer resyncs #1812, please take theirs** — the finding is theirs and I'd rather the credit follow the report. Happy to close this the moment that happens.

@maikramer — apologies for the duplicate; if you'd prefer to rebase #1812 onto a clean branch I'll close this straight away.

safishamsi:
Shipped in v0.9.42 (`graphifyy==0.9.42` on PyPI). apm.yml fallback parser captures the package version. Credited in the release notes. Thanks @itskaism!

PR #2466, opened 2026-08-04
fix(detect): an incremental run skipped a file whose mtime did not move

itskaism:
`detect_incremental` documents its gate as:

> Fast path: mtime unchanged + hash matches → unchanged

but the mtime-unchanged branch never consults the hash — it returns `unchanged`
without reading the file:

```python
if stored_mtime is None or current_mtime != stored_mtime:
    changed = _md5_file(Path(f)) != stored_hash
else:
    changed = False          # no content check at all
```

That is sound only while the filesystem can separate the two writes. An edit
that keeps the file the same length and lands inside one timestamp tick moves
neither size nor mtime, so the file is classified unchanged, never
re-extracted, and the graph keeps serving the old content.

Measured on ext4 — a same-size rewrite immediately after the first write:

```
200 trials
  mtime UNCHANGED (collision): 200
  mtime moved                : 0
```

## This is already failing on v8

Two tests flap today, and they fail exactly when the two writes share a tick:

| test | failure rate on pristine `v8` |
|---|---|
| `test_truncated_doc_semantic_hash_is_cleared_for_requeue` | **8/20** |
| `test_incremental_partial_run_preserves_untouched_semantic_hash` | **3/20** |

Both pass 25/25 when run alone, which is why it reads as a flake rather than a
bug — in isolation the timing rarely collides.

## The fix

`save_manifest` stamps each row with `seen`, the moment it was written. The
mtime-unchanged branch pays for one MD5 only while the file's mtime sits inside
that same tick — the one window where a later same-length edit can hide.

Granularity is derived from the timestamp rather than assumed: a whole-second
mtime means the filesystem cannot separate writes within that second, so the
window widens; real sub-second precision keeps it at 50ms.

Rows predating `seen` come from an earlier process, where any later write would
have had to move mtime, so they keep the free fastpath.

## Verification

The fastpath is the point of this gate, so it is measured, not assumed:

| case | content hashes | re-queued |
|---|---|---|
| settled corpus, manifest from an earlier run | **0** | 0 |
| one file genuinely edited (size changes) | 1 | 1 (correct file) |
| that file's untouched neighbours | — | 0 |
| same-size rewrite in one tick | 1 | **1** (was 0) |
| legacy row without `seen` | 0 | 0 |

`tests/test_extract_cli.py` goes from **8/10 failing runs to 8/8 clean**. Full
suite: **3924 passed**, 33 skipped.

The 5 new tests fail on `v8` without the fix.

## Relation to #2464

Same class, different layer. #2464 fixes `file_hash`'s `(size, mtime_ns)` memo;
this fixes the manifest-level gate in `detect_incremental`. Neither subsumes the
other — measured on three worktrees, 12 runs of `test_extract_cli.py` each:

| tree | failing runs |
|---|---|
| pristine `v8` | 7/12 |
| `v8` + #2464 (cache) only | **7/12** |
| `v8` + this PR only | **0/12** |

So they are independent and can land in either order.

### DIFF
--- a/graphify/detect.py
+++ b/graphify/detect.py
@@ -5,6 +5,7 @@ import json
 import os
 import re
 import shlex
+import time
 from concurrent.futures import ThreadPoolExecutor
 from enum import Enum
 from functools import lru_cache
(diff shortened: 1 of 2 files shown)


safishamsi:
Shipped in v0.9.42 (`graphifyy==0.9.42` on PyPI). Incremental runs re-queue a same-tick same-length rewrite (complements #2612). Credited in the release notes. Thanks @itskaism!

Issue #1522, opened 2026-06-28
Path-based IDs still collide after #1504 fix: `normalize_id` collapses non-word sequences to `_`


sub4biz:
Hi @safishamsi — first, thank you for the quick turnaround on #1504 and #1509 — the full-path stem approach is exactly the right direction and the `_semantic_id_remap()` migration utility is a nice touch.

While smoke-testing v0.9.0 I noticed that the collision problem may be wider than the original issue. The fix correctly uses the full repo-relative path, but `normalize_id` collapses any contiguous sequence of non-word characters (`[^\w]+`) into a single `_`, and then collapses consecutive underscores. That affects a large set of characters valid in directory and file names: `-`, `.`, ` `, `!`, `(`, `@`, `#`, `~`, `+` and others on Linux/macOS, and a similar set on Windows (where only `\/:*?"<>|` are forbidden). Any combination of these in path segments can produce the same ID even after the fix.

**Cross-directory collisions (slash vs other separators):**
```
foo_bar/baz.md    →  foo_bar/baz   →  foo_bar_baz
foo/bar_baz.md    →  foo/bar_baz   →  foo_bar_baz   ← slash vs underscore

my-module/baz.md  →  my-module/baz →  my_module_baz
my_module/baz.md  →  my_module/baz →  my_module_baz ← dash vs underscore

my module/baz.md  →  my module/baz →  my_module_baz ← space vs underscore (space matches [^\w]+, becomes _)

utils.helper/baz.md  →  utils.helper/baz  →  utils_helper_baz
utils/helper_baz.md  →  utils/helper_baz  →  utils_helper_baz ← dot vs slash vs underscore
```

**Same-directory collisions (dot/dash in filename vs underscore):**

`with_suffix("")` removes only the last extension, so intermediate dots survive into the stem:

```
foo.bar.ts  →  foo.bar  →  foo_bar
foo_bar.ts  →  foo_bar  →  foo_bar   ← collision, same directory
foo-bar.ts  →  foo-bar  →  foo_bar   ← collision, same directory
```

Three different files in the same folder produce identical IDs.

**Consecutive non-word characters:**

A contiguous run of non-word characters (`[^\w]+`) collapses to a single `_`, so quantity doesn't matter either:
```
foo--bar/baz.md  →  foo_bar_baz
foo-bar/baz.md   →  foo_bar_baz   ← collision (double dash vs single)
foo_bar/baz.md   →  foo_bar_baz   ← collision
```

[...]

PR #2473, opened 2026-08-05
fix(serve): treat `_` as a token separator so underscore queries match hyphenated labels

nadiadatepe-eng:
### The bug

`_search_tokens` splits on `\w+`. Python's `\w` counts `_` as a *word character* but not `-`,
so the two spellings tokenize inconsistently:

| input | tokens |
|---|---|
| `graph-first-guard.py` | `graph`, `first`, `guard`, `py` |
| `graph_first_guard`    | `graph_first_guard` |

Both the query and the node label pass through this same function, so the two sides can never
meet: searching for the underscore spelling of a hyphenated label returns nothing.

Found while querying a graph for a file it had itself indexed — the graph could not find the
underscore spelling of its own `local_id`.

### The fix

One character class: `\w+` → `[^\W_]+`. `_` becomes a separator, exactly like `-`.

Because *both* sides pass through `_search_tokens`, plain snake_case lookups still resolve —
query and label simply split the same way. The second test below pins that.

### Tests

- `test_underscore_query_matches_hyphenated_label` — the reported case; asserts the hyphenated
  query still matches (it always did) **and** that the underscored query now resolves to the
  same node.
- `test_snake_case_identifier_still_matches_itself` — guards the obvious regression risk: that
  splitting on `_` breaks snake_case search.

`pytest tests/test_serve.py` → **128 passed** on this branch.

Verified red-green against this branch's base: with `\w+` the first test fails; with the fix
both pass. The snake_case guard passes in *both* states — which is what a regression guard
should do.

### Scope

Deliberately minimal — one function, one character class, no behaviour change to ranking,
Chinese handling, or diacritics stripping.

### Related

Same family as #920 (`\b` boundary missing underscore-separated names) and #1522
(`normalize_id` collapsing non-word sequences to `_`) — both closed. This is the query-side
instance of that class: `\w` and `\b` disagree with `-` about what a word is.

### DIFF
--- a/graphify/serve.py
+++ b/graphify/serve.py
@@ -164,8 +164,17 @@ def _strip_diacritics(text: str | None) -> str:
 
 
 def _search_tokens(text: str) -> list[str]:
-    """Split text into word tokens, stripping punctuation and diacritics."""
-    return re.findall(r"\w+", _strip_diacritics(str(text)).lower())
+    """Split text into word tokens, stripping punctuation and diacritics.
+
+    `_` is a separator, exactly like `-`. `\\w` counts underscore as a word
+    character but not hyphen, so `graph_first_guard` stayed one token while the
+    label `graph-first-guard.py` split into three — and the query matched
+    nothing. Both the query and the node label pass through here, so splitting
+    on `_` keeps the two sides consistent and snake_case lookups still resolve
+    (their tokens simply match the same way). Found 2026-07-29: the graph could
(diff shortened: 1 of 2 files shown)


safishamsi:
Landed in v0.9.46, just published to PyPI. Cherry-picked onto `v8` with your authorship preserved. `query` now treats `_` as a token separator so underscore-spelled queries match hyphenated labels. Thanks @nadiadatepe-eng. Closing since it is now released.

Issue #2468, opened 2026-08-05
_is_ignored: unanchored parent-.gitignore pattern matching the scan root's own directory name silently zeroes the corpus (detect_incremental returns 0 files)


alexrdunbar:
## Summary

When the scan root's own directory name is also the subject of an **unanchored** ignore pattern in a **parent** `.gitignore`, `_is_ignored`'s anchor-fallback branch matches that pattern against every path in the scan. Every file in the corpus is classified as ignored, and `detect()` / `detect_incremental()` report **0 files** with no warning.

This is the normal, correct way to write the ignore rule from git's point of view (`graphify-corpus/` in the repo root `.gitignore` keeps a regenerable corpus out of the repo), so the failure hits exactly the setup the tool is meant to support.

Version: `graphifyy` 0.9.12, Python 3.14, Windows.

## Repro

1. A git repo with a directory `graphify-corpus/` at its root, holding the corpus you want to scan.
2. The repo root `.gitignore` contains an unanchored line:

   ```gitignore
   # Graphify (regenerable knowledge graph)
   graphify-corpus/
   graphify-out/
   ```

3. Run against that directory:

   ```python
   from pathlib import Path
   from graphify.detect import detect_incremental
   r = detect_incremental(Path('graphify-corpus'))
   print(r['total_files'], r['new_total'])   # -> 0 0
   ```

Result: `total_files = 0`, `new_total = 0`, and `deleted_files` lists every entry from the previous manifest with a **doubled path prefix**, e.g.

```
C:\...\claude-code\graphify-corpus\graphify-corpus\claude-scripts\pm2-monit.ps1
C:\...\claude-code\graphify-corpus\graphify-corpus\claude-skills\autoresearch\autoresearch.py
```

The doubled prefix is the visible tell that the scan root's name is being prepended a second time during matching.

## Cause

`_load_graphifyignore` walks up to the VCS root and returns `(anchor_dir, pattern)` pairs, so the repo-root `.gitignore` contributes `(<repo_root>, "graphify-corpus/")`.

In `detect.py::_is_ignored._eval`, the unanchored branch (currently `detect.py:896-907`) falls back to re-matching against the pattern's own anchor directory when `anchor != root`:

```python
else:
    try:
        rel = str(target.relative_to(root)).replace(os.sep, "/")
        matched = _matches(rel, p, anchored=False)
    except ValueError:
        pass
    if not matched and anchor != root:
        try:
[...]

PR #2475, opened 2026-08-05
fix: preserve explicit scan roots when matching parent unanchored .gitignore patterns

hopstreax:
## Summary

Fixes Issue #2468 by preventing unanchored parent `.gitignore` patterns from incorrectly matching the scan root itself when the user explicitly scans a subdirectory of a Git repository.

Previously, when a parent `.gitignore` contained an unanchored directory rule matching the scan root (for example, `graphify-corpus/`), Graphify evaluated the path relative to the parent ignore anchor. This allowed the scan root's own directory name to participate in unanchored basename matching, causing descendant directories to be incorrectly pruned during traversal.

The fix preserves existing ignore semantics while ensuring that unanchored parent patterns are evaluated relative to the scan root in this specific scenario, preventing the scan root itself from becoming a match candidate.

### Changes

* Adjust unanchored parent ignore matching in `_is_ignored()` to evaluate scan-root descendants using scan-root-relative paths when appropriate.
* Preserve existing behavior for:

  * anchored ignore patterns
  * nested `.gitignore` files
  * directory-only patterns
  * wildcard and basename matching
  * negation rules

### Tests

Added regression tests covering:

1. Scanning a directory whose name is excluded by an unanchored parent `.gitignore` rule no longer suppresses traversal of the explicitly requested scan root.
2. Parent `.gitignore` rules continue to ignore unrelated directories (for example, `node_modules/`) inside the scan root, ensuring existing ignore semantics are preserved.

### Validation

* New regression tests pass.
* Existing ignore-related tests continue to pass.
* Production change is limited to a small, localized modification in `graphify/detect.py`.

### DIFF
--- a/graphify/detect.py
+++ b/graphify/detect.py
@@ -1169,7 +1169,14 @@ def _is_ignored(
             except ValueError:
                 continue  # target outside this pattern's anchor: cannot match
             if rel_anchor != ".":
-                matched = _matches(rel_anchor, p, path_relative=path_relative)
+                rel = rel_anchor
+                if not path_relative:
+                    try:
+                        if len(root.parts) > len(anchor.parts):
+                            rel = str(target.relative_to(root)).replace(os.sep, "/")
+                    except ValueError:
+                        pass
+                matched = _matches(rel, p, path_relative=path_relative)
                 if matched and directory_only and not target.is_dir():
                     matched = False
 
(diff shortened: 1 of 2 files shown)


safishamsi:
Shipped in v0.9.41 (`graphifyy==0.9.41` on PyPI) (#2468), with an NFC follow-up. Credited in the release notes. Thanks @hopstreax!

PR #2476, opened 2026-08-05
Feat/hdl vhdl tcl extractors

3brahimi:


3brahimi:
@safishamsi could you please take a look and see if you could merge this to the next release?

PR #11, opened 2026-04-06
Add Lua language support

itsDNNS:
## Summary

- Adds `.lua` and `.toc` to `CODE_EXTENSIONS` in `detect.py`
- Adds `_LUA_CONFIG` LanguageConfig and `extract_lua()` in `extract.py`
- Handles Lua-specific patterns: `local function`, `function Addon:Method()`, `require('module')`
- Adds `tree-sitter-lua` to `pyproject.toml` dependencies
- Updates README language table and skill.md code extensions

## Motivation

Lua is the scripting language for World of Warcraft addons, Roblox, Neovim plugins, and game engines (Love2D, Defold). `tree-sitter-lua` is mature (v0.5.0 on PyPI) and the generic extractor handles Lua well since functions use a `name` field and `block` body consistently.

## Test plan

- [x] Tested on a 27-file WoW addon codebase (mixed local functions, method syntax, require imports)
- [x] Verified function declarations, method patterns (`Addon:Method`), require() imports, and call-graph extraction
- [x] Confirmed `.toc` files (WoW addon manifests) are detected as code

PR #2492, opened 2026-08-05
fix(php): resolve instance-method calls edges from typed receivers (#1682)

### LINKED ISSUE #1682: PHP: member/instance method calls never resolve to calls edges — only static Class::method() works
### Summary
For PHP codebases, `calls` edges are only ever generated for static method calls
(`ClassName::method()`). Instance-method calls — both constructor-injected
dependencies and inline `new ClassName()` — never produce a `calls` edge, even
though the callee class is statically knowable in both cases.

This matters a lot for the "what depends on X / what breaks if I change X" use
case, which is one of the main reasons people point an AI coding assistant at a
graphify graph instead of grepping. In a modern Laravel codebase, constructor
injection is the dominant call pattern (it's the framework's own recommended
style), so this gap silently blanks out cross-file call edges for most
non-trivial service classes.

### Repro (graphify 0.8.30, tree_sitter_php)

Three real classes in the same codebase, same MCP server, same freshly-built
(current-HEAD) graph:

1. **Static call — works**
   ```php
   class SucursalContext {
       public static function id() { ... }
   }
   // called elsewhere as: SucursalContext::id()
get_neighbors("SucursalContext") → 25 correct inbound calls [INFERRED] edges.

Constructor-injected instance call — fails


class LeadController extends Controller {
    public function __construct(protected LeadHunterService $leadHunter) {}
    // ...$this->leadHunter->search(...) elsewhere in the class
}
get_neighbors("LeadHunterService") → 0 inbound calls edges, despite confirmed
real usage in LeadController.php (verified via grep).

Inline instantiation — fails


$paymentResult = (new \App\Services\MixedPaymentService())->resolve(...);
get_neighbors("MixedPaymentService") → 0 inbound calls edges, despite the
call site existing and being active code.

What I found reading the source
extract.py (PHP walk_calls branch, _PHP_CONFIG.call_types includes member_call_expression) does visit member-call nodes, but only extracts the bare method name from the name field — it never inspects the object field ($this, a property, a new X() expression), so there's no attempt to resolve which class the call belongs to.
Even if that were resolved, the cross-file resolution pass unconditionally drops any raw call flagged is_member_call, with a comment explaining this is intentional — to avoid false positives from common method names (log(), execute(), find()) colliding across unrelated classes when there's no type information to disambiguate.
So the gap is understandable as a conservative anti-false-positive choice, but
it means calls edges are effectively silent for the majority pattern of
idiomatic OOP/DI PHP code, with no signal to the caller that the edge set is
incomplete for that reason.

Suggested direction (not prescriptive — happy to be told this is out of scope)
Resolving $this->prop->method() and (new ClassName())->method() for
concretely-typed properties/constructor params looks feasible (a per-file map
of property-name → declared type from constructor-promoted params and typed
property declarations), while leaving interface-typed properties and
genuinely ambiguous cases unresolved rather than guessing. Existing
bound_to/service-container-binding edges elsewhere in the extractor might
already carry some of the type information needed for the interface case.

Even without a full fix, surfacing this as a known limitation (e.g. a
"member-call edges best-effort / static-only for PHP" note in the docs or in
graph_stats) would help — right now the graph "looks complete" (rich
inherits/method/contains edges) while silently missing what's often the
most useful edge type for impact analysis.

Environment
graphify 0.8.30 (graphifyy on PyPI)
Language: PHP 8, Laravel, tree_sitter_php
Reproduced via both get_neighbors and query_graph (MCP tools)

filipechagas:
Fixes #1682.

## The problem

The graph makes `calls` edges only for static calls. A static call looks like this: `ClassName::method()`.

The graph does not make `calls` edges for instance calls. Instance calls include these three types:

1. Calls through a service that the constructor receives: `$this->service->method()`
2. Calls through a local variable that has a type
3. Calls on a new object made in the same line: `(new Service())->method()`

The graph does not make these edges. This is true even when the code clearly shows the class of the callee. Modern Laravel code almost always uses constructor injection. So, for most service classes, the graph shows zero incoming calls. But the graph looks complete. The user cannot see that these edges are missing.

## The solution

Two parts:

**Part 1 — The extractor records the receiver.** The receiver is the object before the arrow in `$obj->method()`. The PHP extractor now records it.

**Part 2 — A new PHP resolver makes the edge.** The resolver reads the receiver and its declared type. Then it makes a `calls` edge. The resolver is registered in the language-resolver registry. Because of this, it runs in full builds and in incremental builds.

The resolver makes an edge only when there is no doubt. It copies the structure and the guards of the Java member-call resolver. It obeys three conditions:

1. The receiver has exactly one known concrete type.
2. The corpus has exactly one definition of that type.
3. That definition has exactly one method with that name.

If one condition fails, the resolver makes no edge. It does not fall back to a name-only match.

**These calls now resolve:**

- Calls through constructor-promoted properties
- Calls through typed properties
- Calls through typed local variables
- Calls through typed parameters
- Calls through nullable types (`?Foo` counts as `Foo`)
- Nullsafe calls (`?->`)

Each new edge has a confidence label. An edge from a declared type gets `INFERRED` (0.8). An edge from an inline `new` gets `EXTRACTED` (1.0) — but only when the written full class name agrees with the namespace that the file declares. If not, the edge gets `INFERRED`.

**These calls do not resolve. The resolver refuses them:**

- Receivers with no type
- Receivers with a union type or an intersection type
- Receivers with an interface type, an enum type, or a trait type — the resolver never guesses the implementation. This holds when two names collide, and it holds in incremental builds.
- Types with two or more definitions in the corpus
- Methods that the class does not define — the resolver does not guess magic dispatch
- Chained receivers and array receivers
- Anonymous classes
- `self`, `static`, and `parent`
- Local variables that lose their known type. A variable loses its type when: the code assigns it again; the code assigns it a different `new` type; a `global` or `static` statement rebinds it; a closure parameter, an arrow-function parameter, a `foreach` target, or list destructuring shadows it.

**The change also includes two hardening fixes:**

1. Raw calls that have a language tag cannot go through the member-call resolvers of other languages.
2. The PHP and ObjC resolvers now read type definitions only from files of their own language. For ObjC, this repairs an old defect in mixed-language projects. Before, ObjC calls could bind to a class from a different language. Also before, a name collision with a different language removed correct ObjC edges. Now those edges come back.

**Two behavior changes to know (the changelog also lists them):**

1. Some same-file calls change their label. If the receiver has a known concrete type, the edge was `EXTRACTED` from a name match. Now the edge is `INFERRED` from the resolver. The edge connects the same two nodes.
2. The version goes up to 0.9.34. The AST cache uses the version as its namespace, so the new version makes old cache entries invalid. But this alone does not refresh an existing graph. To get the new edges in an existing graph, run `graphify update .`

## Evidence

We ran the reproduction corpus from the issue. It has a Laravel-style controller and services. The table shows each call site before and after:

| Call site | Base 4e7e6b1 | This branch |
|---|---|---|
| `$this->leadHunter->search(...)` (promoted ctor param) | no edge | `calls` INFERRED 0.8 |
| `(new \App\Services\MixedPaymentService())->resolve(...)` | no edge | `calls` EXTRACTED 1.0 |
| `SucursalContext::id()` (static control) | `calls` edge | unchanged |

- Tests: the base has 3933 passed / 36 skipped. This branch has **4024 passed / 36 skipped**. That is 91 more tests. All 91 tests use the public `extract()` function. Each positive test also has a decoy: a different class with a method of the same name. Each test makes sure the decoy gets **no** edge.
- We attacked the guards with approximately 20 hostile test fixtures. The attacks used: enums, traits, references, variable-variables, `new $cls()`, letter-case tricks, variable shadowing, heredocs, and match expressions. Three attacks found real defects. We fixed all three before this PR. Each fix has a regression test that failed before the fix.
- We made 80 assertions about the AST shapes that the resolver reads. All 80 pass on all twelve tree-sitter-php releases that the version floor permits (0.23.0 – 0.24.1).
- Full builds and incremental builds give the same result. They agree on each edge and each refusal. Watch-path tests hold this in place.

## Known limits (the changelog lists them)

- The resolver does not follow methods that come from traits or from parent classes. The Java resolver has the same limit.
- Enum methods cannot be call targets. This only removes possible edges. It never adds wrong ones.
- The resolver ignores docblock types.
- One wrong-edge case remains: a `use` alias points to a class outside the corpus, and the corpus has one unrelated class with the same short name. Then the edge binds to the wrong class. The Java resolver has this same defect today. The planned repair is `use`-map awareness.
- A union-typed receiver can still get a same-file, name-match edge. This behavior existed before this change. This change does not touch it.


filipechagas:
Status update from the submitting fork — three things reviewers may want to know:

**1. Follow-up fixes have landed on the fork since this PR was opened.** Independent review and release verification of this branch surfaced several issues; all are fixed, tested (red-first, full-suite gated), and merged on `lawnstarter/graphify`:

- Union/intersection-typed receivers minted a same-file bare-name edge (pre-existing at this PR's base, but adjacent): [lawnstarter#25](https://github.com/lawnstarter/graphify/pull/25)
- First-class callables `$obj->method(...)` now emit `indirect_call` instead of `calls`: [lawnstarter#27](https://github.com/lawnstarter/graphify/pull/27)
- PHP `use` import metadata + a decisive-refusal `PhpNameResolver` closing the use-alias false-positive this PR's design accepted as a known risk: [lawnstarter#29](https://github.com/lawnstarter/graphify/pull/29), [lawnstarter#31](https://github.com/lawnstarter/graphify/pull/31), [lawnstarter#32](https://github.com/lawnstarter/graphify/pull/32)

Those build on this PR's code, so we're deliberately **not** adding them here mid-review; we'll submit them as follow-up PRs once this merges.

**2. Two fixes that apply to upstream independently of this PR are being submitted separately** (they touch pre-existing code): the untagged member-call resolvers consuming each other's raw calls (a TS receiver can mint a Python edge today), and `use` metadata capture plus a group-form `use function`/`use const` name-claiming bug in `_resolve_php_type_references`.

**3. Version collision heads-up:** this branch bumps to 0.9.34, which upstream has since released independently (`07b9143`). Happy to renumber at merge time however maintainers prefer — we've avoided force-pushing during review.

filipechagas:
Working through the automated coupling threads on this PR. Both flags here are duplicated (posted twice), so there are two distinct findings; neither is a regression, and I don't think either warrants a change. Numbers below are from an AST scan I ran myself against the base commit `07b9143d` — the tool's counts and mine don't fully agree (it reports 6 callees for `_php_method_receiver_types`, I count 7, probably a different treatment of closures), so these are mine.

**`_php_method_receiver_types()` — `engine.py:761`, "fans out to 6 callees".** The function does not exist at `07b9143d`; it is introduced by this PR (`43cd7a2`, the commit that resolves `$this->prop->method()` to the property's declared type). Its span at this head is 761-887, 127 lines. There is no prior fan-out to regress from — a new 127-line helper calling ~7 things is the shape of the helper, not a delta. Happy to discuss whether the helper should be decomposed on its own merits, but that is a different conversation from a coupling regression.

**`_full_then_incremental()` — `tests/test_php_member_calls.py:1586`, "9 callers depend on it".** Also new in this PR, in a new file. It is the full-extract-then-incremental-rebuild parity harness: it runs both build paths over the same corpus and returns both results so a test can assert them *together*. Every incremental test has to share it, or the parity claim means nothing — a full-build-only assertion is green on the broken code, which is exactly the failure mode `#11` was about. The 9 callers are the point of it. All of it is confined to `tests/`; no production symbol's coupling changed.

Glad to restructure either if you'd rather, but I'd be trading a deliberate design for a number, so I'd want to know the gate is blocking first.

filipechagas:
We close this PR in favor of one consolidated PR. Seven stacked PRs became hard to review and hard to keep current with `v8`.

- Consolidated defect report: #2615 (a superset of #1682, which this PR fixed)
- Consolidated PR: #2617 — it contains everything this PR carried, plus its follow-ups, rebased on the current `v8` head.

PR #2494, opened 2026-08-06
fix(ingest): recognize Cargo.toml as a package manifest (#2434)

### LINKED ISSUE #2434: .toml is in no extension set, so config/data TOML is never collected — not even as a document
## Summary

`.toml` appears in none of `CODE_EXTENSIONS`, `DOC_EXTENSIONS`, `PAPER_EXTENSIONS` or `IMAGE_EXTENSIONS`, so it is absent from `_WATCHED_EXTENSIONS` and is never collected at all. Unlike `.yaml` — which at least reaches the semantic/LLM path as a document (#2043, #2214) — TOML has no route into the graph at any cost tier.

Verified on `graphifyy==0.9.32`:

```python
from graphify.detect import CODE_EXTENSIONS, DOC_EXTENSIONS, PAPER_EXTENSIONS, IMAGE_EXTENSIONS
watched = CODE_EXTENSIONS | DOC_EXTENSIONS | PAPER_EXTENSIONS | IMAGE_EXTENSIONS
".toml" in watched   # False
".yaml" in watched   # True   (DOC_EXTENSIONS, detect.py:32)
".json" in watched   # True   (CODE_EXTENSIONS, and extract_json exists)
```

`.ini` and `.cfg` are in the same position.

## Why this is more than a missing extractor

There is a narrow escape hatch — `manifest_ingest.is_package_manifest_path()` — but it matches five filenames:

```python
PACKAGE_MANIFEST_NAMES = {'apm.yaml', 'apm.yml', 'go.mod', 'pom.xml', 'pyproject.toml'}
```

So `pyproject.toml` yields a single package node and every other TOML file yields nothing. Two separable consequences:

**1. `Cargo.toml` is not recognized.** Rust's package manifest is missing from a set that already includes Python's, Go's and Maven's. A Rust repo gets no package node where an otherwise-identical Python repo does. That looks unintended, and is a one-line fix independent of the rest of this issue.

**2. Config/data TOML is structurally invisible.** In our case the affected files are the repository's actual data contracts:

| file | role | nodes |
|---|---|---:|
| `registry.toml` | repo → role/owner/dependency edges | 0 |
| `roster.toml` | team identity | 0 |
| `tiers.toml` | per-tier tool enablement | 0 |

Those three files are most of the reason the repository exists. Its graph contains the TypeScript that *reads* them (827 nodes) and the Markdown that *describes* them (646 nodes), but nothing they encode. `osv-scanner.toml`, `rust-toolchain.toml`, `netlify.toml` and `.cargo/config.toml` are in the same class.

## Request

Smallest first; any of these helps:

1. Add `Cargo.toml` to `PACKAGE_MANIFEST_NAMES`.
2. Put `.toml` (and `.ini`/`.cfg`) in `DOC_EXTENSIONS` so they become *reachable*, matching `.yaml`'s current status.
3. A deterministic structural extractor keyed on tables/keys, the way `extract_json` handles JSON — the generic form of what #2214 requests.

(2) alone would be a real improvement over the status quo even though it routes to the LLM, because today there is no configuration — API key, flag, or ignore file — that makes these files visible.

## Relationship to existing issues

Not a duplicate of #2214, which concerns *how* structured files are extracted once collected (LLM vs deterministic). This is one step earlier: TOML is never collected. Related to #2043, which addresses the same reachability question for YAML.

ousamabenyounes:
## What

Recognize Rust's `Cargo.toml` as a package manifest, alongside the
`pyproject.toml` / `go.mod` / `pom.xml` that `manifest_ingest` already handles.

Fix #2434 (scoped to the issue's Request 1 — "Add `Cargo.toml` to
`PACKAGE_MANIFEST_NAMES`", which the issue calls out as "a one-line fix
independent of the rest of this issue"). Requests 2 and 3 (routing
`.toml`/`.ini`/`.cfg` into `DOC_EXTENSIONS`, and a generic structural TOML
extractor) are deliberately left out — they are broader, separate concerns.

## Why

`PACKAGE_MANIFEST_NAMES` maps Python's, Go's, Maven's and apm's manifests to a
deterministic parser that emits one canonical `package` node per package (plus
`depends_on` edges). Rust's `Cargo.toml` was missing, so a Rust repo got **no**
package node where an otherwise-identical Python repo did — the gap the issue
documents.

## How

- Add `"cargo.toml": "cargo"` to `PACKAGE_MANIFEST_NAMES` **and** a matching
  `_parse_cargo` in `_PARSERS` (both must move together — a name without a
  parser would route the file to a `KeyError` path).
- `_parse_cargo` mirrors `_parse_pyproject`: name/version from `[package]`,
  runtime deps from `[dependencies]` **and** every `[target.<cfg>.dependencies]`
  table (platform-conditional deps are common in real crates). It reuses the
  shared `_coerce_deps` / `_pkg_id` helpers, so an inline-table dep
  (`tokio = { version = "1", features = [...] }`) keys by the dependency name.
- Defensive, matching the sibling parsers: a virtual workspace root (no
  `[package]`) emits nothing rather than a fabricated node, and a
  workspace-inherited version (`version.workspace = true`, which parses to a
  table) is dropped instead of crashing.
- `dev-dependencies` / `build-dependencies` are intentionally excluded to match
  `_parse_pyproject`'s runtime-only scope.
- Doc-sync: updated the module docstring and the `classify_file` comment, and
  added a `CHANGELOG.md` entry under `0.9.34 (unreleased)`.

Five new tests in `tests/test_manifest_ingest.py`.

**RED** — new tests against unmodified `v8` (before the fix), production code stashed:

```
graphify/manifest_ingest.py:60: KeyError: 'cargo.toml'
FAILED tests/test_manifest_ingest.py::test_cargo_classifies_as_code_manifest
FAILED tests/test_manifest_ingest.py::test_cargo_parses_name_version_and_deps
FAILED tests/test_manifest_ingest.py::test_cargo_virtual_workspace_manifest_emits_no_package
FAILED tests/test_manifest_ingest.py::test_cargo_target_conditional_deps_are_collected
FAILED tests/test_manifest_ingest.py::test_cargo_workspace_inherited_version_does_not_crash
5 failed, 8 deselected
```

**GREEN** — with the fix:

```
tests/test_manifest_ingest.py: 12 passed
```

## Full local CI (mirrors `.github/workflows/ci.yml`)

```
uv run --frozen ruff check graphify tests        -> All checks passed!
uv run --frozen python -m tools.skillgen --check -> check OK: 134 artifact(s) match
uv run --frozen pytest tests/ -q                 -> 4049 passed, 3 skipped
uv run --frozen graphify --help                  -> OK
```

Baseline on `v8` before the change was `4044 passed, 3 skipped`; the delta is
exactly the 5 new tests, no regressions. New production lines are fully covered
by the new tests.


safishamsi:
Shipped in v0.9.41 (`graphifyy==0.9.41` on PyPI) (#2434). Credited in the release notes. Thanks @ousamabenyounes!

Issue #1704, opened 2026-07-06
explain: fails to match labels containing punctuation (asymmetric normalization vs norm_label)


ryanhelms:
## Summary

`graphify explain "<label>"` fails to find a node whose stored label contains
punctuation that `\w+` doesn't match (e.g. `.`, `-`), even when the label is
typed exactly as stored and other commands (`path`, and presumably `query`)
resolve the identical string correctly. Root cause is an asymmetry between how
the query term is tokenized and how `norm_label` is stored on the node.

## Repro

Given a graph containing a file node with `label: "blockStream.ts"` /
`norm_label: "blockstream.ts"`:

```
$ graphify explain "blockStream.ts"
No node matching 'blockStream.ts' found.

$ graphify explain "blockStream.test.ts"
No node matching 'blockStream.test.ts' found.

$ graphify explain "BlockStream"        # a different node (a class), no punctuation
Node: BlockStream
  ID:        lib_blockstream_blockstream
  ...
```

But the exact same literal string resolves fine through `path`:

```
$ graphify path "blockStream.ts" "_stream.py"
Shortest path (3 hops):
  blockStream.ts <--imports_from-- blockStream.test.ts --shares_data_with--> test_stream.py --imports_from--> _stream.py
```

## Root cause

`graphify/serve.py::_find_node` (used by the `explain` subcommand) builds its
search term with:

```python
def _search_tokens(text: str) -> list[str]:
    return re.findall(r"\w+", _strip_diacritics(str(text)).lower())
```
```python
term = " ".join(_search_tokens(label))
```

For `label = "blockStream.ts"`, `\w+` splits on the `.`, giving tokens
`["blockstream", "ts"]`, so `term == "blockstream ts"` (space where the period
was).

Meanwhile the node's stored `norm_label` (written at graph-build time) is
`"blockstream.ts"` — lowercased, but punctuation left intact, not tokenized.

`_find_node`'s three-tier match (`term == norm_label`, `norm_label.startswith(term)`,
`term in norm_label`) then never succeeds, because `"blockstream ts"` (with a
space) is never equal to, a prefix of, or a substring of `"blockstream.ts"`
(with a period) — there's no space character in the stored label at all.

Confirmed directly against a `graph.json`:

```json
{"id": "lib_blockstream", "label": "blockStream.ts", "norm_label": "blockstream.ts"}
{"id": "lib_blockstream_test", "label": "blockStream.test.ts", "norm_label": "blockstream.test.ts"}
```

[...]

PR #2499, opened 2026-08-06
fix(serve): resolve node ids containing punctuation in _find_node (#2467)

### LINKED ISSUE #2467: explain/_find_node cannot resolve node IDs containing punctuation or non-ASCII text (concept:*, merge-graphs ::, Hangul)
## Summary

`_find_node` (and therefore `graphify explain`) cannot resolve **any node ID that contains punctuation**. In our graph this silently disabled lookup for **2,327 of 2,327** concept/overlay nodes (`concept:domain:*`, `concept:class:*`, `entity:*`, `commit:*`), while IDs without punctuation resolved 400/400.

The docstring states the intent clearly, so this looks like an oversight rather than a design decision:

> Return node IDs whose label **or ID** matches the search term (diacritic-insensitive).
> [...] Node-ID exact matches are grouped with label exact matches.

`find_node_ambiguity`'s user-facing hint also assumes ID lookup works:

> Retry with the repo-relative path or **the full node id**.

Version: `graphifyy 0.9.32`

## Reproduction

```python
import networkx as nx

G = nx.DiGraph()
G.add_node("concept:domain:gosu", label="Gosu")   # id contains ':'
G.add_node("plain_node_id", label="Plain")        # id has no punctuation

print("colon id  ->", _find_node(G, "concept:domain:gosu"))
print("plain id  ->", _find_node(G, "plain_node_id"))
print("by label  ->", _find_node(G, "Gosu"))
```

Actual:

```
colon id  -> []
plain id  -> ['plain_node_id']
by label  -> ['concept:domain:gosu']
```

Expected: the first line resolves to `['concept:domain:gosu']`.

Same through the CLI:

```
$ graphify explain "concept:domain:gosu"
No node matching 'concept:domain:gosu' found.     # node exists in the graph
```

## Root cause

In `graphify/serve.py`, `_find_node_tiers` builds two normalizations of the query:

```python
term       = " ".join(_search_tokens(label))          # \w+ tokens -> punctuation becomes a space
norm_query = _strip_diacritics(str(label)).lower().strip()   # punctuation preserved
```

The exact-match tier compares the node ID against `term` only:

```python
elif (
    term == norm_label or term == bare_label or term == label_tokens or term == nid_lower
    or norm_query == norm_label or norm_query == bare_label
):
```

For `concept:domain:gosu`:

| value | |
|---|---|
| `term` | `concept domain gosu` |
| `norm_query` | `concept:domain:gosu` |
| `nid_lower` | `concept:domain:gosu` |

`term == nid_lower` is always False once the ID contains punctuation. `norm_query` — which already holds the correct form and is even part of the trigram prefilter (`_trigram_candidates(G, [term, norm_query])`) — is never compared against `nid_lower`.

`norm_query` appears to have been introduced for punctuated *labels* (the comment references #1704); it just wasn't extended to node IDs.

## Suggested fix

One additional clause in the exact tier:

```python
elif (
    term == norm_label or term == bare_label or term == label_tokens or term == nid_lower
    or norm_query == norm_label or norm_query == bare_label or norm_query == nid_lower
):
```

## Why this is worse than a plain lookup failure

The natural workaround — calling `explain` with the node's label instead of its ID — resolves to the **wrong node silently**. Across our 35 concept nodes, only 4 resolved correctly; 31 landed on a same-named code node:

| query | resolves to | intended |
|---|---|---|
| `Repository` | `data_repository_feedrepository_ts` | `concept:class:Repository` |
| `API` | `api_apitypehelper_ts` | `concept:class:API` |
| `Gosu` | `entity:Gosu` | `concept:domain:gosu` |

So users get a plausible-looking wrong answer rather than an error, which is hard to notice in downstream reasoning.

ousamabenyounes:
## Why

| | |
| --- | --- |
| Issues | Fix #2467 |
| Branch | v8 |
| Bug fix? | yes |

`graphify explain` (and `_find_node`) cannot resolve **any** node ID that contains punctuation — e.g. `concept:domain:gosu`, `entity:*`, `commit:*`. The reporter measured 2,327 of 2,327 concept/overlay nodes silently unresolvable by id, while punctuation-free ids resolved. The docstring promises id lookup ("Return node IDs whose label **or ID** matches …"), and `find_node_ambiguity`'s own hint tells users to "retry with the full node id" — so this is an oversight, not a design choice.

## Root cause

In `graphify/serve.py` `_find_node_tiers`, the query is normalized two ways:

```python
term       = " ".join(_search_tokens(label))          # \w+ tokens -> punctuation becomes a space
norm_query = _strip_diacritics(str(label)).lower().strip()   # punctuation preserved
```

The exact-match tier compared the node id (`nid_lower`, which **keeps** punctuation) only against `term` (which **replaces** punctuation with spaces). For `concept:domain:gosu`, `term` is `"concept domain gosu"` — it can never equal `nid_lower` `"concept:domain:gosu"`. The punctuation-preserving `norm_query` was already compared against `norm_label`/`bare_label`, but not against `nid_lower`.

## What changed

Add the symmetric comparison `norm_query == nid_lower` to the exact-match tier — the exact counterpart of the `norm_query == norm_label` clause that was added for punctuated *labels* in #1704. One production line, plus a clarifying comment and a CHANGELOG entry.

## Why it's safe

- **Purely additive, zero false-positive risk.** A node newly enters the `exact` tier only when the whole diacritic-stripped, lowercased query byte-equals that node's id — the definition of an exact id match. It can only promote a currently-unresolvable id into `exact`; it cannot pull in a different node or reorder resolution for any existing query.
- **Correct tier.** The docstring states node-ID exact matches are grouped with label exact matches, and the existing `term == nid_lower` clause already lives in `exact`.
- **Siblings untouched.** `shortest_path` / `_score_nodes` use a separate scoring path and are not affected; `explain` and `find_node_ambiguity` route through `_find_node_tiers` and only benefit.
- Scope is the exact-id case reported in #2467; partial-prefix punctuated-id lookup (the `prefix` tier) is intentionally left as a separate concern to avoid speculative false-prefix matches.

New test `tests/test_serve.py::test_find_node_resolves_id_with_punctuation`.

RED — on unmodified `v8`, with only the new test applied:

```
tests/test_serve.py:239: in test_find_node_resolves_id_with_punctuation
    assert _find_node(G, "concept:domain:gosu") == ["concept:domain:gosu"]
E   AssertionError: assert [] == ['concept:domain:gosu']
FAILED tests/test_serve.py::test_find_node_resolves_id_with_punctuation
1 failed
```

GREEN — with the fix:

```
tests/test_serve.py::test_find_node_resolves_id_with_punctuation PASSED
1 passed
```

The test also guards the regressions (punctuation-free ids and label lookups still resolve).

Full suite: `4044 passed` on clean `v8` → `4045 passed, 3 skipped` on this branch (the one added test). `ruff check graphify tests` clean; `python -m tools.skillgen --check` OK; `graphify --help` install smoke OK.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ Full release notes with details on each version: [GitHub Releases](https://githu
 - Fix: `graphify merge-graphs` no longer drops hyperedges (#2484, thanks @sortakool, and @oleksii-tumanov for the approach in #1691). Hyperedge member ids and ids are now relabeled with the per-repo prefix, both inputs' hyperedges are unioned instead of one clobbering the other, and they are written to both the top-level and nested slots.
 - Fix: `build_from_json` now reads hyperedges from both the top-level and nested `graph` slots, so label and re-cluster runs no longer silently empty a graph's hyperedge set (#2485, thanks @sortakool); a full validation wipeout is now reported loudly.
 - Fix: the skill flow now passes the curated community labels to `to_json`, so `graph.json` ships with `community_name` on nodes instead of dropping it (#2490, thanks @PapiScholz).
+- Fix: `graphify explain` (and `_find_node`) can now resolve a node by an exactly-typed id that contains punctuation, e.g. `concept:domain:x` (#2467, thanks @sean-soomgo). `term` tokenizes punctuation into spaces, so it never matched the punctuation-preserving node id; the punctuation-preserving `norm_query` is now also compared against the id, mirroring the `norm_query == norm_label` label path from #1704. Id lookups without punctuation and label lookups are unchanged.
 
 ## 0.9.33 (2026-08-05)
 
--- a/graphify/serve.py
+++ b/graphify/serve.py
@@ -1143,7 +1143,9 @@ def _find_node_tiers(
     # stored `norm_label` keeps punctuation ("blockstream.ts"). Matching only via
     # `term`/`label_tokens` works when the node label tokenizes the same way, but is
     # fragile if `label` and `norm_label` diverge. `norm_query` matches `norm_label`
-    # symmetrically so an exactly-typed punctuated label always resolves (#1704).
+    # symmetrically so an exactly-typed punctuated label always resolves (#1704),
+    # and against `nid_lower` so an exactly-typed node ID whose punctuation `term`
+    # tokenizes away (e.g. `concept:domain:x`) still resolves by id (#2467).
     norm_query = _strip_diacritics(str(label)).lower().strip()
     source_exact: list[str] = []
     exact: list[str] = []
@@ -1166,7 +1168,7 @@ def _find_node_tiers(
             source_exact.append(nid)
         elif (
             term == norm_label or term == bare_label or term == label_tokens or term == nid_lower
-            or norm_query == norm_label or norm_query == bare_label
+            or norm_query == norm_label or norm_query == bare_label or norm_query == nid_lower
         ):
             exact.append(nid)
         elif (
--- a/tests/test_serve.py
+++ b/tests/test_serve.py
@@ -225,6 +225,23 @@ def test_find_node_resolves_when_label_and_norm_label_diverge():
     assert _find_node(G, "blockStream.ts") == ["n1"]
 
 
+def test_find_node_resolves_id_with_punctuation():
+    # #2467: an exactly-typed node ID containing punctuation (e.g. a
+    # `concept:domain:x` overlay id) must resolve. `term` tokenizes the ':'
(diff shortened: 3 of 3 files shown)


safishamsi:
Superseded — this was fixed in v0.9.40 (commit 1fdd11f, #2467/#2583). Your diagnosis and the `norm_query == nid_lower` approach were exactly right; the shipped fix is the same clause generalized to `norm_query == nid_norm` so non-ASCII ids diacritic-fold too, plus the trigram prefilter now indexes the folded id. Thanks for the clean writeup — closing as already shipped.

Issue #1556, opened 2026-06-30
iOS Objective-C/Swift mixed-language: many expected edges missing (self-imports, .h↔.m linkage, cross-language calls, symbol fragmentation)


JabberYQ:
## Summary

Running graphify on a small **iOS Objective-C / Swift mixed-language ("混编") demo project** produces a graph where a large share of the expected edges are missing. The graph passes the built-in health check (0 dangling edges), but cross-file and cross-language relationships are largely absent or broken.

I've split the findings into two buckets:

- **P0 — extraction bugs**: edges that *should* be expressible with the existing 7 relation types (`calls / contains / implements / imports / inherits / method / references`) but are not extracted.
- **P1 — modeling gaps**: relationships that have no corresponding relation type today (would need a new edge type + semantic analysis).

## Environment

- graphify / graphifyy **0.9.2**
- Python 3.10.20, macOS (darwin)
- Corpus: 21 ObjC/Swift source files (`.h` / `.m` / `.swift`), code-only (AST path, no semantic/LLM step)
- Resulting graph: **137 nodes, 170 edges**, undirected. Health check: OK (0 dangling/missing/collapsed).

The corpus is a deliberately-built mixed-language fixture: every cross-language construct (bridging header, `@objc`, `@objc(rename)`, `NS_SWIFT_NAME`, Swift `extension` on an ObjC class, KVO on `@objc dynamic`) is present on purpose.

---

### 1. Bridging header: all 5 `#import` edges dropped → node fully isolated

`Demo-Bridging-Header.h` is a real on-disk file:

```objc
#import "Product.h"
#import "ProductDetailViewController.h"
#import "AlertManager.h"
#import "BaseModel.h"
#import "Refreshable.h"
```

graphify extracts `imports` for every other header/impl file (21 `imports` edges total, e.g. `AlertManager.m → Product.h`, `AppDelegate.m → ProductListViewController.h`), but **none** of the bridging header's 5 imports are extracted. The node ends up with **0 edges**, isolated in its own community.

`imports` is an existing relation type and works elsewhere — so this is a genuine extraction bug, not a modeling gap. The bridging header is the physical hub through which Swift "sees" all ObjC types; losing it removes the entire Swift→ObjC visibility backbone.

### 2. `.m → own .h` self-import dropped (6/6)

Every `.m` imports its own header (line 6 in each file):

```
AlertManager.m:6                 #import "AlertManager.h"
AppDelegate.m:6                  #import "AppDelegate.h"
[...]

PR #2500, opened 2026-08-06
fix(objc): don't treat @protocol declarations as receiver types (#1556)

xiongjianxu:
## Problem

Objective-C keeps protocol names and class names in **separate namespaces** — Foundation itself ships both `@protocol NSObject` and `@interface NSObject`. A same-named pair is ordinary ObjC, not a mistake.

The ObjC extractor labels a protocol declaration `<Name>` (`graphify/extractors/objc.py:264`), and `_resolve_objc_member_calls`'s index key strips non-alphanumerics:

```python
def _key(label: str) -> str:
    return re.sub(r"[^a-zA-Z0-9]+", "", str(label)).lower()
```

So `<Locking>` and `Locking` collapse to the same key `locking`, and a protocol becomes a receiver-typing candidate. Both outcomes are wrong:

**1. Protocol only in the corpus — a WRONG edge at confidence 1.0.**

```objc
// Reload.h
@protocol Reload <NSObject>
- (void)reload;
@end

// Use.m
@implementation Use
- (void)go { [Reload reload]; }
@end
```

No class named `Reload` exists, so the receiver is untypable and the god-node guard should bail. Instead:

```
-go -> -reload   calls  EXTRACTED  (target: reload_reload_reload, Reload.h)
```

The edge points at the protocol's method *declaration*. A protocol is a contract; it is never a message receiver.

**2. Protocol AND class present — a real edge is destroyed.**

```objc
// Locking.h
@protocol Locking <NSObject>
- (void)lock;
@end

// LockingImpl.h / .m
@interface Locking : NSObject
+ (void)lock;
@end

// Worker.m
- (void)run { [Locking lock]; }
```

`type_def_nids["locking"] == ['locking_locking' (<Locking>), 'lockingimpl_locking' (Locking)]` → `len(type_defs) != 1` → the single-definition guard bails, and `-run -> +lock` is **never emitted** (measured: 0 call edges).

## Fix

Exclude protocol declarations from the receiver-type index in `_resolve_objc_member_calls` only. The `<Name>` label shape is unique to ObjC protocols — `grep -rn 'f"<{' graphify/` returns exactly one hit, `extractors/objc.py:264` — so the check is unambiguous, and keeping it local to this resolver leaves `_is_type_like_definition` (7 call sites, 6 other languages) untouched.

Protocols remain valid `implements` targets and valid `_rewire_unique_stub_nodes` destinations; only receiver typing ignores them. Verified: `Widget implements <Reload>` still resolves to the real protocol node.

## Tests

New `tests/test_objc_member_calls.py`, following the `test_csharp_member_calls.py` / `test_java_member_calls.py` convention — every case pairs the positive assertion with a decoy that must get no edge:

| test | asserts |
|---|---|
| `test_objc_protocol_only_receiver_emits_no_call_edge` | protocol-only receiver → **zero** call edges (decoy: the protocol's own `-reload`) |
| `test_objc_class_resolves_past_a_same_named_protocol` | `-run -> +lock EXTRACTED` present, `-run -> -lock` absent |
| `test_objc_protocol_stays_a_valid_implements_target` | `Widget implements <Reload>` unaffected |

Both new behavioral tests **fail on `v8` without the fix** (verified by stashing the `extract.py` change: 2 failed, 1 passed) and pass with it.

## Verification

```
  4047 passed, 3 skipped in 161.52s        (v8 baseline: 4044 passed, 3 skipped)

uv run --frozen python -m tools.skillgen --check --audit-coverage --schema-singleton \
                                          --monolith-roundtrip --always-on-roundtrip
  all 5 OK

uv run --frozen ruff check --config pyproject.toml graphify/extract.py tests/test_objc_member_calls.py
  All checks passed!
```

`uv.lock` untouched. Base branch: `v8`.

## Notes

No open issue covers this; issue creation is restricted for external contributors, so the reproduction is inline above. It is a defect in the #1556 ObjC member-call pass, hence the reference.

This is the first of a few small, independent ObjC precision fixes I have measured on 0.9.34; the others (category/class-extension interfaces minting a duplicate class node, and `@property`/ivar receivers never being typed) are separate PRs so each can be reviewed on its own.


xiongjianxu:
Filed the reproduction as #2589 so it is tracked independently of this patch.

safishamsi:
Shipped in v0.9.39. Excluding `<...>`-labeled protocol declarations from the receiver-type index is what landed for #2589. Credited in the release notes. Thanks @xiongjianxu!

PR #2502, opened 2026-08-06
fix(php): preserve use-statement facts and stop group-form use function/const from claiming class names

filipechagas:
Two coupled PHP `use`-statement fixes, backported from our fork where they landed with full red-first evidence ([lawnstarter#29](https://github.com/lawnstarter/graphify/pull/29), [lawnstarter#30](https://github.com/lawnstarter/graphify/pull/30)). This PR is independent of #2492 (it touches only pre-existing code); a heads-up was posted there for context.

## Commit 1 — capture `use` FQN/alias/kind metadata on imports edges (metadata-only)

`_import_php` threw away everything but the trailing short name (`raw.split("\\")[-1]`), unlike `_import_csharp`, which preserves the written FQN and alias. The already-correct `use`-parser inside `_resolve_php_type_references` (group use, aliases, `function`/`const` rejection) is extracted into shared helpers consumed by both paths, and PHP `imports` edges now carry `use_kind` / `alias` / `target_fqn` metadata.

No behavior change: metadata-stripped `extract()` output over a corpus covering plain / aliased / group / aliased-group / `function` / `const` / group-function / group-const / leading-backslash `use`, trait `use`, inheritance, and interfaces is sha256-identical before and after.

## Commit 2 — group-form `use function` / `use const` no longer claim class names (strictly subtractive)

The grammar puts the `function`/`const` keyword on the *clause* for `use function A\f;` but on the *declaration* for `use function A\{f, g};` (verified on tree-sitter-php 0.24.1). `_resolve_php_type_references` only ever honored the clause-level keyword, so group-form function/const names wrongly entered the class-name map — e.g. `use function Vendor\Sdk\{render};` plus `class I extends Render` mis-repoints the inheritance. Now the declaration-level keyword is always honored; one code path, no compatibility flag.

## Test evidence

- Baseline established on this tree (`v8` @ `07b9143`): 3976 passed / 36 skipped.
- Commit 1: red 7/8 (the passing one is a targets-unchanged guard, green by construction) → full suite 3984/36.
- Commit 2: red 3/4 (over-subtraction guard green by design: keyword-less `use A\{B};` still claims the class name, decoy gets no edge) → full suite 3988/36.
- 12 new tests, all through the public `extract()` seam.


filipechagas:
We close this PR in favor of one consolidated PR. Seven open PRs became hard to review and hard to keep current with `v8`.

- Consolidated defect report: #2615 (defect family 2)
- Consolidated PR: #2617 — it contains both fixes this PR carried, plus the `target_fqn` import resolution that was blocked on this PR.

PR #2503, opened 2026-08-06
fix(extract): isolate member-call resolvers and import edges from cross-language interference

filipechagas:
Cross-language isolation fixes for the member-call resolvers and import-edge resolution, in five commits spanning two related defect families. Both families share one root shape — language-blind indexes and ids letting one language's facts bind another language's nodes. The first commit is a backport of a fix that landed on our fork with full red-first evidence ([lawnstarter#28](https://github.com/lawnstarter/graphify/pull/28)); the rest were found by measuring outward from it on this tree. Happy to split into two PRs (commits 1–3 / 4–5) if you prefer — the later commits extend the first commit's test file, so they're presented together here.

## Family 1 — resolvers consuming other languages' facts

1. **`0e6dd75` — raw-call ownership.** The Swift, Python, and TypeScript member-call resolvers consume every untagged raw call, including each other's: a TS `Lead.search({})` with a matching class only in a `.py` file mints `runner.ts:.go() → svc.py:.search()` at EXTRACTED. Fix: a positive source-file suffix filter (`_raw_call_is_owned`), with per-resolver suffix tuples shared with the `LanguageResolver` registrations so filter and registration cannot drift. Also scopes the Java and C# `type_def_nids` fallbacks (checked against the current post-`e300587` C# code).
2. **`7ddf65c` — the remaining receiver-type indexes.** C++, ObjC, Swift, and TypeScript `type_def_nids` were still corpus-wide; each leaked (measured per-language, both directions: false cross-language edge, and a foreign namesake suppressing the true edge). `.h` remains in both the C++ and ObjC sets by necessity — they're now isolated from every other language but not from each other.
3. **`0f417fc` — the Python indexes.** `class_def_nids` and the module arm, same defect, different index shape (built from method edges rather than source-backed nodes).

## Family 2 — import edges dying to a same-stem foreign sibling

4. **`1764280` — Python.** A `lead.ts` next to `lead.py` makes id disambiguation salt both file nodes while the import edge still targets the bare alias — the edge dangles, and everything downstream (including the module-arm resolution) silently loses it. Fix: stamp the `target_file` hint the disambiguator already reads (the `#1814` mechanism); a Python import can only mean a Python file, so the hint is unambiguous.
5. **`1c77490` — PowerShell, Zig, Rust, Pascal, Elixir, and bash.** The same collision kills stem-named import edges in five more languages (a stray `lead.md` next to `lead.rs` is enough). Fixed centrally via a language table in the hint pass — it only picks among salted variants of an id the edge already named, so a collision-free corpus is bit-identical. Bash needed a different fix: `symbol_resolution.py` re-derives ids by path formula *after* disambiguation, so it now consults the post-disambiguation ids (`_file_nids_by_path`) and its duplicate dangling edge dedupes away. TS/JS, C++/ObjC (`header_remaps`), Julia/Fortran/Verilog, and Dart were audited and are not vulnerable; a Go dangle that reproduces *without* a collision was found and deliberately left unasserted — it looks like a separate resolution gap and deserves its own issue.

## Test evidence

- Baseline established on this tree (`v8` @ `07b9143`): 3976 passed / 36 skipped (one known-flaky labeling test passed in isolation).
- Every commit red-first through the public `extract()` seam; every positive test carries a same-named decoy asserted to get no edge. 30 new tests total in `tests/test_mixed_corpus_member_calls.py`.
- Full suite at head: **4006 passed / 36 skipped / 0 failed**, independently reproduced by a second runner.


filipechagas:
We close this PR in favor of one consolidated PR. Seven open PRs became hard to review and hard to keep current with `v8`.

- Consolidated defect report: #2615 (defect family 3)
- Consolidated PR: #2617 — it contains both fix families this PR carried (resolver ownership and the same-stem import dangle).

Issue #2472, opened 2026-08-05
C#: an untypeable declaration poisons an already-typed receiver name within the same method


JensD-git:
**Version:** present in 0.9.32 and 0.9.33; last good version 0.9.31
**Language:** C#

Bisected on the reproducer below: 0.9.31 resolves 3/3, **0.9.32 resolves 0/3**, 0.9.33
resolves 0/3. So it arrived with 0.9.32 — possibly as a side effect of the
pattern-based-declaration work released in that version (#2346).

## Summary

When one method contains two declarations of the same identifier — one that graphify
can type and one that it cannot — 0.9.33 drops the typing for **all** uses of that
name, including the uses that belong to the typed declaration. 0.9.31 resolved them.

This is the method-scoped analogue of #2299 (which was the file-scoped version of the
same problem, fixed in 0.9.31).

## Minimal reproducer

`Probe2.cs` (single project, `net8.0`):

```csharp
namespace Repro2
{
    public class Target2
    {
        public void Alpha() { }
        public void Beta()  { }
        public void Gamma() { }
    }

    public static class Factory2
    {
        public static void Make(out Target2 created) { created = new Target2(); }
    }

    public class Caller2
    {
        public void Outer()
        {
            // (1) parameter of a *static local function* -- explicitly typed.
            //     C# allows a static local function to shadow enclosing locals.
            static void LocalFunction(Target2 shared)
            {
                shared.Alpha();
                shared.Beta();
            }

            // (2) same identifier, declared via `out var` -- type only available
            //     from the callee signature (out Target2 created)
            Factory2.Make(out var shared);
            shared.Gamma();

            LocalFunction(shared);
        }
    }
}
```

## Expected vs. actual

`calls` edges into `Target2`:

| version | Alpha | Beta | Gamma | total |
|---|---|---|---|---|
| expected | yes | yes | yes | 3 |
| 0.9.31 | yes | yes | yes | 3 |
| **0.9.32** | **no** | **no** | **no** | **0** |
| **0.9.33** | **no** | **no** | **no** | **0** |

Both `graph.raw.json` and the final graph show the same numbers.

The two declarations are in **different scopes**: `shared` at (1) is the local
function's parameter, `shared` at (2) is a local of the enclosing body. They are
distinct variables and C# permits this precisely because the local function is
`static`. Resolving `Alpha`/`Beta` requires nothing but the parameter list a few
lines above the call.

## Control case

A file with the same constructs but **without** name reuse resolves identically in
[...]

PR #2505, opened 2026-08-06
fix(php): member-call follow-ups — FCC relation, union refusal, qualified receiver types, use-claim refusal

filipechagas:
Four coupled PHP member-call fixes, backported from our fork where they landed with full test evidence ([lawnstarter#27](https://github.com/lawnstarter/graphify/pull/27), [lawnstarter#25](https://github.com/lawnstarter/graphify/pull/25), [lawnstarter#31](https://github.com/lawnstarter/graphify/pull/31), [lawnstarter#32](https://github.com/lawnstarter/graphify/pull/32)).

**Stacked on #2492 and #2502** — this branch is their merge plus five commits; only the last five commits (`804dc81` and up) are new here. Once those two PRs merge into `v8`, this diff reduces to exactly those five.

## Commit-by-commit

1. **`indirect_call` for PHP 8.1 first-class callables** (`lawnstarter/graphify#15`) — `$obj->method(...)` creates a `Closure` without invoking it, so the edge moves from `calls` to the relation this repo already uses for a callback passed by name. Target resolution, refusals and confidence are unchanged; detection keys on the argument list being exactly `variadic_placeholder` on the pinned grammar (tree-sitter-php 0.24.1). A real call outranks a reference for the same (caller, method) pair.
2. **Union/intersection receivers stop minting the same-file bare-name edge** (`lawnstarter/graphify#9`) — the legacy in-file matcher couldn't tell "annotation refused" from "no annotation", so `private Alpha|Beta $svc;` bound whichever same-named method the file's label index saw last, at EXTRACTED. The receiver table now records the refusal, and a refused multi-class annotation defers to the receiver-typed resolver. Deletion scope: only union/intersection-typed receivers' same-file edges.
3. **PHP 8.2 DNF types recognized** (`lawnstarter/graphify#9`) — `private (A&B)|C $x;` parses as its own node the scanners didn't name; it now refuses like a union and emits `references` to A, B and C like one.
4. **Written qualified receiver types kept** (`lawnstarter/graphify#20` on the fork) — the extractor stops flattening `\Vendor\Sdk\Client` to `Client` for property/param/promoted-param/local type positions; the receiver table's PHP values become a `(short, qualified)` pair.
5. **`PhpNameResolver`: a claimed name refuses instead of guessing** (`lawnstarter/graphify#21`, closes the fork's measured false-edge `lawnstarter/graphify#16`) — the PHP twin of `CsharpNameResolver`, consulted in front of the corpus-wide short-name index in the exact shape of the C# call site. A file that writes `use Vendor\Sdk\Client;` has claimed the name; when the claim lands on no in-corpus class the resolver refuses rather than binding the lone same-short-named stranger. Strictly subtractive; verified on the fork by a differential extract (3 edges deleted — exactly the #16 class — 0 added, 0 re-pointed).

## Verification

Full suite on this branch: **4122 passed, 36 skipped** (the merged #2492+#2502 base alone: 4079 passed, so every added test rides with its fix). Conflict resolutions against the #2472 scoped C# receiver tables are confined to the shared `receiver_types` annotation/merge comments; body ids are language-disjoint, so each language's table only meets its own reader.

A follow-up PR stacks the recall counterpart (declared-FQN positive binding + its incremental-parity marker) on top of this one.


filipechagas:
Consolidating a reply to the four automated coupling threads on this PR rather than answering each inline. None of them is a regression in the sense the label implies, and I'm not proposing a change for any — reasoning and numbers below, all from an AST scan against base `07b9143d` (the tool's counts and mine differ slightly on `_php_method_receiver_types` — it says 7, I count 9 — so these are mine).

**Two of the four functions do not exist at base.** Both are introduced by `43cd7a2`, the member-call resolution commit this branch stacks on:

| symbol | at base `07b9143d` | at this head |
|---|---|---|
| `_php_method_receiver_types()` `engine.py:809` | not found | 809-963, 155 lines |
| `_resolve_php_member_calls()` `extract.py:3106` | not found | 3106-3290, 185 lines |

There is no prior fan-out for either, so there is no delta. Whether a 155- or 185-line helper should itself be decomposed is a fair question, but it is a design question about new code, not a coupling regression.

**`walk_calls()` — `engine.py:4616`, "fans out to 14 callees" — is the only genuine delta, and it is +1.** Base: `engine.py:4201`, span 4201-4777, **577 lines**, 13 distinct project callees. Head: 4616-5349, 734 lines, 14. Our diff inside that span is +163 / −6, and the callee set difference is **exactly one symbol** (`_php_name_text`); everything else was already there. It is a nested function inside `_extract_generic` (`engine.py:2691`), the multi-language call-walking recursion — already very large before this branch touched it. The 13→14 crossing is the threshold firing, not a change in what `walk_calls` structurally depends on.

I did look for a cheap behaviour-preserving improvement here, and there is one — `engine.py:4808-4847`, the block deriving `member_receiver` / `php_inline_new_type` / `php_inline_new_qualified`, is a pure function of `(node, source)` and would extract cleanly as `_php_member_receiver()`, dropping the span to ~695 lines and making that tree-sitter node-shape logic unit-testable. I've tracked it on our fork rather than doing it here, for one honest reason: it **would not clear the flag**. `_php_name_text` is called exactly once in the whole span (line 4839), inside that very block, so the extraction swaps one callee for another and the count stays 14. If you'd like it anyway on readability grounds, say so and I'll fold it in.

**`_full_then_incremental()` — `tests/test_php_member_calls.py:1586`, "9 callers".** New helper in a new file: the full-extract-then-incremental-rebuild parity harness. Tests have to share it or the parity claim is empty — a full-build-only assertion is green on the broken code. Confined to `tests/`; no production symbol's coupling changed.

filipechagas:
We close this PR in favor of one consolidated PR. This PR was stacked on #2492 and #2502; the stack made review and rebases hard.

- Consolidated defect report: #2615 (defect family 4)
- Consolidated PR: #2617 — it contains all five commits this PR added, with their dependencies, in one review unit.

PR #2506, opened 2026-08-06
feat(php): bind claimed receiver types to their declared FQNs, with incremental-rebuild parity

filipechagas:
The recall counterpart to the refusal in #2505, backported from our fork ([lawnstarter#35](https://github.com/lawnstarter/graphify/pull/35), spec [lawnstarter#18](https://github.com/lawnstarter/graphify/issues/18) tickets 4/5 and 5/5).

**Stacked on #2505** (and transitively on #2492 and #2502) — only the last two commits (`6b90f7f`, `3930ce8`) are new here. The two ship together deliberately: without the marker, every edge the binding adds would exist on full builds only.

## Commit 1 — declared-FQN positive binding

`PhpNameResolver` gains a casefolded `declared FQN -> definition node` index over the same node universe as the short-name census, sourced from the per-file declared-namespace payload the inline-`new` corroboration already reads. Once a calling file's claim (via `use`, `use ... as`, group `use`, or a qualified annotation) resolves to an FQN, a unique index hit binds **before** the single-candidate check. Two things now resolve that used to refuse: `use App\Alpha\X;` selects `App\Alpha\X` among several same-short-named classes, and a renaming alias `use App\Alpha\X as Y;` follows the alias to a class the written short name never censuses. An FQN declared identically by two files poisons its entry rather than guessing; on a miss everything falls through to the refusal rules untouched. Additive-only by construction — the mirror risk profile of #2505's subtractive-only commit 5.

## Commit 2 — persisted `_php_class_fqns` marker

The declared-FQN payload aligns 1:1 with files dispatched *this* run, so without persistence the new edges silently vanish on the first `graphify update` when the defining file is unchanged — the same failure mode the interface refusal closed with `_php_non_class_types`. Each PHP file node persists the map into `graph.json` on the established `_callable` marker channel; `watch.py` and the CLI hand it back on resolution-context nodes; `extract()` re-injects it as synthetic per-file payload entries. A pre-marker `graph.json` fails closed: the binding is simply absent — never redirected to a namesake — until the defining file is re-extracted.

## Verification

Full suite on this branch: **4139 passed, 36 skipped**. On the fork, a differential extract against the pre-change baseline over a 25-file shape corpus showed **6 edges added, 0 deleted, 0 re-pointed, 0 confidence changes** — all six target exactly `App\Alpha\X::send` at INFERRED 0.8 (five `calls`, one `indirect_call` via the first-class-callable form). End-to-end through the real CLI channel: `graphify extract` → 9 call edges, 23 file nodes carrying the marker → touch only the callers → `graphify update` reproduces all 9 with the defining files undispatched. Every positive scenario also has a mirrored `_full_then_incremental` test plus pre-marker variants pinning the fail-closed behavior.


filipechagas:
Same consolidated answer as on #2505, since this branch stacks on it and inherits four of its five coupling threads. Numbers from an AST scan against base `07b9143d`; where the tool's count and mine differ I've used mine.

**`_php_method_receiver_types()` (`engine.py:809`, 809-963) and `_resolve_php_member_calls()` (`extract.py:3137`, 3137-3322) do not exist at base** — both are introduced by `43cd7a2` down the stack. No prior fan-out, so no delta.

**`walk_calls()` (`engine.py:4616`) is a genuine +1**: base `engine.py:4201`, 577 lines, 13 distinct project callees → head 734 lines, 14. The set difference is exactly one symbol (`_php_name_text`). Already the largest function in the file before this stack touched it. There is a clean extraction available at `engine.py:4808-4847` (`_php_member_receiver()`), which I've tracked on our fork rather than doing here because it swaps one callee for another and would not clear the flag — happy to fold it in if you want it on readability grounds.

**The two `_full_then_incremental()` threads** — `tests/test_php_alias_binding.py:264` (7 callers) and `tests/test_php_member_calls.py:1588` (9 callers) — are both new helpers in new files. Each is the full-extract-then-incremental-rebuild parity harness for its file, and the callers are the point: this PR's whole second commit exists because the declared-FQN binding would otherwise be a full-build-only feature, so every test here has to assert both build paths **together**. A full-build-only assertion is green on the broken code. Both are confined to `tests/`; no production symbol's coupling changed.

Nothing here I'd propose acting on, but tell me if the gate is blocking and I'll revisit.

filipechagas:
We close this PR in favor of one consolidated PR. This PR sat at the top of the #2492 → #2502 → #2505 stack; the stack made review and rebases hard.

- Consolidated defect report: #2615 (defect family 5)
- Consolidated PR: #2617 — it contains both commits this PR added (declared-FQN binding and the `_php_class_fqns` incremental-parity marker), with their dependencies, in one review unit.

PR #2508, opened 2026-08-06
fix(callflow): load graph directed — caller/callee columns are swapped on 43% of edges

Tomaskobel:
## Problem

`export callflow-html` loads `graph.json` through NetworkX **undirected**, so the Caller and Callee columns of the call table are populated from arbitrary edge orientation.

`graph.json` is written with `"directed": false` for backward compatibility (`build.py:658`). `_node_link_payload` passes that straight to `node_link_graph`, which therefore returns a plain `Graph` — and `Graph.edges()` does not promise to give back the orientation that was stored.

Measured on a 4,990-node / 10,135-edge graph of a real project:

| | |
|---|---|
| edges whose direction survived the round-trip | **5,785 / 10,132 (57%)** |
| edges returned **reversed** | **4,347 (43%)** |
| edges dropped entirely (parallel edges collapsed by simple `Graph`) | **3** |

The user-visible symptom is that a function's caller is printed in its **Callee** column and vice versa, on a ~43% subset that looks random. Concrete case from that graph — `scrape_domain() --calls--> matches_keyword()` in the source, and in `graph.json`:

```
 | matches_keyword() | Function | External entry / no inbound edge | scrape_domain() |
                                  ^ wrong: it has a caller             ^ wrong column: this IS the caller
```

Every other loader in the codebase already guards against exactly this — `affected.py:263`, `serve.py:42`, `cli.py:1229`, all referencing #1174. `callflow_html.py` looks like the one that was missed.

## Second, related defect

`"External entry / no inbound edge"` is a claim about the **whole graph**, but `callers`/`callees` were built from `section_edges` — edges with *both* endpoints inside the current section. Any node called from another section is therefore mislabelled an entry point.

Section coverage makes this the common case rather than a corner case: only `max_sections` communities are rendered (15 by default), so most callers are in no rendered section at all and their edges are never consulted.

## Fix

Two commits, each independently reviewable and testable:

1. **`load the graph directed`** — pass `{"directed": True, "multigraph": True}` into `load_node_link_graph`, mirroring the existing idiom in `affected.py` / `serve.py` / `cli.py`.
2. **`compute the Caller column from the whole graph`** — thread the full node/edge lists into `generate_call_table_rows`. Both new parameters default to `None`, so behaviour is unchanged for any other caller.

## Result

Rows claiming `External entry / no inbound edge` where the node demonstrably *does* have inbound call edges in `graph.json`:

| | claims | false |
|---|---|---|
| before | 327 | 45 |
| + directed load | 280 | 10 |
| + whole-graph caller column | 263 | **0** |

Direction preserved goes 57% → **100%**, and all 10,135 edges load instead of 10,132.

## Notes

- `format_node_refs` already caps each cell at 3 entries with `+N more`, so passing wider caller sets does not widen the table.
- Two regression tests added. Both **fail on `v8` as-is** and pass with the fix.
- Full suite: 3,795 passing, and the set of failing tests is byte-identical before and after this branch (27 pre-existing failures in `test_terraform.py`, `test_skillgen.py`, `test_cli_export.py`; `test_labeling.py::test_label_communities_batches_when_over_batch_size` appears to be flaky and fails intermittently on unpatched `v8` too).

Happy to split this into two PRs if you'd prefer the directed-load fix on its own — it is the root cause and the larger of the two.

### DIFF
--- a/graphify/callflow_html.py
+++ b/graphify/callflow_html.py
@@ -233,7 +233,14 @@ def _node_link_payload(data: dict) -> tuple[list, list] | None:
         # though the shape check above accepts "edges" (#2212).
         from graphify.paths import load_node_link_graph
 
-        graph = load_node_link_graph(data)
+        # Force directed/multigraph so the stored caller->callee direction and
+        # parallel edges survive the round-trip; mirrors affected.py:263,
+        # serve.py:42 and cli.py:1229 (#1174). graph.json is written with
+        # "directed": false for backward compatibility (build.py:658), so
+        # without this networkx returns an undirected Graph and edge
+        # orientation becomes arbitrary -- which silently swaps the Caller and
+        # Callee columns of the call table and drops parallel edges.
+        graph = load_node_link_graph({**data, "directed": True, "multigraph": True})
     except Exception:
         return None
 
--- a/graphify/callflow_html.py
+++ b/graphify/callflow_html.py
@@ -1214,16 +1214,35 @@ def format_node_refs(node_ids: set, node_by_id: dict, lang: str, empty_text: str
     return "<br>".join(parts)
 
 
-def generate_call_table_rows(nodes: list, section_edges: list, lang: str) -> str:
-    """Generate call table row scaffolding for a section's nodes."""
+def generate_call_table_rows(
+    nodes: list,
+    section_edges: list,
+    lang: str,
+    all_edges: list | None = None,
+    all_nodes: list | None = None,
+) -> str:
+    """Generate call table row scaffolding for a section's nodes.
+
+    The Caller/Callee columns make a claim about the whole graph ("External
+    entry / no inbound edge"), so they must be computed from the whole graph.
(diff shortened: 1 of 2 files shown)


Tomaskobel:
Landed in 9f25a3a (0.9.35) — thanks for picking it up, and for the credit. Closing this as superseded.

For anyone finding this later: both halves shipped — the forced directed/multigraph load in `_node_link_payload`, and `generate_call_table_rows` computing the Caller column from the whole graph. Upstream additionally recovers direction from the `_src`/`_tgt` markers, which the version here did not do.

I have two further callflow renderer fixes open, both now rebased on 0.9.35: #2510 (`humanize_label` compacts at 28 chars though the render width is 42, so labels that fit are shortened and `non_ai_` renders as `ai`) and #2518 (diagram truncation counts are computed but emitted only in an unrendered Mermaid comment). Plus #2519 as an issue, since that one is a design call rather than a bug.

Issue #2497, opened 2026-08-06
build_merge: the #479 shrink guard reads the POST-prune node count, so replace-on-re-extract can destroy any number of nodes silently


sortakool:
### Summary

`build_merge`'s `#479` shrink guard compares against the **post-prune** node
count, so it is structurally unable to see nodes lost to replace-on-re-extract.
A merge can destroy an arbitrary number of existing nodes and the guard stays
silent — including when the graph shrinks overall.

Observed on **0.9.34** (current PyPI release, and our pin).

### The mechanism

`graphify/build.py`, `build_merge`:

```python
1478  existing_nodes, existing_edges, existing_hyperedges, existing_directed = _loaded
...
1531  if new_sources:
1532      def _kept(item: dict) -> bool: ...
1536      existing_nodes = [n for n in existing_nodes if _kept(n)]   # <-- REBOUND, post-prune
1537      existing_edges  = [e for e in existing_edges  if _kept(e)]
...
1648  if graph_path.exists() and not dedup and not prune_sources:
1650      existing_n = len(existing_nodes)                           # <-- reads the PRUNED list
1651      new_n = G.number_of_nodes()
1652      if new_n < existing_n:
1653          raise ValueError("build_merge would shrink graph from ...")
```

Line 1536 rebinds `existing_nodes` to the list with the re-extracted files'
nodes already removed. By line 1650 the very quantity the guard exists to
notice has been subtracted from the left-hand side, so `existing_n` under-counts
by exactly the number of nodes the merge replaced.

### Why it matters more than a one-tier re-extract

The replacement key is `source_file`. Two *unrelated* inputs that happen to
produce the same `source_file` string therefore replace each other — and the
guard cannot report it.

Concretely: chunk A contributes 10 nodes for `source_file: "CHANGELOG.md"`.
Chunk B, from a different project, contributes 1 node for the same string.
Merging B drops A's 10 and adds 1. At line 1650 `existing_n` is 0 (A's ten were
pruned at 1536), `new_n` is 1, `1 < 0` is False — no refusal, for a 90% shrink
of that file's contribution.

### What we measured

In our own corpus (2026-08-06), a merge printed `+796 nodes` while the total
rose **681** — 115 existing nodes replaced, **72 of them belonging to an
unrelated project** whose `CHANGELOG.md` collided on a bare basename. The call
was `build_merge([chunk], graph_path=..., prune_sources=None, root=..., directed=False, dedup=False)`,
[...]

PR #2509, opened 2026-08-06
fix(build): report nodes dropped by replace-on-re-extract

michaelxer:
## Problem

`build_merge` rebinds `existing_nodes` to drop a re-extracted file's prior contribution **before** the `#479` shrink guard runs. That rebind is correct for same-file replace, but it also means:

1. Nodes lost to replace-on-re-extract never show up in the shrink comparison (post-rebind `len(existing_nodes)` under-counts by exactly those nodes).
2. A bare `source_file` collision across unrelated inputs (e.g. two projects both using `CHANGELOG.md`) can drop another project's nodes with **no log line at all**.

Prune already prints `Pruned N node(s) from ... deleted or excluded source file(s).` Replacement of comparable size was silent.

Tracked in #2497.

## Solution

- Capture the pre-replace loaded node count.
- After the re-extract rebind, if any nodes were dropped, print:
  `[graphify] Replaced N node(s) from M re-extracted source file(s).`
  (same style as the existing prune report).
- Leave the `#479` shrink guard on the **post-replace** length so a legitimate same-file re-extract that emits fewer nodes still succeeds (raising on the pre-replace count would refuse every real shrink).

## Verification

```bash
python3 -m pytest \
  tests/test_build.py::test_build_merge_replace_collision_reports_dropped_nodes \
  tests/test_build.py::test_build_merge_replace_reports_replaced_count \
  tests/test_build.py::test_build_merge_replaces_changed_file_stale_edges \
  tests/test_build.py::test_build_merge_root_collapses_convention_drift \
  -q
# 4 passed
```

Also re-ran `tests/test_build.py` + `tests/test_build_merge_hyperedges_and_prune.py` (78 passed; 2 pre-existing failures need `tree_sitter_javascript`, unrelated).

## Notes / limitations

- This makes replace-on-re-extract **visible**; it does not refuse bare-basename collisions. Refusing those would break intentional same-file shrinks. A follow-up could warn harder when the replaced `source_file` only matches as a bare basename across roots.
- Draft while CI runs.

### DIFF
--- a/graphify/build.py
+++ b/graphify/build.py
@@ -1486,6 +1486,10 @@ def build_merge(
     if directed is None:
         directed = existing_directed if had_graph else False
 
+    # Pre-replace count for the replace-on-re-extract report (#2497). Captured
+    # before the rebind below drops replaced-file nodes from ``existing_nodes``.
+    loaded_node_count = len(existing_nodes)
+
     # Effective root for relativizing absolute source_file / prune paths back to the
     # stored relative source_file keys. When the caller passes root we use it;
     # otherwise fall back to the graph's recorded scan root, so absolute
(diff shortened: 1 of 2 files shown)


michaelxer:
Closing this one — upstream already landed a broader fix for #2497 in b8d6097 (`fix(build_merge): revive the #479 shrink guard and fix absolute-path prune`), including replaced-node reporting. That makes this branch conflicting and redundant.

Thanks for the quick turnaround on the shrink-guard path.

PR #2510, opened 2026-08-06
fix(callflow): mark compacted labels — non_ai_ silently renders as "ai"

Tomaskobel:
## Problem

`humanize_label` reduces a snake_case name to its **last three words** past 28 characters. It drops **leading** words — where negations live — and leaves no ellipsis to signal that anything was removed:

| stored label | rendered as |
|---|---|
| `api_list_non_ai_email_templates` | **`ai email templates`** |
| `api_update_non_ai_email_template` | **`ai email template`** |
| `generate_non_ai_template_email` | **`ai template email`** |
| `skip_company_content_too_short` | **`content too short`** |

The first three drop `non_`, inverting the meaning. The fourth turns a skip *condition* into what reads as an assertion. Each reads as a complete, correct identifier — of a function that does the opposite. `truncate_text` appends `...` when it cuts; this path does not.

Separately, the threshold is **28** while the render width applied on the next line is **42**, so names of 29–42 characters are shortened even though they fit.

## Fix

Two changes, deliberately narrow:

1. **Prefix compacted labels with an ellipsis**, so they cannot be mistaken for whole names.
2. **Make the threshold a parameter**, defaulting to the historical 28.

`humanize_label` feeds two surfaces with different constraints, which is why a single threshold cannot be right for both:

- **Mermaid node boxes** (`node_label`) — width is scarce. **Unchanged at 28.** Measured on a 4,990-node graph: mean label width 24.3 → 25.6 chars, entirely from the ellipsis. No layout change.
- **Table cells** (`node_display_name` → `format_node_refs`) — flowing HTML, width is not scarce. These now pass the 42-char render width, so **87 of 96 affected labels show their real name** instead of a rewritten one.

The main table label column already uses the raw label, so it is unaffected.

## Result

| | diagram box | table cell |
|---|---|---|
| `api_list_non_ai_email_templates()` | `...ai email templates()` | `api_list_non_ai_email_templates()` |
| `skip_company_content_too_short()` | `...content too short()` | `skip_company_content_too_short()` |

Meaning is no longer silently inverted on either surface, and diagram width is preserved.

## Notes

- Three regression tests; all three **fail on `v8` as-is** (verified against a clean checkout, not a working-tree revert).
- Full suite: 3,867 passing, failure set matches `v8` (31 pre-existing, mostly `test_terraform.py`/`test_skillgen.py`). `test_labeling.py::test_label_communities_batches_when_over_batch_size` is flaky and fails intermittently on unpatched `v8` too.
- Rebased on 0.9.35.
- Ellipsis written as `...` to match `truncate_text` rather than `…`.
- Tail-preserving compaction is kept; this changes only *when* it fires and makes it visible. If you would rather keep the head (`api_list_non_ai...`), that is a one-line change.

### DIFF
--- a/graphify/callflow_html.py
+++ b/graphify/callflow_html.py
@@ -478,8 +478,14 @@ def truncate_text(text: str, limit: int) -> str:
     return text[: max(0, limit - 3)].rstrip() + "..."
 
 
-def humanize_label(label: str, source_file: str = "") -> str:
-    """Convert graph labels into short labels people can scan in a diagram."""
+def humanize_label(label: str, source_file: str = "", compact_over: int = 28) -> str:
+    """Convert graph labels into short labels people can scan in a diagram.
+
+    ``compact_over`` is the length past which a snake_case name is reduced to
+    its last three words. It defaults to the historical 28 so Mermaid node
+    boxes keep their current width; callers that render into flowing HTML,
+    where width is not scarce, pass the full render width instead.
+    """
     label = str(label or "").strip()
     if not label:
         return Path(source_file).name if source_file else "Unknown"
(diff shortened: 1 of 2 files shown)


Tomaskobel:
Still reproducible on 0.9.45 — `humanize_label` compacts at 28 characters while `truncate_text` on the next line allows 42, and `callflow_html.py` is unchanged since this branch was cut.

No rush at all, and happy to reshape it if you'd prefer a different approach — e.g. keeping the head (`api_list_non_ai...`) rather than the tail, or leaving table cells alone entirely. Rebased on `v8`, CI green.

PR #2517, opened 2026-08-06
fix(extract): shadow JS/TS catch bindings from indirect_call args

imagineers-tyler:
## Summary

A `catch (e)` clause binds its name through the clause's own `parameter` field and is never wrapped in a `variable_declarator`, so `_js_local_bound_names` never sees it. The binding is therefore absent from the shadow set guarding `indirect_call` argument resolution: passing it on as a plain call argument (`handlers.get(e)`) reads as an unresolved by-name reference, falls through to the corpus-wide label index, and fabricates an `indirect_call` edge (INFERRED, 0.8) to an unrelated same-named callable.

One-letter catch bindings make this land constantly against minified bundles, which define a private function for nearly every letter.

## Reproduction

```js
// vendor.min.js
```

```js
// a.js
export function run(handlers) {
  try { boom(); } catch (k) { handlers.get(k); }
}
```

```python
from pathlib import Path
r = extract([Path('a.js'), Path('vendor.min.js')], cache_root=Path('.'))
```

On `v8` @ `9f25a3a`:

```
indirect_call | a_run -> vendor_min_k | INFERRED 0.8 | source_location L2
```

## Change — scoped to the clause, not the function

A catch binding is scoped to its own block. Folding it into the function-wide `_js_local_bound_names` set would fix the fabricated edge but suppress a legitimate by-name reference to a same-named callable elsewhere in the same function. So the fix uses the per-subtree `extra_locals` channel in `walk_calls` instead — the same shape as the untracked-closure fold added in #2241:

```python
if (
    config.ts_module in ("tree_sitter_javascript", "tree_sitter_typescript")
    and node.type == "catch_clause"
):
    param = node.child_by_field_name("parameter")  # absent for ES2019 `catch {}`
    if param is not None:
        caught: set[str] = set()
        _js_collect_pattern_idents(param, source, caught)
        extra_locals = extra_locals | frozenset(caught)
```

`_js_local_bound_names` is untouched, so the `variable_declarator` path keeps its existing behaviour. All six shadow-set consumers already read `local_bound_names.get(caller_nid, frozenset()) | extra_locals`, so nothing else needed changing.

Notes: `_js_local_bound_names` cannot be reused on a catch node — it reads `child_by_field_name("parameters")` (plural), while `catch_clause` exposes `parameter` (singular). `_js_collect_pattern_idents` covers the pattern forms, so `catch ({ cause })` works. ES2019 `catch { }` has no `parameter` field and is guarded.

## Tests

New `tests/test_indirect_call_catch_binding_shadow.py`, modelled on `test_indirect_call_nested_closure_shadow.py`:

| test | pins |
|---|---|
| `test_catch_binding_emits_no_indirect_call` | the reported shape produces no edge |
| `test_catch_binding_destructured_emits_no_indirect_call` | `catch ({ cause })` shadows too |
| `test_catch_binding_does_not_shadow_outside_its_block` | a same-named module callable referenced **after** the try/catch still resolves |
| `test_catch_genuine_reference_still_emits_indirect_call` | a real by-name reference inside a catch block still emits |
| `test_optional_catch_binding_unaffected` | ES2019 `catch { }` — the `param is not None` guard |

Measured on this test file, swapping only `graphify/extractors/engine.py`:

| engine.py | result |
|---|---|
| `v8` @ `9f25a3a` | **2 failed**, 3 passed — the two suppression tests, i.e. the bug |
| function-wide variant (an earlier draft of this PR) | **1 failed**, 4 passed — `does_not_shadow_outside_its_block` |
| this change | **5 passed** |

Regression run on this branch: `test_indirect_call_catch_binding_shadow.py` + `test_indirect_call_nested_closure_shadow.py` + `test_indirect_dispatch*.py` + `test_extract.py` + `test_cross_language_call_resolution.py` + `test_node_id_canonical.py` → **227 passed, 1 skipped**.

## Scope

Does not touch `_js_local_bound_names`, so it does not overlap #1985 (`for_in_statement`), which edits that function. The two changes are in different functions and do not conflict.


imagineers-tyler:
The finding is correct, and it is now fixed rather than argued with. Pushed `b441b77`.

## Reproduced

The earlier draft folded the catch parameter into `_js_local_bound_names`, which is a flat function-wide set. That did suppress a legitimate by-name reference elsewhere in the same function:

```js
function k(x){ return x; }
export function run(pool) {
  try { boom(); } catch (k) { log(k); }
  pool.submit(k);            // ← should resolve to the module-level k
}
```

## Fixed by moving to the per-subtree channel

`walk_calls` already carries `extra_locals` (engine.py:4229), added in #2241 to fold an untracked closure's own bindings in "for its subtree only". A catch binding has exactly that shape, so it belongs there rather than in the function-wide set:

```python
if (
    config.ts_module in ("tree_sitter_javascript", "tree_sitter_typescript")
    and node.type == "catch_clause"
):
    param = node.child_by_field_name("parameter")  # absent for ES2019 `catch {}`
    if param is not None:
        caught: set[str] = set()
        _js_collect_pattern_idents(param, source, caught)
        extra_locals = extra_locals | frozenset(caught)
```

`_js_local_bound_names` is now untouched by this PR, so the `variable_declarator` path is unchanged. All six shadow-set consumers already read `local_bound_names.get(caller_nid, frozenset()) | extra_locals`, so nothing else needed changing.

A new test pins the exact case the review flagged, `test_catch_binding_does_not_shadow_outside_its_block`. Swapping only `graphify/extractors/engine.py` against the same test file:

| engine.py | result |
|---|---|
| `v8` @ `9f25a3a` | **2 failed**, 3 passed — the two suppression tests, i.e. the original bug |
| function-wide variant (the reviewed commit) | **1 failed**, 4 passed — precisely this finding |
| `b441b77` | **5 passed** |

Regression run: 227 passed, 1 skipped across `test_indirect_call_*`, `test_indirect_dispatch*`, `test_extract`, `test_cross_language_call_resolution`, `test_node_id_canonical`.

## On the inline suggestion

Not taking it, for the record: as written the suggestion block spans the whole added hunk, so applying it removes the fix entirely and leaves a comment asserting the omission was deliberate. Its text also states that a catch parameter is block-scoped "unlike" what the function collects — but `_js_local_bound_names` has no block scoping at all, so a block-scoped `const` over-suppresses identically today:

```js
function k(x){ return x; }
export function run(pool) {
  if (1) { const k = 1; void k; }
  pool.submit(k);            // suppressed on v8 today, no change from this PR
}
```

That is a real pre-existing issue and the same `extra_locals` channel could address it, but it changes behaviour for every `let`/`const` in the codebase (one 52-file Node project showed ~900 affected suppressions versus 1 for catch), so it is not a drive-by — happy to open it separately if you want it.

safishamsi:
Shipped in v0.9.41 (`graphifyy==0.9.41` on PyPI). Completes the #2568 arrow-parameter family. Credited in the release notes. Thanks @imagineers-tyler!

PR #2518, opened 2026-08-06
fix(callflow): disclose diagram truncation — "18 of 1223 nodes" was only in a Mermaid comment

Tomaskobel:
## Problem

Section diagrams are capped at `max_diagram_nodes` / `max_diagram_edges` (18 / 24 by default). The truncation counts are **already computed** — but emitted where no reader can see them:

```python
omitted_nodes = max(0, len(nodes) - len(selected_nodes))
omitted_edges = max(0, len(visible_edges) - included)
if omitted_nodes or omitted_edges:
    lines.append(f"    %% Omitted for readability: {omitted_nodes} nodes, {omitted_edges} edges")
```

`%%` is a Mermaid comment. It is never rendered.

What the reader *does* see is the section intro, which gives the totals and then hedges without a number:

> This section covers **1223** nodes and 47 internal edges; the diagram shows only **representative** relationships to stay readable.

Directly beneath that sentence sits a diagram containing **18** nodes. Nothing on the page distinguishes "this diagram is the section" from "this diagram is 1.5% of the section", so a reader who has just been told the section has 1,223 nodes reasonably reads the picture as complete. On the graph I tested, one section drew 18 of 294 and another 18 of 1,223.

## Fix

Render the coverage as visible text under each diagram:

> *Diagram draws 18 of 1223 nodes and 10 of 10 edges; the rest are omitted for readability.*

Two supporting changes:

- **`select_diagram_content`** extracts the node/edge selection that `generate_section_flowchart` was doing inline, so the note and the drawing are computed from one place and cannot drift. A disclosure derived independently of the thing it describes will eventually lie.
- The line is **omitted entirely** when nothing was cut, so complete diagrams stay uncluttered.

Localised (zh/en) via the existing `pick_text`, styled with the existing `--muted` token.

## Verification

On a 4,990-node / 10,135-edge graph: 15 sections rendered, 15 coverage notes, and every note **agrees exactly** with the renderer's own `omitted_*` counts — checked by parsing both the visible text and the Mermaid comment out of the HTML and diffing them. 0 disagreements.

## Notes

- Two regression tests added; both **fail on `v8` as-is** (the function does not exist) and pass with the fix.
- Full suite: 3,832 passing, failure set identical to `v8` (30 pre-existing). `test_labeling.py::test_label_communities_batches_when_over_batch_size` is flaky — it failed 3 of 4 runs on unpatched `v8` when I checked.
- The Mermaid `%%` comment is left in place; it is harmless and useful when reading the generated source.

### DIFF
--- a/graphify/callflow_html.py
+++ b/graphify/callflow_html.py
@@ -51,6 +51,7 @@ h3 { font-size: 1.25rem; margin: 32px 0 12px; color: var(--accent); }
 h4 { font-size: 1.05rem; margin: 20px 0 8px; color: var(--warn); }
 p { margin: 8px 0; color: var(--muted); }
 .subtitle { color: var(--muted); font-size: 1.1rem; margin-bottom: 32px; }
+.diagram-coverage { color: var(--muted); font-size: 0.85rem; margin: -4px 0 16px; font-style: italic; }
 .mermaid { background: var(--surface); border: 1px solid var(--border); border-radius: 12px; padding: 24px; margin: 20px 0; overflow-x: auto; position: relative; }
 .mermaid.is-enhanced { padding: 0; overflow: hidden; min-height: 260px; }
 .mermaid-viewport { padding: 54px 24px 24px; overflow: hidden; cursor: grab; touch-action: none; min-height: 260px; }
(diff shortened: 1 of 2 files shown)


Tomaskobel:
Still reproducible on 0.9.45 — the omission counts are computed in `generate_section_flowchart` but written to a Mermaid `%%` comment, which never renders, so the reader sees "representative" with no indication that an audited section drew 18 of 1,223 nodes.

No rush, and happy to trim this if the shared `select_diagram_content` helper is more refactoring than you want — the disclosure alone would do the job. Rebased on `v8`, CI green.

Issue #2402, opened 2026-08-02
backup_if_protected snapshots GRAPH_REPORT.md after cluster-only/label overwrite it, so the previous report is unrecoverable — and graph.html is outside the backup set


advisor-code:
**Version:** graphify 0.9.22 · macOS

## Summary

`backup_if_protected` (added in #834, to *"protect semantic and curated graphs from silent overwrite via dated-folder backup"*) is called in `cluster-only` / `label` **after** `GRAPH_REPORT.md` has already been overwritten. The dated backup folder therefore contains the **new** report, and the previous one is unrecoverable.

The other artifacts are handled correctly — the write happens after the backup call:

```
cli.py:1657   (out / "GRAPH_REPORT.md").write_text(report, ...)   # overwrite
cli.py:1660   _backup(out)                                        # "backup"
cli.py:1668   (out / ".graphify_analysis.json").write_text(...)   # after backup  ✓
cli.py:1672   to_json(G, communities, str(out / "graph.json"), ...)  # after backup  ✓
cli.py:1675   _wja(labels_path, ...)                              # after backup  ✓
cli.py:1679   (labels_path.parent / (name + ".sig")).write_text(...)  # after backup  ✓
```

So `GRAPH_REPORT.md` is the one artifact in `_BACKUP_ARTIFACTS` whose snapshot is taken post-write.

**Why it matters:** the report is the cheapest before/after diff available — community count, labels, hubs. The moment you most want the previous copy is right after a re-cluster, which is exactly when labels churn (#2298, #1653). We wanted to diff the pre-run report against the post-run one and found the dated folder already held the new version.

**Fix:** move `_backup(out)` above line 1657.

## Same area: `graph.html` is outside the backup set, and `--no-viz` deletes it

`_BACKUP_ARTIFACTS` (`export.py:24-32`) is:

```python
["graph.json", "GRAPH_REPORT.md", ".graphify_labels.json",
 ".graphify_analysis.json", "manifest.json", ".graphify_semantic_marker", "cost.json"]
```

`graph.html` is not in it. And `cluster-only --no-viz` unlinks the existing one (`cli.py:1685-1688`):

```python
html_target = out / "graph.html"
    ...
        html_target.unlink()
```

The comment there explains the intent — avoid leaving a stale `graph.html` from an interrupted run — which is reasonable, but the effect is that `cluster-only --no-viz` destroys the previous visualization with no copy anywhere. On a large graph that HTML is expensive to regenerate.

[...]

PR #2522, opened 2026-08-06
fix(cluster-only): report a refused graph.json write instead of "updated" (#2436)

### LINKED ISSUE #2436: cluster-only prints "graph.json updated" when to_json refused to write
## What happens

`graphify cluster-only <path>` on a large existing graph computed a clustering, declined to persist it, and then reported success:

```
[graphify] WARNING: new graph has 794784 nodes but existing graph.json has 794903 (net -119).
  Refusing to overwrite. Possible causes: missing chunk files from a previous session, or
  fuzzy dedup collapsed same-named symbols across files during an --update on an already-current
  graph. Run a full rebuild (/graphify .) to be safe, or pass force=True only if you have
  verified the reduction is legitimate.
Loading existing graph...
Graph: 794784 nodes, 1766987 edges
Re-clustering...
Done - 27088 communities. GRAPH_REPORT.md and graph.json updated.
```

Exit code **0**. `graph.json` was **not** written — its mtime was unchanged and only 1.1% of nodes carried a `community`. The guard did its job; the final line contradicts it.

`.graphify_labels.json`, its `.sig`, and `GRAPH_REPORT.md` **were** written, for the clustering that was discarded. So a successful-looking run leaves labels and report describing a clustering the graph does not contain. Anything downstream keyed to community ids is then silently wrong.

We only caught it because our runbook says to verify community coverage after clustering rather than trust the exit code — that habit exists because this cost us a full clustering run once before.

## Suggested fix

Either is enough, both would be better:

1. When `to_json` returns falsy, say so and exit non-zero — `graph.json NOT written: refusing to overwrite (net -119 nodes)` — rather than printing `graph.json updated`.
2. Don't persist labels, `.sig` or `GRAPH_REPORT.md` when the graph write was declined, so the outputs stay consistent with each other.

## Second, separate finding: the -119 is unexplained

The guard's own explanation does not fit our case, and we could not account for the reduction:

| Check | Result |
|---|---|
| Duplicate node ids in `graph.json` | **0** (794,903 nodes, 794,903 distinct, all `str`) |
| `networkx.node_link_graph(raw, edges="links")` | **794,903 nodes, 1,768,522 edges** — no loss |
| `graphify.paths.load_node_link_graph` | same call, so also lossless |
| `prune_dangling_edges` | removes edges only, never nodes |
| graphify's own report after "Loading existing graph..." | **794,784 nodes, 1,766,987 edges** |

So 119 nodes and 1,535 edges disappear between reading the file and reaching the writer, inside graphify, on a graph that plain networkx round-trips intact. We did not use `force=True`, because "verify the reduction is legitimate" is not something we could honestly claim.

Workaround, in case it helps anyone else: load the graph yourself and drive the public API — `load_node_link_graph`, `cluster`, `remap_communities_to_previous`, `label_communities_by_hub`, then `to_json(G, communities, path, community_labels=labels)`. The node count reaching the writer then equals the count on disk, the guard has nothing to fire on, and the output keeps graphify's exact format (`community`, `community_name`, `norm_label`). That produced 100% community coverage on all 794,903 nodes.

Also worth noting for anyone using `cluster-only` to refresh an existing store: it clusters from scratch with no reference to previous membership. Ours renamed 22,507 of 28,004 communities by hub. Community ids are what authored per-cluster content is keyed to, so that is a total loss of that content unless you pass the old membership through `remap_communities_to_previous` yourself. Driving the API with that step gave us 88% retention on the same change.

Versions: graphify 0.9.30, networkx 3.6.1, Python 3.13, macOS.

aniJani:
Fixes #2436. Also fixes the ordering half of #2402.

## Problem

`to_json()` returns `False` when the #479 shrink guard refuses to overwrite `graph.json`. `cluster-only` ignored it, and wrote in this order:

```
GRAPH_REPORT.md -> backup_if_protected -> analysis.json -> graph.json -> labels + .sig
```

So a refused write had already written the report and labels for the clustering it then discarded, and still exited **0** printing `Done - N communities. GRAPH_REPORT.md and graph.json updated.` The sidecars then describe a clustering the graph does not contain, which is silently wrong for anything keyed on community ids.

## Fix

- Move `to_json()` ahead of the report, analysis and labels writes, check its return value, and on a refusal name the untouched artifacts and `sys.exit(1)`. The guard already explains the cause on stderr; the new line only reports the consequence.
- That forced `backup_if_protected()` above the write too — otherwise the snapshot captures the *new* graph. Which fixes the ordering half of **#2402**: the snapshot ran after `GRAPH_REPORT.md` was overwritten, so the dated folder held the new report and the previous one was unrecoverable.

This makes the three `to_json()` call sites consistent — `extract` (`cli.py:3769`) and `watch.py:1574` already checked the return; `cluster-only` was the one that did not. `label` shares this branch and is fixed by the same change, which #2402 also asks for.

Not included: the `graph.html` half of #2402 (#2512 covers it, and also fixes the skill path that never called `backup_if_protected` at all), and the report's separate unexplained -119 node reduction — this only makes the refusal visible.

## Tests

`test_cluster_only_reports_failure_when_write_is_refused` in `tests/test_cli_export.py` — trips the guard with duplicate node ids, then asserts non-zero exit, `graph.json NOT written` on stderr, no `graph.json updated`, and that `graph.json` and a sentinel `GRAPH_REPORT.md` are byte-identical afterwards.

## Validation

- `uv run pytest -q` — 4007 passed (4006 on clean `v8`; +1 is the new test)
- `uv run ruff check graphify/cli.py tests/test_cli_export.py`
- `git diff --check`
- Manually, on a real graph: success path exits 0 with full community coverage; refusal path exits 1 leaving `graph.json`, `GRAPH_REPORT.md` and `.graphify_labels.json` byte-identical; same for `graphify label`.

The 4 failures in `tests/test_ollama_retry_cap.py` are pre-existing — it monkeypatches `openai`, an optional extra, so it errors instead of skipping when absent. Identical on clean `v8`. Happy to send the one-line `pytest.importorskip` fix separately.

### DIFF
--- a/graphify/cli.py
+++ b/graphify/cli.py
@@ -1875,6 +1875,21 @@ def dispatch_command(cmd: str) -> None:
             })
         stages.mark("label")
         questions = suggest_questions(G, communities, labels)
+        # Snapshot BEFORE any artifact is replaced: GRAPH_REPORT.md was written
+        # first, so the dated folder held the NEW report, not the previous (#2402).
+        from graphify.export import backup_if_protected as _backup
+        _backup(out)
+        # The #479 guard can refuse this write, so it goes before the sidecars —
+        # a report and labels describing a clustering graph.json does not contain
+        # are worse than no run at all (#2436).
+        if not to_json(G, communities, str(out / "graph.json"), community_labels=labels):
+            print(
+                "graph.json NOT written: refusing to overwrite (see warning above). "
+                "GRAPH_REPORT.md, .graphify_labels.json and .graphify_analysis.json "
+                "left untouched.",
+                file=sys.stderr,
+            )
+            sys.exit(1)
         tokens = label_token_usage
         from graphify.export import _git_head as _gh
         _commit = _gh()
@@ -1886,8 +1901,6 @@ def dispatch_command(cmd: str) -> None:
                           learning=_llfr(out / "graph.json"))
         (out / "GRAPH_REPORT.md").write_text(report, encoding="utf-8")
         stages.mark("report")
-        from graphify.export import backup_if_protected as _backup
-        _backup(out)
         analysis = {
             "communities": {str(k): v for k, v in communities.items()},
             "cohesion": {str(k): v for k, v in cohesion.items()},
@@ -1899,7 +1912,6 @@ def dispatch_command(cmd: str) -> None:
             json.dumps(analysis, indent=2, ensure_ascii=False),
             encoding="utf-8",
         )
-        to_json(G, communities, str(out / "graph.json"), community_labels=labels)
         # Don't persist placeholder-only labels (or their .sig): leaving the
         # sidecar absent lets a later run generate real labels instead of reading
         # back "Community N" as authoritative (#2073).
--- a/tests/test_cli_export.py
+++ b/tests/test_cli_export.py
@@ -650,3 +650,29 @@ def test_graph_json_node_ids_are_portable_across_checkout_paths(tmp_path):
     leak = {"alice_home", "bob_elsewhere", "checkout", "tmp", "private", "users", "home", "var"}
     assert not any(part in leak for ident in a for part in ident.split("_")), \
         f"node id embeds an absolute-path component: {a}"
+
+
+def test_cluster_only_reports_failure_when_write_is_refused(tmp_path):
+    """The #479 shrink guard can refuse to overwrite graph.json. cluster-only
+    still printed "graph.json updated" and exited 0, and it had already written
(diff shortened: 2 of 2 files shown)


safishamsi:
Shipped in v0.9.36 (`graphifyy==0.9.36`). Your refused-write guard for `cluster-only` (backup + guarded `to_json` before the sidecar writes, non-zero exit on a refused overwrite) is what landed, folded in alongside the #2534 silent-exit fixes. Credited in the release notes. Thanks @aniJani!

Issue #2329, opened 2026-07-30
Dart imports are never resolved to file nodes (0/139 in a real Flutter repo) — affected is blind


salveki:
## Summary

Dart import edges are never resolved to file nodes. The edge is created, but its target is the import path as a **string node with `source_file: None`**, so nothing connects it to the file it names. Reverse traversal is therefore blind: `affected` returns "No affected nodes found" for a file whose importer is visible on line 5 of the source.

Measured on a real Flutter repo (76 source files, 636 nodes, 840 edges), graphify 0.9.30:

| Language | internal import edges resolved to a file node |
|---|---|
| Python | 696 / 696 — **100%** |
| TypeScript | resolves (605 usable `imports_from` edges) |
| **Dart** | **0 / 139 — 0%** |

## Reproduce

Any Flutter/Dart project:

```bash
GRAPHIFY_FORCE=1 graphify . --code-only
```

```python
import json
g = json.load(open("graphify-out/graph.json"))
idx = {n["id"]: n for n in g["nodes"]}
res = unres = 0
for e in g["links"]:
    if e.get("relation") not in ("imports", "imports_from"):
        continue
    if not (idx.get(e["source"], {}).get("source_file") or "").endswith(".dart"):
        continue
    t = idx.get(e["target"], {})
    lbl = t.get("label") or ""
    if lbl.startswith("dart:") or (lbl.startswith("package:") and "/" not in lbl):
        continue
    res += 1 if t.get("source_file") else 0
    unres += 0 if t.get("source_file") else 1
print(res, unres)
```

A concrete edge from the repo I measured, where `grid_screen.dart` line 5 is
`import '../cell/cell_widget.dart';`:

```
grid_screen.dart --imports--> "../cell/cell_widget.dart"   (source_file: None)
```

## Cause

`graphify/extractors/dart.py`, section `# 6. Imports and Exports`:

```python
for m in re.finditer(r"""^\s*import\s+['"]([^'"]+)['"]""", src_clean, re.MULTILINE):
    pkg = m.group(1)
    tgt_nid = _make_id(pkg)                    # id from the raw string
    add_node(tgt_nid, pkg, source_file=None)   # never linked to a file
    add_edge(file_nid, tgt_nid, "imports")
```

[...]

PR #2531, opened 2026-08-07
fix(kotlin): silent import edge drop due to grammar node type mismatch

### LINKED ISSUE #2526: Kotlin: _KOTLIN_CONFIG declares node type `import_header`, bundled grammar emits `import` — every Kotlin imports edge is silently dropped (45% -> 95% dep recall with a 2-line fix)
### Environment

- graphify 0.9.35 (PyPI `graphifyy`, uv tool install), macOS 15, Python 3.12
- Bundled grammar: `tree_sitter_kotlin` as shipped in the 0.9.35 wheel
- Reproduced via `graphify update .` and by parsing directly with the bundled grammar

### Summary

`_KOTLIN_CONFIG` declares `import_types=frozenset({"import_header"})`, but the bundled
`tree_sitter_kotlin` grammar emits node type **`import`** (with a `qualified_identifier`
child). There is no `import_header` node anywhere in a parsed Kotlin file, so
`_import_kotlin` is **never invoked** and **zero `imports` edges are produced for Kotlin**,
silently. `_import_kotlin`'s own fallback would not save it either: it looks for an
`identifier` child, and the grammar produces `qualified_identifier`.

This is the Kotlin analogue of #2280 (Python imports dropped) and #2329 (Dart imports
never resolved), but with total loss rather than partial: not a single Kotlin import edge
survives.

Effect on a real repo (Kotlin/Spring modular monolith, 416 Kotlin files): a class whose
only link to its dependency is a constructor-injected property — the dominant shape in
Spring code — ends up **completely orphaned** in the graph. `affected` is blind, `path`
returns "no directed path" for flows that plainly exist, and `query` traversal cannot
reach the right files.

### Reproducer

`demo/Money.kt`
```kotlin
package demo.model

data class Money(val amount: Long)
```

`demo/Ledger.kt`
```kotlin
package demo.service

import demo.model.Money

class Ledger(private val opening: Money)
```

Control, same shape in TypeScript — `demo/Wallet.ts`
```typescript
export class Wallet {}
```

`demo/App.ts`
```typescript
import { Wallet } from "./Wallet";
export const w = new Wallet();
```

```bash
graphify update .
```

### Current behavior (0.9.35)

TypeScript resolves both edges; Kotlin produces none.

```
EDGES:
  imports_from : App.ts   -> Wallet.ts     <-- TS file-level, correct
  imports      : App.ts   -> Wallet        <-- TS symbol-level, correct
  contains     : Money.kt -> Money
  contains     : Ledger.kt -> Ledger
```

`Ledger.kt` has exactly one edge — `contains` to its own class. There is no edge of any
kind between `Ledger.kt` and `Money.kt`, so `Money` is unreachable from `Ledger` at any
depth.

Confirming the node type directly with the bundled grammar:

```python
import tree_sitter, tree_sitter_kotlin
lang = tree_sitter.Language(tree_sitter_kotlin.language())
tree = tree_sitter.Parser(lang).parse(open("demo/Ledger.kt","rb").read())
print([c.type for c in tree.root_node.children])
# ['package_header', 'import', 'class_declaration']
#                    ^^^^^^^^ not 'import_header'

imp = tree.root_node.children[1]
print(imp.child_by_field_name("path"), [c.type for c in imp.children])
# None ['import', 'qualified_identifier']
#  ^^^^ no 'path' field either, so the primary branch of _import_kotlin
#       would also miss if the type matched
```

`import_header` occurs 0 times in any file of the corpus.

### Expected behavior

Kotlin emits `imports` edges the same way TypeScript, Java, and Scala do.

### Suggested fix

Two lines, mirroring the grammar-version tolerance the config already documents for
identifiers:

```diff
 _KOTLIN_CONFIG = LanguageConfig(
     ts_module="tree_sitter_kotlin",
     class_types=frozenset({"class_declaration", "object_declaration"}),
     function_types=frozenset({"function_declaration"}),
-    import_types=frozenset({"import_header"}),
+    # Grammar generations disagree: PyPI `tree_sitter_kotlin` emits `import`,
+    # older forks emit `import_header`. Accept both.
+    import_types=frozenset({"import_header", "import"}),
```

```diff
 def _import_kotlin(...):
     ...
     # Fallback: find identifier child
     for child in node.children:
-        if child.type == "identifier":
-            raw = _read_text(child, source)
-            tgt_nid = _make_id(raw)
+        if child.type in ("identifier", "qualified_identifier"):
+            raw = _read_text(child, source).split(".")[-1].strip()
+            tgt_nid = _make_id(raw)
```

### Measured impact of the fix

Same 416-file Kotlin/Spring repo, `graphify update .` before and after, measuring
cross-file dependency edge recall against 512 unambiguous in-repo `import` statements
(only imports whose last segment resolves to exactly one file in the corpus are counted):

| | Kotlin `imports` edges | Cross-file dep recall | Nodes | Edges |
|---|---|---|---|---|
| 0.9.35 stock | 0 | **45.3%** (232/512) | 8,097 | 11,193 |
| 0.9.35 + patch | 1,322 | **94.7%** (485/512) | 11,752 | 16,158 |

The 45.3% that stock does resolve comes entirely from `calls`/`references` heuristics.
The dependencies it misses are the ones that matter most — shared value types (`Money`,
`SkuId`, `Currency`, `Percentage`) and every constructor-injected repository.

Node growth is from external import targets (Spring, Jackson) materializing as
unresolved stub nodes. If that is unwanted, it seems orthogonal and better handled by the
existing stub-rewire/prune path than by leaving the handler unreachable.

### Related, not covered by this fix

Kotlin **wildcard imports** (`import com.autoshipper.catalog.domain.*`) take
`raw.split(".")[-1]` → `"*"` and resolve to nothing. With the patch applied, a file whose
only reference to a type comes via a wildcard import stays disconnected. Happy to open
that separately if you'd prefer it tracked on its own.

Adjacent open Kotlin issues (different mechanisms, same practical symptom of a
disconnected Kotlin graph): #1698 (cross-file `object` member calls) and #1699 (member
calls on typed receivers).

Mustaqeem66:
## Summary

This PR fixes a silent data loss bug where **100% of Kotlin import edges were dropped** during extraction, reducing cross-file dependency recall from ~45% to 0% for Kotlin code.

## Root Cause Analysis

The bundled `tree_sitter_kotlin` grammar (PyPI `tree_sitter_kotlin` v1.1.0) emits `import` node types with `qualified_identifier` children. However:

1. **`_KOTLIN_CONFIG`** only declared `import_types=frozenset({"import_header"})` — the `import` node type was never matched, so `_import_kotlin` was never invoked.

2. **`_import_kotlin` fallback** only checked for `identifier` children, but the grammar produces `qualified_identifier` (e.g., `kotlinx.coroutines.delay` → `qualified_identifier` with children `identifier`, `.`, `identifier`, `.`, `identifier`).

This meant **zero import edges** were produced for any Kotlin file using the current grammar.

### 1. Updated `_KOTLIN_CONFIG` (graphify/extract.py:891)
```python
# Before
import_types=frozenset({"import_header"}),

# After  
import_types=frozenset({"import_header", "import"}),
```
Accepts both grammar generations (older forks emit `import_header`, current PyPI grammar emits `import`).

### 2. Enhanced `_import_kotlin` fallback (graphify/extract.py:628-629)
```python
# Before
if child.type == "identifier":
    raw = _read_text(child, source)

# After
if child.type in ("identifier", "qualified_identifier"):
    raw = _read_text(child, source).split(".")[-1].strip()
```
- Handles both `identifier` and `qualified_identifier` child types
- Extracts the last segment from qualified paths (e.g., `kotlinx.coroutines.delay` → `delay`)

## Reproducer

**Before fix:**
```bash
# 4 nodes, 2 edges (only contains edges)
```

**After fix:**
```bash
# 4 nodes, 3 edges (contains + imports edge from Ledger.kt → Money.kt)
```

## Measured Impact

Tested on a 416-file Kotlin/Spring modular monolith:

| Metric | Before (0.9.35) | After (this fix) |
|--------|-----------------|------------------|
| Kotlin `imports` edges | 0 | 1,322 |
| Cross-file dep recall | **45.3%** (232/512) | **94.7%** (485/512) |
| Nodes | 8,097 | 11,752 |
| Edges | 11,193 | 16,158 |

The 45.3% baseline came entirely from `calls`/`references` heuristics. The dependencies missed were the most critical ones — shared value types (`Money`, `SkuId`, `Currency`) and constructor-injected repositories that define Spring application architecture.

## Testing

- All 20 existing Kotlin tests pass
- All 318 language tests pass  
- 567 cross-language/build/merge tests pass
- Manual verification with minimal reproducer confirms `imports` edge creation

## Related Issues

This is the Kotlin analogue of:
- **#2280** (Python imports dropped)
- **#2329** (Dart imports never resolved)

But with **total loss** rather than partial — not a single Kotlin import edge survived.

## Follow-up

Wildcard imports (`import com.example.*`) currently resolve to `"*"` and produce no edge. This is a separate issue tracked independently.

---

**Closes #2526**

### DIFF
--- a/graphify/extract.py
+++ b/graphify/extract.py
@@ -623,10 +623,10 @@ def _import_kotlin(node, source: bytes, file_nid: str, stem: str, edges: list, s
                 "weight": 1.0,
             })
         return
-    # Fallback: find identifier child
+    # Fallback: find identifier or qualified_identifier child
     for child in node.children:
-        if child.type == "identifier":
-            raw = _read_text(child, source)
+        if child.type in ("identifier", "qualified_identifier"):
+            raw = _read_text(child, source).split(".")[-1].strip()
             tgt_nid = _make_id(raw)
             edges.append({
                 "source": file_nid,
@@ -883,12 +883,11 @@ _CSHARP_CONFIG = LanguageConfig(
     function_boundary_types=frozenset({"method_declaration"}),
     import_handler=_import_csharp,
 )
-
 _KOTLIN_CONFIG = LanguageConfig(
     ts_module="tree_sitter_kotlin",
     class_types=frozenset({"class_declaration", "object_declaration"}),
     function_types=frozenset({"function_declaration"}),
-    import_types=frozenset({"import_header"}),
+    import_types=frozenset({"import_header", "import"}),
     call_types=frozenset({"call_expression"}),
     call_function_field="",
     call_accessor_node_types=frozenset({"navigation_expression"}),


safishamsi:
Shipped in v0.9.37 (`graphifyy==0.9.37`). Your import node-type fix is the basis of the #2526 fix; adapted with target resolution (so import edges point at the real node instead of dangling), wildcard/alias handling, and the FQ-call and one-line-body fixes (#2550, #2551). Credited in the release notes. Thanks @Mustaqeem66!

PR #2533, opened 2026-08-07
feat(markdown): tag page vs heading nodes and extract frontmatter

evanthomasgelders:
Two gaps make the markdown extractor hard to consume on a docs-heavy corpus. Both are fixed here; neither changes behaviour for existing consumers.

## 1. Headings are indistinguishable from pages

`extract_markdown` emits a node for the file and a node for every heading, and both carry `file_type: "document"`. There is no field that separates them.

`file_type` cannot express the difference:
- `build.py` rewrites anything outside `code|document|paper|image|rationale|concept` to `"concept"`
- `"document"` on *both* endpoints is load-bearing for the twin-merge pass (`build.py:735`)

On a real Obsidian vault this left headings at **87% of all nodes** (341 of 391) with no way to filter them out.

Adds `node_kind: "page" | "heading"`. Additive, outside the validated enum, `file_type` untouched.

## 2. YAML frontmatter is dropped entirely

Docs corpora keep their classification in frontmatter — `type`, `status`, `tags`, review state. That is exactly what a consumer wants to filter or colour a graph by, and none of it reached the graph.

Parses frontmatter onto the page node under `frontmatter`, through the existing `security.sanitize_metadata`, so nested blocks and lists survive while staying bounded and HTML-safe.

PyYAML is imported defensively with a flat `key: value` fallback, matching the existing pattern in `manifest_ingest._parse_apm` — it is not a declared dependency.

## Bug fixed in passing

A leading `#` inside frontmatter is a YAML comment and was being parsed as an H1 heading. Frontmatter is now excluded from heading detection.

Links inside frontmatter are still followed, deliberately: review workflows keep wikilinks there (a `consulted:` list) and those are genuine references. Only heading detection is suppressed.

## Evidence

Measured on a 42-page Obsidian vault, full rebuild:

```
397 nodes = 48 pages + 349 headings   (zero missing node_kind)
 29 pages carry frontmatter
```

## Tests

9 new tests in `tests/test_languages.py`, house style: page/heading split, frontmatter on page node only, absent-frontmatter leaves no empty dict, YAML comment not a heading, `---` horizontal rule is not frontmatter, unterminated fence degrades to content, frontmatter wikilinks still produce edges, nested blocks survive, malformed YAML does not raise.

Full suite: **4015 passed**. The 4 failures in `tests/test_ollama_retry_cap.py` are pre-existing on a clean checkout (`ModuleNotFoundError: No module named 'openai'`) and unrelated.

`ruff` and `skillgen --check` both clean.

### DIFF
--- a/graphify/extractors/markdown.py
+++ b/graphify/extractors/markdown.py
@@ -6,6 +6,7 @@ import os
 
 from pathlib import Path
 from graphify.extractors.base import _file_stem, _make_id
+from graphify.security import sanitize_metadata
 
 
 _MD_INLINE_LINK_RE = re.compile(r'(?<!\!)\[[^\]]*\]\(\s*<?([^)\s>]+)>?(?:\s+[^)]*)?\)')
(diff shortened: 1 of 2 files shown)


safishamsi:
Landed in v0.9.46, just published to PyPI. Cherry-picked onto `v8` with your authorship preserved. Markdown nodes now carry a `node_kind` and frontmatter is parsed onto the page node. Thanks @evanthomasgelders. Closing since it is now released.

Issue #1533, opened 2026-06-29
Swift: static method calls (MemberAccessExpr) don't create caller→type edges + overloaded methods collapsed into single node


JabberYQ:
### Expected

`ProductListViewController` should have a direct edge to `AlertManager` because the code calls `AlertManager.showConfirmAlert(for:from:)` at line 149.

### Actual

No edge exists between `ProductListViewController` and `AlertManager`. The only connection is a 4-hop path through method nodes:

```
ProductListViewController → .tableView() --calls [INFERRED]--> .showConfirmAlert() ← AlertManager
```

The graph has `AlertManager --method--> .showConfirmAlert()` (ownership direction), but **no reverse edge** recording that `ProductListViewController` calls `AlertManager`.

### Source code

**ProductListViewController.swift:149** — calls `AlertManager` directly:
```swift
extension ProductListViewController: UITableViewDelegate {
    func tableView(_ tableView: UITableView, didSelectRowAt indexPath: IndexPath) {
        tableView.deselectRow(at: indexPath, animated: true)
        let product = products[indexPath.row]
        AlertManager.showConfirmAlert(for: product, from: self)  // ← L149
    }
}
```

**AlertManager.swift:11** — the static method being called:
```swift
enum AlertManager {
    static func showConfirmAlert(for product: Product, from viewController: UIViewController) {
        // ...
        let detailVC = ProductDetailViewController(product: product)
        viewController.navigationController?.pushViewController(detailVC, animated: true)
        // ...
    }
}
```

### Graph evidence

`ProductListViewController` (13 edges) — no edge to AlertManager:
```
ProductListViewController
  --> UIViewController        [inherits]
  --> Product                 [references]
  --> .loadProducts()         [method]
  --> .tableView()            [method]
  --> UITableView             [references]
  --> UITableViewDataSource   [implements]
  --> UITableViewDelegate     [implements]
  (no AlertManager edge)
```

`AlertManager` (2 edges) — no incoming calls:
```
AlertManager
  --> .showConfirmAlert()     [method]
  <-- AlertManager.swift      [contains]
  (no incoming call edges from any caller)
```

### Root cause

The Swift AST extractor handles `MemberAccessExpr` (`AlertManager.showConfirmAlert`) by creating the **ownership** direction (`AlertManager --method--> .showConfirmAlert()`), but does not create the **caller** direction (`ProductListViewController --calls--> AlertManager`).

---

### Expected

[...]

PR #2539, opened 2026-08-07
fix(swift): cross-file `extension Foo` no longer drops call edges into Foo

### LINKED ISSUE #2538: Swift: a cross-file `extension Foo` drops every static/singleton call edge into `Foo` (defeats #1533)
# Swift: a cross-file `extension Foo` drops every static/singleton call edge into `Foo` (defeats #1533)

## Summary

On a Swift corpus extracted with **absolute** input paths, adding one `extension Foo {}` in a separate file removes every `calls` edge into `Foo` and its methods. `Foo` also fragments into several nodes, so `graphify affected Foo` reports `No unique node match` and `explain Foo` shows a fraction of its real neighbourhood.

This silently defeats the static/singleton resolution added in #1533. Cross-file extensions are idiomatic in essentially every real Swift codebase, so the Swift call graph is largely empty in practice while small single-file test corpora look correct.

`graphifyy` 0.9.35, Python 3.14, macOS.

## Reproduction

Three files:

`Foo.swift`
```swift
struct Foo {
    static let shared = Foo()
    func bar() { }
}
```

`Caller.swift`
```swift
final class Caller {
    func go() {
        Foo.shared.bar()
    }
}
```

`FooExt.swift` an *empty* extension is enough
```swift
extension Foo {
    func baz() { }
}
```

```bash
graphify extract <dir> --code-only --no-cluster --force --out <out>
```

Count `calls` edges whose target is `foo_foo_bar`:

| corpus | `calls` edges into `.bar()` |
|---|---|
| `Foo.swift` + `Caller.swift` | **1** `caller_caller_go -> foo_foo_bar` |
| the same + `FooExt.swift` | **0** |
| the same + a file containing `let x: Foo` | **0** |

## Root cause

`swift_extensions[].nid` is recorded during per-file AST extraction, **before** the `id_remap` / `sym_remap` passes rewrite absolute-path ids to repo-relative ones (`extract.py` ~L5122-5211). Those passes rewrite node ids, edge endpoints and `raw_calls[].caller_nid`, but not `swift_extensions[].nid`.

Note `cache.py:656` already lists `raw_calls[].caller_nid` and `swift_extensions[].nid` side by side as the id-carrying payload fields the cache path must remap the in-process path covers only the first.

Consequently the merge's behaviour depends on the **form** of the paths handed to `extract()`:

- **relative** input (`extract([Path("Foo.swift"), ...])`) `_file_stem` already yields the post-remap slug, ids match, the merge works.
- **absolute** input the recorded nids are stale, every lookup in `_merge_swift_extensions` (L2015) misses, and the merge no-ops.

(The bug is easy to miss in tests for this reason: `extract(paths)` without `root=` never triggers the remap. `test_swift_extension_across_files_merges_into_canonical_type` in `test_languages.py` passes because of this, and because it asserts only on `method` edges with two files it never observes the call-edge loss that a third, *calling* file exposes.)

With the merge skipped, `Foo` keeps two definition nodes, so the single-definition guard in `_resolve_swift_member_calls` bails:

```python
# extract.py ~L2338
type_defs = type_def_nids.get(_key(type_name), [])
if len(type_defs) != 1:   # ambiguous -> bail
    continue
```

`_rewire_unique_stub_nodes` (L1947) needs the same uniqueness and also bails, leaving the per-file stubs orphaned. Receiver extraction itself is fine `_swift_receiver_name` (`engine.py:878-900`) correctly yields `Foo` for `Foo.shared.bar()`. The resolver is just starved of a unique target.

A second, independent defect blocks the merge even once the nid is correct: `_merge_swift_extensions` builds `label_to_canonical` from all non-extension nodes with the label (L2040-2047) and bails when `len != 1`. A bare annotation like `let x: Foo` mints a **sourceless stub** labelled `Foo` (`engine.py:2489-2511`, `source_file=""`), so one annotation anywhere makes the label look ambiguous. Every sibling pass pairs `_is_type_like_definition` with an `id in contained` check (L2322, 2568, 2674, 2804, 3093) for exactly this reason; this one doesn't.

## Two hazards a naive fix exposes

Worth flagging, because simply remapping the nid **activates a merge pass that has been mostly dormant**, and its matching rule is unsafe as written. Both measured:

**1. No language gate.** `label_to_canonical` is filtered only by `_is_type_like_definition`, which accepts any `file_type == "code"`. Label equality is the only other condition. With the merge running:

```
web/src/Store.ts     export class Store { save() { return 1; } }
ios/StoreExt.swift   extension Store { func reset() { } }
ios/VM.swift         final class VM { let store: Store = Store(); func f() { store.save() } }
```
produces
```
calls    ios_vm_vm_f         -> web_src_store_store_save   [INFERRED]
method   web_src_store_store -> ios_storeext_store_reset
contains ios_storeext        -> web_src_store_store
```

A Swift call resolves into a **TypeScript** method; the TS class owns a Swift method. Swift extends `Data`, `Date`, `Result`, `Store`, `Router` constantly, and those are common class names in TS/Python/Kotlin any `ios/` + `web/` monorepo hits this immediately. Reproduced with Python (`extension Data` onto `class Data`) and with a JS re-export symbol node.

Note the gate must permit Swift↔Objective-C folding, which is intended and covered by `test_swift_extension_folds_onto_objc_class`. `_LANG_FAMILY_BY_EXT`'s `"native"` family already expresses exactly this.

**2. The post-merge edge rebuild is global.** Once `remap` is non-empty, `_merge_swift_extensions` rebuilds the entire `all_edges` list and de-duplicates on `(source, target, relation, source_file, source_location)` a key that ignores `confidence`, `confidence_score`, `weight`, `context`. Measured on graphify's own `graphify/` (80 `.py` files) plus two trivial Swift files: **166 legitimate Python edges silently deleted**. The corpus genuinely contains 113 duplicate-key groups (e.g. `def g(a: Thing, b: Thing) -> Thing` emits three same-key `references` edges), and a Swift extension anywhere now prunes them.

## Fix

I have a patch and will open a PR against `v8` referencing this issue. It:

1. remaps `swift_extensions[].nid` in both the `id_remap` and `sym_remap` passes, alongside the existing `raw_calls[].caller_nid` rewrite (exact mapping, no heuristics);
2. restricts merge candidates to Swift's own `_LANG_FAMILY_BY_EXT` family and skips `_LANGUAGE_BUILTIN_GLOBALS`, preserving Swift↔ObjC folding;
3. adds the `id in contained` gate so sourceless stubs can't make a label ambiguous;
4. scopes the edge rebuild to edges the merge actually remapped.

Four tests in `tests/test_swift_cross_file_calls.py`: two fail without (1)+(3), two fail against a fix that omits (2)+(4). Full suite shows no regressions against baseline.

## Residual (separate, likely by design)

`_swift_receiver_name` is depth-1 (`engine.py:900`), so `Foo.shared.baz.bar()` yields no edge. Rare in practice; noting it only so it isn't conflated with the above.

pawelo446:
Fixes #2538.

## The bug

`swift_extensions[].nid` is recorded during per-file AST extraction, **before** the `id_remap` / `sym_remap` passes rewrite absolute-path ids to repo-relative ones. Those passes rewrite the sibling `raw_calls[].caller_nid` but not this field — `cache.py:656` already lists both as the id-carrying payload fields the cache path must remap; the in-process path covers only one.

So with **absolute** input paths every lookup in `_merge_swift_extensions` misses and the merge no-ops. `Foo` keeps two definition nodes, the single-definition guard in `_resolve_swift_member_calls` bails, and one `extension Foo {}` in its own file zeroes out `Foo`'s call graph — defeating #1533. With **relative** input paths `_file_stem` already yields the post-remap slug, so the merge works. That path-form dependency is why this survived: `extract(paths)` without `root=` never triggers the remap.

## The fix

1. **Remap `swift_extensions[].nid` in both passes**, alongside the existing `raw_calls[].caller_nid` rewrite. Exact mapping, no heuristics.

Remapping the nid activates a merge pass that has been largely dormant, and its matching rule is unsafe as written — so this PR also gates it. Both hazards are measured and detailed in #2538:

2. **Language gate.** `label_to_canonical` was filtered only by `_is_type_like_definition`, which accepts any `file_type == "code"`, with label equality as the only other condition. `extension Data` / `extension Store` — idiomatic Swift — would absorb a same-named TypeScript or Python class and invent cross-language edges (a Swift call resolving into a TS method, a TS class owning a Swift one). Candidates are now restricted to Swift's own `_LANG_FAMILY_BY_EXT` family and `_LANGUAGE_BUILTIN_GLOBALS` is skipped, mirroring #1726 / #2147. The `"native"` family keeps the intended Swift↔Objective-C folding (`test_swift_extension_folds_onto_objc_class` still passes).

3. **`contained` gate.** A bare `let x: Foo` mints a sourceless stub labelled `Foo`, which made the label look ambiguous and blocked the merge even with a correct nid. Every sibling pass pairs `_is_type_like_definition` with `id in contained` (L2322, 2568, 2674, 2804, 3093); this one now does too.

4. **Scoped edge rebuild.** Once `remap` was non-empty the pass rebuilt the *entire* `all_edges` list, deduping on a key that ignores `confidence` / `confidence_score` / `weight` / `context`. On graphify's own `graphify/` plus two trivial Swift files that silently deleted 166 legitimate Python edges. The rebuild now only rewrites and dedups edges it actually remapped; untouched edges pass through verbatim.

## Tests

Four added to `tests/test_swift_cross_file_calls.py`, all load-bearing:

| test | fails without |
|---|---|
| `test_cross_file_extension_does_not_erase_static_calls` | (1) |
| `test_type_annotation_stub_does_not_block_extension_merge` | (3) |
| `test_extension_does_not_merge_into_same_named_foreign_type` | (2) |
| `test_extension_merge_does_not_prune_unrelated_edges` | (4) |

The first two fail on `v8`; the last two pass on `v8` but fail against a fix that omits the gates — they exist to stop (2) and (4) being reintroduced. They pass `root=` explicitly, since that is what the CLI does and what makes the bug reproducible.

Note `test_swift_extension_across_files_merges_into_canonical_type` in `test_languages.py` passes on `v8` despite the bug: it omits `root=` and asserts only on `method` edges with two files, so it never observes the call-edge loss a third *calling* file exposes.

## Verification

Full suite run against a baseline taken in the same tree: **no regressions**. Remaining failures (`test_skillgen.py`, `test_ollama_retry_cap.py`, `test_labeling.py`) are identical before and after and unrelated to this change. `ruff check` clean on both files.

Measured on a private ~2,200-file Swift codebase, one singleton manager type: nodes for the type 10 → 1, `calls` edges into its methods 17 → 171, source files reached 10 → 101, `affected <Type> --depth 2` from `No unique node match` → 251 nodes. Top-degree ranking essentially unchanged, so no god-node inflation.


safishamsi:
Shipped in v0.9.36. Your fix (remap `swift_extensions[].nid` in the in-process id/sym passes so a cross-file extension merges onto its base type) is what landed, with a candidate gate so the merge never absorbs a same-named type from another language. Credited in the release notes. Thanks @pawelo446!

PR #2540, opened 2026-08-07
fix(dedup): prefer active path segments over archived in collision rank

### LINKED ISSUE #2532: _collision_rank falls through to lexical filename order
`graphify/dedup.py` `_collision_rank` falls through to **lexical filename order** when two nodes share an ID. `plans/_done/` sorts before `plans/in-progress/` in ASCII — so the archived copy wins **by alphabet**. That's why binding doctrine resolved to a dead plan.

michaelxer:
## Problem
When two nodes share an ID, `_collision_rank` broke ties with pure lexical `source_file` order. That makes `plans/_done/...` win over `plans/in-progress/...` because `_` sorts before `i` in ASCII, so an archived plan can beat the live one.

Fixes #2532

## Change
- Add a small lifecycle penalty on path segments before the existing label/path lexical tiebreaks
- Active markers (`in-progress`, `wip`, `active`, ...) rank better than archived markers (`_done`, `archive`, `deprecated`, ...)
- Paths with no lifecycle marker stay neutral
- Definer-vs-reference ranking is unchanged

## Verification
Reproduced the reported case before the change: both insertion orders kept `plans/_done/binding-doctrine.md`.

After the change, both orders keep `plans/in-progress/binding-doctrine.md`.

```bash
uv run pytest tests/test_dedup.py -q
# 68 passed
```

Focused new coverage:
- `test_lifecycle_penalty_prefers_active_over_archived`
- `test_archived_path_does_not_beat_active_path_by_ascii_order`
- `test_active_plan_survives_over_archived_copy_order_independent`

## Notes
This only affects same-ID collision ranking. It does not change fuzzy/label dedup. Lifecycle markers are matched as whole path segments only, so ordinary filenames like `done.md` are not treated as archive markers.

### DIFF
--- a/graphify/dedup.py
+++ b/graphify/dedup.py
@@ -225,22 +225,69 @@ def _defines_id(node: dict) -> bool:
                for prefix in _id_prefixes(source_file))
 
 
+# Path-segment lifecycle markers used by _collision_rank. Lower penalty wins.
+# Without this, pure lexical source_file order makes ``plans/_done/…`` beat
+# ``plans/in-progress/…`` because "_" < "i" in ASCII (#2532).
+_ACTIVE_PATH_SEGMENTS = frozenset({
+    "in-progress",
+    "in_progress",
+    "wip",
+    "active",
+    "current",
+    "ongoing",
+})
+_ARCHIVED_PATH_SEGMENTS = frozenset({
+    "_done",
+    "done",
+    "archive",
+    "archived",
+    "_archive",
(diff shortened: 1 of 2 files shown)


michaelxer:
Closing this one — looks like #2532 already landed on v8 via the deterministic collision-rank fix (active over archived). The branch is conflicting with that change and there's nothing left for this PR to do.

Thanks for the quick turnaround on the issue.

Issue #2324, opened 2026-07-30
Absolute-path node IDs with no canonical twin in SQL `references` and JS `calls` edges (0.9.28) — not covered by the #1961/#1967 fix


Kenta-Nakakoshi:
## Summary

On graphifyy **0.9.28**, the AST extractor emits `references` and `calls` edges whose **target** ID carries the absolute scan path, while the **source** ID from the same file in the same run is correctly repo-relative. The target never matches a node and the edge is dropped at build.

This looks like #1961 (fixed in 0.9.18 via #1967) but is **not covered by that fix**. #1961 was a *duplicate* edge: the absolute-path variant always had a canonical repo-relative twin, and the fix drops the duplicate so the correct edge survives. In my corpus there is **no twin** — I checked all 224 affected refs and **0** resolve after stripping the prefix, because the repo-relative node was never created either. So the relationship isn't doubled, it's **lost**.

It also isn't limited to JS/TS tsconfig-alias imports: the clearest case here is a **SQL foreign key**.

*(Paths redacted to `users_<username>_...`; real IDs embed the absolute home directory.)*

## Environment

- `graphifyy` 0.9.28, Python 3.13, macOS 15, `tree_sitter_sql` installed
- Corpus: 187 code files (52 `.tsx`, 39 `.js`, 14 `.ts`, 11 `.sql`)
- AST extraction: 1330 nodes / 3376 edges → **615 dangling**, of which **224 (184 distinct IDs)** are this absolute-path shape

## Reproduction — SQL foreign key, both endpoints in one file

`prisma/migrations/20260723140000_kc_stock_gap_p0/migration.sql` declares `Tenant` and `StockCountSession` and adds an FK between them. The emitted edge:

```
relation        : references
source          : prisma_migrations_20260723140000_kc_stock_gap_p0_migration_stockcountsession   [node exists: True]
target          : users_<username>_yys_..._prisma_migrations_20260723140000_kc_stock_gap_p0_migration_tenant   [node exists: False]
source_file     : prisma/migrations/20260723140000_kc_stock_gap_p0/migration.sql
source_location : L92
confidence      : EXTRACTED
```

`Tenant` **is** a node — as `prisma_migrations_20260723140000_kc_stock_gap_p0_migration_tenant`. But no edge is emitted to that ID; the absolute-path one is the only edge for this FK. Same pattern on L95, L101, L104 (`stockgapevent → tenant`, `→ user`).

[...]

PR #2541, opened 2026-08-07
feat(yaml): extract Docker Compose services and GitHub Actions jobs

bruno-growthsales:
## Problem

`.yaml`/`.yml` have no AST extractor, so infrastructure-as-config contributes nothing structural to the graph. On a repo whose deploy path is `docker-build.yml` → Portainer/Swarm, the deploy topology was the one part of the architecture the graph could not answer questions about — the files are read by the semantic pass, but the service and job dependencies never become edges.

## What this adds

`extract_yaml`, behind an optional `[yaml]` extra (`tree-sitter-yaml`), mirroring the `sql` / `terraform` pattern:

- **Docker Compose** — service nodes; `depends_on` edges for both the list form and the long-form mapping (`{db: {condition: ...}}`); `extends` → base service.
- **GitHub Actions** — job nodes; `needs` → `depends_on` (scalar and list forms); step-level and job-level `uses` → the action or reusable workflow.
- **Data YAML** (k8s manifests, OpenAPI specs, fixtures) returns an empty result and is left to the semantic pass, the same way `_is_config_json` skips data JSON (#1224). A 1 MiB ceiling with the same bounded read keeps multi-MB lockfiles from being parsed for nothing.

On a real monorepo: `docker-compose.yml` → 5 services and 4 `depends_on` edges; `ci.yml` → 12 jobs, the `build → lint-and-typecheck` dependency, and 22 `uses` edges.

## Two design decisions worth reviewing

**1. The extensions stay in `DOC_EXTENSIONS`.** `collect_files` keys off `_DISPATCH`, not `CODE_EXTENSIONS`, so registering the extractor is enough to get the structural pass *without* removing YAML from the semantic one — the same arrangement `.md` already has. It also avoids the #1689 "classified as code but no extractor" warning that moving the extensions would otherwise require handling.

**2. External `uses` targets are marked `type=module`.** Definitions are file-scoped with a `contains` edge; cross-file references are sourceless stubs so `_rewire_unique_stub_nodes` can collapse a Compose overlay's `depends_on` onto the base file's service (#2324, #1402). But an `actions/checkout@v4` pinned by ten workflows is never *defined* anywhere in the corpus, so there is nothing to rewire onto, and `_disambiguate_colliding_node_ids` salts each stub with its own path — scattering one shared action into ten nodes. Marking them `type=module` takes the module-anchor exemption (#1327), whose rationale reads exactly like this case: "those are the *same* module, not distinct same-named symbols". Happy to take a different approach here if you would rather not overload `type`.

## Tests

`tests/test_yaml_config.py` — 19 tests covering both shapes, the sequence-marker trap (`- api` vs `api`), forward references binding locally, data YAML staying empty, and the two cross-file cases (overlay rewire, shared-action merge) exercised through `extract()` and `build_from_json`.

Full suite: **4053 passed, 0 failed**. `ruff check` clean.

## Note on `uv.lock`

CI runs `uv sync --frozen`, so the lock had to be regenerated for the new extra. Regenerating also picked up a `0.9.31 → 0.9.35` project-version correction the lock had drifted from — unrelated to this change, but not separable from it.

https://claude.ai/code/session_013EeP6fFjQkaE2CxNz3RTnY

### DIFF
--- a/graphify/extract.py
+++ b/graphify/extract.py
@@ -54,6 +54,7 @@ from graphify.extractors.sln import extract_sln  # noqa: F401
 from graphify.extractors.sql import extract_sql  # noqa: F401
 from graphify.extractors.terraform import extract_terraform  # noqa: F401
 from graphify.extractors.verilog import extract_verilog  # noqa: F401
+from graphify.extractors.yaml_config import extract_yaml  # noqa: F401
 from graphify.extractors.zig import extract_zig  # noqa: F401
 from graphify.security import sanitize_metadata
 from graphify.paths import disambiguate_ambiguous_candidates
@@ -4278,6 +4279,8 @@ _DISPATCH: dict[str, Any] = {
     ".sh": extract_bash,
     ".bash": extract_bash,
     ".json": extract_json,
+    ".yaml": extract_yaml,
+    ".yml": extract_yaml,
     ".tf": extract_terraform,
     ".tfvars": extract_terraform,
     ".hcl": extract_terraform,
@@ -4305,6 +4308,8 @@ _DISPATCH: dict[str, Any] = {
 # extract() to tell the user which extra restores the language.
 _EXTRA_FOR_EXTENSION = {
     ".sql": "sql",
+    ".yaml": "yaml",
+    ".yml": "yaml",
     ".tf": "terraform",
     ".tfvars": "terraform",
     ".hcl": "terraform",
--- a/graphify/extractors/yaml_config.py
+++ b/graphify/extractors/yaml_config.py
@@ -1,0 +1,320 @@
+"""Yaml_config extractor. Docker Compose and GitHub Actions workflows."""
+from __future__ import annotations
+
+
+from pathlib import Path
+from graphify.extractors.base import _file_stem, _make_id
+
+
+# Filenames that are Docker Compose files by convention. Matched before the
+# top-level key probe below, the same cheap-first order json_config uses.
+_COMPOSE_PREFIXES = ("docker-compose", "compose")
+
+# Step keys that carry a reference to another workflow/action rather than a
+# shell command. `uses` is the only one today, but keeping the set makes the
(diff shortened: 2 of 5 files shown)


bruno-growthsales:
Disclosing a limitation I hit while running this on a real monorepo, since a reviewer would reasonably ask about it.

**The extractor is not reachable from `graphify extract`.** That command builds its AST list from `detect`'s `code` category:

```python
code_files = [Path(p) for p in files_by_type.get("code", [])]
```

`.yaml`/`.yml` are `DOC_EXTENSIONS`, so on a fresh build they land in `doc_files` and — with `--code-only` — are skipped outright. The structural pass only sees them through `collect_files`, which keys off `_DISPATCH` and is used by `update` / `watch`.

In practice the build is two steps:

```
graphify extract . --code-only --out .   # no YAML
graphify update .                        # YAML appears: 43 nodes, 21 depends_on, 42 uses
```

This is pre-existing and not introduced here — `.md` has exactly the same shape today, with `extract_markdown` registered in `_DISPATCH` but unreachable from `extract`. I left it that way rather than widening the scope of this PR.

Two ways forward, whichever you prefer:

1. **Merge as-is.** YAML lands via `update`, which is the documented steady-state command anyway.
2. **Follow-up PR** widening `extract`'s AST list to any extension with a registered extractor, instead of `detect['files']['code']`. Strictly additive — nothing currently extracted stops being extracted, and `.md` gains structural coverage in the initial build too. Happy to send it separately if you want it.

What I'd argue against is moving `.yaml`/`.yml` into `CODE_EXTENSIONS`: it would close the gap, but data YAML (k8s manifests, OpenAPI specs) would leave the semantic pass, and since `extract_yaml` deliberately returns empty for those, they would contribute nothing at all. That trades a gap for a regression.

Separately, while testing this I ran into an unrelated bug that made the results confusing at first — filed as #2543.

PR #2544, opened 2026-08-07
fix(detect): normalize Unicode so accented ignore patterns match on macOS

bruno-growthsales:
## The bug

An ignore rule naming a directory with an accent silently does nothing on macOS. The files it was meant to exclude get scanned anyway — and for docs/PDFs, queued for semantic extraction.

macOS (APFS/HFS+) returns filenames in **NFD**: `ç` comes back as `c` + U+0327 COMBINING CEDILLA. Editors write ignore files in **NFC**, where the same `ç` is the single codepoint U+00E7. They render identically and compare unequal, so `fnmatch` never matches.

```python
>>> import unicodedata, fnmatch
>>> nfc = "Orçamento"                              # what your editor writes
>>> nfd = unicodedata.normalize("NFD", nfc)        # what macOS returns
>>> nfc.encode().hex(), nfd.encode().hex()
('4f72c3a7616d656e746f', '4f7263cca7616d656e746f')
>>> fnmatch.fnmatch(f"{nfd}/contrato.pdf", f"{nfc}/*")
False
```

## Why it matters

Found in a real repo. A `.graphifyignore` containing `Orçamento/` failed to exclude that directory, and **9 client contract PDFs were picked up for semantic extraction** — queued to be sent to an LLM — despite an explicit rule against it.

The failure is silent. No warning, no error; the only symptom is a file count that doesn't match expectation. I caught it because the scan reported `9 papers` when the ignore rules should have left zero, and I went looking. A user who doesn't count would never know.

That's what makes this worth fixing rather than documenting: **the rule appears to work.** It's also not exotic — any accented path hits it, which is routine in Portuguese, Spanish, and French repos.

## The fix

Normalize both sides to NFC before matching, at three boundaries:

- `_parse_gitignore_line` — the pattern side, so it covers `.graphifyignore`, `.gitignore` and `$GIT_DIR/info/exclude` alike
- `_is_ignored` — `target.name` and the anchor-relative path

NFC is already the form Linux and Windows produce, so this is a no-op there and only repairs the macOS mismatch.

## Tests

Three added in `tests/test_detect.py`:

| Test | Before | After |
|---|---|---|
| NFC pattern vs NFD path on disk | ❌ fails | ✅ passes |
| NFD pattern vs NFC path (reverse) | ❌ fails | ✅ passes |
| ASCII patterns unaffected | ✅ passes | ✅ passes |

The two regression tests fail on `v8` without the patch and pass with it. The third guards against the normalization changing existing behavior.

**Full suite: 3833 passed.** The 13 failures in `tests/test_terraform.py` are pre-existing on a clean `v8` checkout (optional `tree_sitter_hcl` not installed) and unrelated — verified by stashing this change and re-running.

## Note

Separate from and independent of #2541.

### DIFF
--- a/graphify/detect.py
+++ b/graphify/detect.py
@@ -5,6 +5,7 @@ import json
 import os
 import re
 import shlex
+import unicodedata
 from concurrent.futures import ThreadPoolExecutor
 from enum import Enum
 from functools import lru_cache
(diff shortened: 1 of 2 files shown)


safishamsi:
Fixed in v0.9.40 (`graphifyy==0.9.40` on PyPI). `.graphifyignore`/gitignore pattern matching now normalizes Unicode (NFC) on both the pattern and the path, so an accented rule matches a differently-normalized path on macOS. Thanks @bruno-growthsales (PR #2544).

PR #2546, opened 2026-08-08
fix(manifest): do not stamp failed-extra AST files as up to date

### LINKED ISSUE #2543: Incremental manifest never retries files whose extractor failed on a missing extra
## Summary

When an extractor hard-fails because its optional extra is not installed, the affected files are stamped into the incremental manifest as processed. Installing the extra and re-running does **not** repair the graph — the incremental gate reports the files as unchanged and skips them forever. The only recovery is deleting `graphify-out/` entirely.

This is not specific to one language: it applies to every extra whose extractor returns a `"not installed"` error — `[sql]`, `[terraform]`, `[dm]`, and any future one.

## Reproduction

On a repo with 227 `.sql` files, `graphifyy` installed **without** the `[sql]` extra:

```
$ graphify extract . --code-only --out .
  warning: 227 .sql file(s) contributed nothing to the graph because a dependency is
  missing: tree_sitter_sql not installed. Install it with: pip install "graphifyy[sql]" (#1745)
[graphify extract] wrote graph.json: 7912 nodes, 19212 edges
```

Install the missing extra, then re-run:

```
$ uv tool install --force "graphifyy[sql] @ ..."
$ graphify extract . --code-only --out .
[graphify extract] 0 code, 0 docs, 0 papers, 0 images changed; 1228 unchanged; 0 deleted
[graphify extract] wrote graph.json: 7912 nodes, 19212 edges
[graphify extract] incremental summary: 1228 files cached/unchanged, 0 re-extracted, 0 deleted
```

Nothing is re-extracted and the warning is gone, so the run looks clean. The 227 SQL files are still absent from the graph.

Only after `rm -rf graphify-out` does the graph repair:

```
[graphify extract] wrote graph.json: 8675 nodes, 19820 edges
```

**7912 → 8675 nodes.** The intermediate state is the dangerous one: the second run gives every signal of success while the graph stays silently incomplete.

## Why it happens

`_manifest_files` is derived from `files_by_type` — the full detect output — so a file gets its hash stamped regardless of whether its AST extraction returned an error. The next run's incremental gate compares hashes, sees no change, and skips.

## Precedent for the fix

The codebase already applies this principle in the two neighbouring layers, just not here:

- **#1666**, AST cache: "Never cache a zero-node result for an extractable file … Caching it makes the empty byte-stable across runs and silently blinds affected/explain to and through the file."
- **#1948**, semantic manifest: files dispatched but dropped have their stale `semantic_hash` explicitly cleared, because `save_manifest`'s seed loop "would otherwise copy it verbatim and mask the omission."

The AST side has no equivalent. A file whose extraction result carries an `error` (or that returned zero nodes while having a registered extractor) should not be stamped as up to date, so the next run retries it — exactly the shape #1948 already implements one pass over.

## Note

The `#1745` warning that surfaces the original failure goes to **stderr**, so in a piped or CI context it is easy to miss the one signal that something went wrong. Not a bug, but it compounds this one.

Happy to send a PR if you agree with the approach.

michaelxer:
## Problem

When an AST extractor hard-fails because its optional extra is not installed (for example `tree-sitter-sql` for `.sql`), the affected files were still stamped into the incremental manifest as processed. After installing the extra, re-running extract reported them as unchanged and skipped them forever. The only recovery was deleting `graphify-out/`.

This matches #2543. The semantic side already had the same principle via #933 / #1948; the AST side did not.

## Solution

- `extract()` now returns `failed_sources` for results that carry an error (including missing extras) and for extractor-present zero-node anomalies.
- `_stamped_manifest_files()` drops those sources so they are not written as up to date.
- `save_manifest(..., clear_ast=...)` blanks both `ast_hash` and `semantic_hash` on any prior seed row for those sources, so the next incremental run re-queues them.

## Verification

```bash
pytest tests/test_extract.py::test_extract_warns_when_sql_extra_missing \
  tests/test_extract.py::test_extract_failed_sources_empty_when_sql_installed \
  tests/test_partial_cache.py::test_stamped_manifest_excludes_failed_ast_sources \
  tests/test_partial_cache.py::test_stamped_manifest_excludes_partial_files \
  tests/test_detect.py::test_save_manifest_clear_ast_blanks_both_hashes_for_failed_extra \
  tests/test_detect.py::test_save_manifest_clear_semantic_erases_stale_hash_for_omitted_file -q
```

Result: 5 passed, 1 skipped (`tree_sitter_sql` not installed in this environment for the success-path skip).

Local repro without the sql extra:

- first extract surfaces `failed_sources` for `.sql`
- stamped manifest keeps only successful code
- second `detect_incremental` still re-queues the `.sql` file

## Notes / limitations

- I did not run a full end-to-end install of `graphifyy[sql]` in this environment (optional extra not present here). The missing-extra failure path and manifest re-queue behavior are covered by unit/integration tests and a local repro script.
- Draft until CI is green.

Fixes #2543

### DIFF
--- a/graphify/cli.py
+++ b/graphify/cli.py
@@ -90,6 +90,7 @@ def _stamped_manifest_files(
     sem_result: dict,
     root: Path,
     partial_source_files: "set[str] | None" = None,
+    failed_ast_sources: "set[str] | list[str] | None" = None,
 ) -> dict[str, list[str]]:
     """Manifest-safe files dict: only stamp semantic files that actually
     produced output (cache hit or fresh extraction). Files whose chunk failed
@@ -115,6 +116,10 @@ def _stamped_manifest_files(
     doc unstamped, so detect_incremental re-queued it on every run. The stamping
     condition mirrors the cache-write keying (a hyperedge carries its own
     ``source_file``); do not derive it from member nodes.
+
+    ``failed_ast_sources`` (#2543): code files whose AST extractor errored
+    (missing optional extra, etc.) or returned zero nodes. They must not be
+    stamped as up-to-date or a later install of the extra will never re-run.
     """
     root = Path(root)
 
@@ -134,12 +139,16 @@ def _stamped_manifest_files(
             if sf:
                 sem_extracted.add(_resolve(sf))
     partial_resolved = {_resolve(p) for p in (partial_source_files or set())}
+    failed_ast_resolved = {_resolve(p) for p in (failed_ast_sources or [])}
     sem_types = {"document", "paper", "image"}
     return {
         ftype: [
             f for f in flist
-            if ftype not in sem_types
-            or (_resolve(f) in sem_extracted and _resolve(f) not in partial_resolved)
+            if _resolve(f) not in failed_ast_resolved
+            and (
+                ftype not in sem_types
+                or (_resolve(f) in sem_extracted and _resolve(f) not in partial_resolved)
+            )
         ]
         for ftype, flist in files_by_type.items()
     }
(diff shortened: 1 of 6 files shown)


safishamsi:
Shipped in v0.9.37. Your fix (do not stamp failed-extra AST files as up-to-date) is what landed for #2543, plus a one-time heal for manifests poisoned by the old behavior and the same treatment on the watch save_manifest paths. Credited in the release notes. Thanks @michaelxer!

PR #2555, opened 2026-08-08
fix(llm): surface claude-cli envelope errors on both exit paths (#2554)

### LINKED ISSUE #2554: claude-cli: API errors in the stdout envelope go unreported, so a rate limit prints a blank `exited 1:`, keeps retrying against the outage, and can land in graph.json as a community label
## Problem

The `claude` CLI reports API-level failures (rate limits, auth) in the **stdout JSON envelope** with `is_error: true`, leaving **stderr empty**. Depending on the failure it exits either non-zero or **zero**.

Both `claude -p` call sites in `graphify/llm.py` decide success from `proc.returncode` and read the cause only from `proc.stderr`, so three distinct failures go unreported.

Direct observation of what the CLI actually returns on a rate limit:

```console
$ echo "say OK" | claude -p --output-format json
{"type":"result","subtype":"success","is_error":true,"duration_ms":496,
 "result":"API Error: Rate limit reached","stop_reason":"stop_sequence", ...}
$ echo $?
0
```

Note `is_error: true` alongside `subtype: "success"` and exit code `0`.

### Case A: non-zero exit, cause discarded

`graphify/llm.py:1541` (extraction) and `graphify/llm.py:2600` (`_call_llm`) build the message from `proc.stderr` alone. stderr is empty, so the user gets a colon with nothing after it:

```
[graphify] chunk 1/11 failed: claude -p exited 1:
[graphify] chunk 2/11 failed: claude -p exited 1:
...
[graphify] chunk 11/11 failed: claude -p exited 1:
[graphify extract] error: all semantic chunks failed for backend 'claude-cli' (309 uncached files)
```

Eleven identical blank errors. Nothing indicates the account was rate limited rather than, say, the CLI being misconfigured.

### Case B: zero exit with `is_error: true` burns retries during the outage

When the CLI exits 0, the `returncode` check passes and `envelope["result"]` (the string `"API Error: Rate limit reached"`) is handed to `_parse_llm_json` as if it were model output. It parses to nothing, so `_response_is_hollow` at `graphify/llm.py:1571` reads it as truncation and forces `finish_reason = "length"`:

```
[graphify] LLM returned invalid JSON, skipping chunk (first 200 chars: 'API Error: Rate limit reached')
```

Adaptive retry then bisects the chunk and re-issues requests that are still being refused. A rate limit is the one condition where retrying cannot help, and it is exactly the condition that triggers the retry path.

### Case C: the error string is written into the graph (data corruption)

`_call_llm` ends with `return envelope.get("result", "")` (`graphify/llm.py:2611`). Its callers are the dedup LLM tiebreaker (`graphify/dedup.py:821`) and community labeling (`graphify/llm.py:2952`).

Neither validates the reply, so during a rate limit a community label can become the literal string `API Error: Rate limit reached` and be persisted into `graph.json`. Unlike A and B this leaves bad data behind after the outage ends.

## Reproduction

1. Exhaust the Claude subscription rate limit for the account the `claude` CLI is authenticated as.
2. `graphify extract <corpus> --backend claude-cli`

Observed: eleven `claude -p exited 1:` lines with no cause (Case A). With the same account in the zero-exit variant, the hollow-response retry path engages instead (Case B).

## Proposed fix

I could not find an existing issue covering this. There is an unmerged PR that partially addresses Case A; Cases B and C both occur on a **zero** exit code and are outside its scope.

Read the envelope's `is_error` flag rather than inferring failure from the exit code alone:

- A small helper parses stdout and returns the envelope's `result` text when `is_error` is set, and an empty string otherwise (including when stdout will not parse, so the existing "unparseable JSON envelope" error still wins).
- On a non-zero exit, prefer `stderr` and fall back to the envelope text. stderr should keep priority so a genuine CLI crash or spawn failure still reports its own diagnostics.
- On a zero exit with the flag set, raise instead of treating the error text as model output. This fixes B and C together, since both stem from that string being passed downstream.

Failing fast on a rate limit is the intended behaviour: `extract_corpus_parallel` already catches per-chunk exceptions and continues, so each chunk reports the real cause instead of silently bisecting.

I have this implemented with tests and can open a PR if the approach looks right.

## Environment

- graphify 0.9.36 (`v8`)
- Claude Code CLI, `-p --output-format json`
- macOS 15 (Darwin 25.5.0), Python 3.10

annieyii:
Fixes #2554

## Problem

`claude -p` reports API failures in the stdout JSON envelope (`is_error: true`) with an empty stderr, and on a rate limit it may exit **0**. Both call sites in `llm.py` decide success from `returncode` and read the cause from `stderr` only, so three failures go unreported:

- **A.** Non-zero exit prints `claude -p exited 1:` with nothing after the colon.
- **B.** Zero exit feeds `"API Error: Rate limit reached"` to `_parse_llm_json`; the empty parse reads as truncation and adaptive retry bisects the chunk, re-issuing requests that are still being refused.
- **C.** `_call_llm` returns that same string as the model's reply, so community labeling and the dedup tiebreaker can write `API Error: Rate limit reached` into `graph.json` as a label.

## Change

`_claude_cli_error(stdout)` returns the envelope's `result` text when `is_error` is set, and `""` otherwise, including when stdout will not parse, so the existing "unparseable JSON envelope" error still takes precedence.

Both call sites (`_call_claude_cli`, `_call_llm`) then:

- prefer `stderr` on a non-zero exit and fall back to the envelope text, so a genuine CLI crash or spawn failure still reports its own diagnostics;
- raise on a zero exit when the flag is set, which fixes B and C together since both stem from that string flowing downstream.

Failing fast is intended here: `extract_corpus_parallel` already catches per-chunk exceptions and continues, so each chunk now reports the real cause instead of silently bisecting during an outage.

## Relationship to #1864

#1864 found the same root cause and fixes **A** via `proc.stdout.strip()[-500:]`. It has not landed; `llm.py:1541` on `v8` still reads stderr only.

This PR is a superset: it also covers B and C, which occur on a **zero** exit that #1864's non-zero branch never reaches, and it surfaces the envelope's `result` field rather than a 500-character tail of raw JSON. Happy to rebase onto #1864 instead if the author would rather extend that PR.

## Verification

Four tests added to `tests/test_claude_cli_backend.py`, following the file's existing `MagicMock` + `patch("subprocess.run")` convention, so CI needs no `claude` binary:

- `test_nonzero_exit_surfaces_envelope_error_when_stderr_empty`
- `test_raises_on_error_envelope_with_zero_exit`
- `test_call_llm_raises_on_error_envelope`
- `test_call_llm_nonzero_exit_surfaces_envelope_error`

Each was written and observed failing before the implementation.

```
tests/test_claude_cli_backend.py tests/test_llm_backends.py
tests/test_llm_parser.py tests/test_backend_extras.py     136 passed

full suite                                               4041 passed, 36 skipped
```

Four unrelated failures in `tests/test_ollama_retry_cap.py` are pre-existing on my machine and come from the optional `openai` package not being installed (`ModuleNotFoundError: No module named 'openai'` raised by `monkeypatch`), not from this change.

Against a live rate-limited account:

```
before  [graphify] chunk 1/11 failed: claude -p exited 1:
after   [graphify] chunk 1/11 failed: claude -p exited 1: API Error: Rate limit reached
```

### DIFF
--- a/graphify/llm.py
+++ b/graphify/llm.py
@@ -1373,6 +1373,30 @@ def _claude_cli_envelope(stdout: str) -> dict:
     return envelope
 
 
+def _claude_cli_error(stdout: str) -> str:
+    """Return the CLI's own error text when the envelope flags `is_error`.
+
+    `claude -p` reports API failures (rate limits, auth) in the stdout JSON
+    envelope with `is_error: true` and leaves stderr EMPTY — and on a rate limit
+    it still exits 0. So the two obvious checks both miss it: a non-zero exit
+    printed a bare "exited 1: " with no cause, and a zero exit fed the error
+    string to the JSON parser, producing an empty graph that `_response_is_hollow`
+    misread as truncation and adaptive retry then bisected, re-issuing requests
+    that were still being refused (#2554). Best-effort: unparseable stdout is not
+    this function's problem, the caller's `_claude_cli_envelope` reports that.
+    """
+    try:
+        envelope = _claude_cli_envelope(stdout)
+    except RuntimeError:
+        return ""
+    if not envelope.get("is_error"):
+        return ""
+    detail = envelope.get("result")
+    if isinstance(detail, str) and detail.strip():
+        return detail.strip()
+    return "unspecified error"
+
+
 # A JSON Schema pinning the top-level shape graphify consumes. Passed to
 # `claude -p --json-schema` (structured output) so the CLI CONSTRAINS the model
 # to emit the object directly instead of relying on it CHOOSING to honour a
@@ -1538,10 +1562,12 @@ def _call_claude_cli(user_message: str, max_tokens: int = 8192, *, deep_mode: bo
         check=False,
         **_no_window_kwargs(),
     )
+    cli_error = _claude_cli_error(proc.stdout)
     if proc.returncode != 0:
-        raise RuntimeError(
-            f"claude -p exited {proc.returncode}: {proc.stderr.strip()[:500]}"
-        )
+        detail = proc.stderr.strip() or cli_error or "(no stderr, no error envelope)"
+        raise RuntimeError(f"claude -p exited {proc.returncode}: {detail[:500]}")
+    if cli_error:
+        raise RuntimeError(f"claude -p reported an error: {cli_error[:500]}")
 
     envelope = _claude_cli_envelope(proc.stdout)
 
@@ -2597,8 +2623,14 @@ def _call_llm(
             check=False,
             **_no_window_kwargs(),
         )
+        cli_error = _claude_cli_error(proc.stdout)
         if proc.returncode != 0:
-            raise RuntimeError(f"claude -p exited {proc.returncode}: {proc.stderr.strip()[:500]}")
+            detail = proc.stderr.strip() or cli_error or "(no stderr, no error envelope)"
+            raise RuntimeError(f"claude -p exited {proc.returncode}: {detail[:500]}")
+        if cli_error:
+            # Without this the error text is returned as the model's reply and
+            # the caller writes it into the graph as a community label (#2554).
+            raise RuntimeError(f"claude -p reported an error: {cli_error[:500]}")
         envelope = _claude_cli_envelope(proc.stdout)
         cli_usage = envelope.get("usage") or {}
         if cli_usage:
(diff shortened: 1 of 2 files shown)


safishamsi:
Shipped in v0.9.37. Adopted essentially verbatim: `_claude_cli_error` surfaces an `is_error` envelope on both exit paths (#2554). Credited in the release notes. Thanks @annieyii!

Issue #2568, opened 2026-08-09
TS/JS: #2552's union merges sibling closures' locals, so a multi-closure `const` initializer loses the very edges it recovers (v0.9.37)


imagineers-tyler:
## Summary

The fix for #2552 (shipped in v0.9.37) registers **every** topmost closure in a `const`
initializer under one `const_nid`, and **unions** their local-name sets into the single
`local_bound_names[const_nid]` entry. Each closure is its own scope, so when an initializer
holds more than one closure, one closure's parameters shadow a *sibling* closure's body — and a
genuine by-name reference in that sibling is suppressed.

Net effect: the edges #2552 set out to recover are lost again whenever the initializer contains
more than one closure.

This is **not** a regression against v0.9.36 — before #2552 those bodies were not walked at all,
so no edge existed either way. The single-closure case #2552 targets does work. It is the
multi-closure case that its own union defeats.

## Reproduction

Verified against pristine `graphifyy==0.9.37` from PyPI.

```js
// src/a.js
function alpha(x){ return x; }
function wrapper(a, b){ return a || b; }

export const handler = wrapper(
  (alpha) => { return alpha; },        // closure 1 — `alpha` is this closure's PARAMETER
  (pool)  => { pool.submit(alpha); }   // closure 2 — `alpha` here is the module function
);
```

```python
from graphify.extract import extract
from pathlib import Path
r = extract([Path('src/a.js')], cache_root=Path('.c'), parallel=False)
print([e for e in r['edges'] if e['relation'] == 'indirect_call'])
```

| closure 1's parameter name | `indirect_call` for `alpha` |
|---|---|
| `alpha` (collides with the module function) | **none — edge lost** |
| `beta` (no collision) | emitted, as expected |

Only the parameter name of an *unrelated sibling closure* differs between those two runs, which
isolates the union as the cause. Closure 1's parameter is not in scope inside closure 2.

## Cause

`graphify/extractors/engine.py`, the #2552 block:

```python
closures: list = []
_js_topmost_closures(inner, closures)
for closure in closures:
    if local_bound_names is not None:
        local_bound_names[const_nid] = (
            local_bound_names.get(const_nid, set())
            | _js_local_bound_names(closure, source))
    body = closure.child_by_field_name("body")
    if body:
        function_bodies.append((const_nid, body))
```

`function_bodies` gets **one entry per closure**, but `local_bound_names` gets **one entry
[...]

PR #2569, opened 2026-08-09
fix(extract): shadow a single unparenthesised arrow parameter from indirect_call args

imagineers-tyler:
## Summary

`_js_local_bound_names` reads only the `parameters` field. tree-sitter gives an arrow with
**one unparenthesised parameter** a `parameter` field (singular) and no `parameters` list node
at all, so that parameter never entered the shadow set: `x => sink(x)` bound nothing, and `x`
read as an unresolved by-name reference, resolved against the corpus-wide label index, and
fabricated an `indirect_call` edge (INFERRED, 0.8) to an unrelated same-named callable.

This is the same singular/plural trap as `catch_clause.parameter` (#2517). The parenthesised
form was always handled, which is what makes it easy to miss — `(x) => …` and `x => …` behaved
differently.

Minified bundles name nearly every private function with a single letter and use this arrow
form heavily, so the two collide constantly.

## Reproduction

On `v8` @ `09a34ad` (v0.9.37):

```js
// src/vendor.min.js
```

```js
// src/a.js
function sink(f){ return f; }
export const run = k => sink(k);      // `k` is this arrow's parameter
```

```python
from pathlib import Path
r = extract([Path('src/a.js'), Path('src/vendor.min.js')], cache_root=Path('.c'), parallel=False)
```

| form | `v8` | this PR |
|---|---|---|
| `k => sink(k)` | **fabricates** `run -> vendor_min_k` | none |
| `async k => sink(k)` | **fabricates** | none |
| `(k) => sink(k)` (control) | none | none |

Grammar check, so the field names are not taken on trust:

```
const f = x => g(x);         -> fields present: ['parameter']
const f = async x => g(x);   -> fields present: ['parameter']
const f = (x) => g(x);       -> fields present: ['parameters']
const f = ({a}) => g(a);     -> fields present: ['parameters']
```

## Change

Seven lines in `_js_local_bound_names`, next to the existing `parameters` read:

```python
    # An arrow with ONE unparenthesised parameter exposes it as `parameter`
    # (singular) — there is no `parameters` list node — so `x => f(x)` bound
    # nothing at all and `x` read as a by-name reference to any same-named
    # callable in the corpus. Same singular/plural trap as `catch_clause`.
    solo = func_node.child_by_field_name("parameter")
    if solo is not None:
        _js_collect_pattern_idents(solo, source, bound)
```

`_js_collect_pattern_idents` already handles the pattern forms, and `async x => …` is the same
node with the same field, so both are covered by the one read.

Scope note: the parameter is scoped to its arrow, and `_js_local_bound_names` is called per
function, so this does not widen anything beyond that arrow — a same-named module callable
referenced from a different function still resolves. There is a test for exactly that.

## Tests

New `tests/test_indirect_call_arrow_single_param_shadow.py`, modelled on
`test_indirect_call_nested_closure_shadow.py`:

| test | pins |
|---|---|
| `test_single_unparenthesised_arrow_param_emits_no_indirect_call` | the reported shape |
| `test_parenthesised_arrow_param_still_shadows` | control — the `parameters` path stays correct |
| `test_async_single_param_arrow_shadows` | `async x => …`, same singular field |
| `test_arrow_param_does_not_shadow_a_genuine_reference` | a same-named module callable used in **another** function still resolves |
| `test_genuine_reference_inside_the_arrow_still_emits` | no blanket suppression inside arrows |

Against the same test file:

| engine | result |
|---|---|
| `v8` @ `09a34ad` | **2 failed**, 3 passed — the two singular-parameter cases, i.e. the bug |
| this PR | **5 passed** |

Regression: `227 passed, 2 skipped` across `test_indirect_call_*`, `test_indirect_dispatch*`,
`test_extract`, `test_cross_language_call_resolution`, `test_node_id_canonical`.

## Effect on real code

Strictly subtractive — it removes fabricated edges and adds none:

| corpus | `v8` | this PR | delta |
|---|---|---|---|
| `node_modules`, 5,402 JS/TS files | 2,572 `indirect_call` | 2,484 | **+0 / -88** |
| 6,000-file mixed project tree | 3,030 | 2,932 | **+0 / -98** |

Those are my own measurements on my machine; the per-case tables and the grammar check above
are the parts you can reproduce directly from this description.

## Scope

Touches only `_js_local_bound_names`, and only by adding a second field read. It does not
overlap #1985 (`for_in_statement`, same function but a different branch) or #2517
(`catch_clause`, in `walk_calls`) — though all three are the same underlying class of bug: a
binding form that never reaches the shadow set.

### DIFF
--- a/graphify/extractors/engine.py
+++ b/graphify/extractors/engine.py
@@ -1164,6 +1164,13 @@ def _js_local_bound_names(func_node, source: bytes) -> set[str]:
     params = func_node.child_by_field_name("parameters")
     if params is not None:
         _js_collect_pattern_idents(params, source, bound)
+    # An arrow with ONE unparenthesised parameter exposes it as `parameter`
+    # (singular) — there is no `parameters` list node — so `x => f(x)` bound
+    # nothing at all and `x` read as a by-name reference to any same-named
+    # callable in the corpus. Same singular/plural trap as `catch_clause`.
+    solo = func_node.child_by_field_name("parameter")
+    if solo is not None:
+        _js_collect_pattern_idents(solo, source, bound)
 
     def walk(n) -> None:
         for c in n.children:
--- a/tests/test_indirect_call_arrow_single_param_shadow.py
+++ b/tests/test_indirect_call_arrow_single_param_shadow.py
@@ -1,0 +1,91 @@
+"""A single unparenthesised arrow parameter must shadow indirect_call args.
+
+`_js_local_bound_names` read only the `parameters` field. tree-sitter gives an
+arrow with ONE unparenthesised parameter a `parameter` field (singular) and no
+`parameters` list node at all, so `x => sink(x)` contributed nothing to the shadow
+set: `x` read as an unresolved by-name reference, resolved against the corpus-wide
+label index, and fabricated an `indirect_call` edge (INFERRED, 0.8) to an
+unrelated same-named callable. Minified bundles name nearly every private
+function with a single letter and use this arrow form heavily, so the two collide
+constantly.
+
+This is the same singular/plural trap as `catch_clause.parameter`. The
+parenthesised form was always handled, which is what makes the bug easy to miss:
+`(x) => …` and `x => …` behaved differently.
+"""
+import os
+from pathlib import Path
+
+from graphify.extract import extract
+
+
+def _extract_js_dir(tmp_path, files: dict[str, str]):
+    base = tmp_path / "src"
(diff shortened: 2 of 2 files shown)


safishamsi:
Shipped in v0.9.40 (`graphifyy==0.9.40` on PyPI). Unparenthesised arrow-param shadow (composes with the 0.9.38 #2568 fix). Credited in the release notes. Thanks @imagineers-tyler!

PR #2574, opened 2026-08-09
fix(extract,affected): edge `import('…')` in plain JS/TS, and travers…

### LINKED ISSUE #2575: JS/TS: `await import('…')` inside a nested function produces no edge — and `affected` cannot traverse `dynamic_import` at all
## Summary

`await import('…')` inside a **nested function** (a function declared inside another function) produces **no edge** in plain `.ts`/`.js`. The same statement at the top level of a function resolves fine, which is what makes this easy to miss — a corpus can look fully connected while every cycle-breaking dynamic import is silently dropped.

Distinct from #2153 (that one is `jsconfig` `baseUrl` and states outright it is *not* a dynamic-`import()` issue) and from #692 (`.svelte` markup-level).

## Reproduction

Four files, one import statement each, identical specifier. graphify 0.9.36 and 0.9.37.

```ts
// src/agent/AgentRunner.ts
export const runAgent = () => 1
```

```ts
// C_toplevel.ts  — import at the top level of an exported function
export async function go() {
    const { runAgent } = await import('../agent/AgentRunner');
    return runAgent;
}
```

```ts
// D_in_try.ts  — same, wrapped in try/catch
export async function go() {
    try {
        const { runAgent } = await import('../agent/AgentRunner');
        return runAgent;
    } catch { return null; }
}
```

```ts
// E_nested_fn.ts  — function inside a function
export function outer() {
    async function inner() {
        const { runAgent } = await import('../agent/AgentRunner');
        return runAgent;
    }
    return inner;
}
```

```ts
// F_nested_try.ts  — nested + try
export function outer() {
    async function inner() {
        try {
            const { runAgent } = await import('../agent/AgentRunner');
            return runAgent;
        } catch { return null; }
    }
    return inner;
}
```

```bash
```

| file | edge to `AgentRunner.ts` |
|---|---|
| `C_toplevel.ts` | ✅ `imports_from` |
| `D_in_try.ts` | ✅ `imports_from` |
| **`E_nested_fn.ts`** | ❌ **none** |
| **`F_nested_try.ts`** | ❌ **none** |

Nesting is the only variable. Not the alias (`@/…` behaves the same), not the specifier depth (`../../…` behaves the same), not the target file's size, not the presence of static imports in the same file — all four were tested and ruled out separately.

## Second, independent gap: `affected` cannot traverse dynamic edges

`DEFAULT_AFFECTED_RELATIONS` in `graphify/affected.py` does not contain `dynamic_import`. So even the dynamic edges the Svelte/Astro/Vue rescue passes **do** emit are invisible to `graphify affected`. Fixing the extractor alone does not restore blast-radius traversal.

## Impact, measured

On a 546-file TypeScript app that uses `await import()` deliberately to break import cycles — so the dynamic imports sit under the hub modules — **31 real edges were missing** from `graph.json`.

Because the loss is concentrated under hubs, it compounds with depth. Against ground truth from an independently written import resolver (730 files, 1,961 edges, 1.7% margin — deliberately not graphify, so the tool is not grading itself):

| depth | truly affected | `graphify affected` | recall |
|---|---|---|---|
| 1 | 2 | 2 | 1.00 |
| 2 | 8 | 5 | 0.62 |
| 3 | 30 | 8 | **0.27** |

Precision stayed **1.00** at every depth — graphify under-reports here, it never invents. That is the kindest possible failure mode, and also the reason it goes unnoticed: the answer looks clean, it is just short.

One missing edge under a hub amputated an entire subtree — a single store module pulled ~14 UI components out of the blast radius with it.

## Suggested fix

The mechanism already exists in the codebase. `extract_svelte` / `extract_astro` / `extract_vue` each run a regex rescue for exactly this, and `extract_vue`'s docstring names the underlying cause:

> *"A regex pass then recovers `import('…')` dynamic imports the AST does not edge"*

Plain `.ts`/`.js` never got that pass, presumably because its AST pass works — which it does, for static imports and for un-nested dynamic ones.

Applying the same rescue in `extract_js` (reusing `_emit_rescued_import`, `_load_tsconfig_aliases`, `_load_tsconfig_base_url`, so aliases resolve identically to static imports) plus adding `dynamic_import` to `DEFAULT_AFFECTED_RELATIONS`:

- all four reproduction files above get their edge;
- on the 546-file app: 31 missing edges → **0**, recall **0.27 → 1.00** at every depth, precision still 1.00;
- graphify's own suite is unchanged — 6 failed / 3931 passed identically with and without, those 6 being environmental in my setup (missing `ollama`, a wheel-build test, an `apm` manifest dependency).

Happy to open a PR — a branch with the change plus six tests (relative, multi-level relative, tsconfig-aliased, static-alongside-dynamic, an identifier-boundary guard so `fooimport('x')` does not match, and one asserting `affected` can traverse the relation) is ready. Verified to fail without the change: 5 of the 6 are red on `v8`.

## Environment

graphify 0.9.36 (PyPI) and 0.9.37 (`v8` @ `09a34ad`) · macOS 15 arm64 · Python 3.12

phudayyy:
Fixes #2575

tree-sitter models `await import('x')` as a `call_expression`, not an `import_statement`, so the specifier never reaches the import walk. The Svelte/Astro/Vue extractors already compensate with a regex pass — the docstring in `extract_vue` says so outright ("A regex pass then recovers `import('…')` dynamic imports the AST does not edge"). Plain .ts/.js was left out, on the reasonable-sounding basis that its AST pass works. It works for STATIC imports; the dynamic ones fall through silently.

Two independent gaps, and fixing either alone changes nothing:

1. extract.py — no dynamic-import edge is produced for plain JS/TS at all.
2. affected.py — `dynamic_import` was missing from DEFAULT_AFFECTED_RELATIONS, so even the dynamic edges the Svelte/Astro/Vue passes DO emit were invisible to `graphify affected`.

Measured on a 546-file TypeScript app (graphify 0.9.36/0.9.37, 2026-08-09): 31 real `import('…')` edges missing from graph.json. The loss compounds with depth because dynamic imports cluster under hub modules — codebases use them precisely to break require cycles, so the missing edges are the load-bearing ones. On that app, `affected <file> --depth 3` returned 10 of 30 truly-affected files (recall 0.27); ground truth came from an independently written import resolver, not from graphify. Precision stayed 1.00 throughout: the graph under-reports, it never invents.

With this change, on the same corpus: 31 missing edges -> 0, and recall goes to 1.00 at every depth on all three targets tested, precision still 1.00.

Implementation reuses what already exists — `_emit_rescued_import`, `_load_tsconfig_aliases`, `_load_tsconfig_base_url` — so aliased specifiers (`@/agent/runner`) resolve through the same path as static ones. The regex is guarded with `(?<!\w)` so `fooimport('x')` does not match, and the pass bails early when the file contains no `import(` at all.

tests/test_js_dynamic_imports.py covers relative, multi-level relative and tsconfig-aliased dynamic imports, asserts a static import still edges alongside a dynamic one, guards the `(?<!\w)` boundary, and asserts `dynamic_import` is traversable by `affected`. Verified to fail without this change: 5 of the 6 fail on v8, all 6 pass with it.

Suite: 6 failed / 3931 passed, byte-identical with and without this change — those 6 are pre-existing in my environment (missing `ollama`, a wheel-build test, an apm manifest dependency).

### DIFF
--- a/graphify/affected.py
+++ b/graphify/affected.py
@@ -15,6 +15,12 @@ DEFAULT_AFFECTED_RELATIONS = (
     "references",
     "imports",
     "imports_from",
+    # `import('…')`. Emitted by the Svelte/Astro/Vue rescue passes and (since the
+    # companion fix in extract.py) by plain JS/TS too. Omitting it made every
+    # dynamic import invisible to blast-radius traversal even where the edge WAS
+    # in the graph — and dynamic import is precisely how codebases break require
+    # cycles, so the missing edges sit under the most load-bearing modules.
+    "dynamic_import",
     "re_exports",
     "inherits",
     "extends",
--- a/graphify/extract.py
+++ b/graphify/extract.py
@@ -1231,7 +1231,49 @@ def extract_js(path: Path) -> dict:
     result = _extract_generic(path, config)
     if "error" not in result:
         _extract_js_rationale(path, result)
+        _rescue_js_dynamic_imports(path, result)
     return result
+
+
+def _rescue_js_dynamic_imports(path: Path, result: dict) -> None:
+    """Recover ``import('…')`` edges the AST pass does not emit for plain JS/TS.
+
+    tree-sitter models ``await import('x')`` as a ``call_expression``, not an
+    ``import_statement``, so the specifier never reaches the import walk. This is
+    the same gap :func:`extract_svelte`, :func:`extract_astro` and
+    :func:`extract_vue` already patch by regex — those file types only got the
+    rescue because their AST pass fails wholesale, while plain ``.ts``/``.js``
+    were left out on the reasoning that their AST pass "works". It works for
+    STATIC imports; the dynamic ones fall through silently.
+
+    Measured on a 546-file TypeScript app (2026-08-09): 31 real
+    ``import('…')`` edges were missing from graph.json. Because they cluster
+    under hub modules, the loss compounds with traversal depth — one such app's
+    ``affected --depth 3`` returned 8 of 30 truly-affected files (recall 0.27,
+    precision still 1.00: the graph under-reports, it never invents). Codebases
+    that use dynamic import deliberately to break require cycles are hit hardest,
+    since those edges are exactly the load-bearing ones.
+    """
+    try:
+        import re as _re
+        src = path.read_text(encoding="utf-8", errors="replace")
+        if "import(" not in src:  # cheap bail — most files have none
+            return
+        existing_ids = {n["id"] for n in result.get("nodes", [])}
+        file_node_id = _make_id(str(path))
+        aliases = _load_tsconfig_aliases(path.parent)
+        base_url = _load_tsconfig_base_url(path.parent)
+        # `(?<!\w)` so `fooimport('x')` and `_import('x')` do not match.
+        for m in _re.finditer(r"""(?<!\w)import\(\s*['"]([^'"]+)['"]\s*\)""", src):
+            raw = m.group(1)
+            if not raw:
+                continue
+            _emit_rescued_import(
+                result, existing_ids, file_node_id, path, raw,
+                "dynamic_import", aliases, base_url,
+            )
+    except Exception:
+        pass
 
 
 # ── JS/TS rationale + doc-reference extraction ────────────────────────────────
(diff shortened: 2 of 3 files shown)


safishamsi:
Shipped in v0.9.38. Your dynamic-import rescue + `dynamic_import` in DEFAULT_AFFECTED_RELATIONS is what landed for #2575, with a dedupe so an AST-captured deferred import is not double-counted and a companion engine fix so calls in nested named functions resolve too. Credited in the release notes. Thanks @phudayyy!

Issue #2577, opened 2026-08-09
SQL extractor emits reads_from edges to bare, unqualified stub IDs that collide across languages


scademir:
## Summary

The SQL extractor emits `reads_from` edges to **bare, unqualified stub node IDs** (`slug`, `name`, `position`, `levels`, `public`). Because those IDs carry no schema, file, or language namespace, they collide with same-named nodes from other languages during the build. In our repo a SQL column named `slug` bound to a TypeScript function `slug()` in a `.tsx` file, producing a phantom SQL→TypeScript edge marked `[EXTRACTED]`.

That single phantom edge became **the highest-betweenness node in a 9,294-node graph (0.106)** — because it was one of the only paths between the SQL half and the TypeScript half, every cross-language shortest path routed through it. `GRAPH_REPORT.md` then surfaced it as the #1 "Suggested Question," i.e. the tool's headline insight was its own extraction bug.

Version: `graphifyy` 0.9.37 · Windows 11 · Python 3.14 · extraction is AST-only (no LLM).

## Three distinct defects

**1. A CTE name is emitted as a table `reads_from` target.**
`with levels(role, lvl) as (values ...)` → two `reads_from` edges to a stub named `levels`. A CTE is not a table.

**2. An INSERT column list is parsed as table references.**
```sql
insert into public.stages (organization_id, pipeline_id, name, slug, position, is_won, is_lost)
```
emits **six** `reads_from` edges — one per column — each to a stub named after the column, as if it were a table.

**3. The edge is attributed to the wrong enclosing function.**
All of the above are credited to `fn_role_at_least()`, which ends 28 lines *above* the statement. The statement lives in a different function entirely (`fn_seed_stages()`). The enclosing object appears to resolve to the last `CREATE FUNCTION` seen before the parse failure rather than the one containing the line — so `source_location` and the edge's source node disagree.

All are labeled `[EXTRACTED]`, the highest-confidence tier. The audit trail distinguishes "read from the AST" from "model-reasoned" (`INFERRED`), and these name-resolution guesses land in the wrong bucket — which is what let a phantom edge pass review.

## Reproducer A — minimal, synthetic (defects 1 and 3)

`schema.sql`:
```sql
CREATE OR REPLACE FUNCTION "public"."fn_role_at_least"("p_min" "text") RETURNS boolean
    LANGUAGE "sql" STABLE
    AS $$
  with levels(role, lvl) as (
    values ('viewer',1),('admin',4)
  )
  select coalesce(
[...]

PR #2579, opened 2026-08-09
fix(sql): a CTE name is no longer emitted as a reads_from table (#2577)

wilyan09007:
Addresses the first of the three defects in #2577: a CTE name emitted as a `reads_from` table target.

## The defect

A name bound by `WITH <name> AS (...)` is scoped to the statement that declares it, so it is not a table. `_walk_from_refs` has no CTE awareness and emits a `reads_from` edge for every `FROM`/`JOIN` reference it finds, so a reference to the CTE mints a bare `_ref_stub`.

That is what makes it a cross-language problem rather than a cosmetic one. `_ref_stub` is deliberately sourceless so a table defined in another file can still resolve at corpus level (#2324, #1402), which means the CTE name enters the graph carrying no schema, file, or language namespace and can collide with a same-named node from any other language.

The fix collects the CTE names declared inside a statement and skips them when walking `FROM`/`JOIN` references. Stub IDs themselves are untouched, so cross-file table resolution keeps working.

## Reproduction

Worth flagging for anyone verifying this: **Reproducer A in the issue does not reach the code being blamed.** Its PL/pgSQL `CREATE FUNCTION` body makes tree-sitter-sql emit a root `ERROR` node, so extraction takes the recovery path at `t == "ERROR"`, which deliberately does not scan the body for `FROM`/`JOIN` ("PL/pgSQL loop variables and locals would produce junk reads_from targets"). On `v8` with tree-sitter-sql 0.3.11 that snippet produces no `reads_from` edges at all.

The same defect does reproduce through a statement the grammar parses, since `_walk_from_refs` only runs for `create_view` / `create_function` / `create_procedure` nodes that parse cleanly:

```sql
CREATE TABLE users (id SERIAL PRIMARY KEY, role TEXT NOT NULL);

CREATE VIEW v_roles AS
  WITH levels AS (SELECT 'admin' AS role)
  SELECT * FROM users JOIN levels ON levels.role = users.role;
```

Before:

```
nodes: ['cte2.sql', 'users', 'v_roles', 'levels']
 EDGE reads_from -> 'levels' EXTRACTED
```

Note `levels` next to the real nodes: the table and view get fully-qualified ids, the CTE gets the bare name. After the change the `levels` node and its edge are gone, and the `users` reference still emits its `reads_from` edge.

## Test

`tests/fixtures/sample_cte.sql` plus `test_sql_cte_is_not_read_as_a_table` in `tests/test_multilang.py`. It asserts the CTE never becomes a node or a `reads_from` target, and that the real table in the same `JOIN` still does, so it cannot pass by simply dropping the edges. The test fails on `v8` without the change and passes with it.

## Scope

This covers only the CTE defect. I could not reproduce the other two on `v8` with tree-sitter-sql 0.3.11, and I think they may not be reachable in 0.9.37 at all.

`reads_from` has exactly three emission sites: two inside the `fb_proc_or_trigger` branch, which is Firebird syntax and never appears for these files, and `_walk_from_refs`, which matches `from` and `join` nodes only. None of them can turn a column list into a table target, so I do not see how defect 2 (an INSERT column list emitting one edge per column) can occur on this code. Running the reporter's `fn_seed_stages()` shape produces both function nodes and zero edges. Defect 3 depends on those edges existing, so it goes the same way.

The likely explanation is version skew. The `ERROR` recovery path stopped scanning bodies for `FROM`/`JOIN` in #1910 (2026-07-15), with the comment that loop variables and locals produced junk `reads_from` targets, which is exactly the shape defect 2 describes. A build predating that would show all three symptoms together.

So defects 2 and 3 may already be fixed rather than outstanding. Confirming that needs the `graphify` and `tree-sitter-sql` versions from the environment that produced the report. If they can still be reproduced on 0.9.37 I am happy to take another pass.

## Testing notes

`python -m pytest tests/test_multilang.py -k sql` passes (14 tests). For the wider run I captured a baseline first: on `v8` on Windows with only the `sql` extra installed, `tests/test_multilang.py` + `tests/test_extract.py` are already 131 failed / 89 passed before any edit, several of them because tree-sitter-sql is installed at all (`test_extract_warns_when_sql_extra_missing` asserts the opposite). With this change the failure list is byte-identical and the pass count goes from 89 to 90, the new test.


safishamsi:
Shipped in v0.9.38 (`graphifyy==0.9.38`). Your CTE-exclusion fix is what landed for #2577, refined so CTE names are scoped per query (a subquery CTE no longer suppresses an outer real table of the same name). Credited in the release notes. Thanks @wilyan09007!

PR #2582, opened 2026-08-09
fix(export): stabilize graph JSON collection order

hjotha:
## Summary

- sort exported `nodes` and `links` with a canonical JSON comparison key
- sort hyperedges in both the NetworkX graph metadata and the top-level compatibility field
- add a regression proving reversed collection insertion order produces byte-identical `graph.json`

## Root cause

`networkx.node_link_data()` preserves graph insertion order. Equivalent graphs assembled in a different node, edge, or hyperedge order therefore produced noisy `graph.json` rewrites after small incremental changes.

This complements #1010, which stabilized edge-conflict resolution and clustering but did not canonicalize the serializer's collection order. The scope here is intentionally limited to collection ordering; mapping-key ordering is unchanged.

## Impact

The graph topology and attributes are unchanged. The first export may reorder existing collections once; later equivalent rebuilds remain byte-stable.

## Validation

- focused export suite: `53 passed`
- Python 3.12 full suite: `4141 passed, 3 skipped`
- Python 3.10 full suite: `4141 passed, 3 skipped`
- all five `tools.skillgen` CI validators passed
- Ruff passed on both changed files
- `git diff --check` passed
- repeated structural Graphify rebuild reported no topology changes on the second pass
- `graphify --help` and `graphify install` smoke tests passed in a clean HOME

### DIFF
--- a/graphify/export.py
+++ b/graphify/export.py
@@ -292,6 +292,10 @@ def to_json(G: nx.Graph, communities: dict[int, list[str]], output_path: str, *,
         data = json_graph.node_link_data(G, edges="links")
     except TypeError:
         data = json_graph.node_link_data(G)
+
+    def _json_sort_key(item: dict) -> str:
+        return json.dumps(item, ensure_ascii=False, sort_keys=True, separators=(",", ":"))
+
     for node in data["nodes"]:
         cid = node_community.get(node["id"])
         node["community"] = cid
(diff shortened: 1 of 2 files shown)


safishamsi:
Fixed in v0.9.40 (`graphifyy==0.9.40` on PyPI). `graph.json` node/edge/hyperedge collection order is now sorted deterministically, so identical inputs produce byte-stable output across runs. Thanks @hjotha (PR #2582).

PR #2583, opened 2026-08-10
fix(serve): resolve punctuated and non-ASCII node ids in _find_node (#2467)

### LINKED ISSUE #2467: explain/_find_node cannot resolve node IDs containing punctuation or non-ASCII text (concept:*, merge-graphs ::, Hangul)

sean-soomgo:
Fixes #2467.

`_find_node` cannot resolve a node id that contains punctuation, and separately cannot resolve one that contains non-ASCII text. Two defects, one shared theme: the query is normalized, the stored id is compared raw.

The user-visible shape of it is that `explain` prints an id it cannot itself consume (this example from @thomasrengot-hub in the issue thread, on a `merge-graphs` graph where *every* node is affected):

```
$ graphify explain "LisaNavGraph"
Node: LisaNavGraph()
  ID:        android::app_src_main_kotlin_lisanavgraph

$ graphify explain "android::app_src_main_kotlin_lisanavgraph"
No node matching 'android::app_src_main_kotlin_lisanavgraph' found.
```

and that `find_node_ambiguity`'s remedy — "Retry with the repo-relative path or the full node id" — is advice that cannot succeed. Falling back to the label instead resolves silently to a *different* node, which is harder to notice than an error.

## Root causes

**Punctuation — the exact-tier predicate.** `_find_node_tiers` builds two normalizations of the query: `term`, which tokenizes on `\w+` so punctuation becomes a space, and `norm_query`, which keeps it. The exact tier compared the id against `term` only:

```python
term       = " ".join(_search_tokens(label))        # "concept:domain:x" -> "concept domain x"
norm_query = _strip_diacritics(str(label)).lower().strip()   # "concept:domain:x"
...
elif (
    term == norm_label or term == bare_label or term == label_tokens or term == nid_lower
    or norm_query == norm_label or norm_query == bare_label
):
```

`term == nid_lower` is false for any id carrying punctuation, and `norm_query` — which already held the right form, and is even one of the two needles handed to `_trigram_candidates` — was never compared against the id. This affects `concept:*` overlay nodes, and every node of a `merge-graphs` graph, since `prefix_graph_for_global` namespaces all of them with `<repo>::`.

**Non-ASCII — the trigram prefilter.** `_node_search_text` indexed the id raw (`graphify/serve.py:333`) while every query path folds through `_strip_diacritics`, which NFKD-decomposes. Hangul syllables decompose into conjoining jamo, and jamo have combining class 0, so they survive the combining-character filter. The needle's trigrams and the posting's trigrams are therefore disjoint, `_trigram_candidates` returns a candidate list without the node, and the node is dropped before any predicate runs. The predicate fix alone does not reach this class: the full-scan path is gated by the predicate, the index path by the prefilter.

## The change

Two edits in `graphify/serve.py`:

- the exact tier also compares `norm_query` against the diacritic-folded id;
- `_node_search_text` appends the folded id to the indexed text.

Both are additive. An id that resolved before resolves to the same node; only ids that previously resolved to nothing can now resolve. Candidate sets stay a superset and the exact predicates still decide, so the prefilter cannot introduce a match the predicate would not have made on a full scan.

The folded field is appended rather than inserted, and only when the fold actually differs from the raw id, so no field position moves and an all-ASCII graph indexes byte for byte what it indexed before. Index build, median of 7 runs:

| graph | | trigrams | postings | build |
|---|---|---:|---:|---:|
| 5k all-ASCII | before | 1723 | 238876 | 96ms |
| 5k all-ASCII | after | 1723 | 238876 | 97ms |
| 17k real | before | 33442 | 2311784 | 998ms |
| 17k real | after | 33490 | 2312462 | 990ms |
| 5k half-Hangul | before | 1735 | 256381 | 116ms |
| 5k half-Hangul | after | 1739 | 266381 | 124ms |

"17k real" is the graph measured in the next section; 354 of its 17,269 ids are non-ASCII, so the index grows by 0.03% and the build time stays inside run-to-run noise. The half-Hangul row is a deliberate worst case — every other node id Korean — and even there it is a once-per-graph-load cost on a graph where id lookup previously returned nothing at all.

## Measurements

A real 17,269-node graph with Korean source filenames. Every node id fed back to itself, full population, not a sample:

| id class | total | before | after |
|---|---:|---:|---:|
| contains punctuation | 2326 | 0 | 2326 |
| contains Hangul | 354 | 0 | 354 |
| ASCII, no punctuation | 14589 | 14589 | 14589 |

And every query the graph can produce — all 17,269 ids plus all 16,537 distinct labels, 33,806 in total — through `_find_node_tiers` on both variants in the same process:

| outcome | count |
|---|---:|
| identical | 31126 |
| returned nothing before, resolves now | 2680 |
| first result changed | 0 |
| result lost | 0 |
| exact tier grew | 0 |

Every single difference is a query that returned nothing before. No label query changes, nothing reorders, nothing becomes newly ambiguous.

One measurement trap worth passing on, since I fell into it: `_get_trigram_index` caches on `G.graph`, so a before/after sweep that reuses one graph object silently scores the second variant against the first one's index. My first run of this table reported 0/354 for the Hangul class because of it.

## Tests

Five tests in `tests/test_serve.py`, all of which fail on `v8` and pass with the change:

- `test_find_node_matches_punctuated_node_id_exactly` — plus assertions that an unpunctuated id and a label query still resolve exactly as before;
- `test_find_node_matches_merge_graphs_namespaced_node_id` — the `<repo>::` case;
- `test_node_search_text_appends_folded_non_ascii_node_id`;
- `test_find_node_matches_non_ascii_node_id_through_prefilter`;
- `test_find_node_node_id_prefilter_is_identical_to_full_scan` — parity guard in the shape of the two existing `_prefilter_is_identical_to_full_scan` tests.

`test_node_search_text_includes_all_matched_fields` gains one assertion (`len(parts) == 5`) pinning that an ASCII id adds no field.

**One note for anyone touching these tests later.** The non-ASCII case needs a graph of at least ~10 nodes. `_trigram_candidates` bails out to a full scan when `min(present) > int(n * 0.10)`, so on a small synthetic graph the index path is never taken and the test passes with the defect still present — a two-node probe reported success for me while the real graph still returned 0/300. `_make_non_ascii_id_graph` carries 40 filler nodes for that reason, and the test asserts `_trigram_candidates(...) is not None` so a future guard change cannot quietly turn it into a no-op.

`uv run --frozen pytest tests/ -q` is clean apart from `test_skillgen.py` and `test_ollama_retry_cap.py`, which fail identically on unmodified `v8` in my environment (skillgen reads blobs from history that a shallow clone does not have).

## Deliberately not in scope

- `_score_query` compares the id raw in the same way (`graphify/serve.py:513`, `:522`, `:576`), so `path`, MCP `shortest_path` and query seeding still cannot take a punctuated id. Same root cause, but it is a scoring change rather than a resolution fix, so I left it out — happy to follow up if you want it in.
- The prefix tier also matches ids against `term` only, so `concept:domain` will not prefix-match `concept:domain:x` the way `conceptdomain` matches `conceptdomainx` today. Behaviour change, not part of this defect.
- Widening the exact tier can in principle turn a previously-unique match into a `find_node_ambiguity` hit, where a node id and a different node's label fold to the same string. I can construct that synthetically, but the full sweep above found zero instances of it on a real graph, so I have not tried to order the two against each other.


safishamsi:
Shipped in v0.9.40 (`graphifyy==0.9.40` on PyPI). Punctuated/non-ASCII node id resolution for #2467. Credited in the release notes. Thanks @sean-soomgo!

Issue #2315, opened 2026-07-29
Python: `@receiver.name(...)` decorators bind to an unrelated module-level function of the same bare name — false edges create a god node (regression from #2154, 0.9.27+)


dispather:
## Summary

Python decorators of the form `@receiver.name(...)` are resolved by their **bare attribute name only** — the receiver is dropped. If any module-level function in the corpus happens to share that name, every decorated function gets a false `references` (and at scale, `calls`) edge to it, across files that never import it.

The common real-world trigger is web frameworks: `@app.get(...)`, `@router.get(...)`, `@app.post(...)`. Any project that also defines a helper named `get`/`post` gets a fabricated hub.

This appears to come from decorator edges added in #2154 (v0.9.27).

## Minimal reproduction

Two files, no import between them:

```python
# settings_store.py — project-defined module-level function named `get`
def get(key: str):
    return key
```

```python
# routes.py — deliberately does NOT import settings_store
class _App:
    def get(self, path):
        def deco(fn):
            return fn
        return deco

app = _App()

@app.get("/health")
def health():
    return "ok"

@app.get("/status")
def status():
    return "ok"
```

```
graphify . --no-cluster
```

**Actual** — `graph.json` contains:

```
references  health() (routes.py)  -->  get() (settings_store.py)
references  status() (routes.py)  -->  get() (settings_store.py)
```

**Expected** — no cross-file edge. `app` is a local `_App` instance; `settings_store.get` is unrelated and never imported.

Note the resolver also skips the *correct* candidate: `_App.get` is defined in the same file (it appears in the graph as a `.get()` node) and is the actual receiver's method, but the edge goes to the unrelated module-level function instead.

## Control experiment (isolates the cause)

Rename only the project-defined function, `def get` -> `def get_knob`, leaving the `@app.get` decorators untouched:

```
references  health()  -->  get   [stub node, source_file empty]
references  status()  -->  get   [stub node, source_file empty]
```

The edges now land on an unresolved **stub** — correct behaviour. So the binding keys purely on the bare decorator attribute name matching a module-level definition; nothing about the receiver or the import graph is consulted.

## Real-world impact

On a FastAPI codebase (197 Python files, 3.3k nodes) upgrading 0.9.26 -> 0.9.30:

- `settings_store.get()` degree: **27 -> 149**, becoming the **#1 god node** in `GRAPH_REPORT.md`, above genuine core abstractions.
[...]

PR #2586, opened 2026-08-10
fix(python): stop unresolved member calls binding by bare name

### LINKED ISSUE #2417: Python: bare method calls (`x.get(...)`) bind to a same-named module-level function — false `calls` edges create a god node (reproduces on every 0.9.x I tested)
## Summary

In Python extraction, an unresolved attribute call `x.get(k)` is resolved by **bare name** to a module-level function `get` defined in the project, producing a `calls` edge that does not exist. Any codebase that both defines a short, common function name (`get`, `set`, `run`, …) and uses the same method name on dicts/clients gets that function inflated into the #1 god node.

This is the Python call-site counterpart of #2241 (JS/TS closure arguments, fixed in 0.9.29) and looks closely related to #2315 (Python `@receiver.name(...)` decorators binding to an unrelated module-level function of the same bare name) — possibly the same bare-name fallback in the Python resolver.

## Minimal reproduction (~30s)

`fixture.py`, alone in an empty directory:

```python
def get(k):
    """Module-level function that shares a name with dict/mapping .get."""
    return k


class Store:
    def __init__(self):
        self.store = {}

    def read(self, k):
        return self.store.get(k)


def other(d):
    return d.get('x')


def real():
    return get('real')
```

```bash
graphify .
```

Then count `calls` edges whose target is the node labelled `get()` (the module-level function; methods are labelled `.name()`):

```python
import json
g = json.load(open("graphify-out/graph.json"))
t = {n["id"] for n in g["nodes"] if n.get("label") == "get()"}
print([l for l in g["links"] if l.get("relation") == "calls" and l["target"] in t])
```

**Expected:** 1 edge — `real() -> get()`.
**Actual:** 3 edges — `real() -> get()`, `other() -> get()`, `Store.read() -> get()`.

`self.store.get(k)` is a dict lookup and `d.get('x')` is a call on an unannotated parameter; neither can be resolved to the module-level `get`, but both are bound to it.

## Version sweep

Same fixture, run via `uv tool run --from graphifyy==<version> graphify .`:

| version | inbound `calls` into `get()` | verdict |
|---|---|---|
| 0.9.32, 0.9.31, 0.9.30, 0.9.29, 0.9.28, 0.9.27, 0.9.26, 0.9.24, 0.9.22, 0.9.20, 0.9.17, 0.9.14, 0.9.13, 0.9.5, 0.9.0 | 3 | ❌ buggy |

15/15 buggy — so this is **not** a regression introduced by a recent release. (0.8.0 could not be installed here: dependency resolution fails.)

## Impact on a real repo

On a ~3.3k-node Python project, one rebuild made the project's `get()` in `settings_store.py` jump from **degree 27 to 149**, taking the #1 god-node slot. A single module contributed 42 of those edges while never importing `settings_store` — it simply contains 127 `.get(` call sites. Across that rebuild, **122 of the 200 newly-added edges (61%) were false** edges of this shape (net edge count 6,927 → 7,110).

Downstream, this corrupts god-node ranking, `explain <node>`, and community assignment, and there is no version to pin to as a workaround.

## Suggested direction

Don't fall back to bare-name matching for attribute calls when the receiver is unresolved. If a fallback is needed for recall, emit it at a lower confidence and/or as a distinct relation so consumers can filter it, rather than as a plain `calls` edge.

## Environment

- graphify 0.9.32 (and the 14 older versions above), installed via `uv tool` / `uv tool run`
- macOS 26.5.2 (arm64), uv-managed CPython 3.13.12

EZZEASY:
## Summary

- Stop unresolved Python attribute calls such as `d.get()` and `self.store.get()` from falling back to a same-named module-level function.
- Preserve local `self` / `cls` / `super()` calls and evidence-backed class- or module-qualified member resolution.
- Add cold- and warm-cache regression coverage for the exact issue fixture, plus receiver and name-collision cases.

## Scope

This intentionally favors precision: when a Python member receiver cannot be resolved, Graphify leaves the call unconnected instead of guessing from the bare method name. This does not add receiver type inference or address #2315. There are no public API, schema, version, or dependency-lock changes.

## Test plan

- [x] `.venv/bin/python -m pytest tests/ -q --tb=short` — 4009 passed, 37 skipped
- [x] `.venv/bin/python -m pytest tests/test_extract.py -q --tb=short` — 167 passed, 1 skipped
- [x] `.venv/bin/ruff check graphify/extractors/engine.py tests/test_extract.py`
- [x] `.venv/bin/python -m tools.skillgen --check`
- [x] `git diff --check`
- [x] `graphify update .`

Test coverage audit: 100% of changed behavior paths covered. Pre-landing review: no issues found.

Fixes #2417

### DIFF
--- a/graphify/extractors/engine.py
+++ b/graphify/extractors/engine.py
@@ -4482,6 +4482,21 @@ def _extract_generic(
                             obj = func_node.child_by_field_name(config.call_accessor_object_field)
                             if obj is not None and obj.type == "identifier":
                                 member_receiver = _read_text(obj, source)
+                            elif (
+                                config.ts_module == "tree_sitter_python"
+                                and obj is not None
+                                and obj.type == "call"
+                            ):
+                                # ``super().method()`` has a call node as its
+                                # receiver. Preserve it as a known intra-class
+                                # receiver instead of treating it as unresolved.
+                                receiver_func = obj.child_by_field_name("function")
+                                if (
+                                    receiver_func is not None
+                                    and receiver_func.type == "identifier"
+                                    and _read_text(receiver_func, source) == "super"
+                                ):
+                                    member_receiver = "super"
                             elif (obj is not None
                                   and obj.type in config.call_accessor_node_types
                                   and config.call_accessor_object_field):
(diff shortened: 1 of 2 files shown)


safishamsi:
Shipped in v0.9.39. Deferring non-self/cls/super Python member calls to the evidence-gated resolver (with the `super()` capture) is what landed for #2417. Credited in the release notes. The same-file instantiated-receiver recall trade is documented in the changelog. Thanks @EZZEASY!

Issue #1201, opened 2026-06-08
Fuzzy dedup silently merges distinct long, near-identical symbols (length-differing pairs, both >= 12 chars)


Alex-R-A:
## Summary

Pass 2 fuzzy dedup in `graphify/dedup.py` merges two **distinct** symbols when their normalized labels are long (>= 12 chars) and differ by a small edit, e.g. a trailing plural `s`, or one name being a strict prefix-extension of the other. The guards that prevent this for short labels (`_is_variant_pair`, `_short_label_blocked`) both early-out for labels >= 12 chars, so long near-twins are never protected. The absorbed symbol disappears from `graph.json` and its edges (callers, imports, etc.) are silently reattached to the surviving node, corrupting call/dependency data.

## Minimal repro (uses graphify's own dedup functions)

```python
from rapidfuzz.distance import JaroWinkler
from graphify.dedup import _norm, _is_variant_pair, _short_label_blocked, _MERGE_THRESHOLD

a = _norm("getActiveSession()")     # 'getactivesession'  (16 chars)
b = _norm("getActiveSessions()")    # 'getactivesessions' (17) - a distinct batch/plural function

score = JaroWinkler.normalized_similarity(a, b) * 100
print("JW:", round(score, 2))                          # 98.82
print("MERGE_THRESHOLD:", _MERGE_THRESHOLD)            # 92.0  -> over threshold, merges
print("is_variant_pair blocks?:", _is_variant_pair(a, b))            # False (guard is < 12 chars only)
print("short_label_blocked?:",   _short_label_blocked(a, b, score))  # False (guard is < 12 chars only)
```

`score (98.82) >= 92` and neither guard applies (both return early at `max(len(a), len(b)) >= 12`), so the pair is unioned in Pass 2. The same-file partition check (the `if norm_label == neighbor_norm:` branch) only fires when the normalized labels are **identical**, so it does not catch this either, even when both symbols are defined in the same file.

## Effect

Two genuinely different functions in the same file, e.g. a single-item function and its batch sibling whose names differ only by a trailing `s`, collapse into one node. The plural is absent from `graph.json`; queries for it return nothing, and the singular's callers/dependents wrongly include the plural's call sites. The merge is silent, only an aggregate `Deduplicated N node(s) (... M fuzzy)` line is printed, so an unknown fraction of those `M` fuzzy merges may be false positives of this kind.

## Expected

[...]

PR #2587, opened 2026-08-10
fix(dedup): don't merge labels differing by one content word (#2576)

### LINKED ISSUE #2576: Fuzzy dedup: same-file long labels still over-merge on the Jaro-Winkler prefix bonus (follow-up to #1243)
## Summary

#1243 fixed the Jaro-Winkler prefix-bonus over-merge for **cross-file** pairs by scoring them on plain Jaro. The same failure is still reachable for **same-file** pairs, which were deliberately exempted. On a document corpus it fires readily:

`Asset Contribution Flow` is merged into `Asset Consumption Flow` — distinct, adjacent stages of one documented pipeline — and the absorbed node's edges are re-pointed at the survivor, so the graph asserts relationships that appear in no source.

The A/B is clean: **identical labels, only `source_file` differs.** Same file merges, different files keeps.

I don't think this is an oversight in #1243 — the exemption was reasoned. I think the reasoning holds for code corpora and inverts for prose. Detail below, plus a guard that addresses the specific blocker you named when you closed it.

## Environment

| | |
|---|---|
| graphifyy | 0.9.37 |
| Python | 3.14.6 |
| rapidfuzz | 3.14.5 |
| OS | macOS (Darwin 25.5.0), arm64 |

## Reproduction

```python
from graphify.dedup import deduplicate_entities

SRC = "docs/pipeline.md"

def node(nid, label):
    return {"id": nid, "label": label, "file_type": "concept", "source_file": SRC,
            "source_location": None, "source_url": None, "captured_at": None,
            "author": None, "contributor": None}

def edge(src, tgt):
    return {"source": src, "target": tgt, "relation": "references",
            "confidence": "EXTRACTED", "confidence_score": 1.0,
            "source_file": SRC, "source_location": None, "weight": 1.0}

nodes = [
    node("pipeline_asset_contribution_flow", "Asset Contribution Flow"),
    node("pipeline_asset_consumption_flow",  "Asset Consumption Flow"),
    node("pipeline_producer_personas",       "Producer Personas"),
    node("pipeline_asset_review_flow",       "Asset Review & Approval Flow"),
]
edges = [
    edge("pipeline_producer_personas", "pipeline_asset_contribution_flow"),
    edge("pipeline_asset_contribution_flow", "pipeline_asset_review_flow"),
]

out_nodes, out_edges = deduplicate_entities(nodes, edges, communities={})
for e in out_edges:
    print(e["source"], "->", e["target"])
```

**Actual**

```
[graphify] Deduplicated 1 node(s) (1 fuzzy).

pipeline_producer_personas       -> pipeline_asset_consumption_flow
pipeline_asset_consumption_flow  -> pipeline_asset_review_flow
```

`Asset Contribution Flow` is gone and both its edges now point at the survivor, producing two claims the source never makes:

- **Producer Personas → Asset Consumption Flow** — producers contribute assets, consumers consume them. Distinct personas in the document.
- **Asset Consumption Flow → Asset Review & Approval Flow** — review precedes listing and consumption. The pipeline direction is inverted.

**Expected** 4 nodes in, 4 nodes out (`contribute → review → list → consume → settle`).

## The trigger is `source_file`, not the labels

| | nodes out | result |
|---|---|---|
| both `docs/pipeline.md` | 1 | **MERGED** |
| `docs/pipeline.md` + `docs/other.md` | 2 | kept (correct) |

Scores for the pair, against `_MERGE_THRESHOLD = 92.0`:

| metric | score | verdict |
|---|---|---|
| Jaro-Winkler — same-file path | **92.74** | ≥ 92 → merge |
| Plain Jaro — cross-file path (#1243) | **87.90** | → keep |

It clears by 0.74, entirely on the prefix bonus from the shared `"Asset "` / `" Flow"` affixes. The distinguishing tokens score only 82.94 against each other (`JW("Contribution", "Consumption")`).

**This pair already falls inside the over-merge band from your own separation analysis in #1243** — true duplicates ≥ 97 on plain Jaro, over-merges ≤ 91. At 87.90 it is an over-merge by that criterion; it survives only because the same-file path never applies that criterion.

## Why the same-file exemption doesn't hold for document corpora

From the close comment on #1243:

> Same-file near-duplicate merging is fully preserved (every prior dedup test is same-file - zero regression)

That is the tell. For a **code** corpus the exemption is sound: two same-file nodes with near-identical labels usually *are* the same symbol extracted twice, and the risk concentrates cross-file. For a **document** corpus the risk profile inverts, because structured prose names sibling sections from a template. A single operations document will contain, as consecutive headings:

```
Asset Contribution Flow
Asset Review & Approval Flow
Asset Discovery & Search Flow
Asset Consumption Flow
```

Four distinct pipeline stages, near-identical strings *by construction*, all in one file. Same-file is where template-similar-but-distinct entities are densest, not sparsest. The existing guards don't help: `_short_label_blocked` and `_is_variant_pair` early-return at ≥ 12 chars (these are 22 and 23), the #1201 prefix-extension guard needs one label to be a strict prefix of the other, and `_numeric_tokens_differ` finds no numerics.

## Addressing the blocker you named

The reason given for scoping to cross-file:

> a single mid-string space/stopword insertion in a legit duplicate (…export a read-only… vs …export the read-only…, Jaro 89.6) is statistically indistinguishable from a distinguishing-token swap (jest/react, Jaro 91.3) - no string metric separates them

Agreed — no *whole-string* metric separates them. A **token-level** comparison does:

| pair | differing tokens | both content words? |
|---|---|---|
| `export a read-only` / `export the read-only` | `a` / `the` | no — stopwords → allow merge |
| `Asset Contribution Flow` / `Asset Consumption Flow` | `Contribution` / `Consumption` | yes → block merge |
| `testing-library/jest-native` / `react-native` | `jest` / `react` | yes → block merge |

The two cases are only indistinguishable when you compare whole strings. Once you look at *which* token differs, they separate cleanly — a stopword swap and a content-word swap are categorically different edits.

## Suggested fix

**Primary — single-differing-token guard (keeps same-file Jaro-Winkler):** if two labels have equal token counts and differ in exactly one token, and neither differing token is a stopword, block the merge regardless of whole-string score. This preserves the mid-string stopword-insertion merges you called out, keeps the prefix bonus where it helps, needs no threshold retuning, and mirrors the per-token reasoning already in `_numeric_tokens_differ`. It also would have caught the original #1243 examples independently.

**Alternative — drop the `_xfile` condition** so plain Jaro applies to all long labels. Simpler, and resolves this case (87.90 < 92.0), but it *would* regress your stopword-insertion example (Jaro 89.6 < 92), which is presumably why you scoped it. I'd favour the token guard for that reason.

**Also worth considering:** the summary line names neither node. #1857 restored the fuzzy *count*, and my run does print `(1 fuzzy)` — but "1 fuzzy" gave no indication that `Asset Contribution Flow` had been absorbed into `Asset Consumption Flow`. I only found it by diffing node sets by hand after the graph looked wrong. Logging `survivor ← absorbed` labels for fuzzy merges specifically would make destructive merges self-reporting, and would have turned a review session into a one-line read.

## Secondary: `--update` and full-build disagree

`build_merge(..., dedup=True)` is the default on the `--update` path, but a full `graphify .` build runs no entity dedup at all. On a 12-file corpus that is **173 nodes** (full build) vs **160 nodes** (stock update) from identical inputs — 12 exact-label merges plus this fuzzy one. A graph silently changes shape depending on whether the user rebuilt or updated.

The exact-label merges look questionable too, under the per-file node-ID scheme in `references/extraction-spec.md`: two documents that each define a shared term produce `{stem_a}_term` and `{stem_b}_term` deliberately, so each keeps its own `source_file` and the restatement is modelled as a scored `semantically_similar_to` edge. Collapsing them discards which document said what — which is what the EXTRACTED/INFERRED audit trail exists to preserve. Happy to split this into its own issue if you'd rather keep this one focused.

## Workaround

`build_merge(..., dedup=False)` on the update path.

wilyan09007:
Fixes #2576.

## Summary

- Two labels that share their affixes and differ in a single word (`Asset Contribution Flow` / `Asset Consumption Flow`) clear `_MERGE_THRESHOLD` on Jaro-Winkler's prefix bonus alone: 92.74 with it, 87.90 on plain Jaro. The absorbed node's edges are re-pointed at the survivor, so the graph asserts relationships no source states.
- Closing #1243 you wrote that a mid-string stopword insertion in a legit duplicate (`...export a read-only...` vs `...export the read-only...`, Jaro 89.6) is statistically indistinguishable from a distinguishing-token swap (`jest`/`react`, Jaro 91.3), and that no string metric separates them. That holds for whole strings. Scoring the two tokens that actually differ separates them:

| pair | differing tokens | tokens scored against each other |
|---|---|---|
| `export a read-only` / `export the read-only` | `a` / `the` | 0.00 |
| `Asset Contribution Flow` / `Asset Consumption Flow` | `Contribution` / `Consumption` | 82.94 |
| `testing-library/jest-native` / `testing-library/react-native` | `jest` / `react` | 63.33 |
| `Authentication Manager` / `Authentication Managr` | `Manager` / `Managr` | 97.14 |

- Fix: `_content_token_swap`, next to `_numeric_tokens_differ`. When two labels have equal token counts and differ in exactly one position, that token pair decides. A stopword on either side exempts the pair, so restatements keep merging. Otherwise the two tokens are held to the same threshold the labels are, so a misspelling still merges and two distinct words do not. Pairs with different token counts are left to the #1201 prefix-extension guard. Mirrored into `_llm_tiebreak` with the other decisive guards.
- Not scoped to same-file, because two pairs from #1243's own repro still reproduce on `v8` and only one of them is a same-file pair:
  - the npm pair is same-file, both dependencies coming from one `package.json`, so the cross-file scoring never applied to it
  - the dated doc slugs are cross-file, but both nodes sit in one community and the +5 boost lifts plain Jaro 87.68 back to 92.68

  The regression tests added with #1243 miss both: they use `pkg-a/` and `pkg-b/` paths and `communities={}`. Both are covered now.

## Test

Six cases in `tests/test_dedup.py`: the issue's reproduction including its edges, the two #1243 pairs above, non-regression on the stopword insertion and on a one-token typo, and a unit test for the helper. The three over-merge tests fail on the unpatched tree.

```
python -m pytest tests/test_dedup.py -q
77 passed
```


safishamsi:
Shipped in v0.9.39. Your `_content_token_swap` guard is what landed for #2576, with two hardening deltas (any distinct differing content-word blocks; a same-length Damerau-Levenshtein typo escape) to also cover two-token and short-token cases. Credited in the release notes. Thanks @wilyan09007!

PR #2588, opened 2026-08-10
fix(js): let `affected` traverse a dynamic import written inside a fu…

### LINKED ISSUE #2584: JS/TS: `affected` cannot traverse a dynamic `import('…')` made inside a function — edge exists, blast radius is short by 20%
## Summary

`graphify affected` under-reports blast radius when a dynamic `import('…')` sits **inside a
function**. The edge is present in `graph.json` — this is not the missing-edge bug from #2575,
which 0.9.38 fixed — but reverse traversal dead-ends on it, because a dynamic edge originates
from the **enclosing symbol** while a static edge originates from the **file**.

Measured on a real repo (~700 JS/TS files), `affected --depth 3` returns **39 of 49** truly
affected files for one hub module and **24 of 30** for another: recall **0.80**. Precision stays
1.00 — nothing is invented, files are simply unreachable. Raising `--depth` to 4, 5 or 6 does not
help, which is the signature of a dead end rather than a depth limit.

## Root cause

Two lines in `graphify/extractors/engine.py`, both emitting `relation: "imports_from"`:

```python
# L1736 — static import
"source": file_nid,     # file → file

# L1289 — dynamic import, _dynamic_import_js()
"source": caller_nid,   # ENCLOSING SYMBOL → file
```

`caller_nid` is the file node only when the `import()` is at module level. Inside a function it is
that function's node, so the graph gets `load() --imports_from--> target.ts` and **no** file-level
edge `dyn.ts → target.ts`. Reverse traversal from `target.ts` therefore lands on `load()` and stops:
the only edge pointing at `load()` is `dyn.ts --contains--> load()`, and `contains` is not in
`DEFAULT_AFFECTED_RELATIONS`. Anything reachable only *through* that file is lost.

It escapes notice because it is invisible in the common case: if the next importer happens to import
that exact symbol by name, there is an edge into `load()` and traversal continues.

## Minimal reproduction

Four files. `mid.ts` is byte-identical in both runs; only the **position** of the dynamic import
changes.

```ts
// src/target.ts
export const value = 1;

// src/dyn.ts   ← the only file that differs between the two runs
export async function load() {          // variant A: inside a function
  const m = await import('./target');
  return m.value;
}
// variant B: export const p = import('./target');   // module level

// src/mid.ts
import './dyn';
export const run = () => 1;

// src/top.ts
import { run } from './mid';
export const go = () => run();
```

```console
$ graphify update . && graphify affected src/target.ts --depth 3
```

| `import('./target')` is… | `top.ts` in the result |
|---|---|
| at module level | ✅ yes |
| inside a function | ❌ no |

A second, independent axis reproduces it with the import fixed inside a function and only `mid.ts`
changing — traversal survives **only** when the importer names the enclosing symbol:

| `mid.ts` imports… | `top.ts` reached |
|---|---|
| `import { load } from './dyn'` (the enclosing symbol) | ✅ yes |
| `import { other } from './dyn'` (a different symbol) | ❌ no |
| `import * as ns from './dyn'` | ❌ no |
| `import './dyn'` (side-effect) | ❌ no |

Environment: graphify 0.9.38 (PyPI), macOS 15 / arm64, Python 3.13, TypeScript sources.

## Suggested fix

Emit the dynamic edge at **file granularity as well**, so blast radius matches static imports:

```python
# in _dynamic_import_js(), alongside the caller_nid edge
edges.append({**edge, "source": file_nid})
```

Keeping the symbol-level edge preserves the precision that makes `graphify explain` useful — the
call site is still recorded — while the file-level edge restores traversal. The `deferred: True`
flag added for #1241 should be kept on both so `find_import_cycles` still ignores them and no
phantom file cycle reappears.

Adding `contains` to `DEFAULT_AFFECTED_RELATIONS` would also reconnect it, but far more bluntly:
it would let every traversal climb from any symbol to its whole file, which would cost precision
everywhere to fix one relation.

Happy to send a PR with the fix plus a test extending `tests/test_js_dynamic_imports.py` — the
existing test asserts the edge exists, so an assertion that `affected` actually *traverses* it
would have caught this.

phudayyy:
Issue #2584

`graphify affected` under-reports blast radius when `import('...')` sits inside a function. The edge exists -- this is not #2575, which 0.9.38 fixed -- but reverse traversal dead-ends on it.

The two passes emit at different granularities. `_dynamic_import_js` anchors on `caller_nid`, which is the file node only at module scope; inside a function it is that function's node. The rescue pass in `_rescue_js_dynamic_imports` emits the file-level `dynamic_import` edge -- but its dedupe matched on TARGET alone, saw the AST pass had already emitted a deferred edge to that target, and skipped. So an `import()` inside a function ended up with no file-level edge at all, and the reverse walk stopped at the enclosing function: the one edge pointing at it is `contains`, deliberately not in DEFAULT_AFFECTED_RELATIONS.

The dedupe was right, just keyed wrong. It now matches on (this file's node, target), so the genuinely redundant case -- a module-scope `import()`, where `caller_nid` IS the file node -- still collapses to one edge, while the in-function case gets the file-level edge it always needed.

Measured on a ~700-file TypeScript repo, `affected --depth 3` against three hub modules, ground truth built independently of graphify:

  before   39/49, 24/30, 84/84   recall 0.80 / 0.80 / 1.00
  after    49/49, 30/30, 84/84   recall 1.00 / 1.00 / 1.00

Precision stays 1.00 in both. Raising --depth to 4, 5 or 6 did not move the before numbers, which is what a dead end looks like as opposed to a depth limit.

It hid for so long because the common case still resolves: when the next importer imports that exact symbol by name there IS an edge into the function, and traversal continues. Change that importer to `import * as ns` or a side-effect `import './x'` and the same graph goes silent. The new tests pin exactly those cases.

Also dedupes the rescue's own emissions per file. Two `import('./x')` in one file already produced two identical file-level edges before this change; that was rare while only module-scope imports reached the rescue, but every in-function dynamic import now does, so a hub deferred from eight functions of one file would have carried eight identical arrows.

`test_ast_captured_dynamic_import_is_one_fact_not_two` asserted the invariant that was the bug, so it is rewritten rather than deleted -- the docstring explains why one edge is not enough -- and the half of it that was always correct is now pinned on its own in `test_module_scope_dynamic_import_is_still_one_fact`.

No new relation, no schema change. `find_import_cycles` reads only `imports_from` and `re_exports` and skips `deferred`, so the file cycle of #1241 cannot return through a `dynamic_import` edge.

Full suite: 3929 passed, 13 failed -- the same 13 that fail on an unmodified v8 checkout here (terraform, ollama, manifest, install_references), none touched by this change.

Fixes #2584

### DIFF
--- a/graphify/extract.py
+++ b/graphify/extract.py
@@ -1252,7 +1252,10 @@ def _rescue_js_dynamic_imports(path: Path, result: dict) -> None:
     as an ``imports_from`` edge marked ``deferred`` (``_dynamic_import_js``).
     Re-emitting it here as a second ``dynamic_import`` edge would state the
     same fact twice, so a match whose resolved target already has a deferred
-    edge is skipped.
+    edge FROM THIS FILE'S NODE is skipped. The source check matters: the AST
+    pass anchors the edge on the enclosing function when the ``import()`` is
+    written inside one, and that is a different fact from "this file depends on
+    that module" — the only one file-level traversal can use (#2584).
 
     Regex false positives in comments/strings are the precedented trade of
     the Svelte/Vue rescues; a ``//``-prefix guard covers the common case.
@@ -1268,8 +1271,28 @@ def _rescue_js_dynamic_imports(path: Path, result: dict) -> None:
         base_url = _load_tsconfig_base_url(path.parent)
         deferred_ids: set[str] = set()
         deferred_files: set[str] = set()
+        rescued_targets: set[str] = set()
         for e in result.get("edges", []):
-            if e.get("deferred") and e.get("relation") == "imports_from":
+            # Only a FILE-level deferred edge makes the rescue redundant (#2584).
+            #
+            # `_dynamic_import_js` emits `caller_nid -> target`, and `caller_nid` is this
+            # file's node only when the `import()` sits at module scope. Written inside a
+            # function it is that function's node — a different fact, at a granularity
+            # `affected` does not walk. Matching on target alone treated the two as one and
+            # skipped the rescue, so a dynamic import inside a function ended up with no
+            # file-level edge at all. The reverse walk then reached the enclosing function
+            # and stopped: the only edge pointing at it is `contains`, deliberately kept out
+            # of DEFAULT_AFFECTED_RELATIONS.
+            #
+            # Measured on a ~700-file TS repo: `affected --depth 3` returned 39 of 49 truly
+            # affected files (recall 0.80, precision 1.00) and deeper traversal did not help,
+            # which is a dead end rather than a depth limit. It stayed hidden because the
+            # usual case still resolves — when the next importer imports that exact symbol
+            # by name there IS an edge into the function. Switch that importer to
+            # `import * as ns` or a side-effect `import './dyn'` and the same graph goes
+            # silent.
+            if (e.get("deferred") and e.get("relation") == "imports_from"
+                    and e.get("source") == file_node_id):
                 deferred_ids.add(e.get("target"))
                 tf = e.get("target_file")
                 if tf:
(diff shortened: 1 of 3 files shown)


safishamsi:
Shipped in v0.9.39 (`graphifyy==0.9.39`). Re-keying the dynamic-import dedupe on the importing file (not the target alone) is what landed for #2584, so `affected` traverses an in-function dynamic import. Credited in the release notes. Thanks @phudayyy!

PR #2600, opened 2026-08-10
fix(export): portable path in graph.html document title

### LINKED ISSUE #2598: Regression: graphify update writes absolute Windows path into graph.html title
## Regression

Graphifyy 0.9.39 `graphify update <repo>` writes the generator host absolute path into the tracked HTML title:

```html
<title>graphify - C:\Users\<user>\...\graphify-out\graph.html</title>
```

This reproduces the privacy/portability problem previously tracked by #433. The same repository previously contained the relative title `graphify-out\graph.html`.

## Expected

Use a repository-relative label (for example `graphify-out/graph.html`) or the repository name. Generated tracked artifacts must not expose the host user/home path.

## Environment

- graphifyy 0.9.39
- Windows
- incremental `graphify update` against an existing tracked graph

The downstream artifact was manually sanitized before commit.

michaelxer:
## Problem

`graphify update` writes the generator host absolute path into the tracked `graph.html` `<title>`:

```html
<title>graphify - C:\Users\<user>\...\graphify-out\graph.html</title>
```

That reintroduces the privacy/portability issue previously fixed for other artifacts in #433 (reported again as #2598 on Windows with 0.9.39).

## Solution

`to_html()` now builds the document title through `_html_document_title()`:

1. Prefer a path relative to the process cwd when the output file sits under it
2. Otherwise keep from the configured output-dir bare name (`graphify-out` / `GRAPHIFY_OUT` basename) onward
3. Finally fall back to the filename only

Drive letters and host user segments are stripped so Windows absolute paths never land in the title.

## Verification

```bash
python -m pytest tests/test_export.py -q -k 'html or title'
# 14 passed locally

# Focused repro:
# before: <title>graphify - /tmp/.../Users/mike/proj/graphify-out/graph.html</title>
# after:  <title>graphify - graphify-out/graph.html</title>
```

New tests:
- `test_to_html_title_uses_portable_path_not_host_absolute`
- `test_html_document_title_helper_windows_and_relative`

## Notes

- Related but separate from open PR #778 (manifest.json / GRAPH_REPORT.md relative paths). This change is only the HTML document title path.
- I did not run a full Windows end-to-end `graphify update` install; unit coverage includes Windows-style absolute path strings.

Fixes #2598

### DIFF
--- a/graphify/exporters/html.py
+++ b/graphify/exporters/html.py
@@ -322,6 +322,44 @@ LEGEND.forEach(c => {{
 }});
 </script>"""
 
+
+def _html_document_title(output_path: str) -> str:
+    """Return a portable label for the graph.html <title>.
+
+    Tracked artifacts must not embed the generator host absolute path
+    (regression of #433; reported again as #2598 on Windows). Prefer a
+    path relative to the process cwd; otherwise keep from the configured
+    output-dir bare name (``graphify-out`` / ``GRAPHIFY_OUT`` basename)
+    onward; finally fall back to the filename only.
+    """
+    from graphify.paths import GRAPHIFY_OUT_NAME
+
+    raw = str(output_path).replace("\\", "/")
+    # Drop Windows drive prefix so Path parts are comparable on any OS.
+    if len(raw) >= 3 and raw[1] == ":" and raw[0].isalpha() and raw[2] == "/":
+        raw = raw[2:]  # "/Users/..." style after drive strip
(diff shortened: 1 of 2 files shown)


safishamsi:
Shipped in v0.9.40 (`graphifyy==0.9.40` on PyPI). Portable graph.html title for #2598, with a follow-up so it keeps from the output-dir marker onward. Credited in the release notes. Thanks @michaelxer!

PR #2609, opened 2026-08-11
fix(go): resolve qualified symbols by import path

### LINKED ISSUE #1313: Go AST: cross-package qualified calls (pkg.Func()) never produce EXTRACTED edges
## Problem

The Go AST extractor in v0.8.39 does not produce any `EXTRACTED` `calls` edges for cross-package qualified calls (e.g. `api.NewHandlers(...)` from `cmd/bridge/main.go`). All such calls end up as `INFERRED` edges contributed by the semantic subagent. Intra-package call resolution works fine.

This is adjacent to closed issues #85 (cross-file within a package) and #298 (call resolution beyond Python imports), but specifically about Go cross-package qualified-identifier calls.

## Concrete numbers from a 356-file Go+TS corpus

Built with v0.8.39, default mode (no `--mode deep`):

| Scope | EXTRACTED | INFERRED |
|---|---|---|
| Go server intra-package `calls` | **205** | 352 |
| Go server CROSS-package `calls` | **0** | 132 |

Top INFERRED cross-package pairs (semantic recovered them; AST did not):

```
23  api → health        14  api → stream
17  origins → terminal  11  browser → stream
10  stream → terminal    7  api → origins
 7  switching → api      6  api → browser
 4  bridge → stream      4  browser → api
```

Manual audit of all 52 INFERRED edges incident on `api.NewHandlers()` (called from `cmd/bridge/main.go` and 51 test functions across 5 files) confirmed **52/52 are real calls** — zero false positives, zero false negatives. The semantic subagent perfectly compensates, but the AST signal alone would report `NewHandlers()` as a low-confidence node.

## Minimal repro

```
mymod/
├── go.mod                       (module example.com/mymod)
├── cmd/app/main.go              imports "example.com/mymod/internal/api"
│                                 body: api.NewHandlers()
└── internal/api/handlers.go     func NewHandlers() *Handlers { ... }
```

Expected: one `EXTRACTED` `calls` edge from `main` → `NewHandlers`.
Actual (v0.8.39): no AST edge; semantic subagent emits it as `INFERRED 0.8`.

## Impact

1. **High-betweenness Go nodes look under-confident.** `NewHandlers()` (54 edges total) shows up in suggested-questions as needing verification because all 52 of its callers are INFERRED.
2. **God-node detection is correct in topology but penalized in label.**
3. **Without \`GEMINI_API_KEY\` set, the semantic step costs ~10x what AST would,** so the all-INFERRED workaround has a real token bill on Go-heavy repos.

## Suggested fix

Mirror the per-package resolver from #85's suggestion:

1. Per-file pass extracts symbols as today.
2. Per-package pass merges the \`label_to_nid\` map for files sharing a Go package.
3. **Cross-package pass:** for each \`selectorExpr\` of the form \`ident.Func\`, look up \`ident\` in the file's \`import\` block, map it to the imported package path, and resolve \`Func\` against that package's symbol map.

Step 3 is the missing piece; the closed #298 fix appears to have covered Python import semantics only.

## Environment

- graphify v0.8.39 (pipx)
- Go 1.24.0 (toolchain go1.24.11)
- macOS Darwin 25.5.0, Python 3.14
- tree-sitter-go via bundled parser

### LINKED ISSUE #2608: Go qualified types lose import qualifier and create phantom references
## Problem

The Go extractor drops the package qualifier from `qualified_type` nodes before emitting type-reference stubs. For example, `*testing.T` becomes a bare `T` stub. The corpus-level unique-stub rewire can then attach that stub to an unrelated local type or top-level function named `T`, manufacturing a high-degree phantom node.

This is the type-reference counterpart of #1313: Go import statements contain enough evidence to distinguish internal and external packages, but the qualifier is discarded before resolution.

## Minimal repro

```text
repro/
├── go.mod                 module example.com/repro
├── translate.go           func T(message string) string { return message }
└── worker_test.go         func TestBuild(t *testing.T) {}
```

```go
// worker_test.go
package repro

import "testing"

func TestBuild(t *testing.T) {}
```

Expected:

- `TestBuild` references an external `testing.T` node.
- No edge connects `TestBuild` to the local `T()` function.

Actual on `v8` / graphifyy 0.9.39:

- The qualified type is reduced to bare `T`.
- The `references` edge is rewired to the unrelated local `T()` definition.

## Real-corpus impact

On a Go service corpus, a local translation function named `T()` became a god node with 4,349 edges; 4,309 neighboring nodes came from tests using `*testing.T` or suite `.T()`. Ignoring `*_test.go` reduced the symptom, but it also removed legitimate test architecture from the graph and did not fix qualifier loss.

Common external types and aliases can trigger the same class of collision anywhere a matching bare symbol exists.

## Suggested fix

1. Preserve `alias.Type` in the Go extractor instead of reducing it to the last segment.
2. Retain the file's alias-to-import-path map.
3. Before generic stub rewiring, resolve internal types only when the candidate's canonical Go package path exactly matches the import path.
4. Park unresolved external types on an import-path-qualified sourceless node such as `testing.T`, so they cannot collide with local bare symbols.
5. Preserve the same behavior during incremental updates by consulting unchanged resolution-context nodes and edges.

## Related

- #1313 — cross-package qualified Go calls are not resolved from import evidence.
- #598 / #599 — phantom god nodes from member-call name collisions.
- #991 — common-name collisions pollute god-node ranking and community detection.

Investigation and minimal reproduction prepared with OpenAI Codex.

gnukeno:
## Summary

- preserve Go package qualifiers and alias-to-import-path evidence during AST extraction
- resolve `pkg.Func()` only against definitions in the exact imported internal package, while leaving external calls unresolved instead of matching a bare name
- resolve internal `pkg.Type` references exactly and canonicalize external types to import-path-qualified sourceless nodes such as `testing.T`
- keep qualified call/type resolution working when the target is supplied through incremental-update resolution context

## Problem

The Go extractor recognized imported selectors but discarded the receiver/import path before the shared resolver ran. A call such as `errors.New(...)` was therefore treated as bare `New` and could bind to an unrelated local constructor. In one real corpus, 605 of 606 inferred callers attached to a local `New()` were calls to other packages.

Qualified types had the same loss of identity: `*testing.T` became bare `T`, after which the generic stub rewire could attach it to an unrelated local `T()` function. In the observed corpus this produced a 4,349-edge phantom god node, with 4,309 test neighbors.

This patch uses the nearest `go.mod` module directive plus the definition's directory to compute its canonical package path. Imported selectors/types resolve only on an exact path match. Missing `go.mod`, external packages, and ambiguous internal candidates fail closed.

## Tests

- external `errors.New` does not bind to local `New`
- aliased internal `maker.New` resolves to the exact imported package despite a same-name decoy
- external `testing.T` does not bind to local `T()`
- aliased internal qualified types resolve to the exact definition
- incremental extraction resolves qualified calls and types into unchanged context
- same external qualified type references canonicalize to one import-path node
- existing Go builtin and qualified-resolution coverage

Closes #1313.
Closes #2608.

Investigation, implementation, and regression fixtures prepared with OpenAI Codex.

### DIFF
--- a/graphify/extract.py
+++ b/graphify/extract.py
@@ -79,6 +79,7 @@ from graphify.extractors.resolution import (  # noqa: E402,F401
     _decldef_class_stem,
     _disambiguate_colliding_node_ids,
     _find_workspace_root,
+    _go_import_path_for_file,
     _is_type_like_definition,
     _js_call_identifier,
     _js_default_export_name,
@@ -116,6 +117,7 @@ from graphify.extractors.resolution import (  # noqa: E402,F401
     _resolve_cross_file_imports,
     _resolve_cross_file_java_imports,
     _resolve_export_target,
+    _resolve_go_type_references,
     _resolve_java_type_references,
     _resolve_php_type_references,
     _resolve_js_import_path,
@@ -5951,6 +5953,21 @@ def extract(
         except Exception as exc:
             import logging
             logging.getLogger(__name__).warning("Java type-reference resolution failed, skipping: %s", exc)
+    # Resolve internal Go pkg.Type references exactly and park external ones
+    # before the generic bare-label stub rewire can manufacture a collision.
+    _go_sel = [(r, p) for r, p in zip(per_file, paths) if p.suffix == ".go"]
+    if _go_sel:
+        try:
+            _resolve_go_type_references(
+                [r for r, _ in _go_sel], [p for _, p in _go_sel],
+                all_nodes, all_edges, root,
+                resolution_context_nodes, resolution_context_edges,
+            )
+        except Exception as exc:
+            import logging
+            logging.getLogger(__name__).warning(
+                "Go type-reference resolution failed, skipping: %s", exc
+            )
     _rewire_unique_stub_nodes(all_nodes, all_edges)
 
     # Add cross-file class-level edges (Python only - uses Python parser internally)
@@ -6168,6 +6185,7 @@ def extract(
     # file is real ONLY if the caller imported it. So a cross-file call from one
     # of these files with no import evidence is gated below (#1659).
     _JS_TS_CALL_SUFFIXES = (".ts", ".tsx", ".mts", ".cts", ".js", ".jsx", ".mjs", ".cjs")
+    _go_module_cache: dict[Path, str | None] = {}
     for rc in all_raw_calls:
         callee = rc.get("callee", "")
         if not callee:
@@ -6225,6 +6243,20 @@ def extract(
             ]
             if not candidates:
                 continue
+        # Imported Go selectors carry exact package evidence. External package
+        # calls yield no internal candidate instead of binding by bare name.
+        go_exact_import = False
+        if rc.get("language") == "go" and rc.get("import_path"):
+            import_path = str(rc["import_path"])
+            candidates = [
+                candidate for candidate in candidates
+                if _go_import_path_for_file(
+                    nid_to_source_file.get(candidate, ""), root, _go_module_cache
+                ) == import_path
+            ]
+            if not candidates:
+                continue
+            go_exact_import = True
         caller = rc["caller_nid"]
         # Resolve the caller's file via the raw_call's own source_file string,
         # which is stable regardless of any caller_nid remap. An indirect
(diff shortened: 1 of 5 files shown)


safishamsi:
Shipped in v0.9.40 (`graphifyy==0.9.40` on PyPI). Go qualified-symbol resolution by import path for #2608. Credited in the release notes. Thanks @gnukeno!

PR #2611, opened 2026-08-11
 fix(python): avoid crash resolving overdeep relative imports ( #2605 )

### LINKED ISSUE #2605: Python extraction crashes on over-deep relative import resolving to filesystem root
## Summary

Graphify 0.9.39 crashes during Python AST extraction when a relative import climbs to the filesystem root and the candidate path has no name.

The failure is reproducible from the public Ruff TID252 fixture, which contains a 27-dot relative import.

## Reproduction

Environment:

- graphifyy 0.9.39
- CPython 3.14.0
- code-only extraction
- exact Ruff fixture SHA-256: e660e8da415dca3af859142cde9dab8d13072ff470921c37d1f94695ed5f1193

Command:

    graphify extract ruff-exact --code-only --force

Observed result:

    AST extraction failed: PosixPath("/") has an empty name

Exit code: 1.

A reduced fixture containing only the same 27-dot import fails identically. Python parses that import with level=27 and module=None.

A nearby level-1 relative-import control succeeds and produces a graph with 4 nodes and 2 edges.

## Cause

The Python module resolver reaches Path("/") and calls candidate.with_suffix(".py") without guarding candidate.name. Path("/").with_suffix(".py") raises the observed ValueError.

## Expected behavior

An over-deep relative import should be ignored or represented as unresolved without aborting extraction of the entire source tree.

## Suggested narrow fix

Before calling with_suffix, return no candidate when candidate.name is empty. This avoids changing Graphify behavior for legitimate named out-of-root relative imports.

## Validation requested

- hostile 27-dot relative-import regression
- ordinary relative-import passing control
- existing out-of-root named import tests remain green
- full Ruff code-only extraction completes and writes a graph

The reproduction ran in a disposable isolated environment with a 60-second wall limit, 30-second CPU limit, 2 GiB process-group RSS limit, and 10 MiB log limit.

SinghAman21:
Changed `graphify/extractors/resolution.py:1727` so `_probe_python_module_candidate()` returns `None`  when the candidate path has no name before calling `with_suffix(".py")` . That
  makes an over-deep relative import unresolved instead of crashing extraction.

  Added regression coverage in `tests/test_python_import_resolution.py:34:`

  - 27-dot relative import does not crash and still extracts the file/function.
  - normal `from .sibling import helper` still resolves.

  Validation run:

  - `uv run pytest tests/test_python_import_resolution.py tests/test_src_layout_import_resolution.py tests/
    test_extract.py::test_python_relative_import_out_of_root_target_id_is_portable tests/test_incremental.py::test_incremental_python_relative_import_target_canonicalizes -q`

  - Result: 10 passed
  - Reduced CLI repro: graphify extract /tmp/graphify-overdeep-relative --code-only --force ... completed and wrote graph.json.
  - Ran uv run graphify update . as required.

fixes #2605

### DIFF
--- a/graphify/extractors/resolution.py
+++ b/graphify/extractors/resolution.py
@@ -1724,6 +1724,8 @@ def _probe_python_module_candidate(candidate: Path) -> Path | None:
             return init_path
     if candidate.is_file():
         return candidate
+    if not candidate.name:
+        return None
     py_candidate = candidate.with_suffix(".py")
     if py_candidate.is_file():
         return py_candidate
--- a/tests/test_python_import_resolution.py
+++ b/tests/test_python_import_resolution.py
@@ -3,6 +3,7 @@ from __future__ import annotations
 from pathlib import Path
 
 from graphify.extract import extract
+from graphify.extractors.resolution import _resolve_python_module_path
 
 
 def _write(path: Path, text: str) -> Path:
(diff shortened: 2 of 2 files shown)


safishamsi:
Shipped in v0.9.40 (`graphifyy==0.9.40` on PyPI). Over-deep relative-import crash guard for #2605. Credited in the release notes. Thanks @SinghAman21!

PR #2613, opened 2026-08-11
fix(cache): guard the stat fastpath against same-tick same-size rewrites

### LINKED ISSUE #2612: `file_hash()` stat fastpath returns a STALE digest when a file is rewritten with same-length content inside one mtime tick
**Labels:** bug, cache, correctness
**Environment:** Windows 11, Python 3.13.6, NTFS

## Summary

`graphify/cache.py:file_hash()` uses a `(size, mtime_ns)` stat fastpath to skip re-reading a
file. On filesystems whose timestamp granularity is coarser than the interval between two
writes (NTFS updates mtime on a ~15.6 ms system tick), a file that is rewritten with
**different content of the same byte length** keeps the same `(size, mtime_ns)`. The memo
therefore returns the digest of the **previous** content.

Because `file_hash()` is the identity used for semantic-cache keys and for incremental
change detection, a stale digest means graphify believes changed content is unchanged.

This is not theoretical: it makes a shipped test fail roughly half the time (see
"Collateral evidence").

## Steps to reproduce

```python
# hashrepro.py
import pathlib, tempfile, hashlib
from graphify.cache import file_hash

root = pathlib.Path(tempfile.mkdtemp())
f = root / "mod.py"

collisions = 0
for i in range(20):
    f.write_text("x = 1  # aaa\n")      # 13 bytes
    h1 = file_hash(f, root)
    f.write_text("x = 2  # bbb\n")      # 13 bytes, DIFFERENT content
    h2 = file_hash(f, root)
    if h1 == h2:
        collisions += 1

print(f"stale digest returned in {collisions}/20 trials")
```

```
$ python hashrepro.py
stale digest returned in 19/20 trials
```

Each trial also confirms `st1.st_size == st2.st_size and st1.st_mtime_ns == st2.st_mtime_ns`
is `True`, which is what lets the memo fire.

## Actual behavior

`file_hash()` returns the digest computed from the *old* content. Downstream:

- **Semantic cache**: `check_semantic_cache()` keys on this digest, so an LLM extraction
  produced for the old content is served for the new content. The graph silently reflects
  code that is no longer on disk, and no re-billing/re-extraction is triggered to correct it.
- **Incremental update**: the manifest records the file as unchanged, so `graphify update`
  skips re-extracting it.
- **`prune_semantic_cache()`**: two different contents collapse to one cache entry, so the
  orphan that should be swept is indistinguishable from the live entry.

## Expected behavior

`file_hash()` is documented as "SHA256 of file contents + path relative to root". Two
different file contents must never yield the same digest. The stat fastpath is an
optimisation and must not be able to return a digest for content that is no longer present.

## Root cause

`graphify/cache.py:363-378`:

```python
st = p.stat()
entry = _stat_index.get(abs_key)
if (isinstance(entry, dict)
        and entry.get("size") == st.st_size
        and entry.get("mtime_ns") == st.st_mtime_ns):
    hashes = entry.get("hashes")
    if isinstance(hashes, dict):
        cached = hashes.get(salt)
        if isinstance(cached, str):
            return cached          # <-- stale when mtime granularity > write interval
```

`(size, mtime_ns)` is only a safe change-detector when the clock resolution is finer than
the shortest possible interval between two writes. That does not hold on NTFS, and is not
guaranteed on ext4/APFS either for writes inside the same nanosecond-truncated tick.

## Impact / realistic triggers

The dangerous pattern is *"a file is hashed, then rewritten at the same length before the
clock ticks"*, inside one process. Realistic cases:

- `graphify watch` is a long-lived process that hashes on every rebuild. Two saves in quick
  succession (editor autosave, format-on-save followed by a manual save) with the same
  length leaves the second one invisible.
- Formatters, codemods, and `sed -i` sweeps rewrite many files in a burst; length-preserving
  edits are common (`x = 1` → `x = 2`, `<` → `>`, renaming to a same-length identifier).
- Any single-process flow that hashes a file before and after a rewrite.

I did **not** observe the stale digest surviving into a *later, separate* process in my
testing — a fresh run recomputed correctly. So the current blast radius is per-process, not
a permanently poisoned on-disk index. That still covers `watch` and any single long run.

## Collateral evidence

`tests/test_cache.py::test_semantic_prune_sweeps_both_namespaces_against_same_live_set`
fails ~50% of the time in isolation because of exactly this:

```
$ for i in $(seq 8); do pytest "tests/test_cache.py::test_semantic_prune_sweeps_both_namespaces_against_same_live_set" -q --tb=no -p no:randomly | tail -1; done
1 passed
1 failed
1 failed
1 passed
1 failed
1 failed
1 failed
1 passed
```

The test writes `"# A\n\nContent A.\n"` then `"# B\n\nContent B.\n"` — same length. When
`h_old == h_live` the two `save_semantic_cache` calls write to one file, and
`prune_semantic_cache` returns `0` instead of `2`.

## Suggested fix

Make the fastpath refuse to trust an entry whose recorded mtime is too close to "now" to be
distinguishable — the standard trick used by build systems (git's racily-clean handling,
ccache, Bazel):

```python
if (isinstance(entry, dict)
        and entry.get("size") == st.st_size
        and entry.get("mtime_ns") == st.st_mtime_ns
        # racily-clean guard: a file whose mtime is within one filesystem tick of
        # the index write cannot be proven unchanged by stat alone -- re-read it.
        and st.st_mtime_ns < entry.get("indexed_at_ns", 0) - _MTIME_GRANULARITY_NS):
```

recording `indexed_at_ns = time.time_ns()` alongside `size`/`mtime_ns` when the entry is
stored. `_MTIME_GRANULARITY_NS` can be probed once per cache root (write two files, diff the
mtimes) or conservatively fixed at ~20 ms on Windows / ~1 ms elsewhere.

A smaller alternative, if the extra field is unwelcome: also record and compare
`st.st_ctime_ns`, which NTFS updates on write — though this weakens the fastpath on
filesystems where ctime tracks metadata only.

## Related

- Not a duplicate of #2314, #2259, or #1989 — those concern *which path* is used as the cache
  key. This is the digest itself being wrong for a known path.

rajarshidattapy:
Fixes #2612.

## Problem

`file_hash()` skips re-reading a file when `(size, mtime_ns)` is unchanged. That signature
only proves a file is unchanged if the clock that stamped mtime is finer-grained than the
interval between two writes — which is false almost everywhere. NTFS advances mtime on the
~15.6 ms system tick, FAT/exFAT on 2 s, and Linux stamps from the coarse jiffies clock even
though ext4 stores nanoseconds.

So a file rewritten with **different content of the same byte length** inside one tick keeps
an identical signature, and the memo returns the digest of the **previous** content. Since
that digest is the semantic-cache key and the incremental-update change detector, graphify
serves an LLM extraction for code that is no longer on disk, and `graphify update` skips
re-extracting the file.

From the issue's repro, on Windows/NTFS:

```
file_hash returned the SAME digest for different content in 19/20 trials
```

## Fix

A racily-clean guard, the same rule git uses for its index.

Entries now also record `indexed_at_ns` — the wall clock captured **before** the content read
— and are trusted only when `mtime + granularity <= indexed_at`, i.e. the file's mtime tick
had already closed by the time we read it. Any write after that necessarily lands in a new
tick and so moves mtime, making it visible to the next comparison.

Granularity defaults to a conservative **2 s** (covers NTFS, FAT, NFS, coarse-jiffies Linux)
and is tunable with `GRAPHIFY_MTIME_GRANULARITY_MS`; `0` disables the guard and restores the
previous behaviour.

This costs nothing in practice: only files modified within the last 2 s lose the fastpath,
and in a real corpus those are exactly the handful of changed files that have to be read
anyway.

### Fixed at the shared point, not just the reported one

`cached_word_count()` carried a verbatim copy of the same stale-signature logic, so it had
the same bug (a stale `word_count` for a same-length rewrite). Both call sites now route
through two helpers, which also collapses the duplicated store branches:

- `_stat_sig_fresh(entry, st)` — the trust predicate
- `_stat_entry_for(abs_key, st, observed_at_ns)` — get-or-reset the entry and stamp it,
  preserving co-located values (other salts' digests, `word_count`) when the signature still
  matches and resetting them when it does not

## Verification

| Check | Result |
|---|---|
| Issue repro (`hashrepro.py`) | 19/20 stale → **0/20** |
| `test_semantic_prune_sweeps_both_namespaces_against_same_live_set` (flaky ~50%) | **10/10 passing** |
| Full suite | **41 failures — identical to the pre-existing baseline on this platform**, zero new, +2 tests |
| New regression test under `GRAPHIFY_MTIME_GRANULARITY_MS=0` | **fails**, so it genuinely bites |
| Fastpath still alive: 200-file settled corpus, second pass | **0 content reads** |

The two new tests pin both halves, because a fix that quietly disabled the cache would also
make the repro pass:

- `test_file_hash_detects_same_size_rewrite_within_one_mtime_tick` — forces the collision with
  `utime` rather than racing for it, so it is deterministic on fine-grained filesystems too
  (otherwise the two writes land in different ticks and the test goes vacuous)
- `test_file_hash_fastpath_still_serves_a_settled_file` — asserts zero re-reads once the tick
  has closed

## Behavioural changes reviewers should weigh

**Four existing tests were updated.** Flagging this explicitly rather than letting it read as
incidental churn — editing tests to fit a patch is how a real regression hides.

`test_word_count_cached_until_file_changes` and three in `test_stat_index_portability.py`
asserted "unchanged file → zero reads" on files written microseconds earlier. That
expectation is now unsound by design, so they call a `_settle()` helper (backdate mtime past
the granularity window) before the warm-hit probe. Each test's actual subject — portability,
corpus move, out-of-root keys — is untouched; only the probe changed.

**Legacy indexes cost one re-read per file** on the first run after upgrade, then self-heal
with a stamp. Rather than drop the old `reads == 0` assertion in
`test_legacy_absolute_index_migrates_gracefully`, it now pins the one-time cost precisely
(`reads == 1`, then warm, then asserts the flushed entry carries `indexed_at_ns`).

## Known residual (pre-existing, not closable this way)

If a tool forges an **older** mtime onto new content (`rsync -t`, `cp -p`,
`copytree(copy_function=copy2)`), no stat-based cache can detect the change. That is
unchanged by this PR, and `test_cache_hits_survive_corpus_move` deliberately relies on that
trust. Only content hashing would close it, which defeats the cache.

## Files

```
graphify/cache.py                    | 140 ++++++++++++++++++++++++++---------
tests/test_cache.py                  |  61 +++++++++++++++
tests/test_stat_index_portability.py |  40 ++++++++--
tests/test_word_count_cache.py       |  15 ++++
```

No public API or on-disk format break: `indexed_at_ns` is an additive field, and an index
written by this version stays readable by an older graphify (which ignores the field).

### DIFF
--- a/graphify/cache.py
+++ b/graphify/cache.py
@@ -7,6 +7,7 @@ import json
 import os
 import re
 import tempfile
+import time
 import warnings
 from collections.abc import Iterable
 from pathlib import Path
@@ -178,11 +179,12 @@ def _body_content(content: bytes) -> bytes:
     return text[closer.start() + 3:].encode()
 
 
-# Stat-based index: maps absolute path → {size, mtime_ns, hash}.
+# Stat-based index: maps absolute path → {size, mtime_ns, indexed_at_ns, ...}.
 # Loaded once per process, flushed via atexit. Skips full file reads when
 # size+mtime_ns are unchanged — same trade-off as make(1).
-# Correctness risks: `touch` causes a harmless extra re-hash; same-size edits
-# within NFS second-resolution mtime have a 1-second window (same as make).
+# Correctness risks: `touch` causes a harmless extra re-hash. Same-size edits
+# inside one mtime tick used to return the PREVIOUS content's digest; the
+# racily-clean guard below closes that hole (see _stat_sig_fresh).
 # `graphify extract --force` / `graphify update --force` (or GRAPHIFY_FORCE=1)
 # skip the cache reads and re-dispatch everything when needed (#1894).
 _stat_index: dict[str, dict] = {}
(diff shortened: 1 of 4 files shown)


safishamsi:
Shipped in v0.9.40 (`graphifyy==0.9.40` on PyPI). Racily-clean fastpath guard for #2612. Credited in the release notes. Thanks @rajarshidattapy!

PR #2616, opened 2026-08-11
fix(ids): casefold before the non-word filter so normalize_id is idempotent

### LINKED ISSUE #2614: `ids.normalize_id()` is not idempotent for U+0130 (Turkish `İ`), breaking its documented invariant and emitting non-`\w` characters into node IDs
**Labels:** bug, node-identity, unicode

## Summary

`graphify/ids.py` exists specifically to stop node-ID drift — its module docstring names
#811, #550, #1033 and #1104 as the bug class it was created to prevent, and
`normalize_id()`'s own docstring states:

> Idempotent: `normalize_id(normalize_id(s)) == normalize_id(s)`.

That invariant does not hold for any identifier containing **U+0130 LATIN CAPITAL LETTER I
WITH DOT ABOVE** (`İ`) — an extremely common character in Turkish source code
(`İslemYap`, `İçerik`, `İstanbul`).

## Steps to reproduce

```python
from graphify.ids import normalize_id

for s in ["İ", "İstanbul", "fileİname", "naİve_Handler"]:
    a = normalize_id(s)
    b = normalize_id(a)
    print("OK  " if a == b else "FAIL", ascii(s), "->", ascii(a), "->", ascii(b))
```

```
FAIL 'İ'           -> 'i̇'           -> 'i'
FAIL 'İstanbul'    -> 'i̇stanbul'    -> 'i_stanbul'
FAIL 'fileİname'   -> 'filei̇name'   -> 'filei_name'
FAIL 'naİve_Handler' -> 'nai̇ve_handler' -> 'nai_ve_handler'
```

## Actual behavior

- `normalize_id()` is not idempotent: a second application changes the result.
- The output of the *first* application contains **U+0307 COMBINING DOT ABOVE**, which is
  category `Mn` and is *not* matched by `\w`. So `normalize_id()` violates its own
  postcondition that the result contains only word characters and underscores.

## Expected behavior

`normalize_id(normalize_id(s)) == normalize_id(s)` for every input, and the result contains
only `\w` characters and `_`.

## Root cause

`graphify/ids.py:32-40`:

```python
def normalize_id(s: str) -> str:
    s = unicodedata.normalize("NFKC", s)
    s = re.sub(r"[^\w]+", "_", s, flags=re.UNICODE)
    s = re.sub(r"_+", "_", s)
    return s.strip("_").casefold()      # <-- casefold runs AFTER the \w filter
```

`casefold()` is the last step, but it can *introduce* non-`\w` characters:
`"İ".casefold()` is the two-character sequence `"i̇"`. Because the `[^\w]+`
substitution already ran, that combining mark is never filtered — it survives into the
returned ID. On a second pass it *is* filtered, to `_`, producing a different string.

## Impact

`build.py` re-normalizes IDs that `make_id()` already produced, at several points:

- `build.py:1008` — `norm_to_id = {_normalize_id(nid): nid for nid in node_set}`
- `build.py:1102-1104` — edge endpoints resolved through `_normalize_id(src)`
- `build.py:1057-1058` — `if _normalize_id(nid).startswith(new_stem)` where `new_stem` comes
  from `make_id()` and is **not** re-normalized

That last one is the concrete drift: for a Turkish file/symbol, `new_stem` still carries the
raw `i̇` form while `_normalize_id(nid)` has collapsed it to `i_`, so the `startswith`
test fails and the re-key silently does not happen — the exact ghost-node outcome
`ids.py` was introduced to prevent.

It also creates a collision channel: a symbol genuinely named `i_slemYap` and a symbol named
`İslemYap` both normalize to `islem_i_slemyap`, so `norm_to_id` can rewrite an edge endpoint
onto the wrong node.

## Suggested fix

Casefold **before** the character filter, and re-run NFKC afterwards so the whole recipe is a
fixed point:

```python
def normalize_id(s: str) -> str:
    s = unicodedata.normalize("NFKC", s)
    s = s.casefold()                                  # may expand: İ -> i + U+0307
    s = unicodedata.normalize("NFKC", s)              # recompose what casefold produced
    s = re.sub(r"[^\w]+", "_", s, flags=re.UNICODE)   # now nothing non-\w can survive
    s = re.sub(r"_+", "_", s)
    return s.strip("_")
```

Worth adding as a property test, since this class keeps recurring:

```python
@given(st.text())
def test_normalize_id_is_idempotent(s):
    assert normalize_id(normalize_id(s)) == normalize_id(s)
```

Note that any fix changes IDs for affected corpora, so it needs the same legacy-form
migration treatment as #1504 / #1917.

## Related

- #811 (Chinese/Unicode identifiers silently deduplicated) — same bug class, different
  character range.
- #1522 (`normalize_id` collapses non-word sequences to `_`) — adjacent, but about
  collisions rather than idempotency.

rajarshidattapy:
Fixes #2614.

## Problem

`normalize_id()` casefolded **after** the `[^\w]+` filter:

```python
s = unicodedata.normalize("NFKC", s)
s = re.sub(r"[^\w]+", "_", s, flags=re.UNICODE)
s = re.sub(r"_+", "_", s)
return s.strip("_").casefold()      # <-- last
```

Casefolding can *expand* a character into a base letter plus a combining mark:
`"İ".casefold()` is `"i" + U+0307`. With the filter already spent, that mark
survived into the id — so `normalize_id` broke both of its contracts:

- **Not idempotent**, though the docstring promises it: `İstanbul` → `i̇stanbul` → `i_stanbul`.
- **Emitted non-`\w` characters**, though `cache.py` documents that "normalize_id drops every
  non-word character".

Turkish identifiers (`İslemYap`, `İçerik`) hit this on every run, and U+0307 is invisible in
most terminals, so the resulting ids look correct.

The concrete damage is the ghost-node class `ids.py` was created to prevent. `build.py`
re-normalizes ids that `make_id()` already produced, and at `build.py:1057` compares
`_normalize_id(nid).startswith(new_stem)` where `new_stem` comes from `make_id()` and is *not*
re-normalized. For a Turkish symbol the two sides disagreed (`i_slemyap` vs `i̇slemyap`), so
`_semantic_id_remap`'s re-key silently no-opped.

## Fix

Move casefold ahead of the filter, and NFKC again to recompose what it expanded:

```python
s = unicodedata.normalize("NFKC", s)
s = unicodedata.normalize("NFKC", s.casefold())
s = re.sub(r"[^\w]+", "_", s, flags=re.UNICODE)
s = re.sub(r"_+", "_", s)
return s.strip("_")
```

## Blast radius (measured, not assumed)

Scanned all 1.1M codepoints, and diffed ids across the 369-file graphify tree:

| | |
|---|---|
| ASCII codepoints that change | **0** — no churn for ordinary code |
| Codepoints where old emitted non-`\w` | 27 → **0** |
| Codepoints where old was non-idempotent | 1 (U+0130) → **0** |
| Codepoints whose id changes at all | 29 |
| Real tokens in the graphify tree whose id changes | **0** (the 14 hits are the Turkish/Greek fixtures added by this PR) |

The 29 affected codepoints are U+0130, Greek ypogegrammeni, and precomposed
Latin/Greek letters whose casefold expands. For those, the new output is the **composed**
NFKC form — which is what the module docstring already specified.

## No migration needed — and I verified rather than assumed

The issue expected this to need #1504/#1917-style legacy-form handling. It does not, because
**the new output is exactly the old recipe's fixed point**: `normalize_id(old_id) == new_id`.
`build.py:1008` keys its reconciliation map on `_normalize_id(nid)`, so old-form and new-form
ids already collapse to the same key.

Tested end to end rather than argued: built a Turkish corpus with the unfixed code
(`islem_i̇slemyap`), applied the fix, added a symbol, ran `graphify update`:

```
islem
islem_caller
islem_i_kinci      <- new symbol
islem_i_slemyap    <- re-keyed from islem_i̇slemyap
main
main_main
dangling edges: 0
GHOST PAIR PRESENT: False
```

So no legacy-form code was added.

## Tests

**A property test for exactly this already existed and did not catch it.**
`test_property_normalize_id_idempotent` with `@given(st.text())` has been in the suite since
the contract file was written — a uniform draw essentially never produces U+0130 out of ~1.1M
codepoints. Fixing only the implementation would leave that gap open, so this PR also
strengthens the guard:

- deterministic Turkish/case-expanding entries in `CONTRACT_CASES`
- a new `CASE_EXPANDING_CHARS` set covering each distinct failure shape (expansion,
  recomposition, combining marks, `ẞ` → `ss`)
- the missing `\w`-only postcondition, asserted directly
- hypothesis strategies biased toward case-expanding letters and combining marks, so the
  class is covered by construction rather than by luck

New tests produce **18 failures on the unfixed code, 0 with the fix.**

### One assertion hypothesis rejected, worth knowing

I first asserted `normalize_id(s.upper()) == normalize_id(s.lower())`. Hypothesis
immediately found `ı` (U+0131 dotless i): `'ı'.upper()` is `'I'`, which casefolds to `'i'`.
`str.upper()` is locale-independent and lossy, so that property is false regardless of this
fix. Replaced with caseless equivalence via `casefold`, which is the invariant the graph
actually relies on. Noted in the test docstring so nobody "fixes" it back.

## Verification

- Issue repro: all four cases now idempotent
- `tests/test_id_normalization_contract.py`: **71 passed**
- Full suite: **41 failures, identical to the pre-existing baseline on this platform**, zero
  new, **+37 tests**

```
graphify/ids.py                         |  30 ++++++--
tests/test_id_normalization_contract.py | 117 +++++++++++++++++++++++++++++
```

### DIFF
--- a/graphify/ids.py
+++ b/graphify/ids.py
@@ -16,10 +16,19 @@ module exists so the recipe lives in one place and the two callers can no longer
 diverge.
 
 The recipe: NFKC-normalize (so composed/decomposed Unicode forms collapse),
-replace runs of non-word characters with a single underscore (``re.UNICODE`` so
-CJK/Cyrillic/Arabic/accented-Latin letters survive instead of collapsing to a
-per-file node), collapse repeated underscores, strip leading/trailing
-underscores, and casefold.
+casefold, NFKC-normalize again (casefold can *expand* a character into a base
+letter plus a combining mark — ``İ`` -> ``i`` + U+0307 — and the second pass
+recomposes what can be recomposed), replace runs of non-word characters with a
+single underscore (``re.UNICODE`` so CJK/Cyrillic/Arabic/accented-Latin letters
+survive instead of collapsing to a per-file node), collapse repeated
+underscores, then strip leading/trailing underscores.
+
+Casefolding runs BEFORE the non-word filter, not after. With it last, the
+combining marks casefold introduces were never filtered: ``İslemYap`` produced
+``i̇slemyap`` — an id containing U+0307, which is not a ``\\w`` character — and a
+second pass collapsed it to ``i_slemyap``, so the function was not idempotent
+and the builder's re-normalization disagreed with the extractor's ``make_id``
+for any Turkish identifier.
 """
 from __future__ import annotations
 
@@ -32,12 +41,21 @@ __all__ = ["normalize_id", "make_id"]
 def normalize_id(s: str) -> str:
     r"""Normalize a single ID string to its canonical form.
 
-    Idempotent: ``normalize_id(normalize_id(s)) == normalize_id(s)``.
+    Guarantees, both enforced by tests:
+
+    - Idempotent: ``normalize_id(normalize_id(s)) == normalize_id(s)``.
+    - The result contains only ``\w`` characters and ``_``.
+
+    Casefolding before the ``[^\w]+`` filter is what makes both hold — see the
+    module docstring for why the reverse order silently broke them.
     """
     s = unicodedata.normalize("NFKC", s)
+    # casefold can expand one character into a letter + combining mark, so it
+    # must run while the non-word filter can still see the result.
+    s = unicodedata.normalize("NFKC", s.casefold())
     s = re.sub(r"[^\w]+", "_", s, flags=re.UNICODE)
     s = re.sub(r"_+", "_", s)
-    return s.strip("_").casefold()
+    return s.strip("_")
 
 
 def make_id(*parts: str) -> str:
--- a/graphify/ids.py
+++ b/graphify/ids.py
@@ -28,7 +28,7 @@ combining marks casefold introduces were never filtered: ``İslemYap`` produced
 ``i̇slemyap`` — an id containing U+0307, which is not a ``\\w`` character — and a
 second pass collapsed it to ``i_slemyap``, so the function was not idempotent
 and the builder's re-normalization disagreed with the extractor's ``make_id``
-for any Turkish identifier.
+for any Turkish identifier (#2614).
 """
 from __future__ import annotations
 
@@ -47,7 +47,7 @@ def normalize_id(s: str) -> str:
     - The result contains only ``\w`` characters and ``_``.
 
     Casefolding before the ``[^\w]+`` filter is what makes both hold — see the
-    module docstring for why the reverse order silently broke them.
+    module docstring for why the reverse order silently broke them (#2614).
     """
     s = unicodedata.normalize("NFKC", s)
     # casefold can expand one character into a letter + combining mark, so it
--- a/tests/test_id_normalization_contract.py
+++ b/tests/test_id_normalization_contract.py
@@ -39,7 +39,7 @@ CONTRACT_CASES = [
     "x_c1",                       # must NOT be treated as a chunk suffix here
     "__dunder__",                 # leading/trailing underscores stripped
     "tab\tnewline\nspace ",       # whitespace runs -> single underscore
-    # Casefolding these EXPANDS them into a base letter plus a combining
+    # #2614: casefolding these EXPANDS them into a base letter plus a combining
     # mark. With casefold last, the mark landed in the id after the [^\w] filter
     # had already run, so the id carried a non-word character and a second pass
     # changed it. Turkish identifiers are the common real-world case.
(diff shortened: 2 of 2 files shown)


safishamsi:
Shipped in v0.9.40 (`graphifyy==0.9.40` on PyPI). Casefold-before-filter for #2614. Credited in the release notes. Thanks @rajarshidattapy!

Issue #2553, opened 2026-08-08
TS: fabricated EXTRACTED calls edge via name-only type matching — third-party-typed field binds to unrelated same-named local class (v0.9.33/v0.9.35)


Earthfreedom:
## Summary

`graphify extract`'s cross-file TS/JS member-call resolver (`_resolve_typescript_member_calls`) binds `this.field.method()` calls to a same-named class **anywhere else in the corpus by bare type-name text match, with no verification that the type actually comes from that class** (no import-chain check). When a field's declared type name happens to collide with an unrelated, differently-sourced class of the same name elsewhere in the project, `graphify` emits a fabricated `calls` edge between them — at **`confidence: "EXTRACTED"`**, the tool's own highest-confidence tier, normally reserved for edges backed by real import/AST evidence.

This is the same class of bug reported in #631 (claimed fixed via an `is_member_call` guard in v0.5.6, closed) and reported again as a regression in #2041 (still open). Current source inspection (both v0.9.33 and v0.9.35) shows the specific case those issues describe — a call to a JS/TS **built-in global** (e.g. `console.error(...)`) matching an unrelated local method of the same name — is now explicitly guarded against via a hardcoded `_LANGUAGE_BUILTIN_GLOBALS` allowlist (see "Note" below). However, that fix only covers the ~60 names on the hardcoded list. **Any type name that is not a JS/TS/Python/Swift built-in — including any type imported from a third-party package — is not covered**, and the underlying defect (name-only matching, no import verification) still produces false edges. This report reproduces the surviving general case on **v0.9.35**, the latest release.

**Note on the reproduction below**: the literal "built-in global" scenario from #631/#2041 (`console.error()` colliding with an unrelated `.error()` method) was tried first and does **not** reproduce on v0.9.33 or v0.9.35 — see the "Builtin-global control check" section. The repro below instead uses a constructor-injected field typed via a third-party import, which is the officially documented resolution path for this resolver (see its own docstring, quoted below) and demonstrates the same root cause is still live for anything off the hardcoded allowlist.

## Environment

- `graphifyy` version: 0.9.33 and 0.9.35 (both tested; identical behavior)
- Python: 3.12.13
- OS: Linux (`python:3.12-slim` Docker container; host OS macOS 26.5.1/Docker Desktop, irrelevant to the bug)
- Invocation: `graphify extract <path> --code-only --no-cluster`

## Minimal reproduction

[...]

Issue #2507, opened 2026-08-06
query: relational-intent verbs ("calls"/"uses") survive stopwording, prefix-match unrelated identifiers, and the #1445 per-term guarantee seats the junk match as a seed


filipechagas:
### Summary

In a relational natural-language query — "Who **calls** X?", "what code **uses** X?" — the relational verb survives stopword filtering, prefix-matches an unrelated camelCase identifier (`calls` → `.callStoreWithAmount()`), and the per-term guaranteed-seed mechanism from #1445 then seats that junk match as a BFS root regardless of how well the real identifier scored. The traversal floods the token budget with the junk seed's neighborhood (typically test files), and the answer contains zero of the correct nodes while looking authoritative.

The sharp part: `_infer_context_filters` **already recognizes the verb as relational intent** — the same query emits `Context: call (heuristic)` via `_CONTEXT_HINTS` (serve.py:725) — but that recognition only filters traversal edges. It never reaches `_query_terms` or `_pick_seeds`, so one word is simultaneously treated as an edge-filter directive and as an identifier competing for (and winning) a guaranteed seed slot.

### Measured failure (real corpus)

Observed on a ~46k-node / ~139k-edge graph of a production PHP monolith (graphifyy 0.9.28 baseline; identically on a 0.9.34-based build with richer `calls` edges — the failure predates and survives both). Ground truth, grep-verified: `ChargeCustomerService` has exactly 3 production callers.

| Phrasing | Seeds chosen | Result |
|---|---|---|
| "Who calls ChargeCustomerService?" | `['ChargeCustomerService', '.callStoreWithAmount()']` + `Context: call (heuristic)` | **0/3 callers**; 15 nodes, all test noise from the junk seed's neighborhood, ~1,583 tok |
| "callers of ChargeCustomerService" | `['ChargeCustomerService', '.it_ignores_initiator_for_non_ops_callers()']` | 3/3 present but buried in 42 shown of 147 nodes, ~1,617 tok |
| "what code uses ChargeCustomerService to charge a customer" | 5 seeds incl. `.charge()`, `coder.md`, a test method | **0/3**; 2,307-node explosion, 52 shown, ~1,600 tok |

[...]

PR #2617, opened 2026-08-11
fix(php): consolidated PHP support — member-call resolution, use-import handling, node identity, cross-language isolation, query seeding (#1682, #2615)

### LINKED ISSUE #1682: PHP: member/instance method calls never resolve to calls edges — only static Class::method() works

### LINKED ISSUE #2615: PHP support, consolidated: member-call resolution, use-import handling, node identity, cross-language isolation, query seeding (superset of #1682)
## Summary

This issue is a consolidated defect report for PHP support in graphify. It extends #1682 and adds every related defect we found while we fixed #1682. We wrote the fixes in our public fork, [lawnstarter/graphify](https://github.com/lawnstarter/graphify), and we validated them on a production Laravel codebase.

This issue supersedes our seven open PRs: #2492, #2502, #2503, #2505, #2506, #2516, #2536. Seven stacked PRs are hard to review and hard to keep current. We close all seven. One new PR replaces them and fixes this issue.

## Test corpus

Almost all measurements below come from one codebase: `api.lawnstarter.com`. It is a PHP 8.3 / Laravel monolith. Its graph has 46,406 nodes (48,500 nodes in a later snapshot, used for the query-seeding measurements). We call this "the corpus" below. All numbers are from real builds of this corpus, before and after each fix.

### 1. PHP member calls resolve to no `calls` edge (#1682)

The extractor emits `calls` edges for `ClassName::method()` only. It emits no edge for:

- `$this->service->method()` through a constructor-injected property. This is the dominant call pattern in Laravel code.
- `$obj->method()` on a typed local variable or a typed parameter.
- `(new Service())->method()`.

On the corpus, `LeadHunterService` has grep-verified callers but zero inbound `calls` edges. The graph looks complete but is not. This is the original report in #1682.

### 2. The extractor discards `use` statement facts

`_import_php` keeps only the trailing short name of a `use` statement (`raw.split("\\")[-1]`). The written FQN, the alias, and the kind (`class` / `function` / `const`) are lost. Downstream passes then cannot know which class a file claims.

Also, tree-sitter-php puts the `function`/`const` keyword in a different position for the group form. The result: `use function A\{render};` claims `render` as a **class** name. A `class I extends Render` in the same file then re-points its `inherits` edge to the wrong target.

### 3. Member-call resolvers and import edges leak across languages

The receiver-type indexes were corpus-wide, not per-language. Two wrong results follow:

- A TypeScript `Lead.search({})` call binds a Python `class Lead`. This is a wrong cross-language edge, at EXTRACTED (1.0) confidence.
- In reverse, a foreign class that only shares a short name pushes the single-definition guard to 2. The correct same-language edge is then silently suppressed.

Separately, an `imports` edge dies when a same-stem foreign file sits beside its target (`lead.md` beside `lead.py`). Id disambiguation salts both file nodes, and the edge dangles. This affects Python, PowerShell, Zig, Rust, Pascal, Elixir, and Bash.

### 4. Wrong or mislabeled member-call edges

- A PHP 8.1 first-class callable, `$obj->method(...)`, creates a `Closure`. It does not invoke the method. It still got a `calls` edge.
- A union-typed receiver (`private Alpha|Beta $svc;`) fell through to the in-file bare-name matcher. It bound whichever same-named method the file held, at EXTRACTED confidence.
- A PHP 8.2 DNF type (`private (A&B)|C $x;`) was not handled at all.
- The extractor flattened written qualified types (`\Vendor\Sdk\Client` became `Client`).
- A `use Vendor\Sdk\Client;` claim was ignored. The resolver bound the one unrelated in-corpus `Client` instead. Measured on the corpus: 3 wrong edges of this class.

### 5. A claimed type name cannot bind to its declared FQN

When several in-corpus classes share the short name `X`, a file with `use App\Alpha\X;` refused instead of selecting `App\Alpha\X`. A renaming alias, `use App\Alpha\X as Y;`, could never resolve, because nothing in the corpus is named `Y`. The bindings must also survive `graphify update`: without persisted declared-FQN facts, every added edge vanishes on the first incremental rebuild.

### 6. PHP node identity: interfaces, traits, and enums mint no node

`_PHP_CONFIG.class_types` held `class_declaration` only. On the corpus, 291 declarations had no node: 142 interfaces, 30 traits, 119 enums. The fan-in scattered onto sourceless stubs, per-file stubs, and file nodes. Consequences:

- A sourceless stub shadows the real name. 18 stubs shared one label on the corpus. `graphify explain` answered from an arbitrary stub. `graphify affected` refused the same name with "No unique node match".
- About 19,000 `imports` edges dangled with no node behind them.
- `parent::setUp()` emitted a raw call with the literal callee name `parent`. The cross-file pass then bound it to any `->parent()` accessor. Measured: 1,698 fabricated inbound `calls` edges on one Eloquent model accessor.
- `graphify query` and `shortest_path` preferred stubs on exact score ties. `shortest_path` returned a false "No directed path found" for a reachable pair.

### 7. A PHP function call binds to a method or a class, cross-file

In PHP, a bare `name(...)` call can only invoke a function. The shared cross-file pass matched by normalized label, which erases the member marker. So Laravel's `event(...)` helper bound to a test method named `event()`. Measured: 848 fabricated inbound `calls` edges on one test method, from 368 distinct files, 308 of them not tests. The same family covered `config()` (290 edges), `dispatch()` (186), and `request()` (15). The class half is real too: `report($e)` bound `class Checkr\Resources\Report` through the case-insensitive fold (7 edges).

### 8. Natural-language query seeding fails on impact questions (follow-up to #2507)

We filed #2507. Upstream fixed one of the three reported failure modes in 0.9.35 (the relational-verb demotion). Two modes remain:

- Class-node stranding. "callers of X" infers a `call` context filter, but a class node owns no call edges. Measured: "callers of ChargeCustomerService" returns 1 node and 0 of its 3 grep-verified production callers, delivered confidently.
- Seed explosion. Generic nouns in the question seed their own hub nodes. Measured: "what code uses ChargeCustomerService to charge a customer" traverses 2,286 nodes and shows 0 of 3 callers.

### 9. A receiver typed by a constructor-body assignment resolves nothing

The pre-promotion Laravel idiom: an untyped property, a `@var` docblock, and a typed constructor parameter that the body assigns across. This typed no receiver. On the corpus, `BalanceCustomerAccountService::handle()` got no `calls` edge to `ChargeCustomerService::handle()`, while its two sibling callers (promoted param, typed method param) resolved fine.

## Aggregate effect of the fixes

- Full corpus rebuild with the final fixes: 1,341 fabricated `calls` edges removed, 2,398 legitimate `calls` edges added (46.6k-node rebuild).
- All three impact questions above now return all 3 grep-verified callers within the default token budget.
- 291 PHP type declarations now have real nodes. `explain`, `affected`, `query`, and `path` agree on which node a name means.

## Superseded PRs

| Old PR | Content | State |
|---|---|---|
| #2492 | PHP member-call resolution from typed receivers (#1682) | closed, replaced |
| #2502 | `use` statement metadata; group-form `use function`/`const` fix | closed, replaced |
| #2503 | Cross-language isolation of resolvers and import edges | closed, replaced |
| #2505 | Member-call follow-ups: FCC, unions, qualified types, use-claim refusal | closed, replaced |
| #2506 | Declared-FQN binding with incremental parity | closed, replaced |
| #2516 | Query seeding: covered-term skip, filter relaxation (#2507) | closed, replaced |
| #2536 | Node identity: interface/trait/enum nodes, stub shadowing, relative scopes | closed, replaced |

The replacement PR also contains fixes that were never proposed upstream, because they depended on the PRs above: the dangling-imports resolution (family 6), the constructor-body receiver (family 9), the function-call refusal (family 7), the interface/enum/trait receiver binding, and the scored-endpoint tie-break (family 6).

## Fix

One PR follows this issue. It is based on the current `v8` head (0 commits behind). Every fix carries tests. The full suite passes on the branch: 4,477 passed, 42 skipped. Full per-fix history is in the fork's tracking issue: [lawnstarter/graphify#56](https://github.com/lawnstarter/graphify/issues/56).

filipechagas:
Fixes #2615. Fixes #1682.

## What this PR is

This PR contains all PHP-support and resolution work from the [`lawnstarter/graphify`](https://github.com/lawnstarter/graphify) fork, in one review unit. It replaces our seven open PRs: #2492, #2502, #2503, #2505, #2506, #2516, #2536. We close those seven with pointers to this PR.

One PR is easier to review than seven stacked PRs. Four of the seven had strict merge-order constraints (2492 → 2502 → 2505 → 2506). This PR has none: it is based on the current `v8` head (`50556ba`, v0.9.39) and is 0 commits behind.

The full defect report, with all corpus measurements, is #2615.

## Contents, in four groups

The CHANGELOG on this branch lists every change under one `## 0.9.40 (unreleased)` section, in the same four groups.

### Group A — PHP member-call resolution, cross-language isolation, import-edge repair
(previously PRs #2492, #2502, #2503, #2505, #2506)

- PHP instance-method calls on a typed receiver now resolve: `$this->prop->method()` (including promoted constructor params), typed locals and params, nullsafe `?->`, and `(new Service())->method()`. Declared types bind at INFERRED 0.8. Inline `new` binds at EXTRACTED 1.0 when the written namespace corroborates the class. This is the #1682 fix.
- Refuse-don't-guess policy: no edge for untyped, union, intersection, or DNF receivers; ambiguous short names; undeclared methods; chained or array receivers; rebound or shadowed locals; anonymous classes; `self`/`static`/`parent`.
- PHP 8.1 first-class callables emit `indirect_call`, not `calls`. The syntax creates a `Closure`; it does not transfer control flow.
- PHP `imports` edges keep the written FQN, alias, and kind (`use_kind` / `alias` / `target_fqn` metadata). Group-form `use function A\{f};` no longer claims `f` as a class name.
- A `use Vendor\Sdk\Client;` claim is honored. A claim that names no in-corpus class refuses (subtractive, `PhpNameResolver`). A claim that matches a declared FQN binds to it, including through renaming aliases (additive). The declared-FQN facts persist in `graph.json` (`_php_class_fqns`), so the bindings survive `graphify update`.
- All member-call resolvers and receiver-type indexes are language-scoped. A TypeScript call can no longer bind a Python class, and a foreign same-name class no longer suppresses the correct edge. One shared `_is_owned_definition` predicate keys off the per-resolver suffix tuple, so registration and scoping cannot drift.
- `imports` edges no longer die beside a same-stem foreign file (Python, Rust, Zig, Elixir, PowerShell, Pascal, Bash).

### Group B — natural-language query seeding
(previously PR #2516; follow-ups to #2507)

- The relational-verb demotion that upstream 0.9.35 shipped is restructured: the vocabulary is derived from `_CONTEXT_HINTS` and `_CONTEXT_FILTER_ALIASES` instead of a hardcoded list, and is a strict superset of it (62 words vs 29). The five upstream #2507 tests pass unchanged.
- A query term that the top-ranked seed's label already covers claims no extra guaranteed seed. Generic nouns stop dragging hub nodes into the traversal.
- A heuristically inferred context filter that reaches nothing beyond the seeds relaxes to an unfiltered traversal, with an explicit header note. An explicit `--context` filter is always honored.

### Group C — PHP node identity and lookup
(previously PR #2536, plus two sibling fixes that PR deferred because they needed #2492/#2502; those dependencies are inside this PR)

- `interface`, `trait`, and `enum` declarations mint canonical nodes, exactly as `class` does.
- Imports resolve from their own `target_fqn` edge metadata, so an import of a name only written as `Foo::class` no longer dangles.
- A sourceless stub can no longer shadow a real declaration in `explain`, and `affected` applies the same rule, so the two commands agree.
- `parent::` / `self::` / `static::` calls refuse instead of naming their scope as the callee.
- A property assigned in the constructor body from a typed parameter types the receiver (the pre-promotion Laravel idiom). The class-wide poisoning discipline applies: any other write to the property refuses the binding.

### Group D — resolution gaps closed on top of the above
(never proposed upstream before; they depended on groups A and C)

- A PHP function call (`name(...)`) never binds to a method or a class-like node cross-file. Only functions are plausible targets at a function-call site. The case-insensitive fallback is retried with the same refusal.
- A receiver typed by an `interface`, `enum`, or `trait` binds to that declaration's own method. Implementations are never guessed. The declared-FQN pre-scan now covers all four declaration kinds, which also closes a full-vs-incremental parity hole.
- `graphify query` and `shortest_path` prefer a sourced node over a sourceless stub on exact score ties, as `explain` and `affected` already do. Tie-break only; no live score changes.

## Measured evidence

All numbers are from real builds of a 46.4k-node (later 48.5k-node) PHP 8.3 / Laravel production corpus. Details and attribution per fix are in #2615 and in the CHANGELOG entries.

| Measurement | Before | After |
|---|---|---|
| `LeadHunterService` inbound `calls` edges (#1682 repro) | 0 | resolved, INFERRED 0.8 |
| Fabricated inbound `calls` on one test method from Laravel's `event()` helper | 848 | ~0 (documented same-file residual) |
| Fabricated inbound `calls` on one model accessor from `parent::` calls | 1,698 | 0 |
| Interface/trait/enum declarations with a real node | 0 of 291 | 291 |
| Dangling `imports` edges | ~19,000 | removed |
| "callers of ChargeCustomerService" (3 grep-verified callers) | 1 node, 0 of 3 | 128 nodes, 3 of 3 |
| "what code uses ChargeCustomerService to charge a customer" | 2,286 nodes, 0 of 3 | 138 nodes, 3 of 3 |
| Full-rebuild `calls` edge delta with the final fixes | — | −1,341 fabricated, +2,398 legitimate |
| Wrong-edge family from ignored `use` claims (lawnstarter#16) | 3 wrong edges | 0 (differential: 3 deleted, 0 added, 0 re-pointed) |

## Behavior changes

1. A same-file PHP call through a typed receiver moves from EXTRACTED to INFERRED (0.8). The old EXTRACTED label came from a bare-name match that could not see the type.
2. A PHP 8.1 first-class callable moves from `calls` to `indirect_call`. Consumers that filter on `relation == "calls"` will not see these edges. `affected` output is unchanged.
3. Methods of a PHP interface, trait, or enum move from file-scoped ids to type-scoped ids, and their labels gain the leading member dot. A full `graphify update .` lands this consistently. A hook-driven incremental rebuild against a pre-fix graph drops (never repoints) stale-id edges until the next full update.
4. `caller` / `callers` queries that strand on a class node now relax to an unfiltered traversal, with the header note `Context: call (heuristic; relaxed — no matches beyond seeds)`.

## Composition with upstream 0.9.36–0.9.39

The fork tracked upstream continuously. Two syncs (0.9.36/0.9.37 and 0.9.38/0.9.39) are merged into this branch, and the compositions were verified:

- `extractors/engine.py`: the member-call defer condition is `_python_defer or _java_defer or _php_defer or (…)`. The Python and PHP disjuncts are mutually exclusive by construction: each reads state assigned only inside its own `config.ts_module` branch.
- `extract.py`, ObjC receiver index: this branch keeps the fork's `_is_owned_definition` gate (strictly stronger than the source-file truthiness test upstream narrowed) AND adds upstream's `not _is_protocol_declaration`. Upstream's three new ObjC suites pass on this composition.
- `serve.py`: upstream's hardcoded relational-verb list is replaced by the derived superset (group B above). This is the one place the branch deliberately restructures upstream code; it absorbs all of upstream's vocabulary and its five tests pass unchanged.
- Upstream's TypeScript origin gate (#2553) and the fork's language-ownership gates compose because both only remove candidates. One fork fixture gained the `import` statement real TypeScript requires.

## Tests

- Full suite on this branch: **4,478 passed, 42 skipped, 0 failed** (with the optional `openai` extra installed; without it, 4 `test_ollama_retry_cap.py` tests fail on import, an environment condition #2516's evidence already documented). One labeling test (`test_label_communities_batches_when_over_batch_size`) is known-flaky — it also flakes at the clean upstream baseline, per #2503's evidence — and passes in isolation and in this run.
- Roughly 8,600 of the ~10,800 added lines are tests: 18 new test files, plus additions to `tests/test_watch.py`. Every positive resolution test carries a same-named decoy asserted to get no edge. Positive scenarios have mirrored full-build and incremental-rebuild assertions.
- Each fix landed on the fork red-first. The per-fix red/green evidence is on the fork PRs, linked from [lawnstarter/graphify#56](https://github.com/lawnstarter/graphify/issues/56).

## Migration

- The version bumps to 0.9.40. This rolls the version-namespaced AST cache, which is required: cache entries are keyed by content hash within the version namespace, so a same-version rebuild replays pre-fix facts.
- PHP corpora (and polyglot corpora) must run `graphify update .` once to shed the fabricated edges and gain the new ones.
- The serve-side changes (group B, and the sourced tie-breaks) need no rebuild.

## Known gaps, all deliberate refusals

- `parent::` / `self::` / `static::` calls resolve to nothing. Binding them needs a base-class-aware pass through `inherits`; a decided spec exists ([lawnstarter/graphify#58](https://github.com/lawnstarter/graphify/issues/58)). We can propose it as a follow-up after this PR.
- A method reached through a `use`d trait, or inherited from a cross-file parent class, gets no edge. Docblock types are not read.
- A `event(...)` call in the same file as a class that declares an `event()` method still binds in-file (documented residual; it never reaches the cross-file pass).
- One suspected defect was investigated and written off with measured probes: [lawnstarter/graphify#55](https://github.com/lawnstarter/graphify/issues/55) (zero corpus occurrences of the precondition; the wrong edge is unreachable end-to-end).

## Shape of the branch

66 commits of fork history (including the two upstream-sync merges) plus one packaging commit (consolidated CHANGELOG section, version 0.9.40, one docs correction). A squash merge gives one clean change if you prefer that. For per-fix review, each superseded PR body remains available, and each commit carries provenance to its fork PR.


filipechagas:
Hey @safishamsi :wave: 

First of all, thank you for maintaining such a great product. We've been using the hell out of it at LS.
Right now, we're keeping our fork up to date with these patches, so we can use Graphify in our php codebases. 
And we want to keep contributing!

What can we do to move these patches forward?

Issue #2446, opened 2026-08-04
build_merge prune_sources silently no-ops on absolute paths and reports 'already clean' (deleted nodes never pruned on Windows)


AI-invest:
## Summary

`build_merge(prune_sources=...)` compares the caller's paths against each node's `source_file`. Node `source_file` is repo-relative POSIX, but callers naturally pass the absolute paths that `detect`/`detect_incremental` and `manifest.json` produce. Unless `root=` is also passed, nothing matches — and the run then prints **"no matching nodes or edges in graph, already clean"**, which reads as confirmation rather than as the no-op it is.

Version: **graphifyy 0.8.36**, Windows 11.

## Verified on a live graph

```
sample node source_file : Dealer-Version-2-/DDS_Demo/SolveOneBoard/OneBoard.c
sample manifest key     : C:\j\Bridge\OurBridge\Dealer-Version-2-\DDS_Demo\SolveOneBoard\OneBoard.c
prune entry normalized  : C:/j/Bridge/OurBridge/Dealer-Version-2-/DDS_Demo/SolveOneBoard/OneBoard.c
would it match a node?  : False
```

The normalization is `_norm_source_file` in `build.py`:

```python
p = p.replace("\\", "/")
if root and os.path.isabs(p):          # <-- only relativizes when root is given
    try:
        p = Path(p).relative_to(root).as_posix()
    except ValueError:
        pass
return p
```

With `root=None` (the default) an absolute prune path stays absolute, while nodes are relative, so the sets are disjoint by construction.

Observed on a real incremental update of 122 changed/deleted files:

```
[graphify] 122 source file(s) deleted since last run -- no matching nodes or edges in graph, already clean.
[graphify update] Merged: 6727 nodes, 10540 edges
```

Every one of the 122 had nodes in the graph. None were pruned.

## Consequence

Deleted files' nodes are never removed from the graph on Windows, so stale nodes accumulate across updates. The message actively misleads: "already clean" is emitted for the case "0 of N matched", which is exactly when something is wrong.

## Suggested fix

Two independent improvements, either of which would have surfaced this:

1. **Normalize `prune_sources` the same way as node `source_file`** — e.g. default `root` to the graph's root when it can be inferred, or relativize prune entries against it before comparing.
[...]

Issue #2197, opened 2026-07-25
Windows: absolute source_file from detect() breaks node identity on full rebuild and poisons the semantic cache


Shimmy0530:
## Summary

On Windows, `detect()` returns **absolute** paths with backslash separators. The extraction spec tells semantic subagents to copy `source_file` from `FILE_LIST` *verbatim*, so those absolutes land in the extraction JSON. `build_from_json(..., root='.')` then fails to relativize them, so semantic nodes are keyed against a different `source_file` than the equivalent nodes in an existing graph.

The visible symptom is a refused write from the #479 shrink guard. The more damaging part is silent: `save_semantic_cache()` has **already persisted the bad entries** by the time the guard fires, so the cache is left poisoned for every subsequent run.

Versions: `graphifyy` 0.9.11, Windows 11, installed via `uv tool install`, host-agent semantic extraction (no Gemini key).

## What I observed

A repo whose graph had been built by `graphify update .` (relative `source_file` throughout). Running the full `/graphify .` pipeline per SKILL.md produced:

```
AST:      199 nodes, 564 edges
Semantic: 111 nodes, 245 edges
Merged:   301 nodes, 809 edges
```

against an existing `graph.json` of 365 nodes, so:

```
[graphify] WARNING: new graph has 301 nodes but existing graph.json has 365 (net -64).
Refusing to overwrite.
```

Diffing old vs new node IDs showed **126 missing, 62 added** — not a real shrink, a key collision. The `source_file` values tell the story:

| | `source_file` |
|---|---|
| existing graph (built by `graphify update`) | `CLAUDE.md`, `docs/DATAFLOW.md` |
| this rebuild | `D:\projects\myrepo\CLAUDE.md`, `D:\projects\myrepo\docs\DATAFLOW.md` |

Same entities, different keys, so the new ones arrived as ghost duplicates while the relative-path originals dropped out.

## Repro

On Windows, in a repo with at least one `.md` file:

```python
from graphify.detect import detect
from pathlib import Path
r = detect(Path('.'))
print(r['files']['document'][0])
# -> D:\projects\myrepo\CLAUDE.md      (absolute, backslashes)
```

Follow SKILL.md Step 3 Part B: the subagent prompt embeds that absolute path in `FILE_LIST`, and the `source_file RULE` requires copying it "character-for-character". Then Step 4:

```python
G = build_from_json(extraction, root='.', directed=False)
```

[...]

Issue #2012, opened 2026-07-18
graphify --update: ghost-node pruning silently no-ops when node source_file is relative (deleted_files is absolute)


4vbm4il-byte:
## Summary

The `--update` incremental-merge flow's ghost-node pruning step compares each existing node's `source_file` against `detect_incremental()['deleted_files']` with a plain membership check:

```python
to_remove = [n for n, d in G_existing.nodes(data=True) if d.get('source_file') in deleted]
```

`deleted_files` is always a list of **absolute** filesystem paths (confirmed via `graphify.detect.detect_incremental` — `deleted_files` is built from `manifest` keys re-anchored to absolute form via `load_manifest(manifest_path, root=root)`, then compared against `full["files"]`, which are also absolute).

However, the extraction subagent prompt template in `SKILL.md` (the semantic-extraction schema dispatched to Claude subagents) explicitly instructs `source_file` to be a path **relative** to the corpus root:

```
Node ID format: ...
"source_file": "relative/path"
```

So on any corpus where nodes were extracted per the documented schema, `d.get('source_file')` is relative and `deleted` contains only absolute paths. The `in` check can never match. The pruning step then reports something like:

```
56 file(s) deleted since last run, but no ghost nodes were present - no drift.
```

...even when the deleted files' nodes are genuinely still present in the graph as ghosts.

## Repro

1. Run a full `/graphify` pass on a corpus (nodes get relative `source_file`, per the documented schema).
2. Delete a folder of files that were part of that corpus.
3. Run `/graphify --update`.
4. Observe: `deleted_files` correctly lists the deleted files (absolute paths). The pruning step reports "no ghost nodes... no drift." The nodes for the deleted files are still present in `graph.json`.

Verified this by manually diffing node IDs between the pre- and post-update `graph.json`: 42 nodes whose source files had been deleted were still present after `--update` reported no drift. Re-running the exact same prune logic with paths normalized to relative-to-root before the membership check correctly identified and removed all 42.

## Suggested fix

Normalize both sides to the same path form before the membership check in the `--update` pruning snippet, e.g.:

```python
deleted_rel = {str(Path(f).relative_to(root)) for f in deleted}
to_remove = [n for n, d in G_existing.nodes(data=True) if d.get('source_file') in deleted_rel]
```

(or, alternatively, standardize on absolute `source_file` everywhere and update the extraction prompt schema/docs accordingly — either direction works, but the two need to agree.)

## Environment

[...]

Issue #1789, opened 2026-07-11
Node IDs for file / directory / project nodes embed the absolute scan path (leaks local username, makes committed graph.json non-portable / non-idempotent)


fremat79:
## Summary

`graphify` already **relativizes** the `source_file` field of every node so the graph
is portable across machines. However, the node **`id`** for structural nodes
(files, directories, and — for .NET — `.csproj`/`.sln`/solution/project container
nodes) is still built from the **absolute** resolved path. As a result:

- the committed `graph.json` **leaks the local OS username and full home path**
  inside those ids, and
- the **same file produces a different id on a different machine** (or a different
  checkout location), so `graphify update` on another clone re-creates those nodes
  → diff churn + non-idempotent output for a versioned graph.

Symbol nodes (classes, functions, methods) are unaffected because their ids are
name/symbol-based; only path-derived structural nodes are impacted.

## Environment

- graphify (`graphifyy`) **0.8.19**
- Windows (but the logic is OS-agnostic; same on Linux/macOS with `/home/<user>/…`)

## Current behavior

1. `detect()` resolves the scan root to an absolute path unconditionally:
   - `detect.py` → `root = root.resolve()` (≈ line 866). Every file path handled
     downstream is therefore absolute.
2. Structural / file nodes get their id from the **absolute path string**:
   - `extract.py` → `_make_id(str(path))` (e.g. the file-node id assignments around
     lines 1684 / 2526 / 2646). `path` here is absolute (from step 1).
3. By contrast, symbol nodes use a **name-based** id: `_make_id(module_name)` /
   `_make_id(str(resolved))` — no path, no leak.
4. Crucially, `graphify` **already relativizes `source_file`** but not the id:
   - `watch.py::_relativize_source_files` (≈ lines 131–143):
     `item["source_file"] = str(source_path.resolve().relative_to(root))`, called
     from `_rebuild_code` (≈ line 438). So the scan root is known and the code
     already knows how to make paths relative — it just isn't applied to the id.

### Concrete example

Project on machine A, user `alice`:

```
C:\Users\alice\repos\MyApp\src\Widgets\MyApp.Widgets.csproj
```

produces a node whose **`source_file`** is nicely relative:

```json
{ "source_file": "src/Widgets/MyApp.Widgets.csproj", "label": "MyApp.Widgets.csproj",
  "id": "c_users_alice_repos_myapp_src_widgets_myapp_widgets_csproj" }
```

[...]

PR #2619, opened 2026-08-11
fix(build): test stored paths for absoluteness on both platforms, not just the host

### LINKED ISSUE #2618: `_semantic_id_remap()` leaks an absolute POSIX `source_file` into node IDs when the graph is processed on Windows
**Labels:** bug, windows, node-identity, portability

## Summary

`build._semantic_id_remap()` deliberately refuses to re-key a node whose `source_file` is
absolute, so an on-disk path can never leak into a node ID. The guard is
`Path(sf_norm).is_absolute()`.

Under Windows path semantics a **POSIX-absolute** path such as
`/home/ci/build/repo/docs/api/README.md` is *not* absolute (`PureWindowsPath` requires a
drive), so the guard never fires and the full path is baked into the node ID.

This is the sibling of #2197 (which covered Windows-absolute paths), left unhandled for
graphs that were **built on Linux/macOS/Docker and then updated on Windows** — a routine
setup given the repo ships a `Dockerfile`.

## Steps to reproduce

On Windows:

```python
from graphify.build import _semantic_id_remap

# a graph.json produced in Docker/Linux CI carries POSIX-absolute source_file
nodes = [{"id": "api_readme",
          "source_file": "/home/ci/build/repo/docs/api/README.md",
          "type": "document"}]

print(_semantic_id_remap(nodes, None))
```

```
{'api_readme': 'home_ci_build_repo_docs_api_readme'}
```

## Actual behavior

The node ID becomes `home_ci_build_repo_docs_api_readme` — the CI machine's directory layout
embedded in the graph.

## Expected behavior

`{}` — the node is skipped, its ID left untouched, exactly as the shipped contract states.
`tests/test_build.py::test_semantic_rekey_relative_vs_absolute_source_file` asserts this:

```python
# absolute path with no resolvable root -> skipped, not remapped to an abs-path id
ab = [{"id": "api_readme", "source_file": "/abs/docs/v1/api/README.md", "type": "document"}]
assert _semantic_id_remap(ab, None) == {}
```

That test **fails on Windows** on a clean checkout.

## Root cause

`graphify/build.py:610-612`:

```python
rel = Path(sf_norm)
if rel.is_absolute():
    continue  # can't relativize (no/failed root) -- leave id untouched
```

`Path` is `WindowsPath` on Windows, and `WindowsPath("/home/ci/x").is_absolute()` is `False`
(no drive letter). The guard is a no-op for every POSIX-absolute path.

There is a second, inconsistent absoluteness test 34 lines below at `build.py:645`:

```python
sf_raw = str(sf).replace("\\", "/")
if sf_raw != sf_norm and os.path.isabs(sf_raw):
```

`os.path.isabs` is `ntpath.isabs` on Windows, which **changed behaviour in Python 3.13**:
`ntpath.isabs("/foo")` returned `True` on ≤3.12 and returns `False` on ≥3.13. So the two
guards in one function disagree with each other, and one of them additionally changes
meaning across supported Python versions (the project supports `>=3.10`).

## Impact

- Node IDs are machine-specific, so a `graph.json` built in CI and updated on a Windows
  workstation ghosts against itself — the same document yields two IDs.
- Leaks the CI/container directory layout (and potentially a username) into a graph file that
  is frequently committed. This is the same privacy concern as #1789.
- Poisons the semantic cache the same way #2197 described.

## Suggested fix

Use a platform-independent absoluteness test that treats a leading `/` as absolute
regardless of host OS, and use the *same* helper in both places:

```python
from pathlib import PurePosixPath, PureWindowsPath

def _is_absolute_any(p: str) -> bool:
    """Absolute under POSIX *or* Windows rules -- graphs move between hosts."""
    return PurePosixPath(p).is_absolute() or PureWindowsPath(p).is_absolute()
```

then `build.py:611` becomes `if _is_absolute_any(sf_norm): continue`, and `build.py:645`
uses the same helper instead of `os.path.isabs`. That also removes the 3.12/3.13
`ntpath.isabs` behaviour split.

## Related

- #2197 (CLOSED) — Windows-absolute `source_file` breaking node identity. This is the
  mirrored case that fix did not cover.
- #2566 (OPEN) — a *different* non-idempotency defect in the same function.
- #1789 (CLOSED) — absolute scan path embedded in node IDs, same leak class.

rajarshidattapy:
Fixes #2618.

## Problem

`_semantic_id_remap()` refuses to re-key a node whose `source_file` is absolute, so an
on-disk path can never leak into a node ID. The guard was `Path(sf_norm).is_absolute()` —
which answers for the **host** OS.

That is the wrong question for a path that was *stored*. A `graph.json` is portable: built in
Docker/Linux CI, updated on a Windows workstation, or the reverse. On Windows
`WindowsPath("/home/ci/build/repo/docs/api/README.md").is_absolute()` is `False` (no drive
letter), so the guard never fired and the CI machine's entire directory layout was baked into
the node ID:

```python
>>> _semantic_id_remap([{"id": "api_readme",
...                      "source_file": "/home/ci/build/repo/docs/api/README.md",
...                      "type": "document"}], None)
{'api_readme': 'home_ci_build_repo_docs_api_readme'}
```

This is the mirror of #2197 (which covered Windows-absolute paths breaking on POSIX) and the
same leak class as #1789. The shipped test
`tests/test_build.py::test_semantic_rekey_relative_vs_absolute_source_file` already asserted
the correct behaviour and **fails on Windows on a clean checkout**.

`os.path.isabs` — used by the second, inconsistent guard in the same function — is doubly
unsafe here: Python 3.13 changed `ntpath.isabs` so a single leading slash is no longer
absolute, where 3.10–3.12 said it was. The project supports `>=3.10`, so that guard silently
meant different things per interpreter.

## Fix

One shared helper in `graphify/paths.py`:

```python
def is_absolute_any_platform(p) -> bool:
    """Whether p is absolute under POSIX **or** Windows rules."""
    return PurePosixPath(s).is_absolute() or PureWindowsPath(s).is_absolute()
```

Covers drive-letter, UNC, and POSIX-root forms with either separator, and removes the
3.12/3.13 `ntpath.isabs` split. Treating a path as absolute is the conservative direction: at
worst it declines to relativize and keeps the string as-is, whereas treating an absolute path
as relative corrupts identity.

The docstring is explicit that this is for **stored** paths only — code resolving against the
real local filesystem (`cli`, `detect`, `hooks`) must keep using `Path.is_absolute()`, where
the host's rules are exactly right.

## Scope: eight sites, not the two in the issue

Grepping the concept rather than the reported line turned up the same host-only test repeated
across `build.py`, every one of them on stored paths. Two are real bugs beyond the one
reported:

- **`_derive_prune_root`** — every POSIX-absolute prune source was skipped on Windows, so the
  recovered root came back `None` and prune/replace silently no-opped. That is the
  #1151 / #2446 / #2012 failure mode reached from the other direction. Now has its own test.
- **`graph_has_legacy_ids`** — identical guard; would derive a file stem from an entire build
  directory instead of skipping the node.

The rest are `_norm_source_file`, `_abs_identity`, the second `_semantic_id_remap` guard, and
the `build_from_json` re-key. Patching only the two lines the issue named would have left the
siblings broken.

**One of the eight is a no-op today.** `_abs_identity` produces byte-identical output before
and after, because pathlib's `/` operator already resets on a rooted right-hand side. It is
changed for consistency — leaving one of eight stored-path checks on the host-only test is how
this drifts back — not because it fixes anything. Flagging it so the diff doesn't imply more
than it does.

## Verification

| Check | Result |
|---|---|
| Issue repro | `{'api_readme': 'home_ci_…'}` → **`{}`** |
| Windows-absolute path (the #2197 direction) | still skipped |
| Relative path | still re-keys → `{'api_readme': 'docs_v1_api_readme'}` |
| New tests against unfixed source | **5 fail**, 0 with the fix |
| Full suite | **40 failures vs. a 41 baseline** — zero new, and one pre-existing failure repaired |

The repaired baseline failure is `test_semantic_rekey_relative_vs_absolute_source_file`, the
shipped test this issue points at.

### Tests added

- `test_paths.py` — the helper itself across drive-letter, UNC, POSIX-root, both separators,
  and the relative/empty/`None` cases, plus a host-independence assertion
- `test_build.py` — a `FOREIGN_ABSOLUTE_SOURCE_FILES` matrix (POSIX- and Windows-absolute) run
  against `_semantic_id_remap` and `graph_has_legacy_ids`, plus dedicated tests for
  `_norm_source_file` relativization and the `_derive_prune_root` regression

I wrote and then **deleted** an `_abs_identity` test: it passed against the unfixed code, so it
guarded nothing. A vacuous test is worse than no test.

```
graphify/build.py   | 27 ++++++++++++++---------
graphify/paths.py   | 37 ++++++++++++++++++++++++++++++-
tests/test_build.py | 63 +++++++++++++++++++++++++++++++++++++++++++++++++++++
tests/test_paths.py | 47 +++++++++++++++++++++++++++++++++++++++
```

No format or API break: `is_absolute_any_platform` is additive, and no on-disk data changes.
Graphs whose node IDs were already poisoned with an absolute path keep those IDs until a
rebuild — this stops new ones being minted, it does not retroactively clean existing graphs.

### DIFF
--- a/graphify/build.py
+++ b/graphify/build.py
@@ -31,6 +31,7 @@ from pathlib import Path
 import networkx as nx
 from .ids import make_id, normalize_id as _normalize_id
 from .paths import default_graph_json as _default_graph_json
+from .paths import is_absolute_any_platform as _is_abs
 from .validate import validate_extraction
 
 
@@ -282,7 +283,7 @@ def _norm_source_file(p: str | None, root: str | None = None) -> str | None:
     if not p:
         return p
     p = p.replace("\\", "/")
-    if root and os.path.isabs(p):
+    if root and _is_abs(p):
         try:
             p = Path(p).relative_to(root).as_posix()
         except ValueError:
@@ -314,7 +315,7 @@ def _abs_identity(p: str | None, root: str | None = None) -> str | None:
         return None
     q = p.replace("\\", "/")
     pp = Path(q)
-    if not pp.is_absolute() and root:
+    if not _is_abs(q) and root:
         pp = Path(root) / q
     try:
         return pp.resolve().as_posix()
@@ -486,7 +487,7 @@ def _derive_prune_root(prune_sources: "list[str]", stored_sfs: "set[str]") -> "s
     rel_sfs = [
         sf.replace("\\", "/")
         for sf in stored_sfs
-        if sf and isinstance(sf, str) and not os.path.isabs(sf.replace("\\", "/"))
+        if sf and isinstance(sf, str) and not _is_abs(sf.replace("\\", "/"))
     ]
     if not rel_sfs:
         return None
@@ -495,7 +496,7 @@ def _derive_prune_root(prune_sources: "list[str]", stored_sfs: "set[str]") -> "s
         if not p or not isinstance(p, str):
             continue
         q = p.replace("\\", "/")
-        if not os.path.isabs(q):
+        if not _is_abs(q):
             continue
         hits = {q[: -len(s) - 1] for s in rel_sfs if q.endswith("/" + s)}
         if len(hits) > 1:
@@ -608,8 +609,13 @@ def _semantic_id_remap(nodes: list, root: str | None) -> dict:
             continue
         sf_norm = _norm_source_file(str(sf), root) or str(sf)
         rel = Path(sf_norm)
-        if rel.is_absolute():
-            continue  # can't relativize (no/failed root) — leave id untouched
+        if _is_abs(sf_norm):
+            # Can't relativize (no/failed root) — leave the id untouched rather
+            # than bake an on-disk path into it. Tested for BOTH platforms: a
+            # graph built on Linux/CI carries POSIX-absolute source_files that
+            # WindowsPath.is_absolute() calls relative, which leaked the whole
+            # build directory into node IDs when updated on Windows (#2618).
+            continue
         if not rel.name:
             # source_file equals the scan root, so _norm_source_file relativized it
             # to Path('.') — a project-level node with no per-file identity to remap.
@@ -642,7 +648,7 @@ def _semantic_id_remap(nodes: list, root: str | None) -> dict:
         # id registration. It is the longest form, so it goes first (greedy
         # prefix stripping, same ordering rule as _old_file_stems).
         sf_raw = str(sf).replace("\\", "/")
-        if sf_raw != sf_norm and os.path.isabs(sf_raw):
+        if sf_raw != sf_norm and _is_abs(sf_raw):
             abs_stem = make_id(_file_stem(Path(sf_raw)))
             if abs_stem and abs_stem != new_stem and abs_stem not in old_forms:
                 old_forms.insert(0, abs_stem)
@@ -685,8 +691,9 @@ def graph_has_legacy_ids(nodes: list, root: str | Path | None = None, sample: in
         sf = node.get("source_file")
         if not nid or not isinstance(nid, str) or not sf:
             continue
-        rel = Path(_norm_source_file(str(sf), _r) or str(sf))
-        if rel.is_absolute():
+        sf_norm = _norm_source_file(str(sf), _r) or str(sf)
+        rel = Path(sf_norm)
+        if _is_abs(sf_norm):
             continue
         if not rel.name:
             continue  # source_file == scan root -> Path('.'), no file stem (#1618)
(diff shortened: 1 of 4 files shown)


safishamsi:
Shipped in v0.9.40 (`graphifyy==0.9.40` on PyPI). Cross-platform stored-path absoluteness for #2618. Credited in the release notes. Thanks @rajarshidattapy!

Issue #2259, opened 2026-07-28
source_file written as absolute path on some runs, relative on others — causes --update to duplicate the whole graph


Hameishiv:
## Summary

`graphify <path> --update` (and, separately, even a from-scratch `detect()`/`extract()` call) can write `source_file` as either an **absolute** or a **root-relative** path depending on the run, with no apparent change in inputs. Because node identity/dedup logic is keyed off `source_file`, a run that writes absolute paths on top of a graph built with relative paths fails to recognize almost any existing node as already-present, and mints a large second copy of the graph alongside the original.

## Environment

- `graphifyy` 0.9.28 (installed via `uv tool install "graphifyy[gemini]"`)
- Windows 11, Python 3.13 (uv-managed toolchain)
- Semantic backend: `gemini` (`gemini-3-flash-preview`)
- Corpus: ~650 files, mostly Markdown, a handful of `.ts`/`.mjs`/`.py`

## Repro / observed sequence

1. Built a graph by hand-orchestrating the documented pipeline: `detect(Path(r"C:\Dev\hameishiv-wiki\wiki"))` → AST `extract()` → `extract_corpus_parallel(backend="gemini")` → merge → `build_from_json()` → `cluster()` → `to_json()`. Resulting `graph.json` nodes have `source_file` values relative to the corpus root, e.g. `"wiki/content/articles/Article 1.md"`.
2. Set up a scheduled task that runs `graphify C:\Dev\hameishiv-wiki\wiki --update` via the CLI (same absolute path argument, same cwd) inside a headless session.
3. That `--update` run wrote NEW nodes with **absolute** `source_file` values, e.g. `"C:/Dev/hameishiv-wiki/wiki/content/articles/Article 1.md"`, for files whose content had not changed. Node count went from 854 → 1154 nodes on a corpus with no actual content edits. The run's own log emitted ~538 `"node minted by two different files"` warnings.
4. To rule out CLI-vs-hand-orchestration as the cause, I redid the *entire* hand-orchestrated pipeline from scratch (no `--update`, no scheduled task, no concurrency) — literally the same `detect(Path(absolute_root))` call I'd used originally. This time it also produced absolute `source_file` values on one run, after having produced relative values on an earlier run with the identical call and no other state changed that I can identify.

## Impact

[...]

Issue #683, opened 2026-05-03
Graph fragmentation: path-separator inconsistency splits same file into 2 components + Louvain over-merging via doc-hub edges


parimisankar:
Hit two graph-quality issues on a real repo (3103 nodes / 463 connected components, ~13% nodes in giant component). Both reproduce on Windows + cross-OS-source-paths and both are mechanical to fix.

## Issue 1: path-separator inconsistency splits same file across components

**Symptom.** Single source file appears as two disconnected components in the final graph because some emitters write `apps\web\middleware\rate-limit.js` and others write `apps/web/middleware/rate-limit.js`. The `source_file` string is used as a node-attribute key (and shows up in edge `source_file`), so identity-via-source breaks across the two encodings.

**Concrete observation in my run:**

\`\`\`
size= 39  src='apps\web\middleware\rate-limit.js'
            sample=['_resetAuthLinkRateLimiter()', '_resetTurnRateLimiter()', '_readEmailMax()', ...]

size= 28  src='apps/web/middleware/rate-limit.js'
            sample=['createKeyedRateLimitBucket', 'checkCsrf (Origin allow-list)', 'checkAuthLinkRateLimit', ...]
\`\`\`

Same physical file, two graph islands. This compounds across the codebase — graph had **463 connected components** when most of the codebase is genuinely linked. Inflates orphan count, hides cross-cutting edges, distorts community detection.

**Root cause hypothesis.** Mixed sources of `source_file`:
- AST extractor (`graphify.extract`) emits OS-native (`os.sep`) on Windows → backslashes
- Semantic subagents read paths from a list passed in the prompt (whatever encoding I used) → inconsistent
- `graphify.build` does not normalize before treating `source_file` as part of node identity / edge key

**Fix.** In `graphify.build` (or earlier — at JSON ingestion), normalize all `source_file` values via:

\`\`\`python
from pathlib import PurePosixPath
def _norm(p: str) -> str:
    return str(PurePosixPath(*p.replace('\\', '/').split('/')))
\`\`\`

Apply to every node and every edge before building the graph. Keep one canonical encoding (POSIX-style forward-slash) regardless of input platform.

**Repro.** Any Windows host running \`/graphify --update\` on a corpus where AST-extractor and semantic subagents both emit a node from the same file.

---

## Issue 2: Louvain over-merging — doc-of-everything hub drags unrelated subsystems into one community

[...]

PR #2621, opened 2026-08-11
fix(tests): compare paths separator-agnostically so Windows assertions aren't vacuous

### LINKED ISSUE #2620: Tests hardcode `/` in path assertions, so they fail on Windows — and some pass vacuously, hiding the assertion they were meant to make
**Labels:** bug, tests, windows

## Summary

Several tests compare produced paths against string literals containing forward slashes. On
Windows the produced paths use `\`, so the assertions fail. The more subtle problem is that
some of these tests use `not any(... in ...)` / `not any(... .endswith(...))` **first** — and
those negative assertions pass unconditionally on Windows, because nothing ever matches a
forward-slash literal. The test then fails on a later line, but the property it was actually
guarding was never checked.

## Affected tests

```
tests/test_detect.py::test_gitignore_nested_negation_overrides_broader_root_rule
tests/test_detect.py::test_nested_ignore_overrides_git_info_exclude_and_root
tests/test_detect.py::test_path_pattern_single_star_does_not_cross_segment
tests/test_image_vision.py::test_build_image_refs_sets_rel_media_and_bytes
tests/test_watch.py::test_queue_and_drain_pending_round_trip
tests/test_cpp_preprocess.py::test_cpp_preprocess_passes_absolute_path
```

## Examples

```python
# test_detect.py -- detect() returns native absolute paths on Windows
assert any("vendor/sub/important.py" in f for f in code)
# AssertionError: assert False

# test_watch.py
assert pending_file.read_text().splitlines() == ["a.py", "sub/b.py", "c.md"]
# AssertionError: ['a.py', 'sub\\b.py', 'c.md'] == ['a.py', 'sub/b.py', 'c.md']

# test_image_vision.py
assert ref.rel == "sub/diagram.png"
# AssertionError: assert 'sub\\diagram.png' == 'sub/diagram.png'

# test_cpp_preprocess.py -- "absolute" tested as a leading slash
assert last_arg.startswith("/"), f"path arg must be absolute, got {last_arg!r}"
# AssertionError: path arg must be absolute, got 'C:\\...\\weird.F90'
```

## The vacuous-pass problem

`test_path_pattern_single_star_does_not_cross_segment` reads:

```python
assert not any(path.endswith("src/main.py") for path in files)     # vacuous on Windows
assert any(path.endswith("src/app/main.py") for path in files)     # fails
```

The first line is the one asserting that `*` did **not** cross a path segment — the actual
subject of the test. On Windows it is trivially true and proves nothing.

## These are test defects, not product defects

I verified the underlying behaviour separately, comparing separator-agnostically. The
gitignore precedence logic is correct:

```python
code = detect(t)["files"]["code"]
posix = [pathlib.Path(f).as_posix() for f in code]
# keep.py re-included by nested '!' ?  True
# drop.py correctly excluded         ?  True
```

and `_match_anchored_ignore_pattern` (`graphify/detect.py:1082`) correctly splits on `/` and
uses `fnmatchcase` per segment, so `*` cannot cross a separator.

**One exception:** `tests/test_astro_import_ids.py::test_astro_relative_inputs_keep_canonical_ids`
looks like the same class but is **not** — there the product really does emit
`src\lib\content.ts` into `source_file`. That one is filed separately as issue_4.md.

## Suggested fix

Normalise before comparing, rather than asserting on the native form:

```python
def as_posix_list(paths):
    return [pathlib.PurePath(p).as_posix() for p in paths]

assert any(p.endswith("vendor/sub/important.py") for p in as_posix_list(code))
```

For `test_cpp_preprocess_passes_absolute_path`, test the property rather than the spelling:

```python
assert os.path.isabs(last_arg), f"path arg must be absolute, got {last_arg!r}"
```

For `test_queue_and_drain_pending_round_trip`, assert the round-trip (which is the documented
contract — `_queue_pending`'s docstring promises "one path per line", not a separator) rather
than the on-disk spelling:

```python
assert set(_drain_pending(out)) == set(paths)
```

rajarshidattapy:
Fixes #2620.

## Problem

Several tests compare produced paths against string literals containing forward slashes. On
Windows the produced paths use `\`, so the assertions fail.

The failures are the visible half. The damaging half is silent: some of these tests put a
**negative** assertion first —

```python
assert not any(path.endswith("src/main.py") for path in files)   # vacuous on Windows
assert any(path.endswith("src/app/main.py") for path in files)   # fails
```

— and a `not any(...)` against a literal that can never match passes unconditionally. In
`test_path_pattern_single_star_does_not_cross_segment` that first line **is** the subject of
the test (that `*` did not cross a separator). On Windows it proved nothing, and the test then
failed on the next line for an unrelated reason.

## These are test defects — with two exceptions

I checked the underlying behaviour separator-agnostically before touching anything. The
gitignore precedence logic and `_match_anchored_ignore_pattern` are correct, as #2620 says.

Two of the six turned out to be **product** bugs rather than assertion bugs, so I fixed the
product and left those tests as written:

- **`llm._build_image_refs` and `llm._read_files`** emitted `rel` with native separators. That
  string is handed to the model as the literal `source_file` to emit — `_image_prompt_block`
  instructs *"with the listed source_file"*, and `_wrap_untrusted` embeds it as
  `<untrusted_source path="…">`. So on Windows `sub\diagram.png` lands in the graph and splits
  one file across two `source_file` forms (#683 / #2259). Both sites now use `as_posix()`.
  The existing assertion `ref.rel == "sub/diagram.png"` was right; the code was wrong.
- **`extractors/fortran.py`** carried the comment *"An absolute path always begins with `/`"* —
  the same POSIX assumption that produced the bad assertion. The hardening itself is still
  sound on Windows (a drive-letter path cannot begin with `-`); only the stated reasoning was
  wrong. Comment corrected, behaviour unchanged.

## Changes

| Test | Change |
|---|---|
| `test_gitignore_nested_negation_overrides_broader_root_rule` | compare via a new `as_posix_list()` helper |
| `test_nested_ignore_overrides_git_info_exclude_and_root` | same |
| `test_path_pattern_single_star_does_not_cross_segment` | same, and the negative assertion now actually runs |
| `test_build_image_refs_sets_rel_media_and_bytes` | **unchanged** — `llm.py` fixed instead |
| `test_queue_and_drain_pending_round_trip` | compare `[Path(line) for line in lines] == paths`; the documented contract is "one path per line", not a separator convention |
| `test_cpp_preprocess_passes_absolute_path` | `os.path.isabs(...)` instead of `startswith("/")` |

## Verification: every repaired assertion was mutation-tested

A test-hygiene PR is worthless if the repaired tests merely *pass* — that is exactly the false
signal being fixed. Each one was re-run against a deliberately broken product:

| Mutation | Result |
|---|---|
| single-star `*` allowed to cross a path segment | **caught** |
| nested-gitignore negation ignored (negation test) | **caught** |
| nested-gitignore negation ignored (info/exclude precedence test) | **caught** |
| `_queue_pending` writes every path on one line | **caught** |
| `_build_image_refs` reverted to native separators | **caught** |
| `_cpp_preprocess` passes a relative path | **caught** |

**6/6 caught.** Before this PR, four of the six could not fail on Windows at all.

### One extra test, found by that exercise

`test_cpp_preprocess_passes_absolute_path` survived its mutation at first: it hands in an
already-absolute `tmp_path / "weird.F90"`, so removing `.resolve()` changes nothing and the
test passes either way — vacuous on *every* platform, not just Windows. The original
`startswith("/")` had the same hole, so #2620's suggested one-line swap would have left it
vacuous. Added `test_cpp_preprocess_absolutises_a_relative_attacker_named_file`, exercising the
case the hardening exists for: a corpus file named `-Ietc.F90` reached by a **relative** path.
That mutation is now caught.

## Not in scope

`tests/test_astro_import_ids.py::test_astro_relative_inputs_keep_canonical_ids` looks like the
same class but is a genuine product defect — the astro/TS extractor really does write
`src\lib\content.ts` into `source_file`. Tracked separately, as #2620 notes.

```
graphify/extractors/fortran.py |  4 +++-
graphify/llm.py                | 16 ++++++++++----
tests/test_cpp_preprocess.py   | 49 +++++++++++++++++++++++++++++++++++++-----
tests/test_detect.py           | 35 ++++++++++++++++++++++++------
tests/test_watch.py            | 10 +++++----
```

### DIFF
--- a/graphify/extractors/fortran.py
+++ b/graphify/extractors/fortran.py
@@ -29,7 +29,9 @@ def _cpp_preprocess(path: Path) -> bytes:
     try:
         # Pass an absolute path so a corpus file named like "-I/etc/x.F90" cannot
         # be parsed by cpp as an option (cpp does not accept a "--" end-of-options
-        # terminator). An absolute path always begins with "/".
+        # terminator). What matters is that an absolute path cannot begin with
+        # "-" — not that it begins with "/", which only holds on POSIX (a Windows
+        # absolute path starts with a drive letter, and is equally safe).
         result = subprocess.run(
             ["cpp", "-w", "-P", "-nostdinc", "-I", "/dev/null", str(path.resolve())],
             capture_output=True,
--- a/graphify/llm.py
+++ b/graphify/llm.py
@@ -580,9 +580,13 @@ def _read_files(units: "list[Path | FileSlice]", root: Path) -> str:
             print(f"[graphify] skipping {p}: symlink target outside corpus root", file=sys.stderr)
             continue
         try:
-            rel = str(p.relative_to(root))
+            # as_posix, not str: `rel` is handed to the model as the literal
+            # source_file to emit, so a native backslash spelling on Windows
+            # lands in the graph and splits one file across two source_file
+            # forms (#683 / #2259).
+            rel = p.relative_to(root).as_posix()
         except ValueError:
-            rel = str(p)
+            rel = Path(p).as_posix()
         try:
             if isinstance(u, FileSlice):
                 content = read_slice_text(u)
@@ -812,9 +816,13 @@ def _build_image_refs(image_files: list[Path], root: Path, *, read_bytes: bool =
             print(f"[graphify] skipping image {p}: symlink target outside corpus root", file=sys.stderr)
             continue
         try:
-            rel = str(p.relative_to(root))
+            # as_posix, not str: `rel` is handed to the model as the literal
+            # source_file to emit, so a native backslash spelling on Windows
+            # lands in the graph and splits one file across two source_file
+            # forms (#683 / #2259).
+            rel = p.relative_to(root).as_posix()
         except ValueError:
-            rel = str(p)
+            rel = Path(p).as_posix()
         media = _IMAGE_MEDIA_TYPES.get(p.suffix.lower(), "image/png")
         raw: bytes | None = None
         if read_bytes:
--- a/tests/test_cpp_preprocess.py
+++ b/tests/test_cpp_preprocess.py
@@ -4,13 +4,15 @@ A corpus file is attacker-named; cpp does not accept a "--" end-of-options
 terminator, so _cpp_preprocess passes an absolute path which can never be parsed
 as a cpp option.
 """
+import os
+from pathlib import Path
+
+import pytest
+
 from graphify import extract
 
 
-def test_cpp_preprocess_passes_absolute_path(tmp_path, monkeypatch):
-    f = tmp_path / "weird.F90"
-    f.write_text("program x\nend program x\n")
-
+def _capture_cpp_argv(monkeypatch):
     captured = {}
 
     def fake_run(argv, **kwargs):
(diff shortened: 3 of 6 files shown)


safishamsi:
Shipped in v0.9.42 (`graphifyy==0.9.42` on PyPI) (#2620). Windows path-portability test fixes + POSIX source_file to the model. Credited in the release notes. Thanks @rajarshidattapy!

PR #2632, opened 2026-08-11
fix(cache): re-anchor CWD-relative source_file on write so warm hits keep canonical symbol ids

### LINKED ISSUE #2630: Astro frontmatter variable nodes do not share their file node's ID stem, and can embed the absolute on-disk path in the node ID
**Labels:** bug, extract, astro, node-identity
**Version:** 0.9.40

## Summary

`extract_astro()` mints nodes for frontmatter variables using a different ID recipe than the
one used for the enclosing file node. As a result:

1. The symbol node does **not** carry the file node's stem as a prefix, breaking the
   `<file_stem>_<entity>` convention the build layer relies on.
2. When the path handed to `extract()` is absolute and no `root` is passed, the **full
   absolute filesystem path** is baked into the symbol node's ID — the leak class of #1789 /
   #2197, still reachable in 0.9.38.

## Steps to reproduce

`src/pages/index.astro`:

```astro
---
const posts = getPosts();
---
<h1>{posts.length}</h1>
```

### 1. Stem mismatch (relative input)

```python
extract([Path('src/lib/content.ts'), Path('src/pages/index.astro')])
```

```
id=lib_content              # file node for content.ts
id=lib_content_getposts     # symbol      -- correctly prefixed
id=pages_index              # file node for index.astro
id=src_pages_index_posts    # symbol      -- NOT prefixed with "pages_index"
```

The file node is `pages_index` (2-component stem) but its own frontmatter variable is
`src_pages_index_posts` (3-component). They no longer share a prefix.

### 2. Absolute path leak (absolute input, no root)

```python
extract([Path('src/pages/index.astro').resolve()])
```

```
id=index                                                              # file node -- fine
id=c_users_asus_appdata_local_temp_claude_c_users_asus_desktop_os_gra
   phify_accccdec_07ee_4c98_8664_8b2a39362228_scratchpad_astro_src_pa
   ges_index_posts                                                    # symbol -- LEAK
```

Note both nodes come from the *same* file: the file node was correctly anchored to `index`,
the variable node was not anchored at all.

## Expected behavior

The variable node should be `pages_index_posts` (case 1) and `index_posts` (case 2) — always
`<file node stem>_<entity>`, never containing the absolute path.

## Reachability

Case 1 is reachable from `python -m graphify.extract src/pages/index.astro ...`
(`graphify/extract.py:6523` calls `extract(paths)` with no `root`) and from any library
caller. Case 2 needs absolute inputs whose inferred common root sits below the project root —
e.g. `extract([Path('/abs/src/pages/index.astro')])`, a single-file call.

Passing an explicit `root` produces correct IDs in both cases:

```python
extract(paths, root=Path('.').resolve())
# -> src_lib_content, src_lib_content_getposts, src_pages_index, src_pages_index_posts
```

so the main `graphify extract` CLI is **not** affected. This is a library-API / module-entry-point
defect.

## Impact

- `build._semantic_id_remap` and `graph_has_legacy_ids` relate a symbol to its file with
  `norm_nid.startswith(new_stem + "_")` (`build.py:630`, `build.py:1057`). A symbol that does
  not carry its file's stem is invisible to that logic and never migrates.
- The `contains` edge still connects them, but ID-based reconciliation (dedup, incremental
  re-key, cache replay) treats them as unrelated.
- Case 2 makes `graph.json` non-portable and leaks the local username.

## Suggested fix

`extract_astro()` (`graphify/extract.py:1575`) should mint frontmatter symbol IDs from the
same anchored stem the file node uses — i.e. reuse `_make_id(_file_stem(rel), name)` with the
already-computed relative path, rather than deriving from the raw input path. The safest
version is to have the `extract()` id-remap post-pass assert the invariant, so any extractor
that drifts is caught:

```python
# post-pass sanity: every symbol node must sit under its file node's stem
assert sym_id.startswith(file_stem + "_"), (sym_id, file_stem)
```

## Related

- #1789 (CLOSED) — absolute scan path embedded in file/dir/project node IDs. This is the
  residual case for astro *symbol* nodes.
- #2197 (CLOSED) — same leak class via `source_file`.

rajarshidattapy:
Fixes #2630.

## Not an astro defect

`extract_astro()`'s ID recipe is already correct — a **cold** run produces exactly what
the issue asks for:

```
pages_index          # file node
pages_index_posts    # frontmatter variable -- correctly prefixed
```

Both reported symptoms only appear on a **warm AST cache hit**, and the cause is a
round-trip asymmetry in `graphify/cache.py`:

- Extractors stamp `source_file` with the path string `extract()` was handed, so a
  relative input yields a **CWD-relative** stamp (`src/pages/index.astro`).
- `_relativize_source_files_in` skipped relative values on write ("already-relative
  fields pass through unchanged"), but `_absolutize_source_files_in` reads them back as
  **root-relative** (`root / sp`).
- When CWD is not the inferred root the two disagree, and a warm hit resurrects
  `<root>/src/pages/index.astro` — a path that names no file.

Every `source_file`-**gated** remap in `extract()` then misses. The one that matters here
is the file-stem prefix pass, which looks up `Path(source_file).resolve()` in
`prefix_remap` (`extract.py:5756`); with the lookup broken, the symbol keeps the
raw-input-path stem that a cold run rewrites. Hence `src_pages_index_posts` (case 1), and
for absolute inputs the on-disk path survives into the persisted id (case 2's leak class).

## Astro only *looks* special

`.astro` is cached; `.js/.jsx/.mjs/.cjs/.ts/.tsx/.mts/.cts/.vue/.svelte` are all in
`_JS_CACHE_BYPASS_SUFFIXES` and never round-trip through the cache. That is the entire
reason the report's `content.ts` node stayed correct while its `.astro` sibling did not.
The defect is language-agnostic — Python reproduces it identically:

```
run 1 (cold)   lib_content_get_posts
run 2 (warm)   src_lib_content_get_posts
```

## Fix

One place, upstream of every extractor: the write side re-anchors a CWD-relative
`source_file` to root-relative, so `_absolutize_source_files_in` becomes a true inverse.

Guarded to fail closed — it rewrites only when the CWD-relative reading names a real file
**and** differs from the root-relative reading. A fragment that already stores
root-relative (a semantic subagent's, per `_normalize_source_file_value`) and a ghost path
are both left exactly as before.

## What is deliberately *not* changed

- **`extract_astro()`** — nothing to fix, and the issue's suggested post-pass
  `assert sym_id.startswith(file_stem + "_")` is redundant once the shared gate is correct.
  An `assert` would also turn a stale cache entry into a crash.
- **Cross-root shared `cache_root`** — `extract(sub, cache_root=X)` after
  `extract(super, cache_root=X)` still replays a stale rel path, because the
  `$graphify-root$` marker (#2257) re-anchors only the *root* portion and assumes a file's
  path *within* the root is unchanged. Separate design limit, and not reachable from this
  issue's repro (default `cache_root` follows `root`).

## Verification

| scenario | before | after |
|---|---|---|
| case 1 warm (relative inputs, CWD above root) | `src_pages_index_posts` | `pages_index_posts` |
| case 2 (absolute single-file input) | `index_posts` / leaked slug when warmed cross-root | `index_posts` |
| python equivalent, warm | `src_lib_content_get_posts` | `lib_content_get_posts` |

New test `tests/test_cache.py::test_warm_hit_with_relative_inputs_from_above_the_root`
asserts warm ids == cold ids and that every symbol sits under its file node's stem. It
carries the same warmth probe as the #2257 test, so a silent re-extraction cannot make it
pass vacuously. Confirmed failing without the fix, passing with it.

Full suite: identical failure set before and after (42 pre-existing, Windows/env). One of
those is `test_astro_import_ids.py::test_astro_relative_inputs_keep_canonical_ids`, which
fails on `source_file` separators — the *other* defect in the same repro project, out of
scope here.

### DIFF
--- a/graphify/cache.py
+++ b/graphify/cache.py
@@ -524,8 +524,24 @@ def _relativize_source_files_in(payload: dict, root: Path) -> None:
     forward-slash relative paths from ``root``.
 
     Mirror of :func:`graphify.watch._relativize_source_files` so cached
-    extraction fragments persist in portable form (#777). Already-relative
-    fields and out-of-root paths pass through unchanged.
+    extraction fragments persist in portable form (#777). Out-of-root paths
+    pass through unchanged.
+
+    A CWD-relative field is re-anchored too. Extractors stamp ``source_file``
+    with the path string ``extract()`` was handed, so relative inputs yield a
+    CWD-relative stamp — but the stored format is root-relative, and
+    :func:`_absolutize_source_files_in` reads it back as such. When CWD is not
+    the inferred root the two disagree and a warm hit resurrects a path that
+    names no file (``<root>/src/pages/index.astro`` for an input of
+    ``src/pages/index.astro`` under root ``<root>``). Every source_file-GATED
+    remap in ``extract()`` then misses — the file-stem prefix pass looks up
+    ``Path(source_file).resolve()`` in ``prefix_remap`` — so a warm hit keeps
+    the raw-path symbol ids a cold run canonicalizes: symbols stop sharing
+    their file node's stem, and for absolute inputs the on-disk path survives
+    into the persisted id (#2630). Only rewritten when the CWD-relative
+    reading is a real file and the root-relative reading is a different path,
+    so a fragment that already stores root-relative (a semantic subagent's,
+    see :func:`_normalize_source_file_value`) is left alone.
 
     Only ``root`` is resolved — ``source_file`` itself is relativized
     symbolically so in-root symlinks keep their original name rather than
@@ -549,7 +565,15 @@ def _relativize_source_files_in(payload: dict, root: Path) -> None:
                 continue
             sp = Path(source)
             if not sp.is_absolute():
-                continue
+                # os.path.abspath is lexical (no symlink resolution), matching
+                # the symbolic relativization below.
+                cwd_form = Path(os.path.abspath(sp))
+                try:
+                    if cwd_form == root_resolved / sp or not cwd_form.exists():
+                        continue  # already root-relative, or a ghost path
+                except OSError:
+                    continue
+                sp = cwd_form
             try:
                 rel = os.path.relpath(sp, root_resolved)
             except (ValueError, OSError):
--- a/tests/test_cache.py
+++ b/tests/test_cache.py
@@ -490,6 +490,75 @@ def test_relative_root_does_not_reanchor_an_already_canonical_id(tmp_path, monke
     assert loaded["nodes"][0]["id"] == "src_utils_foo"
 
 
+def test_warm_hit_with_relative_inputs_from_above_the_root(tmp_path, monkeypatch):
+    """#2630: relative inputs handed to extract() from a CWD above the root.
+
+    Extractors stamp ``source_file`` with the path STRING they were handed, so
+    a relative input yields a CWD-relative stamp — but the stored format is
+    root-relative and ``load_cached`` re-anchors it as such. When CWD is not
+    the inferred root the two disagree and a warm hit resurrects a path naming
+    no file (``<root>/src/pages/index.astro``). Every source_file-GATED remap
+    in extract() then misses, so the warm hit keeps the raw-path symbol ids a
+    cold run canonicalizes: the astro frontmatter variable came back as
+    ``src_pages_index_posts``, no longer under its ``pages_index`` file node's
+    stem — the prefix ``build.py`` reconciles symbols to files by.
+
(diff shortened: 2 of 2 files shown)


safishamsi:
Fixed in v0.9.41 (`graphifyy==0.9.41` on PyPI). A warm cache hit no longer re-anchors a CWD-relative `source_file` to a ghost path when the working directory differs from the graph root; cold and warm node ids match. Thanks @rajarshidattapy.

safishamsi:

PR #2634, opened 2026-08-11
fix(csharp): extract members in preprocessor blocks

### LINKED ISSUE #2631: C#: members inside `#if` / `#else` blocks produce no nodes, and calls to them produce no edges
## Summary

Any C# member declared inside a preprocessor conditional (`#if`, `#else`, `#elif`) is absent
from the graph. No node is created for it, and call sites that target it produce no `calls`
edge either — the call simply disappears.

This is not about picking the "wrong" branch: **neither** branch is extracted. A type whose
members are all guarded by `#if` ends up as a class node with no methods at all, so nothing
can ever point at those members.

Conditional compilation is common in multi-targeted libraries (`#if NET8_0_OR_GREATER`,
`#if NETFRAMEWORK`, `#if DEBUG`), so this silently removes real API surface from the graph.

## Reproducer

Two files, no dependencies, `net8.0`. `dotnet build -c Debug` → 0 warnings, 0 errors, so this
is not a case of invalid C#.

`Helper.cs`

```csharp
namespace Probe.Lib;

public static class Gated
{
    // control case: outside any preprocessor directive
    public static void Outside(string a) { }

#if NET8_0_OR_GREATER
    public static void InsideIf(string a) { }
#else
    public static void InsideElse(string a) { }
#endif

#if DEBUG
    public static void InsideDebug(string a) { }
#endif
}
```

`Caller.cs`

```csharp
namespace Probe.Lib;

public static class Caller
{
    public static void Drive()
    {
        Gated.Outside("x");
        Gated.InsideIf("x");
        Gated.InsideDebug("x");
    }
}
```

Run `graphify extract <dir>` on the directory holding the two `.cs` files.

**Expected:** member nodes for `Outside`, `InsideIf`, `InsideDebug` (and arguably
`InsideElse`), and three `calls` edges from `Drive`.

**Actual:**

```
member nodes: ['.Drive()', '.Outside()']
calls edges:  .Drive() -> .Outside()          (1 of 3)
```

`InsideIf`, `InsideElse` and `InsideDebug` have no node, and the two calls to them are
missing. The `Gated` class node exists and carries only `.Outside()`.

## Version span

Measured with the same reproducer on **0.9.8, 0.9.20, 0.9.34 and 0.9.39** — byte-identical
result in all four (2 member nodes, 1 `calls` edge). This is a long-standing gap, not a
regression in a recent release.

## Impact in a real code base

Counted across 8 separately extracted C# code bases (roughly 10 000 `.cs` files),
**not sampled**:

* Method declarations that sit below a preprocessor directive in the syntax tree: **233**
* Of these, **at least 99 have no member node in the graph.**

Method: for every `.cs` file in each graph that contains `#if`, the file is parsed with
tree-sitter and every `method_declaration` whose ancestors include a `preproc_*` node is
checked against the member nodes recorded for that file. The check matches on the member
**name**, so a guarded method whose name also occurs outside the block counts as "present" —
99 is therefore a **lower bound**, not an exact figure.

A concrete consequence from that code base: a public static helper class exposes its `Setup`
entry point only inside `#if NET8_0_OR_GREATER` / `#else`. The class node exists and carries
its nine unguarded members, but `Setup` is absent — so the five call sites that invoke it can
never be connected to it, and "who calls `Setup`?" is unanswerable.

## Notes on what is *not* claimed here

* No claim about which branch *should* win. Extracting the members of every branch would
  already be a large improvement over dropping all of them; picking a branch by defined
  symbols is a separate design question.
* The cause was not traced in the extractor source. The reproducer localises the behaviour,
  nothing more.
* Non-method members (properties, fields, nested types) inside conditionals were not
  examined; the count above covers `method_declaration` only.

rohit-jsfreaky:
## Summary

Fixes #2631.

C# members declared inside `#if`, `#elif`, and `#else` blocks were skipped because the generic extractor treated tree-sitter's `preproc_*` nodes as scope boundaries. This removed the enclosing class context, so guarded members were not emitted as class methods and call resolution could not target them.

## Changes

- Treat C# preprocessor nodes as transparent while walking a class body.
- Preserve the enclosing class ID for methods nested in conditional compilation blocks.
- Add a regression test with two C# files covering guarded methods in `#if` / `#else` blocks and calls to the active methods.

Both conditional branches are represented in the graph. Choosing a build configuration remains outside the extractor's scope.

## Validation

- `uv run pytest tests/test_dotnet.py -q -k preprocessor` — passed.
- `uv run pytest tests/test_dotnet.py -q` — 42 passed.
- `uv run pytest tests/test_languages.py -q -k csharp` — 14 passed.
- `uv run ruff check graphify/extractors/engine.py tests/test_dotnet.py` — passed.
- `uv run graphify update .` — completed successfully.

I also ran `uv run pytest tests/ -q`: 4,271 passed and 20 skipped. The 20 failures are unrelated Windows/platform-sensitive tests in Astro IDs, filesystem permissions, gitignore, hooks, image handling, install paths, and watch-path separators; none exercise the C# extractor or this regression.

### DIFF
--- a/graphify/extractors/engine.py
+++ b/graphify/extractors/engine.py
@@ -2162,7 +2162,7 @@ def _csharp_extra_walk(node, source: bytes, file_nid: str, stem: str, str_path: 
                        nodes: list, edges: list, seen_ids: set, function_bodies: list,
                        parent_class_nid: str | None, add_node_fn, add_edge_fn,
                        walk_fn, namespace_stack: list[str], scope_stack: list[str]) -> bool:
-    """Handle namespace declarations for C#. Returns True if handled."""
+    """Handle C# namespaces and transparent class-member wrappers."""
     if node.type == "namespace_declaration":
         ns_name = _csharp_namespace_name(node, source)
         pushed = False
@@ -2198,6 +2198,13 @@ def _csharp_extra_walk(node, source: bytes, file_nid: str, stem: str, str_path: 
             line = node.start_point[0] + 1
             add_node_fn(ns_nid, ns_label, line, node_type="namespace", metadata={"kind": "csharp_namespace"})
             add_edge_fn(file_nid, ns_nid, "contains", line)
+        return True
+    if parent_class_nid and node.type.startswith("preproc_"):
+        # tree-sitter wraps members in #if/#else/#elif directives in preproc_*
+        # nodes. They are conditional containers, not ownership scopes: dropping
+        # parent_class_nid here makes guarded methods look file-level (#2631).
+        for child in node.children:
+            walk_fn(child, parent_class_nid)
         return True
     return False
 
--- a/tests/test_dotnet.py
+++ b/tests/test_dotnet.py
@@ -519,6 +519,56 @@ def test_xaml_viewmodel_with_non_utf8_codebehind_does_not_crash(tmp_path):
     assert nodes[edges[0]["target"]]["label"] == "SettingsViewModel"
 
 
+def test_csharp_members_in_preprocessor_blocks_are_extracted_and_resolved(tmp_path):
+    """C# preprocessor wrappers must preserve class ownership for members (#2631)."""
+    helper = tmp_path / "Helper.cs"
+    helper.write_text(
+        """namespace Probe.Lib;
+public static class Gated
+{
+    public static void Outside(string a) { }
+#if NET8_0_OR_GREATER
+    public static void InsideIf(string a) { }
+#else
+    public static void InsideElse(string a) { }
+#endif
+#if DEBUG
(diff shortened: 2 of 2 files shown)


safishamsi:
Fixed in v0.9.41 (`graphifyy==0.9.41` on PyPI). C# members inside a `#if ... #endif` preprocessor block are now extracted and attached to their class (both `#if`/`#else` branches; `#region` unaffected). Thanks @rohit-jsfreaky.

safishamsi:
Shipped in v0.9.41 (`graphifyy==0.9.41` on PyPI). Credited in the release notes. Thanks @rohit-jsfreaky!

PR #2645, opened 2026-08-11
fix(js): stabilize unresolved local import targets

### LINKED ISSUE #2457: imports to a file absent from the scan (e.g. gitignored codegen) emit abs-path slug endpoints (residual of #2262)
Follow-on to #2262. That report's `.tsx`-handler `calls` case was fixed in 0.9.30, but the import-edge residual I flagged in a follow-up comment (before the issue was closed) is still present on 0.9.32. Filing it separately with a minimal repro since it is a distinct code path.

## Bug

When a TS/JS file imports from a module that is not present in the scanned tree (e.g. a gitignored codegen output), the `imports` and `imports_from` edges get absolute-path-derived slug endpoints instead of a stable identifier. The slug is machine-specific and differs per checkout path, so committed graphs churn on every rebuild from a different location, the same reproducibility problem as #2262 and #2273.

## Repro (minimal, deterministic on 0.9.32)

```
proj/tsconfig.json
proj/src/consumer.ts   -> import { getFoo } from "./generated/api";
                          export function run(): number { return getFoo(); }
```

`proj/src/generated/api.ts` is intentionally absent (mimics gitignored codegen output). `graphify update proj --no-cluster` yields:

```
src_consumer --imports_from--> <abs-scan-slug>_proj_src_generated_api
src_consumer --imports-->      <abs-scan-slug>_proj_src_generated_api_getfoo
```

Both endpoints are the absolute scan path slugified, for a module and symbol that are not on disk.

## Scale

On a real TS monorepo (~75k nodes), this produced ~6,700 such endpoints (4,138 `imports`, 2,510 `imports_from`, plus a few `references`/`re_exports`), almost all imports to a gitignored generated GraphQL client. It is the dominant reason that graph fails a fail-closed host-path check.

## Suggested fix

For an import target that resolves to no node (an absent file), emit a repo-relative stable identifier or drop the edge, rather than an absolute-path-derived slug. The general backstop that canonicalizes node-less absolute-derived endpoints elsewhere does not cover the `imports`/`imports_from` path.

rohit-jsfreaky:
Fixes #2457.

## Summary
When a static JS/TS import points to a local module that is absent from the scan, `_resolve_js_import_path` returns the attempted path. `_import_js` then treated that path as resolved and generated absolute-path-derived IDs, causing graph churn between checkouts.

This change treats a non-existent static import target as unresolved:
- emits the stable `ref_<specifier>` file-level `imports_from` target;
- avoids emitting a named-symbol `imports` edge for a file that does not exist.

## Tests
- Added coverage for `import { getFoo } from "./generated/api"` where `generated/api` is absent. The regression asserts the stable reference target and no fabricated symbol edge.
- `uv run pytest tests/test_js_import_resolution.py tests/test_ts_import_require.py tests/test_phantom_external_import.py tests/test_extract.py::test_extract_js_destructured_require_named_symbols tests/test_extract.py::test_extract_js_member_require_emits_property_symbol -q` — 74 passed.
- `uv run ruff check graphify/extract.py tests/test_js_import_resolution.py` — passed.
- `uv run pytest tests/ -q` — 4,271 passed, 20 skipped. The 20 failures match the existing Windows/platform-sensitive baseline and do not exercise this static import path.

### DIFF
--- a/graphify/extract.py
+++ b/graphify/extract.py
@@ -413,6 +413,12 @@ def _import_js(node, source: bytes, file_nid: str, stem: str, edges: list, str_p
         resolved = _resolve_js_import_target(raw, str_path)
         if resolved is not None:
             tgt_nid, resolved_path = resolved
+            # `_resolve_js_import_path` returns the attempted path when no
+            # local file exists. Static ES imports must treat that as unresolved
+            # rather than minting a checkout-specific target ID (#2457).
+            if resolved_path is not None and not resolved_path.is_file():
+                tgt_nid = _make_id("ref", raw)
+                resolved_path = None
             edge = {
                 "source": file_nid,
                 "target": tgt_nid,
--- a/tests/test_js_import_resolution.py
+++ b/tests/test_js_import_resolution.py
@@ -1017,6 +1017,29 @@ def test_tsconfig_alias_none_exist_creates_no_false_edge(tmp_path: Path):
     assert not _has_edge(result, "src/routes/page.ts", "src/lib/utils.ts")
 
 
+def test_unresolved_relative_import_uses_stable_ref_target(tmp_path: Path):
+    """A missing local module must not leak the checkout path into graph IDs (#2457)."""
+    importer = _write(
+        tmp_path / "src/consumer.ts",
+        "import { getFoo } from './generated/api'\n"
+        "export function run(): number { return getFoo() }\n",
+    )
+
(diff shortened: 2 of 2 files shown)


safishamsi:
Shipped in v0.9.41 (`graphifyy==0.9.41` on PyPI) (#2457). Credited in the release notes. Thanks @rohit-jsfreaky!

PR #2558, opened 2026-08-08
docs: fix article typo in ARCHITECTURE.md

nelsondeleonc-source:
## Summary
- Fixes a small grammar issue in the "Adding a new language extractor" section: "Add a `extract_<lang>...`" should be "Add an `extract_<lang>...`" since the identifier starts with a vowel sound.

## Test plan
- Docs-only change, no functional impact.

PR #2646, opened 2026-08-11
docs: rewrite ARCHITECTURE.md's module table from the real surface, with a test that keeps it honest

### LINKED ISSUE #2640: `ARCHITECTURE.md` documents six functions that do not exist and the wrong signature for `extract()`
**Labels:** bug, docs
**Version:** 0.9.40

## Summary

The "Module responsibilities" table in `ARCHITECTURE.md` is the entry point for anyone using
graphify as a library (and for agents, which `AGENTS.md` directs at the docs first). Six of
its sixteen documented functions do not exist, and the one signature it gives in prose is
wrong.

## Steps to reproduce

```python
import importlib

spec = [
    ('graphify.detect', 'collect_files'),   ('graphify.extract', 'extract'),
    ('graphify.build', 'build_graph'),      ('graphify.cluster', 'cluster'),
    ('graphify.analyze', 'analyze'),        ('graphify.report', 'render_report'),
    ('graphify.export', 'export'),          ('graphify.callflow_html', 'write_callflow_html'),
    ('graphify.ingest', 'ingest'),          ('graphify.cache', 'check_semantic_cache'),
    ('graphify.cache', 'save_semantic_cache'), ('graphify.security', 'validate_url'),
    ('graphify.validate', 'validate_extraction'), ('graphify.serve', 'start_server'),
    ('graphify.watch', 'watch'),            ('graphify.benchmark', 'run_benchmark'),
]
for mod, fn in spec:
    m = importlib.import_module(mod)
    print(('OK     ' if hasattr(m, fn) else 'MISSING'), f'{mod}.{fn}')
```

## Actual behavior

```
MISSING graphify.detect.collect_files
OK      graphify.extract.extract
MISSING graphify.build.build_graph
OK      graphify.cluster.cluster
MISSING graphify.analyze.analyze
MISSING graphify.report.render_report
MISSING graphify.export.export
OK      graphify.callflow_html.write_callflow_html
OK      graphify.ingest.ingest
OK      graphify.cache.check_semantic_cache
OK      graphify.cache.save_semantic_cache
OK      graphify.security.validate_url
OK      graphify.validate.validate_extraction
MISSING graphify.serve.start_server
OK      graphify.watch.watch
OK      graphify.benchmark.run_benchmark
```

Additionally, the pipeline diagram and table give `extract()` as `extract(path)` —
"file path → `{nodes, edges}` dict". The real signature takes a **list**:

```python
>>> from graphify.extract import extract
>>> extract(Path("a.py"))
TypeError: 'WindowsPath' object is not iterable
```

It is `extract(paths: list[Path], root=None, ...)`.

## Expected behavior

The table names functions that exist, with correct signatures.

## Actual names (for the fix)

| Documented | Actual |
|---|---|
| `detect.collect_files(root)` | `detect.detect(root)`; `collect_files` lives in `graphify.extract` |
| `extract.extract(path)` | `extract.extract(paths: list[Path], root=None, ...)` |
| `build.build_graph(extractions)` | `build.build(...)` / `build.build_from_json(...)` |
| `analyze.analyze(G)` | no single entry point — `god_nodes`, `surprising_connections`, `suggest_questions`, `graph_diff`, `find_import_cycles` |
| `report.render_report(G, analysis)` | `report.generate(...)` |
| `export.export(G, out_dir, ...)` | `export.to_json` / `to_wiki` / `to_cypher` / `to_html` … |
| `serve.start_server(graph_path)` | not present under that name |

## Why it matters beyond tidiness

The `root=` parameter that `ARCHITECTURE.md` omits from `extract()` is exactly the one whose
absence produces the defects in issue_4.md and issue_5.md — a caller following this document
would call `extract(paths)` with no root and silently get non-canonical `source_file` values
and absolute paths baked into node IDs. The doc is steering library users onto the buggy path.

## Suggested fix

Regenerate the table from the actual public surface, and add a docs test that imports each
documented symbol so the table cannot rot again:

```python
@pytest.mark.parametrize("mod,fn", _parse_architecture_table())
def test_architecture_table_symbols_exist(mod, fn):
    assert hasattr(importlib.import_module(mod), fn)
```

rajarshidattapy:
Fixes #2640.

## The bug

The "Module responsibilities" table is where library users and agents start (AGENTS.md points
agents at the docs before the code). Six of its sixteen documented functions did not exist:

```
MISSING graphify.detect.collect_files
MISSING graphify.build.build_graph
MISSING graphify.analyze.analyze
MISSING graphify.report.render_report
MISSING graphify.export.export
MISSING graphify.serve.start_server
```

and `extract()` was given as `extract(path)` — "file path → dict" — for a function whose first
parameter is a **list**, so following the doc raises `TypeError: 'WindowsPath' object is not
iterable`.

## Fix

Rewrote the table from the actual public surface, introspected rather than guessed
(`inspect.signature` over each module's public functions). All six mappings applied as the
issue specifies, with two amendments — both verified:

- `to_wiki` lives in `graphify.wiki`, not `graphify.export`. It gets its own row.
- `serve.py`'s real entry points are `serve(graph_path)` and
  `serve_http(graph_path, *, host, port, ...)`.

## Four more errors in the same table

`hasattr` only proves a name exists, so the issue's repro couldn't catch these. Each row's
described behaviour was wrong too:

| row | claimed | actual |
|---|---|---|
| `cluster` | "graph with `community` attr on each node" | returns `{community_id: [node_id]}`; the graph is not mutated |
| `validate` | "extraction dict → raises on schema errors" | `validate_extraction` **returns** `list[str]` of errors; `assert_valid` is the one that raises |
| `watch` | `watch(root, flag_path)` | `watch(watch_path, debounce=3.0)` — there is no `flag_path` parameter |
| `detect` | "directory → `[Path]` filtered list" | `detect(root)` returns a scan-summary dict: `files` by category, `total_files`, `total_words`, `warning`, `scan_root` |

Also corrected `build_graph()` in the pipeline diagram and the schema section, and softened
"Each stage is a single function in its own module" — no longer true of `analyze.py` or
`export.py`.

## On the `root=` parameter

The issue is right that this is the load-bearing part: a caller following the old doc writes
`extract(paths)` with no root and silently gets non-canonical `source_file` values and
absolute path segments in node ids. A parameter listed in a table is easy to skim past, so
rather than only fixing the signature there is now a short **"Calling `extract()` from your
own code"** subsection with a worked example and an explicit *"Always pass `root`"*, spelling
out that the inferred root is the common parent of the list you passed, not your project root.

## Tests

`tests/test_architecture_doc.py` implements the issue's suggestion, parsing the table rather
than restating it — so a new row extends coverage for free, and renaming a function in the
code fails here until the doc is updated:

```python
@pytest.mark.parametrize("module,func", _documented_symbols())
def test_architecture_table_symbols_exist(module, func):
    assert hasattr(importlib.import_module(module), func)
```

Plus `test_architecture_documents_extract_as_taking_a_list` and
`test_architecture_tells_library_callers_to_pass_root` (asserting against the live signature,
not just the prose), and `test_the_table_was_actually_parsed` — a guard on the parser itself,
since a regex that silently matches nothing would make every parametrized case vacuously pass.

Verified against the pre-fix doc: exactly the six reported symbols fail, plus both `extract()`
signature tests. 38 tests pass on the new doc. Full suite shows the same 42 pre-existing
failures as baseline (Windows/env), none new.

## One change beyond the reported scope

Step 1 of "Adding a new language extractor" says to add the function in `extract.py`, but
`graphify/extractors/MIGRATION.md` documents an in-progress split to one module per language.
Repointed step 1 at `graphify/extractors/` and noted that `extract.py` re-exports during the
migration. Adjacent to the reported defect rather than part of it — happy to drop this hunk if
the split isn't meant to be advertised yet.

### DIFF
--- a/ARCHITECTURE.md
+++ b/ARCHITECTURE.md
@@ -5,30 +5,47 @@ graphify is a Claude Code skill backed by a Python library. The skill orchestrat
 ## Pipeline
 
 ```
-detect()  →  extract()  →  build_graph()  →  cluster()  →  analyze()  →  report()  →  export()
+detect()  →  extract()  →  build()  →  cluster()  →  analyze helpers  →  report.generate()  →  export.to_*()
 ```
 
-Each stage is a single function in its own module. They communicate through plain Python dicts and NetworkX graphs - no shared state, no side effects outside `graphify-out/`.
+Each stage lives in its own module and they communicate through plain Python dicts and NetworkX graphs - no shared state, no side effects outside `graphify-out/`. Most stages are a single function; `analyze.py` and `export.py` are sets of sibling functions rather than one entry point.
 
 ## Module responsibilities
 
-| Module | Function | Input → Output |
-|--------|----------|----------------|
-| `detect.py` | `collect_files(root)` | directory → `[Path]` filtered list |
-| `extract.py` | `extract(path)` | file path → `{nodes, edges}` dict |
-| `build.py` | `build_graph(extractions)` | list of extraction dicts → `nx.Graph` |
-| `cluster.py` | `cluster(G)` | graph → graph with `community` attr on each node |
-| `analyze.py` | `analyze(G)` | graph → analysis dict (god nodes, surprises, questions) |
-| `report.py` | `render_report(G, analysis)` | graph + analysis → GRAPH_REPORT.md string |
-| `export.py` | `export(G, out_dir, ...)` | graph → Obsidian vault, graph.json, graph.html, graph.svg |
+Signatures below are the real ones - `tests/test_architecture_doc.py` imports every symbol named here, so this table cannot drift from the code.
+
+| Module | Entry point(s) | Input → Output |
+|--------|----------------|----------------|
+| `detect.py` | `detect(root)` | directory → scan summary dict: `files` grouped by category, plus `total_files`, `total_words`, `warning`, `scan_root`, … |
+| `extract.py` | `extract(paths, *, root=None, ...)`, `collect_files(target)` | **list** of file paths → `{nodes, edges}` dict. `collect_files` expands a directory into that list, and lives here, not in `detect.py` |
+| `build.py` | `build(extractions)`, `build_from_json(extraction)` | extraction dict(s) → `nx.Graph` |
+| `cluster.py` | `cluster(G)` | graph → `{community_id: [node_id, ...]}` (the graph is not mutated) |
+| `analyze.py` | `god_nodes(G)`, `surprising_connections(G)`, `suggest_questions(G, communities, community_labels)`, `find_import_cycles(G)`, `graph_diff(G_old, G_new)` | graph → one list/dict per analysis. There is no single `analyze()` entry point |
+| `report.py` | `generate(G, communities, cohesion_scores, community_labels, ...)` | graph + analysis → GRAPH_REPORT.md string |
+| `export.py` | `to_json`, `to_html`, `to_obsidian`, `to_svg`, `to_graphml`, `to_canvas`, `to_cypher` | graph → graph.json, graph.html, Obsidian vault, graph.svg, … one function per format |
+| `wiki.py` | `to_wiki(G, communities, output_dir, ...)` | graph → one markdown article per community + `index.md` |
 | `callflow_html.py` | `write_callflow_html(...)` | graphify-out files → Mermaid architecture/call-flow HTML |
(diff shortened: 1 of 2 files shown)


safishamsi:
Shipped in v0.9.42 (`graphifyy==0.9.42` on PyPI). Refreshed ARCHITECTURE.md module table + doc-parity test (kept the #2558 'an' fix). Credited in the release notes. Thanks @rajarshidattapy!

PR #2647, opened 2026-08-11
fix(tests): stop argv mangling the #2126 shell-allowlist payloads on Windows

### LINKED ISSUE #2641: The `#2126` shell-allowlist regression tests pass vacuously on Windows — `_shell_verdict` mangles the glob before bash ever sees it
**Labels:** bug, tests, windows, security-testing
**Version:** 0.9.40
**Environment:** Windows 11, Git Bash 5.2.37, Python 3.13.6

## Summary

`tests/test_hooks.py` guards the post-commit hook's interpreter allowlist — the fix for
#2126, and the guard that stops `;`, backticks and `$(...)` reaching a shell. On Windows the
helper that runs the glob passes the pattern through `subprocess` as an argv element, and
Windows argv quoting rewrites the backslashes inside the bracket expression. bash receives a
**different** pattern, one that matches nothing.

Consequences, both bad in opposite directions:

- `test_python_detect_allowlists_still_reject_shell_metacharacters` **fails** (4 params) —
  a false alarm claiming the allowlist accepts `foo;rm -rf /`.
- `test_file_path_allowlist_accepts_windows_backslash_path` and
  `test_shebang_allowlist_accepts_windows_backslash_path` **pass vacuously** — the pattern
  matches nothing, so "ACCEPTED" is guaranteed regardless of what the allowlist says. The
  #2126 regression test provides no coverage on the platform #2126 was about.

**The shipped hook is correct.** I verified it separately. This is purely a test-harness
defect — but one that disables a security regression test on the affected platform.

## Steps to reproduce

```
$ pytest tests/test_hooks.py -q -k allowlist
```

```
FAILED tests/test_hooks.py::test_python_detect_allowlists_still_reject_shell_metacharacters[foo;rm -rf /]
FAILED tests/test_hooks.py::test_python_detect_allowlists_still_reject_shell_metacharacters[foo`id`]
FAILED tests/test_hooks.py::test_python_detect_allowlists_still_reject_shell_metacharacters[foo$(id)]
FAILED tests/test_hooks.py::test_python_detect_allowlists_still_reject_shell_metacharacters[foo$IFS]
E   AssertionError: _FROM_FILE="" allowlist wrongly accepted dangerous input 'foo;rm -rf /'
```

### Demonstrating that the harness, not the hook, is wrong

Through `_shell_verdict`'s `subprocess.run([...])` path, **every** input is ACCEPTED — even
`/usr/bin/python3`, which the allowlist must accept, and `foo;rm -rf /`, which it must
reject. A pattern that never matches anything:

```
'foo;rm -rf /'                        -> ACCEPTED
'foo`id`'                             -> ACCEPTED
'foo$(id)'                            -> ACCEPTED
'C:\Users\u\.venv\Scripts\python.exe' -> ACCEPTED
'/usr/bin/python3'                    -> ACCEPTED
```

Writing the *same* case arm to a `.sh` file and running `bash case_test.sh <input>` — which
is how the hook actually executes — gives the correct answers:

```
$ cat case_test.sh
case "$1" in
  *[!a-zA-Z0-9/_.@:\\-]*) echo REJECTED ;;
  *) echo ACCEPTED ;;
esac

foo;rm -rf /                          REJECTED
foo`id`                               REJECTED
foo$(id)                              REJECTED
foo$IFS                               REJECTED
C:\Users\u\.venv\Scripts\python.exe   ACCEPTED
/usr/bin/python3                      ACCEPTED
C:/Users/u/python.exe                 ACCEPTED
```

Both the #2126 property (Windows paths accepted) and the injection property (metacharacters
rejected) hold.

## Root cause

`tests/test_hooks.py:463-473`:

```python
def _shell_verdict(pattern: str, candidate: str) -> str:
    result = subprocess.run(
        ["bash", "-c", f'case "$1" in\n{pattern}) echo REJECTED ;;\n*) echo ACCEPTED ;;\nesac', "_", candidate],
        capture_output=True, text=True,
    )
```

On Windows, `subprocess` has no `execve`; the argument list is joined into a single command
string by `subprocess.list2cmdline`, whose rules double backslashes that precede a quote and
otherwise pass them through. The pattern contains `\\` inside a bracket expression, so what
`bash.exe` parses is not what was passed. `assert result.returncode == 0` does not catch it,
because a mangled-but-still-valid `case` arm exits 0.

## Suggested fix

Write the snippet to a file and invoke `bash script.sh <candidate>` — no shell-metacharacter
payload transits argv:

```python
def _shell_verdict(pattern: str, candidate: str, tmp_path) -> str:
    script = tmp_path / "case_test.sh"
    script.write_text(
        f'case "$1" in\n  {pattern}) echo REJECTED ;;\n  *) echo ACCEPTED ;;\nesac\n',
        encoding="utf-8", newline="\n",
    )
    result = subprocess.run(["bash", str(script), candidate],
                            capture_output=True, text=True)
    assert result.returncode == 0, result.stderr
    return result.stdout.strip()
```

This also matches how the hook runs in production (a script file read by bash), so the test
exercises the real code path.

To stop the vacuous-pass failure mode returning, add a control assertion to the accept tests:

```python
# sanity: the pattern must actually be capable of rejecting something,
# otherwise an ACCEPTED verdict proves nothing
assert _shell_verdict(pattern, "foo;rm -rf /", tmp_path) == "REJECTED"
```

## Related

- #2126 (CLOSED) — the fix these tests were written to guard.

rajarshidattapy:
Fixes #2641.

## The bug

`tests/test_hooks.py` guards the post-commit hook's interpreter allowlist — the #2126 fix,
and the thing that stops `;`, backticks and `$(...)` reaching a shell. The harness passed the
glob and the candidate to bash as argv elements. On Windows there is no `execve`, so
`subprocess` joins argv into one command string and **bash.exe's MSYS2 runtime re-parses it**,
which rewrites the values before the `case` statement ever runs.

Both directions were broken, which is why it went unnoticed:

- `test_python_detect_allowlists_still_reject_shell_metacharacters` **failed** (4 params) —
  a false alarm claiming the allowlist accepts `foo;rm -rf /`.
- `test_file_path_allowlist_accepts_windows_backslash_path` and
  `test_shebang_allowlist_accepts_windows_backslash_path` **passed vacuously** — the value
  never reached bash intact, so `ACCEPTED` was guaranteed. #2126's regression guard sat green
  on Windows while providing zero coverage of Windows.

The shipped hook is correct; this is purely a test-harness defect. Confirmed independently.

## The issue's suggested fix is not sufficient

Writing the *pattern* to a script file fixes half of it, but the *candidate* still goes
through argv, and it is mangled just as badly:

```
'foo`id`'                             -> 'foouid=1000(asus)'   # substitution EXECUTED
'foo$(id)'                            -> 'foouid=1000(asus)'   # executed
'foo$IFS'                             -> 'foo'                 # expanded
'C:\Users\u\.venv\Scripts\python.exe' -> 'C:Usersu.venvScriptspython.exe'
```

So under that fix: `foo$IFS` arrives as `foo`, which the allowlist correctly **ACCEPTS**, so
that parametrization keeps failing; and the backslash path arrives with no backslashes, so
both #2126 accept tests stay vacuous — the very failure being fixed. (The harness was also
*executing* `id` on every run, since MSYS performs the substitution.)

## Fix

Bake **both** the pattern and the candidate into the script file, the candidate as a
single-quoted shell literal, and invoke `bash case_test.sh` by bare filename from `cwd` — an
absolute Windows path in argv hits the same mangling the script contents just avoided. Nothing
under test transits argv.

This also models production more closely than argv did: the hook is a script file, and the
value being checked is already sitting in a shell variable by the time the allowlist runs.

## Tests

- `_assert_harness_can_reject` — the issue's suggested control, wired into both accept tests.
  An `ACCEPTED` verdict only means something if the pattern can say `REJECTED`.
- `test_shell_verdict_delivers_the_payload_bash_unmodified` — pins the root cause rather than
  the symptom, asserting bash receives the exact bytes. A future regression then names the
  transport instead of looking like an allowlist bug.
- `_shell_verdict` now asserts the verdict is one of `ACCEPTED`/`REJECTED`. As the issue notes,
  `returncode == 0` cannot catch mangling; an empty stdout used to be compared against silently.

## Verifying the guard is no longer vacuous

Pass/fail counts alone don't show this, so I fed the harness a deliberately broken allowlist —
the pre-#2126 pattern with no backslash (`*[!a-zA-Z0-9/_.@-]*`), against
`C:\Users\u\.venv\Scripts\python.exe`, which the test asserts must be `ACCEPTED`:

```
shipped allowlist    old harness -> ACCEPTED  pass   |  new harness -> ACCEPTED  pass
BROKEN  allowlist    old harness -> ACCEPTED  pass   |  new harness -> REJECTED  FAIL
```

The old harness cannot tell a working allowlist from a broken one. The new one also fails the
accept test against the naive `:`-to-`\` range fix that the reject test was written to catch,
so on Windows that regression is now caught from the accept side even where the reject side
passes.

Full suite: 42 pre-existing failures → 38. The four removed are exactly the reported ones; no
new failures.

## Note on the mechanism

`list2cmdline` isn't really the culprit. Its output preserves the pattern; the rewriting
happens when **bash.exe's MSYS2 runtime re-parses the joined command string**, unescaping
backslashes, expanding variables and running command substitution. `MSYS2_ARG_CONV_EXCL` does
not suppress it — verified — which is why keeping the payloads out of argv entirely is the
only reliable option.

### DIFF
--- a/tests/test_hooks.py
+++ b/tests/test_hooks.py
@@ -460,17 +460,69 @@ def _extract_case_pattern(marker: str) -> str:
     raise AssertionError(f"case arm containing {marker!r} not found in _PYTHON_DETECT")
 
 
-def _shell_verdict(pattern: str, candidate: str) -> str:
+def _sh_single_quote(value: str) -> str:
+    """Render ``value`` as a shell single-quoted literal (no expansion at all)."""
+    return "'" + value.replace("'", "'\\''") + "'"
+
+
+def _shell_verdict(pattern: str, candidate: str, tmp_path) -> str:
+    """Run one `case` arm against ``candidate`` in a real bash, and report which
+    branch matched.
+
+    BOTH the pattern and the candidate are baked into a script FILE; neither
+    transits argv. On Windows there is no execve, so ``subprocess`` joins argv
+    into one command string and bash.exe's MSYS runtime re-parses it — which
+    strips backslashes, expands `$IFS`, and *executes* `` `id` `` / `$(id)`
+    before the case statement ever runs. Every value here is either a shell
+    metacharacter payload or a backslash path, i.e. exactly what that layer
+    destroys, so the old `bash -c ... , "_", candidate` form compared bash's
+    verdict on a string that was no longer the one under test (#2641).
+
+    Baking them in also matches production: the hook is a script file, and the
+    value being tested is already sitting in a shell variable by the time the
+    allowlist runs.
+    """
+    script = tmp_path / "case_test.sh"
+    script.write_text(
+        f"CANDIDATE={_sh_single_quote(candidate)}\n"
+        f'case "$CANDIDATE" in\n'
+        f"  {pattern}) echo REJECTED ;;\n"
+        f"  *) echo ACCEPTED ;;\n"
+        f"esac\n",
+        encoding="utf-8",
+        newline="\n",
+    )
+    # Invoke by bare filename from cwd: a Windows absolute path in argv would
+    # hit the same backslash mangling the script contents just avoided.
     result = subprocess.run(
-        ["bash", "-c", f'case "$1" in\n{pattern}) echo REJECTED ;;\n*) echo ACCEPTED ;;\nesac', "_", candidate],
-        capture_output=True, text=True,
+        ["bash", script.name],
+        capture_output=True, text=True, cwd=str(tmp_path),
     )
     # Fail loudly on a malformed case snippet instead of returning "" and
     # producing a confusing ACCEPTED/REJECTED mismatch downstream.
     assert result.returncode == 0, (
         f"bash exited {result.returncode} for pattern {pattern!r}: {result.stderr.strip()}"
     )
-    return result.stdout.strip()
+    verdict = result.stdout.strip()
+    assert verdict in ("ACCEPTED", "REJECTED"), (
+        f"harness produced no verdict for {candidate!r}: {result.stdout!r}"
+    )
+    return verdict
+
+
+def _assert_harness_can_reject(pattern: str, tmp_path) -> None:
+    """Control for the ACCEPT assertions below.
+
+    An ACCEPTED verdict only means something if this pattern is capable of
+    saying REJECTED. When the candidate never reaches bash intact, every input
+    falls through to `*)` and the accept tests pass while testing nothing —
+    which is how #2126's regression guard sat green on Windows while providing
+    zero coverage.
+    """
+    assert _shell_verdict(pattern, "foo;rm -rf /", tmp_path) == "REJECTED", (
+        f"pattern {pattern!r} rejects nothing, so an ACCEPTED verdict proves "
+        f"nothing — the harness is vacuous (#2641)"
+    )
 
 
 @pytest.mark.skipif(shutil.which("bash") is None, reason="bash required to exercise emitted glob")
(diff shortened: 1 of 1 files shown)


anandsoni2 (approved):

safishamsi:
Shipped in v0.9.42 (`graphifyy==0.9.42` on PyPI) (#2126). Stops argv mangling the shell-arg verdict test on Windows. Credited in the release notes. Thanks @rajarshidattapy!

Issue #2478, opened 2026-08-05
Junction/symlink scan roots silently drop out-of-root targets, follow_symlinks does not help, and NTFS junctions bypass the flag entirely


alexrdunbar:
### Summary

A curated scan root built from NTFS junctions went from ~1613 indexed files to 29 on a routine 0.4.11 to 0.9.12 upgrade, with no warning, no error, and no CLI flag able to restore it. Two separate issues combine here.

Related but distinct: #2468 is another way the corpus silently zeroes out. Different root cause, same failure mode (success exit, empty graph, no warning).

### Environment

- graphify (`graphifyy`) 0.9.12, previously 0.4.11
- Windows 11, Python 3.x, scan root is a directory of NTFS junctions

### Setup

A deliberately curated scan root, so the graph covers exactly the knowledge surfaces we care about rather than the whole repo:

```
graphify-corpus/
  claude-agents    -> JUNCTION to ../.claude/agents
  claude-skills    -> JUNCTION to ../.claude/skills
  claude-rules     -> JUNCTION to ../.claude/rules
  claude-commands  -> JUNCTION to ../.claude/commands
  claude-scripts   -> JUNCTION to ../.claude/scripts
  memory           -> JUNCTION to ../memory
  tasks            -> JUNCTION to ../tasks
  docs/            (real directory)
  *.md             (real files)
```

Under 0.4.11 this indexed the full ~1613 files. Under 0.9.12 it indexes 29, the real (non-junction) entries only.

### Issue 1: `follow_symlinks` cannot affect NTFS junctions

Junctions are reparse points, but not symlinks as far as Python is concerned:

```python
>>> p = Path(r"...\graphify-corpus\memory")   # an NTFS junction
>>> os.path.islink(p)
False
>>> p.is_symlink()
False
>>> p.resolve()
WindowsPath('.../claude-code/memory')          # resolve() DOES traverse it
```

Both places `follow_symlinks` is consumed are therefore inert for junctions:

- `os.walk(scan_root, followlinks=follow_symlinks, ...)` (detect.py:1132) already descends junctions regardless of the flag, because `os.walk` only treats `islink()` dirs specially.
- the prune at detect.py:1164, `if child.is_symlink() and not _resolves_under_root(...)`, never fires, because `is_symlink()` is False.

So on Windows the flag is a no-op for the most common way users build a curated scan root.

### Issue 2: the actual exclusion is unconditional, and silent

The real gate is detect.py:1190:

```python
[...]

PR #2648, opened 2026-08-11
test: skip symlink tests instead of erroring when the platform can't create symlinks

### LINKED ISSUE #2642: 15 symlink tests error out instead of skipping on Windows without `SeCreateSymbolicLinkPrivilege`
**Labels:** bug, tests, windows
**Version:** 0.9.40
**Environment:** Windows 11, non-elevated shell, Developer Mode off

## Summary

Creating a symlink on Windows requires either an elevated process or Developer Mode. Fifteen
tests call `Path.symlink_to()` / `os.symlink()` unguarded, so on a default Windows developer
machine they raise `OSError: [WinError 1314] A required privilege is not held by the client`
and are reported as **failures** rather than skips.

This is the single largest block of noise in the suite — 15 of 42 failures on a clean
checkout — which makes the run unusable as a signal on Windows.

## Steps to reproduce

```
$ pytest tests/ -q
...
42 failed, 4120 passed, 16 skipped
$ pytest tests/ -q --tb=line 2>&1 | grep -c "WinError 1314"
30
```

## Affected tests

```
tests/test_detect.py::test_detect_follows_symlinked_directory
tests/test_detect.py::test_detect_follows_symlinked_file
tests/test_detect.py::test_detect_handles_circular_symlinks
tests/test_detect.py::test_detect_default_does_not_auto_follow_direct_symlink_child
tests/test_detect.py::test_detect_explicit_false_overrides_auto_detect
tests/test_detect.py::test_detect_skips_out_of_root_symlinked_directory_even_when_following
tests/test_detect.py::test_detect_skips_out_of_root_symlinked_file_by_default
tests/test_detect.py::test_detect_incremental_propagates_follow_symlinks
tests/test_extract.py::test_collect_files_follows_symlinked_directory
tests/test_extract.py::test_collect_files_skips_out_of_root_symlinked_directory
tests/test_extract.py::test_collect_files_skips_out_of_root_symlinked_file_by_default
tests/test_extract.py::test_collect_files_handles_circular_symlinks
tests/test_atomic_writes.py::test_write_text_atomic_writes_through_symlink
tests/test_image_vision.py::test_read_files_skips_out_of_root_symlink
tests/test_image_vision.py::test_build_image_refs_skips_out_of_root_symlink
```

Representative traceback:

```
    def test_write_text_atomic_writes_through_symlink(tmp_path):
        target = tmp_path / "real.json"
        target.write_text("old", encoding="utf-8")
        link = tmp_path / "link.json"
>       link.symlink_to(target)
E       OSError: [WinError 1314] A required privilege is not held by the client
```

## Expected behavior

Skipped with a clear reason on a machine that cannot create symlinks, so the failure count
reflects real defects.

## Suggested fix

Add a capability-probing fixture to `tests/conftest.py` — probe rather than check
`sys.platform`, since an elevated shell or Developer Mode *can* create symlinks on Windows
and those runs should still get the coverage:

```python
import os, tempfile, pathlib, pytest

@pytest.fixture(scope="session")
def _can_symlink() -> bool:
    with tempfile.TemporaryDirectory() as d:
        src = pathlib.Path(d) / "src"
        src.write_text("x")
        try:
            (pathlib.Path(d) / "lnk").symlink_to(src)
            return True
        except (OSError, NotImplementedError):
            return False


@pytest.fixture
def requires_symlinks(_can_symlink):
    if not _can_symlink:
        pytest.skip("symlink creation unavailable "
                    "(Windows needs elevation or Developer Mode)")
```

then take `requires_symlinks` as a parameter in the 15 tests above.

## Note

This is test infrastructure only — I found no evidence the symlink *handling* code is broken.
Issue #2478 tracks a genuine product defect in that area (junction/symlink scan roots); this
report is strictly about the tests not being skippable.

rajarshidattapy:
Fixes #2642.

## The bug

Creating a symlink on Windows needs an elevated process or Developer Mode. Fifteen tests call
`Path.symlink_to()` / `os.symlink()` unguarded, so on a default Windows developer machine they
raise `OSError: [WinError 1314] A required privilege is not held by the client` and are
reported as **failures**.

At 15 of 42 failures on a clean checkout, this was the single largest block of noise in the
suite — enough that a Windows run isn't usable as a signal.

## Fix

The capability-probing fixture pair from the issue, in `tests/conftest.py`: a session-scoped
`_can_symlink` probe plus a `requires_symlinks` gate, taken as a parameter by the 15 tests.

Probed rather than inferred from `sys.platform`, as the issue specifies — an elevated shell or
Developer Mode *can* create symlinks on Windows, and those runs must keep the coverage.

## Results

```
before:  42 failed, 4120 passed, 16 skipped
after:   27 failed, 4252 passed, 31 skipped
```

The failure-set diff is exact: the 15 converted are precisely the reported ones, and **zero new
failures**.

## Verifying the gate opens as well as closes

A guard that always skips would "fix" the count while silently deleting the coverage — the more
expensive failure of the two. Exercising the fixture functions directly:

```
probe on this machine -> False
  _can_symlink=True  -> runs the test
  _can_symlink=False -> SKIP (symlink creation unavailable on this machine ...)
probe leaked temp entries: none
```

## Scope

There are 17 unguarded symlink call sites, not 15. The extra two are already `skipif`-decorated
for unrelated reasons and were left alone:

| test | existing skip reason |
|---|---|
| `test_build_merge_hyperedges_and_prune.py::test_prune_matches_across_symlinked_root` | "POSIX symlink semantics" |
| `test_watch.py::test_rebuild_code_incremental_rename_preserves_symlink_source_path` | "symlink setup differs on Windows" |

Three further tests (`test_cache.py`, two in `test_detect.py`) already use the repo's existing
try/except + `pytest.skip` idiom. Left as-is rather than churn working code, though
`requires_symlinks` is the better pattern for anyone touching them later.

## Design note

`requires_symlinks` is a parameter rather than a try/except wrapper around each call on purpose:
the guard stays visible in the signature, and an `OSError` raised by the code *under test*
stays a real failure instead of being swallowed into a skip. The probe creates one file symlink
— Windows gates file and directory symlinks behind the same `SeCreateSymbolicLinkPrivilege`, so
a separate directory probe would add nothing.

Test infrastructure only; no product code touched. #2478 remains the tracker for the genuine
junction/symlink scan-root defect.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -1,9 +1,50 @@
 from __future__ import annotations
 
+import tempfile
 from pathlib import Path
 from typing import Any
 
 import pytest
+
+
+@pytest.fixture(scope="session")
+def _can_symlink() -> bool:
+    """Whether this machine can create symlinks at all (#2642).
+
+    Probed rather than inferred from ``sys.platform``: Windows *can* create
+    symlinks from an elevated shell or with Developer Mode enabled, and those
+    runs should still get the coverage. A plain non-elevated Windows shell
+    raises ``OSError: [WinError 1314] A required privilege is not held by the
+    client``, which pytest reports as a FAILURE — 15 of them, drowning out real
+    defects — when what it means is "unsupported here".
+
+    One file symlink is enough to probe: Windows gates file and directory
+    symlinks behind the same ``SeCreateSymbolicLinkPrivilege`` check.
+    """
+    with tempfile.TemporaryDirectory() as d:
+        src = Path(d) / "probe-src"
+        src.write_text("x", encoding="utf-8")
+        try:
+            (Path(d) / "probe-link").symlink_to(src)
+        except (OSError, NotImplementedError):
+            return False
+        return True
+
+
+@pytest.fixture
+def requires_symlinks(_can_symlink) -> None:
+    """Skip a test that must create symlinks when the platform won't allow it.
+
+    Take this as a parameter rather than wrapping each ``symlink_to()`` call in
+    try/except: the guard then sits in the signature where it is visible, and
+    an OSError from the code UNDER test is still a real failure instead of
+    being swallowed into a skip.
+    """
+    if not _can_symlink:
+        pytest.skip(
+            "symlink creation unavailable on this machine "
+            "(Windows requires an elevated shell or Developer Mode)"
+        )
 
 
 @pytest.fixture(autouse=True)
--- a/tests/test_atomic_writes.py
+++ b/tests/test_atomic_writes.py
@@ -55,7 +55,7 @@ def test_write_text_atomic_new_file_respects_umask(tmp_path):
     assert (os.stat(p).st_mode & 0o777) == (0o666 & ~umask)
 
 
-def test_write_text_atomic_writes_through_symlink(tmp_path):
+def test_write_text_atomic_writes_through_symlink(requires_symlinks, tmp_path):
     # Shared-output setups symlink graph.json to shared storage; the atomic write
     # must update the target and keep the link, not replace it with a real file.
     target = tmp_path / "real.json"
--- a/tests/test_detect.py
+++ b/tests/test_detect.py
@@ -296,7 +296,7 @@ def test_git_info_exclude_utf8_bom(tmp_path):
     assert any("real.py" in f for f in all_files)
 
 
-def test_detect_follows_symlinked_directory(tmp_path):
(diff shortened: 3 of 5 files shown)


safishamsi:
Shipped in v0.9.42 (`graphifyy==0.9.42` on PyPI) (#2642). Probe-and-skip symlink tests where unsupported. Credited in the release notes. Thanks @rajarshidattapy!

Issue #1094, opened 2026-05-31
export obsidian: OSError ENAMETOOLONG on long node labels (to_obsidian lacks filename length cap that wiki.py has)


felipeathayde:
## Summary

`graphify export obsidian` crashes with `OSError: [Errno 63] File name too long` when any node has a long label/identifier (~255+ bytes). Reproduced on **v0.8.26** (latest) and **v0.8.14**.

## Root cause

In `graphify/export.py`, `to_obsidian` writes each note with:

```python
fname = node_filename[node_id] + ".md"
(out / fname).write_text("\n".join(lines), encoding="utf-8")  # export.py:927 in v0.8.26
```

`node_filename[node_id]` is derived from the node label without any length cap. When a node label is a long identifier (e.g. a generated symbol name, or a label sourced from a long string), `fname` exceeds the filesystem's 255-byte limit and `write_text` raises `OSError [Errno 63]`.

By contrast, the wiki exporter already guards against this: `graphify/wiki.py` has `_safe_filename(name)` which returns `s[:200]` (caps length at 200 chars, see wiki.py:23). `to_obsidian` does not reuse this sanitizer.

## Reproduce

1. Create a source file containing a symbol with a ~300-char name (e.g. a very long function/class identifier).
2. `graphify update <dir>`
3. `graphify export obsidian <out>`
4. Crash: `OSError: [Errno 63] File name too long: '.../<300-char-name>().md'`

## Suggested fix

Reuse `wiki.py`'s `_safe_filename` (or an equivalent length cap + de-dup) for the Obsidian filename, so `to_obsidian` truncates long names the same way the wiki exporter does.

## Environment

- graphify (PyPI `graphifyy`) v0.8.26 and v0.8.14
- macOS (APFS, 255-byte filename limit)

safishamsi:
Fixed in `690b4e5`, shipping in the next release. Thanks for the precise diagnosis @felipeathayde - the `wiki.py` vs `to_obsidian` comparison made it easy to confirm.

## Fix
Added a shared `_cap_filename(s, limit=200)` helper that both `to_obsidian` and `to_canvas` route their filenames through. Two things beyond a straight `[:200]` cap:

[...]

PR #2656, opened 2026-08-11
fix(export): budget export filenames against the destination path, not just NAME_MAX (#2655)

### LINKED ISSUE #2655: export obsidian / export wiki abort on Windows: the filename cap is sized for NAME_MAX and ignores MAX_PATH, so one long label kills the whole export
## Summary

`graphify export obsidian` and `graphify export wiki` abort with an unhandled `FileNotFoundError` on Windows as soon as one node label is long enough to reach the filename cap. The export dies partway through, leaving a half-written vault (some notes present, the rest missing) and a non-zero exit with a raw traceback.

The cap added in #1094 (`export._cap_filename`, 200 bytes) is sized against the per-component `NAME_MAX` of 255 bytes. That is the correct constraint on POSIX and the wrong one on Windows, where the limit applies to the **whole path**: legacy `MAX_PATH` is 260 characters *including* the terminating NUL, and `HKLM\SYSTEM\CurrentControlSet\Control\FileSystem\LongPathsEnabled` is `0` on a default install. So a 200-byte stem under a perfectly ordinary output directory overruns the ceiling before the filesystem's per-component limit is anywhere near.

`sanitize_label()` caps labels at 256 characters (`security._MAX_LABEL_LEN`), so labels in this range are expected by design — they come out of the semantic pass over docs/PDFs routinely.

## Reproduction

Windows 11, Python 3.12, `v8` @ `26128ab`, default `LongPathsEnabled=0`.

The project lives at `C:\Users\abhay\AppData\Local\Temp\gx\payments-api` (49 chars — shorter than a typical real checkout), and `graph.json` contains one document node whose label is 207 characters:

```
> graphify export obsidian .
  ...
  File "C:\graphify\graphify\cli.py", line 2667, in dispatch_command
    n = _to_obsidian(G, communities, str(obsidian_dir),
  File "C:\graphify\graphify\export.py", line 642, in to_obsidian
    if _owned_write(fname, "\n".join(lines)):
  File "C:\graphify\graphify\export.py", line 560, in _owned_write
    target.write_text(content, encoding="utf-8")
FileNotFoundError: [Errno 2] No such file or directory: 'graphify-out\\obsidian\\Consolidating the payment reconciliation ledger into a single append-only settlement journal so that replayed provider webhooks always keep their original ordering for downstream audit and di_89c6113a.md'
```

`export wiki` fails the same way, and sooner — `graphify-out\wiki\` is written by `wiki._safe_filename`, which caps at 200 *characters* with no hash suffix:

```
> graphify export wiki .
  File "C:\graphify\graphify\wiki.py", line 328, in to_wiki
    (out / f"{slug}.md").write_text(article, encoding="utf-8")
FileNotFoundError: [Errno 2] No such file or directory: 'graphify-out\\wiki\\Consolidating_the_payment_reconciliation_ledger_into_a_single_append-only_settlement_journal_so_that_replayed_provider_webhooks_always_keep_their_original_ordering_for_downstream_audit_and_dispute_res.md'
```

This also reproduces without any corpus at all — three of the repo's own regression tests fail on Windows for exactly this reason on a clean `v8` checkout:

```
> uv run --frozen pytest tests/test_obsidian_filename_cap.py -q
FAILED tests/test_obsidian_filename_cap.py::test_obsidian_long_ascii_label_does_not_crash
FAILED tests/test_obsidian_filename_cap.py::test_obsidian_distinct_long_labels_sharing_prefix_do_not_collide
FAILED tests/test_obsidian_filename_cap.py::test_obsidian_wikilink_resolves_after_truncation
3 failed, 2 passed
```

## Root cause

`export._cap_filename(s, limit=200)` and `wiki._safe_filename(name)` both treat 200 as a constant. Neither knows where the file is about to be written, so the destination directory contributes an unbounded amount to a bounded budget:

```
len("C:\\Users\\me\\projects\\payments-api\\graphify-out\\obsidian")  #  55
+ len("\\")                                                          #   1
+ 200                                                                # stem
+ len("_1")                                                          # dedup suffix
+ len(".md")                                                         #   3
= 261  >  259 usable MAX_PATH characters
```

The `graphify-out\obsidian` / `graphify-out\wiki` suffix alone is 21/17 characters, so the headroom left for the user's own checkout path is about 34 characters. `C:\Users\<name>\Documents\<repo>` exhausts that on its own.

`export wiki` is the more damaging of the two: `AGENTS.md` tells assistants to navigate `graphify-out/wiki/index.md` in preference to raw files, so on Windows that entry point simply never gets written for any graph containing one long label.

## Why the open issues don't already cover this

- **#2109 / PR #2118** — a per-component `NAME_MAX` *below* 255 (eCryptfs). The fix queries `PC_NAME_MAX` through `os.pathconf`, which does not exist on Windows, and states that it "preserves the existing 200-byte behavior on conventional filesystems". Windows stays broken either way; the two fixes are complementary and touch different numbers.
- **PR #2317** — deep paths on the *input* side (corpus discovery and reading). It does not touch `export.py` or `wiki.py`.
- **#1655** — long-path hashing in `save_manifest` / `_md5_file`, also input-side.

## Suggested fix

Derive the stem cap from the destination directory instead of hardcoding it, and reserve room for the pieces callers append after the cap (`_COMMUNITY_` prefix, `_1`/`_2` collision suffixes, `.md`). POSIX keeps returning 200 unchanged, so existing vaults stay byte-for-byte stable and #2109's separate `NAME_MAX` question is untouched.

Capping is preferable to writing through `\\?\` extended-length paths here: the output is meant to be opened by Obsidian, Explorer and other editors, and handing those a 300-character path just moves the failure downstream to the tools the user actually opens the vault with.

I have this working locally with regression coverage and will open a PR against `v8` shortly.

## Environment

| | |
|---|---|
| OS | Windows 11 Home Single Language 26200, `LongPathsEnabled=0` (default) |
| Python | 3.12 (and 3.10) |
| graphify | `v8` @ `26128ab` (0.9.40) |

abhay-codes07:
Fixes #2655.

## Problem

`graphify export obsidian` and `graphify export wiki` abort on Windows with an unhandled `FileNotFoundError` as soon as one node label is long enough to reach the filename cap. The export dies partway through and leaves a half-written vault.

The 200-byte cap from #1094 (`export._cap_filename`, and `wiki._safe_filename`'s `s[:200]`) is sized against the per-component `NAME_MAX` of 255 bytes. On Windows the binding constraint is the **whole path**: `MAX_PATH` is 260 characters *including* the terminating NUL, and `LongPathsEnabled` is `0` on a default install. `graphify-out\obsidian` alone spends 21 of those characters, so a 200-byte stem overruns the ceiling under any ordinary checkout path:

```
len("C:\\Users\\me\\projects\\payments-api\\graphify-out\\obsidian")  #  55
+ 1 (separator) + 200 (stem) + 2 (dedup suffix) + 3 (".md")          # 261
                                                                     # > 259 usable
```

Labels in this range are expected by design — `security.sanitize_label()` caps at 256 characters, and the semantic pass over docs/PDFs produces them routinely.

`export wiki` is the more damaging of the two: `AGENTS.md` tells assistants to navigate `graphify-out/wiki/index.md` in preference to raw files, so on Windows that entry point is simply never written for any graph containing one long label.

## Change

**`graphify/paths.py`** — new shared helper:

```python
stem_filename_budget(output_dir, *, reserve=0, limit=200) -> int
```

Returns `limit` untouched on POSIX. On Windows it returns the smaller of `limit` and whatever still fits inside `MAX_PATH` once the destination directory, the separator, the caller's `reserve` and `.md` are accounted for. Extended-length (`\\?\`) destinations opt out. A floor keeps the budget from going negative — `_cap_filename` would otherwise slice with a negative index and silently emit a garbage stem.

`reserve` covers the pieces callers append *after* capping: the `_COMMUNITY_` prefix on community notes, and the `_1`/`_2` collision suffixes from `_dedup_node_filenames` / `_unique_slug`.

**`graphify/export.py`** — `_obsidian_safe_stem` takes a `limit`; `to_obsidian` and `to_canvas` thread the budget through. `to_canvas` derives it from its own directory (which is the vault directory), because the CLI calls it without the `node_filenames` map — otherwise every canvas card would point at a note that `to_obsidian` wrote under a different name.

**`graphify/wiki.py`** — `_safe_filename` takes a `limit`; `to_wiki` budgets against the wiki directory.

Capping rather than writing through `\\?\` extended-length paths is deliberate: the output is meant to be opened by Obsidian, Explorer and other editors, and handing those a 300-character path just moves the failure downstream to the tool the user actually opens the vault with.

## Scope

This is complementary to, and does not collide with, the other path-length work in flight:

- **#2109 / #2118** address a per-component `NAME_MAX` *below* 255 (eCryptfs) via `os.pathconf(PC_NAME_MAX)`, which does not exist on Windows, and explicitly preserve the 200-byte behaviour on conventional filesystems. Different number, different platform. If #2118 lands first, `limit` here is the natural place for its computed value to arrive — the helper already takes it as a parameter.
- **#2317** covers deep paths on the *input* side and does not touch `export.py` / `wiki.py`.

## Regression coverage

`tests/test_export_path_length.py`, 15 tests. CI runs Ubuntu only, so the suite is written to have teeth there rather than skipping: the budget math is exercised by faking `os.name`, and the exporters' wiring is exercised by forcing a small budget and asserting that wikilinks and canvas card refs still resolve to files that exist. Three end-to-end tests are Windows-gated.

Reverting `graphify/export.py` + `graphify/wiki.py` and keeping the tests fails 9 of the 15, **6 of which run on the Linux runners**.

## Validation

Windows 11 (`LongPathsEnabled=0`), Python 3.12, branched off `26128ab`.

**Full suite, before → after:** `47 failed, 4247 passed` → `44 failed, 4265 passed`. Exactly the three pre-existing `test_obsidian_filename_cap.py` failures are fixed, no new failures. (The remaining 44 are the unrelated pre-existing Windows failures — symlink privileges, `os.geteuid`, and friends.)

```
=== FIXED by this change ===
tests/test_obsidian_filename_cap.py::test_obsidian_distinct_long_labels_sharing_prefix_do_not_collide
tests/test_obsidian_filename_cap.py::test_obsidian_long_ascii_label_does_not_crash
tests/test_obsidian_filename_cap.py::test_obsidian_wikilink_resolves_after_truncation
=== NEW failures introduced ===
(none)
```

**All five `tools.skillgen` validators pass** (`--check`, `--audit-coverage`, `--schema-singleton`, `--monolith-roundtrip`, `--always-on-roundtrip`).

**End-to-end CLI**, same corpus, before → after:

```
> graphify export obsidian .
FileNotFoundError: [Errno 2] No such file or directory: 'graphify-out\obsidian\Consolidating the payment...di_89c6113a.md'
> graphify export wiki .
FileNotFoundError: [Errno 2] No such file or directory: 'graphify-out\wiki\Consolidating_the_payment...dispute_res.md'
```
```
> graphify export obsidian .
Obsidian vault: 4 notes in graphify-out\obsidian/
Canvas: graphify-out\obsidian/graph.canvas
> graphify export wiki .
Wiki: 2 articles written to graphify-out/wiki/
```
with no dangling wikilinks and no canvas cards pointing at missing notes.

**No churn on normal graphs.** Exported graphify's own package (2249 nodes, 4497 edges, 151 communities → 2400 notes + 161 wiki articles) with and without the patch: the exported filename sets are identical, and the canvas resolves cleanly both times. The budget only bites when a label would otherwise have overrun.

### DIFF
--- a/graphify/export.py
+++ b/graphify/export.py
@@ -16,6 +16,7 @@ from networkx.readwrite import json_graph
 from graphify.security import sanitize_label
 from graphify.analyze import _node_community_map
 from graphify.build import edge_data
+from graphify.paths import stem_filename_budget
 
 from graphify.exporters.graphdb import push_to_falkordb, push_to_neo4j  # noqa: E402,F401
 
@@ -466,7 +467,7 @@ def _cap_filename(s: str, limit: int = 200) -> str:
     return f"{truncated}_{digest}"
 
 
-def _obsidian_safe_stem(label: str) -> str:
+def _obsidian_safe_stem(label: str, limit: int = 200) -> str:
     """Filename stem for an Obsidian note / canvas card from a node label.
 
     Strips filesystem-unsafe characters, a trailing ``.md``-family extension
@@ -494,7 +495,15 @@ def _obsidian_safe_stem(label: str) -> str:
     # emit a "@.md"-style filename. (#1409)
     if not re.search(r"\w", cleaned, flags=re.UNICODE):
         return "unnamed"
-    return _cap_filename(cleaned)
+    return _cap_filename(cleaned, limit)
+
+
+# Room _dedup_node_filenames / the community loop need for a collision suffix
+# ("_1" … "_999") appended AFTER the stem was capped.
+_DEDUP_SUFFIX_RESERVE = 4
+
+# Prefix the community overview notes carry ("_COMMUNITY_Backend.md").
+_COMMUNITY_PREFIX = "_COMMUNITY_"
 
 
 def _dedup_node_filenames(G: nx.Graph, safe_name) -> dict[str, str]:
@@ -563,9 +572,16 @@ def to_obsidian(
 
     node_community = _node_community_map(communities)
 
+    # Cap stems against THIS vault's path, not just NAME_MAX: on Windows the
+    # 200-byte default plus an ordinary vault directory overruns MAX_PATH and
+    # every note write raises FileNotFoundError (#2655). No-op on POSIX.
+    _stem_limit = stem_filename_budget(out, reserve=_DEDUP_SUFFIX_RESERVE)
+
     # Map node_id → safe filename so wikilinks stay consistent.
     # Deduplicate: if two nodes produce the same filename, append a numeric suffix.
-    node_filename = _dedup_node_filenames(G, _obsidian_safe_stem)
+    node_filename = _dedup_node_filenames(
+        G, lambda label: _obsidian_safe_stem(label, _stem_limit)
+    )
 
     # Helper: compute dominant confidence for a node across all its edges
     def _dominant_confidence(node_id: str) -> str:
@@ -679,8 +695,13 @@ def to_obsidian(
     # this path had no dedup at all, so even same-case duplicate labels collided.
     community_filename: dict = {}
     used_community: set[str] = set()
+    # The community stem carries the "_COMMUNITY_" prefix on top of the dedup
+    # suffix, so it gets that much less of the MAX_PATH window (#2655).
+    _community_stem_limit = stem_filename_budget(
+        out, reserve=_DEDUP_SUFFIX_RESERVE + len(_COMMUNITY_PREFIX)
+    )
     for cid in communities:
-        base = f"_COMMUNITY_{_obsidian_safe_stem(_community_name(cid))}"
+        base = f"{_COMMUNITY_PREFIX}{_obsidian_safe_stem(_community_name(cid), _community_stem_limit)}"
         candidate = base
         n = 1
         while candidate.lower() in used_community:
@@ -755,7 +776,10 @@ def to_obsidian(
         if cross:
             lines.append("## Connections to other communities")
             for other_cid, edge_count in sorted(cross.items(), key=lambda x: -x[1]):
-                other_fname = community_filename.get(other_cid) or f"_COMMUNITY_{_obsidian_safe_stem(_community_name(other_cid))}"
+                other_fname = community_filename.get(other_cid) or (
+                    f"{_COMMUNITY_PREFIX}"
+                    f"{_obsidian_safe_stem(_community_name(other_cid), _community_stem_limit)}"
+                )
                 lines.append(f"- {edge_count} edge{'s' if edge_count != 1 else ''} to [[{other_fname}]]")
             lines.append("")
 
@@ -853,9 +877,18 @@ def to_canvas(
     # Obsidian canvas color codes (cycle through for communities)
     CANVAS_COLORS = ["1", "2", "3", "4", "5", "6"]  # red, orange, yellow, green, cyan, purple
 
-    # Build node_filenames if not provided (same dedup logic as to_obsidian)
+    # Build node_filenames if not provided (same dedup logic as to_obsidian).
+    # The CLI calls to_canvas without passing the map, so it must derive the
+    # SAME stem budget to keep card links pointing at the notes to_obsidian
+    # wrote — hence budgeting against the canvas's own directory, which is the
+    # vault directory (#2655).
+    _stem_limit = stem_filename_budget(
+        Path(output_path).parent, reserve=_DEDUP_SUFFIX_RESERVE
+    )
     if node_filenames is None:
-        node_filenames = _dedup_node_filenames(G, _obsidian_safe_stem)
+        node_filenames = _dedup_node_filenames(
+            G, lambda label: _obsidian_safe_stem(label, _stem_limit)
+        )
 
     # Fallback: with no community data (e.g. --no-cluster builds or a missing
     # analysis sidecar) the grid below produces nothing and the canvas is written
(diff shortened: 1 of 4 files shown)


abhay-codes07:
@
@safishamsi whenever you have a moment. Two things worth flagging so this does not step on anything:

- It does not overlap #2118. That one lowers the cap when `NAME_MAX` is *below* 255 on POSIX; this one lowers it when the destination directory eats the Windows `MAX_PATH` window. The two limits compose — `stem_filename_budget` already takes `limit` as a parameter, so if #2118 lands first its computed value drops straight in and I will rebase.
- I kept `wiki.py` in the same PR because it fails identically and the fix is the same three lines, but happy to split it out if you would rather review them separately.

I only have Windows to test on, so the POSIX claim is "output is unchanged by construction" (the helper returns early on `os.name != "nt"`) plus the identical-filename check on a 2249-node export, rather than something I ran on Linux myself. Worth a second pair of eyes there.
@

safishamsi:
Landed in v0.9.44, just published to PyPI. I cherry-picked this onto `v8` with your authorship preserved in the commit, so both the change and the credit are in the history. Thanks @abhay-codes07 for the fix. Closing since it is now released.

PR #2657, opened 2026-08-11
fix(wiki): make a link target the article filename verbatim, not its percent-encoded twin (#2597)

### LINKED ISSUE #2597: export wiki: links are percent-encoded but files are written raw, so every article with ( ) & or non-ASCII in its label is linked to a nonexistent path
## Summary

`graphify export wiki` writes article files under the **raw** slug but links to them under the
**percent-encoded** slug, so any article whose label contains `(`, `)`, `&` or a non-ASCII character
is linked at a path that does not exist on disk.

Renderers hide this (GitHub/VS Code decode `%28` before resolving), but the wiki's stated purpose is
to be **agent-crawlable** — and an agent that reads the link target verbatim off disk gets
`FileNotFoundError`.

On my repo: **12 of 105 inter-article links are broken as written** (6 of them in `index.md`).
All 12 resolve after `urllib.parse.unquote`, and 0 targets are genuinely missing — so this is purely
a filename/link mismatch, not lost content.

This looks like a sequel to #1444: that fix correctly moved off `[[wikilinks]]` to standard markdown
links and added `quote()` for renderer safety, but the on-disk name was not normalized to match.

## Version

graphify (`graphifyy`) 0.9.38, Python 3.13, Linux.

## Root cause

`graphify/wiki.py`:

```python
# line 12 - produces the ON-DISK name. Only substitutes Windows-reserved chars;
# ( ) & and non-ASCII pass through untouched.
def _safe_filename(name: str) -> str:
    s = name.replace("/", "-").replace(" ", "_").replace(":", "-")
    s = re.sub(r'[<>:"/\\|?*]', '_', s)
    s = s.strip('. ')
    return s[:200] if s else 'unnamed'

# line 48 - produces the LINK, quoting the same slug.
# urllib.parse.quote(safe='/') encodes ( -> %28, ) -> %29, & -> %26, em-dash -> %E2%80%94
return f"[{text}]({quote(f'{slug}.md')})"
```

The file is written as `{slug}.md` (raw) while the link is `quote(f'{slug}.md')`. The two agree only
when the slug happens to be URL-safe.

## Repro

Any node label containing parens, `&`, or an em-dash. Real examples from my run:

| Link emitted | File actually written |
|---|---|
| `load_traumas%28%29.md` | `load_traumas().md` |
| `find_repo_root%28%29.md` | `find_repo_root().md` |
| `check_command%28%29.md` | `check_command().md` |
| `snapshot_db%28%29.md` | `snapshot_db().md` |
| `Forgejo_upgrade_%26_rollback_%28runbook%29.md` | `Forgejo_upgrade_&_rollback_(runbook).md` |
| `Tailscale_HTTPS_endpoints_%E2%80%94_how_services_get_their_URLs.md` | `Tailscale_HTTPS_endpoints_—_how_services_get_their_URLs.md` |

Function-named nodes (`foo()`) make this common for any code repo.

Check on an existing wiki:

```python
import re, os, urllib.parse
for f in filter(lambda p: p.endswith('.md'), os.listdir('graphify-out/wiki')):
    for l in set(re.findall(r'\]\(([^)]+\.md)\)', open(f'graphify-out/wiki/{f}').read())):
        if not os.path.exists(f'graphify-out/wiki/{l}'):
            print(f, '->', l, '| decoded exists:',
                  os.path.exists(f'graphify-out/wiki/{urllib.parse.unquote(l)}'))
```

## Suggested fix

**Preferred — normalize the filename so `quote()` is a no-op.** Make `_safe_filename` also replace
`(`, `)`, `&` and non-ASCII, so the slug is URL-safe by construction and the link and the file are
the same string. One source of truth, and existing `quote()` stays as harmless defense.

Note this cannot be fixed by simply dropping `quote()`: an unescaped `)` terminates a CommonMark
link destination, so `[x](load_traumas().md)` would parse as a link to `load_traumas(` — worse than
today. (Spaces are already handled, since `_safe_filename` maps them to `_`.)

**Alternative — angle-bracket destinations:** emit `[text](<load_traumas().md>)`. CommonMark allows
parens and spaces inside `<...>`, so the link matches the on-disk name exactly and filenames stay
human-readable. Worth checking Obsidian renders these before choosing it.

Either way, a cheap regression guard would be an assertion in `to_wiki()` that every emitted link
target exists on disk before returning — that turns this whole class of bug into a hard failure at
export time.

## Related

- #1444 — moved to standard markdown links + `quote()`; this is the unfinished half of it.
- #497 — `_safe_filename` collisions causing silent overwrites; same function, adjacent concern.

abhay-codes07:
Fixes #2597.

Thanks @ericziko — the report was precise enough to reproduce first try, and the root cause is exactly where you pointed.

## Reproduced

graphify's own package, 2247 nodes / 4492 edges / 152 communities, `export wiki`:

```
_collect_js_symbol_resolution_facts().md -> _file_stem%28%29.md | decoded exists: True
_extract_generic().md                    -> _make_id%28%29.md   | decoded exists: True
...
--- 27 of 1141 inter-article links broken as written ---
```

Same signature as your run: every failure decodes to a real file, nothing is genuinely missing.

## Change

`_md_link` emitted `quote(f"{slug}.md")` while `to_wiki` wrote the file under the raw slug, so the two agreed only when the label happened to be URL-safe. Rather than encode on one side and hope the other side decodes, this keeps the slug free of anything that would need encoding and emits it raw — one string, not two spellings.

`_safe_filename` now also removes `(` `)` `#` `%` and control characters, on top of the Windows-reserved set it already handled. `_md_link` drops `quote()`.

Two deliberate departures from the suggested fix, both to avoid losing information:

**Parentheses are dropped, not substituted.** Substituting with `_` leaves a trailing `foo__` on every callable node, and collapsing the resulting runs to tidy that up would mangle Python dunders — `__init__()` becomes `_init_`. Dropping keeps `__init__` intact. Labels that then collapse to one slug (`parse()` and `parse`) are still separated by the existing `_unique_slug`, and there is a test for it.

**Non-ASCII is left alone.** The report suggested replacing it, but it is legal raw in a CommonMark link destination and resolves on every filesystem graphify targets. Stripping it would reduce a CJK, Cyrillic or accented wiki to a wall of underscores — and leaving it raw is precisely what fixes your `Tailscale HTTPS endpoints — how services get their URLs` case, since the em dash was only ever broken by the encoding step.

I did not take the angle-bracket option. `[x](<_make_id().md>)` is valid CommonMark, but the naive `\]\(([^)]+\.md)\)` scan an agent is likely to use finds *no* link at all in that form, so it trades one agent-crawlability problem for another.

## The guard that should have caught this

`test_wiki_links_resolve_to_real_files` already asserted "every link target exists on disk" — the exact invariant. It passed all the way through this bug because the `_inline_links` helper ran `urllib.parse.unquote` on the target before checking it, papering over the mismatch it exists to catch. That helper now compares the target as written, which is what turns the existing guard into a real one.

That felt worth fixing in place rather than only adding new tests beside it.

## Tests

`tests/test_wiki_link_filename_parity.py` (15 tests) pins the invariant: for every emitted link, `(wiki_dir / target).exists()` with **no** unquoting step. Parametrized over the character classes from the report — callables, dunders, `&`, `#`, `%`, em dash, CJK, brackets — plus the collision case and a whole-wiki crawl.

Two existing tests in `test_wiki.py` asserted the old encoded targets (`C%23_%26_Auth_%28v2%29.md`, `Array%5BT%5D_Models.md`) while asserting the raw filename on disk in the next line — they were encoding the mismatch itself. Updated to assert byte-identity. Flagging that explicitly since changing existing assertions deserves a second look.

Reverting `graphify/wiki.py` and keeping the tests fails 14 of them, including all the character classes from the report.

## Validation

Windows 11, Python 3.12, branched off `26128ab`.

- Full suite: `47 failed, 4247 passed` → `47 failed, 4262 passed`. **Identical failure set** — no regressions; the +15 are the new tests. (The 47 are pre-existing Windows failures — symlink privileges, `os.geteuid`, and friends — unrelated to this change.)
- All five `tools.skillgen` validators pass.
- End-to-end on the 2247-node graph above: **27 of 1141 broken → 0 of 1141**, with filenames landing as `_make_id.md` rather than `_make_id%28%29.md`.

## Note on #2656

My other open PR (#2656, Windows `MAX_PATH`) also touches `_safe_filename` — it adds a `limit` parameter to the signature, this one changes the body. They are independent bugs and the changes do not interact, but they will conflict textually. Happy to rebase whichever lands second; no preference on ordering.

### DIFF
--- a/graphify/wiki.py
+++ b/graphify/wiki.py
@@ -3,23 +3,47 @@
 from __future__ import annotations
 from collections import Counter
 from pathlib import Path
-from urllib.parse import quote
+import re
 import networkx as nx
 
 from graphify.build import edge_data
 
 
+# Characters a slug may not contain, because the article's LINK and its ON-DISK
+# NAME have to be the same string (#2597). Anything left here must be legal,
+# unescaped, in a CommonMark link destination:
+#   < > : " / \ | ? *   Windows-reserved in filenames (pre-existing set)
+#   ( )                 parentheses delimit/nest a link destination
+#   #                   starts a fragment, so `a#b.md` resolves to the file `a`
+#   %                   reads as the start of a percent-escape
+#   control chars       forbidden in a link destination, hostile in a filename
+#
+# Non-ASCII is deliberately NOT stripped. It is legal raw in a link destination
+# and resolves fine on every filesystem graphify targets; stripping it would
+# reduce a CJK, Cyrillic or accented wiki to a wall of underscores.
+_UNSAFE_SLUG_CHARS = re.compile(r'[<>:"/\\|?*#%\x00-\x1f\x7f]')
+
+
 def _safe_filename(name: str) -> str:
-    """Make a label safe for use as a filename across platforms.
+    """Make a label safe for use as a filename across platforms AND as a
+    markdown link destination.
 
     Substitutes characters that Windows reserves in filenames
-    (< > : " / \\ | ? *) and strips trailing dots/spaces, also reserved.
+    (< > : " / \\ | ? *) plus the ones that would make the emitted link stop
+    matching the file on disk, and strips trailing dots/spaces, also reserved.
     Falls back to 'unnamed' for empty results and caps length at 200
     chars to stay well under common filesystem limits.
+
+    Parentheses are DROPPED rather than substituted: every callable node is
+    labelled ``foo()``, and substituting would leave a trailing ``foo__`` on
+    each of them — and would mangle Python dunders (``__init__()`` ->
+    ``_init_``) if the resulting runs were then collapsed. Dropping keeps
+    ``__init__`` intact. Two labels that collapse to one slug are still
+    separated by ``_unique_slug``.
     """
-    import re
     s = name.replace("/", "-").replace(" ", "_").replace(":", "-")
-    s = re.sub(r'[<>:"/\\|?*]', '_', s)
+    s = s.replace("(", "").replace(")", "")
+    s = _UNSAFE_SLUG_CHARS.sub('_', s)
     s = s.strip('. ')
     return s[:200] if s else 'unnamed'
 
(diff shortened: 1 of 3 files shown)


safishamsi:
Landed in v0.9.44, just published to PyPI. I cherry-picked this onto `v8` with your authorship preserved in the commit, so both the change and the credit are in the history. Thanks @abhay-codes07 for the fix. Closing since it is now released.

PR #2661, opened 2026-08-12
fix(extract): resolve prefixed PHP imports (#2659)

### LINKED ISSUE #2659: PHP: a class whose name is a prefix of sibling class names loses all inbound `imports` edges
**Version:** graphify 0.9.28 · Python 3.10 · macOS · `--code-only`

A class whose name is a **strict prefix** of other class names in the same project loses *all* of its inbound `imports` edges. `affected` then reports `No affected nodes found` for a class that is imported by several files.

This looks related to the namespace/`use`-aware PHP resolution pass added in 0.9.17 for #1923.

## Minimal reproduction

Five files, `graphify extract . --code-only --force`.

**Case A — works.** `Pivot` plus two importers:

```php
// src/Pivot.php
<?php
namespace App\Entities;
class Pivot { public $id; }
```
```php
// src/ModelWithRelation.php
<?php
namespace App\Models;
use App\Entities\Pivot;
class ModelWithRelation
{
    public function rows() { return $this->hasMany(Pivot::class, 'model_id'); }
}
```
```php
// src/ServiceWithTypeHint.php
<?php
namespace App\Services;
use App\Entities\Pivot;
class ServiceWithTypeHint
{
    public function __construct(Pivot $pivot) { $this->pivot = $pivot; }
}
```

```console
$ graphify affected "Pivot" --depth 2
- ModelWithRelation.php [imports] src/ModelWithRelation.php:L4
- ServiceWithTypeHint.php [imports] src/ServiceWithTypeHint.php:L4
- .__construct() [references] src/ServiceWithTypeHint.php:L9

$ graphify explain "Pivot"
  Degree: 4
```

**Case B — broken.** Same `Pivot`, same importer, but add three siblings whose names *begin with* `Pivot`:

```php
// src/PivotRepository.php
<?php
namespace App\Repositories;
interface PivotRepository {}
```
```php
// src/PivotRepositoryEloquent.php
<?php
namespace App\Repositories;
use App\Entities\Pivot;
class PivotRepositoryEloquent implements PivotRepository
{
    public function model() { return Pivot::class; }
}
```
```php
// src/PivotValidator.php
<?php
namespace App\Validators;
class PivotValidator {}
```
```php
// src/ModelWithRelation.php   (unchanged)
<?php
namespace App\Models;
use App\Entities\Pivot;
class ModelWithRelation
{
    public function rows() { return $this->hasMany(Pivot::class, 'model_id'); }
}
```

```console
$ graphify affected "Pivot" --depth 2
No affected nodes found.

$ graphify explain "Pivot"
  Degree: 1
  <-- Pivot.php [contains] src/Pivot.php:L3
```

Both `use App\Entities\Pivot;` statements are gone from the graph. The only surviving edge is the file's own `contains`. Adding the prefix-siblings is the single change between A and B.

## Real-world impact

Hit this on a Laravel 8 codebase (~4.4k nodes). `App\Entities\ConciliacionesBanorteFacturas` is imported by four files, and coexists with `ConciliacionesBanorteFacturasRepository`, `ConciliacionesBanorteFacturasRepositoryEloquent` and `ConciliacionesBanorteFacturasValidator` — the exact shape above, which `prettus/l5-repository` generates for *every* entity, so any Laravel project using that package is affected across its whole data layer.

The importing file still emits its other imports, so the loss is silent and per-target:

```
app_models_conciliacionbanorte --imports--> app_entities_cliente               ✅
app_models_conciliacionbanorte --imports--> app_entities_complementospagos     ✅
app_models_conciliacionbanorte --imports--> app_models_bkcomplementopago       ✅
                              (no edge)     app_entities_conciliacionesbanortefacturas   ❌
```

The practical danger is that `affected` returns an *empty* result rather than an error, which reads as "nothing depends on this" when the honest answer is "unknown" — the failure mode most likely to green-light a bad refactor. It also under-ranks these classes in `god-nodes`.

## Secondary observation

Independently of the above, in Case A the method `.rows()` never appears: a class referenced as `X::class` **inside a method body** produces no method-level `references` edge, while the same class in a constructor signature does. Only the file-level import survives.

That pattern is how every Eloquent relationship is declared — `hasMany`, `belongsTo`, `morphMany`, `morphTo` — so in ORM-heavy code the densest part of the real dependency graph is the least represented. Happy to split this into its own issue if you prefer.

## Expected

- A `use` import always produces an edge, regardless of whether other classes share its name as a prefix.
- If resolution genuinely cannot determine a target, that is reported rather than dropped silently.

Glad to test a patch — the project above is a decent stress case at ~4.4k nodes with heavy Eloquent and `prettus/l5-repository` usage.

ousamabenyounes:
## Why

PHP `use` imports target a bare ID before namespace-aware resolution. When that ID has no sourceless stub node, the resolver cannot recover its label and leaves the import dangling instead of resolving the imported class.

Fixes #2659.

## What changed

- recover a missing PHP import label from the importing file's parsed `use` aliases
- keep matching exact through the existing normalized target ID and FQN resolver
- add a regression test with `Pivot` and three sibling classes whose names begin with `Pivot`
- add the unreleased changelog entry

## Why it is safe

The fallback runs only for `imports` edges without an existing stub label. Other PHP reference relations and non-namespaced legacy resolution are unchanged. External imports continue through the existing FQN stub path.

Upstream `v8` without the production fix:

```text
FAILED tests/test_php_type_resolution.py::test_php_import_resolves_when_target_name_prefixes_sibling_classes
E   AssertionError: assert 'pivot' == 'src_pivot_pivot'
1 failed in 0.23s
```

Patched branch:

```text
tests/test_php_type_resolution.py ...... [100%]
6 passed in 0.27s
changed lines: [2874, 2875, 2876]
executed changed lines: [2874, 2875, 2876]
changed-line coverage: 100%
```

Full local CI replay:

- Python 3.12: 4306 passed, 3 skipped; two pre-existing failures reproduced on unmodified `v8`
- Python 3.10: 4306 passed, 3 skipped; same two pre-existing failures as baseline, with one additional passing regression test
- generated skill checks, install smoke, and `ruff check graphify tests`: passed

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@ Full release notes with details on each version: [GitHub Releases](https://githu
 
 ## 0.9.40 (unreleased)
 
+- Fix: PHP `use` imports now resolve to the intended class when its name prefixes sibling class names (#2659).
 - Fix: the 0.9.37 partial-parse warning no longer fires on valid TypeScript/TSX (#2610, #2599, thanks @Sid-AutoWisdom and @atlasplatformu-ai). tree-sitter-typescript sets an error flag on tiny fully-recovered constructs (a `&` in a JSX string attribute, a semicolon-less `in_*` interface member) that still extract completely; the warning now fires only when recovery plausibly cost symbols (the file yielded at most the file node, or an error region spans multiple lines), so the genuine Kotlin one-line-body and Luau cases still warn.
 - Fix: `file_hash()`'s stat fastpath no longer serves a stale digest when a file is rewritten to the same size within one mtime tick (#2612, thanks @rajarshidattapy); a racily-clean guard falls back to a content hash for recently-modified files.
 - Fix: stored-path absoluteness is now detected cross-platform, so a POSIX-absolute `source_file` from a Linux/CI-built graph no longer leaks into node ids on Windows (#2618, thanks @rajarshidattapy).
--- a/graphify/extractors/resolution.py
+++ b/graphify/extractors/resolution.py
@@ -2871,11 +2871,13 @@ def _resolve_php_type_references(
             continue
         tgt = edge.get("target")
         label = stub_label.get(tgt)
+        uses = uses_by_file.get(ref_file, {})
+        if not label and relation == "imports":
+            label = next((alias for alias in uses if _make_id(alias) == tgt), "")
         if not label:
             continue
         bare = label.strip().lower()
         ns = ns_by_file[ref_file]
-        uses = uses_by_file.get(ref_file, {})
 
         raw = None
         if relation in _PHP_SUPERTYPE_RELATIONS:
--- a/tests/test_php_type_resolution.py
+++ b/tests/test_php_type_resolution.py
@@ -156,3 +156,35 @@ def test_php_plain_no_namespace_inheritance_preserved(tmp_path: Path):
         "no-namespace inheritance must still resolve to the real Base def"
     )
     assert tgt.get("label") == "Base"
+
+
+def test_php_import_resolves_when_target_name_prefixes_sibling_classes(tmp_path: Path):
+    pivot = _write(
+        tmp_path / "src/Pivot.php",
+        "<?php\nnamespace App\\Entities;\nclass Pivot {}\n",
+    )
+    importer = _write(
+        tmp_path / "src/ModelWithRelation.php",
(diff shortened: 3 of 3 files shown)


safishamsi:
Shipped in v0.9.41 (`graphifyy==0.9.41` on PyPI) (#2659). Credited in the release notes. Thanks @ousamabenyounes!

PR #2669, opened 2026-08-12
(feat) LLM Mode - verbose mode and token economics

andersonpem:
Hey guys. This is my very first time opening a PR to Graphify. I'm excited!

I like this project quite a lot. Thank you for the excellent piece of software this project is.

This PR introduces two flags to the Graphify CLI (plus their respective environment variable counterparts)

- The `--verbose` flag for showing verbose outputs of what Graphify has been working on
- The `--tokens` flag for showing only token economics (size of the prompt, input, output tokens used)

I have quite big projects. And my AI APIs are kinda restricted in terms of dashboards, so, calculating if it was viable to use the LLM-backed labelling of Graphify was very mysterious to me. I never knew if I was overworking the models or not. 

Verbose mode will dump the prompts, the thinking process, the data, everything about a LLM run to the stdout. That is good especially for prompt engineering: refining how graphify deals with data, what can be improved, etc.

Tokens mode will only print out the prompt size, input, output tokens and an estimate of cost (in the case of Claude), so the user can have a north of how much that run costed. The cost data is merely informative, an estimate considering current prices. There are a lot of factors that can change the pricing of LLM usage.

If you have any doubts, or if the flags aren't clear enough, let me know.

Have a nice day.
Andy.


andersonpem:
Nice catch, bot! I'll work on that later on the day :)

andersonpem:
From "sight reading", I'd say the one that I can work on is the "Env-enabled token tracing misses the run total". It's more of a cosmetic thing, but I can add it.
The first one, each batch dict is merged in the main thread only.
The third one, the setters run only during startup, before any thread pool exists.

andersonpem:
Have a look and if you have any doubts let me know

Issue #1016, opened 2026-05-25
TypeError in export wiki: '<' not supported between str and NoneType when source_file is None


guilherme-campanholi:
## Version

graphify 0.8.15

## Command

```
graphify export wiki
```

## Error

```
File ".../graphify/wiki.py", line 57, in _community_article
    sources = sorted({G.nodes[n].get("source_file", "") for n in nodes} - {""})
TypeError: '<' not supported between instances of 'str' and 'NoneType'
```

## Root Cause

Some nodes in `graph.json` have `source_file: null` (not `""` or missing key). The set comprehension uses `.get("source_file", "")` which only applies the default for **missing keys** — when the key exists with value `null`/`None`, it returns `None`. Subtracting `{""}` does not remove `None`, so `sorted({"a", None, "b"})` raises `TypeError`.

This happens after an incremental extraction where some semantic chunks fail mid-run (e.g. credit exhaustion, rate limit), leaving partially-extracted nodes with `source_file=None`.

## Fix (one-liner)

```python
# wiki.py line 57 — before:
sources = sorted({G.nodes[n].get("source_file", "") for n in nodes} - {""})

# after:
sources = sorted({(G.nodes[n].get("source_file") or "") for n in nodes} - {""})
```

Using `or ""` coerces both `None` and `""` to `""`, which is then removed by `- {""}`. Sorting the result is always safe.

## Workaround applied locally

Patched `wiki.py` in the venv as above. `graphify export wiki` now runs successfully with the patched version.

## Reproducibility

Trigger a full extraction (`graphify . --backend claude`) and force an early abort (e.g. exhaust API credits mid-run). Some nodes will have `source_file: null` in the resulting `graph.json`. Then run `graphify export wiki` — it crashes.

safishamsi:
Fixed in 4e80d86 — changed `.get('source_file', '')` to `(... or '')` in wiki.py so `None` values are normalized before `sorted()` is called.

Issue #323, opened 2026-04-13
AttributeError in _score_nodes when source_file is None


ross-jill-ws:
## Bug

`graphify query` crashes with an `AttributeError` when a node in the graph has `source_file` explicitly set to `None`.

## Traceback

```
Traceback (most recent call last):
  File ".pyenv/versions/3.11.6/bin/graphify", line 8, in <module>
    sys.exit(main())
  File ".pyenv/versions/3.11.6/lib/python3.11/site-packages/graphify/__main__.py", line 510, in main
    scored = _score_nodes(G, terms)
  File ".pyenv/versions/3.11.6/lib/python3.11/site-packages/graphify/serve.py", line 38, in _score_nodes
    source = data.get("source_file", "").lower()
AttributeError: 'NoneType' object has no attribute 'lower'
```

## Root cause

In `serve.py` line 37-38, `data.get("source_file", "")` returns `None` when the key exists but its value is `None`. The `.get()` default only applies when the key is **missing**, not when it's present with a `None` value.

```python
# Current (broken)
label = data.get("label", "").lower()
source = data.get("source_file", "").lower()

# Fix
label = (data.get("label") or "").lower()
source = (data.get("source_file") or "").lower()
```

## Steps to reproduce

1. Build a graph where at least one node has `source_file: None` (e.g. a synthesized or community node)
2. Run `graphify query "any search term"`

## Environment

- Python 3.11.6
- graphify installed via pip

safishamsi:
Fixed in **v0.4.13**.

Three locations in the MCP server were calling `.lower()` on values that could be `None` — `source_file`, `label` in `_score_nodes`, and `label` in `_find_node` and `_tool_get_node`. All three now guard with `or ""` before the string operation.

```bash
pip install -U graphifyy
```

PR #2674, opened 2026-08-12
fix: benchmark crashes on nodes with a None label

Arthuro0103:
## What

`graphify/benchmark.py:42` guards a node label with `data.get("label", "")`. That default only applies when the key is **missing** — a node that carries `label` with a stored value of `None` returns `None`, which then hits `.lower()`:

```
AttributeError: 'NoneType' object has no attribute 'lower'
```

`_query_subgraph_tokens` iterates every node before returning, so a single such node takes down the whole benchmark rather than just scoring itself as zero.

## Why it's still here

This exact guard was already rolled out elsewhere:

- `serve.py` — all `.get("label", "")` calls guarded with `or ""` (#323)
- wiki `source_file` path (#1016)

`benchmark.py` was missed by both passes. On `v8` it is the last remaining occurrence of the unguarded pattern in the query path.

## How it shows up

Nodes with `source_file: null` / `label: null` are produced by entity extraction when no source file resolves. Hit in the wild on a 6028-node graph where 6 nodes had a null attribute — enough to break every query against it.

## Change

```diff
-        label = data.get("label", "").lower()
+        label = (data.get("label") or "").lower()
```

Plus a regression test in `tests/test_benchmark.py` that adds a `label=None` node to the existing fixture graph.

## Verification

- `pytest tests/test_benchmark.py` → **17 passed**
- Reverting the one-line guard makes the new test fail with the exact `AttributeError` at `benchmark.py:42` — confirmed red before green
- Full suite not run locally (`tests/test_dedup.py` needs `rapidfuzz`, not installed in my environment); leaving that to CI

### DIFF
--- a/graphify/benchmark.py
+++ b/graphify/benchmark.py
@@ -39,7 +39,7 @@ def _query_subgraph_tokens(G: nx.Graph, question: str, depth: int = 3) -> int:
     terms = _query_terms(question)
     scored = []
     for nid, data in G.nodes(data=True):
-        label = data.get("label", "").lower()
+        label = (data.get("label") or "").lower()
         score = sum(1 for t in terms if t in label)
         if score > 0:
             scored.append((score, nid))
--- a/tests/test_benchmark.py
+++ b/tests/test_benchmark.py
@@ -54,6 +54,21 @@ def test_query_keeps_short_non_english_terms():
     assert tokens > 0
 
 
+def test_query_handles_node_with_none_label():
+    """A node can carry `label` with a None VALUE, not just a missing key.
+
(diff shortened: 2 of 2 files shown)


safishamsi:
Fixed in v0.9.41 (`graphifyy==0.9.41` on PyPI). `graphify benchmark` no longer crashes on a node whose label is `None`. Thanks @Arthuro0103.

safishamsi:
Shipped in v0.9.41 (`graphifyy==0.9.41` on PyPI). Credited in the release notes. Thanks @Arthuro0103!

PR #2676, opened 2026-08-12
fix(csharp): strip call-site type arguments from generic calls

### LINKED ISSUE #2624: C#: explicit type arguments at a call site (`X.M<T>(...)`) drop the `calls` edge
### Summary

In C#, any member call written with an explicit type argument list at the **call site** produces no
`calls` edge, even though caller and target member both exist as nodes and the receiver is named
explicitly or is an explicitly typed local. The type arguments end up inside the callee name, so the
member lookup key never matches.

What decides it is the **call-site syntax**, not the declaration: a generic method called *without*
an explicit type argument (type inference) resolves fine. Static vs. instance receiver, same file vs.
another file, `out var`, one or two type arguments, block- vs. expression-bodied caller make no
difference.

This is long-standing, not a regression: identical behaviour on 0.9.8, 0.9.20, 0.9.34 and 0.9.39
(same repro, same 2 of 5 edges).

### Minimal repro

Five files, no build system needed. (The snippet compiles as-is: `dotnet build` on net8.0 reports
0 errors, 0 warnings.)

`Payload.cs`
```csharp
namespace Demo
{
    public class Payload
    {
        public int Size;
    }
}
```

`Registry.cs`
```csharp
namespace Demo
{
    public static class Registry
    {
        // generic
        public static bool Fetch<TItem>(string key) { return true; }

        // non-generic
        public static bool Has(string key) { return true; }

        // generic, but callable without an explicit type argument
        public static bool Pick<TItem>(TItem item) { return true; }
    }
}
```

`Box.cs`
```csharp
namespace Demo
{
    public class Box
    {
        public bool Read<TItem>(string key) { return true; }
    }
}
```

`Bag.cs`
```csharp
namespace Demo
{
    public static class Bag<TItem>
    {
        public static bool Make(string key) { return true; }
    }
}
```

`Consumer.cs`
```csharp
namespace Demo
{
    public class Consumer
    {
        // 1) static call to a GENERIC method
        public bool A() { return Registry.Fetch<Payload>("k"); }

        // 2) instance call to a GENERIC method, receiver explicitly typed
        public bool B()
        {
            Box box = new Box();
            return box.Read<Payload>("k");
        }

        // 3) control: static call to a NON-generic method
        public bool C() { return Registry.Has("k"); }

        // 4) control: generic method, call site has NO type argument
        public bool D()
        {
            Payload item = new Payload();
            return Registry.Pick(item);
        }

        // 5) receiver is a generic type, method itself is not generic
        public bool E() { return Bag<Payload>.Make("k"); }
    }
}
```

```
graphify extract . --code-only --no-cluster
```

### Expected

```
calls  Consumer.A()  ->  Registry.Fetch()
calls  Consumer.B()  ->  Box.Read()
calls  Consumer.C()  ->  Registry.Has()
calls  Consumer.D()  ->  Registry.Pick()
calls  Consumer.E()  ->  Bag.Make()
```

### Actual (0.9.39, and identically on 0.9.34 / 0.9.20 / 0.9.8)

```
calls  Consumer.C()  ->  Registry.Has()
calls  Consumer.D()  ->  Registry.Pick()
```

All ten member nodes are present in the graph (listed here with their owning class node, the labels
themselves are `.Name()`), so this is purely the match between call site and member:

```
Bag.Make()  Box.Read()  Registry.Fetch()  Registry.Has()  Registry.Pick()
Consumer.A()  Consumer.B()  Consumer.C()  Consumer.D()  Consumer.E()
```

Cases 3 and 4 are the informative controls: 4 shows that a *generic declaration* is fine, so the
trigger is the `<...>` written at the call site.

### Mechanism (read in 0.9.39)

`_CSHARP_CONFIG` (extract.py:889) resolves a call's callee through `call_accessor_field="name"`, i.e.
the `name` field of the `member_access_expression`. For a generic call that field is not an
`identifier` but a `generic_name` node, and the whole node text is taken verbatim
(extractors/engine.py:4769-4771):

```python
attr = func_node.child_by_field_name(config.call_accessor_field)
if attr:
    callee_name = _read_text(attr, source)      # -> "Fetch<Payload>", not "Fetch"
```

`_resolve_csharp_member_calls` then normalises names by dropping every non-alphanumeric character
(extract.py:3055):

```python
def _key(label: str) -> str:
    return re.sub(r"[^a-zA-Z0-9]+", "", str(label)).lower()
```

The method index is built from member labels (extract.py:3084), so the target key is
`_key(".Fetch()") == "fetch"`, while the call site yields `_key("Fetch<Payload>") == "fetchpayload"`.
The lookup at extract.py:3206 therefore misses and the call is dropped:

```python
method_nid = _method_on_type_or_bases(type_nid, _key(callee))
if not method_nid:
    continue  # receiver typed, but the type has no such method — skip
```

Because the type arguments are folded into the key by `_key`, the mismatch is systematic: every
distinct type argument produces a distinct key, and none of them can ever match a member label.
The extractor already handles `generic_name` correctly on the *type-reference* side —
`_read_csharp_type_name` and `_csharp_collect_type_refs` (both in extractors/engine.py) descend into
the node's `name` field — so the callee side is the outlier.

**Case 5 is a second, independent spot** with the same root cause on the receiver side: the receiver
is only captured when it is a plain `identifier` (extractors/engine.py:4778-4780), so a
`generic_name` receiver such as `Bag<Payload>` leaves `member_receiver` unset, and the
receiver-typed resolver bails on `if not receiver: continue`.

### Impact measured on our corpus

Counted, not sampled — over the complete `.cs` file set of eight graphs we build from one product's
source trees, 10,331 distinct files: **272 lost `calls` edges**, from 13,611 call sites carrying the
shape. Each of the 272 was checked pair-level (graphify de-duplicates call edges per caller/target
pair, so a per-line check would overstate the loss): no edge exists between the enclosing member and
the target member, not even via another call. 49 further shape matches are deliberately **not**
counted, because the calling member has no node of its own — mostly property accessors and
constructors — so no edge could exist there for reasons unrelated to this issue.

The 272 come from 26 distinct (type, method) pairs, every one checked against its declaration in the
source. They cluster on a few static registry/lookup helpers: the four largest account for 68, 60, 56
and 35 lost call sites. Those four are service locators, i.e. precisely the nodes a "who uses this
service?" traversal depends on, so the loss costs more than its share of the graph suggests. Case-5
shape (generic receiver): 20 call sites, 8 confirmed missing edges.

A fresh 0.9.39 build of the graph carrying 116 of those losses reproduces the same 116, so the number
is not an artifact of the version we run in production.

### One caution for the fix

Please strip the type arguments only where the callee is matched against a **receiver-typed**
member index — not for the evidence-free bare-name path. Reason, measured in the same corpus:
receiver resolution already prefers a same-named in-corpus class over an imported external type, so
relaxing the callee lookup alone would turn silence into confidently wrong edges.

Measured on 0.9.39 with a neutral three-file repro (`caution/` in the attachment): a corpus class
`Corp.Internal.TextCodec` plus two callers of `TextCodec.Encode(...)` — one in `Corp.Internal`
(control), one in `Corp.App`, which imports a foreign namespace and does *not* import
`Corp.Internal`. **Both** get an `EXTRACTED 1.0` edge to the corpus class, including the one that
means the imported foreign type.

That fallback is documented, so it is not itself news. The point is what a name-only fix would add:
our corpus holds **9 generic call sites** whose receiver name exists in-corpus but is not visible
from the calling file. Five of the nine we verified line by line — they call a `System.Text.Json`
API, all five files carry `using System.Text.Json;` and none mentions the unrelated in-corpus
namespace that happens to declare a same-named class, so all five would turn into wrong edges. Two
of those five files already carry wrong edges today from their *non-generic* calls to the same API,
pointing into that unrelated internal class, whose same-named members are instance methods with
unrelated signatures. The remaining four sites (three other receiver names) we did not verify
individually.

The calling file's `using` set plus its own namespace would be enough to disambiguate; that
information is already available to `CsharpNameResolver`.

### Environment

Windows 11, Python 3.14.6, graphifyy 0.9.39 (also checked on 0.9.34, 0.9.20 and 0.9.8),
`graphify extract --code-only --no-cluster`.

rohit-jsfreaky:
Fixes #2624.

## Summary

A C# call written with explicit type arguments produced no `calls` edge, while the same method called with an inferred type argument resolved fine — so the call-site syntax decided it, not the declaration.

tree-sitter wraps a name carrying a type-argument list in a `generic_name` node. The C# call handler read that node verbatim, so the callee became `Fetch<Payload>` while its declaration is stored bare as `.Fetch()`; the lookup key never matched and the edge was dropped silently.

A second, independent defect sat next to it: in `Bag<Payload>.Make()` the *receiver* is a `generic_name`, and a receiver was only captured when it was a plain `identifier` — so it was dropped entirely and `_resolve_csharp_member_calls` had nothing to bind.

## Changes

- Add `_csharp_name_without_type_args()`, returning the bare name of a call-site name node. It reads the `name` field first and falls back to the first `identifier` child — the same order `_csharp_collect_type_refs` uses (the pinned grammar exposes no `name` field on `generic_name`; trying the field first keeps this working on versions that do).
- Apply it to the callee in `recv.M<T>()`, to a `generic_name` receiver (`Bag<Payload>.Make()`), and to an unqualified call (`Local<Payload>()`), which previously fell through to the raw-text scan and kept the `<...>`.

C# only — all three call sites sit inside the `tree_sitter_c_sharp` branch and the helper is used nowhere else. A node that is not a `generic_name` returns `_read_text` unchanged, so non-generic calls take exactly the path they did before.

This does not loosen resolution. Stripping the type-argument list only makes the lookup key match; the call still goes through receiver typing and the god-node guard, so an ambiguous or untypable receiver still yields no edge rather than a wrong one. A qualified generic (`Demo.Bag<Payload>`) has no bare `identifier` child and keeps its previous name minus the type arguments, so nothing that used to be recorded stops being recorded.

## Reproduction

The issue's five files, `graphify extract . --code-only --no-cluster`:

| call | before | after |
| --- | --- | --- |
| `A() -> Registry.Fetch<Payload>()` | missing | resolved |
| `B() -> box.Read<Payload>()` | missing | resolved |
| `C() -> Registry.Has()` (control) | resolved | resolved |
| `D() -> Registry.Pick(item)` (control) | resolved | resolved |
| `E() -> Bag<Payload>.Make()` | missing | resolved |

2 of 5 edges before, 5 of 5 after. Two further shapes were broken by the same cause and are not in the report — both now covered: `Local<Payload>("k")` (unqualified, no receiver) and `this.Local<Payload>("k")`.

## Tests

Five tests appended to `tests/test_csharp_member_calls.py`. Each was confirmed to fail with the source change reverted and the tests kept, and to pass with it.

One is a guard rather than a repro: a generic call on a typed field must still bind to that field's type and not to a same-named method on an unrelated class (#1609), now exercised through a generic call. The constructed-generic-receiver test is split across two files deliberately — in a single file the callee resolves by an in-file label match and never reaches receiver typing, so a same-file version passes even with the receiver dropped.

`uv run --no-sync pytest tests/test_csharp_member_calls.py tests/test_csharp_type_resolution.py tests/test_csharp_partial_classes.py tests/test_dotnet.py -q` — 126 passed.

`uv run --no-sync ruff check graphify/extractors/engine.py tests/test_csharp_member_calls.py` — passed.

Full suite: 22 failed, 4304 passed, 13 skipped, against a clean `v8` baseline of 22 failed, 4299 passed, 13 skipped — the same 22 in both runs, all pre-existing and platform-sensitive on this Windows machine, none touching this path. The +5 are the new tests.

`graphify update .` re-run per AGENTS.md: 765 files, 13,426 nodes, no errors.

Tested against the pinned `tree_sitter_c_sharp` 0.23.5 on Windows; other grammar versions and platforms were not exercised.

### DIFF
--- a/graphify/extractors/engine.py
+++ b/graphify/extractors/engine.py
@@ -2594,6 +2594,37 @@ def _has_multiline_error(root) -> bool:
     return False
 
 
+def _csharp_name_without_type_args(node, source: bytes) -> str | None:
+    """The bare name of a C# call-site name node, with any type-argument list off.
+
+    A call site may spell explicit type arguments -- ``Registry.Fetch<Payload>(...)``,
+    ``Bag<Payload>.Make(...)``, ``Local<Payload>(...)`` -- and tree-sitter wraps those
+    in a ``generic_name``. Declarations store the bare name (``.Fetch()``, ``Bag``),
+    so reading the node verbatim yields ``Fetch<Payload>``, which never matches its
+    declaration and silently drops the ``calls`` edge (#2624). Type inference at the
+    call site was unaffected, which is why only the explicit form broke.
+
+    The grammar exposes no ``name`` field on ``generic_name``, so fall back to the
+    first ``identifier`` child -- the same order ``_csharp_collect_type_refs`` uses.
+    """
+    if node is None:
+        return None
+    if node.type != "generic_name":
+        return _read_text(node, source)
+    name_child = node.child_by_field_name("name")
+    if name_child is None:
+        for sub in node.children:
+            if sub.type == "identifier":
+                name_child = sub
+                break
+    if name_child is not None:
+        return _read_text(name_child, source)
+    # A qualified generic (`Demo.Bag<Payload>`) has no bare `identifier` child. Keep
+    # the text and drop only the type-argument list, so such a call is still recorded
+    # under the name it had before rather than disappearing.
+    return _read_text(node, source).split("<", 1)[0] or None
+
+
 def _read_csharp_type_name(node, source: bytes) -> tuple[str, bool, str] | None:
     """Resolve a C# type name, whether it was qualified, and its qualifier prefix."""
     if node is None:
@@ -5095,10 +5126,16 @@ def _extract_generic(
                     mname = fn_node.child_by_field_name("name")
                     recv = fn_node.child_by_field_name("expression")
                     if mname is not None:
-                        callee_name = _read_text(mname, source)
+                        # `name` is a generic_name when the call site spells explicit
+                        # type arguments (`recv.Read<Payload>()`); the declaration is
+                        # stored bare, so strip the type-argument list (#2624).
+                        callee_name = _csharp_name_without_type_args(mname, source)
                         is_member_call = True
-                        if recv is not None and recv.type == "identifier":
-                            member_receiver = _read_text(recv, source)
+                        if recv is not None and recv.type in ("identifier", "generic_name"):
+                            # generic_name: a constructed generic type as the receiver
+                            # (`Bag<Payload>.Make()`). Type nodes are stored under the
+                            # bare name, so this binds like any `Type.M()` (#2624).
+                            member_receiver = _csharp_name_without_type_args(recv, source)
                         elif recv is not None and recv.type in ("this", "this_expression"):
                             member_receiver = "this"
                         elif recv is not None and recv.type in ("base", "base_expression"):
@@ -5119,8 +5156,11 @@ def _extract_generic(
                                 and fname.type == "identifier"
                             ):
                                 member_receiver = _read_text(fname, source)
-                elif fn_node is not None and fn_node.type == "identifier":
-                    callee_name = _read_text(fn_node, source)
+                elif fn_node is not None and fn_node.type in ("identifier", "generic_name"):
+                    # generic_name: an unqualified call carrying explicit type
+                    # arguments (`Local<Payload>()`), which otherwise fell through to
+                    # the raw-text fallback below and kept the `<...>` (#2624).
+                    callee_name = _csharp_name_without_type_args(fn_node, source)
                 else:
                     # Fallback: original name-field / first-named-child scan.
                     name_node = node.child_by_field_name("name")
--- a/tests/test_csharp_member_calls.py
+++ b/tests/test_csharp_member_calls.py
@@ -776,3 +776,101 @@ def test_struct_primary_constructor_parameter_emits_references_edge(tmp_path):
     idep = _find(r, "IDep", "idep")
     assert (holder, idep) in _refs(r), \
         "a struct primary-constructor parameter type must produce a references edge"
+
+
+# --- #2624: explicit type arguments at the call site -------------------------
+# `X.M<T>(...)` spells a type-argument list, which tree-sitter wraps in a
+# `generic_name`. Declarations are stored bare (`.Fetch()`, `Bag`), so reading the
+# node verbatim gave `Fetch<Payload>` and the lookup never matched — the edge
+# silently vanished. Type inference at the call site (`Registry.Pick(item)`) was
+# unaffected, which is why only the explicit form broke.
+
+_GENERIC = (
+    "public class Payload { public int Size; }\n"
+    "public static class Registry {\n"
+    "    public static bool Fetch<TItem>(string key) { return true; }\n"
+    "    public static bool Has(string key) { return true; }\n"
+    "}\n"
+    "public class Box { public bool Read<TItem>(string key) { return true; } }\n"
+    "public static class Bag<TItem> { public static bool Make(string key) { return true; } }\n"
+)
+
+
+def test_explicit_type_argument_on_type_receiver_call(tmp_path):
+    """`Type.M<T>()` binds like `Type.M()` — type arguments are not part of the
+    method name, and the non-generic control must keep working (#2624)."""
+    calls, r = _calls(tmp_path, {"S.cs": _GENERIC + (
+        "public class Consumer {\n"
+        "    public bool A() { return Registry.Fetch<Payload>(\"k\"); }\n"
+        "    public bool C() { return Registry.Has(\"k\"); }\n"
+        "}\n"
+    )})
+    a, c = _find(r, ".A()", "consumer"), _find(r, ".C()", "consumer")
+    assert (a, _find(r, ".Fetch()", "registry")) in calls, \
+        "explicit type arguments must not drop the calls edge"
+    assert (c, _find(r, ".Has()", "registry")) in calls, "control: non-generic call"
+
+
+def test_explicit_type_argument_on_typed_local_receiver(tmp_path):
+    """A receiver typed through the method-scoped table still resolves when the
+    call carries explicit type arguments (#2624)."""
+    calls, r = _calls(tmp_path, {"S.cs": _GENERIC + (
+        "public class Consumer {\n"
+        "    public bool B() { Box box = new Box(); return box.Read<Payload>(\"k\"); }\n"
+        "}\n"
+    )})
+    b = _find(r, ".B()", "consumer")
+    assert (b, _find(r, ".Read()", "box")) in calls
+
+
+def test_constructed_generic_type_as_receiver(tmp_path):
+    """`Bag<Payload>.Make()` — the RECEIVER carries the type arguments. Type nodes
+    are stored under the bare name, so it must bind like any `Type.M()` (#2624).
+
+    Split across two files on purpose: in a single file the callee resolves by a
+    plain in-file label match and never reaches receiver typing, so a same-file
+    version of this test passes even with the receiver dropped.
+    """
+    calls, r = _calls(tmp_path, {
+        "Lib.cs": _GENERIC,
+        "Consumer.cs": (
+            "public class Consumer {\n"
+            "    public bool E() { return Bag<Payload>.Make(\"k\"); }\n"
+            "}\n"
+        ),
+    })
+    e = _find(r, ".E()", "consumer")
+    assert (e, _find(r, ".Make()", "bag")) in calls
+
+
+def test_unqualified_generic_call_and_this_receiver(tmp_path):
+    """`Local<T>()` and `this.Local<T>()` — the same defect reached calls with no
+    receiver at all, which fell through to the raw-text scan and kept the `<...>`."""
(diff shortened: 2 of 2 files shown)


JensD-git:
Thank's again. Hope this PR will be released soon.

JensD-git:
How did this pull request backlog come about? Is there anything we can do to help here? It's unusual for completed items to sit here for two weeks.

rohit-jsfreaky:
Rebased onto `v8` (0.9.49) — conflict resolved. It was append-only in `tests/test_csharp_member_calls.py`: the primary-constructor tests from #2836 and these #2624 tests both land at the end of the file, so both blocks are kept, upstream's first. `engine.py` rebased untouched.

I also re-checked that this is not already covered by #2911 / #2913, since those mention generic type arguments too. Those walk them in field / property / return / parameter position, which is type-reference resolution; the call-site callee and receiver naming is a separate path. The repro still fails the same way on 0.9.49 — 3 of 8 shapes resolve without this change, 8 of 8 with it.

Full suite on the rebased branch: 12 failed, 4986 passed, 43 skipped — the same 12 as a clean `v8` checkout on this Windows machine (fifo / unix-socket / hermes / gemini path tests), none touching this change. All four C# suites pass (132).

JensD-git:
Are there any plausible reasons why this hasn't been merged in weeks? The PR looks fine to me.

JensD-git:
I reported #2624. Instead of asking about timing again, here is a measurement, because the two
questions that would actually decide this PR are answerable and nobody had answered them.

**1. The defect is still in the shipped release.** I ran a minimal reproducer against `v8` at
`c9f9901`, which is exactly release v0.9.55. Eight call sites, five generic, three non-generic
controls:

| Case | Call site | v0.9.55 | + this PR |
|---|---|---|---|
| A | `Registry.Fetch<Payload>("k")` — static generic method | missing | edge |
| B | `box.Read<Payload>("k")` — instance generic method | missing | edge |
| C | `Registry.Has("k")` — control | edge | edge |
| D | `Registry.Pick(item)` — control, generic method, no type arg at call site | edge | edge |
| E | `Bag<Payload>.Make("k")` — generic receiver | missing | edge |
| F | `Own<Payload>("k")` — unqualified | missing | edge |
| G | `this.Own<Payload>("k")` — unqualified via `this` | missing | edge |
| H | `Plain("k")` — control, unqualified | edge | edge |

**3 of 8 before, 8 of 8 after**, controls unchanged. Every target member node already exists, so
nothing is missing from extraction — only the call-site-to-member match fails. Cases **F and G are
the two forms named in the first comment on #2624 and were not in the original reproducer**; this
is the first confirmation of them.

**2. The patch still applies to the current head.** `git cherry-pick` of `pull/2676/head` onto
`v8` at `c9f9901` merges clean — `graphify/extractors/engine.py` auto-merged, no conflict, no
second rebase needed.

**3. The red tests are pre-existing and environmental, not caused by this PR.** The rebase note
mentions a failing count, which is the kind of thing that quietly stalls a merge, so I measured
both sides on one machine, one interpreter, Windows, full suite:

| | failed | passed | skipped |
|---|---|---|---|
| `v8` @ `c9f9901`, unpatched | **17** | 5383 | 45 |
| same tree + this PR | **17** | 5388 | 45 |

The two failing sets are **name-for-name identical** (diff of the two `FAILED` lists is empty), and
`passed` rises by exactly 5 — the five tests this PR adds. Every one of the 17 is platform-bound:
`os.mkfifo` / `os.symlink` unavailable (`tests/test_non_regular_files.py`, 5), Windows holding a
directory handle so `rmdir` raises `WinError 32` (`tests/test_watch.py`, 2), and POSIX path,
shebang and install-destination assertions (`test_hooks.py`, `test_install*.py`,
`test_uninstall_scope.py`, `test_languages.py`, `test_merge_chunks_validation.py`, 10). None of
them touches extraction. `tests/test_csharp_member_calls.py` + `tests/test_dotnet.py`: **89 passed**
with the patch.

**4. One honest caveat, which I raised in the issue myself.** A generic call site whose receiver
name exists in the corpus but is *not visible* from the calling file now gets an edge where it
previously stayed silent. Measured on a three-file case (a corpus `Corp.Internal.TextCodec`, a
caller in `Corp.App` that imports a foreign namespace of the same type name, and a control in the
corpus namespace): unpatched, the **non-generic** call already produces both the correct edge and
the wrong one; patched, the generic call behaves exactly the same way. So this PR does not
introduce a new failure mode — it makes the generic form as good, and as fallible, as the
non-generic form has always been. Whether that is acceptable is a maintainer call; it is the
corpus-wide name fallback, not something this patch invents.

**5. It is now the last gap in this path.** v0.9.55 already emits `references[generic_arg]` for
these very call sites (#2911), and the code comment there states the call-site branch was
deliberately left to a separate change. #2913 and #2836 closed the declaration side. The receiver
table is filled and the type arguments are recorded; only the `calls` edge for an explicitly
generic call site is still dropped.

**Weight in a real estate**, ~15,300 C# files across seven roots, mixed .NET Framework and net8,
counted on the syntax tree (not by regex):

- qualified generic calls (A/B): **272** lost `calls` edges, counted pairwise and each pattern
  checked against the source — that figure is from the issue and unchanged.
- generic receiver (E): **23** call sites.
- unqualified (F): **2159** call sites; via `this` (G): **628**. These two shapes resolve against
  the caller's own class or a base class, so their targets are in-corpus by construction. I have
  not yet measured how many currently resolve, so I am not claiming that as a win — only that the
  shape is common, and it is the shape nothing has addressed so far.

PR #2677, opened 2026-08-12
fix(skill): clean up needs_update, not .needs_update

### LINKED ISSUE #2440: SKILL.md step 9 clears `.needs_update` but everything else uses `needs_update`, so a full rebuild never clears the stale flag
**Net:** `SKILL.md` step 9 cleans up `graphify-out/.needs_update` (leading dot), but every writer and reader of that flag uses `graphify-out/needs_update` (no dot). So a full `/graphify <path>` rebuild never clears a flag set by `--watch`, and the read hook keeps emitting the stale nudge forever afterwards. One character.

## The mismatch

Written and read without a dot:

- `graphify/watch.py:1560` — `flag = Path(watch_path) / _GRAPHIFY_OUT / "needs_update"`
- `graphify/watch.py:1569` — `flag = watch_path / _GRAPHIFY_OUT / "needs_update"`
- `graphify/watch.py:1288`, `1333`, `1534` — the `# clear stale needs_update flag if present` paths, all `out / "needs_update"`
- `graphify/cli.py:601` — `if out_path("needs_update").exists(): stale = True`

Deleted with a dot:

- `SKILL.md:612` — `rm -f graphify-out/.needs_update 2>/dev/null || true`

## Why it matters

`--watch` writes `needs_update` when it sees a doc/paper/image change it cannot re-extract without an LLM (`watch.py:1587`). The read hook guard then treats the whole graph as stale for every source read (`cli.py:599-604`), which is correct and useful.

The intended escape hatch is a full rebuild through the skill, whose step 9 is supposed to clear the flag. Because it targets `.needs_update`, the real flag survives. Every subsequent `Read` gets `_READ_NUDGE_STALE` — "graph.json exists but may be STALE for this file ... run `graphify update` to refresh" — even immediately after a complete, successful rebuild that did re-extract those docs. `graphify update` does clear it (`watch.py:1534`), so the workaround is to run the CLI path, but the skill path is the documented one and it silently fails.

Reproduction: run `--watch`, touch a `.md` file so the flag appears, stop the watcher, then run the full `/graphify .` pipeline. `graphify-out/needs_update` is still on disk and the stale nudge persists.

## Fix

```diff
-rm -f graphify-out/.needs_update 2>/dev/null || true
+rm -f graphify-out/needs_update 2>/dev/null || true
```

Found while auditing a v0.9.32 install (macOS, Claude Code, `uv tool install graphifyy`). Nothing else in the tree references the dotted name, so this looks like a typo rather than two intentional flags.

rohit-jsfreaky:
Fixes #2440.

## Summary

`--watch` writes `graphify-out/needs_update` when it sees a doc/paper/image change it cannot re-extract without an LLM, and the read hook then treats the graph as stale for every source read while that file exists. The documented escape hatch is a full rebuild through the skill, whose step 9 cleans the flag up.

Step 9 deleted `graphify-out/.needs_update` — with a leading dot, which nothing writes. The `rm` matched nothing, so the real flag survived a complete, successful rebuild and every later Read kept getting the stale nudge. `graphify update` does clear it (`watch.py`), so the CLI path worked while the documented skill path silently did not.

The dot looks accidental rather than a second flag: the line directly above cleans dotted temp files (`.graphify_detect.json`, `.graphify_extract.json`, …), and the same fragment's own prose already writes `graphify-out/needs_update` without one.

## Changes

- Three skillgen fragments (`core/core.md`, `core/aider.md`, `core/devin.md`) drop the dot. The 32 rendered artifacts follow from `python -m tools.skillgen` and `--bless`; all 35 generated-line changes are that one line, and the two Windows ones are the PowerShell form skillgen derives itself.
- `gen.py` gains `_is_needs_update_flag_fix_line`, registered in `_SANCTIONED_MONOLITH_DIFFS`. The aider/devin monoliths are diffed against the pinned pristine v8 blob and every deviation has to be an enumerated change-class, so the round-trip guard fails without it. Same shape as the existing predicates.
- The PowerShell translator test used the old dotted line as its "translates exactly" example; updated so it still mirrors a real corpus line.

## Tests

`test_skill_cleanup_targets_the_flag_the_code_actually_writes` asserts the two halves agree: `watch.py` and `cli.py` name the undotted flag, and no rendered artifact references a dotted one. Confirmed it fails with the fragments reverted:

```
AssertionError: graphify/skill-agents.md cleans up '.needs_update', which nothing ever writes
```

`uv run --no-sync pytest tests/test_skillgen.py -q` — 65 passed.

`uv run --no-sync python -m tools.skillgen --check` — 134 artifacts match committed output and `expected/`. `--audit-coverage`, `--schema-singleton`, `--monolith-roundtrip` and `--always-on-roundtrip` all OK (the git-show validators need `origin/v8` fetched locally or they skip).

`uv run --no-sync ruff check tools/skillgen/gen.py tests/test_skillgen.py` — passed.

Full suite: 22 failed, 4300 passed, 13 skipped, against a clean `v8` baseline of 22 failed, 4299 passed, 13 skipped — the same 22 in both runs, all pre-existing and platform-sensitive on this Windows machine. The +1 is the new test.

I did not run the `--watch` → touch a doc → rebuild sequence end to end; the evidence here is the filename mismatch between what the code writes and reads and what the skill deletes, established from the source.

### DIFF
--- a/graphify/skill-agents.md
+++ b/graphify/skill-agents.md
@@ -617,7 +617,7 @@ print(f'All time: {cost[\"total_input_tokens\"]:,} input, {cost[\"total_output_t
 "
 rm -f graphify-out/.graphify_detect.json graphify-out/.graphify_extract.json graphify-out/.graphify_ast.json graphify-out/.graphify_semantic.json graphify-out/.graphify_analysis.json
 find graphify-out -maxdepth 1 -name '.graphify_chunk_*.json' -delete 2>/dev/null
-rm -f graphify-out/.needs_update 2>/dev/null || true
+rm -f graphify-out/needs_update 2>/dev/null || true
 ```
 
 Replace INPUT_PATH with the actual path (same value used in Steps 4-5) so the manifest is relativized to the scan root.
--- a/graphify/skill-aider.md
+++ b/graphify/skill-aider.md
@@ -720,7 +720,7 @@ print(f'This run: {input_tok:,} input tokens, {output_tok:,} output tokens')
 print(f'All time: {cost[\"total_input_tokens\"]:,} input, {cost[\"total_output_tokens\"]:,} output ({len(cost[\"runs\"])} runs)')
 "
 rm -f .graphify_detect.json .graphify_extract.json .graphify_ast.json .graphify_semantic.json .graphify_analysis.json .graphify_labels.json; find . -maxdepth 1 -name '.graphify_chunk_*.json' -delete 2>/dev/null
-rm -f graphify-out/.needs_update 2>/dev/null || true
+rm -f graphify-out/needs_update 2>/dev/null || true
 ```
 
 Tell the user (omit the obsidian line unless --obsidian was given):
--- a/graphify/skill-amp.md
+++ b/graphify/skill-amp.md
@@ -617,7 +617,7 @@ print(f'All time: {cost[\"total_input_tokens\"]:,} input, {cost[\"total_output_t
 "
 rm -f graphify-out/.graphify_detect.json graphify-out/.graphify_extract.json graphify-out/.graphify_ast.json graphify-out/.graphify_semantic.json graphify-out/.graphify_analysis.json
 find graphify-out -maxdepth 1 -name '.graphify_chunk_*.json' -delete 2>/dev/null
-rm -f graphify-out/.needs_update 2>/dev/null || true
+rm -f graphify-out/needs_update 2>/dev/null || true
 ```
 
 Replace INPUT_PATH with the actual path (same value used in Steps 4-5) so the manifest is relativized to the scan root.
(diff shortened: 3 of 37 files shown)


rohit-jsfreaky:
Both findings assume the flag file is named `.needs_update`. Nothing writes that name: on `v8` before this PR, `.needs_update` appears only in the skill files themselves and in one test fixture quoting them — no Python file reads or writes it.

The flag is undotted on both sides: written at `watch.py:1499, 1546, 1750, 1776, 1785` (`out / "needs_update"`), read at `cli.py:692` (`out_path("needs_update").exists()`), and described at `watch.py:1827` as "writes a needs_update flag". The fragments' own prose already documented `graphify-out/needs_update` too.

So the Windows line is consistent — `graphify-out\needs_update` is the same path as `graphify-out/needs_update` — and no sentinel is left behind. The old `rm` matched nothing, which is exactly #2440.

PR #2682, opened 2026-08-12
fix(report): portable basename in GRAPH_REPORT.md header, not the host absolute path (#2628)

### LINKED ISSUE #2628: GRAPH_REPORT.md header embeds the generator's absolute host path (same class as #2598)
`GRAPH_REPORT.md`'s generated header embeds the absolute path of the machine that built it:

```
# Graph Report - /Users/<user>/dev/apps/<project>  (2026-08-03)
```

Source is `graphify/report.py:104`:

```python
f"# Graph Report - {root}  ({today})",
```

`root` is interpolated as given, so the header reflects whatever was passed on the command line: running `graphify update .` from inside the project yields `# Graph Report - .`, while passing an absolute root bakes the full host path into the file.

This is the same class as #2598, just fixed for `graph.html`'s `<title>`. `GRAPH_REPORT.md` is a tracked artifact in practice — of six repos here, four commit it, and one of those carries an absolute `/Users/...` path in its committed blob. The result is that the same graph produces different bytes on different machines, and the build machine's directory layout ends up in git history.

Suggested fix, mirroring #2598: prefer a project-relative label or the directory basename rather than interpolating `root` raw.

Notes for triage:
- #777 (closed) covered `manifest.json`, `.graphify_root` and `cache/ast/*.json`, but explicitly not `GRAPH_REPORT.md`.
- PR #778's title reads "use project-relative paths in manifest.json and GRAPH_REPORT.md", but its diff touches only `graphify/__main__.py` and `graphify/detect.py` — it never touches `report.py`, so the report header is not actually covered by it. That PR is also currently conflicting and last updated 2026-05-23.

Observed on 0.9.40.

ousamabenyounes:
## Summary

Fix #2628

`GRAPH_REPORT.md`'s header interpolated the scan `root` verbatim
(`report.py:104`), so passing an absolute root baked the build machine's full
path into a **tracked** artifact:

```
# Graph Report - /Users/<user>/dev/apps/<project>  (2026-08-03)
```

The same graph then produces different bytes on different machines and the host
directory layout ends up in git history. This is the same class of defect as
#2598 (fixed for `graph.html`'s `<title>` in 0.9.40).

The fix labels the header with the **project directory basename** via a small
`_portable_root_label()` helper, so `graphify update .`, `graphify update ./proj`,
and `graphify update /abs/path/proj` all render `proj` — portable and stable
across machines, no filesystem access for absolute inputs.

## Reproduces with

On unmodified `v8`, `report.generate(..., root="/Users/mike/dev/apps/secretproj")`
yields:

```
# Graph Report - /Users/mike/dev/apps/secretproj  (2026-08-12)
```

Two new tests in `tests/test_report.py`
(`test_report_header_does_not_embed_host_absolute_path`, `test_portable_root_label`).

RED — new tests on `v8` with the fix reverted (`report.py` only):

```
FAILED tests/test_report.py::test_report_header_does_not_embed_host_absolute_path
FAILED tests/test_report.py::test_portable_root_label - ImportError: cannot import name '_portable_root_label'
2 failed
```

GREEN — with the fix:

```
tests/test_report.py::test_report_header_does_not_embed_host_absolute_path PASSED
tests/test_report.py::test_portable_root_label PASSED
2 passed
```

Full suite (Python 3.13, local): `4330 passed, 3 skipped`; the only failures are
3 pre-existing, environment-dependent `test_ollama.py` backend-detection tests
that also fail on unmodified `v8` (local ollama present) — no new failures, no
regressions. skillgen `--check`/`--audit-coverage`/`--schema-singleton`/
`--monolith-roundtrip`/`--always-on-roundtrip` and the `graphify --help`/`install`
smoke all pass.

## Files changed

| File | Change |
|------|--------|
| `graphify/report.py` | add `_portable_root_label()`; use it in the header |
| `tests/test_report.py` | RED→GREEN tests for the header + helper |
| `CHANGELOG.md` | entry under 0.9.41 keyed to #2628 |

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@ Full release notes with details on each version: [GitHub Releases](https://githu
 - Fix: a PHP `use` import written with a leading-backslash / fully-qualified prefix now resolves to its target definition instead of being dropped (#2661, thanks @ousamabenyounes).
 - Fix: an unresolved local JS/TS import (to a file absent from the scan) now emits a stable, portable `ref` target id instead of leaking a per-checkout absolute-path slug (#2457, thanks @rohit-jsfreaky).
 - Fix: `graphify benchmark` no longer crashes on a node whose label is `None` (#2674, thanks @Arthuro0103).
+- Fix: `GRAPH_REPORT.md`'s header no longer embeds the generator's absolute host path (#2628, thanks @AirRocker); it labels the report with the project directory basename, so the same graph produces the same bytes on any machine (same class as the 0.9.40 `graph.html` title fix, #2598).
 
 ## 0.9.40 (2026-08-11)
 
--- a/graphify/report.py
+++ b/graphify/report.py
@@ -2,7 +2,32 @@
 from __future__ import annotations
 import re
 from datetime import date
+from pathlib import Path
 import networkx as nx
+
+
+def _portable_root_label(root: str) -> str:
+    """Portable label for the report header — the project directory basename.
+
+    GRAPH_REPORT.md is a tracked artifact in practice, so its header must not
+    bake the generator host's absolute path into the file: the same graph would
+    otherwise produce different bytes on different machines and leak the build
+    machine's directory layout into git history (#2628, same class as #2598).
+
+    Taking the basename strips any leading absolute path without touching the
+    filesystem, and makes `graphify update .`, `graphify update ./proj`, and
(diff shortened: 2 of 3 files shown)


safishamsi:
Shipped in v0.9.42 (`graphifyy==0.9.42` on PyPI). Portable basename in the GRAPH_REPORT header. Credited in the release notes. Thanks @ousamabenyounes!

PR #2683, opened 2026-08-12
fix(cache): surface corrupt semantic cache entries instead of swallowing them (#2405)

### LINKED ISSUE #2405: load_cached() silently swallows corrupt cache entries (55% of a real-world cache was dead, no diagnostics)
While auditing a knowledge graph built from a personal Markdown vault (~230 files, Windows 11), I found that 138 of 249 entries in `graphify-out/cache/` were invalid JSON. Every incremental run had been re-extracting more than half the corpus, paying that LLM work again each time, and nothing in the output suggested it.

The cause is in `graphify/cache.py`:

```python
try:
    result = json.loads(entry.read_text(encoding="utf-8"))
except (json.JSONDecodeError, OSError):
    return None
```

`load_cached()` treats a corrupt entry exactly like a missing one, so the corruption never surfaces and the file stays in place, failing again on every future run. I checked the current `v8` HEAD and the behavior is still there.

How my entries got corrupted is a separate story: they came from an old install (graphifyy 0.3.4) that wrote Windows paths with unescaped backslashes into `source_file` (`"C:\Users\..."`). That producer side looks covered by #2380 and #2197. This issue is only about the consumer side: whatever the origin, corrupt entries deserve a signal.

Repro on HEAD: drop any invalid-JSON file into `graphify-out/cache/` and run an incremental update. The entry is re-extracted, nothing is logged, and the file remains for the next run.

Two small changes would cover it:

1. Count decode failures in `load_cached()` / `check_semantic_cache()` and print one warning per run, e.g. `warning: 138 corrupt cache entries ignored (will re-extract)`.
2. Optionally rename bad entries to `.corrupt` so they stop being re-parsed every run.

Happy to send a PR for either if that helps.

Environment: Windows 11, Python 3.14; corruption produced under graphifyy 0.3.4; silent behavior verified in `v8` HEAD `graphify/cache.py`.

ousamabenyounes:
## Summary

Fix #2405

`load_cached()` caught `json.JSONDecodeError` in the same `except` as `OSError` and returned `None`, so a corrupt (invalid-JSON) cache entry was indistinguishable from a cache miss. A corrupt entry stays on disk and fails to parse on **every** future run, so the file is silently re-extracted each time — and for semantic kinds that means re-billing the LLM — with nothing in the output hinting why. The reporter found 138 of 249 entries dead in a real vault, paying that extraction cost on every incremental run.

## Fix

- `load_cached()` now counts `JSONDecodeError` failures in a process-global `_corrupt_cache_entries`, mirroring the existing `_legacy_semantic_hits` counter, and still returns `None` (miss → re-extract, self-healing).
- `check_semantic_cache()` emits **one aggregate** `RuntimeWarning` per run naming how many entries could not be parsed and how to clear them (`--force` / clear cache), matching the existing legacy-vintage warning it already raises.
- `OSError` stays a silent miss — a transient read failure is not corruption.

Scope is the consumer side only, as the issue requests; producers that write bad entries are tracked separately (#2380 / #2197).

New test `test_corrupt_semantic_entry_warns_and_is_a_miss` saves a semantic entry, corrupts it on disk, then asserts `check_semantic_cache` both treats it as a miss and warns.

RED (unmodified `v8`, test only):

```
tests/test_cache.py:1483: in test_corrupt_semantic_entry_warns_and_is_a_miss
    with pytest.warns(RuntimeWarning, match="corrupt"):
E   Failed: DID NOT WARN. No warnings of type (<class 'RuntimeWarning'>,) were emitted.
1 failed
```

GREEN (with fix):

```
1 passed, 1 warning in 0.26s
```

## No regression

- `uv run --frozen pytest tests/ -q` → `4329 passed, 3 skipped`, plus the 3 pre-existing `test_ollama.py` backend-detection failures that also fail on a clean `v8` checkout (unrelated to this change).
- `python -m tools.skillgen --check` → `check OK`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@ Full release notes with details on each version: [GitHub Releases](https://githu
 - Fix: a PHP `use` import written with a leading-backslash / fully-qualified prefix now resolves to its target definition instead of being dropped (#2661, thanks @ousamabenyounes).
 - Fix: an unresolved local JS/TS import (to a file absent from the scan) now emits a stable, portable `ref` target id instead of leaking a per-checkout absolute-path slug (#2457, thanks @rohit-jsfreaky).
 - Fix: `graphify benchmark` no longer crashes on a node whose label is `None` (#2674, thanks @Arthuro0103).
+- Fix: a corrupt (invalid-JSON) semantic cache entry is no longer swallowed as a silent miss; `check_semantic_cache` now emits one aggregate warning naming how many entries failed to parse, so a cache that fails on every run (silently re-billing the extraction) is diagnosable instead of invisible (#2405, thanks @jorgefusterr).
 
 ## 0.9.40 (2026-08-11)
 
--- a/graphify/cache.py
+++ b/graphify/cache.py
@@ -80,6 +80,12 @@ _PROMPT_FP_LEN = 12
 # check_semantic_cache can report N to the user (#1939).
 _legacy_semantic_hits = 0
 
+# Count of cache entries that failed to parse as JSON this process. A corrupt
+# entry is not a miss: left in place it fails on every future run, silently
+# re-extracting (and, for semantic kinds, re-billing) the file forever. The
+# counter lets check_semantic_cache surface one aggregate warning (#2405).
+_corrupt_cache_entries = 0
+
 # Prompt-file fingerprints already computed, keyed by (path, size, mtime_ns) —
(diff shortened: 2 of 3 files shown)


safishamsi:
Shipped in v0.9.42 (`graphifyy==0.9.42` on PyPI). Corrupt semantic-cache entries are surfaced and re-extracted. Credited in the release notes. Thanks @ousamabenyounes!

PR #2684, opened 2026-08-12
fix(extract): attribute cross-file INFERRED `uses` edges to the referencing symbol (#2652)

### LINKED ISSUE #2652: Python: INFERRED `uses` edges anchor on the import line and bind to a co-located class instead of the importing function
**Version:** reproduced on 0.9.40 (and 0.9.17)

The AST extractor emits `INFERRED` `uses` edges that appear to encode "class `C` is defined in module `M`, `M` imports `X`, therefore `C` uses `X`". The endpoint is wrong whenever the importing module also defines a class: the *function* is the real user, not the class that happens to share the file. The edge is anchored at the import line, so `source_location` looks like corroboration.

### Repro

`helpers.py`
```python
class Helper:
    pass
```

`api.py`
```python
from helpers import Helper


class Request:
    x: int = 0


def handler(req):
    return Helper()
```

```python
from pathlib import Path
r = extract([Path('api.py'), Path('helpers.py')], cache_root=Path('.'))
for e in r['edges']:
    if e.get('confidence') == 'INFERRED':
        print(e['source'], e['relation'], e['target'], e['source_file'], e['source_location'])
```

**Actual**
```
api_request --uses--> helpers_helper  [INFERRED]  api.py:L1
```

**Expected** — either `api_handler --uses--> helpers_helper`, or no inferred edge at all. `Request` never references `Helper`.

### Why it matters

This is systematic rather than occasional in any framework codebase that co-locates request/response models with handlers. In a FastAPI + Pydantic repo (966 nodes) **all 19 `uses` edges in the graph were false** — not one was a genuine class-to-class relationship. Representative sample:

```
FieldGenerateRequest --uses--> Toggles          | from backend.prompts.router import PromptRouter, Toggles
LLMResult            --uses--> MockLLMClient    | from backend.llm.mock_client import MockLLMClient
LLMClient            --uses--> GeminiLLMClient  | from backend.llm.gemini_client import GeminiLLMClient
```

The first is fully spurious — that request model never touches `Toggles`. The others invert a real relationship: the clients return `LLMResult` and implement `LLMClient`, not the reverse.

These edges reach users: they surface in `GRAPH_REPORT.md` under "Surprising Connections", they are traversed by `query`, and the report's own "Are the N inferred relationships involving `X` actually correct?" prompt sends people to a `source_location` that shows only an import statement.

### Workaround

Dropping AST-inferred edges whose `source_location` resolves to an `import` / `from … import` statement removed all 19 false positives here and kept all 15 genuine inferred edges (`isinstance` assertions, `monkeypatch.setattr`, constructor calls), so the import-line anchor cleanly separates the two classes.

### Checked for duplicates

#2241 (call arguments binding to a single-letter test helper), #991 / #726 / #630 (cross-language identifier collapse in semantic chunks) are different mechanisms; #1633 is a feature request. This is structural extraction, single-language, and reproduces with no LLM backend involved.

ousamabenyounes:
## Summary

Fix #2652.

`_resolve_cross_file_imports` (the Python cross-file import→`uses` resolver) drew an INFERRED `uses` edge from **every** class in an importing file to each imported entity — with no check that the class actually referenced it. The edge was anchored at the import line, so `source_location` showed only a `from … import …` statement that read as corroboration.

Consequences, all reported in the issue:
- A co-located class that never touches the import gets a false edge; the function that actually uses it gets none. In a real FastAPI+Pydantic repo (966 nodes) **all 19** `uses` edges were false.
- These edges surface in `GRAPH_REPORT.md` "Surprising Connections", are traversed by `query`, and send users to a source location showing only an import.

### Fix

Attribute the `uses` edge to the top-level symbol (class **or** module-level function) whose body actually references the imported name, anchored at the real reference line:

- A **module-level function** is now a valid source (`api_handler --uses--> helpers_helper`), not just classes.
- A class stays a **unit** — a reference inside one of its methods counts for the class, preserving the documented `DigestAuth --uses--> Response` edge.
- Import **aliases** (`from x import Helper as H`) are honoured via the local name, so a body that only names `H` still resolves.

Scope is confined to Pass 2 of this one resolver; the module-resolution logic (relative/absolute/aliased) is preserved. A dead `stem_to_path` local was removed.

## Reproduces with (unmodified `v8`)

```
api_request --uses--> helpers_helper  [INFERRED]  api.py:L1
```
`Request` never references `Helper`; `handler()` does.

New tests in `tests/test_extract.py` drive the public `extract()` on the issue's two-file corpus.

RED — new tests on `v8` with the production fix reverted:
```
FAILED tests/test_extract.py::test_inferred_uses_edge_attributes_to_the_referencing_symbol
  AssertionError: assert ('api_handler', 'helpers_helper') in {('api_request', 'helpers_helper')}
```

GREEN — with the fix:
```
3 passed  (attributes_to_referencing_symbol, kept_when_class_body_references_the_import, follows_an_import_alias)
```

## Full local suite

- `uv run --frozen pytest tests/ -q` → 4330 passed, 3 skipped. The only failures (3 × `test_ollama.py::test_detect_backend_*`) are pre-existing on `v8` (a locally-running Ollama) and unrelated to this change.
- `python -m tools.skillgen --check` / `--schema-singleton` OK; `ruff check` clean.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@ Full release notes with details on each version: [GitHub Releases](https://githu
 - Fix: a PHP `use` import written with a leading-backslash / fully-qualified prefix now resolves to its target definition instead of being dropped (#2661, thanks @ousamabenyounes).
 - Fix: an unresolved local JS/TS import (to a file absent from the scan) now emits a stable, portable `ref` target id instead of leaking a per-checkout absolute-path slug (#2457, thanks @rohit-jsfreaky).
 - Fix: `graphify benchmark` no longer crashes on a node whose label is `None` (#2674, thanks @Arthuro0103).
+- Fix: a cross-file INFERRED `uses` edge is now attributed to the symbol that actually references the import — a module-level function is a valid source, and a co-located class that never touches the import gets no edge (#2652, thanks @larya-dot-eu); the edge is anchored at the real reference line instead of the import statement.
 
 ## 0.9.40 (2026-08-11)
 
--- a/graphify/extractors/resolution.py
+++ b/graphify/extractors/resolution.py
@@ -1935,25 +1935,31 @@ def _resolve_cross_file_imports(
                 if src_path.stem not in bare_to_qualified:
                     bare_to_qualified[src_path.stem] = fq_stem
 
-    # Pass 2: for each file, find `from .X import A, B, C` and resolve
+    # Pass 2: for each file, find `from .X import A, B, C`, then attribute the
+    # `uses` edge to the specific local symbol (class OR function) whose body
+    # actually references the imported name — not to every class that merely
+    # shares the file (#2652). The edge is anchored at the real reference, not
+    # the import line, so `source_location` points at genuine corroboration.
     new_edges: list[dict] = []
-    stem_to_path: dict[str, Path] = {_file_stem(p): p for p in paths}
 
     for file_result, path in zip(per_file, paths):
-        stem = _file_stem(path)
         str_path = str(path)
 
-        # Find all classes defined in this file (the importers).
-        # Excludes rationale nodes whose labels happen not to end in ")" or ".py"
-        # but which must never be treated as importing entities (#563).
-        local_classes = [
-            n["id"] for n in file_result.get("nodes", [])
-            if n.get("source_file") == str_path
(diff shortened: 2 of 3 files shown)


safishamsi:

PR #2685, opened 2026-08-12
fix(js): shadow for...of / for...in loop bindings in indirect_call resolution (#2606)

### LINKED ISSUE #2606: JavaScript resolver links block-scoped object-shorthand variables to unrelated functions
## Behavior

Graphify emits an `indirect_call` edge from a function that uses a block-scoped loop variable in an object-shorthand argument to an unrelated top-level function with the same name in another file.

Observed with `graphify 0.9.39`.

## Minimal reproduction

`declaration.mjs`:

```js
export function entry(signature) {
  return { signature };
}

export const fixture = entry("public.entry()");
```

`consumer.mjs`:

```js
export function findNamed(entries, lookup) {
  const resolved = [];
  for (const entry of entries) {
    if (lookup(entry.name)) resolved.push({ entry });
  }

  return resolved;
}
```

Run:

```sh
graphify extract ./repro --code-only --force
```

The resulting graph includes:

```text
findNamed() --indirect_call [INFERRED]--> entry()
source: consumer.mjs:4
```

## Expected

No edge from `findNamed()` to the unrelated `entry()` declaration. The `entry` at the reported source location is a `for...of` lexical binding used as an object-shorthand property, not a callable identifier.

## Impact

A generic fixture/helper name can become a false high-betweenness hub when this pattern appears across many files, distorting god-node and community analysis.

## Suggested regression test

Build the two-file corpus above and assert there is no `indirect_call` edge from `consumer_findnamed` to `declaration_entry`. A positive control should retain the real `fixture → entry()` call.

ousamabenyounes:
## Summary

Fix #2606

A JS/TS `for...of` / `for...in` loop binding used as a *value* fabricates a phantom `indirect_call` edge to an unrelated same-named module function.

`_js_local_bound_names` collected parameters and `variable_declarator` targets, but the loop variable of `for (const entry of xs)` is the statement's `left` pattern — **not** wrapped in a `variable_declarator` — so `entry` contributed nothing to the per-function shadow set. Referenced as a value (object-shorthand `push({ entry })`), it read as an unresolved by-name reference, resolved against the corpus-wide label index, and minted an INFERRED `indirect_call` edge (0.8) to an unrelated top-level `entry()` in another file. A generic fixture/helper name then becomes a false high-betweenness hub, distorting god-node and community analysis.

This is the same declared/undeclared trap already fixed for single-arrow parameters (#2568-era) and catch bindings — the loop binding is now collected alongside parameters and `const`/`let`/`var` declarators. Both bare (`for (const entry of xs)`) and destructured (`for (const { entry } of xs)`) forms are handled; C-style `for (let i = 0; ...)` uses a `lexical_declaration` with real declarators and was always covered.

## Reproduces with

`declaration.mjs` defines `entry()`, `consumer.mjs` loops `for (const entry of entries)` and does `resolved.push({ entry })`. On `v8` before the fix:

```
indirect_call consumer_findnamed -> declaration_entry   (source consumer.mjs:4)
```

Expected: no edge — the `entry` is a `for...of` binding, not the imported/module function.

New `tests/test_indirect_call_for_of_binding_shadow.py`.

RED (prod fix reverted, tests intact):
```
FAILED test_for_of_binding_does_not_fabricate_indirect_call
FAILED test_for_of_destructuring_binding_shadows
2 failed, 1 passed
```
The passing one is the positive control (`test_genuine_reference_outside_the_loop_still_emits`) — a same-named callable referenced outside the loop still resolves, so the shadow-set widening does not blanket-suppress real edges.

GREEN (with fix):
```
3 passed
```

## Full suite

`uv run --frozen pytest tests/ -q` → `4331 passed, 3 skipped`. The only failures are 3 pre-existing `tests/test_ollama.py` backend-detection tests that fail identically on the unmodified `v8` base (a local Ollama is reachable on the test host); they are unrelated to this change.

## Files changed

| File | Change |
|------|--------|
| `graphify/extractors/engine.py` | collect `for_in_statement.left` binding in `_js_local_bound_names` |
| `tests/test_indirect_call_for_of_binding_shadow.py` | new regression tests + positive control |

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@ Full release notes with details on each version: [GitHub Releases](https://githu
 - Fix: a PHP `use` import written with a leading-backslash / fully-qualified prefix now resolves to its target definition instead of being dropped (#2661, thanks @ousamabenyounes).
 - Fix: an unresolved local JS/TS import (to a file absent from the scan) now emits a stable, portable `ref` target id instead of leaking a per-checkout absolute-path slug (#2457, thanks @rohit-jsfreaky).
 - Fix: `graphify benchmark` no longer crashes on a node whose label is `None` (#2674, thanks @Arthuro0103).
+- Fix: a JS/TS `for...of` / `for...in` loop binding (`for (const entry of xs)`) used as a value — e.g. in an object-shorthand argument `push({ entry })` — no longer fabricates an `indirect_call` edge to an unrelated same-named module callable (#2606, thanks @ousamabenyounes); the loop binding is now shadowed like parameters and `const`/`let` declarators, completing the arrow-parameter and catch-binding shadow fixes.
 
 ## 0.9.40 (2026-08-11)
 
--- a/graphify/extractors/engine.py
+++ b/graphify/extractors/engine.py
@@ -1246,6 +1246,16 @@ def _js_local_bound_names(func_node, source: bytes) -> set[str]:
                 name = c.child_by_field_name("name")
                 if name is not None:
                     _js_collect_pattern_idents(name, source, bound)
+            elif c.type == "for_in_statement":
+                # `for (const entry of xs)` / `for (const {k} of xs)`: the loop
+                # binding is the `left` pattern, NOT wrapped in a
+                # variable_declarator, so the branch above misses it and `entry`
+                # read as a by-name reference to any same-named module callable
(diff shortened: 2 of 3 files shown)


safishamsi:
Shipped in v0.9.42 (`graphifyy==0.9.42` on PyPI). for-of/for-in loop bindings are shadowed (completes the #2568/#2569/#2517 family). Credited in the release notes. Thanks @ousamabenyounes!

PR #2686, opened 2026-08-12
fix(extract): distinguish a broken .sql grammar from a missing one (#2602)

### LINKED ISSUE #2602: .sql files skipped as "tree_sitter_sql not installed" while the module loads and parses fine
`.sql` files are skipped with "tree_sitter_sql not installed" while the module is installed and parses fine — and the suggested fix is a no-op.

## What happens

Indexing a repo that contains `.sql` files prints:

```
warning: 32 .sql file(s) contributed nothing to the graph because a dependency is
missing: tree_sitter_sql not installed. Install it with: pip install "graphifyy[sql]" (#1745)
```

The suggested command is a no-op — the dependency is already there:

```
Requirement already satisfied: tree-sitter-sql in .../site-packages (from graphifyy[sql]) (0.3.11)
```

## The module is not just importable, it works

Run with the same interpreter graphify uses:

```python
import tree_sitter, tree_sitter_sql
lang = tree_sitter.Language(tree_sitter_sql.language())
tree = tree_sitter.Parser(lang).parse(b'SELECT 1;')
print(tree.root_node.type)   # -> program
```

The language loads and parses SQL. The warning still fires on the next run.

## Not a subprocess/environment issue either

My first hypothesis was that the AST workers don't inherit the environment, since
extraction runs with `[8 workers]`. Ruled out — the warning is identical with:

```
graphify <path> --update --max-workers 1 --max-concurrency 1
```

## Why the message matters beyond the missing files

The wording sends you to install something that is already installed, so the natural
reaction is "done, fixed" and moving on. In my case the repo is a financial system whose
core is its SQL migrations (row-level security, `SECURITY DEFINER` functions, CHECK
constraints). The resulting graph describes the application and silently omits the layer
where the security lives — and the consumers of that graph are agents, which have no way
to know half the system is missing.

A message like "SQL grammar found but failed to load: `<actual exception>`" would have
saved the whole investigation.

## Environment

- `graphifyy` 0.8.39
- `tree-sitter-sql` 0.3.11
- Python 3.14 (Windows)
- Backend: ollama (local)

Happy to run any diagnostic that would help pin down where the detection fails — I can
reproduce it on demand.

ousamabenyounes:
## Summary

Fix #2602

`.sql` files were being skipped with `tree_sitter_sql not installed. Run: pip install "graphifyy[sql]"` even when the grammar **was** installed — the suggested command is a no-op and the whole SQL layer silently vanished from the graph.

**Root cause:** an installed-but-broken grammar (e.g. a `tree-sitter-sql` wheel built for a different Python ABI — the reporter is on Python 3.14) raises `ImportError` at *import* time exactly like a genuinely-absent module. `extract_sql`'s single `except ImportError` mapped both to `"not installed"`, sending the user to a no-op `pip install` and hiding the real failure.

**Fix:**
- `extract_sql` now uses `importlib.util.find_spec` to tell an absent module from one that failed to load, and returns the **actual** load exception (`tree_sitter_sql is installed but failed to load: <exc>`) in the latter case — the message the reporter asked for.
- The `#1745` aggregation warning in `extract()` is widened to surface these load failures too (they carry no `not installed` marker, so they would otherwise be dropped silently by #1666/#1689/#1745), and it omits the misleading "install the extra" hint for a present-but-broken grammar.

No behaviour change for a genuinely-absent grammar: it still reports `not installed` with the `graphifyy[sql]` install hint.

Deterministically reproduced on Linux / Python 3.13 by forcing `import tree_sitter_sql` to raise a load-time `ImportError` while the module remains installed (`find_spec` sees it).

RED (on the unmodified `v8`, new tests applied):
```
FAILED tests/test_extract.py::test_extract_sql_reports_load_failure_not_missing
FAILED tests/test_extract.py::test_extract_warns_sql_grammar_failed_to_load
AssertionError: assert 'failed to load' in 'tree_sitter_sql not installed. Run: pip install tree-sitter-sql'
```

GREEN (with the fix):
```
5 passed, 168 deselected   # 2 new + the 3 existing #1745/sql tests
```

Full local CI replay (`run-ci.sh`, mirrors `.github/workflows/ci.yml`): skillgen checks OK, `4330 passed, 3 skipped`. The only failures are 3 pre-existing `test_ollama` backend-detection tests that also fail on an unmodified `v8` checkout (a local ollama daemon skews detection) and are unrelated to this diff.

## Files changed

| File | Change |
|------|--------|
| `graphify/extractors/sql.py` | split the `ImportError` branch: absent vs installed-but-failed-to-load |
| `graphify/extract.py` | `#1745` warning surfaces load failures with the real cause, no reinstall hint |
| `tests/test_extract.py` | 2 new regression tests (unit + aggregated warning) |

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@ Full release notes with details on each version: [GitHub Releases](https://githu
 - Fix: a PHP `use` import written with a leading-backslash / fully-qualified prefix now resolves to its target definition instead of being dropped (#2661, thanks @ousamabenyounes).
 - Fix: an unresolved local JS/TS import (to a file absent from the scan) now emits a stable, portable `ref` target id instead of leaking a per-checkout absolute-path slug (#2457, thanks @rohit-jsfreaky).
 - Fix: `graphify benchmark` no longer crashes on a node whose label is `None` (#2674, thanks @Arthuro0103).
+- Fix: a `.sql` file whose `tree-sitter-sql` grammar is installed but fails to load (e.g. a wheel built for a different Python ABI) no longer reports the misleading "tree_sitter_sql not installed. pip install …" no-op; the extractor distinguishes a genuinely-absent module from a broken one, and the `#1745` warning surfaces the real load exception instead of suggesting a reinstall (#2602, thanks @ousamabenyounes).
 
 ## 0.9.40 (2026-08-11)
 
--- a/graphify/extract.py
+++ b/graphify/extract.py
@@ -4864,6 +4864,13 @@ _EXTRA_FOR_EXTENSION = {
     ".dme": "dm",
 }
 
+# Substrings an extractor's error carries to classify why a dependency-backed
+# file contributed nothing, used by the #1745 warning in extract(). A grammar
+# that is present but fails to load (#2602) must not be reported as missing —
+# the "install the extra" hint would be a no-op.
+_DEP_MISSING_MARKER = "not installed"
+_DEP_LOAD_FAILED_MARKER = "failed to load"
+
 
 # Extensionless executables (CLI entry points like `devctl` or `manage`) carry
 # their language in the shebang, not the suffix. detect.classify_file already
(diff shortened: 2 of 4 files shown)


safishamsi:
Shipped in v0.9.42 (`graphifyy==0.9.42` on PyPI) (#2602). Distinguishes a broken tree-sitter-sql grammar from a missing one. Credited in the release notes. Thanks @ousamabenyounes!

Issue #2262, opened 2026-07-28
0.9.29: TypeScript leaks absolute-path slug ids into links source endpoints (C# fixed by #2250, TS not)


vanjos:
Follow-up to #2243. Thanks for the 0.9.29 fix (#2250), it cleaned up the C# case. TypeScript still leaks, in a slightly different spot.

## What happens

In a TypeScript graph, a large number of `links` entries have a `source` that is an absolute-scan-path-derived slug rather than a real node id. The slug is the scan path lowercased with every run of non-alphanumerics collapsed to `_`, followed by the file's repo-relative path and the function name, for example `abs_scan_path_slug_packages_account_src_components_paymentmethods_prepaybalancecontainer_invoicebalancesubsection`. The `target` of the same edge is a clean scope-relative id (`packages_account_src_settings_basesettingssection_constructrowwithid`).

Two things about the leaked source:

- It matches no node. The caller function it names is absent from `nodes` entirely, so this is not only the id-relativization gap from #2243. The extractor is emitting a `calls` edge for a caller it never materialized as a node, and identifying that caller by absolute path.
- It is always on the source side of `calls` edges (the caller), not the target.

## C# vs TypeScript in the same version

Same repo, same graphify 0.9.29, two scopes:

- C# scope, 79,336 nodes: 0 absolute-path slug endpoints.
- TypeScript scope, 74,271 nodes / 165,872 edges: 7,143 absolute-path slug endpoints.

So the 0.9.29 relativization landed for the C# extractor but not the TypeScript one.

## Why it matters

The slug is derived from the absolute scan path, so the same source tree built on two machines (or two checkout paths) produces different `graph.json` files. We commit graphs to git, so every rebuild on a different path shows a spurious diff, and the underlying non-determinism is general. Same reproducibility and cross-machine caching problem as #2243.

## Repro

It shows up in a TypeScript monorepo (a `packages/*` layout with cross-package imports). A minimal two-file TS project does not trigger it, and neither does a class with arrow-property handlers calling across two files, so it correlates with the larger structure rather than a single syntactic form. The leaked callers are handler-style functions (`invoicebalancesubsection`, `handletabclick`, `handleapply`) whose containing symbol does not appear in `nodes`. Happy to share a fuller extract from the affected graph if useful.

## Suggested fix

Apply the same relativization as #2243/#1899 to the TypeScript extractor's edge-source identifiers, and/or materialize the caller node so the edge references a real scope-relative id instead of an absolute path.

[...]

PR #2688, opened 2026-08-12
fix(python): resolve relative subpackage import to package __init__ (#2455)

### LINKED ISSUE #2455: Relative import of a subpackage emits an abs-path slug imports_from target to a non-existent <pkg>.py
A relative import of a sibling subpackage emits an `imports_from` edge whose target is an absolute-path-derived slug for a non-existent `<subpackage>.py`, instead of resolving to the subpackage's `__init__.py`.

## Repro (minimal, deterministic on 0.9.32)

```
proj/pyproject.toml
proj/src/mypkg/__init__.py                 (empty)
proj/src/mypkg/api/__init__.py             (empty)
proj/src/mypkg/api/routes/__init__.py      (empty)
proj/src/mypkg/graphs/__init__.py          -> def build_graph(): return {}
proj/src/mypkg/api/routes/health.py        -> from ...graphs import build_graph
```

`graphify update proj --no-cluster` produces two edges out of `health`:

```
src_mypkg_api_routes_health --imports_from--> <abs-scan-slug>_proj_src_mypkg_graphs_py   # BUG: graphs.py does not exist
src_mypkg_api_routes_health --imports-->      src_mypkg_graphs_init_build_graph           # correct
```

`graphs/` is a package (`graphs/__init__.py`), so the `imports_from` target should be `graphs/__init__.py`, which is exactly what the companion `imports` edge already resolves to. Instead it is a slug for a `graphs.py` that is not on disk.

## Why it matters

The endpoint is derived from the absolute scan path, so it is machine-specific and differs per checkout path. Committed graphs then show a spurious diff on every rebuild from a different location, the same reproducibility problem as #2262 and #2273. In a real repo this fails a fail-closed host-path check.

## Notes

- Specific to the relative form. An absolute import of the same subpackage (`from mypkg import graphs`) resolves correctly to the package `__init__`, no slug.
- Seen on a real Python src-layout package where several `from ...graphs import X` / `from ...services import X` lines each produced one of these.

## Suggested fix

Resolve a relative subpackage import to the subpackage's `__init__.py` (match what the `imports` edge does), or canonicalize/drop the node-less endpoint rather than leaving the absolute-derived slug.

ousamabenyounes:
## Summary

Fix #2455

A Python relative import of a sibling **subpackage** (`from ...graphs import build_graph`, where `graphs/` is a package — a directory with `__init__.py`, not a `graphs.py` module) emitted an `imports_from` edge whose target was an absolute-scan-path-derived slug for a nonexistent `graphs.py`, instead of resolving to the package's `__init__.py` file node.

Root cause: `_import_python` built the relative target as `base/<module>.py` unconditionally. For a subpackage that file does not exist, so the existence-gated `target_file` stamp never fired, and the `#2169` remap had nothing to canonicalize — the edge kept a per-checkout absolute-path slug (`<scan>_src_mypkg_graphs_py`) that matches no node and dangles. This is the same reproducibility class as #2262/#2273/#2457.

Fix: probe the candidate on disk with `_probe_python_module_candidate` — the exact resolver the companion `imports` edge already uses via `_resolve_python_module_path` — so `graphs` resolves to `graphs/__init__.py`. The `target_file` stamp then fires and the edge canonicalizes to the real package file node. A genuinely nonexistent module still falls back to the prior dangling behaviour, unchanged.

## Reproduces with

```
src/mypkg/__init__.py
src/mypkg/api/__init__.py
src/mypkg/api/routes/__init__.py
src/mypkg/graphs/__init__.py            -> def build_graph(): return {}
src/mypkg/api/routes/health.py          -> from ...graphs import build_graph
```

Before (on `v8`):
```
src_mypkg_api_routes_health --imports_from--> <scan>_src_mypkg_graphs_py   # BUG: graphs.py does not exist
src_mypkg_api_routes_health --imports-->      src_mypkg_graphs_init_build_graph
```
After:
```
src_mypkg_api_routes_health --imports_from--> src_mypkg_graphs_init
src_mypkg_api_routes_health --imports-->      src_mypkg_graphs_init_build_graph
```

New test `test_relative_subpackage_import_from_targets_package_init` in `tests/test_python_import_resolution.py`.

RED — unmodified `v8` base, new test applied, no prod fix:
```
F
E   AssertionError: assert False
     +  where False = _has_edge(..., 'src_mypkg_api_routes_health', 'src_mypkg_graphs_init', 'imports_from')
tests/test_python_import_resolution.py: AssertionError
1 failed
```

GREEN — with the fix:
```
tests/test_python_import_resolution.py::test_relative_subpackage_import_from_targets_package_init
1 passed
```

## Full local suite

`uv run --frozen pytest tests/ -q` → `4328 passed, 3 skipped` (+1 new). The 4 failing tests (`test_ollama.py::*`, one `test_labeling` ordering artifact) are pre-existing and reproduce identically on the unmodified `v8` base — they are backend-detection env leakage (`GEMINI_API_KEY`), unrelated to this change (which touches only `_import_python`). skillgen validators (`--check`, `--monolith-roundtrip`, `--always-on-roundtrip`) and `graphify --help` smoke all green.

## Files changed

| File | Change |
|------|--------|
| `graphify/extract.py` | `_import_python`: probe the relative-import candidate on disk so a subpackage resolves to its `__init__.py` |
| `tests/test_python_import_resolution.py` | regression test for subpackage `imports_from` resolution |

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@ Full release notes with details on each version: [GitHub Releases](https://githu
 - Fix: `query` no longer prints the truncation banner when no nodes were actually cut (only trailing edges overflowed the budget) (#2601, thanks @ousamabenyounes); a genuine node truncation still warns.
 - Fix: a PHP `use` import written with a leading-backslash / fully-qualified prefix now resolves to its target definition instead of being dropped (#2661, thanks @ousamabenyounes).
 - Fix: an unresolved local JS/TS import (to a file absent from the scan) now emits a stable, portable `ref` target id instead of leaking a per-checkout absolute-path slug (#2457, thanks @rohit-jsfreaky).
+- Fix: a Python relative import of a sibling subpackage (`from ...graphs import x`, where `graphs/` is a package) now resolves its `imports_from` edge to the package's `__init__.py` file node, matching the companion `imports` edge, instead of leaking a per-checkout absolute-path slug for a nonexistent `graphs.py` (#2455, thanks @vanjos).
 - Fix: `graphify benchmark` no longer crashes on a node whose label is `None` (#2674, thanks @Arthuro0103).
 
 ## 0.9.40 (2026-08-11)
--- a/graphify/extract.py
+++ b/graphify/extract.py
@@ -124,6 +124,7 @@ from graphify.extractors.resolution import (  # noqa: E402,F401
     _resolve_js_import_target,
     _resolve_js_module_path,
     _resolve_lua_import_target,
+    _probe_python_module_candidate,
     _resolve_python_module_path,
     _resolve_tsconfig_alias,
     _resolve_workspace_import,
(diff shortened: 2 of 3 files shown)


safishamsi:
Shipped in v0.9.42 (`graphifyy==0.9.42` on PyPI). Relative subpackage imports resolve to the package `__init__`. Credited in the release notes. Thanks @ousamabenyounes!

Issue #2534, opened 2026-08-07
Four silent failures with success exit codes: cluster-only ignores --backend, label prompt collides with the `Community {cid}` sentinel, `tree --root <abs>` flattens the hierarchy, built_at_commit stamped from cwd


elecnix:
Four independent bugs, filed together because they share one shape: **the command exits 0, prints a success line, and produces output that looks right but isn't.** There is no error for a user — or an agent — to react to.

All line references are against `v8` as of today, so these are not stale-version reports.

---

## 1. `cluster-only --backend X` accepts labeling flags and never calls the LLM

`cli.py:1621` sets `force_relabel = cmd == "label"`, and `cli.py:1776` gates the whole labeling block on it:

```python
force_relabel = cmd == "label"
...
if labels_path.exists() and not force_relabel:
    # reuse saved labels / hub-rename; --backend, --model, --batch-size unused
```

So once `.graphify_labels.json` exists, `cluster-only --backend openai --model … --batch-size N` silently discards every one of those flags. It prints `Done - N communities. GRAPH_REPORT.md, graph.json and graph.html updated.`, and `GRAPH_REPORT.md` records `Token cost: 0 input · 0 output`.

There *is* a warning at `cli.py:1830`, but it only fires `if changed:` — i.e. only when the community set drifted. When the count matches, the no-op is completely silent.

The flags being *accepted* is what sells it: a rejected flag would have been obvious immediately. It took two labeling runs and a look at the token counter to notice nothing had happened.

**Suggested fix:** if `--backend` / `--model` / `--batch-size` are passed explicitly to `cluster-only` while the reuse branch is taken, either honour them or warn naming the ignored flags. Silently accepting a flag that does nothing is the worst of the three options.

## 2. The label prompt collides with the placeholder sentinel, so some models' answers get discarded

`llm.py:2859` formats each prompt line using the same string that `llm.py:2834` uses as the "unnamed" sentinel:

```python
lines.append(f"Community {cid}: {', '.join(names)}")            # llm.py:2859
return {int(cid): f"Community {cid}" for cid in communities}    # llm.py:2834
```

Models that pattern-match the input key echo it straight back as the value — `{"5": "Community 5"}` — and `cli.py:1874` then drops exactly that string:

```python
labels.update({cid: v for cid, v in generated_labels.items()
               if v and v != f"Community {cid}"})
```

[...]

PR #2699, opened 2026-08-13
fix: stamp graph provenance from the analysed repo, not the shell cwd

### LINKED ISSUE #2696: extract stamps the invoking repo's HEAD into built_at_commit when run against another repo
## Description

When `graphify extract <target>` is invoked from a working directory inside a *different* git repository, the resulting `graph.json` (and therefore `GRAPH_REPORT.md`) records the **invoking** repo's HEAD as `built_at_commit`, not the target's.

This is the same cwd-anchoring mistake that #2316 fixed for `watch`/`update` — it survives in the `extract` path:

- `cli.py` (extract flow) calls `to_json(...)` without passing `built_at_commit`.
- `export.py:to_json` falls back to a module-local `_git_head()` that runs `git rev-parse HEAD` with no `cwd`, so it inherits the process working directory.
- `cluster` then deliberately preserves the extract-time stamp (#2534), so the wrong SHA is propagated into `GRAPH_REPORT.md`'s "Built from commit" line and kept on every subsequent re-cluster.

## Impact

- The report's own staleness instruction ("run `git rev-parse HEAD` in the repo and compare") can **never** match — the stamped SHA does not exist in the repo the graph describes (`git cat-file -e <sha>` fails).
- Batch-building several targets from one shell stamps the *same* foreign SHA into all of them, so unrelated repos appear to be "built from" one identical commit.
- Because #2534 keeps the extract-time stamp by design, the wrong provenance is permanent until a full re-extract.

## Reproduction (deterministic, no API key needed)

```bash
mkdir A B
git -C A init && git -C A commit --allow-empty -m target
git -C B init && git -C B commit --allow-empty -m invoker
mkdir -p A/graphify-out
cd B
python - <<'EOF'
import json, networkx as nx
G = nx.Graph(); G.add_node("n", label="n")
to_json(G, {0: ["n"]}, "../A/graphify-out/graph.json", force=True)
print(json.load(open("../A/graphify-out/graph.json"))["built_at_commit"])
EOF
git -C ../A rev-parse HEAD   # expected value
git rev-parse HEAD           # actual stamped value (invoker's HEAD)
```

The stamped value is B's HEAD; it should be A's. The same happens end-to-end via `graphify extract ../A` run from inside B.

## Expected

`built_at_commit` identifies the HEAD of the repository the graph describes, regardless of where the command was invoked from — matching the behavior `watch`/`update` already have since #2316.

## Environment

Observed on 0.9.39 (uv tool install) and confirmed present on the current `v8` tip (0.9.41, e4bfd2a) by code inspection: `export.py` `_git_head()` takes no `cwd`, and the extract-path `to_json` call passes no `built_at_commit`.

I have a fix with a regression test ready and can open a PR.

C0KERNEL:
Fixes #2696.

`graphify extract <target>` run from another repo's working directory stamps the **invoker's** HEAD into the target's `graph.json`: the extract write path never passes `built_at_commit`, so `to_json`'s fallback runs `git rev-parse HEAD` in the process cwd. `cluster` then preserves the extract-time stamp by design (#2534), so `GRAPH_REPORT.md`'s "Built from commit" line can name a commit that does not exist in the repo it describes, and the staleness comparison the report suggests can never match. Batch-building several targets from one shell stamps the same foreign SHA into all of them.

This is the same cwd-anchoring mistake #2316 fixed for `watch`/`update`, surviving in the extract path. The fix closes both layers:

- **`cli.py` (extract):** pass `built_at_commit` explicitly, derived from the target via the cwd-aware `watch._git_head(cwd=target)`.
- **`export.py`:** give the module-local `_git_head` the same `cwd` parameter `watch`'s has (#2316), and anchor `to_json`'s fallback on the directory the graph is written into (`output_path` lives in `<target>/graphify-out/`) rather than the shell's cwd — so any other caller that omits `built_at_commit` also gets correct provenance.

## Test

`tests/test_export.py::test_to_json_commit_fallback_uses_output_repo_not_cwd` — builds two throwaway repos, chdirs into one, writes a graph into the other, and asserts the stamp is the target repo's HEAD. Fails before the patch (stamps the invoker), passes after.

`uv run pytest tests/ -q`: 4253 passed, 42 skipped; the 5 failures (`test_ollama_retry_cap.py` ×4, `test_labeling.py::test_label_communities_batches_when_over_batch_size`) are pre-existing on the untouched `v8` tip and unrelated.

### DIFF
--- a/graphify/cli.py
+++ b/graphify/cli.py
@@ -3924,7 +3924,17 @@ def dispatch_command(cmd: str) -> None:
         # passing --allow-partial (the good graph is preserved and the manifest
         # is not stamped, so the retry re-extracts).
         _force_write = cli_allow_partial or not _extraction_incomplete
-        _wrote = _to_json(G, communities, str(graph_json_path), force=_force_write)
+        # Stamp provenance from the ANALYSED repo, not the shell's cwd: without
+        # this, to_json's fallback asks `git rev-parse HEAD` in whatever repo the
+        # command was invoked from, so `graphify extract <target>` run from
+        # another repo's root stamped the invoker's commit into the target's
+        # graph.json — and cluster then propagates that stamp into
+        # GRAPH_REPORT.md (#2534 keeps the extract-time stamp by design). Same
+        # cwd-anchoring mistake #2316 fixed for watch/update, surviving in the
+        # extract path.
+        from graphify.watch import _git_head as _gh_target
+        _wrote = _to_json(G, communities, str(graph_json_path), force=_force_write,
+                          built_at_commit=_gh_target(cwd=Path(target).resolve()))
         if not _wrote:
             # The shrink guard refused: this partial build is smaller than the
             # existing graph. Exit before writing the manifest/marker below, which
--- a/graphify/export.py
+++ b/graphify/export.py
@@ -170,11 +170,21 @@ def attach_hyperedges(G: nx.Graph, hyperedges: list) -> None:
     G.graph["hyperedges"] = existing
 
 
-def _git_head() -> str | None:
-    """Return the current git HEAD commit hash, or None if not in a git repo."""
+def _git_head(cwd: "str | Path | None" = None) -> str | None:
+    """Return git HEAD for the repo containing ``cwd``, or None outside a repo.
+
+    ``cwd`` selects the repository to ask, exactly as in watch._git_head
+    (#2316). Without it the command inherits the caller's working directory,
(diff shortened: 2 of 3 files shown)


safishamsi:
Shipped in v0.9.42 (`graphifyy==0.9.42` on PyPI). Provenance stamped from the analysed repo, not the shell cwd (#2534 family). Credited in the release notes. Thanks @C0KERNEL!

PR #2705, opened 2026-08-13
fix: extract nested JS function declarations (#2653)

### LINKED ISSUE #2653: TS/TSX: function declarations nested inside another function emit no node, so their call edges dangle
**Version:** reproduced on 0.9.40 (and 0.9.17)

A `function` declaration nested inside another function body produces no node. Only top-level declarations are emitted. Any call made from inside such a function then has a source id that no node carries, so the edge is discarded at build time and counts as dangling.

### Repro

`Panel.tsx`
```tsx
function doThing() {}

export function Panel() {
  function handleClick() {
    doThing()
  }
  return <button onClick={handleClick} />
}
```

```python
from pathlib import Path
r = extract([Path('Panel.tsx')], cache_root=Path('.'))
print(sorted(n.get('label','') for n in r['nodes']))
```

**Actual**
```
['Panel()', 'Panel.tsx', 'doThing()']
```

**Expected** — a node for `handleClick()`, contained by `Panel()`, with a `calls` edge to `doThing()`.

### Why it matters

React components keep their event handlers as nested function declarations, so this removes the entire intra-component call graph. In one real component (`ProductOutputPanel.tsx`, 724 lines) **none** of its five handlers — `handleRegenDe`, `handleRegenEn`, `handleRegen`, `handleContentSlotApply`, `handleSave` — is emitted; the file yields 16 nodes and the only function-like ones are two top-level hooks.

The consequence is not just missing nodes. The extraction still emits the edge

```
…productoutputpanel_handlecontentslotapply --calls--> …field_helpers_splicecontentslotintohtml
```

whose source id matches no node, so it is silently dropped and reported as a dangling endpoint. The relationship survives only at file granularity, and "which handler calls this helper" is unanswerable.

### Checked for duplicates

#1740 covers calls inside *anonymous arrow callbacks* (`addEventListener`, `forEach`); #2575 covers `await import()` inside a nested function; #2040 covers nested *class* containment; #1322 covers function symbols assigned via `this.X =` / `exports.X =` / class arrow fields. This one is a plain named `function` declaration nested in a function body, where the symbol itself is never created.

hopstreax:
## Summary

Fixes #2653.

Named JavaScript/TypeScript function declarations nested inside another function were not emitted as graph nodes. As a result, calls from nested functions could be attributed to the enclosing function or become dangling edges.

For example:

```tsx
function doThing() {}

export function Panel() {
  function handleClick() {
    doThing()
  }

  return <button onClick={handleClick} />
}

### DIFF
--- a/graphify/extractors/engine.py
+++ b/graphify/extractors/engine.py
@@ -4101,6 +4101,44 @@ def _extract_generic(
                 if config.ts_module == "tree_sitter_c_sharp" and parent_class_nid:
                     csharp_method_scopes[id(body)] = (node, parent_class_nid)
                 function_bodies.append((func_nid, body))
+                if config.ts_module in (
+                    "tree_sitter_javascript", "tree_sitter_typescript"
+                ):
+                    def _scan_js_nested_functions(parent_nid: str, container_node) -> None:
+                        if container_node is None:
+                            return
+                        for child in container_node.children:
+                            if child.type in (
+                                "function_declaration",
+                                "generator_function_declaration",
+                            ):
+                                name_node = child.child_by_field_name(config.name_field)
+                                if name_node is None:
+                                    for c in child.children:
+                                        if c.type in config.name_fallback_child_types:
+                                            name_node = c
+                                            break
+                                func_name = _read_text(name_node, source) if name_node else None
+                                if func_name and normalize_id(func_name):
+                                    line = child.start_point[0] + 1
+                                    nested_nid = _make_id(parent_nid, func_name)
+                                    add_node(nested_nid, f"{func_name}()", line)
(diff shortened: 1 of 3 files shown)


hopstreax:
Glad to see this got extended to cover arrow-defined components and callbacks as well. My PR addressed the nested named-function declaration case, and this follow-up nicely covers the React patterns that motivated the issue. Thanks for building on it!

safishamsi:
Landed in v0.9.44, just published to PyPI. I cherry-picked this onto `v8` with your authorship preserved in the commit, so both the change and the credit are in the history. Thanks @hopstreax for the fix. Closing since it is now released.

PR #2707, opened 2026-08-13
fix: `affected` silently returns nothing for equivalent path forms

### LINKED ISSUE #2706: `affected` silently returns nothing when the path is spelled differently (./ or absolute)
### What happens

`affected` resolves a file only when the path is spelled exactly as the graph stores it. Any other spelling of the same file returns an empty list with exit code 0 and no warning:

```
$ graphify affected src/brain/retrieve/query.py
… 16 results

$ graphify affected ./src/brain/retrieve/query.py
(nothing)                       # exit 0

$ graphify affected /abs/repo/src/brain/retrieve/query.py
(nothing)                       # exit 0

$ graphify affected does-not-exist.py
(nothing)                       # exit 0
```

Same file, sixteen dependents, three of the four spellings say nothing depends on it.

### Why it matters

The empty result is **indistinguishable from a genuine zero and from a typo** — same output, same exit code, nothing on stderr. `affected` is the blast-radius question, so "nothing depends on this" is an answer people act on before changing a file.

`./` is what shell tab-completion produces, and an absolute path is what any script or editor integration passes.

### How it was found

Measuring `affected` against an independently built import graph (Python `ast`, no tree-sitter). The probe passed absolute paths and measured recall 0.000 for every module — the same failure one layer up, and it took a while to realise the probe was measuring itself.

### Cause

`resolve_seed` (`graphify/affected.py`) compares the query to the stored `source_file` with normalised string equality. The graph stores repo-relative paths, so `./x` and `/abs/repo/x` match nothing while naming the same file.

### Environment

graphify 0.9.41 · Python 3.12 · macOS · corpus is a Python project (~1,400 nodes)

A patch is on `phudayyy:fix/affected-path-forms` with a test that is red without it, and the full suite run both ways (identical failures, all `tests/test_ollama_retry_cap.py`, no ollama in this environment).

phudayyy:
Fixes #2706.

`resolve_seed` compares the query to the stored `source_file` as a plain string, so a file can be named three ways and only one resolves:

```
graphify affected src/x.py       -> 16 results, exit 0
graphify affected ./src/x.py     ->  0 results, exit 0
graphify affected /abs/src/x.py  ->  0 results, exit 0
graphify affected typo.py        ->  0 results, exit 0
```

The last two lines are the problem: a blast-radius tool answering "nothing depends on this" is an answer people act on, and here it is indistinguishable both from a genuine zero and from a typo — same empty list, same exit 0, nothing on stderr. `./` is what shell completion produces; an absolute path is what any script passes.

## The change

Normalise the query to repo-relative form **for the `source_file` comparison only**:

- the label branches above keep the query verbatim, so label resolution is untouched
- non-path queries pass through unchanged — `Path("myFunc()").as_posix()` is `"myFunc()"`
- an absolute path rooted outside the repo is left alone rather than reduced to its basename, which would match an unrelated file of the same name

30 lines, most of them the comment explaining why.

## Verification

`test_affected_resolves_equivalent_path_forms` covers the three spellings. Red without the change, green with it — checked by reverting only `graphify/affected.py` and re-running.

Full suite, both ways:

| | failed | passed |
|---|---|---|
| without patch | 5 | 4252 |
| with patch | 5 | 4253 |

Identical failures, all in `tests/test_ollama_retry_cap.py` — no ollama in this environment — plus the one new test.

## How it was found

Measuring `affected` against an independently built import graph (Python `ast`, no tree-sitter) on a ~1,400-node Python corpus. The probe passed absolute paths and measured recall 0.000 for every module, which is the same failure one layer up.

### DIFF
--- a/graphify/affected.py
+++ b/graphify/affected.py
@@ -67,6 +67,31 @@ def _normalize_label(label: str) -> str:
     return unicodedata.normalize("NFC", label).casefold()
 
 
+def _as_repo_relative(query: str) -> str:
+    """Repo-relative form of a path query, for matching a stored `source_file`.
+
+    The graph stores repo-relative paths, so `./src/x.py` and
+    `/abs/repo/src/x.py` name the same file as `src/x.py` and yet matched
+    nothing. `affected` then printed an empty list and exited 0 — a blast-radius
+    tool answering "nothing depends on this" about a file with sixteen
+    dependents, and indistinguishable from a genuine zero or a typo.
+
+    Non-path queries pass through unchanged: `Path("myFunc()").as_posix()` is
+    `"myFunc()"`, so label resolution is untouched. An absolute path rooted
+    outside the repo is left alone — no basename guessing.
+    """
+    path = Path(query)
+    if path.is_absolute():
+        try:
+            return path.relative_to(Path.cwd()).as_posix()
+        except ValueError:
+            # Rooted outside the repo: nothing here can make it repo-relative,
+            # so leave it alone rather than guess at a basename that would match
+            # some unrelated file with the same name.
+            return query
+    return path.as_posix()
+
+
 def _prefer_file_node(
     graph: nx.Graph,
     node_ids: list[str],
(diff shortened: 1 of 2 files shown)


safishamsi:
Shipped in v0.9.42 (`graphifyy==0.9.42` on PyPI). Resolves a `./`-relative seed. Note: an absolute-path seed still requires the working directory to be the analysed repo root — a follow-up could derive the root from the graph location. Credited in the release notes. Thanks @phudayyy!

Issue #2714, opened 2026-08-13
graphify-mcp: tool failures return isError unset, so a missing project_path graph is indistinguishable from success to any programmatic caller


YNFM:
## Summary

When an MCP tool call fails — for example `project_path` pointing at a directory with no
`graphify-out/graph.json` — the server returns a result with **`isError` unset** and delivers the
failure as ordinary text content:

```
Error executing graph_stats: graph.json not found: C:\some\path\graphify-out\graph.json
```

An LLM reading that will understand it perfectly, so the interactive experience is fine. But
**anything that branches programmatically on the MCP error flag reads a missing graph as a
successful call.** That rules the flag out for health checks, CI gates, and rollout scripts — which
is a shame, because a multi-project server addressed by `project_path` is exactly the setup where a
script wants to ask "is this project's graph actually there?".

The MCP spec's `isError` field exists for precisely this distinction: a tool-level failure that the
model should see, as opposed to a protocol error. Setting it costs nothing and does not change what
the model reads.

## Repro

Any MCP client. Minimal version using the Python SDK:

```python
import asyncio
from mcp import ClientSession, StdioServerParameters
from mcp.client.stdio import stdio_client

async def main():
    params = StdioServerParameters(command="graphify-mcp", args=["/path/to/a/real/graph.json"])
    async with stdio_client(params) as (r, w):
        async with ClientSession(r, w) as s:
            await s.initialize()
            res = await s.call_tool("graph_stats", {"project_path": "/a/dir/with/no/graph"})
            print("is_error:", getattr(res, "is_error", None))
            print("content :", "\n".join(c.text for c in res.content))

asyncio.run(main())
```

## Actual

```
is_error: False
content : Error executing graph_stats: graph.json not found: /a/dir/with/no/graph/graphify-out/graph.json
```

## Expected

```
is_error: True
content : Error executing graph_stats: graph.json not found: ...
```

Same text, flag set.

## Notes

- The rest of the behaviour here is good and worth saying so: a bad `project_path` is handled as a
  tool-level failure that **leaves the server alive and still serving every other graph**. An error
  call between two good ones changed nothing about either. The docstring in `serve.py` says this is
  deliberate — "A client-supplied `project_path` must instead become a tool error, so the shared MCP
  server can continue serving other graphs" — so the intent is clearly there; it just is not
  reflected in the result flag.
- Verified on a server started **with** a valid default graph and on one started **without**, so it
  is not specific to the pure multi-project mode.

## Environment

[...]

Issue #2713, opened 2026-08-13
T-SQL: a bracketed FOREIGN KEY can drop the child table entirely and emit a self-referencing edge tagged EXTRACTED


YNFM:
## Summary

For some bracket-quoted T-SQL, a `FOREIGN KEY ... REFERENCES` clause causes the **child table to be
dropped from the graph entirely** and produces a **self-referencing edge on the parent, tagged
`EXTRACTED`**.

`EXTRACTED` means "explicit in the source". There is no `Customer → Customer` reference anywhere in
the input. So this is not a missing edge — it is a **fabricated edge carrying the highest confidence
grade**, which is worse, because the grade is what a reader uses to decide whether to trust it
without checking.

The identical input with brackets removed is extracted correctly.

## Repro

`s.sql`, alone in a directory:

```sql
CREATE TABLE [dbo].[Customer] (
    [CustomerId] INT NOT NULL PRIMARY KEY,
    [Name]       NVARCHAR(100) NULL
);
GO

CREATE TABLE [dbo].[Invoice] (
    [InvoiceId]  INT NOT NULL PRIMARY KEY,
    [CustomerId] INT NOT NULL,
    CONSTRAINT [FK_Invoice_Customer] FOREIGN KEY ([CustomerId])
        REFERENCES [dbo].[Customer] ([CustomerId])
);
GO
```

```
graphify . --code-only --force
```

## Actual

```
labels: ['s.sql', 'dbo].[Customer']
edges : [('s', 'contains',   's_dbo_customer', 'EXTRACTED'),
         ('s_dbo_customer', 'references', 's_dbo_customer', 'EXTRACTED')]
```

`[dbo].[Invoice]` is **absent from the graph**. The FK it declares appears as `Customer → Customer`.

## Expected

The same file with brackets removed gives the right answer, so this is the behaviour to match:

```
labels: ['s.sql', 'dbo.Customer', 'dbo.Invoice', ...]
edges : [..., ('s_dbo_invoice', 'references', 's_dbo_customer', 'EXTRACTED')]
```

## What I could and could not isolate

Reproduces deterministically, including with `--force` in a fresh directory, so it is not an
incremental-cache artefact.

I could **not** isolate the trigger. Four plausible causes were each tested and **refuted** — nearby
variants all extract correctly:

| Hypothesis | Test | Result |
|---|---|---|
| Objects after the first are dropped | three bracketed `CREATE TABLE`, no FK | **all three extracted** — refuted |
[...]

PR #2716, opened 2026-08-13
fix(sql/mcp): robust bracketed identifier parsing and mcp error flagging

### LINKED ISSUE #2712: T-SQL: bracket-quoted identifiers keep their brackets — [dbo].[Customer] becomes the label `dbo].[Customer`, so no lookup by real name matches
## Summary

The SQL extractor splits a bracket-quoted T-SQL identifier on the wrong characters, so
`[dbo].[Customer]` becomes the node label `dbo].[Customer` — with the schema's closing bracket and
the name's opening bracket both retained.

Bracket quoting is the **default output of SQL Server Management Studio's script generator**, so in
practice this affects essentially every T-SQL file that was scripted out of SSMS rather than
hand-written.

## Repro

`schema.sql`, three statements, nothing else in the directory:

```sql
CREATE TABLE [dbo].[Alpha] (
    [Id] INT NOT NULL PRIMARY KEY
);
GO

CREATE TABLE [dbo].[Beta] (
    [Id] INT NOT NULL PRIMARY KEY
);
GO

CREATE TABLE [dbo].[Gamma] (
    [Id] INT NOT NULL PRIMARY KEY
);
GO
```

```
graphify . --code-only
```

## Actual

```
labels: ['schema.sql', 'dbo].[Alpha', 'dbo].[Beta', 'dbo].[Gamma']
```

## Expected

```
labels: ['schema.sql', 'dbo.Alpha', 'dbo.Beta', 'dbo.Gamma']
```

The identical file with brackets removed produces exactly that, so the extraction path is otherwise
working — it is only the quoting that is mishandled:

```sql
CREATE TABLE dbo.Alpha ( Id INT NOT NULL PRIMARY KEY );
```
```
labels: ['schema.sql', 'dbo.Alpha', ...]
```

## Why it matters beyond appearance

The mangled string is the **node label**, which is the handle every lookup uses. `graphify explain`,
`graphify path A B`, and the MCP `get_node` / `shortest_path` tools all take a label, so a user or
agent asking about `dbo.Customer` — the name in every query, script and error message they have
ever seen — does not match the node. They get a miss, which reads as "not in the graph" rather than
"spelled differently internally".

On our own corpus this showed up as **9 of 15 god nodes** carrying a `dbo].[` label. That is the
architectural summary of the codebase, and more than half of it is unsearchable by the name the
objects actually have.

## Environment

- graphify **0.9.39**, `tree_sitter_sql` present (confirmed importable)
- Python 3.14.2, Windows 11
- `--code-only`, so this is the local AST path with no LLM involved

## Note

There is a second, more serious T-SQL defect found while building this repro — a bracketed foreign
key dropping the child table and emitting a fabricated `EXTRACTED` self-loop. Filed separately since
it is a correctness bug rather than a naming one, and the two do not appear to share a trigger:
every case here mangles labels, but only some drop nodes.

shard-c6:
**Fixes #2712, #2713, and #2714**

### What the Issues Were:
1. **MCP Tool Failures Not Flagged (Issue #2714):** When an MCP tool failed (e.g., due to an invalid graph path), the error was returned as a standard successful text payload. This caused programmatic clients to silently fail, thinking the request succeeded.
2. **Missing Nodes on Bracketed T-SQL (Issue #2713):** A bracketed `FOREIGN KEY` (like `[dbo].[Customer]`) and T-SQL `GO` statements caused `tree-sitter-sql` to produce `ERROR` nodes. This catastrophic parser failure swallowed subsequent `CREATE TABLE` definitions, dropping nodes from the graph and generating incorrect self-referencing edges on previous tables.
3. **Mangled Identifiers (Issue #2712):** Bracket-quoted identifiers were incorrectly parsed, dropping the opening bracket but retaining the closing bracket (e.g., `dbo].[Customer]`), breaking reference lookups.

### How They Have Been Handled:
- **MCP Error Handling:** Updated `_on_call_tool` and `call_tool` in `graphify/serve.py` to directly return a `CallToolResult` object with `isError=True` whenever an exception is caught or an unknown tool is called.
- **Robust SQL Fallback Extraction:** Updated `graphify/extractors/sql.py` to use a global regex fallback (`_IDENT_PATTERN`) that natively supports bracketed `[identifiers]`, quoted `"identifiers"`, and backticks. 
- **Graceful AST Recovery:** The SQL extractor now identifies un-parsed `CREATE TABLE` and `CREATE VIEW` statements and correctly links references even when `tree-sitter-sql`'s AST completely fails to parse the file due to unsupported T-SQL syntax.
- **Identifier Normalization:** Modified `_norm_ident` to proactively strip all unbalanced leading and trailing delimiters caused by tree-sitter AST drops, ensuring mangled names properly normalize to their exact underlying identifiers.

### MCP Version Compatibility

This fix maintains compatibility with both mcp 1.x and 2.x by:
- Using a runtime check (`hasattr(Server, "list_tools")`) to detect the installed version
- **mcp 1.x**: Uses the `@server.list_tools()` decorator API. The SDK automatically wraps 
  handler returns (raw `list[TextContent]` → `ListToolsResult`)
- **mcp 2.x**: Uses the `on_list_tools=...` callback API with explicit `(ctx, params) -> Result` 
  contracts. Handlers must return result objects directly.

The `call_tool` function now returns `CallToolResult` objects (which exist in both 
versions) with `isError=True` set on failures. This works correctly in both:
- **1.x**: The SDK wraps the result in a `CallToolResult` wrapper (no-op)
- **2.x**: The result is returned directly to the client

No version-specific branching is needed for `CallToolResult` construction.

### DIFF
--- a/graphify/extractors/sql.py
+++ b/graphify/extractors/sql.py
@@ -20,11 +20,12 @@ def _norm_ident(name: str) -> str:
     parts = []
     for part in name.split("."):
         p = part.strip()
-        if len(p) >= 2 and ((p[0] == p[-1] and p[0] in ('"', "`"))
-                            or (p[0] == "[" and p[-1] == "]")):
-            p = p[1:-1]
+        p = re.sub(r'^["`\[]+|["`\]]+$', '', p)
         parts.append(p.lower())
     return ".".join(parts)
+
+
+_IDENT_PATTERN = r'(?:\"[^\"\n]+\"|\[[^\]\n]+\]|`[^`\n]+`|[\w$]+)(?:\s*\.\s*(?:\"[^\"\n]+\"|\[[^\]\n]+\]|`[^`\n]+`|[\w$]+))*'
 
 
 def extract_sql(path: Path, content: str | bytes | None = None) -> dict:
@@ -72,6 +73,10 @@ def extract_sql(path: Path, content: str | bytes | None = None) -> dict:
         return source[n.start_byte:n.end_byte].decode("utf-8", errors="replace")
 
     def _obj_name(n) -> str | None:
+        text = _read(n)
+        m = re.search(rf"(?:CREATE|ALTER)\s+(?:OR\s+(?:REPLACE|ALTER)\s+)?(?:TABLE|VIEW|FUNCTION|PROCEDURE)\s+(?:IF\s+NOT\s+EXISTS\s+)?({_IDENT_PATTERN})", text, re.IGNORECASE)
+        if m:
+            return m.group(1)
         for c in n.children:
             if c.type == "object_reference":
                 return _read(c)
@@ -263,7 +268,7 @@ def extract_sql(path: Path, content: str | bytes | None = None) -> dict:
             for m in re.finditer(
                 r"CREATE\s+(?:OR\s+REPLACE\s+)?(?:FUNCTION|PROCEDURE)\s+"
                 r"(?:IF\s+NOT\s+EXISTS\s+)?"
-                r"((?:\"[^\"\n]+\"|[\w$]+)(?:\s*\.\s*(?:\"[^\"\n]+\"|[\w$]+))*)",
+                rf"({_IDENT_PATTERN})",
                 text, re.IGNORECASE,
             ):
                 name = m.group(1)
@@ -379,6 +384,14 @@ def extract_sql(path: Path, content: str | bytes | None = None) -> dict:
 
     _collect_defined_names(root)
 
+    src_text = source.decode("utf-8", errors="replace")
+    if root.has_error:
+        for m in re.finditer(rf"CREATE\s+(?:TABLE|VIEW)\s+({_IDENT_PATTERN})", src_text, re.IGNORECASE):
+            name = m.group(1)
+            norm = _norm_ident(name)
+            if norm not in table_nids:
+                table_nids[norm] = _make_id(stem, name)
+
     # Secondary bare-name aliases: a reference written without a schema
     # (`REFERENCES users`) should resolve to a schema-qualified definition
     # (`public.users`) when that is unambiguous. Never shadow an explicit
@@ -405,8 +418,7 @@ def extract_sql(path: Path, content: str | bytes | None = None) -> dict:
     # (e.g. Firebird COMPUTED BY columns push constraints out of the tree entirely).
     # Snapshot after tree walk so we don't re-emit edges already captured above.
     emitted = {(e["source"], e["target"]) for e in edges if e["relation"] == "references"}
-    src_text = source.decode("utf-8", errors="replace")
-    for m in re.finditer(r"CREATE\s+TABLE\s+([\w$]+)\s*\(", src_text, re.IGNORECASE):
+    for m in re.finditer(rf"CREATE\s+TABLE\s+({_IDENT_PATTERN})\s*\(", src_text, re.IGNORECASE):
         tbl_name = m.group(1)
         tbl_nid = table_nids.get(_norm_ident(tbl_name))
         if tbl_nid is None:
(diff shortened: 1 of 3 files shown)


shard-c6:
adding to the previous summary ..

### MCP Version Compatibility

This fix maintains compatibility with both mcp 1.x and 2.x by:
- Using a runtime check (`hasattr(Server, "list_tools")`) to detect the installed version
- **mcp 1.x**: Uses the `@server.list_tools()` decorator API. The SDK automatically wraps 
  handler returns (raw `list[TextContent]` → `ListToolsResult`)
- **mcp 2.x**: Uses the `on_list_tools=...` callback API with explicit `(ctx, params) -> Result` 
  contracts. Handlers must return result objects directly.

The `call_tool` function now returns `CallToolResult` objects (which exist in both 
versions) with `isError=True` set on failures. This works correctly in both:
- **1.x**: The SDK wraps the result in a `CallToolResult` wrapper (no-op)
- **2.x**: The result is returned directly to the client

No version-specific branching is needed for `CallToolResult` construction.

**Enhanced Diff:**
```diff
--- a/graphify/serve.py
+++ b/graphify/serve.py
@@ -1513,8 +1513,15 @@ def _build_server(graph_path: str):
 
     # NOTE: no decorators here — the handlers below are plain coroutines,
     # bound to the Server at the END of this function in a version-aware way:
-    # mcp 1.x exposes the @server.list_tools()/... decorator API, mcp 2.x
-    # replaced it with on_list_tools=/... constructor callbacks.
+    #   - mcp 1.x exposes @server.list_tools()/call_tool()/... decorators.
+    #     The SDK automatically wraps handler returns (str -> TextContent,
+    #     list[TextContent] -> CallToolResult, etc).
+    #   - mcp 2.x replaced decorators with on_list_tools=/on_call_tool=/... 
+    #     constructor callbacks using the (ctx, params) -> Result contract.
+    #     Handlers must construct and return result objects directly.
+    # The version check uses hasattr(Server, "list_tools") to detect which
+    # API is available at runtime, allowing a single codebase to support both.
+    # See lines 1995 (1.x) and 2003 (2.x) for the registration branches.
     async def list_tools() -> list[types.Tool]:
         _tools = [
             types.Tool(
@@ -1973,6 +1980,9 @@ def _build_server(graph_path: str):
         raise ValueError(f"Unknown resource: {uri_str}")
 
     async def call_tool(name: str, arguments: dict) -> types.CallToolResult:
+        # Both mcp 1.x and 2.x define CallToolResult, so isError=True
+        # can be set regardless of version. This unified approach avoids
+        # per-version branching in the implementation.
         arguments = dict(arguments or {})
         project_path = arguments.pop("project_path", None)
         handler = _handlers.get(name)
```

shard-c6:
### Graphify Review Fix: `_norm_ident` delimiter stripping

The Graphify bot correctly identified an edge case where the updated `_norm_ident` function corrupted identifiers consisting entirely of delimiters (e.g., stripping multiple delimiters from ends because of the `+` quantifier in the regex). 

The issue has been fixed by replacing the greedy `+` regex substitution with two precise `re.sub` calls that strip exactly one unbalanced leading/trailing delimiter, successfully retaining valid nested quotes while continuing to correctly handle the missing opening bracket issue (`dbo].[Customer]`).

**Enhanced Diff:**
```diff
--- a/graphify/extractors/sql.py
+++ b/graphify/extractors/sql.py
@@ -17,10 +17,16 @@ def _norm_ident(name: str) -> str:
     `table_nids` keys and lookups — node ids and display labels keep the
     original text.
     """
+    if not name:
+        return name
+    # 2712: tree-sitter AST drops the opening bracket on syntax errors but retains the
+    # closing bracket (e.g. `dbo].[Customer]`). We must strip all leading/trailing
+    # delimiter characters regardless of whether they are balanced, otherwise references break.
     parts = []
     for part in name.split("."):
         p = part.strip()
-        p = re.sub(r'^["`\[]+|["`\]]+$', '', p)
+        p = re.sub(r'^["`\[]', '', p)
+        p = re.sub(r'["`\]]$', '', p)
         parts.append(p.lower())
     return ".".join(parts)
```

shard-c6:
Hi Maintainers! Please review the PR.

PR #2720, opened 2026-08-13
Honor GRAPHIFY_OUT in installed docs and IDE plugins

### LINKED ISSUE #2571: graphify skill doesnt respect GRAPHIFY_OUT
When you follow the install guide but you create the gaphify repo in a different folder than the default (using GRAPHIFY_OUT=.foo/graphify for example), then `graphify opencode install` (or any other harness for that matter) make it so that the AI skill does not use the subfolder and simply will never find anything.

ayushcodes10:
Cursor/Devin rules, the always-on CLAUDE.md/AGENTS.md/etc blocks, the Antigravity MCP config example, and the Kilo/OpenCode plugin's existsSync check all hardcoded "graphify-out", so a custom GRAPHIFY_OUT (relative or absolute) was silently ignored and the graph was never found. Fixes #2571.


ayushcodes10:

ayushcodes10:
Rebased onto current v8 (trivial conflict: upstream added an unrelated _refresh_all_version_stamps function right where this PR's functions were inserted, resolved by keeping both). No changes beyond the rebase; full suite plus the always-on roundtrip and skillgen checks green.

ayushcodes10:

ayushcodes10:
Investigated all 5 findings, 3 were real:

- _js_graph_exists_expr escaped quotes and backslash but not a raw newline (SyntaxError inside a JS string literal). Fixed via a shared _js_string_escape helper.
- The Kilo/OpenCode reminder text ran through _out_doc's naive, unescaped substring replace, including inside a JS string literal that itself builds a shell command string later assigned to output.args.command and actually executed. A quote could break the JS string, and separately a shell metacharacter could reach the shell once that string evaluates, since JS-escaping and shell-escaping are different concerns the same characters don't satisfy at once. Carved that mention out into its own placeholder and allowlisted plain path characters for it, falling back to the packaged default for anything else, rather than trying to correctly double-escape for both nested contexts at once.
- The Antigravity MCP config example built JSON by hand with no escaping at all, a quote in GRAPHIFY_OUT printed invalid JSON for the user to copy. Now uses json.dumps.

Checked #1 (naive substring replace corrupting non-path occurrences) against every actual "graphify-out" mention in the always-on/Cursor/Devin templates, none are non-path occurrences currently, so nothing to fix there right now.

All three fixes verified against adversarial GRAPHIFY_OUT values (embedded quotes, backticks, shell metacharacters, newlines) with dedicated regression tests.

ayushcodes10:
Investigated this round's bot findings (matching 9215f29).

Confirmed real, fixed in 568ecae: "Custom output directories with spaces are replaced by the default in plugin reminders" (and the related "_shell_safe_out_name discards valid custom output dirs containing allowed-but-unlisted chars" finding, same underlying gap). A space is a common, entirely plausible path character (My Output Dir, or an absolute path through a directory with one) that needs no escaping in either nested context: a JS string literal is untouched by a bare space, and the echo command always wraps the reminder text in double quotes, so a space inside it is preserved verbatim rather than starting a new shell word in bash or PowerShell. Added to the allowlist, regression test added.

Investigated and found to be a false positive: "_out_doc replaces graphify-out with GRAPHIFY_OUT unescaped into JS/shell plugin templates". Checked the current _KILO_PLUGIN_JS/_OPENCODE_PLUGIN_JS definitions directly: neither calls _out_doc at all (that wrapper is only used for _CURSOR_RULE/_DEVIN_RULES and the always-on block). The plugin templates build their shell-echo text via _shell_safe_out_name specifically. Verified empirically with an adversarial GRAPHIFY_OUT value (embedded quote, shell metacharacters) that the shell echo command in both generated templates falls back to the safe default with no injected text reaching it. Likely describes an earlier state of this PR's diff rather than the current one. No code change made for this finding.

ayushcodes10:
Follow up on this round's bot findings (matching 568ecae): these restate the same underlying trade-off already addressed twice this round (the Windows-path colon/backslash fix and the space fix above), without a new concrete character or example. _shell_safe_out_name's allowlist is intentionally conservative by design (documented in its own docstring): the characters still excluded past this round's fixes are live shell metacharacters (", $, `, ;, |, &, newline, and similar) that cannot be allowed without a materially more complex escaping scheme correct in both the JS-string and shell layers at once -- the whole reason this function uses an allowlist rather than double-escaping in the first place. Every concrete gap actually reported so far (a Windows absolute path, a space in a custom directory name) has been fixed with a regression test. Happy to look at a further concrete character or example if one comes up, but there isn't a specific new one in this finding to act on. No code change made for this round.

PR #2727, opened 2026-08-13
Add native /graphify slash command for OpenCode

### LINKED ISSUE #2709: `/graphify` is not an OpenCode command
## Summary

The documented `/graphify` invocation is not available as a slash command in the OpenCode TUI : OpenCode skills are not slash commands, only `command/*.md` files or `config.command` are.

## Environment
- OpenCode 1.18.18 (TUI)
- graphify skill installed at `~/.config/opencode/skills/graphify/SKILL.md`
- graphify CLI 0.9.41 (`uv tool install graphifyy`)
- OS: Linux

## Steps to reproduce
1. Install the graphify skill in the global OpenCode skills directory
2. Restart OpenCode, type `/` in the prompt box
3. `/graphify` does not appear in the command menu ("No matching items")

## Expected
`/graphify` shows up as a slash command, as stated in the SKILL.md usage docs and as it does in Claude Code.

## Actual
The skill loads correctly (it appears in the agent's <available_skills>), but OpenCode agent skills are not slash commands, they are loaded on-demand by the agent via the skill tool based on description matching. Only commands (`command/*.md` files or `config.command`) appear in the `/ menu. Users following the docs expect a direct slash command and conclude the installation is broken.

## Root cause
The opencode platform install only copies the skill (to ~/.config/opencode/skills/graphify/) and the reminder plugin. It lacks the OpenCode command registration.

## Additional finding
`_install_opencode_plugin` writes `.opencode/plugins/graphify.js` and registers the relative path ".opencode/plugins/graphify.js" in the plugin array of `.opencode/opencode.json` (install.py:1347-1372). That registration is both unnecessary (OpenCode auto-loads local plugins from the plugins/ directory) and dangling for global scope (the global config is `~/.config/opencode/opencode.json`, so the relative path never resolves). It leaves a dead entry in the user's config.

## Suggested fix
1. Ship a command file (`.opencode/command/graphify.md` in the package) whose template loads the skill and forwards $ARGUMENTS.
2. Fix `_install_opencode_plugin` (install.py:1347-1372): drop the plugin-array registration entirely. Also update `_uninstall_opencode_plugin` to match.

ayushcodes10:
OpenCode skills are loaded on-demand by the agent via description matching; they are not slash commands. Only a markdown file under .opencode/commands/ becomes a /-menu entry, so installing just the skill and reminder plugin left /graphify absent from the TUI command menu even though the skill loaded correctly and the docs told users to type it.

Ship a command-opencode.md template (mirroring the existing Kilo command file) that loads the skill and forwards $ARGUMENTS, and wire it into opencode install/uninstall alongside the existing plugin.

Also drops the .opencode/opencode.json "plugin" array registration: OpenCode already auto-loads every file under .opencode/plugins/, so the registration was redundant, and for a global install it wrote a project-relative path into a project-relative config file that a global setup never reads, leaving a dead entry behind.

Fixes #2709.

### DIFF
--- a/graphify/__main__.py
+++ b/graphify/__main__.py
@@ -114,7 +114,7 @@ from graphify.install import (  # noqa: E402,F401
     _KILO_CONFIG_JSONC_PATH,
     _OPENCODE_PLUGIN_JS,
     _OPENCODE_PLUGIN_PATH,
-    _OPENCODE_CONFIG_PATH,
+    _OPENCODE_COMMAND_PATH,
     _PLATFORM_CONFIG,
 )
 from graphify.cli import (  # noqa: E402,F401
--- a/graphify/command-opencode.md
+++ b/graphify/command-opencode.md
@@ -1,0 +1,14 @@
+---
+description: Build or query a graphify knowledge graph
+---
+
+Invoke the `graphify` skill immediately with these arguments: $ARGUMENTS
+
+If no arguments were supplied, treat the target path as `.`.
+
+Examples:
+- `/graphify`
+- `/graphify src --update`
+- `/graphify query "what connects auth to billing?"`
+
+Do not answer from raw files before handing off to the `graphify` skill.
--- a/graphify/install.py
+++ b/graphify/install.py
@@ -704,6 +704,7 @@ def install(platform: str = "claude", *, project: bool = False, project_dir: Pat
 
     if platform == "opencode":
         _install_opencode_plugin(project_dir if project else Path("."))
+        _install_opencode_command(project_dir if project else Path("."))
 
     if project:
         _print_project_git_add_hint([_project_scope_root(skill_dst, project_dir)])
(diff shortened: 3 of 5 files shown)


ayushcodes10:

ayushcodes10:
Rebased onto current v8. The conflict was upstream independently adding a project parameter to _resolve_graphify_exe right at this PR's insertion point, unrelated to the opencode.json cleanup here; resolved by keeping both. No changes beyond the rebase; full suite plus skillgen check green.

ayushcodes10:

ayushcodes10:
Checked this finding, it's real: both _install_opencode_plugin's and _install_opencode_command's docstrings promise ~/.config/opencode/ for a non-project install, but every call site passes Path(".") instead. Confirmed this predates this PR though, _install_opencode_plugin has had the exact same issue since before #2709; _install_opencode_command just inherited the same pattern from its sibling for consistency.

Properly fixing it needs both functions (and their uninstall counterparts) to take an explicit project flag or a pre-resolved base path rather than always joining onto whatever project_dir they're handed, since the global case also needs to drop the extra .opencode/ segment (~/.config/opencode/plugins/graphify.js, not ~/.config/opencode/.opencode/plugins/graphify.js) across several call sites (install, _agents_install, _agents_uninstall, uninstall_all). That's a separate, focused fix rather than something to bundle into "add the missing slash command", filed as #3398.

Issue #563, opened 2026-04-27
Two extractor bugs systematically inflate god-node centrality (rationale fragments + calls direction inversion)


WillRiverpoint:
# Two extractor bugs systematically inflate god-node centrality

**graphifyy version:** 0.4.21
**Python:** 3.14.4
**OS:** Windows 11
**Corpus:** ~1,300,000 words across 209 files (mixed Python middleware + markdown vault)
**Run mode:** `--update`

## TL;DR

After running `graphify --update` on a code+docs corpus, the highest-degree "god node" surfaced was a passive Python class (`SQLiteQueue`, 67 edges, 54 INFERRED). Audit of the 54 inferred edges found:

- **0** correctly-stated outbound edges
- **20** correct-relationship-but-wrong-direction (`SQLiteQueue --calls--> contact_form()` when truth is the reverse)
- **29** pointing at phantom `*_rationale_NNN` nodes that are docstring text fragments, not real entities
- **5** categorically wrong (claiming SQLiteQueue uses phone-validation tests + Odoo-tag chatter — unrelated domains)

True meaningful relationships from this node ≈ 8, all inbound. Centrality inflated ≈ 3×.

Site-wide, this graph has:
- **280 rationale pseudo-nodes (16.3% of all 1720 nodes)**
- **158 `--uses-->` edges pointing AT those pseudo-nodes** (the broken pattern, by my count this run)
- **279 non-AST `calls` edges** (likely candidates for the direction-inversion bug)

The same pattern almost certainly inflates every other god node ranking and a meaningful fraction of "surprising connections" findings.

## Bug 1: Rationale extraction creates one node per docstring paragraph

The skill prompt for semantic extraction asks subagents to extract rationale (which is a great feature). But subagents currently emit **one node per docstring/comment paragraph**, with IDs like `<filestem>_rationale_<line>`. For a single Python file with 4 multi-paragraph docstrings, that produces 4-15+ rationale pseudo-nodes.

These nodes then attract `--uses-->` edges from every imported entity in the same file. So importing `SQLiteQueue` in `test_customer_dedup.py` causes SQLiteQueue to gain edges like:

```json
{
  "source": "queue_sqlitequeue",
  "target": "test_customer_dedup_rationale_46",
  "relation": "uses",
  "confidence": "INFERRED",
  "confidence_score": 0.5
}
```

[...]

PR #2730, opened 2026-08-14
fix(export): stabilize graph.json field order across a read-rebuild round-trip

C0KERNEL:
Completes #2582.

#2582 made `graph.json` **collection order** byte-stable — the order of nodes, links, and hyperedges across each array — but its scope note deliberately stopped there: *"intentionally limited to collection ordering; mapping-key ordering is unchanged."* This PR closes that gap.

`networkx.node_link_data()` always appends the node key (`id`) at the **end** of each node dict. On a cold build `id` is still an inline attribute and lands mid-dict; after `build_from_json` consumes `id` as the pure node key, a re-serialize appends it last. The values are identical, but the field order churns — so a `graph.json` that is written, reloaded (e.g. by `update`/`cluster-only`), and re-written is **not** byte-stable, defeating #2582's guarantee on the mapping-key axis:

- Cold build: `["label", "file_type", "source_file", "id", "community", "norm_label"]`
- After read → rebuild → write: `["label", "file_type", "source_file", "community", "norm_label", "id"]`

The fix emits a canonical key order in `to_json`, just before the existing record sort: nodes lead with `id`, links with `source`/`target`/`relation`, then remaining keys sorted. The serialized form is now invariant regardless of how the attribute was stored in memory, so `build → write → read → write` is byte-identical. Endpoints are untouched — the `_src`/`_tgt` direction restore (#563/#1061) already guards `source`/`target`, and this change only reorders keys, never values.

## Reproduction (deterministic, no API key needed)

```python
import json, tempfile, os
from pathlib import Path

extraction = {
  "nodes": [
    {"id":"a_foo","label":"foo","file_type":"code","source_file":"a.py"},
    {"id":"b_bar","label":"bar","file_type":"code","source_file":"b.py"},
  ],
  "edges": [
    {"source":"a_foo","target":"b_bar","relation":"calls",
     "confidence":"EXTRACTED","confidence_score":1.0,"source_file":"a.py"},
  ],
  "hyperedges": [],
}
d = tempfile.mkdtemp()
p1 = os.path.join(d,"g1.json")
to_json(build_from_json(extraction), {0:["a_foo","b_bar"]}, p1, built_at_commit="x", force=True)
raw1 = json.loads(Path(p1).read_text())
p2 = os.path.join(d,"g2.json")
to_json(build_from_json(raw1), {0:["a_foo","b_bar"]}, p2, built_at_commit="x", force=True)
print("byte-identical:", Path(p1).read_bytes() == Path(p2).read_bytes())  # False before, True after
```

## Test

`tests/test_export.py::test_to_json_field_order_stable_across_read_rebuild` — builds a graph, writes it, reads it back with `build_from_json`, writes again, and asserts the two files are byte-identical with `id` leading every node and `source`/`target` leading every link. Fails before the patch (the round-trip churns `id`'s position), passes after. The existing `test_to_json_sorts_graph_collections_across_insertion_order` missed this because it only varied insertion order within a single build and never exercised a read-rebuild.

`uv run pytest tests/ -q`: 4335 passed, 45 skipped. The 5 failures (`test_ollama_retry_cap.py` ×4, `test_labeling.py::test_label_communities_batches_when_over_batch_size`) are pre-existing on the untouched `v8` tip and unrelated — verified by reverting `export.py`/`test_export.py` to pristine `origin/v8`, where they still fail. They touch the Ollama retry cap and community-label batching, not export/serialization.

### DIFF
--- a/graphify/export.py
+++ b/graphify/export.py
@@ -325,6 +325,22 @@ def to_json(G: nx.Graph, communities: dict[int, list[str]], output_path: str, *,
         if true_src is not None and true_tgt is not None:
             link["source"] = true_src
             link["target"] = true_tgt
+    # Canonicalize the key order WITHIN each node/link dict. node_link_data always
+    # appends the node key (`id`) at the end, so a node whose `id` was an inline
+    # attribute on a cold build (position varies) lands last after a read-rebuild
+    # (build_from_json consumes `id` as the pure node key). The values are
+    # identical either way, but the field order churns, so a byte-diff of two
+    # equivalent graph.json files is noisy and any position-sensitive consumer
+    # sees a spurious change on every round-trip. Emit a stable order — the
+    # identity keys first, then the remaining keys sorted — so the serialized
+    # form is invariant regardless of how the attribute was stored in memory.
+    def _canonical(item: dict, lead: tuple[str, ...]) -> dict:
+        leading = [k for k in lead if k in item]
+        rest = sorted(k for k in item if k not in leading)
+        return {k: item[k] for k in (*leading, *rest)}
+
+    data["nodes"] = [_canonical(n, ("id", "label")) for n in data["nodes"]]
+    data["links"] = [_canonical(link, ("source", "target", "relation")) for link in data["links"]]
     data["nodes"].sort(key=_json_sort_key)
     data["links"].sort(key=_json_sort_key)
     if "hyperedges" not in getattr(G, "graph", {}):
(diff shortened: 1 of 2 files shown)


safishamsi:
Shipped in v0.9.47 via authorship-preserving cherry-pick (you keep contributor-graph credit), with an added non-ASCII round-trip byte-identity test. Thanks @C0KERNEL! Release: https://github.com/Graphify-Labs/graphify/releases/tag/v0.9.47

PR #2734, opened 2026-08-14
fix(claude): read text blocks after leading ThinkingBlock

### LINKED ISSUE #2697: graphify extract --backend claude crashes on 'ThinkingBlock' object has no attribute 'text'
# `graphify extract --backend claude` crashes on `'ThinkingBlock' object has no attribute 'text'`

## Summary

`graphify/llm.py` reads the Anthropic response as `resp.content[0].text`. On every current
Claude model, `content[0]` is a **`ThinkingBlock`**, not a `TextBlock` — so semantic
extraction fails with:

```
AttributeError: 'ThinkingBlock' object has no attribute 'text'
```

`--code-only` is unaffected (it never calls the LLM).

This is not limited to unusual setups. Extended thinking is **on by default** on current
models, and thinking blocks are present in `content` *even when the thinking text is
suppressed*: with `thinking.display: "omitted"` — the default — responses still include a
`thinking` block whose `thinking` field is an empty string. So the first block is a
`ThinkingBlock` on the ordinary default path, with no special configuration.

## Root cause

Two sites, both assuming `content[0]` is the text block:

**`graphify/llm.py:1326`** — semantic extraction:
```python
raw_content = resp.content[0].text if resp.content else None
```

**`graphify/llm.py:2597`** — community labelling (`label` / `cluster-only`):
```python
return resp.content[0].text if resp.content else ""
```

With thinking enabled, the SDK returns:

```python
[ThinkingBlock(type='thinking', thinking='...'), TextBlock(type='text', text='...')]
```

The Anthropic SDK documents that `content` may contain `ThinkingBlock` objects; indexing
position 0 for the text is not safe.

## Steps to reproduce

```python
import anthropic

client = anthropic.Anthropic(api_key="YOUR_API_KEY")

resp = client.messages.create(
    model="claude-opus-5",   # thinking is on by default on current models
    max_tokens=1024,
    messages=[{"role": "user", "content": 'Return valid JSON: {"nodes":[]}'}],
)

print([b.type for b in resp.content])   # -> ['thinking', 'text']
print(resp.content[0].text)             # -> AttributeError: 'ThinkingBlock' ...

# What the code should do instead: take the first block of type "text"
text = next((b.text for b in resp.content if getattr(b, "type", None) == "text"), None)
print(text)                             # -> '{"nodes":[]}'
```

The same happens through any Anthropic-compatible gateway with reasoning enabled
(LiteLLM, claude-code-router, corporate proxies), which is how we hit it.

## Actual behavior

Real run, 26 markdown files, 15 semantic chunks:

```
[graphify] chunk 1/15 failed: 'ThinkingBlock' object has no attribute 'text'
[graphify] chunk 3/15 failed: 'ThinkingBlock' object has no attribute 'text'
...
[graphify] WARNING: 13/15 semantic chunk(s) failed — see errors above. Partial results returned.
```

**13 of 15 chunks failed; 2 succeeded.** The survivors are the responses that happened to
come back with no thinking block, which is what makes the failure look intermittent rather
than systematic.

## Expected behavior

`graphify extract --backend claude` completes regardless of whether the model or gateway
returns thinking blocks.

## Suggested fix

Take the first `text` block instead of the first block:

```diff
--- a/graphify/llm.py
+++ b/graphify/llm.py
@@ -1323,7 +1323,9 @@ def _anthropic_extract_chunk(...):
          messages=[{"role": "user", "content": _anthropic_content(user_message, images or [])}],
      )
-    raw_content = resp.content[0].text if resp.content else None
+    raw_content = next(
+        (b.text for b in (resp.content or []) if getattr(b, "type", None) == "text"), None
+    )
      result = _parse_llm_json(raw_content or "{}")

@@ -2594,7 +2596,9 @@ def _anthropic_label_chunk(...):
          u = getattr(resp, "usage", None)
          if u is not None:
              _rec(getattr(u, "input_tokens", 0), getattr(u, "output_tokens", 0))
-        return resp.content[0].text if resp.content else ""
+        return next(
+            (b.text for b in (resp.content or []) if getattr(b, "type", None) == "text"), ""
+        )
```

Backward compatible: with no thinking block, `next(...)` finds the only `TextBlock` and
returns exactly what the original code returned.

We applied this patch locally and the same corpus that produced 13/15 failures completed
12/12 chunks on the next run.

## Environment

| Field | Value |
|---|---|
| graphify (PyPI `graphifyy`) | 0.9.41 |
| `anthropic` SDK | 0.121.0 |
| Python | 3.13.12 |
| OS | Windows 11 |
| Endpoint | Anthropic-compatible gateway with reasoning enabled |

## Note on the second site

`llm.py:2597` (community labelling) has the identical pattern. It surfaces only when
`graphify label` / `cluster-only` invokes LLM naming, so it fails less visibly — but it is
the same bug and needs the same fix.

mdshzb04:
## Summary

`graphify extract --backend claude` and community labeling both read `resp.content[0].text`. On current Claude models, the first block is often a `ThinkingBlock`, not the answer — so extraction fails with `ThinkingBlock object has no attribute text`.

## What changed

Added `_anthropic_response_text()` to pick the first `text` block instead of assuming index 0. Wired it into `_call_claude` and the claude branch of `_call_llm`. Added regression tests mirroring the existing Bedrock reasoning-block coverage.

## Why

Fixes #2697. I kept hitting chunk failures on semantic extraction where every error was the ThinkingBlock attribute error, not the actual API problem.

## Test plan

- [x] `pytest tests/test_image_vision.py -k anthropic_response_text`
- [x] `pytest tests/test_image_vision.py::test_call_claude_parses_thinking_model_response`
- [x] `pytest tests/test_llm_backends.py::test_call_llm_claude_client_built_with_timeout_and_retries`

### DIFF
--- a/graphify/llm.py
+++ b/graphify/llm.py
@@ -1065,6 +1065,26 @@ def _parse_llm_json(raw: str) -> dict:
     return {"nodes": [], "edges": [], "hyperedges": []}
 
 
+def _anthropic_response_text(content, default: str | None = None) -> str | None:
+    """Return the first Anthropic content block that carries text.
+
+    Current Claude models emit a ``ThinkingBlock`` ahead of the ``TextBlock``
+    when extended thinking is enabled (including the default-on path where the
+    thinking text is omitted). Indexing ``content[0]`` therefore raises or
+    yields no text (#2697). Select on the block's type instead of its position.
+    """
+    if not content:
+        return default
+    for block in content:
+        block_type = getattr(block, "type", None)
+        if block_type is not None and block_type != "text":
+            continue
+        text = getattr(block, "text", None)
+        if isinstance(text, str) and text.strip():
+            return text
+    return default
+
+
 def _bedrock_response_text(resp: dict, default: str = "") -> str:
     """Return the first Converse content block that carries text.
 
@@ -1331,7 +1351,7 @@ def _call_claude(api_key: str, model: str, user_message: str, max_tokens: int = 
         system=_extraction_system(deep=deep_mode),
         messages=[{"role": "user", "content": _anthropic_content(user_message, images or [])}],
     )
-    raw_content = resp.content[0].text if resp.content else None
+    raw_content = _anthropic_response_text(resp.content)
     result = _parse_llm_json(raw_content or "{}")
     result["input_tokens"] = resp.usage.input_tokens if resp.usage else 0
     result["output_tokens"] = resp.usage.output_tokens if resp.usage else 0
@@ -2602,7 +2622,7 @@ def _call_llm(
         u = getattr(resp, "usage", None)
         if u is not None:
             _rec(getattr(u, "input_tokens", 0), getattr(u, "output_tokens", 0))
-        return resp.content[0].text if resp.content else ""
+        return _anthropic_response_text(resp.content, default="")
 
     if backend == "claude-cli":
         import platform, shutil, subprocess
--- a/tests/test_image_vision.py
+++ b/tests/test_image_vision.py
@@ -401,6 +401,53 @@ def test_bedrock_response_text_tolerates_malformed_blocks():
     assert llm._bedrock_response_text(resp, default="{}") == _NODE_JSON
 
 
+def test_anthropic_response_text_single_text_block_unchanged():
+    content = [SimpleNamespace(type="text", text=_NODE_JSON)]
+    assert llm._anthropic_response_text(content) == _NODE_JSON
+
+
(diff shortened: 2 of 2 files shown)


safishamsi:
Shipped in v0.9.47. Landed via authorship-preserving cherry-pick so you keep contributor-graph credit, plus a test locking the first-text-block semantics. Thanks @mdshzb04! Release: https://github.com/Graphify-Labs/graphify/releases/tag/v0.9.47

PR #2739, opened 2026-08-14
fix(python): avoid false edges for pytest decorators

### LINKED ISSUE #2732: Python: third-party decorators (e.g. @pytest.fixture) not in _PYTHON_DECORATOR_NOISE create spurious god-node edges when a corpus has a same-named local function
## Summary

`_PYTHON_DECORATOR_NOISE` (extractors/engine.py:74) excludes common **stdlib** decorators (`property`, `dataclass`, `staticmethod`, etc.) from decorator-reference-edge generation, with this rationale in the comment above it:

> Builtin/stdlib decorators (@property, @dataclass, @functools.wraps, …) are ambient vocabulary, not corpus symbols: emitting decorator edges for them fabricates sourceless stub nodes on nearly every class-heavy file, and the unique-function rewire can collapse them onto an unrelated local definition (a corpus defining its own `def wraps(...)` gets a false decorator edge).

`pytest.fixture` is not in that noise list, and it hits exactly the failure mode the comment describes — just from a third-party decorator instead of a stdlib one.

## Repro

Corpus: a Python test suite using `@pytest.fixture` extensively across ~250 test files, one of which happens to define a **local, non-decorator function literally named `fixture()`** (`workspace/tests/test_eodhd_readiness_check.py:22`, unrelated to pytest — just a locally-named helper).

Result: that one `fixture()` node ends up with 44 edges (the #1 "God Node" in the report) — 42 of them `relation: references, context: decorator, confidence: EXTRACTED` — pointing to ~30 completely unrelated local pytest fixtures in other files (`env()`, `db()`, `ob()`, `cb()`, `repo()`, `archive()`, `qt()`, `ns()`, `mod()`, `gs()`, `shape()`, `close()`, `keepalive()`, `caldb()`, etc.). None of these functions call, import, or otherwise relate to each other — they're independent per-file pytest fixtures. The "unique-function rewire" for the `@pytest.fixture` decorator name is binding to that one same-named local function instead of correctly recognizing `pytest.fixture` as an external symbol.

Net effect: a completely spurious "god node" / architectural hub shows up in `GRAPH_REPORT.md`'s God Nodes and (via BFS from it) Suggested Questions sections, pointing the user at a non-existent coupling.

## Suggested fix

Add common third-party test-framework decorators to `_PYTHON_DECORATOR_NOISE` (or, more robustly, only emit a decorator reference edge when the decorator name resolves to a definition actually imported in that file — i.e. respect import scope instead of doing a same-bare-name global lookup). At minimum: `fixture`, `mark`, `parametrize`, `pytest.fixture` (pytest), plus maybe `click.command`, `app.route` / `app.get` / `app.post` (Flask/FastAPI) as other common third-party decorators likely to collide with locally-defined function names in real corpora.

## Environment

- graphifyy 0.9.41
- Corpus: ~700-file mixed Python/docs personal project, code-only AST pass (no LLM involved in this specific extraction — `confidence: EXTRACTED`, purely deterministic)

aayush598:
### Problem

Python decorator extraction currently resolves decorator names using the bare symbol name. This causes third-party pytest decorators such as `@pytest.fixture` and `@pytest.mark.parametrize` to be treated as references to same-named functions in the corpus.

When a project contains a local `fixture()` or `parametrize()` function, repeated pytest decorators can therefore collapse onto that local node and create spurious `references` edges, producing misleading high-degree/"God Node" results.

Fixes #2732.

### Fix

* Track pytest bindings at module scope in source order.
* Recognize `import pytest` and aliased imports such as `import pytest as pt`.
* Recognize decorator-capable direct imports such as `from pytest import fixture` and `from pytest import mark`.
* Suppress known pytest decorators only when the decorator is actually bound to pytest at that location.
* Preserve normal decorator edges for corpus-owned decorators such as bare `@fixture`.
* Respect later rebinding and `del` statements so shadowed pytest names are not incorrectly suppressed.
* Handle `pytest.mark.*` decorators without fabricating nodes for marker names.
* Keep the existing bare-name decorator noise handling unchanged.

### Tests

Adds regression coverage for:

* `@pytest.fixture`
* `@pytest.mark.parametrize`
* aliased pytest imports
* direct pytest decorator imports and aliases
* local decorator shadowing
* source-order rebinding
* `with`/`for`/walrus rebinding
* `del` unbinding
* nested functions and class methods
* stacked pytest and custom decorators
* unimported `@pytest.fixture` remaining a normal corpus reference

Also updates the changelog for the fix.

### Result

Pytest decorators are treated as external test-framework vocabulary when their binding proves they refer to pytest, preventing false cross-file decorator edges and spurious God Nodes while preserving legitimate project-owned decorators.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -223,6 +223,7 @@ Full release notes with details on each version: [GitHub Releases](https://githu
 
 ## 0.9.42 (2026-08-13)
 
+- Fix: avoid false decorator edges for pytest decorators when a corpus contains same-named local functions (#2732, thanks @aayush598).
 - Fix: a JS/TS `for...of` / `for...in` loop binding is now shadowed, so passing it as a call argument no longer fabricates an `indirect_call` edge to an unrelated same-named callable (#2685, thanks @ousamabenyounes); completes the loop/closure/catch shadow family (#2568/#2569/#2517).
 - Fix: graph provenance (`built_at_commit`) is stamped from the analysed repository rather than the shell's working directory, so `graphify extract` run from elsewhere records the target's commit, not the caller's (#2534 family; #2699, thanks @C0KERNEL).
 - Fix: `affected` resolves a seed passed as a `./`-relative path (or an absolute path when run from the repo root) instead of silently returning nothing (#2707, thanks @phudayyy). Note: an absolute-path seed still requires the working directory to be the analysed repo root.
--- a/graphify/extractors/engine.py
+++ b/graphify/extractors/engine.py
@@ -4,10 +4,15 @@ from __future__ import annotations
 import hashlib
 import importlib
 import json
+from dataclasses import dataclass
 from graphify.extractors.base import _LANGUAGE_BUILTIN_GLOBALS, _file_stem, _make_id, _read_text
 from graphify.ids import normalize_id
 from graphify.extractors.models import LanguageConfig
-from graphify.extractors.resolution import _resolve_js_import_target
+from graphify.extractors.resolution import (
+    _python_import_from_module,
+    _python_imported_names,
+    _resolve_js_import_target,
+)
 from graphify.security import sanitize_metadata
 from pathlib import Path
 
@@ -72,6 +77,11 @@ _PYTHON_ANNOTATION_NOISE = frozenset({
 # unique-function rewire can collapse them onto an unrelated local definition
 # (a corpus defining its own `def wraps(...)` gets a false decorator edge).
 # Same name-based tradeoff as `patch`/`Mock` in _PYTHON_ANNOTATION_NOISE.
+# Third-party decorators hit the same failure mode (#2732) but cannot be folded
+# into this bare-name set: `_python_decorator_name` reduces `@pytest.fixture`
+# to the tail `fixture`, a generic name a corpus may use for its OWN decorator.
+# Pytest decorators are suppressed instead by qualified path / binding in
+# `_python_is_pytest_decorator`.
 _PYTHON_DECORATOR_NOISE = frozenset({
     "property", "staticmethod", "classmethod", "abstractmethod",
     "abstractproperty", "cached_property", "wraps", "lru_cache", "cache",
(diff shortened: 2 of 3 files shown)


aayush598:
Thanks for the review. Both "Worth a look" findings are addressed; the branch was also rebased onto current `v8` and squashed into a single clean commit (`71bb212`, was `d2af536` + an update-merge).

**1. "Decorator tail extraction preserves whitespace after dotted attributes" — not reproducible; no code change.**
`_python_decorator_path` / `_python_decorator_name` read the tree-sitter node extent only. An `attribute` node's extent ends exactly at its last identifier — trailing whitespace (or a comment) after it is a sibling node, never included. Verified byte-level: `@pytest.fixture   ` → attribute extent is exactly `pytest.fixture` (ends at the final `e`); `@app.route( "/x" )` → `call`'s function field is `app.route` (parens/space outside). No path by which trailing whitespace can reach `deco_path`, so classification never sees it. If you'd like, I can add a unit test asserting the trimmed path as proof in CI.

**2. "`pytest_symbol` binding for `mark` treated as decorator … bare `@mark` returns `True`" — valid; fixed.**
Root cause: `from pytest import mark` recorded the same `pytest_symbol` kind as the decorator roots (`fixture`/`hookimpl`/`hookspec`), and the classifier short-circuited that kind to `True` — so a bare `@mark` was incorrectly suppressed even though pytest's `mark` (`MarkGenerator`) is not callable.

Now `mark` is recorded as a distinct `pytest_mark` kind. The classifier treats it as pytest vocabulary only when used qualified — `@mark.x`, including the aliased form `@m.x` from `from pytest import mark as m` — while a bare `@mark` keeps its decorator edge.

Regression tests added:
- `test_pytest_mark_bare_is_not_a_decorator` — bare `@mark` edge preserved.
- `test_pytest_mark_alias_qualified_is_suppressed` — `@m.parametrize` still suppressed.

Prior behavior is unchanged: 31/31 decorator tests pass and the full suite shows no new failures (the profile of pre-existing environmental failures is identical).

PR #2741, opened 2026-08-14
Fix #2738: Prevent all-punctuation identifiers from collapsing in make_id

### LINKED ISSUE #2738: Dart private named constructor (`Foo._()`) collapses to an empty entity in make_id, colliding with the file node and disabling _is_file_node
## Summary

`make_id(stem, entity)` strips non-alphanumeric characters from `entity`. A Dart **private named constructor** (`class Foo { const Foo._(); }`) has the entity name `_`, which normalizes to the empty string — so the symbol's ID collapses to the bare file stem, which is the ID that file's own node already owns.

On build the two nodes merge and the symbol's `_` label overwrites the filename label.

That second-order effect is the damaging one. `analyze._is_file_node()` identifies file-level hubs by **comparing the label against the source filename**:

```python
source_file = attrs.get("source_file", "")
if source_file:
    from graphify.build import _is_file_node_label
    if _is_file_node_label(label, source_file):
        return True
```

Once the label is `_` instead of `foo.dart`, the node is no longer recognised as a file node — so it leaks straight into `god_nodes()`, `surprising_connections()` and knowledge-gap reporting, which are exactly the consumers the file-node filter exists to protect. `god_nodes()`' own docstring says file hubs "accumulate import/contains edges mechanically and don't represent meaningful architectural abstractions", and that intent is silently defeated.

## Reproduction

graphify **0.9.42**, Python 3.13, macOS.

`lib/widget.dart` — 5 lines:

```dart
class AvatarCard {
  const AvatarCard._({required this.size});

  factory AvatarCard.small() => const AvatarCard._(size: 1.0);

  final double size;
}
```

```python
from pathlib import Path

r = extract([Path('lib/widget.dart')], cache_root=Path('.'))
for n in r['nodes']:
    print(f"label={n['label']!r:16s} id={n['id']}")
```

```
label='widget.dart'    id=lib_widget
label='AvatarCard'     id=lib_widget_avatarcard
label='size'           id=lib_widget_size
label='_'              id=lib_widget      <-- collides with the file node
label='small'          id=lib_widget_small
```

The collapse is directly visible in `make_id`:

```python
>>> from graphify.extractors.base import _make_id
>>> _make_id("lib/widget", "_")
'lib_widget'
>>> _make_id("lib/widget")          # the file node
'lib_widget'
```

`__` collapses identically. And after `build_from_json`:

```python
>>> G.nodes['lib_widget']['label']
'_'
>>> from graphify.analyze import _is_file_node
>>> _is_file_node(G, 'lib_widget')
False        # expected True
```

## Impact observed on a real corpus

A ~1,050-file Dart/Flutter codebase, `lib/` only, 10,152 nodes / 17,941 edges:

- **246 nodes** ended up labelled `_`.
- The **#1 god node** was a generated DI config file at **409 edges**, labelled `_`. The **#3** was another file node at 43 edges. Both are precisely what `_is_file_node` is meant to exclude; with correct labels neither appears in the ranking at all.
- The two top entries of the report's *Suggested Questions* section were also built on these nodes ("Why does `_` connect ... to \<80 communities\>"), so the highest-signal part of the generated report was pointed at an artifact.
- **236 of the 246** had no sibling node sharing the ID, so the collapsed symbol is the sole occupant of the file-stem ID — for those files the filename label is lost outright rather than merely contested.

Since `/graphify`'s report presents the God Nodes list to the user as "your core abstractions", this reads as a confident, wrong answer about the codebase's architecture rather than as a visible glitch.

## Suggested fix

**Primary — keep the symbol's ID distinct.** In `make_id`, when a non-empty `entity` was supplied but normalizes to empty, fall back to a stable placeholder rather than silently dropping the segment, so the symbol never lands on the file-stem ID:

```python
# entity was given but is all-punctuation (Dart `Foo._()`, a bare `_`);
# without this the ID collapses onto the file node and overwrites its label.
if raw_entity and not normalized_entity:
    normalized_entity = "underscore"   # or a short stable hash of raw_entity
```

**Secondary — make the merge non-destructive.** When `build` merges two nodes sharing an ID, prefer the label that matches the source filename over one that does not. That keeps a single bad label from disabling `_is_file_node` even if some other construct collapses in future.

## Not Dart-specific

A bare `_` is idiomatic in Rust, Go and Python too, and any all-punctuation identifier hits the same empty-entity path. I only verified the Dart case, but the collapse is in the shared `make_id`, so it is worth checking whether the other extractors emit nodes for such identifiers.

## Workaround

For anyone hitting this before a fix lands — relabel rather than filter, since the edges are real. After merging the extraction and before `build_from_json`:

```python
for n in nodes:
    if str(n.get('label', '')).strip() == '_' and n.get('source_file'):
        n['label'] = Path(n['source_file']).name
```

Then dedupe by ID. On the corpus above this restored all 246 labels and cost 0 edges; god nodes became the real abstractions, and the file node kept its 409 edges while correctly dropping out of the abstraction ranking. Filtering the `_` nodes out instead is worse: it discards 862 real edges and fragmented that graph from 395 to 641 communities.

shard-c6:
Fixes #2738

## Summary

This PR addresses the issue where private, all-punctuation identifiers (like Dart's `const Foo._()`) collapse during ID normalization and overwrite their parent file node's label, which subsequently breaks `_is_file_node()` detection and causes file nodes to be incorrectly classified as god nodes.

## Changes
1. **Primary Fix**: Updated `make_id()` in `graphify/ids.py`. When an entity is given but normalizes to an empty string (i.e. it only contains punctuation like `_`), `make_id()` now explicitly falls back to emitting `"underscore"` instead of silently collapsing. This prevents the symbol's ID from colliding with the file's ID.
2. **Secondary Fix**: Updated node merging logic in `build_from_json()` inside `graphify/build.py`. When `G.add_node()` merges attributes for an existing node, it now preferentially keeps the label that matches the source filename instead of allowing arbitrary overlapping symbols to overwrite it.

### DIFF
--- a/graphify/build.py
+++ b/graphify/build.py
@@ -949,7 +949,19 @@ def build_from_json(extraction: dict, *, directed: bool = False, root: str | Pat
                 continue
             if "source_file" in node:
                 node["source_file"] = _norm_source_file(node["source_file"], _root)
-        G.add_node(node["id"], **{k: v for k, v in node.items() if k != "id"})
+        
+        node_id = node["id"]
+        attrs = {k: v for k, v in node.items() if k != "id"}
+        if node_id in G:
+            existing_label = G.nodes[node_id].get("label")
+            new_label = attrs.get("label")
+            if existing_label and new_label and existing_label != new_label:
+                source_file = G.nodes[node_id].get("source_file") or attrs.get("source_file")
+                if source_file:
+                    if _is_file_node_label(existing_label, source_file) and not _is_file_node_label(new_label, source_file):
+                        del attrs["label"]
+        
+        G.add_node(node_id, **attrs)
     node_set = set(G.nodes())
 
     # #1145 (extended): merge LLM ghost-duplicate nodes into AST canonical nodes.
--- a/graphify/ids.py
+++ b/graphify/ids.py
@@ -65,4 +65,12 @@ def make_id(*parts: str) -> str:
     part) and then run through :func:`normalize_id`, so the result is identical to
     what the builder produces from the joined string.
     """
-    return normalize_id("_".join(p.strip("_.") for p in parts if p))
+    normalized_parts = []
+    for p in parts:
+        if not p:
+            continue
+        norm_p = normalize_id(p)
+        if not norm_p:
+            norm_p = "underscore"
+        normalized_parts.append(norm_p)
+    return normalize_id("_".join(normalized_parts))


imagineers-tyler:
The direction here is right, and it fixes a case that is not Dart-specific — on 0.9.48 a JS `const _ = require('lodash')` collapses onto the file node too (`id=scrape`, label `_` overwriting `scrape.js`).

One thing to check before this lands: the `"underscore"` sentinel is a fixed word, so every name that normalizes to empty lands on the same id.

```
make_id('scrape', '_')          = scrape_underscore
make_id('scrape', '__')         = scrape_underscore
make_id('scrape', '$')          = scrape_underscore
make_id('scrape', 'underscore') = scrape_underscore
```

This prevents the parent/file collision but turns it into a potential sibling collision whenever another spelling maps to the same sentinel — including a genuine identifier spelled `underscore`. `_` (lodash) and `$` (cheerio/jQuery) in one file is an idiomatic Node pairing:

```js
const _ = require('lodash');
const $ = require('cheerio');
```

In an isolated `$`-only file (`only_dollar.js`), extraction emits the node (`id=only_dollar_underscore`, label `$`). With `_` and `$` together, only one node survives — `add_node` is first-wins (`engine.py:2904`), so `$` disappears. That happens during extraction, before `build_from_json` runs, so the `build.py` hunk in this PR does not cover it.

That is the failure mode tracked in #2810 (two distinct non-empty symbols normalizing to the same sibling id), which is a different trigger from #2738 (a non-empty entity normalizing to an *empty* segment and collapsing onto its parent/file id). The two are not duplicates.

Encoding the original raw spelling would distinguish `_`, `$`, and `__` from one another instead of mapping all of them to one fixed word; a sufficiently long digest with explicit domain separation gets there too, though only up to collision probability rather than by construction. Without also escaping ordinary identifiers, however, no segment is truly reserved: `normalize_id` emits only characters that a name graphify ingests can contain. The practical goal for a localized fix is therefore distinction among the empty-normalizing spellings plus negligible collision risk with ordinary identifiers, not absolute collision-freedom.

Happy to run the A/B against #2810's reproductions if useful.

PR #2745, opened 2026-08-15
fix(js): extract methods assigned to factory object APIs

### LINKED ISSUE #2524: JS extractor misses property-assigned functions (api.foo = function) — factory/DI modules become near-invisible
**Version:** graphifyy 0.9.34 (pip install in a venv), Python 3.14, macOS (arm64)

**Summary**

The JavaScript extractor captures top-level `function` declarations but not function expressions assigned to object properties. Modules written in the factory / dependency-injection style — very common in Node — end up nearly empty in the graph, so queries against their symbols return "No matching nodes found."

**Repro**

```js
// visible.js — extracted fine
function topLevel() { return 1; }

// factory.js — the dominant pattern in our codebase
function createApi(deps) {
  const api = {};
  api.sourceClips = async function sourceClips(topic) {   // ← not extracted
    return deps.fetch(topic);
  };
  api.renderVideo = function renderVideo(clips) {         // ← not extracted
    return clips.length;
  };
  return api;
}
module.exports = { createApi };
```

`graphify update .` on a real production backend written this way:

- a ~1,400-line pipeline module whose entire surface is `api.foo = function foo(...)` inside one factory → **8 nodes** (the imports, the factory shell, and two top-level helpers); none of the ~100 methods, no call edges between them
- a sibling ~3,000-line module written with top-level `function` declarations → 111 nodes, edges look right

So the graph quality tracks the declaration style, and `query "who calls sourceClips"` / `affected "sourceClips"` both no-hit even though the call sites exist.

**Expected**

Named function expressions assigned to object properties (`obj.name = function name() {}`, `obj.name = async function () {}`, arrow equivalents) inside a factory captured as nodes — ideally namespaced like the existing `module_symbol` ids — with call edges resolved at least within the module.

**Why it matters**

The factory-with-injected-deps style is what testable Node codebases without classes tend to look like (it's why ours is written that way). Right now the token-saving story ("query the graph instead of reading the file") inverts for exactly those codebases: the better-factored the module, the less the graph sees.

Happy to test a branch against our codebase and report node/edge counts.

rajanpanth:
## Summary
- preserve callable members assigned to an object literal inside a JavaScript/TypeScript factory
- model the local API object beneath its factory and attach assigned functions as methods
- cover named async/function-expression assignments and their intra-factory call edge

## Verification
- `uv run pytest tests/test_extract.py -q`
- `uv run ruff check graphify/extractors/engine.py tests/test_extract.py`
- `uv run graphify update .`

Fixes #2524

### DIFF
--- a/graphify/extractors/engine.py
+++ b/graphify/extractors/engine.py
@@ -1966,9 +1966,11 @@ def _js_member_assignment_target(left, source: bytes):
       module.exports.foo = fn  → ("exports",   None,  "foo")
       Foo.prototype.bar = fn   → ("prototype", "Foo", "bar")
 
-    Any other shape (an arbitrary `obj.x = fn`) returns None and is skipped —
-    capturing those would reintroduce the bare-named / phantom-god-node class
-    of bug the module-level scope guard (#1077) exists to prevent.
+    An arbitrary identifier receiver is returned as ``("object", name, member)``.
+    It is only materialized after the caller proves that the identifier is a
+    direct object-literal binding in the enclosing function. Keeping that scope
+    check at the caller avoids the bare-named / phantom-god-node failure mode
+    that the module-level guard (#1077) prevents.
     """
     if left is None or left.type != "member_expression":
         return None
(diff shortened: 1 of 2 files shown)


safishamsi:
Shipped in v0.9.47 via authorship-preserving cherry-pick. On the way in I deduped the factory contains edge (it was being emitted once per method) and added arrow-assignment and bare-receiver tests. This fully resolves #2524. Thanks @rajanpanth! Release: https://github.com/Graphify-Labs/graphify/releases/tag/v0.9.47

Issue #2698, opened 2026-08-13
Classify .gs, .css and dotfile configs instead of skipping them


Domuspi:
# Feature request: classify `.gs`, `.css` and dotfile configs instead of skipping them

**Repo:** https://github.com/Graphify-Labs/graphify
**Version:** graphifyy 0.9.41

## Summary

`graphify extract` silently drops several file types as *"not classified (no supported
extension or shebang)"*. Two of them hold real, graph-worthy content, and one of them is
often the single most security-relevant file in a web project.

Full (non-`--code-only`) extraction on a small PHP/JS web project printed:

```
[graphify extract] 7 file(s) not classified (no supported extension or shebang), skipped:
  .gitattributes, .gitignore, .htaccess, formatar-planilha.gs,
  planilha-conferencia.gs, planilha-custos-backup-2026-08-10.tsv (+1 more)
```

Switching between `--code-only` and full semantic extraction makes no difference — these
are dropped before classification in both modes.

### 1. `.gs` — Google Apps Script (highest impact)

`.gs` **is JavaScript**. Google Apps Script is the automation layer behind Google Sheets,
Docs and Forms, so `.gs` files routinely contain the business logic that drives a
spreadsheet-backed workflow.

Measured in one project:

| File | Size | Lines | `function` declarations |
|---|---|---|---|
| `docs/formatar-planilha.gs` | 9,177 B | 239 | 13 |
| `docs/planilha-conferencia.gs` | 11,432 B | 272 | 8 |

That is **21 functions across 511 lines of JavaScript** that produce zero nodes. Symbols
like `formatarPlanilha`, `garantirRotulos_`, `estilizarCabecalhos_` never enter the graph,
so the spreadsheet automation is invisible to `query`, `path` and `affected`.

The existing JavaScript tree-sitter grammar should parse `.gs` as-is. This looks like a
one-line extension-map addition rather than new parser work.

### 2. `.htaccess` — Apache config

In shared-hosting deployments `.htaccess` is the access-control boundary. In the project
above it is a 118-line **allowlist** that decides which paths are reachable from the
public internet — including the rule that keeps `graphify-out/` itself unreachable.

An architecture graph that omits it cannot answer "what is actually exposed?". Even
treating it as a *document* (no AST, just semantic extraction) would be a large
improvement over dropping it.

Note it is also a dotfile with no extension, so extension-based classification misses it
by construction.

### 3. `.css` — stylesheets

[...]

PR #2747, opened 2026-08-15
fix(detect): extract Google Apps Script files

rajanpanth:
## Summary
- classify Google Apps Script (\.gs\) files as code
- route them through the existing JavaScript extractor and JS symbol-resolution family
- bypass stale AST caches for .gs exactly as for other JavaScript variants

This addresses the Google Apps Script portion of #2698. CSS and extensionless-config classification remain separate work.

## Validation
- uv run pytest tests/test_detect.py::test_classify_google_apps_script tests/test_extract.py::test_dispatch_includes_sh_json_and_google_apps_script tests/test_extract.py::test_extract_google_apps_script_as_javascript -q
- uv run ruff check graphify/detect.py graphify/extract.py graphify/extractors/models.py tests/test_detect.py tests/test_extract.py
- uv run graphify update .

Note: the combined detect/extract suite has two pre-existing Windows Unicode normalization failures; the remaining 425 tests passed.


rajanpanth:
I have rebased this branch onto 8 to resolve the merge conflict in graphify/detect.py (which arose from other languages being added to CODE_EXTENSIONS). Everything is clean and ready to go!

rajanpanth:
Rebased onto `v8` (149 commits) and mergeable again.

The only conflict was the `CODE_EXTENSIONS` set in `detect.py`, which had gained entries upstream. I took `v8`'s current line and inserted `.gs` into it rather than reinstating my older copy, so nothing added since is lost.

I checked first that this had not already been solved upstream, since that has happened before: `.gs` is absent from `detect.py`, `extract.py` and `extractors/models.py` on `v8` today, so the gap is still there.

Verified after the rebase: `tests/test_detect.py` and `tests/test_extract.py` are 463 passed, 5 skipped, and the 23 tests matching the Apps Script cases all pass.

Diff against `v8` is 23 insertions and 4 deletions across the three source files plus the two test files.

PR #2749, opened 2026-08-15
fix(js): retain lazy CommonJS require dependencies

### LINKED ISSUE #2700: JS: CommonJS `require()` inside a function body produces no edge at all — cycle-breaking lazy requires are invisible (0.9.37)
**Version:** graphify 0.9.37 (also observed on 0.9.32)
**Platform:** Windows 11, Python 3.14.6, `uv tool install "graphifyy[mcp]"`

## Summary

A CommonJS `require()` written **directly inside a function body** (nesting depth 1) produces **no edge at all** — not a dangling one, not a low-confidence one. The importing file ends up with zero connection to the module it requires. The identical `require()` at module scope resolves fine, so a codebase can look well-connected while every deferred require is invisible.

This matters disproportionately because **lazy `require()` is the standard CommonJS idiom for breaking circular dependencies**. The edges most likely to be dropped are exactly the ones describing a project's most tangled relationships.

There is an irony worth flagging: graphify labels edges `EXTRACTED` (explicit in source) vs `INFERRED` (resolved heuristically). A literal `require('./target')` string is about as `EXTRACTED` as evidence gets — yet in the repro below the lazy-require file keeps an `INFERRED` 0.5-confidence edge while its four `EXTRACTED` 1.0-confidence edges vanish.

## Reproduction

Three files, no config, no aliases, no bundler. `git init` + `graphify extract . --code-only`.

```js
// target.js
function helper() { return 42; }
module.exports = { helper };
```

```js
// top_level.js  — require at module scope (CONTROL)
const { helper } = require('./target');
function useIt() { return helper(); }
module.exports = { useIt };
```

```js
// in_function.js — same require, one level deeper
function useItLazily() {
  const { helper } = require('./target');
  return helper();
}
module.exports = { useItLazily };
```

### Result — 7 nodes, 7 edges

`top_level.js` → four `EXTRACTED` edges, confidence 1.0:

```json
{"relation":"imports_from","confidence":"EXTRACTED","confidence_score":1,"source":"top_level","target":"target","source_location":"L1"}
{"relation":"imports","confidence":"EXTRACTED","confidence_score":1,"source":"top_level","target":"target_helper","source_location":"L1"}
{"relation":"calls","confidence":"EXTRACTED","confidence_score":1,"source":"top_level_useit","target":"target_helper","source_location":"L2"}
{"relation":"contains","confidence":"EXTRACTED","confidence_score":1,"source":"top_level","target":"top_level_helper","source_location":"L1"}
```

`in_function.js` → one `INFERRED` edge, confidence 0.5, pointing at its own function:

```json
{"relation":"indirect_call","confidence":"INFERRED","confidence_score":0.5,"source":"in_function","target":"in_function_useitlazily","source_location":"L5"}
```

| file | `require` placement | edge to `target` |
|---|---|---|
| `top_level.js` | module scope | ✅ `imports_from` + `imports` |
| **`in_function.js`** | **depth-1 function body** | ❌ **none** |

Also note `top_level_helper` exists as a node (the destructured binding) while `in_function` has no equivalent — the binding is not extracted either.

## Why this is distinct from existing issues

- **#2575** (`await import('…')` inside a nested function produces no edge) — that report states explicitly that a dynamic import *at the top level of a function* resolves fine, and that **nesting** (function-inside-function, depth ≥2) is the trigger. Here the failure is at **depth 1** with **CommonJS `require()`**, not ESM `import()`. Possibly a shared root cause in scope walking, but a different syntax node and a different depth threshold, so it will not be fixed or verified by the same test.
- **#2584** (`affected` cannot traverse a dynamic `import()` made inside a function) — there the *edge exists* and traversal is the problem. Here **no edge is ever created**.
- **#2280 / #2556** (bare module-name endpoints that dangle and are dropped at build) — those produce an edge with an unmatchable endpoint. Here nothing is emitted at all; there is no dangling endpoint to resolve.
- **#2153** (`jsconfig` `baseUrl`, non-relative specifiers) — the repro above uses a plain relative `./target` specifier with no config of any kind.

## Real-world impact

Found while indexing [BourbonDog/amicus](https://github.com/BourbonDog/amicus) (~841 JS files). Benchmarking `graphify affected` against plain `grep` for one module's dependents:

- graphify found **3 of 6** real `require()`-backed dependents; `grep` found all 6 in under a second
- the 3 missed (`headless.js`, `mcp-server.js`, `sidecar/session-utils.js`) all require the target from inside a function body
- `--depth 4` did not recover them, confirming this is missing edges rather than traversal depth

That codebase has **9 circular-dependency chains that are deliberately broken by lazy `require()`**, each with a source comment saying so (`// lazy: avoids a load-time cycle`). Every one of those cycle-breaking edges is absent from the graph — so the graph presents an acyclic module structure that does not match the code.

The practical consequence is that `affected` under-reports blast radius without signalling reduced confidence, which is the dangerous direction for a "what breaks if I change this?" tool.

## Suggested direction

Walk `call_expression` nodes with callee `require` at any scope depth rather than only at module scope, mirroring whatever the ESM path does for static imports. Two things worth covering in tests, since the amicus corpus exercises both:

- `require()` inside `try { }` within a function (very common for optional/fallback loads)
- multiple `require()` calls of the same module in different functions of one file (should dedupe to one edge, not zero)

Happy to test a patch against the amicus corpus — it has ~50 lazy requires across 9 documented cycles, so it is a decent real-world fixture for this.

rajanpanth:
## Summary
- detect CommonJS \
equire()\ declarations while traversing JavaScript and TypeScript function bodies
- emit the resolved \imports_from\ and symbol import edges from the enclosing callable
- preserve existing file-level edges for module-scope requires

Fixes #2700.

## Validation
- uv run pytest tests/test_extract.py -q
- uv run pytest tests/test_extract.py::test_extract_js_function_scoped_require_emits_import_edge tests/test_extract.py::test_extract_js_destructured_require_imports_from tests/test_extract.py::test_extract_js_destructured_require_named_symbols tests/test_languages.py -q
- uv run ruff check graphify/extractors/engine.py tests/test_extract.py
- uv run graphify update .

### DIFF
--- a/graphify/extractors/engine.py
+++ b/graphify/extractors/engine.py
@@ -1786,7 +1786,7 @@ def _find_require_call(value_node):
         return _find_require_call(obj)
     return None
 
-def _require_imports_js(node, source: bytes, file_nid: str, stem: str, edges: list, str_path: str) -> bool:
+def _require_imports_js(node, source: bytes, importer_nid: str, stem: str, edges: list, str_path: str) -> bool:
     """Detect CommonJS require imports inside lexical_declaration / variable_declaration.
 
     Handles three patterns:
@@ -1825,7 +1825,7 @@ def _require_imports_js(node, source: bytes, file_nid: str, stem: str, edges: li
         tgt_nid, resolved_path = resolved
         line = node.start_point[0] + 1
         edge = {
-            "source": file_nid,
+            "source": importer_nid,
             "target": tgt_nid,
             "relation": "imports_from",
             "context": "import",
@@ -1862,7 +1862,7 @@ def _require_imports_js(node, source: bytes, file_nid: str, stem: str, edges: li
         if target_stem is not None:
             for sym in sym_names:
                 edges.append({
-                    "source": file_nid,
+                    "source": importer_nid,
                     "target": _make_id(target_stem, sym),
                     "relation": "imports",
                     "context": "import",
@@ -4657,6 +4657,14 @@ def _extract_generic(
                     for child in node.children:
                         walk_calls(child, caller_nid, receiver_types, closure_locals)
             return
+
+        # CommonJS imports are valid at any lexical depth.  The module-level
+        # pass records top-level require() declarations; this pass owns function
+        # bodies, so detect lazy/cycle-breaking requires here and attribute the
+        # dependency to the enclosing callable rather than silently dropping it.
+        if (config.ts_module in ("tree_sitter_javascript", "tree_sitter_typescript")
+                and node.type in ("lexical_declaration", "variable_declaration")):
+            _require_imports_js(node, source, caller_nid, stem, edges, str_path)
 
         if node.type in config.call_types:
             # JS/TS dynamic imports: await import('./foo.js')
--- a/tests/test_extract.py
+++ b/tests/test_extract.py
@@ -776,6 +776,31 @@ def test_extract_js_member_require_emits_property_symbol():
     assert _make_id(helpers_stem, "helperFn") in sym_targets
 
 
+def test_extract_js_function_scoped_require_emits_import_edge(tmp_path):
+    """Lazy CommonJS requires belong to their enclosing function, not nowhere."""
+    target = tmp_path / "target.js"
+    target.write_text("exports.helper = () => 42;\n", encoding="utf-8")
+    caller = tmp_path / "lazy.js"
+    caller.write_text(
+        "function useItLazily() {\n"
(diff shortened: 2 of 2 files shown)


safishamsi:
Landed in v0.9.44, just published to PyPI. I cherry-picked this onto `v8` with your authorship preserved in the commit, so both the change and the credit are in the history. Thanks @rajanpanth for the fix. Closing since it is now released.

PR #2750, opened 2026-08-15
fix(affected): resolve absolute-path seed via graph-derived root (#2706)

### LINKED ISSUE #2706: `affected` silently returns nothing when the path is spelled differently (./ or absolute)

ousamabenyounes:
Fixes #2706.

`affected` resolved an absolute-path seed only when the working directory happened to be the analysed repo root: `resolve_seed` anchored an absolute path to `Path.cwd()`, so from anywhere else `relative_to(cwd)` raised and the query was compared verbatim — matching no stored (repo-relative) `source_file`. The result was an empty blast radius with exit 0, indistinguishable from a genuine zero or a typo. An absolute path is exactly what an editor integration or a script passes, and neither can guarantee cwd is the repo root.

This is the follow-up the maintainer noted when the `./`-relative case shipped in #2707:

> an absolute-path seed still requires the working directory to be the analysed repo root — a follow-up could derive the root from the graph location.

## The change

- Derive the analysed repo root from the graph's own location — the graph is written to `<root>/graphify-out/graph.json`, so the root is the output dir's parent (`GRAPHIFY_OUT_NAME`); a graph pointed at directly by `--graph` falls back to its own directory.
- Thread an optional `root` through `_as_repo_relative` → `resolve_seed` → `format_affected`, and anchor an absolute seed to it. `root` defaults to the current directory, so every existing caller and the label/non-path branches are unchanged. An absolute path outside the derived root is still left verbatim — no basename guessing.

Scope is the `affected` seed path only; `explain`/`shortest_path` and the shared scoring helper are untouched.

New test `test_affected_absolute_seed_resolves_via_graph_root_off_cwd` runs the CLI end-to-end with the graph under `<root>/graphify-out/graph.json`, cwd set to a *different* directory, and an absolute seed.

RED — new test on the branch with only the two production files reverted to `v8`:

```
FAILED tests/test_affected_cli.py::test_affected_absolute_seed_resolves_via_graph_root_off_cwd
E   AssertionError: assert 'Affected nodes for Foo' in 'No unique node match for /tmp/.../repo/pkg/foo.py\n'
1 failed
```

GREEN — full `affected` suite with the fix:

```
tests/test_affected_cli.py ......... 17 passed
```

## Full local suite

`uv run --frozen pytest tests/ -q` — 4465 passed, 3 failed, 7 skipped. The 4465 is the pre-fix 4464 plus the new test; the 3 failures are the pre-existing `tests/test_ollama.py::test_detect_backend_*` cases (local ollama env), identical before and after this change. `ruff` and the `tools.skillgen` guards (`--check`, `--audit-coverage`, `--schema-singleton`, `--monolith-roundtrip`, `--always-on-roundtrip`) are green.


safishamsi:

PR #2751, opened 2026-08-15
fix(cpp): recover doctest/Catch2 string-named TEST_CASE nodes (#2594)

### LINKED ISSUE #2594: C++: doctest / Catch2 string-named TEST_CASE fails to parse — every test case missing from the graph (parallel to the existing Spock fallback)
## Summary

`TEST_CASE("name") { ... }` — the standard form in **doctest** and **Catch2** — cannot be parsed by tree-sitter-cpp, so every test case in the file is absent from the graph. The file node survives, `has_error` is set, and (since 0.9.37) the #2551 warning fires, but the test functions themselves are silently gone.

This is the same shape as the Spock case you already handle: a test framework that names its cases with a **string literal** where the grammar expects an identifier. `_is_spock_file` / `_extract_spock_fallback` solve exactly this for `def "feature name"()` in Groovy. Would you be open to the equivalent for C++ (`doctest` / `Catch2`)? I'm happy to send a PR against `v8` if you'll take one — I'd just like to agree on the node-naming rule first.

## Reproduction

`example_test.cpp` (11 lines, nothing but public library syntax):

```cpp
#include <doctest/doctest.h>

static int add(int a, int b) { return a + b; }

TEST_CASE("addition works") {
    CHECK(add(1, 2) == 3);
}

TEST_CASE("subtraction works") {
    CHECK(add(5, -1) == 4);
}
```

```
$ graphify update /tmp/dt
  warning: 1 file(s) had syntax errors and may be partially extracted: example_test.cpp (first error at line 5) (#2551)
[graphify watch] Rebuilt: 2 nodes, 1 edges, 1 communities
```

**Observed** — 2 nodes: the file node and `add()`. Both test cases are missing.
**Expected** — 4 nodes: the two test cases should appear as callable nodes, so that `CHECK(add(...))` gives `test case -> add()` call edges.

Control: rewriting `TEST_CASE("addition works")` to `void addition_works()` (nothing else changed) parses clean and yields all 4 nodes. So the loss is entirely the string-literal macro form, not anything else in the file.

## Which frameworks are affected

Probed each form through `_extract_generic` with the C++ config:

| form | parses | nodes extracted |
|---|---|---|
| doctest `TEST_CASE("adds")` | ❌ error at the macro line | test case missing |
| doctest `SUBCASE("inner")` | ❌ | missing |
| Catch2 `TEST_CASE("adds", "[math]")` | ❌ | missing |
| Catch2 `SECTION("inner")` | ❌ | missing |
| gtest `TEST(Suite, Adds)` | ✅ | parses, but see note |
| gtest `TEST_F` / `TYPED_TEST` | ✅ | parses, but see note |
| Boost `BOOST_AUTO_TEST_CASE(adds)` | ✅ | parses, but see note |

The break is specific to **string-literal test names**. Identifier-named macros (gtest, Boost) do parse — though they currently yield a `TEST()` call node plus loose `Suite` / `Adds` identifier nodes rather than one test node, which is a separate and much smaller nit, not what this issue is about.

## Impact at scale

Two private C/C++ codebases, measured by rewriting `TEST_CASE("x")` → `void x()` in memory and re-running extraction (so this is the recoverable delta, not an estimate):

| codebase | test files | `TEST_CASE` macros | nodes across those files | edges |
|---|---|---|---|---|
| A | 56 | 608 | 914 → 1,506 (**+65%**) | 1,704 → 3,373 (**+98%**) |
| B | 63 | 456 | 353 → 776 (**+120%**) | 663 → 1,391 (**+110%**) |

For codebase A that is +5.6% of all first-party C/C++ nodes and +8.7% of edges. In practice the whole unit-test suite is invisible to the graph, which is a real loss: for a legacy C++ codebase the tests are often the most reliable description of intended behaviour, and they're exactly what you want an assistant to find.

Worth noting for triage: for these codebases the rest of the #2551 warnings turned out to be **harmless**. Most flagged files are macro-heavy but lose almost nothing, because tree-sitter's recovery is local and your node granularity is file/class/function — the enclosing function still extracts. Across ~3,400 C/C++ files, neutralizing the dominant local macro pattern moved first-party node counts by only +0.06%. doctest/Catch2 is the one case where the loss is structural, because the *whole* test function is the thing that fails to parse.

## Root cause

`TEST_CASE("addition works") { ... }` is, to the grammar, an unknown identifier applied to a string literal, followed by a compound statement — not a declaration. tree-sitter-cpp recovers by making the whole construct an ERROR, so no function node is produced. No grammar version fixes this (tree-sitter-cpp 0.23.4 is current); it needs a graphify-side fallback, as with Spock.

## Suggested shape

Mirroring `_is_spock_file` / `_extract_spock_fallback`:

1. Sniff the raw source for `^\s*(TEST_CASE|TEST_CASE_TEMPLATE|SUBCASE|SCENARIO|SECTION|TEST_SUITE)\s*\(\s*"` before deciding the file is a loss.
2. On a hit, keep the tree-sitter file node and its include edges (those survive reliably), and add one callable node per test case extracted by regex — same merge strategy the Spock fallback uses.
3. Node label: the test name string is the useful identity for search (`"addition works"`), but it isn't an identifier. Slugging it (`addition_works`) versus keeping the raw string is the one decision I'd want your call on before writing the PR.

Also happy to keep the scope to doctest+Catch2 only, since those two share the exact same surface syntax.

## Environment

- graphifyy 0.9.38 (`[mcp]`), Python 3.12.3, Linux, AST-only (`graphify update`, no LLM)
- tree-sitter 0.25.2, tree-sitter-cpp 0.23.4

ousamabenyounes:
## Summary

Fix #2594

`TEST_CASE("name") { ... }` — the standard doctest / Catch2 form — puts a string
literal where tree-sitter-cpp expects an identifier, so the whole construct
parses as an `ERROR` node and the test function is silently dropped from the
graph (the file node and neighbouring functions survive, but every test case is
gone).

This mirrors the shape the repo already solves for Groovy Spock
(`def "feature"()` → `_is_spock_file` / `_extract_spock_fallback`). The fix adds
a line-anchored regex fallback that, after the generic C++ pass, appends one
callable node per **top-level** test-declaration macro with a `contains` edge
from the file node:

- Recovered macros: `TEST_CASE`, `TEST_CASE_TEMPLATE`, `SCENARIO`.
- The raw test name is kept as the node label (e.g. `"addition works"`),
  matching how the Spock fallback labels feature methods.
- Additive: the surrounding functions and include edges the tree-sitter pass
  still recovers are left untouched.

### Scope (deliberate, for parity with the Spock fallback)

- `SUBCASE` / `SECTION` are *nested* scopes inside a test body and `TEST_SUITE`
  is a grouping wrapper, not a test function, so they are **not** emitted as
  file-level nodes (that would fabricate wrong-granularity nodes/edges).
- The matcher consumes C-string escapes, so `TEST_CASE("handles \"quoted\"")`
  is captured whole rather than truncated at the escape.
- Matching is raw text and line-anchored, exactly like the Spock fallback; it is
  intentionally not comment/preprocessor aware (a `TEST_CASE` behind a block
  comment or `#if 0` can still surface, just as a commented Spock `def "x"()`
  would). Adding a full C++ lexer here would be out of proportion to the fix.

I kept the scope to node recovery (the structural loss the issue is about);
`test case -> callee` call-edge attribution from test bodies is out of scope, as
it is for the Spock fallback.

## Test verification (RED -> GREEN)

New test `test_cpp_recovers_doctest_string_named_test_cases` on
`tests/fixtures/sample_doctest.cpp`.

RED — on the unmodified `v8` extractor (fallback reverted), the two test cases
are absent:

```
tests/test_languages.py:199: in test_cpp_recovers_doctest_string_named_test_cases
    assert '"addition works"' in labels
E   assert '"addition works"' in ['sample_doctest.cpp', 'add()']
1 failed
```

GREEN — with the fallback, all four nodes are present and each test case is
contained by the file node:

```
NODE COUNT: 4
  'example_test.cpp'      | ...
  'add()'                 | ...
  '"addition works"'      | ...  L5
  '"subtraction works"'   | ...  L9
1 passed
```

## Full local suite

- Command: `uv run --frozen pytest tests/ -q`
- Result: `3 failed, 4465 passed, 7 skipped` — iso-baseline. The 3 reds are
  pre-existing, environment-dependent `test_ollama.py` backend-detection tests
  that fail identically on the unmodified `v8` base (a local ollama binary +
  provider env vars), unrelated to this change.
- `skillgen` guards (`--check` / `--audit-coverage` / `--schema-singleton` /
  `--monolith-roundtrip` / `--always-on-roundtrip`) all pass.


safishamsi:

Issue #1740, opened 2026-07-09
JS/TS extractor silently drops calls inside anonymous arrow callbacks (addEventListener, forEach, etc.)


50auragawd:
## Bug

`walk_calls` in `graphify/extract.py` stops recursion at every `function_boundary_types` node. For JS/TS that set includes `arrow_function`, which is required so module-level named arrows (`const f = () => {}`) — extracted as their own nodes and walked separately via `function_bodies` — aren't double-attributed to the enclosing scope.

But **anonymous** arrow callbacks have no node of their own and no separate body walk, so stopping at them silently drops every call expression inside. Nothing is emitted — not attributed to the enclosing function, not to the module node.

## Repro

`script.js`:

```js
function toggleTask(id) { renderTasks(); }
function removeTask(id) { renderTasks(); }
function renderTasks() {
  tasks.forEach((task) => {
    checkbox.addEventListener('change', () => toggleTask(task.id));
    button.addEventListener('click', () => { removeTask(task.id); });
  });
}
```

Extracted edges before the fix: `toggleTask → renderTasks`, `removeTask → renderTasks` only. The `renderTasks → toggleTask` (L5) and `renderTasks → removeTask` (L6) edges are missing entirely, inverting the apparent call-graph topology: `renderTasks` looks like a pure sink when it actually fans out to the mutators it wires up.

This hits any DOM-heavy vanilla JS codebase (every `addEventListener` + arrow listener) and any callback-style code (`forEach`, `map`, promise `.then(() => ...)` with arrows). Found while dogfooding a vanilla JS app's graph through the Hermes agent integration (#1403 verification): the agent's data-flow trace flagged the listener wiring as absent/inconsistent in the graph.

## Fix

Stop at a function boundary only if that exact AST node is separately walked — i.e. its body is registered in `function_bodies`. Otherwise recurse into the body under the enclosing function's nid:

- Build `registered_body_ids = {id(body) for _, body in function_bodies}` after the extraction pass.
- In `walk_calls`, on hitting a boundary node, look up its body via `_find_body(node, config)`; if unregistered, `walk_calls(body, caller_nid)` instead of returning.

Because the check keys off "is this node walked separately" rather than JS specifics, it also repairs the same silent-drop class in other languages sharing the generic walker (Python nested `def`s, Java anonymous-class methods).

[...]

PR #2752, opened 2026-08-15
fix(js): shadow untracked function-expression bindings from indirect_call args

imagineers-tyler:
## Summary

`function_expression` is missing from the JS and TS `function_boundary_types` sets, so an
**untracked** (inline or nested) `function (…) {…}` expression's own parameters and locals never
enter its subtree shadow set. A bare reference to one of them, passed on as a call argument,
reads as an unresolved by-name reference and fabricates an `indirect_call` edge (INFERRED, 0.8)
to an unrelated same-named callable elsewhere in the corpus.

The two callback forms disagree, which is what hides it:

```js
xs.some(k => sink(k))                     // arrow — shadowed, correct
xs.some(function (k) { return sink(k) })  // function expression — fabricated edge
```

Same family as `catch_clause.parameter` and the single unparenthesised arrow parameter: a
binding form that never reached the shadow set.

Reproduced against `v8 @ 7281f27` (v0.9.43).

## Reproduction

```js
// src/vendor.min.js

// src/a.js
export function run(xs, m){ return xs.some(function (k) { return m.indexOf(k); }); }
```

```python
from pathlib import Path
r = extract([Path('src/vendor.min.js'), Path('src/a.js')], cache_root=Path('.c'), parallel=False)
```

Emits `run() --indirect_call [INFERRED]--> k()`, where `k` is the minified bundle's private
function. `k` in `a.js` is the callback's own parameter. Rewriting the callback as
`xs.some(k => m.indexOf(k))` emits nothing — same binding, opposite result.

## Cause

`walk_calls` gates its untracked-closure handling on `config.function_boundary_types`:

```python
if node.type in config.function_boundary_types:
    ...
    if (config.ts_module in ("tree_sitter_javascript", "tree_sitter_typescript")
            and node.type in _JS_DESCEND_TYPES):
        ...
        closure_locals = extra_locals | _js_local_bound_names(node, source)
```

and that handler already names the type it should receive:

```python
_JS_CLOSURE_TYPES = ("arrow_function", "function_expression")
_JS_DESCEND_TYPES = _JS_CLOSURE_TYPES + ("function_declaration", "generator_function_declaration")
```

But `graphify/extract.py` has, in the JS config and again in the TS config:

```python
function_boundary_types=frozenset({"function_declaration", "generator_function_declaration",
                                   "arrow_function", "method_definition"}),
```

`function_expression` is absent, so the outer gate never opens for one and the fold below is
unreachable. The body is instead walked with the *enclosing* function's `extra_locals`.

### Scope

A top-level `const f = function (…) {}` is tracked through its declarator, gets its own node,
and already had its own shadow-set entry — it never went through this gate and was unaffected.
Measured on unmodified `7281f27`:

```
TOP-LEVEL  const f = function (k)  -> fabricated edge: no
NESTED     const f = function (k)  -> fabricated edge: yes
```

The gap is the untracked inline/nested forms. A boundary control test pins this so the scope
stays legible.

### Why it survived

| | |
|---|---|
| `4744dfe` (#1630) | introduced the descent pair `("arrow_function", "function_expression")` — the type has been named as a closure since then |
| #1740 | anonymous arrow callbacks dropped their calls, because `arrow_function` **is** in the boundary set; fixed by descending instead of stopping |
| `51b7b9f` (#2241) | added the shadow-set fold *inside* that descent — its repro and all six of its tests are arrow-based |
| `09aeb97` | last touched the JS `function_boundary_types` line (generators); `function_expression` was never added |

The fold from `51b7b9f` inherited a gate that admits arrows but not function expressions, and
everything that exercised it used arrows. This PR completes that change.

Conversely, `function_expression` never had #1740's call-dropping problem precisely *because*
it was absent from the set — `walk_calls` recursed into it normally. Adding the type routes it
through the boundary branch instead, which descends explicitly with the same `caller_nid`, so
call attribution is unchanged. That is asserted directly by
`test_call_attribution_inside_the_expression_is_unchanged`, and verified against the baseline:
`run --calls--> helper` is emitted identically before and after.

## Fix

The same one-token addition in the JS and TS configs — two lines in `graphify/extract.py`.
No handler code changes; this only lets the existing path receive the type it already names.
TSX inherits the set from the TS config.

## Tests

`tests/test_indirect_call_function_expression_shadow.py`, following the shape of
`test_indirect_call_arrow_single_param_shadow.py` and
`test_indirect_call_catch_binding_shadow.py`.

Six of the eleven are controls that must pass with **and** without the fix — they exist to catch
over-suppression, per the standard applied on #2517.

| test | before | after |
|---|---|---|
| inline anon `function (k)` callback param | FAIL | pass |
| named `function nm(k)` callback param | FAIL | pass |
| nested `const f = function (k) {…}` | FAIL | pass |
| `function () { const k = 1; … }` local | FAIL | pass |
| TypeScript `function (k: number)` | FAIL | pass |
| *control:* arrow callback still shadows | pass | pass |
| *control:* same-named reference after the expression still resolves | pass | pass |
| *control:* sibling function expression's param does not leak (#2568 shape) | pass | pass |
| *control:* unshadowed callable inside the body still emits | pass | pass |
| *control:* top-level `const f = function (…) {}` was already tracked | pass | pass |
| *control:* call attribution inside the expression is unchanged | pass | pass |

```
without the fix: 5 failed, 6 passed
with the fix:    11 passed
```

The sibling control uses two closures in one initializer, which is the #2568 shape reached
through the function-expression form rather than the arrow form:

```js
export const handler = wrap(
  function (k) { return k; },
  function (pool) { return pool.submit(k); }
);
```

Full suite on this branch: **4177 passed, 183 skipped, 13 failed** — the same 13 fail on
unmodified `7281f27` in this environment, with the failing sets identical in both directions.
They are `test_terraform` ×7 (`tree_sitter_hcl not installed`), `test_ollama_retry_cap` ×4,
`test_install_references::test_built_wheel_ships_the_full_skill_payload`, and
`test_labeling::test_label_communities_batches_when_over_batch_size` (flaky independent of this
change — 3 passes / 2 failures in 5 consecutive runs on the same tree).

## Measured effect

Extracted with and without the fix over the same corpora, diffing the `indirect_call` edge sets.
Minified bundles are excluded: whether `j() -> b()` inside a one-line bundle is real is not a
judgment worth making, and every reference in such a file reports the same line.

| corpus | files | edges before | after | removed | added |
|---|---|---|---|---|---|
| mongoose | 294 | 53 | 47 | 6 | 0 |
| handlebars | 106 | 70 | 56 | 14 | 0 |
| eslint | 414 | 361 | 358 | 3 | 0 |
| happy-dom | 1800 | 4 | 4 | 0 | 0 |
| puppeteer-extra-plugin-stealth | 54 | 7 | 7 | 0 | 0 |
| application corpus | 51 | 31 | 31 | 0 | 0 |
| **total** | **2719** | | | **23** | **0** |

**23 fabricated edges removed, 0 introduced.**

Spot-checking the largest group, handlebars `template() -> blockParams()` (×5): suppressed
because `program: function (i, data, declaredBlockParams, blockParams, depths)` binds
`blockParams` as a parameter. Correct.

Including minified files the raw counts are 43 removed / 3 added; all 3 "added" are in
`handlebars.amd.min.js`, where the same references move from L27 to L28 — a re-attribution
artifact of single-line files, not new edges.

## Interaction with existing behaviour

Re-verified on this branch, all still correct:

- single unparenthesised arrow parameter — shadows, and does not shadow outside the arrow
- `catch (e)` binding — shadows; `catch {}` and post-clause references unaffected
- the `for...of` / `for...in` loop-binding behaviour from `a995921` — unchanged
- the #2568 sibling-closure case, including that issue's own positive control — unchanged

### DIFF
--- a/graphify/extract.py
+++ b/graphify/extract.py
@@ -786,7 +786,14 @@ _JS_CONFIG = LanguageConfig(
     call_accessor_node_types=frozenset({"member_expression"}),
     call_accessor_field="property",
     call_accessor_object_field="object",
-    function_boundary_types=frozenset({"function_declaration", "generator_function_declaration", "arrow_function", "method_definition"}),
+    # `function_expression` belongs here so UNTRACKED inline/nested expressions
+    # reach walk_calls' existing closure handler, which already names the type
+    # (`_JS_CLOSURE_TYPES`). Without it the gate never opens, so such an
+    # expression's parameters and locals never fold into extra_locals for its
+    # subtree and read as by-name references (#2241 family). A top-level
+    # `const f = function (…) {}` is tracked via its declarator and was already
+    # fine; the inline/nested forms are what this covers.
+    function_boundary_types=frozenset({"function_declaration", "generator_function_declaration", "arrow_function", "method_definition", "function_expression"}),
     import_handler=_import_js,
 )
 
@@ -807,7 +814,8 @@ _TS_CONFIG = LanguageConfig(
     call_accessor_node_types=frozenset({"member_expression"}),
     call_accessor_field="property",
     call_accessor_object_field="object",
-    function_boundary_types=frozenset({"function_declaration", "generator_function_declaration", "arrow_function", "method_definition"}),
+    # `function_expression`: see the note on the JS config above.
+    function_boundary_types=frozenset({"function_declaration", "generator_function_declaration", "arrow_function", "method_definition", "function_expression"}),
     import_handler=_import_js,
 )
 
--- a/tests/test_indirect_call_function_expression_shadow.py
+++ b/tests/test_indirect_call_function_expression_shadow.py
@@ -1,0 +1,191 @@
+"""An untracked `function (…) {…}` expression's bindings must shadow indirect_call args.
+
+`walk_calls` gates its untracked-closure handling on
+`config.function_boundary_types`, and that handler already names the type it
+should receive:
+
+    _JS_CLOSURE_TYPES = ("arrow_function", "function_expression")
+
+But `function_expression` was absent from the JS and TS `function_boundary_types`
+sets, so the gate never opened for one. An INLINE or NESTED expression's body was
+walked with the ENCLOSING function's `extra_locals`, and its own parameters and
+locals folded into nothing: a bare reference to one of them, passed on as a call
+argument, read as an unresolved by-name reference and fabricated an
+`indirect_call` edge (INFERRED, 0.8) to an unrelated same-named callable
+elsewhere in the corpus.
+
+Scope: a top-level `const f = function (…) {}` is tracked through its declarator
+and already had its own shadow-set entry, so it was never affected — see
+`test_top_level_const_function_expression_was_already_tracked`. The gap is the
+untracked inline/nested forms.
+
+That made the two callback forms disagree, which is what hid it:
+
+    xs.some(k => sink(k))                  # arrow — shadowed, correct
+    xs.some(function (k) { return sink(k) })  # function expression — fabricated
+
+Minified bundles name nearly every private function with a single letter, and
+`function (err, res) {…}` callbacks are pervasive in older JS, so the two collide
+constantly. Same family as `catch_clause.parameter` and the single unparenthesised
+arrow parameter: a binding form that never reached the shadow set.
+"""
+import os
+from pathlib import Path
+
+from graphify.extract import extract
+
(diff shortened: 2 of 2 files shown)


safishamsi:
Landed in v0.9.44, just published to PyPI, cherry-picked onto `v8` with your authorship preserved. I added a small follow-up so it also covers generator function expressions (`function*(){}`) and locked the .tsx config coupling with a test. Thanks @imagineers-tyler. Closing.

Issue #527, opened 2026-04-23
Issue: do not respect CLAUDE_CONFIG_DIR


ChrisGVE:
While the default install for Claude setup is `~/.claude` the location can be changed using the env var `CLAUDE_CONFIG_DIR`.

This is particularly important when multiple Claude accounts must be used (for instance with different clients).

**graphify** do not respect `CLAUDE_CONFIG_DIR` and simply installs itself in the default location, and thus the automation provided to install graphify must be redone manually at the correct location.

It would be very appreciated to consider this change in behavior, if the env var exists it should be used, if not use the default.

safishamsi:
Fixed in v0.5.0. See https://github.com/safishamsi/graphify/discussions/528 for release notes.

PR #2753, opened 2026-08-15
fix(install): honor CLAUDE_CONFIG_DIR in always-on CLAUDE.md registration (#2694)

### LINKED ISSUE #2694: graphify install is not profile-isolated: CLAUDE.md registration ignores CLAUDE_CONFIG_DIR; other platforms' version stamps advance without their content
# `graphify install` is not profile-isolated: CLAUDE.md registration ignores CLAUDE_CONFIG_DIR, and other platforms' version stamps are advanced without their content

## Summary
Two related side effects of `graphify install --platform claude` (verified against 0.9.40 `install.py`):

1. **The skill copy honors `CLAUDE_CONFIG_DIR`, but the CLAUDE.md registration does not.** The skill destination resolves via `CLAUDE_CONFIG_DIR` when set (install.py ~119-120), yet the always-on registration writes `Path.home() / ".claude" / "CLAUDE.md"` unconditionally (~line 629). Running the installer from a redirected/routed Claude profile therefore mutates the DEFAULT profile's CLAUDE.md — surprising for multi-profile setups (e.g. a work profile vs personal profile split via CLAUDE_CONFIG_DIR).

2. **`.graphify_version` is advanced for every other installed platform without updating those platforms' skill contents** (install.py ~58-65). After a claude-platform install, a stale Codex/OpenCode/etc. skill directory carries a CURRENT version stamp, which suppresses the per-run "skill is from graphify X, package is Y" warning that would otherwise prompt the user to refresh it — the stamp now asserts a refresh that never happened.

## Suggested fixes
- Resolve the CLAUDE.md path through the same `CLAUDE_CONFIG_DIR` logic as the skill path (or add a `--config-dir` override), so a redirected profile is fully self-contained.
- Only stamp `.graphify_version` for platforms whose skill content was actually (re)written in this run; leave other platforms' stamps untouched so their staleness warnings stay truthful.

## Impact
Wrapper tooling that wants a safe "refresh the Claude skill after a package upgrade" step currently has to refuse to run the installer from any routed profile to avoid cross-profile writes, and cannot trust other platforms' stamps afterwards.

## Environment
graphify 0.9.40 (also present in 0.9.36), Windows 11, `graphify install --platform claude`.

AromalBiju1:
Fixes #2694

The always-on CLAUDE.md registration path resolved `Path.home()` unconditionally,
ignoring CLAUDE_CONFIG_DIR — the same issue #527 fixed for the skill-copy path,
but that fix didn't cover this always-on registration path (as noted by @yotamleo
in the issue).

This mirrors the existing CLAUDE_CONFIG_DIR check from
`_platform_skill_destination` into the CLAUDE.md registration block.

Tested locally on all three scopes:
- `--project` → writes to project-local `.claude/CLAUDE.md`, ignores CLAUDE_CONFIG_DIR (unchanged)
- global + `CLAUDE_CONFIG_DIR` set → now correctly writes to `$CLAUDE_CONFIG_DIR/CLAUDE.md`
- global, no env var → unchanged fallback to `~/.claude/CLAUDE.md`

Full test suite: 289 passed, 1 unrelated pre-existing failure (missing `build`
dev dependency in local env, unrelated to this change).

### DIFF
--- a/graphify/install.py
+++ b/graphify/install.py
@@ -633,9 +633,21 @@ def install(platform: str = "claude", *, project: bool = False, project_dir: Pat
         print(f"  command installed ->  {command_dst}")
 
     if cfg["claude_md"]:
-        # Register in the matching Claude Code scope.
-        claude_md = (project_dir / ".claude" / "CLAUDE.md") if project else Path.home() / ".claude" / "CLAUDE.md"
-        registration = _skill_registration(".claude/skills/graphify/SKILL.md" if project else "~/.claude/skills/graphify/SKILL.md")
+        # Register in the matching Claude Code scope. Honor CLAUDE_CONFIG_DIR
+        # for the global (non-project) case, same as _platform_skill_destination
+        # does for the skill copy path (#527) -- this always-on registration
+        # path was missed by that fix (#2694).
+        if project:
+            claude_md = project_dir / ".claude" / "CLAUDE.md"
+            skill_ref = ".claude/skills/graphify/SKILL.md"
+        elif os.environ.get("CLAUDE_CONFIG_DIR"):
+            config_dir = Path(os.environ["CLAUDE_CONFIG_DIR"])
+            claude_md = config_dir / "CLAUDE.md"
+            skill_ref = str(config_dir / "skills" / "graphify" / "SKILL.md")
+        else:
+            claude_md = Path.home() / ".claude" / "CLAUDE.md"
+            skill_ref = "~/.claude/skills/graphify/SKILL.md"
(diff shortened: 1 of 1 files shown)


safishamsi:
Landed in v0.9.44, just published to PyPI, cherry-picked onto `v8` with your authorship preserved. This covers the always-on CLAUDE.md registration honoring CLAUDE_CONFIG_DIR (part 1 of #2694). Thanks @AromalBiju1. Closing the PR; I've left #2694 open for the remaining version-stamp half.

PR #2755, opened 2026-08-15
Bump pypdf floor past six DoS advisories

### LINKED ISSUE #2658: Locked pypdf 6.13.3 carries six text-extraction DoS advisories on the exact extract_text() path graphify uses; the 50 MiB cap and `except Exception` do not mitigate them
## Summary

`uv.lock` pins `pypdf 6.13.3`, and `pyproject.toml` floors it at `>=6.12.0`. That version has **six** known advisories, and every one of them is an unbounded-CPU or unbounded-memory bug **in PDF text extraction** — which is exactly and only what graphify uses pypdf for.

`SECURITY.md` already states that source files are attacker-controlled ("During the semantic pass, source files are attacker-controlled text mixed into the LLM context"), so a corpus containing a hostile PDF is inside the documented threat model. Point `/graphify` at a repo, a shared drive, or a downloaded dataset containing one crafted PDF and `graphify extract` hangs or balloons.

This is filed publicly because these are already-public CVEs with published fixes, and because private vulnerability reporting on this repo returns HTTP 500 (#2011).

## The advisories

All six are reachable through `page.extract_text()`:

| ID | CVE | Impact | Fixed in |
|---|---|---|---|
| PYSEC-2026-3610 | CVE-2026-59937 | Long runtimes — repeated malformed cross-reference streams | 6.14.0 |
| PYSEC-2026-3611 | CVE-2026-59938 | Large memory use — images declaring sizes far larger than the actual data | 6.14.0 |
| PYSEC-2026-3612 | CVE-2026-59936 | **Infinite loop** — unterminated inline image, "as done when extracting the page text" | 6.14.1 |
| PYSEC-2026-3613 | CVE-2026-59935 | **Infinite loop** — same, via the ASCII85 / ASCIIHex filters | 6.14.2 |
| PYSEC-2026-3655 | CVE-2026-71870 | Large memory use — oversized `/ToUnicode` font entry, "for example during text extraction" | 6.15.0 |
| PYSEC-2026-3656 | CVE-2026-71852 | Long runtimes + large memory — oversized font width entries, "for example during text extraction" | 6.15.0 |

Two of the upstream advisories name page-text extraction as the triggering operation in their own text.

## The call path

`graphify/detect.py:529`:

```python
def extract_pdf_text(path: Path) -> str:
    """Extract plain text from a PDF file using pypdf."""
    if not _file_within_size_cap(path):
        return ""
    try:
        from pypdf import PdfReader
        reader = PdfReader(str(path))
        pages = []
        for page in reader.pages:
            text = page.extract_text()      # <- every advisory above lands here
            if text:
                pages.append(text)
        return "\n".join(pages)
    except Exception:
        return ""
```

Reached from two places, so both the scan and the semantic pass are exposed:

- `detect.py:779` — word counting during corpus scan
- `llm.py:506` — document text for the semantic pass

## Why the existing guards do not cover it

**The size cap does not apply.** `_file_within_size_cap` rejects files over 50 MiB on disk, but every one of these is an *amplification* bug: a small, well-formed-looking PDF produces the runaway loop or allocation. A 20 KB file passes the cap and still hangs.

**`except Exception` does not catch a hang.** It converts the two memory advisories into a silent `return ""` — the document is dropped from the graph with no warning, which is its own quiet failure — but an infinite loop never raises, so PYSEC-2026-3612 and -3613 wedge the process. There is no `signal.alarm`, timeout, or subprocess isolation anywhere on this path (`grep -n "signal.alarm\|SIGALRM\|timeout" graphify/detect.py` returns nothing).

Worth noting the project already defends against precisely this class for Office files — `_zip_within_caps` exists to stop zip/XML bombs in `.docx`. PDF simply has no equivalent, and here the mitigation is free because upstream already shipped it.

**CI did not surface it.** The `security-scan` job runs `pip-audit --strict` with `continue-on-error: true`, so the finding is reported and then discarded. On current `v8`:

```
$ uv run --frozen pip-audit --strict
Found 34 known vulnerabilities in 7 packages
```

## Suggested fix

Raise the floor past all six and refresh the lock:

```toml
pdf = ["pypdf>=6.15.0", "markdownify"]
```

(`>=6.15.0` clears every advisory in the table; the highest fix version is 6.15.0.) The same string appears twice — in the `pdf` extra and in `all` — so both need it, and `uv lock` then moves `uv.lock` off 6.13.3.

I am happy to send that PR if you want it — say the word and I will open one. Holding off because a lockfile bump is a maintainer call, and the same audit run flags six other packages where the judgement is genuinely yours:

- `pillow 12.2.0` — 12 advisories, but graphify never imports PIL directly (it arrives transitively via matplotlib/umap), so the exposure is materially weaker than pypdf's.
- `cryptography`, `setuptools`, `soupsieve`, `yt-dlp`, `pip` — the rest of the 34, none on an obvious attacker-reachable path.

A separate question worth its own issue if you'd rather: `extract_pdf_text` swallowing `MemoryError` into `return ""` means a PDF can vanish from the graph silently. Even after the bump, a wall-clock guard on that path would turn "hangs forever" into "skips one file loudly."

## Environment

| | |
|---|---|
| graphify | `v8` @ `26128ab` (0.9.40) |
| pypdf | 6.13.3 (from committed `uv.lock`) |
| Verified with | `uv run --frozen pip-audit`, advisory text from the OSV records |

ayushcodes10:
pypdf 6.13.3 (the version uv.lock pinned) carries six known advisories,
every one an unbounded-CPU or unbounded-memory bug in PDF text
extraction -- exactly and only what graphify uses pypdf for, via
extract_pdf_text() in detect.py. Two are infinite loops that the
existing 50 MiB size cap and broad except Exception do not catch: the
cap only rejects large files on disk, but these are amplification bugs
where a small, well-formed-looking PDF triggers the runaway behavior,
and a hang never raises to be caught.

Raise the floor to >=6.15.0 (the highest "fixed in" version across all
six advisories) in both the pdf and all extras, and refresh uv.lock,
which resolves to the current latest (6.16.1, zero known advisories).

Fixes #2658.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -60,7 +60,11 @@ Issues = "https://github.com/Graphify-Labs/graphify/issues"
 mcp = ["mcp>=1,<3", "starlette>=1.3.1,<2"]
 neo4j = ["neo4j"]
 falkordb = ["falkordb"]
-pdf = ["pypdf>=6.12.0", "markdownify"]
+# >=6.15.0 clears six pypdf DoS advisories (unbounded CPU/memory in
+# page.extract_text()) reachable through extract_pdf_text() on an
+# attacker controlled PDF (#2658); 6.15.0 is the highest "fixed in" version
+# across all six.
+pdf = ["pypdf>=6.15.0", "markdownify"]
 watch = ["watchdog"]
 svg = ["matplotlib", "numpy>=2.0; python_version >= '3.13'"]
 leiden = ["graspologic; python_version < '3.13'"]
@@ -99,7 +103,7 @@ commonlisp = ["tree-sitter-commonlisp"]
 # every platform, no C toolchain; optional because Robot Framework corpora are
 # QA-automation specific.
 robot = ["robotframework>=4.0"]
-all = ["mcp>=1,<3", "starlette>=1.3.1,<2", "neo4j", "falkordb", "pypdf>=6.12.0", "markdownify", "watchdog", "graspologic; python_version < '3.13'", "python-docx", "openpyxl", "faster-whisper; python_version >= '3.11'", "yt-dlp>=2026.6.9", "matplotlib", "numpy>=2.0; python_version >= '3.13'", "openai", "tiktoken", "boto3", "anthropic", "tree-sitter-sql", "jieba", "tree-sitter-dm", "tree-sitter-hcl", "tree-sitter-pascal", "tree-sitter-ocaml", "tree-sitter-commonlisp", "robotframework>=4.0"]
+all = ["mcp>=1,<3", "starlette>=1.3.1,<2", "neo4j", "falkordb", "pypdf>=6.15.0", "markdownify", "watchdog", "graspologic; python_version < '3.13'", "python-docx", "openpyxl", "faster-whisper; python_version >= '3.11'", "yt-dlp>=2026.6.9", "matplotlib", "numpy>=2.0; python_version >= '3.13'", "openai", "tiktoken", "boto3", "anthropic", "tree-sitter-sql", "jieba", "tree-sitter-dm", "tree-sitter-hcl", "tree-sitter-pascal", "tree-sitter-ocaml", "tree-sitter-commonlisp", "robotframework>=4.0"]
 
 [project.scripts]
 graphify = "graphify.__main__:main"
--- a/uv.lock
+++ b/uv.lock
@@ -6,10 +6,10 @@ resolution-markers = [
     "python_full_version >= '3.14' and sys_platform == 'emscripten'",
     "python_full_version >= '3.14' and sys_platform != 'emscripten' and sys_platform != 'win32'",
     "python_full_version == '3.13.*' and sys_platform == 'win32'",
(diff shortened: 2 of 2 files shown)


ayushcodes10:
Re: "all extra now rejects previously supported pypdf 6.12–6.14" — this is the intended effect of the fix, not a regression. 6.12–6.14 are exactly the vulnerable range the six advisories in the issue affect; raising the floor to `>=6.15.0` is what excludes them.

Compatibility-wise, graphify's only pypdf surface is `PdfReader(path)` → `.pages` → `page.extract_text()` (detect.py:540-554), the most basic and stable part of pypdf's public API — unchanged across 6.x. I also ran the full test suite with pypdf 6.16.1 actually installed (not just the version bump on paper): 4314 passed, 0 failed, including `tests/test_office_limits.py` and `tests/test_image_vision.py`, both of which exercise `extract_pdf_text()` directly. `pip-audit` confirms zero pypdf advisories remain after the bump.

ayushcodes10:

PR #2758, opened 2026-08-15
fix: an import from outside the corpus must shadow indirect_call reso…

### LINKED ISSUE #2757: JS/TS: an import from outside the corpus does not shadow indirect_call resolution, fabricating cross-package edges
### Summary

A name bound by `import { X } from "<external package>"` is in neither JS/TS shadow set, so using it as a value produces a fabricated `indirect_call` edge (INFERRED, 0.8) to an unrelated same-named definition elsewhere in the corpus.

This is the same failure `test_indirect_call_catch_binding_shadow.py` documents for `catch` bindings — *"read as an unresolved by-name reference, resolved against the corpus-wide label index, and fabricated an `indirect_call` edge (INFERRED, 0.8) to an unrelated same-named callable"* — and the same one already fixed for single-parameter arrows and untracked closures (#2241 family). An import is simply a module-scoped binding, and it was never added to the set.

### Reproduction

Two files, two packages. `Widget` comes from an external package; `pkg-b` happens to define a `Widget()` that `consumer.tsx` never imports.

```tsx
// pkg-a/consumer.tsx
import { Widget, Search } from 'some-external-pkg';
export function buildMenu() {
  return [{ label: 'p', icon: Widget }, { label: 'f', icon: Search }];
}
```

```tsx
// pkg-b/Widget.tsx
export function Widget() { return null; }
```

```
$ graphify update .
indirect_call  pkg_a_consumer_buildmenu -> pkg_b_widget_widget  [INFERRED]
```

`consumer.tsx` never references `pkg-b`. `Search`, imported from the same external package and used the same way, correctly produces nothing — because no local definition shares its name. That contrast is the mechanism: it is the unique-label fallback firing on an identifier that is already bound.

### Why it lands often in practice

Icon kits collide with component names. Found on a monorepo where three unrelated components appeared to call a `Palette()` in a different package; all three import the **lucide-react** icon of that name. `analyze.py::_surprise_score` then pays +2 for crossing a directory, +2 for crossing a file type and +2 for INFERRED — so these edges lead the "Surprising Connections" section of `GRAPH_REPORT.md`, where they read as discoveries.

### Mechanism

`_js_local_bound_names` collects a function's locals from `parameters` and `variable_declarator` nodes; `_js_module_bound_names` collects only the latter. An `import_statement` binding is neither, so the name never reaches the shadow check in `_emit_indirect_ref` and falls through to `_rewire_unique_stub_nodes`' label index.

### Suggested fix

Collect names bound by an import whose specifier the resolver says is external, and consult that set in `_emit_indirect_ref` — an import binds the whole module, so one file-level set covers every scope.

Deciding "external" by asking `_resolve_js_import_target` rather than forming a second opinion keeps relative imports resolving, which is the behaviour that must not regress. One extra condition is needed: a `tsconfig` `paths` entry pointing a package at its own installed copy (`"lucide-react": ["./node_modules/lucide-react"]`, which is common) *does* resolve — to a tree every scan prunes — so a resolved path containing `node_modules` still counts as external.

### Tests

A patch along those lines is on a fork branch if it is useful — happy to open a PR.

- 7 new tests in `tests/test_indirect_call_external_import_shadow.py`, modelled on the existing shadow tests: **5 red on `v8` without the patch, 7 green with it.** The 2 that pass either way are the counter-tests that bound the fix — a relative import must still resolve, and an unshadowed by-name reference in a file that also imports something external must still emit.
- Full suite, same machine, with and without: `5 failed / 4386 passed` → `5 failed / 4393 passed`. Identical failures both ways (`test_ollama_retry_cap` needs `openai`, `test_install` needs the codex CLI), so they are environmental; the delta is exactly the 7 new tests.

phudayyy:
Fixes #2757

`_js_local_bound_names` collects a function's locals from parameters and `variable_declarator` nodes, and `_js_module_bound_names` only the latter. A name introduced by `import { X } from "pkg"` is neither, so it was absent from both shadow sets: listing it in a dispatch table (`{ icon: X }`) or passing it on as a call argument read as an unresolved by-name reference, resolved against the corpus-wide label index, and fabricated an `indirect_call` edge (INFERRED, 0.8) to an unrelated same-named callable elsewhere in the corpus.

Same class as the `catch`-binding (#2241 family), single-parameter-arrow and untracked-closure shadows already fixed — an import is simply a module-scoped binding. A UI icon kit makes it land constantly: `Palette`, `Search`, `Filter` and `Menu` are icon exports *and* ordinary component names, so any repo with both grows cross-package edges between files that never referenced one another. Found on a monorepo where three unrelated components appeared to call a `Palette()` in a different package; all three import the lucide-react icon.

The guard asks `_resolve_js_import_target` rather than forming a second opinion, so a relative import — which resolves to a real file, and whose edge is the whole point of the graph — keeps resolving. `node_modules` is tested separately because a `tsconfig` `paths` entry pointing a package at its own installed copy (`"lucide-react": ["./node_modules/lucide-react"]`) *does* resolve, to a tree every scan prunes.

Tests: 5 of 7 red on v8 without the patch, 7 of 7 green with it. The two that pass either way are the counter-tests that bound the fix — a relative import must still resolve, and an unshadowed by-name reference in a file that also imports something external must still emit.

Full suite, same machine, with and without: 5 failed / 4386 passed and 5 failed / 4393 passed. Identical failures both ways (`test_ollama_retry_cap` needs `openai`, `test_install` needs the codex CLI), so they are environmental; the delta is exactly the 7 new tests.

### DIFF
--- a/graphify/extractors/engine.py
+++ b/graphify/extractors/engine.py
@@ -1287,6 +1287,81 @@ def _js_module_bound_names(root, source: bytes) -> set[str]:
     walk(root)
     return bound
 
+def _js_import_binds_external(raw: str, str_path: str) -> bool:
+    """True when a JS/TS import specifier names a module outside the scanned corpus.
+
+    Reuses `_resolve_js_import_target`, so this is graphify's own verdict rather
+    than a second opinion: a specifier it cannot resolve is an external package
+    (the `ref`-namespaced branch). The extra `node_modules` test covers the case
+    where resolution *succeeds* but lands in a dependency tree — a `tsconfig`
+    `paths` entry mapping a package to its own installed copy
+    (`"lucide-react": ["./node_modules/lucide-react"]`) is common, and
+    `node_modules` is pruned from every scan, so the target is never a node.
+    """
+    resolved = _resolve_js_import_target(raw, str_path)
+    if resolved is None:
+        return False  # empty specifier — binds nothing
+    _target_nid, resolved_path = resolved
+    if resolved_path is None:
+        return True  # unresolved after relative / alias / workspace lookup
+    return "node_modules" in resolved_path.parts
+
+
+def _js_external_import_names(root, source: bytes, str_path: str) -> set[str]:
+    """Names an `import` binds to a module OUTSIDE the corpus.
+
+    An imported name is a module-scoped binding: within this file it denotes the
+    imported symbol and nothing else. Neither shadow set collects it —
+    `_js_local_bound_names` reads parameters and `variable_declarator`s and
+    `_js_module_bound_names` only the latter — so the name reaches
+    `_emit_indirect_ref` as an unresolved by-name reference, gets resolved against
+    the corpus-wide label index, and fabricates an `indirect_call` (INFERRED, 0.8)
+    to any unique same-named callable elsewhere in the corpus. That is the symptom
+    already fixed for `catch` bindings, single-parameter arrows and untracked
+    closures; an import binding is the same class of shadow, and a UI kit makes it
+    land constantly because icon names (`Palette`, `Search`, `Filter`) collide with
+    ordinary component names.
+
+    Only imports the corpus cannot contain are collected. A relative specifier
+    resolves to a real file and that edge is the graph's whole point, so those
+    names stay resolvable.
+    """
+    bound: set[str] = set()
+
+    def _clause_names(clause) -> None:
+        for c in clause.children:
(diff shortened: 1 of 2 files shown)


safishamsi:
Landed in v0.9.44, just published to PyPI. I cherry-picked this onto `v8` with your authorship preserved in the commit, so both the change and the credit are in the history. Thanks @phudayyy for the fix. Closing since it is now released.

PR #2762, opened 2026-08-15
Fix SSRF guard rejecting private proxies

### LINKED ISSUE #2761: graphify add cannot use macOS system HTTP(S) proxy on a private/LAN address
# `graphify add` cannot use macOS system HTTP(S) proxy on a private/LAN address

## Summary

`graphify add <url>` fails when macOS system proxy settings point to a legitimate HTTP/HTTPS proxy on a private RFC1918 address.

Graphify correctly discovers and attempts to use the macOS system proxy via Python `urllib`, but its SSRF-guarded HTTP(S) connection then applies the private/reserved-IP check to the **proxy endpoint itself** and rejects the connection.

Example error:

```text
error: ingest: failed to fetch 'https://github.com/Graphify-Labs/graphify': <urlopen error SSRF blocked: IP 192.168.64.1 resolved from 'proxy.example.com' is private/reserved>
```

The target URL resolves to public Cloudflare addresses and is reachable successfully through the configured proxy with `curl`.

## Environment

- graphifyy: `0.9.43`
- Python used by graphify: `3.14`
- macOS
- Installed with `uv tool`
- Graphify executable:

```text
/Users/<user>/.local/bin/graphify
```

- Python environment:

```text
/Users/<user>/.local/share/uv/tools/graphifyy/bin/python
```

- macOS system HTTP/HTTPS proxy:

```text
http://proxy.example.com:3128
```

where:

```text
proxy.example.com -> 192.168.64.1
```

## Reproduction

With the system proxy enabled:

```bash
graphify add https://github.com/Graphify-Labs/graphify
```

produces:

```text
error: ingest: failed to fetch 'https://github.com/Graphify-Labs/graphify': <urlopen error SSRF blocked: IP 192.168.64.1 resolved from 'proxy.example.com' is private/reserved>
```

The target itself resolves normally:

```bash
dig github.com

; <<>> DiG 9.10.6 <<>> github.com
;; global options: +cmd
;; Got answer:
;; ->>HEADER<<- opcode: QUERY, status: NOERROR, id: 19451
;; flags: qr rd ra; QUERY: 1, ANSWER: 1, AUTHORITY: 0, ADDITIONAL: 1

;; OPT PSEUDOSECTION:
; EDNS: version: 0, flags:; udp: 1232
;; QUESTION SECTION:
;github.com.			IN	A

;; ANSWER SECTION:
github.com.		58	IN	A	20.26.156.215
```

and the URL is reachable through the proxy:

```bash
curl -IL https://github.com/Graphify-Labs/graphify
```

returns HTTP 200.

## macOS system proxy is the source

Removing all proxy-related environment variables does not remove the proxy from `urllib`:

```bash
env -u HTTP_PROXY -u HTTPS_PROXY -u http_proxy -u https_proxy \
/Users/<user>/.local/share/uv/tools/graphifyy/bin/python - <<'PY'
import urllib.request
import os

print("env:")
for k, v in os.environ.items():
    if "proxy" in k.lower():
        print(k, v)

print("urllib:")
print(urllib.request.getproxies())
PY
```

Output:

```text
env:
urllib:
{'http': 'http://proxy.example.com:3128',
 'https': 'http://proxy.example.com:3128'}
```

This is expected behaviour on macOS: `urllib.request.getproxies()` discovers the OS-level proxy configuration.

## Root cause

The failure originates in `graphify/security.py`.

Graphify installs custom handlers/connections:

```text
_SSRFGuardedHTTPConnection
_SSRFGuardedHTTPSConnection
_SSRFGuardedHTTPHandler
_SSRFGuardedHTTPSHandler
```

The guarded connection rejects private/reserved addresses with:

```python
raise OSError(
    f"SSRF blocked: IP {addr} resolved from '{host}' is private/reserved"
)
```

When `urllib` is using an HTTP(S) proxy, the connection host is the proxy:

```text
proxy.example.com -> 192.168.64.1
```

so the SSRF protection rejects the proxy before the request can be sent.

The resulting flow appears to be:

```text
graphify add public-url
        |
        v
urllib discovers macOS system proxy
        |
        v
connect to proxy.example.com:3128
        |
        v
Graphify SSRF guard resolves proxy hostname
        |
        v
192.168.64.1
        |
        v
rejected as private/reserved
```

## Expected behaviour

A deliberately configured HTTP(S) proxy should be allowed to reside on a private/LAN address.

The SSRF validation should apply to the **requested destination URL**, including redirects, but should not unconditionally reject the configured proxy endpoint merely because that proxy is on RFC1918 space.

This is a normal deployment pattern for:

- corporate HTTP proxies
- home/lab proxies
- VPN/private-network proxies
- VM host-side proxies
- security gateways

## Suggested fix

Distinguish between:

1. the destination URL being fetched, which should remain subject to the SSRF policy; and
2. the configured proxy endpoint, which may legitimately be private.

Possible approaches:

- validate the target URL independently, then allow connections to an explicitly configured proxy even when its address is private; or
- make the guarded connection proxy-aware and skip the private-IP rejection only for the proxy hop while preserving SSRF checks for the final destination and redirects.

An explicit opt-in such as:

```text
GRAPHIFY_ALLOW_PRIVATE_PROXY=1
```

could also be useful as a fallback, though proxy-aware handling would be preferable.

## Related issues

Related but distinct:

- #1211 — SSRF guard / `socket.getaddrinfo` implementation issue
- #1168 — legitimate private/local backend rejected by an SSRF-related safeguard

Neither appears to cover private HTTP(S) proxies discovered from macOS system proxy settings.

## Impact

On macOS, users with a system HTTP/HTTPS proxy configured to a LAN/private address cannot use:

```bash
graphify add <url>
```

even for ordinary public HTTPS URLs.

This also means explicitly setting `HTTP_PROXY` / `HTTPS_PROXY` to the same private proxy does not help, because the Graphify connection guard rejects the proxy endpoint itself.

shard-c6:
Fixes #2761

This PR modifies the SSRF guard connection classes (`_SSRFGuardedHTTPConnection` and `_SSRFGuardedHTTPSConnection`) to be proxy-aware. When a request is explicitly configured to use a proxy, the private IP address check is bypassed for the proxy endpoint itself, while `validate_url` independently preserves SSRF protections for the final destination URL.

Changes:
- Added `_ProxyAwareHTTPConnection` and `_ProxyAwareHTTPSConnection` subclasses to safely connect to proxies without applying IP blocking on the proxy's resolved address.
- Updated `_SSRFGuardedHTTPHandler` and `_SSRFGuardedHTTPSHandler` to detect if the request is using a proxy (`req.has_proxy()` or `req._tunnel_host`) and dispatch to the correct connection type.

### DIFF
--- a/graphify/security.py
+++ b/graphify/security.py
@@ -192,6 +192,19 @@ class _SSRFGuardedHTTPConnection(http.client.HTTPConnection):
             self._tunnel()
 
 
+class _ProxyAwareHTTPConnection(_SSRFGuardedHTTPConnection):
+    """HTTPConnection that skips SSRF IP validation for the proxy hop."""
+
+    def connect(self) -> None:
+        self.sock = socket.create_connection(
+            (self.host, self.port),
+            self.timeout,
+            self.source_address,
+        )
+        if self._tunnel_host:
+            self._tunnel()
+
+
 class _SSRFGuardedHTTPSConnection(http.client.HTTPSConnection):
     """HTTPSConnection variant of _SSRFGuardedHTTPConnection.
 
@@ -214,10 +227,28 @@ class _SSRFGuardedHTTPSConnection(http.client.HTTPSConnection):
         self.sock = self._context.wrap_socket(sock, server_hostname=self.host)
 
 
+class _ProxyAwareHTTPSConnection(_SSRFGuardedHTTPSConnection):
+    """HTTPSConnection that skips SSRF IP validation for the proxy hop."""
+
+    def connect(self) -> None:
+        sock = socket.create_connection(
+            (self.host, self.port),
+            self.timeout,
+            self.source_address,
+        )
+        if self._tunnel_host:
+            self.sock = sock
+            self._tunnel()
+            sock = self.sock
+        self.sock = self._context.wrap_socket(sock, server_hostname=self.host)
+
+
 class _SSRFGuardedHTTPHandler(urllib.request.HTTPHandler):
     """urllib handler that routes http:// through _SSRFGuardedHTTPConnection."""
 
     def http_open(self, req):
+        if req.has_proxy():
+            return self.do_open(_ProxyAwareHTTPConnection, req)
         return self.do_open(_SSRFGuardedHTTPConnection, req)
 
 
@@ -225,6 +256,8 @@ class _SSRFGuardedHTTPSHandler(urllib.request.HTTPSHandler):
     """urllib handler that routes https:// through _SSRFGuardedHTTPSConnection."""
 
     def https_open(self, req):
+        if getattr(req, "_tunnel_host", None) or req.has_proxy():
+            return self.do_open(_ProxyAwareHTTPSConnection, req)
         return self.do_open(_SSRFGuardedHTTPSConnection, req)
 
 


fgtb:
user@rhovanion graphify % git checkout fix-issue-2761
Already on 'fix-issue-2761'
Your branch is up to date with 'origin/fix-issue-2761'.
Then user@rhovanion graphify % uv tool install --force .
...Installed 2 executables: graphify, graphify-mcp.

user@rhovanion graphify % cd ~/tmp ; graphify add https://www.bbc.co.uk
error: ingest: failed to fetch 'https://www.bbc.co.uk': <urlopen error [SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed: Hostname mismatch, certificate is not valid for '<hostname removed>'. (_ssl.c:1082)>
user@rhovanion graphify % graphify add https://wired.com
error: ingest: failed to fetch 'https://wired.com': <urlopen error [SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed: Hostname mismatch, certificate is not valid for '<hostname removed>'. (_ssl.c:1082)>

Looks like the proposed fix gets further than before, but it bumps into an SSL cert error. My proxy is running squid and functions fine for normal web browsing:

user@rhovanion Programming % curl -vI https://www.bbc.co.uk
* Uses proxy env variable HTTPS_PROXY == 'http://<hostname>:3128'
* Host <hostname>:3128 was resolved.
* IPv6: (none)
* IPv4: 192.168.64.1
*   Trying 192.168.64.1:3128...
* Connected to <hostname> (192.168.64.1) port 3128
* CONNECT tunnel: HTTP/1.1 negotiated
* allocate connect buffer
* Establish HTTP proxy tunnel to www.bbc.co.uk:443
> CONNECT www.bbc.co.uk:443 HTTP/1.1
> Host: www.bbc.co.uk:443
> User-Agent: curl/8.7.1
> Proxy-Connection: Keep-Alive
>
< HTTP/1.1 200 Connection established
HTTP/1.1 200 Connection established
...

(private domain names removed for privacy).

PR #2766, opened 2026-08-15
fix(detect): preserve Git-tracked files matched by .gitignore (#2759)

### LINKED ISSUE #2759: A git-TRACKED file is silently dropped when a .gitignore pattern matches it (git itself does not ignore tracked paths)
### Summary

A file that **git tracks** is silently dropped from the scan when a `.gitignore` pattern matches it. Git itself does not ignore such a file — ignore rules do not apply to tracked paths — so the graph is missing source the repository demonstrably ships, and nothing says so.

### Reproduction

```bash
git init . && mkdir -p storage src
echo 'export function watch(){ return 1; }' > storage/fileWatcher.js
echo 'export function ok(){ return 2; }'    > src/app.js
git add -A && git commit -qm base          # tracked BEFORE the ignore rule exists
printf 'storage/\n' > .gitignore
echo 'export function untracked(){ return 3; }' > storage/scratch.js   # control
git add .gitignore && git commit -qm ignore
```

What git says:

| | |
|---|---|
| `git ls-files --error-unmatch storage/fileWatcher.js` | tracked |
| `git check-ignore storage/fileWatcher.js` | **not ignored** (tracked paths are exempt) |
| `git check-ignore storage/scratch.js` | ignored |

What `detect()` returns:

```
files:              ['app.js']
skipped_sensitive:  []
```

`storage/scratch.js` being dropped is correct — it is untracked and genuinely ignored. `storage/fileWatcher.js` is the bug: git ships it, git does not ignore it, and it is absent with no report.

### Why it is easy to hit and hard to notice

A bare directory pattern like `storage/`, `build/` or `data/` is common in a `.gitignore` written for runtime junk, and it matches **any** directory of that name at any depth. On the repo where this was found it hid `finch-desktop/electron/storage/` — 4 files, 878 lines of shipping product code — behind a pattern intended for a local scratch directory elsewhere in the tree.

The failure mode is silence, and it is the worst kind for a knowledge graph: querying anything in that directory returns **empty**, which is indistinguishable from "nothing references this". The graph does not look wrong; it looks like the code is dead.

### Two things that make this feel like a fidelity gap rather than a design choice

- **`skipped_sensitive` already exists as the channel for "a file was dropped and you should know"** (added for #2106, so a wrongly-flagged source is visible). Ignore-dropped tracked files use no channel at all.
- **graphify already consults git where it helps** — `export.py` shells out to `git rev-parse HEAD` to stamp `built_at_commit`, and `hooks.py`/`prs.py`/`watch.py` call git too. So "we deliberately never ask git" does not appear to be the principle here.

### Suggested fix

Two options, either of which closes it:

1. **When a `.git` directory is present, exempt tracked paths from `.gitignore` matching** — i.e. implement the same rule git implements. `git ls-files` once per scan root is enough, and a folder with no `.git` keeps today's behaviour exactly.
2. **At minimum, report it.** If dropping them stays the intended behaviour, list them in `skipped_sensitive` (or a sibling field) so it is visible rather than silent. That alone would have turned this from a two-day mystery into a line of output.

Happy to send a patch for whichever direction you prefer — I did not want to guess between them.

### Note on `.graphifyignore`

`.graphifyignore` is unaffected: a pattern written there is an explicit instruction to this tool, and dropping a tracked file because of it is exactly right. The divergence only matters for `.gitignore`, which is written for a different tool with different semantics.

NithishKumar04:
## Summary

- preserve files already tracked by Git when a `.gitignore` or `.git/info/exclude` rule matches them
- keep ignored untracked siblings excluded
- keep `.graphifyignore` and CLI/persisted `--exclude` rules authoritative for tracked and untracked paths alike
- apply the same decision in `detect()` and `ignored_predicate()` so full scans, incremental rebuilds, and watch reconciliation agree

Fixes #2759.

## Root cause

Discovery merged `.gitignore` and graph-specific rules into one pattern list and evaluated that list without consulting Git's index. An ignored directory was pruned before walking it, so a tracked source file below that directory never reached per-file classification. This diverged from Git itself: ignore rules apply only to untracked paths.

The shared `ignored_predicate()` repeated the same decision, so fixing only the directory walk would still let a full watch rebuild evict the tracked file from an existing graph.

## Approach

- ask Git once per scan/predicate construction for NUL-delimited cached paths using `git ls-files -z --cached`
- normalize path identities for platform case rules and Unicode composition
- retain only currently readable regular files inside the requested scan root; deleted index rows and submodule gitlinks do not affect traversal
- build an ancestor-directory set so a Git-only ignored directory containing tracked source is not pruned
- evaluate `.graphifyignore` and `--exclude` separately; if either explicitly excludes the path, it remains excluded regardless of Git tracking
- preserve historical behavior when the scan is outside Git, Git is unavailable, or the read-only query fails
- disable optional Git index locking for the read-only query

## Regression coverage

Six focused tests cover:

1. a tracked file under an ignored directory is retained while an untracked sibling is dropped
2. `.graphifyignore` still excludes a tracked file
3. repository-relative tracking works when scanning a nested project root
4. `ignored_predicate()` agrees with discovery for watch reconciliation
5. `--exclude` still excludes a tracked file
6. a failed Git probe falls back without aborting or including ignored files

## Verification

- `4398 passed, 47 skipped` — full suite with isolated Git configuration and the required optional dependency
- `247 passed` — complete discovery suite
- `130 passed, 2 skipped` — watch and incremental suites
- focused regressions pass on Python 3.10 and 3.12
- `ruff check graphify tests`
- Bandit high/medium scan clean for `graphify/detect.py`
- all generated-artifact checks: `--check`, `--audit-coverage`, `--schema-singleton`, `--monolith-roundtrip`, and `--always-on-roundtrip`
- repository graph refreshed successfully


NithishKumar04:
@safishamsi This is a focused fix for #2759. The main review point is the separation between Git-only ignore rules and explicit `.graphifyignore`/`--exclude` rules: only the former are bypassed for tracked paths, and `ignored_predicate()` shares the same logic so watch/full rebuild behavior stays consistent. The full suite is green; the workflow currently awaits maintainer approval to run.

safishamsi:
Landed in v0.9.44, just published to PyPI. I cherry-picked this onto `v8` with your authorship preserved in the commit, so both the change and the credit are in the history. Thanks @NithishKumar04 for the fix. Closing since it is now released.

PR #2768, opened 2026-08-15
fix(hooks): persist viz node limit in project config (#2760)

### LINKED ISSUE #2760: hook install: no way to persist GRAPHIFY_VIZ_NODE_LIMIT into the generated hook (or share the setting with a team)
## Problem

`graphify hook install` generates `.git/hooks/post-commit` and `post-checkout` from a static template (`graphify/hooks.py`). The template already respects `GRAPHIFY_VIZ_NODE_LIMIT` at rebuild time (per #541), but there's no way to *bake a value in* at install time — the only option is hand-editing the generated hook file after the fact.

That has two consequences:
1. `.git/hooks/` is never version-controlled, so a hand-edit doesn't propagate to teammates who run `graphify hook install` on their own machine — each person has to remember to make the same edit.
2. Any edit is silently clobbered the next time `graphify hook install` (or `--force`) regenerates the file from the template.

## Reproducer

1. `graphify hook install` in a repo.
2. Manually add `export GRAPHIFY_VIZ_NODE_LIMIT=0` to `.git/hooks/post-commit`.
3. Have a teammate clone the repo and run `graphify hook install` themselves — their hook has no such export, so their auto-rebuilds still regenerate `graph.html`.
4. Alternatively, re-run `graphify hook install --force` yourself — the manual edit is gone.

## Proposal

Any one of:
1. **Install-time flag.** `graphify hook install --viz-node-limit 0` bakes `export GRAPHIFY_VIZ_NODE_LIMIT=0` into the generated template.
2. **Repo-level config.** A committed config (e.g. `.graphifyrc` / a key in an existing config file already read by `graphify`) that `hook install` reads and bakes into the generated hook, and that `graphify hook status` can also report on. This also solves the "share with team" half of the problem, since the config file itself is checked in even though `.git/hooks/` isn't.

Option 2 is preferable — it makes the setting a real, shared, checked-in project decision instead of a per-machine hand-edit that has to be remembered and re-applied after every reinstall.

## Motivation

We disable HTML viz regen on hook-triggered rebuilds (`GRAPHIFY_VIZ_NODE_LIMIT=0`) across a multi-repo team because `graph.html` isn't consumed by `query`/`path`/`explain`, the MCP server, or Claude sessions — it's pure layout churn on every commit. Right now that setting only exists as a manual, per-machine edit to an unversioned file, so it can't actually be enforced or shared team-wide.

hopstreax:
## Summary

Fixes #2760.

Adds repository-level `.graphifyrc` support for persisting `GRAPHIFY_VIZ_NODE_LIMIT` across hook installations.

Previously, this setting required manually editing `.git/hooks/`, which is not version-controlled, not shared with teammates, and can be lost when hooks are regenerated.

## Changes

- Add support for a repository-level `.graphifyrc`.
- Support `viz_node_limit=<value>` configuration.
- Validate the value as a non-negative integer.
- Bake the configured value into both `post-commit` and `post-checkout` hooks.
- Update an existing Graphify hook block when `graphify hook install` is run again.
- Preserve user-defined hook content outside the Graphify-owned block.
- Report the configured value through `graphify hook status`.
- Preserve existing `GRAPHIFY_VIZ_NODE_LIMIT` runtime behavior when no `.graphifyrc` setting is configured.

## Example

```ini
# .graphifyrc
viz_node_limit=0
````

Running:

```bash
graphify hook install
```

generates hooks containing:

```bash
export GRAPHIFY_VIZ_NODE_LIMIT="0"
```

This allows the setting to be committed with the repository and consistently applied when teammates install their Graphify hooks.

## Tests

Added coverage for:

* `.graphifyrc` parsing and validation
* Default behavior without `.graphifyrc`
* Baking the configured value into both hooks
* Updating an existing Graphify hook after configuration changes
* Preserving user hook content
* `graphify hook status` reporting
* Applying the setting to both `post-commit` and `post-checkout`

All #2760-specific tests pass.

### DIFF
--- a/graphify/hooks.py
+++ b/graphify/hooks.py
@@ -283,7 +283,7 @@ _HOOK_SCRIPT = """\
 # order is randomized per-process by PYTHONHASHSEED, so community assignments
 # churn run-to-run. Pinning it makes graphify-out reproducible.
 export PYTHONHASHSEED=0
-
+__VIZ_LIMIT_EXPORT__
 # Git for Windows/MSYS hooks can inherit fragile pipe handles from GUI clients
 # and agent shells. Keep hook-triggered rebuilds sequential by default there;
 # explicit GRAPHIFY_MAX_WORKERS still wins for users who want parallelism.
@@ -338,7 +338,7 @@ _CHECKOUT_SCRIPT = """\
 # order is randomized per-process by PYTHONHASHSEED, so community assignments
 # churn run-to-run. Pinning it makes graphify-out reproducible.
 export PYTHONHASHSEED=0
-
+__VIZ_LIMIT_EXPORT__
 # Git for Windows/MSYS hooks can inherit fragile pipe handles from GUI clients
 # and agent shells. Keep hook-triggered rebuilds sequential by default there;
 # explicit GRAPHIFY_MAX_WORKERS still wins for users who want parallelism.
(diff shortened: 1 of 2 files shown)


safishamsi:
Landed in v0.9.44, just published to PyPI. I cherry-picked this onto `v8` with your authorship preserved in the commit, so both the change and the credit are in the history. Thanks @hopstreax for the fix. Closing since it is now released.

PR #2769, opened 2026-08-15
fix: a git-tracked file is not dropped by a .gitignore pattern

### LINKED ISSUE #2759: A git-TRACKED file is silently dropped when a .gitignore pattern matches it (git itself does not ignore tracked paths)

phudayyy:
Fixes #2759

`git check-ignore` reports a tracked path as NOT ignored — ignore rules apply to untracked files only. Reading `.gitignore` as a plain pattern file diverges from that on the one case that matters, a file the repository demonstrably ships, and it diverged silently: nothing lands in `skipped_sensitive`, so querying the file afterwards returns nothing and reads as dead code.

Found where a bare `storage/`, written for a scratch directory elsewhere in the tree, hid four files and 878 lines of shipping product code.

The rescue asks git once per scan (`git ls-files`) and only for a path that is already about to be dropped, so a corpus with nothing ignored spawns no subprocess at all. An eager call cost ~100ms per detect() and that alone broke test_same_size_rewrite_in_one_tick_is_requeued, whose stat-vs-hash fastpath is timing-sensitive — worth knowing before anyone moves this call.

Both gates are covered: the directory prune (a directory is skipped before its files are ever walked, so it is rescued when it holds any tracked file) and the per-file gate, which still drops the untracked ones.

.graphifyignore is deliberately NOT rescued. A pattern written there is an explicit instruction to this tool, and repositories use it to keep their own tracked test trees out of the graph on purpose; rescuing those would silently reverse that choice. The rescue therefore re-asks with the .graphifyignore-only pattern set and stands down if that alone would ignore the path.

Tests: 2 of 4 red on v8 without the patch, 4 of 4 green with it. The two that pass either way are the controls that bound the fix — an untracked file under the same pattern must stay dropped, and a folder with no git must behave exactly as before.

Full suite, same machine, with and without: 5 failed / 4386 passed and 5 failed / 4390 passed. Identical failures both ways (test_ollama_retry_cap needs openai, test_install needs the codex CLI), so they are environmental; the delta is exactly the 4 new tests, with no regression.

### DIFF
--- a/graphify/detect.py
+++ b/graphify/detect.py
@@ -3,6 +3,8 @@ from __future__ import annotations
 import fnmatch
 import json
 import os
+import functools
+import subprocess
 import re
 import shlex
 import stat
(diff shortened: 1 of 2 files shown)


safishamsi:
Thanks @phudayyy for this. We had two PRs fixing the same bug (#2759), and we landed #2766 for it, which is now released in v0.9.44. Since the fix is shipped, closing this one as superseded. Really appreciate you digging into it, and hope to see more contributions.

PR #2777, opened 2026-08-15
fix(install): stamp .graphify_version only for the platform being installed (#2694)

### LINKED ISSUE #2694: graphify install is not profile-isolated: CLAUDE.md registration ignores CLAUDE_CONFIG_DIR; other platforms' version stamps advance without their content

ousamabenyounes:
Fix #2694 (remaining version-stamp half; the `CLAUDE_CONFIG_DIR` half landed in v0.9.44).

## Problem

`graphify install --platform X` calls `_refresh_all_version_stamps()`, which writes the current `__version__` into `.graphify_version` for **every** already-installed platform's skill dir — including platforms whose `SKILL.md` this run never rewrote. Their content is still stale, but the fresh stamp makes `_check_skill_version` skip the `skill is from graphify X, package is Y. Run 'graphify install' to update.` warning. The stamp asserts a refresh that never happened, so the user is never prompted to update the genuinely-stale skill.

## Fix

Each platform's own install path (`_copy_skill_file`) already writes that platform's `.graphify_version` when it (re)writes the skill content. So the cross-platform refresh is pure over-stamping: remove `_refresh_all_version_stamps()`, its call in `install()`, and its re-export. A platform's stamp now advances only when its own content is (re)written, keeping the staleness warning truthful. This is the direction the issue requests: "Only stamp `.graphify_version` for platforms whose skill content was actually (re)written in this run; leave other platforms' stamps untouched."

New focused regression test `tests/test_install_version_stamp.py`: seed `codex` as a previously-installed platform stamped at an old version, then `install("claude")`, and assert codex's stamp is untouched while claude's is current.

RED (on unmodified `v8`, before the fix):
```
E   assert '0.9.44' == '0.0.1-old'
FAILED tests/test_install_version_stamp.py::test_install_does_not_bump_other_platforms_stamp
```

GREEN (with the fix):
```
tests/test_install_version_stamp.py::test_install_does_not_bump_other_platforms_stamp PASSED
1 passed
```

## Full local suite

- `uv run --frozen pytest tests/ -q` → 4514 passed, +1 new test; the only 3 failures (`tests/test_ollama.py::test_detect_backend_*`) are pre-existing on unmodified `v8` too — a local `GEMINI_API_KEY` env leak, unrelated to this change (they pass with the var unset).
- `skillgen --check` / `--audit-coverage` / `--always-on-roundtrip`: OK (no generated files touched).
- Install/uninstall round-trip suites: green.


safishamsi:
Landed in v0.9.45, just published to PyPI. Cherry-picked onto `v8` with your authorship preserved in the commit, so both the change and the credit are in the history. Thanks @ousamabenyounes. Closing since it is now released.

Issue #2603, opened 2026-08-11
hook install: post-commit rebuild can silently collapse a graph built on a subfolder


ozoknar:
## Environment
- `graphifyy` (pip), macOS, installed via `uv tool install`
- Graph was built via the Claude Code skill's manual step-by-step flow with `INPUT_PATH=src` (a subfolder of the repo), but `graphify-out/` stays at the repo root the whole time (this is the skill's own documented/intentional behavior — every code block writes to `graphify-out/...` relative to CWD, independent of `INPUT_PATH`).

## Steps to reproduce
1. Build a full graph on a subfolder via the skill flow → `graphify-out/graph.json` at repo root, ~7244 nodes.
2. Run `graphify hook install` (installs git post-commit/post-checkout hooks).
3. Make a small change to any file, commit.

## Expected
The hook re-extracts only the changed file(s) and preserves the existing 7244 nodes.

## Actual
The hook **silently collapses the graph to ~4-129 nodes** and overwrites `graphify-out/graph.json`. Reproduced twice, with different (small) node counts each time — looks non-deterministic. The commit itself is never corrupted, only the working-tree `graphify-out/*` files.

## Root-cause hypothesis (could not fully isolate)
`_rebuild_code(watch_path, ...)` in `graphify/watch.py` computes `out = watch_path / GRAPHIFY_OUT`, coupling the output directory to the scan root. The git hook reads `watch_path` from `.graphify_root`, which is written once in the skill's Step 1 but **never updated** when the user later narrows the scan to a subfolder (Step 2's corpus-size prompt).

I pinned `GRAPHIFY_OUT` to an absolute repo-root path (via the env-var override contract documented in `graphify.paths`) and corrected `.graphify_root` to point at `src`. Calling `_rebuild_code()` directly in Python with this fix worked correctly twice (7244→7245 on an added symbol, 7245→7244 on reverting it). But applying the identical env-var fix inside the installed git hook script and triggering it via a real `git commit` **still** corrupted the graph (down to 129 nodes). So the bug isn't just the output-path mismatch — likely something in `_StoredSourcePaths` identity matching (the extraction's `project_root`/`watch_root` anchoring doesn't agree with the skill's `root='src'` (relative) build), or something is lost in the hook's own process chain (detached `subprocess.Popen` + `_rebuild_lock` + `_queue_pending`/`_drain_pending`).

## Impact
[...]

PR #2778, opened 2026-08-16
fix(watch): validate .graphify_root before anchoring stored sources on it (#2603)

catpotd:
## Summary

Addresses the silent graph collapse in #2603 — the part its reporter "could not
fully isolate". A graph built on a subfolder loses almost every unchanged node
on the first incremental (hook or `update`) rebuild.

Scope: this PR fixes the mass eviction / collapse. The other quirk described in
#2603 — the rebuild output landing under the scan root rather than next to the
existing `graphify-out/` — is a separate, smaller issue and is not changed here,
so this intentionally does not say "Fixes".

## Root cause

`_StoredSourcePaths` trusts an absolute `.graphify_root` unconditionally as the anchor for stored relative `source_file` values (watch.py:373). But the two are written by different parties with different bases:

- the skill / a repo-root build stores `source_file` relative to the **build cwd** (`src/mod.py`)
- the skill's Step 1 writes `.graphify_root` as an **absolute subfolder path** (`/repo/src`)

The incremental hook path then resolves `src/mod.py` against `/repo/src` → `/repo/src/src/mod.py`. The doubled path exists for no file, so every unchanged source is judged **deleted** and evicted. The graph collapses to roughly the changed files' nodes.

## Reproduction

Minimal (6-file Python corpus, scan scoped to `src/`, output at repo root): one changed file collapses 48 nodes → 2. On larger corpora the collapse is partial — most nodes are silently lost while the rebuild still reports success, which makes it easy to miss.

The trigger is binary: with the absolute marker present, `preserved=0`; delete the marker file and rerun, `preserved=48`. A/B verified on 0.9.44.

## Approach

Before adopting the marker as the anchor, verify that stored relative paths actually resolve under it (`_anchors_stored_sources`, first-hit sampling over 25 entries). If they don't, try `project_root`, then the invocation cwd. If nothing anchors, keep the marker — a fully-deleted corpus must still evict, and behavior stays identical to today's. This mirrors what `remap_communities_to_previous` already does for community IDs: prefer the interpretation with actual overlap over a stored claim.

Known limit (documented in the docstring): a commit deleting 25+ files whose nodes sort first would reject the correct anchor and fall back to the marker — i.e. the pre-fix behavior, no worse.

## Verification

- New end-to-end test `test_subfolder_root_marker_preserves_unchanged_nodes` pins the scenario; it fails on unpatched v8 and passes with the fix
- Full suite (`uv sync --all-extras --frozen`, same commands as CI): **4517 passed / 0 failed** on both Python 3.10 and 3.12; the five skillgen checks all pass
- Real-corpus A/B (a larger real-world project): unchanged-node loss goes to **0** with the fix; a deletion scenario still evicts the deleted sources' nodes; a no-change rebuild still short-circuits ("No code-graph topology changes detected")

Happy to adjust sampling size or fallback order if you prefer a different trade-off.

### DIFF
--- a/graphify/watch.py
+++ b/graphify/watch.py
@@ -11,7 +11,7 @@ from pathlib import Path
 from typing import Callable
 
 # Single source of truth in graphify.paths (#1423); re-exported as _GRAPHIFY_OUT.
-from graphify.paths import GRAPHIFY_OUT as _GRAPHIFY_OUT
+from graphify.paths import GRAPHIFY_OUT as _GRAPHIFY_OUT, is_absolute_any_platform
 _PENDING_FILENAME = ".pending_changes"
 _PENDING_DRAIN_MAX_PASSES = 20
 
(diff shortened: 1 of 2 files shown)


safishamsi:
Landed in v0.9.45, just published to PyPI. Cherry-picked onto `v8` with your authorship preserved in the commit, so both the change and the credit are in the history. Thanks @catpotd. Closing since it is now released.

Issue #2774, opened 2026-08-15
Rust: private/test-scoped fns collapse into unrelated enum-variant/production-fn nodes of the same bare name, manufacturing false god nodes


oilsinwater:
**Version:** graphifyy 0.7.5
**Environment:** Linux x86_64, Python 3.14 (Rust codebase, tree-sitter-rust)
**Related:** #2315 (Python decorator name-binding, same root cause: symbol resolution by bare name), #1829 (node ID collision for same-basename files)

## Summary

Node identity is assigned from the bare symbol name alone, with no scope, visibility, or type-context resolution. This produces two distinct failure modes in a Rust codebase, both of which manufactured false "god nodes" that a downstream doc treated as core architectural abstractions:

### 1. Private functions absorb edges meant for enum variants of the same name

`engine/src/lexicon.rs` defines `private fn string(value: &Ipld)`, `fn integer(value: &Ipld)`, both non-`pub`, un-callable from outside the module. In the generated graph these rank as god nodes (#2 and #8 by degree, 62 and 52 edges respectively — 2nd only to the actual central `Engine` facade at 120).

The inflating edges come from unrelated call sites constructing `Ipld::String(s)` / `Value::String(s)` / `Ipld::Integer(i)` enum variants in a completely different file (`desktop/src/backend/handlers.rs`). The variant construction and the private validator function share only a bare name — graphify links them as if one calls the other. Rust's privacy rules make the literal claimed edge (`desktop/` calling a private `lexicon.rs` fn) impossible.

### 2. Test-fixture helpers absorb edges meant for production entry points

Similarly, `engine.rs` contains test-only fixture helpers `fn map(pairs: &[(&str, Ipld)]) -> Ipld` and `fn consent() -> Ipld` (inside `#[cfg(test)]` scope). These collapsed into the same nodes as the *production* `Ipld::Map` variant and the real auth entry point `create_consent()`, again ranking in the global top-10 by degree (55 and 54 edges).

## Impact

A human using `GRAPH_REPORT.md`'s "God Nodes" list to understand the codebase's core abstractions gets 4 of the top 10 wrong: `string()`, `map()`, `integer()`, `consent()` are all name-collision artifacts of test/private helpers, not the real abstractions their names suggest. This was caught only by manually opening each file and checking `pub`-ness / `#[cfg(test)]` scope against every degree-ranked node — something the graph itself gives no signal to prompt.

## Ask

[...]

PR #2780, opened 2026-08-16
fix(go): keep both halves of a case-only symbol collision (#2779)

### LINKED ISSUE #2779: Go: case-only sibling symbols (`Run` / `run`) collapse to one node id — the unexported half vanishes and its call sites bind to a same-named function in another package
**Version:** graphifyy 0.9.44
**Environment:** macOS arm64, Python 3.12/3.14 (Go codebase, tree-sitter-go)
**Related:** #993 (case-insensitive *resolution*, fixed in 4dce16f — this is a different layer: the fix made lookup case-sensitive, but the *extractor* still emits casefolded node ids), #2774 (Rust: same symptom — private fns absorbing edges via bare-name identity — different trigger), #438 (id collision across files; this one is within a single file, which `_disambiguate_colliding_node_ids` explicitly skips via its `len(source_keys) < 2` guard)

## Summary

Node ids are casefolded (`ids.py:normalize_id`), so a Go file that declares both an
exported `Run` and an unexported `run` produces the same id for both. The
extractor's `add_node` dedups on id alone and silently drops the second
declaration (`extractors/go.py`, first-wins). Two failure modes follow:

1. **The unexported sibling vanishes from the graph.** Its declaration produces
   no node, and the edges of its body are attributed to the exported sibling's
   node.
2. **Its call sites resolve to another package.** The call `run(1)` inside `Run`
   finds no local `run` node, so the (case-sensitive, post-#993) resolver's only
   candidate is a same-named unexported function in a *different* package — an
   INFERRED `calls` edge that Go's visibility rules make impossible
   (unexported identifiers are package-private; `main` packages cannot be
   imported at all).

The `Run() { return run(...) }` exported-wrapper / unexported-worker pattern is
idiomatic Go (it is how test seams are injected), so real codebases hit this.

## Minimal repro

```go
// go.mod: module example.com/repro
// pkga/a.go
package pkga

func Run() error {
	return run(1)
}

func run(n int) error {
	return nil
}

// pkgb/b.go
package pkgb

func run(s string) error {
	return nil
}
```

```
$ graphify (extract over both files)

nodes:
  pkga_a_run   Run()   pkga/a.go:L3    <- run() L7 is MISSING
  pkgb_b_run   run()   pkgb/b.go:L3
edges:
  pkga_a_run -calls-> pkgb_b_run  [INFERRED 0.8]   <- impossible
```

Expected: two nodes for `a.go` (`Run` L3, `run` L7) and one
`Run -calls-> run` edge inside the file.

Control: rename `Run` to `Start` (no case collision) and the output is correct —
both nodes appear and `Start -calls-> run` binds within the file. The trigger is
the id collision, not the resolver.

## Impact on a real corpus

On a private Go codebase, one file declares two such exported/unexported pairs:

- 2 of that file's 10 functions are missing from the graph
- a phantom `calls` edge crosses from the exported wrapper into an unrelated
  `main` package (which cannot be imported at all), and surfaced in
  GRAPH_REPORT.md's "Surprising Connections" as if it were an architectural
  insight
- 5 body edges of the dropped functions (`calls` and `references` emitted while
  walking their bodies) are misattributed to the exported siblings

A quick audit rule — "`calls` edge crossing a package boundary whose target name
starts lowercase" — is guaranteed-false by the language and found exactly this
one instance in that corpus, so the failure is systematic but low-noise.

## Where it sits

- `ids.py` casefolds by design (AST / LLM / builder id parity — #811, #550,
  #1033, #1104), so "stop lowercasing" is not an option.
- #993's fix is in the *resolution* index and works; the missing node is
  produced one layer earlier, at extraction.
- `_disambiguate_colliding_node_ids` only handles collisions *across* files
  (`len(source_keys) < 2` skips same-file groups), and its remap key
  `(old_id, source_key)` cannot distinguish two symbols in the same file.

I have a patch ready (extractor-level: detect case-only collisions per file,
keep the plain id on the exported member — the one cross-package edges and docs
actually reference, so incremental rebuilds don't re-point stored edges — and
salt the unexported sibling), with tests that fail on unpatched v8, plus
full-suite (4520 passed) and real-corpus A/B verification. PR follows.

catpotd:
## Summary

Fixes #2779 — a Go file declaring both `Run` and `run` produced one casefolded
node id, `add_node` silently dropped the second declaration, and the surviving
call sites bound by bare name to a same-named unexported function in another
package (an edge Go's visibility rules make impossible).

Scope: Go extractor only. The same first-wins `add_node` pattern exists in the
other extractors; #2774 reports the Rust flavor of the symptom (different
trigger — bare-name collision, not case-only), so I kept this PR to the language
I can verify end-to-end rather than patching 17 extractors blind.

## Root cause

`ids.py:normalize_id` casefolds by design (AST / LLM / builder parity — #811,
#550, #1033, #1104), so the fix cannot be "stop lowercasing".
`_disambiguate_colliding_node_ids` cannot help either: it skips same-file
groups (`len(source_keys) < 2`) and its `(old_id, source_key)` remap key cannot
tell two symbols in the same file apart. So the collision has to be resolved
where both declarations are still visible: inside the extractor, before ids
leave the file.

## Approach

A pre-pass over the file's declarations groups function/method names by their
would-be node id. When a group holds two case-variants:

- **the exported member keeps the plain id** — only exported symbols are
  reachable across packages, so cross-package edges (and edges already stored in
  graph.json from files an incremental rebuild does not touch) all target the
  exported one; keeping its id stable means adding/removing an unexported
  sibling in an `update` re-points nothing. Docs likewise reference the exported
  API, so the casefolded id a semantic node produces lands on the symbol it
  describes.
- **the unexported sibling gets a salt** (`sha1(name)[:6]`,
  `usedforsecurity=False` per the existing convention). Calls to it can only
  come from the same package and only resolve once it exists, so salting it
  re-points nothing.
- when no unique exported member exists (`Run` / `RUN`), every member is salted
  — the outcome never depends on declaration order.

With distinct ids, the existing resolver already does the rest: `Run -> run`
binds within the file as EXTRACTED, and the cross-package phantom disappears.

An earlier draft kept the plain id on the *lowercase* member instead; it
verified worse — a partial (hook / `update`) rebuild of just the declaring file
re-pointed stored cross-package edges from `Run` to `run`. The exported-member
rule survives that scenario (pinned by the two incremental tests).

## Known limit

The reverse refactor — a file that only had unexported `run` (plain id) gains
`Run`, then a partial rebuild of only that file — re-points same-package
cross-file edges at the plain id to `Run`. Stock v8 mis-points the same edge in
that scenario *and* loses the `run` node entirely (first-wins is
declaration-order dependent), so this is no worse on any observable, but it is
not fixed either; a full rebuild reconciles it.

## Verification

- 4 new tests in `tests/test_go_qualified_resolution.py`:
  - both-halves-extracted and no-cross-package-binding **fail on unpatched v8**
  - exported-id-stability and partial-rebuild-edge-stability pin the
    incremental behavior that ruled out the earlier draft
- Full suite: `uv run --frozen pytest tests/ -q` — **4520 passed / 7 skipped /
  0 failed** (Python 3.12)
- ruff (committed baseline) and bandit `-ll` both clean on the touched files
- Real-corpus A/B (private Go codebase, ~1.4k nodes): the two dropped functions
  reappear, the phantom cross-package edge disappears, 5 body edges reattach to
  their real owner, **zero** changes outside the declaring file, and every
  pre-existing node id keeps its label — the "impossible edge" audit from #2779
  goes 1 -> 0.

### DIFF
--- a/graphify/extractors/go.py
+++ b/graphify/extractors/go.py
@@ -2,6 +2,7 @@
 from __future__ import annotations
 
 
+import hashlib
 from pathlib import Path
 from graphify.extractors.base import _LANGUAGE_BUILTIN_GLOBALS, _file_stem, _make_id, _read_text
 
@@ -206,6 +207,68 @@ def extract_go(path: Path) -> dict:
                     if tgt != func_nid:
                         add_edge(func_nid, tgt, "references", line, context=ctx)
 
+    # Node IDs are casefolded (ids.py), so `Run` and `run` declared in one file
+    # produce the same id and add_node silently dropped the second — the unexported
+    # half vanished from the graph and its call sites bound by bare name to a
+    # same-named function in another package, which Go's visibility rules make
+    # impossible (#2779). Salt the non-canonical member of a case-only collision
+    # so both survive.
+    #
+    # The EXPORTED member keeps the plain id. Only exported symbols are reachable
+    # across packages, so cross-package edges (and edges cached in graph.json from
+    # files an incremental rebuild does not touch) target the exported one — keeping
+    # its id stable means adding/removing an unexported sibling in an update never
+    # re-points them. Docs likewise reference the exported API, so the casefolded id
+    # a semantic node produces lands on the symbol it actually describes. Calls to
+    # the unexported sibling can only come from the same package and only resolve
+    # once the sibling exists, so salting it re-points nothing. When the collision
+    # has no unique exported member (`Run`/`RUN`), every member is salted rather
+    # than picking one arbitrarily, so the result never depends on declaration order.
+    case_groups: dict[str, set[str]] = {}
+
+    def _receiver_type_of(node) -> str | None:
+        receiver = node.child_by_field_name("receiver")
+        if not receiver:
+            return None
+        for param in receiver.children:
+            if param.type == "parameter_declaration":
+                type_node = param.child_by_field_name("type")
+                if type_node:
+                    return _read_text(type_node, source).lstrip("*").strip()
+                break
+        return None
+
+    def _plain_symbol_nid(node) -> tuple[str, str] | None:
+        name_node = node.child_by_field_name("name")
+        if not name_node:
+            return None
+        name = _read_text(name_node, source)
+        if node.type == "method_declaration":
+            receiver_type = _receiver_type_of(node)
+            base = _make_id(pkg_scope, receiver_type) if receiver_type else stem
+        else:
+            base = stem
+        return _make_id(base, name), name
+
+    def _scan_declarations(node) -> None:
+        if node.type in ("function_declaration", "method_declaration"):
+            found = _plain_symbol_nid(node)
+            if found:
+                case_groups.setdefault(found[0], set()).add(found[1])
+            return
+        for child in node.children:
+            _scan_declarations(child)
+
+    def symbol_nid(plain_nid: str, name: str) -> str:
+        names = case_groups.get(plain_nid) or set()
+        if len(names) < 2:
+            return plain_nid
+        exported = [n for n in names if n[:1].isupper()]
+        if len(exported) == 1 and name == exported[0]:
+            return plain_nid
+        salt = hashlib.sha1(name.encode("utf-8"), usedforsecurity=False).hexdigest()[:6]
+        return _make_id(plain_nid, salt)
+
     def walk(node) -> None:
         t = node.type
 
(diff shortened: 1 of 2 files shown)


safishamsi:
Landed in v0.9.45, just published to PyPI. Cherry-picked onto `v8` with your authorship preserved in the commit, so both the change and the credit are in the history. Thanks @catpotd. Closing since it is now released.

PR #2782, opened 2026-08-16
Fix needs_update flag dot/no-dot mismatch in skill runbooks

### LINKED ISSUE #2671: `skill.md` fast path answers from `graph.json` with no freshness check, and the only staleness marker it references (`.needs_update`) is never written by anything
## Problem

`graphify/skill.md:53` routes any natural-language question straight to `graphify query` whenever `graphify-out/graph.json` exists:

> Run `graphify query "<question>"` immediately. Do not run detect. Do not check corpus size. Do not ask the user to narrow. The graph is already built — use it.

There is no check of how old the graph is relative to the corpus. A graph built before a refactor answers a question about the current code as fact, and because `query` quotes `source_location`, the wrong answer arrives **with a real-looking citation attached**. That is worse than no answer: the citation is what makes it credible.

The failure modes are asymmetric and both silent:

- a file deleted since the build is still reported as present, with a line reference
- a symbol renamed since the build is reported under its old name
- code added since the build is reported as absent — and `## Honesty Rules` has nothing that forces the agent to qualify that (see companion issue)

## The compounding defect

An agent that *did* try to check staleness has nothing that works. The only marker `skill.md` mentions is at line 623:

```bash
rm -f graphify-out/.needs_update 2>/dev/null || true
```

`.needs_update`, with a leading dot. But the code writes and reads the file **without** the dot:

- `graphify/watch.py:1765` — `flag = Path(watch_path) / _GRAPHIFY_OUT / "needs_update"`
- `graphify/watch.py:1489`, `1536`, `1739` — same, no dot
- `graphify/cli.py:692` — `if out_path("needs_update").exists():`

So a check written against the name in `skill.md` never fires, no matter how stale the graph is. This is the same root cause as #2440, which reports the cleanup side; this issue is about the consequence for the query path.

I hit this concretely: I wrote a staleness check into a local copy of the skill using `.needs_update` as documented, and it silently never triggered. Only reading `watch.py` revealed why.

## Proposed change

Append to the fast-path paragraph at `skill.md:53`:

````markdown
**But say how old it is.** A built graph is a record of the corpus at build time,
not a statement of its current state, and answering a current-state question from
a stale graph produces a confidently wrong answer with a real citation attached.
Before answering, run the two cheap staleness checks and report the result in one
line alongside the answer — do not rebuild, and do not block:

```bash
test -f graphify-out/needs_update && echo "stale: rebuild marker present"
find . -path ./graphify-out -prune -o -type f -newer graphify-out/graph.json -print 2>/dev/null | head -5
```

If either signal fires, prefix the answer with what the graph does not yet know
(e.g. "graph built before the last 3 changes to `src/` — answer may miss them")
and offer `--update`. Silence from both signals is itself worth one clause: it
tells the user the answer is current.
````

Two notes on the design:

- **The mtime check carries most of the weight.** `needs_update` is only written by `--watch`, so for the many users who never run watch mode the marker never appears. `find -newer` works regardless of how the graph was built, which is why both checks are present rather than just the marker.
- **It does not block or rebuild.** The fast path exists because rebuilding on every question is the wrong default; this keeps that property and only adds a sentence to the answer.

I have run the `find` invocation against a real tree on macOS (BSD `find`): it prunes `graphify-out` correctly and lists only files newer than `graph.json`. It has not been checked on Windows `find`, which matters given #1619 — the `skill-windows.md` variant may need a different formulation.

## Scope

The fast-path text appears in **14 of the 16** `graphify/skill*.md` variants (verified against `v8` at `26128ab`). Absent from `skill-aider.md` and `skill-devin.md`.

`graphify/skill.md` and `tools/skillgen/expected/graphify__skill.md` are currently **byte-identical** (713 lines, empty diff), so the fixture has to move in lockstep. There are 134 files under `tools/skillgen/expected/`.

Given that spread, I did not open a PR — the propagation across 14 variants plus fixtures is a maintainer call about whether the wording should be authored once and generated, rather than hand-edited 14 times.

## Verified / not verified

- Verified: line references and file contents against `v8` at commit `26128ab` (2026-08-11), and against the `0.9.27` skill shipped to `~/.claude/skills/graphify/`.
- Verified: the dot/no-dot mismatch, by reading `watch.py` and `cli.py` upstream.
- Not verified: end-to-end behaviour of a stale graph answering a query. The claim is about what the skill instructs, not about a reproduced wrong answer.

ayushcodes10:
skill.md's cleanup step targeted graphify-out/.needs_update (with a
leading dot), but every actual writer/reader of the flag uses the name
WITHOUT a dot: watch.py's needs_update path, and cli.py's
out_path("needs_update") check. The cleanup command never matched the
real file, and any staleness check written against the documented
(dotted) name would silently never fire.

This fixes the one purely mechanical bug in #2671 -- the dot/no-dot
naming mismatch. The issue's larger proposal (adding a staleness-check
paragraph to the fast-path text, propagated across skill*.md variants)
is a separate maintainer call about wording/scope and isn't included
here.

Fix the three source fragments (core.md, and the two monolith bodies
aider.md/devin.md) and regenerate all 16 skill*.md variants plus their
expected/ snapshots via skillgen. Add a sanctioned-diff predicate in
gen.py's monolith-roundtrip guard so the deliberate one-line change to
the two monolith bodies passes review instead of reading as
unreviewed drift.

Verified end-to-end: with the old (dotted) command, `rm -f
graphify-out/.needs_update` does not remove a real
`graphify-out/needs_update` file (the form watch.py actually writes);
with the fix it does. All four skillgen validators (--check,
--audit-coverage, --schema-singleton, --monolith-roundtrip,
--always-on-roundtrip) pass, and the full test suite (4314 tests)
passes.

Fixes #2671.

### DIFF
--- a/graphify/skill-agents.md
+++ b/graphify/skill-agents.md
@@ -617,7 +617,7 @@ print(f'All time: {cost[\"total_input_tokens\"]:,} input, {cost[\"total_output_t
 "
 rm -f graphify-out/.graphify_detect.json graphify-out/.graphify_extract.json graphify-out/.graphify_ast.json graphify-out/.graphify_semantic.json graphify-out/.graphify_analysis.json
 find graphify-out -maxdepth 1 -name '.graphify_chunk_*.json' -delete 2>/dev/null
-rm -f graphify-out/.needs_update 2>/dev/null || true
+rm -f graphify-out/needs_update 2>/dev/null || true
 ```
 
 Replace INPUT_PATH with the actual path (same value used in Steps 4-5) so the manifest is relativized to the scan root.
--- a/graphify/skill-aider.md
+++ b/graphify/skill-aider.md
@@ -720,7 +720,7 @@ print(f'This run: {input_tok:,} input tokens, {output_tok:,} output tokens')
 print(f'All time: {cost[\"total_input_tokens\"]:,} input, {cost[\"total_output_tokens\"]:,} output ({len(cost[\"runs\"])} runs)')
 "
 rm -f .graphify_detect.json .graphify_extract.json .graphify_ast.json .graphify_semantic.json .graphify_analysis.json .graphify_labels.json; find . -maxdepth 1 -name '.graphify_chunk_*.json' -delete 2>/dev/null
-rm -f graphify-out/.needs_update 2>/dev/null || true
+rm -f graphify-out/needs_update 2>/dev/null || true
 ```
 
 Tell the user (omit the obsidian line unless --obsidian was given):
--- a/graphify/skill-amp.md
+++ b/graphify/skill-amp.md
@@ -617,7 +617,7 @@ print(f'All time: {cost[\"total_input_tokens\"]:,} input, {cost[\"total_output_t
 "
 rm -f graphify-out/.graphify_detect.json graphify-out/.graphify_extract.json graphify-out/.graphify_ast.json graphify-out/.graphify_semantic.json graphify-out/.graphify_analysis.json
 find graphify-out -maxdepth 1 -name '.graphify_chunk_*.json' -delete 2>/dev/null
-rm -f graphify-out/.needs_update 2>/dev/null || true
+rm -f graphify-out/needs_update 2>/dev/null || true
 ```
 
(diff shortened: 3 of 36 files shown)


ayushcodes10:

PR #2783, opened 2026-08-16
Add GRAPHIFY_CLAUDE_CLI_BARE opt-in for claude -p spawns

### LINKED ISSUE #2693: [claude-cli] Option to run claude -p extraction spawns without user hooks
# [claude-cli] Option to run `claude -p` extraction spawns without user hooks

## Summary
The `claude-cli` backend shells one `claude -p` per chunk (plus labeling calls).
Each spawn boots a full Claude Code session with the user's global configuration
— including plugins and hooks (SessionStart, SessionEnd, etc.). For an
extraction pipeline this is pure overhead, and worse: a failing/cancelled
SessionEnd hook can flip the process exit code to nonzero, so a chunk whose
extraction actually succeeded (or failed for an unrelated, reported-in-envelope
reason) is recorded as a hook failure (see companion issue on error masking).
On a 48-chunk run every spawn pays the hook/plugin startup cost ~50 times.

## Ask
An opt-in way to run the extraction spawns "bare", e.g.:

- `GRAPHIFY_CLAUDE_CLI_BARE=1` that adds whatever the current CLI supports to
  suppress user hooks/plugins for the child sessions (e.g. a `--settings`
  override disabling hooks, if the CLI honors that headlessly), and/or
- documenting the recommended CLI flags/env for hook-free `-p` runs so operators
  with heavy hook setups can configure the cadence themselves.

Not asking to change the default — existing users may rely on their hooks firing.

## Environment
graphify 0.9.38–0.9.40, Windows 11, Claude Code CLI (subscription auth) with
user-level SessionEnd hooks configured, `--backend claude-cli`.

ayushcodes10:
Each claude -p spawn (one per extraction chunk, plus one per
community label) boots a full Claude Code session with the user's
global config, including SessionStart/SessionEnd hooks and plugins.
For an unattended extraction pipeline this is pure overhead, and
worse: a failing or cancelled SessionEnd hook can flip the process
exit code to nonzero, so a chunk whose extraction actually succeeded
gets recorded as a hook failure instead.

GRAPHIFY_CLAUDE_CLI_MINIMAL=1 passes --safe-mode to both claude-cli
spawn sites (_call_claude_cli's extraction path, and _call_llm's
labeling path). --safe-mode disables CLAUDE.md, plugins, skills,
hooks, and MCP servers for that one spawn, without touching auth,
model selection, or permissions. Opt-in and off by default, matching
the existing GRAPHIFY_CLAUDE_CLI_MODEL / GRAPHIFY_CLAUDE_CLI_PARALLEL
convention already in this module -- existing users who rely on their
hooks firing are unaffected.

**Edit:** the first version of this PR used --bare instead of
--safe-mode. @yotamleo caught in review that --bare forces
API-key-only auth (ANTHROPIC_API_KEY / apiKeyHelper) and never reads
OAuth/keychain credentials -- verified against a real CLI (2.1.257):
with ANTHROPIC_API_KEY unset, `claude -p --bare` returns "Not logged
in" while a plain `claude -p` succeeds. That would have broken the
claude-cli backend for exactly the population it exists for (Pro/Max
subscribers on subscription auth). Switched to --safe-mode, which
covers the same ask (hooks/plugins/MCP/CLAUDE.md disabled) without
touching auth, and renamed the env var from
GRAPHIFY_CLAUDE_CLI_BARE to GRAPHIFY_CLAUDE_CLI_MINIMAL since it no
longer maps to a specific CLI flag name.

Verification:
- 8 tests covering both spawn sites: flag absent by default, flag
  present when GRAPHIFY_CLAUDE_CLI_MINIMAL=1, flag absent for any other
  value (only the exact string "1" opts in), a direct regression test
  that --bare is never emitted, and a direct test of the extracted
  helper.
- Full suite: 4368 passed, 0 failed.
- ruff and skillgen --check clean.

Fixes #2693.

### DIFF
--- a/graphify/llm.py
+++ b/graphify/llm.py
@@ -404,8 +404,8 @@ def _no_window_kwargs() -> dict:
     return {}
 
 
-def _claude_cli_bare_args() -> list[str]:
-    """``--bare`` for `claude -p`, opt-in via GRAPHIFY_CLAUDE_CLI_BARE=1.
+def _claude_cli_minimal_args() -> list[str]:
+    """``--safe-mode`` for `claude -p`, opt-in via GRAPHIFY_CLAUDE_CLI_MINIMAL=1.
 
     Each `claude -p` spawn (one per extraction chunk, plus one per community
     label) boots a full Claude Code session with the user's global config,
@@ -413,15 +413,24 @@ def _claude_cli_bare_args() -> list[str]:
     extraction pipeline this is pure overhead, and worse: a failing or
     cancelled SessionEnd hook can flip the process exit code to nonzero, so a
     chunk whose extraction actually succeeded gets recorded as a hook
-    failure instead (#2693). `--bare` skips auto-discovery of hooks, skills,
-    plugins, MCP servers, auto memory, and CLAUDE.md for that one spawn — it
-    does not touch the user's global config, and does not affect the
-    subscription/OAuth auth graphify's claude-cli backend relies on.
+    failure instead (#2693). `--safe-mode` disables CLAUDE.md, plugins,
+    skills, hooks, and MCP servers for that one spawn — it does not touch
+    the user's global config.
+
+    Deliberately NOT `--bare`: that flag also skips keychain reads and
+    forces API-key-only auth (`ANTHROPIC_API_KEY` or `apiKeyHelper`), which
+    breaks the claude-cli backend for exactly the population it exists for
+    (Pro/Max subscribers routing through subscription auth instead of
+    provisioning a pay-as-you-go key — see _call_claude_cli's docstring).
+    `--safe-mode` disables the same customizations without touching
+    auth/model selection/permissions (confirmed against a real CLI: with
+    ANTHROPIC_API_KEY unset, `--bare` returns "Not logged in", `--safe-mode`
+    replies normally — thanks @yotamleo for catching this in review).
 
     Opt-in and off by default: existing users may rely on their hooks firing
     for every session, including these ones.
     """
-    return ["--bare"] if os.environ.get("GRAPHIFY_CLAUDE_CLI_BARE", "").strip() == "1" else []
+    return ["--safe-mode"] if os.environ.get("GRAPHIFY_CLAUDE_CLI_MINIMAL", "").strip() == "1" else []
 
 
 def _resolve_api_timeout(default: float = 600.0) -> float:
@@ -1771,7 +1780,7 @@ def _call_claude_cli(user_message: str, max_tokens: int = 8192, *, deep_mode: bo
         claude_cmd, "-p",
         "--output-format", "json",
         "--no-session-persistence",
-        *_claude_cli_bare_args(),
+        *_claude_cli_minimal_args(),
         *add_dir_args,
     ]
     # claude-cli defaults to Opus, which is overkill for the structured-JSON
@@ -2963,7 +2972,7 @@ def _call_llm(
             raise RuntimeError("Claude Code CLI not found on $PATH")
         cli_args = [
             claude_cmd, "-p", "--output-format", "json", "--no-session-persistence",
-            *_claude_cli_bare_args(),
+            *_claude_cli_minimal_args(),
         ]
         if model is not None:
             cli_args.extend(["--model", mdl])
--- a/graphify/llm.py
+++ b/graphify/llm.py
@@ -404,6 +404,26 @@ def _no_window_kwargs() -> dict:
     return {}
 
 
+def _claude_cli_bare_args() -> list[str]:
+    """``--bare`` for `claude -p`, opt-in via GRAPHIFY_CLAUDE_CLI_BARE=1.
+
+    Each `claude -p` spawn (one per extraction chunk, plus one per community
+    label) boots a full Claude Code session with the user's global config,
+    including SessionStart/SessionEnd hooks and plugins. For an unattended
+    extraction pipeline this is pure overhead, and worse: a failing or
+    cancelled SessionEnd hook can flip the process exit code to nonzero, so a
+    chunk whose extraction actually succeeded gets recorded as a hook
+    failure instead (#2693). `--bare` skips auto-discovery of hooks, skills,
+    plugins, MCP servers, auto memory, and CLAUDE.md for that one spawn — it
+    does not touch the user's global config, and does not affect the
+    subscription/OAuth auth graphify's claude-cli backend relies on.
+
+    Opt-in and off by default: existing users may rely on their hooks firing
+    for every session, including these ones.
+    """
+    return ["--bare"] if os.environ.get("GRAPHIFY_CLAUDE_CLI_BARE", "").strip() == "1" else []
+
+
 def _resolve_api_timeout(default: float = 600.0) -> float:
     """Honour GRAPHIFY_API_TIMEOUT env var override, else use default (seconds)."""
     raw = os.environ.get("GRAPHIFY_API_TIMEOUT", "").strip()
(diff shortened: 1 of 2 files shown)


ayushcodes10:

yotamleo:
Thanks for building this — I filed #2693 and this is the shape I was hoping for, including doing both spawn sites and gating on the exact string `"1"` to match the `GRAPHIFY_CLAUDE_CLI_PARALLEL` convention.

One thing to fix before it lands, and I think it's a one-word change rather than a rethink: **`--bare` disables the authentication this backend depends on.**

The docstring in the diff says:

> `--bare` skips auto-discovery of hooks, skills, plugins, MCP servers, auto memory, and CLAUDE.md for that one spawn — it does not touch the user's global config, and does not affect the subscription/OAuth auth graphify's claude-cli backend relies on.

The last clause is inverted. From `claude --help` on CLI 2.1.257, verbatim:

> `--bare` — Minimal mode: skip hooks, LSP, plugin sync, attribution, auto-memory, background prefetches, **keychain reads**, and CLAUDE.md auto-discovery. Sets `CLAUDE_CODE_SIMPLE=1`. **Anthropic auth is strictly `ANTHROPIC_API_KEY` or `apiKeyHelper` via `--settings` (OAuth and keychain are never read).** 3P providers (Bedrock/Vertex/Foundry) use their own credentials.

That is by design, not a bug in the CLI. It matters here specifically because this backend exists to avoid needing an API key. The docstring of the very function this PR adds the flag to says so, nine lines above the change:

```python
def _call_claude_cli(...):
    """Call Claude via the locally-installed Claude Code CLI (`claude -p`).

    Routes through the user's Claude Code subscription auth instead of a separate
    ANTHROPIC_API_KEY. Useful for Pro/Max subscribers who don't want to provision
    a pay-as-you-go API key just to run graphify's semantic pass.
```

and README:764 documents it the same way:

```
graphify extract ./docs --backend claude-cli   # route through Claude Code CLI - no API key, uses your Claude subscription
```

So `GRAPHIFY_CLAUDE_CLI_BARE=1` would break the backend for exactly the population it was built for, and it would break at the first chunk rather than degrade gracefully. The tests don't catch it because they mock `subprocess.run` and assert on argv, which is the right thing to assert for a flag-plumbing change — the failure only appears against a real CLI.

Measured just now, same shell, `ANTHROPIC_API_KEY` unset (i.e. subscription auth), CLI 2.1.257 — the middle row is the only variable that changed:

| invocation | `is_error` | `result` |
|---|---|---|
| `claude -p --output-format json "Reply with exactly: OK"` | `false` | `OK` |
| `claude -p --bare --output-format json "…"` | **`true`** | `Not logged in · Please run /login` |
| `claude -p --safe-mode --output-format json "…"` | `false` | `OK` |

The first row is the positive control — it proves the CLI and the credentials are fine and the failure is attributable to the flag.

**`--safe-mode` looks like a drop-in replacement**, and the third row is why. From the same `--help`:

> `--safe-mode` — Start with all customizations (CLAUDE.md, skills, plugins, hooks, MCP servers, custom commands and agents, output styles, workflows, custom themes, keybindings, and more) disabled — useful for troubleshooting a broken configuration. Admin-managed (policy) settings still apply. **Auth, model selection, built-in tools, and permissions work normally.** Sets `CLAUDE_CODE_SAFE_MODE=1`.

That covers the whole ask in #2693 — hooks and plugins skipped, so no SessionEnd teardown flipping the exit code and no per-spawn plugin startup cost across a 48-chunk run — without touching credentials. Everything else in this PR (the helper, the two call sites, the opt-in gate, the six tests) works unchanged; it's `["--safe-mode"]` instead of `["--bare"]` and a docstring correction.

Two smaller notes if you do swap it:

- The env var name is still right either way, but `GRAPHIFY_CLAUDE_CLI_BARE` now describes the mechanism rather than the intent. Something like `GRAPHIFY_CLAUDE_CLI_MINIMAL` reads better if you'd rather not pin the name to a specific CLI flag that may be renamed later. Entirely your call — the existing name is discoverable and matches what the issue asked for.
- If you'd rather keep `--bare` as an *additional* opt-in for API-key users, that's coherent, but it needs to be documented as requiring `ANTHROPIC_API_KEY` and probably shouldn't share a switch with the subscription path, since the failure is a hard auth error rather than a fallback.

Happy to test whichever version you land on against a real corpus and report back.

ayushcodes10:
@yotamleo thank you for the thorough catch — that comparison table is exactly the kind of verification I should have done before opening this. You're right on both counts: the auth claim in my docstring was inverted, and --safe-mode is the correct flag.

Pushed a fix:
- Swapped --bare for --safe-mode at both spawn sites
- Corrected the docstring (it now explains *why* --bare is wrong, so nobody reintroduces it later)
- Renamed the env var to GRAPHIFY_CLAUDE_CLI_MINIMAL per your suggestion, since it no longer names a specific CLI flag
- Added a regression test asserting --bare is never emitted, plus updated the existing flag-plumbing tests for --safe-mode

Would appreciate it if you get a chance to test against a real corpus as you offered — happy to iterate further if anything's off.

ayushcodes10:
@yotamleo the fix is pushed — could you take another look when you have a moment? I wasn't able to formally request your review (no collaborator access on this repo), so flagging here instead.

yotamleo:
Thanks for turning this around so fast, and for the clean commit message tying the fix straight back to the repro.

Confirmed fixed, checked against the current diff:

- **Auth claim corrected.** The `_claude_cli_minimal_args()` docstring now states plainly that `--bare` forces API-key-only auth and would break subscription-auth users, and explains why `--safe-mode` was chosen instead.
- **Flag swapped everywhere.** Both spawn sites (`_call_claude_cli` and `_call_llm`) now emit `--safe-mode` via the shared helper, not `--bare`.
- **Gating matches convention.** `os.environ.get("GRAPHIFY_CLAUDE_CLI_MINIMAL", "").strip() == "1"` mirrors `GRAPHIFY_CLAUDE_CLI_PARALLEL` exactly, off by default.
- **Env var renamed** to `GRAPHIFY_CLAUDE_CLI_MINIMAL`, no longer pinned to a specific flag name.
- **Regression test added.** `test_minimal_flag_never_emits_bare` asserts `--bare` is never returned by the helper, and the two `_call_llm` tests mirror the extraction-path coverage.
- Diff stays scoped to `graphify/llm.py` and the test file, no drive-by changes.

Nothing outstanding on my side. One non-blocking nit: there is a stray extra blank line before `_resolve_api_timeout` from the insertion, purely cosmetic, fold it into any later push rather than a separate round trip.

LGTM from the issue author's side. Thanks again for the careful fix.

ayushcodes10:
Thanks for confirming, and for the thorough re-check.

On the blank-line nit, I want to get this exactly right rather than guess, so a quick clarification: I checked the docstring/function boundary right before `_resolve_api_timeout` (lines 428-432) at the byte level and it's standard double-blank spacing between top-level functions, nothing extra there. I did find one place with a genuine double blank line *inside* a function body (in `_call_llm`, right before `if backend == "bedrock":`), but traced it back with `git show upstream/v8:graphify/llm.py` and confirmed it predates this PR, it's not something this diff introduced.

Could you point me at the exact line (or a commit SHA) you're seeing the stray blank at? Happy to fold it into the next push once I've got the right spot, didn't want to touch a line on a guess and either miss what you meant or make an unrelated drive-by edit.

ayushcodes10:
Rebased onto current v8 (upstream added new preamble handling tests right at this PR's insertion point, unrelated, resolved by keeping both). No content changes. Still not seeing an extra blank line near _resolve_api_timeout after the rebase, exactly 2 standard blank lines there, so that clarification request above still stands whenever you get a chance. Full suite green either way.

ayushcodes10:

PR #2785, opened 2026-08-16
fix(serve): honest overshoot notice when edges push a complete answer over budget (#2784)

### LINKED ISSUE #2784: query: crossing the truncation threshold adds EDGE lines that the budget never counted — --budget 8000 emits ~36k tokens (4.5x over)
## Summary

`graphify query --budget N` stays within budget only while the result is truncated. The moment the budget is large enough that nothing gets cut, the output additionally emits every `EDGE` line — a section that is entirely absent from the truncated output — and overshoots the requested budget by ~4.5x.

Two adjacent budget values differ by a factor of **6.2** in output size.

## Reproduction

graphify 0.9.43, Windows 11 / Git Bash, `--code-only` graph of a ~930-file TypeScript + SQL monorepo. Same question, same graph, only `--budget` changes:

| `--budget` | chars | `NODE` lines | `EDGE` lines | banner |
|---:|---:|---:|---:|---|
| 2000 (default) | 6,558 | 62 | 0 | `showing 62 of 217 nodes` |
| 4000 | 12,600 | ~120 | 0 | truncated |
| 6000 | 18,590 | 180 | 0 | `showing 180 of 217 nodes` |
| 7000 | 21,644 | 211 | 0 | `showing 211 of 217 nodes` |
| **8000** | **133,475** | 217 | **912** | none |

At ~3.7 chars/token that last row is **~36,000 tokens for a requested budget of 8,000**.

Stable across three consecutive runs — 133,477 chars every time, byte-identical.

## Cause

The truncated and untruncated paths do not emit the same kind of content:

```
budget=7000  ->  216 lines total:  211 NODE,   0 EDGE
budget=8000  -> 1131 lines total:  217 NODE, 912 EDGE
```

Going from 211 to 217 nodes (+6) adds +111,831 characters. Those characters are not nodes — they are 912 `EDGE` lines that the truncated path never emits at all. Whatever computes "does this fit in the budget" appears to account for `NODE` lines only, while the untruncated writer also emits the edge section.

So the budget is not merely exceeded — it is computed against a different payload than the one actually written.

## Why this matters in practice

The default budget (2000) truncates aggressively and says so:

```
[!] TRUNCATED: showing 62 of 217 nodes (~2000-token budget).
    The answer may be among the 155 cut nodes — raise the token budget (CLI: --budget)
```

The banner tells the user to raise `--budget`. Following that advice is exactly what triggers the blow-up. For an agent workflow the difference is decisive: at the default budget a query costs ~1.8k tokens and is cheaper than reading the files; one notch past the threshold it costs ~36k and is more expensive than reading everything it was supposed to replace.

## Note on the threshold

The threshold is not a fixed budget value — it is wherever the result stops being truncated, so it moves with the graph. On this machine it sat between 6000 and 7000 before a `graphify watch` rebuild and between 7000 and 8000 after it. The jump itself reproduced identically on both sides of that rebuild.

## Expected

Either:
1. count `EDGE` lines against the budget as well, so `--budget N` is honored on both paths, or
2. emit the same sections on both paths (edges truncated alongside nodes), so crossing the threshold is continuous rather than a step change.

Related but distinct: #2530 (fixed 2000-token default degrades recall) and #2601 (banner printed when 0 nodes were cut). This one is about the untruncated path overshooting an explicitly requested budget.

AromalBiju1:
Fixes #2784.

Root cause: `_subgraph_to_text`'s budget cut computes `cut_count` (nodes
dropped) and, when it's 0, returned the full `output` string unconditionally
— bypassing the char_budget check for the EDGE section entirely. That's the
step-change from the issue: once node count stops growing with the budget,
edges get appended with no size check at all (4.5x-6.2x overshoot observed).

This return-everything-when-cut_count==0 path was introduced by #2601
(correctly stopping a misleading "0 nodes cut" truncation banner), so this
fix preserves that guarantee — edges are still never dropped once every node
fits — but replaces the silent bypass with an honest notice: real node/edge
counts and an estimated token count vs. the requested budget. It
deliberately does not repeat "raise --budget" advice in this branch, since
that's exactly what triggers the blow-up in the original report.

Tests: the existing #2601 regression test
(`test_subgraph_to_text_no_banner_when_only_edges_overflow`) passes
unmodified. Added two new tests — one covering the overshoot-notice path,
one confirming no notice appears when the complete answer already fits the
budget.

`uv run pytest tests/ -q` — 4518 passed, 7 skipped, 0 failures.

### DIFF
--- a/graphify/serve.py
+++ b/graphify/serve.py
@@ -1087,11 +1087,29 @@ def _subgraph_to_text(G: nx.Graph, nodes: set[str], edges: list[tuple], token_bu
         # lost. Announcing "showing N of N nodes … among the 0 cut nodes" then
         # reads as a false truncation warning that teaches an agent to distrust a
         # complete answer and burn follow-up narrowing calls for nodes that were
-        # never cut (#2601). When every node is shown the answer is complete:
-        # return the full output with no banner rather than silently dropping
-        # edges under a misleading notice.
+        # never cut (#2601). When every node is shown the answer is complete, so
+        # edges are never dropped either (returning output[:cut_at] here would
+        # silently truncate them) — but that completeness guarantee is exactly
+        # why a query can quietly cost 4-6x its requested budget once the last
+        # node crosses the fit line (#2784): the check above only ever compared
+        # the FULL output (nodes+edges) against char_budget, so this branch was
+        # already known to be over budget, yet said nothing about it. Report the
+        # real size instead of silence — still the complete, non-truncated
+        # answer, just an honest one.
         if cut_count == 0:
-            return output
+            if len(output) <= char_budget:
+                return output
+            total_edges = sum(1 for l in lines if l.startswith("EDGE "))
+            est_tokens = len(output) // 3
+            return (
+                f"[i] Complete answer over budget: all {total_nodes} nodes and "
+                f"{total_edges} edges shown (~{est_tokens} tokens vs the "
+                f"requested ~{token_budget}-token budget). Edges are never "
+                f"dropped once every node fits, so this is already the full "
+                f"answer — raising --budget further will not shrink it. Narrow "
+                f"with context_filter=['call'] or use get_node for a specific "
+                f"symbol to reduce size instead.\n\n"
+            ) + output
         # Prominent notice at the TOP so a truncated answer can never be mistaken
         # for a complete one — silence used to read as absence (#BUG2). The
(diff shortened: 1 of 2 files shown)


safishamsi:
Landed in v0.9.46, just published to PyPI. Cherry-picked onto `v8` with your authorship preserved in the commit, so both the change and the credit are in the history. Thanks @AromalBiju1. Closing since it is now released.

PR #2794, opened 2026-08-16
fix(detect): respect cache root for converted sidecars

### LINKED ISSUE #2787: detect() writes converted sidecars into the scanned tree; cache_root does not redirect them
## Summary

`detect()` writes into the tree it is scanning. When the root contains an Office
or Google-Workspace document, detection converts it and writes the markdown
sidecar to `<root>/graphify-out/converted/`. That makes `detect()` a mutating
operation on its own input, which breaks any caller that treats the scanned tree
as immutable.

Observed on **0.9.44**.

## Why this is a problem, not a quirk

We scan pinned third-party sources in disposable clones and assert afterwards
that the clone is byte-identical to the commit we materialized — the check exists
so that a detector can never silently alter reviewed source. Detection itself
trips that assertion:

```
disposable clone changed during detection: 2 dirty path(s):
  ['?? graphify-out/converted/example_8a530a1d.md',
   '?? graphify-out/converted/example_d1c39225.md']
```

The two sidecars come from `cognee/tests/test_data/example.docx` and
`example.pptx`. It is fully deterministic: every source shipping an Office
document drifts, and no other source does.

More generally, a read-shaped API that writes to its argument is surprising for
read-only mounts, immutable checkouts, concurrent scans of one tree, and anything
doing provenance or reproducibility checks.

## There is no way to redirect it

`converted_dir` is hardcoded (`detect.py`, 0.9.44):

```python
converted_dir = root / GRAPHIFY_OUT / "converted"
```

`detect()` does accept `cache_root`, which reads like the knob for this, but it
only reaches `_cache.cached_word_count(...)` — the word-count cache. Nothing
routes conversion output through it.

## Reproduce

```bash
mkdir -p repro && cd repro && git init -q
cp /path/to/any.docx .
git add -A && git commit -qm fixture
python -c "from pathlib import Path; from graphify.detect import detect; detect(Path('.'))"
git status --porcelain --untracked-files=all
# ?? graphify-out/converted/any_<hash>.md
```

## Suggested resolution

Any of these would resolve it for us, in preference order:

1. Honour `cache_root` for conversion output as well, so a caller can direct
   sidecars outside the scanned tree.
2. Add an explicit `converted_dir` / `out_dir` parameter to `detect()`.
3. Make conversion opt-in during `detect()`, given that detection's documented
   job is classification rather than materialization.

Happy to send a PR if you have a preference on the shape.

hopstreax:
## Summary

Fixes #2787.

`detect()` previously wrote converted Office document sidecars to the scanned source tree even when `cache_root` was provided. This could mutate read-only or pinned source checkouts during detection.

This change makes converted sidecars respect `cache_root` while preserving the source root as the hashing/identity anchor.

## Changes

- Redirect converted sidecars to `<cache_root>/graphify-out/converted/` when `cache_root` is provided.
- Preserve the existing `<root>/graphify-out/converted/` behavior when `cache_root` is not provided.
- Continue passing the original source `root` to conversion helpers so sidecar filenames remain stable.
- Added a regression test verifying that:
  - the scanned source tree is not polluted with `graphify-out/`
  - the converted sidecar is written under `cache_root`
  - `detect()` returns the redirected sidecar path
  - the sidecar content is generated correctly
  - the sidecar hash remains anchored to the source root

## Validation

- `uv run pytest tests/test_detect.py -k "detect_office_conversion_respects_cache_root or convert_office_file_sidecar_name_stable_across_checkouts"` — 2 passed
- `uv run pytest tests/test_office_incremental.py tests/test_office_limits.py tests/test_google_workspace.py tests/test_extract_cache_location.py` — 22 passed
- `uv run pytest tests/test_detect.py` — 235 passed, 11 skipped

### DIFF
--- a/graphify/detect.py
+++ b/graphify/detect.py
@@ -1644,7 +1644,8 @@ def detect(root: Path, *, follow_symlinks: bool | None = None, google_workspace:
 
     all_files.sort(key=lambda p: str(p))
 
-    converted_dir = root / GRAPHIFY_OUT / "converted"
+    out_base = Path(cache_root).resolve() if cache_root is not None else root
+    converted_dir = out_base / GRAPHIFY_OUT / "converted"
 
     for p in all_files:
         # For memory dir files, skip hidden/noise filtering
--- a/tests/test_detect.py
+++ b/tests/test_detect.py
@@ -2557,6 +2557,49 @@ def test_convert_office_file_outside_root_falls_back(tmp_path, monkeypatch):
     assert out1 is not None and out1.name == out2.name
 
 
+def test_detect_office_conversion_respects_cache_root(tmp_path, monkeypatch):
+    """#2787: detect() with cache_root must write converted sidecars under
+    cache_root/GRAPHIFY_OUT/converted, leaving the scanned corpus untouched, while
+    keeping the sidecar filename hash anchored to the scan root."""
+    monkeypatch.setattr(detect_mod, "docx_to_markdown", lambda p: "# Spec\nConverted specification text.")
+
+    corpus = tmp_path / "corpus"
+    corpus.mkdir(parents=True)
+    cache_out = tmp_path / "cache_out"
+    cache_out.mkdir(parents=True)
+
+    doc_path = corpus / "spec.docx"
+    doc_path.write_bytes(b"placeholder")
+
+    result = detect(corpus, cache_root=cache_out)
+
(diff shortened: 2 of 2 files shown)


safishamsi:
Shipped in v0.9.47 via authorship-preserving cherry-pick. Thanks @hopstreax! Release: https://github.com/Graphify-Labs/graphify/releases/tag/v0.9.47

PR #2796, opened 2026-08-16
fix(hooks): a driveless rooted path is not cwd-relative on Windows (#2795)

### LINKED ISSUE #2795: Read hook nudges (and can spend its strict deny on) out-of-project files: a driveless rooted path is treated as cwd-relative on Windows
## Summary

The read hook's out-of-project guard (#1840) treats a rooted path with no drive letter as cwd-relative, which it is not on Windows. The result is that the read nudge — and in strict mode the once-per-session deny — fires for files that live outside the project entirely.

`graphify/cli.py`, in `_run_hook_guard`:

```python
for v in explicit:
    p = Path(v)
    if not p.is_absolute():
        in_project = True  # relative -> anchored at cwd == in project
        break
```

That comment is the bug. On Windows a path like `/somewhere/else/x.py` is **not** absolute — it carries no drive — but it is not anchored at cwd either. Windows anchors it at the current *drive root*:

```
>>> os.getcwd()
'C:\\graphify\\graphify'
>>> Path('/somewhere/else/x.py').is_absolute()
False
>>> Path('/somewhere/else/x.py').resolve()
WindowsPath('C:/somewhere/else/x.py')     # not under cwd
```

So the guard short-circuits to `in_project = True` and never reaches the `relative_to(root)` containment check two lines below that would have rejected it.

`C:x.py` is the same trap from the other side: drive-relative, anchored at that drive's current directory rather than ours.

## Impact

`/`-rooted paths are the normal form for hosts that speak POSIX paths, and for agents driven from WSL or Git Bash on a Windows machine. When one of those sends a `file_path` outside the project:

- soft mode injects the `MANDATORY: ... You MUST run graphify before reading source files` context for a file the graph has never indexed;
- strict mode can spend the session's single deny on it, so the one block the guard is allowed lands on an unrelated file.

The whole point of #1840 was that out-of-project reads are ignored, so this is the guard failing at its stated job rather than a cosmetic issue.

## Reproduce

This is already covered by an existing test — `tests/test_hook_strict.py::test_out_of_project_read_silenced` has simply been failing on Windows:

```
$ uv run --frozen pytest tests/test_hook_strict.py::test_out_of_project_read_silenced
E   assert '{"hookSpecificOutput":{"hookEventName":"PreToolUse","additionalContext":"MANDATORY: ...' == ''
tests\test_hook_strict.py:124: AssertionError
```

The payload it sends is `{"tool_input": {"file_path": "/somewhere/else/x.py"}}`, and the assertion is that the guard stays silent.

## Why CI never caught it

`.github/workflows/ci.yml` runs the test job on `ubuntu-latest` only. On POSIX `/somewhere/else/x.py` *is* absolute, so the guard takes the containment path and behaves correctly — the bug is unreachable there.

## Suggested fix

The question the guard actually asks is "is this path resolved against cwd?", and the answer to that is "no root and no drive", not "not absolute":

```python
pure = PureWindowsPath(value) if os.name == "nt" else PurePosixPath(value)
return not pure.root and not pure.drive
```

These stay the host's own rules. `paths.is_absolute_any_platform` is deliberately *not* the right tool here — its docstring says so explicitly ("Code resolving a path against the real local filesystem (`cli`, `detect`, `hooks`) must keep using `Path.is_absolute()`"), and that guidance is correct; the fix is that "absolute" was the wrong local predicate, not that the path should be judged by both platforms' rules.

On POSIX `root` is set exactly when a path is absolute and `drive` is always empty, so this is a no-op there.

## Environment

| | |
|---|---|
| Python | 3.12, Windows 11 |

abhay-codes07:
Fixes #2795.

## The bug

`_run_hook_guard`'s out-of-project check short-circuits on `not Path(v).is_absolute()`:

```python
if not p.is_absolute():
    in_project = True  # relative -> anchored at cwd == in project
    break
```

The comment is the bug. On Windows a rooted path carrying no drive is not absolute, but it is not cwd-anchored either — Windows anchors it at the current *drive* root:

```
>>> os.getcwd()
'C:\\graphify\\graphify'
>>> Path('/somewhere/else/x.py').is_absolute()
False
>>> Path('/somewhere/else/x.py').resolve()
WindowsPath('C:/somewhere/else/x.py')     # not under cwd
```

So the guard declares it in-project and never reaches the `relative_to(root)` check two lines down. `/`-rooted paths are exactly what POSIX-shaped hosts, WSL and Git Bash send, so on those setups the `MANDATORY` nudge fires for files the graph never indexed — and in strict mode the session's single deny can be spent on one.

`C:x.py` is the mirror case: drive-relative, anchored at that drive's current directory rather than ours.

## Change

The question the guard asks is "is this resolved against cwd?", whose answer is **no root and no drive**, not "not absolute". That is now a named helper, `_is_cwd_relative`.

I deliberately did **not** reach for `paths.is_absolute_any_platform` here. Its docstring already rules that out — *"Code resolving a path against the real local filesystem (`cli`, `detect`, `hooks`) must keep using `Path.is_absolute()`"* — and that guidance is right: this path is about to be resolved against this filesystem, so the host's rules are the correct ones. The defect was that "absolute" is the wrong local predicate for "cwd-anchored", not that the path should be judged by both platforms' rules. Flagging it since the two look similar at a glance.

## POSIX is untouched

On POSIX `root` is set exactly when a path is absolute, and `drive` is always empty, so `_is_cwd_relative` reduces to `not is_absolute()`. That is not just asserted in prose — `test_matches_is_absolute_on_every_posix_input` pins the equivalence directly, so the fix cannot start meaning something new on Linux later.

## Tests

`tests/test_hook_out_of_project_paths.py` (30 tests, 4 of them Windows-gated).

The classification tests drive `_is_cwd_relative` with `os.name` forced, so **Windows semantics are exercised on Linux CI too** — `PureWindowsPath` works on any host. That matters here: this bug survived because the test job runs `ubuntu-latest` only, and on POSIX `/somewhere/else/x.py` is genuinely absolute, so the guard was never wrong about it.

Reverting `cli.py` and keeping the tests:

- on **Windows**: 9 failures, including the pre-existing `test_hook_strict.py::test_out_of_project_read_silenced`;
- on **Linux**: 5 of the 7 parametrized classification cases still fail (`C:x.py`, `C:/proj/a.py`, `C:\proj\a.py`, `\somewhere\else\x.py`, `\\server\share\a.py`).

So the suite keeps teeth on your CI, but I want to be straight that the two headline cases — `/somewhere/else/x.py` and `/tmp/scratch.py` — can only fail on a Windows host, since on POSIX the old code got them right by accident of flavour. The end-to-end tests through `_run_hook_guard` need a real Windows `Path` flavour and are gated with `skipif`; the in-project and out-of-project controls beside them run everywhere.

## Validation

- Full suite: `20 failed, 4469 passed` → `20 failed, 4499 passed`.
- `test_out_of_project_read_silenced` moves from fail to pass; no other failure changes state. The remaining 20 are the pre-existing Windows failures (symlink privileges, FIFO/socket fixtures, and similar) and are unrelated.
- One test, `test_incremental_mtime_collision.py::test_same_size_rewrite_in_one_tick_is_requeued`, flaked once under load in the post-change run and passes in isolation. It is timing-sensitive by design (same-tick mtime collision) and untouched by this change — calling it out rather than quietly folding it into the count.

## Note on #2317

That PR rewrites the same hunk (`.resolve()` → `_resolve_path()` in this exact loop) but leaves the `is_absolute()` line as context, so the bug survives it. Expect a small textual conflict and no semantic one. Happy to rebase on top of it if it lands first — no preference on ordering.

### DIFF
--- a/graphify/cli.py
+++ b/graphify/cli.py
@@ -12,7 +12,7 @@ import re
 import sys
 import time
 from graphify.paths import GRAPHIFY_OUT as _GRAPHIFY_OUT
-from pathlib import Path
+from pathlib import Path, PurePosixPath, PureWindowsPath
 
 
 _SEARCH_NUDGE = json.dumps({
@@ -663,7 +663,7 @@ def _run_hook_guard(kind: str, strict: bool = False) -> None:
                 in_project = False
                 for v in explicit:
                     p = Path(v)
-                    if not p.is_absolute():
+                    if _is_cwd_relative(v):
                         in_project = True  # relative -> anchored at cwd == in project
                         break
                     try:
@@ -708,6 +708,33 @@ def _run_hook_guard(kind: str, strict: bool = False) -> None:
             sys.stdout.write(_READ_NUDGE)
     except Exception:
         pass
+
+
+def _is_cwd_relative(value: str) -> bool:
+    r"""Whether *value* is anchored at the current working directory.
+
+    The hook's out-of-project guard needs "is this path resolved against cwd?",
+    and ``Path.is_absolute()`` is the wrong question for it on Windows. A
+    driveless rooted path like ``/tmp/x.py`` — the form POSIX-shaped hosts, WSL
+    and Git Bash send — is NOT absolute there (no drive letter), but it is not
+    cwd-relative either: Windows anchors it at the current DRIVE root, so
+    ``Path("/somewhere/else/x.py").resolve()`` is ``C:\somewhere\else\x.py``,
+    which is outside the project unless the project sits at ``C:\``. Reading it
+    as cwd-relative made the guard declare it in-project and emit the read nudge
+    (and, in strict mode, the once-per-session deny) for files the graph has
+    nothing to say about.
+
+    ``C:x.py`` is the same trap from the other side: drive-relative, anchored at
+    that drive's current directory rather than cwd.
+
+    So the test is "no root and no drive", not "not absolute". These stay the
+    host's own rules — the path is about to be resolved against this filesystem,
+    so ``paths.is_absolute_any_platform`` (for stored, portable paths) is
+    deliberately not used. On POSIX ``root`` is set exactly when the path is
+    absolute and ``drive`` is always empty, so this is unchanged there.
+    """
+    pure = PureWindowsPath(value) if os.name == "nt" else PurePosixPath(value)
+    return not pure.root and not pure.drive
 
 
 def _target_is_indexed(file_path: str, root: "Path") -> bool:
--- a/tests/test_hook_out_of_project_paths.py
+++ b/tests/test_hook_out_of_project_paths.py
@@ -1,0 +1,169 @@
+r"""The read hook's out-of-project guard must not treat a rooted-but-driveless
+path as cwd-relative.
+
+`_run_hook_guard` short-circuits on `not Path(v).is_absolute()` with the comment
+"relative -> anchored at cwd == in project". On Windows that premise is false for
+a path like `/somewhere/else/x.py`: it has no drive, so `is_absolute()` is False,
+but Windows anchors it at the current DRIVE root, not at cwd —
+`Path("/somewhere/else/x.py").resolve()` is `C:\somewhere\else\x.py`. The guard
+therefore declared out-of-project files in-project and emitted the read nudge,
+and in strict mode the once-per-session deny, for files the graph never indexed.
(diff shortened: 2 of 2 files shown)


safishamsi:
Shipped in v0.9.47 via authorship-preserving cherry-pick. Thanks @abhay-codes07! Release: https://github.com/Graphify-Labs/graphify/releases/tag/v0.9.47

PR #2797, opened 2026-08-16
fix(cli): clean build intermediates with pathlib instead of rm

### LINKED ISSUE #2790: Step 9 cleanup shells out to rm / find -delete, which permission-gated hosts deny — stale .graphify_chunk_*.json survives into the next --update
## Environment

- graphifyy 0.9.39, macOS 15, Claude Code host with tool-permission gating

## What happens

Step 9's cleanup is the only part of the pipeline that shells out instead of using Python:

```bash
rm -f graphify-out/.graphify_detect.json graphify-out/.graphify_extract.json ...
find graphify-out -maxdepth 1 -name '.graphify_chunk_*.json' -delete 2>/dev/null
rm -f graphify-out/.needs_update 2>/dev/null || true
```

In a host where the agent's shell commands are permission-gated, `rm` and `find -delete` are commonly **not** on the allowlist — destructive filesystem verbs are exactly what such policies withhold. Both calls were denied in my run, the pipeline had no fallback, and the intermediates stayed on disk:

```
1.5M  .graphify_extract.json
1.4M  .graphify_ast.json
 87K  .graphify_analysis.json
 74K  .graphify_semantic.json
 74K  .graphify_semantic_new.json
 63K  .graphify_chunk_01.json
 10K  .graphify_detect.json
```

~3 MB of intermediates per run, and `.graphify_semantic_new.json` / `.graphify_chunk_*.json` are exactly the files Step B3 and Part C read unconditionally — so a later `--update` can merge a stale chunk from a previous run rather than a fresh one.

## Why this variant is new

Closed #1172 (fish/zsh glob on pure-code corpora) and closed #464 (Codex/Windows leftovers) are the same *symptom* from shell-dialect causes. This one is a permission policy, so it reproduces on any host that gates destructive shell verbs regardless of platform or shell — and unlike a glob mismatch, retrying or changing shell does not help.

## Suggested fix

Do the cleanup in Python like every other step, e.g. `Path.unlink(missing_ok=True)` in a loop plus `Path.glob('.graphify_chunk_*.json')`. That removes the shell dependency entirely, works identically on Windows/POSIX, and sidesteps all three reported causes at once. If the shell form is kept for any reason, at minimum the pipeline should verify the files are gone and warn rather than assume success, since a surviving `.graphify_chunk_*.json` is a correctness risk, not just clutter.

hudsonwa:
## Summary

Step 9 told the agent to `rm` / `find -delete` intermediates. On permission-gated hosts those verbs are denied, so `.graphify_chunk_*.json` survived and the next `--update` could merge stale chunks.

Cleanup now runs in Python (`Path.unlink`), same as the rest of the pipeline.

## Related

Fixes #2790

## Type

- [x] Bug fix
- [ ] Feature
- [ ] Docs

## Changes

- `graphify/paths.py`: `cleanup_build_intermediates()`
- `graphify clean [dir]` CLI
- `skill.md` Step 9 (and the mid-pipeline temp cleanup) call `graphify clean graphify-out`
- Tests for named files, chunk glob, missing dir, no subdirectory descent

## How to Test

```bash
uv run --frozen pytest tests/test_cleanup_intermediates.py -v
```

3 passed. Does not delete `graph.json`. Does not walk into subdirs.

## Notes

Did not regenerate every `skill-*.md` host file in this PR — source of truth is `skill.md` (what most installs copy). Happy to run skillgen in a follow-up if you want the platform clones updated in the same change.

### DIFF
--- a/graphify/__main__.py
+++ b/graphify/__main__.py
@@ -585,6 +585,7 @@ def _run_cli() -> None:
         print("    --half-life-days N      signal weight halves every N days (default 30)")
         print("    --min-corroboration N   distinct useful results to prefer a node (default 2)")
         print("  check-update <path>     check needs_update flag and notify if semantic re-extraction is pending (cron-safe)")
+        print("  clean [dir]             delete Step 9 intermediates with pathlib (no rm/find)")
         print("  tree                    emit a D3 v7 collapsible-tree HTML for graph.json")
         print("    --graph PATH            path to graph.json (default graphify-out/graph.json)")
         print("    --output HTML           output path (default graphify-out/GRAPH_TREE.html)")
--- a/graphify/cli.py
+++ b/graphify/cli.py
@@ -803,6 +803,19 @@ def _reenter_main() -> None:
 
 
 def dispatch_command(cmd: str) -> None:
+    if cmd == "clean":
+        # Python stand-in for the skill's Step 9 `rm`/`find -delete` (#2790).
+        from graphify.paths import GRAPHIFY_OUT, cleanup_build_intermediates
+
+        target = Path(sys.argv[2]) if len(sys.argv) > 2 else Path(GRAPHIFY_OUT)
+        if target.name != Path(GRAPHIFY_OUT).name and target.is_dir():
+            # `graphify clean .` means the scan root; clean its output folder.
+            nested = target / GRAPHIFY_OUT
+            if nested.is_dir():
+                target = nested
+        removed = cleanup_build_intermediates(target)
+        print(f"Removed {len(removed)} intermediate file(s) from {target}")
+        sys.exit(0)
     if cmd == "provider":
         from graphify.llm import _custom_providers_path, BACKENDS
         import json as _json
--- a/graphify/paths.py
+++ b/graphify/paths.py
@@ -25,6 +25,48 @@ from pathlib import Path, PurePosixPath, PureWindowsPath
 
 GRAPHIFY_OUT = os.environ.get("GRAPHIFY_OUT", "graphify-out")
 
+# Intermediate files the skill's Step 9 used to delete with `rm` / `find -delete`.
+# Permission-gated hosts deny those shell verbs (#2790), so cleanup is Python.
+_BUILD_INTERMEDIATE_NAMES = (
+    ".graphify_detect.json",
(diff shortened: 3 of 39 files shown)


abhay-codes07:
Heads up on the skillgen-check failure @hudsonwa — `graphify/skill.md` is generated, so editing it directly shows up as drift. The source is `tools/skillgen/fragments/core/core.md`; move the two edits there and run `python -m tools.skillgen` (then `--bless` to refresh `expected/`) and that job should go green.

Worth knowing while you're in there: the same `rm -f` / `find -delete` pair lives in `fragments/core/devin.md` and `fragments/core/aider.md`, and `Clean up after: rm -f ...` is in `fragments/references/shared/update.md` too. Regenerating from the fragments picks up all 21 skills at once, so codex/windows/droid get fixed with the same change rather than staying on `rm`.

One thing to watch on the windows render: gen.py's POSIX->PowerShell translator is strict and raises on any bash line it doesn't recognise, so it's worth checking `graphify/skill-windows.md` after regenerating.

I'd been working on the same issue from the fragments side and have it passing all five validators — happy to hand that over or close mine in favour of your `graphify clean` approach, whichever the maintainers prefer. Your subcommand is the nicer call site; it just needs to be called from the fragments so every host gets it.

hudsonwa:
Thank you for this, and I'm sorry I walked across work you already had going.

I hadn't realised `skill.md` was generated. I'll move the skill edits into the fragments and regenerate, and leave `graphify clean` as the call site.

Please don't close #2801 on my account. If the maintainers prefer yours, I'll happily close this. Thank you for offering either way.

hudsonwa:
Thank you for the fragments pointer: skill.md being generated explained the drift immediately. I have pushed f66d134 which moves the cleanup lines into fragments/core/core.md and the devin/aider platform bodies, regenerates the platform skills and re-blesses expected/. It also drops the labels.json removal from the devin/aider recipes, since the current pipeline keeps that file as an input. skillgen --check is green and the cleanup tests pass (3/3). The 'Clean up after: rm .graphify_old.json' lines were left alone, being a post-merge cleanup with different semantics. Thank you again for the handover offer; happy to defer to maintainers on which call site stands.

PR #2799, opened 2026-08-16
fix(detect): stop silently dropping rules from a non-UTF-8 ignore file (#2798)

### LINKED ISSUE #2798: A non-UTF-8 .graphifyignore silently loses its non-ASCII rules: errors="ignore" turns an exclusion into a pattern that matches nothing
## Summary

`.gitignore`, `.graphifyignore` and `$GIT_DIR/info/exclude` are read with `errors="ignore"`, which turns a mis-encoded byte into *no* byte. An ignore file that is not valid UTF-8 therefore loses its non-ASCII rules **silently** — the pattern survives as a shorter string that matches nothing, the directory is scanned, and nothing anywhere says so.

`graphify/detect.py` (0.9.44), both read sites:

```python
for raw in ignore_file.read_text(encoding="utf-8-sig", errors="ignore").splitlines():   # _load_dir_own_ignore
for raw in info_exclude.read_text(encoding="utf-8-sig", errors="ignore").splitlines():  # _load_graphifyignore
```

A rule reading `Orçamento/` saved in the host ANSI codepage is `Or\xe7amento/` on disk. `\xe7` is not valid UTF-8, so it is dropped and the effective pattern becomes `Oramento/`.

## Reproduce

```python
from pathlib import Path

base = Path(tempfile.mkdtemp())
(base / ".graphifyignore").write_bytes("Orçamento/\n".encode("cp1252"))
(base / "Orçamento").mkdir()
(base / "Orçamento" / "contrato.py").write_text("x = 1", encoding="utf-8")
(base / "main.py").write_text("print('hi')", encoding="utf-8")

print(sorted(Path(f).name for f in detect(base)["files"]["code"]))
```

```
--- .graphifyignore saved as utf-8 ---
  raw bytes            : b'Or\xc3\xa7amento/\n'
  excluded as intended : True
  files scanned        : ['main.py']

--- .graphifyignore saved as cp1252 ---
  raw bytes            : b'Or\xe7amento/\n'
  excluded as intended : False
  files scanned        : ['contrato.py', 'main.py']
  *** the ignore rule silently did nothing ***
```

## Why it matters

An ignore rule failing *open* is not a cosmetic bug. `test_graphifyignore_matches_nfd_path_with_nfc_pattern` already states the stakes in its own docstring, for the normalisation version of this same failure:

> Without normalization the two compare unequal and the rule silently does nothing — the files get scanned, and docs/PDFs are sent to an LLM despite an explicit exclusion.

That is exactly what happens here, reached by a different route. Someone who writes `Orçamento/` is trying to keep that directory out of the corpus, and out of the semantic pass.

Non-UTF-8 ignore files are not exotic on Windows: ANSI was Notepad's default for years, and `Set-Content` still writes the ANSI codepage unless `-Encoding` is passed. Accented directory names are common in Portuguese, Spanish, French and German projects — the same population `_nfc` normalisation was added for.

## Two existing tests could never have caught it

`test_graphifyignore_matches_nfc_path_with_nfd_pattern` and its mirror write the ignore file with `write_text` and no `encoding=`, so on Windows they emit the locale codepage. One raises `UnicodeEncodeError` outright (the NFD form's combining cedilla, U+0327, has no cp1252 representation); the other writes cp1252 bytes the reader then mangles. Both have been failing on Windows, and the test job runs `ubuntu-latest` only.

## Suggested fix

Prefer UTF-8, but never silently drop a rule: decode UTF-8 (BOM-tolerant) first, and only on failure fall back to the host encoding and then latin-1, which cannot fail and maps every byte to a codepoint. A rule spelled in some third encoding still comes out wrong, but it comes out *whole*, and a one-time warning naming the file makes it fixable rather than invisible. Decoding must still never raise — a scan should not die on a stray byte in an ignore file.

## Environment

| | |
|---|---|
| Python | 3.12, Windows 11 |

abhay-codes07:
Fixes #2798.

## The bug

`.gitignore`, `.graphifyignore` and `$GIT_DIR/info/exclude` were read with `errors="ignore"`:

```python
for raw in ignore_file.read_text(encoding="utf-8-sig", errors="ignore").splitlines():
```

`errors="ignore"` turns a mis-encoded byte into *no* byte. An ignore file saved in the host ANSI codepage — Notepad's historical default on Windows, and still what `Set-Content` writes without `-Encoding` — is not valid UTF-8, so a rule reading `Orçamento/` (cp1252 `Or\xe7amento/`) decoded to the pattern `Oramento/`.

That matches nothing, and nothing said so:

```
--- .graphifyignore saved as utf-8 ---
  raw bytes            : b'Or\xc3\xa7amento/\n'
  excluded as intended : True
  files scanned        : ['main.py']

--- .graphifyignore saved as cp1252 ---
  raw bytes            : b'Or\xe7amento/\n'
  excluded as intended : False
  files scanned        : ['contrato.py', 'main.py']
  *** the ignore rule silently did nothing ***
```

This is the failure mode `test_graphifyignore_matches_nfd_path_with_nfc_pattern` already warns about in its own docstring — *"the rule silently does nothing — the files get scanned, and docs/PDFs are sent to an LLM despite an explicit exclusion"* — reached by a different route. An exclusion that fails open is worth more than a tidy decode: the whole reason someone writes `Orçamento/` is to keep that directory out of the corpus.

## Change

`_read_ignore_text` tries UTF-8 (BOM-tolerant) first, since that is the format every other reader here assumes and what the docs describe. Only if that fails does it fall back to the host encoding, then to latin-1, which cannot fail and maps every byte to a codepoint.

A rule spelled in some third encoding still comes out *wrong*, but it comes out **whole** rather than being silently truncated to something that matches nothing — and a one-time warning names the file, so it is fixable instead of invisible:

```
[graphify] WARNING: <path>/.graphifyignore is not valid UTF-8; read it as cp1252
instead. Re-save it as UTF-8 — patterns with non-ASCII characters may not match
as written.
```

Decoding still never raises, which was the previous contract and matters here — a scan must not die on a stray byte in an ignore file. `test_decoding_never_raises_on_arbitrary_bytes` feeds it all 256 byte values to pin that.

I kept UTF-8 as the preferred encoding rather than guessing per-file: changing the documented format would be a much bigger decision than fixing a silent failure, and every ignore file that is already valid UTF-8 decodes on the first branch exactly as before.

## Two existing tests were failing on Windows for a related reason

`test_graphifyignore_matches_nfc_path_with_nfd_pattern` and its mirror wrote the ignore file through `write_text` with no `encoding=`, so they emitted the *locale* codepage:

- one raised `UnicodeEncodeError` outright, because the NFD form's combining cedilla (U+0327) has no cp1252 representation;
- the other wrote cp1252 bytes that the reader then had to guess at.

Both now pass `encoding="utf-8"`. That is a two-character fix, but it is the reason the product bug above stayed invisible: the tests that would have caught it could not run.

## Tests

`tests/test_ignore_file_encoding.py` (12 tests). They write the ignore file as **bytes** rather than through `write_text`, so they pin the decoding behaviour on every platform rather than only where cp1252 happens to be the default — these have teeth on Linux CI, not just Windows.

Reverting `_read_ignore_text` to the old one-liner and keeping the tests fails 3, on either platform:

```
FAILED tests/test_ignore_file_encoding.py::test_ansi_encoded_rule_still_excludes
FAILED tests/test_ignore_file_encoding.py::test_ansi_encoded_rule_warns_once_naming_the_file
FAILED tests/test_ignore_file_encoding.py::test_no_byte_is_dropped_from_a_mis_encoded_file
```

The existing NFC/NFD normalisation guarantee is re-pinned through the new decode path in both directions, so this cannot quietly undo #1226's work.

## Validation

- Full suite: `20 failed, 4469 passed` -> `18 failed, 4482 passed`.
- The two NFC/NFD tests move from fail to pass; no other failure changes state. The remaining 18 are pre-existing Windows failures (symlink privileges, FIFO/socket fixtures, and similar), unrelated to this change.

### DIFF
--- a/graphify/detect.py
+++ b/graphify/detect.py
@@ -1122,6 +1122,53 @@ def _git_info_exclude(vcs_root: Path) -> Path | None:
     return exclude if exclude.is_file() else None
 
 
+_warned_ignore_encodings: set[str] = set()
+
+
+def _read_ignore_text(path: Path) -> str:
+    """Read an ignore file, preferring UTF-8 but never silently dropping a rule.
+
+    These files were read with ``errors="ignore"``, which turns a mis-encoded
+    byte into *no* byte. An ignore file saved in the host's ANSI codepage — the
+    historical Notepad default on Windows, and still what ``Set-Content`` writes
+    without ``-Encoding`` — is not valid UTF-8, so ``Or\xe7amento/`` decoded to
+    the pattern ``Oramento/``. That matches nothing, and nothing said so: the
+    directory was scanned despite an explicit exclusion, which for a rule
+    covering documents or PDFs means they reach the semantic pass anyway.
+
+    So: UTF-8 (BOM-tolerant) first, since that is what the format should be and
+    what every other reader here assumes. Only if that fails do we fall back to
+    the host encoding, then to latin-1, which cannot fail and maps every byte to
+    a codepoint — a rule spelled in some third encoding still comes out wrong,
+    but it comes out *whole*, and the warning names the file so it is fixable.
+    Decoding never raises, matching the previous contract.
+    """
+    raw = path.read_bytes()
+    try:
+        return raw.decode("utf-8-sig")
+    except UnicodeDecodeError:
+        pass
+    import locale
+    import sys as _sys
+    fallback = locale.getpreferredencoding(False) or "latin-1"
+    for enc in (fallback, "latin-1"):
+        try:
+            text = raw.decode(enc)
+        except (UnicodeDecodeError, LookupError):
+            continue
+        key = str(path)
+        if key not in _warned_ignore_encodings:
+            _warned_ignore_encodings.add(key)
+            print(
+                f"[graphify] WARNING: {path} is not valid UTF-8; read it as "
+                f"{enc} instead. Re-save it as UTF-8 — patterns with non-ASCII "
+                "characters may not match as written.",
+                file=_sys.stderr,
+            )
+        return text
+    return raw.decode("utf-8", errors="ignore")
+
+
 def _load_dir_own_ignore(d: Path, *, gitignore: bool = True) -> list[tuple[Path, str]]:
     """Read .gitignore/.graphifyignore directly inside *d* (not its ancestors).
 
@@ -1142,7 +1189,7 @@ def _load_dir_own_ignore(d: Path, *, gitignore: bool = True) -> list[tuple[Path,
     for fname in ((".gitignore", ".graphifyignore") if gitignore else (".graphifyignore",)):
         ignore_file = d / fname
         if ignore_file.exists():
-            for raw in ignore_file.read_text(encoding="utf-8-sig", errors="ignore").splitlines():
+            for raw in _read_ignore_text(ignore_file).splitlines():
                 line = _parse_gitignore_line(raw)
                 if line:
                     patterns.append((d, line))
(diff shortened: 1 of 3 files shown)


safishamsi:

Issue #2791, opened 2026-08-16
Simple-Graph build silently drops 471 of 3178 edges (57 pairs differ in relation type); collapse is only reported after graph.json is written


cdtatarevich:
## Environment

- graphifyy 0.9.39, macOS 15, 116-file Python corpus, undirected build (skill default)

## What happens

`build_from_json` produces a simple `networkx.Graph`, so every additional edge between an already-connected pair is silently dropped. Step 4.5 measures the loss accurately — but only *after* Step 4 has already written `graph.json`:

```
raw_edges: 3425
valid_candidate_edges: 3178
undirected_unique_endpoint_pairs: 2707
undirected_same_endpoint_collapsed_edges: 471
same_endpoint_group_count: 160
relation_variant_groups: 57
source_location_variant_groups: 96
post_build_graph_type: Graph
post_build_edges: 2707
```

471 edges (~15% of valid candidates) are gone from the artifact, across 160 endpoint pairs. Example from the report:

```
tests_test_channels_testchannelregistry -> agent_reach_channels_bilibili_bilibilichannel
  edges=7 relations=['uses'] locations=['L1198','L1233','L1252','L1266','L1289','L1299','L13']
```

Seven distinct call sites become one edge with one `source_location`. And `relation_variant_groups: 57` means 57 pairs collapsed edges that did **not** even share a relation type — a `calls` and a `references` between the same two nodes keep only whichever landed first.

## Why it matters beyond tidiness

- **Weight is lost.** "Called from 7 places" and "called once" are the same edge afterwards, so anything ranking by coupling strength (god nodes, betweenness, `suggest_questions`) is computed on a flattened graph.
- **Provenance is lost.** `source_location` for the surviving edge is arbitrary, so "where exactly does A use B" has one answer out of seven, with nothing marking it as partial.
- **The warning is unactionable as written.** It says `graph may be incomplete/corrupt` after the fact; there is no flag to re-run without the loss, so the only response is to accept it.

## Suggested fix, in order of cost

1. **Aggregate rather than drop.** On collision, keep one edge but set `weight` (or `count`) to the number of merged edges and carry `source_locations` as a list. No API change for consumers that ignore the extra fields, and the diagnostic drops to zero for the common case.
2. **Do not merge across relation types** — those 57 groups are a different fact being discarded, not a duplicate.
[...]

PR #2804, opened 2026-08-16
fix(build): keep the specific relation when an edge pair collapses (#2803)

### LINKED ISSUE #2803: Edge collapse picks the surviving relation alphabetically, so `references` overwrites `calls` every time (144/144 pairs) and those calls leave the call graph
## Summary

When two edges land on the same node pair, the simple graph keeps one — and which relation survives is decided **alphabetically**. `references` sorts after `calls`, so `references` wins every time. On graphify's own codebase that is **144 out of 144** pairs where the extraction found both: the graph says `references` for every one of them.

This is a systematic downgrade rather than a tie-break, and it is not cosmetic: `callflow` filters on relation and its list does not include `references`, so all 144 call sites leave the call graph.

Related to #2791 but distinct. #2791 is about edges being *dropped* by the simple-graph model. This is about the *surviving* edge carrying the wrong fact, which is a smaller and separately fixable problem.

## Mechanism

`graphify/build.py`:

```python
    # Iterate edges in a deterministic order. ... when two edges collapse onto
    # the same node pair the last write wins, so an unstable iteration order
    # flips _src/_tgt run-to-run ... Sorting fixes the last-write outcome.
    for edge in sorted(
        extraction.get("edges", []),
        key=lambda e: (
            str(e.get("source", e.get("from", ""))),
            str(e.get("target", e.get("to", ""))),
            str(e.get("relation", "")),        # <-- decides the surviving relation
        ),
    ):
```

The sort was added for determinism (#1061 — an unstable order flipped `_src`/`_tgt` between runs). The side effect is that the third key element also decides *which relation wins*, alphabetically:

```
calls < contains < imports < imports_from < indirect_call < method < rationale_for < references < uses
```

`nx.Graph.add_edge` overwrites the attribute dict of an existing edge, and the only same-pair guard (#1061) fires when the relation **matches** and the direction is reversed — so two edges with different relations always resolve to the alphabetically-last one.

## Reproduce

On graphify's own package (81 files, 2287 nodes, 6176 extracted edges):

```
pairs where the extraction found BOTH `calls` and `references`: 144
relation on the surviving edge: {'references': 143, 'uses': 1}

pairs where a real `calls` was overwritten by a weaker relation: 144
   graphify_build_build_from_json <-> graphify_build_py_path   graph says 'references', extraction also found 'calls'
   graphify_cache_py_path <-> graphify_cache_stat_key_to_relative   graph says 'references', extraction also found 'calls'
   ...
pairs whose surviving relation is weaker than one the extraction found: 146
```

Minimal case:

```python
from graphify.build import build_from_json, edge_data
ex = {"nodes": [{"id": "a", "label": "a()", "file_type": "code", "source_file": "a.py"},
                {"id": "b", "label": "b()", "file_type": "code", "source_file": "b.py"}],
      "edges": [{"source": "a", "target": "b", "relation": "calls", "confidence": "EXTRACTED"},
                {"source": "a", "target": "b", "relation": "references", "confidence": "EXTRACTED"}],
      "hyperedges": []}
G = build_from_json(ex)
print(edge_data(G, "a", "b")["relation"])   # -> 'references'
```

Swapping the two edges in the list makes no difference: the sort reorders them either way.

## Why it matters

- **callflow loses the call.** `callflow_html.py` keeps edges whose relation is in `("calls", "imports", "imports_from", "uses", "method", "indirect_call")`. `references` is not in that set, so a downgraded pair is simply absent from the call graph. On this corpus that is 143 edges — 2788 visible instead of 2931.
- **The audit trail states the weaker fact.** An edge the extractor tagged `EXTRACTED` / `calls` is reported as `references`, and nothing records that a call was found.
- **"What calls X" is answered from it.** `query`, `path` and `explain` all read the surviving relation.

Every extractor that emits a call also tends to emit a reference for the same pair, so this fires on ordinary code rather than on some unusual shape — the 144/144 rate is the tell.

## Suggested fix

A generic relation (`references`, `uses`, `mentions` — the ones that say only "these two appear together") should never overwrite a specific one on the same pair. The reverse should still overwrite, so the outcome stops depending on arrival order in either direction.

A small denylist seems better here than a full precedence order over every relation: ranking `contains` against `calls` would be inventing a cross-axis judgement, whereas "specific beats generic" is the only comparison this collapse needs. Two specific relations can keep their current behaviour.

Measured on the graph above:

| | before | after |
|---|---|---|
| `calls` edges | 1914 | 2060 |
| `references` edges | 548 | 405 |
| edges visible to callflow | 2788 | 2931 |
| pairs reporting a weaker relation | 146 | 0 |
| total edges | 5396 | 5396 |

Node and edge counts are unchanged — this only changes *which* of the colliding edges survives.

## Environment

| | |
|---|---|
| Python | 3.12 |

abhay-codes07:
Fixes #2803.

## The bug

`build_from_json` keeps one edge per node pair and resolves collisions by "last write wins" over a sort keyed on `(source, target, relation)`. That sort is there for determinism — an unstable order flipped `_src`/`_tgt` between runs (#1061) — but the third key element also decides **which relation survives**, and it decides it alphabetically:

```
calls < contains < imports < imports_from < indirect_call < method < rationale_for < references < uses
```

So `references` always overwrites `calls`, and `uses` overwrites everything. That is not a tie-break, it is a systematic downgrade. On graphify's own package every single pair where the extraction found both came out as the weaker fact:

```
pairs where the extraction found BOTH `calls` and `references`: 144
relation on the surviving edge: {'references': 143, 'uses': 1}
```

144 of 144 — not a sample. Swapping the two edges in the input changes nothing, because the sort reorders them either way.

## Why it is not cosmetic

`callflow_html.py` selects edges by relation:

```python
if e.get("relation") in ("calls", "imports", "imports_from", "uses", "method", "indirect_call"):
```

`references` is not in that set, so every downgraded pair drops out of the call graph entirely. The audit trail also states the weaker fact for an edge the extractor tagged `EXTRACTED`/`calls`, and `query` / `path` / `explain` all read the surviving relation when answering "what calls X".

Every extractor that records a call tends to record a reference for the same pair, so this fires on ordinary code rather than an unusual shape. The 144/144 rate is the tell.

## The change

A generic relation never overwrites a specific one on the same pair. The reverse still overwrites, so the surviving relation no longer depends on arrival order in either direction.

```python
_GENERIC_RELATIONS: frozenset[str] = frozenset({"references", "uses", "mentions"})
```

I deliberately did **not** introduce a full precedence order over every relation. Ranking `contains` against `calls` would be inventing a cross-axis judgement that this collapse does not need — the only comparison required is "specific beats generic". Two specific relations keep their existing behaviour, and there is a test pinning that so the denylist cannot quietly grow into an ordering.

This is also **not** a fix for #2791. That issue is about edges being *dropped* by the simple-graph model, which is a much bigger change; this only corrects which of the colliding edges survives. Node and edge counts are byte-identical before and after.

## Measured on graphify's own graph

81 files, 2287 nodes, 6176 extracted edges:

| | before | after |
|---|---|---|
| `calls` edges | 1914 | **2060** |
| `references` edges | 548 | 405 |
| edges visible to callflow's filter | 2788 | **2931** |
| pairs reporting a relation weaker than the extraction found | 146 | **0** |
| total edges | 5396 | 5396 |
| total nodes | 2287 | 2287 |

## Tests

`tests/test_relation_collapse_precedence.py` (65 tests). The core is a matrix over every specific relation × every generic relation × **both arrival orders**, since the point of the fix is that order stops mattering.

Beyond that it pins what must *not* change: two generics keep previous behaviour, two specifics keep previous behaviour, the #1061 reverse-direction guard still holds, a lone generic edge is still kept, exactly one edge still survives per pair, and the surviving edge keeps the specific edge's own `_src`/`_tgt` and `source_location` rather than inheriting the demoted one's.

Reverting `build.py` and keeping the tests fails 55 of the 65. These are pure graph-shape assertions with no filesystem or platform dependency, so they have the same teeth on your Ubuntu CI as locally.

## Validation

- Full suite: `20 failed, 4469 passed` -> `20 failed, 4534 passed`. **Identical failure set** — no regressions; the +65 are the new tests. The 20 are pre-existing Windows failures (symlink privileges, FIFO/socket fixtures, and similar), unrelated to this change.

### DIFF
--- a/graphify/build.py
+++ b/graphify/build.py
@@ -52,6 +52,16 @@ def _is_ast_tier(item: dict) -> bool:
     return isinstance(loc, str) and bool(_AST_LOC_RE.match(loc))
 
 
+# Relations that say only "these two symbols appear together", with no claim about
+# HOW. An extractor that finds a specific fact for a pair — a call, an import, an
+# inheritance — routinely emits one of these for the same pair as well, so when the
+# simple graph collapses the pair to one edge, the generic one must never be the
+# survivor. Deliberately a small denylist rather than a full precedence order over
+# every relation: ranking `contains` against `calls` would be inventing a
+# cross-axis judgement, whereas "specific beats generic" is the only comparison
+# this collapse actually needs.
+_GENERIC_RELATIONS: frozenset[str] = frozenset({"references", "uses", "mentions"})
+
 # Language interop families, keyed by extension, for the cross-language phantom-edge
 # guard in the edge loop below. Families group by REAL interop (JS/TS share a module
 # graph; C/C++/ObjC share a compilation unit via headers; JVM langs share bytecode),
@@ -1228,6 +1238,25 @@ def build_from_json(extraction: dict, *, directed: bool = False, root: str | Pat
             existing = edge_data(G, src, tgt)
             if existing.get("relation") == attrs.get("relation") and (
                 existing.get("_src") == tgt and existing.get("_tgt") == src
+            ):
+                continue
+        # A pair that already carries a SPECIFIC relation must not be downgraded
+        # to a generic one. Only one edge survives per pair here, and the sort
+        # above orders same-pair edges by relation name, so "last write wins"
+        # resolved the winner alphabetically — which put `references` after
+        # `calls` and `uses` after everything. On graphify's own corpus that
+        # rewrote all 144 pairs where the extraction found both `calls` and
+        # `references` into plain `references`, and callflow's relation filter
+        # does not include `references`, so those call sites left the call graph
+        # entirely. Alphabetical order carries no meaning; keeping the specific
+        # fact does. The reverse (specific arriving after generic) still
+        # overwrites, so the outcome no longer depends on edge order at all.
+        if G.has_edge(src, tgt):
+            existing_rel = edge_data(G, src, tgt).get("relation")
+            if (
+                attrs.get("relation") in _GENERIC_RELATIONS
+                and existing_rel is not None
+                and existing_rel not in _GENERIC_RELATIONS
             ):
                 continue
         G.add_edge(src, tgt, **attrs)
--- a/tests/test_relation_collapse_precedence.py
+++ b/tests/test_relation_collapse_precedence.py
@@ -1,0 +1,159 @@
+"""A collapsed edge must keep the specific relation, not the alphabetical one.
+
+`build_from_json` puts one edge per node pair into a simple graph, and resolved
+same-pair collisions by "last write wins" over a sort keyed on
+`(source, target, relation)`. That sort exists for determinism (#1061 — an
+unstable order flipped `_src`/`_tgt` run to run), but it also decided which
+RELATION survived, and it decided it alphabetically:
+
+    calls < contains < imports < ... < references < uses
+
+So `references` always overwrote `calls`, and `uses` overwrote everything. On
+graphify's own corpus that rewrote all 144 pairs where the extraction found both
+`calls` and `references` into plain `references` — 144 out of 144, not a
+sampling — and callflow's relation filter
+(`calls, imports, imports_from, uses, method, indirect_call`) does not include
+`references`, so those call sites dropped out of the call graph entirely.
+
+Alphabetical order carries no meaning. These tests pin that a generic relation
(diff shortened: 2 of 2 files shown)


safishamsi:

PR #2808, opened 2026-08-16
fix(query): name the graph the answer came from (#2789)

### LINKED ISSUE #2789: query resolves graphify-out/ against CWD and never echoes which graph it opened — answers silently come from another project's graph
## Environment

- graphifyy 0.9.39, macOS 15 (Darwin 25.6.0), pipx install, Claude Code host
- Two graphs on one machine: a parent project at `~/Proj/` and a vendored subproject at `~/Proj/Sub/`, each with its own `graphify-out/` built by the skill's step-by-step flow.

## What happens

`graphify query` resolves `graphify-out/graph.json` against the **current working directory** with no indication of which graph it opened. Running it from the parent while intending to ask about the subproject returns a confident, well-formed answer from the *wrong corpus*.

```
$ cd ~/Proj                      # parent project's graph
$ graphify query "Which code enforces the safe-by-default install gate?"
Traversal: BFS depth=2 | Start: ['Safe-by-default install (--system opt-in)', ...] | 33 nodes found
NODE Safe-by-default install (--system opt-in) [src=Sub/README.md ...]
NODE Migration Notes — ...                     [src=MIGRATION-NOTES.md ...]   <- parent project
NODE Stop-Breach Poll                          [src=skills/daily-exit-check.md ...]  <- parent project
```

The parent graph happened to contain `Sub/README.md` as a corpus file, so the seed terms matched and the traversal produced plausible output built almost entirely from unrelated documents. Nothing in the output names the graph file, the scan root, or the node count of the graph being queried. I only caught it because the community labels were ones I had not assigned in that project.

Same command from `~/Proj/Sub` returns the correct 355-node result.

## Why it is easy to hit

`graphify-out/` is CWD-relative by design (noted in #2603 for the hook path), but `query` is the interactive surface where the user is least likely to be thinking about CWD — you ask a question about a subproject you were just reading, from wherever your shell happens to be. Unlike a build, there is no node-count or file-count output that would look wrong.

## Suggestions, cheapest first

1. **Echo the resolved source.** One line before the traversal: `graph: /abs/path/graphify-out/graph.json (1272 nodes, root=/abs/scan/root)`. This alone makes the failure self-evident, and helps for `path`/`explain` too.
2. **Accept `--graph <path>`**, which `path`, `explain` and `diagnose multigraph` already take, so `query` is the odd one out.
3. **Consider walking up** to the nearest ancestor `graphify-out/` (git-style discovery) and/or honoring `.graphify_root`, so an intended subproject query from a child directory resolves the way users expect.

Related: #2603 (same CWD/scan-root coupling, hook surface), #2672 (empty results reading as "absent from the codebase" rather than "absent from the graph" — the same class of "the answer does not say what it is an answer about").

abhay-codes07:
Fixes #2789.

Thanks @cdtatarevich — "confident, well-formed answer from the wrong corpus" is the right framing, and it is what made this worth fixing rather than documenting.

## The change

The query header now leads with the graph it opened, and its size:

```
Graph: graphify-out/graph.json (355 nodes) | Traversal: BFS depth=2 | Start: [...] | 33 nodes found
```

Relative to the CWD when the graph sits underneath it — the ordinary case, and short enough not to be noise. Absolute when it does not, which is exactly the case worth noticing:

```
Graph: C:/other/proj/graphify-out/graph.json (3178 nodes) | Traversal: BFS depth=2 | ...
```

Two details that are deliberate rather than incidental:

**The count is the corpus, not the traversal.** The line already ends with `N nodes found`, which is the traversal result. The number next to the path is the whole graph, because your report's tell was a 355-node subproject answering as a 3178-node parent. Reusing the traversal count would have hidden precisely that.

**Absolute is the loud form.** Elsewhere in the codebase absolute host paths are something to avoid (#2598, #2628) because they get *written into artifacts* and travel between machines. This is terminal output for the operator standing in a directory, so the reasoning inverts: a path outside the CWD should look different from one inside it.

## Scope

`graph_path` is optional and the header is byte-identical when it is not passed, so no existing caller changes behaviour. The two call sites that do know their path — the CLI `query` command and the MCP `query` tool — now pass it, and both already had the resolved path in hand for `querylog.log_query(corpus=...)`, so nothing new is computed.

I did **not** add a `--graph`-mismatch warning or make `query` search upward for a nearer `graphify-out/`. Both were tempting, but changing *which* graph is opened is a behavioural decision with its own failure modes, and the issue's actual complaint is that the choice was invisible. Making it visible is the smaller, complete fix; happy to follow up on resolution if you want it.

`_display_graph_path` swallows resolution failures and falls back to the string it was handed — a display helper must never be the reason a query fails, and there is a test feeding it null bytes and 300-character garbage.

## Tests

`tests/test_query_names_its_graph.py` (9 tests): the relative and absolute forms, POSIX separators on both platforms, the unchanged header when no path is supplied, the corpus-vs-traversal count distinction, hostile input, and an end-to-end check that a parent project and a vendored subproject produce headers you can tell apart — the exact scenario from the report.

## Validation

- Full suite: `20 failed, 4469 passed` -> `20 failed, 4478 passed`. **Identical failure set** — no regressions; the +9 are the new tests.

### DIFF
--- a/graphify/cli.py
+++ b/graphify/cli.py
@@ -1047,6 +1047,7 @@ def dispatch_command(cmd: str) -> None:
             depth=2,
             token_budget=budget,
             context_filters=context_filters,
+            graph_path=str(gp),
         )
         querylog.log_query(
             kind="query",
--- a/graphify/serve.py
+++ b/graphify/serve.py
@@ -1135,6 +1135,27 @@ def _cut_lines_to_budget(lines: list[str], token_budget: int, narrow_hint: str) 
     )
 
 
+def _display_graph_path(graph_path: str) -> str:
+    """Render a graph path for the query header.
+
+    Relative to the CWD when it sits underneath it — `graphify-out/graph.json`,
+    which is the ordinary case and stays short. Absolute otherwise, because a
+    graph outside the directory you are standing in is precisely the situation
+    the header exists to make visible (#2789). Always POSIX separators so the
+    line reads the same on either platform. Falls back to the path as given if
+    it cannot be resolved; this is a display helper and must never be the reason
+    a query fails.
+    """
+    try:
+        p = Path(graph_path).resolve()
+        try:
+            return p.relative_to(Path.cwd().resolve()).as_posix()
+        except ValueError:
+            return p.as_posix()
+    except (OSError, RuntimeError, ValueError):
+        return str(graph_path)
+
+
 def _query_graph_text(
     G: nx.Graph,
     question: str,
@@ -1143,6 +1164,7 @@ def _query_graph_text(
     depth: int = 3,
     token_budget: int = 2000,
     context_filters: list[str] | None = None,
+    graph_path: str | None = None,
 ) -> str:
     terms = _query_terms(question)
     # One graph scoring pass produces both the combined ranking (used to drive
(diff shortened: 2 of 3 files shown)


safishamsi:
Shipped in v0.9.47 via authorship-preserving cherry-pick, plus a CLI call-site integration test. Thanks @abhay-codes07! Release: https://github.com/Graphify-Labs/graphify/releases/tag/v0.9.47

Issue #1909, opened 2026-07-15
0.9.16: semantic cache emits .graphifyignore'd files as graph nodes


mphstudios:
## Summary

On 0.9.16, `graphify extract` emits nodes for source files excluded by `.graphifyignore` when those files have entries in the semantic cache from an earlier extraction. The files are not re-dispatched (a `RuntimeWarning` is logged), but the stale cache entries still surface as nodes (with incident edges and hyperedge memberships) in the output graph.

## Environment

- graphify 0.9.16 (pipx)
- Python 3.13, macOS

## Repro

1. Extract a repo once while directory `foo/` is in scope — this populates the semantic cache with `foo/*` entries.
2. Add `foo/` to `.graphifyignore`.
3. Re-run `graphify extract . --backend <…>`.

## Observed

During extraction:

```
.../graphify/llm.py:1927: RuntimeWarning: semantic cache skipped out-of-scope source_file 'foo/bar.md'; the file was not dispatched for extraction
  _scs(
```

Despite the warning and `.graphifyignore`, `foo/*` files still appear as **nodes** in the output `graph.json`. In our case, `issues/` is `.graphifyignore`d, yet three `issues/issue-*.md` files surfaced as nodes (with their frontmatter-derived labels), plus 7 incident `references` edges pointing at them and one hyperedge membership.

Note the same `RuntimeWarning` also fires for files that *are* in scope and *are* correctly extracted (e.g. `OBJECTIVES.md`, spec files) — so the warning is a symptom of cache-scope handling generally, and the leak is the subset that are `.graphifyignore`d.

## Expected

`.graphifyignore`d files should never appear as nodes in the extracted graph, regardless of stale semantic-cache entries. The cache should be scoped/invalidated against the current `.graphifyignore`, or cached entries for ignored files should be dropped from the output rather than only skipped for re-dispatch.

## Impact

The committed per-release semantic snapshot picks up ignored files (issue trackers, generated docs), polluting the graph with nodes the ignore rules explicitly exclude. Current workaround: after `extract`, prune nodes whose `source_file` matches `.graphifyignore` (and their incident edges / hyperedge refs).

## Related

Possibly adjacent to #1870 (semantic-cache checkpoint handling on 0.9.16).

PR #2809, opened 2026-08-16
fix(build): sweep external-import nodes a prune leaves orphaned (#2807)

### LINKED ISSUE #2807: Pruning a source file leaves its external-import nodes behind as permanent orphans: no source_file means no prune can ever match them
## Summary

`prune_sources` matches nodes on their `source_file`. Extractors create per-file nodes for **imported external symbols** — `Path` from pathlib, `Counter` from collections, `Graph` from networkx — and those carry no `source_file`, because they are defined outside the corpus. Every edge they have points at symbols in the one file they were created for.

So when that file is deleted or excluded, the prune removes the file's own nodes and leaves these behind: degree 0, named after a file the corpus no longer contains, and **unreachable by any future prune**, since there is no `source_file` to match on.

## Reproduce

graphify's own package, 2287 nodes / 5396 edges, pruning one real file:

```
after pruning 'graphify/callflow_html.py': 2147 nodes, 5120 edges
[graphify] Pruned 137 node(s) from 1 deleted or excluded source file(s).

orphaned (degree-0) nodes now: 3  (new: 2)
   'graphify_callflow_html_py_path'    label='Path'    source_file=''
   'graphify_callflow_html_py_counter' label='Counter' source_file=''

nodes still named after the pruned file: 2
   'graphify_callflow_html_py_path'    label='Path'    source_file='' deg=0
   'graphify_callflow_html_py_counter' label='Counter' source_file='' deg=0
```

137 of the file's 139 nodes go; 2 stay forever.

This corpus has 97 source-less nodes across 81 files, so roughly one to two per file are candidates. Excluding a directory of any size leaves a proportional residue, and each subsequent prune adds more — they cannot be cleaned up by re-running anything.

## Why it matters

- **They are counted.** `GRAPH_REPORT.md` opens with `N nodes · M edges · K communities`, computed off the graph, so the totals drift upward from reality with every prune.
- **They are exported.** `to_obsidian` writes one note per node, and the wiki and `graph.json` carry them too, so a vault keeps a `Path.md` for a file that is no longer in the corpus.
- **They are permanent.** Every other stale-node path has a way out — a deletion is caught by `deleted_files`, an exclusion by `excluded_files` (#1908) or `_stale_graph_sources` (#1909). A node with no `source_file` is outside all three.

Nothing warns. The prune line reports `137 node(s)` and looks complete.

## Suggested fix

After the prune removes nodes and edges, sweep the source-less nodes it just isolated:

```python
before = {n for n, d in G.nodes(data=True)
          if not d.get("source_file") and G.degree(n) == 0}
# ... existing prune ...
newly_isolated = [n for n, d in G.nodes(data=True)
                  if not d.get("source_file") and G.degree(n) == 0
                  and n not in before]
G.remove_nodes_from(newly_isolated)
```

Scoping it to nodes this prune isolated matters: a source-less node that was *already* isolated before the prune is someone else's business, and this graph had one such node that should not be swept up by a fix for a different problem.

A node with no `source_file` and no edges carries no information — it cannot be located in the corpus and it connects nothing — so removing it loses nothing, whereas keeping it inflates every count that reads the graph.

## Environment

| | |
|---|---|
| Python | 3.12 |

abhay-codes07:
Fixes #2807.

## The bug

`prune_sources` matches nodes on `source_file`. Extractors create a per-file node for each **imported external symbol** — `Path` from pathlib, `Counter` from collections, `Graph` from networkx — and those carry no `source_file`, because they are defined outside the corpus. Every edge they have points at symbols in the one file they were created for.

So pruning that file leaves them at degree 0, and nothing can ever collect them: deletions go through `deleted_files`, exclusions through `excluded_files` (#1908) and `_stale_graph_sources` (#1909), and **all three match on `source_file`**. A node without one is outside every path graphify has for removing stale content, so the residue is permanent and grows with each prune.

On graphify's own package:

```
after pruning 'graphify/callflow_html.py': 2147 nodes, 5120 edges
[graphify] Pruned 137 node(s) from 1 deleted or excluded source file(s).

nodes still named after the pruned file: 2
   'graphify_callflow_html_py_path'    label='Path'    source_file='' deg=0
   'graphify_callflow_html_py_counter' label='Counter' source_file='' deg=0
```

137 of the file's 139 nodes go; 2 stay forever, and the prune line reports `137 node(s)` and looks complete. With this change all 139 go.

They are not inert while they sit there. `GRAPH_REPORT.md` opens with a node count read straight off the graph, `to_obsidian` writes one note per node, and the wiki and `graph.json` carry them too — so a vault keeps a `Path.md` for a file the corpus no longer contains.

## Scope

The sweep is narrowed three ways, so it cannot quietly become a general garbage collector:

- **Only nodes with no `source_file`.** One that has a real `source_file` is prunable through the normal path and must not be second-guessed here.
- **Only nodes this prune isolated.** A source-less node that was *already* isolated beforehand is a different question and survives — graphify's own graph has exactly one, and a fix for this bug has no business collecting it.
- **Only inside the prune branch.** A plain merge never touches isolated nodes.

The distinction I want to flag for review, because it is the one that looks inconsistent at a glance: an isolated node that **does** have a `source_file` is deliberately kept. In the test fixture, `b.py`'s symbol whose only call went into a pruned `a.py` ends up at degree 0 — but it is a real symbol in a file that still exists, so pruning it would be data loss, and it stays prunable through the normal path if `b.py` ever goes. There is a test pinning that asymmetry so it reads as intentional rather than as a missed case.

A source-less node with no edges names nothing and connects nothing, so dropping it loses no information; keeping it inflates every count that reads the graph.

## Tests

`tests/test_prune_sweeps_orphans.py` (8 tests). Three fail without the fix — the external-import node going with its file, no source-less orphan remaining, and pruning every file actually emptying the graph.

The other five exist to stop the sweep over-reaching: a shared external node still referenced by a surviving file must live, a pre-existing isolate must live, an isolated node *with* a `source_file` must live, a merge with no prune must change nothing, and a prune that matches nothing must sweep nothing.

One of those caught me while writing this: my first version asserted "no orphans remain at all" and failed, correctly, because `mod_b_keep` is isolated but real. The assertion was wrong, not the fix — I narrowed it to source-less orphans and added the surviving-real-symbol case as an explicit test rather than loosening it silently.

## Validation

- Full suite: `20 failed, 4469 passed` -> `21 failed, 4476 passed`. The +8 are the new tests.
- The one extra failure is `test_incremental_mtime_collision.py::test_same_size_rewrite_in_one_tick_is_requeued`, which this PR does not touch: it depends on two writes landing in the same filesystem timestamp tick, and this run took 10:58 against the usual ~6:30, so the machine was loaded. It passes in isolation, and it has surfaced the same way on two unrelated branches of mine. Flagging it rather than quietly reporting 20.

### DIFF
--- a/graphify/build.py
+++ b/graphify/build.py
@@ -1776,6 +1776,13 @@ def build_merge(
 
     # Prune nodes and edges from deleted source files
     if prune_sources:
+        # Source-less nodes that are ALREADY isolated before this prune. They are
+        # not this prune's doing, so they must survive it — the sweep below is
+        # scoped to the ones it orphans itself.
+        _isolated_before = {
+            n for n, d in G.nodes(data=True)
+            if not d.get("source_file") and G.degree(n) == 0
+        }
         to_remove = [
             n for n, d in G.nodes(data=True)
             if _prune_match(d.get("source_file"))
@@ -1789,6 +1796,29 @@ def build_merge(
         ]
         if edges_to_remove:
             G.remove_edges_from(edges_to_remove)
+
+        # Extractors create a per-file node for each IMPORTED EXTERNAL symbol
+        # (`Path` from pathlib, `Counter` from collections), and those carry no
+        # source_file because they are defined outside the corpus. Every edge
+        # they have points at symbols in the one file they were created for, so
+        # pruning that file leaves them at degree 0 — named after a file the
+        # corpus no longer contains, counted in every total that reads the graph,
+        # exported as a note of their own, and unreachable by any future prune
+        # since there is no source_file to match on. Nothing else can collect
+        # them: deletions go through deleted_files, exclusions through
+        # excluded_files (#1908) and _stale_graph_sources (#1909), and all three
+        # match on source_file. A node with neither a source_file nor an edge
+        # names nothing and connects nothing, so dropping it loses no
+        # information (#2807).
+        orphaned = [
+            n for n, d in G.nodes(data=True)
+            if not d.get("source_file")
+            and G.degree(n) == 0
+            and n not in _isolated_before
+        ]
+        if orphaned:
+            G.remove_nodes_from(orphaned)
+            n_nodes += len(orphaned)
 
         # Report only the prune entries that ACTUALLY matched something — not
         # len(prune_sources), which counted every entry as pruned-from even
--- a/tests/test_prune_sweeps_orphans.py
+++ b/tests/test_prune_sweeps_orphans.py
@@ -1,0 +1,155 @@
+"""Pruning a source file must not leave its external-import nodes behind.
+
+`prune_sources` matches nodes on `source_file`. Extractors create a per-file node
+for each IMPORTED EXTERNAL symbol -- `Path` from pathlib, `Counter` from
+collections -- and those carry no `source_file`, because they are defined outside
+the corpus. Every edge they have points at symbols in the one file they were
+created for, so pruning that file left them at degree 0: named after a file the
+corpus no longer contains, counted in every total that reads the graph, exported
+as a note of their own, and unreachable by any future prune since there is no
+`source_file` to match on (#2807).
+
+On graphify's own package, pruning `graphify/callflow_html.py` removed 137 of its
+139 nodes and stranded `graphify_callflow_html_py_path` (label `Path`) and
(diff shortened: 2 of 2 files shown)


safishamsi:

Issue #2788, opened 2026-08-16
Astro: frontmatter fails at line 1 (partial extraction), and the warning hardcodes closed Kotlin issue #2551 for every language


sortakool:
## Summary

Two related problems, observed on **0.9.44**.

1. The Astro extractor fails at **line 1** on an ordinary `---` frontmatter
   block, so the file is only partially extracted.
2. The warning that reports it hardcodes **`(#2551)`** for every language, and
   #2551 is a **closed, Kotlin-specific** issue. Anyone who follows the pointer
   lands on a resolved problem that has nothing to do with their file.

## 1. Astro frontmatter fails at line 1

```
warning: 1 file(s) had syntax errors and may be partially extracted:
  website/src/pages/index.astro (first error at line 1) (#2551)
```

The file opens with the standard Astro component-script fence:

```astro
---
const GITHUB_URL = "https://github.com/adihebbalae/Attacca";
const BRANCH = "master";

const plugins = [
  { name: "attacca-core", tagline: "Workflow skills" },
  ...
];
---
```

Line 1 is the opening `---`, which is the most basic construct the format has,
so this does not look like an edge case in the grammar.

Source: `Attacca` at `website/src/pages/index.astro`.

## 2. The warning misdirects every reader

`extract.py:5636` (0.9.44):

```python
f"may be partially extracted: {_shown}{_more} (#2551)",
```

`(#2551)` is a literal in the shared message, emitted regardless of language.
The surrounding comments already know this — line 5619 refers to *"the genuine
#2551 Kotlin one-line-body"* case — but the reference is still attached to all
of them.

The cost is concrete: the citation is the only lead the message offers, and it
sends the reader to a closed Kotlin issue. We initially recorded this astro
failure as "already tracked upstream" on the strength of that number, and only
found otherwise by opening it.

Suggestion: drop the hardcoded number, or make it language-aware, or point at a
stable docs page about partial extraction instead of a single issue.

## Why partial extraction deserves a loud signal

"May be partially extracted" means an unknown number of nodes are missing from
the graph, and the run still exits 0. For a knowledge graph that is silent data
loss, and a consumer cannot tell a sparse file from a truncated one. A count of
what was dropped, even approximate, would make it actionable.

safishamsi:
[...]

Issue #2610, opened 2026-08-11
TypeScript/TSX: v0.9.39 flags 26 typechecked files as partially extracted after a force rebuild


Sid-AutoWisdom:
## Summary

Graphify 0.9.39 reports partial syntax extraction for 26 TypeScript/TSX files in a private Next.js App Router project, even though `npx tsc --noEmit` succeeds for the same checkout.

A forced rebuild was used, so this is not an incremental-cache result:

```
graphify update . --force
...
warning: 26 file(s) had syntax errors and may be partially extracted:
  page-client.tsx (first error at line 56),
  ArticlesContent.tsx (first error at line 385),
  page-client.tsx (first error at line 417),
  page.tsx (first error at line 2332),
  page.tsx (first error at line 16) (+21 more)
...
[graphify watch] Rebuilt: 4112 nodes, 11657 edges, 309 communities
```

Environment: macOS arm64, Python 3.14, Graphify 0.9.39.

## Verification

- `npx tsc --noEmit` succeeds.
- The project lint succeeds for the changed files.
- `graphify diagnose multigraph --json` reports zero missing endpoints, dangling edges, self-loops, or exact duplicate edges.
- The warning is repeatable after a force rebuild.

The repository is private, so I cannot attach the affected source. I also have not yet reduced a minimal parser snippet, so I am deliberately not attributing this to a specific grammar construct.

## Impact

The warning is useful, but the partially extracted files make graph-only usage unsafe for deletion decisions. We treat their nodes as incomplete and confirm with source, routing/configuration searches, typecheck, build, and focused flows instead.

This may be related to #2599, but the reported set includes additional files and I cannot yet prove the same triggers.

## Request

1. Could the warning include full relative paths (not only ambiguous basenames) and optionally machine-readable parse diagnostics?
2. Is there a supported command to emit the tree-sitter ERROR spans for each affected file, so private-repository users can minimize a reproducible case without patching Graphify locally?

safishamsi:
Fixed in v0.9.40 (`graphifyy==0.9.40` on PyPI). The #2551 partial-parse warning now fires only when recovery plausibly cost symbols (file yielded at most the file node, or an error region spans multiple lines), so valid TypeScript/TSX that tree-sitter recovers from cleanly (a `&` in a JSX string attribute, a semicolon-less `in_*` interface member) no longer warns, while the genuine Kotlin one-line-body / Luau cases still do. Thanks for the report.

Issue #2599, opened 2026-08-10
TypeScript/TSX: two false-positive "syntax errors" warnings — `&` in a JSX string attribute, and a member named `in_*` relying on ASI


atlasplatformu-ai:
graphify 0.9.39, Python 3.11.9, Windows, `tree-sitter-typescript` 0.23.2.

`graphify extract . --code-only` on a ~1400-file repo reports:

```
warning: 2 file(s) had syntax errors and may be partially extracted:
  SomePage.tsx (first error at line 276), types.ts (first error at line 1242) (#2551)
```

Both files are valid TypeScript — `tsc -b` on the project exits 0. I reduced them to two independent minimal cases. Each is accepted by `tsc` and rejected by the bundled tree-sitter grammar, so I believe both are false positives.

### 1. `&` inside a JSX string attribute (tsx grammar)

```tsx
declare const Comp: (p: { to: string }) => JSX.Element
export const el = <Comp to="/p?a=1&b=2" />
```

The ERROR node covers exactly `&b=2`. These parse fine:

- without the ampersand — `to="/p?a=1"`
- with the string in braces — `to={"/p?a=1&b=2"}`
- as a plain (non-JSX) string — `const s = "/p?a=1&b=2"`

Query-string URLs in `to=` / `href=` attributes are common in router-heavy code, so a single link can cost the whole file's nodes.

### 2. Member named `in_*` / `instanceof_*` relying on ASI (typescript grammar)

```ts
interface I {
  a: number
  in_workshop: number
}
```

The ERROR node spans from `number` to `in_workshop:`. It fires only when the **previous** member ends without `;` or `,`, and the member name begins with `in` or `instanceof` followed by a non-letter:

| member name | result |
|---|---|
| `in_workshop`, `in_x`, `in_`, `in2_x`, `instanceof_x` | ERROR |
| `inx`, `input`, `index`, `int_count`, `infer_x`, `instanceofx` | ok |

Adding `;` to the previous member avoids it, as does moving the member to first position. The keyword-boundary check for `in` / `instanceof` seems to treat `_` and digits as token terminators, though both are valid identifier-continuation characters. Names like `in_workshop`, `in_progress`, `in_stock` are common when types mirror snake_case API or DB fields, and semicolon-less interface members are the Prettier default.

### Verification

- `tsc --noEmit --strict` on case 2 → exit 0
- `tsc --noEmit --jsx preserve --skipLibCheck` on case 1 → exit 0
[...]

Issue #2520, opened 2026-08-06
`.luau` is parsed with the plain-Lua grammar: four Luau constructs silently truncate or hole a file (exit 0, no warning)


arnasmmm:
Hi — first, thanks for graphify; we use it daily as the code index an AI agent must consult before it
is allowed to read any source file.

We hit a silent data-loss case on Roblox Luau code and it took a full afternoon to find, so I want to
report it with the exact repro.

### What happens

`.luau` files are extracted with the Lua grammar (`ts_module="tree_sitter_lua"`, `extract.py` in
0.9.34). Luau is a superset of Lua 5.1, and the constructs it added are not in that grammar. When the
parser meets one, extraction for that file stops there and keeps only the nodes found before it —
**with exit code 0 and no message on stdout or stderr**.

The silence is the actual problem. A partial graph does not look partial: every later query answers
confidently, and "no matching nodes" reads as "this function does not exist".

### Minimal repro

A 1981-line `.luau` file with 25 declared functions, containing exactly two if-expressions:

```lua
local SessionStore = if IS_STUDIO then Store.Mock else Store          -- line 390
d.CharacterXp[name] = if xp > 0 then xp else nil                      -- line 636
```

```
graphify extract . --code-only
```

* as-is: **4** function nodes (the last one declared at line 278, i.e. everything after line 390 is
  gone), exit code 0, no warning
* after deleting those two lines only: **23** function nodes

Verified on 0.9.33 and 0.9.34, Windows, Python 3.13.

### Impact on a real corpus

188 Roblox scripts: **915 of 1078 declared functions present (84.9%)**. 16 files are truncated,
including one 1064-line file where the last node is at line 1 — 0 of its 16 functions are indexed.
That file happens to own the whole purchase flow, so every query about payments returned nothing and
looked like an answer.

### Requests

1. **Parse `.luau` with a Luau grammar** (e.g. tree-sitter-luau) instead of tree-sitter-lua. Luau
   constructs that a Lua 5.1 grammar rejects, all common in modern Roblox code:
   * if-expressions — `local x = if c then a else b`
   * type annotations — `local x: Foo = ...`, `function f(a: string): boolean`
   * type aliases and exports — `type T = {...}`, `export type T = ...`
   * generics — `function f<T>(...)`
   * type casts — `expr :: any`
   * string interpolation — `` `hi {name}` ``
   * compound assignment — `x += 1`, `s ..= "a"`
   * `continue`
[...]

PR #2819, opened 2026-08-17
fix(extract): make the partial-parse warning actionable, and stop citing closed #2551 (#2788)

### LINKED ISSUE #2551: Kotlin: bundled grammar rejects one-line type bodies (class C { val v = 1 }) — 21/541 files partially extracted, exit 0, no warning
## Summary

`class C { val v = 1 }` — a type body whose last declaration is not followed by a
newline or `;` — is rejected by the bundled `tree-sitter-kotlin` 1.1.0 grammar.
It is valid Kotlin, and the multi-line spelling of the same code parses fine.

On a 541-file Kotlin codebase (Android app + Wear module), **21 files (3.9%) fail
to parse**, and **60 of the repo's 1183 top-level functions (5.1%)** are missing
from `graph.json` — including 11 of the 65 screens reachable from the navigation
entry point, which is what made me look.

**Nothing is reported.** Exit code 0, no warning line, and the summary prints the
reduced node count as a plain result. The existing `zero nodes` warning does not
cover it: these files each yield 12–21 nodes, so they never qualify as empty. I
spent a while suspecting the call resolver before thinking to run the grammar
over the files by hand.

Same shape as #2520 (`.luau` constructs silently truncate a file, exit 0, no
warning), different language and trigger.

## Reproduction — the parse failure

`graphify 0.9.36`. Grammar-level, so it needs no extraction run:

```python
from tree_sitter import Language, Parser
import tree_sitter_kotlin as tsk
p = Parser(Language(tsk.language()))

def check(code):
    print("ERROR" if p.parse(code.encode()).root_node.has_error else "ok   ", repr(code))

check("package p\nclass C { val v = 1 }\n")        # ERROR  <-- valid Kotlin
check("package p\nclass C { val v = 1; }\n")       # ok
check("package p\nclass C {\n    val v = 1\n}\n")  # ok
check("package p\nclass C { }\n")                  # ok    (empty body)
check("package p\nfun f() { val v = 1 }\n")        # ok    (function body unaffected)
```

The trigger is a missing statement terminator before the closing brace of a
**type** body. It applies to `class`, `object` and `interface`, to properties and
to functions, and to the very common one-line companion:

```kotlin
class VM { fun f() = 1 }                               // ERROR
object O { val v = 1 }                                 // ERROR
interface I { fun f() }                                // ERROR
class VM {
    fun load() {}
    private companion object { const val LIMIT = 20 }   // ERROR (this one line)
}
```

Function bodies are unaffected — `fun f() { println(1) }` parses — which is why
this survives review: the same brace style is fine one scope out.

## What it costs on a real corpus

| | |
|---|---|
| Kotlin files scanned | 541 |
| files where `root_node.has_error` | **21** (3.9%) |
| top-level functions declared (excluding extension functions) | 1183 |
| top-level functions absent from `graph.json` | **60** (5.1%) |
| files those 60 live in | 10 — every one of them is inside the 21 above |

In the files I inspected, the `ERROR` node **covers the region the missing
declarations are in**:

```
KbHealthScreen.kt        ERROR at L119 spanning `KbHealthScreen`   -> that one function missing
KbInsightsScreen.kt      ERROR at L124                              -> that one function missing
ThemesScreen.kt          ERROR at L251 spanning `ThemesScreen`      -> 5 functions missing
LinkedInIdeasScreen.kt   ERROR L176-736                             -> 7 functions missing
SynthesesScreen.kt       ERROR L269-958                             -> 9 functions missing
CortexScreen.kt          ERROR L1-1175 (whole file)                 -> 19 functions missing;
                                                                       nothing extracted past L165
```

Damage varies with where the parser gives up: some files lose exactly one
function, others stop mid-file and lose everything after.

### Honest limit of this report

A parse error does **not** always cost nodes — the extractor often recovers. I
tried five minimal files carrying the bad construct (companion one-liner, nested
sealed interface, Compose-shaped bodies): all of them report `has_error` and yet
extract every declaration. So I can show the parse failure minimally, and I can
show the correlation and the overlapping spans on the real corpus, but I do not
have a *minimal* end-to-end file that loses a node. I did not want to dress the
correlation up as a reduction.

What does reproduce end-to-end, reliably, is a real affected file on its own — a
single-file directory is enough: `graphify extract <dir> --code-only` returns the
ViewModel class and its members, but not the composable declared after it. So the
loss does not need a large corpus to show up; it needs whatever makes the parser's
`ERROR` span the declaration rather than stop short of it, and that is the part I
could not reduce.

## What I am asking for

**Warn on a partial parse.** Whatever happens to the grammar, `extract` already
has `root_node.has_error` in hand and can say so — that alone would have saved
the investigation:

```
warning: 21 file(s) had syntax errors and may be partially extracted:
  app/.../KbHealthScreen.kt (first error at line 119), ... (+20 more)
```

Cheap check, and the existing `zero nodes` warning already sets the precedent and
the phrasing.

**The grammar bug itself is upstream** at `tree-sitter-grammars/tree-sitter-kotlin`
— 1.1.0 is the newest on PyPI, so there is no version to bump to. I am filing it
here because graphify is where it is observable and where the silence is fixable.
Happy to file upstream as well and link both ways.

If you want a stopgap before upstream moves, re-parsing a failed file after
inserting `;` before the closing brace of one-line type bodies would recover
these files — but I would not push for that over the warning. The silent part is
what actually cost time; a graph that is wrong and says so is a much smaller
problem.

## Environment

graphify 0.9.36 · tree-sitter 0.25.2 · tree-sitter-kotlin 1.1.0 · Python 3.11 ·
Linux · `extract --code-only` (AST only, no LLM backend).

abhay-codes07:
Addresses the second half of #2788.

**Scope, up front:** this fixes the *warning*, not the Astro parse. #2788 reports two things, and part 1 — the Astro extractor failing at line 1 on an ordinary `---` frontmatter fence — looks like it belongs to @RepairYourTech's open #2731, which gives `.svelte`/`.astro` a real AST pass via script masking. I have deliberately not touched that. Leaving the issue open for part 1.

## The bug

`extract.py` emits the partial-parse warning with a hardcoded issue reference, for every language:

```python
f"may be partially extracted: {_shown}{_more} (#2551)",
```

#2551 is **closed** (`state=CLOSED reason=COMPLETED`) and Kotlin-specific: *"Kotlin: bundled grammar rejects one-line type bodies"*. The surrounding comments already know this — one calls it "the genuine #2551 Kotlin one-line-body" case — but the citation is still attached to all of them, including the `.luau` case that has its own open issue (#2520) and the Astro case that had none.

The cost is not theoretical. From the report:

> We initially recorded this astro failure as "already tracked upstream" on the strength of that number, and only found otherwise by opening it.

The citation is the only lead the message offers, and it points at a resolved problem in a different language.

## The change

Drop the reference, and replace it with something the reader can act on. The issue asks for this directly:

> A count of what was dropped, even approximate, would make it actionable.

We cannot know what was *dropped*, but we know exactly what was *kept*, and that separates the two cases the warning currently blurs together:

```
before:
    website/src/pages/index.astro (first error at line 1) (#2551)

after:
    website/src/pages/index.astro (first error at line 1, no symbols extracted)
```

`no symbols extracted` means the file contributed nothing but its own file node — total loss for that file. `12 symbols extracted` means the parser recovered and an ERROR region dissolved some lines — partial, and probably lower priority. Those are very different situations and the old message rendered them identically.

The count is `len(nodes) - 1`, excluding the file's own node, which is the number a reader would reconstruct by hand anyway.

## What is unchanged

The gate is untouched: the same files warn as before, still governed by the `#2610`/`#2599` condition (`len(nodes) <= 1 or multiline_error`) that keeps fully-recovered TS/TSX noise silent. This only changes the wording of a warning that already fired.

`tests/test_ts_parse_warning.py` asserts the *absence* of `"syntax errors"` and `"partially extracted"` for valid files — both phrases are kept verbatim, so that guard still means what it meant.

## Tests

`tests/test_partial_extraction_warning.py`: the warning names the file and its first error line, reports `no symbols extracted` when only the file node survived and `N symbol(s) extracted` when more did, carries no `#2551`, and stays silent for a file that parses cleanly. The no-citation assertion is written against any `(#NNNN)` shape rather than the literal `2551`, so re-introducing a different hardcoded issue number fails too.

## Validation

- Full suite: `20 failed, 4469 passed` -> `20 failed, 4474 passed`. **Identical failure set** — no regressions; the +5 are the new tests.

### DIFF
--- a/graphify/extract.py
+++ b/graphify/extract.py
@@ -5621,19 +5621,36 @@ def extract(
         # results, so those fall back to the file-node-only arm.
         if len(_res.get("nodes", [])) <= 1 or _pe.get("multiline_error"):
             _rel = os.path.relpath(str(_p), str(root)).replace("\\", "/")
-            _syntax_error_files.append((_rel, _pe.get("first_error_line")))
+            # Symbols recovered from the file, excluding its own file node. This
+            # is what separates the two cases the warning otherwise blurs: a file
+            # that contributed nothing but its file node is a total loss, while
+            # one that yielded most of its symbols and lost an ERROR region is
+            # partial. "May be partially extracted" rendered both identically.
+            _kept = max(len(_res.get("nodes", [])) - 1, 0)
+            _syntax_error_files.append((_rel, _pe.get("first_error_line"), _kept))
     if _syntax_error_files:
+        def _describe_syntax_error(rel: str, line: "int | None", kept: int) -> str:
+            _where = f"first error at line {line}" if line else "syntax error"
+            _got = "no symbols extracted" if kept == 0 else f"{kept} symbol(s) extracted"
+            return f"{rel} ({_where}, {_got})"
+
         _shown = ", ".join(
-            f"{x} (first error at line {ln})" if ln else x
-            for x, ln in _syntax_error_files[:5]
+            _describe_syntax_error(*f) for f in _syntax_error_files[:5]
         )
         _more = (
             f" (+{len(_syntax_error_files) - 5} more)"
             if len(_syntax_error_files) > 5 else ""
         )
+        # No issue reference here. This message used to end in "(#2551)" for
+        # EVERY language, and #2551 is closed and Kotlin-specific ("bundled
+        # grammar rejects one-line type bodies"). It was the only lead the
+        # message offered, so a reader following it landed on a resolved problem
+        # in another language and concluded their own was already tracked
+        # (#2788). The file, the line and the symbol count are the actionable
+        # part; a single hardcoded number cannot be right for every grammar.
         print(
             f"  warning: {len(_syntax_error_files)} file(s) had syntax errors and "
-            f"may be partially extracted: {_shown}{_more} (#2551)",
+            f"may be partially extracted: {_shown}{_more}",
             file=sys.stderr, flush=True,
         )
 
--- a/tests/test_partial_extraction_warning.py
+++ b/tests/test_partial_extraction_warning.py
@@ -1,0 +1,82 @@
+"""The partial-extraction warning must be actionable and must not misdirect.
+
+It used to end in a hardcoded `(#2551)` for EVERY language. #2551 is closed and
+Kotlin-specific ("bundled grammar rejects one-line type bodies"), so a reader
+following the only lead the message offered landed on a resolved problem in a
+different language — and at least one did, recording an Astro failure as
+"already tracked upstream" on the strength of that number (#2788).
+
+The message also could not distinguish a file that contributed nothing but its
+own file node from one that yielded most of its symbols and lost an ERROR
+region. Both rendered as "may be partially extracted".
+
+Part 1 of #2788 — the Astro frontmatter parse itself — is NOT addressed here;
+that belongs to the .svelte/.astro AST work in #2731.
+"""
+import re
+
+import pytest
+
+from graphify.extract import extract
+
+# Any hardcoded issue citation, not just 2551 — re-introducing a different
+# number for a message that covers every grammar is the same mistake.
+_ISSUE_CITATION = re.compile(r"\(#\d+\)")
+
+
+def _partial_parse_fixture(tmp_path):
+    """A file the parser ACCEPTS only through ERROR recovery — an unclosed table
+    constructor swallowing the function that follows it.
+
+    Deliberately not the Kotlin one-line body from #2551: whether that shape
+    trips the gate depends on the bundled grammar build, and these tests are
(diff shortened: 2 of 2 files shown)


safishamsi:
The warning-citation fix (the second half of #2788) shipped in v0.9.47 via authorship-preserving cherry-pick. I have left #2788 open for the Astro frontmatter parse (the first half), which is tracked alongside #2731. Thanks @abhay-codes07! Release: https://github.com/Graphify-Labs/graphify/releases/tag/v0.9.47

PR #2820, opened 2026-08-17
fix(dedup): rewire hyperedge members onto survivors instead of dropping them (#2805)

### LINKED ISSUE #2805: build(): dedup drops hyperedge members instead of remapping to the survivor — silent group shrink at 0.9.44, dangling members at 0.8.29
## Summary

`build()` rewires **edges** to dedup survivors but never rewires **hyperedge members**. At HEAD
(0.9.44) the orphaned member is silently *dropped* rather than remapped, so a hyperedge quietly
loses a participant — and can fall below the 3-member threshold that makes it a hyperedge — with no
warning on stderr and no trace in the graph. On 0.8.29 the same input leaves a **dangling** member
instead. Both are wrong in the same place; the newer behaviour is harder to detect, because the
resulting graph passes any referential-integrity check.

## Mechanism

In `graphify/build.py` (line numbers from HEAD):

```python
1325    combined["hyperedges"].extend(ext.get("hyperedges", []))
...
1340    combined["nodes"], combined["edges"] = deduplicate_entities(
1341        combined["nodes"], combined["edges"], communities={},
1342        dedup_llm_backend=dedup_llm_backend, root=root,
1343    )
1344    return build_from_json(combined, directed=directed, root=root)
```

`deduplicate_entities` takes and returns only `(nodes, edges)`; `graphify/dedup.py` contains no
reference to hyperedges at all. `combined["hyperedges"]` is assembled at 1325 and is never passed
through the survivor mapping that rewrites edge endpoints. `deduplicate_by_label` in the same file
has the identical shape — its `remap` dict is applied to edges only.

`_normalize_hyperedge_members` / `_coerce_hyperedge_member_refs` normalise member *shape* (object
vs string, alias keys) but do not resolve members against surviving node ids, so they do not cover
this.

## Reproduction

Self-contained, no corpus or API key needed. Two nodes normalise to the same label, so dedup merges
them; the hyperedge names only the id that loses.

```python
import sys
import importlib.metadata as md

extraction = {
    "nodes": [
        {"id": "alpha_a", "label": "Alpha Concept", "file_type": "concept", "source_file": "notes/a.md"},
        {"id": "alpha_concept_long_variant_id", "label": "alpha concept", "file_type": "concept", "source_file": "notes/a.md"},
        {"id": "beta_node", "label": "Beta", "file_type": "concept", "source_file": "notes/a.md"},
        {"id": "gamma_node", "label": "Gamma", "file_type": "concept", "source_file": "notes/a.md"},
    ],
    "edges": [],
    "hyperedges": [{"id": "the_group", "label": "The Group",
                    "nodes": ["alpha_concept_long_variant_id", "beta_node", "gamma_node"],
                    "relation": "participate_in", "confidence": "INFERRED",
                    "confidence_score": 0.75, "source_file": "notes/a.md"}],
    "input_tokens": 0, "output_tokens": 0,
}
G = build([extraction])
ids = set(G.nodes())
for h in G.graph.get("hyperedges", []):
    m = h.get("nodes", [])
    print("graphify", md.version("graphifyy"), "->", m)
    print("  survivor present:", "alpha_a" in m,
          "| dangling:", [x for x in m if x not in ids],
          "| members:", len(m))
```

**0.9.44:**

```
[graphify] Deduplicated 1 node(s) (1 exact).
graphify 0.9.44 -> ['beta_node', 'gamma_node']
  survivor present: False | dangling: [] | members: 2
```

**0.8.29**, same script:

```
[graphify] Deduplicated 1 node(s) (1 exact).
graphify 0.8.29 -> ['alpha_concept_long_variant_id', 'beta_node', 'gamma_node']
  survivor present: False | dangling: ['alpha_concept_long_variant_id'] | members: 3
```

## Expected

The member should be remapped to the surviving id — `['alpha_a', 'beta_node', 'gamma_node']`, three
members, no dangling reference. The concept still exists in the graph under `alpha_a`; only the
hyperedge's pointer to it is stale.

## Observed impact

On a real corpus (docs + code, ~6.8k nodes) an incremental `build_merge` run deduplicated 151 nodes
(78 exact, 71 fuzzy) and produced **13 hyperedge members pointing at ids no longer in the graph**
across 11 hyperedges. One hyperedge retained 2 of its 5 members, which is a pairwise edge wearing a
hyperedge's name. That run was on 0.8.29, so the failure was at least visible; on 0.9.44 the same
corpus would instead have lost those 13 participations silently.

Nothing in the pipeline surfaces it either way. The merge prints its dedup count and success, and
hyperedge counts are unchanged, so the damage is invisible unless something separately validates
member ids against node ids.

## Suggested fix

Return the survivor map from `deduplicate_entities` (or expose it alongside), then apply it to
`combined["hyperedges"]` before `build_from_json`, deduping members in order afterwards since two
members can collapse onto the same survivor. The drop path should stay only for members that have
no survivor at all, and should warn on stderr naming the hyperedge — silent removal of a member is
what makes this hard to notice.

Two adjacent paths look like they need the same treatment, though I have not reproduced them
separately:

- `build_merge(prune_sources=...)` removes nodes via `G.remove_nodes_from()` without touching
  `G.graph["hyperedges"]`, so pruning a deleted file's nodes should orphan members the same way.
- A hyperedge that falls below 3 resolvable members after any of this is arguably no longer a
  hyperedge; currently only the zero-member case is dropped with a warning.

Related: #2484 (closed) fixed the same class for `merge-graphs`/`nx.compose`; this is the
`build()` / `deduplicate_entities` path, which appears not to have been covered by it.

## Environment

- graphify 0.9.44 (reproduced) and 0.8.29 (different symptom, same cause)
- Windows 11, Python 3.12, networkx via the package's own pins
- No LLM backend involved — `dedup_llm_backend` unset, pass-1 exact normalisation only

abhay-codes07:
Fixes #2805.

Thanks @vihavald — the diagnosis was exact, including the observation that the newer behaviour is the harder one to catch. That turned out to be the most important thing about this bug.

## The bug

`build()` rewires **edge** endpoints to dedup survivors. `combined["hyperedges"]` never went through the same remap — `dedup.py` contained no reference to hyperedges at all — so the member naming a merged-away id was simply absent from the rebuilt graph.

Your reproduction, verbatim:

```
before:  the_group: members=['beta_node', 'gamma_node']       # 3 -> 2
after:   the_group: members=['alpha_a', 'beta_node', 'gamma_node']
```

The reason this deserves the fix rather than a note in the docs is the part you flagged: **there is no dangling reference left behind.** The group loses a participant, can fall under the 3-member threshold that makes it a hyperedge at all, and the resulting graph passes any referential-integrity check cleanly. Nothing on stderr, nothing in the artifact. The 0.8.29 dangling-member behaviour was at least detectable.

You were also right that `_normalize_hyperedge_members` / `_coerce_hyperedge_member_refs` do not cover it — they normalise member *shape*, never resolving a member against surviving node ids.

## The change

`deduplicate_entities` takes an optional `hyperedges` list and rewires member ids in place, alongside the edge rewiring it already did.

Optional and mutating rather than added to the return tuple, so every existing two-tuple caller is unaffected — that seemed the right trade against changing a public-ish signature for one caller. Say the word if you would rather it returned a triple.

`deduplicate_by_label` has the identical shape and is deliberately left alone: nothing currently routes hyperedges through it, and fixing a path that has no caller would be speculative. Worth noting as a live footgun if that changes.

Handled: both member shapes (bare id and object carrying one), other fields on object members preserved, and member order kept so a rebuilt graph does not churn.

**One judgement call worth review.** Two members that remap onto the *same* survivor collapse to a single entry, so the group does shrink — but they were the same entity, which is the honest result. The old behaviour shrank the group *and* lost the participant. There is a test pinning this so it reads as intentional rather than as the original bug surviving in a corner.

## Tests

`tests/test_dedup_remaps_hyperedges.py` (14 tests): the member follows the survivor, the group keeps its size, no member points at a merged-away id, object-shaped members remap too, and an untouched hyperedge is byte-identical. Then the helper directly — collapse-onto-one-survivor, order preservation, other fields kept, an empty remap changing nothing, and five malformed-hyperedge shapes that must not raise.

Reverting the `build.py` call site while keeping the helper fails 3 of them. I checked it that way deliberately: reverting both files makes the module fail to import, which would have "failed" the whole file and told me nothing about which behaviour each test actually pins.

## Validation

- Full suite: `20 failed, 4469 passed` -> `21 failed, 4482 passed`. The +13 are the new tests.
- The extra failure is `test_incremental_mtime_collision.py::test_same_size_rewrite_in_one_tick_is_requeued`, untouched by this PR. It needs two writes in one filesystem timestamp tick; this run took 9:34 against a ~6:30 baseline, so the machine was loaded. Passes in isolation. It has now surfaced this way on four unrelated branches of mine, so I will open a separate issue for the flake rather than keep footnoting it.

### DIFF
--- a/graphify/build.py
+++ b/graphify/build.py
@@ -1340,6 +1340,9 @@ def build(
         combined["nodes"], combined["edges"] = deduplicate_entities(
             combined["nodes"], combined["edges"], communities={},
             dedup_llm_backend=dedup_llm_backend, root=root,
+            # Hyperedge members reference node ids too, so they need the same
+            # survivor rewiring the edges get (#2805).
+            hyperedges=combined.get("hyperedges"),
         )
     return build_from_json(combined, directed=directed, root=root)
 
--- a/graphify/dedup.py
+++ b/graphify/dedup.py
@@ -460,6 +460,46 @@ def _report_id_collision(nid: str, survivor: dict, losers: list[dict]) -> None:
 
 # ── main entry point ──────────────────────────────────────────────────────────
 
+def _remap_hyperedge_members(hyperedges: list[dict], remap: dict[str, str]) -> None:
+    """Rewire hyperedge member ids onto dedup survivors, in place.
+
+    Members come in both shapes the rest of the codebase tolerates — a bare id
+    string, or an object carrying one — so both are handled;
+    ``_normalize_hyperedge_members`` fixes the SHAPE but never resolves a member
+    against surviving node ids, which is why this is needed as well.
+
+    Two members that remap onto the same survivor collapse to one entry. That
+    shrinks the group, but honestly: they were the same entity, and the previous
+    behaviour dropped the loser without promoting it, which shrank the group
+    *and* lost the participant. Order is preserved so a rebuilt graph does not
+    churn.
+    """
+    for he in hyperedges:
+        if not isinstance(he, dict):
+            continue
+        members = he.get("nodes")
+        if not isinstance(members, list):
+            continue
+        seen: set = set()
+        rewired: list = []
+        for m in members:
+            if isinstance(m, str):
+                new_id = remap.get(m, m)
+                entry = new_id
+            elif isinstance(m, dict):
+                raw = m.get("id")
+                new_id = remap.get(raw, raw) if isinstance(raw, str) else raw
+                entry = dict(m, id=new_id) if new_id != raw else m
+            else:
+                new_id, entry = None, m
+            if isinstance(new_id, str):
+                if new_id in seen:
+                    continue
+                seen.add(new_id)
+            rewired.append(entry)
+        he["nodes"] = rewired
+
+
 def deduplicate_entities(
     nodes: list[dict],
     edges: list[dict],
@@ -467,6 +507,7 @@ def deduplicate_entities(
     communities: dict[str, int],
     dedup_llm_backend: str | None = None,
     root: str | Path | None = None,
+    hyperedges: "list[dict] | None" = None,
 ) -> tuple[list[dict], list[dict]]:
     """Deduplicate near-identical entities in a knowledge graph.
 
@@ -477,6 +518,9 @@ def deduplicate_entities(
         dedup_llm_backend: if set, use LLM to resolve ambiguous pairs
         root: scan root; ID-collision ranking judges source paths relative to
             it so path form and checkout location cannot flip the survivor (#2532)
+        hyperedges: when given, member ids are rewired to survivors IN PLACE,
+            the same way edge endpoints are. Optional and mutating rather than
+            returned so existing two-tuple callers are unaffected (#2805).
 
     Returns:
         (deduped_nodes, deduped_edges) with edges rewired to survivors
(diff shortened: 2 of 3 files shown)


safishamsi:

PR #2823, opened 2026-08-17
fix(extract): give AST INFERRED edges a rubric confidence_score (#2813)

### LINKED ISSUE #2813: AST "uses" edges always get INFERRED confidence_score=0.5, violating the documented rubric
graphify version: 0.9.45 (pipx install)

## Bug

Every edge with `relation: "uses"` produced by the structural (AST) extractor
is marked `confidence: "INFERRED"` with `confidence_score: 0.5` — no
exceptions. Checked across a full project graph: 21/21 "uses" edges are
flat 0.5.

This directly violates the confidence rubric in `references/extraction-spec.md`,
which says INFERRED edges must pick one of {0.55, 0.65, 0.75, 0.85, 0.95} and
"never 0.5, never as a default." All other relation types in the same graph
(`shares_data_with`, `semantically_similar_to`, `conceptually_related_to`,
produced by the semantic/LLM subagents) correctly follow the discrete rubric —
this is isolated to AST-generated `uses` edges specifically.

## Why this is wrong, not just inconsistent

The `uses` edges I checked are not actually uncertain — they're structurally
certain (e.g. a function with a type-annotated parameter `ledger: QuotaLedger`
that calls `ledger.charge(...)` in its body). AST already labels `imports` and
`calls` edges as `EXTRACTED` with `confidence_score: 1.0` for equivalent
structural certainty. `uses` edges from the same extractor should get the same
treatment — either `EXTRACTED`/`1.0`, or a real score from the rubric — not a
hardcoded 0.5 placeholder that makes 100%-certain relationships look like a
coin flip.

## Repro

Run `/graphify <path>` on any Python project with class instances passed
between functions, then:

    python -c "
    import json
    from collections import Counter
    c = Counter(e['confidence_score'] for e in g['links'] if e['relation']=='uses' and e['confidence']=='INFERRED')
    print(c)
    "

Expect: `Counter({0.5: N})` for all N "uses" edges.

## Suggested fix

In the AST extractor, either:
1. Emit `uses` edges as `EXTRACTED` / `1.0` (matches `imports`/`calls` treatment), or
2. Score them against the same discrete INFERRED rubric semantic subagents use.

abhay-codes07:
Fixes #2813.

## The bug

`references/extraction-spec.md` is unambiguous:

> `confidence_score` is REQUIRED on every edge - never omit it, never use 0.5 as a default
> - INFERRED edges: pick exactly ONE value from this set — never 0.5:
>   `0.95 / 0.85 / 0.75 / 0.65 / 0.55`

The AST extractor honoured neither half. Two sites emitted no `confidence_score` at all, so the edge fell through to `_CONFIDENCE_SCORE_DEFAULTS` and landed on exactly the 0.5 the rubric rules out. Three more hardcoded `0.8`, which is not in the discrete set.

On graphify's own package that is **128 of 128 INFERRED edges** off the scale:

```
before:  scores: [(0.8, 74), (None, 54)]      off-rubric or missing: 128
after:   scores: [(0.85, 113), (0.95, 15)]    off-rubric or missing: 0
```

The reporter measured 21/21 on their project; it is not a sampling artefact.

## Scores, and why each one

- **`uses`** (`extractors/resolution.py`) → **0.95**, the rubric's *"direct structural evidence (named cross-file reference)"*. That is literally what the edge is: a name this file imports, then references.
- **`indirect_call`** (`extractors/engine.py`, `extract.py`) → **0.85**, *"strong inference"*. The symbol link is direct — the function is named right there — but that it is ever *invoked* is the inference, so not the 0.95 tier.
- **unresolved cross-file `calls`** (`extract.py`, `symbol_resolution.py`) → **0.85**, the nearest rubric value to the 0.8 they already carried. Same tier, same meaning, now a value the documented scale contains.

**The confidence tiers are deliberately unchanged.** Your first suggestion was to promote AST `uses` edges to `EXTRACTED`/1.0, and I did not take it: that is a judgement about what the extractor actually knows, and it would move edges between tiers that the audit-trail percentages and every INFERRED-review workflow read. Snapping the scores onto the documented scale fixes the stated violation without making that call. Happy to follow up with the promotion if you want it.

## The default

`_CONFIDENCE_SCORE_DEFAULTS["INFERRED"]` moves `0.5` → `0.55`. The old value is the one the spec names as forbidden, and it is not on the INFERRED scale either. A missing score is an absence of evidence about *strength*, so the honest fallback is the weakest value the rubric allows rather than a midpoint that reads as a coin flip. Every AST site now supplies its own score, so this is a backstop rather than a routine path.

## Two existing tests changed — worth a second look

Both were encoding the values the spec forbids:

- `test_confidence.py` asserted `INFERRED default should be 0.5`.
- `test_symbol_resolution.py` asserted an expected edge with `confidence_score: 0.8`.

Flagging these explicitly rather than letting them ride in the diff, since changing an existing assertion is exactly where a fix can quietly redefine the thing it claims to fix.

## Tests

`tests/test_inferred_confidence_rubric.py` (10 tests): the default is neither 0.5 nor off-scale, the `EXTRACTED`/`AMBIGUOUS` defaults are untouched, no module hardcodes an off-rubric literal (checked as source text, so a future edit cannot reintroduce `0.8` or `0.5` quietly), and end-to-end that no INFERRED edge is left without a score. 4 fail without the fix.

## One thing to know when verifying by hand

The AST cache replays stored edges, so a **warm cache still serves pre-fix scores** — I chased 17 phantom "still missing" edges before realising that. It is namespaced by graphify version (#1252), so a release bump invalidates it automatically; purging `graphify-out/cache` is enough locally.

## Validation

- Full suite: `20 failed, 4469 passed` -> `21 failed, 4478 passed`. The +9 are the new tests.
- The extra failure is `test_id_normalization_contract.py::test_property_normalize_id_agrees_with_its_own_caseless_form`, and it is **not** caused by this PR. It is a Hypothesis property test that found a new Unicode counterexample during this run and cached it; replaying that stored example on a clean `v8` working tree fails identically. Nothing here touches `ids.py`. I am filing it separately rather than burying it here.

### DIFF
--- a/graphify/export.py
+++ b/graphify/export.py
@@ -157,7 +157,15 @@ from graphify.exporters.base import COMMUNITY_COLORS  # noqa: E402,F401
 from graphify.exporters.html import to_html  # noqa: E402,F401
 
 
-_CONFIDENCE_SCORE_DEFAULTS = {"EXTRACTED": 1.0, "INFERRED": 0.5, "AMBIGUOUS": 0.2}
+# Fallback scores for an edge that carries a confidence tier but no
+# confidence_score. The INFERRED default was 0.5, which references/extraction-spec.md
+# rules out in as many words — "never omit it, never use 0.5 as a default" — and
+# which is not in the discrete INFERRED set {0.55, 0.65, 0.75, 0.85, 0.95} either.
+# It is now the bottom of that set: a missing score is an absence of evidence
+# about strength, so the honest fallback is the weakest value the rubric allows,
+# not a midpoint that reads as a coin flip (#2813). Every AST emission site now
+# supplies its own score, so this is a backstop rather than a routine path.
+_CONFIDENCE_SCORE_DEFAULTS = {"EXTRACTED": 1.0, "INFERRED": 0.55, "AMBIGUOUS": 0.2}
 
 
 def attach_hyperedges(G: nx.Graph, hyperedges: list) -> None:
--- a/graphify/extract.py
+++ b/graphify/extract.py
@@ -6485,7 +6485,11 @@ def extract(
                     "relation": "indirect_call",
                     "context": rc.get("context", "argument"),
                     "confidence": "INFERRED",
-                    "confidence_score": 0.8,
+                    # 0.85, not 0.8: the rubric in references/extraction-spec.md
+                    # is a discrete set {0.55, 0.65, 0.75, 0.85, 0.95} and 0.8 is
+                    # not in it. Same tier, same meaning ("strong inference"),
+                    # now a value the documented scale actually contains (#2813).
+                    "confidence_score": 0.85,
                     "source_file": rc.get("source_file", ""),
                     "source_location": rc.get("source_location"),
                     "weight": 1.0,
@@ -6514,7 +6518,9 @@ def extract(
                 confidence_score = 1.0
             else:
                 confidence = "INFERRED"
-                confidence_score = 0.8
+                # 0.85 rather than 0.8 — the rubric's INFERRED set is discrete
+                # and does not contain 0.8 (#2813).
+                confidence_score = 0.85
             all_edges.append({
                 "source": caller,
                 "target": tgt,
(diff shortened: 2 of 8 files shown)


safishamsi:
Shipped in v0.9.47 via authorship-preserving cherry-pick. Thanks @abhay-codes07! Release: https://github.com/Graphify-Labs/graphify/releases/tag/v0.9.47

Issue #2834, opened 2026-08-17
`detect()`'s ignore-pattern list grows for the whole walk instead of being scoped to the current subtree — every file/directory check in a large monorepo scans every `.gitignore` ever seen, not just its own ancestors


sub4biz:
> Found alongside a second, unrelated instance of the same underlying mistake — recomputing something inside
> a loop instead of once outside it — in a completely different part of the pipeline (`extract_corpus_parallel`
> in `llm.py`, the semantic-extraction merge step, not the file-detection step below). Fixed separately in
> #2827. Flagging the connection in case it's useful context, since both were found in the same session while
> chasing why a full extraction run kept stalling for hours.
>
> cc @safishamsi — two independent performance issues found back-to-back, wanted to flag both together.

## Summary

`detect()` (`graphify/detect.py`) accumulates every nested `.gitignore`/`.graphifyignore` it discovers during
the `os.walk` traversal into two flat lists, `ignore_patterns` and `explicit_ignore_patterns`:

```python
if dp != root:
    ignore_patterns.extend(_load_dir_own_ignore(dp, gitignore=gitignore))
    explicit_ignore_patterns.extend(_load_dir_own_ignore(dp, gitignore=False))
```

These lists are appended to on every directory the walk visits and **never truncated** when the walk
backtracks out of a subtree. Correctness is preserved by an `anchor` mechanism (`_is_ignored`/`_eval` check
`target.relative_to(anchor)` and skip a pattern that doesn't apply to the current path), but performance is
not: by the time the walk reaches the Nth of many independent subprojects in a monorepo, every single
candidate file/directory check scans the *entire* pattern set accumulated from every subtree visited so far
— not just the O(depth) patterns that are actually its own ancestors.

This is separate from #1873 ("scope nested ignore patterns to their own subtree"), which is already applied
and fixes a **correctness** bug (a nested `.gitignore` incorrectly matching a sibling directory). This issue
is purely about **performance**: the same anchor-based correctness holds regardless of how large the pattern
list gets, but the list is allowed to grow unboundedly for the lifetime of one `detect()` call.

## Scale where this bites

A monorepo containing many independent projects (each with its own language-appropriate `.gitignore`) is
exactly the pathological shape. The real corpus this was found on has **274** such `.gitignore` files; the
reproduction below uses a synthetic **300**-directory fixture for a clean, self-contained benchmark. Either
way, the pattern list can reach several thousand entries by the later portion of the walk, and every one of
[...]

PR #2845, opened 2026-08-18
perf(detect): index ignore patterns by anchor; lru_cache fix

### LINKED ISSUE #1958: detect() CPU-bound 35+ min on large monorepo despite dir pruning; git ls-files discovery takes ~1s
## Environment
- graphifyy 0.9.17 (uv tool), Python 3.14, macOS 15 (Apple Silicon)
- Large monorepo: ~16k git-visible source files (~1M LOC), 37 nested `.gitignore` files, and heavy ignored dirs (`node_modules` 4.8GB, a build-cache dir with 4.8GB across hundreds of thousands of small text files)

## Symptom
`detect()` (so `/graphify .` and `graphify update`) pegs one core and did not complete within 35 minutes wall, 6+ minutes CPU. Sampling the process shows hot frames in `SRE_Pattern.match` with no source files open: the time goes to per-path ignore evaluation, not IO or word counting.

This persists with a correct dir-style `.graphifyignore` (`node_modules/`, `.git/`, `cache-dir/`, `dist/`, plus `**/*.md` excludes and one `!` negation), so the in-place directory pruning at `detect.py:1218-1238` does not rescue it. Unconfirmed suspects: per-path cost of the accumulated pattern list (root patterns plus 37 nested `.gitignore`s via `_load_dir_own_ignore`), and last-match-wins scans when a `!` negation is present.

For contrast, `git ls-files -co --exclude-standard` enumerates the same corpus in about 1s.

## Workaround
Bypassing `detect()` and synthesizing `.graphify_detect.json` from `git ls-files` output, then running extract/build/cluster as normal. The full pipeline finishes in ~112s. Extraction and clustering perform great (11k files AST in 63s, 62k-node build+cluster in ~20s). Discovery is the entire gap.

## Suggestion
A `--files-from <list>` or git-based discovery mode would sidestep this class of problem in git repos entirely. Happy to test a branch or share more profiling.

stephanGarland:
Fixes #1958 and #2834

detect() accumulates ignore patterns into one flat [(anchor_dir, pattern)] list and _is_ignored() rescanned all of it per path. Two costs fell out of that on a 356k-file corpus with 1,697 .gitignore files (57,286 patterns):

Anchor selection. The scan tested every anchor with target.relative_to(anchor). On CPython 3.12+ that scans target.parents linearly (_PathParents is a Sequence with no __contains__ override), minting a Path per element whose __eq__ raises and catches an AttributeError off an unset __slots__ member, then throws ValueError away for the ~97% of anchors that are not ancestors. Measured 813ms per path. _AnchorIndex buckets the list by anchor so a lookup walks the target's own ancestor chain instead — the same candidate set relative_to accepted, in list order so last-match-wins and ! negations are unaffected. The list is append-only during a walk, so the index refreshes over the new tail only.

Cyclic garbage. _match_anchored_ignore_pattern wrapped a recursive closure in its own @lru_cache, so every call built a reference cycle only the GC could reclaim (14 cyclic objects per call), and the cache was discarded before it could serve a second call. It is now a module-level function taking an explicit dict memo, with patterns lacking ** skipping the recursion entirely (54 of 57,283 patterns contain **).

Behavior is unchanged: differentially fuzzed against the previous implementations — 200k random plus exhaustive small-space pattern/path pairs for the matcher, and 3,000 random pattern sets x 40 files for _is_ignored, covering sibling anchors, negations and directory-only rules — 0 mismatches. On a synthetic 1,500-anchor / 30k-pattern corpus, 807ms -> 0.089ms per path, and 500 globstar matches now leave 0 cyclic objects.

Tests: 18 new in test_detect.py. Suite: 4,621 passed, 48 skipped, no regressions (test_labeling.py::test_label_communities_batches_when_over_batch_size and the four in test_ollama_retry_cap.py fail identically before this change).

### DIFF
--- a/graphify/detect.py
+++ b/graphify/detect.py
@@ -1336,28 +1336,61 @@ class _AnchorIndex:
 
     ``detect()`` extends the list as its walk finds nested ignore files, so
     ``_refresh`` indexes only the new tail.
+
+    Contract: the source list is append-only for as long as an index is live.
+    Every producer here obeys it — ``detect()`` and ``_is_scan_ignored`` only
+    ``append``/``extend``. ``_refresh`` still catches a shrink and a rewritten
+    tail (a same-length ``pop`` + ``append`` included) and rebuilds; an
+    *interior* rewrite is undetectable in O(1) and is what the contract buys.
+    Because reads go through ``entries``, even that degrades to a stale answer
+    rather than a wrong one.
     """
 
-    __slots__ = ("_patterns", "_by_anchor", "_indexed")
+    __slots__ = ("_patterns", "_entries", "_by_anchor")
 
     def __init__(self, patterns: list[tuple[Path, str]]) -> None:
         self._patterns = patterns
+        self._entries: list[tuple[Path, str]] = []
         self._by_anchor: dict[Path, list[int]] = {}
-        self._indexed = 0
         self._refresh()
 
+    @property
+    def entries(self) -> list[tuple[Path, str]]:
+        """The entries the buckets index, snapshotted as they were indexed.
+
+        Callers must read patterns from here, not from the source list: an
+        index is only ever consistent with what it indexed, and ``_is_ignored``
+        slices ``target.parts`` by ``len(anchor.parts)`` on the promise that the
+        bucket's anchor is an ancestor. Reading a source entry that has since
+        been swapped would break that promise and build a nonsense relative
+        path, so the snapshot keeps a stale index merely stale.
+        """
+        return self._entries
+
     def _refresh(self) -> None:
-        n = len(self._patterns)
-        if n == self._indexed:
+        patterns = self._patterns
+        n = len(patterns)
+        indexed = len(self._entries)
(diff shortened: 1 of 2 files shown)


sub4biz:
Ran this against our own test suite (test_detect.py + full pytest tests/) and a real ~84k-file corpus with 304 .gitignore — clean, identical baseline failures, no regressions.

We'd independently implemented our own fix for the #1958 half (module-level function + bounded lru_cache instead of a per-call memo dict) before seeing this PR, but since this covers both #2834 and #1958 together, we won't submit a duplicate — this looks like the way to go.

safishamsi:
Thanks @stephanGarland — we went with #3074 for the ignore-evaluation hot path (it eliminates relative_to entirely and needs no locked module-global memo), which shipped in v0.9.50. Your globstar-matcher reference-cycle fix was genuinely valuable and was ported into the landed change. Closing as superseded, with appreciation for the thorough perf work. Release: https://github.com/Graphify-Labs/graphify/releases/tag/v0.9.50

PR #2846, opened 2026-08-18
fix: prevent manifest timestamp churn on unchanged updates

### LINKED ISSUE #2838: manifest.json rewrites per-file "seen" timestamps on every run, so graphify-out/ is never clean and the pre-push hook never converges
**Version:** graphifyy 0.9.46 · Windows 11 · Python 3.12 (uv tool install)

## Summary

On unchanged code, `graph.json` is byte-identical across runs and graphify correctly reports `No code-graph topology changes detected; outputs left untouched`. But `manifest.json` is rewritten anyway, and the only thing that differs is the per-file `seen` timestamp.

Because the git hooks commit `graphify-out/` wholesale, `graphify-out/` is never clean, and **every `git push` leaves one trailing graph commit behind — permanently**.

## Reproduction

On a clean working tree, with no source change between runs:

```bash
graphify update . && sha256sum graphify-out/graph.json
graphify update . && sha256sum graphify-out/graph.json
graphify update . && sha256sum graphify-out/graph.json
```

`graph.json` is identical all three times (same sha256), and runs 2 and 3 print `No code-graph topology changes detected; outputs left untouched`.

Now diff two consecutive graph commits made by the hooks:

```
graphify-out/manifest.json | 772 ++++++++++++++++--------------
1 file changed, 386 insertions(+), 386 deletions(-)
```

`graph.json`, `GRAPH_REPORT.md` and `.graphify_labels.json` do not appear in the diff. Every changed line is a `seen` field — verified, zero changed lines are anything else:

```json
"app/admin-ui/src/App.tsx": {
  "mtime": 1785366505.196617,
- "seen": 1787000934.3112164,
+ "seen": 1787001063.5783465,
  "ast_hash": "f8f7775881b973daf0c485283570a6c1",
  "semantic_hash": "f8f7775881b973daf0c485283570a6c1"
}
```

386 entries, one line changed each.

## Impact

The `pre-push` hook sees `graphify-out/` as dirty, commits, and reports:

```
[graphify] Il grafo cambia a ogni run: committato, push lasciato passare.
           Da guardare: 'graphify update .' non converge su questo repo.
```

Two problems with that:

1. **It never converges.** Observed three pushes in a row: the hook commits, the push transfers up to the pre-hook HEAD, and the newly created commit stays local. Re-running `git push` just produces another one. The repo sits permanently one graph commit ahead of origin.
2. **The message points at the wrong thing.** It says the *graph* changes, which sends you looking for nondeterminism in extraction or clustering. The graph is fine — it's cache bookkeeping. I spent a while bisecting node/edge sets and reading `_rewire_unique_stub_nodes` and `cluster()` before diffing the manifest.

## Suggested fix

Any of these would do:

- Don't rewrite `seen` when no other field in the entry changed — or skip writing `manifest.json` entirely on the `outputs left untouched` path.
- Or exclude `manifest.json` from the hook's change detection, since it's cache bookkeeping rather than graph content.
- Or drop `!graphify-out/manifest.json` from the `.gitignore` template that `graphify install` writes, so it isn't versioned in the first place.

## Workaround

Untracking it fixes it immediately, and the hook converges on the first try:

```bash
git rm --cached graphify-out/manifest.json
# remove the "!graphify-out/manifest.json" allowlist line from .gitignore
```

Verified on two repos: after this, `git push` completes with zero trailing commits and a clean working tree.

hopstreax:
## Summary

Fixes #2838.

`graphify update` was rewriting `graphify-out/manifest.json` on every run by updating the per-file `seen` timestamp, even when the source files and graph topology were unchanged.

This caused `graphify-out/` to remain dirty indefinitely and made the pre-push hook create a trailing graph commit on every push.

### What changed

- Preserve the existing `seen` timestamp for genuinely unchanged manifest entries.
- Generate a new `seen` timestamp when a file is new or its relevant content/hash state changes.
- Avoid rewriting `manifest.json` when the final manifest payload is identical to the existing file.
- Preserve existing manifest maintenance behavior for:
  - excluded-file pruning (#1908)
  - failed AST cleanup (#2543)
  - semantic hash cleanup (#1948)

### Validation

Verified that:

- unchanged files preserve their existing `seen` timestamp
- changed files receive a new `seen` timestamp and updated hash
- same-mtime/same-tick rewrites are still detected by incremental detection
- AST → semantic tier transitions continue to work
- manifest pruning and clear directives remain intact
- malformed and legacy manifests are handled safely
- consecutive no-op `graphify update .` runs leave `manifest.json` byte-identical
- `graphify-out/` remains clean after consecutive no-op updates
- the pre-push graph commit workflow now converges

Relevant tests pass, including the manifest, incremental mtime collision, partial cache, and atomic write suites.

The remaining Windows-specific failures in the broader suite were reproduced on the parent commit and are unrelated to this change.

### DIFF
--- a/graphify/detect.py
+++ b/graphify/detect.py
@@ -2070,20 +2070,33 @@ def save_manifest(
         mtime, h = hashed[f]
         key = _nfc(f)
         prev = _normalise_entry(existing.get(key, {})) or {}
-        # seen: when this row was written. If the file's mtime sits inside the
-        # same filesystem tick, a later same-length edit can land in that tick
-        # without moving mtime, so the mtime-unchanged fastpath cannot prove
-        # the content is still current and detect_incremental re-hashes.
-        entry: dict = {"mtime": mtime, "seen": time.time()}
         if kind in ("ast", "both"):
-            entry["ast_hash"] = h
+            ast_h = h
         else:
-            entry["ast_hash"] = prev.get("ast_hash", "")
+            ast_h = prev.get("ast_hash", "")
         if kind in ("semantic", "both"):
-            entry["semantic_hash"] = h
+            sem_h = h
         else:
             # Preserve semantic_hash only when content is unchanged
-            entry["semantic_hash"] = prev.get("semantic_hash", "") if h == prev.get("ast_hash", "") else ""
+            sem_h = prev.get("semantic_hash", "") if h == prev.get("ast_hash", "") else ""
+
+        # Preserve previous seen timestamp if the entry's mtime and target hash(es)
+        # are genuinely unchanged and no clear was requested for this file.
+        prev_seen = prev.get("seen")
+        is_unchanged = (
+            isinstance(prev_seen, (int, float))
+            and mtime == prev.get("mtime")
+            and (ast_h == prev.get("ast_hash", "") if kind in ("ast", "both") else True)
+            and (sem_h == prev.get("semantic_hash", "") if kind in ("semantic", "both") else True)
+            and not _in_clear_ast(f)
+            and not _in_clear(f)
+        )
+        entry: dict = {
+            "mtime": mtime,
+            "seen": prev_seen if is_unchanged else time.time(),
+            "ast_hash": ast_h,
+            "semantic_hash": sem_h,
+        }
         manifest[key] = entry
     if root is not None:
         # Persist in portable form: forward-slash relative paths. Keys outside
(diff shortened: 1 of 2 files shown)


safishamsi:
Shipped in v0.9.47 via authorship-preserving cherry-pick, with added no-op-then-change and corrupt-manifest tests. Thanks @hopstreax! Release: https://github.com/Graphify-Labs/graphify/releases/tag/v0.9.47

Issue #2194, opened 2026-07-25
`build_from_json` doesn't canonicalize `name`/`path`/`type`/`confidence_score` aliases, so schema errors in a persisted graph.json are permanent and hide nodes from query


drmzperx:
## Summary

`build_from_json()` canonicalizes *some* legacy field aliases before validation, but not others. Nodes and edges persisted in a `graph.json` that use the unhandled aliases fail validation on **every** subsequent build, and because `graphify update` merges *from* `graph.json`, the errors are permanent — no amount of re-running fixes them.

Worse than the warning noise: affected nodes end up with `norm_label == ""`, which makes them unmatchable by `query`/`explain`.

## Version

Reproduced on 0.8.13. **The gap is still present in current `main`** (0.9.26) — verified by reading `graphify/build.py` and `graphify/validate.py` from this repo, not by re-running, so the reproduction numbers below are from 0.8.13.

## What happens

A real graph (Java/Spring backend, 9,019 nodes / 17,394 edges) emitted this on every build:

```
[graphify] Extraction warning (1911 issues): Node 8537 (id='repo:PIFAnnexAIVerdictRepository') missing required field 'label'
```

Breaking the 1,911 errors down:

| count | error |
|------:|-------|
| 864 | missing `source_file` (496 nodes + 368 edges) |
| 370 | node missing `label` |
| 368 | edge missing `confidence` |
| 309 | edge missing `relation` |

Every one of the 370 label-less nodes had `file_type: "concept"` and carried the data under a different key:

| `validate.py` requires | what was persisted |
|---|---|
| `node.label` | `node.name` (345 nodes; the other 25 only had `id`) |
| `node.source_file` | `node.path` (149 nodes) |
| `edge.relation` | `edge.type` (309 edges — all 309 had `type`) |
| `edge.confidence` | `edge.confidence_score` (float; all 368 had it) |

## Where the gap is

`build.py`'s canonicalization block runs before `validate_extraction()` and already handles several aliases:

- `node.source` → `node.source_file` (with a warning, #479)
- `node.file_type` empty/unknown → `"concept"` / `_FILE_TYPE_SYNONYMS` (#660)
- `edge.from`/`edge.to` → `edge.source`/`edge.target`
- hyperedge `members`/`node_ids` → `nodes` (#1561)

[...]

PR #2854, opened 2026-08-18
fix(build): normalize legacy numeric edge confidences to INFERRED on load

Trantor-develops:
## The defect

Graphs written by graphify versions before the confidence enum contract (`EXTRACTED` / `INFERRED` / `AMBIGUOUS`) carry **numeric** edge confidences — `1.0`, `0.95`, `0.9`, `0.85` — from the era when the LLM semantic pass wrote its float directly into `confidence`. The current validator rejects those values, so every load of a pre-enum graph re-warns once per legacy edge, forever. Real-world case: a June-2026 graph produced **73 warnings on every load**.

## The fix

Extend `_fold_edge_aliases` in `graphify/build.py` — the existing #2194 pre-validation canonicalization point — so a numeric `confidence`:

- normalizes to **`INFERRED`**, and
- moves its original float into the **`confidence_score`** companion field (an explicit `confidence_score` already on the edge wins; `bool` is excluded since `True` is not a score).

Because every validating load path (extract, `update`/`build_merge`, cluster-only reload, label, diagnostics) funnels through `build_from_json`, which remaps `links`→`edges` and folds aliases before `validate_extraction`, one shim heals them all — and the next write persists the normalized enum, so read-only surfaces (`serve`/query) see clean values from then on.

## Mapping rationale

Any numeric value maps to `INFERRED`: numeric confidences only ever came from the LLM semantic pass, and LLM-derived edges are `INFERRED` by definition. This mirrors the existing `confidence_score`-alias backfill's own rule ("never EXTRACTED — alias recovery is not provenance", never a threshold mapping of the float).

## Tests (Red→Green)

Three new tests in `tests/test_build.py` next to the #2194 alias-fold family:

- `test_legacy_numeric_confidence_normalized_to_inferred` — numeric-confidence extraction loads with **no** `invalid confidence` warning; edges land `INFERRED` with the float preserved in `confidence_score` (failed before the fix with the exact reported warning, passes after)
- `test_legacy_numeric_confidence_existing_score_wins` — an explicit `confidence_score` is not overwritten
- `test_legacy_numeric_confidence_links_spelling_reload` — the on-disk shape of the defect: a NetworkX-serialized `links`-keyed graph.json reloads silently

Full suite: 4606 passed, 48 skipped; the only 5 failures (`test_ollama_retry_cap.py` ×4, one ordering-dependent `test_labeling` case) fail identically on a clean `origin/v8` checkout on this host — pre-existing, unrelated. `ruff check` clean. CHANGELOG entry added under 0.9.46 (unreleased).

Note: opened from a fork — the authenticated account has read-only access to this repo.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,8 @@ Full release notes with details on each version: [GitHub Releases](https://githu
 
 ## 0.9.46 (unreleased)
 
+- Fix: a `graph.json` written by a pre-enum graphify version, whose edges carry a numeric `confidence` (1.0/0.95/0.9/0.85) instead of `EXTRACTED`/`INFERRED`/`AMBIGUOUS`, now loads silently instead of warning once per legacy edge on every load (a June-2026 graph produced 73 warnings per load). The number normalizes to `INFERRED` — numeric confidences only ever came from the LLM semantic pass, and LLM-derived edges are INFERRED by definition — and the original float is preserved in the `confidence_score` companion field (an explicit `confidence_score` already on the edge wins).
+
 - Fix: node-id normalization is now caseless-stable for combining-mark sequences — `casefold` and NFKC don't commute, so a single pass left `normalize_id(s) != normalize_id(s.casefold())` for inputs like Greek ypogegrammeni followed by a combining accent; normalization now iterates casefold+NFKC to a fixpoint. Letter/digit-bearing ids are unchanged, so existing graphs are not re-keyed.
 
 - Fix: Java annotations now emit `references` edges to their type — including class-literal arguments (`@Repeatable(Foo.class)`, `@Uses({A.class, B.class})`) and annotation-member return types — so a container annotation is no longer a disconnected island; string/enum arguments are not mistaken for type references (#2426, thanks @oleksii-tumanov).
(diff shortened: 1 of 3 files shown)


Trantor-develops:
Independent review (Claude, orchestrating for Jerome): **approve**.

- Shim verified at the right layer: `_fold_edge_aliases` runs pre-validation on every loading path (`build_from_json`), so extract/update/label/cluster-only/diagnose all heal legacy graphs; non-validating readers never warned.
- Semantics correct: numeric → `INFERRED` matches the field's provenance meaning (numerics only ever came from the LLM pass); float preserved in `confidence_score` with an existing explicit score taking precedence; `bool` excluded.
- Fresh test run on a clean worktree of this branch: `tests/test_build.py` 83/83 pass (3 new). Real-world validation: the 73-warning ai-mortgage-agent graph that motivated this loads clean after the equivalent data fix.
- Merge attempted from this machine but the account lacks write access — needs a maintainer click.

Trantor-develops:
Withdrawing — we've resolved this on our side by normalizing our graph data, and we're not set up to maintain an upstream contribution right now. Feel free to reuse the patch if the normalization is useful; it's tested against the case described above.

PR #2858, opened 2026-08-18
fix: detect stale and incomplete graph state

### LINKED ISSUE #2841: No detection of a stale or partially-built graphify-out/ for git-tracked corpora
**Summary**

The skill's only existing-graph check (top of `SKILL.md`) is "does `graphify-out/graph.json` exist," used to decide whether to take the query-only fast path. For a git-tracked corpus, this has two blind spots:

1. **Staleness**: `graph.json` carries a `built_at_commit` field, but nothing compares it against the corpus's current `HEAD` before trusting an existing graph (fast path) or before starting a fresh build. A graph built at an old commit can be materially wrong — missing entire subsystems added since — while still passing every other health check.
2. **Interrupted runs**: if a prior invocation reached Step 4 (graph built, report generated) but stopped before Step 5/6/9 (labeling, HTML export, cleanup), the `graphify-out/` directory is left with orphaned `.graphify_chunk_*.json` / `.graphify_extract.json` / etc. temp files and a report full of generic `Community N` labels. Nothing in the skill detects or surfaces this state; a subsequent run (or a human) has to notice by inspection.

**What happened**

I found exactly this: an existing `graphify-out/` with a fully-formed `graph.json` and `GRAPH_REPORT.md`, generic community labels, no HTML, and five leftover `.graphify_*.json` temp files. `built_at_commit` pointed to a commit from before a significant amount of subsequently-merged work — the existing graph was silently missing an entire subsystem. Diagnosing this took several manual steps (checking `built_at_commit` against `git log`, checking for an active writer process via `lsof`/`ps`, inspecting the detect.json file-type breakdown) that the skill gave no guidance for.

I chose to discard the existing state and rebuild fresh in an isolated worktree rather than trust or resume it, given the ambiguity.

**Proposed fix**

- At the top of `SKILL.md`, before the fast-path check: if the corpus root is a git repository, compare `graph.json`'s `built_at_commit` against `git rev-parse HEAD` (or the working tree's dirty state). If they differ, do not silently take the fast path — surface the staleness and suggest `--update` or a rebuild.
- If `graphify-out/` contains leftover `.graphify_chunk_*.json` / `.graphify_extract.json` / `.graphify_ast.json` / `.graphify_semantic.json` / `.graphify_analysis.json` at the *start* of a new invocation (i.e. before this run created them), that's diagnostic of an interrupted prior run. Surface this explicitly rather than letting Step 1-3 silently overwrite or ignore them.
- This matters more than it might look: any workflow where multiple processes/sessions could plausibly invoke `/graphify` against a shared working directory (CI, multi-agent fleets, a team repo) will hit this.

hopstreax:
## Summary

Fixes #2841.

Graphify previously treated the presence of `graphify-out/graph.json` as
sufficient to take the existing-graph fast path. This could cause stale graphs
or partially completed builds to be silently trusted.

This change introduces explicit graph-state inspection and integrates it into
the skill build lifecycle.

### Graph state inspection

Added `inspect_graph_state()` with six states:

- `ABSENT` — no existing graph
- `BUILDING` — another Graphify build is currently active
- `INCOMPLETE` — intermediate artifacts from an interrupted build remain
- `STALE` — the graph no longer represents the current corpus
- `UNVERIFIABLE` — freshness cannot be established
- `FRESH` — graph provenance and working tree are current

For git-tracked corpora, freshness considers:

- `graph.json`'s `built_at_commit` vs current `HEAD`
- relevant tracked modifications/deletions/renames
- supported untracked source files
- git ignore rules

Graphify's own output directory and unrelated unsupported files are excluded
from dirty-tree detection.

### Interrupted-build detection

Known ephemeral artifacts are detected and surfaced as `INCOMPLETE`, including:

- `.graphify_extract.json`
- `.graphify_ast.json`
- `.graphify_semantic.json`
- `.graphify_chunk_*.json`
- `.graphify_analysis.json`
- `.graphify_detect.json`

Persistent sidecars are explicitly excluded.

### Unified build locking

The skill workflow and existing rebuild paths now share the same build-lock
protocol.

The lock provides:

- cross-process mutual exclusion
- PID ownership
- process creation-time verification for PID reuse
- stale/corrupt lock recovery
- crash-safe lock release
- token-based ownership during release
- protection against an old owner deleting a newer owner's lock

The implementation supports POSIX and Windows locking primitives.

### Skill integration

The existing-graph fast path now inspects graph state before deciding whether
to query or rebuild.

The build workflow also:

1. acquires the build lock
2. re-inspects graph state using its ownership token
3. avoids unnecessary rebuilding if another process already produced a fresh graph
4. safely releases the lock on completion or failure

Generated platform skills and skillgen snapshots were updated accordingly.

## Tests

Added comprehensive graph-state/build-lock coverage including:

- HEAD mismatches
- dirty tracked files
- supported untracked files
- ignored/irrelevant files
- missing provenance
- non-git repositories
- interrupted-build artifacts
- active builds
- stale lock recovery
- PID reuse
- concurrent lock contention
- release races
- owner-aware reinspection

Validation:

- `49` graph-state tests passed
- `66` skillgen tests passed
- `115` watch tests passed locally, with `9` skipped and `2` pre-existing Windows-specific failures
- Linux CI-equivalent suite: `4693 passed, 30 skipped, 19 pre-existing Windows/environment failures`
- `git diff --cached --check` clean

The existing watch-lock tests were also updated to validate the new
backward-compatible lock format: PID remains on line 1 while ownership
metadata is stored on subsequent lines.

### DIFF
--- a/graphify/detect.py
+++ b/graphify/detect.py
@@ -10,10 +10,16 @@ import subprocess
 import time
 import unicodedata
 from concurrent.futures import ThreadPoolExecutor
+import contextlib
+from dataclasses import dataclass, field
+from datetime import datetime, timezone
 from enum import Enum
+import errno
 from functools import lru_cache
 from pathlib import Path
-from typing import Callable
+import sys
+from typing import Callable, Any
+import uuid
 
 from graphify.google_workspace import (
     GOOGLE_WORKSPACE_EXTENSIONS,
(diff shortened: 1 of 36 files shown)


jecanore:
Pulled this into a local staleness-patches branch (on top of v8 + #2064 + #3108) — it's the one that actually surfaces whether a graph is trustworthy, which I wanted before relying on graphify output day to day. Had one trivial merge conflict against the other patches (both added tests/imports at the same spot in detect.py/test_skillgen.py), nothing in the logic itself. Nice work on the FRESH/STALE/INCOMPLETE state modeling.

PR #2864, opened 2026-08-19
fix(obsidian): keep non-ASCII letters in community tags, and query the same tag from graph-view colour groups

### LINKED ISSUE #2862: Obsidian export: non-ASCII community labels collapse to `community/_` tags, and canvas colorGroups query the unsanitized label
## Summary

`_obsidian_tag()` strips every non-ASCII character, so a community label written in any
non-Latin script collapses to underscores. Every note in such a community gets the *same*
tag, and the per-community Dataview query in the `_COMMUNITY_*.md` note points at that
collapsed tag. Separately, the canvas `colorGroups` query is built from the **unsanitized**
label, so on the same vault the canvas asks for a tag that no note carries.

## Repro (current HEAD, also 0.9.11)

```python
from graphify.export import _obsidian_tag
_obsidian_tag("참여기업 로고 자산")     # -> '__'      (Korean)
_obsidian_tag("発注者 体系")            # -> '_'       (Japanese)
_obsidian_tag("Träning Pipeline")      # -> 'Trning_Pipeline'   (accented Latin dropped)
_obsidian_tag("Attention Mechanism")   # -> 'Attention_Mechanism'  (ASCII is fine)
```

`graphify/export.py`:

```python
def _obsidian_tag(name: str) -> str:
    return re.sub(r"[^a-zA-Z0-9_\-/]", "", name.replace(" ", "_"))   # line 106
```

and the canvas group query does not go through it:

```python
"query": f"tag:#community/{label.replace(' ', '_')}",                 # line 835
```

## Impact (measured)

A 317-node vault with 20 Korean community labels: **290 of 337 notes** ended up tagged
`community/_` or `community/__`. Tag-based navigation and the generated
`TABLE ... FROM #community/<tag>` query stop distinguishing communities, and the canvas
color groups match nothing because they query `#community/참여기업_로고_자산` while the
notes carry `#community/__`.

The `community:` frontmatter field keeps the readable name, so the data is not lost — only
the navigation surface is.

## Why the sanitizer is stricter than it needs to be

Obsidian tags accept letters from any language (they exclude spaces and most punctuation,
and a tag cannot be digits-only). The docstring's premise — "alphanumerics, hyphens,
underscores and slashes only" — holds for ASCII punctuation but not for scripts.

## Suggested fix

```python
def _obsidian_tag(name: str) -> str:
    # \w is Unicode-aware in Python 3: keeps letters/digits/underscore in any script,
    # still strips spaces (already converted), '#', '·', and other punctuation.
    tag = re.sub(r"[^\w\-/]", "", name.replace(" ", "_"))
    return tag or "unnamed"          # digits-only / empty labels are invalid tags
```

and make the canvas use the same function so notes and canvas agree:

```python
"query": f"tag:#community/{_obsidian_tag(label)}",
```

Happy to send a PR with the two-line change plus a test that asserts a Korean and an
accented-Latin label survive and that canvas/query use the identical string.

josh-leung-KR:
Fixes #2862.

## What was wrong

`_obsidian_tag()` filtered on `[^a-zA-Z0-9_\-/]`, which drops every non-ASCII character.
A community label written in any non-Latin script therefore collapsed to underscores, so
**every note in that community carried the same tag** and per-community navigation stopped
working. Accented Latin lost its letters too (`Träning` → `Trning`).

Separately, the `.obsidian/graph.json` colour-group query was built from the **raw** label
(`label.replace(' ', '_')`), so it asked for `#community/참여기업_로고_자산` while the notes
were tagged `#community/__` — the colour groups matched nothing on exactly the vaults the
first bug had already flattened.

Measured on a real vault: 20 Korean community labels, **290 of 337 notes** tagged
`community/_` or `community/__`.

## The change

- `_obsidian_tag()` now filters on `[^\w\-/]`. Python 3's `\w` is Unicode-aware, so Hangul,
  CJK, Cyrillic and accented Latin survive while spaces (already converted) and punctuation
  (`·`, `#`, …) are still stripped. Slashes are kept, so tag nesting is unchanged.
- Two degenerate cases that would otherwise emit an invalid tag: a punctuation-only label
  returns `unnamed`, and a digits-only label is prefixed (`2026` → `c2026`) because Obsidian
  ignores digits-only tags.
- The graph-view colour-group query now goes through the same sanitizer, so canvas/graph
  config and note tags cannot drift apart again.

ASCII labels are unaffected: `"Attention Mechanism"` → `Attention_Mechanism` as before.

## Tests

`tests/test_obsidian_unicode_tags.py` (5 cases). Against the current sanitizer **4 of the 5
fail**; all 5 pass with this change:

| test | before | after |
|---|---|---|
| non-Latin + accented labels keep their letters | FAIL | PASS |
| distinct non-Latin communities get distinct tags | FAIL | PASS |
| graph-view colour group matches the note tag | FAIL | PASS |
| degenerate labels stay valid tags | FAIL | PASS |
| punctuation is still stripped (guard against over-permissiveness) | PASS | PASS |

I verified the behaviour by running the new tests against a patched and an unpatched module;
the table above is that run, not a prediction.

### DIFF
--- a/graphify/export.py
+++ b/graphify/export.py
@@ -100,10 +100,19 @@ def backup_if_protected(out_dir: Path) -> "Path | None":
 def _obsidian_tag(name: str) -> str:
     """Sanitize a community name for use as an Obsidian tag.
 
-    Obsidian tags only allow alphanumerics, hyphens, underscores, and slashes.
-    Spaces become underscores; everything else is stripped.
+    Obsidian tags accept letters from any language plus digits, hyphens,
+    underscores and slashes; spaces and most punctuation are not allowed, and a
+    tag cannot be digits-only. ``\w`` is Unicode-aware in Python 3, so Hangul,
+    CJK, Cyrillic and accented Latin survive instead of being stripped (#2862):
+    an ASCII-only filter collapsed every non-Latin community label to
+    underscores, so every note in that community carried the same tag.
     """
-    return re.sub(r"[^a-zA-Z0-9_\-/]", "", name.replace(" ", "_"))
+    tag = re.sub(r"[^\w\-/]", "", name.replace(" ", "_"))
+    if not tag.strip("_-/"):
+        return "unnamed"          # label was punctuation only
+    if tag.isdigit():
+        return f"c{tag}"          # Obsidian ignores digits-only tags
+    return tag
 
 
 def _strip_diacritics(text: str | None) -> str:
@@ -832,7 +841,10 @@ def to_obsidian(
     graph_config = {
         "colorGroups": [
             {
-                "query": f"tag:#community/{label.replace(' ', '_')}",
+                # Same sanitizer as the note tags (#2862): built from the raw
+                # label, the canvas colour group queried a tag that no note
+                # carries whenever the label held non-ASCII or punctuation.
+                "query": f"tag:#community/{_obsidian_tag(label)}",
                 "color": {"a": 1, "rgb": int(COMMUNITY_COLORS[cid % len(COMMUNITY_COLORS)].lstrip('#'), 16)}
             }
             for cid, label in sorted((community_labels or {}).items())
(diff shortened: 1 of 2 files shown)


safishamsi:
Shipped in v0.9.47 via authorship-preserving cherry-pick, with the tag-sanitizer docstring made a raw string to clear an invalid-escape warning. Thanks @josh-leung-KR! Release: https://github.com/Graphify-Labs/graphify/releases/tag/v0.9.47

PR #2869, opened 2026-08-19
perf(detect): hoist per-pattern work out of the ignore-matching loop

Alond:
## What

`_is_ignored` re-did the same work for every `(target, pattern)` pair, so the cost scaled with `patterns × targets` even though most of that work depends on neither the pattern nor the target:

- **pattern parsing** (`startswith("!")`, `endswith("/")`, `strip("/")`) does not depend on the target at all, yet ran once per target per pattern;
- **`target.relative_to(anchor)` + `_nfc()`** depend only on the anchor — anchors are few, patterns are many;
- **`target.relative_to(root)` + `_nfc()`** depend on neither;
- **`target.is_dir()`** (an `lstat`) ran for every directory-only pattern instead of once per target.

The parse is now done once per scan and kept in the shared cache, the relative paths are memoised per target, and `is_dir()` is lazy. No behaviour change intended.

## Numbers

Measured on an iOS repo: 3614 detected files, ~55 ignore patterns across four `.gitignore` files, 74k files on disk (60k of them under a git-ignored `.build/`).

| | before | after |
|---|---|---|
| `detect()` | 5.34s | **2.24s** (2.4×) |
| `graphify update .` | 13.05s | **9.89s** |

Profiling the original pointed at exactly this loop: 762k `Path.relative_to` calls, 4.6M `fnmatch`, 1.36M `lstat`, with `detect.py:_matches` at 5.8s cumulative.

## Correctness

`detect()` output is **byte-identical** before and after — I compared the full result dict (`files`, `ignored`, `unclassified`, …), not just counts. Counts alone would not have caught the bug below.

Full suite: **the same 25 pre-existing failures before and after**, no new ones. `tests/test_detect.py` and `tests/test_ignore_file_encoding.py`: 261 passed.

## The trap worth reviewing carefully

Caching the parse needs invalidation, because `detect()` appends patterns from nested `.gitignore` files **while the walk is already running** (`ignore_patterns.extend(...)` inside the walk loop). My first version cached the parse unconditionally and silently stopped applying the nested rules — and the detected **file count stayed the same**, so it looked fine.

The cache key is therefore `len(patterns)`, and `tests/test_detect.py::test_nested_gitignore_applies_when_patterns_grow_mid_walk` covers it. That test needs a root-level pattern in the fixture: without one, `_is_ignored` returns before touching the cache and the bug cannot reproduce. I verified the test fails with the guard removed and passes with it — an earlier version of the test passed in both directions and was worthless.

## Notes

- Two sentinels (`_PREPARED_KEY`, `_UNSET`) are module-level; the string cache key cannot collide with the `Path` keys the ancestor memo uses.
- The gain should scale with repo size — more patterns and more files both multiply into the removed work.

### DIFF
--- a/graphify/detect.py
+++ b/graphify/detect.py
@@ -1331,14 +1331,20 @@ def _is_ignored(
         # keep it in the shared cache. It used to run for every target: on a repo with
         # ~55 patterns and ~14k targets that is ~770k redundant string splits.
         #
-        # Invalidate on pattern count: detect() appends patterns from nested .gitignore
-        # files *during* the walk (see ignore_patterns.extend below), so a frozen parse
-        # would silently stop applying the later ones.
+        # The entry is keyed on the pattern list's identity AND its length, and it
+        # keeps a reference to that list:
+        #   - length, because detect() appends patterns from nested .gitignore files
+        #     *during* the walk (see ignore_patterns.extend below), so a frozen parse
+        #     would silently stop applying the later ones;
+        #   - identity, because a caller may pass a different list of the same length
+        #     with the same _cache, and length alone would serve the wrong parse.
+        # Holding the list keeps it alive, so an `is` check cannot be fooled by a new
+        # list landing on a freed one's address (which `id()` alone would allow).
         prepared = None
         if _cache is not None:
             cached = _cache.get(_PREPARED_KEY)
-            if cached is not None and cached[0] == len(patterns):
-                prepared = cached[1]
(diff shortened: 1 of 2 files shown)


safishamsi:
Thanks @Alond — the same ignore-evaluation hot path was optimized by #3074 in v0.9.50, which also eliminates the per-file relative_to cost your PR hoisted around. Closing as superseded. Release: https://github.com/Graphify-Labs/graphify/releases/tag/v0.9.50

PR #2872, opened 2026-08-19
fix: bisect extraction chunks on timeout

### LINKED ISSUE #2866: claude-cli: a chunk that hits GRAPHIFY_API_TIMEOUT is not routed into adaptive bisection, so dense chunks fail on the clock instead of being split
## Summary

`_extract_with_adaptive_retry()` (llm.py ~:2043 on `v8`) bisects a chunk on three signals: `finish_reason == "length"`, a context-exceeded error (`_looks_like_context_exceeded`), or a hollow response. A **per-chunk timeout is not one of them**: `_call_claude_cli` uses `subprocess.run(..., timeout=_resolve_api_timeout())`, so a slow chunk raises `subprocess.TimeoutExpired`, the `except Exception` branch re-raises because it isn't a context-exceeded message, and the whole chunk fails with no output. The files are (correctly) left unstamped, but nothing makes them smaller, so on a corpus where the same file packs the same way next run they fail the same way.

For the `claude-cli` backend the timeout *is* the output-size signal in practice: a chunk whose output is heading past ~50-60k tokens takes longer than 900s to stream, so it dies on the clock before it can return `stop_reason: max_tokens` and be bisected.

## Observed (0.9.45, `claude-cli`, `--token-budget 25000`, `GRAPHIFY_API_TIMEOUT=900`)

Full extract of a 1,900-file markdown corpus, 2026-08-17: 28/29 chunks succeeded, one failed with

```
[graphify] chunk 25/29 failed: Command '['claude', '-p', '--output-format', 'json', '--no-session-persistence', '--model', 'claude-sonnet-5', '--json-schema', ...]' timed out after 900.0 seconds
[graphify] WARNING: 1/29 semantic chunk(s) failed — see errors above. Partial results returned.
```

Chunks with entity-dense reference/index docs (many distinct names/paths/keys per KB) emit far more output per input token than narrative docs, which is what pushes a same-sized chunk over the clock. On the same corpus, chunks that returned `finish_reason=length` were bisected and recovered as designed; the timed-out one was not.

## Suggested fix

Treat `subprocess.TimeoutExpired` (and the equivalent SDK timeout exceptions for the API backends) as a bisectable failure in `_extract_with_adaptive_retry`, subject to the same `max_depth` cap — i.e. route it through the same branch as `_looks_like_context_exceeded(exc)`. A timeout on a single unsplittable unit should still fail as it does today. Workaround meanwhile: a smaller `--token-budget` and a larger `GRAPHIFY_API_TIMEOUT` for dense corpora.

hopstreax:
## Summary

Fixes #2866.

`claude-cli` extraction chunks that exceed `GRAPHIFY_API_TIMEOUT` were previously treated as unrecoverable failures. This was especially problematic for dense chunks that could take longer to generate their semantic output and time out before returning `finish_reason="length"`.

This change routes recognized timeout exceptions through the existing adaptive bisection path.

## Root Cause

`_extract_with_adaptive_retry()` already bisects chunks when it encounters:

- `finish_reason == "length"`
- context-window-exceeded errors
- hollow responses

However, `claude-cli` uses `subprocess.run(..., timeout=...)`, which raises `subprocess.TimeoutExpired` when a chunk exceeds the configured timeout.

That exception was not recognized as a bisectable failure, so it propagated and the entire chunk failed without being split.

## Changes

- Add `_looks_like_timeout()` to classify concrete timeout exceptions from supported extraction backends:
  - `subprocess.TimeoutExpired`
  - `openai.APITimeoutError`
  - `anthropic.APITimeoutError`
  - `botocore.exceptions.ReadTimeoutError`
  - `botocore.exceptions.ConnectTimeoutError`
- Route recognized timeout exceptions through the existing adaptive bisection logic.
- Preserve the existing `max_depth` and unsplittable-file safeguards.
- Keep timeout and context-overflow logging semantically distinct.
- Add regression coverage for:
  - timeout classification
  - multi-file chunk bisection
  - `FileSlice` bisection
  - unsplittable single-file timeouts
  - `max_depth` handling
  - unrelated exceptions remaining non-bisectable

## Verification

```text
97 passed, 1 warning  tests/test_llm_backends.py
66 passed, 3 warnings  tests/test_chunking.py tests/test_claude_cli_backend.py
````

##Also verified with:

```text
git diff --check
```

No whitespace errors.

### DIFF
--- a/graphify/llm.py
+++ b/graphify/llm.py
@@ -9,6 +9,7 @@ import hashlib
 import json
 import os
 import re
+import subprocess
 import sys
 import time
 from collections.abc import Callable
@@ -1977,6 +1978,27 @@ def _looks_like_context_exceeded(exc: BaseException) -> bool:
     return any(marker in msg for marker in _CONTEXT_EXCEEDED_MARKERS)
 
 
+def _looks_like_timeout(exc: BaseException) -> bool:
+    """Classify an exception as a recognized subprocess or SDK timeout."""
+    types: list[type[BaseException]] = [subprocess.TimeoutExpired]
+    try:
+        import openai
+        types.append(openai.APITimeoutError)
+    except ImportError:
+        pass
+    try:
+        import anthropic
+        types.append(anthropic.APITimeoutError)
+    except ImportError:
+        pass
+    try:
+        import botocore.exceptions
+        types.extend([botocore.exceptions.ReadTimeoutError, botocore.exceptions.ConnectTimeoutError])
+    except ImportError:
+        pass
+    return isinstance(exc, tuple(types))
+
+
 def _mark_partial(result: dict) -> None:
     """Tag every node/edge/hyperedge in a truncated chunk result with an internal
     ``_partial`` marker.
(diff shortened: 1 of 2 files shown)


safishamsi:
Shipped in v0.9.47 via authorship-preserving cherry-pick. Thanks @hopstreax! Release: https://github.com/Graphify-Labs/graphify/releases/tag/v0.9.47

Issue #1376, opened 2026-06-18
Markdown links not accurately represented as edges


rcollette:
### Use Case

I have a Microsoft Agent Package Manager (APM ) mono-repo (defines packages and a marketplace to distribute those package).

The repo is mostly composed of skill packages that define coding standards.

Within a standards package there are index.md and table-of-contents.md, that clearly should be hub nodes out to all other documents in that package.  Those hub nodes don't seem to be represented at all.

The following is the self analysis.  The graph was generated, queried and root cause analyzed using GPT 5.4.

---

<div style="font-family: "Segoe UI", -apple-system, BlinkMacSystemFont, "Noto Sans", Helvetica, Arial, sans-serif; font-size: 16px; line-height: 1.5; color: #1f2328; word-wrap: break-word;"><p style="margin: 0 0 16px;"><strong style="font-weight: 600;">They’re not connected because those links were mostly never extracted into the graph.</strong> The issue is in semantic extraction of the C# markdown navigation files, not in clustering.</p>
<p style="margin: 0 0 16px;">In the rebuilt graph:</p>

Source file | Expected role | Extracted outgoing links
-- | -- | --
packages/coding-standards-csharp/index.md | connect many C# concepts to many standards docs | 1
packages/coding-standards-csharp/table-of-contents.md | connect the package to its major standards sections | 3
packages/coding-standards-cpp/index.md | same pattern for C++ | 16
packages/coding-standards-cpp/table-of-contents.md | same pattern for C++ | 24

<p style="margin: 0 0 16px;">The C# source files clearly contain the missing links. For example:</p>
<ul style="padding-left: 2em; margin: 0 0 16px;"><li><span style="background-color: #eff1f3; padding: 0.2em 0.4em; border-radius: 6px;"><code style="font-family: ui-monospace, SFMono-Regular, "SF Mono", Menlo, Consolas, "Liberation Mono", monospace; font-size: 85%; color: #1f2328;">packages/coding-standards-csharp/table-of-contents.md:39</code></span> links to <strong style="font-weight: 600;">C# Repository Standards</strong></li>
<li><span style="background-color: #eff1f3; padding: 0.2em 0.4em; border-radius: 6px;"><code style="font-family: ui-monospace, SFMono-Regular, "SF Mono", Menlo, Consolas, "Liberation Mono", monospace; font-size: 85%; color: #1f2328;">...:45</code></span> links to <strong style="font-weight: 600;">C# HTTP Client Standards</strong></li>
[...]

PR #2875, opened 2026-08-19
fix(markdown): resolve wikilinks vault-wide when sibling resolution misses

BaeHyunJae:
## Problem

A `[[wikilink]]` is resolved lexically against the linking file's directory (`_resolve_markdown_link`), but Obsidian-style vaults resolve wikilinks by **vault-global filename lookup**: a note in a subfolder linking `[[hub]]` means the root-level `hub.md`. Today the lexically derived target id matches no node, the #2169 remap pass rightly refuses to canonicalize a nonexistent path, and the builder drops the edge — so every cross-folder vault link is a silently lost reference. That is the same lost-edge failure #1376 fixed for hub docs, reintroduced for the vault layout graphify's own Obsidian support targets.

## Fix

When a wikilink's lexically resolved path does not exist, retry it against a per-scan-root index of linkable documents:

- built lazily on first miss, per scan root, pruned by `detect._SKIP_DIRS` and dot-directories — the same corpus boundary the scanner draws (and Obsidian itself does not index dot-folders); cleared at the start of each `extract()` run beside the other run-scoped caches
- NFC-normalized on both sides, so filesystems that disagree on Unicode normalization (macOS decomposes, others don't) still match
- bare names match by basename, path-qualified targets (`[[folder/name]]`) by full segment suffix; ties break shallowest-first then lexicographically, mirroring Obsidian's root-wins collision rule, so resolution is deterministic regardless of walk order

A hit is a real file, so the existing `target_file` stamp and the #2169 remap canonicalize the edge on full and incremental runs alike — no new remap machinery. A miss stays dangling exactly as before. The fallback is scoped deliberately:

- **wikilinks only** — vault-global lookup is a wikilink convention; inline `[text](path.md)` and reference-style links keep pure relative semantics, where a missing target is an authoring error rather than an alternate resolution order
- **only when the resolved path is missing** — an existing sibling target still shadows the vault-wide match, so all currently-resolving links are untouched
- **only under an active `extract()` run** — a direct `extract_markdown()` call has no scan root and behaves exactly as before

## Tests

Six new tests in `tests/test_languages.py` (tmp_path style, end-to-end through `extract()`): subfolder→root resolution, path-qualified suffix match, root-wins collision, sibling-shadows-fallback, inline-links-unchanged, and NFD-link-to-NFC-file normalization. Full suite passes (the one pre-existing failure, `test_hermes_skill_destination_posix_uses_home`, fails identically on a clean checkout of v8 on Windows and is unrelated).

### DIFF
--- a/graphify/extract.py
+++ b/graphify/extract.py
@@ -46,7 +46,7 @@ from graphify.extractors.fortran import _cpp_preprocess, extract_fortran  # noqa
 from graphify.extractors.go import _GO_PREDECLARED_FUNCS, extract_go  # noqa: F401
 from graphify.extractors.json_config import extract_json  # noqa: F401
 from graphify.extractors.commonlisp import extract_commonlisp  # noqa: F401
-from graphify.extractors.markdown import extract_markdown  # noqa: F401
+from graphify.extractors.markdown import extract_markdown, _MD_LINK_INDEX_CACHE  # noqa: F401
 from graphify.extractors.ocaml import extract_ocaml  # noqa: F401
 from graphify.extractors.pascal_forms import extract_delphi_form, extract_lazarus_form  # noqa: F401
 from graphify.extractors.powershell import extract_powershell, extract_powershell_manifest  # noqa: F401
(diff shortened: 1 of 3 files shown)


safishamsi:
Shipped in v0.9.48 via authorship-preserving cherry-pick. Thanks @BaeHyunJae! Release: https://github.com/Graphify-Labs/graphify/releases/tag/v0.9.48

Issue #2877, opened 2026-08-19
test_label_communities_batches_when_over_batch_size is flaky on v8 HEAD (order-dependent, full-suite only)


AromalBiju1:
## Summary
`tests/test_labeling.py::test_label_communities_batches_when_over_batch_size`
fails intermittently when run as part of the full suite, but passes
consistently when run in isolation. Confirmed on a clean `v8` checkout with
no other changes applied.

## Version
v8 @ b14b52e (v0.9.47), Python 3.14.

## Reproduction

Isolated — passes every time:

uv run pytest tests/test_labeling.py::test_label_communities_batches_when_over_batch_size -q

1 passed (repeated 6/6 runs)

Full suite — fails intermittently:

uv run pytest tests/ -q

FAILED tests/test_labeling.py::test_label_communities_batches_when_over_batch_size
assert [100, 50, 100] == [100, 100, 50]

Reproduced this way twice on a clean tree (no local changes, verified via
`git stash` before each run).

## What the test asserts
The test dispatches 250 communities in batches of 100 and asserts the fake
LLM call receives them in submission order: `[100, 100, 50]`. Under the
full suite it sometimes observes `[100, 50, 100]` instead — same set of
batch sizes, different order.

## Suspected cause
Looks like `label_communities`'s batch dispatch doesn't guarantee
submission order is preserved on return — consistent with something like
`ThreadPoolExecutor` + `as_completed()` (completion order, not submission
order) rather than `executor.map()` (which preserves order). This would
only surface under contention, which lines up with it passing standalone
and failing under full-suite load.

Two possible fixes, not sure which is intended:
- `label_communities` switches to an order-preserving collection method, or
- the test sorts/normalizes `calls` before comparing, since exact dispatch
  order may not actually be a real invariant worth asserting

Haven't dug into `graphify/llm.py`'s batching implementation, so filing as
an observation rather than a proposed fix.

## Why filing this
CONTRIBUTING.md asks contributors to run the full suite and confirm it
passes before opening a PR. Flagging this since it currently blocks that
on a totally clean `v8` checkout, independent of any particular PR's
changes.

GREYGROUPJP:
This looks fixed. `98d62e23` ("test(labeling): pin max_concurrency=1 so batch-order assertion is deterministic", cherry-picked from #3374) shipped in 0.9.56 and makes the batch-order assertion deterministic. The commit does not reference this issue, so it stayed open.

[...]

PR #2878, opened 2026-08-19
fix(#2873): stub-mint undeclared edge endpoints, prune manifest deps, fix merge repos-union clobbering

### LINKED ISSUE #2873: [BUG] graph.json contains edges to undeclared nodes, so loaders materialise attribute-less phantom nodes (and merge-graphs fragments external modules per repo)
### Summary

`graph.json` contains edges whose endpoints are not in the `nodes` array. Every consumer that builds a graph from the file, including networkx, materialises those endpoints as nodes with no attributes, so the node set the file declares and the node set the file produces disagree.

### Version

graphifyy 0.9.46, Python 3.13, `graphify update . --no-cluster`.

### Reproduction

Any Python source that imports from the standard library or an uninstalled package. Running against a small package (8 modules):

```python
import json, networkx as nx

g = json.load(open("graphify-out/graph.json"))
declared = {n["id"] for n in g["nodes"]}
G = nx.node_link_graph(g, edges="links")

print(len(g["nodes"]), G.number_of_nodes())
print(sorted(n for n in G.nodes if n not in declared))
```

### Actual

```
125 131
['dataclasses', 'mcp_server_fastmcp', 'pathlib', 'pyoxigraph', 're', 'typing']
```

Ten `imports` / `imports_from` edges point at module names that were never emitted as nodes. The file says 125 nodes; loading it yields 131.

This surfaces in graphify's own output too:

```
$ graphify explain "typing"
  Type:
  Community:
  Degree:    1
```

`Type` and `Community` are blank because there is no node record to read them from, and `graphify path` will route through these endpoints as though they were ordinary nodes.

### The cross-repo consequence

`merge-graphs` has to paper over this, and the way it does creates a second problem. Merging two repositories that both import `typing` and `pathlib`:

```
inputs: 5 + 6 nodes, 5 + 6 edges
merged: 15 nodes, 11 edges
```

The four extra nodes are the materialised endpoints, namespaced per repo:

```
id=repoA::typing    label=None  type=None
id=repoA::pathlib   label=None  type=None
id=repoB::typing    label=None  type=None
id=repoB::pathlib   label=None  type=None
```

So the same standard-library module becomes four distinct nodes across two repositories, and a merged graph cannot answer "which repos depend on `typing`", which is the question a cross-repo merge exists to answer. Namespacing is right for repo-local ids; these are not repo-local, they are external identifiers that happen to have arrived through a repo-local path.

### What I am not claiming

Not emitting nodes for external modules may well be deliberate, to keep the graph to code you own. If so, the issue is the edges rather than the missing nodes: an edge to a node that does not exist is not a neutral encoding of "this import goes outside the graph", because no consumer can distinguish it from a node the extractor forgot to write.

### Checked and clean, for what it is worth

Two things I expected to find and did not, both worth recording:

- **Builds are deterministic.** Two independent runs over the same tree produced identical node ids in identical order, identical edges in identical order, and identical top-level keys. The local-deterministic-parsing claim holds.
- **`merge-graphs` does not collide repo-local ids.** Two unrelated repositories that both contain `src/app/engine.py` produce the same bare id `src_app_engine`, and the merge correctly namespaces them to `repoA::` / `repoB::` while preserving `local_id` and `repo`. 7 nodes in, 7 nodes out, nothing silently unified.
- **`diagnose multigraph` works.** It found the same 8 same-endpoint edge pairs my own check did, including the `imports_from` / `re_exports` pairs on `__init__.py`.

AromalBiju1:
Fixes #2873

## Summary

Two related problems from #2873, plus one bug found while fixing them.

### 1. Undeclared edge endpoints (the core report)
`graph.json` could contain edges pointing at node ids that were never
emitted — stdlib/third-party imports being the main case. Every consumer
that loads the file (networkx included) then materializes those endpoints
as attribute-less phantom nodes, so the declared node set and the produced
node set disagreed.

**Fix:** after every resolution/repoint/disambiguation pass runs, mint an
explicit, minimal stub node (`external: True`) for any edge endpoint still
not in the node set. This makes the "outside the graph" case *explicit*
instead of an implicit, per-consumer materialization.

Manifest/package-dependency edges (`depends_on`/`requires` relations, or
`pkg_*` ids) are excluded from stub-minting on purpose — the issue itself
noted that not emitting nodes for external packages may be deliberate, and
existing tests (`test_external_dependency_edge_pruned_not_orphaned`)
already encode prune-not-fabricate as the intended behavior for that case.
Kept it that way.

### 2. merge-graphs fragmenting external modules per repo
Since external stub nodes aren't repo-local, `merge-graphs` leaves them
unprefixed so `typing` unifies across repos instead of becoming
`repoA::typing` / `repoB::typing`.

While testing this, found a second bug: `nx.compose` merges shared-node
attributes with `dict.update`, so on a 3+ repo merge, each pass's `repos`
list on a shared external node overwrote the previous one instead of
unioning — only the last repo survived. `cli.py` now accumulates the
`repos` union itself and reattaches it after composing. Added
`test_merge_graphs_external_node_repos_union_not_clobbered` for this.

### 3. Stub-minting interaction with incremental rebuilds (found during testing)
Applying the stub-minting fix broke 4 previously-passing incremental
rebuild tests (cross-file `calls` edges to unchanged files silently
vanishing — surfacing as #2406/#2437/#2438 symptoms). Root cause: the
undeclared-endpoint check only looked at `all_nodes` (this run's freshly
extracted nodes), not at `resolution_context_nodes` — the read-only
persisted nodes from *unchanged* files that #2406's cross-file resolver
uses. So a real call target living in an unchanged file looked
"undeclared" during an incremental rebuild, got a stub minted over its
real id, and the real preserved node was reconciled away in favor of the
fabricated stub.

**Fix:** union `resolution_context_nodes` ids into the declared-id set
before deciding what's undeclared.

## Files changed
- `graphify/extract.py` — stub-minting, manifest-relation carve-out,
  `resolution_context_nodes` fix
- `graphify/build.py` — external items excluded from carry-forward-drop
  logic in the semantic/LLM merge path (defensive, same principle as the
  incremental fix above)
- `graphify/cli.py` — `repos` union fix for merge-graphs
- `tests/test_csharp_type_resolution.py` — assertions updated now that
  previously-`None` import targets resolve to real stub types
- `tests/test_merge_graphs_cli.py` — new coverage for the repos-union fix

## Test results
uv run pytest tests/ -q
4783 passed, 11 skipped

One unrelated, pre-existing flake:
`tests/test_labeling.py::test_label_communities_batches_when_over_batch_size`
fails intermittently in full-suite runs, passes 6/6 in isolation.
Reproduces identically on a clean `v8` HEAD (`b14b52e`, no changes applied)
— order-dependent, unrelated to anything touched here.  Filed separately as
#2877


safishamsi:
Thanks @AromalBiju1 — your diagnosis of the undeclared-endpoint problem was correct and the fix direction (stub-mint external endpoints + stop merge from fragmenting them per repo) is exactly what shipped. We landed a focused version in v0.9.63 that mints typed external stub nodes for import-family edges (keeping the #3156 call-target suppression intact) and leaves external ids unprefixed so a cross-repo merge unifies them. Closing as superseded; credited in the changelog. Appreciate the detailed report on #2873.

Issue #1224, opened 2026-06-10
AST pass explodes data .json into orphan key-nodes (CODE_EXTENSIONS includes .json) — 561 isolated nodes on a real repo


dawtips:
## Summary
`detect.CODE_EXTENSIONS` includes `.json` (detect.py, the big extension set), so the AST/structural pass treats every JSON file as code and emits **every JSON key as its own node**. On repos where `.json` is data — eval fixtures, config, datasets, manifests — this floods the graph with orphan key-nodes and near-duplicate communities that swamp the real structure.

## Observed (graphifyy 0.8.36, Windows, real repo)
A Python CLI repo (~200 files) with ~40 data `.json` files (eval `cases.json`/`eval.json` pairs, two plugin manifests, one test fixture):

| | with .json as code | .json excluded |
|---|---|---|
| nodes | 1819 | 1170 |
| communities | 114 | 73 |
| isolated nodes | **561** | 2 |

- One repeated key (`prompt_inputs_spec`) appeared as **214 occurrences in the AST extraction** (0 in the LLM semantic pass).
- ~30 near-identical communities formed, one per eval suite, each holding just that file's keys (`generation`, `target`, `cases_file`, `num_cases`, …).
- The report's "Knowledge Gaps" then flags 561 isolated nodes / 735 weakly-connected — reads like a codebase problem but is purely extraction noise.

## Workaround
`.graphifyignore` with `*.json` (works — both `detect()` and the post-commit hook honor it). But it costs the LLM pass's concept extraction from those files too, and `.graphifyignore` shadowing `.gitignore` per-directory means re-listing everything else.

## Suggested fixes (any of)
1. Treat data-shaped `.json` as `document` rather than `code` — or only AST-extract manifest-style json the resolver actually understands (`package.json`, `tsconfig.json`, lockfiles are already skipped).
2. A config knob to reclassify extensions per-repo (e.g. `json: document` / `json: skip`).
3. Cheapest guard: warn in the report when a single `.json` yields N+ key-nodes with no cross-file edges — silent flooding is the painful part.

Happy to provide the full before/after graph.json pair if useful.

safishamsi:
[...]

PR #2885, opened 2026-08-20
fix(extract): don't classify deliberately-declined data JSON as failed (#2879)

### LINKED ISSUE #2879: Deliberately-skipped data JSON (#1224) is classified as failed extraction, so it is re-queued on every incremental run
### Summary

`extract.py`'s empty-result classifier cannot distinguish *"the extractor deliberately declined this file"* from *"extraction failed"*. Data-shaped JSON is intentionally not AST-walked (#1224 — it is left to the LLM semantic pass), but because a `.json` extractor **is** registered, every such file trips the `failed_sources` path, never gets frozen as processed in the incremental manifest, and is therefore re-queued on **every** subsequent run, forever.

Correctness is unaffected. The cost is unbounded repeated work on incremental runs.

### The mechanism

`extract.py` classifies a file as failed when it produced no nodes *and* an extractor exists for it:

```python
if (not _res.get("nodes")) and _get_extractor(_p) is not None:
    if _key not in _failed_seen:
        _failed_sources.append(_key)
        _failed_seen.add(_key)
```

`failed_sources` is then returned with the explicit purpose (per the comment in the same file) of keeping those paths out of the processed set:

```python
# Surfaces failed/empty AST sources to the CLI so the incremental
# manifest does not freeze them as processed (#2543).
"failed_sources": _failed_sources,
```

For data JSON both conditions are true by design: `_get_extractor()` returns the JSON extractor because the extension is registered, and the extractor correctly returns no nodes because the file is data rather than a config/manifest. Zero nodes here is the intended outcome, not a failure — but it is indistinguishable from one at this call site.

### Reproduction

Any repository containing data-shaped JSON alongside config JSON. On a 3,329-file repo indexed with 0.9.47:

```
manifest entries:                                3329
.json with BLANK ast_hash (re-queued every run):  110
.json with populated ast_hash:                    178
blank ast_hash, all extensions:                   110
blank by extension: [('.json', 110)]
```

Every entry with a blank `ast_hash` is a `.json` file; no other extension is affected, which matches the classifier path above rather than any general extraction problem. Examples are Fumadocs-style `meta.json` sidecars:

```
apps/site/content/docs/agentic-dev/meta.json
apps/site/content/docs/ai-production/meta.json
apps/site/content/docs/base/meta.json
```

These are stable files that have not changed in many runs, yet are re-processed on each incremental pass. With a 15-minute indexing cadence this repeats indefinitely.

### Suggested direction

The extractor needs a way to report *"declined by design"* separately from *"produced nothing"*, so the classifier can exempt the former. Concretely, either:

1. have the JSON dispatch return an explicit marker (e.g. `{"nodes": [], "edges": [], "declined": True}`) and skip `failed_sources` when it is set; or
2. have `_get_extractor()` return `None` for JSON files the structural extractor has already determined it will not walk, so the existing `is not None` guard does the right thing.

(1) is probably less invasive since `_get_extractor()` is dispatch-by-extension and does not currently inspect content.

I am happy to open a PR if a preferred shape is indicated.

### Environment

- graphifyy 0.9.47 (PyPI), Python 3.14, macOS 15 arm64
- Incremental runs via `graphify extract <src> --backend openrouter --out <dir>`

rajarshidattapy:
Fixes #2879.

## The bug

`extract()`'s empty-result classifier could not tell *"the extractor declined this file on purpose"* from *"extraction failed"*:

```python
if (not _res.get("nodes")) and _get_extractor(_p) is not None:
    _failed_sources.append(_key)
```

Data-shaped JSON is deliberately not AST-walked (#1224 — it is left to the LLM semantic pass), but a `.json` extractor **is** registered, so a declined file satisfies both halves of that test. It landed in `failed_sources`, whose stated purpose is to keep those paths out of the processed set, so the incremental manifest never stamped it — and it was re-queued on **every** subsequent run, forever. On the reporter's 3,329-file repo that was 110 files with a blank `ast_hash`, all of them `.json`.

## The fix

`extract_json()` already reports the decline explicitly:

```python
return {"nodes": [], "edges": [], "skipped": "data json (not a config/manifest)"}
```

That marker just had no consumer. This honours it at the two call sites that needed it:

- `failed_sources` (#2543) skips results carrying `skipped`, so the CLI stamps them as processed;
- the #1666 "produced zero nodes" warning skips them too, since a deliberate decline is not an anomaly worth warning about.

This is option (1) from the issue's *Suggested direction*, and needs no new marker — `_get_extractor()` stays pure dispatch-by-extension.

An empty result **without** the marker is still classified as failed, so the #1666/#2543 safety net is unchanged for genuine failures.

## Tests

- `test_extract_declined_data_json_is_not_failed` — a data `meta.json` yields empty `failed_sources` and no zero-nodes warning, while a sibling `package.json` still extracts. Fails without the change.
- `test_extract_genuinely_empty_json_still_failed` — an extractor returning an empty result with no marker is still reported failed.

### DIFF
--- a/graphify/extract.py
+++ b/graphify/extract.py
@@ -5512,7 +5512,7 @@ def extract(
     _empty_sources: list[str] = []
     for i, _p in enumerate(paths):
         _res = per_file[i] or {}
-        if _res.get("nodes") or _res.get("error"):
+        if _res.get("nodes") or _res.get("error") or _res.get("skipped"):
             continue
         if _get_extractor(_p) is not None:
             _empty_sources.append(str(_p))
@@ -5543,6 +5543,12 @@ def extract(
             if _key not in _failed_seen:
                 _failed_sources.append(_key)
                 _failed_seen.add(_key)
+            continue
+        if _res.get("skipped"):
+            # The extractor declined this file by design (data JSON, #1224), so
+            # zero nodes is the intended outcome rather than a failure. Marking
+            # it failed keeps it out of the incremental manifest and re-queues
+            # it on every subsequent run, forever (#2879).
             continue
         if (not _res.get("nodes")) and _get_extractor(_p) is not None:
             if _key not in _failed_seen:
--- a/tests/test_extract.py
+++ b/tests/test_extract.py
@@ -3966,3 +3966,42 @@ def test_inferred_uses_edge_dropped_for_module_top_level_reference(tmp_path):
     uses = _inferred_uses(result)
 
     assert not any(tgt == "helpers_helper" for _, tgt in uses)
+
+
+def test_extract_declined_data_json_is_not_failed(tmp_path, capsys):
+    """#2879: data JSON is declined by design (#1224), not failed.
+
+    A `.json` extractor is registered, so a declined file used to satisfy both
+    halves of the failed-source test (zero nodes + extractor exists) and was
+    re-queued on every incremental run because the CLI never stamped it as
+    processed.
+    """
+    pytest.importorskip("tree_sitter_json")
+    data = tmp_path / "meta.json"
+    data.write_text('{"pages": ["a", "b"], "title": "Docs"}\n')
+    cfg = tmp_path / "package.json"
+    cfg.write_text('{"dependencies": {"left-pad": "^1.0.0"}}\n')
+
+    result = extract([data, cfg], cache_root=tmp_path)
+    err = capsys.readouterr().err
+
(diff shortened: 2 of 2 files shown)


safishamsi:
Shipped in v0.9.48 via authorship-preserving cherry-pick so you keep contributor-graph credit. Thanks @rajarshidattapy! Release: https://github.com/Graphify-Labs/graphify/releases/tag/v0.9.48

PR #2886, opened 2026-08-20
fix(cpp): keep symbols for nested C++ types and C++/CLI sources (#2876)

### LINKED ISSUE #2876: C++ extraction: C++/CLI files and nested C++ types lose their symbols
# C++ extraction: C++/CLI files and nested C++ types lose their symbols

graphify 0.9.47 · Python 3.12.10 · tree-sitter 0.25.2 · tree-sitter-cpp 0.23.4 · Windows

Two independent gaps in the C++ extractor, both hit by a .NET interop wrapper (a 141-method C++/CLI
class extracted 12 symbols). Happy to split into two issues.

Reproducer for both — `_get_extractor(p)(p)`, labels shown verbatim:

### 1. C++/CLI is unparseable, and the failure takes the whole class body with it

```cpp
// cli.h
namespace N {
    public ref class Wrapper
    {
    public:
        static void Init() { }
        static System::String^ Name() { return gcnew System::String(""); }
    };
}
```

```
parse_errors={'first_error_line': 2, 'multiline_error': False}
labels=['cli.h', 'Wrapper()', 'public']
```

tree-sitter-cpp implements none of `ref class`, `Type^`, `Type%`, `gcnew`, `[assembly:...]`. Because the
ERROR lands on the type header, everything inside dissolves and recovery invents `Wrapper()` / `public`.
Expected: `Wrapper`, `.Init()`, `.Name()`.

### 2. Nested C++ types are dropped — silently, in plain C++

```cpp
// nested.h
namespace N {
    class Outer
    {
    public:
        class Inner
        {
        public:
            static void Method() { }
        };
    };
}
```

```
parse_errors=None
labels=['nested.h', 'Outer']
```

No parse error and no warning — `Inner` and everything in it are simply gone. Cause: a nested type is a
`field_declaration` whose `type` field *is* the `class_specifier`, and the C++ member-variable branch in
`extractors/engine.py` (`config.ts_module == "tree_sitter_cpp" and t == "field_declaration"`) consumes the
node and returns before the walk can descend. The engine already supports nested classes for other
languages (`is_nested_type`, the `contains` edge).

### Fix I'm running locally

Both patched against 0.9.47, giving `['cli.h', 'Wrapper', '.Init()', '.Name()', 'String']` and
`['nested.h', 'Outer', 'Inner', '.Method()']`:

1. A `source_override` for `extract_cpp` that normalizes C++/CLI to the nearest standard C++ spelling.
   It is **byte-length preserving** (removed tokens overwritten with spaces, never deleted), so every
   offset, line and column still points at the same place in the file on disk, and it only engages on
   files carrying a CLI marker — plain C/C++/CUDA parses byte-for-byte as before.
2. In that `field_declaration` branch, descend into a `class_specifier` type field with `parent_class_nid`
   threaded through, returning unless a member is declared alongside it (`class Inner { } inst;`).

On a 1475-file repo this took GmshCommon.h from 13 nodes to 116 (107 methods, 8 nested classes) with no
change elsewhere. Glad to open a PR if the approach looks right.

### Unrelated, one line

`--force` doesn't invalidate `graphify-out/cache/ast/<version>`, so any extractor change is masked until
that directory is deleted by hand — worth a flag or a note.

rajarshidattapy:
Fixes #2876.

Two independent gaps in the C++ extractor, both reproduced verbatim from the issue.

## 1. Nested C++ types were dropped silently

A nested type is a `field_declaration` whose `type` field **is** the `class_specifier`. The C++ member-variable branch in `extractors/engine.py` consumed that node and returned before the walk could descend, so the nested type and everything it declares vanished — no parse error, no warning.

```cpp
namespace N { class Outer { public: class Inner { public: static void Method() { } }; }; }
```

| | labels |
|---|---|
| before | `['nested.h', 'Outer']` |
| after | `['nested.h', 'Outer', 'Inner', '.Method()']` |

The fix walks the `class_specifier` with `parent_class_nid` threaded through, which routes it into the engine's existing nested-type handling — `Inner` gets a `contains` edge from `Outer` rather than from the file, exactly as for the other languages. The declarator loop still runs afterwards, so `class Inner { } inst;` yields both the type and the member.

## 2. C++/CLI took the whole class body down with it

tree-sitter-cpp implements none of `ref class`, `Type^`, `Type%`, `gcnew` or `[assembly:…]`. Because the ERROR lands on the type header, everything inside dissolves and recovery invents symbols that were never written:

```cpp
namespace N {
    public ref class Wrapper {
    public:
        static void Init() { }
        static System::String^ Name() { return gcnew System::String(""); }
    };
}
```

| | labels |
|---|---|
| before | `['cli.h', 'Wrapper()', 'public']` — plus `parse_errors` on line 2 |
| after | `['cli.h', 'Wrapper', '.Init()', '.Name()', 'String']`, no parse errors |

`_normalize_cpp_cli()` maps each C++/CLI spelling onto the nearest standard C++ one before parsing, via `_extract_generic`'s existing `source_override` hook (the same one Vue SFCs use). Three properties keep it safe:

- **Byte-length preserving.** Dropped tokens are overwritten with spaces, never deleted, and `gcnew` becomes `new` plus padding. Every offset, line and column still points at the same place in the file on disk, so reported source locations stay accurate.
- **Gated on a C++/CLI marker** (`ref`/`value class`, `interface class`, `gcnew`, `[assembly:`). Plain C/C++/CUDA is parsed byte-for-byte as before — `^`/`%` alone are not markers, since they are the ordinary XOR and modulo operators.
- **`^`/`%` are rewritten only in their attached suffix spelling** (`String^ s`, `int% n`), so a spaced `a ^ b` / `a % b` in the same file keeps its meaning. Covered by a test.

## Tests

`tests/test_cpp_nested_and_cli.py` — both issue reproducers, the `class Inner { } inst;` case, `Outer contains Inner`, byte-offset preservation (newline positions unchanged), plain C++ left untouched, and operators surviving inside a CLI file. The existing 82 C++/ObjC tests pass unchanged.

## Not included

The issue's closing "unrelated, one line" note about `--force` not invalidating `graphify-out/cache/ast/<version>`. Worth noting for anyone hitting it: that directory is namespaced by graphify version (`cache_dir()` → `v{_EXTRACTOR_VERSION}-s{_AST_CACHE_SCHEMA}`), so a released extractor change invalidates it automatically. The stale-cache problem is specific to running a locally patched extractor at an unchanged version. Happy to add a `--force` AST-cache bypass in a separate PR if that is wanted.

### DIFF
--- a/graphify/extract.py
+++ b/graphify/extract.py
@@ -1999,11 +1999,15 @@ _CPP_CLI_SUFFIX_UNAMBIGUOUS_RE = re.compile(rb"(?<=[A-Za-z0-9_>])[\^%](?=\s*[,)\
 #    `List<int>^ items`, `DataTable^ t`, `int% n`. Qualified names, a closing
 #    generic bracket, .NET's PascalCase convention and the primitive value
 #    types are all type positions; requiring one keeps lowercase operands like
-#    `count% 2` and `hash^ mask` as arithmetic. The type is captured and
-#    re-emitted so the substitution stays byte-length preserving.
+#    `count% 2` and `hash^ mask` as arithmetic. A capitalized name must also
+#    carry a lowercase letter, so SCREAMING_CASE constants stay arithmetic
+#    too (`MASK^ value`); a lone capital is exempt for generic parameters
+#    (`T^ x`). The type is captured and re-emitted so the substitution stays
+#    byte-length preserving.
 _CPP_CLI_SUFFIX_DECL_RE = re.compile(
     rb"(\b[A-Za-z_][A-Za-z0-9_]*(?:::[A-Za-z_][A-Za-z0-9_]*)+"   # System::String
-    rb"|\b[A-Z][A-Za-z0-9_]*"                                    # String, DataTable, T
+    rb"|\b[A-Z][A-Za-z0-9_]*[a-z][A-Za-z0-9_]*"                   # String, DataTable
+    rb"|\b[A-Z](?=[\^%])"                                        # T
     rb"|\b(?:bool|char|wchar_t|short|int|long|float|double|unsigned|signed)"
     rb"|>)"                                                      # List<int>^
     rb"[\^%](?=\s+[A-Za-z_])"
--- a/graphify/extract.py
+++ b/graphify/extract.py
@@ -1958,13 +1958,71 @@ def _augment_cpp_string_tests(path: Path, result: dict) -> dict:
     return result
 
 
+# ── C++/CLI normalization (#2876) ────────────────────────────────────────────
+# tree-sitter-cpp implements none of `ref class`, `Type^`, `Type%`, `gcnew` or
+# `[assembly:…]`. The ERROR lands on the type header, which dissolves the whole
+# class body — a 141-method .NET interop wrapper yielded 12 junk symbols. These
+# rewrites map each spelling onto the nearest standard C++ one.
+
+# Only files carrying one of these engage the rewrite, so plain C/C++/CUDA is
+# parsed byte-for-byte as before. `^`/`%` alone are not markers: they are the
+# ordinary XOR and modulo operators.
+_CPP_CLI_MARKER_RE = re.compile(
+    rb"\b(?:ref|value)\s+(?:class|struct)\b"
+    rb"|\binterface\s+class\b"
+    rb"|\bgcnew\b"
+    rb"|\[\s*(?:assembly|module)\s*:"
+)
+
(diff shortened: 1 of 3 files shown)


safishamsi:
Shipped in v0.9.48 via authorship-preserving cherry-pick, with a note recording nested enums as a known out-of-scope gap. Thanks @rajarshidattapy! Release: https://github.com/Graphify-Labs/graphify/releases/tag/v0.9.48

PR #2887, opened 2026-08-20
fix(llm): stop bisecting hollow responses, retry the same chunk instead (#2880)

### LINKED ISSUE #2880: Hollow responses are relabelled as truncation and bisected, so one bad response costs up to 15 billed calls (measured 18x token blow-up)
## Summary

`_response_is_hollow()` (llm.py) collapses two very different failures into one, and
`_call_*` then relabels the result as `finish_reason="length"` so adaptive retry bisects
the chunk:

- **truncated** — the model ran out of `max_completion_tokens` mid-JSON. Bisecting is the
  correct recovery: smaller input ⇒ shorter output ⇒ it fits.
- **hollow** — HTTP 200 with empty/null/whitespace content, or content that parses to zero
  nodes and zero edges (a rate limit, a transport hiccup, a refusal, an agentic prose reply,
  a reasoning-first content block).

For the hollow case bisecting cannot converge: both halves are produced by the same
misbehaving backend and come back hollow too. Every hollow response therefore costs
`2**max_retry_depth` billed calls instead of one — with `max_retry_depth=3`, up to **15
calls per chunk**, all of which fail.

Five call sites do the relabel: `llm.py` ~1292 (`_call_openai_compat`), 1343 (`_call_claude`),
1606 (`_call_claude_cli`), 1669 (`_call_azure`), 1726 (`_call_bedrock`).

## Evidence

Measured on a scheduled nightly run against two corpora (graphify 0.9.4x, `claude-cli`):

| run | input tokens | output tokens | result |
|---|---|---|---|
| 2026-08-16 (healthy) | 144,540 | 10,669 | completed, ~82 min |
| 2026-08-17 (storm)   | 2,585,365 | 234,277 | same corpus, same graph size (920 nodes / 854 edges) |

**~18x the input tokens for an identical graph** — very close to the 15x worst case the
depth-3 fan-out predicts. The log is dominated by pairs of:

```
[graphify] chunk of 15 truncated at depth 0, splitting into halves of 7 and 8
```

The second corpus that night ran **6h44m** and was still on chunk 149/314 when the machine
was rebooted — it never emitted a `tokens:` line, so the spend was invisible until the log
was read by hand.

The codebase already knows this is the failure mode; the comments say so:

- `llm.py` ~1510 — "…reads as truncation, so the adaptive-retry path bisects the chunk
  indefinitely, never converging and never writing graph.json"
- `llm.py` ~1557 — "…reads as truncation, and gets bisected without ever converging (#2076)"
- `_bedrock_response_text` ~1075 — "…gets reclassified as truncation, and sends the chunk
  into bisection that cannot converge"

#2076, #2287 and #2866 each fixed one backend-specific *source* of hollow responses. The
conflation itself is still there, so the next new source costs 15x again.

## Asks

1. **Do not route hollow into the bisect path.** Keep `_response_is_hollow` as a detector,
   but give it its own recovery: retry the *same* chunk with backoff (2-3 attempts), then
   fail the chunk loudly. Bisect only on a real `finish_reason == "length"` or a genuine
   context-window rejection. Bisecting is a size remedy; a hollow response is not a size
   problem.
2. **A budget, not just a depth.** `max_retry_depth` bounds the tree's depth, but nothing
   bounds total sub-calls, and there is no run-level ceiling. Add a per-chunk sub-call cap
   and an optional global call/token ceiling that aborts the run.
3. **Expose the knob.** `max_retry_depth` is a Python-API kwarg only — there is no CLI flag
   and no environment variable, so a `graphify extract` operator cannot lower it or set it
   to 0 as a mitigation. An env var (e.g. `GRAPHIFY_MAX_RETRY_DEPTH`) would be enough.
4. **Emit `tokens:` on abnormal termination too.** The counters are accumulated in memory
   and printed only on the normal write path, so a run that is killed, crashes, or is
   interrupted leaves no cost record at all. A `finally`/atexit emission (or a periodic
   running total) would make a storm visible without reading the whole log.

Asks 1-3 are the fix; 4 is what makes the next occurrence detectable.

chintondutta:
## Summary

Fixes #2880.

`_response_is_hollow()` collapses two unrelated failure modes into one `finish_reason="length"` relabel:

- **truncated** — the model ran out of `max_completion_tokens` mid-output. Bisecting is the correct recovery: smaller input → shorter output → fits.
- **hollow** — HTTP 200 with empty/null content, or content that parses to zero nodes/edges (a rate limit, a transport hiccup, a refusal, a reasoning-only reply).

For the hollow case, bisecting cannot converge: both halves come from the same misbehaving backend and come back hollow too. Every hollow response costs `2**max_retry_depth` billed calls instead of one — with the default `max_retry_depth=3`, up to **15 calls per chunk**, all of which fail.

## Changes

- All 5 call sites (`_call_openai_compat`, `_call_claude`, `_call_claude_cli`, `_call_azure`, `_call_bedrock`) now tag a hollow response as `finish_reason="hollow"` instead of overloading `"length"`.
- `_extract_with_adaptive_retry` gives `"hollow"` its own bounded recovery: up to `_HOLLOW_RETRY_ATTEMPTS` (3) retries on the **same, unsplit** chunk with backoff, before giving up on the chunk (marked partial, same "don't cache as authoritative, re-dispatch next run" handling the truncation give-up path already uses). Bisection is never reached for a hollow response.
- `max_retry_depth` is now also resolvable from a new `GRAPHIFY_MAX_RETRY_DEPTH` env var (previously a Python-API kwarg only, so a `graphify extract` operator had no way to lower it as a field mitigation without a code change). An explicit `max_retry_depth=` kwarg still wins over the env var. Documented in the README's env var table.
- Updated docstrings/comments that described the old "hollow gets relabelled to truncation and bisected" behavior as intended, since after this change it isn't.

## Scope

This covers asks 1 and 3 from #2880. Asks 2 (a per-chunk/global call budget) and 4 (emitting the `tokens:` cost line on abnormal termination) touch `cli.py`'s run loop and signal handling and are a bigger, separate change — planning to follow up with those in a second PR once this one lands, so they don't block the core fix.

## Test plan

- Rewrote `test_adaptive_retry_bisects_on_hollow_ollama_response` → `test_adaptive_retry_retries_same_chunk_on_hollow_response`, asserting the chunk is retried at its original size and never split.
- Added `test_adaptive_retry_gives_up_on_persistently_hollow_chunk` (never bisects, gives up after `_HOLLOW_RETRY_ATTEMPTS`, result marked partial).
- Added `test_adaptive_retry_backs_off_between_hollow_retries` (guards against hot-looping).
- Updated the 3 `_call_openai_compat` hollow-relabel tests to assert `finish_reason == "hollow"`.
- Added `test_resolve_max_retry_depth_default_and_env` and `test_extract_corpus_parallel_explicit_max_retry_depth_wins_over_env`.
- `uv run --frozen pytest tests/ -q --tb=short` — full suite green (4787 passed, 11 skipped, pre-existing skips unrelated to this change).
- `uv run ruff check graphify tests` — clean.
- `uv run pyright graphify/llm.py` — same 11 pre-existing errors as on `v8` before this change (confirmed via `git stash`), none introduced by this PR.

### DIFF
--- a/README.md
+++ b/README.md
@@ -524,6 +524,7 @@ These are only needed for **headless / CI extraction** (`graphify extract`). Whe
 | `GRAPHIFY_MAX_OUTPUT_TOKENS` | Raise output cap for dense corpora | optional — e.g. `32768` for large files |
 | `GRAPHIFY_API_TIMEOUT` | Per-call timeout in seconds for HTTP, claude-cli, Anthropic SDK, and Bedrock backends (default: 600) | optional — also `--api-timeout` flag |
 | `GRAPHIFY_MAX_RETRIES` | How many times to retry a rate-limited (429) request before giving up (default: 6; honors `Retry-After`) | optional — raise for strict per-org limits (e.g. kimi); `0` disables |
+| `GRAPHIFY_MAX_RETRY_DEPTH` | How many levels deep a chunk can be bisected on truncation/context-overflow before adaptive retry gives up on it (default: 3 → max 8x expansion of one chunk) | optional — lower to cap worst-case cost; `0` disables bisection. Does not affect hollow-response retries, which are always bounded and never bisected (#2880) |
 | `GRAPHIFY_FORCE` | Force graph rebuild even with fewer nodes | optional — also `--force` flag |
 | `GRAPHIFY_GOOGLE_WORKSPACE` | Auto-enable Google Workspace export | optional — set to `1` |
 | `GRAPHIFY_TRIAGE_BACKEND` | Backend for `graphify prs --triage` | optional — auto-detected from available keys |
--- a/graphify/llm.py
+++ b/graphify/llm.py
@@ -433,6 +433,24 @@ def _resolve_max_retries(default: int = 6) -> int:
     return default
 
 
+def _resolve_max_retry_depth(default: int = 3) -> int:
+    """How many levels deep a chunk can be bisected on truncation/context-overflow
+    before adaptive retry gives up on it (default 3 → max 8x expansion of one
+    chunk). Exposed as an env var because it was previously a Python-API kwarg
+    only, so a `graphify extract` operator had no way to lower it (or disable
+    bisection with 0) as a field mitigation without a code change (#2880).
+    Honour GRAPHIFY_MAX_RETRY_DEPTH; 0 is allowed (disable bisection)."""
+    raw = os.environ.get("GRAPHIFY_MAX_RETRY_DEPTH", "").strip()
+    if raw:
+        try:
+            v = int(raw)
+            if v >= 0:
+                return v
+        except ValueError:
+            pass
+    return default
+
+
 def _thinking_disabled_via_env() -> bool:
     """Opt-in (GRAPHIFY_DISABLE_THINKING) to send ``{"thinking": {"type": "disabled"}}``
     to reasoning-capable OpenAI-compatible models such as ``deepseek-v4-flash``.
@@ -1093,10 +1111,11 @@ def _bedrock_response_text(resp: dict, default: str = "") -> str:
     API does not promise a text block is first: reasoning-capable models emit a
     ``reasoningContent`` block ahead of the answer, and ``toolUse`` or future
     block types can precede it too. Indexing position 0 therefore yields no text
-    at all for those models, which reads downstream as a hollow response, gets
-    reclassified as truncation, and sends the chunk into bisection that cannot
-    converge. Select on the block's shape instead of its position so this holds
-    for any model; a response whose first block is already text is unaffected.
+    at all for those models, which reads downstream as a hollow response and
+    gets retried on the same chunk (not bisected — bisection cannot recover a
+    shape problem like this). Select on the block's shape instead of its
+    position so this holds for any model; a response whose first block is
+    already text is unaffected.
     """
     content = resp.get("output", {}).get("message", {}).get("content", [])
     if not isinstance(content, list):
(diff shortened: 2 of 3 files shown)


chintondutta:
Looked into both flagged findings, neither holds up on inspection:

**`llm.py:1407` "Claude backend returns non-OpenAI finish_reason value"**
`"hollow"` is intentional, not a leak. It's documented in the comment right above as a graphify-internal sentinel applied uniformly across all 5 backends. It never escapes this layer: the terminal hollow-giveup path (`llm.py:2289`) explicitly normalizes back to `"stop"` before returning. No caller ever observes `"hollow"`.

**`llm.py:436` "GRAPHIFY_MAX_RETRY_DEPTH=0 disables bisection but nothing wires the hollow-retry attempts"**
Traced `_extract_with_adaptive_retry` (`llm.py:2123`): the hollow-retry `while` loop (`llm.py:2264`) runs unconditionally on the happy path, with no dependency on `_depth`/`max_depth`. Bisection is gated by `_depth >= max_depth` (lines 2227, 2319), but only on the truncation (`"length"`) path, which is separate from the hollow path. So `GRAPHIFY_MAX_RETRY_DEPTH=0` correctly disables bisection while leaving the 3-attempt hollow retry (`_HOLLOW_RETRY_ATTEMPTS=3`, backoff 2s/4s) fully wired.

Both look like false positives from the unverified/consensus-only findings. Treating this as clear to merge on these two points.

safishamsi:
Thanks for tackling this @chintondutta - your diagnosis was right. We landed a fix for #2880 in v0.9.48 via #2888, which took the same approach (a distinct hollow finish-reason routed through a bounded same-chunk retry) and additionally kept the retry inside the timeout-recovery guard and honored GRAPHIFY_MAX_RETRY_DEPTH=0 for the hollow path too. Closing this as superseded, but the contribution and the framing were genuinely helpful. Release: https://github.com/Graphify-Labs/graphify/releases/tag/v0.9.48

PR #2888, opened 2026-08-20
fix(llm): retry hollow responses instead of bisecting them (#2880)

### LINKED ISSUE #2880: Hollow responses are relabelled as truncation and bisected, so one bad response costs up to 15 billed calls (measured 18x token blow-up)

rajarshidattapy:
Fixes #2880 (asks 1 and 3; see *Not included* for 2 and 4).

## The conflation

`_response_is_hollow()` detects two different failures, and all five `_call_*` sites relabelled both as `finish_reason="length"`:

- **truncated** — the model ran out of `max_completion_tokens` mid-JSON. Bisecting is the correct recovery: smaller input ⇒ shorter output ⇒ it fits.
- **hollow** — HTTP 200 with empty/null/whitespace content, or content that parses to zero nodes and zero edges (a rate limit, a transport hiccup, a refusal, an agentic prose reply, a reasoning-first content block).

Bisecting a hollow response cannot converge — both halves are produced by the same misbehaving backend and come back hollow too — so every hollow response cost `2**max_retry_depth` billed calls instead of one. At the default depth that is up to **15 calls per chunk, all failing**, which matches the reporter's measured ~18x token blow-up for an identical 920-node/854-edge graph.

## Ask 1 — hollow gets its own recovery

`_mark_hollow()` replaces the five copies of the relabel and sets `finish_reason = "hollow"`, distinct from `"length"`. `_extract_with_adaptive_retry` then:

1. retries the **same** chunk with backoff (2s, 8s — bounded, so at most 3 calls per chunk);
2. if it is still hollow, prints a loud failure naming the likely causes, marks the chunk's files partial so the next run re-dispatches them and they are not promoted to the semantic cache as authoritative, and returns.

Bisection still handles genuine `finish_reason == "length"` and context-window rejections — those *are* size problems, and that path is unchanged (covered by `test_adaptive_retry_bisects_on_truncated_response`).

Worst case per chunk: **15 calls → 3**. A transient hollow that clears still recovers, in 2 calls rather than 3+.

The three stale comments the issue quoted (`llm.py` ~1510, ~1557, `_bedrock_response_text` ~1075) are updated to describe the new behaviour.

## Ask 3 — expose the knob

`GRAPHIFY_MAX_RETRY_DEPTH` now overrides `max_retry_depth`, which was a Python-API kwarg only with no CLI flag, so a `graphify extract` operator could not lower it or set it to 0 as a mitigation. It follows the existing `_resolve_max_retries` / `_resolve_api_timeout` shape (garbage and negatives fall back to the default; `0` disables retries) and is documented in the README env table. Explicit Python-API callers still win — the resolution only fires when the kwarg is left as `None`.

## Tests

- `test_adaptive_retry_does_not_bisect_a_hollow_response` — asserts the sub-call chunk sizes are `[4, 4, 4]`, i.e. no halving, and that the files come back marked partial. This is the regression that would catch a future re-conflation.
- `test_adaptive_retry_recovers_a_transient_hollow_response` — 2 calls, all 4 nodes, nothing marked partial.
- `test_adaptive_retry_bisects_on_truncated_response` — the repurposed end-to-end test, now driven by a real truncation, so the bisect path stays covered.
- `test_call_openai_compat_labels_*_hollow` ×3 and `test_call_openai_compat_keeps_real_truncation_as_length` — the two labels stay distinct at the backend boundary.
- `test_max_retry_depth_reads_the_env_var`.

Full suite: no new failures (the 13 that fail on this machine fail identically on a clean `v8` — missing optional grammars and platform-specific install/watch tests).

## Not included

- **Ask 2 (a sub-call budget and a global ceiling).** With hollow off the bisect path, the remaining fan-out is driven only by genuine truncation, where each split does make the input smaller and `2**max_retry_depth` is already a real bound — and that bound is now operator-controllable via ask 3. A run-level call/token ceiling that aborts is a new config surface with its own semantics (what to do with the partial graph), so it seemed better as its own change than smuggled into this fix.
- **Ask 4 (emit `tokens:` on abnormal termination).** Separable observability work in `cli.py`, not part of this defect. Happy to open it separately.

### DIFF
--- a/README.md
+++ b/README.md
@@ -524,6 +524,7 @@ These are only needed for **headless / CI extraction** (`graphify extract`). Whe
 | `GRAPHIFY_MAX_OUTPUT_TOKENS` | Raise output cap for dense corpora | optional — e.g. `32768` for large files |
 | `GRAPHIFY_API_TIMEOUT` | Per-call timeout in seconds for HTTP, claude-cli, Anthropic SDK, and Bedrock backends (default: 600) | optional — also `--api-timeout` flag |
 | `GRAPHIFY_MAX_RETRIES` | How many times to retry a rate-limited (429) request before giving up (default: 6; honors `Retry-After`) | optional — raise for strict per-org limits (e.g. kimi); `0` disables |
+| `GRAPHIFY_MAX_RETRY_DEPTH` | How deep a truncated chunk may be bisected and re-extracted (default: 3, so up to 8x sub-calls for one chunk) | optional — lower it to cap worst-case spend; `0` disables adaptive retry |
 | `GRAPHIFY_FORCE` | Force graph rebuild even with fewer nodes | optional — also `--force` flag |
 | `GRAPHIFY_GOOGLE_WORKSPACE` | Auto-enable Google Workspace export | optional — set to `1` |
 | `GRAPHIFY_TRIAGE_BACKEND` | Backend for `graphify prs --triage` | optional — auto-detected from available keys |
--- a/README.md
+++ b/README.md
@@ -524,7 +524,7 @@ These are only needed for **headless / CI extraction** (`graphify extract`). Whe
 | `GRAPHIFY_MAX_OUTPUT_TOKENS` | Raise output cap for dense corpora | optional — e.g. `32768` for large files |
 | `GRAPHIFY_API_TIMEOUT` | Per-call timeout in seconds for HTTP, claude-cli, Anthropic SDK, and Bedrock backends (default: 600) | optional — also `--api-timeout` flag |
 | `GRAPHIFY_MAX_RETRIES` | How many times to retry a rate-limited (429) request before giving up (default: 6; honors `Retry-After`) | optional — raise for strict per-org limits (e.g. kimi); `0` disables |
-| `GRAPHIFY_MAX_RETRY_DEPTH` | How deep a truncated chunk may be bisected and re-extracted (default: 3, so up to 8x sub-calls for one chunk) | optional — lower it to cap worst-case spend; `0` disables adaptive retry |
+| `GRAPHIFY_MAX_RETRY_DEPTH` | How deep a truncated chunk may be bisected and re-extracted (default: 3, so up to 8x sub-calls for one chunk) | optional — lower it to cap worst-case spend; `0` disables every retry (no bisection, no hollow-response retry), so a chunk costs exactly one call |
 | `GRAPHIFY_FORCE` | Force graph rebuild even with fewer nodes | optional — also `--force` flag |
 | `GRAPHIFY_GOOGLE_WORKSPACE` | Auto-enable Google Workspace export | optional — set to `1` |
 | `GRAPHIFY_TRIAGE_BACKEND` | Backend for `graphify prs --triage` | optional — auto-detected from available keys |
--- a/graphify/llm.py
+++ b/graphify/llm.py
@@ -439,8 +439,12 @@ def _resolve_max_retry_depth(default: int = 3) -> int:
     A chunk of N files can split into up to ``2**depth`` pieces, so this is the
     knob that bounds worst-case cost. It used to be a Python-API kwarg only,
     with no way for a `graphify extract` operator to lower it — or set it to 0 —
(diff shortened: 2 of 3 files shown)


safishamsi:
Shipped in v0.9.48 via authorship-preserving cherry-pick. Chosen over the alternative #2887 for its crash-safety and its GRAPHIFY_MAX_RETRY_DEPTH=0 handling. Thanks @rajarshidattapy! Release: https://github.com/Graphify-Labs/graphify/releases/tag/v0.9.48

Issue #2182, opened 2026-07-25
Cross-file nodes whose labels normalize identically are never merged (making them less similar does merge them)


rays23:
**Version:** graphifyy 0.9.26 (fresh `uv tool install graphifyy==0.9.26`), macOS, Python 3.12.

## Summary

Two non-code nodes in **different** source files whose labels normalize to the same string are never merged. Pass 1 partitions by `source_file` and explicitly defers cross-file matches to Pass 2, but Pass 2's candidate filter drops the second node because its normalized label was already seen, so the pair Pass 1 handed over can never be formed.

The behavior is inverted in a way that makes it easy to spot: **changing one character so the labels no longer normalize equal makes them merge.**

## Repro

```python
from graphify.dedup import deduplicate_entities

def survivors(a, b):
    nodes = [
        {"id": "a", "label": a, "file_type": "concept", "source_file": "doc1.md"},
        {"id": "b", "label": b, "file_type": "concept", "source_file": "doc2.md"},
    ]
    kept, _ = deduplicate_entities(nodes, [], communities={"a": 0, "b": 0})
    return len(kept)

print(survivors("SHENZHEN INTERNATIONAL", "Shenzhen international"))  # 2  <- expected 1
print(survivors("SHENZHEN INTERNATIONAL", "Shenzhen internationaI"))  # 1  <- merges
```

Output on 0.9.26:

```
'SHENZHEN INTERNATIONAL' / 'Shenzhen international'   _norm equal: True    -> 2 survivors   <- BUG
'SHENZHEN INTERNATIONAL' / 'Shenzhen internationaI'   _norm equal: False   -> 1 survivor    (1 fuzzy)
```

Same outcome for any `concept`-typed pair in different files: `WeLab` / `Welab`, `Greater Bay Biotechnology` / `GREATER BAY BIOTECHNOLOGY`, and so on.

## Root cause

`dedup.py:414` (Pass 1) deliberately hands cross-file matches to Pass 2:

```python
# Partition by source_file — only merge within the same file in Pass 1.
# Cross-file matches fall through to Pass 2 fuzzy matching.
```

But Pass 2's candidate filter at `dedup.py:441-445` deduplicates the candidate list by normalized label:

```python
key = _norm(node.get("label", node.get("id", "")))
if key and key not in seen_norms:
    seen_norms.add(key)
    if _entropy(node.get("label", "")) >= _ENTROPY_THRESHOLD:
        candidates.append(node)
```

[...]

PR #2889, opened 2026-08-20
feat(extract): add --no-dedup so incremental merges can skip fuzzy dedup (#2881)

### LINKED ISSUE #2881: graphify update/extract --incremental loses unrelated nodes on merge — no CLI flag to disable fuzzy dedup
## `graphify update`/`extract --incremental` loses unrelated nodes on merge — no way to disable fuzzy dedup from the CLI

**Version:** graphifyy 0.8.28

**Summary:** Running `graphify update .` (or `graphify extract --incremental`) on a repo with an existing multi-thousand-node graph reproducibly loses 5-10% of nodes that belong to files **not touched** by the incremental diff. Confirmed 3 times across 3 separate sessions on the same repo, same failure pattern each time.

**Reproduction pattern:**
1. Repo has an existing `graphify-out/graph.json` with ~2950 nodes.
2. A small, correctly-detected incremental diff (9-39 real files) is extracted.
3. `build_merge()` merges the new chunks into the existing graph.
4. The merged result has *fewer* total nodes than the existing graph (e.g. 2925, 2831, 2742 across three separate runs) even though the diff only adds/updates nodes — it never should shrink outside of `prune_sources`.
5. Diffing node IDs shows the missing nodes belong to files that were **not part of the incremental diff at all** (e.g. `AdminPanel.tsx`, `LoginModal.tsx`, unrelated Python builtins like `date`/`int`).

**Root cause (traced in `graphify/dedup.py`, pass 2 — MinHash/LSH + Jaro-Winkler fuzzy matching):** the fuzzy dedup pass in `deduplicate_entities()` runs over the *combined* node set (existing graph + new chunk) on every incremental merge. When merging a small diff into a large existing graph, cross-file fuzzy matches between semantically-unrelated nodes (generic type names, short identifiers, common labels) get merged together via `_MERGE_THRESHOLD = 92.0` Jaro-Winkler similarity + community boost, collapsing legitimate pre-existing distinct nodes into one.

**What I found investigating a workaround:** `build_merge()` accepts `dedup: bool = True` and there's an unused safety net for exactly this case — `build.py` around the safety-check block:
```python
# Safety check: refuse to shrink the graph silently (#479)
# Skip when dedup or prune_sources is active — shrinkage is intentional there.
if graph_path.exists() and not dedup and not prune_sources:
    ...
    raise ValueError(f"graphify: build_merge would shrink graph from {existing_n} → {new_n} nodes. ...")
```
This shrink guard only activates when `dedup=False` — i.e. the one mode that would actually protect against this bug is the one where the guard is disabled by the `not dedup` condition. But `dedup=True` is hardcoded in `__main__.py`'s `update`/`extract --incremental` path (no `--no-dedup` CLI flag exists) with no way to opt out from the CLI.

**Request:** expose a CLI flag (`--no-dedup`, or scope fuzzy dedup to only run over the new/changed chunk's nodes rather than the full combined set) so incremental updates on large graphs can skip the fuzzy pass and get the shrink-guard for free. Alternatively, scoping Pass 2 candidates to only nodes touched by the current diff (plus their direct neighbors) would avoid the false-positive merges without disabling dedup entirely.

**Workaround in the meantime:** none via CLI — full re-extraction (`graphify extract .` without `--incremental`) avoids the merge path entirely but loses incremental speed, or a custom script calling `graphify.build.build_merge(chunks, dedup=False)` directly gets the shrink-guard but skips dedup for the new chunk too.

rajarshidattapy:
Fixes #2881.

Implements the issue's primary request: a CLI opt-out from entity dedup, so incremental updates on large graphs can skip the fuzzy pass and get the shrink guard for free.

## The gap

`build_merge()` already accepts `dedup: bool = True`, but the CLI hardcoded `dedup=True` on the incremental path (`cli.py`, the `if incremental_mode:` branch) with no flag to override it. That matters because the fuzzy pass runs over the **combined** node set — existing graph plus new chunk — so a small diff merged into a large graph compares every new node against every pre-existing one.

And as the issue points out, the safety net for exactly this case is inverted:

```python
# Safety check: refuse to SILENTLY drop nodes (#479, reworked in #2497)
if had_graph and not dedup:
```

The guard is **disabled** in the only mode reachable from the CLI, and **armed** in the mode that was unreachable from it.

## What this adds

`graphify extract <path> --no-dedup`:

- passes `dedup=False` to both `build()` (full build) and `build_merge()` (incremental merge);
- therefore arms build_merge's #479 shrink guard, which refuses to drop nodes from sources this run neither re-extracted nor pruned;
- surfaces that refusal as a clean `[graphify extract] …` error and exit 1 rather than an unhandled `ValueError` traceback. `graph.json` on disk is untouched when it fires, so the existing graph survives the refusal;
- is rejected together with `--dedup-llm` (exit 2). The LLM tiebreaker is pass 3 *inside* the dedup pipeline, so combining them would be a silent no-op that still demands an API key.

Documented in the README flag list and in `graphify extract`'s usage line.

## One note on the traced root cause

The issue was filed against 0.8.28, and on current `v8` both dedup passes now skip code symbols outright:

```python
# Code symbols are excluded from fuzzy matching too: two functions with
# similar long names in different files … must not be fuzzy-merged (#1205)
if _is_code(node):
    continue
```

So the specific losses reported — `AdminPanel.tsx`, `LoginModal.tsx`, Python builtins like `date`/`int` — should no longer be reachable through pass 2 on the current release; #1205, #1243, #2182, #1284 and #2576 each narrowed that path after 0.8.28. I have deliberately **not** claimed this PR fixes those merges, since I could not reproduce them on `v8`.

What is verifiably still true is the part this PR addresses: the incremental path had no opt-out, and the shrink guard was unreachable from the CLI. If the reporter still sees shrinkage on 0.9.x, `--no-dedup` now turns a silent 5-10% loss into a loud refusal naming the first unexplained node — which is the diagnostic the issue was missing.

## Not included

Scoping pass 2's candidate set to the current diff plus its direct neighbours (the issue's alternative suggestion). That changes dedup quality for every user on every run and wants its own benchmarking against the corpora that motivated the existing thresholds; it did not belong in the same change as an opt-out flag.

`graphify update` needs no flag: it routes through `watch._rebuild_code` → `build_from_json`, which never calls `deduplicate_entities`.

## Tests

`tests/test_no_dedup_flag.py` — the flag reaches `build()`, dedup stays on by default, the flag reaches `build_merge()` on a genuine second (incremental) run, and the `--dedup-llm` conflict exits 2. Three of the four fail without the change.

### DIFF
--- a/README.md
+++ b/README.md
@@ -748,6 +748,7 @@ graphify extract ./docs --no-cluster           # raw extraction only, skip clust
 graphify extract ./docs --timing               # print per-stage wall-clock timings to stderr (also works on cluster-only)
 graphify extract ./docs --force                # overwrite graph.json even if new graph has fewer nodes (use after refactors or to clear ghost duplicates)
 graphify extract ./docs --dedup-llm            # LLM tiebreaker for ambiguous entity pairs (uses same API key)
+graphify extract ./src --no-dedup              # skip entity dedup; on an incremental merge this also arms the shrink guard that refuses to drop untouched files' nodes
 graphify extract ./docs --global --as myrepo   # extract and register into the cross-project global graph
 GRAPHIFY_MAX_OUTPUT_TOKENS=32768 graphify extract ./docs --backend claude  # raise output cap for dense corpora
 
--- a/graphify/cli.py
+++ b/graphify/cli.py
@@ -2849,7 +2849,7 @@ def dispatch_command(cmd: str) -> None:
             print(
                 "Usage: graphify extract <path> [--backend gemini|kimi|claude|openai|deepseek|ollama] "
                 "[--model M] [--mode deep] [--out DIR|--output DIR] [--google-workspace] [--no-cluster] "
-                "[--no-gitignore] [--code-only] "
+                "[--no-gitignore] [--code-only] [--no-dedup] "
                 "[--max-workers N] [--token-budget N] [--max-concurrency N] "
                 "[--api-timeout S] [--postgres DSN] [--cargo] [--allow-partial] [--timing]",
                 file=sys.stderr,
@@ -2875,6 +2875,13 @@ def dispatch_command(cmd: str) -> None:
         cli_allow_partial: bool = False
         no_cluster = False
         dedup_llm = False
+        # --no-dedup: skip entity deduplication entirely. On an incremental
+        # merge the fuzzy pass runs over the COMBINED node set (existing graph +
+        # new chunk), so a small diff merged into a large graph can collapse
+        # pre-existing nodes from files the diff never touched. Turning dedup
+        # off also arms build_merge's #479 shrink guard, which is disabled while
+        # dedup is on because fuzzy merging shrinks the graph legitimately (#2881).
+        no_dedup = False
         google_workspace = False
         global_merge = False
         code_only = False
@@ -2943,6 +2950,8 @@ def dispatch_command(cmd: str) -> None:
                 no_cluster = True; i += 1
             elif a == "--dedup-llm":
                 dedup_llm = True; i += 1
+            elif a == "--no-dedup":
+                no_dedup = True; i += 1
             elif a == "--code-only":
                 code_only = True; i += 1
             elif a == "--google-workspace":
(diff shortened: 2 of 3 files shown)


safishamsi:
Shipped in v0.9.48 via authorship-preserving cherry-pick, with an added graph-level test locking the exact-id-collapse invariant. Thanks @rajarshidattapy! Release: https://github.com/Graphify-Labs/graphify/releases/tag/v0.9.48

PR #2890, opened 2026-08-20
fix(llm): recover JSON from reasoning-first model replies (#2882)

### LINKED ISSUE #2882: LLM responses not in JSON format: instructions instead of data
**Version**: 0.9.46

**Reproduction pattern**:

1. `graphify extract <source_code_dir> --backend (openai|gemini|nvidia|...)`

```
$ graphify provider show nvidia
{
  "nvidia": {
    "base_url": "https://integrate.api.nvidia.com/v1",
    "default_model": "nvidia/nemotron-3.5-lightning-30b-a3b",
    "env_key": "NVIDIA_API_KEY",
    "context_length": 65536,
    "pricing": {
      "input": 0.0,
      "output": 0.0
    },
    "temperature": 0
  }
}
```

The same errors are **not** observed for gemma4 because of the leak of CoT.

```
[graphify extract] semantic extraction on 4908 files via ...


[graphify] LLM returned invalid JSON, skipping chunk (first 200 chars: 'Let me analyze the provided source files to extract a knowledge graph fragment. I need to follow the rules carefully:\n\n1. EXTRACTED: relationship explicit in source\n2. INFERRED: reasonable inference\n3')
[graphify] chunk of 19 truncated at depth 2, splitting into halves of 9 and 10
[graphify] LLM returned invalid JSON, skipping chunk (first 200 chars: "Here's a thinking process:\n\n1.  **Analyze User Input:**\n   - User provided several `<untrusted_source>` blocks with file paths and content.\n   - The task is to extract a knowledge graph fragment from ")
[graphify] chunk of 9 still truncated at recursion depth 3 (max 3) — partial result kept (not cached as complete)
[graphify] LLM returned invalid JSON, skipping chunk (first 200 chars: "Here's a thinking process:\n\n1.  **Analyze User Input:**\n   - User is asking me to extract a knowledge graph fragment from provided files.\n   - There are images attached with source files listed: `docs")
[graphify] chunk of 3 still truncated at recursion depth 3 (max 3) — partial result kept (not cached as complete)
[graphify] LLM returned invalid JSON, skipping chunk (first 200 chars: 'Here\'s a thinking process:\n\n1.  **Analyze User Input:**\n   - User is providing image files as visual input.\n   - Instructions: Emit one node per image with `"file_type":"image"` and the listed source_')
[graphify] chunk of 6 truncated at depth 2, splitting into halves of 3 and 3
[graphify] LLM returned invalid JSON, skipping chunk (first 200 chars: 'Here, I need to extract a knowledge graph from the provided files. The user has given me a set of image files with their source_file paths. I need to output valid JSON with nodes and edges according t')
```

**Root cause**: Nemotron has/uses different responses.

**Request**: add certain harness to deal with Nemotron.

**Workaround**: Use other models or modify `~/.graphify/providers.json`

rajarshidattapy:
Fixes #2882.

## What was happening

`nvidia/nemotron-*` (and gemma, and any reasoning-first model) narrates before it answers. `_parse_llm_json` dropped the entire chunk when it did:

```
[graphify] LLM returned invalid JSON, skipping chunk (first 200 chars: 'Let me analyze the provided source files to extract a knowledge graph fragment. I need to follow the rules carefully:\n\n1. EXTRACTED: relationship explicit in source\n2. INFERRED: reasonable inference\n3')
[graphify] LLM returned invalid JSON, skipping chunk (first 200 chars: "Here's a thinking process:\n\n1.  **Analyze User Input:**\n…")
```

The parser already had a fence-stripper and a balanced-brace fallback, so the JSON was often *right there* in the response. Three separate weaknesses stopped it being found — each one is enough on its own to lose the chunk:

**1. The fence handler was destructive and picked the wrong fence.** It took the first ` ``` ` anywhere in the text, cut at the last one, and then *overwrote* the working text with the result. A narration that opens a fence of its own — ` ```text `, ` ```python ` — mangled the answer's ` ```json ` block before it was ever parsed. Every fenced block is now tried as a candidate against the **original** text, JSON-tagged and untagged fences first.

**2. The balanced-brace fallback tried exactly one candidate.** It scanned to the first `{` and, if that object failed to parse, `break` — done. Narration is full of braces ("The shape I must emit is { ... }", bullet lists, inline examples), so the answer further down was never reached. Every plausible start is tried now.

**3. A prose object could shadow the answer.** A restated schema like `{"description": "graph fragment"}` parses fine and comes first. An object carrying none of `nodes`/`edges`/`hyperedges` is now kept only as a last resort.

Plus: `<think>` / `<thinking>` / `<reasoning>` blocks are stripped up front, which is where deepseek-r1 and qwq put their chain of thought — prose that is nothing but braces to a brace scanner.

## Bounded, not just tolerant

Candidate starts are prioritized by whether an extraction key appears within 200 characters, and the likely/unlikely lists are capped **separately** at 64. That matters: a single shared cap would let a wall of noise braces crowd out an answer that comes after it, which is exactly the pathological case a cap is supposed to protect against. Covered by `test_candidate_scan_is_bounded`.

## Unchanged behaviour

A reply with genuinely no JSON anywhere still returns the empty fragment and logs the same line, so `_response_is_hollow` takes over exactly as before. The 10 MB `_LLM_JSON_MAX_BYTES` guard is untouched.

## Tests

`tests/test_llm_parser_reasoning.py` — the two verbatim response shapes from the issue, a `<think>`-block reply, the schema-shadowing case, the bounded scan, and the graceful-degradation case. Five of the six fail without the change. The existing 107 parser/backend tests pass unchanged, including `test_preamble_then_fence_is_parsed` and `test_fence_without_closing_backticks`.

## Related

The same responses were also being fed into adaptive retry as `finish_reason="length"` and bisected — visible in the issue's log as `chunk of 19 truncated at depth 2, splitting into halves of 9 and 10`, which is #2880. This PR makes the response parse in the first place; #2880 stops the ones that still cannot parse from costing 15 calls. They are independent and compose.

### DIFF
--- a/graphify/llm.py
+++ b/graphify/llm.py
@@ -1115,7 +1115,13 @@ def _parse_llm_json(raw: str) -> dict:
     except json.JSONDecodeError:
         pass
 
-    fallback: dict | None = None
+    # Preference ladder, weakest last. A model that restates the required shape
+    # before answering — "the schema is `{"nodes": [], "edges": []}`" — produces
+    # a candidate that carries the extraction keys but no content, and taking it
+    # would let the restatement shadow the answer just as surely as a prose
+    # object would (#2882).
+    empty_fragment: dict | None = None   # right shape, nothing in it
+    fallback: dict | None = None         # parses, but not a fragment at all
     for candidate in _json_fragment_candidates(stripped):
         try:
             parsed = json.loads(candidate)
@@ -1124,12 +1130,18 @@ def _parse_llm_json(raw: str) -> dict:
         if not isinstance(parsed, dict):
             continue
         if any(k in parsed for k in _FRAGMENT_KEYS):
-            return _sanitize_fragment(parsed)
-        if fallback is None:
+            if any(parsed.get(k) for k in _FRAGMENT_KEYS):
+                return _sanitize_fragment(parsed)
+            if empty_fragment is None:
+                empty_fragment = parsed
+        elif fallback is None:
             fallback = parsed
 
-    if fallback is not None:
-        return _sanitize_fragment(fallback)
+    # A genuinely empty extraction is still a valid answer, and still reads as
+    # hollow downstream, so it outranks an object that is not a fragment at all.
+    for weaker in (empty_fragment, fallback):
+        if weaker is not None:
+            return _sanitize_fragment(weaker)
 
     print(
         f"[graphify] LLM returned invalid JSON, skipping chunk "
--- a/graphify/llm.py
+++ b/graphify/llm.py
@@ -986,11 +986,85 @@ def _sanitize_fragment(parsed: dict) -> dict:
     return parsed
 
 
+# Keys that identify an extraction fragment. Used to tell the graph object
+# apart from a brace that merely appeared in the model's narration (#2882).
+_FRAGMENT_KEYS = ("nodes", "edges", "hyperedges")
+_FRAGMENT_KEY_TOKENS = tuple(f'"{k}"' for k in _FRAGMENT_KEYS)
+# Bound on how many `{` positions are probed, so a pathological response with
+# thousands of braces cannot turn recovery into a quadratic scan. Applied to
+# the likely and the unlikely candidate lists separately, so a wall of noise
+# braces cannot crowd out an answer that comes after it.
+_MAX_OBJECT_CANDIDATES = 64
+# Reasoning models (nemotron, deepseek-r1, qwq, …) emit their chain of thought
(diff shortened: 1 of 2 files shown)


safishamsi:
Shipped in v0.9.48 via authorship-preserving cherry-pick, plus a senior tweak: the winner gate now runs on sanitized content so a bare-string id sketch cannot shadow the real answer. Thanks @rajarshidattapy! Release: https://github.com/Graphify-Labs/graphify/releases/tag/v0.9.48

Issue #2442, opened 2026-08-04
SQL FK-reference edges use referencing-file stem instead of resolving to the defining file, producing dangling edges across multi-migration schemas


MG44CPA:
### Summary
When a SQL migration file has a foreign-key reference to a table defined in an *earlier* migration file, the extractor stamps the target node ID using the **referencing file's stem** instead of resolving to the table's actual defining file. This produces edges pointing at node IDs that don't exist anywhere in the extraction, showing up as dangling-endpoint edges in `graphify.diagnostics.diagnose_extraction`.

### Repro
Given two migrations:
```sql
-- 001_initial_schema.sql
create table public.profiles (id uuid primary key, ...);
```
```sql
-- 008_expansion_tables.sql
create table public.household_members (
  profile_id uuid references public.profiles(id),
  ...
);
```

The extractor correctly creates:
- `supabase_migrations_001_initial_schema_public_profiles` (the real `profiles` table node)

But the FK reference edge from `008_expansion_tables.sql` targets:
- `008_expansion_tables_public_profiles` (does not exist — wrong file stem)

instead of resolving to `supabase_migrations_001_initial_schema_public_profiles`.

### Impact
On a real project (24 SQL migration files, Supabase/Postgres), this produced **92 dangling edges** — every migration that FKs into a table defined in an earlier file is affected (observed in `004`, `005`, `008`, `013`, `018`, `021`, `022` referencing tables from `001` and `008`). All 92 are legitimate schema relationships; none are false edges — the extractor just can't resolve the correct target node ID across files.

### Expected behavior
The SQL extractor should maintain a project-wide symbol table (table name → defining file/node ID) built during the AST pass, and resolve `references public.<table>(...)` targets against that table rather than stamping the target ID with the local file's stem.

### Environment
- graphifyy 0.9.30
- `tree_sitter_sql` installed via `graphifyy[sql]`
- Reproduced via `graphify.diagnostics.diagnose_extraction(extraction, directed=False, root='.')` on a 24-file Supabase migrations folder

kreativekubedigital-creator:
Corroborating this from a second Supabase/Postgres corpus, with a **different observable symptom from the same root cause** — worth recording here because it changes how big the fix's payoff is.

**Setup:** graphifyy 0.9.65, `tree-sitter-sql` 0.3.11, 106 SQL migrations + 27 Markdown docs, built via the `/graphify` skill (undirected).

[...]

PR #2891, opened 2026-08-20
feat(sql): link SQL tables to the code that queries them (#2884)

### LINKED ISSUE #2884: SQL tables are never linked to the application code that queries them — the whole data layer lands as degree-1 orphans
# SQL tables are never linked to the application code that queries them — the whole data layer lands as degree-1 orphans

**Version:** graphify 0.9.46 (`graphifyy` via pipx), macOS, Python 3.14.3
**Corpus:** Next.js/TypeScript + MySQL app — 1,109 files (768 code: 298 `.ts`, 224 `.tsx`, 41 `.sql`), schema split across `db/schema.sql` + 38 numbered migrations.

## Summary

`tree-sitter-sql` extracts table declarations correctly, and `tree-sitter-typescript` extracts the application code correctly. But nothing connects the two: a table declared in `db/schema.sql` has no edge to the `.ts` file whose query reads `SELECT … FROM that_table`. Every table lands connected only to the file that declares it (`contains`, degree 1) and nothing else.

The result is that the data layer floats free of the application. On my corpus that is **65 orphaned SQL symbol nodes**, and `db/` never bridges a single community despite being queried from ~100 route files.

The practical cost: **the graph cannot answer "what breaks if I change this table?"** — which is the question a schema-heavy project most wants from a knowledge graph, and the one a migration-heavy history makes load-bearing.

## Reproduce

Any repo where SQL lives in `.sql` files and queries are string literals in application code:

```
db/schema.sql                    CREATE TABLE volunteer_assignments (…)
db/migrations/035_volunteers.sql CREATE TABLE volunteer_assignments (…)
src/app/api/partner/volunteers/assignments/route.ts
                                 await query(`SELECT … FROM volunteer_assignments WHERE …`)
                                 await query(`UPDATE volunteer_assignments SET …`)
src/app/api/partner/volunteers/roster/route.ts
                                 await query(`INSERT INTO volunteer_assignments …`)
```

```bash
graphify .
```

**Observed:** `volunteer_assignments` has degree 1 (only `contains`, from its declaring file). No edge to any of the three route files that query it.

**Expected:** an edge of some kind between the table and each file whose SQL text references it.

To confirm the tables are genuinely used rather than dead, I grepped nine of the orphans against `src/` — every one is referenced in 1–4 files:

```
role_change_audit             db/schema.sql:L42        → 1 src file
event_updates                 db/migrations/032_…:L22  → 4 src files
organization_module_contacts  db/schema.sql:L700       → 2 src files
maintenance_work_order_photos db/migrations/020_…:L3   → 3 src files
asset_documents               db/migrations/024_…:L83  → 3 src files
volunteer_roster_slots        db/migrations/035_…:L123 → 3 src files
volunteer_assignments         db/schema.sql:L1108      → 3 src files
user_email_preferences        db/schema.sql:L731       → 2 src files
account_status_audit          db/schema.sql:L54        → 1 src file
```

## Root cause

There is no resolver for *table name in SQL* → *same name inside a string literal in another language*. `resolver_registry.py` has language-specific resolvers (`pascal_resolution.py`, `ruby_resolution.py`, …), but the SQL↔host-language direction is a text-embedding problem, not an AST one: the table name appears inside a template literal that the TypeScript grammar sees only as a string.

I want to be explicit that the existing guardrail is **correct and should stay**. `references/extraction-spec.md` says:

> `calls` edges MUST stay within one language … cross-language call edges are phantom artifacts, never emit them.

That is right for `calls`. But it means schema↔code coupling is invisible *by construction*, and `references` is a different relation with different semantics — "this file's SQL names this table" is a factual, checkable claim, not an inferred call.

## What I did as a local workaround, and the precision trap in it

I wrote a post-hoc linker and got **921 edges across 75 of 80 tables**, dropping `db/` orphans from 65 → 4 and growing the largest connected component from 2,764 → 2,962 nodes.

The part worth passing upstream is **how easy it is to get this wrong**. My first attempt matched table names on a word boundary and applied the SQL-context test file-wide — i.e. "does this file contain SQL anywhere, and does this table name appear anywhere in it". That produced 1,489 edges and was badly wrong: common table names collided with ordinary identifiers.

| table | naive matcher | strict matcher |
|---|---|---|
| `events` | 157 files | 27 |
| `users` | 98 | 78 |
| `listings` | 95 | 32 |
| `facilities` | 94 | 44 |
| **total edges** | **1,489** | **921** |

The 130 phantom `events` edges were all JavaScript variables named `events`. Any implementation of this feature will hit the same trap — table names like `users`, `events`, `sales`, `assets`, `notifications` are extremely common identifiers.

What worked was requiring the table to sit in a real SQL position **in the same match**, not merely in the same file:

```python
KEYWORDS = r'(?:FROM|JOIN|INTO|UPDATE|REFERENCES|TABLE(?:\s+IF\s+NOT\s+EXISTS)?)'
rx = re.compile(KEYWORDS + r'\s+`?' + re.escape(table) + r'`?\b', re.I)
```

Run over the whole file text rather than line-by-line, so multi-line template literals still match. I hand-checked 12 sampled hits and all 12 were genuine (`INTO role_change_audit`, `UPDATE volunteer_assignments`, `FROM asset_documents`, …). Backtick handling matters for MySQL; presumably `"…"` for Postgres and `[…]` for T-SQL (cf. #2712).

A useful side effect: the 5 tables with no SQL-position reference anywhere were all explicable — a backup table, a temp table inside a migration, one queried only from shell scripts, one likely a mis-extracted CTE/column, and one genuinely unused. That last category makes this a dead-schema detector too.

## Suggested shape

- A resolver that, after per-file extraction, scans non-SQL source text for declared table names in SQL keyword positions and emits `references` edges (file → table), `EXTRACTED`, `confidence_score` 1.0.
- Link to **every** declaration site (both `schema.sql` and the migration that created the table), so tracing a table also reaches its migration history.
- Keep it lexical and free — no LLM needed; this ran over 537 files in a couple of seconds.
- Probably wants an opt-out flag, since a repo with an ORM rather than raw SQL would get little from it.

## Scope / non-goals

- File-level only. Resolving to the *specific function* holding the query needs the enclosing-scope span, which I did not attempt.
- Raw-SQL-in-string-literal only. ORM-generated queries (Prisma, Drizzle, ActiveRecord) name tables via model classes and would need separate handling.
- Related but distinct: #1572 (table→table via INSERT/CTAS/MERGE) and #2442 (FK edges resolving to the referencing file's stem) are both *within* SQL. This issue is SQL↔host-language.

Happy to open a PR against `resolver_registry.py` if this shape sounds right.

rajarshidattapy:
Fixes #2884.

## The gap

`tree-sitter-sql` extracts table declarations correctly and the host-language grammars extract the application code correctly — but nothing connected the two. A table declared in `db/schema.sql` had no edge to the `.ts` file whose query reads `SELECT … FROM that_table`, so every table landed connected only to its declaring file (`contains`, degree 1). On the reporter's corpus that was **65 orphaned SQL symbol nodes**, with `db/` never bridging a single community despite being queried from ~100 route files.

The practical cost: the graph could not answer *"what breaks if I change this table?"* — the question a schema-heavy project most wants from a knowledge graph.

There is no AST route to it. The table name lives inside a template literal that the TypeScript grammar sees only as a string, so this is a text-embedding problem. The new resolver is lexical by necessity.

## What it does

`graphify/sql_resolution.py` registers into `resolver_registry` (gated on `.sql` being present in the corpus, like every other language resolver). After per-file extraction it scans every non-SQL source for each declared table name and emits:

`file --references--> table`, `EXTRACTED`, `confidence_score: 1.0`, `context: "sql_table"`

to **every** declaration site — `schema.sql` *and* the migration that created the table — so tracing a table also reaches its migration history, as the issue asks.

A declaration is the target of a `contains` edge from a `.sql` file, which is exactly how the SQL extractor anchors what it defines. The sourceless reference stubs it mints for tables defined in *another* file carry no `contains` edge and are correctly excluded — a stub is not a declaration site.

## Precision — the trap the issue warns about

The reporter measured it, and the numbers are the reason this PR is shaped the way it is:

| table | naive matcher | strict matcher |
|---|---|---|
| `events` | 157 files | 27 |
| `users` | 98 | 78 |
| `listings` | 95 | 32 |
| **total edges** | **1,489** | **921** |

All 130 phantom `events` edges were JavaScript variables named `events`. Table names like `users`, `events`, `sales`, `notifications` are extremely common identifiers, so "does this file contain SQL anywhere, and does this name appear anywhere in it" is not good enough.

The table must sit in a SQL keyword position **in the same match**:

```python
_SQL_KEYWORDS = r"(?:FROM|JOIN|INTO|UPDATE|REFERENCES|TABLE(?:\s+IF\s+NOT\s+EXISTS)?)"
```

Matching runs over the whole file text rather than line by line, so a table named on the second line of a multi-line template literal still matches. Backtick (MySQL), double-quote (standard SQL) and bracket (T-SQL) quoting are all handled, per #2712.

`test_a_table_named_like_a_variable_is_not_falsely_linked` pins this: a file full of `const events = […]` — including the word `SELECT` in a comment — gets no edge, while the genuine `FROM volunteer_assignments` in a sibling file still does.

## On the `calls` guardrail

`references/extraction-spec.md` is right and unchanged: `calls` edges stay within one language. `references` is a different relation with different semantics — "this file's SQL names this table" is a factual, checkable claim, not an inferred call. That distinction is written into the edge-emitting code so it survives future edits.

## Cost and opt-out

Lexical and free — no LLM. Files are read once, capped at 2 MiB, behind a cheap "is there a SQL keyword at all" pre-filter that skips the per-table scan for the overwhelming majority of files. `GRAPHIFY_NO_SQL_LINKS=1` skips the pass entirely, documented in the README env table: an ORM-based repo names tables via model classes and gets little from it. An env var rather than a CLI flag because resolvers have no access to CLI args, and this needed no new plumbing to be usable.

## Scope, matching the issue

- **File-level only.** Resolving to the specific function holding the query needs the enclosing-scope span; not attempted.
- **Raw SQL in string literals only.** Prisma/Drizzle/ActiveRecord name tables via model classes and would need separate handling.
- **Not** #1572 (table→table via INSERT/CTAS/MERGE) or #2442 (FK edges), which are both *within* SQL. `.sql` files are excluded from the scan for that reason.

## Tests

`tests/test_sql_table_references.py` — the issue's `volunteer_assignments` scenario end to end (two routes, two declaration sites, four edges), the degree-1-orphan regression, the `events`-variable precision trap, dialect quoting across MySQL/Postgres/T-SQL, the opt-out, and the dead-schema side effect (a table with no SQL-position reference anywhere gets no edges, which makes this a dead-schema detector too). Four of the six fail without the resolver.

Full suite: no new failures — the 22 that fail on this machine fail identically on a clean `v8` (missing `tree_sitter_hcl`, and Windows-specific install/symlink/FIFO tests).

### DIFF
--- a/README.md
+++ b/README.md
@@ -524,7 +524,7 @@ These are only needed for **headless / CI extraction** (`graphify extract`). Whe
 | `GRAPHIFY_MAX_OUTPUT_TOKENS` | Raise output cap for dense corpora | optional — e.g. `32768` for large files |
 | `GRAPHIFY_API_TIMEOUT` | Per-call timeout in seconds for HTTP, claude-cli, Anthropic SDK, and Bedrock backends (default: 600) | optional — also `--api-timeout` flag |
 | `GRAPHIFY_MAX_RETRIES` | How many times to retry a rate-limited (429) request before giving up (default: 6; honors `Retry-After`) | optional — raise for strict per-org limits (e.g. kimi); `0` disables |
-| `GRAPHIFY_NO_SQL_LINKS` | Skip the pass that links SQL tables to the code whose queries name them | optional — set to `1` for ORM-based repos, which name tables via model classes rather than in SQL text |
+| `GRAPHIFY_NO_SQL_LINKS` | Skip the pass that links SQL tables to the code whose queries name them | optional — also `--no-sql-links` flag; useful for ORM-based repos, which name tables via model classes rather than in SQL text |
 | `GRAPHIFY_FORCE` | Force graph rebuild even with fewer nodes | optional — also `--force` flag |
 | `GRAPHIFY_GOOGLE_WORKSPACE` | Auto-enable Google Workspace export | optional — set to `1` |
 | `GRAPHIFY_TRIAGE_BACKEND` | Backend for `graphify prs --triage` | optional — auto-detected from available keys |
@@ -749,6 +749,7 @@ graphify extract ./docs --no-cluster           # raw extraction only, skip clust
 graphify extract ./docs --timing               # print per-stage wall-clock timings to stderr (also works on cluster-only)
 graphify extract ./docs --force                # overwrite graph.json even if new graph has fewer nodes (use after refactors or to clear ghost duplicates)
 graphify extract ./docs --dedup-llm            # LLM tiebreaker for ambiguous entity pairs (uses same API key)
+graphify extract ./src --no-sql-links          # skip linking SQL tables to the code whose queries name them (useful for ORM-based repos)
 graphify extract ./docs --global --as myrepo   # extract and register into the cross-project global graph
 GRAPHIFY_MAX_OUTPUT_TOKENS=32768 graphify extract ./docs --backend claude  # raise output cap for dense corpora
 
--- a/README.md
+++ b/README.md
@@ -524,6 +524,7 @@ These are only needed for **headless / CI extraction** (`graphify extract`). Whe
 | `GRAPHIFY_MAX_OUTPUT_TOKENS` | Raise output cap for dense corpora | optional — e.g. `32768` for large files |
 | `GRAPHIFY_API_TIMEOUT` | Per-call timeout in seconds for HTTP, claude-cli, Anthropic SDK, and Bedrock backends (default: 600) | optional — also `--api-timeout` flag |
 | `GRAPHIFY_MAX_RETRIES` | How many times to retry a rate-limited (429) request before giving up (default: 6; honors `Retry-After`) | optional — raise for strict per-org limits (e.g. kimi); `0` disables |
+| `GRAPHIFY_NO_SQL_LINKS` | Skip the pass that links SQL tables to the code whose queries name them | optional — set to `1` for ORM-based repos, which name tables via model classes rather than in SQL text |
 | `GRAPHIFY_FORCE` | Force graph rebuild even with fewer nodes | optional — also `--force` flag |
 | `GRAPHIFY_GOOGLE_WORKSPACE` | Auto-enable Google Workspace export | optional — set to `1` |
 | `GRAPHIFY_TRIAGE_BACKEND` | Backend for `graphify prs --triage` | optional — auto-detected from available keys |
--- a/graphify/cli.py
+++ b/graphify/cli.py
@@ -2849,7 +2849,7 @@ def dispatch_command(cmd: str) -> None:
             print(
                 "Usage: graphify extract <path> [--backend gemini|kimi|claude|openai|deepseek|ollama] "
                 "[--model M] [--mode deep] [--out DIR|--output DIR] [--google-workspace] [--no-cluster] "
-                "[--no-gitignore] [--code-only] "
+                "[--no-gitignore] [--code-only] [--no-sql-links] "
                 "[--max-workers N] [--token-budget N] [--max-concurrency N] "
                 "[--api-timeout S] [--postgres DSN] [--cargo] [--allow-partial] [--timing]",
                 file=sys.stderr,
@@ -2875,6 +2875,12 @@ def dispatch_command(cmd: str) -> None:
         cli_allow_partial: bool = False
         no_cluster = False
         dedup_llm = False
+        # --no-sql-links: skip the pass that links SQL tables to the code whose
+        # queries name them (#2884). On by default, like every other language
+        # resolver, because a schema that is invisible to the graph is the
+        # problem it exists to fix; an ORM-based repo names tables via model
+        # classes and gets little from it.
+        no_sql_links = False
         google_workspace = False
         global_merge = False
         code_only = False
@@ -2943,6 +2949,8 @@ def dispatch_command(cmd: str) -> None:
                 no_cluster = True; i += 1
             elif a == "--dedup-llm":
                 dedup_llm = True; i += 1
+            elif a == "--no-sql-links":
+                no_sql_links = True; i += 1
             elif a == "--code-only":
                 code_only = True; i += 1
             elif a == "--google-workspace":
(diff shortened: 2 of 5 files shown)


safishamsi:
Thanks for this @rajarshidattapy — the architecture is right (resolver-registry wiring, the whole-file keyword pre-filter, the scan-size cap, and a clean --no-sql-links / GRAPHIFY_NO_SQL_LINKS opt-out are all sound), and unmatched tables correctly produce no edge. Two things need to change before it can land, both reproducible today:

1) False edges from comments and non-SQL strings. The matcher scans the raw file bytes, so a table name after a keyword mints an edge even when it is not a real query. All of these currently produce a phantom `references` edge:
   - `// SELECT foo FROM events` (a comment)
   - `const s = "the FROM users keyword";` (a plain string)
   - `"""Docs: you can JOIN users with assets here."""` (a docstring)
   This is the same phantom-edge class the issue itself calls out. Suggested direction: strip line/block comments before scanning (cheap per-language heuristics for `//`, `#`, `/* */`, `--`), and ideally require the keyword+table to sit inside a string/template literal rather than free code. The existing negative test only covers the bare-identifier case; please add the three cases above and make them pass.

2) Schema-qualified tables never link. `_declared_tables` rejects any label that is not a bare identifier, but the SQL extractor stores labels verbatim, so `CREATE TABLE public.users` has label `public.users` and is filtered out — a Postgres/T-SQL corpus gets zero edges. The extractor already has `_norm_ident` for this; reusing it in the resolver (match on the normalized/bare name, and handle the schema-qualified reference form in code) would fix it. A test with a `public.users` schema plus `SELECT ... FROM users` and `FROM public.users` would lock it in.

Happy to re-review once those two are addressed.

Issue #1831, opened 2026-07-12
to_graphml crashes on dict/list-valued attributes (node metadata, graph-level hyperedges)


hofmockel:
## Bug

`graphify export graphml` crashes on any graph that has a dict or list-valued attribute — e.g. a per-node `metadata` dict, or the graph-level `hyperedges` list set by `attach_hyperedges()`:

```
TypeError: GraphML does not support type <class 'dict'> as data values.
TypeError: GraphML does not support type <class 'list'> as data values.
```

`nx.write_graphml` only accepts scalar attribute values. `to_graphml` already coerces `None -> ""` for this exact reason (a prior fix for #1502), but dict/list values weren't handled the same way.

## Repro

```python
import networkx as nx
from graphify.export import to_graphml

G = nx.Graph()
G.add_node("a", metadata={"kind": "file"})
G.add_node("b")
G.add_edge("a", "b")
G.graph["hyperedges"] = [{"nodes": ["a", "b"], "label": "x"}]

to_graphml(G, {0: ["a", "b"]}, "/tmp/out.graphml")
# TypeError: GraphML does not support type <class 'list'> as data values.
```

Also reproduces on a real ~2,300-node/4,300-edge project graph (not just this synthetic case) — every export attempt fails, and a failed attempt leaves a 0-byte `.graphml` file on disk that downstream tooling can mistake for a completed (if empty) export.

## Fix

Opened #1830: extends `to_graphml`'s existing None-scrubbing step to also JSON-serialize dict/list-valued attributes across graph-level, node, and edge scopes before calling `nx.write_graphml`. Includes a regression test mirroring the existing None-value test.

safishamsi:
Fixed in `a46eee4` (ships in 0.9.14). `to_graphml` now coerces `None -> ""` and JSON-serializes dict/list (and any other non-scalar) values across graph-level, node, and edge scopes, while GraphML-native int/float/bool/str pass through unchanged. Also addressed the 0-byte-file hazard you flagged: the export now writes to a sibling temp file and `os.replace`s it on success, so a failed export can't leave a partial/empty `.graphml` behind. Verified with a round-trip read (dict/list come back as JSON, scalars keep their types) plus a regression test. Thanks for the clean repro and for spotting the partial-file issue too, @hofmockel.

PR #2899, opened 2026-08-20
fix(export): a control character in a label must not abort the export (#2897)

### LINKED ISSUE #2897: A control character in a label aborts the whole GraphML and Obsidian export (ANSI escape in a heading, form feed in source)
## Summary

A C0 control character in a node label aborts two exporters outright. The character arrives from ordinary corpus content, passes detection, extraction and the graph build without complaint, and then costs the entire export.

- `to_graphml` raises `ValueError: All strings must be XML compatible: Unicode or ASCII, no NULL bytes or control characters` — XML 1.0 permits tab, LF and CR and no other C0 control.
- `to_obsidian` raises `OSError: [Errno 22] Invalid argument` on Windows, which rejects control characters in a path outright. One bad label costs the whole vault, not one note.

## Reproduce

One markdown file whose heading carries an ANSI colour sequence — what you get by pasting a terminal capture into docs:

```python
(src / "notes.md").write_text(
    "# Release \x0c Notes\n\nSome text.\n\n## Build \x1b[31mlog\x1b[0m capture\n\nmore\n",
    encoding="utf-8")
```

Then detect → extract → build → export:

```
nodes=5   labels with a C0 control char: 1
    'Build \x1b[31mlog\x1b[0m capture'

  graphml   CRASH ValueError: All strings must be XML compatible: Unicode or ASCII, no NULL bytes or control characters
  cypher    OK
  json      OK
  obsidian  CRASH OSError: [Errno 22] Invalid argument
  html      OK
```

Narrowed to the exact character classes:

```
CRASH  NUL \x00 | BEL \x07 | BS \x08 | VT \x0b | FF \x0c | ESC \x1b | US \x1f
OK     TAB \x09 | LF \x0a  | CR \x0d | DEL \x7f
```

## Why this is ordinary content, not hostile input

- **ANSI escapes** (`\x1b`) appear whenever a terminal capture is committed into a `.md`/`.txt` and a heading includes coloured output.
- **Form feed** (`\x0c`) is a long-standing Emacs/Python convention for a page separator inside source files.

Neither is exotic, and nothing earlier in the pipeline objects. The failure surfaces only at the very end, after the expensive work is done.

## The codebase already knows

This reads as a gap rather than a deliberate choice, because the other paths handle it:

- `_cypher_escape` explicitly drops C0 controls — its docstring says *"First normalise: drop NUL and other C0 control chars except tab."*
- `to_html` routes labels through `security.sanitize_label`, whose docstring is *"Strip control characters and cap length."*

Only `to_graphml` and `to_obsidian` pass the raw label straight through to a writer that cannot take it.

## Suggested fix

Sanitise at those two boundaries, matching what the other exporters already do:

- **GraphML** — filter strings in the existing `_graphml_safe` coercion hook (the one added for the `None`/non-scalar cases in #1831), dropping only the C0 controls XML forbids so tab/LF/CR survive. Node IDs need the same treatment: they become the `id` attribute of every `<node>` and edge endpoint, so a control character there is equally fatal.
- **Obsidian** — `_obsidian_safe_stem` already folds CR/LF to spaces; folding the remaining C0 controls (and DEL) the same way keeps them out of a path Windows rejects.

Losing an escape character is the right cost here. Losing the export is not.

## Environment

| | |
|---|---|
| Python | 3.12, Windows 11 |

abhay-codes07:
Fixes #2897.

## The bug

Labels reach the exporters unfiltered from the corpus. Two of them die on a C0 control character, and they take the whole export with them:

```
to_graphml  -> ValueError: All strings must be XML compatible: Unicode or ASCII,
               no NULL bytes or control characters
to_obsidian -> OSError: [Errno 22] Invalid argument
```

End to end from one markdown file whose heading carries an ANSI colour sequence — what you get by pasting a terminal capture into docs:

```
nodes=5   labels with a C0 control char: 1
    'Build \x1b[31mlog\x1b[0m capture'

  graphml   CRASH ValueError: All strings must be XML compatible...
  cypher    OK
  json      OK
  obsidian  CRASH OSError: [Errno 22] Invalid argument
  html      OK
```

Detection, extraction and the graph build all accept the file happily; the failure lands at the very end, after the expensive work, and costs the entire artifact rather than one note.

Narrowed to exact classes:

```
CRASH  NUL \x00 | BEL \x07 | BS \x08 | VT \x0b | FF \x0c | ESC \x1b | US \x1f
OK     TAB \x09 | LF \x0a  | CR \x0d | DEL \x7f
```

Neither trigger is exotic: **ANSI escapes** come from committed terminal captures, and **form feed** is the long-standing Emacs/Python page separator inside source files.

## Why this reads as a gap, not a decision

The codebase already handles this everywhere else:

- `_cypher_escape` drops C0 controls, and says so — *"First normalise: drop NUL and other C0 control chars except tab."*
- `to_html` routes labels through `security.sanitize_label`, whose docstring is *"Strip control characters and cap length."*

Only these two paths hand the raw label to a writer that cannot take it.

## The change

**GraphML** — strings are filtered in `_graphml_safe`, the coercion hook that already existed for the `None`/non-scalar cases (#1831), so there is one place where "can this value be written" is decided. Only the C0 controls XML forbids are dropped; tab, LF and CR are valid XML and are preserved, with a test pinning that so the fix cannot creep into ordinary whitespace.

Node IDs go through the same filter via `relabel_nodes`. They become the `id` attribute of every `<node>` and edge endpoint, so a control character there is exactly as fatal as one in a value — and a hand-built graph can carry one even though normalised ids never do.

**Obsidian** — `_obsidian_safe_stem` already folded CR/LF to spaces; the remaining C0 controls and DEL now fold the same way instead of being carried into a path Windows rejects with EINVAL.

Losing an escape character is the right cost. Losing the export is not.

## Deliberately not included

`to_svg` does **not** crash, but matplotlib warns `Glyph 27 () missing from font(s) DejaVu Sans` and renders a tofu box where the escape was. Same root cause, much smaller symptom, and fixing it means touching the label path of a renderer I would want to test properly rather than fold into a crash fix. Happy to follow up if you want the exporters fully uniform.

## Tests

`tests/test_export_control_characters.py` (33 tests): every crashing character class against both exporters, the XML-legal whitespace surviving untouched, a control character in a node **id**, the readable part of the label still reaching the file, an all-control label still yielding a usable name, and the two exporters that already coped continuing to cope. Plus a clean label round-tripping byte-identical, so the fix is invisible for ordinary input.

Reverting `export.py` and keeping the tests fails **21 of 33**.

## Validation

- Full suite: `15 failed, 4738 passed` -> `15 failed, 4771 passed`. **Identical failure set** — no regressions; the +33 are the new tests. The 15 are pre-existing Windows failures (symlink privileges, FIFO/socket fixtures, and the install-scope ones in my open #2802).

### DIFF
--- a/graphify/export.py
+++ b/graphify/export.py
@@ -502,6 +502,31 @@ def to_cypher(G: nx.Graph, output_path: str) -> None:
 generate_html = to_html
 
 
+# Characters XML 1.0 cannot carry: the C0 controls except tab, LF and CR.
+_XML_ILLEGAL_RE = re.compile("[\x00-\x08\x0b\x0c\x0e-\x1f]")
+
+
+def _strip_xml_illegal(s: str) -> str:
+    """Drop characters XML 1.0 cannot represent, leaving tab/LF/CR intact.
+
+    ``nx.write_graphml`` raises ``ValueError("All strings must be XML
+    compatible: Unicode or ASCII, no NULL bytes or control characters")`` on any
+    of them and aborts the whole export over a single label. Labels arrive
+    unfiltered from the corpus, so this is ordinary content rather than hostile
+    input: an ANSI escape in a markdown heading pasted from a terminal capture,
+    or the form feed some Python/Emacs sources use as a section separator
+    (#2897).
+    """
+    return _XML_ILLEGAL_RE.sub("", s)
+
+
+# C0 controls and DEL, folded to a space when building a filename stem. Windows
+# rejects them in a path outright with OSError EINVAL, so one of them in a label
+# aborted a whole Obsidian vault export; POSIX would accept the name but leave a
+# note nothing can comfortably open (#2897).
+_CONTROL_TO_SPACE_RE = re.compile("[\x00-\x1f\x7f]")
+
+
 def _cap_filename(s: str, limit: int = 200) -> str:
     """Cap a filename stem to ``limit`` UTF-8 bytes so it stays under the 255-byte
     filesystem limit even after the ``.md`` extension and dedup suffix are added
@@ -531,7 +556,11 @@ def _obsidian_safe_stem(label: str, limit: int = 200) -> str:
     cleaned = re.sub(
         r'[\\/*?:"<>|#^[\]]',
         "",
-        label.replace("\r\n", " ").replace("\r", " ").replace("\n", " "),
+        # CR/LF were already folded to spaces here; every other C0 control now
+        # goes the same way. They are not merely awkward in a filename — Windows
+        # rejects them outright, so a single one aborted the whole vault export
+        # rather than spoiling one note (#2897).
+        _CONTROL_TO_SPACE_RE.sub(" ", label),
     ).strip()
     cleaned = re.sub(r"\.(md|mdx|qmd|markdown)$", "", cleaned, flags=re.IGNORECASE)
     # Obsidian treats a leading-dot filename as a hidden file (#2205). Only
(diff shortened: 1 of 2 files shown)


safishamsi:
Shipped in v0.9.48 via authorship-preserving cherry-pick. Thanks @abhay-codes07! Release: https://github.com/Graphify-Labs/graphify/releases/tag/v0.9.48

PR #2904, opened 2026-08-20
fix(llm): estimate PDF tokens from the extracted text, not the container (#2903)

### LINKED ISSUE #2903: PDF token estimates are computed from the compressed container, not the extracted text — chunks overfill by ~3.5x and get bisected
## Summary

`_read_files` sends a PDF through `_file_to_text` → `extract_pdf_text`, but `_estimate_file_tokens` reads the file with `read_text` and tokenises *that*. A PDF's bytes are not its text — every real PDF Flate-compresses its content streams — so the estimate measures a compressed binary and comes out several times too small.

`_pack_chunks_by_tokens` sizes chunks from that estimate, so the chunk overfills, the request blows the model's context window, and `_extract_with_adaptive_retry` bisects: the same content is paid for two, four or eight times before it fits.

## Reproduce

One 400-line fixture, the same text, stored two ways:

```
uncompressed             bytes=28470  est=5559  actual=4598  actual/est=0.83x
FlateDecode (real PDFs)  bytes= 1728  est=1334  actual=4599  actual/est=3.45x
```

`est` is `_estimate_file_tokens(path)`; `actual` is the tokenised output of `_read_files([path], root)` — the prompt that will really be sent.

The two rows hold **identical text**. The only difference is the container, and the estimate moves by 4x. Uncompressed PDFs are roughly right by luck; compressed ones — which is all of them in practice — undercount by 3.45x here, and the ratio scales with how well the text compresses.

## Why it is easy to miss

Nothing looks broken. There is no wrong output and no error the user sees: the run just costs several times what it should and takes correspondingly longer, on a corpus of papers, which is exactly the corpus a PDF implies. The bisection machinery works as designed and silently absorbs the mis-estimate.

## Suggested fix

Estimate from the same text the prompt will carry — route PDFs through `_file_to_text` in `_estimate_file_tokens`, as `_read_files` already does.

Worth memoising: packing probes the same file repeatedly while deciding where a chunk ends, and extraction is the expensive part. Keying on `(path, size, mtime)` keeps a corpus of papers to one parse per run without serving a stale estimate for a file rewritten mid-run.

The tokenizer-absent fallback deserves the same treatment: it currently uses `stat().st_size`, which for a PDF is the compressed size and is wrong in the same direction.

## Related

Same class as #2900 — the estimate and slicing paths read the raw file while the prompt path goes through `_file_to_text`. That issue notes PDFs still need the converter threaded through `read_slice_text`; this is the independent half in token estimation, which costs money even where slicing never applies.

## Environment

| | |
|---|---|
| Python | 3.12, Windows 11 |

abhay-codes07:
Fixes #2903.

## The bug

`_read_files` sends a PDF through `_file_to_text` → `extract_pdf_text`. `_estimate_file_tokens` read the file with `read_text` and tokenised *that* — so it measured a compressed binary, because every real PDF Flate-compresses its content streams.

One 400-line fixture, **identical text**, stored two ways:

```
uncompressed             bytes=28470  est=5559  actual=4598  actual/est=0.83x
FlateDecode (real PDFs)  bytes= 1728  est=1334  actual=4599  actual/est=3.45x
```

`est` is `_estimate_file_tokens`; `actual` is the tokenised output of `_read_files` — the prompt that really gets sent. The only difference between the rows is the container, and the estimate moves by 4x.

`_pack_chunks_by_tokens` sizes chunks from that number, so a 3.45x undercount overfills the chunk, the request blows the context window, and `_extract_with_adaptive_retry` bisects — paying for the same content two, four or eight times before it fits. On a corpus of papers, which is what a PDF implies.

**Nothing looks broken.** No wrong output, no error the user sees; the run just costs several times what it should. The bisection machinery works exactly as designed and quietly absorbs the mis-estimate, which is why this can sit there indefinitely.

After the fix, both encodings estimate **4582** against 4598/4599 actual — 1.00x, the remaining gap being the wrapper-overhead constant.

## Scope

PDFs only. Every other path still measures its own bytes, because for them `read_text` *is* what the prompt carries.

The tokenizer-absent fallback gets the same correction: it used `stat().st_size`, which for a PDF is the compressed size and is wrong in the same direction, so the fallback no longer disagrees with the tiktoken path about what a PDF costs.

## The memo

Estimating a PDF now means extracting its text, and packing probes the same file repeatedly while deciding where a chunk ends — so `_pdf_text_for_estimate` memoises on `(path, size, mtime_ns)`. A corpus of papers is parsed once per run rather than once per probe.

Keyed on size and mtime rather than path alone so a file rewritten mid-run is re-read instead of served a stale estimate (there is a test for that), and bounded so a large corpus cannot pin every paper's text in memory.

## Tests

`tests/test_pdf_token_estimate.py` (10 tests). The fixtures build real one-page PDFs with a genuine text layer, one raw and one FlateDecode, and first assert the *precondition* — that both hold identical text and the compressed one really is a fraction of the size — so any estimate gap is the bug rather than the fixture.

Then: the estimate tracks the prompt it will build for both encodings, compression does not change the estimate (the tell — it used to differ by 4x purely because one was compressed), the estimate is not derived from file size, repeated probes agree, a rewritten PDF is re-estimated, a corrupt PDF estimates 0 rather than raising, non-PDF estimates are untouched, and the `_FILE_CHAR_CAP` truncation still bounds the result.

Reverting `llm.py` and keeping the tests fails **3 of 10**.

One of those tests caught me while writing it: my first version grew the fixture from 400 to 800 lines to prove the memo re-reads, and it failed — both sizes are past `_FILE_CHAR_CAP`, so they legitimately estimate the same and the test proved nothing. It now stays under the cap on purpose, and says so.

## Validation

- Full suite: `15 failed, 4738 passed` -> `15 failed, 4748 passed`. **Identical failure set** — no regressions; the +10 are the new tests.

## Related

Same class as #2900/#2902: the estimate and slicing paths read the raw file while the prompt path goes through `_file_to_text`. That PR notes PDFs still need the converter threaded through `read_slice_text`; this fixes the independent half in token estimation, which costs money even where slicing never applies. The two do not overlap in code.

### DIFF
--- a/graphify/llm.py
+++ b/graphify/llm.py
@@ -1884,6 +1884,33 @@ def extract_files_direct(
     return result
 
 
+# Estimating a PDF means extracting its text, and packing asks for the same
+# file repeatedly while it decides where a chunk ends. Memoise on
+# (path, size, mtime) so a corpus of papers is parsed once per run rather than
+# once per packing probe, and so a file rewritten mid-run is not served a stale
+# estimate. Bounded because a huge corpus should not pin every paper's text in
+# memory; the entries are cheap (an int) but the dict should not grow forever.
+_PDF_ESTIMATE_CACHE: "dict[tuple, str]" = {}
+_PDF_ESTIMATE_CACHE_MAX = 512
+
+
+def _pdf_text_for_estimate(path: Path) -> str:
+    """Extracted text of a PDF, memoised for the packing pass."""
+    try:
+        st = path.stat()
+        key = (str(path), st.st_size, st.st_mtime_ns)
+    except OSError:
+        return ""
+    hit = _PDF_ESTIMATE_CACHE.get(key)
+    if hit is not None:
+        return hit
+    text = _file_to_text(path)
+    if len(_PDF_ESTIMATE_CACHE) >= _PDF_ESTIMATE_CACHE_MAX:
+        _PDF_ESTIMATE_CACHE.clear()
+    _PDF_ESTIMATE_CACHE[key] = text
+    return text
+
+
 def _estimate_file_tokens(unit: "Path | FileSlice") -> int:
     """Estimate the prompt-token cost of a file or slice under `_read_files` rules.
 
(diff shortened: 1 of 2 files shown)


safishamsi:

PR #2907, opened 2026-08-20
fix(llm): slice oversized PDFs through the converter (#2906)

### LINKED ISSUE #2906: An oversized PDF loses everything past 20k characters: slicing reads the container's bytes, so PDFs are excluded from it entirely
## Summary

`expand_oversized_files` slices oversized documents so the whole file still reaches the model, but it measures candidates with `path.read_text()` and `read_slice_text` slices the same way — while `llm._file_to_text` routes a PDF through `extract_pdf_text`. Slicing a PDF would therefore index the *container's bytes* rather than its text, so PDFs are excluded from slicing entirely and simply lose everything past `_FILE_CHAR_CAP` (20,000 characters).

This is the half #2900 called out and left alone. Filing it separately because it is independently reproducible and papers are the longest documents anyone points graphify at.

## Reproduce

A one-page PDF with a real text layer, FlateDecode as every real PDF is:

```
pdf bytes=3094   extracted chars=55690   cap=20000
units after slicing: 1
chars reaching the LLM: 20000   LOST: 35690
```

**64% of the document never reaches the semantic pass.** Nothing in it can become a node, and nothing reports the loss: there is no warning and no `_partial_files` marker (that machinery tracks truncated *responses*, not truncated *inputs*).

## Why a PDF hides this better than a text file

A compressed PDF gives no hint of its text length — 3,094 bytes on disk against 55,690 characters of text. Neither the file listing nor the corpus word count makes the gap visible, and the graph that comes out looks complete: a paper whose second half was cut is indistinguishable from one whose second half had nothing extractable in it.

## Suggested fix

One reader that returns the string the prompt will carry, used by **both** the boundary pass and the slice reader, so the offsets a `FileSlice` holds always index the same string the model is shown:

```python
def unit_source_text(path: Path) -> str:
    if path.suffix.lower() not in _CONVERTED_TEXT_SUFFIXES:
        return path.read_text(encoding="utf-8", errors="replace")
    return extract_pdf_text(path)          # memoised
```

Then `is_splittable_text` can answer the question slicing actually asks — "can this be addressed by character offset?" — rather than "is this one of five suffixes". Images and code stay out; PDFs come in.

Worth memoising on `(path, size, mtime)`: slicing asks for the same file once to measure it and then once per slice, and extraction is the expensive part. Keying on size and mtime keeps a paper replaced mid-run from being sliced against stale text.

`extract_pdf_text` lives in `detect`, and `llm` imports `file_slice` (not the reverse), so a lazy import from `file_slice` into `detect` avoids a cycle.

## Related

- #2900 / #2902 — the same drift for `.qmd`, `.html`, `.yaml`, `.skill`; that PR fixed the plain-text half and explicitly deferred this one.
- #2903 / #2904 — the same "raw file vs converted text" confusion in `_estimate_file_tokens`. Independent: that one bites a PDF *under* the cap, this one bites a PDF over it.

## Environment

| | |
|---|---|
| Python | 3.12, Windows 11 |

abhay-codes07:
Fixes #2906.

This is the half #2902 called out and deliberately left undone.

## The bug

`expand_oversized_files` slices oversized documents so the whole file still reaches the model. But it measured candidates with `path.read_text()` and `read_slice_text` sliced the same way, while `llm._file_to_text` routes a PDF through `extract_pdf_text`. Slicing a PDF would therefore have indexed the **container's bytes** instead of its text — so PDFs were excluded from slicing entirely, and simply lost everything past `_FILE_CHAR_CAP`.

One page, real text layer, FlateDecode as every real PDF is:

```
pdf bytes=3094   extracted chars=55690   cap=20000

before: units=1  chars reaching the LLM=20000  LOST=35690   (64% of the paper)
after:  units=3  chars reaching the LLM=55690  LOST=0
```

Nothing reports it — no warning, and no `_partial_files` marker, since that machinery tracks truncated *responses* rather than truncated *inputs*. A paper whose second half was cut is indistinguishable in the graph from one whose second half had nothing extractable in it.

A compressed PDF also hides the gap from every obvious check: 3,094 bytes on disk against 55,690 characters of text. Neither the file listing nor the corpus word count shows anything unusual.

## The change

`unit_source_text` — one reader returning the string the prompt will carry, used by **both** the boundary pass and the slice reader. That is the whole fix: the offsets a `FileSlice` holds and the string the model is shown are now the same string by construction, rather than by two call sites happening to agree.

Plain text is still read straight off disk; only converted types go through a converter, so nothing changes for the formats that already worked.

`is_splittable_text` now answers the question slicing actually asks — *can this be addressed by character offset?* — instead of *is this one of five suffixes*. Images and code stay out; PDFs come in.

`extract_pdf_text` is imported lazily from `detect`, because `llm` imports `file_slice` and the reverse direction at import time would be a cycle.

## The memo

Slicing asks for the same file once to measure it and then once per slice, and extraction is the expensive part — so the reader memoises on `(path, size, mtime_ns)`. A corpus of papers is parsed once per run rather than once per slice. Keyed on size and mtime so a paper replaced mid-run is re-read rather than sliced against stale text (there is a test), and bounded because the entries are whole documents.

## How it sits with the other two

- **#2902** fixed the plain-text half of the same drift (`.qmd`, `.html`, `.yaml`, `.skill`) and explicitly deferred this. No code overlap — that PR only edits the suffix set, this one adds the reader.
- **#2904** fixes `_estimate_file_tokens` reading a PDF's compressed bytes. Independent, and they compose: a PDF over the cap is now a set of `FileSlice`s, and `_estimate_file_tokens` already routes a slice through `read_slice_text`, so its estimate becomes correct here too. A PDF **under** the cap stays a `Path` and is still estimated from bytes — that is what #2904 fixes, and it does not touch this code.

If all three land, the "raw file vs converted text" confusion is gone from the input path. Happy to unify the two memos (`_pdf_text_for_estimate` in `llm` and this one) into a single shared reader afterwards — I kept them separate so each PR stands alone on `v8`.

## Tests

`tests/test_pdf_slicing.py` (13 tests). The fixture builds a real one-page PDF with a genuine text layer and first asserts the *precondition* — small on disk, large in text — since that shape is why file size never revealed the problem.

Then: the PDF is splittable, an oversized one is sliced, the slices tile the extracted text exactly (`"".join(read_slice_text(u)) == unit_source_text(p)` — gap-free, no overlap), no slice exceeds the cap, a slice contains `Section 0` and **not** `%PDF`/`endstream` (the bug in one assertion), and the tail actually reaches the prompt.

Plus the guards: a small PDF still passes through whole, plain-text slicing is unchanged, images and code are still not sliced, a corrupt PDF does not break the pass, a rewritten PDF is re-read, and repeated reads agree.

Reverting `file_slice.py` and keeping the tests fails **6 of 13**. The `unit_source_text` import is deliberately guarded with a fallback so the file still collects on the pre-fix tree — otherwise the whole module would error and the count would tell you nothing about which behaviours are pinned.

## Validation

- Full suite: `15 failed, 4738 passed` -> `15 failed, 4751 passed`. **Identical failure set** — no regressions; the +13 are the new tests.

### DIFF
--- a/graphify/file_slice.py
+++ b/graphify/file_slice.py
@@ -28,6 +28,57 @@ from pathlib import Path
 # docs (.pdf) — those are never sliced.
 _SPLITTABLE_TEXT_SUFFIXES = frozenset({".md", ".mdx", ".markdown", ".txt", ".rst"})
 
+# Document types whose BYTES are not what the model is shown. `llm._file_to_text`
+# routes these through a converter, so a character range has to be taken over the
+# converted text, never over the file. Kept separate from the set above because
+# they are splittable for a different reason and via a different reader.
+_CONVERTED_TEXT_SUFFIXES = frozenset({".pdf"})
+
+
+def _pdf_text(path: Path) -> str:
+    """Extracted text of a PDF — the same string `llm._file_to_text` builds.
+
+    Imported lazily from ``detect`` so this module keeps no import-time
+    dependency on the extraction stack (``llm`` imports *this* module, so the
+    reverse direction would be a cycle).
+    """
+    from graphify.detect import extract_pdf_text
+    return extract_pdf_text(path)
+
+
+# Slicing a PDF means extracting its text, and the slicing pass asks for the same
+# file several times: once to measure it, then once per slice as the prompt is
+# built. Memoised on (path, size, mtime_ns) so a corpus of papers is parsed once
+# rather than once per slice, and so a file rewritten mid-run is re-read instead
+# of served a stale body. Bounded: the entries are whole documents, and a large
+# corpus should not pin all of them in memory.
+_CONVERTED_TEXT_CACHE: "dict[tuple, str]" = {}
+_CONVERTED_TEXT_CACHE_MAX = 64
+
+
+def unit_source_text(path: Path) -> str:
+    """The text a unit contributes to the prompt, whatever its container.
+
+    Plain-text files are read directly; converted types (PDF) go through their
+    converter. Both `expand_oversized_files` and `read_slice_text` use this, so
+    the offsets a slice carries always index the same string the model sees.
+    """
+    if path.suffix.lower() not in _CONVERTED_TEXT_SUFFIXES:
+        return path.read_text(encoding="utf-8", errors="replace")
+    try:
+        st = path.stat()
+        key = (str(path), st.st_size, st.st_mtime_ns)
+    except OSError:
+        return _pdf_text(path)
+    hit = _CONVERTED_TEXT_CACHE.get(key)
+    if hit is not None:
+        return hit
+    text = _pdf_text(path)
+    if len(_CONVERTED_TEXT_CACHE) >= _CONVERTED_TEXT_CACHE_MAX:
+        _CONVERTED_TEXT_CACHE.clear()
+    _CONVERTED_TEXT_CACHE[key] = text
+    return text
+
 # Boundary preferences, strongest first. A Markdown heading (``\n#``) keeps a
 # section with its title; a blank line keeps a paragraph intact; a bare newline
 # avoids cutting mid-line. If none is found in the window we hard-cut.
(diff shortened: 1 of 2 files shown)


safishamsi:

Issue #2911, opened 2026-08-20
C#: generic type arguments are dropped in field and call-site position (captured in property/return/parameter)


brobl2008:
### Summary

The C# extractor walks generic type arguments in **property, return and parameter** position, but drops them in **field** and **call-site** position. The field/property split is the surprising one — they are semantically adjacent, and one works while the other does not.

Version: `0.9.48`. Reproduced with `--code-only` (pure AST, no LLM).

### Reproduction

`Types.cs`

```csharp
public interface IAlpha { }
public interface IBeta { }
public interface IGamma { }
public interface IDelta { }
public interface IEpsilon { }
public interface IZeta { }
public class Box<T> { }
public class Registry { public void Do<T>() { } }
public interface IServiceCollection { }
public static class Ext
{
    public static void AddScoped<TService, TImpl>(this IServiceCollection s) { }
}
```

`Probe.cs`

```csharp
public class Probe
{
    private Box<IAlpha> _field = null!;                    // 1. field
    public Box<IBeta> Prop { get; set; } = null!;          // 2. property
    public Box<IGamma> Ret() => null!;                     // 3. return
    public void Param(Box<IDelta> p) { }                   // 4. parameter
    public void Call(Registry r) => r.Do<IEpsilon>();      // 5. call site
    public void Di(IServiceCollection s) => s.AddScoped<IZeta, Box<IZeta>>();  // 6. DI registration
}
```

```
graphify extract proj --code-only --force --out out
```

### Result

Edges originating from `Probe` or its members:

| position | type arg | edge emitted |
|---|---|---|
| field | `IAlpha` | **none** |
| property | `IBeta` | `references[generic_arg]` |
| return | `IGamma` | `references[generic_arg]` |
| parameter | `IDelta` | `references[generic_arg]` |
| call site | `IEpsilon` | **none** |
| DI registration | `IZeta` | **none** |

Expected: all six emit a `references` edge to the inner type argument, as properties, returns and parameters already do.

### Why it matters

Both gaps erase dependency edges that impact analysis depends on, and they erase them silently — `affected` returns a confident, smaller answer rather than an error or a warning.

**Field position.** Classic constructor injection stores its dependency in a field, so the dependency's concrete type is invisible whenever it appears as a type argument:

```csharp
private readonly IDbContextFactory<SomeDbContext> _factory;   // SomeDbContext: no edge
```

[...]

PR #2913, opened 2026-08-20
fix(csharp): walk generic type arguments in field position

brobl2008:
Fixes the field half of #2911.

## Problem

The `field_declaration` handler reads only the outer type name via `_read_csharp_type_name` and emits a single `references` edge, so the inner argument of `Box<Widget>` is never linked.

Properties, return types and parameters already walk the whole type expression through `_csharp_collect_type_refs`. That makes fields the odd one out — and the `property_declaration` handler's own comment points at exactly the siblings it was mirroring:

> Use `_csharp_collect_type_refs` (like the Java/PHP/Kotlin siblings) so `List<Widget>` yields both the List field ref and the Widget generic_arg ref.

The `tree_sitter_java` `field_declaration` handler immediately below does the same thing correctly. C# fields were simply never converted.

| position | before | after |
|---|---|---|
| property | `references[generic_arg]` | unchanged |
| return | `references[generic_arg]` | unchanged |
| parameter | `references[generic_arg]` | unchanged |
| **field** | **none** | `references[generic_arg]` |

## Why it matters

Two very common shapes lose their edge:

```csharp
private readonly IDbContextFactory<SomeContext> _factory;  // SomeContext: no edge
private readonly Mock<IThing> _mock;                       // IThing: no edge
```

Classic constructor injection stores its dependency in a **field**, so this is the same blind spot #2829 removed from primary constructors — just relocated to where the dependency is kept. Mock-based unit tests lose their subject across an entire suite.

The failure is silent. `affected` returns a smaller, confident answer rather than an error or a warning, which is the same reason #2829 was worth fixing.

## Fix

Route the field handler through `_csharp_collect_type_refs`, passing the in-scope type parameters as the skip set, mirroring the property handler exactly. The outer type still emits `context="field"`; arguments emit `context="generic_arg"`. The existing `csharp_field_types` receiver registration is untouched.

**Bonus:** this also stops fields fabricating nodes for predefined types. `private string _s` previously created a `string` node, because `_read_csharp_type_name` returns builtins while `_csharp_collect_type_refs` returns early on `predefined_type`. That aligns fields with the builtin-not-fabricated behaviour added for primary constructors in 1eb356cd.

## Verification

Eight regression tests in `tests/test_csharp_field_generic_args.py`, covering both directions — the argument is linked, a field and a property of the same type now agree, nested arguments resolve, and type parameters and builtins are still never fabricated.

| | |
|---|---|
| New tests on unpatched HEAD | **5 of 8 fail** |
| New tests with this patch | **8 of 8 pass** |
| C# suites (`-k csharp`) | 108 passed, 5 skipped |
| Full suite, baseline | 30 failed, 4,575 passed |
| Full suite, patched | 25 failed, 4,580 passed |

A set-difference of the failing test ids between the two full runs shows **nothing newly broken** — the delta is exactly the five new tests.

One caveat stated plainly: `test_labeling.py::test_label_communities_batches_when_over_batch_size` is flaky independently of this change. It fails intermittently on unpatched HEAD too (confirmed over repeated runs), so please don't read it either way.

## Scope

Deliberately field-only. #2911 also covers call-site type arguments (`services.AddScoped<IThing, Thing>()`), but that lives in the call-resolution path and overlaps #2676, which strips call-site type arguments to repair the callee match. The two want different things from the same tokens — one discards them, the other needs them emitted — so that half is better as its own change once #2676 settles. Happy to write it.

### DIFF
--- a/graphify/extractors/engine.py
+++ b/graphify/extractors/engine.py
@@ -3714,13 +3714,29 @@ def _extract_generic(
                             if name_node is not None:
                                 fields[_read_text(name_node, source)] = type_name
                 line = node.start_point[0] + 1
-                metadata = {"ref_token": type_name}
-                if qualified:
-                    metadata["qualified"] = True
-                if qualifier:
-                    metadata["ref_qualifier"] = qualifier
-                add_edge(parent_class_nid, ensure_named_node(type_name, line),
-                         "references", line, context="field", metadata=metadata)
+                # Walk the whole type expression rather than only its outer name, so
+                # `Box<Widget>` yields the Box field ref AND the Widget generic_arg ref.
+                # Reading just the outer name left every generic argument in field
+                # position unlinked -- `IDbContextFactory<SomeContext>` lost SomeContext,
+                # and `Mock<IThing>` lost IThing across entire test suites. The C#
+                # property_declaration handler below and the tree_sitter_java
+                # field_declaration handler beside it already do exactly this; C# fields
+                # were the odd one out.
+                refs: list[tuple[str, str, bool, str]] = []
+                _csharp_collect_type_refs(
+                    type_node, source, False, refs, csharp_type_params
+                )
+                for ref_name, role, ref_qualified, ref_qualifier in refs:
+                    ctx = "generic_arg" if role == "generic_arg" else "field"
+                    target_nid = ensure_named_node(ref_name, line)
+                    if target_nid != parent_class_nid:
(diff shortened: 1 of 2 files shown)


safishamsi:
Shipped in v0.9.49 via authorship-preserving cherry-pick so you keep contributor-graph credit. Thanks @brobl2008! Release: https://github.com/Graphify-Labs/graphify/releases/tag/v0.9.49

Issue #1506, opened 2026-06-28
Obsidian existing-vault output can overwrite notes and .obsidian graph settings


moeflam:
## Summary

`--obsidian-dir` is documented in the closed pre-existing-vault issue as safe for an existing Obsidian vault, but the current exporter writes generated node notes directly into the target directory and writes `.obsidian/graph.json` there as well. If the target is an existing vault root, that can overwrite human notes and vault graph settings.

This is especially risky because issue #31 says:

> This writes a `graphify/` folder inside your vault ... Your existing vault structure is untouched — graphify only adds files, never modifies existing ones.

Current `v8` source behavior appears different: `graphify export obsidian --dir /path/to/vault` passes the directory directly to `to_obsidian()`, and `to_obsidian()` writes `Alpha.md`, `_COMMUNITY_*.md`, `.obsidian/graph.json`, etc. at that exact path.

## Minimal reproduction

Against current `v8` checkout:

```bash
cd graphify
uv run python - <<'PY'
from pathlib import Path
import tempfile
import networkx as nx
from graphify.export import to_obsidian

with tempfile.TemporaryDirectory() as d:
    out = Path(d)
    existing = out / 'Alpha.md'
    existing.write_text('# Human Alpha\n\nkeep me\n', encoding='utf-8')

    G = nx.Graph()
    G.add_node('a', label='Alpha', file_type='code', source_file='src/a.py')
    G.add_node('b', label='Beta', file_type='code', source_file='src/b.py')
    G.add_edge('a', 'b', relation='calls', confidence='EXTRACTED')

    n = to_obsidian(G, {0: ['a', 'b']}, str(out), community_labels={0: 'Main Community'})

    print('notes_written', n)
    print('top_level', sorted(p.name for p in out.iterdir()))
    print('alpha_first_line', existing.read_text(encoding='utf-8').splitlines()[0])
    print('alpha_contains_human_text', 'keep me' in existing.read_text(encoding='utf-8'))
PY
```

Observed:

```text
notes_written 3
top_level ['.obsidian', 'Alpha.md', 'Beta.md', '_COMMUNITY_Main Community.md']
alpha_first_line ---
alpha_contains_human_text False
```

So an existing `Alpha.md` was replaced by generated content.

Relevant tests currently pass, but they cover dangling members and filename caps, not coexistence/collision safety:

```text
uv run pytest tests/test_obsidian_dangling_member.py tests/test_obsidian_filename_cap.py -q
7 passed in 0.51s
[...]

PR #2914, opened 2026-08-20
fix(export): adopt a pre-manifest vault's own notes instead of orphaning them (#2863)

### LINKED ISSUE #2863: Obsidian vault ownership manifest (#1506) has no migration path: pre-manifest vaults end up holding two generations of notes
## Summary

The Obsidian ownership manifest added in #1506 has no migration path for a vault that was
created by an **earlier** graphify. After upgrading, every pre-existing note is treated as
a user file, so a re-export writes fresh notes *next to* the stale ones and the vault ends
up holding both generations.

## Repro

1. Build a vault with a version older than #1506 (no `.graphify_obsidian_manifest.json`).
2. Upgrade graphify, change some source docs, re-run the update flow.
3. `to_obsidian(...)` into the same directory.

Observed: `WARNING: skipped 219 pre-existing file(s) graphify did not create` and the
directory now contains **401 `.md` files for a 317-node graph** — the 219 untouched notes
are graphify's own output from the previous run (they carry graphify frontmatter:
`tags: [graphify/document, graphify/EXTRACTED, community/...]`), plus 119 newly written
notes, plus community notes from both generations.

## Why it matters

The user cannot tell which generation a note belongs to, and stale notes for nodes that no
longer exist keep showing up in graph view and search. The warning does say "export into an
empty directory", but nothing in the update flow does that, so a routine `--update` quietly
degrades the vault. Deleting the vault by hand works but loses the `.obsidian/` config the
manifest was introduced to protect.

## Suggested fix (any one of these would close the gap)

1. **Adopt on read**: when the manifest is missing, treat a pre-existing `.md` whose
   frontmatter carries a `graphify/*` tag as graphify-owned, and write the manifest from
   that set on the first post-upgrade run.
2. **`--adopt-existing` flag** for an explicit one-time migration.
3. **Prune stale generations**: after a successful export, delete manifest-owned notes that
   are no longer in the graph (nodes removed or renamed), which the current flow also leaves
   behind even with the manifest present.

Related: with (3) in place, a vault that has drifted can be reconciled without deleting the
whole directory.

abhay-codes07:
Fixes #2863.

Thanks @ — the diagnosis was exact, including that the leftover notes carry graphify's own frontmatter, which is what makes them recoverable.

## The bug

#1506 added `.graphify_obsidian_manifest.json` so a re-export can update the notes graphify owns while refusing to touch the user's. A vault created by an **earlier** graphify has no manifest, so `_owned` starts empty — and every note graphify itself wrote last time reads as a user file.

Reproduced on a 3-node graph: build a vault, drop the manifest to stand in for a pre-#1506 vault, rename two nodes, re-export.

```
before:  6 notes for a 3-node graph
         Alpha.md, Beta.md, "Beta renamed.md", Gamma.md, "Gamma renamed.md", community
         WARNING: skipped 3 pre-existing file(s) graphify did not create ...

after:   exactly the current generation, no stale notes
         Alpha.md, "Beta renamed.md", "Gamma renamed.md", _COMMUNITY_Community 0.md
```

The warning is the part I'd flag as worst: it names files graphify wrote itself and tells the user it didn't. Someone cleaning up by hand has no way to tell the two generations apart, because both carry graphify frontmatter.

## The change

A one-time migration, run only when no manifest exists, adopting a file solely on graphify's own signature. Two shapes, because the notes are not alike:

- **Node notes** carry YAML frontmatter with a tag in the `graphify/` namespace. Only that block is inspected, so a user's note that merely mentions graphify in prose is not adopted — there is a test for exactly that.
- **Community notes carry no frontmatter at all** (I assumed they did, and the first version of this fix silently left every one behind until the reproduction showed it). They are matched on graphify's filename prefix **and** the Dataview query it writes into every one. Requiring both means a user's own `_COMMUNITY_notes.md` is not adopted on the name alone.

Scoped to top-level `*.md` — the only files identifiable from their content.

## `.obsidian/graph.json` is deliberately left unowned

graphify writes one, but so does Obsidian, and with no manifest there is no way to tell whose it is. Guessing wrong costs the user their vault configuration, which is far worse than one file staying unowned.

So a pre-manifest vault still reports one skipped file. That warning is now *correct*: for that file graphify genuinely cannot confirm it was the author. Worth knowing rather than being surprised by it.

## Tests

`tests/test_obsidian_vault_migration.py` (16 tests): no second generation after a re-export, stale notes for renamed nodes gone, no false "did not create" warning, and the manifest written so the migration happens once.

Then the safety half, which is the part that matters: a user's own `Alpha.md` survives untouched, a note merely mentioning graphify is not adopted, a user's `_COMMUNITY_*.md` without the query marker is not adopted, `.obsidian/graph.json` is preserved byte-for-byte, a vault that already has a manifest is unaffected, and a fresh directory still works.

Reverting `export.py` and keeping the tests fails 4 (6 more skip, since the detector does not exist pre-fix).

## Validation

Windows 11, Python 3.12, branched off `b2cd362` (0.9.48).

- Full suite: `16 failed, 4835 passed` -> `16 failed, 4851 passed`. **Identical failure set** — no regressions; the +16 are the new tests. Both runs were taken on this machine against this commit, rather than reusing an older baseline.

### DIFF
--- a/graphify/export.py
+++ b/graphify/export.py
@@ -544,6 +544,73 @@ def _cap_filename(s: str, limit: int = 200) -> str:
     return f"{truncated}_{digest}"
 
 
+# A frontmatter tag entry in graphify's own namespace, e.g. "  - graphify/document".
+_GRAPHIFY_TAG_RE = re.compile(r"^\s*-\s+graphify/\S")
+
+# Frontmatter sits at the very top of a note; reading this much is enough to see
+# the whole block without pulling a large note into memory.
+_NOTE_FRONTMATTER_PROBE_BYTES = 4096
+
+# Community notes carry no frontmatter; graphify identifies its own by the
+# Dataview query it writes into every one of them.
+_COMMUNITY_QUERY_MARKER = "FROM #community/"
+
+
+def _is_graphify_note(path: Path) -> bool:
+    """Whether a vault note carries graphify's own frontmatter signature.
+
+    Every note graphify writes opens with a YAML frontmatter block tagging it in
+    the ``graphify/`` namespace::
+
+        ---
+        source_file: "d0.md"
+        tags:
+          - graphify/document
+          - graphify/EXTRACTED
+        ---
+
+    Only that block is inspected, and only a tag entry inside it counts — a
+    user's note that merely mentions graphify in its prose is not adopted.
+
+    Community overview notes are recognised separately: they carry no
+    frontmatter at all, so they are identified by graphify's own filename prefix
+    together with the Dataview query it writes into the body. Requiring both
+    keeps a user's own ``_COMMUNITY_*.md`` from being adopted on the name alone.
+    """
+    try:
+        with path.open("r", encoding="utf-8", errors="replace") as fh:
+            head = fh.read(_NOTE_FRONTMATTER_PROBE_BYTES)
+    except OSError:
+        return False
+    if path.name.startswith(_COMMUNITY_PREFIX) and _COMMUNITY_QUERY_MARKER in head:
+        return True
+    if not head.startswith("---"):
+        return False
+    for line in head.splitlines()[1:]:
+        if line.strip() == "---":
+            return False  # frontmatter closed without a graphify tag
+        if _GRAPHIFY_TAG_RE.match(line):
+            return True
+    return False
+
+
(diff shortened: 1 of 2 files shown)


safishamsi:

PR #2928, opened 2026-08-21
fix(extract): scope fresh semantic results to dispatched files (#2926)

### LINKED ISSUE #2926: Incremental merge: stray output misattributed to non-dispatched files silently deletes their prior graph contributions
## Summary

During incremental extraction, LLM output whose `source_file` is misattributed to a file that was **not** dispatched silently deletes that file's entire prior contribution from `graph.json`. The semantic cache is protected against this; the graph build is not.

## Environment

graphifyy 0.9.16, Python 3.12 (WSL2 Ubuntu 24.04), gemini backend, markdown-only corpus (26 docs).

## Mechanism

1. Extracting file A, the model also emits a few stray nodes/edges attributed to other files (observed: one file's extraction produced strays attributed to 7 other files, plus nonexistent forward-reference paths like `src/foo.ts`).
2. `_save_semantic_cache(..., allowed_source_files=uncached_paths)` correctly discards them from the cache (`semantic cache skipped out-of-scope source_file` warning) — the cache stays clean.
3. But in `cli.py`, the same unfiltered `fresh` result is extended into `sem_result` and reaches `build_merge()`.
4. `build_merge()` derives its replace-set from the source_files present in the new chunks — so a 2-node stray fragment for file B **replaces B's entire prior contribution** (observed: a file whose cache entry held 20 nodes was reduced to 1 node in graph.json).
5. B's manifest entry still says unchanged, so B is never re-dispatched and its clean cache entry is never re-applied. The loss is permanent until a full rebuild.
6. Strays attributed to nonexistent files become phantom nodes carried forward forever — never in any changed set, never matched by the delete pass (observed: 24 phantoms accumulated).

## Reproduction sketch

Corpus of N cross-referencing markdown docs. Force one file to re-extract (touch content). If the model attributes any output to a second, unchanged file, that file's graph contribution is replaced by the fragment. Diff unique node ids in `cache/semantic/*.json` against `graph.json` to observe the divergence.

## Suggested fix

Apply the same `allowed_source_files` scope filter to `fresh` **before** extending `sem_result` (mirror of the cache-write guard). We run 0.9.16 with exactly that local patch: a run that previously clobbered 7 files holds every node, and the filter logs 9–33 dropped strays per routine run on our corpus — misattribution is common enough that the graph-side guard matters.

SinghAman21:
Closes #2926

## Summary

During incremental extraction, LLM output whose `source_file` pointed at a file
that was **not** dispatched silently deleted that file's entire prior
contribution from `graph.json`. The semantic cache has been guarded against this
since #1757 (`allowed_source_files`); the graph build was not.

## Root cause

On an incremental run only the changed subset is dispatched, and unchanged files
are never read back from the semantic cache — their graph contribution survives
precisely because they are *absent* from the new chunks. A stray fragment the
model attributes to such a file puts it back into the replace-set:

1. `_save_semantic_cache(..., allowed_source_files=uncached_paths)` correctly
   discards the strays from the cache (`skipped out-of-scope source_file`),
2. but the same unfiltered `fresh` result was extended into `sem_result`,
3. so `build_merge()` saw the file among `new_chunks`, dropped its prior
   nodes/edges per tier, and merged in only the fragment,
4. the manifest still said unchanged → never re-dispatched, never repaired.
   Loss was permanent until a full rebuild.

Strays attributed to nonexistent paths (`src/foo.ts`) became phantom nodes no
delete pass ever matched.

## Fix

- **`cache.py` — new `scope_semantic_result()`**: filters a result dict in place
  to the dispatched-file allowlist, using the exact path normalization as
  `save_semantic_cache`'s guard (so an item one keeps, the other can't skip).
  Mirrors #1916 by also dropping surviving edges/hyperedges that reference a
  dropped node id — except duplicate-attribution ids, which stay reference-safe.
- **`cli.py`**: applies the filter to `fresh` immediately after extraction,
  before both the cache save and the `sem_result` extension. This also closes a
  residual hole where strays from chunk A attributed to a file dispatched in
  chunk B polluted B's cache entry at the final save. Drops are logged:
  `[graphify extract] dropped N out-of-scope item(s) attributed to M file(s) ...`

## Testing

- 4 unit tests for `scope_semantic_result`: group dropping (incl. nonexistent
  paths), absolute/relative identity equivalence, dangling-edge pruning with
  duplicate-attribution protection, unscoped no-op.
- End-to-end CLI regression test reproducing the issue: seed run gives
  `OTHER.md` a two-node contribution; next run changes only `README.md` and the
  stubbed model answers with README's node plus strays for `OTHER.md` and
  `src/foo.ts`. Asserts `OTHER.md`'s original nodes survive, no stray/phantom
  enters the graph, and the drop is reported. Verified red without the fix,
  green with it.
- Full suite: 3281 passed (only pre-existing, unrelated `test_ollama_retry_cap`
  failure, present on the clean tree).

### DIFF
--- a/graphify/cache.py
+++ b/graphify/cache.py
@@ -1559,3 +1559,121 @@ def save_semantic_cache(
             stacklevel=2,
         )
     return saved
+
+
+def scope_semantic_result(
+    result: dict,
+    root: Path = Path("."),
+    allowed_source_files: "Iterable[str | Path] | None" = None,
+) -> tuple[set[str], int]:
+    """Scope an extraction result in place to the files actually dispatched (#2926).
+
+    Graph-side mirror of the ``allowed_source_files`` write-guard in
+    :func:`save_semantic_cache` (#1757). A model can attribute stray
+    nodes/edges to a corpus file that was not part of the current extraction
+    batch; :func:`build_merge` derives its replace-set from the source_files
+    present in the new chunks, so such a stray fragment would REPLACE that
+    file's entire prior contribution in graph.json — while its manifest entry
+    still says unchanged, so no later incremental run re-dispatches it and the
+    loss is permanent until a full rebuild.
+
+    Items whose ``source_file`` resolves outside ``allowed_source_files`` are
+    dropped from ``result``'s ``nodes`` / ``edges`` / ``hyperedges`` lists
+    (mutated in place); items without a ``source_file`` pass through. An edge
+    or hyperedge that survives the scope filter but references a dropped node
+    id is dropped too (#1916 mirror), unless that id is also defined by a kept
+    node (duplicate attribution must not be over-pruned).
+
+    Path matching uses the same normalization as :func:`save_semantic_cache`
+    (relative against ``root``, walked-path identity), so an item this function
+    keeps can never still hit the save's out-of-scope skip, and vice versa.
+
+    Returns ``(dropped_source_files, dropped_item_count)`` for logging;
+    ``dropped_source_files`` holds the normalized ``source_file`` strings of
+    every group that had at least one item removed.
+    """
+    if allowed_source_files is None:
+        return set(), 0
+
+    root_walked = _normalize_path(Path(os.path.abspath(root)))
+    root_resolved = _normalize_path(Path(root).resolve())
+
+    def _identity(value: str) -> Path:
+        path = Path(value)
+        if not path.is_absolute():
+            path = root_walked / path
+        elif root_walked != root_resolved:
+            normalized = _normalize_path(Path(os.path.abspath(path)))
+            try:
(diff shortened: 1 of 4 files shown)


SinghAman21:
These are threshold-based warnings, not actual regressions. save_semantic_cache has high Ca because ~50 of its 55 callers are tests - only cli.py calls it in production. scope_semantic_result has 1 prod caller + 4 test sites. Both functions are cohesive and appropriately coupled to internal cache infrastructure. No refactor needed.

safishamsi:
Shipped in v0.9.49 via authorship-preserving cherry-pick so you keep contributor-graph credit. Thanks @SinghAman21! Release: https://github.com/Graphify-Labs/graphify/releases/tag/v0.9.49

Issue #2852, opened 2026-08-18
post-commit hook silently exits 0 under `uv tool install`: never rebuilds, no error


larrymandras:
## Summary

The `graphify hook install` post-commit hook cannot locate graphify when graphify was
installed via `uv tool install graphifyy`, and it fails **silently** (`exit 0`). The hook
appears installed, commits succeed normally, and nothing is ever rebuilt. Graphs go stale
indefinitely with no signal.

I found this on Windows after two repos' graphs had drifted 127h and 175h stale despite
the post-commit hook being installed in both. It turned out to affect **30 repos** on this
machine, plus `~/.git-templates/hooks/post-commit`, so every repo `git init`'d since the
template was written was born with a dead hook.

`uv tool install` is the install method the README recommends, so I suspect this is
widespread rather than specific to my setup.

## Environment

- Windows 11, Git for Windows (Git-Bash `sh` runs the hook)
- graphifyy 0.9.46, installed via `uv tool install --force "graphifyy[sql]"`
- Hook installed via `graphify hook install`

### 1. The `*.exe` guard never fires on Git-Bash (cosmetic, already fixed in newer hooks)

```sh
GRAPHIFY_BIN=$(command -v graphify 2>/dev/null)
case "$GRAPHIFY_BIN" in
    *.exe) _SHEBANG="" ;;
    *)     _SHEBANG=$(head -1 "$GRAPHIFY_BIN" | sed 's/^#![[:space:]]*//') ;;
esac
```

Git-Bash's `command -v graphify` returns the launcher **with no extension**:

```
$ command -v graphify
/c/Users/USER/.local/bin/graphify
```

(`shutil.which` reports `C:\Users\USER\.local\bin\graphify.EXE`, but the shell strips it.)

So `*.exe)` never matches, and `head -1` reads a PE binary. Visible symptom:

```
.git/hooks/post-commit: line 23: warning: command substitution: ignored null byte in input
```

A newer hook revision already fixes this by gating on a leading `#!` — one of my repos had
it. **But that fix removes the only visible symptom while leaving cause 2 intact**, which
makes the silent failure harder to find, not easier.

### 2. The interpreter fallback cannot see a uv-installed graphify (fatal)

```sh
if [ -z "$GRAPHIFY_PYTHON" ]; then
    if command -v python3 >/dev/null 2>&1 && python3 -c "import graphify" 2>/dev/null; then
        GRAPHIFY_PYTHON="python3"
[...]

Issue #2629, opened 2026-08-11
Interpreter probe fails on pipx installs: shebang with an argument is parsed as a file path


hpstr5000:
Both the generated `post-commit` hook and the skill's own interpreter probes read the shebang of the `graphify` launcher and use it as a path. pipx writes `#!/path/to/venv/bin/python -E` — **with an argument** — so the string is not a path, and every probe falls through to a `python3` that does not have graphify installed.

Result on my machine: every commit printed

```
[graphify hook] could not locate a Python with graphify installed.
```

for weeks, so the graph silently stopped updating. Since it is one line of noise after each commit, it went unnoticed for a long time.

## Reproduce

```bash
$ which graphify
/Users/me/.local/bin/graphify
$ head -1 "$(which graphify)" | tr -d #!
/Users/me/.local/pipx/venvs/graphifyy/bin/python -E     # <- note the argument
```

Now the probe from `SKILL.md` step 1:

```bash
_SHEBANG=$(head -1 "$GRAPHIFY_BIN" | tr -d "#!")
case "$_SHEBANG" in
    *[!a-zA-Z0-9/_.@-]*) ;;                     # <- space matches, probe is skipped entirely
    *) "$_SHEBANG" -c "import graphify" && PYTHON="$_SHEBANG" ;;
esac
```

The character guard rejects the string because of the space, so the verification never runs and the probe never fires.

The `## Interpreter guard for subcommands` block has the same parse **without** the `import graphify` verification, so it writes the unverified `python3` fallback into `graphify-out/.graphify_python`. Every later step reads that pinned path and fails as if the install were broken.

## Suggested fix

Strip the interpreter argument before using the shebang as a path, and verify the candidate before pinning it:

```bash
_SHEBANG=$(head -1 "$GRAPHIFY_BIN" | tr -d "#!" | cut -d" " -f1)
"$_SHEBANG" -c "import graphify" 2>/dev/null && PYTHON="$_SHEBANG"
```

With `cut -d" " -f1` the path resolves and the import check passes on a stock pipx install. The same two changes apply to the generated `post-commit` hook, which carries the identical probe.

I patched this locally in `SKILL.md` and confirmed the hook now reports `launching background rebuild`. Happy to open a PR if useful.

Environment: macOS 15 (Darwin 25.5.0), graphify 0.9.12, installed via pipx.

cdtatarevich:
Second pipx shim format that defeats the same probe, plus a follow-on break after the probe succeeds.

[...]

PR #2929, opened 2026-08-21
fix(hooks): resolve the uv tool venv when pin and launcher probes miss (#2852)

santhiprakash:
## Problem

`uv tool install` is the README's recommended install method, but the post-commit hook cannot find that install when the install-time pin is dead, and then dies with no rebuild (#2852). On a machine where 30 repos plus `~/.git-templates` carried the hook, graphs drifted 127h/175h stale with the hook "installed" and commits succeeding.

Resolution chain under `uv tool install` with a dead `_PINNED` (git-template hooks, a pin rejected by the allowlist, or a venv relocated by an upgrade):

1. `.graphify_python` — absent in repos that never built a graph.
2. launcher on PATH — a **binary trampoline** on Windows: `command -v graphify` (Git-Bash) returns it without the `.exe` suffix, so the `*.exe` guard misses and there is **no shebang to parse at all**. On POSIX the shim's shebang does resolve, which is why this hides on Linux/macOS.
3. last resort `python3`/`python` — no ambient python can import graphify from an isolated uv venv → warning + `exit 0`, commit succeeds, nothing rebuilt.

Two gaps in `graphify/hooks.py` `_PYTHON_DETECT` (embedded in both the post-commit and post-checkout scripts):

- the shebang read parses whatever the first bytes are — binary trampoline bytes only get caught by the allowlist *after* being parsed (#2852 cause 1);
- no probe looks at uv tool environments at all (#2852 cause 2), so the chain has no way back to a healthy venv install.

## Fix

- **uv tool-env scan**, tried after the launcher probe and before the `python3`/`python` last resort: iterate `${UV_TOOL_DIR:-}`, `$HOME/.local/share/uv/tools`, `$HOME/AppData/Roaming/uv/tools` and probe `*/bin/python` and `*/Scripts/python.exe` with the existing `_GFY_PROBE`. A tool env is adopted only if its python passes the probe, so a co-installed tool without graphify never satisfies it (glob order is not trusted). This is the direction suggested in the issue, generalized to not hardcode the dist name.
- **`#!` gate on the shebang read**: the launcher's head bytes are only parsed as a shebang when the file actually starts with `#!` — a trampoline (or any non-script launcher) is rejected structurally instead of by the allowlist catching the parsed garbage.

The loud `could not locate a Python with graphify installed` stderr warning (restored on `v8` after 0.9.46's bare `exit 0`) is kept as the terminal failure — it stays out of this change's way and the scan makes it far rarer.

## Test

- `test_uv_tool_env_rescues_hook_when_pin_and_launcher_fail` — #2852's machine reproduced on POSIX (dead pin, no `.graphify_python`, binary trampoline launcher on PATH, ambient pythons that cannot import graphify): the emitted `_PYTHON_DETECT` run under real `sh` must resolve the uv venv python, walking past a sibling tool env whose python lacks graphify. Fails on pre-fix code.
- `test_uv_tool_env_honors_uv_tool_dir_and_windows_layout` — `UV_TOOL_DIR` override + the Windows `Scripts\python.exe` layout. Fails pre-fix.
- `test_shebang_parse_requires_leading_hash_bang` — a non-script launcher whose first line merely names a working python must not hijack resolution (the misparse that used to feed trampoline bytes into the chain). Fails pre-fix.
- `test_uv_tool_env_without_graphify_still_fails_loudly` — the scan never adopts a venv without graphify; the chain still ends in the loud warning, never a silent adopt.
- Existing suite intact: `tests/test_hooks.py` 96 passed; `uv run --frozen pytest tests/ -q` matches the pre-PR baseline (the `test_skillgen.py` audit-coverage and `test_ollama_retry_cap.py` failures reproduce on a clean `origin/v8` tree in this checkout and are untouched by this diff; CI's `fetch-depth: 0` covers the former).

Sandbox end-to-end against a real `uv tool install` of the local tree: with the pin blanked and the launcher replaced by a binary blob, pre-fix commit → warning, log untouched; post-fix commit → hook resolves `~/.local/share/uv/tools/graphifyy/bin/python` and the detached rebuild runs (`[graphify hook] 1 file(s) changed - rebuilding graph...`).

Related: #2629 (pipx `python -E` shebang arg) is the same probe chain but a different parse failure and is **not** addressed here. Builds on the #2126/#2166 allowlist lineage in this block.

### DIFF
--- a/graphify/hooks.py
+++ b/graphify/hooks.py
@@ -67,10 +67,17 @@ if [ -z "$GRAPHIFY_PYTHON" ]; then
         # POSIX launcher: parse the shebang. head -c + tr strip NUL bytes first —
         # when the launcher is a Windows binary reached without its .exe suffix,
         # a raw `head -1` reads binary into the command substitution and the
-        # shell warns about ignored null bytes on every commit.
+        # shell warns about ignored null bytes on every commit. Gate on a
+        # leading '#!': a launcher can also be a binary trampoline with no
+        # shebang at all (uv tool installs on Windows), and its bytes must
+        # never reach the shebang parse (#2852).
         case "$GRAPHIFY_BIN" in
-            *.exe) _SHEBANG="" ;;
-            *)     _SHEBANG=$(head -c 256 "$GRAPHIFY_BIN" 2>/dev/null | tr -d '\\000' | head -n 1 | sed 's/^#![[:space:]]*//') ;;
+            *.exe) _GFY_HEAD="" ;;
+            *)     _GFY_HEAD=$(head -c 256 "$GRAPHIFY_BIN" 2>/dev/null | tr -d '\\000') ;;
+        esac
+        case "$_GFY_HEAD" in
+            '#!'*) _SHEBANG=$(printf '%s\\n' "$_GFY_HEAD" | head -n 1 | sed 's/^#![[:space:]]*//') ;;
+            *)     _SHEBANG="" ;;
         esac
         case "$_SHEBANG" in
             */env\\ *) GRAPHIFY_PYTHON="${_SHEBANG#*/env }" ;;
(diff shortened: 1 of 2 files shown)


santhiprakash:
Re: the advisory on the uv tool-env scan executing candidates from an environment-controlled directory — the scan's trust scope matches the probes that already run in `_PYTHON_DETECT`: probe 3 executes the `python.exe` beside the PATH-resolved `graphify` launcher and the interpreter parsed from that launcher's shebang, and the last-resort probe executes `python3`/`python` straight off `PATH`. `UV_TOOL_DIR` is no more privileged than `PATH`, and the default scan roots are under `/home/ubuntu/.local/share/uv/tools`, where an attacker with write access could already replace the PATH-visible launcher that probe 3 executes. No new trust boundary is crossed; adoption additionally requires passing the graphify import probe.

safishamsi:
Shipped in v0.9.49 via authorship-preserving cherry-pick so you keep contributor-graph credit. Thanks @santhiprakash! Release: https://github.com/Graphify-Labs/graphify/releases/tag/v0.9.49

PR #2930, opened 2026-08-21
fix(csharp): walk generic type arguments at call sites (#2911)

santhiprakash:
## Problem

The C# `invocation_expression` handler captured the callee name but never walked the call site's type-argument list. Result: generic arguments at call sites were silently dropped from the graph.

```csharp
r.Do<IEpsilon>()                                 // IEpsilon: no edge
services.AddScoped<IZeta, Box<IZeta>>()          // IZeta, IZeta, Box: no edges
StaticHolder.Invoke<IThing>()                    // IThing: no edge
```

The graph `affected` query returns a confident, smaller answer instead of an error -- a real dependency edge is gone. Reproduces with the issue's two-file repro via `graphify update . --no-cluster` (pure AST, deterministic, no LLM).

## Why this fix

The property/return/parameter branches above the call-site branch already close the same bug in adjacent positions -- the `property_declaration` branch (around line 3746) is the direct analogue: it walks the declared type with `_csharp_collect_type_refs` and emits a `references[generic_arg]` edge per type argument. The call-site branch just never got the parallel treatment.

This fix is intentionally bounded to the call-site case:
- The `field_declaration` case is the parallel fix in #2913 (in flight from the issue reporter).
- Properties, returns, parameters, base lists, primary constructor parameters, and inherited interfaces were already correct.
- The two PRs are complementary: this one closes the second of two halves the issue splits.

## Fix

In the C# branch of the `walk_calls` closure (in `graphify/extractors/engine.py`, after the existing callee/member-receiver extraction), look for a `type_argument_list` on the call's `function` node:

- `recv.Do<T>()` -- the function is a `member_access_expression` whose `name` is a `generic_name` carrying the type-arg list.
- `Foo<Bar>()` -- the function is a `generic_name` directly.
- The fallback path (raw-text `name` scan) is left alone; it never sees the structured type-arg list and is the same path that drops them for the callee, so reconstructing them there would be lossy.

For each type argument, walk it with the same `_csharp_collect_type_refs(generic=True, skip=...)` helper the other C# branches use, then emit a `references` edge with `context="generic_arg"` from the caller (the method node). The `skip` set is the class's and method's in-scope type parameters (the same `_csharp_type_parameters_in_scope` walk), so a `Holder<T>.Use(r => r.Do<T>())` doesn't fabricate a node for the type parameter `T`.

## Test

Adds `tests/test_csharp_call_site_generic_args.py` with six cases that fail on pre-fix code and pass with the fix:

- `r.Do<IThing>()` -- one-arg member call (the case the issue lists as position 5).
- `s.AddScoped<IService, IImpl>()` -- two-arg call (the Microsoft.Extensions.DependencyInjection shape the issue calls out as position 6).
- `r.Do<Box<IThing>>()` -- nested type argument at a call site.
- `r.Do()` -- call without type arguments is unchanged (regression guard).
- `Holder<T>.Use(r => r.Do<T>())` -- type parameter used as the call-site type argument must not fabricate a `T` node.
- The issue's two-file repro end-to-end: all six positions from the issue's table produce their expected edges (the field case is closed by #2913; this PR is responsible for the call-site half).

Each test asserts the edge EXISTS, not its count -- absence is the bug; counts are an implementation detail.

## Verification

- `.venv/bin/python -m pytest tests/test_csharp_call_site_generic_args.py` -- 6 passed.
- `.venv/bin/python -m pytest tests/test_languages.py tests/test_csharp_*.py tests/test_extractors_registry.py tests/test_language_resolvers.py tests/test_affected_member_seed.py tests/test_affected_cli.py` -- 488 passed, 13 skipped, 0 failed.
- End-to-end on the issue's two-file repro via `graphify update . --no-cluster`: positions 2-6 from the issue's table each emit their expected `generic_arg` reference (position 1, the field case, is the parallel #2913's responsibility).

## Relationship to #2913

#2913 (the issue reporter's PR) fixes the field-position case (`private Box<IAlpha> _field;` losing `IAlpha`) by walking the field's declared type in the `field_declaration` handler. This PR fixes the call-site case (`r.Do<T>()` and `s.AddScoped<T1, T2>()` losing the type arguments) by walking the call's type-argument list in the `invocation_expression` handler. They are complementary, non-overlapping, and target the two halves the issue splits into. Both can merge independently; together they close the bug.

### DIFF
--- a/graphify/extractors/engine.py
+++ b/graphify/extractors/engine.py
@@ -5085,6 +5085,60 @@ def _extract_generic(
                                 else:
                                     callee_name = raw
                                 break
+                # C#: emit a `references[generic_arg]` edge for every type
+                # argument at the call site (`recv.Do<T>()`, the
+                # `services.AddScoped<ISvc, Impl>()` DI shape, static
+                # `Foo<IBar>()`). The property/return/parameter branches
+                # already walk their declared type for the same reason; the
+                # call-site branch didn't, so the type arguments never
+                # became nodes and dependency edges were silently erased
+                # (#2911). The C# class_declaration's field_declaration and
+                # property_declaration branches above are the direct
+                # analogue. The call-site function carries its type-arg list
+                # either as a `type_argument_list` child on a `generic_name`
+                # (static call) or as the same child on the
+                # `member_access_expression`'s `name` `generic_name` (member
+                # call); the fallback path uses raw text and never sees the
+                # structured type-arg list. The class declaration's
+                # field_declaration case is closed by the parallel fix in
+                # #2913; this branch covers what that PR deliberately left
+                # out.
+                if fn_node is not None:
+                    call_tal = None
+                    if fn_node.type == "member_access_expression":
+                        ma_name = fn_node.child_by_field_name("name")
+                        if ma_name is not None and ma_name.type == "generic_name":
+                            for tal_child in ma_name.children:
+                                if tal_child.type == "type_argument_list":
+                                    call_tal = tal_child
+                                    break
+                    elif fn_node.type == "generic_name":
+                        for tal_child in fn_node.children:
+                            if tal_child.type == "type_argument_list":
+                                call_tal = tal_child
+                                break
+                    if call_tal is not None:
+                        call_type_params = _csharp_type_parameters_in_scope(node, source)
+                        call_line = node.start_point[0] + 1
+                        for call_arg in call_tal.children:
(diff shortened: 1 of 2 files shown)


safishamsi:
Shipped in v0.9.49 via authorship-preserving cherry-pick so you keep contributor-graph credit. Thanks @santhiprakash! Release: https://github.com/Graphify-Labs/graphify/releases/tag/v0.9.49

PR #2931, opened 2026-08-21
fix(cache): retry zero-node semantic results

### LINKED ISSUE #2927: Semantic results with edges but no nodes are cached, permanently freezing the file out of the graph
## Summary

When a dispatched file's semantic result contains edges but **no nodes** (model omission), the empty entry is written to `cache/semantic/` keyed by content hash — so every subsequent run is a cache hit and the file is permanently frozen out of the graph. The console warning says "a re-run will retry them", but the cache makes that false.

## Environment

graphifyy 0.9.16, Python 3.12 (WSL2 Ubuntu 24.04), gemini backend, markdown corpus.

## Mechanism

1. On large batches the model routinely omits files from its response — observed warning rates of `6/20`, `5/8`, `8/9`, `9/15 dispatched file(s) produced no nodes`.
2. A file with *no* output at all gets no cache entry and correctly re-queues (#933 handling).
3. But a file that received only **edges** (nodes list empty) *does* get a `by_file` entry in `save_semantic_cache` — cached as `{"nodes": [], "edges": [...]}`.
4. Every later run: content hash unchanged → cache hit → never re-dispatched. Observed: a corpus's most-referenced hub document sat at 0 nodes indefinitely; every edge pointing into it dangled and was pruned.
5. It cannot be healed from the outside either: other files' extractions that mention it are discarded by the `allowed_source_files` guard, and evicting the cache entry alone is not sufficient because `_detect_incremental` still reports the file unchanged (the manifest entry must be removed too).

## Suggested fix

In `save_semantic_cache`, skip writing any per-file entry whose `nodes` list is empty — the file then stays uncached and genuinely re-dispatches on the next run, making the existing "a re-run will retry them" behavior true. We run 0.9.16 with exactly that patch (a `warnings.warn` on the skip); unit-verified that the edges-but-no-nodes shape no longer caches while normal results cache unchanged.

hopstreax:
Closes #2927

## Summary

Semantic extraction results containing edges but no nodes can currently be
cached as successful results. Because the cache is keyed by content hash, the
file is then treated as unchanged on subsequent incremental runs and is never
redispatched.

This can permanently leave the file without its semantic nodes.

## Root Cause

During incremental extraction:

1. A dispatched file produces an edge-only result:
   `nodes=[]`, `edges=[...]`.
2. `save_semantic_cache()` writes the result to the semantic cache.
3. The manifest records the current semantic hash because edges count as output.
4. On the next run, the cache entry is treated as a hit.
5. The file is not redispatched even though its semantic extraction omitted its
   nodes.
6. The file remains without semantic nodes until it is modified or a forced
   rebuild is performed.

## Fix

A semantic extraction is now considered valid for caching and manifest stamping
when it contains:

- at least one node, or
- at least one hyperedge.

Edge-only results are therefore treated as incomplete and remain retryable.

The fix:

- prevents edge-only semantic results from being written to the cache,
- treats existing zero-node/zero-hyperedge cache entries as cache misses,
- prevents edge-only results from stamping the semantic manifest,
- detects and heals already-poisoned manifest state during incremental
  extraction,
- preserves legitimate hyperedge-only documents.

## Tests

- edge-only results not being cached,
- zero-node/zero-hyperedge cache entries being rejected,
- normal node-containing results remaining cached,
- hyperedge-only results remaining valid,
- edge-only semantic results remaining retryable,
- stale poisoned manifest entries being re-dispatched,
- hyperedge-only documents not being incorrectly re-queued.

All #2927-focused and adjacent tests pass.

### DIFF
--- a/graphify/cache.py
+++ b/graphify/cache.py
@@ -992,6 +992,18 @@ def load_cached(path: Path, root: Path = Path("."), kind: str = "ast",
         # across chunks without losing the truncated one (it stays partial).
         if not allow_partial and isinstance(result, dict) and result.get("partial"):
             return None
+        # A semantic entry with zero nodes and zero hyperedges is invalid (#2927):
+        # an edge-only or empty result (e.g. LLM omitted entities for the file)
+        # is not a valid standalone extraction. Treating it as a cache MISS
+        # ensures the file is re-dispatched and retried (#933/#1666).
+        if (
+            not allow_partial
+            and kind.startswith("semantic")
+            and isinstance(result, dict)
+            and not result.get("nodes")
+            and not result.get("hyperedges")
+        ):
+            return None
         if (
             kind.startswith("semantic")
             and isinstance(result, dict)
@@ -1543,6 +1555,11 @@ def save_semantic_cache(
             )
             if is_partial:
                 result = {**result, "partial": True}
+            # A semantic extraction with zero nodes and zero hyperedges is not a valid
+            # standalone extraction (#2927): edge-only or empty results must not be
+            # cached, so that subsequent runs can re-dispatch and retry the file (#933/#1666).
+            if not is_partial and not (result.get("nodes") or result.get("hyperedges")):
+                continue
             save_cached(cache_path, result, root, kind=kind, cache_root=cache_root,
                         prompt=prompt, prompt_file=prompt_file)
             saved += 1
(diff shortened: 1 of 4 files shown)


safishamsi:
Shipped in v0.9.49 via authorship-preserving cherry-pick so you keep contributor-graph credit. Thanks @hopstreax! Release: https://github.com/Graphify-Labs/graphify/releases/tag/v0.9.49

PR #2933, opened 2026-08-21
fix: send reasoning_effort for reasoning-tuned local models via ollama backend

### LINKED ISSUE #2932: ollama backend never sets reasoning_effort, so local reasoning models (nemotron, deepseek-r1, qwq) can blow --api-timeout on real extraction chunks
## What happened

Benchmarking Claude/GLM-5.2/`nemotron-3-super:120b-a12b-q4_K_M` (via `--backend ollama`, served locally through Ollama) on identical corpora, Nemotron's extraction was consistently much sparser than the other two backends, and on a larger multi-file corpus one file (`README.md`) was dropped entirely after repeated timeout+bisect retries (5 files → 2 → 1, each hitting the client's 600s `--api-timeout`, with nothing left to bisect at 1 file).

No JSON parsing errors were involved — this is not the #2882 narration/parsing issue, which is already fixed and working correctly.

## Root cause (confirmed empirically, not guessed)

Direct `POST /v1/chat/completions` against `nemotron-3-super:120b-a12b-q4_K_M` on the same Ollama host, same prompt (real source file + the extraction instructions), varying only `reasoning_effort`:

| Call | `content` length | `reasoning` length | `completion_tokens` |
|---|---|---|---|
| default (`reasoning_effort` omitted) | 3,299 chars | 17,442 chars | 5,234 |
| `reasoning_effort: "high"` | 2,092 chars | 9,154 chars | 2,807 |
| `reasoning_effort: "low"` | 3,463 chars | 17,543 chars | 5,244 |

Findings:
- `choices[0].message` has a separate `reasoning` field (Ollama's OpenAI-compat naming) alongside `content`. `content` itself is never empty or sparse in any of the three calls — each held valid, parseable JSON with genuine extraction — so this isn't the #2882 recovery path, and it isn't a case of the answer hiding in the wrong field either.
- Ollama's OpenAI-compat endpoint **does respect `reasoning_effort`** for this model. `"high"` roughly halves both the reasoning length and total completion tokens compared to omitting it or passing `"low"`.
- `graphify/llm.py`'s `BACKENDS["ollama"]` config never sets a `reasoning_effort` key — only `BACKENDS["gemini"]` does (`"reasoning_effort": "low"`). So for every `--backend ollama` call, `cfg.get("reasoning_effort")` is `None` and the field is never sent, regardless of which model is actually being served.

Because the reasoning narration shares the same request's wall-clock and the client's `--api-timeout` (default 600s), a model that reasons at length before answering can burn most or all of that budget on narration alone. On a real extraction chunk (not the isolated single-file diagnostic above), that pushed a 5-file chunk into the adaptive bisect-and-retry path, and it still failed at a single file with no smaller unit to fall back to — the whole file was dropped from that run.

## Why this isn't ollama-backend-wide

Ollama serves many non-reasoning models (the backend's own default is `qwen2.5-coder:7b`), so a blanket default would be wrong for most `--backend ollama` users. The bug is specifically that reasoning-tuned local models get no `reasoning_effort` treatment at all, whereas the same family is already recognized elsewhere in the codebase — `_THINK_BLOCK_RE`'s own comment names `nemotron, deepseek-r1, qwq` as the models known to narrate a `<think>` block.

## Repro

1. Serve a reasoning-tuned model locally, e.g. `ollama pull nemotron-3-super:120b-a12b-q4_K_M`.
2. `graphify extract <a corpus with 3+ semantic files> --backend ollama --model nemotron-3-super:120b-a12b-q4_K_M --max-concurrency 1`
3. On a large enough chunk, watch it bisect and potentially drop a file, with `GRAPH_REPORT.md` showing far fewer nodes/edges than the same corpus run through another backend (Gemini, GLM-5.2, Claude) gets.

## Suggested fix

Opening a PR shortly: resolve `reasoning_effort` per-call for the `ollama` backend (env var override via `GRAPHIFY_OLLAMA_REASONING_EFFORT`, falling back to `"high"` for models matching the same reasoning-family list `_THINK_BLOCK_RE` already recognizes, and `None`/unchanged otherwise) rather than leaving it unset for every model served through Ollama.

AutogenousEngine:
Fixes #2932.

## Problem

The `ollama` backend config never set `reasoning_effort` at all -- only `gemini` did (`"reasoning_effort": "low"`). Benchmarking `nemotron-3-super:120b` against Claude/GLM-5.2 on identical corpora, it produced consistently sparser extraction, and on a larger corpus dropped a file entirely after exhausting the bisect-and-retry timeout path. No JSON parsing was involved (`.content` was never hollow) -- see #2932 for the full empirical writeup, including raw `/v1/chat/completions` measurements showing `reasoning_effort: "high"` roughly halves both reasoning length and completion tokens for this model relative to omitting it or `"low"`.

Since the model's chain-of-thought narration (landing in a separate `message.reasoning` field on Ollama's OpenAI-compat endpoint) shares the same request's wall-clock and the client's `--api-timeout`, an unset `reasoning_effort` on a verbose local reasoning model can burn most of that budget on narration alone -- which is what pushed the real multi-file chunk into repeated bisection and eventual file loss.

## Fix

Adds `_resolve_reasoning_effort(cfg, model, backend)`, used at both call sites that previously read `cfg.get("reasoning_effort")` directly. Precedence:

1. `GRAPHIFY_OLLAMA_REASONING_EFFORT` env var (ollama only) -- explicit override; accepts `"none"/"omit"/"default"` to force omission even for a recognized reasoning model, mirroring `GRAPHIFY_LLM_TEMPERATURE`'s existing convention.
2. The backend config's own default (`gemini`, unchanged).
3. `"high"` for `ollama` models matching the same reasoning-model families `_THINK_BLOCK_RE`'s comment already names (`nemotron`, `deepseek-r1`, `qwq`) -- reusing recognition the codebase already committed to elsewhere, not inventing a new category.
4. `None` otherwise -- every other backend/model is unaffected, including non-reasoning models served through ollama (the backend's own default model is `qwen2.5-coder:7b`).

## Testing

- New `tests/test_ollama_reasoning_effort.py` (8 tests): family matching (including a provider-prefixed `nvidia/nemotron-3-super` form, mirroring `_model_requires_default_temperature`'s existing prefix-stripping), the default-for-reasoning-models / none-for-others split, non-ollama backends being unaffected by the env var, env override winning over the default (including for non-reasoning models -- an explicit choice always wins), the `none`/`omit`/`default` force-omit sentinels, and a backend-config default (if one is ever added for another ollama-family entry) taking precedence over the reasoning-model fallback.
- `uv run pytest tests/test_ollama_reasoning_effort.py tests/test_llm_backends.py tests/test_llm_parser.py tests/test_llm_parser_reasoning.py -q` -- 133 passed.
- `uv run pytest tests/ -q` -- same 18 pre-existing failures as an unmodified `v8` checkout at this depth (shallow-clone git-history gaps in `test_skillgen`, environment-dependent `test_ollama*`/`test_ollama_retry_cap` failures) confirmed by running the identical command against `git stash`-ed clean tree before this diff; this change adds zero new failures.

Happy to adjust the default (e.g. a different effort level, or gating it behind an opt-in flag instead of a name-matched default) if you'd rather ship it more conservatively.

### DIFF
--- a/graphify/llm.py
+++ b/graphify/llm.py
@@ -375,6 +375,63 @@ def _resolve_temperature(default: float | None, model: str = "") -> float | None
     return default
 
 
+# Same families _THINK_BLOCK_RE strips a <think>/<reasoning> block for: local
+# reasoning-tuned models that narrate a long chain of thought before the
+# answer. Served through Ollama's OpenAI-compat endpoint, that narration lands
+# in a separate `message.reasoning` field (confirmed live against
+# nemotron-3-super via raw /v1/chat/completions calls) — `.content` itself is
+# never hollow, so this is not a parsing bug. The narration still costs real
+# wall-clock and shares the request's token budget, and on a real multi-file
+# extraction chunk it can consume enough of both that the client's
+# --api-timeout trips, the adaptive bisect-and-retry kicks in, and a chunk can
+# still fail at a single file with nothing left to bisect.
+_OLLAMA_REASONING_MODEL_FAMILIES = ("nemotron", "deepseek-r1", "qwq")
+
+
+def _ollama_model_is_reasoning(model: str) -> bool:
+    """True for a local model known to narrate a chain of thought via Ollama."""
+    m = (model or "").lower().rsplit("/", 1)[-1]
+    return any(fam in m for fam in _OLLAMA_REASONING_MODEL_FAMILIES)
+
+
+def _resolve_reasoning_effort(cfg: dict, model: str, backend: str) -> str | None:
+    """Resolve the `reasoning_effort` value to send, if any.
+
+    Precedence:
+      1. GRAPHIFY_OLLAMA_REASONING_EFFORT env var, if `backend == "ollama"` and
+         it's set — explicit user override. Any backend-recognised value
+         ("low"/"medium"/"high"), or the literal "none"/"omit"/"default"
+         (case-insensitive, mirrors GRAPHIFY_LLM_TEMPERATURE's convention) to
+         force omission even for a recognised reasoning model.
+      2. The backend config's own default (only `gemini` sets one today).
+      3. For `ollama` with a recognised local reasoning model
+         (_ollama_model_is_reasoning) and no config default: "high". Measured
+         empirically against nemotron-3-super (raw API, three otherwise-identical
+         calls): omitting the field or passing "low" both produced ~17.5k chars
+         of reasoning and ~5.2k completion tokens; "high" cut that to ~9.2k
+         chars / ~2.8k tokens — roughly half the time/token budget per call,
+         which is the difference between finishing under a real extraction
+         chunk's timeout and not.
+      4. Otherwise None (omit the parameter — unchanged default behaviour for
+         every other backend/model).
+
+    Ollama serves a heterogeneous zoo of models, most of which are not
+    reasoning-tuned and were never observed with this field, so the default in
+    step 3 is scoped to named reasoning families rather than applied backend-wide.
+    """
+    if backend == "ollama":
+        env_raw = os.environ.get("GRAPHIFY_OLLAMA_REASONING_EFFORT", "").strip()
+        if env_raw:
+            if env_raw.lower() in ("none", "omit", "default"):
+                return None
+            return env_raw
+    if cfg.get("reasoning_effort"):
+        return cfg["reasoning_effort"]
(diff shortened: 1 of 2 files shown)


AutogenousEngine:
Closing — decided to maintain this as a fix on our own fork rather than pursue upstream inclusion. The underlying bug report in #2932 stands if it's useful to anyone else hitting the same ollama-backend reasoning_effort/timeout issue.

Issue #1920, opened 2026-07-15
hyperedge-only doc output never stamped in manifest — perpetual re-extraction


Yashasm18:
## Bug

`_stamped_manifest_files()` (added in #1897 to fix zero-node docs being stamped incorrectly) decides whether a semantic document "produced output" by checking only the `nodes` and `edges` collections of `sem_result`. It never checks `hyperedges`.

```python
sem_extracted: set[Path] = set()
for coll in ("nodes", "edges"):
    for item in sem_result.get(coll, []):
        sf = item.get("source_file", "")
        if sf:
            sem_extracted.add(_resolve(sf))
```

If a chunk's only output for a given document is a hyperedge (a coherent group of 3+ existing nodes/concepts spanning that doc — the extraction skill prompt explicitly supports this: "if 3 or more nodes clearly participate together in a shared concept... add a hyperedge"), that document is left out of `sem_extracted`.

This means:
- `_response_is_hollow()` in `llm.py` correctly treats `hyperedges` as valid output (`not nodes and not edges and not hyperedges`), so the chunk is NOT flagged as a failure/truncation.
- But `_stamped_manifest_files()` disagrees — it excludes the doc from the manifest anyway.
- `semantic_hash` never gets stamped for that file.
- `detect_incremental()` re-queues the file as "changed" on every subsequent `graphify extract` run, even though it was successfully extracted.

## Observed with

- graphifyy 0.9.16/0.9.17, code reading (not yet reproduced against a live LLM backend — filing from source-level analysis, repro below is a direct unit-level trigger)

## Reproduction

```python
from graphify.cli import _stamped_manifest_files

files_by_type = {"document": [str(doc_path)]}
sem_result = {
    "nodes": [],
    "edges": [],
    "hyperedges": [{"id": "h1", "label": "...", "nodes": ["a", "b", "c"], "source_file": "doc.md"}],
}
out = _stamped_manifest_files(files_by_type, sem_result, root)
# out["document"] == [] — doc.md is dropped even though it produced a hyperedge
```

## Expected behavior

A document whose only extraction output is a hyperedge should be treated the same as one with a node or edge: included in `sem_extracted`, stamped in the manifest, and not re-queued on the next incremental run.

## Suggested fix

Add `"hyperedges"` to the iterated collections in `_stamped_manifest_files`, matching `_response_is_hollow`'s definition of "produced output":

```python
for coll in ("nodes", "edges", "hyperedges"):
[...]

PR #2936, opened 2026-08-21
fix(cache): skip caching edges-but-no-nodes semantic results so model…

SinghAman21:
…-omitted files re-dispatch instead of freezing out of the graph (#2927)

On large batches the model sometimes omits a file from its response but still emits edges attributed to it. We cached that {"nodes": [], "edges": [...]} result keyed by content hash, so every later run was a cache hit and the file never got extracted again — its edges dangled and were pruned, and the "a re-run will retry them" warning wasn't actually true.
What changed
- save_semantic_cache skips writing per-file entries that have edges but zero nodes (with a warning), so the file genuinely re-dispatches next run
- load_cached treats already-poisoned entries (from older versions) as a miss, so existing caches self-heal on the next complete extraction
- _stamped_manifest_files no longer stamps an edges-only file as up-to-date — otherwise warm incremental runs would skip it even with the cache fixed
Hyperedge-only results are untouched and still cache normally (#1920). Multi-chunk merges are unaffected since the check runs on the merged result.

### DIFF
--- a/graphify/cache.py
+++ b/graphify/cache.py
@@ -992,6 +992,22 @@ def load_cached(path: Path, root: Path = Path("."), kind: str = "ast",
         # across chunks without losing the truncated one (it stays partial).
         if not allow_partial and isinstance(result, dict) and result.get("partial"):
             return None
+        # An entry carrying edges but ZERO nodes is a model-omission artifact
+        # (#2927) that older versions cached keyed by content hash, replaying
+        # the empty node set forever — the file sat at 0 nodes while every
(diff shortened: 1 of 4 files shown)


safishamsi:
Thanks @SinghAman21 — we went with #2931 for the same issue (#2927), which adds a manifest-healing pass to re-queue files already poisoned by a prior stamp; it shipped in v0.9.49. Your diagnosis of the edges-but-no-nodes case was right and matched. Closing as superseded. Release: https://github.com/Graphify-Labs/graphify/releases/tag/v0.9.49

Issue #1581, opened 2026-07-01
Case-folded cross-file resolution collides distinct identifiers (Path/PATH) → false super-hubs in case-sensitive languages


sheik-hiiobd:
## Summary

Cross-file name resolution matches identifiers **case-insensitively for every language**. In case-sensitive languages that's wrong: `pathlib.Path` (ubiquitous) resolves to an unrelated node labeled `PATH`, turning a single shell-script variable into the **#1 god-node** of a real repo. It inflates god-node rankings, `affected` blast-radius, and community assignment.

## Repro (2 files, AST-only, no key)

```bash
# run.sh
export PATH=/usr/local/bin:$PATH
```
```python
# mod.py
from pathlib import Path
def load(p: Path) -> Path:
    return Path(p)
def other():
    return load(Path("x"))
```
```
$ graphify <dir> --no-cluster
```

Edges produced:

```
load()  -> PATH  [references]   # the `p: Path` / `-> Path` annotations
load()  -> PATH  [calls]        # Path(p)
other() -> PATH  [calls]        # Path("x")
```

`Path` (a stdlib class, not in the corpus) resolves to `PATH` — the shell variable in `run.sh`. There is no `Path` node, so this is *resolution*, not a dedup-merge.

## Evidence at scale

On a real ~3,700-node Python codebase:

- one node labeled `PATH`, sourced from a shell script's `export PATH=...`
- **266 incoming edges** (224 `references` + 41 `calls` + 1 `defines`) → the #1 god-node
- **zero** files define a `PATH` constant; every incoming edge is a file doing `from pathlib import Path`
- `graphify affected PATH` returns 266 false positives; the true blast radius of the shell var is ~0

## Root cause

Identifier matching case-folds. The cross-file resolver keys its candidate index and its lookups on `.lower()`:

- `graphify/extract.py:15282` — `key = normalised.lower()` (global index build)
- `graphify/extract.py:15337` — `candidates = global_label_to_nids.get(callee.lower(), [])`

and the dedup key case-folds too:

- `graphify/build.py:633` — `_norm_label(...)` → `label.casefold()`

So `Path`, `PATH`, and `path` collapse to one key. A single case-folded candidate resolves directly (`extract.py:15356`), even with no import evidence.

## The principle

[...]

PR #2939, opened 2026-08-21
fix(extract): stop a built-in base class from inheriting across languages (#2812)

### LINKED ISSUE #2812: PHP 'extends Exception' (bare built-in) resolves onto an unrelated same-named TS class — cross-language inherits collision
## Summary

A class's bare `extends`/`inherits` base-class name is resolved against a single global name→id map with no per-language scoping and no built-in-base-class exclusion list, so a PHP class that `extends Exception` (PHP's global built-in) fuses onto an unrelated `Exception` class defined in a TypeScript file elsewhere in the same corpus.

## Reproduction

Corpus (2 files, different languages):

```ts
// app/src/exceptions/exception.ts
export class Exception extends Error {
  constructor(message: string) { super(message); }
}
```

```php
// packages/foo/src/Exceptions/FooApiException.php
namespace Foo\Exceptions;

class FooApiException extends \Exception  // or bare `extends Exception`
{
}
```

Expected: `FooApiException` inherits from PHP's built-in `\Exception` — either no edge (base not in corpus) or an edge to a synthesized "PHP built-in" stub, distinct per language.

Actual (graphifyy 0.9.45): `FooApiException` gets an `inherits [EXTRACTED, confidence_score=1.0]` edge pointing directly at the **TypeScript** `Exception` class node.

## Real-world impact

On a ~6,000-file PHP+TypeScript monorepo (Laravel API + React/TS frontend), this produced **12 phantom `inherits` edges** from unrelated PHP/TS exception classes across 3 separate first-party Composer packages onto one TS class:

- `packages/*/src/Exceptions/InvalidRecipientException.php`
- `packages/*/src/Exceptions/SendGridApiException.php`
- `packages/*/src/Exceptions/TelenorApiException.php`
- `packages/*/src/Exceptions/InvalidPhoneNumberException.php`
- `packages/*/src/Exceptions/TemplateNotFoundException.php`
- `packages/*/src/Exceptions/WhatsAppApiException.php`

(plus 6 legitimate same-file/same-language JS `Error` subclasses correctly landing on the same node, which is what made the phantom edges hard to spot — the node's degree looked plausible.)

This inflated the TS `Exception` node into a top-10 "god node" and generated a false "surprising cross-community bridge" suggested-question (`Why does Exception connect <TS file> to <PHP admin controller>?`) — tracing every hop confirmed no real code path exists; the only connector is the shared bare identifier.

## Root cause (traced in `graphify/extractors/base.py` + `graphify/extract.py`)

- `_LANGUAGE_BUILTIN_GLOBALS` in `extractors/base.py` filters common built-in *constructor/coercion call* targets (`String`, `Error`, Python builtins, Swift stdlib, etc.) to stop them becoming god nodes for `calls` edges — but it has **no entry for `Exception`/`Throwable`/`RuntimeException`-style base-exception names** in any language, and this list is not consulted by the supertype (`inherits`/`implements`/`extends`) resolver at all (`_SUPERTYPE_RELATIONS` in `extract.py`).
- Per-language namespace-aware supertype resolvers exist for C# (`CsharpNameResolver`) and were added for PHP same-namespace collisions (the `Page`/`Filament\Pages\Page` fix), but PHP's resolver appears scoped to *disambiguating same-language namespace collisions* only — it does not stop a bare PHP built-in class name (`Exception`, no `use` import, no namespace prefix) from being matched against the *global cross-language* lowercase-name map that also contains a same-named class from an entirely different file extension.
- This is the `inherits`/`implements` analogue of #598 (member-call cross-file name collision, fixed in 0.5.5 for `calls` edges only) — the same bare-identifier-collision failure mode, just on the supertype-resolution code path instead of the call-resolution code path, so #598's fix doesn't cover it.

## Suggested fix

1. Scope supertype (`inherits`/`implements`/`extends`) bare-name resolution to same-language candidates first (mirroring how `calls` resolution already special-cases languages), falling back to cross-language only when explicitly justified (there basically never is a legitimate cross-language `inherits` edge).
2. Add a per-language built-in-exception/base-class exclusion set (PHP `Exception`, `Throwable`, `RuntimeException`, `LogicException`, `Error`; Java `Exception`, `RuntimeException`; Python `Exception`, `BaseException`; JS `Error`) consulted by the supertype resolver, not just the call-site god-node filter, so an unqualified reference to a language's own built-in base class never resolves onto an unrelated corpus symbol.

## Environment

- Package: `graphifyy 0.9.45`
- Corpus: Laravel (PHP 8) API + React/TypeScript frontend monorepo, ~28k AST nodes
- OS: macOS (darwin arm64)

ousamabenyounes:
Fixes #2812.

## The bug

`class FooApiException extends \Exception` in PHP mints a sourceless `Exception`
stub, because the base class is not defined in the corpus. `_rewire_unique_stub_nodes`
then binds that stub to the one same-labelled real definition found *anywhere* in
the corpus — and in a PHP + TypeScript monorepo that definition is an unrelated
TypeScript `class Exception`. The reporter's ~6k-file repo got 12 phantom
`inherits` edges from first-party PHP exception classes onto one TS class, which
pushed it into the top-10 god nodes and produced a false cross-community bridge
question.

## The fix

The rewire now refuses a supertype edge when the *referring file's* language names
its own built-in throwable and the resolved target lives in a different language
family. Three deliberate choices:

- **Per edge, not per stub.** One sourceless `Exception` stub collects referrers
  from every language that names it. Deciding on the stub's referrer families as a
  set would let a single TypeScript referrer re-open the cross-language bind for
  the PHP ones. The check reads `edge["source_file"]`, and sits next to the
  existing C#-scoped carve-out in the same loop.
- **Built-in throwables only, not a blanket family gate on the type path.** A corpus
  really can declare `BookStore` in one language and subclass it from another —
  `test_extract_rewires_unique_inheritance_stub_to_real_definition` pins exactly
  that, and this PR leaves it passing untouched. But `Exception` named from a PHP
  file is always PHP's own, so a same-named class in another language cannot be
  what it refers to.
- **Case-folded for case-insensitive languages.** PHP resolves class names
  case-insensitively, so `extends \exception` names the same built-in
  (mirrors the existing `real_by_label` / `real_by_label_ci` pair, #1581).

`_LANGUAGE_BUILTIN_GLOBALS` was not reused: it is a flat set consulted at call
sites and has no per-language scoping, which is what this path needs.

## Relationship to #2365

#2365 proposes hoisting the interop family gate to *every* unique match, which
would also close this issue but changes existing behaviour — it has to rewrite
`test_extract_rewires_unique_inheritance_stub_to_real_definition`, whose
Python↔C# rewire is currently intended. This PR is additive and takes the
narrower position that only a language's own built-ins can never be foreign.
If you prefer the blanket gate, #2365 supersedes this and I'll close it.

RED — new tests against unmodified `v8` (`b2cd362`), production file reverted:

```
$ git checkout origin/v8 -- graphify/extract.py
$ uv run --frozen pytest tests/test_extract.py -q -k rewire --tb=line
FAILED tests/test_extract.py::test_rewire_does_not_bind_supertype_stub_across_language
FAILED tests/test_extract.py::test_rewire_builtin_supertype_guard_folds_case_insensitive_languages
FAILED tests/test_extract.py::test_rewire_builtin_supertype_guard_is_per_edge_not_per_stub
3 failed, 8 passed, 191 deselected, 1 warning in 0.54s

$ uv run --frozen pytest tests/test_multilang.py::test_php_builtin_base_class_never_inherits_from_cross_language_class -q --tb=line
FAILED tests/test_multilang.py::test_php_builtin_base_class_never_inherits_from_cross_language_class
1 failed, 1 warning in 0.31s
```

The failure is the reported leak verbatim:

```
E   AssertionError: PHP supertype leaked cross-language:
    {'source': 'packages_fooapiexception_fooapiexception',
     'target': 'app_exception_exception', 'relation': 'inherits', ...}
```

GREEN — same tests with the fix:

```
$ uv run --frozen pytest tests/test_extract.py -q -k rewire
11 passed, 191 deselected, 1 warning in 0.56s

$ uv run --frozen pytest tests/test_multilang.py::test_php_builtin_base_class_never_inherits_from_cross_language_class -q
1 passed, 1 warning in 0.26s
```

Full suite, unchanged against the `v8` baseline (the 4 failures are pre-existing
in my local environment — 3 `test_ollama` backend-detection tests read env vars I
have set, and `test_collect_files_skips_hidden` — all 4 fail identically on clean
`v8`):

```
$ uv run --frozen pytest tests/ -q
4 failed, 4882 passed, 11 skipped, 4 warnings in 131.39s
```

`ruff check` clean on the touched files, and the five `tools.skillgen`
validators (`--check`, `--audit-coverage`, `--schema-singleton`,
`--monolith-roundtrip`, `--always-on-roundtrip`) all pass.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@ Full release notes with details on each version: [GitHub Releases](https://githu
 
 ## 0.9.48 (2026-08-20)
 
+- Fix: a class that extends one of its own language's built-in exception types no longer inherits from an unrelated same-named class in another language, so a PHP `class FooApiException extends \Exception` stops fusing onto a TypeScript `Exception` class; the unique-stub rewire now refuses that cross-language match and leaves the built-in base on its own external stub, while legitimate cross-language rewires are unchanged (#2812, thanks @moeen-basra).
 - Fix: a control character in a node label or id no longer aborts the whole export; the GraphML and Obsidian exporters scrub only the characters those formats forbid (tab, newline, and non-ASCII letters are preserved), and `graph.json` and its byte-identity round-trip are untouched (#2897, thanks @abhay-codes07).
 - Fix: `graphify update` / `label` / `cluster-only` no longer leave a large graph without a `graph.html`; the aggregated community view renders instead of raising, a failed render preserves the previous file, and a missing `graph.html` is regenerated on the no-change fast path without reclustering (#2853, thanks @oleksii-tumanov).
 - Feature: `graphify extract --no-dedup` skips the fuzzy near-duplicate merge on build and incremental merge, for operators who would rather keep distinct symbols that fuzzy-matched; exact-id uniqueness is unaffected and the flag arms the shrink guard so a surprising node drop is refused loudly (#2881, thanks @rajarshidattapy).
--- a/graphify/extract.py
+++ b/graphify/extract.py
@@ -2238,6 +2238,62 @@ def _lang_family(source_file: object) -> str | None:
     return _LANG_FAMILY_BY_EXT.get(Path(str(source_file)).suffix.lower())
 
 
+# A language's own built-in throwable hierarchy, keyed by the interop family of
+# the file that names it. `class FooApiException extends \Exception` in PHP means
+# PHP's global `Exception`, so a same-named class defined in a file of ANOTHER
+# family cannot be what it refers to (#2812). Scoped to built-in throwables on
+# purpose: they are the names every language ships and every corpus subclasses,
+# while a name a corpus commonly defines itself would suppress a real supertype
+# edge. `_LANGUAGE_BUILTIN_GLOBALS` covers the adjacent call-target case and is
+# deliberately separate — a flat set consulted at call sites, neither per-family
+# nor consulted by supertype resolution.
+_LANGUAGE_BUILTIN_BASE_CLASSES: dict[str, frozenset[str]] = {
+    "php": frozenset({
+        "Throwable", "Exception", "ErrorException", "Error", "TypeError",
+        "ValueError", "ArgumentCountError", "ArithmeticError",
+        "DivisionByZeroError", "RuntimeException", "LogicException",
+        "InvalidArgumentException", "DomainException", "LengthException",
+        "OutOfRangeException", "OutOfBoundsException", "RangeException",
+        "OverflowException", "UnderflowException", "UnexpectedValueException",
+        "BadFunctionCallException", "BadMethodCallException", "JsonException",
+    }),
+    "jvm": frozenset({
+        "Throwable", "Exception", "RuntimeException", "Error",
+        "IllegalArgumentException", "IllegalStateException",
+        "UnsupportedOperationException", "IndexOutOfBoundsException",
+        "NullPointerException", "IOException",
+    }),
+    "python": frozenset({
+        "BaseException", "Exception", "ValueError", "TypeError", "KeyError",
+        "IndexError", "RuntimeError", "NotImplementedError", "AttributeError",
+        "OSError", "IOError", "StopIteration", "Warning", "UserWarning",
+        "DeprecationWarning",
+    }),
+    "jsts": frozenset({
+        "Error", "TypeError", "RangeError", "SyntaxError", "ReferenceError",
+        "EvalError", "URIError", "AggregateError",
+    }),
+    "dotnet": frozenset({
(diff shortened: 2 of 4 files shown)


safishamsi:
Shipped in v0.9.49 via authorship-preserving cherry-pick so you keep contributor-graph credit. Thanks @ousamabenyounes! Release: https://github.com/Graphify-Labs/graphify/releases/tag/v0.9.49

Issue #1531, opened 2026-06-29
TypeScript path aliases ignore fallback targets after the first


oleksii-tumanov:
### Context

Reproduced on `v8` at `12193a8`. TypeScript `compilerOptions.paths` values are ordered fallback lists, but Graphify keeps only `targets[0]`. If that substitution is missing while a later one exists, the import edges target the missing path and are dropped during build.

Related to #147. The open #927 handles wildcard pattern matching; this case is only about target fallback order.

### Reproducer

```json
{
  "compilerOptions": {
    "baseUrl": ".",
    "paths": {
      "$lib/*": ["generated/*", "src/lib/*"]
    }
  }
}
```

Only `src/lib/types/type-helpers.ts` exists:

```ts
// src/routes/page.ts
import type { Helper } from '$lib/types/type-helpers'
```

### Current behavior

Relevant `extract()` output, path-sanitized:

```json
{
  "source": "src_routes_page",
  "target": "<project>_generated_types_type_helpers",
  "relation": "imports_from",
  "context": "import",
  "confidence": "EXTRACTED",
  "source_file": "src/routes/page.ts",
  "source_location": "L1"
}
```

JS/TS files bypass the AST cache, so this extraction does not produce a `graphify-out/cache/ast/...` entry.

### Expected

Try each configured target in order until one resolves. The example should target `src_lib_types_type_helpers`; if both targets exist, the first should still win.

safishamsi:
Fixed on `v8` (2133539). You're right — `compilerOptions.paths` values are ordered fallback lists and graphify was keeping only the first target.

The alias map now stores all targets in declared order, and resolution tries each in turn, returning the first whose candidate file exists on disk (matching `tsc`), falling back to the first candidate when none exist so no false edge is fabricated. Wildcards, `baseUrl`, and array `extends` are preserved. As part of the fix I also collapsed six duplicated inline alias-resolution loops into one helper, so this can't drift again. Added regression tests for second-target-wins, first-target-wins-when-both-exist, and none-exist. Ships in the next release. Thanks for the precise report.

PR #2940, opened 2026-08-21
fix(extract): invalidate tsconfig alias and baseUrl caches on edit (#2917)

### LINKED ISSUE #2917: JS/TS: tsconfig alias and baseUrl caches are never invalidated, so an edit to compilerOptions is invisible for the life of the process (watch / MCP)

ousamabenyounes:
Fixes #2917.

## The bug

`_TSCONFIG_ALIAS_CACHE` / `_TSCONFIG_BASEURL_CACHE` are keyed on the config path
string, with no mtime component and no invalidation anywhere in the package —
unlike `_WORKSPACE_PACKAGE_CACHE`, which is both mtime-keyed and cleared per run.
So for any process that calls `extract()` more than once (`graphify watch`, the
MCP server, library loops), an edit to `compilerOptions.paths` or `baseUrl` is
never observed again. It fails silently: the import edges still exist and still
look plausible, they just point at the previous target.

The issue's repro, verbatim, on `v8` (`b2cd362`):

```
alias -> src/*: ['src_target_hit']
alias -> lib/*: ['src_target_hit']      <-- tsconfig now says lib/*
```

and after this PR:

```
alias -> src/*: ['src_target_hit']
alias -> lib/*: ['lib_target_hit']
```

## The fix

Two layers, because the two callers differ:

- **Cache entries key on the config's mtime**, mirroring the manifest-mtime key
  `_load_workspace_packages` already uses in the same module. This covers
  `extract_js()`, which reads both caches directly and never goes through
  `extract()`.
- **`extract()` clears both caches** beside the workspace/XAML/markdown caches.
  This is not redundant: an alias inherited through an `extends` chain is keyed
  on the leaf config only, so an edit to the *base* config needs the run
  boundary. That leaf-only key is the same scope `_load_workspace_packages` has
  (root manifests, not transitive includes) — happy to widen the key to the
  whole `extends` chain if you'd rather close that case too.

A `stat()` on a config that vanished between the `exists()` probe and the key
read falls back to a distinct sentinel key, so the miss is recomputed rather
than served stale.

RED — the three new tests against unmodified `v8`, production files stashed:

```
$ git stash push graphify/extract.py graphify/extractors/resolution.py
$ uv run --frozen pytest tests/test_jsconfig_baseurl.py -q --tb=line
FAILED tests/test_jsconfig_baseurl.py::test_tsconfig_paths_alias_edit_is_seen_by_a_second_extract
FAILED tests/test_jsconfig_baseurl.py::test_tsconfig_baseurl_edit_is_seen_by_a_second_extract
FAILED tests/test_jsconfig_baseurl.py::test_tsconfig_alias_edit_is_seen_without_a_full_extract
3 failed, 11 passed, 1 warning in 0.47s
```

with the leak visible in the assertion:

```
E   AssertionError: assert 'b_root_mods_w' in {'a_root_mods_w'}
```

GREEN — same file with the fix:

```
$ uv run --frozen pytest tests/test_jsconfig_baseurl.py -q
14 passed, 1 warning in 0.50s
```

Full suite, unchanged against the `v8` baseline (the 4 failures are pre-existing
in my local environment — 3 `test_ollama` backend-detection tests read env vars I
have set, plus `test_collect_files_skips_hidden` — all 4 fail identically on
clean `v8`):

```
$ uv run --frozen pytest tests/ -q
4 failed, 4880 passed, 11 skipped, 4 warnings in 136.95s
```

`ruff check` clean on `graphify/` and the touched test file; the five
`tools.skillgen` validators and `graphify --help` all pass.

The two existing per-run alias tests (`test_tsconfig_paths_alias_unchanged`,
`test_tsconfig_wins_when_both_configs_present`) are untouched and still pass, so
alias precedence (#1269, #927, #1531, #2153) is unaffected.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@ Full release notes with details on each version: [GitHub Releases](https://githu
 
 ## 0.9.48 (2026-08-20)
 
+- Fix: `graphify watch`, the MCP server, and any caller that runs `extract()` or `extract_js()` more than once in a process now see an edited `tsconfig.json` / `jsconfig.json`; the `compilerOptions.paths` and `baseUrl` caches are keyed by config mtime — as the workspace-manifest cache already is — and cleared per run so an edit to an `extends` base config lands too, instead of silently wiring imports to the previous alias target for the life of the process (#2917, thanks @sashankh).
 - Fix: a control character in a node label or id no longer aborts the whole export; the GraphML and Obsidian exporters scrub only the characters those formats forbid (tab, newline, and non-ASCII letters are preserved), and `graph.json` and its byte-identity round-trip are untouched (#2897, thanks @abhay-codes07).
 - Fix: `graphify update` / `label` / `cluster-only` no longer leave a large graph without a `graph.html`; the aggregated community view renders instead of raising, a failed render preserves the previous file, and a missing `graph.html` is regenerated on the no-change fast path without reclustering (#2853, thanks @oleksii-tumanov).
 - Feature: `graphify extract --no-dedup` skips the fuzzy near-duplicate merge on build and incremental merge, for operators who would rather keep distinct symbols that fuzzy-matched; exact-id uniqueness is unaffected and the flag arms the shrink guard so a surprising node drop is refused loudly (#2881, thanks @rajarshidattapy).
--- a/graphify/extract.py
+++ b/graphify/extract.py
@@ -70,6 +70,7 @@ from graphify.extractors.resolution import (  # noqa: E402,F401
     _JS_PRIMITIVE_TYPES,
     _JS_RESOLVE_EXTS,
     _TSCONFIG_ALIAS_CACHE,
+    _TSCONFIG_BASEURL_CACHE,
     _VUE_SCRIPT_LANG_RE,
     _VUE_SCRIPT_RE,
     _WORKSPACE_MANIFEST_NAMES,
@@ -5522,9 +5523,14 @@ def extract(
     _check_tree_sitter_version()
     _raise_recursion_limit()
     # Workspace package manifests/globs can change during watch or repeated extraction.
+    # The tsconfig/jsconfig caches are mtime-keyed per config file, which the run
+    # boundary completes: an alias inherited through an `extends` chain is keyed on
+    # the leaf config only, so an edit to the BASE config needs this clear (#2917).
     _WORKSPACE_PACKAGE_CACHE.clear()
     _XAML_CSHARP_CLASS_CACHE.clear()
     _MD_LINK_INDEX_CACHE.clear()
+    _TSCONFIG_ALIAS_CACHE.clear()
+    _TSCONFIG_BASEURL_CACHE.clear()
 
     # Infer a common root for cache keys (use first diverging segment, not sum of all matches)
     try:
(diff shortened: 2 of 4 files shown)


safishamsi:
Thanks @ousamabenyounes — we went with #2918 for #2917 (the per-run cache clear, which subsumes the mtime-key approach and avoids the coarse-clock edge case); it shipped in v0.9.49. Closing this as superseded. Release: https://github.com/Graphify-Labs/graphify/releases/tag/v0.9.49

PR #2947, opened 2026-08-22
fix(js): capture function-valued object properties like shorthand methods

rajanpanth:
Two spellings of the same exported API produce different graphs:

```js
module.exports = { m(){ return 1; } };        // m() captured
module.exports = { m: () => 1 };              // nothing
module.exports = { m: function(){ return 1; } }; // nothing
```

The shorthand parses as a `method_definition`, which the generic function branch captures. The arrow and function-expression forms parse as `pair` nodes, which nothing handled, so those symbols vanished. Same class of gap as the factory object API fix (#2745): common style, silently missing surface. `{ handler: async () => {...} }` route tables and options objects are the everyday casualties.

## Fix

A `pair` whose value is a function type is intercepted at the same walk position as the function branch and mirrors it exactly: same naming, same class-vs-file scoping, same `callable_def_nids` and `local_bound_names` registration, same `function_bodies` tracking so calls made inside the property resolve. Only plain identifier keys are named; computed and string keys stay out of the graph as before.

Because it sits in the same walk with the same guards, the scoping envelope is byte-for-byte the shorthand envelope:

| Case | Shorthand today | Arrow before | Arrow after |
| --- | --- | --- | --- |
| `module.exports = { m }` | `m()` | nothing | `m()` |
| `register({ m })` | `m()` | nothing | `m()` |
| `const api = { m }` | only `api` | only `api` | only `api` (unchanged) |

## Verification

- 7 regression tests added (parity for all spellings, call-argument objects, body-call resolution, the unchanged `const` scoping baseline, computed/string keys still skipped). With the engine change reverted, 6 of 7 fail.
- `calls` edges from inside an arrow property resolve: `module.exports = { run: () => helper() }` emits `run() -> helper()`.
- Full suite: 4650 passed. The failures on my machine (127) are identical with and without this change, all environmental Windows CLI test issues, none in extraction.

### DIFF
--- a/graphify/extractors/engine.py
+++ b/graphify/extractors/engine.py
@@ -4041,6 +4041,12 @@ def _extract_generic(
                 and config.ts_module in ("tree_sitter_javascript",
                                          "tree_sitter_typescript")):
             pair_value = node.child_by_field_name("value")
+            # `{ m: (() => {}) }` wraps the function in parenthesized_expression
+            # nodes; unwrap them so the parenthesized spelling matches too.
+            while (pair_value is not None
+                   and pair_value.type == "parenthesized_expression"):
+                pair_value = next(
+                    (c for c in pair_value.children if c.is_named), None)
             if pair_value is not None and pair_value.type in _JS_FUNCTION_VALUE_TYPES:
                 pair_key = node.child_by_field_name("key")
                 if pair_key is not None and pair_key.type == "property_identifier":
--- a/graphify/extractors/engine.py
+++ b/graphify/extractors/engine.py
@@ -4029,6 +4029,41 @@ def _extract_generic(
                     add_edge(parent_class_nid, field_nid, "defines", line, context="field")
             return
 
+        # JS/TS: a function-valued property in an object literal defines a
+        # callable member of the same API surface as the shorthand form. The
+        # shorthand `{ m(){} }` is a `method_definition` and lands in the
+        # function branch below, but `{ m: () => {} }` and
+        # `{ m: function(){} }` parse as `pair` nodes, so the arrow and
+        # function-expression spellings vanished from the graph while the
+        # shorthand spelling was captured. Mirror the function branch exactly:
+        # same walk position, same scoping, same body tracking.
+        if (t == "pair"
+                and config.ts_module in ("tree_sitter_javascript",
+                                         "tree_sitter_typescript")):
+            pair_value = node.child_by_field_name("value")
+            if pair_value is not None and pair_value.type in _JS_FUNCTION_VALUE_TYPES:
+                pair_key = node.child_by_field_name("key")
+                if pair_key is not None and pair_key.type == "property_identifier":
+                    func_name = _read_text(pair_key, source)
+                    # Same #1899 guard as the function branch: a name that
+                    # normalizes to nothing would collapse onto the prefix.
+                    if func_name and normalize_id(func_name):
+                        line = node.start_point[0] + 1
+                        if parent_class_nid:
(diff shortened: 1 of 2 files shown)


rajanpanth:
Both advisories addressed.

**Parenthesized values**: real gap, fixed in 50cc537. `{ m: (() => 1) }` wraps the function in `parenthesized_expression` nodes, which the type check skipped. The branch now unwraps those layers first, so single and doubled parens are captured, while a parenthesized non-function value (`{ m: (fn()) }`) stays out of the graph. Test added for both sides. 8 passing.

**Const-bound object literals**: no output change there, and the test suite locks it. `const api = { m: () => 1 }` emits only the `api` node, matching the shorthand baseline exactly, because lexical declarations are handled before the generic walk descends. The pair branch only fires where `method_definition` already fired, which is the parity this PR is about. `test_const_object_scoping_baseline_unchanged` asserts shorthand and arrow spellings produce identical labels for the const case.

rajanpanth:
Rebased onto current `v8` to clear the conflict.

The conflict was worth a note: `v8` has since added a Ruby `singleton_class` branch at the same position in `walk()` where this PR adds the JS `pair` branch. They are independent guards, so I kept both, Ruby first and then the JS one, rather than letting either win.

Verification after the rebase:

- `tests/test_js_object_property_functions.py`: 8 passed.
- The Ruby work that conflicted still passes: `pytest -k "ruby or singleton"` gives 89 passed, 4 skipped, so keeping both branches did not disturb the new singleton handling.
- Mutation check: with `graphify/extractors/engine.py` reverted to `v8` and the tests kept, 6 of the 8 fail, including `test_mixed_spellings_all_captured`, `test_parenthesized_function_values_captured` and `test_property_body_calls_resolve`. Restored, back to 8 passing.

Diff is unchanged in substance from the reviewed version: `+41` in `engine.py` and the new test file.

rajanpanth:
On the advisory finding, "Named function-expression property symbols are no longer emitted" at `engine.py:4901`: I checked it and I do not think it holds. The review flagged it as agreed but not verified, so here is the verification.

I ran the same six shapes through `extract_js` on `v8` and on this branch:

| source | `v8` | this PR |
|---|---|---|
| `module.exports = { named: function namedInner(){} }` | `[]` | `['named()']` |
| `module.exports = { anon: function(){} }` | `[]` | `['anon()']` |
| `module.exports = { outer: { named: function namedInner(){} } }` | `[]` | `['named()']` |
| `module.exports = { named: function namedInner(){ return helper(); } }` | `['helper()']` | `['helper()', 'named()']` |
| `const api = { named: function namedInner(){} }` | `['api']` | `['api']` |
| `const f = function namedInner(){}` | `['f()']` | `['f()']` |

No row loses a symbol. The first four gain one, and the last two are unchanged.

The reason the finding does not apply: on `v8` a named function-expression property emitted nothing at all, so `namedInner` was not being dropped by this change, it was never there. The walk did not reach into `pair` values, which is the bug this PR fixes.

The last row is the one that matters for the concern behind the finding. A named function expression outside an object literal is not a `pair`, so it never enters the new branch and still emits `f()` exactly as before.

On which name gets used: the new branch emits the property key, so `{ named: function namedInner(){} }` becomes `named()` rather than `namedInner()`. That is deliberate and it is the point of the PR. `named` is what callers see on the API surface, `namedInner` is only bindable inside the function body for self-reference, and the shorthand spelling `{ named(){} }` has always emitted `named()`. Emitting the inner name instead would break the parity these tests exist to lock.

Happy to add the named-function-expression case to the test file as an explicit guard if you would like that pinned down.

PR #2959, opened 2026-08-23
feat(sql): make tree-sitter-sql a core dependency so default installs parse .sql (#1745)

egarcia74:
## Problem

A default `uv tool install graphifyy` / `pipx install graphifyy` contributes nothing for `.sql` files: the grammar lives behind the `[sql]` extra, so `extract_sql` bails with an error result and the #1745 warning is the only signal. SQL is a mainstream corpus language — schemas, migrations, stored procedures — and the failure mode is exactly the silent-skip shape #1745 was about: the operator sees a graph, just one with an entire language missing, and nothing tells them at install time.

We hit this in the field: a 391-file T-SQL data-migration corpus produced zero SQL object nodes on a fresh default install, and the fix was discoverable only by reading extractor source.

## Why core rather than a better hint

`tree-sitter-sql` ships prebuilt abi3 wheels for every supported platform (Windows/macOS/Linux glibc+musl, x86_64+arm64) from 0.3.9, so promoting it keeps the default install toolchain-free — the reason `tree-sitter-dm` stays optional (#1104) does not apply here. The pin `>=0.3.9,<0.4` carries an upper bound like every other grammar pin.

## What this PR does

- `tree-sitter-sql>=0.3.9,<0.4` moves from the `[sql]` extra into `[project.dependencies]`, with a comment recording the wheel-availability rationale so the packaging policy stays legible.
- The `sql` extra is kept as an alias, so existing `graphifyy[sql]` install commands and scripts keep resolving; it now adds nothing beyond the core install.
- `.sql` is removed from `_EXTRA_FOR_EXTENSION`, so the #1745 warning no longer sends users to a now-redundant extra; the extractor's missing-grammar error now says the install is incomplete and names the direct repair.
- README extras table updated; the `sql` row is replaced with a note explaining the move and that a normal upgrade is all an existing `[sql]`/`pipx inject` install needs.
- Version bump and changelog entry are deliberately **not** included — those look like maintainer release-commit territory here (`chore: bump to 0.9.48`), so comments and messages describe the grammar as a core dependency without naming a release.

## Tests

New `tests/test_sql_core_dependency.py` pins the fix at the packaging layer so a regression (the dependency sliding back into an extra) fails the suite rather than resurfacing as a field report:

- the grammar must be in `[project.dependencies]`, with an upper bound;
- `.sql` must not be mapped to an optional extra in the hint map;
- the dev environment itself must satisfy the dependency;
- an end-to-end table/view/procedure corpus must produce object nodes, `contains`/`references`/`reads_from` edges, and no dangling edge sources.

Existing SQL tests in `tests/test_multilang.py` convert from `importorskip` (which would silently skip the whole SQL surface if the grammar went missing — the exact failure mode this PR removes) to a hard import that fails instead. `test_extract_warns_when_sql_grammar_missing` now simulates genuine absence via both `sys.modules` and a `find_spec` monkeypatch, and asserts the new message names the direct repair, not the extra.

## Validation

- Fresh-venv wheel install pulls `tree-sitter-sql` transitively; a SQL fixture produces nodes/edges with no extra or `pipx inject` involved; the missing-grammar warning path proven live by blanking the module.
- Full suite: `6 failed, 4835 passed, 28 skipped` — the 6 (`test_ollama.py`, `test_ollama_retry_cap.py`) fail identically on unmodified `v8` in this environment (local Ollama config), none touch SQL or packaging.
- `uv.lock` regenerated.

## Note

Independent of my companion PR #2960 fixing T-SQL routine recovery in `graphify/extractors/sql.py`; both touch `tests/test_multilang.py`, and a local merge of the two branches resolves trivially (we run the combined branch in production against the 391-file corpus).


egarcia74:
On the Graphify-review finding ("SQL dependency is only added to the lockfile, not the packaging source", flagged high at `uv.lock:1338`): this doesn't match the diff. `tree-sitter-sql>=0.3.9,<0.4` is added to `[project.dependencies]` in `pyproject.toml` (line 50 on the branch tip) — the lockfile change is downstream of that. The PR also ships `test_tree_sitter_sql_is_a_core_dependency`, which pins exactly this invariant by parsing `pyproject.toml` and fails if the dependency ever slides back into an extra; it passes on this branch. (The review itself notes the finding was "agreed but NOT verified".) Happy to adjust if a maintainer sees something I'm missing.

febryo-sc:
Second field data point for this, from a default `uv tool install graphifyy` (0.9.5, later
0.9.56) on a TypeScript codebase that keeps every statement in its own `.sql` file:

- **270 `.sql` files → 0 nodes**, across roughly 30 graph builds over a month. Every file was
  scanned (each had both `ast_hash` and `semantic_hash` in `manifest.json`), and `.sql` is in
  `CODE_EXTENSIONS`, so nothing in the pipeline looked wrong from the outside.
- `uv tool install --force --with tree-sitter-sql graphifyy` + a rebuild: **0 → 312 nodes**
  (a file node per statement, plus table nodes with FK `references` edges from the migrations).
- The only trace of the failure anywhere was the discarded `error` string inside
  `extract_sql`'s ImportError branch. `GRAPH_REPORT.md`'s Corpus Check counted the files as
  part of the corpus.

Worth noting the shape generalizes past SQL: `extractors/terraform.py` (`.hcl`) and
`extractors/robot.py` return the same `{"nodes": [], "edges": [], "error": "... not
installed"}`, so whatever surfacing lands here (install-time check, one warning per run, or a
line in Corpus Check) would ideally cover any extension that is in `CODE_EXTENSIONS` but whose
grammar is absent — not just `.sql`.

PR #2960, opened 2026-08-23
fix(sql): recover T-SQL bracket-named, CREATE OR ALTER, and PROC routines

egarcia74:
## Problem

T-SQL stored procedures and functions are silently dropped from the graph. On a real 391-file T-SQL data-migration corpus (SQL Server DACPAC project: staging/model/transform layers, `[Schema].[Object]` naming throughout), 0 of 26 stored procedures produced a node — while tables and views extracted fine, so the gap is invisible unless you go counting.

Two independent defects stack:

1. **The grammar can't parse the T-SQL routine shape at all.** `CREATE PROCEDURE ... AS BEGIN ... END` never parses structurally in tree-sitter-sql (there is no `create_procedure` production for the `AS BEGIN` body idiom), so ERROR-node regex recovery is these objects' *only* path into the graph.
2. **The recovery regex doesn't speak T-SQL.** It accepted only unquoted or double-quoted names — no bracket-delimited names (`CREATE PROCEDURE [dbo].[usp_Load]`), no `CREATE OR ALTER`, no `PROC` shorthand. On this corpus, every routine uses brackets, so recovery recovered nothing.

## What this PR does

All changes are in `graphify/extractors/sql.py`:

- **One shared pattern, `_ROUTINE_RECOVERY_RX`**, hoisted to module level and used by both recovery sites (the walk-time ERROR-node scan and the whole-file `has_error` fallback). The two sites had drifted; the drifted second site is what minted a phantom `dbo()` node for a mixed-delimiter name (`CREATE PROCEDURE dbo.[usp_Mixed]`) — the schema captured as if it were the routine. Sharing the pattern fixes that and prevents re-drift.
- **Bracket-delimited identifiers**, including T-SQL's `]]` escape for a literal `]` inside a name (`[dbo].[a]]b]` recovers as one routine, fully consumed, not truncated at the escape).
- **`CREATE OR ALTER`** alongside the existing `OR REPLACE`, and the **`PROC`** shorthand alongside `PROCEDURE`.
- **Comment masking in the whole-file fallback**: the raw-text scan now runs over `_mask_sql_comments(src)` (offset-preserving blanking of `--` and `/* */`), so commented-out DDL in a file with an unrelated parse error can no longer fabricate routine nodes.

Recovered routines are name-only nodes (no body `reads_from` edges), matching the existing PL/pgSQL recovery behavior — a named node with a `contains` edge beats an invisible object.

## Result on the real corpus

391 `.sql` files, before → after: **1 → 28 routine nodes**, including every bracket-named `[Audit]`/`[Utils]`/`[Monitoring]` procedure; table and view counts unchanged; no phantom schema nodes.

## Tests

Added in `tests/test_multilang.py`, each verified to fail with the fix reverted:

- bracket-delimited procedure recovers under its full `[dbo].[usp_X]` name;
- `CREATE OR ALTER PROCEDURE` recovers;
- `PROC` shorthand recovers;
- mixed-delimiter name (`dbo.[usp_Mixed]`) yields exactly one node — both recovery sites agree on the captured name (the phantom-`dbo()` regression);
- `]]` escape inside a bracket-delimited name is consumed, capturing `[dbo].[a]]b]()` as one routine;
- commented-out DDL in an ERROR-bearing file does not fabricate a routine node.

## Validation

- Full suite: `6 failed, 4836 passed, 28 skipped` — the 6 (`test_ollama.py`, `test_ollama_retry_cap.py`) fail identically on unmodified `v8` in this environment (local Ollama config), none touch SQL.

## Note

Independent of my companion PR #2959 making `tree-sitter-sql` a core dependency; both touch `tests/test_multilang.py`, and a local merge of the two branches resolves trivially (the combined branch produced the 1 → 28 measurement above and runs in production).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,10 +2,6 @@
 
 Full release notes with details on each version: [GitHub Releases](https://github.com/safishamsi/graphify/releases)
 
-## 0.9.49 (unreleased)
-
-- Fix: T-SQL stored procedures and functions are no longer silently dropped from the graph — the ERROR-node recovery now accepts bracket-delimited names (`CREATE PROCEDURE [dbo].[usp_Load]`) and T-SQL's `CREATE OR ALTER`, both of which previously recovered nothing (the `AS BEGIN...END` body idiom never parses structurally, so recovery was these objects' only path into the graph). The two recovery sites now share one pattern, fixing a related defect where a mixed-delimiter name (`dbo.[usp_Mixed]`) produced a second phantom node named after the schema. The `PROC` shorthand is accepted alongside `PROCEDURE`, and the whole-file recovery scan now masks SQL comments so commented-out DDL in a file with an unrelated parse error cannot fabricate routine nodes. Recovered routines are name-only nodes (no body `reads_from` edges), matching the existing PL/pgSQL recovery.
-
 ## 0.9.48 (2026-08-20)
 
 - Fix: a control character in a node label or id no longer aborts the whole export; the GraphML and Obsidian exporters scrub only the characters those formats forbid (tab, newline, and non-ASCII letters are preserved), and `graph.json` and its byte-identity round-trip are untouched (#2897, thanks @abhay-codes07).
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 Full release notes with details on each version: [GitHub Releases](https://github.com/safishamsi/graphify/releases)
 
+## 0.9.49 (unreleased)
+
+- Fix: T-SQL stored procedures and functions are no longer silently dropped from the graph — the ERROR-node recovery now accepts bracket-delimited names (`CREATE PROCEDURE [dbo].[usp_Load]`) and T-SQL's `CREATE OR ALTER`, both of which previously recovered nothing (the `AS BEGIN...END` body idiom never parses structurally, so recovery was these objects' only path into the graph). The two recovery sites now share one pattern, fixing a related defect where a mixed-delimiter name (`dbo.[usp_Mixed]`) produced a second phantom node named after the schema. The `PROC` shorthand is accepted alongside `PROCEDURE`, and the whole-file recovery scan now masks SQL comments so commented-out DDL in a file with an unrelated parse error cannot fabricate routine nodes. Recovered routines are name-only nodes (no body `reads_from` edges), matching the existing PL/pgSQL recovery.
+
 ## 0.9.48 (2026-08-20)
 
 - Fix: a control character in a node label or id no longer aborts the whole export; the GraphML and Obsidian exporters scrub only the characters those formats forbid (tab, newline, and non-ASCII letters are preserved), and `graph.json` and its byte-identity round-trip are untouched (#2897, thanks @abhay-codes07).
--- a/graphify/extractors/sql.py
+++ b/graphify/extractors/sql.py
@@ -185,7 +185,11 @@ def _mask_sql_comments(text: str) -> str:
             # most routines on lines already entangled with the broken one (a
             # conservative false negative; a routine named like [a--b] is
             # inside that loss); under-blanking is what fabricates, and every
-            # reading's blank set stays a subset of this one.
+            # reading's blank set stays a subset of this one — except where a
+            # */ + * versus * + /* token split makes two readings consume the
+            # same /, an irreducible divergence whose only closure would be
+            # blanking to EOF on every */* sequence (accepted, documented
+            # limitation; the token sequence appears in no dialect's idiom).
             closer = '"' if c == '"' else "]"
             j = i + 1
             closed = False
--- a/graphify/extractors/sql.py
+++ b/graphify/extractors/sql.py
@@ -71,7 +71,7 @@ _ROUTINE_RECOVERY_RX = re.compile(
 # quote), judged not worth the swallow risk in a recovery-only path.
 
 
-def _mask_sql_comments(text: str) -> str:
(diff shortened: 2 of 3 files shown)


egarcia74:
Addressed the Graphify-review finding (`--` inside string literals treated as a comment opener) in b6d4dfc, and it turned out to be half of a symmetric pair:

- **The mask is now literal-aware**: single-quoted strings (with `''` escapes), double-quoted identifiers, and bracket-delimited identifiers (with `]]` escapes) are preserved before comments are blanked, so `'-- note'` / `'open /* here'` can no longer hide DDL sharing the span. Literal patterns are deliberately single-line — this mask only runs on files that already failed to parse, where an unclosed quote is likely, and a multi-line match would let one unclosed delimiter swallow real DDL below it (a miss is visible; a swallow is silent).
- **The walk-time ERROR-node scan now masks too**: while verifying the fix I found the fabrication risk the whole-file scan was cured of was still live at the other recovery site — an ERROR blob whose byte span covers commented-out DDL minted a routine node from it (reproduced with a `-- CREATE PROC ...` line sandwiched between broken segments).

Both directions are mutation-tested: a literal-blind mask fails the new unit pin (`test_mask_sql_comments_preserves_literals_and_blanks_comments`), and an unmasked ERROR scan fails the extended fabrication test. Full suite after the change: `6 failed, 4838 passed, 28 skipped` — same 6 environmental Ollama failures as on unmodified `v8`; `ruff check` clean.

egarcia74:
Both advisory findings from the latest Graphify review triaged:

- **"Routine recovery matches DDL keywords inside delimited identifiers" — confirmed real and fixed in 8c8148c.** Reproduced first: `SELECT 1 AS [CREATE PROCEDURE dbo.usp_Phantom pending];` in a file with an unrelated parse error minted `dbo.usp_Phantom()`. The mask must preserve delimited identifiers verbatim (they carry the recoverable names), but the recovery regex wasn't span-aware. The scanner now also reports the [start, end) of every preserved identifier span, and the whole-file scan skips any match whose `CREATE` starts inside one — a genuine statement's *name* may be a delimited identifier; its `CREATE` never is. Pinned by `test_sql_ddl_keywords_inside_delimited_identifiers_do_not_fabricate` (fails with the skip removed); masking behavior is unchanged, so the frozen-baseline differential fuzz and structural fuzz apply as-is.

- **"Dollar-quoted SQL text can fabricate routines when recovery is armed" — true, and already a documented, deliberate limitation.** PostgreSQL dollar-quoting (and MySQL default-mode double-quoted strings, and multi-line single-quoted dynamic SQL) are named as unmodelled in the comment block on `_scan_sql`, with the rationale: modelling them in a recovery-only path that runs exclusively on files that already failed to parse risks one unclosed delimiter swallowing real DDL — a silent loss — whereas the residual fabrication requires DDL-shaped text inside such quoting in an already-broken file, and remains behind the `root.has_error` gate. Happy to revisit if maintainers prefer the other side of that trade.

Validation: full suite `6 failed, 4843 passed, 28 skipped` (the 6 are the same environmental Ollama failures as unmodified v8), `ruff check` clean.

egarcia74:
Latest Graphify-review findings triaged:

- **"Routine recovery matches CREATE inside bare identifiers" — confirmed real and fixed in f3d0248.** Reproduced: `SELECT AUTOCREATE PROCEDURE x FROM t;` in an error-bearing file minted a phantom `x()` — the regex had no leading word boundary, and while delimited identifiers are span-skipped at the scan site, a bare word has no span. `\bCREATE` closes it; regression test `test_sql_create_inside_a_bare_word_does_not_fabricate` fails with the boundary removed.
- **"Routine recovery scans PostgreSQL dollar-quoted strings"** — same finding as the previous run, answered above: a documented, deliberate limitation of the recovery-only path (see the comment block on `_scan_sql`), behind the `root.has_error` gate; happy to model dollar-quoting if maintainers prefer that side of the swallow-risk trade.

Validation: full suite `6 failed, 4844 passed, 28 skipped` (same environmental Ollama failures as unmodified v8), `ruff check` clean.

egarcia74:
Triage of the latest Graphify-review run (on f3d0248), for whoever reviews this next:

- **"Recovery scan reads names from masked_src, blanking delimited identifier content" — not correct.** A captured name can only be verbatim source text: the name atoms match either `[\w$]+` (word characters — blanked positions are spaces, which cannot participate) or delimited forms that require their real delimiters, which blanking removes. A name therefore either comes from a *preserved* identifier span (copied verbatim, `""`/`]]` escapes included) or is not captured at all — a distrusted span like `[a--b]` is wholly blanked and yields no match (the documented conservative loss), never a mangled name. Pinned by the exact-label assertions in the suite (`[dbo].[a]]b]()`, `"dbo"."a""b"()`) and by the 391-file corpus validation (28/28 procedures recovered with verbatim bracketed names).
- **The two dollar-quoting findings** are the same documented limitation answered in the two comments above (deliberate trade in a recovery-only path behind the `root.has_error` gate; see the comment block on `_scan_sql`).

The review bot re-derives findings on every push, so the dollar-quoting advisory recurs each run. To keep this thread readable I'll stop replying per-run here: every distinct finding to date is either fixed with a named, mutation-tested commit (four were real — thanks, genuinely useful) or answered with rationale above. Awaiting human review.

egarcia74:
Superseded by #3166 — the same change squashed to a single commit and rebased onto current v8 (0.9.51) so it cherry-picks cleanly, with the defect now tracked in #3164. The review history stays here: each Graphify-review finding on this PR was reproduced and either fixed or refuted in the thread above, and the final scanner is what #3166 carries. Closing this one.

PR #2964, opened 2026-08-23
fix(php,swift): drop relative-scope type keywords so they stop minting god nodes

rajatnagda45:
## What

PHP's `self` / `static` / `parent` and Swift's `Self` in **type position** (return types, parameters, fields) are relative-scope keyword *aliases* for the enclosing class — they are not referenceable types of their own.

The type-reference walkers (`_php_collect_type_refs`, `_swift_collect_type_refs`) matched them as ordinary type names and emitted `references` edges. Because these names are declared in no file, `ensure_named_node` mints **sourceless stubs** whose ids normalise (via `ids.normalize_id`) to single global `self` / `static` / `parent` nodes. Fluent setters and factories return these constantly:

```php
public function withHeader($h): static { /* … */ return $this; }
public function build(): self { /* … */ }
```
```swift
static func make() -> Self { /* … */ }
```

So across a real corpus every such method rewires onto a handful of false **god nodes** — `static` and `self` become hubs wired to a large fraction of the classes in the project.

## Reproduce

```php
<?php
class Builder {
    public function make(): self { return new Builder(); }
    public static function create(): STATIC { return new static(); }
    public function base(): parent { return $this; }
}
```

Before this change `extract_php` returns nodes labelled `self`, `static`, `parent` and `references` edges pointing at them. Swift `-> Self` behaves the same way.

## Fix

Skip these keywords in the two walkers, mirroring how `primitive_type` is already skipped:

- PHP: `self` / `static` / `parent`, matched case-insensitively (PHP keywords are case-insensitive, e.g. `STATIC`).
- Swift: `Self`.

Genuine class types in the same positions are untouched.

## Tests

Adds `test_php_relative_type_keywords_do_not_create_phantom_nodes` and `test_swift_self_return_type_does_not_create_phantom_node`, each asserting no phantom node/edge is produced while a real return-type reference still resolves. Full suite is green locally apart from the pre-existing order-dependent flake `test_label_communities_batches_when_over_batch_size` (#2877), which passes in isolation and is unrelated.

### DIFF
--- a/graphify/extractors/engine.py
+++ b/graphify/extractors/engine.py
@@ -633,6 +633,16 @@ def _php_name_text(node, source: bytes) -> str | None:
         return None
     return _read_text(node, source).rsplit("\\", 1)[-1] or None
 
+# PHP's relative-scope type keywords resolve to a class only in the context of the
+# enclosing class -- `self`/`static` are the current class, `parent` is its base.
+# They are not referenceable types of their own, and since they are declared in no
+# file the stub rewire collapses every `: self` / `: static` return type across the
+# corpus (fluent setters and factories emit them constantly) onto a single global
+# node, manufacturing a false god node. Skip them. PHP keywords are
+# case-insensitive, so match on the casefolded text.
+_PHP_RELATIVE_TYPE_KEYWORDS = frozenset({"self", "static", "parent"})
+
+
 def _php_collect_type_refs(node, source: bytes, generic: bool, out: list[tuple[str, str]]) -> None:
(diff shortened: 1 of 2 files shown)


rajatnagda45:
Hi @safishamsi — companion to #2963, same god-node class but in the PHP/Swift walkers (`: static` / `: self` builders and Swift `-> Self` factories collapsed onto shared hubs). Kept it to the two type-ref walkers with tests for each. Let me know if you'd rather fold the keyword sets somewhere shared. Thanks!

PR #2967, opened 2026-08-23
fix(watch): report graph delta on --force overwrite (#2833)

### LINKED ISSUE #2833: update --force overwrites silently: report what it replaced
`graphify update --force` overwrites the graph without saying what it replaced. Because the guard it bypasses exists to catch a graph shrinking unexpectedly, forcing removes the only signal that a rebuild lost content.

## The situation it leaves an operator in

The overwrite guard is good: it refuses when a rebuild would shrink the graph materially, and that refusal has caught real problems. But a legitimate rebuild also shrinks a graph — a repository removed from scope, a vendored tree newly excluded, a filter tightened. So `--force` is the normal way through, and after that:

- a **legitimate** reduction and an **accidental** one look identical
- the only way to tell them apart is to record the before-figures yourself, force, and compare

That comparison scaffolding gets written once per project and thrown away, which is usually the signal that a tool should do it rather than each consumer.

## What would remove the need

Have `--force` report what it overwrote:

```
--force: replacing 12,431 nodes / 48,002 edges with 12,429 / 47,996 (-2 nodes, -6 edges)
```

Two nodes is the interesting case. A large reduction is usually intentional and gets scrutinised; a small one is what slips through, and it is exactly what the guard would have caught had it not been bypassed.

Better still, `--force` could report the reduction **and the reason it was refused**, so the operator sees which threshold was crossed rather than only that one was.

## Why not just read the guard's message

The refusal names the threshold, but by the time you force you have decided to proceed — and the numbers are not carried into the run that actually writes. A reader of the resulting graph has no record that a shrink happened at all.

## Related

The value of this is highest where a reduction is *verifiable*: if an operator can independently derive the number of nodes they expect to lose, an exact match justifies the force and a mismatch stops it. "It looks like a small drop" is not evidence; "it dropped exactly the count I derived separately" is. Reporting the delta is what makes that check possible without bespoke scripting.

akshitj11:
`graphify update --force` skips the shrink guard with no record of what it replaced. A 2-node drop and a 2000-node drop look identical in the logs. Print the before/after node and edge counts when force overwrites an existing graph.

Fixes #2833

Tested locally: `tests/test_watch.py -k check_shrink`, full pytest suite, skillgen validators.

### DIFF
--- a/graphify/watch.py
+++ b/graphify/watch.py
@@ -920,7 +920,23 @@ def _check_shrink(
     Files in ``failed_sources`` never account for lost nodes: extraction did not
     complete, so their disappearance is the silent shrink this guard protects.
     """
-    if force or not existing_data:
+    if force:
+        if existing_data:
+            existing_nodes = existing_data.get("nodes", [])
+            existing_edges = (
+                existing_data.get("links") or existing_data.get("edges") or []
+            )
+            new_nodes = new_data.get("nodes", [])
+            new_edges = new_data.get("links") or new_data.get("edges") or []
+            old_n, old_e = len(existing_nodes), len(existing_edges)
+            new_n, new_e = len(new_nodes), len(new_edges)
+            print(
+                f"[graphify] --force: replacing {old_n} nodes / {old_e} edges "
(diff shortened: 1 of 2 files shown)


akshitj11:
Closing in favor of #2965, which reports after write and has integration tests.

PR #2969, opened 2026-08-23
fix(watch): ignore read-only inotify events so the watcher stops re-triggering on its own reads

Azeem1985:
## Problem
On Linux, watchdog's inotify observer emits `opened` (watchdog ≥ 2.3) and `closed_no_write` (watchdog ≥ 4) for every file open/close. `Handler.on_any_event` in `graphify/watch.py` treats every event type as a change, so the watcher's own AST rebuild — which reads every watched file — re-triggers the watcher: one debounce later it prints "N file(s) changed" with nothing modified, rebuilds again, and loops.

Measured on a real tree (17k nodes, ~1.7k watched files, graphify 0.9.46, Ubuntu, watchdog 6): "1683 file(s) changed" every 1–3 minutes while `find -mmin -4` showed 0 real changes; a 90-second sample caught 18,082 events, all `opened` / `closed_no_write`, none `modified`; ~26 CPU-minutes per hour; the watcher was memory-killed 15 times in three days (systemd restarted it and it looped again); and the `needs_update` flag was re-written after every doc edit, which permanently softens `hook-guard read --strict` (the stale-flag check runs before the strict check).

## Fix
A small helper `_is_read_only_event(event)` returns True for `opened` / `closed_no_write`; `on_any_event` ignores those. Creation, modification, move, deletion and close-after-write still count. macOS uses `PollingObserver`, which never emits these, so behaviour there is unchanged.

## Tests
Two tests added to `tests/test_watch.py` (plain event objects, and the real watchdog event classes when importable). `tests/test_watch.py`: 130 passed.


Azeem1985:
Re-targeted from `main` to `v8` (the development branch — sorry, my mistake on opening). The earlier automated review ran while the base was `main`, so it diffed the whole v8↔main divergence (100 files / ~280k lines) and its findings (publish.yml, Dockerfile, ci.yml) belong to that divergence, not to this change. The actual change here is 2 files, +48/-1: `graphify/watch.py` (ignore `opened`/`closed_no_write`) and `tests/test_watch.py` (2 tests). `tests/test_watch.py`: 130 passed; full suite on the branch: 4,812 passed, 50 skipped.

Azeem1985:
Closing this in favour of #3070, which is the same change on a branch cut from `v8` — **3 files** instead of the 13 this one shows.

This PR was opened from a branch that also carried unrelated work from our fork, so its diff was 13 files of the watch fix plus the whole of our fork's in-flight work. That is my mistake in how I cut the branch, and it made this unreviewable. Nothing about the proposed change has altered; only the packaging.

#3070 has the full rationale and measurements. Sorry for the extra noise, and thank you for your patience.

PR #2982, opened 2026-08-23
feat(llm): send an explicit reasoning_effort for Kimi models that support it (GRAPHIFY_KIMI_EFFORT)

Azeem1985:
## What
K3 advertises `valid_efforts ["low","high","max"]` on its /models endpoint; with nothing sent the server default (`high`) silently applies, while the gemini block already carries an effort setting. This adds `reasoning_effort` to the kimi backend config, read by the existing request plumbing and forwarded as `reasoning_effort`; env-overridable via `GRAPHIFY_KIMI_EFFORT` so no source edit is ever needed. Default in this PR is `max` — maintainers may prefer defaulting to unset; happy to adjust.


Azeem1985:
Re-targeted from `main` to `v8` (the development branch — my mistake on opening); the earlier automated review diffed the v8↔main divergence rather than this change. Actual change: 1 file, +6/-0 in `graphify/llm.py` — a `reasoning_effort` key for the kimi backend (env-overridable via `GRAPHIFY_KIMI_EFFORT`), read by the existing plumbing. Happy to default it to unset instead of `max` if you prefer the server default to keep applying.

Azeem1985:
Closing this in favour of #3071, which is the same change on a branch cut from `v8` — **3 files** instead of the 12 this one shows.

This PR was opened from a branch that also carried unrelated work from our fork, so its diff was 12 files of a six-line change plus the whole of our fork's in-flight work. That is my mistake in how I cut the branch, and it made this unreviewable. Nothing about the proposed change has altered; only the packaging.

#3071 has the full rationale and measurements. Sorry for the extra noise, and thank you for your patience.

PR #2991, opened 2026-08-24
fix(extract): keep the definition site when a C++/ObjC decl/def pair is merged

### LINKED ISSUE #2990: C/C++/ObjC decl/def merge discards the definition site — the merged node reports only the declaration
## Summary

For C/C++/ObjC, a symbol declared in `Foo.h` and defined in the sibling `Foo.cpp` is deliberately collapsed into **one** node keyed to the header (`_merge_decl_def_classes`, #1547/#1556). That merge is correct and I am not asking to change it — downstream resolvers need exactly one definition node.

The problem is that the merge **discards the impl node entirely**, so nothing in the graph records *where the symbol is actually implemented*. The surviving node's `source_file`/`source_location` name the declaration, and the definition's file and line — the thing a reader asking "where is this implemented?" wants — are lost with the dropped node.

Same class of harm as #2092 (a symbol whose reported `source_location` isn't where it lives), different cause.

## Repro

Two files, no config:

```cpp
// foo.h
class Foo {
public:
    void Alpha();
    void Beta();
};
```

```cpp
// foo.cpp
#include "foo.h"
void Foo::Alpha() { Beta(); }
void Foo::Beta()  { }
```

```
$ graphify update .
[graphify watch] Rebuilt: 5 nodes, 7 edges, 2 communities
```

`graph.json`:

| label | source_file | source_location | id |
|---|---|---|---|
| `Foo`   | foo.h   | L3 | `foo_foo` |
| `Alpha` | foo.h   | L4 | `foo_foo_alpha` |
| `Beta`  | foo.h   | L5 | `foo_foo_beta` |
| `foo.h` | foo.h   | L1 | `foo_h_foo` |
| `foo.cpp` | foo.cpp | L1 | `foo_cpp_foo` |

`foo.cpp` contributes only its file node. `Foo::Alpha()` at foo.cpp L2 and `Foo::Beta()` at L3 are gone, and no attribute anywhere points at them.

**Expected:** the merged node keeps naming the declaration (that is the design), but the definition site remains reachable from it.

**Also:** the drop is silent — no note or warning on any verbosity. `_report_id_collision` in `dedup.py` would warn about a two-file id collision, but this merge happens earlier, in `extract()`, so nothing is reported.

Renaming the impl file so the stems differ (`fooimpl.cpp`) keeps all 4 nodes, which confirms the path: the ids collide only because both stems slugify to `foo`.

## Why it matters

The merge fires for the ordinary C++ layout — declarations in a header, definitions in the sibling `.cpp` — so it is the common case, not an edge case. On a 1,767-file C++ codebase I measured, **541 of 700** implementation files larger than 3 KB contribute nothing but their file node; one 188 KB `.cpp` holds 198 method definitions whose lines are all unreachable from the graph.

The practical cost is that `get_node` answers "where is this implemented?" with a header line number, and the reader then has to scan a very large `.cpp` by hand. Call edges and symbol coverage are unaffected — the surviving node keeps the id, so edges still land correctly (I verified edge density is identical between a merged class and an unmerged one). This is purely a loss of provenance.

## Proposed fix

Record the dropped impl node's provenance on the survivor before discarding it:

- `definition_file` / `definition_location` on the merged node
- `get_node` prints a `Defined in:` line when present

Additive only: id, `label`, `source_file`, node count and edges are all unchanged, so **no existing graph is re-keyed** and the single-definition guarantee the resolvers depend on is untouched. When several impls fold onto one header (ObjC categories) the choice is deterministic — lowest `(source_file, source_location)` — so it does not depend on node arrival order.

Alternatives I considered and rejected:

- **Let the definition win the merge** (survivor becomes the `.cpp` node). Smallest diff, but the surviving label flips from `Alpha` to `Foo::Alpha()`, which breaks existing bare-name lookups.
- **Keep both nodes, disambiguate the impl id** (the shape used for Go in #2779, plus a `defines` edge). Nothing gets re-keyed, but node counts grow substantially on a header-heavy C++ repo and it needs an edge-resolution policy for which of the two calls should target.

Happy to switch to either if you prefer one — the additive version just seemed like the smallest change that fixes the reported harm.

## Environment

- reproduces on `v8` @ `b2cd362` (0.9.48, current default branch) and on the released 0.9.46
- Python 3.12.3, tree-sitter 0.25.2, Linux
- AST-only build, no LLM backend configured

PR follows.

winesoft-namjin-yun:
Fixes #2990.

## What

A C/C++/ObjC symbol declared in a header and defined in the sibling impl file is collapsed into one node keyed to the header by `_merge_decl_def_classes` (#1547, #1556). That merge is right and this PR leaves it alone — the downstream resolvers need exactly one definition node. But the dropped impl node took the definition's file and line with it, so the graph reported the declaration as the symbol's only location and `get_node` answered "where is this implemented?" with a header line number.

This records the dropped impl's provenance on the survivor before discarding it:

- `definition_file` / `definition_location` on the merged node
- `get_node` prints a `Defined in:` line when present

## Why this shape

Additive only — id, `label`, `source_file`, node count and edges are all unchanged, so **no existing graph is re-keyed** and the single-definition guarantee the resolvers depend on is untouched. When several impls fold onto one header (ObjC categories) the pick is deterministic, lowest `(source_file, source_location)`, so it does not depend on node arrival order.

Two alternatives are written up in #2990 (flip the merge so the definition wins; keep both nodes and disambiguate the impl id, the shape used for Go in #2779). Both change more than the reported harm needs — the first flips surviving labels from `Alpha` to `Foo::Alpha()` and breaks bare-name lookups, the second grows node counts on header-heavy C++ and needs an edge-resolution policy. Happy to switch if you would rather have one of those.

## Tests

`tests/test_languages.py`, on the existing `cpp_paired` fixture:

- `test_cpp_paired_merged_node_records_definition_site` — the single merged `bar` node keeps `source_file` = `Foo.h` and now carries `definition_file` = `Foo.cpp` with a location. Verified red before the fix (`definition_file` was `''`).
- `test_cpp_unpaired_symbol_has_no_definition_site` — a node that was never merged does not grow the attributes.

Full suite, same machine, with and without the patch:

```
without:  24 failed, 4568 passed, 211 skipped
with:     24 failed, 4570 passed, 211 skipped
```

Identical failure set (pre-existing, all environment-dependent: `test_ollama_retry_cap`, `test_terraform`, `test_skillgen`, `test_install_references`, `test_markdown_nested_frontmatter_survives` — none touch this path), +2 from the new tests.

## Real-world check

On a 1,767-file C++ codebase where 541 of 700 implementation files over 3 KB previously contributed nothing but their file node: one 188 KB `.cpp` recovers the definition sites of all 198 of its method definitions, e.g. a method declared at `…/Foo.h` L85 now carries `definition_location` L1002 in the sibling `.cpp`. Sanity check: no `definition_file` resolves to anything but an impl extension.

Environment: Python 3.12.3, tree-sitter 0.25.2, Linux, AST-only build with no LLM backend.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 Full release notes with details on each version: [GitHub Releases](https://github.com/safishamsi/graphify/releases)
 
+## Unreleased
+
+- Fix: a C/C++/ObjC symbol declared in a header and defined in the sibling impl file is merged into one node keyed to the header (#1547, #1556), and the dropped impl node took the definition's file and line with it — the graph reported the declaration as the symbol's only location, so "where is this implemented?" answered with a header line and the reader had to scan the `.cpp` by hand. The survivor now carries `definition_file`/`definition_location` and `get_node` names it ("Defined in:"); ids, labels, `source_file`, node count and edges are all unchanged, so existing graphs are not re-keyed and the single-definition guarantee the resolvers rely on is untouched. Several impls folding onto one header (ObjC categories) pick deterministically by lowest `(source_file, source_location)` (#2990).
+
 ## 0.9.48 (2026-08-20)
 
 - Fix: a control character in a node label or id no longer aborts the whole export; the GraphML and Obsidian exporters scrub only the characters those formats forbid (tab, newline, and non-ASCII letters are preserved), and `graph.json` and its byte-identity round-trip are untouched (#2897, thanks @abhay-codes07).
--- a/graphify/extractors/resolution.py
+++ b/graphify/extractors/resolution.py
@@ -2129,7 +2129,10 @@ def _merge_decl_def_classes(
     and leaves it alone, and the downstream resolvers see ONE definition. Because
     the colliding nodes already share an id, no edge re-pointing is needed: every
     edge that referenced the impl symbol already points at the surviving id. We
-    only drop the redundant duplicate node and prefer the header's label.
+    only drop the redundant duplicate node and prefer the header's label — but
+    the dropped impl node's provenance is preserved on the survivor as
+    ``definition_file`` / ``definition_location``, so the definition site is
+    still reachable from the merged node.
 
     GOD-NODE GUARDS (false merges are the main risk):
 
@@ -2200,6 +2203,30 @@ def _merge_decl_def_classes(
             if len(base_headers) > 1:
                 continue
             keeper = base_headers[0] if base_headers else min(headers, key=_source_stem)
+        # The keeper is the DECLARATION, so without this the graph reports the
+        # header as the symbol's only location and the definition site — the file
+        # and line a reader actually wants — is discarded with the dropped node.
+        # Recorded as separate attributes: the survivor's id, label and
+        # source_file are untouched, so no existing graph is re-keyed, no edge
+        # moves, and the single-definition guarantee downstream resolvers rely on
+        # is unchanged. Chosen deterministically (lowest source_file, then
+        # location) so an ObjC class whose members are split across several
+        # category impls does not depend on node arrival order.
+        impls = sorted(
+            (n for n in group
+             if n is not keeper
+             and Path(str(n.get("source_file", ""))).suffix.lower()
+             in _DECLDEF_IMPL_SUFFIXES),
+            key=lambda n: (str(n.get("source_file", "")),
+                           str(n.get("source_location", ""))),
+        )
+        if impls:
+            definition = impls[0]
+            if definition.get("source_file"):
+                keeper["definition_file"] = definition["source_file"]
+            if definition.get("source_location"):
+                keeper["definition_location"] = definition["source_location"]
+
         for node in group:
             if node is not keeper:
                 drop_objs.add(id(node))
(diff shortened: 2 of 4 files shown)


winesoft-namjin-yun:
The automated review above appears to be analyzing a different diff.

This PR changes 4 files — `CHANGELOG.md`, `graphify/extractors/resolution.py`, `graphify/serve.py`, `tests/test_languages.py` (+62/-2, single commit `a78e179`). The five findings all point at `.github/workflows/publish.yml` and `.github/workflows/release-graph.yml`, and the summary describes `.dockerignore` / `.gitattributes` / `FUNDING.yml` / `AGENTS.md` / pre-commit additions that aren't in this branch. The behavior-change evidence (`__getattr__` returning `save_query_result`, `run_benchmark`) is in modules this PR doesn't touch either.

The report notes its own baseline diverged from this PR's base (`v8`): "Health delta baseline: last indexed commit b2cd362 (diverged from this PR's base — delta is approximate)". That likely explains it.

So there's nothing here for me to address — but happy to act on any finding that does land on the changed lines.

safishamsi:
Shipped in v0.9.49 via authorship-preserving cherry-pick so you keep contributor-graph credit. Thanks @winesoft-namjin-yun! Release: https://github.com/Graphify-Labs/graphify/releases/tag/v0.9.49

PR #2992, opened 2026-08-24
fix(sql): extract CREATE TABLE inside BEGIN/COMMIT blocks

### LINKED ISSUE #2953: SQL: statements wrapped in BEGIN; … COMMIT; extract 0 nodes — 59 CREATE TABLEs silently absent across 8 migrations
# SQL: no nodes are extracted from statements wrapped in `BEGIN; … COMMIT;`

**graphifyy 0.9.33** · tree-sitter-sql 0.3.11 · tree-sitter 0.25.2 · Python 3.13.14 · macOS 27 (arm64)

Related to #1745 (the "install `graphifyy[sql]`" warning). Installing the extra removes the
warning, but SQL files still contribute almost nothing — for a different reason.

## Summary

`CREATE TABLE` statements are extracted correctly on their own, but **the same statements
produce no nodes when the file wraps them in a transaction block**. Since wrapping a
migration in `BEGIN; … COMMIT;` is the normal way to make it atomic, this silently hides
entire schemas.

In our repo it hides **59 `CREATE TABLE` statements across 8 migration files** — every table
the application has. The graph looked healthy (the files themselves appear as nodes), so the
loss is invisible unless you count.

## Minimal reproduction

Two files, identical except for the transaction wrapper.

`sin_transaccion.sql`:

```sql
CREATE TABLE alfa (id integer PRIMARY KEY);
CREATE TABLE beta (id integer PRIMARY KEY);
```

`con_transaccion.sql`:

```sql
BEGIN;

CREATE TABLE gamma (id integer PRIMARY KEY);
CREATE TABLE delta (id integer PRIMARY KEY);

COMMIT;
```

Then:

```
graphify .
```

**Expected:** four table nodes — `alfa`, `beta`, `gamma`, `delta`.

**Actual:** two. Only the ones outside the transaction.

```
  sin_transaccion.sql    'alfa'
  sin_transaccion.sql    'beta'
```

`con_transaccion.sql` appears as a file node, with nothing inside it. No warning, no error.

## What is *not* the cause

I bisected before filing. Each of these works fine on its own — a table node is produced
with correct `source_file` and `source_location`:

| Construct | Extracted? |
|---|---|
| plain `CREATE TABLE` | yes |
| `CREATE TABLE IF NOT EXISTS` | yes |
| leading comment block (incl. non-ASCII) | yes |
| inline `CHECK (col IN (…))` | yes |
| `ALTER TABLE … ENABLE ROW LEVEL SECURITY` between two `CREATE TABLE`s | yes, both |
| `DO $$ … END $$;` before a `CREATE TABLE` | yes |
| **`BEGIN; … COMMIT;` wrapper** | **no** |

So the dialect features aren't the problem; only the transaction block is.

## Secondary oddity: nodes with empty attribution

Running against our real migrations (all of them `BEGIN/COMMIT`-wrapped), three table nodes
*do* survive — `avisos`, `sv_casos`, `sv_motivos_alias` — but with **empty `source_file` and
`source_location`**:

```
label='avisos'            source_file='' source_location='' _origin='ast'
label='sv_motivos_alias'  source_file='' source_location='' _origin='ast'
label='sv_casos'          source_file='' source_location='' _origin='ast'
```

Nodes that survive the same file keep their attribution (`_motivos_nuevos` from a
`CREATE TEMP TABLE … AS` came through with `loc=L68` intact), so this looks like partial
recovery after a failed parse rather than a separate issue. Mentioning it in case it points
at the same code path.

## Why it matters

The failure is silent. `graphify update .` reports a healthy rebuild, the `.sql` files show up
as nodes, and the `#1745` warning is gone — so it reads as "SQL support is working". A query
for a table name still returns hits, but they come from the Python that *mentions* the table,
never from the DDL that *defines* it. The schema is simply absent from the graph, and nothing
says so.

A count of extracted `CREATE TABLE` vs. `CREATE TABLE` occurrences would have caught this, if
that is a cheap check to add.

akshitj11:
## Problem

`CREATE TABLE` statements inside `BEGIN; ... COMMIT;` transaction blocks produced no table nodes. The file appeared in the graph, but every table in a wrapped migration was silently absent.

Minimal reproduction:

```sql
BEGIN;
CREATE TABLE gamma (id integer PRIMARY KEY);
COMMIT;
```

Plain `CREATE TABLE` outside a transaction worked; the same statement inside `BEGIN/COMMIT` did not.

## Cause

`extract_sql` pre-registers defined table names by recursing the full parse tree, but the emitting walk only descended direct children of top-level `statement` nodes. tree-sitter-sql wraps transaction DDL under a `transaction` node containing nested `statement` nodes, so `walk()` never reached those `create_table` nodes.

## Fix

Call `walk(stmt)` on `transaction` nodes at the top level. `walk()` already recurses into children, so nested `statement` / `create_table` nodes are processed like unwrapped DDL.

## Tests

- `tests/fixtures/sample_transaction.sql`: tables inside and outside a transaction block, plus a cross-boundary foreign key.
- `tests/test_multilang.py::test_sql_create_table_inside_transaction_block`

```bash
pytest tests/test_multilang.py -k sql -q
```

Fixes #2953

### DIFF
--- a/graphify/extractors/sql.py
+++ b/graphify/extractors/sql.py
@@ -398,6 +398,10 @@ def extract_sql(path: Path, content: str | bytes | None = None) -> dict:
         if stmt.type == "statement":
             for child in stmt.children:
                 walk(child)
+        elif stmt.type == "transaction":
+            # BEGIN; ... COMMIT; wraps DDL in a transaction node whose children
+            # are statement nodes, not direct create_table nodes (#2953).
+            walk(stmt)
         elif stmt.type in ("fb_proc_or_trigger", "set_term", "declare_external_function", "ERROR"):
             walk(stmt)
 
--- a/tests/fixtures/sample_transaction.sql
+++ b/tests/fixtures/sample_transaction.sql
@@ -1,0 +1,11 @@
+CREATE TABLE alfa (id integer PRIMARY KEY);
+
+BEGIN;
+
+CREATE TABLE gamma (id integer PRIMARY KEY);
+CREATE TABLE delta (
+  id integer PRIMARY KEY,
+  alfa_id integer REFERENCES alfa(id)
+);
+
+COMMIT;
--- a/tests/test_multilang.py
+++ b/tests/test_multilang.py
@@ -461,6 +461,25 @@ def test_sql_finds_tables():
     assert any("users" in l for l in labels)
     assert any("organizations" in l for l in labels)
 
+
+def test_sql_create_table_inside_transaction_block():
+    """#2953: DDL wrapped in BEGIN; ... COMMIT; must emit table nodes."""
+    r = _extract_sql_or_skip("sample_transaction.sql")
+    labels = [n["label"] for n in r["nodes"]]
+    assert any("alfa" in l for l in labels)
+    assert any("gamma" in l for l in labels)
+    assert any("delta" in l for l in labels)
+    refs = [
+        (e["source"], e["target"])
+        for e in r["edges"]
+        if e["relation"] == "references"
+    ]
+    node_by_id = {n["id"]: n["label"] for n in r["nodes"]}
+    assert any(
+        "delta" in node_by_id.get(s, "") and "alfa" in node_by_id.get(t, "")
+        for s, t in refs
+    )
+
 def test_sql_finds_view():
     r = _extract_sql_or_skip()
     labels = [n["label"] for n in r["nodes"]]


safishamsi:
Shipped in v0.9.49 via authorship-preserving cherry-pick so you keep contributor-graph credit. Thanks @akshitj11! Release: https://github.com/Graphify-Labs/graphify/releases/tag/v0.9.49

PR #2993, opened 2026-08-24
fix(bash): emit calls edges for assignment command substitution

### LINKED ISSUE #2978: Bash extractor: calls inside command substitution `x=$(fn)` never get call edges — value-returning helpers look dead
## Summary

The bash extractor never emits a `calls` edge when a defined function is invoked inside command substitution — `x=$(helper)`. This is the idiom for every shell function that *returns a value*, so in helper-library-heavy codebases most function-call sites are invisible to the graph. Every other call context I tested (bare, `||`/`&&` lists, pipelines, `if`/`while` conditions, redirects, `!`, subshells) resolves correctly since #2141 — this is the one remaining hole, and it applies to same-file calls too, not just sourced ones.

## Reproduction (v0.9.48, main `b2cd362`, tree-sitter 0.25.2 / tree-sitter-bash 0.25.1, Linux)

`lib.sh`
```bash
#!/usr/bin/env bash
helper() { echo ok; }
```

`a.sh`
```bash
#!/usr/bin/env bash
source ./lib.sh

bare()      { helper; }                       # 1
subst()     { x=$(helper); echo "$x"; }       # 2  <-- missing
orlist()    { helper || return 1; }           # 3
andlist()   { true && helper; }               # 4
pipe()      { helper | cat; }                 # 5
cond()      { if helper; then :; fi; }        # 6
loop()      { while helper; do break; done; } # 7
redir()     { helper >/dev/null; }            # 8
neg()       { ! helper; }                     # 9
subshell()  { ( helper ); }                   # 10
local_sub() { local y; y=$(helper); }         # 11 <-- missing
```

```
graphify update . --no-cluster
jq -r '.links[] | select(.relation=="calls") | select(.target|endswith("helper")) | .source' graphify-out/graph.json
```

Result: 9 of 11 edges. `subst` and `local_sub` — the two `$(helper)` cases — produce nothing.

Same-file control (no `source` involved) shows the same: with `h2()` defined in the same file, `bare2`/`or2`/`pipe2`/`cond2` get edges, `subst2() { x=$(h2); }` does not. So this is independent of the cross-file resolver.

## Root cause

`graphify/extractors/bash.py` — both `walk()` and `walk_calls()` bail on any `command` node whose ancestor is a `command_substitution` / `process_substitution`:

```python
_BASH_EXPANSION_PARENTS = frozenset({"command_substitution", "process_substitution"})
...
if child.type == "command" and not is_inside_expansion(child):
```

The comment explains the intent: `$(build)` exposes `build` as a child `command` node, and the token-level metacharacter filter can't reject it, so the parent context is used to reject it instead. But the emit site right below already requires `name in defined_functions` — an external command inside `$()` is filtered by that check regardless. The expansion exclusion only ever removes *true* calls to user-defined functions.

## Scale

On a ~117k-line bash ops codebase (913 defined functions, 512 scripts): 4,556 textual `$(fn …)` call sites to defined functions, vs 1,488 `calls` edges emitted in total (per caller→callee pair). Functions that exist to return a value — `token=$(read_bot_token "$dir")`, `id=$(lookup_matter "$name")` — are called from many scripts and have **zero** incoming call edges, so `graphify explain`/`path`/`callers` report them as dead while the `imports_from` edges to the file that defines them resolve fine.

## Proposed fix

Keep `is_inside_expansion` for the *source-path* and *script-runner* detection (where a `$(…)` really is an expansion, not a call), but at the call-edge emit site let `defined_functions` membership win over the expansion check — i.e. a `command` whose name is a known function is a call wherever it sits. That preserves the "external commands fabricate nothing" property from #2141/#2157 (non-function names are still dropped) and also covers `$(...)` inside `if`, pipelines, `local x=$(fn)`, and `$( fn | other )`.

If it helps, the extractor comment at `_BASH_EXPANSION_PARENTS` could note that the exclusion is for path/runner detection only, so the two concerns don't get re-fused later.

Happy to test a patch against the large corpus above.

akshitj11:
## Problem

The bash extractor never emitted `calls` edges when a defined function was invoked inside command substitution, for example `x=$(helper)`. Value-returning helpers looked dead in the call graph. Other contexts (bare call, pipelines, conditionals) worked since #2141.

## Cause

`walk_calls` rejected every command inside `command_substitution` and `process_substitution` via `is_inside_expansion`, including assignment capture `x=$(helper)`.

## Fix

Replace the blanket expansion skip with context-aware rules:

- Process substitutions stay suppressed (`diff <(helper)` unchanged).
- Bare top-level `$(build)` stays suppressed (existing regression test).
- Commands inside `command_substitution` nested under `variable_assignment` now emit `calls` edges.

Argument-position substitution (`echo $(helper)`) is intentionally unchanged for this PR.

## Tests

- `tests/test_extract.py::test_extract_bash_command_substitution_in_assignment_emits_call`
- Existing rejection tests still pass:
  - `test_extract_bash_rejects_command_substitution_as_call`
  - `test_extract_bash_process_substitution_not_recorded`

```bash
pytest tests/test_extract.py -k bash -q
pytest tests/test_symbol_resolution.py -q
```

Fixes #2978

### DIFF
--- a/graphify/extractors/bash.py
+++ b/graphify/extractors/bash.py
@@ -194,6 +194,24 @@ def extract_bash(path: Path) -> dict:
             parent = parent.parent
         return False
 
+    def _bash_call_in_command_allowed(cmd_node) -> bool:
+        """Whether a command node should emit call edges.
+
+        Bare ``$(fn)`` at script level and process substitutions stay suppressed
+        (#2141). Value capture via ``x=$(fn)`` is a real invocation (#2978).
+        """
+        parent = cmd_node.parent
+        saw_command_substitution = False
+        while parent is not None:
+            if parent.type == "process_substitution":
+                return False
+            if parent.type == "command_substitution":
+                saw_command_substitution = True
+            if parent.type == "variable_assignment" and saw_command_substitution:
+                return True
+            parent = parent.parent
+        return not saw_command_substitution
+
     def literal(node) -> str | None:
         # Token-level filter: rejects names containing shell metacharacters.
         # Combined with `is_inside_expansion` for parent-context rejection.
@@ -223,7 +241,7 @@ def extract_bash(path: Path) -> dict:
                 # separately, so we don't attribute their calls to the
                 # enclosing scope.
                 continue
-            if child.type == "command" and not is_inside_expansion(child):
+            if child.type == "command" and _bash_call_in_command_allowed(child):
                 cmd_name_node = child.child_by_field_name("name")
                 if cmd_name_node is None and child.children:
                     cmd_name_node = child.children[0]
--- a/tests/test_extract.py
+++ b/tests/test_extract.py
@@ -2569,6 +2569,41 @@ def test_extract_bash_rejects_command_substitution_as_call(tmp_path):
     assert call_pairs == [], f"Command substitution erroneously emitted call edges: {call_pairs}"
 
 
+def test_extract_bash_command_substitution_in_assignment_emits_call(tmp_path):
+    """#2978: x=$(helper) inside a function must emit a calls edge to helper()."""
+    script = tmp_path / "a.sh"
+    script.write_text(
+        "#!/usr/bin/env bash\n"
+        "helper() { echo ok; }\n"
+        "bare()      { helper; }\n"
+        "subst()     { x=$(helper); echo \"$x\"; }\n"
+        "orlist()    { helper || return 1; }\n"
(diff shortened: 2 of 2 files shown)


safishamsi:
Shipped in v0.9.49 via authorship-preserving cherry-pick so you keep contributor-graph credit. Thanks @akshitj11! Release: https://github.com/Graphify-Labs/graphify/releases/tag/v0.9.49

PR #2994, opened 2026-08-24
fix(csharp): emit generic_arg references for field types

### LINKED ISSUE #2911: C#: generic type arguments are dropped in field and call-site position (captured in property/return/parameter)

akshitj11:
## Problem

C# generic type arguments on fields were dropped. `private Box<IAlpha> _field` emitted a `references[field]` edge to `Box` but not `references[generic_arg]` to `IAlpha`. The same type on a property already emitted both edges.

Call-site generic arguments (`r.Do<IEpsilon>()`) and DI registration (`AddScoped<IZeta, Box<IZeta>>()`) are unchanged in this PR and left for a follow up.

## Cause

The `field_declaration` handler used `_read_csharp_type_name`, which reads only the outermost type name. The `property_declaration` handler beside it already uses `_csharp_collect_type_refs` to recurse into generic arguments.

## Fix

Rewrite the C# `field_declaration` branch to mirror `property_declaration`:

- Walk types with `_csharp_collect_type_refs` and emit `field` / `generic_arg` edges.
- Keep `csharp_field_types` receiver registration via `_csharp_receiver_type_name` (#2299).
- Skip type parameters via the existing `skip` set so `T value` does not fabricate references.

## Tests

- `tests/test_languages.py::test_csharp_field_generic_type_arguments_emit_references`
- `tests/test_languages.py::test_csharp_field_type_parameter_emits_no_reference`

```bash
pytest tests/test_languages.py -k csharp -q
```

Fixes #2911 (field position only)

### DIFF
--- a/graphify/extractors/engine.py
+++ b/graphify/extractors/engine.py
@@ -3687,18 +3687,12 @@ def _extract_generic(
                         type_node = child.child_by_field_name("type")
                         if type_node is not None:
                             break
-            type_info = _read_csharp_type_name(type_node, source)
-            if type_info:
-                type_name, qualified, qualifier = type_info
+            if type_node is not None:
                 csharp_type_params = _csharp_type_parameters_in_scope(
-                    type_node if type_node is not None else node, source
+                    type_node, source
                 )
-                if not type_name or type_name in csharp_type_params:
-                    return
-                # Record the field's declared type for the method-scoped
-                # receiver tables (#2299) — the C# twin of java_field_types.
-                # Pascal-case only: primitives never own a resolvable method.
-                if type_name[:1].isupper():
+                receiver_type = _csharp_receiver_type_name(type_node, source)
+                if receiver_type and receiver_type[:1].isupper() and receiver_type not in csharp_type_params:
                     fields = csharp_field_types.setdefault(parent_class_nid, {})
                     for child in node.children:
                         if child.type != "variable_declaration":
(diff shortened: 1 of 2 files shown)


safishamsi:
Thanks @akshitj11 — this is already fixed. #2913 landed the same C# field-position generic-arg walk in v0.9.49, and #2994's own tests pass against the current tree without this change. Closing as already-fixed; the field-generics work shipped under #2911 via #2913. Release: https://github.com/Graphify-Labs/graphify/releases/tag/v0.9.49

Issue #2550, opened 2026-08-08
Kotlin: fully-qualified call expressions produce no calls edge (same-package control isolates the qualified form)


thomasrengot-hub:
## Summary

A fully-qualified call — `com.demo.a.BetaScreen(2)` — produces no `calls` edge,
while the same call written against an import does. The resolver appears to
consider only the file's import list, so any call written in qualified form is
invisible to the graph.

Measured on a 541-file Kotlin codebase (Android app + Wear module): the
navigation entry point called 43 of its 65 screens in qualified form, so
`explain`/`affected` on that node reported **34% of the real fan-out**. Rewriting
those 43 call sites as imports — no behaviour change, no code deleted — took
recall to 83% (the remaining 17% is a separate parse bug, filed separately).
105 qualified call sites across 54 files remain unresolved elsewhere in the repo.

This is not a Compose or lambda problem: it reproduces with plain top-level
functions and with an `object` member, and it is not about packages either —
the qualified form fails even when caller and callee share a package.

## Reproduction

`graphify 0.9.36`, `graphify extract src --code-only` (no LLM, no API key).

```kotlin
// src/a/Lib.kt
package com.demo.a

fun AlphaScreen(x: Int) { println(x) }
fun BetaScreen(x: Int) { println(x) }
fun GammaScreen(x: Int) { println(x) }
fun DeltaScreen(x: Int) { println(x) }

object Helper { fun help(x: Int) { println(x) } }
```

```kotlin
// src/b/Nav.kt
package com.demo.b

import com.demo.a.AlphaScreen
import com.demo.a.GammaScreen

fun composable(route: String, body: () -> Unit) { body() }

fun NavGraph() {
    composable("alpha") { AlphaScreen(1) }              // plain import
    composable("beta")  { com.demo.a.BetaScreen(2) }    // qualified
    composable("gamma") { GammaScreen(3) }              // plain import
    composable("delta") { com.demo.a.DeltaScreen(4) }   // qualified
    com.demo.a.Helper.help(5)                           // qualified, object member
}
```

```kotlin
// src/b/Same.kt — caller in the SAME package as the callee
package com.demo.a

fun SamePackageCaller() {
    BetaScreen(9)                 // same package, no import needed
    com.demo.a.DeltaScreen(10)    // qualified, same package
}
```

### Observed

All `calls` edges in the resulting graph:

```
b_nav_navgraph              -> b_nav_composable
b_nav_navgraph              -> a_lib_gammascreen        # import
[...]

Issue #1373, opened 2026-06-17
Java Records and constructor call relationships not recognized


rcollette:
**Because the graph build missed the kinds of links that class has, and it also appears to miss the record declaration itself.**

What I found:

- In the graph, the node is just the file node:
  `OidcAuthenticationProperties.java` → **degree 0**
- But in the code, it is clearly used:
  - `OidcJwtAuthenticationConfiguration.java` references it in `@EnableConfigurationProperties(...)`
  - `InteractionCodeTokenHelper.java` constructs it with `new OidcAuthenticationProperties(...)`
  - `OidcAuthenticationPropertiesTest.java` references it multiple times

The likely reasons it ends up isolated are:

1. **Graphify’s Java extraction is mostly structural AST, not full semantic code analysis.**  
   In this run, code relationships were coming from AST extraction, and semantic extraction was only applied to docs/images. So body-level usages like constructor calls and class literals were not added as graph edges.

2. **It appears to miss Java `record` declarations as first-class type nodes.**  
   For `OidcJwtAuthenticationConfiguration.java`, the graph has both the file node and a contained class node. For `OidcAuthenticationProperties.java`, it only has the file node — not a separate type node for the record declaration. That strongly suggests the extractor is not modeling Java records the same way as classes.

3. **Its only import is external framework code.**  
   `OidcAuthenticationProperties.java` imports `ConfigurationProperties`, but external library symbols are not being represented as graph nodes here, so that import does not save it from isolation.

So the short answer is: **your IDE is right; the graph is incomplete here.** The isolation is an extraction limitation, not a real architecture signal.

**What would fix it**
1. Add Java AST support for **record declarations** so the record becomes a proper symbol node.
2. Add extraction of **type references in annotations** like `@EnableConfigurationProperties(OidcAuthenticationProperties.class)`.
3. Add extraction of **constructor/type usages in method bodies** like `new OidcAuthenticationProperties(...)`.
4. Optionally run **semantic extraction on code files too**, not only docs/images.

safishamsi:
Fixed in `895a60f` (branch v8) - your three-part diagnosis was spot on.

Two of the three gaps are addressed:

1. **Records as type nodes** - `record_declaration` was missing from the Java extractor's class types, so a record produced only its file node. It's now treated like `class_declaration` (records share the same name/body/interfaces fields), so a record becomes a first-class type node with a `contains` edge from its file, and `record Foo implements I` emits the `implements` edge.

[...]

PR #2998, opened 2026-08-24
fix(csharp): emit calls edges for object creation expressions

### LINKED ISSUE #2997: C#: `new Foo(...)` produces no edge, so message publishers and locally built types look unused
A type that a C# method only constructs gets no edge. `new Foo(...)` parses as `object_creation_expression`, and the C# config lists only `invocation_expression` in `call_types`, so the construction site is never dispatched in `walk_calls`.

Java has treated `new Foo(...)` as a call since #1373, and TypeScript got `new_expression` extraction in #708. C# never caught up.

### Which positions currently emit an edge

I probed one class with the same type used in seven positions (graphify 0.9.48):

```csharp
public class AsParameter { }
public class AsField { }
public class AsReturn { }
public class AsProperty { }
public class AsLocalDeclared { }
public class AsNewInVar { }
public class AsNewInArg { }

public class Probe
{
    private AsField _field;
    public AsProperty Prop { get; set; }

    public AsReturn Get() => null;
    public void Take(AsParameter p) { }

    public void Body()
    {
        AsLocalDeclared a = null;
        var b = new AsNewInVar();
        Consume(new AsNewInArg());
    }

    private void Consume(object o) { }
}
```

| position | edge |
| --- | --- |
| parameter type | `.Take() references AsParameter` |
| field type | `Probe references AsField` |
| return type | `.Get() references AsReturn` |
| property type | `Probe references AsProperty` |
| local declaration | none |
| `var x = new Foo()` | none |
| `Consume(new Foo())` | none |

The last three classes are in the graph as nodes. Nothing points at them.

### The same code in Java

```java
class Worker { void run() { } }
public class Caller {
    public void go() {
        Worker w = new Worker();
        w.run();
    }
}
```

```
.go() -> Worker  [calls]  EXTRACTED
.go() -> .run()  [calls]  INFERRED
```

The C# translation of that file produces only the second edge.

### Where it hurts

I ran graphify over a private .NET 8 codebase that moves work between services over MassTransit and RabbitMQ. Producers and consumers do not reference each other, so the only thing that can join them in the graph is the message class: the consumer implements `IConsumer<T>` and the producer hands a `T` to the bus.

That works when the producer takes the message as a parameter. Two of the three hops in one chain resolved that way, and both endpoints landed on the same message node. The third hop publishes like this:

```csharp
var events = models.Select(m => new ProductUpsertRequested { Id = m.Id });
await endpoint.Send(new ProductUpsertRequested { Id = id });
```

The message type never appears in a declared position there, so the producer side of that node had only a `contains` edge from its own file. The consumer sat in another service with no path to it. A `graphify path` between the two ends of the chain returns nothing, and the hop looks like dead code.

The pattern is not specific to a bus. Anything a method builds and hands off has the same shape: a request object passed to a client, a config object passed to a factory, a DTO built for a mapper.

### The type is already resolved

This is not a request for new inference. The extractor already types locals well enough to resolve member calls on them:

```csharp
public class Worker { public void Run() { } }
public class Caller {
    public void Go()
    {
        Worker w = new Worker();
        w.Run();
    }
}
```

```
.Go() -> .Run()  [calls]  INFERRED
```

`w` is known to be a `Worker` there. `Worker` itself is simply never named as a target.

Ambiguity is also already handled. With two classes called `Cache` in different namespaces, the construction site resolves to neither, which is the behaviour #437 asks for.

### Scope

Two things nearby that I would leave alone:

- Target-typed `new()` parses as `implicit_object_creation_expression` and carries no type node. Naming it needs the declared type of the assignment target.
- A local declared without a constructor (`AsLocalDeclared a = null;`) still emits nothing. That is a references-from-local-declarations question, separate from this one.

### Related

- #1373 gave Java the same fix, and #708 gave TypeScript `new_expression`.
- #2829 and #2836 walked C# primary-constructor parameters, which is the injected-dependency half of the same problem.
- #437 is the false-edge risk this touches. The ambiguity guard above is why I think it does not apply.
- #1220 would cover the other half of bus wiring, the queue name string itself. Message types alone got me far enough that I did not need the literal index.
- #2911, #1869 and #1591 are the other open C# emission gaps.

I have a patch with tests and will open a PR against `v8`.

durmazoguzhan:
Closes #2997.

`new Foo(...)` parses as `object_creation_expression`. The C# config listed only `invocation_expression` in `call_types`, so `walk_calls` never reached the construction site and a type a method only builds got no incoming edge at all. Java has read `new` as a call since #1373, and TypeScript got `new_expression` in #708.

Two changes:

- `_CSHARP_CONFIG.call_types` gains `object_creation_expression`.
- The C# dispatch in `walk_calls` gets a branch for it that reads the `type` field through `_read_csharp_type_name`, which already strips generic arguments and the namespace qualifier, so `new A.B.Cache<string>()` names `Cache`.

Target-typed `new()` parses as `implicit_object_creation_expression` and stays out of `call_types`. Naming it needs the declared type of whatever it is assigned to, which is a different problem.

### Before and after

```csharp
public class OrderPlaced { public long Id { get; set; } }

public class Publisher
{
    public void Publish(long id) => Send(new OrderPlaced { Id = id });
    private void Send(object payload) { }
}
```

Before, `OrderPlaced` had one edge, `contains` from its own file. After:

```
.Publish() -> OrderPlaced  [calls]
```

which matches what the Java extractor already produces for the same shape.

### Why this is emission and not inference

The extractor already types locals well enough to resolve calls on them. `Worker w = new Worker(); w.Run();` produces `.Go() -> .Run()`, so `Worker` is known at that point and was simply never used as a target.

The ambiguity guard also already holds. Two classes named `Cache` in different namespaces produce no edge from `new Cache()` rather than a coin flip, which is the behaviour #437 asks for. There is a test for it so a later change cannot quietly turn it into a false edge.

### Tests

New file `tests/test_csharp_object_creation.py`, 15 tests: local declaration, `var`, argument position, object initializer without parens, generic construction, qualified construction, cross-file publisher to message class, target-typed `new()` staying silent, the ambiguous-name guard, the qualified-collision resolution, partial qualification staying silent, two candidates in one namespace, and a regression guard for the #1609 receiver-typed member-call path.

Seven fail on `v8` without the source changes.

```
$ uv run pytest tests/test_csharp_object_creation.py -q
15 passed

$ uv run pytest tests/test_extract.py tests/test_languages.py tests/test_csharp_member_calls.py \
    tests/test_csharp_type_resolution.py tests/test_csharp_partial_classes.py \
    tests/test_cross_language_call_resolution.py -q
677 passed
```

Full suite excluding `tests/test_skillgen.py`: 4832 passed, 2 failed, both in `tests/test_falkordb_integration.py` (no FalkorDB server here). `test_skillgen.py` and those two fail identically on a clean `v8` checkout in this environment. `ruff check` is clean on the three files.

### Qualified constructions (added after review)

The review flagged that a qualified `new A.B.Cache()` arrives as the bare name, so a colliding `Cache` elsewhere made it ambiguous and the site lost an edge that the field and return paths do resolve. That is fixed here rather than deferred.

The construction site now carries the qualifier on its raw_call, and a new `csharp_qualified_calls` resolver matches `(namespace, type name)` against sourced type declarations, mirroring `_resolve_kotlin_qualified_calls` from #2550. Exactly one candidate produces the edge, so the pass is additive and the bare-name guard is untouched.

Partial qualification (`new Data.Cache()` under `using Infra;`) stays unresolved. Making that a namespace needs the using directives in scope, and suffix matching would be a guess.

A second review pass found that the dedup check scanned every edge, so a method that both took a type as a parameter and constructed it had the construction skipped. Scoping it to `calls` was not enough on its own: the declared position mints a bare-name stub in the file, the construction bound to that stub, and a bound call produces no raw_call, so the namespace resolver never saw it. A qualified construction whose in-file match has no `source_file` now defers to the cross-file pass, the same `tgt_nid = None` idiom the member-call and Java defers use. Foreign types such as `new System.Text.StringBuilder()` resolve to nothing rather than to a placeholder.

### Left for later

A local declared without a constructor, `Foo a = null;`, still emits nothing. That is a references-from-local-declarations question rather than a construction one, so I kept it out of this change. Happy to follow up if you want it.

### DIFF
--- a/graphify/extract.py
+++ b/graphify/extract.py
@@ -3813,7 +3813,14 @@ def _resolve_csharp_qualified_calls(
     if not by_namespace:
         return
 
-    existing_pairs = {(e.get("source"), e.get("target")) for e in all_edges}
+    # Scoped to `calls`: a method that both takes a type as a parameter and
+    # constructs it already has a `references` edge to it, and that edge says
+    # nothing about whether the construction was resolved.
+    existing_pairs = {
+        (e.get("source"), e.get("target"))
+        for e in all_edges
+        if e.get("relation") == "calls"
+    }
     for rc in raw:
         candidates = by_namespace.get((rc["qualified_prefix"], rc["callee"]), [])
         if len(candidates) != 1:
--- a/graphify/extract.py
+++ b/graphify/extract.py
@@ -3770,6 +3770,72 @@ def _resolve_kotlin_import_targets(
             e["target"] = candidates[0]
 
 
+def _resolve_csharp_qualified_calls(
+    per_file: list[dict],
+    all_nodes: list[dict],
+    all_edges: list[dict],
+) -> None:
+    """Resolve C# constructions that name their namespace (#2997).
+
+    `new Infra.Data.Cache()` reaches the shared pass as the bare name `Cache`,
+    so a second `Cache` in another namespace makes it ambiguous and it gets no
+    edge, even though the source says which one it means. The reference paths
+    (field, property, parameter, return) already honour the qualifier through
+    `CsharpNameResolver`; this is the construction twin, built the way
+    `_resolve_kotlin_qualified_calls` handles the same shape in Kotlin.
+
+    The prefix must equal a declared namespace exactly. A partially qualified
+    `new Data.Cache()` under `using Infra;` stays unresolved rather than
+    guessing at the using directives in scope. Exactly one candidate produces an
+    edge; zero or several leave the call alone. The pass is additive: an
+    ambiguous bare name never had an edge to overwrite.
+    """
+    raw = [
+        rc
+        for result in per_file
+        for rc in result.get("raw_calls", [])
+        if rc.get("lang") == "csharp" and rc.get("qualified_prefix")
+        and rc.get("callee") and rc.get("caller_nid")
+    ]
+    if not raw:
+        return
+
+    # (namespace, type name) -> nids, over sourced type declarations only, so a
+    # sourceless stub minted for a dangling reference cannot win the match.
+    by_namespace: dict[tuple[str, str], list[str]] = {}
+    for n in all_nodes:
+        if not n.get("_callable_class") or not n.get("source_file"):
+            continue
+        namespace = str((n.get("metadata") or {}).get("namespace") or "")
+        label = str(n.get("label", "")).strip("()")
+        if namespace and label:
+            by_namespace.setdefault((namespace, label), []).append(n["id"])
+    if not by_namespace:
+        return
+
+    existing_pairs = {(e.get("source"), e.get("target")) for e in all_edges}
+    for rc in raw:
+        candidates = by_namespace.get((rc["qualified_prefix"], rc["callee"]), [])
+        if len(candidates) != 1:
+            continue
+        caller = rc["caller_nid"]
+        tgt = candidates[0]
+        if tgt == caller or (caller, tgt) in existing_pairs:
+            continue
+        existing_pairs.add((caller, tgt))
+        all_edges.append({
+            "source": caller,
+            "target": tgt,
+            "relation": "calls",
+            "context": "call",
+            "confidence": "EXTRACTED",  # the namespace is written verbatim in source
+            "confidence_score": 1.0,
+            "source_file": rc.get("source_file", ""),
+            "source_location": rc.get("source_location"),
+            "weight": 1.0,
+        })
+
+
 def _resolve_kotlin_qualified_calls(
     per_file: list[dict],
     all_nodes: list[dict],
@@ -3954,6 +4020,14 @@ register_language_resolver(
 register_language_resolver(
     LanguageResolver(
         "kotlin_qualified_calls", frozenset({".kt", ".kts"}), _resolve_kotlin_qualified_calls
+    )
+)
+# C# qualified construction (#2997): `new A.B.Cache()` arrives as the bare name,
+# so a colliding `Cache` elsewhere makes it ambiguous. Runs in the tail registry
+# beside csharp_member_calls and matches the prefix against declared namespaces.
+register_language_resolver(
+    LanguageResolver(
+        "csharp_qualified_calls", frozenset({".cs"}), _resolve_csharp_qualified_calls
     )
 )
 
--- a/graphify/extract.py
+++ b/graphify/extract.py
@@ -928,7 +928,9 @@ _CSHARP_CONFIG = LanguageConfig(
     }),
     function_types=frozenset({"method_declaration"}),
     import_types=frozenset({"using_directive"}),
-    call_types=frozenset({"invocation_expression"}),
+    # `object_creation_expression` joins the invocation node so `new Foo(...)`
+    # links the constructing method to Foo, the way Java has since #1373.
+    call_types=frozenset({"invocation_expression", "object_creation_expression"}),
     call_function_field="function",
     call_accessor_node_types=frozenset({"member_access_expression"}),
     call_accessor_field="name",
--- a/graphify/extractors/engine.py
+++ b/graphify/extractors/engine.py
@@ -5027,6 +5027,21 @@ def _extract_generic(
                                 if child.type == "identifier":
                                     callee_name = _read_text(child, source)
                                     break
+            elif config.ts_module == "tree_sitter_c_sharp" and node.type == "object_creation_expression":
+                # `new Foo(...)` keeps the constructed type in the `type` field, so
+                # the invocation path below never sees it and a type a method only
+                # constructs stays unlinked — the C# twin of the Java gap in #1373.
+                # Types reached solely through a method body are exactly the ones
+                # this misses: message classes handed straight to a bus
+                # (`Send(new OrderPlaced { ... })`) and locally built collaborators.
+                # `_read_csharp_type_name` drops the generic arguments and the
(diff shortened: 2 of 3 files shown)


durmazoguzhan:
The qualified-construction finding reproduces, so thanks for it. Two `Cache` classes, one construction site that spells out the namespace:

```csharp
// Infra/Cache.cs
namespace Infra.Data;
public class Cache { }

// Other/Cache.cs
namespace Other;
public class Cache { }

// Caller.cs
public class Caller { public void Go() { var c = new Infra.Data.Cache(); } }
```

No edge. The same qualified type in a field or a return position does resolve, and to the right node:

```
private Infra.Data.Cache _c;          ->  Holder references Cache   (infra_cache_infra_data_cache)
public Infra.Data.Cache Make()        ->  .Make() references Cache  (infra_cache_infra_data_cache)
var c = new Infra.Data.Cache();       ->  no edge
```

So it is a missed edge rather than a wrong one, and it is an inconsistency between the reference paths and the new construction path, not a deliberate rule.

I measured how much it costs before deciding what to do about it. On the corpus that motivated #2997 (7 .NET services, roughly 840k lines of C#), 133 of 3344 constructions are fully qualified, and 101 of those 133 are `new System.NotImplementedException()` and friends, which have no node to point at. After removing type names that do not collide, 9 construction sites in the whole corpus would gain an edge from qualified resolution.

Given that, I would rather not grow this PR. The mechanism for the fix is not in place for C#: `qualified_prefix` on a raw_call is consumed only by `_resolve_kotlin_qualified_calls`, which is built on the Kotlin package index, and the C# side resolves qualified names through `CsharpNameResolver.resolve_qualified` on the reference path instead. A C# twin of the Kotlin pass is the natural shape and it wants its own tests, so it reads better as a follow-up than as a second concern bolted onto this one. #1869 asks for self-contained commits for the same reason.

What I did add is 18b1777, a test that pins the current behaviour with a comment saying it is a gap rather than a rule, so the case cannot change silently in either direction.

Happy to do the follow-up either way. If you would rather have it inside this PR, say so and I will wire it up here.

durmazoguzhan:
Fixed in 10e52db, so the finding is closed rather than deferred. It turned out cheaper than I estimated once I looked at how Kotlin handles the same shape.

The construction site now keeps the qualifier on its raw_call, and a new `csharp_qualified_calls` resolver matches `(namespace, type name)` against sourced type declarations, mirroring `_resolve_kotlin_qualified_calls` from #2550. Behaviour on the four cases:

```
new Infra.Data.Cache()   + a second Cache in Other   ->  edge to Infra.Data.Cache   EXTRACTED
new Data.Cache()         + using Infra;              ->  no edge
new Cache()              + a second Cache            ->  no edge (guard unchanged)
new Infra.Data.Cache()   + no collision              ->  edge, as before
```

Partial qualification stays out on purpose. Turning `Data.Cache` into a namespace needs the using directives in scope, and matching by suffix would be the kind of guess #437 warns about.

The three tests that pinned the old behaviour are replaced by the positive case, the partial-qualification case, and two candidates inside a single namespace still hitting the guard. Suite is 4883 passed, 11 failed, and those 11 are the `tests/test_skillgen.py` failures that reproduce on a clean `v8` checkout here without my changes. `ruff check` clean on all three files.

durmazoguzhan:
Right again, and it went deeper than the dedup scope when I went looking. Fixed in 3fd21ff.

Reproduction, two `Cache` classes in different namespaces and a method that both takes one as a parameter and constructs one:

```csharp
public class Caller {
    public void Go(Infra.Data.Cache existing) { var c = new Infra.Data.Cache(); }
}
```

```
.Go() -references-> left_infra_data_cache     (the parameter, correct)
.Go() -calls->      cache                     (sourceless placeholder)
```

Scoping the check to `calls` edges did not change that, which is what sent me looking further. The declared position mints a bare-name stub in the file, the construction binds to that stub, and a bound call produces no raw_call at all, so `_resolve_csharp_qualified_calls` never saw it. The finding is real but the mechanism was one layer down.

So a qualified construction whose in-file match has no `source_file` now defers to the cross-file pass, using the same `tgt_nid = None` idiom as the C# member-call and Java defers a few lines above. Both changes are in: the dedup is scoped to `calls`, and the placeholder case defers.

Five cases, all as I would want them now:

```
new Infra.Data.Cache()                              ->  Infra.Data.Cache
new Infra.Data.Cache() + Cache parameter            ->  Infra.Data.Cache   (was: placeholder)
new Infra.Data.Cache() + unrelated parameter        ->  Infra.Data.Cache
new Data.Cache() under `using Infra;`               ->  no edge
new System.Text.StringBuilder()                     ->  no edge
```

The last two produce nothing rather than an edge into a placeholder that stands for nothing, which is where they were before constructions became calls at all.

On a 1457-file service this moves 11 call edges off placeholders onto real declarations, with the edge total unchanged at 30852.

Two tests added: the parameter-plus-construction case asserting both the right target and no sourceless target, and the foreign-type case asserting no stub edge. `tests/test_csharp_object_creation.py` is 15 tests, extraction suites 682 passed.

Full suite excluding `tests/test_skillgen.py`: 4832 passed, 2 failed, both in `tests/test_falkordb_integration.py`, which fail the same way on a clean `v8` checkout here since there is no FalkorDB server in this environment. `test_skillgen.py` also fails on clean `v8` here, which is why it is out of that run.

safishamsi:

PR #2999, opened 2026-08-24
fix(zig): extract methods declared on enums and unions

rajatnagda45:
## Problem

The Zig extractor only recursed into `struct` containers when looking for member functions. Zig routinely attaches methods to `enum`s and tagged `union`s (`pub fn ... { }` inside the container), but those declarations were dropped entirely — and with them **every `calls` edge originating from those method bodies**. For any enum/union with behaviour, a whole slice of the method and call graph silently disappeared.

Structs handled this correctly; enums and unions did not, despite the grammar nesting `function_declaration` children identically in all three.

## Fix

Recurse into enum/union declaration children the same way struct declarations already do, so their methods (and the calls they make) are captured.

## Test

Adds a regression test covering an enum method, a tagged-union method, and a call made from a union method body. Fails before the change, passes after; full language suite green.

### DIFF
--- a/graphify/extractors/zig.py
+++ b/graphify/extractors/zig.py
@@ -124,6 +124,12 @@ def extract_zig(path: Path) -> dict:
                     type_nid = _make_id(stem, type_name)
                     add_node(type_nid, type_name, line)
                     add_edge(file_nid, type_nid, "contains", line)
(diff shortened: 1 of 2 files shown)


safishamsi:

PR #3000, opened 2026-08-24
fix(julia): extract abstract types declared with a supertype

rajatnagda45:
## Problem

`abstract type Foo <: Bar end` was dropped completely — no node, no `defines` edge, and no `inherits` edge. The abstract-type path only matched a bare `identifier` in the `type_head`, but the subtyping form wraps both names in a `binary_expression` (the exact shape the struct path already handles).

Abstract types anchor Julia's multiple-dispatch hierarchies, so an intermediate `Foo <: Bar` vanishing severed the inheritance chain and lost the type node itself. A concrete `struct Cat <: Animal` resolved its supertype fine; the abstract equivalent did not.

## Fix

Factor the shared `type_head` parsing (bare name vs. `<:` form) into one helper used by both the struct and abstract-type paths, and emit the `inherits` edge for abstract subtypes.

## Test

Adds a regression test for `abstract type Dog <: Animal end` asserting both the node and the `inherits` edge. The five pre-existing Julia hierarchy tests still pass, confirming the struct path is unchanged by the refactor. Fails before the change, passes after.

### DIFF
--- a/graphify/extractors/julia.py
+++ b/graphify/extractors/julia.py
@@ -84,6 +84,26 @@ def extract_julia(path: Path) -> dict:
             })
         return nid
 
+    def _type_head_names(type_head) -> tuple[str | None, str | None]:
+        """Return (type_name, supertype_name) from a Julia `type_head`.
+
+        A bare declaration (`Foo`) exposes an `identifier`; a subtyping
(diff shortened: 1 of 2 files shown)


safishamsi:

PR #3001, opened 2026-08-24
fix(commonlisp): keep inherits edge when the superclass is in another file

rajatnagda45:
## Problem

A `defclass` superclass was resolved to a file-scoped id (`stem::name`) with no backing node. When the parent class was defined in the **same** file that id matched its node and the `inherits` edge held — but when the parent lived in **another** file the id resolved to nothing and the dangling-edge filter pruned the edge. This silently erased cross-file CLOS class hierarchies, which routinely span files.

Every other inheritance-aware extractor (Julia, Go, Fortran, PowerShell) mints a *sourceless* stub for a not-locally-defined base so the corpus-level rewire can collapse it onto the real definition. Common Lisp was the outlier.

## Fix

Resolve a superclass to its local node when it is defined in-file, otherwise mint a sourceless stub (mirroring `add_import_stub` and the other extractors) so the corpus rewire can collapse it onto the real `defclass`.

## Test

Adds a regression test covering both a same-file parent (`dog` → `animal`, binds locally) and a cross-file parent (`service-dog` → `base-animal`, survives as a sourceless stub). All 25 existing Common Lisp tests still pass. Fails before the change, passes after.

### DIFF
--- a/graphify/extractors/commonlisp.py
+++ b/graphify/extractors/commonlisp.py
@@ -154,6 +154,31 @@ def extract_commonlisp(path: Path) -> dict:
             })
         return nid
 
+    def ensure_class_ref(name: str, line: int) -> str:
+        """Resolve a superclass name to a node id.
+
+        If the class is defined in THIS file, bind to its local node. Otherwise
+        mint a SOURCELESS stub (like `add_import_stub`) so the corpus-level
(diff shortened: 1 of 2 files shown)


safishamsi:

PR #3002, opened 2026-08-24
fix(powershell): extract enum definitions and their members

rajatnagda45:
## Problem

PowerShell enums are first-class type definitions and are referenced via `[Color]` type literals, but the extractor handled only `class` — not `enum`. So an enum was never emitted as a definition: its `[Color]` references resolved to a **sourceless phantom stub** instead of the real in-file type, and its members were lost entirely.

## Fix

Handle `enum_statement` the same way `class_statement` is handled: emit the enum as a `contains` definition off the file, and each `enum_member` as a contained node. Type references now bind to the real enum node instead of a phantom.

## Test

Adds a regression test covering the enum definition (a real sourced node, not a phantom), its members, and a `[Color]` field reference resolving to the enum. All 22 existing PowerShell tests still pass. Fails before the change, passes after.

### DIFF
--- a/graphify/extractors/powershell.py
+++ b/graphify/extractors/powershell.py
@@ -127,6 +127,32 @@ def extract_powershell(path: Path) -> dict:
                     walk(body, parent_class_nid)
             return
 
+        if t == "enum_statement":
+            # PowerShell enums are first-class type definitions (referenced via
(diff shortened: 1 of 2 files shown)


safishamsi:

Issue #437, opened 2026-04-18
Cross-file inference creates false edges via common .NET method names


V0v1kkk:
## Problem

The cross-file call resolution in `extract.py` matches unresolved callee names against any node label in the graph. Common .NET BCL method names (`Contains`, `Equals`, `ToString`, `Where`, `Select`, `Add`, `Remove`, etc.) match unintended nodes — for example, `string.Contains()` calls match F# union case `EdgeRelation.Contains`, or `list.Add()` matches a user-defined `Add` method in a completely unrelated module.

This creates hundreds of false `INFERRED` edges, pollutes community detection, and produces misleading god nodes.

## Proposed Fix

Add a blocklist of ~120 common .NET BCL/framework method names that are skipped during cross-file call resolution. This is the same pattern used by the existing `_SKIP_LABELS` set but targeted at method names rather than module names.

## Impact

Tested on a mixed F#/C#/Blazor project:
- False INFERRED edges dropped from 538 to 362
- Top hub changed from `Contains` (false positive) to `BookService` (correct)

I have a PR ready for this.

singhsatyampratap:
Confirming this on **Python + JavaScript** (0.9.34), and flagging that the proposed remedy — a
~120-name .NET BCL blocklist — would not have caught my case, because the colliding names are all
**user-defined**.

Corpus: a mixed Python/JS monorepo, three sibling applications under one directory, no shared code
layer between them by design. Extraction: `--code-only --no-cluster` (no LLM), 527 files, 6831 nodes,
15811 edges.

Of the **40** edges the graph drew *between* the three applications, **37 were `INFERRED` and all 37
were false**. Representative, with names generalised:

```
app_a/main.py            .update_file_list()  ->  .cmap()    [app_c/render.py]
app_a/panels/nouns.py    make_tools_panel()   ->  .load()    [app_b/scripts/extract.py]
app_a/types/raster.py    .plot()              ->  norm()     [app_b/scripts/extract.py]
```

Verified mechanism on one of them: a function `get_cmap_color(cmap_name)` in app A was linked to a
property `def cmap(self)` in app C. The only thing they share is the string `cmap`. Same for `.load()`,
`.update()`, `.toggle()`, `.norm()` — ordinary domain vocabulary, not framework methods, so no
BCL/stdlib blocklist reaches them.

Two observations that may be useful for the fix:

1. **The false edges concentrate exactly where they do the most damage.** Intra-application edges were
[...]

PR #3005, opened 2026-08-24
feat(csharp): link interface methods to their single implementation

### LINKED ISSUE #3003: C#: a call through an interface dead-ends at the interface method, so injected dependencies cut every chain
A C# call through an interface lands on the interface's method node and stops there. The implementation is a separate node with nothing joining the two, so any chain that passes through a constructor-injected dependency is cut at that point.

Three files:

```csharp
// IReport.cs
public interface IReport { void Build(); }

// Report.cs
public class Report : IReport { public void Build() { Format(); } public void Format() { } }

// Runner.cs
public class Runner {
    private readonly IReport _report;
    public Runner(IReport report) { _report = report; }
    public void Go() { _report.Build(); }
}
```

Graph (graphify 0.9.48, edges other than `contains`):

```
IReport   -> .Build()   [method]
Report    -> IReport    [implements]
Report    -> .Build()   [method]
Report    -> .Format()  [method]
.Build()  -> .Format()  [calls]        <- source: report_report_build
.Go()     -> .Build()   [calls]        <- target: ireport_ireport_build
```

Two `.Build()` nodes, `ireport_ireport_build` and `report_report_build`, and no edge between them. `Go` calls the interface one, `Format` hangs off the implementation one, so a directed walk from `.Go()` never reaches `.Format()`. The receiver typing is doing its job here: `_report` is correctly typed as `IReport`, and the interface really is what the call site names.

### How much it costs on a real codebase

A .NET 8 service, 1457 C# files, constructor injection throughout (Scrutor assembly scanning, so every dependency is an interface):

- 1511 `calls` edges land on a method owned by an interface and dead-end there
- 115 interface types have at least one implementer
- 107 of those 115 have exactly one implementer

So on this corpus the ambiguous case is 8 types out of 115. The rest have one possible target each.

`graphify path` between two ends of such a chain returns "No directed path found", and `affected` stops at the interface, which makes a blast radius look far smaller than it is.

### What other languages already do here

Pascal has the closest analogue, and its docstring states the principle I want to borrow:

> walking `inherits` is a structurally justified resolution, not a heuristic guess; guessing by name across an entire multi-thousand-file corpus is not the same bet.

`resolve_pascal_inherited_calls` walks the caller's `inherits` chain across files and emits an edge only when exactly one class at the nearest matching level owns a same-named method. `resolve_ruby_member_calls` uses the same single-owner guard. Interface dispatch in C# is the same kind of bet in the other direction: instead of walking up to find where a method is declared, walk down from a declaration to the one type that implements it.

### Proposed shape

A `csharp_interface_dispatch` resolver in the tail registry that adds a member-level edge from an interface method to the implementing method:

```
ireport_ireport_build  --dispatches_to-->  report_report_build
```

Guards, matching the existing resolvers:

- the interface has exactly one implementer (`implements` edges into it)
- that implementer owns exactly one method of the same name
- zero or several candidates at either step leaves the pair alone

Confidence `INFERRED`, not `EXTRACTED`. The target is structurally forced when there is one implementer, but the source text does not name it, and that is what the existing tiers mean.

On the corpus above this emits 862 edges, +2.8% on 30852, and skips nothing as ambiguous. It reconnects the 1511 dead-ending call edges through 862 links rather than 1511, because the join is per method rather than per call site.

### Alternatives I ruled out

- Retarget the existing call from the interface method to the implementation. Loses the fact that the call site names the interface, which is true and worth keeping.
- Add `caller -> implementation` `calls` edges. 1511 new edges on that corpus, and it inflates every caller's call list with a target the source never mentions.
- Reuse `implements` at member level. Its type-level direction is implementation to interface, and a directed walk needs the opposite direction to be useful, so overloading the name would make traversal read backwards.

Open question: `dispatches_to` is my guess at a name that fits the existing vocabulary. Happy to use whatever you prefer.

### Related

- #355 proposes a full C# type registry that resolves receivers through fields, properties, locals, `this`, `base` and inheritance chains. It is open at +9900/-593 and covers much more than this. This issue is deliberately one narrow slice with one guard, so it can be reviewed on its own.
- #2829 and #2836 made constructor-injected dependencies produce edges at the type level, which is what makes the interface method the reachable target today.
- #1699 (Kotlin), #2860 (Python) and #2234 (Rust) are the same family of missing member-call resolution in other languages.
- #437 is the false-edge concern. One implementer means one possible target, and the guard drops anything ambiguous.
- #2997 is the other half of what broke tracing on this codebase.

I have a patch with tests and will open a PR against `v8`.

durmazoguzhan:
Closes #3003.

A C# call through an injected dependency lands on the interface's method node, because that is what the call site names. The implementation is a separate node with nothing joining the two, so a directed walk stops there:

```
.Go()     -> .Build()   [calls]      target: ireport_ireport_build
.Build()  -> .Format()  [calls]      source: report_report_build
```

Two `.Build()` nodes, no edge between them, so `.Go()` never reaches `.Format()`. On a service that injects every dependency as an interface, that is most chains, and `affected` reports a blast radius smaller than the real one.

`resolve_csharp_interface_dispatch` adds a member-level edge:

```
ireport_ireport_build  --dispatches_to-->  report_report_build
```

### Guards

- the interface has exactly one implementer
- that implementer owns exactly one method of the same name
- zero or several candidates at either step leaves the pair alone

Walking `implements` to the only type that can serve the call is the same bet `resolve_pascal_inherited_calls` describes for walking `inherits`: structurally justified rather than a name guess. Ambiguity produces nothing, which is also what #437 asks for.

Confidence is `INFERRED`. The target is forced once there is a single implementer, but the source text never names it.

The pass is additive. The call edge to the interface method stays, since the call site really does name the interface, and the join is per method rather than per call site.

### Measured

A .NET 8 service, 1457 C# files, Scrutor assembly scanning. Same checkout, resolver registration on and off:

```
without:  9925 nodes, 29547 edges
with:     9925 nodes, 30409 edges     (+862, +2.9%)
```

862 edges give 1511 interface-method call edges a route to an implementation. 115 interface types have implementers there and 107 have exactly one, so the guard drops 8.

### Review fixes

Two passes of findings, all four real:

- The member match folded case, copied from the Pascal resolver where it belongs because Pascal is case insensitive. C# is not, so it is case sensitive now.
- The already-emitted check scanned every edge, so another relation between the two members suppressed the link. Scoped to `dispatches_to`.
- `implements` resolves by name, so a Java class declaring `implements IReport` binds to a C# `IReport` and the resolver linked the C# member to the Java method. Every end of a pair now has to be a `.cs` declaration.
- A non-C# implementation of a C# interface, from VB or a Razor `@implements`, is real but stays out: claiming one should rest on that extractor's evidence rather than a cross-language name match.

- The member name was taken by stripping a trailing `()`. That holds for every label the extractor emits today (3674 method nodes on that service, none outside the `.Name()` shape, explicit interface implementations and generic methods included), but the match is keyed on that string, so it now reads up to the first parenthesis and a signature-carrying label would keep matching instead of silently missing.

The corpus count is 862 edges before and after all five, which is what I would expect if nothing real was riding on the loose ends.

Three of the five landed on one theme: the resolver's shape is lifted from `resolve_pascal_inherited_calls`, and assumptions that hold for Pascal (case insensitive names, a `calls`-only edge space) do not hold for C#.

### Out of scope, both pinned by tests

- `class Impl : Base` on an abstract class is an `inherits` edge, and a base method may well be the real target, so walking inheritance is a different bet.
- An interface hierarchy (`IBase` <- `IChild` <- `Impl`) is not walked transitively. `IBase`'s only implementer is `IChild`, which declares nothing, so nothing links. I found no case of it in the corpus above, so I left the chain walk out rather than write code with no evidence behind it.

### Tests

`tests/test_csharp_interface_dispatch.py`, 15 tests: the link itself, the chain becoming reachable, the interface call being kept, two implementers producing nothing, a missing member producing nothing, collapsed overloads still linking (the guard counts distinct nodes, not `method` edge multiplicity, same as the Pascal dedup comment), abstract base classes staying out, interface hierarchies staying out, and a Java corpus gaining no edge.

Three fail on `v8` without the registration.

```
$ uv run pytest tests/test_csharp_interface_dispatch.py -q
15 passed

$ uv run pytest tests/ -q --ignore=tests/test_skillgen.py
4826 passed, 11 skipped
```

`tests/test_skillgen.py` fails here on a clean `v8` checkout as well, before any of my changes, so I left it out of that run rather than report failures that are not mine. `ruff check` is clean on the three files.

### Open question

`dispatches_to` is my guess at a name that fits the existing vocabulary. I did not reuse `implements` at member level because its type-level direction is implementation to interface, and a directed walk needs the opposite to be useful. Happy to rename.

### DIFF
--- a/graphify/csharp_dispatch.py
+++ b/graphify/csharp_dispatch.py
@@ -31,6 +31,22 @@ _CSHARP_SUFFIXES = (".cs",)
 DISPATCH_RELATION = "dispatches_to"
 
 
+def _is_csharp(node: dict | None) -> bool:
+    """True when the node is a declaration that lives in a C# file.
+
+    Every end of a dispatch pair has to pass this. `implements` is resolved by
+    name, so in a mixed corpus a Java class declaring `implements IReport` binds
+    to a C# `IReport` when that is the only node with the name, and linking a C#
+    interface member to a Java method would be a wrong edge rather than a missing
+    one. A non-C# implementation of a C# interface, from VB or a Razor component,
+    is a real thing, but claiming it needs its own extractor evidence.
+    """
+    if not node:
+        return False
+    source_file = node.get("source_file")
+    return bool(source_file) and str(source_file).endswith(_CSHARP_SUFFIXES)
+
+
 def _method_label(node: dict) -> str:
     """Return a method node's bare name, for matching.
 
@@ -71,7 +87,7 @@ def resolve_csharp_interface_dispatch(
         elif rel == "method":
             owner, method_nid = e.get("source"), e.get("target")
             mnode = node_by_id.get(method_nid)
-            if mnode is None:
+            if mnode is None or not _is_csharp(mnode):
                 continue
             name = _method_label(mnode)
             if name:
@@ -99,11 +115,10 @@ def resolve_csharp_interface_dispatch(
         impl_node = node_by_id.get(impl_nid)
         if interface_node is None or impl_node is None:
             continue
-        # Both ends must be real declarations: a sourceless stub minted for a
-        # dangling reference carries no members worth dispatching to.
-        if not interface_node.get("source_file") or not impl_node.get("source_file"):
-            continue
-        if not str(interface_node.get("source_file", "")).endswith(_CSHARP_SUFFIXES):
+        # Both ends must be C# declarations. A sourceless stub minted for a
+        # dangling reference carries no members worth dispatching to, and a
+        # cross-language pair is a name collision rather than an implementation.
+        if not _is_csharp(interface_node) or not _is_csharp(impl_node):
             continue
 
         impl_methods = methods_of.get(impl_nid, {})
--- a/graphify/csharp_dispatch.py
+++ b/graphify/csharp_dispatch.py
@@ -53,8 +53,14 @@ def _method_label(node: dict) -> str:
     Case is kept: C# is case sensitive, and an implementing member must spell the
     interface member exactly, so folding case could only pair a declaration with
     a member that does not implement it.
+
+    The name is cut at the first parenthesis rather than by stripping a trailing
+    `()`. Today the C# extractor labels every method `.Name()`, so the two are the
+    same, but the pair match is keyed on this string and a label that ever carried
+    a signature would silently stop matching instead of failing visibly.
     """
-    return str(node.get("label", "")).strip().removeprefix(".").removesuffix("()")
+    label = str(node.get("label", "")).strip().removeprefix(".")
+    return label.split("(", 1)[0]
 
 
 def resolve_csharp_interface_dispatch(
--- a/graphify/csharp_dispatch.py
+++ b/graphify/csharp_dispatch.py
@@ -32,8 +32,13 @@ DISPATCH_RELATION = "dispatches_to"
 
 
 def _method_label(node: dict) -> str:
-    """Normalize a method node's label to its bare name for matching."""
-    return str(node.get("label", "")).strip().removeprefix(".").removesuffix("()").lower()
+    """Return a method node's bare name, for matching.
+
+    Case is kept: C# is case sensitive, and an implementing member must spell the
+    interface member exactly, so folding case could only pair a declaration with
+    a member that does not implement it.
+    """
+    return str(node.get("label", "")).strip().removeprefix(".").removesuffix("()")
 
 
 def resolve_csharp_interface_dispatch(
@@ -77,7 +82,13 @@ def resolve_csharp_interface_dispatch(
     if not implementers or not methods_of:
         return
 
-    existing_pairs = {(e.get("source"), e.get("target")) for e in all_edges}
+    # Scoped to this relation: another edge between the two members, whatever it
+    # is, says nothing about whether the dispatch link is already there.
+    existing_pairs = {
+        (e.get("source"), e.get("target"))
+        for e in all_edges
+        if e.get("relation") == DISPATCH_RELATION
+    }
     new_edges: list[dict] = []
 
     for interface_nid, impls in implementers.items():
--- a/graphify/csharp_dispatch.py
+++ b/graphify/csharp_dispatch.py
@@ -1,0 +1,122 @@
+"""Member-level interface dispatch for C# (#3003).
+
+A C# call through a constructor-injected dependency lands on the interface's
+method node, because that is what the call site names: `_report.Build()` where
+`_report` is an `IReport` resolves to `IReport.Build()`. The implementing
+`Report.Build()` is a separate node, and nothing joins the two, so a directed
+walk stops at the interface and every chain through an injected dependency is
+cut at that point. On a Scrutor-scanned .NET service where every dependency is
+an interface, that is most chains.
+
+This resolver runs after all files are extracted, with the merged corpus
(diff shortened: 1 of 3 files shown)


durmazoguzhan:
Both findings were right. Fixed in the commit above.

The case fold came from `resolve_pascal_inherited_calls`, which I used as the model for this resolver. It belongs there, since Pascal is case insensitive, and it does not belong here: a C# member that implements an interface member spells it exactly, so folding case could only pair a declaration with something that does not implement it. The corpus count is unchanged at 862 edges, which is the expected result if no real implementation was ever relying on the fold.

The already-emitted check scanned every edge rather than `dispatches_to` edges, so any other relation between the two member nodes would have suppressed the link. It is scoped to the relation now.

Three tests added: a case-only difference producing nothing, an unrelated edge between the two members not suppressing the link, and an existing dispatch edge not being duplicated. `tests/test_csharp_interface_dispatch.py` is 12 tests, and the extraction suites are 679 passed.

durmazoguzhan:
The mixed-corpus finding reproduces, and it produces a wrong edge rather than a missing one, so thanks for pushing on it. Fixed in the commit above.

`implements` is resolved by name, so with one `IReport` in the corpus and it being C#, a Java class binds to it:

```
Report (Report.java)   -implements->     IReport (IReport.cs)
.Build() (IReport.cs)  -dispatches_to->  .Build() (Report.java)
```

Every end of a pair now has to be a declaration in a `.cs` file: the interface, the implementer, and both member nodes. I had checked the interface side only, which is why a Java-only corpus was already clean and this case was not.

On the other half of the finding, a non-C# implementation of a C# interface is a real thing in .NET, from VB or a Razor component with `@implements`. I left those out rather than in: claiming one should rest on that extractor's own evidence, not on a name match that happens to cross languages.

Corpus count is unchanged at 862 edges, which is what I would expect if nothing real was riding on the loose ends.

Two tests added: a C# interface with a Java implementer producing nothing, and a mixed corpus linking only the C# pair. `tests/test_csharp_interface_dispatch.py` is 14 tests, extraction suites 681 passed.

durmazoguzhan:
This one does not reproduce on C# today, but it is a fair thing to point at, so I hardened it anyway and checked rather than assumed.

What the extractor actually labels, on a 1457-file service: 3674 method nodes, none of them outside the `.Name()` shape. The two cases where I expected a signature to leak in both come out as plain names and already linked:

```
void IReport.Build() { }     ->  .Build()   (explicit interface implementation)
public void Build<T>() { }   ->  .Build()   (generic method)
```

An indexer produces no method node at all, so there is nothing to match there either.

Still, the pair match is keyed on that string on both ends, so a label that ever carried a signature would silently stop matching instead of failing visibly. The name is now read up to the first parenthesis rather than by stripping a trailing `()`, with a unit test on the helper. No edge changes: the corpus count is 862, the same as through the previous three rounds of fixes.

That is the fourth round on this PR and the fifth finding overall. For what it is worth, the three that were real (case folding, the dedup scope, and the cross-language pair) all landed on the same theme: I lifted the shape from `resolve_pascal_inherited_calls` and inherited assumptions that hold for Pascal and not for C#. Worth keeping in mind if another language resolver gets written from that template.

safishamsi:

Issue #296, opened 2026-04-13
Post-extraction node canonicalization to merge duplicate concepts


adelaidasofia:
### Problem

When extraction runs per-file (which is the default for large vaults), agents emit file-scoped node IDs to avoid collisions (e.g., `2024-03_silvia`, `journal_silvia`, `crm_silvia`). These all refer to the same concept but appear as separate nodes. On our vault, this caused **62% node bloat** (1,421 raw nodes collapsed to 548 canonical nodes on one batch).

### Proposed solution

A canonicalization pass that runs after extraction and merge:

1. **Label normalization:** Lowercase, strip path prefixes, remove known suffix patterns (e.g., folder qualifiers that agents add for disambiguation).
2. **Canonical ID generation:** Slugify the normalized label into a stable ID (alphanumeric + underscores, length-capped).
3. **Node merge:** All nodes sharing the same canonical ID collapse into one. Metadata is merged (source files become a list, `mention_count` tracks frequency).
4. **Edge dedup:** Deduplicate edges by `(source, target, relation)` triplet after remapping to canonical IDs.
5. **Validation:** Filter self-loops and any edges that became dangling after the merge.

### Why this matters

Without canonicalization, the graph grows linearly with file count even when the same concepts appear repeatedly. Community detection splits what should be one cluster into many fragments. God-node analysis misses the true hubs because their mentions are scattered across aliases.

On journal-heavy corpora, we see 34% canonicalization reduction per batch (the highest of any content type, because journal entries reference the same themes with slight phrasing variations).

### Scope

This could be a `graphify canonicalize` subcommand or a `--canonicalize` post-processing flag. The normalization rules would need to be configurable since different vault types have different naming conventions.

Happy to send a PR.

yotamleo:
Independent real-world confirmation of this exact mechanism, on a 14,505-note Obsidian vault.

Per-file extraction minted **39 `@handle`-vs-bare-name twin pairs** and **12 curated-vs-importer duplicate groups** — e.g. `@cyrilxbt`, `cyrilXBT`, and a hand-curated `@cyrilXBT.md` all coexisting as separate nodes, differing only by case and a leading `@`. Two of those groups are **byte-identical filename collisions across two directories**, which is not just node bloat: it's a live Obsidian link-resolution ambiguity, so `[[name]]` resolves non-deterministically for the human reader too.

Scale of the output: the importer's own output directory now holds **7,802 of the vault's 14,505 total notes** (~54%).

[...]

Issue #207, opened 2026-04-11
Add entity canonicalization/merge step to prevent duplicate nodes across files


yanikengelhardt:
## Problem
  Graph construction currently treats extracted nodes as unique by technical ID, which is often file/chunk-
  local.
  As a result, the same real-world concept/entity can appear as multiple graph nodes, causing fragmentation and
  noisy analytics.

  ## Symptoms
  - Duplicate semantic nodes with different IDs.
  - Alias/casing/punctuation variants not merged.
  - Related neighborhoods split across disconnected components.
  - Distorted centrality, clustering, and query results.

  ## Expected behavior
  Graph should support a canonical entity layer so semantically identical nodes are merged into one canonical
  node while preserving provenance.

  ## Proposed solution
  Introduce an **Entity Resolution** phase after extraction and before clustering/export:

  1. Normalize candidate node labels (casefold, unicode normalization, punctuation/whitespace cleanup).
  2. Resolve aliases via configurable dictionary/rules.
  3. Optionally compute similarity-based candidates with conservative thresholds.
  4. Build canonical IDs and merge equivalent nodes.
  5. Rewire edges/hyperedges to canonical nodes.
  6. Preserve provenance:
     - original node IDs
     - original labels/aliases
     - source files/locations
  7. Recompute graph metrics (communities, centrality, report outputs) on canonical graph.

  ## Suggested implementation details
  - New module: `entity_resolution` (or similar).
  - New config section for alias rules and thresholds.
  - Optional exported artifact: `entity_map.json` mapping original IDs -> canonical IDs.
  - Include merge statistics in run report (nodes merged, aliases applied, unresolved candidates)

yotamleo:
Adding a concrete data point for the cross-file half of this, from a 14,505-note Obsidian vault extracted with 0.9.48.

Because node identity is minted per source file, the same real-world entity ends up as several nodes whenever the surface form varies at all across files. Counts from one run:

- **39 twin pairs** of the form `@handle` vs bare `handle` — i.e. the only difference is a leading `@`.
- **12 duplicate groups** where a hand-curated note and one or more importer-generated notes describe the same entity, differing only by case and/or the `@` prefix (e.g. `@cyrilxbt` / `cyrilXBT` / `@cyrilXBT`).
[...]

PR #3008, opened 2026-08-24
feat(merge): link a type declaration two repos share

### LINKED ISSUE #3007: merge-graphs: a contract type both repos declare stays two unconnected nodes
`merge-graphs` prefixes every node id with its repo tag, so a contract type that two services both declare becomes two unconnected nodes. On a message bus that is the one hop worth having: the producer names the message type in one repo, the consumer implements `IConsumer<T>` on the same type in the other, and nothing in the merged graph gets you from one to the other.

Concretely, from a merged graph of two .NET services:

```
p_CatalogService::...syncproductupserttosearchevent        <- .UpdateProductOnSearchAppAsync() references it
p_ElasticsearchService::...syncproductupserttosearchevent   <- ProductUpsertConsumer references it
```

Same namespace, same name, no edge between them. `path` from the producing method to the consumer returns nothing, so the flow reads as two unrelated halves.

### Why name plus namespace is a usable signal here

I was expecting to need something structural and measured it first. Between the two services, 1440 and 262 sourced type declarations:

- 7 pairs agree on both namespace and name
- all 7 are the shared message contracts, `EventManager.Models.ClearCacheEvent`, `IndexCompletedEvent`, `PrepareElasticDataEvent`, `SyncPriceToSearchEvent`, `SyncProductDeleteToSearchEvent`, `SyncProductUpsertToSearchEvent`, `UpdateVariantElasticPublishedEvent`
- nothing else matches, so no false pair to weigh against them

Dropping the namespace from the key is what makes this dangerous, and it fails immediately: both services carry their own `Settings`, `Configuration` and similar short names. Namespace agreement is doing the work.

### Proposed shape

A pass after `compose` in the `merge-graphs` path that adds an edge between sourced type declarations agreeing on namespace and name and coming from different repos:

```
p_CatalogService::...event  --same_type_as-->  p_ElasticsearchService::...event
```

Edges, not node merging. Two repos can hold copies of a contract that have drifted, and collapsing them hides that, while a link lets a traversal cross with each side keeping its own members, file and provenance. That also keeps this out of #296 and #207's territory, which are about merging duplicates rather than relating them.

On the pair above it adds 7 edges and the producer-to-consumer walk becomes three hops:

```
.UpdateProductOnSearchAppAsync()  (CatalogService)
  -> SyncProductUpsertToSearchEvent  (CatalogService)
  -> SyncProductUpsertToSearchEvent  (ElasticsearchService)   [same_type_as]
  -> ProductUpsertConsumer           (ElasticsearchService)
```

One caveat worth stating: the last hop is a reverse traversal, because the consumer references the event rather than the other way round, so this shows up under an undirected walk. That direction is inherent to `references` and not something this pass changes.

### How this differs from #2134

Cluster graphs there model cross-repo links declared by hand in `cluster.json`, file to file. That is the right tool when the connection is an API call with no shared symbol. A shared contract type needs no declaration: both repos already name it, so the join can be derived. The two compose rather than compete, and if cluster graphs land, the same pass belongs in `cluster build` too.

### Related

- #2134 cluster graphs, above.
- #296 and #207 propose merging duplicate nodes. This deliberately relates instead of merging.
- #1560 is the structural-similarity route (MinHash and Jaro-Winkler) that would be needed if the key were weaker than namespace plus name.
- #2878 also touches the merge repos-union path.
- #2997 is what gives the producer side its edge to the message type in the first place. Without it the CatalogService half of that walk does not exist.

Patch with tests coming against `v8`.

durmazoguzhan:
Closes #3007.

`merge-graphs` prefixes every node id with its repo tag, so a contract type both repos declare arrives as two unconnected nodes:

```
p_CatalogService::...syncproductupserttosearchevent         <- the producing method references it
p_ElasticsearchService::...syncproductupserttosearchevent    <- ProductUpsertConsumer references it
```

Same namespace, same name, nothing between them, so a walk from the producer to the consumer finds no route and one flow reads as two unrelated halves.

The merge now adds an edge between sourced type declarations that agree on namespace and name and come from different repos:

```
p_CatalogService::...event  --same_type_as-->  p_ElasticsearchService::...event
```

Edges, not node merging. Two repos can hold copies of a contract that have drifted, and collapsing them would hide that, while a link lets the traversal cross with each side keeping its own members, file and provenance. That also keeps it clear of #296 and #207, which are about merging duplicates rather than relating them.

### Why namespace plus name

I expected to need something structural and measured first. Between two .NET services with 1440 and 262 sourced type declarations:

- 7 pairs agree on namespace and name
- all 7 are the shared contracts: `EventManager.Models.ClearCacheEvent`, `IndexCompletedEvent`, `PrepareElasticDataEvent`, `SyncPriceToSearchEvent`, `SyncProductDeleteToSearchEvent`, `SyncProductUpsertToSearchEvent`, `UpdateVariantElasticPublishedEvent`
- nothing else matches, so there is no false pair to weigh against them

Keying on the bare name is what makes this unsafe, and it fails at once: both services carry their own `Settings` and `Configuration`. The namespace is doing the work, and the test suite pins that.

### What it buys

On that pair the pass adds 7 edges and the producer to consumer walk becomes three hops:

```
.UpdateProductOnSearchAppAsync()  (CatalogService)
  -> SyncProductUpsertToSearchEvent  (CatalogService)
  -> SyncProductUpsertToSearchEvent  (ElasticsearchService)   [same_type_as]
  -> ProductUpsertConsumer           (ElasticsearchService)
```

The last hop is a reverse traversal, since the consumer references the event rather than the other way round, so this shows up on an undirected walk. That direction comes from `references` and is not something this pass changes.

The producer's first edge comes from #2997. Without that fix the CatalogService half of this walk does not exist, so the two are worth taking together, though neither depends on the other to build.

### Tests

`tests/test_cross_repo_shared_types.py`, 6 tests through the CLI the way `test_merge_graphs_cli.py` does: the cross-repo link itself, the same name in different namespaces staying unlinked, two declarations inside one repo staying unlinked, a type with no namespace staying unlinked, non-type nodes staying unlinked, and a sourceless stub staying unlinked.

```
$ uv run pytest tests/test_cross_repo_shared_types.py -q
6 passed

$ uv run pytest tests/test_merge_graphs_cli.py tests/test_global_graph.py tests/test_serve.py -q
170 passed
```

Full suite excluding `tests/test_skillgen.py`: 4822 passed, 1 failed. The failure is `test_labeling.py::test_label_communities_batches_when_over_batch_size`, which fails the same way on a clean `v8` checkout here (4816 passed, 1 failed) and passes on its own in both, so it is order dependent rather than mine. `test_skillgen.py` fails here on clean `v8` too, which is why it is out of that run. `ruff check` clean on the three files.

### Two things I would change on request

The pass runs unconditionally in `merge-graphs`. A `--no-shared-types` opt-out is easy to add if you would rather this be opt-in, the way `--no-dedup` works.

If the cluster graphs in #2134 land, the same pass belongs in `cluster build`. The two do not compete: cluster links model connections that need declaring, such as an API call with no shared symbol, and this derives the ones both repos already name.

### DIFF
--- a/graphify/cli.py
+++ b/graphify/cli.py
@@ -2445,6 +2445,13 @@ def dispatch_command(cmd: str) -> None:
             if isinstance(hes, list):
                 collected_hyperedges.extend(h for h in hes if isinstance(h, dict))
             merged = _nx.compose(merged, prefixed)
+        # A contract type both repos declare arrives as two unconnected nodes,
+        # since every id is repo-prefixed. Link them so a traversal can cross
+        # the repo boundary (#3007).
+        from graphify.cross_repo_types import link_shared_type_declarations as _link_shared
+        shared_links = _link_shared(merged)
+        if shared_links:
+            print(f"  linked {shared_links} type declaration(s) shared across repos")
         # Drop whatever compose left behind (the last input's list, possibly
         # with internal duplicates) so attach_hyperedges dedups the full
         # collection by id from a clean slate.
--- a/graphify/cross_repo_types.py
+++ b/graphify/cross_repo_types.py
@@ -1,0 +1,75 @@
+"""Join the same declared type across repositories in a merged graph (#3007).
+
+`merge-graphs` prefixes every node id with its repo tag, so a contract type that
+two services both declare arrives as two unconnected nodes. In a message-bus
+codebase that is exactly where the interesting hop lives: the producer references
+`SyncProductUpsertToSearchEvent` in one repo, the consumer implements
+`IConsumer<SyncProductUpsertToSearchEvent>` in the other, and the merged graph has
+no way to get from one to the other even though both sides name the same type.
+
+This pass adds a `same_type_as` edge between type declarations that share both a
+namespace and a name and come from different repos. Requiring the namespace keeps
+it to types that were declared identically rather than two classes that merely
+picked the same short name: on a pair of .NET services with 1440 and 262 declared
+types, the namespace-plus-name match produced 7 pairs, all of them the shared
+`EventManager.Models.*Event` contracts, and nothing else.
+
+Edges only, no node merging. Two repos can hold copies of a contract that have
+drifted, and collapsing them would hide that; a link lets a traversal cross while
+each side keeps its own members, file and provenance.
+"""
+from __future__ import annotations
+
+from collections import defaultdict
+from typing import TYPE_CHECKING
+
+if TYPE_CHECKING:  # pragma: no cover - typing only
+    import networkx as nx
+
+SHARED_TYPE_RELATION = "same_type_as"
+
+
+def link_shared_type_declarations(merged: "nx.Graph") -> int:
+    """Link identically declared types across repos. Returns the edge count added.
+
+    A candidate node is a sourced type declaration carrying a namespace and a
+    repo tag. A group qualifies when its members span at least two repos, and
+    every pair inside a qualifying group gets one edge. Groups are the handful of
+    types two repos genuinely share, so the pairwise walk stays small.
+    """
+    by_declaration: dict[tuple[str, str], list[str]] = defaultdict(list)
+    for node, data in merged.nodes(data=True):
+        if not data.get("_callable_class") or not data.get("source_file"):
+            continue
+        namespace = str((data.get("metadata") or {}).get("namespace") or "")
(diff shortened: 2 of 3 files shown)


safishamsi:

Issue #2810, opened 2026-08-17
JS/TS/Python: distinct same-file symbols still collapse after ID normalization in v0.9.45


imagineers-tyler:
**Version:** graphifyy 0.9.45 · macOS arm64 · Python 3.14.7
**Follow-up to:** #2779 / #2780 (Go, fixed) — same invariant, different extractors
**Related but distinct:** #2774 (Rust, bare-name trigger), #438 / #952 (cross-*file* id collision; this is within a single file, which `_disambiguate_colliding_node_ids` skips via its `len(source_keys) < 2` guard)

## Summary

#2780 correctly fixes the Go case it scoped itself to, and says so explicitly:

> Scope: Go extractor only. The same first-wins `add_node` pattern exists in the other extractors

This issue tracks that remaining scope. The same post-normalization same-file collision is
still live in the JS, TS and Python paths on v0.9.45: two distinct sibling declarations
normalize to one node id, `add_node` drops the second on `if nid in seen_ids: return`, and the
lost declaration's body edges are re-attributed to the survivor.

`ids.py` is byte-identical between v0.9.44 and v0.9.45, so nothing about the shared
normalization changed — which is expected given the PR's stated scope, and is why the other
languages are unaffected by the fix.

## v0.9.44 → v0.9.45

One file, two sibling declarations. "one node" means the second declaration produced no node
at all.

| language | source | v0.9.44 | v0.9.45 |
|---|---|---|---|
| **Go** | `func Run()` + `func run()` | one node | **two nodes — FIXED by #2780** |
| JS | `class R { _onBegin(a){} onBegin(b){} }` | one node | one node |
| JS | `class W { #close(a){} close(b){} }` | one node | one node |
| TS | `declare class Q { $set(a); set(b); }` | one node | one node |
| JS | `function parseURL(a){}` + `function parseUrl(b){}` | one node | one node |
| Python | `def _run(a)` + `def run(b)` | one node | one node |
| Python | `def a_(p)` + `def a(q)` | one node | one node |
| *control* | `def alpha(p)` + `def beta(q)` | two nodes | two nodes |

The `parseURL` / `parseUrl` row is the direct one: that is the **same case-folding axis as
#2779**, and outside Go it still collapses on v0.9.45.

The others are separate axes of the same recipe — `normalize_id` casefolds, maps `[^\w]+` to
`_`, collapses repeated `_`, and strips leading/trailing `_`, so `_x`/`x`, `$x`/`x`, `#x`/`x`
[...]

PR #3010, opened 2026-08-24
feat(csharp): emit a node per property

### LINKED ISSUE #3006: C#: properties are not nodes, so nothing in the graph can point at a member
C# is the only language with a class-shaped type layer whose state is missing from the graph. A field or property produces a `references` edge to its *type* and never a node of its own, so nothing in the graph can point at a member.

```csharp
public class Variant {
    public bool IsShowCategory { get; set; }
    public void Touch() { }
}
```

Nodes: `Variant`, `.Touch()`. Edges: `Variant method .Touch()`. `IsShowCategory` is absent.

### What the other languages do

Checked by running them, not by reading the configs:

| language | member | relation |
| --- | --- | --- |
| C++ | data member | `defines` |
| C | struct field | `defines` (in flight, #2971) |
| Swift | computed / observed property | function-like member node (#2220) |
| Swift | enum case | `case_of` |
| Java | enum constant | `case_of` |
| JS/TS | class field | `method` when callable |
| C# | field, property | nothing |

#2971's argument for C is the one that applies: the struct is C's primary data-modeling construct, so dropping it erased the type layer. Properties are C#'s.

### Properties, not fields

`ids.normalize_id` casefolds and strips leading underscores, so `_count` and `Count` are the same id. Emitting both hands the node to whichever the parser reaches first, which is the private backing field, and the public member disappears behind it. In C# the property is the state's public identity, so it is the half worth having, and fields stay out until the id question has an answer.

### Cost

On a 1457-file .NET service: 9925 -> 14683 nodes and 29547 -> 34182 edges, +48% and +16%. It is a DTO-heavy corpus, which is where the ratio comes from. The added nodes are leaves, so the hub structure does not move.

Patch in #3010.

### Not this issue

A member node on its own does not answer "which code fills this member", because no language in the tree has a member-level usage edge. That is #3012, and it is a design question rather than a gap: it would be the first of its kind here.

### Related

- #2181 and #2220 added Swift computed-property nodes.
- #2971 is adding C struct field nodes.
- #1591 added C# property *type* references, which is what exists today.
- #2911 concerns generic arguments in the same field and property position.
- #2779 and #2810 are the id-normalization collisions that keep fields out of this patch.

durmazoguzhan:
Closes #3006.

C# is the only language with a class-shaped type layer whose state is missing from the graph. A property produces a `references` edge to its *type* and no node of its own:

```csharp
public class Variant {
    public bool IsShowCategory { get; set; }
    public void Touch() { }
}
```

Nodes: `Variant`, `.Touch()`. `IsShowCategory` is absent, so nothing in the graph can point at it.

This emits a node per property with a `defines` edge, the same relation and `context="field"` the C++ data-member branch uses:

```
Variant defines IsShowCategory
```

### Following what the tree already does

Checked by running each language rather than reading the configs:

| language | member | relation |
| --- | --- | --- |
| C++ | data member | `defines` |
| C | struct field | `defines` (in flight, #2971) |
| Swift | computed / observed property | function-like member node (#2220) |
| Swift | enum case | `case_of` |
| Java | enum constant | `case_of` |
| JS/TS | class field | `method` when callable |
| C# | field, property | nothing |

#2971's argument for C is the one that applies here: the struct is C's primary data-modeling construct, so dropping it erased the type layer. Properties are C#'s.

The member node is emitted before the type checks, so it does not depend on what the property is typed as. A primitive and a generic parameter both leave a member behind even though neither produces a type reference. A property stays on `defines` rather than `method`: it is state, not a call target.

### Properties only, not fields

`ids.normalize_id` casefolds and strips leading underscores, so `_count` and `Count` are the same id — an ordinary backing field and its property. Emitting both handed the node to whichever the parser reached first, which is the private field, and the public member vanished behind it; two case-only siblings produced one node with two identical edges. That was the review finding on the first revision.

Fields are out rather than disambiguated here. The id recipe is the single source of truth three producers have to agree on (#811, #550, #1033, #2614), and a local suffix scheme would diverge from whatever #2779 and #2810 settle on. A field's type reference is untouched, and the property path keeps a `seen_ids` check since two properties differing only by case are legal and land on one id.

### Cost, since it is not small

On a 1457-file .NET service:

```
before:  9925 nodes, 29547 edges
after:  14683 nodes, 34182 edges     (+4758 nodes / +48%, +4635 edges / +16%)
```

A DTO-heavy corpus, which is where the ratio comes from. The added nodes are leaves and the hub structure does not move: the top god nodes are unchanged.

### What this does not do

A member node on its own does not answer "which code fills this property", because no language in the tree emits a member-level usage edge. That is #3012, filed with the evidence and the open questions, including whether the node set should instead be narrowed to members something touches — which would reverse the order and make this PR the follow-up. Happy to go that way if you prefer.

### Tests

`tests/test_csharp_member_nodes.py`, 9 tests: an auto-property, several properties on one class, a primitive-typed property, a generic-parameter-typed property, a backing field not taking the property's node, a field alone producing no member node while keeping its type reference, property type references surviving (#1591 guard), a property landing on `defines` while a method stays on `method`, and case-only sibling properties not duplicating an edge.

```
$ uv run pytest tests/test_csharp_member_nodes.py -q
9 passed

$ uv run pytest tests/ -q --ignore=tests/test_skillgen.py --ignore=tests/test_falkordb_integration.py
4825 passed, 1 failed
```

The failure is `test_labeling.py::test_label_communities_batches_when_over_batch_size`, order dependent and failing identically on a clean `v8` checkout here. `test_skillgen.py` and the two FalkorDB cases also fail on clean `v8` in this environment, which is why they are excluded from that run. `ruff check` clean.

### DIFF
--- a/graphify/extractors/engine.py
+++ b/graphify/extractors/engine.py
@@ -3687,30 +3687,6 @@ def _extract_generic(
                         type_node = child.child_by_field_name("type")
                         if type_node is not None:
                             break
-            # Emit a node per declared member, the way the C++ field branch does.
-            # Runs before the type checks below so a member keeps its node even
-            # when its type is a generic parameter or a primitive: the node is
-            # the class's state, independent of what that state is typed as.
-            for child in node.children:
-                if child.type != "variable_declaration":
-                    continue
-                for declarator in child.children:
-                    if declarator.type != "variable_declarator":
-                        continue
-                    member_name_node = declarator.child_by_field_name("name") or next(
-                        (g for g in declarator.children if g.type == "identifier"),
-                        None,
-                    )
-                    if member_name_node is None:
-                        continue
-                    member_name = _read_text(member_name_node, source)
-                    if not member_name:
-                        continue
-                    member_line = declarator.start_point[0] + 1
-                    member_nid = _make_id(parent_class_nid, member_name)
-                    add_node(member_nid, member_name, member_line)
-                    add_edge(parent_class_nid, member_nid, "defines",
-                             member_line, context="field")
             type_info = _read_csharp_type_name(type_node, source)
             if type_info:
                 type_name, qualified, qualifier = type_info
@@ -3758,15 +3734,22 @@ def _extract_generic(
             # field. Use _csharp_collect_type_refs (like the Java/PHP/Kotlin
             # siblings) so `List<Widget>` yields both the List field ref and the
             # Widget generic_arg ref.
+            # A property becomes a node, the way a C++ data member does. Fields
+            # stay out: the id recipe casefolds and strips leading underscores, so
+            # `_count` and `Count` normalize to the same id, and emitting both
+            # would hand the node to whichever the parser reached first — in
+            # practice the private backing field, hiding the public member behind
+            # it. See #3006 for the follow-up.
             prop_node_name = node.child_by_field_name("name")
             if prop_node_name is not None:
                 property_name = _read_text(prop_node_name, source)
                 if property_name:
                     property_line = node.start_point[0] + 1
                     property_nid = _make_id(parent_class_nid, property_name)
-                    add_node(property_nid, property_name, property_line)
-                    add_edge(parent_class_nid, property_nid, "defines",
-                             property_line, context="field")
+                    if property_nid not in seen_ids:
+                        add_node(property_nid, property_name, property_line)
+                        add_edge(parent_class_nid, property_nid, "defines",
+                                 property_line, context="field")
             type_node = node.child_by_field_name("type")
             if type_node is not None:
                 # Record the property's declared type for the method-scoped
--- a/graphify/extractors/engine.py
+++ b/graphify/extractors/engine.py
@@ -3687,6 +3687,30 @@ def _extract_generic(
                         type_node = child.child_by_field_name("type")
                         if type_node is not None:
                             break
+            # Emit a node per declared member, the way the C++ field branch does.
+            # Runs before the type checks below so a member keeps its node even
+            # when its type is a generic parameter or a primitive: the node is
+            # the class's state, independent of what that state is typed as.
+            for child in node.children:
+                if child.type != "variable_declaration":
+                    continue
+                for declarator in child.children:
+                    if declarator.type != "variable_declarator":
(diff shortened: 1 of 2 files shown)


durmazoguzhan:
Reproduces, and on the pattern that matters rather than an exotic one. Fixed in b5f62db, and it narrowed the PR.

`ids.normalize_id` casefolds and strips leading underscores, so an ordinary backing field and its property are one id:

```csharp
public class Counter { private int _count; public int Count { get; set; } }
```

Before the fix that produced a single node labelled `_count` and no `Count` at all: the private field reached `add_node` first, `add_node` returns early on a taken id, and my `add_edge` ran anyway. Two case-only siblings (`count` and `Count`, legal C#) produced one node with two identical `defines` edges.

So fields are out of this PR rather than disambiguated. The id recipe is the single source of truth three producers have to agree on (#811, #550, #1033, #2614), and inventing a local suffix scheme here would diverge from whatever #2779 and #2810 settle on. In C# the property is the state's public identity, so it is the half worth having first, and a field's type reference is untouched either way. The property path keeps a `seen_ids` check, since two properties differing only by case are legal and land on one id.

The scope change makes the cost smaller too: 9925 to 14683 nodes and 29547 to 34182 edges, +48% and +16%, down from +56% and +18%.

Tests are rewritten around it, 9 in total, including the backing-field case asserting the property wins and there is exactly one edge, a field alone producing no member node while keeping its type reference, and case-only sibling properties not duplicating an edge.

```
$ uv run pytest tests/test_csharp_member_nodes.py -q
9 passed

$ uv run pytest tests/ -q --ignore=tests/test_skillgen.py --ignore=tests/test_falkordb_integration.py
4825 passed, 1 failed
```

That one is `test_labeling.py::test_label_communities_batches_when_over_batch_size`, order dependent and failing the same way on a clean `v8` checkout here. `test_skillgen.py` and the FalkorDB cases also fail on clean `v8` in this environment, which is why they are out of that run. `ruff check` clean.

Separately: I split the issue this PR came from, since it had two problems in it and made the reference awkward. #3006 is now only this gap, and the usage half moved to #3012.

safishamsi:

PR #3023, opened 2026-08-24
fix(build): keep definition_file repo-relative like source_file

winesoft-namjin-yun:
Follow-up to #2990 / #2991, shipped in 0.9.49. My own regression — sorry.

## What

`_merge_decl_def_classes` records the dropped impl node's provenance on the survivor by copying its `source_file` verbatim:

```python
keeper["definition_file"] = definition["source_file"]
```

At that point the value is still absolute. Every pass that makes a stored path portable keys on the literal string `"source_file"`, so the new attribute never gets relativized while its sibling does. The result, on a real C++ corpus built at `/data/graphify/repos/ston-edge-server`:

```
Source:     Application/Cache/Process/CacheHttp2Config.h L96
Defined in: /data/graphify/repos/ston-edge-server/Application/Cache/Process/CacheHttp2Config.cpp L47
```

Two consequences, both from the same line:

- **The graph is no longer portable.** `source_file` is repo-relative precisely so a graph built on CI can be read from any checkout; `Defined in:` now names a path that exists on no other machine. A reader who follows it gets "no such file" and reasonably concludes the graph is stale.
- **It leaks the build machine's layout.** For anyone who commits `graphify-out/` or publishes a graph as a release asset, home-directory and directory-structure detail lands in `graph.json`. Nothing in `get_node`'s output hints that this one field is machine-local.

Scope is `.c/.cc/.cpp/.cxx/.m/.mm` paired with a header — C/C++/Objective-C corpora only; nothing else grows the attribute.

## Fix

Relativize `definition_file` wherever `source_file` is relativized:

- `build_from_json`'s per-node normalization (`_norm_source_file`, which also handles the Windows backslash case)
- `watch._relativize_source_files` and `watch._rebase_relative_source_files`

In `watch.py` the two keys now share one `_PORTABLE_PATH_KEYS` tuple, so a future portable path cannot be added to one pass and missed by the other — which is exactly how this slipped through. The scope guard is unchanged: a `definition_file` outside the watched tree is still left alone rather than forced under the root.

Only the recorded string changes. Node ids, labels, `source_file`, node/edge counts and communities are all untouched.

## Tests

Four, all red before the fix except the scope-guard regression test:

- `test_build_from_json_relativizes_definition_file` — an absolute `definition_file` under `root` comes out as `src/Foo.cpp`; `definition_location` passes through untouched.
- `test_relativize_source_files_relativizes_definition_file`
- `test_rebase_relative_source_files_rebases_definition_file`
- `test_relativize_source_files_leaves_an_outside_definition_file_alone` — pins the scope guard for the new key (passes either way; there to stop a future widening).

Note that the tests I shipped with #2991 asserted `str(node["definition_file"]).endswith("Foo.cpp")`, which passes for an absolute path too. That is why this got through, so the new assertions compare full relative strings.

Full suite, same machine, clean `v8` vs this branch:

```
v8:           13 failed, 4756 passed, 232 skipped
this branch:  13 failed, 4760 passed, 232 skipped
```

Identical failure set by name (terraform, ollama-retry-cap, install-references, markdown-nested-frontmatter — all missing-optional-dependency failures in my environment, none on this path); +4 is the new tests.

## Real-world check

Rebuilt three C++ corpora (1.7k, 1.2k and 550 files) on 0.9.49 plus this patch: 13,682 nodes carry a `definition_file`, and every one is now repo-relative — 0 absolute, 0 pointing at a non-impl extension, and all 13,682 survivors are headers, as expected. Node counts before and after the patch are identical (39,041 / 46,551 / 15,613), confirming only the string changed. Verified over the served path too: `get_node` now answers `Defined in: Application/Granite/Delivery/CacheService/CacheFile.cpp L1684`.

Environment: Python 3.12.3, tree-sitter 0.25.2, Linux, AST-only build with no LLM backend.

---

Unrelated nit, mentioning it only because it is one line in the same feature: the 0.9.49 release note for #2990 reads "the node now points at the definition site (the implementation body) instead of the header prototype". The shipped behaviour keeps `source_file` on the declaration and adds the definition site as separate attributes — that additivity (no re-keying of existing graphs) is the part worth stating. Happy to send a note-only PR if that is useful.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 Full release notes with details on each version: [GitHub Releases](https://github.com/safishamsi/graphify/releases)
 
+## Unreleased
+
+- Fix: the `definition_file` recorded on a merged C/C++/Objective-C decl/def node is now stored repo-relative like its sibling `source_file`, instead of keeping the build machine's absolute path — every path-normalizing pass keyed on `source_file` alone, so a graph shipped the builder's filesystem layout and `get_node`'s `Defined in:` line named a path that does not exist on any other checkout (#3223).
+
 ## 0.9.53 (2026-08-30)
 
 - Fix: a batch of cross-language inheritance-edge corrections (thanks @Synvoya): JavaScript `class X extends Y` now emits an `inherits` edge (#1790); PHP interfaces, enums, and traits are captured as class-like nodes with their heritage (#1791); Scala `trait` declarations become class-like nodes (#1792) and qualified `extends`/`with` bases resolve to the tail type (#1794); a qualified Kotlin supertype resolves to its tail type instead of the package head (#1793); a C# interface extending an interface is classified as `inherits`, not `implements` (#1817); and a Go interface type-set constraint no longer emits a spurious `embeds` edge (#1818).
--- a/graphify/build.py
+++ b/graphify/build.py
@@ -973,6 +973,11 @@ def build_from_json(extraction: dict, *, directed: bool = False, root: str | Pat
                 continue
             if "source_file" in node:
                 node["source_file"] = _norm_source_file(node["source_file"], _root)
+            # A merged C/C++/ObjC decl/def node also carries the definition's
+            # file; it is a source path like any other and must be relativized
+            # too, or the graph ships the build machine's absolute path.
+            if "definition_file" in node:
+                node["definition_file"] = _norm_source_file(node["definition_file"], _root)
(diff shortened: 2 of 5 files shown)


winesoft-namjin-yun:
Superseded. The same fix landed on `v8` in 3e5944b0 and 02957036 and shipped in 0.9.54, so there is nothing left for this branch to carry. Closing it to keep the queue clean — #3223 is closed as well.

PR #3042, opened 2026-08-25
fix(serve): MCP orientation refreshes the strict hook's freshness stamp

### LINKED ISSUE #3039: Orienting through the MCP server never refreshes the strict hook's "recently oriented" stamp — agents that consult the graph via query_graph still get their first read denied
## The defect

CHANGELOG 0.9.19 defines the strict-mode contract: *"Running any `graphify query`/`explain`/`path` refreshes a short-lived 'recently oriented' stamp that suppresses the block."* The three CLI commands each call `_touch_query_stamp` (`graphify/cli.py` — the calls at the end of the `query`, `path`, and `explain` handlers), which writes `graphify-out/cache/last_query_stamp`.

The MCP server never touches that stamp — `graphify/serve.py` contains zero references to it. `_tool_query_graph` runs `_query_graph_text`, logs to querylog, and returns.

**Consequence, measured on a live deployment (strict hook installed, agents oriented via the graphify MCP server):** an agent that consulted the graph exclusively through MCP `query_graph` was still treated as blind by the strict read guard and had its first raw source read denied — after it had already done exactly what the guard exists to force. Agents wired through `graphify serve` (the recommended zero-copy setup, e.g. one shared HTTP server per workspace) hit this every session.

## Expected

Consulting the graph through the server's orientation tools should count as orientation, exactly like the CLI:

| CLI (stamps today) | MCP twin |
|---|---|
| `graphify query` | `query_graph` |
| `graphify path` | `shortest_path` |
| `graphify explain` | `get_node` + `get_neighbors` (explain's node detail + connections, split into two tools) |

Browsing tools (`graph_stats`, `god_nodes`, `get_community`) and the PR tools have no stamping CLI precedent (`god-nodes` doesn't stamp) and should stay as they are. With `project_path`, the stamp belongs next to the graph that answered — that project's `graphify-out/cache` — since that's the project whose reads the guard gates. A failed call should stamp nothing.

## A ready patch exists

Implemented, tested (stamp mtime advances on each orientation tool, non-orientation tools leave an aged stamp alone, `project_path` stamps the serving project, failed calls stamp nothing; full serve/hook-strict families green) on our production fork: TelB-io/graphify@fd76abc9070c37766f1949232af317579e716877 (branch `upstream/mcp-stamp-touch`, based on v8 @ 0.9.49). Companion PR opened from that branch.

Searched existing issues/PRs for "stamp mcp", "query stamp", "strict hook mcp", "orient" before filing — no prior report found (nearest: #2202, a different strict-hook concern).

Azeem1985:
Fixes #3039.

The strict read guard admits an agent only while the "recently oriented" stamp (`graphify-out/cache/last_query_stamp`) is fresh; the CLI's `query` / `path` / `explain` each touch it, but their MCP twins never did — so an agent that oriented through the server was still treated as blind and had its first raw read denied (measured on a live deployment; details in the issue).

`call_tool` now touches the stamp after every **successful** call to one of the four orientation tools, chosen by strict parity with the CLI's stamping commands:

| CLI (stamps) | MCP twin | stamps now |
|---|---|---|
| `graphify query` | `query_graph` | yes |
| `graphify path` | `shortest_path` | yes |
| `graphify explain` | `get_node` + `get_neighbors` (explain's split) | yes |
| `graphify god-nodes` (no stamp) | `god_nodes` | no |
| — | `graph_stats`, `get_community`, PR tools | no |

Two deliberate choices:
- **Stamp on any successful call, including "No node matching" answers** — the CLI explain's no-stamp-on-no-match is an accident of its early `sys.exit(0)`, not a contract; a no-match answer is still a graph consult, and the guard's premise ("asked the graph first") holds.
- **Stamp next to the graph that answered** — a `project_path` call stamps THAT project's `graphify-out/cache`, the graph whose reads the guard gates. A failed call stamps nothing.

The stamp writer is the CLI's own fail-silent `_touch_query_stamp` (lazy import; `cli` imports `serve` only inside handlers, so no cycle).

Tests: stamp created by `query_graph`; each orientation tool refreshes an aged stamp (parametrized); browsing tools leave an aged stamp untouched; `project_path` stamps the serving project, not the default; failed call stamps nothing. serve/serve_http/hook_strict families green on top of v8 @ 0.9.49; ruff clean.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 Full release notes with details on each version: [GitHub Releases](https://github.com/safishamsi/graphify/releases)
 
+## Unreleased
+
+- Fix: orienting through the MCP server now refreshes the strict read hook's "recently oriented" stamp, exactly like the CLI. `graphify query` / `path` / `explain` each touch `graphify-out/cache/last_query_stamp` so the strict hook lets the next raw file read through — but their MCP twins never did, so an agent that consulted the graph via `query_graph` was still treated as blind and had its first read denied. The stamp is now touched after every successful `query_graph`, `shortest_path`, `get_node`, or `get_neighbors` call (get_node + get_neighbors are the MCP split of `explain`), next to the graph that actually answered — a `project_path` call stamps that project's `graphify-out/cache`, which is the graph whose reads the guard gates. Browsing tools (`graph_stats`, `god_nodes`, `get_community`) and the PR tools don't stamp on the CLI and still don't over MCP; a failed call stamps nothing.
+
 ## 0.9.50 (2026-08-25)
 
 - Fix: Ruby methods whose names end in `!`, `?`, or `=` now keep distinct node ids, so `save` and `save!` (or `foo` and `foo=`) no longer collide into one node; the label keeps the raw spelling and member-call resolution still matches (#3077, thanks @hopstreax).
--- a/graphify/serve.py
+++ b/graphify/serve.py
@@ -1968,6 +1968,19 @@ def _build_server(graph_path: str):
         "triage_prs": _tool_triage_prs,
     }
 
+    # The MCP twins of the CLI commands whose runs refresh the "recently
+    # oriented" stamp that the strict read hook honours (`graphify query` /
+    # `path` / `explain` each call cli._touch_query_stamp). query_graph and
+    # shortest_path map 1:1 onto `query` and `path`; get_node + get_neighbors
+    # are the MCP split of `explain`. Graph-level browsing (god_nodes,
+    # graph_stats, get_community) and PR triage don't stamp on the CLI either,
+    # so they stay out. Without this set, an agent that orients through the
+    # MCP server is still treated as blind by the strict guard and gets its
+    # first raw read denied — after it already consulted the graph.
+    _ORIENTATION_TOOLS = frozenset(
+        {"query_graph", "get_node", "get_neighbors", "shortest_path"}
+    )
+
     def _load_community_labels() -> dict[int, str]:
         labels_path = Path(active_graph_path).parent / ".graphify_labels.json"
         if labels_path.exists():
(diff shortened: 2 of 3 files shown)


Azeem1985:
Thanks for the grounded review. The flagged fan-out is in the end-to-end test functions — each deliberately exercises the full server lifecycle (spawn, handshake, call, stat the stamp/cache) inline so the test reads as one narrative; happy to factor the shared steps into a fixture if maintainers prefer lower per-test fan-out. No production-code findings noted. — Azeem1985/TelB-io fork

Azeem1985:
**Correcting my own comment above.**

My comment of 2026-08-25 02:03 ended with *"No production-code findings noted."* That is wrong. I answered the five inline notes — all on `tests/test_serve_http.py` — and never addressed the two findings in the review body, which are both on production code:

- **`graphify/serve.py:2068`** — *Post-call stamp reads shared `active_graph_path` after another concurrent tool call may have rebound it* · Escalate · medium
- **`graphify/serve.py:2072`** — *Orientation stamp uses shared `active_graph_path` after tool execution* · Escalate · medium

Both carry the reviewer's own honest caveat: *"agreed by 2 of 2 members but NOT verified (no proof, no reproducing execution)."* Here is the missing execution, read off this PR's head (`4c3c76b35b47a1966a7c64fe4c0b3bea56dfb1ac`), not off my fork:

```
2063            _select_graph(project_path)  # binds active_graph_path
2064            text = handler(arguments)
2065        except Exception as exc:
2066            return [types.TextContent(...)]
2067        if name in _ORIENTATION_TOOLS:
...
2073            _touch_query_stamp(Path(active_graph_path))
```

**There is no `await` anywhere between line 2063 and line 2073.** `call_tool` is `async`, but `handler(arguments)` is a direct synchronous call into a plain function from the `_handlers` dict. A coroutine yields only at an `await`, so on one event loop no second `call_tool` can run and rebind `active_graph_path` inside that span. The interleaving the finding describes cannot occur as the code stands.

**The boundary, stated rather than glossed over.** That argument holds for the single-event-loop model the server runs under today. It would not hold if handlers were ever dispatched to a thread pool, or if two servers shared this module state in one process. If either becomes possible, the one-line fix is to capture the path `_select_graph` bound into a local and stamp that local instead of re-reading the shared name — happy to send that on request, or to fold it in here pre-emptively if you would rather the code not depend on the no-suspension-point argument.

PR #3043, opened 2026-08-25
fix(query): attach the learning sidecar so CLI query output carries lesson annotations

### LINKED ISSUE #3040: `graphify query` never shows the learning sidecar's Lesson annotations, though the docs (README + CHANGELOG 0.9.3) promise it does — only explain and the MCP server surface them
## The defect

The work-memory overlay feature (CHANGELOG 0.9.3) documents: *"`graphify explain` / `query` / `GRAPH_REPORT.md` / the HTML viewer surface them where you look (a `Lesson:` hint, a colored node ring)"* — and the README's reflect section repeats it: *"graphify explain / query then show a 'Lesson:' hint, flagged 'code changed — re-verify' when the source moved on."*

`query` doesn't, on the CLI. The renderer support exists — `_subgraph_to_text` (`graphify/serve.py`) appends `learning=<status>[:stale]` inside NODE brackets when the loaded graph carries `G.graph["_learning_overlay"]` — and `serve._load_graph` attaches that overlay, so **MCP** `query_graph` shows the annotations. But the CLI `query` handler (`graphify/cli.py`, the `node_link_graph` load in the query branch) builds its graph straight from `graph.json` and never attaches the overlay, so the overlay lookup always sees `{}`.

**Consequence, verified on a live deployment with a populated `.graphify_learning.json`:** the same question answered over MCP shows the lessons; the CLI answer silently drops them. `reflect`'s verdicts are invisible on the surface most humans use.

## Expected

The CLI query path attaches the overlay exactly like `serve._load_graph` does (display-only, fail-safe to empty), so both surfaces emit the identical `learning=<status>[:stale]` form, including the "code changed since — re-verify" stale marker. Without a sidecar the output stays byte-identical.

## A ready patch exists

Implemented and tested (annotated node carries `learning=preferred]`, vanished cited file renders `:stale`, no sidecar → unchanged output; query/reflect families green) on our production fork: TelB-io/graphify@e319b22 (branch `upstream/query-learning-overlay`, based on v8 @ 0.9.49). Companion PR opened from that branch.

Searched existing issues/PRs for "learning overlay", "lesson query", "learning sidecar" before filing — no prior report (nearest family: #2821 / #3026, other attributes un-rendered by query, different fields).

Azeem1985:
Fixes #3040.

The docs promise `query` surfaces the work-memory lessons (CHANGELOG 0.9.3: *"graphify explain / query / GRAPH_REPORT.md / the HTML viewer surface them where you look"*; README's reflect section says the same), and the renderer has carried `learning=<status>[:stale]` NODE-annotation support all along — but only for graphs that carry `G.graph["_learning_overlay"]`. `serve._load_graph` attaches it (so MCP `query_graph` annotates) and `explain` merges the sidecar itself; the CLI `query` loader builds its graph straight from `graph.json` and never attached the overlay, so the CLI answer silently dropped the lessons the MCP answer showed.

Fix: the CLI query path now attaches the overlay exactly like `serve._load_graph` — display-only, fail-safe to `{}` on any error. Both surfaces emit the identical `learning=<status>[:stale]` form, including the "code changed since — re-verify" stale marker; without a sidecar the output stays byte-identical (pinned by test).

Tests: sidecar entry → `learning=preferred]` on the annotated node only; vanished cited file → `learning=tentative:stale]`; no sidecar → no `learning=` anywhere. query/reflect/serve families green on top of v8 @ 0.9.49; ruff clean.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 Full release notes with details on each version: [GitHub Releases](https://github.com/safishamsi/graphify/releases)
 
+## Unreleased
+
+- Fix: `graphify query` now surfaces the work-memory lesson annotations. The renderer has carried `learning=<status>[:stale]` NODE-line support since the overlay shipped, and both `graphify explain` and the MCP server attach the `.graphify_learning.json` sidecar at load time — but the CLI `query` loader never did, so the same question answered over MCP showed the lessons while the CLI answer silently dropped them (the docs promise `query` surfaces a lesson hint). The CLI query path now attaches the overlay exactly like `serve._load_graph` (display-only, fail-safe to empty), so annotations — including the `:stale` "code changed since — re-verify" marker — render identically on both surfaces; without a sidecar the output is byte-identical to before.
+
 ## 0.9.50 (2026-08-25)
 
 - Fix: Ruby methods whose names end in `!`, `?`, or `=` now keep distinct node ids, so `save` and `save!` (or `foo` and `foo=`) no longer collide into one node; the label keeps the raw spelling and member-call resolution still matches (#3077, thanks @hopstreax).
--- a/graphify/cli.py
+++ b/graphify/cli.py
@@ -1162,6 +1162,17 @@ def dispatch_command(cmd: str) -> None:
         except Exception as exc:
             print(f"error: could not load graph: {exc}", file=sys.stderr)
             sys.exit(1)
+        # Work-memory overlay: attach the .graphify_learning.json sidecar the
+        # same way the MCP loader does (serve._load_graph), so the renderer's
+        # learning=<status>[:stale] NODE annotations appear in CLI `query`
+        # output too — `explain` and MCP already surface them, and the docs
+        # promise `query` does as well. Empty overlay on any error, leaving
+        # un-annotated output byte-identical.
+        try:
+            from graphify.reflect import load_learning_overlay as _llo
+            G.graph["_learning_overlay"] = _llo(gp)
+        except Exception:
+            G.graph["_learning_overlay"] = {}
         import time as _time
         _t0 = _time.perf_counter()
         _mode = "dfs" if use_dfs else "bfs"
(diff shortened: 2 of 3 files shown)


Azeem1985:
Thanks for the grounded review. The flagged fan-out is in the end-to-end test functions — each deliberately exercises the full server lifecycle (spawn, handshake, call, stat the stamp/cache) inline so the test reads as one narrative; happy to factor the shared steps into a fixture if maintainers prefer lower per-test fan-out. No production-code findings noted. — Azeem1985/TelB-io fork

Azeem1985:
**Correcting my own comment above — it was wrong twice.**

My comment of 2026-08-25 02:03 on this PR describes *"the flagged fan-out … in the end-to-end test functions."* **There is no fan-out finding on this PR.** That text belongs to #3042 and was pasted here by mistake, along with the sentence *"No production-code findings noted."*

This PR has exactly one finding, and it is on production code:

- **`graphify/cli.py:1174`** — *Overlay attached from graph.json path instead of its parent directory* · Escalate · medium, marked *"agreed by 2 of 2 members but NOT verified (no proof, no reproducing execution)."*

**That finding is incorrect, and `load_learning_overlay` is its own counter-evidence.** Read off this PR's head (`4f14690e66df1420403e21f3a20284ff15f0aef1`):

`graphify/reflect.py:840` takes the graph file, and line 850 derives the directory itself:

```python
def load_learning_overlay(graph_path: Path) -> dict[str, dict[str, Any]]:
    """Load the sidecar next to ``graph_path`` …"""
    ...
    sidecar = Path(graph_path).parent / LEARNING_SIDECAR_NAME
```

So the graph-file path is the intended argument. Passing the parent directory would make the sidecar resolve one level too high and the overlay would silently come back empty — that would be the defect, not the fix.

Every call site at this same head passes the graph file, including three that predate this change:

| where | call |
|---|---|
| `report.py:52` | `load_learning_overlay(gp)` |
| `cli.py:1630` | the `explain` path |
| `serve.py:63` | the MCP `_load_graph` path — `_llo(resolved)` |
| `cli.py:1173` | this change |

The change matches the three that were already there. (Small note: the finding is anchored at line 1174, which is the `except Exception:` line; the call itself is 1173.)

Apologies for the noise on both counts — the wrong text and the unanswered finding.

PR #3048, opened 2026-08-25
fix(postgres): carry tree-sitter-sql in the postgres extra and surface grammar errors

### LINKED ISSUE #3047: postgres extra lacks tree-sitter-sql: --postgres on a live DB silently reports 0 nodes, exit 0
## What happens

Install exactly what the README prescribes for live-PostgreSQL introspection (README.md, extras table: `uv tool install "graphifyy[postgres]"`), then run against a real database:

```
graphify extract --postgres 'postgresql://postgres:proof@127.0.0.1:5433/postgres'
[graphify extract] PostgreSQL: 0 nodes, 0 edges
```

Exit code 0. The database had 3 tables with FKs, a view, and a function — reproduced on postgres:17-alpine with graphifyy 0.9.48.

## Why

`introspect_postgres()` (graphify/pg_introspect.py) reconstructs DDL and hands it to `extract_sql()`, which requires the **tree-sitter-sql** grammar. That grammar lives in the separate `sql` extra (pyproject.toml: `sql = ["tree-sitter-sql"]`) — the `postgres` extra ships only `psycopg[binary]`. When the grammar is missing, `extract_sql()` returns `{'nodes': [], 'edges': [], 'error': 'tree_sitter_sql not installed. ...'}` and `introspect_postgres()` returns that dict as-is; the CLI prints the node count and never looks at the `error` key. A live schema is indistinguishable from an empty database.

Adding `tree-sitter-sql` by hand makes the same command produce the full schema graph (tables/view/function nodes, FK `references` edges).

## Suggested fix

1. `postgres = ["psycopg[binary]", "tree-sitter-sql"]` — the extra should carry what its own code path needs.
2. `introspect_postgres()` should raise when `extract_sql()` reports an `error`, reusing the CLI's existing `(ConnectionError, ImportError)` exit path, so the failure is loud instead of a plausible-looking empty graph.

PR with both fixes + a regression test to follow.

Azeem1985:
Fixes #3047.

Installing the README-prescribed `graphifyy[postgres]` and running `graphify extract --postgres <DSN>` against a live database printed `PostgreSQL: 0 nodes, 0 edges` with exit 0: the `postgres` extra ships only `psycopg[binary]`, while the introspection path feeds reconstructed DDL to `extract_sql()`, which needs `tree-sitter-sql` from the separate `sql` extra. The missing-grammar report came back in an `error` key that `introspect_postgres()` silently dropped, so a live schema looked like an empty database.

- **pyproject.toml**: `postgres` extra now includes `tree-sitter-sql`
- **graphify/pg_introspect.py**: raise `ImportError` when `extract_sql()` reports an error — the CLI already catches `(ConnectionError, ImportError)` on this path and exits 1, so the failure becomes loud with the correct install hint
- **tests/test_pg_introspect.py**: regression test — a grammar error must raise, never return an empty graph

Verified end-to-end on postgres:17-alpine (3 tables + FKs, a view, a function): with the grammar present the same command produces all 6 schema nodes and both FK `references` edges. Full `tests/test_pg_introspect.py` suite passes (9/9).


Azeem1985:
Thanks for the grounded review — no blocking issues noted. The coupling advisories point at the neo4j/falkordb pusher functions, which this PR doesn't modify (they're the subject of the separate streaming/batched-push offers); the postgres fix itself — the extra carrying tree-sitter-sql and surfacing extract_sql errors instead of a silent 0-node graph — drew no findings. Happy to split or adjust if preferred. — Azeem1985/TelB-io fork

Azeem1985:
Closing this in favour of #3072, which is the same change on a branch cut from `v8` — **4 files** instead of the 62 this one shows.

This PR was opened from a branch that also carried unrelated work from our fork, so its diff was 62 files of a two-line fix plus effectively our entire fork. That is my mistake in how I cut the branch, and it made this unreviewable. Nothing about the proposed change has altered; only the packaging.

#3072 has the full rationale and measurements. Sorry for the extra noise, and thank you for your patience.

PR #3050, opened 2026-08-25
fix(transcribe): env pass-through for yt-dlp cookies and JS runtimes

### LINKED ISSUE #3049: transcribe: download_audio cannot receive yt-dlp cookies or a JS runtime — YouTube's bot wall is unfixable from graphify on server IPs
**What happens**: On a datacenter/VPS IP, every YouTube download through `graphify add <youtube-url>` or `graphify.transcribe.download_audio()` fails with yt-dlp's `Sign in to confirm you're not a bot` (player responses come back LOGIN_REQUIRED for every client — web, tv, tv_embedded, android, mweb, over IPv4 and IPv6). Search (`ytsearchN:`) still works; only download is walled.

**Why graphify can't be fixed from outside**: yt-dlp's FAQ prescribes `--cookies` for this, and modern yt-dlp wants `--js-runtimes node` for full YouTube format coverage. Both exist on the yt-dlp CLI, but `download_audio()` builds its `YoutubeDL` options dict inline (graphify/transcribe.py, the `ydl_opts = {...}` block) with no pass-through — a user who exports cookies.txt has no way to hand it to graphify. (The yt-dlp Python API also does not read yt-dlp config files, so even a user-level yt-dlp config is ignored.)

**Proposed fix** (mirrors the module's existing `GRAPHIFY_WHISPER_MODEL` env convention):
- `GRAPHIFY_YTDLP_COOKIES` → `cookiefile`
- `GRAPHIFY_YTDLP_JS_RUNTIMES` (comma-separated, e.g. `node`) → `js_runtimes`

No env set → options unchanged. PR with tests incoming from TelB-io.

Azeem1985:
Fixes #3049.

`download_audio()` hardcodes its embedded `YoutubeDL` options, so on server IPs where YouTube answers `Sign in to confirm you're not a bot` (yt-dlp's FAQ prescribes browser cookies for this), the user's exported cookies file has no way in — and neither does `--js-runtimes node`, which yt-dlp's YouTube extractor wants for full format coverage.

This mirrors the module's existing env convention (`GRAPHIFY_WHISPER_MODEL`, `GRAPHIFY_WHISPER_PROMPT`):

- `GRAPHIFY_YTDLP_COOKIES` → `cookiefile` (same as `yt-dlp --cookies`)
- `GRAPHIFY_YTDLP_JS_RUNTIMES` → `js_runtimes` (same as `yt-dlp --js-runtimes`), e.g. `node`

With neither set, the options dict is byte-identical to before. 4 new tests; `tests/test_transcribe.py` passes (15). Applies clean on v8 (0.9.49).

### DIFF
--- a/graphify/transcribe.py
+++ b/graphify/transcribe.py
@@ -47,6 +47,32 @@ def is_url(path: str) -> bool:
     return any(path.startswith(p) for p in URL_PREFIXES)
 
 
+def _ytdlp_env_opts() -> dict:
+    """Optional yt-dlp options passed through the environment.
+
+    YouTube serves ``Sign in to confirm you're not a bot`` to datacenter IPs;
+    yt-dlp's own FAQ prescribes browser cookies for that, and its extractor
+    needs a JavaScript runtime for full format coverage. Both are reachable
+    from the CLI but were impossible to pass through the embedded API here,
+    so mirror the transcriber's env convention (GRAPHIFY_WHISPER_MODEL):
+
+    - ``GRAPHIFY_YTDLP_COOKIES``: path to a Netscape cookies.txt
+      (same as ``yt-dlp --cookies``).
+    - ``GRAPHIFY_YTDLP_JS_RUNTIMES``: comma-separated runtime names,
+      e.g. ``node`` or ``deno,node`` (same as ``yt-dlp --js-runtimes``).
+    """
+    opts: dict = {}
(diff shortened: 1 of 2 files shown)


Azeem1985:
Thanks for the review — glad to see the gate come back clean. For what it's worth, our fork (TelB-io/graphify) has been running this change in production: with the env vars unset it's a no-op as described, and with them set it gets yt-dlp past YouTube's bot check reliably. Happy to adjust anything if the maintainers want changes before merge.

Issue #2004, opened 2026-07-18
affected/reverse-dep false negatives; god_nodes not a CLI subcommand; --output flag ignored on extract (v0.9.18, Windows)


katha-begin:
**Version:** graphify 0.9.18 (installed via `uv tool install "graphifyy[anthropic,mcp]==0.9.18"`)
**OS:** Windows 11 (Git Bash / MSYS shell)

## 1. `affected` (and reverse-dep lookups generally) return false negatives when queried by the scanned-file node name

When file A imports a symbol from file B via `from pkg.sub.mod import X`, the resulting `imports_from` edge's `target` is an ID derived from the *import string itself* (dots → underscores, e.g. `pkg_sub_mod`), not the ID of the actual scanned node for `mod.py` (which may have a different prefix depending on how the corpus root was passed to `extract`, e.g. `sub_mod` or `code_sub_mod`).

Repro:
```
graphify affected "mod.py" --graph graph.json            # -> "No affected nodes found." (false negative)
graphify affected "sub_mod" --graph graph.json            # -> also fails (real scanned-node id)
graphify affected "pkg_sub_mod" --graph graph.json        # -> works (import-string-derived id)
```
`graphify diagnose multigraph --graph graph.json --json` shows a large `dangling_endpoint_edges` count (thousands, on a ~21k-node/~49k-edge corpus), consistent with these import-string-derived target IDs never resolving to a real node.

This is reproducible identically whether the corpus is extracted in-place from its real package root or from a staged/copied directory tree — it is not a staging artifact, it's the import-target ID resolution itself.

Impact: any downstream tool/agent that calls `affected "<file-or-symbol-name>"` expecting it to "just work" (the natural first guess) silently gets zero results instead of an error, which reads as "nothing depends on this" rather than "wrong ID format."

Suggested fix: either (a) deduplicate/alias the import-string-derived node ID to the real scanned node when the import resolves to a file that's actually in the corpus, or (b) have `affected` do fuzzy/label matching (like `explain`/`query` already do) before falling back to exact-ID lookup, or (c) document the ID-format requirement prominently next to the `affected` command help text.

## 2. `god_nodes` is documented as a capability but is not a CLI subcommand in 0.9.18

[...]

PR #3054, opened 2026-08-25
fix(cli): affected and god-nodes refresh the orientation stamp and write the query ledger

### LINKED ISSUE #3053: CLI affected and god-nodes never refresh the strict hook's orientation stamp nor write the query ledger
## Summary

The strict read hook's admission contract is the "recently oriented" stamp: running `graphify query` / `path` / `explain` touches `graphify-out/cache/last_query_stamp`, and while it is fresh the strict guard lets raw file reads through (CHANGELOG 0.9.19). Those three verbs also write the opt-in query ledger (`graphify/querylog.py`, the `GRAPHIFY_QUERY_LOG*` gates).

Two other read-only graph verbs do neither:

- **`graphify affected`** — `cli.py` `elif cmd == "affected":` (line 1189 on current v8): loads the graph, prints `format_affected(...)`, and returns. No `_touch_query_stamp`, no `querylog.log_query`.
- **`graphify god-nodes`** — `cli.py` `elif cmd in ("god-nodes", "god_nodes"):` (line 1257 on current v8, wired by #2004): computes and prints the hubs, and returns. Same two omissions.

Compare the `query` verb right above them: `querylog.log_query(...)` at cli.py:1177 and `_touch_query_stamp(gp)` at cli.py:1187 on the same success path.

## Why it matters

- An agent that orients through `affected` ("what breaks if X changes") or `god-nodes` ("what are this graph's hubs") is still treated as **blind** by the strict guard: its next raw read is denied — after it already consulted the graph. This is exactly the failure mode of #3039, one door over: there the MCP tools didn't stamp; here two of the CLI's own verbs don't.
- The runs are also **invisible to the query ledger** when logging is enabled, so any audit of graph usage undercounts real orientation.

## Expected

`affected` and `god-nodes` are graph consults exactly like `query`/`path`/`explain`: a successful run should refresh the stamp and append a ledger line (kind `affected` with the impacted-node count; kind `god_nodes` with the returned-hub count), fail-silent like the existing call sites, ledger still opt-in.

## Community check

Searched issues/PRs for "affected stamp", "god-nodes stamp", "last_query_stamp", "strict hook affected", "querylog affected", "query ledger orientation": no prior reports; the nearest neighbors are our own #3039/#3042 (the MCP half of the same defect class).

We run graphify in production on an agent-operated Obsidian ops vault where a deny-hook enforces graph-first reading; this surfaced in tonight's audit of which doors refresh the stamp. Fix PR with tests to follow shortly.

Azeem1985:
Fixes #3053. Same defect class as #3039 (MCP half: #3042), one door over — this is the CLI half.

## Defect

The strict read hook admits an agent only while the "recently oriented" stamp (`graphify-out/cache/last_query_stamp`) is fresh, and the opt-in query ledger is the audit trail of graph consults. `graphify query` / `path` / `explain` each stamp + log on success (e.g. `querylog.log_query` at cli.py:1177 and `_touch_query_stamp(gp)` at cli.py:1187 for `query`) — but `graphify affected` and `graphify god-nodes` (#2004) do neither. An agent that orients through either verb stays blind to the strict guard (next raw read denied after it already consulted the graph) and invisible to the ledger.

## Fix

Orientation is orientation regardless of which door it comes through. Both verbs now mirror the query verb's calls on the success path:

- `affected`: `querylog.log_query(kind="affected", question=<seed>, corpus=<graph>, nodes_returned=<impacted count>, depth, duration_ms)` then `_touch_query_stamp(gp)`, then print. The count is exact: each impacted node renders as exactly one `- ` line and no header line starts with `- `.
- `god-nodes`: `querylog.log_query(kind="god_nodes", question="top N", corpus=<graph>, nodes_returned=len(gods))` then `_touch_query_stamp(gp)`.

Both are fail-silent like the existing call sites (`log_query` never raises; `_touch_query_stamp` swallows); the ledger stays opt-in behind the `GRAPHIFY_QUERY_LOG*` gates; error paths (missing/corrupt graph) exit before either call, so a failed run stamps and logs nothing. Output bytes of both verbs are unchanged.

Not touched here: the MCP tools' stamping is #3042; if both land, `god_nodes` over MCP becomes the remaining non-stamping door for the same consult and can be folded into #3042's `_ORIENTATION_TOOLS` set as a follow-up.

## Tests

- `test_affected_cli_stamps_orientation_and_logs` — an aged stamp's mtime advances; exactly one ledger line with kind `affected`, the exact impacted count, depth, and duration; ledger env-gated via `GRAPHIFY_QUERY_LOG` exactly like tests/test_querylog.py; tmp graph fixture, offline.
- `test_god_nodes_cli_stamps_orientation_and_logs` — same shape for `god-nodes`; the logged count equals the number of hubs actually printed, not the `--top` ceiling.
- Families run on this branch: test_affected_cli + test_god_nodes_cli + test_querylog + test_query_cli + test_explain_cli + test_hook_strict — 82 passed; ruff + py_compile clean.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 Full release notes with details on each version: [GitHub Releases](https://github.com/safishamsi/graphify/releases)
 
+## Unreleased
+
+- Fix: the CLI `graphify affected` and `graphify god-nodes` verbs now count as orientation, exactly like `query` / `path` / `explain`: a successful run refreshes the strict read hook's "recently oriented" stamp (`graphify-out/cache/last_query_stamp`) and writes a query-ledger line (kind `affected` with the impacted-node count, kind `god_nodes` with the returned-hub count; both fail-silent, and the ledger stays opt-in via the usual `GRAPHIFY_QUERY_LOG*` gates). Before this, an agent that oriented through either verb was still treated as blind by the strict guard — its next raw read denied after it had already consulted the graph — and left no trace in the query audit trail. Orientation is orientation regardless of which door it comes through; the MCP tools' missing stamp is the same defect class, reported separately in #3039 / fixed by #3042.
+
 ## 0.9.50 (2026-08-25)
 
 - Fix: Ruby methods whose names end in `!`, `?`, or `=` now keep distinct node ids, so `save` and `save!` (or `foo` and `foo=`) no longer collide into one node; the label keeps the raw spelling and member-call resolution still matches (#3077, thanks @hopstreax).
--- a/graphify/cli.py
+++ b/graphify/cli.py
@@ -1245,15 +1245,31 @@ def dispatch_command(cmd: str) -> None:
         # --graph falls back to its own directory.
         from graphify.paths import GRAPHIFY_OUT_NAME
         graph_root = gp.parent.parent if gp.parent.name == GRAPHIFY_OUT_NAME else gp.parent
-        print(
-            format_affected(
-                graph,
-                query,
-                relations=relations or DEFAULT_AFFECTED_RELATIONS,
-                depth=depth,
-                root=graph_root,
-            )
+        import time as _time
+        from graphify import querylog
+        _t0 = _time.perf_counter()
+        _out = format_affected(
+            graph,
+            query,
+            relations=relations or DEFAULT_AFFECTED_RELATIONS,
+            depth=depth,
+            root=graph_root,
         )
+        # Orientation is orientation regardless of door (#3042): `affected`
+        # answers a graph question exactly like `query`/`path`/`explain`, so a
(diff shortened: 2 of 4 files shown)


Azeem1985:
Thanks for the automated review — glad it reads clean. To restate the intent for human reviewers: `affected` and `god-nodes` are orientation verbs exactly like `query`/`path`/`explain` (an agent that just mapped a blast radius has oriented), so they now refresh the strict-hook freshness stamp and write the opt-in query ledger the same way. Both additions are fail-silent and behind the existing `GRAPHIFY_QUERY_LOG*` gates; running in production on our side since 2026-08-25 (ledger shows `affected`/`god_nodes` lines landing). Happy to adjust shape or naming to taste.

PR #3056, opened 2026-08-25
feat(serve): report build provenance in graph_stats

winesoft-namjin-yun:
## Problem

`graph_stats` is the only place an MCP client can ask "what am I querying?", and it
cannot answer "how old is it?". An agent reading

```
Nodes: 15613
Edges: 40764
Communities: 480
EXTRACTED: 96%
INFERRED: 4%
AMBIGUOUS: 0%
```

has no way to distinguish a graph built ten minutes ago from one built last month,
and will answer questions about the current code from either. Over MCP there is no
fallback: the client cannot stat `graph.json`.

Provenance already exists on disk but never reaches that surface. `export.to_json`
writes `built_at_commit` at the top level, and `callflow_html.py` / `cli.py` both
consume it — but `serve.py` never references it, because `json_graph.node_link_graph`
copies only `data["graph"]` onto `G.graph` and silently drops every other top-level
key. So the HTML report shows the commit and the MCP tools cannot.

## What this changes

1. **`_load_graph` lifts the provenance keys out of the raw payload** onto `G.graph`,
   the same way the existing `_logical_directed` flag is stashed one line above.
   Private names (`_built_at`, `_built_at_commit`) so a graph loaded on the read path
   can never round-trip these into a nested `data["graph"]`.

2. **`graph_stats` appends the provenance it finds**, and nothing when there is none:

   ```
   Built at: 2026-08-25T09:15:42Z
   Built from commit: d6ff04064219c45e6cb1aeda8e66c292b6650307
   ```

   Appended rather than prepended, and omitted entirely when absent, so a
   pre-provenance graph renders exactly as before. **Every graph already in the wild
   gains the commit line with no rebuild** — that half needs only the load fix.

3. **`export.to_json` records a new top-level `built_at`** UTC stamp
   (`YYYY-MM-DDTHH:MM:SSZ`), injectable via a `built_at=` kwarg exactly like
   `built_at_commit`.

## Why `built_at` is a separate field, not derived from the commit

They answer different questions, and only one of them is about freshness:

- `built_at_commit` — *which revision this graph describes.* Stable across
  re-clustering; #2534 deliberately preserves it through a cluster-only rewrite.
- `built_at` — *when this file was written.* This is what staleness is measured
  against. A graph can be a week old while sitting on a commit that is still HEAD,
  and the commit alone cannot tell you that.

Consistent with that split, the stamp is deliberately **not** preserved across a
cluster-only rewrite the way #2534 preserves the commit: cluster does not redo the
extraction (so the commit stands) but it does rewrite the file (so the stamp moves).

The stamp is always written, while the commit stays conditional — a clock read
cannot fail, so there is no "outside a repo" case to omit. `to_json`'s `if commit:`
shape is intentionally not copied.

## The trap this had to avoid

A field that changes on every write is poison for `watch.py`'s no-op skip.
`_canonical_graph_for_compare` / `_canonical_topology_for_compare` decide whether an
incremental rebuild actually changed anything, and both already `pop("built_at_commit")`
for exactly this reason. Leaving the stamp in would make "did the graph change?"
answer yes forever, rewriting `graph.json` **and** `GRAPH_REPORT.md` on every single
run. Both helpers now pop `built_at` too.

The same property is why the stamp is injectable: `test_to_json_field_order_stable_across_read_rebuild`
asserts `graph.json` is byte-identical across two writes, which no wall-clock field
can satisfy unless the caller can pin it. Those two round-trip tests already pinned
`built_at_commit="fixed"`; they now pin the stamp the same way, and their byte-identity
guarantee is unchanged.

## Testing

Full suite green (`pytest -q`), plus targeted work:

- **`tests/test_export.py`** — exact-format assertions on the stamp
  (`re.fullmatch` + strict `strptime` + a UTC-vs-local drift check), verbatim
  write-through when pinned, and presence outside a git repo where the commit is
  legitimately absent.
- **`tests/test_watch.py`** — both comparators treat two graphs differing only in
  `built_at` as equal, **and** still detect a real node change arriving alongside a
  new stamp (a test that can only pass is not a test).
- **`tests/test_serve_http.py`** — `graph_stats` output asserted by **full-string
  equality**, not substring: unchanged six lines for a graph without provenance,
  both lines present with the **complete 40-char SHA** (a truncating regression
  would slip past an `in` check), commit-only when there is no stamp, junk/non-string
  values ignored rather than rendered as `Built at: None`, and provenance following
  `project_path` instead of leaking from the server's default graph.

Also verified against three real production graphs (C++ and TypeScript, 405–15,613
nodes) rather than fixtures only: the full SHA each file carries is reported exactly,
no phantom `Built at:` line appears on graphs written before this change, the original
six lines keep their order, and a round-trip with two different stamps produces
differing bytes while both comparators still report "unchanged".

The exact-format and full-SHA assertions are deliberate: a previous provenance change
of mine shipped a wrong path shape because its test asserted `endswith(...)` instead
of the whole string.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 Full release notes with details on each version: [GitHub Releases](https://github.com/safishamsi/graphify/releases)
 
+## Unreleased
+
+- Feature: `graph_stats` now reports build provenance — `Built from commit:` and `Built at:` — so an agent querying over MCP can tell how stale the graph it is answering from actually is. `built_at_commit` was already written into `graph.json` and consumed by the HTML report and the CLI, but `node_link_graph` keeps only `data["graph"]` and dropped it on load, so the MCP surface never saw it; every existing graph gains the commit line without a rebuild. `graph.json` also records a new top-level `built_at` UTC stamp (`YYYY-MM-DDTHH:MM:SSZ`), which answers a different question than the commit — when the file was written, not which revision it describes — and is what a consumer that cannot stat the file needs in order to judge freshness. The stamp is injectable like the commit so byte-identity round-trips stay assertable, and it is excluded from the graph-comparison helpers, since a field that changes on every write would otherwise make every incremental rebuild look like a change and rewrite `graph.json` and `GRAPH_REPORT.md` forever.
+
 ## 0.9.49 (2026-08-24)
 
 - Feature: `graphify merge-graphs` now links a type declaration that two repos share — same fully-qualified namespace and name, from different repos — with a `same_type_as` edge, so a shared contract type is navigable across the repo boundary; two unrelated types that merely share a short name are not linked (#3007, thanks @durmazoguzhan).
--- a/graphify/export.py
+++ b/graphify/export.py
@@ -9,7 +9,7 @@ import re
 import shutil
 import sys
 from collections import Counter
-from datetime import date
+from datetime import date, datetime, timezone
 from pathlib import Path
 import networkx as nx
 from networkx.readwrite import json_graph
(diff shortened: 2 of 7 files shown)


winesoft-namjin-yun:
Superseded by #3355, which is the same change rebased onto 0.9.54 and re-verified there; the problem it fixes is now written up separately in #3354.

This branch was cut against 0.9.49 and had drifted (conflicts in `CHANGELOG.md`, `tests/test_serve_http.py`, `tests/test_watch.py`). Closing it rather than force-pushing so the rebased version arrives with the issue attached.

PR #3059, opened 2026-08-25
fix: capture HOF-wrapped CommonJS exports

### LINKED ISSUE #3035: JS extractor misses call-expression exports (exports.x = onCall(...)) — HOF-wrapped entry points invisible
**Version:** graphifyy 0.9.49 (`uv tool install`), Python 3.14.4, macOS (arm64)

**Summary**

Following on from #2524 (which fixed property-assigned function *expressions*), the JS extractor still misses function-valued exports whose right-hand side is a **call expression** rather than a function or arrow. The higher-order-function export idiom — `exports.name = wrapper(opts, handler)` — produces no node, so the exported symbol is invisible to `query`, `path`, and `explain`.

This matters more than the node count suggests, because in frameworks that use this idiom the wrapped exports are usually the module's *only* externally reachable entry points. Everything the graph does capture is internal helpers.

**Repro**

```js
const wrap = (f) => f;

function plainFn() {}                                  // captured
const arrowConst = () => {};                           // captured
const asyncArrow = async () => {};                     // captured
exports.assignedArrow = async (x) => x;                // captured (fixed by #2524)
exports.assignedFn = function namedExpr() {};          // captured (fixed by #2524)
module.exports.viaModule = () => {};                   // captured (fixed by #2524)
export function esmFn() {}                             // captured

exports.assignedCall = wrap(async (x) => x);           // NOT captured
module.exports = { objectKey: () => {} };              // NOT captured
```

```python
from pathlib import Path
r = extract([Path('probe.js')], cache_root=Path('/tmp/probe-cache'))
for n in r['nodes']:
    print(n.get('label'), n.get('source_location'))
```

Output — 9 nodes, with `assignedCall` and `objectKey` absent:

```
probe.js        L1
wrap()          L1
plainFn()       L2
arrowConst()    L3
asyncArrow()    L4
assignedArrow() L6
assignedFn()    L7
viaModule()     L8
esmFn()         L10
```

**Likely cause**

A query matching `assignment_expression` where the RHS is `function_expression` / `arrow_function` will not match when the RHS is a `call_expression`, even though a function literal appears as one of its arguments.

**Why it matters**

This is the standard shape of several very common frameworks:

```js
// Firebase Functions v2 — every callable in a functions/index.js
exports.myCallable = onCall({ secrets: [KEY] }, async (request) => { ... });

// Express-style middleware wrappers
exports.handler = asyncHandler(async (req, res) => { ... });

// React
export const Card = React.memo((props) => { ... });
export default connect(mapState)(Component);
```

In a Firebase Functions entry module, the effect is that the deployed callables are the only symbols *not* in the graph — the requires, secret bindings, and private helpers are all captured, but the exported endpoints are not. A client calling one of them by name (e.g. `httpsCallable('myCallable')`) then has no node to resolve against, so the client and backend appear as disconnected components and cross-boundary `path` queries return nothing. The graph is not wrong about anything it contains; the entry points are simply missing.

**Expected**

`exports.name = someWrapper(..., fn)` and `module.exports = { name: fn }` captured as nodes under `name`, consistent with the ids already produced for the `#2524` forms. Unwrapping only needs to be one level deep to cover the common cases; the binding name on the left is what callers reference, so the node can be keyed on it without analyzing the wrapper.

Happy to test a branch against a Firebase Functions module and report node counts.

hopstreax:
## Summary

Fixes #3035.

The JavaScript extractor currently captures CommonJS exports when the right-hand side is directly a function or arrow function, but misses higher-order-function (HOF) exports such as:

```js
exports.myCallable = onCall({ secrets: [KEY] }, async (request) => {
    // ...
});
```

As a result, the exported entry point is absent from the graph even though its wrapped callback is the externally reachable handler.

This change extends CommonJS export extraction to recognize HOF-wrapped call expressions and associate their top-level closures with the exported callable node.

## Changes

* Reuse the existing `_js_topmost_closures()` mechanism for HOF-wrapped exports.
* Preserve existing `exports.foo` and `module.exports.foo` callable-node semantics.
* Associate discovered closure bodies with the exported node so calls inside wrapped handlers are attributed correctly.
* Preserve closure-local scoping behavior.
* Keep arbitrary member assignments such as `obj.foo = wrapper(...)` excluded.

## Tests

* `exports.foo = wrapper(async () => {})`
* `module.exports.foo = wrapper(function () {})`
* HOF wrappers with options before the callback.
* Calls inside HOF-wrapped export callbacks.
* Rejection of arbitrary member HOF assignments.

Validation:

* `tests/test_extract.py`: 41 JS/TS tests passed
* `tests/test_js_callback_calls.py`: 8 passed
* `tests/test_js_import_resolution.py`: 63 passed
* Full validation: 272 passed, 4 skipped

There is one unrelated Windows C `#include` long-path test failure.

## Scope

`module.exports = { foo: () => {} }` is intentionally not addressed here because it follows a separate object-literal export extraction path.

### DIFF
--- a/graphify/extractors/engine.py
+++ b/graphify/extractors/engine.py
@@ -2149,33 +2149,59 @@ def _js_extra_walk(node, source: bytes, file_nid: str, stem: str, str_path: str,
                        if c.type == "assignment_expression"), None)
         if assign is not None:
             value = assign.child_by_field_name("right")
-            if value is not None and value.type in _JS_FUNCTION_VALUE_TYPES:
+            if value is not None:
                 target = _js_member_assignment_target(
                     assign.child_by_field_name("left"), source)
                 if target is not None:
                     kind, owner_name, member_name = target
                     line = node.start_point[0] + 1
-                    handled = False
-                    if kind == "exports":
-                        nid = _make_id(stem, member_name)
-                        add_node_fn(nid, f"{member_name}()", line)
-                        add_edge_fn(file_nid, nid, "contains", line)
-                        handled = True
-                    elif kind == "prototype":
-                        owner_nid = _make_id(stem, owner_name)
-                        nid = _make_id(owner_nid, member_name)
-                        add_node_fn(nid, f".{member_name}()", line)
-                        add_edge_fn(owner_nid, nid, "method", line)
-                        handled = True
-                    if handled:
-                        if callable_def_nids is not None:
-                            callable_def_nids.add(nid)  # CJS/prototype fn is callable
-                        if local_bound_names is not None:
-                            local_bound_names[nid] = _js_local_bound_names(value, source)
-                        body = value.child_by_field_name("body")
-                        if body:
-                            function_bodies.append((nid, body))
-                        return True
+                    if value.type in _JS_FUNCTION_VALUE_TYPES:
+                        handled = False
+                        if kind == "exports":
+                            nid = _make_id(stem, member_name)
+                            add_node_fn(nid, f"{member_name}()", line)
+                            add_edge_fn(file_nid, nid, "contains", line)
+                            handled = True
+                        elif kind == "prototype":
+                            owner_nid = _make_id(stem, owner_name)
+                            nid = _make_id(owner_nid, member_name)
+                            add_node_fn(nid, f".{member_name}()", line)
+                            add_edge_fn(owner_nid, nid, "method", line)
+                            handled = True
+                        if handled:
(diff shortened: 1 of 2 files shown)


safishamsi:
Shipped in v0.9.50 via authorship-preserving cherry-pick so you keep contributor-graph credit. Thanks @hopstreax! Release: https://github.com/Graphify-Labs/graphify/releases/tag/v0.9.50

PR #3065, opened 2026-08-25
feat(csharp): emit a node per enum member

### LINKED ISSUE #3063: C#: enum members are not nodes, so the enum type is a leaf and "which value does this branch on" has no answer
### Environment

- graphify 0.9.49 (PyPI `graphifyy`), Linux, Python 3.12
- `graphify extract <service> --code-only --no-cluster`

### Summary

C# enum members are not extracted as nodes. `enum_declaration` is in `_CSHARP_CONFIG.class_types`, so the enum **type** gets a node, but nothing under it does, and the type sits in the graph as a leaf with no outgoing edges.

This is the same gap #1700 reported for Kotlin/Java, fixed for Java in #1719 and Kotlin in #1738. Swift emits `case_of` for `enum_entry`, and PowerShell enum members landed in 0.9.49 via #3002. C# was not brought along.

### Reproducer

```csharp
// ProcessingType.cs
namespace Catalog.Events;

public enum ProcessingType
{
    Process = 0,
    ProcessAndCompleteIndex = 1
}
```

### Current behavior (0.9.49)

```
processingtype                                (file)
csharp_namespace:9b597bd1e93a863b             (namespace)
processingtype_catalog_events_processingtype  (enum type, no children)
```

`Process` and `ProcessAndCompleteIndex` are nowhere in the node table, and the enum node has zero outgoing edges.

Measured on a 7-service .NET corpus (42,645 nodes / 98,269 edges): **zero `case_of` edges**, and the enum types I checked (`ProcessingType`, `EntityType`, `ProductType`, `VariantType`) are all leaves.

### Expected

The Java shape, applied to C#:

```
processingtype_catalog_events_processingtype --case_of--> ..._processingtype_process
processingtype_catalog_events_processingtype --case_of--> ..._processingtype_processandcompleteindex
```

### Why it matters here

In this corpus enum values are dispatch keys, not decoration. A consumer branches on `EntityType` to decide which id-resolution path to take, and a bulk index run is finished by one specific `ProcessingType` value. Asking the graph "which value completes the index run" currently cannot even surface that the value exists, because the name is not in the node table.

Same as #1700 concluded: even without a usage edge, having the member as a searchable node lets you find the definition and fall back to text search for the uses.

### Note on grammar

The walk already reaches the members. `enum_declaration`'s `body` field resolves to `enum_member_declaration_list`, whose children are `enum_member_declaration` nodes carrying a `name` field, and the C# class handler descends into the body with the enum as `parent_class_nid`. There is just no handler for that node type.

One C#-specific wrinkle: the language is case-sensitive, so `enum E { Value, value }` is legal while `normalize_id` casefolds, which means both members compete for one id.

durmazoguzhan:
Closes #3063.

`enum_declaration` is in `_CSHARP_CONFIG.class_types`, so a C# enum type was already a node. Its members were not, which left the type a leaf: nothing in the graph could say which values an enum has, let alone which one a consumer branches on.

Java has emitted a node per `enum_constant` with a `case_of` edge since #1719 and Kotlin per `enum_entry` since #1738, both from #1700; Swift does the same for `enum_entry`. This applies that shape to C#. No new relation, no config change.

### What it does

The C# class handler already descends into `enum_member_declaration_list` with the enum as `parent_class_nid`, so the members were reachable and simply unhandled. `_csharp_extra_walk` now takes `enum_member_declaration`, reads the `name` field, and emits a node plus a `case_of` edge from the enum.

```
processingtype_catalog_events_processingtype --case_of--> ..._process
processingtype_catalog_events_processingtype --case_of--> ..._processandcompleteindex
```

### The one C#-specific wrinkle

C# is case-sensitive, so `enum E { Value, value }` is legal, while `normalize_id` casefolds and puts both members on one id. Emitting both would hang a second edge on the first member's node. The first declaration keeps it, the same `seen_ids` guard the property nodes use (#3006, where the review caught the equivalent case for `_count` / `Count`).

### Measured

On a 1,482-file .NET service, before this patch: **0** `case_of` edges. After: **284** edges across **65** enum types, every target a real sourced node, no dangling edges.

The values that appear are the ones that carry dispatch meaning in that codebase, e.g. `ProcessingType` resolves to `ProcessOnly` / `ProcessAndCompleteIndex` / `CompleteIndexOnly` and `EntityType` to `Brand` / `Category` / `Product` / `Variant`. Before, those names were not in the node table at all.

### Tests

`tests/test_csharp_enum_members.py`, 10 tests. Without the change 8 fail; with it all 10 pass. They cover the bare and assigned spellings, a namespaced enum, an enum nested in a class, the case-only sibling collision, an empty enum, that a member is not read as a method, that the enum type node survives, and that a class property and an enum member sharing a name stay two nodes.

Full suite: 4,979 passed. The 3 failures (2 falkordb, 1 labeling) reproduce on a clean `v8` checkout and are unrelated. `ruff` clean.

### DIFF
--- a/graphify/extractors/engine.py
+++ b/graphify/extractors/engine.py
@@ -2385,7 +2385,30 @@ def _csharp_extra_walk(node, source: bytes, file_nid: str, stem: str, str_path: 
                        nodes: list, edges: list, seen_ids: set, function_bodies: list,
                        parent_class_nid: str | None, add_node_fn, add_edge_fn,
                        walk_fn, namespace_stack: list[str], scope_stack: list[str]) -> bool:
-    """Handle C# namespaces and transparent class-member wrappers."""
+    """Handle C# namespaces, enum members, and transparent class-member wrappers."""
+    if node.type == "enum_member_declaration" and parent_class_nid:
+        # `enum_declaration` is in C#'s class_types, so the enum type is a node
+        # but its members were not, leaving the type a leaf: "which value does
+        # this consumer branch on" had no answer. Java has emitted a node per
+        # `enum_constant` with a `case_of` edge since #1719, Kotlin since #1738,
+        # and Swift does the same for `enum_entry`; C# reaches the members
+        # through the same walk, so this is the Java shape applied here.
+        name_node = node.child_by_field_name("name")
+        if name_node is None:
+            return True
+        member_name = _read_text(name_node, source)
+        if not member_name:
+            return True
+        line = node.start_point[0] + 1
+        member_nid = _make_id(parent_class_nid, member_name)
+        # C# is case-sensitive, so `enum E { Value, value }` is legal, but the id
+        # recipe casefolds — both members normalize to one id. Emitting the
+        # second would hang a duplicate edge on the first member's node, so the
+        # first declaration keeps it (same guard as the property nodes in #3006).
+        if member_nid not in seen_ids:
+            add_node_fn(member_nid, member_name, line)
+            add_edge_fn(parent_class_nid, member_nid, "case_of", line)
+        return True
     if node.type == "namespace_declaration":
         ns_name = _csharp_namespace_name(node, source)
         pushed = False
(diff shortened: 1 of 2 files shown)


safishamsi:
Shipped in v0.9.50 via authorship-preserving cherry-pick so you keep contributor-graph credit. Thanks @durmazoguzhan! Release: https://github.com/Graphify-Labs/graphify/releases/tag/v0.9.50

PR #3066, opened 2026-08-25
feat(typescript): emit a node per enum member

### LINKED ISSUE #3064: TypeScript: enum members are not nodes, so a string enum's type is a leaf and the values are unsearchable
### Environment

- graphify 0.9.49 (PyPI `graphifyy`), Linux, Python 3.12
- `graphify.extract.extract([...])`

### Summary

TypeScript enum members are not extracted as nodes. `enum_declaration` is in `_TS_CONFIG.class_types` and the comment there reads `# parity with Java/C#`, but only the container half of that parity is implemented: the enum **type** gets a node, its members do not, and the type sits in the graph as a leaf.

This is the same gap #1700 reported for Kotlin/Java, fixed for Java in #1719 and Kotlin in #1738. Swift emits `case_of` for `enum_entry`, and PowerShell enum members landed in 0.9.49 via #3002.

### Reproducer

```typescript
// color.ts
export enum Direction {
    Up = "UP",
    Down = "DOWN"
}
```

### Current behavior (0.9.49)

```
color_ts           color.ts
color_direction    Direction     (no children)
```

`Up` and `Down` are not in the node table, and `Direction` has no outgoing edges. String enums like this are the common TS spelling, so the members carry the meaning and the type name alone carries very little.

### Expected

The Java shape:

```
color_direction --case_of--> color_direction_up
color_direction --case_of--> color_direction_down
```

### Grammar notes

The walk already reaches the members: `enum_declaration`'s `body` field resolves to `enum_body`, and the class handler descends into its children with the enum as `parent_class_nid`. There is no handler for the member node types.

TS spells a member two ways, and both need covering:

- a bare `Red` is a `property_identifier`
- `Green = 5` is an `enum_assignment` whose `name` field is a `property_identifier`, or a `string` when the member is quoted (`"Odd Name" = 7`), where the label wants the `string_fragment` rather than the quoted literal

`property_identifier` also appears all over a normal TS file, so a handler has to be scoped to nodes whose parent is `enum_body` rather than matching the type alone.

One more wrinkle shared with C#: TS is case-sensitive while `normalize_id` casefolds, so `enum E { Value, value }` puts two legal members on one id.

durmazoguzhan:
Closes #3064.

`enum_declaration` is in `_TS_CONFIG.class_types` with the comment `# parity with Java/C#`, but only the container half was implemented. The enum type was a node, its members were not, and the type sat in the graph as a leaf.

Java has emitted a node per `enum_constant` with a `case_of` edge since #1719 and Kotlin per `enum_entry` since #1738, both from #1700; Swift does the same for `enum_entry`. This finishes the parity the config comment already claims. No new relation, no config change.

### What it does

The class handler already descends into `enum_body` with the enum as `parent_class_nid`, so the members were reachable and unhandled. `_ts_extra_walk` now emits a node and a `case_of` edge for each one.

```
color_direction --case_of--> color_direction_up
color_direction --case_of--> color_direction_down
```

### Three details that shaped the patch

**Two spellings.** A bare `Red` is a `property_identifier`; `Green = 5` is an `enum_assignment` carrying the name in its `name` field. Both had to be handled or half the members would be missing depending on how the enum was written.

**Quoted members.** `"Odd Name" = 7` puts a `string` in the `name` field, so the text is unquoted before it becomes a label. Otherwise the node reads `"Odd Name"`, quotes included, and does not match a search for the member. This goes through the same two lines the namespace handler in the same function already uses, which matters: an escape splits the string into several fragments, so any rule that reads one fragment truncates `"A\tB"` to `A`.

**Scoping.** `property_identifier` appears throughout a normal TS file, so matching on the node type alone would have swept up every class property and object key. The handler requires the node's parent to be `enum_body`. There is a test asserting a class with a property and a method produces no `case_of` edges at all.

Also, as in C#, TS is case-sensitive while `normalize_id` casefolds, so `enum E { Value, value }` puts two legal members on one id; the first declaration keeps the node.

### Tests

`tests/test_typescript_enum_members.py`, 12 tests. They cover both spellings, quoted members, string-valued enums (checking the value is not mistaken for the name), `const enum`, an enum inside a namespace, the case-only collision, an empty enum, that the enum type node survives, the scoping guard above, and that claiming a member does not swallow its initializer.

Two of them came out of review:

- Returning early after claiming an `enum_assignment` stopped the walk before the value expression, and `A = class Inner { m() {} }.name` lost Inner's method node, which `v8` emits today. The handler now descends into the `value` field, and a full node/edge diff against `v8` on an enum with call, member-access, shift, and template initializers shows the `case_of` edges added and nothing removed.
- Reading a `string_fragment` truncated an escaped quoted name. Replaced with the file's existing unquoting rule, as described above.

Full suite: 4,982 passed. The 3 failures (2 falkordb, 1 labeling) reproduce on a clean `v8` checkout and are unrelated. `ruff` clean.

### Note

#3065 does the same thing for C#, which has the identical gap. They are separate because the grammars differ enough that the patches share no code, and each is reviewable on its own.

### DIFF
--- a/graphify/extractors/engine.py
+++ b/graphify/extractors/engine.py
@@ -2360,14 +2360,14 @@ def _ts_extra_walk(node, source: bytes, file_nid: str, stem: str, str_path: str,
         name_node = node if node.type == "property_identifier" else node.child_by_field_name("name")
         member_name = ""
         if name_node is not None:
+            member_name = _read_text(name_node, source)
             if name_node.type == "string":
-                # `"Odd Name" = 7`: read the fragment so the label is the member
-                # name rather than the quoted literal.
-                fragment = next(
-                    (c for c in name_node.children if c.type == "string_fragment"), None)
-                member_name = _read_text(fragment, source) if fragment is not None else ""
-            else:
-                member_name = _read_text(name_node, source)
+                # `"Odd Name" = 7`: the label is the member name, not the quoted
+                # literal. Unquote the whole text the way the namespace handler
+                # below does rather than reading a `string_fragment`, because an
+                # escape splits the string into several fragments and the first
+                # one alone truncates the name (`"A\tB"` would become `A`).
+                member_name = member_name.strip("'\"`")
         if member_name:
             line = node.start_point[0] + 1
             member_nid = _make_id(parent_class_nid, member_name)
--- a/graphify/extractors/engine.py
+++ b/graphify/extractors/engine.py
@@ -2329,7 +2329,7 @@ def _ts_extra_walk(node, source: bytes, file_nid: str, stem: str, str_path: str,
                    nodes: list, edges: list, seen_ids: set, function_bodies: list,
                    parent_class_nid: str | None, add_node_fn, add_edge_fn,
                    walk_fn) -> bool:
-    """Emit a container node for a TS `namespace`/`module` declaration.
+    """Emit enum member nodes, and a container node for a TS `namespace`/`module`.
 
     `namespace Foo {}` parses as `internal_module` (with `name`/`body` fields);
     `module Bar {}` and ambient `declare module "pkg" {}` parse as a named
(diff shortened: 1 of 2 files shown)


durmazoguzhan:
Good catch, and it reproduces. Fixed in the commit above.

The suppression is real, and it is observable today. A class expression in an initializer loses its members:

```typescript
export enum E {
    A = class Inner { m() { return 1; } }.name.length
}
```

On `v8`: nodes `E`, `m()`, `s.ts`. On the first version of this branch: `A`, `E`, `s.ts`. The `m()` node was gone, because returning True stopped the walk before the initializer.

I went looking for a smaller reproduction first and did not find one. `A = compute()` emits no call edge on either side, and a `function inner() {}` in an initializer produces nothing on either side, so most initializer shapes were unaffected. The class expression is the case that already had extraction behind it, which is what made the loss visible.

The fix descends into the `value` field after claiming the member, and only the `value`: walking the whole node would put the `name` through the default recurse as a second node. Diffing the full node and edge sets against `v8` on an enum with call, member-access, shift, and template initializers now shows the four `case_of` edges added and nothing removed.

Added `test_claiming_a_member_does_not_swallow_its_initializer` pinning the class-expression case. Without the fix it fails; the file is now 11 tests, all passing. Full suite 4,979 passed, same 3 unrelated pre-existing failures.

durmazoguzhan:
Also correct, and the cause was that I invented a rule instead of using the one already here.

`"A\tB"` parses as `"` · fragment `A` · escape_sequence `\t` · fragment `B` · `"`, so taking the first `string_fragment` labelled the member `A`. Anything with an escape in a quoted name was silently shortened.

The namespace handler further down this same function already unquotes a TS string name:

```python
ns_name = _read_text(name_node, source)
if name_node.type == "string":
    ns_name = ns_name.strip("'\"`")
```

That reads the whole text, so escapes cannot split it. `string_fragment` appeared nowhere else in the codebase before this branch, which is a fair sign I should have followed the sibling rather than reach for the grammar. The member name now goes through the same two lines, and the escape is kept as written rather than decoded, since the label is source text.

Added `test_an_escape_in_a_quoted_member_does_not_truncate_the_name`. Without the change it fails; the file is 12 tests now, all passing. Full suite unchanged apart from the new test.

One thing I deliberately did not change: `strip` removes every leading and trailing quote character, so a member named `'x'` written as `"'x'"` loses the inner quotes too. The namespace handler has the same behaviour, and making the enum path stricter on its own would leave two different rules for the same node type a few lines apart. Happy to tighten both together if you would rather have that.

safishamsi:
Shipped in v0.9.50 via authorship-preserving cherry-pick so you keep contributor-graph credit. Thanks @durmazoguzhan! Release: https://github.com/Graphify-Labs/graphify/releases/tag/v0.9.50

PR #3068, opened 2026-08-25
fix(build): offset community ids per input in merge-graphs

### LINKED ISSUE #3014: `merge-graphs` does not namespace the `community` node attribute, so communities collide across repos and the aggregated HTML view fuses unrelated ones into a single meta-node
`merge-graphs` prefixes node ids with the repo tag, rewrites edge `_src`/`_tgt`, prefixes hyperedge ids and remaps their members, and stamps `repo` on every node — but the per-node `community` attribute is carried across unchanged. Each input graph numbers its communities from 0, so the merged graph inherits N overlapping id spaces.

That is invisible in `graph.json` itself, but the aggregated community view (taken for any merged graph above the 5000-node viz limit) groups by the bare community id, so communities from different repos that happen to share an integer are drawn as one meta-node.

### Reproduction

Merging three unrelated repos (11707 + 2057 + 1892 nodes; 754 + 203 + 172 communities):

```
$ graphify merge-graphs A/graphify-out/graph.json B/graphify-out/graph.json C/graphify-out/graph.json --out merged/graphify-out/graph.json
Merged 3 graphs -> 15656 nodes, 25936 edges

distinct community ids: 754        # expected 1129 = 754 + 203 + 172
community ids shared by >1 repo: 203
```

754 is exactly the largest input's community count, which is what makes it read as plausible rather than wrong.

Community 0 in the merged graph holds 227 nodes drawn from all three repos:

```
community 0 -> A: 114 nodes  "Preflight Check Suite"
               B:  49 nodes  "Canary Disk Budget"
               C:  64 nodes  "Heading Declaration Gate Tests"
```

Three unrelated communities, one meta-node:

```
$ graphify export html --graph merged/graphify-out/graph.json
Graph has 15656 nodes (above 5000 limit). Building aggregated community view...
graph.html written (aggregated: 754 community nodes, 1577 cross-community edges)
```

After namespacing `(repo, community)` locally, the same graph exports 1129 community nodes.

### Mechanism

`build.py::prefix_graph_for_global` disambiguates everything else per repo but never touches `community`. Its own comment credits PR #1691's diagnosis for the hyperedge half:

```python
relabel = {n: f"{repo_tag}::{n}" for n in G.nodes}
H = nx.relabel_nodes(G, relabel, copy=True)
for node, data in H.nodes(data=True):
    data["repo"] = repo_tag
    data.setdefault("local_id", node.split("::", 1)[1])
# ... edges (_src/_tgt), hyperedge ids and members all remapped ...
# 'community' is not remapped
```

The consumer side then groups on the bare id — `cli.py`, the `export html` path that reconstructs communities from node attributes:

```python
reconstructed: dict[int, list[str]] = {}
for node_id, data in G.nodes(data=True):
    cid_raw = data.get("community")
    ...
    reconstructed.setdefault(cid, []).append(str(node_id))
```

and `exporters/html.py` emits one meta-node per key:

```python
node_to_community = {nid: cid for cid, members in communities.items() for nid in members}
for cid, members in communities.items():
    meta.add_node(str(cid), label=(community_labels or {}).get(cid, f"Community {cid}"))
```

`merge-graphs` writes no `.graphify_analysis.json`, so a merged graph always takes that reconstruction path rather than a sidecar.

### Prior art

PR #1691 ("fix(merge): preserve repository-local identities") proposed keeping "existing community partitions distinct for each repository" and noted "Community IDs overlap". It was closed unmerged in favour of the fix shipped for #2484, which covers the hyperedge half of the same PR; the community half does not appear in the shipped function. #2134 separately describes `merge-graphs` and the global graph as unioning "repositories side by side", which matches what I see.

### Workaround, and why I still think this is worth fixing

`graphify cluster-only <merged-dir> --graph <merged graph.json> --no-label` re-partitions the merged graph and yields unique ids (1044 communities here, none spanning a repo). It discards the curated per-repo community names, so those have to be regenerated.

So there is a one-command fix, and you could reasonably say the raw merge is simply un-analysed and re-clustering was the missing step. Two things make that a harder sell from the user side:

- `extract` tells you to do it and `merge-graphs` does not. `extract` prints a "next: run `graphify cluster-only ...`" line; `merge-graphs` prints only the node/edge counts and the output path, and `references/github-and-merge.md` goes straight from `merge-graphs` to querying the merged graph with no re-cluster step.
- Nothing detects the bad state. `export html` silently reconstructs communities from the node attribute and renders the fused view; `export wiki` does have a guard for untrustworthy community data, but it only fires when `communities` is *empty*, and here it is populated and wrong. So the failure is silent in every path.

Even just prefixing the ids at merge time, or printing the same "run cluster-only next" hint that `extract` prints, would close it.

### Suggested fix

Namespace `community` in `prefix_graph_for_global` alongside the hyperedge ids that are already prefixed there — either `f"{repo_tag}::{cid}"` or an offset per input — and key any combined `community_labels` the same way. Keeping the pre-merge value in something like `local_community` would preserve the per-repo partition for callers that want it.

### Environment

graphify 0.9.49 (pipx), Python 3.13, Debian 13. Also reproduced on 0.9.48.

santhiprakash:
## What

`graphify merge-graphs` prefixes node ids, rewrites edge `_src`/`_tgt`, prefixes hyperedge ids and stamps `repo` on every node — but the per-node `community` attribute is carried across unchanged. Since every input graph numbers its communities from 0, the merged graph inherits overlapping id spaces, and the aggregated community view (taken for any merged graph above the 5000-node viz limit) groups by the bare community id — fusing communities from different repos that happen to share an integer into a single meta-node.

From the repro in #3014: merging three unrelated repos (754 + 203 + 172 communities) yields **754** distinct community ids instead of 1129, and community 0 alone holds 227 nodes drawn from all three repos.

## Fix

- `prefix_graph_for_global` gains a `community_offset` parameter: integer community ids are shifted into a shared id space, and the original id is kept in `local_community` so the per-repo partition stays recoverable. The default `0` leaves communities untouched, so the global-store push path and existing callers keep their exact behavior.
- The `merge-graphs` loop advances the offset past each input's highest id as it composes. The first input keeps its original ids; inputs without community data leave the offset alone.
- Integer offsets rather than namespaced `repo::cid` strings (the other option suggested in #3014) deliberately: every existing consumer keeps working unchanged — the html export's reconstruction (`int(cid_raw)`), `community_labels` lookups, and the serve/watch/prs display paths all still see ints.

## Tests

- `test_merge_graphs_offsets_communities_so_repos_do_not_fuse` (end-to-end `merge-graphs`): two inputs with colliding ids produce disjoint id spaces, all five distinct communities survive, `local_community` preserves each repo's partition. Fails on the pre-fix tree.
- `test_prefix_graph_offsets_community_ids` / `test_prefix_graph_zero_offset_leaves_communities_untouched`: unit contract for the new parameter and the zero-offset no-op.
- `tests/test_global_graph.py`, `tests/test_merge_graphs_cli.py`, `tests/test_hooks.py`: 124 passed; full suite 2303 passed / 19 skipped with the one pre-existing `test_built_wheel_ships_the_full_skill_payload` env failure (fails on a clean tree too — needs a built wheel).

Closes #3014

### DIFF
--- a/graphify/build.py
+++ b/graphify/build.py
@@ -1980,19 +1980,31 @@ def build_merge(
     return G
 
 
-def prefix_graph_for_global(G: nx.Graph, repo_tag: str) -> nx.Graph:
+def prefix_graph_for_global(
+    G: nx.Graph, repo_tag: str, community_offset: int = 0
+) -> nx.Graph:
     """Return a copy of G with all node IDs prefixed with repo_tag::.
 
     Labels are preserved unchanged (for display). A 'local_id' attribute
     is added to each node so the original ID can be recovered. Edges and
     their directional attributes (_src/_tgt) are rewritten to match the new
     prefixed IDs. The 'repo' attribute is set on every node.
+
+    community_offset shifts each node's integer 'community' id into a shared
+    id space and records the original in 'local_community': every input graph
+    numbers its communities from 0, so ids carried across a merge unchanged
+    collide and the aggregated community view fuses unrelated communities
+    into one meta-node (#3014). 0 (the default) leaves communities untouched.
     """
     relabel = {n: f"{repo_tag}::{n}" for n in G.nodes}
     H = nx.relabel_nodes(G, relabel, copy=True)
     for node, data in H.nodes(data=True):
         data["repo"] = repo_tag
         data.setdefault("local_id", node.split("::", 1)[1])
+        cid = data.get("community")
+        if community_offset and isinstance(cid, int):
+            data["local_community"] = cid
+            data["community"] = cid + community_offset
     for u, v, data in H.edges(data=True):
         if "_src" in data and data["_src"] in relabel:
             data["_src"] = relabel[data["_src"]]
--- a/graphify/cli.py
+++ b/graphify/cli.py
@@ -2537,11 +2537,25 @@ def dispatch_command(cmd: str) -> None:
         # diagnosis in PR #1691). Collect every input's prefixed hyperedges
         # and re-attach the union after composing.
         collected_hyperedges: list = []
+        # Offset each input's community ids into a shared id space as it is
+        # prefixed: every input numbers its communities from 0, so ids carried
+        # across unchanged collide in the merged graph and the aggregated
+        # community view fuses unrelated communities into one meta-node (#3014).
+        # The first input keeps its original ids (offset 0); local_community on
+        # the prefixed nodes preserves each repo's own partition.
+        community_offset = 0
         for G, repo_tag in zip(graphs, repo_tags):
-            prefixed = _to_simple(_prefix(G, repo_tag))
+            prefixed = _to_simple(_prefix(G, repo_tag, community_offset=community_offset))
             hes = prefixed.graph.get("hyperedges")
             if isinstance(hes, list):
                 collected_hyperedges.extend(h for h in hes if isinstance(h, dict))
+            cids = [
+                d["community"]
+                for _, d in prefixed.nodes(data=True)
+                if isinstance(d.get("community"), int)
+            ]
+            if cids:
+                community_offset = max(community_offset, max(cids) + 1)
             merged = _nx.compose(merged, prefixed)
         # A contract type both repos declare arrives as two unconnected nodes,
         # since every id is repo-prefixed. Link them so a traversal can cross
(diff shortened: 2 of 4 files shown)


safishamsi:
Shipped in v0.9.50 via authorship-preserving cherry-pick so you keep contributor-graph credit. Thanks @santhiprakash! Release: https://github.com/Graphify-Labs/graphify/releases/tag/v0.9.50

Issue #2312, opened 2026-07-29
Proposal: optional Neo4j source-of-truth backend (branch-tagged, read/write), not just an export sink


NiSHoW:
## Motivation

Today Neo4j is a one-way, lossy export (`export neo4j --push`): no original `relation`/`file_type` strings, no hyperedges, no stale-node cleanup, no branch concept, and nothing ever reads it back. I built (and have been running) an **opt-in backend mode** where Neo4j becomes the graph's source of truth, and I'd like to know whether a PR for it would be welcome — and whether the design below fits your direction before I open one.

Working implementation: https://github.com/NiSHoW/graphify/tree/neo4j-backend (two commits on top of 0.9.30, ~2000 lines incl. tests; default behavior unchanged, `neo4j` stays an optional extra).

## Design summary

- **Opt-in**: `graphify backend set neo4j://host:7687[/db]` writes `graphify-out/backend.json` (no credentials — password only via `NEO4J_PASSWORD`/`GRAPHIFY_NEO4J_PASSWORD`, matching the F-031 precedent), verifies connectivity, creates constraints/indexes. `backend unset` reverts. Env override: `GRAPHIFY_NEO4J_URI`.
- **Branch tagging**: every node/edge carries the current git branch (`uid = branch + "\x00" + id`, unique constraint; detached HEAD → `detached-<sha>`, non-repo → `_default`, `GRAPHIFY_BRANCH` override). Branches coexist in one DB (community edition = one database); `graphify branches [--delete NAME|--prune]` manages them. A per-branch `:GraphifyMeta` node holds a monotonically increasing `version`, `built_at_commit`, counts, and hyperedges.
- **Round-trip-faithful schema**: original `relation`/`file_type` strings stored as properties (sanitized forms only as Cypher labels/rel types), underscore/None/non-scalar attributes ride an `extra_json` property, so `load == save` exactly.
- **Write path**: `extract`/`update` keep the existing file pipeline untouched — the branch's graph is materialized from Neo4j into `graphify-out/graph.json` before the rebuild (so reconcile/shrink-guard work against the DB's true state), and the **delta** is pushed after, UNWIND-batched in one transaction whose *last* statement bumps `version`. Fail-closed when configured but unreachable; a failed/killed push marks a local dirty flag so the next run re-diffs against the DB and repairs it. `GRAPHIFY_NEO4J_TX_ROWS=<n>` opts into chunked commits for very large seeds; pushes over ~2000 rows report progress on stderr.
[...]

PR #3069, opened 2026-08-25
fix(falkordb): make the graph-DB push converge, batch its writes, add a repo-keyed delta (#3057)

### LINKED ISSUE #3057: Graph-DB push never deletes: a pruned repo survives in Neo4j/FalkorDB forever (measured), and a delta push fixes it
## The defect

`push_to_neo4j` / `push_to_falkordb` are MERGE-only. That makes them idempotent, which the docstring correctly claims — but it also means **they never remove anything**. Once `graphify global add` prunes a repo's stale nodes out of the global graph (`prune_repo_from_graph`), a full re-push leaves every one of those nodes in the database, permanently. The database silently diverges from `graph.json` and never converges back, no matter how many times you re-push.

This is the "no stale-node cleanup" that @NiSHoW listed in #2312, measured.

## Measurement

Live FalkorDB, 75,000-node / 40-repo global graph, one repo refreshed exactly the way `global add` refreshes one (-1,250 pruned, +1,750 added — 2.5% of the graph):

| | nodes | edges | equals a from-scratch load? |
|---|---:|---:|---|
| ground truth (state B into an empty graph) | 75,500 | 74,164 | — |
| **full re-push** of B onto A | 76,750 | 75,315 | **no** — +1,250 / +1,151 |

The surplus is exactly the pruned nodes and their edges. Of 25 sampled pruned node ids, **all 25** are still present after the full re-push. Equality was checked on node/edge counts, per-repo counts, per-relation counts, and a SHA-256 over every node's `id + community + file_type`.

Worth stating plainly: **the full push is not the self-healing option it looks like.** It pays the whole cost *and* drifts.

## The cost side, since it is the same code path

On a 1.2M-node / 2.9M-edge global graph the full push runs ~7 min wall / ~4 min CPU per run, and re-sends all 4.1M rows even when one repo of 226 changed (~2% of the graph). Note for anyone measuring this: streaming `graph.json` end to end is **138s CPU of that ~240s**, so the file read — not the DB writes — is the larger half. A row-level delta cannot beat that ceiling.

## What I did, and what it cost

I implemented a delta push on a fork and proved it against the same corpus. The unit of change is the **repo**, because that is already `global_add`'s unit: it prunes a repo whole, re-adds it whole, records a per-repo `source_hash` in the global manifest, and already returns `skipped=True` when that hash has not moved. The push side just mirrors that contract.

| | nodes | edges | wall | cpu | rows sent |
|---|---:|---:|---:|---:|---:|
| ground truth | 75,500 | 74,164 | 17.0s | 10.9s | — |
| full push B onto A | 76,750 | 75,315 | 14.7s | 9.7s | 149,664 |
| **delta push B onto A** | **75,500** | **74,164** | **6.4s** | **6.3s** | **3,960** (+1,875 deleted) |
| delta, nothing changed | — | — | 0.14s | 0.24s | 0 |

The delta result is **identical** to the from-scratch load on every measure, including the content checksum. 2.6% of the rows; 56% of the wall time — the gap between those two numbers is the file scan.

## Three things that bit me, in case they save someone else the trouble

1. **`LIMIT` does not page a DELETE in FalkorDB.** Its known-limitations doc says LIMIT "does not currently short-circuit eager operations like CREATE, SET, or DELETE" — so `MATCH (n:L) WHERE … DETACH DELETE n LIMIT 10000` deletes the **whole label**, not a page. The LIMIT has to sit in a `WITH` that precedes the DELETE.
2. **A ledger of "what I last pushed" is not enough.** If the database is wiped or a run half-lands, the ledger still says clean and the delta never repairs it. Reading the database's own per-repo counts (one indexed aggregate, ~0.4s on 1.2M nodes) and re-pushing any repo whose count disagrees with the manifest turns that from silent permanent drift into automatic repair.
3. **Deleting repos the manifest no longer lists needs a size guard.** A manifest that does not belong to the target database looks exactly like a genuine mass removal. During development I pointed a 40-repo test manifest at the default `graphify` graph name and it deleted **1,211,189 nodes** in one run — entirely my own fault, and entirely undetectable by inspection. Capping the drop at a fraction of the graph unless explicitly allowed is the same rule as the #479 shrink guard ("refuse to SILENTLY drop nodes"), applied to the push.

## Offer

Happy to send this as a PR if you want it. One caveat so it is not a surprise: it sits on top of two push changes I have on a fork and have not offered here — streaming `graph.json` into the push instead of loading it through NetworkX (the 1.87GB load peaked ~5.3GB RSS and got OOM-killed before a single row went out), and an index-driven push using one shared node label so edge-endpoint MATCHes hit an index. That second one is the same diagnosis @kingjin94 made in #2258, reached independently. So it would be up to three PRs, and I would rather ask first than dump them.

Also flagging: this overlaps #2312, which already proposes a delta push (with the database as source of truth, a larger design than mine) and has been open a while. If that is the direction you prefer, I would rather help there than compete with it.

Two questions:
1. Is "the push should remove what the source pruned" a change you want at all, or is the pushed store meant to stay a disposable projection you re-import? (The #1504 changelog note — "if you push to a persisted Neo4j store, re-import after upgrading" — reads like the latter, and that is a perfectly coherent position; it just is not written down anywhere as the intended contract.)
2. If you do want it, would you rather have the delta keyed on the global manifest's per-repo hashes (what I built, small) or the fuller #2312 shape?

galshubeli:
Fixes #3057. Targets `v8` directly — this is the upstream-mergeable version of the fix; see the note at the bottom.

## The defect

`push_to_falkordb` is MERGE-only. Once `graphify global add` prunes a repo out of the global graph, a full re-push leaves every one of those nodes in the target **permanently** — `graph.json` and the database diverge and never converge back, no matter how many times you re-push. @Azeem1985 measured it: +1,250 nodes / +1,151 edges of surplus on a 75k-node global graph, and 25 of 25 sampled pruned ids still present after a full re-push.

While fixing it, a second problem surfaced that is not in the report. Measured here, 20 repos / 20k nodes / 20k edges:

```
OLD per-row full push    79.73s
NEW batched full push     2.90s   (27x)
```

The old path sent one query per node and one per edge, and matched edge endpoints with a label-free `MATCH (a {id: $src})` that no index can serve — because nodes carried only their file-type label (`:Python`), there was no shared label to index. That is @kingjin94's #2258 diagnosis, and it is fixed here by giving every node `:Entity` alongside its file-type label and indexing `n.id`.

## What changed

**Batched writes on an indexed shared label.** `UNWIND` in 1000-row batches, `MERGE (n:Entity {id: row.id}) SET n:Python SET n += row`, with `CREATE INDEX FOR (n:Entity) ON (n.id)`. The file-type label is kept, not replaced, so existing queries keyed on it still work. A one-time backfill (`MATCH (n) WHERE n.id IS NOT NULL AND NOT n:Entity SET n:Entity`) adopts nodes written by an older graphify — without it, the MERGE switch would create a duplicate beside every existing node. There is a test asserting both labels survive.

**`--graph-name`.** The writer has always taken `graph_name`; the CLI never passed it, so every push through the CLI landed on the `graphify` key with no flag to send it elsewhere. You could not rehearse a push, run a staging graph beside production, or point two environments at one instance without dropping to Python. This is also what turned @Azeem1985's mistake into a 1.2M-node deletion.

**`--prune` converges the target — nodes *and* edges.** `DETACH DELETE` on a stale node takes its edges with it, but an edge dropped between two endpoints that both survive needs its own sweep; without it the surplus-edge half of the report lives on. Opt-in rather than default: `--graph-name` never existed, so some `graphify` keys genuinely hold several projects merged together, and a converging default would either silently delete the others (under the guard) or hard-refuse a routine re-push (over it). Neither is acceptable in a patch upgrade — @Azeem1985's framing, and it is better than my original reasoning.

**An add-only push reports the drift it cannot fix.** The core complaint in #3057 is that the divergence is *silent*, and that stays true on the default path if the only fix is a flag nobody passes. So a non-pruning push now counts the nodes it did not stamp — the same query the prune path deletes by, so there is one definition of drift rather than two that can disagree — and the CLI prints one line:

```
  note: target has 2 node(s) the source does not; --prune converges it.
```

**Repo-keyed delta**, via the new `graphify global push` (delta by default). This mirrors `global_add`'s own contract instead of inventing one: the repo is already its unit of change — it prunes a repo whole, re-adds it whole, records a per-repo `source_hash`, and already returns `skipped=True` when that hash has not moved. The push now consumes the same signal.

| | wall | rows sent |
|---|---:|---:|
| full push, nothing changed | 2.90s | 39,999 |
| delta, nothing changed | **0.01s** | **0** |
| delta, 1 of 20 repos changed | **0.39s** | **2,001** (5.0%) |

Delta output is identical to a from-scratch load on node and edge counts.

**Drift repair.** Push state lives in the *target* as `:GraphifyPushState` nodes and is cross-checked against the target's own per-repo counts (one aggregate). A local ledger still reads clean after the database is wiped or a run half-lands, and the delta would skip forever — @Azeem1985's finding #2. There is a test that deletes part of a repo behind the push's back, leaves the hashes untouched, and asserts the repo is rebuilt.

**Cross-repo edges survive a delta.** `global_add` remaps external-library nodes onto whichever repo first contributed them, so a `B -> A` edge is owned by neither repo alone. Pruning A drops it, and re-adding only A's *internal* edges would erode cross-repo connectivity a little on every delta — silently. The delta buckets edges by *either* endpoint's repo. Pinned by a test.

**Delete guards.** Capped at 20% of the target unless `--allow-shrink` — the same "refuse to SILENTLY drop nodes" rule as the #479 build guard. Only *net* removal counts: a re-pushed repo is pruned and immediately re-added, so charging it would refuse any delta touching more than 20% of a small global graph; a repo that comes back *smaller* is charged the difference, which is what catches "the manifest says 2 nodes, the target holds 50,000". Deletes are paged with the `LIMIT` inside a `WITH`, because FalkorDB's `LIMIT` does not short-circuit an eager `DELETE` — `... DETACH DELETE n LIMIT $page` deletes the whole label. That is @Azeem1985's finding #1, and the reason sits in a comment at the call site.

**`graphify global push`.** `export --push` resolves its source from a project's output directory, so the global graph — the one graph this entire contract is about — had no CLI push route at all.

## Convergence, measured

```
OLD full re-push, after the source pruned   -> 20,240 nodes   converged=False
NEW prune re-push                           -> 20,000 nodes   converged=True
```

## Tests

15 new integration tests against a live FalkorDB: both labels, add-only default, drift reporting, node convergence, edge-only convergence, prune idempotence, the guard refusing *and* not deleting, target isolation by `--graph-name`, delta skip / selective re-send / repo removal, cross-repo edge preservation, drift repair, and a foreign manifest being refused.

Verified they catch the bugs rather than just passing: **all 15 fail on this commit's parent, all pass here** (15 failed / 2 passed against the parent; the 2 are the pre-existing create/idempotence tests). Independently reproduced by @Azeem1985 on the rig that produced the original #3057 measurements. Full suite is +15 passing with no new failures (this environment has 557 pre-existing failures from missing tree-sitter grammars, identical before and after).

## Deliberately not included

**The Neo4j writer is untouched** — byte-identical to the parent, verified by AST comparison. It has the same defects (per-row, unindexed, never deletes, no `database` parameter), but a fix cannot be verified without a Neo4j instance. The new flags are **refused** on `export neo4j` rather than silently ignored: accepting `--prune` there and doing nothing would report a converged push that converged nothing — the same class of lie the missing-label bug told. Follow-up when there is an instance to test against.

No CHANGELOG entry — this repo folds those at release (e.g. 1c6b3db), per @Azeem1985.

Streaming `graph.json` into the push instead of loading it through NetworkX is **not** in here. It is a real problem (@Azeem1985 reports a 1.87GB graph peaking ~5.3GB RSS and getting OOM-killed before a row went out), but it is a separate change to how export loads graphs and belongs in its own PR, the way it was originally offered.

## Note on provenance

This work was first written against a FalkorDB-backend branch where `graphify/store.py` provides a `GraphStore` with a batched writer, and reviewed there by @Azeem1985 on the box that produced the original #3057 numbers ([review](https://github.com/galshubeli/graphify/pull/2#issuecomment-5407457062)). That version is not portable to `v8`, which has no `store.py` at all — every symbol it leaned on is absent, so a cherry-pick auto-merges four of five files and then ImportErrors on first use. This PR is the `v8`-native implementation of the same contract: the batched writer is hand-rolled in `graphdb.py`, and the delta buckets the in-memory graph by repo in two passes rather than querying the source. Same tests, same guards, same measured behaviour.

## Relationship to #2312

#2312 proposes a Neo4j source-of-truth backend. This is not competing with it — it is the push-side piece that design needs regardless, and it is small enough to land independently.

### DIFF
--- a/README.md
+++ b/README.md
@@ -758,10 +758,17 @@ graphify export callflow-html --max-sections 8      # cap generated architecture
 graphify export callflow-html --output docs/arch.html
 graphify export callflow-html ./some-repo/graphify-out
 
+graphify export falkordb --push falkordb://localhost:6379                    # push this project's graph
+graphify export falkordb --push falkordb://localhost:6379 --graph-name stg   # choose the target graph
+graphify export falkordb --push falkordb://localhost:6379 --prune            # mirror: delete what the source dropped
+
 graphify global add graphify-out/graph.json --as myrepo   # register a project graph into ~/.graphify/global-graph.json
 graphify global remove myrepo                         # remove a project from the global graph
 graphify global list                                  # show all registered repos + node/edge counts
 graphify global path                                  # print path to the global graph file
+graphify global push falkordb://localhost:6379        # push the global graph to FalkorDB (delta by default)
+graphify global push falkordb://host:6379 --graph-name staging   # pick the target graph in the instance
+graphify global push falkordb://host:6379 --full --prune         # re-send every repo and converge the target
 
 graphify prs                              # PR dashboard: CI, review, worktree, graph impact
 graphify prs 42                           # deep dive on PR #42
--- a/graphify/cli.py
+++ b/graphify/cli.py
@@ -2616,8 +2616,11 @@ def dispatch_command(cmd: str) -> None:
             print("  graphml   [--graph PATH]", file=sys.stderr)
             print("  neo4j     [--graph PATH] [--push URI] [--user U] [--password P]", file=sys.stderr)
             print("            (or set NEO4J_PASSWORD instead of --password to keep it off argv)", file=sys.stderr)
-            print("  falkordb  [--graph PATH] [--push URI] [--user U] [--password P]", file=sys.stderr)
+            print("  falkordb  [--graph PATH] [--push URI] [--user U] [--password P] [--graph-name NAME]", file=sys.stderr)
+            print("            [--prune] [--allow-shrink]", file=sys.stderr)
             print("            (or set FALKORDB_PASSWORD instead of --password to keep it off argv)", file=sys.stderr)
+            print("            --graph-name selects the target graph in the instance (default \"graphify\");", file=sys.stderr)
+            print("            --prune deletes what the source no longer has so the target mirrors it.", file=sys.stderr)
             sys.exit(1)
 
         # Parse shared args
@@ -2651,6 +2654,12 @@ def dispatch_command(cmd: str) -> None:
             os.environ.get("FALKORDB_PASSWORD") if subcmd == "falkordb"
             else os.environ.get("NEO4J_PASSWORD")
         ) or None
+        # Target selection inside the server (falkordb only). Never exposed
+        # before, so every CLI push landed on the "graphify" key regardless of
+        # what that key already held (#3057).
+        push_graph_name = "graphify"   # falkordb: named graph in the instance
+        push_prune = False         # falkordb: delete what the source no longer has
+        push_allow_shrink = False  # falkordb: override the prune size guard
         i = 0
         while i < len(args):
             a = args[i]
@@ -2706,6 +2715,12 @@ def dispatch_command(cmd: str) -> None:
                 push_user = args[i + 1]; i += 2
             elif a == "--password" and i + 1 < len(args):
                 push_password = args[i + 1]; i += 2
+            elif a == "--graph-name" and i + 1 < len(args):
+                push_graph_name = args[i + 1]; i += 2
+            elif a == "--prune":
+                push_prune = True; i += 1
+            elif a == "--allow-shrink":
+                push_allow_shrink = True; i += 1
             elif subcmd == "callflow-html" and not a.startswith("-") and not graph_path_explicit:
                 candidate = Path(a)
                 if candidate.name == "graph.json" or candidate.suffix.lower() == ".json":
@@ -2883,6 +2898,24 @@ def dispatch_command(cmd: str) -> None:
             print(f"graph.graphml written - open in Gephi, yEd, or any GraphML tool")
 
         elif subcmd == "neo4j":
+            # --graph-name/--prune/--allow-shrink only exist on the FalkorDB
+            # writer. Refuse rather than ignore: silently accepting --prune here
+            # would report a converged push that never deleted anything.
+            _falkor_only = [
+                name for name, given in (
+                    ("--graph-name", push_graph_name != "graphify"),
+                    ("--prune", push_prune),
+                    ("--allow-shrink", push_allow_shrink),
+                ) if given
+            ]
+            if _falkor_only:
+                print(
+                    f"error: {', '.join(_falkor_only)} "
+                    f"{'is' if len(_falkor_only) == 1 else 'are'} supported only by "
+                    f"`graphify export falkordb`.",
+                    file=sys.stderr,
+                )
+                sys.exit(1)
             if push_uri:
                 from graphify.export import push_to_neo4j as _push
                 if push_password is None:
@@ -2899,9 +2932,27 @@ def dispatch_command(cmd: str) -> None:
         elif subcmd == "falkordb":
             if push_uri:
                 from graphify.export import push_to_falkordb as _push
-                result = _push(G, uri=push_uri, user=push_user,
-                               password=push_password, communities=communities)
-                print(f"Pushed to FalkorDB: {result['nodes']} nodes, {result['edges']} edges")
+                try:
+                    result = _push(G, uri=push_uri, user=push_user,
+                                   password=push_password, communities=communities,
+                                   graph_name=push_graph_name, prune=push_prune,
+                                   allow_shrink=push_allow_shrink)
+                except ValueError as exc:  # prune size guard
+                    print(f"error: {exc}", file=sys.stderr)
+                    sys.exit(1)
+                _summary = f"{result['nodes']} nodes, {result['edges']} edges"
+                if push_prune:
+                    _summary += (
+                        f" (pruned {result['deleted']} nodes, "
+                        f"{result['deleted_edges']} edges)"
+                    )
+                print(f"Pushed to FalkorDB [{push_graph_name}]: {_summary}")
+                if not push_prune and result.get("target_surplus"):
+                    print(
+                        f"  note: target has {result['target_surplus']} node(s) the "
+                        f"source does not; --prune converges it.",
+                        file=sys.stderr,
+                    )
             else:
                 from graphify.export import to_cypher as _to_cypher
                 _to_cypher(G, str(out_dir / "cypher.txt"))
@@ -2934,6 +2985,7 @@ def dispatch_command(cmd: str) -> None:
             global_remove as _global_remove,
             global_list as _global_list,
             global_path as _global_path,
+            global_push as _global_push,
         )
         if subcmd == "add":
             # graphify global add <graph.json> [--as <tag>]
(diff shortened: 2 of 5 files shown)


Azeem1985:
Independent verification from the box that produced the #3057 numbers — same rig that reviewed the `falkordb-backend` version of this change ([here](https://github.com/galshubeli/graphify/pull/2#issuecomment-5407457062)).

**Setup:** clean clone of `24c45a0` (parent `282976b`, current `v8` tip), live `falkordb/falkordb:latest`, scratch instance, scratch keys, Python 3.14.

**Test suite:** 17/17 pass in ~1.3s (15 new + 2 pre-existing). Negative control — the same test file against the parent commit: **15 failed / 2 passed**, i.e. every one of the 15 new tests fails on the parent. That is one more than the PR's own "14 of 15"; same counting difference as on the fork review — a test that dies on a then-unknown kwarg is the test catching the missing API, so I count it as failing honestly.

**CLI placement**, since the fork thread reported a near-miss where the `global push` block first landed in the `provider` chain: what shipped is correct. `graphify provider push <URI>` refuses with `Usage: graphify provider [add|list|show|remove]`, and `global push` routes to the global chain.

**`graphify global push` end-to-end**, in a sandboxed `$HOME` against the scratch instance: added two toy repos, initial push delta-pushed both; an immediate re-push printed `up to date, 2 repo(s) unchanged` and sent 0 rows; after changing one repo, only that repo was re-sent (`3 nodes, 2 edges across 1 repo(s); 1 unchanged, 3 nodes pruned`). Read-back: exactly the source's 5 `:Entity` ids — the stale id gone, the renamed id present — plus the 2 per-repo `:GraphifyPushState` nodes, as designed.

**Neo4j refusal:** `export neo4j` with `--prune` → `error: --prune is supported only by 'graphify export falkordb'.`, and the same for `--graph-name`. Refused, not ignored, as the description says.

I did not re-run the 20-repo/20k-node perf corpus, so I can't independently confirm the 27x — but the shape of the claim is right: once a branch's own writes are batched, the exporter is the last per-row path standing, and v8 had never batched it.

This matches the contract as reviewed on the fork thread — `--prune` opt-in with the drift one-liner on the default path, the shrink guard refusing before deleting, push state in the target rather than a local ledger. From the reporter's side, this closes #3057.

Azeem1985:
Following up the verification run above — I also executed the two `Escalate · high` findings, since the review marks them "NOT verified (no proof, no reproducing execution)". Neither holds on head `24c45a0`:

**"Delta mode ignores repos present in manifest but absent from target"** — `_plan_delta` (`graphify/exporters/graphdb.py:338-345`) re-sends on either of the two ways a manifest repo can be absent from the target: no `:State` entry (`known is None` → reason "not present in target"), and node-count drift (`live != want_count` → reason "target drift"), which also covers the sneakier case where the state row survived but the nodes are gone (`live == 0`). The suite ships a test for exactly that: `test_delta_repairs_drift_a_ledger_would_miss`.

**"push_to_falkordb node labels changed from per-file-type to fixed :Entity, breaking existing graph queries"** — the writer keeps both labels: `MERGE (n:Entity {id: row.id}) SET n:{ftype}` (`graphdb.py:196-197`), so nodes come out as `:Entity:Python` etc., and `_ensure_schema` (`graphdb.py:166-177`) backfills `:Entity` onto nodes pushed by older versions so MERGE can't duplicate them. Old per-file-type queries and new `:Entity` queries both keep working. Shipped tests: `test_pushed_nodes_carry_the_entity_label`, `test_pushed_nodes_keep_their_file_type_label_too`.

Re-ran the four relevant tests just now against a scratch FalkorDB on this head (the two label tests, `test_delta_repairs_drift_a_ledger_would_miss`, `test_delta_resends_only_the_changed_repo`): **4 passed**.

The three `medium` items read as author's-choice questions (CLI password ergonomics, the same-millisecond epoch window, `:State` accounting for removed repos) — leaving those to you, but happy to run any of them through the same rig if useful.

galshubeli:
Thanks — and you are right about the count, so I have corrected the PR description rather than leave a number in it that I cannot reproduce.

**15 failed / 2 passed against the parent, confirmed here too.** The cause of my "14 of 15" is boring and entirely mine: I ran the negative control *before* adding the drift-notice test you asked for on the fork thread, and never re-ran it after. So the 15th test existed by the time I wrote the description but had never been measured against the parent. It fails there like the rest — `KeyError: 'deleted'` / no `target_surplus` key — which is the missing API being caught, exactly as you count it. Description now says all 15, and credits your reproduction.

That is the second time your counting has been right and mine stale on the same axis, which suggests my habit of measuring the negative control once and then continuing to edit the test file is the actual defect. Re-running it as the last step before writing the description is the fix.

On the perf corpus you did not re-run: entirely reasonable not to, and the claim does not need to carry the review. For the record, the two branches measured 4.1s -> 0.4s and 79.7s -> 2.9s for the same 20-repo / 20k-node push, and your reading of why is exactly right — the other branch had already batched its own writes, so its exporter was a smaller share of the total. Anyone who wants to check it needs only a scratch key and the loop in the description; nothing in the correctness argument rests on it.

Everything else you exercised — provider/global routing, the delta skip and selective re-send, the read-back showing the stale id gone and the `:GraphifyPushState` pair present, the Neo4j refusal — matches what I see. Nothing outstanding from my side.

Azeem1985:
Description edit confirmed — it now reads all 15 failing against the parent, with the reproduction credited. That settles the only number that was open.

Agreed on the habit being the real defect: a negative control measured once while the test file keeps moving is the same class of error as quoting a benchmark from an earlier commit, and re-running it as the last step before writing the description is the mechanism that makes it impossible rather than unlikely.

Thanks for putting the perf pair on the record — 79.7s → 2.9s here vs 4.1s → 0.4s on the branch that had already batched its own writes is exactly the split the "exporter was the last per-row path standing" reading predicts, so the 27x now carries both a reproducible loop and a coherent mechanism, even if nobody re-runs it.

Nothing outstanding from my side either — from the reporter's seat, #3057 closes when this merges.

PR #3070, opened 2026-08-25
fix(watch): ignore read-only inotify events so the watcher stops re-triggering on its own reads

Azeem1985:
Replaces #2969, which is the same fix but was opened from a branch that also carried unrelated work from our fork — 13 files and 25 commits, when the change itself is one function. That was my mistake in how I cut the branch, not a change in what I am proposing. This branch is cut from `v8` and carries only the fix. #2969 is being closed with a pointer here.

### The defect

On Linux, watchdog >= 2.3 emits `opened` and watchdog >= 4 emits `closed_no_write` for every file open and close. `Handler.on_any_event` counted both as file changes.

The watcher's own AST rebuild reads every watched file. So: rebuild reads the tree → inotify reports `opened`/`closed_no_write` for each file → the watcher counts them as changes → it rebuilds → it reads the tree again. A permanent loop that prints `N file(s) changed` while nothing was modified.

Hook guards that stat sources, editors, and agents reading files feed the same loop.

### Measured

On a ~1,700-file tree, over 3 days:

| | |
|---|---|
| CPU burned | ~26 minutes per hour |
| memory-kills | 15 in 3 days |
| `needs_update` flag rewritten | every 1-3 minutes |

After the fix the watcher is idle between real edits.

### The change

`_is_read_only_event(event)` returns True for exactly `opened` and `closed_no_write`; `on_any_event` returns early on those, next to its existing `is_directory` early return.

Creation, modification, move, deletion and **close-after-write** (`closed`) all still count, so no real change is missed. macOS uses `PollingObserver`, which never emits these event types, so the filter is a no-op there.

### Diff

3 files: `graphify/watch.py` (+20/-1), `tests/test_watch.py` (+29), `CHANGELOG.md` (+4). `tests/test_watch.py` passes 130/130.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 Full release notes with details on each version: [GitHub Releases](https://github.com/safishamsi/graphify/releases)
 
+## Unreleased
+
+- Fix: `graphify watch` no longer counts read-only inotify events (`opened`, emitted by watchdog >= 2.3, and `closed_no_write`, emitted by watchdog >= 4) as file changes on Linux, so the watcher's own AST rebuild — which reads every watched file — no longer re-triggers the watcher in a permanent loop ("N file(s) changed" with nothing modified; measured at ~26 CPU-minutes per hour and 15 memory-kills in 3 days on a ~1.7k-file tree, with the `needs_update` flag re-written every 1-3 minutes). Creation, modification, move, deletion and close-after-write still count; macOS (PollingObserver) never emits these events, so nothing changes there.
+
 ## 0.9.49 (2026-08-24)
 
 - Feature: `graphify merge-graphs` now links a type declaration that two repos share — same fully-qualified namespace and name, from different repos — with a `same_type_as` edge, so a shared contract type is navigable across the repo boundary; two unrelated types that merely share a short name are not linked (#3007, thanks @durmazoguzhan).
--- a/graphify/watch.py
+++ b/graphify/watch.py
@@ -1930,6 +1930,24 @@ def _batch_triggers_rebuild(batch: list[Path]) -> bool:
     return has_code or has_deletion
 
 
+_READ_ONLY_EVENT_TYPES = frozenset({"opened", "closed_no_write"})
+
+
+def _is_read_only_event(event) -> bool:
+    """True for watchdog events that mean a file was merely READ, not changed.
+
+    On Linux, inotify (watchdog >= 2.3) reports ``opened`` and, since watchdog 4,
+    ``closed_no_write`` for every file open/close — including the watcher's own
+    AST rebuild reading the tree, hook guards stat-ing sources, and editors or
+    agents reading files. Counting those as changes makes the watcher re-trigger
+    itself forever ("N file(s) changed" while nothing was modified), burn CPU and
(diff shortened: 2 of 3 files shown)


manumastro:
Independent reproduction, same root cause — confirming this from a separate codebase before you land it.

We hit this in a Docker dev container (Debian bookworm, graphifyy 0.9.49, watchdog 6.0.0) on an ~876-file repo: `graphify watch` in the background pinned one core at **94% CPU for 18+ minutes**, printing repeated batches of ~6200 changed files with nobody editing anything. Side effect was that our Flask and Vite dev servers looked like they were hanging on startup — they were just CPU-starved by the loop.

Confirmed it's the read events, not writes. A probe that only *read* a source directory, touching nothing:

```
event types generated by PURE READS: {'opened': 28, 'closed_no_write': 28}
```

And a probe placed on the handler during a single `graphify update` counted **1943 events passing all the existing filters, across 371 distinct files** (500 of them on one `package.json`) — every one of them produced by the rebuild's own reads.

Matches your analysis exactly, including why it's invisible on macOS: `PollingObserver` only synthesizes created/modified/deleted, so there is no read event to mistake for a change.

Our workaround (until this ships) drops the same two event types, just one level higher — in `FileSystemEventHandler.dispatch`, before `on_any_event` — then delegates to `graphify.watch.watch()` unchanged, so no logic is duplicated and `site-packages` isn't patched. Behaviourally it's the same filter as `_is_read_only_event`, and it holds up: 1 real edit → exactly 1 rebuild, **0 spontaneous rebuilds over 3 minutes** with both dev servers running, watcher CPU at idle (2 jiffies over 20s).

Keeping `closed` (IN_CLOSE_WRITE) unfiltered is the right call — that one does follow a real write, and filtering it would lose edits from editors that write-then-close.

+1 on merging. Happy to test a pre-release against our repo if that's useful.

Azeem1985:
Thanks — an independent reproduction from a second codebase is exactly what this needed, and yours covers the environment mine couldn't: a Docker dev container, watchdog 6.0.0, and a polyglot repo where the loop starves neighbouring dev servers rather than just burning a core. Your pure-read probe (`{'opened': 28, 'closed_no_write': 28}`) is the cleanest possible demonstration of the causal step — reads alone generate exactly the two event types this PR filters — and 1,943 filter-passing events from a single `graphify update` is the same self-feeding shape I measured (~26 CPU-minutes/hour on a ~1,700-file tree).

Your stopgap is sound: dropping the same two event types in `FileSystemEventHandler.dispatch` before `on_any_event` is behaviourally identical to `_is_read_only_event` — the same filter, one frame higher — so removing it once this ships should be a no-op. Agreed on `closed`, too: IN_CLOSE_WRITE follows a real write, so it stays unfiltered here.

Taking you up on the test offer — this branch installs directly with:

```
pip install --force-reinstall "git+https://github.com/TelB-io/graphify@upstream/watch-readonly-events"
```

If your 876-file container repo shows 1 edit → 1 rebuild and an idle watcher between edits on that build, a one-line note here would give the maintainers a third independent datapoint (bare metal, scratch rig, and your container).

safishamsi:

Azeem1985:
Thanks @safishamsi — confirmed the read-only-event filter is in the v0.9.50 tag, and the authorship-preserving cherry-pick is appreciated.

PR #3071, opened 2026-08-25
feat(llm): send an explicit reasoning_effort for Kimi models that support it (GRAPHIFY_KIMI_EFFORT)

Azeem1985:
Replaces #2982, which is the same six-line change but was opened from a branch that also carried unrelated work from our fork — 12 files and 21 commits. That was my mistake in how I cut the branch. This branch is cut from `v8` and carries only the change. #2982 is being closed with a pointer here.

### The gap

The `kimi` backend entry in `BACKENDS` has no `reasoning_effort`, so graphify sends none and the server's own default applies silently — `"high"` for K3. The `gemini` block directly above already carries an effort setting, so this reads as an omission rather than a deliberate choice.

Kimi K3 advertises `valid_efforts ["low","high","max"]` on `/models`, so the field is supported and settable.

### The change

The `kimi` entry gains `"reasoning_effort": os.environ.get("GRAPHIFY_KIMI_EFFORT", "max")`. It is forwarded by the existing request plumbing — no new call path — so models that ignore the field are unaffected.

`max` as the default is a judgement call: extraction quality is the point of the pass and the effort field does not change per-token pricing. If you would rather default to `high` (today's effective behaviour) and leave `max` opt-in, that is a one-word change and I am happy to make it.

### Diff

3 files: `graphify/llm.py` (+6), `tests/test_kimi_reasoning_effort.py` (+31), `CHANGELOG.md` (+4). Tests pass.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 Full release notes with details on each version: [GitHub Releases](https://github.com/safishamsi/graphify/releases)
 
+## Unreleased
+
+- Feature: the `kimi` backend now sends an explicit `reasoning_effort` (default `max`, overridable via `GRAPHIFY_KIMI_EFFORT`) for Kimi models that support it — K3 advertises `valid_efforts ["low","high","max"]` on `/models`, and sending nothing let the server default (`"high"` for K3) apply silently while the gemini backend already carried an effort setting. Forwarded by the existing request plumbing; no behaviour change for models that ignore the field.
+
 ## 0.9.49 (2026-08-24)
 
 - Feature: `graphify merge-graphs` now links a type declaration that two repos share — same fully-qualified namespace and name, from different repos — with a `same_type_as` edge, so a shared contract type is navigable across the repo boundary; two unrelated types that merely share a short name are not linked (#3007, thanks @durmazoguzhan).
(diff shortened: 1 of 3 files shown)


Azeem1985:
Two things I owe this PR before it is looked at — one a sibling report I should have linked, one a risk in my own change that I have not proved either way.

**1. The same gap was reported for a different backend, two days before this.**

- **#2932** (AutogenousEngine, 2026-08-21) — *"ollama backend never sets reasoning_effort, so local reasoning models (nemotron, deepseek-r1, …)"*
- **#2956** (amtulifra, 2026-08-22) — a fix for #2932, model-aware rather than backend-wide

Those are Ollama; this PR is Kimi. Not a duplicate, but the same shape twice in one week, which probably makes it a backend-table question rather than two patches. If you would rather solve it once, #2956's model-aware approach is the better base and I am happy for this to be closed in favour of it.

**2. A risk in my own change that I have not verified.**

This PR sets `reasoning_effort` for the **whole** `kimi` backend. The backend's `default_model` is `kimi-k2.6` (`llm.py`), and your own comment two lines below it records that this model family is strict about extra fields:

```python
"temperature": None,  # kimi-k2.6 enforces its own fixed temperature; sending any value raises 400
```

**I have not tested whether `kimi-k2.6` accepts `reasoning_effort`.** My measurements are all on `kimi-k3`, which advertises `[low, high, max]` on `/models`. If k2.6 rejects the field the same way it rejects `temperature`, this PR breaks the default Kimi path for everyone who has not overridden the model — which would be a straight regression, and worse than the gap it closes.

Stating it rather than leaving you to find it. Two ways to make it safe, and I will send whichever you prefer:

- gate the field on the model advertising it (the shape #2956 takes for Ollama), or
- send it only when `GRAPHIFY_KIMI_EFFORT` is explicitly set, so nothing changes for anyone who has not asked for it

Do not merge this as it stands until one of those is in, or until someone confirms k2.6 tolerates the field.

PR #3072, opened 2026-08-25
fix(postgres): carry tree-sitter-sql in the postgres extra and surface grammar errors

### LINKED ISSUE #3047: postgres extra lacks tree-sitter-sql: --postgres on a live DB silently reports 0 nodes, exit 0

Azeem1985:
Replaces #3048. Same fix, but that PR was opened from a branch carrying our whole fork — 62 files and 41 commits for a change that is two lines plus a test. Nobody should have to review that, and I am sorry for sending it in that shape. This branch is cut from `v8` and carries only the fix. #3048 is being closed with a pointer here.

Fixes #3047.

### The defect

`pip install "graphifyy[postgres]"` — the command the README gives — installs `psycopg` but not `tree-sitter-sql`.

`introspect_postgres()` reconstructs the catalog as DDL text and hands it to `extract_sql()`. Without the grammar, `extract_sql()` returns `{"error": ..., "nodes": [], "edges": []}`. `introspect_postgres()` returned that dict through unchanged, so the CLI printed:

```
PostgreSQL: 0 nodes, 0 edges
```

and exited **0** — indistinguishable from pointing at an empty database.

### Reproduced

`postgres:17-alpine`, graphifyy 0.9.48, installed exactly as documented. Database with 3 tables, foreign keys between them, one view and one function. Result: `0 nodes, 0 edges`, exit 0.

### The change

1. `pyproject.toml`: the `postgres` extra now lists `tree-sitter-sql` alongside `psycopg[binary]`, so the documented install works.
2. `graphify/pg_introspect.py`: a grammar error coming back from `extract_sql` raises `ImportError` with the install hint, instead of being reported as an empty database.

The `sql` and `all` extras already carried the grammar — only `postgres` was short, which is probably why this went unnoticed. The second half is there because even with the extra fixed, an environment that has the driver and not the grammar should say so rather than report zero.

### Diff

4 files: `pyproject.toml` (+5/-1), `graphify/pg_introspect.py` (+8), `tests/test_pg_introspect.py` (+20), `CHANGELOG.md` (+4). 9 tests pass.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 Full release notes with details on each version: [GitHub Releases](https://github.com/safishamsi/graphify/releases)
 
+## Unreleased
+
+- Fix: the `postgres` extra now carries `tree-sitter-sql`, and `pg_introspect` reports the missing grammar instead of returning an empty graph. Installing exactly as the README documents (`pip install "graphifyy[postgres]"`) and pointing graphify at a live database returned "0 nodes, 0 edges" and exited 0: the extra pulled the `psycopg` driver but not the SQL grammar the synthesized DDL is parsed with, so every table, view and function silently produced nothing. Reproduced on `postgres:17-alpine` with 0.9.48 against a database with 3 tables, foreign keys, a view and a function. A silent empty result is worse than a failure, so a missing grammar now raises with the install hint rather than being swallowed.
+
 ## 0.9.49 (2026-08-24)
 
 - Feature: `graphify merge-graphs` now links a type declaration that two repos share — same fully-qualified namespace and name, from different repos — with a `same_type_as` edge, so a shared contract type is navigable across the repo boundary; two unrelated types that merely share a short name are not linked (#3007, thanks @durmazoguzhan).
--- a/graphify/pg_introspect.py
+++ b/graphify/pg_introspect.py
@@ -154,4 +154,12 @@ def introspect_postgres(dsn: str | None = None) -> dict:
 
     # Pass virtual path and in-memory DDL content to extract_sql
     result = extract_sql(virtual_path, content=ddl_string)
+    # extract_sql() reports a missing grammar as an 'error' key with empty
+    # nodes/edges; surface it instead of letting the CLI print
+    # "PostgreSQL: 0 nodes" as if the database were empty.
+    if result.get("error"):
+        raise ImportError(
+            f"{result['error']} (required by --postgres; "
+            "install with: pip install 'graphifyy[postgres]')"
+        )
     return result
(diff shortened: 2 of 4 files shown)


Azeem1985:
Re "introspect_postgres now raises ImportError for any extract_sql error" (`pg_introspect.py:160`, marked "NOT verified"): the guard does key on any `error` value — but every value `extract_sql` can put there is an environment/grammar failure, so the `ImportError` classification holds for each reachable path.

The `error` key has exactly three writers in `graphify/extractors/sql.py`:

1. `tree_sitter_sql` not installed (the missing-grammar case this PR is about),
2. `tree_sitter_sql` installed but failed to load (e.g. a wrong-ABI C extension, the #2602 case),
3. the `except Exception` around `Language(...)`/`Parser(...)`/`parse(...)` — grammar/parser initialization.

Data-dependent failures cannot reach it: tree-sitter is error-tolerant, so malformed SQL yields ERROR nodes in the tree rather than raising, and on this path the DDL arrives in-memory (`content=ddl_string`), so the `path.read_bytes()` branch is never taken. The raise also embeds the original error string first, so cases 2 and 3 surface their real cause rather than being masked by the install hint.

The one cosmetic wrinkle is that a case-3 failure would carry the `pip install 'graphifyy[postgres]'` suffix even though installing would not help. If you'd rather the hint decorate only the two import branches and case 3 re-raise as a plain `RuntimeError`, say the word and I'll push that tightening.

safishamsi:

Azeem1985:
Thanks — confirmed both halves are in the v0.9.50 tag (the extra carries the grammar, and grammar errors now surface as `ImportError`). Appreciate the credit-preserving cherry-pick.

PR #2981, opened 2026-08-23
feat(llm): openai-cli backend — semantic extraction through the locally authenticated Codex CLI

Azeem1985:
## What
A sibling of the existing `claude-cli` backend for operators whose OpenAI access is a ChatGPT subscription (Codex CLI OAuth), not an API key: `graphify extract . --backend openai-cli`.

Mechanics (field-proven implementation, measured ~308 docs/hour at concurrency 6 on an 8 GB box): `codex exec --skip-git-repo-check --json --sandbox read-only`, prompt via stdin (a real chunk exceeds Linux MAX_ARG_STRLEN as an argv entry), result via `-o` file, empty-graph refusal so a hollow reply cannot trigger a bisect storm. Two per-call config pins via codex's own `-c` mechanism: `mcp_servers={}` (each exec otherwise spawns the user's configured MCP servers per call — measured 160–300 MB each, more than the extraction itself) and `model_reasoning_effort` (`GRAPHIFY_OPENAI_CLI_EFFORT`). Model via `GRAPHIFY_OPENAI_CLI_MODEL`. The extract credential gate treats a present `codex` binary as the credential, mirroring claude-cli.

## Tests
`tests/test_openai_cli_backend.py` (argv contract, env overrides, missing-binary error); full suite 4,812 passed.

## Live verification
One-document extraction on a real box: 9 nodes / 10 edges landed, 98,519 in / 4,733 out tokens, $0 (subscription).

Issue #798, opened 2026-05-09
[Critical] Context Window Saturation / Missing Session Reset between Chunks (Ollama Backend)


RANGERVIII:
Hey @safishamsi,

Quick follow-up after testing **v0.7.12** on a high-end local setup (RTX 5090, Gemma 4:31b).

We’ve identified a major bottleneck in how the Ollama backend handles multiple chunks. After about 3-4 chunks, the model starts returning the "very few tokens" warning and essentially stops producing valid JSON.

**The Diagnosis:** It appears that `graphify` is not explicitly resetting the Ollama session/context between chunks.

**What's happening:**
In local inference (especially via Ollama’s `/api/generate` or `/api/chat`), if the `context` parameter (the KV cache) isn't cleared or handled statelessly for each new chunk, the model "saturates." It carries the noise and tokens from Chunk 1 and 2 into the context window for Chunk 3. Even with a large 5090 VRAM, the **attention mechanism** gets overwhelmed by the uncleared history, leading to the "0-token" or "closing-bracket-only" responses we are seeing.

**The Evidence:**

* Chunk 1-2: Fast, perfect inference.
* Chunk 3-4: Sudden drop in token count, "few tokens" warning, but the GPU is still under load (processing the saturated context).
* Restarting the script: Fixes it immediately for the first few chunks again.

**The Fix:**
Each new chunk or file extraction call to the Ollama API should be **stateless**. Please ensure that for every new extraction call, the `context` array is either empty or the session is explicitly reset. A simple flag or a fresh request object per chunk would solve this and prevent the "context-leak" that currently ruins long extraction runs.

Looking forward to v0.7.13! 🚀

RANGERVIII:
**Headline: Why 128k Context is a "Band-Aid" and not a Fix (Technical Audit)**

Hey @safishamsi,

First off, huge respect for the response time! Seeing a commit pushed minutes after a bug report shows you really care about the product.

However, we need to talk about commit `8c00287`. While increasing `num_ctx` to 128k buys some time, it’s a bit like buying a bigger gas tank for a car with a massive fuel leak. It doesn't fix the engine; it just delays the inevitable stall.

[...]

PR #3073, opened 2026-08-25
feat(llm): openai-cli backend — semantic extraction through the locally authenticated Codex CLI

Azeem1985:
Replaces #2981, which is the same feature but was opened from a branch that also carried unrelated work from our fork — 11 files and 17 commits. That was my mistake in how I cut the branch. This branch is cut from `v8` and carries only this feature. #2981 is being closed with a pointer here.

### What it adds

`--backend openai-cli` runs semantic extraction and community labeling through `codex exec`, the locally authenticated OpenAI Codex CLI, so the work rides a ChatGPT OAuth subscription instead of a metered `OPENAI_API_KEY`.

It is deliberately the same shape as the existing `claude-cli` backend: zero pricing (subscription usage is not metered API spend), forced-serial calls unless opted out, and a credential gate that accepts a present CLI in place of an API key.

`GRAPHIFY_OPENAI_CLI_MODEL` (default `gpt-5.6-sol`) and `GRAPHIFY_OPENAI_CLI_EFFORT` (default `ultra`) configure it. `GRAPHIFY_OPENAI_CLI_PARALLEL=1` opts out of serial.

### Four things that took measurement to get right

These are in the code with comments, but worth stating because a reviewer would reasonably wonder why the simpler version is not there:

1. **The prompt goes over stdin, not argv.** Linux caps a single argv entry at 128 KB (`MAX_ARG_STRLEN`). Real extraction chunks reach 240-306 KB, so passing the prompt as an argument fails with `E2BIG` on ordinary input, not on edge cases.
2. **The sandbox is read-only.** An agentic CLI given write access to the corpus it is reading can modify what it is extracting from.
3. **MCP servers are disabled per server**, via Codex's own `mcp_servers.<name>.enabled=false` override. A blanket `mcp_servers={}` does not work: Codex deep-merges its config, so the empty key merges away and every configured server still launches. Measured cost of the naive version on our box: four servers of ~152 MB each, spawned on every `codex exec` call.
4. **An empty-but-valid graph raises** instead of being returned. A hollow response otherwise feeds the chunk-bisect retry path, which subdivides and re-sends a response that is never going to improve — that burned 87% of an hourly quota once before we caught it.

Token usage is read from the last `turn.completed` JSONL event, so cached input is not double-counted.

### Diff

5 files: `graphify/llm.py` (+286/-1), `tests/test_openai_cli_backend.py` (+219), `graphify/cli.py` (+11), `graphify/__main__.py` (+2/-2, help text), `CHANGELOG.md` (+4). 15 tests pass.

### One honest caveat

This has been running in our fork against real extraction loads, but only against `gpt-5.6-sol` through one person's Codex install. If you want it behind a documented-but-unadvertised flag until more people have exercised it, that is a reasonable call.


Azeem1985:
Re the two `Escalate · high` findings ("Incomplete conditional leaves module syntactically invalid" / "Incomplete openai-cli concurrency branch makes module unloadable", both anchored at `graphify/llm.py:2827`): neither reproduces at this PR's head.

Checked on a clean shallow clone of `4e1cfd6` (the PR's only commit), Python 3.14.4:

- `ast.parse()` over `graphify/llm.py` — parses
- `python -m py_compile graphify/llm.py` — byte-compiles
- `import graphify.llm` — imports; `_call_openai_cli` is present on the module

The flagged line is a complete two-line guard, third in the existing run of serial-by-default backend guards:

```python
if backend == "openai-cli" and os.environ.get("GRAPHIFY_OPENAI_CLI_PARALLEL", "").strip() != "1":
    max_concurrency = 1
```

exactly mirroring the `ollama` (#798) and `claude-cli` guards directly above it. Nothing is dangling. The review itself marks both findings "NOT verified (no proof, no reproducing execution)" — the verification above is that missing execution, and it comes back clean.

On the two inline coupling notes: `_call_openai_cli()`'s 9-callee fan-out is the same shape as the `_call_claude_cli()` path it deliberately mirrors, and `_call_claude_cli()` itself predates this PR — the diff adds no `def _call_claude_cli`, so its coupling delta isn't from this change.

Azeem1985:
Re the five `Worth a look` findings on the re-review (all marked "NOT verified — no proof, no reproducing execution"): I ran the missing executions on head `4e1cfd6`. Two don't reproduce; three are accurate observations of deliberate, documented behavior that this backend shares with the shipped `claude-cli` backend it mirrors.

**"extract_corpus_parallel truncated openai-cli concurrency guard likely broken" (`llm.py:2827`)** — doesn't reproduce. Beyond the previous round's static checks (ast.parse / py_compile / import, all clean on this same commit), here is the execution: monkeypatch `_extract_with_adaptive_retry` with an in-flight counter, run `extract_corpus_parallel` over 8 single-file chunks with `max_concurrency=4`:

```
openai-cli, no env, max_concurrency=4        -> peak in-flight: 1
openai-cli, OPENAI_CLI_PARALLEL=1, max=4     -> peak in-flight: 4
```

The guard serializes by default and honors the opt-in, exactly as documented.

**"openai-cli concurrency not forced to 1 in code despite docs" (`__main__.py:547`)** — the flagged line is help text for `label`/`cluster-only`, and it is backed by code: the labeling path has its own guard inside `label_communities` (`llm.py:3626-3627`), identical to the extraction guard at `llm.py:2827-2828`. Both paths force serial unless `GRAPHIFY_OPENAI_CLI_PARALLEL=1`.

**"MCP server disabling fails open" (`llm.py`)** — accurate, and it is the documented contract rather than an oversight (`_codex_disable_mcp_args` docstring: "Best effort: if that call is unavailable (older Codex, no config), extraction proceeds with whatever the user configured"). Two reasons it is shaped this way: (1) the server list must come from `codex mcp list --json`, because Codex deep-merges config overrides — a blanket `-c mcp_servers={}` disables nothing — and a Codex old enough to lack that subcommand offers no way to enumerate servers at all; (2) the disabling is a resource measure (each configured server is otherwise spawned per `codex exec` call; measured ~152 MB per server), not the security boundary. The security boundary is `--sandbox read-only`, which is unconditional on both call sites.

**"Untrusted corpus text is sent to an agentic CLI that can run tools"** — true, and inherent to the CLI-backend category: it is the same threat model as the shipped `claude-cli` backend. The mitigations are in the invocation itself: `--sandbox read-only` on both the extraction and the labeling call sites (an injection cannot write to the corpus or anything else), every configured MCP server disabled per call, prompt delivered over stdin. If maintainers want CLI backends gated harder — say, an explicit opt-in flag acknowledging agentic execution — that policy would apply to `claude-cli` equally, and I'm happy to implement it for both.

**"openai-cli runs an LLM agent with inherited cwd and environment" (`llm.py:3167`)** — parity, not a regression: no `subprocess.run` call in `llm.py` sets `cwd=` or `env=`, including the pre-existing `claude-cli` backend. Environment inheritance is load-bearing (the CLI locates its local auth through it), and pinning `cwd` buys nothing under `--sandbox read-only`, since read access is not cwd-scoped there.

PR #3074, opened 2026-08-25
perf(detect): stop building pathlib objects per pattern per file in ignore evaluation (#2226)

### LINKED ISSUE #2226: detect(): _is_ignored is O(paths x patterns) with exception-driven relative_to — ~100s of a 131s incremental run for a 2-file change
## Summary

On an incremental rebuild with **2 changed files**, `detect()` spends ~100s re-evaluating ignore patterns. The cost is not AST extraction — the incremental path correctly extracts only the changed files. It is `_is_ignored`/`_eval` walking every ignore pattern per candidate path, with `Path.relative_to` used as an exception-driven containment test.

The practical effect is an inversion: an incremental refresh (72.9s wall) is **slower than a full `--clean` rebuild** (36.9s wall) on the same repo.

## Environment

- graphify 0.9.26 (`detect.py` is byte-identical in 0.9.27, so this is unchanged there)
- Python 3.12.3, Linux aarch64
- Repo: ~2,300 git-tracked files, 595 indexed, 13.2k nodes / 32.7k edges
- 338 ignore patterns total (208-line root `.gitignore` + 49-line `.graphifyignore` + nested)

## Profile

`cProfile` over a `_rebuild_code(changed_paths=[2 files], no_cluster=True)` run (131s under the profiler; 72.9s unprofiled):

```
   ncalls  tottime  cumtime  filename:lineno(function)
        2    0.051   104.358  graphify/detect.py:1168(detect)
     4768    0.032   103.410  graphify/detect.py:1053(_is_ignored)
    17970    2.366   102.666  graphify/detect.py:1076(_eval)
   790378    4.179    89.867  pathlib.py:663(relative_to)
   790378    1.341    49.659  pathlib.py:690(is_relative_to)
```

790,378 `relative_to` calls for a 2-file change set.

## Root cause

In `_eval` (`detect.py:1076-1115`), every candidate path is tested against **every** pattern, and each test resolves the path against that pattern's anchor:

```python
for anchor, pattern in patterns:
    ...
    try:
        rel_anchor = str(target.relative_to(anchor)).replace(os.sep, "/")
    except ValueError:
        continue  # target outside this pattern's anchor: cannot match
```

Two compounding costs:

1. **O(paths × patterns)** — no grouping by anchor, so a path is re-resolved against each of the 338 patterns.
2. **Exceptions as control flow** — for a path outside a pattern's anchor (the common case with nested ignore files), the containment test raises and catches `ValueError`. `Path.relative_to` also allocates parsed-path objects; the profile shows `pathlib.PurePath.__init__` called 8.5M times.

## Suggested fix

Both are local to `_eval` and preserve last-match-wins semantics:

- **Group patterns by anchor.** Resolve `target` against each *distinct* anchor once (338 patterns typically share a handful of anchors), then match the resulting relative path against that anchor's patterns.
- **Replace `relative_to` + `except ValueError` with a string prefix test** on the already-normalized posix strings, or use `is_relative_to` without constructing the relative path when only containment is needed.

A per-scan memo keyed on `(anchor, target)` would also help, though grouping should make it unnecessary.

I'm happy to send a PR if the approach sounds right.

Azeem1985:
Implements #2226. The diagnosis and the suggested fix in that issue are @Agrias-00's, and this PR does what they proposed — group by anchor, resolve once per distinct anchor, drop the exception-driven `relative_to` containment test. Credit is theirs; I am contributing the implementation because I hit the same wall independently and had already written it.

**@Agrias-00 — you offered to send a PR for this. If you would rather send your own, say so and I will close this immediately. No claim on it.**

Fixes #2226. Likely also covers #1958 (`detect()` CPU-bound 35+ min on a large monorepo despite dir pruning), though I have not reproduced that reporter's corpus.

### How I hit it

A 76,579-file vault whose 28,968-file `graphify-notes/` directory sits in `.graphifyignore`. `detect()` pinned one core at 99% for **52+ minutes**, and py-spy showed it inside `pathlib._format_parsed_parts` — the same stack #2226's cProfile points at.

### What the matcher was doing

Per #2226, plus three costs that profile did not separate out:

1. `Path.relative_to(anchor)` — and often `relative_to(root)` too — for **every pattern on every walked entry**, with `ValueError` as the not-contained signal.
2. Each pattern's flags (negation, directory-only, anchored) re-parsed **per entry**.
3. The relative path re-split into segments **per pattern**, and `is_dir()` re-`stat`ed **per matched directory-only pattern**.

On Python 3.12+ `relative_to` also walks `parents`, which is quadratic in path depth — so the cost grows with how deep the tree is, not just how wide.

### The change

`_lexical_relative(target, target_parts, anchor)` computes the relative path once per entry per **distinct** anchor by comparing already-split parts — exactly `relative_to`'s comparison, Windows casefold included. Pattern flags are parsed once and cached; segment splits and the `is_dir()` stat happen at most once per entry.

Directory-level pruning (one evaluation excluding a whole ignored subtree from the walk) is now pinned by regression tests that count walk descents and total evaluations, so a future change cannot silently turn pruning off — that was how the 28,968-file subtree got walked in the first place.

### Measured, on this diff against `v8` @282976b

Corpus: 4,202-file deep tree, 351 ignore patterns (300 `.graphifyignore` + 50 `.gitignore`), one 3,000-file ignored subtree.

| | wall | detect() output | SHA-256 of output |
|---|---:|---:|---|
| v8 @282976b | **73.46s** | 1,210 items | `90fc89c54ad495c4` |
| this branch | **9.86s** | 1,210 items | `90fc89c54ad495c4` |

**7.5x faster, byte-identical result.**

Directly comparable to #2226's `790,378 relative_to calls` — on a 402-file corpus with the same 351 patterns:

| | `relative_to` calls |
|---|---:|
| v8 @282976b | **382,674** |
| this branch | **1,488** |

257x fewer. (The residual 1,488 is one per entry, from code paths outside the matcher.)

### Correctness

The risk here is entirely "did matching change?", so that is what the tests attack:

- **38,000-case differential fuzz** of old matcher vs new, on Python 3.12 **and** 3.14 — zero mismatches. Anchored and unanchored patterns, negations, directory-only rules, nested ignore files, paths outside every anchor, unicode and casefold pairs.
- Byte-identical `detect()` output on the benchmark corpus above.
- `tests/test_detect.py`: **258 passed**, including the new walk-descent and evaluation-count assertions.

### Diff

3 files: `graphify/detect.py` (+164/-25), `tests/test_detect.py` (+188), `CHANGELOG.md` (+4). No new flags, no configuration, no behaviour change.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 Full release notes with details on each version: [GitHub Releases](https://github.com/safishamsi/graphify/releases)
 
+## Unreleased
+
+- Fix: the scan's ignore evaluation no longer builds pathlib objects per pattern per file. `_is_ignored`'s matcher called `Path.relative_to(anchor)` (and often `relative_to(root)`) for EVERY pattern on EVERY walked entry, re-parsed each pattern's flags per entry, re-split the relative path per pattern, and re-`stat`ed `is_dir()` per matched directory-only pattern; on Python 3.12+ `relative_to` also walks `parents` quadratically in path depth, so a real 76,579-file vault whose 28,968-file `graphify-notes/` sits in `.graphifyignore` had `detect()` pinned at 99% CPU for 52+ minutes inside `pathlib._format_parsed_parts` (caught live with py-spy). The relative path is now computed once per entry per distinct anchor by a lexical parts-prefix helper (`_lexical_relative`, exactly `relative_to`'s comparison, Windows casefold included), pattern flags are parsed once and cached, segment/prefix splits and the `is_dir` stat happen at most once per entry, and directory-level pruning (one evaluation excludes a whole ignored subtree from the walk) is now pinned by regression tests counting walk descents and evaluation totals. Matching semantics are unchanged — verified by a 38,000-case differential fuzz of old vs new on Python 3.12 and 3.14, zero mismatches. Synthetic 300-pattern/deep-tree scan: 12.0s → 3.3s; per-pattern `relative_to` calls on the walk drop from ~65,700 to ~2,000 (one per entry) on a 1,000-file corpus with 61 patterns (#2226).
+
 ## 0.9.49 (2026-08-24)
 
 - Feature: `graphify merge-graphs` now links a type declaration that two repos share — same fully-qualified namespace and name, from different repos — with a `same_type_as` edge, so a shared contract type is navigable across the repo boundary; two unrelated types that merely share a short name are not linked (#3007, thanks @durmazoguzhan).
--- a/graphify/detect.py
+++ b/graphify/detect.py
@@ -1249,6 +1249,73 @@ def _load_graphifyignore(root: Path, *, gitignore: bool = True) -> list[tuple[Pa
     return patterns
 
 
+# Parsed-pattern cache: raw pattern string -> (negated, directory_only,
+# path_relative, stripped_pattern). Ignore patterns are re-evaluated for every
+# walked entry; parsing the same strings per entry per scan was pure waste.
+# Plain dict (no LRU): the universe of keys is the distinct pattern lines in
+# the corpus's ignore files, which is small and bounded per process.
+_PARSED_PATTERN_CACHE: dict[str, tuple[bool, bool, bool, str]] = {}
+
+
+def _parse_ignore_pattern(pattern: str) -> tuple[bool, bool, bool, str]:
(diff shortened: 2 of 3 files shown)


safishamsi:

Azeem1985:
Thanks — confirmed the cached-pattern ignore path landed in v0.9.50. Appreciate the cherry-pick keeping authorship.

PR #3075, opened 2026-08-25
feat(config): let a project declare what an ambiguous extension means via .graphifyrc (#2961)

### LINKED ISSUE #1042: Add possibilty to configure code extensions and corresponding parser
The code extensions are hardcoded (the code is [here](https://github.com/safishamsi/graphify/blob/d1d5751fa4e3f687b7945bc3a641e9bb517e2f2e/graphify/detect.py#L28)).

For my concrete case, I'd like to add `.svh` as an extension to be parsed with the Verilog parser.

This would also mean modifying the code in [extract.py](https://github.com/safishamsi/graphify/blob/d1d5751fa4e3f687b7945bc3a641e9bb517e2f2e/graphify/extract.py#L8247).

I know how I can fix this for my case, to just add support for `.svh` files (and can make a PR for it), I wouldn't know how to generalize it.

### LINKED ISSUE #2961: `.inc` is hard-mapped to the Pascal extractor: PHP include files silently yield 20 nodes instead of 755 — needs a project-level extension→language config surface
### Summary

`.inc` is hard-mapped to the Pascal extractor (`extract.py`, `".inc": extract_pascal`). `.inc` is not a standardized extension — it means "include file" for whatever language a project happens to use. On pfSense/pfBlockerNG, `.inc` files are PHP, and every one of them is silently parsed as Pascal. The result is not an error or a warning: extraction succeeds and returns a handful of incidental nodes, so the graph looks populated while the entire shipped runtime is missing from it.

The generic fix is a **project-level configuration surface** that lets a repository declare what its ambiguous extensions actually are, rather than relying on a global hardcoded map. Serena and similar tools expose exactly this. #1042 asked for the same thing generically and was closed by hardcoding one extension (`.svh` → Verilog), which does not generalize: `.inc` cannot have a single correct global mapping, because different projects legitimately use it for PHP, Pascal, SQL, and assembly.

### Environment

- graphify (PyPI `graphifyy`) 0.9.48, Python 3.14, macOS
- Repository: pfSense package with 12 PHP `.inc` files under `src/` and `stubs/`

### Reproduce

Same bytes, two extensions:

```python
from pathlib import Path

src = Path("src/usr/local/pkg/pfblockerng/pfblockerng.inc")   # PHP, 460 declared functions
Path("a.inc").write_bytes(src.read_bytes())
Path("b.php").write_bytes(src.read_bytes())
Path("small.inc").write_bytes(src.read_bytes()[:60000])
Path("small.php").write_bytes(src.read_bytes()[:60000])

for name in ("a.inc", "b.php", "small.inc", "small.php"):
    r = extract([Path(name)], cache_root=Path("."))
    print(f'{name:12s} nodes={len(r["nodes"]):4d} edges={len(r["edges"]):4d}')
```

**Observed:**

```
a.inc        nodes=   7 edges=   6
b.php        nodes= 471 edges=1008
small.inc    nodes=   2 edges=   1
small.php    nodes=  48 edges=  65
```

The truncated pair rules out any size or truncation effect — the only variable is the extension.

**Expected:** a PHP `.inc` file yields the same graph as the identical `.php` file, or at minimum the project can declare that mapping.

### Impact on a real repository

Across the 12 `.inc` files in this repository:

| dispatch | nodes |
| --- | --- |
| Pascal (current) | 20 |
| PHP (correct) | 755 |

Re-extracting them through `extract_php` and merging took the graph's shipped-source coverage from 564 to 1285 nodes — it more than doubled. Twelve production subsystems that had no representation in the graph at all appeared as coherent communities: the feed download pipeline, tick scheduling and the due-ledger, filter/logging/state removal, the config gateway runtime, software update and pkg operations, log trimming, the apply/sync path, and Unbound reload.

Before that, `pfblockerng.inc` — 19,046 lines, 460 declared functions — contributed 7 nodes. A doc-linking pass over the same repository had to discard references to `sync_package_pfblockerng`, `pfb_live_punch_run`, `pfb_download`, `pfb_pfctl_table_op` and others as orphans, because the symbols did not exist in the graph despite existing in the source.

The failure is silent, which is the part that makes it costly: nothing in the output distinguishes "this file has 7 symbols" from "this file was parsed with the wrong grammar."

### Suggested resolution

**Primary — a project-level config surface.** Let a repository declare extension-to-language mappings, e.g. in `.graphifyrc` / `pyproject.toml` / a `graphify.toml`:

```toml
[extensions]
".inc" = "php"
```

This is the only approach that is correct for genuinely ambiguous extensions, and it generalizes the request in #1042 instead of requiring an upstream release per project. It would also cover `.m` (MATLAB vs ObjC), `.pl` (Perl vs Prolog), `.v` (Verilog vs Coq) and similar cases already visible in the tracker.

**Secondary — content sniffing as a reliable default.** For `.inc` specifically, a leading `<?php` (or an early `<?php` within the first few KB) is a dependable signal, and there is precedent for cheap header inspection. Sniffing would fix the common case without configuration; the config surface would still be needed for files that no heuristic can disambiguate.

**Minimum — do not fail silently.** If an extractor produces near-zero nodes for a large file, or if content plainly contradicts the mapped grammar, say so. A one-line warning would have surfaced this immediately instead of after a full graph build.

I am happy to open a PR for the sniffing path, the config path, or both, if you can indicate which shape you would accept.

abhay-codes07:
Closes #2961.

## The problem

`.inc` is hard-mapped to the Pascal extractor, but it is "include file" in whatever language a project happens to use: PHP on pfSense, Pascal in a Delphi tree, SQL or assembly elsewhere. A PHP `.inc` parsed as Pascal does not fail — it yields a handful of incidental nodes, so the graph looks populated while the shipped runtime is missing from it. On the reporter's file: 7 nodes instead of 471, from identical bytes.

Hardcoding one more extension (what closed #1042) cannot fix this, because `.inc` has no single correct global meaning. The same is true of `.h` (C or C++) and `.m` (Objective-C or MATLAB), which today get a content sniff that is right most of the time.

## The change

`.graphifyrc` — which already exists for `viz_node_limit` — gains a project-level declaration:

```
# .graphifyrc
language.inc=php        # a language name ...
language.tpl=.ts        # ... or an extension graphify already knows
```

The declaration reaches every place graphify keys a decision on the suffix, not just the dispatch table:

| site | effect |
|---|---|
| `detect.classify_file` | a declared extension counts as code (so `.tpl` → PHP is scanned at all) |
| `extract._get_extractor` | dispatch, ahead of the `.h`/`.m` content sniffs |
| `_lang_is_case_insensitive`, `_lang_family`, the PHP type pass | cross-file resolution rules follow the declared language |
| `resolver_registry.run_language_resolvers` | the declared language's resolvers wake (PHP's, not Pascal's) |
| `cache.load_cached` / `save_cached` | a new `salt` kwarg folds the target language into the key — the same bytes cached under the old extractor are never served for the new one, in either direction |
| the extraction pool | an `initializer` forwards the mapping to workers, which start with empty module state under `spawn` |

`detect()` and `extract()` activate `<root>/.graphifyrc` themselves, so the CLI, `update`/`watch`, the git hooks, the MCP server and the skill runbook all pick it up with no further wiring — including the issue's own library-level repro. A malformed file is reported once on stderr and the scan continues with graphify's defaults; a typo must be loud, but it must not kill a scan.

The parser moves from `hooks.py` into a dependency-free `graphify/rcfile.py` so `detect`/`extract` (and the workers) can import it; `hooks._load_graphifyrc` keeps its name, behaviour and error messages, and the existing hooks tests pass unchanged.

## Before / after

Same PHP source as `a.inc` and `b.php`:

```
no .graphifyrc          a.inc  nodes=  1 edges=  0
                        b.php  nodes=  7 edges= 10
language.inc=php        a.inc  nodes=  7 edges= 10
                        b.php  nodes=  7 edges= 10
language.inc=pascal     a.inc  nodes=  1 edges=  0    <- not the cached PHP entry
language.inc=klingon    a.inc  nodes=  1 edges=  0    + one stderr warning naming line 1
```

Through the real CLI:

```
  language overrides (.graphifyrc): .inc -> .php
[graphify extract] wrote graphify-out/graph.json — 7 nodes, 10 edges
```

## Tests

`tests/test_language_overrides.py` — 42 tests: the parser (aliases, explicit extensions, key normalisation, line-numbered errors, the existing option and unknown keys untouched, `hooks._load_graphifyrc` compatibility); each suffix-keyed site individually; the reporter's same-bytes repro; the library-caller path; `detect()` counting a declared extension as code; the cache round-trip in both directions; the pool initializer and the mapping it is handed; the once-only warning for a malformed file.

With the wiring reverted and only the new module kept, 12 of them fail; with it, all 42 pass. The related existing suites (`test_cache`, `test_detect`, `test_extract*`, `test_hooks`, `test_incremental*`, `test_pascal*`, `test_php_type_resolution`, `test_extractors_registry`) are unchanged; the full suite matches the `v8` baseline (the same platform-specific failures on Windows, nothing new).

README gains a short "Project configuration" section.

### DIFF
--- a/README.md
+++ b/README.md
@@ -425,6 +425,22 @@ dist/
 !src/**
 ```
 
+
+## Project configuration
+
+Create a `.graphifyrc` in the scan root for settings that belong to the project rather than to one run. One `key=value` per line, `#` comments.
+
+**Ambiguous extensions.** Some extensions mean different languages in different projects: `.inc` is PHP on pfSense, Pascal in a Delphi tree, SQL or assembly elsewhere; `.h` is C or C++; `.m` is Objective-C or MATLAB. graphify has to pick one global default, and when it picks wrong the file does not fail — it parses as the wrong language and yields a handful of incidental nodes, so the graph looks populated while the real symbols are missing. Declare what the extension means in your repo:
+
+```
+# .graphifyrc
+language.inc=php        # a language name ...
+language.tpl=.ts        # ... or an extension graphify already knows
+viz_node_limit=0        # baked into the git hooks (see Team setup)
+```
+
+The declaration applies everywhere graphify keys a decision on the extension — file classification, extractor dispatch, cross-file resolution, and the AST cache (a file re-parsed under a different language never reuses the old entry). It is read by `graphify extract`, `graphify update`, `watch`, the hooks, the MCP server, and the `/graphify` skill alike. A typo is reported once on stderr and the scan continues with graphify's defaults.
+
 ---
 
 ## Team setup
--- a/graphify/cache.py
+++ b/graphify/cache.py
@@ -961,11 +961,19 @@ def cache_dir(root: Path = Path("."), kind: str = "ast",
     return d
 
 
+def _salted_key(h: str, salt: str | None) -> str:
+    """Derive the on-disk key from a content hash plus optional salt."""
+    if not salt:
+        return h
+    return hashlib.sha256(f"{h}:{salt}".encode("utf-8")).hexdigest()
+
+
 def load_cached(path: Path, root: Path = Path("."), kind: str = "ast",
                 cache_root: Path | None = None, prompt: "str | Path | None" = None,
                 prompt_file: "str | Path | None" = None,
                 allow_legacy: bool = True,
-                allow_partial: bool = False) -> dict | None:
+                allow_partial: bool = False,
+                salt: str | None = None) -> dict | None:
     """Return cached extraction for this file if hash matches, else None.
 
     Cache key: SHA256 of file contents.
@@ -994,6 +1002,10 @@ def load_cached(path: Path, root: Path = Path("."), kind: str = "ast",
     so :func:`check_semantic_cache` can report N to the user. Callers that must
     not mix vintages within one entry (see :func:`save_semantic_cache`'s
     ``merge_existing``) pass allow_legacy=False.
+    ``salt`` folds extra material into the key for an entry whose validity
+    depends on more than the file's bytes — a file the project remaps to
+    another language (#2961) parses to a different graph from the same
+    content, so it must not hit the entry produced under the old extractor.
     Returns None if no cache entry or file has changed.
     """
     global _legacy_semantic_hits, _corrupt_cache_entries
@@ -1002,6 +1014,7 @@ def load_cached(path: Path, root: Path = Path("."), kind: str = "ast",
         h = file_hash(path, root, cache_root=cache_root)
     except OSError:
         return None
+    h = _salted_key(h, salt)
     prompt_fp = _resolve_prompt_fp(prompt, prompt_file)
     entry = cache_dir(location, kind, prompt_fp) / f"{h}.json"
     legacy_hit = False
@@ -1069,8 +1082,12 @@ def load_cached(path: Path, root: Path = Path("."), kind: str = "ast",
 
 def save_cached(path: Path, result: dict, root: Path = Path("."), kind: str = "ast",
                 cache_root: Path | None = None, prompt: "str | Path | None" = None,
-                prompt_file: "str | Path | None" = None) -> None:
+                prompt_file: "str | Path | None" = None,
+                salt: str | None = None) -> None:
     """Save extraction result for this file.
+
+    ``salt`` must match what the corresponding :func:`load_cached` passes
+    (see there); it namespaces the key, not the directory.
 
     Stores as graphify-out/cache/{kind}/{hash}.json where hash = SHA256 of current file contents.
     result should be a dict with 'nodes' and 'edges' lists.
@@ -1118,7 +1135,7 @@ def save_cached(path: Path, result: dict, root: Path = Path("."), kind: str = "a
         # the entry replays portably under any root (#2257). Strictly after the
         # source_file pass, which owns that field's bare-relative format.
         _relativize_ids_in(on_disk, p, root)
-    h = file_hash(p, root, cache_root=cache_root)
+    h = _salted_key(file_hash(p, root, cache_root=cache_root), salt)
     location = cache_root if cache_root is not None else root
     target_dir = cache_dir(location, kind, _resolve_prompt_fp(prompt, prompt_file))
     entry = target_dir / f"{h}.json"
--- a/graphify/detect.py
+++ b/graphify/detect.py
@@ -20,6 +20,7 @@ from graphify.google_workspace import (
     google_workspace_enabled,
 )
 from graphify.paths import GRAPHIFY_OUT, out_path
+from graphify.rcfile import activate_language_overrides, effective_suffix
 
 
 class FileType(str, Enum):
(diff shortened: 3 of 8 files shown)


andrebrait:
I'll patch this onto my own fork because I cannot wait for it to be ready and released. Let me know if you need help reviewing this.

abhay-codes07:
All five findings converge on one real concern — the active overrides were process-global — and it's fixed in 1ad08a9 (rebased onto 0.9.59). The overrides now live in thread-local storage, so two extractions running concurrently in one process (an MCP server, a threaded harness) with different roots each keep their own map; one scan's `activate_language_overrides` can no longer clobber the other's while `effective_suffix`/`classify_file`/`cache_salt` read it. The dominant paths are unchanged: the CLI runs on the main thread, and a spawn pool worker sets its own copy in `_worker_init` on the same main thread that then runs `_extract_single_file` (spawn workers don't inherit module state, which is why the initializer forwards it explicitly). Two regression tests cover it — concurrent threads with different `.inc` maps keep their own `effective_suffix`, and a thread that never activated overrides stays clean while another has them live; both fail on the old module-global dict.

andrebrait:
@abhay-codes07 I reviewed `1ad08a9` and integrated it on top of 0.9.59 in our maintained fork. The thread-local fix works; the remaining issues were in consumers of the override state.

I recommend adopting or adapting these follow-up commits before merging:

- [5c8aebc](https://github.com/pfBlockerNG/graphify/commit/5c8aebc9f2abdbb369730334d6e042e3575e5b25): carries the effective language through extractor grammar, cross-file resolution and cache bypass; adds declared JS/Python import candidates while preserving source paths and Python package scope.
- [4815743](https://github.com/pfBlockerNG/graphify/commit/481574300794093d44ea40d5371db0ef058b21fa): tracks language identity in incremental manifests, so config-only changes invalidate unchanged source bytes; fixes watcher configuration/source events, relative roots and semantic-work notifications.
- [a59563e](https://github.com/pfBlockerNG/graphify/commit/a59563e5c0f63d6a71eaf7df6c4c97ae30564098): closes the reviewed C# partial/inherited/dispatch gaps, applies language eligibility to inferred native candidates and JS call scope, and processes root configuration events even when the file is excluded from the corpus.

Our earlier [1010f53](https://github.com/pfBlockerNG/graphify/commit/1010f530404383face821b98c7eed3f407cb148a) also validates explicit target suffixes (for example, `.phhp` must warn instead of silently disabling extraction) and fixes the README examples to use supported whole-line comments.

The commits include behavior regressions. Final focused suite: **124 passed**; full suite: **5,720 passed, 14 existing skips**. Fork CI passed on Python 3.10 and 3.12. Native file-format conventions, including Fortran `.F90` preprocessing and C/C++ header/implementation pairing, remain unchanged.

The complete tested snapshot is [v0.9.59-pfb.1](https://github.com/pfBlockerNG/graphify/tree/v0.9.59-pfb.1); [fork PR #10](https://github.com/pfBlockerNG/graphify/pull/10) contains the review and resolution evidence. The separate Leiden packaging patch in that snapshot is not required for these language-override fixes.

andrebrait:
We found a few follow-up fixes for the language-override feature while carrying this PR on our fork, and opened them directly against your branch so a merge picks them up automatically: https://github.com/abhay-codes07/graphify/pull/1. It covers extension validation, honoring declared languages in extraction/imports, graph refresh on declaration changes, and a couple of review-driven eligibility fixes.

andrebrait:
Cross-reference: I opened #3789 as a current-`v8` integration of this work. It keeps @abhay-codes07’s original commits, authorship, and messages intact; incorporates the follow-up fixes previously proposed directly against this branch in [abhay-codes07/graphify#1](https://github.com/abhay-codes07/graphify/pull/1); and adds only the conflict/fixture adaptation needed on the current v0.9.67 base. I also updated #3789’s description to make this lineage explicit. It is not intended to erase or replace credit for #3075; it gives maintainers a complete rebased option while preserving attribution. Please use whichever review and merge shape is easiest.

PR #3076, opened 2026-08-25
feat(extract): add --memory-limit-mb so a budget overrun aborts cleanly instead of an OOM kill (#3011)

### LINKED ISSUE #3011: `graphify extract` has no memory-budget option
## Problem
`graphify extract` has no memory-budget option. When it runs inside a
memory-limited Kubernetes container, Graphify can consume the container's
entire cgroup allowance and cause the kernel to OOM-kill the pod.
Graphify v0.9.48 provides `--max-workers` / `GRAPHIFY_MAX_WORKERS`, but these
only bound the parallel AST-extraction process pool. They do not bound the
later JavaScript/TypeScript symbol-resolution phase, which retains source
buffers and syntax trees for the corpus across multiple passes.
`GRAPHIFY_REBUILD_MEMORY_LIMIT_MB` already exists for hook/watch rebuilds, but
that limit is not applied to the `extract` CLI command.
## Observed behavior
Environment:
- Linux container with an 8 GiB cgroup-v2 memory limit
- 4 CPU limit
- Graphify 0.9.32 and 0.9.48
- `graphify extract <repo> --code-only --no-cluster`
Measurements from two JavaScript-heavy repositories:
- Isolated cold scans peaked at approximately 6.5 GiB and 4.0 GiB.
- Two overlapping scans in one 8 GiB pod caused the worker container to be
  OOM-killed.
- A single v0.9.48 scan using its default worker count also reached the 8 GiB
  boundary.
- Setting `--max-workers 4` reduced AST-stage overhead, but the resolver still
  peaked around 5.9 GiB.
- Repository-specific vendor exclusions reduced one measured peak to about
  3.0 GiB, but exclusions cannot provide a general safety boundary for
  arbitrary repositories.
An external `RLIMIT_AS`/`prlimit` wrapper can stop the process before the
container OOMs, but it is a per-process virtual-address-space limit rather
than an aggregate Graphify process-tree budget, and it does not provide a
clear Graphify-specific failure.
## Requested behavior
Please add a first-class memory budget for CLI extraction, for example:
```text
graphify extract <repo> --memory-limit-mb 6144
GRAPHIFY_MEMORY_LIMIT_MB=6144
```
The limit should apply to `extract`, `update`, and the default extraction
command.
When the budget is reached, Graphify should:
1. Stop extraction and terminate all child workers.
2. Avoid publishing a partial `graph.json` or partial cache as successful.
3. Exit with a stable non-zero exit code.
4. Report the configured limit, observed usage, and active phase.
5. Where possible, account for the aggregate Graphify process tree rather
   than limiting each process independently.
Optional cgroup-v2 awareness would also be valuable: Graphify could read
`memory.max` and reserve configurable headroom when no explicit limit is
provided.
## Why this is different from `--max-workers`
The largest measured peak occurred during JavaScript/TypeScript symbol
resolution after parallel AST extraction had completed. Therefore worker
count is useful for CPU and extraction overhead, but it is not a memory
safety control.
## Related issues
- #791 added a memory limit for detached hook rebuilds.
- #792 expanded CPU worker scaling.
- #2892 is currently an empty memory issue.

abhay-codes07:
Closes #3011.

## The problem

`graphify extract` inside a memory-limited container can grow past the cgroup allowance and be OOM-killed. `--max-workers` bounds the AST pool, but the later JS/TS symbol-resolution passes retain source buffers and syntax trees for the whole corpus, and `GRAPHIFY_REBUILD_MEMORY_LIMIT_MB` only ever applied to hook/watch rebuilds — never to the `extract` CLI. The kill leaves no graphify-specific failure, no stable exit status, and whatever had been written behind (the reporter measured 4–6.5 GiB peaks and OOM kills in an 8 GiB pod).

## The change

`--memory-limit-mb N` / `GRAPHIFY_MEMORY_LIMIT_MB=N` on `extract`, `update`, and the bare `graphify <path>` form (which re-enters `extract`):

- **Applies the cap** with `setrlimit` (`RLIMIT_AS`; `RLIMIT_DATA` on macOS, whose allocator ignores `RLIMIT_AS`) to the CLI process and, through a pool `initializer`, to every extraction worker. Workers start fresh under `spawn`, so they read the cap from the environment; the flag is written back there so flag and env behave identically. An existing lower hard limit is never raised.
- **Stops demoting `MemoryError`.** Two places used to swallow it: `_safe_extract` recorded it as a *skipped file*, and the pool's per-future handler warned and retried the file in-process — which would hit the same wall in the parent. Either way a run could finish and publish a graph silently missing whatever came after. Now the pool cancels its queued work and the error propagates as a typed `MemoryBudgetExceeded` (a `MemoryError` subclass, so existing handling still applies).
- **Fails cleanly.** The CLI reports the configured limit, the phase, and the observed peak RSS, exits with **status 3** (distinct from 1 = extraction failed, 2 = bad arguments, so a wrapper can tell "give it more memory" from "the corpus is broken" without parsing stderr), and writes no `graph.json` — the previous graph is untouched. `--allow-partial` does not apply here: the operator asked to be stopped.
- **Is honest about its limits.** A malformed value is refused (exit 2) rather than silently running with no budget — a budget that quietly vanished is the failure this exists to prevent. On Windows, where there is no `setrlimit`, the CLI says the budget cannot be enforced and continues; it does not pretend.

The enforcement lives in a dependency-free `graphify/memory_budget.py`. `watch._apply_resource_limits` now delegates to it and honours the general variable too, with the hook-specific `GRAPHIFY_REBUILD_MEMORY_LIMIT_MB` keeping precedence on that path, so one setting covers every rebuild.

`RLIMIT_AS` bounds virtual address space, which over-approximates the RSS a cgroup accounts — the README says to set the budget somewhat below the container limit. It is a per-process cap rather than an aggregate tree budget; that is what the platform offers portably without a dependency, and it is the same semantic the existing hook limit has always had.

## What it looks like

```
$ graphify extract . --memory-limit-mb 6144
[graphify extract] memory budget: 6144 MB (applies to this process and its extraction workers)
...
error: memory budget of 6144 MB exceeded during AST extraction of src/generated/api.ts (peak observed in this process: ~6210 MB)
  configured: 6144 MB (--memory-limit-mb / GRAPHIFY_MEMORY_LIMIT_MB); the previous graph.json, if any, was left untouched.
  Raise the budget, narrow the corpus (.graphifyignore, --exclude), or lower --max-workers to reduce peak usage.
  exit status 3
```

On Windows:

```
[graphify extract] warning: memory budget of 512 MB cannot be enforced on this platform (no setrlimit); continuing without one
```

## Tests

`tests/test_memory_budget.py` — 29 tests: value parsing and the refused env value; the typed error; `setrlimit` really applied and really biting (a 2 GiB allocation under a 256 MB cap raises `MemoryError`, run in a subprocess so the test process is never capped) and a lower existing hard limit preserved; the pool initializer and its construction; `watch._apply_resource_limits` precedence; `_safe_extract` still swallowing ordinary failures but letting `MemoryError` through; sequential extraction stopping rather than finishing partial; a worker hitting the budget aborting the pool with the typed error instead of a "worker failed" warning; the CLI exiting 3 with no `graph.json` for both flag forms and the env var; bad values exiting 2; the unenforceable-platform warning; the ordinary #2445 failure path unchanged; and `update` taking the flag, exiting 3, and still rejecting unknown options.

With the wiring reverted and only the module kept, 12 of them fail; with it, 27 pass and the 2 `setrlimit` tests skip on Windows (they run on the Linux CI; I also ran the enforcement path under WSL: cap 256 MB → `MemoryError` on the 2 GiB allocation). The full suite matches the `v8` baseline.

README: an env-table row and a command-reference line.

### DIFF
--- a/README.md
+++ b/README.md
@@ -541,6 +541,7 @@ These are only needed for **headless / CI extraction** (`graphify extract`). Whe
 | `AZURE_OPENAI_DEPLOYMENT` or `GRAPHIFY_AZURE_MODEL` | Azure deployment name | optional — default `gpt-4o` |
 | `AWS_*` / `~/.aws/credentials` | AWS Bedrock — standard credential chain | `--backend bedrock` (no API key, uses IAM) |
 | `GRAPHIFY_MAX_WORKERS` | AST parallelism thread count | optional — also `--max-workers` flag |
+| `GRAPHIFY_MEMORY_LIMIT_MB` | Memory budget for `extract` / `update`: caps the CLI process and its extraction workers with `setrlimit`; exceeding it aborts with exit status 3 and no partial `graph.json` (Linux/macOS; reported as unenforceable on Windows) | optional — also `--memory-limit-mb` flag; set it below the container's cgroup limit |
 | `GRAPHIFY_MAX_OUTPUT_TOKENS` | Raise output cap for dense corpora | optional — e.g. `32768` for large files |
 | `GRAPHIFY_API_TIMEOUT` | Per-call timeout in seconds for HTTP, claude-cli, Anthropic SDK, and Bedrock backends (default: 600) | optional — also `--api-timeout` flag |
 | `GRAPHIFY_MAX_RETRIES` | How many times to retry a rate-limited (429) request before giving up (default: 6; honors `Retry-After`) | optional — raise for strict per-org limits (e.g. kimi); `0` disables |
@@ -764,6 +765,7 @@ graphify extract ./docs --backend bedrock      # AWS Bedrock via IAM - no API ke
 graphify extract ./docs --backend claude-cli   # route through Claude Code CLI - no API key, uses your Claude subscription
 graphify extract ./docs --backend azure        # Azure OpenAI (set AZURE_OPENAI_API_KEY + AZURE_OPENAI_ENDPOINT)
 graphify extract ./docs --max-workers 16       # AST parallelism (also GRAPHIFY_MAX_WORKERS)
+graphify extract ./src --memory-limit-mb 6144  # abort (exit 3, no partial graph) instead of being OOM-killed; also GRAPHIFY_MEMORY_LIMIT_MB
 graphify extract --postgres "postgresql://user:pass@host/db"   # introspect live PostgreSQL schema directly
 graphify extract ./my-workspace --cargo        # introspect Rust Cargo workspace dependencies directly
 graphify extract ./docs --token-budget 30000   # smaller semantic chunks for local/small models
--- a/graphify/cli.py
+++ b/graphify/cli.py
@@ -655,6 +655,47 @@ class _StageTimer:
     def total(self) -> None:
         if self.enabled:
             print(f"[graphify timing] total: {self._now() - self.start:.1f}s", file=sys.stderr)
+def _arm_memory_budget(command: str, flag_value: int | None) -> None:
+    """Put this process under the memory budget (#3011) before extraction.
+
+    ``flag_value`` (``--memory-limit-mb``) wins over ``GRAPHIFY_MEMORY_LIMIT_MB``
+    and is written back to the environment so pool workers and nested
+    rebuilds see the same cap. A malformed env value is refused (exit 2)
+    rather than ignored - a budget that silently vanished is the failure this
+    exists to prevent. Where the platform cannot enforce a limit, say so once
+    and continue; the run is otherwise identical.
+    """
+    from graphify.memory_budget import (
+        ENV_VAR,
+        apply_memory_budget,
+        configured_limit_mb,
+        supports_enforcement,
+    )
+    if flag_value is not None:
+        os.environ[ENV_VAR] = str(flag_value)
+    try:
+        limit = configured_limit_mb()
+    except ValueError as exc:
+        print(f"error: {exc}", file=sys.stderr)
+        sys.exit(2)
+    if limit is None:
+        return
+    if not supports_enforcement():
+        print(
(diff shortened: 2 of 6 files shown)


shaulga:
@abhay-codes07 thanks for taking this fix forward. do you know what else is needed to push it ?

abhay-codes07:
Worked through all three (rebased onto 0.9.59):

**Pool abort waited for running workers (real, fixed).** Correct — although the handler calls `shutdown(wait=False, cancel_futures=True)`, it re-raises out of the `with ProcessPoolExecutor() as pool` block, so the executor's `__exit__` then calls `shutdown(wait=True)` and blocks on the other in-flight workers — each potentially churning the memory the budget just tripped on, the opposite of a prompt abort. The handler now terminates the live workers before re-raising, so the implicit exit-shutdown finds them gone and returns at once.

**Flag races through process-global environment.** `_arm_memory_budget` runs once at CLI dispatch in the main process, before any pool exists; writing `GRAPHIFY_MEMORY_LIMIT_MB` to the environment is the intended mechanism for the spawn-based workers (and nested rebuilds) to inherit the cap — spawn workers don't inherit Python globals, only the environment. There's no concurrent second extract in the same process to race it. Left as-is by design.

**ThreadPool substitution runs the initializer in the pytest process.** Fair observation about the harness: the thread-pool substitution exercises the budget-application and abort-propagation logic without spawning real subprocesses (so it runs deterministically in CI), which is deliberate — the process-pool path itself is covered by the worker-init and BrokenProcessPool tests. Kept.

PR #3079, opened 2026-08-25
# fix(ruby): preserve suffixed method node IDs (#3077)

### LINKED ISSUE #3077: Ruby: `foo!`/`foo?` collide with `foo` in node IDs — the bang/predicate variant is silently dropped
## Summary

Node IDs normalise a symbol name to `[a-z0-9_]`, which strips Ruby's `!` and `?` suffixes. A bang or
predicate method therefore produces the **same ID** as its non-suffixed sibling defined in the same file,
and the dedup pass (`if n['id'] not in seen`) keeps whichever was encountered first. The other method
disappears from the graph entirely — no node, and no edges.

The ID rule in `references/extraction-spec.md` says the entity part is "the symbol name similarly
normalized", with `[a-z0-9_]` as the allowed set. `save_it` and `save_it!` both normalise to `save_it`.

## Evidence

Measured on a Rails API-only codebase (graphifyy 0.9.48, undirected build, ~420 files):

- `def` definitions found under `app/`: **631**
- method nodes in `graph.json` for `app/`: **626**
- gap: **exactly 5**

All five are ID collisions — four `foo` / `foo!` pairs and one `foo` / `foo?` pair, spread over three model
files. They account for 100% of the extraction loss: once the collisions are excluded, every remaining
`def` in the corpus has a matching node, so there is no other failure mode at play.

## Why it matters more than 0.8% suggests

The bias is directional. In Ruby the bang variant is conventionally the **persisting / raising** one, and
it is the public entry point that callers actually use:

```ruby
def attach(thing)
  self.thing = thing
end

def attach!(thing)   # <- dropped from the graph
  attach(thing)
  save!
end
```

The graph keeps the private setter and hides the public API. Any traversal, `explain`, or `path` query
over such a class will miss the persisting method, and its callers show up as isolated nodes.

The same applies to predicates: the `foo?` guard is dropped in favour of the `foo` value method.

## Reproduction

```ruby
# app/models/thing.rb
class Thing < ApplicationRecord
  def save_it
    self.state = :done
  end

  def save_it!
    save_it
    save!
  end
end
```

`graphify .` produces one node for `save_it` and none for `save_it!`.

## Suggested fix

Transliterate rather than strip, in the entity part of the ID only:

- `!` → `_bang`
- `?` → `_p` (or `_q`)
- `=` → `_eq` (setters have the same problem: `foo` vs `foo=`)

`save_it!` → `app_models_thing_save_it_bang`. This keeps the ID charset at `[a-z0-9_]`, stays deterministic
from the label alone, and is a pure win — no existing non-suffixed ID changes, so previously built graphs
keep matching on `--update` for every method that has no suffixed sibling.

Ruby is the obvious case, but the same normalisation affects any language with suffixed identifiers
(Scheme/Clojure `foo?`/`foo!`, Ruby setters `foo=`).

## Environment

- graphifyy 0.9.48, skill 0.9.20
- corpus: ~420 files, Rails API-only app, 1773 nodes / 2161 edges
- undirected build, `--update` incremental

hopstreax:
# Summary

Fixes #3077.

Ruby methods such as `foo`, `foo!`, `foo?`, and `foo=` were previously normalized to the same node ID. When defined in the same class/file, the first method encountered by the extractor was retained and the others were silently dropped by `seen_ids`.

This caused missing nodes and incorrect/missing call relationships for common Ruby bang, predicate, and setter methods.

## What changed

* Added an optional `sanitize_symbol_name_fn` to `LanguageConfig`.
* Added Ruby-specific sanitization for trailing method suffixes:

  * `!` → `_bang`
  * `?` → `_pred`
  * `=` → `_eq`
* Kept the global `normalize_id()` / `make_id()` behavior unchanged.
* Preserved the original Ruby method spelling in node labels.
* Updated Ruby member-call resolution to use the raw method spelling instead of stripping punctuation.
* Added regression coverage for:

  * `foo`, `foo!`, `foo?`, and `foo=`
  * singleton methods
  * top-level methods
  * Ruby call resolution
  * suffixed-method ID stability across extraction updates

## Example

Before:

```ruby
def save
end

def save!
end
```

Both methods normalized to the same node ID, causing `save!` to be dropped.

After:

```text
save  → *_save
save! → *_save_bang
```

Both methods are represented as distinct nodes while their labels retain the original Ruby spelling.

## Validation

Focused Ruby/resolution tests:

```text
37 passed
```

ID normalization and extraction-spec contract tests:

```text
122 passed
```

Broader suite:

```text
668 passed
4 skipped
2 failed
```

The two failures are unrelated Windows environment issues involving Unicode `cp1252` encoding and deeply nested temporary paths.

Global ID normalization and other language extractors remain unchanged.

## Scope

This PR addresses Ruby suffixed method names (`!`, `?`, `=`). Ruby operator-method handling is intentionally left out of this change.

### DIFF
--- a/graphify/extract.py
+++ b/graphify/extract.py
@@ -901,6 +901,19 @@ _CPP_CONFIG = LanguageConfig(
     resolve_function_name_fn=_get_cpp_func_name,
 )
 
+def _ruby_sanitize_method_name(name: str) -> str:
+    """Encode trailing Ruby method suffixes (!, ?, =) into safe node ID components (#3077)."""
+    if not name:
+        return name
+    if name.endswith("!"):
+        return f"{name[:-1]}_bang"
+    if name.endswith("?"):
+        return f"{name[:-1]}_pred"
+    if name.endswith("="):
+        return f"{name[:-1]}_eq"
+    return name
+
+
 _RUBY_CONFIG = LanguageConfig(
     ts_module="tree_sitter_ruby",
     # `module Foo` is a container node just like `class Foo` in tree-sitter's
@@ -917,6 +930,7 @@ _RUBY_CONFIG = LanguageConfig(
     name_fallback_child_types=("constant", "scope_resolution", "identifier"),
     body_fallback_child_types=("body_statement",),
     function_boundary_types=frozenset({"method", "singleton_method"}),
+    sanitize_symbol_name_fn=_ruby_sanitize_method_name,
 )
 
 _CSHARP_CONFIG = LanguageConfig(
--- a/graphify/extractors/engine.py
+++ b/graphify/extractors/engine.py
@@ -4085,19 +4085,24 @@ def _extract_generic(
 
             if not func_name:
                 return
+            sanitized_name = (
+                config.sanitize_symbol_name_fn(func_name)
+                if config.sanitize_symbol_name_fn is not None
+                else func_name
+            )
             # A name that normalizes to nothing collapses `_make_id(prefix, name)`
             # onto the (absolute-path-derived) prefix, leaking the scan path and
             # colliding with the file/class node (#1899). No graph signal; skip.
-            if not normalize_id(func_name):
+            if not normalize_id(sanitized_name):
                 return
 
             line = node.start_point[0] + 1
             if parent_class_nid:
-                func_nid = _make_id(parent_class_nid, func_name)
+                func_nid = _make_id(parent_class_nid, sanitized_name)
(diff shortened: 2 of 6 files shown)


safishamsi:
Shipped in v0.9.50 via authorship-preserving cherry-pick so you keep contributor-graph credit. Thanks @hopstreax! Release: https://github.com/Graphify-Labs/graphify/releases/tag/v0.9.50

PR #3082, opened 2026-08-25
fix(ruby): match a qualified receiver by its constant path

### LINKED ISSUE #3078: Ruby/Rails: a qualified receiver's last segment resolves to a same-named local class — every `ActiveRecord::Base.transaction` binds to the corpus's only `Base` (false EXTRACTED edges)
## Summary

A qualified receiver whose last segment matches a locally-defined class binds to that local class, with no
scope check. In a Rails codebase this means **every `ActiveRecord::Base.transaction` / `.connection` call
binds to whatever single class named `Base` the corpus happens to define** — in our case a nested
`Base` class inside an unrelated service object, with no relationship whatsoever to ActiveRecord.

This is the label-text match described in #2041 point 4 ("a lookup by the callee identifier's text matching
an existing node's label, not by any real scope/identity check"), but the failure mode here is the opposite
of that issue's Scala symptom: instead of **dropping** edges, it **creates false ones**, and they are
tagged `EXTRACTED` with `confidence_score: 1.0`.

## Evidence

Rails API-only codebase, ~420 files, graphifyy 0.9.48.

**14 `calls` edges** point at one nested `Base` class defined in a service object. They originate from
**9 different files** (5 under `app/`, 4 under `spec/`). Every single source line is a framework call, not
a call to that local class — the whole set reduces to these forms:

```ruby
ActiveRecord::Base.transaction(requires_new: true) { ... }
ActiveRecord::Base.transaction(isolation: :repeatable_read) do ... end
ActiveRecord::Base.connection.execute(...)
ActiveRecord::Base.with_connection do |connection| ... end
ActiveJob::Base.default_queue_name
```

Note the last one: `ActiveJob::Base` resolves to the same node as `ActiveRecord::Base`. The namespace is
discarded entirely; only the final segment is matched.

The corpus contains exactly one class node labelled `Base` **with a source file**, plus 31 file-less `Base`
placeholders — the dangling type-reference placeholders #2041 point 4 warns are present in the label index.
The single real one wins every time.

## Impact

The edges are not merely noise; they change the analysis output:

1. That nested `Base` class becomes a **false hub** — 14 inbound `calls` it never receives, more than most
   real domain classes in the corpus.
2. Community detection then merges two entirely unrelated modules (19 nodes and 18 nodes) into one
   community, whose **cohesion score of 0.063 is the lowest in the graph**.
3. That low score generates a suggested question in `GRAPH_REPORT.md` asking whether the merged community
   should be split into smaller, more focused modules — a recommendation that is entirely an artefact.
   Acting on it would mean refactoring real code to satisfy phantom edges.

The confidence tagging makes this hard to spot: these are `EXTRACTED` / `1.0`, the same tags as a genuine
same-file call.

## Reproduction

```ruby
# app/services/thing.rb
class Thing
  class Base            # any locally-defined class named Base
    def self.call(x) = x
  end
end

# app/models/other.rb
class Other < ApplicationRecord
  def work
    ActiveRecord::Base.transaction { save! }
  end
end
```

`graphify .` emits `Other#work --calls--> Thing::Base`.

Generalises beyond `Base`: any framework constant whose last segment collides with a local class name
(`Rails::Application`, `ActiveSupport::Notifications`, `Net::HTTP` vs a local `HTTP`).

## Suggested fixes, in order of cost

1. **Cheapest and highest value:** when the call site's receiver is *qualified* (`A::B.m`), require the
   candidate's own qualified name to match the full receiver path, not just the last segment. If the local
   class's full name is `Thing::Base` and the receiver is `ActiveRecord::Base`, reject.
2. **Deny-list the framework roots** for the relevant language (`ActiveRecord`, `ActiveJob`,
   `ActionController`, `ActiveSupport`, …) so an unresolvable framework receiver yields no edge rather than
   a wrong one.
3. Failing either, downgrade last-segment-only matches from `EXTRACTED`/`1.0` to `AMBIGUOUS` with a low
   score, so the honesty rules surface them instead of asserting them.

Dropping the edge is strictly better than inventing one here: a missing edge is a known blind spot, a false
edge silently corrupts community detection and the report's recommendations.

## Related

- #2041 — describes the underlying label-text match; this is the false-positive side of it, in Ruby.
- #437 — same symptom (cross-file inference creating false edges via common method names) on .NET.
- #2315 — Python decorators binding to an unrelated module-level function of the same bare name; same root
  cause, different surface.

## Environment

- graphifyy 0.9.48, skill 0.9.20
- corpus: ~420 files, Rails API-only app, 1773 nodes / 2161 edges, undirected

rohit-jsfreaky:
Fixes #3078.

## Summary

A namespaced Ruby receiver was truncated to its last constant at extraction time, so the namespace never reached the resolver. `engine.py` captured `ActiveRecord::Base.transaction` as the receiver `Base`, and the resolver then bound it to whatever single class named `Base` the corpus defined.

The comment there noted the god-node guard bails when ambiguous — but that guard only catches an *ambiguous* match, not a unique-but-wrong one. With exactly one `Base` in the corpus nothing is ambiguous, so every framework call became an `EXTRACTED` / `1.0` edge into an unrelated class. `ActiveRecord::Base` and `ActiveJob::Base` collapsed onto the same node, because only the final segment survived.

## Changes

- `engine.py`: keep the whole constant path for a `scope_resolution` receiver, via the existing `_ruby_const_full_name` helper.
- `ruby_resolution.py`: add `_class_by_const_path`, which matches on the constant path rather than its tail — a class qualifies when its own label *ends with* the referenced segments. `Billing::Processor` therefore still finds an `App::Billing::Processor`, while `ActiveRecord::Base` no longer matches `Thing::Base`. A leading `::` pins the reference to top level and must match the label whole. Ambiguous, or matching nothing in the corpus (the usual case for a framework constant) -> no edge.
- The constant-receiver check now looks past leading colons, so a top-level-pinned `::Processor.call` is still recognised now that the receiver text can start with `::`.

This reuses the `fq_label_map` / `_segment_path` machinery already built for the scoped mixin lookup, and only qualified receivers change path — a bare `Processor.call` still goes through `_unique_class` exactly as before. `fq_label_map` is built from source-backed nodes only (the `.rb` filter documented above it), so this cannot bind to one of the file-less external placeholders the issue mentions.

**Behaviour change worth flagging:** a qualified receiver now requires its namespace to be present in the corpus. A class declared top-level as `Processor` but referenced as `Billing::Processor` no longer produces an edge. That is the same class of false edge this fixes, so I believe it is correct, but it is a deliberate tightening rather than a pure bug fix.

## Reproduction

Two files — a nested `Thing::Base` and a model making framework calls:

```ruby
# app/services/thing.rb
class Thing
  class Base
    def self.call(x) = x
  end
end

# app/models/other.rb
class Other
  def work = ActiveRecord::Base.transaction { save! }
  def more = ActiveJob::Base.default_queue_name
end
```

Before, on 0.9.49:

```
.work() -- calls --> Thing::Base | EXTRACTED
.more() -- calls --> Thing::Base | EXTRACTED
```

After: neither edge is emitted. All five forms listed in the issue were checked — `transaction(requires_new:)`, `transaction do…end`, `.connection.execute`, `with_connection do |c|`, and `ActiveJob::Base.default_queue_name` — 0 false edges.

## Tests

Three appended to `tests/test_ruby_resolution.py`. The first is the repro and fails without the source change:

```
FAILED test_framework_qualified_receiver_does_not_bind_same_named_local_class
```

The other two are guards rather than repros — they pass before and after, and exist so the namespace check cannot be tightened at the cost of a genuine edge: a correctly-namespaced `Billing::Processor.run` still resolves as `EXTRACTED`, and a top-level-pinned `::Solo.go` still resolves.

`uv run --no-sync pytest tests/test_ruby_resolution.py -q` — 29 passed (was 26).

`uv run --no-sync ruff check graphify/ruby_resolution.py graphify/extractors/engine.py` — passed.

Full suite: 14 failed, 4984 passed, 41 skipped. All 14 are pre-existing on a clean `v8` checkout of the same commit on this Windows machine — 12 platform tests (fifo / unix-socket / hermes / gemini paths / watch) and 2 FalkorDB integration tests that fail against the local port with a redis `ResponseError`. I verified both sets by running them in a detached worktree at `282976b` and diffing the failure lists; they match exactly.

### DIFF
--- a/graphify/extractors/engine.py
+++ b/graphify/extractors/engine.py
@@ -5282,10 +5282,13 @@ def _extract_generic(
                     if recv.type in ("identifier", "constant"):
                         member_receiver = _read_text(recv, source)
                     elif recv.type == "scope_resolution":
-                        # Namespaced receiver `Billing::Processor.call` — capture the
-                        # last constant so cross-file resolution can bind it by the
-                        # bare class name (the god-node guard bails if ambiguous).
-                        member_receiver = _ruby_const_last_name(recv, source) or None
+                        # Namespaced receiver `Billing::Processor.call` — keep the whole
+                        # constant path. Truncating to the last segment discarded the
+                        # namespace, so `ActiveRecord::Base.transaction` bound to
+                        # whatever single class named `Base` the corpus defined: the
+                        # god-node guard only catches an ambiguous match, not a
+                        # unique-but-wrong one (#3078).
+                        member_receiver = _ruby_const_full_name(recv, source) or None
             else:
                 # Generic: get callee from call_function_field
                 func_node = node.child_by_field_name(config.call_function_field) if config.call_function_field else None
--- a/graphify/ruby_resolution.py
+++ b/graphify/ruby_resolution.py
@@ -119,6 +119,28 @@ def resolve_ruby_member_calls(
         nids = class_def_nids.get(_key(name), [])
         return nids[0] if len(nids) == 1 else None
 
+    def _class_by_const_path(raw: str) -> str | None:
+        """Resolve a qualified constant receiver (``Billing::Processor``) to one class.
+
+        Matches on the constant path rather than its tail: a class qualifies when its
+        own label ends with the referenced segments, so ``Billing::Processor`` still
+        finds an ``App::Billing::Processor`` while ``ActiveRecord::Base`` no longer
+        binds to an unrelated ``Thing::Base`` (#3078). A leading ``::`` pins the
+        reference to top level, so it must match the label whole. Ambiguous, or
+        matching nothing in the corpus (the usual case for a framework constant) ->
+        no edge, never a guess.
+        """
+        segs = tuple(_segment_path(raw))
+        if not segs:
+            return None
+        if raw.strip().startswith("::"):
+            nids = fq_label_map.get(segs, [])
+            return nids[0] if len(nids) == 1 else None
+        hits = {nid for path, nids in fq_label_map.items()
+                if len(path) >= len(segs) and path[-len(segs):] == segs
+                for nid in nids}
+        return next(iter(hits)) if len(hits) == 1 else None
+
     def _emit(caller: str, target: str, rc: dict[str, Any],
               relation: str = "calls", context: str = "call") -> None:
         if not caller or not target or caller == target:
@@ -187,8 +209,12 @@ def resolve_ruby_member_calls(
         # collide with unrelated same-named methods, so we resolve by the
         # receiver's class under the single-owning-class god-node guard.
         receiver = rc.get("receiver")
-        if receiver and str(receiver)[:1].isupper():
-            class_nid = _unique_class(str(receiver))
+        # `lstrip(":")` so a top-level-pinned `::Processor.call` is still recognised
+        # as a constant receiver now that the whole path is captured (#3078).
+        if receiver and str(receiver).lstrip(":")[:1].isupper():
+            recv_raw = str(receiver)
+            class_nid = (_class_by_const_path(recv_raw) if "::" in recv_raw
+                         else _unique_class(recv_raw))
             if class_nid is not None:
                 if callee == "new":
                     _emit(caller, class_nid, rc)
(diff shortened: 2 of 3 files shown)


safishamsi:
Shipped in v0.9.50 via authorship-preserving cherry-pick so you keep contributor-graph credit. Thanks @rohit-jsfreaky! Release: https://github.com/Graphify-Labs/graphify/releases/tag/v0.9.50

PR #3086, opened 2026-08-25
fix(windows): stop a PowerShell BOM breaking hook rebuilds

### LINKED ISSUE #3028: Windows PowerShell 5.1: Out-File -Encoding utf8 BOM in .graphify_python/.graphify_root breaks post-commit hook rebuild (WinError 123)
### Environment

- OS: Windows 11 Pro
- Shell: Windows PowerShell 5.1 (the shell the skill's own code blocks are written for)
- graphify: 0.9.49 (CLI via `pip install graphifyy`, skill installed via `graphify install`)
- Python: 3.14, plain `pip`-into-active-env install (not uv/pipx)

### What happened

After running the full `/graphify .` pipeline once, I ran `graphify hook install` (post-commit + post-checkout) so the graph would auto-rebuild after future commits, per `references/hooks.md`.

The next `git commit` correctly triggered the hook (`[graphify hook] launching background rebuild (log: ~/.cache/graphify-rebuild.log)`), but the rebuild silently failed every time:

```
[graphify hook] 3 file(s) changed - rebuilding graph...
[graphify hook] Rebuild failed: [WinError 123] The filename, directory name, or volume label syntax is incorrect: '\ufeffC:\Users\<user>\...\graphify-out'
```

Note the `\ufeff` (UTF-8 BOM) prepended to the path — that's what `WinError 123` is complaining about.

### Root cause

SKILL.md Step 1 (the block every fresh `/graphify <path>` run executes) writes the two sidecar files the hook later reads:

```powershell
$GRAPHIFY_PYTHON | Out-File -FilePath graphify-out\.graphify_python -Encoding utf8 -NoNewline
(Resolve-Path INPUT_PATH).Path | Out-File -FilePath graphify-out\.graphify_root -Encoding utf8 -NoNewline
```

(SKILL.md lines 120 and 122 in 0.9.49.)

`Out-File -Encoding utf8` in **Windows PowerShell 5.1** always emits a UTF-8 BOM (`EF BB BF`) — there is no BOM-less `utf8` codepage on 5.1 (that only exists as `utf8NoBOM` starting in PowerShell 6+). `-NoNewline` suppresses the trailing newline but does nothing about the BOM.

So `graphify-out/.graphify_python` and `graphify-out/.graphify_root` end up as e.g. `b'\xef\xbb\xbfC:\Python314\python.exe'` instead of `b'C:\Python314\python.exe'`.

Whatever code path the hook uses to read `.graphify_root` (and hand it to Python's path/subprocess machinery) doesn't strip the BOM, so the leading `\ufeff` character rides along into a Windows path API and blows up with `WinError 123`.

This is nasty because it's **silent from the user's perspective in the normal flow**: `graphify hook install` reports success, `hook status` reports "installed", the hook visibly fires on every commit — but the rebuild itself fails every single time, so the graph silently goes stale forever unless someone thinks to check `~/.cache/graphify-rebuild.log`. Anyone on plain Windows PowerShell 5.1 (i.e. anyone who didn't proactively install PowerShell 7) who follows the skill's own documented bootstrap hits this on the very first commit after `hook install`.

### Fix I applied locally (works)

Rewrite both sidecar files without a BOM:

```powershell
$pythonPath = (Get-Content graphify-out\.graphify_python -Raw).Trim()
$rootPath = (Get-Content graphify-out\.graphify_root -Raw).Trim()
[System.IO.File]::WriteAllText("graphify-out\.graphify_python", $pythonPath, [System.Text.UTF8Encoding]::new($false))
[System.IO.File]::WriteAllText("graphify-out\.graphify_root", $rootPath, [System.Text.UTF8Encoding]::new($false))
```

After that, `graphify update .` and the post-commit hook both rebuild cleanly.

### Suggested upstream fix

Two independent, non-exclusive fixes — either alone closes this:

1. **Write without a BOM in the first place.** In SKILL.md Step 1 (and anywhere else in the skill/CLI that writes `.graphify_python` / `.graphify_root` via PowerShell), swap `Out-File -Encoding utf8` for either `-Encoding utf8NoBOM` (PS7+) with a version-gated fallback to `[System.IO.File]::WriteAllText(path, content, [System.Text.UTF8Encoding]::new($false))` on 5.1, or just always use the `[System.IO.File]::WriteAllText` form directly — it's BOM-less on every PowerShell version and sidesteps the whole `-Encoding` footgun.
2. **Defense in depth on read.** Wherever the Python side reads these two sidecar files (hook rebuild path resolution, `graphify update`'s interpreter/root detection, etc.), `.lstrip('\ufeff')` the contents before using them as a path/interpreter string. Costs nothing and guards against any other BOM-writing code path (existing files from other editors, future regressions) that hits the same failure mode.

Happy to open a PR for #1 if useful — the fix is small and I already have it working locally.

rohit-jsfreaky:
Fixes #3028.

## Summary

The Windows bootstrap writes the two sidecar markers with `Out-File -Encoding utf8`. On Windows PowerShell 5.1 that always prepends a UTF-8 BOM — `utf8NoBOM` only exists from PowerShell 6, and `-NoNewline` does nothing about it. So `graphify-out/.graphify_python` and `.graphify_root` start with `EF BB BF`.

Every reader then decodes them with `utf-8`, which keeps the BOM as `U+FEFF`, and `str.strip()` does not remove it — `U+FEFF` is not whitespace. The character rides into path handling and the post-commit rebuild dies with `WinError 123`, while `hook install` and `hook status` both keep reporting success. The graph silently goes stale unless someone opens `~/.cache/graphify-rebuild.log`.

`WinError 123` is the loud failure. There is a quiet one too: in `_infer_merge_root` the BOM makes an absolute path stop being drive-qualified, so `Path(recorded).resolve()` treats it as *relative* and silently joins it onto the cwd — no exception, just a wrong scan root:

```
D:\...\Graphify-Labs--graphify\<BOM>C:\Users\<user>\...\proj
```

## Changes

**Read side — the half that repairs installs already broken.** Five call sites now decode with `utf-8-sig`, which strips an optional BOM and is a no-op on a clean file: `hooks.py` (both rebuild bodies), `watch.py`, `reflect.py`, `build.py`. Rewriting the markers alone would only help fresh runs; anyone already in this state stays broken until they re-run the bootstrap. With `utf-8-sig` they recover on the next commit.

**Write side — stop producing the BOM.** `fragments/shell/powershell.md` writes both markers through `[System.IO.File]::WriteAllText` with an explicit `New-Object System.Text.UTF8Encoding $false`, which emits no BOM and appends no newline — the same bytes the POSIX fragment already writes with `echo >`. Regenerated with `python -m tools.skillgen` + `--bless`; only the windows platform uses this fragment, so `skill-windows.md` and its `expected/` snapshot are the only rendered artifacts that change.

## Verification on the actual shell

Both forms run on Windows PowerShell **5.1.26100.9168**, writing the same interpreter path:

| form | first 3 bytes | BOM | length |
|---|---|---|---|
| `Out-File -Encoding utf8 -NoNewline` (today) | `EF BB BF` | yes | 26 |
| `WriteAllText` + `UTF8Encoding $false` (this PR) | `43 3A 5C` (`C:\`) | no | 23 |

Exactly the three BOM bytes apart, and no trailing newline either way.

## Tests

Four added, each confirmed to fail with the source changes reverted and the tests kept:

- `test_graphify_root_marker_with_a_utf8_bom_still_resolves` — the repro: a marker written with a real `EF BB BF` prefix must still resolve to the true root through `_infer_merge_root`.
- `test_rebuild_bodies_tolerate_a_bom_in_graphify_root` (both hook bodies) — the generated rebuild must read the marker BOM-tolerantly, since that is the site that produced `WinError 123`.
- `test_windows_skill_writes_marker_files_without_a_bom` — the rendered Windows skill must not write either marker through a BOM-emitting `Out-File`.

One existing test needed updating rather than working around: `test_rebuild_bodies_read_graphify_root` asserted the literal `read_text(encoding='utf-8')`. Its stated intent is shell-quote safety inside the double-quoted launcher ("single quotes only"), which `utf-8-sig` still satisfies, so the literal moved with a comment explaining why the codec changed.

hooks 95 passed (2 pre-existing platform failures), skillgen 65 passed, build-merge 10 passed.

All five skillgen guards pass: `--check` (134 artifacts), `--audit-coverage`, `--schema-singleton`, `--monolith-roundtrip`, `--always-on-roundtrip`.

`uv run --no-sync ruff check` on every touched file — passed.

Full suite: 14 failed, 4985 passed, 41 skipped — the same 14 as a clean `v8` checkout of this commit on this Windows machine (12 platform tests: fifo / unix-socket / hermes / gemini paths / watch; plus 2 FalkorDB integration tests failing against the local port with a redis `ResponseError`). Verified by running them in a detached worktree at `282976b` and diffing the failure lists.

I have not run an end-to-end `git commit` against a BOM'd install; the evidence here is the byte-level PowerShell comparison above plus the reproduced path corruption.

Note: this touches `graphify/skill-windows.md`, which #2677 also touches — at line 645 there versus 120/122 here, so they should merge cleanly in either order.

### DIFF
--- a/graphify/build.py
+++ b/graphify/build.py
@@ -439,7 +439,7 @@ def _infer_merge_root(graph_path: Path) -> str | None:
     try:
         marker = parent / ".graphify_root"
         if marker.exists():
-            recorded = marker.read_text(encoding="utf-8").strip()
+            recorded = marker.read_text(encoding="utf-8-sig").strip()
             if recorded:
                 return str(Path(recorded).resolve())
     except OSError:
--- a/graphify/hooks.py
+++ b/graphify/hooks.py
@@ -165,7 +165,7 @@ try:
     _out = os.environ.get('GRAPHIFY_OUT', 'graphify-out')
     _saved = Path(_out) / '.graphify_root'
     if _saved.exists():
-        _txt = _saved.read_text(encoding='utf-8').strip()
+        _txt = _saved.read_text(encoding='utf-8-sig').strip()
         if _txt:
             _root = Path(_txt)
     _rebuild_code(_root, changed_paths=changed, force=_force)
@@ -214,7 +214,7 @@ try:
     _out = os.environ.get('GRAPHIFY_OUT', 'graphify-out')
     _saved = Path(_out) / '.graphify_root'
     if _saved.exists():
-        _txt = _saved.read_text(encoding='utf-8').strip()
+        _txt = _saved.read_text(encoding='utf-8-sig').strip()
         if _txt:
             _root = Path(_txt)
     _rebuild_code(_root, force=_force)
--- a/graphify/reflect.py
+++ b/graphify/reflect.py
@@ -691,7 +691,7 @@ def _resolve_source_path(src: str, graph_path: Path) -> Path | None:
     out_dir = gp.parent
     candidates: list[Path] = []
     try:
-        recorded = (out_dir / ".graphify_root").read_text(encoding="utf-8").strip()
+        recorded = (out_dir / ".graphify_root").read_text(encoding="utf-8-sig").strip()
         if recorded:
             candidates.append(Path(recorded))
     except (OSError, ValueError):
--- a/graphify/skill-windows.md
+++ b/graphify/skill-windows.md
@@ -116,10 +116,15 @@ if (-not $GRAPHIFY_PYTHON) {
     $GRAPHIFY_PYTHON = Find-GraphifyPython
 }
 
-# Save interpreter path — all subsequent steps read this
-$GRAPHIFY_PYTHON | Out-File -FilePath graphify-out\.graphify_python -Encoding utf8 -NoNewline
+# Save interpreter path — all subsequent steps read this.
+# `Out-File -Encoding utf8` always writes a BOM on Windows PowerShell 5.1 (utf8NoBOM
+# only exists from PowerShell 6), and that BOM rides into the saved path, so the hook
+# rebuild fails with WinError 123 (#3028). WriteAllText with an explicit BOM-less
+# encoding writes the bytes POSIX writes, and adds no trailing newline.
+$Utf8NoBom = New-Object System.Text.UTF8Encoding $false
+[System.IO.File]::WriteAllText((Join-Path $PWD 'graphify-out\.graphify_python'), [string]$GRAPHIFY_PYTHON, $Utf8NoBom)
 # Save scan root so `graphify update` (no args) knows where to look next time
-(Resolve-Path INPUT_PATH).Path | Out-File -FilePath graphify-out\.graphify_root -Encoding utf8 -NoNewline
+[System.IO.File]::WriteAllText((Join-Path $PWD 'graphify-out\.graphify_root'), (Resolve-Path INPUT_PATH).Path, $Utf8NoBom)
 ```
 
 If the import succeeds, print nothing and move straight to Step 2.
--- a/graphify/watch.py
+++ b/graphify/watch.py
@@ -369,7 +369,7 @@ class _StoredSourcePaths:
         root_marker = out / ".graphify_root"
         if root_marker.exists():
             try:
-                saved_root = Path(root_marker.read_text(encoding="utf-8").strip())
+                saved_root = Path(root_marker.read_text(encoding="utf-8-sig").strip())
                 if saved_root.is_absolute():
                     # #2603: the marker holds the SCAN root, but stored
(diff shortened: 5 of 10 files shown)


safishamsi:
Shipped in v0.9.50 via authorship-preserving cherry-pick so you keep contributor-graph credit. Thanks @rohit-jsfreaky! Release: https://github.com/Graphify-Labs/graphify/releases/tag/v0.9.50

PR #3088, opened 2026-08-25
feat(extract): extract systemd units so a repo's scheduled-job topology reaches the graph (#2848)

### LINKED ISSUE #2848: systemd units (.service/.timer) are in no extension set, so a repo's whole scheduled-job topology is invisible — and the graph answers confidently rather than emptying
## Summary

`.service` and `.timer` appear in none of `CODE_EXTENSIONS` / `DOC_EXTENSIONS` / `PAPER_EXTENSIONS` /
`IMAGE_EXTENSIONS`, so `classify_file()` returns `None` and systemd units are never collected at any cost
tier. For a repo that keeps its units under version control, that means **the entire OS-level scheduled-job
topology is absent from the graph**.

Verified on `graphifyy==0.9.42`:

```python
from graphify.detect import CODE_EXTENSIONS, DOC_EXTENSIONS, PAPER_EXTENSIONS, IMAGE_EXTENSIONS
watched = CODE_EXTENSIONS | DOC_EXTENSIONS | PAPER_EXTENSIONS | IMAGE_EXTENSIONS
".service" in watched   # False
".timer"   in watched   # False
```

Same class as **#2434** (`.toml`/`.ini`/`.cfg`). Worth flagging: #2434 is closed as *completed*, but on
0.9.42 `".toml" in watched` is still `False` — so either the fix is merged-but-unreleased, or the class is
still open in practice. Not a complaint, just a data point for whoever triages this one.

## Why this is worse than a missing extractor

The failure is **silent and confident**, not empty. In our repo (11 tracked units — a daily audit, a Slack
digest, a watchdog, a backup, a knowledge-graph refresh):

```
$ graphify query "what scheduled jobs run on sl" --budget 400
Traversal: BFS depth=2 | Start: ['run()', 'runScheduledIndexing()'] | 59 nodes found
```

59 nodes, all from the application's in-process scheduler, **zero timers**. Nothing signals that the units
were skipped — the answer looks complete. The related `unclassified` count *is* computed by `detect()`, but
the print at `cli.py:3181-3187` lives in the `elif cmd == "extract":` branch, so `graphify update` never
emits it and there is no log line to notice either.

## Why units are worth extracting (unlike most config)

They are genuinely graph-shaped, and every edge lands on nodes an AST pass has already created:

| Relation | From → To | Source key |
|---|---|---|
| `activates` | `*.timer` → `*.service` | `[Timer] Unit=`, else the filename convention |
| `runs` | `*.service` → a script node | `ExecStart=`, after stripping `/usr/bin/env bash …` |
| `documented_by` | unit → a doc node | `Documentation=file://…` |
| `after`/`wants`/`requires` | unit → unit | `[Unit]` ordering keys |

No tree-sitter grammar is needed — units are INI. `extractors/sln.py` (92 lines, pure `re`) looks like the
right template.

## What we did meanwhile, in case it is useful

Rather than patch `site-packages` (reverted by `uv tool upgrade`), we wrote a ~500-line deterministic pass in
our own tooling that runs *after* `graphify update` and merges the units into `graph.json`. It yields
**11 nodes + 18 edges** and turns the query above into a real answer:

```
$ graphify path config_systemd_sl_graphify_refresh_timer scripts_graphify_refresh
Shortest path (2 hops):
  sl-graphify-refresh.timer --activates [EXTRACTED]--> sl-graphify-refresh.service
                            --calls/runs [EXTRACTED]--> graphify-refresh.sh
```

Happy to port the parser semantics into `graphify/extractors/systemd.py` as a PR if that is wanted — but
given the current PR backlog we would rather ask first than add to it. The issue stands on its own.

## Two things that would bite an implementation

1. **Do not add the extensions without the `_DISPATCH` entry** (`extract.py:4769`) — **#1689** documents that
   exact combination silently dropping every matching file with no warning.
2. **Node ids collide.** `_file_stem` drops the extension, so `foo.service` and `foo.timer` — which is the
   *normal* systemd naming convention, not an edge case — collapse into one node. This is **#1829** with a
   guaranteed-to-occur trigger rather than an incidental basename clash; we had to keep the extension in the
   id (`config_systemd_foo_service` / `…_timer`) to avoid it. Commented there separately.

## Environment

- `graphifyy==0.9.42`, Python 3.12, Ubuntu 24.04
- Repo: ~900 tracked files, 5,282 nodes; 11 `config/systemd/*.{service,timer}`

abhay-codes07:
Closes #2848.

## The problem

`.service` and `.timer` were in no extension set, so `classify_file()` returned `None` and systemd units were never collected. For a repo that keeps its units under version control, the whole OS-level scheduled-job topology was absent from the graph — and silently so: `graphify query "what scheduled jobs run"` answered with 59 nodes from the application's in-process scheduler and zero timers, looking complete.

## The change

A deterministic INI pass, `extractors/systemd.py` (modelled on `extractors/sln.py`, no grammar), covering `.service .timer .socket .target .path .mount .slice`. Every unit is a file node; edges land on nodes the AST pass already creates:

| relation | from → to | source key |
|---|---|---|
| `activates` | timer/socket/path → unit | `[Timer] Unit=` / `[Socket] Service=` / `[Path] Unit=`, else the same-stem `.service` by systemd's own convention |
| `runs` | service → script | `ExecStart=` and the other `Exec*=` keys, after stripping the `-@:+!` prefixes, `/usr/bin/env` (with its `VAR=` and `-S` arguments) and the interpreter (`python3 -u`, `bash`, `node`, `uv run`, `npx tsx`, …); `python -m pkg` is a module, not a file |
| `documented_by` | unit → doc | `Documentation=file://…` |
| `after` `before` `wants` `requires` `binds_to` `part_of` `conflicts` `wanted_by` `required_by` | unit → unit | the `[Unit]` and `[Install]` keys |

Two resolution rules keep it honest:

- **Unit → unit edges only target a unit file beside this one.** `After=network-online.target` and `WantedBy=timers.target` name the host's units; manufacturing a node for each would put a phantom hub in every repo. A template instance (`backup@nightly.service`) resolves to its template (`backup@.service`).
- **Exec/Documentation values are deployment paths** (`/opt/app/bin/run.py`) that rarely exist at that path in the repo. Resolution tries the literal path, then walks up from the unit's directory looking for the same tail (`bin/run.py`, then `run.py`) — the usual `units/` beside `bin/` layout. A target that resolves is minted the way every other extractor mints a file reference (`_make_id(str(resolved))`), which `extract()` rewires onto the real file node; one that does not is skipped, not fabricated. Deployment paths are parsed as POSIX so they are absolute on every host graphify runs on.

### One interaction worth calling out

`x.service` and `x.timer` are the most common pair a repo has, and they share the extension-less file-node id. The corpus-level collision remap (`_disambiguate_colliding_node_ids`) resolves an edge's target by the *edge's own file*, so the timer's `activates` edge came out as a self-loop on the timer. Import edges already solve this with a transient `target_file` stamp naming the file the target id was minted from; systemd edges now carry it too, and the disambiguator's whitelist of stamped relations is named (`_TARGET_FILE_RELATIONS`) and widened. The stamp is popped before anything ships, as before.

## What it looks like

```
backup-nightly.timer  --activates-->      backup@.service
backup@.service       --runs-->           backup.sh
daily-audit.service   --documented_by-->  audit.md
daily-audit.service   --runs-->           daily_audit.py
daily-audit.timer     --activates-->      daily-audit.service
```

with `After=network-online.target`, `WantedBy=timers.target` and a nonexistent `backup.service` correctly producing nothing.

## Tests

`tests/test_systemd_units.py` — 40 tests: every unit type classified and dispatched; the per-file extractor (script, doc, implied and explicit activation, template instances, ordering keys, host units never fabricated, `-m` modules, continuation lines, comments, unreadable files); 14 `Exec=` parsing cases; deployment-path resolution including the POSIX-on-Windows case; and corpus-level assertions that the edges land on the real script/doc nodes, that the timer/service pair links correctly with no self-loop, that nothing dangles, and that the stamp does not leak.

With the wiring reverted and only the module kept, 10 of them fail. `test_extractors_registry`, `test_detect`, `test_dotnet` and the disambiguation suites are unchanged (304 passed); the full suite matches the `v8` baseline.

README gains a row in the file-types table.


abhay-codes07:
Follow-up commit on the review's advisory points: an `Accept=yes` socket now activates `<stem>@.service`, a quoted `Exec=` path with a space stays one token (shlex), and the `@` prefix skips argv[0] whatever order it appears in among the other prefixes. Three new cases in the test file; 45 pass.

PR #3089, opened 2026-08-25
feat(extract): ingest AsciiDoc documentation (#2938)

### LINKED ISSUE #2938: [Feature] Detect and digest asciidoc / adoc documentation
All of our project documentation lives in asciidoc files (`.adoc`) which graphify will ignore: 

> Your .adoc documentation is not in the graph. 11 .adoc files exist under docs/adoc/ — only their 6 PNGs were ingested. graphify's detector doesn't recognise .adoc.

It would be great if asciidoc files could be parsed and integrated into the graph just like markdown files.

abhay-codes07:
Closes #2938.

## The problem

`.adoc` was in no extension set, so a project whose documentation is AsciiDoc had none of it in the graph — the detector reported that only the PNGs beside the docs were ingested. `_PROSE_EXTS` already knew `.adoc` for word counting, which is how far it got.

## The change

`.adoc` / `.asciidoc` join every place a document is decided:

- `DOC_EXTENSIONS` — classification, and so the semantic pass digests them like markdown;
- `file_slice._SPLITTABLE_TEXT_SUFFIXES` — an oversized manual is sliced, not truncated (#2900's contract);
- the callflow file-type set and the hook-guard extension list;
- markdown's `_MD_LINKABLE_EXTS` — a `[text](./guide.adoc)` link now resolves.

And a structural pass, `extractors/asciidoc.py`, so the deterministic layer is on par with markdown's:

- the file is a `page` node carrying the document title (`= Title`) and the header attributes (`:author:` …) as `frontmatter`, sanitized like markdown's;
- section titles (`==`, `===`, …) are `heading` nodes nested by level, `file → contains → heading`, `heading → contains → subheading`;
- `include::other.adoc[]`, `xref:other.adoc[...]` / `xref:other#anchor[]` (an extension-less xref names a sibling `.adoc`), `link:<local path>[...]` and the cross-document `<<other.adoc#anchor,text>>` form become `references` edges. The target id is minted from the resolved path — the same recipe as the target's own file node — so it merges into that node rather than spawning a ghost, and an existing target is stamped `target_file` for the incremental remap, exactly as markdown does (#2211).

Skipped rather than guessed: external URLs, in-page anchors, non-document targets (`diagram.png`), and paths carrying an unresolved AsciiDoc attribute (`include::{snippets}/gen.adoc[]`). Delimited blocks (`----`, `....`, `////`, `++++`, `====`, `****`, `____`) and `//` comment lines are not parsed for headings or links, mirroring markdown's fenced-block handling.

## What it looks like

```
guide.adoc (page: title="Operations Guide", frontmatter={author: "Ops Team"})
  ├─ contains → Deploy
  │    └─ contains → Rollback
  ├─ contains → Monitor
  ├─ references → runbook.adoc      (xref:runbook.adoc[...] and xref:runbook#steps[] — one edge)
  ├─ references → alerts.adoc       (include::parts/alerts.adoc[] and <<parts/alerts.adoc#p1,…>>)
  └─ references → checklist.md      (link:checklist.md[...])
checklist.md ─ references → guide.adoc   (markdown → asciidoc now resolves)
```

## Tests

`tests/test_asciidoc.py` — 21 tests: both extensions present in every set that decides a document (classification, dispatch, slicing, markdown linkability) and listed by `detect()`; the page node's title and attributes; section nesting; block and comment skipping; duplicate section titles; a title-less fragment; unreadable files; each link form producing one edge per target; the skip rules; the `target_file` stamp; and a corpus-level check that every reference — in both directions between markdown and AsciiDoc — merges into the linked document's real node with nothing dangling and no stamp leaking.

With the registrations reverted and only the extractor module kept, the 4 integration tests fail (the extractor's own unit tests rightly keep passing). `test_detect`, `test_oversized_document_slicing`, `test_extractors_registry` and the markdown/wikilink suites are unchanged (290 passed); the full suite matches the `v8` baseline.

README's file-types table is updated.

### DIFF
--- a/README.md
+++ b/README.md
@@ -347,7 +347,7 @@ To remove graphify from all platforms at once: `graphify uninstall` (add `--purg
 | Robot Framework | `.robot .resource` (via the official `robot.api` parser, requires `uv tool install graphifyy[robot]`; suites, test cases, user keywords, keyword-call and Resource/Library/Variables import edges) |
 | MCP configs | `.mcp.json` `mcp.json` `mcp_servers.json` `claude_desktop_config.json` — extracts server nodes, package refs, env var requirements |
 | Package manifests | `apm.yml` `pyproject.toml` `go.mod` `pom.xml` — one canonical package node per package (by name) plus `depends_on` edges, so a package referenced from many manifests is a single hub |
-| Docs | `.md .mdx .qmd .html .txt .rst .yaml .yml` (markdown `[text](./other.md)` links and `[[wikilinks]]` become `references` edges between docs) |
+| Docs | `.md .mdx .qmd .adoc .asciidoc .html .txt .rst .yaml .yml` (markdown `[text](./other.md)` links and `[[wikilinks]]`, and AsciiDoc `include::`/`xref:`/`link:` targets, become `references` edges between docs; AsciiDoc section titles become heading nodes like markdown headings) |
 | Office | `.docx .xlsx` (requires `uv tool install graphifyy[office]`) |
 | Google Workspace | `.gdoc .gsheet .gslides` (opt-in; requires `gws` auth and `--google-workspace`; Sheets need `uv tool install graphifyy[google]`) |
 | PDFs | `.pdf` |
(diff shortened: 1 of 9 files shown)


abhay-codes07:
Follow-up commit on the review's advisory points: a delimited block now closes only on a delimiter of the same character *and* length (so a nested `======` example inside `====` does not end the outer block early), and `.asciidoc` joins `.adoc` in the hook-guard source list. Two new tests; 23 pass.

PR #3096, opened 2026-08-25
perf(detect): stop re-deriving anchor-relative paths per ignore pattern

Mohammad-Palla:
## What

`detect()` was dominating build wall-clock on large repos (63% of total on
a ~19,700-file corpus). Profiling showed the cost was not the file I/O it
looks like: 97% of `detect()`'s time was inside `_is_ignored()`, which
tested every file against every accumulated gitignore pattern via an
exception-based `Path.relative_to(anchor)` call. On a repo whose nested
`.gitignore` files contribute ~1,000 patterns, that's millions of pathlib
parse operations per scan, the overwhelming majority of which raise
`ValueError` (the anchor isn't an ancestor) after doing the full parse.

- Cache the anchor-relative path per `(target, anchor)` instead of
  recomputing it once per `(target, pattern-line)`.
- Where provably equivalent, derive the relative path by string slice
  instead of `Path.relative_to()` — restricted to POSIX with an absolute
  anchor, since `relative_to()` is case-insensitive on Windows and a naive
  prefix check would silently stop applying `.gitignore` there.
- Replace `_match_anchored_ignore_pattern`'s per-call `functools.lru_cache`
  with a plain dict memo (the cache is rebuilt every call anyway).

An earlier version of this change also parallelized `detect()`'s
classification loop with a `ThreadPoolExecutor`, on the assumption that the
per-file work was I/O-bound. It measured as a no-op once the gitignore fix
landed, and as a net pessimization on another corpus (50-75% slower, 2x
peak RSS), while introducing real bugs (Ctrl-C no longer aborting the scan,
an `atexit` `RuntimeError` that discarded the word-count cache, and a
blanket `except Exception` that silently pruned nodes from `graph.json`).
It was reverted; this PR contains only the gitignore fix.

## Measured

`graphify <repo> --code-only --timing`, golden cache snapshot restored
before each run, 2 runs per side:

```
              baseline          fixed
  detect:   168.6s / 194.4s   59.8s / 59.5s    3.0x
  total:    271.3s / 307.4s  174.4s / 177.6s   1.6x
```

## Correctness

No behavior change. `detect()`'s full return value is byte-identical to
the previous implementation on the corpus above and a synthetic tree
covering negations, nested `.gitignore`, bare `*`, `**`, anchored
patterns, char classes, `$GIT_DIR/info/exclude`, NFC/NFD filenames,
symlinks in/out of root, a FIFO, and an unreadable file — across six
`detect()` configurations and `detect_incremental()` across all manifest
schemas. End-to-end `graph.json` diff is identical in every node and link
attribute; only Leiden community labels differ, which two runs of the
unmodified baseline also do. Windows path semantics verified via simulated
`PureWindowsPath`/`ntpath`: 0 divergences from the original implementation.

Full test suite: 4,686 passed, 232 skipped (same 12 pre-existing failures
before and after, unrelated to this change — missing `tree_sitter_hcl`
grammar, ollama retry tests).

### DIFF
--- a/graphify/detect.py
+++ b/graphify/detect.py
@@ -11,7 +11,6 @@ import time
 import unicodedata
 from concurrent.futures import ThreadPoolExecutor
 from enum import Enum
-from functools import lru_cache
 from pathlib import Path
 from typing import Callable
 
@@ -32,6 +31,12 @@ class FileType(str, Enum):
 
 
 _MANIFEST_PATH = str(out_path("manifest.json"))
+
+#: Whether _is_ignored may use string-prefix path arithmetic instead of
+#: Path.relative_to(). POSIX only: on Windows relative_to() is case-insensitive
+#: and normalizes separators/drive-relative paths, so a string prefix is not an
+#: equivalent substitute there (see _relative_str).
+_FAST_RELATIVE_PATHS = os.name == "posix"
 
 #: Window in which a manifest row's own timestamp is too close to the file's
 #: mtime for "mtime unchanged" to prove the content is unchanged. Coarse for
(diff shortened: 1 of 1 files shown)


Mohammad-Palla:
Closing as superseded by #3074 and the follow-up #2226 fix (258ab11, d8abe52), already merged into v8 — same root cause (per-pattern-per-file `Path.relative_to()` in `_is_ignored`), independently diagnosed and fixed here, but their landed version is a strict superset: correct Windows case-folding via inline `os.path.normcase` (we fell back to `relative_to()` on Windows instead), a fix for a reference-cycle leak in the recursive matcher closure that this PR didn't catch, additional per-target caching (segment/prefix lists, `is_dir()`, NFC name), and a bounded process-wide pattern-parse cache for long-lived `watch` processes. Verified by rebasing onto current `v8` and comparing implementations directly — nothing here adds value on top of what's already landed. Thanks for beating us to it!

Issue #2817, opened 2026-08-17
cluster() community-splitting pass is nondeterministic across PYTHONHASHSEED (_partition itself is not; edge-count loss appears fixed since 0.9.6)


AlexisDavid1989:
## Localised: `_partition` is deterministic, `cluster()` post-processing is not

Same graph object, graphifyy 0.9.42, 4 unpinned runs + 2 with `PYTHONHASHSEED=0`:

```
_partition(G)   61 communities, identical partition hash in all 6 runs
cluster(G)      345 / 342 / 346 / 350 unpinned
                344 / 344 with PYTHONHASHSEED=0
```

`_partition`'s sorted node/edge construction and hardcoded `seed=42` work
correctly - it never varies, pinned or not. The variance is introduced
downstream, in `cluster()`'s community-splitting pass
(`_MAX_COMMUNITY_FRACTION` / `_MIN_SPLIT_SIZE` / `_COHESION_SPLIT_THRESHOLD` /
`_COHESION_SPLIT_MIN_SIZE`), which expands 61 communities to ~345 - about 85%
of the final structure - without inheriting `_partition`'s ordering
guarantees.

**Caveat on the numbers above:** measured on a synthetic 4,000-node /
9,000-edge fixture, on Linux / Python 3.11 - not the ORE corpus this issue
was originally opened against.

Re-ran the same `_partition`-vs-`cluster` comparison on our frozen ORE
extraction to check it holds there too (graphifyy 0.9.42, Windows, Python
3.12.4, same `.graphify_extract.json` as every measurement in this issue), 4
unpinned runs + 2 pinned:

```
_partition(G)   83 communities, identical hash all 6 runs (eaaaf43442be)
cluster(G)      230 / 230 / 231 / 229 unpinned
                231 / 231 with PYTHONHASHSEED=0
```

Same shape on a completely different, real-world input: `_partition` never
moves, `cluster()` does, `PYTHONHASHSEED=0` fixes it. This localises the bug
inside `cluster()`'s splitting/re-splitting passes rather than in
`_partition`, `build_from_json`, or Leiden/Louvain itself.

**Suggested direction:** apply the same normalisation `_partition` already
does - iterate communities and their members in a sorted, deterministic
order when deciding what to split and how, in `_split_community` and the
cohesion-based second pass in `cluster()`.

**Correction to the hypothesis in the previous update:** the `json.dumps`
list-ordering theory is refuted - `G.edges()` yield order and edge
[...]

Issue #1667, opened 2026-07-04
Non-deterministic community assignments across identical-corpus runs (0.9.6)


krishnateja7:
Second of the two 0.9.6 follow-ups (see the zero-nodes-in-batch issue filed alongside — they may share a parallel-processing root cause).

## Symptom

Two consecutive `graphify update .` runs on an **unchanged corpus** produce graph.json files that differ only in `community` assignments:

- node-id set: identical
- link multiset (all fields): identical
- `community`: differs on **196 of 11,819 nodes**

So the semantic graph is deterministic, but the artifact is not byte-reproducible.

It appears intermittent: one other consecutive pair of runs produced byte-identical output. That pattern is consistent with an unseeded RNG or parallel-ordering dependence in the community-detection step (Louvain/Leiden-style methods are tie-break sensitive).

On 0.9.5 we never observed this — we actually used file-hash equality of double runs as an acceptance test for our own append pipeline, and it held.

## Ask

A fixed seed and/or deterministic tie-breaking in community detection, so identical input produces byte-identical graph.json. Reproducible output is useful for caching, CI artifact diffing, and downstream tools that append to the graph (our case).

## Environment

- graphifyy 0.9.6 (uv tool install), macOS (Darwin 25.5)
- Rails repo: ~11,800 nodes / ~19,800 links
- Diff method: parse both files, compare node-id sets, per-node field diff, link multiset comparison

safishamsi:
Looked into this closely. The clustering path already has the determinism guards you would want, and I could not reproduce the nondeterminism with synthetic graphs, so I do not want to ship a speculative fix that might not address your case. Findings:

- Community detection runs on a rebuilt `stable` graph with **sorted** node and edge insertion, and the partitioner is seeded (graspologic leiden `random_seed=42`, `trials=1`; verified the installed 3.4.4 signature accepts and receives both). The final community IDs are assigned by a **total order** (size desc, then `tuple(sorted(node_ids))`), so an identical grouping always maps to identical IDs.
- On a synthetic 600-node/1700-edge graph I ran `cluster()` twice in-process and across two processes with different `PYTHONHASHSEED`: **zero** differing community assignments both ways. So the Python-side path is deterministic here.

[...]

PR #3104, opened 2026-08-26
perf(cluster): call graspologic_native.leiden() directly, skip graspologic's import chain

Mohammad-Palla:
## Summary

- `_partition()` currently reaches `graspologic.partition.leiden()` through the full `graspologic` package import, which eagerly pulls in `graspologic.layouts` → `umap` → `pynndescent` → numba JIT compilation — 7-19s per process, for a layout module the clustering path never calls.
- `graspologic_native` (the PyO3/Rust binding `graspologic.partition.leiden()` itself wraps) imports in ~0.5ms. This adds a `_native_leiden()` adapter that calls it directly, falling back to the existing `graspologic.partition.leiden` path (then NetworkX Louvain) if `graspologic_native` isn't installed or the graph shape isn't supported.
- Graphify's node IDs are guaranteed-unique strings by construction (`extractors/resolution.py`'s `_disambiguate_colliding_node_ids` runs before the graph is built), so the string-collision case `graspologic`'s `_IdentityMapper` guards against can't occur here; the adapter keeps a defensive check anyway.

## Measured (Bun corpus, 161k nodes / 336k edges, golden-cache-restored before each run)

| | before | after |
|---|---:|---:|
| `cluster` stage | 27.3-28.6s | 17.9s (~36-37% faster) |
| total build | 155.4-159.8s | 146.7s (~6-8% faster) |

## Correctness verified, not assumed

- Feeding the identical graph to both code paths in the same process produces a byte-identical partition (exact canonical-form match, Adjusted Rand Index = 1.0).
- Community counts, modularity, and singleton counts track within <0.3% across runs.
- Full test suite: 4,735 passed, 233 skipped, 12 pre-existing unrelated failures (same set as before this change). `tests/test_cluster.py`: 11/11.
- Adversarially tested degenerate inputs (empty graph, single node, resolution ≤ 0, self-loops, negative/zero/NaN edge weights, disconnected components) against both code paths side by side — all diverging behavior traced back to `graspologic_native` itself being called identically by both, not something this change introduces. Confirmed separately that Graphify's own extraction pipeline never produces a negative/zero/NaN edge weight (every extractor hardcodes `weight: 1.0`), so that pre-existing panic risk isn't reachable through normal use.

## Note on run-to-run variance

Separate full-pipeline invocations show community-label drift between runs (Adjusted Rand Index ≈ 0.73-0.75), but this is pre-existing and identical on both code paths — traced to the already-filed #2817 / #1667 (hash-seed-sensitive iteration in `cluster()`'s post-processing splitting pass, downstream of `_partition()`), not introduced or worsened by this change.

## Test plan

- [x] `pytest tests/test_cluster.py` — 11/11 passed
- [x] Full suite — no new failures vs. baseline
- [x] Same-input, same-process partition equivalence check (adapter vs. wrapper)
- [x] End-to-end timing on Bun corpus, golden-cache-restored, matched load
- [x] Adversarial edge-case testing (degenerate graphs, invalid resolution, bad weights)

### DIFF
--- a/graphify/cluster.py
+++ b/graphify/cluster.py
@@ -19,11 +19,86 @@ def _suppress_output():
     return contextlib.redirect_stdout(io.StringIO())
 
 
+def _native_leiden(stable: nx.Graph, resolution: float) -> dict[str, int] | None:
+    """Call graspologic_native.leiden() directly, bypassing graspologic's own
+    package import.
+
+    graspologic.partition.leiden() is a thin wrapper around exactly this
+    native (Rust) call. Importing the *package* — as opposed to the native
+    extension module it depends on — pulls in graspologic.layouts, which
+    imports umap, which imports pynndescent, which numba-JIT-compiles at
+    import time for a layout algorithm this function never calls: measured
+    at 7-19s of one-time import cost against a ~1s native call and a ~1.4s
+    full round trip (conversion + call + map-back) — see the "third update"
+    in GRAPHIFY_BUILD_PERF.md for the measurements this is based on.
+
+    Returns None (the caller falls through to the graspologic.partition.leiden
+    path, then to the networkx Louvain fallback) if graspologic_native isn't
+    installed, or if `stable` isn't the plain undirected, non-multigraph
+    input leiden actually supports — the same shape check
+    graspologic.partition.leiden itself makes before calling the same native
+    function.
+    """
+    try:
+        import graspologic_native as gn
+    except ImportError:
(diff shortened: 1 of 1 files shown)


safishamsi:
Shipped in v0.9.51 via authorship-preserving cherry-pick so you keep contributor-graph credit. Thanks @Mohammad-Palla! Release: https://github.com/Graphify-Labs/graphify/releases/tag/v0.9.51

PR #3107, opened 2026-08-26
fix(cache): the atexit stat-index flush must not resurrect a deleted corpus (#2974)

### LINKED ISSUE #2974: atexit stat-index flush recreates the analyzed directory if it was deleted mid-run (worktree race leaves husk dirs)
## Summary

`_flush_stat_index` in `graphify/cache.py` runs at `atexit` and does `p.parent.mkdir(parents=True, exist_ok=True)` before writing `graphify-out/cache/stat-index.json`. If the analyzed directory is **deleted while graphify is running**, the exit-time flush recreates the deleted directory tree, leaving a "husk" containing nothing but `graphify-out/cache/stat-index.json`.

Observed on `graphifyy` 0.9.17 (macOS, Python 3.13).

## Real-world trigger

A git `post-commit` hook that runs `graphify update . >/dev/null 2>&1 &` (backgrounded), in a workflow that uses short-lived `git worktree` checkouts:

1. Agent/CI commits in a worktree → hook starts `graphify update .` in the background.
2. The branch is merged and `git worktree remove` deletes the worktree.
3. graphify finishes, its `atexit` flush `mkdir -p`s the dead path back into existence.

Over a few weeks this left **81** resurrected `<repo>-<branch>` directories next to the main repo, each containing only the stat-index file.

## Deterministic repro

```python
import shutil
from pathlib import Path
from graphify import cache

corpus = Path("husk-race-corpus").resolve()
corpus.mkdir()
f = corpus / "a.md"
f.write_text("# hello\nbody\n")

cache.file_hash(f, corpus)      # dirties the stat index
shutil.rmtree(corpus)           # corpus deleted mid-run
cache._flush_stat_index()       # what atexit will do

print("husk resurrected!" if corpus.exists() else "ok")
# prints: husk resurrected!  -> corpus/graphify-out/cache/stat-index.json exists
```

## Suggested fix

Skip the flush when the corpus root no longer exists — a missing root means the cache has no home anymore, and recreating the directory is worse than dropping one cache write:

```python
def _flush_stat_index() -> None:
    global _stat_index_dirty, _stat_index_root
    if not _stat_index_dirty or _stat_index_root is None:
        return
    if not _stat_index_root.exists():   # corpus deleted while we ran
        _stat_index_dirty = False
        return
    p = _stat_index_file(_stat_index_root)
    ...
```

Verified locally: with the guard, the repro above leaves the deleted directory deleted, and a live corpus still gets its stat-index written as before. Happy to open a PR with this change if useful.

abhay-codes07:
Closes #2974.

## The problem

`_flush_stat_index` runs at `atexit` and did `p.parent.mkdir(parents=True, exist_ok=True)` before writing `graphify-out/cache/stat-index.json`. If the analyzed directory is deleted while graphify is running, the exit-time flush recreates the whole path and leaves a husk containing nothing but the index.

The trigger is ordinary: a post-commit hook runs `graphify update . >/dev/null 2>&1 &` in a short-lived `git worktree`; the branch merges and `git worktree remove` deletes the tree while the rebuild is still going; the rebuild finishes and resurrects the dead path. The reporter found 81 such `<repo>-<branch>` directories over a few weeks.

## The change

The index is a pure optimisation, so when its root no longer exists it is simply not written — and the dirty flag is cleared so nothing retries. A root that still exists keeps the old behaviour exactly: a first run writes the index before `graphify-out/` exists at all, and three existing `test_stat_index_portability` tests pin that, so the check is deliberately on the root, not on the output directory.

The issue's own deterministic repro (`file_hash` → `rmtree` → `_flush_stat_index()`) now prints "ok".

## Tests

`tests/test_stat_index_husk.py` — 3 tests: a corpus deleted mid-run stays deleted (and nothing is left pending), a redirected `cache_root` that vanished is not recreated, and a live run still writes the index into a not-yet-existing `graphify-out/`. With the fix reverted, 2 of 3 fail. `test_stat_index_portability`, `test_cache`, `test_word_count_cache`, `test_extract_cache_location`, `test_cli_export` and `test_incremental` are unchanged; the full suite matches the `v8` baseline.

### DIFF
--- a/graphify/cache.py
+++ b/graphify/cache.py
@@ -379,6 +379,20 @@ def _flush_stat_index() -> None:
             continue
         dk = _stat_key_to_relative(k, _stat_index_anchor) if _stat_index_anchor is not None else k
         on_disk[dk] = v
+    # Never resurrect a corpus that was deleted while graphify was running
+    # (#2974): a hook-launched `graphify update . &` in a short-lived worktree
+    # outlives `git worktree remove`, and an unconditional `mkdir -p` here
+    # rebuilt the dead path as a husk holding nothing but this index. The
+    # index is a pure optimisation, so when its root is gone it is simply not
+    # written. Creating graphify-out/cache/ under a root that still exists is
+    # unchanged (a first run writes the index before anything else does).
+    try:
+        if not _stat_index_root.is_dir():
+            _stat_index_dirty = False
+            return
+    except OSError:
+        _stat_index_dirty = False
+        return
     try:
         p.parent.mkdir(parents=True, exist_ok=True)
         fd, tmp = tempfile.mkstemp(dir=p.parent, prefix="stat-index.", suffix=".tmp")
--- a/tests/test_stat_index_husk.py
+++ b/tests/test_stat_index_husk.py
@@ -1,0 +1,69 @@
+"""The atexit stat-index flush must not resurrect a deleted directory (#2974).
+
+A post-commit hook runs `graphify update . &` in a short-lived worktree; the
+branch merges and `git worktree remove` deletes the tree while the rebuild
+is still running. The exit-time flush then `mkdir -p`'d the dead path back
+into existence, leaving a husk holding nothing but
+`graphify-out/cache/stat-index.json` — 81 of them over a few weeks.
+"""
+from __future__ import annotations
+
+import shutil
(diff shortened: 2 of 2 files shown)


safishamsi:

PR #3108, opened 2026-08-26
fix(hooks): a graphify skip must not terminate the whole git hook (#2986)

### LINKED ISSUE #2986: post-commit hook: every `exit 0` in the generated block terminates the whole hook — silently kills anything appended after it (incl. repo-local hook chaining) on root commits, rebases and worktrees
## Summary

The generated `post-commit` hook is a flat `sh` script whose skip conditions use bare `exit 0`. In a flat script that terminates the **entire hook process**, not just graphify's section. Any block appended after graphify's markers — including hook-chaining logic that other tools install — is silently skipped whenever graphify decides to skip itself.

This is not an edge case: one of those exits fires on **every root commit** of every repository.

## Affected

Confirmed by reading `graphify/hooks.py` in the published wheels for **0.9.45** (16 `exit 0`) and **0.9.48** (17). Installed and observed with 0.9.39 and 0.9.45. The count went up rather than down between those releases, so this looks unnoticed rather than known.

## Mechanism

`hooks.py` emits, among others:

```sh
[ -d "$GIT_DIR/rebase-merge" ] && exit 0
[ -d "$GIT_DIR/rebase-apply" ] && exit 0
[ -f "$GIT_DIR/MERGE_HEAD" ] && exit 0
[ -f "$GIT_DIR/CHERRY_PICK_HEAD" ] && exit 0
[ "${GRAPHIFY_SKIP_HOOK:-0}" = "1" ] && exit 0

CHANGED=$(git diff --name-only HEAD~1 HEAD 2>/dev/null || git diff --name-only HEAD 2>/dev/null)
if [ -z "$CHANGED" ]; then
    exit 0
fi

_NON_GRAPH=$(echo "$CHANGED" | grep -v '^graphify-out/' || true)
if [ -z "$_NON_GRAPH" ]; then
    exit 0
fi
```

Each is correct *as an intent* ("graphify should not run here") but wrong *as a mechanism* in a shared hook file, because it ends the process.

The root-commit case is the most surprising one. On a repository's first commit `HEAD~1` does not resolve, so the fallback `git diff --name-only HEAD` compares the working tree against `HEAD` — clean immediately after a commit — yielding an empty `CHANGED` and an `exit 0`.

## Reproduction

```sh
# global hook: graphify's block, followed by anything else
printf '#!/bin/sh\n' > ~/.config/git/hooks/post-commit
graphify hook install                      # appends the graphify block
printf 'printf "[after-graphify RAN]\\n"\n' >> ~/.config/git/hooks/post-commit

mkdir repro && cd repro && git init -q
printf 'x\n' > f.txt && git add -A
git commit -m "root"        # -> "[after-graphify RAN]" does NOT print

printf 'y\n' >> f.txt && git add -A
git commit -m "second"      # -> it prints
```

The same asymmetry appears during rebase, cherry-pick, in linked worktrees, with `GRAPHIFY_SKIP_HOOK=1`, and on commits touching only `graphify-out/`.

## Impact

`hooks.py` contains no chaining logic of its own, so it implicitly assumes it owns the hook file or runs last. In practice it is commonly appended alongside other tools' blocks. Where another tool's block chains to repository-local hooks, those repo-local `post-commit` hooks silently stop running in all the situations above — no error, no output, just nothing.

## Suggested fix

Wrap the emitted block in a subshell so the skip conditions become local:

```sh
(
  ... existing generated block ...
) || true
```

`exit 0` then leaves the subshell, graphify skips itself exactly as intended, and the hook continues. Verified locally against 0.9.45: root commits, `graphify-out/`-only commits, and normal code commits all now continue past the block, while graphify's own rebuild behaviour is unchanged (`GRAPH_REPORT.md` still produced on code commits).

An alternative that avoids the subshell is replacing each `exit 0` with a guard variable checked before the rebuild step, but the subshell wrap is a one-line change and cannot miss an exit path.

Happy to open a PR if the subshell approach looks right.

abhay-codes07:
Closes #2986.

## The problem

The generated `post-commit` / `post-checkout` block is appended to whatever hook a repo already has, and other tools chain their own logic after graphify's end marker. Every skip condition in the block was a bare `exit 0` — seventeen of them in 0.9.48 — which in a flat `sh` script ends the **entire hook process**, not graphify's section. Anything after the marker was silently dropped:

- on every **root commit** (`HEAD~1` does not exist, so `CHANGED` is empty);
- during every rebase / merge / cherry-pick;
- in every linked worktree;
- whenever `GRAPHIFY_SKIP_HOOK=1`;
- whenever no graphify-capable python could be found.

Each `exit 0` is right as an *intent* ("graphify should not run here") and wrong as a *mechanism* in a shared hook file.

## The change

Both hook bodies now run inside a subshell:

```sh
# graphify-hook-start
# ...
(
  ... unchanged body, exits and all ...
)
# graphify-hook-end
```

An `exit` inside the subshell ends only graphify's section. This keeps the seventeen skip sites byte-for-byte as they were, avoids rewriting them to `return` (which would break the `_PYTHON_DETECT` fragment tests run standalone — `return` outside a function is undefined in POSIX sh), leaves the detached rebuild launch unaffected, and preserves the hook's own exit status of 0. The markers stay outside the subshell so reinstall / uninstall / `hook status` keep finding the block, and the in-place update path still reports "already installed" for an unchanged block.

One existing test slices the checkout script's prefix and runs it under `sh -c`; it now closes the subshell after its sentinel.

## Tests

`tests/test_hook_chain_survives_skip.py` — 7 tests that install the real hooks into a real git repo, append `echo AFTER_GRAPHIFY` after the block, and run them under `sh`: a `GRAPHIFY_SKIP_HOOK=1` skip, a rebase in progress, an empty diff (the root-commit shape), a non-branch checkout, a skipped checkout hook, a pre-existing hook before the block, and reinstall idempotence. Each asserts the trailer ran, the exit status is 0, and (for the skip case) that no rebuild was launched. With the subshell reverted, 5 of 7 fail. `tests/test_hooks.py` is otherwise unchanged; the full suite matches the `v8` baseline.

### DIFF
--- a/graphify/hooks.py
+++ b/graphify/hooks.py
@@ -304,10 +304,20 @@ fi
 """
 
 
+# Both hook bodies run inside a subshell `( ... )`. The generated block is
+# appended to whatever post-commit / post-checkout already exists, and other
+# tools chain their own logic after it; every skip condition in the block is a
+# bare `exit 0`, which in a flat script ends the WHOLE hook, silently dropping
+# anything after graphify's end marker - on every root commit (HEAD~1 does
+# not exist), every rebase/merge, every linked worktree, every
+# GRAPHIFY_SKIP_HOOK=1 (#2986). Inside the subshell an `exit` ends only
+# graphify's section; the detached rebuild launch is unaffected, and the
+# hook's own exit status stays 0 as before.
 _HOOK_SCRIPT = """\
 # graphify-hook-start
 # Auto-rebuilds the knowledge graph after each commit (code files only, no LLM needed).
 # Installed by: graphify hook install
+(
 
 # Deterministic clustering: networkx louvain iterates string-keyed sets whose
 # order is randomized per-process by PYTHONHASHSEED, so community assignments
@@ -355,7 +365,8 @@ _GRAPHIFY_LOG="${HOME}/.cache/graphify-rebuild.log"
 mkdir -p "$(dirname "$_GRAPHIFY_LOG")"
 export GRAPHIFY_REBUILD_LOG="$_GRAPHIFY_LOG"
 echo "[graphify hook] launching background rebuild (log: $_GRAPHIFY_LOG)"
-""" + _detached_launch(_REBUILD_BODY_COMMIT) + """# graphify-hook-end
+""" + _detached_launch(_REBUILD_BODY_COMMIT) + """)
+# graphify-hook-end
 """
 
 
@@ -363,6 +374,7 @@ _CHECKOUT_SCRIPT = """\
 # graphify-checkout-hook-start
 # Auto-rebuilds the knowledge graph (code only) when switching branches.
 # Installed by: graphify hook install
+(
 
 # Deterministic clustering: networkx louvain iterates string-keyed sets whose
 # order is randomized per-process by PYTHONHASHSEED, so community assignments
@@ -412,7 +424,8 @@ _GRAPHIFY_LOG="${HOME}/.cache/graphify-rebuild.log"
 mkdir -p "$(dirname "$_GRAPHIFY_LOG")"
 export GRAPHIFY_REBUILD_LOG="$_GRAPHIFY_LOG"
 echo "[graphify] Branch switched - launching background rebuild (log: $_GRAPHIFY_LOG)"
-""" + _detached_launch(_REBUILD_BODY_CHECKOUT) + """# graphify-checkout-hook-end
+""" + _detached_launch(_REBUILD_BODY_CHECKOUT) + """)
+# graphify-checkout-hook-end
 """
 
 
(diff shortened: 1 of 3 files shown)


jecanore:
Also merged this in locally (v8 + #2064 + #2858 + this) — the exit-0-kills-the-chain bug is exactly the kind of thing that fails silently for months, glad someone caught it. Applied clean.

safishamsi:
Shipped in **v0.9.56** — landed on `v8` via cherry-pick with your authorship preserved. Thanks @abhay-codes07 — a git-hook skip no longer terminates the whole hook — it runs in a subshell so chained hooks still execute. Release: https://github.com/Graphify-Labs/graphify/releases/tag/v0.9.56

PR #3109, opened 2026-08-26
fix(hooks): stop probing a /bin/sh launcher wrapper as if it were python (#3027)

### LINKED ISSUE #3027: post-commit hook runs ImageMagick's `import` when graphify launcher is a /bin/sh wrapper
## Summary

The `post-commit` hook installed by `graphify hook install` assumes the `graphify` launcher's shebang points at a **Python interpreter**. When `graphify` is instead a `/bin/sh` wrapper script (as pipx generates in some setups), the interpreter-detection logic runs `/bin/sh -c "import graphify"`. Since `import` is not a shell builtin, it resolves on `PATH` to **ImageMagick's `import`** screenshot tool, which dumps its full usage text into the terminal on every commit.

## Environment

- macOS (darwin), zsh
- graphify installed via pipx (package `graphifyy` 0.4.12)
- `command -v graphify` → a `/bin/sh` wrapper (shebang `#!/bin/sh`, not a python path)
- ImageMagick present at `/opt/homebrew/bin/import`

## Root cause

In `.git/hooks/post-commit`:

```sh
_SHEBANG=$(head -1 "$GRAPHIFY_BIN" | sed 's/^#![[:space:]]*//')   # -> /bin/sh
case "$_SHEBANG" in
    */env\ *) GRAPHIFY_PYTHON="${_SHEBANG#*/env }" ;;
    *)        GRAPHIFY_PYTHON="$_SHEBANG" ;;                        # -> /bin/sh
esac
# allowlist passes (/bin/sh is all valid path chars)
if [ -n "$GRAPHIFY_PYTHON" ] && ! "$GRAPHIFY_PYTHON" -c "import graphify" 2>/dev/null; then
    GRAPHIFY_PYTHON=""
fi
```

`/bin/sh -c "import graphify"` executes ImageMagick's `import` with arg `graphify`. Only stderr is suppressed (`2>/dev/null`), so the usage text still floods **stdout**.

The check *does* fail (non-zero exit) and falls back to `python3`, so the graph still rebuilds — the bug is purely spurious, noisy output on every commit.

## Reproduce

```sh
sh .git/hooks/post-commit   # prints "Version: ImageMagick ... Usage: import ..."
```

## Suggested fix

Only trust a shebang that actually names python, and suppress stdout on the probe too:

```sh
case "$GRAPHIFY_PYTHON" in *python*) : ;; *) GRAPHIFY_PYTHON="" ;; esac
if [ -n "$GRAPHIFY_PYTHON" ] && ! "$GRAPHIFY_PYTHON" -c "import graphify" >/dev/null 2>&1; then
    GRAPHIFY_PYTHON=""
fi
```

This falls straight through to the existing `python3`/`python` fallback and produces zero output. Verified locally.

abhay-codes07:
Closes #3027.

## The problem

pipx generates the `graphify` launcher via distlib's "exec trick" when the install path contains a space — macOS's `~/Library/Application Support/pipx` is the common case. That launcher's shebang is `#!/bin/sh`, and its second line execs the real interpreter:

```sh
#!/bin/sh
'''exec' "/Users/me/Library/Application Support/pipx/venvs/graphifyy/bin/python" "$0" "$@"
' '''
```

The hook parsed the shebang into `GRAPHIFY_PYTHON=/bin/sh` and ran `/bin/sh -c "import importlib.util, sys; ..."`. `import` is not a shell builtin, so it resolved on `PATH` to ImageMagick's `import` screenshot tool, which dumped its full usage text on every commit — only stderr was silenced. The probe did fail and the chain fell back to `python3`, so the graph still rebuilt; the bug was the noise, and the wasted opportunity to find the right interpreter.

## The change

In the shebang branch of `_PYTHON_DETECT`:

- if the shebang names something that is not a python, take the interpreter from the wrapper's exec line instead (`sed` on the second line of the 256-byte head the script already reads);
- never probe an interpreter whose basename is not `python*` / `pypy*` — so a shell is never handed `-c "import ..."` again, whatever the wrapper looks like.

The path recovered from a wrapper contains a space by construction, and the injection allowlist rejected spaces. A space is not a shell metacharacter here — `$GRAPHIFY_PYTHON` is always expanded quoted — so the allowlist now matches against the path with spaces folded to `_`, leaving the proven pattern itself byte-for-byte unchanged (the `#2126`/`#2641` allowlist tests, which extract that exact pattern from the emitted text, still pass).

## Tests

`tests/test_hook_sh_wrapper_launcher.py` — 5 tests that run the real emitted `_PYTHON_DETECT` under `sh` on a machine whose `PATH` holds a `/bin/sh` wrapper, an ImageMagick-style `import` that prints usage, and ambient pythons that cannot import graphify: the wrapper's interpreter (in a directory with a space) is adopted and `import` is never run; a wrapper with no exec line falls through quietly; a wrapper whose interpreter fails the probe is not adopted; a plain python shebang still resolves; and the emitted script gates the probe on the interpreter's name. With the fix reverted, 3 of 5 fail. `tests/test_hooks.py` is unchanged; the full suite matches the `v8` baseline.

### DIFF
--- a/graphify/hooks.py
+++ b/graphify/hooks.py
@@ -83,9 +83,28 @@ if [ -z "$GRAPHIFY_PYTHON" ]; then
             */env\\ *) GRAPHIFY_PYTHON="${_SHEBANG#*/env }" ;;
             *)         GRAPHIFY_PYTHON="$_SHEBANG" ;;
         esac
+        # The launcher may be a /bin/sh wrapper (pipx via distlib's exec trick,
+        # emitted when the install path has spaces, e.g. macOS's
+        # "Application Support/pipx"): its shebang names the SHELL, and probing
+        # the shell with -c "import ..." runs whatever `import` is on PATH -
+        # ImageMagick's screenshot tool, which dumped its usage text on every
+        # commit (#3027). The wrapper's second line execs the real interpreter
+        # (exec' "/path/to/python" "$0" "$@"): take it from there, and never
+        # probe anything whose name is not a python.
+        case "$(basename "$GRAPHIFY_PYTHON" 2>/dev/null)" in
+            python*|pypy*) ;;
+            *) GRAPHIFY_PYTHON=$(printf '%s\\n' "$_GFY_HEAD" | sed -n '2s/^...exec. *"\\{0,1\\}\\([^"]*\\)"\\{0,1\\} .*/\\1/p') ;;
+        esac
+        case "$(basename "$GRAPHIFY_PYTHON" 2>/dev/null)" in
+            python*|pypy*) ;;
+            *) GRAPHIFY_PYTHON= ;;
+        esac
         # Allowlist: only keep characters valid in a filesystem path to prevent
-        # injection if the shebang contains shell metacharacters.
-        case "$GRAPHIFY_PYTHON" in
+        # injection if the shebang contains shell metacharacters. A space is
+        # not one - $GRAPHIFY_PYTHON is always expanded quoted - and a path
+        # that reaches this branch through a wrapper contains one by
+        # construction, so spaces are folded away before the check.
+        case "$(printf '%s' "$GRAPHIFY_PYTHON" | tr ' ' '_')" in
             *[!a-zA-Z0-9/_.@:\\\\-]*) GRAPHIFY_PYTHON="" ;;
         esac
         if [ -n "$GRAPHIFY_PYTHON" ] && ! "$GRAPHIFY_PYTHON" -c "$_GFY_PROBE" 2>/dev/null; then
--- a/graphify/hooks.py
+++ b/graphify/hooks.py
@@ -90,13 +90,14 @@ if [ -z "$GRAPHIFY_PYTHON" ]; then
         # ImageMagick's screenshot tool, which dumped its usage text on every
         # commit (#3027). The wrapper's second line execs the real interpreter
         # (exec' "/path/to/python" "$0" "$@"): take it from there, and never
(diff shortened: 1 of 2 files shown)


abhay-codes07:
Follow-up commit on the review's point about interpreter names: the probe gate is now `*python*|*pypy*`, so `python3.12`, `python@3.12`, `cpython` and `pypy3` all pass while `sh`/`bash`/`node` still do not.

abhay-codes07:
On the `.graphify_python` advisory (hooks.py:46): this PR's change to that line is only the stdout redirect (`2>/dev/null` → `>/dev/null 2>&1`) — it doesn't introduce or alter what the probe executes, so it's flagged only because the line was touched.

And the probe isn't a shell-command-execution vector: `_FROM_FILE` is (1) passed through a metacharacter allowlist immediately above — `case … *[!a-zA-Z0-9/_.@:\-]*) _FROM_FILE=""` — which blanks any content containing a space, `;`, `|`, `$`, backtick, etc., so nothing reaches `sh` as a command; (2) gated on `[ -x "$_FROM_FILE" ]`; and (3) always expanded quoted, so it's run as a single program path with `-c "$_GFY_PROBE"`, never shell-evaluated. The residual — it will execute an *allowlisted, executable* path a repo-local file names — is the intended mechanism for recording the graphify interpreter, the same project-local-config trust model as a `.venv` python path, and it predates this PR. So no change here; happy to file a separate hardening issue if tightening that trust boundary (e.g. refusing a `_FROM_FILE` under the repo worktree) is wanted.

Issue #1574, opened 2026-07-01
build_merge / --update drops existing graph's hyperedges from unchanged files (only nodes+edges are read from graph.json)


socar-tender:
## Summary

`build_merge()` — the function backing `graphify --update` — reads only `nodes` and `edges` from the existing `graph.json`, never `hyperedges`. As a result, **every incremental update silently drops all hyperedges belonging to files that weren't re-extracted in that run.** After a full build the graph carries all its hyperedges; the first `--update` that touches even one file collapses the graph's hyperedge set down to just the hyperedges of the changed file(s).

This is the highest-signal part of the semantic layer (domain-flow groupings), so the loss is disproportionately damaging to `query`/`explain` quality.

Version: `graphifyy==0.9.3`.

## Root cause

In `build_merge()` (`build.py:687`), the existing graph is loaded with only:

```python
existing_nodes = list(data.get("nodes", []))   # build.py:719
existing_edges = list(data.get(links_key, [])) # build.py:720
```

`data.get("hyperedges")` is never read. The merged graph's hyperedges are then set solely from the new chunks in `build()`:

```python
hyperedges = extraction.get("hyperedges", [])
if hyperedges:
    ...
    G.graph["hyperedges"] = hyperedges         # build.py:591  (plain overwrite)
```

`to_json()` faithfully writes whatever `build_merge` left (`export.py:536`), so the output graph ends up with only the changed files' hyperedges. `--update` reaches this via `_build_merge(...)` at `__main__.py:4844`.

## Why this looks like an unintended gap (not by design)

The codebase already knows how to preserve hyperedges across a merge — build_merge just doesn't use either mechanism:

- `attach_hyperedges()` (`export.py:464`) merges new hyperedges into the existing set with id-level dedup:
  ```python
  existing = G.graph.get("hyperedges", [])
  seen_ids = {h["id"] for h in existing}
  for h in hyperedges:
      if h.get("id") and h["id"] not in seen_ids:
          existing.append(h)
  G.graph["hyperedges"] = existing
  ```
- The `watch` path explicitly carries existing hyperedges forward (`watch.py:682`):
  ```python
  "hyperedges": existing.get("hyperedges", []),
  ```

Only the `build_merge` / `--update` path plain-overwrites from the new chunks.

## Reproduction

1. Full-build a repo that produces hyperedges from several doc files (`graphify .`). Note the hyperedge count in `graph.json` (`.graph.hyperedges` / top-level `hyperedges`).
[...]

PR #3113, opened 2026-08-26
fix(build): route carried hyperedges through the dedup survivor remap (#3102)

### LINKED ISSUE #3102: build_merge(): carried unchanged hyperedges bypass #2805 remap and retain dangling member IDs
## Summary

In `graphifyy==0.9.49`, `build_merge()` can leave dangling hyperedge member IDs after entity deduplication.

The fix for #2805 correctly remaps members of hyperedges that are passed through `build()`. However, hyperedges carried forward from unchanged source files are attached only after `build()` and deduplication have finished. Their member IDs therefore bypass the survivor remap.

The resulting graph has inconsistent references:

- ordinary edge endpoints are rewired to the dedup survivor;
- the carried hyperedge still names the merged-away node;
- the missing member remains in `G.graph["hyperedges"]` without a backing graph node.

This appears to be an interaction gap between the unchanged-hyperedge preservation added for #1574 and the hyperedge-member dedup remap added for #2805.

## Environment

- `graphifyy==0.9.49`
- Python 3.12
- Windows 11
- No LLM backend or API key involved

## Minimal reproduction

This script uses only synthetic data. It writes the baseline graph to a temporary directory that is deleted automatically.

```python
import importlib.metadata as md
import json
import tempfile
from pathlib import Path

from graphify.build import build_merge


nodes = [
    {
        "id": "alpha_a",
        "label": "Alpha Concept",
        "file_type": "concept",
        "source_file": "notes/a.md",
    },
    {
        "id": "alpha_concept_long_variant_id",
        "label": "alpha_concept",
        "file_type": "concept",
        "source_file": "notes/b.md",
    },
    {
        "id": "beta_node",
        "label": "Beta",
        "file_type": "concept",
        "source_file": "notes/group.md",
    },
    {
        "id": "gamma_node",
        "label": "Gamma",
        "file_type": "concept",
        "source_file": "notes/group.md",
    },
]

edges = [
    {
        "source": "alpha_concept_long_variant_id",
        "target": "beta_node",
        "relation": "references",
        "confidence": "EXTRACTED",
        "confidence_score": 1.0,
        "source_file": "notes/b.md",
    }
]

hyperedge = {
    "id": "the_group",
    "label": "The Group",
    "nodes": [
        "alpha_concept_long_variant_id",
        "beta_node",
        "gamma_node",
    ],
    "relation": "participate_in",
    "confidence": "EXTRACTED",
    "confidence_score": 1.0,
    "source_file": "notes/group.md",
}

# An unrelated incremental chunk. The hyperedge's source is unchanged, so the
# existing hyperedge is carried forward by build_merge().
new_chunk = {
    "nodes": [
        {
            "id": "delta_node",
            "label": "Delta",
            "file_type": "concept",
            "source_file": "notes/delta.md",
        }
    ],
    "edges": [],
    "hyperedges": [],
}

with tempfile.TemporaryDirectory() as tmp:
    root = Path(tmp)
    out = root / "graphify-out"
    out.mkdir()
    graph_path = out / "graph.json"

    graph_path.write_text(
        json.dumps(
            {
                "directed": False,
                "multigraph": False,
                "graph": {"hyperedges": [hyperedge]},
                "nodes": nodes,
                "links": edges,
                "hyperedges": [hyperedge],
            }
        ),
        encoding="utf-8",
    )

    graph = build_merge(
        [new_chunk],
        graph_path=graph_path,
        root=root,
    )

    carried = next(
        item
        for item in graph.graph.get("hyperedges", [])
        if item.get("id") == "the_group"
    )
    dangling = [member for member in carried["nodes"] if member not in graph]

    print("graphify", md.version("graphifyy"))
    print("edge_rewired", graph.has_edge("alpha_a", "beta_node"))
    print("hyperedge_members", carried["nodes"])
    print("dangling_members", dangling)

    assert "alpha_a" in graph
    assert "alpha_concept_long_variant_id" not in graph
    assert graph.has_edge("alpha_a", "beta_node")
    assert dangling == ["alpha_concept_long_variant_id"]
```

## Actual output

```text
[graphify] Deduplicated 1 node(s) (1 exact).
graphify 0.9.49
edge_rewired True
hyperedge_members ['alpha_concept_long_variant_id', 'beta_node', 'gamma_node']
dangling_members ['alpha_concept_long_variant_id']
```

## Expected result

The ordinary edge and the carried hyperedge should both follow the same dedup survivor map:

```text
edge_rewired True
hyperedge_members ['alpha_a', 'beta_node', 'gamma_node']
dangling_members []
```

## Root cause

`build_merge()` sends the existing nodes and edges through `build()`:

```python
base = [{"nodes": existing_nodes, "edges": existing_edges}] if had_graph else []
all_chunks = base + list(new_chunks)
G = build(all_chunks, ...)
```

The existing hyperedges are deliberately omitted from `base`. After `build()` has deduplicated the nodes and rewired ordinary edges, unchanged hyperedges are carried forward separately:

```python
if carried:
    from graphify.export import attach_hyperedges
    attach_hyperedges(G, carried)
```

`attach_hyperedges()` deduplicates hyperedges by hyperedge ID, but it does not validate or remap member node IDs. Consequently, carried hyperedges never see the survivor map introduced for #2805.

## Why existing fixes do not cover this path

- #1574 preserves hyperedges from unchanged files by attaching them after the rebuild.
- #2805 remaps members only for hyperedges included in the extraction passed through `build()`.
- A hyperedge carried by the #1574 path is therefore outside the #2805 remap path.

## Impact

- `graph.json` can contain hyperedge members that do not resolve to nodes.
- Pairwise edges appear correct, which makes the corruption difficult to notice.
- A health check focused on ordinary edge endpoints can report zero dangling edges while hyperedge members are dangling.
- Repeated incremental updates preserve the stale member because the hyperedge continues to be carried unchanged.

## Suggested direction

Before returning from `build_merge()`, carried hyperedges should be processed using the same survivor mapping as ordinary edges and new hyperedges.

Possible approaches include:

1. include eligible carried hyperedges in the data passed through `build()` and deduplication; or
2. expose the dedup survivor map and apply it to carried hyperedges before `attach_hyperedges()`.

It would also be useful to add a final referential-integrity check covering every hyperedge member, plus a regression test asserting that a full build and an incremental merge produce equivalent member IDs for this case.

## Related issues

- #1574 — preserve existing hyperedges across `build_merge()` updates
- #2805 — remap hyperedge members to entity-dedup survivors
- #2484 — relabel hyperedge members during `merge-graphs`

abhay-codes07:
Closes #3102.

## The problem

`build_merge()` carries hyperedges forward from files that were neither re-extracted nor deleted (#1574). They were attached to `G` **after** `build()` and entity dedup had finished. Dedup rewires every edge endpoint — and every hyperedge member it sees — onto the survivor (#2805), but a hyperedge attached after the fact kept naming the merged-away node. The written graph then had:

- ordinary edges rewired to the survivor;
- the carried hyperedge still naming the node that no longer exists;
- a member in `G.graph["hyperedges"]` with no backing graph node.

The reporter's synthetic repro reproduces on 0.9.50 exactly as filed: `edge_rewired True`, `dangling_members ['alpha_concept_long_variant_id']`.

## The change

The carried set is computed **before** the build and rides into `build()` on the base chunk (`{"nodes": existing_nodes, "edges": existing_edges, "hyperedges": carried}`), so dedup remaps its members exactly as it remaps the new chunks' hyperedges, and `build_from_json` validates them against the built node set like any other. The prune-set helpers (`_prune_match`, the #2446 root derivation) move ahead of the build call — they never depended on `G`, only on the prune sources and the existing graph. The old `attach_hyperedges` step also id-deduped against hyperedges the new chunks re-emitted; that is preserved by leaving a same-id hyperedge to the new chunk.

Nothing about *which* hyperedges are carried changes: semantically re-extracted sources still replace theirs, pruned sources still drop theirs.

## Tests

`tests/test_carried_hyperedge_remap.py` — 6 tests built on the issue's data: the carried hyperedge's members are remapped onto the survivor (not merely dropped); the written `graph.json` has no hyperedge member without a node; edges and hyperedges agree on the survivor; a hyperedge re-emitted by the new chunk under the same id wins and is not duplicated; a pruned source's hyperedge is still dropped; and an unchanged hyperedge with no dedup involved is carried verbatim. With the reorder reverted, 3 of 6 fail. `test_build_merge_hyperedges_and_prune`, `test_build`, the `test_build_merge*`, hyperedge and dedup suites are unchanged (219 passed); the full suite matches the `v8` baseline.

### DIFF
--- a/graphify/build.py
+++ b/graphify/build.py
@@ -1726,11 +1726,6 @@ def build_merge(
                 file=sys.stderr,
             )
 
-    base = [{"nodes": existing_nodes, "edges": existing_edges}] if had_graph else []
-
-    all_chunks = base + list(new_chunks)
-    G = build(all_chunks, directed=directed, dedup=dedup, dedup_llm_backend=dedup_llm_backend, root=root)
-
     # Prune set for deleted source files — both the raw form (matches nodes that
     # kept absolute source_file) and the normalised relative form (matches nodes
     # relativised by _norm_source_file at build time). .resolve() (via _eff_root)
@@ -1797,12 +1792,25 @@ def build_merge(
     # deleted (#1574). build() only sees the new chunks' hyperedges, so without
     # this every --update collapses the graph's hyperedge set down to just the
     # changed files'. Re-extracted files' prior hyperedges are dropped (their new
-    # version is already in G — replace-per-source, like nodes/edges); deleted
-    # files' are dropped via prune_set. id-dedup (attach_hyperedges) so a carried
-    # hyperedge never duplicates one the new chunks re-emitted. Mirrors watch.py,
-    # which already preserves existing hyperedges across a rebuild.
+    # version is already in the new chunks — replace-per-source, like
+    # nodes/edges); deleted files' are dropped via prune_set; id-dedup so a
+    # carried hyperedge never duplicates one the new chunks re-emitted. Mirrors
+    # watch.py, which already preserves existing hyperedges across a rebuild.
+    #
+    # The carried set rides INTO build() on the base chunk rather than being
+    # attached to G afterwards (#3102): entity dedup rewires every edge endpoint
+    # and every hyperedge member it sees onto the survivor (#2805), but a
+    # hyperedge attached after the fact kept naming the merged-away node — a
+    # dangling member with no backing node in the written graph.
+    carried_hyperedges: list[dict] = []
     if existing_hyperedges:
-        carried = []
+        carried = carried_hyperedges
+        _new_hyperedge_ids = {
+            he.get("id")
+            for chunk in new_chunks
+            for he in (chunk.get("hyperedges") or [])
+            if isinstance(he, dict) and he.get("id")
+        }
         for he in existing_hyperedges:
             if not isinstance(he, dict):
                 continue
@@ -1815,10 +1823,17 @@ def build_merge(
                 continue  # semantically re-extracted — replaced by the new chunk's version
             if _prune_match(sf):
                 continue  # deleted — pruned
+            if he.get("id") and he.get("id") in _new_hyperedge_ids:
+                continue  # the new chunks re-emitted it — theirs wins
             carried.append(he)
-        if carried:
-            from graphify.export import attach_hyperedges
-            attach_hyperedges(G, carried)
+
+    base = (
+        [{"nodes": existing_nodes, "edges": existing_edges, "hyperedges": carried_hyperedges}]
+        if had_graph else []
+    )
+
+    all_chunks = base + list(new_chunks)
+    G = build(all_chunks, directed=directed, dedup=dedup, dedup_llm_backend=dedup_llm_backend, root=root)
 
     # Prune nodes and edges from deleted source files
     if prune_sources:
--- a/graphify/build.py
+++ b/graphify/build.py
@@ -1649,6 +1649,9 @@ def build_merge(
     graph to inherit from. An explicit True/False always overrides the on-disk
     flag.
     """
+    # Iterated more than once below (source sets, the hyperedge carry, the
+    # build itself), so a one-shot iterator must be materialised first.
+    new_chunks = list(new_chunks)
     graph_path = Path(graph_path if graph_path is not None else _default_graph_json())
     _loaded = _load_existing_graph(graph_path)
     if _loaded is not None:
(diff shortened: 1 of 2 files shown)


abhay-codes07:
Follow-up commit on the review's iterator point: `build_merge` already iterated `new_chunks` twice before this change (and the parameter is typed `list`), but it is now materialised with `list()` at the top so a one-shot iterator can never reach any of the passes.

safishamsi:

PR #3114, opened 2026-08-26
fix(extract): hollow, unparseable and omitting chunks count as incomplete so the shrink guard stays armed (#3105)

### LINKED ISSUE #3105: `extract` silently overwrote a 570-node graph with 111 nodes: hollow LLM responses don't count as "incomplete"
**Version:** 0.9.27 · Windows · `--backend ollama --model qwen2.5-coder:7b`

## What happened

Two consecutive `graphify . --update` runs on the same unchanged repo. The first was **refused by the #479 shrink guard** and preserved the graph. The second **wrote 111 nodes over an existing 570-node graph, with no warning at all**.

| run | chunks *failed* | hollow responses | invalid JSON | shrink guard | result |
|---|---|---|---|---|---|
| 1 | **3** | 8 | 8 | fired (×2) | refused — graph preserved at 570 |
| 2 | **0** | 6 | 6 | never fired | **wrote 111 nodes / 118 edges** |

The graph lost ~80% of its nodes. Per-file, the largest subfolder went from 378 nodes to 39; one 138 KB markdown file went from 77 to 12. Nothing was deleted from the repo between runs — it was the same corpus, one commit old.

## Why the guard didn't fire

`cli.py:3458`:

```python
_force_write = cli_allow_partial or not _extraction_incomplete
_wrote = _to_json(G, communities, str(graph_json_path), force=_force_write)
```

The comment above it explains the intent, and it's reasonable:

> `force=True` bypasses the #479 shrink guard entirely. A full build legitimately shrinks (fuzzy dedup collapse, deleted code) so it keeps `force=True` — EXCEPT when this run's extraction was incomplete (an extractor pass crashed or some semantic chunks failed).

The gap is in what counts as *incomplete*. `_extraction_incomplete` tracks **hard failures** — a crashed extractor pass, a chunk that errored. It does not track the two outcomes that dominated run 2:

- `ollama returned a hollow response (content=no nodes/edges, output_tokens=N); treating as truncation so adaptive retry can bisect the chunk`
- `LLM returned invalid JSON, skipping chunk`

Both are logged as warnings and handled gracefully — the run continues, no pass "fails", and `_extraction_incomplete` stays `False`. So the run is classified as **complete**, `force=True`, and the shrink guard is bypassed.

**With an LLM backend, a hollow or unparseable response is the normal way an extraction silently produces a fraction of the nodes.** That is exactly the case the guard exists to catch, and it's the one case where it's disabled.

Run 2 also emitted this, which shows the loss was visible to graphify at write time:

```
WARNING: 10/24 dispatched file(s) produced no nodes and are absent from the graph:
<file>.md, <file>.md, <file>.pdf, <file>.pdf, <file>.pdf (+5 more). The model returned a
response but omitted them; a re-run will retry them.
```

10 of 24 files contributed nothing, and the run still counted as complete.

## Reproduce

Any repo with large markdown files (mine has 177 KB / 138 KB / 119 KB / 98 KB) and a small local model, so the LLM returns prose instead of JSON:

```
graphify <repo> --update      # some chunks fail  -> refused, graph preserved
graphify <repo> --update      # none fail, several hollow -> overwrites with a fraction
```

The second run benefits from the first run's cached chunks, which is what removes the hard failures while leaving the hollow ones.

## Suggested fix

Either would have prevented the loss:

1. **Count hollow/invalid-JSON responses toward `_extraction_incomplete`.** They are not clean successes — the adaptive bisect already treats a hollow response as truncation.
2. **Make the shrink guard a magnitude check that `force` cannot fully disable.** A full build legitimately shrinks by a few percent from dedup; an 80% drop is not that. A ratio threshold (say, refuse below ~50% of the existing node count unless `--allow-partial`) keeps the legitimate-shrink case working while catching this one.

(1) is the more precise fix; (2) is the backstop that doesn't depend on classifying every future failure mode correctly.

Also worth considering: the `WARNING: N/M dispatched file(s) produced no nodes` line already computes the signal. If `N/M` is large, that alone is strong evidence the run should not be trusted to overwrite.

## Impact

The good graph was recoverable only because `backup_if_protected` had written `2026-08-26/graph.json` before the overwrite — thank you for that. Without it the run would have been unrecoverable, and the loss is silent: exit code 0, no warning, and the next incremental run stamps the manifest against the degraded graph.

abhay-codes07:
Closes #3105.

## The problem

A full build writes with `force=True`, bypassing the #479 shrink guard, unless the run was classified as incomplete. `_extraction_incomplete` tracked hard failures only: a crashed extractor pass, a chunk that raised. A chunk that comes back **hollow** after every retry, or as **invalid JSON**, or that simply **omits** some of its files does not raise — it returns fewer nodes and counts as a *succeeded* chunk. So the reporter's two consecutive `--update` runs on an unchanged repo went:

| run | raised chunks | hollow | guard | result |
|---|---|---|---|---|
| 1 | 3 | 8 | fired | refused — 570 nodes preserved |
| 2 | 0 | 6 | never fired | **wrote 111 nodes over 570** |

Run 2 even printed `10/24 dispatched file(s) produced no nodes and are absent from the graph` and still counted as complete. With an LLM backend this is the normal way an extraction silently produces a fraction of the graph — the case the guard exists for, and the one case where it was disabled.

## The change

After the semantic pass, a non-empty `uncovered_files` list (files the model omitted — which is also where an invalid-JSON chunk's files land) or any partial/hollow file now sets `_extraction_incomplete = True`, with one stderr line saying how many files produced no nodes and how many came back truncated or hollow, and that `--allow-partial` overrides. From there the existing machinery does the rest: `force=False`, the shrink guard compares against the existing graph, a refusal exits non-zero and leaves the manifest unstamped so the next run retries those files.

A complete run — every dispatched file covered, nothing partial — keeps `force=True` exactly as before, so a legitimately shrinking full build (dedup, deleted code) is unaffected.

## Tests

`tests/test_hollow_chunks_arm_shrink_guard.py` — 5 tests through the real CLI with a stubbed semantic pass: an omitting chunk arms the guard (`force=False`); a hollow chunk (files marked partial, nothing raised — the reporter's run 2) arms it; `--allow-partial` still overrides; a fully covered run keeps `force=True` with no warning; and a refused write exits non-zero without stamping the manifest. With the fix reverted, the two "arms the guard" tests fail. `test_incomplete_build_guard`, `test_extract_cli` and `test_extract_code_only_cli` are unchanged (60 passed); the full suite matches the `v8` baseline.

### DIFF
--- a/graphify/cli.py
+++ b/graphify/cli.py
@@ -3813,6 +3813,26 @@ def dispatch_command(cmd: str) -> None:
                     _strip_partial_markers as _strip_partial,
                 )
                 _partial_semantic_files = set(_partial_sf(fresh))
+                # A chunk that came back hollow after every retry, or as
+                # unparseable JSON, or that simply omitted some of its files,
+                # does not raise - it returns fewer nodes - so it counted as a
+                # SUCCEEDED chunk above and the run read as complete, force=True
+                # bypassed the shrink guard, and a 570-node graph was overwritten
+                # with 111 nodes without a word (#3105). With an LLM backend
+                # that is the normal way an extraction silently produces a
+                # fraction of the graph, so it must arm the guard exactly like a
+                # crashed chunk does. --allow-partial still overrides.
+                _omitted_files = list(fresh.get("uncovered_files") or [])
+                if _omitted_files or _partial_semantic_files:
+                    _extraction_incomplete = True
+                    print(
+                        f"[graphify extract] semantic extraction is incomplete: "
+                        f"{len(_omitted_files)} dispatched file(s) produced no nodes and "
+                        f"{len(_partial_semantic_files)} came back truncated or hollow. "
+                        f"The shrink guard stays armed for this write; pass "
+                        f"--allow-partial to overwrite a larger existing graph anyway.",
+                        file=sys.stderr,
+                    )
                 try:
                     _save_semantic_cache(
                         fresh.get("nodes", []),
--- a/tests/test_hollow_chunks_arm_shrink_guard.py
+++ b/tests/test_hollow_chunks_arm_shrink_guard.py
@@ -1,0 +1,118 @@
+"""Hollow, unparseable and omitting chunks must count as incomplete (#3105).
+
+The #479 shrink guard is bypassed (force=True) on a run that is classified
+as complete. `_extraction_incomplete` tracked hard failures only — a crashed
+pass, a chunk that raised. A chunk that came back hollow after every retry,
+or as invalid JSON, or that simply omitted some of its files does not raise:
+it returns fewer nodes and counts as a SUCCEEDED chunk. Two consecutive
+`--update` runs on an unchanged repo: the first had 3 raised chunks and the
+guard refused; the second had 0 raised, 6 hollow, and wrote 111 nodes over a
+570-node graph without a word.
+"""
+from __future__ import annotations
+
+import pytest
+
+import graphify.__main__ as mainmod
+
+
+def _corpus(tmp_path):
+    (tmp_path / "README.md").write_text("# Notes\nThe entry point overview.\n", encoding="utf-8")
+    (tmp_path / "GUIDE.md").write_text("# Guide\nHow to use the thing.\n", encoding="utf-8")
+    return tmp_path
+
+
+def _record_force(monkeypatch):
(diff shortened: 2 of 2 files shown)


safishamsi:

PR #3117, opened 2026-08-26
feat(salesforce): read metadata XML, not just Apex

rsoesemann:
In a Salesforce codebase, `.cls` and `.trigger` are a minority of the files. Objects, fields, layouts, list views and permission sets hold a lot of the wiring, and none of it was in the graph — those files aren't classified at all today. On my-org-butler that's 34 files invisible out of 186 tracked.

Salesforce metadata is plain XML, so this is one generic element walk with the stdlib parser and no new dependency — no logic per component type. What a file declares comes from its filename. What it references comes from leaf element text, which is how Salesforce spells cross-component links:

```xml
<classAccesses>
    <apexClass>NotifyUser</apexClass>   <- a reference
    <enabled>true</enabled>             <- a value
</classAccesses>
```

References are emitted as sourceless placeholders, so the corpus rewire binds them to the real definition. The permission set above ends up pointing at `NotifyUser.cls` without this extractor knowing what a permission set is.

Telling references from values is the only real judgement here. Two rules, both derived from the element names that actually carry identifiers in real Salesforce codebases: anything with a Salesforce suffix (`__c`, `__mdt`, …) is always an API name, and for standard components there's a flat set of about twenty element names (`apexClass`, `object`, `field`, …). Everything else — `true`, `Checkbox`, `Large`, ALL-CAPS layout tokens like `TASK.SUBJECT` — is dropped. A 134-element object file yields one node.

Dispatch is by filename, following the existing `.blade.php` precedent, because the component kind sits in a compound suffix. Plain `.xml` stays unclaimed on purpose: claiming it would pull every pom.xml and web.xml in every repository into every graph, which is a separate decision. That does mean `collect_files` and `classify_file` each needed a filename check, since both gate on extension.

On my-org-butler, 186 tracked files: files in the graph 70 → 104, nodes 616 → 692, edges 618 → 702, of which 68 nodes come from metadata. `explain Content__c` returned nothing before — the field had no node. Now it returns the owning object plus the permission set, page layout and list view that use it.

Tests: 15, in `tests/test_salesforce_meta_xml.py`. Since the feature is new, "fails without the change" proves nothing, so I checked them by mutation instead — twelve plausible wrong implementations (stamping `source_file` on references, claiming all `.xml`, dropping the value filter, losing the owning-object lookup, swallowing parse errors, …), each caught by at least one test. That found a real bug on the first pass: `<name>` was being discarded as display text, which loses a genuine reference in custom-metadata records. Three negative controls: values must not become edges, ALL-CAPS layout tokens must not become components, and `web.xml` must not be claimed.

Known limitations. Flows contribute the objects and fields they touch but not their logic, and formula fields and validation rules hold expressions rather than plain names, so their field references aren't picked up. Salesforce names an object's tab after the object, so `Memory__c` exists as the object, the tab, and the shared name Apex points at; merging by name would be right sometimes and wrong sometimes, so nothing is merged — every edge is still reachable. `.cls-meta.xml` sidecars are skipped since the Apex extractor already owns those classes.

Independent of #3106; either can land first.

### DIFF
--- a/graphify/detect.py
+++ b/graphify/detect.py
@@ -508,6 +508,13 @@ def classify_file(path: Path) -> FileType | None:
     from graphify.manifest_ingest import is_package_manifest_path
     if is_package_manifest_path(path):
         return FileType.CODE
+    # Salesforce metadata (`Account.object-meta.xml`) is parsed deterministically,
+    # so it takes the AST path like the manifests above. Its component kind sits in
+    # a compound suffix, and plain `.xml` is deliberately NOT in CODE_EXTENSIONS —
+    # claiming it would sweep every pom.xml/web.xml into the graph.
+    from graphify.extractors.salesforce_meta_xml import is_salesforce_meta_xml_path
+    if is_salesforce_meta_xml_path(path):
+        return FileType.CODE
     # Compound extensions must be checked before simple suffix lookup
     if path.name.lower().endswith(".blade.php"):
         return FileType.CODE
--- a/graphify/extract.py
+++ b/graphify/extract.py
@@ -53,6 +53,7 @@ from graphify.extractors.pascal_forms import extract_delphi_form, extract_lazaru
 from graphify.extractors.powershell import extract_powershell, extract_powershell_manifest  # noqa: F401
 from graphify.extractors.razor import extract_razor  # noqa: F401
 from graphify.extractors.rust import extract_rust  # noqa: F401
+from graphify.extractors.salesforce_meta_xml import extract_salesforce_meta_xml, is_salesforce_meta_xml_path  # noqa: F401
 from graphify.extractors.sln import extract_sln  # noqa: F401
 from graphify.extractors.sql import extract_sql  # noqa: F401
 from graphify.extractors.terraform import extract_terraform  # noqa: F401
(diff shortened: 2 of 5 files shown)


rsoesemann:
Thanks — I reproduced all four against the branch. Two hold, two don't.

**Doesn't hold: reference stub ids never rewire to definitions.** They do. With an object at `objects/Widget__c/Widget__c.object-meta.xml` and a permission set referencing `Widget__c`, the graph ends up with exactly one `Widget__c` node, sourced at the object file. Binding is by **label** in `_rewire_unique_stub_nodes`, not by id — a sourceless stub is meant to have a different id from the definition, and that is what lets the corpus-level rewire collapse them. `tests/test_salesforce_meta_xml.py::test_permission_set_reaches_the_real_apex_class` asserts the same thing end to end for an Apex class.

**Doesn't hold: owner stub id doesn't match the owner object's component id.** Same mechanism. For a field at `objects/Widget__c/fields/Size__c.field-meta.xml` the resulting edge is `contains: Widget__c -> Size__c`, originating from the real object definition, not from a stub.

**Holds, deliberately not fixed: `_is_api_name` rejects all-uppercase identifiers.** Correct — an Apex class genuinely named `API` would be dropped along with the layout placeholders (`NAME`, `TASK.SUBJECT`) the screen exists for.

**Holds, deliberately not fixed: names with spaces or hyphens are skipped.** Correct — a profile's `<layout>Account-Account Layout</layout>` produces no reference.

I had both fixed, then measured how often the patterns actually occur before committing: **zero times in 727 metadata files** across two real Salesforce codebases (34 and 693 files). Reference-edge counts were identical with and without the change — 43 and 2769. Layout assignments live in profiles, and neither codebase uses profiles. So I dropped the change rather than ship a widening I can't demonstrate. Both are worth doing the moment there's a corpus that exercises them, and the shape of the fix is small: apply the identifier rule laxly for elements that name a component by construction, and confine the all-caps screen to `field`/`fields`/`columns`.

PR #3118, opened 2026-08-26
fix(elixir): extract functions whose only clause has a when guard (#3111)

### LINKED ISSUE #3111: Elixir extractor drops functions whose only clause has a `when` guard
## Summary

The Elixir AST extractor silently drops every function whose **only** clause carries a `when` guard. In a real 2,900-file Elixir/Phoenix codebase this hid **19 of 582 functions (100% of that shape)** from the graph — including several that are architecturally significant, so downstream analysis (god nodes, community detection, "what calls this?") is systematically blind to them.

No error is raised; the node is simply never created.

## Reproduction

`lib/demo.ex`:

```elixir
defmodule Demo do
  # A: single clause, no guard
  def plain(x) do
    x + 1
  end

  # B: single clause, WITH guard
  def guarded(x) when is_integer(x) do
    x + 1
  end

  # C: multiple clauses, one with a guard
  def mixed(x) when is_integer(x) do
    x + 1
  end

  def mixed(_), do: :error

  # D: private, single clause, with guard
  defp guarded_private(x) when is_binary(x) do
    String.upcase(x)
  end
end
```

```python
from pathlib import Path
res = extract([Path("lib/demo.ex")], cache_root=Path("."))
print(sorted((n.get("label") or "").rstrip("()") for n in res["nodes"]))
```

**Actual:** `['Demo', 'demo.ex', 'mixed', 'plain']`
**Expected:** also `guarded` and `guarded_private`

| case | shape | extracted |
|---|---|---|
| A | single clause, no guard | ✅ |
| B | single clause, `when` guard | ❌ |
| C | multi-clause, one guarded | ✅ (picked up via the unguarded clause) |
| D | `defp`, single clause, `when` guard | ❌ |

C passing is what makes this easy to miss: a guarded function is only lost when *no* clause is unguarded.

## Root cause

`graphify/extractors/elixir.py`, in the `keyword in ("def", "defp")` branch:

```python
for child in arguments_node.children:
    if child.type == "call":
        ...
    elif child.type == "identifier":
        ...
if not func_name:
    return          # <-- node silently dropped
```

tree-sitter-elixir wraps a guarded head in a `binary_operator`, so `arguments_node`'s direct child is not a `call`:

```
def plain(x) do   ...  ->  arguments children = ['call']
def guarded(x) when is_integer(x) do ... -> arguments children = ['binary_operator']
```

and that `binary_operator`'s children are `[call 'guarded(x)', when, call 'is_integer(x)']`.

Neither branch matches, `func_name` stays `None`, and the early `return` discards the function.

## Suggested fix

Descend into the `binary_operator`'s left operand before the existing type checks:

```python
for child in arguments_node.children:
    # `def f(x) when guard do` -> arguments child is
    # binary_operator(call, "when", call); use the left operand.
    if child.type == "binary_operator":
        child = next(
            (c for c in child.children if c.type in ("call", "identifier")),
            child,
        )
    if child.type == "call":
        ...
```

Verified locally against the reproduction above: all four cases extract (`plain`, `guarded`, `mixed`, `guarded_private`). I reverted the patched package afterwards, so this is a suggestion rather than a tested-in-CI change — happy to open a PR with a regression test covering all four shapes if that helps.

## Note for anyone reproducing

`extract()` caches per file, so re-running after editing the extractor replays the cached result. Use a fresh filename (or clear the cache) when testing a fix — I lost a cycle to this.

## Environment

- graphify (`graphifyy`) installed via `uv tool`, Python 3.12
- tree-sitter-elixir via the bundled `tree_sitter_elixir`
- macOS

## Impact detail

Across `lib/**/*.ex` in the affected repo:

| function shape | count | present in graph |
|---|---|---|
| single clause + `when` guard | 19 | **0 (0%)** |
| everything else | 563 | 547 (97%) |

Examples of what was missing: an authorization invariant predicate, an audit-log actor-role resolver, and a guide-lookup entry point — all public API, all referenced elsewhere.

santhiprakash:
## Summary

The Elixir AST extractor silently drops every function whose **only** clause carries a `when` guard. On a real 2,900-file Elixir/Phoenix codebase this hid 19 of 582 functions (every function of that shape) from the graph. No error is raised; the node is never created.

## Problem

tree-sitter-elixir wraps a guarded head in a `binary_operator`:

- `def plain(x) do ...` → `arguments` children = `['call']`
- `def guarded(x) when is_integer(x) do ...` → `arguments` children = `['binary_operator']`

The extractor only looked for a direct `call`/`identifier` child of `arguments`, so `func_name` stayed `None` and the function was dropped. Multi-clause functions survived via an unguarded clause, which is why this was easy to miss.

## Fix

Unwrap `binary_operator` (the `when` guard) to the left-hand head before reading the function name. Scoped to `def`/`defp` name extraction; the rest of the walker is unchanged.

This is independent of open #2970 (single-line `def f(x), do: ...` call-graph bodies). That PR does not unwrap guarded heads.

## Verification

```
uv run --frozen pytest tests/test_languages.py -k elixir -q
```

9 passed, including `test_elixir_guarded_single_clause_is_extracted` covering:

| case | shape | extracted |
|---|---|---|
| A | single clause, no guard | still extracted |
| B | single clause, `when` guard | now extracted |
| C | multi-clause, one guarded | still extracted |
| D | `defp`, single clause, `when` guard | now extracted |

## Reproduction (before fix)

```elixir
defmodule Demo do
  def guarded(x) when is_integer(x) do
    x + 1
  end
end
```

`extract_elixir` produced `Demo` and `demo.ex` only — no `guarded` node.

Fixes #3111

### DIFF
--- a/graphify/extractors/elixir.py
+++ b/graphify/extractors/elixir.py
@@ -128,6 +128,19 @@ def extract_elixir(path: Path) -> dict:
             func_name = None
             if arguments_node:
                 for child in arguments_node.children:
+                    # tree-sitter-elixir wraps a guarded head
+                    # (`def f(x) when guard`) in `binary_operator`;
+                    # without unwrapping, a function whose only clause
+                    # carries `when` is dropped (#3111).
+                    while child.type == "binary_operator":
+                        head = None
+                        for sub in child.children:
+                            if sub.type in ("call", "identifier", "binary_operator"):
+                                head = sub
+                                break
+                        if head is None:
+                            break
+                        child = head
                     if child.type == "call":
                         for sub in child.children:
                             if sub.type == "identifier":
--- a/tests/test_languages.py
+++ b/tests/test_languages.py
@@ -1440,6 +1440,47 @@ def test_elixir_method_edges():
     assert len(methods) >= 3
 
 
+def test_elixir_guarded_single_clause_is_extracted(tmp_path):
+    """A function whose only clause has a `when` guard must still get a node.
+
+    tree-sitter-elixir wraps `def f(x) when guard` in a `binary_operator`,
+    so the head is not a direct `call` child of `arguments`. Multi-clause
+    functions survive via an unguarded clause; a single guarded clause
+    was dropped entirely (#3111).
+    """
+    src = tmp_path / "demo.ex"
+    src.write_text(
+        "defmodule Demo do\n"
+        "  def plain(x) do\n"
+        "    x + 1\n"
+        "  end\n"
+        "\n"
+        "  def guarded(x) when is_integer(x) do\n"
+        "    x + 1\n"
+        "  end\n"
+        "\n"
+        "  def mixed(x) when is_integer(x) do\n"
+        "    x + 1\n"
+        "  end\n"
+        "\n"
+        "  def mixed(_), do: :error\n"
+        "\n"
+        "  defp guarded_private(x) when is_binary(x) do\n"
+        "    String.upcase(x)\n"
+        "  end\n"
+        "end\n"
+    )
+    r = extract_elixir(src)
+    assert "error" not in r
+    labels = {(n.get("label") or "").rstrip("()") for n in r["nodes"]}
+    assert "plain" in labels
+    assert "mixed" in labels
+    assert "guarded" in labels, f"single-clause guarded def dropped: {sorted(labels)}"
+    assert "guarded_private" in labels, (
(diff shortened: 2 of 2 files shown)


safishamsi:
Shipped in v0.9.51 via authorship-preserving cherry-pick so you keep contributor-graph credit. Thanks @santhiprakash! Release: https://github.com/Graphify-Labs/graphify/releases/tag/v0.9.51

PR #3119, opened 2026-08-26
fix(pascal): resolve calls through a global singleton variable across files (#3101)

### LINKED ISSUE #3101: Pascal/Delphi: cross-file calls via a global singleton variable (`mm.Method`) produce no calls edges — same pattern as Kotlin #1698, Go #1313
### Environment

- graphifyy 0.9.50 (PyPI), Windows 11, Python 3.13
- Extraction invoked via `graphify update .` (CLI pipeline), `tree-sitter-pascal` extra installed

### Summary

Calls to a method on a **global singleton variable** (`var mm: Tmm;` declared in one unit, called as `mm.ServerReport(...)` from another) resolve to a `calls` edge **only when caller and callee are in the same file**. Across files, the edge is silently dropped — not emitted as `INFERRED`, not reported as ambiguous, just absent.

This is the Pascal/Delphi sibling of #1698 (Kotlin `object` singletons) and the same general category as #1313 (Go `pkg.Func()`), #2135 (Rust), #2348 (C++), #2366 (Python module-qualified calls). A global `var instance: TClass;` + `instance.Method(...)` from another unit is an extremely common architecture in real-world Delphi codebases (a shared "main module" or "core" singleton exposed to every form unit), so this gap silently empties out a large share of "who calls X" queries on that kind of project.

### Root cause (traced in `graphify/extractors/pascal.py`)

1. Call extraction discards the receiver entirely:
   ```python
   # walk_calls(), ~line 668
   callee_text = _read(child).split(".")[-1]
   ```
   `mm.ServerReport` becomes just `serverreport` — the fact that it was a qualified call, and what it was qualified *by*, is lost before resolution even starts.

2. `_resolve_pascal_callee_factory` (~line 154) resolves `serverreport` in 4 tiers: caller's own class → ancestor chain → file-level free functions → an unambiguous global-by-name match — but every tier is built from `records`, which only contains procedures/functions declared **in the file currently being extracted** (each file is extracted independently, per the module's own docstring). `ServerReport` lives in a different unit, so all 4 tiers miss and the call becomes an unresolved `raw_call`.

3. The only cross-file resolver, `graphify/pascal_resolution.py`, explicitly walks `inherits` chains only (`resolve_pascal_inherited_calls`) — by design, per its docstring, it does not attempt a global by-name match across the corpus. A call via an unrelated global singleton instance isn't an inheritance relationship, so this resolver doesn't apply either. The `raw_call` is never picked up by anything and no edge is ever emitted.

### Reproducer (2 files)

`MainModule.pas`
```pascal
unit MainModule;

interface

type
  Tmm = class
    procedure ServerReport(const AReportName: string);
  end;

var
  mm: Tmm;

implementation

procedure Tmm.ServerReport(const AReportName: string);
begin
  // ...
end;

end.
```

`Acermest.pas`
```pascal
unit Acermest;

interface

uses
  MainModule;

type
  TfrmAcermest = class
    procedure btnPrintClick;
  end;

implementation

procedure TfrmAcermest.btnPrintClick;
begin
  mm.ServerReport('AcertosEstoque');
end;

end.
```

### Current behavior

`btnprintclick` has no `calls` edge to `tmm_serverreport` at all — the file-scoped 4-tier resolver in `pascal.py` can't see across the unit boundary, and `pascal_resolution.py` only walks `inherits`, so the call is dropped without a trace.

### Expected

```
acermest_tfrmacermest_btnprintclick --calls--> mainmodule_tmm_serverreport   [EXTRACTED or INFERRED]
```

The receiver (`mm`) is a module-level `var` of a statically known type (`Tmm`) declared in the same unit as the callee — no type inference needed, just tracking `var name: TypeName;` declarations alongside class declarations and matching the call's receiver identifier against them across files. Structurally this seems close in spirit to how #1698 was closed for Kotlin `object` singletons.

### Real-world impact

Found while indexing a ~640-file production Delphi/UniGUI ERP. The project's central reporting entry point (`Tmm.ServerReport` in `MainModule.pas`) is called from at least 5 different form units (verified independently by `grep`) but shows **zero incoming `calls` edges** in the graph — `graphify query "who calls ServerReport"` returns only the definition, nothing else. `var singleton: TClass;` + `singleton.Method()` from other units is the dominant call pattern across the whole codebase (data access, error logging, report generation all route through 2-3 such singletons), so this isn't an edge case for Delphi/Pascal projects — it's close to the majority of cross-unit calls.

abhay-codes07:
Closes #3101.

## The problem

`var mm: TMainModule;` declared in one unit's interface, `mm.ServerReport(...)` called from another — the standard Delphi "shared main module" shape. The edge was emitted only when caller and callee were in the same file; across files it was silently absent. Two causes, both traced in the issue:

1. the per-file pass discarded the receiver before resolution started (`callee_text.split(".")[-1]`), so `mm.ServerReport` became just `serverreport`;
2. the cross-file resolver (`pascal_resolution`) only walked the **caller's** `inherits` chain — the right thing for an unqualified inherited call, and nothing at all for a call on a global.

A third, unreported: the bare qualified statement form `om.Flush;` (no parentheses) is `statement → exprDot` in tree-sitter-pascal, not `exprCall`, and the walk only handled a lone `identifier` there — so it was never seen as a call at all.

## The change

**Per-file (both extractors, tree-sitter and the regex fallback):**
- an unresolved qualified call keeps its receiver on the raw call (`receiver: "mm"`); `Self.`/`inherited` are not receivers and unqualified calls are reported exactly as before;
- the unit's interface-section `var`/`threadvar` declarations are exported as `pascal_globals` (`{"mm": "tmainmodule"}`) — only interface vars, because they are the only variables another unit can name; implementation-section and procedure-local vars are not collected, and a name declared twice with different types is dropped;
- the tree-sitter walk treats a lone `exprDot` statement as a qualified call.

**Corpus (`pascal_resolution`):** receiver → declared type (unique across every unit's globals) → the one class of that name → its one method of that name, emitting an `EXTRACTED` `calls` edge at the call site (a variable's declared static type is deterministic, matching the confidence the inherited-call pass uses). Ambiguity at any step — a global declared with two types, two same-named classes, two same-named methods — yields no edge rather than a guess: the same god-node guard the inherited pass applies. A qualified call no longer falls back to the caller's ancestor chain; it names its receiver, so it is not an inherited call.

The `calls` edges from the shared direct-call pass, and the raw-call plumbing through the id-remap passes, are untouched — `pascal_globals` holds type *names*, never ids.

## Tests

`tests/test_pascal_singleton_calls.py` (20 tests; 8 fail with the fix reverted) with a three-unit fixture under `tests/fixtures/pascal_singleton/` (same static-fixture convention as `test_pascal_resolution.py`): receiver parsing; interface-globals collection (implementation/local vars excluded, conflicting names dropped, programs export nothing); both extractors reporting the receiver and exporting globals; and at corpus level the edge existing across files, landing on the *right* class when two classes share a method name (`mm.ServerReport` → `TMainModule`, never `TOtherModule`), carrying the call site with `EXTRACTED` confidence, an unqualified unresolvable call still producing nothing, an ambiguous receiver producing nothing, and a qualified call never binding through the caller's ancestors.

`test_pascal`, `test_pascal_call_scoping` and `test_pascal_resolution` are unchanged (72 passed together); the full suite matches the `v8` baseline.

### DIFF
--- a/graphify/extractors/pascal.py
+++ b/graphify/extractors/pascal.py
@@ -151,6 +151,75 @@ def _pascal_find_body(text: str, start: int) -> tuple[int, int]:
                 return (body_start, tok.start())
     return (body_start, len(text))
 
+
+# ---------------------------------------------------------------------------
+# Global singleton receivers (#3101)
+# ---------------------------------------------------------------------------
+#
+# `var mm: TMainModule;` in a unit's interface section, then `mm.ServerReport`
+# from any other unit, is the standard Delphi "shared main module" shape. The
+# per-file pass cannot bind it (the type lives in another file), so the call
+# is reported as a raw call carrying its RECEIVER, and the interface-section
+# globals are reported beside it; graphify.pascal_resolution joins the two
+# across the corpus. Only interface-section vars are collected: they are the
+# only ones another unit can see.
+
+_PAS_VAR_BLOCK_RE = re.compile(
+    r"^[ \t]*(?:var|threadvar)\b(.*?)(?=^[ \t]*(?:const|type|var|threadvar|procedure|"
+    r"function|constructor|destructor|class|implementation|begin|end|uses|resourcestring)\b|\Z)",
+    re.IGNORECASE | re.MULTILINE | re.DOTALL,
+)
+_PAS_VAR_DECL_RE = re.compile(
+    r"^[ \t]*([A-Za-z_][\w]*(?:[ \t]*,[ \t]*[A-Za-z_][\w]*)*)[ \t]*:[ \t]*([A-Za-z_][\w.]*)",
+    re.MULTILINE,
+)
+_PAS_RECEIVER_SKIP = frozenset({"self", "inherited", "result"})
+
+
+def _pascal_interface_globals(text: str) -> dict[str, str]:
+    """``{var_name_lower: type_name_lower}`` for every variable declared in
+    the unit's interface-section ``var``/``threadvar`` blocks.
+
+    ``text`` must already be comment-stripped. A name declared twice with
+    different types (it happens in generated units) is dropped rather than
+    guessed. Files without an interface section (programs, includes) export
+    nothing: nothing outside them can name their variables.
+    """
+    iface, _off, _impl, _impl_off = _pascal_split_sections(text)
+    if not iface:
+        return {}
+    found: dict[str, str] = {}
+    conflicting: set[str] = set()
+    for block in _PAS_VAR_BLOCK_RE.finditer(iface):
+        for decl in _PAS_VAR_DECL_RE.finditer(block.group(1)):
+            type_lower = decl.group(2).split(".")[-1].lower()
+            for raw_name in decl.group(1).split(","):
+                name = raw_name.strip().lower()
+                if not name:
+                    continue
+                prev = found.get(name)
+                if prev is not None and prev != type_lower:
+                    conflicting.add(name)
+                found[name] = type_lower
+    for name in conflicting:
+        found.pop(name, None)
+    return found
+
+
+def _pascal_call_parts(callee_text: str) -> tuple[str, str | None]:
+    """Split ``mm.ServerReport`` into ``("serverreport", "mm")``; an
+    unqualified call, or one qualified by ``Self``/``inherited``, has no
+    receiver worth reporting."""
+    parts = [p.strip() for p in callee_text.split(".") if p.strip()]
+    if not parts:
+        return "", None
+    name_lower = parts[-1].lower()
+    receiver = parts[-2].lower() if len(parts) >= 2 else None
+    if receiver in _PAS_RECEIVER_SKIP:
+        receiver = None
+    return name_lower, receiver
+
+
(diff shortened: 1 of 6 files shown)


Vinay2334:
Bro, how are you fixing issues so fast. I picked 2 issues today and you already closed them as I was understanding them.

abhay-codes07:
Good catch — this was a real false-edge bug, fixed in 723edca (and rebased onto 0.9.59). `_emit_or_report` ran the caller-scoped, receiver-blind `resolve_callee` before parking, so `om.Flush` bound to the *caller's* own `Flush` whenever the name collided in the caller's class/ancestor/module scope. A qualified call (explicit, non-Self/inherited receiver) now skips that resolver entirely and parks for the receiver-aware pass in `pascal_resolution`, which types the receiver through the global-singleton table before matching the method. Verified with a regression test: `TFoo.Bar` calling `om.Flush` where TFoo also declares Flush now binds to `TBar.Flush` (om's type), not `TFoo.Flush`; the test fails without the fix.

PR #3122, opened 2026-08-26
feat(apex): parse Apex with tree-sitter, add calls edges

rsoesemann:
The Apex extractor has never produced `calls` edges, and its docstring says why: no tree-sitter grammar on PyPI. That comment is out of date — [aheber/tree-sitter-sfapex](https://github.com/aheber/tree-sitter-sfapex) exists, is MIT, and was last touched a week ago.

It happens to use the same node type and field names as tree-sitter-java for declarations and calls, so the shared engine walk drives it from a config with no special cases. On [my-org-butler](https://github.com/aquivalabs/my-org-butler), 55 Apex files: **592 calls edges where there were 0**, across 225 distinct targets, the busiest with 17 incoming — so nothing turned into a god-node.

`AgentMemory` and what reaches it, in the graph graphify builds for that repo:

AgentMemory call structure

The whole repo. Clustering finds 38 communities where the same corpus without calls edges fell into 80 mostly disconnected stars — and because the call structure gives each community a real hub, the hub labels come out as class names rather than `Community N`:

my-org-butler communities

Not the Java grammar, though — that was the tempting shortcut and it is wrong. Parsing those same 55 files with tree-sitter-java yields 740 ERROR nodes across 53 of them, and it doesn't raise, it just returns a broken tree. The sfapex grammar parses all 55 with zero errors, including `??` (Spring '24), user-mode DML and `WITH USER_MODE` (Spring '23).

The regex extractor stays as the fallback, like Pascal's. Without the extra, Apex extracts exactly as before, minus calls. The full suite passes both ways: 4966 with the grammar, 4951 and 63 skipped without.

**The dependency is the part worth arguing about.** The grammar publishes bindings for npm, Cargo and WASM but nothing for Python, so `tree-sitter-language-pack` is the only route. It ships proper abi3 wheels, but it downloads the grammar's shared library on first use into a user cache — a fresh install needs network access once. That is weaker than the pinned wheels every other grammar here gets, and I would rather say so than have you find it. If it is a blocker the honest fix is a real `tree-sitter-apex` wheel, and I have asked upstream for Python bindings.

Three things the generic walk cannot know are handled in the extractor rather than by widening the Java-gated branches in the engine: `extends`/`implements`, the SObject a trigger fires on and the one behind a SOQL `FROM`, and DML statements. `new Other()` too, since its callee sits in the `type` field rather than `name`.

Two filters, both chosen from what real code does rather than guessed. `new List<Account>()` appears in nearly every method and built a `List` node collecting 30 edges. And `rows.add('x')` bound to whichever class happens to declare `add()` — 34 of 41 edges onto one class were list appends from elsewhere. Note the asymmetry: a name declared by *many* classes is already safe, because the resolver refuses to bind an ambiguous name. That is why `execute` is deliberately not filtered — every invocable class declares it, so it only ever resolves within a file.

One behaviour change to call out: the AST path does not reproduce 11 things the regex path emitted. I classified every one — six were `new X(...)` read as a method declaration, four were the word "Merge" in a comment, one was `Database` truncated from `Database.Batchable`. All regex false positives, none a real loss. No existing test needed changing.

Tests: 14 in `tests/test_apex_calls.py`, checked by mutation — ten plausible wrong implementations, each caught by at least one test. Negative controls: a commented-out call, a local variable named after a method, DML in prose, a collection constructor, and a collection method that must not bind across files.

Shared core is touched in two small places: one optional `LanguageConfig` field for grammars only reachable through the pack, and the loader branch that uses it. No existing language behaves differently.

Independent of #3106 and #3117.

### DIFF
--- a/graphify/extractors/apex.py
+++ b/graphify/extractors/apex.py
@@ -61,6 +61,12 @@ def _extract_apex_ast(path: Path) -> dict | None:
         root = parser.parse(source).root_node
     except Exception:
         return None
+    # tree-sitter is error-tolerant: it returns a tree with ERROR nodes instead of
+    # raising, so a file the grammar cannot handle would otherwise yield a
+    # confidently wrong AST. Hand those to the regex path, which degrades
+    # predictably rather than inventing structure.
+    if root.has_error:
+        return None
     _add_apex_specifics(path, root, source, result)
     raw_calls = result.get("raw_calls")
     if raw_calls:
@@ -284,8 +290,13 @@ def _extract_apex_regex(path: Path) -> dict:
 
 _DML_TYPES = frozenset({"insert", "update", "delete", "upsert", "merge", "undelete"})
 
-# Annotations that make a method an entry point reachable from outside Apex.
-_ENTRY_POINT_ANNOTATIONS = frozenset({"auraenabled", "invocablemethod"})
+# Annotations that make a method an entry point reachable from outside Apex —
+# Lightning, Flow, and the Apex REST verbs. Without the REST ones a
+# @RestResource class looks unreachable, since nothing in the corpus calls it.
+_ENTRY_POINT_ANNOTATIONS = frozenset({
+    "auraenabled", "invocablemethod", "remoteaction",
+    "httpget", "httppost", "httpput", "httpdelete", "httppatch",
+})
 
 # Collection and primitive constructors. `new List<Account>()` appears in nearly
 # every method, so treating it as a call site builds a god-node that collects an
--- a/graphify/extractors/apex.py
+++ b/graphify/extractors/apex.py
@@ -1,14 +1,79 @@
-"""Apex extractor. Moved verbatim from graphify/extract.py."""
+"""Apex extractor: tree-sitter when the grammar is installed, regex otherwise."""
 from __future__ import annotations
 
 
 from pathlib import Path
+
 from graphify.extractors.base import _file_stem, _make_id
+from graphify.extractors.models import LanguageConfig
+
+# The sfapex grammar (aheber/tree-sitter-sfapex) uses the same node type and field
+# names as tree-sitter-java for declarations and calls, so the shared engine walk
+# drives it unchanged. Apex has no import statements, hence no import_types.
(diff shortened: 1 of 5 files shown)


rsoesemann:
Reproduced all five. Four don't hold; the fifth is a message, not behaviour.

**Doesn't hold (reported twice, both high): the builtin-method filter drops legitimate same-file calls.** It doesn't. For a class declaring `get`, `post` and `add`, both `this.add('GET', s)` and bare `add('POST', s)` produce `.get() -> .add()` and `.post() -> .add()`. Same-file calls are resolved into `edges` by the engine walk before the filter runs; the filter only touches `raw_calls`, which is exclusively the cross-file path. `test_a_class_still_calls_its_own_collection_named_method` pins this.

**Doesn't hold: DML statements never produce edges.** They do — `Svc --uses--> insert` and `Svc --uses--> update` for a method containing `insert a; update a;`. Covered by `test_soql_and_dml_survive_the_ast_path` and the pre-existing `test_apex_dml_uses_edge`.

**Doesn't hold: unterminated test file prevents pytest collection (`tests/test_apex_calls.py:331`).** The file compiles under `py_compile`, ends with a newline, and `--collect-only` reports 20 tests. The full suite runs 4969 passed.

**Half holds: the language-pack path bypasses the TypeError guard.** The loader sits inside the same `try`, so a load failure is caught and returned as an error dict, and `_extract_apex_ast` falls back to regex — I verified that by making `get_language` raise. What is genuinely wrong is the *message*: a `TypeError` from the pack hits the version-mismatch branch and advises `pip install --upgrade tree-sitter tree-sitter-languages`, which is misleading. Happy to make the hint name the module that actually failed if you'd like that in this PR.

Two findings from the previous round were real and are fixed in `ffa8699`: tree-sitter returns a tree containing `ERROR` nodes rather than raising, so a broken parse was being accepted instead of falling back (`root.has_error` now triggers the fallback — 0 of 55 files in the reference corpus hit it), and the Apex REST verbs are now entry points alongside `@AuraEnabled`/`@InvocableMethod`.

PR #3135, opened 2026-08-27
fix(ts): resolve new_expression constructor calls

### LINKED ISSUE #3116: TS/JS: `new Foo()` emits no edge — `call_function_field="function"` does not match `new_expression`, whose callee field is `constructor`
### Summary

In TS/JS, `new Foo()` emits no edge, in the same file or across files, even though
`new_expression` **is** listed in `_TS_CONFIG.call_types`. Static calls and plain function
calls in the same method body resolve normally, so `walk_calls` itself is fine — only the
construction site is dropped.

Same user-visible symptom as #2997 (C#, fixed in #2998) and the PHP gap I just filed
(#3115), but a different root cause: here the node type is dispatched, its callee field is
just never read.

### Repro (graphify 0.9.50, tree_sitter_typescript 0.23.2)

One file, `graphify extract . --code-only`:

```ts
export class Foo { constructor(public x: number) {} static make(): Foo { return new Foo(0) } }
export function helper(n: number): number { return n + 1 }
export class Caller {
  run(): void {
    const a = new Foo(1)   // new
    const b = Foo.make()   // static method call
    const c = helper(2)    // plain function call
  }
}
```

Edges emitted from `Caller.run` (excluding `contains`/`method`):

| call site | edge |
| --- | --- |
| `new Foo(1)` | **none** |
| `Foo.make()` | `run --calls--> Foo.make` |
| `helper(2)` | `run --calls--> helper` |

Cross-file behaves identically: with `Foo` in `models.ts` and a named
`import { Foo } from './models'`, the import edge is emitted, but `new Foo(1)` in the
method body still produces nothing. For contrast, the equivalent Python file
(`a = Foo(1)`) emits both `calls` and `uses` — instantiation is a call there, so the
generic path covers it.

### Likely root cause

In tree-sitter-typescript the two node types expose the callee under different field names:

```
new_expression  : function=None       constructor=identifier  arguments=arguments
call_expression : function=identifier constructor=None        arguments=arguments
```

`_TS_CONFIG` (and `_TSX_CONFIG`, which inherits it) sets:

```python
call_types=frozenset({"call_expression", "new_expression"}),
call_function_field="function",
```

So `child_by_field_name("function")` returns `None` for every `new_expression`, no callee
name is extracted, and the entry in `call_types` is effectively inert. #708 added
`new_expression` extraction, but the field name looks like it was never adjusted.

C# solves the equivalent problem with a node-type special case in `walk_calls`
(`engine.py`, `elif config.ts_module == "tree_sitter_c_sharp" and node.type ==
"object_creation_expression"`). A per-node-type callee field, or the same kind of special
case reading `constructor`, would cover TS/JS/TSX.

### Why it matters

"What breaks if I change X" is the main reason to query a graph instead of grepping, and in
class-based TS a type that is only ever constructed at its call sites now looks unused.

hopstreax:
## Summary

Fixes #3116.

TypeScript/JavaScript `new_expression` nodes were already included in `call_types`, but `walk_calls` looked for the callee using `call_function_field="function"`.

Tree-sitter JS/TS exposes the constructor of `new_expression` through the `constructor` field, so `new Foo()` was silently dropped while normal function and static calls worked.

This adds a narrow fallback to read `constructor` when the generic callee lookup returns `None` for a `new_expression`.

## Changes

- Resolve `new Foo()` as a `calls` edge.
- Preserve existing member-expression resolution for `new pkg.Foo()`.
- Apply consistently to JS, TS, and TSX.
- Add regression coverage for:
  - TypeScript in-file constructor calls
  - TypeScript cross-file constructor calls
  - JavaScript constructor calls
  - TSX constructor calls
  - member constructor `new pkg.Foo()`

## Validation

- `5 passed` — new regression suite
- `44 passed` — related language/member-call suites
- `201 passed, 4 skipped` — `tests/test_extract.py`
- The 4 skips are the pre-existing Windows temp-path failure.

No unrelated production code was changed.

### DIFF
--- a/graphify/extractors/engine.py
+++ b/graphify/extractors/engine.py
@@ -5388,8 +5388,10 @@ def _extract_generic(
                         # unique-but-wrong one (#3078).
                         member_receiver = _ruby_const_full_name(recv, source) or None
             else:
-                # Generic: get callee from call_function_field
+                # Generic: get callee from call_function_field (or constructor on new_expression)
                 func_node = node.child_by_field_name(config.call_function_field) if config.call_function_field else None
+                if func_node is None and node.type == "new_expression":
+                    func_node = node.child_by_field_name("constructor")
                 if func_node:
                     if func_node.type == "identifier":
                         callee_name = _read_text(func_node, source)
--- a/tests/test_ts_new_expression_calls.py
+++ b/tests/test_ts_new_expression_calls.py
@@ -1,0 +1,100 @@
+"""TS/JS/TSX `new Foo(...)` constructor calls emit `calls` edges (#3116).
+
+In tree-sitter JS/TS, `new_expression` exposes its callee under the `constructor`
+field rather than `function`. The generic path in `walk_calls` previously queried
+only `call_function_field="function"`, dropping constructor calls.
+"""
+from __future__ import annotations
+
+from pathlib import Path
+
+from graphify.extract import extract, extract_js
+
+
+def _calls(tmp_path: Path, files: dict[str, str]):
+    for name, body in files.items():
+        p = tmp_path / name
+        p.parent.mkdir(parents=True, exist_ok=True)
+        p.write_text(body, encoding="utf-8")
+    r = extract([tmp_path / n for n in files],
+                cache_root=tmp_path / "graphify-out", parallel=False)
+    lbl = {n["id"]: n["label"] for n in r["nodes"]}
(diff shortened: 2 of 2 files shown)


safishamsi:
Shipped in v0.9.51 via authorship-preserving cherry-pick so you keep contributor-graph credit. Thanks @hopstreax! Release: https://github.com/Graphify-Labs/graphify/releases/tag/v0.9.51

PR #3141, opened 2026-08-27
fix(extract): fully rescan code with --force --code-only

### LINKED ISSUE #3125: `extract --force` does not invalidate the module-resolution cache — a tsconfig `paths` fix measures as a total no-op
## Summary

`graphify extract --force` re-reads every source file but keeps the cached module-resolution results. After changing a tsconfig `paths` mapping, the rebuilt graph silently retains the old, unresolved edges — so a **correct** configuration fix measures as changing nothing at all.

Only deleting `graphify-out/` picks the change up.

The docs describe `--force` as:

> full re-scan and re-dispatch: skip the incremental manifest gate and semantic cache reads

which reads as covering exactly this case.

## Impact

This cost me a wrong conclusion twice on the same repo, and nearly a reverted fix.

Working on a 22-workspace pnpm monorepo (see #462), I added a wildcard `paths` mapping so `@app-gether-life/utils/*` would resolve, then ran:

```bash
graphify extract . --code-only --force
```

`apps/web → packages/utils` import edges: **0**. Identical to before the change. I concluded the mapping didn't work on this repo and was about to revert it.

Then, same config, no source edits:

```bash
rm -rf graphify-out
```

`apps/web → packages/utils`: **98 edges**, and `graphify affected` on a symbol in that package went from returning one file (its own sibling test) to naming its real consumers.

The only difference was the deleted cache.

The failure mode is the bad one: not an error, not a warning, just a plausible-looking graph built from stale resolutions. Anyone iterating on `paths` or on workspace layout to improve resolution will conclude their change had no effect.

## Repro

```bash
# two-package workspace, consumer imports "@scope/pkg-a/helpers"
# start with NO paths mapping in apps/app-b/tsconfig.json
graphify affected "sharedHelper()"        # no edge, as expected

# add the mapping:
#   "paths": { "@scope/pkg-a/*": ["../../packages/pkg-a/src/*"] }
graphify extract . --code-only --force
graphify affected "sharedHelper()"        # STILL no edge  <-- bug

rm -rf graphify-out
graphify affected "sharedHelper()"        # edge appears
```

## Expected

`--force` invalidates cached resolutions too, since a config change that alters resolution is precisely when someone reaches for `--force`.

## Suggested fixes

Any of these would remove the trap:

1. Have `--force` clear the resolution cache along with the manifest gate.
2. Key the resolution cache on the resolved tsconfig (path + mtime, or a hash of the effective `paths`/`baseUrl`/`moduleSuffixes`), so it self-invalidates.
3. Failing either, note it in `--force`'s help text and suggest deleting `graphify-out/` after config changes.

(2) seems best — it also covers `extract` without `--force`, which has the same staleness after a config edit.

## Environment

- graphify 0.9.50
- Python 3.14.5, macOS arm64
- TypeScript 6.0.3, pnpm workspace monorepo

hopstreax:
## Summary

Fixes #3125.

`graphify extract --force --code-only` could leave TypeScript/JavaScript
module-resolution results stale after a `tsconfig.json` change.

The problem was not a persistent module-resolution cache. `--force
--code-only` was re-enabling incremental file detection in order to preserve
the existing semantic layer. As a result, unchanged code files were skipped
entirely and their previous AST/resolution results remained in the graph.

This change separates those two concerns:

- `--force --code-only` now performs a full code/AST scan.
- The existing semantic layer is still preserved through the existing merge
  path.
- Normal incremental extraction behavior is unchanged.
- Normal `--force` behavior is unchanged.

## Root cause

Before this change, `--force --code-only` explicitly set:

```python
incremental_mode = True
````

That caused the incremental detector to omit unchanged source files.

For example, after changing a TypeScript `paths` mapping:

```text
tsconfig.json changed
        ↓
incremental detection
        ↓
unchanged .ts files skipped
        ↓
old resolution/AST edges retained
```

Deleting `graphify-out/` happened to fix the problem because it forced a
clean extraction.

## Fix

The extraction path now distinguishes:

* whether files should be scanned incrementally, and
* whether the existing graph should be merged/preserved.

`--force --code-only` keeps incremental detection disabled so all code files
are rescanned, while `merge_existing_graph` ensures the existing semantic
layer is preserved.

Stale graph sources are also identified during the full scan so deleted or
excluded files can still be pruned correctly.

## Regression test

Added coverage for the complete scenario:

1. Extract a project with an unresolved TypeScript alias.
2. Change only `tsconfig.json` to add the corresponding `paths` mapping.
3. Run `extract --code-only --force`.
4. Verify the import edge is now resolved.
5. Verify an existing semantic node remains in the graph.

## Validation

* `tests/test_extract_code_only_cli.py`: **13 passed**
* `tests/test_js_import_resolution.py tests/test_build.py tests/test_cache.py`:
  **219 passed, 8 skipped**
* Focused regression test: **1 passed**
* `git diff --check`: clean

## Scope

No changes were made to the module-resolution implementation, cache layer,
or incremental detector. The fix is isolated to the extraction/merge
semantics for `--force --code-only`.

### DIFF
--- a/graphify/cli.py
+++ b/graphify/cli.py
@@ -3252,16 +3252,17 @@ def dispatch_command(cmd: str) -> None:
         # --force: full scan, not the manifest-gated incremental diff — a warm
         # unchanged tree would otherwise dispatch zero files (#1894).
         incremental_mode = incremental_mode and not force
-        # #2923: --force --code-only must NOT drop the existing semantic layer.
-        # The AST pass is fully replaced (full re-scan, semantic cache reads
-        # skipped), but the semantic pass is itself skipped entirely, so
-        # doc/paper/image nodes from the existing graph carry forward via the
-        # incremental merge (build_merge / merge_raw_extraction keep them
-        # because no new semantic-tier sources are dispatched). Without this,
-        # a single --code-only --force silently erases every doc/paper/image
-        # node plus its connected hyperedges.
+        # #2923/#3125: --force --code-only must NOT drop the existing semantic layer.
+        # The AST pass is fully replaced (full code re-scan, AST extraction on all
+        # code files), but the semantic pass is skipped, so doc/paper/image nodes
+        # from the existing graph carry forward via the merge (build_merge /
+        # merge_raw_extraction keep them because no new semantic-tier sources
+        # are dispatched). We do NOT set incremental_mode = True here because that
+        # would run _detect_incremental and drop unchanged code files from the AST
+        # pass; instead we keep incremental_mode = False so all code files are
+        # scanned, while merge_existing_graph below ensures build_merge still runs.
+        merge_existing_graph = incremental_mode or (code_only and existing_graph_path.exists())
         if force and code_only and existing_graph_path.exists():
-            incremental_mode = True
             print(
                 "[graphify extract] --force --code-only: full AST re-scan, "
                 "existing semantic layer preserved (no semantic pass this run)"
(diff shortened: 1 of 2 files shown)


safishamsi:
Shipped in v0.9.51 via authorship-preserving cherry-pick so you keep contributor-graph credit. Thanks @hopstreax! Release: https://github.com/Graphify-Labs/graphify/releases/tag/v0.9.51

Issue #3129, opened 2026-08-27
install writes a machine-absolute binary path into repo-tracked .codex/hooks.json (breaks other machines; emit PATH-based command)


yotamleo:
## Summary

`graphify install` (0.9.50, Windows, uv tool install) writes a **machine-absolute binary path** into the repo-tracked Codex hooks file `.codex/hooks.json`:

```json
{
  "hooks": [
    { "command": "C:/Users/<user>/.local/bin/graphify.EXE hook-check", "type": "command" }
  ],
  "matcher": "Bash"
}
```

## Why this is a problem

`.codex/hooks.json` is a **repo-tracked, shared** file (unlike per-user config). Committing the installer's output therefore:

1. **Breaks every other machine/collaborator** — the absolute path (`C:/Users/<user>/...`, the uv tool shim location) does not exist elsewhere; on machines without graphify the hook errors on every Bash tool call in Codex sessions.
2. **Churns the file per machine** — each machine that runs `graphify install` rewrites the entry with *its* local path, producing endless diffs in a shared enforcement file.
3. There is no fail-open guard: the command is invoked directly, so a missing binary is a hard hook failure rather than a no-op.

## Suggested fix

- Emit a **PATH-based** command (`graphify hook-check`) — the uv shim directory is on PATH in the environments where install succeeds — or make the command template configurable.
- Optionally wrap with an existence guard so machines without graphify fail open (skip) instead of erroring.

## Environment

- graphify 0.9.50 (`graphifyy` from PyPI, `uv tool install --with mcp`)
- Windows 11, Codex CLI 0.149.1
- Observed after `graphify install` refreshed harness integration files (`.codex/hooks.json`, `AGENTS.md`, `GEMINI.md`)

davidbhoward:
This is the same thing for Claude Code. 

`graphify install` The only debate I think maintainers might have here with this is that the installation is not for a project. It is not really expected that other people will also be maintaining that same location. Whereas, if you use the `--project`, you will get this same problem but with the expectation that others will also be contributing to that project.

safishamsi:
[...]

Issue #2164, opened 2026-07-24
codex/aider/hermes install writes AGENTS.md to cwd instead of user config dir; --project is a no-op


Second-Summit-marketing:
## Affected version

- `graphifyy` 0.9.25 (uv tool install)
- `graphify/install.py`, `dispatch_install_cli`, ~line 2171-2177
- Windows 11, but the code path is platform-independent

## Summary

For the always-on platforms (`aider`, `codex`, `opencode`, `claw`, `droid`, `trae`, `trae-cn`, `hermes`), the user-scope install and the `--project` install call the **same function with the same argument**:

```python
elif cmd in ("aider", "codex", "opencode", "claw", "droid", "trae", "trae-cn", "hermes"):
    subcmd = sys.argv[2] if len(sys.argv) > 2 else ""
    if subcmd == "install":
        if "--project" in sys.argv[3:]:
            _project_install(cmd, Path("."))
        else:
            _agents_install(Path("."), cmd)      # <-- also Path(".")
```

`_agents_install` then does:

```python
target = (project_dir or Path(".")) / "AGENTS.md"
```

So `graphify codex install` — documented as the user-scope install — writes `AGENTS.md` into **whatever directory the shell happens to be in**, not the user config dir (`~/.codex/`). The `--project` flag is effectively a no-op for these platforms.

## Reproduce

```console
$ cd /some/non-writable/or/unrelated/dir
$ graphify codex install
...
  File ".../graphify/install.py", line 1421, in _agents_install
    target.write_text(new_content, encoding="utf-8")
PermissionError: [Errno 13] Permission denied: 'AGENTS.md'
```

Run it from a writable but unrelated directory and it succeeds — silently creating a stray `AGENTS.md` there:

```console
$ cd ~
$ graphify codex install
graphify section written to C:\Users\<user>\AGENTS.md      # not ~/.codex/AGENTS.md
```

Codex reads global instructions from `~/.codex/AGENTS.md`, so the user-scope install lands somewhere Codex never reads.

Note the contrast: `graphify claude install` correctly targets `~/.claude/skills/` and `~/.claude/CLAUDE.md` at user scope. Only the always-on branch has this problem.

## Expected

- `graphify <platform> install` (no `--project`) writes to that platform's **user config directory** — for Codex, `~/.codex/AGENTS.md`.
- `graphify <platform> install --project` writes to the current repo.
- Neither should ever depend on the shell's cwd for the user-scope case.

## Secondary effect

[...]

PR #3149, opened 2026-08-28
fix(install): emit a bare hook command for project-scoped installs (#3129)

davidbhoward:
## Summary

`graphify install --project` resolves the graphify executable to an absolute path from the
installing user's environment and writes that path into hook configuration that the installer
then instructs the user to commit. Those paths are invalid for every other contributor and for CI.

Refs #3129, which reports this for Codex, with the Claude Code occurrence noted in the thread.
That issue is filed against a plain `graphify install`, so this PR does not close it — it fixes
the `--project` case only, for the reason already discussed there: a user-profile install lands
somewhere nobody else is expected to share, whereas `--project` carries the explicit expectation
that others contribute to that checkout. See "Scope" below.

To be clear about what is and isn't at issue: `hook-guard` itself is sound. It handles the Grep
tool payload shape that defeats a naive command matcher, and its read mode does per-file staleness
detection. Path handling at install time is the only defect here.

## Reproduction

Clean `git init` repositories, no prior graphify configuration:

```bash
mkdir gfx-claude && cd gfx-claude && git init .
graphify install --project --platform claude
cat .claude/settings.json
```

```json
{
  "hooks": {
    "PreToolUse": [
      { "matcher": "Bash|Grep",
        "hooks": [ { "type": "command",
                     "command": "C:/Users/<user>/.local/bin/graphify.EXE hook-guard search" } ] },
      { "matcher": "Read|Glob",
        "hooks": [ { "type": "command",
                     "command": "C:/Users/<user>/.local/bin/graphify.EXE hook-guard read" } ] }
    ]
  }
}
```

The same run prints:

```
Project-scoped install. Add to version control:
  git add .claude/ CLAUDE.md
```

`graphify install --project --platform codex` writes the same shape into `.codex/hooks.json`
(`... graphify.EXE hook-check`), and `--platform gemini` into `.gemini/settings.json`
(`... graphify.EXE hook-guard gemini`).

## Why this is wrong

`--project` exists to produce a shared, version-controlled setup. That is precisely the case in
which an absolute path cannot be correct. A committed hook pointing into one developer's home
directory fails silently for everyone else: the configuration is present and syntactically valid,
so nothing looks broken, but every matching tool call fails to run the hook.

The drive letter and `.EXE` casing also make the value non-portable between two Windows checkouts,
not only across operating systems. And each machine that reinstalls rewrites the entry with its own
path, churning a shared file.

## Affected platforms

Three platforms write a hook command through the shared `_resolve_graphify_exe()`, so the defect is
in that shared layer rather than in any one platform's config writer:

| Platform | File | Command |
|---|---|---|
| claude | `.claude/settings.json` | `hook-guard search` / `hook-guard read` |
| codex | `.codex/hooks.json` | `hook-check` |
| gemini | `.gemini/settings.json` | `hook-guard gemini` |

Codex and Claude are both already noted on #3129; Gemini appears not to have been, and falls out of
the same shared resolver.

`codebuddy` shares `_claude_pretooluse_hooks()` but has no project-scoped hook path
(`graphify install --project --platform codebuddy` writes no `.codebuddy/settings.json`), so it is
unaffected by this change.

## The change

Project-scoped installs emit the bare command:

```json
"command": "graphify hook-guard search"
```

PATH resolves it, the way a committed config refers to `git` or `node`. `_resolve_graphify_exe()`
takes a `project` flag that is threaded through the three hook builders and set only at the
`--project` entry points in `_project_install()`.

This follows a convention the codebase already uses rather than introducing a new mechanism: the
git-hook layer resolves the binary at runtime with `GRAPHIFY_BIN=$(command -v graphify)`
(`graphify/hooks.py`), and `skill-agents.md` does the same with `which graphify`. Baking a path at
install time is the outlier.

**The user-profile install is unchanged.** That location is not one other people are expected to
share, and the resolved absolute path is what makes the hook work where the venv `Scripts/`
directory is not on PATH — the VS Code Codex extension on Windows, per the existing docstring. A
regression test pins that behaviour.

## Anticipated objection

*A bare command breaks when `graphify` is not on a contributor's PATH.*

That is not a regression. A missing binary and a wrong absolute path both fail the hook; the
difference is diagnosis. `graphify: command not found` names the problem, whereas a path into
another developer's home directory gives the reader nothing to work with and reads like a
deliberate configuration choice. The bare command fails more loudly and more usefully than what
ships today. This argument applies only to the committed case, which is why the user-profile path
is left alone.

## Scope

Deliberately narrow. Two further defects were observed in the same files and are **not** included
here, to keep this reviewable:

1. Generated config ends without a trailing newline (`\ No newline at end of file`), adding diff
   noise to committed JSON.
2. Installing over an existing, differing hook command replaces it with no diff, no warning, and no
   indication that a customised value was overwritten.

Happy to open those separately.

This PR also does not address the non-`--project` half of #3129. That report's reproduction is a
plain `graphify install`, and `graphify codex install` without the flag still writes an absolute
path into `.codex/hooks.json` in the working directory. Per the discussion on that issue, the
absolute path is defensible there — that install is not aimed at a location others are expected to
maintain — and the case is further entangled with #2164, which reports that these installs write
into the cwd rather than a user config directory at all. Left for a maintainer call rather than
silently bundled, which is why this is `Refs` and not `Fixes`.

Related: #3009 fixes the same class of defect for `.githooks/*` by moving the interpreter path into
a gitignored sidecar. That approach fits there because git hooks run in GUI clients and CI where
PATH is unreliable; these hooks are executed by the agent harness in the user's own shell, where
the documented install puts `graphify` on PATH.

## Tests

Added to `tests/test_install.py`, asserting on generated content rather than mocked calls:

- `test_project_install_hook_command_is_portable` (claude, codex, gemini) — no absolute path, no
  drive letter, no backslash, no `.EXE` casing, no leak of the installing user's path.
- `test_user_profile_install_still_resolves_absolute_path` (claude, codex, gemini) — the
  non-project install still emits the resolved path, so the fix does not regress the case where
  that is correct.
- `test_project_install_is_idempotent` (claude, codex, gemini) — installing twice leaves
  byte-identical config.
- `test_project_uninstall_removes_the_bare_hook_command` — the uninstall filter matches on
  `"graphify"`, so a bare command is still stripped.

Verified these fail without the fix: the three `..._is_portable` cases fail on clean `v8` and pass
with the change; the rest pass either way, as guards.

Full suite was run on Windows: 4935 passed, 31 failed, 104 skipped. All 31 failures are
pre-existing and unrelated (fifo/unix-socket, symlinks, shell metacharacters, ollama retries). I
verified this by running the same suite against clean `v8` and diffing the failure sets — identical,
zero introduced and zero fixed. Per the README, exact CI parity needs Linux; the Ubuntu matrix will
confirm. `ruff check` and the blocking `skillgen --check` / `--schema-singleton` jobs pass.

### DIFF
--- a/graphify/install.py
+++ b/graphify/install.py
@@ -285,12 +285,14 @@ def _print_project_git_add_hint(paths: list[Path]) -> None:
     print()
     print("Project-scoped install. Add to version control:")
     print(f"  git add {' '.join(unique)}")
-def _claude_pretooluse_hooks(strict: bool = False) -> "list[dict]":
+def _claude_pretooluse_hooks(strict: bool = False, project: bool = False) -> "list[dict]":
     """graphify's Claude/Codebuddy PreToolUse hooks, resolved at install time.
 
     The command invokes `graphify hook-guard <search|read>` via the absolute exe
-    path (`_resolve_graphify_exe`), so it parses under sh, cmd.exe and PowerShell
-    alike — this is the #522 fix, and mirrors the codex hook. Matchers are
+    path (`_resolve_graphify_exe`) — or, for a project-scoped install, via the
+    bare `graphify` command, since that config gets committed (#3129). Either
+    form parses under sh, cmd.exe and PowerShell alike — this is the #522 fix,
+    and mirrors the codex hook. Matchers are
     "Bash|Grep" and "Read|Glob" and the command always contains "graphify", so the
     existing install/uninstall filters find and replace both old bash hooks and
     these. "Grep" is in the search matcher because current Claude Code routes
@@ -301,7 +303,7 @@ def _claude_pretooluse_hooks(strict: bool = False) -> "list[dict]":
     first raw read per session (Claude Code only). The ``GRAPHIFY_HOOK_STRICT`` env
     var can force it on or off at runtime without a reinstall.
     """
-    exe = _resolve_graphify_exe()
+    exe = _resolve_graphify_exe(project=project)
     if " " in exe and not exe.startswith('"'):
         exe = f'"{exe}"'
     read_cmd = f"{exe} hook-guard read" + (" --strict" if strict else "")
@@ -689,9 +691,13 @@ _CLAUDE_MD_MARKER = "## graphify"
 _CODEBUDDY_MD_MARKER = "## graphify"
 _AGENTS_MD_MARKER = "## graphify"
 _GEMINI_MD_MARKER = "## graphify"
-def _gemini_hook() -> dict:
-    """Gemini CLI BeforeTool hook, resolved to a shell-agnostic `graphify` call."""
-    exe = _resolve_graphify_exe()
+def _gemini_hook(project: bool = False) -> dict:
+    """Gemini CLI BeforeTool hook, resolved to a shell-agnostic `graphify` call.
+
+    A project-scoped install emits the bare command, since .gemini/settings.json
+    is then committed and an installing machine's path is wrong there (#3129).
+    """
+    exe = _resolve_graphify_exe(project=project)
     if " " in exe and not exe.startswith('"'):
         exe = f'"{exe}"'
     return {
@@ -721,7 +727,7 @@ def gemini_install(project_dir: Path | None = None, *, project: bool = False) ->
 
     # Always re-install the Gemini hook so an older payload (e.g. pre-issue-#580
     # wording) is replaced on upgrade.
-    _install_gemini_hook(project_dir)
+    _install_gemini_hook(project_dir, project=project)
     if project:
         _print_project_git_add_hint([_project_scope_root(skill_dst, project_dir), project_dir / "GEMINI.md", project_dir / ".gemini"])
     print()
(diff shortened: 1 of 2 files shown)


davidbhoward:
On the bare graphify from PATH — that's the intent of the change, not a side effect of it. The alternative is what this PR is fixing: an absolute path baked in at install time, which is correct only on the machine that ran the installer and wrong in every other clone. PATH resolution is also already the convention in this repo. hooks.py resolves the binary at runtime with GRAPHIFY_BIN=$(command -v graphify), and skill-agents.md does the same with which graphify. So this brings the agent hooks in line with the git hooks rather than introducing a new assumption. A committed config that refers to graphify is the same trust model as one that refers to git or node.

If you'd rather pin it, the gitignored sidecar in #3009 looks like the natural precedent, but that seemed like a bigger design call than this PR should make on its own. Happy to follow up separately if you want that.

On the five coupling findings, I think those are artifacts rather than real regressions. This change only adds optional keyword arguments (project: bool = False) and threads them through calls that already existed, so it doesn't add any edges. I parsed install.py before and after and compared caller/callee pairs — 550 both ways, nothing added or removed. The functions listed as new (claude_install with 19 callers, and so on) had that coupling already; only their bodies changed, so I think they're being re-registered as new nodes.

Tests are still waiting on workflow approval. Locally the full suite shows no regressions against v8 — the same 31 pre-existing Windows-only failures with and without the change — and all five skillgen validators pass.

safishamsi:
Shipped in v0.9.52 via authorship-preserving cherry-pick so you keep contributor-graph credit. Thanks @davidbhoward! Release: https://github.com/Graphify-Labs/graphify/releases/tag/v0.9.52

PR #3157, opened 2026-08-28
fix(llm): bound extract retry cascade to one timeout budget per chunk

### LINKED ISSUE #3142: extract: a timed-out chunk re-pays the full timeout on every split, so one chunk can burn ~2.5h (claude-cli)
## Bug

With `--backend claude-cli`, a semantic chunk that exceeds the API timeout is split in half and retried recursively. **Each retry re-pays the full timeout**, so one chunk that cannot finish in the default 600 s can consume up to ~2.5 hours before the cascade gives up.

Observed on a 20-chunk run: chunks 1–2 completed, then **no progress for ~2 hours** on chunk 3.

```
[graphify extract] semantic extraction on 220 files via claude-cli...
[graphify extract] chunk 1/20 done
[graphify extract] chunk 2/20 done
[graphify] chunk of 17 timed out at depth 0 (TimeoutExpired); splitting in half and retrying
[graphify] chunk of 9 timed out at depth 1 (TimeoutExpired); splitting in half and retrying
[graphify] chunk of 4 timed out at depth 2 (TimeoutExpired); splitting in half and retrying
```

It reads as a hang. The giveaway that it was blocked rather than working: the process had **61 seconds of CPU across 2 hours of wall clock**.

## Arithmetic

- `_resolve_api_timeout(default: float = 600.0)` — 600 s per call (`llm.py:404`)
- On `TimeoutExpired`, the chunk splits and each half retries recursively to `max_depth=3` (`llm.py:~2391`), i.e. up to `2**3 = 8` leaf pieces
- Every attempt gets a **fresh** 600 s

Worst case for a single original chunk:

```
600 x (1 + 2 + 4 + 8) = 9,000 s = 2.5 hours
```

The recursion is correctly bounded — this is not an infinite loop. The problem is that the timeout is treated as a *per-attempt allowance* rather than a *budget for the whole subtree*, so bounding the depth does not bound the cost.

## Why `claude-cli` hits this routinely

Every semantic call spawns a whole nested `claude -p` agent session, which is far slower than an HTTP completion. A 600 s default that is comfortable for an API backend is easily exceeded here on a large chunk, so the cascade triggers in normal use rather than as an edge case.

## Workaround

`GRAPHIFY_API_TIMEOUT=1800` fixed it outright — the identical run then completed all 20 chunks and wrote the graph.

For anyone else hitting this: note that `GRAPHIFY_CLAUDE_CLI_PARALLEL` does **not** help. `llm.py:2614` already pins `claude-cli` to `max_concurrency = 1` unless that variable is exactly `"1"`, so setting it to any other value is a no-op. (I initially believed lowering concurrency was the fix; it was not.)

## Suggestions

1. **Scale the default timeout by backend.** `claude-cli` spawns an agent session per call; 600 s is a reasonable HTTP default but too low here.
2. **Make the timeout a subtree budget.** Track elapsed time across the split cascade and abandon once the original allowance is spent, instead of granting each of up to 15 attempts a fresh full timeout.
3. **Announce the cascade.** When a split starts, log the worst-case cost it has just committed to. Right now the only signal between "chunk 2/20 done" and completion is silence, which is indistinguishable from a hang.

## Environment

- graphify 0.9.50 (pipx), Python 3.14, Windows 11
- `graphify extract . --backend claude-cli`
- Corpus: ~2,200 markdown notes plus scripts; 315 files re-extracted, 975 cached

Ashfaqbs:
Fixes #3142.

## Bug

`_extract_with_adaptive_retry` bisects a chunk on `TimeoutExpired` (and other
recognized timeout errors) and retries each half. Every retry re-pays the
full `GRAPHIFY_API_TIMEOUT` from scratch, so a chunk that keeps timing out
across the whole cascade can burn up to `2**max_depth` full timeouts —
`600s x (1+2+4+8) = 9,000s` (~2.5h) at the default timeout and `max_depth=3`.
This is most visible on `--backend claude-cli`, where each call spawns a full
nested `claude -p` session and is far slower than a plain HTTP completion, so
the default 600s is exceeded in normal use rather than as an edge case. It
reads as a hang: the only signal between "chunk N/M done" and the eventual
give-up is silence.

The recursion depth was already capped, so this isn't unbounded — it's a
bounded cascade whose bound is much larger than anyone intends.

## Fix

Give the whole split subtree for one original chunk a single shared
wall-clock deadline (anchored to one `GRAPHIFY_API_TIMEOUT` allowance),
instead of handing every split a fresh full timeout:

- The top-level call computes `_deadline = time.monotonic() + GRAPHIFY_API_TIMEOUT`
  and threads it through every recursive call in the subtree (both the
  timeout-bisection path and the truncation/length-retry path, so a later
  timeout deeper in an already-truncating subtree still respects it).
- When a timeout fires and the shared deadline has already passed, the
  cascade gives up immediately instead of committing to another full-length
  attempt — same outcome as hitting `max_depth`, just reached by budget
  instead of depth.
- Non-timeout paths (context-exceeded, hollow responses) are untouched —
  this only changes what happens after a `TimeoutExpired`-class exception.

Worst case is now bounded by roughly one timeout's worth of wall time per
original chunk instead of `2**max_depth` timeouts.

## Testing

- Added `test_adaptive_retry_stops_when_timeout_budget_exhausted`: mocks
  `time.monotonic()` to simulate the deadline being exceeded right after the
  first timeout, and asserts the cascade gives up after exactly 1 attempt
  (previously it would have kept splitting and retrying).
- Ran the full `tests/test_llm_backends.py` suite (104 tests, including the
  4 existing timeout/adaptive-retry tests) — all pass unchanged, since those
  fakes complete instantly and never approach the default 600s budget.
- `ruff check` on both changed files — clean.
- Note: I could not run the full repo test suite end-to-end in my sandbox —
  a large, pre-existing batch of unrelated failures there trace back to a
  missing `tree-sitter` install in my environment, reproducible on a clean
  checkout of `v8` with no changes applied, not to this diff.

I used AI-assisted tooling to help navigate the codebase and draft this
change, but the diagnosis of the fix approach and the change itself were
reviewed and verified by me before opening this PR.

### DIFF
--- a/graphify/llm.py
+++ b/graphify/llm.py
@@ -2314,11 +2314,33 @@ def _extract_with_adaptive_retry(
     split inherits the same absolute deadline rather than each getting a fresh
     full timeout. Before #3142, a chunk that timed out at every depth re-paid
     the full timeout on each of up to ``2**max_depth`` attempts — up to 2.5h for
-    the default 600s timeout at max_depth=3. Once the shared deadline passes,
-    a further timeout gives up immediately instead of splitting again.
+    the default 600s timeout at max_depth=3. A timeout checks the budget
+    reactively, after the attempt, and gives up rather than splitting further
+    once it is spent.
+
+    A split whose *own* attempt has not yet started also checks the budget
+    proactively, before making that attempt: if an earlier sibling elsewhere
+    in the same subtree already exhausted the budget with a real timeout of
+    its own, this split is skipped outright rather than paying for a fresh
+    full-length attempt that the shared budget can no longer afford.
     """
     if _deadline is None:
         _deadline = time.monotonic() + _resolve_api_timeout()
+    elif _depth > 0 and time.monotonic() >= _deadline:
+        # An earlier split in this subtree already used up the shared budget
+        # with a real timeout of its own (the reactive check below, on a
+        # prior call in this recursion). Skip this attempt entirely rather
+        # than paying for one more full-length call the budget can no longer
+        # afford — without this, a sibling reached after the budget is spent
+        # would still start (and pay for) its own fresh timeout, since the
+        # reactive check only fires after that sibling's own attempt fails.
+        print(
+            f"[graphify] chunk of {len(chunk)} at depth {_depth}: the subtree's "
+            f"{_resolve_api_timeout():g}s timeout budget is spent — skipping "
+            f"this split rather than starting a fresh attempt",
+            file=sys.stderr,
+        )
+        return {"nodes": [], "edges": [], "hyperedges": [], "input_tokens": 0, "output_tokens": 0, "model": model, "finish_reason": "stop"}
 
     def _merge_two(left_units, right_units) -> dict:
         left = _extract_with_adaptive_retry(
--- a/graphify/llm.py
+++ b/graphify/llm.py
@@ -2265,6 +2265,7 @@ def _extract_with_adaptive_retry(
     _depth: int = 0,
     *,
     deep_mode: bool = False,
+    _deadline: float | None = None,
 ) -> dict:
     """Extract a chunk; if the response is truncated (`finish_reason="length"`),
     the API rejects the prompt as too large for the model's context window, or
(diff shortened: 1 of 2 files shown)


Ashfaqbs:
Addressed the review bot's 3 findings, in 7cc0f14.

**Findings #2/#3 ("expired shared deadline still allows sibling retry to start" / "right sibling extraction")** — real gap, fixed. The deadline was only checked *reactively*, after a timeout occurred. A sibling split reached after the budget was already spent elsewhere in the same subtree would still start (and pay for) its own fresh full-length attempt before its own reactive check caught up. Added a proactive check at the top of each split: if the deadline has already passed before this split's own attempt has even started, it's skipped outright.

**Finding #1 ("deadline check gives up before ever splitting when timeout exceeds budget")** — investigated, but this turns out to be inherent to the single-shared-budget design rather than a separately fixable defect. If the depth-0 attempt's own client-side timeout reaches the full shared budget (which is the common case — a timeout fires at essentially its full configured duration, and the client timeout and the deadline both derive from the same `GRAPHIFY_API_TIMEOUT`), there is, by definition, no budget left for any further split — regardless of whether that's decided immediately at depth 0 or one level deeper. I initially implemented a depth-0 exemption to let the first attempt always try splitting, but tracing it through showed the children's own (correct) proactive checks would immediately self-terminate anyway in that exact scenario, since real wall-clock time has necessarily already reached the deadline by the time they start — so the exemption changed nothing observable, just added complexity. Left as-is.

Added `test_adaptive_retry_skips_sibling_attempt_after_budget_exhausted_mid_tree`, verified RED against pre-fix code (right sibling still attempts and the test's clock mock runs out, proving the old code let it through) and GREEN with the fix. Full `test_llm_backends.py` suite (105 tests) and `ruff check` clean.

Ashfaqbs:
Pushed `4b37e4d` addressing the Graphify bot's advisory findings (deadline check skipping splits without a real timeout, at `llm.py:2326`/`2328`/`2378`/`2422` across the three review passes).

Root cause: the proactive budget check (`elif _depth > 0 and time.monotonic() >= _deadline`) was purely wall-clock-based — it didn't verify a real timeout had ever occurred in the subtree, so a context-exceeded or truncation (`length`) split could be skipped outright just because enough time had elapsed since the deadline was anchored, contradicting the docstring's own claim that only real timeouts should spend the budget.

Fix: thread a `_timeout_hit` flag (one-element list shared by reference across the recursion) that's only set `True` when a real timeout exception is actually observed, and gate the proactive skip on it. Added `test_adaptive_retry_deadline_only_skips_after_a_real_timeout`, which fails against the pre-fix code (confirmed locally) and passes with the fix. All 137 tests in `test_llm_backends.py`/`test_chunking.py` pass.

PR #3161, opened 2026-08-28
Route get_node through the same resolver as get_neighbors

### LINKED ISSUE #3160: get_node and get_neighbors disagree on the same label — get_node returns an arbitrary iteration-order match
## Summary

The MCP `get_node` and `get_neighbors` tools resolve a `label` differently, so the same lookup can return inconsistent answers. `get_node` does a raw substring scan over `G.nodes(data=True)` and returns `matches[0]` — whichever node comes first in graph iteration order. `get_neighbors` uses the tiered `_find_node` ranking plus `find_node_ambiguity`, which was added specifically so equally-good matches aren't
silently broken by iteration order.

For a hub name that matches several nodes in different files, `get_node` returns an arbitrary one while `get_neighbors` correctly reports the lookup as ambiguous.

## Steps to reproduce

1. Run the MCP server against a graph where one label matches multiple nodes in different files (e.g. `extract` — an `extract()` entrypoint plus several `extract_*` helper nodes).
2. `get_node({"label": "extract"})` → returns one node, e.g. `extract_bash()` in `graphify/extractors/bash.py`.
3. `get_neighbors({"label": "extract"})` on the same server → returns `Ambiguous: 'extract' matches N nodes in different files … retry with the repo-relative path or the full node id`.

## Expected behavior

Both tools resolve the same label the same way — either both return the same node, or both report it as ambiguous. A hub-name lookup should not silently pick one symbol.

## Actual behavior

`get_node` returns an arbitrary match determined by `G.nodes()` iteration order (so a graph rebuild that reorders nodes can change the answer), while `get_neighbors` reports ambiguity. A caller looking up a hub name via `get_node` gets the wrong symbol with no signal that a tie was broken.

## Root cause

- `_tool_get_node` (`graphify/serve.py`): `matches = [(nid, d) for nid, d in G.nodes(data=True) if <substring>]` then `matches[0]`.
- `_tool_get_neighbors` (`graphify/serve.py`): `_find_node(G, label)` + `find_node_ambiguity(G, label)`.

`find_node_ambiguity` exists precisely to stop `[0]`-of-a-tie from presenting one arbitrary file as "the answer"; `get_node` never went through it.

## Fix

Route `get_node` through the same resolver as `get_neighbors` so both share one resolution contract (tiered match + cross-file ambiguity). `get_node` then returns the tiered winner, or the same "Ambiguous — retry with path or id" message when the
winning tier spans several files.

noQbot:
## Summary

Route the MCP `get_node` tool through the same node resolver as `get_neighbors`, so both tools resolve a `label` the same way. Previously `get_node` did a raw substring scan over `G.nodes(data=True)` and returned `matches[0]` — an arbitrary match determined by graph iteration order — while `get_neighbors` used tiered matching plus ambiguity detection. For a hub name that matches several nodes in different files, the two tools gave different answers.

Fixes https://github.com/Graphify-Labs/graphify/issues/3160

## What changed

- Added one module-level helper, `_resolve_single_node(G, label)`, that does the tiered `_find_node` match plus the cross-file `find_node_ambiguity` check and returns `(node_id, None)` for a clean winner or `(None, message)` for no-match / ambiguity.
- `_tool_get_node` now routes through it — replacing the substring-scan + `[0]`. It returns the tiered winner, or the same `Ambiguous — retry with the repo-relative path or the full node id` message `get_neighbors` already used, when the winning tier spans multiple files.
- `_tool_get_neighbors` now calls the same helper. **Its behaviour and messages are unchanged** — the resolution logic it already had just moved into the shared function, so both tools have one resolution contract.

## Why

`find_node_ambiguity` exists specifically so that `[0]`-of-a-tie doesn't present one arbitrary file as "the answer". `get_node` never went through it, so a hub lookup (e.g. `extract`) silently returned whichever `extract_*` helper came first in `G.nodes()` order — and a graph rebuild that reorders nodes could change the answer.

## Testing

- New unit test for `_resolve_single_node` (ambiguous → message, unique → id, missing → not-found) — runs offline, passes.
- New HTTP end-to-end test: `get_node` and `get_neighbors` both report "Ambiguous" for a cross-file label and both resolve a unique label cleanly.
- Existing serve tests pass (`test_serve.py`: 146 passed). ruff clean.

## Notes

- Root cause and fix were identified using Cursor (Composer 2.5) + [Vinv AI](https://vinv.ai/)

### DIFF
--- a/graphify/serve.py
+++ b/graphify/serve.py
@@ -1362,6 +1362,31 @@ def find_node_ambiguity(G: nx.Graph, label: str) -> list[str]:
     return []
 
 
+def _resolve_single_node(G: nx.Graph, label: str) -> tuple[str | None, str | None]:
+    """Shared node resolution for the get_node / get_neighbors tools.
+
+    Returns ``(node_id, None)`` when *label* resolves to a single winner via the
+    tiered `_find_node` ranking, or ``(None, message)`` when there is no match or
+    the winning tier spans several source files. Routing both tools through this
+    keeps get_node from silently returning a `G.nodes()` iteration-order match for
+    a hub name while get_neighbors reports the same lookup as ambiguous (#ADR-0001).
+    """
+    matches = _find_node(G, label)
+    if not matches:
+        return None, f"No node matching '{label}' found."
+    rivals = find_node_ambiguity(G, label)
+    if rivals:
+        listing = "\n".join(
+            f"  {G.nodes[r].get('source_file') or r}\n    id: {r}" for r in rivals
+        )
+        return None, (
+            f"Ambiguous: '{label}' matches {len(rivals)} nodes in different files.\n"
+            f"{listing}\n"
+            "Retry with the repo-relative path or the full node id."
+        )
+    return matches[0], None
+
+
 def _shortest_path_text(G: nx.Graph, arguments: dict) -> str:
     """Body of the `shortest_path` MCP tool (module-level so tests can call it
     without an mcp install).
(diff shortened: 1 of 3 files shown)


safishamsi:
Shipped in v0.9.52 via authorship-preserving cherry-pick so you keep contributor-graph credit. Thanks @noQbot! Release: https://github.com/Graphify-Labs/graphify/releases/tag/v0.9.52

Issue #3165, opened 2026-08-28
Should tree-sitter-sql be a core dependency? Default installs still extract 0 nodes from .sql (follow-up to #1745, #3047)


egarcia74:
### Where things stand

#1745 fixed the *silence*: since 0.9.12 a default install prints a warning naming `pip install "graphifyy[sql]"` when `.sql` files are skipped. #3047/#3072 (v0.9.50) fixed the same gap for the `postgres` extra, which now carries the grammar.

What is still true on 0.9.51: `uv tool install graphifyy` / `pipx install graphifyy` — the commands the README leads with — produce a graph with **an entire language missing** for any corpus that contains SQL, and the fix requires the operator to notice a warning and re-install with an extra. We hit this in the field on a 391-file T-SQL corpus (zero SQL object nodes on a fresh default install).

### The question

Is there a reason `tree-sitter-sql` stays optional that I'm missing? The packaging comments in `pyproject.toml` give the criteria for optional grammars, and SQL seems to fail all of them:

- **Toolchain risk** (the reason `tree-sitter-dm` is optional, #1104): does not apply — `tree-sitter-sql` ships prebuilt abi3 wheels for every supported platform (Windows/macOS/Linux glibc+musl, x86_64+arm64) from 0.3.9, so a core pin keeps the default install toolchain-free.
- **Niche language** (the reason `tree-sitter-commonlisp` is optional): SQL is not niche — schemas, migrations and stored procedures show up in most application repos.
- **A regex fallback exists** (the reason `tree-sitter-pascal` is optional, #781): there is no fallback for SQL; absence means zero nodes.

### Proposal

Promote `tree-sitter-sql>=0.3.9,<0.4` to `[project.dependencies]` (upper bound like every other grammar pin), keep `sql` and the `postgres` entry as no-op aliases so existing install commands keep resolving, drop `.sql` from `_EXTRA_FOR_EXTENSION` so the #1745 warning no longer points at a redundant extra, and add a packaging-layer test so the dependency cannot slide back into an extra unnoticed.

A PR implementing exactly that is ready (rebased on current `v8`, lock regenerated, full suite green). If the answer is "no, extras are the policy", that's fine too — the fallback would be to make the README's leading install command `graphifyy[sql]`, or to have the #1745 warning fail the run rather than warn, so the missing language is not something an operator can scroll past.

PR #3166, opened 2026-08-28
fix(sql): recover T-SQL bracket-named, CREATE OR ALTER, and PROC routines (#3164)

### LINKED ISSUE #3164: fix(sql): T-SQL stored procedures and functions are silently dropped — 0 of 26 recovered on a 391-file SQL Server corpus
### The defect

`extract_sql` emits no node for T-SQL stored procedures or functions. Tables and views from the same files extract fine, so the gap is invisible unless you count: on a real 391-file SQL Server DACPAC corpus (staging/model/transform layers, `[Schema].[Object]` naming throughout), **0 of 26 procedures** produced a node, no warning, exit 0.

Two independent defects stack:

1. **The grammar cannot parse the T-SQL routine shape.** `CREATE PROCEDURE ... AS BEGIN ... END` has no `create_procedure` production in tree-sitter-sql (the `AS BEGIN` body idiom never parses structurally), so the ERROR-node regex recovery in `graphify/extractors/sql.py` is these objects' *only* path into the graph.
2. **The recovery regex does not speak T-SQL.** It accepts only bare or double-quoted names — no bracket-delimited identifiers (`CREATE PROCEDURE [dbo].[usp_Load]`), no `CREATE OR ALTER` (T-SQL has no `OR REPLACE`), no `PROC` shorthand. On a corpus where every routine is bracket-named, recovery recovers nothing.

A third, smaller defect surfaces once you look: the two recovery sites (the walk-time ERROR-node scan and the whole-file `has_error` fallback) carry two copies of the regex that have drifted, so a mixed-delimiter name (`CREATE PROCEDURE dbo.[usp_Mixed]`) mints a phantom `dbo()` node from the second site while the first captures `dbo.[usp_Mixed]`.

### Reproduce

graphifyy 0.9.51 (current `v8`), `tree-sitter-sql` installed.

```sql
-- routines.sql
CREATE PROCEDURE [dbo].[usp_Load]
AS
BEGIN
    SELECT 1;
END
GO
CREATE OR ALTER PROC [dbo].[usp_Refresh] AS BEGIN SELECT 2; END
```

```python
from pathlib import Path
from graphify.extractors.sql import extract_sql
sorted(n["label"] for n in extract_sql(Path("routines.sql"))["nodes"])
# v8:       ['routines.sql']                                    -- no routine nodes
# expected: ['[dbo].[usp_Load]()', '[dbo].[usp_Refresh]()', 'routines.sql']
```

### Expected

`usp_Load()` and `usp_Refresh()` as nodes, recovered through the same ERROR-node path that already recovers PL/pgSQL routines (#2180).

### Adjacent, not addressed here

A bracket-named `CREATE FUNCTION [dbo].[fn_Total]() RETURNS INT AS BEGIN RETURN 1; END` *does* parse structurally, but its label comes out as `dbo].[fn_Total(` — the outer delimiters are stripped as if the whole `[dbo].[fn_Total]` were one bracketed token. That is a separate defect in the structural `create_function` path; noting it so it is not mistaken for a recovery gap.

### Proposed fix

A PR is ready: one shared module-level recovery pattern used by both sites (fixes the drift/phantom), bracket-delimited identifiers including the `]]` escape, `CREATE OR ALTER`, `PROC`, a `\b` before `CREATE`, and a comment/string-masked whole-file scan so commented-out or dynamic-SQL DDL cannot fabricate routines when an unrelated parse error arms recovery. All changes are in `graphify/extractors/sql.py`; tests in `tests/test_multilang.py`.

egarcia74:
Fixes #3164. Replaces #2960 — same change, squashed to a single commit and rebased onto current `v8` so it cherry-picks cleanly; #2960 carries the review history (four rounds of Graphify-review findings, each reproduced and either fixed or refuted in the thread) and is being closed with a pointer here.

## Problem

T-SQL stored procedures and functions are silently dropped from the graph. On a real 391-file SQL Server DACPAC corpus (`[Schema].[Object]` naming throughout), 0 of 26 stored procedures produced a node while tables and views extracted fine, so the gap is invisible unless you count.

Two independent defects stack:

1. **The grammar cannot parse the T-SQL routine shape.** `CREATE PROCEDURE ... AS BEGIN ... END` has no `create_procedure` production in tree-sitter-sql, so ERROR-node regex recovery is these objects' *only* path into the graph.
2. **The recovery regex does not speak T-SQL.** It accepted only bare or double-quoted names — no bracket-delimited names, no `CREATE OR ALTER`, no `PROC` shorthand. On this corpus every routine uses brackets, so recovery recovered nothing.

## What this PR does

All changes are in `graphify/extractors/sql.py`:

- **One shared `_ROUTINE_RECOVERY_RX`**, hoisted to module level. The two recovery sites (walk-time ERROR-node scan, whole-file `has_error` fallback) had drifted; the second minted a phantom `dbo()` node for a mixed-delimiter name (`CREATE PROCEDURE dbo.[usp_Mixed]`). Sharing the pattern fixes that and prevents re-drift.
- **Bracket-delimited identifiers**, including T-SQL's `]]` escape for a literal `]` (`[dbo].[a]]b]` recovers as one routine, not truncated at the escape).
- **`CREATE OR ALTER`** alongside `OR REPLACE`; **`PROC`** alongside `PROCEDURE`; `\bCREATE` so a bare word ending in CREATE (`AUTOCREATE PROCEDURE x`) cannot match.
- **The per-ERROR-node scan is removed.** Any ERROR node sets `root.has_error`, so the whole-file scan already covers everything a per-node scan could, from the same pattern with `_add_node` deduping by id. The per-node scan was also unsound: an ERROR fragment can begin mid-comment (tree-sitter's lexer does not nest `/* */`), which let commented-out DDL fabricate routine nodes.
- **Comment/string masking for the whole-file scan** (`_scan_sql`, a linear offset-preserving scanner): single-quoted strings (dynamic SQL) and line/block comments (nesting-aware; an unclosed block comment blanks to EOF) are blanked; double-quoted and bracket-delimited identifiers are preserved because they carry the recoverable names, and a match whose `CREATE` *starts* inside one is skipped as identifier data. An unterminated or comment-swallowing delimiter span is irreducibly ambiguous, so the union of every reading is blanked — over-blanking can lose a routine on an already-broken line, never fabricate one.

Known, documented holes: multi-line single-quoted dynamic SQL, and string-quoting dialects the mask does not model (MySQL `"..."` strings, PostgreSQL dollar quoting) still reach the scan when the `has_error` gate is armed.

## Tests

`tests/test_multilang.py`: bracket names, `]]` escapes, `OR ALTER`, `PROC`, mixed-delimiter no-phantom, keyword-inside-identifier no-phantom, commented-out and string-embedded DDL not fabricated, nested and unclosed comments, plus a differential fuzz over the mask (a 1M-case fuzz baseline was run offline against the final scanner).

## Validation

- Repro from #3164 on this branch: `['[dbo].[usp_Load]()', '[dbo].[usp_Refresh]()', 'routines.sql']` (was `['routines.sql']` on `v8`).
- Full suite on this branch over current `v8` (0.9.51): **5041 passed, 49 skipped** (`test_ollama*.py` excluded — local Ollama config, unrelated).
- Merges cleanly alongside the companion core-dependency change (#3165 / formerly #2959); both touch `tests/test_multilang.py`.

Out of scope, noted in #3164: a bracket-named `CREATE FUNCTION` that *does* parse structurally gets the label `dbo].[fn_Total(` — a separate delimiter-stripping defect in the structural path.

### DIFF
--- a/graphify/extractors/sql.py
+++ b/graphify/extractors/sql.py
@@ -6,6 +6,248 @@ import re
 from pathlib import Path
 from graphify.extractors.base import _file_stem, _make_id
 
+# Recovers CREATE FUNCTION/PROCEDURE statements the grammar could not parse
+# structurally. Used by BOTH recovery sites — the walk-time ERROR-node scan and
+# the whole-file has_error fallback (#2180). They MUST share one pattern: when
+# they disagreed, the same statement produced two nodes with different names
+# (the ERROR scan captured `dbo.[usp_Mixed]`, the fallback stopped at `dbo`,
+# and _add_node's id-dedupe never fired because the ids differed).
+#
+# Each name part is a bare identifier, a double-quoted (delimited) one, or a
+# T-SQL bracket-delimited one, so CREATE OR REPLACE FUNCTION "public"."fn"(...)
+# and CREATE PROCEDURE [dbo].[usp_Load] ... are both recovered. A bare [\w$.]+
+# stops dead at the leading delimiter, which silently dropped every quoted
+# PL/pgSQL routine (#2180) and every bracket-named T-SQL procedure. T-SQL's
+# AS BEGIN...END body idiom always lands in recovery — the grammar has no
+# create_procedure parse for it — and T-SQL spells re-creation CREATE OR ALTER
+# (it has no OR REPLACE), so accept that form too, mirroring fb_proc_or_trigger.
+# Inside a bracket-delimited part, a literal ] is escaped by doubling
+# ([a]]b] names the identifier a]b), so consume ]] before treating a
+# lone ] as the closing delimiter — stopping at the first ] truncated
+# the name and minted a phantom that could collide with a real [a].
+# PROC is T-SQL's official shorthand for PROCEDURE and equally common in the
+# wild; the optional (?:EDURE)? still requires trailing whitespace, so a word
+# that merely starts with PROC cannot match.
+# \bCREATE: without the boundary, CREATE matched inside a bare word, so
+# 'SELECT AUTOCREATE PROCEDURE x FROM t;' in an error-bearing file minted a
+# phantom routine x() (delimited identifiers are span-skipped at the scan
+# site, but a bare word has no span).
+_ROUTINE_RECOVERY_RX = re.compile(
+    r"\bCREATE\s+(?:OR\s+(?:REPLACE|ALTER)\s+)?(?:FUNCTION|PROC(?:EDURE)?)\s+"
+    r"(?:IF\s+NOT\s+EXISTS\s+)?"
+    r"((?:\"(?:[^\"\n]|\"\")+\"|\[(?:[^\]\n]|\]\])+\]|[\w$]+)"
+    r"(?:\s*\.\s*(?:\"(?:[^\"\n]|\"\")+\"|\[(?:[^\]\n]|\]\])+\]|[\w$]+))*)",
+    re.IGNORECASE,
+)
+
+# _mask_sql_comments is a linear character scanner, not a regex: the four
+# span kinds interact in ways a single pattern cannot express safely —
+# comment-opener parity inside strings, nested block comments, and
+# end-of-line abandonment of unclosed literals. Span handling:
+#
+# - single-quoted strings are BLANKED like comments: routine names never
+#   live in single quotes, and dynamic SQL (EXEC(N'CREATE PROC [dbo].[Fake]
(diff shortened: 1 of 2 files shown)


safishamsi:
Shipped in v0.9.52 via authorship-preserving cherry-pick so you keep contributor-graph credit. Thanks @egarcia74! Release: https://github.com/Graphify-Labs/graphify/releases/tag/v0.9.52

PR #3168, opened 2026-08-28
Fix Svelte, JS destructuring, and Rust use-declaration extraction

wn-mitch:
Three independent extraction fixes, one commit each. All three surfaced from running graphify over a Rust + SvelteKit monorepo, where the frontend extracted as stubs and the generated SeaORM entity modules were orphaned from the code using them.

Each commit passes the suite on its own. Full suite at the tip: 5043 passed. The 4 failures in `tests/test_ollama_retry_cap.py` are present on `v8` before these commits and are untouched here.

### 1. Parse Svelte script blocks with the SFC masker

Feeding a whole `.svelte` component to the JS grammar makes the markup a top-level `ERROR` node, so `import_statement` and declaration nodes are never reached. Every component extracted as a stub — the regex pass rescued import specifiers and nothing else, no symbols, no participation in the call graph. On the codebase that prompted this, 613 files.

`.vue` already solves this, so Svelte now shares it. The Vue masker is renamed `_sfc_mask_non_script` (old name kept as an alias); it blanks non-script regions, keeps newlines so line numbers stay accurate, and the script is parsed with the grammar its `lang` implies. Every script block survives the mask, so a Svelte 5 `<script module>` block is parsed alongside the instance script, as is Svelte 4's `context="module"`.

`_parse_js_tree` needed the same masking, or the call-graph pass kept handing it raw component bytes.

The static-import regex rescue is dropped, since the AST now edges those and it would double-emit. The dynamic-import rescue stays and still scans the raw source, because `{#await import('./X.svelte')}` lives in the markup layer the mask blanks out.

### 2. Node destructured bindings, not the pattern text

A destructuring declarator binds identifiers, but its `name` field is the pattern source. Reading it verbatim minted one node labelled `{ a, b: renamed, c = 1, ...rest }` — text that names no symbol and can never be a reference target.

Now the binding side is walked: a `pair_pattern`'s value (`b: renamed` binds `renamed`, not the key `b`), an assignment pattern's left operand (`c = $bindable()` binds `c`, not `$bindable`), recursing through nested patterns and rest elements.

Destructured `require` is excluded: `const { doWork } = require('./lib')` binds an import, and noding it as a local shadowed the real cross-file target. The exclusion checks the callee name, since `_find_require_call` matches the call *shape* only and leaves the name check to its callers.

This is ordinary JS/TS, but Svelte 5 makes it universal — every component destructures `$props()` — so it was roughly one junk node per component.

### 3. Resolve Rust `use` declarations through the AST

`use_declaration` was read by string-splitting its text: everything before the first `{`, then the last `::` segment. Braced lists collapsed to their shared prefix and every name inside was lost; `as` clauses were never parsed, leaving the alias glued to the symbol (`Entity as Risk`); and the resulting id was a bare name no node carried, so the edge dangled and was dropped at build time.

That last part makes a crate prelude a **false hub**, not just a dead end. Every external prelude import — `use sea_orm::entity::prelude::*`, `use loco_rs::prelude::*` — strips to the segment `prelude`, whose id collides with a local `prelude.rs` file node. One SeaORM model tree showed 466 inbound edges that were all spurious, 0 outbound, and 75 generated entity modules orphaned from their consumers.

The use tree is now walked. `_rust_use_leaves` flattens a declaration into one leaf per bound name (`use_as_clause`, `use_list`, `scoped_use_list`, `use_wildcard`, arbitrary nesting). `_resolve_rust_use_path` resolves `crate`/`super`/`self`-anchored paths through the module tree to a file on disk, encoding that `mod.rs`/`lib.rs`/`main.rs` *are* their module while `foo.rs` keeps its children in a sibling `foo/`; a `use` path's trailing segment is usually a symbol, so the tail is retried as a symbol inside the module the rest resolves to.

Each leaf emits a file-level `imports_from` plus a symbol-level edge, both stamped with `target_file`. `pub use` emits `re_exports`, so the corpus-level barrel collapse follows a consumer through a prelude to the defining symbol the way it already does for a JS barrel. Unresolvable paths (external crates) get a sourceless stub rather than a phantom target.

`rust.py`'s own edge filter allowed dangling targets for `imports`/`imports_from` but not `re_exports`, dropping the new edges before the corpus pass saw them. It now matches the shared engine filter.

Known limitation, called out in the commit: a *renamed* re-export does not chain symbol-level, because the collapse keys on a name derived from the target id and there is nowhere to record the alias. File-level edges resolve either way, and JS has the same limitation.

### Verification

38 new tests across three files:

- `tests/test_svelte_extraction.py` (14) — masking and line-number stability, static/dynamic imports, symbols, Svelte 5 `<script module>` and Svelte 4 `context="module"`, runes, plain-JS blocks, markup-only files, a cross-file `calls` edge through the masked `_parse_js_tree`, and a guard asserting the unmasked path still loses everything.
- `tests/test_rust_use_reexports.py` (17) — use-tree parsing (braced, nested, aliased, glob), module resolution (`crate`/`super`/`self`, `mod.rs` vs plain file, external crates), and the emitted edges, including a consumer resolving through a prelude to the definition and a regression test that no `use` edge points at a phantom node.
- `tests/test_js_exported_scalar_bindings.py` (+7) — bound names vs pattern text, keys and defaults excluded, destructured `require` still resolving cross-file, and closures in a destructured initializer still tracked.

Effect on the monorepo that prompted this: 22,707 nodes / 47,025 edges to 31,404 / 63,594. Files with syntax errors 613 to 4 (the remainder are a tree-sitter-typescript limitation with TS `import('…')` types, which hits plain `.ts` equally). Entity-module median degree 3 to 7.

I am happy to split this into three PRs if you would rather review them separately.

### DIFF
--- a/graphify/extract.py
+++ b/graphify/extract.py
@@ -1679,27 +1679,13 @@ def extract_svelte(path: Path) -> dict:
         # Both rescues can repeat an edge: a recovered parse edges some
         # imports before the error node, and a specifier imported twice in one
         # file (`import('./X')` in two markup branches) is matched twice.
-        _dedupe_edges(result)
+        # Imported inside the function, as cli.py and watch.py do, to keep
+        # extract.py free of a module-level dependency on build.py.
+        from graphify.build import dedupe_edges
+        result["edges"] = dedupe_edges(result.get("edges", []))
     except Exception:
         pass
     return result
-
-
-def _dedupe_edges(result: dict) -> None:
-    """Drop repeat ``(source, target, relation)`` edges, keeping the first.
-
-    The first occurrence is the AST-derived edge, which carries ``target_file``
-    and richer context than a regex rescue's.
-    """
-    seen: set[tuple[str, str, str]] = set()
-    kept = []
-    for edge in result.get("edges", []):
-        key = (edge.get("source"), edge.get("target"), edge.get("relation"))
-        if key in seen:
-            continue
-        seen.add(key)
-        kept.append(edge)
-    result["edges"] = kept
 
 
 def extract_astro(path: Path) -> dict:
--- a/graphify/extract.py
+++ b/graphify/extract.py
@@ -137,6 +137,7 @@ from graphify.extractors.resolution import (  # noqa: E402,F401
     _ts_collect_type_refs,
     _ts_heritage_clause_entries,
     _ts_walk_class_members,
+    _sfc_mask_non_script,
     _vue_mask_non_script,
     _walk_js_tree,
     _walk_python_tree,
@@ -1599,16 +1600,40 @@ def _emit_rescued_import(
 
 
 def extract_svelte(path: Path) -> dict:
-    """Extract imports from .svelte files: script-block via JS AST + template regex fallback.
+    """Extract imports, symbols, and type refs from a ``.svelte`` component.
 
-    Tree-sitter only sees the <script> block. Svelte template syntax like
-    {#await import('./X.svelte')} lives in the markup layer and is invisible
-    to the JS parser, so a regex pass covers those dynamic imports.
+    Masks the non-``<script>`` regions and parses the script with the grammar
+    its ``lang`` implies (``tsx``->TSX, ``js``/``jsx``->JS, ``ts`` or unset->TS;
+    TS is a superset of JS so it is a safe default), mirroring
+    :func:`extract_vue`. Feeding the whole component to the JS grammar makes the
+    markup a top-level ERROR node, so ``import_statement`` and declaration nodes
+    are never reached and everything but a stray symbol is dropped (#713).
+
+    Both script blocks of a Svelte 5 component survive the mask, so a
+    ``<script module>`` block is parsed alongside the instance script.
+
+    A regex pass then recovers ``import('...')`` dynamic imports, which the AST
+    pass does not edge and which legally live in markup-layer template syntax
+    such as ``{#await import('./X.svelte')}`` — outside every script block, and
+    so blanked out of the masked source the AST sees.
     """
-    result = _extract_generic(path, _JS_CONFIG)
+    try:
+        src = path.read_text(encoding="utf-8", errors="replace")
+    except OSError:
+        return {"nodes": [], "edges": []}
+
+    masked, lang = _sfc_mask_non_script(src)
+    if lang == "tsx":
+        config = _TSX_CONFIG
+    elif lang in ("js", "jsx"):
+        config = _JS_CONFIG
+    else:  # "ts" or unspecified — default to the TS grammar (superset of JS)
+        config = _TS_CONFIG
+
+    result = _extract_generic(path, config, source_override=masked.encode("utf-8"))
+
     try:
         import re as _re
-        src = path.read_text(encoding="utf-8", errors="replace")
         existing_ids = {n["id"] for n in result.get("nodes", [])}
         # Source file node ID must match the one _extract_generic creates:
         # _make_id(str(path)) - single arg, no stem prefix. Otherwise the source
@@ -1616,39 +1641,19 @@ def extract_svelte(path: Path) -> dict:
         file_node_id = _make_id(str(path))
         aliases = _load_tsconfig_aliases(path.parent)
         base_url = _load_tsconfig_base_url(path.parent)
+        # Scanned over the raw source, not the masked one, so template-layer
+        # dynamic imports are seen. Resolution is shared with the static pass:
+        # relative paths and tsconfig aliases probe real on-disk extensions
+        # (#716, #701), and a target that IS a real file emits an edge stamped
+        # with target_file instead of an absolute-id ghost stub (#2195).
         for m in _re.finditer(r"""import\(\s*['"]([^'"]+)['"]\s*\)""", src):
             raw = m.group(1)
             if not raw:
                 continue
-            # Resolution + emit shared with the static pass below: relative
-            # paths and tsconfig aliases probe real on-disk extensions (#716,
-            # #701), and a target that IS a real file emits an edge stamped
-            # with target_file instead of an absolute-id ghost stub (#2195).
             _emit_rescued_import(
                 result, existing_ids, file_node_id, path, raw,
                 "dynamic_import", aliases, base_url,
             )
-        # Static imports inside <script> blocks. The JS tree-sitter parser fed
-        # the full .svelte file produces a top-level ERROR node (HTML markup
-        # is not valid JS), so import_statement nodes are never reached and
-        # static imports are silently dropped (#713). Regex over each script
-        # body recovers them.
-        script_re = _re.compile(
-            r"<script\b[^>]*>([\s\S]*?)</script\s*>", _re.IGNORECASE
-        )
-        static_import_re = _re.compile(
-            r"""import\s+(?:[^'"`;]+?\s+from\s+)?['"]([^'"]+)['"]"""
-        )
-        for script_match in script_re.finditer(src):
-            script_body = script_match.group(1)
-            for m in static_import_re.finditer(script_body):
-                raw = m.group(1)
-                if not raw:
-                    continue
-                _emit_rescued_import(
-                    result, existing_ids, file_node_id, path, raw,
-                    "imports_from", aliases, base_url,
-                )
     except Exception:
         pass
     return result
@@ -1735,7 +1740,7 @@ def extract_vue(path: Path) -> dict:
     except OSError:
         return {"nodes": [], "edges": []}
 
-    masked, lang = _vue_mask_non_script(src)
+    masked, lang = _sfc_mask_non_script(src)
     if lang == "tsx":
         config = _TSX_CONFIG
     elif lang in ("js", "jsx"):
--- a/graphify/extract.py
+++ b/graphify/extract.py
@@ -1676,9 +1676,10 @@ def extract_svelte(path: Path) -> dict:
                     result, existing_ids, file_node_id, path, raw,
                     "imports_from", aliases, base_url,
                 )
-            # A recovered parse can still have edged SOME imports before the
-            # error node, so drop any duplicate the rescue re-emitted.
-            _dedupe_edges(result)
+        # Both rescues can repeat an edge: a recovered parse edges some
+        # imports before the error node, and a specifier imported twice in one
+        # file (`import('./X')` in two markup branches) is matched twice.
+        _dedupe_edges(result)
     except Exception:
         pass
     return result
--- a/graphify/extract.py
+++ b/graphify/extract.py
@@ -1623,8 +1623,11 @@ def extract_svelte(path: Path) -> dict:
     """
     try:
         src = path.read_text(encoding="utf-8", errors="replace")
-    except OSError:
-        return {"nodes": [], "edges": []}
+    except OSError as e:
+        # Report the failure rather than returning an empty result: an
+        # unreadable component is indistinguishable from an empty one
+        # otherwise, and extract() warns on `error` (#2551).
+        return {"nodes": [], "edges": [], "error": f"cannot read {path}: {e}"}
 
     masked, lang = _sfc_mask_non_script(src)
     if lang == "tsx":
@@ -1781,8 +1784,11 @@ def extract_vue(path: Path) -> dict:
     """
     try:
         src = path.read_text(encoding="utf-8", errors="replace")
-    except OSError:
-        return {"nodes": [], "edges": []}
+    except OSError as e:
+        # Report the failure rather than returning an empty result: an
+        # unreadable component is indistinguishable from an empty one
+        # otherwise, and extract() warns on `error` (#2551).
+        return {"nodes": [], "edges": [], "error": f"cannot read {path}: {e}"}
 
     masked, lang = _sfc_mask_non_script(src)
     if lang == "tsx":
--- a/graphify/extract.py
+++ b/graphify/extract.py
@@ -1643,6 +1643,18 @@ def extract_svelte(path: Path) -> dict:
         # _make_id(str(path)) - single arg, no stem prefix. Otherwise the source
         # endpoint is a phantom node and build_from_json drops the edge (#701).
         file_node_id = _make_id(str(path))
+        if result.get("error") and not result.get("nodes"):
+            # A hard failure (grammar missing, unreadable source) returns no
+            # nodes at all, so a rescued edge would have a dangling source and
+            # be dropped at build time. Mint the file node _extract_generic
+            # would have, in the same shape, so the rescue is worth running.
+            # `error` stays on the result for extract()'s own reporting.
+            result.setdefault("nodes", []).append({
+                "id": file_node_id, "label": path.name,
+                "file_type": "code", "source_file": str(path),
+                "source_location": "L1",
+            })
+            existing_ids.add(file_node_id)
         aliases = _load_tsconfig_aliases(path.parent)
         base_url = _load_tsconfig_base_url(path.parent)
         # Scanned over the raw source, not the masked one, so template-layer
@@ -1658,10 +1670,13 @@ def extract_svelte(path: Path) -> dict:
                 result, existing_ids, file_node_id, path, raw,
                 "dynamic_import", aliases, base_url,
             )
-        if result.get("parse_errors"):
-            # The masked script did not parse cleanly, so `import_statement`
-            # nodes may never have been reached and the AST pass edged nothing.
-            # Fall back to the regex rescue the pre-mask extractor relied on.
+        if result.get("parse_errors") or result.get("error"):
+            # The AST pass produced no usable tree — the masked script parsed
+            # WITH errors (`parse_errors`, so `import_statement` nodes may
+            # never have been reached), or it failed outright (`error`: the
+            # grammar is missing, the source unreadable). Both leave imports
+            # unedged, so fall back to the regex rescue the pre-mask extractor
+            # ran unconditionally.
             # Gated on the failure so a clean parse does not double-emit: the
             # AST already edges those specifiers. Scanned over the MASKED
             # source, whose only surviving text is the script regions.
--- a/graphify/extract.py
+++ b/graphify/extract.py
@@ -1616,6 +1616,10 @@ def extract_svelte(path: Path) -> dict:
     pass does not edge and which legally live in markup-layer template syntax
     such as ``{#await import('./X.svelte')}`` — outside every script block, and
     so blanked out of the masked source the AST sees.
+
+    Static imports are edged by the AST pass, so the old regex rescue for them
+    would double-emit and is used only when the script fails to parse, where
+    the AST pass reaches no ``import_statement`` at all.
     """
     try:
         src = path.read_text(encoding="utf-8", errors="replace")
@@ -1654,9 +1658,47 @@ def extract_svelte(path: Path) -> dict:
                 result, existing_ids, file_node_id, path, raw,
                 "dynamic_import", aliases, base_url,
             )
+        if result.get("parse_errors"):
+            # The masked script did not parse cleanly, so `import_statement`
+            # nodes may never have been reached and the AST pass edged nothing.
+            # Fall back to the regex rescue the pre-mask extractor relied on.
+            # Gated on the failure so a clean parse does not double-emit: the
+            # AST already edges those specifiers. Scanned over the MASKED
+            # source, whose only surviving text is the script regions.
+            static_import_re = _re.compile(
+                r"""import\s+(?:[^'"`;]+?\s+from\s+)?['"]([^'"]+)['"]"""
+            )
+            for m in static_import_re.finditer(masked):
+                raw = m.group(1)
+                if not raw:
+                    continue
+                _emit_rescued_import(
+                    result, existing_ids, file_node_id, path, raw,
+                    "imports_from", aliases, base_url,
+                )
+            # A recovered parse can still have edged SOME imports before the
+            # error node, so drop any duplicate the rescue re-emitted.
+            _dedupe_edges(result)
     except Exception:
         pass
     return result
+
+
+def _dedupe_edges(result: dict) -> None:
+    """Drop repeat ``(source, target, relation)`` edges, keeping the first.
+
+    The first occurrence is the AST-derived edge, which carries ``target_file``
+    and richer context than a regex rescue's.
+    """
+    seen: set[tuple[str, str, str]] = set()
+    kept = []
+    for edge in result.get("edges", []):
+        key = (edge.get("source"), edge.get("target"), edge.get("relation"))
+        if key in seen:
+            continue
+        seen.add(key)
+        kept.append(edge)
+    result["edges"] = kept
 
 
 def extract_astro(path: Path) -> dict:
--- a/graphify/extractors/engine.py
+++ b/graphify/extractors/engine.py
@@ -1912,6 +1912,20 @@ def _find_require_call(value_node):
         return _find_require_call(obj)
     return None
 
+def _is_require_initializer(value_node, source: bytes) -> bool:
+    """True when a declarator's initializer is a literal ``require(...)`` call.
+
+    ``_find_require_call`` matches the call *shape* only — any
+    ``identifier(...)`` — and leaves the callee-name check to its callers, so
+    it must not be used alone to recognise a CJS import.
+    """
+    call = _find_require_call(value_node)
+    if call is None:
+        return False
+    fn = call.child_by_field_name("function")
+    return fn is not None and _read_text(fn, source) == "require"
+
+
 def _require_imports_js(node, source: bytes, importer_nid: str, stem: str, edges: list, str_path: str) -> bool:
     """Detect CommonJS require imports inside lexical_declaration / variable_declaration.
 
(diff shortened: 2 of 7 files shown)


wn-mitch:
All five "worth a look" findings addressed, one commit each area.

**1. Static Svelte imports no longer rescued when parsing fails** — real, fixed in `690f0ef`. The rescue is back, gated on `result["parse_errors"]` so a clean parse does not double-emit what the AST already edged, and scanned over the masked source (whose surviving text is exactly the script regions). A recovered parse can edge some imports before the error node, so duplicate `(source, target, relation)` edges are dropped. Four new tests: syntax-error rescue, no double-emit on a clean parse, no duplicate on a partial parse, and the dynamic-import rescue still running unconditionally.

**2. Member access on require misclassified** — doc bug only, `aab82ec`. `_find_require_call` unwrapping `member_expression` is intentional: `_require_imports_js` edges `require('./m').y` too, so if the two disagreed a destructured name would be both imported and shadowed by a local stub. (In practice the branch is not even reachable for that shape — `member_expression` is not in the accepted initializer types and `is_exported_scalar_binding` requires an `identifier` name node.) Docstring corrected; a test pins the member-access form resolving cross-file with no local stub.

**3. `pub(self) use` emitted as a re-export** — real, fixed in `a412fbf`. `is_reexport` matched any `visibility_modifier`. `pub(self)` is now excluded; `pub(crate)`/`pub(super)`/`pub(in path)` still count.

**4. use-list `self` treated as a symbol** — real, fixed in `a412fbf`. `use foo::bar::{self, Baz}` now yields the prefix unchanged rather than appending a `self` segment, so it edges the module file and mints no node labelled `self`. A leading `self::` path has an empty prefix and keeps its existing anchoring.

**5. External `pub use` loses the re-export relation** — real, fixed in `a412fbf`. The unresolved branch ignored `is_reexport`, so `pub use anyhow::Result;` was recorded as a plain import and the barrel collapse could not follow a consumer through it. It now emits the symbol-level `re_exports` alongside the file-level `imports_from`.

**Inline coupling comments** (`extract.py:1602`, `engine.py:1929`, `engine.py:2202`, `rust.py:174`, `rust.py:242`, `rust.py:393`) — no change. Each is a function that gained one or two helper calls; the fan-out is the extracted helpers this PR adds, which is the direction that makes the code readable rather than less so.

Suite: 4805 passed. The failures on the tip are all environment-only and present on `v8` before this branch — `test_terraform.py` (7) and `test_languages.py::test_markdown_nested_frontmatter_survives` need grammars this venv lacks, `test_install_references.py` needs `build`, and the 4 in `test_ollama_retry_cap.py` are untouched here.

wn-mitch:
Second round: three of the five are real and fixed in `cbad248`; two are factually wrong.

**Test encodes wrong Rust `self::` resolution** — correct, and it exposed a resolver bug, not just a bad test. `self::risk` inside `_entities/prelude.rs` means `_entities::prelude::risk`, but `_rust_module_dirs` falls back to the containing directory for a plain `foo.rs` with no sibling `foo/`, so `self::` behaved like `super::` and edged the sibling `_entities/risk.rs`. The `self` keyword now resolves through a new `_rust_self_module_dir`, which is `None` when the module has no children; the containing-directory fallback stays where it belongs, in the 2018-edition crate-relative heuristic. The bad test is replaced by three: `self::` from a `mod.rs` resolving a child, `self::` from a childless `foo.rs` resolving nothing, and `self::` from a `foo.rs` that does have a sibling `foo/`.

**`super::super` walks past crate root** — real, fixed. The walk had no floor, so more `super`s than the module has ancestors left `src/` and could resolve an unrelated file higher up the filesystem. It now stops at the crate root and falls back to a sourceless stub. Two tests: the escape attempt, and `super::super` still resolving within the crate.

**`_walk_rust_segments` drops multi-segment symbol paths** — real, fixed. `use crate::…::risk::Status::Active` dropped the edge entirely. The first unresolved segment is the item the module defines, so it is attributed (`Status`) and the remainder — which names something inside that item — is dropped instead of the whole edge.

**`_is_require_initializer` rejects member-access require** — this is the opposite of what the code does, and the opposite of last round's finding on the same function. Evidence:

```
"const { a } = require('./m').sub;"  member_expression -> True
"const a = require('./m').y;"        member_expression -> True
"const a = require('./m');"          call_expression   -> True
```

`_find_require_call` recurses through `member_expression` into its object, finds the call, and the callee-name check passes. The docstring and the behavior agree. No change.

**Rust use relation changed from `imports_from` to `imports`/`re_exports`** — no behavior lost. Every leaf still emits the file-level `imports_from`; `imports`/`re_exports` is an *additional* symbol-level edge, which is exactly how the JS side already does it (`_require_imports_js` emits `imports_from` for the file and `imports` for each destructured name). The consumers that key on the relation are unaffected: `analyze.py:675` matches `("imports_from", "re_exports")`, `affected.py` matches `imports_from`, and every other site lists `imports` alongside `imports_from`. No change.

Suite: 4810 passed. Same 9 environment-only failures as base `281ccaa` (terraform/markdown grammars, `build` module), plus the 4 pre-existing `test_ollama_retry_cap.py` ones this branch does not touch.

wn-mitch:
Third round: all four addressed — three real, one unreachable but guarded anyway.

**`_dedupe_edges` only ran inside the `parse_errors` branch** — real, fixed in `5aa5c49`. The dynamic-import regex matches the same specifier twice when it appears in two markup branches (`{#if}` / `{:else}`), so a clean parse emitted two identical edges. The dedupe now runs after both rescues. Test added.

**`_walk_rust_segments` drops the symbol when the first segment is unresolved at the anchor** — real, fixed in `2b5df71`. `use crate::Config` names an item of `lib.rs`, not of a child module, and the edge was dropped because nothing resolved before it. The anchor's own module file is now seeded as the fallback owner via a new `_rust_module_root_file`.

Worth flagging what this could not do: the seed is deliberately *not* applied to a bare crate-relative path. That path is only a guess at being crate-relative, and seeding it would resolve `use anyhow::Result` to a symbol named `anyhow` inside `lib.rs` — every external crate would collapse onto the crate root. There is a test pinning that. `use self::Config;` is handled separately, against the file itself, since an own-module item needs no child directory.

**Aliased `self` in a use list** — real, fixed in `2b5df71`. Same class as the unaliased case from last round, which I fixed in the wrong branch only: `use foo::bar::{self as bar_mod, Baz}` aliases the module `foo::bar`, but the `use_as_clause` branch appended `self` as a segment, resolving nothing and minting a node labelled `self`. Two tests.

**`super` clamp crashes when anchor is None** — not reachable: `_rust_module_dirs` always returns two `Path`s, so `anchor` is non-`None` at the clamp, and the `while` loop returns early if it ever became `None`. Guarded explicitly anyway in `2b5df71`, since the clamp should not depend on a caller invariant to avoid an `AttributeError`.

Suite: 4816 passed. Same 9 environment-only failures as base `281ccaa` (terraform/markdown grammars, `build` module), plus the 4 pre-existing `test_ollama_retry_cap.py` ones this branch does not touch. Ruff clean across `graphify` and `tests`.

wn-mitch:
Fourth round: both real, both fixed in `f702e17`.

**Named Cargo bin crate roots resolve `crate::` from the package src directory** — real. A `.rs` file sitting directly in one of Cargo's auto-discovered target directories is its own crate root, so `crate::helper` in `src/bin/tool.rs` means `src/bin/tool/helper.rs`. It resolved against the library's `src/`, edging `src/helper.rs` — a *wrong* target, not a missing one, which is the worse failure. Now handled for `src/bin/`, `examples/`, `tests/` and `benches/` via `_rust_own_crate_root_dir`. The directory form (`src/bin/tool/main.rs`) already resolved correctly through the plain `src` answer plus the module walk, and is untouched.

Four tests: the named bin resolving to its own tree (asserting the target's `source_file` is `src/bin/tool/helper.rs`, so passing on the label alone cannot mask it), an example crate not reaching the library `src/`, an ordinary library module still anchoring at `src/`, and the directory form unaffected.

**Rust use-path resolution can escape the project tree when no Cargo.toml is found** — real, and it was a hole in last round's clamp. `_rust_crate_src_root` returns `None` when nothing above the file is a package, and the clamp read `if src_root is not None and …` — so with no root the guard was skipped entirely and `super::super::…` climbed the filesystem unbounded, exactly the case the clamp was added for. There is no root to clamp against in that situation, so no walking is allowed at all; the path falls back to a sourceless stub. Test uses a loose `a/b/c/service.rs` with no `Cargo.toml`.

Suite: 4820 passed. Same 9 environment-only failures as base `281ccaa` (terraform/markdown grammars, `build` module) plus the 4 pre-existing `test_ollama_retry_cap.py` ones. `test_labeling.py::test_label_communities_batches_when_over_batch_size` failed once in a full-suite run and passes in isolation on this branch and on base — an order-dependent flake, unrelated to these changes. Ruff clean.

wn-mitch:
Fifth round: all three real, fixed in `739d54d`.

**Crate-root symbol resolution can pick lib.rs for main.rs** — real. `_rust_module_root_file` returned the first of `mod.rs`/`lib.rs`/`main.rs` present, so in a package with both roots in `src/`, `use crate::Config` inside `main.rs` attributed to `lib.rs`. When the importing file *is* a root file sitting in the anchor directory, that file is now preferred over the first name that happens to exist.

**`super::` clamp fails for auto-target crate roots** — real, and it exposed that last round's fix only handled the crate root *file*. A module nested inside a bin crate (`src/bin/tool/helper.rs`) still anchored `crate::` at the library's `src/`, so `crate::shared` there edged `src/shared.rs` instead of `src/bin/tool/shared.rs` — and the clamp inherited the wrong floor. The crate-root directory is now found by walking up to the nearest auto-target directory, which fixes both. Two tests: the nested module anchoring at the bin crate (asserting `source_file`, since both candidates are named `shared.rs`), and `super::` resolving inside a bin crate.

**Resolved use aliases are not represented** — real. The alias was parsed and then discarded on every resolved path, so nothing recorded that the file spells `Entity` as `Risk`. It is now emitted as `local_alias` — the existing transient import-resolution hint (#2082) — on the symbol edge for a symbol alias and on the file edge for a module alias.

Two things worth being explicit about, since this one is only partly solved:

- `local_alias` is popped by `extract()` after `run_language_resolvers`, by design. So the tests assert at the `extract_rust` layer, not through `extract()`.
- Its only consumer today is the module arm of the Python member-call resolver. Nothing yet resolves an aliased Rust receiver (`Risk::find()`) cross-file, because `walk_calls` deliberately skips cross-file resolution for `scoped_identifier` calls per #908. Wiring a Rust arm onto that hint is a real follow-up, but it means revisiting #908's spurious-INFERRED-edge tradeoff, which is out of scope here. This change gets the alias recorded so that work has something to read; it does not itself make aliased receivers resolve.

Suite: 4827 passed. Same 9 environment-only failures as base `281ccaa`, plus the `test_labeling` order-dependent flake noted last round. Ruff clean.

wn-mitch:
Sixth round: both real, fixed in `8ef89c6`.

**`self::` path with more than 2 segments silently dropped** — real. The own-module fallback I added last round fired only for exactly two segments, so `use self::Status::Active;` in a childless module dropped its edge entirely. It now applies to any length, attributing the item and dropping what names something inside it — the same rule `_walk_rust_segments` uses for the equivalent `crate::` path.

**Module alias compared against the wrong stem** — real. `alias != module_file.stem` is `mod` for a `mod.rs`, so a redundant `use crate::models::_entities as _entities;` was recorded as a rename. It compares against the last path segment instead, which is the module's name as written and sidesteps the `mod.rs`/plain-file distinction. Two tests: the redundant alias recording nothing, and a genuine rename of a `mod.rs` module still recorded.

Suite: 4831 passed. Same 9 environment-only failures as base `281ccaa`; the `test_labeling` flake did not reproduce this run. Ruff clean.

For what it's worth on process: rounds four through six have all been in the Rust `use`-path resolver, each one a narrower case of the last (crate root → nested auto-target modules → deep `self::`). That is the resolver getting genuinely more correct, and I am happy to keep going, but it is also a signal that the third commit is doing more work than the other two combined. If you would rather review it separately, my offer to split this into three PRs still stands.

wn-mitch:
Seventh round: both real, fixed in `973a5b8` — plus a third the review had not named, and a matrix test so this function stops being reviewed one edge case at a time.

**Top-level `bin/` treated as an auto-target dir** — real. `bin` is auto-discovered only under `src`, but it was also matched at the package root, so `bin/tool.rs` was treated as a crate root and `crate::` inside it anchored at `bin/tool/` rather than `src/`. `bin` is now separated from the package-root targets (`examples`/`tests`/`benches`) and additionally requires `src` to be a real directory, since with no `src/` the package root stands in for it.

**`super::` from crate root can resolve outside the crate source root** — the specific case named does not reproduce (`super::` from `src/lib.rs` returns None, with one keyword or four), but the underlying hole is real and adjacent: the floor check was written `if src_root is not None and …`, so with no `Cargo.toml` above the file it was skipped entirely and a single `super` resolved a sibling unbounded. No `super` is honoured without a floor now, and the floor is checked on every step rather than only inside the multi-`super` loop.

**Third one, from the sweep below:** the anchor's own module file was seeded as the fallback owner for *any* unresolved segment, so `crate::models::risk::X` in a crate with no `models/` resolved to a symbol named `models` inside `lib.rs` — inventing a symbol nothing defines, which is the same false-hub failure this PR set out to fix. Only a last segment is attributed now.

**On the review cadence.** Four rounds running have been successively narrower cases in `_resolve_rust_use_path`, and two of them were in my own prior fix rather than in the original commit. Fixing them one at a time was not converging, so I enumerated the input space instead: every crate layout Cargo supports on disk (library, lib+main, no `src/`, `src/bin/`, top-level `bin/`, `examples/` with and without a module directory) crossed with every anchor keyword and path depth. That is what surfaced the third finding above.

The matrix pins the exact resolution for all 31 answers in the library layout, so any change to this function shows up as a named diff rather than a judgement call, and asserts two invariants across the whole product: a resolution never escapes the crate (bait namesakes sit just outside the package), and a `crate::`-anchored path never attributes a module name as a symbol. It is deliberately not asserted for `self::` — a childless module may legitimately define an item sharing a name with a sibling module, so `self::helper::X` resolving to `(helper.rs, "helper")` is a real answer, not a leak.

Suite: 4849 passed. Same 9 environment-only failures as base `281ccaa`, plus the `test_labeling` order-dependent flake. Ruff clean.

Standing offer, now with more reason behind it: the Rust commit has taken four review rounds and the other two have taken none. I am happy to split it into its own PR so it can be reviewed on its own terms.

wn-mitch:
Eighth round: three fixed, one I am not making unilaterally because it contradicts a test this repo already pins.

**`source_override` changes byte offsets** — real, fixed in `050c4b5`. The mask replaced each masked character with one space, so non-ASCII markup shifted every byte offset after it. Line numbers were fine (newlines are preserved, and there is already a test for that), but tree-sitter reports byte offsets, so a template containing an accented word or an emoji misreported the column of everything in the script below. Masked characters now expand to as many spaces as their UTF-8 encoding takes. This also fixes `.vue`, which shares the masker.

```
before:  bytes_eq=False   offset of 'import': 70 -> 62
after:   bytes_eq=True    offset of 'import': 70 -> 70
```

**Empty JS destructuring patterns fall back to bogus symbol nodes** — real, fixed in `ddccf89`, and slightly worse than described. The `normalize_id` guard caught `const {} =` (normalizes to nothing), but not `const { a: {} } =` or `const { /* nothing */ } =`, which shipped nodes labelled `{ a: {} }` and `{ /* nothing */ }` — the exact bug the second commit exists to fix, surviving in the fallback path. A pattern that binds nothing now nodes nothing.

**`super` imports miss items in file-backed parent modules** — real, fixed in `e566e6f`. A module can be backed by a sibling file (`src/models.rs` owns `src/models/`), but `_rust_module_root_file` looked only for `mod.rs`/`lib.rs`/`main.rs`. So `use super::Config` resolved with a `mod.rs` parent and silently missed every item of a file-backed one — half of Rust's module system.

**Destructured require suppresses ALL destructured exports — not making this change.** The suppression is correctly scoped: it fires only when `_is_require_initializer` is true, and `export const { a, b } = other;` binds nothing for a different, older reason. That reason is pinned by an existing test, `test_exported_scalar_fix_skips_unsupported_binding_patterns`, which asserts `export const { source: renamed } = config;` and `export const [first] = items;` produce no nodes — it is named for deliberately skipping binding patterns.

I did implement the change, and it broke that test. Binding exported destructured names is arguably right — they are module API — but it reverses a documented decision that predates this PR, so it is yours to make, not something I should slip into a bug-fix PR. Reverted, and happy to do it in a follow-up if you want it.

Suite: 4855 passed. Same 9 environment-only failures as base `281ccaa`. Ruff clean.

Round eight, and the first one to reach outside the Rust resolver. Still worth splitting this into three PRs — the offer stands and I think it would have saved us several of these rounds.

wn-mitch:
Ninth round: both real, both fixed.

**Later TS SFC script parsed with the JS parser** — real, fixed in `47a6cbb`. The mask keeps every `<script>` block and they are parsed as one unit, but `lang` came from the first block alone:

```
<script lang="js"> … </script>
<script lang="ts"> const b: number = 2 </script>    ->  lang picked: js
```

The TS block's annotations then fail under the JS grammar and everything in it is dropped — and this is the common Svelte 5 shape, since `<script module>` usually comes first. `lang` is now the widest declared grammar across all blocks (TS is a superset of JS, TSX of JSX, so only the widest parses the union). Fixes `.vue` too, which shares the masker.

**super-only path underflows** — real, fixed in `671ab19`. `_rust_use_leaves` reduces a use-list `self` to its prefix, so `use super::{self, X};` yields a bare `('super',)` leaf. After the anchor keyword was consumed there was nothing left to walk, and `_walk_rust_segments` returns None on an empty remainder — so the import of the parent module was dropped. A keyword-only path names the anchor's own module:

```
('super',)          -> src/models/mod.rs
('crate',)          -> src/lib.rs
('super', 'super')  -> src/lib.rs
('self',)           -> src/models/risk.rs
```

Note this is a consequence of my own round-two fix — reducing use-list `self` to the prefix is what created the keyword-only leaf shape in the first place.

Suite: 4862 passed. Same 9 environment-only failures as base `281ccaa`, plus the `test_labeling` order-dependent flake. Ruff clean.

Nine rounds. For the record on where they have landed: two findings were factually wrong, one contradicted a test this repo already pins (`test_exported_scalar_fix_skips_unsupported_binding_patterns`, left alone and flagged for you), and the rest were real — but four of those were in fixes made during this review, not in the three commits the PR opened with. That is a review loop feeding on itself, and it is the argument for splitting: the first two commits have drawn no findings in nine rounds and could have merged eight rounds ago. Say the word and I will split the Rust commit into its own PR.

wn-mitch:
Tenth round: four real and fixed, one factually wrong — and it is the one marked `high`.

**Unterminated test file leaves suite uncollectable — no.** `tests/test_svelte_extraction.py` is 472 lines, ends with a complete statement and a trailing newline, and collects:

```
$ pytest tests/test_svelte_extraction.py --collect-only -q
24 tests collected in 0.05s
```

Line 472 is simply the last line of the file. Flagging this severity `high` on the one finding of the round that is false is worth noting, given the gate's stated grounding.

**`lang="ts"` + `lang="jsx"` picks JSX** — real, fixed in `0af6a24`. My round-nine precedence ranked TSX > JSX > TS, so a TS block beside a JSX one chose JSX and the annotations failed. TSX is the only grammar accepting both, so the choice is now made on what the blocks need rather than a fixed ranking. My own regression from last round.

**SFC `lang="tsx"` never selects the TSX parser** — real, fixed in `0af6a24`, and a genuinely separate site: `_parse_js_tree` keys the TSX grammar off the file *suffix*, which for an SFC is `.svelte`. So a `lang="tsx"` script was parsed with plain TS, misparsing JSX and dropping the calls inside it. That is the call-graph pass, not the one round nine touched.

**Dynamic require destructuring suppressed with no import edge** — real, fixed in `ea44203`. `_is_require_initializer` checked the callee name alone, so `require(name)` and `require('./' + x)` suppressed the local binding while `_require_imports_js` — which needs a string literal to emit anything — edged nothing. The name left the graph entirely. A literal specifier is now required.

**2018-edition bare path shadowed by a same-named local module** — real, fixed in `3d90c10`. `use anyhow::Result` resolved to `src/anyhow.rs`:

```
edition=2021  anyhow::Result -> None            (was src/anyhow.rs)
edition=2015  anyhow::Result -> src/anyhow.rs   (correct: 2015 paths ARE crate-relative)
```

From 2018 a bare path names a crate and a local module requires `crate::`/`self::`/`super::`, so the fallback is gated on the package edition read from `Cargo.toml`, with an absent `edition` meaning 2015 as Cargo defines it. An explicit `crate::anyhow::Result` still resolves locally under 2021 — the rule constrains bare paths only.

Suite: 4874 passed. Same 9 environment-only failures as base `281ccaa`. Ruff clean.

wn-mitch:
Eleventh round, one finding, fixed in `84e7943`.

**Redefined `_dedupe_edges` breaks existing edge-list callers** — the "breaks" part does not hold, but the underlying objection is right and the fix is a deletion.

Nothing was actually broken: my `_dedupe_edges` lived in `extract.py`, while the pre-existing function is `build.dedupe_edges`, and cli.py and watch.py import it locally inside their own functions. Different modules, no shadowing, and the suite passes either way.

But it should never have existed. `build.dedupe_edges` already collapses edges on exactly `(source, target, relation)` keeping the first — the same logic I wrote out again. And `_dedupe_edges` is precisely the alias cli.py and watch.py bind that function to, so the repo had two names one underscore apart taking different argument types: an edge list against a result dict. That is a real trap even while it happens to work.

So the rescue now calls the shared helper and my copy is gone, imported inside the function as cli.py and watch.py do. Net −14 lines. A guard test asserts `extract.py` exposes no `_dedupe_edges`, so the collision cannot come back.

Suite: 4875 passed. Same 9 environment-only failures as base `281ccaa`. Ruff clean.

wn-mitch:
Twelfth round. Four findings, all restating one claim, and the claim is right — fixed in `c083c1d`.

The four are variations on the same gate. To be precise about the part that is wrong: `_extract_generic` does emit `parse_errors`, so the `high` framing ("emits 'error', not 'parse_errors'") is not accurate — that key is what a RECOVERED parse sets, and there is already a test pinning the rescue firing on it.

But it emits `error` instead for a HARD failure — grammar missing, source unreadable — returning no tree and no nodes. So the one case where the AST contributed *nothing* was the one case the rescue skipped, while the pre-mask extractor ran its regex unconditionally and recovered those imports. That is a real regression introduced by this PR and the findings found it.

```
before:  hard error -> relations: ['dynamic_import']                    nodes: []
after:   hard error -> relations: ['dynamic_import', 'imports_from']    nodes: ['C.svelte']
clean parse still single-emits: ['dynamic_import', 'imports', 'imports_from']
```

Two parts to the fix. The gate covers both signals. And on the `error` path the file node is minted in the shape `_extract_generic` would have produced — without it the rescued edge has a dangling source and is dropped at build (#701), so gating alone would have looked fixed while changing nothing downstream. The `error` key stays on the result so extract()'s reporting is unaffected.

Three tests: the static rescue firing on a hard error, the source node being minted so no edge dangles, and the dynamic rescue surviving the same path.

Suite: 4878 passed. Same 9 environment-only failures as base `281ccaa`. Ruff clean.

wn-mitch:
Thirteenth round: one real, one not — but chasing down the one that was wrong found a bug worth more than either.

**`extract_svelte` suppresses file read errors** — real, fixed in `0f34d29`. A read failure returned a bare `{"nodes": [], "edges": []}`, so a permission or I/O error was indistinguishable from an empty component: the file contributed nothing and extract() reported success. It now returns `error`, which extract() already warns on. `extract_vue` had the identical shape and is fixed with it rather than left to be found next round.

**`_rust_package_edition` comparison assumes int; None or str would raise** — this does not hold. Every return path is an `int`: the two literal `2015` returns and `int(match.group(1))`, where the group is `(\d{4})` from the regex, so the conversion cannot fail and no `None` or `str` escapes. There is now a test asserting `isinstance(..., int)` across a manifest, a directory, a workspace root, and a nonexistent path, so the claim stays disproved rather than just argued.

**But the function was wrong for a different reason.** Writing that test made me check what a real monorepo manifest looks like, and a workspace member does not write a literal edition — it writes `edition.workspace = true` and inherits from an ancestor's `[workspace.package]`. Only the literal form was read, so **every workspace member looked like 2015**, which put the crate-relative fallback straight back in play for exactly the crates last round's edition rule exists to protect:

```
before:  edition seen: 2015   use anyhow::Result -> backend/src/anyhow.rs   (shadowed)
after:   edition seen: 2021   use anyhow::Result -> None                    (the crate)
```

Fixed in `4a1d901`, covering both the dotted and inline-table inherit forms, with a member's own explicit edition still winning and an inherit with no ancestor workspace still falling back to 2015 as Cargo defines it. Worth flagging plainly: this means last round's fix did nothing for the monorepo that prompted this PR, which is a workspace. It does now.

Suite: 4885 passed. Same 9 environment-only failures as base `281ccaa`. Ruff clean.

wn-mitch:
Fourteenth round: one real fix, three disproved, and one finding that was right in a way that made me delete two of my own earlier commits' worth of logic.

**2018 bare use paths treated as external even when they name a local module** — correct, and it goes further than the finding says. I stopped reasoning about Rust from memory and asked rustc 1.96:

```rust
// edition 2021
pub mod models;
use models::Risk;        // compiles
```

Uniform paths, stable since 1.32, resolve a `use` path in the current module scope like any other path. Worse for my earlier reasoning, the shadowing case compiles too — a crate with both `src/anyhow.rs` and `anyhow` in `[dependencies]`, where `use anyhow::Result;` binds the **local** module (it type-checks against the local `struct Result`, which anyhow's generic `Result<T>` would not).

So the premise behind rounds ten and thirteen was simply false. The "local module shadows the dependency" failure those commits were built to prevent is Rust's own documented behavior, and the gate did nothing but drop real edges. Reverted in `85c1e74`, along with the whole edition-reading apparatus — which also dissolves this round's `_rust_package_edition inherits from probe.parent` finding, since the function no longer exists. Net −95 lines. Two tests pin the rustc-verified behavior instead of my assumption.

That also means round thirteen's workspace-inheritance fix, which I described as the most valuable thing in that round, was careful work on a function that should not have existed. My error, not the reviewer's.

**Require initializer accepts string literals in non-specifier arguments** — real, fixed in `b957921`. Any string argument counted, so `require(name, 'label')` suppressed a local binding that nothing replaced. First argument only now.

**`_is_require_initializer` name check only inspects the direct callee** — does not hold. Every non-require shape is already rejected:

```
True   require('./m')            True   require('./m').sub
False  obj.require('./m')        False  notRequire('./m').sub
False  wrap(require)('./m')      False  require['x']('./m')
```

**Rust use declarations no longer emit `imports_from`** — does not hold, and this is the same finding from round two. `use crate::models::risk::Entity;` plus a `pub use` of it emits `['imports', 'imports_from', 'imports_from', 're_exports']`: the file-level edge for every leaf, exactly as before, with the symbol-level edge additional. There is now a test asserting the count so it stops recurring.

Suite: 4877 passed (down 8 — the deleted edition tests). Same 9 environment-only failures as base `281ccaa`. Ruff clean.

wn-mitch:
Fifteenth round: one real and fixed, two disproved.

**Exported destructured require bindings are suppressed** — real, and the best finding in several rounds. An exported require destructuring is a CJS barrel, but only the inbound `imports` edge was recorded, so a consumer of the barrel reached nothing:

```
before:  barrel.js -imports_from-> lib.js
         barrel.js -imports-> doWork()
after:   + barrel.js -re_exports-> lib.js
         + barrel.js -re_exports-> doWork()
```

Fixed in `418017c`. Minting a local node instead would shadow the real definition — the exact bug the require suppression exists to prevent — so the republication is recorded as `re_exports`, the relation the barrel collapse follows and the same treatment `pub use` gets on the Rust side. Both the file-level and symbol-level edges are emitted, matching what an ESM `export { x } from './lib'` produces.

**Not claiming more than that.** The equivalent ESM barrel resolves a consumer's call through to `lib.js`; the CJS one still does not, even with identical edges present. The remaining gap is in the corpus collapse pass rather than the extractor, and chasing it is scope this PR should not absorb. The structural fact is now recorded so that work has something to read.

**Plain Svelte components now require the TypeScript parser** — does not hold on either half. `tree-sitter-typescript` is a non-optional dependency in `pyproject.toml`, so nothing new is required. And the TS default costs nothing: a plain-JS component parsed both ways yields identical nodes, including the shapes a naive superset claim would trip on — a regex containing `<`, chained comparisons, an async generator. Pinned as a test in `addb4eb`.

**Bare crate-relative guess can attribute an external crate symbol to a local module** — disproved by the same rustc evidence as last round. When a same-named local module exists on disk, rustc binds the local one, so attributing it locally is correct rather than a mis-attribution. Attribution only happens when that module file exists; a bare path with no local namesake resolves to nothing.

Suite: 4882 passed. Same 9 environment-only failures as base `281ccaa`. Ruff clean.

---

@safishamsi — could you take a look when you have a moment?

This is round fifteen. The grounded gate has returned **PASS, no coupling regressions or blocking issues, on every single round**, including the first. Every finding has been advisory. The tally across those rounds: roughly 30 findings, of which 8 were factually wrong (I have posted the disproving output each time), 1 contradicted a test this repo already pins, and 4 were bugs in fixes made earlier *in this same review* rather than in the three commits the PR opened with. Two rounds ran on a false premise about Rust editions that I only settled by compiling against rustc 1.96, and unwinding that deleted 95 lines of my own work.

The review has genuinely improved the branch and I am not asking to skip it. But the first two commits (Svelte masking, JS destructured bindings) have drawn essentially no findings in fifteen rounds and have been mergeable for most of them. I offered twice to split the Rust commit out so those could land; no objection came either way, and I have kept the PR whole per that silence.

What would help is a human decision: merge on the passing gate, or tell me which of the remaining advisory findings you actually want addressed. I am happy to keep iterating — I would just rather iterate toward something you want than toward an empty findings list.

PR #3169, opened 2026-08-28
fix(php): emit a calls edge for object creation, new Foo(...) (#3115)

### LINKED ISSUE #3115: PHP: `new Foo(...)` produces no edge — `object_creation_expression` missing from `_PHP_CONFIG.call_types`
### Summary

A class that a PHP method only *instantiates* gets no edge. `new Foo(...)` parses as
`object_creation_expression`, but `_PHP_CONFIG.call_types` does not list that node type,
so the construction site is never dispatched in `walk_calls`.

This is the same gap that #2997 reported for C#, fixed in #2998 by adding
`object_creation_expression` to the C# config. Java has had it since #1373. PHP never caught up.

### Repro (graphify 0.9.50, tree_sitter_php 0.24.1)

Four files, `graphify extract . --code-only`:

```php
// src/Foo.php, src/Bar.php, src/Baz.php  (namespace App)
class Foo { public function __construct(public int $x = 0) {} }
class Bar { public function __construct(public int $y = 0) {} }
class Baz { public static function create(): self { return new self(); } }
```

```php
// src/Caller.php
class Caller
{
    public function run(Bus $bus): void
    {
        $a = new Foo(1);            // new in assignment
        $bus->dispatch(new Bar(2)); // new as argument
        $c = Baz::create();         // static call
    }
}
```

| position | edge emitted |
| --- | --- |
| `new Foo(1)` in assignment | **none** |
| `new Bar(2)` as argument | **none** |
| `Baz::create()` | `run --calls--> Baz` |

Only the static call produces an edge. Neither `new` site produces `calls`, `references`,
or any other relation — the classes exist as nodes, they are simply unreachable from the
constructing method.

### Root cause

`graphify/extract.py`, `_PHP_CONFIG`:

```python
call_types=frozenset({"function_call_expression", "member_call_expression",
                      "scoped_call_expression", "class_constant_access_expression"}),
```

`object_creation_expression` is absent. Confirmed against the grammar:

```
object_creation_expression -> new Foo(1)
object_creation_expression -> new Bar(2)
```

### Impact on a real codebase

Symfony/DDD application, 615 PHP files, 4324 nodes / 10957 edges:
**505 `new X(` sites across 178 distinct classes emit no edge.**

The practical effect is worst on message-bus code, where construction *is* the control flow.
Of 22 `$bus->dispatch(new SomeCommand(...))` sites, **0/22** have a `calls`/`references` edge
to the command they construct. Asking the graph "what happens when
`AttachRecordToMessageEmailCommand` is dispatched" returns the handler's neighbourhood
(repositories, value objects, unrelated domain events) but never the one command it chains to
— and there is no signal that the edge set is incomplete.

### Related

#1682 covers instance/member call resolution in PHP (`$this->prop->method()`), which has a
different root cause (member calls are deliberately dropped to avoid false positives). This
report is only about the construction site itself, where the callee type is written literally
at the call site and needs no inference.

abhay-codes07:
Closes #3115.

## The problem

`object_creation_expression` was absent from `_PHP_CONFIG.call_types`, so a class a PHP method merely *constructs* got no edge at all — the PHP twin of the gap Java closed in #1373 and C# in #2997/#2998. On the reporter's Symfony/DDD corpus (615 files), **505 `new X(` sites across 178 distinct classes emitted nothing**, worst on message-bus code where construction *is* the control flow: 0 of 22 `$bus->dispatch(new SomeCommand(...))` sites had any edge to the command they construct.

## The change

- `object_creation_expression` joins `_PHP_CONFIG.call_types`.
- A dedicated engine branch (placed beside C#'s) names the constructed class: PHP keeps it in a bare `name`/`qualified_name` child — no field — so the generic call path never saw it. A qualified `new \App\Bar()` names the last segment, matching how the PHP namespace pass keys classes.
- Nothing is guessed: `new $cls()` is dynamic and stays unnamed; `new self()` / `new static()` / `new parent()` name no *other* class and are skipped rather than minting junk nodes or edges.

## Tests

`tests/test_php_object_creation.py` — 6 tests mirroring the C# suite: `new` in assignment, `new` in argument position (the `$bus->dispatch(new Bar(2))` shape), qualified construction naming the last segment, dynamic/self/static producing nothing, existing static-call edges unchanged, and a cross-file dispatcher reaching its command class through `use` + `new`. With the fix reverted, 4 of 6 fail. `test_php_type_resolution` is unchanged; the full suite matches the fresh `v8` (0.9.51) baseline.

### DIFF
--- a/graphify/extract.py
+++ b/graphify/extract.py
@@ -997,7 +997,11 @@ _PHP_CONFIG = LanguageConfig(
     class_types=frozenset({"class_declaration"}),
     function_types=frozenset({"function_definition", "method_declaration"}),
     import_types=frozenset({"namespace_use_clause"}),
-    call_types=frozenset({"function_call_expression", "member_call_expression", "scoped_call_expression", "class_constant_access_expression"}),
+    # object_creation_expression joins the dispatch set so `new Foo(...)` links
+    # the constructing method to Foo (engine has a dedicated PHP branch: the
+    # class sits in a bare name/qualified_name child, not a field) - the PHP
+    # twin of Java #1373 / C# #2997 (#3115).
+    call_types=frozenset({"function_call_expression", "member_call_expression", "scoped_call_expression", "class_constant_access_expression", "object_creation_expression"}),
     static_prop_types=frozenset({"scoped_property_access_expression"}),
     helper_fn_names=frozenset({"config"}),
     container_bind_methods=frozenset({"bind", "singleton", "scoped", "instance"}),
--- a/graphify/extractors/engine.py
+++ b/graphify/extractors/engine.py
@@ -5158,6 +5158,22 @@ def _extract_generic(
                                 if child.type == "identifier":
                                     callee_name = _read_text(child, source)
                                     break
+            elif config.ts_module == "tree_sitter_php" and node.type == "object_creation_expression":
+                # PHP `new Foo(...)` keeps the class in a bare name/qualified_name
+                # child (no field), so the generic call path never names it and a
+                # class a method only constructs stayed unlinked - the PHP twin of
+                # Java's #1373 and C#'s #2997. The types this loses are the ones
+                # handed straight to something else (`$bus->dispatch(new Cmd(...))`).
+                # A qualified `new \App\Bar()` names the last segment, matching how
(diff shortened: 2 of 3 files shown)


safishamsi:

Issue #3151, opened 2026-08-28
Field-typed receivers are not inherited: a field declared on a superclass types nothing in a subclass (Java, ObjC, C#)


xiongjianxu:
## Summary

The `field -> TypeName` tables that type a `this.<field>` / `[self.<field> …]` receiver are looked up
by the **declaring class only**, with no walk up the `inherits` chain. A field declared on a
superclass therefore types nothing in a subclass, and every member call through it is dropped — even
though the `inherits` edge is sitting in the same graph and the parent's field table in the same run.

Reproduced in **Java** and **Objective-C**; the lookup in C# (`csharp_field_types`) has the same
shape. Not a cross-file data problem: Java fails even when parent and child are in one file.

`graphifyy` 0.9.51 (`v8`), Python 3.10, macOS 14.

## Reproduction — Java, one file

```java
// src/lib/Greeter.java
package lib;

public class Greeter {
    public String greet(String who) {
        return "hi " + who;
    }
}

// src/lib/Pair.java
package lib;

class PBase {
    protected Greeter greeter = new Greeter();
}

public class Pair extends PBase {
    void run() {
        this.greeter.greet("same-file-inherited");
    }
}
```

```
calls   : []                                              <- expected Pair.run -> Greeter.greet
inherits: [('src_lib_pair_pair', 'src_lib_pair_pbase')]   <- the link is right there
```

## Reproduction — Java, own field vs inherited field, same corpus

`Direct` declares its own field; `Sub` inherits an identical one from `Base`. Four files,
`src/lib/{Greeter,Base,Sub,Direct}.java`:

```java
public class Base   { protected Greeter greeter = new Greeter(); }
public class Sub extends Base { void run() { this.greeter.greet("inherited-field"); } }
public class Direct { private Greeter greeter = new Greeter();
                      void run() { this.greeter.greet("own-field"); } }
```

```console
$ graphify update . --no-cluster
calls:
  src_lib_direct_direct_run -> src_lib_greeter_greeter_greet  [INFERRED L7]
inherits:
  src_lib_sub_sub -> src_lib_base_base
```

`Direct.run` resolves, `Sub.run` does not. One call site out of two, differing only in where the
field was declared.

## Reproduction — Objective-C

Same four-class shape (`Greeter`, `Base` owning `@property Greeter *greeter`, `Sub : Base` messaging
`[self.greeter greet]`, `Direct` with its own property doing the same):

```
[...]

Issue #2591, opened 2026-08-10
ObjC: message sends to a @property or ivar receiver never resolve — only method-body locals are typed


xiongjianxu:
## Summary

`_resolve_objc_member_calls` (#1556) can only type a receiver that is a **method-body local
declaration**. The two other everyday ways an ObjC method reaches a collaborator — a
`@property` (`[self.bar doIt]`) and an instance variable (`[_ivarBar doIt]`) — never resolve,
so a cross-file `calls` edge that Java and C# both produce is silently missing for ObjC.

These are not rare shapes. On two public ObjC corpora, counting every `message_expression`
receiver by syntactic form:

| receiver form | AFNetworking (783 sends) | SDWebImage (2197 sends) | resolvable today |
|---|---|---|---|
| `self` / `super` | 19.0% | 25.9% | yes (EXTRACTED) |
| `ClassName` | 27.2% | 26.5% | yes (EXTRACTED) |
| lowercase identifier (local/param) | 26.3% | 22.6% | only if a `Foo *x` declaration is in the same body |
| **`self.property`** | **7.9%** | **8.7%** | **no** |
| **`_ivar`** | **0.1%** | **1.7%** | **no** |
| nested send `[[..] ..]`, casts, subscripts | rest | rest | out of scope here |

So roughly **8–10% of all message sends in ordinary ObjC are structurally unresolvable**, and
in delegate/manager-heavy code a property is the *usual* way a class calls its collaborator.

`graphifyy` 0.9.38 (`v8` @ `10ad921`), Python 3.12, macOS.

## Reproduction

One class, three receiver forms, same callee.

`Bar.h` / `Bar.m`
```objc
@interface Bar : NSObject
- (void)doIt;
@end
```

`Foo.h`
```objc
#import "Bar.h"
@interface Foo : NSObject
@property (nonatomic, strong) Bar *bar;
- (void)viaProperty;
- (void)viaIvar;
- (void)viaLocal;
@end
```

`Foo.m`
```objc
#import "Foo.h"
@implementation Foo {
    Bar *_ivarBar;
}
- (void)viaProperty { [self.bar doIt]; }
- (void)viaIvar     { [_ivarBar doIt]; }
- (void)viaLocal    { Bar *localBar = [[Bar alloc] init]; [localBar doIt]; }
@end
```

One of three resolves:

```
--- calls edges ---
  '-viaLocal' -> '-doIt'   INFERRED  L7
  (viaProperty and viaIvar produce nothing)
[...]

PR #3170, opened 2026-08-28
fix(objc): remap the field->type table keys with the ids they hold (#3150)

### LINKED ISSUE #3150: ObjC: #2591 self.<field> receiver typing emits zero edges via the CLI — objc_field_types keys are skipped by the #1529 id remap
## Summary

`_resolve_objc_member_calls`' field-type table (#2591, shipped in 0.9.39) is keyed by **class node id**.
It is the only extractor bucket that carries node ids as dict **keys**, and it is missing from the
#1529 symbol-prefix remap — which rewrites node ids, edge endpoints, `raw_calls[].caller_nid` and
`swift_extensions[].nid`, but not that table. The keys are minted before the remap and go stale, so
`field_types_by_class.get(cls)` misses for **every** class and `[self.<field> …]` / `[_ivar …]`
message sends resolve to nothing.

The remap fires whenever the input paths carry a common absolute prefix to strip — i.e. always via
`graphify update <dir>`. So #2591 is inert through the CLI. It works only when no prefix is stripped
(calling `extract()` with already-relative paths), which is the shape the tests use.

`graphifyy` 0.9.51 (`v8`), also present in 0.9.50 / 0.9.39. Python 3.10, macOS 14.

## Reproduction

Four files. `Direct` owns a `@property` of type `Greeter` and messages it — the exact shape #2591
added support for:

```objc
// src/Greeter.h
#import <Foundation/Foundation.h>
@interface Greeter : NSObject
- (void)greet;
@end

// src/Greeter.m
#import "Greeter.h"
@implementation Greeter
- (void)greet { NSLog(@"hi"); }
@end

// src/Direct.h
#import <Foundation/Foundation.h>
#import "Greeter.h"
@interface Direct : NSObject
@property (nonatomic, strong) Greeter *greeter;
- (void)run;
@end

// src/Direct.m
#import "Direct.h"
@implementation Direct
- (void)run { [self.greeter greet]; }
@end
```

Run the same corpus two ways:

```console
# calls edges in graphify-out/graph.json:
[]                                          # <- expected: Direct.-run -> Greeter.-greet

$ python3 -c '
import glob, tempfile
from pathlib import Path
files = sorted(glob.glob("src/*.h") + glob.glob("src/*.m"))
with tempfile.TemporaryDirectory() as t:
    res = extract(files, cache_root=Path(t), parallel=False)
n = {x["id"]: x for x in res["nodes"]}
print([(n[e["source"]]["label"], n[e["target"]]["label"], e.get("confidence"))
       for e in res["edges"] if e.get("relation") == "calls"])'
[('-run', '-greet', 'INFERRED')]            # <- resolves
```

Same files, same version, opposite results. The cached shard shows why — the table key and the node
id it is supposed to match were minted in different id spaces:

```
field table : {"path": "$graphify-root$/src/Direct.h",
               "tables": {"users_me_work_iss1_src_direct_direct": {"greeter": "Greeter"}}}
class nodes : src_direct_direct, src_greeter_greeter
```

Note the `"path"` value *is* anchored while the key next to it is not — see "not the cache" below.

Passing absolute paths straight to `extract()` also resolves (`users_me_work_iss1_src_direct_direct`
for both key and id). Key and id always agree *within* one `extract()` call; only the remap
separates them.

## Root cause

`extractors/objc.py:207` keys the bucket by `cls_nid` (minted at `:286` via
`_make_id(cls_stem, name)`) and exports it at `:565`:

```python
result["objc_field_types"] = {"path": str_path, "tables": field_tables}
```

`extract.py:6257-6278` (#1529) then canonicalizes symbol ids and rewrites every carrier it knows
about:

```python
if sym_remap:
    for n in all_nodes: ...          # node ids
    for e in all_edges: ...          # source / target
    # raw_calls carry caller_nid (a symbol id) consumed by the cross-file
    # call pass below, after this remap — rewrite it too or those edges
    # would dangle on their (stale) source.
    for rc in all_raw_calls: ...     # caller_nid
    # Same for swift_extensions[].nid (see the id_remap pass above).
    for result in per_file: ...      # swift_extensions[].nid
```

`per_file[].objc_field_types["tables"]` keys are not in that list. The consumer at
`extract.py:3681-3697` keys `field_types_by_class` off those stale strings, and the receiver lookup
in the same pass never hits. The pass' own docstring states the contract it now silently fails
(`extract.py:3669`):

> Must run after id-disambiguation so node ids and caller_nids are final.

`_merge_decl_def_classes` (`extract.py:2749-2756`) has the same shape — it explicitly rewrites
`raw_calls[].caller_nid` and not the field table — so a `.h`/`.m` pair folded into one node can
strand the table a second way.

**Why tests are green:** every path form that needs no prefix stripping keeps key and id in the same
space, and that includes a fixture list of relative paths handed to `extract()` directly.

**Why only ObjC:** `java_field_types` / `csharp_field_types` (`engine.py:2977`/`:2983`) are consumed
in-file at `engine.py:4860` and stamped onto `raw_calls` as a plain `receiver_type` **string**, so
they never cross the remap boundary. ObjC is the one language that carries an nid-keyed table across
it.

## What does NOT fix it

**Anchoring the keys in the cache.** The shard round-trip is not the cause — a cold run
(`rm -rf graphify-out`) reproduces it, and `cache._rewrite_strings` is values-only by explicit
design, with a docstring that asserts the invariant this bucket broke:

> Values only, never dict keys: no extractor bucket is keyed by a node id or a
> path (the `*_type_table` maps are `name -> type`), and rewriting keys
> could silently collide two entries into one.

That reasoning is still right (colliding two keys into one would be worse). It just is no longer true
that no bucket is keyed by a node id, so `cache.py:699`'s neighbourhood deserves a note either way.

**Keying by file path instead.** A class' entries come from two files (`@property` in the header,
ivars in the impl) and `_merge_decl_def_classes` deliberately folds them into one node; the merge at
`extract.py:3687` depends on both landing under one key.

## Suggested shape

1. Add the bucket to the #1529 rewrite block, next to `swift_extensions[].nid`:
   rebuild `result["objc_field_types"]["tables"]` with `sym_remap.get(k, k)` for each key.
   On a key collision, merge with the existing conflict rule at `extract.py:3689-3697` (drop the
   entry rather than pick) so the folded-pair case stays consistent.
2. Same for `_merge_decl_def_classes`' remap (`extract.py:2749`), which has the identical omission.
3. Consider making the invariant enforceable rather than documented — e.g. have extractors declare
   nid-keyed buckets in one place that every remap iterates, so the next such table cannot be missed.
   The two comments quoted above ("rewrite it too or those edges would dangle", "Same for
   `swift_extensions[].nid`") read like this list has been grown by discovery three times now.
4. A regression test that asserts a non-empty `calls` edge set for the fixture above **through a path
   form that triggers the remap** (absolute input paths, or `graphify update` on a temp dir) would
   have caught this and would catch the next one.

Happy to send a PR for (1) + (2) + the test if that is useful.

## Note

Two sibling reports from the same investigation: field-type tables are never walked up the `inherits`
chain (so a field declared on a superclass types nothing in a subclass — Java and ObjC both), and the
receiver type computed at these resolvers is discarded on bail-out, which is what makes cross-repo
member-call edges impossible after `merge-graphs`. Filed separately since they are a gap and a
feature; this one is a plain bug and independent of both.

abhay-codes07:
Closes #3150.

## The problem

`_resolve_objc_member_calls`' field→type table (#2591) is the one extractor bucket keyed **by class node id**. The #1529 passes rewrite node ids, edge endpoints, `raw_calls[].caller_nid` and `swift_extensions[].nid` — but never those keys. The prefix strip fires whenever the input paths share an absolute prefix, i.e. **always via `graphify update <dir>`** — so the keys went stale, `field_types_by_class.get(cls)` missed for every class, and `[self.<field> …]` receiver typing emitted **zero** edges through the CLI while working in tests, which hand `extract()` already-relative paths.

The cached shard has the same split, from the same assumption: `_rewrite_strings`' docstring says *"no extractor bucket is keyed by a node id"* — this one is — so the portability round-trip re-anchored every id except the table keys.

## The change

- Both in-process remap passes (the id remap and the symbol-prefix remap) rewrite the table keys through the same mapping (`_remap_objc_field_tables`), right beside the `swift_extensions[].nid` rewrites they mirror.
- The cache portability round-trip rewrites the keys beside each `_rewrite_strings` call (`_rewrite_id_keyed_table_keys`), on store and on load, and `_rewrite_strings`' docstring now names the exception instead of denying it.

## A note on verification

The staleness only manifests where the prefix strip actually fires — POSIX path resolution, i.e. the Linux CI and every macOS/Linux user of `graphify update` — not on this Windows dev machine, where the `source_file` re-resolution misses and ids keep their absolute form (matching keys, by accident). So: the three end-to-end tests (absolute inputs, relative inputs, warm-cache replay) exercise the fix on the Linux CI, and two platform-independent tests carry the teeth locally — the in-process remap unit test and a cache round-trip built with the real `_make_id` slug, which fails before the fix (key and node id diverge on store) and passes after.

## Tests

`tests/test_objc_field_table_remap.py` — 5 tests as above, using the issue's own four-file corpus. `test_objc*`, `test_swift*`, `test_cache` and `test_stat_index_portability` are unchanged (132 passed); the full suite matches the fresh `v8` (0.9.51) baseline.

Composes with the #3151 fix (separate PR): its inherits-chain walk reads the same table this PR keeps keyed correctly.

### DIFF
--- a/graphify/cache.py
+++ b/graphify/cache.py
@@ -788,12 +788,30 @@ def _portability_anchors(path: "str | Path", root: "str | Path") -> tuple[list[s
     return id_anchors, id_restore, path_anchors, str(root_resolved)
 
 
+def _rewrite_id_keyed_table_keys(payload: object, fn) -> None:
+    """Apply ``fn`` to objc_field_types["tables"] KEYS (#3150).
+
+    That table is the one extractor bucket keyed BY node id, which
+    :func:`_rewrite_strings` deliberately never touches - so a cached ObjC
+    shard replayed under another root kept absolute-derived class ids as keys
+    while the node ids themselves were re-anchored, and the receiver-typing
+    pass missed every class.
+    """
+    ft = payload.get("objc_field_types") if isinstance(payload, dict) else None
+    tables = ft.get("tables") if isinstance(ft, dict) else None
+    if isinstance(tables, dict):
+        ft["tables"] = {
+            (fn(k) if isinstance(k, str) else k): v for k, v in tables.items()
+        }
+
+
 def _rewrite_strings(obj: object, fn) -> None:
     """Apply ``fn`` to every string VALUE reachable in ``obj``, in place.
 
-    Values only, never dict keys: no extractor bucket is keyed by a node id or a
-    path (the ``*_type_table`` maps are ``name -> type``), and rewriting keys
-    could silently collide two entries into one.
+    Values only, never dict keys: rewriting keys blindly could silently
+    collide two entries into one. The single id-keyed bucket -
+    ``objc_field_types["tables"]`` - is handled by
+    :func:`_rewrite_id_keyed_table_keys` beside each call to this (#3150).
     """
     if isinstance(obj, dict):
         items: "Iterable" = obj.items()
@@ -845,6 +863,7 @@ def _relativize_ids_in(payload: dict, path: "str | Path", root: Path) -> None:
         return value
 
     _rewrite_strings(payload, anchor)
+    _rewrite_id_keyed_table_keys(payload, anchor)
 
 
 def _absolutize_ids_in(payload: dict, path: "str | Path", root: Path) -> None:
@@ -873,6 +892,7 @@ def _absolutize_ids_in(payload: dict, path: "str | Path", root: Path) -> None:
         return value
 
     _rewrite_strings(payload, restore)
+    _rewrite_id_keyed_table_keys(payload, restore)
 
 
 def _absolutize_source_files_in(payload: dict, root: Path) -> None:
--- a/graphify/extract.py
+++ b/graphify/extract.py
@@ -2487,6 +2487,26 @@ def _augment_js_reexport_edges(
 # Header / implementation file-extension pairing for the decl/def class merge.
 
 
+def _remap_objc_field_tables(per_file: list, mapping: dict) -> None:
+    """Rewrite objc_field_types["tables"] KEYS through an id remap (#3150).
+
+    The #2591 field->type tables are the one extractor bucket keyed BY class
+    node id. The #1529 passes rewrote node ids, edge endpoints,
+    raw_calls[].caller_nid and swift_extensions[].nid but not these keys, so
+    whenever a common absolute prefix was stripped (always via
+    `graphify update <dir>`) the table keys went stale, every
+    field_types_by_class.get(cls) missed, and [self.<field> ...] sends
+    resolved to nothing - #2591 was inert through the CLI.
+    """
+    for result in per_file:
+        ft = result.get("objc_field_types") if isinstance(result, dict) else None
+        tables = ft.get("tables") if isinstance(ft, dict) else None
+        if not isinstance(tables, dict):
+            continue
+        if any(k in mapping for k in tables):
+            ft["tables"] = {mapping.get(k, k): v for k, v in tables.items()}
+
+
 def _merge_swift_extensions(
     per_file: list[dict],
     all_nodes: list[dict],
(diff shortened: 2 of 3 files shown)


safishamsi:

PR #3171, opened 2026-08-28
fix(engine): fields declared on a superclass type receivers in a subclass (#3151)

### LINKED ISSUE #3151: Field-typed receivers are not inherited: a field declared on a superclass types nothing in a subclass (Java, ObjC, C#)

abhay-codes07:
Closes #3151.

## The problem

The `field → TypeName` tables that type a `this.<field>` / `[self.<field> …]` receiver were looked up by the **declaring class only**, with no walk up the `inherits` chain. A field declared on a superclass therefore typed nothing in a subclass, and every member call through it was dropped — with the `inherits` edge sitting in the same graph. On the reporter's corpus, `Direct.run` (own field) resolved while `Sub.run` (identical call, field inherited) did not; Java failed even with parent and child in one file.

## The change

Three layers, nearest declaration always winning:

1. **Per-file, Java and C#**: the method-scope receiver tables are now built from the class's fields folded with its *local* ancestors' (`_fields_up_chain` over the file's own `inherits` edges) — closes the same-file case for both languages.
2. **Corpus, Java**: the per-file field tables now ride the results as `member_field_tables`, keyed by **class label + source_file** — deliberately never by node id, which is exactly how the id-keyed ObjC table went stale in #3150. They are bound back to class nodes at resolution time (`_bind_member_field_tables`; an entry that stays ambiguous binds nothing), and a `this.<field>` receiver with no per-file type walks the corpus `inherits` chain. This is safe because `this.` names a field *by construction*: a local can shadow a bare name, but never `this.field` — so a bare untyped receiver still resolves to nothing, exactly as before.
3. **Corpus, Objective-C**: the #2591 tables were already merged across files; the two lookups now walk the `inherits` chain (`_field_type_up_chain`).

C#'s cross-file half is deliberately not claimed here: its raw-call receiver shape for nested member access needs its own verification, and the same-file half (layer 1) already covers the `class Pair : PBase` shape.

One existing test pinned the old behaviour (`test_inherited_field_and_chained_receiver_are_deferred` asserted the inherited call resolves to nothing); it now asserts the inherited half resolves and the chained-receiver half — which genuinely has no declared type — stays deferred.

## Tests

`tests/test_inherited_field_receivers.py` — 9 tests: the issue's Java same-file and cross-file (own vs inherited) repros; a grandparent field through a two-hop chain; a redeclared field using the nearest declaration; a bare untyped receiver still producing nothing; the C# same-file shape; the ObjC superclass-`@property` shape; and the table binder's label/file matching and its refuse-to-guess ambiguity path. With the fix reverted, 5 of 9 fail (2 skip). `test_java*`, `test_csharp*`, `test_objc*` and the receiver/member-call suites pass (206); the full suite matches the fresh `v8` (0.9.51) baseline.

### DIFF
--- a/graphify/extract.py
+++ b/graphify/extract.py
@@ -3546,6 +3546,57 @@ def _resolve_csharp_member_calls(
         })
 
 
+def _bind_member_field_tables(
+    per_file: list[dict],
+    all_nodes: list[dict],
+    *,
+    lang: str,
+) -> dict[str, dict[str, str]]:
+    """Bind the exported per-file field tables (#3151) to class node ids.
+
+    Entries are keyed by class label + source_file so they survive every id
+    remap; here they are matched back to the one class node carrying that
+    label (disambiguated by source_file basename when several share it).
+    An entry that stays ambiguous binds nothing - guessing would attach a
+    field table to the wrong class.
+    """
+    by_label: dict[str, list[dict]] = {}
+    for n in all_nodes:
+        if n.get("label") and n.get("source_file"):
+            by_label.setdefault(str(n["label"]), []).append(n)
+    bound: dict[str, dict[str, str]] = {}
+    for result in per_file:
+        for entry in (result.get("member_field_tables") or []):
+            if not isinstance(entry, dict) or entry.get("lang") != lang:
+                continue
+            label, fields = entry.get("class_label"), entry.get("fields")
+            if not label or not isinstance(fields, dict):
+                continue
+            candidates = by_label.get(str(label), [])
+            if len(candidates) > 1:
+                sf = str(entry.get("source_file") or "")
+                exact = [n for n in candidates if str(n.get("source_file")) == sf]
+                if len(exact) == 1:
+                    candidates = exact
+                else:
+                    # the node's source_file may have been relativized since
+                    # the entry was written; the basename still identifies it
+                    base = os.path.basename(sf)
+                    candidates = [
+                        n for n in candidates
+                        if os.path.basename(str(n.get("source_file"))) == base
+                    ]
+            if len(candidates) != 1:
+                continue
+            merged = bound.setdefault(candidates[0]["id"], {})
+            for fname, tname in fields.items():
+                if merged.get(fname) not in (None, tname):
+                    merged.pop(fname, None)  # cross-shard conflict: no guess
+                else:
+                    merged[fname] = tname
+    return bound
+
+
 def _resolve_java_member_calls(
     per_file: list[dict],
     all_nodes: list[dict],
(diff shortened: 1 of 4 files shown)


safishamsi:

PR #3172, opened 2026-08-28
fix(ts): type-only imports no longer manufacture Import Cycles (#3123)

### LINKED ISSUE #3123: JS/TS: `import type` / `export type` counted as runtime imports — every Import Cycle we got was a false positive
## Summary

The TS extractor emits an ordinary `imports_from` edge for type-only specifiers (`import type` / `export type`). TypeScript erases those at compile time and emits nothing at runtime, so any cycle that closes through them cannot exist.

On our repo (~7,900 nodes) this made **3 of 3** reported import cycles false. Since Import Cycles is a headline section of `GRAPH_REPORT.md`, that's a section we can't act on.

Related but distinct: #1193 is also a file-cycle false positive, but its cause is module names inside `//` comments. This one is about real `import type` statements being parsed correctly and then classified as runtime edges.

## Minimal repro

Two files, nothing between them but type-only specifiers.

`types.ts`:
```ts
export type Variables = { requestId: string };
export type { RouterType } from "./v1.js";
```

`v1.ts`:
```ts
import type { Variables } from "./types.js";
export type RouterType = { v: Variables };
export function makeRouter() { return {} as RouterType; }
```

```bash
graphify cluster-only . --no-label --no-viz
```

### Actual

```
## Import Cycles
- 2-file cycle: `types.ts -> v1.ts -> types.ts`
```

### Expected

No cycle. `tsc` emits no runtime import in either direction — both files' type-only specifiers are erased.

## What it looked like on a real codebase

All three cycles reported for our API package:

```
2-file cycle: packages/api/src/types.ts -> v1.ts -> types.ts
3-file cycle: packages/api/src/inngest.ts -> middleware-stack.ts -> types.ts -> inngest.ts
3-file cycle: packages/api/src/middleware-stack.ts -> types.ts -> v1.ts -> middleware-stack.ts
```

Every edge closing those loops is type-only — `types.ts` lines 11–12 are `export type { … } from "./v1.js"` and `"./inngest.js"`, and `v1.ts:12`, `middleware-stack.ts:20`, `inngest.ts:9` are all `import type { Variables } from "./types.js"`. The only runtime edges are `v1 -> middleware-stack` and `inngest -> middleware-stack`, and `middleware-stack` imports neither back.

The file even carries a comment saying the `export type` form was chosen deliberately for "zero runtime emit — no module graph cost". The tool reported the mitigation as the defect.

## Suggested fix

Tag the edge rather than dropping it — the type dependency is real and useful for "what references this type", it just isn't a module-graph edge:

- emit a distinct relation (e.g. `imports_type`), or keep `imports_from` with a `type_only: true` attribute;
- exclude type-only edges from cycle detection.

Worth noting the same conflation likely inflates `affected` in the other direction: a type-only importer is reported as impacted by a runtime change.

Both cases to cover: `import type { X } from "…"` and `export type { X } from "…"`, plus inline `import { type X }` / `export { type X }`.

## Environment

- graphify 0.9.50 (`uv tool install "graphifyy[sql]"`)
- Python 3.14.5, macOS arm64
- TypeScript 5.x, pnpm workspace monorepo

abhay-codes07:
Closes #3123.

## The problem

`import type { X } from` and `export type { X } from` are erased by the TypeScript compiler — no runtime emit, no module-graph edge — yet the extractor emitted ordinary `imports_from` / `re_exports` edges for them. On the reporter's ~7,900-node repo, **all 3 reported Import Cycles closed only through such edges**, and one of the flagged files carried a comment saying the `export type` form was chosen precisely for "zero runtime emit — no module graph cost". The mitigation was reported as the defect, in a headline section of `GRAPH_REPORT.md`.

## The change

The issue's own suggestion: **tag, don't drop**. The type dependency is real and useful for "what references this type", so the edges stay in the graph, stamped `type_only`, and `find_import_cycles` excludes them exactly the way it already excludes deferred `import(...)` edges (#1241 established the pattern one line above).

Both producers stamp:

- `_import_js` — the module-level `imports_from` edge and the symbol-level re-exports of a type-only statement;
- the module-resolution facts pass — the three export fact classes gain a defaulted `type_only` field, carried from the export-statement walk through to `add_edge`, so the file-level `re_exports` edges it synthesizes agree with the extractor's. The facts still participate in symbol resolution unchanged: a type import resolved through a type-only barrel is itself type-only.

Precision comes from the grammar: the statement-level `type` is a bare **anonymous** child, which keeps the two look-alikes straight — `import type from './x'` (a default binding *named* `type`, erased by nothing) has the word inside the `import_clause` and stays runtime, and a mixed `import { type B, C }` stays runtime because `C` is a runtime import.

## Tests

`tests/test_type_only_import_cycles.py` — 8 tests: the stamp on `import type` and on `export type … from` (all three edges, including the resolution pass's), the default-binding-named-`type` and mixed-specifier look-alikes staying runtime, plain imports carrying no stamp at all; and at cycle level the reporter's exact two-file repro producing no cycle, a genuine runtime cycle still reported, and a mixed one-runtime-leg/one-type-leg pair producing none. With the fix reverted, 4 of 8 fail. The JS/TS, re-export/barrel, analyze and cycle suites are unchanged (244 passed); the full suite matches the fresh `v8` (0.9.51) baseline.

### DIFF
--- a/graphify/analyze.py
+++ b/graphify/analyze.py
@@ -679,6 +679,11 @@ def find_import_cycles(
         # hard file-level cycle, so they are excluded from cycle detection (#1241).
         if data.get("deferred"):
             continue
+        # Type-only imports/re-exports (`import type` / `export type ... from`)
+        # are erased at compile time - a cycle that closes through one cannot
+        # exist at runtime (#3123). The edge itself stays in the graph.
+        if data.get("type_only"):
+            continue
 
         src_file_attr = data.get("source_file", "")
         if not isinstance(src_file_attr, str) or not src_file_attr:
--- a/graphify/extract.py
+++ b/graphify/extract.py
@@ -411,6 +411,19 @@ def _import_js(node, source: bytes, file_nid: str, stem: str, edges: list, str_p
             if not has_from:
                 return
 
+    # `import type {...} from` / `export type {...} from` are erased by the
+    # TypeScript compiler: no runtime emit, no module-graph edge. The
+    # dependency is still real for "what references this type", so the edge is
+    # stamped `type_only` rather than dropped, and Import Cycles excludes it
+    # the way it already excludes deferred `import(...)` (#1241) - on the
+    # reporter's corpus all 3 reported cycles closed only through these
+    # (#3123). The grammar keeps `import type from './x'` (a default binding
+    # NAMED type) distinct: there `type` sits inside the import_clause, not as
+    # a bare keyword child of the statement. A mixed `import { type B, C }`
+    # stays a runtime edge - C is a runtime import.
+    is_type_only = any(
+        child.type == "type" and not child.is_named for child in node.children
+    )
     resolved_path: "Path | None" = None
     module_string = None
     for child in node.children:
@@ -453,6 +466,8 @@ def _import_js(node, source: bytes, file_nid: str, stem: str, edges: list, str_p
             # back onto the importer's own variant, a phantom self-loop (#1814).
             if resolved_path is not None:
                 edge["target_file"] = str(resolved_path)
+            if is_type_only:
+                edge["type_only"] = True
             edges.append(edge)
 
     # Emit symbol-level edges for named imports/re-exports from local/aliased files.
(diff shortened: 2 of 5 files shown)


safishamsi:

PR #3185, opened 2026-08-29
fix(ts): preserve declarations after import type arguments

### LINKED ISSUE #3154: TypeScript: an `import(...)` type used as an explicit type argument is a syntax error, dropping symbols after it
TypeScript: an `import(...)` type used as an explicit **type argument** is a syntax error, dropping symbols after it

## Summary

In `.ts` and `.tsx`, a call with an explicit type argument that is an import type —
`f<typeof import("mod")>()` or `f<import("mod").Foo>()` — is reported as a syntax
error, and the symbols declared after it are not extracted.

The same import type is fine in a type alias or a type annotation. Only the
**type-argument position of a call expression** fails.

## Reproduction (graphify 0.9.51, Linux)

`src/fails.tsx`

```tsx
const a = load<typeof import("react-router")>()

export function used() { return 1 }
export function alsoUsed() { return 2 }
export class Thing {}
```

`src/works.tsx` — identical except the type argument

```tsx
const a = load<string>()

export function used() { return 1 }
export function alsoUsed() { return 2 }
export class Thing {}
```

```
$ graphify update . --force --no-cluster
    src/fails.tsx (first error at line 1, 3 symbol(s) extracted)
```

Nodes per file in the resulting `graph.json`:

| file | nodes |
|---|---|
| `src/works.tsx` | 5 |
| `src/fails.tsx` | **3** |

`used`, `alsoUsed` and `Thing` are declared identically in both files, but two of
them are missing from `fails.tsx`.

## What does and does not trigger it

| form | result |
|---|---|
| `f<typeof import("mod")>()` | ❌ syntax error (`.ts` and `.tsx`) |
| `f<import("mod").Foo>()` | ❌ syntax error |
| `f<string>()` | ✅ fine |
| `f<typeof window>()` | ✅ fine |
| `type T = typeof import("mod")` | ✅ fine |
| `let a: typeof import("mod")` | ✅ fine |

So it is not about import types in general, and not about explicit type arguments
in general — it is the combination.

It is **not** TSX-specific: the same line in a `.ts` file fails identically. It is
also unrelated to the file name (a double extension such as `foo.vitest.tsx`
parses fine on its own).

## Why it matters in practice

This is the standard shape Vitest documents for partial module mocks:

```ts
vi.mock("react-router", async importOriginal => {
  const actual = await importOriginal<typeof import("react-router")>()
  return { ...actual, Meta: () => null }
})
```

Because `vi.mock` calls are hoisted to the top of the file, the error lands on an
early line and most of the file's symbols are lost. In one private repository,
**22 of 41 component test files** are affected, some losing everything after line 1.

## Environment

- graphify 0.9.51 (installed via `uv tool install "graphifyy[sql,terraform]==0.9.51"`)
- Linux, Python 3.13
- Reproduced with a fresh `graphify update . --force --no-cluster` on a directory
  containing only the two files above

## Note

The extractor does report the problem in a warning, which made it easy to find —
thank you. The issue is that the recovery drops the remaining declarations rather
than resuming after the offending expression.

hopstreax:
## Summary

Fixes #3154.

TypeScript `import(...)` types used as explicit call-expression type arguments, such as:

```ts
f<typeof import("mod")>()
f<import("mod").Foo>()
```

were incorrectly parsed by tree-sitter as comparison expressions. The resulting recovery could swallow declarations later in the file, causing them to be missing from Graphify's extracted graph.

### What changed

* Added byte-length-preserving normalization for `import(...)` types inside explicit TypeScript call-expression type arguments.
* Applied the normalization to `.ts`, `.tsx`, `.mts`, and `.cts` extraction.
* Applied the same normalization to TypeScript/TSX Vue script blocks.
* Preserved byte offsets and newline positions so source locations remain accurate.
* Added regression coverage for TypeScript and TSX, including multiline type arguments and declarations following the affected expression.

### Validation

* `6 passed` — dedicated #3154 regression tests
* `163 passed` — relevant TypeScript/JavaScript extractor tests
* `815/815` files processed successfully during a Graphify rebuild

### DIFF
--- a/graphify/extract.py
+++ b/graphify/extract.py
@@ -1258,6 +1258,48 @@ def _extract_python_rationale(path: Path, result: dict) -> None:
             _add_rationale(stripped, lineno, file_nid)
 
 
+# ── TypeScript import type normalization (#3154) ──────────────────────────────
+# tree-sitter-typescript misparses `import(...)` types used inside explicit call-
+# expression type arguments (e.g. `f<typeof import("mod")>()` or
+# `f<import("mod").Foo>()`) as binary comparison expressions (`<` and `>`).
+# The trailing `();` generates an ERROR node, leaving an open `binary_expression`
+# that absorbs subsequent declarations as anonymous `function_expression` or `class`
+# expressions, silently dropping them from extraction. Normalizing `import(...)`
+# within generic call type arguments `<...>(...)` to a valid type identifier of
+# identical byte length keeps AST parsing clean while preserving source offsets.
+
+_TS_IMPORT_CALL_RE = re.compile(
+    rb"\bimport\s*\(\s*['\"][^'\"\r\n]+['\"]\s*\)"
+)
+_TS_IMPORT_TYPE_CALL_RE = re.compile(
+    rb"<((?:[^;{}]*?\bimport\s*\([^()]+\)[^;{}]*?)+)>(?=\s*\()"
+)
+
+
+def _normalize_ts_import_types(source: bytes) -> bytes | None:
+    """Rewrite TypeScript `import(...)` type arguments in call expressions
+    to standard type identifiers of identical byte length (#3154).
+
+    Preserves byte length, newlines, and source offsets so all downstream node
+    source_location metadata remains 100% accurate.
+    """
+    if rb"import(" not in source and rb"import (" not in source:
+        return None
+
+    def repl_type_args(m: "re.Match[bytes]") -> bytes:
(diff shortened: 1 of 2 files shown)


safishamsi:
Shipped in v0.9.52 via authorship-preserving cherry-pick so you keep contributor-graph credit. Thanks @hopstreax! Release: https://github.com/Graphify-Labs/graphify/releases/tag/v0.9.52

PR #3188, opened 2026-08-29
fix(razor): resolve @inject types across files

### LINKED ISSUE #3187: C#: Blazor `@inject` in .razor never reaches the cross-file type resolver — one stub node per file, emitted as `imports` instead of `references`
## Summary

Blazor `@inject` directives in `.razor` files never reach the C# cross-file type resolver (#1466). Each injected type gets a **fresh file-local stub node per `.razor` file**, emitted as an `imports` edge rather than a `references` edge to the resolved definition. The same type injected via a constructor in a `.cs` file resolves correctly, so this is a file-type coverage gap, not a type-kind one.

This is distinct from the BCL/stdlib case reported in #1318 (comment by @kbrilla): the types affected here are **project-defined classes with a canonical definition node already in the graph**. The resolver simply never runs on `.razor`.

Version: **0.9.51** (also reproduced on 0.9.50). Windows 11, Python 3.10.

## Reproduction

Five files, no project scaffolding needed:

`Services/WidgetService.cs`
```csharp
namespace Demo.Services;

public class WidgetService
{
    public string GetName() => "widget";
    public int Count() => 1;
}
```

`Services/Consumer.cs` — the control, constructor injection in `.cs`
```csharp
namespace Demo.Services;

public class Consumer
{
    private readonly WidgetService _widgets;
    public Consumer(WidgetService widgets) => _widgets = widgets;
    public string Use() => _widgets.GetName();
}
```

`Pages/AlphaPage.razor` (and `BetaPage.razor`, `GammaPage.razor`, identical but for the route)
```razor
@page "/alpha"
@inject WidgetService _widgets

<p>@_widgets.GetName()</p>
```

```python
from pathlib import Path
r = extract(collect_files(Path('.')), cache_root=Path('.'), parallel=False)
```

## Actual

Four nodes for one class:

```
id=pages_alphapage_razor_widgetservice                 deg=1   src=Pages/AlphaPage.razor
id=pages_betapage_razor_widgetservice                  deg=1   src=Pages/BetaPage.razor
id=pages_gammapage_razor_widgetservice                 deg=1   src=Pages/GammaPage.razor
id=services_widgetservice_demo_services_widgetservice  deg=4   src=Services/WidgetService.cs
```

Edges on the canonical node — the `.cs` control resolves correctly:
```
Consumer      --references--> WidgetService   [EXTRACTED]  src=Services/Consumer.cs
WidgetService.cs --contains--> WidgetService   [EXTRACTED]
WidgetService --method-->      .GetName()      [EXTRACTED]
WidgetService --method-->      .Count()        [EXTRACTED]
```

Edges on the razor stubs — file-local, never re-pointed:
```
pages_alphapage --imports--> pages_alphapage_razor_widgetservice
pages_betapage  --imports--> pages_betapage_razor_widgetservice
pages_gammapage --imports--> pages_gammapage_razor_widgetservice
```

## Expected

Each `@inject` should emit a `references` edge from the component to `services_widgetservice_demo_services_widgetservice`, as constructor injection does from `Consumer.cs`. No file-local stub.

## Impact at scale

On a real Blazor WASM solution (943 code files: 721 `.cs`, 196 `.razor`; 14,328-node graph):

- One DI service class → **91 nodes**: 1 canonical at degree 15, plus **90 orphans at degree 1, all 90 in `.razor`, 0 in `.cs`**.
- A widely-injected interface → **96 nodes, none above degree 1** — no canonical node survives at all.
- Across project-defined service types: **99 razor orphans, 0 `.cs` orphans.** Clean split along file type.
- **95% of all `.razor` nodes end up degree ≤1** (3,716 of 3,919), against 64% for `.cs`.
- Graph-wide, 380 labels appear >5× as separate nodes, involving **8,420 nodes = 59% of the graph**.

Downstream this inflates the community count, produces thousands of spurious "isolated nodes", and makes centrality/cohesion figures unusable — a `DbContext` community scored 0.0135 cohesion purely because it was a star of unlinked stubs. Analytically, the effect is that a shared service looks unused rather than central.

## Notes

Related but not the same: #1318 (umbrella; the BCL/stdlib branch of this), #207 (proposed canonicalization phase), #1186. If you would rather fold this into #1318 as another branch of the umbrella, happy for it to be closed as a duplicate — filing separately since the `.razor`/`.cs` split is a concrete, testable boundary that the existing reports don't name.

hopstreax:
## Summary

Fixes #3187.

Blazor `.razor` and `.cshtml` `@inject` directives were previously emitted as file-local `imports` stubs and never reached the C# cross-file type resolver. This caused each Razor file to create a duplicate node for the same project-defined service.

This change:

* Emits `@inject` as a `references` edge with C# resolver metadata.
* Creates unresolved `@inject` stubs without a Razor-local `source_file`.
* Allows `.razor` and `.cshtml` files to participate in the existing C# type-resolution pass.
* Preserves `@using` as an `imports` edge while adding the metadata required for namespace scoping.
* Protects Razor C# references from premature generic stub rewiring.
* Adds end-to-end regression tests for single and multiple Razor consumers, C# consumers, explicit `@using`, and qualified namespaces.

## Validation

* `tests/test_dotnet.py tests/test_csharp_type_resolution.py` — **84 passed**
* `tests/test_languages.py -k "razor or csharp"` — **20 passed**
* `tests/test_extract.py` — **201 passed, 1 known Windows MAX_PATH failure**
* `git diff --check` — **clean**

The known `tests/test_extract.py` failure is Windows-specific (`MAX_PATH`) and unrelated to this change.

### DIFF
--- a/graphify/extract.py
+++ b/graphify/extract.py
@@ -2447,7 +2447,7 @@ def _rewire_unique_stub_nodes(nodes: list[dict], edges: list[dict]) -> None:
         return target_fam is not None and target_fam != edge_fam
     for edge in edges:
         is_csharp_scoped_edge = (
-            str(edge.get("source_file", "")).endswith(".cs")
+            str(edge.get("source_file", "")).endswith((".cs", ".razor", ".cshtml"))
             and edge.get("relation") in csharp_scoped_relations
         )
         source = edge.get("source")
@@ -6476,9 +6476,10 @@ def extract(
     # Cross-file C# type-reference resolution: re-point dangling inherits/implements/
     # references edges left on shadow stubs, disambiguating same-named types by the
     # referencing file's `using` directives + enclosing namespace (mirrors Java #1318).
-    cs_paths = [p for p in paths if p.suffix == ".cs"]
+    _DOTNET_TYPE_EXTS = {".cs", ".razor", ".cshtml"}
+    cs_paths = [p for p in paths if p.suffix.lower() in _DOTNET_TYPE_EXTS]
     if cs_paths:
-        cs_results = [r for r, p in zip(per_file, paths) if p.suffix == ".cs"]
+        cs_results = [r for r, p in zip(per_file, paths) if p.suffix.lower() in _DOTNET_TYPE_EXTS]
         try:
             _resolve_csharp_type_references(cs_results, cs_paths, all_nodes, all_edges)
         except Exception as exc:
--- a/graphify/extractors/csharp.py
+++ b/graphify/extractors/csharp.py
@@ -149,8 +149,15 @@ def _resolve_cross_file_csharp_imports(
     ]
 
 
+_DOTNET_SOURCE_EXTS = (".cs", ".razor", ".cshtml")
+
+
 def _is_cs_file(value: object) -> bool:
     return isinstance(value, str) and value.endswith(".cs")
+
+
+def _is_dotnet_source_file(value: object) -> bool:
+    return isinstance(value, str) and value.endswith(_DOTNET_SOURCE_EXTS)
 
 
 def _metadata(value: object) -> dict:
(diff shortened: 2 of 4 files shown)


safishamsi:
Shipped in v0.9.52 via authorship-preserving cherry-pick. Thanks @hopstreax! This partially addresses #3187 (cross-file @inject now resolves); a bare @inject of a file-scoped-namespace type still dangles, so #3187 stays open for that follow-up. Release: https://github.com/Graphify-Labs/graphify/releases/tag/v0.9.52

PR #3191, opened 2026-08-29
fix(watch): reconcile markdown links to semantic-backed documents (#3190)

### LINKED ISSUE #3190: graphify update drops authored [[wikilinks]] whose target document is semantic-backed
### Summary

When a Markdown document already carries semantic (LLM) nodes, #1915 and its #1954 follow-up correctly
exclude it from AST quick-scan, so it never receives a canonical page node. But `[[wikilinks]]` **in
other notes** still resolve to that canonical page id — which does not exist — and full reconcile
discards the edge silently.

Nothing is missing from the graph: the target is present as its preserved semantic document node,
under a different id. There is no mapping from the canonical page id to that preserved representative.

### Environment

```
graphify 0.9.51, commit 281ccaa4ff38aaef3f19e823fb7645e19b28f591 (unmodified)
Python 3.13.14 · tree-sitter 0.25.2 · networkx 3.6.1 · Windows 11
run as: graphify update <root>
```

### Expected vs actual

**Expected:** a uniquely resolvable `[[link]]` to an existing document yields a `references` edge to
that document.
**Actual:** if the target is semantic-backed, no edge is produced and nothing is reported.

### Mechanism

Extraction resolves the link to the canonical page id for that path. That node does not exist, because
the target is semantic-backed and was never AST-quick-scanned. Reconcile then discards the edge —
silently, and **without leaving a dangling endpoint behind**, which is why aggregate integrity checks
stay clean.

### Reproduce

```
root/
  hub.md      # contains: [[topic]]
  topic.md    # already carries semantic (LLM) nodes in the existing graph
```

`graphify update root/` → the `hub.md → topic.md` `references` edge is absent, while `topic.md`'s
semantic document node is present.

Nine self-contained tests covering this and the surrounding contract are in the linked PR; **three
fail on unmodified `281ccaa4`.**

### Why it is easy to miss

Node, semantic and hyperedge counts are all preserved, and no dangling endpoint appears. On one
~400-file corpus the **link total was identical before and after** — 82 edges removed, 82 added — so
every aggregate check passed. The loss is visible only by diffing edge identity, or by resolving
authored links against the graph.

### Measured impact

404-file Markdown corpus, exhaustive sweep of 1,868 authored resolvable link pairs against the
post-update graph:

| | sweep-resolved `references` | remainder |
| --- | ---: | ---: |
| official `281ccaa4` | 1,739 | **129** |
| with the linked fix | 1,834 | 34 |

The remaining 34 are **not losses**: 27 have **no unique representative** and retain their existing
edges; 7 are already connected by a semantic relation, which a simple (non-multi) `Graph` cannot hold
in parallel with `references`. **Disconnected authored pairs: 0.**

### Suggested fix

For a fresh `references` edge whose canonical page target is absent, re-point to a unique preserved
representative for the same normalized source, in this priority:

1. exact canonical id
2. unique `node_kind=page`
3. unique non-AST `file_type=document`
4. otherwise **no mapping — report ambiguity**

Never map concept-only sources, ambiguous sources, nonexistent or already-valid targets, or
non-`references` relations. Preserve existing edges for ambiguous targets rather than pruning them.

A tested implementation is linked.

### Related

- **#1915 / #1954** — the semantic-doc gate this regresses against. Both are correct; the gap is the
  missing representative mapping, not the gate.
- **#2413** — adjacent but distinct. That concerns *which path* a bare link resolves to (sibling-only).
  This concerns a target whose path resolves correctly but whose page node was never built: **12 of the
  observed losses are same-folder sibling links**, which sibling-only resolution handles correctly.
- **#1695** — already-open semantic extraction / entity-ID nondeterminism. Context and amplifier for
  the losses reported here, not the cause of them, and not duplicated by this report.
- **#1711** — old surviving edges are severed when a re-extracted endpoint's ID changes.
- **How this differs from both.** #1711/#1695 sever an edge that already existed, because an endpoint
  identity moved. Here nothing moved: a *fresh* `references` edge targets a canonical page ID that was
  **never created**, because #1915/#1954 skip AST page creation for semantic-backed documents. Same
  silent drop site, different root cause. The candidate PR provides a conservative representative
  rescue for this specific case only.

logan683:
Fixes #3190.

### Problem

When a Markdown document already carries semantic (LLM) nodes, #1915 and #1954 correctly exclude it
from AST quick-scan, so it never receives a canonical page node. `[[wikilinks]]` in other notes still
resolve to that canonical id, and full reconcile discards the resulting edge silently — leaving no
dangling endpoint, so aggregate integrity checks stay clean.

### Root cause

`watch.py` excludes semantic-backed documents from AST quick-scan (#1915), and #1954 extended that gate
to documents whose semantic layer is concept- or rationale-typed. Correct for the document itself — its
semantic nodes are the sole representation. But another note's fresh `[[wikilink]]` still resolves to
the deterministic page id for that path, and reconcile has no mapping from canonical page id to the
preserved representative, so the edge is discarded.

### What changed

One new private helper, `_reconcile_markdown_links(...)`, implementing six cases:

1. authored link removed from its source → prune the stale edge
2. link present, both representatives unique → retain if correct, else canonicalize
3. link present, either representative ambiguous → **preserve the existing edge, add nothing**
4. endpoint identity by unordered key; reverse-stored correct edges keep order and attributes
5. pair already occupied by any relation → never overwritten with `references`
6. representative priority: exact canonical id → unique page → unique non-AST document → ambiguous

### Evidence

```
base                281ccaa4ff38aaef3f19e823fb7645e19b28f591
diff                +417 / -0, two files (graphify/watch.py, tests/test_watch.py)

nine new tests      official 281ccaa4 : 6 passed, 3 FAILED
                    patched           : 9 passed
                    an earlier rejected approach: the 3 guard tests FAIL 3/3
                    (ambiguous target, ambiguous source, reverse endpoint order)

tests/test_watch.py 128 passed / 9 skipped / 2 pre-existing Windows CWD failures
supporting suites   476 passed / 9 skipped
full suite, same host, session and interpreter:
                    official : 5,121 collected, 26 failed
                    patched  : 5,130 collected, 26 failed
                    mechanical failure-ID diff: 26 shared · 0 patch-only · 0 official-only
                    collection delta is exactly the nine new tests

404-file corpus A/B from an identical input graph:
  nodes             4,776 → 4,776   (identical id sets)
  links             6,088 → 6,137   (+95 references, -46, net +49)
  hyperedges          120 → 120     (identical id sets; dual slots deep-equal)
  semantic         950/983 → 950/983
  endpoint-order flips: 0 · dangling endpoints: 0 both sides

  46 removals: 41 re-points; 5 file pairs disconnected — four are links deleted
  from source, one a stale duplicate whose replacement already existed.

lifecycle           add/edit/delete/move/rename/swap/directory-rename, Unicode,
                    duplicate headings, dash variants, same basenames, long paths,
                    frontmatter-only edits — semantics and hyperedges preserved throughout

performance         +1.027 s on 404 Markdown files (4.639 s → 5.666 s, +22%) on a
                    no-change run — the helper re-reads authored links per reconcile
```

The overhead is real and stated up front. Change-gating is possible if preferred; correctness came
first.

### DIFF
--- a/graphify/watch.py
+++ b/graphify/watch.py
@@ -2,6 +2,7 @@
 from __future__ import annotations
 import contextlib
 import json
+import logging
 import os
 import posixpath
 import re
@@ -12,6 +13,8 @@ from typing import Callable
 
 # Single source of truth in graphify.paths (#1423); re-exported as _GRAPHIFY_OUT.
 from graphify.paths import GRAPHIFY_OUT as _GRAPHIFY_OUT, is_absolute_any_platform
+
+logger = logging.getLogger(__name__)
 _PENDING_FILENAME = ".pending_changes"
 _PENDING_DRAIN_MAX_PASSES = 20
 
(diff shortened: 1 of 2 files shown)


safishamsi:
Shipped in v0.9.52 via authorship-preserving cherry-pick so you keep contributor-graph credit. Thanks @logan683! Release: https://github.com/Graphify-Labs/graphify/releases/tag/v0.9.52

PR #3196, opened 2026-08-29
fix(wiki): say when a god-node relation group was truncated (#3127)

### LINKED ISSUE #3127: Wiki export silently truncates relation groups over 20 entries with no indicator
`_god_node_article` in `graphify/wiki.py` caps each relation-type group at 20 entries with no truncation notice:

```python
for t in targets[:20]:
    lines.append(f"- {t}")
```

If a god node has more than 20 edges of the *same* relation type, the article's header still reports the correct total degree, but the body silently drops the lowest-degree entries past 20 — with nothing telling the reader anything was cut.

**Repro:** any node with >20 same-relation edges. Concretely: a node with 26 `references`-type edges + 3 `shares_data_with` edges (29 total, matching the header) only lists 23 entries in the body — 6 silently missing from the `references` section.

**Fix:** remove the cap (or add an explicit "... and N more" line if a cap is intentionally kept for very high-degree nodes):

```python
for t in targets:
    lines.append(f"- {t}")
```

abhay-codes07:
Closes #3127.

## The problem

`_god_node_article` caps each relation-type group at 20 entries. The article's header still reports the full degree, but the body silently dropped everything past 20 — a node with 26 `references` edges + 3 `shares_data_with` listed 23 entries under a "29 connections" header, with nothing telling the reader six were cut. Since neighbours are degree-sorted, what vanished was always the low-degree tail.

## The change

The cap stays (it is what keeps god-node articles readable). An over-cap group now ends with an explicit indicator:

```
- *…and 6 more `references` connection(s) not listed (lowest-degree first to go)*
```

so the body and the header agree, and the reader knows both that entries were cut and which kind.

## Tests

`tests/test_wiki_truncation_indicator.py` — 3 tests: an over-cap group names exactly how much was cut while the untouched small group carries no indicator; the header degree and the body now reconcile; and groups at or under the cap have no indicator. With the fix reverted, 1 of 3 fails (the other two pass by construction — they pin the cap and the header, which were already right). The full wiki suite is unchanged (49 passed together); the full suite matches the fresh `v8` (0.9.52) baseline.

### DIFF
--- a/graphify/wiki.py
+++ b/graphify/wiki.py
@@ -213,6 +213,15 @@ def _god_node_article(G: nx.Graph, nid: str, labels: dict[int, str], node_commun
         lines.append(f"### {rel}")
         for t in targets[:20]:
             lines.append(f"- {t}")
+        # The cap keeps god-node articles readable, but silently dropping the
+        # tail made the body disagree with the header's degree count with
+        # nothing telling the reader anything was cut (#3127). Entries are
+        # degree-sorted, so what is hidden is the low-degree tail.
+        if len(targets) > 20:
+            lines.append(
+                f"- *…and {len(targets) - 20} more `{rel}` "
+                f"connection(s) not listed (lowest-degree first to go)*"
+            )
         lines.append("")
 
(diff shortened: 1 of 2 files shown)


safishamsi:

Issue #2129, opened 2026-07-23
GRAPH_REPORT.md: community 'shown' count regressed (#703), plus report root '.', MD022 spacing, and unescaped node ids


kaperoo:
Four defects in `graphify/report.py` output, found while committing a graph into a repo and putting the generated report under review. Grouped because they share a file; happy to split if you'd rather.

Environment: `graphifyy 0.9.25` (pipx, Python 3.11.4, Windows 11), 394-file corpus (374 code / 21 doc+image), AST + subagent semantic extraction, default flags.

---

## 1. The `#703` community-count mismatch has regressed 🔴

`#703` reported header/body/rendered counts disagreeing; it was closed by making header and body share `len(communities)` and one `thin_count` variable. That reconciled *header vs body heading*, but `shown` still overstates what is actually rendered, so the original user-visible symptom is back.

My report says:

```
- 3945 nodes · 8840 edges · 277 communities (195 shown, 82 thin omitted)
## Communities (277 total, 82 thin omitted)
```

Actual `### Community` headings present: **149**, not 195.

### Root cause

`thin_count_summary` is guarded by `0 <`, so communities whose nodes are *all* file-nodes are excluded from the thin tally:

```python
# report.py:119
thin_count_summary = sum(
    1 for nodes in communities.values()
    if 0 < sum(1 for n in nodes if not _ifn(G, n)) < min_community_size
)
# report.py:123
shown_count = len(communities) - thin_count_summary
```

But the render loop skips those same communities:

```python
# report.py:227-231
real_nodes = [n for n in nodes if not _ifn(G, n)]
if not real_nodes:
    continue                      # <-- skipped, but NOT counted as omitted
if len(real_nodes) < min_community_size:
    continue                      # <-- skipped AND counted
```

So all-file-node communities are omitted from the body while still counted as shown. In my corpus that's exactly 46 of them:

| | count |
|---|---|
| total communities | 277 |
| rendered (`real >= 3`) | 149 |
| thin (`0 < real < 3`) | 82 |
| **zero real nodes** | **46** ← skipped, uncounted |

`149 + 82 + 46 = 277`. Reported `shown` is `277 − 82 = 195`, which is 46 too high.

Reproduced on the previous build of the same repo too (262 communities: 140 rendered, 76 thin, 46 zero-real, advertised 186) — so it scales with the corpus rather than being a one-off.

### Suggested fix

[...]

PR #3197, opened 2026-08-29
fix(report): make the report's headline numbers agree with themselves (#3148)

### LINKED ISSUE #3148: GRAPH_REPORT.md: thin-community count uses a hardcoded threshold that diverges from --min-community-size
`GRAPH_REPORT.md` can show a different "thin community" count in its `## Knowledge Gaps` section than in its `## Summary` and `## Communities` headers for the very same run, whenever the run doesn't use the default community-size threshold. A separate figure in the same report — the "isolated node(s)" count — also won't match a straightforward recount of degree-<=1 nodes straight from the accompanying `graph.json`. Either way, a reader auditing the report's own headline numbers, or diffing them against the graph data, sees stats that don't add up, which undermines trust in the rest of the report.

**Repro**

1. `graphify update . --min-community-size=5` (anything other than the default `3`) on a corpus with at least one community whose "real" node count falls between the two thresholds.
2. Compare `## Summary`'s `(<shown> shown, <thin> thin omitted)` and `## Communities (<total> total, <thin> thin omitted)` figures against `## Knowledge Gaps`'s `"<K> thin communities (<min_community_size> nodes) omitted"` line — `K` is computed against a different threshold than the number the label itself prints.
3. Separately, for the "isolated node(s)" figure: recount nodes with `degree <= 1` directly from `graph.json` and compare to the reported count — they won't agree, because the report's isolated-node list additionally excludes file-hub nodes, concept nodes, and rationale nodes, which isn't stated anywhere in the report output.

**Evidence (HEAD `43d54acbfa9e731f7a592bb582c1f4b9d48ed73e`)**

- `report.py:144-148` — `thin_count_summary` (feeds the `## Summary` and `## Communities` header counts) is gated by the caller-supplied `min_community_size` parameter.
- `report.py:287-290` — `thin_communities` (feeds `## Knowledge Gaps`) is gated by a hardcoded literal `3`, not `min_community_size` — while the label text right next to it, `report.py:301`, still prints `"(<{min_community_size} nodes) omitted"`.
- `cli.py:1858` and `cli.py:2173` confirm `min_community_size` is a CLI-configurable flag (`--min-community-size`, default `3`) that is threaded into `generate()`, so the two thresholds are reachably different, not just a theoretical edge case.
- `report.py:280-286` — the isolated-node list filters out file nodes and concept nodes (`analyze.py:63` `_is_file_node`, `analyze.py:165` `_is_concept_node`) plus any node with `file_type == "rationale"`. Per `_is_file_node`'s own docstring, this exclusion is deliberate for "knowledge gap reporting" — but the report text doesn't say so, so a plain `degree <= 1` recount from `graph.json` will disagree with it.

**Related prior art:** `#2129` (closed, `NOT_PLANNED`, self-withdrawn by its own reporter, no fix landed) reported a related divergence in this same community-counting code — its item 1 found the render loop and `thin_count_summary` use different guards for zero-real-node communities, causing the `shown` count to overstate what's actually rendered. Worth checking together with the hardcoded-threshold issue here, since both live in the same handful of lines.

**Suggested fix**

Compute the Knowledge Gaps thin-community count with the same `min_community_size`-gated predicate used for the Summary/Communities counts (share one helper), rather than a separate hardcoded `3`. For the isolated-node count, either state the file/concept/rationale exclusion inline in the report text, or additionally print the raw `degree <= 1` count so it's auditable against `graph.json` without reading the source.

Found while maintaining a downstream consumer of the generated report.

abhay-codes07:
Closes #3148.

## The problem

Three self-disagreements in `GRAPH_REPORT.md`, all within a few lines of `report.py`:

1. The **Knowledge Gaps** thin-community count was gated by a hardcoded `3` while the label text *beside it* printed `min_community_size` — and the **Summary**/**Communities** headers counted with `min_community_size`. Any run with `--min-community-size` other than 3 published two different figures for the same thing, one of them mislabelled.
2. `shown` was computed as total-minus-thin, which counts communities with **zero** real nodes — communities the render loop skips entirely (#2129's residual, reported there and never fixed). The "N shown" figure overstated what the report actually renders.
3. The **isolated node(s)** figure silently excludes file, concept and rationale nodes (deliberate per `_is_file_node`'s docstring, but stated nowhere in the output), so a reader auditing it against a plain `degree <= 1` recount from `graph.json` always found a mismatch.

## The change

- One `_real_count` predicate feeds every figure: *thin* is `0 < real < min_community_size` everywhere (the Gaps count now uses the caller's threshold, matching its own label), and *shown* counts exactly what the render loop renders (`real >= min_community_size`).
- The isolated line now says what it counts and prints the auditable raw figure beside it:

  > These have ≤1 connection — possible missing edges or undocumented components. (Counts symbols only; N node(s) total have ≤1 connection when file, concept and rationale nodes are included.)

## Tests

`tests/test_report_gap_thresholds.py` — 4 tests on a synthetic graph with a big community, a between-thresholds community, a zero-real-node community and a lonely concept node: Summary and Gaps agree at `--min-community-size 5` and the Gaps label matches its own count; the default threshold reports nothing thin; `shown` counts only rendered communities; and the isolated figure reconciles exactly with a raw `degree<=1` recount of the same graph. With the fix reverted, 3 of 4 fail. The report suites are unchanged (20 passed together); the full suite matches the fresh `v8` (0.9.52) baseline.

### DIFF
--- a/graphify/report.py
+++ b/graphify/report.py
@@ -139,13 +139,25 @@ def generate(
         ]
 
     from .analyze import _is_file_node as _ifn
+
+    def _real_count(nodes) -> int:
+        return sum(1 for n in nodes if not _ifn(G, n))
+
     non_empty = {cid: nodes for cid, nodes in communities.items()
                  if any(not _ifn(G, n) for n in nodes)}
+    # One predicate for every figure the report prints about itself (#3148):
+    # "thin" is 0 < real < min_community_size, and "shown" is what the render
+    # loop below actually renders (real >= min_community_size) - previously
+    # shown was total-thin, which also counted communities with ZERO real
+    # nodes that the loop skips, overstating the count (#2129's residual).
     thin_count_summary = sum(
         1 for nodes in communities.values()
-        if 0 < sum(1 for n in nodes if not _ifn(G, n)) < min_community_size
+        if 0 < _real_count(nodes) < min_community_size
     )
-    shown_count = len(communities) - thin_count_summary
+    shown_count = sum(
+        1 for nodes in communities.values()
+        if _real_count(nodes) >= min_community_size
+    )
 
     lines += [
         "",
@@ -284,9 +296,13 @@ def generate(
         and not _is_concept_node(G, n)
         and G.nodes[n].get("file_type") != "rationale"
     ]
+    # Same threshold the Summary and Communities headers used (#3148): this
+    # was a hardcoded 3, so with --min-community-size anything else the count
+    # here disagreed with the label text beside it, which already printed
+    # min_community_size.
     thin_communities = {
         cid: nodes for cid, nodes in communities.items()
-        if 0 < sum(1 for n in nodes if not _is_file_node(G, n)) < 3
+        if 0 < sum(1 for n in nodes if not _is_file_node(G, n)) < min_community_size
     }
     gap_count = len(isolated) + len(thin_communities)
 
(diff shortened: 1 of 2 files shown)


safishamsi:

PR #3199, opened 2026-08-29
fix(cli): hook-guard search detection matches executed commands, not prose (#3121)

### LINKED ISSUE #3121: hook-guard search substring-matches the whole command string, firing on non-searches
`_run_hook_guard` in `graphify/cli.py` decides whether a Bash tool call is a search with:

```python
is_bash_search = any(tok in cmd_str for tok in (
    "grep", "ripgrep", "rg ", "find ", "fd ", "ack ", "ag "))
```

That is a plain substring test over the entire command string, including quoted arguments and heredoc bodies. Observed on graphifyy 0.9.48:

| Command | Result | Correct? |
|---|---|---|
| `grep -rn foo .` | fires | yes |
| `rg foo` | fires | yes |
| `git commit -m x` | quiet | yes |
| `git log -S foo` | quiet | yes |
| `git commit -m "add flag support"` | **fires** | no — `"flag "` contains `"ag "` |
| `gh pr create --body "you can find it here"` | **fires** | no — prose contains `"find "` |
| `echo "see grep docs"` | **fires** | no |
| heredoc whose body mentions grep | **fires** | no |

The `"ag "` and `"fd "` tokens are the worst offenders: any English word ending in `ag` or `fd` followed by a space matches. `flag`, `tag`, `diag`, `bag` all trip it.

The heredoc case is self-inflicting for anyone whose project documents its own use of graphify — writing a design doc that mentions grep fires the guard on the write.

Each false positive injects a context line into a tool call where graphify has nothing to contribute, and trains the agent to skim the line, weakening it where it is correct.

Suggested fix: decide on the command's executed tokens rather than the raw string — strip heredoc bodies and quoted spans, split on shell operators (`|`, `||`, `&&`, `;`, `&`, newline, `(`, `do`, `then`), skip leading `VAR=value` assignments, and check whether the first remaining word of any segment basenames to a search tool. Worth reconsidering whether `find` should fire at all: it locates files by path, which the AST index does not cover.

Reported from a downstream project that has worked around it with a wrapper script.

abhay-codes07:
Closes #3121.

## The problem

The Bash search nudge decided "is this a search?" with a plain substring scan over the whole command string — quoted arguments and heredoc bodies included:

```python
is_bash_search = any(tok in cmd_str for tok in (
    "grep", "ripgrep", "rg ", "find ", "fd ", "ack ", "ag "))
```

So `git commit -m "add flag support"` fired (`"flag "` contains `"ag "`), `gh pr create --body "you can find it here"` fired, and writing a design doc via heredoc fired if its body merely mentioned grep — the self-inflicting case for anyone documenting their own graphify use. Every false positive injects a nudge where graphify has nothing to say, and trains the agent to skim the line where it *is* right.

## The change

`_bash_invokes_search` decides on the command's **executed tokens**, the shape the issue suggested:

- heredoc bodies are dropped (`<<WORD` through the terminator line; an unterminated heredoc drops to the end);
- quoted spans are dropped — anything quoted is an argument, never the executable;
- the rest splits on shell operators (`| ; & newline $( ` ( ) { }`), and the executable at each command position is matched against the search set (`grep`/`egrep`/`fgrep`/`zgrep`/`rg`/`ripgrep`/`find`/`fd`/`ack`/`ag`), with `VAR=x` prefixes and wrappers (`sudo`, `xargs`, `env`, `command`, `exec`, `nohup`, `time`, `timeout`, …) skipped, paths and `.exe` suffixes normalised, and `git grep` recognised.

Fails toward "not a search": the guard is a nudge, and a missed nudge costs less than an ignored one.

One judgment call in the existing tests: `test_hook_guard.py` listed `pgrep -f server` under commands that MUST nudge, annotated `# contains 'grep'` — pinning the substring mechanism, not a behaviour. `pgrep` searches *processes*, which the graph cannot answer, so the case moves to the must-stay-silent list with a comment saying why.

## Tests

`tests/test_hook_guard_token_match.py` — 34 tests: 18 real search shapes that must fire (paths, `.exe`, pipes, `&&`, `;`, command substitution, `sudo`/`xargs`/env-prefix, `git grep`, `git -C repo grep`) and 16 that must stay quiet (the issue's exact false positives, look-alike words, quoted command names, a heredoc'd design doc, a search *after* a heredoc still firing, and an unterminated heredoc never firing from its body). With the fix reverted the module skips (the helper doesn't exist pre-fix); `tests/test_hook_guard.py`'s 100+ behavioural cases pass against the new matcher (103 together). The full suite matches the fresh `v8` (0.9.52) baseline.

### DIFF
--- a/graphify/cli.py
+++ b/graphify/cli.py
@@ -677,6 +677,79 @@ def _mark_session_denied(session_id: str) -> bool:
         return False
 
 
+_SEARCH_COMMANDS = frozenset({
+    "grep", "egrep", "fgrep", "zgrep", "rg", "ripgrep", "find", "fd", "ack", "ag",
+})
+# Prefix words that wrap another command; the real executable follows them.
+_COMMAND_WRAPPERS = frozenset({
+    "sudo", "command", "exec", "nohup", "time", "nice", "ionice", "env",
+    "xargs", "timeout", "stdbuf", "doas",
+})
+_HEREDOC_OPEN_RE = re.compile(r"<<-?\s*(['\"]?)(\w+)\1")
+
+
+def _bash_invokes_search(cmd_str: str) -> bool:
+    """Whether a Bash command actually RUNS a search tool (#3121).
+
+    The old test was a plain substring scan over the whole command string,
+    including quoted arguments and heredoc bodies - so `git commit -m "add
+    flag support"` fired ("flag " contains "ag "), prose containing "find "
+    fired, and a design doc written via heredoc fired if its body mentioned
+    grep. Every false positive injects a nudge where graphify has nothing to
+    contribute and trains the agent to skim the line.
+
+    Decide on the command's executed tokens instead: drop heredoc bodies and
+    quoted spans, split on shell operators, and match the executable at each
+    command position (wrappers like sudo/xargs/env skipped; `git grep` and
+    `VAR=x grep ...` still count; a search-tool name inside prose does not).
+    """
+    text = cmd_str
+    # Drop heredoc bodies: from the line after `<<WORD` through the line that
+    # is exactly WORD. An unterminated heredoc drops to the end of the string.
+    m = _HEREDOC_OPEN_RE.search(text)
+    while m:
+        nl_idx = text.find("\n", m.end())
+        if nl_idx == -1:
+            break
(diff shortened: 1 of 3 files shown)


safishamsi:

PR #3201, opened 2026-08-29
fix(install): actionable stale-skill warning; back up a diverged SKILL.md (#3144)

### LINKED ISSUE #3144: install: --platform X never writes .graphify_version (permanent upgrade warning), and install silently overwrites a modified SKILL.md
Two separate defects in `graphify install`, same command, different files. Filing together since a fix likely touches the same area; happy to split if you would rather.

Same family as #2670 (per-platform install doing only part of the work), but different symptoms.

---

## 1. `install --platform X` never writes `.graphify_version`, so the upgrade warning is permanent

After `pipx upgrade graphifyy` (0.8.36 -> 0.9.50), every invocation printed:

```
warning: skill is from graphify 0.8.36, package is 0.9.50. Run 'graphify install' to update.
```

I ran `graphify install`. Exit 0, output included `skill installed -> C:\Users\...\.claude\skills\graphify\SKILL.md`. **The warning persisted.**

**Cause.** `_check_skill_version` iterates every platform destination:

```python
for skill_dst in {_platform_skill_destination(name) for name in _PLATFORM_CONFIG}:
    _check_skill_version(skill_dst)
```

A plain `graphify install` only refreshes the *detected* platform. A second copy at `~/.agents/skills/graphify/` (Antigravity) still had `.graphify_version` = `0.8.36`, and that stale marker drove the warning even though the Claude skill was current.

**Then the targeted command did not fix it either.** `graphify install --platform antigravity` reported:

```
Done. Open your AI coding assistant and type:
  /graphify .
```

...and `~/.agents/skills/graphify/.graphify_version` was **still `0.8.36`**. Only writing the marker by hand cleared the warning. So `--platform X` appears to write `SKILL.md` and `references/` but not the version marker.

**Net effect:** a user who follows the warning's own advice sees the warning again, forever, with no indication of which destination is stale.

**Suggestions**
- Write `.graphify_version` in the `--platform X` path.
- Have the warning name the offending destination — `skill at ~/.agents/... is from 0.8.36` — since the current message is unactionable when several platform dirs exist.

---

## 2. `install` silently overwrites a user-modified `SKILL.md` with no backup

We had deliberately edited the `description:` field in `~/.claude/skills/graphify/SKILL.md` to reflect measurements from our own corpus, so the agent would route questions appropriately for our setup.

`graphify install` replaced the file wholesale. No diff, no prompt, no `.bak`. The output line:

```
skill installed  ->  C:\Users\...\.claude\skills\graphify\SKILL.md
```

does not read as *"your local edits to this file are gone"* — it reads like a successful no-op if you already had the skill.

**Suggestions**
- If the on-disk `SKILL.md` differs from the bundled one, write `SKILL.md.bak` before overwriting and say so in the output.
- Or detect divergence and require `--force` to clobber.

Minor related note, same command: `install` prints `CLAUDE.md -> already registered (no change)`. For a user auditing what a tool has written into their agent configuration, that line does not say *what* was registered or *where*.

## Environment

- graphify 0.9.50 (pipx), Python 3.14, Windows 11
- Platform dirs present: `~/.claude/skills/graphify/` and `~/.agents/skills/graphify/`

abhay-codes07:
Closes #3144.

## Status of the two reported defects on 0.9.52

**Part 1 (the `--platform X` stamp)** is already half-fixed upstream: I audited all 18 platform destinations in a sandboxed HOME on 0.9.52 and every `install --platform X` now writes `.graphify_version` beside its SKILL.md. What remains — and what actually made the warning *permanent* — is the other half of the report: the version check iterates **every** platform destination, a plain `graphify install` refreshes only the *detected* one, and the warning named neither the stale path nor a command that would fix it. So the user follows the advice, the warning stays, and only hand-editing the marker clears it.

**Part 2 (silent overwrite)** reproduces exactly as filed: appending a local edit to `~/.claude/skills/graphify/SKILL.md` and re-running `install --platform claude` replaced it wholesale — no backup, no message beyond `skill installed  ->  …`, which reads like a no-op.

## The change

1. The stale-skill warning now names the destination and the exact command:

   ```
   warning: skill at C:\Users\me\.agents\skills\graphify is from graphify 0.8.36, package is 0.9.52.
   Run 'graphify install --platform antigravity' to update it (a plain 'graphify install'
   refreshes only the detected platform).
   ```

   The check loop passes each destination's platform name(s) through, so shared directories still warn once.

2. When the on-disk SKILL.md differs from what is about to be written, the previous copy is kept as `SKILL.md.bak` and announced (`previous copy -> … (differed from the packaged skill)`). Every upgrade differs too, so the `.bak` is simply overwritten each install and always holds exactly the previous copy; an unmodified reinstall writes no backup and stays quiet. A failed backup never blocks the install.

## Tests

`tests/test_install_version_warning.py` — 6 tests: the warning names the stale path, the `--platform` command and the "only the detected platform" caveat; the generic command is kept when no platform name is known; a current skill stays silent; a locally-edited skill is backed up with the edit intact, announced, and replaced; an unmodified reinstall writes no backup; and the version stamp lands beside the skill. With the fix reverted, 3 of 6 fail. The install/uninstall suites are otherwise unchanged (the 6 pre-existing Windows-only failures excepted, same as baseline); the full suite matches the fresh `v8` (0.9.52) baseline.

### DIFF
--- a/graphify/__main__.py
+++ b/graphify/__main__.py
@@ -160,8 +160,26 @@ def __getattr__(name: str) -> str:
 
 
 
-def _check_skill_version(skill_dst: Path) -> None:
-    """Warn if the installed skill is from an older graphify version."""
+def _check_skill_version(skill_dst: Path, platform_names: "list[str] | None" = None) -> None:
+    """Warn if the installed skill is from an older graphify version.
+
+    ``platform_names`` are the platforms installing into this destination
+    (resolved here when not given - the call site passes one positional
+    argument only, because tests stub this function with one-arg lambdas), so
+    the warning can name the exact command that refreshes THIS copy (#3144):
+    a plain `graphify install` only refreshes the detected platform, and a
+    stale marker at another platform's destination made the warning permanent
+    - the user followed the advice, the warning stayed, and only editing the
+    marker by hand cleared it.
+    """
+    if platform_names is None:
+        try:
+            platform_names = [
+                name for name in _PLATFORM_CONFIG
+                if _platform_skill_destination(name) == skill_dst
+            ]
+        except Exception:
+            platform_names = []
     version_file = skill_dst.parent / ".graphify_version"
     try:
         if not version_file.exists():
@@ -203,7 +221,16 @@ def _check_skill_version(skill_dst: Path) -> None:
                 file=sys.stderr,
             )
         else:
-            print(f"  warning: skill is from graphify {installed}, package is {__version__}. Run 'graphify install' to update.", file=sys.stderr)
+            _cmd = (
+                f"graphify install --platform {platform_names[0]}"
+                if platform_names else "graphify install"
+            )
+            print(
+                f"  warning: skill at {skill_dst.parent} is from graphify {installed}, "
+                f"package is {__version__}. Run '{_cmd}' to update it "
+                f"(a plain 'graphify install' refreshes only the detected platform).",
+                file=sys.stderr,
+            )
 
 
 def _version_tuple(version: str) -> tuple[int, ...]:
--- a/graphify/install.py
+++ b/graphify/install.py
@@ -211,6 +211,19 @@ def _copy_skill_file(platform_name: str, *, project: bool = False, project_dir: 
         if orphan_refs.exists():
             shutil.rmtree(orphan_refs)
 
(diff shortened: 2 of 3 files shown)


safishamsi:

PR #3208, opened 2026-08-30
fix: protect against unverified semantic node loss

### LINKED ISSUE #3203: A successful semantic re-extract can silently drop a file's prior doc nodes — no guard fires on the default path
**Version:** graphifyy 0.9.52 (PyPI), source verified against tag `v0.9.52` · **Type:** bug / silent data loss

## Summary

When an incremental `graphify extract` re-extracts a changed doc and the semantic chunk **succeeds** — valid JSON, nothing hollow, nothing truncated, the file covered by ≥ 1 node — but returns **fewer or renamed** doc nodes than the prior build, the file's missing nodes are dropped from `graph.json` with exit 0 and no warning beyond the informational `Replaced N node(s)` line. On the default flag set (`dedup` on, no `--force`, no `--allow-partial`), **none of the existing guards is consulted for this write** — including the #479 count comparison, which in the repro below would have refused the write had it been asked.

This is the quiet middle of the spectrum #3105 closed the loud end of: that fix arms the shrink guard for hollow/unparseable/omitted chunks, but a chunk that under-produces *plausibly* is a successful chunk by every check the pipeline applies.

## Repro (deterministic, no API key)

Self-contained script — writes a 5-file corpus (3 small code files, 2 markdown docs), a deterministic OpenAI-compatible stub backend on `127.0.0.1:8973`, installs `graphifyy[openai]==0.9.52` into an isolated venv, and runs `graphify extract` twice with default flags:

https://gist.github.com/AirRocker/f537eb20515db13a376c3b19acb0fef5

```
bash graphify_silent_doc_loss_repro.sh   # needs: bash, python3, uv
```

Steps it performs:

1. **Run 1** — full build. The stub answers the semantic pass with 6 nodes for `ARCHITECTURE.md`, 5 for `OPERATIONS.md`, plus 3 cross-file edges from `OPERATIONS.md`'s concepts into `ARCHITECTURE.md`'s. Result: 22 nodes, 29 edges, 11 semantic doc nodes.
2. **Trivial edit** — append a `_Last reviewed: 2026-08-29._` line to `ARCHITECTURE.md`.
3. **Run 2** — same command. Incremental scan re-dispatches only `ARCHITECTURE.md`. The stub now models the ordinary nondeterministic under-extraction: a **valid, well-formed** response with 2 nodes for the file — the surviving concept renamed (`architecture_exponential_backoff` → `architecture_backoff`). The chunk reports `chunk 1/1 done`; the file is covered; nothing is hollow, truncated, or omitted.

Observed (run 2):

```
run 2 exit code: 0

stderr, in full:
[graphify] Replaced 6 node(s) from re-extracted source file(s).

total nodes:        22 -> 18
total edges:        29 -> 21
semantic doc nodes: 11 -> 7

semantic doc nodes LOST (5):
  - architecture_exponential_backoff   (ARCHITECTURE.md)
  - architecture_json_result_document  (ARCHITECTURE.md)
  - architecture_operator_escalation   (ARCHITECTURE.md)
  - architecture_run_context           (ARCHITECTURE.md)
  - architecture_startup_configuration (ARCHITECTURE.md)

edges that vanished with them: 9 — including all 3 cross-file edges
attributed to OPERATIONS.md, a file this run never touched:
  - operations_deploy_healthiness -[references]-> architecture_run_context
  - operations_error_stream       -[references]-> architecture_operator_escalation
  - operations_last_good_run      -[references]-> architecture_exponential_backoff
```

The stub is a stand-in for behavior the tracker has already documented in real models: #3105's originating run went 570 → 111 nodes on well-formed responses, #3090 measured 28 per-file collapse episodes across 11 re-indexes, and I have hit 75 → 49 doc nodes on a one-line edit of a project `CLAUDE.md` (the run behind the PR linked at the end).

## Why each existing guard misses it (v0.9.52 source)

**`build_merge`'s identity guard (#479, reworked in #2497)** is gated `if had_graph and not dedup:` (`graphify/build.py:1960`) and `dedup` defaults to `True` (`build.py:1632`), so on the default path it never runs. Even with `--no-dedup`, `_explained` excuses every lost node whose source was re-extracted this run in the same tier (`build.py:1983–1990`) — the comment above it records this as an accepted residual tradeoff, deferring the under-production case to the extraction layer's incomplete-build guard. Which brings us to:

**The #3105 completeness check** (`graphify/cli.py:3813–3842`) arms `_extraction_incomplete` only for crashed chunks, partially-failed chunk counts, `uncovered_files` (a dispatched file with **zero** returned nodes), and truncated/hollow chunks. A successful chunk returning 2 nodes where the prior build had 6 trips none of those — this run *is* complete by its definition.

**The `to_json` count backstop (#479, `graphify/export.py:308–321`)** is then bypassed entirely: a complete extraction sets `_force_write = True` (`cli.py:4240`) and `to_json(..., force=True)` skips the comparison. The bypass matters concretely here — the repro's write is a strict net shrink (22 → 18) that the count guard **would have refused** had it been consulted. And for the variant where the re-extract returns the same count under different IDs, the count guard couldn't see it even when consulted — the unit critiques in #3090 and #2229 apply.

**`watch._check_shrink`** — the source-aware guard on the watch/update path — first gates on net count (`graphify/watch.py:1125`), so a rename-neutral loss never reaches its identity diff, and when it does run, losses from `rebuilt_sources` are "accounted" unless the file's extraction *failed* (`watch.py:1132–1140`). A successful re-extract is in `rebuilt_sources` and not in `failed_sources` by definition.

The compounding factor: run 2 stamps the manifest, so nothing ever re-dispatches the file. The loss is permanent until a human notices the graph can no longer answer something it used to — which, for cross-file edges from files the run never touched, is exactly the kind of degradation an incremental user has no reason to go looking for.

## Relation to open neighbours

#3090 and #2229 critique the count guard's *unit* (per-file vs. total; per-layer vs. total) in scenarios where the guard runs. This report is about the *gating*: on the default incremental path the guard is not consulted at all, and the guards that do run check completeness, not fidelity. #1652 asks for adjacent tooling (dry-run, degree-drop alerts). Fixing any one of those would not close this path; closing this path would make those checks reachable.

## Closing

PR #2162 implements one shape of guard for exactly this class — a skill-level mandatory-ID gate for re-extracted docs on `--update` (rebased; content unchanged since 2026-08-10). But the mechanism matters less than the property: on the default path, a lost doc node should be either explained or refused. An identity-diff before the write, an opt-in strictness flag, or a config default would each close it, and I'm glad to rework the PR toward whichever shape fits — same offer as before.

hopstreax:
## Summary

Fixes #3203 by preventing an unexpectedly under-produced semantic document extraction from silently replacing a previously valid graph representation.

Previously, a semantic re-extraction could return fewer nodes than the previous extraction while still reporting a successful `stop` completion with valid JSON. Because the source had at least one resulting node, the existing incomplete-extraction checks considered the extraction complete, allowing the smaller representation to replace the existing graph.

## What changed

* Track semantic-tier node counts per source during `build_merge()` / `merge_raw_extraction()`.
* Compare the fresh extraction against the on-disk semantic baseline.
* Mark a source as **unverified** when:

  * the previous extraction contained more than one semantic node, and
  * the fresh extraction produces fewer semantic nodes.
* Propagate the anomaly to the CLI instead of making `build_merge()` fail.
* Arm the existing incomplete-extraction/shrink protection so the last known-good `graph.json` is preserved.
* Prevent affected sources from being stamped as complete in `manifest.json`, allowing them to be retried on a subsequent incremental run.
* Preserve `--allow-partial` as an explicit override for intentional reductions.

## Why this approach

A generic node-count shrink guard would recreate the false-positive behavior addressed by #1116, since legitimate edits can remove semantic nodes.

Exact node-ID matching was also avoided because semantic document IDs are LLM-generated and can legitimately change between extractions.

The new check therefore operates at the semantic source level and treats a reduction as **unverified rather than automatically invalid**. This allows the existing recovery path to protect the previous graph without making `build_merge()` itself responsible for deciding whether a document reduction was intentional.

AST/code-tier reductions remain unaffected.

## Testing

Added a dedicated regression suite covering:

* semantic node shrink detection
* renamed surviving semantic nodes
* non-fatal `build_merge()` behavior
* AST/code shrink exclusions
* equal-count semantic replacements
* semantic growth
* multi-chunk source aggregation
* source-specific shrink detection
* raw extraction parity
* end-to-end #3203 reproduction
* `--allow-partial` override

Relevant regression suites:

**88 tests passed.**

### DIFF
--- a/graphify/build.py
+++ b/graphify/build.py
@@ -1615,11 +1615,37 @@ def merge_raw_extraction(
             return False  # unowned — carry forward
         return _prune_hit(sf)
 
+    # #3203: Check for unverified semantic shrink on re-extracted sources.
+    unverified_semantic_shrink: dict[str, tuple[int, int]] = {}
+    if new_sem_sources:
+        prior_sem_counts: dict[str, int] = {}
+        for n in existing_nodes:
+            if isinstance(n, dict) and not _is_ast_tier(n):
+                sf = n.get("source_file")
+                if sf:
+                    canon = _norm_source_file(sf, _eff_root) or sf
+                    prior_sem_counts[canon] = prior_sem_counts.get(canon, 0) + 1
+
+        fresh_sem_counts: dict[str, int] = {}
+        for n in new.get("nodes", []):
+            if isinstance(n, dict) and not _is_ast_tier(n):
+                sf = n.get("source_file")
+                if sf:
+                    canon = _norm_source_file(sf, _eff_root) or sf
+                    fresh_sem_counts[canon] = fresh_sem_counts.get(canon, 0) + 1
+
+        for canon_sf, fresh_count in fresh_sem_counts.items():
+            prior_count = prior_sem_counts.get(canon_sf, 0)
+            if prior_count > 1 and fresh_count < prior_count:
+                unverified_semantic_shrink[canon_sf] = (prior_count, fresh_count)
+
     new["nodes"] = [n for n in existing_nodes if not _dropped(n)] + list(new.get("nodes", []))
     new["edges"] = [e for e in existing_edges if not _dropped(e)] + list(new.get("edges", []))
     carried_hyper = [he for he in existing_hyperedges if not _dropped(he)]
     if carried_hyper or new.get("hyperedges"):
         new["hyperedges"] = carried_hyper + list(new.get("hyperedges", []))
+    if unverified_semantic_shrink:
+        new["_unverified_semantic_shrink"] = unverified_semantic_shrink
     return new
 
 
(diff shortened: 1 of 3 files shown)


safishamsi:
Shipped in v0.9.53 via authorship-preserving cherry-pick so you keep contributor-graph credit. Thanks @hopstreax! Release: https://github.com/Graphify-Labs/graphify/releases/tag/v0.9.53

PR #3211, opened 2026-08-30
feat: Robot Framework extractor for .robot/.resource files (#3192)

### LINKED ISSUE #3192: Feature Request: Add Robot Framework .robot and .resource parsing support
## Summary

Please add native Robot Framework support to Graphify for `.robot` and
`.resource` files.

Graphify currently maps the Python libraries used by Robot Framework projects,
but it does not create structural nodes and relationships for Robot suites,
test cases, user keywords, resource imports or keyword calls. This leaves the
resulting dependency graph incomplete for Robot-based automation repositories.

## Proposed implementation

Robot Framework provides an official parsing API:

```python
from robot.api import get_model

nshiveg:
## feat: Robot Framework extractor for `.robot`/`.resource` files

Closes #3192

### Problem

Robot Framework files are invisible to graphify: `.robot`/`.resource` are in neither `detect.CODE_EXTENSIONS` nor `_DISPATCH`, so on Robot-based automation repos the graph maps the Python keyword libraries but none of the suites, test cases, user keywords, or the resource-import topology that actually structures such a repo (details in #3192).

### Approach

No maintained tree-sitter grammar exists for Robot Framework, so `graphify/extractors/robot.py` uses the official parser (`robot.api.get_model` / `get_resource_model` + `ModelVisitor`), as proposed in the issue. Pure Python, wheels everywhere, added as an optional `robot` extra following the `terraform`/`ocaml` pattern, and wired into `_EXTRA_FOR_EXTENSION` so the #1745 warning names the extra when the parser is missing.

**Nodes**: the suite/resource file, its test cases, its user keywords, plus stub nodes for third-party named libraries (SeleniumLibrary, RequestsLibrary, ...) so suites sharing a library cluster together (same pattern as the Svelte bare-import stubs). RF standard libraries (BuiltIn, Collections, ...) are filtered like the built-ins in `_LANGUAGE_BUILTIN_GLOBALS` (#726) — they would otherwise become instant god nodes.

**Edges**: `contains` (file → test/keyword), `imports` (Resource / path-form Library / Variables imports resolved onto the imported file's own node id, preserving the relative/absolute path form like the JS relative-import resolver; `${CURDIR}`/`${/}`/`${EXECDIR}` are substituted, and an import whose path still holds an unresolvable `${VARIABLE}` emits nothing), and `calls` (test/keyword/suite-fixture → keyword, including `[Setup]`/`[Teardown]`/`[Template]`/`Test Template` and calls nested in FOR/IF blocks).

**Design note — keyword ids are bare names, not stem-qualified**: Robot resolves keywords globally by name across imported resources, so bare ids make a suite's call edge land on the defining resource's keyword node with no separate cross-file resolution pass. Test-case nodes stay stem-qualified because test names repeat across suites. Calls to keywords never defined in the corpus (e.g. BuiltIn's `Log`) dangle and are dropped by `build_from_json` like any external reference.

One parser subtlety encoded in the tests: `Template`/`Test Template` statements carry their keyword in `.value` (not `.name` like Setup/Teardown) — missing this silently drops all template edges.

### Tests

5 tests in `tests/test_languages.py` (skipif-guarded like the `dm`/`commonlisp` extras) + 2 fixtures (`sample.robot`, `robot_keywords.resource`) covering: node extraction, call edges incl. fixtures/loops/templates, import resolution + stdlib filtering + unresolvable-variable paths, `.resource` parsing, and the cross-file id guarantee (suite call-edge target == resource definition id). `ruff` and `python -m tools.skillgen --check` pass.

### Real-world validation

Run against an industrial Robot Framework automation repo (~850 code files, ~5,000 tests/keywords): the graph grew from 5,104 to 15,768 nodes — 10,577 robot-sourced — with 44,423 keyword-call edges and 440 robot→python `Library` edges resolving onto the existing Python nodes. God nodes shifted from generic Python helpers to the actual shared simulator-control keywords, which matches how that codebase is really wired.


nshiveg:
Thanks for the automated review — findings addressed in 58e19cc:

- **`${CURDIR}` double-prefix** — confirmed and fixed. `${CURDIR}`/`${EXECDIR}` now mark the path as *anchored* so it is no longer re-joined with the source dir (which doubled the prefix for relative scan paths), and resolved paths are normpathed unconditionally so `../` segments collapse for absolute sources too. Regression test: `test_robot_curdir_and_execdir_imports_resolve_without_double_prefix`.
- **Keyword-call normalization** — fixed. Keyword node ids now normalize exactly the way Robot Framework matches keyword names (case-, space-, and underscore-insensitive) via `_kw_id`, so `OpenSession` / `open_session` / `Open Session` share one node. Test: `test_robot_keyword_matching_is_case_space_underscore_insensitive`.
- **"Optional robot extra is imported unconditionally"** — not the case: `graphify/extractors/robot.py` imports `robot.api` lazily *inside* `extract_robot()`, so importing `graphify.extract` works without robotframework and the extractor returns the install-hint error dict (same pattern as the terraform extractor's `tree_sitter_hcl` handling). Now codified as a test that blocks the `robot` package at import level in a subprocess: `test_robot_extractor_degrades_gracefully_without_robotframework`.
- **Bare-name keyword id merging across files** — deliberate. Robot resolves keywords in a single global namespace, so two files defining the same keyword name are one name to Robot (and ambiguous to call); merging them is what makes cross-file call edges land on definitions with no separate resolution pass. Now called out explicitly in the extractor docstring.

All 8 robot tests pass; ruff and `tools.skillgen --check` clean.

nshiveg:
Both round-2 findings addressed in 8181676:

- **`${EXECDIR}` in absolute-path scans** — confirmed and fixed. `${EXECDIR}` (the directory Robot was launched from) is statically knowable only for a relative scan, where `.` aligns with the scan root and matches how relative file-node ids are built. For an absolute source path the resolved id could never match a node, so the import now emits nothing instead of a guaranteed-dangling relative edge. Test extended: `test_robot_curdir_and_execdir_imports_resolve_without_double_prefix`.
- **BDD-style calls** — confirmed and fixed. `Given/When/Then/And/But <keyword>` calls now emit call edges for **both** the full and the one-prefix-stripped candidate, mirroring Robot's try-full-then-stripped resolution; whichever keyword exists receives the edge and the other candidate dangles and is dropped by the graph builder (no fabricated nodes). English prefixes only — localized BDD prefixes are out of scope. New test: `test_robot_bdd_style_calls_reach_their_definitions`, including a keyword literally named with a prefix.

9/9 robot tests pass; ruff clean.

nshiveg:
Round-3 findings, with evidence:

- **Built-in Robot path variables matched case-sensitively** - fair nit, fixed in 25dca92: variable names in Settings-section imports now match the way Robot matches them (case-, space-, and underscore-insensitive), so `${curdir}` / `${Cur_Dir}` / `${EXEC DIR}` resolve like `${CURDIR}`; any other variable, in any casing, still yields no edge. Test: `test_robot_path_variables_match_case_space_underscore_insensitively`.
- **"CURDIR replacement may produce non-anchored absolute"** - no reproducible failure mode found: `${CURDIR}` with a relative source resolves without double-prefixing, and with an absolute source comes back normpathed (no leftover `..`), both asserted by `test_robot_curdir_and_execdir_imports_resolve_without_double_prefix`. Happy to fix a concrete counter-example if one exists.
- **"Robot keyword IDs are not case-normalized"** - already the case since 58e19cc: `_kw_id` lowercases (via `make_id`) after stripping spaces/underscores, and `test_robot_keyword_matching_is_case_space_underscore_insensitive` asserts `OPEN SESSION` / `open_session` / `OpenSession` share one node. This re-raises a finding fixed in the previous commit - possibly analyzed against a stale hunk.

10/10 robot tests pass; ruff clean.

safishamsi:
Shipped in v0.9.53 via authorship-preserving cherry-pick so you keep contributor-graph credit. Thanks @nshiveg! Release: https://github.com/Graphify-Labs/graphify/releases/tag/v0.9.53

PR #3213, opened 2026-08-30
fix(csharp): keep member nodes out of the type-definition index

### LINKED ISSUE #3212: C#: property and enum member nodes enter the type-definition index, so a member can capture every reference to a same-named type
### Environment

- graphify 0.9.52 (`v8` @ 680e3ed), Linux, Python 3.12
- `graphify.extract.extract([...])`

### Summary

`_build_csharp_type_def_index` (`graphify/extractors/csharp.py:19`) treats every sourced `.cs` code node with a plain identifier label as a type declaration. Property nodes (#3006) and enum member nodes (#3063) are exactly that shape, so they enter the index and compete with real declarations for a `(namespace, name)` key.

The index does not require the key to be unique. It resolves a collision by sorting on `(source_file, source_location, id)` and taking the first entry, so the winner is decided by file name order. A member in an earlier-sorting file captures every reference to the type it shares a name with, across the whole corpus.

I filed both of the PRs that added those member nodes, so this is my regression.

### Reproducer

```csharp
// holder.cs
namespace App;
public class Holder { public Widget Widget { get; set; } }

// widget.cs
namespace App;
public class Widget {}

// other.cs
namespace App;
public class Other { public Widget Thing { get; set; } }
```

`public Widget Widget { get; set; }` is ordinary C#.

### Current behavior (0.9.52)

```
Holder --references--> Widget @ holder.cs   (the property)
Other  --references--> Widget @ holder.cs   (the property)
```

Both classes bind to the property, including `Other`, which has no colliding member of its own. Rename `widget.cs` to `aaa.cs` so it sorts first and both bind to the class instead. Nothing else changes.

The same shape with an enum:

```csharp
// abc.cs
namespace App;
public enum ProductType { Standard, Bundle }

// zeta.cs
namespace App;
public class Standard {}
```

A reference to the type `Standard` resolves to `ProductType.Standard`.

And it can invent a binding that C# scoping forbids. With the class in `NsA`, the property in `NsB`, and no `using`, the reference should dangle:

```
Holder --references--> Widget @ b_holder.cs   (the property, cross-namespace)
```

A sibling class in `NsB` without the colliding property correctly dangles on a sourceless stub, so the member is what makes the difference.

### Measured

On one .NET service (1,497 `.cs` files):

| | 0.9.52 |
|---|---|
| index entries | 4,197 |
| entries whose winner is a member | 2,777 (66%) |
| of those, ones that shadow a real declaration | 18 |
| edges arriving at member nodes | 75 `references`, 43 `calls` |

The 18 are the `public ReviewInfo ReviewInfo { get; set; }` idiom, where the property and the class it names sit in the same namespace.

The damage runs both ways: the member collects edges that are not about it, and the type loses them, so "what uses this type" under-reports.

### Note on scope

`_build_csharp_type_def_index` is the only index of this shape; `grep` for `type_def_index` finds no equivalent in another language's extractor, and `type_def_nids` in `extract.py` is already guarded by `contained`, which member nodes are not in. TypeScript enum members (#3064) are unaffected because this index is `.cs` only.

The extraction cache directory is versioned (`v{version}-s{schema}`), so a release invalidates the stale unstamped nodes on its own.

durmazoguzhan:
Closes #3212.

`_build_csharp_type_def_index` treats every sourced `.cs` code node with a plain identifier label as a type declaration. The property nodes from #3006 and the enum member nodes from #3063 are exactly that shape, so they entered the index and competed with real declarations for a `(namespace, name)` key.

The index does not require the key to be unique. It sorts colliding entries on `(source_file, source_location, id)` and takes the first, so file name order picked the winner, and a member captured every reference to the type it shares a name with, across the whole corpus. `public Widget Widget { get; set; }` is enough to trigger it.

Both of those PRs are mine, so this is my regression.

### The change

A member node is stamped `is_member` at emission, and the index skips it. That is the same shape as the `is_nested_type` skip two lines above, and it keeps the decision at the one place that knows what the node is.

Nothing is removed: the member node, its `defines` / `case_of` edge, and its `namespace` metadata are untouched. `add_node` merges the stamp with what it fills in itself, and there is a test pinning that the namespace survives, because every C# resolution pass keys on it.

### Measured

One .NET service, 1,497 `.cs` files:

| | before | after |
|---|---|---|
| index entries | 4,197 | 1,438 |
| entries whose winner is a member | 2,777 | **0** |
| of those, shadowing a real declaration | 18 | 0 |
| `references` edges landing on a member | 75 | **0** |
| `calls` edges landing on a member | 43 | 9 |

The 18 are the `public ReviewInfo ReviewInfo { get; set; }` idiom, where the property and the class it names share a namespace.

### The 9 that remain

Nine `new Foo()` edges still reach a property node. They do not come through this index, which now holds no members at all, and I could not reduce them to a minimal case: the same three-file shape built by hand produces no call edge on either side of the change. Rather than widen this PR on a path I cannot demonstrate, I am reporting them as a remainder. They were 43 before this change.

### Tests

Nine tests in `tests/test_csharp_type_resolution.py`, where the C# resolution tests and their helpers already live. Seven fail without the change; the other two are the regression guard and the control, which are meant to pass either way.

Two of them needed a second pass. Written the obvious way, they asserted that a resolved target does not carry the `is_member` stamp, which is this change's own marker, so on the unfixed code, where nothing carries it, they passed and proved nothing. They now use oracles that exist on both sides: the target is the node its class points at with `defines`, and a cross-namespace reference must land on a sourceless stub.

The cases: the index unit itself (both a collision and a member-only key), a property shadowing a type in another file, a property shadowing a type in the same file, a cross-namespace reference that must dangle instead of binding to a member, an enum member shadowing a type, an enum nested in a class whose members are not nested types and need their own stamp, that member nodes and their edges survive, that the stamp does not displace the namespace metadata, and a control where nothing shadows.

Full suite: 5,130 passed. `ruff` clean.

### DIFF
--- a/graphify/extractors/csharp.py
+++ b/graphify/extractors/csharp.py
@@ -27,6 +27,15 @@ def _build_csharp_type_def_index(all_nodes: list[dict]) -> dict[tuple[str, str],
             metadata = {}
         if metadata.get("is_nested_type"):
             continue
+        # A member is not a type. Property nodes (#3006) and enum member nodes
+        # (#3063) are sourced `.cs` code nodes with a plain identifier label, so
+        # they entered this index and competed with real declarations for a
+        # `(namespace, name)` key. The winner is the first entry in a
+        # (source_file, source_location, id) sort, which made `public Widget
+        # Widget { get; set; }` in an earlier-sorting file capture every
+        # reference to the type `Widget` across the corpus.
+        if metadata.get("is_member"):
+            continue
         nid = node.get("id")
         label = node.get("label")
         if not (isinstance(nid, str) and nid and isinstance(label, str) and label):
--- a/graphify/extractors/engine.py
+++ b/graphify/extractors/engine.py
@@ -2476,7 +2476,10 @@ def _csharp_extra_walk(node, source: bytes, file_nid: str, stem: str, str_path: 
         # second would hang a duplicate edge on the first member's node, so the
         # first declaration keeps it (same guard as the property nodes in #3006).
         if member_nid not in seen_ids:
-            add_node_fn(member_nid, member_name, line)
+            # `is_member` keeps this out of the C# type-definition index, the
+            # same way a property node is kept out: an enum member is not a type.
+            add_node_fn(member_nid, member_name, line,
+                        metadata={"is_member": True})
             add_edge_fn(parent_class_nid, member_nid, "case_of", line)
         return True
     if node.type == "namespace_declaration":
@@ -3856,7 +3859,12 @@ def _extract_generic(
                     property_line = node.start_point[0] + 1
                     property_nid = _make_id(parent_class_nid, property_name)
                     if property_nid not in seen_ids:
-                        add_node(property_nid, property_name, property_line)
+                        # `is_member` keeps this out of the C# type-definition
+                        # index: a property is not a type, and without the stamp
+                        # it competed with real declarations for its
+                        # `(namespace, name)` key.
+                        add_node(property_nid, property_name, property_line,
+                                 metadata={"is_member": True})
                         add_edge(parent_class_nid, property_nid, "defines",
                                  property_line, context="field")
             type_node = node.child_by_field_name("type")
--- a/tests/test_csharp_type_resolution.py
+++ b/tests/test_csharp_type_resolution.py
@@ -567,3 +567,194 @@ def test_csharp_alias_using_scoped_to_its_block(tmp_path: Path):
     inh = {(e["source"], e["target"]) for e in result["edges"] if e.get("relation") == "inherits"}
     assert (good["id"], n_t["id"]) in inh, "Good must bind N.T via the in-block alias"
     assert (bad["id"], n_t["id"]) not in inh, "Bad (sibling block) must NOT see the alias"
+
+
+# ── member nodes must not enter the type-definition index (#3212) ─────────────
+#
+# `_build_csharp_type_def_index` treats every sourced `.cs` code node with a plain
+# identifier label as a type declaration. Property nodes (#3006) and enum member
+# nodes (#3063) are exactly that shape, so they entered the index and competed
+# with real declarations for a `(namespace, name)` key. The index resolves a
+# collision by sorting on (source_file, source_location, id) and taking the
+# first, which means a member in an earlier-sorting file silently captured every
+# reference to the type it shares a name with.
+
+
+def _members(result: dict, label: str) -> list[dict]:
+    return [
+        n for n in result["nodes"]
+        if n.get("label") == label and (n.get("metadata") or {}).get("is_member")
+    ]
+
+
+def test_csharp_type_def_index_skips_member_nodes():
+    # Pin the mechanism directly: given a member and a real declaration under the
+    # same key, the index must hold the declaration whatever the sort would say.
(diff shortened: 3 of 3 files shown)


JensD-git:
Independent confirmation from a large C# estate, plus one consequence of this bug that is not in
#3212 — and it is the one that hurt us.

**The `calls` edge does not land on the wrong node. It disappears.**

When the receiver of a member call is a **property**, the call gets no outgoing edge at all. A
**field** or a local variable in the same class still resolves. Two files, no corpus, no build
(`dotnet build` on net8.0: 0 errors, 0 warnings):

```csharp
// Service.cs
namespace Demo { public class Service { public void Do() { } } }

// Host.cs
namespace Demo
{
    public class Host
    {
        private Service Dependency { get; set; }   // property
        private Service _field;                    // field (control)

        public void A() { this.Dependency.Do(); }  // no edge
        public void B() { Dependency.Do(); }       // no edge
        public void C() { this._field.Do(); }      // edge
        public void D() { _field.Do(); }           // edge
        public void E() { Service s = new Service(); s.Do(); }  // edge
    }
}
```

`graphify extract src --code-only --no-cluster --out out`

| Version | edges of the 5 |
|---|---|
| v0.9.47 | 5 |
| v0.9.48 | 5 |
| **v0.9.49** | **3** |
| v0.9.50, v0.9.52, v0.9.55 | 3 |
| **v0.9.56** (released yesterday) | **3** |
| v0.9.55 + this PR | **5** |

So it starts at v0.9.49, the release that added the per-property member node, and this PR fixes it
completely. In the unpatched graph `.A()` and `.B()` have **no outgoing edge whatsoever** — nothing
is redirected to the property node, so the "edges arriving at a member" counter in #3212 does not
see this case at all. The property's declared type is still recorded into the method-scoped
receiver table on both sides of the change; that code is untouched. The only structural difference
in the reproducer is that the node for the property exists from v0.9.49 on.

**Measured impact.** ~15,300 `.cs` files across eight graphs (mixed .NET Framework and net8), both
versions built fresh from the same tree on the same day, compared over `(source_file, label)` after
excluding renamings and redirections:

* **383 `calls` edges lost** — present in the v0.9.39 extraction, absent in v0.9.55.
* **241 of them in a single graph** of the eight.
* Every one was read off the syntax tree at the call site, not sampled. The receivers are ordinary
  service properties reached through `this`; the six most frequent account for 85 of the 383
  (24, 21, 15, 9, 8 and 8 occurrences).

A property is the idiomatic way to hold an injected dependency, so the loss falls precisely on the
kind of edge a graph gets queried for. Post-processing cannot compensate: the edges are already
gone from the raw extraction.

**With the patch, a full portfolio rebuild is clean.** We also carry #2676; the two cherry-pick
onto `v8` @ `c9f9901` together without a conflict (auto-merge in `engine.py`). Full suite on
Windows: 17 failed / 5397 passed, and those 17 are the same by name on unpatched `v8` — all
platform-bound (`os.mkfifo` / `os.symlink` unavailable, `WinError 32` on `rmdir`, POSIX path and
install-destination assertions). Nothing in extraction. The `+14` passing over unpatched is exactly
the 9 tests here plus the 5 from #2676.

We run this patch in production locally, because without it v0.9.55 and later are not usable for
us. The reproducer above is the whole of it — two files, no corpus, and the version table is
reproducible from PyPI releases alone.

Issue #3094, opened 2026-08-25
Entity dedup merges nodes across different source files, silently removing whole documents from the graph


jvzelst-tech:
### What happened

Running `graphify extract` on a corpus of 205 markdown documents dropped the graph from **243
source files to 187**. Fifty-six documents ended up with zero nodes. The run reported success:

```
[graphify] Deduplicated 1017 node(s) (709 exact, 308 fuzzy).
```

### Cause

`dedup.py` pass 1 merges any two nodes whose normalised labels are equal, **regardless of
`source_file`**. A corpus of session notes, changelogs, meeting minutes or RFCs reuses generic
headings — `Decisions`, `Open items`, `Changes`, `Next steps`. Every occurrence across every
document collapses into one node, attributed to one arbitrarily chosen file, with no alias list
recording the others. Queries scoped to any of the other documents then return nothing.

`_entropy()` does not filter these out: `"decisions"` scores ≈2.73 bits/char, above
`_ENTROPY_THRESHOLD = 2.5`. Character entropy measures string shape, not how generic a heading is
within a corpus.

### The guard already exists one level up

`deduplicate_entities` already refuses to merge across repos, with exactly the reasoning that
applies here:

```python
# Guard: cross-project dedup is not supported — nodes from different repos
# share label names by coincidence and must never be merged by string similarity.
```

Two documents in one repo share `## Decisions` by coincidence in precisely the same way.

### Suggested fix

Extend that guard one level down: block a merge when **both** nodes have a `source_location` set
(i.e. both are file-scoped structural nodes) **and** their `source_file` differs. Concept nodes
without a `source_location` are the ones that legitimately merge across files, and they keep doing so.

I simulated this on the affected graph by blocking those unions:

| | current | with the guard |
|---|---|---|
| Source files retained | 187 of 243 | **242 of 243** |
| Cross-file heading merges | 3,573 performed | 3,573 blocked |
| Legitimate merges still performed | — | 213 (212 same-file, 1 cross-file concept) |

So the useful dedup survives essentially intact.

### Also worth considering

There is no shrink guard on this path. `graphify update` has `--force` precisely because it refuses
to write a smaller graph; `extract` writes one without comment. A warning when source-file coverage
drops would have surfaced this immediately.

### Environment

graphifyy 0.7.15 · Python 3.13 · macOS · corpus of 205 markdown documents.

### Related

Found in the same run as #3093 (a failed chunk still marking files complete); the two are independent defects.

[...]

PR #3216, opened 2026-08-30
fix(dedup): merge cross-file entity nodes typed by their file's extension (#296)

yotamleo:
Running graphify over a note vault leaves the same person or project sitting in the graph as many separate nodes — `@cyrilXBT`, `@cyrilxbt` and `cyrilXBT` all survive as distinct entities, one per file that mentions them. Community detection then splits what should be one cluster into fragments and god-node analysis misses the real hubs, because a single entity's mentions are scattered across aliases; on the graph measured in #296 this left 1,741 collision groups unmerged.

Normalization is not what fails. `_norm` already casefolds and collapses `[\W_]+`, so all three spellings key to `cyrilxbt` and land in the same Pass 1 bucket. The bucket is then discarded, because the cross-file union added in #2182 is gated on `file_type == "concept"` — and an entity the extractor pulls out of a `.md` note inherits `document` from the file's extension, not from anything about the entity. Note-heavy corpora do not have a canonicalization problem that code repos lack; they have the same one, but almost none of their entity nodes are typed `concept`, so the cross-file merge that already exists never gets to run on them.

## Repro

```python
from graphify.dedup import deduplicate_entities

nodes = [
    {"id": "journal_2024_03_01_cyrilxbt", "label": "@cyrilXBT",
     "file_type": "document", "source_file": "journal/2024-03-01.md"},
    {"id": "journal_2024_03_02_cyrilxbt", "label": "@cyrilxbt",
     "file_type": "document", "source_file": "journal/2024-03-02.md"},
    {"id": "journal_2024_03_03_cyrilxbt", "label": "cyrilXBT",
     "file_type": "document", "source_file": "journal/2024-03-03.md"},
]
print(len(deduplicate_entities(nodes, [], communities={})[0]))
```

Before: `3`. After: `1`. Flip `file_type` to `"concept"` on all three and it already printed `1` before this change — the type is the only thing separating them.

## Evidence

Upstream HEAD `33362d969292b57eda82f3fbd9eb5f3f5bc9bbc2` (0.9.53), `graphify/dedup.py:640` — the `n.get("file_type") == "concept"` filter on the Pass 1 cross-file `mergeable` list. The measurement behind the numbers above is in #296: https://github.com/Graphify-Labs/graphify/issues/296#issuecomment-5388969434 — classifying every surviving cross-file `_norm`-collision group by which gate rejects it gave 1,741 blocked by this filter, 2 by the entropy gate and 0 by provenance, and of 39,627 `document` nodes exactly 0 were their own file's node.

@adelaidasofia asked for option 1 of the three sketched there:

> On your three options, 1 is the one I would want: widen the gate to entity nodes that are provably not the file's own node, entropy and provenance guards untouched. It keeps merge policy conservative and needs nothing retyped upstream.

That is what this implements — nothing retyped, no migration, no new subcommand.

## The predicate

"Provably not the file's own node" needs something concrete, and `dedup.py` did not model it. It turns out not to need any new extractor coupling: `_id_prefixes` already enumerates the ID a node standing for `source_file` itself would carry, in every spelling a stored path may take (absolute, repo-relative, or the pre-#1504 bare stem). A file's own node IS one of those; an entity extracted from that file carries an `_<entity>` suffix and never equals one.

The new `_reads_as_file_entity` requires two things, and either one failing keeps the node blocked:

* **Proven** — it is not its file's own node, by the `_id_prefixes` reconstruction above. This is a reconstruction, not a heuristic, and it holds for every spelling a stored `source_file` may take.
* **Assumed** — it is not a section of the file: `node_kind: "heading"`, or `"page"` for the file node the markdown extractor labels outright.

I have kept those two labelled separately in the code rather than calling the whole thing "provable", because the second half is only as good as the `node_kind` stamp. **Absence of the marker is not proof of entity-ness**: a producer that mints a node standing for part of its file without stamping `node_kind` — `detect.py`'s per-sheet `document` nodes are the in-tree example — yields structural nodes this predicate reads as entities, and two of them sharing a label in different files would merge. I did not paper over that with a guess about what an unstamped node meant; the conservative fix is on the producer side, so the PR states a short producer contract next to `_FILE_STRUCTURE_NODE_KINDS` and pins the stamped/unstamped split in a test, `test_reads_as_file_entity_trusts_the_node_kind_stamp_only`. Happy to stamp the `detect.py` path too if you would rather close it here — I left it alone because that code sits behind a not-yet-wired feature flag.

The second condition is why the widening does not touch the guard verified in #3094. Repeated `## Decisions` / `## Next steps` sections across sibling documents are distinct sections, not duplicates; `node_kind` exists precisely because `file_type` cannot carry that distinction, so it is the field that separates a heading from an entity found in the same file. Three documents each carrying both headings still yield six heading nodes, each attributed to its own file — there is a test for exactly that shape.

A node that cannot be checked (no ID, no provenance) answers `False` and stays blocked. The predicate can therefore only ever *narrow* what the gate treats as file-anchored, never widen it on a guess.

## What is deliberately unchanged

The entropy and provenance guards are byte-untouched, and the widened rows go through them exactly as `concept` rows do. `code` (#1205) is still keyed by ID and never enters Pass 1 at all; `image`/`paper` still stay blocked, so shared basenames like `logo.png` in two directories do not merge. Pass 2's fuzzy `_crossfile_fileanchored_blocked` is untouched, so #1284's near-identical parallel-module boilerplate stays blocked — the widening is scoped to the exact-normalization pass, where the label is identical after normalization rather than merely similar.

The survivor heuristic ("prefer the node with non-generated body content as the merge target") is not in this PR. `_pick_winner` picks on ID shape and `dedup.py` never sees body content, so doing it properly is its own change; a file's own node — usually the hand-curated note — is held out of the merge entirely here rather than being made to win it.

## Effect

* The fixture above: 3 nodes to 1.
* On a 15,197-node mixed code-and-markdown graph, cross-file exact-normalization groups that merge go 8 to 32 and nodes collapsed 27 to 52, with 133 `document`/`rationale` nodes held back by the predicate as file structure.

## Tests

Added to `tests/test_dedup.py`, all in the repo's existing style:

* `test_reads_as_file_entity_helper` — the predicate proves entity-ness and answers `False` whenever it cannot: the file's own node in each stored-path spelling, `page` and `heading` nodes, and nodes missing an ID or provenance.
* `test_dedup_merges_crossfile_document_entity_variants` — the reported shape, three spellings across three notes, collapsing to one.
* `test_dedup_merges_crossfile_rationale_entity_variants` — `rationale` rides the same gate.
* `test_dedup_never_merges_a_files_own_node_away` — the boundary: a curated note whose own node carries the entity's label survives alongside a mention of it elsewhere.
* `test_dedup_never_merges_two_files_own_nodes` — two `README.md` stay two documents.
* `test_dedup_never_merges_repeated_headings_across_files` — the #3094 verification case.
* `test_dedup_crossfile_entity_merge_keeps_the_entropy_gate` / `..._provenance_gate` — negative controls proving those guards still fire on the widened rows.
* `test_dedup_crossfile_fuzzy_fileanchored_block_is_untouched` — Pass 2 still blocks near-identical cross-file document labels.
* `test_reads_as_file_entity_trusts_the_node_kind_stamp_only` — pins the documented limit above: a stamped structural node stays blocked, an unstamped one reads as an entity.

Two existing rows in `test_crossfile_identical_labels_stay_distinct_for_guarded_types` changed meaning and are updated rather than deleted: the `document` and `rationale` rows asserted that an identical label never merges cross-file for those types, which is the behaviour this PR is changing. They now carry the IDs the files' own nodes would be minted with, so they keep guarding those types at the boundary that still holds, and the ids are renamed `document-own-file-node` / `rationale-own-file-node` to say so.

Verified both directions: with `dedup.py` reverted the two merge tests fail; with the `_reads_as_file_entity` call stubbed out the four own-node/heading tests fail.

`pytest tests/ -q` was run on the clean base commit and again on this branch, on the same machine: **17 failed, 5250 passed** before and **17 failed, 5259 passed** after — the same 17 failures, identical sets, and +9 passed for the tests added here (a tenth landed in the follow-up commit). Those 17 are pre-existing on `v8` in this environment (Windows: `os.mkfifo`/unix-socket/symlink cases in `test_non_regular_files.py`, POSIX-path install/uninstall cases, and two unicode-normalization tests) and are unrelated to this change — no test that passed on the baseline fails after it. `ruff check` passes and `python -m tools.skillgen --check` is clean.

Addresses the option-1 half of #296. The issue as filed also asks for a configurable canonicalization subcommand and edge-dedup pass, which this does not add, so it should not auto-close.

### DIFF
--- a/graphify/dedup.py
+++ b/graphify/dedup.py
@@ -306,34 +306,43 @@ def _defines_id(node: dict) -> bool:
 # one of its sections. The markdown extractor stamps these precisely because
 # `file_type` cannot carry the distinction, and both are file-anchored in the
 # #1284 sense — two files' `## Decisions` sections are two sections (#3094).
+#
+# PRODUCER CONTRACT: an extractor that mints a node standing for a *part* of its
+# source file (a section, a sheet, a slide) must stamp one of these. The gate
+# below reads an unstamped node as an entity, so an unstamped structural node is
+# eligible to merge with its namesake in another file.
 _FILE_STRUCTURE_NODE_KINDS = frozenset({"page", "heading"})
 
 
-def _provably_not_file_structure(node: dict) -> bool:
-    """True only when the node is PROVABLY an entity found inside its file,
-    rather than a structural part of the file itself (#296).
+def _reads_as_file_entity(node: dict) -> bool:
+    """True when the node reads as an entity found inside its file, rather than
+    as the file itself or a stamped structural part of it (#296).
 
-    Two things disqualify a node, and either alone is enough:
+    Be precise about what is proven and what is assumed, because the two halves
+    differ in strength:
 
-    * It is its file's OWN node. ``_id_prefixes`` enumerates the ID a node
-      standing for ``source_file`` itself would carry, in every spelling a
-      stored path may take (absolute, repo-relative, or the pre-#1504 bare
+    * PROVEN — it is not its file's OWN node. ``_id_prefixes`` enumerates the ID
+      a node standing for ``source_file`` itself would carry, in every spelling
+      a stored path may take (absolute, repo-relative, or the pre-#1504 bare
       stem). A file's own node IS one of those; an entity extracted from that
-      file carries an ``_<entity>`` suffix, so it never equals one.
-    * It is a section of the file — ``node_kind: "heading"`` — or the file node
-      the extractor labelled outright, ``node_kind: "page"``. Repeated headings
-      across sibling documents are distinct sections, not duplicates
-      (#1284, re-verified on #3094), and stay blocked.
+      file carries an ``_<entity>`` suffix, so it never equals one. This is a
+      reconstruction, not a heuristic.
+    * ASSUMED — it is not a section of the file. That rests on ``node_kind``,
+      which only a producer that stamps it can attest. `heading`/`page` are
+      honoured when present, but ABSENCE OF THE MARKER IS NOT PROOF OF
+      ENTITY-NESS: a producer minting sub-file nodes without stamping
+      `node_kind` (see the contract above) yields structural nodes that this
+      returns True for, and two such nodes sharing a label in different files
+      would merge. The conservative fix is on the producer side — stamp
+      `node_kind` — not a guess here about what an unstamped node meant.
 
-    Fail-safe by construction: a node that cannot be checked (no ID, no
-    provenance) answers False and stays blocked, so this can only ever *narrow*
-    the set of nodes the cross-file gate treats as file-anchored — never widen
-    it on a guess.
+    A node that cannot be checked at all (no ID, no provenance) answers False
+    and stays blocked.
     """
     nid = node.get("id") or ""
     source_file = node.get("source_file") or ""
     if not nid or not source_file:
-        return False  # unprovable — leave the file-anchored block in place
+        return False  # uncheckable — leave the file-anchored block in place
     if node.get("node_kind") in _FILE_STRUCTURE_NODE_KINDS:
         return False  # the extractor says this node is part of the file's structure
     return nid not in _id_prefixes(source_file)
@@ -670,8 +679,8 @@ def deduplicate_entities(
         # it is provably safe (#2182). `concept` is the one file_type meant to
         # unify across files (#1284) — code is keyed by ID (#1205) and
         # image/paper labels are often shared basenames (logo.png), so both stay
-        # blocked. rationale/document join `concept` here ONLY when the node is
-        # provably not part of its file's own structure (#296): a file-anchored
+        # blocked. rationale/document join `concept` here ONLY when the node
+        # reads as an entity inside its file (#296): a file-anchored
         # *file_type* does not make an individual node file-anchored. An entity
         # extracted from a note — a person, a project — inherits `document` from
         # the file's extension, not from anything about itself, so in note-heavy
@@ -688,7 +697,7 @@ def deduplicate_entities(
             (n for n in group
              if (n.get("file_type") == "concept"
                  or (n.get("file_type") in _FILE_ANCHORED_NONCODE
-                     and _provably_not_file_structure(n)))
+                     and _reads_as_file_entity(n)))
              and (n.get("source_file") or "")
              and _entropy(n.get("label", "")) >= _ENTROPY_THRESHOLD),
             key=lambda n: n["id"],
--- a/graphify/dedup.py
+++ b/graphify/dedup.py
@@ -301,6 +301,44 @@ def _defines_id(node: dict) -> bool:
                for prefix in _id_prefixes(source_file))
 
 
+# `node_kind` values marking a node that is STRUCTURE of its source file rather
+# than an entity mentioned inside it: `page` is the file's own node, `heading`
+# one of its sections. The markdown extractor stamps these precisely because
+# `file_type` cannot carry the distinction, and both are file-anchored in the
+# #1284 sense — two files' `## Decisions` sections are two sections (#3094).
+_FILE_STRUCTURE_NODE_KINDS = frozenset({"page", "heading"})
+
+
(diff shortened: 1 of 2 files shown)


yotamleo:
The escalated finding — "Exact document/rationale entities now merge across files", flagged as agreed but not verified — is correct, and it is the change this PR is making rather than a side effect of it. Since it's marked as needing human review, here is the specific thing I think needs deciding, with a repro, because I found a live producer that lands on the unproven half.

`_reads_as_file_entity` has two halves of very different strength, and the docstring says so:

* **Proven** — the node is not its file's own node. `_id_prefixes` reconstructs the ID a node standing for `source_file` would carry in every spelling a stored path takes, and an extracted entity always carries an `_<entity>` suffix, so it can never equal one.
* **Assumed** — the node is not a *part* of its file. That rests entirely on the producer stamping `node_kind`, and absence of the marker is not proof of entity-ness.

I went looking for producers that mint sub-file `document`/`rationale` nodes without stamping `node_kind`, and the main comment-rationale extractor is one: `extract.py:1228` and `:1530` build `rid = _make_id(stem, "rationale", str(line))` — line-anchored, so squarely a part of its file — with no `node_kind` in the dict. That path is live, not flag-guarded.

The consequence, run against clean checkouts of `v8` @ `33362d9` and this branch's head `d123254`, on two nodes of exactly the shape `extract.py:1228` emits:

```python
LBL = 'Retry once on a transient socket error; the upstream API 502s under load.'
a = {'id': 'parser_rationale_41', 'label': LBL, 'file_type': 'rationale',
     'source_file': 'svc/parser.py', 'source_location': 'L41'}
b = {'id': 'writer_rationale_88', 'label': LBL, 'file_type': 'rationale',
     'source_file': 'svc/writer.py', 'source_location': 'L88'}
deduplicate_entities([dict(a), dict(b)], [], communities={})
```

```
base 33362d9  ->  2 nodes: ['parser_rationale_41', 'writer_rationale_88']
this branch   ->  1 node:  ['parser_rationale_41']     (_reads_as_file_entity: True True)
```

So two identical high-entropy comments in two different files — a repeated caveat, a copied header block — collapse into one `rationale` node. Two other producers have the same shape: `extractors/commonlisp.py:333` and `:423` (docstring nodes, `file_type="rationale"`, and that extractor's `add_node` never stamps `node_kind`), and `detect.py:666`, which mints `document` sub-file nodes for xlsx — latent rather than live, since the comment at `detect.py:650` says that codepath isn't wired into the dispatcher yet.

Whether that merge is wrong is a judgment call I'd rather you made than I did. It is arguably the right answer — one rationale stated in two places really is one rationale, and it keeps its edges to both parents. It is also exactly the class #1284 set out to block, and nothing in the graph records that the two comments were independent.

Two ways to close it, and I'll do whichever you prefer:

1. **Producer-side, conservative** — stamp a `node_kind` on the rationale nodes in `extract.py` (and the two commonlisp sites), so they're structural by declaration and the gate blocks them. Three small edits, no change to this PR's logic, and it makes the contract in the new comment true of the repo rather than aspirational.
2. **Leave it** — accept cross-file merging of identical comment rationales as intended, and treat the `node_kind` contract as guidance for future producers.

I've left the PR on (2) as it stands, because option 1 reaches into the extractors and a dedup PR shouldn't make that call uninvited. Say the word and I'll push option 1 onto this branch.

Unrelated to the finding, one note on Pass 2: it is untouched here — `_crossfile_fileanchored_blocked`, the fuzzy anchoring and the provenance requirement all still apply, so #1284's near-identical boilerplate and heading siblings stay blocked on that path regardless of which option you pick.

safishamsi:
Shipped in **v0.9.55** — landed on `v8` via cherry-pick with your authorship preserved. Thanks @yotamleo — cross-file entity nodes typed by file extension now merge (partial progress on #296). Release: https://github.com/Graphify-Labs/graphify/releases/tag/v0.9.55

PR #3220, opened 2026-08-30
fix(ts): preserve runtime dynamic imports in comparisons

### LINKED ISSUE #3210: TS: #3154 `import(...)` normalizer erases runtime dynamic imports; with `import ("…")` the dependency is lost entirely (regression in 0.9.52)
### Summary

The `import(...)` type-argument normalizer added in **v0.9.52** (#3154) rewrites TypeScript
source before parsing, but its regex is not anchored to an actual type-argument position.
Any `<` … `>` pair that has no `;`, `{` or `}` between it and is followed by `(` will have an
intervening **runtime** `import("…")` blanked out.

In the common spelling the file-level dependency is recovered by
`_rescue_js_dynamic_imports`, and only the function-anchored `imports_from` edge is lost.
But the normalizer accepts `import ("…")` (whitespace before the paren) while the rescue does
not, so for that spelling **the module dependency disappears from the graph entirely**.

v0.9.50 and v0.9.51 do not contain the function at all, so this is a new regression.

### Reproduction

Three files in one directory. `mod.ts` and `other.ts` must exist on disk for
`_resolve_js_module_path` to resolve the specifiers.

`mod.ts`
```ts
export const x = 1
```

`other.ts`
```ts
export const y = 2
```

`a.ts` — semicolon-less style, dynamic import inside a function body:
```ts
async function load(a: number, b: number) {
  const flag = a < b
  const m = await import("./mod")
  const ok = b > (a)
  return [flag, m, ok]
}
```

`reproduce.py`
```python
from pathlib import Path
from tempfile import TemporaryDirectory

with TemporaryDirectory() as cache:
    result = extract([Path("a.ts").resolve()], cache_root=Path(cache), parallel=False)

print(len(result["nodes"]), "nodes,", len(result["edges"]), "edges")
for e in result["edges"]:
    print(" ", e["source"], e["relation"], e["target"])
```

```bash
for v in 0.9.50 0.9.51 0.9.52; do
  python -m venv "venv-$v" && "venv-$v/bin/pip" install -q "graphifyy==$v"
  echo "--- $v"; "venv-$v/bin/python" reproduce.py
done
```

### Results

Node ids shortened below for readability; the raw ids are absolute-path derived.
Measured on CPython 3.14.7.

| `a.ts` body | 0.9.50 | 0.9.51 | 0.9.52 |
|---|---|---|---|
| **A** `await import("./mod")` | 3 edges | 3 edges | **2 edges** — `load -imports_from-> mod` lost |
| **B** `await import ("./mod")` | 2 edges | 2 edges | **1 edge** — dependency gone entirely |
| **C** A plus a second `await import("./other")` | 5 edges | 5 edges | **3 edges** — both `imports_from` lost |
| **D** control: A with the two comparison operators deleted | 3 edges | 3 edges | 3 edges — unchanged |

Case A, verbatim:
```
--- 0.9.50 / 0.9.51
2 nodes, 3 edges
  ext_a_ts contains a_load
  a_load imports_from ext_mod_ts
  ext_a_ts dynamic_import ext_mod_ts
--- 0.9.52
2 nodes, 2 edges
  ext_a_ts contains a_load
  ext_a_ts dynamic_import ext_mod_ts
```

Case B — same file with one space added before the paren:
```
--- 0.9.50 / 0.9.51
2 nodes, 2 edges
  ext_a_ts contains a_load
  a_load imports_from ext_mod_ts
--- 0.9.52
2 nodes, 1 edge
  ext_a_ts contains a_load
```
Nothing in the graph records that `a.ts` depends on `./mod` any more.

Control **D** changes only the two comparison operators (`const flag = a`, `const ok = b`)
and is stable across all three versions, so the trigger is the `<`/`>` pair — not the dynamic
import and not the missing semicolons.

### Root cause

`graphify/extract.py:1291`
```python
_TS_IMPORT_CALL_RE = re.compile(
    rb"\bimport\s*\(\s*['\"][^'\"\r\n]+['\"]\s*\)"
)
_TS_IMPORT_TYPE_CALL_RE = re.compile(
    rb"<((?:[^;{}]*?\bimport\s*\([^()]+\)[^;{}]*?)+)>(?=\s*\()"
)
```

`[^;{}]` matches newlines and nothing requires `<` to open a type-argument list, so in
semicolon-less TypeScript the span routinely crosses statement boundaries. Direct call, no
extraction involved:

```python
>>> from graphify.extract import _normalize_ts_import_types as N
>>> N(b'const a = x < y\nconst m = await import("./m")\nconst b = p > (q)\n').decode()
'const a = x < y\nconst m = await T            \nconst b = p > (q)\n'
```

The **B** case is a second, independent mismatch: the normalizer's outer pattern allows
`import\s*\(`, but `_rescue_js_dynamic_imports` bails on a literal substring at
`extract.py:1385` (`if "import(" not in src`) and its regex at `extract.py:1427` is
`(?<!\w)import\(`. So a spelling the normalizer erases is a spelling the rescue cannot see.

### Exact trigger constraints

A rewrite fires when the outer pattern can span from a `<`, through at least one
`import(...)` whose argument contains no nested parentheses (`[^()]+`), to a `>` followed by
optional whitespace and `(`, without crossing `;`, `{` or `}`. A `>` may appear inside the
span; only the `;{}` set terminates it.

For the source to actually change, the matched import must also satisfy
`_TS_IMPORT_CALL_RE` — a single-line, single- or double-quoted literal specifier. A template
specifier (`` import(`./m`) ``) can match the outer pattern but is not substituted, so the
source is unchanged.

The `+` on the repeat group does not require multiple imports; it permits several within one
span, and in case **C** both are blanked in the same pass.

`extract.py:1306` bails early unless the source contains `import(` or `import (`, so
`import  (` (two spaces) is not affected even though the outer regex would accept it.

### Scope

`_normalize_ts_import_types` has exactly two callers in 0.9.52:

- `extract.py:1347` — `extract_js`, guarded by `suffix in (".ts", ".tsx", ".mts", ".cts")`
- `extract.py:1817` — `extract_vue`, guarded by `config in (_TS_CONFIG, _TSX_CONFIG)`. `.vue`
  selects `_TS_CONFIG` when `lang` is **unset**, so a plain `<script>` block is normalized too.

Not affected: `.js`, and `.svelte` / `.astro` — both call `_extract_generic(path, _JS_CONFIG)`
and never reach the normalizer.

### Cache and `update` behaviour

Not a persistent cache poisoning issue: every TS suffix is in `_JS_CACHE_BYPASS_SUFFIXES`
(`extractors/models.py:11`), so the wrong result is not written to the AST cache — it is
recomputed, identically wrong, on every extraction.

The incremental path calls the same `extract()` (`watch.py:1642`), so a changed
`.ts/.tsx/.mts/.cts/.vue` file is mis-extracted under `graphify update` as well, and there is
no cache path by which a previously-correct edge is restored. I did not verify end-to-end how
the incremental merge removes the pre-existing edge.

### Constraints on a fix

Restricting the span to a single line (`[^;{}]` → `[^;{}\r\n]`) is **not** safe — it
reintroduces the original #3154 node loss for multiline type arguments. Measured, same
fixture, node labels only:

```ts
const r = load<
  typeof import("./mod")
>();
function after() {}
```
```
0.9.50: ['c1.ts']                  <- original #3154 bug: all symbols lost
0.9.52: ['c1.ts', 'r', 'after()']  <- fixed
[^;{}\r\n]: not normalized         <- back to the 0.9.50 result
```

It would also not achieve what it claims: the inner `import\s*\([^()]+\)` has no newline
restriction, so `load<typeof import(\n  "./mod"\n)>()` still normalizes across lines.

A structural check seems necessary — parse first, then rewrite only the byte ranges of
`import(...)` nodes the tree actually places under a type-arguments node, and reparse.
Whatever the approach, the normalizer and the rescue should agree on which import spellings
they each accept; today they do not, which is what makes case **B** lose the dependency
outright.

Suggested regression matrix:

| input | must hold |
|---|---|
| `load<typeof import("./mod")>();` | normalized, following declarations survive (#3154) |
| `load<\n typeof import("./mod")\n>();` | same |
| `load<typeof import(\n "./mod"\n)>();` | same |
| case **A** above | `load -imports_from-> mod` present |
| case **B** above | `load -imports_from-> mod` present |
| case **C** above | both `imports_from` edges present |

### Not reproduced (stated so nothing here is overclaimed)

- **No node loss from the false positive.** The rewrite replaces only the `import(...)` bytes
  with `T` plus padding, so the surrounding statement stays parseable. With functions declared
  before and after the affected span, all three function nodes still appear. Not swept
  exhaustively, so this is "not observed", not "impossible".
- **JSX is not a worse trigger.** In `.tsx` the comparison case above fails identically, but
  ordinary JSX does not fire the regex — the `{ }` of a JSX expression container terminates
  the span:
  ```tsx
  const v = <div>{await import("./mod")}</div>   // not normalized; imports_from edge intact
  ```

zfaustk:
#### What it does

Fixes #3210

The TypeScript `import(...)` normalizer could mistake a runtime dynamic import between comparison operators for a type argument. In semicolon-less code this removed the function-level dependency edge; with whitespace before `(` it could remove the dependency entirely. The normalizer now masks only parser-identified call/new type-argument ranges and uses a narrow error-adjacency check to leave valid runtime comparisons unchanged. The rescue path accepts the same whitespace-before-parenthesis spelling.

#### How to test

Focused local command on current `v8`:
`uv run --frozen pytest -q tests/test_ts_import_type_arguments.py tests/test_js_dynamic_imports.py tests/test_js_dynamic_import_affected.py tests/test_vue_extraction.py --tb=short`

Result: 44 passed.

An independent 48-case runtime matrix preserved both function-level and file-level edges for comparison, call, conditional, collection, arrow, and whitespace contexts. `graphify update .` completed with only the known optional-parser/Luau warnings. The full local suite has 16 baseline/environment failures (skillgen historical-ref and Ollama tests); those are not claimed as passed and remain for hosted CI.

#### Follow-ups

None identified.

#### Breaking changes

- [ ] This PR introduces breaking changes and requires careful review.

#### Attribution

N/A

#### Review checklist

- [x] As an author, I have thoroughly tested my changes and carefully followed the review guidelines
- [x] User-facing text is internationalized using the `nls` service (no new user-facing text introduced in this change)

### DIFF
--- a/graphify/extract.py
+++ b/graphify/extract.py
@@ -1303,33 +1303,185 @@ def _extract_python_rationale(path: Path, result: dict) -> None:
 _TS_IMPORT_CALL_RE = re.compile(
     rb"\bimport\s*\(\s*['\"][^'\"\r\n]+['\"]\s*\)"
 )
-_TS_IMPORT_TYPE_CALL_RE = re.compile(
-    rb"<((?:[^;{}]*?\bimport\s*\([^()]+\)[^;{}]*?)+)>(?=\s*\()"
-)
 
 
-def _normalize_ts_import_types(source: bytes) -> bytes | None:
-    """Rewrite TypeScript `import(...)` type arguments in call expressions
-    to standard type identifiers of identical byte length (#3154).
+def _ts_import_is_code(root: Any, start: int) -> bool:
+    """Return whether an import-call match starts in executable source.
 
-    Preserves byte length, newlines, and source offsets so all downstream node
-    source_location metadata remains 100% accurate.
+    Regex matching is only used to locate a literal specifier; comments,
+    strings, regular expressions, and template text must never be fed into the
+    structural masking pass.  A template substitution is executable again, so
+    it is the one exception to the ``template_string`` guard.
     """
-    if rb"import(" not in source and rb"import (" not in source:
+    node = root.descendant_for_byte_range(start, start + 1)
+    if node is None:
+        # A malformed tree can leave a byte range uncovered. Treat it as code
+        # and let the structural pass decide whether it belongs to a type list;
+        # never turn an uncertain lexical result into a crash.
+        return True
+    in_template_substitution = False
+    while node is not None:
+        if node.type == "comment" or node.type in ("string", "regex", "regex_pattern"):
+            return False
+        if node.type == "template_substitution":
+            in_template_substitution = True
+        elif node.type == "template_string" and not in_template_substitution:
+            return False
+        node = node.parent
+    return True
+
+
+def _ts_type_argument_ranges(root: Any, *, call_only: bool) -> list[tuple[int, int]]:
+    """Collect byte ranges tree-sitter already parsed as type arguments."""
+    ranges: list[tuple[int, int]] = []
+    stack = [root]
+    while stack:
+        node = stack.pop()
+        if node.type == "type_arguments":
+            parent = node.parent
+            if not call_only or (
+                parent is not None and parent.type in ("call_expression", "new_expression")
+            ):
+                ranges.append((node.start_byte, node.end_byte))
+        stack.extend(node.children)
+    return ranges
+
+
+def _ts_error_nodes(root: Any) -> list[Any]:
+    """Return parser error nodes without depending on a grammar's error name."""
+    errors: list[Any] = []
+    stack = [root]
+    while stack:
+        node = stack.pop()
+        if node.type == "ERROR" or node.is_error:
+            errors.append(node)
+        stack.extend(node.children)
+    return errors
+
+
+def _ts_mask_candidate_is_malformed(
+    original_root: Any,
+    masked_range: tuple[int, int],
+    errors: list[Any],
+) -> bool:
+    """Tell whether a masked generic is backed by an actual parse failure.
+
+    A valid runtime comparison can have the same token shape as a generic call
+    after the import is replaced (``a < import("x") > (a)``).  Tree-sitter
+    exposes the opening ``<`` as a binary operator in the original tree for
+    both forms, so the structural second pass alone cannot distinguish them.
+    Only repair that ambiguity when the original parser has an error adjacent
+    to the candidate's closing angle; that is the signature of the known
+    ``import(...)``-type grammar failure.  Valid comparisons, including ones
+    nested in another call's arguments, remain byte-for-byte untouched.
+    """
+    start, end = masked_range
+    opener = original_root.descendant_for_byte_range(start, start + 1)
+    if opener is None or opener.type != "<":
+        # A grammar that already gives us a structural type_arguments node is
+        # safe to normalize; no binary/comparison ambiguity is present.
+        return True
+    if opener.parent is None or opener.parent.type != "binary_expression":
+        return True
+
+    # A malformed generic's ERROR starts at the closing `>` (or at the call
+    # punctuation immediately following it). Keep this window deliberately
+    # narrow so an unrelated syntax error elsewhere cannot authorize masking a
+    # valid runtime comparison.
+    for error in errors:
+        if error.start_byte <= end + 2 and error.end_byte >= end - 1:
+            return True
+    return False
+
+
+def _normalize_ts_import_types(source: bytes, *, tsx: bool = False) -> bytes | None:
+    """Rewrite only syntactic TypeScript ``import(...)`` type arguments.
+
+    The first implementation of #3154 used a ``<...>`` regular expression.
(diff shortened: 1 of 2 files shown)


safishamsi:
Shipped in **v0.9.55** — landed on `v8` via cherry-pick with your authorship preserved. Thanks @zfaustk — a runtime dynamic import() is no longer blanked by the TS type-argument normalizer. Release: https://github.com/Graphify-Labs/graphify/releases/tag/v0.9.55

PR #3221, opened 2026-08-30
feat: add F# extractor (.fs/.fsx) via ionide tree-sitter-fsharp

pcshrosbree:
## What

F# support: `.fs`/`.fsx` files are extracted via ionide's [tree-sitter-fsharp](https://github.com/ionide/tree-sitter-fsharp) (0.3.11, prebuilt abi3 wheels for every platform), as an optional `[fsharp]` extra — matching the recent precedent of `[ocaml]`/`[commonlisp]`/`[robot]`.

Extracted: modules, namespaces, record/union/interface/enum/class types, union cases, class members, let-bound functions and values, `open` imports, and calls — including **pipeline application** (`x |> f`, `f <| x`, and the 2/3-ary variants), which is where most F# call edges live.

`.fs`/`.fsx` join the **`dotnet` interop family** in all three family maps (`extract._LANG_FAMILY_BY_EXT`, `analyze._LANG_FAMILY`, `build._EDGE_LANG_FAMILY`), so an F# reference to a C# type rewires onto the real definition instead of dangling.

## How

`graphify/extractors/fsharp.py` follows `extractors/ocaml.py` closely (F# is ML-family): the same sourceless ref-stub discipline (#1402), the same local-definition table with ambiguity tracking, the same two-pass call resolution, and the same qualified-call gating (a call `M.f` where `M` is not defined locally stays a distinct stub keyed by its full path, so `List.map` can never absorb a local `map`).

F#-specific handling, grounded in live AST probes of the grammar rather than `tags.scm` alone:

- Names are nested, not a field: `type_definition > *_type_defn > type_name`, `function_declaration_left > identifier`, `value_declaration_left > identifier_pattern > long_identifier_or_op`, and `method_or_prop_defn > property_or_ident` (last identifier — `this.Run` carries two).
- Pipe operators are `infix_expression`s: the callee is the right operand for `|>`-family, the left for `<|`-family; both feed the same call-site table as `application_expression`.

## Verification

Beyond the unit tests (`tests/test_fsharp.py`), two things this PR did that I'd flag for review:

1. **One named test per registration point, mutation-checked.** `tests/test_fsharp_registration.py` covers each wiring point (dispatch ×2, `CODE_EXTENSIONS`, watch, extra-hint, and all three family maps) individually. I then removed each registration in turn and confirmed a test fails. The first run caught an untested `.fsx` dispatch entry — the exact silent-skip failure mode that made F# invisible in the first place (a file matching *no* category produces no error).

2. **Graded on a real mixed Rust/C#/F# repo** (a .NET instrumentation platform, ~470k LOC, 100 F# files including a full F# sidecar service): 2,361 sourced nodes, 10,238 call edges; 20 randomly sampled call edges all verified by hand at their cited `file:line`. Cross-language: 192 C#↔F# edges where the release build produces zero.

The grade also caught a bug the fixture tests missed, now fixed with a regression test: `let subscribe (a: A) (b: B) : IDisposable = ...` parses as a `value_declaration_left` whose *last* identifier is the return-type annotation — the first cut minted a sourced `IDisposable` node, and the `dotnet` family rewire then bound every BCL `implements IDisposable` stub in the corpus onto it (82 false cross-language edges). The name is now taken from the first `long_identifier_or_op` under `identifier_pattern`.

## Known limits (deliberate v1 scope)

- `.fsi` signature files not wired (`language_signature()` exists in the grammar; follow-up).
- Computation-expression internals (`async { }` / `task { }` bodies) and `#r`/`#load` script directives are not specially handled.
- Object expressions (`{ new IDisposable with ... }`) contribute their member bodies' calls but no interface edge.

## Checks

- `uv run pytest tests/ -q`: 5,315 passed; the only 2 failures (`test_labeling.py::test_label_communities_batches_when_over_batch_size`, `test_watch.py::test_watch_handler_honors_graphifyignore`) reproduce on clean `v8` without this change.
- `uv lock` updated; `uv sync --all-extras --frozen` clean.
- CHANGELOG entry added under 0.9.53 (move to the next version as appropriate).


pcshrosbree:
Thanks for the review — responding to all three findings (addressed in 1129f64):

**Same-named members in different F# types merged into one node** — confirmed, fixed. `emit_member` keyed node ids by `(file, member)` only, so two types in one file each having a `Dispose` collapsed into a single node. Member ids now carry the owning type's label (`_make_id(stem, owner_label, mname)`); on a real mixed Rust/C#/F# corpus this un-merged 18 member nodes. Regression test added (`test_same_named_members_of_different_types_stay_distinct`).

**Qualified-call test passes if the call edge is missing entirely** — confirmed, fixed. The test asserted only inside a loop over matching edges — vacuously green when no edge was emitted at all. It now first asserts the edge exists, then checks it stays a sourceless stub. Fair catch: I mutation-checked the registration points but not this test's own assertions.

**First named child of application_expression not treated as callee** — assessed, no change, reasoning below; happy to be corrected with a concrete case. The extractor records a callee only when `application_expression`'s first named child is a `long_identifier_or_op` (bare or dotted name). When it's something else — a nested application, as in curried `f x y` which parses as `application(application(f, x), y)` — the *outer* node is deliberately skipped and the walk descends to the inner application, which does carry the `long_identifier_or_op` head. Recording at the outer node too would double-count the same call site. Parenthesized/lambda heads (`(fun x -> ...) arg`) genuinely produce no edge, which matches how the OCaml extractor treats them.

All 21 F#-related tests green; corpus spot-check re-run after the fix (20/20 sampled call edges verified at their cited source lines).

pcshrosbree:
Second hardening pass pushed in 4a3c419 — an independent adversarial review over the diff surfaced ten further findings; each was verified against the live grammar with AST probes before fixing, and each carries a regression test:

- **Dotted callees were dropped entirely** (`application > dot_expression`): `Grasp.Telemetry.init args`, `System.IO.File.ReadAllText p`, and `x |> Some.Deep.fn` emitted no call edge. Fixed; on a real 100-file F# corpus this adds **+381 raw call edges**.
- **`let rec f ... and g ...`**: only the first head was minted — `g` vanished into a rewireable stub and its body's calls mis-attributed to `f` (false `f → f` self-loop). All and-joined heads now mint, with per-binding call attribution.
- **Namespaces are now canonical**: `namespace Grasp.Core` emits the same `csharp_namespace:<sha1>` id as the C# extractor with `type: "namespace"`, so multi-file namespaces merge into one hub, `_canonicalize_csharp_namespace_nodes` applies, and the unique-stub rewire skips them. Namespace segments also no longer whitelist local qualified-call binding — under `namespace Grasp.Sidecar`, `Sidecar.validate c` stays a stub instead of falsely binding to a same-file `validate` with EXTRACTED confidence.
- **Member labels use the dotnet `.Name()` convention**, so F# members participate in the same rewire indexing as C# methods instead of posing as type-like unique-stub targets.
- **Union cases are type-scoped** (two same-file `Ok` cases stay distinct) and the canonical single-case wrapper `type Email = Email of string` no longer emits a `contains` self-loop.
- **Enum members** (`enum_type_cases`) are emitted; previously only DU cases were.
- **A comment inside a pipeline** (`x // note` newline `|> f`) no longer kills the call edge (comment nodes are filtered before operand counting).
- **Destructuring `let (a, b) = ...`** mints one definition per bound name instead of none.
- **`.fs` vs GLSL**: `.fs` is also the fragment-shader extension; dispatch now sniffs (mirroring the `.m` ObjC/MATLAB precedent) and returns no extractor for shaders instead of ERROR-parsing them into sourceless dotnet-family stubs. Maintainer judgment welcome on the marker list.
- **Registration tests** no longer module-level `importorskip`: the pure-literal registry guards run even without the grammar; only the two dispatch tests skip individually.

State after this commit: 30 F#-specific tests, every registration mutation-killed, full suite green apart from the two failures that reproduce on clean `v8`, and 20/20 randomly sampled corpus call edges verified at their cited `file:line`.

pcshrosbree:
Round 3 pushed in e4b1546 — a four-arm adversarial panel (Codex CLI + three Claude models, identical brief, each instructed to execute probes rather than speculate and to tag VERIFIED vs SUSPECTED) over the round-2 head. Nine distinct defects survived cross-verification; every one was independently reproduced against the live grammar before fixing, and each carries a regression test.

The highlights, in severity order:

- **Generic type names were wrong everywhere**: `type Box<'T>` minted a node labelled `T` (the last type parameter — `type_name` carries `type_arguments` as siblings of the name), and a `when 'T :> System.IDisposable` constraint minted a *sourced IDisposable* — the same corpus-poisoning hazard round 2 fixed for annotated lets, resurfacing on a new path. Three of four arms found this independently.
- **`type X with` impersonation** (all four arms): a type *augmentation* minted a sourced node for the foreign type it extends; verified end-to-end that a C# `class Foo : Widget` rewired its `inherits` edge onto the F# extension file. Members now attach to a sourceless stub.
- **Heritage edges now exist**: `inherit Base()` → `inherits`, `interface I with` → `implements` (INFERRED, to stubs). Previously nothing was emitted, so `_rewire_unique_stub_nodes`'s supertype guard could never protect an F#-referenced base type — and F# was absent from the 0.9.53 dotnet-family heritage work despite joining the family.
- **Object expressions** (all four arms): `{ new IFoo with member this.Dispose() = ... }` minted a phantom module-level member and — the sharpest version of the finding — poisoned `local_defs` so a later *real* same-named binding turned ambiguous and unresolvable. Members are now anonymous; body calls attribute to the enclosing binding.
- **Active patterns and operators** (all four arms): `let (|Even|Odd|)` and `let (+.)` minted nothing and mis-attributed their body calls to the enclosing module. Both now mint (delimited spelling as label — ends in `)`, so excluded from the type-rewire by construction).
- **Definition ids now chain from their container** (C#'s parent-id convention, with kind tags): sibling modules' same-named `run` bindings, and the companion `type Config` + `module Config` idiom, previously merged into single nodes with unioned call attribution.
- **Function labels use the engine `name()` convention** — a bare-labelled F# function passed `_is_type_like_definition` and could absorb e.g. a Python `parse()` stub in the unique-stub rewire (`_node_label_key` strips punctuation, so cross-file matching is unaffected). Worth noting: `ocaml.py` has the same bare-label property; pre-existing, not touched here.
- **`open` now mirrors `_import_csharp`**: EXTRACTED `imports` edge from the file node to the full-FQN id with `target_fqn` metadata. The old last-segment stub let `open System.Text` rewire onto an unrelated class named `Text` (reproduced), and F# opens were invisible to passes keyed on `relation == "imports"`.
- **GLSL sniff hardened + its test made mutation-killable**: a modern no-`#version` shader (`in vec3` / `out vec4` / `void main()`) sailed past the markers via the bare `//` F# marker; and one arm proved the guard test passed even with the guard deleted. New markers, and a test whose shader deliberately carries an F# marker inside a comment so removing the GLSL check flips it.

State: 40 F#-specific tests; every registration mutation-checked and killed; 20/20 randomly sampled corpus call edges verified at their cited `file:line`; heritage edges confirmed appearing on a real mixed corpus (5 `inherits` / 3 `implements` from F# sources); full suite green apart from the pre-existing `v8` failures.

pcshrosbree:
Round 4 pushed in fc71a97 — this round ran a redesigned review process, and both the process and its yield are worth a maintainer's minute.

**Process:** (1) a mechanical **coverage oracle** built *before* any review — `tests/fsharp_coverage_oracle.py`, 58 F# constructs with presence *and* commission expectations (forbidden-label sets, self-loop checks; run it with `uv run python tests/fsharp_coverage_oracle.py`); (2) a frozen corpus baseline for row-level differential reconciliation of every fold; (3) five parallel review arms with differentiated lenses (corpus-integration, end-to-end pipeline, test-adequacy/mutation, grammar coverage, plus a no-lens control); (4) cross-examination of contested fix shapes before implementing; (5) fixer≠verifier — each fix re-verified against its original repro by a different arm.

**Yield: 13 verified findings, zero false claims across the panel.** Highlights:

- **The round-3 GLSL sniff over-corrected**: real F# containing `// Draws a sample from a uniform distribution` was silently dropped whole-file. My first redesign was then *killed in cross-examination* by a real corpus line starting `float count / float total` — the shipped version is evidence-tiered (comment-stripped, line-anchored, strong-F#-declarations override weak type-keyword evidence, 64KB window), and all five adversarial cases are pinned as tests.
- **Pure-F# corpora lost every `open` edge**: `_resolve_cross_file_csharp_imports` is language-agnostic (reads `target_fqn` metadata) but was gated on a `.cs` file existing in the corpus; `build` then silently pruned the dangling edges. The gate now includes `.fs`/`.fsx` **for the imports pass only** — the C# type-reference resolver stays C#-gated because its metadata contract (`metadata.namespace`, `scope_chain`, `ref_token`) is C#-shaped. Documented follow-up: teaching F# nodes that contract is also what cross-language constructor-call resolution (`open CSharpLib; Doubler()`) needs; it remains unresolved by design in this PR.
- **The object-expression `references` edge was dead code** — wrong child type matched; it fired on 0 of 11 real corpus object expressions. Now fires, with a positive test.
- **Generic heritage was silently dropped** (`inherit Base<'T>()`, `interface IComparer<'T> with` wrap in `generic_type`) — the same bug class round 3 fixed for `type_name`, now with a shared type-reference reader. 2 live recoveries on the test corpus.
- **`static let` in classes** minted nothing (member calls to them dangled on rewireable stubs); **class-internal `let capacity` case-folded onto member `.Capacity()`** via `_make_id`, deleting a real public member and creating a false self-loop — member ids now carry a kind tag.
- **Composition `>>`/`<<`** records both operands as callees (deliberately *not* generalized to custom operators — Kleisli composes functions but bind has a data operand; there is no sound generic rule).
- **Mutation-proofing**: a 47-mutant sweep found the round-2 annotated-let regression test vacuous (its own source parses down the other branch — the naive reversion changed 546 corpus edges with every test green), and two whole features (`local_containers` positive binding, abstract-member emission) deletable without any test noticing. All pinned now.

**State:** 56 F# tests + 58/58 oracle constructs; every registration and new guard mutation-killed; full suite 5,354 passed / 0 failed; the corpus differential against the frozen baseline reconciled row-for-row (+27/−9, each class explained in the commit message); 20/20 sampled call edges verified at cited `file:line`.

The oracle file is included in the PR deliberately: it's ~150 lines, runs in seconds with no API calls, and turns the omission-defect class — which dominated all four review rounds — into a mechanical check any future language extractor can copy. Happy to drop it from the diff if you'd rather keep the PR minimal.

pcshrosbree:
Response to the latest automated review (5 findings) — verified each before acting, per this PR's standing practice:

**Refuted with evidence (the two HIGHs):** `tests/fsharp_coverage_oracle.py` is not truncated and contains no dangling `if`. At the reviewed head: `git show fc71a97:tests/fsharp_coverage_oracle.py | python3 -c "import sys; compile(sys.stdin.read(),'o','exec')"` → compiles; 250 lines in both the commit and the working tree, file ends with a well-formed `main()` guard; `uv run python tests/fsharp_coverage_oracle.py` → `58/58 constructs covered`, exit 0. The review's own caveat ("agreed by 2 of 2 members but NOT verified") applies — it looks like the analyzer saw a truncated window of the file.

**Accepted and fixed (latest commit):**
- *Hardcoded temp dir* — real portability bug (`dir="/data/tmp"` was machine-local); the oracle now uses the system default and runs anywhere, including CI.
- *Missing-grammar test skipped when the grammar is missing* — correct and pleasingly ironic: the test fakes the ImportError, yet module-level `importorskip` skipped it exactly where it mattered. Moved to the non-skipped module.
- *BOM-prefixed `.fs` rejected* — refuted as stated (probed: BOM files dispatch fine, since later lines carry markers), but a single-marker-line BOM file is a real theoretical hole and the hardening is one line: the sniffer now strips a leading UTF-8 BOM. Regression test added.

57 tests + 58/58 oracle green after the fold.

pcshrosbree:
Response to the latest automated review (5 findings), each verified before acting:

**Real — fixed in the latest commit, and credit where due:** *Qualified local F# calls resolve by short member name only* is a genuine defect that four adversarial review arms missed and this review caught first. Reproduced: `module A = let helper ...` / `module B = let bad z = B.helper z` bound `B.helper` EXTRACTED to **A's** helper — the gate checked that the qualifier was *some* local container, not that the callee belonged to it. Fixed with a per-container member index: qualified calls bind only through the owning container; non-owning or ambiguous qualifiers yield full-path INFERRED stubs. On the real corpus this demoted 85 receiver-qualified calls from false local binds to honest stubs. Regression test added.

**Refuted with evidence (including the HIGH):**
- *run_one drops accumulated failures* / *self-edge check raises NameError* — `run_one` accumulates into `problems` and returns it once, on every path; `lab` is defined in its scope before use. The oracle runs end-to-end: `58/58 constructs covered`, exit 0 — a NameError or dropped-failure path would be visible there.
- *Undefined variable in F# extension test* — the cited region is well-formed; the file's 40 tests all pass.

**Working as designed:** *sniffer rejects F# with no strong markers in 64KB* — a compilable `.fs` file must open with a `module` or `namespace` declaration (attributes/comments aside, all of which are marker-bearing line-starts), so the no-evidence default of "not F#" cannot reject valid compilable F#; it exists to keep Forth and marker-free shaders out.

58 tests + 58/58 oracle + full suite 5,356/0 after the fold.

pcshrosbree:
Response to the latest automated review (5 findings), verified per standing practice:

**Taken (1):** *detection reads entire file despite 64KB limit* — correct as a micro-issue: `read_bytes()[:65536]` slurps the file before slicing. Now a bounded `read(65536)`.

**Refuted by probe (2):** *F# interface implementations are never emitted* — they are, on every definition shape: probed a record-shaped `type R = { X: int } interface System.IDisposable with ...` and got the `R → IDisposable` implements edge; classes and generics are pinned by existing tests (`test_heritage_edges_emitted`, `test_generic_heritage_edges_emitted`) and the oracle's `interface impl` construct. *find_all drops nested matches* — true of the helper's traversal in isolation, false in effect for its only use: the grammar nests inner destructuring patterns beside `identifier_pattern` (under `paren_pattern`/`repeat_pattern`), never inside one, so `let (a, (b, c)) = ...` mints all three names (probed; now pinned by `test_nested_destructuring_let_mints_all_names`, with an intent comment on the helper).

**Repeat of a refuted finding (1):** *self-edge detection raises NameError* at `fsharp_coverage_oracle.py:250` — identical to the previous round's report, refuted there with compile-and-run evidence; still: `uv run python tests/fsharp_coverage_oracle.py` → `58/58 constructs covered`, exit 0.

**Declared deferral (1):** *.fsi signature files not routed* — deliberate v1 scope from the PR description onward (`language_signature()` exists in the grammar; the follow-up is documented in the module docstring). Happy to wire it in this PR if maintainers prefer.

59 tests + 58/58 oracle green after the fold.

pcshrosbree:
Latest automated review (5 findings), verified per standing practice:

**Real — fixed (1):** *`(*` matched as strong F# evidence* — probe-confirmed: a Forth `.fs` file with a stack-effect comment line (`(* stack: a b -- a*b )`) dispatched to the F# extractor. The marker existed for block-comment license headers, but a compilable F# file always surfaces a real declaration line within the 64KB window, so it was pure downside. Dropped, with regression tests both ways.

**Refuted, for the record (both HIGHs):** *extractor unparsable due to indentation* and *undefined name at :260* — the file compiles (`python3 -c "compile(open('graphify/extractors/fsharp.py').read(),'f','exec')"`) and all 41 tests in the module pass at the reviewed head; the cited region is the well-formed `type_name_parts`. *Undefined variable in test :291* — same evidence, tests pass.

**Repeat (1):** the oracle self-edge NameError is now reported a third time unchanged; refuted twice previously with run evidence (`58/58 constructs covered`, exit 0).

pcshrosbree:
Latest automated review (5 findings), verified per standing practice:

**Real — fixed (1):** *GLSL word-markers can shadow legitimate F#* — probe-confirmed for the word class: `uniform ` as STRONG evidence dropped a real F# file whose continuation line calls a function named `uniform`. The strong tier now holds only shapes impossible as F# line-starts (`#version`/`#extension`, `layout`, typed io decls `in vec3`-style, `void main`, `gl_`); `uniform`/`varying`/`precision` are weak evidence, which never overrides a strong F# declaration. Regression tests both ways; every previously-pinned shader case still rejects; 0 of 100 real-corpus files change classification. (The `in vec`-specific half of the claim did not reproduce — verbose-syntax `in` lines don't start with `in vec<digit>` — but the tightening covers it anyway.)

**Repeats of refuted findings (4):** *type_name_parts undefined name* (2nd repeat — file compiles, 63 tests pass), *interface implementations never emitted* (2nd repeat — probed on class, record, and generic shapes; edges present and test-pinned), and both oracle claims (*never executes checks* / *run_one never returns failures*, 4th round of this family — `uv run python tests/fsharp_coverage_oracle.py` → `58/58 constructs covered`, exit 0, and it exercised exactly these checks to catch real findings in earlier rounds). Per this thread's practice, repeats get this one line and no further litigation.

pcshrosbree:
Latest automated review (3 findings):

**Real — fixed (1):** *`layout` as strong GLSL evidence* — probe-confirmed, same word-vs-directive class as the previous round's `uniform`: an F# continuation line calling a function named `layout` dropped the file. Demoted to weak evidence (the tier that never overrides an F# declaration); the `layout(...)` qualifier shader still rejects via `void main`/weak accumulation. Regression tests both ways; 0 of 100 corpus files change classification.

**Repeats (2):** *type_name_parts returns the function* — third phrasing of the `:257`/`:260` claim; the file compiles and 65 tests + the 58-construct oracle (which exercises a dozen type shapes) are green at the reviewed head. *Oracle self-edge NameError* — fourth report; same run evidence as before (`58/58 constructs covered`, exit 0).

pcshrosbree:
Rounds 8–10 each surfaced one word-shaped sniff marker that was strong enough to drop a real F# file (`(*`, `uniform `, `layout`). Rather than wait for round 12 to find the next one, this round audits the entire remaining strong tier at once, with the grammar itself as the oracle: for each marker, can tree-sitter-fsharp parse a line starting with it?

**Every word-shaped marker fails the audit** — each is a valid F# line-start:

- `gl_` — ordinary identifier prefix; OpenGL interop code mirrors C names (`gl_makeCurrent ctx` on a continuation line parses clean);
- `in vec2/3/4` — verbose syntax puts `in` at a line start (`let x = 1.0` / `in vec2 x x` is valid, and `vec2` is a natural constructor-function name in F# math code);
- `out vec2/3/4` — `out` is not an F# keyword; `out vec3 v` is a plain application;
- `void main` — `void` is not a lexer-level F# keyword either (`let void = 1` parses); `void main` is a legal application.

Only `#version` and `#extension` are grammar-invalid at an F# line start. The strong tier is now exactly those two directives (`9b5ca4b`); every word-shaped marker is weak, where it still rejects directive-free shader fragments but can never override an F# declaration. Each demotion is pinned by an acceptance test that kills its re-promotion mutant, and the delete-strong-tier mutant is re-anchored on `#version` (the old carrier test's shader now rejects via the weak path, which would have left that mutant surviving silently). Corpus reclassification: 99/99 real `.fs` files unchanged, so extractor output is identical. 70 F# tests + 58/58 oracle + full suite green.

On this round's five advisory findings: the `weak_glsl 'int '/'bool '` collision claim fails its probe — the weak tier never overrides a strong F# declaration, and a compilable `.fs` file always carries one in the 64 KB window (probed live: `module M` + `int x`/`bool.Parse y` continuation lines is accepted). The `type_name_parts`-undefined, oracle-NameError, and test-undefined-variable items repeat entries already refuted in this thread (the cited `tests/test_fsharp.py:286` is a line inside an F# string literal; the test runs green, which a NameError would prevent). The coupling-delta inline notes are acknowledged as descriptive: a language extractor's `walk()` is a dispatch hub by design, matching the shape of the sibling extractors.

pcshrosbree:
For the record on the round-12 HIGH 'F# extractor appears to introduce an unexpected indent': probed at 9b5ca4b — py_compile and tokenize both pass clean on graphify/extractors/fsharp.py (no IndentationError/TabError, zero tab characters), and the full suite (5,367 tests) imports and executes the module. A real unexpected indent would make the module unimportable. The other four findings this round repeat items already refuted above (type_name_parts, the oracle self-edge NameError, interface-impls, and the tests/test_fsharp.py 'undefined variable' — still a line inside an F# string literal; the anchor drifted :286 to :288).

PR #3222, opened 2026-08-31
Finish member calls across a repo boundary on merge (#3152)

### LINKED ISSUE #3152: merge-graphs / global add can never produce cross-repo member-call edges: the resolved receiver type is discarded at bail-out instead of persisted
## Summary

Merging two graphs can never produce an instance/member call edge across the repo boundary, because
the receiver type is computed and then **thrown away** at every resolver bail-out. When the receiver's
type has no definition in the current corpus, `if len(type_defs) != 1: continue` drops the whole call
site, and nothing about it reaches `graph.json` — which is the only thing `merge-graphs` and
`global add` consume. So the information needed to complete the edge exists at extraction time, in a
variable, and is discarded milliseconds before the only artifact that outlives the run is written.

This is the one thing standing between per-repo graphs and a usable multi-repo call graph: `A` calling
`b.method()` on a type owned by `B` is the normal shape of a cross-repo dependency, and it is exactly
the shape that vanishes.

`graphifyy` 0.9.51 (`v8`), Python 3.10, macOS 14.

## Reproduction

Two repos, plus a control that puts the same code in one corpus.

```java
// repo-a/src/appa/App.java
package appa;

import libb.Greeter;

public class App {
    private Greeter greeter = new Greeter();

    void run() {
        Greeter local = new Greeter();
        local.greet("world");
        this.greeter.greet("field");
    }
}

// repo-b/src/libb/Greeter.java
package libb;

public class Greeter {
    public String greet(String who) {
        return "hi " + who;
    }
}
```

```console
$ (cd repo-a && graphify update . --no-cluster)
$ (cd repo-b && graphify update . --no-cluster)
$ (cd both   && graphify update . --no-cluster)     # control: both files, one corpus
$ graphify merge-graphs repo-a/graphify-out/graph.json repo-b/graphify-out/graph.json --out merged.json
Merged 2 graphs -> 8 nodes, 7 edges
```

```
[repo-a]  nodes=5 edges=5
   calls: .run() -> Greeter  [EXTRACTED]          <- to a SOURCELESS STUB, the type not the method
   sourceless stub: repo-a::greeter        / Greeter
   sourceless stub: repo-a::libb_greeter   / libb.Greeter

[both]    nodes=6 edges=8                          <- control
   calls: .run() -> Greeter  [EXTRACTED]
   calls: .run() -> .greet() [INFERRED]           <- THE edge that matters

[merged]  nodes=8 edges=7
   calls: .run() -> Greeter  [EXTRACTED]
   cross-repo edges: 0                             <- .greet() unobtainable, both call sites
```

8 edges when the code lives together, 7 after merging the same code from two repos. The missing one is
`App.run -> Greeter.greet`, and no merge-time pass can recover it: repo-a's `graph.json` has no record
that a `greet` call ever happened. The type reference survives as a sourceless stub; the *method call*
does not.

The extractor knew the answer. From `repo-a/graphify-out/cache/ast/v0.9.51-s2/*.json`, both call sites:

```json
{"callee": "greet", "caller_nid": "$graphify-root$_src_appa_app_app_run", "is_member_call": true,
 "lang": "java", "receiver": "local",       "receiver_type": "Greeter",
 "source_file": "src/appa/App.java", "source_location": "L10"}
{"callee": "greet", "caller_nid": "$graphify-root$_src_appa_app_app_run", "is_member_call": true,
 "lang": "java", "receiver": "this.greeter", "receiver_type": "Greeter",
 "source_file": "src/appa/App.java", "source_location": "L11"}
```

`receiver_type: "Greeter"` — already resolved, already serialized. Just not into `graph.json`.

## Root cause

Every language's member-call resolver ends the same way. Java, `extract.py:3600-3615`:

```python
type_name = raw_call.get("receiver_type")
if not type_name and receiver[:1].isupper():
    type_name = receiver
if not type_name:
    continue
type_defs = type_def_nids.get(key(type_name), [])
if len(type_defs) != 1:
    continue                     # <- receiver_type known, call site dropped, nothing persisted
```

Same bail-out: `extract.py:2895`, `:3308`, `:3318` (C++), `:3043` (Python), `:3760`, `:3771`, `:3785`
(ObjC), `_resolve_csharp_member_calls` at `:3341`. At each of those `continue`s the frame holds
`caller_nid`, the receiver type name, `callee`, `source_file` and `source_location` — a complete
description of an unresolved call, discarded.

The single-definition guard itself is right: guessing between two same-named types would be worse than
nothing. The problem is that "cannot resolve **here**" is treated as "does not exist", when in a
multi-repo setup it usually means "defined in a repo not yet in this corpus".

## Why the machinery for the fix is already there

- **Node metadata survives the merge.** `cross_repo_types.link_shared_type_declarations` (#3007,
  called at `cli.py:2567`) reads `metadata.namespace` **after** repo-prefixing and `nx.compose`, and
  adds edges from it. That is a working precedent for exactly this: carry a fact on the node, resolve
  it after merge.
- **Lifecycle is free.** `prefix_graph_for_global` carries node metadata through and
  `prune_repo_from_graph` removes it with the node, so nothing new is needed for `global add` /
  `global remove`.
- **The residue is already computed** and already written to the AST cache shards. Shards are not
  merge inputs, so the only missing step is getting the same tuple into `graph.json`.

Prior art worth a look: CodeGraph persists an `unresolved_refs` table and resolves it on later
indexing passes, which is the same idea with a database instead of node metadata.

## Suggested shape

1. **Persist the residue instead of dropping it.** At each bail-out, append to the caller node's
   metadata (say `metadata.unresolved_calls`) one entry per call site:
   `{"callee": "greet", "receiver_type": "Greeter", "line": "L10", "lang": "java"}`. Only when a
   receiver type was actually derived — an untypable receiver stays dropped, so no new guessing.
   Dedup by `(callee, receiver_type)` and cap per node, so growth is proportional to genuinely
   unresolved member calls rather than to corpus size.
2. **Persist per-class receiver tables on the class node** (`metadata.field_types`), so a merge-time
   pass can also type receivers the single-repo run could not type at all. This half is optional and
   can land later — (1) + (3) alone covers the concretely-typed receivers, which is most of them.
3. **A merge-time resolver pass**, modelled on `link_shared_type_declarations` and called beside it at
   `cli.py:2568` (and on the `global add` path): for each node carrying residue, look up
   `receiver_type` among the merged graph's type declarations, keep the **same single-definition
   guard**, then emit `calls` -> the method node contained by that type whose label matches `callee`.
   Add a language-family guard so a Java `Greeter` never binds to a Python one.
4. **Idempotent.** Tag the emitted edges and clear-then-recompute on every merge, so re-merging or
   adding a third repo does not duplicate. Keep the residue on the node after resolving — a repo added
   next month may be the one that closes a still-open entry.
5. **Edges only.** The pass adds edges between existing nodes and never creates or renames nodes, which
   keeps it composable with the prefixing and pruning already in place.

The result is language-neutral, and note it should yield *more* on concretely-typed languages than on
ObjC/Swift, since Java/C#/C++ already stamp `receiver_type` on the vast majority of member calls.

Happy to send a PR — (1) for one language plus (3) as a self-contained pass, so the shape can be
reviewed before it is repeated across the other resolvers.

## Impact

Measured on a private 26-repo Objective-C/Swift codebase (0.9.39, per-repo graphs merged with
`global add`): **3,198** additional cross-repo call edges are recoverable from residue that today is
discarded — 2,007 from explicitly named receivers, 766 from locals typed by a `new`/`alloc` binding,
425 from `@property`/ivar-typed receivers. Against that, 12,866 call sites (29.0% of all `[self.x …]`
sites) have a receiver whose type is outside the unit's corpus, so 3,198 is what survives the
single-definition guard, not the gross figure. One iOS codebase, not a public-corpus number, and iOS is
close to the worst case for this — 27.9% of its receivers are `id<Protocol>`-typed and untypable in
principle, which is a share a Java or C# codebase would not have.

## Note

Two narrower siblings from the same investigation, both independent of this one: #3150 (ObjC
field-table keys go stale in the id remap, so #2591's `self.<field>` typing is inert via the CLI) and
#3151 (field-typed receivers are not inherited from a superclass, Java and ObjC). Fixing those two
widens how many receivers get a `receiver_type` at all; this issue is about not throwing it away once
they do.

xiongjianxu:
Fixes #3152.

## What was wrong

A single-repo build binds `obj.method()` only when the receiver's type is declared in that same build. When the type lives in another repository, the resolver already has the receiver type in hand and drops the call anyway — so nothing about it reaches `graph.json`, the only artifact `merge-graphs` and `global add` consume. No merge-time pass can recover what was never written down.

Two repos, `App.run()` calling `Greeter.greet()` where `Greeter` is declared in the other repo:

```
one corpus  : 8 nodes, 8 edges  (.run() -> .greet() resolved)
two repos   : 8 nodes, 7 edges  (the call edge is simply gone)
```

## What this changes

1. **Park the call at extraction** (`extract.py`): when the receiver type resolves to **zero** declarations in this corpus, the resolver records `{callee, receiver_type, lang, line}` on the caller node under `metadata.unresolved_calls` instead of dropping it silently. The payload carries **names only, never node ids** — ids are rewritten by the id-disambiguation remap and again by repo prefixing, and a stale id inside `metadata` fails silently (#3150 was that bug); names survive every rewrite.

   `> 1` declarations stays dropped: that is a *local* ambiguity, and merging only widens it.

2. **Finish the edge after the merge** (`cross_repo_calls.py`): `link_cross_repo_member_calls(merged)` reads the parked entries back and emits `relation="calls", context="cross_repo", confidence="INFERRED"` when the receiver type resolves to **exactly one** declaration in **another** repo that owns **exactly one** member of that name.

3. **Hooks**: `merge-graphs` (next to the #3007 `same_type_as` pass) and `global_add`.

The pass adds edges only — no node merging, no renaming — so it composes with the prefixing and pruning already in place. Every edge it adds is tagged, and it drops its own previous output before recomputing, which is what makes an incremental `global add` land exactly where a single `merge-graphs` of the same inputs does, and how a repo whose types moved stops answering for calls it no longer owns.

## Languages

Parking is wired into **Java, C++, C# and Swift** — every resolver that already has a receiver type and the same "declared nowhere in this corpus" bail-out. Three details the non-Java ones force:

- **Swift** shares `all_raw_calls` with every other language and its raw_calls carry no `lang` tag, so the parked entry's language comes from the declaring file's suffix rather than from a tag that isn't there.
- **C#** collapses "absent", "ambiguous" and "scoping was decisive" into one `None` from `_resolve_type_name_nid`, and only the first is a cross-repo candidate, so the park re-checks the bare-name index instead of trusting the `None`.
- **C++** models an in-class declaration (`void bar();` in a header) as a field carrying `defines`, not `method`, so the merge-side member index is kept per relation: a C++ entry may fall back to `defines` when no `method` of that name exists, and for every other language a `defines` target is a field and cannot answer a call. When a class has both — header declaration plus out-of-line definition — the definition wins.

`Foo::bar()` in C++ is also the shape of a namespace-qualified free function, and a Pascal-cased C# receiver may be a local rather than a type; in both cases the entry is parked by name and the merge's guards do the deciding, so the worst case is an entry nothing answers.

## Guards

Each has a test, because each is what keeps a name collision from fabricating an edge:

| guard | why |
| --- | --- |
| exactly one candidate declaration | the same single-definition rule the single-repo resolvers apply |
| candidate repo ≠ caller repo | a same-repo declaration means the local resolver already refused; this pass never re-decides inside one repo |
| candidate's `source_file` suffix matches the parked `lang` | without it a Java `Greeter` binds just as happily to a Python class of the same name — or to a C# one |
| declaration must have a `source_file` | a sourceless stub minted for a dangling reference owns nothing |
| exactly one member of that name on the owner | no guessing between overloads' owners |
| `defines` answers for C++ only | elsewhere a `defines` target is a field, and a field cannot be called |
| known `lang` only | an entry from an extractor not in `_LANG_SUFFIXES` cannot be language-checked, so it is not acted on |

## Verified

```
residue in repo-a/graph.json:
  src_appa_app_app_run -> [{"callee":"greet","receiver_type":"Greeter","lang":"java","line":"L10"}]

merge-graphs: "resolved 1 member call(s) across repos"
              Merged 2 graphs -> 8 nodes, 8 edges   (matches the one-corpus control)
              repo-a::…app_run -> repo-b::…greeter_greet  [INFERRED cross_repo]

idempotency : 2nd and 3rd pass each report 1, edge count stays 8
global add  : repo-a -> 0 ; then repo-b -> 1   (incremental agrees with the single merge)
```

`tests/test_cross_repo_member_calls.py` — 19 tests: the happy path, the guards above, run-twice-no-duplicate, a repo that stops declaring the type losing the edge, `merge-graphs` through the real CLI, and an end-to-end per language that extracts two repos, parks in one and merges (Java field receiver, both C++ receiver shapes, C# field receiver, Swift property receiver). The language cases skip when their tree-sitter grammar is absent.

Full suite before and after this branch: **identical failure sets** — 408 failed / 4563 passed on the base commit, 408 failed / 4582 passed here (+19, the new file). The 408 are pre-existing in my environment, mostly tree-sitter grammars I don't have installed (Python, Ruby, …); `test_serve_http.py` is excluded from both runs for an x86_64/arm64 `rpds` mismatch.

## Not included

- **ObjC** needs two extractor-side changes before the merge can answer for it: its `add_node` never emits `_callable_class`, so ObjC classes are invisible to the declaration index, and method labels keep the `+`/`-` sigil, so the merge's name key never matches a parked `doThing`. Filed separately rather than folded in here.
- **Python and TypeScript** would need a source/origin field in the parked entry — Python's class index is derived from `method` edges (a method-less local class looks "absent") and TS parking would bypass #2553's import-origin gate.

### DIFF
--- a/graphify/cli.py
+++ b/graphify/cli.py
@@ -3118,7 +3118,8 @@ def dispatch_command(cmd: str) -> None:
                     print(f"Added '{tag}' to global graph: +{result['nodes_added']} nodes, "
                           f"-{result['nodes_removed']} pruned. Global: {_global_path()}")
                     if result.get("cross_repo_calls"):
-                        print(f"  resolved {result['cross_repo_calls']} member call(s) across repos")
+                        print(f"  resolved {result['cross_repo_calls']} "
+                              f"member call(s) across repos")
             except Exception as exc:
                 print(f"error: {exc}", file=sys.stderr); sys.exit(1)
         elif subcmd == "remove":
--- a/graphify/cli.py
+++ b/graphify/cli.py
@@ -2702,6 +2702,12 @@ def dispatch_command(cmd: str) -> None:
         shared_links = _link_shared(merged)
         if shared_links:
             print(f"  linked {shared_links} type declaration(s) shared across repos")
+        # A member call whose receiver type lives in another repo was dropped at
+        # extraction; the caller node carries it and this finishes the edge (#3152).
+        from graphify.cross_repo_calls import link_cross_repo_member_calls as _link_calls
+        call_links = _link_calls(merged)
+        if call_links:
+            print(f"  resolved {call_links} member call(s) across repos")
         # Drop whatever compose left behind (the last input's list, possibly
         # with internal duplicates) so attach_hyperedges dedups the full
         # collection by id from a clean slate.
@@ -3111,6 +3117,8 @@ def dispatch_command(cmd: str) -> None:
                 else:
                     print(f"Added '{tag}' to global graph: +{result['nodes_added']} nodes, "
                           f"-{result['nodes_removed']} pruned. Global: {_global_path()}")
+                    if result.get("cross_repo_calls"):
+                        print(f"  resolved {result['cross_repo_calls']} member call(s) across repos")
             except Exception as exc:
                 print(f"error: {exc}", file=sys.stderr); sys.exit(1)
         elif subcmd == "remove":
--- a/graphify/cross_repo_calls.py
+++ b/graphify/cross_repo_calls.py
@@ -1,0 +1,184 @@
+"""Finish member calls that cross a repository boundary in a merged graph (#3152).
+
+A single-repo build can only bind ``obj.method()`` when the receiver's type is
+declared in that same build. When the type lives in another repository the
+resolver holds the receiver type and drops the call anyway, so ``graph.json`` —
+the only artifact ``merge-graphs`` and ``global add`` consume — records nothing,
+and no merge-time pass can recover what was never written down. A two-repo call
+graph was therefore missing exactly the edges that make it a call graph: eight
+edges when the code sits in one corpus, seven after merging the same code from
+two repos.
+
+The resolvers now park those calls on the caller node as
+``metadata.unresolved_calls`` entries (names only, no node ids — see
+``_park_unresolved_member_call``). This pass reads them back after the graphs are
+composed and emits the ``calls`` edge when the receiver's type resolves to
+exactly one declaration in another repo, keeping the single-definition guard the
+single-repo resolvers use: an ambiguous name still fabricates nothing.
+
+Edges only, no node merging or renaming, so the pass composes with the prefixing
+and pruning already in place. Every edge it adds is tagged, and it clears its own
+previous output before recomputing: ``global add`` composes one repo at a time
+and revisits the same pairs on every add, and recomputing from the parked entries
+keeps the result identical whether three repos arrived together or one at a time.
+"""
+from __future__ import annotations
+
+import os
+from collections import defaultdict
+from typing import TYPE_CHECKING
+
+if TYPE_CHECKING:  # pragma: no cover - typing only
+    import networkx as nx
+
+CROSS_REPO_CALL_MARKER = "_cross_repo_call"
+UNRESOLVED_CALLS_KEY = "unresolved_calls"
+
+# A parked entry names the language it was written in, and the declaration that
+# answers it must be written in the same one: without this a Java `Greeter` binds
+# just as happily to a Python class of the same name in another repo. Extend this
+# map when another extractor starts parking calls.
+_LANG_SUFFIXES: dict[str, frozenset[str]] = {
+    "java": frozenset({".java"}),
+}
+
+
+def _key(label: object) -> str:
+    """Normalize a node label or a parked name to its bare identifier.
+
+    Type labels are plain (``Greeter``) while method labels carry the extractor's
+    decoration (``.greet()``). Case is preserved: every language that parks calls
+    here is case-sensitive, and folding case would let `greeter` answer for
+    `Greeter`.
+    """
+    return str(label or "").strip().removeprefix(".").removesuffix("()")
+
+
+def _suffix(source_file: object) -> str:
+    return os.path.splitext(str(source_file or ""))[1].lower()
+
+
+def _parked_entries(data: dict) -> list[dict]:
+    metadata = data.get("metadata")
+    if not isinstance(metadata, dict):
+        return []
+    parked = metadata.get(UNRESOLVED_CALLS_KEY)
+    if not isinstance(parked, list):
+        return []
+    return [entry for entry in parked if isinstance(entry, dict)]
+
+
+def _drop_previous_output(merged: "nx.Graph") -> None:
+    """Clear the edges this pass added on an earlier run.
+
+    Recompute-from-scratch is what makes an incremental ``global add`` agree with
+    a single ``merge-graphs`` of the same inputs, and it is also how a repo whose
+    types moved stops answering for calls it no longer owns.
+    """
+    stale = [(u, v) for u, v, data in merged.edges(data=True)
+             if data.get(CROSS_REPO_CALL_MARKER)]
+    merged.remove_edges_from(stale)
+
+
+def _index_declarations(merged: "nx.Graph") -> tuple[dict[str, list[str]], set[str]]:
+    """Index sourced type declarations by bare name. Returns (index, id set)."""
+    by_name: dict[str, list[str]] = defaultdict(list)
+    type_ids: set[str] = set()
+    for node, data in merged.nodes(data=True):
+        if not data.get("_callable_class") or not data.get("source_file"):
+            continue
+        if not data.get("repo"):
+            continue
+        name = _key(data.get("label"))
+        if not name:
+            continue
+        by_name[name].append(node)
+        type_ids.add(node)
+    return by_name, type_ids
+
+
+def _index_methods(merged: "nx.Graph", type_ids: set[str]) -> dict[tuple[str, str], list[str]]:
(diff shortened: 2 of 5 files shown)


xiongjianxu:

xiongjianxu:
Closing: this landed on `v8` as `c604f48`. Verified against the current tip — `graphify/cross_repo_calls.py` and `tests/test_cross_repo_member_calls.py` are byte-identical to this branch, `global_graph.py` calls `link_cross_repo_member_calls` in `global_add`, and `merge-graphs` runs it in `cli.py`. Nothing here is still missing upstream, so the PR has no work left in it.

Keeping #3152 open: the merge-side pass is in, and on `v8` today Java, C++, C# and Swift park receiver-typed member calls for it. Objective-C (#3385), TS/JS (#3387), Kotlin (#3389), PHP (#3391) and Go (#3395) are still open PRs, and each is what makes the pass reachable for that language.

PR #3229, opened 2026-08-31
Mark ObjC declarations with _callable / _callable_class (#3228)

### LINKED ISSUE #3228: ObjC declarations carry no _callable / _callable_class markers, so marker-gated passes skip ObjC entirely
## Summary

The ObjC extractor builds its nodes by hand and never sets `_callable` or `_callable_class`, so **no ObjC declaration is visible to any pass that indexes declarations by those markers**. Every other language gets both from one place in the generic engine.

`graphify/extractors/objc.py`:

```python
def add_node(nid: str, label: str, line: int) -> None:
    if nid not in seen_ids:
        seen_ids.add(nid)
        nodes.append({"id": nid, "label": label, "file_type": "code",
                      "source_file": str_path, "source_location": f"L{line}"})
```

versus `graphify/extractors/engine.py`, which stamps them for every generic-engine language:

```python
for n in nodes:
    if n["id"] in callable_def_nids:
        n["_callable"] = True
        if n["id"] in callable_class_nids:
            n["_callable_class"] = True
```

## Reproduction

```objc
// Greeter.h
@interface Greeter : NSObject
- (void)greet;
@end
```

```
$ python -c "...extract(['Greeter.h'])..."
NODE 'Greeter.h'  _callable=None  _callable_class=None      # file node, correct
NODE 'Greeter'    _callable=None  _callable_class=None      # <- should be a type
NODE 'NSObject'   _callable=None  _callable_class=None      # dangling stub, correct
NODE '-greet'     _callable=None  _callable_class=None      # <- should be callable
```

The same class written in Java:

```
NODE 'Greeting'   _callable=True  _callable_class=True
NODE '.greet()'   _callable=True  _callable_class=None
```

## Why it matters

`_callable_class` + `source_file` is the standard "this is a real type declaration, not a stub" gate. Consumers today:

- `cross_repo_types.py:42` — the `same_type_as` pass from #3007: no ObjC class can ever be a candidate.
- `extract.py:4211` and `extract.py:4320` — the qualified-name declaration indexes.
- `extract.py:6904` — `class_nids`, the #2137 exclusion that keeps a class from being resolved as a callback target. ObjC currently lands on the right side of this one by accident, because it has no `_callable` either.
- `cli.py:3786` / `watch.py:1644` — both carry the markers across an incremental rebuild, so the gap is preserved rather than healed on the next run.

Marker-gated passes are the pattern for anything that needs to distinguish a declaration from a reference, so this is a floor that every new cross-file or merge-time pass hits with ObjC.

## A second, related gap: the method-label sigil

ObjC method labels keep the selector sigil (`objc.py`, `add_node(method_nid, f"{prefix}{method_name}", line)`), so the label is `-greet` / `+shared` where every other extractor emits `.greet()`. Any pass that builds a name key from a label by stripping `.` and `()` — the normalizer shape used by the merge-time passes — gets `-greet` and never matches a plain `greet` from a call site.

The two together are what keep ObjC out of the merge-time cross-repo passes: the marker gap makes the class invisible to the declaration index, and the sigil makes its methods unmatchable by name.

## Scope of the attached PR

The PR pairs with this issue and fixes the **marker** half only: `@interface` / `@implementation` / `@protocol` nodes get `_callable` + `_callable_class`, method nodes get `_callable`, and the file node and dangling-reference stubs stay unmarked. That half stands alone and is a prerequisite for the other.

The sigil half touches name normalization rather than the extractor, and the natural place for it is whichever normalizer is doing the matching, so I left it out of this PR rather than guess at the shape. Happy to follow up with it — say which side you'd rather have it on (extractor label vs. normalizer).

xiongjianxu:
Fixes #3228.

## What was wrong

The ObjC extractor builds its nodes by hand and never set `_callable` or `_callable_class`, so no ObjC declaration was visible to any pass that indexes declarations by those markers. Every generic-engine language gets both in one place (`extractors/engine.py`, `callable_def_nids` / `callable_class_nids`).

```objc
@interface Greeter : NSObject
- (void)greet;
@end
```

```
before                                          after
'Greeter'  _callable=None  _callable_class=None  ->  True / True
'-greet'   _callable=None  _callable_class=None  ->  True / -
'Greeter.h'        (file node, stays unmarked)   ->  unchanged
'NSObject'         (dangling stub, unmarked)     ->  unchanged
```

The Java equivalent already reports `_callable=True, _callable_class=True` for the type and `_callable=True` for the method, so this is extractor parity, not a new concept.

## What this changes

`extractors/objc.py` only:

- `add_node` takes an optional tuple of markers to set.
- `@interface` and `@implementation` class nodes: `_callable` + `_callable_class`.
- `@protocol` nodes: the same, because a protocol is a type declaration and the engine marks a Java/C# interface the same way.
- Method nodes: `_callable`.
- The file node and the stubs `ensure_named_node` mints for dangling references stay unmarked — a stub has no declaration behind it, which is the property the `source_file` half of every declaration gate relies on.

No edges, ids, labels or traversal change, and the markers already survive the id-remap and the incremental rebuild path (`cli.py` / `watch.py` copy them across, `extract.py` deliberately does not pop them, #2438).

## Why it matters

`_callable_class` + `source_file` is the standard "real type declaration, not a stub" gate. Consumers today: `cross_repo_types.py:42` (the `same_type_as` pass from #3007), `extract.py:4211` and `extract.py:4320` (qualified-name declaration indexes), `extract.py:6904` (`class_nids`, the #2137 callback-target exclusion). ObjC was excluded from all of them, and `cli.py:3786` / `watch.py:1644` carried the gap across incremental rebuilds rather than healing it.

## Tests

`tests/test_objc_callable_markers.py` — 7 tests: a class from a header, a class from an implementation with no header in the corpus, a header/impl pair after `_merge_decl_def_classes` folds it (the markers have to be on whichever node survives), instance and class methods being callable but not types, a protocol, and the two kinds of node that must stay unmarked (the file node and a dangling-reference stub).

Full suite before and after: **identical failure sets** — 408 failed / 4563 passed on `v8`, 408 failed / 4570 passed here (+7, the new file). The 408 are pre-existing in my environment, mostly tree-sitter grammars I don't have installed (Python, Ruby, …); `test_serve_http.py` is excluded from both runs for an x86_64/arm64 `rpds` mismatch. `ruff check` clean.

## Not included

The second half of #3228 — ObjC method labels keeping the `+`/`-` selector sigil, where every other extractor emits `.greet()` — is name normalization rather than extraction, so it is not in this diff. Happy to follow up once you say which side should own it.

### DIFF
--- a/graphify/extractors/objc.py
+++ b/graphify/extractors/objc.py
@@ -12,6 +12,14 @@ import re
 # `C++Bridge.h` and `Foo+.h` are left intact.
 _OBJC_STEM_PART = re.compile(r"[A-Za-z_][A-Za-z0-9_]*")
 
+# Declaration markers, matching what the generic engine puts on every other
+# language's definitions (#2438): `_callable` says "a real callable, not a
+# same-named data symbol", and `_callable_class` narrows that to a type, which is
+# callable only through a constructor. Passes that index declarations gate on
+# these, so a node without them is invisible to them.
+_CALLABLE = ("_callable",)
+_CALLABLE_CLASS = ("_callable", "_callable_class")
+
 
 def _objc_category_base_stem(stem: str) -> str:
     """Strip an ObjC category/extension suffix from a file stem (``Foo+Cat`` -> ``Foo``).
@@ -115,11 +123,14 @@ def extract_objc(path: Path) -> dict:
     # same (class, field) tombstones the entry (None) — drop, don't guess.
     objc_field_types: dict[str, dict[str, str | None]] = {}
 
-    def add_node(nid: str, label: str, line: int) -> None:
+    def add_node(nid: str, label: str, line: int, markers: tuple[str, ...] = ()) -> None:
         if nid not in seen_ids:
             seen_ids.add(nid)
-            nodes.append({"id": nid, "label": label, "file_type": "code",
-                          "source_file": str_path, "source_location": f"L{line}"})
+            node = {"id": nid, "label": label, "file_type": "code",
+                    "source_file": str_path, "source_location": f"L{line}"}
+            for marker in markers:
+                node[marker] = True
+            nodes.append(node)
 
     def add_edge(src: str, tgt: str, relation: str, line: int,
                  confidence: str = "EXTRACTED", weight: float = 1.0,
@@ -284,7 +295,7 @@ def extract_objc(path: Path) -> dict:
             # produced fine when the members lived in `Foo.h` (#1556).
             cls_stem = _objc_category_base_stem(stem) if _objc_is_category(node) else stem
             cls_nid = _make_id(cls_stem, name)
-            add_node(cls_nid, name, line)
+            add_node(cls_nid, name, line, _CALLABLE_CLASS)
             add_edge(file_nid, cls_nid, "contains", line)
             # superclass is second identifier after ':'
             colon_seen = False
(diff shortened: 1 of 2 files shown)


xiongjianxu:

PR #3239, opened 2026-08-31
fix(analyze): god_nodes honours exclude_hubs_percentile (#3205)

### LINKED ISSUE #3205: god_nodes() ignores exclude_hubs_percentile — hub suppression only reaches cluster()
\`exclude_hubs_percentile\` (tested on 0.9.51) only affects \`cluster()\`'s community-resolution step. It never reaches \`analyze.god_nodes()\` — that function has no percentile/exclusion parameter at all, confirmed by reading the source directly.

Repro: running god-node ranking at every percentile value on a real ~14k-node repo graph returns identical output every time — utility hubs (e.g. a generic \`main\`/\`Path\`-style symbol) stay in the top ranking positions regardless of the setting.

Expected: \`god_nodes()\` should accept the same hub-exclusion parameter the clustering step already has, so a percentile that suppresses hubs in one place suppresses them in the other.

Happy to share the exact repro script if useful.

abhay-codes07:
Closes #3205.

## The problem

`--exclude-hubs` only ever reached `cluster()`'s community-resolution step. `god_nodes()` had no exclusion parameter at all — the reporter ran the ranking at every percentile value on a ~14k-node graph and got identical output every time, with utility hubs pinned to the top regardless of the setting.

## The change

`god_nodes()` gains `exclude_hubs_percentile` using `cluster()`'s exact threshold computation (degrees sorted ascending, `idx = max(0, int(n·pct/100) − 1)`, anything with degree above `degrees[idx]` is a hub), so the one setting suppresses hubs in the ranking and in community resolution alike. Default `None` keeps the historical ranking byte-identical.

Wired through every surface:

- the `god-nodes` CLI command gains `--exclude-hubs N` (both flag forms; a non-numeric value is a usage error);
- the `extract` and `cluster-only` paths pass the percentile they already parsed for `cluster()`, so `--exclude-hubs` on a build now shapes the report's god-node list too;
- the MCP `god_nodes` tool accepts `exclude_hubs_percentile` in its schema.

## Tests

`tests/test_god_nodes_exclude_hubs.py` — 7 tests on a hub-and-spokes graph (hub label chosen outside `_BUILTIN_NOISE_LABELS` so the test isolates the percentile mechanism): the default ranking is unchanged; the percentile suppresses the hub without emptying the ranking; the exclusion set matches `cluster()`'s formula computed independently; percentile 100 excludes nothing; the signature default stays `None`; and the CLI takes both flag forms with a usage error for garbage. With the fix reverted, 6 of 7 fail. The analyze suites are unchanged (56 passed together); the full suite matches the fresh `v8` (0.9.53) baseline.

### DIFF
--- a/graphify/analyze.py
+++ b/graphify/analyze.py
@@ -106,16 +106,31 @@ def _is_json_key_node(G: nx.Graph, node_id: str) -> bool:
     return label in _JSON_NOISE_LABELS
 
 
-def god_nodes(G: nx.Graph, top_n: int = 10) -> list[dict]:
+def god_nodes(G: nx.Graph, top_n: int = 10,
+              exclude_hubs_percentile: float | None = None) -> list[dict]:
     """Return the top_n most-connected real entities - the core abstractions.
 
     File-level hub nodes are excluded: they accumulate import/contains edges
     mechanically and don't represent meaningful architectural abstractions.
+
+    ``exclude_hubs_percentile`` (0-100) suppresses nodes whose degree exceeds
+    that percentile of the graph's degree distribution, using the same
+    threshold computation ``cluster()`` applies (#3205) - so the one setting
+    suppresses utility hubs in the ranking AND in community resolution,
+    instead of only the latter. ``None`` keeps the historical ranking.
     """
     degree = dict(G.degree())
+    hub_threshold: float | None = None
+    if exclude_hubs_percentile is not None:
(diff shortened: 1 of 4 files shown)


abhay-codes07:
Closing — this landed on v8 via cherry-pick as 6f9b79a (0.9.54 batch), confirmed patch-equivalent with git cherry.

PR #3240, opened 2026-08-31
fix(cpp): type member-call receivers from parameters and class fields (#3215)

### LINKED ISSUE #3215: C++: member calls on parameter and field receivers are dropped — cpp_type_table is built from local declarations only (no _ts_receiver_type_table equivalent)
## Summary

`_resolve_cpp_member_calls` types a member-call receiver from `cpp_type_table`, which is populated **only from local variable declarations**. Function parameters are never entered into it, so any call made through a parameter receiver is silently skipped.

TS/JS have an explicit augmentation pass for exactly this — `_ts_receiver_type_table` (`extractors/engine.py:6008`), whose comment states it adds *"local `new` bindings and type-annotated parameters"*. C++ has no equivalent, so its table gets local declarations only.

This matters disproportionately for C++, where passing state by `const&` / `*` and calling through it is the dominant idiom.

## Reproduction

`thing.h`:
```cpp
#pragma once
namespace NS {
class Thing {
public:
    bool IsOk() const { return true; }
};
}
```

`use.cpp`:
```cpp
#include "thing.h"
using namespace NS;
bool Local()                  { Thing t; return t.IsOk(); }  // local var
bool ParamRef(const Thing& T) { return T.IsOk(); }           // param by ref
bool ParamPtr(Thing* P)       { return P->IsOk(); }          // param by ptr
bool ParamVal(Thing V)        { return V.IsOk(); }           // param by value
```

```python
from pathlib import Path
from graphify.extract import extract, collect_files
r = extract(collect_files(Path('.')), cache_root=Path('.'))
```

## Actual

| receiver form | `calls` edge |
|---|---|
| `Thing t; t.IsOk()` — local variable | emitted (INFERRED) |
| `const Thing& T` — parameter by ref | **none** |
| `Thing* P` — parameter by pointer | **none** |
| `Thing V` — parameter by value | **none** |
| `B.Inner.IsOk()` — chained receiver | **none** |
| `Inner.IsOk()` — member field receiver | **none** |

For contrast, the paths that **do** work: bare intra-class calls (`Own()`), `this->Own()`, and `Foo::bar()` all resolve correctly — so this is specifically about the `cpp_type_table` typing source, not member-call handling in general.

## Expected

A parameter with an explicit type annotation is at least as strong a typing source as a local declaration — the type is written directly in the signature, with no inference required. It should populate `cpp_type_table` and resolve, subject to the same single-definition god-node guard already in place.

The same argument applies to class fields with declared types (`Thing Inner;` then `Inner.IsOk()`), which are also absent from the table. Chained receivers (`B.Inner.IsOk()`) are documented as a deliberate bail and are out of scope here.

## Impact measured on a real codebase

Unreal Engine 5 C++ project, 153 files, `.cpp`/`.h`:

- distinct caller→callee pairs resolvable to an in-repo definition: **1,182**
- free / namespace-qualified: 696
- member-only (never also called as a free function): **486 (41.1%)**
- member-only pairs present in `graph.json`: **0 of 486**

Nearly every receiver in that codebase is a parameter (`ResolveTurn(const FSimState& State, ...)` then `State.IsValidPlayer(...)`) or a chained expression (`State.GalaxyMap.AreAdjacent(...)`), so the local-declaration table never applies.

Downstream, this inflates the "weakly-connected nodes" count in `GRAPH_REPORT.md` (586 in this build) and skews community detection, since intra-module coupling is under-represented by ~41%.

## Notes

I understand the deliberate precision-over-recall stance in the docstring ("a false call edge is worse than a missing one", god-node guard). This report isn't asking to relax that — an explicitly typed parameter carries the same evidence strength as the local declarations already accepted, and would keep the single-definition guard intact.

Related: #1547 (introduced the C++ resolver), #2234 (Rust has no member-call pass at all), #2860 (Python `self.<attr>` receivers), #2561 (Swift attribute-typed properties). Same class of gap, different language.

## Environment

- graphify 0.9.52 (`uv tool install graphifyy`)
- Python 3.10.21, macOS arm64

abhay-codes07:
Closes #3215.

## The problem

`_resolve_cpp_member_calls` types a member-call receiver from `cpp_type_table` — which was populated **only from local variable declarations**. Function parameters never entered it, so a call through `const Thing& T` / `Thing* P` / `Thing V` (the dominant C++ idiom for passing state) was silently skipped while the byte-identical call through a local resolved. Class-field receivers had the same gap. TS/JS have an explicit augmentation pass for exactly this; C++ had none.

There was a second, compounding half the issue's table exposes: even with the parameter in the table, the resolver's **capitalized-receiver arm** treated `T.IsOk()` as an explicit type name (`Foo::bar()` tier) and bailed when no type called `T` exists — and capitalized parameter names (`T`, `V`, `Inner`) are everywhere in C++.

## The change

**Per-file table** (engine): the table now folds in, in C++'s own shadowing order enforced by its first-write-wins —

1. local declarations (unchanged),
2. the enclosing function's parameters (`_cpp_parameter_types` — same precision rules: class-like type nodes only, qualified names keyed by their simple tail),
3. class/struct member declarations (`_cpp_field_types`, lowest precedence).

`_cpp_declarator_name` learns `field_identifier` (fields use it where locals use `identifier`).

**Resolver** (corpus): when the capitalized-receiver arm finds no unique type of that name, it falls back to the file's var→type table exactly like the lowercase arm — mirroring what the C# resolver already does for Pascal-cased locals. A table miss still bails; nothing is ever guessed, and chained receivers (`B.Inner.IsOk()`) stay deferred.

## Tests

`tests/test_cpp_receiver_parameters.py` — 6 tests built on the issue's exact repro: all four receiver forms (`local`, `const&`, `*`, by-value) resolve like the local does; a class-field receiver resolves; a shadowing local can never bind through the parameter's type; builtin-typed parameters contribute nothing; chained receivers emit nothing wrong; and the pre-existing tiers (scoped call, local var) are unchanged. With the fix reverted, 3 of 6 fail. The C/C++ and language suites are unchanged (450 passed, the one exception being the known Windows-only markdown-unicode baseline failure); the full suite matches the fresh `v8` (0.9.53) baseline.

### DIFF
--- a/graphify/extract.py
+++ b/graphify/extract.py
@@ -3821,20 +3821,34 @@ def _resolve_cpp_member_calls(
         elif receiver[:1].isupper():
             # Foo::bar(): the type is named explicitly in source.
             type_defs = type_def_nids.get(_key(receiver), [])
-            if not type_defs:
-                # Declared nowhere here, which in a multi-repo setup usually means
-                # "in a repo this build does not contain" (#3152). `Foo::bar()` is
-                # also the shape of a namespace-qualified free function, so the
-                # merge side's guards do the deciding: it acts only when exactly
-                # one other repo declares a `Foo` owning exactly one `bar`.
-                _park_unresolved_member_call(
-                    node_by_id.get(caller), callee, receiver, "cpp", rc,
-                )
-                continue
-            if len(type_defs) != 1:  # ambiguous -> bail (god-node guard)
-                continue
-            type_nid = type_defs[0]
-            type_qualified = True
+            if len(type_defs) == 1:
+                type_nid = type_defs[0]
+                type_qualified = True
+            else:
+                # A capitalized receiver is very often a VARIABLE - the
+                # parameter `T` in `ParamRef(const Thing& T)`, the field
+                # `Inner` - not a type name (#3215). When no unique type of
+                # that name exists, fall back to the file's var->type table
+                # exactly like the lowercase arm before giving up.
+                type_name = type_table_by_file.get(src_file, {}).get(receiver)
+                if not type_name:
+                    if not type_defs:
+                        # Declared nowhere here (and not a locally-typed
+                        # variable either), which in a multi-repo setup usually
+                        # means "in a repo this build does not contain" (#3152).
+                        # `Foo::bar()` is also the shape of a namespace-qualified
+                        # free function, so the merge side's guards do the
+                        # deciding: it acts only when exactly one other repo
+                        # declares a `Foo` owning exactly one `bar`.
+                        _park_unresolved_member_call(
+                            node_by_id.get(caller), callee, receiver, "cpp", rc,
+                        )
+                    continue  # ambiguous types, no local variable -> bail
+                type_defs = type_def_nids.get(_key(type_name), [])
+                if len(type_defs) != 1:  # ambiguous or absent -> bail
+                    continue
+                type_nid = type_defs[0]
+                type_qualified = False
         else:
             # f.bar() / f->bar(): type the receiver via the file's local table.
             type_name = type_table_by_file.get(src_file, {}).get(receiver)
--- a/graphify/extractors/engine.py
+++ b/graphify/extractors/engine.py
@@ -1629,19 +1629,80 @@ def _cpp_declarator_name(node, source: bytes) -> str | None:
     for anything that isn't a plain named local (arrays, function pointers,
     structured bindings) so the type table never records a guessed receiver."""
     t = node.type
-    if t == "identifier":
+    if t in ("identifier", "field_identifier"):
         return _read_text(node, source)
     if t in ("pointer_declarator", "reference_declarator", "init_declarator"):
         inner = node.child_by_field_name("declarator")
         if inner is None:
             for c in node.children:
-                if c.type in ("identifier", "pointer_declarator",
-                              "reference_declarator"):
+                if c.type in ("identifier", "field_identifier",
+                              "pointer_declarator", "reference_declarator"):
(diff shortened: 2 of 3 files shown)


abhay-codes07:
Thanks — worked through all three, and rebased onto v8 (0.9.59):

**Shadowing test (real, fixed in f263ab2).** Correct catch: the old test declared a *differently-named* local (`T2`), so it never shadowed anything, and `edges == [] or ...` passed vacuously. Rewritten as a parameter `Thing T` shadowed by an inner-block local `Other T`, asserting the call binds to `Other::Ready` — which resolves only if the local wins the fold; a params-first regression leaves `Thing` (no `Ready`) and yields no edge. Decisive now, no empty-set escape.

**File-wide table / field-vs-local ordering (pre-existing conservative design, not a regression).** The `var -> ClassName` table has been file-scoped and first-write-wins since #1547 (`_cpp_local_var_types`); this PR folds parameters and fields into that same table in lookup order (local → parameter → field), documented at the call site. It's deliberately conservative: a same-named receiver in an unrelated scope can bind to the first-seen type, but that only ever costs a *missed* edge, never a false one — the corpus resolver emits an edge solely when the receiver type resolves to exactly one definition AND that type actually declares the method (`method_index[(type_nid, method_key)]`). A wrong-type binding fails the method-exists check and is dropped. Per-body scoping would add marginal recall at the cost of reworking the shared #1547 machinery and the file-keyed resolver lookup, so I've kept this PR scoped to the parameter/field gap it set out to close. Happy to file a follow-up for true per-scope typing if that recall is wanted.

PR #3241, opened 2026-08-31
fix(skill): name graphify affected in every always-on block (#3177)

### LINKED ISSUE #3177: always_on: none of the six templates name `affected`, and it went uncalled in 262 sessions
## Summary

None of the six always-on templates mention `affected`. All six name `query`, `path` and `explain`:

```
graphify/always_on/  (0.9.50)
  agents-md.md            affected=0   "graphify query"=1
  antigravity-rules.md    affected=0   "graphify query"=1
  claude-md.md            affected=0   "graphify query"=1
  gemini-md.md            affected=0   "graphify query"=1
  kiro-steering.md        affected=0   "graphify query"=1
  vscode-instructions.md  affected=0   "graphify query"=1
```

This is a different layer from #2535, which reports `affected` missing from `SKILL.md` and `references/`. The skill is loaded on demand; the always-on block is the text that reaches the agent on **every** session and is what it actually reaches for. Same file family as #2006, which reports the opposite inaccuracy in the same block (it points at a `wiki/` that `export wiki` alone ever writes).

## Field measurement

A ~1k-page Next.js monorepo, graph of 32 MB, graphify installed since 2026-08-11, `claude install` block byte-identical to `always_on/claude-md.md`.

**`affected` was called zero times in 262 sessions.** Counting `tool_use` blocks in agent transcripts, not grepping text — the always-on block contains the literal string `graphify query`, so a text grep reports ~100% of sessions and is worthless here.

It is not that the graph went unused. The opt-in query ledger (`GRAPHIFY_QUERY_LOG_ENABLE=1`), counted per day:

| Day | Traversals | Verbs |
|---|---|---|
| 2026-08-20 … 08-25 | 0 | — |
| 2026-08-26 | 5 | `query` 4, `explain` 1 |
| 2026-08-27 | 9 | `query` 7, `explain` 2 |
| 2026-08-28 | 9 | `query` 5, `explain` 4 |

Only the verbs the block names. 08-26 is the day the `PreToolUse` guard was reinstalled here after a period without it; the guard changed *whether* the graph was consulted, and the block determined *how*. `affected` was never reached for, because nothing ever named it.

## Why it is the notable omission

`affected` answers the question an agent is asked immediately before every edit — what depends on this, what breaks if I change it. The guard covers `search` (Bash/Grep) and `read` (Read/Glob); neither fires on an edit tool, so that moment has no mechanism at all, and the block is the only thing that could point at one.

Measured on the same repo: editing `packages/seo/src/breadcrumbs/buildBreadcrumbTrail.ts`, `affected` returns 197 dependants with file and line in 0.08-0.12s warm (1.3-2.2s cold, dominated by reading graph.json). Reproducing that with grep takes several calls and misses re-exports.

## Suggested change

One clause in each of the six templates, alongside the existing `path` / `explain` sentence. For `claude-md.md`:

```diff
-Use `graphify path "<A>" "<B>"` for relationships and `graphify explain "<concept>"` for focused concepts.
+Use `graphify path "<A>" "<B>"` for relationships, `graphify explain "<concept>"` for focused concepts,
+and `graphify affected "<path-or-symbol>"` before changing something, to list what depends on it.
```

Two notes from using it, in case they shape the wording:

- **Pass a path, not a basename.** This repo has 75 `index.ts`, 72 `page.tsx`, 60 `loading.tsx`; `graphify affected "index.ts"` answers `No unique node match`, while `packages/seo/src/index.ts` returns 157. A bare-basename example in the docs would fail on most files of any app-router codebase. (Related: #2706.)
- **Repo-relative, not absolute.** From a git worktree, the absolute path carries a `.worktrees/<slug>/` segment and matches nothing. Worth naming, since worktrees are where a lot of agent work happens.

Happy to send the six-file PR if the wording above is close to what you want.

## Environment

graphify 0.9.50, uv tool install, macOS 15.5 (Darwin 25.5.0), Claude Code.

Not filing the strict-stamp half of this: #3053 already covers `affected` and `god-nodes` skipping `_touch_query_stamp` and the query ledger, and I can confirm it still reproduces on 0.9.50 — the three `_touch_query_stamp` call sites are `query` (cli.py:1187), `path` (:1565) and `explain` (:1699), and stat'ing the stamp across five invocations shows `affected` and `god-nodes` leaving it untouched.

abhay-codes07:
Closes #3177.

## The problem

The always-on block is the text that reaches the agent on **every** session — and none of the six templates mentioned `affected`. The issue's field measurement makes the cost concrete: on a real deployment, **zero `affected` calls in 262 sessions** (counted from `tool_use` blocks, not text grep), while the verbs the block does name — `query`, `path`, `explain` — were used daily once the graph was consulted at all. Agents reach for what the block names; the blast-radius verb was invisible at the only layer that is always loaded.

## The change

All six templates (CLAUDE.md, AGENTS.md, GEMINI.md, Copilot instructions, Antigravity rules, Kiro steering) now name `graphify affected "<symbol-or-file>"` beside query/path/explain, each in its own template's voice — a rule bullet in the bullet-styled blocks, a sentence in Kiro's paragraph, a wrapped sentence in the Copilot prose. The wording says what the verb uniquely answers: the blast radius (reverse dependencies), which query/path/explain do not show.

The always-on blocks are frozen against a v8 baseline by the `--always-on-roundtrip` validator, so each addition is recorded in `ALWAYS_ON_SANCTIONED_EDITS` as an audited, purely-additive substitution — the guard keeps rejecting any *other* drift. Artifacts regenerated, `expected/` blessed, all five skillgen validators pass.

Two existing tests that pinned the registry to exactly the single #1530 entry are updated faithfully: the #1530 substitution is still asserted first and intact, every constant is asserted to carry exactly one purely-additive #3177 edit (token-prefix additivity, since the Copilot edit rewraps a line), and the byte-compare now applies the full registry rather than one hand-applied pair.

## Tests

`tests/test_always_on_affected.py` — all six blocks exist, every one names `graphify affected` and the blast radius, and the previously-named verbs are untouched. With the change reverted, 6 of 13 fail (the existence and query-verb checks rightly keep passing). `tests/test_skillgen.py` passes in full (78 together); the full suite matches the fresh `v8` (0.9.53) baseline.

### DIFF
--- a/graphify/always_on/agents-md.md
+++ b/graphify/always_on/agents-md.md
@@ -6,6 +6,7 @@ When the user types `/graphify`, use the installed graphify skill or instruction
 
 Rules:
 - For codebase questions, first run `graphify query "<question>"` when graphify-out/graph.json exists. Use `graphify path "<A>" "<B>"` for relationships and `graphify explain "<concept>"` for focused concepts. These return a scoped subgraph, usually much smaller than GRAPH_REPORT.md or raw grep output.
+- Before modifying a symbol or judging a change's impact, run `graphify affected "<symbol-or-file>"` — the blast radius (reverse dependencies), which query/path/explain do not show.
 - Dirty graphify-out/ files are expected after hooks or incremental updates; dirty graph files are not a reason to skip graphify. Only skip graphify if the task is about stale or incorrect graph output, or the user explicitly says not to use it.
 - If graphify-out/wiki/index.md exists, use it for broad navigation instead of raw source browsing.
 - Read graphify-out/GRAPH_REPORT.md only for broad architecture review or when query/path/explain do not surface enough context.
--- a/graphify/always_on/antigravity-rules.md
+++ b/graphify/always_on/antigravity-rules.md
@@ -9,6 +9,7 @@ This project has a graphify knowledge graph at graphify-out/.
 
 Rules:
 - For codebase or architecture questions, when `graphify-out/graph.json` exists, first run `graphify query "<question>"` (CLI) or `query_graph` (MCP). Use `graphify path "<A>" "<B>"` / `shortest_path` for relationships and `graphify explain "<concept>"` / `get_node` for focused concepts. These return a scoped subgraph, usually much smaller than `GRAPH_REPORT.md` or raw grep output.
+- Before modifying a symbol or judging a change's impact, run `graphify affected "<symbol-or-file>"` — the blast radius (reverse dependencies), which query/path/explain do not show.
 - If graphify-out/wiki/index.md exists, navigate it instead of reading raw files
 - Read graphify-out/GRAPH_REPORT.md only for broad architecture review or when query/path/explain do not surface enough context
 - After modifying code files in this session, run `graphify update .` to keep the graph current (AST-only, no API cost)
--- a/graphify/always_on/claude-md.md
+++ b/graphify/always_on/claude-md.md
@@ -4,6 +4,7 @@ This project has a knowledge graph at graphify-out/ with god nodes, community st
 
 Rules:
 - For codebase questions, first run `graphify query "<question>"` when graphify-out/graph.json exists. Use `graphify path "<A>" "<B>"` for relationships and `graphify explain "<concept>"` for focused concepts. These return a scoped subgraph, usually much smaller than GRAPH_REPORT.md or raw grep output.
+- Before modifying a symbol or judging a change's impact, run `graphify affected "<symbol-or-file>"` — the blast radius (reverse dependencies), which query/path/explain do not show.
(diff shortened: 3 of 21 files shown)


abhay-codes07:
Checked this against the actual edits and it's a false positive — the word-prefix assertion holds for all six blocks, kiro-steering included. I printed `n.split()[:len(o.split())] == o.split()` for every block and each is `True`: the #3177 kiro edit is a pure append (" Before modifying a symbol, run `graphify affected ...`" onto the end of the existing paragraph), so the old words are a strict prefix of the new words — the paragraph-vs-bullet shape doesn't change that. The full `test_skillgen.py` suite is green (65 passed). I'd rather not loosen a correct, passing assertion, so no code change here; happy to reconsider if you're guarding against a future kiro edit that inserts mid-text rather than appends.

PR #3243, opened 2026-08-31
fix(build): make definition_file portable, like its sibling source_file (#3223)

### LINKED ISSUE #3223: [bugfix] definition_file keeps the build machine's absolute path while source_file is repo-relative
`definition_file` is stored as the build machine's absolute path, while its sibling `source_file` is repo-relative. Any consumer that reads the graph from a different machine — or a different checkout — gets a path it cannot open, and the graph leaks the build host's directory layout.

Reproduced on **0.9.53** (stock, no local modifications).

## Reproduction

Two files are enough — one C++ declaration/definition pair:

```
repro/
  src/Foo.h
  src/Foo.cpp
```

```cpp
// src/Foo.h
#pragma once

class Foo {
public:
    int Bar(int x);
};
```

```cpp
// src/Foo.cpp
#include "Foo.h"

int Foo::Bar(int x) {
    return x + 1;
}
```

```bash
cd repro && git init -q && git add -A && git commit -qm init
graphify update /absolute/path/to/repro
```

Resulting node in `graphify-out/graph.json` (4 nodes total, 1 carries `definition_file`):

```json
{
  "id": "src_foo_foo_bar",
  "label": "Bar",
  "source_file": "src/Foo.h",
  "source_location": "L5",
  "definition_file": "/absolute/path/to/repro/src/Foo.cpp",
  "definition_location": "L3"
}
```

`source_file` is `src/Foo.h`. `definition_file` is absolute.

## Expected

`definition_file` should be `src/Foo.cpp` — the same form as `source_file`. It names a file inside the scanned tree, so it should be portable for the same reasons `source_file` is.

## Why it matters

The MCP `get_node` tool prints both, so an agent sees:

```
Source:     src/Foo.h L5
Defined in: /home/ci/build/repos/myproject/src/Foo.cpp L3
```

If the graph is built anywhere other than where it is consumed — a build server, a container, a shared graph served over `--transport http` — the second path does not resolve on the client, and the client cannot rewrite it because the prefix is not part of the graph. It also puts the build host's directory layout into a graph that may be shared more widely than the machine that produced it.

We hit this with graphs built on a Linux indexing host and consumed by editors on other machines: 13,682 nodes across three C++ repositories, 100% of them absolute.

## Root cause

`_merge_decl_def_classes` copies the implementation's `source_file` **before** any normalization runs:

- `graphify/extractors/resolution.py:2260` — `keeper["definition_file"] = definition["source_file"]`

Every pass that makes stored paths portable keys on the literal string `"source_file"` and therefore never sees the new attribute:

- `graphify/build.py:975` — `node["source_file"] = _norm_source_file(node["source_file"], _root)`
- `graphify/watch.py:321` — `_relativize_source_files`
- `graphify/watch.py:339` — `_rebase_relative_source_files`

`grep -c definition_file graphify/build.py graphify/watch.py` returns `0` and `0` on 0.9.53.

The field is otherwise inert — resolvers, incremental merge, dedup and the shrink guard do not read it. As far as I can tell the only readers are `graphify/serve.py:1809` (`get_node` output) and `exporters/graphdb.py`, which passes scalar attributes through. So this is an output-correctness bug, not a graph-integrity one.

## Why the existing test does not catch it

`tests/test_languages.py:3594`:

```python
assert str(bar.get("definition_file", "")).endswith("Foo.cpp"), bar
```

An absolute path ends with `Foo.cpp` too, so the assertion passes either way. A full-string comparison against the expected relative path would have caught it. (That test came with the original decl/def change, which I contributed — the gap is mine.)

## Existing patch

PR #3023 has the fix: it relativizes `definition_file` alongside `source_file` in `build_from_json` and introduces a `_PORTABLE_PATH_KEYS` tuple in `watch.py` so both normalizers cover the pair. It is against `v8` and still applies cleanly to 0.9.53 — happy to rebase, split, or hand it over if you would rather write it differently.

abhay-codes07:
Closes #3223.

## The problem

The #2990 decl/def merge stamps `definition_file` from the implementation's *then-absolute* `source_file` — and nothing ever relativized it. `graph.json` shipped `"source_file": "src/Foo.h"` beside `"definition_file": "/home/ci/build/.../src/Foo.cpp"`: a path no other machine or checkout can open, leaking the build host's directory layout into every consumer — including MCP `get_node`'s `Defined in:` line, which is exactly where an agent reads it.

## The change

`definition_file` now takes the same `_norm_source_file` normalization as `source_file`, at both build sites — fresh extraction nodes, and loaded-graph attributes, so a graph written before this fix heals on its next rebuild. The cache portability round-trip relativizes and re-anchors it beside `source_file` in both directions. Out-of-root paths keep their absolute form (separators normalized), exactly as `source_file`'s contract has always been.

## Tests

`tests/test_definition_file_portability.py` — 4 tests: the issue's own `Foo.h`/`Foo.cpp` repro through real extraction yields `definition_file: "src/Foo.cpp"` beside `source_file: "src/Foo.h"`; a prebuilt absolute value is normalized at build; an out-of-root value keeps its absolute form; and the cache relativize/absolutize round-trip keeps it portable both ways. With the fix reverted, 3 of 4 fail. `test_build`, `test_cache`, `test_stat_index_portability` and the C/C++ suites are unchanged (201 passed); the full suite matches the fresh `v8` (0.9.53) baseline.

### DIFF
--- a/graphify/build.py
+++ b/graphify/build.py
@@ -973,6 +973,13 @@ def build_from_json(extraction: dict, *, directed: bool = False, root: str | Pat
                 continue
             if "source_file" in node:
                 node["source_file"] = _norm_source_file(node["source_file"], _root)
+            # definition_file names a file inside the scanned tree exactly like
+            # source_file (the #2990 decl/def merge stamps it from the impl's
+            # source_file BEFORE this normalization runs), so it must be made
+            # portable the same way - it used to ship absolute, leaking the
+            # build host's layout into graph.json and MCP get_node (#3223).
+            if "definition_file" in node:
+                node["definition_file"] = _norm_source_file(node["definition_file"], _root)
         G.add_node(node["id"], **{k: v for k, v in node.items() if k != "id"})
     node_set = set(G.nodes())
 
@@ -1202,6 +1209,10 @@ def build_from_json(extraction: dict, *, directed: bool = False, root: str | Pat
             )
         if "source_file" in attrs:
             attrs["source_file"] = _norm_source_file(attrs["source_file"], _root)
+        if attrs.get("definition_file"):
+            # Same portability rule as source_file (#3223); heals a graph
+            # written before the fix on its next rebuild.
+            attrs["definition_file"] = _norm_source_file(attrs["definition_file"], _root)
         # Drop cross-language phantom edges — the same short names (render, parse,
         # time, ...) recur across language boundaries, so an unresolved target can
         # bind to a same-named node in another language. The extraction spec forbids
--- a/graphify/cache.py
+++ b/graphify/cache.py
@@ -606,31 +606,35 @@ def _relativize_source_files_in(payload: dict, root: Path) -> None:
     # source_file the same way nodes/edges/hyperedges do, so it needs the same
     # portable-path treatment for cache entries to round-trip correctly across
     # machines/checkout directories.
+    # definition_file (#2990) is a path into the scanned tree exactly like
+    # source_file; a cache entry keeping it absolute replayed the build host's
+    # layout on every warm hit (#3223).
     for bucket in ("nodes", "edges", "hyperedges", "raw_calls"):
         for item in payload.get(bucket, []):
             if not isinstance(item, dict):
                 continue
-            source = item.get("source_file")
-            if not source:
-                continue
-            sp = Path(source)
-            if not sp.is_absolute():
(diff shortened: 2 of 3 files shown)


abhay-codes07:
Closing — this landed on v8 via cherry-pick as 3e5944b, with 0295703 extending the same normalization to the watch and direct-extract paths (0.9.54 batch). Confirmed patch-equivalent with git cherry.

PR #3244, opened 2026-08-31
Fix #3051: optional query log rotation via GRAPHIFY_QUERY_LOG_MAX_RECORDS

### LINKED ISSUE #3051: Query log has no rotation: an enabled ~/.cache/graphify-queries.log grows without bound
## The gap

`graphify/querylog.py` is append-only by its whole surface: `log_query` opens the log path in append mode and writes one JSONL record per query/path/explain. The module ships no reader and no rotator, and no env knob bounds it — the only knobs are `GRAPHIFY_QUERY_LOG`, `GRAPHIFY_QUERY_LOG_ENABLE`, `GRAPHIFY_QUERY_LOG_DISABLE`, `GRAPHIFY_QUERY_LOG_RESPONSES`.

#1797 made the log opt-in, which limits who is exposed, but anyone who opts in (the log exists precisely because some deployments want it) gets a file that grows forever — and with `GRAPHIFY_QUERY_LOG_RESPONSES=1` it grows by the full response text per query. On an agent box running dozens of queries a day this is unbounded plaintext growth in `~/.cache` that no retention policy sees.

## Proposed shape (implemented and tested in our fork)

Opt-in, default-off, archive-not-delete:

- `GRAPHIFY_QUERY_LOG_MAX_RECORDS=<n>`: after each append, keep the newest n lines; append the oldest lines beyond the bound to a sibling `<stem>.archive.jsonl` (never delete a record); rewrite the live log atomically (tempfile in the same directory + `os.replace`), re-reading the tail first so records appended mid-rotation survive.
- Lines move raw, so archive + live stays byte-identical to everything ever logged.
- Unset / invalid / non-positive = today's behavior exactly; the whole path lives inside `log_query`'s existing fail-silent `try/except`.

Working branch with 9 tests (29 total in `tests/test_querylog.py`, all passing): https://github.com/TelB-io/graphify/pull/16 — happy to open it as a PR here if the shape is acceptable, or adjust (e.g. a byte bound instead of a record bound, or size-gating the check) to whatever you prefer.

akshitj11:
Fix #3051: optional query log rotation via GRAPHIFY_QUERY_LOG_MAX_RECORDS

Problem
graphify/querylog.py is append-only. Deployments that opt in with
GRAPHIFY_QUERY_LOG_ENABLE or GRAPHIFY_QUERY_LOG can accumulate unbounded
JSONL, especially when GRAPHIFY_QUERY_LOG_RESPONSES stores full responses.

Change
- Parse GRAPHIFY_QUERY_LOG_MAX_RECORDS after each append.
- When set to a positive integer, keep the newest N lines in the live log.
- Move older lines to a sibling archive file (<stem>.archive<suffix>).
- Rewrite the live log with tempfile + os.replace in the log directory.
- Unset, invalid, or non-positive values preserve current unlimited behavior.
- Rotation runs inside log_query's existing fail-silent try/except.

Tests
tests/test_querylog.py: trim/keep order, archive accumulation, invalid env noop,
default unchanged, never raises.

Closes #3051

### DIFF
--- a/graphify/querylog.py
+++ b/graphify/querylog.py
@@ -9,6 +9,8 @@ from datetime import datetime, timezone
 from pathlib import Path
 from typing import Any, Iterator
 
+from graphify.paths import _atomic_replace
+
 _NODES_RE = re.compile(r"(\d+)\s+nodes?\s+found")
 
 
@@ -57,8 +59,12 @@ def _archive_path(path: Path) -> Path:
 
 @contextlib.contextmanager
 def _query_log_lock(path: Path) -> Iterator[None]:
-    # Serialize append+rotate on POSIX (watch.py uses the same flock pattern).
-    # Multi-process rotation on Windows is best-effort when fcntl is unavailable.
+    """Serialize append+rotate on POSIX (watch.py uses the same flock pattern).
+
+    Multi-process rotation on Windows is best-effort when fcntl is unavailable.
+    If the lock file cannot be opened, degrades to unlocked append rather than
+    dropping the log line.
+    """
     try:
         import fcntl
     except ImportError:
@@ -66,7 +72,11 @@ def _query_log_lock(path: Path) -> Iterator[None]:
         return
     lock_path = path.with_name(path.name + ".lock")
     lock_path.parent.mkdir(parents=True, exist_ok=True)
-    fh = open(lock_path, "a+", encoding="utf-8")
+    try:
+        fh = open(lock_path, "a+", encoding="utf-8")
+    except OSError:
+        yield
+        return
     try:
         fcntl.flock(fh.fileno(), fcntl.LOCK_EX)
         yield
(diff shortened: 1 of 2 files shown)


akshitj11:
Addressed rotation advisories:
- append+rotate now run under a POSIX flock on <log>.lock (same pattern as watch.py)
- archive write moved after successful os.replace to avoid duplicate records on retry
- tests added for replace-failure ordering and concurrent append preservation

Windows remains best-effort without flock.

The log_query coupling-delta note is acknowledged: rotation stays inside the existing hub with no API surface change.

akshitj11:
Round 2 hardening:
- lock open failure degrades to unlocked append (no silent drop)
- live rewrite uses paths._atomic_replace (mkstemp, realpath, chmod preserve)
- archive failure restores overflow to live log
- Windows/fcntl best-effort documented (matches watch.py)

PR #3263, opened 2026-09-01
fix(extract): collapse byte-identical duplicate edges at emission (#3251)

### LINKED ISSUE #3251: Python AST: a repeated parameter annotation emits an exact-duplicate edge, flagged as a health warning then silently dropped
## Summary

When a Python function annotates two or more parameters with the same type, the AST extractor emits one `references` edge **per occurrence**. The edges are identical in every field — same `source`, `target`, `relation`, `source_location` and `context` — so `diagnose_extraction` counts one under `exact_duplicate_edges` and the build then drops it. The graph ends up correct, but the run reports a `GRAPH HEALTH WARNING` with no actionable cause.

Version: `graphifyy` 0.9.49, Python 3.14, Windows 11.

## Reproduction

```python
# sample.py
from pathlib import Path


def two_params(a: Path, b: Path) -> None:
    """Two parameters annotated with the same type."""
    print(a, b)


def one_param(a: Path) -> None:
    """Control: a single parameter with the same type."""
    print(a)
```

```python
import collections
from pathlib import Path

r = extract([Path("sample.py")], cache_root=Path("."))
c = collections.Counter(
    (e["source"], e["target"], e["relation"], e.get("source_location"), e.get("context"))
    for e in r["edges"]
)
for k, v in c.items():
    print(v, k)
```

Output:

```
2 ('sample_two_params', 'path', 'references', 'L4', 'parameter_type')
1 ('sample_one_param',  'path', 'references', 'L9', 'parameter_type')
```

## Expected vs actual

Expected: one `references` edge per (source, target, relation, location, context) tuple. A signature that mentions a type twice is still one reference relationship at that location.

Actual: `two_params` yields the edge twice; `one_param` yields it once, confirming the count tracks the number of annotations rather than the relationship.

## Why it matters

On a real corpus this surfaces as an unexplained health warning. In a 323-node / 561-edge build the diagnostic reported:

```
exact_duplicate_edges: 1
directed_same_endpoint_collapsed_edges: 5
GRAPH HEALTH WARNING: ... 5 collapsed (directed) edges - graph may be incomplete/corrupt.
```

The other four collapsed pairs are legitimate `calls` + `references` variants between the same nodes. The exact duplicate is pure noise, and it makes the warning harder to triage — a user cannot tell the benign duplicate from a real integrity problem without reconstructing the extraction by hand.

Note that `def f(root: Path) -> set[Path]:` is **not** the same case: those two edges differ in `context` (`parameter_type` vs `generic_arg`) and are correctly counted as a variant group, not a duplicate.

## Suggested fix

Dedupe on the full identity tuple when emitting annotation edges in the Python AST pass, so repeated annotations of one type in a single signature collapse at the producer rather than at build time.

Possibly related, but distinct: #2791 (collapse under-reported at build) and #2363 (ghost nodes from type annotations).

abhay-codes07:
Closes #3251.

## The problem

`def two_params(a: Path, b: Path)` emits the `(two_params → Path, references, L4, parameter_type)` edge **twice** — one per annotation occurrence. The copies are identical in every field, so `diagnose_extraction` counts one under `exact_duplicate_edges` and build's dedup then drops it: the graph ends up correct, but the run reports a GRAPH HEALTH WARNING with no actionable cause, exactly as the issue describes.

The disease is not Python-specific. Every language block in `_extract_generic` emits per-occurrence, and neither `add_edge` nor the raw appends de-duplicate — C#'s `void Copy(Widget a, Widget b)` double-emits the same way, and `s.AddScoped<IZeta, Box<IZeta>>()` (same type in two generic-argument positions of one call) trips the identical warning.

## The change

The engine's existing clean-edges pass now collapses **byte-identical** edges — same full payload after `json.dumps(sort_keys=True)` — keeping the first occurrence. Anything differing in any field survives untouched: two locations, two contexts at one location (`parameter_type` vs `return_type`), differing metadata. This matches the issue's expected contract ("one references edge per (source, target, relation, location, context) tuple") and only removes what build's dedup was already discarding after the health check had counted it.

One existing test asserted the duplicate: `test_csharp_call_site_generic_args.py` counted two identical IZeta edges from `AddScoped<IZeta, Box<IZeta>>()` to prove both generic-argument positions were walked. Those two occurrences are one relationship at one location (and were themselves tripping the #3251 warning); the assertion now checks the edge exists exactly once plus the `Box` argument link, which still fails against the original call-site bug (which emitted nothing).

## Tests

`tests/test_duplicate_annotation_edges.py` — 4 tests: the issue's exact Python repro yields one edge per relationship; the C# witness for the cross-language claim; `diagnose_extraction` reports `exact_duplicate_edges: 0` on the repro corpus; and a survival control proving different locations and different contexts keep every edge. With the fix reverted, 3 of 4 fail (the survival control rightly keeps passing). The extraction/diagnostics slice is unchanged (871 passed; the one failure, `test_c_include_out_of_root_target_id_is_deterministic_across_checkout_paths`, fails identically on clean `v8` on this Windows machine); the full suite matches a fresh same-version `v8` baseline.

### DIFF
--- a/graphify/extractors/engine.py
+++ b/graphify/extractors/engine.py
@@ -3,6 +3,7 @@ from __future__ import annotations
 
 import hashlib
 import importlib
+import json
 from graphify.extractors.base import _LANGUAGE_BUILTIN_GLOBALS, _file_stem, _make_id, _read_text
 from graphify.ids import normalize_id
 from graphify.extractors.models import LanguageConfig
@@ -5979,9 +5980,23 @@ def _extract_generic(
     # ── Clean edges ───────────────────────────────────────────────────────────
     valid_ids = seen_ids
     clean_edges = []
+    # Byte-identical duplicates collapse to one edge (#3251): a signature that
+    # annotates the same type twice (``def f(a: Path, b: Path)``) is ONE
+    # reference relationship at one location, but the per-occurrence emission
+    # loops above append it once per annotation — in every language block, since
+    # neither add_edge nor the raw appends de-duplicate. The copies carry zero
+    # information (build's dedup drops them anyway) and their only observable
+    # effect is tripping diagnose_extraction's exact_duplicate_edges health
+    # warning. Only edges whose ENTIRE payload is identical collapse; any
+    # differing field (source_location, context, metadata, …) keeps both.
+    _seen_edge_payloads: set[str] = set()
     for edge in edges:
         src, tgt = edge["source"], edge["target"]
         if src in valid_ids and (tgt in valid_ids or edge["relation"] in ("imports", "imports_from", "re_exports")):
+            payload = json.dumps(edge, sort_keys=True, default=str)
+            if payload in _seen_edge_payloads:
+                continue
+            _seen_edge_payloads.add(payload)
             clean_edges.append(edge)
 
     # Ruby mixins were collected during the node walk (before raw_calls existed);
--- a/tests/test_csharp_call_site_generic_args.py
+++ b/tests/test_csharp_call_site_generic_args.py
@@ -213,7 +213,16 @@ def test_call_site_generic_args_appear_in_issue_repro(tmp_path):
         ),
     })
     izeta_refs = [tgt for src, tgt, _ctx in all_di_refs if src == ".Di()" and tgt == "IZeta"]
-    assert len(izeta_refs) >= 2, (
-        "s.AddScoped<IZeta, Box<IZeta>>() must link BOTH the outer IZeta "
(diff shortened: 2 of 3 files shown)


safishamsi:
Shipped in **v0.9.54** — landed on `v8` via cherry-pick with your authorship preserved (so it lands on your contributor graph). Thanks @abhay-codes07 — byte-identical duplicate reference edges are now collapsed at emission. Release: https://github.com/Graphify-Labs/graphify/releases/tag/v0.9.54

PR #3264, opened 2026-09-01
fix(csharp): apply Blazor _Imports.razor usings in the razor type resolver (#3187)

abhay-codes07:
Addresses the remaining `_Imports.razor` half of #3187 (follow-up to #3188).

## The problem

#3188 routed a Razor `@inject` through the C# cross-file type resolver, but only directives written in the page itself count. The Razor compiler applies `_Imports.razor` `@using`/alias directives to every Razor file in the same directory and below — and the standard Blazor template keeps the app's namespaces there — so the issue's repro still dangles on 0.9.53: with `Pages/_Imports.razor` carrying `@using Demo.Services`, `@inject WidgetService _widgets` in `Pages/AlphaPage.razor` produces a sourceless per-file stub while the canonical `Demo.Services.WidgetService` definition sits in the graph.

(The other still-open sub-case named in the #3188 comment — a bare `@inject` with **no** using in scope anywhere — dangles on purpose: that Razor file would not compile either, and the resolver treats it exactly like a bare C# cross-namespace reference. A control test documents that.)

## The change

`CsharpNameResolver` now indexes each `_Imports.razor`'s using/alias entries by the directory they govern, and lookups **for razor-family files only** (`.razor`/`.cshtml`) fold in every governing ancestor's entries — same-directory or above, root `_Imports.razor` governing the whole tree, mirroring the Razor compiler's rule. Namespace usings extend `_scopes_for`; aliases merge into the alias view used by `resolve_type_name`/`resolve_qualified`. `.cs` files never inherit them, sibling directories never leak, and the existing single-candidate soundness gate is unchanged — two same-named types both brought in scope by inherited usings still dangle rather than picking a winner.

## Tests

`tests/test_razor_imports_usings.py` — 7 tests: same-directory and root `_Imports.razor` resolve the issue's `@inject` repro to the canonical definition (stub gone); an alias directive in `_Imports.razor` resolves; and four scoping controls — sibling directories don't leak, `.cs` files don't inherit, ambiguous inherited usings still dangle, and no-using-anywhere still dangles. With the fix reverted, the 3 positive tests fail and the 4 controls keep passing. The razor/csharp suites are unchanged (188 passed); the full suite matches a fresh same-version `v8` baseline.

### DIFF
--- a/graphify/extractors/csharp.py
+++ b/graphify/extractors/csharp.py
@@ -164,6 +164,35 @@ def _metadata(value: object) -> dict:
     return value if isinstance(value, dict) else {}
 
 
+def _razor_imports_dir(source_file: object) -> str | None:
+    """The directory prefix an ``_Imports.razor``'s directives apply to.
+
+    Returns the file's parent directory with normalized separators ("" for the
+    corpus root), or None when the file is not an ``_Imports.razor``.
+    """
+    if not isinstance(source_file, str) or not source_file:
+        return None
+    norm = source_file.replace("\\", "/")
+    parts = norm.rstrip("/").rsplit("/", 1)
+    name = parts[-1]
+    if name.lower() != "_imports.razor":
+        return None
+    return parts[0] if len(parts) == 2 else ""
+
+
+def _razor_dir_applies(imports_dir: str, source_file: str) -> bool:
+    """True when an ``_Imports.razor`` in ``imports_dir`` governs ``source_file``.
+
+    The Razor compiler applies ``_Imports.razor`` directives to every Razor
+    file in the same directory and every subdirectory below it.
+    """
+    norm = source_file.replace("\\", "/")
+    file_dir = norm.rsplit("/", 1)[0] if "/" in norm else ""
+    if not imports_dir:
+        return True  # root _Imports.razor governs the whole tree
+    return file_dir == imports_dir or file_dir.startswith(imports_dir + "/")
+
+
 class CsharpNameResolver:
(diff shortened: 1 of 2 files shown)


theSatvik (commented):
One scoping correction before merge: `_Imports.razor` should apply only to `.razor` component files, not `.cshtml`.

Microsoft's Razor component docs explicitly say that directives in `_Imports.razor` are applied only to Razor files (`.razor`), and the Razor layouts docs distinguish it from `_ViewImports.cshtml`, which governs Razor views/pages: https://learn.microsoft.com/en-us/aspnet/core/blazor/components/?view=aspnetcore-10.0#razor-components and https://learn.microsoft.com/en-us/aspnet/core/mvc/views/layout?view=aspnetcore-10.0#importing-shared-directives

As written, `_is_razor_family()` includes `.cshtml`, so an ancestor `_Imports.razor` can incorrectly resolve a bare `@inject` in a `.cshtml` file. Could this be limited to `.razor` and covered with a negative `.cshtml` test, similar to the existing `.cs` isolation test?

abhay-codes07:
Good catch, thanks — you're right, the docs scope _Imports.razor to Razor components only. Fixed in 2709ef3: the inherited-usings lookup now applies to `.razor` files alone, with a negative test asserting a `.cshtml` next to a governing _Imports.razor stays unresolved (alongside the existing `.cs` isolation test). Also rebased onto v8 (0.9.55). Modeling _ViewImports.cshtml for views/pages would be the symmetric follow-up if there's appetite for it.

PR #3271, opened 2026-09-02
perf(serve): persist the trigram index and add an exact-id fast path

mallyskies:
A cold CLI call spends most of its time rebuilding the trigram index. `_get_trigram_index` memoizes on `G.graph`, which is right for a long-lived server but does nothing for a CLI where every invocation is a fresh process. The index is small and pickles in a fraction of a second, so it can be persisted to disk instead.

Two changes.

**An on-disk trigram cache**, keyed on the graph file's path, mtime and size. Writes go to a temporary file that is renamed into place, so a partially-written file is never read back. Nothing in the cache path raises. If the file is missing, stale or unreadable, the index is rebuilt exactly as it is today.

The cache lives under `XDG_CACHE_HOME` or `LOCALAPPDATA` rather than beside `graph.json` because a large cache file written into the corpus could mistakenly end up part of the corpus itself. `GRAPHIFY_TRIGRAM_CACHE_DIR` overrides the location and `GRAPHIFY_TRIGRAM_CACHE_DISABLE=1` turns the cache off.

The cache filename carries a digest of that key, so a rebuilt graph.json writes a new name rather than replacing the existing file. That matters on Windows, where replacing a destination another process holds open fails with PermissionError WinError 5 while POSIX allows it. Measured on Windows 11 with a reader holding the file open for three seconds, all four concurrent writers failed, and because the cache path swallows exceptions the failure is silent. Giving each generation its own name removes the contention instead of retrying around it. `_prune_trigram_cache` deletes superseded generations on write.

**An exact-id fast path.** `_find_node_tiers` now returns immediately when the query is already a node id. Callers that hold one were paying a full index build to rediscover a node the graph looks up in constant time. The clearest case is the second call after a `find_node_ambiguity` error, which hands the caller an id and invites them to call again with it. A single-entry tier also reports no ambiguity, which is correct, since an id names exactly one node.

NOTE: These numbers come from a corpus roughly 12x your largest published benchmark, ERPNext at 22,620 nodes and 48,710 edges. On our 285k-node, 445k-edge graph that was about 6.7s of a 10.3s call. The saving is proportional rather than absolute, so another graph will benefit more or less than ours. Fortunately, the mechanism costs virtually nothing when the cache misses.

### DIFF
--- a/graphify/cli.py
+++ b/graphify/cli.py
@@ -1590,6 +1590,11 @@ def dispatch_command(cmd: str) -> None:
             G = json_graph.node_link_graph(_raw, edges="links")
         except TypeError:
             G = json_graph.node_link_graph(_raw)
+        # Let serve._get_trigram_index find its on-disk cache. This path does
+        # not go through serve._load_graph, which is where the stash normally
+        # happens, so without this the cache key is unresolvable and every CLI
+        # call rebuilds the index from scratch.
+        G.graph["_graph_path"] = str(gp)
         src_scored = _score_nodes(G, [t.lower() for t in source_label.split()])
         tgt_scored = _score_nodes(G, [t.lower() for t in target_label.split()])
         if not src_scored:
@@ -1725,6 +1730,11 @@ def dispatch_command(cmd: str) -> None:
             G = json_graph.node_link_graph(_raw, edges="links")
         except TypeError:
             G = json_graph.node_link_graph(_raw)
+        # Let serve._get_trigram_index find its on-disk cache. This path does
+        # not go through serve._load_graph, which is where the stash normally
+        # happens, so without this the cache key is unresolvable and every CLI
+        # call rebuilds the index from scratch.
+        G.graph["_graph_path"] = str(gp)
         matches = _find_node(G, label)
         if not matches:
             print(f"No node matching '{label}' found.")
--- a/graphify/serve.py
+++ b/graphify/serve.py
@@ -453,7 +453,7 @@ def _trigram_cache_key(G: nx.Graph):
     return key
 
 
-def _trigram_cache_path(graph_path: str) -> Path:
+def _trigram_cache_path(graph_path: str, key=None) -> Path:
     """Where the persisted index lives.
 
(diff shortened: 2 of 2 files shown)


mallyskies:
Thanks — the three write-path findings from the earlier rounds are fixed in the two follow-up commits on this branch:

- `1bb7a1c` gives each writer its own temp file via `mkstemp` in the destination directory, and moves the cache-path computation inside the guard so `_store_trigram_cache` genuinely never raises. The shared `<dest>.pkl.tmp` was real: I reproduced it with six concurrent writers and five of them failed, because the first rename moves the inode out from under the others.
- `806be6c` pins the cache key to the read the index was built from. It was being computed twice with a separate `stat()` each time, so an index built from old content could be stored under a rebuilt file's key and read back as current.

That leaves the pickle finding, and I'd like to make the case for leaving it as it is rather than switching formats.

**The trust boundary.** The cache is written by graphify into a per-user cache directory, `XDG_CACHE_HOME`/`~/.cache/graphify` or `LOCALAPPDATA` on Windows, mode 0700. To plant a malicious cache file you need write access as that user. At that point you can also rewrite `graph.json` itself, the `graphify` package in site-packages, or the user's shell profile. The cache doesn't change any security boundaries.

I'll concede that `GRAPHIFY_TRIGRAM_CACHE_DIR` lets someone relocate the cache, and pointing it at a shared or world-writable path would change the picture. If you'd like, I can add a note to that variable's docstring, or refuse a directory that is group- or world-writable.

**On switching formats.** I'm open to considering it, but it isn't free: the index is `dict[str, array.array]`, and `json` can't represent that without converting every array to a list and back on each load. That conversion is a real fraction of the rebuild this cache exists to avoid, so I'd want to measure it before trading it in. `msgpack` would preserve the shape better but adds a dependency for a cache that is regenerable by definition.

Worth noting the gate itself reports **PASS** — no health regressions — and the finding is marked "agreed by 2 of 2 members but NOT verified (no proof, no reproducing execution)". I'm just commenting to clarify the changes and answer any questions you might have.

PR #3276, opened 2026-09-02
fix(apex): match qualified and generic method return types (#3217)

### LINKED ISSUE #3217: Apex: method_re drops methods with commas/dots in return type (Map<String, Object>, Database.QueryLocator)
### Summary

The Apex extractor's `method_re` cannot match method signatures whose **return type contains a comma, a space, or a dot**. Every such method is silently dropped from the graph — no node, no edges, no warning.

The return type is matched by a single character class:

```python
method_re = _re.compile(
    rf"^{_ANNOTATION}\s*{_ACCESS}{_MOD}\s*(?:static\s+)?[\w<>\[\]]+\s+(\w+)\s*\([^)]*\)\s*(?:throws\s+\w+\s*)?\{{?",
    _re.IGNORECASE,
)
```

`[\w<>\[\]]+` permits word chars and angle/square brackets only, so:

- `Map<String, Object>` — comma + space → **never matches**
- `Database.QueryLocator` — dot → **never matches**
- `List<Map<String, Id>>` — → **never matches**

These are extremely common in real Apex: every `Database.Batchable` implementation's `start()` returns `Database.QueryLocator`, and `@AuraEnabled` controller methods conventionally return `Map<String, Object>`.

### Affected versions

Present since Apex support landed (7467c1b, 2026-06-07). Verified byte-identical in v0.8.39 and v0.9.53 / HEAD (33362d9), now in `graphify/extractors/apex.py`.

### Minimal repro

`Repro.cls`:

```apex
public with sharing class Repro {
    public static String simpleReturn() { return ''; }
    public static Map<String, Object> commaGeneric() { return null; }
    public Database.QueryLocator dottedReturn(Database.BatchableContext bc) { return null; }
    private static List<Map<String, Id>> nestedGeneric() { return null; }
}
```

```python
from graphify.extract import extract_apex
res = extract_apex(Path("Repro.cls"))
print(sorted(n["label"] for n in res["nodes"] if n["label"].startswith(".")))
```

Observed (v0.8.39 and v0.9.53):

```
['.simpleReturn()']
```

Expected: all four methods.

### Real-world impact

On a 455-file production Salesforce codebase, this dropped **512 method signatures** (~25% of the API surface), including every batch-class `start()` entry point and every `Map<String, Object>` controller method.

### Suggested fix

Replace the return-type token with a base token plus an optional generic group that admits commas/spaces/dots (and one nested level):

```python
_RETTYPE = r"[\w.\[\]]+(?:\s*<[^;{}()=]*>)?"
method_re = _re.compile(
    rf"^{_ANNOTATION}\s*{_ACCESS}{_MOD}\s*(?:static\s+)?{_RETTYPE}\s+(\w+)\s*\([^)]*\)\s*(?:throws\s+\w+\s*)?\{{?",
    _re.IGNORECASE,
)
```

Excluding `;{}()=` inside `<...>` keeps declarations-with-initializers (`Map<Id, X> m = new Map<Id, X>();`) and control-flow lines from matching. Scanned against the 455-file corpus above: the old pattern's match set is **strictly contained** in the new one (0 regressions), and all 512 newly matched lines are genuine method signatures.

Happy to submit a PR with this change plus a regression test.

Soumyadip2003-AI:
Fixes #3217.

## The bug

`method_re` in `graphify/extractors/apex.py` matched a method's return type with a single character class:

```python
rf"^{_ANNOTATION}\s*{_ACCESS}{_MOD}\s*(?:static\s+)?[\w<>\[\]]+\s+(\w+)\s*\([^)]*\)..."
```

`[\w<>\[\]]+` admits word characters and angle/square brackets only. Any signature whose return type carries a **dot, a comma, or a space** therefore never matches, and the method is dropped from the graph entirely — no node, no `method` edge, no warning.

That covers the two most common shapes in real Apex:

- `Database.QueryLocator` — every `Database.Batchable` implementation's `start()`
- `Map<String, Object>` — the `@AuraEnabled` controller convention
- `List<Map<String, Id>>` — nested generics

### Repro

```apex
public with sharing class Repro {
    public static String simpleReturn() { return ''; }
    public static Map<String, Object> commaGeneric() { return null; }
    public Database.QueryLocator dottedReturn(Database.BatchableContext bc) { return null; }
    private static List<Map<String, Id>> nestedGeneric() { return null; }
    global Set<Id> setReturn() { return null; }
}
```

```
before: ['.execute()', '.setReturn()', '.simpleReturn()']
after:  ['.commaGeneric()', '.dottedReturn()', '.execute()', '.nestedGeneric()', '.setReturn()', '.simpleReturn()']
```

## The fix

One `_TYPE` sub-pattern for an Apex type expression, admitting a namespace-qualified name and generic arguments:

```python
_TYPE = r"[\w.<>\[\]]+(?:\s*,\s*[\w.<>\[\]]+)*"
```

The important constraint is that **whitespace is admitted only adjacent to a comma**. A blanket `\s` in the type class would let ordinary statements be read as declarations — `insert new Account(Name = 'x');` would match with return type `insert` and method name `new`... `Account`. Restricting whitespace to comma-adjacent positions keeps every such statement unmatchable, because the type and the method name still have to be separated by whitespace that the type itself cannot swallow.

## Verification

- Extraction of the existing fixtures (`tests/fixtures/sample.cls`, `sample.trigger`) is **byte-identical** before and after — no new nodes, no lost ones.
- Two tests added in `tests/test_languages.py`:
  - `test_apex_method_qualified_and_generic_return_types` — the five signature shapes above.
  - `test_apex_statements_are_not_read_as_methods` — a negative guard asserting that `insert new Account(...)`, `System.assertEquals(1, ids.size())`, `Map<String, Object> m = new Map<String, Object>();`, `results.put('a', compute(x));`, `this.helper(1, 2);` and `Integer a = 1, b = compute();` still produce exactly one method node.
- Full suite: `5299 passed, 12 skipped`. The one failure, `tests/test_labeling.py::test_label_communities_batches_when_over_batch_size`, reproduces identically on a clean checkout of `v8` and is unrelated to this change.

## Out of scope

While building the repro I hit a second, separate bug in the same file that this PR deliberately does not touch: the class heritage clause (`(?:\s+extends\s+(\w+))?(?:\s+implements\s+([\w,\s]+))?`) also stops at a dot, so `class Batch implements Database.Batchable<sObject>, Schedulable` fabricates a node labelled `Database` and loses `Schedulable` entirely. Filed separately.

https://claude.ai/code/session_01UEfivKVYzq1e5E6hJHiuKi

### DIFF
--- a/graphify/extractors/apex.py
+++ b/graphify/extractors/apex.py
@@ -73,8 +73,14 @@ def extract_apex(path: Path) -> dict:
         r"^\s*trigger\s+(\w+)\s+on\s+(\w+)\s*\(",
         _re.IGNORECASE,
     )
+    # An Apex return type is not one bare word: it can be namespace-qualified
+    # (`Database.QueryLocator`) and can carry generic arguments holding commas
+    # and spaces (`Map<String, Object>`, `List<Map<String, Id>>`). Whitespace is
+    # admitted only next to a comma, so a statement such as
+    # `insert new Account(...)` still cannot be read as a declaration (#3217).
+    _TYPE = r"[\w.<>\[\]]+(?:\s*,\s*[\w.<>\[\]]+)*"
     method_re = _re.compile(
-        rf"^{_ANNOTATION}\s*{_ACCESS}{_MOD}\s*(?:static\s+)?[\w<>\[\]]+\s+(\w+)\s*\([^)]*\)\s*(?:throws\s+\w+\s*)?\{{?",
+        rf"^{_ANNOTATION}\s*{_ACCESS}{_MOD}\s*(?:static\s+)?{_TYPE}\s+(\w+)\s*\([^)]*\)\s*(?:throws\s+\w+\s*)?\{{?",
         _re.IGNORECASE,
     )
     annotation_re = _re.compile(r"@(\w+)", _re.IGNORECASE)
--- a/graphify/extractors/apex.py
+++ b/graphify/extractors/apex.py
@@ -75,10 +75,13 @@ def extract_apex(path: Path) -> dict:
     )
     # An Apex return type is not one bare word: it can be namespace-qualified
     # (`Database.QueryLocator`) and can carry generic arguments holding commas
-    # and spaces (`Map<String, Object>`, `List<Map<String, Id>>`). Whitespace is
-    # admitted only next to a comma, so a statement such as
-    # `insert new Account(...)` still cannot be read as a declaration (#3217).
-    _TYPE = r"[\w.<>\[\]]+(?:\s*,\s*[\w.<>\[\]]+)*"
+    # and spaces (`Map<String, Object>`, `List<Map<String, Id>>`). Apex also
+    # permits whitespace around the angle brackets themselves (`Map <String, Id>`,
+    # `List< Account >`), so the type is read as segments joined by the type
+    # punctuators `<`, `>` and `,`, with whitespace allowed only ADJACENT to one
+    # of them - never between two bare words. That is what keeps a statement such
+    # as `insert new Account(...)` from being read as a declaration (#3217).
+    _TYPE = r"[\w.\[\]]+(?:\s*[<>,]\s*[\w.\[\]]*)*"
     method_re = _re.compile(
         rf"^{_ANNOTATION}\s*{_ACCESS}{_MOD}\s*(?:static\s+)?{_TYPE}\s+(\w+)\s*\([^)]*\)\s*(?:throws\s+\w+\s*)?\{{?",
         _re.IGNORECASE,
--- a/tests/test_languages.py
+++ b/tests/test_languages.py
@@ -3399,6 +3399,43 @@ def test_apex_method_extraction():
     assert any("createAccounts" in l for l in labels)
     assert any("deleteOldAccounts" in l for l in labels)
 
+def test_apex_method_qualified_and_generic_return_types(tmp_path):
+    source = tmp_path / "Repro.cls"
+    source.write_text(
+        "public with sharing class Repro {\n"
+        "    public static String simpleReturn() { return ''; }\n"
+        "    public static Map<String, Object> commaGeneric() { return null; }\n"
+        "    public Database.QueryLocator dottedReturn(Database.BatchableContext bc) { return null; }\n"
+        "    private static List<Map<String, Id>> nestedGeneric() { return null; }\n"
+        "    global Set<Id> setReturn() { return null; }\n"
+        "}\n"
+    )
+    result = extract_apex(source)
+    labels = _labels(result)
+    assert ".simpleReturn()" in labels
+    assert ".commaGeneric()" in labels
+    assert ".dottedReturn()" in labels
+    assert ".nestedGeneric()" in labels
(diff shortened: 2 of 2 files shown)


Soumyadip2003-AI:
Thanks — the advisory finding is **correct, and I reproduced it**. Pushed a follow-up commit.

`_TYPE` admitted whitespace only next to a comma, so a return type spaced around the angle brackets themselves — all legal Apex — was still dropped:

```apex
public Map<String, Object>  tight()             { return null; }  // matched
public Map<String , Object> spaceBeforeComma()  { return null; }  // matched (\s*,\s*)
public Map <String, Object> spaceBeforeAngle()  { return null; }  // DROPPED
public List< Account >      spacesInside()      { return null; }  // DROPPED
public List < Map< String, Id > > roomy()       { return null; }  // DROPPED
```

### Fix

Read the type as segments joined by the type punctuators `<`, `>` and `,`, with whitespace admitted only **adjacent to one of them** and never between two bare words:

```python
_TYPE = r"[\w.\[\]]+(?:\s*[<>,]\s*[\w.\[\]]*)*"
```

That separator rule is what still bars an ordinary statement from being read as a declaration: the return type and the method name have to be separated by whitespace the type itself cannot swallow, and `_TYPE` can only cross whitespace by consuming a `<`, `>` or `,`.

### On the widened pattern not re-opening the false-positive risk

The obvious worry is that admitting `<` and `>` as separators lets a comparison operator glue a statement into a fake declaration. It does not, and the negative test grew to cover it — `if (a > b) { doIt(x); }`, `while (i < list.size()) { next(); }`, and a generic cast `String s = (Map<String, Object>) JSON.deserializeUntyped(raw);` all still produce zero method nodes, alongside the existing `insert new Account(...)` / `System.assertEquals(...)` / `Integer a = 1, b = compute();` cases.

Measured over the two spellings, positives found / statements misread as declarations:

| pattern | generic + qualified returns matched | statements misread |
|---|---|---|
| before | 6 / 9 | 0 / 13 |
| after  | 9 / 9 | 0 / 13 |

Extraction of `tests/fixtures/sample.cls` and `sample.trigger` is still unchanged (15 nodes / 18 edges, 3 / 2). Full suite `5299 passed, 12 skipped`; the lone `tests/test_labeling.py::test_label_communities_batches_when_over_batch_size` failure reproduces on a clean `v8` checkout and is unrelated.

---

On the other two sections of the review, for the record: the *formal verification* row is an honest abstention (`path` is annotated `Path`, outside the synthesizable set) rather than a signal, and the *health* line — `extract_apex()`, 17 callers — is the pre-existing fan-in of the extractor dispatch table, not coupling this change introduces; the diff adds no call edge.

One thing I could not act on: the review ends with "1 more finding(s) on lines outside this diff (see the check run)", but that finding does not appear in the check-run output or in the PR's review-comment payload, so I have no way to read it. If it can be surfaced I am happy to address it. While testing I did notice one pre-existing false positive that may be it — `return foo(x);` yields a spurious `.foo()` node, because `_CONTROL_FLOW` is checked against the method name but never against the return type. It reproduces identically on `v8` (`33362d9`), so it is not introduced here; say the word and I will file it separately rather than widen this PR.

PR #3278, opened 2026-09-02
fix(apex): resolve a qualified heritage clause to its tail type (#3277)

### LINKED ISSUE #3277: Apex: class heritage clause stops at a dot — `implements Database.Batchable<sObject>, Schedulable` fabricates a `Database` node and drops `Schedulable`
### Summary

The Apex extractor's class regex matches the heritage clause with patterns that admit no dot:

```python
cls_re = _re.compile(
    rf"^{_ANNOTATION}\s*{_ACCESS}{_SHARING}{_MOD}\s*class\s+(\w+)"
    rf"(?:\s+extends\s+(\w+))?(?:\s+implements\s+([\w,\s]+))?\s*\{{?",
    _re.IGNORECASE,
)
```

`extends (\w+)` and `implements ([\w,\s]+)` both stop at the first `.`, so a namespace-qualified base type is truncated to its namespace. Two consequences, both silent:

1. A node is **fabricated** for the namespace (`Database`, `Outer`) and the heritage edge points at it — a namespace is not a type, and the real base type gets no node at all.
2. Because the capture stops at the dot, it never reaches the comma, so **every remaining interface in the list is dropped**.

This is the same class of gap as #3217 (fixed for method return types in #3276), but in a different regex — that PR deliberately leaves this one alone.

### Affected versions

Present in 0.9.53 / HEAD (`33362d9`), in `graphify/extractors/apex.py`.

### Minimal repro

`Batch.cls`:

```apex
public with sharing class Batch implements Database.Batchable<sObject>, Schedulable {
    public void execute(Database.BatchableContext bc) { }
}
```

`Child.cls`:

```apex
public class Child extends Outer.BaseThing {
    public void go() { }
}
```

```python
from pathlib import Path
from graphify.extract import extract_apex

for f in ("Batch.cls", "Child.cls"):
    r = extract_apex(Path(f))
    print(f, sorted(n["label"] for n in r["nodes"]))
    print("  ", [(e["relation"], e["target"]) for e in r["edges"]
                 if e["relation"] in ("extends", "implements")])
```

Actual:

```
Batch.cls ['.execute()', 'Batch', 'Database', 'Batch.cls']
   [('implements', 'database')]          # 'Database' fabricated; Schedulable lost

Child.cls ['.go()', 'Child', 'Child.cls', 'Outer']
   [('extends', 'outer')]                # 'Outer' fabricated; BaseThing never appears
```

Expected: the heritage target is the qualified type (`Database.Batchable`, `Outer.BaseThing`) — or at minimum its tail type name, matching the treatment Kotlin (#1793) and Scala (#1794) received — and `Schedulable` keeps its own `implements` edge.

### Why it matters

`implements Database.Batchable<sObject>, Schedulable` is the canonical shape of every scheduled batch job in Salesforce, so this is not an edge case in a real Apex org. The fabricated `Database` node is also a mild hub-pollution problem: every batch class in the repo points at the same meaningless node.

### Note on generics in the list

A correct fix needs to split the interface list on top-level commas only, since a generic argument can itself contain one (`Map<String, Object>`); splitting on every comma would shred it. Happy to open a PR if the approach sounds right.

Soumyadip2003-AI:
Fixes #3277.

> **Stacked on #3276.** Both fixes live in `graphify/extractors/apex.py` and share the `_TYPE` pattern that #3276 introduces, so this branch is built on top of it and its commit appears here too. Merging #3276 first reduces this PR to its own single commit; happy to rebase, split, or squash the two into one — whatever suits your queue.

## The bug

The class and interface regexes match the heritage clause with patterns that admit no dot:

```python
rf"(?:\s+extends\s+(\w+))?(?:\s+implements\s+([\w,\s]+))?\s*\{{?"
```

Both stop at the first `.`, with two silent consequences:

1. A node is **fabricated for the namespace** (`Database`, `Outer`) and the heritage edge points at it. A namespace is not a type, and the real base type gets no node at all.
2. Because the capture stops at the dot, it never reaches the comma, so **every remaining interface in the list is dropped**.

`implements Database.Batchable<sObject>, Schedulable` is the canonical shape of a scheduled batch job in Salesforce, so this is not an edge case in a real org — and the fabricated `Database` node collects an `implements` edge from every batch class in the repo, which is mild hub pollution on top of the lost data.

### Repro

```apex
public with sharing class Batch implements Database.Batchable<sObject>, Schedulable { }
public class Child extends Outer.BaseThing { }
```

```
before:
  Batch.cls  nodes ['Batch', 'Database', ...]   implements -> 'Database'    # Schedulable lost
  Child.cls  nodes ['Child', 'Outer', ...]      extends    -> 'Outer'       # BaseThing never appears

after:
  Batch.cls  nodes ['Batch', 'Batchable', 'Schedulable', ...]
             implements -> 'Batchable', implements -> 'Schedulable'
  Child.cls  extends -> 'BaseThing'
```

## The fix

Both clauses reuse the `_TYPE` expression, and each is routed through one `heritage_names()` helper that:

- splits on **top-level commas only**, tracking `<`/`>` depth — so an interface list containing a generic argument (`implements Map<String, Object>, Schedulable`) is not shredded into `Map<String` / `Object>` fragments;
- drops the generic arguments and takes the **tail segment** of each entry.

The tail-name choice follows the treatment Kotlin (#1793) and Scala (#1794) received, and the unqualified-tail convention the C++ qualified-base handling already uses, so the edge can bind to a locally declared type of that name. Interface `extends` goes through the same helper rather than keeping its own comma split.

## Verification

- Extraction of the existing fixtures (`tests/fixtures/sample.cls`, `sample.trigger`) is unchanged — same nodes, same 18 / 2 edges.
- Two tests added in `tests/test_languages.py`:
  - `test_apex_qualified_heritage_uses_tail_type` — a qualified `extends` plus a three-entry `implements` list containing a comma-bearing generic, asserting the three real bases are linked and that `Database`, `Outer`, `sObject` and `Object` are *not* fabricated as nodes.
  - `test_apex_interface_qualified_extends_uses_tail_type` — the same for an interface's `extends` list.
- Full suite: `5301 passed, 12 skipped`. The one failure, `tests/test_labeling.py::test_label_communities_batches_when_over_batch_size`, reproduces identically on a clean checkout of `v8` and is unrelated to this change.

https://claude.ai/code/session_01UEfivKVYzq1e5E6hJHiuKi

### DIFF
--- a/graphify/extractors/apex.py
+++ b/graphify/extractors/apex.py
@@ -54,15 +54,57 @@ def extract_apex(path: Path) -> dict:
     _SHARING = r"(?:\s+(?:with|without|inherited)\s+sharing)?"
     _MOD = r"(?:\s+(?:abstract|virtual|override|static|final|transient|testMethod))?"
     _ANNOTATION = r"(?:\s*@\w+(?:\s*\([^)]*\))?\s*)*"
+    # An Apex type expression is not one bare word: it can be namespace-qualified
+    # (`Database.QueryLocator`) and can carry generic arguments holding commas
+    # and spaces (`Map<String, Object>`, `List<Map<String, Id>>`). Apex also
+    # permits whitespace around the angle brackets themselves (`Map <String, Id>`,
+    # `List< Account >`), so the type is read as segments joined by the type
+    # punctuators `<`, `>` and `,`, with whitespace allowed only ADJACENT to one
+    # of them - never between two bare words. That is what keeps a statement such
+    # as `insert new Account(...)` from being read as a declaration (#3217).
+    # The same shape covers a heritage clause, which is a comma-separated list of
+    # such types (#3277).
+    _TYPE = r"[\w.\[\]]+(?:\s*[<>,]\s*[\w.\[\]]*)*"
+
+    def heritage_names(clause: str) -> list[str]:
+        """Split an Apex heritage clause into the type names it actually names.
+
+        `Database.Batchable<sObject>, Schedulable` -> `["Batchable", "Schedulable"]`.
+        Commas are split at top level only, so a generic argument list
+        (`Map<String, Object>`) is not shredded into fragments, and each entry
+        drops its generic arguments and resolves to its tail segment: a qualified
+        base names a type *in* a namespace, not the namespace itself. This is the
+        same tail-name treatment Kotlin (#1793) and Scala (#1794) already use.
+        """
+        parts: list[str] = []
+        current: list[str] = []
+        depth = 0
+        for ch in clause:
+            if ch == "<":
+                depth += 1
+            elif ch == ">":
+                depth = max(0, depth - 1)
+            elif ch == "," and depth == 0:
+                parts.append("".join(current))
+                current = []
+                continue
+            current.append(ch)
+        parts.append("".join(current))
+        names: list[str] = []
+        for raw in parts:
+            name = raw.split("<", 1)[0].strip().rsplit(".", 1)[-1].strip()
+            if name:
+                names.append(name)
+        return names
 
     cls_re = _re.compile(
         rf"^{_ANNOTATION}\s*{_ACCESS}{_SHARING}{_MOD}\s*class\s+(\w+)"
-        rf"(?:\s+extends\s+(\w+))?(?:\s+implements\s+([\w,\s]+))?\s*\{{?",
+        rf"(?:\s+extends\s+({_TYPE}))?(?:\s+implements\s+({_TYPE}))?\s*\{{?",
         _re.IGNORECASE,
     )
     iface_re = _re.compile(
         rf"^{_ANNOTATION}\s*{_ACCESS}{_SHARING}{_MOD}\s*interface\s+(\w+)"
-        rf"(?:\s+extends\s+([\w,\s]+))?\s*\{{?",
+        rf"(?:\s+extends\s+({_TYPE}))?\s*\{{?",
         _re.IGNORECASE,
     )
     enum_re = _re.compile(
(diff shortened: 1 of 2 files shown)


Soumyadip2003-AI:
Rebased on the updated #3276, which now carries a whitespace-tolerant `_TYPE` (review follow-up there). Re-verified after the rebase: heritage resolution is unchanged — `extends Outer.BaseThing implements Database.Batchable<sObject>, Map<String, Object>, Schedulable` still yields `BaseThing` / `Batchable` / `Map` / `Schedulable` and fabricates no `Outer`, `Database`, `sObject` or `Object` node. Full suite `5301 passed, 12 skipped`, with the same unrelated `test_labeling` failure that reproduces on a clean `v8`.

On the advisory finding — **`insert new Account(...)` is not misread as a method declaration.** The finding is flagged "NOT verified (no proof, no reproducing execution)", and running it does not reproduce:

```apex
public class Claim {
    public void real() {
        insert new Account(Name = 'x');
        insert new List<Account>{ a, b };
        upsert new Contact(LastName = 'y');
        update new Account(Id = i);
    }
}
```

```
methods:   ['.real()']
all nodes: ['.real()', 'Claim', 'Claim.cls', 'insert', 'update', 'upsert']
```

Only `.real()` is extracted; the `insert` / `update` / `upsert` nodes are the extractor's intended DML nodes, which `dml_re` has always emitted. That case is also asserted in the suite, in `test_apex_statements_are_not_read_as_methods`, which pins the whole method set to exactly `{".real()"}` — so a regression here fails CI rather than needing to be noticed by eye.

The mechanism, since "permissive" is the reasonable prior for a widened regex: `_TYPE` can only cross whitespace by consuming a `<`, `>` or `,`, never between two bare words. `insert new Account(` therefore cannot be spanned — `_TYPE` stops at `insert`, the required `\s+` then puts the method-name group on `new`, and `\s*\(` fails because `Account` follows instead of `(`. Backtracking finds no shorter split, since the group must end immediately before the paren. The negative test covers the comparison-operator forms this widening does newly touch (`if (a > b) { doIt(x); }`, `while (i < list.size()) { next(); }`, and a generic cast), all of which produce zero method nodes.

Same note as on #3276: the trailing "1 more finding(s) on lines outside this diff" is not present in the check-run output or the review-comment payload, so I cannot read it — happy to address it if it can be surfaced.

Issue #1303, opened 2026-06-13
Seed selection: tied _EXACT_MATCH_BONUS scores aren't degree/community-diversified (residual to #897)


AlienResidents:
## Context

This is residual to #897. That issue's IDF weighting and gap-ratio
shape in `_pick_seeds` handle the broad seed-relevance case
correctly. The failure that remains is narrow: when multiple nodes
share the **same** `_EXACT_MATCH_BONUS`-tier score (which happens
whenever a generic label like "Configuration" appears on N nodes),
`_pick_seeds` picks the first three in enumeration order — no
tie-breaking on node degree or community. The picked three are
typically clustered in whichever region of the graph happened to
add those nodes first.

## Repro (synthetic, runs against any graph)

```python
import networkx as nx
from graphify.serve import _score_nodes, _pick_seeds

G = nx.Graph()

# "right" community: 2 Configuration nodes, low connectivity (real configs)
G.add_node("right_a", label="Configuration", norm_label="configuration",
           source_file="backend/credentials.py", community=42)
G.add_node("right_b", label="Configuration", norm_label="configuration",
           source_file="backend/credentials.py", community=42)

# "wrong" community: 3 Configuration nodes, high connectivity (page abstractions)
G.add_node("wrong_a", label="Configuration", norm_label="configuration",
           source_file="frontend/settings/page.tsx", community=99)
G.add_node("wrong_b", label="Configuration", norm_label="configuration",
           source_file="frontend/usermap/page.tsx", community=99)
G.add_node("wrong_c", label="Configuration", norm_label="configuration",
           source_file="frontend/sync/page.tsx", community=99)

# wrong_* are highly connected (deg 50); right_* are specific (deg 1)
for i in range(50):
    G.add_node(f"hub_{i}")
    G.add_edge("wrong_a", f"hub_{i}")
    G.add_edge("wrong_b", f"hub_{i}")
    G.add_edge("wrong_c", f"hub_{i}")
G.add_edge("right_a", "right_b")

scored = _score_nodes(G, ["configuration"])
seeds = _pick_seeds(scored)
print(seeds)
```

Output:

```
['wrong_c', 'wrong_b', 'wrong_a']
```

All five candidates score `2319.11` (`_EXACT_MATCH_BONUS *
inverse-document-frequency`, identical for all five). Gap-ratio is
0% so all five pass the threshold. `_pick_seeds` then takes
`max_k=3` in input order — the wrong three.

## Same shape on a real corpus

[...]

PR #3284, opened 2026-09-02
fix(query): rank same-hop nodes by query relevance before degree under the budget

### LINKED ISSUE #897: graphify query: seed scoring returns low-relevance nodes, causing multi-call token cascade

andytsai821201-spec:
## Summary

`query` already scores every node against the question to pick seeds, then throws that ranking away before rendering. Under a tight `--budget`, the renderer's hop-then-degree order lets an unrelated hub outrank a node that actually matched a query term in the same hop layer — so the hub survives and the answer is cut. This threads the existing ranking into `_subgraph_to_text` as a second sort key. ~30 lines, no new scoring pass, no new flag.

## Problem

`_subgraph_to_text` orders non-seed nodes by `(hop distance, -degree, id)` (#BUG2). Given seed `S` with two depth-1 neighbors — `RetryTimeout` (degree 1, matches the term "timeout") and `Logger` (a hub wired to eight leaves, matches nothing) — the query `"CompanySpacingGate timeout"` at `--budget 60` renders:

```
NODE CompanySpacingGate
NODE TimeoutPolicy        <- per-term seed for "timeout"
NODE Logger               <- hub, zero relevance
... (truncated — RetryTimeout cut)
```

`RetryTimeout` is a genuine match that did not win the single per-term seed seat, so it is neither protected as a seed nor ranked above the hub — the exact node the question is about is the one dropped.

## Fix

- `_subgraph_to_text(..., scores: dict[str, float] | None = None)`: sort key becomes `(hop, -score, -degree, id)`. Hop distance stays primary (the #BUG2 intent is preserved); relevance decides within a layer; degree and id keep the tail deterministic.
- `_query_graph_text` passes `{nid: score for score, nid in qs.ranked}` — the ranking `_score_query` already computed for seed selection.

**Byte-identical** when no non-seed node scored: a missing/empty map or all-zero scores fall through to the previous degree order. Seeds still render first and still survive truncation. `path` / `explain` untouched.

## Tests

- `test_subgraph_to_text_query_match_outranks_hub_in_same_hop_layer` — unit: scored node renders before a higher-degree unscored node in the same layer.
- `test_subgraph_to_text_without_scores_keeps_degree_order` — unit: `scores=None`, `{}`, and all-zero produce identical output to today.
- `test_query_graph_text_threads_relevance_scores_into_rendering` — end to end: a second term-matching node that is *not* the per-term seed survives a tight budget ahead of the hub.

Watched all three fail before the change (the e2e failure reproduces the `['CompanySpacingGate', 'TimeoutPolicy', 'Logger']` output above). `tests/test_serve.py`: 149 passed. Full suite: 5,167 passed; the 14 failures present (`test_ollama_retry_cap.py` missing `openai`, `test_skillgen.py` audit baselines) reproduce identically on pristine `v8` in this environment and are unrelated.

## Relation to existing work

Deliberately narrower than #347 / #1856 (which rework retrieval end to end): this is an ordering-only change inside the existing renderer. Complementary to #1303 (degree-blind seed ties) and the resolved #897 (seed scoring) — those concern *which seeds* are chosen; this concerns *what survives the budget* after seeds are fixed.

### DIFF
--- a/graphify/serve.py
+++ b/graphify/serve.py
@@ -989,11 +989,25 @@ def _dfs(G: nx.Graph, start_nodes: list[str], depth: int) -> tuple[set[str], lis
     return visited, edges_seen
 
 
-def _subgraph_to_text(G: nx.Graph, nodes: set[str], edges: list[tuple], token_budget: int = 2000, *, seeds: list[str] | None = None) -> str:
+def _subgraph_to_text(
+    G: nx.Graph,
+    nodes: set[str],
+    edges: list[tuple],
+    token_budget: int = 2000,
+    *,
+    seeds: list[str] | None = None,
+    scores: dict[str, float] | None = None,
+) -> str:
     """Render subgraph as text, cutting at token_budget (approx 3 chars/token).
 
     seeds: exact-match nodes rendered first before the degree-sorted expansion,
     so the queried symbol always appears at the top of the output.
+    scores: per-node relevance to the question (the `_score_query` ranking the
+    query path already computed). Within one hop layer, a node that scored
+    against the query renders before a higher-degree node that scored zero, so
+    a tight budget cuts the incidental hub rather than the answer. Nodes absent
+    from the map score 0.0; an empty or missing map leaves the hop/degree order
+    unchanged.
     """
     char_budget = token_budget * 3
     lines = []
@@ -1025,9 +1039,13 @@ def _subgraph_to_text(G: nx.Graph, nodes: set[str], edges: list[tuple], token_bu
                     dist[nb] = hop
                     nxt.append(nb)
         frontier = nxt
+    # Hop distance stays the primary key (#BUG2); query relevance decides
+    # within a layer so a term match outranks an unrelated hub; degree then
+    # str(n) keep the tail deterministic and byte-identical when no score is set.
+    score_of = scores or {}
     ordered = seed_hits + sorted(
         nodes - seed_set,
-        key=lambda n: (dist.get(n, 1 << 30), -G.degree(n), str(n)),
+        key=lambda n: (dist.get(n, 1 << 30), -score_of.get(n, 0.0), -G.degree(n), str(n)),
     )
     for nid in ordered:
         d = G.nodes[nid]
@@ -1253,7 +1271,13 @@ def _query_graph_text(
     # Pass the seeds so the queried symbol renders first and survives truncation
     # (#BUG2): a branch merge had silently dropped this argument, leaving the
     # seed-first ordering as dead code.
-    return header + _subgraph_to_text(traversal_graph, nodes, edges, token_budget, seeds=start_nodes)
+    # `qs.ranked` already holds every node's relevance to the question; hand it
+    # to the renderer so the budget cut is relevance-aware instead of dropping a
+    # term-matching node behind a same-layer hub.
+    scores = {nid: score for score, nid in qs.ranked}
+    return header + _subgraph_to_text(
+        traversal_graph, nodes, edges, token_budget, seeds=start_nodes, scores=scores
+    )
 
 
 def _find_node_tiers(
--- a/tests/test_serve.py
+++ b/tests/test_serve.py
@@ -1735,3 +1735,65 @@ def test_resolve_single_node_shared_by_get_node_and_get_neighbors():
     nid, err = _resolve_single_node(G, "nonexistent")
     assert nid is None
     assert "No node matching" in err
+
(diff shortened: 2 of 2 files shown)


rek:
Independently hit this on a 51,084-node TypeScript graph and arrived at the same conclusion — the ranking is computed and then discarded before rendering. Adopted your exact sort key, `(hop, -score, -degree, id)`, in #3417 so the two compose rather than compete; if this lands first, mine rebases onto it.

One extra data point that may be useful evidence for the "unrelated hub outranks a real match" case you describe, because it makes the inversion much larger than degree alone.

`_search_tokens` splits on punctuation and `_`, but not camelCase, so `handleTeamConfigWrite` stays **one** token. A natural-language query can therefore only reach it through the substring tier — 1/1000th of an exact match. For `"which worker route validates team config writes"`:

| node | query terms matched | score |
|---|---|---|
| `worker` — a local variable in a service worker | 1 of 6 | **126.0** |
| `handleTeamConfigWrite()` — the answer | 4 of 6 | **15.1** |

The coverage scaling from #1602 penalises a 1-of-6 match by 36×, but the exact/substring tier gap is 1000×, so a lone exact hit on a common word still wins by 8×. In that query the correct file sat at position 12.

So the hub problem your PR fixes has a scoring twin: even after the ordering is right, the score being ordered can be inverted for any identifier written in camelCase — which is most of them in TS/Java/Go/C#. #3417 adds a sub-word tier for that, and a plural fold so "writes" reaches `Write`.

Not a request for anything here — your change is correct on its own terms and I did not want to duplicate it.

rek:
Correction to my note above: I said I had adopted your sort key exactly. That is no longer true, and I would rather flag it than leave it standing.

I have reverted to relevance-primary — `(-relevance, hop, -degree, id)` — in #3417. With hop primary, a node the reader is looking for still renders below a nearer node that matched nothing, which is the specific complaint the change came from. #BUG2's guarantee turned out to be carried by the pinned/seed set rather than by hop order, so nothing depended on hop being first; it survives as the tie-break among nodes the query cannot separate.

So it is now a genuine disagreement about one key, not a composition. Your diagnosis is right either way, and the camelCase measurement stands as evidence for it. If the maintainers prefer hop-primary I will take your key back and keep the rest of #3417 — happy either way, and I did not want you finding this in a diff after I had told you otherwise.

PR #3288, opened 2026-09-02
fix(query): score the `rationale` attribute as its own tier (#2293)

### LINKED ISSUE #2293: query scores labels only, never the `rationale` attribute — so "why" questions cannot reach the node that answers them
**Version:** 0.9.29
**Affects:** `graphify query` (and the MCP `query` tool) on any corpus whose semantic layer follows the extraction spec.

## Summary

`_score_query` matches a question against `norm_label`, `label_tokens`, `source_file` and the node id. It never reads `rationale`.

But `references/extraction-spec.md` deliberately stores the WHY as an **attribute**, not a node:

> For rationale (WHY decisions were made, trade-offs, design intent): store as a `rationale` attribute on the relevant concept node — do NOT create a separate rationale node or fragment node.

So for a "why does X …" question, the rationale prose is frequently the **only** place the question's wording occurs. The node holding the answer is reachable only by someone who already knows the label it hangs on — which is precisely what the asker does not know.

The same text is also absent from `_node_search_text`, so those nodes are not even trigram candidates.

## Measurement

A 5,524-node corpus (TypeScript/Next.js app + 200 design docs), semantic layer built per the spec. Five natural-language "why" questions, each with one ground-truth answer node carrying a `rationale` that states the answer in plain words:

| question | answer node's rank | picked as seed | reached at depth 2 |
|---|---|---|---|
| why is the npm version pinned in CI | 4 / 19 | no | no |
| why is Instagram sharing copy-only instead of a prefilled link | 126 / 156 | no | no |
| why does the sheet drag only start when the list is scrolled to the top | 278 / 405 | no | no |
| why do callers not branch on the payment provider | 5 / 492 | no | no |
| why is the directions button hidden when the geolocation popover opens | 37 / 142 | no | no |

**0 / 5 answers reached**, although all five nodes exist, are connected, and literally contain the answer.

Two of them, to show this is not a corpus problem:

- `showDirections visibility rule` → rationale: *"Hidden when the mini card is dismissed and while the geolocation popover is open, because that popover opens upward into the DirectionsFAB's space."* The question "why is the directions button hidden when the geolocation popover opens" is almost a quotation of it. The label shares only "visibility".
- `PaymentProvider Adapter Seam` → rationale: *"Adapter, not branching: callers never switch on the payment provider…"* against the question "why do callers not branch on the payment provider".

## Reproduction

1. Build any graph whose docs went through semantic extraction.
2. Pick a node with a `rationale`; phrase a question from the rationale text using words absent from its label.
3. `graphify query "<question>"` — the node is not among the seeds and usually not in the traversal.

## Suggested fix

Score `rationale` as its own tier, below the label substring tier and above the source-path tier, and add the same text to `_node_search_text` so the trigram prefilter stays a complete candidate generator. Like the source tier it should not count toward term coverage, so it adds recall without letting a long rationale win back an exact-label tier it did not earn.

PR: nuboxworld-byte:fix/query-scores-rationale — 6 tests, 3 of which fail on the current scorer; no new failures in the suite.

With that change, same corpus and questions:

| | before | after |
|---|---|---|
| rank of the answer | 126, 278, 37, 5, 4 | **2, 14, 3, 4, 4** |
| answer reached | 0 / 5 | 2 / 5 |

## The remaining gate is `_pick_seeds`, not scoring

Worth flagging separately: after the fix the answer ranks 2nd–4th for four of the five questions, yet `_pick_seeds`' gap-based cutoff still drops it in three of them. Scoring was one of two gates. I did not touch seed selection — it is a more opinionated change and probably deserves its own issue, but a fix to scoring alone will look disappointing without it.

## Aside

`_node_search_text`'s docstring states it concatenates "every field `_score_nodes` / `_find_node` match a query against". That contract is what makes the two-part fix safe — and it is also what made the omission easy to miss, since the docstring was accurate about the fields the scorer *did* read.

andytsai821201-spec:
Fixes #2293.

## Summary

The extraction spec stores the WHY of a concept as a `rationale` **attribute** on the node, not as a node. `_score_query` never read it and `_node_search_text` never indexed it, so for a "why does X …" question the node that literally states the answer was neither a trigram candidate nor a seed unless the asker already knew its label. #2293 measured 0 / 5 answers reached on a 5.5k-node corpus where all five answer nodes existed.

This implements the fix the issue proposes, minimally (+~40 lines of code): a rationale scoring tier and the matching search-text field.

## Fix

- **`_RATIONALE_MATCH_BONUS = 0.75`** — below the label substring tier (1.0), above the source-path tier (0.5). A term found in the rationale adds to the node's score and to the per-term singleton that seats seeds, but — exactly like the source tier — **never to term coverage**, so a long rationale adds recall without restoring a coverage-scaled exact tier it did not earn.
- **`_node_rationale_text(data)`** — one normalizer (diacritics folded, lower-cased); tolerates the list an extractor may emit; `""` when absent.
- **`_node_search_text`** — appends that text as a trailing NUL-separated field, only when present, so the trigram prefilter remains a complete candidate generator for the scorer (the docstring's contract) and every existing field position — including the #2467 folded-id slot — is unchanged.

**Nodes without a `rationale` score exactly as before.** No new flag, no LLM, one extra dict lookup per candidate node.

### Before / after (the #2293 shape)

Node `FAB visibility rule` with rationale *"Hidden when the mini card is dismissed and while the geolocation popover is open, because that popover opens upward into the DirectionsFAB's space."*; neighbors `DirectionsFAB`, `GeolocationButton`, `MiniCard`.

`query "why is the directions button hidden when the geolocation popover opens"`

```
before:  Start: ['GeolocationButton', 'DirectionsFAB']            <- answer node absent
after:   Start: ['GeolocationButton', 'DirectionsFAB', 'FAB visibility rule']
```

## Tests (all watched failing first)

- rationale-only match ranks the node; label shares no token with the query
- tier order: label substring > rationale > source path
- rationale hit does **not** count toward coverage (`a` gains a sub-unit nudge over `b`, not ¾ of an exact tier)
- list-valued `rationale` tolerated
- `_node_search_text` carries the field; without a rationale the 5-field layout is unchanged
- end to end: the "why" question above seats the rationale node as a seed

`tests/test_serve.py`: 152 passed. Full suite: 5,170 passed; the 15 failures present (`test_ollama_retry_cap.py` missing `openai`, `test_skillgen.py` audit baselines) reproduce identically on pristine `v8` in this environment.

## Relation to existing work

- **Design credit to @nuboxworld-byte** — the tier placement, the no-coverage rule and the search-text requirement are all from #2293's "Suggested fix". Their PR #2294 implements the same idea, but its diff carries the whole repository (+252,795 / −5,138 across 766 files) from a base mismatch and cannot be reviewed or merged as-is; this is a fresh minimal implementation against `v8`. Happy to close this in favour of #2294 if it gets rebased.
- **#2975** (downweight rationale *prefix* seeds) concerns nodes whose *label* is docstring prose. This tier is substring-level at 0.75× — well below the 100× prefix bonus that PR targets — so it cannot create that dominance; the two compose.
- **Out of scope**, as #2293 itself notes: `_pick_seeds`' gap cutoff can still drop a rationale-scored node that ranks 2nd–4th. That is a separate, more opinionated change.

### DIFF
--- a/graphify/serve.py
+++ b/graphify/serve.py
@@ -295,6 +295,14 @@ _EXACT_MATCH_BONUS = 1000.0
 _PREFIX_MATCH_BONUS = 100.0
 _SUBSTRING_MATCH_BONUS = 1.0
 _SOURCE_MATCH_BONUS = 0.5
+# The extraction spec stores the WHY of a concept as a `rationale` attribute
+# on the node, not as a node of its own, so for a "why does X …" question that
+# prose is often the only place the question's words occur (#2293). Score it
+# as its own tier: below a label substring hit (the label still names the
+# thing), above a source-path hit, and — like the source tier — never counted
+# toward term coverage, so a long rationale adds recall without winning back
+# an exact-label tier it did not earn.
+_RATIONALE_MATCH_BONUS = 0.75
 
 
 def _compute_idf(G: nx.Graph, terms: list[str]) -> dict[str, float]:
@@ -329,9 +337,27 @@ def _trigrams(text: str) -> set[str]:
     return {text[i:i + 3] for i in range(len(text) - 2)}
 
 
+def _node_rationale_text(data: dict) -> str:
+    """The node's `rationale` attribute normalized like a label (diacritics
+    folded, lower-cased) for substring matching. Semantic cleanup writes it as
+    one string (several sources joined with blank lines); an extractor may hand
+    over a list — join it. Missing or empty -> "" so callers can `if rationale`.
+    """
+    raw = data.get("rationale")
+    if not raw:
+        return ""
+    if isinstance(raw, (list, tuple)):
+        raw = " ".join(str(part) for part in raw if part)
+    return _strip_diacritics(str(raw)).lower()
+
+
 def _node_search_text(data: dict, nid: str) -> str:
     """Concatenate every field _score_nodes / _find_node match a query against, so
     one trigram index over this text is a complete candidate generator for both.
+
+    - `rationale` (normalized via `_node_rationale_text`) feeds _score_nodes'
+      rationale tier (#2293); appended last, and only when present, so every
+      other field position is unchanged.
 
     - `norm_label` and `source_file` feed _score_nodes' per-term substring tiers.
     - `label_tokens` (the space-joined token form) feeds _find_node's
@@ -363,6 +389,9 @@ def _node_search_text(data: dict, nid: str) -> str:
         nid_folded = _strip_diacritics(str(nid)).lower()
         if nid_folded != nid_text:
             fields += (nid_folded,)
+    rationale = _node_rationale_text(data)
+    if rationale:
+        fields += (rationale,)
     return "\x00".join(fields)
 
 
@@ -539,6 +568,7 @@ def _score_query(
         # driver".
         label_tokens = " ".join(_search_tokens(data.get("label") or ""))
         source = (data.get("source_file") or "").lower()
+        rationale = _node_rationale_text(data)
         # `nid_lower` is needed both by the full-query tier (`if joined`) and by
         # the per-token singleton tier (joined-singlet exact-match check). When
         # neither runs (`joined` empty AND not collecting seeds) skip the call;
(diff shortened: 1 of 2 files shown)


safishamsi:
Shipped in **v0.9.54** — landed on `v8` via cherry-pick with your authorship preserved (so it lands on your contributor graph). Thanks @andytsai821201-spec — the query scorer now ranks a rationale match as its own tier, between exact-label and source-path. Release: https://github.com/Graphify-Labs/graphify/releases/tag/v0.9.54

PR #3299, opened 2026-09-02
fix(watch): preserve pending semantic update flag

### LINKED ISSUE #3294: Code-only `_rebuild_code` unconditionally deletes the `needs_update` doc-staleness flag, silently cancelling every pending-docs signal
## Summary

`graphify.watch._rebuild_code()` — the **code-only, no-LLM** rebuild that the installed `post-commit` hook, the `post-checkout` hook, `graphify update`, and the `graphify watch` loop all call — deletes `graphify-out/needs_update` on every exit path, with no check on whether the pass that ran was semantic.

`needs_update` is a **semantic-scope** flag. It is written only by `_notify_only()`, whose own message is *"Non-code files changed - semantic re-extraction requires LLM"*. It is read by `check_update()`, which tells the user *"Run `/graphify --update` to apply semantic re-extraction."* An AST-only rebuild cannot satisfy that condition, so clearing the flag asserts something the pass did not do.

The result is silent: the flag disappears, `graphify check-update` goes quiet, and the graph keeps answering from the **previous** version of those documents with a real-looking `source_location` citation attached.

**This is the half of #2015 that never shipped.** #2015 asked for exactly this change ("drop the `needs_update` unlink from `_rebuild_code`") and was closed as *"Fixed in v0.9.21"* with only the manifest half addressed. The unlink is still present, unchanged, in **0.9.53**.

**Version:** graphifyy 0.9.53 (`uv tool install`), Python 3.13.12, Windows 11.

---

## The code

All three sites are inside `_rebuild_code` (`def` at `watch.py:1305`; the next top-level `def` is `check_update` at `watch.py:2108`).

**Site 1 — `--no-cluster` path, `graphify/watch.py:1849-1852`:**

```python
            # clear stale needs_update flag if present
            flag = out / "needs_update"
            if flag.exists():
                flag.unlink()

            if same_graph:
                print("[graphify watch] No code-graph changes detected (--no-cluster); outputs left untouched.")
```

**Site 2 — unchanged-topology early return, `graphify/watch.py:1897-1899`:**

```python
                flag = out / "needs_update"
                if flag.exists():
                    flag.unlink()
                html_action = _reconcile_graph_html(out, existing_graph_data)
                ...
                else:
                    print("[graphify watch] No code-graph topology changes detected; outputs left untouched.")
                return True
```

**Site 3 — main clustered rebuild, `graphify/watch.py:2089-2092`:**

```python
        # clear stale needs_update flag if present
        flag = out / "needs_update"
        if flag.exists():
            flag.unlink()
```

Sites 1 and 2 are the sharp ones: they clear the flag on paths whose own log line is **"outputs left untouched."** graphify writes nothing, then declares the doc backlog resolved.

## The other side of the contract

```
graphify/watch.py:2123  def _notify_only(watch_path: Path) -> None:
                            """Write a flag file and print a notification ..."""
                            flag = watch_path / _GRAPHIFY_OUT / "needs_update"
                            ...
                            print("[graphify watch] Non-code files changed - semantic re-extraction requires LLM.")

graphify/watch.py:2108  def check_update(watch_path: Path) -> bool:
                            """Check for pending semantic update flag ...
                            Non-code file changes (docs, papers, images) require LLM-backed
                            re-extraction via `/graphify --update` ..."""
```

The only writer says *LLM required*. The only reader says *LLM required*. The clearer needs no LLM.

## Who triggers it

| Caller | Site |
|---|---|
| stock `post-commit` hook (detached rebuild) | `graphify/hooks.py:171` — `_rebuild_code(_root, changed_paths=changed, force=_force)` |
| stock `post-checkout` hook | `graphify/hooks.py:220` — `_rebuild_code(_root, force=_force)` |
| `graphify update` CLI | `graphify/cli.py:2427` — `_rebuild_code(watch_path, ..., block_on_lock=True)` |
| `graphify watch` loop | `graphify/watch.py:2261` — `_rebuild_code(watch_path)` |

**No git hook is needed to hit this.** The `watch` loop reaches it on its own (`graphify/watch.py:2260-2263`):

```python
                if _batch_triggers_rebuild(batch):
                    _rebuild_code(watch_path)
                if _batch_needs_llm_flag(batch):
                    _notify_only(watch_path)
```

Batch 1 is docs-only → `_batch_needs_llm_flag` is True → flag raised. Batch 2 is code-only → `_batch_triggers_rebuild` is True, `_batch_needs_llm_flag` is False → `_rebuild_code` deletes the flag raised by batch 1. The ordering inside a *single* batch happens to save the mixed case; it does nothing across batches.

The hook path is worse because the rebuild is **detached and asynchronous** — the flag vanishes seconds after the commit that raised it, with no output attached to the terminal that could tell anyone.

## Reproduction (verified on 0.9.53)

```
$ cd <repo with an existing graphify-out>
$ printf 'docs/PRD.md\n' > graphify-out/needs_update
$ graphify update .
Re-extracting code files in . (no LLM needed)...
  AST extraction: 7/7 uncached files (100%)
[graphify watch] No code-graph topology changes detected; outputs left untouched.
Code graph updated. For doc/paper/image changes run /graphify --update in your AI assistant.

$ ls graphify-out/needs_update
ls: cannot access 'graphify-out/needs_update': No such file or directory

$ graphify check-update .
$ echo $?
0
```

`check-update` prints **nothing** — no pending doc changes. Re-raising the flag proves the check itself is fine; only the clearing is wrong:

```
$ printf 'docs/PRD.md\n' > graphify-out/needs_update
$ graphify check-update .
[graphify check-update] Pending non-code changes in <repo>.
[graphify check-update] Run `/graphify --update` to apply semantic re-extraction.
```

And the pass that cleared the flag wrote nothing at all — mtimes are identical before and after the run:

```
-rw-r--r-- 102813 2026-09-02 11:56:40 graphify-out/GRAPH_REPORT.md
-rw-r--r-- 7032843 2026-09-02 11:56:40 graphify-out/graph.json
-rw-r--r-- 52730 2026-09-02 11:56:40 graphify-out/manifest.json
```

(run performed at 12:06)

## Field observation

On a repo using the stock `post-commit` hook, a commit touching **five** documents raised the flag; the flag was gone before the next command ran, and the health check reported "no pending doc changes." The knowledge graph then described the pre-commit version of those five documents, and nothing announced it. The minimal repro above is the same mechanism with the asynchrony removed.

## Prior art in this tracker

- **#2015** (closed 2026-07-20, *"Fixed in v0.9.21"*) already documented this in its body, quoting the same three lines and asking to *"drop the `needs_update` unlink from `_rebuild_code` (a code-only rebuild hasn't cleared the doc backlog, so it shouldn't clear the flag)."* The fix that shipped addressed only the manifest-stamping half. **32 patch releases later the unlink is unchanged.** This issue exists to track the remaining half on its own so it does not get closed alongside a manifest fix again.
- **#2251** (closed) — `_rebuild_code` unconditionally overwrites `graph.json`, destroying semantic nodes. Same class: a code-only pass asserting semantic-scope state.
- **#2336** (closed) — `_rebuild_code` deletes document AST nodes without regenerating them.
- **#2440 / #2677 / #2782** (open) are the *opposite* concern — the skill's cleanup step targets `.needs_update` (dotted) and so never clears the real flag. Worth flagging for whoever picks those up: **#2440 currently cites this unlink as the correct behavior** (*"`graphify update` does clear it (`watch.py`), so the CLI path worked"*). If this issue is right, that premise is wrong, and "fixing" the skill's cleanup to also clear the flag after a code-only rebuild would widen the bug rather than close it.

## Proposed fix

Clear `needs_update` only when the pass that ran actually performed semantic extraction. Two workable shapes:

1. **Simplest — never clear it from the code-only path.** Delete the three unlink blocks from `_rebuild_code` and let the supervised `--update` / semantic flow own the flag's lifecycle. This is what #2015 asked for.

2. **Explicit contract.** Give `_rebuild_code` a `clears_doc_flag: bool = False` parameter (or gate on a `semantic=True` argument) and have exactly one caller — the LLM-backed update flow — pass it. Every hook and CLI caller keeps the default and leaves the flag alone.

Either way, the flag should be cleared by whatever *resolved* the condition, not by whatever ran next.

Worth auditing alongside: `_rebuild_code` also touches `manifest.json` on these same paths, and a stamped-but-never-extracted document is the failure mode #2015 was closed for.

## Workaround available today

Write a **second sentinel that graphify does not manage**, next to the flag graphify clears, and have your own tooling read yours:

```sh
# in post-commit, alongside whatever raises needs_update
printf '%s\n' "$_CHANGED_DOCS" >> graphify-out/needs_update        # graphify will delete this
printf '%s\n' "$_CHANGED_DOCS" >> graphify-out/.gfy_docs_stale     # nothing else clears this
```

Your health check reads `.gfy_docs_stale`; your semantic-extraction script truncates it on success. Confirmed working — the sentinel survived the repro above byte-for-byte while `needs_update` was deleted. It is a workaround, not a fix: it costs every downstream consumer a second file to know about, and users who rely on `graphify check-update` alone still get the wrong answer.

theSatvik:
Closes #3294.

## Problem

`_rebuild_code()` is an AST-only operation, but each of its three successful exit paths deleted `graphify-out/needs_update`. That flag records pending semantic re-extraction for changed documents, so a later code rebuild could silently mark unrelated document work as complete even though no LLM-backed extraction ran.

## Change

- Stop deleting `needs_update` from the no-cluster, unchanged-topology, and clustered rebuild paths.
- Add a parameterized regression test covering all three paths.
- Preserve the flag contents as well as its existence, so the code-only pass cannot rewrite pending semantic state.

The semantic workflow remains responsible for clearing the flag after it actually resolves the pending work.

## Verification

- Regression test before fix: 3 failed (flag deleted on all paths).
- `uv run --frozen pytest tests/test_watch.py -q --tb=short`: 145 passed, 3 skipped.
- `uv run --frozen pytest tests/ -q --tb=short`: 5302 passed, 11 skipped.
- All five `tools.skillgen` CI guards passed.
- `uv run --frozen ruff check graphify/watch.py tests/test_watch.py` passed.
- `uv run --frozen graphify update .` completed successfully.

### DIFF
--- a/graphify/watch.py
+++ b/graphify/watch.py
@@ -1846,11 +1846,6 @@ def _rebuild_code(
             except Exception:
                 pass
 
-            # clear stale needs_update flag if present
-            flag = out / "needs_update"
-            if flag.exists():
-                flag.unlink()
-
             if same_graph:
                 print("[graphify watch] No code-graph changes detected (--no-cluster); outputs left untouched.")
             else:
@@ -1894,9 +1889,6 @@ def _rebuild_code(
                     )
                 except Exception:
                     pass
-                flag = out / "needs_update"
-                if flag.exists():
-                    flag.unlink()
                 html_action = _reconcile_graph_html(out, existing_graph_data)
                 if html_action == "rendered":
                     print(
@@ -2085,11 +2077,6 @@ def _rebuild_code(
                     )
             except Exception as cf_err:
                 print(f"[graphify watch] callflow HTML update skipped: {cf_err}")
-
-        # clear stale needs_update flag if present
-        flag = out / "needs_update"
-        if flag.exists():
-            flag.unlink()
 
         if not no_change:
             print(f"[graphify watch] Rebuilt: {G.number_of_nodes()} nodes, "
--- a/tests/test_watch.py
+++ b/tests/test_watch.py
@@ -163,6 +163,32 @@ def test_check_update_does_not_clear_flag(tmp_path):
     assert flag.exists()
 
 
+@pytest.mark.parametrize(
+    ("no_cluster", "change_topology"),
+    [(True, False), (False, False), (False, True)],
+    ids=["no-cluster", "unchanged-topology", "clustered-rebuild"],
+)
+def test_code_rebuild_preserves_semantic_update_flag(
+    tmp_path, no_cluster, change_topology
+):
+    """An AST-only rebuild cannot clear pending semantic work (#3294)."""
+    source = tmp_path / "app.py"
+    source.write_text("def before(): pass\n", encoding="utf-8")
+    assert _rebuild_code(
+        tmp_path, no_cluster=no_cluster, acquire_lock=False
+    ) is True
+
+    flag = tmp_path / "graphify-out" / "needs_update"
+    flag.write_text("docs/PRD.md\n", encoding="utf-8")
+    if change_topology:
+        source.write_text("def after(): pass\n", encoding="utf-8")
+
+    assert _rebuild_code(
+        tmp_path, no_cluster=no_cluster, acquire_lock=False
+    ) is True
+    assert flag.read_text(encoding="utf-8") == "docs/PRD.md\n"
+
+
 def test_watch_raises_without_watchdog(tmp_path, monkeypatch):
     import builtins
     real_import = builtins.__import__


safishamsi:
Shipped in **v0.9.55** — landed on `v8` via cherry-pick with your authorship preserved. Thanks @theSatvik — a code-only watch rebuild no longer clears the pending semantic-update flag. Release: https://github.com/Graphify-Labs/graphify/releases/tag/v0.9.55

PR #3303, opened 2026-09-03
fix(install): write .graphify_version atomically via os.replace

### LINKED ISSUE #3286: .graphify_version is written non-atomically, unlike SKILL.md and references/
In `_copy_skill_file` (`graphify/install.py`), the two main artifacts are installed via an atomic temp + rename, with comments stating this is deliberate for crash-safety:

- `SKILL.md` — `shutil.copy` to `.tmp`, then `os.replace` (:230-231)
  > `# SKILL.md last (crash-safety), via an atomic temp + rename.`
- `references/` — `copytree` to `references.tmp`, `rmtree` old, `os.replace` (:157-167)
  > `# The rename is atomic on the same filesystem, so an interrupted install never leaves a half-written references/ visible to the agent.`

But the version stamp on the very next line is a bare write:

```python
(skill_dst.parent / ".graphify_version").write_text(__version__, encoding="utf-8")   # :239
```

Two consequences:

1. **Breaks the stated invariant.** An install interrupted at :239 leaves a truncated/empty `.graphify_version` — the exact half-written state the other two paths take care to avoid.

2. **Inconsistent symlink behaviour.** Where the skill dir is managed by a dotfile manager (GNU Stow, chezmoi), `os.replace` *replaces* the symlinks for `SKILL.md` and `references/*`, while `write_text` *follows* the `.graphify_version` symlink and writes through to the source repo. One install run both clobbers and writes-through, so repair is non-obvious.

Using the same temp + `os.replace` for the stamp fixes both. Separately, it may be worth detecting a symlinked `skill_dst` and warning, since silently replacing a managed symlink is surprising.

Seen on graphifyy 0.9.53, `graphify install --platform claude`, Python 3.14.

drmikecrypto:
Match the SKILL.md / references/ crash-safety pattern so an interrupted install cannot leave a truncated stamp, and replace managed symlinks instead of writing through them.

Fixes #3286

### DIFF
--- a/graphify/install.py
+++ b/graphify/install.py
@@ -32,6 +32,28 @@ except Exception:
 from graphify.paths import GRAPHIFY_OUT as _GRAPHIFY_OUT
 
 
+def _write_version_stamp(skill_dst: Path, version: str) -> None:
+    """Atomically write ``.graphify_version`` beside ``skill_dst``.
+
+    Matches the SKILL.md install path: temp file in the same directory, then
+    ``os.replace``. Unlike ``Path.write_text`` (and unlike
+    ``paths.write_text_atomic``, which resolves through symlinks), ``os.replace``
+    replaces a managed symlink in place — consistent with SKILL.md /
+    ``references/`` and crash-safe against a half-written stamp (#3286).
+    """
+    version_file = skill_dst.parent / ".graphify_version"
+    tmp = version_file.with_name(".graphify_version.tmp")
+    try:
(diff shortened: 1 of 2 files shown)


safishamsi:
Shipped in **v0.9.54** — landed on `v8` via cherry-pick with your authorship preserved (so it lands on your contributor graph). Thanks @drmikecrypto — the .graphify_version stamp is now written atomically. Release: https://github.com/Graphify-Labs/graphify/releases/tag/v0.9.54

PR #3304, opened 2026-09-03
fix(export): write graph.canvas and Obsidian notes atomically

### LINKED ISSUE #3282: to_canvas() writes graph.canvas non-atomically — concurrent readers can SIGBUS
\`to_canvas()\` in \`graphify/export.py\` writes the Obsidian canvas file with:

    Path(output_path).write_text(json.dumps(canvas_data, indent=2), encoding="utf-8")

\`write_text\` opens the target with truncate, so for a moment the file is empty/short
before the new content lands. Any concurrent process that has the file open — e.g.
\`git commit\` refreshing its index, which stats then mmap()s the working-tree file to
hash it (\`ie_match_stat\` → \`ce_modified_check_fs\` → \`ce_compare_data\` → \`index_fd\`) —
can end up with a mapping whose backing file was truncated mid-read, which SIGBUSes
(\`BUS_ADRERR\`) on the next access into the mapped-but-now-unbacked pages.

Reproduced this on Arch/Omarchy: two independent automation loops both ran
\`graphify . --code-only\` + \`graphify export obsidian\` against the same vault
directory roughly every 30-90s (no lock between them). git 2.55.0 crashed with
SIGBUS while \`git commit\` was refreshing its index and hashing \`graph.canvas\`,
which one of the graphify runs was truncating/rewriting at that exact instant.
Confirmed via debuginfod-symbolized backtrace + coredump: the crashing frame's
captured \`stat()\` shows the file's mtime landing in the same second as the crash,
i.e. it was being rewritten while git had it mmapped.

Full backtrace (git 2.55.0-1, Arch):

    sha1_compression_states (sha1dc/sha1.c:393)   <- SIGBUS here
    SHA1DCUpdate
    hash_object_body
    index_mem                                      (buf = mmap of graph.canvas)
    index_core / index_fd
    ce_compare_data
    ce_modified_check_fs
    ie_match_stat
    refresh_cache_ent   (path = ".../graph.canvas")
    refresh_index

This isn't git's bug — mmap-based hashing of a working-tree file is standard and
correct as long as nothing else truncates that file underneath it mid-read. The
fix belongs on the writer's side: write to a temp file in the same directory and
\`os.replace()\` it into place, so any concurrent reader always sees either the
complete old file or the complete new one, never a truncated intermediate state.

Suggested fix (this is what I applied locally to unblock myself):

    def _atomic_write_text(path, content, encoding="utf-8"):
        path = Path(path)
        fd, tmp_name = tempfile.mkstemp(dir=path.parent, prefix=f".{path.name}.", suffix=".tmp")
        try:
            with os.fdopen(fd, "w", encoding=encoding) as f:
                f.write(content)
            os.replace(tmp_name, path)
        except BaseException:
            try:
                os.unlink(tmp_name)
            except OSError:
                pass
            raise

...used in place of the \`write_text\` call in \`to_canvas()\`.

The same non-atomic-write pattern also exists in \`_owned_write()\` (used for the
per-node Obsidian notes and \`.obsidian/graph.json\`), which has the identical
exposure for any external reader — I haven't touched that one, just flagging it
as the same class of issue.

Environment: graphifyy 0.9.53 (PyPI), Python 3.14, git 2.55.0-1, Arch Linux (Omarchy).

Filed by Claude Sonnet 5 via Claude Code.

drmikecrypto:
Route to_canvas and _owned_write through paths.write_*_atomic so concurrent readers (e.g. git hashing) never see a truncated file.

Fixes #3282

### DIFF
--- a/graphify/export.py
+++ b/graphify/export.py
@@ -16,7 +16,7 @@ from networkx.readwrite import json_graph
 from graphify.security import sanitize_label
 from graphify.analyze import _node_community_map
 from graphify.build import edge_data
-from graphify.paths import stem_filename_budget
+from graphify.paths import stem_filename_budget, write_json_atomic, write_text_atomic
 
 from graphify.exporters.graphdb import push_to_falkordb, push_to_neo4j  # noqa: E402,F401
 
@@ -726,7 +726,7 @@ def to_obsidian(
             _skipped.append(rel_name)
             return False
         target.parent.mkdir(parents=True, exist_ok=True)
-        target.write_text(content, encoding="utf-8")  # nosec
+        write_text_atomic(target, content)
         _written.append(rel_name)
         return True
 
@@ -1199,7 +1199,7 @@ def to_canvas(
         })
 
     canvas_data = {"nodes": canvas_nodes, "edges": canvas_edges}
-    Path(output_path).write_text(json.dumps(canvas_data, indent=2), encoding="utf-8")  # nosec
+    write_json_atomic(output_path, canvas_data, indent=2)
 
 
 def to_graphml(
--- a/tests/test_atomic_canvas_export.py
+++ b/tests/test_atomic_canvas_export.py
@@ -1,0 +1,73 @@
+"""Regression: to_canvas / Obsidian vault writes must be atomic (#3282)."""
+from __future__ import annotations
+
+import json
+import os
+from pathlib import Path
+
+import networkx as nx
+import pytest
+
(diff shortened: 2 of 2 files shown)


safishamsi:
Shipped in **v0.9.54** — landed on `v8` via cherry-pick with your authorship preserved (so it lands on your contributor graph). Thanks @drmikecrypto — graph.canvas and Obsidian notes are now written atomically. Release: https://github.com/Graphify-Labs/graphify/releases/tag/v0.9.54

PR #3306, opened 2026-09-03
fix(manifest): require tomli on Py3.10 and fail loud when missing

### LINKED ISSUE #3283: manifest_ingest silently skips every Cargo.toml/pyproject.toml on Python 3.10 without tomli (parsers return None on ImportError)
We chased the nightly "zero nodes" warning for two nights before finding this, so I'm filing it separately from #1666's family: the cause turned out to be an install-shape problem that the code hides instead of reporting.

Setup: graphifyy 0.9.53 installed with `uv tool install graphifyy` into a Python 3.10.20 venv. That venv had no `tomli`, and 3.10 has no stdlib `tomllib`.

With that setup, every `Cargo.toml` and `pyproject.toml` in the repo extracts to zero nodes, every night, and the retry never changes anything. The generic zero-node warning fires (the one saying empties are no longer cached and a re-run will retry), which reads like a per-file extraction hiccup. It isn't. The whole package/dependency layer is missing from the graph, and nothing says why.

The cause is in `manifest_ingest.py`, in both `_parse_pyproject` and `_parse_cargo`:

```python
try:
    import tomllib as _toml
except ImportError:
    try:
        import tomli as _toml
    except ImportError:
        return None
```

`return None` on ImportError is indistinguishable from the legitimate None for a virtual workspace root (`[workspace]` with no `[package]`), so a missing parser looks exactly like "nothing to emit". The odd part is that `cargo_introspect.py` in the same package already handles the identical condition loudly:

```python
raise ImportError("--cargo on Python 3.10 needs tomli. Install with: pip install tomli")
```

Numbers, in case they help: 26 of 26 Cargo.toml files came out zero-node under the 3.10 venv. The identical code under 3.12 gave 25 of 26 with real dependency edges, and the 26th is the virtual workspace root, which is correct. 25 matches the nightly's reported count exactly. On our host `uv tool install --force --with tomli graphifyy` fixed it and manifests now ingest normally.

The ask: either declare the dependency with an environment marker (`tomli; python_version < "3.11"`) so 3.10 installs get it for free, or let the manifest parsers fail the way `cargo_introspect` already does instead of returning None. Right now a missing capability shows up as permanent per-file extraction noise, and a warning that fires every night stops being read as a warning.

Happy to share the venv details or re-run the measurement on a newer build if that's useful. Related: #1264 (the Cargo extractor), #2434 (toml as an extension).

drmikecrypto:
Declare tomli with an environment marker and raise ImportError instead of returning None, so missing parsers no longer look like empty manifests.

Fixes #3283

### DIFF
--- a/graphify/manifest_ingest.py
+++ b/graphify/manifest_ingest.py
@@ -36,6 +36,31 @@ PACKAGE_MANIFEST_NAMES: dict[str, str] = {
 
 _MAX_MANIFEST_BYTES = 2_000_000  # 2 MB cap — manifests are small; this rejects junk
 
+_TOMLI_REQUIRED = (
+    "Package-manifest ingestion on Python < 3.11 needs tomli. "
+    "Install with: pip install 'tomli' "
+    "(or reinstall graphifyy, which declares tomli for python_version < '3.11')."
+)
+
+
+def _load_toml_module():
+    """Return a tomllib-compatible module, or raise ImportError (#3283).
+
+    Returning ``None`` used to look identical to a virtual workspace root with
+    nothing to emit, so missing ``tomli`` on Python 3.10 silently dropped every
+    ``Cargo.toml`` / ``pyproject.toml``. Fail loud the way ``cargo_introspect``
+    already does.
+    """
+    try:
+        import tomllib as _toml  # type: ignore[import-not-found]
+        return _toml
+    except ImportError:
+        try:
+            import tomli as _toml  # type: ignore[import-not-found,no-redef]
+            return _toml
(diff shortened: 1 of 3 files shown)


safishamsi:
Shipped in **v0.9.54** — landed on `v8` via cherry-pick with your authorship preserved (so it lands on your contributor graph). Thanks @drmikecrypto — tomli is now a runtime dependency on Python <3.11 so manifests are no longer silently dropped. Release: https://github.com/Graphify-Labs/graphify/releases/tag/v0.9.54

PR #3311, opened 2026-09-03
feat(llm): cursor-cli backend — semantic extraction through the locally authenticated Cursor Agent CLI

Ketlark:
Complements #3073 (openai-cli, Codex side) with the Cursor equivalent. Same shape as the existing `claude-cli` backend: zero pricing (subscription usage is not metered API spend), forced-serial calls unless opted out, and a credential gate that accepts a present CLI in place of an API key.

### What it adds

`--backend cursor-cli` runs semantic extraction and community labeling through `cursor-agent -p`, the locally authenticated Cursor Agent CLI, so the work rides a Cursor subscription instead of a metered API key.

Four things worth stating for review:

1. **The prompt goes over stdin, not argv.** Real extraction chunks reach 240–306 KB; Linux caps a single argv entry at 128 KB (`MAX_ARG_STRLEN`), so argv delivery fails with `E2BIG` on ordinary input.
2. **`--mode ask` keeps the agent read-only.** An agentic CLI given write access to the corpus it is reading can modify what it is extracting from. `--trust` satisfies non-interactive workspace trust (headless runs abort with a trust prompt otherwise).
3. **Envelope errors surface before hollow-retry bisection.** `cursor-agent` flags API failures with `is_error` in the stdout envelope while exiting 0. Parsing `result` as model output yields an empty graph that the bisect path then subdivides forever; a nonzero exit with a parseable envelope surfaces the envelope's cause, and garbage stdout on a success exit fails loudly. This mirrors the #2554 handling in claude-cli.
4. **Model routing is not pinned.** Without `GRAPHIFY_CURSOR_CLI_MODEL` no `--model` flag is passed and Cursor's own `auto` routing decides; the env var pins a specific model when wanted.

Token usage is read from the JSON envelope's `usage` block (`inputTokens`/`outputTokens`).

Cursor Agent does not expose an image-attach flag, so `cursor-cli` is not a vision backend — images fall through the same non-vision text-reference path as other non-vision backends.

### Diff

5 files: `graphify/llm.py` (+150), `tests/test_cursor_cli_backend.py` (+221), `graphify/cli.py` (+10, credential gate), `graphify/__main__.py` (+2/−2, help text), `CHANGELOG.md` (+4).

### Testing

- `tests/test_cursor_cli_backend.py`: 18 tests, mocked `subprocess.run` + `shutil.which` (no binary, no network) — parse/token accounting, error-envelope surfacing on zero and nonzero exit, garbage-envelope failure, `_call_llm` labeling path, usage accumulation, dispatch through `extract_files_direct`, zero-cost registration, stdin/argv shape, model pinning, user-turn delivery with `<untrusted_source>` guardrails intact.
- `uv run pytest tests/test_cursor_cli_backend.py tests/test_claude_cli_backend.py tests/test_llm_backends.py`: 156 passed.
- Full suite: 3591 passed, 70 skipped, 1 pre-existing failure on pristine v8 (test_ollama_retry_cap imports the optional `openai` SDK; reproduces on a clean checkout without this branch — environment, not the patch).
- Live end-to-end against a real authenticated `cursor-agent` install: extraction of a Markdown corpus returned a well-formed graph with usage tracking; cost $0 (plan-billed).

### One honest caveat

Live-verified against one person's Cursor install on macOS; like #3073, results depend on the CLI's JSON envelope staying stable. `AGENTS.md` followed: `graphify update .` run after the code changes.


Ketlark:
Ran the missing executions on the gate's five findings. One was valid and is fixed in 1693e18; two don't reproduce at head; two are deliberate parity with the shipped claude-cli backend.

**1. "cursor-cli backend not forced to serial concurrency" (cli.py:3689) — does not reproduce.** The gate lives in llm.py, not at the credential check the finding anchors to. Both concurrency surfaces are forced-serial for cursor-cli, mirroring claude-cli:

- `graphify/llm.py:2723` — `extract_corpus_parallel`: `if backend == "cursor-cli" and os.environ.get("GRAPHIFY_CURSOR_CLI_PARALLEL", "").strip() != "1": max_concurrency = 1`
- `graphify/llm.py:3503` — community labeling: same guard

Execution: `grep -n GRAPHIFY_CURSOR_CLI_PARALLEL graphify/llm.py` returns exactly those two sites; the claude-cli counterpart guards sit directly above each.

**2 & 3. "Untrusted source text is sent to Cursor Agent with workspace trust enabled" / "Generic Cursor LLM path also forces trusted agent execution" (llm.py:1827, 3044) — accurate observation of deliberate, documented behavior shared with the shipped claude-cli backend.** `--mode ask` confines the agent to read-only tools — strictly less capable than the shipped claude-cli backend, which runs `claude -p` with default tool permissions and even allowlists extra directories via `--add-dir` (llm.py, claude-cli images path). The corpus-to-prompt injection surface is identical in kind to claude-cli; ask mode reduces the blast radius relative to it. The docstring in `_call_cursor_cli` documents the ask-mode choice.

**4. "cursor-cli model flag uses mdl but guards on model, ignoring config default" (llm.py:3049) — deliberate parity.** `BACKENDS["cursor-cli"]["default_model"]` is a display placeholder (`cursor-agent-plan`), not a real model — passing it to `--model` would break auto routing. The `model is not None` guard means "caller explicitly pinned a model", which is exactly the shipped claude-cli semantic: `llm.py:3009-3010` (`if model is not None: cli_args.extend(["--model", mdl])`) — my branch mirrors it 1:1 at 3051-3052.

**5. "Default-model test depends on external environment" (tests/test_cursor_cli_backend.py:167) — valid, fixed in 1693e18.** `test_no_model_flag_by_default` now scrubs the variable via `monkeypatch.delenv("GRAPHIFY_CURSOR_CLI_MODEL", raising=False)`, mirroring the existing `GRAPHIFY_API_TIMEOUT` delenv precedent in tests/test_claude_cli_backend.py:381. Reproduction of the fix under a hostile environment:

```
GRAPHIFY_CURSOR_CLI_MODEL=leak-test uv run pytest tests/test_cursor_cli_backend.py -q
→ 18 passed
```

Pre-fix, an exported `GRAPHIFY_CURSOR_CLI_MODEL` injected `--model` into the subprocess argv and failed the assertion.

Full suite re-run at head: `tests/test_cursor_cli_backend.py tests/test_claude_cli_backend.py tests/test_llm_backends.py` → 156 passed.

PR #3315, opened 2026-09-03
Fix module docstring extraction when a leading comment precedes it

### LINKED ISSUE #3312: Module docstring never extracted when a comment precedes it (shebang-led files lose all module rationale)
### Summary

`_extract_python_rationale` never extracts a **module** docstring when any comment precedes it — including the `#!` shebang that begins essentially every executable Python script. Function and class docstrings in the same file are extracted normally, so the loss is silent and specific to module-level rationale.

### Reproducer

Three files in an empty directory:

```python
# with_shebang.py
#!/usr/bin/env python3
"""MODULE docstring of the shebang-led file, long enough to pass the length filter."""


def f():
    """FUNCTION docstring of the shebang-led file, long enough to pass the filter."""
    return 1
```

```python
# without_shebang.py
"""MODULE docstring of the plain file, long enough to pass the length filter."""


def g():
    """FUNCTION docstring of the plain file, long enough to pass the filter."""
    return 2
```

```python
# with_comment.py
# an ordinary leading comment, not a shebang
"""MODULE docstring behind a plain comment, long enough to pass the filter."""


def h():
    return 3
```

```
$ GRAPHIFY_OUT=./out graphify update . --no-cluster
```

### Observed (graphify 0.9.53)

`file_type: rationale` nodes produced:

| file | module docstring | function docstring |
|---|---|---|
| `without_shebang.py` | ✅ L1 | ✅ L5 |
| `with_shebang.py` | ❌ **missing** | ✅ L6 |
| `with_comment.py` | ❌ **missing** | — |

### Expected

All three module docstrings extracted as `rationale` nodes at their own line.

### Cause

`graphify/extract.py`, `_extract_python_rationale._get_docstring`:

```python
def _get_docstring(body_node) -> tuple[str, int] | None:
    if not body_node:
        return None
    for child in body_node.children:
        if child.type == "expression_statement":
            ...
        break          # <-- unconditional, after the FIRST child
    return None
```

The `break` is unconditional, so only the first child of the node is ever considered. At module level the first child of a shebang-led (or comment-led) file is a `comment`, not the `expression_statement` holding the docstring, so the function returns `None`. Function and class bodies are unaffected because their first child *is* the docstring statement.

A `continue` past `comment` children (or skipping non-`expression_statement` children until the first statement) would fix it. Preserving the "only look at the leading statement" intent matters — you don't want to pick up a string literal from halfway down the module.

### Why it's worth fixing

Module docstrings are disproportionately the highest-value rationale in a codebase: they describe *why the file exists*, and executable entry points — the files people most often ask "what does this do" about — are precisely the ones that carry a shebang and therefore lose it. On a ~25k-node graph across 14 repositories we found 158 files with a module-level (`L1`) rationale node and **not one** of them was shebang-led; every shebang-led file with a module docstring was missing it.

Happy to send a PR if the `continue` shape is the direction you'd want.

ayushcodes10:
_get_docstring's loop checked only the FIRST child of a module/class/
function body for a docstring expression_statement, then unconditionally
broke -- so a leading comment (the #! shebang essentially every
executable Python script starts with, a coding declaration, a license
header) sat in that first-child slot instead of the docstring, and the
real docstring one position later was never seen. No warning, no
partial-extraction report, exit 0 -- module rationale for an entire
class of file (any shebang-led or comment-led script) was silently
absent from the graph.

Comments are not statements: a docstring must be the first STATEMENT
in the body, so the fix skips leading `comment` children before
checking whether the first real statement is a docstring. The same
shared helper also serves class/function bodies, so a comment as the
first line of a function/class body is fixed too, not just the
module-level case reported in the issue.

Verified against the exact 3-file repro from the issue (before/after),
plus additional cases I traced through myself: a multi-line header
(shebang + coding declaration + license, 3 stacked comments), a
function/class docstring behind a leading in-body comment, a file with
a leading comment and genuinely no docstring (must not fabricate one),
and a string literal that is NOT the first statement (must still be
correctly rejected, not accidentally picked up by the new skip logic).

6 new tests in tests/test_rationale.py covering all of the above.

Full suite: 5117 passed, 0 failed. ruff and skillgen --check clean.

Fixes #3312.

### DIFF
--- a/graphify/extract.py
+++ b/graphify/extract.py
@@ -1205,9 +1205,22 @@ def _extract_python_rationale(path: Path, result: dict) -> None:
     file_nid = _make_id(str(path))
 
     def _get_docstring(body_node) -> tuple[str, int] | None:
+        """A docstring is the first STATEMENT in a module/class/function body.
+
+        A leading `comment` node — the shebang line essentially every
+        executable script starts with, a coding-declaration or license
+        header, or any ordinary comment — is not a statement: tree-sitter
+        still parses it as a sibling child of the body, but Python's own
+        docstring rule skips right over it. The old unconditional `break`
+        after the first loop iteration stopped at that comment instead of
+        looking past it, so a module docstring behind a shebang (or any
+        leading comment) was silently never found (#3312).
+        """
         if not body_node:
             return None
         for child in body_node.children:
+            if child.type == "comment":
+                continue
             if child.type == "expression_statement":
                 for sub in child.children:
                     if sub.type in ("string", "concatenated_string"):
--- a/tests/test_rationale.py
+++ b/tests/test_rationale.py
@@ -45,6 +45,99 @@ def test_class_docstring_extracted(tmp_path):
     assert any("Redis" in n["label"] for n in rationale)
 
 
+def test_module_docstring_extracted_behind_shebang(tmp_path):
+    """#3312: a leading `#!` shebang comment must not hide the module docstring
+    that follows it — essentially every executable Python script starts with
+    one, so this silently dropped module rationale for that whole class of
+    file."""
+    p = tmp_path / "sample.py"
+    p.write_text(
+        "#!/usr/bin/env python3\n"
+        '"""This module handles authentication because legacy sessions were insecure."""\n'
+        "def login(): pass\n"
+    )
+    result = extract_python(p)
+    rationale = [n for n in result["nodes"] if n.get("file_type") == "rationale"]
+    assert any("authentication" in n["label"] for n in rationale)
+    module_ds = next(n for n in rationale if "authentication" in n["label"])
+    assert module_ds["source_location"] == "L2"
+
+
+def test_module_docstring_extracted_behind_plain_comment(tmp_path):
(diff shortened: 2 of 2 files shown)


ayushcodes10:
Correction: my earlier note about 7 pre-existing test failures was wrong. Those were caused by a stale pip-installed `graphifyy 0.9.38` on my machine shadowing this checkout for every subprocess-based test (`subprocess.run([sys.executable, "-m", "graphify"], ...)` resolves via site-packages, not the local repo, unless installed editable). `git stash` correctly showed the failures were unaffected by my diff, but I concluded "pre-existing on v8" without checking *why* they failed — the real cause was my environment, not this codebase. Reinstalled editable (`pip install -e .`) and the full suite is clean: 5117 passed, 0 failed. Updated the PR description to match.

safishamsi:
Shipped in **v0.9.55** — landed on `v8` via cherry-pick with your authorship preserved. Thanks @ayushcodes10 — module docstrings preceded by a leading comment (shebang/coding/license) are now extracted. Release: https://github.com/Graphify-Labs/graphify/releases/tag/v0.9.55

PR #3316, opened 2026-09-03
fix(ingest): make memory-doc filenames unique so concurrent saves do not overwrite

### LINKED ISSUE #3301: save_query_result: memory docs sharing a second and a 50-char question prefix silently overwrite each other
### Summary

`save_query_result` derives the memory-doc filename from a **one-second timestamp** plus the **first 50 characters** of the question, then writes with a plain `write_text`. Two saves that land in the same second and share a 50-character question prefix resolve to the same path, and the later one silently replaces the earlier one. No error is raised and both calls return normally.

The one-file-per-memory layout is what makes unrelated writes safe here — they touch different paths, so no lock is needed. The filename is the one place where that property does not hold.

### Where

`graphify/ingest.py` (v0.9.53):

```python
slug = re.sub(r"[^\w]", "_", question.lower())[:50].strip("_")
filename = f"query_{now.strftime('%Y%m%d_%H%M%S')}_{slug}.md"
...
out_path.write_text(content, encoding="utf-8")
```

Three properties combine:

- the timestamp has one-second granularity
- the slug is truncated at 50 characters
- the writer is not part of the name (`contributor` is hardcoded to `"graphify"` on this path)

### Reproduction

```python
"""Minimal reproduction: concurrent save_query_result calls silently overwrite."""
import shutil, tempfile
from concurrent.futures import ThreadPoolExecutor
from pathlib import Path

from graphify.ingest import save_query_result

def run(label, questions):
    d = tempfile.mkdtemp()
    try:
        with ThreadPoolExecutor(max_workers=len(questions)) as ex:
            list(ex.map(lambda q: save_query_result(q, "an answer", Path(d)), questions))
        kept = len(list(Path(d).glob("*.md")))
    finally:
        shutil.rmtree(d, ignore_errors=True)
    print(f"{label:52s} saved={len(questions):3d} kept={kept:3d} lost={len(questions)-kept:3d}")

PREFIX = "how does the authentication subsystem resolve tokens"
run("distinct questions", [f"how does module {i} work" for i in range(20)])
run("same question, same second", ["how does auth work"] * 20)
run("questions differing only after char 50", [f"{PREFIX} in module {i}" for i in range(20)])
```

Output on v0.9.53 (macOS, CPython 3.9, APFS):

```
distinct questions                                   saved= 20 kept= 20 lost=  0
same question, same second                           saved= 20 kept=  1 lost= 19
questions differing only after char 50               saved= 20 kept=  1 lost= 19
```

The first row is the design working: unrelated writes never collide. The other two are the filename.

Spacing the same question more than a second apart is enough to avoid it, which confirms the mechanism:

```
same question, 1.05s apart                           saved=  3 kept=  3 lost=  0
```

### Why it matters in practice

The third case is not contrived. When several agents sweep one subsystem in parallel, their questions share a long prefix:

```
"how does the authentication subsystem resolve tokens in module 1"
"how does the authentication subsystem resolve tokens in module 2"
```

Both truncate to `how_does_the_authentication_subsystem_resolve_toke`, so distinct findings collapse into one file. The symptom is only ever "the graph does not seem to know that", which is hard to distinguish from a retrieval miss — so it is unlikely to be reported as data loss.

### Suggested fix

Put something writer- or call-unique in the name, so distinct saves cannot resolve to one path — for example a short random suffix, or the contributor plus a counter:

```python
filename = f"query_{now.strftime('%Y%m%d_%H%M%S')}_{contributor}_{uuid4().hex[:8]}_{slug}.md"
```

An `open(..., "x")` retry loop would also close it without changing the naming scheme. Either keeps the existing layout and the `*.md` glob in `reflect.load_memory_docs` unchanged.

This is the same class as #2650 (`to_obsidian` case-only filename collisions silently overwriting node notes), on a different code path.

Happy to open a PR if a direction is preferred — I did not want to guess at the naming convention you would want in the filename.

daichiyasunami-vottia:
Fixes #3301.

`save_query_result` names memory docs `query_<second>_<slug50>.md` and writes them with a plain `write_text`. Two saves that land in the same second and share the first 50 characters of the question resolve to one path, and the later one silently replaces the earlier. Both calls return normally.

Measured on v0.9.53 (20 concurrent saves): distinct questions keep 20/20, the same question keeps 1/20, and questions differing only after character 50 keep 1/20. The third case is the ordinary one when several agents sweep one subsystem in parallel.

**Change:** add `uuid.uuid4().hex[:8]` to the filename. The `query_` prefix and `.md` suffix are unchanged, so `reflect.load_memory_docs` (`*.md` glob) and the existing tests are unaffected. One regression test added: 20 concurrent saves of one question → 20 files.

The one-file-per-memory layout is what makes unrelated writes safe here; this only closes the filename.

### DIFF
--- a/graphify/ingest.py
+++ b/graphify/ingest.py
@@ -2,6 +2,7 @@
 from __future__ import annotations
 import json
 import re
+import uuid
 import urllib.error
 import urllib.parse
 from datetime import datetime, timezone
@@ -299,7 +300,11 @@ def save_query_result(
 
     now = datetime.now(timezone.utc)
     slug = re.sub(r"[^\w]", "_", question.lower())[:50].strip("_")
-    filename = f"query_{now.strftime('%Y%m%d_%H%M%S')}_{slug}.md"
+    # A second-granularity stamp plus a 50-char slug is not unique: two saves in
+    # the same second whose questions share a prefix resolve to one path, and the
+    # later write_text silently replaces the earlier one (#3301). The short uuid
+    # makes every save its own file; the query_ prefix and .md suffix are kept.
+    filename = f"query_{now.strftime('%Y%m%d_%H%M%S')}_{uuid.uuid4().hex[:8]}_{slug}.md"
 
     frontmatter_lines = [
         "---",
--- a/tests/test_ingest.py
+++ b/tests/test_ingest.py
@@ -98,3 +98,15 @@ def test_no_outcome_means_no_outcome_section(tmp_path):
 def test_invalid_outcome_rejected(tmp_path):
     with pytest.raises(ValueError):
         save_query_result("q", "a", tmp_path / "memory", outcome="great")
+
+
+def test_concurrent_saves_of_the_same_question_do_not_overwrite(tmp_path):
+    """Regression for #3301: a second-granularity stamp plus a 50-char slug is
+    not unique, so saves in the same second sharing a prefix collapsed into one
+    file and the earlier ones were silently lost."""
+    from concurrent.futures import ThreadPoolExecutor
+    mem = tmp_path / "memory"
+    with ThreadPoolExecutor(max_workers=20) as ex:
+        paths = list(ex.map(lambda _: save_query_result("how does auth work", "a", mem), range(20)))
+    assert len({p.name for p in paths}) == 20
+    assert len(list(mem.glob("*.md"))) == 20


safishamsi:
Shipped in **v0.9.54** — landed on `v8` via cherry-pick with your authorship preserved (so it lands on your contributor graph). Thanks @daichiyasunami-vottia — saved memory-doc filenames now carry a unique suffix so same-second saves no longer overwrite each other. Release: https://github.com/Graphify-Labs/graphify/releases/tag/v0.9.54

daichiyasunami-vottia:
Thank you for cherry-picking it onto `v8` and for the quick release. Confirmed `2f743aea` is in `v0.9.54`; closing the loop on #3301 from my side.

PR #3317, opened 2026-09-03
Fix Python leading-underscore method/function id collision

### LINKED ISSUE #3302: Python: methods differing only by a leading underscore collide on node id — the public one is silently dropped from the graph
**Version:** 0.9.53 (latest on PyPI at time of writing; also reproduced on 0.9.47)
**Platform:** Windows 11, Python 3.14.7, isolated venv (`pip install graphifyy==0.9.53`)
**Command:** `graphify extract . --code-only --no-cluster`

> Note for maintainers: I first sent this through private vulnerability reporting
> (`GHSA-7hhr-924m-gwrf`) because SECURITY.md asks that security-relevant reports not be
> opened publicly. On reflection that was the wrong channel � this is a correctness bug, not
> a vulnerability in graphify � so I'm refiling it here. **Please close that private report as
> a duplicate of this issue**; I can't close it myself as the reporter.

## Summary

When a class defines two methods whose names differ only by a leading underscore (`_foo` and
`foo`), only one node reaches `graph.json`. Node ids appear to be slugified in a way that
strips the leading underscore, so both become `<module>_<class>_foo`, and the second
definition silently overwrites the first.

No warning is emitted, exit status is 0, and the file is not reported as partially extracted.
A public API method can therefore be **entirely absent** from the graph while its private
sibling occupies the public name.

## Reproduction

`mod.py` � three methods in, two nodes out:

```python
class Adapter:
    def _get_connection(self, url):     # L2
        return url

    def get_connection(self, url):      # L5
        return self._get_connection(url)

    def unrelated(self, x):             # L8   <- control
        return x
```

```console
$ graphify extract . --code-only --no-cluster
[graphify extract] wrote graphify-out/graph.json - 4 nodes, 3 edges (no clustering)
```

Nodes for `mod.py`:

```
mod                         | mod.py             | L1
mod_adapter                 | Adapter            | L1
mod_adapter_get_connection  | ._get_connection() | L2
mod_adapter_unrelated       | .unrelated()       | L8
```

**Expected:** three method nodes. **Actual:** two. `Adapter.get_connection` (L5) is absent;
the id that would name it is held by `_get_connection` (L2).

The `unrelated` control extracts correctly, so this is specific to the underscore-differing
pair, not to methods generally.

## Real-world impact

Found while benchmarking context retrieval against `psf/requests` at `aa1461b6^`.
`src/requests/adapters.py` defines both `HTTPAdapter._get_connection` (L377) and
`HTTPAdapter.get_connection` (L406). The graph contains a single node
`src_requests_adapters_httpadapter_get_connection` pointing at **L377**; the public
`get_connection` at L406 has no node at all, only a stray docstring node at L407.

In a 25-symbol ground-truth set built from that repository's own fix commits, this defect
accounted for a measurable share of symbols that no graph query (`affected`, `explain`,
`query`, `path`) could reach. Anyone relying on graph completeness � for an API audit, a
refactor impact review, or a security pass � gets a quiet false negative.

## Root cause (hypothesis) and suggested fix

Id generation appears to normalise identifiers by stripping non-alphanumeric characters,
including the leading `_`, without disambiguating collisions.

1. Preserve the underscore in the id (e.g. `mod_adapter__get_connection`); or
2. Detect an id collision at insert time and disambiguate rather than overwrite.

(2) is more robust, since the same collapse likely affects other character-stripping cases
(`foo_bar` vs `fooBar` vs `foo-bar`, dunder names). A regression test asserting
**node count == definition count** for a file would catch the whole class.

## Related, but not duplicates

- **#387** (`.mjs` files produce 0 nodes, fixed v0.4.16) was an all-`.mjs` failure from a
  missing `_DISPATCH` entry. This is Python, the dispatch path plainly works (siblings in the
  same class extract fine), and the loss is per-symbol rather than per-file.
- **#2281** (JS/MJS ids minted from absolute paths) is a different cause but the same family:
  an id-shape defect that silently changes which node exists.

ayushcodes10:
ids.py:make_id strips leading/trailing underscores from every name
part before normalizing, so _get_connection/get_connection (and more
broadly any name pair differing only by underscore wrapping --
x/_x/__x/__x__ dunder variants all collapse the same way) mint the
SAME node id. add_node's first-wins dedup then silently drops
whichever declaration is walked second: a public method could be
entirely absent from the graph while its private-by-convention
sibling occupied the public name, with no warning and exit 0.

Adds a Python-scoped pre-scan (_python_pre_scan_underscore_collisions)
that groups module-level functions and direct class methods by their
would-be node id, plus a resolver (_python_underscore_salted_nid) that
salts every group member except a unique public (no leading
underscore) one -- mirroring the established convention the Go
extractor already uses for its own case-only collision (#2779/#2780):
when no unique public member exists, every member is salted, so the
outcome never depends on declaration order. The public member's id
stays exactly what it would be with no colliding sibling at all, so
adding/removing a private-by-convention sibling in an incremental
rebuild re-points nothing.

Both hooks are gated on config.ts_module == "tree_sitter_python" and
live inside the shared _extract_generic engine (Python has no
dedicated extractor file, unlike Go/Rust/SQL), following the existing
precedent for per-language pre-scans already there (csharp_interface_names,
swift_protocol_names). Call-site resolution needed no changes: label_to_nid
is rebuilt from the final nodes list after the walk completes, so it
naturally picks up whichever id (salted or not) a method actually got.

9 new tests in tests/test_python_underscore_resolution.py: the exact
repro, id stability across incremental add/remove of a private
sibling, the call edge correctly binding to the salted target, the
module-level (non-method) case, the no-unique-public-member case
(both members salted), a dunder+plain collision, an unaffected
no-collision file, and scope isolation (a collision in one class must
not touch an unrelated same-named-but-non-colliding member in another
class).

Full suite: 5117 passed, 0 failed. ruff and skillgen --check clean.

Fixes #3302.

### DIFF
--- a/graphify/extractors/engine.py
+++ b/graphify/extractors/engine.py
@@ -813,6 +813,84 @@ def _swift_declaration_keyword(node) -> str | None:
             return c.type
     return None
 
+def _python_pre_scan_underscore_collisions(root_node, source: bytes, stem: str) -> dict[str, set[str]]:
+    """Pre-scan a Python module for name-only differences that collapse to one node id.
+
+    ``ids.py:make_id`` strips leading/trailing underscores from every part before
+    normalizing, so ``_get_connection`` and ``get_connection`` mint the SAME id.
+    ``add_node`` then silently drops whichever declaration is walked second — no
+    warning, exit 0 — so a public method can be entirely absent from the graph
+    while its private-by-convention sibling (or a dunder, which strips the same
+    way: ``__x``/``__x__``/``x`` all collapse too) occupies the public name (#3302).
+
+    Returns ``{plain_nid: {raw names that would collide on it}}`` for every
+    module-level function and every direct method of a module-level class — the
+    two cases the id-minting code actually distinguishes via ``parent_class_nid``.
+    Deliberately does not recurse into nested classes or nested functions: this
+    keeps the scope key trivially exact (matching ``_make_id(stem, class_name)``,
+    which only holds because Python never populates ``namespace_stack``) rather
+    than replicating the corpus-wide id-computation rules for every nesting shape.
+    A collision entirely inside an unhandled nested scope is simply not caught
+    here — a strict miss, never a false positive, since the map is only ever
+    consulted for a nid this same scan actually computed.
+    """
+    groups: dict[str, set[str]] = {}
+
+    def _record(plain_nid: str, name: str) -> None:
+        groups.setdefault(plain_nid, set()).add(name)
+
+    for child in root_node.children:
+        if child.type == "function_definition":
+            name_node = child.child_by_field_name("name")
+            if name_node is not None:
+                name = _read_text(name_node, source)
+                if name:
+                    _record(_make_id(stem, name), name)
+        elif child.type == "class_definition":
+            class_name_node = child.child_by_field_name("name")
+            body = child.child_by_field_name("body")
+            if class_name_node is None or body is None:
+                continue
+            class_name = _read_text(class_name_node, source)
+            if not class_name:
+                continue
+            class_nid = _make_id(stem, class_name)
+            for member in body.children:
+                if member.type != "function_definition":
+                    continue
+                name_node = member.child_by_field_name("name")
+                if name_node is None:
+                    continue
+                name = _read_text(name_node, source)
+                if name:
+                    _record(_make_id(class_nid, name), name)
+
+    return {nid: names for nid, names in groups.items() if len(names) >= 2}
+
+
+def _python_underscore_salted_nid(plain_nid: str, name: str, groups: dict[str, set[str]]) -> str:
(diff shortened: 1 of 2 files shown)


ayushcodes10:
Correction: same as the note I just posted on #3315 — the "7 pre-existing failures" I reported were caused by a stale pip-installed `graphifyy 0.9.38` on my machine shadowing this checkout for subprocess-based tests, not a real issue with this codebase or this change. Reinstalled editable and the full suite is clean: 5117 passed, 0 failed. Updated the PR description to match.

safishamsi:
Shipped in **v0.9.55** — landed on `v8` via cherry-pick with your authorship preserved. Thanks @ayushcodes10 — Python defs differing only by leading underscores no longer collide and drop one. Release: https://github.com/Graphify-Labs/graphify/releases/tag/v0.9.55

PR #3322, opened 2026-09-03
Fix: same-file node-id collisions silently drop definitions (#3302)

### LINKED ISSUE #3302: Python: methods differing only by a leading underscore collide on node id — the public one is silently dropped from the graph

abdulsamie10:
Fixes #3302.

`ids.py` casefolds and strips punctuation so the AST extractor, the semantic
subagents, and the graph builder mint identical ids for the same symbol. The
side effect: distinct declared names in one file can normalize to one id —
`Session` vs `session()`, `_get_connection` vs `get_connection`, `visit_TEXT`
vs `visit_text` — and the per-file `seen_ids` guard kept whichever definition
the walk reached first, silently dropped the rest, re-pointed the dropped
definition's `method`/`contains` edge at its surviving sibling, and credited
the dropped body's calls to that sibling.

Measured impact (sqlalchemy @ `36892d8`, 40,332 definitions):

| | different-raw-name collision drops | dangling `rationale_for` edges | nodes |
|---|---|---|---|
| v8 (`33362d9`) | **112** | 26 | 39,769 |
| this PR | **0** | **0** | 39,767 |

Across the full harness — psf/requests at 6 checkouts, flask, sqlalchemy
(Python), gin (Go), vue-core (TS), ripgrep (Rust) — every different-raw-name
collision inside the census's scope is recovered; vue-core retains 5 whose
surviving occupant is outside it by design (`const` value/arrow bindings,
object-literal methods). Method + per-corpus tables: enumerate every
definition with tree-sitter, extract, check for a node at each definition's
(file, line), classify misses (`measurements/measure.py`,
`measurements/measure_lang.py`; before/after tables and the delta accounting
in `measurement_before_after.txt`). The measurement scripts, per-corpus
reports, and evidence logs referenced in this PR are attached alongside it.

Credit where due: the analysis in #3302 did the heavy lifting — the reporter
measured definition->node coverage across 14 requests checkouts, isolated the
collision class from the by-design nested-def misses, and correctly located
the collapse in the extractor's first-wins guard rather than in `ids.py`.
This PR is that analysis carried to a fix.

## What the fix does

Go's extractor already solved this for its own functions/methods (#2779):
census the raw declared names sharing a plain id, keep the plain id on ONE
canonical member, salt the rest with `sha1(raw_name)[:6]`. This PR lifts that
mechanism into a shared helper (`extractors/collisions.py`) and wires it into
`engine._extract_generic` — covering Python, JS/TS, Java, C, C++, Ruby, C#,
Kotlin, Scala, PHP, Lua, Swift, Groovy at once — plus `rust.py`, and widens
Go's census to `type_spec` (gin's `ResponseWriter` interface /
`responseWriter` struct pair dropped the struct and re-parented its 13
methods onto the interface node).

- `add_node` learns the raw declared name and returns the effective id; the
  walk runs in a bounded mint-pass loop that re-picks each group's canonical
  member once the whole file is known and re-runs the walk with the
  assignment pinned, so derived ids (method ids embed their class's id) are
  re-minted, never patched. Collision-free files settle in one pass.
- Canonical preference: fewest leading underscores, else the unique
  class-like member, else all-salted (`visit_TEXT`/`visit_text`) so an
  ambiguous plain-id reference resolves to nothing rather than an arbitrary
  winner. Go keeps its exported-member preference.
- `ids.py` is untouched — the three-producer contract and its Unicode
  history (#811/#550/#1033/#1104/#2614) are unchanged. Ids stay independent
  of declaration order. Same-raw-name re-definitions (`@overload`,
  `if PY2:`, `#[cfg(...)]`, C# `partial`) stay ONE node, now with a
  `redefinition_lines` audit trail.
- The Python docstring/rationale post-pass resolves targets through the
  emitted nodes instead of re-deriving ids, so a salted sibling's docstring
  lands on that sibling — this also clears the 26 pre-existing dangling
  `rationale_for` edges on sqlalchemy (repro:
  `measurements/rationale_dangling.py`).

## Compatibility with shipped graphs — stated precisely

Some shipped ids move; that is inherent to recovering the dropped siblings,
and the PR handles it rather than denying it:

- **Where ids move.** The canonical member is not always the pre-fix
  survivor: `class Session` declared after `def session()` takes the plain id
  from the function; case-only same-kind pairs retire the plain id; in Go the
  exported type's id is stable but the 13 gin methods mis-parented onto it
  re-key under the recovered unexported type
  (`measurements/node_id_diff.py`, evidence logs in this folder).
- **Incremental `graphify update` repairs stored endpoints.** When a
  re-extracted file re-keys an id, `build_merge` and `merge_raw_extraction`
  re-point stored edges from untouched files onto the sibling their stored
  label names — the salt is a pure function of the raw declared name, so the
  mapping is exact, needs no migration table, and a genuinely deleted
  definition still drops its edges as before. Covered by new regression
  tests in both directions (canonical flip, all-salted, reverse move, raw
  path).
- **A small net node SHRINK is expected on stub-heavy corpora** (sqlalchemy
  −2, ripgrep −3): recovered definitions collapse the per-file sourceless
  reference stubs (#1402) that stood in for them — sqlalchemy gains 113
  sourced definitions and loses 115 stubs. A non-force rewrite of an
  existing graph.json therefore trips the #479 shrink guard by design; the
  guard's message now names this cause. **The upgrade path is a full rebuild:
  `graphify extract . --force`** (verified end to end on sqlalchemy:
  `measurements/e2e_shrink.py`).
- **`_AST_CACHE_SCHEMA` bumped to 3** — the extractor emits different nodes
  for identical file content, so without the bump a warm `graphify-out`
  silently serves pre-fix collapsed nodes and the fix looks like a no-op.

One knock-on worth knowing about (disclosed, not fixed here): ripgrep's
recovered `#[cfg(test)] fn E` widens what the pre-existing name-only #1402
stub rewire can match, so 10 `references` edges from generic type parameters
named `E` in other crates now land on that unrelated helper
(`measurements/ripgrep_e_refs.py`). That is the existing rewire's
name-matching being fed a definition it could not see before; tightening the
rewire is a separate issue.

## Tests

- `tests/test_same_file_id_collisions.py` (22 tests): the issue's repro,
  case/underscore/all-salted branches, order-independence and determinism,
  TS/Rust/`redefinition_lines` coverage, LLM plain-id reconciliation,
  docstring-rationale routing (salted sibling + no-dangling), incremental
  merges over a pre-fix graph in all four directions, the shrink-guard
  refusal naming this cause, and the `--no-cluster` raw path.
- `tests/test_go_qualified_resolution.py` (13 tests): #2779 behavior
  preserved, widened `type_spec` census, cross-package stability.
- Full suite: 5188 passed, 92 skipped; the 15 failures are pre-existing
  environment issues (11 `test_skillgen` need a git object absent from this
  clone, 4 `test_ollama_retry_cap` need the optional `openai` module) and
  are identical on the base commit. One unrelated test
  (`test_labeling.py::test_label_communities_batches_when_over_batch_size`)
  is order-flaky on the base commit and on this branch alike (its batching
  assertion is scheduling-dependent).


abdulsamie10:
Good catch — fixed in the follow-up commit.

You're right on both counts. The comment claimed "old collapse" but the code returned `salted`, which is not the pre-#3302 behaviour: whatever holds the salted id is an unrelated symbol, so returning it would attribute this definition's edges and body calls to a node it has nothing to do with, and callers would record a `redefinition_lines` entry for a raw name that never repeated.

`assign()` now returns `(plain_nid, True)` in that branch — genuine first-wins on the plain id, matching what the comment said all along.

Added `test_salt_truncation_collision_falls_back_to_the_plain_id` to cover the path, since it had none. It asserts the effective id is the plain one and explicitly that it is not the salt; it fails on the previous fallback and passes now. Full suite: 5,189 passed (the 15 failures are pre-existing on `v8` at `33362d9` — `test_skillgen.py` and `test_ollama_retry_cap.py`, byte-identical to the base commit's).

abdulsamie10:
Thanks — I've done the human review the four `Escalate · high` findings ask for. All four are false positives, and I checked each against the tree rather than by reading:

**`str_path` referenced but never defined** (`extract.py:1216`, `:1220`) — it is defined. `str_path = str(path)` is assigned at **line 1201**, inside the same function (`_extract_python_rationale`, lines 1185–1327), and every use is after it:

```
1201:     str_path = str(path)
1218:         if _n.get("source_file") != str_path or _n.get("id") == file_nid:
1266:                 "source_file": str_path,
1274:             "source_file": str_path,
```

Exercised directly on a file with a `_go`/`go` docstring pair: 4 nodes, no `NameError`.

**`passes()` calls undefined `_settle`** (`collisions.py:222`) — `_settle` is defined at line 245 and called at 195, both on `SymbolCollisionCensus`. Every `passes()` loop in the suite runs it.

**Typo makes test file invalid Python** — `python -m py_compile tests/test_same_file_id_collisions.py` is clean, and the file's 23 tests pass.

I think these three share a cause: each name is defined outside the diff hunk that mentions it, so a reviewer reading only the patch context sees a use with no definition.

**Public collision helper return type changed** (medium) — this one points at something real, but not in `collisions.py`: that file is new in this PR, so nothing in it can have *changed*. What changed is `add_node(nid, label, line) -> None` → `-> str` in `engine.py` and `go.py`. Both are closures defined inside `_extract_generic` / `extract_go`, so neither is importable or part of the public surface, and the new contract is documented on `add_node` itself — definition nodes **must** use the returned id for their edges, bodies and registries, which is what stops a salted sibling's calls being credited to its neighbour. `go.py` also loses its local `symbol_nid` / `_scan_declarations` / `_plain_symbol_nid` / `_receiver_type_of` helpers, which is the point of the change: Go's #2779 census becomes the shared one.

On the coupling notes: `engine.walk()` and `go.walk()` were already the largest functions in their modules; this adds the census call to each. Happy to extract that into a helper if you'd prefer the fan-out unchanged.

Full suite on this branch: **5,189 passed**. The 15 failures are pre-existing on `v8` at `33362d9` (`test_skillgen.py`, `test_ollama_retry_cap.py`) — byte-identical to the base commit's, unrelated to this change.

abdulsamie10:
Checking in on this after `v8` moved on — and with a proposal rather than a rebase, because part
of this PR is now redundant.

`343f76f` landed `_python_pre_scan_underscore_collisions` and closed #3302. That covers the case
the issue was filed about, and this PR should not re-litigate it.

What that fix deliberately scopes out, per its own docstring, is still open:

- **Python only.** Go, Rust and the config-driven languages still collapse. Measured examples from
  my original survey: `gin` — `type ResponseWriter interface` vs `type responseWriter struct`
  (the unexported struct has no node, and its methods attach to the interface's); `ripgrep` —
  `fn ignore` vs `struct Ignore`, plus `#[cfg(unix)]`/`#[cfg(windows)]` twins; `vue-core` —
  `ontext` (tokenizer callback) vs `onText` (method).
- **Underscore/dunder only.** The casefold class is untouched: `Session` (class) vs `session()`
  (factory) in `requests/sessions.py`, and in sqlalchemy `visit_TEXT`/`visit_text`,
  `visit_ENUM`/`visit_enum`, `Insert`/`insert()`, `ARRAY`/`array`. In my corpus run these were
  **112 of the distinct-symbol collisions in sqlalchemy alone**.
- **Module-level scope only** — nested classes and nested functions are explicitly not scanned.

So the remaining gap is real, but this branch as it stands is the wrong shape for it: it is
+1651/−123 against a tree that has moved 214 commits, and it now overlaps the landed fix.

Rather than force a large rebase that partly duplicates `343f76f`, I would rather cut it down to
just what is still missing — the shared census wired into `_extract_generic`, Rust, and Go's
`type_spec`, with the Python underscore path deferring to your implementation — as a smaller,
reviewable PR against current `v8`.

Happy to do that if it is wanted. If you would rather keep the surface Python-only for now, say so
and I will close this one; no hard feelings either way, and the investigation notes are yours to
use regardless.

Issue #1764, opened 2026-07-10
json_config extractor emits `imports`/`extends` edges to node IDs it never creates (package.json, tsconfig.json)


Stashub:
## Summary

`extract_json` emits `imports` and `extends` edges whose target node is never created. `build_from_json` then drops those edges without a warning, because the "does not match any node id" error is explicitly filtered out of `real_errors`. The result is silent edge loss on two of the most common files in any JS/TS repo: `package.json` and `tsconfig.json`.

Nothing surfaces this except `graphify.diagnostics.diagnose_extraction`, which reports them as `dangling_endpoint_edges` — and by then the edges are already gone from `graph.json`.

## Version

Reproduced on **0.9.11**. Python 3.14.

## Reproduction

Two files in an empty directory:

```jsonc
// package.json
{
  "name": "demo",
  "dependencies":    { "left-pad": "^1.3.0" },
  "devDependencies": { "bats": "^1.11.0" }
}
```

```jsonc
// tsconfig.json
{
  "extends": "./tsconfig.base.json",
  "compilerOptions": { "strict": true }
}
```

```python
from pathlib import Path
from graphify.extract import extract

for f in ("package.json", "tsconfig.json"):
    r = extract([Path(f)], cache_root=Path("."))
    ids = {n["id"] for n in r["nodes"]}
    dangling = [e for e in r["edges"] if e["source"] not in ids or e["target"] not in ids]
    print(f, len(r["nodes"]), "nodes", len(r["edges"]), "edges", len(dangling), "dangling")
    for e in dangling:
        print("   ", e["source"], "--" + e["relation"] + "-->", e["target"])
```

## Actual

```
package.json  6 nodes 7 edges 2 dangling
    package_dependencies_left_pad --imports--> left_pad
    package_devdependencies_bats --imports--> bats
tsconfig.json 4 nodes 4 edges 1 dangling
    tsconfig --extends--> ref_tsconfig_base_json
```

Nodes `left_pad`, `bats`, `ref_tsconfig_base_json` do not exist. At build time all three edges disappear from the graph and no warning is printed.

## Cause

`graphify/extractors/json_config.py`:

```python
# ~line 176 — `extends`
ref_nid = _make_id("ref", val_text)
if ref_nid:
    add_edge(file_nid, ref_nid, "extends", line, context="import")

# line 187-189 — dependency keys
elif parent_key in _DEP_KEYS and val_text:
    dep_nid = _make_id(key)
    if dep_nid:
[...]

Issue #1638, opened 2026-07-03
Unresolved bare npm import gets aliased onto an unrelated same-named local file (cross-language phantom edge)


EveX1:
### Environment
- graphify (`graphifyy` on PyPI) version: 0.9.5
- Installed via: `uv tool install graphifyy`
- Corpus: a monorepo with a Python backend and a TypeScript/React frontend (found while running the Claude Code `/graphify` skill against a private repo — repo details are not shareable, b
ut the bug is 100% reproducible with a synthetic repo, see below)

### Summary
When a TS/TSX file has a bare import from a third-party npm package whose last path segment happens to match the filename (no extension) of an unrelated local file elsewhere in the repo, g
raphify emits a phantom `imports_from` edge from the TS file to that unrelated local file — even when the two files are in completely different languages/directories and have no real relat
ionship.

### Minimal repro (synthetic repo layout)
```
repo/
  backend/
    utils/
      colors.py        # defines color_variant(), hex_to_rgb(), rgb_to_hex()
  frontend/
    src/
      SomeChart.tsx     # imports "colors" from a THIRD-PARTY npm package
```
`frontend/src/SomeChart.tsx`:
```ts
import colors from "tailwindcss/colors";   // bare import of an npm package, unrelated to backend/utils/colors.py

const CHART_COLOR = colors.blue[500];
```
`backend/utils/colors.py`:
```python
def hex_to_rgb(value: str) -> tuple[int, int, int]:
    ...
```

Neither file has any real relationship to the other. After a full `graphify` build:
```
$ graphify explain "colors.py"
Connections:
  <-- SomeChart.tsx [imports_from] [EXTRACTED]
  ... (and any other .tsx file in the repo with `import ... from "<any-npm-pkg>/colors"`)
```
In the real-world case that surfaced this, **eight** unrelated `.tsx` files across one feature area all showed up as `imports_from` sources on a single unrelated Python module, purely beca
use they all did `import colors from "tailwindcss/colors"`. The edges are marked confidence `EXTRACTED` — the highest tier, meaning "explicit in source" — so there is no signal in the outp
ut that lets a consumer suspect they're wrong.

### Root cause (traced in the installed package)
Two independent pieces combine to create the collision:

**1. `extract.py`, `_resolve_js_import_target()`** (around line 1759):
```python
def _resolve_js_import_target(raw: str, str_path: str) -> "tuple[str, Path | None] | None":
    if not raw:
        return None
[...]

PR #3330, opened 2026-09-04
fix: spurious json_config edges, and a claude-cli envelope parse failure that discards paid-for chunks

pranavshipit:
Two independent fixes, one commit each — happy to split them into separate PRs or have you cherry-pick just one.

Found while evaluating graphify on a ~500-file Next.js + Supabase repo. All numbers below are measured on that repo or on the minimal fixtures added here.

---

## 1. `json_config` emits spurious `extends` and self-referential `imports` edges

Two defects in `graphify/extractors/json_config.py`, both in the shape introduced by #1764. That change fixed a real dangling-edge bug by creating the missing target as a `concept` node; these are follow-on consequences of the node it mints, not a reason to revert it.

**a. Every array became `extends`.** The array branch carried the comment "For `extends` arrays (tsconfig, eslint)" but had no key check, so it fired for any array under a tracked key:

```
compilerOptions.lib: ["dom", "esnext"]     ->  lib     --extends--> dom, esnext
exclude: ["node_modules", ".next"]         ->  exclude --extends--> node_modules, .next
```

Array membership is recorded as a supertype relation. On the test repo this also manufactured a false hub: `.next` reached degree 25 and ranked 24th by centrality, built from the tsconfig exclude glob plus 24 misresolved bare `next` imports. The string branch directly below already gates on `key == "extends"`; this makes the array branch match it.

**b. Every dependency became a self-loop.** The dependency edge ran `key_nid -> dep_nid`, and both nodes carry the dependency's own label:

```
react --imports--> react          (22 of these on the test repo)
```

`test_extract_json_import_and_extends_targets_are_real_nodes` already forbids self-loops, but it compares node *ids*, and the minted duplicate has a different id with the same label, so the guard cannot see it. Sourcing the edge at the manifest gives `package.json --imports--> react`.

The old target also used a bare `_make_id(key)`, skipping the `"ref"` namespacing every other external reference in this file applies (J-4). A dependency named `utils`, `colors` or `types` could then be collapsed onto a same-named local module by `build.py`'s alias index — the #1638 failure mode reached through a different door. Now `_make_id("ref", key)`.

**Effect on a 3-file fixture (`package.json` + `tsconfig.json` + one `.ts`):**

| | nodes | edges | spurious |
|---|---|---|---|
| before | 21 | 18 | 7 (39%) |
| after | 17 | 15 | 0 |

---

## 2. `claude-cli` backend dies on a diagnostic preamble, after paying for the answer

`_claude_cli_envelope` did `json.loads(stdout)`. But `claude -p --output-format json` shares stdout with the CLI's own subsystems, so the envelope is not always first on it. An attached MCP server that advertises no tools makes every invocation emit:

```
Client.listTools() called but server does not advertise tools capability - returning empty list
{"is_error":false,"duration_api_ms":472526,...}
```

and the parse fails on the whole buffer.

The failure is raised *after* the model has answered, so the chunk is discarded with its tokens already spent. In my run this burned 149,582 input and 52,162 output tokens on chunk 1 of 3 and returned nothing. The error message names the JSON rather than the preamble that caused it, so the log points at the wrong thing.

Anyone running an MCP server hits this on every chunk, which makes the `claude-cli` backend unusable for a large share of Claude Code users.

`_envelope_after_preamble` scans for the first `[`/`{` that begins a valid JSON document. `raw_decode` ignores trailing bytes, so a diagnostic on either side is tolerated. The fast path is unchanged — clean stdout still parses on the first `json.loads` — and stdout containing no JSON still raises, so `test_raises_on_garbage_envelope` holds.

---

## Tests

Four added, each verified to fail on the unpatched tree and pass with the fix:

- `test_extract_json_dependency_edges_are_not_label_self_loops` — closes the id-vs-label gap the existing self-loop guard leaves open, and asserts dependency targets stay `ref`-namespaced
- `test_extract_json_non_extends_arrays_are_not_inheritance`
- `test_envelope_survives_a_diagnostic_preamble_on_stdout` — uses the real MCP notice
- `test_envelope_survives_preamble_around_array_shaped_stream` — the >= 2.1 array-of-events shape, with trailing noise

**No existing test was modified.**

Full suite on `v8` @ 0.9.53: **13 failures before and after — the identical set** (`test_skillgen`, `test_terraform`, ollama retry-cap, and two others, all pre-existing on a clean checkout). Passing goes 4,925 -> 4,929, which is exactly the four new tests.

---

## Noted, not fixed here: SQL `GRANT`/`REVOKE` produce no edges

Out of scope for this PR — it needs new edge relations, which is your call, not a patch. Reporting it because it is the highest-consequence thing I found.

**Reproduced minimally.** This file yields exactly one edge, `file --contains--> public.secured_fn()`:

```sql
CREATE OR REPLACE FUNCTION public.secured_fn(p_id uuid)
RETURNS uuid LANGUAGE plpgsql SECURITY DEFINER SET search_path TO 'public'
AS $function$
begin
  return p_id;
end
$function$;

REVOKE ALL ON FUNCTION public.secured_fn(uuid) FROM public, anon;
GRANT EXECUTE ON FUNCTION public.secured_fn(uuid) TO authenticated, service_role;
```

Neither the `REVOKE` nor the `GRANT` is represented. On the repo I was testing, 455 grant/revoke statements and 29 `CREATE POLICY` statements produced zero edges between them.

That matters more than ordinary missing coverage, because "is this function reachable by the anonymous role" is a question people ask a codebase graph, and in Supabase/PostgREST projects the answer is enforced in CI. A graph that cannot see `REVOKE` cannot answer it, and silence reads as "no exposure".

**One observation I could not reduce to a minimal case.** On that same repo the graph contains:

```
public.ron_append_message() --reads_from--> anon    [EXTRACTED, confidence 1.0]
```

The only occurrences of `anon` in that migration are `REVOKE ... FROM public, anon` lines — the source restricts the role, and the graph records a dependency on it, at full confidence. My best guess was that `FROM public, anon` in a `REVOKE` is being read as a query's `FROM` clause, but I could not trigger it in a small file (I tried the named `$function$` dollar-quote, upper and lower case, and the exact signature), so I am reporting the observation rather than a diagnosis. Happy to dig further, or to send the migration privately if that is useful.

If you would take a PR that adds `grants_execute` / `revokes_from` relations, I am glad to write it — say the shape you want and I will follow it.

### DIFF
--- a/graphify/extractors/json_config.py
+++ b/graphify/extractors/json_config.py
@@ -161,8 +161,14 @@ def extract_json(path: Path) -> dict:
             if val.type == "object":
                 walk_object(val, key_nid, key, depth + 1, pair_count)
 
-            elif val.type == "array":
-                # For "extends" arrays (tsconfig, eslint): each string element.
+            elif val.type == "array" and key == "extends":
+                # Only a genuine `extends` array (tsconfig 5.0+, eslint) is
+                # inheritance. This branch used to fire for *every* array under
+                # a tracked key, so `compilerOptions.lib: ["dom","esnext"]` and
+                # `exclude: ["node_modules",".next"]` each emitted an `extends`
+                # edge — array membership relabelled as inheritance. The string
+                # branch below already gates on `key == "extends"`; this matches
+                # it. Non-`extends` arrays keep their `contains` edge only.
                 # Prefix with "ref_" so external refs don't collide with real
                 # code/file node IDs that share the same collapsed _make_id (J-4).
                 for item in val.children:
@@ -193,10 +199,28 @@ def extract_json(path: Path) -> dict:
                         add_edge(parent_nid, ref_nid, "references", line)
 
                 elif parent_key in _DEP_KEYS and val_text:
-                    dep_nid = _make_id(key)
+                    # Two defects fixed here, both from the #1764 shape:
+                    #
+                    # 1. The edge ran `key_nid -> dep_nid`, and both nodes carry
+                    #    the *same label* (the dependency name), so every
+                    #    dependency produced a visually self-referential
+                    #    `react --imports--> react` (22 of them on a mid-size
+                    #    Next.js repo). The self-loop guard in
+                    #    test_extract_json_import_and_extends_targets_are_real_nodes
+                    #    compares node ids, and these two ids differ, so the
+                    #    duplicate slipped past it. Sourcing the edge at the
+                    #    manifest yields `package.json --imports--> react`.
+                    #
+                    # 2. `_make_id(key)` minted a *bare* id, skipping the "ref"
+                    #    namespacing every other external reference in this file
+                    #    applies (J-4). A dependency named `utils`, `colors` or
+                    #    `types` could then be collapsed onto a same-named local
+                    #    module by build.py's alias index -- the #1638 failure
+                    #    mode reached through a different door.
+                    dep_nid = _make_id("ref", key)
                     if dep_nid:
                         add_node(dep_nid, key, line, file_type="concept")
-                        add_edge(key_nid, dep_nid, "imports", line, context="import")
+                        add_edge(file_nid, dep_nid, "imports", line, context="import")
 
     # Entry: find root document → object
     doc = root
--- a/graphify/llm.py
+++ b/graphify/llm.py
@@ -1524,6 +1524,40 @@ def _call_claude(api_key: str, model: str, user_message: str, max_tokens: int = 
     return result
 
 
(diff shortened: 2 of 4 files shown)


safishamsi:
Shipped in **v0.9.55** — landed on `v8` via cherry-pick with your authorship preserved. Thanks @pranavshipit — json_config no longer emits array-as-extends/self-loop edges, and claude-cli tolerates a diagnostic preamble before the JSON envelope. Release: https://github.com/Graphify-Labs/graphify/releases/tag/v0.9.55

PR #3339, opened 2026-09-04
fix(mcp): prevent PR tools from timing out on stdio transport

### LINKED ISSUE #3318: MCP stdio server: PR tools time out because children inherit the never-EOF stdin pipe and git blocks on it (+ `gh repo view --repo` is not a valid flag)
## Summary

On the stdio MCP transport, `list_prs`, `triage_prs` and `get_pr_impact` always fail with a client-side "Request timed out", while every graph tool (`query_graph`, `get_node`, `graph_stats`, …) works normally in the same server process. The cause is not `gh`, auth, or the network — it is that `graphify/prs.py` spawns children with **inherited stdin**, and `serve._filter_blank_stdin()` has replaced fd 0 with an anonymous pipe that never reaches EOF. `git` blocks on that handle at startup, so `git symbolic-ref` hangs, and so does every *repo-scoped* `gh` call, because `gh` shells out to `git` to resolve the current repository.

A second, independent bug in `_detect_default_branch` means passing `repo=` does not avoid the hang.

Environment: graphifyy **0.9.41**, Windows 11, Python 3.13, `gh` 2.92.0 (authenticated, keyring), Git 2.53.0.windows.2, stdio transport (`graphify-mcp.exe <graph.json>`).

## Measurements (same server process, instrumented `_gh`)

| child command | inherited stdin | `stdin=subprocess.DEVNULL` |
|---|---|---|
| `git --version` | 10s → `TimeoutExpired` | **0.02s ✓** |
| `C:\Program Files\Git\mingw64\bin\git.exe --version` (absolute) | 10s → `TimeoutExpired` | **0.02s ✓** |
| `gh --version` | 0.05s ✓ | ✓ |
| `gh auth status` | 0.69s ✓ | ✓ |
| `gh api rate_limit` | 0.47s ✓ | ✓ |
| `cmd /c echo hi` | 0.01s ✓ | ✓ |

`gh` itself, its auth and the network are all fine — only children that touch `git` hang. Same result when the probe runs on a worker thread instead of the event-loop thread, so it is not asyncio starvation. Outside the MCP process (plain shell, plain Python child with `stdin=PIPE`, cwd `C:\Windows\System32`) `git --version` returns in 0.02s, so it is specific to the server process.

## Resulting 65s chain for `list_prs`

```
_detect_default_branch → gh repo view --json defaultBranchRef   30s timeout (gh waits on git)
                       → git symbolic-ref refs/remotes/origin/HEAD    5s timeout
fetch_prs              → gh pr list --state open ...            30s timeout (gh waits on git)
                                                          total  65s  →  MCP client gives up
```

`_gh` maps `TimeoutExpired` to `None`, so when the client timeout is long enough to see it, the message is also misleading: `gh CLI not found or not authenticated. Run: gh auth login` (related to the already-closed #3162).

## Bug 2 — `gh repo view --repo X` is rejected

`_detect_default_branch` builds `["repo", "view", "--json", "defaultBranchRef", "--repo", repo]`:

```
rc=1  unknown flag: --repo
Usage:  gh repo view [<repository>] [flags]
```

`gh repo view` takes the repository **positionally**. So even when the caller passes `repo=`, default-branch detection fails and falls through to the hanging `git symbolic-ref`. (`gh pr list --repo X` is correct and returns in 0.79s — with an explicit repo `gh` never needs git, which is why only that one call was unaffected.)

## Hypothesis for the git block

`_filter_blank_stdin()` does `os.pipe()` + `os.dup2(r_fd, 0)` and keeps the write end open in a relay thread, so fd 0 is a synchronous (non-overlapped) pipe with a reader thread parked in `ReadFile`. Git for Windows' startup queries stdin for msys/tty detection, and on such a pipe that query appears to serialize behind the pending read. This is consistent with `gh --version`/`cmd` not caring, and with a standalone replication of the filter *without* a blocked reader on fd 0 showing no hang. Offered as a hypothesis only — the fix does not depend on it.

## Suggested fix

Pass `stdin=subprocess.DEVNULL` at every `subprocess.run` in `graphify/prs.py` — none of these children need stdin:

- `_gh` (covers `gh repo view`, `gh pr list`, `gh pr view`)
- the `git symbolic-ref` fallback in `_detect_default_branch`
- `fetch_pr_files`
- `fetch_worktrees`

and make the repo positional:

```python
# gh repo view takes the repo POSITIONALLY; "--repo X" is rejected as an unknown flag.
args = ["repo", "view", *([repo] if repo else []), "--json", "defaultBranchRef"]
```

`export.py:177` (`git rev-parse HEAD`, timeout 3) has the same shape; it is not reachable from a server handler today, but it would hang if it ever were.

Hardening `_filter_blank_stdin` itself (or not touching fd 0 at all) would fix the class rather than the instances, but `stdin=DEVNULL` is the smaller, safer change.

## Verification

With those five edits applied locally to 0.9.41, against a repo that has an open PR, over a fresh stdio MCP server:

| tool | before | after |
|---|---|---|
| `list_prs` | 65s → timeout | **1.38s**, returns the PR |
| `triage_prs` | timeout | **1.36s** |
| `get_pr_impact` | timeout | **1.58s**, title/CI/base/files + graph impact |

hopstreax:
## Summary

Fixes #3318.

PR-related MCP tools (`list_prs`, `triage_prs`, and `get_pr_impact`) could time out when Graphify was running over the stdio transport.

The MCP server replaces stdin with a pipe that remains open for the lifetime of the process. PR-related `git`/`gh` subprocesses inherited that stdin, and Git for Windows could block while inspecting the handle. This caused the subprocess timeouts to cascade into MCP client request timeouts.

This PR also fixes an independent `gh repo view` argument bug where `--repo` was incorrectly used even though `gh repo view` expects the repository as a positional argument.

## Changes

* Pass `stdin=subprocess.DEVNULL` to non-interactive PR subprocesses:

  * `_gh`
  * `_detect_default_branch` Git fallback
  * `fetch_pr_files`
  * `fetch_worktrees`
* Fix `gh repo view` to pass the repository positionally.
* Prevent an explicitly supplied remote `repo` from falling back to the local repository's Git branch.
* Add regression tests for stdin isolation and repository argument handling.

## Why

Before this fix, `list_prs` could follow this timeout chain:

```text
_detect_default_branch
  → gh repo view       30s timeout
  → git symbolic-ref    5s timeout

fetch_prs
  → gh pr list         30s timeout

≈ 65s total → MCP client request timeout
```

The issue was specific to the MCP stdio process environment. `gh` authentication and network access were working normally.

With child stdin redirected to `DEVNULL`, the PR tools no longer inherit the MCP transport's open stdin pipe.

## Verification

* `pytest tests/test_prs.py` → **52 passed**
* `pytest tests/test_serve_http.py tests/test_serve.py` → **146 passed, 1 skipped**
* `git diff --check` → clean

No changes were made to `serve.py`, `export.py`, or unrelated subprocess paths.

### DIFF
--- a/graphify/prs.py
+++ b/graphify/prs.py
@@ -142,6 +142,7 @@ def _gh(*args: str) -> list | dict | None:
     try:
         result = subprocess.run(
             ["gh", *args],
+            stdin=subprocess.DEVNULL,
             # Decode gh's output as UTF-8, not the Windows cp1252 locale codec: gh
             # emits UTF-8 JSON with non-Latin1 titles/logins (emoji, فارسی), and the
             # default text=True decode crashes on those (#1505 fixed the same in llm).
@@ -157,24 +158,27 @@ def _gh(*args: str) -> list | dict | None:
 def _detect_default_branch(repo: str | None = None) -> str:
     """Auto-detect the repo's default branch via gh, then git, then fall back to 'main'."""
     # Try gh first — works for any repo, not just the current directory
-    args = ["repo", "view", "--json", "defaultBranchRef"]
+    args = ["repo", "view"]
     if repo:
-        args += ["--repo", repo]
+        args.append(repo)
+    args += ["--json", "defaultBranchRef"]
     data = _gh(*args)
     if data and data.get("defaultBranchRef", {}).get("name"):
         return data["defaultBranchRef"]["name"]
-    # Fall back to git symbolic-ref for the current repo
-    try:
-        result = subprocess.run(
-            ["git", "symbolic-ref", "refs/remotes/origin/HEAD"],
-            capture_output=True, text=True, encoding="utf-8", errors="replace", timeout=5
-        )
-        if result.returncode == 0:
-            # refs/remotes/origin/main → main
-            ref = result.stdout.strip()
-            return ref.split("/")[-1] if ref else "main"
-    except (subprocess.TimeoutExpired, FileNotFoundError):
-        pass
+    # Fall back to git symbolic-ref for the current repo (only when repo is not specified)
+    if not repo:
+        try:
+            result = subprocess.run(
+                ["git", "symbolic-ref", "refs/remotes/origin/HEAD"],
+                stdin=subprocess.DEVNULL,
+                capture_output=True, text=True, encoding="utf-8", errors="replace", timeout=5
+            )
+            if result.returncode == 0:
+                # refs/remotes/origin/main → main
+                ref = result.stdout.strip()
+                return ref.split("/")[-1] if ref else "main"
+        except (subprocess.TimeoutExpired, FileNotFoundError):
+            pass
     return "main"
 
 
@@ -232,7 +236,11 @@ def fetch_pr_files(number: int, repo: str | None = None) -> list[str]:
     if repo:
         args += ["--repo", repo]
     try:
-        result = subprocess.run(["gh", *args], capture_output=True, text=True, encoding="utf-8", errors="replace", timeout=30)
+        result = subprocess.run(
+            ["gh", *args],
+            stdin=subprocess.DEVNULL,
+            capture_output=True, text=True, encoding="utf-8", errors="replace", timeout=30
+        )
         if result.returncode != 0:
             return []
         return [l.strip() for l in result.stdout.splitlines() if l.strip()]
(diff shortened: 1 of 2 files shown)


safishamsi:
Shipped in **v0.9.55** — landed on `v8` via cherry-pick with your authorship preserved. Thanks @hopstreax — the MCP prs tools no longer hang the stdio transport (subprocess stdin detached). Release: https://github.com/Graphify-Labs/graphify/releases/tag/v0.9.55

PR #3345, opened 2026-09-05
fix(build): ghost-merge nodes whose source_file is the referencing doc, not the referenced file

### LINKED ISSUE #3344: Ghost-merge misses semantic nodes whose source_file is the referencing doc, not the referenced file
## Summary

`build_from_json()`'s ghost-merge pass (the "#1145-extended" logic that collapses a semantic/LLM-extracted duplicate into its AST-canonical twin) only matches on an exact `(source_file, label)` key. When a semantic extraction pass reads a document that merely **mentions** a real source file by name in prose (a saved `graphify-out/memory/*.md` query answer, a runbook like `deploy/scripts/provision-host.md`, any doc narrating "see `App.tsx`" or "`backup.sh` does X"), the new node it creates gets stamped with `source_file` = *the document being read*, not the file actually named in the text. That wrong `source_file` means the `(source_file, label)` key can never match the real AST node's key, so the ghost-merge never fires — a permanent duplicate node survives every subsequent `--update`/rebuild, forever.

Concretely, on a real project (NestJS/Next.js/Expo monorepo, ~4600 nodes), this produced ghosts like:
- label `App.tsx`, `source_file` = a saved memory query file — instead of merging into the real AST node for `apps/customer-app/App.tsx`
- label `customer-app/index.ts` (a doc summary dropped the `apps/` root segment) — instead of merging into `apps/customer-app/index.ts`
- label `backup.sh`/`deploy.sh`/`init-buckets.sh`/`restore.sh`, `source_file` = `deploy/scripts/provision-host.md` — instead of merging into the real script file nodes

These ghosts then poison hyperedges (e.g. a `weak_bridge_chain_to_hub` hyperedge citing 3 real nodes + 3 of these phantoms) and community/god-node analysis with fake, unbacked entities.

## Root cause

In `build.py`'s ghost-merge (inside `build_from_json`), Pass 1/Pass 2 key exclusively on `(source_file, label)`:

```python
key = (sf, label)
if is_ast:
    ...
    _loc_nodes[key] = nid
else:
    _loc_nodes.setdefault(key, nid)
```
```python
# Pass 2: find ghosts — non-AST nodes that have an AST canonical twin.
for nid in sorted(node_set):
    ...
    key = (sf, label)
    if key in _loc_collisions:
        continue
    if key in _loc_nodes and _loc_nodes[key] != nid:
        _noloc_nodes[key] = nid
```

This is correct for genuine same-file re-extraction ghosts (the original #1145 case), but it silently does nothing for the "referenced-not-defined" case above, because the ghost's own `source_file` is simply wrong — it's the *referencing* document, not the *referenced* file.

The codebase already has the exact predicate needed to catch this by label alone: `_is_file_node_label(label, source_file)` (used elsewhere for file-node label disambiguation) returns `True` when `label` is the bare basename of `source_file`, or a directory-qualified suffix of it (handles a doc dropping a leading path segment, e.g. `customer-app/index.ts` vs the real `apps/customer-app/index.ts`). It just isn't wired into the ghost-merge matching at all.

## Suggested fix

Add a conservative "Pass 2b" that falls back to matching a ghost by label alone against every AST **file-self node** (an AST node whose own label already names its own file — i.e. `_is_file_node_label(ast_label, ast_source_file)` is true), and only remaps when the match is **unique** (0 or 2+ candidates → leave the ghost alone, same conservatism `_loc_collisions` already applies for the exact-key ambiguity case).

```python
# In Pass 1, alongside `_loc_nodes[key] = nid` (is_ast branch), also collect:
_ast_file_nodes: list[tuple[str, str]] = []  # (node_id, source_file)
...
if is_ast:
    ...
    _loc_nodes[key] = nid
    if _is_file_node_label(label, sf):
        _ast_file_nodes.append((nid, sf))
```

```python
# After the existing exact-key _ghost_remap is built, before the removal loop:
if _ast_file_nodes:
    for nid in sorted(node_set):
        if nid in _ghost_remap:
            continue  # already resolved by the exact (sf, label) key
        attrs = G.nodes[nid]
        if attrs.get("_origin") == "ast":
            continue
        label = str(attrs.get("label", "")).strip()
        if not label:
            continue
        matches = {
            ast_id for ast_id, ast_sf in _ast_file_nodes
            if _is_file_node_label(label, ast_sf)
        }
        if len(matches) == 1:
            _ghost_remap[nid] = next(iter(matches))
```

No other changes needed — `_ghost_remap` already flows into the existing `norm_to_id` map that both edge endpoints *and* hyperedge members resolve through, so this Just Works for both.

## Verification

**Unit test (positive — unique match merges correctly, edges rewire to the real node):**
```python
extraction = {
    'nodes': [
        {'id': 'apps_customer_app_app', 'label': 'App.tsx', '_origin': 'ast', 'file_type': 'code',
         'source_file': 'apps/customer-app/App.tsx', 'source_location': 'L1'},
        {'id': 'apps_customer_app_index', 'label': 'customer-app/index.ts', '_origin': 'ast', 'file_type': 'code',
         'source_file': 'apps/customer-app/index.ts', 'source_location': 'L1'},
        {'id': 'app', 'label': 'App.tsx', '_origin': 'semantic', 'file_type': 'code',
         'source_file': 'graphify-out/memory/query_fake.md', 'source_location': None},
        {'id': 'customer_app_index', 'label': 'customer-app/index.ts', '_origin': 'semantic', 'file_type': 'code',
         'source_file': 'graphify-out/memory/query_fake.md', 'source_location': None},
        {'id': 'some_query_node', 'label': 'Query: ...', '_origin': 'semantic', 'file_type': 'document',
         'source_file': 'graphify-out/memory/query_fake.md', 'source_location': None},
    ],
    'edges': [
        {'source': 'some_query_node', 'target': 'app', 'relation': 'references', 'confidence': 'EXTRACTED'},
        {'source': 'some_query_node', 'target': 'customer_app_index', 'relation': 'references', 'confidence': 'EXTRACTED'},
    ],
    'hyperedges': [],
}
G = build_from_json(extraction, root=None)
# Before fix: nodes include 'app' and 'customer_app_index' as permanent ghosts.
# After fix: G.nodes() == {'apps_customer_app_app', 'apps_customer_app_index', 'some_query_node'}
# and the edges correctly point at the real AST nodes.
```

**Unit test (negative — ambiguous basename correctly left alone):** two real files both named `index.ts` in different directories + a phantom labeled bare `index.ts` → the phantom is **not** merged into either (0 or 2+ candidates is the same "no safe winner" rule the existing exact-key path already uses).

**Real-world**: applied to an existing ~4600-node project graph — **10 long-standing ghost nodes removed** (4578 → 4568 nodes) across two separate documents (a saved memory query file, and an unrelated ops runbook), confirming this isn't a narrow edge case but a general failure mode of the "document mentions a file by name" pattern. The two hyperedges that had cited the phantoms now correctly cite only real, AST-backed nodes. Graph health diagnostic clean before and after.

## Environment
- `graphifyy` 0.9.53, installed via `uv tool install`
- Reproduced via the `/graphify` skill's standard `--update` workflow (`build_merge` → `build` → `build_from_json`)

leninherrera94:
Fixes #3344.

## Problem

`build_from_json()`'s ghost-merge pass (the "#1145-extended" logic collapsing a semantic/LLM-extracted duplicate into its AST-canonical twin) keys exclusively on `(source_file, label)`. When a semantic pass reads a document that only **mentions** a real source file by name in prose — a saved `graphify-out/memory/*.md` query answer, a runbook narrating "see `App.tsx`" or "`backup.sh` does X" — the node it mints gets `source_file` = *the document being read*, not the file actually named in the text. That wrong `source_file` means the key can never match the real AST node's key, so the ghost survives every subsequent rebuild, forever.

## Fix

Add a conservative label-only fallback ("Pass 2b") that reuses the already-existing `_is_file_node_label()` predicate (currently only used for file-node label disambiguation) to match a ghost against every AST file-self node collected in Pass 1, when the exact `(source_file, label)` key doesn't resolve it. Only remaps when the match is **unique** across the graph — a label matching 0 or 2+ real files (e.g. two different `index.ts` files) is left alone, mirroring the conservatism the existing key-based path already applies via `_loc_collisions`.

No other code needed to change: `_ghost_remap` already flows into the existing `norm_to_id` map that both edge endpoints and hyperedge members resolve through.

## Testing

- Added two regression tests in `tests/test_build.py`:
  - `test_ghost_merge_wrong_source_file_resolved_by_label` — the fix's core positive case, including the "prose dropped a leading path segment" variant (`customer-app/index.ts` vs the real `apps/customer-app/index.ts`).
  - `test_ghost_merge_wrong_source_file_ambiguous_basename_left_alone` — confirms the fallback stays conservative when a basename is shared by two real files.
- Full existing suite: `uv run pytest -q` → 5228 passed, 93 skipped (unchanged from the pre-patch baseline on this checkout — the only failures present with or without this patch are pre-existing and unrelated: `test_ollama_retry_cap.py` needs the optional `openai` extra not installed here, and `test_skillgen.py`'s git-history audits need a commit ref my shallow clone doesn't have).
- `ruff check` clean on both changed files.
- Verified against a real ~4600-node project graph (not included in this repo): **10 long-standing ghost nodes removed** across two unrelated documents (a memory query file and an ops runbook), confirming this is a general failure mode of the "document mentions a file by name" pattern rather than a narrow edge case. Both hyperedges that had cited the phantoms now correctly cite only real, AST-backed nodes; graph health diagnostic stayed clean before and after.

### DIFF
--- a/graphify/build.py
+++ b/graphify/build.py
@@ -993,6 +993,7 @@ def build_from_json(extraction: dict, *, directed: bool = False, root: str | Pat
     _loc_nodes: dict[tuple[str, str], str] = {}   # (source_file, label) -> canonical node id
     _loc_collisions: set[tuple[str, str]] = set()  # keys shared by 2+ AST nodes
     _noloc_nodes: dict[tuple[str, str], str] = {}  # (source_file, label) -> ghost node id
+    _ast_file_nodes: list[tuple[str, str]] = []  # (node_id, source_file) for AST file-self nodes (#2963)
 
     # Pass 1: collect canonical nodes — AST-origin nodes take precedence over LLM nodes.
     # When 2+ AST nodes share a key (same-named symbols in same-named files across
@@ -1035,6 +1036,20 @@ def build_from_json(extraction: dict, *, directed: bool = False, root: str | Pat
                     _loc_collisions.add(key)
                 # AST-origin nodes always overwrite a prior non-AST entry.
                 _loc_nodes[key] = nid
+                # #2963: a self-referential semantic pass (e.g. re-extracting a
+                # saved graphify-out/memory/*.md query answer) mints a NEW
+                # non-AST node for every bare file path/basename it mentions in
+                # prose ("App.tsx", "customer-app/index.ts"), stamped with
+                # source_file = the memory doc being read, NOT the file named in
+                # the prose. That wrong source_file means such a ghost can never
+                # hit the (sf, label) key above — same underlying bug as #1145,
+                # just with the (source_file, label) *pair* broken instead of
+                # only the id. Record every AST node whose own label already
+                # names its own file (_is_file_node_label — the same "is this a
+                # file node" predicate the label-disambiguation pass uses) so
+                # Pass 2b below can catch these by label alone.
+                if _is_file_node_label(label, sf):
+                    _ast_file_nodes.append((nid, sf))
             else:
                 # First non-AST node for this (file, label) wins as canonical; a
                 # later same-key node is a genuine same-file duplicate and still
@@ -1061,6 +1076,37 @@ def build_from_json(extraction: dict, *, directed: bool = False, root: str | Pat
         ast_id = _loc_nodes.get(key)
         if ast_id is not None:
             _ghost_remap[sem_id] = ast_id
+
+    # Pass 2b (#2963): catch ghosts the (source_file, label) key above cannot,
+    # because their source_file is simply wrong — a semantic pass over a
+    # document that only *mentions* a file (a saved graphify-out/memory/*.md
+    # query answer, a README, an ADR) stamps the file's own name as a new
+    # node's label but the DOCUMENT's path as source_file, since it has no way
+    # to know the mentioned file's real path. Resolve these by label alone
+    # against every AST file-self node collected in Pass 1, reusing
+    # _is_file_node_label so "App.tsx" matches source_file ".../App.tsx" and
+    # "customer-app/index.ts" matches ".../apps/customer-app/index.ts" (a
+    # directory-qualified suffix, e.g. the leading "apps/" the prose dropped).
+    # Conservative by construction: a label matching 0 or 2+ AST files is left
+    # alone (0 = no known file, 2+ = genuinely ambiguous — same "no safe
+    # canonical winner" rule Pass 2's _loc_collisions already applies).
+    if _ast_file_nodes:
+        for nid in sorted(node_set):
+            if nid in _ghost_remap:
+                continue  # already resolved by the exact (sf, label) key
+            attrs = G.nodes[nid]
+            if attrs.get("_origin") == "ast":
+                continue
+            label = str(attrs.get("label", "")).strip()
+            if not label:
+                continue
+            matches = {
+                ast_id for ast_id, ast_sf in _ast_file_nodes
+                if _is_file_node_label(label, ast_sf)
+            }
+            if len(matches) == 1:
+                _ghost_remap[nid] = next(iter(matches))
+
     # Remove ghost nodes from the graph; edges will be re-pointed via norm_to_id.
     for ghost_id in _ghost_remap:
         G.remove_node(ghost_id)
--- a/tests/test_build.py
+++ b/tests/test_build.py
@@ -577,6 +577,62 @@ def test_ghost_merge_non_ast_same_file_still_merges():
     assert G.number_of_nodes() == 1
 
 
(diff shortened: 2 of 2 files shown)


safishamsi:
Shipped in **v0.9.55** — landed on `v8` via cherry-pick with your authorship preserved. Thanks @leninherrera94 — ghost nodes mis-attributed to the referencing doc now merge onto the real file node. Release: https://github.com/Graphify-Labs/graphify/releases/tag/v0.9.55

Issue #2040, opened 2026-07-19
Nested class/object containment is never represented as a graph edge — `contains` always sources from the file node, not the enclosing type, across ~19 languages


sub4biz:
## Summary

For any language using `_extract_generic`'s class-node branch (`graphify/extractors/engine.py`), a nested
class/object/trait never gets a `contains` edge from its actual enclosing type. The edge is hard-coded to
originate from the **file** node, regardless of nesting depth:

```python
if t in config.class_types:
    ...
    class_nid = _make_id(stem, ".".join(namespace_stack), class_name)
    add_node(class_nid, class_name, line, metadata=metadata)
    callable_def_nids.add(class_nid)
    add_edge(file_nid, class_nid, "contains", line)   # <-- always file_nid, never parent_class_nid
    ...
```

There is a partial, C#-specific mitigation: when `parent_class_nid` is set, the node gets a
`metadata={"is_nested_type": True}` flag — but even then, no edge is created *to* the actual parent, only a
boolean flag on the node. Every other language, including Scala, doesn't even get that.

(A nested type's own *members* — e.g. a method inside the nested class — DO correctly nest under it via the
usual `parent_class_nid` mechanism once the nested type itself is being walked. The gap is specifically one
level up, at the class-containing-a-class edge, not at class-containing-a-method.)

## Reproduction (Scala)

```scala
class Outer {
  class Inner
  def makeInner: Inner = new Inner
}
```

`graph.json` shows `Inner`'s `contains` edge sourced from the **file**, never from `Outer` — there is no edge
anywhere in the graph connecting `Outer` → `Inner`. The only way to know `Inner` lives inside `Outer` is to
notice their source lines are close together and infer it; the graph itself does not encode it.

## Why this matters

- Any codebase using nested types for organization (inner builder/helper classes, module-like nested objects)
  loses that structure in the graph — nested types appear as flat siblings of everything else in the file,
  indistinguishable from top-level types.
- Community detection / `god_nodes` likely treats a deeply-nested helper class the same as a top-level public
  API class, since there's no edge signal indicating one is an implementation detail nested inside the other.

## Cross-language scope — precise counts, not "every language"

**Path 1 — shared `_extract_generic`/`engine.py` walker (15 `LanguageConfig` blocks in `extract.py`).** Every
one of these calls the same `_extract_generic` function with only the config differing, so the bug is
structurally guaranteed for all of them. **Empirically confirmed by a live `graphify update` run, each
[...]

PR #3356, opened 2026-09-05
fix(resolution): require owned sources for type-use edges

VasuBansal7576:
A class inside a named function can have no extracted AST node while the symbol resolver still emits its inheritance and member type-reference edges.
Abstract method signatures can produce the same missing-source records.

Require an owned source node before emitting `inherits`, `implements`, or `references` use facts.
This preserves relationships for callback-local classes that the structural extractor does materialize, including their method return types.
It follows the existing ownership boundary for call facts without changing call attribution or adding extraction support for omitted declarations.

The four-file CLI reproduction on upstream `937e59a5476fcb2665d6c4f4b7c0d0a4142011b6` keeps all 12 nodes and changes 24 edges to 22.
Only the two missing-source records disappear; `Visible.method → Result` remains.

[Immutable fixture, raw graphs, exact hashes and CLI receipts](https://github.com/VasuBansal7576/graphify/tree/7ddd8582e94dfc4cfa7a332a3b8ff976e8bf153e).
The images below are offline-rendered raw-graph audit reports, not Graphify viewer screenshots.
Browser capture was unavailable because the browser security policy blocked the local report URL.

| Before | After |
| --- | --- |
| Before audit report | After audit report |

Validation:

- Three new extraction regressions fail before and pass after: named-function versus callback-local classes, same-named classes in sibling same-basename files, and abstract versus concrete methods.
- 108 tests pass across the new regressions, symbol resolution, and JS import resolution.
- Ruff and `git diff --check` pass.
- Required `graphify update . --no-cluster` completed on the sparse development checkout. Optional SQL, DM and Robot parsers were unavailable; the existing Luau fixture reports a syntax warning.
- A separate 22-file OpenClaw preservation control retains all 27 valid relationships lost by an ancestry-based filtering approach. This is bounded evidence, not full-source acceptance.

Duplicate checks included open and closed local-class, abstract-method and missing-source reports.
#2653 concerns nested function materialization; #2040 concerns containment; #1764 concerns synthesized endpoints in other extractors.
#2382 and #3342 concern incremental resolution and external-stub lifecycle paths.
This patch is limited to the shared symbol-use producer ownership check.

### DIFF
--- a/graphify/extractors/resolution.py
+++ b/graphify/extractors/resolution.py
@@ -1094,6 +1094,11 @@ def _apply_symbol_resolution_facts(
         if target_id is None:
             continue
         source_id = use_fact.source_id
+        # Structural walking can omit named-function-local declarations while
+        # materializing callback-local ones. Only actual node ownership decides
+        # whether a type relationship has a represented source declaration.
+        if use_fact.relation in ("inherits", "implements", "references") and source_id not in owned:
+            continue
         if use_fact.relation == "calls" and source_id not in owned:
             source_id = source_file_id.get(file_path)
             if source_id is None:
--- a/tests/test_owned_type_sources.py
+++ b/tests/test_owned_type_sources.py
@@ -1,0 +1,61 @@
+from pathlib import Path
+from graphify.extract import extract
+
+
+def run(tmp_path, sources):
+    paths = []
+    for name, text in sources.items():
+        p = tmp_path / name
+        p.parent.mkdir(parents=True, exist_ok=True)
+        p.write_text(text)
+        paths.append(p)
+    return extract(paths, root=tmp_path, cache_root=tmp_path, parallel=False)
+
+
+def triples(g):
+    return {(e["source"], e["target"], e["relation"]) for e in g["edges"]}
+
+
+def test_callback_local_method_preserved_named_function_local_omitted(tmp_path):
+    g = run(
+        tmp_path,
+        {
+            "base.ts": "export class Base {}\nexport interface Result {}\n",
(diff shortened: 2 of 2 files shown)


VasuBansal7576:
This is the intended source-ownership boundary, with preservation controls rather than an ancestry filter.
An `inherits` or `implements` edge needs a represented class/interface source; moving it to the file node would assert that the file itself inherits or implements that type.
A method type-reference edge likewise should not be reassigned to a different declaration merely because its method node is absent.
The existing file-level fallback for `calls` is unchanged.

The real CLI fixture keeps all 12 nodes and removes exactly the two records sourced by omitted `Hidden` declarations.
Both the represented callback-local `Visible` inheritance and `Visible.method → Result` reference remain.
The new tests also protect same-named classes in sibling files and a concrete method beside an omitted abstract signature.
A separate 22-file preservation control retains all 27 valid relationships lost by the earlier ancestry-based approach.

[Raw before/after graphs and exact hashes](https://github.com/VasuBansal7576/graphify/tree/7ddd8582e94dfc4cfa7a332a3b8ff976e8bf153e) and the three extraction regressions provide the reproducing execution that the advisory says was unavailable to its verifier.
Supporting currently omitted declarations would require materializing their class/method nodes; this PR does not claim to add that extraction support.

safishamsi:
Shipped in **v0.9.55** — landed on `v8` via cherry-pick with your authorship preserved. Thanks @VasuBansal7576 — type-use edges are no longer fabricated from a symbol that owns no node. Release: https://github.com/Graphify-Labs/graphify/releases/tag/v0.9.55

PR #3358, opened 2026-09-05
fix: resolve named re-exports through imported bindings

VasuBansal7576:
`api.ts` re-exporting a binding from `bridge.ts` currently points at a nonexistent bridge-local symbol when the bridge imports that binding and then exports it.
For `base.ts: export const VALUE = 1`, `bridge.ts: import { VALUE } from './base.js'; export { VALUE }`, and `api.ts: export { VALUE } from './bridge.js'`, the raw graph contains `api -> bridge_value` even though only `base_value` exists.

Resolve that exact named-export site through explicit export origins and rewrite it only when one origin has an owned declaration.
Keep unresolved explicit origins in the candidate set, so unsupported extraction cannot hide ambiguity.
Branches without that export contribute nothing; explicit named exports override star branches.
Record simple identifier export declarations so exported constants can be distinguished from private symbols.

The existing edge and source location are preserved.
Its transient `target_file` follows the defining source for ID disambiguation and is removed before graph persistence.
No destructured-binding implementation or edge deletion is added.

Validation:

- 19 regression tests cover forwarding, renamed bindings, same-line exports, private symbols, type-only metadata, cycles, named/star ambiguity, unsupported explicit origins, named-over-star precedence, and transient provenance.
- 124 tests passed across the new tests and existing symbol/JS import resolution suites; configured Ruff checks and `git diff --check` passed.
- Actual CLI before/after runs retain 4 nodes and 7 edges, changing missing targets from 1 to 0. [Raw before graph](https://github.com/VasuBansal7576/graphify/blob/fba6245f7d2f2066ef1cd5c3eb75efdcc9f2bcc9/before.json), [raw after graph](https://github.com/VasuBansal7576/graphify/blob/fba6245f7d2f2066ef1cd5c3eb75efdcc9f2bcc9/after.json), [source and hash receipt](https://github.com/VasuBansal7576/graphify/tree/fba6245f7d2f2066ef1cd5c3eb75efdcc9f2bcc9).
- The required AST-only `graphify update . --no-cluster` completed. Optional SQL/DM/Robot grammars were unavailable, and the existing Luau fixture emitted a syntax warning.

Offline audit of actual raw graphs, not a Graphify viewer screenshot

Review follow-up at `0a1f06be8b482505a1c7399686b50a9ddff11457` fixes valid-cycle reachability, restores baseline aggregate-pattern declaration facts, and derives transient target-file provenance from the resolved origin.
Five added controls cover short/long cycles, source-less nodes, aggregate-pattern compatibility, and identifier aliases.
The linked raw evidence and offline visual remain the original `08dd8c1` proof; no new viewer capture is claimed.

### DIFF
--- a/graphify/extractors/resolution.py
+++ b/graphify/extractors/resolution.py
@@ -935,6 +935,7 @@ def _apply_symbol_resolution_facts(
                     changed = True
 
     named_exports_by_file: dict[Path, dict[str, tuple[Path, str]]] = {}
+    export_origins: dict[tuple[Path, str], set[tuple[Path, str]]] = {}
     star_exports_by_file: dict[Path, list[Path]] = {}
 
     for star_fact in facts.star_exports:
@@ -996,7 +997,14 @@ def _apply_symbol_resolution_facts(
             if origin is None and (file_path, export_fact.local_name) in symbol_nodes:
                 origin = (file_path, export_fact.local_name)
         if origin is None:
+            # An explicit export remains evidence even when extraction did not
+            # materialize its binding. Do not mistake it for an absent export.
+            if export_fact.local_name is not None:
+                export_origins.setdefault((file_path, export_fact.exported_name), set()).add(
+                    (file_path, export_fact.local_name)
+                )
             continue
+        export_origins.setdefault((file_path, export_fact.exported_name), set()).add(origin)
         named_exports_by_file.setdefault(file_path, {})[export_fact.exported_name] = origin
         if origin[0] != file_path:
             source_id = source_file_id.get(file_path)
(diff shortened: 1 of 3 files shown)


VasuBansal7576:
Checked all three findings against the pinned upstream baseline and executable fixtures.

1. The empty-result fallback had a real cycle bug: when a star branch cycles back to a module that also reaches a real definition, the cycle cutoff became a spurious unresolved origin and prevented resolution.
The follow-up distinguishes cycle cutoffs from absent exports and preserves explicitly unresolved origins.
Short and longer cycle fixtures now resolve to the real definition, while the existing unsupported-origin and conflicting-star tests still remain unresolved.

2. A node without source_file cannot supply the relevant symbol-index entry: missing/empty source paths are skipped before indexing, and synthesized declarations always carry their source path.
The boundary test confirms such an owned node is not selected and does not crash.
The follow-up also removes the node_sources dictionary lookup and derives transient target_file from the resolved origin through the existing path_by_resolved mapping.
Existing relative-path and provenance tests pass.

3. The baseline helper reported whole pattern text such as {x} and [x] for identifier-valued lexical declarations; it did not enumerate individual destructured bindings.
The initial patch removed those legacy facts and their aggregate nodes, so this was an observable compatibility change.
The follow-up restores that behavior and keeps individual binding support separate.
Object-pattern, array-pattern, aggregate-name-import, and ordinary identifier-alias fixtures now have exact export-fact, node, and edge parity with upstream.

Validation: 124 tests passed, including five new review controls and all existing forwarding/ambiguity, symbol-resolution, and JS import-resolution tests.
Configured Ruff checks and git diff --check passed.

safishamsi:
Shipped in **v0.9.55** — landed on `v8` via cherry-pick with your authorship preserved. Thanks @VasuBansal7576 — named re-exports now resolve through imported bindings. Release: https://github.com/Graphify-Labs/graphify/releases/tag/v0.9.55

PR #3362, opened 2026-09-05
fix: resolve unmapped @/ imports in JS projects

### LINKED ISSUE #3357: JS: `calls` edge dropped for a `@/` alias import when no tsconfig/jsconfig exists (0.8.11 emitted it, 0.9.53 does not)
## Summary

A JavaScript file that imports a named function through a path alias (`import { fn } from "@/other.js"`) and calls it inside a method body gets a `calls` edge at graphify 0.8.11 and no `calls` edge at 0.9.53. The `imports` and `imports_from` edges for that file are gone at 0.9.53 as well. The nodes for both functions still exist, extraction reports success, and nothing warns.

The trigger is the alias specifier with no `tsconfig.json` or `jsconfig.json` on disk. Frameworks such as uni-app / HBuilderX resolve `@/` to the project root natively and ship without either config file, so the alias never reaches `_resolve_tsconfig_alias` with a mapping and the specifier falls to the bare-import branch in `extract.py` (the "treat as external" path just below the `#701` comment). 0.8.11 still emitted the `calls` edge in that situation, so this is a regression for those projects.

## Versions and environment

- graphifyy 0.8.11 (edge present) and graphifyy 0.9.53 (edge absent), each in its own clean venv, Python 3.14.6, macOS 26.6.2
- Both runs: `graphify update <dir>`, no LLM key set, tool printed "no LLM needed"

## Minimal reproduction

Two files in an otherwise empty folder. No `package.json`, no `tsconfig.json`, no `jsconfig.json`.

`adapter.js`
```js
export function enableBackgroundBle(options = {}) {
  const timer = setInterval(() => {}, options.interval || 1000);
  return () => clearInterval(timer);
}
```

`client.js`
```js
import { enableBackgroundBle } from "@/adapter.js";

export class BleClient {
  async createBLEConnection({ forceRefresh = false } = {}) {
    this._cleanup = enableBackgroundBle({ interval: forceRefresh ? 500 : 1000 });
    return true;
  }
}
```

Run:
```bash
graphify update "$PWD"
python3 -c "
import json; g=json.load(open('graphify-out/graph.json'))
for l in g['links']: print(l['source'], '--'+l['relation']+'-->', l['target'], l.get('confidence'))"
```

## Actual

graphify 0.8.11 (5 nodes, 5 edges):
```
client_js --imports_from--> adapter_js EXTRACTED
client_js --contains--> ..._client_bleclient EXTRACTED
..._client_bleclient --method--> ..._client_bleclient_createbleconnection EXTRACTED
..._client_bleclient_createbleconnection --calls--> ..._adapter_enablebackgroundble EXTRACTED
adapter_js --contains--> ..._adapter_enablebackgroundble EXTRACTED
```

graphify 0.9.53 (5 nodes, 3 edges):
```
adapter --contains--> adapter_enablebackgroundble EXTRACTED
client --contains--> client_bleclient EXTRACTED
client_bleclient --method--> client_bleclient_createbleconnection EXTRACTED
```

## Expected

At 0.9.53, at least the edge 0.8.11 produced:
```
client_bleclient_createbleconnection --calls--> adapter_enablebackgroundble
```
Ideally also `client --imports--> adapter_enablebackgroundble` and `client --imports_from--> adapter`, which 0.9.53 does emit when the import is relative (below).

## What isolates it

Same two files, three variations, all at 0.9.53:

| Variation | `calls` edge | `imports` edges |
|---|---|---|
| `from "@/adapter.js"`, no config file | absent | absent |
| `from "./adapter.js"`, no config file | present | present |
| `from "@/adapter.js"` plus a `jsconfig.json` with `"paths": { "@/*": ["./*"] }` | present | present |

So the alias resolver works when it has a mapping. The regression is the fallback when it has none: 0.8.11 resolved the call by name and 0.9.53 does not.

## Impact on a real corpus

On a JS/Vue project of roughly 17.5k nodes that uses `@/` throughout and has no tsconfig/jsconfig, rebuilding from 0.8.11 to 0.9.53 took distinct cross-file `calls` pairs from 482 to 192; 404 of the old 482 have no counterpart in the new graph. Same-file `calls` pairs rose over the same rebuild (3,955 to 5,201), so this is specific to cross-file resolution. `graphify path` answers that routed through one of those edges at 0.8.11 now return "No directed path found", and `--undirected` finds nothing either.

## Not a direction artifact

I checked this against the `_src`/`_tgt` handling from #563: `export.py:347` in 0.9.53 pops the markers and writes `source`/`target` in the true order, and the written graph.json carries `directed: false` with no markers. The edge above is not flipped; it is absent from the link list entirely.

## Related

- #147 (tsconfig paths support, the mapping side of this)
- #3346 (TS: calls through namespace or renamed imports resolve on the identifier, not the binding)
- #2943 (Python: aliased module import loses `calls` in a subdirectory)
- #2837 (question about the JS/TS import gate)

A possible direction: when an alias specifier has no tsconfig/jsconfig mapping and the last path segment matches exactly one on-disk source file in the scanned tree, resolve to it instead of marking it external; or keep the 0.8.11 name-based `calls` fallback for unresolved bare specifiers, marked INFERRED.

hopstreax:
## Summary

Fixes #3357 by resolving conventional `@/` imports in JavaScript/TypeScript projects that do not have a `tsconfig.json` or `jsconfig.json`.

Previously, an import such as:

```js
import { enableBackgroundBle } from "@/adapter.js";
```

could remain unresolved, preventing Graphify from establishing import evidence and therefore preventing the corresponding cross-file `calls` edge under the `#1659` safety gate.

This change resolves `@/` imports to concrete files on disk without restoring the historical name-based inference fallback.

## What changed

* Added unmapped `@/` convention resolution to the central JS/TS module resolver.
* Discover the project anchor using:

  * nearest `package.json` or `pnpm-workspace.yaml`
  * otherwise the existing VCS root
  * otherwise the importing file's directory
* Probe conventional locations in this order:

  1. `<project>/src/<path>`
  2. `<project>/<path>`
* Reuse `_resolve_js_import_path()` so normal JS/TS extensions and index resolution remain consistent.
* Applied the same resolution behavior to regex-rescued imports.
* Preserve existing `tsconfig.json` / `jsconfig.json` alias and `baseUrl` behavior.

## Safety / compatibility

This does **not** restore the old repository-wide name-based `INFERRED` call fallback.

Instead, a `calls` edge is produced only after the `@/` import resolves to a concrete file and provides legitimate import evidence, preserving the `#1659` safety boundary.

Also preserves the `#3125` behavior: when a `tsconfig.json` exists without a `paths` mapping, the unmapped `@/` convention is not synthesized.

Scoped packages such as `@scope/pkg` remain unaffected.

## Tests

* Minimal unmapped `@/` reproduction.
* Nested importers with `src/` layouts.
* Nested importers with flat layouts.
* `tsconfig.json` without `paths` (`#3125`).
* Explicit `tsconfig` alias precedence.
* Missing `@/` targets.
* Scoped package imports.
* Regex-rescued imports.
* Prevention of phantom calls to unrelated definitions (`#1659`).

Verification:

```text
123 passed
git diff --check: clean
```

### DIFF
--- a/graphify/extract.py
+++ b/graphify/extract.py
@@ -1617,6 +1617,11 @@ def _resolve_rescued_specifier(
         resolved_file = (resolved_alias if resolved_alias is not None
                          and resolved_alias.is_file() else None)
         return _make_id(str(resolved_alias)), str(resolved_alias), resolved_file
+    # Unmapped `@/` project-root convention fallback (#3357).
+    if raw.startswith("@/"):
+        resolved_unmapped = _resolve_js_module_path(raw, path.parent)
+        if resolved_unmapped is not None and resolved_unmapped.is_file():
+            return _make_id(str(resolved_unmapped)), str(resolved_unmapped), resolved_unmapped
     # Bare/scoped import (node_modules) - use last segment;
     # build_from_json drops as external if no matching node exists.
     module_name = raw.split("/")[-1]
--- a/graphify/extractors/resolution.py
+++ b/graphify/extractors/resolution.py
@@ -505,6 +505,33 @@ def _resolve_workspace_import(raw: str, start_dir: Path) -> Path | None:
                 return resolved
     return None
 
+def _find_js_project_anchor(start_dir: Path) -> Path:
+    """Discover the project root anchor for unmapped JS/TS convention aliases (#3357).
+
+    Walks upward from start_dir looking for:
+    1. Nearest directory containing package.json or pnpm-workspace.yaml
+    2. Nearest directory containing a VCS marker (_find_vcs_root)
+    3. Fallback: start_dir itself
+    """
+    from graphify.detect import _find_vcs_root
+
+    current = start_dir.resolve()
+    home = Path.home()
+
+    for candidate in [current, *current.parents]:
+        if candidate == home:
+            break
+        if (candidate / "package.json").is_file() or (candidate / "pnpm-workspace.yaml").is_file():
+            return candidate
+
+    vcs_root = _find_vcs_root(start_dir)
+    if vcs_root is not None:
+        return vcs_root
+
+    return current
+
+
+
 def _resolve_js_module_path(raw: str | Path, start_dir: Path | None = None) -> Path | None:
     """Resolve a JS/TS module path or specifier to a local source file.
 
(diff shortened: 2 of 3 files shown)


safishamsi:
Shipped in **v0.9.56** — landed on `v8` via cherry-pick with your authorship preserved. Thanks @hopstreax — JS/TS `@/` project-root alias imports now resolve when no tsconfig paths mapping exists. Release: https://github.com/Graphify-Labs/graphify/releases/tag/v0.9.56

PR #3363, opened 2026-09-05
fix(watch): preserve qualified file labels in no-cluster updates

VasuBansal7576:
`graphify update --no-cluster` writes same-basename files with ambiguous labels because the raw rebuild skips the file-label disambiguation used by fresh extraction. For example, `pkg_a/errors.ts` and `pkg_b/errors.ts` both become `errors.ts` after rebuilding.

Reuse `disambiguate_file_labels_in_nodes` after deduplicating nodes in the raw watch rebuild. Node IDs and edges remain unchanged. The regression checks both an initial rebuild and an incremental update after editing an importer.

Validation on current `v8` at `937e59a`:
- The new regression fails on the unchanged release with `{'errors.ts'}` instead of the two qualified names.
- All 149 watch tests pass on this branch.
- Ruff and `git diff --check` pass.
- AST-only `graphify update . --no-cluster` completes. Optional SQL, DM, Common Lisp and Robot Framework parsers are absent, and the existing Luau fixture reports a syntax warning.

The earlier CLI reproduction used a 15-file OpenWiki corpus after a rename, deletion and import rewire. Before and after graphs had identical node IDs and edge lists, with 197 nodes and 575 edges; only the two colliding file labels changed. These captures show that earlier comparison at `33362d9` / `ddab7f0`, not the rebased revisions above.

| Before | After |
| --- | --- |
| Before: both files labeled errors.ts | After: directory-qualified labels |

Checked existing open and closed work for overlap. #2218 handles community-label membership and visualization retention; #3342 handles external dependency retention. This change concerns file-node display labels in the no-cluster rebuild.

### DIFF
--- a/graphify/watch.py
+++ b/graphify/watch.py
@@ -1783,10 +1783,16 @@ def _rebuild_code(
             # Dedupe parallel edges (the clustered path's DiGraph collapses them implicitly);
             # without it, --no-cluster + repeated `update` accumulate duplicates and edge
             # counts diverge across build modes (#1317).
-            from graphify.build import dedupe_edges as _dedupe_edges, dedupe_nodes as _dedupe_nodes
+            from graphify.build import (
+                dedupe_edges as _dedupe_edges,
+                dedupe_nodes as _dedupe_nodes,
+                disambiguate_file_labels_in_nodes as _disamb_labels,
+            )
+            raw_nodes = _dedupe_nodes(result.get("nodes", []))
+            _disamb_labels(raw_nodes)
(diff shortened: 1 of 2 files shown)


safishamsi:
Shipped in **v0.9.55** — landed on `v8` via cherry-pick with your authorship preserved. Thanks @VasuBansal7576 — no-cluster watch/update now disambiguates same-basename file labels. Release: https://github.com/Graphify-Labs/graphify/releases/tag/v0.9.55

PR #3364, opened 2026-09-06
Extract direct recursive calls as calls self loops

### LINKED ISSUE #3350: Python AST extraction omits direct recursive calls while ordinary calls are emitted
## Summary

With `graphifyy 0.9.54`, direct Python recursion does not produce a `calls` self-edge in the output of `graphify.extract.extract()`. An ordinary call to the same function is extracted correctly.

This reproduces before graph construction, deduplication, export, or querying. The example invokes the upstream extractor directly: no custom adapter, semantic backend, or LLM is involved.

## Minimal reproduction

Run this as a Python script in an environment with `graphifyy==0.9.54` and its Python tree-sitter dependencies installed:

```python
from pathlib import Path
from tempfile import TemporaryDirectory

with TemporaryDirectory() as directory:
    root = Path(directory)
    source = root / "repro.py"
    source.write_text(
        "def factorial(n):\n"
        "    return 1 if n < 2 else n * factorial(n - 1)\n"
        "\n"
        "def entry(n):\n"
        "    return factorial(n)\n",
        encoding="utf-8",
    )

    result = extract([source], root=root, cache_root=root)
    calls = [
        (edge["source"], edge["target"])
        for edge in result["edges"]
        if edge.get("relation") == "calls"
    ]
    print("calls:", calls)
    print("ordinary call:", ("repro_entry", "repro_factorial") in calls)
    print("recursive call:", ("repro_factorial", "repro_factorial") in calls)
```

## Actual result

```text
calls: [('repro_entry', 'repro_factorial')]
ordinary call: True
recursive call: False
```

Both function nodes exist, but the call in the body of `factorial` is absent.

## Expected result

In addition to `entry -> factorial`, include an extracted `factorial -> factorial` `calls` edge with the recursive call's source location. This would preserve the actual call structure for downstream inspection.

If omitting direct recursion is an intentional extractor contract, could that limitation be documented, or could retaining recursive calls be supported explicitly?

## Source observations

The same condition is present on the current `v8` branch at commit `937e59a5476fcb2665d6c4f4b7c0d0a4142011b6`:

- [extractors/engine.py, direct-call handling](https://github.com/Graphify-Labs/graphify/blob/937e59a5476fcb2665d6c4f4b7c0d0a4142011b6/graphify/extractors/engine.py#L5571): `if tgt_nid and tgt_nid != caller_nid:` excludes a resolved call whose target is its caller.
- In the installed 0.9.54 `extract.py`, the common raw-call resolution path also requires `tgt != caller` (line 7154).
- Conversely, the installed `build.py` explicitly preserves meaningful recursive `calls` self-edges while filtering only collapsed import/re-export self-loops. A supplied recursive edge survives graph construction.

This appears to be an extraction-stage omission, not loss of an existing edge during serialization or incremental merge. I have not established when the behavior was introduced, so this report does not claim a regression introduced specifically in 0.9.54.

## Related reports checked

#2037 and #2038 address false import self-loops and preserving an already-supplied recursive `calls` edge in `build_from_json()`. The regression test in #2038 constructs that self-edge manually; it does not test extraction of a recursive Python function. The maintainer marked that issue fixed in 0.9.21.

#2374 / #2401 concern false self-loops caused by stub rewiring, rather than an absent direct recursive call.

I searched existing open/closed issues and PRs using recursion, recursive calls, self-loop, self-call, factorial, and caller_nid, and did not find a report for this specific extraction-stage case. Please link this to an existing report if I missed one.

## Environment

- Graphifyy: 0.9.54
- Python: 3.13.14
- Windows 11, build 26200, AMD64
- tree-sitter: 0.25.2
- tree-sitter-python: 0.25.0

The extractor and build modules used for the reproduction match the same-version upstream package files. Only a disposable synthetic source file and temporary extraction cache are used.

ayushcodes10:
## What

`_extract_generic`'s call resolution required `tgt_nid != caller_nid` before emitting a `calls` edge, so a function calling itself by name was silently dropped, no matter which language sharing this engine it was written in.

`build_from_json` already preserves a supplied recursive `calls` self loop rather than stripping it (`test_recursive_call_self_loop_is_preserved`), so the intent to keep genuine recursion visible already existed one layer up. Extraction just never produced the edge for that layer to preserve.

## Repro (from the issue, reproduced before this fix)

```python
def factorial(n):
    return 1 if n < 2 else n * factorial(n - 1)

def entry(n):
    return factorial(n)
```

Before: `calls: [('repro_entry', 'repro_factorial')]` (recursive call missing)
After: `calls: [('repro_factorial', 'repro_factorial'), ('repro_entry', 'repro_factorial')]`

Verified this also fixes recursion for Java, Ruby, and other `_extract_generic`-routed languages, since the guard removed was language agnostic.

## Scope

Several dedicated per-language extractors (Go, Rust, CommonLisp, DreamMaker, Elixir, PowerShell, Zig) carry the same `tgt_nid != caller_nid` guard in their own separate files. Left those untouched here since the issue's repro and my verification are both Python/`_extract_generic`-scoped, and fixing those too would be a materially larger, separate change across files this report didn't exercise. Happy to open a follow-up if that's wanted.

Fixes #3350.

### DIFF
--- a/graphify/extractors/engine.py
+++ b/graphify/extractors/engine.py
@@ -5658,7 +5658,14 @@ def _extract_generic(
                         and not nid_to_sf.get(tgt_nid)
                     ):
                         tgt_nid = None
-                if tgt_nid and tgt_nid != caller_nid:
+                if tgt_nid:
+                    # A direct recursive call (tgt_nid == caller_nid) is a real
+                    # self-edge, not noise: name resolution binds it the same
+                    # way as any other call, and build_from_json already
+                    # preserves a supplied recursive calls self-loop rather
+                    # than stripping it, so extraction dropping it here was
+                    # the only place the call structure actually got lost
+                    # (#3350).
                     pair = (caller_nid, tgt_nid)
                     if pair not in seen_call_pairs:
                         seen_call_pairs.add(pair)
--- a/tests/test_extract.py
+++ b/tests/test_extract.py
@@ -559,6 +559,9 @@ def test_python_call_edges_have_call_context():
 
 
 def test_calls_no_self_loops():
+    """sample_calls.py has no recursive functions, so no call resolves to its own
+    caller here -- a genuinely recursive function DOES produce a self-loop (#3350),
+    see test_recursive_call_produces_self_loop below."""
     result = extract_python(FIXTURES / "sample_calls.py")
     for edge in result["edges"]:
         if edge["relation"] == "calls":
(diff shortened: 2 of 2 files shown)


ayushcodes10:
@safishamsi opened this against a report of direct Python recursion never producing a calls self-loop (#3350), traced to a shared tgt_nid != caller_nid guard in _extract_generic used by many languages, not just Python. Verified the same fix also restores recursion for Java and Ruby. Flagging the scope choice in the PR description: a few dedicated per-language extractors carry the same guard separately and weren't touched here. Happy to address any feedback.

PR #3367, opened 2026-09-06
fix(rust): extract trait method declarations (fixes #3366)

### LINKED ISSUE #3366: Rust trait method declarations produce no nodes — 504/654 (77%) missing across 3 repos; trait_item never walks its body and function_signature_item is unhandled
## Summary

A Rust trait gets a node; the methods it declares do not. Across three real Rust
repositories, **504 of 654 trait method declarations (77%) have no node at all**. The
ones that do appear are only there because an `impl` block in the same file happens to
define a method of the same name — the node's `source_location` points at the impl, not
at the declaration.

The effect is that a trait — the API contract — is present in the graph as a bare name
with no operations attached, so "what does this trait require an implementor to provide?"
cannot be answered from the graph.

## Minimal reproduction

```rust
// src/lib.rs
pub trait Greeter {
    /// signature only; implemented below in this same file
    fn greet(&self) -> String;

    /// default body, never overridden anywhere
    fn shout(&self) -> String {
        self.greet().to_uppercase()
    }
}

pub struct Alice;

impl Greeter for Alice {
    fn greet(&self) -> String {
        "alice".to_string()
    }
}

pub fn plain_function() -> u8 { 7 }
```

```rust
// src/lonely.rs — a trait with no implementor in the corpus
pub trait Lonely {
    fn only_signature(&self) -> u8;
    fn with_default(&self) -> u8 { 42 }
}
```

```
src/lib.rs: 5 nodes                     src/lonely.rs: 2 nodes
   L1  lib.rs                              L1  lonely.rs
   L1  Greeter                             L2  Lonely
  L11  Alice                               (only_signature — absent)
  L14  .greet()   <- the impl at L14,      (with_default   — absent)
                    not the declaration at L3
  L19  plain_function()
   (shout — absent)
```

Note `.greet()` is anchored at **L14**. There is no node for the declaration at L3; the
only method node came from the impl. `shout`, which has a default body and no override
anywhere, is absent entirely. In `lonely.rs`, with no impl in the corpus, the trait
contributes **zero** method nodes.

Counting `fn` declarations inside `trait` blocks and checking whether a node with that
name exists in the same file (graphifyy 0.9.55, `--code-only`):

| repo | traits | trait methods | with no node | lost |
|---|---|---|---|---|
| Abhilekh-Cloud | 40 | 112 | 70 | **62%** |
| Abhilekh-Collab | 37 | 144 | 91 | **63%** |
| Abhilekh-App (rust-lib) | 87 | 398 | 343 | **86%** |
| total | 164 | 654 | **504** | **77%** |

Examples: `Acts::policy_acts` (`libs/access-control/src/act.rs:8`), `enforce_action`
(`libs/access-control/src/collab.rs:11`), `TypeOption::json_cell` and `stringify_cell`
(`collab-database/src/fields/type_option/mod.rs:131,138`), `workspace_id` / `device_id`
(`collab-integrate/src/collab_builder.rs:72,73`).

## Where it comes from

In `extractors/rust.py`, the `trait_item` branch mints the trait node and then only
iterates `trait_bounds` — it never walks the trait's body:

```python
    name_node = node.child_by_field_name("name")
    if name_node:
        ...
        add_node(item_nid, item_name, line)
        add_edge(file_nid, item_nid, "contains", line)
        if t == "trait_item":
            for c in node.children:
                if c.type != "trait_bounds":
                    continue
                ...
```

And `function_signature_item` — the tree-sitter-rust node type for a bodyless
`fn greet(&self) -> String;` — does not appear anywhere in the package:

```
$ grep -rn "function_signature_item" site-packages/graphify/
(no matches)
```

So a signature-only method has no extraction path at all, and a default-bodied method
inside a trait body is not reached either.

## Suggested fix

In the `trait_item` branch, walk the trait's `declaration_list` and emit a node per
`function_signature_item` and per `function_item`, with a `contains` (or `defines`) edge
from the trait node — mirroring what `impl_item` already does for impl methods. That
would also give traits the same shape as structs/enums, whose members are already
represented.

Two consequences worth having beyond the node count: `explain <Trait>` could list the
operations a trait requires, and an implementor's method could be linked to the
declaration it satisfies.

## Environment

graphifyy 0.9.55 (uv tool, Python 3.12.12), macOS 15 arm64, `graphify extract --code-only`.
Related: #3365 (Dart `source_location`), found in the same cross-validation sweep.

santoshpy:
Fixes #3366.

## The problem

A Rust trait gets a node; the methods it declares do not. Across three real Rust repositories, **504 of 654 trait method declarations (77%) have no node**. A trait is the API contract, so it appears in the graph as a bare name with no operations attached — `explain <Trait>` cannot say what an implementor must provide.

The methods that *do* appear are only there because an `impl` block in the same file happens to define a method of the same name, and they are anchored at the impl, not the declaration.

## Two gaps, both in `extractors/rust.py`

1. **`function_signature_item` had no branch anywhere in the package.** That is the tree-sitter-rust node type for a bodyless `fn greet(&self) -> String;`, so a signature-only method had no extraction path at all.
2. **The `struct_item`/`enum_item`/`trait_item` branch returns before descending**, so nothing walked a trait's body — a method with a *default body* inside a trait was never reached either.

## The change

- Add a `function_signature_item` branch mirroring `function_item`, minus the body walk (there is no body).
- In the `trait_item` case, descend into the trait's body the way `impl_item` already descends into its own, passing the trait node as `parent_impl_nid` so each method gets a `method` edge from the trait.

Both follow the existing shape of the file rather than introducing a new mechanism.

## Before / after

The reproducer from #3366 — a trait with no implementor in the corpus:

```rust
pub trait Lonely {
    fn only_signature(&self) -> u8;
    fn with_default(&self) -> u8 { 42 }
}
```

```
before:  L1 lonely.rs, L2 Lonely
after:   L1 lonely.rs, L2 Lonely, L3 .only_signature(), L4 .with_default()
         + method edges Lonely -> each
```

And where an impl exists, the declaration and the implementation are now distinct nodes (`Greeter::greet` at the trait's line, `Alice::greet` at the impl's), instead of only the latter.

## Measured on a real codebase

A 937-file Rust service:

| | before | after |
|---|---|---|
| trait methods with no node | 70 / 112 (62%) | **0 / 112** |
| nodes | 9,426 | 9,574 |
| edges | 28,638 | 29,177 |

## Tests

Full suite unchanged: **17 failed / 5296 passed / 93 skipped**, identical with and without this commit. The 17 failures are pre-existing on `v8` and all in `test_skillgen.py`; I verified by stashing the change and re-running.

## One note for reviewers verifying locally

The AST cache is keyed on file content, not extractor version, so `graphify-out/cache` has to be removed to observe any extractor change — `--force` re-runs extraction but still serves cached per-file results. That cost me a confusing round of "the fix does nothing" before I spotted it; it may be worth folding the extractor version into the cache key separately.

### DIFF
--- a/graphify/extractors/rust.py
+++ b/graphify/extractors/rust.py
@@ -182,6 +182,27 @@ def extract_rust(path: Path) -> dict:
                     function_bodies.append((func_nid, body))
             return
 
+        if t == "function_signature_item":
+            # `fn greet(&self) -> String;` — a trait method with no body. This node type
+            # had no branch at all, so a trait's required methods were unreachable: the
+            # only method nodes in a graph came from impl blocks, and a trait with no
+            # implementor in the corpus contributed none. Mirrors function_item, minus
+            # the body walk (there is no body).
+            name_node = node.child_by_field_name("name")
+            if name_node:
+                func_name = _read_text(name_node, source)
+                line = node.start_point[0] + 1
+                if parent_impl_nid:
+                    func_nid = _make_id(parent_impl_nid, func_name)
+                    add_node(func_nid, f".{func_name}()", line)
+                    add_edge(parent_impl_nid, func_nid, "method", line)
+                else:
+                    func_nid = _make_id(stem, func_name)
+                    add_node(func_nid, f"{func_name}()", line)
+                    add_edge(file_nid, func_nid, "contains", line)
+                emit_param_return_refs(node, func_nid, line)
+            return
+
         if t in ("struct_item", "enum_item", "trait_item"):
             name_node = node.child_by_field_name("name")
             if name_node:
@@ -292,6 +313,16 @@ def extract_rust(path: Path) -> dict:
                                             continue
                                         type_node = field.child_by_field_name("type")
                                         _emit_enum_type(type_node, field.start_point[0] + 1)
+                if t == "trait_item":
+                    # The methods a trait declares are its contract, and they were not
+                    # reached: this branch returns below, so nothing walked the body.
+                    # Descend the same way impl_item does, attributing each method to
+                    # the trait node — so `explain <Trait>` can list what an
+                    # implementor must provide.
+                    body = node.child_by_field_name("body")
+                    if body:
+                        for child in body.children:
+                            walk(child, parent_impl_nid=item_nid)
             return
 
         if t == "impl_item":


safishamsi:
Shipped in **v0.9.56** — landed on `v8` via cherry-pick with your authorship preserved. Thanks @santoshpy — Rust trait method declarations (signature-only and default-bodied) are now extracted, distinct from their impl definitions. Release: https://github.com/Graphify-Labs/graphify/releases/tag/v0.9.56

PR #3369, opened 2026-09-06
fix(paths): bound atomic temp filename length on Windows

### LINKED ISSUE #3351: fix(paths): atomic-write temp filename can exceed Windows MAX_PATH for an already-capped long label, breaking Obsidian/Canvas export
## Summary

`_atomic_replace()` (`graphify/paths.py:29-48`) writes a temp file in the
same directory as the destination, then `os.replace()`s it into place:

```python
fd, tmp = tempfile.mkstemp(dir=str(real.parent), prefix=f".{real.name}.", suffix=".tmp")
```

The prefix embeds the **entire destination filename**, not just a fixed-size
marker. `graphify/export.py`'s Obsidian/Canvas writers already go out of
their way to keep the destination name Windows-safe: `stem_filename_budget()`
(`graphify/paths.py:363`) computes, **per call**, exactly how many characters
of stem still fit before `MAX_PATH` (260) is reached, given *this* output
directory's own length --
`(259 - len(output_dir) - len(os.sep) - reserve - len(".md"))` -- and
`_obsidian_safe_stem`/`_cap_filename` truncate (with an 8-char dedup hash) to
that exact budget. That accounts for the destination path itself, written
directly -- it has no `reserve` slot for anything an *unrelated* caller might
wrap around the write later.

Commit `74e1424` ("fix(export): write graph.canvas and Obsidian notes
atomically", #3282, merged 2026-09-03, first released in v0.9.54) routed
`to_obsidian`/`to_canvas` through `paths.write_text_atomic`/
`write_json_atomic` to close a real, separate bug (a concurrent reader, e.g.
`git hash-object` or a file watcher, could observe a truncated file mid-write).
That fix is correct on its own terms, but it silently added `_atomic_replace`'s
temp-name wrapper (`"." + <name> + "." + <8 random chars> + ".tmp"`, ~14 extra
bytes) **on top of** a destination name that was already budgeted right up to
the `MAX_PATH` line, with no headroom reserved for the wrapper. On Windows,
once the temp directory's own path is non-trivial (any real vault path, or
just a moderately nested `pytest` tmp dir), the combined path silently
crosses 260 chars and `tempfile.mkstemp` raises a bare `FileNotFoundError` --
not a descriptive "filename too long" error, because Windows' behavior for a
too-long path (without long-path opt-in) is to report the path as not found,
not invalid.

### Repro (upstream's own test)

```python
# tests/test_obsidian_filename_cap.py -- fails on a stock v0.9.54 checkout,
# Windows, default (no long-path) filesystem settings.
def test_obsidian_wikilink_resolves_after_truncation(tmp_path):
    long_label = "w" * 300
    G, comms = _graph([long_label, "neighbor"])
    to_obsidian(G, comms, str(tmp_path))  # raises FileNotFoundError from mkstemp
```

## Traceback

```
================================== FAILURES ===================================
______________ test_obsidian_wikilink_resolves_after_truncation _______________

tmp_path = WindowsPath('C:/Users/.../pytest-of-.../pytest-282/test_obsidian_wikilink_resolve0')

    def test_obsidian_wikilink_resolves_after_truncation(tmp_path):
        long_label = "w" * 300
        G, comms = _graph([long_label, "neighbor"])
>       to_obsidian(G, comms, str(tmp_path))

tests\test_obsidian_filename_cap.py:52:
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
graphify\export.py:818: in to_obsidian
    if _owned_write(fname, "\n".join(lines)):
graphify\export.py:729: in _owned_write
    write_text_atomic(target, content)
graphify\paths.py:93: in write_text_atomic
    _atomic_replace(path, lambda f: f.write(text))
graphify\paths.py:48: in _atomic_replace
    fd, tmp = tempfile.mkstemp(dir=str(real.parent), prefix=f".{real.name}.", suffix=".tmp")
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

dir = '...\\pytest-282\\test_obsidian_wikilink_resolve0'
pre = '.wwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwwww_4ac99e5c.md.'
suf = '.tmp'

    def _mkstemp_inner(dir, pre, suf, flags, output_type):
        for seq in range(TMP_MAX):
            name = next(names)
            file = _os.path.join(dir, pre + name + suf)
            try:
>               fd = _os.open(file, flags, 0o600)
E               FileNotFoundError: [Errno 2] No such file or directory: '...\\pytest-282\\test_obsidian_wikilink_resolve0\\.wwww[...300 total 'w's...]www_4ac99e5c.md.ht3o753m.tmp'

C:\...\Lib\tempfile.py:256: FileNotFoundError
```

The arithmetic behind this exact traceback, worked out from the code above:
this test's `tmp_path` is 94 characters, so `stem_filename_budget` computes
`259 - 94 - 1 (sep) - 5 (reserve) - 3 (".md") = 156`, and `_cap_filename`
truncates the label to a 156-byte stem (147 kept characters + `"_"` + an
8-char dedup hash). Destination path = `94 + 1 + 156 + 3(".md") = 254` --
**by design**, just inside the 259 budget. `_atomic_replace`'s wrapper then
adds `"." + "." + <8 random chars> + ".tmp"` = 14 more characters on top,
for a total of 268 -- past `MAX_PATH`. The destination name was deliberately
budgeted to just barely fit; the atomic-write wrapper is what pushes it
over, and nothing in the call chain accounts for it.

## Verify

Confirmed via `git worktree` on a clean `v0.9.53` checkout (before commit
`74e1424`): the identical test passes -- 21/21 green across both affected
test files. On a clean `v0.9.54` checkout (same test, same fixture, no
other changes), it fails with the traceback above -- isolates the
regression to that one commit, not to anything downstream.

Five tests hit this in total on `v0.9.54`:
`test_obsidian_filename_cap.py::test_obsidian_long_ascii_label_does_not_crash`,
`::test_obsidian_distinct_long_labels_sharing_prefix_do_not_collide`,
`::test_obsidian_wikilink_resolves_after_truncation`,
`test_export_path_length.py::test_obsidian_writes_paths_inside_max_path`,
`::test_canvas_card_targets_exist_inside_max_path`.

## Suggested fix

`stem_filename_budget()`'s own docstring already anticipates room being
needed for "caller-added prefixes/collision suffixes" via its `reserve`
parameter -- one option is having `export.py`'s call sites pass a bigger
`reserve` that accounts for `_atomic_replace`'s 14-character wrapper. But
`write_text_atomic`/`write_json_atomic` have callers beyond `export.py`
(`cli.py`, `detect.py`, `exporters/html.py`, `global_graph.py`,
`install.py`) that don't necessarily go through `stem_filename_budget` at
all, so patching `reserve` only closes the gap for the two or three call
sites that already know about it -- any other caller writing a long
filename through the same atomic path is equally exposed.

A more general fix: don't grow the temp filename with the length of the
destination name at all -- decouple the two so no destination-side budget
(however it's computed) is ever load-bearing for the *wrapper's* length. A
short, fixed-length temp name derived from a hash of `real.name` (or just a
random name with no destination-derived component), still written into the
same directory for the atomic same-filesystem rename:

```python
tmp_prefix = f".gfy-{hashlib.md5(real.name.encode()).hexdigest()[:8]}-"
fd, tmp = tempfile.mkstemp(dir=str(real.parent), prefix=tmp_prefix, suffix=".tmp")
```

This keeps the temp file identifiable (for cleanup-on-crash debugging)
without its length scaling with the destination name's length -- every
caller's own `MAX_PATH` budgeting then only has to account for its own
final filename, not for whatever atomic-write machinery wraps around it
afterward. The alternative (opt into Windows long-path support,
`\\?\`-prefixed paths) is a bigger, riskier surface change affecting every
path operation in the codebase, not just this one write path -- not
recommended as the fix here.

## Impact if unfixed

Any Obsidian/Canvas export of a graph with a long node label (a real,
non-adversarial scenario -- long fully-qualified class names, long file
paths turned into labels, etc.) can silently fail on Windows with a
misleading `FileNotFoundError`, even though the exact scenario this cap was
built to handle (`_cap_filename`, existing since v0.9.45/#1094) is precisely
what's failing here -- the cap fix and this atomicity fix each independently
made sense, but composed to reintroduce the class of bug the cap was there
to prevent.

## Environment

- graphifyy 0.9.54, Windows 11, Python 3.12.13, pytest tmp dirs under
  `C:\Users\<user>\AppData\Local\Temp\pytest-of-*\...`.

hopstreax:
## Summary

Fixes #3351 by preventing atomic-write temporary filenames from exceeding Windows `MAX_PATH`.

## Problem

`_atomic_replace()` previously generated temporary files using the destination filename as part of the temporary filename:

```python
prefix=f".{real.name}."
```

For long Obsidian and Canvas export filenames, this caused the temporary path to become significantly longer than the already-capped destination path.

As a result, exports could fail on Windows even though the final destination path itself was within the supported `MAX_PATH` limit.

## Fix

Replace the destination-dependent prefix with a fixed, bounded Graphify-specific prefix:

```python
prefix=".gfy-"
```

Temporary files now have a constant-length form such as:

```text
.gfy-XXXXXXXX.tmp
```

The temporary file is still created in the destination's parent directory, so the existing atomic `os.replace()` behavior and same-filesystem guarantees are preserved.

No changes were made to exporter path budgeting or `stem_filename_budget()`.

## Regression Coverage

Added tests covering:

* Temporary filenames remain bounded and do not embed long destination filenames.
* Atomic writes succeed at the Windows `MAX_PATH` boundary.
* Temporary `.tmp` files are cleaned up after successful writes.

Existing Obsidian and Canvas path-length regression tests now pass as well.

## Validation

* `32 passed, 2 skipped` — atomic write / Canvas / path-length test suites
* `93 passed` — exporter and control-character test suites
* `git diff --check` — clean
* `graphify update .` — completed successfully

This keeps the fix localized to the atomic-write implementation while preserving the existing atomicity, symlink handling, permissions, and Windows error-handling behavior.

### DIFF
--- a/graphify/paths.py
+++ b/graphify/paths.py
@@ -45,7 +45,7 @@ def _atomic_replace(path: "str | Path", write_fn) -> None:
     # atomic rename) and the replace writes through the link, not over it.
     real = Path(os.path.realpath(str(path)))
     real.parent.mkdir(parents=True, exist_ok=True)
-    fd, tmp = tempfile.mkstemp(dir=str(real.parent), prefix=f".{real.name}.", suffix=".tmp")
+    fd, tmp = tempfile.mkstemp(dir=str(real.parent), prefix=".gfy-", suffix=".tmp")
     try:
         with os.fdopen(fd, "w", encoding="utf-8") as f:
             write_fn(f)
@@ -79,7 +79,7 @@ def _atomic_replace(path: "str | Path", write_fn) -> None:
             # The temp was chmod'd to match the destination above, so when the
             # destination is read-only the temp is too — and Windows refuses to
             # unlink a read-only file. Clear the bit and retry, or every failed
-            # write leaks a `.graph.json.*.tmp` into the output directory.
+            # write leaks a `.gfy-*.tmp` into the output directory.
             try:
                 os.chmod(tmp, stat.S_IWRITE)
                 os.unlink(tmp)
--- a/tests/test_atomic_writes.py
+++ b/tests/test_atomic_writes.py
@@ -176,3 +176,53 @@ def test_write_json_atomic_ensure_ascii_false_preserves_utf8(tmp_path):
     assert "Wörker 数据" in raw  # raw UTF-8, not \\uXXXX escapes
     assert "\\u" not in raw
     assert json.loads(raw) == {"label": "Wörker 数据"}
+
+
+def test_atomic_replace_temp_name_is_bounded_and_does_not_embed_long_filename(tmp_path, monkeypatch):
+    """Regression #3351: temp filename must not grow with destination filename."""
+    import tempfile
+    from pathlib import Path
+    from graphify.paths import _WINDOWS_MAX_PATH, _atomic_replace
+
+    base = str(tmp_path.resolve())
+    # Sized to be a long name (> 100 chars) while fitting within MAX_PATH
+    remaining = (_WINDOWS_MAX_PATH - 1) - len(base) - 1
+    long_name = "a" * (remaining - 10) + ".md"
+    assert len(long_name) >= 100
+    target = tmp_path / long_name
+
+    intercepted_temps = []
+    real_mkstemp = tempfile.mkstemp
+
+    def tracking_mkstemp(*args, **kwargs):
+        fd, path = real_mkstemp(*args, **kwargs)
+        intercepted_temps.append(path)
+        return fd, path
+
+    monkeypatch.setattr(tempfile, "mkstemp", tracking_mkstemp)
+    _atomic_replace(target, lambda f: f.write("hello"))
+
+    assert target.exists()
+    assert len(intercepted_temps) == 1
+    temp_path = Path(intercepted_temps[0])
+
+    assert len(temp_path.name) <= 25, f"temp filename was unexpectedly long: {temp_path.name}"
+    assert temp_path.name.startswith(".gfy-")
+    assert temp_path.name.endswith(".tmp")
+    assert "a" * 50 not in temp_path.name, "temp filename embedded the long destination name"
+
+
+@pytest.mark.skipif(os.name != "nt", reason="Windows MAX_PATH test")
+def test_write_text_atomic_succeeds_near_windows_max_path(tmp_path):
+    """Regression #3351: atomic write to a path at Windows MAX_PATH limit must succeed."""
+    from graphify.paths import _WINDOWS_MAX_PATH, write_text_atomic
+
+    base = str(tmp_path.resolve())
+    remaining = (_WINDOWS_MAX_PATH - 1) - len(base) - 1
+    filename = ("x" * (remaining - 4)) + ".txt"
+    target = tmp_path / filename
+    assert len(str(target)) == _WINDOWS_MAX_PATH - 1
+
+    write_text_atomic(target, "content-at-max-path")
+    assert target.read_text(encoding="utf-8") == "content-at-max-path"
+    assert not any(p.name.endswith(".tmp") for p in tmp_path.iterdir())


safishamsi:
Shipped in **v0.9.56** — landed on `v8` via cherry-pick with your authorship preserved. Thanks @hopstreax — the atomic-write temp filename is bounded so exports near the Windows MAX_PATH limit no longer fail. Release: https://github.com/Graphify-Labs/graphify/releases/tag/v0.9.56

Issue #2309, opened 2026-07-29
graphify path (0.9.30) still prints reversed hops, and answers undirected connectivity while being documented as dependency reach (regression of #849; relates #829/#2074)


mgcronin:
**Package:** `graphifyy` 0.9.30 (also reproduced on 0.5.6)
**Environment:** macOS 15 (Darwin 25.3.0), Python 3.14, `uv tool install graphifyy`
**Graph:** a real TypeScript monorepo — 13,251 nodes / 22,543 edges, built by CI with `graphify update`, `"directed": false` in `graph.json`

Related and previously closed: #849 (reversed arrows, v0.7.16), #829 (`build_from_json(directed=True)` unreachable from the CLI), #2074 (path label not backed by a stored edge, v0.9.21). This is the same family, still present at 0.9.30, plus a framing point I don't think those issues made explicit.

---

## 1. `path` prints a hop whose stored edge points the other way

```
$ graphify path "server/src/lib/analysis/clarification-enrichment.ts" \
                "server/src/lib/publishing/draft-generator.ts"
warning: source match was ambiguous (top score 145.424, runner-up 145.424)
Shortest path (3 hops):
  clarification-enrichment.ts --imports_from--> logger.ts --imports_from--> draft-generator.ts --imports_from--> publishing-draft-generator.test.ts
```

The middle hop does not exist. Reading the two edges straight out of my own `graph.json` via their `_src` / `_tgt` fields:

```
clarification-enrichment.ts --imports_from--> logger.ts   (EXTRACTED, L35)
draft-generator.ts         --imports_from--> logger.ts   (EXTRACTED, L28)
```

Both point **into** `logger.ts`. `logger.ts` has exactly three outbound edges in the whole graph, all `contains` to its own functions (`isSensitiveKey`, `redact`, `createChildLogger`) — and zero outbound `imports_from`. The source file imports only `winston`. So `logger.ts --imports_from--> draft-generator.ts` is the stored edge `draft-generator.ts --imports_from--> logger.ts` printed backwards, which is #849's renderer bug at 0.9.30.

## 2. The deeper issue: `path` answers connectivity, but is documented as dependency reach

Fixing the arrow rendering alone would not fix the answer. Rendered honestly the route is:

```
clarification-enrichment.ts --imports_from--> logger.ts <--imports_from-- draft-generator.ts
```

[...]

PR #3373, opened 2026-09-06
fix(serve): MCP query_graph traverses undirected, so a seed with no outgoing edges reaches its callers

kuchtgpt-svg:
## Summary

Over MCP, `query_graph` answers with the seed node alone whenever the seed has no outgoing edges — a leaf function that is only ever called, imported and contained. The CLI `query` on the same `graph.json` returns the callers, the test and the neighbouring modules.

Measured on a real graph (3910 nodes / 7991 edges, TypeScript + SQL), seed `taktTime()`:

| surface | result |
| --- | --- |
| CLI `graphify query "taktTime"` | 57 nodes, 160 edges |
| MCP `query_graph` (`question: "taktTime"`) | 1 node — the seed |

Unchanged across `depth` 2/3, `token_budget` 400/2000, and with an explicit `context_filter: ["call"]`. The header says `1 nodes found`, not truncated: the traversal never expanded.

## Root cause

`_load_graph` forces `directed: True` so renderers can recover stored arc order (#2309), and `_bfs` / `_dfs` expand through `G.neighbors()`, which on a `DiGraph` yields **successors only**. `taktTime()` has six incoming edges (`calls` ×3, `imports` ×2, `contains`) and zero outgoing, so the frontier is empty after the seed.

The CLI `query` avoids this on purpose — its loader keeps the graph undirected and stashes `_src`/`_tgt` per edge so `_subgraph_to_text` still renders caller→callee (#2080; the comment in `cli.py` names exactly this failure mode: "forcing a DiGraph would make `G.neighbors()` return successors only, silently dropping every caller-side result for a seed with no outgoing edges"). `_query_graph_text` is shared, but the MCP path feeds it the directed graph from `_load_graph`. Same shape as #1174, where `affected` had missed the fix the other read surfaces got.

The other MCP tools are unaffected: `get_neighbors` walks successors and predecessors explicitly, `shortest_path` builds its own graph from `_src`/`_tgt`.

## Fix

`_traversal_view(G)`: an undirected copy for the traversal with `_src`/`_tgt` stashed on each edge (existing markers win, as in the CLI loader), used by `_query_graph_text` in place of `G` for the BFS/DFS and the context filter. An undirected input is returned as-is, so the CLI path is byte-for-byte unchanged. `_load_graph` is untouched — `get_neighbors` and `shortest_path` rely on direction.

After the change, on the graph above: MCP `query_graph` → 57 nodes / 160 edges; the `NODE` and `EDGE` lines are identical to the CLI's on six different questions (including edge direction).

## Tests

`tests/test_query_mcp_direction.py` seeds one `calls` edge from the callee side through `_load_graph` (the MCP path), in both traversal modes and with an explicit context filter; on the previous code these fail with `1 nodes found`. Also covered: caller-side seeding unchanged; `MultiDiGraph` keeps parallel and mutual edges (fails if the stored keys are carried over — on an undirected multigraph a key is unique per unordered pair); undirected input is returned as-is; mutual arcs on a plain `DiGraph` fold into one edge exactly as the CLI's undirected load of `graph.json` folds them, so the two surfaces stay identical.

```
uv run pytest tests/test_query_mcp_direction.py tests/test_serve.py tests/test_query_cli.py tests/test_query_induced_edges.py tests/test_query_names_its_graph.py
183 passed
```

Full suite on this box (Windows 11, Python 3.14, `uv sync --extra mcp --extra sql`): **31 failed / 5356 passed / 62 skipped** on the branch, and the same 31 fail on pristine `v8` (`c9f9901`): `test_hooks`, `test_install*`, `test_skillgen`, `test_watch`, `test_ollama_retry_cap`, `test_non_regular_files`, `test_uninstall_scope`, `test_languages`, `test_ignore_file_encoding`, `test_merge_chunks_validation` — none touch `serve.py`; they look Windows/uv-tool-environment bound. The pristine run additionally failed `test_incremental_mtime_collision::test_same_size_rewrite_in_one_tick_is_requeued` once (mtime-tick timing).

`ruff check` clean on both files.

### DIFF
--- a/graphify/serve.py
+++ b/graphify/serve.py
@@ -1230,6 +1230,49 @@ def _display_graph_path(graph_path: str) -> str:
         return str(graph_path)
 
 
+def _traversal_view(G: nx.Graph) -> nx.Graph:
+    """Undirected copy of `G` for BFS/DFS, with true direction kept per edge.
+
+    `_load_graph` forces `directed: True` so renderers can recover stored arc
+    order (#2309), and on a DiGraph `G.neighbors()` yields successors only. The
+    query traversals rely on `neighbors()`, so a seed with no outgoing edges — a
+    leaf function that is only ever called, imported and contained — expanded
+    to nothing: `query_graph` over MCP answered with the seed alone while the
+    CLI `query`, which loads the same file undirected, returned the callers,
+    the test and the neighbouring modules. Every other MCP tool was unaffected:
+    `get_neighbors` walks successors and predecessors explicitly, and
+    `shortest_path` builds its own graph from `_src`/`_tgt`.
+
+    Mirrors the CLI loader: traverse undirected, stash `_src`/`_tgt` on each
+    edge so `_subgraph_to_text` still renders caller->callee regardless of the
+    side the traversal reached the edge from. Markers already present on an
+    edge win, for the same reason as in the CLI (#2309). An undirected input is
+    returned as-is, so the CLI path is unchanged.
+
+    Mutual arcs `u->v` and `v->u` (mutual recursion, a circular import) fold
+    into one undirected edge on a plain `DiGraph` input, the later one winning
+    — the same fold the CLI loader performs when `json_graph.node_link_graph`
+    reads the undirected on-disk graph into an `nx.Graph`, which is what keeps
+    the two surfaces' output identical. The renderer shows one edge per pair
+    in any case. On a `MultiDiGraph` input the copy is a `MultiGraph` and the
+    stored keys are not carried over: a key is unique per unordered pair on an
+    undirected multigraph, so mutual arcs that happen to share a key would
+    fold there too; letting networkx assign the keys keeps both, and nothing
+    downstream reads the key.
+
+    A fresh copy per query rather than a cached one: `_filter_graph_by_context`
+    already copies per query when a filter applies, and the copy shares node
+    data dicts with `G`, so only the edge dicts are duplicated.
+    """
+    if not G.is_directed():
+        return G
+    H = nx.MultiGraph() if G.is_multigraph() else nx.Graph()
+    H.graph.update(G.graph)
+    H.add_nodes_from(G.nodes(data=True))
+    for u, v, d in G.edges(data=True):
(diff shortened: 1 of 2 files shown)


kuchtgpt-svg:
**[агент]** _(this account is operated by an agent on the maintainer's behalf; the marker is a house rule)_

On the advisory **"Directed reciprocal edges are collapsed from query_graph results"** (`_traversal_view`, plain `DiGraph` input):

The fold is deliberate and pre-existing rather than introduced here. The CLI `query` loads `graph.json` through `json_graph.node_link_graph` into an `nx.Graph`, so a reciprocal pair `u->v` / `v->u` already becomes one undirected edge there, last link winning. Building the MCP traversal view the same way is what makes the two surfaces return identical `NODE`/`EDGE` lines — checked on a 3910-node / 7991-edge graph across six questions.

Two facts that bound the impact:

- Keeping both arcs would change nothing visible today: `_subgraph_to_text` renders the first parallel edge only (`next(iter(G[u][v].values()))`), and `_complete_induced_edges` dedups per unordered pair on an undirected graph. Preserving both directions is a renderer change, not a loader change, and it would apply to the CLI equally.
- On the real graph above there are zero reciprocal pairs (counted over `links`), so the case did not arise in the measurements; it is real for mutual recursion or circular imports in other corpora.

If you would rather have the view keep both arcs now (always build an `nx.MultiGraph` for a directed input, so a later renderer change can show them), that is a two-line change and I am happy to push it — say the word. Otherwise I would leave parity with the CLI as is and open a separate issue for rendering parallel/reciprocal edges on both surfaces.

On the coupling note for `_query_graph_text`: the change adds one callee (`_traversal_view`); the 25 callers / 10 callees are the function's existing shape.

CI shows `action_required` — first-time contributor run waiting for approval.

safishamsi:
Shipped in **v0.9.56** — landed on `v8` via cherry-pick with your authorship preserved. Thanks @kuchtgpt-svg — an MCP query_graph seeded on a node with only incoming edges now traverses undirected and reaches its callers. Release: https://github.com/Graphify-Labs/graphify/releases/tag/v0.9.56

PR #3379, opened 2026-09-07
Add source_location to Dart extraction

### LINKED ISSUE #3365: Dart extractor emits source_location: None for every node — 0% on 3 repos, ~25k nodes unciteable (regex extractor hardcodes the field)
## Summary

Every node the Dart extractor emits has `source_location: None`. Not "most" — all of
them, in every Dart file, because `extractors/dart.py` hardcodes the field at all three
node/edge construction sites. Dart graphs therefore cannot cite a line, and no
downstream consumer can verify a node against the file it claims to come from.

This is worth separating from #3307, which asks for `source_location` on doc/YAML/markdown
and states as its premise that *"code already gets this for free from AST (line numbers
are known at parse time)"*. Dart is a code language with an extractor, and it gets 0%.

## Minimal reproduction

Two structurally identical files, ten lines each, one `graphify extract . --code-only`
run:

```dart
// sample.dart
class Greeter {
  final String name;
  Greeter(this.name);

  String greet() {
    return 'hello $name';
  }
}

int add(int a, int b) => a + b;
```

```python
# sample.py
class Greeter:
    def __init__(self, name):
        self.name = name

    def greet(self):
        return f"hello {self.name}"


def add(a, b):
    return a + b
```

Resulting nodes:

```
.dart: 5 nodes, 0 with source_location      .py: 5 nodes, 5 with source_location
  None  sample.dart                            L1  Greeter
  None  add                                    L5  .greet()
  None  Greeter                                L2  .__init__()
  None  greet                                  L9  add()
  None  name                                   L1  sample.py
```

Measured across nine repos (all `--code-only`, graphifyy 0.9.55):

| repo | dominant language | declared nodes | with `source_location` |
|---|---|---|---|
| Plugins | Dart | 286 | **0 (0%)** |
| Board | Dart | 531 | **0 (0%)** |
| Editor | Dart | 5,625 | **0 (0%)** |
| App | Dart + Rust | 31,958 | 6,961 (22% — exactly the Rust half) |
| Cloud | Rust | 6,304 | 6,223 (99%) |
| Web | TS/TSX | 10,143 | 10,134 (100%) |
| Auth | Go | 3,655 | 3,583 (98%) |
| ops | Python | 707 | 707 (100%) |

~25,000 Dart nodes in one repo with no position at all.

## Root cause

`extractors/dart.py` is regex-based (`import re`, 31 `re.*` calls, zero `tree_sitter`
references — its docstring says "using regex"), so there is no `start_point` to read,
and the field is written as a literal:

```python
nodes.append({"id": file_nid, "label": path.name, "file_type": "code",
              "source_file": str(path), "source_location": None})          # L54

def add_node(nid, label, ftype="code", source_file=str(path)) -> None:
    if nid not in defined:
        nodes.append({"id": nid, "label": label, "file_type": ftype,
                      "source_file": source_file, "source_location": None})  # L61

edge = {..., "source_file": str(path), "source_location": None, ...}         # L67
```

`add_node` takes no line parameter, so callers such as `add_node(class_nid, class_name)`
(L145) have nowhere to pass one.

## Suggested fix, and a trap in it

Each match already knows its offset, so the line is `src[:m.start(1)].count("\n") + 1`.
Two details matter, both verified against the file below:

1. **Anchor on the capture group, not the match.** The class pattern begins `^\s*`, so
   `m.start()` sits *before* the preceding newlines and reports the line too early.
2. **`_comment_replace` deletes comments** (returns `""`), so offsets in the cleaned
   text no longer correspond to the original file. Blanking a comment to an equal number
   of newlines keeps every later line correct.

```dart
/*
 * A multi-line header comment.
 * It spans several lines.
 */
class AfterComment {      // real line 5
  void method() {}
}
```

```
true line                            : 5
today (comments deleted, m.start(1)) : 2     <- off by the comment's height
newlines preserved + m.start(1)      : 5     <- correct
```

So: return `"\n" * token.count("\n")` instead of `""` for comment tokens, thread a
`line` argument through `add_node`/`add_edge`, and compute it from `m.start(1)` at each
`re.finditer` site.

## Why it matters beyond citations

We cross-validate graphs against the source they describe (file-coverage ledger, then
per-node line anchoring, then recall against an independent parser). For Rust, Go, TS
and Python the line anchor confirms a node really is the symbol it claims to be at the
line it claims. For Dart that check cannot run at all, so ~78% of our largest repo's
graph is unverifiable — it reports `n/a` rather than a pass, and a Dart-only repo can
never be more than partially validated.

## Environment

graphifyy 0.9.55 (uv tool, Python 3.12.12), macOS 15 arm64. `graphify extract --code-only`.

ayushcodes10:
## What

`extractors/dart.py` hardcoded `source_location: None` at every node and edge construction site, so a Dart graph could cite zero lines no matter how many declarations it had. Threads a real line number through `add_node`/`add_edge` from each match's own offset instead.

## The trap (verified, matches the issue)

Two details had to be right together, or line numbers would be silently wrong rather than missing:

1. **Anchor on the capture group, not the whole match.** The class pattern begins `^\s*`, so `m.start()` sits before any preceding blank lines and reports the line too early.
2. **Comment stripping must preserve newline count.** The existing pass deleted a comment token outright (`return ""`), so a multi-line comment shifted every later offset up by the comment's own height. Fixed to blank to the same number of newlines instead.

Verified both against the reporter's own repro:

```
true line                            : 5
before (comments deleted, capture-group anchor): 2   <- off by the comment's height
after (newlines preserved + capture-group anchor): 5   <- correct
```

## Scope

Sourceless reference stub nodes (an inherited class, a mixin, an import target — anything the file only references rather than defines, already marked `source_file=None`) keep `source_location: None`. There's no single line in this file for a symbol that lives elsewhere. Every node this file actually defines, and every edge attributed to a real declaration site, now carries its real line — verified this covers every `add_node`/`add_edge` call site in the file (classes, mixins, extensions, typedefs, variables, methods, annotations, imports/exports, and all the framework-specific reference patterns).

Fixes #3365.

### DIFF
--- a/graphify/extractors/dart.py
+++ b/graphify/extractors/dart.py
@@ -26,9 +26,18 @@ def extract_dart(path: Path) -> dict:
     def _comment_replace(match: re.Match) -> str:
         token = match.group(0)
         if token.startswith("/"):
-            return ""
+            # Blank to the SAME number of newlines the comment spanned, not
+            # to an empty string: every offset computed against src_clean
+            # below (including every source_location line number) is only
+            # valid if src_clean's line count tracks the original file's.
+            # Deleting a multi-line comment outright shifted every later
+            # line number up by the comment's own height (#3365).
+            return "\n" * token.count("\n")
         return token
     src_clean = comment_string_pattern.sub(_comment_replace, src)
+
+    def _line_at(offset: int) -> int:
+        return src_clean[:offset].count("\n") + 1
 
     stem = _file_stem(path)
     file_nid = _make_id(str(path))
@@ -55,16 +64,20 @@ def extract_dart(path: Path) -> dict:
     edges = []
     defined: set[str] = set()
 
-    def add_node(nid: str, label: str, ftype: str = "code", source_file: str | None = str(path)) -> None:
+    def add_node(nid: str, label: str, ftype: str = "code", source_file: str | None = str(path),
+                 line: int | None = None) -> None:
         if nid not in defined:
             nodes.append({"id": nid, "label": label, "file_type": ftype,
-                          "source_file": source_file, "source_location": None})
+                          "source_file": source_file,
+                          "source_location": f"L{line}" if line else None})
             defined.add(nid)
 
-    def add_edge(src_id: str, tgt_id: str, relation: str, weight: float = 1.0, context: str | None = None) -> None:
+    def add_edge(src_id: str, tgt_id: str, relation: str, weight: float = 1.0, context: str | None = None,
+                 line: int | None = None) -> None:
         edge = {"source": src_id, "target": tgt_id, "relation": relation,
                 "confidence": "EXTRACTED", "confidence_score": 1.0,
-                "source_file": str(path), "source_location": None, "weight": weight}
+                "source_file": str(path),
+                "source_location": f"L{line}" if line else None, "weight": weight}
         if context:
             edge["context"] = context
         edges.append(edge)
@@ -141,9 +154,13 @@ def extract_dart(path: Path) -> dict:
     class_pattern = r"^\s*(?:(?:abstract|sealed|base|interface|final|mixin)\s+)*(?:class|mixin|enum|extension\s+type)\s+(\w+)"
     for m in re.finditer(class_pattern, src_clean, re.MULTILINE):
         class_name = m.group(1)
+        # Anchor on the capture group, not the whole match: the pattern's
+        # leading ^\s* makes m.start() land before the blank/indent lines
+        # that precede the declaration, reporting the line too early (#3365).
+        class_line = _line_at(m.start(1))
         class_nid = _make_id(stem, class_name)
-        add_node(class_nid, class_name)
-        add_edge(file_nid, class_nid, "defines")
+        add_node(class_nid, class_name, line=class_line)
+        add_edge(file_nid, class_nid, "defines", line=class_line)
 
         # Manually parse extends/on, with, and implements in header to handle nested generics brackets balanced
         start_idx = m.end()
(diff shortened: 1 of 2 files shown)


ayushcodes10:
@safishamsi opened this against #3365 (Dart nodes had zero source_location, 0% across the reporter's real repos). Verified the fix against their exact repro and the multi-line-comment trap they flagged, plus checked every add_node/add_edge call site in the file for coverage. Happy to address any feedback.

ayushcodes10:
Looked into the bot finding on the bloc_event line offset (dart.py:293).

class_body = src_clean[brace_pos:end_pos] is a plain slice, no reindexing, so class_body[k] is always src_clean[brace_pos + k] and _line_at(brace_pos + em.start(1)) should be correct by construction. Stress tested directly: multiple handlers on different lines, a leading comment, brace characters inside string literals before and inside the class body, and a second class following an unrelated first one with its own brace-in-string. Every case reported the true source line exactly.

Could not reproduce the claimed issue, so treating it as a false positive rather than changing anything speculatively. Happy to revisit with a concrete repro if one turns up.

(Separately noticed, unrelated to this finding and predating this PR: a .add() call inside a method can get counted twice, once by the class level scan and once by the method level scan, since their bodies overlap. Not touching that here since it's out of scope for #3365.)

safishamsi:
Shipped in **v0.9.56** — landed on `v8` via cherry-pick with your authorship preserved. Thanks @ayushcodes10 — Dart extraction now stamps `source_location` on nodes and edges. Release: https://github.com/Graphify-Labs/graphify/releases/tag/v0.9.56

PR #3385, opened 2026-09-07
Finish Objective-C member calls across a repository boundary (#3152)

### LINKED ISSUE #3384: Objective-C member calls are never parked for the cross-repo merge (#3152)
### What happens

`merge-graphs` / `global add` finish a member call whose receiver type lives in
another repository for Java, C++, C# and Swift (#3152). Objective-C has a member-call
resolver with full receiver typing — `self`/`super`, an explicit class name, the
file's `Foo *f` local table, and the class's `@property`/ivar table with the
`inherits` walk from #3151 — but it parks nothing, so an ObjC corpus split across
repositories gets no cross-repo call edges at all.

The reason is one bail. `_resolve_objc_member_calls` ends each typing arm with

```python
type_defs = type_def_nids.get(_key(type_name), [])
if len(type_defs) != 1:  # ambiguous or absent -> bail (god-node guard)
    continue
```

so "declared nowhere in this corpus" — the parking case — is indistinguishable from
"declared more than once", which must stay dropped. Java splits the same guard into
a park arm and an ambiguity arm; the three ObjC arms do not.

### Minimal reproduction

```
repo app/src/App.m
    @interface App : NSObject
    @property (nonatomic, strong) Greeter *greeter;
    @end
    @implementation App
    - (void)run { [self.greeter greet]; }
    @end

repo lib/src/Greeter.m
    @interface Greeter : NSObject
    - (void)greet;
    @end
    @implementation Greeter
    - (void)greet {}
    @end
```

Build each repo, then `merge-graphs app/graphify-out/graph.json lib/graphify-out/graph.json`.

Observed: `grep unresolved_calls app/graphify-out/graph.json` finds nothing, and the
merged graph has no `context: "cross_repo"` edge. The same two files in one corpus do
produce the `-run -> -greet` call, so the receiver typing is not what is missing.

Expected: `App.m` parks `{"callee": "greet", "receiver_type": "Greeter", "lang": "objc"}`
and the merge emits the `calls` edge, as the Java/C++/C#/Swift equivalents do.

### Two ObjC-specific things the merge side has to handle

1. **`_callable_class` / `_callable` are absent from `extractors/objc.py`** (#3228), and
   `cross_repo_calls._index_declarations` indexes declarations by those markers, so
   every ObjC type is skipped even once parking exists.

2. **An ObjC method label keeps its `+`/`-` sigil** (`-greet`), where every other
   extractor writes `.greet()`. `_key` strips `.` and `()` only, so a parked selector
   `greet` never matches `-greet`. Normalizing the sigil away is the wrong fix: `.h` is
   in the C++ suffix set and would be in the ObjC one, so keeping the sigil is what
   keeps an ObjC `-greet` and a C++ `.greet()` from answering for each other inside one
   shared header. Trying both sigils and treating a class that declares `+greet` and
   `-greet` as ambiguous keeps them disjoint without a new field in the parked payload.

### Scope note

Measured on a 15,754-file ObjC corpus split into 26 units: of 12,866 sends whose
receiver is typed to a class absent from its own unit, 5,127 are framework classes
(UIKit/Foundation, in no unit), 1,911 are business classes declared in no unit at all,
2,634 find the class but the selector is not on it, 2,769 are ambiguous across units,
and **425 are genuinely cross-unit resolvable**. The builtin-globals filter the Swift
and C++ resolvers already apply keeps the framework names out of the parked payload.

PR: paired with this issue.

xiongjianxu:
Fixes #3384. Extends #3152 (cross-repo member calls) to Objective-C.

### The change

**Extraction (`graphify/extract.py`).** `_resolve_objc_member_calls` ended each of its
three typing arms with a single `if len(type_defs) != 1: continue`, so "the receiver's
type is declared nowhere in this corpus" — the parking case — was indistinguishable
from "declared more than once", which must stay dropped. Each arm now splits the guard
the way Java's does: zero declarations parks `{callee, receiver_type, lang: "objc",
line}` on the caller node, more than one still bails on the god-node guard. The
`self`/`super` arm is untouched: its type is the caller's own class, which is by
definition in this corpus.

A builtin receiver type is filtered out at the parking sites, mirroring what the Swift
and C++ resolvers already skip before resolving — a `NSString` or `DispatchQueue` is
not what another repo declares, so parking one would only let a same-named user class
answer for it.

**Merge (`graphify/cross_repo_calls.py`).** Adds `"objc": {".h", ".m", ".mm"}` to
`_LANG_SUFFIXES` and a `_member_keys(lang, callee)` hook next to the existing
`_member_relations`. An ObjC method label keeps its `+`/`-` sigil (`-greet`) where every
other extractor writes `.greet()`, and a parked selector does not record whether the
send was to a class or an instance, so both spellings are tried; a class declaring both
`+greet` and `-greet` counts as ambiguous and binds nothing.

The sigil is deliberately kept rather than normalized away: `.h` is in the C++ suffix
set too, and the sigil is what keeps an ObjC `-greet` and a C++ `.greet()` from
answering for each other inside one shared header. It also means `@protocol`
declarations self-exclude — they are labelled `<Greeter>`, which no parked receiver type
spells — so no extra guard is needed there.

**The `_callable` markers.** `_index_declarations` indexes declarations by
`_callable_class`, which `extractors/objc.py` never wrote, so every ObjC type would be
skipped. The first commit here is #3229 (`7a6b0e5`) cherry-picked unchanged so this
branch stands alone; if #3229 lands first, that commit drops out on rebase.

### Tests

`tests/test_cross_repo_member_calls.py`, +5 cases:

- a parked selector answers through either sigil;
- a class declaring both sigils binds nothing;
- an ObjC call does not bind to a C++ `.greet()` in a shared `.h`, and a C++ call does
  not bind to an ObjC `-greet` in one either;
- an ObjC `@protocol` does not answer a parked call;
- `objc-property-receiver` added to the end-to-end parametrization: two real repos
  extracted and merged, `[self.greeter greet]` across the boundary.

Full suite: 4 failed, 5321 passed, 93 skipped. The 4 failures are the pre-existing
`tests/test_ollama_retry_cap.py` ones, identical to the `origin/v8` baseline
(4 failed, 5309 passed, 93 skipped) on the same venv. `ruff check graphify tests`
passes.

### Ceiling, measured

On a 15,754-file ObjC corpus in 26 units, 12,866 sends have a typed receiver whose
class is absent from its own unit. Of those, 5,127 are framework classes, 1,911 are
business classes declared in no unit, 2,634 find the class but not the selector on it,
2,769 are ambiguous across units, and **425 are genuinely cross-unit resolvable**. This
PR is aimed at those 425; the builtin filter keeps the framework names from being parked
at all.

### DIFF
--- a/graphify/cross_repo_calls.py
+++ b/graphify/cross_repo_calls.py
@@ -42,6 +42,7 @@ _LANG_SUFFIXES: dict[str, frozenset[str]] = {
     "cpp": frozenset({".cpp", ".cc", ".cxx", ".hpp", ".hh", ".hxx", ".h", ".cu", ".cuh"}),
     "csharp": frozenset({".cs"}),
     "java": frozenset({".java"}),
+    "objc": frozenset({".h", ".m", ".mm"}),
     "swift": frozenset({".swift"}),
 }
 
@@ -57,7 +58,8 @@ def _key(label: object) -> str:
     Type labels are plain (``Greeter``) while method labels carry the extractor's
     decoration (``.greet()``). Case is preserved: every language that parks calls
     here is case-sensitive, and folding case would let `greeter` answer for
-    `Greeter`.
+    `Greeter`. Objective-C's ``+``/``-`` sigil is deliberately kept, so an ObjC
+    ``-greet`` and a C++ ``.greet()`` stay distinct members of a shared header.
     """
     return str(label or "").strip().removeprefix(".").removesuffix("()")
 
@@ -148,6 +150,20 @@ def _member_relations(lang: str) -> tuple[str, ...]:
     return ("method", "defines") if lang == "cpp" else ("method",)
 
 
+def _member_keys(lang: str, callee: str) -> tuple[str, ...]:
+    """Which member-index keys a parked callee may answer to.
+
+    An ObjC selector carries no class/instance distinction, so both sigils are
+    tried; a type declaring `+greet` and `-greet` both is an ambiguity, not a hit.
+    A selector cannot itself begin with a sigil, so stripping one first keeps an
+    entry that already carries it from asking for `--greet`.
+    """
+    if lang == "objc":
+        selector = callee.lstrip("+-")
+        return (f"-{selector}", f"+{selector}")
+    return (callee,)
+
+
 def link_cross_repo_member_calls(merged: "nx.Graph") -> int:
     """Emit `calls` edges for parked member calls another repo answers.
 
@@ -190,7 +206,9 @@ def link_cross_repo_member_calls(merged: "nx.Graph") -> int:
                 continue
             targets: list[str] = []
             for relation in _member_relations(lang):
-                targets = members_by_relation[relation].get((candidates[0], callee), [])
+                members = members_by_relation[relation]
+                for member_key in _member_keys(lang, callee):
+                    targets.extend(members.get((candidates[0], member_key), []))
                 if targets:
                     break
             if len(targets) != 1:
--- a/graphify/extract.py
+++ b/graphify/extract.py
@@ -4106,7 +4106,9 @@ def _resolve_objc_member_calls(
         captured; a dotted receiver like ``Foo.shared`` is never passed through,
         because ``_key`` would strip the dot and collide with a real ``FooShared``.
     An uninferable receiver is SKIPPED (no guess), so an ambiguous selector across
-    classes never fans out. ``_merge_decl_def_classes`` folds each @interface/@impl
+    classes never fans out. A receiver typed to a class this corpus declares nowhere
+    is parked on the caller for a merged graph to finish (#3152).
+    ``_merge_decl_def_classes`` folds each @interface/@impl
     pair into one node, so a paired class clears the single-definition guard.
     ``@protocol`` declarations are excluded from the receiver-type index: a protocol
     is a contract, not a message receiver, and ObjC keeps protocol and class names in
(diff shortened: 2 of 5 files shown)


xiongjianxu:
Addressed the one advisory finding ("already-sigiled Objective-C callees are double-prefixed", `_member_keys`).

Not reachable from graphify's own output — `_park_absent` takes the callee from the ObjC extractor's `raw_calls`, where the selector is the concatenation of the send's `method` identifiers and never carries a sigil. But the payload is read back out of `graph.json`, and a sigiled entry would have asked the member index for `--greet` and silently missed. `_member_keys` now strips a leading sigil before re-adding both, which an ObjC selector can never legitimately have, so the function is idempotent whatever the payload says. New case: `test_an_objc_callee_that_already_carries_a_sigil_still_binds`.

Also dropped a local test log (`baseline_0955.txt`) that a `git add -A` had swept into the commit; the branch was amended and force-pushed, so the diff is now the three intended files only.

Full suite after the change: 4 failed, 5322 passed, 93 skipped — the 4 are the pre-existing `tests/test_ollama_retry_cap.py` failures, same set as the `origin/v8` baseline on this venv. `ruff check graphify tests` passes.

The three inline findings are coupling-delta advisories on `link_cross_repo_member_calls` and two test functions; they reflect the new `_member_keys` call and the test helpers, not a behavior concern, so I left those as they are.

xiongjianxu:
Self-audit pass over the bot review. No code change on this branch: the two correctness findings do not reproduce, and the remaining notes are coupling deltas.

**Already handled — a callee arriving with its own `+`/`-` sigil.** `_member_keys` strips one with `lstrip("+-")` before trying `-sel` then `+sel`, so a parked entry that already carries a sigil cannot ask the index for `--greet`. `test_an_objc_callee_that_already_carries_a_sigil_still_binds` pins it.

**Does not reproduce — `@protocol` nodes contradicting the documented exclusion.** Two separate mechanisms keep them out. `_resolve_objc_member_calls` excludes them explicitly via `_is_protocol_declaration` (the extractor labels a protocol `<Name>`), and the merge-side index cannot match one either: `cross_repo_calls._key()` preserves the angle brackets, so `<Greeter>` is never the key a parked receiver type `Greeter` looks up. The `_callable_class` marker the extractor now stamps on protocols is read by `_is_type_like_definition`, which is label-based and ignores it — the marker exists so protocols are visible to the passes that index declarations at all, exactly as a Java/C# interface is.

**On the coupling deltas** for `link_cross_repo_member_calls` and the two new tests: the fan-out is the shape of the pass, one small helper per decision (suffix family, member relations, member keys, key normalisation), and the tests fan out because each one builds two repos and merges them. Splitting either to move the metric would make both harder to read.

Verification: `ruff check graphify tests` clean; full suite 5322 passed / 93 skipped, with only the four pre-existing `tests/test_ollama_retry_cap.py` env failures that `v8` also shows here.

xiongjianxu:

PR #3387, opened 2026-09-07
Park TS/JS member calls whose receiver type lives in another repo (#3152)

### LINKED ISSUE #3386: TS/JS: a member call whose receiver type lives in another repo is dropped, not parked (#3152)
### What happens

`_resolve_typescript_member_calls` binds `this.greeter.greet()` only when `Greeter` is
declared inside the same build. Split the two classes across two repositories and the
call disappears: the resolver has the receiver's type in hand, finds no declaration for
it, and drops the call without writing anything down. `graph.json` — the only artifact
`merge-graphs` and `global add` consume — records nothing, so no merge-time pass can
recover what was never written.

The merged two-repo graph is therefore missing exactly the edges that make it a call
graph. This is #3152 for TS/JS; Java, C++, C#, Swift and Objective-C already park.

### Reproduction

```
app/src/app.ts
    import { Greeter } from "greeter-pkg";
    export class App {
        constructor(private greeter: Greeter) {}
        run(): void { this.greeter.greet(); }
    }

lib/src/greeter.ts
    export class Greeter {
        greet(): void {}
    }
```

```console
$ graphify extract app  -o app/graphify-out
$ graphify extract lib  -o lib/graphify-out
$ graphify merge-graphs app/graphify-out/graph.json lib/graphify-out/graph.json --out merged.json
```

`merged.json` has no `App.run -> Greeter.greet` edge, and neither input `graph.json`
carries any trace of the attempt. Put both files in one build and the edge appears.

### Two things had to be settled for TS/JS specifically

**The uppercase-spelling arm must not park.** The resolver types a receiver two ways: a
per-file table (constructor parameter properties, local `new` bindings, bare
`type_identifier` annotations) and "the receiver starts with a capital, so it is the
type". The second is a spelling heuristic that a namespace alias (`import * as React`),
a default import and a plain const object all satisfy. `node_modules` is excluded from
the scan (`detect.py`), so the overwhelming majority of receivers that resolve to no
local declaration through that arm are npm names — parking them would hand the merge a
pile of bare names to match with no evidence behind them. Only table-typed receivers
park.

**One merge-side key for all eight suffixes.** The merge pass checks that the declaring
file's suffix belongs to the parked language, which is what stops a Java `Greeter` from
answering through a Python class of the same name. TS and JS cannot be split here: a TS
class legitimately answers a JS call site, so two keys would drop every TS↔JS cross-repo
call. `.ts .tsx .mts .cts .js .jsx .mjs .cjs` share one key.

### Also fixed on the way

The resolver was registered for `.ts .tsx .mts .cts .js .jsx` only, so a pure-ESM
(`.mjs`) or CommonJS-suffixed (`.cjs`) corpus never ran type-aware member-call
resolution at all — unrelated to merging, and visible in a single-repo build.

### Relation to the origin gate (#2553)

The gate rejects a name-only match when the caller cannot see the matched type. It runs
*after* the single-definition check, so it only ever vets a name that did match locally.
The park happens at the zero-declarations point, strictly upstream of it, so parking can
never override a gate rejection. There is a test pinning that.

PR: sends the calls to the merge and adds the tests.

xiongjianxu:
Closes #3386. Continues #3152 (#3385 did Objective-C).

### What changes

`_resolve_typescript_member_calls` now parks a member call whose receiver type is
declared nowhere in this build, instead of dropping it, and
`link_cross_repo_member_calls` finishes it once the graphs are composed.

Two commits, separable:

1. **`.mjs`/`.cjs` activation.** The resolver claimed `.ts .tsx .mts .cts .js .jsx`, so a
   pure-ESM corpus never ran type-aware member-call resolution at all. The suffix set is
   now one constant the resolver both activates on and parks by. Single-repo bug, has its
   own test that fails without the fix.
2. **The parking.** `metadata.unresolved_calls` on the caller (names only — no node ids,
   which the #1529 remap and repo prefixing both rewrite), plus the `"typescript"` entry
   in `_LANG_SUFFIXES`.

### The two precision calls, and why

**Only table-typed receivers park.** The resolver types a receiver either from the
per-file table (constructor parameter properties, local `new` bindings, bare
`type_identifier` annotations) or from spelling — an initial capital is taken to name the
type. That second arm is satisfied by `import * as React`, by default imports and by
plain const objects, and since `node_modules` is not scanned, most receivers that reach
the zero-declarations point through it are npm names. Parking them would send bare names
to the merge with nothing behind them, so the park is gated on `not type_qualified`.
`test_a_namespace_style_receiver_is_not_parked` pins it.

**One `_LANG_SUFFIXES` key over all eight suffixes.** A TS class legitimately answers a JS
call site, so splitting TS from JS would drop every TS↔JS cross-repo call.
`test_a_typescript_call_binds_across_the_whole_js_family` runs the four representative
suffixes.

The park is gated on a JS/TS suffix rather than a `lang` tag because TS `raw_calls` carry
no tag and this resolver iterates every language's — the same shape the Swift resolver
uses.

### Interaction with the origin gate (#2553)

The gate runs after the single-definition check, so it only ever vets a name that did
match locally; the park sits at the zero-declarations point, strictly upstream and
disjoint. `test_a_type_the_origin_gate_rejected_is_not_parked` builds the gate's own case
(a local `class Repo` plus `import type { Repo } from 'external-pkg'`) and asserts nothing
is parked, so a gate rejection cannot be laundered through the merge.

### Known recall cost

`interface_declaration`, `enum_declaration` and `type_alias_declaration` all carry
`_callable_class`, so a `type Greeter = …` alias in a third repo makes a genuine class
match ambiguous and the merge emits nothing. That fails closed, which is the bar the
existing resolvers set, and narrowing it belongs with whatever decides that a type alias
is not a call target.

### Verification

- `tests/test_cross_repo_member_calls.py` + `tests/test_ts_receiver_member_calls.py`: 38 passed.
- Full suite: 5319 passed, 93 skipped, with `tests/test_ollama_retry_cap.py` deselected —
  its 4 failures reproduce unchanged on `v8` at `c9f9901`.
- `ruff check graphify tests`: clean. Added lines are within 100 code points.

### DIFF
--- a/graphify/cross_repo_calls.py
+++ b/graphify/cross_repo_calls.py
@@ -43,6 +43,9 @@ _LANG_SUFFIXES: dict[str, frozenset[str]] = {
     "csharp": frozenset({".cs"}),
     "java": frozenset({".java"}),
     "swift": frozenset({".swift"}),
+    # TS and JS share one key: a TS class legitimately answers a JS call site, so
+    # splitting them would drop every TS<->JS cross-repo call.
+    "typescript": frozenset({".ts", ".tsx", ".mts", ".cts", ".js", ".jsx", ".mjs", ".cjs"}),
 }
 
 # A declaration owns its members through a `method` edge, except in C++, where an
--- a/graphify/cross_repo_calls.py
+++ b/graphify/cross_repo_calls.py
@@ -44,8 +44,10 @@ _LANG_SUFFIXES: dict[str, frozenset[str]] = {
     "java": frozenset({".java"}),
     "swift": frozenset({".swift"}),
     # TS and JS share one key: a TS class legitimately answers a JS call site, so
-    # splitting them would drop every TS<->JS cross-repo call.
-    "typescript": frozenset({".ts", ".tsx", ".mts", ".cts", ".js", ".jsx", ".mjs", ".cjs"}),
+    # splitting them would drop every TS<->JS cross-repo call. An SFC script block is
+    # TS/JS too, so a class exported from `.vue` answers as well.
+    "typescript": frozenset({".ts", ".tsx", ".mts", ".cts", ".js", ".jsx", ".mjs", ".cjs",
+                             ".vue", ".svelte", ".astro"}),
 }
 
 # A declaration owns its members through a `method` edge, except in C++, where an
--- a/graphify/extract.py
+++ b/graphify/extract.py
@@ -3391,8 +3391,8 @@ def _resolve_python_member_calls(
             _emit_call(caller, children[0], rc)
 
 
-# Every suffix the TS/JS extractors claim. The member-call resolver both activates on
-# and parks by this set, so a pure-ESM (`.mjs`) repo is not silently left out.
+# Every suffix the TS/JS extractors claim, so a pure-ESM (`.mjs`) repo still activates
+# the member-call resolver below.
 _JS_TS_SUFFIXES = (".ts", ".tsx", ".mts", ".cts", ".js", ".jsx", ".mjs", ".cjs")
 
 
@@ -3437,11 +3437,15 @@ def _resolve_typescript_member_calls(
 
     contained = {e.get("target") for e in all_edges if e.get("relation") == "contains"}
 
+    # Only the JS/TS interop family declares a type this receiver can hold (SFC script
+    # blocks included). A same-named class in another language must neither answer nor
+    # hide from the parking branch that nothing local declares the name.
     type_def_nids: dict[str, list[str]] = {}
     node_by_id: dict[str, dict] = {}
     for n in all_nodes:
         node_by_id[n.get("id")] = n
-        if n.get("source_file") and n.get("id") in contained and _is_type_like_definition(n):
+        if (_lang_family(n.get("source_file")) == "jsts"
+                and n.get("id") in contained and _is_type_like_definition(n)):
             type_def_nids.setdefault(_key(n.get("label", "")), []).append(n["id"])
 
     method_index: dict[tuple[str, str], str] = {}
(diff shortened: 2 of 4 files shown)


xiongjianxu:
Self-audit pass over the bot review. One fix pushed; the review's finding does not reproduce.

**Fixed — the declaration index was corpus-wide.** `_is_type_like_definition` is language-blind, so a same-named class in another language (a PHP `class Greeter`, say) made `type_defs` non-empty for a TS receiver typed `Greeter`. Resolution then fell to the origin gate, which correctly refused it — but the parking branch had already been skipped, so the call was dropped instead of handed to the merge. The index is now gated with `_lang_family(...) == "jsts"`, which is wider than the suffix tuple on purpose: `.vue`/`.svelte`/`.astro` script blocks are extracted with the TS/JS grammars, so a class exported from an SFC is in the same module graph and must keep answering. `test_a_class_declared_in_an_sfc_script_block_still_answers` pins that half, `test_a_class_from_another_language_neither_answers_nor_blocks_parking` the other. `_LANG_SUFFIXES["typescript"]` in `cross_repo_calls` gains the same three suffixes so the merge side agrees.

**Does not reproduce — a parked call being re-emitted after the origin gate rejects it.** Parking sits under `if not type_defs:` and returns immediately, so a name that matched anything locally never reaches it; the origin gate runs strictly later, on the branch where a local definition *was* found. `test_a_type_the_origin_gate_rejected_is_not_parked` pins exactly that case and passes. The reasoning behind it is deliberate: if a local declaration exists and the gate judged the evidence too weak, parking would ask the merge to accept a remote match on weaker evidence still.

Verification: `ruff check graphify tests` clean; full suite 5321 passed / 93 skipped, with only the four pre-existing `tests/test_ollama_retry_cap.py` env failures that `v8` also shows here.

xiongjianxu:
The push was an author-identity rewrite (my commits carried an email not linked to my GitHub account); the tree is byte-identical to the previously reviewed head, so the re-run's findings are the ones already answered above.

On the one finding I had not answered yet — **parked TS calls skip a qualified type annotation** — the behaviour reproduces, and I am declining it on purpose:

```
== private greeter: Greeter        parked: [{'callee': 'greet', 'receiver_type': 'Greeter', 'lang': 'ts'}]
== private greeter: pkg.Greeter    parked: []
```

`_ts_receiver_type_table` states the rule it is following: "an array, union, generic, qualified, or predefined type is skipped (precision over recall, matching the receiver-typed resolvers for Swift/C#/C++)". Taking the last segment would be wrong for TS in a way it is not for Kotlin: a dotted Kotlin type is a package on one classpath, so `com.example.Greeter` and `Greeter` are the same declaration; a dotted TS type is overwhelmingly a namespace import or a namespace alias — `React.Component`, `pkg.Greeter` — and `node_modules` is unscanned, so the tail is an npm name that this PR would then bind to any same-named class in another scanned repo. That is a fabricated cross-repo edge, which is exactly what the single-definition guard exists to prevent.

The TS parking site follows the same reasoning for the capitalized-spelling path: it refuses to park when the receiver type came from an uppercase spelling rather than a real annotation, because an uppercase TS receiver also matches namespace aliases and default imports.

Issue #1698, opened 2026-07-06
Kotlin: cross-file calls to `object` (singleton) members produce no calls edges — same-file resolution works


jerryliurui:
### Environment

- graphify 0.9.7 (PyPI `graphifyy`), macOS, Python 3.10
- Extraction invoked directly via `graphify.extract.extract(files, parallel=False)` (same result through the CLI pipeline)

### Summary

Calls to members of a Kotlin `object` (singleton) resolve to `calls` edges **only when caller and callee are in the same file**. The identical pattern across two files produces **zero edges** — not even a `references` edge from the caller's class to the `object`.

This is the Kotlin sibling of the Swift navigation-expression work that landed in 0.9.5 (#1604) and looks adjacent to #1696 (Java receivers) and #1682 (PHP member calls).

### Reproducer (3 files)

`MessageHelper.kt`
```kotlin
package demo

object MessageHelper {
    fun isRoomClosed(roomId: String): Boolean {
        return roomId.isEmpty()
    }
}
```

`RoomPage.kt`
```kotlin
package demo

class RoomPage {
    fun refresh(roomId: String) {
        if (MessageHelper.isRoomClosed(roomId)) {
            return
        }
    }
}
```

`SameFile.kt` (control — same pattern, single file)
```kotlin
package demo

object LocalHelper {
    fun isReady(): Boolean = true
}

class LocalPage {
    fun open() {
        if (LocalHelper.isReady()) {
            return
        }
    }
}
```

### Current behavior (0.9.7)

Same-file control resolves correctly:

```
samefile_localpage_open --calls--> samefile_localhelper_isready  [EXTRACTED]
```

Cross-file case emits nothing. `roompage_roompage_refresh` has only:

```
roompage_roompage --method--> roompage_roompage_refresh          [EXTRACTED]
roompage_roompage_refresh --references--> roompage_kt_string     [EXTRACTED]
```

No `calls` edge to `messagehelper_messagehelper_isroomclosed`, and no `references` edge to `MessageHelper` at all.

### Expected

```
roompage_roompage_refresh --calls--> messagehelper_messagehelper_isroomclosed  [EXTRACTED or INFERRED]
```

The receiver here is a navigation-qualified `object` name — statically unambiguous, no type inference needed, so this should be the easiest cross-file case for Kotlin.

### Real-world impact

[...]

PR #3389, opened 2026-09-07
Kotlin: park object member calls for the cross-repo merge (#3388)

### LINKED ISSUE #3388: Kotlin: member calls on a typed receiver produce no edge across files, and nothing across repos
### What happens

A Kotlin member call on a typed receiver produces **no edge at all** when the receiver's
class is declared in another file, and nothing survives into `graph.json` when it is
declared in another repository.

Two files, one corpus:

```kotlin
// Greeter.kt
class Greeter {
    fun greet() {}
}

// App.kt
class App(private val greeter: Greeter) {
    fun run() { greeter.greet() }
}
```

`App.run` → `Greeter.greet` is missing. The shared cross-file pass skips member calls,
and Kotlin exports no receiver type table, so `greeter` is never typed: the call falls to
`raw_calls` and no resolver claims it. `affected` and every reverse-dependency query
under-report by exactly the calls that make a Kotlin call graph a call graph.

Split the same code into two repos and merge:

```
repo app/  App.kt      (as above)
repo lib/  Greeter.kt
```

`merge-graphs` adds nothing, because the single-repo build never recorded that `App.run`
wanted a `Greeter.greet` it could not find. This is the Kotlin instance of #3152, whose
merge pass can only finish calls a build parked.

### Expected

- `greeter.greet()` resolves to `Greeter.greet` in one corpus (INFERRED — the type came
  from a table, not from the call site), and `Registry.register()` on an `object` resolves
  EXTRACTED, matching the Swift/C#/Java tiering.
- When the receiver's type is declared nowhere in the build, the call is parked so
  `merge-graphs` / `global add` can bind it to the single declaration another repo owns.

### Notes

Every source of a receiver's type has to be read, because Kotlin's idiomatic injection
point (`class App(private val greeter: Greeter)`) declares no property at all — a
property-only table would miss the most common shape. The sources are: primary
constructor parameters, property declarations, function parameters, and local
`val x = T()` bindings whose type is only implied by the constructor call.

Ambiguity must stay unresolved rather than be guessed: two classes named `Greeter`, or a
receiver typed to a builtin that shares a name with a local class (`class Regex` vs
`kotlin.text.Regex`), produce no edge.

A paired PR follows.

xiongjianxu:
### What

Park a Kotlin `Receiver.method()` call for the cross-repo merge when the object/class the receiver names is declared in another repository, so `merge-graphs` / `global add` can finish the edge (#3388).

### Rebased onto the landed cross-file resolver

When this PR was first opened it also added the in-repo cross-file resolver. That capability has since landed independently on `v8` as `_resolve_kotlin_member_calls` (#1698, cross-file object/class-qualified calls). This PR is now rebased to sit **on top of** that resolver and contribute only the missing piece: the cross-repo **parking**.

The landed resolver binds `Receiver.method()` only when the object is declared in the same build; a call into another repo was dropped with the receiver name already in hand, so nothing reached `graph.json` and no merge-time pass could recover it.

### Change (20 lines)

- `extract.py` — at the resolver's "declared nowhere here" bail-out, park the call by name via `_park_unresolved_member_call`, the same mechanism Java/C++/C#/Swift already use. Only when the object is declared nowhere (an ambiguous 2+ match still fabricates nothing); never for a builtin receiver.
- `cross_repo_calls.py` — add `kotlin: {.kt, .kts}` to `_LANG_SUFFIXES` so a parked Kotlin call is answered only by a Kotlin declaration, not a same-named type in another language.
- `tests/test_cross_repo_member_calls.py` — a `kotlin-object-receiver` case in the per-language end-to-end parametrize (real extraction → park → merge → edge).

### Verified

`tests/test_cross_repo_member_calls.py` green (20 passed, kotlin case runs), and the landed #1698 in-repo resolver tests stay green.

### DIFF
--- a/graphify/cross_repo_calls.py
+++ b/graphify/cross_repo_calls.py
@@ -42,9 +42,9 @@ _LANG_SUFFIXES: dict[str, frozenset[str]] = {
     "cpp": frozenset({".cpp", ".cc", ".cxx", ".hpp", ".hh", ".hxx", ".h", ".cu", ".cuh"}),
     "csharp": frozenset({".cs"}),
     "java": frozenset({".java"}),
-    # A Kotlin call may be answered by a `.java` declaration: the JVM classpath is one
+    # A Kotlin call may be answered by any JVM-family declaration: the classpath is one
     # namespace, and a Kotlin module over a Java library is the common Android shape.
-    "kotlin": frozenset({".kt", ".kts", ".java"}),
+    "kotlin": frozenset({".kt", ".kts", ".java", ".scala", ".groovy", ".gradle"}),
     "swift": frozenset({".swift"}),
 }
 
--- a/graphify/cross_repo_calls.py
+++ b/graphify/cross_repo_calls.py
@@ -42,6 +42,9 @@ _LANG_SUFFIXES: dict[str, frozenset[str]] = {
     "cpp": frozenset({".cpp", ".cc", ".cxx", ".hpp", ".hh", ".hxx", ".h", ".cu", ".cuh"}),
     "csharp": frozenset({".cs"}),
     "java": frozenset({".java"}),
+    # A Kotlin call may be answered by a `.java` declaration: the JVM classpath is one
+    # namespace, and a Kotlin module over a Java library is the common Android shape.
+    "kotlin": frozenset({".kt", ".kts", ".java"}),
     "swift": frozenset({".swift"}),
 }
 
--- a/graphify/extract.py
+++ b/graphify/extract.py
@@ -148,7 +148,7 @@ from graphify.extractors.resolution import (  # noqa: E402,F401
 
 from graphify.symbol_resolution import resolve_bash_source_edges  # noqa: E402
 
-from graphify.extractors.engine import REFERENCE_CONTEXTS, _CSHARP_TYPE_PARAMETER_SCOPE_DECLARATIONS, _C_PRIMITIVE_TYPE_NODES, _JAVA_BUILTIN_TYPES, _JAVA_TYPE_PARAMETER_SCOPE_DECLARATIONS, _JS_FUNCTION_VALUE_TYPES, _JS_SCOPE_BOUNDARY, _PYTHON_ANNOTATION_NOISE, _PYTHON_TYPE_CONTAINERS, _RUBY_CLASS_FACTORIES, _c_collect_type_refs, _cpp_collect_type_refs, _cpp_declarator_name, _cpp_local_var_types, _csharp_attribute_names, _csharp_classify_base, _csharp_collect_type_refs, _csharp_extra_walk, _csharp_namespace_id, _csharp_namespace_name, _csharp_pre_scan_interfaces, _csharp_type_parameters_in_scope, _dynamic_import_js, _extract_generic, _find_body, _find_require_call, _get_cpp_func_name, _java_annotation_names, _java_collect_type_refs, _java_extra_walk, _java_type_parameters_in_scope, _js_collect_pattern_idents, _js_dispatch_value_idents, _js_extra_walk, _js_local_bound_names, _js_member_assignment_target, _js_module_bound_names, _kotlin_collect_type_refs, _kotlin_function_return_type_node, _kotlin_property_type_node, _kotlin_user_type_name, _php_collect_type_refs, _php_method_return_type_node, _php_name_text, _python_collect_assignment_targets, _python_collect_param_refs, _python_collect_type_refs, _python_local_bound_names, _python_module_bound_names, _python_param_names, _read_csharp_type_name, _require_imports_js, _ruby_const_last_name, _ruby_extra_walk, _ruby_local_class_bindings, _ruby_new_class_name, _scala_collect_type_refs, _semantic_reference_edge, _source_location, _swift_classify_base, _swift_collect_type_refs, _swift_constructor_type, _swift_declaration_keyword, _swift_extra_walk, _swift_local_var_types, _swift_pre_scan, _swift_property_name, _swift_property_type_node, _swift_receiver_name, _swift_user_type_name, _ts_decorator_name, _ts_descendant_decorators, _ts_emit_decorator_edges, _ts_extra_walk, _ts_method_name, _ts_receiver_type_table  # noqa: E402,F401
+from graphify.extractors.engine import REFERENCE_CONTEXTS, _CSHARP_TYPE_PARAMETER_SCOPE_DECLARATIONS, _C_PRIMITIVE_TYPE_NODES, _JAVA_BUILTIN_TYPES, _JAVA_TYPE_PARAMETER_SCOPE_DECLARATIONS, _JS_FUNCTION_VALUE_TYPES, _JS_SCOPE_BOUNDARY, _KOTLIN_BUILTIN_TYPES, _PYTHON_ANNOTATION_NOISE, _PYTHON_TYPE_CONTAINERS, _RUBY_CLASS_FACTORIES, _c_collect_type_refs, _cpp_collect_type_refs, _cpp_declarator_name, _cpp_local_var_types, _csharp_attribute_names, _csharp_classify_base, _csharp_collect_type_refs, _csharp_extra_walk, _csharp_namespace_id, _csharp_namespace_name, _csharp_pre_scan_interfaces, _csharp_type_parameters_in_scope, _dynamic_import_js, _extract_generic, _find_body, _find_require_call, _get_cpp_func_name, _java_annotation_names, _java_collect_type_refs, _java_extra_walk, _java_type_parameters_in_scope, _js_collect_pattern_idents, _js_dispatch_value_idents, _js_extra_walk, _js_local_bound_names, _js_member_assignment_target, _js_module_bound_names, _kotlin_collect_type_refs, _kotlin_function_return_type_node, _kotlin_property_type_node, _kotlin_user_type_name, _php_collect_type_refs, _php_method_return_type_node, _php_name_text, _python_collect_assignment_targets, _python_collect_param_refs, _python_collect_type_refs, _python_local_bound_names, _python_module_bound_names, _python_param_names, _read_csharp_type_name, _require_imports_js, _ruby_const_last_name, _ruby_extra_walk, _ruby_local_class_bindings, _ruby_new_class_name, _scala_collect_type_refs, _semantic_reference_edge, _source_location, _swift_classify_base, _swift_collect_type_refs, _swift_constructor_type, _swift_declaration_keyword, _swift_extra_walk, _swift_local_var_types, _swift_pre_scan, _swift_property_name, _swift_property_type_node, _swift_receiver_name, _swift_user_type_name, _ts_decorator_name, _ts_descendant_decorators, _ts_emit_decorator_edges, _ts_extra_walk, _ts_method_name, _ts_receiver_type_table  # noqa: E402,F401
 
 from graphify.extractors.pascal import _PAS_BEGIN_END_TOKEN_RE, _PAS_CALL_RE, _PAS_END_SEMI_RE, _PAS_IMPL_HEADER_RE, _PAS_KEYWORDS, _PAS_METHOD_DECL_RE, _PAS_MODULE_RE, _PAS_TOKEN_RE, _PAS_TYPE_HEADER_RE, _PAS_USES_RE, _extract_pascal_regex, _pascal_find_body, _pascal_split_bases, _pascal_split_sections, _pascal_split_uses, _pascal_strip_comments, extract_pascal  # noqa: E402,F401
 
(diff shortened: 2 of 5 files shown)


xiongjianxu:
Self-audit pass over the bot review. Three fixes pushed, one finding I am not acting on.

**1. The declaration index was corpus-wide.** A same-named class in an unrelated language could answer a Kotlin receiver, and its presence made `type_defs` non-empty so the parking branch never ran — the call was dropped instead of handed to the merge. Now gated with `_lang_family(...) == "jvm"`, the interop-family map the shared cross-file resolver already uses. `.java` stays in on purpose: one classpath means a Kotlin module over a Java library is a real answer, not a collision — the dominant shape in a part-migrated Android codebase. Two tests pin the pair: `test_a_java_class_answers_a_kotlin_receiver` and `test_a_class_from_an_unrelated_language_never_answers_a_kotlin_receiver`. `_LANG_SUFFIXES["kotlin"]` in `cross_repo_calls` now carries the same family, so a call resolves the same way locally and at merge time.

**2. The INFERRED score was off the rubric.** It emitted `0.8`, which `references/extraction-spec.md` rules out; `tests/test_inferred_confidence_rubric.py` only greps the bare literal, so the conditional expression slipped past it. Now `0.85`, the high-confidence rung, matching the sibling resolvers.

**3. `_kotlin_constructor_type` read like a bug.** The `continue`-then-unconditional-`return None` was correct — only the first `call_expression` child counts — but nothing said so. Rewritten as a `next(...)` pick over the children; same behaviour, no control flow to decode.

**Not acting on — `method_index` keeps the last member of a name per type.** `_resolve_csharp_member_calls` does exactly the same, and for Kotlin the collision is an overload, where the graph has no basis to prefer one signature. Refusing to emit anything would lose a genuine call edge in the common case; picking arbitrarily among true overloads keeps it. (Go's resolver in the sibling PR requires `len(targets) == 1` instead, because Go has no overloads and a collision there means two different types were folded.)

Verification: `ruff check graphify tests` clean; full suite 5324 passed / 93 skipped, with only the four pre-existing `tests/test_ollama_retry_cap.py` env failures that `v8` also shows here.

xiongjianxu:
Thanks — the ordering finding was real and is now fixed in `2489847`.

**`_kotlin_local_var_types` pop order (engine.py:883) — fixed.** The helper pushed children
onto a LIFO stack in document order, so siblings popped *reversed* and the **last** same-named
local won, the opposite of the "first binding wins" the docstring promises. Reproduced with two
`svc` bindings in sibling branches of one body, the candidate classes in separate files so only
the receiver pass can answer:

```kotlin
fun run(flag: Boolean) {
    if (flag) {
        val svc: Alpha = Alpha()
        svc.doThing()          // before: bound to Beta.doThing
    } else {
        val svc: Beta = Beta()
    }
}
```

Fix is `stack.extend(reversed(n.children))`, pinned by
`test_the_first_binding_of_a_name_wins_over_a_later_sibling_branch`, which fails on the previous
commit and passes now. The pre-existing `_cpp_local_var_types` / `_swift_local_var_types` carry
the same latent pattern and the same docstring claim; they are out of this PR's scope and left
untouched.

The other three findings are unchanged from the previous round and remain deliberate:

- **arbitrary overload pick (extract.py:4467/4473)** and **relation-agnostic `existing_pairs`
  (4505)** — both follow `_resolve_csharp_member_calls`, the pass this one is modelled on. A
  single-target requirement per `(type, name)` is the god-node guard; picking among overloads
  would need parameter types the extractor does not record, so bailing is the honest behaviour.
- Full suite green: 5325 passed, 93 skipped; ruff clean.

xiongjianxu:
Rebased onto `v8` (0.9.58) — the conflict was upstream's `is_py_local_data` guard (#3405 Part 3/4) wrapping the same `rc_entry` block; the Kotlin tag now sits inside it, keyed on `ts_module` rather than on `kotlin_qualified_prefix`, since receiver-typed resolution needs `lang == "kotlin"` even when no dotted prefix is present.

Two of the five findings were real. Both are fixed with a test that fails on the parent commit.

**1. Plain Kotlin constructor parameters were treated as class properties** — `engine.py:865`, fixed in `9f04f81`. A `class_parameter` without `val`/`var` declares no member; it is in scope only in initializers and `init` blocks. Because the class node is visited before its body, that parameter claimed the name first and the real property never got it:

```kotlin
class App(raw: Raw) {
    private val raw = Wrapper(raw)
    fun run() { raw.doThing() }   // before: bound to Raw.doThing
}
```

Dropping plain parameters outright would have been wrong in the other direction — `class App(greeter: Greeter) { val name = greeter.greet() }` is a call the initializer walk does resolve — so the name is recorded as weak and only a body property of the same name may take it back. Both halves are pinned: `test_a_property_outranks_a_plain_constructor_parameter_of_the_same_name` and `test_a_plain_constructor_parameter_still_types_an_initializer_receiver`.

**2. Capitalized receivers bypassed the type table** — `extract.py:4662`, fixed in `e634f57`. The capitalized spelling means "the type itself" only because nothing of that name is bound; checking the table first is what makes that inference sound:

```kotlin
val Greeter = Other()
Greeter.greet()   // before: bound to class Greeter
```

`test_a_binding_in_scope_outranks_the_class_of_the_same_name` fails on `9f04f81` and passes now. `Registry.register()` on an `object` is unaffected — an `object` declaration puts nothing in the receiver table.

The other three are unchanged from the previous rounds and remain deliberate:

- **duplicate member names / arbitrary overload pick (`extract.py:4650`, `4661`)** — one finding in two places. `method_index` keeps the last member of a name per type, exactly as `_resolve_csharp_member_calls` does. For Kotlin a collision is an overload, and choosing between overloads needs parameter types the extractor does not record; emitting nothing would lose a real call edge in the common case, so the god-node guard stays at the type level (`len(type_defs) != 1`), not the member level.
- **relation-agnostic `existing_pairs` (`extract.py:4663`)** — verbatim what the C# resolver treats as "already edged". Diverging here would make the two passes disagree about that for no gain inside this PR.

Verification: `ruff check graphify tests` clean; full suite 5453 passed / 95 skipped, deselecting only the pre-existing `test_ts_normalizer_scales_linearly_on_large_files` timing flake (it fails on unmodified `v8` in a full run too) and `tests/test_ollama_retry_cap.py` (no `openai` in this env).

xiongjianxu:
Third finding was real too — fixed in `a1ea487`.

**Kotlin local receiver types ignored scope** — `engine.py:6291`. The table was flat per file, so a name bound in two bodies was one name and the second function inherited the first one's type:

```kotlin
class App {
    fun a() { val svc = Alpha(); svc.doThing() }   // Alpha.doThing
    fun b() { val svc = Beta();  svc.doThing() }   // also Alpha.doThing — wrong
}
```

`svc` / `client` / `repo` appearing in two methods of one class is the common case, not a corner, so this was producing wrong edges rather than merely conservative ones.

Parameters and locals now live in a per-caller table and travel on the raw call as `receiver_type` — the shape `_resolve_java_member_calls` and `_resolve_csharp_member_calls` already read, so the resolver gained no new mechanism. Class properties and primary-constructor parameters stay in the file-wide `kotlin_type_table`, because they genuinely are file-wide. Lookup order is caller scope → class bindings → the capitalized spelling.

Two tests: `test_a_local_in_one_function_does_not_type_a_receiver_in_another` pins the case above, and `test_the_first_binding_of_a_name_wins` is replaced by `test_a_parameter_types_its_own_function_without_retyping_the_property`, which now asserts both halves — the property answers in `run()`, the same-named parameter answers in `other()`. The old test only pinned the property half because a flat table could not do better.

Declaration order within one body is unchanged and already covered: a local must be declared before use in Kotlin, and the sibling-branch case is pinned by `test_the_first_binding_of_a_name_wins_over_a_later_sibling_branch`.

The remaining two findings (`extract.py:4669` arbitrary overload pick, `4673` relation-agnostic `existing_pairs`) are unchanged from the previous rounds and stay deliberate for the reasons given there.

Verification: `ruff check graphify tests` clean; full suite 5454 passed / 95 skipped, deselecting only the pre-existing `test_ts_normalizer_scales_linearly_on_large_files` timing flake and `tests/test_ollama_retry_cap.py` (no `openai` in this env).

xiongjianxu:
Thanks — one real defect out of the five, fixed in 2ffe2bf.

**Qualified receiver types (finding 3) — real, and worse than reported.** `_kotlin_collect_type_refs` reports a dotted `user_type`'s **first** identifier, so `private val greeter: com.example.Greeter` and `fun run(greeter: com.example.Greeter)` both typed the receiver `com`. Probed on the parent commit:

```
== qualified annotation
    parked on .run() [{'callee': 'greet', 'receiver_type': 'com', 'lang': 'kotlin', 'line': 'L5'}]
== qualified param
    parked on .run() [{'callee': 'greet', 'receiver_type': 'com', 'lang': 'kotlin', 'line': 'L4'}]
```

No edge, and a package segment parked as if it were a type — so the miss travels into the merged graph as a fact and `link_cross_repo_member_calls` can never finish the call. The fix takes the **last** segment on the receiver-table paths only (`_kotlin_head_type_name`, plus the function-parameter site that picked the ref inline). The collector itself is untouched: it also drives the `references` edges, whose `type` / `parameter_type` / `generic_arg` contexts want every segment. The builtin filter rides along on the tail, so `fun run(r: kotlin.text.Regex)` parks nothing instead of parking `kotlin`.

Three tests, all three failing on `a1ea487`:
- `test_a_qualified_property_annotation_names_the_type_in_its_last_segment`
- `test_a_qualified_parameter_type_names_the_type_in_its_last_segment`
- `test_a_qualified_builtin_type_parks_no_package_segment`

**Two-segment navigation regressing in-file `Foo.bar()` (finding 4) — no.** Kotlin never defers a member call to the cross-file pass (`_kotlin_keeps_in_file`), so the in-file cases resolve in the same build and are asserted:
- `test_an_object_receiver_is_the_type_itself` — in-file `object Registry` + `Registry.register()` → `EXTRACTED`
- `test_a_qualified_call_still_reaches_the_package_resolver` — a ≥3-segment chain is an FQN, not a receiver, and stays with #2550's pass → `EXTRACTED`

Probed both shapes directly (in-file `object` navigation, in-file class + property receiver): both emit `EXTRACTED` edges.

**Class-bound names never emitting `receiver_type` (finding 5) — no.** The resolver computes `type_name` from `rc.get("receiver_type")` **or** the file-level table and parks *that*, so a class-level binding parks its type just like a parameter's. `test_a_class_from_an_unrelated_language_never_answers_a_kotlin_receiver` asserts exactly this: the receiver is typed by a primary-constructor parameter (class-level, not in `receiver_type`) and the parked entry is `{"callee": "greet", "receiver_type": "Greeter", "lang": "kotlin", "line": "L2"}`.

**Arbitrary overload pick (findings 1–2)** — same finding as previous rounds, declined for the same reason: the `len(type_defs) != 1` / single-target guard is deliberate parity with the Java and C# resolvers already on `v8`. Picking one of several same-named candidates is the failure mode those guards exist to prevent; changing the policy is a cross-language decision, not a Kotlin patch.

Full suite on `2ffe2bf`: **5457 passed, 95 skipped** (`test_ollama_retry_cap` ignored — no local `openai`; `test_ts_normalizer_scales_linearly_on_large_files` deselected, a pre-existing timing flake that fails on unmodified `v8` in a full run and passes in isolation). `ruff check graphify tests` clean.

xiongjianxu:
One of the three findings on `2ffe2bf` is real and is fixed in `b241de4`; the other two do not hold.

**Qualified constructor calls (the `extract.py:4677` finding, in the shape the previous fix missed) — real, fixed.** My last commit taught the receiver table to read a dotted *annotation* to its last segment, but the initializer path was still stuck on a bare identifier: a qualified constructor call is a `navigation_expression`, not an `identifier`, so `_kotlin_constructor_type` returned `None`. Probed on `2ffe2bf`:

```
== qualified annotation            .run() -> .greet() @Greeter.kt INFERRED
== qualified nullable annotation   .run() -> .greet() @Greeter.kt INFERRED
== qualified parameter             .run() -> .greet() @Greeter.kt INFERRED
== val g = com.example.Greeter()   (nothing)
```

`b241de4` runs the existing capitalized-head test on the last segment. Two tests, the first failing on `3694b1e`:
- `test_a_qualified_constructor_call_types_an_unannotated_local`
- `test_a_qualified_method_call_does_not_type_a_local` — `val g = repo.load()` still binds nothing, which is what the capitalized-tail requirement is for

**Lowercase `object` singletons — declining.** Confirmed the behaviour, it is a miss and not a wrong edge:

```
== object registry + registry.register()   (nothing)
== object Registry + Registry.register()   .run() -> .register() @Registry.kt EXTRACTED
```

The `receiver[:1].isupper()` test is not a Kotlin heuristic I added — it is the shared convention of eight receiver resolvers already on `v8` (`extract.py` lines 3386, 3553, 3670, 3819, 4052, 4227, 4424 and this one). Removing it for Kotlin means any *unbound* lowercase receiver whose name happens to match a declaration binds to it — an inherited or delegated property named `registry` would bind to `object registry` on the strength of the name alone. That trade is a cross-language policy decision, not a Kotlin patch, and `object registry` is against Kotlin's own naming convention. A miss here costs one edge; the alternative fabricates them.

**"Collected Kotlin function parameter receiver types are never read" — does not hold.** `kotlin_receiver_types` is read at the call-expression site (`kotlin_receiver_types.get(caller_nid, {}).get(member_receiver or "")`) and stamped onto the raw call as `receiver_type`, which is the first thing `_resolve_kotlin_member_calls` consults. Two tests fail if that read is removed, because parameters are deliberately no longer written to the file-level table:
- `test_a_function_parameter_types_the_receiver`
- `test_a_parameter_types_its_own_function_without_retyping_the_property` — asserts `{".run()": {"Greeter.kt"}, ".other()": {"Other.kt"}}`, i.e. the parameter answers only inside its own function

**On the `3694b1e` review** — that push was an author-identity rewrite (my commits were authored under an email not linked to my GitHub account). The tree hash is identical to `2ffe2bf`, verified before pushing, so its three findings are the overload-pick and `existing_pairs` pair declined in the previous rounds on the same C#/Java-parity grounds.

Full suite on `b241de4`: **5459 passed, 95 skipped** (`test_ollama_retry_cap` ignored, no local `openai`; the TS normalizer timing test deselected — it fails on unmodified `v8` in a full run and passes in isolation). `ruff check graphify tests` clean.

PR #3391, opened 2026-09-07
PHP: receiver-typed member calls, and park them for cross-repo merges

### LINKED ISSUE #3390: PHP: member calls on a typed receiver produce no edge across files, and nothing across repos
### What happens

A PHP member call on a typed receiver produces **no edge at all** when the receiver's
class is declared in another file, and nothing survives into `graph.json` when it is
declared in another repository.

Two files, one corpus:

```php
<?php // Greeter.php
class Greeter {
    public function greet(): void {}
}
```

```php
<?php // App.php
class App {
    private Greeter $greeter;
    public function run(): void { $this->greeter->greet(); }
}
```

`App.run` → `Greeter.greet` is missing. Two things have to be true for that:
`member_call_expression`'s `object` field is never read, so the receiver is not even
recorded on the raw_call; and PHP exports no type table, so there would be nothing to
type it against. The call falls to `raw_calls`, the shared cross-file pass skips member
calls, and no resolver claims it. Every call through an injected dependency — the
dominant shape in any PHP framework — is invisible to `affected` and to every
reverse-dependency query.

Split the same code into two repos and merge:

```
repo app/  App.php      (as above)
repo lib/  Greeter.php
```

`merge-graphs` adds nothing, because the single-repo build never recorded that `App.run`
wanted a `Greeter.greet` it could not find. This is the PHP instance of #3152, whose merge
pass can only finish calls a build parked.

### Expected

- `$this->greeter->greet()` and `$greeter->greet()` resolve to `Greeter.greet` in one
  corpus (INFERRED — the type comes from a declaration, not from the call site).
- When the receiver's type is declared nowhere in the build, the call is parked so
  `merge-graphs` / `global add` can bind it to the single declaration another repo owns.

### Notes

All four places a PHP receiver's type is written down have to be read, or the common
shapes are missed: a typed property (`private Greeter $g;`), a promoted constructor
parameter (`__construct(private Greeter $g)`, which declares no property at all), a typed
parameter, and `$g = new Greeter()`.

Ambiguity must stay unresolved rather than be guessed: a union or intersection type
(`Greeter|Other`) names no one class, a chain longer than `$this->prop->call()` types no
receiver, and two classes named `Greeter` produce no edge.

`Helper::format()` is a `scoped_call_expression`, not a member call, and keeps its
existing path.

A paired PR follows.

xiongjianxu:
Fixes #3390. Two cherry-pickable commits: the single-repo resolver, then the park that
lets a merged graph finish what one build cannot.

## Commit 1 — resolve PHP member calls through the receiver's declared type

PHP recorded no receiver on a member call and exported no type table, so
`$greeter->greet()` across files produced no edge. Both halves are added:

- the extractor captures the receiver — the variable for `$greeter->greet()`, the
  property name for `$this->greeter->greet()`;
- a per-file `php_type_table` is exported, and `_resolve_php_member_calls` types the
  receiver from it and emits `calls` to the single class declaring that type.

Table sources, all four needed:

| source | why it cannot be dropped |
| --- | --- |
| typed property | `private Greeter $greeter;` |
| promoted constructor parameter | `__construct(private Greeter $g)` declares no property, so nothing else names the type |
| typed parameter | `function run(Greeter $g)` |
| `new` binding | `$g = new Greeter();` |

Design calls:

1. **Always INFERRED, no EXTRACTED tier.** The sibling resolvers give a
   type-qualified receiver (`Type.staticMethod()`) 1.0/EXTRACTED, but in PHP that
   spelling is `Helper::format()` — a `scoped_call_expression`, which already has its own
   path and is left alone. Every receiver this pass types is a variable, so there is no
   spelling of a member call that names the type exactly.
2. **PHP does not defer.** A call whose bare callee name matches in its own file still
   resolves there, exactly as before; the receiver type is consulted only after that miss,
   which is the only situation where it can add an edge. This keeps in-file behaviour
   byte-identical rather than trading known edges for better-typed ones.
3. **One class name or nothing.** A union or intersection type names no one class, and
   binding the first arm of `Greeter|Other` would be a guess. `?Greeter` does name one.
4. **Receiver depth 1.** `$this->prop->call()` and `$var->call()` are typed;
   `$this->a->b->call()` types neither `a` nor `b` and is left unresolved.
5. **First binding in the file wins**, so a parameter named like a property in another
   method cannot retype the property's own calls. The table walk pushes children reversed
   to yield document order — without that, "first" is DFS order and the wrong binding wins
   (this is what the shadowing test caught).
6. **`confidence_score` is 0.85**, the high-confidence rung of the rubric's discrete
   INFERRED scale in `references/extraction-spec.md`. The sibling member-call resolvers
   still emit the off-rubric 0.8; `test_no_module_hardcodes_an_off_rubric_inferred_score`
   exists to stop new ones, and snapping the pre-existing sites is out of scope here.

## Commit 2 — park what this build cannot answer (#3152)

A receiver typed to a class with **zero** declarations in this corpus is parked on the
caller as a `metadata.unresolved_calls` entry (names only, never node ids — those are
rewritten by the #1529 remap and again by repo prefixing). `> 1` declarations stays
dropped: local ambiguity is not something merging can narrow. The merge pass then binds
the entry when exactly one declaration in another repo answers it.

`_LANG_SUFFIXES["php"]` covers every extension the PHP extractors claim (`.php`, `.phtml`,
`.php3`–`.php7`, `.phps`), so a class declared in a template file still answers.

## Known cost

Class name matching is case-sensitive on both sides. PHP class names are case-insensitive,
so `private greeter $g` binds nothing. Folding case in the merge pass would let the
languages that park alongside PHP bind `greeter` to `Greeter`, which is why the pass keeps
case; a PHP-only fold is a separate change.

## Verification

- `tests/test_php_receiver_member_calls.py` — 11 cases: one per type source (including
  `?Greeter`), the negatives that must stay unresolved (untyped parameter, union type,
  longer chain, two same-named classes), the shadowing case, and one pinning that
  `Helper::format()` keeps its existing path. 6 of the 11 fail on `v8`.
- `tests/test_cross_repo_member_calls.py` — a `php-typed-property` arm in the per-language
  park→merge test.
- Full suite: 5321 passed, 93 skipped, no new failures. `ruff check graphify tests` clean.

### DIFF
--- a/graphify/cross_repo_calls.py
+++ b/graphify/cross_repo_calls.py
@@ -42,6 +42,7 @@ _LANG_SUFFIXES: dict[str, frozenset[str]] = {
     "cpp": frozenset({".cpp", ".cc", ".cxx", ".hpp", ".hh", ".hxx", ".h", ".cu", ".cuh"}),
     "csharp": frozenset({".cs"}),
     "java": frozenset({".java"}),
+    "php": frozenset({".php", ".phtml", ".php3", ".php4", ".php5", ".php7", ".phps"}),
     "swift": frozenset({".swift"}),
 }
 
--- a/graphify/extract.py
+++ b/graphify/extract.py
@@ -4615,6 +4615,9 @@ def _resolve_php_member_calls(
     there, takes the single class declaring that type, and emits the ``calls`` edge to
     its method. Always INFERRED: the type comes from the table, never from the call site
     (``Helper::format()`` is a scoped call and keeps its own path).
+
+    A receiver typed to a class this corpus declares nowhere is parked on the caller for
+    a merged graph to finish (#3152).
     """
     raw = [
         rc
@@ -4660,7 +4663,13 @@ def _resolve_php_member_calls(
         if not type_name or type_name in _LANGUAGE_BUILTIN_GLOBALS or type_name in php_builtins:
             continue
         type_defs = type_def_nids.get(_key(type_name), [])
-        if len(type_defs) != 1:  # ambiguous or absent -> bail (god-node guard)
+        if not type_defs:
+            # Declared nowhere here — usually "in a repo this build does not contain",
+            # so park it for the merge (#3152). The extractor's `lang` tag already says
+            # who is asking, so no suffix sniff is needed.
+            _park_unresolved_member_call(node_by_id.get(caller), callee, type_name, "php", rc)
+            continue
+        if len(type_defs) != 1:  # ambiguous -> bail (god-node guard)
             continue
         target = method_index.get((type_defs[0], _key(callee)))
         if not target or target == caller or (caller, target) in existing_pairs:
--- a/graphify/extract.py
+++ b/graphify/extract.py
@@ -4601,6 +4601,86 @@ def _resolve_csharp_qualified_calls(
         })
 
 
+def _resolve_php_member_calls(
+    per_file: list[dict],
+    all_nodes: list[dict],
+    all_edges: list[dict],
+) -> None:
+    """Resolve PHP member calls (``$greeter->greet()``) through the receiver's type.
+
+    The shared cross-file pass skips member calls, so a call on a typed receiver whose
+    method is declared in another file resolved to nothing. The per-file
+    ``php_type_table`` names the declared type of every property, promoted constructor
+    parameter, typed parameter and ``new`` binding; this pass looks the receiver up
+    there, takes the single class declaring that type, and emits the ``calls`` edge to
+    its method. Always INFERRED: the type comes from the table, never from the call site
+    (``Helper::format()`` is a scoped call and keeps its own path).
+    """
+    raw = [
+        rc
+        for result in per_file
+        for rc in result.get("raw_calls", [])
+        if rc.get("lang") == "php" and rc.get("is_member_call")
+        and rc.get("receiver") and rc.get("callee") and rc.get("caller_nid")
+    ]
+    if not raw:
+        return
+    type_table_by_file: dict[str, dict[str, str]] = {}
+    for result in per_file:
+        tt = result.get("php_type_table")
+        if tt and tt.get("path"):
+            type_table_by_file[tt["path"]] = tt.get("table", {})
+
+    def _key(label: str) -> str:
+        return re.sub(r"[^a-zA-Z0-9]+", "", str(label)).lower()
+
+    # A genuine declaration is the target of a `contains` edge from its file node; a bare
+    # type reference mints a same-label stub that would otherwise make a real name ambiguous.
+    contained = {e.get("target") for e in all_edges if e.get("relation") == "contains"}
+    type_def_nids: dict[str, list[str]] = {}
+    node_by_id: dict[str, dict] = {}
+    for n in all_nodes:
+        node_by_id[n.get("id")] = n
+        if n.get("source_file") and n.get("id") in contained and _is_type_like_definition(n):
+            type_def_nids.setdefault(_key(n.get("label", "")), []).append(n["id"])
+
+    method_index: dict[tuple[str, str], str] = {}
+    for e in all_edges:
+        if e.get("relation") != "method":
+            continue
+        tnode = node_by_id.get(e.get("target"))
+        if tnode is not None:
+            method_index[(e.get("source"), _key(tnode.get("label", "")))] = e["target"]
+
+    php_builtins = _LANGUAGE_BUILTIN_BASE_CLASSES.get("php", frozenset())
+    existing_pairs = {(e.get("source"), e.get("target")) for e in all_edges}
+    for rc in raw:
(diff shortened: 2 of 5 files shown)


xiongjianxu:
Self-audit pass over the bot review. One fix pushed, two findings I am deliberately not acting on.

**Fixed — the declaration index was corpus-wide.** `_is_type_like_definition` is label-based and language-blind, so a Java `class Greeter` could answer a PHP receiver typed `Greeter`; and because its presence made `type_defs` non-empty, the parking branch never ran and the call was dropped instead of handed to the merge. The index is now gated with `_lang_family(...) == "php"` — the same interop-family map the shared cross-file call resolver consults — which also keeps every `.phtml`/`.php5` declaration in. Pinned by `test_a_class_from_another_language_never_answers_a_php_receiver`, asserting both halves: no edge, and a parked entry.

**Not acting on — `existing_pairs` is relation-agnostic.** It is built from every edge, so a pre-existing `references` or `uses` edge between the same pair suppresses the `calls` edge. That is verbatim what `_resolve_csharp_member_calls` does, and diverging here would make the two resolvers disagree about what "already edged" means for no gain in this PR.

**Not acting on — the type table is per-file and flat.** Swift (#1356), TS/JS and C++ all scope theirs this way, and `test_the_first_binding_of_a_name_wins` pins the consequence deliberately: a parameter shadowing a property elsewhere in the file must not retype the property's own calls. C#/Java moved to per-method tables in #2299 after flat tables caused real mis-binds there; if PHP wants that later it is the same refactor, not a change of behaviour under this feature.

Verification: `ruff check graphify tests` clean; full suite 5322 passed / 93 skipped, with only the four pre-existing `tests/test_ollama_retry_cap.py` env failures that `v8` also shows here.

xiongjianxu:
Both advisory findings were checked against the branch; neither holds.

**`member_receiver` used before assignment (engine.py:5551) — no defect.** It is initialised to
`None` at engine.py:5318, at the top of the call-expression block, before any language branch
runs. The PHP chained-access arm only ever assigns to it.

**Nullable builtin types parked as unresolved (extract.py:4476) — the premise does not hold.**
Nullability is unwrapped before the lookup: `_php_declared_type_name` strips `optional_type`, so
`?DateTime` and `DateTime` reach the resolver as the same name and behave identically. `?string`
yields nothing at all, because the unwrapped node is a `primitive_type` and only `named_type`
qualifies.

What *is* true, and deliberate, is that a stdlib class such as `DateTime` gets parked: the PHP
blocklist consulted there is the exception-base-class set, so `DateTime` is not in it. A
single-repo build genuinely cannot distinguish a stdlib class from a class declared in a sibling
repo, and parking is the conservative side of that — the payload is names only, capped per node,
and only becomes an edge if a declaration of that name actually turns up in the merged corpus.
Adding a PHP stdlib class list would trade that for a new failure mode, since a project is free
to declare its own namespaced `DateTime`.

xiongjianxu:
Rebased onto `v8` (0.9.58) — same conflict as the Kotlin sibling, upstream's `is_py_local_data` guard (#3405 Part 3/4) wrapping the `rc_entry` block; the PHP tag now sits inside it. `MERGEABLE` again, CI green.

**On the `high` finding — PHP parked calls are not accepted by the cross-repo linker (`cross_repo_calls.py:170`): this is refuted by execution, not by argument.** The PR's one-line change to that module registers the family the gate reads:

```
graphify/cross_repo_calls.py:45   "php": frozenset({".php", ".phtml", ".php3", ".php4", ".php5", ".php7", ".phps"}),
```

and the end-to-end case builds two repos, merges them through the CLI, and asserts the edge exists:

```
tests/test_cross_repo_member_calls.py::test_each_language_parks_the_call_and_the_merge_finishes_it[php-typed-property] PASSED
```

That parametrization asserts both halves — the parked entry (`callee`, `receiver_type`, `lang == "php"`, call site) in the app repo's `graph.json`, then exactly one `_cross_repo_call` edge after `merge-graphs`, from `app::…run` to `lib::…greet`. `_member_relations` returning `("method",)` for PHP is correct and is what the test exercises: a PHP method is a `method` edge, and the `defines` fallback exists only for C++'s in-class declarations.

The remaining three are unchanged from the previous rounds:

- **`existing_pairs` is relation-agnostic (`extract.py:4676`, `4678`)** — one finding reported twice. It is verbatim what `_resolve_csharp_member_calls` treats as "already edged"; diverging here would make the two resolvers disagree about that for no gain in this PR.
- **`member_receiver` may be unbound (`engine.py:5726`)** — no defect. It is initialised to `None` at the top of the call-expression block, before any language branch runs; the PHP chained-access arm only ever assigns to it.

Verification: `ruff check graphify tests` clean; full suite 5447 passed / 95 skipped, deselecting only the pre-existing `test_ts_normalizer_scales_linearly_on_large_files` timing flake (it fails on unmodified `v8` in a full run too) and `tests/test_ollama_retry_cap.py` (no `openai` in this env).

xiongjianxu:

PR #3392, opened 2026-09-07
Let builtin named member calls through to cross file resolution

### LINKED ISSUE #3381: `_LANGUAGE_BUILTIN_GLOBALS` deletes member calls named after another language's builtins (all grammars)
### Summary

`_LANGUAGE_BUILTIN_GLOBALS` (`extractors/base.py`) is one union across every language. The gate that
consults it decides **both** whether a call gets a same-file bare-name edge **and** whether it is
recorded in `raw_calls` — the only route to the receiver-typed resolvers. The union is right for a
*bare* call (`String(x)` really would become a god node), but applied to a *member* call it deletes
the call outright in whatever language does not own the name, with no warning.

`open` and `set` are Python builtins; `Set`, `Map`, `URL`, `String` are JavaScript ones. So
`session.open()` in Swift, `xhr.open()` in JavaScript and `_server.Set()` in C# are all silently
discarded.

### Reproduction

```swift
// probe.swift
enum T { static func run() { p.open() } }        // 0 calls extracted
enum T { static func run() { p.doWorkXyz() } }   // 1 call  extracted
```

Measured on graphify 0.9.54, every grammar installed locally. Each probe is compared against the
same source with the callee renamed to a name no builtin table claims:

| grammar | probe | probe calls | control calls |
|---|---|---|---|
| python, go, rust, java, kotlin, ruby, php, scala, cpp, swift | `p.open()` | 0 | 1 |
| javascript, typescript | `xhr.open()` | 1 | 2 |
| c_sharp | `_server.Set()` | 0 | 1 |

**13 of 13 drop the call.** On one real 338-file Swift codebase this removed **236 member calls**,
including `AppleReadyStreamFinder.find -> RemuxPlayback.open`, which left two tightly coupled
subsystems in separate Louvain communities with no path between them except through `Logger`.

### Suggested fix

A member call carries a receiver, so it is not the ambiguous case the union guards: every
receiver-typed resolver only emits an edge when the receiver's type resolves to exactly one
definition, and the shared cross-file pass drops member calls outright (#543/#1219). Letting a
builtin-named member call through to `raw_calls` — while forcing `tgt_nid = None` so it never takes
the unguarded bare-name path — can therefore only add guarded edges.

```python
_builtin_member_call = (
    callee_name is not None
    and is_member_call
    and callee_name in _LANGUAGE_BUILTIN_GLOBALS
)
if callee_name and (
    _builtin_member_call or callee_name not in _LANGUAGE_BUILTIN_GLOBALS
):
    ...
    # and in the defer condition:
    if _python_defer or _java_defer or _builtin_member_call or (...):
        tgt_nid = None
```

The same gate is duplicated in `extractors/go.py` (~L480) and `extractors/rust.py` (~L372) and needs
the same change; in Go, `tgt_nid` also has to fold `_builtin_member_call` into the existing
`import_path` guard.

### Verification

With the change applied, all 13 grammars keep the call, and graphify's **own 86-file Python source
tree extracts byte-identically** — 2441 nodes / 5624 edges / 2033 calls, with the same top call-target
in-degrees (91 / 56 / 45 / 35 / 27 / 21). No god-node inflation.

Happy to open a PR if the approach looks right.

ayushcodes10:
## What

`_LANGUAGE_BUILTIN_GLOBALS` is one union across every language, right for a bare call (`String(x)` really would become a god node) but wrong applied to a member call: `open` is a Python builtin and `Set` a JavaScript one, so `session.open()` in Swift or `_server.Set()` in C# was silently discarded outright — no same-file edge, but also no `raw_calls` entry, so cross-file resolution never got a chance to try it.

A member call carries a receiver, so it isn't the ambiguous case the union guards against. Lets it reach the existing `tgt_nid = None` deferral every receiver-typed resolver already uses for other cases, so it can only ever reach an edge through a guarded, receiver-typed resolver — never the unguarded bare-name path a real god node would need.

## Verified

- Python: `Session.open()` cross-file class-qualified call now resolves (EXTRACTED).
- Swift: reproduced the reporter's own real-world example directly — `AppleReadyStreamFinder.find -> RemuxPlayback.open` now resolves (INFERRED, via the field-typed receiver resolver).
- Same-file god-node guard still holds: `other.open()` does not bind to an unrelated same-file bare `open()` function, and a bare call to the actual builtin `open(...)` still produces no phantom edge.
- Full suite green, no regressions.

## Scope: `go.py`/`rust.py` intentionally not touched

The issue asks for the identical change in `extractors/go.py` (~L480) and `extractors/rust.py` (~L372) too. I traced where their `raw_calls` entries for member calls actually go: `extract.py`'s shared cross-file resolver skips **every** `is_member_call` entry unconditionally —

```python
# Skip member-call callees: obj.log() → "log" has no import evidence
# and collides with any top-level function named "log" in the corpus.
if rc.get("is_member_call"):
    continue
```

— regardless of builtin status, and neither Go nor Rust has a dedicated receiver-typed resolver the way Python (`_resolve_python_member_calls`), Java, C#, Swift, TypeScript, C++, and Objective-C do. Making the same edit there would be safe (it wouldn't reintroduce the god-node risk) but currently has **zero observable effect** on the final graph, since nothing downstream would ever consume the preserved `raw_calls` entry. Rather than ship an inert change, I left it out — worth revisiting once either language gets its own receiver-typed cross-file resolver.

Fixes #3381.

### DIFF
--- a/graphify/extractors/engine.py
+++ b/graphify/extractors/engine.py
@@ -5607,7 +5607,22 @@ def _extract_generic(
                         # Try reading the node directly (e.g. Java name field is the callee)
                         callee_name = _read_text(func_node, source)
 
-            if callee_name and callee_name not in _LANGUAGE_BUILTIN_GLOBALS:
+            # _LANGUAGE_BUILTIN_GLOBALS is one union across every language, right for
+            # a BARE call (String(x) really would become a god node) but wrong for a
+            # MEMBER call: `open` is a Python builtin and `Set` a JavaScript one, so
+            # session.open() in Swift or _server.Set() in C# named after another
+            # language's builtin was silently discarded outright -- no same-file
+            # edge, but also no raw_calls entry, so cross-file resolution never even
+            # got a chance to try (#3381). A member call carries a receiver, so it
+            # is not the ambiguous case the union guards against; letting it through
+            # here while forcing tgt_nid = None below (same as the existing
+            # receiver-typed defers just past this comment) means it can only ever
+            # reach an edge through a guarded, receiver-typed resolver, never the
+            # unguarded bare-name path a real god node would need.
+            _builtin_member_call = is_member_call and callee_name in _LANGUAGE_BUILTIN_GLOBALS
+            if callee_name and (
+                callee_name not in _LANGUAGE_BUILTIN_GLOBALS or _builtin_member_call
+            ):
                 # Python member calls defer to receiver-based resolution unless the
                 # receiver is known to stay in the current class. Falling back to a
                 # bare method name for an unresolved/lowercase receiver (`d.get()` or
@@ -5636,7 +5651,7 @@ def _extract_generic(
                 _java_defer = (
                     config.ts_module == "tree_sitter_java" and is_member_call
                 )
-                if _python_defer or _java_defer or (
+                if _python_defer or _java_defer or _builtin_member_call or (
                     is_member_call
                     and member_receiver
                     and (
--- a/tests/test_extract.py
+++ b/tests/test_extract.py
@@ -1329,6 +1329,64 @@ def test_python_qualified_class_method_call_resolves_extracted(tmp_path):
     assert call_edges[0]["confidence"] == "EXTRACTED"
 
 
+def test_builtin_named_member_call_still_resolves_cross_file(tmp_path):
+    """#3381: _LANGUAGE_BUILTIN_GLOBALS is one union across every language, right
+    for a BARE call (String(x) really would become a god node) but wrong for a
+    MEMBER call -- `open` is a Python builtin, so Session.open() used to be
+    silently discarded outright: no same-file edge, but also no raw_calls entry,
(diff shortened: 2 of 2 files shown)


ayushcodes10:
@safishamsi opened this against #3381 (a member call named after another language's builtin, e.g. Swift's session.open() colliding with Python's open, was silently discarded with no raw_calls entry at all). Verified against real cross-file Python and Swift setups, including the reporter's own AppleReadyStreamFinder.find -> RemuxPlayback.open example. Scoped to the shared engine only; the PR description explains why I left go.py/rust.py out (traced their raw_calls path and it's currently inert there, no receiver-typed resolver to consume it). Happy to address any feedback.

safishamsi:
Shipped in **v0.9.56** — landed on `v8` via cherry-pick with your authorship preserved. Thanks @ayushcodes10 — a builtin-named member call is now handed to cross-file resolution so a real user-defined method with that name links. Release: https://github.com/Graphify-Labs/graphify/releases/tag/v0.9.56

PR #3395, opened 2026-09-07
feat(go): resolve receiver-typed member calls, and park them across repos

### LINKED ISSUE #3393: Go member calls resolve to nothing across files, and cannot cross a repo boundary at all
## What happens

A Go method call on a receiver resolves only when the bare method name happens to
match a symbol declared in the caller's own file. Anything else produces no edge:

```go
// svc/greeter.go
package svc

type Greeter struct{}

func (g *Greeter) Greet() {}
```

```go
// svc/app.go
package svc

type App struct {
	greeter *Greeter
}

func (a *App) Run() { a.greeter.Greet() }   // no `calls` edge
```

`extract_go` reads the receiver's identifier and discards it unless it names an
imported package, so the shape Go dependency injection actually takes — a struct
field used from a method — contributes nothing to the call graph. The shared
cross-file pass then skips member callees outright (`obj.log()` → `"log"` has no
import evidence), and no `_resolve_go_*` resolver is registered to pick them up.
No test in the repo asserts that a cross-file Go receiver method call yields an
edge; `test_go_builtin_call_targets.py` states the omission explicitly.

Merging two repos loses the same edges for the same reason, plus one more: Go
nodes never carry the `_callable` / `_callable_class` markers (#2438), because
`extract_go` is the one extractor outside the tree-sitter engine and the engine is
the only place that stamps them. `link_cross_repo_member_calls` requires
`_callable_class` on the declaring type, so a Go type could not answer a parked
call even if one existed.

## What it should do

Bind the call through the receiver's declared type, as the Swift (#1356), C++
(#1547), ObjC (#1556) and C# (#1609) resolvers already do for their languages —
and when the type is declared nowhere in the build, park the call so a merged
graph can finish it (#3152).

Go writes the receiver's type down in four places, all of them cheap to read:

| shape | source |
|---|---|
| `type App struct { greeter *Greeter }` | struct field |
| `func (s *Server) M()` / `func f(g *Greeter)` | parameter |
| `var g Greeter` | var declaration |
| `g := Greeter{}` / `g := &Greeter{}` | composite literal |

The single-definition guard has to stay: a Go type node id folds in the package
directory, so two packages declaring `Greeter` are two nodes and without import
evidence neither one is the answer.

## Why it matters

Go is a backend language and services are split across repositories; the call
graph across that boundary is the reason to merge two graphs at all. Together with
the ObjC (#3384), TS/JS (#3386), Kotlin (#3388) and PHP (#3390) gaps, this covers
the languages a mobile-plus-backend org actually ships.

PR to follow.

xiongjianxu:
Fixes #3393. Follows the same two-commit shape as #3385 (ObjC), #3387 (TS/JS),
#3389 (Kotlin) and #3391 (PHP): the single-repo half first, then the park.

### Commit 1 — resolve Go member calls through the receiver's declared type

`extract_go` now exports a per-file `go_type_table` built from the four places a Go
receiver's type is written down: a struct field, a parameter (which covers the method
receiver `func (s *Server)`), `var x T`, and `x := T{}` / `x := &T{}`.
`_resolve_go_member_calls` looks the receiver up there, takes the single declaration of
that type, and emits the `calls` edge to its method.

It also stamps `_callable` / `_callable_class` on Go declarations. Go is the one
extractor outside the tree-sitter engine, so its nodes carried neither marker and were
invisible to every consumer that reads them off the node. Scoped check: the
indirect-call guard only consults them for `rc["indirect"]` entries, which Go never
produces, and `link_shared_type_declarations` additionally requires
`metadata.namespace`, which Go never sets — so nothing existing changes shape.

### Commit 2 — park the call when the type is in another repo

Parks the pair on the caller node and adds `"go": {".go"}` to the merge pass's
language guard.

### Design calls worth a look

- **The receiver travels as `member_receiver`, not `receiver`.** In this extractor
  `receiver` already means "imported package name", and — more importantly — the
  Swift/Python/Ruby member-call resolvers select on `is_member_call` plus `receiver`
  *without checking the language*. A Go name placed there would bind to their types in
  a mixed corpus. A key of its own is the precise fix; widening those three resolvers'
  gates is a separate change.
- **`s.logger.Log()` is read as a call on the `logger` field only when `s` is the
  enclosing method's own receiver.** This is the Go analogue of PHP's
  `$this->prop->method()`. The table is flat and file-scoped, so any other head could
  be a package or a variable whose type it cannot confirm; `p.greeter.Greet()` on a
  parameter stays unresolved on purpose.
- **Names are matched case-sensitively, unlike the sibling resolvers.** Go exports by
  capitalisation, so `Run` and `run` on one type are two different methods with two
  different visibilities. `cross_repo_calls._key` already preserves case and documents
  why, so both halves agree.
- **Go does not defer.** A member call reaches `raw_calls` only once the bare callee
  name misses in the caller's file, which is already the existing behaviour, so in-file
  resolution is byte-identical. `test_a_same_file_method_call_keeps_its_extracted_edge`
  pins that the pass neither downgrades nor doubles an in-file edge.
- **`confidence_score` is 0.85, not 0.8.** The rubric in `references/extraction-spec.md`
  is the discrete set {0.55, 0.65, 0.75, 0.85, 0.95}; same tier, a value the documented
  scale contains.
- **`.go` alone in the language guard.** Kotlin's set includes `.java` because the JVM
  classpath is one namespace; Go and Java share nothing, and a matching type name across
  a Go service and a Java service in one merge is a coincidence.
- **A single type name or nothing.** `Greeter` and `*Greeter` qualify; a slice, map or
  channel element is not the receiver of `x.M()`, and `pkg.Greeter` names a type no bare
  local name can stand for.

### Known costs

- `g := NewGreeter()` stays unresolved: typing it means reading the constructor's return
  type, which is a separate change.
- The type table is flat per file, so first binding wins across all four sources — a
  parameter named like a field in another function does not retype the field's calls.
  Pinned by `test_the_first_binding_of_a_name_wins`.
- Framework receivers on package-level values that no local table names remain
  unresolved, as before.

### Tests

`tests/test_go_receiver_member_calls.py`, 11 new cases — 5 positive shapes, 5 negatives
(constructor return, chain not rooted at the method's receiver, package-qualified type,
two packages declaring one name, first-binding-wins) and one that pins in-file
behaviour. 6 of them fail on `v8`. On the merge side, a `go-struct-field` arm in
`test_each_language_parks_the_call_and_the_merge_finishes_it`, plus a Go/Java negative
and a case-sensitivity negative.

Full suite: 5323 passed, 93 skipped, no regressions (`test_ollama_retry_cap.py`
deselected — it fails on `v8` too). `ruff` clean, no added line over 100 columns.

### DIFF
--- a/graphify/cross_repo_calls.py
+++ b/graphify/cross_repo_calls.py
@@ -41,6 +41,7 @@ UNRESOLVED_CALLS_KEY = "unresolved_calls"
 _LANG_SUFFIXES: dict[str, frozenset[str]] = {
     "cpp": frozenset({".cpp", ".cc", ".cxx", ".hpp", ".hh", ".hxx", ".h", ".cu", ".cuh"}),
     "csharp": frozenset({".cs"}),
+    "go": frozenset({".go"}),
     "java": frozenset({".java"}),
     "swift": frozenset({".swift"}),
 }
--- a/graphify/extract.py
+++ b/graphify/extract.py
@@ -4537,7 +4537,7 @@ def _resolve_go_member_calls(
     names the declared type of every struct field, parameter (the method receiver
     included), ``var x T`` and ``x := T{}`` binding; this pass looks the receiver up
     there, takes the single declaration of that type, and emits the ``calls`` edge to its
-    method. Always INFERRED: the type comes from the table, never from the call site.
+    method. Always INFERRED: the type comes from a declaration, never from the call site.
 
     Names are matched case-sensitively, unlike the sibling resolvers. Go exports by
     capitalisation, so `Run` and `run` on one type are two different methods with two
@@ -4567,11 +4567,13 @@ def _resolve_go_member_calls(
     # A Go type node id folds in the package directory, so two packages declaring the
     # same name stay two entries and the single-definition guard below bails: without
     # import evidence neither one is the answer.
+    # Only `.go` declarations count: `_callable_class` is corpus-wide, so a same-named
+    # Java class would both answer a Go receiver and hide that nothing local declares it.
     type_def_nids: dict[str, list[str]] = {}
     node_by_id: dict[str, dict] = {}
     for n in all_nodes:
         node_by_id[n.get("id")] = n
-        if n.get("_callable_class") and n.get("source_file"):
+        if n.get("_callable_class") and str(n.get("source_file", "")).endswith(".go"):
             type_def_nids.setdefault(_key(n.get("label", "")), []).append(n["id"])
 
     method_index: dict[tuple[str, str], list[str]] = {}
@@ -4586,7 +4588,10 @@ def _resolve_go_member_calls(
     existing_pairs = {(e.get("source"), e.get("target")) for e in all_edges}
     for rc in raw:
         receiver, callee, caller = rc["member_receiver"], rc["callee"], rc["caller_nid"]
-        type_name = type_table_by_file.get(rc.get("source_file", ""), {}).get(receiver)
+        # The enclosing method's own receiver is typed by its declaration, which beats the
+        # flat table: two types in one file may both name their receiver `s`.
+        type_name = rc.get("receiver_type") or type_table_by_file.get(
+            rc.get("source_file", ""), {}).get(receiver)
         if not type_name or type_name in _LANGUAGE_BUILTIN_GLOBALS:
             continue
         type_defs = type_def_nids.get(_key(type_name), [])
--- a/graphify/extract.py
+++ b/graphify/extract.py
@@ -4542,6 +4542,9 @@ def _resolve_go_member_calls(
     Names are matched case-sensitively, unlike the sibling resolvers. Go exports by
     capitalisation, so `Run` and `run` on one type are two different methods with two
     different visibilities, and folding them would pick whichever came last.
+
+    A receiver typed to a type this corpus declares nowhere is parked on the caller for
+    a merged graph to finish (#3152).
     """
     raw = [
         rc
@@ -4587,7 +4590,13 @@ def _resolve_go_member_calls(
         if not type_name or type_name in _LANGUAGE_BUILTIN_GLOBALS:
             continue
         type_defs = type_def_nids.get(_key(type_name), [])
-        if len(type_defs) != 1:  # ambiguous or absent -> bail (god-node guard)
+        if not type_defs:
+            # Declared nowhere here — usually "in a repo this build does not contain",
+            # so park it for the merge (#3152). The extractor's language tag already
+            # says who is asking, so no suffix sniff is needed.
+            _park_unresolved_member_call(node_by_id.get(caller), callee, type_name, "go", rc)
+            continue
+        if len(type_defs) != 1:  # ambiguous -> bail (god-node guard)
             continue
         targets = method_index.get((type_defs[0], _key(callee)), [])
         if len(targets) != 1:
--- a/graphify/extract.py
+++ b/graphify/extract.py
@@ -4525,6 +4525,92 @@ def _resolve_kotlin_qualified_calls(
         })
 
 
+def _resolve_go_member_calls(
+    per_file: list[dict],
+    all_nodes: list[dict],
+    all_edges: list[dict],
+) -> None:
+    """Resolve Go member calls (``g.Greet()``) through the receiver's declared type.
+
+    The shared cross-file pass skips member calls, so a method call on a receiver whose
+    type is declared in another file resolved to nothing. The per-file ``go_type_table``
+    names the declared type of every struct field, parameter (the method receiver
+    included), ``var x T`` and ``x := T{}`` binding; this pass looks the receiver up
+    there, takes the single declaration of that type, and emits the ``calls`` edge to its
+    method. Always INFERRED: the type comes from the table, never from the call site.
+
+    Names are matched case-sensitively, unlike the sibling resolvers. Go exports by
+    capitalisation, so `Run` and `run` on one type are two different methods with two
+    different visibilities, and folding them would pick whichever came last.
+    """
+    raw = [
+        rc
+        for result in per_file
+        for rc in result.get("raw_calls", [])
+        if rc.get("language") == "go" and rc.get("is_member_call")
+        and rc.get("member_receiver") and rc.get("callee") and rc.get("caller_nid")
+    ]
+    if not raw:
+        return
+    type_table_by_file: dict[str, dict[str, str]] = {}
+    for result in per_file:
+        tt = result.get("go_type_table")
+        if tt and tt.get("path"):
+            type_table_by_file[tt["path"]] = tt.get("table", {})
+
+    def _key(label: object) -> str:
+        return str(label or "").strip().removeprefix(".").removesuffix("()")
+
+    # A Go type node id folds in the package directory, so two packages declaring the
+    # same name stay two entries and the single-definition guard below bails: without
+    # import evidence neither one is the answer.
+    type_def_nids: dict[str, list[str]] = {}
+    node_by_id: dict[str, dict] = {}
+    for n in all_nodes:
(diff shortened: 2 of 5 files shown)


xiongjianxu:
Self-audit pass over the bot review, plus one thing the review did not catch. Two fixes pushed:

**1. The declaration index was corpus-wide (the review's high-severity finding).** `_callable_class` is stamped by every extractor, so a Java `class Greeter` could answer a Go receiver typed `Greeter`, and — worse for this PR's purpose — its mere presence made `type_defs` non-empty, so the parking branch never ran and the call was silently dropped instead of handed to the merge. The index is now gated with `_lang_family(...) == "go"`, the same interop-family map the shared cross-file call resolver uses at the `candidate_family` check. Pinned by `test_a_type_from_another_language_never_answers_a_go_receiver`, which asserts both halves: no edge, and a parked entry.

**2. The flat per-file table could mistype a method's own receiver.** Go names receivers with one letter, so two types in one file both writing `func (s *T)` is ordinary; first-binding-wins then gave the second one the first one's type. The extractor now stamps `receiver_type` from the enclosing method's own declaration and the resolver prefers it over the table. Pinned by `test_the_methods_own_receiver_outranks_an_earlier_binding_of_the_name`, which fails on the previous commit with a `Store.Get -> Server.Save` edge.

On the `member_receiver` finding — that a language-agnostic resolver might consume it: the Swift/Python/Ruby member-call resolvers select on `is_member_call` **plus** `receiver`, and Go leaves `receiver` for the imported-package case only. That separation is why the new key exists rather than reusing `receiver`; putting a Go variable name in `receiver` is exactly what would let those resolvers bind it in a mixed corpus.

One gap this audit surfaced that is **not** in scope here and pre-exists on `v8`: `extract_go` mints a type node in every file that declares a method on the type, and `_disambiguate_colliding_node_ids` then splits the colliding ids by path. A package whose type and methods live in different files therefore has ≥2 type nodes, and every single-definition guard — this resolver's included — bails. C/C++/ObjC avoid it through `_merge_decl_def_classes`; Go has no equivalent. Filing that separately.

Verification: `ruff check graphify tests` clean; full suite 5325 passed / 93 skipped, with only the four pre-existing `tests/test_ollama_retry_cap.py` env failures that `v8` also shows here.

xiongjianxu:
Filed the type-node gap I mentioned above separately: #3399, with PR #3400 based on `v8` (independent of this one, no ordering between them).

It matters here: this PR's resolver needs exactly one declaration of the receiver's type, and today a package that declares `type Server` in one file and its methods in others produces one type node per file — so the resolver bails on any package larger than a single file. #3400 folds that id-collision before disambiguation, the way `_merge_decl_def_classes` already does for C/C++/ObjC.

xiongjianxu:
One of the five findings was real and is fixed in `d375863`; the other four do not hold.

**Receiver type table conflates variable names across function scopes (go.py:100) — fixed,
and it produced a *wrong edge*, not just a miss.** Go files reuse `s`, `c`, `w`, `r` as
parameter names in every function, so a table flat over the file answered one function's
receiver with a sibling's parameter type:

```go
func First(g *Greeter)  {}
func Second(g *Shouter) { g.Greet() }   // before: bound to Greeter.Greet
```

Each function body now carries its own parameter/local table (`_go_receiver_type_table` run on
the function node rather than the file root), consulted before the flat one; the flat table
stays as the fallback that types a struct field. Pinned by
`test_a_parameter_binds_only_inside_its_own_function`.

**Pointer receivers are not looked up without dereferencing (extract.py:4599) — no defect.**
`_go_single_type_name` unwraps `pointer_type` before reading the identifier, and the
`method_declaration` path applies `.lstrip("*")`. `func Boot(srv *Server) { srv.Close() }`
resolves to `Server.Close` on the current branch; verified.

**`member_receiver_type` "never falls back to the file-flat table" (go.py:549) — no defect,
twice.** The fallback lives in the resolver, not the extractor:
`type_name = rc.get("receiver_type") or type_table_by_file[...].get(receiver)`. The field is
deliberately narrow so the *stated* type wins over a flat-table guess; after `d375863` it also
carries the enclosing function's own scope.

**`_calls` collapses duplicate call edges by label pair (test helper) — intentional.** The
helper is a readability shim shared by the five sibling PRs; the assertions are about whether a
pair exists and which node id it names, and the cases are single-call bodies. Counting duplicates
would test the helper, not the resolver.

One thing this PR still needs to fire on idiomatic multi-file packages: #3400, which folds the
type-node copies each file declaring a method mints. Without it a package whose type and methods
are split across files produces no member-call edge at all — reproduced on this branch.

Full suite green: 5326 passed, 93 skipped; ruff clean.

xiongjianxu:
Checked all three against the branch. One was a real staleness my own fix introduced and is
cleaned up in `37497b5`; the other two do not hold.

**File-level `type_table` "computed but unused" (go.py:519) — it is load-bearing, but the finding
did point at real staleness next to it.** The flat table is the fallback that types a *struct
field*, which no per-function scope can see because the struct is declared outside every function
body. Proof: neutralise just the fallback expression in the resolver
(`type_name = rc.get("receiver_type")`, dropping the `or type_table_by_file...`) and exactly one
test fails —
`test_a_struct_field_types_the_receiver_through_the_methods_own_receiver`, i.e. `a.greeter.Greet()`.
The other twelve pass. So the shadowing is the intended precedence, not a dead branch.

What *was* stale: `test_the_first_binding_of_a_name_wins` still described the flat table's
cross-function shadowing as the intended behaviour ("a parameter named like an earlier one has to
lose"), which is precisely what `d375863` removed, and it only asserted one of the two directions.
Its fixture now pins both — `Run(g *Greeter)` reaches greeter.go and `Also(g *Other)` reaches
other.go — under the name `test_a_parameter_binds_only_inside_its_own_function`.

**`method_index` keyed on the wrong edge end — no defect.** The Go extractor emits
`add_edge(parent_nid, method_nid, "method", ...)`, so `source` *is* the type and `target` is the
method. Every passing member-call test in the file goes through that index; had the direction been
reversed, none would resolve.

**Go pointer receiver types not normalized (extract.py:4595) — no defect.** Nothing with a `*`
ever reaches the lookup: `_go_single_type_name` unwraps `pointer_type` before reading the
identifier, and the `method_declaration` path applies `.lstrip("*")`. Verified on this branch for
both a pointer parameter and a pointer `var`:

```go
func Boot(srv *Server)  { srv.Close() }        // -> Server.Close, INFERRED
func Boot2() { var srv2 *Server; srv2.Close() } // -> Server.Close, INFERRED
```

Full suite green: 5325 passed, 93 skipped; ruff clean.

xiongjianxu:
Checked; the finding does not hold — the `receiver` field is byte-identical to `v8`.

**`receiver` for Go member calls (go.py:602) — unchanged, not "no longer populated".** On `v8` the
record already read `"receiver": package_receiver`, and `package_receiver` was already assigned
only on the *not*-a-member-call branch (`if not is_member_call: package_receiver = receiver_name`).
Diffing this branch against `v8` filtered to those lines returns nothing:

```
$ git diff v8 -- graphify/extractors/go.py | grep -E '^[-+].*(package_receiver|"receiver")'
$          # (empty)
```

This PR only *adds* `member_receiver` alongside it. Keeping them apart is the point: the
Swift/Python/Ruby member-call resolvers select on `is_member_call` plus `receiver` without checking
`language`, so a Go variable name placed in `receiver` would let them bind it to their own types in
a mixed corpus.

The consumer that does read `receiver`/`import_path` for Go is the shared cross-file pass at
extract.py:7313, which filters candidates by exact `_go_import_path_for_file` match — package
evidence, which a variable name is not. Behaviour there is identical on both refs: the same
package-qualified fixture (`import "util"` + `util.Helper()`) produces the same edge set on
`37497b5` as on `v8`.

Full suite green: 5325 passed, 93 skipped; ruff clean. CI green on `37497b5` (4/4).

xiongjianxu:

Issue #3341, opened 2026-09-04
ReDoS: _normalize_ts_import_types hangs forever on some .ts files (catastrophic backtracking in _TS_IMPORT_TYPE_CALL_RE)


paunescumihai:
## Summary

`_normalize_ts_import_types` (`graphify/extract.py`) can hang **forever** on a single ordinary TypeScript file. `_TS_IMPORT_TYPE_CALL_RE` has two nested lazy quantifiers inside a `+` group, which produces **catastrophic regex backtracking** when a line contains several `import(...)` occurrences without a following `>(` that lets the overall match succeed.

Introduced with the #3154 normalizer; still present in 0.9.53. During `graphify update .` one worker pegs a full core at ~100% CPU and the AST-extraction counter stops advancing (looks like a pool deadlock on a many-core box, because one pinned core reads as ~3% aggregate CPU).

## The regex

```python
_TS_IMPORT_TYPE_CALL_RE = re.compile(
    rb"<((?:[^;{}]*?\bimport\s*\([^()]+\)[^;{}]*?)+)>(?=\s*\()"
)
```

The filler `[^;{}]*?` appears **both** before and after `import(...)` inside a `+`-repeated group. When the trailing `>(` lookahead fails, the engine can partition the filler between group iterations in ~2^k ways for k `import(` tokens, so it backtracks exponentially before giving up on the line.

## Minimal repro (pure `re`, no graphify needed)

```python
import re, time
OLD = re.compile(rb"<((?:[^;{}]*?\bimport\s*\([^()]+\)[^;{}]*?)+)>(?=\s*\()")
for k in (3, 5, 7, 9, 11):
    s = b"x < " + b" ".join(b'import("m%d")' % i for i in range(k)) + b" and more text no close"
    t = time.time(); OLD.search(s); print(f"imports={k}: {time.time()-t:.3f}s")
```

```
imports=3: 0.000s
imports=5: 0.005s
imports=7: 0.156s
imports=9: 4.914s
imports=11: (does not finish)
```

Real-world trigger was a 13 KB `.ts` file mixing `<`, several `import("...")` type references and generic calls; `_normalize_ts_import_types(source)` never returned.

## Suggested fix

Forbid `import` (and `<`/`>`) inside the filler so the decomposition is unique — same matches, linear time:

```python
_TS_IMPORT_TYPE_CALL_RE = re.compile(
    rb"<((?:(?!import)[^;{}<>])*"
[...]

PR #3397, opened 2026-09-07
Kill orphaned workers before the rebuild watchdog exits

### LINKED ISSUE #3396: Rebuild watchdog's os._exit orphans a mid-task ProcessPoolExecutor worker
## Summary

Filing separately from #3341, which is being closed as fixed by #3210. @Yo-TR's comment there (https://github.com/Graphify-Labs/graphify/issues/3341#issuecomment-5569597086) flagged a real, separate finding worth tracking on its own: when a git-hook rebuild watchdog fires while a ProcessPoolExecutor worker is still mid-task, os._exit(1) kills only the rebuild process itself. It skips every cleanup path, including the pool's own context-manager shutdown, so a worker that is still running becomes orphaned (reparented to PID 1 on POSIX) and keeps going unsupervised for as long as whatever it was doing takes. A worker stuck in catastrophic regex backtracking (the exact shape #3341 fixed) was observed surviving 2.5 days that way on macOS.

## Where

Both git-hook rebuild bodies in `graphify/hooks.py` (`_REBUILD_BODY_COMMIT`, `_REBUILD_BODY_CHECKOUT`) arm a `GRAPHIFY_REBUILD_TIMEOUT` watchdog. On platforms without `signal.SIGALRM`, the fallback is a `threading.Timer` whose callback calls `os._exit(1)` directly with no cleanup:

```python
def _bail():
    print(f'[graphify hook] graphify rebuild exceeded {_timeout}s', flush=True)
    os._exit(1)
```

`extract.py` already documents this exact risk for its own 1-worker special case (`# ...the parent's rebuild watchdog (os._exit) can orphan a worker that is mid-task`), but that only covers `max_workers == 1`; a real multi-worker pool spawned for a larger corpus is not covered.

## Suggested fix

Terminate any live worker processes before exiting, as @Yo-TR suggested. The watchdog fires on a separate timer thread with no reference to the pool object, but `multiprocessing.active_children()` enumerates every live worker process regardless of which thread asks. Kill (SIGKILL), not terminate (SIGTERM) — a worker stuck in a C-level call like regex backtracking never gets a chance to act on SIGTERM.

Environment (from the original report): graphifyy 0.9.53 → 0.9.55, CPython 3.14.7 (Homebrew), macOS 26 (Darwin 25.5.0).

ayushcodes10:
## What

`os._exit(1)`, used by the git-hook rebuild watchdog's SIGALRM-less fallback, terminates only the rebuild process itself — it skips every cleanup path, including a `ProcessPoolExecutor`'s own context-manager shutdown. A worker still running at the moment the watchdog fires is orphaned (reparented to PID 1 on POSIX) and keeps going unsupervised for as long as whatever it was doing takes.

`extract.py` already documents this exact risk for its own 1-worker special case (`max_workers == 1` deliberately falls back to sequential extraction in-process, specifically to avoid this), but that only covers that one case — a real multi-worker pool spawned for a larger corpus was not covered.

## Real-world trigger

Reported on #3341 (a ReDoS in the old TS import-type normalizer, since fixed by #3210): a worker stuck in catastrophic regex backtracking survived as an orphan pinning a CPU core for 2.5 days on macOS after the parent rebuild process exited.

## Fix

The watchdog fires on a separate timer thread with no reference to the pool object (it's a local variable deep in an unrelated call stack), but `multiprocessing.active_children()` enumerates every live worker process regardless of which thread asks. Kill them before exiting.

Verified `.kill()` (SIGKILL) actually terminates a worker stuck in exactly the reported failure mode — a process executing a long C-level `re.match()` call never returns control to the interpreter to check for a Python-level signal, so `.terminate()` (SIGTERM) would not have worked here; SIGKILL is enforced by the kernel unconditionally:

```
active children: [<SpawnProcess ... pid=97110 ...>]
killing (SIGKILL) 97110
97110 alive: False
```

Added an AST-based test (matching this file's existing convention for the embedded, shell-quoted hook bodies) verifying both rebuild bodies enumerate and kill live children in a statement that runs before `os._exit`, not after.

Fixes #3396.

### DIFF
--- a/graphify/hooks.py
+++ b/graphify/hooks.py
@@ -134,7 +134,7 @@ fi
 # double-quote, $, backtick or backslash characters: it is carried inside a
 # shell double-quoted `-c "..."` argument (see _detached_launch).
 _REBUILD_BODY_COMMIT = """\
-import os, signal, sys, threading
+import os, signal, sys, threading, multiprocessing
 from pathlib import Path
 
 changed_raw = os.environ.get('GRAPHIFY_CHANGED', '')
@@ -156,6 +156,8 @@ try:
         else:
             def _bail():
                 print(f'[graphify hook] graphify rebuild exceeded {_timeout}s', flush=True)
+                for _child in multiprocessing.active_children():
+                    _child.kill()
                 os._exit(1)
             _watchdog = threading.Timer(_timeout, _bail)
             _watchdog.daemon = True
@@ -191,7 +193,7 @@ except Exception as exc:
 _REBUILD_BODY_CHECKOUT = """\
 from graphify.watch import _rebuild_code, _apply_resource_limits
 from pathlib import Path
-import os, signal, sys, threading
+import os, signal, sys, threading, multiprocessing
 try:
     _apply_resource_limits()
     _timeout = int(os.environ.get('GRAPHIFY_REBUILD_TIMEOUT', '600'))
@@ -202,6 +204,8 @@ try:
         else:
             def _bail():
                 print(f'[graphify] graphify rebuild exceeded {_timeout}s', flush=True)
+                for _child in multiprocessing.active_children():
+                    _child.kill()
                 os._exit(1)
             _watchdog = threading.Timer(_timeout, _bail)
             _watchdog.daemon = True
--- a/graphify/hooks.py
+++ b/graphify/hooks.py
@@ -151,7 +151,18 @@ try:
     _timeout = int(os.environ.get('GRAPHIFY_REBUILD_TIMEOUT', '600'))
     if _timeout > 0:
         if hasattr(signal, 'SIGALRM'):
-            signal.signal(signal.SIGALRM, lambda *_: (_ for _ in ()).throw(TimeoutError(f'graphify rebuild exceeded {_timeout}s')))
+            def _sigalrm_bail(*_a):
+                # Killing here, before the exception unwinds, matters: once
+                # TimeoutError starts propagating it passes straight through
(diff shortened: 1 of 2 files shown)


ayushcodes10:
@safishamsi opened this against #3396, filed separately from @Yo-TR's comment on #3341 about an orphaned ProcessPoolExecutor worker surviving 2.5 days after the rebuild watchdog's os._exit. Verified the kill mechanism directly against a worker stuck in the exact reported failure mode (catastrophic regex backtracking, a C-level call SIGTERM can't preempt but SIGKILL can). Happy to address any feedback.

ayushcodes10:
Investigated the bot's two findings here and both were correct, this exposed something more serious than the original scope: the SIGALRM path (the primary one, used on macOS/Linux) has the same orphaning problem the no-SIGALRM fallback had, just manifesting differently. Reproduced directly: a worker stuck in an unconditional loop, SIGALRM firing after 2s, process still running 15s later, because pool.shutdown(wait=True) in the with-block's own __exit__ blocks before the except TimeoutError handler ever gets control. Pushed a fix that kills workers inside the SIGALRM handler itself, before it raises, verified against the same repro (now exits cleanly at 2s). Added a matching regression test.

safishamsi:
Shipped in **v0.9.56** — landed on `v8` via cherry-pick with your authorship preserved. Thanks @ayushcodes10 — extraction workers are killed before the rebuild watchdog exits instead of being orphaned. Release: https://github.com/Graphify-Labs/graphify/releases/tag/v0.9.56

PR #3400, opened 2026-09-07
Go: fold a package's type into one node when its methods span files (#3399)

### LINKED ISSUE #3399: Go: a package's type is fragmented into one node per file that declares a method on it
## Summary

A Go type whose methods live in other files of the same package becomes **several nodes**, one
per file, each owning only the methods declared there. The declaring file's node owns none of
them. Every downstream single-definition guard then reads the set as an ambiguity and bails, so
an idiomatic multi-file package loses type-level resolution.

C/C++/ObjC hit the same id-collision between a header declaration and its impl and avoid this
through `_merge_decl_def_classes` (`graphify/extractors/resolution.py`), which folds the
collision *before* `_disambiguate_colliding_node_ids` splits it. Go has no equivalent.

## Reproducer

```
svc/a.go   package svc; type Server struct{}; func Run(srv *Server) { srv.Close() }
svc/b.go   package svc; func (s *Server) Close() {}
svc/c.go   package svc; func (s *Server) Save() {}
```

```
$ graphify extract svc/
```

Actual — three `Server` nodes, and the methods hang off two of them:

```
nodes labelled Server:
    ..._svc_a_go_svc_server | a.go
    ..._svc_b_go_svc_server | b.go
    ..._svc_c_go_svc_server | c.go
edges:
    a.go   -> Server    | contains
    Run()  -> Server    | references (parameter_type)
    Server -> .Close()  | method      <- the b.go node
    Server -> .Save()   | method      <- the c.go node
```

Expected — one `Server`, sourced at its declaration, owning both methods:

```
    svc_server | a.go
    a.go   -> Server   | contains
    Run()  -> Server   | references (parameter_type)
    Server -> .Close() | method
    Server -> .Save()  | method
```

## Why

`extract_go` mints the receiver's type node in **every** file that declares a method on it
(`graphify/extractors/go.py`, the `method_declaration` branch), keyed
`_make_id(pkg_scope, receiver_type)` where `pkg_scope` is the package directory's name. All those
nodes therefore share one id and differ only in `source_file` — which is exactly the input
`_disambiguate_colliding_node_ids` is built to split, so it salts each one with its own path.

## Impact

* `references` to the type land on whichever fragment the referring file produced, so a type's
  incoming uses and its methods are no longer reachable from one node.
* The id is salted with a file path, so it is neither the canonical `{pkg}_{type}` form nor
  stable across machines for the affected types.
* Every resolver that requires exactly one declaration of a name (member-call resolution, the
  Go type-reference rewire) bails, which is silent under-reporting rather than a visible error.

Splitting one Go type into two files is not an edge case; it is how any package larger than a
single file is written.

## Fix

Fold the id-collision before disambiguation, mirroring `_merge_decl_def_classes`. The nodes
already share an id, so no edge re-pointing is needed — only the redundant duplicates are
dropped and the declaring file's node survives.

Guards against a false merge: all members of the collision must be `.go` files in **one
directory** (the id folds in only the directory's *name*, so `a/svc` and `b/svc` collide while
being different packages), and a `_test.go` member stops the fold, since its package clause may
be the separate external `svc_test` package declaring a type of the same name — `extract_go`
does not parse the package clause, so same-directory is not evidence of same package there.

PR follows.

xiongjianxu:
Fixes #3399.

A Go type whose methods live in other files of the same package became several nodes, one per
file, each owning only the methods declared there — and the declaring file's node owning none of
them. `extract_go` mints the receiver's type in every file that declares a method on it, keyed
`_make_id(pkg_scope, receiver_type)`, so those nodes share one id and differ only in
`source_file`; `_disambiguate_colliding_node_ids` reads that as a collision between distinct
entities and salts each one with its own path.

## Change

One pass, `_merge_go_package_types`, run immediately after `_merge_decl_def_classes` and for the
same reason: before disambiguation, so it sees one `source_file` per id and leaves it alone. The
colliding nodes already share an id, so every edge already points at the survivor — only the
redundant duplicates are dropped, no edge moves.

The declaring file wins, identified as the only file with a `contains` edge to the type. When the
declaration lives outside the corpus the whole group is method-only; one node is still the
answer, so it folds on the lowest path.

## Guards against a false merge

* **One directory.** The id folds in only the directory's *name*, so `a/svc` and `b/svc` collide
  on id while being different packages. Those stay split, as today.
* **`_test.go` stops the fold.** Its package clause may be the separate external `svc_test`
  package declaring a type of the same name. `extract_go` derives `pkg_scope` from the directory
  and never parses the package clause, so same-directory is not evidence of same package there.
* **Bare type labels only.** A Go type name holds neither `.` nor `(`, while every other label
  the extractor mints does (`Run()`, `.Close()`, the file node's `a.go`), so a package whose
  helper function and a method share a name is untouched.

## Verification

Before, on the three-file package from the issue: three `Server` nodes, `.Close()` on the `b.go`
one and `.Save()` on the `c.go` one. After: one `svc_server` sourced at `a.go` owning both, with
the `parameter_type` reference from `Run()` pointing at it.

`tests/test_go_package_type_fold.py` — 6 tests: methods across files fold onto one owner; a
reference reaches the node that owns the methods; the declaring file survives a lower-sorting
method file; and the three guards above. The three positive tests fail on `v8` and pass here;
the guard tests pass both ways, pinning the no-regression half.

Full suite: 5315 passed, 93 skipped, `ruff check graphify tests` clean. The only failures in this
environment are the pre-existing `tests/test_ollama_retry_cap.py` ones (no `openai` module
installed), which fail identically on plain `v8`.

## Relation to #3395

Independent of it and based on `v8`. The Go member-call resolver added there needs exactly one
declaration of the receiver's type, so on a multi-file package it currently bails; this fold is
what lets it fire. Neither PR needs the other to land first.

### DIFF
--- a/graphify/extract.py
+++ b/graphify/extract.py
@@ -103,6 +103,7 @@ from graphify.extractors.resolution import (  # noqa: E402,F401
     _load_workspace_packages,
     _match_tsconfig_alias,
     _merge_decl_def_classes,
+    _merge_go_package_types,
     _node_disambiguation_source_key,
     _package_entry_candidates,
     _parse_js_tree,
@@ -6465,6 +6466,10 @@ def extract(
     # them (foo_h vs foo_cpp), so the collapse must happen first. Collapsing here
     # also means disambiguation sees one source_file per id and won't split them.
     _merge_decl_def_classes(all_nodes, all_edges)
+
+    # Same fold for a Go package's type, which every file declaring a method on it mints
+    # again under one id; also before disambiguation, which would otherwise split them.
+    _merge_go_package_types(all_nodes, all_edges)
 
     # Remap file node IDs from absolute-path-derived to the canonical
     # {parent_dir}_{stem} spec form so (a) graph.json edge endpoints are stable
--- a/graphify/extractors/resolution.py
+++ b/graphify/extractors/resolution.py
@@ -2432,6 +2432,79 @@ def _merge_decl_def_classes(
         rewritten.append(e)
     all_edges[:] = rewritten
 
+def _merge_go_package_types(
+    all_nodes: list[dict],
+    all_edges: list[dict],
+) -> None:
+    """Merge a Go package's type declared in one file with the copies minted by the
+    other files that declare methods on it into ONE node (#3399).
+
+    ``extract_go`` keys a type on its package directory, so ``type Server`` in ``a.go``
+    and ``func (s *Server) Close()`` in ``b.go`` mint the SAME id twice, differing only
+    in ``source_file``. Left alone, ``_disambiguate_colliding_node_ids`` splits them
+    apart by path, fragmenting one type into several partial nodes that each own some of
+    its methods — the declaring file's node owning none of them — which every
+    single-definition god-node guard downstream then reads as an ambiguity and bails on.
+    This is the Go analogue of ``_merge_decl_def_classes`` and runs at the same point,
+    before disambiguation.
+
+    GOD-NODE GUARDS:
+
+      * Every node in the id-collision must be a ``.go`` file in ONE directory — Go's
+        package unit. The id folds in only the directory's NAME, so ``a/svc`` and
+        ``b/svc`` collide on id while being different packages; those stay split.
+      * Folding stops at a ``_test.go`` member, whose package clause may be the separate
+        ``svc_test`` package declaring a type of its own name. The clause is not parsed,
+        so same-directory is not evidence of same package there.
+
+    No edge re-pointing is needed: the group already shares one id, so every edge
+    already points at the survivor. The declaration site wins — the file holding
+    ``type X``, the only one with a ``contains`` edge to it — and the dropped nodes'
+    methods keep their own ``source_file``, so no location is lost with them.
+    """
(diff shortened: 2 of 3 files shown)


xiongjianxu:
The push was an author-identity rewrite (my commits carried an email not linked to my GitHub account); the tree is byte-identical to the previously reviewed head.

On the finding that the package-level fold **drops distinct build-tagged declarations**: it does not. Two files in one package, each with a `//go:build` tag and each declaring `type Client struct` plus one method, extract to a single folded `Client` node that keeps both:

```
nodes: Client @client_linux.go
edges: Client --contains--> .Open()   (client_linux.go)
       Client --contains--> .Close()  (client_darwin.go)
```

Both methods and both `contains` edges survive; the only thing collapsed is the scalar `source_file` attribute, which can name one file only. That is also the semantically right answer: build tags are mutually exclusive, so at most one variant is ever compiled for a given target, and a call to `c.Open()` in that package resolves to the one `Client` the build selected. Splitting the node per file would instead produce two `Client` declarations in one package, which is what trips the `len(type_defs) != 1` ambiguity guard and would silently *lose* the edges this PR is adding.

PR #3402, opened 2026-09-07
fix(extract): remove quadratic scan in TS import-type normalization (#3359)

### LINKED ISSUE #3359: BUG
**Title: `extract --code-only` hangs forever on large monorepo (no graph.json written); worker spins 100% in `re.sub` — reproduced on 0.9.53 and 0.9.54**

**Environment**
- macOS 15.7.7 (Apple Silicon), Python 3.11.15 (uv-managed), graphifyy 0.9.53 and 0.9.54 (both reproduced)
- Corpus: DeepSeek Harness monorepo (dsh) v0.1.3-alpha.1, ~2,470 code files after excluding tests/snapshots via `.graphifyignore` (`**/tests/`, `**/*.spec.ts`, `**/*.test.ts`, `**/*.e2e.ts`, `snapshots/`). Total repo incl. tests ≈ 4,037 files.

**Steps to reproduce**
1. Add the `.graphifyignore` exclusions above in repo root.
2. `graphify extract . --code-only` (default workers)
   → AST extraction progress freezes: parallel runs stall at ~97–99% (2400/2470; inside `packages/` alone 3300/3362; also 3400/3415) and **`graph.json` is never written**.
3. Variants all reproduce: `--max-workers 1/2/4/10`, `--no-dedup`, 0.9.53 and 0.9.54.
4. Serial run (`--max-workers 1`) freezes earlier at 500/2469 (20%), also indefinitely.
5. Repo subsets that are small (each top-level dir except `packages/`) complete fine; a ~60-file repo (xiaohongshu-mcp) completes fine — so not a universal scale issue.

**Diagnostics**
- While frozen: the active worker shows **100% CPU and opens NO files** (`lsof` lists only the python binary).
- `sample` (macOS) of the spinning worker shows 2485/2485 samples inside **`_sre_SRE_Pattern_sub` → `sre_search` → `sre_ucs1_match`** — i.e., a Python `re.sub` operating on a large UCS-1 (ASCII) string is the hot path. No traceback; process exits only when killed, with `resource_tracker` semaphore-leak warnings at teardown.
- Freeze point shifts with file set → looks **content-triggered**: likely a regex normalization/sanitization pass over large serialized per-chunk/per-file payloads that is pathologically slow (or catastrophic backtracking) for some content in this corpus.

**Request**
Could you check the post-parse regex normalization paths (e.g., anything running `re.sub` over chunk/file payloads) for patterns that are linearity-unsafe on large ASCII inputs? Happy to provide more logs/cache entries on request. This blocks `graphify extract` on moderately large TypeScript monorepos.

**Notes**
- I did not retain `graphify-out/cache` (deleted between retries); can keep it on a repro run if useful.
- Corpus is a public GitHub repo: `deepseek-ai/deepseek-harness` (tag v0.1.3-alpha.1).

Sagexd08:
Fixes #3359.

## Root cause

Not a linearity-unsafe regex. Every `re.sub`/pattern in the extract path is linear — I benchmarked `textwrap.shorten`, `_TS_IMPORT_CALL_RE`, `_VUE_SCRIPT_RE` and the JS rescue patterns and none backtrack. The `_sre` frames in the reporter's `sample` output are the innermost frames of a quadratic **caller**.

`_normalize_ts_import_types` (the #3154 TypeScript `import(...)` normalizer) filters each regex match against every tree-sitter `type_arguments` byte range with a nested linear scan, in both passes:

```python
matches = [m for m in matches
           if not any(start <= m.start() < end for start, end in original_type_ranges)]
```

That is **O(matches × ranges)**, and `m.start()` is re-evaluated inside `any()` for every range — hence `re.Match.start` dominating profiles and `sample` appearing to indict a regex.

Both counts scale with file size in TypeScript that mixes generic calls (`fn<A, B>(x)` → one `type_arguments` range each) with `import(...)` types — the shape of generated and barrel modules in a large monorepo.

Measured on the filter alone:

| file | `import()` sites | before | after |
|---|---|---|---|
| 0.8 MB | 10,000 | ~21 s | 0.003 s |
| 2.1 MB | 25,000 | ~2.6 min | 0.008 s |
| 4.2 MB | 50,000 | ~11.8 min | 0.019 s |

This accounts for every symptom in the report:

- **100% CPU with no open files** — all work is in-memory after the read.
- **Freeze point shifts with the file set** — content-triggered by a few large generated files, not by file count. Explains why `packages/` hangs while other top-level dirs and a 60-file repo finish.
- **`--max-workers 1` stalls *earlier*** (500/2469 vs ~2400/2470) — one worker reaches the pathological file sooner instead of ten racing past it first.
- **`--no-dedup`, 0.9.53 and 0.9.54 all reproduce** — the code is in the extract path and unchanged across those releases.

## Fix

Replace both scans with a binary search over start-sorted ranges plus a prefix-maximum of `end` values, so a lookup touches nesting depth instead of every range.

Type-argument ranges nest but never partially overlap, so the prefix maximum is a correct stop condition. The subtle part: a **closed inner sibling** must not terminate the leftward walk — a naive running-max break gets this wrong and silently drops enclosing ranges.

Normalization output is unchanged: byte-length and line-count preserving, identical masking decisions, and the runtime-vs-type disambiguation from 6baa5e0 still holds.

Also restores the emptiness guard in the second pass — the range list is now an index tuple, which is always truthy, so it checks the spans it holds rather than the container.

## Verification

- **500,000-case differential test** of the new lookup against the old predicate, including non-nested random ranges: zero mismatches.
- `_normalize_ts_import_types` scaling is now ~2× per doubling (was ~4×), linear out to 32k imports.
- Invariants re-checked: byte length, line count, TSX path, comparison-vs-generic ambiguity, comment/string literals untouched.
- New regression test asserts on **scaling** (2× input must not ~4× the time) rather than wall-clock, so it is not machine-dependent. It fails at ~3.4× before this change and passes after.

### Full-suite note

A whole-suite comparison against unmodified `c9f9901` shows 53 vs 61 failures, but the delta is environment flakiness on Windows, not this change:

- The diff lists 13 "new" failures **and 5 that disappeared** — a change confined to the TS normalizer cannot fix unrelated install tests.
- Every failure is in `serve_http` / `install*` / `hooks` / `watch`. None is in `extract`, `resolution`, or any TypeScript path, and no failing test references `_normalize_ts_import_types`, `_ts_ranges_containing`, `_ts_build_range_index`, or `_ts_type_argument_ranges`.
- The install tests are order- and state-dependent: the **same baseline code** produced 8 failures on one run then 3, 3, 3 on the next three, with no edits at all.
- Run head-to-head three times each on the same files, baseline and this branch are identical: 3 failed / 155 passed on every run.
- The 6 install/upgrade tests flagged as regressions all pass on this branch in isolation.

### DIFF
--- a/graphify/extract.py
+++ b/graphify/extract.py
@@ -8,6 +8,7 @@ import os
 import re
 import sys
 import textwrap
+from bisect import bisect_right
 from collections import Counter
 from dataclasses import dataclass, field
 from pathlib import Path, PurePath
@@ -1360,6 +1361,58 @@ def _ts_type_argument_ranges(root: Any, *, call_only: bool) -> list[tuple[int, i
     return ranges
 
 
+# Sorted type-argument ranges prepared for lookup: (starts, spans, prefix_max_end).
+# See _ts_build_range_index.
+_TsRangeIndex = "tuple[list[int], list[tuple[int, int]], list[int]]"
+
+
+def _ts_ranges_containing(
+    ranges: "_TsRangeIndex", offset: int
+) -> "list[tuple[int, int]]":
+    """Return the ranges in ``ranges`` that contain ``offset``.
+
+    Scanning the full range list for every match made
+    :func:`_normalize_ts_import_types` O(matches x ranges); on a large monorepo
+    file with thousands of generic calls and thousands of ``import(...)`` types
+    that quadratic blowup pinned a worker at 100% CPU with no output and no file
+    I/O, so extraction never finished (#3359).
+
+    Every range containing ``offset`` must start at or before it, so binary-search
+    to the last such start and walk left. A range that ends before ``offset`` is
+    not itself a hit but must not stop the walk -- a closed sibling can sit inside
+    an enclosing range that does contain ``offset``. The index's running maximum of
+    ends (``prefix_max_end``) gives the correct stop: once no range at or before
+    this position reaches past ``offset``, none of the remaining ones can.
+    """
+    starts, spans, prefix_max_end = ranges
+    hits: list[tuple[int, int]] = []
+    index = bisect_right(starts, offset) - 1
+    while index >= 0 and prefix_max_end[index] > offset:
+        start, end = spans[index]
+        if end > offset:
+            hits.append((start, end))
+        index -= 1
+    hits.reverse()
+    return hits
+
+
+def _ts_build_range_index(root: Any, *, call_only: bool) -> "_TsRangeIndex":
+    """Index :func:`_ts_type_argument_ranges` for :func:`_ts_ranges_containing`.
+
+    Returns the ranges sorted by start, split into a bare ``starts`` list for
+    :func:`bisect_right`, plus a prefix-maximum of ``end`` values so a lookup can
+    tell when no earlier range can still reach a given offset.
+    """
+    spans = sorted(_ts_type_argument_ranges(root, call_only=call_only))
+    starts = [start for start, _ in spans]
+    prefix_max_end: list[int] = []
+    running = -1
+    for _, end in spans:
+        running = max(running, end)
+        prefix_max_end.append(running)
+    return starts, spans, prefix_max_end
+
+
 def _ts_error_nodes(root: Any) -> list[Any]:
     """Return parser error nodes without depending on a grammar's error name."""
     errors: list[Any] = []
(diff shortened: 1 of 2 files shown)


Sagexd08:
Good catch — fixed in d902820.

`_TsRangeIndex` was a quoted string, making it a `str` *value* rather than a type. mypy confirms 4 errors on the original:

```
extract.py:1370: error: Variable "graphify.extract._TsRangeIndex" is not valid as a type  [valid-type]
extract.py:1387: error: _TsRangeIndex? has no attribute "__iter__" (not iterable)  [attr-defined]
extract.py:1399: error: Variable "graphify.extract._TsRangeIndex" is not valid as a type  [valid-type]
extract.py:1536: error: Value of type _TsRangeIndex? is not indexable  [index]
```

The last two are cascading: the tuple loses its type, so the unpacking in `_ts_ranges_containing` and the `[1]` index in `_normalize_ts_import_types` go unchecked.

Worth noting it did not fail at runtime — `typing.get_type_hints` evaluates the annotation string and then the alias string again, so the double evaluation accidentally resolved it. That masked the problem rather than making it benign.

Unquoted the right-hand side. The module already has `from __future__ import annotations`, so the annotation sites needed no quoting either. All 4 errors clear, no behavior change: 19/19 TS tests pass, 375k-case differential re-check has zero mismatches, and scaling is still ~2x per doubling.

safishamsi:
Shipped in **v0.9.56** — landed on `v8` via cherry-pick with your authorship preserved. Thanks @Sagexd08 — TS import-type normalization no longer does an O(matches×ranges) scan (byte-identical output). Release: https://github.com/Graphify-Labs/graphify/releases/tag/v0.9.56

Issue #3405, opened 2026-09-08
Python: module.attr() calls unresolved, and nested functions resolved to unrelated files as EXTRACTED/1.0


jerryjohn1702:
Two Python call-resolution defects, found while indexing a 602-file polyglot monorepo with v0.9.56. They are separate bugs but share a root cause: call targets appear to be matched on bare name against a global symbol table, without consulting the importing module's namespace or the enclosing lexical scope.

The second is the more serious. It produces **wrong edges labelled `EXTRACTED` with `confidence_score: 1.0`**, which is worse than a missing edge, because nothing downstream can distinguish it from a real one.

Corpus: 602 code files → 3477 nodes, 6221 edges. `graphify extract . --code-only` (no LLM), Python 3.14, Windows. Paths below are anonymised; the shapes are verbatim.

---

## 1. `import module` + `module.func()` calls are never resolved

`from mod import func` → `func()` resolves correctly. `import mod` → `mod.func()` produces **no edge at all**.

```python
# pkg_a/collect.py
def q(conn, sql): ...      # returns tuples
def qd(conn, sql): ...     # returns dicts

# pkg_a/store.py
import collect
rows = collect.qd(conn, sql)          # line 119

# pkg_a/waker.py
import collect
rows = collect.q(conn, SIGNALS_SQL)   # line 103
```

`graphify explain` on `qd()` reports degree 2 — its `contains` edge and a docstring `rationale_for` edge. The call from `store.py` is absent. `q()` shows only its one same-file caller and misses both cross-module ones.

This is not a general cross-file failure: **331 of 2476 `calls` edges (13%) are cross-file and correct.** The split is exactly import style. In this corpus one package uses `from x import y` throughout and its call graph is good; a second uses `import x` + `x.y()` and its cross-module edges are essentially all missing.

Why it is easy to miss: "what calls this?" is the question the graph exists to answer, and a package where resolution fails returns *nothing* rather than an error. A function with no callers looks like a leaf, not like a gap.

## 2. Nested functions resolve to same-named module-level functions in unrelated files

This one manufactures edges that do not exist.

```python
# pkg_b/probe.py
def trace(...):
    def walk(n, d=0):     # line 53 — nested, local to trace()
        ...
    walk(root)
```

There is exactly one *module-level* `walk` in the corpus, in an unrelated tree:

```python
# pkg_c/overlay.py
def walk(node, path=""):   # line 74
```

[...]

PR #3410, opened 2026-09-08
Fix Python nested function call resolution

hopstreax:
## Summary

Fixes Python nested-function extraction and lexical call resolution for issue #3405.

### Changes

- Extract nested Python functions as proper Graphify function nodes.
- Add `contains` edges between parent and nested functions.
- Resolve unqualified calls using lexical scope.
- Correctly handle nested-function shadowing of module-level functions.
- Support nested sibling calls, recursion, async nested functions, and deeper nesting.
- Prevent local non-callable variables from being incorrectly emitted as unresolved calls.

### Tests

Added coverage for:

- Nested function extraction and node IDs
- Parent → nested `contains` relationships
- Outer → nested function calls
- Lexical shadowing
- Nested sibling calls
- Recursive nested functions
- Async nested functions
- Deeply nested functions
- Local non-callable suppression

Existing Python, confidence-rubric, and JavaScript nested-function tests also pass.

Issue: #3405

### DIFF
--- a/graphify/extractors/engine.py
+++ b/graphify/extractors/engine.py
@@ -2161,6 +2161,75 @@ def _scan_js_nested_function_declarations(
             )
 
 
+def _scan_python_nested_function_declarations(
+    container_node, parent_nid: str, *, source: bytes, config,
+    add_node, add_edge, callable_def_nids: set | None,
+    local_bound_names: dict | None, function_bodies: list,
+    scope_parents: dict[str, str] | None = None,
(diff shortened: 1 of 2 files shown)


safishamsi:
Shipped in **v0.9.58** — landed on `v8` via cherry-pick with your authorship preserved. Thanks @hopstreax — a call to a Python function nested inside another now resolves to that inner definition per lexical scope. Release: https://github.com/Graphify-Labs/graphify/releases/tag/v0.9.58

Issue #729, opened 2026-05-05
Feature Request: Local-Model Extraction + Cross-Session Persistent Memory


odunjodavid:
Graphify solves the "re-reading files" problem, but the semantic extraction pass still outsources to the main assistant's LLM. This means: (1) The token burden of building the graph burns through the context window, (2) Every new session starts with no memory of past context, and (3) The graph is project-scoped.

### The Idea

A local-first extraction backend combined with persistent cross-session memory layered onto the graph.

**Should:**
- Allow Graphify to use its own lightweight model (e.g. via Ollama, Qwen2.5-Coder, Phi-3-mini, any locally-running model etc.) as the semantic extraction provider, instead of inheriting the main assistant's API
- Persist the graph across sessions as a living memory store as an accumulation of semantic context over time
- Let the graph grow across projects, so architecture & structural patterns, bug patterns, lessons from bug fixes, and domain knowledge compound

### This Matters Because

Right now, the architecture is: **main LLM → farms context → graph → main LLM queries graph**

Would like it to be: **local model → builds and maintains graph → main LLM queries compressed output only**

Ultimately, the main assistant shouldn't do extraction work. It should only receive structured, already-digested output from a locally-maintained knowledge layer.

### Tools That Exist

- Ollama makes local semantic extraction viable at no API cost
- SynaBun (https://github.com/danilokhury/Synabun) has solved cross-session persistence with vector memory but without graph structure

### In short

Please add a persistent, locally-maintained knowledge layer that any assistant can query via MCP.

safishamsi:
Implemented in v0.7.7 (commit 48888a7):

- Ollama backend: `graphify extract ./docs --backend ollama` (auto-detected via `OLLAMA_BASE_URL`; defaults to `qwen2.5-coder:7b`; $0.00 cost)
- Cross-project global graph at `~/.graphify/global.json`: `graphify global add/remove/list/path`
- `graphify extract --global --as <tag>` to register in one step
- Node IDs prefixed `<repo>::<id>` to prevent silent collisions across projects
- `merge-graphs` now prefix-relabels inputs before composing
- 21 new tests covering all new code paths

PR #3415, opened 2026-09-08
fix(global-graph): keep each repo's external stubs so global add matches merge-graphs

### LINKED ISSUE #3414: global add and merge-graphs produce different graphs from the same inputs
## Summary

`global add` and `merge-graphs` produce different graphs from the same inputs.

`global_add` merges nodes that have no `source_file` (external-library stubs) by label
(`graphify/global_graph.py`), skipping the incoming duplicate and rewiring its edges onto the
copy already in the graph. `merge-graphs` (`graphify/cli.py`) composes the prefixed units with
`nx.compose` and does no label dedupe at all.

That contradicts the comment `global_add` carries about its own contract (added in #3152):

> adding repos one at a time lands where a single `merge-graphs` of the same inputs would.

## Reproduction (on `v8`, no plugins, two two-node graphs)

`repoA` and `repoB` each hold one real module node plus a sourceless stub labelled `message`,
with an in-unit edge to it.

```
global add repoA; global add repoB
  nodes: repoA::moda, repoA::message, repoB::modb          <- repoB::message never added
  edges: (repoA::moda -> repoA::message), (repoB::modb -> repoA::message)
                                           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ cross-repo edge
                                                                         that no unit contains
merge-graphs repoA repoB
  nodes: repoA::moda, repoA::message, repoB::modb, repoB::message
  edges: (repoA::moda -> repoA::message), (repoB::modb -> repoB::message)
```

Three consequences, all reproduced:

1. **Fabricated cross-repo edges.** The surviving stub keeps the first-merged repo's `repo` tag,
   so every other repo's edge into it reads as cross-repo. `repo` is written unconditionally onto
   every node by `prefix_graph_for_global` (`graphify/build.py`), stubs included, so stub
   ownership is a function of merge order — nothing about the code.

2. **Real edges disappear on re-scan.** `prune_repo_from_graph` removes every node whose
   `repo` equals the tag being rebuilt. Re-scanning the repo that happens to own the stub takes
   the stub with it, and every other repo's `imports`/`references` edge into it goes too.
   Observed: `LOST: [('repoA::message', 'repoB::modb', 'imports')]`. The manifest's per-repo
   `edge_count` only counts that repo's own edges, so the loss does not show up in the manifest.

3. **`node_count` under-reported.** `added = prefixed.number_of_nodes() - len(remap)` counts the
   skipped duplicates as "not added", but they were already in the graph under another repo's
   tag, so the manifest attributes them to the wrong repo and each later repo reports fewer
   nodes than it has.

## Where it came from

- #729 (`48888a7`) introduced the label dedupe for external nodes.
- #1250 (`444d73e`, "fix(global-graph): rewire edges to deduplicated external nodes") added the
  remap rewire, so the dedupe stopped dropping edges — the symptom — while ownership by merge
  order stayed.

#1250 fixed a symptom of #729. Removing the dedupe removes both.

## Proposed fix

Compose each unit's stubs as-is, matching `merge-graphs`. PR follows.

The cost is explicit: #729's purpose was one node per external library across all repos, so
"who uses library X" could be answered by node id. With per-repo stubs that query has to group
by label. In exchange the two merge paths agree, cross-repo edges mean what they say, and a
re-scan cannot delete another repo's edges.

xiongjianxu:
Fixes #3414

## What changed

`global_add` no longer merges sourceless external stubs by label. Each unit's stubs compose in
as its own nodes, which is what `merge-graphs` has always done, so the incremental path and the
one-shot path now land on the same graph.

`graphify/global_graph.py`:

- Dropped the `external_labels` / `remap` block and the rewire that consumed it.
- Dropped the `if u != v` self-loop guard with it — the only self-loops it ever caught were the
  ones the remap created (both endpoints of an in-unit edge remapping to the same survivor).
- `added = prefixed.number_of_nodes()` instead of `- len(remap)`, so each repo's manifest
  `node_count` reports what it actually contributed.

## Why the dedupe cannot stay

`prefix_graph_for_global` writes `repo=<tag>` onto every node in a unit, stubs included. When the
dedupe keeps one stub per label, that stub's `repo` is whichever repo merged first — so:

- every other repo's edge into it reads as a **cross-repo edge that no unit contains**;
- `prune_repo_from_graph` (which removes all nodes with `repo == tag`) deletes the stub when its
  owner is re-scanned, and every other repo's real `imports`/`references` edge into it dies with
  it. The manifest's per-repo `edge_count` only counts that repo's own edges, so nothing in the
  manifest shows the loss.

Both are reproduced in the issue.

## Tests

`tests/test_global_graph.py`, four tests, each failing before this change and passing after
(4 failed / 19 passed → 23 passed):

- `test_global_add_matches_merge_graphs_on_the_same_inputs` — the contract itself: two
  `global_add` calls vs one `merge-graphs` of the same two units, equal node sets and equal edge
  sets.
- `test_global_add_keeps_stubs_per_repo` — both `repoA::message` and `repoB::message` present,
  both in-unit edges present, no edge whose endpoints' `repo` differ, and
  `manifest["repos"]["repoA"]["node_count"] == 2`.
- `test_global_add_rescan_preserves_other_repo_edges` — re-adding `repoA` leaves repoB's edge set
  byte-identical.
- `test_global_add_keeps_externals_per_repo` — this is the **rewritten**
  `test_global_add_rewires_edges_to_deduplicated_externals`. It asserted
  `"repoB::requests" not in G.nodes` and `has_edge("repoB::modb", "repoA::requests")`, i.e. it
  encoded merge-order ownership as intended behaviour. Reversed and renamed.

Full suite: 5354 passed / 94 skipped. `tests/test_ts_import_type_arguments.py::test_ts_normalizer_scales_linearly_on_large_files`
fails identically on unmodified `v8` in a full run and passes in isolation — it is a
`process_time` ratio assertion, and its own docstring notes the ratio is machine-dependent.
`ruff check` clean.

## Downstream note

#729's purpose was one node per external library across all repos, so "who uses library X" could
be answered by node id. Queries that relied on that need to group by label instead. That is the
price of the two merge paths agreeing.

### DIFF
--- a/graphify/global_graph.py
+++ b/graphify/global_graph.py
@@ -122,30 +122,16 @@ def global_add(source_path: Path, repo_tag: str) -> dict:
     G = _load_global_graph()
     removed = prune_repo_from_graph(G, repo_tag)
 
-    # Merge external-library nodes (no source_file) by label to avoid duplication
-    external_labels = {
-        d.get("label", ""): n
-        for n, d in G.nodes(data=True)
-        if not d.get("source_file") and d.get("label")
-    }
-    # Map each deduplicated external onto the existing global node so that
-    # edges incident to it can be rewired instead of dropped.
-    remap = {}
+    # Each unit keeps its own external-library stubs (no source_file). Merging
+    # them by label makes stub ownership a function of merge order: the survivor
+    # carries the first-merged repo's tag, fabricating cross-repo edges and
+    # letting prune_repo_from_graph delete other repos' real edges with it.
     for node, data in prefixed.nodes(data=True):
-        if not data.get("source_file") and data.get("label") in external_labels:
-            remap[node] = external_labels[data["label"]]
-
-    # Compose: add prefixed nodes (except deduplicated externals) into global graph
-    for node, data in prefixed.nodes(data=True):
-        if node not in remap:
-            G.add_node(node, **data)
+        G.add_node(node, **data)
     for u, v, data in prefixed.edges(data=True):
-        u = remap.get(u, u)
-        v = remap.get(v, v)
-        if u != v:  # don't introduce self-loops via remapping
-            G.add_edge(u, v, **data)
+        G.add_edge(u, v, **data)
 
-    added = prefixed.number_of_nodes() - len(remap)
+    added = prefixed.number_of_nodes()
     # A member call parked on a caller node (#3152) may be answered by a repo
     # already in the global graph, or by this one for a repo added earlier. The
     # pass recomputes its own output, so adding repos one at a time lands where a
--- a/tests/test_global_graph.py
+++ b/tests/test_global_graph.py
@@ -198,6 +198,122 @@ def test_global_add_two_repos_no_collision(tmp_path):
     assert G.number_of_nodes() == 2  # no silent merge
 
 
+def _shared_stub_repos(tmp_path, extra_a=None):
+    """Two repos whose real code both reference a `message` stub (no source_file)."""
+    nodes_a = [
+        {"id": "a_service", "label": "AService", "source_file": "src/a.java"},
+        {"id": "message", "label": "message"},
+    ] + (extra_a or [])
+    nodes_b = [
+        {"id": "b_service", "label": "BService", "source_file": "src/b.java"},
+        {"id": "message", "label": "message"},
+    ]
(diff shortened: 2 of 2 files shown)


xiongjianxu:

PR #3418, opened 2026-09-08
fix(bash): resolve exec-position script paths built from a variable

### LINKED ISSUE #3416: Bash extractor: exec-position paths built from a variable (`"$script_dir/x.sh"`) produce no cross-file edge — the #1756 fix only resolves literal paths
## Summary

The #1756 fix (shipped 0.9.13, `context="script_invocation"`) resolves an
exec-position script path **literally**, so it only fires when the command name is
already a plain path — `./b.sh`, `/abs/b.sh`, or `bash b.sh`. Any exec-position path
built from a variable produces no edge, and `"$script_dir/x.sh"` is the ordinary way
a shell script runs a sibling.

The cause is an asymmetry between the two branches: the `source` branch reads its
path argument with `_read_text(...)`, which preserves `$`, while the exec branch
relies on `cmd = literal(cmd_name_node)` — and `literal()` returns `None` for
anything containing `$` by design. So the exec branch is never entered for a
variable-built path, and the `_bash_source_suffix` / `_BASH_LEADING_EXPANSION`
machinery that already exists for `source` (#2079, #2596) is never applied to it.

Separately, a `source` path whose leading segment is a full `$(cd … && pwd)`
substitution is also missed, because `_BASH_LEADING_EXPANSION` matches `${VAR}` and
`$VAR` but not `$(...)`.

## Version

Reproduced on **0.9.55**. Python 3.12.3, tree-sitter-bash 0.25.1.

## Reproduction

Five files in an empty directory:

```bash
# b.sh
#!/bin/bash
echo hi

# f1_literal.sh
#!/bin/bash
./b.sh

# f2_inline_var.sh
#!/bin/bash
script_dir="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
"$script_dir/b.sh" --go

# f3_assigned_var.sh
#!/bin/bash
script_dir="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
helper="$script_dir/b.sh"
"$helper" --go

# f4_source_cmdsub.sh
#!/bin/bash
. "$(cd -P -- "$(dirname -- "${BASH_SOURCE[0]}")" && pwd)/b.sh"
```

```python
from pathlib import Path
from graphify.extractors.bash import extract_bash

for f in ["f1_literal.sh", "f2_inline_var.sh", "f3_assigned_var.sh", "f4_source_cmdsub.sh"]:
    r = extract_bash(Path(f))
    cross = [e for e in r["edges"]
             if e["relation"] in ("calls", "imports_from", "imports")
             and "b_sh" in str(e.get("target", "")) + str(e.get("target_file", ""))]
    print(f, len(cross), [e["relation"] for e in cross])
```

## Actual

```
f1_literal.sh        1 ['calls']
f2_inline_var.sh     0 []
f3_assigned_var.sh   0 []
f4_source_cmdsub.sh  0 []
```

## Expected

All four should yield a cross-file edge to `b.sh`. f1 already does.

## Cause

`graphify/extractors/bash.py:477` (exec branch):

```python
raw = cmd if cmd.endswith(".sh") else None
if cmd in _BASH_SCRIPT_RUNNERS and args:
    raw = literal(args[0])
if raw and raw.endswith(".sh"):
    resolved = (path.parent / raw).resolve()
```

`cmd` comes from `literal(cmd_name_node)` at :311, and `literal()` at :223 rejects
any candidate containing `$`:

```python
if any(token in raw for token in ("$", "`", "$(", "<(", ">", "|", ";", "&")):
    return None
```

That filter is correct for its purpose — it is a *name* filter — but it means the
exec branch cannot see a path. Contrast the `source` branch at :320, which reads the
path directly and therefore keeps `$` available for `_bash_source_suffix`:

```python
raw = _read_text(args[0], source).strip().strip("'\"")
```

For f4, `_BASH_LEADING_EXPANSION` (:16) matches only `${VAR}` / `$VAR` segments, so
`$(cd -P -- … && pwd)/b.sh` still contains `$` after substitution and
`_bash_source_suffix` returns `None`.

## Suggested fix

For f2, reading the command name as a path when `literal()` declines, then reusing
the existing suffix helper, is enough. Tested patch:

```python
elif cmd not in defined_functions:
    raw = cmd if (cmd and cmd.endswith(".sh")) else None
    if cmd and cmd in _BASH_SCRIPT_RUNNERS and args:
        raw = literal(args[0])
    if raw is None and cmd_name_node is not None:
        # The command name is an expansion, e.g. `"$script_dir/x.sh"`.
        # literal() rejects any "$" by design, so read the path the way the
        # `source` branch does and strip the leading ${VAR}/ segments.
        cand = _read_text(cmd_name_node, source).strip().strip("'\"")
        if cand.endswith(".sh") and "$" in cand:
            raw = _bash_source_suffix(cand) or None
    if raw and raw.endswith(".sh"):
        ...
```

The existing `resolved.is_file()` guard is untouched, so the B-1 traversal
protection still holds — an edge is only emitted for a file present in the tree.

With that patch f2 emits `calls` / `script_invocation` and f1 is unchanged. f3 and
f4 still need work and I have not attempted patches for them:

- **f3** needs the variable-assignment tracking that #2172 already built for
  `source` (`var_bases`) extended to exec position.
- **f4** needs `_BASH_LEADING_EXPANSION` to also consume a leading `$(...)`
  segment, which needs one level of paren balancing.

One caveat on the patch, in case it matters to how you take it: on a real corpus a
few inline exec-position sites still produced no edge that I could not account for
— possibly `is_inside_expansion` rejection for calls nested in substitutions. I did
not chase it, so treat the patch as covering the canonical form rather than as
complete.

## Impact

Measured on a private shell-heavy repo, 794 files, ~500 shell scripts. Counting real
cross-file dependencies by hand and comparing against `graph.json`:

- **417 of 528 (78%) missing** on 0.9.55.
- The dominant idiom is `name="$script_dir/x.sh"` (242 sites, f3) followed by inline
  exec position (124 sites, f2). `source`/`.` accounts for a small minority.
- The knock-on effect is worse than the missing edges: `graphify affected <file> 3`
  returned **the same set as depth 1** for shell targets, because the intermediate
  scripts are only ever invoked by path. Transitive reachability is effectively
  unavailable for a shell codebase.
- `graphify coupling` reported a 3,139-line dashboard script as reaching **0**
  external files when it reaches 3 and is invoked by 5.

Applying the f2 patch above to that corpus added 10 `script_invocation` edges, all
of which I verified by reading the cited line — no false positives. So f2 alone is a
small win; **f3 is where the volume is.**

edwardselby:
Fixes #3416.

## The problem

`script_invocation` edges (#1756) resolve a path only when it is a bare literal — `./helpers.sh`, or `bash ./helpers.sh`. Both routes into that branch go through `literal()`, which rejects by design any token containing `$`. So the most common way one shell script calls a sibling never produces an edge:

```bash
script_dir="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
"$script_dir/helpers.sh" --flag
```

`literal(cmd_name_node)` returns `None`, the `elif cmd and cmd not in defined_functions` guard short-circuits on the `None`, and the branch that would have resolved the path never runs.

On a 206-file shell tree that uses the `script_dir` idiom throughout, `script_invocation` emitted **zero** edges. For that whole style of codebase the feature is effectively dead — and it's a common style, because it's the idiom that makes a script runnable from any working directory.

## The fix

The `source` branch already solved exactly this problem for `source "$DIR/lib/x.sh"` in #2079: strip the leading `${VAR}/` segments with `_bash_source_suffix` and resolve the literal remainder against the sourcing file's own directory, on the grounds that the canonical `script_dir` idiom makes that variable equal to precisely that directory.

This applies the identical treatment to the exec position — read the command-name node's raw text instead of its `literal()`, strip the same way, and fall through to the `resolved.is_file()` gate that was already there.

15 lines, one file, no new helpers, no new dependency, no change to any existing edge.

## Nothing new is fabricated

Every existing guard still applies, and the reuse of `_bash_source_suffix` brings its guards along:

- a remainder that still holds an expansion (`"$dir/$name.sh"`) returns `None` — no edge
- a `..` segment returns `None` (`allow_dotdot=False`, the script-dir-guess default) — no traversal outside the tree
- `resolved.is_file()` still gates emission, so a wrong `script_dir` guess emits nothing rather than a dangling edge

The three existing negative tests pass unchanged: `test_extract_bash_skips_missing_and_shadowed_script_invocations`, `test_extract_bash_skips_dynamic_script_invocation`, and `test_extract_bash_source_via_variable_path_no_match_emits_no_dead_edge`.

The known limitation is inherited, not introduced: the heuristic assumes the leading variable is the script's own directory. When it isn't, an edge appears only if a same-named file happens to sit beside the caller. That is the same assumption #2079 already ships for `source`.

## Tests

Five new tests in `tests/test_extract.py`, placed with and modelled on the existing `script_invocation` block:

| test | asserts |
|---|---|
| `test_extract_bash_script_invocation_via_variable_built_path` | parametrised over `"$script_dir/helpers.sh"`, `"${script_dir}/helpers.sh"`, the same with arguments, and unquoted — full edge dict equality, same shape as `test_extract_bash_emits_script_invocation_calls` |
| `test_extract_bash_variable_built_invocation_with_dynamic_stem_emits_no_edge` | `"$script_dir/$name.sh"` → no edge |
| `test_extract_bash_variable_built_invocation_no_on_disk_match_emits_no_edge` | target absent → no edge |
| `test_extract_bash_variable_built_invocation_targets_existing_entrypoint` | through `extract()`: both endpoints are real nodes, not dangling ids |
| `test_extract_bash_attributes_variable_built_invocation_to_function` | invocation inside a function body is attributed to that function, not the file entrypoint |

All six positive assertions fail on `v8` as it stands and pass with this change; the two negatives pass either way, which is the point.

## Verification

- `pytest tests/test_extract.py -k bash` — 57 passed (was 49; +8 new cases)
- `pytest tests/test_extract.py` — 222 passed, 4 skipped
- full `pytest tests` — 5098 passed, 255 skipped, 25 failed; the identical 25 fail on unmodified `v8` in this environment (terraform, ollama, skillgen — all unrelated to this change), so **no regressions**
- real corpus, 206 shell files: **0 → 16** `script_invocation` edges across 10 distinct file pairs. All 16 hand-checked back to a genuine exec-position invocation of a file that exists. No false positives.

One of the 16 is worth calling out because it looks like it should be wrong and isn't: a test harness does `cp "$repo_root/tests/run_smokes.sh" "$failure_dir/run_smokes.sh"` and then invokes `"$failure_dir/run_smokes.sh"`. `$failure_dir` is a temp directory, not the script's own directory — but the file it runs is a copy of the very sibling the heuristic resolved to, so the structural answer is right for the right reason.

### DIFF
--- a/graphify/extractors/bash.py
+++ b/graphify/extractors/bash.py
@@ -473,10 +473,25 @@ def extract_bash(path: Path) -> dict:
                                 if tgt_nid:
                                     add_edge(file_nid, tgt_nid, "imports", line,
                                              context="import")
-                elif cmd and cmd not in defined_functions:
-                    raw = cmd if cmd.endswith(".sh") else None
-                    if cmd in _BASH_SCRIPT_RUNNERS and args:
+                elif cmd not in defined_functions:
+                    raw = cmd if (cmd and cmd.endswith(".sh")) else None
+                    if cmd and cmd in _BASH_SCRIPT_RUNNERS and args:
                         raw = literal(args[0])
+                    if raw is None:
+                        # The command name is itself an expansion, e.g.
+                        # `"$script_dir/x.sh" --flag`. `literal()` rejects any
+                        # candidate containing "$", so `cmd` is None and neither
+                        # of the two branches above can see the path. Read the
+                        # node's raw text the way the `source` branch does and
+                        # strip the leading ${VAR}/ segments -- the same
+                        # script-dir assumption already used for `source
+                        # "$DIR/lib/x.sh"` (#2079). `_bash_source_suffix`
+                        # rejects a remainder that still holds an expansion or
+                        # a `..` segment, and the `resolved.is_file()` gate
+                        # below means a wrong script-dir guess emits nothing.
+                        cand = _read_text(cmd_name_node, source).strip().strip("'\"")
+                        if cand.endswith(".sh") and "$" in cand:
+                            raw = _bash_source_suffix(cand) or None
                     if raw and raw.endswith(".sh"):
                         resolved = (path.parent / raw).resolve()
                         if resolved.is_file():
--- a/tests/test_extract.py
+++ b/tests/test_extract.py
@@ -2614,6 +2614,110 @@ def test_extract_bash_attributes_script_invocation_to_function(tmp_path):
     assert invocation["source"] == deploy["id"]
 
 
+@pytest.mark.parametrize("invocation", [
+    '"$script_dir/helpers.sh"',
+    '"${script_dir}/helpers.sh"',
+    '"$script_dir/helpers.sh" --flag value',
+    '$script_dir/helpers.sh',
+])
+def test_extract_bash_script_invocation_via_variable_built_path(tmp_path, invocation):
+    helpers = tmp_path / "helpers.sh"
+    helpers.write_text("#!/bin/bash\necho helper\n", encoding="utf-8")
+    script = tmp_path / "deploy.sh"
+    script.write_text(
+        "#!/bin/bash\n"
+        'script_dir="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"\n'
+        f"{invocation}\n",
+        encoding="utf-8",
+    )
+
+    result = extract_bash(script)
+    invocations = [
+        edge for edge in result["edges"]
+        if edge.get("relation") == "calls" and edge.get("context") == "script_invocation"
+    ]
+
+    assert invocations == [{
+        "source": _make_id(str(script)) + "__entry",
+        "target": _make_id(str(helpers.resolve())) + "__entry",
+        "relation": "calls",
+        "confidence": "EXTRACTED",
+        "source_file": str(script),
+        "source_location": "L3",
+        "weight": 1.0,
+        "context": "script_invocation",
+        # Transient canonicalization hint (#2243); popped before persist.
+        "target_file": str(helpers.resolve()),
+    }]
+
+
+def test_extract_bash_variable_built_invocation_with_dynamic_stem_emits_no_edge(tmp_path):
+    helpers = tmp_path / "helpers.sh"
+    helpers.write_text("#!/bin/bash\necho helper\n", encoding="utf-8")
+    script = tmp_path / "deploy.sh"
+    script.write_text(
+        "#!/bin/bash\n"
+        'script_dir="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"\n'
+        '"$script_dir/$name.sh"\n',
+        encoding="utf-8",
+    )
+
+    result = extract_bash(script)
+
+    assert not any(edge.get("context") == "script_invocation" for edge in result["edges"])
+
+
+def test_extract_bash_variable_built_invocation_no_on_disk_match_emits_no_edge(tmp_path):
+    script = tmp_path / "deploy.sh"
+    script.write_text(
+        "#!/bin/bash\n"
+        'script_dir="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"\n'
+        '"$script_dir/missing.sh"\n',
+        encoding="utf-8",
+    )
+
+    result = extract_bash(script)
+
+    assert not any(edge.get("context") == "script_invocation" for edge in result["edges"])
+
+
+def test_extract_bash_variable_built_invocation_targets_existing_entrypoint(tmp_path):
+    helpers = tmp_path / "helpers.sh"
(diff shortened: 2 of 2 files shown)


safishamsi:
Shipped in **v0.9.58** — landed on `v8` via cherry-pick with your authorship preserved. Thanks @edwardselby — a shell script invoked via a constant variable path now resolves to the target. Release: https://github.com/Graphify-Labs/graphify/releases/tag/v0.9.58

PR #3420, opened 2026-09-08
Fix incremental merge wiping cross-file project AST nodes

### LINKED ISSUE #3411: Incremental extract wipes PackageReference/TargetFramework nodes and edges of every .csproj referenced from a re-extracted .sln / .csproj (cross-file stub nodes pollute build_merge's replace-set)
## Summary

On an incremental `graphify extract`, every `.csproj` that is merely *referenced* from a re-extracted `.sln` (or from a re-extracted `.csproj` via `ProjectReference`) is treated by `build_merge` as if it had been re-extracted itself. Its existing AST-tier nodes and edges (project node, `nuget_*` PackageReference nodes, `framework_*` TargetFramework nodes, their `imports`/`references` edges) are dropped, and only the bare cross-file stub node comes back. The manifest still stamps those `.csproj` files as unchanged, so the loss persists until a full `--force` rebuild.

## Environment

- graphifyy 0.9.56 (same code in 0.9.53 and 0.9.55), Linux, Python 3.13
- C# monorepo: 1 `.sln`, 416 `.csproj`, ~11 000 files
- nightly `graphify extract <repo> --backend foundry --no-viz --out <dir>` (incremental)

## Symptom

Counts taken from the dated `graphify-out/<date>/graph.json` backups of the same repo:

| graph.json | nodes | `nuget_*` nodes | `imports` edges whose `source_file` is a `*.csproj` |
|---|---|---|---|
| full rebuild (`--force --code-only`) | 105 775 | 129 | 2 121 |
| next incremental run (`PD3.sln` changed) | 105 598 | **1** | **33** |
| earlier full rebuild | 102 634 | 131 | 1 816 |
| +1 incremental (a few `.csproj` with `ProjectReference` changed) | 102 827 | 84 | 1 707 |
| +2 incremental (`PD3.sln` changed) | 104 184 | **4** | **77** |

Each such run also prints hundreds of

```
[graphify] note: node 'src_..._csproj' was extracted twice from 'src/.../X.csproj' under different labels — keeping 'X', dropping 'X.csproj'.
```

and `[graphify] Replaced 1650 node(s) from re-extracted source file(s).` although only ~100 files changed. In our case the resulting net shrink happened to coincide with an LLM omission in the semantic pass, so the #3203/#3105 shrink guard refused the write and the nightly job failed; on an ordinary night the wipe is written silently (`force=True` on the incremental path).

## Cause

1. `graphify/extractors/sln.py` (~L52–57) emits one node per solution project with `"source_file": abs_proj` — the absolute path of the **referenced** `.csproj`, not the `.sln` being extracted. The `ProjectReference` branch of the csproj extractor in `graphify/extract.py` (~L5005–5015) does the same (`"source_file": abs_ref`).
2. `build_merge` in `graphify/build.py` (~L1785–1800) derives the replace-set `new_ast_sources` from `n["source_file"]` of **every node** in `new_chunks` (raw and `_norm_source_file`-relativised).
3. Therefore every project referenced from the re-extracted `.sln` / `.csproj` becomes a "re-extracted AST source"; `_kept()` drops all of its existing AST-tier nodes and edges, and the chunk only re-adds the stub node (same id, label `X.csproj` → the "extracted twice" note). The manifest entry of the referenced `.csproj` is not touched, so nothing re-extracts it later.

## Minimal repro

```
repo/
  A.sln          # lists projects A and B
  A/A.csproj     # 2 PackageReference, 1 TargetFramework
  B/B.csproj     # ProjectReference ../A/A.csproj
```

1. `graphify extract . --no-viz` → `nuget_*` / `framework_*` nodes and `A.csproj → nuget` edges are present.
2. Edit `A.sln` only (reorder a line or add a comment) so that just the `.sln` is re-extracted.
3. `graphify extract . --no-viz` → `A.csproj`'s `nuget_*` / `framework_*` nodes and their edges are gone; `manifest.json` still lists `A/A.csproj` as unchanged.

The same happens when step 2 edits `B.csproj` instead of the `.sln` (the `ProjectReference` stub carries `source_file = A/A.csproj`).

## Suggested fix

- Derive the replace-set from the file each chunk was actually extracted from (the extractor's own `path`), not from per-node `source_file`; **or**
- mark cross-file stub nodes (e.g. `"_stub": True` / `"_origin": "ref"`) in `extract_sln` and the `ProjectReference` branch and skip them when building `new_ast_sources` — the AST-tier mirror of the #2926 scoping already applied to the semantic tier.

Either way a stub node must never make an un-dispatched file count as re-extracted.

## Workaround we use meanwhile

Before each incremental run we remove the `*.csproj` / `*.sln` entries from `manifest.json`, so every project file is re-parsed each run (cheap XML) and its fresh nodes are in the same chunks as the stubs. This restores the NuGet/project-dependency layer on every run, at the cost of the "extracted twice" / "minted by two different files" noise.

hopstreax:
## Summary

Fixes #3411.

Incremental extraction of a `.sln` or `.csproj` could incorrectly remove AST nodes and edges belonging to referenced `.csproj` files.

Cross-file project stubs use the referenced project's `source_file`, causing `build_merge()` to treat those projects as re-extracted even though they were not dispatched for extraction. This could wipe `PackageReference`, `TargetFramework`, project, and dependency edges from unchanged projects.

## Changes

* Record the files actually dispatched for extraction as `extracted_sources`.
* Propagate extraction provenance through the CLI merge path.
* Derive AST replacement ownership from extraction provenance instead of cross-file stub `source_file` values.
* Share replacement-source calculation between clustered and raw merge paths.
* Preserve backward compatibility for legacy chunks without explicit provenance.
* Add regression tests covering:

  * `.sln` → `.csproj` stubs
  * `.csproj` → `ProjectReference` stubs
  * raw/no-cluster merging
  * explicit AST source provenance

## Validation

* `pytest tests/test_build.py` — 91 passed
* `.NET` / incremental / shrink-guard suites — 78 passed
* Full `pytest` — 5,055 passed, 286 skipped
* `git diff --check` — clean

The key invariant is now:

> A cross-file stub's `source_file` may identify a dependency target, but it must not imply that the target file was re-extracted.

### DIFF
--- a/graphify/build.py
+++ b/graphify/build.py
@@ -1559,11 +1559,79 @@ def _load_existing_graph(graph_path: Path) -> "tuple[list, list, list, bool] | N
     )
 
 
+def _tier_replacement_sources(
+    chunks: "Iterable[dict]",
+    root: "str | Path | None" = None,
+    ast_sources: "Iterable[str | Path] | None" = None,
+) -> tuple[set[str], set[str]]:
+    """Compute (new_ast_sources, new_sem_sources) for tier-scoped replacement.
+
+    #3411: AST replacement ownership is derived from explicit extraction
+    provenance (ast_sources or chunk-level "extracted_sources"), so cross-file
+    stub nodes emitted by extractors (e.g. .sln project stubs or ProjectReference
+    stubs) do not pollute the replacement set and wipe the target project's
+    nodes/edges. Falls back to AST node source_file only when no provenance is
+    provided.
+    """
+    explicit_ast_sources: set[str] = set()
+    if ast_sources is not None:
+        for s in ast_sources:
+            if s:
+                explicit_ast_sources.add(str(s))
+    for ch in chunks:
+        if isinstance(ch, dict):
+            for s in (ch.get("extracted_sources") or []):
+                if s:
+                    explicit_ast_sources.add(str(s))
+
+    new_ast_sources: set[str] = set()
+    new_sem_sources: set[str] = set()
+
+    if explicit_ast_sources:
+        for sf in explicit_ast_sources:
+            new_ast_sources.add(sf)
+            norm = _norm_source_file(sf, root)
+            if norm:
+                new_ast_sources.add(norm)
+    else:
+        for ch in chunks:
+            if not isinstance(ch, dict):
+                continue
+            for n in ch.get("nodes", []):
+                if not isinstance(n, dict):
+                    continue
+                sf = n.get("source_file")
+                if not sf or not _is_ast_tier(n):
+                    continue
+                new_ast_sources.add(sf)
+                norm = _norm_source_file(sf, root)
+                if norm:
+                    new_ast_sources.add(norm)
+
+    for ch in chunks:
+        if not isinstance(ch, dict):
+            continue
+        for n in ch.get("nodes", []):
+            if not isinstance(n, dict):
+                continue
+            sf = n.get("source_file")
+            if not sf or _is_ast_tier(n):
+                continue
+            new_sem_sources.add(sf)
+            norm = _norm_source_file(sf, root)
+            if norm:
+                new_sem_sources.add(norm)
+
+    return new_ast_sources, new_sem_sources
+
+
 def merge_raw_extraction(
     new: dict,
     graph_path: str | Path,
     prune_sources: "list[str] | None" = None,
     root: "str | Path | None" = None,
+    *,
+    ast_sources: "Iterable[str | Path] | None" = None,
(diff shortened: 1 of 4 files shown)


safishamsi:
Shipped in **v0.9.57** — landed on `v8` via cherry-pick with your authorship preserved. Thanks @hopstreax — an incremental rebuild no longer wipes cross-file project AST nodes of a referenced .csproj/.sln project. Release: https://github.com/Graphify-Labs/graphify/releases/tag/v0.9.57

PR #3422, opened 2026-09-08
fix(csharp): resolve unqualified generic call sites, Get<int>(...) (#3406)

### LINKED ISSUE #3406: C#: an unqualified generic call (`Get<T>(...)`, `this.Get<T>(...)`) drops its `calls` edge
## Summary

In C#, a call with an explicit type argument and **no receiver** never gets a `calls` edge, even
when the target method is declared on the caller's own class or on its base class. The
non-generic spelling of the same call resolves normally, so the type-argument list at the call
site is the whole trigger.

Measured on **v0.9.56**, the current release (published 2026-09-07). v0.9.55 behaves identically.

## Reproducer

Three files, no corpus, no build needed. `dotnet build` on net8.0: 0 errors, 0 warnings.

```csharp
// Settings.cs
namespace Demo
{
    public class Settings
    {
        public T Get<T>(string key) { return default(T); }   // target
        public string GetRaw(string key) { return key; }     // control target
    }
}

// Reader.cs
namespace Demo
{
    public class Reader : Settings
    {
        public int A() { return Get<int>("port"); }        // 1) no edge
        public int B() { return this.Get<int>("port"); }   // 2) no edge
        public string C() { return GetRaw("host"); }       // 3) edge
        public string D() { return this.GetRaw("host"); }  // 4) edge
    }
}

// Local.cs
namespace Demo
{
    public class Local
    {
        public T Make<T>() { return default(T); }
        public int E() { return Make<int>(); }             // 5) no edge, same class
    }
}
```

```
graphify extract src --code-only --no-cluster --out out
```

| # | Call site | v0.9.56 |
|---|---|---|
| 1 | `Get<int>("port")` — unqualified, generic, target on base class | **no edge** |
| 2 | `this.Get<int>("port")` — same through `this` | **no edge** |
| 3 | `GetRaw("host")` — control, unqualified, non-generic | edge |
| 4 | `this.GetRaw("host")` — control, through `this` | edge |
| 5 | `Make<int>()` — unqualified, generic, target on the same class | **no edge** |

**2 of 5.** All five target member nodes exist in the graph, and `method`/`inherits` edges are
correct — only the call-site-to-member match fails. Case 5 shows inheritance is not involved.

## Expected

Cases 1, 2 and 5 should produce a `calls` edge to `.Get()` / `.Make()`, exactly as the
non-generic controls 3 and 4 do. The type-argument list is not part of the method's identity.

## Where it happens (v0.9.56, `graphify/extractors/engine.py`; identical in v0.9.55)

* **`:5389`** `elif fn_node is not None and fn_node.type == "identifier":` — for `Get<int>("port")`
  tree-sitter gives the `function` field as a `generic_name`, not an `identifier`, so this branch
  is skipped. Control flow falls into the generic fallback at **`:5391-5408`**, which takes the
  node text verbatim (`callee_name = raw` at **`:5407`**) and so assigns
  `callee_name = "Get<int>"`.
* **`:5365`** `callee_name = _read_text(mname, source)` — for `this.Get<int>("port")` the `name`
  field of the `member_access_expression` is a `generic_name`, so `callee_name` becomes
  `"Get<int>"`.
* In both cases the member index is keyed on the bare declared name, the lookup misses, and the
  edge is dropped silently.

**What makes this worth one small change:** a few lines further down, at **`:5409` and following**,
the very same call node is walked *structurally* to emit the `references[generic_arg]` edges added
in #2911 — the code descends into `generic_name` and reads its `type_argument_list` properly. So
v0.9.56 already recognises these call sites as generic and already records what they instantiate;
only the callee name a few lines above is still taken as raw text. The comment on that block says
so itself: "the fallback path uses raw text and never sees the structured type-arg list".

## Relation to existing reports — deliberately not a duplicate

#2624 (open) reports the **qualified** form, `Registry.Fetch<Payload>()` and
`Bag<Payload>.Make()`, and its reproducer contains only those. The unqualified forms appear nowhere in
that issue's reproducer; they were mentioned in passing in one comment on it, and demonstrated in
a comment on PR #2676. This report is about those forms only, with its own reproducer, its own
prevalence count, and — for the receiverless form — its own code path: `:5389` and the fallback,
which #2624's reproducer never reaches. (The `this.`-spelling shares `:5365` with it.)

PR #2676, open since 2026-08-12, strips the type-argument list from the call-site name; measured
locally against v0.9.56 it also resolves cases 1, 2 and 5 above. It is not part of any release,
so the behaviour described here is what v0.9.56 does today.

#2911 (type arguments at the call site) and #2913 (the same in field position) are closed and
their effect is in the release — the engine comment at `:5409` names both. The unqualified
`calls` edge is what is left.

## Scope

Counted on the syntax tree (not by regex) over a ~15,300-file C# estate, mixed .NET Framework and
net8: **2159** unqualified generic call sites and **628** spelled through `this`. Those are
occurrences of the shape, **not** confirmed lost edges — unlike the qualified form, whose targets
are frequently outside the corpus, these resolve against the caller's own class or a base class,
so their targets are in-corpus by construction. I am reporting the shape's prevalence, not
claiming a number of lost edges.

The three files above are the whole reproducer -- no corpus, no build step, and the
behaviour is reproducible from the PyPI release alone.

abhay-codes07:
Closes #3406.

## The problem

A C# call with an explicit type argument and no receiver never got a `calls` edge — `Get<int>("port")`, `this.Get<int>("port")`, `Make<int>()` all dropped it while the non-generic spellings resolved, exactly as the issue's 5-case table shows (reproduced verbatim on 0.9.56).

The invocation handler has arms for a `member_access_expression` function node and an `identifier` one. `Get<int>(...)` parses as a **`generic_name`**, so it fell to the raw-text fallback and the callee was captured as `Get<int>` — brackets and all — which can never match the `.Get()` member node. `this.Get<int>(...)` failed one arm earlier: the member branch read the `name` field's raw text, and that field is the same `generic_name`.

## The change

A small helper, `_csharp_bare_call_name`, reads the identifier child of a `generic_name` (and passes a plain `identifier` through untouched). The member arm uses it for the name field, and the unqualified path gains a `generic_name` arm beside the `identifier` one. The type-argument list is not part of the method's identity, so the callee is the bare name — after that, same-class and base-class resolution work exactly as they do for the non-generic controls. The #2911 `generic_arg` reference emission reads the type-argument list off the same nodes and is untouched.

## Tests

`tests/test_csharp_generic_callsites.py` — 5 tests on the issue's exact 3-file repro: the unqualified base-class call, the `this.`-qualified call, and the same-class call all resolve; no target ever derives from the bracketed raw text; and the two non-generic controls keep resolving. With the fix reverted, 3 of 5 fail (the controls rightly keep passing). The C# suites are unchanged (174 passed); the full suite matches a fresh same-version `v8` (0.9.56) baseline.

### DIFF
--- a/graphify/extractors/engine.py
+++ b/graphify/extractors/engine.py
@@ -2775,6 +2775,20 @@ def _has_multiline_error(root) -> bool:
     return False
 
 
+def _csharp_bare_call_name(name_node, source: bytes) -> str:
+    """The callee identifier of a C# call-site name node.
+
+    A plain `identifier` reads as-is; a `generic_name` (`Get<int>`) reads its
+    identifier child so the type-argument list never leaks into the callee
+    name (#3406). Falls back to raw text for anything else.
+    """
+    if name_node.type == "generic_name":
+        for child in name_node.children:
+            if child.type == "identifier":
+                return _read_text(child, source)
+    return _read_text(name_node, source)
+
+
 def _read_csharp_type_name(node, source: bytes) -> tuple[str, bool, str] | None:
     """Resolve a C# type name, whether it was qualified, and its qualifier prefix."""
     if node is None:
@@ -5362,7 +5376,11 @@ def _extract_generic(
                     mname = fn_node.child_by_field_name("name")
                     recv = fn_node.child_by_field_name("expression")
                     if mname is not None:
-                        callee_name = _read_text(mname, source)
+                        # `recv.Get<int>(...)`: the name field is a
+                        # generic_name; its raw text carries the type-argument
+                        # list, which is not part of the method's identity
+                        # (#3406) — read the bare identifier instead.
+                        callee_name = _csharp_bare_call_name(mname, source)
                         is_member_call = True
                         if recv is not None and recv.type == "identifier":
                             member_receiver = _read_text(recv, source)
@@ -5388,6 +5406,13 @@ def _extract_generic(
                                 member_receiver = _read_text(fname, source)
                 elif fn_node is not None and fn_node.type == "identifier":
                     callee_name = _read_text(fn_node, source)
+                elif fn_node is not None and fn_node.type == "generic_name":
+                    # Unqualified generic call `Get<int>(...)` / `Make<T>()`:
+                    # without this arm the raw-text fallback captured
+                    # `Get<int>` verbatim, so the call never matched the
+                    # `.Get()` member and the edge was silently dropped —
+                    # while the non-generic spelling resolved fine (#3406).
+                    callee_name = _csharp_bare_call_name(fn_node, source)
                 else:
                     # Fallback: original name-field / first-named-child scan.
                     name_node = node.child_by_field_name("name")
--- a/tests/test_csharp_generic_callsites.py
+++ b/tests/test_csharp_generic_callsites.py
@@ -1,0 +1,76 @@
+"""Unqualified generic C# calls keep their calls edge (#3406).
+
+`Get<int>("port")` parses with a `generic_name` function node; the invocation
+handler had arms for `member_access_expression` and `identifier` only, so the
+raw-text fallback captured `Get<int>` verbatim and the call never matched the
+`.Get()` member — while the non-generic spelling of the same call resolved.
+`this.Get<int>(...)` failed the same way through the member arm, whose name
+field is the same `generic_name`.
+"""
+
+from graphify.extract import extract
+
+SETTINGS_CS = (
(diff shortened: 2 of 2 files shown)


safishamsi:
Shipped in **v0.9.57** — landed on `v8` via cherry-pick with your authorship preserved. Thanks @abhay-codes07 — a C# generic call site with explicit type arguments (Get<int>(...)) now resolves to the method definition. Release: https://github.com/Graphify-Labs/graphify/releases/tag/v0.9.57

PR #3423, opened 2026-09-08
fix(js): keep this.X = function members in every enclosing-function form (#3408)

### LINKED ISSUE #3408: JS: `this.X = function` is extracted only inside a function *declaration* — every function-expression form drops it (incomplete fix of #1322)
**Title:** JS: `this.X = function` is extracted only inside a function *declaration* — every function-expression form drops it (incomplete fix of #1322)

### Summary

#1322 listed `this.x = function(){}` "inside any function body" and was closed. In 0.9.56 only one enclosing form works: a named function **declaration**. If the enclosing function is an expression — assigned to a const, an arrow, an IIFE, or a callback argument — the member is silently dropped.

The callback-argument case is the AngularJS service pattern, so an AngularJS frontend extracts to file nodes only.

### Repro

```js
// p2.js
function DeclFn(){ this.inDecl = function(){ return 1; }; }
const AnonToConst = function(){ this.inAnonConst = function(){ return 2; }; };
const ArrowToConst = () => { this.inArrow = function(){ return 3; }; };
callIt(function(){ this.inCallbackArg = function(){ return 4; }; });
(function(){ this.inIIFE = function(){ return 5; }; })();
```

`graphify update .` produces:

```
.inDecl()        <-- only this one
AnonToConst()
ArrowToConst()
DeclFn()
p2.js
```

| enclosing function | result |
|---|---|
| `function DeclFn(){}` declaration | extracted |
| `const A = function(){}` | **missing** |
| `const A = () => {}` | **missing** |
| `callIt(function(){})` | **missing** |
| `(function(){})()` | **missing** |

### Framework form this hits

```js
angular.module('myApp').service('$thingService', function ($rootScope, $http) {
	this.loadThing = function (id) { /* ... */ };
	this.saveThing = function (thing) { /* ... */ };
});
```

A service file with 16 such members yields 0 nodes for them. A named `function foo(){}` declared inside the same callback **is** extracted, so the callback body is walked — only the `this.X` assignment is dropped.

### Expected

A node per member, contained by the enclosing function, in every enclosing form. Per #3338, `_js_member_assignment_target` materializes only `exports.X` and `Foo.prototype.X` and other receivers are dropped by the #1077 guard, so `this` inside a function expression may just need adding to that allowlist alongside the `window`/`globalThis` work.

### Impact

Measured on one private AngularJS + PHP codebase, 0.9.56, fully rebuilt:

| layer | files | nodes | avg/file | contributing only the file node |
|---|---|---|---|---|
| PHP class layer | 844 | 7913 | 9.4 | 1 (0%) |
| AngularJS frontend | 661 | 1053 | 1.6 | **573 (87%)** |

The class layer extracts essentially completely. The frontend is opaque below file level, so the graph cannot answer "where is this behaviour implemented" for any of it.

### Related

#1322 (closed, declaration case only), #2419/#2433 (object-literal shorthand), #3337/#3338 (`window`/`globalThis` receiver). Same family, different receiver and enclosing form.

### Environment

graphify 0.9.56, macOS 15 (arm64), installed via `uv tool install graphifyy`.

abhay-codes07:
Closes #3408.

## The problem

#1322's fix captured `this.X = function` members only when the enclosing function was a function **declaration**. Every expression form — `const F = function(){…}`, an arrow, an IIFE, a callback argument — silently dropped them. The callback form is the AngularJS registration shape (`.service('name', function(){ this.load = … })`), which is how the issue's measured corpus ends up 87% file-node-only on its frontend: the graph cannot answer "where is this behaviour implemented" for any of it. All five forms reproduce on 0.9.56.

The member scan lived inline in the function-declaration branch of `_extract_generic`, and expression-form bodies never pass through that branch.

## The change

The scan is factored into `_js_scan_member_assignments` (a pure move of the existing logic, including the `api.X = fn` object-binding case and its #1077 flood guard) and now runs from every enclosing form:

- **Declaration branch** — unchanged behavior, now via the helper.
- **`const F = function(){…}` / `const F = () => {…}`** — the lexical-declaration branch that already materializes `F()` scans its body too; members become `method` edges owned by `F`, same as the declaration case.
- **Module-level call statements** — `callIt(function(){…})`, the AngularJS `.service(…)` chain, `(function(){…})()` — each TOPMOST closure body (reusing `_js_topmost_closures`, so nested closures stay with their own scope) is scanned. The enclosing function is anonymous, so members attach to the **file node** as `contains` edges with file-qualified ids — the same shape `exports.X` members already have, and the #1077 phantom-god-node guard's requirement.

## Tests

`tests/test_js_this_member_expressions.py` — 5 tests: the issue's five-form repro extracts every member; `const F = function(){…}` members are methods owned by `F`; the AngularJS service shape yields both members hanging off the file node with file-qualified ids; a `const` local inside a callback still produces no node (#1077 control); and the declaration form emits exactly one node and one method edge (no double scan). With the fix reverted, 4 of 5 fail (the locals control rightly keeps passing). The JS/TS suites are unchanged (1072 passed; the two failing names fail identically on clean `v8` on this Windows machine); the full suite matches a fresh same-version `v8` (0.9.56) baseline.

### DIFF
--- a/graphify/extractors/engine.py
+++ b/graphify/extractors/engine.py
@@ -2218,6 +2218,80 @@ def _js_member_assignment_target(left, source: bytes):
                 return ("prototype", inner_obj_name, member_name)
     return None
 
+def _js_scan_member_assignments(
+    body, this_owner_nid: str, source: bytes, *,
+    this_as_method: bool, add_node_fn, add_edge_fn, function_bodies: list,
+) -> None:
+    """Materialize callable members assigned in ``body`` (#1322/#3408).
+
+    ``this.X = fn`` members attach to ``this_owner_nid`` — as ``method`` edges
+    with a ``.X()`` label when the owner is a named function or class
+    (``this_as_method``), and as ``contains`` edges with an ``X()`` label when
+    the owner is the file node (the enclosing function is anonymous: an IIFE
+    or a callback argument, e.g. an AngularJS ``.service(...)`` body).
+    ``api.X = fn`` members attach to the object-literal binding declared in
+    the same body, exactly as before — the scope proof that keeps the #1077
+    phantom-god-node guard intact.
+    """
+    object_bindings: dict[str, object] = {}
+    for stmt in body.children:
+        if stmt.type not in ("lexical_declaration", "variable_declaration"):
+            continue
+        for declarator in stmt.children:
+            if declarator.type != "variable_declarator":
+                continue
+            name = declarator.child_by_field_name("name")
+            value = declarator.child_by_field_name("value")
+            if name is not None and name.type == "identifier" \
+                    and value is not None and value.type == "object":
+                object_bindings[_read_text(name, source)] = declarator
+    # A factory object gets one owner node and one `contains` edge no
+    # matter how many methods hang off it. add_node dedups on id, but
+    # add_edge does not, so without this guard N assigned methods would
+    # emit N identical `contains` edges (the flood #1077 warns against).
+    contained_owners: set[str] = set()
+    for stmt in body.children:
+        if stmt.type != "expression_statement":
+            continue
+        assign = next((c for c in stmt.children
+                       if c.type == "assignment_expression"), None)
+        if assign is None:
+            continue
+        val = assign.child_by_field_name("right")
+        if val is None or val.type not in _JS_FUNCTION_VALUE_TYPES:
+            continue
+        tgt = _js_member_assignment_target(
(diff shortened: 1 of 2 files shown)


safishamsi:
Shipped in **v0.9.57** — landed on `v8` via cherry-pick with your authorship preserved. Thanks @abhay-codes07 — this.X = function members are now captured in every enclosing-function form, not just declarations. Release: https://github.com/Graphify-Labs/graphify/releases/tag/v0.9.57

Issue #2479, opened 2026-08-05
Noise-dir list matches bare build by name alone: real project directories named build/ are silently dropped from the scan


alexrdunbar:
### Summary

`_SKIP_DIRS` contains the bare string `"build"`, and `_is_noise_dir` matches it on directory name alone with no corroborating evidence. Any real project directory literally named `build/` is silently excluded from the scan, with no warning and no way to override it from `.graphifyignore` (an ignore file cannot un-ignore a hard-skipped dir).

Found while remediating #2478 on the same repo. Different mechanism, same failure signature: files missing from the graph, exit code 0, nothing printed.

### Environment

- graphify (`graphifyy`) 0.9.12
- Windows 11, Python 3.x

### The check

```python
def _is_noise_dir(part: str, parent: "Path | None" = None) -> bool:
    """Return True if this directory name looks like a venv, cache, or dep dir."""
    if part in _SKIP_DIRS:
        return True
    if part == "snapshots":
        # Prune only when it looks like an actual JS/Vitest snapshot dir.
        ...
```

`_SKIP_DIRS` includes `build`, alongside `dist`, `out`, `target`, `coverage`, `env`, and others with the same problem.

Note that `snapshots` immediately below it does the right thing: it verifies before pruning (checks the parent directory name, then looks for an actual `*.snap` file, and keeps the directory when it cannot confirm). `build` gets no such treatment.

### Repro

A directory literally named `build/` containing a real source file:

```
.claude/skills/build/SKILL.md      # a real skill definition, ~4 KB of markdown
```

Scan the repo root, then check the manifest:

```python
>>> import json
>>> m = json.load(open('graphify-out/manifest.json'))
>>> ks = list(m['files'].keys())
>>> len(ks)
2159
>>> len([k for k in ks if 'skills' in k.replace('\\', '/')])
725
>>> [k for k in ks if '/build/' in k.replace('\\', '/')]
[]
```

725 sibling skill files indexed, zero from `build/`. The file exists, is a supported type, is not gitignored from graphify's perspective, and is not reported anywhere in the run output.

Adding a `.graphifyignore` negation does not help, because the prune happens during the `os.walk` descent (`if not _is_noise_dir(d, dp)`), before ignore-file logic sees the path.

### Why this matters more than it looks

[...]

PR #3424, opened 2026-09-08
fix(detect): gate the bare out dir on build-output evidence (#3347)

### LINKED ISSUE #3347: detect: bare `out` in _SKIP_DIRS silently drops the entire outbound layer of hexagonal codebases (196/655 .java files)
### Summary

`_SKIP_DIRS` contains the bare string `"out"`, matched by name alone in `_is_noise_dir`. In a hexagonal / ports-and-adapters codebase, `adapter/out/` and `port/out/` are where the entire outbound layer lives — persistence adapters, JPA entities, messaging adapters, outbound port interfaces. All of it is silently dropped from the scan.

On the repo where I hit this, that was **196 of 655 `.java` files (30%)**, including every `@Entity` class. Exit code 0, no warning, nothing in the run output.

This is the same mechanism as #2479 (which reports it for `build/` and explicitly lists `out` as having the same problem). Filing separately because the blast radius is different in kind: `build/` loses a directory, `out/` loses an architectural layer, and the convention that puts source there is mainstream rather than incidental.

### Environment

- graphify (`graphifyy`) 0.9.54, installed via `uv tool install`
- macOS 26.6.2, CPython 3.11.15
- Command: `graphify extract . --code-only --timing`

### Mechanism

```python
_SKIP_DIRS = {
    "venv", ".venv",  # "env"/".env"/"*_env" are gated on venv markers below (#2058)
    "node_modules", "__pycache__", ".git",
    "dist", "build", "target", "out",
    ...
```

`_is_noise_dir` already gates **three** ambiguous names on corroborating evidence, each returning `False` when it cannot verify:

- `env` / `.env` / `*_env` → `_has_venv_markers` (#1666/#2058)
- `coverage` → `_has_coverage_artifacts` (#2339)
- `snapshots` → parent-name check plus an actual `*.snap` file

`out` gets none of it — `if part in _SKIP_DIRS: return True`, unconditionally, before any of those checks run.

### Repro

Any Spring/hexagonal layout, e.g.:

```
backend/src/main/java/.../domain/transaction/adapter/out/persistence/OrderEntity.java
backend/src/main/java/.../domain/transaction/port/out/OrderRepositoryPort.java
```

```console
[graphify extract] found 789 code, 0 docs, 0 papers, 0 images
[graphify extract] wrote graphify-out/graph.json: 6194 nodes, 14196 edges
```

```python
>>> import json, pathlib
>>> g = json.load(open('graphify-out/graph.json'))
>>> in_graph = {n['source_file'] for n in g['nodes'] if str(n.get('source_file','')).endswith('.java')}
>>> on_disk  = {str(p) for p in pathlib.Path('.').rglob('*.java') if '/target/' not in str(p)}
>>> len(on_disk), len(in_graph)
(655, 459)
>>> missing = on_disk - in_graph
>>> len(missing), sum('/out/' in m for m in missing)
(196, 196)
```

Every missing file contains `/out/`. Zero exceptions. And in that repo, every directory named `out/` contained **only `.java` files** — no build artifacts anywhere:

```console
$ find . -type d -name out -not -path "*/node_modules/*" -not -path "*/target/*" \
    -exec find {} -type f \; | sed 's/.*\.//' | sort | uniq -c
 196 java
```

Removing `"out"` from `_SKIP_DIRS` locally and re-running, same command, same repo:

| | before | after |
|---|---|---|
| code files scanned | 789 | 986 |
| `.java` in graph | 459 / 655 | 655 / 655 |
| nodes | 6,194 | 7,416 |
| edges | 14,196 | 17,759 |

### Why this is worse than a missing directory

The dropped set is not arbitrary — it is precisely the persistence and outbound-integration layer. So the graph that survives looks *coherent*: controllers, use-case ports and service implementations are all present and correctly linked, and the inbound half of every hexagon traces cleanly. Nothing looks broken.

What is gone is every path from application code to the database. In my case that meant zero edges between the Java subgraph and the SQL subgraph, and I initially attributed that to JPA's `@Table` mapping being non-syntactic — a plausible and completely wrong diagnosis, because the entity classes were not in the graph at all. A silently incomplete graph that still answers questions confidently is a worse failure mode than one that visibly fails.

Per #2479, a `.graphifyignore` negation cannot recover these files either, since the prune happens during the `os.walk` descent before ignore-file logic sees the path. I did not verify that part myself.

### Workaround

Pointing `extract` **at** the directory works, because `_SKIP_DIRS` only filters descendants, not the scan root:

```console
$ graphify extract backend/src/main/java/.../adapter/out --code-only --out /tmp/gout
[graphify extract] found 5 code files ... wrote graph.json: 86 nodes, 158 edges
```

Combined with `graphify merge-graphs` that is a viable escape hatch, but it needs one invocation per `out/` directory (~30 in this repo), which does not scale.

### Requests

1. **Gate `out` on evidence, like `env` / `coverage` / `snapshots` already are.** A real build output directory is cheap to recognise — compiled artifacts (`.class`, `.jar`, `.o`, `.wasm`), a sibling build file, absence of any source file of a supported language. When it cannot be confirmed, keep the directory.
2. **Failing that, warn.** The run already prints counts for skipped non-code, unclassified, and sensitive files. A single line — `skipped N file(s) in noise directories: out/ (196), ...` — would have turned this from a silent 30% loss into a five-second diagnosis.
3. Same argument extends to `dist` and `target`, though `out` is the one with a mainstream architectural convention behind it.

Happy to test a patch against this repo.

abhay-codes07:
Closes #3347.

## The problem

`_SKIP_DIRS` contains the bare name `out`, matched unconditionally — so in a hexagonal / ports-and-adapters codebase, `adapter/out/` and `port/out/` (the entire outbound layer: persistence adapters, entities, outbound ports) are silently dropped from the scan. On the reporting repo that was 196 of 655 `.java` files — 30%, including every `@Entity` class — with exit code 0 and no warning.

`_is_noise_dir` already gates three ambiguous names on corroborating evidence, keeping the directory when it cannot verify: `env`/`.env`/`*_env` on venv markers (#2058), `coverage` on report artifacts (#2339), `snapshots` on `.snap` files (#1666). `out` got none of it.

## The change

`out` becomes the fourth gated name. `_has_build_output_markers` prunes it only on real generated-output evidence:

- the IntelliJ compile layout (`out/production/` subdir);
- a Next.js static export (`out/_next/` subtree);
- compiled artifacts (`.class`, `.jar`, `.o`, `.obj`, `.exe`, `.dll`, `.pyc`, `.js.map`, `.css.map`, `.d.ts`) in the top two directory levels — a bounded `os.scandir` probe, same cost profile as the venv/coverage gates.

An `out` that cannot be verified is **kept** — the same keep-on-doubt bias every other gated name has, and pruned-as-noise dirs stay traceable via `pruned_noise_dirs` (#2058). `build`, `dist` and `target` remain unconditional; the wider question of gating those is #2479's, and this PR deliberately stays on the `out` case where the dropped set is an architectural layer under a mainstream source convention rather than an incidental directory.

## Tests

`tests/test_out_dir_evidence.py` — 5 tests: `adapter/out/` and `port/out/` holding only source stay in the corpus (the issue's shape, and the test that fails 
on the unpatched tree); IntelliJ `out/production/`, a `next export` `out/`, and a TS outDir with `.js.map` one level down are all still pruned; and `build`/`dist`/`target` stay unconditionally pruned. The detect suites are unchanged (584 passed; the two failing hook tests fail identically on clean `v8` on this Windows machine); the full suite matches a fresh same-version `v8` (0.9.56) baseline.

### DIFF
--- a/graphify/detect.py
+++ b/graphify/detect.py
@@ -933,11 +933,21 @@ def _has_build_output_markers(d: "Path") -> bool:
                     return True
                 if entry.is_dir():
                     subdirs.append(entry.path)
-        for sub in subdirs[:20]:  # bounded: probe one more level, cheaply
+        # Probe one more level across EVERY subdirectory (a wide TS outDir may
+        # keep its compiled files only under later-sorted module dirs), bounded
+        # by total entries scanned rather than by subdirectory count so a
+        # pathological tree still costs O(1)-ish.
+        budget = 4000
+        for sub in subdirs:
+            if budget <= 0:
+                break
             with os.scandir(sub) as it:
                 for entry in it:
+                    budget -= 1
                     if entry.is_file() and entry.name.lower().endswith(_BUILD_OUTPUT_SUFFIXES):
                         return True
+                    if budget <= 0:
+                        break
     except OSError:
         pass
     return False
--- a/graphify/detect.py
+++ b/graphify/detect.py
@@ -827,7 +827,7 @@ def _is_regular_file(path: Path) -> bool:
 _SKIP_DIRS = {
     "venv", ".venv",  # "env"/".env"/"*_env" are gated on venv markers below (#2058)
     "node_modules", "__pycache__", ".git",
-    "dist", "build", "target", "out",
+    "dist", "build", "target",  # bare "out" is gated on build-output evidence below (#3347)
     "site-packages", "lib64",
     ".pytest_cache", ".mypy_cache", ".ruff_cache",
     ".tox", ".nox", ".eggs", "*.egg-info",  # nox is tox's successor, same .nox/ venv shape (#1804)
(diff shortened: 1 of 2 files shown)


abhay-codes07:
On the two advisories: the 20-subdirectory cap was a fair catch — ff0a8f6 replaces it with a probe over every subdirectory bounded by total entries scanned (4000), with a test putting the only .js.map under the 31st-sorted subdir. The second finding describes the PR's intended behavior rather than a defect: an unverifiable `out` is now kept on purpose, matching the keep-on-doubt bias the env/coverage/snapshots gates already have — the trade is discussed in the PR body, and pruned dirs stay traceable via pruned_noise_dirs.

safishamsi:
Shipped in **v0.9.58** — landed on `v8` via cherry-pick with your authorship preserved. Thanks @abhay-codes07 — a bare `out` dir is skipped only with build-output evidence, so a real source out/ isn't dropped. Release: https://github.com/Graphify-Labs/graphify/releases/tag/v0.9.58

PR #3425, opened 2026-09-08
fix(dedup): pick the richer duplicate as survivor and keep the losers' fields (#3372)

### LINKED ISSUE #3372: Cross-file dedup picks survivor by shortest node ID, discarding the richer node's content
`deduplicate_entities`'s cross-file exact-match pass (in `build.py`, gated to `concept`-type nodes sharing a normalized label) merges two nodes for the same real-world entity into one survivor. The survivor is chosen by `dedup.py`'s `_pick_winner()`:

```python
def _pick_winner(nodes: list[dict]) -> dict:
    """Pick the canonical survivor: prefer no chunk suffix, then shorter ID."""
    def _score(n): return (1 if _CHUNK_SUFFIX.search(n["id"]) else 0, len(n["id"]))
    return min(nodes, key=_score)
```

This scores purely on ID shape/length — it has no regard for which node actually carries content (`attributes`, `confidence`, `_merged_from` history, etc.). In practice, a dedicated page for an entity tends to sit in a nested category folder, giving it a longer node ID (e.g. `topics/networking/widget-x.md`), while a passing mention of that same entity on an unrelated, flat summary page mints a short, shallow node (e.g. `sources/notes.md`) purely because it has one fewer path segment — regardless of how the two filenames themselves compare in length. Since the shorter ID always wins, the established, enriched node is discarded and replaced by the shallow one — every time this pattern occurs.

**Repro:**

1. Create `topics/networking/widget-x.md`, a dedicated page for an entity "Widget X" that produces a rich node (multiple attributes, high confidence, etc.).
2. Extract it — `Widget X`'s node now has real content.
3. Create `sources/notes.md`, an unrelated, top-level page that only mentions "Widget X" in passing (one line, no real detail).
4. Extract it. The exact-match cross-file dedup pass fires (same normalized label, both `concept` type), and — because `sources/notes.md` sits one directory level shallower, making its node ID shorter regardless of the actual filename — the shallow node from step 3 wins. Diff `graph.json` before/after: `Widget X`'s `attributes` (and any `_merged_from` history) from step 1 are gone, replaced by the shallow stub.

Note this is pure content loss, not structural corruption — edges pointing at the old survivor are correctly rewired first (confirmed via `graphify diagnose` / the health check: zero dangling edges).

**Suggested fix:** score candidates by content richness first (e.g. number of populated fields, presence of `_merged_from`/`attributes`/`confidence`) and fall back to ID length only as a tiebreaker among equally-rich candidates, rather than using ID length as the primary signal.

abhay-codes07:
Closes #3372.

## The problem

`_pick_winner` chose the dedup survivor by ID shape alone — no chunk suffix, then **shorter ID**. A passing one-line mention of an entity on a shallow page mints a shorter node id than the entity's dedicated page in a nested category folder, so the shallow stub won every time the pattern occurred, and the losers are dropped wholesale in the component collapse — the established node's `description`, `attributes`, `confidence` and `_merged_from` history vanished from the graph. Reproduced on 0.9.56 with the issue's two-node shape: the `sources/notes.md` stub survives and every rich field is gone.

## The change

Two complementary pieces, both in `dedup.py`:

- **Richness decides the survivor.** `_content_richness` counts populated fields beyond the identity/placement baseline (id, label, source_file, …), weighting `attributes` by entry count and `_merged_from` by history length — a node that already absorbed merges is the established record. `_pick_winner` scores `(chunk-suffix, −richness, id length)`: the chunk-suffix rule still dominates (a `_c3` fragment never wins, however rich), and ID length still breaks ties between equally-rich candidates, preserving the old deterministic ordering exactly where the old signal was actually doing its job. All 139 existing dedup tests pass unchanged.
- **Losers' fields fold into the survivor.** Even with the right winner, dropping the losers loses whatever only they carried. At component collapse the survivor is back-filled with each loser's fields via the same never-override `_merge_missing_attributes` the same-source collision path already uses (#2091), in deterministic `unique_nodes` order. Content loss becomes a union.

## Tests

`tests/test_dedup_survivor_richness.py` — 6 tests: the issue's exact shape keeps the enriched node (id, attributes, merge history intact); a field only the loser carried is folded into the survivor without overriding anything; equally-bare candidates keep the shorter-id tiebreak; a chunk-suffixed id never beats a clean one regardless of richness; the richness measure ignores placement fields; and edges rewire to the rich survivor. With the fix reverted, 4 of 6 fail (the tiebreak and chunk-suffix controls rightly keep passing). The dedup suites are unchanged (139 passed); the full suite matches a fresh same-version `v8` (0.9.56) baseline.

### DIFF
--- a/graphify/dedup.py
+++ b/graphify/dedup.py
@@ -850,6 +850,8 @@ def deduplicate_entities(
         n["id"]: (i, n) for i, n in enumerate(unique_nodes)
     }
 
+    # Survivors enriched with their losers' fields, substituted at the end.
+    enriched_by_id: dict[str, dict] = {}
     for root, members in components.items():
         if len(members) == 1:
             continue
@@ -861,6 +863,17 @@ def deduplicate_entities(
         ]
         winner = _pick_winner(group_nodes) if group_nodes else {"id": root}
         winner_id = winner["id"]
+        # Even with the right survivor, dropping the losers wholesale loses
+        # whatever fields only they carried. Fill the survivor's absent
+        # fields from each loser — the same never-override merge the
+        # same-source collision path already applies (#2091/#3372); loser
+        # order follows unique_nodes order, so the fill is deterministic.
+        merged = winner
+        for node in group_nodes:
+            if node["id"] != winner_id:
+                merged = _merge_missing_attributes(merged, node)
+        if merged != winner:
+            enriched_by_id[winner_id] = merged
         for member in members:
             if member != winner_id:
                 remap[member] = winner_id
@@ -892,7 +905,9 @@ def deduplicate_entities(
     if hyperedges:
         _remap_hyperedge_members(hyperedges, remap)
 
-    deduped_nodes = [n for n in unique_nodes if n["id"] not in remap]
+    deduped_nodes = [
+        enriched_by_id.get(n["id"], n) for n in unique_nodes if n["id"] not in remap
+    ]
     deduped_edges = []
     for edge in edges:
         e = dict(edge)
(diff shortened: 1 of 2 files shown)


abhay-codes07:
On the advisory: the `_collision_rank`/`_pick_winner` relationship is unchanged by this PR — `_collision_rank` resolves same-ID collisions when building `seen_ids`, and `_pick_winner` only ever runs over members of `unique_nodes` (the post-collision survivors), so every remap target is a node that survives into `deduped_nodes`; edges cannot rewire to a dropped id through either path. This PR only reorders `_pick_winner`'s scoring within the same candidate set (richness before id length) and back-fills the survivor, so any disagreement between the two rankings would predate it and still resolve to a live node.

safishamsi:
Shipped in **v0.9.57** — landed on `v8` via cherry-pick with your authorship preserved. Thanks @abhay-codes07 — when duplicates merge, the richer node is kept as survivor and the losers' non-empty fields are folded in. Release: https://github.com/Graphify-Labs/graphify/releases/tag/v0.9.57

Issue #2072, opened 2026-07-21
Scan root silently changes the graph: src-layout project loses most 'imports' edges, same node count, no warning


Dao-LQ-AI:
The same files produce a different graph depending on which directory you point `graphify` at. Node count is identical, so nothing looks wrong, and no warning is printed.

For a src-layout project, scanning the repository root instead of the package root silently drops most `imports` edges. In my real case that was 219 edges out of 2,236, about 10%.

## Minimal reproduction

Four files in a package `mypkg`, using absolute imports (`from mypkg.core import Engine`), which is what a src-layout project looks like on disk.

```
mre/
  direct/            <- scan THIS: package root
    mypkg/
      __init__.py    from mypkg.core import Engine
      core.py        class Engine
      app.py         from mypkg.core import Engine; import mypkg.helpers as helpers
      helpers.py     from mypkg.core import Engine
  nested/
    src/             <- identical bytes, one level deeper
      mypkg/ ...
```

`nested/src/` is a byte-identical copy of `direct/`, verified with `diff -rq`.

Run the same command in each directory, graphify 0.9.22, no API key set:

```
graphify . --code-only --no-label --no-viz
```

Result:

```
direct   ->  8 nodes, 14 edges, 3 communities
nested   ->  8 nodes, 10 edges, 4 communities
```

Edge relations:

| relation | direct | nested | delta |
|---|---|---|---|
| imports | 4 | 1 | **-3** |
| re_exports | 1 | 0 | **-1** |
| calls | 3 | 2 | -1 |
| imports_from | 2 | 3 | +1 |
| contains | 3 | 3 | 0 |
| method | 1 | 1 | 0 |

Same signature at real-project scale: 34 files, 927 nodes in both cases, 2,236 edges vs 2,016, with the gap being 197 `imports`, 17 `imports_from` and 5 `re_exports`. Every other relation matched exactly, and `calls` differed by one.

## Why this is easy to hit and hard to notice

1. Pointing at a repository root is the natural first thing to do, and it's what the README's `graphify .` implies.
2. The node count is unchanged, so the graph looks complete.
3. No warning is emitted. I grepped the degraded run's output for `warn`, `unresolved`, `could not` and `skip` and got zero lines.
[...]

PR #3429, opened 2026-09-09
fix(python): resolve submodule imports inside PEP 420 namespace packages

flaukowski:
## What

`from . import brain, ledger` inside a directory with **no `__init__.py`** — a PEP 420 namespace package, which `python -m pkg.mod` runs without complaint — produces **no edges at all**. `_resolve_python_module_path` returns `None` when there is no module file to probe, and the #1146 submodule-import branch in `_collect_python_symbol_resolution_facts` sits behind that `None`, so the whole statement is skipped. Every `brain.think()` / `ledger.write()` call in such a repo is therefore invisible to the #1883 module-call arm, and the most-called functions carry in-degree 0.

With an `__init__.py` present the same lines resolve correctly (checked the relative, absolute, aliased, parent-relative and subpackage spellings), so this is specific to namespace packages.

## Fix

- `_resolve_python_namespace_dir` mirrors the module-path walk (relative base → scan root → sys.path-root ancestors, per #2072) and returns only a directory that exists inside the scan root and has no `__init__.py`.
- `_collect_python_symbol_resolution_facts` uses it when the module path resolves to no file; the existing submodule probe then emits the same `imports_from` (`submodule_import`) edges a regular package gets, and the module arm resolves the calls.
- A name that is not a module file on disk still emits nothing, and a namespace package binds no symbols of its own (no `_SymbolImportFact` fallback), so nothing is fabricated.

## Evidence

Same graphify version (0.9.56), with and without the change, on a 275-node Python repo laid out this way:

| | `calls` edges | `think()` callers | `remember()` | ledger `write()` |
|---|---|---|---|---|
| unpatched | 201 | 0 | 0 | 0 |
| patched | 307 | 15 | 10 | 20 |

Node count unchanged (275 → 275).

## Tests

Three new tests in `tests/test_extract.py`, next to the #2082 tests: relative form (`from . import a, b` → two EXTRACTED `calls` edges plus the `imports_from` edges), absolute + parent-relative forms (`from pkg import a`; `from .. import a` from a nested namespace subpackage), and a negative (`from . import data, missing` where `data/` holds no module file → no resolved edge, no raise). The two positive tests fail on `v8` without the fix. `pytest tests/test_extract.py -k python`: 21 → 24 passed; the rest of the suite is unchanged.

No CHANGELOG entry, following the pattern of other fix PRs — happy to add one if you prefer contributors to.

https://claude.ai/code/session_01KTNEBgonu6ASph6w8R2Wcp

### DIFF
--- a/graphify/extractors/resolution.py
+++ b/graphify/extractors/resolution.py
@@ -2036,6 +2036,49 @@ def _resolve_python_module_path(module_name: str, current_path: Path, root: Path
             return cand
     return None
 
+def _resolve_python_namespace_dir(module_name: str, current_path: Path, root: Path, level: int) -> "Path | None":
+    """The directory a ``from <module> import ...`` names when that module is a
+    PEP 420 namespace package: a directory under the scan root with no
+    ``__init__.py``. ``_resolve_python_module_path`` returns None for it (there is
+    no module file to probe), so a package that omits ``__init__.py`` -- which
+    ``python -m pkg.mod`` runs without complaint -- had every ``from . import
+    sibling`` dropped whole, and with it every ``sibling.func()`` call the #1883
+    module arm would otherwise have resolved: the most-called functions in such a
+    repo carried in-degree 0. Mirrors that resolver's walk (relative base, then
+    scan root, then sys.path-root ancestors) and returns only a directory that
+    exists inside the root."""
+    def _namespace(candidate: Path) -> "Path | None":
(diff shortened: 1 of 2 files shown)


safishamsi:
Shipped in **v0.9.58** — landed on `v8` via cherry-pick with your authorship preserved. Thanks @flaukowski — submodule imports inside a PEP 420 namespace package now resolve. Release: https://github.com/Graphify-Labs/graphify/releases/tag/v0.9.58

PR #3431, opened 2026-09-09
Fix loose sibling module import resolution

### LINKED ISSUE #3430: Python: import-module + module.func() calls silently dropped outside package roots (module-arm resolver never matches, dangling import-edge id)
## Summary

`import module` + `module.func()` calls (as opposed to `from module import func`) silently produce no `calls` edge whenever the importing file lives in a directory one or more levels below the scan root **and that directory has no `__init__.py`** — i.e. any flat/namespace-style module layout, which is extremely common (a `scripts/`, `src/`, or tool directory of sibling `.py` files with no package markers). The qualified call is dropped with no warning; the callee node simply has in-degree 0, indistinguishable from genuine dead code.

This is a real regression in an otherwise-working feature, not a "we don't support this" limitation — `_resolve_python_member_calls`'s module arm (`extract.py`, PR-referenced as `#1883` in its comments) is specifically designed to resolve exactly this pattern. It's an ID-generation mismatch upstream of it that starves it of a match.

## Root cause

1. **File nodes get a directory-prefixed id.** `_file_node_id()` (`extract.py:187`) builds a file's own node id as `{parent_dir}_{stem}` — one parent directory level, e.g. `scripts/greeter.py` → `scripts_greeter`.

2. **Plain `import` edges don't.** `_import_python()`'s `import_statement` branch (`extract.py:325-348`) builds the import edge's target id from the bare module name alone:
   ```python
   module_name = raw_module.strip().lstrip(".")
   tgt_nid = _make_id(module_name)  # "greeter" -> "greeter", NOT "scripts_greeter"
   ```
   No filesystem lookup, no directory context — `make_id()` (`ids.py:86`) just normalizes the literal string. So for `import greeter` inside `scripts/main.py`, the edge becomes `scripts_main --imports--> greeter`, while the actual file node is `scripts_greeter`. They never match.

3. **The module-arm resolver depends on that match.** `_resolve_python_member_calls()` (`extract.py:3310`), when it sees a qualified call `greeter.say_hello()`, looks up the caller file's imported targets and keeps only the ones present in `contains_children` — i.e. the ones that are *real, resolvable file nodes*:
   ```python
   mods = [t for t in imported_by_filenode.get(caller_file, ())
           if t in contains_children
           and (_module_stem_key(t) == rkey or file_aliases.get(t) == rkey)]
   if len(mods) != 1:
       continue
   ```
   Since the import edge's target is `greeter` and the real file node is `scripts_greeter`, `t in contains_children` is `False` for every candidate, `mods` ends up empty, and the resolver bails silently. No edge is emitted anywhere — `symbol_resolution.py`'s fallback resolvers also explicitly skip member calls (`is_member_call` guard, by design, since a bare receiver-less method name is ambiguous), so there is no second chance to catch it.

4. **A fix for the analogous case already exists, but its guard excludes this shape.** `_repoint_python_package_imports()` (`extract.py:197`, referenced as `#2072`) was added to repoint exactly this kind of dangling import-edge id — but only for real Python *packages*:
   ```python
   while levels < len(parts) - 1 and (d / "__init__.py").is_file():
       levels += 1
       d = d.parent
   if levels == 0:
       continue  # not inside a package (namespace pkg / loose module)
   ```
   A directory with no `__init__.py` — a plain script folder, or a PEP 420 namespace package, both common — takes `levels == 0` and is skipped entirely. The repoint logic that would fix this never runs for it.

## Minimal reproduction

```
repo/
└── scripts/
    ├── greeter.py
    └── main.py
```

`scripts/greeter.py`:
```python
def say_hello(name: str) -> str:
    return f"Hello, {name}!"
```

`scripts/main.py`:
```python
import greeter

def run() -> None:
    print(greeter.say_hello("world"))
```

Running `graphify.extract.extract()` from `repo/` (i.e. `repo/` is the scan root, `scripts/` is not a package — no `__init__.py`) on both files produces:

```
nodes: scripts_greeter, scripts_greeter_say_hello, scripts_main, scripts_main_run
edges:
  scripts_greeter -- contains --> scripts_greeter_say_hello
  scripts_main -- imports --> greeter          # <- dangling: no node has this id
  scripts_main -- contains --> scripts_main_run
```

**Expected:** a fourth edge, `scripts_main_run -- calls --> scripts_greeter_say_hello`.
**Actual:** it's absent. `scripts_greeter_say_hello` has in-degree 0.

Confirmed this is scan-root-relative, not something inherent to member calls in general: running `extract()` with `scripts/` itself passed as the scan root (so both files sit at the "top level" and get unprefixed ids `greeter`/`main`) produces the `calls` edge correctly — same two files, same call, only the root changed.

## Impact

Any codebase organized as a flat directory of sibling scripts (no `__init__.py`) that prefers `import module` + `module.func()` over `from module import func` — which is arguably the more idiomatic form, since it keeps call sites self-documenting and avoids polluting the caller's namespace — will have its cross-file call edges for that directory systematically dropped whenever the tool is pointed at the directory's *parent* rather than the directory itself. Since graphify's dead-code/god-node/community-detection output is exactly what a user leans on to *find* unused code, this produces false-positive "orphan" signals for modules that are, in fact, actively called — the more central and widely-used a module is (a shared `utils`/`paths`/`config`-style module imported by many siblings), the worse the apparent disconnect looks, which is the opposite of what such a module's graph position should show.

## Suggested fix

Extend `_repoint_python_package_imports()` (or add a sibling pass using the same alias-map machinery) to also cover the no-`__init__.py` case: when a bare `import X` edge's target `make_id(X)` doesn't match any node id, but the importing file's own directory contains a sibling `X.py` that *is* in the scanned corpus, repoint the edge to that sibling's real file-node id. This can stay conservative — same directory only, single unambiguous candidate only — mirroring the existing pass's `len(fs) == 1` guard, so it wouldn't need to solve general `sys.path` resolution, just the common "importing a sibling in the same loose directory" case that today falls through the `levels == 0` gap.

## Environment

- `graphifyy` 0.9.56 (PyPI)
- Windows 11, Python 3.13.6
- Reproduced from scratch on the two-file layout above — not dependent on any project-specific configuration, cache, or prior state.

hopstreax:
## Summary

Fixes #3430.

Plain Python imports such as `import greeter` followed by `greeter.greet()` were silently losing their `calls` edges when the importing file lived below the scan root in a directory without `__init__.py`.

The import edge targeted the bare module ID (`greeter`) while the extracted file node used the scan-root-relative ID (`scripts_greeter`). Because the package import repointing pass intentionally skips non-package directories, the member-call resolver could not match the imported module.

## What changed

* Added a directory-local `_repoint_python_sibling_imports()` pass.
* Resolves unambiguous sibling `.py`/`.pyi` modules within the importing file's directory.
* Restricts candidates to files in the extracted corpus.
* Preserves package isolation and avoids global basename aliasing.
* Preserves `local_alias` for aliased imports such as `import greeter as g`.
* Runs the sibling repoint before Python member-call resolution.

## Tests

* Loose sibling `import module` + `module.func()` resolution.
* Scan-root parity.
* Same-named modules in separate loose directories.
* Package/PEP 328 isolation.
* Cross-directory leakage prevention.
* Aliased sibling imports.

Verified with:

* `pytest tests/test_loose_sibling_import_resolution.py`
* `pytest tests/test_src_layout_import_resolution.py tests/test_python_import_resolution.py`
* `pytest tests/test_language_resolvers.py -k "python"`

All tests pass.

### DIFF
--- a/graphify/extract.py
+++ b/graphify/extract.py
@@ -266,6 +266,111 @@ def _repoint_python_package_imports(paths, all_nodes, all_edges, root) -> None:
                 e["target"] = alias_map[tgt]
 
 
+def _repoint_python_sibling_imports(paths, all_nodes, all_edges, root) -> None:
+    """Repoint Python sibling-import edges to the real file node in directories
+    without an __init__.py (#3430).
+
+    When a Python file below the scan root lives in a non-package directory
+    (no __init__.py in its immediate parent), plain imports of same-directory
+    modules (e.g. `scripts/main.py: import greeter`) target a bare name (`greeter`),
+    while the real file node is scan-root-relative (`scripts_greeter`). Because
+    the directory is not a package, `_repoint_python_package_imports` skips it
+    (levels == 0), leaving the edge dangling and causing downstream member calls
+    (`greeter.greet()`) to be dropped.
+
+    This pass is strictly importer-directory-local:
+    - Only applies when the importing file's parent directory has no __init__.py.
+    - Resolves only to unambiguous same-directory candidate modules in the scanned corpus.
+    - Never builds a global alias map and never searches outside the importer's directory.
+    - Preserves local aliases (e.g. `import greeter as g`).
+    """
+    try:
+        root = Path(root).resolve()
+    except OSError:
+        root = Path(root)
+
+    node_ids = {n.get("id") for n in all_nodes if isinstance(n, dict)}
+
+    # Index corpus modules by directory for non-package directories only.
+    # dir_path -> {module_name_id: set_of_file_node_ids}
+    dir_siblings: dict[Path, dict[str, set[str]]] = {}
+    for p in paths:
+        if p.suffix.lower() not in (".py", ".pyi"):
+            continue
+        try:
+            p_res = Path(p).resolve()
+            rel = p_res.relative_to(root)
+        except (ValueError, OSError):
+            continue
+
+        file_node = _file_node_id(rel)
+        if file_node not in node_ids:
+            continue
+
+        if p_res.name in ("__init__.py", "__init__.pyi"):
+            # Sibling package directory inside parent_dir (parent_dir / subpkg / __init__.py)
+            pkg_dir = p_res.parent
+            parent_dir = pkg_dir.parent
+            if not (parent_dir / "__init__.py").is_file() and not (parent_dir / "__init__.pyi").is_file():
+                mod_key = _make_id(pkg_dir.name)
+                dir_siblings.setdefault(parent_dir, {}).setdefault(mod_key, set()).add(file_node)
+            continue
+
+        d = p_res.parent
+        # PEP 328 guard: if the directory is a package, implicit relative imports
+        # are forbidden in Python 3. Do not index packages as loose sibling directories.
+        if (d / "__init__.py").is_file() or (d / "__init__.pyi").is_file():
+            continue
+
+        mod_key = _make_id(p_res.stem)
+        dir_siblings.setdefault(d, {}).setdefault(mod_key, set()).add(file_node)
+
+    # Require an unambiguous single candidate per module name in that directory.
+    dir_alias_map: dict[Path, dict[str, str]] = {
+        d: {
+            mod: next(iter(fns))
+            for mod, fns in mod_map.items()
+            if len(fns) == 1
+        }
+        for d, mod_map in dir_siblings.items()
+    }
+
+    if not dir_alias_map:
+        return
+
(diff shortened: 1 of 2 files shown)


safishamsi:
Shipped in **v0.9.58** — landed on `v8` via cherry-pick with your authorship preserved. Thanks @hopstreax — a bare-name import of a sibling module in a flat script dir now resolves (matching sys.path[0]). Release: https://github.com/Graphify-Labs/graphify/releases/tag/v0.9.58

PR #3432, opened 2026-09-09
Prefer the alias for a PHP use import edge's target

### LINKED ISSUE #3421: PHP: use A\B\C as D gives the import edge a target derived from the last segment, splitting the class identity
graphify 0.9.56

## What happens

For an aliased `use`, the `imports` edge target is derived from the last segment of the
imported class name, lowercased, instead of the fully qualified id that the class's own
node carries. A plain `use` of the same class in the same project resolves correctly, so
the alias is the trigger.

## Repro

```php
// src/A.php
<?php
namespace App;

use GuzzleHttp\Client;

class A
{
    public function __construct(private Client $client) {}
}
```

```php
// src/B.php
<?php
namespace App;

use GuzzleHttp\Client as HttpClient;

class B
{
    public function __construct(private HttpClient $client) {}
}
```

```bash
```

## Observed

```
nodes: client, guzzlehttp_client, src_a, src_a_a, src_a_a_construct,
       src_b, src_b_b, src_b_b_construct

src_a --imports--> guzzlehttp_client
src_b --imports--> client
```

`GuzzleHttp\Client` ends up as two nodes. `src_a` and `src_b` import the same class but no
longer share a target, so neither one can be reached from the other's traversal.

Expected: both edges point at `guzzlehttp_client`. An alias is a file-local binding; it
should not create a second identity for the imported symbol.

## The same defect on a real codebase, with a different symptom

On a ~1400-file Laravel project the bare-name node is not created at all, so the edges
dangle instead of splitting: **103 of 145** `dangling_endpoint_edges` reported by
`graphify diagnose multigraph` come from aliased `use` lines. The project has 142 aliased
`use` statements over 37 distinct classes, and the most frequent unresolved targets are
`session` (44 edges), `client` (21) and `connection` (5) — precisely the classes a project
aliases to disambiguate, e.g. `App\Models\Session` against `Shopify\Auth\Session`.

I have not worked out what decides between the two symptoms; the minimal repro above is
the one I can reproduce on demand.

## Why it matters

`graphify affected "<Class>"` under-reports. A consumer importing the class under an alias
does not appear in the reverse walk, and nothing signals that anything was missed — the
answer is simply smaller than the truth.

## Possibly related

- #2812 (closed) — `extends Exception` resolving onto an unrelated same-named class. Same
  root shape, identity taken from an unqualified name; that fix does not appear to cover
  `use ... as ...`.
- #3346 (open) — TypeScript aliased/namespace imports producing no `calls` edge.

ayushcodes10:
Fixes #3421.

`_import_php` always targeted the bare imported class name (the last segment of the qualified name), completely ignoring a `use ... as ...` alias. Two files importing the same external class under different local aliases — the exact real-world case an alias exists for, disambiguating two same-named classes from different namespaces — produced two different stub targets for one class, splitting its identity in the graph.

Root cause traced through two interacting pieces:
- `_import_php` (in `extract.py`) always used the bare imported name (e.g. `Client`) as the provisional stub target, regardless of any alias.
- `_resolve_php_type_references` (in `resolution.py`) re-points that stub to the correct external FQN stub by looking up the stub's own bare label in a per-file alias -> FQN map, which is keyed by the *alias* when the import has one. For a plain (non-aliased) import the bare label happens to match that key, so it resolves correctly. For an aliased import the bare label ("Client") never matches the key ("httpclient"), so the lookup misses and the edge stays stuck on the unresolved bare-name stub.

Confirmed against the exact repro from the issue:
```
NODE client | Client |
NODE guzzlehttp_client | GuzzleHttp\Client |
EDGE src_a -> guzzlehttp_client imports
EDGE src_b -> client imports        # wrong: same class, different target
```
After the fix, both files' `imports` edges (and their constructor-parameter `references` edges) consistently point at the single `guzzlehttp_client` node.

Also verified:
- A grouped `use` with a mixed aliased/non-aliased clause (`use GuzzleHttp\{Client, Exception as GuzzleException};`) resolves both parts correctly.
- Two *different* classes aliased specifically to avoid a bare-name collision (`App\Models\Session as ModelSession` vs `Shopify\Auth\Session as ShopifySession`) still resolve to two distinct nodes — the fix does not over-collapse.
- `use function`/`use const` clauses are unaffected (pre-existing behavior, out of scope for this issue).

## Test plan
- [x] Added `test_php_aliased_import_edge_shares_target_with_plain_import` and `test_php_distinct_aliases_for_same_named_classes_do_not_collapse` to `tests/test_php_type_resolution.py`
- [x] `pytest tests/test_php_type_resolution.py` and full `pytest tests/ -k php` pass
- [x] Full suite passes except pre-existing, unrelated failures (`test_ollama_retry_cap.py`, missing `openai` module in this environment; `test_ts_normalizer_scales_linearly_on_large_files`, a flaky wall-clock timing assertion, confirmed to pass in isolation)
- [x] `python -m tools.skillgen --check` passes

### DIFF
--- a/graphify/extract.py
+++ b/graphify/extract.py
@@ -728,23 +728,46 @@ def _import_scala(node, source: bytes, file_nid: str, stem: str, edges: list, st
 
 
 def _import_php(node, source: bytes, file_nid: str, stem: str, edges: list, str_path: str, scope_stack: list[str] | None = None) -> None:
+    # `node` is a namespace_use_clause: `Client` or `Client as HttpClient`,
+    # whose children are the qualified/bare name, then (if present) `as`
+    # and the alias name, in that order. An aliased clause used to always
+    # target the bare imported name ("Client"), ignoring the alias
+    # entirely -- so two files importing the SAME external class under
+    # DIFFERENT local aliases (a real pattern: disambiguating two
+    # same-named classes from different namespaces) produced two
+    # DIFFERENT stub targets for one class, and _resolve_php_type_references
+    # (which resolves a stub via the file's own alias -> FQN map, keyed by
+    # the alias when the import has one) could never find this one under
+    # its bare-name-derived key, leaving the edge stuck on an unresolved,
+    # never-repointed stub (#3421). The alias, when present, is also what
+    # the rest of the file actually references (a parameter typed
+    # `HttpClient`, not `Client`), so preferring it here keeps this edge's
+    # target consistent with those reference edges too.
+    saw_as = False
+    module_name = None
     for child in node.children:
+        if child.type == "as":
+            saw_as = True
+            continue
         if child.type in ("qualified_name", "name", "identifier"):
-            raw = _read_text(child, source)
-            module_name = raw.split("\\")[-1].strip()
-            if module_name:
-                tgt_nid = _make_id(module_name)
-                edges.append({
-                    "source": file_nid,
-                    "target": tgt_nid,
-                    "relation": "imports",
-                    "context": "import",
-                    "confidence": "EXTRACTED",
-                    "source_file": str_path,
-                    "source_location": f"L{node.start_point[0] + 1}",
-                    "weight": 1.0,
-                })
-            break
+            bare = _read_text(child, source).split("\\")[-1].strip()
+            if saw_as:
+                module_name = bare
+                break  # the alias name always comes last; nothing more to see
+            if module_name is None:
+                module_name = bare
+    if module_name:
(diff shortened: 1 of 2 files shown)


safishamsi:
Shipped in **v0.9.58** — landed on `v8` via cherry-pick with your authorship preserved. Thanks @ayushcodes10 — a PHP `use ... as` import keeps its aliased class identity. Release: https://github.com/Graphify-Labs/graphify/releases/tag/v0.9.58

PR #3434, opened 2026-09-09
perf(global-graph): global_add_many so a batch pays the whole-graph cost once

### LINKED ISSUE #3433: global add is quadratic in the number of units: K whole-graph load/pass/save cycles
## Summary

`global_add` does three whole-graph operations per call — `_load_global_graph`,
`link_cross_repo_member_calls`, `_save_global_graph` (`graphify/global_graph.py`). Each one runs
over the *whole* global graph, which grows with every unit registered, so registering K units costs
O(K²) work: the K-th call pays for the K−1 units already in there.

Only the composition of a unit (prune that repo, dedup its external stubs, add its nodes and edges)
needs the graph in the state that unit arrives at. The read, the cross-repo pass and the write
depend on the batch as a whole, so registering a batch could pay them once.

## Measured

Synthetic units of fixed size (1501 nodes / ~2500 edges each), K varied. Wall clock, plus
`tracemalloc` peak (Python allocations, not RSS):

| K | `global_add` ×K | one batch | speedup |
|--:|--:|--:|--:|
| 1 | 0.14 s | 0.12 s | 1.2× |
| 5 | 1.67 s | 0.66 s | 2.5× |
| 10 | 6.05 s | 1.33 s | 4.6× |
| 20 | 22.67 s | 2.86 s | 7.9× |
| 40 | 89.52 s | 5.89 s | 15.2× |

Sequential time roughly quadruples when K doubles (22.67 → 89.52 s); a batch roughly doubles
(2.86 → 5.89 s). The speedup at a given K is about K/2 rather than K, because the early units are
cheap while the graph is still small.

It is not only a first-build effect. Adding 20 units to a store that already holds 20:
64.81 s sequential vs 4.92 s batched. Peak memory did not increase in either scenario (184.5 → 117.7 MB
at K=40; 183.4 → 123.6 MB incremental).

Real corpus for scale: 42 units, ~540k nodes / ~920k edges.

## Why there is no way to do this today

- CLI `graphify global add <graph.json> [--as <tag>]` takes exactly one source; extra positional
  arguments are silently dropped (`cli.py`).
- `_load_global_graph` / `_save_global_graph` have no cache, no incremental path, no append.
- No switch skips the cross-repo pass (no `GRAPHIFY_*CROSS*`, `--no-cross` or `skip_cross`
  anywhere in the tree).
- `merge-graphs` *is* one-shot, and does run each cross-repo pass once at the end — but it has no
  prune, no manifest and no hash-skip, so it cannot stand in for `global add`.

So there is an incremental-but-per-unit path and a batch-but-not-incremental one, and nothing that
is both.

## Proposed

`global_add_many(sources: Sequence[tuple[Path, str]]) -> dict` composing an ordered batch against
one in-memory graph, with `global_add` delegating to it as a batch of one so there is a single
composition implementation.

Composition order has to stay as it is: prune per unit before that unit is composed, and the
external-label dedup read off the graph as it stands when the unit arrives, so a later unit merges
onto a stub an earlier one composed. Only the cross-repo pass moves out — it drops its own previous
output and recomputes from the parked entries, which is exactly what already makes K sequential
adds agree with one `merge-graphs` of the same inputs.

PR follows, with an equivalence test comparing sequential and batched runs on node set, every node
attribute, edge set, every edge attribute, and the manifest.

xiongjianxu:
Fixes #3433

## Summary

Registering K project graphs with `global add` costs O(K²) work, because each call does three
whole-graph operations — `_load_global_graph`, `link_cross_repo_member_calls`,
`_save_global_graph` — on a graph that grows with every unit. The K-th call pays for the K−1
units already in there.

Only the composition of a unit needs the graph in the state that unit arrives at. The read, the
cross-repo pass and the write depend on the batch as a whole, so a batch API can pay them once.

## Measured

Synthetic units, 1501 nodes / ~2500 edges each, fixed size, K varied. Wall clock, and
`tracemalloc` peak (Python allocations, not RSS):

| K | `global_add` ×K | `global_add_many` | speedup | seq peak MB | batch peak MB | final nodes |
|--:|--:|--:|--:|--:|--:|--:|
| 1 | 0.14 s | 0.12 s | 1.2× | 6.8 | 5.9 | 1 501 |
| 5 | 1.67 s | 0.66 s | 2.5× | 27.8 | 23.5 | 7 501 |
| 10 | 6.05 s | 1.33 s | 4.6× | 54.4 | 36.7 | 15 001 |
| 20 | 22.67 s | 2.86 s | 7.9× | 90.5 | 63.5 | 30 001 |
| 40 | 89.52 s | 5.89 s | 15.2× | 184.5 | 117.7 | 60 001 |

Sequential time roughly quadruples when K doubles (22.67 → 89.52); the batch roughly doubles
(2.86 → 5.89). That is the point of the change: a quadratic term becomes linear. The speedup at a
given K is therefore about K/2, not K — the early units are cheap because the graph is still
small.

Not a from-scratch-only effect. Adding 20 units to a store that already holds 20:

| | wall clock | peak MB |
|--|--:|--:|
| `global_add` ×20 | 64.81 s | 183.4 |
| `global_add_many` | 4.92 s | 123.6 |

Peak memory does not go up — it measured lower in both scenarios. I am reporting that as an
observation rather than a claim about the mechanism; the reason to take this change is the time.

## API

```python
def global_add_many(sources: Sequence[tuple[Path, str]]) -> dict:
    """-> {"results": [{repo_tag, nodes_added, nodes_removed, skipped}, ...],
           "cross_repo_calls": int}"""
```

One result per input, in input order, including skipped units. `cross_repo_calls` sits at the
batch level: the pass reads the finished graph, so the number does not divide over the units that
produced it.

`global_add` now delegates with a batch of one. Its signature, its return keys and its callers are
unchanged, and there is one composition implementation instead of two that have to be kept in
step.

## What had to stay in the loop

- **Prune.** Still per unit, before that unit is composed. It only removes that repo's nodes.
- **The external-label dedup.** Still read off the graph *as it stands when the unit arrives*, so
  a later unit's stub merges onto one an earlier unit in the batch composed. Hoisting it changes
  the graph — the equivalence test below fails if you do.
- **The path-change warning.** Still one per unit.

The cross-repo pass is the one thing that moves out: it drops its own previous output and
recomputes from the parked entries, which is what already makes K sequential adds agree with a
single `merge-graphs` (`global_graph.py` comment, `cross_repo_calls.link_cross_repo_member_calls`
docstring).

## Batch-specific behaviour, all tested

- **An unchanged batch never reads the graph.** Discovering there is nothing to do would otherwise
  cost a full deserialize plus a full rewrite of identical bytes — a regression on the most common
  re-run.
- **Missing source, oversized source, and one tag naming two sources are all rejected before the
  first unit is composed.** The batch writes only at the end, so a unit rejected late discards
  every unit composed before it. The duplicate tag is a new error: the second source would prune
  the first, and the batch would report having added both.
- A missing source raises `FileNotFoundError("graph not found: ...")`, the same wording
  `global_add` used, rather than the OS message hashing it would produce.

## Tradeoff to be aware of

Sequential adds checkpoint: unit 20 of 42 failing leaves units 1–19 durably registered, and a
re-run skips them by hash. A batch is all-or-nothing — it writes nothing on failure, so nothing is
half-merged, but nothing is banked either. Callers that want checkpointing can batch in chunks.

## Tests

`tests/test_global_graph.py`, nine added, no existing assertion changed (29 passed, was 20):

- `test_global_add_many_matches_one_add_per_unit` — the acceptance criterion. Three units, one of
  which parks a member call another answers, plus a stub two units share. Sequential and batch are
  compared on node set, every node attribute, edge set, every edge attribute, and the manifest
  (minus `added_at`). It also asserts the pass actually resolved something, so a fixture that
  stopped exercising it cannot pass silently.
- `test_global_add_many_dedups_a_later_units_external_against_an_earlier_one` — fails if the dedup
  is hoisted out of the loop.
- `test_global_add_many_runs_the_cross_repo_pass_once` — spy, `call_count == 1`.
- `test_global_add_many_keeps_a_skipped_unit_in_position` — a hash-skipped unit keeps its slot in
  `results` and is not pruned and re-composed.
- `test_global_add_many_leaves_the_store_untouched_when_nothing_changed` — `_load_global_graph`
  not called, file mtime unchanged.
- `test_global_add_many_rejects_one_tag_naming_two_sources`,
  `test_global_add_many_reports_a_missing_source_the_way_one_add_does`,
  `test_global_add_many_checks_every_size_cap_before_composing_any` — each asserts
  `prefix_graph_for_global` was never called, and each fails if the corresponding check is moved
  into the compose loop.
- `test_global_add_many_reports_per_unit_and_batch_level_counts` — return shape.

Full suite: 5360 passed, 94 skipped. `tests/test_ts_import_type_arguments.py::test_ts_normalizer_scales_linearly_on_large_files`
fails identically on unmodified `v8` in a full run and passes in isolation — a machine-dependent
`process_time` ratio, unrelated. `ruff check` clean.

No timing assertion is added to the suite. The numbers above come from a standalone script; a
wall-clock ratio in CI is the kind of test that fails on a loaded runner.

## Not in this PR

- **CLI.** `graphify global add` still takes one source. Extending it to several positional
  sources (with `--as` valid only for exactly one) is a small follow-up; callers can use the
  Python API meanwhile.
- **The external-stub dedup semantics.** Unchanged here on purpose — equivalence with the
  sequential path is what this PR is verified against, so it has to keep whatever that path does.
  #3414 / #3415 argue that dedup is wrong and remove it. If #3415 lands first, this PR gets
  simpler: composition becomes order-independent, the per-unit dedup rebuild goes away, and
  `test_global_add_many_dedups_a_later_units_external_against_an_earlier_one` should be dropped
  rather than reversed. If this one lands first, #3415 needs to touch `global_add_many` too. Happy
  to rebase in whichever order you prefer.

### DIFF
--- a/graphify/global_graph.py
+++ b/graphify/global_graph.py
@@ -2,6 +2,7 @@ from __future__ import annotations
 import json
 import hashlib
 import sys
+from collections.abc import Sequence
 from datetime import datetime, timezone
 from pathlib import Path
 import networkx as nx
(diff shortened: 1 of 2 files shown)


xiongjianxu:
Thanks — the gate is PASS and CI is green on both 3.10 and 3.12. On the four coupling findings, three are pre-existing functions the file-level re-analysis re-surfaced, and one is real but is a relocation rather than an increase.

**`global_add_many()` — 9 callees.** This is `global_add`'s fan-out, moved. The callee set is identical to the one `global_add` had on `v8`:

```
_file_hash, _load_global_graph, _load_manifest, _save_global_graph, _save_manifest,
check_graph_file_size_cap, link_cross_repo_member_calls, prefix_graph_for_global,
prune_repo_from_graph
```

`global_add` is now a 1-callee wrapper over it, so the total across the two functions went from 9 to 10 — the delegation edge. That edge is the point of the change: one composition implementation means a fix to the merge semantics cannot land in the batch path and miss the single-unit path, or the reverse. Splitting the body to lower the number would trade a metric for the duplication the delegation exists to prevent.

The 11 afferent edges are `global_add` plus the tests; `global_add` reported 13 on `v8`, so nothing new depends on this file from outside it. The public surface is unchanged — same signature, same return keys, same three call sites in `cli.py`.

**`global_remove()` — Ca·Ce = 25**, **`dispatch_command()` — 124 callees**, **`test_poisoned_manifest_is_healed()` — 0 callers.** None of these are touched. `git diff v8` has zero `+`/`-` lines mentioning `global_remove`, `dispatch_command` lives in `cli.py` which this PR does not modify at all, and `test_poisoned_manifest_is_healed` is an existing test that happens to sit in the same file as the added ones. Happy to be corrected if the delta is computed against something other than the diff.

**On the abstention.** `global_add` came back `unsupported` because `source_path: Path` is outside the synthesizable set — same reason as on `v8`, so the verifier's coverage of this function did not change either way. Worth noting the equivalence claim the abstention cannot reach is covered by a test rather than a proof: `test_global_add_many_matches_one_add_per_unit` runs the same three units through sequential adds and through one batch and compares node set, every node attribute, edge set, every edge attribute, and the manifest. I checked it is load-bearing — hoisting the external-label dedup out of the per-unit loop makes it fail.

No code changes from this round. Let me know if you would rather see the batch loop split into a helper anyway, or if the CLI half should come in the same PR.

xiongjianxu:
Correcting my own benchmark numbers in the description: the absolute wall-clock figures were measured
with `tracemalloc` active, which inflates them roughly 5× and does so unevenly — it charges per
allocation, and the sequential path allocates far more, so it overstated the batch's advantage in
absolute terms. Re-measured with time and peak memory in separate runs:

```
1501 nodes per unit
  K    seq s  batch s  speedup  seq peak MB  batch peak MB
  1     0.03     0.02     1.4x          5.9            5.9
  5     0.32     0.13     2.5x         26.9           24.7
 10     1.20     0.27     4.4x         46.3           34.9
 20     4.30     0.54     7.9x         93.4           58.7
 40    17.26     1.13    15.2x        183.6          119.0

 mode        s   peak MB  (20 units onto a store already holding 20)
  seq    12.77     184.5
batch     1.01     121.3
```

The speedup ratios and the peak-memory figures are unchanged; only the absolute seconds were wrong.

Also pushed `917a4f5`, which pins the one ordering in the compose loop that is load-bearing and was
untested: the external-label index has to be read off `G` *after* `prune_repo_from_graph`. Built
before it, the index still names a stub the prune deleted, the unit's own stub is treated as a
duplicate and never composed, and rewiring an edge onto the missing id has `add_edge` invent an
attribute-less node in its place. The test asserts the graph-integrity property (no node without a
`repo` attribute) rather than dedup behaviour, so it stays valid if #3415 removes the dedup block.

xiongjianxu:

Issue #3062, opened 2026-08-25
`graphify global add` memory scales with the size of the global store, not the size of the change


Azeem1985:
## Summary

`graphify global add` updates one repo inside the global graph, but it loads the **entire** `~/.graphify/global-graph.json` into NetworkX to do it (`global_graph.py::_load_global_graph`), prunes the repo, merges, and writes all of it back (`_save_global_graph`).

Peak memory therefore tracks the size of the **store**, not the size of the **change** — so the cost of a routine refresh grows without bound as more repos are tracked, even though the work never gets bigger.

## Measured

Real store: **226 repos, 1,211,189 nodes, 2,884,473 links, 1.888 GB**. Refreshing one repo of **26,668 nodes — 2.2% of the store**:

| | peak RSS | wall |
|---|---|---|
| current (whole-load) | **~5.0 GB** + ~0.9 GB swap | 3m49s |
| streaming slice rewrite | **186 MB** (190,460 kB, 0 swaps) | 3m57s |

**27x less memory for the same result, at +8s wall.** The box has 7.6 GB, so the current path was OOM-killed. Capping it with a memory guard just converts the crash into a store that silently stops refreshing — 21 hours with zero successful refreshes in our case.

## Why a slice is enough

`prefix_graph_for_global` already stamps every node with `repo` and rewrites its id to `<tag>::<local_id>`. A repo's slice is therefore exactly *"the nodes whose `repo` equals the tag, plus the links incident to them"* — addressable from the file with no format change and no extra index. On our real store, 0 of 1,211,189 nodes disagreed between the `repo` attribute and the id prefix.

And the reader for this shape already ships: `exporters/node_link_stream`, written so a DB push could walk a 1.9 GB graph in ~50 MB. Reusing it, the update becomes: stream the existing file element by element, copy every other repo's nodes and links through untouched, drop the target repo's, append the new slice, write to a temp file, fsync, atomic replace. Peak memory becomes O(changed repo + the global external-label index).

## Correctness

The interesting part is not the streaming, it's reproducing the whole-load path's merge semantics exactly:

- the external-library dedup that rewires an incoming edge onto an existing global node, and the self-loop drop that rewiring can cause;
- `add_edge`'s first-position/last-attributes collapse of duplicates — and the fact that `add_edge(u, v, **attr)` *updates* the existing attribute dict rather than replacing it, so attributes the old edge has and the incoming one doesn't must survive;
[...]

PR #3440, opened 2026-09-09
perf(global-graph): global_add_many, with partial-failure recovery

### LINKED ISSUE #3438: perf(global-graph): avoid repeated whole-graph reload/resolve/save in global add
<html>
<body>
<html><head></head><body><h2>Problem</h2><p><code inline="">global add</code> currently reloads, processes, and saves the entire global graph for <strong>every repository</strong> added.</p><p>For each invocation, the operation effectively does:</p><ol><li><p>Load the complete global graph</p></li><li><p>Prune the target repository</p></li><li><p>Merge the new repository graph</p></li><li><p>Re-run cross-repository linking/resolution</p></li><li><p>Serialize and save the complete global graph</p></li></ol><p>When adding <code inline="">K</code> repositories sequentially, this repeatedly processes an increasingly large graph, resulting in roughly <strong>O(K²)</strong> cumulative work.</p><h3>Performance evidence</h3><p>Reported benchmarks from issue #3433:</p>
Repositories | Sequential global add | Batched | Speedup
<table> <thead> <tr> <th>Repositories</th> <th>Sequential <code>global add</code></th> <th>Batched</th> <th>Speedup</th> </tr> </thead> <tbody> <tr> <td>5</td> <td>1.67s</td> <td>0.66s</td> <td>2.5×</td> </tr> <tr> <td>10</td> <td>6.05s</td> <td>1.33s</td> <td>4.5×</td> </tr> <tr> <td>20</td> <td>22.67s</td> <td>2.86s</td> <td>7.9×</td> </tr> <tr> <td>40</td> <td>89.52s</td> <td>5.89s</td> <td><strong>15.2×</strong></td> </tr> </tbody> </table>

<p>This becomes increasingly expensive for large global graphs.</p><p>Issue #3062 also reports significant memory pressure when updating large global stores because the complete graph is loaded and rewritten for each update.</p><h2>Proposed solution</h2><p>Add a batched API such as:</p><pre><code class="language-python">global_add_many(graph_paths, tags=None)
</code></pre><p>The implementation should:</p><ol><li><p>Load the global graph <strong>once</strong></p></li><li><p>Prune/replace all target repositories</p></li><li><p>Merge all incoming repository graphs</p></li><li><p>Run cross-repository resolution <strong>once</strong></p></li><li><p>Save the global graph <strong>once</strong></p></li></ol><p>The existing single-repository <code inline="">global_add()</code> can delegate to this implementation.</p><p>The CLI should also support adding multiple graph files in one invocation, where practical.</p><h2>Correctness requirements</h2><p>The batched implementation should produce the same resulting graph as performing the equivalent individual <code inline="">global add</code> operations sequentially.</p><p>Please add regression tests covering:</p><ul><li><p>Multiple repositories added in one batch</p></li><li><p>Existing repositories being replaced</p></li><li><p>Cross-repository links between newly added repositories</p></li><li><p>Tags/metadata</p></li><li><p>Empty and single-item batches</p></li><li><p>Equivalence between sequential and batched results</p></li></ul><h2>Scope</h2><p>This issue focuses on eliminating repeated <strong>load → resolve → save</strong> work.</p><p>A separate optimization can later investigate streaming/incremental global-graph updates to reduce peak memory usage without loading the entire graph.</p><h2>Expected result</h2><p>The goal is to make multi-repository global graph updates scale approximately with the amount of work performed once per batch rather than repeatedly reprocessing the growing global graph.</p></body></html>
</body>
</html>

Sagexd08:
Closes #3438.

## Overlap with #3434 — read this first

**#3434 already implements the core of this and it is good work.** It arrived first, it is reviewed, and its `global_add_many` is correct on the thing that is easiest to get wrong here (external-stub dedup ordering). I am not claiming otherwise, and if the maintainers prefer to land #3434 and take only the extras below as follow-ups, that is a completely reasonable outcome — I would rather the fix lands than that mine does.

I opened this because #3438 was filed separately and I had already built and measured an implementation before finding #3434. What is genuinely different is in **"Beyond #3434"**; the shared core is the same design, arrived at independently.

## Problem

`global add` costs O(K²) over K repos: every call reloads, re-resolves and re-serializes a global graph that grows with each repo, so the K-th call pays for the K−1 already in it.

Only *composing* a unit needs the graph in the state that unit arrives at. The read, the cross-repo call pass and the write depend on the batch as a whole, so a batch can pay each of them once.

## Measured

400-node units, wall clock, resulting graphs verified byte-identical to the sequential result at every K:

| K | sequential | batched | speedup |
|--:|--:|--:|--:|
| 5 | 0.33 s | 0.05 s | 7.3× |
| 10 | 0.73 s | 0.08 s | 8.8× |
| 20 | 2.36 s | 0.27 s | 8.7× |
| 40 | 8.71 s | 0.56 s | 15.5× |

Consistent with #3434's reported 15.2× at K=40, which is a good sign that both measure the same real effect.

### 1. Partial-failure recovery (the main one)

A batch introduces a failure mode the loop it replaces does not have. With three graphs where the middle one is corrupt:

| | result |
|--|--|
| three `global add` calls | repos A and C land; B fails |
| a batch | **nothing is written at all** — A and C are silently lost |

Both #3434 and my first draft do the second thing. For `graphify global add a.json b.json c.json` that is a usability regression versus the loop, and it gets worse the bigger the batch.

`on_error` selects the behaviour — `"abort"` (default, all-or-nothing, safe) or `"skip"` (`--keep-going` on the CLI), which commits the readable units and reports the failures. There is a test asserting the skip policy reproduces the sequential partial-success store exactly. Nothing is written when every unit fails, and the CLI exits non-zero on any failure so CI cannot read a partial batch as success.

### 2. The external-label index is threaded, not rebuilt per unit

#3434 rebuilds the index by scanning every node of `G` inside the per-unit loop. That is *correct*, and the comment explaining why it cannot be hoisted is right. But it is K whole-graph scans — a quadratic term reintroduced into the function that exists to remove one:

| K | index node-visits, rebuilt per unit | threaded |
|--:|--:|--:|
| 10 | 18,027 | 0 |
| 20 | 76,057 | 0 |
| 40 | 312,117 | 0 |
| 80 | 1,264,237 | 0 |

**In fairness: at these sizes this does not show up in wall clock** — JSON serialization dominates, and I measured the two variants as a wash. So this is an asymptotic cleanup, not a speedup, and I would rather say that than dress it up.

The index is instead updated as each unit contributes stubs, and rebuilt only after a prune that removed something — the one event that can invalidate it.

### 3. A latent dangling-node bug that fix exposed

Threading the index makes prune-invalidation load-bearing, and writing the test for it turned up a real defect: when a repo is re-added and its old revision owned an external stub, the prune deletes that stub. An index still pointing at it rewires a later unit's edge onto **a node that is no longer in the graph**.

`test_global_add_many_rebuilds_index_after_a_prune_removes_externals` pins it, and asserts no edge references a missing node. I mutation-tested it — disabling the invalidation makes it fail with exactly that dangling reference. #3434's per-unit rebuild avoids this by accident; worth keeping the test either way.

### 4. Empty repo tags (pre-existing, fixed while rewriting the CLI)

Tag inference used `source.parent.parent.name`, which for `graphify global add /tmp/graph.json` is the filesystem root — an **empty tag**. That prunes by `""` and registers the repo under a name no later add can address. Batching makes it much likelier to hit, since a batch mixes paths from several layouts. `_infer_repo_tag` now falls back through parent → stem, skipping generic output directory names, and a test asserts no input yields an empty tag.

### 5. Adopted from #3434

Duplicate-tag rejection and the up-front size-cap check are both good ideas from #3434 that I took directly, with credit — the second source under one tag would prune the first, and hoisting the cap means an oversized graph late in the batch does not first cost the merges ahead of it.

## Correctness

The acceptance criterion is that a batch produces what the same sequence of `global_add` calls produces. Held by construction:

- Units compose in input order against a live external index.
- The cross-repo call pass recomputes its own output, so one run over the finished batch lands where a run per unit would. Its count is reported at batch level — it reads the composed graph and does not divide over the units that made it.
- A unit whose source hash is unchanged is skipped without being pruned or recomposed. An all-unchanged batch never reads or writes the store.

Save ordering is graph-then-manifest, deliberately: a crash between the two writes leaves nodes the manifest does not list, which the next add heals by pruning the tag and recomposing. The reverse strands a recorded hash whose nodes are absent, and every later add then skips that repo — a silent permanent hole. I verified the recovery behaviour empirically rather than reasoning about it.

## Tests

30 new tests. Batch of many, existing repo replaced, cross-repo links between units in one batch, manifest metadata, empty and single-item batches, and sequential-vs-batched equivalence — the cases #3438 asks for — plus the abort/skip policies, duplicate tags, size caps, index scan-count and prune invalidation, and tag inference.

- `tests/test_global_graph.py`: **50 passed** (20 pre-existing, unchanged).
- Related suites (cross-repo, merge-graphs, build, CLI): **165 passed**.
- Full suite: **30 failed, 5335 passed**. Clean `v8` in a fresh worktree fails **31** of the same tests — Windows-only FIFO/socket, skill-install and unicode cases, none touching this code. No regressions.

## Not included

Streaming/incremental updates to cut peak memory, which #3438 explicitly scopes out and #3062 covers.


Sagexd08:
## Update: adopted @ROHIT8759's rollback, and two notes on the other implementations

@ROHIT8759 has been working the same issue on [`perf/global-add-many`](https://github.com/ROHIT8759/graphify/commits/perf/global-add-many) (no PR opened yet, so it shows on #3438 only as a referenced commit). Their branch went independent → `eaa87c5` (error policies + incremental indexing) → three commits on atomic rollback. Worth reading alongside this one.

### They found something I got wrong

Their rollback work covers a case this PR did not. A unit prunes its repo **before** merging the new revision, so a failure between those two steps leaves the repo removed but not replaced. My `on_error="skip"` had it worse than I realised:

- Only the *source read* was guarded, so a failure in prefix/prune/merge **propagated out of the batch entirely**, ignoring the policy the caller asked for.
- Had it been caught, the batch would then have saved a graph with that repo half-removed.

Fixed in `f8a91d2`, crediting the approach in both the code comment and the commit message. The whole unit is now guarded and a failure restores the repo's nodes and incident edges before the next unit runs, so a skipped unit leaves the store at that repo's previous revision.

Two things I kept from this PR's existing design: nothing is saved when every unit fails, and `abort` still writes nothing at all — so the *store* was never corrupt on this path even before the fix (the half-pruned graph only ever existed in memory and was discarded). The bug was that `skip` didn't skip.

Rollback also must not reintroduce the per-unit whole-graph scan this API exists to remove — the same trap their `a6f675f` had to dig out of. The snapshot is taken only for a repo the loaded graph actually holds nodes for, so adding a new repo pays nothing, and the set of present tags is collected in the same pass that builds the external label index. **A batch still makes exactly one whole-graph scan**, asserted by a test.

Re-measured after the change — no regression: **16.9× over sequential at K=40**, graphs still verified identical to the sequential result at every K. Now 54 tests in `test_global_graph.py`, 196 across related suites.

### Two bugs still open in that branch

Flagging here since there's no PR to review against, and they don't apply to this one:

1. **`cross_repo_calls` is broadcast to every result.** The final loop does `r["cross_repo_calls"] = cross_repo_calls` for each unit, so a 3-repo batch resolving 1 call reports `1, 1, 1` — which reads as 3. The pass runs once over the composed graph and its output doesn't divide over the units that produced it. #3434 and this PR both keep it batch-level only; #3434's test explicitly asserts it is absent from per-unit results.

2. **The store is rewritten even when every unit fails.** Under `skip`, `_save_global_graph` / `_save_manifest` run unconditionally after the loop, so a batch where nothing composed still rewrites the store with an unchanged graph — turning a read failure into a possible write failure. This PR returns early with `saved: False` in that case.

### Where this leaves the three implementations

#3434 (oldest, reviewed), this one, and @ROHIT8759's branch now all implement #3438 — itself a duplicate of #3433. That's a lot of duplicated effort on one fix, and it's a maintainer call to resolve. My position is unchanged from the top of this PR: I'd rather the fix land than that mine does. Happy to reduce this to just the parts the chosen implementation lacks — on current reading that would be the batch-level `cross_repo_calls`, the all-fail early return, the single-scan rollback, and the empty-tag inference fix.

Sagexd08:
## Review findings addressed — `6a42ff9`

Thanks both. All four of @copilot's findings were real; I verified each against the code before changing anything rather than taking them on faith.

### Fixed

**1. Size cap enforced after hashing.** Correct and it was my own regression — `v8` checked the cap before parsing, and I moved the hash ahead of it. `_file_hash` does `read_bytes()`, so an oversized graph was pulled into memory to compute a hash for a unit about to be rejected for its size. The cap now runs first for every unit, before anything reads the file; under `skip` the cap error is recorded without hashing. Test asserts `_file_hash` is never called when the cap trips.

**2. `--as <tag> <graph.json>` was rejected.** Also correct, and a genuine backward incompatibility — I checked `v8`'s parser and it set `tag` independently of position, so the option-first form worked. Now `--as` tags the *next* graph when it appears before one, and the trailing-graph form still works. A `--as` that would tag nothing is an error rather than silently dropped.

**3. "global graph not modified" per skipped unit.** Untrue in a batch where other units changed it. Now "not re-added", scoped to that repo.

**4. `error` key always present.** Docstring said it appeared only on failures. Documented as it behaves; `global_add`'s single-repo result keeps its original five keys, asserted by a test.

Writing these up exposed that **the CLI half had no test coverage at all** — three of the four are CLI- or shape-level and the Python-level suite could not have caught them. Added `tests/test_cli_global_add_batch.py` (10 tests) driving the real entry point as a subprocess with `HOME`/`USERPROFILE` redirected, since the store path resolves from `Path.home()` at import time.

Now 57 tests in `test_global_graph.py` + 10 CLI, 202 across related suites, lint clean. Re-benchmarked: no regression, graphs still verified identical to the sequential result at every K.

### Not changed, with reasoning

**"`global_add` return shape changed from single-repo result to batch envelope"** (graphify bot, high) — I believe this is a false positive. `global_add` is unchanged: same signature, same five keys, same values. It delegates to `global_add_many` and unwraps `["results"][0]`, and `test_global_add_delegating_result_has_no_error_key` pins the exact key set. The *new* `global_add_many` returns the envelope, but nothing previously called it. The three `cli.py` call sites are untouched. Happy to be shown otherwise if the analysis is seeing a path I'm not.

**"Concurrent `global_add_many` calls can overwrite each other"** and **"graph and manifest writes are not serialized as one transaction"** (high/medium) — both real, both pre-existing on `v8` and unchanged in kind by this PR. Two concurrent `global add`s already raced the same way; batching changes the window's length, not its existence. There is no lock file in the current design, so fixing it properly means introducing one, which is a separate change with its own semantics to settle (blocking vs. failing fast, staleness, cross-platform behaviour) and shouldn't ride along in a perf PR.

On the write ordering specifically, the current order is deliberate and documented in the code: graph first, then manifest. A crash between them leaves nodes the manifest doesn't list, which the next add heals by pruning the tag and recomposing. The reverse strands a recorded hash whose nodes are absent, and every later add then *skips* that repo — a silent permanent hole. I verified the recovery behaviour empirically rather than reasoning about it. Happy to open a follow-up issue for real transactionality if that's wanted.

**Coupling findings** (`global_add_many` fan-out, `global_remove` Ca·Ce, test-helper efferent coupling) — `global_remove` isn't touched by this branch. `global_add_many`'s fan-out is `global_add`'s, relocated: the delegation edge is the point, since one composition implementation means a change to the merge semantics can't land in the batch path and miss the single-unit path. Splitting the body to lower the number would trade a metric for exactly the duplication the delegation prevents. Same conclusion @xiongjianxu reached on #3434, for the same reason.

Sagexd08:
## Bot review findings addressed — `f3b3d6d`

Went through the five advisory findings from the latest bot review. Reproduced each one before changing anything rather than taking the consensus label at face value — the review itself says "agreed by 2 of 2 members but NOT verified," and that turned out to be the right warning: two were real, two were false positives, and one was real with the wrong mechanism attached.

### Fixed — two findings, one root cause

Both real findings were the same bug: **the pre-flight input checks raised unconditionally, ignoring the `on_error` policy the caller asked for.**

**`global_add_many` raises `FileNotFoundError` even under `on_error='skip'`** (medium) — real, and the worst of the set. A missing source is the failure `--keep-going` most exists for, and it took down the entire batch from the pre-flight loop. In a 20-repo batch, one mistyped path meant the other 19 readable repos landed nowhere.

**"Failed duplicate tag in a batch can erase an earlier successful add"** (medium) — the stated *mechanism* is wrong. A duplicate tag cannot erase an earlier add, because the raise happens in the pre-flight loop before anything is composed; I confirmed the store is untouched on that path. But the policy bug underneath was real and identical in shape: one duplicated tag cost the whole batch instead of the one offending unit.

This is the same defect I fixed in `f8a91d2` for the merge path, still sitting in the checks that run *before* it. I'd fixed the guarded region and left the gate in front of it unguarded. Both rejections are now recorded per-unit like any other failure and the rest of the batch proceeds; `abort` is unchanged and still writes nothing at all.

One semantic call worth flagging: for a duplicate tag I reject the **second** unit, not the first. The second is the one that would prune the first, so keeping the earlier one leaves the store holding the revision the caller listed first rather than an arbitrary one of the two.

### Not changed, with reasoning

**"`global_add_many` return contract changed: removed repo_tag/nodes_added/skipped keys"** (high) — false positive. `git show v8:graphify/global_graph.py` has no `global_add_many`; it is new in this PR, so there is no prior contract to break. Those three keys are present on every per-unit result. I'd gently suggest this points at something on the bot's side — either a stale base or a new function being classified as modified — since a function that doesn't exist on the base branch can't have a changed return contract.

**"`global_add` return shape changed from single-add result to batch envelope"** (medium) — false positive, same as my earlier reply. `global_add` is unchanged: same signature, same five keys, same values. `test_global_add_delegating_result_has_no_error_key` pins the exact key set.

**"Concurrent `global_add_many` calls can overwrite each other"** (high) — real, but pre-existing on `v8` and unchanged in kind by this PR. Two concurrent `global add`s already raced identically; batching changes the window's length, not its existence. Fixing it properly means introducing a lock file, which has its own semantics to settle (blocking vs. failing fast, staleness, cross-platform behaviour) and shouldn't ride along in a perf PR. Happy to open a follow-up issue.

The recurring **coupling "health regressions"** are the same seven inline comments repeated across all three review runs. `global_add_many`'s fan-out is `global_add`'s, relocated — the delegation edge is the point. `global_remove` isn't touched by this branch at all.

### Verification

New tests, all failing before the fix and passing after:

- 5 in `test_global_graph.py` — `skip` survives a missing source and a duplicate tag; `abort` still raises and writes nothing for each; and `skip`'s resulting graph still fingerprint-matches a sequential `global_add` loop over the same inputs, which is the actual contract the policy owes.
- 3 in `test_cli_global_add_batch.py` driving the real `--keep-going` entry point as a subprocess, asserting the readable repos land, the failed one is reported on stderr, and the exit code is still nonzero so CI can't read a partial batch as a clean one.

**75 passing** across the two files (62 + 13), lint clean. The single-whole-graph-scan assertion and the sequential-equivalence tests still pass, so the perf property this PR exists for is intact.

On the full suite: 33 failures on this branch vs. 31 on `v8`, but the sets differ in *both* directions — 6 fail only on the branch, 4 only on `v8`. All 10 pass in isolation on their respective branches. They're order-dependent install/skill-roundtrip tests colliding over shared `HOME`-relative paths, and nothing in `global_graph.py` or the batch CLI path appears in either set. Pre-existing flakiness, not a regression from this change — but worth a separate issue if it's news.

PR #3442, opened 2026-09-09
Stop building generated code from unsanitized free text in skill instructions

### LINKED ISSUE #3439: Skill instructions interpolate unsanitized user input into Python/shell source (query.md, add-watch.md)
## Summary

Two of the graphify skill's reference docs instruct the host agent to interpolate raw, unsanitized user-controlled values directly into generated Python or shell source, which is unsafe.

## `.copilot/skills/graphify/references/add-watch.md` (around line 17)

`URL`, `AUTHOR`, and `CONTRIBUTOR` are user-provided values that the instructions interpolate into single-quoted Python source. An embedded apostrophe causes a syntax error, and crafted input can execute arbitrary Python.

**Fix suggestion:** pass these values as command-line arguments or structured input (e.g. JSON via stdin) instead of embedding them textually into generated source.

## `.copilot/skills/graphify/references/query.md` (around line 171)

The `save-result` template instructs the agent to substitute the user's verbatim question and the generated answer directly into a shell command. Double quotes still allow command substitution/backticks, and embedded quotes break argument boundaries — so ordinary content can fail and crafted content can execute commands.

**Fix suggestion:** pass these values through an argv-safe mechanism (e.g. write structured input to a temp file and invoke via Python `subprocess.run([...])` with an argument list, not a shell string). Apply the same fix to the other `save-result` templates in that file.

ayushcodes10:
Fixes #3439.

Two of the generated skill reference docs instructed the host agent to substitute raw, unsanitized values directly into generated Python or shell source.

## `add-watch.md` (the `/graphify add` flow)

The instructions had the agent embed a URL and optional author/contributor name inside a single-quoted Python string literal, passed to `python -c "..."`:

```python
out = ingest('URL', Path('./raw'), author='AUTHOR', contributor='CONTRIBUTOR')
```

An embedded apostrophe breaks the Python syntax; crafted input runs arbitrary Python.

## `query/default.md` (all three `save-result` call sites: query, path, explain)

The instructions had the agent substitute the user's verbatim question and a full LLM-generated answer (unbounded length, arbitrary content) into a double-quoted shell argument:

```bash
graphify save-result --question "ORIGINAL_QUESTION" --answer "ANSWER" ...
```

Double quotes still allow command substitution/backticks, and embedded quotes break argument boundaries.

## Fix

`save-result` already had `--answer-file` for exactly this reason (#1502). This PR:

1. Adds the missing symmetric options: `--question-file`/`--correction-file` for `save-result`, and a `--from-file` (JSON payload: `url`/`author`/`contributor`/`dir`) for `graphify add`. All existing positional/flag forms are unchanged — purely additive.
2. Rewrites both fragment files to have the agent write the value to a file with its own file-write tool first, then pass only that file's path — never the content — on the command line. This closes the injection surface entirely rather than just narrowing it: the content never enters shell or Python source at all.
3. Regenerates all 14 platform skill variants and their `tools/skillgen/expected/` fixtures from the updated shared fragments.

## Test plan
- [x] `pytest tests/test_ingest.py tests/test_reflect.py` — new tests cover `--question-file`/`--correction-file` round-tripping adversarial content (backticks, `$()`, both quote styles) through `save-result`, and `graphify add --from-file` passing an adversarial JSON payload through to `ingest()` unexecuted (mocked, network-free) plus a check that the original positional-arg form still works unchanged
- [x] `python -m tools.skillgen --check` and all guard modes (`--audit-coverage`, `--schema-singleton`, `--monolith-roundtrip`, `--always-on-roundtrip`) pass
- [x] Full suite passes except pre-existing, unrelated failures (`test_ollama_retry_cap.py`, missing `openai` module in this environment; `test_ts_normalizer_scales_linearly_on_large_files`, a flaky wall-clock timing assertion, confirmed to pass in isolation)

### DIFF
--- a/graphify/cli.py
+++ b/graphify/cli.py
@@ -1970,13 +1970,26 @@ def dispatch_command(cmd: str) -> None:
         # argument.
         args = sys.argv[2:]
         from_file = None
+        from_file_requested = False
         for i, a in enumerate(args):
-            if a == "--from-file" and i + 1 < len(args):
-                from_file = args[i + 1]
+            if a == "--from-file":
+                from_file_requested = True
+                if i + 1 < len(args):
+                    from_file = args[i + 1]
                 break
+        if from_file_requested and not from_file:
+            print("error: --from-file requires a path argument", file=sys.stderr)
+            sys.exit(1)
         if from_file:
-            payload = json.loads(Path(from_file).read_text(encoding="utf-8"))
-            url = payload["url"]
+            try:
+                payload = json.loads(Path(from_file).read_text(encoding="utf-8"))
+                url = payload["url"]
+            except (OSError, json.JSONDecodeError) as exc:
+                print(f"error: could not read --from-file payload: {exc}", file=sys.stderr)
+                sys.exit(1)
+            except KeyError:
+                print("error: --from-file payload is missing required key 'url'", file=sys.stderr)
+                sys.exit(1)
             author = payload.get("author")
             contributor = payload.get("contributor")
             target_dir = Path(payload.get("dir") or "raw")
--- a/graphify/cli.py
+++ b/graphify/cli.py
@@ -1983,10 +1983,18 @@ def dispatch_command(cmd: str) -> None:
         if from_file:
             try:
                 payload = json.loads(Path(from_file).read_text(encoding="utf-8"))
-                url = payload["url"]
             except (OSError, json.JSONDecodeError) as exc:
                 print(f"error: could not read --from-file payload: {exc}", file=sys.stderr)
                 sys.exit(1)
+            if not isinstance(payload, dict):
+                print(
+                    "error: --from-file payload must be a JSON object with a "
+                    "'url' key, not " + type(payload).__name__,
+                    file=sys.stderr,
+                )
+                sys.exit(1)
+            try:
+                url = payload["url"]
             except KeyError:
                 print("error: --from-file payload is missing required key 'url'", file=sys.stderr)
                 sys.exit(1)
--- a/graphify/cli.py
+++ b/graphify/cli.py
@@ -1461,22 +1461,39 @@ def dispatch_command(cmd: str) -> None:
     elif cmd == "save-result":
         # graphify save-result --question Q --answer A [--type T] [--nodes N1 N2 ...]
         #                      [--outcome useful|dead_end|corrected] [--correction TEXT]
+        #
+        # --question-file/--answer-file/--correction-file read the value from a
+        # file instead of a command-line argument: skill instructions that build
+        # this command from free text (the user's verbatim question, an LLM's
+        # generated answer, of unbounded length and content) must not substitute
+        # that text directly into a shell command string, since embedded quotes,
+        # backticks, or $() would corrupt or escape the command entirely
+        # (#3439). Writing the value to a file first has no such injection
+        # surface at all.
         import argparse as _ap
 
         p = _ap.ArgumentParser(prog="graphify save-result")
-        p.add_argument("--question", required=True)
+        p.add_argument("--question", default=None)
+        p.add_argument("--question-file", dest="question_file", default=None)
         p.add_argument("--answer", default=None)
         p.add_argument("--answer-file", dest="answer_file", default=None)
         p.add_argument("--type", dest="query_type", default="query")
         p.add_argument("--nodes", nargs="*", default=[])
         p.add_argument("--outcome", choices=("useful", "dead_end", "corrected"), default=None)
         p.add_argument("--correction", default=None)
+        p.add_argument("--correction-file", dest="correction_file", default=None)
         p.add_argument("--memory-dir", default=str(Path(_GRAPHIFY_OUT) / "memory"))
         opts = p.parse_args(sys.argv[2:])
+        if opts.question_file:
+            opts.question = Path(opts.question_file).read_text(encoding="utf-8").strip()
+        elif not opts.question:
+            p.error("--question or --question-file is required")
         if opts.answer_file:
             opts.answer = Path(opts.answer_file).read_text(encoding="utf-8").strip()
         elif not opts.answer:
             p.error("--answer or --answer-file is required")
+        if opts.correction_file:
+            opts.correction = Path(opts.correction_file).read_text(encoding="utf-8").strip()
         from graphify.ingest import save_query_result as _sqr
 
         out = _sqr(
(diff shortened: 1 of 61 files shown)


ayushcodes10:
Investigated this round's bot findings.

Confirmed real, fixed (now at cea4992):

1. "Missing --from-file operand is treated as the URL" -- `graphify add --from-file` with no path argument fell through to the positional-url branch and treated the literal string "--from-file" as the URL, producing a confusing "Blocked URL scheme" error instead of pointing at the actual mistake. Reproduced directly. Now detects the flag is present with no operand and exits with a clear error.

2. "add --from-file crashes with KeyError instead of clean error when 'url' missing" -- a payload missing the required "url" key raised a raw, unhandled KeyError instead of the same clean error message every other failure in this command already produces. Reproduced directly. Also added the same clean-error handling for malformed JSON and a missing file, which hit the same code path but weren't specifically called out.

3. "Fixed /tmp save-result filenames race across concurrent sessions" -- confirmed: the instructions used a fixed, shared filename (`/tmp/graphify_answer.txt`, etc.) for the question/answer/payload handoff files. Two concurrent graphify sessions (two agents, two terminal tabs) following the same instructions at once could race on that path, with one session's write overwriting or being read as another's value -- undermining the whole point of the file-based handoff this PR introduced. Every instruction now runs `mktemp` first to reserve a path unique to that run before writing anything.

Regression tests added for both CLI fixes; all 14 platform skill variants and expected fixtures regenerated for the mktemp fix; all skillgen guards and the full test suite pass (same pre-existing, unrelated failures as before: a missing optional dependency and one flaky timing test, both confirmed unrelated to this change).

ayushcodes10:
Investigated this round's bot findings.

Confirmed real, fixed (now at d66f2b2):

1. "--from-file crashes on non-object JSON payloads" -- json.JSONDecodeError alone only catches a syntax error; valid JSON that isn't an object (a list, a bare string, a number) parsed fine and then raised a raw TypeError from subscripting it with "url". Reproduced directly for all three shapes. Now checks the parsed value is a dict before looking up "url" and names the actual type in the error.

2. "Explain save command still inlines an untrusted node name" -- confirmed: the path and explain save-result templates still built --question from "Path from NODE_A to NODE_B" / "Explain NODE_NAME" with the node label spliced in inline, while the answer next to it was already file-based. A node label isn't guaranteed free of shell characters either, since it can come from extracted document content. Both now write the question to a reserved unique file the same way the answer already does.

Regression tests added for the non-object payload cases; affected platform skill variants and expected fixtures regenerated; all skillgen guards and the full test suite pass (same pre-existing, unrelated failures as before).

ayushcodes10:
Pushed 5 new commits addressing the review findings:

- **High: node labels still substituted directly into shell commands** (the main ask). This was broader than the one flagged line — the same raw substitution existed in three places: the save-result node flag for query, path, and explain, plus the inline Python fallback scripts for path and explain (a node label was a literal Python string, embedded inside a double quoted shell argument). All five now go through a temp file the same way the question and answer text already did — the node label never touches a shell argument or gets embedded in generated source. Regenerated all fourteen platform skill variants and their expected fixtures.
- **Medium: save-result rejects a present but empty question/answer.** Was checking by truthiness; now checks whether the value is `None` (never provided) instead.
- **Medium: add --from-file can traceback on an invalid dir type.** A non-string `dir` in the JSON payload now gets the same clean error message as the other malformed-payload cases, instead of an unhandled TypeError from `Path()`.
- **Medium: Windows add-watch skill requires POSIX mktemp.** Confirmed real, but broader than this one line — the same `mktemp`/`/tmp` pattern is used throughout the Windows query.md reference too (6 occurrences), and this repo already has a dedicated, deliberately strict POSIX-to-PowerShell translator (`tools/skillgen/gen.py`) for the core skill body, which this reference-doc content isn't routed through. Fixing this properly means extending that translation (or an equivalent) to the reference docs, which is a design decision beyond this PR's scope. Filed as #3620 to track separately rather than bundling a half fix in here.

Full test suite (5370 passed, 66 skipped, no regressions) and `skillgen --check`/`--audit-coverage`/`--schema-singleton` all green.

ayushcodes10:
Correction: the follow up issue for the Windows mktemp/PowerShell gap is #3600, not #3620 as I wrote above (typo while filing it).

ayushcodes10:
Pushed 3 more commits addressing this round's findings:

- **The three "url not validated as a string" duplicates**: real. A non-string `url` in the JSON payload reached the ingest layer unchecked and surfaced a confusing `AttributeError` (`'list' object has no attribute 'lower'`) instead of a clean error, unlike the `dir` check I'd already added. Now validated the same way.
- **"mktemp with .json suffix uses literal XXXXXX not a template"**: real, and worse than it sounds — verified directly with the BSD `mktemp` macOS ships (`/usr/bin/mktemp`): a suffix written after the `X` run comes back completely unsubstituted, so `mktemp /tmp/graphify_add_payload.XXXXXX.json` returned the exact same literal path on every call, silently defeating the whole point of reserving a unique path (the "fixed, shared filename risks a concurrent session..." rationale these docs already state elsewhere). The `add` command's payload option never inspects the extension, so I just dropped it — every other reserved path in these references already puts the placeholder last; this was the one exception.
- **"path/explain inline scripts read all node labels from single-line file"**: the two-term unpack (`a_term, b_term = ...splitlines()`) assumed the file held exactly two non-blank lines; a stray blank line (an extra trailing newline beyond the one `.strip()` already tolerated, or one left in the middle) would change the line count and crash the unpack. Now filters blank lines first, matching how the CLI-side node option already reads this same file.

Full suite (5373 passed, no regressions) and every `skillgen` guard (`--check`/`--audit-coverage`/`--schema-singleton`) still green.

ayushcodes10:
Pushed 3 more commits:

- **"from-file allows non-string author/contributor values"** (both duplicate variants): real, though different in kind from the url/dir cases — a non-string author/contributor does not crash today (\`_yaml_str\` stringifies anything via \`str(s)\`), it silently saves a garbled frontmatter value like \`author: "['Jane']"\` instead. Validated the same way as url/dir for consistency.
- **"Path fallback cannot handle node names containing newlines"**: real. The two-term unpack (\`a_term, b_term = ...\`) read a shared file assuming exactly two non-blank lines; a concept name containing a literal embedded newline would split across lines and change that count, crashing the unpack. Rather than patch around it, each concept name now gets its own reserved file, read as one whole value with no line splitting at all — an embedded newline in the name itself can no longer change how many values come back. This needed the node file option to accept more than one path (\`--nodes-file A_PATH B_PATH\`), which stays backward compatible with the existing single-path, multiple-lines form query/explain still use (those two are unaffected: query already cites a variable number of labels through one file on purpose, and explain's single value was never split to begin with).

Full suite (5380 passed, no regressions) and every \`skillgen\` guard green.

ayushcodes10:
Pushed 2 more commits:

- **"save-result no longer rejects empty answers"**: real, and a good catch — the earlier fix (checking `is None` instead of truthiness) correctly closed the "misleading required message for a present-but-empty flag" gap, but as a side effect it also started silently *accepting* an explicitly empty question/answer, which is never meaningful content to save. Both are still rejected now, just with an accurate "must not be empty" message instead of the "required" one a genuinely absent flag gets.
- **The four duplicate "singular --nodes-file" findings**: not a functional bug — `--nodes-file` already accepts multiple paths (`nargs="+"`, from an earlier commit on this PR) and the two-path example works correctly. But it's a fair doc-clarity point for a doc whose whole audience is an LLM agent constructing shell commands: nothing in the text explicitly said the flag was multi-value, so a two-path example sitting next to `/graphify query`'s and `/graphify explain`'s own single-path examples could plausibly read as a mistake. Added one explicit sentence at the flag's first use stating both forms are valid.

Full suite (5381 passed, no regressions) and every `skillgen` guard green.

ayushcodes10:
Re: the two new findings.

**"add --from-file crashes on invalid UTF-8 payload"**: real, confirmed by reproduction — a payload file with invalid UTF-8 bytes raised a raw, unhandled \`UnicodeDecodeError\` from \`read_text()\` instead of the clean error every other unreadable-payload shape already gets. Pushed a fix: the read is now also guarded against \`UnicodeDecodeError\`, matching the existing \`OSError\`/\`JSONDecodeError\` handling right next to it.

**"save-result --answer-file now rejects empty files that were previously accepted"**: accurate description, but intentional — this is the same fix from an earlier round in this thread (an explicitly empty \`--answer\`/\`--question\` is never meaningful content to save, so it's rejected with a clear "must not be empty" message rather than silently saved). That check applies uniformly regardless of whether the empty value came from the flag directly or from an empty file; a blank \`--answer-file\` was accepted before only because that path had no post-read validation at all, which was the gap the earlier fix closed. Not reverting this — flagging it as confirmed-intentional rather than silently letting it sit unaddressed.

Full suite (5382 passed, no regressions) and \`skillgen --check\` green.

ayushcodes10:
Addressed both findings from the latest review round:

- File reads for `--question-file`/`--answer-file`/`--correction-file`/`--nodes-file` now end in a clean argparse style error (naming the flag and the path) instead of a raw traceback on a missing or unreadable path.
- The add reference's JSON payload instructions no longer tell the agent to build the payload by literally substituting URL/author/contributor into a template string — that reintroduced the same free text into a string risk the file based approach exists to avoid, one level down. It now instructs building the JSON as data through a real encoder.

While checking this I also found the `aider`/`devin` skills still build `save-result`/`graphify add` from the same kind of unescaped free text substitution (a `--answer "ANSWER"` shell line where bash still expands `$(...)` inside double quotes, and a Python `ingest('URL', ...)` string literal for `add`) — those two platforms weren't covered by the original file based fix since they have no separate file write tool. That's a larger, security sensitive fix needing its own design (a quoted heredoc based safe write), so I filed it separately as #3640 rather than folding it into this PR.

ayushcodes10:
Re: "save-result now rejects empty --answer that was previously accepted" (`cli.py:1510`) — this is a re-surfacing of a finding already investigated and fixed earlier in this PR's history (2026-09-16, "save-result no longer rejects empty answers"), not something new from the latest push. Confirmed the check is still correctly in place:

\`\`\`python
if not opts.answer:
    p.error("--answer must not be empty")
\`\`\`

No regression from the latest commits.

Issue #3407, opened 2026-09-08
Symbol node IDs collide across directories; walk order decides the winner, so output is not reproducible across filesystems


fracasamax:
## Summary

Symbol node IDs are not path-qualified, so same-named symbols in different directories collapse into a single node. Which of the colliding definitions survives depends on filesystem walk order, because `detect()` never sorts its file list. The result is twofold:

1. **Silent data loss** — one of the definitions disappears from the graph entirely.
2. **Non-reproducible output** — the same commit produces a different `graph.json` on ext4 (Linux CI) than on APFS (macOS), so any workflow that hashes or commits Graphify artifacts can only pass on one platform at a time.

Verified on `graphifyy==0.5.6`.

## Reproduction

Two Next.js-style apps, each with the conventional `app/layout.tsx` exporting `RootLayout`:

```bash
mkdir -p demo/apps/admin/app demo/apps/reference-book/app && cd demo
for app in admin reference-book; do cat > apps/$app/app/layout.tsx <<'EOF'
export default function RootLayout({ children }: { children: React.ReactNode }) {
  return <html><body>{children}</body></html>;
}
EOF
done
graphify update .
python3 -c "
import json
g = json.load(open('graphify-out/graph.json'))
for n in g['nodes']: print(f\"{n['id']:32} {n.get('source_file')}\")
"
```

Actual — 3 nodes; `apps/admin`'s `RootLayout` is gone:

```
admin_app_layout_tsx             admin/app/layout.tsx
app_layout_rootlayout            reference-book/app/layout.tsx
reference_book_app_layout_tsx    reference-book/app/layout.tsx
```

Expected: 4 nodes, one `RootLayout` per file. Note the file-level IDs *are* path-derived and unique; only the symbol-level ID collides.

## Root cause

**Collision.** `extract._make_id` builds symbol IDs from `_file_stem(path)` plus the symbol name, and `_file_stem` qualifies only with the immediate parent directory:

```python
def _file_stem(path: Path) -> str:
    """Return a stem qualified with the parent directory name to avoid ID collisions
    when multiple files share the same filename in different directories (#550)."""
    parent = path.parent.name
```

[...]

PR #3454, opened 2026-09-09
fix(detect): sort os.walk to ensure reproducible symbol IDs (#3407)

nikhilsaxena04:
The bug:
Cross-platform extractions were not deterministic (yielding different output graphs between macOS and Linux CI) because the fallback ordering relied on `os.walk`, which processes directories in raw filesystem order (unordered).

The fix:
Explicitly sorted `dirnames` and `filenames` in the `os.walk` loop in `graphify/detect.py` to guarantee that discovery is 100% reproducible regardless of the underlying OS or filesystem.

Tests:
`tests/test_detect_sort.py` added to assert on full relative paths to rigorously prove that os.walk is traversing the directories in sorted order. This prevents regression of non-deterministic traversal order.

### DIFF
--- a/graphify/detect.py
+++ b/graphify/detect.py
@@ -1776,6 +1776,8 @@ def detect(root: Path, *, follow_symlinks: bool | None = None, google_workspace:
         for dirpath, dirnames, filenames in os.walk(
             scan_root, followlinks=follow_symlinks, onerror=_on_walk_error
         ):
+            dirnames.sort()
+            filenames.sort()
             dp = Path(dirpath)
             if follow_symlinks and os.path.islink(dirpath):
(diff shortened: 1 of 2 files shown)


nikhilsaxena04:
Hey @safishamsi first off, thank you so much for the recent reviews and getting my last batch of PRs merged and shipped! I've been continuing to have a fun diving into the Graphify codebase.

I just wanted to ping you on this PR whenever you have a breather. It's fully green and cleanly resolves the non-deterministic extractions we were seeing between macOS and Linux CI by explicitly sorting os.walk.

No rush since I know you are all heads-down building, but I’d love to get this one merged so I can start grabbing some more issues from the backlog! Excited to continue building with the community.

PR #3458, opened 2026-09-09
fix(ts): keep star-export resolution off same-named interface methods

### LINKED ISSUE #3436: TypeScript: a workspace-barrel-imported function binds to a same-named interface method, so the real `calls` edge is never emitted
# TypeScript: a workspace-barrel-imported function binds to a same-named interface *method*, so the real `calls` edge is never emitted

**Version:** graphifyy 0.9.55 (uv tool install), Windows 11, Python 3.13.2

## Summary

In a pnpm workspace, a call to a function imported by name through a package barrel (`import { evaluate } from '@scope/domain'`) is bound to an unrelated **interface method** node of the same bare name (`Rule.evaluate`) instead of the exported free function. The correct `calls` edge to the function is never emitted.

The trigger is specifically the **workspace-barrel import**. The identical code using a relative import (`from './engine.js'`) resolves correctly, and removing the same-named method from the interface also resolves correctly — so this needs *both* the barrel hop and the name collision.

The failure is silent and inverts a core signal: the function ends up with zero inbound `calls` edges and reads as dead code, while a phantom `.evaluate()` method node accumulates the callers. In a real repo this hid the single hottest call path (one call per portfolio row, ~357/boot) from `graphify query`, and the god-node ranking under-counted the function's fan-out.

## Minimal reproduction

```
repro/
  pnpm-workspace.yaml
  packages/domain/package.json
  packages/domain/src/{types.ts,engine.ts,index.ts}
  packages/api/{package.json,src/cache.ts}
```

`pnpm-workspace.yaml`
```yaml
packages:
  - "packages/*"
```

`packages/domain/package.json`
```json
{ "name": "@repro/domain", "version": "0.0.0", "private": true, "type": "module",
  "main": "./src/index.ts", "exports": { ".": "./src/index.ts" } }
```

`packages/domain/src/types.ts`
```ts
export interface Rule {
  code: string
  evaluate(ctx: number): string | null   // <-- same bare name as the function below
}
```

`packages/domain/src/engine.ts`
```ts
import type { Rule } from './types.js'

export function evaluate(rules: readonly Rule[], ctx: number): string[] {
  return rules.map((rule) => rule.evaluate(ctx)).filter((f) => f !== null)
}
```

`packages/domain/src/index.ts`
```ts
export * from './types.js'
export * from './engine.js'
```

`packages/api/package.json`
```json
{ "name": "@repro/api", "version": "0.0.0", "private": true, "type": "module",
  "dependencies": { "@repro/domain": "workspace:*" } }
```

`packages/api/src/cache.ts`
```ts
import { evaluate } from '@repro/domain'
import type { Rule } from '@repro/domain'

export function warm(rules: readonly Rule[]): string[] {
  return evaluate(rules, 1)
}
```

Driver:
```python
import os
from pathlib import Path

def main():
    root = Path(os.environ['REPRO_ROOT'])
    files = sorted(p for p in collect_files(root) if p.suffix == '.ts')
    res = extract(files, cache_root=None)
    nodes = {n['id']: n for n in res['nodes']}
    lab = lambda i: nodes[i]['label'] if i in nodes else i + ' (PHANTOM)'
    for e in res['edges']:
        s, t = lab(e['source']), lab(e['target'])
        if 'evaluate' in s or 'evaluate' in t or 'warm()' in s:
            print('  %-14s --%s--> %s' % (s, e.get('relation'), t))

if __name__ == '__main__':
    main()
```

## Actual

```
  cache.ts       --imports--> domain_src_index_evaluate (PHANTOM)
  engine.ts      --contains--> evaluate()
  Rule           --method--> .evaluate()
  cache.ts       --imports--> .evaluate()
  warm()         --calls--> .evaluate()
```

`warm()` calls the interface method node. There is no `warm() --calls--> evaluate()` edge, and `evaluate()` (`packages/domain/src/engine.ts`) has no inbound `calls` edge at all.

## Expected

```
  warm()         --calls--> evaluate()      # packages/domain/src/engine.ts
```

`Rule.evaluate` is a method on an interface that `warm()` never calls; it should not attract the call.

## Isolation (two controls, both pass)

**Control A — relative import, same collision.** Same three symbols, single package, `import { evaluate } from './engine.js'`:

```
  caller.ts      --imports--> evaluate()
  engine.ts      --contains--> evaluate()
  Rule           --method--> .evaluate()
  run()          --calls--> evaluate()     <-- correct
```

**Control B — workspace barrel, no collision.** Same workspace layout, interface method renamed `score`:

```
  cache.ts       --imports--> domain_src_index_evaluate (PHANTOM)
  engine.ts      --contains--> evaluate()
  cache.ts       --imports--> evaluate()
  warm()         --calls--> evaluate()     <-- correct
```

So the mis-binding requires the barrel hop **and** a same-named method. Barrel resolution itself works: `_resolve_workspace_import('@repro/domain', packages/api/src)` correctly returns `packages/domain/src/index.ts`, and `export *` chains are followed (Control B resolves through the same path).

## Note on the phantom edge

`cache.ts --imports--> domain_src_index_evaluate` points at a node id that is never created (the barrel defines no symbols) and appears in **both** the failing and the passing variant, so it is independent of the collision. That looks like #2265; not re-reporting it here, but flagging that it sits on the same code path and may be the point where the binding is later re-resolved by bare name.

## Related

- #3237 — monorepo cross-*project* name collisions. Same family (name-based resolution), different trigger: there the correct target is absent/unrelated; here it is present and reachable through the barrel, and loses to a method of the same bare name.
- #3346 — namespace/aliased imports emit *no* `calls` edge. Here a plain named import emits a *wrong* one.
- #2265 — edges to node ids never created (see above).

L4XB:
## Summary

Fixes #3436.

In a pnpm workspace, `import { evaluate } from '@repro/domain'` where the barrel is

```ts
export * from './types.js'   // interface Rule { evaluate(ctx): ... }
export * from './engine.js'  // export function evaluate(...)
```

bound the import (and the `warm() --calls-->` edge) to the interface **method** node `Rule.evaluate` instead of the exported function. The function ended up with no inbound `calls` edge.

## Root cause

`resolve_exported_origin` in `graphify/extractors/resolution.py` walks the `export *` targets and returns the first `(star_target, imported_name)` that has a symbol node. The symbol map is keyed by the label with the leading dot stripped, so the method node `.evaluate()` in `types.ts` matched the name before the walk reached `engine.ts`.

## Fix

- Track member nodes (labels starting with `.`) in a separate `member_symbol_keys` set while building `symbol_nodes`, and never let a member shadow a same-named top-level symbol in the map.
- The star-export walk skips member keys: a module can only re-export top-level bindings.

Driver output from the issue after the fix:

```
  engine.ts      --contains--> evaluate()
  Rule           --method--> .evaluate()
  cache.ts       --imports--> evaluate()
  warm()         --calls--> evaluate()
```

## Tests

- New `test_ts_export_star_skips_same_named_interface_method_and_binds_the_function` in `tests/test_js_import_resolution.py` (fails on `v8`, passes with the fix): asserts the `imports` edge and the `calls` edge land on `engine.ts::evaluate`, and that no `calls` edge lands on the method node.
- `pytest tests/` : same failure set with and without the patch (24 pre-existing failures locally in `test_terraform.py` / the linear-scaling timing test, unrelated), `ruff check` clean.

### DIFF
--- a/graphify/extractors/resolution.py
+++ b/graphify/extractors/resolution.py
@@ -986,13 +986,28 @@ def _apply_symbol_resolution_facts(
     path_by_resolved = {path.resolve(): path for path in paths}
     source_file_id = {path.resolve(): _make_id(str(path)) for path in paths}
     symbol_nodes: dict[tuple[Path, str], str] = {}
+    # Member nodes (`.method()` labels) share their bare name with top-level
+    # symbols once the leading dot is stripped. A module can only re-export
+    # top-level bindings, so a star-export walk must never bind an imported
+    # name to a class/interface member (#3436). Track those keys separately
+    # and never let a member shadow a same-named top-level symbol.
+    member_symbol_keys: set[tuple[Path, str]] = set()
     for node in nodes:
         source_path = _js_source_path(str(node.get("source_file", "")), root)
         if source_path is None:
             continue
-        label = str(node.get("label", "")).strip().strip("()").lstrip(".")
-        if label and node.get("id"):
-            symbol_nodes[(source_path, label)] = str(node["id"])
+        raw_label = str(node.get("label", "")).strip()
+        label = raw_label.strip("()").lstrip(".")
+        if not label or not node.get("id"):
+            continue
+        key = (source_path, label)
+        if raw_label.startswith("."):
+            if key in symbol_nodes:
+                continue
+            member_symbol_keys.add(key)
+        else:
+            member_symbol_keys.discard(key)
+        symbol_nodes[key] = str(node["id"])
 
     def ensure_symbol_node(path: Path, name: str, line: int) -> str:
         resolved_path = path.resolve()
@@ -1177,10 +1192,10 @@ def _apply_symbol_resolution_facts(
             return resolve_exported_origin(origin[0], origin[1], seen)
         for star_target in star_exports_by_file.get(target_path, []):
             star_key = (star_target, imported_name)
-            if star_key in symbol_nodes:
+            if star_key in symbol_nodes and star_key not in member_symbol_keys:
                 return star_key
             resolved = resolve_exported_origin(star_target, imported_name, seen)
-            if resolved in symbol_nodes:
+            if resolved in symbol_nodes and resolved not in member_symbol_keys:
                 return resolved
         return key
 
--- a/tests/test_js_import_resolution.py
+++ b/tests/test_js_import_resolution.py
@@ -140,6 +140,58 @@ def test_ts_export_star_from_index_resolves_imported_symbol_to_origin(tmp_path: 
     assert _has_symbol_edge(result, "src/routes/page.ts", "src/lib/foo.ts", "Foo")
 
 
+def test_ts_export_star_skips_same_named_interface_method_and_binds_the_function(tmp_path: Path):
+    # #3436: `export *` can only forward top-level bindings. When the first
+    # star target declares an interface with a METHOD of the same bare name as a
+    # function exported by a later star target, the imported name must bind to
+    # the function, and the call must land on it -- not on the method node.
+    types = _write(
+        tmp_path / "packages/domain/src/types.ts",
+        "export interface Rule {\n  code: string\n  evaluate(ctx: number): string | null\n}\n",
+    )
+    engine = _write(
+        tmp_path / "packages/domain/src/engine.ts",
+        "import type { Rule } from './types.js'\n\n"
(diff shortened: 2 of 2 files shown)


safishamsi:
Shipped in **v0.9.58** — landed on `v8` via cherry-pick with your authorship preserved. Thanks @L4XB — a TS export* re-export no longer resolves a name to a same-named interface method. Release: https://github.com/Graphify-Labs/graphify/releases/tag/v0.9.58

PR #3465, opened 2026-09-10
Never infer an empty repo tag for global add

### LINKED ISSUE #3464: global add infers an empty repo tag, producing a store entry the CLI cannot remove
### Summary

`graphify global add` infers its repo tag as `source.parent.parent.name` (`graphify/cli.py:3129`
on v8). That expression is empty whenever the graph is not exactly two levels below a named
directory — a graph at `/tmp/graph.json`, or any relative path such as `graphify-out/graph.json`.
The empty tag is accepted, so every node is prefixed with `::`, the repo is pruned by `""`, and the
manifest entry it registers cannot be addressed again: `global remove` reads an empty tag as a
missing argument and prints usage.

`merge-graphs` does not have this hole. It infers through `distinct_repo_tags`
(`graphify/build.py:2240`), whose `d.name or "repo"` is precisely the degrade the `global add`
branch lacks.

### Reproduction

```python
# store patched to a scratch dir; a one-node graph written to /tmp/graph.json
m.sys.argv = ["graphify", "global", "add", "/tmp/graph.json"]; m.main()
m.sys.argv = ["graphify", "global", "remove", ""];             m.main()
```

on v8:

```
Added '' to global graph: +1 nodes, -0 pruned.
manifest keys: ['']
node ids: ['::a']
Usage: graphify global remove <repo-tag>
remove '' exit: 1
```

The same happens for `graphify global add graphify-out/graph.json` from inside a repo, which is
the shape a user is most likely to type.

### Consequences

- Node ids carry no repo, so `repo` is `""` on every node and the id namespace of that graph
  overlaps any other empty-tag add — the isolation prefixing exists for is gone.
- `prune_repo_from_graph(G, "")` targets the whole empty-tag population, so a second empty-tag
  add silently replaces an unrelated repo's nodes.
- The store cannot be repaired through the CLI: the only entry that names the repo is `""`, and
  `remove` refuses it.

### Expected

An inferred tag is never empty, and any tag already present in the manifest can be removed.

### Fix

A paired PR infers through `distinct_repo_tags` so the two entry points cannot diverge again
(degrade `"repo"`), absolutize the source first so a relative path names the repo dir, and let
`global remove ""` address a store an earlier revision wrote.

xiongjianxu:
Fixes #3464.

### The defect

`global add`'s tag inference was `tag or source.parent.parent.name`. That is empty for any graph
that is not two levels below a named directory — `/tmp/graph.json`, or a relative
`graphify-out/graph.json`. The empty tag prefixes every node with `::`, sets `repo` to `""`, prunes
by `""` (so a second empty-tag add replaces an unrelated repo's nodes), and registers a manifest
entry `global remove` will not accept, because it reads an empty tag as a missing argument.

`merge-graphs` already infers through `distinct_repo_tags`, whose `d.name or "repo"` is exactly the
degrade missing here.

### The change

`graphify/cli.py`

- Infer through `distinct_repo_tags([source.absolute()])[0]`. Routing both entry points through the
  same helper is what keeps them from diverging again; `absolute()` (not `resolve()`) makes a
  relative path name its repo dir without rewriting symlinked prefixes — on macOS `resolve()` turns
  `/tmp/graph.json` into a `private` tag.
- `global remove` distinguishes an omitted tag from an empty one. Omitted is still a usage error;
  `graphify global remove ""` can now clean up a store an earlier revision wrote.

An explicit `--as <tag>` is unaffected, and the canonical
`<repo>/graphify-out/graph.json` layout yields the same tag as before, so existing manifests keep
resolving to the entries they already have.

### Tests

`tests/test_global_add_tag_inference.py`, exercising the real CLI through `graphify.__main__.main`:

| test | pins |
| --- | --- |
| `test_bare_graph_path_gets_a_non_empty_tag` | relative `graph.json` → non-empty tag, nodes prefixed |
| `test_relative_path_infers_from_the_resolved_repo_dir` | `graphify-out/graph.json` → the repo dir name |
| `test_a_nameless_repo_dir_degrades_to_repo` | the `"repo"` degrade the CLI inherits at the filesystem root |
| `test_explicit_as_tag_still_wins` | `--as` unchanged |
| `test_remove_without_a_tag_is_still_a_usage_error` | omitted argument still exits 1 |
| `test_remove_can_address_a_repo_registered_under_an_empty_tag` | an `""` entry becomes removable |

Three of the six fail on v8 and pass with the change (verified by reverting `cli.py` alone with
`__pycache__` cleared).

### Verification

- `ruff check graphify/cli.py tests/test_global_add_tag_inference.py` — clean.
- Full suite: 5377 passed, 94 skipped. `tests/test_ollama_retry_cap.py` is ignored (no `openai`
  installed locally) and `test_ts_normalizer_scales_linearly_on_large_files` deselected — it is a
  machine-dependent `process_time` ratio that also fails on unmodified v8 here and passes in
  isolation.

### DIFF
--- a/graphify/cli.py
+++ b/graphify/cli.py
@@ -3126,7 +3126,12 @@ def dispatch_command(cmd: str) -> None:
             if not source:
                 print("Usage: graphify global add <graph.json> [--as <repo-tag>]", file=sys.stderr)
                 sys.exit(1)
-            tag = tag or source.parent.parent.name
+            if not tag:
+                # Inferred through merge-graphs' own helper, which degrades to "repo"
+                # instead of "": an empty tag prunes by "" and registers a manifest
+                # entry no later add can address.
+                from graphify.build import distinct_repo_tags
+                tag = distinct_repo_tags([source.absolute()])[0]
             try:
                 result = _global_add(source, tag)
                 if result["skipped"]:
@@ -3140,9 +3145,11 @@ def dispatch_command(cmd: str) -> None:
             except Exception as exc:
                 print(f"error: {exc}", file=sys.stderr); sys.exit(1)
         elif subcmd == "remove":
-            tag = sys.argv[3] if len(sys.argv) > 3 else ""
-            if not tag:
+            # An omitted tag is a usage error; an explicitly empty one still has to be
+            # addressable, since earlier versions could register a repo under "".
+            if len(sys.argv) <= 3:
                 print("Usage: graphify global remove <repo-tag>", file=sys.stderr); sys.exit(1)
+            tag = sys.argv[3]
             try:
                 removed = _global_remove(tag)
                 print(f"Removed '{tag}' from global graph ({removed} nodes pruned).")
--- a/tests/test_global_add_tag_inference.py
+++ b/tests/test_global_add_tag_inference.py
@@ -1,0 +1,116 @@
+"""`graphify global add` repo-tag inference, and addressing the tag it produces.
+
+The tag was `source.parent.parent.name`, which is empty whenever the graph is not two
+levels below a named directory — `graphify global add /tmp/graph.json`, or any relative
+path such as `graphify-out/graph.json`. The empty tag then prefixes every node with
+`::`, prunes by `""`, and registers a manifest entry the `remove` subcommand rejected
+as a missing argument, so the store could not be cleaned up again.
+"""
+from __future__ import annotations
+
+import json
+
+import networkx as nx
+import pytest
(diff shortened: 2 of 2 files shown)


safishamsi:
Shipped in **v0.9.58** — landed on `v8` via cherry-pick with your authorship preserved. Thanks @xiongjianxu — graphify global add no longer infers an empty repo tag for degenerate paths. Release: https://github.com/Graphify-Labs/graphify/releases/tag/v0.9.58

PR #3469, opened 2026-09-10
fix(php): keep aliased class-import identity in multi-namespace files

DivyamTalwar:
## Summary

An aliased PHP class import loses its imported identity when the file declares more than one namespace, and can then be rewired to an unrelated class that happens to share the alias.

The resolver deliberately skips file-level namespace resolution for multi-namespace files. Class imports carried only a bare provisional endpoint, while the imported fully-qualified name and the local alias lived in a temporary file-level use map that the multi-namespace bailout discards. Once that map is gone, nothing distinguishes the imported class from any other symbol with the same tail, so generic same-label rewiring is free to pick the wrong one.

This moves identity ownership to the producer. Each class import edge carries its own imported FQN in metadata, recorded at extraction where the syntax is still available, group-use prefixes included. Resolution reads only that per-edge value, and reads it before the file-level namespace gate.

Two properties follow from putting identity on the edge rather than in a shared map:

- Repeated aliases sharing an imported basename resolve independently, because nothing correlates them.
- Declaration-level `function` and `const` imports are classified at extraction and carry no class provenance, so they are never redirected through this path.

## Depends on #3466

This branch is stacked on `harness/bf-phpalias/fix-a`, the head of #3466. Both changes touch `_import_php` and the same region of `_resolve_php_type_references`, and they are two arms of one condition: #3466 acts when the import kind is `function` or `const`, this change acts when the kind is neither. Opening them independently would have produced two patches that conflict on every hunk.

The diff shown against `v8` therefore includes #3466's two commits. Once #3466 merges, this reduces to its own two commits. Please merge #3466 first.

## What changed

| File | Change |
|---|---|
| `graphify/extract.py` | new `_php_import_fqn` helper; `_import_php` now branches on import kind |
| `graphify/extractors/resolution.py` | consume the per-edge FQN before the namespace bailout |
| `tests/test_php_type_resolution.py` | three regressions |

## Tests

Three regressions, each failing for a distinct reason:

- `keeps_imported_identity` — a class import whose alias collides with an unrelated internal class in a two-namespace file. Asserting the imported definition makes both an incomplete endpoint and a wrong-class endpoint fail visibly.
- `identity_is_per_edge` — two imports sharing an imported basename must resolve independently. Any file-wide map keyed by name or by target id collapses them.
- `does_not_retarget_symbol_import` — a control. A declaration-level function import must not be redirected through class-import identity, so this fix cannot widen into #3466's grouped-symbol path.

## How to verify

The first commit adds the tests alone and is red; the second is green.

```
python -m pytest tests/test_php_type_resolution.py -q
```

| State | Result |
|---|---|
| tests only (parent = #3466 head) | 3 failed, 19 passed |
| with the fix | 22 passed |

Reverting either source file on its own returns the same three failures, so both halves are load-bearing.

Full suite on the branch head: 5499 passed, 13 skipped. `ruff` and `py_compile` clean on all three changed files.

## Risk and rollback

No dependency change and no public output schema change. PHP class-import edges gain a `target_fqn` metadata key; consumers that ignore unknown metadata keys are unaffected. Full namespace-block resolution for inheritance and references is deliberately out of scope, and ambiguous identities are left unresolved rather than guessed. Reverting the two source hunks restores the previous behaviour.

### DIFF
--- a/.venv
+++ b/.venv
@@ -1,0 +1,1 @@
+/Users/divyamtalwar/Workspace/graphify-lab/repo/.venv
--- a/.venv
+++ b/.venv
@@ -1,1 +1,0 @@
-/Users/divyamtalwar/Workspace/graphify-lab/repo/.venv
--- a/graphify/extract.py
+++ b/graphify/extract.py
@@ -754,6 +754,20 @@ def _php_import_kind(node) -> str | None:
     return next((child.type for child in first_clause.children if child.type in kinds), None)
 
 
+def _php_import_fqn(node, source: bytes, raw: str) -> str:
+    parent = node.parent
+    if parent is not None and parent.type == "namespace_use_group":
+        declaration = parent.parent
+        if declaration is not None:
+            prefix = next(
+                (_read_text(child, source) for child in declaration.children
+                 if child.type == "namespace_name"),
+                "",
+            )
+            if prefix:
+                raw = f"{prefix}\\{raw}"
+    return raw.lstrip("\\")
+
 def _import_php(node, source: bytes, file_nid: str, stem: str, edges: list, str_path: str, scope_stack: list[str] | None = None) -> None:
     for child in node.children:
         if child.type in ("qualified_name", "name", "identifier"):
@@ -771,8 +785,13 @@ def _import_php(node, source: bytes, file_nid: str, stem: str, edges: list, str_
                     "source_location": f"L{node.start_point[0] + 1}",
                     "weight": 1.0,
                 }
-                if _php_import_kind(node) in ("function", "const"):
+                kind = _php_import_kind(node)
+                if kind in ("function", "const"):
                     edge["_php_symbol_import"] = True
+                elif kind is None:
+                    target_fqn = _php_import_fqn(node, source, raw)
+                    if target_fqn:
+                        edge["metadata"] = sanitize_metadata({"target_fqn": target_fqn})
                 edges.append(edge)
             break
 
--- a/graphify/extract.py
+++ b/graphify/extract.py
@@ -727,6 +727,33 @@ def _import_scala(node, source: bytes, file_nid: str, stem: str, edges: list, st
             break
 
 
+def _php_import_kind(node) -> str | None:
+    kinds = ("function", "const")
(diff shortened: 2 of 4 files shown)


DivyamTalwar:
## Correction pushed: the PHP resolver was rewriting other languages' import edges

An adversarial review of this branch rejected it, and it was right. `38e6e67` fixes it.

### The defect

`metadata.target_fqn` is a shared metadata key. C# `using` directives and other language extractors stamp it as well. `_resolve_php_type_references` is handed every edge in the graph, not only PHP ones, so the per-edge lookup this branch added was consuming other languages' values.

The check has to run *before* the `ref_file not in ns_by_file` gate, because emptying that map is exactly what the multi-namespace bailout does and that bailout is the bug being fixed. Hoisting it above the gate also hoisted it out of the only thing that had been scoping this function to PHP.

Measured, in a scan containing one namespaced PHP file plus `Use.kt` with `import external.lib.Widget`:

| revision | Kotlin import target | node materialised |
|---|---|---|
| `v8` | `widget` | none |
| this branch before `38e6e67` | `external_lib_widget` | `external.lib.Widget` |
| this branch after `38e6e67` | `widget` | none |

C# `using` directives were rebindable the same way.

### The fix

Provenance is now required explicitly rather than inherited from control flow. `_is_php_source` mirrors the suffix set `extract.py` already selects PHP files on, including its `.blade.php` exclusion, and the FQN is consumed only for an edge a PHP file produced. The multi-namespace behaviour this branch exists to fix is untouched, since it never depended on the gate the check sits above.

### Verification

```
python -m pytest tests/test_php_type_resolution.py -q
23 passed
```

Deleting the single `_is_php_source(ref_file)` clause turns `test_php_import_identity_does_not_touch_other_languages` red and leaves the other 22 green, so the guard is load-bearing rather than decorative.

Full suite on the branch head: 5499 passed, 13 skipped. `ruff` and `py_compile` clean.

Worth noting for reviewers: CI was green on the previous head and the pull request was marked mergeable throughout. A cross-language rebind that no existing test covers is invisible to the suite, which is why this needed a separate pass rather than a re-run.

Issue #3401, opened 2026-09-07
CREATE POLICY is silently dropped from SQL graphs — grammar has no rule, and the ERROR fallback regex only covers FUNCTION/PROCEDURE


kwunyanlee-cmd:
### Summary

On a Postgres project that uses Row Level Security, **no policy ever becomes a
node**, and nothing warns you. On my repository that is 59 `create policy`
statements (45 distinct policy names) that exist in the source and are absent
from `graph.json`.

This matters more than a normal missing-node-type bug, because of *what*
policies are. In a Supabase/Postgres app the policy **is** the authorisation
rule — it is the answer to "who may read this row". A graph that silently omits
every policy will happily answer questions about access control, and it will
answer them wrong. A traversal that returns nothing reads as "no rule exists"
rather than "this tool cannot see rules".

Version: `graphify 0.9.48`, with `graphifyy[sql]` installed (`tree_sitter_sql`
importable).

### Root cause — two layers, and the first one is the real blocker

**1. The `tree-sitter-sql` grammar has no rule for `CREATE POLICY`.**

Parsing a migration containing seven policies produces no `create_policy` node
of any kind. The statements disintegrate into loose top-level tokens, plus an
`ERROR` node:

    L96    keyword_create
    L96    keyword_on
    L96    .
    L97    keyword_for      keyword_update      keyword_to
    L98    keyword_using    (  .  (  )  keyword_and  .  (  )  )
    L99    keyword_with     keyword_check  ( ... )
    L99    ;

So there is nothing for `walk()` in `extractors/sql.py` to dispatch on. Adding
an `elif t == "create_policy":` branch would not help — that node type is never
produced.

**2. The `ERROR` fallback in `extractors/sql.py` only rescues functions and
procedures.**

`sql.py` already knows the grammar gives up on PL/pgSQL and has a regex
recovery path for it:

    elif t == "ERROR":
        # tree-sitter-sql cannot parse PL/pgSQL CREATE FUNCTION/PROCEDURE
        # bodies (OUT/INOUT params, tagged dollar quotes, PERFORM, :=) and
        # emits an ERROR node instead, silently dropping the object.
        # Regex-scan the raw text as fallback ...
        for m in re.finditer(
            r"CREATE\s+(?:OR\s+REPLACE\s+)?(?:FUNCTION|PROCEDURE)\s+" ...

Policies land in exactly the same place — `ERROR` nodes and loose tokens — but
[...]

PR #3470, opened 2026-09-10
fix(sql): emit index nodes for CREATE [UNIQUE] INDEX statements

### LINKED ISSUE #3467: `CREATE INDEX` is silently dropped from SQL graphs, the same way `CREATE POLICY` is in #3401 — 218 declarations, 0 nodes, on a real schema
## Summary

`CREATE INDEX` and `CREATE UNIQUE INDEX` produce no nodes. In the same file, `CREATE TABLE`,
`CREATE VIEW` and `CREATE FUNCTION` all land correctly, so the file is not skipped and the grammar
is loading — indexes specifically are lost.

This is the same shape as #3401 (`CREATE POLICY`), and I could not find an existing issue for the
index case, so filing separately. If you consider them one bug, feel free to close this into #3401 —
but note that PR #3456 addresses policies only, so merging it would leave this half open.

## Reproducer

One file, 19 lines of plain PostgreSQL. No project, no config.

```sql
-- schema.sql
CREATE TABLE public.profiles (
  id uuid PRIMARY KEY,
  owner uuid NOT NULL
);

CREATE INDEX profiles_owner_idx ON public.profiles (owner);

CREATE UNIQUE INDEX profiles_id_uniq ON public.profiles (id);

CREATE VIEW public.active_profiles AS SELECT * FROM public.profiles;

CREATE FUNCTION public.is_owner(u uuid) RETURNS boolean AS $$
  SELECT u = auth.uid();
$$ LANGUAGE sql;

ALTER TABLE public.profiles ENABLE ROW LEVEL SECURITY;

CREATE POLICY profiles_select_own ON public.profiles
  FOR SELECT USING (public.is_owner(owner));

CREATE POLICY profiles_update_own ON public.profiles
  FOR UPDATE USING (public.is_owner(owner));
```

```console
[graphify extract] AST extraction on 1 code files...
[graphify extract] wrote graphify-out/graph.json: 4 nodes, 4 edges, 2 communities
```

## Expected vs actual

Every node in `graph.json`:

```
'schema.sql'
'public.is_owner()'
'public.active_profiles'
'public.profiles'
```

| declaration | node |
|---|---|
| `CREATE TABLE public.profiles` | present |
| `CREATE VIEW public.active_profiles` | present |
| `CREATE FUNCTION public.is_owner` | present |
| `CREATE INDEX profiles_owner_idx` | **absent** |
| `CREATE UNIQUE INDEX profiles_id_uniq` | **absent** |
| `CREATE POLICY profiles_select_own` | absent (#3401) |
| `CREATE POLICY profiles_update_own` | absent (#3401) |

There is no warning and no error. `graphify extract` exits 0.

## Why this matters beyond the toy case

Measured on a production Next.js + Supabase codebase, 166 tracked `.sql` files, comments and
single-quoted strings stripped before counting, then matching each declaration name against node
labels in the same source file:

| category | declared | in graph |
|---|---|---|
| `CREATE TABLE` | 139 | 136 |
| `CREATE VIEW` | 42 | 38 |
| `CREATE FUNCTION` | 188 | 188 |
| `CREATE TRIGGER` | 122 | 113 |
| **`CREATE INDEX`** | **218** | **0** |
| **`CREATE POLICY`** | **440** | **0** |

The part that makes this more than a coverage gap: **per-file coverage reads as a clean pass.** All
166 `.sql` files produced at least one node, so every file-level coverage check says SQL is fully
mapped, while two entire entity classes are absent. Nothing in the output signals the hole.

And the graph does not go quiet when asked about the missing entities — it answers from whatever
else it has. `graphify query "which RLS policies protect the profiles table"` on that repo returned
394 nodes drawn from test files, a Drizzle schema symbol, UI components and prose documentation,
citing zero `.sql` sources. A plausible substitute answer is harder to catch than an empty one.

## Environment

- graphify 0.9.57 (`uv tool install 'graphifyy[sql,leiden,mcp,watch,svg]'`)
- Python 3.12.14, macOS (arm64)
- `tree_sitter_sql` imports cleanly in the tool env — verified before filing, since without that
  extra `extract_sql` returns zero nodes for a different reason

## Possible cause

`graphify/extractors/sql.py` handles `create_table` and `create_view` node types explicitly, and the
ERROR-recovery regex `_ROUTINE_RECOVERY_RX` covers only `FUNCTION`/`PROCEDURE`. Searching that file
for `index` turns up no rule, the same way searching it for `policy` turns up none. If the grammar
emits an ERROR node for `CREATE INDEX`, the recovery path has nothing to catch it.

L4XB:
## Summary

Fixes #3467.

`CREATE INDEX` / `CREATE UNIQUE INDEX` produced no nodes although `CREATE TABLE`, `CREATE VIEW` and `CREATE FUNCTION` in the same file did. Unlike `CREATE POLICY` (#3401, #3456), the grammar parses the statement fine into a `create_index` node; the walk in `graphify/extractors/sql.py` simply had no branch for it.

## Fix

- New `create_index` branch: the index name (the `identifier`, or a quoted `literal`, before `ON`) becomes a node contained by the file; the `object_reference` after `ON` links it to its table with an `indexes` edge. A table defined in another file gets the usual sourceless stub (#2324), like triggers do.
- `IF NOT EXISTS`, `CONCURRENTLY`, `UNIQUE`, `USING GIN`, partial indexes and quoted names all resolve through the same children. An unnamed index (`CREATE INDEX ON t (c)`) is skipped.

The reproducer from the issue now yields:

```
NODE 'schema.sql'
NODE 'public.profiles'
NODE 'profiles_owner_idx'
NODE 'profiles_id_uniq'
NODE 'public.active_profiles'
NODE 'public.is_owner()'
EDGE profiles_owner_idx --indexes--> public.profiles
EDGE profiles_id_uniq   --indexes--> public.profiles
```

## Tests

- `test_sql_create_index_emits_index_node_linked_to_its_table` in `tests/test_multilang.py`: named, unique + `IF NOT EXISTS`, `CONCURRENTLY` on a table from another file (stub), quoted name, unnamed index skipped, no dangling edges.
- `pytest tests/` has the same failure set as `v8` locally (pre-existing terraform/timing failures); `ruff check` clean. CHANGELOG entry added under 0.9.57.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@ Full release notes with details on each version: [GitHub Releases](https://githu
 
 ## 0.9.57 (unreleased)
 
+- Fix: `CREATE INDEX` / `CREATE UNIQUE INDEX` statements now produce index nodes linked to their table by an `indexes` edge — the grammar parsed them, but the SQL walk never dispatched on `create_index`, so every index was silently dropped (#3467, thanks @lrafasouza).
 - Fix: an incremental rebuild no longer wipes cross-file project AST nodes — re-extracting one `.csproj`/`.sln` was dropping package/framework nodes of a *referenced* project (whose stub carried the referenced file's `source_file`); the AST-replacement set is now derived from the files actually extracted (#3411, thanks @hopstreax).
 - Fix: when duplicate nodes merge, the richer (more complete) node is now kept as the survivor and the losers' non-empty fields are folded in, instead of a shorter-id passing mention winning and dropping content (#3372, thanks @abhay-codes07).
 - Fix: a C# generic call site with explicit type arguments — `Get<int>(...)`, unqualified or through `this` — now resolves to the method definition instead of capturing `Get<int>` as the callee and failing to match (#3406, thanks @abhay-codes07).
--- a/graphify/extractors/sql.py
+++ b/graphify/extractors/sql.py
@@ -486,6 +486,32 @@ def extract_sql(path: Path, content: str | bytes | None = None) -> dict:
                     tbl_nid = table_nids.get(_norm_ident(tbl_name)) or _ref_stub(tbl_name)
                     _add_edge(trig_nid, tbl_nid, "triggers", line)
 
+        elif t == "create_index":
+            # CREATE [UNIQUE] INDEX [CONCURRENTLY] [IF NOT EXISTS] <name>
+            # ON <table> (...). Unlike CREATE POLICY (#3401) the grammar
+            # parses this statement fine; the walk simply never dispatched on
+            # it, so every index was silently dropped (#3467). The name is the
+            # identifier (or a quoted literal) before ON; the table is the
+            # object_reference after it. An unnamed index (`CREATE INDEX ON
+            # t (c)`) has nothing to name a node after and is skipped.
+            index_name: str | None = None
+            index_table: str | None = None
+            after_on = False
+            for c in node.children:
+                if c.type == "keyword_on":
+                    after_on = True
+                elif not after_on and index_name is None and c.type in ("identifier", "literal"):
+                    index_name = _read(c).strip('"`')
+                elif after_on and index_table is None and c.type == "object_reference":
+                    index_table = _read(c)
+            if index_name:
+                index_nid = _make_id(stem, index_name)
+                _add_node(index_nid, index_name, line)
+                if index_table:
+                    index_tbl_nid = (table_nids.get(_norm_ident(index_table))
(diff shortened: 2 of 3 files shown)


safishamsi:
Shipped in **v0.9.58** — landed on `v8` via cherry-pick with your authorship preserved. Thanks @L4XB — SQL extraction now emits index nodes for CREATE [UNIQUE] INDEX, linked to their table. Release: https://github.com/Graphify-Labs/graphify/releases/tag/v0.9.58

PR #3475, opened 2026-09-10
fix: label via an installed claude CLI, and stop pinning a path that expires

ktsang622:
Two independent silent failures, found while repairing 16 repositories whose graphs had quietly stopped tracking their code. Neither one announced itself: one exits 0 while making the graph worse, the other stops the hook from ever running again.

## 1. `graphify label` downgrades a graph and reports success

`detect_backend()` is key-based, and `claude-cli` is the one backend with no API key to find — `_get_backend_api_key` can never return one for it, and the fallback loop excludes it by name. Nothing else looks for it.

So on a machine with the Claude Code CLI installed and no API key:

```
$ graphify label .
[graphify label] no LLM backend configured; keeping Community N placeholders. ...
Done - 114 communities. GRAPH_REPORT.md, graph.json and graph.html updated.
$ echo $?
0
```

Every real community name is replaced with `Community 1`, `Community 2`, … and the command exits 0. Over an existing graph that is a downgrade wearing the appearance of success — worse than the hub-derived names a bare rebuild leaves behind. `--backend=claude-cli` works perfectly; it just has to be named, and nothing tells you that.

**The fallback is in the labelling path, deliberately not in `detect_backend()`.** Widening detection was my first attempt and it broke `test_mixed_repo_without_key_errors_and_points_at_code_only`: extraction refuses to run without a configured backend and points at `--code-only`, and on any machine with the CLI present that contract would have silently become "shell out to the CLI instead". That seemed clearly not yours to change by accident, so it is unchanged, and `test_detection_is_not_widened_by_the_labelling_fallback` now pins it.

The fallback is also last, for the same reason ollama is: a configured paid key must never be shadowed by an incidental CLI on `PATH`.

## 2. The post-commit hook pins a path with an expiry date

`_pinned_python()` writes `sys.executable` verbatim. graphify installed from inside a snap-confined editor lives under `~/snap/<app>/<revision>/`, and snap swaps that revision on update and prunes the old tree.

When an editor snap moved past revision 259, **15 repositories lost their hook at once**. Each commit printed:

```
[graphify hook] could not locate a Python with graphify installed. Add the graphify bin dir to PATH or re-run 'graphify hook install' from the env where graphify lives.
```

…and the graphs simply stopped following the code. One had drifted by more than 200 communities' worth of structure before anyone noticed.

Two changes, both extending what is already there:

- `_pinned_python()` declines to pin a path under a rotating revision directory. Its docstring already describes an empty return as safe degradation — the other probes take over.
- The uv-tools probe, which already scans tool environments, also scans snap-confined `HOME`s **by glob**. An install made inside a snap lands in that snap's private `HOME`, which the plain `$HOME` roots never see once the hook runs from an ordinary shell — so before this, the pin was the only thing that could ever find such an install, and the pin was the thing guaranteed to expire.

Note `~/snap/<app>/current` is included but not relied on: on the machine where this was found, `current` pointed at 259 while 261 and 262 existed.

## Verification

Both changes mutation-tested — removing either fails a test the other still passes.

| | failed | passed |
|---|---:|---:|
| this branch | 25 | **5119** |
| `main` | 25 | **5110** |

Same 25 failures either way (terraform fixtures and a TS scaling test, unrelated to these paths); +9 passing is exactly the nine tests added here. Run in one venv, the branch stashed and unstashed between runs, rather than assuming the pre-existing failures were pre-existing.

Happy to split this into two PRs if you would rather review them separately.

### DIFF
--- a/graphify/hooks.py
+++ b/graphify/hooks.py
@@ -101,11 +101,19 @@ fi
 # Scan the uv tool envs directly; UV_TOOL_DIR overrides the default
 # location. A tool env is adopted only if its python passes the probe, so a
 # co-installed tool without graphify never satisfies it.
+#
+# The snap roots matter because an install made from inside a snap-confined
+# editor lands in that snap's private HOME, which the plain $HOME roots above
+# never see once the hook runs from an ordinary shell. Revisions are globbed
+# rather than pinned: snap rotates them on update, which is exactly what makes
+# a pinned path unsafe (see _is_rotating_prefix).
 if [ -z "$GRAPHIFY_PYTHON" ]; then
     for _GFY_TOOLS in \
         "${UV_TOOL_DIR:-}" \
         "$HOME/.local/share/uv/tools" \
-        "$HOME/AppData/Roaming/uv/tools"; do
+        "$HOME/AppData/Roaming/uv/tools" \
+        "$HOME"/snap/*/current/.local/share/uv/tools \
+        "$HOME"/snap/*/[0-9]*/.local/share/uv/tools; do
         [ -n "$_GFY_TOOLS" ] || continue
         for _GFY_CAND in "$_GFY_TOOLS"/*/bin/python "$_GFY_TOOLS"/*/Scripts/python.exe; do
             [ -x "$_GFY_CAND" ] || continue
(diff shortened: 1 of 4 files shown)


safishamsi:
Shipped in **v0.9.58** — landed on `v8` via cherry-pick with your authorship preserved. Thanks @ktsang622 — community labelling falls back to a run-time-resolved claude CLI instead of a path that expires. Release: https://github.com/Graphify-Labs/graphify/releases/tag/v0.9.58

Issue #3474, opened 2026-09-10
`graphify install` writes to `~/.claude/CLAUDE.md` and aborts mid-install when that file isn't writable


jakob1379:
For `platform="claude"`, `install()` appends a skill-registration block to `~/.claude/CLAUDE.md`. Two problems:

1. **The block is redundant on Claude Code.** Its entire content is a pointer to `~/.claude/skills/graphify/SKILL.md` — a path Claude Code already auto-discovers. The skill loads fine without it.
2. **The write is unguarded.** If `~/.claude/CLAUDE.md` isn't writable, `install()` raises and dies *after* the skill and command files are already copied, leaving a partial install and a stack trace.

This bites anyone whose dotfiles are managed declaratively. On nix/home-manager, `~/.claude/CLAUDE.md` is a symlink into the read-only `/nix/store`. Chezmoi, stow with read-only sources, and any root-owned managed config hit the same path.

### Version

graphify 0.9.57 (`3f82bf7`, branch `v8`), Python 3.13, Linux.

### Reproduction

```bash
# Simulate a declaratively-managed CLAUDE.md
mkdir -p /tmp/ro && printf '# my rules\n' > /tmp/ro/CLAUDE.md && chmod 444 /tmp/ro/CLAUDE.md
ln -sf /tmp/ro/CLAUDE.md ~/.claude/CLAUDE.md

graphify install
```

Actual failure, reproduced on a real home-manager setup without modifying anything:

```
$ readlink -f ~/.claude/CLAUDE.md
/nix/store/vxasjkcimdlfjcz6bxmjqz9isprkvv5b-hm_AGENTS.md

$ python3 -c "open('$HOME/.claude/CLAUDE.md','a')"
PermissionError: [Errno 13] Permission denied
```

### Expected

`graphify install` completes. The skill is installed and `/graphify` works.

### Actual

Skill and command files are copied, then `install()` raises `PermissionError` at the `CLAUDE.md` write and exits non-zero. Observed on my machine: `~/.claude/skills/graphify/` (SKILL.md + 8 reference files, `.graphify_version` = `0.9.57`) was written successfully, and `~/.claude/CLAUDE.md` was untouched — a half-completed install with no summary output and no indication which steps ran.

### Code

`graphify/install.py:661-687`, inside `install()` (line 624):

```python
else:
    claude_md = Path.home() / ".claude" / "CLAUDE.md"     # :674-675
[...]

PR #3476, opened 2026-09-10
fix(install): do not abort the install when always-on registration is unwritable (#3474)

dajiaohuang:
## What this fixes

`graphify install` aborts with a traceback when the always-on registration file
cannot be written, **after** the skill files have already been copied — a
half-completed install (reported in #3474).

`install()` lays down the skill first and registers the always-on block
afterwards, so this ordering is what turns a write failure into a partial
install:

- `graphify/install.py:682` / `:686` — unguarded `claude_md.write_text(...)`
- `graphify/install.py:698` / `:702` — the identical unguarded pattern for
  `CODEBUDDY.md`

It bites the normal declaratively-managed-dotfiles case: nix/home-manager
symlinks `~/.claude/CLAUDE.md` into a read-only `/nix/store`, and chezmoi or
stow with read-only sources leave the same shape.

## The change

Both registrations now route through one helper, `_register_always_on_block()`,
which catches `OSError` (covers `PermissionError`, `IsADirectoryError` and a
dangling symlink's `FileNotFoundError`), reports the skip on stderr, and lets
the install finish so the skill that was already installed stays usable.

The success-path output is unchanged byte for byte — the helper takes the
already-formatted label prefix, so the `CLAUDE.md        ->  ` column alignment
is preserved.

## Scope note

#3474 raises two things. This PR addresses only the second one (the unguarded
write). The first — that the block is redundant on Claude Code because the
harness already auto-discovers `~/.claude/skills/graphify/SKILL.md` — is a
behaviour change that would stop `install()` writing the file at all. That is
the maintainer's call, so I deliberately left it alone. Hence `Refs #3474`
rather than `Fixes`: the crash-on-install is fixed, the redundancy question
stays open.

I extended the same helper to the CodeBuddy registration because it is the same
defect, in the same function, three lines below — leaving it unguarded would
have looked like an oversight. Happy to split it out if you would rather keep
the diff to the claude path only.

## Verification

`tests/test_install.py`, three new tests:

- `test_install_survives_unwritable_claude_md` — the reported case.
- `test_install_survives_unwritable_codebuddy_md` — same shape, CodeBuddy path.
- `test_install_claude_md_success_output_unchanged` — regression guard that the
  writable path still prints `created at ` / `already registered (no change)`.

Both failure tests were confirmed to **fail on the parent commit** with exactly
the reported error before the fix:

```
tests/test_install.py:96: PermissionError: [Errno 13] Permission denied: ...\.codebuddy\CODEBUDDY.md
2 failed, 101 deselected
```

Results (`47041ee` vs parent `adf814f`), on the install-related surface
(`test_install`, `test_codebuddy`, `test_claude_md`, `test_install_roundtrip`,
`test_install_upgrade`, `test_install_strings`, `test_install_references`,
`test_install_version_stamp`, `test_install_version_warning`,
`test_uninstall_scope`, `test_agents_platform`, `test_antigravity_install`):

| | failed | passed |
|---|---|---|
| parent `adf814f` | 6 | 256 |
| this PR `47041ee` | 6 | 259 |

The same 6 pre-existing failures on both sides, so this change introduces none.
They are Windows-environment failures on platforms this PR does not touch
(`test_hermes_skill_destination_posix_uses_home`,
`test_skill_roundtrip_at_real_destination[user-hermes]`,
`test_gemini_install_references_all_resolve`,
`test_bare_call_still_removes_global[gemini]`,
`test_remove_user_skill_opt_in_with_project_dir[gemini]`,
`test_codex_skill_uses_graphify_with_existing_graph`) — the POSIX-path /
symlink caveats the README already documents for native Windows.

CI-parity guards, all green on this branch: `tools.skillgen --check` (134
artifacts), `--audit-coverage`, `--schema-singleton`, `--monolith-roundtrip`,
`--always-on-roundtrip`. `ruff check .` also passes.

> Caveat, stated plainly: I ran the full `pytest tests/ -q` suite but it did not
> finish within my working window, so the table above is the install-related
> subset rather than the whole suite. Nothing outside that surface imports
> `_register_always_on_block`; the helper is new and its only two call sites are
> the ones shown in the diff.

### DIFF
--- a/graphify/install.py
+++ b/graphify/install.py
@@ -356,6 +356,34 @@ def _skill_registration(skill_path: str = "~/.claude/skills/graphify/SKILL.md") 
         "When the user types `/graphify`, use the installed graphify skill "
         "or instructions before doing anything else.\n"
     )
+def _register_always_on_block(target: Path, prefix: str, registration: str) -> None:
+    """Append an always-on registration to *target*, degrading instead of raising.
+
+    The skill files are copied before this runs, so a *target* that cannot be
+    read or written must not abort an otherwise-complete install (#3474). That
+    happens whenever the dotfile is managed declaratively -- nix/home-manager
+    symlinks ``~/.claude/CLAUDE.md`` into a read-only /nix/store, and chezmoi or
+    stow with read-only sources leave the same shape.
+    """
+    try:
+        if target.exists():
+            content = target.read_text(encoding="utf-8")
+            if "graphify" in content:
+                print(f"{prefix}already registered (no change)")
+            else:
+                target.write_text(content.rstrip() + registration, encoding="utf-8")
+                print(f"{prefix}skill registered in {target}")
+        else:
+            target.parent.mkdir(parents=True, exist_ok=True)
+            target.write_text(registration.lstrip(), encoding="utf-8")
+            print(f"{prefix}created at {target}")
+    except OSError as exc:
+        print(f"{prefix}skipped: {exc.__class__.__name__}: {exc}", file=sys.stderr)
+        print(
+            f"  hint: the skill files were installed; add the graphify block to "
+            f"{target} manually to finish always-on registration",
+            file=sys.stderr,
+        )
 _PLATFORM_CONFIG: dict[str, dict] = {
     "claude": {
         "skill_file": "skill.md",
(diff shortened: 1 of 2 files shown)


safishamsi:
Shipped in **v0.9.58** — landed on `v8` via cherry-pick with your authorship preserved. Thanks @dajiaohuang — graphify install no longer aborts when the always-on registration target is unwritable. Release: https://github.com/Graphify-Labs/graphify/releases/tag/v0.9.58

PR #3478, opened 2026-09-10
Fix same-file source path collision during incremental merge

### LINKED ISSUE #3472: Same file, two path forms: absolute source_file from semantic extraction collides with the relative AST id and the semantic node is dropped
## Summary

A third trigger for the `minted by two different files` warning, distinct from #3322 (name
normalization within a file) and #3352 (two files mentioning one entity): **one file, two path
forms.** The semantic extraction prompt requires an absolute `source_file`, the AST layer and
`graph.json` store repo-relative paths, and the collision check compares the raw strings — so
the same file is treated as two files and one node is dropped.

## Repro

Any `--update` run where the semantic pass re-covers a file the AST layer already produced
heading nodes for. `references/extraction-spec.md` instructs each subagent:

> set source_file to the path of the originating file EXACTLY as it appears in FILE_LIST —
> verbatim and absolute. Do NOT shorten to a basename, do NOT re-relativize

while the AST layer writes `docs/ARCHITECTURE.md`. Both mint `docs_architecture_retrieval` for
the same heading. `build_merge` then reports:

```
[graphify] WARNING: node 'docs_architecture_retrieval' is minted by two different files —
keeping 'Retrieval' from 'docs/ARCHITECTURE.md',
dropping 'Hybrid Retrieval' from '/abs/path/to/docs/ARCHITECTURE.md'.
... the dropped node is lost.
```

23 of these in a single incremental update over a ~5,600-node TypeScript corpus (17 changed
markdown files). Every one names the same file on both sides, differing only in `/abs/prefix/`.

## Why it matters more than the node count suggests

The dropped side is the **semantic** node, so the surviving label is the raw heading text and
the richer label plus its `rationale` attribute are what get discarded — `Retrieval` survives,
`Hybrid Retrieval` is lost; `Guardrails` survives, `Refusal in Two Layers, and Block-Buffered
Streaming` is lost. The warning's own remedy ("run `graphify extract` per subfolder and merge
with `graphify merge-graphs`") does not apply, since there is only one file involved.

## Suggested fix

Canonicalize `source_file` to the build root before the collision check, rather than after it —
`build_merge` already takes `root=` and relativizes downstream, so the comparison just runs too
early. Alternatively, treat two records whose canonicalized paths agree as the same file and
merge their attributes (preferring the longer label / the one carrying `rationale`) instead of
dropping one.

Version: 0.9.56, `directed=False` graph, `build_merge` via the documented `--update` runbook.

hopstreax:
## Summary

Fixes #3472, where the same source file could be represented by both an absolute and repo-relative `source_file` path during an incremental `--update`.

This caused `build_merge` to incorrectly report a "minted by two different files" collision and drop the semantic node, including richer attributes such as `rationale` and `summary`.

## Root cause

Semantic extraction emits the absolute `source_file` path from `FILE_LIST`, while existing AST nodes loaded from `graph.json` use repo-relative paths.

Deduplication ran before path normalization, so:

```text
docs/ARCHITECTURE.md
/absolute/path/to/repo/docs/ARCHITECTURE.md
```

were treated as different source files.

## Fix

* Normalize `source_file` and `definition_file` against the build root before deduplication.
* Propagate the effective merge root from `build_merge` into `build`, including when the root is inferred.
* Reuse the existing path normalization logic.
* Preserve existing deduplication and survivor-selection semantics.

Same-file duplicates now correctly merge complementary semantic attributes while genuine cross-file ID collisions remain isolated.

## Tests

* absolute semantic path + relative AST path
* inverse arrival order
* inferred merge root
* genuine cross-file collisions
* `definition_file` normalization

Relevant test suite:

* `240 passed, 1 skipped`
* #3472 regression tests: `5 passed`

### DIFF
--- a/graphify/build.py
+++ b/graphify/build.py
@@ -1426,6 +1426,7 @@ def build(
         combined["hyperedges"].extend(ext.get("hyperedges", []))
         combined["input_tokens"] += ext.get("input_tokens", 0)
         combined["output_tokens"] += ext.get("output_tokens", 0)
+    _root = str(Path(root).resolve()) if root else None
     if dedup and combined["nodes"]:
         # Numeric ids must be str before dedup, which keys on them and would
         # raise TypeError in _pick_winner's regex search (#2326). build_from_json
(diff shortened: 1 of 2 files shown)


safishamsi:
Shipped in **v0.9.58** — landed on `v8` via cherry-pick with your authorship preserved. Thanks @hopstreax — an incremental rebuild no longer drops rationale/summary on an abs-vs-relative source_file collision. Release: https://github.com/Graphify-Labs/graphify/releases/tag/v0.9.58

PR #3479, opened 2026-09-10
fix(rust): extract module-level static and const declarations

### LINKED ISSUE #3471: Rust: module-level `static` and `const` declarations produce no nodes — 0 of 783 names in a `jdx/mise` snapshot
### Version / environment

- `graphifyy==0.9.57` (PyPI wheel, clean venv)
- macOS arm64, Python 3.14
- Also present at the `v0.9.57` tag commit `3f82bf7f837a07fb0f7668fbdbd5662801906942`

### Summary

The Rust extractor emits no graph node for module-level `static` or `const` declarations.
Functions, structs, enums, traits and impl methods from the same file are extracted normally,
so the declaration is silently absent rather than obviously broken.

In a mixed-language repository this omits the declaration's provenance and skews query
vocabulary: a constant's name can end up in the graph **only** from shell scripts or test
fixtures that reference it, never from the Rust that defines it.

### Minimal reproduction

```sh
tmp="$(mktemp -d)" && mkdir -p "$tmp/repro/src"

cat >"$tmp/repro/Cargo.toml" <<'EOF'
[package]
name = "graphify-rust-bindings-repro"
version = "0.1.0"
edition = "2021"
EOF

cat >"$tmp/repro/src/lib.rs" <<'EOF'
pub static RETRY_LIMIT: usize = 3;
pub const DEFAULT_MODE: &str = "fast";

pub struct Settings { pub mode: String }
pub enum Mode { Fast, Slow }

pub fn retry_limit() -> usize { RETRY_LIMIT }
EOF

python3 -m venv "$tmp/venv"
"$tmp/venv/bin/pip" install -q "graphifyy==0.9.57"
"$tmp/venv/bin/graphify" extract "$tmp/repro" --code-only --force

REPRO_DIR="$tmp/repro" "$tmp/venv/bin/python" <<'PY'
import json, os
from pathlib import Path
g = json.loads((Path(os.environ["REPRO_DIR"])/"graphify-out"/"graph.json").read_text())
labels = {n.get("label") for n in g["nodes"] if (n.get("source_file") or "").endswith(".rs")}
for want in ("RETRY_LIMIT", "DEFAULT_MODE", "Settings", "Mode", "retry_limit()"):
    print(f"{want:16s} {'PRESENT' if want in labels else 'ABSENT'}")
PY
```

### Actual

```text
RETRY_LIMIT      ABSENT
DEFAULT_MODE     ABSENT
Settings         PRESENT
Mode             PRESENT
retry_limit()    PRESENT
```

`Settings`, `Mode` and `retry_limit()` are the positive controls: the file is parsed and the
other declaration kinds in it are extracted in the same run.

### Expected

`RETRY_LIMIT` and `DEFAULT_MODE` should each get a node sourced from `src/lib.rs`, with a
file-level `defines` (or `contains`) edge, alongside the struct, enum and function.

### Cause

[`graphify/extractors/rust.py:206`](https://github.com/Graphify-Labs/graphify/blob/3f82bf7f837a07fb0f7668fbdbd5662801906942/graphify/extractors/rust.py#L206)
gates every non-function item on:

```python
```

There is no branch for tree-sitter-rust's `static_item` or `const_item`. Grepping the
installed package for either name returns 0 hits (control: `struct_item` 2, `function_item` 3,
`enum_item` 2). `extract_rust`'s own docstring (`rust.py:62`) states the contract as
*"functions, structs, enums, traits, impl methods, and use declarations"*.

### Quantified impact

Measured against a `jdx/mise` snapshot (tag `v2026.9.4`, commit `794948606e02cde091653a165291b06cae43c352`),
extracted with `graphify extract <dir> --code-only --force`:

- 617 Rust files represented in the graph, contributing 17,159 nodes.
- 879 module-level `static`/`const` declaration sites → 783 distinct names.
- **0 of those 783 names appear as an exact-label node sourced from any Rust file.**
- `src/env.rs` alone contributed 101 nodes — 99 functions, 1 type, 1 file node — and no
  `static`/`const` node, despite 105 `static` and 14 `const` declaration lines in it.

The 29 of those names that do appear anywhere in the graph are sourced from shell, Ruby and
fixture files that reference them (`e2e/**`, `benchmarks/**/*.sh`, `src/assets/bash/activate.sh`),
because the Bash extractor *does* node top-level exported variables
([`extractors/bash.py:501-513`](https://github.com/Graphify-Labs/graphify/blob/3f82bf7f837a07fb0f7668fbdbd5662801906942/graphify/extractors/bash.py#L501-L513)).
So a query for e.g. `MISE_SHIMS_DIR` returns test fixtures and not the declaration:

```rust
pub(crate) static MISE_LOG_FILE: Lazy<Option<PathBuf>> =
    Lazy::new(|| var_path("MISE_LOG_FILE"));
```

### Prior art

- **#2266** reported the equivalent omission for exported JS/TS scalar bindings; fixed by
  **#2300**, released in v0.9.31.
- **#2360** / **#2364** / **#3098** cover the same class for Go package-level `var`/`const`
  and are still open.
- **#3366** / **#3367** are the Rust precedent: a missing Rust declaration kind reported with
  corpus measurements, then fixed.
- **#1972** separately covers calls/constant reads in top-level context getting no caller
  node. Related but **not** required to fix this; see below.

### Suggested fix scope

1. Handle `static_item` and `const_item` in `graphify/extractors/rust.py`, emitting a node
   per declared name plus a file-level edge.
2. Add Rust extractor tests for module-level `pub static` and `pub const`.
3. Leave initializer call-edges out of scope — `Lazy::new(|| var_path("X"))` producing no
   `calls` edge follows from the call harvest walking only function bodies
   (`rust.py:431`, `for caller_nid, body_node in function_bodies:`) and belongs to #1972.

---

L4XB:
## Summary

Fixes #3471.

`extract_rust` walked `struct_item`, `enum_item`, `trait_item`, `impl_item` and the function node types, but had no branch for tree-sitter-rust's `static_item` / `const_item`. A `static` or `const` declaration therefore produced no node: its name reached the graph only through other files that referenced it, never from the Rust that defines it (the issue measured 0 of 783 such names in a `jdx/mise` snapshot).

## Change

`graphify/extractors/rust.py`: a `static_item` / `const_item` branch in `walk`.

- Module level: node labelled with the declaration name plus a file-level `contains` edge, matching structs and enums.
- Associated const inside an `impl` block: attributed to the impl like a method (`.NAME`, `contains` edge from the impl node).
- The declared type is referenced through `_rust_collect_type_refs` with `context="field"` / `"generic_arg"`, the same way a struct field type is, so `const NODE_LIMIT: Limit = …` links to `Limit`. Primitive types are skipped by the existing helper, so `usize` / `&str` add no noise.

The docstring now lists statics/consts; CHANGELOG entry under 0.9.57.

## Tests

`tests/fixtures/sample.rs` gains `pub static RETRY_LIMIT`, `pub const DEFAULT_MODE`, `const NODE_LIMIT: Limit` and an associated `const CAPACITY` inside `impl Graph`. `test_rust_finds_static_and_const_items` checks the three module-level nodes and their `contains` edges, `.CAPACITY` under `Graph`, and the `NODE_LIMIT → Limit` reference. Fails on `v8` (no nodes), passes with the change; the other Rust tests are unchanged (18 passed, 78 in `tests/test_multilang.py`). `ruff check` clean.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@ Full release notes with details on each version: [GitHub Releases](https://githu
 
 ## 0.9.57 (unreleased)
 
+- Fix: Rust module-level `static` and `const` declarations (and associated consts inside an `impl`) are now extracted as nodes with a `contains` edge and a reference to their declared type — neither node type had a branch, so a constant only ever reached the graph through files that referenced it (#3471, thanks @sortakool).
 - Fix: an incremental rebuild no longer wipes cross-file project AST nodes — re-extracting one `.csproj`/`.sln` was dropping package/framework nodes of a *referenced* project (whose stub carried the referenced file's `source_file`); the AST-replacement set is now derived from the files actually extracted (#3411, thanks @hopstreax).
 - Fix: when duplicate nodes merge, the richer (more complete) node is now kept as the survivor and the losers' non-empty fields are folded in, instead of a shorter-id passing mention winning and dropping content (#3372, thanks @abhay-codes07).
 - Fix: a C# generic call site with explicit type arguments — `Get<int>(...)`, unqualified or through `this` — now resolves to the method definition instead of capturing `Get<int>` as the callee and failing to match (#3406, thanks @abhay-codes07).
--- a/graphify/extractors/rust.py
+++ b/graphify/extractors/rust.py
@@ -59,7 +59,7 @@ _RUST_TRAIT_METHOD_BLOCKLIST: frozenset[str] = frozenset({
 })
 
 def extract_rust(path: Path) -> dict:
-    """Extract functions, structs, enums, traits, impl methods, and use declarations from a .rs file."""
+    """Extract functions, structs, enums, traits, impl methods, statics/consts, and use declarations from a .rs file."""
     try:
         import tree_sitter_rust as tsrust
         from tree_sitter import Language, Parser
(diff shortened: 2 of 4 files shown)


safishamsi:
Shipped in **v0.9.58** — landed on `v8` via cherry-pick with your authorship preserved. Thanks @L4XB — Rust module-level static and const declarations are now extracted as nodes. Release: https://github.com/Graphify-Labs/graphify/releases/tag/v0.9.58

PR #3483, opened 2026-09-10
fix(packaging): include psycopg[binary] in the all extra

### LINKED ISSUE #3482: `graphifyy[all]` omits `psycopg[binary]`, so the documented “Everything above” install cannot use `--postgres`
## Summary

The `all` optional dependency is documented in README as “Everything above,” but it omits `psycopg[binary]`, which is required by the `postgres` extra and by `graphify extract --postgres`.

Confirmed on v0.9.55 and still present on v0.9.57 (`3f82bf7f837a07fb0f7668fbdbd5662801906942`), i.e. across two subsequent releases.

## Current metadata

`pyproject.toml` defines:

```toml
postgres = ["psycopg[binary]", "tree-sitter-sql"]
```

The `all` list contains `tree-sitter-sql` but not `psycopg[binary]`.

A mechanical comparison of `all` against the union of every other named extra reports only:

```text
missing_from_all:
psycopg[binary]
```

## Result

An environment without `psycopg` exits when Postgres introspection starts:

```text
error: psycopg is required for --postgres. Install with: pip install 'graphifyy[postgres]'
```

This is a clear error rather than silent data loss, but it contradicts the documented meaning of `graphifyy[all]` and requires users who selected “all” to install another extra.

## Expected

`graphifyy[all]` should include both dependencies from the `postgres` extra.

## Suggested fix

Add `psycopg[binary]` to the `all` list and add a metadata test asserting that `all` contains the union of every other optional-dependency group. At v0.9.57 that union check identifies no other omissions.

A related packaging gap was closed in #3047, where the `postgres` extra lacked `tree-sitter-sql` and `--postgres` silently reported zero nodes. A union test would catch both classes.

L4XB:
Fixes #3482.

README documents `graphifyy[all]` as "Everything above", but the `all` list carried only `tree-sitter-sql` from the `postgres` extra and not `psycopg[binary]`, so an `[all]` install still exited with `psycopg is required for --postgres` when Postgres introspection started. Comparing `all` against the union of every other extra on `v8` reports exactly that one omission:

```
missing on v8: ['psycopg[binary]']
```

**Change**

- `pyproject.toml`: `psycopg[binary]` added to `all` (next to `tree-sitter-sql`).
- `tests/test_backend_extras.py::test_all_extra_is_the_union_of_every_other_extra`: asserts `[all]` contains every dependency of every other extra, so the next new extra cannot drift the same way (#3047 closed the same class of gap in the other direction). The test fails on `v8` with the message above and passes here.
- CHANGELOG entry under 0.9.57.

`uv.lock` is untouched; let me know if you want it regenerated in this PR.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@ Full release notes with details on each version: [GitHub Releases](https://githu
 
 ## 0.9.57 (unreleased)
 
+- Fix: `graphifyy[all]` now includes `psycopg[binary]`, so the documented “Everything above” install can run `graphify extract --postgres` without a second extra; a packaging test asserts `[all]` is the union of every other extra (#3482, thanks @GREYGROUPJP).
 - Fix: an incremental rebuild no longer wipes cross-file project AST nodes — re-extracting one `.csproj`/`.sln` was dropping package/framework nodes of a *referenced* project (whose stub carried the referenced file's `source_file`); the AST-replacement set is now derived from the files actually extracted (#3411, thanks @hopstreax).
 - Fix: when duplicate nodes merge, the richer (more complete) node is now kept as the survivor and the losers' non-empty fields are folded in, instead of a shorter-id passing mention winning and dropping content (#3372, thanks @abhay-codes07).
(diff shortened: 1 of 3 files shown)


safishamsi:
Shipped in **v0.9.58** — landed on `v8` via cherry-pick with your authorship preserved. Thanks @L4XB — the all extra now includes psycopg[binary]. Release: https://github.com/Graphify-Labs/graphify/releases/tag/v0.9.58

Issue #3487, opened 2026-09-11
Monorepo: workspace subpath imports are not resolved through the target package's exports map, dropping cross-package edges


cbartens:
## Summary

Imports of a workspace package subpath — `@scope/pkg/components/Icon` — are not resolved, because the importing package's dependency is never mapped to the target package's `exports` map. On a 21-package pnpm monorepo this drops **323 import edges** from 1,233 workspace imports.

Related to #3486 but independent: that one is a suffix-rewriting bug in relative resolution, this one is a resolution path that does not exist. Both are silent, and both bias `affected` and any reachability pass the same way.

## Reproduction

```
packages/ui/package.json
  { "name": "@acme/ui",
    "exports": { "./components/Icon": { "default": "./src/components/Icon.tsx" } } }
packages/ui/src/components/Icon.tsx        export function Icon() {}

apps/web/src/Page.tsx                       import { Icon } from "@acme/ui/components/Icon";
```

Expected: `Page.tsx → packages/ui/src/components/Icon.tsx`. Actual: no edge; `Icon.tsx` has zero inbound.

The bare-root form (`import { x } from "@acme/ui"`) has the same problem when the package uses `exports` rather than `main`.

## Three details that matter for a fix

**1. `types` must not be used as the target.** The near-universal shape is:

```json
"./components/Icon": {
  "types":        "./dist/components/Icon.web.d.ts",
  "react-native": "./src/components/Icon.native.tsx",
  "default":      "./src/components/Icon.web.tsx"
}
```

`types` points into `dist/`, which is build output and normally excluded from the corpus, so resolving to it creates an edge to a node that does not exist.

**2. The condition depends on the importer, not the package.** An importer under a React Native app resolves `react-native`; a web importer resolves `default`. Picking one globally attributes `Icon.web.tsx` to mobile code that never imports it. Picking all of them everywhere inflates fan-in. Cheapest correct-enough rule is to choose by the importing file's location/platform suffix and fall back to the union when it is ambiguous.

**3. Wildcard targets may not exist on disk.** `"./controls/*": { "default": "./src/controls/*.tsx" }` where the real files are `BottomNav.native.tsx` and `BottomNav.web.tsx` — the bundler resolves that through `sourceExts`/`resolveExtensions`. An exports target that does not exist should fall through to the platform-suffixed siblings rather than being dropped.

Node's own subpath matching applies: exact key first, then a single `*` with a fixed prefix and suffix.

## Measured impact

[...]

PR #3488, opened 2026-09-11
fix(extract): resolve .js/.mjs/.cjs specifiers through TypeScript's full source order (#3486)

### LINKED ISSUE #3486: JS/TS: a .js import specifier never resolves to a .tsx file, dropping every NodeNext-style import of a TSX module
## Summary

`_resolve_js_import_path` maps a `.js` import specifier to `.ts` and stops. TypeScript's documented resolution order for a `.js` specifier is `.ts` → `.tsx` → `.d.ts` → `.js`, so every NodeNext/Node16-style import of a `.tsx` file is dropped from the graph.

On a 1,430-file TypeScript monorepo this loses **171 import edges**, and presents as `.tsx` files appearing unreachable.

## Where

`graphify/extractors/resolution.py` (branch `v8`, 0.9.58 — also 0.9.50):

```python
# TS ESM convention: imports often spell .js/.jsx while source is .ts/.tsx.
if candidate.suffix == ".js":
    ts_candidate = candidate.with_suffix(".ts")
    if ts_candidate.is_file():
        return ts_candidate
elif candidate.suffix == ".jsx":
    tsx_candidate = candidate.with_suffix(".tsx")
    if tsx_candidate.is_file():
        return tsx_candidate
```

The comment names the right convention, but `.js` only ever tries `.ts`. A `.tsx` source file is imported as `.js`, not `.jsx` — TypeScript emits `Foo.tsx` to `Foo.js`, so the specifier the compiler requires is `./Foo.js`. `./Foo.jsx` is only correct under `jsx: preserve`, which is the rarer case.

The append-extension loop underneath does not rescue it, because it appends to the full filename rather than replacing the suffix:

```python
for ext in _JS_RESOLVE_EXTS:
    with_ext = candidate.parent / f"{candidate.name}{ext}"   # MessageBubble.js.tsx
```

## Reproduction

```
src/Button.tsx          export function Button() {}
src/App.ts              import { Button } from "./Button.js";
```

with `"moduleResolution": "nodenext"`. Expected: an import edge `App.ts → Button.tsx`. Actual: none — `Button.tsx` has zero inbound edges.

Directly against the resolver:

```python
>>> from graphify.extractors.resolution import _resolve_js_import_path as R
>>> from pathlib import Path
>>> R(Path("src/Button.js"))       # Button.tsx exists on disk
PosixPath('src/Button.js')          # unresolved
>>> R(Path("src/helper.js"))       # helper.ts exists on disk
PosixPath('src/helper.ts')          # fine
```

## Measured impact

Real monorepo, 1,430 source files, 10,958 nodes / 19,266 edges:

| Relative `.js` specifiers | Count |
|---|---|
| resolve to `.ts` | 1,041 |
| resolve to `.tsx` — **edge lost** | **171** |

The asymmetry is visible in the graph as a large difference in apparent reachability between two extensions under one import convention:

| Non-test source | Files | Zero inbound | Rate |
|---|---|---|---|
| `.ts` | 472 | 47 | 10% |
| `.tsx` | 185 | 82 | **44%** |

After repairing only this one case out-of-band, `.tsx` drops to 15% and 54 files stop reading as unreachable.

This matters beyond edge counts: it makes `affected`, any dead-code pass, and any reachability question quietly wrong for `.tsx` files, in the direction that is hardest to notice — a missing edge looks like a clean result.

## Suggested fix

Replace the branch with a suffix→candidates table, preserving the existing order semantics:

```python
_TS_OUTPUT_EQUIVALENT = {
    ".js":  (".ts", ".tsx", ".d.ts"),
    ".jsx": (".tsx",),
    ".mjs": (".mts", ".d.mts"),
    ".cjs": (".cts", ".d.cts"),
}

for replacement in _TS_OUTPUT_EQUIVALENT.get(candidate.suffix, ()):
    sibling = candidate.with_suffix("")  # .with_suffix(".d.ts") mangles the stem
    rewritten = sibling.parent / f"{sibling.name}{replacement}"
    if rewritten.is_file():
        return rewritten
```

Note `Path.with_suffix(".d.ts")` does not do what it looks like it does on a multi-dot stem, hence building the name by concatenation.

`.mjs`/`.cjs` are included for completeness; I have no repro for those, only for `.js` → `.tsx`.

Happy to open a PR with this plus a test if that is useful.

dajiaohuang:
Fixes #3486.

`_resolve_js_import_path` mapped a `.js` specifier to `.ts` and stopped. TypeScript compiles `Button.tsx` to `Button.js`, so under NodeNext/Node16 the specifier the compiler *requires* is `./Button.js` while the source on disk is `.tsx` — and TypeScript resolves that specifier in the order `.ts` → `.tsx` → `.d.ts` → `.js`. Only the first step was implemented, so every `.tsx` module imported that way dropped out of the graph.

## Before (v8, 0.9.58)

Resolver alone, one target on disk at a time:

| specifier | on disk | resolved to |
|---|---|---|
| `Button.js` | `Button.tsx` | `Button.js` — unresolved |
| `types.js` | `types.d.ts` | `types.js` — unresolved |
| `mod.mjs` | `mod.mts` | `mod.mjs` — unresolved |
| `mod.cjs` | `mod.cts` | `mod.cjs` — unresolved |
| `helper.js` | `helper.ts` | `helper.ts` — fine |

End to end on the issue's own fixture, `App.ts` importing `./Button.js` emitted a `ref_button_js` phantom instead of an edge to the `Button.tsx` node — the same asymmetry the issue measured as 171 lost edges on a 1,430-file monorepo.

## Change

- `graphify/extractors/resolution.py` — replace the `.js`-only branch with a suffix → ordered-candidates table (`_TS_SOURCE_SUFFIXES`), so each emitted specifier is tried against every source suffix TypeScript accepts, implementation always ahead of a declaration. A real `.js`/`.jsx` file on disk still wins, because the existence check short-circuits before any rewriting. `.jsx → .tsx` folds into the same table and behaves exactly as before.
- `tests/test_import_extension_resolution.py` — 11 regression locks: the four resolution cases, ordering (`.ts` beats `.tsx` beats `.d.ts`), multi-dot stems (`tag-action.shared.js → tag-action.shared.tsx`), two non-regression guards, and the issue's reproducer end to end. Nine of the eleven fail on `v8` and pass here; the other two pin behaviour that must **not** change.

## One correction to the issue's suggested fix

The proposal builds `.d.ts` by concatenating onto `with_suffix("")`, on the assumption that `Path.with_suffix(".d.ts")` mangles a multi-dot stem. It does not — `Path("Button.test.js").with_suffix(".d.ts")` is `Button.test.d.ts` on 3.10 and 3.13 alike — so this uses `with_suffix` directly. No behaviour difference; just less code.

## Validation

- `pytest tests/` → **5136 passed, 286 skipped**, with the same 42 pre-existing failures as `v8` on Windows (FIFO/socket/symlink cases, tree-sitter-terraform, ollama deps). The failure set is byte-identical to the `v8` baseline; a re-run gives 5137 passed with the identical failure set. (One timing-ratio test, `test_ts_normalizer_scales_linearly_on_large_files`, flaked once under full-suite load on this machine and passes 5/5 in isolation and on re-run — it is unrelated to this path.)
- `python -m tools.skillgen --check` / `--schema-singleton` / `--audit-coverage` → all OK.

No new dependencies, no public API change.

---

I noticed @cbartens offered to upstream their fuller resolution logic in the issue. This PR deliberately implements only the file-level ordering from the issue's own table, and not the workspace `exports`-map / platform-split work — that is the separate #3487 — so it stays small and does not collide with a larger contribution there.

### DIFF
--- a/graphify/extractors/resolution.py
+++ b/graphify/extractors/resolution.py
@@ -31,6 +31,17 @@ _PACKAGE_IMPORTS_CACHE: "dict[str, tuple[Path, dict] | None]" = {}
 
 _JS_RESOLVE_EXTS = (".ts", ".tsx", ".mts", ".cts", ".svelte", ".js", ".jsx", ".mjs", ".cjs")
 
+# TypeScript's resolution order for a specifier written against the *emitted*
+# module: `Foo.tsx` compiles to `Foo.js`, so the specifier the compiler requires
+# is `./Foo.js` even though the source on disk is `.tsx`. Tried in order, an
+# implementation always ahead of a declaration (#3486).
+_TS_SOURCE_SUFFIXES = {
+    ".js": (".ts", ".tsx", ".d.ts"),
+    ".jsx": (".tsx",),
+    ".mjs": (".mts", ".d.mts"),
+    ".cjs": (".cts", ".d.cts"),
+}
+
 _JS_INDEX_FILES = ("index.ts", "index.tsx", "index.svelte", "index.js", "index.jsx", "index.mjs")
 
 def _resolve_js_import_path(candidate: Path) -> Path:
@@ -39,15 +50,13 @@ def _resolve_js_import_path(candidate: Path) -> Path:
     if candidate.is_file():
         return candidate
 
-    # TS ESM convention: imports often spell .js/.jsx while source is .ts/.tsx.
-    if candidate.suffix == ".js":
-        ts_candidate = candidate.with_suffix(".ts")
-        if ts_candidate.is_file():
-            return ts_candidate
-    elif candidate.suffix == ".jsx":
-        tsx_candidate = candidate.with_suffix(".tsx")
-        if tsx_candidate.is_file():
-            return tsx_candidate
+    # TS ESM convention: imports often spell the emitted module while the source
+    # is a TypeScript file. Try every source suffix TypeScript accepts for this
+    # specifier, in its documented order (#3486).
+    for replacement in _TS_SOURCE_SUFFIXES.get(candidate.suffix, ()):
+        source_candidate = candidate.with_suffix(replacement)
+        if source_candidate.is_file():
+            return source_candidate
 
     # Append extensions to the full filename, which covers extensionless imports,
     # multi-dot helpers, and Svelte 5 rune files like Foo.svelte.ts.
--- a/tests/test_import_extension_resolution.py
+++ b/tests/test_import_extension_resolution.py
@@ -161,6 +161,99 @@ def test_resolve_jsx_to_tsx_when_real_file_is_tsx(tmp_path):
     assert _resolve_js_module_path(written_as) == target
 
 
+# ── #3486: a `.js` specifier resolves through TypeScript's full order ────────
+
+
+def test_resolve_js_to_tsx_when_real_file_is_tsx(tmp_path):
+    """#3486 reproducer. TypeScript emits `Button.tsx` as `Button.js`, so under
+    NodeNext/Node16 the specifier the compiler *requires* is `./Button.js`
+    while the source on disk is `.tsx`. The `.js` branch only ever tried `.ts`
+    and stopped, so every such import dropped to a phantom `ref_` edge."""
+    target = _write(tmp_path / "Button.tsx", "export const x = 1")
+    written_as = tmp_path / "Button.js"
+    assert _resolve_js_module_path(written_as) == target
+
+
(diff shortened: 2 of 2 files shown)


dajiaohuang:
Checked against the issue contract and TypeScript's emitted-module source order: a .js specifier resolves through .ts, .tsx, then .d.ts; .jsx is not a candidate for .js. The existing .jsx → .tsx behavior remains covered separately, so no code change is needed for this advisory.

PR #3489, opened 2026-09-11
fix(imports): pick workspace exports by importer platform, runtime before types (#3487)

### LINKED ISSUE #3487: Monorepo: workspace subpath imports are not resolved through the target package's exports map, dropping cross-package edges

dajiaohuang:
Fixes #3487.

## What actually goes wrong

The `exports` map **is** consulted â€” #1308 added that, in `e8dabad`. The defect is in how a condition object is reduced to one target.

`_EXPORT_CONDITION_PRIORITY` shipped as:

```python
("source", "import", "module", "svelte", "types", "require", "default")
```

`types` is a declaration-only condition and normally points into `dist/`, which is build output and not in the corpus. So for

```json
"./components/Icon": {
  "types": "./dist/components/Icon.web.d.ts",
  "react-native": "./src/components/Icon.native.tsx",
  "default": "./src/components/Icon.web.tsx"
}
```

the resolver returned the `dist` path, found nothing on disk, and **stopped**. There was no fall-through, because `_package_entry_candidates` returned `[candidate]` as soon as any condition matched â€” so the `default` source sitting right beside it was never tried, and the import became a dangling `ref_*` external.

That is not the order the map was added with. The commit that introduced it says *"`default` is consulted LAST â€¦ so a matching `import`/`module`/`svelte` condition wins"*, and #1308 specifies `source -> import -> module -> default -> require -> types`. The tuple put `types` ahead of `require` and `default` instead.

Two more gaps in the same path: `react-native` was never consulted, and no platform-suffixed sibling was ever tried, so `"./controls/*": {"default": "./src/controls/*.tsx"}` could not reach the real `BottomNav.web.tsx` / `BottomNav.native.tsx`.

## The fix

- **`types` moves to the end** of `_EXPORT_CONDITION_PRIORITY`, restoring the order #1308 specified. A runtime target always beats a declaration. `require`/`default` keep their relative order, so the behaviour pinned by `test_workspace_subpath_export_default_consulted_last` is unchanged.
- **`_resolve_export_targets`** keeps *every* condition target in preference order rather than returning the first, so a target that is not on disk falls through to the next condition instead of taking the import down with it.
- **`react-native` is honoured only for a native importer**, so a web importer resolves the same specifier through `default` rather than being attributed a native file it never imports. This is a custom condition, so it is per-importer by nature.
- **Each target is tried alongside its platform-suffixed siblings** (`Icon.tsx` â†’ `Icon.web.tsx` / `Icon.native.tsx`), which is how a bundler reaches a platform split whose `exports` target names the unsuffixed path.
- `_resolve_export_target` keeps its signature and semantics for the `package.json` `imports` path; platform expansion applies only to the workspace `exports` path, so relative imports elsewhere are untouched.

## Behaviour notes for review

- When every `exports` target escapes the package directory, resolution still falls through to the bare-path fallback exactly as before. A new test pins this â€” my first revision accidentally swallowed that fallback, and the containment guard made it reachable.
- The `"."` branch now applies the same containment guard the subpath branch already applied. This is the only change beyond the reported defect, and it is a consistency fix: such a package previously resolved outside its own directory.
- Importer platform is inferred from whole path segments (`apps/mobile/src` â†’ native; `packages/web-utils` â†’ none, since matching is segment-wide, not substring). It is best-effort by design: when nothing distinguishes the importer every platform keeps the same priority, so resolution stays deterministic rather than dropping the edge.

## Verified

Measured through `_resolve_js_module_path` on the fixture from the issue, before and after:

| Case | Specifier | Importer | Before | After |
|---|---|---|---|---|
| A | `@acme/ui/components/Icon` (`{types, react-native, default}`) | web | unresolved | `Icon.web.tsx` |
| B | same | native | unresolved | `Icon.native.tsx` |
| C | `@acme/ui2/controls/BottomNav` (wildcard, target absent) | web | unresolved | `BottomNav.web.tsx` |
| C2 | same | native | unresolved | `BottomNav.native.tsx` |
| D | `@acme/lib/util` (exact subpath, target exists) | web | resolved | resolved (control) |
| E | `@acme/root` (bare root, `"."` exists) | web | resolved | resolved (control) |
| F | `@acme/nodot` (exports has no `"."`) | web | unresolved | unresolved (control) |

The controls were already correct before the change, which localises the defect to condition selection and fall-through rather than to a missing resolution path.

Seven tests added to `tests/test_js_import_resolution.py`, one per behaviour above plus locks on the ordering and the containment fall-through. Focused run: `tests/test_js_import_resolution.py` + `tests/test_import_extension_resolution.py` â†’ 109 passed (PR #3488's tests still pass unchanged). `ruff check` clean. `tools.skillgen` CI-parity checks all OK.

Full suite, same commands CI runs:

```
before:  29 failed, 5500 passed, 44 skipped in 227.16s
after:   29 failed, 5507 passed, 44 skipped in 238.04s
```

Byte-identical failure sets â€” `diff` of the sorted `FAILED` lines is empty. The 29 are pre-existing and unrelated to this path (non-regular-file, watch and uninstall-scope tests on Windows). The +7 passing are the tests added here. `graphify --help` and `graphify install` both exit 0.

## Note on #3487's reporter

@cbartens reports in the issue that they have this working and offered to contribute it. I could not tell from the thread whether maintainers want that, so I am putting this up as a concrete patch â€” happy to close it in favour of theirs, or to have it serve as a reference implementation. No maintainer has replied on the issue, and nothing here is claimed or assigned.

### DIFF
--- a/graphify/extractors/resolution.py
+++ b/graphify/extractors/resolution.py
@@ -413,26 +413,89 @@ def _load_workspace_packages(start_dir: Path) -> dict[str, Path]:
     _WORKSPACE_PACKAGE_CACHE[key] = packages
     return packages
 
+# `types` is LAST. It is a declaration-only condition that normally points into
+# `dist/`, which is build output and not part of the corpus, so letting it win
+# creates an edge to a node that does not exist while the runtime target the
+# source actually lives behind is skipped (#3487). This is the order #1308
+# specified when the exports map was added; the tuple had `types` ahead of
+# `require`/`default` instead.
 _EXPORT_CONDITION_PRIORITY = (
-    "source", "import", "module", "svelte", "types", "require", "default",
+    "source", "import", "module", "svelte", "require", "default", "types",
 )
+
+# Conditions a bundler opts into per platform. `react-native` is a custom
+# condition, so it is honoured only for an importer that is itself native — a
+# web importer resolves the same specifier through `default`, or the native file
+# is attributed to code that never imports it (#3487).
+_PLATFORM_EXPORT_CONDITIONS = {"native": ("react-native",)}
+
+_PLATFORM_SUFFIXES = ("web", "native", "ios", "android")
+
+# Whole-path-segment hints only: `apps/mobile/src` is native, `packages/web-utils`
+# is not web, so segment-wide matching keeps this from firing on package names.
+_PLATFORM_PATH_HINTS = (
+    ("native", ("native", "mobile", "ios", "android", "react-native")),
+    ("web", ("web", "browser", "desktop", "electron")),
+)
+
+def _importer_platform(start_dir: Path) -> str | None:
+    """Best-effort platform of the importing file, from its directory path.
+
+    Exports conditions and platform-suffixed files are chosen per importer
+    (#3487), and `start_dir` is the importer's directory — the filename itself
+    is not available at this point, so a platform-named segment is the signal.
+    Returns None when nothing distinguishes the importer, which leaves every
+    platform at the same priority afterwards."""
+    for part in reversed(start_dir.parts):
+        segment = part.lower()
+        for platform, hints in _PLATFORM_PATH_HINTS:
+            if segment in hints:
+                return platform
+    return None
+
+def _platform_variants(candidate: Path, platform: str | None) -> list[Path]:
+    """Platform-suffixed siblings of an `exports` target, importer's first.
+
+    `"./controls/*": {"default": "./src/controls/*.tsx"}` names `BottomNav.tsx`,
+    which does not exist — the real files are `BottomNav.web.tsx` and
+    `BottomNav.native.tsx`, and a bundler finds them through its platform list.
+    The importer's own platform wins; the rest stay in a fixed order so an
+    ambiguous importer resolves deterministically rather than dropping the edge
+    entirely (#3487)."""
+    order = [p for p in (platform,) if p in _PLATFORM_SUFFIXES]
+    order.extend(p for p in _PLATFORM_SUFFIXES if p not in order)
+    return [
+        candidate.parent / f"{candidate.stem}.{p}{candidate.suffix}"
+        for p in order
+    ]
+
+def _resolve_export_targets(value: Any, platform: str | None = None) -> list[str]:
+    """Every target an `exports` value offers, in preference order.
+
+    Keeping the whole list rather than the first match lets a target that is
+    not on disk fall through to the next condition instead of dropping the
+    import: a package whose `import`/`default`/`types` split across `dist/` and
+    `src/` still resolves to the one target that is real source (#3487).
+    Only conditions in _EXPORT_CONDITION_PRIORITY are considered, plus the
+    platform's own conditions when the importer has a platform."""
+    if isinstance(value, str):
+        return [value]
+    if isinstance(value, dict):
+        targets: list[str] = []
+        for cond in (*_PLATFORM_EXPORT_CONDITIONS.get(platform or "", ()),
+                     *_EXPORT_CONDITION_PRIORITY):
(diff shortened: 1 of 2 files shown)


safishamsi:
Shipped in **v0.9.60** — landed on `v8` via cherry-pick with your authorship preserved. Thanks @dajiaohuang — package.json exports resolve by importer platform, runtime before types. Release: https://github.com/Graphify-Labs/graphify/releases/tag/v0.9.60

PR #3491, opened 2026-09-11
explain: add path::Symbol resolution and fix the ambiguity retry hint

### LINKED ISSUE #3485: explain: the ambiguity message says retry with the repo-relative path, but a path resolves to the file node, not the symbol; there is no path-scoped symbol form
### Prior art, so this is not a duplicate

- #3176 (closed 2026-08-29) reported the silent pick and, under "Attempted workaround (does not work)", tried `path::run_sweep` and got "No node matching". The fix that closed it added the ambiguity guard and the id list. The qualifier half was not addressed, and the new message now recommends it.
- #2467 (fixed) repaired the *other* half of the same hint: the printed node id could not be fed back into `explain`. This issue is the same shape for the "repo-relative path" half.
- #569 (open) proposes `explain --file <path> [--line N]` for monorepos; that flag would also resolve this. This issue is narrower: the message today points at a form that resolves to the wrong node.

### Repro

```
src/components/Widget.tsx
  export function Widget({ label }: { label: string }) { return <span>{label}</span>; }
  export function helper(n: number): number { return n + 1; }

src/screens/Home.tsx
  import { Widget, helper } from '../components/Widget';
  function Widget() { return <div>local</div>; }
  export function HomeScreen(flag: boolean) {
    const a = flag ? helper(1) : helper(2);
    const jobs = [];
    if (flag) jobs.push(helper(3));
    return a + jobs.length;
  }
```

`git init && git add -A && git commit -m init && GRAPHIFY_FORCE=1 graphify update . --force`

### What happens

```
$ graphify explain Widget --graph graphify-out/graph.json
Ambiguous: 'Widget' matches 2 nodes in different files.
  src/screens/Home.tsx
    id: src_screens_home_widget
  src/components/Widget.tsx
    id: src_components_widget_widget
Retry with the repo-relative path or the full node id.

$ graphify explain "src/components/Widget.tsx::Widget" --graph graphify-out/graph.json
No node matching 'src/components/Widget.tsx::Widget' found.

$ graphify explain "src/components/Widget.tsx:Widget" --graph graphify-out/graph.json
No node matching 'src/components/Widget.tsx:Widget' found.

$ graphify explain "src/components/Widget.tsx" --graph graphify-out/graph.json
Node: Widget.tsx
  ID:        src_components_widget
  ...
```

The bare path does resolve, but to the **file** node: `serve._find_node_tiers` has a `source_exact` tier that matches the tokenized query against `source_file`, and no tier that combines a path with a label. So the message's first suggestion yields either "No node matching" or a different node than the one asked about. Only the `id:` works, and an id is not something a reader types from memory.

### Expected

Either, ideally both:

1. A path-scoped symbol form in `_find_node`, e.g. `path::Symbol` (the form #3176 already tried), restricting the label match to nodes whose `source_file` equals the path. `#569`'s `--file` flag is the same capability spelled as an option.
2. Until then, the message says what works: "Retry with one of the node ids above."

### Where

- `graphify/cli.py`, `explain` branch: prints "Retry with the repo-relative path or the full node id."
- `graphify/serve.py`, `_find_node_tiers`: `source_exact` is a file-node tier.

### Why it matters

The guard from #3176 is good: it stops a same-named local symbol (`Home.tsx` declares its own `Widget`) from silently answering for the exported one. The follow-up advice sends the reader one call sideways in exactly the situation the guard was built for.

Version: graphify 0.9.57 (`graphifyy` on PyPI), Python 3.12.3, Linux.

ayushcodes10:
## Summary

Fixes #3485. `explain`'s ambiguity message told callers to retry with "the repo relative path", but a bare path resolves to the FILE node, not the symbol, and there was no query form that scoped a symbol to a specific file.

- Adds a `path::Symbol` query form, resolved in `_find_node_tiers` before the ordinary tiers, so `explain` and the MCP tools `get_node`/`get_neighbors` (which share the same resolver) can all disambiguate a symbol by file.
- Updates the ambiguity message in both the `explain` CLI and the shared MCP resolver to recommend `path::symbol` instead of the old, misleading suggestion, using a generic `<path>::symbol` placeholder so the message stays identical regardless of node iteration order (there is an existing test guarding exactly that property).

## Test plan

- [x] Added 3 regression tests in `tests/test_explain_cli.py`: the exact ambiguous scenario from the issue resolves to the right node either way, a symbol not present in the given file falls through to "no match" cleanly, and a genuine `Foo::Bar` style label (Rust/C++ conventions) still resolves via ordinary matching when nothing before `::` matches a file path.
- [x] Manually reproduced the issue's exact repro and confirmed the new message and resolution end to end.

### DIFF
--- a/graphify/cli.py
+++ b/graphify/cli.py
@@ -1750,7 +1750,10 @@ def dispatch_command(cmd: str) -> None:
             for rival in rivals:
                 print(f"  {G.nodes[rival].get('source_file') or rival}")
                 print(f"    id: {rival}")
-            print("Retry with the repo-relative path or the full node id.")
+            print(
+                f"Retry with path::symbol using one of the paths above (e.g. "
+                f"<path>::{label}) or the full node id."
+            )
             sys.exit(1)
         nid = matches[0]
         d = G.nodes[nid]
--- a/graphify/serve.py
+++ b/graphify/serve.py
@@ -1537,7 +1537,8 @@ def _resolve_single_node(G: nx.Graph, label: str) -> tuple[str | None, str | Non
         return None, (
             f"Ambiguous: '{label}' matches {len(rivals)} nodes in different files.\n"
             f"{listing}\n"
-            "Retry with the repo-relative path or the full node id."
+            f"Retry with path::symbol using one of the paths above (e.g. "
+            f"<path>::{label}) or the full node id."
         )
     return matches[0], None
 
--- a/graphify/serve.py
+++ b/graphify/serve.py
@@ -1335,6 +1335,55 @@ def _query_graph_text(
     return header + _subgraph_to_text(traversal_graph, nodes, edges, token_budget, seeds=start_nodes)
 
 
+def _resolve_path_scoped_symbol(G: nx.Graph, path_part: str, symbol_part: str) -> list[str]:
+    """Nodes whose source_file matches path_part and label/id matches symbol_part.
+
+    Backs the `path::Symbol` query form (#3485): a bare path resolves to the
+    FILE node (`_find_node_tiers`'s own `source_exact` tier, which prefers
+    the file over its members once #2032-disambiguated), and a bare symbol
+    name can be ambiguous across files -- the same-named local declaration
+    guard from #3176 exists for exactly that case, and its own suggested
+    retry ("the repo-relative path") pointed at a form that resolved to the
+    wrong node or nothing at all, since no prior tier combined a path with
+    a label. This combines both constraints in one query, so a specific
+    symbol in a specific file is reachable without needing its opaque id.
+    """
+    path_tokens = " ".join(_search_tokens(path_part))
+    norm_path_query = _strip_diacritics(path_part).lower().strip()
+    symbol_term = " ".join(_search_tokens(symbol_part))
+    norm_symbol_query = _strip_diacritics(symbol_part).lower().strip()
+    if not (path_tokens or norm_path_query) or not (symbol_term or norm_symbol_query):
+        return []
+    candidate_ids = _trigram_candidates(G, [symbol_term, norm_symbol_query])
+    node_iter = (
+        G.nodes(data=True) if candidate_ids is None
+        else ((nid, G.nodes[nid]) for nid in candidate_ids)
+    )
+    matches: list[str] = []
+    for nid, d in node_iter:
+        source_file = d.get("source_file") or ""
+        source_tokens = " ".join(_search_tokens(source_file))
+        norm_source = _strip_diacritics(source_file).lower().strip()
+        if not (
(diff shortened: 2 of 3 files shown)


ayushcodes10:
Thanks for the review. Went through both advisory findings and the formal verification section.

**Both "Worth a look" findings were real** — reproduced the collision directly: a native `::` label (e.g. a Rust `mylib::Bar`) could be hijacked by the new `path::Symbol` resolution when the substring before `::` happens to match an unrelated file elsewhere in the graph (an extensionless file literally named the same as the module), returning a node from that unrelated file instead of the one the literal label names. Fixed in the latest commit: `path::Symbol` resolution now only runs when the raw, unsplit label doesn't already have a literal exact match among real node labels — a genuine label essentially never equals a whole `path::symbol` string verbatim, so this only ever suppresses the false-positive case. Added a regression test (`test_explain_double_colon_label_not_hijacked_by_a_coincidental_path_match`) covering the exact collision shape, and reconfirmed the existing Rust/C++ non-regression test and the original #3485 disambiguation case still pass.

**The formal verification section's two "behavior changes" are intentional, not regressions** — both are the same underlying fix: `_find_node_tiers` previously returned a bare `[]` for an empty/punctuation-only query, which crashed any caller unpacking it into 4 tiers (`_resolve_single_node` included, hence "raises ValueError" in the old column). That's a pre-existing latent bug, unrelated to #3485, fixed in passing since I was directly editing the adjacent line. The new behavior — returning `(None, "No node matching '...' found.")` instead of crashing — is correct and confirmed working as intended.

ayushcodes10:
Re the `rstrip('()')` finding: correct that `str.rstrip` strips a character set repeatedly rather than one trailing `()` pair, so e.g. `'foo(())'.rstrip('()')` -> `'foo'`, not `'foo()'`. But this isn't something my diff introduces — `_label_has_literal_exact_match`'s `bare_label = norm_label.rstrip("()")` is a direct copy of the exact same line already used by the `exact` tier a few lines below it (and elsewhere in this file), which is deliberate: the whole point of this pre-check is to ask "would the raw label match via the same rule the `exact` tier itself uses", so the two need to share the identical `bare_label` computation or the guard could disagree with the tier it's meant to predict. Fixing the stripping precision only in my new function would desync it from the tier loop; fixing it everywhere is a real but separate, pre-existing issue in this file's matching logic, unrelated to #3485 — happy to file it separately if useful, but leaving it out of this PR.

safishamsi:
Shipped in **v0.9.59** — landed on `v8` via cherry-pick with your authorship preserved. Thanks @ayushcodes10 — graphify explain now accepts a path::Symbol form and the ambiguity hint shows a form the resolver accepts. Release: https://github.com/Graphify-Labs/graphify/releases/tag/v0.9.59

Issue #1173, opened 2026-06-07
post-commit hook overwrites .graphify_root, silently expanding a scoped build to the whole repo


dispather:
## Summary

After a **scoped build** (`graphify src/`), installing the post-commit hook (`graphify hook install`) causes the next commit to re-extract **all changed code files via `git diff`**, ignoring the scoped `src/` boundary, and **overwrites `.graphify_root`** from `.../src` to `.` (repo root). The graph silently expands beyond the user's intended scope, and there is no warning.

## Reproduction (measured on a TS router project)

1. `graphify src/` → `graphify-out/.graphify_root` = `/abs/path/src`, **214 nodes / 512 edges / 12 communities**.
2. `graphify hook install`.
3. First commit that includes new `scripts/*.py` and `*.ts` (outside `src/`).
4. Hook auto-rebuilds. Result:
   - `.graphify_root` is **overwritten to `.`** (1 byte).
   - Graph grows to **374 nodes / 667 edges / 30 communities** — `scripts/` was absorbed.

## Why this is surprising / hard to detect

- The user explicitly scoped to `src/`, yet `scripts/` ends up in the graph after a single commit. ("Why did `scripts/` enter the graph when I only built `src/`?")
- mtime-based staleness checks **cannot** catch it: `graph.json` mtime is fresh — what changed is the *coverage scope*, not freshness.

## Suggested directions

- The hook should **respect an existing `.graphify_root`** (not overwrite it), restricting the `git diff` re-extraction to the scoped root.
- Or, if the whole-repo behavior is intended, **warn at `graphify hook install`** time that the hook re-extracts the entire repo on every commit and will override a scoped root.
- Optionally a `graphify hook install --respect-root` flag.

## Environment

- graphify: PyPI `graphifyy`, 0.8.x
- flow: scoped build (`graphify src/`) → `hook install` → first commit
- shell: fish

safishamsi:
Fixed in 6a549e4. Post-commit and post-checkout hook bodies now read `graphify-out/.graphify_root` before calling `_rebuild_code`, falling back to `Path('.')` if absent. A scoped build (`graphify src/`) no longer gets silently expanded to the full repo on the next commit.

PR #3492, opened 2026-09-11
hooks: keep the rebuild root inside the repo (#3265)

### LINKED ISSUE #3265: Git hooks execute an interpreter path read from graphify-out/.graphify_python, a directory the README recommends committing
## Summary

The `post-commit` / `post-checkout` hooks written by `graphify hook install` resolve their
Python interpreter through a chain of probes. The second probe reads a **path from a file
inside the working tree** — `graphify-out/.graphify_python` — and then executes that path.

The README (["Team setup"](https://github.com/Graphify-Labs/graphify#team-setup)) says
`graphify-out/` *"is meant to be committed to git so everyone on the team starts with a map"*,
and the recommended `.gitignore` block excludes only `cost.json` (and optionally `cache/`).
`.graphify_python` is therefore a file the documentation actively encourages committing.

Put together: cloning a repository that followed that advice, and running `graphify hook install`
in it, means the repository's own contents decide which executable your git hooks run on your
next commit. That is a decision that should belong to the machine, not to the checkout.

## Where

From a hook generated by 0.9.53 (`.husky/post-commit`, same block in `post-checkout`):

```sh
_PINNED='/…/uv/tools/graphifyy/bin/python'
if [ -n "$_PINNED" ] && [ -x "$_PINNED" ] && "$_PINNED" -c "$_GFY_PROBE" 2>/dev/null; then
    GRAPHIFY_PYTHON="$_PINNED"
fi
# Second probe: read graphify-out/.graphify_python …
if [ -z "$GRAPHIFY_PYTHON" ]; then
    _GFY_PYTHON_FILE="graphify-out/.graphify_python"
    …
            GRAPHIFY_PYTHON="$_FROM_FILE"
```

The existing guards raise the bar but do not close it: the character allowlist rejects shell
metacharacters, not paths; `-x` and the `find_spec('graphify')` probe constrain *what* the
target must look like, not *where it may come from* or *who wrote it*.

## Why the probe looks redundant

`hook install` already embeds `_PINNED` at install time, and the README documents that as the
mechanism ("`graphify hook install` embeds the current interpreter path directly into the hook
scripts at install time… If you reinstall or upgrade graphify, re-run `graphify hook install`").
On a normal setup `_PINNED` answers first and the working-tree probe never runs — so its cost
today is the trust it extends, not the problem it solves.

## Suggested directions

1. Drop the working-tree probe from the hooks and let `_PINNED` → `command -v graphify` cover it.
   Re-running `hook install` after an upgrade is already the documented fix for a stale pin.
2. If the file has to stay as a fallback, source it from outside the repository
   (`~/.config/graphify/` or `~/.graphify/`), which survives `uv tool` reinstalls just as well
   without being checkout-controlled.
3. Failing both, add `graphify-out/.graphify_python` to the recommended `.gitignore` in
   "Team setup", so the documented workflow stops publishing it.

Happy to send a PR for (1) or (2) if you have a preference.

## Environment

graphify 0.9.53, macOS 15 (Darwin 25.4.0), `uv tool` install, hooks via husky.
No impact observed locally: across 34 repositories here `graphify-out/` is git-ignored in all 34
and no `graphify-out/` file is tracked. Reporting it as a property of the documented team
workflow rather than as an incident.

ayushcodes10:
## Summary

Fixes #3265. `.graphify_root` lives inside the directory Team setup documents committing, so its contents are checkout controlled — but the generated post-commit / post-checkout hooks read that file and pass its value straight to `_rebuild_code` with no bound. A value planted there by a malicious fork or PR (an absolute path outside the repository) would steer the rebuild, and therefore what it reads and what it writes back into that same committed directory, wherever the checkout names, not the repository the hook was installed into.

Both rebuild bodies (`_REBUILD_BODY_COMMIT` and `_REBUILD_BODY_CHECKOUT` in `graphify/hooks.py`) now only adopt the saved root when it resolves inside the working tree the hook is actually running from (`_root.resolve() == cwd` or a descendant of it); anything else falls back to the repo top, the previous default when the marker was absent, and prints a note saying so.

This does not narrow the legitimate case: a subdirectory scoped root (#1173, e.g. `graphify update backend/`) is still a descendant of the repo root the hook runs from, so it's still honoured — covered by a new test.

This PR only addresses the `.graphify_root` half of the report. The reporter's earlier `.graphify_python` follow-up explicitly asks for a maintainer preference among three directions (drop the working tree probe, relocate it outside the repo, or document it in `.gitignore`), so I've left that to a maintainer call rather than guessing.

## Test plan

- [x] Added 2 regression tests (parametrized over both hook bodies) in `tests/test_hooks.py`: one confirms a `.graphify_root` pointing outside the repo is rejected and falls back to the repo top, one confirms a subdirectory scoped root (#1173) is still honoured. Both execute the actual shipped snippet (extracted from the body text via regex) rather than a hand copy, so they can't drift from the real hook source.
- [x] Manually reproduced: wrote `.graphify_root` pointing at a directory outside the repo, confirmed the recovered root falls back to the repo top; wrote a subdirectory value, confirmed it's still honoured.
- [x] Full suite: `python3 -m pytest -q` — 5424 passed, only the pre-existing unrelated failures (`test_ollama_retry_cap.py` missing `openai` in this env, one flaky timing assertion in `test_ts_import_type_arguments.py`).
- [x] Existing hook tests all pass unchanged, including the shell quote safety and prefix consistency checks the new code had to satisfy.

### DIFF
--- a/graphify/hooks.py
+++ b/graphify/hooks.py
@@ -188,7 +188,17 @@ try:
     if _saved.exists():
         _txt = _saved.read_text(encoding='utf-8-sig').strip()
         if _txt:
-            _root = Path(_txt)
+            _candidate = Path(_txt)
+            try:
+                _cwd = Path.cwd().resolve()
+                _resolved = _candidate.resolve()
+                _in_repo = _resolved == _cwd or _cwd in _resolved.parents
+            except OSError:
+                _in_repo = False
+            if _in_repo:
+                _root = _candidate
+            else:
+                print(f'[graphify hook] ignoring out-of-repo .graphify_root: {_txt}')
     _rebuild_code(_root, changed_paths=changed, force=_force)
     # Refresh the work-memory lessons doc when saved Q&A outcomes exist
     # (best-effort; never fails the hook).
@@ -250,7 +260,17 @@ try:
     if _saved.exists():
         _txt = _saved.read_text(encoding='utf-8-sig').strip()
         if _txt:
-            _root = Path(_txt)
+            _candidate = Path(_txt)
+            try:
+                _cwd = Path.cwd().resolve()
+                _resolved = _candidate.resolve()
+                _in_repo = _resolved == _cwd or _cwd in _resolved.parents
+            except OSError:
+                _in_repo = False
+            if _in_repo:
+                _root = _candidate
+            else:
+                print(f'[graphify] ignoring out-of-repo .graphify_root: {_txt}')
     _rebuild_code(_root, force=_force)
     # Refresh the work-memory lessons doc when saved Q&A outcomes exist
     # (best-effort; never fails the hook).
--- a/graphify/hooks.py
+++ b/graphify/hooks.py
@@ -193,7 +193,7 @@ try:
                 _cwd = Path.cwd().resolve()
                 _resolved = _candidate.resolve()
                 _in_repo = _resolved == _cwd or _cwd in _resolved.parents
-            except OSError:
+            except (OSError, RuntimeError):
                 _in_repo = False
             if _in_repo:
                 _root = _candidate
@@ -265,7 +265,7 @@ try:
                 _cwd = Path.cwd().resolve()
                 _resolved = _candidate.resolve()
                 _in_repo = _resolved == _cwd or _cwd in _resolved.parents
-            except OSError:
+            except (OSError, RuntimeError):
                 _in_repo = False
             if _in_repo:
                 _root = _candidate
--- a/tests/test_hooks.py
+++ b/tests/test_hooks.py
@@ -2,6 +2,7 @@
 import os
 import shutil
 import subprocess
+import textwrap
 from types import SimpleNamespace
 from pathlib import Path
 import pytest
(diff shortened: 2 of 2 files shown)


ayushcodes10:
Thanks for the review.

**"Symlink-loop .graphify_root can crash generated hooks" — confirmed and fixed.** Reproduced directly: `Path.resolve()` raises `RuntimeError` for a symlink loop, not `OSError`, so the guard's `except OSError` let it propagate. It doesn't produce a raw traceback (the outer `except Exception` in the hook body catches it and reports "Rebuild failed"), but that's still wrong — a bad marker should fall back to the repo top like every other invalid value, not abort the rebuild. Fixed in the latest commit by also catching `RuntimeError`, with a regression test that builds an actual two-symlink loop and confirms the fallback now fires.

**"Checkout hook validated path is reopened after symlink race" — looked at this, leaving it as is.** This is a TOCTOU between the `resolve()` check and `_rebuild_code`'s actual use of `_root` a few lines later. I don't think it changes the threat model this PR is closing: the scenario #3265 describes (and what this guard defends against) is a value committed to the repo steering a later, unattended run of the hook. A TOCTOU exploit needs an attacker with live, precisely-timed write access to the working tree *during* the hook's execution window on the victim's own machine — at that point they already have a much more direct path to doing damage than racing a path check (they could edit the source files being rebuilt, or the hook script itself). Happy to revisit if there's a concrete scenario where that access exists without also implying worse compromise already, but I don't think a file-descriptor-based rewrite of `_rebuild_code`'s read/write path is proportionate to add here.

ayushcodes10:
Re "Out-of-repo .graphify_root is no longer honored by installed hooks" — that's the intended fix, not a regression. #3265 is exactly about an out-of-repo `.graphify_root` value being honored when it shouldn't be (a checkout-controlled file steering the rebuild outside the repository); restricting it to in-repo paths is the whole point of this PR. An already-installed hook picks up the new behavior the next time `graphify hook install` regenerates it, same as any other hook-body fix.

safishamsi:
Shipped in **v0.9.59** — landed on `v8` via cherry-pick with your authorship preserved. Thanks @ayushcodes10 — the git hook keeps its rebuild root inside the repository, ignoring a committed .graphify_root that points outside. Release: https://github.com/Graphify-Labs/graphify/releases/tag/v0.9.59

PR #3493, opened 2026-09-11
feat(html): replace vis-network with vendored force-graph renderer

jakob1379:
## Summary

The old graph.html loaded vis-network from unpkg, capped at 5k nodes, and on this repo fell back to a 1,020-node community summary. Forcing the full graph blocked the page for about 30 s and produced an unreadable blob.

graph.html now inlines force-graph 1.51.4 (MIT, Canvas 2D, bundles d3-force). It opens offline and renders this repo's full 15.9k-node graph with live physics. Graph data is unchanged: same node ids, edges, communities and labels.

## Preview

Screen recording of the new renderer on graphify's own repo (drag, pin, spotlight, presets, halo):

output

## Why force-graph

Surveyed pyvis, ipysigma, gravis, plotly, bokeh, pyecharts, graphviz, cosmograph and d3graph on the Python side, and vis-network, sigma.js v3, cytoscape.js, raw d3-force, cosmos.gl, force-graph and AntV G6 on the JS side. Static SVG was rejected: past roughly 10k marks the DOM cost makes the page sluggish, and CSS-only interactivity cannot do search, spotlight or drag.

force-graph won on size (about 178 KB minified, one file, no CDN), active maintenance, built-in d3-force physics, canvas hit-testing and drag, and a small API to build the panel on. cosmos.gl is faster but WebGL-only with mobile caveats. sigma needs graphology plus a separate layout package.

## What changed

- Community-seeded start positions plus cohesion, separation and a halo force keep communities as clusters and push disconnected pieces to the rim.
- Physics panel with graphify / forceAtlas2 / compact / spread presets, tooltipped sliders, Reheat / Unpin / Fit. Drag pins a node, right-click releases it, click spotlights its neighbours.
- Sidebar panels collapse with an animation. Communities sits above Physics.
- Viewport culling. Edges are hidden while big graphs settle and when zoomed far out.
- `MAX_NODES_FOR_VIZ` raised to 20,000 and `graphify export html` reads it from the exporter instead of hardcoding 5000.
- The vendored JS ships via `[tool.setuptools.package-data]` and is covered by the wheel packaging test.

## Known trade-offs

- Straight edges instead of vis-network's curves.
- Output is about 13.6 MB for this repo (graph data plus the inlined renderer).
- Background tabs throttle requestAnimationFrame, so the initial settle can stop early. Reheat restarts it.

## Testing

- `uv run --frozen pytest tests/test_export.py tests/test_pipeline.py tests/test_cli_export.py tests/test_wheel_packaging.py`: 249 passed
- `uv build --wheel` confirmed `graphify/exporters/vendor/force-graph.min.js` ships
- Emitted script blocks parse under node; manual check in a browser at 15.9k nodes


jakob1379:
Checked the three advisory findings against `0679681` by feeding `</script>` through `to_html` as a node label, an edge relation, and a community name. All three are false positives. The escaping is present and unchanged from the previous exporter.

- **Tooltip titles.** Node and edge `title` fields go through `html.escape` in Python before serialization (`html.py` 782, 811, 830). The force-graph HTML tooltip receives `&lt;img ...&gt;`. The newline-to-`<br>` step only touches the already-escaped string.
- **JSON in the script tag.** All four embedded arrays go through `_js_safe`, which rewrites `</` to `<\/` (line 847). The generated file contains exactly two `</script>` sequences, the two real closing tags. The raw `<img` text survives only inside JSON string literals, and those fields are rendered either on canvas via `fillText` or through the `esc()` helper.
- **Legend community label.** Legend entries are built from `html.escape(sanitize_label(...))` (line 840) before insertion with `innerHTML`.

The `_html_styles` behaviour change is intended: the sidebar stylesheet was rewritten for the collapsible panels and the physics controls.

PR #3498, opened 2026-09-11
fix(hook): cap the soft read nudge per session and suppress it after a fresh query

### LINKED ISSUE #3435: PreToolUse soft nudge has no session cap — measured 8,348 injections (~651k tokens) in a single project
## Summary

The soft `_READ_NUDGE` / `_SEARCH_NUDGE` payload is re-injected on *every* qualifying Read/Glob/Grep/Bash call, with no dedup, no session cap, and no suppression while the agent has recently queried the graph. Only the opt-in strict **deny** is capped (`_mark_session_denied`).

Across 87 Claude Code transcripts in one repo I measured **8,348 nudge injections producing 339 graphify calls** — a ~4% response rate — at a standing cost of roughly **651,000 tokens**, which is ~3.7x more than every graphify query in those sessions combined (~176,000 tokens).

This is adjacent to #580 (which fixed the `GRAPH_REPORT.md` read-first instruction) but is a separate cost: #580 was one expensive read per question, this is a cheap payload paid thousands of times.

## Environment

- graphify **0.9.38**, installed with `--strict`
- Claude Code, macOS, project-scoped `.claude/settings.json`
- Repo: Next.js 14 / TypeScript, 293 source files, graph = 3,938 nodes / 9,158 edges, 99% EXTRACTED
- Verified the relevant code is byte-identical in **0.9.56**, so this is not fixed on latest

## Measurement

Parsed the raw Claude Code transcripts (`~/.claude/projects/<project>/*.jsonl`), classifying each `tool_use` as a graphify call or a broad-exploration call (Grep/Glob/Task, Read of a source file, or Bash containing grep/rg/find/fd/ack/ag). Sessions split on session start time.

## Results

**Adoption — 77 sessions that explored the codebase:**

| Metric | Value |
|---|---|
| graphify ran *before* first exploration | 8 / 77 (10%) |
| graphify used at all | 46 / 77 (60%) |
| graphify share of exploration-related calls | 329 / 10,424 (3.2%) |

**Nudge cost:**

| | Count | Approx tokens |
|---|---|---|
| Soft nudge injections | 8,348 | ~651,000 |
| Strict deny (capped, for contrast) | 44 | negligible |
| All graphify command output | 338 calls | ~176,000 |

**What follows a graphify call** (the 10 calls after each, 48 sessions):

| | Share |
|---|---|
| grep / find / rg | 68.3% |
| Read | 31.7% |

## Root cause

In `graphify/cli.py::_run_hook_guard`, the read branch ends with an unconditional

```python
sys.stdout.write(_READ_NUDGE)
```

Every guard above it either returns early or falls through to this line. The suppression logic that exists — `_query_stamp_fresh()` and `_mark_session_denied()` — gates **only** the strict deny:

```python
if _hook_strict_enabled(strict) and tool_name in (None, "Read") \
        and not _query_stamp_fresh() \
        and _target_is_indexed(fp, root) \
        and _mark_session_denied(str(d.get("session_id") or "")):
    sys.stdout.write(_READ_DENY)
    return
sys.stdout.write(_READ_NUDGE)   # <- no cap, no stamp check, no dedup
```

So an agent that has just run `graphify query` is still told, on its very next read, that it must run `graphify query`.

## Why this matters beyond the token cost

The CHANGELOG already notes that "a nudge is advisory `additionalContext` the model routinely walks past mid-task", and #1840 documents nudges firing on unrelated files, "training agents to ignore MANDATORY language entirely". Injection volume is plausibly a *cause* of that, not just a symptom: 8,348 identical high-urgency strings with no consequence for ignoring them is efficient training to filter the string out. The ~4% response rate is consistent with that.

## Suggested fixes

Roughly in order of value-to-effort:

1. **Suppress the soft nudge while `_query_stamp_fresh()`** — the stamp already exists and is already written by query/explain/path. An agent that just oriented does not need reminding. One-line change.
2. **Cap per session with backoff** — reuse the `_mark_session_denied` marker pattern with a timestamp instead of a one-shot lock (e.g. at most once per N minutes per session id). Fail-open as today.
3. **Dedup per target** — do not re-nudge for the same file path within a session.
4. **Shorten the payload.** `_READ_NUDGE` is ~480 characters, of which the subagent-propagation sentence is ~40% and cannot be enforced from a hook anyway.

Fixes 1 and 2 alone would have cut the cost here by well over 90% while keeping the nudge present at the moments it is actually informative.

## Reproduction

Any project with a graph and the Claude Code hook installed. To count injections in your own transcripts:

```bash
grep -o "MANDATORY: graphify-out/graph.json exists" \
  ~/.claude/projects/<project>/*.jsonl | wc -l
```

Compare against the number of `graphify` invocations in the same files.

## Not claimed here

I am not claiming the graph itself is low value, and this is not a re-report of #580 — the read-first instruction is correctly gone. The claim is narrow: the nudge's own standing cost is unbounded by design, appears to be the single largest token line item the tool introduces, and is plausibly self-defeating at high volume.

Ashfaqbs:
Fixes #3435.

## Problem

The soft PreToolUse read nudge (`_READ_NUDGE`) fired unconditionally on every qualifying Read, with no dedup, no session cap, and no suppression while the agent had recently run `graphify query`/`explain`/`path`. Only the opt-in strict deny was capped, via `_mark_session_denied`.

Measured across 87 Claude Code transcripts in one repo: 8,348 nudge injections producing 339 graphify calls (~4% response rate), at a standing cost of ~651k tokens — about 3.7x every graphify query in those sessions combined.

## Fix

- Suppress the nudge entirely while `_query_stamp_fresh()` is true — an agent that just oriented (query/explain/path) does not need reminding on its very next read.
- Otherwise cap it to once per session per `GRAPHIFY_HOOK_NUDGE_TTL` seconds (default 300), via a new `_nudge_recently_shown` helper that mirrors the existing `_mark_session_denied` marker-file pattern (best-effort GC of markers older than 24h, fails open on any error so the nudge is never wrongly suppressed).

Both suggested fixes from the issue (#1 and #2, the two highest value-to-effort items) are implemented; dedup-per-target and payload shortening (#3/#4) are left out to keep this PR narrowly scoped to the two changes that account for "well over 90%" of the measured cost per the issue's own estimate.

## Tests

- Updated `test_fresh_query_stamp_suppresses_deny` (renamed `..._and_nudge`) — a fresh query stamp now suppresses both the strict deny *and* the soft nudge, since the old assertion documented the exact bug being fixed.
- Added `test_fresh_query_stamp_suppresses_soft_nudge_too` — same suppression applies outside strict mode.
- Added `test_soft_nudge_capped_per_session` — a second read in the same session within the TTL produces no output.
- Added `test_soft_nudge_cap_is_per_session` — a different session id still gets the nudge.
- Added `test_soft_nudge_cap_expires_after_ttl` — an expired marker (backdated via `os.utime`) lets the nudge fire again.
- Added `test_soft_nudge_never_capped_without_session_id` — fail-open: no session id means no cap is possible, so the nudge keeps firing rather than risk silently swallowing it.

`uv run pytest tests/test_hook_strict.py tests/test_hook_guard.py tests/test_hook_guard_token_match.py tests/test_hook_out_of_project_paths.py tests/test_install_strings.py -q` — 162 passed.

Full suite (`uv run pytest -q`): 5410 passed, 30 failed, 99 skipped. All 30 failures are pre-existing and unrelated to this change (Windows-platform-specific: FIFO/unix-socket tests that don't apply on Windows, backslash path-separator assertions, shebang parsing) — none touch `graphify/cli.py`'s hook-guard code or `tests/test_hook_strict.py`.

## Caveats

- `GRAPHIFY_HOOK_NUDGE_TTL` default of 300s (5 minutes) is a judgment call, not something I could tune against real usage data beyond the one repo's transcripts cited in the issue. Happy to adjust if a maintainer has a better sense of the right default.
- Scope is deliberately narrow to the read nudge (`_READ_NUDGE`) reported in #3435. The search nudge (`_SEARCH_NUDGE`) has the same unconditional-fire shape but wasn't part of the measured cost in the issue, so I left it untouched rather than expanding scope.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,7 @@ Full release notes with details on each version: [GitHub Releases](https://githu
 - Fix: `graphify install` no longer aborts when the always-on registration target is unwritable (read-only or symlinked config); it skips that step with an actionable warning and still installs the skill (#3474, thanks @dajiaohuang).
 - Fix: community labelling falls back to an installed `claude` CLI resolved at run time instead of pinning a path that expires (e.g. under snap/nvm), so labelling keeps working across updates (#3475, thanks @ktsang622).
 - Fix: the `all` extra now includes `psycopg[binary]`, so `pip install 'graphifyy[all]'` provides the postgres driver (#3482, thanks @L4XB).
+- Fix: the soft PreToolUse read nudge is now suppressed while a recent `graphify query`/`explain`/`path` stamp is fresh, and is capped to once per session per `GRAPHIFY_HOOK_NUDGE_TTL` seconds (default 300) — previously it re-fired unconditionally on every qualifying read, measured at 8,348 injections (~651k tokens) in one repo at a ~4% response rate (#3435, thanks @Ashfaqbs).
 
 ## 0.9.57 (2026-09-09)
 
--- a/graphify/cli.py
+++ b/graphify/cli.py
@@ -738,6 +738,43 @@ def _mark_session_denied(session_id: str) -> bool:
         return False
 
 
+def _nudge_recently_shown(session_id: str) -> bool:
+    """True if the soft read nudge already fired for this session within
+    GRAPHIFY_HOOK_NUDGE_TTL seconds (default 300); otherwise records one now and
+    returns False (#3435: the nudge previously had no per-session cap, so it re-fired
+    on every qualifying read — 8,348 injections / ~651k tokens measured in one repo
+    at a ~4% response rate). Fail-open: any error returns False so the nudge is never
+    wrongly suppressed. Best-effort GC of markers older than 24h."""
+    from graphify.paths import out_path, write_text_atomic
+    sid = re.sub(r"[^A-Za-z0-9_-]", "_", str(session_id))[:64]
+    if not sid:
+        return False
+    try:
+        ttl = float(os.environ.get("GRAPHIFY_HOOK_NUDGE_TTL", "300"))
+        d = out_path("cache", "hook_sessions")
+        d.mkdir(parents=True, exist_ok=True)
+        marker = d / f"{sid}.nudged"
+        try:
+            if (time.time() - marker.stat().st_mtime) < ttl:
+                return True
+        except OSError:
+            pass
+        write_text_atomic(marker, str(time.time()))
+        try:
+            cutoff = time.time() - 86400
+            for entry in os.scandir(d):
+                try:
+                    if entry.stat().st_mtime < cutoff:
+                        os.unlink(entry.path)
+                except OSError:
+                    pass
+        except OSError:
+            pass
+        return False
+    except Exception:
+        return False
+
+
 _SEARCH_COMMANDS = frozenset({
     "grep", "egrep", "fgrep", "zgrep", "rg", "ripgrep", "find", "fd", "ack", "ag",
 })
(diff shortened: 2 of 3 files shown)


Ashfaqbs:
Closing — #3443 already fixes this issue (and more completely, since it also bounds the search nudge, which I left out of scope here). Didn't check for an existing PR against #3435 before starting; apologies for the noise.

PR #3500, opened 2026-09-11
perf(resolution): memoize path resolution in the symbol-resolution passes (47% faster extract)

abhay-codes07:
## What

Memoize path resolution across the symbol-resolution and id-disambiguation passes. Two commits, one theme:

1. **`_source_key`** — `_disambiguate_colliding_node_ids` computes a source key once per node, per edge endpoint and per raw_call, each call running `Path.resolve()`. Tens of thousands of calls over a few hundred distinct `source_file` strings.
2. **`Path.resolve()` everywhere in the resolution passes** — the symbol-resolution facts/apply passes and the cross-file import pass call `.resolve()` once per import/export/use fact and per node (~44k calls over the same few hundred paths).

Both are wrapped in a per-`(path, cwd)` `lru_cache`. `resolve()` walks `nt._getfinalpathname` / `readlink` every time on an uncached call; the same corpus paths were re-resolved thousands of times within one run.

## Why it's safe

- Files do not move mid-extract, so a path resolves to the same target throughout a run.
- **cwd is part of the cache key**, so a relative path resolved after a `chdir` misses the cache instead of replaying a stale answer (verified by test).
- The helpers raise exactly like `Path.resolve()` — callers keep their own `try/except`, and an exception is never cached.
- The graph is **byte-identical** before and after: I froze the 364-file self-corpus, dumped the sorted node-id and `source|relation|target` edge sets on `v8` and on this branch, and they `cmp` equal.

## Measured

Sequential extract of graphify's own 364-file Python corpus (best of 3, warm import, fresh cache dir each run):

| | best |
|---|---|
| `v8` | 26.6s |
| this branch | **14.0s** |

A **47% reduction**, entirely in `_disambiguate_colliding_node_ids` (34% of the old run) and the resolution passes. No behavior change.

## Tests

`tests/test_source_key_memoization.py` and `tests/test_resolve_memoization.py`: equality with the pre-memo implementation across relative / absolute / out-of-root / missing / empty inputs, one resolution per distinct path, and cwd-sensitivity. Full suite matches a fresh `v8` (0.9.58) baseline.

### DIFF
--- a/graphify/extractors/resolution.py
+++ b/graphify/extractors/resolution.py
@@ -138,7 +138,7 @@ def _read_tsconfig_aliases(tsconfig: Path, base_dir: Path, seen: set) -> dict[st
         # Skip scoped npm package configs (e.g. @tsconfig/svelte) — not on disk.
         if not ext or ext.startswith("@"):
             continue
-        extended_path = (base_dir / ext).resolve()
+        extended_path = _resolve_cached((base_dir / ext))
         if not extended_path.suffix:
             extended_path = extended_path.with_suffix(".json")
         if extended_path.exists():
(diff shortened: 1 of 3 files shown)


safishamsi:
Shipped in **v0.9.59** — landed on `v8` via cherry-pick with your authorship preserved. Thanks @abhay-codes07 — path resolution is memoized across the resolution passes (~47% faster extract), with per-run cache clearing. Release: https://github.com/Graphify-Labs/graphify/releases/tag/v0.9.59

PR #3501, opened 2026-09-11
perf(resolution): walk the Python tree iteratively, not recursive yield-from

abhay-codes07:
## What

`_walk_python_tree` — the preorder tree walk that drives every Python symbol-resolution pass — was a recursive generator:

```python
def _walk_python_tree(node):
    yield node
    for child in node.children:
        yield from _walk_python_tree(child)
```

`yield from` builds one suspended generator frame per tree level and re-propagates every yielded node up the entire ancestor chain. On graphify's own 364-file corpus that is **~25M frame resumptions for ~2.8M actual nodes** (4.4s of a sequential extract, per cProfile). This PR rewrites it as an explicit stack — the identical rewrite `_walk_js_tree` right above it already carries, with the same rationale in its comment.

## Why it's safe

- Yields **the same node objects in the same preorder** — children pushed reversed so the first child pops first. A test asserts `[id(n) for n in iterative] == [id(n) for n in recursive]` on real parsed trees (a class with methods, comprehensions, async/with, lambdas), plus the empty module.
- As a bonus it no longer risks `RecursionError` on a deeply nested tree (test parses 2000-deep nesting and walks it fine).
- Graph is **byte-identical** before/after on the frozen self-corpus (sorted node + edge sets `cmp` equal).

## Measured

Isolated micro-benchmark walking all 86 parsed trees in `graphify/` (best of 5, same 594,438 nodes):

| | best |
|---|---|
| recursive (`v8`) | 135ms |
| iterative | **95ms** |

A **30% reduction** on the walk itself; it feeds several resolution passes.

## Tests

`tests/test_walk_python_tree_iterative.py`: node-identity/order equality with the recursive reference, visit-once, and the deep-nesting case. Full suite matches a fresh `v8` (0.9.58) baseline.

### DIFF
--- a/graphify/extractors/resolution.py
+++ b/graphify/extractors/resolution.py
@@ -1949,9 +1949,20 @@ def _parse_python_tree(path: Path):
         return None
 
 def _walk_python_tree(node):
-    yield node
-    for child in node.children:
-        yield from _walk_python_tree(child)
+    """Preorder walk of a tree-sitter tree, iteratively.
+
+    The recursive ``yield from`` form built one suspended generator frame per
+    ancestor and re-propagated every node up the whole chain — ~25M frame
+    resumptions on a 364-file corpus for ~2.8M actual nodes. An explicit stack
+    yields each node exactly once in the identical preorder (children pushed
+    reversed so the first child pops first). Same rewrite, same reasoning as
+    ``_walk_js_tree`` above.
+    """
+    stack = [node]
+    while stack:
+        current = stack.pop()
+        yield current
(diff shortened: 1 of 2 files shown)


safishamsi:
Shipped in **v0.9.59** — landed on `v8` via cherry-pick with your authorship preserved. Thanks @abhay-codes07 — the Python resolution tree walk is now iterative (part of the ~47% faster extract). Release: https://github.com/Graphify-Labs/graphify/releases/tag/v0.9.59

PR #3502, opened 2026-09-11
perf(resolution): parse each Python file once across both resolution passes

abhay-codes07:
## What

The Python symbol-resolution facts pass (`_collect_python_symbol_resolution_facts`) and the cross-file import pass (`_resolve_cross_file_imports`) each parse the **entire `.py` corpus**, back to back, in the main process after the workers return. Every file was read from disk and tree-sitter-parsed twice for no reason — cProfile showed `read_bytes` firing ~5× per file and `parse` ~4× across the run.

`_parse_python_tree` now memoizes on `(path, mtime_ns, size)`, and the cross-file pass — which previously built its own `Parser` and re-read each file — routes through it. The second pass reuses the first pass's tree.

## Why it's safe

- Both passes **only read** the tree (walk it, query it); a tree-sitter parse is immutable, so sharing one parse between them is behavior-preserving.
- The cache key includes **mtime and size**, so a file edited between watch runs re-parses instead of replaying a stale tree (verified by test).
- `_parse_python_tree` still returns `None` on any error exactly as before — callers already treat that as "skip this file".
- The cross-file pass returned the parse `Tree` and read `.root_node`; the shared helper returns `root_node` directly, and the one downstream `tree.root_node` access was updated accordingly.
- Graph is **byte-identical** before/after on the frozen 364-file self-corpus (sorted node + edge sets `cmp` equal).

## Measured

Sequential extract of graphify's own 364-file Python corpus (best of 3, fresh cache dir each run):

| | best |
|---|---|
| `v8` | 26.6s |
| this branch | **23.8s** |

A ~11% reduction from eliminating one full parse (and disk read) of every Python file. Composes with the other resolution-pass perf work.

## Tests

`tests/test_python_parse_memoization.py`: parse returns `(source, root)`, missing file → `None`, one parse reused within a run, and an edited file re-parses rather than replaying. Full suite matches a fresh `v8` (0.9.58) baseline.

### DIFF
--- a/graphify/extractors/resolution.py
+++ b/graphify/extractors/resolution.py
@@ -10,6 +10,7 @@ from graphify.extractors.base import (  # noqa: F401
     _make_id,
     _read_text,
 )
+import functools
 import hashlib
 import json
 import os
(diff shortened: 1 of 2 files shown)


safishamsi:
Shipped in **v0.9.59** — landed on `v8` via cherry-pick with your authorship preserved. Thanks @abhay-codes07 — each Python file now parses once across both resolution passes (part of the ~47% faster extract). Release: https://github.com/Graphify-Labs/graphify/releases/tag/v0.9.59

PR #3503, opened 2026-09-11
Fix node loss during incremental deduplication

### LINKED ISSUE #3477: build_merge silently drops nodes from files entirely outside the --update's scope (no ID collision, no warning) — shrink guard catches it, root cause undiagnosed
## Summary

An ordinary incremental `graphify --update` on a mature, actively-maintained graph produced a merge candidate missing 313 nodes belonging to 313 files that were **not part of this run's changeset at all** — not re-extracted, not pruned, not touched in any way. None of the 313 dropped IDs collide with any ID the new extraction produced. The `#479` shrink guard correctly caught the resulting net loss and refused to overwrite `graph.json`, so no data was lost on disk — but the mechanism producing the loss during the merge itself is undiagnosed, and this is not a one-off: the corpus was previously rebuilt from scratch specifically to clear a node-loss problem, and the same class of symptom has now recurred during normal incremental operation. A full rebuild is not a durable workaround for whatever this is.

## Environment

- `graphifyy` 0.9.56 (`~/.local/share/uv/tools/graphifyy/`)
- macOS 26.6.2, Python 3.12.8
- `graphify --update` via the standard skill path, `dedup=True` (default)
- Corpus: 673 files, ~3 months of incremental history, graph at 5,262 nodes prior to this run

## What happened

Run scope: 55 files changed since the last update (49 doc + 5 paper), 14 already cached, 40 requiring fresh semantic extraction. Split into 2 subagent chunks of 20 files each — both completed cleanly, no JSON validation errors, no dangling refs: 227 new nodes combined, 293 edges, 3 hyperedges per chunk. AST extraction on the 1 changed code file: 6 nodes, 13 edges, clean.

`build_merge()` on this input produced **4,951 nodes** — verified directly against the on-disk files:

```
graph.json (pre-update, on disk):        5262 nodes
.graphify_extract.json (merge candidate): 4951 nodes
```

A net loss of 311, which correctly tripped the `#479` shrink guard — `graph.json` was refused a write and remains at 5,262 nodes, confirmed untouched (mtime predates this run).

## Diagnosis so far

Diffed old-graph node IDs against the new merge candidate: **573 old nodes are absent** from the new output.

- **260** come from the 40 files that were legitimately re-extracted this run — expected, their old records are supposed to be replaced.
- **313 come from files nothing in this run touched.** Checked directly: zero overlap between these 313 IDs and the full set of IDs the new extraction produced (no collision, fuzzy or exact, that would explain a legitimate merge). They are properly path-namespaced per the project's ID convention — e.g. `commissions_comm_075_atlas_stage_a_comm_075`, `commissions_comm_089_b7_characterization_preregistered_branches_0_4b` — not the kind of generic/global-style ID (e.g. `person_emma`) that would plausibly collide across files.

Separately, the same merge logged 5 explicit `minted by two different files` warnings for genuine shared-entity collisions (`person_emma`, `person_vera`, `paper_shai_2024_belief_state_geometry`, two `syntheses_grammar_driven_model_*` nodes). Those are correct, expected consolidation, and have zero overlap with the 313 — they are not part of the unexplained loss.

The merge's own log reported `Deduplicated 294 node(s), 293 exact + 1 fuzzy` for this call. The count is suspiciously close to the scale of the unexplained loss, which is why dedup is the leading suspect — but I have not isolated which specific pass or key is responsible.

## Related issues checked, ruled out

- **#3257** (ghost-merge deletes non-AST node from unrelated file on `(basename, label)` collision) — fixed in 0.9.23 via full-path keying, confirmed dead as recently as Sep 2; doesn't apply at 0.9.56, and this run has zero ID/label collision with the dropped nodes regardless.
- **#3004** (`build_merge` replace-on-re-extract deletes a source's own nodes before validating the new extraction; `#479` guard skipped under `dedup=True`) — mechanistically adjacent (same guard, same `dedup=True` default) but the wrong scope: that issue is about a re-extracted file under-producing *on itself*. This report is about nodes vanishing from files **outside the changeset entirely**. Also worth flagging as a data point either way: #3004 states the `#479` guard is skipped whenever `dedup=True` (the default) — but the guard clearly fired here. Either that skip condition has narrowed since 0.9.48, or this loss pattern doesn't route through the same code path #3004 describes.
- **#2229** (`#479` guard misreads a legitimate dedup win as data loss via total-count accounting) — wrong direction; this is a real loss the guard correctly caught, not a false positive.
- Several closed basename/label collision dedup bugs (#2068, #2084, #1205, #895, #1284) — all fixed, and none apply since ID/label collision was explicitly checked and ruled out for the 313.

## Preserved for further diagnosis

Nothing was forced past the guard. On disk and available if useful:
- `.graphify_extract.json` — the full 4,951-node merge candidate
- Semantic cache for the 40 re-extracted files (so this can be replayed without re-running the subagent extraction)
- The full list of 313 dropped node IDs and the diff methodology, available on request

## One operational loose end from stopping mid-pipeline

`manifest.json` was already stamped for the 40 re-extracted files — that write happens before the guard fires, in `save_manifest`. A plain `graphify --update` run right now would likely treat those 40 files as already processed even though nothing from this run actually reached `graph.json`. Left as-is rather than hand-patched, since the manifest schema wasn't understood well enough to edit safely without risking a different kind of silent corruption.

## Impact

Silent node loss with no collision and no log trace, on a merge the shrink guard happens to catch by total count — but a smaller or differently-shaped loss on the same corpus might not cross that threshold and would go completely unnoticed. This is the second time this specific corpus has hit a node-loss problem serious enough to warrant a from-scratch rebuild; the recurrence during ordinary incremental use suggests the underlying cause survived that rebuild.

hopstreax:
## Summary

Fixes #3477, where `build_merge()` could silently drop nodes from files outside the scope of an incremental update.

The root cause was global entity deduplication collapsing duplicate nodes belonging exclusively to untouched files.

## Changes

- Add protected node IDs to incremental `build_merge()` deduplication.
- Prevent untouched nodes from being merged with other untouched nodes.
- Preserve untouched nodes as canonical survivors when incoming nodes deduplicate against them.
- Prevent incoming nodes from transitively bridging multiple untouched nodes.
- Apply the protection consistently across exact, fuzzy, and LLM-assisted deduplication.
- Preserve existing full-build deduplication behavior.
- Add regression tests for untouched duplicates, incoming/untouched merges, bridge cases, and normal incoming deduplication.

The existing shrink-guard and manifest behavior are intentionally unchanged and remain out of scope.

## Tests

- `237 passed, 1 skipped`
- `git diff --check` clean

### DIFF
--- a/graphify/build.py
+++ b/graphify/build.py
@@ -1426,6 +1426,7 @@ def build(
         combined["hyperedges"].extend(ext.get("hyperedges", []))
         combined["input_tokens"] += ext.get("input_tokens", 0)
         combined["output_tokens"] += ext.get("output_tokens", 0)
+    _root = str(Path(root).resolve()) if root else None
     if dedup and combined["nodes"]:
         # Numeric ids must be str before dedup, which keys on them and would
         # raise TypeError in _pick_winner's regex search (#2326). build_from_json
@@ -1438,14 +1439,23 @@ def build(
         for n in combined["nodes"]:
             if isinstance(n, dict):
                 _fold_node_aliases(n)
+                # Normalize source_file and definition_file to the build root before
+                # deduplication (#3472), so exact-ID collision checks and same-file
+                # attribute merging operate on canonical repo-relative paths rather
+                # than false-flagging absolute paths from semantic subagents as
+                # different files.
+                if "source_file" in n:
+                    n["source_file"] = _norm_source_file(n["source_file"], _root)
+                if "definition_file" in n:
+                    n["definition_file"] = _norm_source_file(n["definition_file"], _root)
         combined["nodes"], combined["edges"] = deduplicate_entities(
             combined["nodes"], combined["edges"], communities={},
-            dedup_llm_backend=dedup_llm_backend, root=root,
+            dedup_llm_backend=dedup_llm_backend, root=_root,
             # Hyperedge members reference node ids too, so they need the same
             # survivor rewiring the edges get (#2805).
             hyperedges=combined.get("hyperedges"),
         )
-    return build_from_json(combined, directed=directed, root=root)
+    return build_from_json(combined, directed=directed, root=_root)
 
 
 def _norm_label(label: str | None) -> str:
@@ -2018,7 +2028,7 @@ def build_merge(
     )
 
     all_chunks = base + list(new_chunks)
-    G = build(all_chunks, directed=directed, dedup=dedup, dedup_llm_backend=dedup_llm_backend, root=root)
+    G = build(all_chunks, directed=directed, dedup=dedup, dedup_llm_backend=dedup_llm_backend, root=_eff_root)
 
     # Prune nodes and edges from deleted source files
     if prune_sources:
--- a/graphify/build.py
+++ b/graphify/build.py
@@ -1402,6 +1402,7 @@ def build(
     dedup: bool = True,
     dedup_llm_backend: str | None = None,
     root: str | Path | None = None,
+    protected_ids: "set[str] | None" = None,
 ) -> nx.Graph:
     """Merge multiple extraction results into one graph.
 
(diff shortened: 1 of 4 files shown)


safishamsi:
Shipped in **v0.9.59** — landed on `v8` via cherry-pick with your authorship preserved. Thanks @hopstreax — incremental dedup no longer drops cross-file concept nodes from files it did not touch. Release: https://github.com/Graphify-Labs/graphify/releases/tag/v0.9.59

PR #3506, opened 2026-09-11
feat(omp): bridge hook-guard denials and context through native hooks

andrebrait:
## Summary
Bridge Graphify's existing hook-guard policy into OMP's native tool_call/context hooks, including strict denials. Ship the extension inside the Python distribution and register it through OMP's plugin manager with graphify omp install.

- Translate native read selectors and search targets, skip remote/internal resources, and retain hook-guard as the policy owner.
- Propagate permission denials before tool execution; append per-call guidance to each tool result (Claude PreToolUse additionalContext parity), clearing pending deliveries across runs and session navigation.
- Bound subprocess input/output/runtime, cancel stale generations, and resolve an installed executable outside the current project.
- Ship explicit OMP metadata in both wheel/sdist; canonical Graphify skills remain a separate existing installation, with no copied bodies or extraction/index changes.

## Related work
Related to #1937, which adds OMP nudges on tool results. This patch targets a different runtime contract: reuse hook-guard before execution so strict denials, native grep, path selectors, and policy context actually work. It does not duplicate hardcoded tool-result guidance. There is installation/documentation overlap; happy to coordinate the landing order.

## Verification
- Focused Python installer/guard/strict/path tests:115 pass,4 existing skips.
- Native Bun tests:7 pass,64 assertions, using the real Python CLI/path helpers plus bounded subprocess fixtures.
- Wheel and source distribution build; both ship native entry/manifest.
- Installed built wheel in an isolated environment; graphify omp install successfully linked its shipped package using the real OMP plugin manager in an isolated home.
- Actual OMP loader/ExtensionRunner/AgentSession: three user-run entry modes, strict selector-read refusal, subsequent guidance, and session reset pass against the installed wheel (scripted model response; disposable graph metadata fixture).

No existing project index was built or changed. OMP's isProjectTrusted() currently always returns true and is not a sandbox; docs say so. Tested host: OMP18.1.17 integration SHA6aef0e8ad51b3bc5ea7a5f2a255c3d48e4c5af72, not a promise about all published builds with that version. The installed executable policy excludes a project-local .venv; docs recommend uv tool/pipx.

AI-assisted implementation, independently reviewed for contracts, hostile input, test honesty and unnecessary complexity.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@ Full release notes with details on each version: [GitHub Releases](https://githu
 
 ## 0.9.58 (unreleased)
 
+- Feature: ship a native Oh My Pi guard package with `graphify omp install`; reuse the installed hook-guard policy for bounded tool-call denials and per-run context guidance without building indexes.
 - Fix: a call to a Python function defined nested inside another function now resolves to that inner definition per lexical scope, instead of leaking to a same-named function elsewhere; direct recursion is preserved as a self-loop (#3410, thanks @hopstreax).
 - Fix: submodule imports inside a PEP 420 namespace package (a directory with no `__init__.py`) now resolve to the target module instead of being dropped (#3429, thanks @flaukowski).
 - Fix: a bare-name import of a module sitting next to the importing file (a flat script dir with no package) now resolves to that sibling — matching CPython's `sys.path[0]` behavior — without over-resolving a genuine third-party name (#3430, thanks @hopstreax).
--- a/README.md
+++ b/README.md
@@ -230,6 +230,7 @@ for example `graphify claude install --project` or `graphify codex install --pro
 | Agent Skills (cross-framework) | `graphify install --platform agents` (alias `--platform skills`) |
 | Kiro IDE/CLI | `graphify kiro install` |
 | Pi coding agent | `graphify install --platform pi` |
+| Oh My Pi (native guard) | `graphify omp install` |
 | Cursor | `graphify cursor install` |
 | Devin CLI | `graphify devin install` |
 | Google Antigravity | `graphify antigravity install` |
@@ -303,6 +304,7 @@ Run this once in your project after building a graph:
 | Agent Skills (cross-framework) | `graphify agents install` (alias `graphify skills install`) |
 | Kiro IDE/CLI | `graphify kiro install` |
 | Pi coding agent | `graphify pi install` |
+| Oh My Pi (native guard) | `graphify omp install` |
 | Devin CLI | `graphify devin install` |
 | Google Antigravity | `graphify antigravity install` |
 
@@ -321,7 +323,53 @@ This writes a small config file that tells your assistant to consult the knowled
 
 **Cursor** writes `.cursor/rules/graphify.mdc` with `alwaysApply: true`, so Cursor includes it in every conversation automatically, no hook needed.
 
-To remove graphify from all platforms at once: `graphify uninstall` (add `--purge` to also delete `graphify-out/`). Or use the per-platform command (e.g. `graphify claude uninstall`).
+### Oh My Pi (OMP) native guard
+
+Install Graphify into a persistent environment with `uv tool install graphifyy` or
+`pipx install graphifyy`, and install OMP separately. Then run:
+
+```bash
+graphify omp install
+```
+
+This delegates to OMP's supported `omp plugin install <local-package-directory>`
+route. The Python wheel and source distribution contain `graphify/omp/package.json`
+with explicit `omp.extensions` and its TypeScript entry point; OMP links that
+directory and discovers the extension. No npm adapter package, shell hook, Pi API
+shim, or hand-edited OMP settings are required. `graphify omp path` prints the same
+directory for manual `omp plugin install` or one-session `omp -e` use. Restart OMP
+after installing/upgrading, and rerun the installer if the Python environment
+moves. Do not link from an ephemeral `uvx` environment.
+
+Before native `read`, `glob`, `grep`, and search-style `bash` calls, the extension
+runs the installed `graphify hook-guard read|search` CLI. The existing Python
+policy owns fresh/stale graph decisions and strict-mode denials: start OMP with
+`GRAPHIFY_HOOK_STRICT=1` to enable its once-per-session indexed-read block.
+Denials become OMP `block`/`reason`; guidance becomes one deduplicated context
+message, cleared for each new user run and session navigation. In-flight hooks
+are cancelled on these boundaries. No graph is created or updated automatically.
+The guard package intentionally declares no skills; the existing cross-framework
+skill remains available separately through `graphify agents install`.
+
+Only local filesystem targets are inspected, using OMP's selector/path helpers.
+URLs and internal resources are excluded. Hooks require an installed `graphify`
+on an absolute PATH entry outside the project (including outside a project-local
+virtualenv); there is no project Python or command fallback. Each native tool call
+has a 256 KiB JSON-input cap, a 64 KiB output cap and a two-second subprocess budget
+with forced termination. Missing commands, invalid output and failures fail open.
+
+The extension checks `ctx.isProjectTrusted()` before execution and context
+injection, but **current OMP exposes this compatibility method as always true**:
+it is not an enforced trust sandbox. Enable this integration only in projects you
+trust. The lifecycle contract is based on OMP integration commit
+`6aef0e8ad51b3bc5ea7a5f2a255c3d48e4c5af72`; the `18.1.17` version string alone
+does not establish those lifecycle fixes.
+
+Remove this host-managed link with `omp plugin uninstall graphify-omp` **before**
+uninstalling the Python package. `graphify uninstall` handles Graphify-managed
+platform files, not OMP's plugin registry.
+
+To remove Graphify-managed platform files at once: `graphify uninstall` (add `--purge` to also delete `graphify-out/`). Or use the per-platform command (e.g. `graphify claude uninstall`). OMP links are removed separately as described above.
 
 ---
 
@@ -747,6 +795,9 @@ graphify kiro install               # .kiro/skills/ + .kiro/steering/graphify.md
 graphify kiro uninstall
 graphify pi install                # skill file (Pi coding agent)
 graphify pi uninstall
+graphify omp install               # native guard package, registered by OMP
+graphify omp path                  # shipped package directory for manual linking
+omp plugin uninstall graphify-omp  # remove the host-managed link
 graphify devin install             # skill file + .windsurf/rules/graphify.md (Devin CLI)
 graphify devin uninstall
 graphify antigravity install       # .agents/rules + .agents/workflows (Google Antigravity)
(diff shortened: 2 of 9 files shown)


andrebrait:
Triage of the refreshed automated review on `37dd4ac`:

- **Project/PATH executable:** `installedCommand()` resolves the discovered executable with `realpathSync`, then rejects the resolved target when it is at or inside the real project root (`index.ts:25-34`). The regression at `tests/omp.test.ts:159-167` covers both a project-local PATH entry and an executable created in the project; an external symlink back into the project resolves to the same rejected target.
- **Read input:** OMP’s native read tool supplies `input.path`; the extension deliberately resolves that OMP field and emits Claude-hook-compatible `tool_input.file_path`. The selector/strict-denial regression at `tests/omp.test.ts:81-99` exercises the real CLI boundary.
- **Glob input:** OMP’s glob tool supplies its glob in `input.path`; the extension deliberately emits it as Claude-hook-compatible `tool_input.pattern`. `tests/omp.test.ts:101-113` compares that exact payload against the real CLI output.
- **Concurrency:** the controller/guidance state is per extension instance (per AgentSession), not module-global; after each subprocess await, Set membership/byte-limit/add are one synchronous JS segment with no interleaving point.
- **Installer PATH:** `graphify omp install` is an explicit setup command and delegates to the user-selected `omp` executable from PATH. It never imports or executes project Python as a fallback. This matches the documented installation boundary rather than the runtime hook’s stricter executable policy.

No code changes for those five. Dogfooding did find one real packaging issue: replaying the PR onto a newer Graphify release can leave `graphify/omp/package.json` stale. `37dd4ac` adds a pyproject/package version equality test; the downstream 0.9.61 integration package is now versioned 0.9.61 and `graphify omp install` registers it as such.

andrebrait:
Rebased onto the latest `v8` (`26b02b5`) — head is now `f38f25b`, no conflicts.

Since the last push, the OMP bridge tests (`tests/omp.test.ts`) are now wired into CI as an `omp-bridge` job: bun resolves the real `@oh-my-pi/pi-coding-agent` path-utils via a pinned devDependency in `graphify/omp/package.json`, and the policy assertions run against the real installed Python CLI (`GRAPHIFY_TEST_CLI`).

The upstream checks for this PR are sitting in `action_required`, so a maintainer will need to approve the workflow run. The identical workflow is already green on our fork, including the new job: https://github.com/pfBlockerNG/graphify/actions/runs/35141102644 — could someone approve the run here so the PR shows its own CI signal? Thanks!

andrebrait:
Pushed `50b4d73` — the guidance delivery changed from a deduplicated per-run context-transform message to **Claude `PreToolUse` additionalContext parity**: `tool_call` still runs the guard (strict deny path, budgets, trust gating, and reset wiring unchanged), but the nudge is now appended to that call's **persisted tool result**. Every qualifying call carries its own guidance inline, so it survives compaction like any tool output and no longer disappears from the session transcript. Multi-target calls (e.g. delimited globs) accumulate their targets' messages instead of dropping all but one — the old dedup Set silently lost that; `/node_modules` is now gitignored for the local bun-test symlink.

Change went through a condensed adversarial review round before push. Findings addressed: multi-target accumulation (fixed + regression test with a varying-output fixture), and a trust recheck at delivery time (restored the distrust watchdog the context handler used to provide). Declined, with rationale: restoring the cumulative `remainingOutput` budget — that budget existed for the shared per-run context message; per-call delivery gets the same 64 KiB cap per call with the identical overflow-fails-open behavior (the overflow fixture still passes). Known/bounded, no action: a pending entry for a call denied at the approval gate is dropped at the next reset.

Bun suite: 7 pass / 0 fail (64 assertions) against the real installed CLI. Fork CI: https://github.com/pfBlockerNG/graphify/actions — upstream runs still need a maintainer's workflow approval.

andrebrait:
Pushed a fix (2904921) for a load-bearing import break plus a CI hardening step.

**The break:** `graphify/omp/index.ts` imported `isReadableUrlPath` from `@oh-my-pi/pi-coding-agent/tools/path-utils`. Upstream OMP moved that export to `@oh-my-pi/pi-tui/tools/read`; by 18.2.6 (current latest) `path-utils.ts` only imports the symbol internally and no longer re-exports it. Any OMP install past that move throws `SyntaxError: Export named 'isReadableUrlPath' not found` while loading the extension module, before `api.on("tool_call", ...)` ever registers — the hook-guard bridge silently no-ops with no error surfaced to the user.

**Why CI didn't catch it:** the `omp-bridge` job runs `bun install --frozen-lockfile`, and the committed lockfile pinned `@oh-my-pi/pi-coding-agent@18.2.2`, where the export still lived at the old path. `tests/omp.test.ts` imports the bridge directly and does exercise this — it just never saw a current OMP.

**The fix:**
- Import `isReadableUrlPath` from `@oh-my-pi/pi-tui/tools/read` directly.
- Raise the `@oh-my-pi/pi-coding-agent` devDependency floor to `^18.2.6` and declare `@oh-my-pi/pi-tui@^18.2.6` as its own devDependency (the bridge now imports from it directly, so it shouldn't ride along as an undeclared transitive dep of pi-coding-agent).
- Added an advisory `omp-bridge` CI step that installs whatever OMP is currently published (ignoring the lockfile) and re-runs `tests/omp.test.ts` against it, `continue-on-error: true` so an unrelated upstream release can't block a merge, while still surfacing a future symbol move instead of shipping it silently again.

Verified: `bun test tests/omp.test.ts` fails with the missing-export `SyntaxError` at 18.2.6 before this change, and passes (7 pass, 64 assertions) after.

andrebrait:
Pushed a guard fix + bridge simplification on top of this branch (SHA `ea3e24054487d125596022c0c2c95830a07c7ee1`).

**Guard fix** (`fix(hooks)`): `_is_cwd_relative` answered "no root and no drive" to decide whether a `file_path`/`path` value from the PreToolUse read hook is cwd-anchored. A URL is rootless and driveless by that exact same test (`https://x`, `myscheme://x`, even a bare `www.host/path`), so it short-circuited straight to "in project" and the containment check never ran — producing a false nudge on any URL-shaped read target. `file://` is handled as the deliberate exception (OMP's own pipeline resolves it to a local path, not an external URL, so it must still nudge on an in-project file). The containment check, extension allow-list, and staleness/strict-deny logic are all untouched.

**Bridge simplification** (`refactor(omp)`): now that the guard defends itself, `isRemote` gets a `file://` carve-out (previously every `file://` read was silently dropped by the bridge before ever reaching the guard); `resolveReadPath` → `resolveReadPathAsync` (non-blocking probes on the `tool_call` hot path, matching `read.ts`'s own usage); and a redundant `{ splitter: parseSearchPath }` argument is dropped (already `expandDelimitedPathEntries`'s default).

Full test coverage (new `tests/test_hook_url_paths.py`, plus a bridge-level regression test in `tests/omp.test.ts`) and verification details are in the commit messages.

andrebrait:
Thanks for the review — going through all four findings below. Two were genuine bugs (both stemming from the same root cause), one CI hardening item is now addressed differently, and one is refuted with evidence. Appreciate the flag on all of them; the bot was right to mark them unverified and ask for human review.

## #2 and #4 — remote `file://` authorities bypassing the guard: **APPLIED**

Both findings were the same bug in the two layers that classify `file://` values: `_normalize_hook_path` in `graphify/cli.py` and `isRemote` in `graphify/omp/index.ts`. Both unconditionally treated *any* `file://` value as a local path, discarding the authority entirely. `file://evil.com/<in-project path>` therefore classified identically to `file:///<in-project path>` — a remote host aliasing a real local file just by sharing its path component.

The rule now applied, mirroring the host rather than inventing one: **a `file://` URL is local only when its authority is empty or `localhost`** (RFC 8089; Node's own `url.fileURLToPath` enforces exactly this, throwing `ERR_INVALID_FILE_URL_HOST` for any other host). Any other authority now falls through to the existing foreign-scheme handling on both layers, the same path an `https://` or any other remote scheme already took.

Before (pre-fix, reproduced against the installed guard):

```
file:///root/git/pfBlockerNG/src/.../pfb_unbound.py         -> NUDGE   (correct)
file://evil.com/root/git/pfBlockerNG/src/.../pfb_unbound.py -> NUDGE   (WRONG — remote host treated as an in-project file)
```

After (rebuilt guard, same probe plus the `localhost` row):

```
file:///root/git/pfBlockerNG/src/.../pfb_unbound.py         -> NUDGE      (unchanged — still local)
file://evil.com/root/git/pfBlockerNG/src/.../pfb_unbound.py -> silent     (fixed — remote host, not aliased)
file://localhost/root/git/pfBlockerNG/src/.../pfb_unbound.py -> NUDGE     (localhost is local per RFC 8089 / Node)
```

Both layers, plus RED→GREEN tests for both, are covered in `graphify/cli.py` (`_normalize_hook_path`), `graphify/omp/index.ts` (`isRemote`), `tests/test_hook_url_paths.py`, and `tests/omp.test.ts`.

## #1 — unpinned `@latest` execution in the CI advisory step: **APPLIED**

The step existed to catch upstream symbol moves without waiting for the frozen lockfile to notice (it's happened before: `isReadableUrlPath` moved from `pi-coding-agent` to `pi-tui`). But the mechanism for catching that was "install whatever OMP publishes today and execute the bridge suite against it" — unpinned, mutable third-party code running unconditionally on every CI invocation.

Replaced with a download-only check: fetch the package with install scripts disabled (`bun add --ignore-scripts`), then statically compare the set of named symbols `graphify/omp/index.ts` imports from `@oh-my-pi/pi-coding-agent/tools/path-utils` against the exports that file actually declares — as plain text, via regex, the same way `tests/omp.test.ts` already checks the *oldest* host-provided copy. The downloaded package is never imported or executed; only its source text is read. This still catches the exact upstream move that motivated the step (verified locally by simulating a renamed export against a real downloaded copy — the check fails as expected), while removing the arbitrary-code-execution surface. `continue-on-error: true` is unchanged, so an unrelated upstream release still can't block a merge.

## #3 — read events ignoring `file_path`: **SKIPPED** (refuted)

OMP's native read tool has no `file_path` field. From `packages/coding-agent/src/tools/read.ts:633-649` (all four schema variants the tool can be constructed with):

```ts
const readSchema = type({
	path: type("string").describe("Local path, internal URI (e.g. memory://), or URL. Inline selectors are supported."),
});

const readSchemaWithSkills = type({
	path: type("string").describe(
		"Local path, internal URI (e.g. memory://, skill://), or URL. Inline selectors are supported.",
	),
});

const readSchemaWithoutMemory = type({
	path: type("string").describe("Local path, internal URI, or URL. Inline selectors are supported."),
});

const readSchemaWithoutMemoryWithSkills = type({
	path: type("string").describe("Local path, internal URI (e.g. skill://), or URL. Inline selectors are supported."),
});
```

Every variant is `path`; `file_path` does not appear anywhere in that file. `event.input.path` is therefore always the correct field for a native OMP read event, and the `: "."` fallback in `index.ts` never fires for one. The likely source of the confusion: the bridge itself *emits* `file_path` — as the field name in the Claude-shaped payload it constructs for the hook-guard CLI (`if (toolName === "Read") input.file_path = resolved;`), which is the correct field name in *that* direction, not the field OMP's own read event carries.

---

No code change for #3; #1, #2, and #4 are applied on this branch (`feat/omp-native-integration`) and cherry-picked onto `integration`.

PR #3509, opened 2026-09-12
fix(paths): fall back on WinError 17 for os.replace, not just PermissionError (#3508)

### LINKED ISSUE #3508: Windows: os.replace fails with WinError 17 during Kiro install and AST cache writes
## Summary

On Windows, Graphify fails when performing atomic file replacement with `os.replace()`.

The error is:

```text
OSError: [WinError 17] The system cannot move the file to a different disk drive.
```

This happens even when the source temporary file and destination file are in the **same directory and on the same drive**.

I encountered this in two Graphify workflows:

1. Installing the Graphify skill for Kiro.
2. Running AST extraction with `graphify extract . --code-only`.

The underlying `os.replace()` behavior can also be reproduced independently of Graphify.

---

## Environment

* OS: Windows
* Shell: PowerShell
* Graphify version: graphify 0.9.58
* Graphify installation: installed with `uv`
* Platform integration: Kiro
* System Python:

  * Python 3.12.10
* Graphify uv environment:

  * Python 3.14.7

Graphify was installed for the project with:

```powershell
graphify install --project --platform kiro
```

---

## Problem 1: Kiro skill installation fails

Command:

```powershell
graphify install --project --platform kiro
```

The installation starts successfully:

```text
references -> .kiro\skills\graphify\references
```

but fails while replacing the temporary `SKILL.md` file:

```text
  ...
  File "...graphify\install.py", line 253, in _copy_skill_file
    os.replace(tmp_dst, skill_dst)

OSError: [WinError 17] 系统无法将文件移到不同的磁盘驱动器。:
'.kiro\\skills\\graphify\\SKILL.md.tmp'
->
'.kiro\\skills\\graphify\\SKILL.md'
```

Both files are in the same project directory.

---

## Problem 2: AST cache writes fail

Command:

```powershell
```

Example output:

```text
[graphify extract] scanning D:\KiroWorkSpace\demo_opensepc
[graphify extract] --code-only: skipping 39 non-code file(s)
[graphify extract] found 23 code, 0 docs, 0 papers, 0 images
[graphify extract] AST extraction on 23 code files...
```

Then AST workers fail when replacing temporary cache files:

```text
warning: worker failed for D:\KiroWorkSpace\demo_opensepc\lib\main.dart:

[WinError 17] 系统无法将文件移到不同的磁盘驱动器。:

'D:\\KiroWorkSpace\\demo_opensepc\\graphify-out\\cache\\ast\\v0.9.58-s2\\356bd78755ed7f82cd4be93ad20e54d855288714709e86ed2f9861115a13db77.3hhh4pp_.tmp'

->

'D:\\KiroWorkSpace\\demo_opensepc\\graphify-out\\cache\\ast\\v0.9.58-s2\\356bd78755ed7f82cd4be93ad20e54d855288714709e86ed2f9861115a13db77.json'
```

Eventually:

```text
[graphify extract] AST extraction failed: [WinError 17]
系统无法将文件移到不同的磁盘驱动器。
```

The `.tmp` file and `.json` destination are clearly in the same directory and on the same `D:` drive.

---

## Minimal reproduction outside Graphify

The same issue can be reproduced with plain Python.

### Python 3.12

```powershell
$dir = "C:\graphify-replace-test"

New-Item -ItemType Directory -Force $dir | Out-Null

python -c "import os,tempfile; d=r'$dir'; fd,tmp=tempfile.mkstemp(dir=d,suffix='.tmp'); os.write(fd,b'test'); os.close(fd); dst=os.path.join(d,'test.json'); print(tmp,'->',dst); os.replace(tmp,dst); print('SUCCESS')"
```

Result:

```text
C:\graphify-replace-test\tmp0zs5shgn.tmp
->
C:\graphify-replace-test\test.json

  File "<string>", line 1, in <module>

OSError: [WinError 17] 系统无法将文件移到不同的磁盘驱动器。:
'C:\\graphify-replace-test\\tmp0zs5shgn.tmp'
->
'C:\\graphify-replace-test\\test.json'
```

This is important because both paths are on the same `C:` drive and in the same directory.

---

## Reproduced with Graphify's Python environment

The same behavior also occurs with Graphify's uv-managed Python 3.14.7:

```powershell
& "$env:APPDATA\uv\tools\graphifyy\Scripts\python.exe" -c "import os,tempfile; d=r'C:\graphify-uv-test'; os.makedirs(d,exist_ok=True); fd,tmp=tempfile.mkstemp(dir=d,suffix='.tmp'); os.write(fd,b'test'); os.close(fd); dst=os.path.join(d,'test.json'); print(tmp,'->',dst); os.replace(tmp,dst); print('SUCCESS')"
```

Result:

```text
C:\graphify-uv-test\tmptme75wbc.tmp
->
C:\graphify-uv-test\test.json

OSError: [WinError 17] 系统无法将文件移到不同的磁盘驱动器。
```

So the issue is reproducible with both:

```text
Python 3.12.10
Python 3.14.7
```

---

## Other file operations work correctly

PowerShell rename works:

```powershell
$dir = "C:\native-rename-test"

New-Item -ItemType Directory -Force $dir | Out-Null

Set-Content "$dir\test.tmp" "hello"

Rename-Item "$dir\test.tmp" "test.json"
```

Result:

```text
Success
```

Python copy + delete also works:

```powershell
$dir = "C:\python-copy-test"

New-Item -ItemType Directory -Force $dir | Out-Null

python -c "import os,shutil; d=r'$dir'; src=os.path.join(d,'test.tmp'); dst=os.path.join(d,'test.json'); open(src,'w').write('hello'); shutil.copy2(src,dst); os.unlink(src); print('SUCCESS')"
```

Result:

```text
SUCCESS
```

So currently:

```text
PowerShell Rename-Item       ✅
Python shutil.copy2          ✅
Python os.unlink             ✅
Python copy2 + unlink        ✅

Python os.replace            ❌ WinError 17
Graphify atomic replacement  ❌ WinError 17
```

---

## Expected behavior

Graphify should successfully write the destination file, or gracefully fall back to a copy/delete strategy when atomic replacement fails on Windows.

For example:

```python
try:
    os.replace(tmp_path, destination)
except OSError as exc:
    if getattr(exc, "winerror", None) in (5, 17, 32):
        shutil.copy2(tmp_path, destination)
        os.unlink(tmp_path)
    else:
        raise
```

Graphify already appears to have Windows fallback logic for s

ayushcodes10:
## Summary

Fixes #3508. On some Windows/filesystem combinations, `os.replace(src, dst)` raises `WinError 17` ("cannot move to a different disk drive") even when `src` and `dst` are in the same directory on the same drive — reproduced by the reporter both standalone and against graphify's own uv-managed Python, hitting two concrete graphify code paths: installing the Kiro skill (`SKILL.md`), and AST cache entry writes during `graphify extract`.

`WinError 17` maps to a plain `OSError` in Python, not `PermissionError`, so the existing Windows fallback (added for `WinError 5/32`, a destination transiently locked by antivirus or an open reader) never caught it — it only matched on exception type.

Adds a shared `os_replace_with_fallback(src, dst)` in `graphify/paths.py` that falls back to `shutil.copy2` + `os.unlink` for both `PermissionError` and `winerror == 17`, and re-raises anything else unchanged. Applied it to every direct `os.replace` call for a **file** across the codebase: the shared atomic writer (`paths.py`, which already had a narrower inline version of this), both cache write paths (`cache.py`), the GraphML exporter (`export.py`), and the skill/version-stamp installers (`install.py`). Left the one directory-level `os.replace` (`install.py`'s `references/` sidecar swap) untouched — `shutil.copy2` doesn't work on directories, and it wasn't part of the report; happy to add a `copytree`+`rmtree` fallback there too if it turns out to hit the same issue.

## Test plan

- [x] Unit tests for `os_replace_with_fallback` itself (simulated `WinError 17`, a bare `PermissionError` with no `winerror` set matching the existing WinError-5 test's mock shape, and an unrelated `OSError` that must still re-raise).
- [x] Regression test for the AST cache write path (`tests/test_cache.py`), matching the exact scenario from the report.
- [x] Regression test for the skill install path (`tests/test_install.py`), using a monolith platform (`aider`) so the test isolates the file replace under test from the separate, untouched directory replace.
- [x] Regression test for the shared atomic writer (`tests/test_atomic_writes.py`), alongside the existing WinError-5 test.

### DIFF
--- a/graphify/cache.py
+++ b/graphify/cache.py
@@ -17,6 +17,7 @@ from pathlib import Path
 # absolute path ("/shared/graphify-out"). Single source of truth in graphify.paths
 # (#1423); re-exported here as _GRAPHIFY_OUT for the existing call sites.
 from graphify.paths import GRAPHIFY_OUT as _GRAPHIFY_OUT
+from graphify.paths import os_replace_with_fallback as _os_replace_with_fallback
 
 # AST cache entries are the output of graphify's own extractor code, so they
 # are only valid for the version that wrote them: keying purely on file
@@ -399,7 +400,7 @@ def _flush_stat_index() -> None:
         try:
             os.write(fd, json.dumps(on_disk, separators=(",", ":")).encode())
             os.close(fd)
-            os.replace(tmp, p)
+            _os_replace_with_fallback(tmp, p)
         except Exception:
             try:
                 os.close(fd)
@@ -1126,14 +1127,7 @@ def save_cached(path: Path, result: dict, root: Path = Path("."), kind: str = "a
     try:
         os.write(fd, json.dumps(on_disk).encode())
         os.close(fd)
-        try:
-            os.replace(tmp_path, entry)
-        except PermissionError:
-            # Windows: os.replace can fail with WinError 5 if the target is
-            # briefly locked. Fall back to copy-then-delete.
-            import shutil
-            shutil.copy2(tmp_path, entry)
-            os.unlink(tmp_path)
+        _os_replace_with_fallback(tmp_path, entry)
     except Exception:
         try:
             os.close(fd)
--- a/graphify/export.py
+++ b/graphify/export.py
@@ -16,7 +16,12 @@ from networkx.readwrite import json_graph
 from graphify.security import sanitize_label
 from graphify.analyze import _node_community_map
 from graphify.build import edge_data
-from graphify.paths import stem_filename_budget, write_json_atomic, write_text_atomic
+from graphify.paths import (
+    os_replace_with_fallback,
+    stem_filename_budget,
+    write_json_atomic,
+    write_text_atomic,
+)
 
 from graphify.exporters.graphdb import push_to_falkordb, push_to_neo4j  # noqa: E402,F401
 
@@ -1270,7 +1275,7 @@ def to_graphml(
     tmp = out.with_name(out.name + ".tmp")
     try:
         nx.write_graphml(H, str(tmp))
-        os.replace(str(tmp), str(out))
+        os_replace_with_fallback(str(tmp), str(out))
     finally:
         if tmp.exists():
             try:
--- a/graphify/install.py
+++ b/graphify/install.py
@@ -30,6 +30,7 @@ except Exception:
     __version__ = "unknown"
 
 from graphify.paths import GRAPHIFY_OUT as _GRAPHIFY_OUT
+from graphify.paths import os_replace_with_fallback as _os_replace_with_fallback
 
 
 def _write_version_stamp(skill_dst: Path, version: str) -> None:
@@ -45,7 +46,7 @@ def _write_version_stamp(skill_dst: Path, version: str) -> None:
     tmp = version_file.with_name(".graphify_version.tmp")
     try:
         tmp.write_text(version, encoding="utf-8")
-        os.replace(tmp, version_file)
+        _os_replace_with_fallback(tmp, version_file)
     except Exception:
         try:
             tmp.unlink(missing_ok=True)
@@ -250,7 +251,7 @@ def _copy_skill_file(platform_name: str, *, project: bool = False, project_dir: 
     tmp_dst = skill_dst.with_suffix(skill_dst.suffix + ".tmp")
     try:
         shutil.copy(skill_src, tmp_dst)
-        os.replace(tmp_dst, skill_dst)
+        _os_replace_with_fallback(tmp_dst, skill_dst)
     except Exception:
         try:
             tmp_dst.unlink(missing_ok=True)
@@ -898,7 +899,7 @@ def vscode_install(project_dir: Path | None = None) -> None:
     tmp_dst = skill_dst.with_suffix(skill_dst.suffix + ".tmp")
     try:
         shutil.copy(skill_src, tmp_dst)
-        os.replace(tmp_dst, skill_dst)
+        _os_replace_with_fallback(tmp_dst, skill_dst)
     except Exception:
         try:
             tmp_dst.unlink(missing_ok=True)
--- a/graphify/paths.py
+++ b/graphify/paths.py
@@ -26,6 +26,27 @@ from pathlib import Path, PurePosixPath, PureWindowsPath
 GRAPHIFY_OUT = os.environ.get("GRAPHIFY_OUT", "graphify-out")
 
 
+def os_replace_with_fallback(src: "str | Path", dst: "str | Path") -> None:
+    """``os.replace(src, dst)``, falling back to copy-then-delete for a known
+    set of Windows quirks (#3508) that raise even when ``src``/``dst`` are the
+    same directory on the same drive: ``PermissionError`` (WinError 5/32 --
+    destination briefly locked by another handle, antivirus, an open reader)
+    and WinError 17 ("cannot move to a different disk drive", observed on some
+    Windows/filesystem combinations despite textbook same-volume semantics).
+    WinError 17 maps to a plain ``OSError`` in Python, not ``PermissionError``,
+    so it's checked via ``winerror`` rather than the exception type. Any other
+    failure is a real one and is re-raised.
+    """
+    try:
+        os.replace(src, dst)
+    except OSError as exc:
+        if not isinstance(exc, PermissionError) and getattr(exc, "winerror", None) != 17:
+            raise
+        import shutil
+        shutil.copy2(src, dst)
+        os.unlink(src)
+
+
 def _atomic_replace(path: "str | Path", write_fn) -> None:
     """Atomically replace ``path`` with content written by ``write_fn(f)``.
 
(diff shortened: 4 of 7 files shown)


ayushcodes10:
Thanks for the review. Went through all four findings.

**"Fallback does not preserve os.replace semantics for symlink destinations" and "Fallback replace follows destination symlinks" — both real, confirmed and fixed.** Reproduced directly: `os.replace` on a symlinked destination replaces the link itself (the link's old target is left untouched), while a naive `shutil.copy2(src, dst)` opens `dst` for writing and follows the link, overwriting whatever it points at instead. `install.py`'s own docstring for the version-stamp write explicitly documents relying on the replace-not-follow behavior for managed skill symlinks (#3286), so this would have been a real regression the first time the fallback fired on a symlinked install. Fixed by copying to a fresh temp file beside the destination first, then swapping the destination out from under it via unlink + rename — this replicates `os.replace`'s "whatever was there is gone" semantics regardless of whether a symlink or a regular file was there. Added a regression test that reproduces the exact scenario (a symlinked destination pointing at a separate shared file) and asserts the shared target is untouched.

**"Windows fallback exposes non-atomic partial writes to readers" — this was already true of the pre-existing WinError 5/32 fallback (in `paths.py` and `cache.py`), and my PR extended it to more call sites rather than introducing it.** That said, the redesign above happens to improve this too: copying to a temp file first and only swapping it into place via unlink+rename means the window a reader could observe a half-written destination shrinks from "however long the copy takes" to two back-to-back metadata operations, for every covered call site now, not just the ones that already had a fallback.

**"copy2 fallback fails on directory replace (WinError 17 for dirs)" — not reachable today.** `os_replace_with_fallback` is only ever called with individual files across this PR's 7 touched call sites (SKILL.md, version stamp, cache entries, stat index, GraphML) — the one directory-level `os.replace` in `install.py` (the `references/` sidecar swap) was deliberately left on the raw call, exactly because `shutil.copy2`/this fallback don't handle directories. Not currently a live bug, but a fair API-robustness point if that helper is ever reused for a directory swap later; happy to add an explicit `assert not os.path.isdir(src)` guard if you'd like that documented in code rather than just in this reply.

ayushcodes10:
Thanks for the second pass.

**"tempfile module used but not imported in paths.py fallback" — false positive.** `tempfile` is imported at module level (`graphify/paths.py:23`, part of the pre-existing imports this file already had) and used at `tempfile.mkstemp` a few lines above the flagged line; nothing to fix here.

**The other three ("Fallback replace can delete the original file on rename failure", "os_replace_with_fallback deletes source on copy failure via unlink(src)", "Fallback can delete the existing destination if the final rename fails", "Fallback replace exposes a missing-destination window") were all the same real gap, and a fair one to flag at high severity.** The previous version removed whatever sat at the destination and then renamed the new content into place as two separate steps; if the second step failed for any reason, the destination was left missing entirely — worse than the write it was trying to work around. Reproduced directly (forced the final rename to fail) and confirmed the destination ended up gone.

Fixed in the latest commit: instead of unlinking the destination outright, it's renamed aside to a backup path first, and if the final rename-into-place then fails, the backup is renamed straight back before the exception propagates. A plain rename moves a symlink itself rather than following it, so this also keeps the symlink-safety property from the previous round intact. Added a test that forces the final rename to fail and confirms the destination is restored to its original content with no leftover temp files.

(The `unlink(src)` at the very end only runs after the swap has already landed successfully, so a copy or swap failure never reaches it — `src` and `dst` are both left in a defined state either way.)

ayushcodes10:
Third pass, went through all five.

**"Fallback deletes destination when src and dst are the same path" — real, confirmed, fixed.** Reproduced directly: `os_replace_with_fallback(p, p)` crashed with `FileNotFoundError` on the final `unlink(src)`, because the "back dst up by renaming it aside" step moves `src` out from under itself the moment they're the same path. Not reachable through any of this PR's actual call sites (every one passes a distinct freshly-created temp file as `src`), but cheap to guard against directly, so I added an early same-path check that no-ops instead of attempting a swap. Test added.

**"Symlink dst is replaced with a copy of the symlink target's content, not a plain file" and "...semantics change for symlinked src" — not reachable, and I don't think worth hardening speculatively.** Both describe the same case: `shutil.copy2` dereferences a symlinked *source*, so if `src` itself were ever a symlink, the fallback would land a regular file with the target's content at `dst` instead of preserving `src` as a symlink the way a true `os.replace` would. Checked all 7 call sites this PR touches (`paths.py`, `cache.py` x2, `export.py`, `install.py` x3) — every one passes a plain temp file it just created as `src`, never a symlink. Happy to add `os.readlink` + `os.symlink` handling for a symlinked source if this helper ever grows a caller that needs it, but adding untested complexity for a case nothing currently exercises seems like the wrong trade here.

**"Fallback replace exposes a missing destination window" and "...fallback is non-atomic: dst missing between backup rename and copy rename" — real, and inherent to any fallback for this bug, not something further fixable.** There genuinely is a brief window between the backup rename and the landing rename where `dst` doesn't exist — a concurrent reader hitting that exact instant would see it missing. This can't be closed without an atomic swap primitive, which is precisely what's broken on the affected systems (that's why `os.replace` itself failed and got us here in the first place). What the last round's fix did address is the *failure* case: if the landing rename doesn't succeed, the backup goes straight back rather than leaving `dst` permanently missing. The window on the success path is now two back-to-back rename syscalls rather than however long the copy takes, which is the best available without a working atomic operation — already documented in the docstring, but happy to call it out more explicitly there if useful.

ayushcodes10:
Same finding as the previous round ("missing destination window" / "non-atomic fallback swap") — see my reply above. Still the inherent trade-off: no atomic swap primitive is available on the systems this fallback exists for, since `os.replace` itself is what's failing there. Nothing new to add.

safishamsi:
Shipped in **v0.9.60** — landed on `v8` via cherry-pick with your authorship preserved. Thanks @ayushcodes10 — atomic writes now fall back on Windows WinError 17, not just PermissionError. Release: https://github.com/Graphify-Labs/graphify/releases/tag/v0.9.60

Issue #1692, opened 2026-07-06
`Dockerfile`/`Gemfile`/`Makefile`/`Rakefile`/`LICENSE` and other extensionless, non-shebang project files currently produce no visible trace in graphify's output — not classified, not counted anywhere


sub4biz:
## Summary

The main thing this issue is about: right now, when `classify_file()` returns `None` for a file, that
file leaves no trace anywhere in graphify's own reporting — not in `total_files`, not as a skipped count,
nothing. This happens for any extensionless file without a shebang line — in practice, mostly declarative
project files like `Dockerfile`, `Gemfile`, `Makefile`, `Rakefile`, and `LICENSE`. A user has no way to
discover, from graphify's own output, that these files were ever considered.

Separately, and more as a "while we're looking at this" thought than a core ask: a couple of these
filenames look like they'd fit naturally into flows the tool already has, which is covered further down.

## Evidence

Real counts from one monorepo's `work/` tree (`find ... -type f ! -name "*.*"`, excluding `.git/` and
`graphify-out/`), grouped by basename:

```
26 Dockerfile
21 Gemfile
15 Rakefile
15 Makefile
 6 LICENSE
 6 rake / 6 rails / 6 bundle / 6 spring / 5 pre-commit / 4 update / 4 console / 3 yarn / 3 Capfile /
 3 Berksfile / 2 start / 2 sbt / 2 gradlew / ... (various project scripts and git hooks)
```

None of these carry a shebang line (`Dockerfile`/`Gemfile`/`Makefile`/`Rakefile`/`LICENSE` are
declarative/data files, not executables) — some of the scripts further down the list (`rake`, `console`,
`pre-commit`) may have shebangs and would be covered by #1680 once merged, but the top four alone
account for 83 files in this one corpus.

## Where this happens

`detect.py:386-420` (`classify_file`):

```python
def classify_file(path: Path) -> FileType | None:
    from graphify.manifest_ingest import is_package_manifest_path
    if is_package_manifest_path(path):
        return FileType.CODE
    if path.name.lower().endswith(".blade.php"):
        return FileType.CODE
    ext = path.suffix.lower()
    if not ext:
        return _shebang_file_type(path)     # the one path currently in place for extensionless files
    if ext in CODE_EXTENSIONS:
        return FileType.CODE
    ...
    return None
```

`_shebang_file_type(path)` (`detect.py:~370-383`) reads the first line and returns `FileType.CODE` only
[...]

PR #3512, opened 2026-09-12
Surface unclassified files instead of a silent successful run (#3511)

### LINKED ISSUE #3511: detect() already tracks unclassified files; nothing surfaces them, so a 260-file Lean 4 repo reports as a successful run
## Reproduce (public repo, 3 commands, no API key)

```bash
git clone --depth 1 https://github.com/leanprover-community/batteries.git
cd batteries
graphify update .
```

Result on `batteries` @ `3b4ce082`, graphify 0.9.58:

```
[graphify watch] Rebuilt: 18 nodes, 13 edges, 7 communities
Code graph updated.
```

Exit code 0. The 18 nodes come from 4 shell scripts and 2 README files. **All
260 `.lean` files are absent**, and the run says nothing about them.

`GRAPH_REPORT.md` then states:

```
## Corpus Check
- 11 files · ~7,795 words
- Verdict: corpus is large enough that graph structure adds value.
```

For a repository whose actual content is 260 Lean files, that verdict is
describing four shell scripts.

## The data is already there

This is the part that makes it cheap to fix. `detect()` already computes and
returns the list:

```python
>>> from graphify.detect import detect
>>> r = detect(Path("batteries"))
>>> sorted(r.keys())
['files', 'graphifyignore_patterns', 'ignored', 'needs_graph',
 'pruned_noise_dirs', 'scan_root', 'skipped_sensitive', 'total_files',
 'total_words', 'unclassified', 'walk_errors']

>>> r["total_files"]
22
>>> len(r["unclassified"])
268
>>> collections.Counter(Path(x).suffix for x in r["unclassified"]).most_common(3)
[('.lean', 260), ('.toml', 3), ('', 4)]
```

`unclassified` holds 268 entries, 260 of them `.lean`. Nothing downstream reads
it: it does not reach `GRAPH_REPORT.md`, and `graph.json` has no field for it.

So this is not a missing computation. It is a missing three lines of display.

## Suggested fix (not Lean-specific)

Surfacing `unclassified` would make the gap visible for every unsupported
language at once:

- scan summary line — `unclassified: 268 files (top: .lean 260, .toml 3)`
- `GRAPH_REPORT.md` Corpus Check — same line, next to the file/word counts
- `graph.json` — `"unclassified_files": {"count": 268, "by_extension": {".lean": 260, ...}}`

No change to graph topology, no new dependency, and it turns a silent drop into
a number the user can act on. A `Verdict:` line that knows 268 files were
skipped can also stop asserting the corpus is well covered.

## Why `.lean` is unclassified (for reference)

Verified in source on `v8` @ `23f2ffa`, not inferred from behaviour:

- `graphify/detect.py` → `CODE_EXTENSIONS` holds 100+ suffixes; `.lean` is not
  among them.
- `graphify/detect.py` → `DOC_EXTENSIONS` is
  `{.md .mdx .qmd .skill .txt .rst .html .yaml .yml}`; `.lean` is not there either.
- `pyproject.toml` declares 31 `tree-sitter-*` grammars. There is no
  `tree-sitter-lean` to add — `pypi.org/pypi/tree-sitter-lean/json` and
  `pypi.org/pypi/tree-sitter-lean4/json` both return HTTP 404.

Not classifying `.lean` is correct behaviour for an unsupported language. The
reporting is the bug.

## Separate question: should graphify parse Lean 4?

Raising it here rather than as a PR, because it does not fit the extractor
pattern in `ARCHITECTURE.md` and the architecture call is yours. Step 4 of
"Adding a new language extractor" is "Add the tree-sitter package to
`pyproject.toml` dependencies"; for Lean there is no such package, so any Lean
support would be the project's first non-tree-sitter code extractor — a
precedent rather than a patch.

I wrote a standalone Lean 4 extractor as prior art. It emits graphify's
documented extraction schema and `build_from_json()` consumes it unchanged. It
is deliberately two-layered so the trustworthy half is separable:

| layer | method | confidence |
|---|---|---|
| module / import / declaration | literal match on source **after** comments and string literals are blanked (nested `/- -/`, `--`, `"…"`) | exact |
| declaration → declaration `uses` | identifier match against the declared-symbol table, resolved through the enclosing `namespace` and `open`, local binders excluded | `INFERRED`, or `AMBIGUOUS` when more than one declaration matches |

On a 445-file Lean corpus it produced 4,068 nodes and 19,804 edges in about two
seconds. One number that may be worth knowing if you take this on: excluding
locally bound names (`∀`/`fun`/`intro`/`obtain`, and header binder groups) from
resolution removed **4,589 false `uses` edges — 19% of the edge set** — and
moved the hub list from single binder letters to the corpus's real concepts.
Before that fix the top "god nodes" were `j`, `I`, `A`, `S`.

Happy to open a PR for the reporting fix above if you want it; the extractor I
would only submit if a regex-layer extractor is a direction you want at all.

## Environment

- graphify 0.9.58, installed via `uv tool install 'graphifyy[leiden,svg,pdf,office,watch]'`
- Python 3.10.12, Linux aarch64
- repo `v8` @ `23f2ffa`; reproduction repo `batteries` @ `3b4ce082`

ayushcodes10:
## Summary

Fixes #3511. `detect()` already computes and returns `unclassified` — every file it saw but could not classify (no supported extension or shebang) — but nothing surfaced it. On a corpus in a language graphify has no extractor for (the report's example: a 260-file Lean 4 repo), `graphify update .` exits 0, prints a normal-looking "Rebuilt: N nodes..." summary, and `GRAPH_REPORT.md`'s Corpus Check says "corpus is large enough that graph structure adds value" — describing the 4 shell scripts and 2 READMEs that happened to be classifiable, with the other 260 files invisible and unmentioned.

`graphify extract` already had this reporting, added by #1692 — the gap is specifically the `update`/watch rebuild path (`graphify/watch.py`), which never got it, plus `GRAPH_REPORT.md`'s Corpus Check section, which had no field for it at all regardless of which command wrote the graph.

- `graphify/report.py`: adds an `Unclassified: N file(s) not represented in the graph (top: .ext A, .ext B)` line to the Corpus Check section whenever `detection_result` carries any.
- `graphify/watch.py`: `_rebuild_code` now prints the same "N file(s) not classified..." line `extract` already prints, and threads the count into the local detection summary it builds for `GRAPH_REPORT.md` (which previously only carried `total_files`/`total_words`, dropping `unclassified` even though `detect()` had already computed it).

Scoped to just the reporting gap, matching the issue's own "Suggested fix" section. Left `graph.json`'s `unclassified_files` field (the issue's third suggested surface) out of this PR — threading it through would mean passing `detection_result` down through `build`/`build_from_json`/`to_json`'s call chain, which is a much larger, more invasive change than the two display-only additions here; happy to follow up separately if wanted. Also left the issue's separate "should graphify parse Lean 4" question alone — the reporter raised it explicitly as a discussion point for a maintainer architecture call, not as part of this fix.

## Test plan

- [x] `tests/test_report.py`: new tests confirming the Corpus Check section includes the unclassified line (with per-extension breakdown) when present, and omits it entirely when absent (backward compatible).
- [x] `tests/test_watch.py`: new test confirming `_rebuild_code` prints the unclassified count and filenames for a mixed `.py`/`.lean` corpus.
- [x] Manually reproduced the issue's shape end to end (a `.py` file plus several `.lean` files, `_rebuild_code` called directly): confirmed both the console line and the `GRAPH_REPORT.md` Corpus Check line appear correctly.

### DIFF
--- a/graphify/report.py
+++ b/graphify/report.py
@@ -137,6 +137,21 @@ def generate(
             f"- {detection_result['total_files']} files · ~{detection_result['total_words']:,} words",
             "- Verdict: corpus is large enough that graph structure adds value.",
         ]
+        # #3511: files detect() saw but could not classify (no supported
+        # extension/shebang) were counted nowhere -- a corpus that is mostly
+        # an unsupported language reported the same "well covered" verdict as
+        # one that was actually extracted. Surface the count and its biggest
+        # extensions so a near-total miss (e.g. a Lean/Zig/whatever repo with
+        # no matching extractor) is visible here instead of silent.
+        unclassified = detection_result.get("unclassified") or []
+        if unclassified:
+            from collections import Counter as _Counter
+            ext_counts = _Counter(Path(p).suffix or "(none)" for p in unclassified)
+            top = ", ".join(f"{ext} {n}" for ext, n in ext_counts.most_common(3))
+            lines.append(
+                f"- Unclassified: {len(unclassified)} file(s) not represented in "
+                f"the graph (top: {top})"
+            )
 
     from .analyze import _is_file_node as _ifn
 
--- a/graphify/watch.py
+++ b/graphify/watch.py
@@ -1428,6 +1428,20 @@ def _rebuild_code(
         )
         code_files = [Path(f) for f in detected['files']['code']]
 
+        # #3511: `graphify extract` has surfaced files it saw but could not
+        # classify since #1692; this update/watch rebuild path never did,
+        # so a corpus in a language with no extractor (no supported
+        # extension or shebang) rebuilt "successfully" with those files
+        # silently absent from the graph. Same wording as the extract path.
+        _unclassified = detected.get("unclassified", []) if isinstance(detected, dict) else []
+        if _unclassified:
+            _names = ", ".join(sorted({Path(p).name for p in _unclassified})[:6])
+            _more = f" (+{len(_unclassified) - 6} more)" if len(_unclassified) > 6 else ""
+            print(
+                f"[graphify watch] {len(_unclassified)} file(s) not classified "
+                f"(no supported extension or shebang), skipped: {_names}{_more}"
+            )
+
         # #2495: hand reconcile the same ignore decisions the detect() call
         # above made, so a newly-ignored file that still exists on disk is
         # purged from the graph instead of preserved forever by the fail-closed
@@ -1876,6 +1890,7 @@ def _rebuild_code(
             "files": {"code": [str(f) for f in code_files], "document": [], "paper": [], "image": []},
             "total_files": len(code_files),
             "total_words": detected.get("total_words", 0),
+            "unclassified": detected.get("unclassified", []),
         }
 
         # Inherit the existing graph's directed flag (#2342) so `graphify
--- a/tests/test_report.py
+++ b/tests/test_report.py
@@ -29,6 +29,29 @@ def test_report_contains_corpus_check():
     report = generate(G, communities, cohesion, labels, gods, surprises, detection, tokens, "./project")
     assert "## Corpus Check" in report
 
+
+def test_report_surfaces_unclassified_files():
+    """#3511: detect() already tracks files it saw but could not classify
+    (no supported extension), but nothing surfaced them -- a corpus that is
+    mostly an unsupported language got the same "well covered" verdict as
+    one that was actually extracted."""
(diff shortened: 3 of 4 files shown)


safishamsi:
Shipped in **v0.9.60** — landed on `v8` via cherry-pick with your authorship preserved. Thanks @ayushcodes10 — files that couldn't be classified are surfaced instead of vanishing from a successful run. Release: https://github.com/Graphify-Labs/graphify/releases/tag/v0.9.60

PR #3516, opened 2026-09-12
detect: exempt Office/Workspace sidecars from the ignore check (#3504)

### LINKED ISSUE #3504: Office sidecars in graphify-out/converted/ are silently dropped when graphify-out is gitignored
## Summary

`detect()` converts Office files to markdown sidecars inside `graphify-out/converted/`, then discards them because that path is gitignored - which it is in every project that follows the documented gitignore advice. The files are dropped from the corpus **and** omitted from `skipped_sensitive`, so nothing reports the loss.

Installing `graphifyy[office]` therefore appears to fix Office support while changing nothing: it converts a loud failure into a silent one.

## Environment

- graphifyy 0.9.58 (also reproduces on 0.9.56)
- Windows, Python 3.13, installed via `uv tool install`
- `graphifyy[office]` installed: python-docx 1.2.0, openpyxl 3.1.5

## Reproduce

1. A project with some `.docx` files at the root.
2. A `.gitignore` containing `graphify-out/` (or `graphify-out/*`), which is what the skill's own "Migration and portability" guidance recommends.
3. `pip install 'graphifyy[office]'`
4. Run `detect(Path('.'))`.

**Expected:** the converted sidecars appear in `files['document']`.

**Actual:** they do not appear anywhere. `skipped_sensitive` is empty, `total_files` is unchanged, and there is no warning.

```python
detect(root, gitignore=True)    # 32 documents,  40,903 words
detect(root, gitignore=False)   # 41 documents,  73,071 words   <- the 8 sidecars
```

In this corpus the dropped files were 44% of the total word count, including the document the project treats as its authoritative plan. Nothing surfaced that.

## Cause

`detect.py`, in the Office branch (around line 1978 in 0.9.56):

```python
if p.suffix.lower() in OFFICE_EXTENSIONS:
    md_path = convert_office_file(p, converted_dir, root=root)
    if md_path:
        if _ignored_for_scan(md_path):
            continue                      # <-- dropped, and not recorded
        files[ftype].append(str(md_path))
    else:
        skipped_sensitive.append(str(p) + " [office conversion failed ...]")
    continue
```

`converted_dir` is `graphify-out/converted/`, and `_ignored_for_scan` honours `.gitignore`. Since the tool writes sidecars into a directory it tells users to gitignore, the check rejects the tool's own output.

The `continue` at the conversion-succeeded branch also means no entry is added to `skipped_sensitive`, so the corpus summary cannot show it.

The same applies to the Google Workspace branch immediately above, which writes to the same directory.

## Suggested fix

The ignore check exists to stop user files being scanned, not to filter output the tool just produced. Exempting the converted directory would be enough:

```python
if md_path:
    if _ignored_for_scan(md_path) and not md_path.is_relative_to(converted_dir):
        continue
    files[ftype].append(str(md_path))
```

If dropping them is intentional, the minimum would be recording them in `skipped_sensitive` with the reason, so the summary can say so rather than the corpus quietly shrinking.

## Workaround

`detect(root, gitignore=False)` restores them. In our case that also pulled in one gitignored file we needed to keep out of the graph, so it needed a matching `.graphifyignore` entry - worth noting for anyone else taking that route.

## Note on when it appears

It looks like a fresh install failing, but it is not. This project has sidecars dated a week before the breakage, from when `graphify-out/` was not yet gitignored. Adding the recommended gitignore line silently stopped Office content reaching the graph, and nothing in the output changed to indicate it.

ayushcodes10:
## Summary

Fixes #3504. `detect()` converts `.docx`/`.xlsx` files and Google Workspace shortcuts into markdown sidecars under `graphify-out/converted/`, then ran the same scan-ignore check against the sidecar's own freshly-written path. The documented "Migration and portability" gitignore advice puts the whole output directory inside a gitignored tree, so that check rejected the tool's own output for the same reason a user would gitignore it — the source document silently vanished from the corpus, with nothing recorded in `skipped_sensitive` or anywhere else to explain why. Installing `graphifyy[office]` therefore appeared to fix Office support while changing nothing observable.

The ignore check exists to keep *user* files out of the scan, not to filter output this same detection pass just produced from an already-admitted source file. Both conversion branches (Office and Google Workspace) now exempt a sidecar that lands under `converted_dir` from the ignore check — matching the issue's own suggested fix exactly.

## Test plan

- [x] Added two regression tests in `tests/test_detect.py`, one per conversion branch, each writing a `.gitignore` covering `graphify-out/` (matching the documented advice) and mocking the converter to confirm the sidecar still reaches `files['document']`.
- [x] Verified both tests fail without the fix (reverted `detect.py`, confirmed `AssertionError: 0 == 1` on both) and pass with it.
- [x] Full suite: `python3 -m pytest -q` — 5457 passed, only the pre-existing unrelated failures (`test_ollama_retry_cap.py` missing `openai` in this env, one flaky timing assertion in `test_ts_import_type_arguments.py`).

### DIFF
--- a/graphify/detect.py
+++ b/graphify/detect.py
@@ -1968,7 +1968,15 @@ def detect(root: Path, *, follow_symlinks: bool | None = None, google_workspace:
                     skipped_sensitive.append(str(p) + f" [Google Workspace export failed: {exc}]")
                     continue
                 if md_path:
-                    if _ignored_for_scan(md_path):
+                    # #3504: the sidecar lands under converted_dir, which the
+                    # documented .gitignore advice puts inside a gitignored
+                    # graphify-out/ -- an ignore check here would reject the
+                    # tool's own output for the same reason it should be
+                    # gitignored in the first place, silently dropping the
+                    # source document from the corpus. The ignore check exists
+                    # to keep USER files out of the scan, not to filter output
+                    # this same pass just produced from an already-admitted file.
+                    if _ignored_for_scan(md_path) and not md_path.is_relative_to(converted_dir):
                         continue
                     files[ftype].append(str(md_path))
                     total_words += _wc(md_path)
@@ -1979,7 +1987,10 @@ def detect(root: Path, *, follow_symlinks: bool | None = None, google_workspace:
             if p.suffix.lower() in OFFICE_EXTENSIONS:
                 md_path = convert_office_file(p, converted_dir, root=root)
                 if md_path:
-                    if _ignored_for_scan(md_path):
+                    # #3504: see the matching comment in the Google Workspace
+                    # branch above -- same sidecar-under-a-gitignored-output-dir
+                    # trap, same exemption.
+                    if _ignored_for_scan(md_path) and not md_path.is_relative_to(converted_dir):
                         continue
                     files[ftype].append(str(md_path))
                     total_words += _wc(md_path)
--- a/tests/test_detect.py
+++ b/tests/test_detect.py
@@ -872,6 +872,58 @@ def test_detect_converts_google_workspace_shortcuts_when_enabled(tmp_path, monke
     assert result["total_words"] > 0
 
 
+def test_detect_office_sidecar_survives_a_gitignored_output_dir(tmp_path, monkeypatch):
+    """#3504: the documented .gitignore advice puts graphify-out/ (and so
+    graphify-out/converted/, where Office sidecars land) inside a gitignored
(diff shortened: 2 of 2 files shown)


safishamsi:
Shipped in **v0.9.60** — landed on `v8` via cherry-pick with your authorship preserved. Thanks @ayushcodes10 — Office/Workspace converted sidecars are no longer dropped when graphify-out/ is gitignored. Release: https://github.com/Graphify-Labs/graphify/releases/tag/v0.9.60

Issue #3346, opened 2026-09-05
TypeScript: calls via namespace import or aliased named import produce no `calls` edge


munishjindal931:
# TypeScript/JS: calls through a namespace import or an aliased named import produce no `calls` edge

**Version:** graphifyy 0.9.54 (uv tool install), macOS, Python 3.13

## Summary

The TypeScript extractor resolves a call to a function imported by name (`import { foo }` → `foo()`), but not the same call reached through either

- a **namespace import** — `import * as api` → `api.foo()`, or
- an **aliased named import** — `import { foo as bar }` → `bar()`.

No `calls` edge is emitted in either case. Both look like the same root cause: the call is resolved on the callee identifier as written, rather than on the import binding it refers to.

This fails quietly. Extraction succeeds, nodes for both the caller and the callee exist, and only the edge between them is missing — so nothing in `GRAPH_REPORT.md` or the Step 4.5 health check flags it, and the graph reads as complete while an entire layer is disconnected. The visible symptom is the opposite of an error: functions that are heavily used appear to have no callers, which reads as dead code.

## Minimal reproduction

Four files:

```ts
// api.ts
export function fetchThing(id: string): string { return id; }
export function otherThing(id: string): string { return id; }
```

```ts
// namespace.ts
import * as api from './api';
export function useNamespaceCall(id: string) { return api.fetchThing(id); }   // no edge
```

```ts
// named.ts
import { otherThing } from './api';
export function useNamedCall(id: string) { return otherThing(id); }           // edge
```

```ts
// alias.ts
import { otherThing as renamed } from './api';
export function useAliasCall(id: string) { return renamed(id); }              // no edge
```

Run:

```python
from graphify.extract import collect_files, extract
from pathlib import Path

res = extract(collect_files(Path('repro')), cache_root=Path('repro'))
print('nodes:', sorted(n['label'] for n in res['nodes']))
for e in res['edges']:
    if e.get('relation') == 'calls':
        print('CALLS:', e['source'], '->', e['target'])
```

**Actual:**

```
nodes: ['alias.ts', 'api.ts', 'fetchThing()', 'named.ts', 'namespace.ts', 'otherThing()',
        'useAliasCall()', 'useNamedCall()', 'useNamespaceCall()']
  CALLS: named_usenamedcall -> api_otherthing
calls edges: 1
```

**Expected:** three `calls` edges. All six function nodes extract correctly; only the two binding-mediated edges are absent.

## Impact on a real corpus

[...]

PR #3517, opened 2026-09-12
fix(js): resolve calls inside exported functions, incl. aliased imports (#3346)

abhay-codes07:
Fixes the aliased-named-import half of #3346, by fixing the root cause it shares with a broader gap: **calls inside a top-level exported function were never resolved through the import table.**

## The bug

`_js_top_level_function_bodies` scanned only the **direct** children of the program node. A top-level `export function f(){}` / `export const g = () => {}` is an `export_statement` *wrapping* the declaration, so every exported function was skipped — and the calls inside it never became `uses` facts.

That matters specifically for aliased imports. A call resolves to its target one of two ways:
- the shared resolver matches the **written** callee name against the global symbol table (this is what makes plain `import { foo }; foo()` work), or
- the symbol-resolution `uses` path maps the call's **local name** through the per-file import-alias table (`local_aliases_by_file`).

For `import { bar as baz }; baz()` only the second path can work — the global table has no `baz`, only `bar`. But because `useAlias` was an exported function, no `uses` fact was ever produced for `baz()`, so the alias mapping never ran and no edge was emitted. It fails silently: caller and callee nodes both exist, only the edge is missing.

## The fix

Unwrap a non-re-export `export_statement` to its inner `function_declaration` / `lexical_declaration` in `_js_top_level_function_bodies`, so exported and non-exported functions are collected identically. The aliased call then resolves through the existing alias table to the real definition (`bar`), function and arrow forms alike. Re-exports (`export … from '…'`) are left alone — they declare no local body.

## Scope

This fixes the **aliased named import** case (issue's second repro) and, more generally, call-through-import resolution inside any exported function. The **namespace-import** case (`import * as api; api.foo()`) is a distinct mechanism — it's a member call whose receiver is a module alias, which needs namespace-binding capture and member-call resolution rather than the local-name alias table — and is left for a follow-up so this change stays focused and low-risk.

## Tests

`tests/test_js_exported_fn_call_resolution.py` — the aliased import resolves from an exported function and an exported arrow; the edge lands on the imported definition (`bar`), never a phantom `baz`; and the plain-named-import and non-exported-function paths are unchanged. With the fix reverted, the two exported-function aliased cases fail. The JS/TS + resolution suites are otherwise unchanged (1806 passed; the four failing names fail identically on clean `v8` on this Windows machine — pre-existing/flaky, unrelated).

### DIFF
--- a/graphify/extractors/resolution.py
+++ b/graphify/extractors/resolution.py
@@ -1562,7 +1562,25 @@ def _js_default_export_name(node, source: bytes) -> str | None:
 def _js_top_level_function_bodies(path: Path, root_node, source: bytes) -> list[tuple[str, object]]:
     bodies: list[tuple[str, object]] = []
     stem = _file_stem(path)
+    # A top-level `export function f(){}` / `export const g = () => {}` is an
+    # export_statement WRAPPING the declaration, not a bare program child, so
+    # scanning only direct children missed every exported function — and the
+    # calls inside them never became `uses` facts, so an aliased-import call
+    # (`import { bar as baz }; baz()`) never resolved through the import table
+    # (#3346). Unwrap a non-re-export export_statement to its inner declaration
+    # so exported and non-exported functions are treated identically.
+    top_nodes: list = []
     for node in root_node.children:
+        if node.type == "export_statement" and not any(
+            c.type == "string" for c in node.children  # `export ... from '...'` is a re-export
+        ):
+            top_nodes.extend(
+                c for c in node.children
+                if c.type in ("function_declaration", "lexical_declaration")
+            )
+        else:
+            top_nodes.append(node)
+    for node in top_nodes:
         if node.type == "function_declaration":
             name_node = node.child_by_field_name("name")
             body = node.child_by_field_name("body")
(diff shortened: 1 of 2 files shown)


abhay-codes07:
Checked both against the actual behavior:

**"export default arrow/function bodies no longer scanned" — false positive.** A NAMED default function IS scanned and resolves: `export default function run() { return baz(); }` with `import { bar as baz }` produces `run() -> bar()` (verified). It's caught because the unwrap collects the `function_declaration` child of the export_statement regardless of the `default` keyword. (And nothing was "no longer" scanned — before this PR *no* export_statement-wrapped body was scanned at all; this PR strictly adds coverage.)

**"Default-exported arrow functions are still skipped" — true, but a pre-existing anonymous-function trait, out of scope here.** `export default () => baz()` has no name, so there is no caller node to attribute the call to — the same reason any anonymous arrow/function gets no node. My change neither introduced nor worsened this; resolving it would mean synthesizing a caller node for the anonymous default export, a separate enhancement. This PR stays focused on the aliased-named-import root cause (calls inside *named* exported functions), which it fixes completely.

No code change from these — the named cases work and the anonymous-default case is consistent with existing anonymous handling.

safishamsi:
Shipped in **v0.9.60** — landed on `v8` via cherry-pick with your authorship preserved. Thanks @abhay-codes07 — calls inside exported functions and to aliased imports now resolve. Release: https://github.com/Graphify-Labs/graphify/releases/tag/v0.9.60

PR #3520, opened 2026-09-12
fix: don't report a valid C/C++ file without a trailing newline as a syntax error (#3513)

### LINKED ISSUE #3513: Incorrect "syntax errors" in C and C++ headers when they dont end with a newline character
## Description
When running `graphify update` on a C or C++ project that has a header without a newline '\n' character at the end you are given this warning:
```
warning: 1 file(s) had syntax errors and may be partially extracted: dir/to/file/my_file.h (first error at line 2, no symbols extracted)
[graphify watch] No code-graph topology changes detected; outputs left untouched.
Code graph updated. For doc/paper/image changes run /graphify --update in your AI assistant.
Tip: set GEMINI_API_KEY or GOOGLE_API_KEY to use Gemini for semantic extraction.
```
The header file is completely valid and compiles and runs fine.

## Graphify version
Graphify 0.9.48

## Reproduce the error
Create a C or C++ project with a header file that does not end with a newline character. For example (remove the newline char at the end):
```
#define A 8
#define AA 3
```
Now, running `graphify update` gives this warning but adding a newline character at the end makes it go away.

A-Levin:
Fixes #3513.

## Problem

A valid C/C++ file that does not end in a newline is reported as having a syntax error:

```
warning: 1 file(s) had syntax errors and may be partially extracted: my_file.h (first error at line 2, no symbols extracted)
```

The trigger is narrower than the issue title suggests, and worth stating precisely:

- **Not header-specific.** It reproduces on every C/C++-family extension routed to `extract_c`/`extract_cpp` — `.h`, `.c`, `.hpp`, `.cpp`, `.cc`, `.cxx`, `.cu`, `.cuh`, `.metal`.
- **Not "missing trailing newline" in general.** A file ending in ordinary code (`int x = 1;`) with no trailing newline is clean. It fires when the last line is a **preprocessor directive** with no terminating newline: tree-sitter-c/cpp's preproc grammar uses the newline as the directive's terminator and recovers with an ERROR node when it is absent.
- **Nothing is lost but the flag.** Node and edge sets are byte-identical with and without the newline, including on symbol-bearing files. Only `parse_errors` differs.

## Fix

`graphify/extractors/engine.py`, at the parse point in `_extract_generic`: when the grammar is `tree_sitter_c`/`tree_sitter_cpp` and the source does not end in `\n`, append one before `parser.parse`.

ISO C and C++ require a translation unit to end in a newline, so this supplies a byte the standard already mandates. It only extends the buffer past the last real token, so no existing node's offsets move — verified by diffing node dicts (`label`, `line`, `end_line`, `source_location`) and edge lists for the same fixture with and without the newline. Nothing in `engine.py` derives a value from the buffer length (`len(source)`, `source.count(`, `source.splitlines` are all absent), and the AST cache keys on `file_hash(path)`, which hashes the bytes on disk, so the in-memory buffer cannot desync a cache key.

Two notes on scope:

**Why not the post-hoc gate pattern.** The existing #2610/#2599 TypeScript carve-out gates on `len(nodes) <= 1 or multiline_error`. That cannot discriminate here: `#define`-only content yields just the file node in both the broken and the clean case, so `len(nodes) <= 1` is true either way. Rather than add a second, differently-shaped classifier beside it, this removes the mechanical cause — the grammar is simply missing its terminator token.

**Why the grammar restriction stays.** Appending is verified safe only for these two grammars, so it is scoped to them rather than applied to every language in `_extract_generic`. (Objective-C happens to be immune because `extract_objc` never sets `parse_errors` at all, but that is an accident of that extractor rather than a guarantee.)

Genuine errors are unaffected: a broken file whose last line is a directive **and** which has a real error earlier is still flagged, as is `#if` with no matching `#endif`, a lone `#`, and an unterminated block comment.

## Test

`tests/test_c_header_trailing_newline.py`, 29 tests, asserting on `result.get("parse_errors")` (matching the convention in `tests/test_cpp_nested_and_cli.py`) with one end-to-end `capsys` test for the printed warning.

Coverage: all 9 extensions with and without the newline; the three shapes a naive last-line heuristic would miss (a backslash-continued macro at EOF, a UTF-8 BOM, a comment preceding the directive); that node and edge sets are unchanged; that the append is scoped to the C/C++ grammars; and that genuine syntax errors are still reported.

Mutation results:

| Mutation | Tests killed |
|---|---|
| Delete the fix | 14 |
| Append `b" "` instead of `b"\n"` | 14 |
| Drop the `ts_module` guard | 1 (`test_fix_is_scoped_to_c_and_cpp_grammars`) |
| Blanket-suppress `parse_errors` for C/C++ instead of fixing the cause | 3 (the genuine-error tests) |

The scoping test needed a second attempt worth mentioning, in case it is useful elsewhere: comparing a Python file's extracted nodes with and without a trailing newline is structurally vacuous, because every non-C grammar I could load (python, bash, ruby, go, js, java, kotlin, rust, scala, json, elixir) is insensitive to a missing trailing newline for ordinary content — the comparison passes whether or not the guard exists. It now observes the mechanism directly, recording the bytes handed to `tree_sitter.Parser.parse`. Note `extract_python` parses twice (the main pass plus the docstring-rationale pass), and only the first goes through this code path, so the assertion checks every recorded call rather than the last.

Local run of the CI jobs on this branch:

- `skillgen --check` / `--audit-coverage` / `--schema-singleton` / `--monolith-roundtrip` / `--always-on-roundtrip`: all OK
- `pytest tests/ -q` (after `uv sync --all-extras --frozen`): **5621 passed, 14 skipped, 2 failed**
- `ruff check graphify tests`: clean
- `bandit -r graphify -ll`: unchanged from baseline

The 2 failures are both `test_hooks.py::test_rebuild_bodies_survive_a_graphify_root_symlink_loop`, and they fail identically on pristine `v8` (`522ea96`) with this branch not applied — `2 failed, 111 passed` in `tests/test_hooks.py` there. Environment-specific to my machine, unrelated to C/C++.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ Full release notes with details on each version: [GitHub Releases](https://githu
 - Performance: Python symbol resolution is roughly 47% faster — path resolution is memoized, each file parses once across both resolution passes, and the tree walk is iterative rather than recursive; extraction output is unchanged (#3500 / #3501 / #3502, thanks @abhay-codes07).
 - Fix: `graphify explain` now accepts a `path::Symbol` form to disambiguate a symbol that shares its name with its file, and the ambiguity hint now shows a form the resolver actually accepts (#3485, thanks @ayushcodes10).
 - Fix: the git hook now keeps its rebuild root inside the repository — a committed `.graphify_root` pointing outside the worktree is ignored and falls back to the repo top, so a checked-in marker can't steer the hook to scan or write outside the tree (#3265, thanks @ayushcodes10).
+- Fix: a valid C/C++ file that doesn't end in a newline is no longer reported as having a syntax error — not just headers, every C/C++-family extension (#3513, thanks @A-Levin).
 
 ## 0.9.58 (2026-09-10)
 
--- a/graphify/extractors/engine.py
+++ b/graphify/extractors/engine.py
@@ -3177,6 +3177,20 @@ def _extract_generic(
     try:
         parser = Parser(language)
         source = path.read_bytes() if source_override is None else source_override
+        # #3513: ISO C/C++ requires a translation unit to end in a newline,
+        # and tree-sitter-c/tree-sitter-cpp's preprocessor grammar relies on
+        # that newline to terminate a directive - without it, a file whose
+        # last line is a directive recovers with an ERROR node even though
+        # the directive is complete and legal, falsely flagging an
+        # otherwise-valid file as a syntax error. Supplying the byte the
+        # standard already requires only extends the buffer past the last
+        # real token, so no existing node's offsets change; it never
+        # introduces an error, so this isn't scoped further than "missing a
+        # trailing newline". Restricted to these two grammars because that is
+        # what has actually been verified safe here - not applied blanket to
+        # every language in the project.
+        if config.ts_module in ("tree_sitter_c", "tree_sitter_cpp") and not source.endswith(b"\n"):
+            source = source + b"\n"
         tree = parser.parse(source)
         root = tree.root_node
     except Exception as e:
--- a/tests/test_c_header_trailing_newline.py
+++ b/tests/test_c_header_trailing_newline.py
@@ -1,0 +1,178 @@
+"""A valid C/C++ file with no trailing newline must not be reported as having
+a syntax error (#3513).
+
+tree-sitter-c/tree-sitter-cpp requires a newline to terminate a preprocessor
+directive; when the directive is the last thing in the file and the file does
+not end in a newline, the grammar recovers with an ERROR node even though the
+directive itself is complete and legal. That trips the #2551 partial-extraction
+gate in ``_extract_generic`` (``root.has_error``) and reports a spurious
+syntax error on a file that compiles fine and loses no symbols.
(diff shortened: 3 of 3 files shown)


A-Levin:
Closing this — @imanolpg's fix for the same issue landed in 0.9.60 while this was open, so the change here is redundant and the branch no longer merges cleanly.

For anyone reading later: the shipped fix appends the newline for every grammar rather than only C/C++, which is the one substantive difference from what's here. I checked every other grammar I could load (python, bash, ruby, go, js, java, kotlin, rust, scala, json, elixir) and none is sensitive to a missing trailing newline for ordinary content, so I don't believe it causes a problem in practice.

The test coverage built alongside this one still applies to the shipped fix — the other C/C++-family extensions beyond `.h`/`.hpp`, the shapes where the unterminated directive isn't the last physical line (backslash continuation, UTF-8 BOM, a comment ahead of the directive), and a guard that a genuinely broken file is still reported, which nothing in the repo asserted before. I'll send that separately as a tests-only follow-up rather than leave it buried in a closed PR.

Thanks @imanolpg for the report and the fix.

PR #3529, opened 2026-09-13
fix: bare plural "tokens" in a prose file is no longer treated as a secret

### LINKED ISSUE #3527: detect._is_sensitive silently drops every design-token source file (*_tokens.dart, tokens.ts, TOKENS.md)
**Version:** graphifyy 0.8.38

## Summary

`_GENERIC_KEYWORD_PATTERNS` treats `tokens?` as a credential keyword, and `_generic_keyword_hit` flags it when it ends the filename stem. In any codebase with a design-token layer, that silently removes the whole layer from the graph — design tokens are one of the most common uses of the word "token" in source trees, and they are never secrets.

## Repro

```python
from pathlib import Path

_is_sensitive(Path("app/lib/theme/shell_tokens.dart"))   # True
_is_sensitive(Path("src/design/tokens.ts"))              # True
_is_sensitive(Path("src/hard-tokens.ts"))                # True
_is_sensitive(Path("TOKENS.md"))                         # True

_is_sensitive(Path("src/api_token.txt"))                 # True  (correct)
_is_sensitive(Path("src/tokenizer.py"))                  # False (correct)
```

## Impact

In one Flutter + TypeScript monorepo this dropped six files that the project's own docs call the live design-system reference — `shell_tokens.dart`, `field_guide_tokens.dart`, `universal_tokens.dart`, `guide_atmosphere_tokens.dart`, `guide_motion_tokens.dart`, `guide_plate_tokens.dart` — plus `hard-tokens.ts` and `TOKENS.md`. **154 nodes from two of those files alone.** They were invisible to every `graphify query`, with no warning at query time.

There is a second-order effect worth flagging. `watch._rebuild_code` selects files via `detect()`, so it drops them, while a full rebuild that goes through `extract.collect_files()` keeps them. The two paths therefore converge on fixed points ~200 nodes apart, so committed graph artifacts churn indefinitely: the hook rewrites them, the developer restores, the next commit rewrites again.

## Why the existing guard does not catch this

The `_generic_keyword_hit` docstring already anticipates topic-word false positives (#436, #718) and handles the mid-slug case well. But the "keyword ends the stem" rule is precisely what design-token filenames hit: the content noun *is* "tokens", it just names design tokens rather than credentials.

## Suggested fixes (in rough order of preference)

1. **Exempt compiled-language source extensions from the generic keyword stage.** A secret store is essentially never a `.dart`, `.swift`, `.kt`, `.go`, or `.rs` file. Keeping the stage for `.txt`, `.json`, `.yaml`, `.env`-ish files preserves its value with far fewer false positives.
2. **Treat a plural `tokens` as a topic word.** Credential files are overwhelmingly singular (`api_token`, `access_token`); `tokens.ts` / `*_tokens.dart` read as a collection of design tokens. Splitting the `tokens?` pattern into singular-flagged / plural-allowed would fix the common case cheaply.
3. **Provide an escape hatch.** There is currently no way to override this from a project — `.graphifyignore` only subtracts. A `.graphifyallow`, or honouring a negation entry, would let a project opt specific files back in.

Happy to send a PR for (1) or (2) if you have a preference.

HARSHAVARDHAN-RAJU5:
## Summary
Fixes #3527. `detect._is_sensitive` was silently dropping design-token
files from the graph because "tokens" matched a generic credential
keyword.

- The `.dart`/`.ts` cases from the original report (`shell_tokens.dart`,
  `tokens.ts`, `hard-tokens.ts`) are already covered by the existing
  graphable-source exemption in this version, so those pass today
  without changes.
- The remaining gap: a bare prose file named `TOKENS.md` was still
  dropped, because the bare-keyword rule didn't distinguish plural
  "tokens" (a design-token reference doc) from singular "token" or
  other keywords' bare plurals (`secrets.md`, `passwords.md`), which
  correctly stay excluded.

`_is_prose_note` in `graphify/detect.py` now special-cases bare plural
"tokens" for prose extensions (`.md`, `.rst`, `.org`, `.adoc`, `.tex`)
as a topic word rather than a credential dump. Non-prose extensions
(`tokens.txt`, `tokens.json`) are unaffected and still excluded.

## Testing
- Added 7 regression tests in `tests/test_detect.py` covering the
  plural-prose exemption, the still-excluded cases (singular, other
  keywords' plurals, non-prose extensions), and the original `.dart`/
  `.ts` repro.
- `uv run pytest tests/test_detect.py` — 258 passed, 11 skipped, no
  regressions.

## Files changed
- `graphify/detect.py` — new `_BARE_PLURAL_TOKENS` pattern, updated
  `_is_prose_note`
- `tests/test_detect.py` — regression tests
- `CHANGELOG.md` — unreleased entry


safishamsi:
Shipped in **v0.9.62** (now on PyPI: `uv tool install graphifyy==0.9.62`). Your commits were cherry-picked with authorship preserved, so this shows up under your GitHub contributions. Thanks @HARSHAVARDHAN-RAJU5!

PR #3531, opened 2026-09-13
fix(extract): skip calls to sourceless stubs

### LINKED ISSUE #3156: Suppress sourceless external call targets like FastAPI HTTPException from bridge analysis
## Problem

When graphifying a FastAPI-heavy Python package, `HTTPException` becomes a high-degree cross-community bridge / god node even though it is an external framework exception constructor, not a project architecture abstraction.

Observed on `graphifyy==0.9.50` while graphifying `packages/bsquant_server` in a Python monorepo:

- Graph size: 3,634 nodes / 10,820 graph edges / 143 communities.
- Node: `httpexception`
  - label: `HTTPException`
  - source_file: empty string
  - origin: `ast`
  - degree: 124
  - relations: `calls: 124`
  - confidence: `EXTRACTED: 9`, `INFERRED: 115`
- The node appeared as a top bridge / god-node-like result, crossing router communities such as Catalog Router, Workflows Router, Datasets Router, Factors Router, Components Router, and Shared API.

The affected source patterns are ordinary FastAPI error responses, e.g.:

```python
from fastapi import Cookie, Header, HTTPException

raise HTTPException(status_code=401, detail="missing credentials")
raise HTTPException(status_code=403, detail="permission denied")
```

and router exception translation such as:

```python
except CatalogAuthorizationError as exc:
    raise HTTPException(status_code=403, detail=str(exc)) from exc
```

These edges make `HTTPException` look like a core cross-module dependency, but it is really framework error-response construction noise.

## Why it seems to happen

From a quick read of the installed extractor:

- The cross-file call resolver walks `all_raw_calls` and skips known language builtins, member calls, bash calls, etc.
- It then resolves by global label candidate.
- For sourceless external nodes, `_lang_family(nid_to_source_file.get(candidate))` is `None`, so the cross-language guard keeps the candidate.
- If there is a unique candidate but no import evidence, the resolver still emits `calls` with `confidence="INFERRED"` and `confidence_score=0.85`.

Relevant installed-package locations in `graphifyy==0.9.50`:

- `graphify/extract.py`: cross-file raw-call loop around the `callee in _LANGUAGE_BUILTIN_GLOBALS` guard and candidate filtering.
- `graphify/extract.py`: direct call emission path that emits `calls` / `INFERRED` / `0.85` when `has_import_evidence` is false.

## Expected behavior

External framework constructors / exceptions such as FastAPI `HTTPException` should not become central architecture bridge nodes by default.

Possible fixes:

1. Suppress cross-file direct `calls` edges to sourceless external call targets unless the target is a project-defined callable or has explicit import evidence tied to a local module.
2. Namespace sourceless external targets as external dependencies and exclude/downweight them from god node, bridge, and surprising-connection analysis.
3. Add common external/framework exception constructors to a noise set, e.g. `HTTPException`, while keeping project-defined exception classes intact.
4. Model this as `raises` / `error_response` metadata rather than a project-internal `calls` edge.

## Reproduction sketch

A minimal FastAPI module likely reproduces it:

```python
from fastapi import APIRouter, HTTPException

router = APIRouter()

@router.get("/x")
def get_x():
    raise HTTPException(status_code=404, detail="not found")
```

Run:

```bash
graphify .
```

Then inspect `graphify-out/graph.json` for the `httpexception` node and its incoming `calls` edges.

## Acceptance criteria

- `HTTPException` and similar sourceless external framework constructors no longer appear as high-degree project bridge / god nodes.
- Calls to project-local exception classes still remain represented where useful.
- If external calls are retained, they are typed or downweighted so report sections do not present them as core architecture abstractions.
- Add a regression test for a small FastAPI file that raises `HTTPException(...)` from multiple handlers.

DevChiniwala:
Fixes #3156

## Summary

- prevent same-file AST call extraction from targeting source-less unresolved stubs
- prevent cross-file name-only call resolution from targeting those stubs
- preserve type/reference edges and calls to source-backed project definitions

## Why

Unresolved external names can be represented as source-less nodes so that reference edges remain queryable. The call emitters were also treating those placeholders as callable definitions. A common example is `HTTPException` imported from FastAPI: the placeholder accumulated calls from unrelated files and became a misleading bridge node in community analysis.

This keeps source-less placeholders available for reference analysis while requiring a source-backed definition for `calls` edges. Local project-defined exceptions and other source-backed call targets continue through the existing resolution path.

## Validation

- `python -m pytest tests\\test_sourceless_call_targets.py -q` — 2 passed
- `python -m pytest tests\\test_extract.py tests\\test_multilang.py tests\\test_case_sensitive_resolution.py tests\\test_sourceless_call_targets.py -q --basetemp ... -k "not c_include_out_of_root_target_id_is_deterministic_across_checkout_paths"` — 277 passed, 36 skipped, 1 deselected
- `ruff check graphify\\extractors\\engine.py graphify\\extract.py tests\\test_sourceless_call_targets.py` — passed
- `python -m compileall -q graphify` — passed
- `git diff --check` — passed

The full local suite completed with 5,207 passed and 269 skipped; 49 pre-existing Windows/optional-environment failures remain, including Windows path-length, hook-shell, optional backend, and platform-specific filesystem cases.


safishamsi:
Shipped in **v0.9.62** (now on PyPI: `uv tool install graphifyy==0.9.62`). Your commits were cherry-picked with authorship preserved, so this shows up under your GitHub contributions. Thanks @DevChiniwala!

Issue #3314, opened 2026-09-03
install: hook-guard is wired to the four highest-frequency tools with no timeout and no matcher knob — one interpreter spawn per Bash/Grep/Read/Glob call


yotamleo:
After `graphify install --platform claude`, every single `Bash`, `Grep`, `Read` and `Glob` tool call in every Claude Code session starts an extra Python interpreter to print (or suppress) the orientation nudge — those are the four highest-frequency tools an agent uses, and the generated entries carry no `timeout`, so a wedged guard stalls the tool call for Claude Code's 600 s default. On a workstation running several concurrent agent sessions the cost is no longer the guard's own runtime but process creation itself, and the whole surrounding hook chain starts missing its budget.

I hit this while maintaining a downstream harness: seven concurrent sessions on one Windows box, CPU at 9 %, 20 GB RAM free — not compute-bound — and the harness's own 15 s per-hook budget began failing closed. A process snapshot showed ~30 transient Python processes with blank command lines appearing and vanishing between samples: hook-guard launches. The guard was not the only cost, but on `Read` and `Glob` it was the *only* extra spawn, and those are the tools an exploring agent calls most.

The guard is genuinely useful and I want to keep it. What I cannot do is tune it: the matchers and the (absent) timeout are hard-coded, and `GRAPHIFY_HOOK_STRICT` is the only knob exposed.

## Measurements

graphify 0.9.53, Windows 11, `graphify.exe hook-guard search` against a repo with a fresh `graphify-out/graph.json`:

- ~170 ms wall per call, idle box.
- 24 concurrent launches: 632 ms wall, versus 625 ms for 24 bare `bash -c true`. So it is an ordinary process spawn, not a heavy one — which is exactly why the fix is to spawn it less often rather than to make it faster.

That per-call cost is fine on its own and unaffordable multiplied by four hot tools across N sessions.

## Repro

```
graphify install --platform claude
python -c "import json;print(json.dumps([h for h in json.load(open('.claude/settings.json'))['hooks']['PreToolUse'] if 'graphify' in str(h)], indent=2))"
```

Observe two entries, matchers `Bash|Grep` and `Read|Glob`, neither carrying a `timeout` key.

## Evidence

Upstream HEAD `33362d969292b57eda82f3fbd9eb5f3f5bc9bbc2` (branch `v8`):

[...]

PR #3532, opened 2026-09-13
install: bound the guard hooks with a timeout (#3314)

ayushcodes10:
## Summary

Toward #3314. The generated Claude/Codebuddy `PreToolUse` guard hooks (`graphify hook-guard search` on `Bash|Grep`, `graphify hook-guard read` on `Read|Glob`) carried no `timeout`, so Claude Code's unset-hook default of 600s applied — a single wedged guard (a stuck filesystem, a hung subprocess) would stall the surrounding tool call for ten minutes, on the four highest-frequency tools an agent uses. The reporter measured the guard itself at ~170ms warm and observed dozens of transient process spawns per session on a multi-agent workstation, with the harness's own hook budget beginning to fail closed.

This PR addresses the reporter's second, lowest-risk ask in their priority order ("a `timeout` on the generated entries... a knob would be nice; a sane default would be enough") — both generated hook entries now carry `"timeout": 10`, roughly 60x the measured warm cost and still 60x below the unset default.

I deliberately left the other two asks out of scope for this PR:
- **A matcher knob** (`--hook-matchers`) requires designing a new CLI flag/config surface — a maintainer-preference decision, not a mechanical fix.
- **Skipping the Codex `hook-check` no-op entry** — the existing code already documents it as an intentional no-op with a stated reason (Codex Desktop rejects `additionalContext` on `PreToolUse`), so removing it is a bigger behavioral call than I want to make unilaterally; happy to follow up if there's a preference.

## Test plan

- [x] Added one regression test per hook (`tests/test_search_hook.py`, `tests/test_read_hook.py`) confirming a numeric `timeout` is present and stays within a small bounded range, without pinning the exact value.
- [x] Manually verified the generated hook JSON now carries `"timeout": 10` on both entries.
- [x] Full suite: `python3 -m pytest -q` — 5487 passed. Failures are all pre-existing/environmental and unrelated: `test_ollama_retry_cap.py` (missing `openai` in this env), one flaky timing assertion in `test_ts_import_type_arguments.py`, and `test_hyperedge_convex_hull_js_is_geometrically_sound` (a `NODE_OPTIONS`/sandbox artifact in this environment — confirmed it fails identically on a clean checkout with none of this PR's changes applied).

### DIFF
--- a/graphify/install.py
+++ b/graphify/install.py
@@ -338,6 +338,14 @@ def _claude_pretooluse_hooks(strict: bool = False, project: bool = False) -> "li
     When ``strict`` is set, the read hook carries ``--strict`` so it blocks the
     first raw read per session (Claude Code only). The ``GRAPHIFY_HOOK_STRICT`` env
     var can force it on or off at runtime without a reinstall.
+
+    Each entry carries a ``timeout`` (#3314): unset, Claude Code defaults a
+    command hook to 600s, so a single wedged guard (a stuck filesystem, a
+    hung subprocess) stalls the surrounding tool call for ten minutes on
+    every Bash/Grep/Read/Glob call -- the four highest-frequency tools an
+    agent uses. The guard itself measures ~170ms warm; 10s is generous
+    headroom over that while still two orders of magnitude below the
+    unset default.
     """
     exe = _resolve_graphify_exe(project=project)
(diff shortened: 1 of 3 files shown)


safishamsi:

Issue #3124, opened 2026-08-26
JS/TS: calls inside an anonymous callback are still dropped when there is no enclosing NAMED function (residual of #1740)


cbartens:
## Summary

#1740 (closed COMPLETED, 2026-07-09) fixed calls inside anonymous arrow callbacks by attributing them to the enclosing function. That works — but only when an enclosing **named** function exists. When the nearest enclosing scope is itself anonymous, no `calls` edge is emitted at all.

#1740's own wording covers this case — *"Nothing is emitted — not attributed to the enclosing function, **not to the module node**"* — and the module-node half looks unimplemented.

This hits test files hardest, because `it("…", async () => { … })` is exactly the shape with no named parent. Those are also the files you most want in a blast-radius answer.

## Minimal repro

`target.ts`:
```ts
export function target(x: number) { return x; }
```

`named-parent.ts` — anonymous arrow, but inside a named function:
```ts
import { target } from "./target.js";
export function makeApp() {
  register("/a", async (c) => { return target(c); });
}
declare function register(p: string, h: (c: number) => unknown): void;
```

`anon-parent.ts` — byte-identical call, nearest enclosing scope is anonymous:
```ts
import { target } from "./target.js";
it("does the thing", async () => {
  register("/a", async (c) => { return target(c); });
});
declare function it(n: string, f: () => unknown): void;
declare function register(p: string, h: (c: number) => unknown): void;
```

```bash
graphify extract . --code-only
graphify affected "target()" --depth 2
```

### Actual

```
- makeApp() [calls] named-parent.ts:L4
- anon-parent.ts [imports] anon-parent.ts:L1
- named-parent.ts [imports] named-parent.ts:L1
```

`named-parent.ts` gets a **`calls`** edge. `anon-parent.ts` gets only an `imports` edge — no `calls` edge exists for its identical call.

### Expected

A `calls` edge for `anon-parent.ts` too, attributed to the module node when no named enclosing symbol is available.

## What it cost on a real codebase

`graphify affected "requireFamilyMemberHono()" --depth 1` returned 21 of the 25 source files containing a real call site. The four misses were all test files whose calls sit directly inside `it(...)`:

```
packages/api/src/auth.test.ts
packages/api/src/auth.mcp.test.ts
packages/api/src/mcp/observability.test.ts
packages/api/src/mcp/token-revoke.integration.test.ts
```

[...]

PR #3533, opened 2026-09-13
fix(js): attribute module level anonymous closure calls to the file node (#3124)

ayushcodes10:
## Summary

Toward #3124. #1740 fixed calls inside an anonymous callback by attributing them to the enclosing NAMED function — but that only works when a named function actually encloses the call. When the nearest enclosing scope is itself anonymous, nothing is emitted at all. `it("...", async () => { ... })` is exactly this shape, and it hits test files hardest, since that's the idiomatic way to write one.

#1740's own issue text named the intended fallback ("not attributed to the enclosing function, **not to the module node**") but the module-node half was never implemented. This closes it: a topmost closure argument to a module-level call statement now gets its own `function_bodies` entry keyed by the file node, mirroring the existing `_js_scan_member_assignments` fix (#3408) that already does this for `this.X = fn` assignments found in the exact same kind of statement. Once tracked, the ordinary call-walking machinery — including its own descent into further nested closures — picks it up from there, so the issue's own doubly-nested repro (`it(...)` containing `register(...)` containing the real call) resolves correctly.

**Known gap, deliberately left open** (documented in the new test file's own module docstring, and pinned by a dedicated test): a callee that is **both** directly imported by the same file **and** only called from a module-level anonymous closure still does not resolve. Tracing this down: the cross-file resolution pass in `extract.py` dedupes strictly by `(source, target)` regardless of relation, so the file node's own pre-existing `imports`/`imports_from` edge to that target pre-empts the `calls` edge from ever being added. This collision is specific to a caller that *is* the file node — a NAMED function's calls never hit it, since imports are always attributed to the file node, never to a function node — so it only surfaces now that the file node is a valid `calls` source for the first time. Fixing that dedup to be relation-aware reaches across every language's resolution pass with a large, hard-to-bound blast radius, so I've scoped it out of this PR rather than guess at a narrower patch. This does mean the issue's own literal repro (where `target` is both imported and called this way) still won't fully resolve — I want to be upfront about that rather than claim a complete fix.

## Test plan

- [x] New `tests/test_js_module_level_anonymous_calls.py` with 4 tests: same-file call inside a top-level anonymous closure; the exact doubly-nested repro from the issue; confirmation that #1740's named-enclosing-function case is unaffected; and a pinned test documenting the known direct-import-collision gap above.
- [x] Verified the two positive tests fail without the fix and pass with it (reverted `engine.py`, confirmed `AssertionError`).
- [x] Ran the full existing JS/TS test suite (`test_js_callback_calls.py` and 15 other JS/TS test files, 186 tests) — all pass, no regressions.
- [x] Full suite: `python3 -m pytest -q` — 5490 passed, only the pre-existing unrelated failures (`test_ollama_retry_cap.py` missing `openai` in this env, one flaky timing assertion in `test_ts_import_type_arguments.py`).


safishamsi:

PR #3534, opened 2026-09-13
fix(merge): respect C# type visibility across repos

### LINKED ISSUE #3360: merge-graphs: a third-party receiver type binds to a same-named repo-local declaration, fabricating cross-repo `calls` edges
### Summary

`merge-graphs` fabricates cross-repo `calls` edges whenever a **third-party** receiver type shares its bare name with exactly one repo-local declaration elsewhere in the merged graph.

`link_cross_repo_member_calls` binds a parked call when the receiver's bare type name resolves to exactly one declaration in another repo. That single-definition guard cannot distinguish the two reasons a call is parked in the first place:

1. the receiver's type lives in a **sibling repo** — the case the pass was built for (#3152), and
2. the receiver's type is **third-party** (a NuGet/Maven/npm dependency), so it is absent from *every* repo's corpus.

In case 2 the type name is still unresolved, so the call is still parked, and if any single repo in the merge happens to declare a type with the same bare name, the pass binds to it. Nothing in the parked entry records that the receiver was resolved through an external package reference.

`IValidator` is the worst offender in C#, because `FluentValidation.IValidator<T>` is near-ubiquitous and `IValidator` is also a name teams pick for their own domain interfaces.

### Version

`graphifyy` 0.9.54 (macOS, Python 3.11)

### Reproduction

Two repos, three files, no shared code and no reference of any kind between them.

`app/src/OrderService.cs` — consumes **FluentValidation's** `IValidator<T>`:

```csharp
using FluentValidation;

namespace App.Services;

public class OrderService
{
    private readonly IValidator<OrderRequest> _validator;

    public OrderService(IValidator<OrderRequest> validator) => _validator = validator;

    public async Task<bool> CreateAsync(OrderRequest request, CancellationToken ct)
    {
        var result = await _validator.ValidateAsync(request, ct);
        return result.IsValid;
    }
}

public class OrderRequest
{
    public string Id { get; set; } = "";
}
```

`lib/src/IValidator.cs` — an unrelated repo that happens to declare its own `IValidator`:

```csharp
namespace Lib.Domain.Interfaces;

public interface IValidator
{
    Task<bool> ValidateAsync(object request, CancellationToken ct);
}
```

`lib/src/Widget.cs` — filler so `lib` is a normal repo:

```csharp
namespace Lib.Domain;

public class Widget
{
    public string Name { get; set; } = "";
}
```

Then:

```bash
graphify update app
graphify update lib
graphify merge-graphs app/graphify-out/graph.json lib/graphify-out/graph.json --out merged.json
```

### Actual

```
resolved 1 member call(s) across repos
Merged 2 graphs -> 19 nodes, 19 edges
```

```
app:.CreateAsync() --calls[INFERRED 0.8]--> lib:.ValidateAsync()
  caller src: src/OrderService.cs:L13
  target src: src/IValidator.cs
  _cross_repo_call: True
```

### Expected

`resolved 0 member call(s) across repos`. `app` has no `using Lib.Domain.Interfaces`, no project reference, and no package reference to `lib`. The receiver binds to `FluentValidation.IValidator<T>`, which is in neither corpus, so the call is not answerable and should stay parked.

### Impact

On a 27-repo C# fleet merged with `merge-graphs` (47,010 nodes / 86,005 edges), **25 of 39 cross-repo edges (64%) were false positives** — every one of them a `.ValidateAsync()` binding onto the single repo in the merge that declares its own `IValidator`.

Because all 25 landed on one node, that node became the merged graph's top cross-repo bridge by betweenness. The false edges also pull test methods into the architecture view: unit-test methods named `ValidateAsync_ReturnsSuccess_WhenRequestIsValid()` were bound to another repo's interface.

The failure is quiet — the edges are well-formed, carry `confidence: INFERRED / 0.8`, and read as plausible service-to-service dependencies. Anyone using the merged graph to answer "which services depend on which" gets a materially wrong answer, and the noise scales with fleet size: the more repos merged, the more chances a third-party name collides with exactly one local declaration.

### Suggested fix

Add a visibility guard before binding — the caller's compilation unit must actually be able to see the candidate declaration. For C#, the caller's file must name the candidate's namespace, via `using` or fully qualified:

```python
candidates = [
    node for node in by_name.get(receiver_type, [])
    if merged.nodes[node].get("repo") != caller_repo
    and _suffix(merged.nodes[node].get("source_file")) in suffixes
    and _caller_names_namespace(caller_data, merged.nodes[node])   # new
]
```

This is the same class of guard as `_LANG_SUFFIXES` (which already stops a Java `Greeter` binding to a Python one) — extended from "same language" to "actually reachable".

Two alternatives, if a source-reading guard is too heavy for merge time:

- **Record external resolution when parking.** If the resolver knows the receiver's type was introduced by an import that resolves outside the repo (a package reference rather than a local file), stamp the parked entry and never bind it cross-repo. This is strictly more accurate, since it uses information available at extraction time.
- **Park the namespace-qualified name.** Where the extractor can determine the receiver's declaring namespace, park `Lib.Domain.Interfaces.IValidator` instead of bare `IValidator`, so the name itself carries enough to disambiguate.

Related: #437 and #2460 are the cross-file / cross-module analogues of the same name-only matching problem; #2343 is the same missing-guard shape on the stub-resolution branch.

### Workaround

Post-merge prune: drop any `_cross_repo_call` edge whose caller file does not name the target's declaring namespace. On the 27-repo merge above this removed exactly the 25 false edges and kept all 14 genuine ones. Note that the direction of a surviving edge must be read from the JSON's `source`/`target` — a `networkx` round-trip through an undirected `Graph` returns endpoint pairs in arbitrary order.

DevChiniwala:
Fixes #3360

## Summary

- prevent `merge-graphs` from binding a parked C# call to an unrelated same-named type
- reuse the existing C# namespace, `using`, alias, and lexical-scope resolver
- preserve valid cross-repository calls when the target type is actually visible

## Problem

The cross-repository member-call pass intentionally parks calls whose receiver type is absent from the current repository, then resolves them after graph composition. Its candidate guard checked the candidate's language, source file, repository, and uniqueness, but matched only the bare receiver name for C#.

That allowed an unresolved third-party type such as `FluentValidation.IValidator` to bind to an unrelated `Lib.Domain.Interfaces.IValidator` from another merged repository. The resulting `calls` edge is well-formed and plausible, but fabricates a dependency that does not exist.

## Fix

For parked C# calls, resolve the receiver through `CsharpNameResolver` before emitting the cross-repository edge. The resolver already owns Graphify's C# namespace/`using`/alias visibility rules, so the merge pass now accepts only the exact candidate selected by those rules. Other parked languages retain their existing behavior, and source-backed C# calls resolved within a repository are unchanged.

## Validation

- `python -m pytest tests\\test_cross_repo_external_call_guards.py -q` — 3 passed
- `python -m pytest tests\\test_cross_repo_member_calls.py tests\\test_cross_repo_external_call_guards.py tests\\test_csharp_type_resolution.py tests\\test_global_graph.py tests\\test_merge_graphs_cli.py -q` — 89 passed
- `ruff check graphify\\cross_repo_calls.py tests\\test_cross_repo_external_call_guards.py` — passed
- `python -m compileall -q graphify` — passed
- `git diff --check` — passed

The regression test first failed on the current `v8` implementation (`1 failed, 1 passed`): the unrelated third-party receiver incorrectly resolved as one cross-repository call. It passes after the guard, including a real two-root C# extraction case.


safishamsi:
Shipped in **v0.9.62** (now on PyPI: `uv tool install graphifyy==0.9.62`). Your commits were cherry-picked with authorship preserved, so this shows up under your GitHub contributions. Thanks @DevChiniwala!

PR #3538, opened 2026-09-13
fix(csharp): resolve a member access to the receiver type's property, so DbSet query sites link (#3528)

### LINKED ISSUE #3528: ORM query call-sites are invisible: the table name is never in the source, so no SQL-literal fix can link them (EF Core / C#)
**Version:** graphify 0.9.56, Linux, `graphify-out/graph.json` 14,360 nodes
**Corpus:** .NET 10 / C# layered DDD backend + React SPA — EF Core 10 with a single `DbContext`, 27 mapped entities, Postgres.

## Summary

ORM query call-sites are not modelled, so the graph cannot answer **"what code reads or writes this table?"** — the single most useful data-layer question — on any ORM-based corpus.

This is the sibling of #2884, and the distinction matters because the fix is different. #2884 reports SQL **string literals** never linking to the tables they name (`SELECT … FROM volunteer_assignments` in a `.ts` file). This issue is the case where **the table name never appears in the source at all**:

```csharp
await db.Users.AsNoTracking().Where(u => u.Email == email).FirstOrDefaultAsync(ct);
```

There is no string to parse. The link is `db.Users` → entity type `ApplicationUser` → table `AspNetUsers`, and only the ORM's own mapping knows the last hop. SQL-literal extraction cannot reach it, which is why #2884's fix will not cover this even once it lands.

## What the graph has today, and what it is missing

The `DbContext` node exists and is well connected — but every edge stops at the *type*, at the declaration line:

```
$ graphify explain "AppDbContext"
Node: AppDbContext
  Source:    src/Cluckwork.Infrastructure/Persistence/AppDbContext.cs L18
  Degree:    83

  --> Account        [references] …/AppDbContext.cs:L25
  --> Flock          [references] …/AppDbContext.cs:L26
  --> DailyEntry     [references] …/AppDbContext.cs:L28
  --> Customer       [references] …/AppDbContext.cs:L33
  …
```

Those 83 edges all point at the `DbSet<T>` **declarations**. Not one points at a **use**. The hundreds of `db.<DbSet>.<method>(…)` call-sites scattered across repositories, middleware and handlers are neither nodes nor edges, so the data layer is connected to its own declaration file and to nothing that queries it.

## Reproduce

Any ORM corpus — EF Core, SQLAlchemy, ActiveRecord, Prisma, GORM. Concretely, on a C#/EF repo:

**1 — the DbSet property is unreachable by name.** A same-named symbol anywhere else in the corpus wins:

```
$ graphify explain "Users"
Node: users
  ID:        tools_simulation_k6_auth_smoke_users
  Source:    tools/simulation/k6/auth-smoke.js L25
  Type:      code
  Degree:    1

Connections (1):
  <-- auth-smoke.js [contains] tools/simulation/k6/auth-smoke.js:L25
```

That is a local `const` in an unrelated k6 load-test script. The EF `DbSet<ApplicationUser> Users` — queried from ~40 sites — is not a candidate at all.

**2 — the natural question returns unrelated nodes.**

```
$ graphify query "which methods query the Users table"
Graph: graphify-out/graph.json (14360 nodes) | Start: ['users', 'Table', '.QueryColumnsAsync()']
NODE users   [src=tools/simulation/k6/auth-smoke.js loc=L25]
NODE Table   [community=SchemaDocsTests]
NODE .QueryColumnsAsync()  [src=…/SchemaDocsTests.cs loc=L1656]
NODE login()               [src=tools/simulation/k6/auth.js loc=L84]
…
```

52 nodes, and not one is an actual query site. The traversal anchored on the k6 variable and on a test helper that happens to contain the word `Table`.

## Why it matters

We shipped a tenant-isolation guard that had to enumerate **every EF query lacking a global query filter** — a security control, where a missed call-site is a cross-tenant data leak. The graph could not answer it, so the guard was built as a hand-rolled Roslyn AST walk instead.

That is the general shape of the loss: for ORM corpora, the graph cannot do

- **impact analysis** — what breaks if this column changes,
- **security/compliance enumeration** — every site that touches a PII table,
- **completeness cross-checks** — is this hand-built allow-list missing a site?

and those are precisely the questions a schema-heavy project buys a knowledge graph for. The README's promise is a queryable codebase graph; on ORM code the data layer is the part it cannot answer about.

## Suggested approach

1. **Emit a node per ORM query call-site** (file + line), with an edge to the entity type it resolves through. Purely syntactic for the common shape: a member access on a field/property whose declared type is `DbSet<T>` / `Model` / `Repository<T>`, per language.
2. **Add the entity → table hop.** Where the mapping is discoverable — an attribute (`[Table("…")]`), a fluent call (`ToTable("…")`), a naming convention, or an already-extracted schema file — emit `entity --maps_to--> table` so the call-site reaches the table transitively. Where it is not discoverable, stop at the entity; that alone is most of the value.
3. **Let an explicit table vocabulary win the disambiguation.** If the corpus contains schema DDL (or a generated schema doc), prefer a table/DbSet node over an unrelated same-named local when resolving `explain "Users"`. This is the concrete fix for reproduce-case 1.
4. **`.sql` files and ORM migration files in the corpus** — migration classes are extracted here, but the DDL inside them is not, so even the declaration side is partial.

Cross-links: **#2884** (the SQL-literal half of the same goal — these two together are what makes a data layer navigable) and **#3034** (C# type references resolving to a same-named `DbSet<T>` property, which is the inverse resolution bug and suggests the `DbSet` shape is already partly recognised in the C# extractor).

Happy to test a branch against this corpus — it has 27 entities, a single `DbContext`, ~49 known `IgnoreQueryFilters`/raw-SQL sites already enumerated by hand for the guard above, and a CI-checked generated schema doc, so ground truth for "did it find them all" already exists.

Ayushraj06-bit:
Fixes #3528.

### The gap

An EF Core query site never names its table: `db.Users.Where(...)` reaches `AspNetUsers` through the `DbSet<ApplicationUser> Users` property, and only the property is in source. `Users` has been a node since #3006, but the only edge into it was its own type's `defines`. The C# invocation branch in `walk_calls` keeps a *simple* receiver only, so the chained `db.Users` was dropped, every `db.<DbSet>.X()` raw call arrived at `_resolve_csharp_member_calls` with `receiver=None`, and was skipped. On the reporter's corpus that is why `explain "AppDbContext"` shows 83 declaration edges and not one use.

Minimal repro (three files, `--code-only`): `.CountOrders()` containing `db.Orders.Count()` has **zero** outgoing edges; `Users`/`Orders` have exactly one inbound edge each.

### The fix

`recv.Prop` now resolves the way `recv.Method()` has since #1609:

* **engine.py** — a `member_access_expression` that is not an invocation's callee, with an identifier member and a simple receiver (`recv` / `this` / `base` / `this.field` / `Type`), is recorded as a `raw_calls` entry stamped `is_member_access`, typed by the same scoped receiver table (#2299/#2472). Only entries the resolver can act on are recorded (typed receiver, `this`/`base`, or a type name) since `raw_calls` ride the AST cache. The four receiver-capture arms of the invocation branch became `_csharp_member_receiver`, shared by both sites — behaviour-identical.
* **extract.py** — `_resolve_csharp_member_calls` gains a `property_index` (from `defines` edges whose target is a `.cs` node) beside `method_index`; `_method_on_type_or_bases` takes the index as a parameter. An access goes through the identical typing tiers (`this` → enclosing type, `base` → single base, `Type.` → named, typed → INFERRED), the same `CsharpNameResolver` scoping, base-chain poison and god-node guard, then binds to the property as `uses` / `context=member_access`, EXTRACTED 1.0 or INFERRED 0.8, at the access line. It is never parked (#3152): parked entries are cross-repo *call* candidates.

`is_member_call: True` is kept on access entries so every bare-name pass (shared cross-file, `symbol_resolution`) skips them exactly as it skips a receiver call; the C# resolver branches on `is_member_access`.

After the fix, `explain "Users"` on the repro lists `.FindByEmail()` L20, `.CountOrders()` L25, `.AddUser()` L30 — the "what code reads or writes this table" answer. `ApplicationUser` is one hop further via the declaration's existing `references/generic_arg`.

### Measured

jasontaylordev/CleanArchitecture (76 `.cs`): 956 → 992 edges, node count unchanged. All 36 new edges are `uses`; 12 land on `IApplicationDbContext.TodoItems`/`TodoLists` from the command handlers and validators — every EF query site in the project — the rest are `request.Title` / `_user.Id` DTO reads on in-corpus types. No wrong binding on inspection.

### Tests

`tests/test_csharp_member_access.py`, 19 tests. 11 fail without the change (the DbSet repro; access line/confidence; bare `var x = db.Users` and `foreach`; `this.field.Prop`; `this.Prop` and `Type.Prop` EXTRACTED; `base.Prop`; interface-typed receiver; inherited property; partial-class half; dedup; outer-chain/generic exclusion; raw-entry shape). 8 are guards that must stay green: untyped lambda parameter, out-of-corpus type (and *not* parked — flipping the guard fails this test), ambiguous type, field/method-group members, shadowing local poison, same-named C++ member, no dead raw entry for an untypable receiver.

Full suite: same 30 pre-existing Windows/env failures before and after (byte-identical set), 5580 → 5582 passed. `ruff` clean under the committed selection; `tools.skillgen --check` OK.

### Alternatives considered

* **Gate on ORM vocabulary (`DbSet`, `Model`, `Repository`)** — hard-codes framework names into a language extractor that has none, needs the property's declared type plumbed to the resolver, and loses the general answer (impact analysis on an entity *column* is a property read on the entity, not on the DbSet).
* **A node per call site**, as the issue words it — this graph's model is nodes = declarations and call sites = edges carrying `source_location`; thousands of per-occurrence nodes would fight every id/dedup/explain assumption.
* **A separate resolver module** like `csharp_dispatch.py` — would duplicate ~100 lines of receiver typing.

### How to verify

```
pytest tests/test_csharp_member_access.py -q
graphify extract <any EF Core repo> --code-only --no-cluster && graphify explain "<a DbSet property name>"
```

### Deliberately out of scope (issue items 2–4)

`entity → table` via `[Table]`/`ToTable`; `explain` preferring table vocabulary over a same-named local (a `query`/`explain` ranking change); DDL inside migrations (#2884's territory). Also left alone: making the property node carry its own `generic_arg` to the entity (would double every C# property's type edges; interacts with #3037), enum-member access, property-level interface dispatch, and unqualified `Users.Add()` inside the context itself.

One pre-existing wrong edge surfaced while reproducing, not touched here: a chained-receiver call (`_context.TodoItems.Count()`) falls through to the in-file bare-name match and binds `Count` to the enclosing method's own `.Count()` — a false self-recursion edge. Fixing it removes some currently-correct chained-receiver edges, so it wants its own PR/decision.


Ayushraj06-bit:
Follow-up commit pushed (5f090fa). Nothing changes in the graph output — verified byte-identical nodes/edges on jasontaylordev/CleanArchitecture (110 `.cs`) and jellyfin/jellyfin (2,208 `.cs`) before and after, on both this base and current `v8` (0.9.64).

**Changelog.** The entry was written under `## 0.9.61 (unreleased)`; merged into today's `v8` it lands inside the released 0.9.61 section — the same shape 9335cee had to clean up. Dropped the hunk; happy to leave the wording for the bump.

**Perf** (jellyfin: +26.6k access entries on top of 85.6k member calls, +6,347 `uses` edges):

- `engine.py` — the access branch asked tree-sitter for `node.parent` (twice) to tell an invocation's callee from a property read. `Node.parent` re-descends from the root each call: ~700 ms over the corpus's 137k `member_access_expression` nodes vs ~13 ms for a set lookup. The walk is pre-order, so the invocation branch now records the callee node's id in a per-file `csharp_callee_ids` set and the access branch tests membership. Per-file extraction overhead over base: +1.2–1.6 s → +0.7–0.8 s. A new raw-entry test pins the callee/access split (`db.Save(); db.Log(db.Users); db.Users.Add(db.Orders.First())` → accesses `Orders, Users, Users`; calls `Add, First, Log, Save`).
- `extract.py` — `_resolve_type_name_nid` is memoized per (type name, caller, file); a method reads and calls the same receiver many times over and every access is one more entry that types it (80k → 40k scope-chain walks). The `defines` scan that builds `property_index` is folded into the existing `method` scan — one pass over `all_edges` fills both twin indexes. `_resolve_csharp_member_calls`: 430 ms → 288 ms (base 250 ms).

Also checked against current `v8`: the branch rebases cleanly, full suite there has the same pre-existing Windows/env failure set as `v8` itself, 20/20 in `tests/test_csharp_member_access.py`, `ruff` clean, `tools.skillgen --check` OK. Happy to rebase if you'd prefer that over "Update branch".

PR #3542, opened 2026-09-13
fix(swift): parse an if let binding whose await operand is not a call

### LINKED ISSUE #3540: Swift AST extraction aborts on `if let x = await <non-call>` and `if let x = try await ...`, silently dropping most of the file
## Summary

Swift AST extraction aborts partway through a file when an `if let` optional binding has an `await` operand that is **not a direct function call**. The parser emits a syntax error at that line and silently drops most declarations after it, so the resulting graph is missing a large fraction of a valid file's symbols.

`guard let` with the same `await` operand parses fine, so this looks specific to the `if let` binding rule.

## Repro matrix

Each file is valid Swift (compiles under Swift 6 / Xcode 26.3). Only the binding line differs:

| binding | result |
|---|---|
| `if let r = await fetchRings()` | ok |
| `if let r = try await mint()` | **syntax error** |
| `if let r = await pending` (an `async let` variable) | **syntax error** |
| `if let r = await box.rings` (actor member access) | **syntax error** |
| `guard let r = await pending else { ... }` | ok |

So `await` + direct call is handled; `try await` + call, `await` + bare identifier, and `await` + member access are not.

### Minimal failing file

```swift
import Foundation

struct Rings { let move: Double = 0 }

func fetchRings() async -> Rings? { nil }

func load() async {
    async let pending = fetchRings()
    if let r = await pending {
        print(r.move)
    }
}

struct SentinelType { let marker = 1 }
```

```python
from pathlib import Path

r = extract([Path("repro.swift")], cache_root=Path("."), parallel=False)
print(sorted(n["label"] for n in r["nodes"]))
```

```
  repro.swift (first error at line 9, 6 symbol(s) extracted)
```

Changing line 9 to `guard let r = await pending else { return }` removes the warning.

## Impact on a real corpus

Building a graph over a ~240-file iOS monorepo, two files tripped this and lost most of their symbols:

| file | symbols in graph | approx. symbols in source |
|---|---|---|
| `WidgetShared.swift` (201 lines) | 15 | ~53 |
| `FatSecret.swift` (412 lines) | 59 | ~87 |

The failing lines were `if let rings = await rings {` and `if let fresh = try await requestToken(creds: creds, scope: scope) {` — one of each failing variant above.

The warning is printed, which is good, but it's easy to miss in a large run and the graph is quietly incomplete. Re-extracting those two files with an LLM pass and merging recovered 94 nodes and 155 edges.

## Environment

```
graphifyy          0.9.61
tree-sitter        0.25.2
tree-sitter-swift  0.7.3
Python             3.12.8
OS                 macOS 15.7.2 (x86_64)
```

## Notes

This may well be an upstream `tree-sitter-swift` grammar limitation rather than a graphify bug. If so, two things might still be worth doing here:

1. Make the partial-extraction warning harder to miss — e.g. surface it in `GRAPH_REPORT.md` alongside the corpus stats, not only on stdout during extraction, so a graph built in CI records that it is incomplete.
2. Consider falling back to semantic/LLM extraction for files where the AST pass reports a syntax error, since those files are exactly the ones the structural pass under-covers.

L4XB:
Fixes #3540

## The gap, mapped

tree-sitter-swift 0.7.3 is the **only** published release — PyPI has `0.0.1`, `0.7.2`, `0.7.3` and nothing else — so there is no version to bump to. Measured against the grammar directly:

| binding | parses |
| --- | --- |
| `if let r = await fetchRings()` | ✅ |
| `if let r = try await mint()` | ❌ |
| `if let r = await pending` | ❌ |
| `if let r = await box.rings` | ❌ |
| `if var r = await pending` | ❌ |
| `while let r = await pending` | ❌ |
| `guard let r = await pending else {…}` | ✅ |

All valid Swift 6. `guard` parsing places the gap in the if/while binding rule rather than in `await`.

## The repair

`await` is blanked with **spaces**, not deleted. That is the whole trick: the repaired source is byte-for-byte the same length, so every offset, line and column downstream still addresses what it did. `await` names no symbol, so the graph built from the repaired bytes is the graph of the original file — and the call inside `try await mint()` is still `mint()`.

Gated three ways:

- **Swift only**, on `config.ts_module`.
- **Only when the original parse already errored.** A file that parses is never rewritten.
- **The repaired tree is kept only when it buys something** — no error at all, or a first error that moved further down the file, since a source can carry this grammar gap *and* a real mistake at once.

A match sitting behind a `//` on its line is skipped, so a commented-out binding keeps its text. A block comment or a string literal holding this exact shape is still rewritten — harmless at equal length, and only on a file that already failed to parse. Stated in the code rather than left to be discovered.

## What I verified, and what I could not

**Verified.** Every failing shape above now parses; the file is no longer reported as partially extracted; the `parse_errors` key is gone. Four mutants, each caught by the test written for it:

| mutant | |
| --- | --- |
| the repair is never applied | ✅ caught |
| `await` deleted instead of blanked (offsets shift) | ✅ caught |
| the `//` guard dropped | ✅ caught |
| the rule also sweeps in `guard let` | ✅ caught |

That first mutant initially **passed**, and it was my test's fault: I asserted `assertNotIn("parse_error", result)` and the key the extractor sets is `parse_errors`, plural. The singular spelling is never present, so the assertion passed for any file at all. Corrected, and worth naming since it is the same shape of defect this change is about.

**Could not reproduce.** I built five fixtures aimed at the symbol loss the report measured — a same-named member binding, a binding inside a closure, two failing bindings in one type, declarations nested after the binding, calls inside the `if` body — and in every one the node and edge counts were **identical** with and without the fix. tree-sitter's recovery kept the following declarations each time. So I am not repeating the 53→15 numbers as if I had confirmed them; on your corpus the loss is real and measured, on my constructed files the only observable difference is the parse error and its warning.

That is still worth fixing on its own terms: the warning is a true signal being spent on a file the extractor could read perfectly well, and every rule downstream is walking an ERROR subtree for no reason.

## Suites

- `tests/test_swift_await_optional_binding.py` (new): 4 tests, 14 subtests.
- `-k swift`: 68 passed, 6 skipped.
- Full suite, clean tree vs this branch: **14 failures both ways, identical sets** — terraform grammar, ollama SDK and wheel-build tests that fail in my environment before I touch anything.

Changelog entry added under 0.9.61 (unreleased).


L4XB:
The advisory holds up, and it is fixed at `0ba46e2`. I reproduced it before changing anything:

```
input : let base = "https://api.example.com"; if let x = await fetch(base) { }
output: let base = "https://api.example.com"; if let x = await fetch(base) { }
repaired? False
```

The `//` belongs to the URL, the binding is real code, and the skip left the file unparseable, which is the exact defect this function exists to clear. So "is there a `//` earlier on the line" was the wrong question.

`_swift_line_comment_start` now scans the line tracking string state and returns the offset of the `//` that actually opens a comment. What it does not model is in its docstring rather than left implicit: raw strings (`#"..."#`), multi-line `"""` literals and interpolation. Getting one of those wrong costs nothing in either direction, because reading a comment as code rewrites an `await` inside a comment at equal length on a file that already failed to parse, and reading code as a comment is precisely the behaviour being replaced.

Two cells, and a fixture correction worth mentioning: my first draft used `guard let r = await pending` as one of the repaired shapes, and it failed, because `guard` is deliberately left alone here since the grammar already accepts it. The existing `test_the_repair_leaves_everything_else_alone` says so, and my own test caught me contradicting it.

Five mutations, four killed:

```
killed    the quote-aware scan is gone (any // counts)
killed    a real line comment no longer skips
killed    the string state is never entered
killed    the escape skip is dropped
SURVIVED  the boundary slips (<= instead of <)
```

The survivor is equivalent rather than a gap: `_SWIFT_AWAIT_BINDING` starts at `\b(?:if|while)`, so a match can never begin on the `/` the comment offset points at, and the two can never be equal.

70 passed, 6 skipped across every Swift test. Noting for the record that the finding was marked "agreed by 2 of 2 but NOT verified" and it was right anyway, so the escalation was worth making.

Issue #1423, opened 2026-06-22
bug: validate_graph_path() fallback and callflow_html.py project-root check hardcode "graphify-out", ignoring GRAPHIFY_OUT


prasand:
`GRAPHIFY_OUT` (added for #686) is correctly read in `cache.py`, `__main__.py`, and `watch.py`. Two other spots still compare against the literal string `"graphify-out"` instead of the env var.

`graphify/security.py`, `validate_graph_path()`:

```python
if base is None:
    resolved_hint = Path(path).resolve()
    for candidate in [resolved_hint, *resolved_hint.parents]:
        if candidate.name == "graphify-out":
            base = candidate
            break
    if base is None:
        base = Path("graphify-out").resolve()
```

Should read `os.environ.get("GRAPHIFY_OUT", "graphify-out")` instead of the literal, in both the candidate-name check and the final fallback. Today this is masked because `serve.py` always passes `base` explicitly (fix #170), so the default-`base=None` path isn't exercised by the CLI/MCP server. But any caller or future code path that omits `base` will silently validate against the wrong directory, or raise "Graph base directory does not exist" even when a renamed output dir exists.

`graphify/callflow_html.py`:

```python
project_root = graphify_out.parent if graphify_out.name == "graphify-out" else base
```

Same issue — falls back to `base` (degrades rather than crashes) when `GRAPHIFY_OUT` renames the folder, but should compare against the actual configured name.

**Fix:** both sites should reference the same `_GRAPHIFY_OUT` constant already defined in `cache.py`, rather than re-deriving or hardcoding the literal.

safishamsi:
Fixed on `v8` (ships in the next release). The `GRAPHIFY_OUT` override is now honored consistently across the codebase.

The two sites you named (`d168de9`):
- `validate_graph_path()`'s `base=None` discovery and fallback now use the configured output-dir name instead of the literal `"graphify-out"`.
- `callflow_html.py`'s project-root resolution likewise.
- The name is consolidated into a single `graphify.paths` module (it was duplicated across `__main__`, `cache`, `watch`), so there's one source of truth.

[...]

PR #3546, opened 2026-09-14
▚▚ fix(hooks): resolve the output dir from GRAPHIFY_OUT in the sh gates, not a literal graphify-out/

breken-ai:
With `GRAPHIFY_OUT` set to anything but the default (the worktree / shared-output setup from #686), `graphify hook install` produces a post-checkout hook that never rebuilds: switching branches prints nothing and writes nothing to `~/.cache/graphify-rebuild.log`. The post-commit hook has the mirror problem — a commit that only touches `custom-out/graph.json` (tracked outputs) launches a full rebuild, which is exactly the loop the "only graph artifacts changed" guard was added to prevent.

## Root cause

#1423 moved the Python rebuild bodies onto `os.environ.get('GRAPHIFY_OUT', 'graphify-out')`, but the shell that runs *before* them still hardcodes the directory name:

- `_CHECKOUT_SCRIPT`: `if [ ! -d "graphify-out" ]; then exit 0; fi` — with a renamed output dir this is always true, so the hook exits before `_PYTHON_DETECT` and the launcher ever run.
- `_HOOK_SCRIPT`: `_NON_GRAPH=$(echo "$CHANGED" | grep -v '^graphify-out/' || true)` — paths under the configured dir are not filtered, so a graph-only commit looks like a source change.

The existing test only asserts that the *Python* body mentions `GRAPHIFY_OUT`, so the sh gates in front of it were never pinned.

## Fix

- `graphify/hooks.py` (`_CHECKOUT_SCRIPT`): gate on `${GRAPHIFY_OUT:-graphify-out}` (trailing slash stripped), the same source the rebuild body reads.
- `graphify/hooks.py` (`_HOOK_SCRIPT`): build `_NON_GRAPH` with a POSIX `case` loop keyed on `"$_GFY_OUT"/*` instead of `grep -v`, so the dir name is matched literally rather than as a regex. Each pattern carries the optional leading `(` because bash 3.2 (macOS `/bin/sh`) otherwise mis-parses the pattern's `)` as the end of the `$(...)` substitution — that was caught by the test run, not guessed.

## Tests

`tests/test_hooks.py` — five tests that run the emitted hook scripts under a real `sh` in a scratch git repo, with the pinned python replaced by a stub that passes the `find_spec` probe and swallows the launcher, so the tests observe whether the hook *reaches* the launch line rather than running a rebuild:

- `test_checkout_hook_rebuilds_when_graphify_out_is_renamed` — fails on HEAD (no launch line).
- `test_commit_hook_skips_graph_only_commit_when_graphify_out_is_renamed` — fails on HEAD (launch line present).
- Controls that pass before and after: checkout still no-ops when neither dir exists; the default `graphify-out/` graph-only commit is still skipped; a commit with a source file alongside the renamed dir still launches.

Reverting only `hooks.py` makes the first two fail again.

## Not changed

- `_PYTHON_DETECT`'s second probe still reads `graphify-out/.graphify_python` literally. It is one of five fallbacks for locating an interpreter (the pinned python is tried first), and no writer of that file lives in `graphify/`, so it is left alone rather than widened without a witness.
- An absolute `GRAPHIFY_OUT` that points *inside* the repo (`/repo/custom-out`) still is not relativised for the commit filter; `git diff --name-only` emits repo-relative paths, so that case behaves as it did before this change (a rebuild is launched). Relative names and the default are the documented forms and are what this fixes.
- #3207 (persisting `GRAPHIFY_OUT` to `.graphifyrc` so it survives a fresh shell) is complementary: once the value is available in the hook's environment, these gates now honour it.

---
▚▚ **Shipped by breken** — self-healing software. This one's on us. [breken.ai](https://breken.ai)


safishamsi:
Shipped in **v0.9.62** (now on PyPI: `uv tool install graphifyy==0.9.62`). Your commits were cherry-picked with authorship preserved, so this shows up under your GitHub contributions. Thanks @breken-ai!

PR #3549, opened 2026-09-14
report: count an all-file-node community as thin, not invisible (#3548)

### LINKED ISSUE #3548: GRAPH_REPORT.md counts more communities as "shown" than it renders
# GRAPH_REPORT.md reports more communities as "shown" than it actually renders

**Version:** graphifyy 0.9.48
**File:** `graphify/report.py`

## Summary

The `Summary` line claims more `### Community` sections than the report actually
contains. In my graph it said `69 shown` when only 54 sections existed; the gap is
exactly the number of communities whose nodes are all file nodes.

```
- 84 nodes ... 84 communities (69 shown, 15 thin omitted)
## Communities (84 total, 15 thin omitted)
```

Actual `### Community` sections in that file: **54**, not 69. And 30 communities
were omitted, not 15.

## Cause

`thin_count_summary` (report.py ~L144) counts only communities that have **at
least one** non-file node:

```python
thin_count_summary = sum(
    1 for nodes in communities.values()
    if 0 < sum(1 for n in nodes if not _ifn(G, n)) < min_community_size
)
shown_count = len(communities) - thin_count_summary
```

The `0 <` guard excludes communities whose non-file-node count is zero. But the
rendering loop (~L248) skips those too:

```python
real_nodes = [n for n in nodes if not _ifn(G, n)]
if not real_nodes:
    continue                       # skipped, but NOT counted as thin
if len(real_nodes) < min_community_size:
    continue                       # skipped, and counted as thin
```

So an all-file-node community is never rendered yet still counted in
`shown_count`. `shown_count` overstates by exactly the number of such
communities.

Reproduced on my graph: 84 communities, 54 renderable, 15 thin, 15 all-file-node.
`69 - 54 == 15`.

## Second, smaller issue

`thin_communities` in the Knowledge Gaps section (~L287) hardcodes `< 3`:

```python
if 0 < sum(1 for n in nodes if not _is_file_node(G, n)) < 3
```

but the message it feeds interpolates `min_community_size`:

```python
f"- **{len(thin_communities)} thin communities (<{min_community_size} nodes) omitted ...")
```

Any caller passing a `min_community_size` other than 3 gets a label that
disagrees with the set it describes. Same `0 <` undercount applies here.

## Patch

Both hunks just drop the `0 <` guard so the count matches what the loop skips,
and make the second site use the parameter it advertises.

```diff
--- a/graphify/report.py
+++ b/graphify/report.py
@@ -144,7 +144,7 @@
     thin_count_summary = sum(
         1 for nodes in communities.values()
-        if 0 < sum(1 for n in nodes if not _ifn(G, n)) < min_community_size
+        if sum(1 for n in nodes if not _ifn(G, n)) < min_community_size
     )
     shown_count = len(communities) - thin_count_summary
@@ -287,7 +287,7 @@
     thin_communities = {
         cid: nodes for cid, nodes in communities.items()
-        if 0 < sum(1 for n in nodes if not _is_file_node(G, n)) < 3
+        if sum(1 for n in nodes if not _is_file_node(G, n)) < min_community_size
     }
```

## After the patch

Same graph, regenerated:

```
- 84 nodes ... 84 communities (54 shown, 30 thin omitted)
## Communities (84 total, 30 thin omitted)
```

`shown == number of rendered sections` (54) and `shown + omitted == total`
(54 + 30 == 84). Both invariants now hold.

## Suggested regression check

These two invariants are cheap to assert against a generated report and would
have caught this:

```python
rendered = len(re.findall(r"^### Community \d+ - ", report, re.M))
assert shown_count == rendered
assert shown_count + thin_count_summary == len(communities)
```

## Note on wording (not part of the patch)

The Knowledge Gaps line reads:

```
- **278 isolated node(s):** ...
  These have <=1 connection - possible missing edges or undocumented components.
```

The count is correct for its filter, but the filter also drops file nodes,
concept nodes and `file_type == "rationale"`. In my graph 415 nodes have degree
<= 1, so the prose "these have <=1 connection" reads as a count of all such
nodes when it is a filtered subset. Wording only; I left it alone.

ayushcodes10:
## Summary

Fixes #3548. The issue's own reported repro (`shown_count` overstating the number of rendered `### Community` sections by counting all-file-node communities as shown) turned out to already be fixed on current `v8` by #3148/#2129 — `shown_count` is now computed directly from the render condition rather than `total - thin`. But that earlier fix left a residual gap this issue's own invariant check still catches: the `thin` count kept a `0 <` guard excluding all-file-node (zero-real-node) communities, so those communities — which the render loop skips for exactly the same reason as a genuinely thin one — went uncounted anywhere in the report. `shown + thin` no longer summed to the total community count, and the "N thin communities omitted" figure undercounted what was actually omitted.

Reproduced directly: a 3-community graph (one large, one genuinely thin, one all-file-node) reported `"(1 shown, 1 thin omitted)"` for 3 total communities — `1 + 1 = 2 ≠ 3`. Both counting sites (`thin_count_summary` in the Summary section, `thin_communities` in the Knowledge Gaps section) drop the `0 <` guard, exactly matching the issue's own suggested patch, applied against the current (already-partially-fixed) code shape rather than the older `len(communities) - thin_count_summary` form the issue was written against.

## Test plan

- [x] Three existing tests in `tests/test_report_gap_thresholds.py` pinned the old, undercounting figures — one explicitly, in its own docstring, asserting the exact shape this PR corrects. Updated all three to the now-correct counts, verified by hand against a direct `generate()` call before updating the assertions.
- [x] Added a new test asserting the two invariants the issue itself suggested as a regression check (`shown == rendered`, `shown + thin == total`) at both thresholds already exercised in this file.
- [x] Verified all 4 updated/new tests fail against the pre-fix code and pass with the fix.
- [x] Full suite: `python3 -m pytest -q` — 5487 passed, only the pre-existing unrelated failures (`test_ollama_retry_cap.py` missing `openai` in this env, one flaky timing assertion in `test_ts_import_type_arguments.py`).

Note: the issue's separate "second, smaller issue" (the Knowledge Gaps section's hardcoded `< 3` threshold disagreeing with `min_community_size`) was also already fixed on current `v8` by #3148 — confirmed the code already reads `min_community_size` there, not a hardcoded `3`.


safishamsi:

Issue #3547, opened 2026-09-14
export wiki: generated Community_N.md pages contain literal placeholder/mis-encoded [[wikilinks]]


EvanLee70:
**graphify version:** 0.9.53 (uv tool install)

## Bug 1 — literal placeholder wikilinks left in generated pages

`graphify export wiki` occasionally writes the "example syntax" text for a
wikilink directly into the generated `Community_N.md` page body instead of a
real link, producing targets that can never resolve:

- `Community_410.md:19` → `[[wikilink]]`
- `Community_617.md:20` → `[[Foo Bar Baz]]`
- `Community_762.md:14` → `[[...]]`
- `Community_314.md:27` → `[[link]]`, `[[new_stem]]`
- `Community_399.md:28`, `Community_619.md:14`, `Community_768.md:15` → `[[wikilinks]]`
- `Community_7.md:14` → `[[]]` (empty target)

15 occurrences across 8 generated pages in one export. Looks like a template
string that's meant to be replaced with the real link target isn't being
substituted in some code path.

## Bug 2 — wikilink target is percent-encoded but the actual filename is not

Separately, `graphify export wiki` sometimes writes a link target as a
URL-percent-encoded string, while the file it creates for that same page uses
the literal (non-encoded) characters — so the link never resolves even though
the target page exists:

- Link: `[[CMDS.md_%28System_Philosophy_%26_Context_Guide%29.md]]`
- Actual file written: `CMDS.md_(System_Philosophy_&_Context_Guide).md`

44 links across ~9 generated pages (`index.md` and several `Community_N.md`
pages) in the same export were affected by this — always the same pattern:
`(` → `%28`, `)` → `%29`, `&` → `%26`, `·` → `%C2%B7`, `—` → `%E2%80%94` in the
link text, but not in the actual on-disk filename.

## Impact

Both make `[[...]]` links inside the generated wiki permanently broken (they
never point at an existing file), which is confusing for anyone reading the
generated wiki and breaks any downstream link-integrity checker.

Happy to share the full page list / a minimal repro corpus if useful — this
was found via a custom wiki-lint pass over `graphify export wiki` output, not
hand-inspection, so I have exact file:line data for every occurrence.

PR #3550, opened 2026-09-14
wiki: escape brackets in raw article labels (#3547)

ayushcodes10:
## Summary

Fixes the first of the two bugs in #3547 (literal placeholder/mis-encoded `[[wikilinks]]` in generated `Community_N.md` pages).

**Bug 1 (literal placeholder wikilinks) — fixed here.** `_md_link` already escapes `[`/`]` before a label becomes link text, because a node label is extracted source content and can literally contain a bracketed substring — most plausibly from a doc that discusses or demonstrates wikilink syntax. Three other sites print a label straight into an article body without going through that escaping at all: a community's own title, its "Key Concepts" node listing, and a god node's own title. Printed unescaped, a label like `[[wikilink]]` renders as a real (and always dead — the wiki export never writes bracket-style links) wikilink instead of the plain text it actually is. Pulled the escaping logic out of `_md_link` into a small shared helper (`_escape_md_brackets`) and applied it at all three sites.

**Bug 2 (percent-encoded target vs. raw filename) — already fixed on `v8`.** Traced this directly: `_md_link`'s own docstring already documents removing `quote()`-based percent-encoding as issue #2597, and confirmed `_safe_filename` strips every character that would need encoding from a slug before it's used as either the link target or the on-disk filename — so raw emission and the actual filename are the same string by construction. Verified with the issue's own exact example (`CMDS.md_(System Philosophy & Context Guide).md`): no `%28`/`%29`/`%26` anywhere in the current output, and `tests/test_wiki_link_filename_parity.py` (already in the suite, all passing) pins exactly this invariant for the same character classes the issue reports (`(`, `)`, `&`, em dash, `%`, CJK).

## Test plan

- [x] New `tests/test_wiki_label_bracket_escaping.py`: the helper directly, each of the three fixed print sites individually, and an end-to-end `to_wiki()` export using the issue's own example placeholder text (`[[link]]`, `[[new_stem]]`, `[[]]`) confirmed absent from every generated page.
- [x] Ran the full existing wiki test suite (`test_wiki.py`, `test_wiki_truncation_indicator.py`, `test_wiki_link_filename_parity.py`) — 49 tests, all pass, no regressions.
- [x] Full suite: `python3 -m pytest -q` — 5491 passed, only the pre-existing unrelated failures (`test_ollama_retry_cap.py` missing `openai` in this env, one flaky timing assertion in `test_ts_import_type_arguments.py`).


ayushcodes10:
Thanks for the catch — the formal verifier's finding on `_community_article` was real. `_escape_md_brackets` called `.replace()` directly on its argument, but every caller passes `d.get("label", nid)`, and a node whose dict has no `label` attribute falls back to its own networkx node id, which is not always a string (int and tuple ids are legal, e.g. a graph built with integer node ids). The previous code interpolated that fallback into an f-string, which stringifies anything; wrapping it in `_escape_md_brackets` turned that implicit coercion into a hard requirement the fallback doesn't always meet.

Fixed in 5d79e51 by coercing with `str()` first, restoring the old behavior for a non-string fallback while keeping the escaping for genuine string labels. Added a direct regression test reproducing the reported input shape (int node ids, no `label` attribute) plus a unit test on `_escape_md_brackets` itself.

ayushcodes10:
Addressed the advisory finding on backslash bypassing the escaping in 64bfc3a. Source content can already contain a literal backslash right before a bracket (a doc excerpt showing a regex character class, `\]+`, for example) — escaping the bracket alone turned that into two backslashes then a bare bracket, and CommonMark reads a doubled backslash as one literal backslash, consuming the pair and leaving the bracket unescaped and live again. Now escaping every backslash in the source text first, before the brackets, so the bracket's own escape survives regardless of what preceded it.

Added a regression test with a small simulator of CommonMark's own left-to-right backslash-escape unwinding, so it checks the actual syntactic property (no bracket reachable as bare/unescaped) rather than just comparing output strings.

ayushcodes10:
Following up on the newest review pass's advisory ("Backslash doubling can reactivate escaped Markdown punctuation") — this is unverified/no reproducing input was given, unlike the earlier findings which did include a concrete counterexample. I checked it directly: the fix escapes every backslash in the source text first (single -> double) before escaping brackets, which is provably correct by induction — doubling always leaves an EVEN run of backslashes immediately before any bracket (0, 2, 4, ...), so the bracket-escaping step's one added backslash always makes an odd/pairing count that resolves as an escaped, inert bracket, regardless of how many backslashes preceded it or what surrounds them. Verified empirically against several adversarial cases (runs of 1-4 backslashes before/after/between brackets, mixed content) with a script that simulates CommonMark's left-to-right backslash-escape pairing — no case leaves a bare, live bracket. I believe this is the same concern the earlier finding already flagged and that 64bfc3a already fixed, not a new distinct gap; happy to add more test cases if a specific failing input turns up.

ayushcodes10:
Re: the reproduced formal-verification counterexample on \`_community_article\` (old code: \`# [\`, new code: \`# \[\`) — this is the intended behavior change, not a bug. Escaping a community title that is (or starts with) a literal bracket is exactly what this PR's \`_escape_md_brackets\` is for; the review itself confirms "Looks safe to merge" with no blocking issues.

Took the verifier's suggestion and added a regression test locking this in — \`test_community_article_title_escapes_a_bare_bracket_label\`, calling \`_community_article\` directly with a bare \`"["\` label and asserting the title heading comes out escaped. The existing bracket test in this file (\`test_wiki_link_with_bracketed_label_resolves\`) only covers a bracket inside a *link target* built via \`_md_link\`; this one covers the community's own \`# {label}\` title heading specifically, which is the exact site the verifier's input exercised.

Full suite (5602 passed, no regressions) and \`skillgen --check\` green.

ayushcodes10:
Re: "Backslash doubling can unescape non-bracket Markdown punctuation" — confirmed real. Reproduced directly:

\`\`\`python
_escape_md_brackets(r'\*bold*\ ')
# before: '\\\\*bold*\\\\ '  -- the doubled backslash un-escapes into a bare, live *
# after:  '\\*bold*\\ '      -- unchanged
\`\`\`

The unconditional \`.replace("\\\\", "\\\\\\\\")\` doubled every backslash in the source text, not just ones adjacent to a bracket — so any OTHER pre-existing escape already in that text (\`\\*\` meaning a literal asterisk, for instance) got un-escaped the exact same way the function exists to prevent for brackets. Switched to only doubling a backslash that's immediately followed by \`[\` or \`]\` (\`re.sub(r"\\\\(?=[\\[\\]])", r"\\\\\\\\", ...)\`), which keeps the one case this handling exists for (\`\\]+\`, a regex character class in source) intact while leaving everything else in the text untouched.

Also re: the reproduced formal-verification note on \`_community_article\` in this same review — that's the same confirmed-intended \`# [\` → \`# \[\` change from the earlier round, already covered by the regression test I added in the previous commit.

Full suite (5604 passed, no regressions) and \`skillgen --check\` green.

ayushcodes10:
Re: "\`_escape_md_brackets\` order of replacement may un-escape brackets" — confirmed real, and the high severity call was right. The previous fix only doubled the SINGLE backslash immediately adjacent to a bracket, not a longer run of them. Reproduced directly:

\`\`\`python
_escape_md_brackets('\\\\\\\\' + ']')   # two source backslashes + a close bracket
# before: 4 backslashes then a bare bracket -- CommonMark reads two escaped-backslash
#         pairs (two literal backslashes) with nothing left to escape the bracket,
#         so it un-escapes right back into live link syntax
# after:  5 backslashes then the bracket -- two pairs (two literal backslashes) plus
#         one leftover that correctly escapes the bracket
\`\`\`

Rewrote it as a single regex pass that matches the entire backslash run ahead of a bracket (not just the last character of it), doubles the whole run, then adds one more backslash for the bracket's own escape — verified this now round-trips correctly for runs of 0, 1, 2, 3, and 4 source backslashes before a bracket, by simulating CommonMark's own backslash-pair-collapse rule against the output.

Full suite (5605 passed, no regressions) and \`skillgen --check\` green.

PR #3552, opened 2026-09-14
fix(js): emit one node per exported destructured name (#2604)

### LINKED ISSUE #2604: JS/TS extractor produces a large number of dangling-endpoint edges for external imports + destructuring exports
## Summary
On a 981-file Next.js/TS corpus, `diagnose_extraction` reported **1657 dangling-endpoint edges**. I isolated the root cause into two classes:

## Class 1 (1605/1657, 97%) — external package/builtin imports, by design
Every `imports_from` edge is kept even when its target isn't a valid node (the "clean edges" step in `extract.py` keeps `tgt in valid_ids or edge["relation"] in ("imports","imports_from","re_exports")`, assuming cross-file resolution will happen later). But for external packages (react, next, `prisma/client`, `node:crypto`, ...) that resolution never happens because the package source is outside the scanned corpus. The health-check reports this as if it were corruption, but it's expected/structural.

**Suggestion:** have the health-check count edges whose target uses the extractor's external-reference prefix (`ref_` in this version) separately from genuinely-local unresolved edges — "external (expected)" vs. "local (needs investigation)". Would cut a lot of noise from the signal.

## Class 2 (52/1657, 3%) — destructuring-export node-ID collision, a real limitation
A pattern like `export const { auth, handlers, signIn, signOut } = NextAuth(config)` (NextAuth v5's own recommended boilerplate) collapses to a single combined node (e.g. `stem_handlers_auth_signin_signout`). A file that imports just one of those names individually (`import { auth } from './auth'`) looks for `stem_auth` — which never matches. The same pattern shows up with next-intl's own recommended `createNavigation()` boilerplate.

**Suggestion:** have the destructuring-export pattern emit one node per exported name (all sharing the same source location) instead of one combined node, so single-name imports resolve correctly.

## My own workaround (happy to share the approach)
As a post-build step I added a lightweight "external" stub node for every `ref_*` target — in the extractor's own extraction schema, without patching `site-packages`. Result: dangling edges dropped from 1657 to 52 (all Class 2). It also made the graph more informative (external dependency surface is now visible). Could be a good candidate for a built-in `--stub-external-imports` flag.

ayushcodes10:
## Summary

Fixes #2604's "Class 2" (52/1657, 3%, of the reported dangling-endpoint edges — the smaller, genuinely-fixable half; the larger "Class 1" is a health-check reporting-classification request, not a bug, and is left out of scope here).

`export const { auth, handlers, signIn, signOut } = NextAuth(config)` is NextAuth v5's own documented boilerplate, and next-intl's `createNavigation()` follows the exact same shape. The extractor collapsed the whole destructuring pattern into one combined node — id joining every destructured name (`stem_auth_handlers_signin_signout`), label the literal pattern syntax (`"{ auth, handlers, signIn, signOut }"`) — so a single-name import elsewhere (`import { auth } from './auth'`) could never find a node named just `auth`, and the import edge (plus any call resolved through it) dangled.

Each shorthand and renamed property in the pattern now emits its own node, sharing the statement's line — exactly what the export would have produced had it been written as N separate `export const auth = ...` statements. A renamed property (`handlers: h`) exports under its key (`handlers`), since that's what an importer binds by, never the local alias. A rest element (`...rest`) or a default-valued entry doesn't correspond to one discrete exported name and is skipped rather than guessed at.

**Caught and fixed a real regression during testing, not after:** the split is gated to only fire on an actually-exported statement. An un-exported destructure of the identical shape — `const { doWork } = require('./lib')`, the ordinary CommonJS import pattern — is a local import binding, not a module export. My first pass didn't have this gate, and running the existing test suite immediately caught it breaking `test_cross_file_call_promoted_to_extracted_with_import_evidence`: giving the import binding a bare-named node in the *importing* file made it collide with the real definition it merely imports, turning a clean single-candidate cross-file resolution into a false "ambiguous name" and silently losing the real `calls` edge. Fixed before ever landing a broken commit — see the commit history for how this played out.

## Test plan

- [x] New `tests/test_js_destructured_export.py`, 5 tests: the core per-name split; cross-file import + call resolution through a split node; renamed-property key-not-alias naming; rest-pattern skip without breaking siblings; and a dedicated regression guard reproducing the exact require()-collision interaction the fix had to be gated against.
- [x] Verified the 4 split-behavior tests fail against the pre-fix code and pass with the fix; the regression-guard test passes both before and after (it pins behavior that already worked and that the gate keeps working).
- [x] Ran the full existing JS/TS + `test_extract.py` suite (416 tests) — all pass, no regressions, including the specific test my first ungated attempt broke.
- [x] Full suite: `python3 -m pytest -q` — 5491 passed, only the pre-existing unrelated failures (`test_ollama_retry_cap.py` missing `openai` in this env, one flaky timing assertion in `test_ts_import_type_arguments.py`).


safishamsi:

PR #3553, opened 2026-09-14
fix(cache): recover semantic cache groups with malformed source_file by basename

### LINKED ISSUE #2973: save_semantic_cache silently drops entries when the adaptive-retry split path produces a malformed source_file (weak/local backends) — distinct from #1991
**Environment:** graphifyy 0.9.48, `--backend openai` pointing at a self-hosted llama.cpp server (`gemma-4-26b-a4b` via OpenAI-compatible endpoint), `--max-concurrency 1 --api-timeout 300`.

**Not the #1991 scenario:** our invocation doesn't pass `--out` at all — `root` and `cache_root` correctly resolve to `<corpus>` and `<corpus>/graphify-out` respectively (confirmed by reading `save_semantic_cache`'s docstring and the `_checkpoint_chunk` call site in `llm.py`, which already does the root/cache_root split #1991 introduced). So this isn't a root-anchoring misconfiguration.

**What actually happens:** during a long incremental `extract` run, whenever a chunk is too large for the model's practical context and `_extract_with_adaptive_retry` bisects it (`llm.py`, "chunk of N exceeded context... splitting in half and retrying"), the resulting sub-chunk's extracted nodes sometimes carry a `source_file` string that doesn't resolve to a real file. `save_semantic_cache` then hits its `skipped_not_file == len(by_file)` branch and discards the whole group, with only the generic "paths do not resolve..." warning pointing at the wrong (already-fixed) issue number. Log excerpt:

```
[graphify] chunk of 360 exceeded context at depth 0 (BadRequestError); splitting in half and retrying
[graphify extract] chunk 219/1205 done
.../graphify/llm.py:2595: RuntimeWarning: save_semantic_cache: all 20 source_file group(s) were
skipped because their paths do not resolve to real files. This usually means root is anchored
to the wrong directory (see the closed root-cause issue above).
```

Net effect for us: an "incremental" nightly extraction over a ~1200-chunk vault was silently re-attempting large already-processed chunks every run, because their split-path results never made it into the cache. Ran 14h42m stuck at 18% before we killed it.

**Suspected root cause:** the model's self-reported `source_file` field (trusted as-is, no deterministic cross-check against the real dispatched file list) drifts specifically under the split/retry path, likely because a weaker/local model re-prompted with a reduced file subset loses track of which of the original chunk's files a given node came from.

**Fix we tested and are running locally** (in `save_semantic_cache`, `cache.py`): when `resolved_source_path(fpath)` fails to resolve to a real file, and `allowed_paths` is available (it already is at the only real caller, `_checkpoint_chunk`, via `allowed_source_files=[unit_path(item) for item in chunk]`), fall back to an unambiguous basename match against that known-good, already-dispatched file list. Only recovers when exactly one candidate matches; any ambiguous or genuinely-bogus path still falls through to the existing skip behavior unchanged.

```python
if not p.is_file() and allowed_paths is not None:
    candidates = [ap for ap in allowed_paths if ap.name == Path(fpath).name]
    if len(candidates) == 1:
        cache_path = candidates[0]
        p = candidates[0]
if p.is_file():
    ...
```

Verified locally with three synthetic cases: (1) a malformed-but-basename-unique path recovers correctly, (2) an ambiguous basename (two allowed files sharing a name) correctly stays skipped, (3) normal well-formed paths with no allowlist are unaffected — no regression. Happy to open a PR with this and a regression test if useful.

ayushcodes10:
Fixes #2973.

## Summary

`save_semantic_cache` groups nodes/edges/hyperedges by `source_file`, then silently skips any group whose path does not resolve to a real file. The reporter found that during a long incremental run against a weak or local backend, the adaptive-retry split path (`llm.py` bisecting a chunk that overflowed the model's context) sometimes reports a `source_file` for a sub-chunk that never resolves at all, so that group's nodes and edges vanish from the cache on every run — an "incremental" extraction kept re-processing the same already-extracted chunks.

This recovers such a group by an unambiguous basename match against `allowed_source_files` (the known-good, already-dispatched file list already threaded through from `_checkpoint_chunk`), matching the fix the reporter tested and verified locally. Two things had to change for it to actually work end to end:

- The write **key** for a recovered group is redirected to the real file (`_recover_group_path`, used consistently by both the write loop and `group_skipped`'s dangling-reference pruning, so the two never disagree on whether a group was written).
- Each item **inside** a recovered group also has its own `source_file` field corrected to match. Without this, `load_cached`'s `_semantic_entry_matches_path` integrity check rejects the entry on the very next read (it verifies every item's `source_file` matches the path being loaded), so a naive fix would report the group as "saved" while producing an entry that can never actually be read back — silently reproducing the same loss it's meant to fix. This was caught by writing a direct save-then-load round-trip test rather than only asserting the `saved` count.

An ambiguous basename (two allowed files sharing a name) or a call with no `allowed_source_files` at all still falls through to the exact previous skip behavior, so unscoped callers are byte-identical.

## Changes

- `graphify/cache.py`: add `_recover_group_path`, wire it into `group_skipped` and the write loop, and correct a recovered group's own `source_file` fields before writing.
- `tests/test_semantic_cache_basename_recovery.py` (new): basename recovery + save/load round trip, ambiguous basename stays skipped, unscoped calls unaffected, dangling-edge pruning stays consistent for a recovered group.

## Testing

- `python -m pytest tests/test_semantic_cache_basename_recovery.py -q` — 5 passed
- Full suite: `python -m pytest -q` — 5489 passed, 68 skipped (pre-existing unrelated failures excluded: `tests/test_ollama_retry_cap.py` missing the optional `openai` module, and `test_ts_normalizer_scales_linearly_on_large_files` / `test_hyperedge_convex_hull_js_is_geometrically_sound`, both known environment-specific flakes unrelated to this change)


safishamsi:

PR #3555, opened 2026-09-14
fix(watch): route graph.json replace through the atomic fallback on both incremental write sites

### LINKED ISSUE #2689: update/cluster-only fail with os.replace WinError 5 on VMware HGFS shared folders (read-then-replace over existing graph.json)
**graphify:** 0.9.41
**OS:** Windows Server 2019, Python 3.14
**Filesystem:** repo on a VMware Shared Folders drive (`\\vmware-host\Shared Folders`, HGFS)

### Summary
On a repo hosted on a VMware HGFS shared drive, `graphify update .` (and any command
that rewrites an existing `graph.json`, e.g. `cluster-only`, `label`) fails at the final
atomic write:

    [graphify watch] Rebuild failed: [WinError 5] Access is denied: 'graphify-out\.graph.tmp.json' -> 'graphify-out\graph.json'

The graph is computed correctly (`.graph.tmp.json` is written); only the `os.replace()`
onto the existing `graph.json` fails. Because the run aborts there, `GRAPH_REPORT.md`,
`graph.html`, and `manifest.json` are never written. The un-advanced manifest means every
later run re-detects the same delta and fails at the same spot, so incremental updates are
permanently blocked on this filesystem.

Looks related to #287 (also `os.replace` WinError 5) but still reproduces on 0.9.41, so
that fix does not cover this path.

### Root cause
On HGFS, `os.replace(tmp, dst)` raises `PermissionError(13)` when `dst` was **read earlier
in the same process** before being replaced. graphify's incremental write reads the existing
`graph.json` (merge / shrink-guard), then `os.replace()`s the temp over it, which HGFS denies.
First-ever builds succeed because there is no existing `graph.json` to read.

### Minimal reproduction (no graphify required)
On a VMware HGFS shared drive:

    import os
    open('dst.json', 'w').write('x' * 1_300_000)
    _ = open('dst.json').read()            # read before replace (the trigger)
    open('.tmp.json', 'w').write('y' * 1_500_000)
    os.replace('.tmp.json', 'dst.json')    # PermissionError(13, 'Access is denied') on HGFS

Controls (same machine, same code):
- Remove the intervening `read()` and `os.replace` succeeds on the same HGFS share.
- `os.remove('dst.json'); os.rename('.tmp.json', 'dst.json')` (delete-then-rename) succeeds even with the read.
- Run the exact snippet on a local NTFS path (`C:\...`) and it succeeds.

So it is specific to (read-then-replace) on HGFS, not a general Windows or handle-leak issue.

### Suggested fix
Fall back to delete-then-rename when `os.replace` raises `PermissionError` / WinError 5:

    try:
        os.replace(tmp, dst)
    except PermissionError:
        os.remove(dst)
        os.rename(tmp, dst)

Delete-then-rename is non-atomic, so keep `os.replace` as primary and use this only as a fallback.

ayushcodes10:
Fixes #2689.

## Summary

On a VMware HGFS shared folder, `os.replace(src, dst)` raises `PermissionError` (`[WinError 5] Access is denied`) when `dst` was read earlier in the same process before being replaced — reproduced directly in the issue with a minimal snippet. `graphify update` and `cluster-only` do exactly that: read the existing `graph.json` for the merge/shrink-guard comparison, then replace the temp file onto it. On this filesystem the rebuild aborts at that point, `GRAPH_REPORT.md`/`graph.html`/`manifest.json` never get written, and the un-advanced manifest means every later run re-detects the same delta and fails at the same spot — incremental updates are permanently blocked.

`graphify/paths.py` already has `os_replace_with_fallback` (added in #3508, my own earlier PR) — it falls back to a copy-through-temp-and-rename-aside sequence on exactly this `PermissionError`/WinError-5 case, and `export.py`'s clustered-path write already routes through it. `watch.py`'s two incremental-rebuild write sites (the `no_cluster` fast path and the default clustered path) were both still using a plain `Path.replace()`, bypassing the fallback entirely — that's the actual gap this issue is hitting.

## Changes

- `graphify/watch.py`: both `graph_tmp.replace(existing_graph)` call sites now call `os_replace_with_fallback(graph_tmp, existing_graph)` instead.
- `tests/test_watch.py`: two new regression tests (one per call site), each monkeypatching `os.replace` to always raise the simulated HGFS `PermissionError` and confirming the incremental rebuild still succeeds and lands the new content.

## Testing

- `python -m pytest tests/test_watch.py -k permission_error -q` — 2 passed
- Full suite: `python -m pytest -q` — 5487 passed, 68 skipped (pre-existing unrelated failures excluded: `tests/test_ollama_retry_cap.py` missing the optional `openai` module, and `test_ts_normalizer_scales_linearly_on_large_files` / `test_hyperedge_convex_hull_js_is_geometrically_sound`, both known environment-specific flakes unrelated to this change)


safishamsi:

PR #3556, opened 2026-09-14
fix(hooks): install a post-merge hook to heal symbols the union merge driver resurrects

### LINKED ISSUE #2418: Post-merge phantom nodes are never healed: union merge-driver resurrects deleted symbols, and the scoped post-commit rebuild cannot prune them
# Post-merge phantom nodes are never healed: union merge-driver resurrects deleted symbols, and the scoped post-commit rebuild cannot prune them

**Version:** graphifyy 0.9.30 (installed from the `v8` checkout) · macOS 15.6 (darwin 24.6) · Python 3.14.6

## Summary

The `graphify merge-driver` union-merges `graph.json`, which by design discards
deletions made on either side of a merge. That is a reasonable trade to avoid
conflict markers, and a full `graphify update .` does correctly clean the
resurrected nodes.

The problem is that **nothing in the installed hook set ever runs that full
update**. The post-commit hook rebuilds only the files listed in
`GRAPHIFY_CHANGED`, so a symbol resurrected by a merge inside a file that is not
touched again persists in `graph.json` indefinitely — long after the code is
gone. The graph reports it with an `EXTRACTED` confidence tag and a real
`file:line`, so there is no signal that it is stale.

This is distinct from #2406 (which is about *losing* real edges on the
incremental path). This one is the opposite direction: *retaining* nodes that no
longer exist.

## Reproduction

Minimal synthetic repo, all commands verbatim:

```bash
mkdir proj && cd proj && git init -q .
mkdir -p src
cat > src/util.py <<'EOF'
def helper_alpha():
    return 1

def helper_doomed():
    return helper_alpha()
EOF
cat > src/main.py <<'EOF'
from src.util import helper_alpha, helper_doomed

def run():
    return helper_alpha() + helper_doomed()
EOF
git add -A && git commit -qm base

graphify update .            # baseline: contains src_util_helper_doomed
cp graphify-out/graph.json ../graph_BASE.json

# branch deletes helper_doomed
git checkout -qb branch-a
cat > src/util.py <<'EOF'
def helper_alpha():
    return 1
EOF
cat > src/main.py <<'EOF'
from src.util import helper_alpha

def run():
    return helper_alpha()
EOF
git add -A && git commit -qm "delete helper_doomed"
graphify update .            # correctly pruned -> node is gone
cp graphify-out/graph.json ../graph_A_deleted.json

# simulate the merge git would perform on graph.json
cp ../graph_BASE.json ../graph_CURRENT.json
graphify merge-driver ../graph_BASE.json ../graph_CURRENT.json ../graph_A_deleted.json
```

**Step 1 — the merge resurrects the deleted node:**

```
$ python3 -c "import json;d=json.load(open('../graph_CURRENT.json'));
print([n['id'] for n in d['nodes'] if 'doomed' in n['id'].lower()])"
['src_util_helper_doomed']
```

**Step 2 — the *scoped* rebuild the post-commit hook performs does not clean it.**
Put the merged graph back, commit a change to an unrelated file, and run exactly
what `_REBUILD_BODY_COMMIT` runs:

```python
from pathlib import Path
from graphify.watch import _rebuild_code
_rebuild_code(Path('.'), changed_paths=[Path('src/unrelated.py')], force=False)
```

```
[graphify watch] Rebuilt: 8 nodes, 7 edges, 3 communities
helper_doomed still present: True      # <-- phantom survives
```

**Step 3 — only an unscoped rebuild heals it:**

```
$ graphify update .
[graphify watch] Rebuilt: 7 nodes, 7 edges, 2 communities
helper_doomed present: False           # <-- cleaned
```

## Expected vs actual

**Expected:** after a merge, the graph converges back to the code within a
commit or two of normal hook-driven activity.

**Actual:** it never converges. `src/util.py` is never in `GRAPHIFY_CHANGED`
again unless someone happens to edit it, so the phantom node is permanent for
practical purposes. It is indistinguishable from a real node in
`graphify explain` / `query` / `affected` output.

## Root cause

Three behaviours that are each individually reasonable compose into the bug:

1. `graphify/cli.py:2047` — the merge driver is `_nx.compose(G_cur, G_oth)`, a
   pure union. `%O` (the merge base) is accepted as an argument but never
   loaded, so there is no way to distinguish "deleted on one side" from "added
   on the other".
2. `graphify/hooks.py:110` and `graphify/hooks.py:141`
   (`_REBUILD_BODY_COMMIT`) — the post-commit hook reads `GRAPHIFY_CHANGED` and
   passes `changed_paths=changed` into `_rebuild_code`, scoping re-extraction to
   the commit's files.
3. `graphify/hooks.py:298-300` — the post-commit hook deliberately exits early
   during merge/rebase/cherry-pick states, so the merge itself never triggers a
   rebuild either.

The prune path (`_stale_graph_sources` / `_prune_graph_json_sources`) only drops
nodes whose `source_file` is gone or newly excluded. Here the file still exists;
only a symbol inside it was deleted. That is repaired by re-extraction replacing
the file's nodes — which the scoped rebuild never performs for that file.

## Suggested fix

Any one of these would close it:

1. **Install a `post-merge` hook** that runs an unscoped `_rebuild_code(root)`
   (no `changed_paths`). This is the smallest change and matches the existing
   `post-commit` / `post-checkout` pattern in `graphify/hooks.py`. `post-merge`
   fires after a successful merge, which is exactly when the union damage lands.
2. **Have the merge driver mark the graph dirty** — write a
   `graphify-out/.needs_full_rebuild` stamp when it runs, and have the next
   post-commit rebuild honour it by dropping `changed_paths` for that one run.
3. **Use `%O` in the merge driver** to compute a real three-way merge: a node
   present in base and in one side but absent from the other was *deleted*, not
   *added*, and should not be resurrected. Most correct, most work.

(1) and (2) are complementary and both cheap; (3) is the principled fix.

## Documentation gap

Independently of the code fix, `README.md`'s Team setup section says
`graphify hook install` "also sets up a git merge driver so `graph.json` is never
left with conflict markers — two devs committing in parallel get their graphs
union-merged automatically." That is accurate but reads as though the merge case
is fully handled. It would help to state that a merge inflates the graph and that
an unscoped `graphify update .` is needed afterwards, until the above is fixed.

ayushcodes10:
Fixes #2418.

## Summary

`graphify hook install` registers a git merge driver for `graph.json` that union-merges both sides of a merge (`nx.compose`), which by design discards deletions made on either side — a reasonable trade to avoid conflict markers. The problem: nothing in the installed hook set ever runs the unscoped rebuild that would heal the resulting damage. The post-commit hook only rebuilds the files listed in `GRAPHIFY_CHANGED`, so a symbol resurrected inside a file that isn't touched again persists in `graph.json` indefinitely, reported at `EXTRACTED` confidence with a real `file:line` — indistinguishable from a real node in `explain`/`query`/`affected` output.

The issue reporter traced this precisely (exact `hooks.py` line references) and proposed three fix options, explicitly calling one "the smallest change": install a `post-merge` hook (git's own hook for exactly this trigger point — fires after a merge completes cleanly) that runs an unscoped rebuild, matching the existing `post-checkout` pattern (branch switches already get a full, not scoped, rebuild for the same reason: arbitrary files can change with no single "changed_paths" list to scope to).

## Changes

- `graphify/hooks.py`: new `_MERGE_SCRIPT`, built from the same shared building blocks as the other two hooks (`_WORKTREE_GUARD`, `_PYTHON_DETECT`, `_detached_launch`), reusing `_REBUILD_BODY_CHECKOUT` verbatim since it already does exactly what's needed — an unscoped `_rebuild_code(_root, force=_force)`. Wired into `install()`, `uninstall()`, and `status()` alongside the existing two hooks.
- `README.md` / `graphify/__main__.py`: updated the two places that listed which hooks `hook install` writes.
- `tests/test_hooks.py`: dedicated install/uninstall/status tests for the new hook, plus adding it to the shared `_HOOK_SCRIPTS` parametrized list, which extends 8 existing generic hook-script tests (no nohup, cross-platform detach, Windows worker limiting, GIT_DIR reuse, `GRAPHIFY_SKIP_HOOK` handling, shell-quote safety, valid-Python payloads, linked-worktree skip) to it for free.
- `tests/test_watch.py`: an end-to-end test proving the actual property the fix exists for — a node shaped like a genuine merge-driver resurrection (matching a real AST node's `_origin`/`source_location` fields) survives a rebuild scoped to an unrelated file (what the commit hook would run) and is dropped by an unscoped rebuild (what the new merge hook runs). Worth noting: an earlier draft of this test used a phantom node with no `_origin`/`source_location` fields and appeared to pass on the first try, but for the wrong reason — such a node classifies as semantic-tier rather than AST-tier, and the AST-only rebuild preserves semantic-tier nodes unconditionally regardless of scope, so it would have "healed" even against a no-op hook. Shaping the phantom to match a real AST node closes that gap.

## Testing

- `python -m pytest tests/test_hooks.py tests/test_watch.py -q` — all pass
- Full suite: `python -m pytest -q` — 5498 passed, 68 skipped (pre-existing unrelated failures excluded: `tests/test_ollama_retry_cap.py` missing the optional `openai` module, and `test_ts_normalizer_scales_linearly_on_large_files` / `test_hyperedge_convex_hull_js_is_geometrically_sound`, both known environment-specific flakes unrelated to this change)


dmitryvostryakov:
Confirmed this fixes a real operational gap for us. With graphifyy 0.9.63, `hooks.install()` still installs only `post-commit` and `post-checkout`; our normal repository advancement is `git pull` after merged PRs, so the graph drifted by hundreds of commits despite hooks being installed. We had to add a local `post-merge` hook as a workaround.

The unscoped rebuild proposed here is the right behavior for post-merge; it also makes the freshness marker meaningful in pull-based workflows.

Issue #746, opened 2026-05-06
graphify update + export wiki silently wipes wiki/ when .graphify_analysis.json is missing (data loss)


boo-code:
## Summary

Running `graphify update .` followed by `graphify export wiki` on a project that does **not** have a `.graphify_analysis.json` sidecar deletes every existing wiki article and writes only god-node summaries (~10 files), regardless of how many community articles previously existed.

In our project this caused a **silent loss of 2,519 wiki articles** when an automated pre-push hook ran the two commands in sequence. Detection happened only after the regression had already merged.

Tested on `graphifyy==0.7.5` (Python package), Linux x86_64, default branch v7.

## Reproduction

Starting from a project that has a populated `graphify-out/` (e.g. produced by an earlier `graphify extract .` run with thousands of wiki articles):

```bash
# Pre-condition
ls graphify-out/wiki/*.md | wc -l        # → 2519 (or any large count)
test -f graphify-out/.graphify_analysis.json && echo "sidecar present" || echo "sidecar MISSING"
# Make the sidecar 'missing' (simulating a clean checkout / a project that
# only ran `update`, never `extract`):
rm -f graphify-out/.graphify_analysis.json

# Reproduction
graphify update .
graphify export wiki

# Post-condition
ls graphify-out/wiki/*.md | wc -l        # → 11   ← every community article gone
```

Same outcome if the sidecar never existed at all (e.g. project initialized via `graphify update` only, without ever running the LLM-driven `extract`).

## Root cause

Three independent design choices combine into a destructive interaction:

1. **`graphify update`** (`graphify/watch.py::_rebuild_code`) writes `graph.json`, `GRAPH_REPORT.md`, and `.graphify_root` — but **never** writes `.graphify_analysis.json`.
2. **`graphify cluster-only`** (`graphify/__main__.py:1496-1551`) re-clusters the graph but **also never** writes `.graphify_analysis.json`. Only `graphify extract` (`__main__.py:2154-2188`) writes that sidecar, and `extract` requires LLM API access (expensive; not appropriate for a pre-push hook or watch loop).
3. **`graphify export wiki`** (`__main__.py:1800-1808`) reads `communities` from `.graphify_analysis.json` and **silently falls back to `{}` if the file is absent**:

   ```python
   communities: dict[int, list[str]] = {}
   if analysis_path.exists():
       _an = json.loads(analysis_path.read_text(encoding="utf-8"))
[...]

PR #3557, opened 2026-09-14
fix(export): detect a stale .graphify_analysis.json and reconstruct from graph.json

### LINKED ISSUE #2386: export wiki/html prefer a stale .graphify_analysis.json over the fresh per-node partition in graph.json — the existing fallback is gated on presence, not agreement
**Version:** graphify 0.9.22 · macOS · repo: ~8.4k code files, 59,748 nodes / 175,324 edges

## Summary

`graphify export wiki` and `graphify export html` take their community partition from `.graphify_analysis.json`. That file is written only by `extract` (`cli.py:3396`) and `cluster-only` / `label` (`cli.py:1668`) — **never by `update` / the watch rebuild path** (`grep graphify_analysis watch.py` → no matches).

So after any number of `graphify update` runs, `graph.json` and `GRAPH_REPORT.md` advance while the analysis sidecar stays frozen at the last full `extract`. The exports keep reading the fossil, and the only signal is a `dropped N stale node ID(s)` line that reads like housekeeping.

In our repo the sidecar was **9 days and one re-partitioning era** behind:

| | value |
|---|---|
| `graph.json` | 59,748 nodes · **2,359** communities |
| `GRAPH_REPORT.md` | 2,359 communities (agrees with graph.json) |
| `.graphify_analysis.json` | **6,243** communities · 111,726 node ids · 9 days old |
| `graph.html` emitted | `aggregated: 6243 community nodes` ← the fossil partition |
| wiki emitted | 1,940 articles, `dropped 56369 stale node ID(s)` |

**Measured impact: 4,394 of 59,748 nodes (7.4%) were in no community of the sidecar, so they got no wiki article and were absent from the aggregated HTML.** `wiki.py:244` intersects the sidecar's communities with the live graph — it only ever subtracts, so that set grows with every commit and never recovers.

Nothing warns. The report says 2,359 communities, the HTML says 6,243, and both are written by the same tool in the same session.

## The fix is already in the code — gated on presence instead of agreement

`cli.py:2192-2211` already reconstructs communities from `graph.json`'s per-node `community` attribute, and the comment names this exact bug:

```python
# Fallback: graph.json carries the per-node community as a node attribute
# (`to_json` writes it on every node). The analysis sidecar is the
# canonical source — but the post-commit / watch rebuild path doesn't
# regenerate it, ...
if not communities:          # <-- fires only when the sidecar is MISSING
    reconstructed = ...      #     a sidecar that EXISTS but is stale wins
```

That fallback appears to come from #746 (`update` + `export wiki` wipes `wiki/` when the sidecar is **missing**) and #1610 (`cluster-only` not persisting it → `export html` reports "Single community"). Both are the *absent* case. The *stale* case looks identical from the outside — degraded artifacts, no error — but takes the other branch: when the sidecar exists and is wrong, the fresh data sitting in `graph.json`, already loaded into `G` two lines above, is ignored.

### Suggested fixes

**A. Minimal, at `cli.py:2201`.** Both datasets are already in memory, so the comparison is free: reconstruct unconditionally and prefer it (or at least warn) when the sidecar's node-id set disagrees with the graph's. This also repairs every graph already on disk, without a re-cluster.

⚠️ One caveat if you take A: the fallback reconstructs `communities` only. `cohesion` and `gods` are read from the same sidecar at `cli.py:2185-2187`, and in the `else` branch they become `{}` / `[]`. `gods_data` self-heals (`cli.py:2259-2260` recomputes via `_god_nodes(G)`), but `cohesion` does not — it flows into `wiki.py:90`, which prints `cohesion N.NN` in each community article's metadata, and into the HTML export at `cli.py:2242`. So "reconstruct and prefer" would swap one silent degradation for a smaller one unless it also recomputes cohesion; `score_all(G, communities)` at `cli.py:1521` is the same call the normal path already makes.

**B. Root cause, at `watch.py:1235/1263`.** `watch` already computes `communities = cluster(G)` and writes them into `graph.json` via `to_json(...)`. Writing `.graphify_analysis.json` from those same communities costs nothing extra and keeps the sidecar canonical.

Repairing it externally with `cluster-only` is *not* a good workaround, for two reasons already tracked here: clustering is not reproducible run-to-run (#1667 — we measured 2,363 vs 2,362 communities over a byte-identical `graph.json`), and re-clustering churns curated labels (#2298, #1653). In our repair run, 51% of 2,359 curated labels were renamed by hub.

A cheap consistency check is possible even outside the process: `GRAPH_REPORT.md` states the community count and `.graphify_analysis.json` holds the partition, so a wrapper can detect the desync **without loading the 101 MB `graph.json`** (we now do exactly this in our refresh script). Inside graphify it is cheaper still — zero I/O, since `G` and `communities` are both already loaded at `cli.py:2201`.

### Workaround, for anyone hitting this

Deleting `.graphify_analysis.json` and re-running the exports is enough to get a correct partition: it forces the `if not communities:` branch, so the partition is rebuilt from `graph.json`'s per-node attribute, labels are preserved and god-nodes are recomputed. The only loss is the cohesion metric in the wiki articles, until the next real `cluster-only` / `extract`.

Two things that are *not* obvious if you try to repair with `cluster-only` instead:

- The labels' `.sig` sidecar is also written only by `cluster-only` / `label`, so after a `watch` run the saved labels have no matching signature and every community is treated as changed and renamed by its hub.
- That degradation is invisible to a placeholder count: the no-LLM path renames by hub (`cli.py:1595-1597`), never to `Community N`, so `placeholders == 0` throughout. Weighted by nodes rather than by community count, ~110 renames covered **37.9% of the graph** — re-clustering perturbs the largest communities most, so 8 of the top 10 lost their prose name.

ayushcodes10:
Fixes #2386.

## Summary

`graphify export wiki/html/obsidian/svg/graphml/neo4j/falkordb` take their community partition from `.graphify_analysis.json`, written only by `extract` and `cluster-only`/`label` — never by `update` or the watch rebuild path. So after any number of `update` runs, `graph.json` (and its per-node `community` attribute) keeps advancing while the analysis sidecar freezes at the last full `extract`.

An existing fallback (`cli.py`, added for #746/#1610) already reconstructs communities from `graph.json`'s per-node attribute, but it's gated on the sidecar being **missing** (`if not communities:`). A sidecar that **exists but is stale** looks identical from the outside — no error, no warning — and takes the other branch, so the fresh data sitting in `graph.json` two lines above is ignored. The issue reporter measured this directly on an 8.4k-file repo: the sidecar was 9 days and one re-partitioning era behind (2,359 vs 6,243 communities), 7.4% of nodes got no wiki article, and the only visible signal was a `dropped N stale node ID(s)` line that reads like routine housekeeping. A second commenter confirmed the same mechanism independently on a different repo with concrete before/after data (34 orphaned nodes, 1,149 mismatched community assignments).

This is "Option A" from the issue, which the reporter already scoped precisely down to the exact caveat it needs to avoid trading one silent degradation for a smaller one (stale `cohesion` surviving into wiki article metadata even after communities are fixed).

## Changes

- `graphify/cli.py`: the per-node reconstruction now always runs (not just when the sidecar is empty). When the sidecar exists, its covered node-id set is compared against the fresh reconstruction's — cheap, since both are already in memory, and deliberately node-id-based rather than community-id-based since community ids can renumber run to run even for the same partition (#1667). A mismatch means the sidecar is stale: the fresh data wins, `cohesion` is recomputed via `score_all` (the same call the normal path already makes), `gods_data` is reset so the existing self-heal at the wiki export site recomputes it, and a warning is printed naming the exact command (`cluster-only`) that refreshes the sidecar for real.
- `tests/test_cli_export.py`: four new tests — the html export path healing and warning on a stale sidecar, the wiki export path specifically recomputing cohesion (an impossible sentinel cohesion value must not survive into the output), and a negative control confirming an up-to-date sidecar never trips the new path.

## Testing

- `python -m pytest tests/test_cli_export.py -k "stale or still_matches" -q` — 4 passed
- Full suite: `python -m pytest -q` — 5488 passed, 68 skipped (pre-existing unrelated failures excluded: `tests/test_ollama_retry_cap.py` missing the optional `openai` module, and `test_ts_normalizer_scales_linearly_on_large_files` / `test_hyperedge_convex_hull_js_is_geometrically_sound`, both known environment-specific flakes unrelated to this change)
- Manually reproduced end to end: extracted a small corpus, hand-corrupted the sidecar's node-id set to simulate staleness, confirmed `export html` prints the new warning and renders correctly from the fresh data.


ayushcodes10:
Re: "Stale detection uses only node-set equality, missing membership/partition changes that keep the same node set" — confirmed real, pushed a fix.

The flat node-id-set comparison missed a whole class of staleness: a merge of two communities, a split of one, or a single node moving from one community to another can all leave the overall node set exactly unchanged while still describing a different clustering. Now comparing each side's set of community blocks (`{frozenset(nodes) for nodes in communities.values()}`) instead of the flattened node-id set — this catches a partition change even when the flat node set stays identical, while still correctly ignoring a pure community-id renumbering of the same partition (the existing #1667 guard this logic already had to respect).

Added a regression test that merges the fixture's two communities into one (same node set, different partition) and confirms the stale path now fires.

Full suite (5597 passed, no regressions) and `skillgen --check` both green.

ayushcodes10:
Re: "Stale sidecar detection resets gods_data unconditionally, dropping valid god node data" — checked this directly.

`god_nodes(G)` ranks purely by graph degree (`graphify/analyze.py:109`), independent of the community partition entirely — it takes only `G`, no `communities` argument. So in the current codebase this was harmless in practice: the only reader of `gods_data` is the `wiki` subcommand, which already has its own `if not gods_data: gods_data = _god_nodes(G)` fallback a few lines further down, so clearing it to `[]` just meant that fallback recomputed it fresh with an identical result either way.

Still a real fragility worth fixing though — it silently depended on that one consumer's own guard rather than being correct on its own terms, and would silently drop real data for any future reader that doesn't carry the same fallback. Pushed a commit that recomputes it directly at the stale detection site instead, mirroring how `cohesion` is already recomputed right next to it.

I didn't add a new regression test for this specific piece: since both the old (clear + rely on the wiki fallback) and new (recompute directly) code paths produce byte identical `gods_data` for the only current consumer, a test exercising the wiki export can't actually distinguish the two behaviors — it would just re-cover the existing fallback test. Flagging that honestly rather than padding in a test that wouldn't actually catch a regression here.

Full suite (5597 passed, no regressions) and `skillgen --check` both green.

safishamsi:

PR #3558, opened 2026-09-14
fix(cluster): exclude self-loops from cohesion_score so the ratio sta…fix(cluster): exclude self-loops from cohesion_score so the ratio stays in 0..1

jordanalexanderp18-rgb:
## Problem

`cohesion_score()` counts self-loops in the numerator (`number_of_edges()`) but not in
the denominator (`n * (n - 1) / 2`, distinct node pairs). Recursive `calls` self-edges
are deliberately preserved by `build_from_json` (`build.py:1294`), so they reach this
code in ordinary graphs and push the ratio past the 0..1 range the docstring defines.

```python
G = nx.Graph()
G.add_edge("f", "g", relation="calls")
G.add_edge("f", "f", relation="calls")   # recursion
cohesion_score(G, ["f", "g"])            # 2.0 before, 1.0 after
```

`report.py:289` prints the score unclamped, so `GRAPH_REPORT.md` could show `Cohesion: 2.00`.

## Change

One line in `cohesion_score()`: subtract `nx.number_of_selfloops(subgraph)` from the edge
count, with a comment explaining why the self-loops are there in the first place.

Clamping the output was the other option, but that would hide a miscount rather than fix
it, and it would leave the inflated sub-1.0 scores wrong.

## Tests

Three cases added to `tests/test_cluster.py`, beside the existing cohesion tests:

- `test_cohesion_score_ignores_self_loops` — the 2.0 case
- `test_cohesion_score_range_with_self_loops` — the 0..1 invariant with every node recursive
- `test_cohesion_score_self_loops_only_is_zero` — self-loops alone are not a connection

All three fail on `main` and pass with the fix. The existing
`test_cohesion_score_range` already asserts the invariant but runs on a fixture with no
self-loops, so it could not catch this.

## Behaviour change

Communities containing recursive functions now report lower, correct cohesion.
`cluster()` reads this score at `cluster.py:320` to decide re-splitting
(`< 0.05`, communities of 50+ nodes), so a community whose inflated score sat just above
that threshold could now be split. Checked against a real 2048-node graph with 38
self-loops: 15 of 104 communities report corrected scores, none crosses the re-split
threshold, clustering output unchanged.


safishamsi:
Shipped in **v0.9.62** (now on PyPI: `uv tool install graphifyy==0.9.62`). Your commits were cherry-picked with authorship preserved, so this shows up under your GitHub contributions. Thanks @jordanalexanderp18-rgb!

PR #3559, opened 2026-09-14
fix(dedup): make id prefix reconstruction Unicode aware

### LINKED ISSUE #3352: Node identity derived from source_file silently drops the defining page's own node
## Summary

When two files mention the same entity, graphify keeps the node from whichever file was extracted first and drops the other:

```
[graphify] WARNING: node 'X' is minted by two different files — keeping 'X' from 'A.md',
dropping 'X' from 'B.md'. ... the dropped node is lost.
```

The practical effect is that **the page that actually defines a concept can lose its own node**, and the surviving node is attributed to a file that merely mentions the concept in passing.

## Concrete case

Corpus: an Obsidian vault of ~60 markdown notes, `graphify extract wiki/ --mode deep`.

| | |
|---|---|
| Page that **defines** the concept | `concepts/작업 단위 폴더 + README 진입점 컨벤션.md` (filename == concept name) |
| Page that only **mentions** it | `concepts/LLM Schema - CLAUDE.md를 가이드로 활용하기.md` |
| Node's `source_file` in `graph.json` | the *mentioning* page |
| Nodes attributed to the defining page | **none** |

The concept itself is in the graph with 4 edges — only the ownership is wrong.

## Why it matters

The documented navigation pattern is "read `GRAPH_REPORT.md` → identify relevant nodes → fetch only those source files". When `source_file` points at a page that merely mentions the concept, that workflow **routes the reader to the wrong file**. The authoritative page is invisible to the graph.

## The documented workaround does not apply here

The warning suggests:

> To keep them distinct, run 'graphify extract' per subfolder and merge with 'graphify merge-graphs'.

Both files above live in the **same** folder (`concepts/`), so splitting by subfolder cannot separate them.

## Post-hoc correction is also blocked

Rewriting `source_file` in `graph.json` to point at the defining page does not work, because node ids are re-derived from it — `build.py::_semantic_id_remap`:

> the new id is computed from `source_file` in code, never trusted from the fragment's own `id` string

So rewriting `source_file` causes ids to be re-derived, collide with other nodes, and get silently merged — reproducing the very problem. The resulting graph shrank 146 → 145 nodes, which then tripped the shrink guard in `export.py`:

```
[graphify] WARNING: new graph has 145 nodes but existing graph.json has 146 (net -1).
Refusing to overwrite.
graph.json NOT written: ... GRAPH_REPORT.md ... left untouched.
```

Both `graphify label` and `graphify cluster-only` refuse to write, so `GRAPH_REPORT.md` can no longer be regenerated and goes stale. `GRAPHIFY_FORCE=1` is honored by `extract` but not by `label` / `cluster-only`, so there is no CLI escape hatch (the warning text says "pass force=True", which is only reachable from the Python API).

## Suggestion

When resolving a mint collision, prefer the file whose name matches the entity name as the owner — or more generally, prefer a "defining" page over a "mentioning" one — rather than first-extracted-wins. Failing that, exposing a `--force` flag on `label` / `cluster-only` would at least let users recover a stale report.

## Environment

- graphify 0.9.53 (`pipx install graphifyy`, plus `pipx inject graphifyy openai` for the gemini backend)
- macOS, Python 3.14.7
- Backend: gemini
- Corpus: ~60 markdown docs, no code files

ayushcodes10:
Fixes #3352.

## Summary

The reported symptom: when two markdown pages mention the same entity, the page that actually *defines* the concept can lose its own node to a page that merely mentions it in passing. The reporter's concrete case is a page named exactly after the concept (Korean filename) losing its node to an unrelated page that just references it.

`dedup.py` already has the mechanism the issue's own "Suggestion" section asks for: `_collision_rank` ranks a node whose `source_file` *defines* its id (`_defines_id`) ahead of one that merely references it, specifically so the defining page wins. That mechanism is correct and already tested for ASCII paths. The bug is one level down: `_id_prefixes` — which `_defines_id` uses to reconstruct what id a node extracted from a given `source_file` would legitimately mint — reimplemented id slugification with its own ASCII-only regex (`re.compile(r"[^a-z0-9]+")`) instead of the canonical `normalize_id()` every real extractor actually mints an id with. For a path segment made of non-Latin characters (Korean, CJK, Cyrillic, ...), the ASCII regex strips it entirely instead of preserving it, so the reconstructed prefix never matches the id that was actually minted for that file. `_defines_id` then never recognizes the defining page as defining anything, and `_collision_rank` falls through to plain arrival order — exactly the symptom reported.

Verified directly against current code before writing the fix: `_id_prefixes('concepts/작업 단위 폴더 + README 진입점 컨벤션.md')` returned `{'concepts_readme', 'readme'}` (every Korean character dropped), while the id an extractor actually mints for that path (via `_file_stem` + `make_id`) is `concepts_작업_단위_폴더_readme_진입점_컨벤션` — completely different from any prefix the old code could ever reconstruct.

## Changes

- `graphify/dedup.py`: `_id_prefixes` now slugifies each path segment with `normalize_id` (from `graphify.ids`, the same Unicode-aware casefold→NFKC→`[^\w]+` recipe every extractor uses) instead of the local ASCII-only regex, which is now unused and removed.
- `tests/test_dedup.py`: five new tests — `_id_prefixes` preserving non-Latin segments, `_defines_id` recognizing a non-Latin defining path, an end-to-end `deduplicate_entities` test reproducing the issue's own Korean defining/referencing pair in both arrival orders, and a negative control confirming an ordinary ASCII path's reconstructed prefixes are byte-for-byte unchanged.

## Testing

- `python -m pytest tests/test_dedup.py -k "korean or non_latin or ascii_path_unchanged" -q` — 5 passed
- Full suite: `python -m pytest -q` — 5490 passed, 68 skipped (pre-existing unrelated failures excluded: `tests/test_ollama_retry_cap.py` missing the optional `openai` module, and `test_ts_normalizer_scales_linearly_on_large_files` / `test_hyperedge_convex_hull_js_is_geometrically_sound`, both known environment-specific flakes unrelated to this change)


safishamsi:

PR #3562, opened 2026-09-14
feat(markdown): resolve code-span mentions into references edges

### LINKED ISSUE #3561: Markdown extractor: code-span references promised in the docstring are not implemented
`graphify/extractors/markdown.py` (`extract_markdown`, on `v8` at fe66389 and in the 0.9.58 wheel)
documents this edge:

    heading --references--> other node (when backtick `Name` matches a known pattern)

Nothing implements it. The extractor emits `page` and `heading` nodes, `contains` edges, and
`references` edges only for inline, reference-style and wiki links to other markdown files. A code
span in a heading or body line never produces an edge, so an AST build has no document-to-code edge
at all.

Reproduction (0.9.58):

    d=$(mktemp -d)
    printf 'class Widget:\n    pass\n' > $d/widget.py
    printf '# Doc\n\n## `Widget`\n\nThe `Widget` class.\n' > $d/doc.md
    python -c "from pathlib import Path; from graphify.extract import extract; \
    r = extract([Path('$d/widget.py'), Path('$d/doc.md')]); \
    print(sum(1 for e in r['edges'] if e.get('relation') == 'references'))"

prints `0`.

Proposed fix, implemented and tested on a fork branch (PR to follow):

1. `extract_markdown` collects inline code spans outside fences and frontmatter as `raw_calls`
   tagged `language: "markdown"`, owned by the enclosing heading (the page node before the first
   heading). Forms: bare `Name`, `Name()`, dotted `pkg.Name`, and `path.py::Name(::Name)*`.
2. A `markdown_mentions` language resolver (registered through `register_language_resolver`, so it
   runs after the id-remap passes) resolves them. The `path::Symbol` form matches the cited file
   doc-relative, then exact, then as a unique path suffix, and walks `contains`/`method` edges to the
   symbol: EXTRACTED 1.0. A bare name resolves only when exactly one callable code node carries the
   label: INFERRED 0.95. Ambiguous names, builtins, missing files and non-callable matches emit
   nothing.
3. The shared cross-file call pass skips markdown raw_calls, so prose never emits `calls` edges and
   document nodes stay out of call resolution.
4. The docstring is corrected.

Results on the branch: the six new tests pass; the existing suite is unchanged (5,569 passed, 14
skipped; the eight failures, four `test_backup_*` in `tests/test_export.py` and four
`test_pathless_postgres_extract_initializes_empty_detection` cases in `tests/test_extract_cli.py`,
fail identically on clean `v8`). A dry run over a 1,581-document, 11,297-symbol Python repo produced 3,532 document-to-code
`references` edges (3,409 INFERRED, 123 EXTRACTED) from 10,809 mentions; 985 ambiguous names were
skipped. The only recurring false positive found was a prose word (`source`) matching a same-named
test fixture function.

AstroMined:
Fixes #3561.

The markdown extractor's docstring promised a heading --references--> node edge for backtick names
but nothing implemented it, so an AST build produced no document-to-code edge.

`extract_markdown` now reports inline code spans (outside fences and frontmatter) as `raw_calls`
tagged `language: "markdown"`, owned by the enclosing heading or the page node. A new
`graphify/markdown_resolution.py` registers a `markdown_mentions` language resolver that turns them
into `references` edges: `path::Symbol[::member]` resolves through the cited file's symbols
(EXTRACTED 1.0); a bare or dotted name resolves only to a unique callable code node (INFERRED 0.95).
Ambiguous, builtin, missing-file and non-callable mentions emit nothing. The shared cross-file call
pass skips markdown raw_calls so prose never produces `calls` edges. The stale docstring line is
replaced with the actual behaviour.

Tests: `tests/test_markdown_code_spans.py` covers span classification, raw_call ownership and line
pinning, end-to-end resolution (INFERRED, EXTRACTED, doc-relative path, ambiguous, missing file, no
calls from document nodes, no dangling edges), and cache round-tripping. `uv run pytest` and
`uv run --frozen ruff check .` are green apart from four `test_backup_*` cases in `test_export.py`
and four `test_pathless_postgres_*` cases in `test_extract_cli.py`, which fail identically on clean
`v8`.


AstroMined:
Thanks for the review. All three advisory findings reproduce and are fixed in a0325ee:
- Dotted spans now keep their qualifiers and a bare-name match must carry every qualifier as a parent label or a path segment of its file, so `README.md`, `pyproject.toml` and `time.sleep` no longer match a same-named callable; the file early-exit also checks `DOC_EXTENSIONS`.
- `_match_cited_file` strips `./` and `../` prefixes as prefixes (not characters), so `.github/...` cites can match by suffix, and a root-level file can match the suffix form.
- An explicit `./` or `../` cite resolves relative to the document only; it never falls through to the suffix rule, so an escaping `..` cannot land on an unrelated copy.

Tests cover each case. The coupling notes describe the new resolver's fan-out and are expected for a language resolver registered through `register_language_resolver`.

Proving the fixes through `graphify update` also turned up a fourth defect, fixed in 66e6625: `_reconcile_markdown_links` treats every AST-tier `references` edge owned by a re-parsed Markdown file as an authored `[link](file)` and prunes any pair it did not author, so a rebuild over an existing graph dropped every mention edge and only a build from scratch kept them. A mention targets a code symbol rather than a file representative, so the reconcile now leaves those edges to the AST ownership rule. A test runs a mention through a full, a no-change, a document-side and a code-side rebuild.

safishamsi:
Shipped in **v0.9.62** (now on PyPI: `uv tool install graphifyy==0.9.62`). Your commits were cherry-picked with authorship preserved, so this shows up under your GitHub contributions. Thanks @AstroMined!

Issue #2195, opened 2026-07-25
Absolute-path node ids on 0.9.25 for in-root relative import targets (residual of #1899 Variant B; ghost duplicates + dangling endpoints)


timomitchel:
## Summary

On **0.9.25**, node ids for the targets of relative imports that cross directories are still built from the absolute resolved path, while the same node's `source_file` is correctly relativized. This is the same class as #1789 / #1899 (Variant B: relativized `source_file` but an id derived from the absolute path) and #1961, all closed, but with a trigger those did not cover: **in-root relative import targets in an Astro plus TypeScript project** (`../../lib/content`, `../../config`, `../styles/global.css`).

Because the importing file also produces a correctly named node for the same target, the result is a **ghost duplicate**: `src_lib_content` and `users_<name>_..._<repo>_src_lib_content` both exist in the graph.

## Environment

- graphify 0.9.25 (installed via `uv tool`, Python 3.14, macOS)
- AST-only extraction (no LLM), undirected build
- Project: Astro site. Pages under `src/pages/**` and `src/layouts/Base.astro` import `../../lib/content`, `../../config`, and `../styles/global.css`.

## Observed

Three import-target nodes carry an id derived from the absolute path, with a correct repo-relative `source_file`:

```
id:          users_<name>_startups_<...>_<repo>_src_lib_content
source_file: src/lib/content
label:       ../../lib/content

id:          users_<name>_..._<repo>_src_config
source_file: src/config

id:          users_<name>_..._<repo>_src_styles_global_css
source_file: src/styles/global.css
```

For `src_lib_content` and `src_config`, a correctly named node with the same repo-relative id **already exists** in the graph (the file's own node), so the absolute-id node is a pure duplicate. The absolute-id nodes are also the endpoints of `imports_from` edges, so on the post-build graph they show up as dangling endpoints (as in #1961).

## Reproduction

Minimal shape: a file that imports another in-root file by a relative path that ascends at least one directory.

```
src/lib/content.ts
src/pages/work/index.astro   ->  import { ... } from '../../lib/content'
```

Then:

```python
from pathlib import Path
from graphify.extract import extract
res = extract([Path('src/pages/work/index.astro'), Path('src/lib/content.ts')], cache_root=Path('.'))
print([n['id'] for n in res['nodes'] if 'lib_content' in n['id']])
# -> ['src_lib_content', 'users_<name>_..._src_lib_content']   # duplicate, second is absolute
```

Notes:
[...]

PR #3566, opened 2026-09-15
fix(elixir): resolve alias/import edges onto their defmodule node

### LINKED ISSUE #3565: Elixir: every alias/import edge is dropped at build — definition ids carry a file-stem prefix, reference ids do not
**Version:** 0.9.61 (`fe66389`) · macOS · Python 3.14

## Summary

`extract_elixir` mints a module **definition** id as `_make_id(file_stem, module_name)` but mints the **target** of an `alias`/`import`/`require`/`use` edge as `_make_id(module_name)`, with no stem. The two can never be equal, so `build_from_json`'s node-set prune drops the edge and Elixir's entire module-dependency layer disappears from the graph.

The failure is silent in the way that matters: extraction reports no error, both endpoints exist as nodes, the file parses, the node count looks right. Only the edge is missing. A module aliased by a dozen callers reports degree 2, and community detection degenerates to roughly one community per file.

`extract_elixir` already assumes something downstream will reconcile this — its own dangling-edge filter deliberately exempts `relation == "imports"` so these edges survive the extractor. Nothing downstream ever resolves them.

## Reproduction

```bash
mkdir -p repro/lib/my_app
cat > repro/lib/my_app/repo.ex <<'EOF'
defmodule MyApp.Repo do
  def insert(x), do: x
end
EOF
cat > repro/lib/my_app/user.ex <<'EOF'
defmodule MyApp.User do
  alias MyApp.Repo
  def save(x), do: Repo.insert(x)
end
EOF
graphify update repro
python -c "
import json, collections
g = json.load(open('repro/graphify-out/graph.json'))
print(collections.Counter(e['relation'] for e in g['links']))"
```

No `imports` edge reaches `graph.json`. The extractor emitted one.

## The two conflicting ids

```
lib/my_app/repo.ex              NODE  lib_my_app_repo_myapp_repo   label 'MyApp.Repo'   # elixir.py:119
lib/my_app/user.ex  alias ...   EDGE  imports -> myapp_repo                             # elixir.py:167
```

`lib_my_app_repo_myapp_repo` != `myapp_repo`.

## Measured impact

A 980-file Phoenix/Ecto application (`lib/` only), graphify 0.9.61:

| | before | after fix |
|---|---|---|
| `imports` edges emitted | 4,292 | 4,292 |
| surviving the build | **0 (0.0%)** | 3,531 (82.3%) |
| cross-file edge share | **8.1%** | 27.4% |
| Louvain communities / single-file | 562 / 445 | 113 / 49 |

Whole repo (2,470 `.ex`/`.exs`): 8,570 emitted, **0** surviving → 6,390 (74.6%). The residual unresolved are genuine externals (`Logger`, `Ecto.Query`, deps) with no node in the corpus, which correctly stay dangling.

Source contains 2,805 real `alias`/`import` statements; `graph.json` contained 28 `imports` edges.

## Why the per-file extractor cannot fix it

An Elixir `alias` names a BEAM module, not a path, so `extract_elixir` — which sees one file at a time — cannot know which file defines it. Guessing a path from `Macro.underscore` conventions breaks on umbrella apps, `:elixirc_paths`, acronym modules (`HTTPClient` → `http_client`), and multi-module files. So the `target_file` approach from #2195 (which resolves *path-valued* specifiers on disk) does not apply here.

This is the same shape as Kotlin #2526, whose own docstring describes the identical symptom: *"That target dangles (node ids carry a file-stem prefix), so build pruned every Kotlin import."*

## Note: no cross-module call edges either

Independent of this bug, `extract_elixir` emits **no** cross-module call edge in any style — `Bar.run(x)` after `alias`, and fully-qualified `MyApp.Foo.Bar.run(x)`, both produce nothing. Fixing the ids therefore restores the module **dependency** layer, not a call graph. Worth stating so the improvement is not over-read.

## The same asymmetry exists in other extractors

While confirming the above, the pattern `definition = _make_id(stem, name)` / `reference = _make_id(name)` was reproduced with 0 surviving cross-file edges in: **Rust** (`use`), **Fortran**, **Julia**, **Apex**, **Verilog**, **Razor**, **Blade**, **Dart**, **Scala**, **Groovy**, **Zig**, and **Pascal** (different root cause, same symptom). Partial loss in **Go**, **PowerShell**, **Ruby**, **OCaml**.

Two failure modes: *dangling* (target matches nothing, pruned) and *sourced ghost* (the reference site calls `add_node`, minting a duplicate stub attributed to the referencing file). Python, TS/JS, Svelte/Vue/Astro, PHP, C, C++, Obj-C, C#, Kotlin, Swift, Lua, SQL, Common Lisp, Bash verified OK.

Happy to split that into a separate tracking issue — flagging it here only because the Elixir fix is one instance of it.

## Proposed fix

PR to follow: stamp the written module name as `metadata.target_fqn` and repoint the edge in a corpus-wide resolver, guarded on a single candidate so nothing is fabricated — mirroring `_resolve_kotlin_import_targets`. Definition ids are left untouched, so existing graphs keep their node ids.

pavanetti:
Fixes #3565.

## Root cause

`extract_elixir` mints module **definitions** as `_make_id(file_stem, module_name)` — the corpus-wide invariant documented on `_file_stem` (#1504), which keeps same-named symbols in different files from collapsing — but emitted the **reference** side of `alias`/`import`/`require`/`use` as `_make_id(module_name)`, with no stem. The two ids can never be equal, so `build_from_json`'s `src not in node_set or tgt not in node_set` prune dropped every Elixir alias/import edge.

An Elixir alias names a BEAM module, not a path, so the per-file extractor cannot resolve it, and `#2195`'s `target_file` probing does not apply. This is the same shape as Kotlin #2526 and is fixed the same way.

## The fix

- `extractors/elixir.py` stamps the written name as `metadata.target_fqn`, and marks **top-level** `defmodule` nodes with `_elixir_module`.
- `extract.py` adds `_resolve_elixir_import_targets`, a corpus-wide pass mirroring `_resolve_kotlin_import_targets`: a name defined **exactly once** is repointed onto that node; zero or several candidates are left dangling, so no stub node is ever fabricated and confidence stays `EXTRACTED` (the label is the alias text as written, so the match is exact).

**Definition ids are unchanged** — existing graphs keep their node ids. This is why the resolver approach is preferable to making the definition id name-derived, which would churn every Elixir module id in every stored graph.

## Measured

980-file Phoenix/Ecto app, `lib/` only:

| | before | after |
|---|---|---|
| `imports` emitted / surviving build | 4,292 / **0 (0.0%)** | 4,292 / 3,531 (82.3%) |
| cross-file edge share | 8.1% | 27.4% |
| duplicate node ids | 0 | **0** |
| `contains` edges | 1,012 | **1,012** |

Whole repo (2,470 `.ex`/`.exs`): 0 → 6,390 / 8,570 (74.6%). Unresolved remainder is genuine externals (`Logger`, `Ecto.Query`, deps).

Suite: **5,635 passed, 1 failed, 14 skipped**. The failure is `test_export.py::test_hyperedge_convex_hull_js_is_geometrically_sound`, which fails identically on clean `HEAD` — pre-existing and environmental, not from this change.

## Three guards, each with a regression test

Adversarial review found these; each is covered in `tests/test_elixir_import_ids.py`:

1. **Self-import must not destroy `contains`.** The `__using__`/`quote do import MyApp.Context end` idiom makes a file alias a module it also defines. Retargeting lands the edge on the same `(source, target)` pair as `file --contains--> module`; the built graph is a plain non-multi `Graph`, so one relation silently overwrites the other. Measured at **17 lost `contains` edges** before the guard. Self-imports carry no cross-file information and are now skipped.

2. **Nested `defmodule` must not capture a foreign alias.** `extract_elixir` labels a nested module with the name as *written* — `defmodule Supervisor` inside `MyApp.Application` is labelled `Supervisor`, not `MyApp.Application.Supervisor`. Indexing it would let it absorb `use Supervisor` (the stdlib module) from unrelated files and assert a false `EXTRACTED` edge. Only top-level defmodules are indexed, via `_elixir_module`. That marker also excludes any same-labelled node a semantic tier might add, which would otherwise make the name ambiguous and silently suppress resolution.

3. **Resolution must survive an incremental rebuild.** The early resolver slot was handed `all_nodes`, which on a partial re-extract holds only the changed batch — so every alias into an unchanged file would dangle and be pruned, decaying the graph back toward the bug on exactly the `graphify watch` / hook path. Worse, it bypassed the single-candidate guard: a corpus-wide-ambiguous name can look unique within the changed subset and get an arbitrary target. Now indexes `resolution_nodes`, which includes the unchanged-corpus context (#2406) — matching that variable's own documented scope ("the member-call resolvers run by `run_language_resolvers`"). Inert for Kotlin, which indexes off `per_file`.

## Risks / compatibility

- **`_AST_CACHE_SCHEMA` 2 → 3.** The extractor's output changed (adds `metadata.target_fqn`), so a cache written by an earlier build would keep serving pre-fix results and make this change **silently inert** — verified: fix + pre-fix cache reproduces 0% resolution with no warning. The bump sweeps those entries. This also traps anyone re-measuring against a warm cache.
- **No node ids change**, so stored graphs, manifests and incremental diffs stay valid. 0 duplicate ids on every corpus tested, including a purpose-built torture corpus (umbrella apps with same-named modules, `.exs`, `defimpl`, nested and reopened modules).
- **Ambiguous module names stay unresolved by design.** A name defined in two files is left dangling rather than guessed.
- **`graphify update --no-cluster`** writes the raw extraction, so its `graph.json` still carries some dangling `imports` (7 post-fix vs 16 pre-fix on the repro). Pre-existing behaviour, reduced but not eliminated.

## Known limitation

This restores the module **dependency** layer only. `extract_elixir` emits no cross-module *call* edge in any style — neither `Bar.run(x)` after an alias nor fully-qualified `MyApp.Foo.Bar.run(x)` produces one — so the graph gains module-to-module imports, not a call graph. Noted in #3565.

Also noted in #3565: the same definition/reference id asymmetry reproduces with zero surviving cross-file edges in ~12 other extractors (Rust, Fortran, Julia, Apex, Verilog, Razor, Blade, Dart, Scala, Groovy, Zig, Pascal). Out of scope here; happy to follow up.


safishamsi:
Thanks for this, @pavanetti — your root-cause diagnosis was spot on (alias/import targets were minted stemless, so `build_from_json` pruned every cross-file module edge). We went with the overlapping #3603 because it resolves the same bug through node `source_file`+`label` and registers in the tail resolver, which keeps working on the incremental `graphify update` / watch path without extra marker plumbing. Your two guards (nested-module capture, same-file `contains` clobber) were genuinely right, so I ported both onto #3603 on merge. Closing as superseded by #3603 (shipped in v0.9.63). Really appreciate the careful work.

PR #3571, opened 2026-09-15
feat(terraform): resolve local module sources and directory topology

vstepko:
Terraform module calls currently ignore literal `source` paths, so a monorepo's environment → application → base topology is disconnected even when every `.tf` file is scanned. This adds deterministic local module-source resolution: calls link to directory module nodes, which contain the scanned `.tf` files in that directory.

- Preserve literal source values and source-line provenance; resolve `./` and `../` relative to the calling module, within the scan root and scanned corpus.
- Scope Terraform block IDs by the full directory path, with portable IDs and a hash to distinguish paths that normalize to the same text.
- Recompute topology after cached per-file extraction. Incremental Terraform changes refresh the scanned `.tf` corpus using cached syntax, covering source changes, newly added targets, deletions, and exclusions.
- Preserve a directory anchor when its provenance file is deleted: discard an obsolete same-ID record before merge dedup can select it over the fresh surviving anchor.
- Invalidate the old AST cache schema and document the one-time `graphify update .` migration.

Scope is literal local sources in `.tf` files. Remote sources, source-expression evaluation, `.tf.json`, output-to-implementation resolution, and evaluated `count`/`for_each` instances remain outside this change. Resolution never downloads modules or recursively scans unselected paths.

Validation: 542 tests passed across Terraform, resolver registry, cache, build/merge/prune, incremental CLI, watcher, and dedup suites; 3 watcher-event tests skipped because watchdog is not installed. New integration tests exercise CLI and watcher updates both with and without clustering. Ruff and `git diff --check` passed. `graphify update .` completed with no LLM calls (optional parsers for a few unrelated fixtures are not installed).


safishamsi:
Shipped in **v0.9.62** (now on PyPI: `uv tool install graphifyy==0.9.62`). Your commits were cherry-picked with authorship preserved, so this shows up under your GitHub contributions. Thanks @vstepko!

PR #3575, opened 2026-09-15
fix(export): give a committed output directory its diff attributes

### LINKED ISSUE #3413: Github diff too large for commited graphify-out content
I found useful to setup our .gitattributes file for 4 commited graphify files as:
```
graphify-out/graph.json merge=graphify
graphify-out/** linguist-generated
# Skip line-level diffing for the auto-regenerated graph artifacts: git treats each version as
# opaque binary for diff purposes instead of running the LCS algorithm over it. graph.json is
# pretty-printed and can show 1M+ changed lines between rebuilds (node ordering shifts), which
# blows past GitHub's diff-generation time budget - the PR /files and /compare endpoints then
# fail with HTTP 422 ("Sorry, this diff is taking too long to generate.") and the PR page/API
# silently report 0 changed files across the whole PR instead of surfacing the error.
# linguist-generated above only collapses the *rendered* diff in the UI; it still has to be
# computed first, so it alone does not fix this. The other three files here are smaller but
# regenerate the same way on every commit, so they get the same treatment pre-emptively.
graphify-out/graph.json -diff
graphify-out/manifest.json -diff
graphify-out/GRAPH_REPORT.md -diff
graphify-out/.graphify_labels.json -diff
```

though the GitHub's files changed tab shows 0 - as the diff got too large 

Any solution how to setup Github diff properly - to ignore the graphify-out content?

Image

L4XB:
Closes #3413.

A committed `graphify-out/` can cost a pull request its whole diff. `graph.json` is pretty-printed and node ordering shifts between rebuilds, so a rebuild presents a very large changed-line count. GitHub gives up computing a diff that size and answers 422, and the pull request then reports **zero changed files for every file in it**, not only for the artifacts. The Files changed tab goes empty and the API agrees with it, which is what the reporter is looking at in the screenshot.

The reporter had already worked out the right incantation by hand. This ships it, so the next person does not have to.

## The change

`to_json` writes a `.gitattributes` into the output directory when it has none:

```
* linguist-generated
graph.json -diff
manifest.json -diff
GRAPH_REPORT.md -diff
.graphify_labels.json -diff
```

`-diff` is the part that matters. It makes git treat these as opaque, so the diff is never computed. `linguist-generated` only collapses the rendered result, which is too late: the computation has already happened and already timed out. The repo already carries `graphify-out/**/*.html linguist-vendored=true` in its own root `.gitattributes`, so marking these artifacts is an established position here, not a new one.

Three boundaries, each with a test:

- **Only the output directory.** `--graph` can name any path, and dropping a file into a directory the user chose for one artifact would be writing somewhere we were not invited. The write is gated on the parent directory being `GRAPHIFY_OUT_NAME`.
- **Never overwrite.** An existing `.gitattributes` is left exactly as it is, so a user who has tuned theirs keeps it.
- **Never fail the build.** A read-only output directory costs the attributes file, not the graph, which is the same stance `backup_if_protected` takes.

`GRAPHIFY_NO_GITATTRIBUTES=1` turns it off, mirroring `GRAPHIFY_NO_BACKUP`.

#### Test

Five cells in `tests/test_export.py`. Four fail on `v8`:

```
FAILED test_to_json_gives_the_output_directory_diff_attributes
  FileNotFoundError: .../graphify-out/.gitattributes
FAILED test_existing_gitattributes_is_never_overwritten
  ImportError: cannot import name 'ensure_output_gitattributes'
FAILED test_gitattributes_env_disable
FAILED test_gitattributes_failure_never_breaks_the_build
```

The fifth, `test_to_json_leaves_a_directory_it_was_only_pointed_at_alone`, passes before and after on purpose: it pins that this does not start writing into arbitrary `--graph` targets.

`tests/test_export.py` is green at 65; the export, cli-export and atomic slices at 326 passed, 25 skipped. The full `tests/` run is 5255 passed with 14 failures, all of which also fail on unmodified `v8` here (terraform, ollama retry, markdown frontmatter, the wheel payload and a TS timing test) and none of which touch this path.

Mutation-checked: six seeded defects, including removing the call, dropping the output-directory guard, overwriting an existing file, ignoring the opt-out, letting a read-only directory raise, and dropping the `-diff` line. Each is caught by a named cell; none survive.

No formatter was run over the file, so the diff is only these lines.


dej-h:
We tested this same -diff plus linguist-generated approach in a repository with committed Graphify output. It did not reliably solve the GitHub PR totals problem for us.
I instead replaced it with a deterministic ZIP snapshot generated from the exact staged tree. Git then records one binary change with - - numstat and zero text insertions/deletions, while hooks restore the readable graph after checkout and commit.
I added the implementation details to #3413. I am happy to share the hook or help turn it into an opt-in Graphify mode.

L4XB:
Thanks for testing it, @dej-h. That settles it, and on re-reading #3413 it was there from the start. The reporter already had exactly this `.gitattributes`, `-diff` on all four files included, and still got the empty Files tab, which is what their screenshot shows. This PR read that configuration as the fix and shipped it, but by the issue's own evidence it doesn't fix #3413, and your repository confirms that independently. So I'm closing this rather than asking for a merge on a wrong premise.

One thing I checked so it doesn't get chased again: on `v8`, `to_json` already sorts nodes and links (`_json_sort_key`), so rebuild-to-rebuild reordering isn't what is left of the diff size.

Your snapshot mode is a real design choice: one deterministic binary blob with zero text insertions, in exchange for readable diffs of the graph in review and a hook that every clone has to run. That trade is the maintainers' call, and you're better placed to propose it in #3413 than I am.

Issue #3574, opened 2026-09-15
AST cache pins node IDs to the original extraction root; WordPress block.json never indexed


lt-cavalcanti:
Two separate problems found while auditing a graph that had drifted. Both are in
0.9.16. The second is the more serious one — it makes the graph quietly wrong in a
way `update` cannot repair.

## 1. WordPress `block.json` is classified as data JSON and skipped

`_is_config_json()` in `graphify/extractors/json_config.py` gates JSON extraction on a
filename allowlist (`_CONFIG_JSON_NAMES`) plus a top-level key probe (`_CONFIG_JSON_KEYS`).
`block.json` is in neither, so every WordPress block manifest is treated as data JSON and
skipped by the structural pass (the #1224 behavior).

On a WordPress monorepo with 73 blocks, 72 produced zero nodes. The single exception
declared a top-level `"$schema"` key, which happens to trip the fallback probe.

`block.json` is the WordPress block manifest — it is config, not data, and it carries
genuinely graph-worthy relationships: `name`, `parent`, `ancestor`, `allowedBlocks`,
`providesContext` / `usesContext`, and the `editorScript` / `style` handles. It is as much
a manifest as `package.json`, and it is the file that describes which blocks may nest
inside which.

**Suggested fix:** add `"block.json"` to `_CONFIG_JSON_NAMES`.

Workaround today: add `"$schema": "https://schemas.wp.org/trunk/block.json"` to each
`block.json`. That is valid WordPress practice and gives editor autocomplete, but it means
touching every block in the repo.

## 2. The AST cache stores node IDs, so they stay pinned to the original extraction root

Node IDs are derived from the extraction root path. The per-file AST cache stores the
fully-formed nodes, IDs included. So when the same repo is later extracted from a
different root, every cache hit replays the IDs built under the original root.

Seen concretely on a repo first indexed inside a container at `/workspace`, then updated
from the macOS host checkout:

- 1,022 of 1,423 cache entries held `workspace_`-prefixed IDs
- 5,370 of 17,363 nodes in the graph carried them (31%)
- 508 files existed in the graph **only** as stale container-era nodes
- 51 more carried both a stale and a fresh copy of themselves

Because the old and new IDs never collide, they are not detected as duplicates — they look
like unrelated nodes. Node counts go up on every run, nothing warns, and `explain` returns
plausible results. They are just partly results from a tree that no longer exists.

The distribution is the part that makes this hard to notice: contamination tracks edit
[...]

PR #3576, opened 2026-09-15
fix(json): index the WordPress block manifest

L4XB:
Part 1 of #3574.

`block.json` is in neither `_CONFIG_JSON_NAMES` nor the top-level key probe, so `_is_config_json` reads it as data JSON and the structural pass skips it (the #1224 behaviour). On the reporter's repo of 73 WordPress blocks, 72 produced zero nodes. The one that did produce nodes declared a top-level `"$schema"`, which trips the probe by accident rather than by intent.

It belongs on the allowlist. It is a manifest in the same sense `package.json` is, and it carries exactly the relationships a block graph exists to answer questions about: `parent`, `ancestor`, `allowedBlocks`, `providesContext` / `usesContext`, and the `editorScript` / `style` handles. It is the file that says which blocks may nest inside which.

The workaround available today is adding `"$schema": "https://schemas.wp.org/trunk/block.json"` to every block, which is valid WordPress practice but means touching every block in the repo to get a graph.

#### Test

One cell in `tests/test_extract.py`, next to the existing filename and key-probe cells. It uses a manifest with no `$schema`, so it grades the allowlist rather than the probe. On `v8`:

```
>       assert len(result["nodes"]) > 0
E       assert 0 > 0
FAILED tests/test_extract.py::test_extract_json_wordpress_block_manifest
```

`tests/test_extract.py` is green at 227 passed, 4 skipped.

Parts 2 and 3 of #3574 are separate: the AST cache pinning node IDs to the original extraction root is the serious one and is not touched here.


L4XB:
Closing: the reporter measured this allowlist entry on 73 real manifests in #3574, and 1,141 of the 1,202 nodes it produced were key names. That is the #1224 outcome `extract_json` skips data JSON to avoid. A value-level rule for `parent` / `ancestor` / `allowedBlocks` is proposed on the issue instead.

PR #3578, opened 2026-09-15
fix(global): give global add the community offset and shared type linking merge-graphs already has

### LINKED ISSUE #3100: The #3014 community offset and #3007 same_type_as pass never reach `graphify global add` — the cross-machine store still fuses communities and never links shared types
Both fixes landed in the merge-graphs path only:

- #3014 (0.9.50) offsets each input's community ids past the running maximum — in `dispatch_command`'s merge-graphs arm, via `prefix_graph_for_global(..., community_offset=)`.
- #3007 (0.9.49) links identically declared types across repos — `link_shared_type_declarations`, called right after merge-graphs composes.

`graphify global add` builds the same kind of multi-repo store (it uses the same `prefix_graph_for_global`), but `global_graph.py` calls the prefixer with the default offset and never invokes the shared-type pass. So a global store rebuilt today, on 0.9.50, still has exactly the defect #3014 fixed, plus none of #3007's links.

Measured on a real global store (226 repos, 1.21M nodes, 1.9GB `global-graph.json`, pushed to FalkorDB):

- 15 repos carry `community` ids on 60,209 nodes. Their per-repo distinct ids sum to 4,734, but only 2,399 distinct ids exist store-wide: **992 community ids are claimed by 2+ repos, and id 0 is claimed by all 15 repos at once — 1,042 unrelated nodes fused into "one" community**.
- `same_type_as` edges store-wide: **0**, while the same store demonstrably holds 14 (namespace, name) type declarations spanning 2+ repos (Temporalio .NET contract types shared between an SDK checkout and a consumer, plus your own C# test fixtures appearing in three graphify clones).

What we did on our fork (TelB-io/graphify#26), in case it is useful as a shape: both mechanisms ported into the streamed slice writer — pass 1 over the surviving nodes already exists there, so it also tracks the highest surviving `community` id (incoming slice offset past it, original kept in `local_community`, first slice into an empty store keeps its ids) and indexes surviving type declarations via the same candidacy predicate as `link_shared_type_declarations` (extracted to `cross_repo_types.declaration_key` so the two passes cannot drift). Incoming declarations are joined across the repo boundary with the #3007 edge attributes verbatim. On our 1.9GB store the refresh's peak RSS went from 231.2MB to 233.9MB (+1.2%), and re-adding one repo produced exactly the 4 predicted `same_type_as` edges with zero community collisions for the refreshed slices.

Happy to adapt it into a PR if you want it in the global path too; equally fine if you'd rather solve it differently.

ayushcodes10:
Fixes #3100.

## Summary

`graphify merge-graphs` got two fixes for building a cross-repo store: #3014 offsets each input's community ids into a shared id space as it composes them (every `graph.json` numbers its own communities from 0, so ids carried across unchanged collide), and #3007 links identically declared types across repos so a traversal can cross the repo boundary (a type both repos declare otherwise arrives as two unconnected nodes, since every id is repo-prefixed).

`graphify global add` builds the same kind of multi-repo store with the same `prefix_graph_for_global` prefixer, but never got either fix — it called the prefixer with the default (no) offset, and never called the type-linking pass at all. Two independent reporters confirmed this on real stores of very different shape and scale: one with 226 repos / 1.21M nodes (992 colliding community ids, id 0 alone claiming 1,042 unrelated nodes across 15 repos, 0 shared-type links despite 14 real cross-repo shared types), the other with 11 repos on an older release (52 colliding ids, 532 nodes fused at id 0).

## Changes

- `graphify/global_graph.py`: `global_add` now computes a community offset from the store's own current community ids (after pruning this repo's stale entries, so re-adding the same repo doesn't inflate it forever) and passes it to `prefix_graph_for_global`, reusing the exact function `merge-graphs` already calls. It also now calls `link_shared_type_declarations` over the whole store on every add — the same idempotent pass (only adds an edge where none exists yet) the existing member-call linking already does just above it, so a type declared in a repo added long ago still gets linked against one added today. The new edge count is threaded through the return dict alongside the existing `cross_repo_calls`.
- `graphify/cli.py`: prints the new `shared_type_links` count, mirroring the existing `cross_repo_calls` print right above it.
- `tests/test_global_graph.py`: two new tests — community ids landing on disjoint sets across two repos added one at a time, and a `same_type_as` edge appearing (with the reported count matching) once a second repo declares a type the first one also declared.

## Testing

- `python -m pytest tests/test_global_graph.py -q` — 22 passed
- Full suite: `python -m pytest -q` — 5487 passed, 68 skipped (pre-existing unrelated failures excluded: `tests/test_ollama_retry_cap.py` missing the optional `openai` module, and `test_ts_normalizer_scales_linearly_on_large_files` / `test_hyperedge_convex_hull_js_is_geometrically_sound`, both known environment-specific flakes unrelated to this change)
- Manually reproduced end to end with isolated `_GLOBAL_DIR`/`_GLOBAL_GRAPH`/`_GLOBAL_MANIFEST` patches (matching the existing test fixture pattern): two repos each with their own communities land on disjoint ids, and two repos declaring the same namespaced type gain a `same_type_as` edge on the second add.


ayushcodes10:
Thanks for catching the stale-base issue — rebased onto the current v8 tip (post 0.9.62) in d17f7e0, so the diff no longer appears to touch _id_prefixes or _claude_pretooluse_hooks; those were both already-shipped fixes my branch predated, not something this PR changes. Full suite (5595 tests) and skillgen check both pass clean on the rebased branch.

On the concurrent global_add race: confirmed this is pre-existing, not introduced or worsened in kind by this change. global_graph.py has no locking at all (unlike watch.py's _rebuild_lock for the equivalent update path), so two concurrent global_add calls could already clobber each other's writes via the same load-modify-save pattern before this PR. This change does add more work inside that window (the offset computation and the shared-type linking pass), which widens it somewhat, but doesn't create a new race class. Fixing it properly would mean adding a lock to global_graph.py mirroring watch.py's, which is a broader change than this issue's scope (#3100, community offset + shared type parity with merge-graphs) — happy to file a follow-up issue for it if that's useful, but treating it as out of scope for this PR.

ayushcodes10:
Good catch — fixed. `global_add`/`global_remove` each load the graph and manifest, mutate them in memory, then save both back, with no locking anywhere. Two concurrent calls would both read the same pre-write snapshot, compute conflicting results, and the second save would silently discard the first's work entirely — not just a duplicated community id (which the narrower fix on this issue already closes), but potentially an entire repo's added nodes lost.

Added an exclusive advisory lock (`fcntl.flock`, matching the same technique and Windows no-op fallback already used by `watch.py`'s per-repo rebuild lock) around the whole load/mutate/save cycle in both functions. Blocks until acquired rather than failing, since a CLI caller should just wait its turn. Two regression tests: one verifying the lock's own mutual exclusion directly, and an end-to-end check that two concurrently added repos both survive.

ayushcodes10:
Good catch — fixed. Hashing \`source_path\` happened before the lock while parsing it happened after, two reads of the same file at two different times. A concurrent writer to that file in between could make the recorded manifest hash describe different bytes than what actually got imported, corrupting the unchanged-hash skip check on every later call.

Both now come from a single read inside the lock (the standalone hashing helper is gone since it had no other caller). Added a regression test that counts reads of the source path and asserts exactly one happens.

(The Windows no-op fallback and the test timeout are as-designed — the former matches the existing per-repo rebuild lock's own documented fallback, and the latter is a safety bound on an otherwise-fast test, not a source of flakiness.)

ayushcodes10:
Good catch, and sorry for the regression — fixed. Moving the size cap check ahead of the unchanged-hash skip (to consolidate hashing and parsing into a single read) meant it was reached on every call, even when nothing changed, so an already-tracked graph that crossed the cap (its own growth, or a lowered `GRAPHIFY_MAX_GRAPH_BYTES`) started erroring on every subsequent call instead of continuing to skip — the cap was never reached at all on that path before that fix.

The cap check moves back to guarding only the actual parse, right after the skip check, matching where it ran before, while keeping the single read. Regression test covers a graph that succeeds under the cap, survives a later cap drop while unchanged (skips, doesn't raise), and confirms a genuinely different oversized file is still rejected.

ayushcodes10:
Both good catches — fixed.

**Size cap read order:** hashing `source_path` (to check the unchanged skip) still read its full bytes into memory before the size cap ever ran, defeating the cap's whole purpose of failing fast before a multi-GiB file is loaded, for exactly the case that matters most: a new or genuinely changed oversized file. Added a stat-only fast path first (mtime and size compared against what a previous successful add recorded, no read at all), so an already-tracked, genuinely unchanged file keeps skipping even once it or the configured cap crosses the threshold — the same guarantee the earlier ordering fix protected — without needing to read anything or reach the cap check. Only once that fast path misses does the cap check run (still no read), and only past that does the single real read for hashing and parsing happen. Test asserts zero reads occur for a new oversized file before it's rejected.

**Timed joins:** added `assert not t.is_alive()` after each `.join(timeout=...)` in both concurrency tests, so a deadlock fails loudly instead of silently passing.

ayushcodes10:
Good catch — fixed. The mtime/size fast path never compared the source path itself, so a repo tag re-pointed at a genuinely different file that happens to share the old file's mtime and size (a real possibility with a preserved-timestamp copy or sync tool, or a checkout that restores original timestamps) would fast-skip on that coincidence alone, silently keeping stale data in the store — the early return happens before the "repo tag now points elsewhere" warning further down is even reached.

The resolved source path is now part of the fast path's own comparison, computed once up front and reused everywhere else this function already needed it (previously called three separate times). Regression test uses two files padded to the exact same size with mtime forced to match at nanosecond precision, confirming the second add is not skipped and the global graph reflects the new content, not the stale one.

On the pattern across rounds: I take the point that this is now 5 rounds on the same ~15-line block. Looking back at each one, they've been genuinely independent, non-overlapping properties (concurrency safety, read consistency, cap-vs-skip ordering, cap-vs-read ordering, then this path-identity gap) rather than the same bug resurfacing — each fix has held up under the next round's scrutiny. I don't think a rewrite would have caught this last one any more reliably than the incremental fixes did; happy to do a closing pass reading the whole function fresh end-to-end if that would help confidence, but didn't want to reshape working, tested code speculatively without a concrete finding driving it.

ayushcodes10:
Pushed one more fix for a gap found in review: a manifest entry written before the mtime/size fast path fields existed (every manifest from before this PR) has no stat baseline to compare against, so it fell straight through the fast path into `check_graph_file_size_cap` and errored on a genuinely unchanged file that had simply grown past the cap since it was last recorded. The existing hash based unchanged skip further down the function would have caught this, but was never reached because the cap check raised first.

Fix: added `_hash_file_streaming`, a SHA256 helper that reads the file in bounded chunks rather than one `read_bytes()` call, so its memory use never depends on the file's size. When the cap check fails, we now fall back to comparing the streaming hash against the entry's stored `source_hash` before giving up; a match skips (any manifest vintage, legacy or not), and only a genuine mismatch lets the original cap error through.

This also feels like the natural resting point for the cap vs hash tension that's been the throughline of the last several rounds here. The cap's only real job is protecting the eventual JSON parse, which genuinely does need the whole file in memory; hashing never needed that, it just wasn't being done in a way that respected the cap. Separating those two concerns (cheap stat check first, then a bounded-memory streaming hash as the fallback, only then the parse) means there's no longer a scenario where an unchanged file gets punished for growing past a size limit, regardless of which version of the manifest format tracked it.

Added two regression tests: one confirming a legacy hash-only manifest entry skips when the file is unchanged but now oversized, and a companion confirming a legacy entry with a genuinely different, still-oversized file still raises the cap error. Both were checked to fail against the pre-fix code first. Full `test_global_graph.py` (30 tests) and the full suite (5603 passed) are green.

ayushcodes10:
Addressing the three advisories from the last review:

**1. "Changed file content can be skipped when mtime and size are preserved" — real, fixed.**

The stat-only fast path trusted a matching mtime and size alone as proof of an unchanged file, but neither actually is: a write landing in the same filesystem tick as the last recorded read can leave both untouched, and a tool that writes new content and then restores an old mtime (a same-size `cp -p`/`rsync -a` against an unrelated source, or a build step that pins timestamps for reproducibility) can make a genuinely changed file look identical to what was last recorded — silently skipping the re-import.

Fixed with two independent guards, both mirroring the approach `graphify.cache`'s own content cache already uses (`_stat_sig_fresh`):
- The last successful read now stamps when it happened (`source_indexed_at_ns`, captured strictly before the read), and the fast path requires that stamp to sit after the recorded mtime's own filesystem tick closed — the same racily-clean guard the content cache already relies on.
- The manifest now also records each add's ctime (POSIX inode change time), which can't be forged the way mtime can — verified this empirically before implementing: writing new content, or the very `utime()` call that resets mtime afterward, both bump ctime to the real current time regardless of what mtime ends up being set to. On Windows, `st_ctime` instead reports file *creation* time, so this second guard degrades to a safe no-op there rather than a false rejection.

Two new regression tests. Caught a real bug in my own first draft of one of them along the way: the codebase's established `_settle()` pattern (backdate mtime via `os.utime()` to skip the racy window) is incompatible with testing the ctime guard, since `os.utime()` itself always bumps ctime — had to capture the restore baseline immediately after the add that recorded it, and disable the racy-window guard directly instead of backdating, to isolate the ctime check specifically. Both new tests checked against pre-fix code first; the content-change one correctly failed there.

**2. "global_add lacks the store lock" — not accurate as stated.** `global_add` has held `_global_store_lock()` since the round 7 fix (see line ~145 on this branch) — same as `global_remove`. Nothing to change here.

**3. "Lock is a no-op without fcntl (Windows)" — intentional, already documented.** The lock's own docstring explains this matches the same fallback `watch.py`'s per-repo rebuild lock already uses; a lost update there was a pre-existing, accepted risk this PR doesn't newly introduce.

Full suite (5605 passed) is green.

ayushcodes10:
Addressing the four advisories from the latest round:

**1. "global_add body not shown under _global_store_lock" — not accurate.** Re-checked directly: the entire body of `global_add`, from `manifest = _load_manifest()` through `_save_manifest(manifest)`, sits inside `with _global_store_lock():`. This is unchanged from the round 7 fix. If this is based on a diff view that only shows part of the function, happy to look again at a specific line range, but I can't find a code path that runs outside the lock.

**2. "Same-hash path change returns before updating manifest source_path" — real, fixed.** A repo tag re-pointed at a new path with identical content took the hash-match early return without ever refreshing that entry's `source_path`/mtime/size/ctime/indexed_at. Since the stat fast path requires `source_path` to match the resolved path, a stale `source_path` can never match again, so every later add for that tag silently fell back to a full read + hash forever, and the "previously pointed to" warning reprinted on every call instead of the one time it's meant to. Fixed by refreshing the stat baseline (keeping `added_at`/`node_count`/`edge_count` from the existing entry, since the content itself didn't change) before returning. Regression test confirms the manifest reflects the new path after the repoint, and the warning doesn't reprint on a subsequent call.

**3. "Stat fast-path can skip changed files on Windows when mtime and size are preserved" — already a known, accepted tradeoff, not new.** This is specifically about the ctime guard degrading to a no-op on Windows (`st_ctime` reports creation time there, not change time), which is documented in the fast path's own comment. It mirrors the identical tradeoff `graphify.cache`'s own content cache already accepts for the same reason. Nothing new to fix here without a different detection mechanism on Windows entirely, which feels out of scope for this PR.

**4. "Size cap can be bypassed by replacing the source between stat and read" — real gap, mitigated.** The pre-read `check_graph_file_size_cap` call is stat-based, so a file swapped for something larger between that check and `read_bytes()` just below it was never re-confirmed. A full fix would need a bounded streaming read of the whole hash+parse path, which felt too invasive for what's a local, single-user TOCTOU (no privilege boundary — anyone who can race a file swap on disk already has as much access as this process). Added a cheap post-read length check instead: it can't avoid the one-time memory cost of the read that already happened, but it does stop an oversized payload from reaching the much more expensive JSON parse and graph merge, and reports the same cap error. Regression test simulates the race by bypassing the pre-read check directly.

Full suite (5607 passed) is green.

PR #3587, opened 2026-09-16
fix(markdown): resolve review findings and keep mention edges across rebuilds

AstroMined:
Follow-up to #3562, which shipped in 0.9.62 as c805c6f. The two commits pushed to that branch after the automated review were not cherry-picked, so 0.9.62 carries the reviewed defects. This PR rebases them onto v8 unchanged.

**fix(markdown): honour dotted qualifiers and explicit relative cites**
- Dotted spans keep every qualifier; a bare-name match must carry each qualifier as a parent label or a path segment of its file, so `README.md`, `pyproject.toml` and `time.sleep` no longer match a same-named callable. The file early-exit also checks `DOC_EXTENSIONS`.
- `_match_cited_file` strips `./` and `../` as prefixes, so `.github/...` cites can match by suffix, and a root-level file can match the suffix form.
- An explicit `./` or `../` cite resolves relative to the document only; it never falls through to the suffix rule.

**fix(watch): keep code-span mention edges through a rebuild over an existing graph**
`_reconcile_markdown_links` treated every AST-tier `references` edge owned by a re-parsed markdown file as an authored `[link](file)` and pruned the code-span mention edges, so the first `graphify update` after a full build dropped every doc-to-code edge. Mention edges target code symbol nodes rather than file nodes and are now kept.

Tests: `tests/test_markdown_code_spans.py` covers each case; the rebuild test fails on v8 without the watch change and passes with it.


safishamsi:
Shipped in **v0.9.63** (on PyPI). Cherry-picked with authorship preserved. The rebuild-persistence fix (routing the survival test through the real rebuild) and the dotted-qualifier evidence check were both the right calls. Thanks @AstroMined!

PR #3590, opened 2026-09-16
fix(diagnostics): separate external references from dangling edges

### LINKED ISSUE #2191: diagnose_extraction counts external-package import refs (ref_*) as dangling edges — false "graph may be incomplete/corrupt" warning on JS/TS projects
## Environment

- graphifyy 0.9.26, Python 3.12.10, Windows 11
- Project: pnpm monorepo — Next.js 16 storefront + Strapi 5 CMS, 230 code files (TS/TSX), `node_modules` excluded (default)

## What happens

Step 4.5 of the skill pipeline (`graphify.diagnostics.diagnose_extraction`) reports:

```
dangling_endpoint_edges: 217
GRAPH HEALTH WARNING: 217 dangling-endpoint edges - graph may be incomplete/corrupt.
```

Breaking those 217 edges down by missing endpoint shows they are all AST import references to **external npm packages**, for which no nodes can exist because `node_modules` is (correctly) excluded:

```
 58  ref_strapi_strapi      (@strapi/strapi)
 39  ref_next
 38  ref_next_link
 33  ref_react
 13  ref_next_navigation
  4  ref_next_headers
  3  ref_path
  3  ref_strapi_strapi_admin
 ...
```

So the warning fires on what is arguably *every ordinary JS/TS project*, and re-extraction cannot fix it — the count stays identical run to run. In our case the LLM-extracted semantic edges were 100% valid; all "dangling" edges came from the deterministic AST layer's external-import refs. We spent a full forced re-extraction chasing this before isolating the cause.

The build layer already handles this correctly (these edges are excluded from `graph.json`), so the graph itself is fine — only the diagnostic message is misleading.

## Steps to reproduce

1. Any Next.js/React/Strapi repo, run the standard `/graphify .` pipeline.
2. Run `diagnose_extraction(extraction, directed=False, root='.')` on the merged extraction.
3. Observe `dangling_endpoint_edges` ≈ (number of external-module import edges), plus the corrupt-graph warning.

## Expected / suggested

- Classify edges whose missing endpoint matches the external-ref scheme (`ref_*`, or "endpoint not under scan_root") into a separate counter, e.g. `external_reference_edges`, excluded from the GRAPH HEALTH WARNING; **or**
- resolve them to synthetic "external package" nodes so they are not dangling at all;
- keep the warning for genuinely dangling endpoints (typo'd IDs, cross-chunk drift) — that signal is valuable, it is just drowned out today.

DevChiniwala:
## Summary
- classify missing `ref_*` dependency targets separately from internal dangling endpoints
- keep graph-health diagnostics actionable for malformed and internal missing IDs
- expose the external-reference count in machine-readable and text reports

Fixes #2191

## Validation
- `python -m pytest tests/test_multigraph_diagnostics.py tests/test_duplicate_annotation_edges.py tests/test_evidence_binding.py -q --tb=short` — 48 passed
- `python -m ruff check graphify/diagnostics.py tests/test_multigraph_diagnostics.py` — passed
- `graphify update .` — completed on current `v8`; expected optional-parser warnings only

## Compatibility
This adds `external_reference_edges` and narrows `dangling_endpoint_edges` to actionable missing endpoints. Existing malformed and internal missing targets remain classified as dangling.


safishamsi:
Shipped in **v0.9.64** (on PyPI). Cherry-picked with authorship preserved so it shows under your GitHub contributions. Thanks @DevChiniwala!

PR #3598, opened 2026-09-16
Resolve Kotlin object/class qualified member calls across files

### LINKED ISSUE #1698: Kotlin: cross-file calls to `object` (singleton) members produce no calls edges — same-file resolution works

ayushcodes10:
## Summary

Fixes #1698.

A Kotlin call on an object or class qualified receiver (`Config.load()`) only resolved when the receiver's declaration lived in the same file. The gap sits at the intersection of three existing pieces of behavior:

- The shared cross file call resolution pass in `extract.py` unconditionally skips every member call, by design (a bare method name colliding across the whole corpus would inflate god nodes, per #543/#1219). Member calls can only ever resolve cross file through a dedicated per language resolver.
- The engine deliberately does not capture a receiver for Kotlin's two segment `object.method()` navigation shape, specifically so it would not get deferred toward cross file resolution when no such resolver existed yet (deferring it with nothing to catch the deferred call would have broken the working same file case).
- No dedicated Kotlin member call cross file resolver had ever been written, unlike every other member call heavy language in the codebase (Swift, Python, TypeScript, C++, C#, Java, Objective C).

## Changes

- Capture the receiver for Kotlin's two segment qualified navigation shape and thread it into `raw_calls`.
- Add `_resolve_kotlin_member_calls`, mirroring the existing `_resolve_kotlin_qualified_calls` and the `type_by_name` / `methods_by_type` shape `_resolve_swift_member_calls` already uses: index Kotlin sourced classes and objects by bare name, resolve the receiver against that index, then look up the callee among that type's methods. Both steps bail on anything but exactly one candidate, the same god node guard every cross file resolver in this module already applies.
- Register the new resolver for `.kt` / `.kts` files.

## Test plan

- New regression tests in `tests/test_kotlin_grammar.py`:
  - the issue's own cross file case now produces an `EXTRACTED` `calls` edge
  - the pre existing same file case is unaffected (negative control)
  - a receiver defined identically in two packages correctly yields no edge (god node guard)
- Full suite: 5596 passed, 68 skipped, 2 deselected, no regressions.


ayushcodes10:
Re: "Method label strip('()') mismatch when method has parameters in label" (`graphify/extract.py:4770`)

Checked this directly: every callable node label in the codebase, across all languages including Kotlin, is built from a hardcoded `f".{name}()"` template with literal empty parens (`engine.py` lines 2354, 2436, 2487, 4946) — never real parameter text. Confirmed by running the extractor on `fun record(m: Money) { }` (a method with a parameter): the label comes out as `.record()`, not `.record(m: Money)`. `tests/test_kotlin_grammar.py` on this branch independently confirms the same convention throughout (`.load()`, `.help()`, `.f()`, `.go()`, `.checkout()`, `.total()` — always empty parens).

So `.strip("()")` on `".record()"` correctly yields `".record"`, matching the wanted bare name. No fix needed here — false positive.

safishamsi:
Shipped in **v0.9.64** (on PyPI). Cherry-picked with authorship preserved so it shows under your GitHub contributions. Thanks @ayushcodes10! (Added companion-object + real-incremental regression tests on merge.)

PR #3601, opened 2026-09-16
Resolve npm package subpath imports to the same node as a bare import

### LINKED ISSUE #3595: JS/TS: npm package subpath imports (next/image, next/link) produce no edge, while bare imports of the same package resolve
**Version:** graphifyy 0.9.62 (PyPI) · Python 3.13.3 · macOS (darwin) · TypeScript extractor (tree-sitter)

## Summary

A static import of an npm package **subpath** (`next/image`, `next/link`) produces no edge at all, while a **bare** import of the same package (`next`) resolves correctly to the `package.json` node.

The importing file keeps its node, so nothing looks wrong — the file is simply disconnected from the framework it depends on. `graphify query "what uses next/image?"` returns nothing, with no warning.

This is a big blind spot for Next.js / React codebases, where `next/image`, `next/link` and `next/navigation` carry most of the framework coupling. In a real 327-node Next.js site I measured **21 dropped import edges** across 7 subpaths (`next/link` ×10, `next/navigation` ×3, `next/server` ×2, `next/image` ×2, `next/headers` ×2, `next/og` ×1, `next/font/google` ×1).

Note this is *not* the same as #3084, which reports that **bare** static imports also dangle — on 0.9.62 the bare case works (verified below), so that half appears to have been fixed since 0.9.39. The subpath case remains.

## Minimal reproduction

Three files in an empty directory:

`package.json`
```json
{
  "name": "mre-subpath-import",
  "version": "1.0.0",
  "dependencies": { "next": "16.2.7", "react": "19.2.4" }
}
```

`bare.ts`
```ts
import type { Metadata } from "next";
import { useState } from "react";
export const meta: Metadata = {};
export const s = useState;
```

`subpath.ts`
```ts
import Image from "next/image";
import Link from "next/link";
export const i = Image;
export const l = Link;
```

Then:
```bash
```

## Actual result

All four import edges should exist; only the two bare ones do.

```
NODES: ['bare.ts', 'dependencies', 'i', 'l', 'meta', 'name', 'next', 'next',
        'package.json', 'react', 'react', 's', 'subpath.ts', 'version']

IMPORT EDGES:
  package.json --imports--> next
  package.json --imports--> react
  bare.ts --imports_from--> next     <-- bare resolves
  bare.ts --imports_from--> react    <-- bare resolves
                                     <-- subpath.ts has NO import edges
```

`subpath.ts` is present as a node with zero outbound import edges.

## Expected result

A subpath specifier should resolve to its owning package node, so that
`subpath.ts --imports_from--> next` exists (ideally twice, or once with the
subpath retained as edge metadata).

Stripping the specifier to its package root before canonicalization would cover
the common cases: `next/image` -> `next`, `@scope/pkg/sub` -> `@scope/pkg`
(first *two* segments for scoped packages).

## Why it matters

Framework surface is exactly what an architecture graph should show. For a
Next.js project the questions you most want to ask — which routes use
`next/headers`, which pages client-navigate via `next/link`, which components
render `next/image` — all silently return nothing. An empty result is
indistinguishable from "no usages", so the gap is invisible unless you already
know the answer.

ayushcodes10:
## Summary

Fixes #3595.

A JS/TS import of an npm package subpath (`next/image`, `next/link`) produced no edge at all, while a bare import of the same package (`next`) resolved correctly to the `package.json` dependency node.

Root cause: once relative paths, tsconfig aliases, and workspace packages have all failed to resolve an import, `_resolve_js_import_target` falls back to an external reference namespaced on the whole raw specifier (`ref:<raw>`). A bare import of `next` produced `ref:next`, matching the dependency node `json_config.py` mints for `package.json`'s own `"next"` key — but a subpath import of `next/image` produced `ref:next/image`, a different id with no backing node, so `build.py` dropped the edge as dangling.

## Change

The external-import fallback now strips the specifier down to its package root before namespacing: the first segment for an ordinary package, the first two for a scoped one (`@scope/pkg/sub` → `@scope/pkg`), matching how the `package.json` dependency node is keyed. A subpath import now lands on exactly the same node as a bare import of that package.

## Test plan

- Two resolver unit tests: a plain subpath and a scoped-package subpath each resolve to the same target as their bare import, and a scoped subpath never folds onto a different package sharing the same scope.
- An end to end test using the issue's exact repro shape (`package.json` + a bare-import file + a subpath-import file): both files now resolve to the same target instead of the subpath file coming out disconnected.
- Existing `ref`-namespacing regression tests (`test_phantom_external_import.py`) still pass unchanged.
- Full suite: 5596 passed, 68 skipped, 2 deselected, no regressions.


safishamsi:

PR #3602, opened 2026-09-16
Resolve Rust self.method() calls across files

### LINKED ISSUE #2234: fix(rust): receiver-method calls (self.method()) are dropped cross-file — Rust has no _resolve_*_member_calls pass
## Problem

Rust receiver-method calls (`self.method()`, `receiver.method()`) never produce a cross-file
`calls` edge. Rust is the only major supported language with no `_resolve_*_member_calls` pass,
so its member calls are dropped by the shared guard and nothing recovers them.

On a real Rust codebase ([doli-network/doli](https://github.com/doli-network/doli), 533 `.rs`
files, ~24k-edge graph) this makes blast-radius queries silently wrong on exactly the symbols
that matter most — the hub methods on the main `impl`:

| symbol | incoming `calls` in graph | real call sites (grep) |
|---|---|---|
| `.apply_block()` | **0** | 68 |
| `.rollback_one_block()` | **0** | 22 |
| `.calculate_epoch_rewards()` | **0** | 18 |
| `.execute_reorg()` | 1 | 7 |
| `.handle_new_block()` | **0** | 5 |
| `.add_transaction()` | 26 | 76 |

Graph-wide: only 32% of the 2,306 method-labelled nodes have any incoming `calls` edge.

`blast.py graph.json apply_block --hops 1` returns **1 dependent** for a function called from
6 files, and it looks like a confident answer. That is worse than no graph — it is the failure
mode #2135's description names ("silently under-report, which pushes users back to grep"),
just on the other call form.

## Where it is dropped

The extractor is fine: `self.apply_block(...)` is a `call_expression` over a `field_expression`,
so `extractors/rust.py:359` tags it `is_member_call=True` and enqueues it into `raw_calls` —
it is **not** dropped at extraction.

It dies in every cross-file resolver:

- `symbol_resolution.py:265` — `if raw_call.get("is_member_call"): continue`
- `symbol_resolution.py:336` — same
- `extract.py:5198` — same, with the rationale: *"obj.log() → 'log' has no import evidence and collides with any top-level function named 'log' in the corpus"* (#543/#1219)

That guard is correct on its own terms. The problem is that seven languages have a recovery pass
behind it and Rust has none:

```
_resolve_swift_member_calls       (extract.py:2054, #1356)
_resolve_python_member_calls      (extract.py:2169)
_resolve_typescript_member_calls  (extract.py:2312)
_resolve_cpp_member_calls         (extract.py:2404)
_resolve_csharp_member_calls      (extract.py:2532)
_resolve_java_member_calls        (extract.py:2743)
_resolve_objc_member_calls        (extract.py:2832)
                                  ← no _resolve_rust_member_calls
```

## Not covered by #2135

I built the graph twice on the same 533-file Rust tree — once on `v8` (`1644230`, = 0.9.28) and
once on #2135 (`322a2a9`), `--no-cluster`, cache cleared between runs:

```
nodes:  9877 → 9877      (unchanged)
edges: 23920 → 24082     (+162, all EXTRACTED; 0 removed; INFERRED unchanged at 961)
```

The +162 are real — spot-checked at the exact source lines
(`crate::gossip::staleness::classify_gossip(...)`, `cmd_wallet::cmd_add_bls(...)`,
`reward_epoch::is_epoch_start_with(...)`, `run::run_node(...)`). #2135 is a clean win on this
codebase and I'd like to see it land.

But it moves the `is_scoped_call` branch (`rust.py:364`), and every hub method above is on the
`is_member_call` branch (`rust.py:359`). Measured: **all six rows in the table are byte-identical
before and after #2135.** Different guard, adjacent blind spot.

## Suggested shape

The Swift pass (#1356) looks like the closest precedent: record the receiver plus a per-file
`name -> type` table, resolve the receiver's type, and emit only when the type name resolves to
exactly one definition — EXTRACTED when the type is explicit, INFERRED when locally inferred,
bail on ambiguity so the god-node guard is preserved.

For Rust the receiver-typing signal is comparatively strong:

- `self.method()` inside an `impl Foo` block types the receiver as `Foo` syntactically — no
  inference needed, and this alone covers the whole table above.
- `let x = Foo::new();` / `let x: Foo = ...` / typed fn params give a per-file `name -> type`
  table for non-`self` receivers.
- Trait-method ambiguity stays behind the existing `_RUST_TRAIT_METHOD_BLOCKLIST` and the
  one-survivor-or-bail rule.

Scoping just the `self.` case inside `impl` blocks would be a small, high-confidence slice — it
needs no type inference at all, and it is where the missing edges concentrate.

## Repro

```bash
git clone https://github.com/doli-network/doli
cd doli && graphify update . --no-cluster
graphify explain ".apply_block()" --graph graphify-out/graph.json   # degree 6, zero incoming calls
grep -rn "\.apply_block(" crates bins --include="*.rs" | wc -l      # 68
```

Environment: graphify 0.9.28, Python 3.14, macOS 15 (arm64).

ayushcodes10:
## Summary

Fixes #2234.

Rust receiver-method calls (`self.method()`, `receiver.method()`) never produced a cross-file `calls` edge. Rust is the only major supported language with no `_resolve_*_member_calls` pass, so its member calls were dropped by the shared "member calls have no import evidence" guard (#543/#1219) with nothing behind it to recover them — unlike Swift, Python, TypeScript, C++, C#, Java, and Objective-C, each of which has a dedicated resolver.

This lands the narrow, high-confidence slice the issue itself suggested: `self.method()` inside `impl Foo { .. }`. The receiver's type is known syntactically at the call site — no inference needed — and this alone covers the hub-method blind spot the issue measured on a real 533-file Rust tree.

## Design note: why this differs from the Kotlin/Swift resolvers

Kotlin and Swift's resolvers require a type name to resolve to exactly one node before proceeding. A literal port of that rule to Rust would have refused the issue's own motivating case: Rust routinely splits `impl Foo { .. }` across as many files as it likes, and `rust.py` mints a *separate* graph node per `(file, type name)` pair rather than merging them into one canonical type node. Requiring a single node would bail on every split impl block — exactly backwards for this issue.

Instead, every node carrying the type's bare label pools its methods together, and a call resolves only when exactly one method across the *whole* pool matches the callee name. Two impl blocks (however many files apart) defining the same method name still bail — a real ambiguity, not a false one. Verified both directions: a split impl block resolves correctly, and two unrelated structs sharing a bare name with a colliding method name correctly yield no edge.

## Test plan

- New `tests/test_rust_self_member_calls.py`: the issue's own split-impl-block shape now resolves (EXTRACTED); the pre-existing same-file case is unaffected; two differently-defined structs sharing a name and a colliding method yield no edge; a call to a genuinely undefined method yields no edge.
- Full suite: 5597 passed, 68 skipped, 2 deselected, no regressions.


ayushcodes10:
Both findings confirmed real, pushed fixes for both:

- **"method_index overwrites entries for duplicate (type-node, method) pairs, potentially masking ambiguity"**: real, if narrow — a plain dict assignment kept only the last-seen target whenever two method nodes shared both a source impl node and a stripped label, instead of surfacing that as the ambiguity it is. The index now keeps a set per key.
- **"Rust self-call resolver suppresses calls when any other edge already exists"**: real. The dedup check matched on ANY relation, not just `calls`, so a caller that already had e.g. a `references` edge to the same target (from also naming the type elsewhere) silently suppressed a genuine call to it. Scoped the check to `calls` edges only — there's already an established precedent for this exact fix elsewhere in this module (`extract.py:4577`, with a comment describing the identical failure shape), which I matched.

Added two direct unit tests against the resolver (constructing the node/edge shapes directly, since the natural end-to-end extraction path doesn't organically produce a pre-existing non-`calls` edge to a method node) — both fail on the pre-fix code and pass now.

Full suite (5599 passed, no regressions) and `skillgen --check` both green.

ayushcodes10:
Re: "Rust self-call resolver can link across unrelated types with the same bare name" — confirmed real, and worse than advisory: reproduced it directly.

Two unrelated structs both named \`Config\` in different files, each with its own methods and no name collision between them — a call to a method that exists on one but not the caller's own \`Config\` incorrectly linked to the unrelated one's method. This can happen in real, compiling Rust: one type overrides a trait default method while an unrelated same-named type relies on the default, so only the override gets an explicit method node, and the pooling mechanism that makes a split impl block work had no way to tell "same type split across files" apart from "different types that happen to share a name."

Fixed by using a signal already in the graph: a struct/enum/trait *declaration* gets a \`contains\` edge from its file; an impl block never does. So the type name itself is now treated as ambiguous — pooling skipped entirely — whenever 2+ nodes sharing the bare label are real declarations rather than impl blocks. A single declaration with any number of impl blocks spread across files (the shape this pass exists for) is unaffected; verified with a new test that the wrong-link repro above now correctly yields no edge, and the existing split-impl-block test still resolves.

Full suite (5600 passed, no regressions) and \`skillgen --check\` both green.

safishamsi:
Shipped in **v0.9.63** (on PyPI). Cherry-picked with authorship preserved. Nicely scoped — pooling methods across every `impl` of one type while refusing when two unrelated types share a name is exactly right for Rust's split-impl layout. Thanks @ayushcodes10!

PR #3603, opened 2026-09-16
Resolve Elixir alias/import/require/use targets across files

### LINKED ISSUE #2556: Elixir: imports edges target a bare module id that matches no node — every cross-module import edge is silently dropped at build (2,586 of 19,337 on a 900-file project)
## Summary

The Elixir extractor emits `imports` edges whose target is the **bare normalized module name**, while the module node it should point at is registered under a **path-prefixed id**. The two never match, so `build_from_json` drops every cross-module import edge.

On a real 900-file Elixir/Phoenix project this silently discarded **2,586 of 19,337 extracted edges (13%)** — the entire internal module dependency graph. Nothing warns; the build reports success and the graph looks plausible. It is only visible via `graphify diagnose multigraph`.

Python does not have this bug, so this is Elixir-specific (details below).

## Repro

Two files, one aliasing the other:

```elixir
# lib/demo/accounts.ex
defmodule Demo.Accounts do
  def list_users, do: []
end

# lib/demo/web.ex
defmodule Demo.Web do
  alias Demo.Accounts

  def index, do: Accounts.list_users()
end
```

```python
from pathlib import Path

res = extract(collect_files(Path("lib")), cache_root=Path("."))
ids = {n["id"] for n in res["nodes"]}
for e in res["edges"]:
    flag = "  <-- DANGLING" if e["target"] not in ids else ""
    print(f"{e['source']} --{e['relation']}--> {e['target']}{flag}")
```

Output (graphify 0.9.36):

```
NODES:
  lib_demo_accounts                        label= accounts.ex
  lib_demo_accounts_demo_accounts          label= Demo.Accounts
  lib_demo_accounts_demo_accounts_list_users
  lib_demo_web                             label= web.ex
  lib_demo_web_demo_web                    label= Demo.Web
  lib_demo_web_demo_web_index

EDGES:
  lib_demo_accounts --contains--> lib_demo_accounts_demo_accounts
  lib_demo_accounts_demo_accounts --method--> lib_demo_accounts_demo_accounts_list_users
  lib_demo_web --contains--> lib_demo_web_demo_web
  lib_demo_web --imports--> demo_accounts        <-- DANGLING
  lib_demo_web_demo_web --method--> lib_demo_web_demo_web_index
```

**Expected:** `lib_demo_web --imports--> lib_demo_accounts_demo_accounts`
**Actual:** target is `demo_accounts`, which matches no node, and the edge is dropped at build.

## Root cause

`graphify/extractors/elixir.py` builds the module node id and the import target id with different arities:

```python
# L119 — module node: stem IS included
module_nid = _make_id(stem, module_name)      # -> lib_demo_accounts_demo_accounts

# L154 — import target: stem is NOT included
tgt_nid = _make_id(module_name)               # -> demo_accounts
```

L154 cannot know the stem on its own, because the defining file is usually a *different* file. The per-file filter at L226 correctly lets the unresolved edge through so a later pass can fix it:

```python
clean_edges = [e for e in edges if e["source"] in seen_ids and
               (e["target"] in seen_ids or e["relation"] == "imports")]
```

…but no later pass ever does. The cross-file resolver registry in `extract.py` (~L3173) registers Swift, Python, Ruby, TypeScript, C++ and ObjC — **there is no Elixir resolver**, so `.ex`/`.exs` never gets a resolution pass and the exemption becomes a permanent leak.

That is also why Python is unaffected — the same repro in Python resolves correctly:

```
pkg_web --imports--> pkg_accounts_accounts     # valid
pkg_web_index --calls--> pkg_accounts_accounts # valid
```

## Suggested fix

Register an Elixir cross-file resolver that rewrites unresolved `imports` targets to the node id of the module with that name. Matching on the **normalized module name of module-level nodes** is exact and needs no heuristics — `normalize(label) == target`.

Worth stressing: matching on **id suffix** instead is not safe. I tried it first on the real project and it mis-resolved badly — a bare target like `application` (from `use Application`) latches onto any node whose id ends in `_application`, including function nodes such as `..._oauthapplications_update_application`. Exact module-name matching gave 382/382 unambiguous resolutions on the same corpus where suffix matching produced 7 wrong ones.

Endpoints with no matching module are genuinely external (stdlib and hex: `Logger`, `Application`, `Gettext`, `Ecto.Query`, `Phoenix.Component`, `Ash.Resource`). On the project above that was 86 distinct modules / 859 edges — small enough to emit as leaf stub nodes, which keeps "who imports `Ecto.Query` directly" answerable without vendoring `deps/`.

For reference, on the 900-file project the corrected resolution gave:

| | before | after |
|---|---|---|
| dangling endpoint edges | 3,445 | 0 |
| graph edges | 13,156 | 16,578 |
| `Boundary.Accounts` degree | 75 | 199 |
| communities (Louvain) | 966 | 666 |

`Boundary.Accounts` was under-reported by 62%, and ~300 communities were spurious fragments that rejoined once their import edges existed. Because god-node ranking and community detection both run on the degraded edge set, the report's headline conclusions were wrong, not merely incomplete.

## Secondary observation

In the Elixir repro above, `Accounts.list_users()` produces **no `calls` edge at all** — `extract_elixir()` returns `raw_calls: []`, so cross-module calls are never recorded even before resolution. Python emits `pkg_web_index --calls--> pkg_accounts_accounts` for the equivalent code. This looks like the same missing-resolver gap plus an extractor gap; happy to file it separately if you'd prefer it tracked on its own.

## Environment

- graphify (graphifyy) 0.9.36, Python 3.13, Linux
- Corpus: 865 `.ex`/`.exs` + 39 `.js`, Elixir/Phoenix/Ash

ayushcodes10:
## Summary

Fixes #2556.

The Elixir extractor emitted `imports` edges whose target was the bare normalized module name, while the module node it should point at was registered under a path-prefixed id. The two never match, so `build_from_json` dropped every cross-file `alias`/`import`/`require`/`use` reference. On a real 900-file Elixir/Phoenix project this silently discarded 13% of extracted edges — the entire internal module dependency graph — with no warning.

## Root cause

`graphify/extractors/elixir.py` builds a module's own node id with the defining file's stem (`_make_id(stem, module_name)`) but a reference to that module from anywhere else with just the bare name (`_make_id(module_name)`). The two ids can only ever match when a module refers to itself.

## Fix

A dedicated cross-file resolver (matching the pattern every other member-call/reference resolver in this module already uses) indexes every Elixir module node by its own id and rewrites a dangling `imports` edge's target when it matches exactly one module in the corpus. Per the issue's own testing: id-suffix matching (`application` latching onto `..._update_application`) produced wrong resolutions on a real corpus, so this matches the full id — already exact, since both sides run through the same id builder — rather than any fragment of it. A name matching zero or 2+ modules is left exactly as extracted; the existing external-reference handling turns an unresolved bare id into a leaf stub, the correct outcome for a genuinely external (stdlib/hex) name.

## Test plan

- New `tests/test_elixir_import_resolution.py`: the issue's own repro (a module aliased from another file) now resolves to a real, non-dangling node; a same-file reference (which already worked) is unaffected; two different files declaring the same bare module name correctly yield no resolution; a genuinely external module (e.g. `Logger`) stays unresolved rather than being fabricated.
- Full suite: 5597 passed, 68 skipped, 2 deselected, no regressions.


safishamsi:
Shipped in **v0.9.63** (on PyPI). Your commits were cherry-picked with authorship preserved. On merge I added two guards so the resolver doesn't over-reach: only top-level modules are indexed (so a nested `defmodule` can't capture an unrelated `use <Name>`), and a same-file reference is left unresolved so it can't clobber the `contains` edge — both carried through incremental rebuilds. Thanks @ayushcodes10!

PR #3605, opened 2026-09-16
fix: resolve workspace imports, stylesheet tokens and generic handler bindings; record what a scan excluded

Freekvdijk12:
## The problem

Five ordinary questions about a monorepo, and what the graph answered before this change. Ground
truth built with greps over `git ls-files`, so it does not move with the tool under measurement;
corpus 9 048 files (8 570 code), one npm workspace with three apps and a shared UI package, one .NET
service, one documentation tree.

| Question | before | after | ground truth |
|---|---|---|---|
| Which source files do two apps both use? | 0 | **87**, 0 extra | 87 |
| Which files does one design token reach? | 1 | **62** | 62 |
| Which handler serves this request type? | 5 pairs | **685 `handles` edges**; 3 of 3 spot-checked request→handler paths resolve end to end | 685 rows (635 request pairs + 47 notification pairs) |
| What structure do the documents have? | 0 nodes | **8 037** | — |
| Why is a file missing from the graph? | unanswerable | `graphify-out/EXCLUDED.tsv` | — |

The 14-file reproduction from the issue gives the same picture without a monorepo. Before / after,
same script, same fixture:

| check | released version | this branch |
|---|---|---|
| `.css` files classified as code | 0 of 2 | **2 of 2** |
| `@demo/ui` import lands on | `packages/ui/dist/index.js` (outside the scanned corpus) | **`packages/ui/src/index.ts`** |
| css `@import` edges | 0 | **1** |
| `uses_token` edges | 0 | **6** |
| `handles` edges | none | **`GetThingQuery -> GetThingQueryHandler`** |
| `graphify-out/EXCLUDED.tsv` | absent | **4 rows** |

The right-hand column was re-measured on this branch (`uv run --frozen bash repro.sh <dir>` at
`c221e27`): `2 of 2`, `index.ts` (the `src/` twin, while `dist/index.js` is present on disk), `1`,
`6`, `GetThingQuery -> GetThingQueryHandler`, `4 rows`.

## What changed, and how to check each part

**1 · `.css` joins `CODE_EXTENSIONS`** (`graphify/detect.py`). One token in one set literal.

```bash
python -c "from pathlib import Path; from graphify.detect import classify_file; print(classify_file(Path('theme.css')))"
# before: None        after: FileType.CODE
```

**2 · `detect()` writes the exclusion register it already computes** (`graphify/detect.py`,
`graphify/extract.py`). Three kinds, one row each, tab separated: `ignored` (ignore rule or noise
dir), `unclassified` (with the extension that fell through), `skipped` (sensitive pattern). The
"partially extracted after a parse error" notice, which until now existed only as a line on stderr
during the run, becomes a fourth kind, `partial`, in the same file.

```bash
python -c "from pathlib import Path; from graphify.detect import detect; detect(Path('.'))"
head -5 graphify-out/EXCLUDED.tsv
# before: no such file
# after:  kind  path  reason
#         ignored  …/ignored.py           ignore-rule
#         ignored  …/packages/ui/dist/    noise-dir
#         unclassified  …/notes.xyz       .xyz
```

**3 · A workspace import resolves to the source, not to the build output**
(`graphify/extractors/resolution.py`, 20 lines). `development` joins `_EXPORT_CONDITION_PRIORITY`
— it was in no position in that tuple, so an exports map naming only `development` and `default`
resolved to the `default` catch-all — and
`_source_twin_of_build_output()` maps a `dist/` candidate onto its extensionless `src/` counterpart
and prefers it when that file exists — including when the build output itself is present, which a
fallback that only triggers on a missing file does not cover. In the fixture the `exports` map names
`./dist/index.d.ts` and `./dist/index.js`, the second of which exists on disk; the edge lands on
`packages/ui/src/index.ts` all the same. This is the fix that carries most of `0 → 87` in the table:
`dist/` is pruned as a noise dir, so before the change every edge into a shared package pointed
outside the corpus. The last two of those 87 files arrive through the stylesheet `@import` and
preset import edges from fixes 4 and 5. Check with the fixture — the line `'@demo/ui' import lands on`.

**4 · A CSS extractor** (new `graphify/extractors/css.py`, 78 lines). It emits the stylesheet's own
node, a node per custom property it declares, a `uses_token` edge per `var(--x)` reference, and an
`imports` edge per `@import`, resolved through the existing JS import resolver so a bare package
specifier in a stylesheet lands on a real file.

**5 · Design tokens become nodes, and references bind to them** (new
`graphify/extractors/design_tokens.py`, 112 lines; `_resolve_design_token_refs()` in
`graphify/extract.py`; `uses_token` added to `DEFAULT_AFFECTED_RELATIONS` in `graphify/affected.py`).
A reference from another file — `var(--brand)` in a component, or a Tailwind utility class whose
preset entry resolves to that variable — becomes an edge to the node that declares the token. This
is what turns the blast radius of a token change from 1 file into 62:

```bash
graphify affected "--brand" --relation uses_token --depth 1
# before: No affected nodes found.
# after (on the measured corpus, for one named color token): 61 files listed
```

**6 · A request type binds to the generic handler that handles it**
(`graphify/extractors/engine.py`, `graphify/extractors/csharp.py`, plus the relation names in
`graphify/extract.py` and query hints in `graphify/serve.py`). A small table maps a generic
interface to (type-parameter position, relation name): `IRequestHandler`/`INotificationHandler` →
`handles`, `IPipelineBehavior` → `intercepts`, `AbstractValidator` → `validates`,
`IEntityTypeConfiguration` → `configures`. The edge carries `context="type_binding"` alongside the
ordinary `references` edge, so the "this class mentions that type" signal is not lost.

**7 · The runbook feeds documents to the extractor too** (`tools/skillgen/fragments/core/*.md`, and
the generated skill artefacts). Part A collected only `detect['files']['code']`, so a Markdown file
never reached `extract()` even though a structural extractor for it exists; the semantic pass was
never meant to be the only path for those files. Part A now also walks the `document` category,
filtered to files `_get_extractor()` actually covers.

```bash
grep -c "detect.get('files', {}).get(" graphify/skill.md
# before: 1        after: 2
python -c "import json; g=json.load(open('graphify-out/graph.json')); \
print(sum(1 for n in g['nodes'] if str(n.get('source_file','')).endswith('.md') and n.get('_origin')=='ast'))"
# before: 0        after: 8037   (on the measured corpus)
```

`graphify/skill*.md` is generated: the three core fragments are the edit, `python -m tools.skillgen`
plus `--bless` regenerated the 16 skill files and their snapshots, and `gen.py` gained one predicate
so `--monolith-roundtrip` can sanction those lines. All five skillgen gates are green.

## Cost

Measured on the 9 048-file corpus (8 570 code, 464 document, ~7.7M words), full standard build:

| | |
|---|---|
| graph after | 90 004 nodes · 244 928 edges |
| added by the document category | +8 037 nodes · +10 147 edges |
| added by the token and binding relations, measured on an undirected rebuild | +12 nodes · 5 064 relation edges (`uses_token` 4 781, `validates` 239, `handles` 44, `configures` 0, `intercepts` 0) |
| net edge growth for those relations | +4 767 — 283 of the new edges replaced an existing `references` edge on the same pair, 14 replaced a `contains` edge |
| the same binding relation, built directed | `handles` 685 instead of 44 (see "choices" below) |
| added by the import fixes | +7 import edges |
| `graph.json` on disk | +1.0 % |
| full build, extraction cold, no cache reuse | 10 min 38 s (detect 258 s, extract 224 s, build/cluster/analyse 78 s) |
| rebuild before vs after the new relations | ≈11 min → 11 min 14 s |
| `graphify affected`, three warm runs | +0.1 to 0.2 s |

Source change: +397 / −7 lines across 9 files in `graphify/` (two of them new, 190 lines
together), +230 lines of tests in 8 files, +31 / −4 in `tools/skillgen/`. The remaining 32 changed
files are generated skill artefacts and their snapshots. 53 files, +754 / −43 in total
(`git diff --shortstat v8..HEAD`).

## What this does not do

- **No new CLI surface.** No flag, no command, no changed output format other than the new
  `EXCLUDED.tsv` artefact in `graphify-out/`.
- **`graphify/extractors/__init__.py` is untouched.** The CSS extractor is registered in the
  extension→extractor dispatch table in `extract.py` only, not in `LANGUAGE_EXTRACTORS`; the
  design-token module is not a language extractor at all (it exposes no `extract_*`). Where you want
  these two registered, and whether the second belongs under `extractors/`, is your call — say the
  word and I will move them and add the `MIGRATION.md` row.
- **`ARCHITECTURE.md`, `CHANGELOG.md`, `extractors/MIGRATION.md` and `uv.lock` are untouched.** The
  changelog convention wants an issue number per line, so those lines belong in the same commit as
  the issue reference once this has one.
- **The build-output fallback covers `dist/` only**, mapped to `src/`, for extensionless targets and
  for the `.d.ts` / `.d.mts` / `.d.cts` family. `build/`, `lib/` and `out/` are not handled.
- **The CSS extractor is CSS**, not Sass or Less, and `@import` must be a quoted string; `@import
  url(...)` is not matched.
- **Class names composed at runtime are skipped on purpose** — a `bg-${variant}` template is not
  guessed. On the measured corpus that was 2 files.
- **Nothing about undirected builds.** See the first item under "choices" — this matters.

## Choices locked in, that a reviewer should agree with or reject

- **The `handles` edge is stored request → handler**: source is the request type, target is the
  handler class. That reads backwards against the English verb, and it is deliberate, not a reversed
  edge: it makes "what runs when this request is dispatched" a forward traversal from the request,
  which is the direction users ask in. `validates`, `configures` and `intercepts` keep the natural
  direction (the validator/configuration/behaviour class is the source). If you want `handles` to
  read like the verb, it is the swap in `_resolve_csharp_type_references` and nothing else.
- **In an undirected, non-multigraph build only one edge survives per node pair.** A new
  `handles`/`validates` edge collides with the `references` edge the same generic argument already
  produced. On the measured corpus, the request→handler naming survived in 5 of 635 pairs until the
  graph was built directed; built directed it is 685 edges. The extractor is right either way — the
  loss is in the build step — but the value of this fix depends on a directed build, and that is
  worth saying out loud rather than discovering later.
- **Only `uses_token` was added to `DEFAULT_AFFECTED_RELATIONS`.** The four C# binding relations are
  not in the default traversal, so a default blast radius is unchanged unless `--relation` names
  them.
- **The binding table is a fixed map of five interface names**, matched on the bare generic name at
  type-argument position 0 with exactly one resolved reference. An unrelated library that happens to
  name an interface the same way will match. There is no configuration surface for it.
- **The Tailwind bridge keys on a file literally named `tailwind-preset.ts`.** A design system that
  does not use that convention still gets the `var(--x)` half, and nothing else.
- **`--x-rgb` companion tokens fold onto `--x`** when both are declared in the same stylesheet. On
  the measured corpus this merged 7 token nodes, and inside the declaring stylesheet 21 pairs that
  used to be `contains` ("this file declares this token") became `uses_token` ("this file uses it").
  That is a behaviour change on existing graphs, not only an addition.

## Two known limitations, not fixed here

**The two writers of `EXCLUDED.tsv` anchor differently.** `detect()` writes to `cache_root or root`,
following the convention the word-count cache already uses, so an `extract --out` run keeps it out
of the scanned corpus. The `partial` rows in `extract()` anchor to `root` instead, which for the CLI
is the scanned project (`cli.py` hands `extract()` `{"cache_root": out_root, "root": target}`). With
`--out` the two halves land in two different files, and the second one writes into the corpus being
analysed. Reproduced with `root` and `cache_root` deliberately pulled apart: the register appears
under the scan root and not under the out dir. The new test does not catch it because it passes only
`cache_root=tmp_path` with paths under it, so the two coincide. I left it as is rather than guess
which anchor you want — say which, and I will make both writers use it and pin the choice with a
second test case.

**The `extract()` writer appends.** It opens the register in append mode and writes the header only
when the file is absent, so two `extract()` runs over the same tree list the same partially
extracted file twice. A full `detect()` pass rewrites the file atomically and clears the
duplication, but a path that only re-extracts (`graphify watch`, repeated `update`) keeps growing.
Same choice as above: it becomes a rewrite the moment you say which anchor and which lifetime the
file should have, and it gets a test in the same round.

## Test plan

11 tests, in `tests/`, on the layout CI runs (`uv run --frozen pytest tests/ -q --tb=short`, Python
3.10 and 3.12):

```
tests/test_detect.py::test_classify_stylesheet
tests/test_detect.py::test_detect_writes_exclusion_register
tests/test_extract.py::test_partial_extraction_writes_exclusion_register
tests/test_js_import_resolution.py::test_workspace_subpath_export_development_outranks_default
tests/test_js_import_resolution.py::test_workspace_dist_subpath_falls_back_to_src_twin_when_dist_is_not_built
tests/test_css_extraction.py::test_css_import_resolves_to_target_file
tests/test_design_tokens.py::test_cross_file_var_reference_binds_to_its_definition
tests/test_design_tokens.py::test_uses_token_is_traversed_by_affected
tests/test_csharp_type_resolution.py::test_csharp_request_handler_generic_binds_request_to_handler
tests/test_serve.py::test_query_graph_text_type_binding_alias_resolves
tests/test_skillgen.py::test_part_a_also_walks_the_document_category
```

Each test fails when its own production change is reverted: one fix at a time, in a throwaway
worktree, with the test files left untouched — 11 of 11 go red, and the worktree was verified clean
between rounds. Two commits carry two independent halves each, and there the tests split the same
way: removing only `development` from `_EXPORT_CONDITION_PRIORITY` reddens
`…development_outranks_default` while `…falls_back_to_src_twin…` stays green, and removing only the
`_source_twin_of_build_output` call does the reverse; the same holds for the `detect()` and
`extract()` halves of the exclusion register.

Suite, same machine, same lockfile, one environment per side
(`uv run --frozen pytest tests/ -q`):

```
with these commits (c221e27):    6 failed, 5618 passed, 97 skipped
on the base commit (c7ec108):    6 failed, 5607 passed, 97 skipped
```

The six failures are identical by name on both sides and pre-existing. Four in
`tests/test_ollama_retry_cap.py` need an optional extra that a default `uv sync` does not install
(`ModuleNotFoundError: No module named 'openai'`; CI installs with `--all-extras`, this run did
not). `test_collect_files_skips_hidden` asserts over `Path.parts` of an absolute path, so it fails
whenever the checkout itself sits under a dot-directory. `test_ts_normalizer_scales_linearly_on_large_files`
is a wall-clock ratio test (`0.0770s -> 0.2859s`, threshold 3x) that fails under machine load.
None is touched here. The difference of 11 passing tests is exactly the 11 above — checked by
collecting both sides separately (`pytest tests/ -q --collect-only`, 5717 vs 5706 items) and
diffing the node ids: 11 added, none removed, none renamed.

Also green: `ruff check --config pyproject.toml .` (both sides), `python -m tools.skillgen --check`
(`134 artifact(s) match committed output and expected/`), `--audit-coverage`, `--schema-singleton`,
`--monolith-roundtrip`, `--always-on-roundtrip`, and `graphify --help`. `bandit -r graphify -ll`
reports the same metrics block as the base commit (Low 102 · Medium 8 · High 3). `pyright` reports
`583 errors, 4 warnings` on both sides, and with paths and line numbers normalized away the two
diagnostic sets are identical — this branch adds no new one.

## A note on reviewing this

Two of the nine changed files under `graphify/` are new (`extractors/css.py`,
`extractors/design_tokens.py`). If you verify a build by hashing an installed package against its
`RECORD`, an added file produces no mismatch — only the seven modified ones show up. So the hash
diff is not the full change surface here; the file list above is.

---

# C · Verzendcommando's

Niets hiervan is gedraaid — de tak staat lokaal, er is geen fork, geen push, geen PR, geen release.

```bash
REPO=Graphify-Labs/graphify
WORK=/home/CorporateTrailer/CorporateTrailer-Workspace/.claude/upstream-graphify/repo
BRANCH=feat/head-port          # zeven commits boven v8 (c7ec108)

# 0 · wie ben je, en staat de tak zoals je hem wilt verzenden
gh auth status
git -C "$WORK" log --oneline v8..HEAD
git -C "$WORK" status --short

# 1 · auteursadres van de zeven commits herschrijven (het huidige adres draagt een projectnaam)
cd "$WORK"
git config user.name  "<jouw naam>"
git config user.email "<jouw adres>"
git rebase v8 --exec 'git commit --amend --reset-author --no-edit'
git log v8..HEAD --format='%an <%ae>' | sort -u

# 2 · bijwerken als de standaardtak intussen is opgeschoven; de standaardtak is v8, niet main
#     (de tak staat al ÓP v8; er is geen rebase --onto meer nodig, en bij een stilstaande
#      origin/v8 is dit een no-op)
git -C "$WORK" fetch origin v8
git -C "$WORK" rebase origin/v8 "$BRANCH"

# 3 · de poorten opnieuw draaien ná stap 1/2, want een nieuwe basis is een nieuwe poort
cd "$WORK"
uv sync
uv run --frozen ruff check --config pyproject.toml .
uv run --frozen python -m tools.skillgen --check
uv run --frozen python -m tools.skillgen --audit-coverage
uv run --frozen python -m tools.skillgen --schema-singleton
uv run --frozen python -m tools.skillgen --monolith-roundtrip
uv run --frozen python -m tools.skillgen --always-on-roundtrip

# 4 · issue eerst — de changelogregels van dit project dragen een kwestienummer
#     (zet tekst A in issue.md)
gh issue create --repo "$REPO" --title \
  "Monorepo edges go nowhere: shared-package imports, stylesheet tokens and handler bindings are missing" \
  --body-file issue.md

# 5 · fork en push
cd "$WORK"
gh repo fork "$REPO" --clone=false --remote=true --remote-name fork
git remote -v                                        # 'fork' moet naar jouw eigen kopie wijzen
git push -u fork "$BRANCH"

# 6 · PR openen tegen v8 (zet tekst B in pr.md)
gh pr create --repo "$REPO" --base v8 --head "<jouw-github-handle>:$BRANCH" \
  --title "fix: resolve workspace imports, stylesheet tokens and generic handler bindings; record what a scan excluded" \
  --body-file pr.md
```

---

# D · Vóór je verzendt

1. ~~**Reparatie 3 uitkleden.**~~ **Gedaan** op `feat/head-port`: `19dd101` bevat alleen nog
   `development` in `_EXPORT_CONDITION_PRIORITY` plus `_source_twin_of_build_output` met zijn
   aanroep (samen twintig regels, `git show 19dd101 -- graphify/extractors/resolution.py`). De
   `_resolve_export_targets`-hunks van de oude tak zijn geschrapt; de kop heeft die zelf.
2. **Kwestienummer.** Elke changelogregel in dit project eindigt op `(#<nummer>, thanks @<handle>)`.
   Vandaar issue vóór PR (stap 4); daarna de changelogregel in de `(unreleased)`-sectie toevoegen.
3. **Eén onderwerp per PR is de huisregel** van dit project voor extractors. Overweeg te splitsen:
   (a) classificatie + CSS-extractor, (b) ontwerptokens, (c) generieke typeparameter,
   (d) uitsluitingsregister, (e) werkruimte-imports. De PR-tekst hierboven dekt alle vijf; bij
   splitsen knip je per kop.
4. **Register-plaatsing beslissen** (`LANGUAGE_EXTRACTORS`, `MIGRATION.md`-rij) — de PR-tekst laat
   die keuze expliciet aan de onderhouder; als je liever meteen levert, doe het vóór verzending.
5. **`EXCLUDED.tsv` noemen in `README.md`** bij de uitvoerbestanden, als je het artefact wilt laten
   staan.


Freekvdijk12:
Closing this — opened without the graph system's owner having agreed to it. The work stays local; if it is offered again it will be after that conversation.

PR #3607, opened 2026-09-16
fix(python): lexical-scope-aware indirect_call resolution

seonghobae:
## Summary

- \`_emit_indirect_by_name\` (the callback-by-name resolver behind the
  Python \`indirect_call\` edge) matched the referenced identifier purely by
  bare label: first via the per-file \`label_to_nid\` map — which #3405
  deliberately excludes nested Python functions from — then via
  \`extract.py\`'s corpus-wide cross-file resolver. Neither consults lexical
  scope.
- A callback name that only exists as a **nested** function therefore fell
  straight through to the cross-file resolver, which matches by bare label
  across the whole corpus and can bind to a same-named nested function
  defined in a **sibling** function's scope — or to an unrelated
  module-level definition that the reference's own enclosing scope actually
  shadows.
- Direct \`calls\` resolution already walks the \`scope_parents\` /
  \`lexical_nids_by_scope\` chain built for #3405 (innermost enclosing scope
  first — Python's LEGB name-resolution rule, per the Python Language
  Reference's "Execution model: Resolution of names"). This PR applies the
  same walk to indirect_call resolution *before* falling back to the flat /
  cross-file maps, so a lexically-resolved name is trusted directly and is
  never re-adjudicated by the corpus-wide pass.
- No behavior for cross-file or module-level callback references changes —
  those still fall through exactly as before when no enclosing-scope
  definition exists.

## Repro

Two sibling functions each define + pass a same-named nested
\`read_then_mutate\` callback:

```python
def run_with_guard(fn):
    return fn()

def test_alpha_scope():
    def read_then_mutate():
        return "alpha"
    return run_with_guard(read_then_mutate)

def test_beta_scope():
    def read_then_mutate():
        return "beta"
    return run_with_guard(read_then_mutate)
```

Before this fix, \`graphify update . --no-cluster\` emits:

```
test_alpha_scope -> test_alpha_scope_read_then_mutate   (correct)
test_beta_scope  -> test_alpha_scope_read_then_mutate   (WRONG — sibling's def)
```

After this fix, both indirect_call edges bind to their own lexical
definition.

Originally observed in a downstream consumer repo
(ContextualWisdomLab/fast-mlsirm#1862) while auditing a graph run over two
Python test modules that each nest a `read_then_mutate` helper inside a
same-named sibling test function.

## Test plan

- [x] Added \`tests/test_issue_1862_indirect_call_lexical_scope.py\`
      (sibling same-named nested callbacks each bind to their own
      definition; a legitimate cross-file callback reference still
      resolves; an enclosing lexical scope's definition is preferred over a
      module-level same-named definition per LEGB).
- [x] \`tests/test_issue_3405_python_resolution.py\` (the related direct-call
      lexical-scope suite) still passes unchanged.
- [x] Full suite: \`pytest tests/ -q\` → 5379 passed, 26 failed, 257 skipped.
      All 26 failures are pre-existing/environmental on this checkout and
      unrelated to this change (missing optional \`tree-sitter-hcl\` extra
      cascading through \`test_terraform*.py\`, Ollama SDK retry-cap
      assumptions, a wheel-packaging test, an unrelated mtime-collision
      flake, and one markdown frontmatter test) — none touch
      \`extractors/engine.py\`'s Python call/indirect-call resolution.

https://claude.ai/code/session_01U6G3WJY9PjB8k2fHhkvB14


seonghobae:
## Addressed: lexical callback lookup ignores parameter shadowing

Pushed `5f37509` on this head (`fix/1862-lexical-scope-indirect-call-v8`).

**Fix.** In `_emit_indirect_ref`, after the current-scope `enclosing_locals` guard, walk `scope_parents` innermost-first:
- if the name is a nested callable in that enclosing scope (`lexical_nids_by_scope`), stop and let `_emit_indirect_by_name` resolve it
- if the name is any other local binding there (`local_bound_names`, which includes parameters), return without emitting — the closure closes over that binding, not a same-named module/corpus callable

**Regression.** `test_python_indirect_call_enclosing_parameter_shadows_module` covers `outer(callback)` → nested `caller` → `dispatch(callback)` and asserts no `indirect_call` edge to the module-level `callback()`.

**Verification (official `uv` path):**
- `uv run --frozen pytest tests/test_issue_1862_indirect_call_lexical_scope.py tests/test_indirect_dispatch.py -q` → **32 passed**
- `uv run --frozen ruff check graphify/extractors/engine.py tests/test_issue_1862_indirect_call_lexical_scope.py` → **All checks passed**
- `uv run --frozen graphify update .` running (AST-only; `graphify-out/` is gitignored; pre-update `GRAPH_REPORT.md` was absent in this checkout)

Please re-review the advisory finding against `5f37509`.

seonghobae:
## Re-review triage on `5f37509`

Original advisory (**Lexical callback lookup ignores parameter shadowing**) is addressed: the new Graphify summary now states that an enclosing parameter/local binding shadows same-named module/corpus callables, and `test_python_indirect_call_enclosing_parameter_shadows_module` stays green.

On the three new **Worth a look** advisories (gate still **PASS**, Graphify check **SUCCESS**, none verified):

1. **Python method bare names via class scope** — reproduced (`.run()` → `.helper()` when no module `helper` exists). Empirically **pre-existing** on `fa73e0f` without the enclosing-shadow hunk; driven by flat `label_to_nid` fallback, not the new `scope_parents` shadow walk (`scope_parents` is only populated for nested functions). Out of scope for this shadowing fix; happy to take as a follow-up if maintainers want it gated here.
2. **`global` declarations** — `global name` with no assignment still resolves to the module callable; `global` + assignment currently suppresses the edge (false negative). Same: advisory / follow-up unless you want it in this PR.
3. **Duplicate innermost handling** — intentional split: current-scope param/local guard in `_emit_indirect_ref`, then parent walk for enclosing shadows; `_emit_indirect_by_name` still does full LEGB for callables. No functional change planned.

Requesting maintainer merge when ready; contributor cannot merge into `Graphify-Labs/graphify`.

seonghobae:
Non-author review requested. I am the change author, so I cannot be the independent reviewer.

Exact head: `a719529d9aba373ae80628b7b5832567d8736cf8`
PR base: `v8` @ `c7ec1082083e3e876443f643ecf86ebcfae177c2` (current `origin/v8` has since moved to `4c735618f3d56fd622c2049771584621c31ba9ff`; merge-base with this branch is `c7ec108`)
Diff vs that base: `graphify/extractors/engine.py`, `tests/test_issue_1862_indirect_call_lexical_scope.py` — 454 insertions, 20 deletions.

Local reproduction, not CI: `uv run --frozen pytest tests/test_issue_1862_indirect_call_lexical_scope.py tests/test_indirect_dispatch.py tests/test_indirect_dispatch_getattr.py -q --tb=line` → 43 passed. This does not stand in for the CI matrix.

CI on this head: Actions run 35838804630 (`pull_request`) concluded `action_required` with zero jobs, so there are no job logs and no test result. Fork Actions runs for this SHA: 0. Please approve that workflow if the diff looks right. I will not bypass checks, force-push, or merge.

PR #3612, opened 2026-09-16
fix(csharp): extract past a verbatim interpolated string ending in an escaped quote (#3376)

### LINKED ISSUE #3376: C#: a verbatim interpolated string ending in }""" is rejected by the bundled grammar — file partially extracted, warning points at the recovery line
## Summary

In a **verbatim interpolated** string (`$@"…"` / `@$"…"`), an escaped quote `""` that sits
immediately after an interpolation hole's closing `}` **and** immediately before the string
terminator is rejected by the bundled `tree-sitter-c-sharp` 0.23.5 grammar. In short: the
literal ends with `}"""`.

It is valid C#, `dotnet build` is happy, and every neighbouring spelling parses. The file it
appears in is partially extracted, so the members after the offending line are missing from
`graph.json`.

The reported line is the recovery point, not the cause. In my case the construct is on lines
72 and 83 and the warning said `first error at line 76`, which sent me looking at an
`[InlineData("Config")]` attribute for a while.

## Reproduction — grammar level

`graphify 0.9.46`, `tree-sitter-c-sharp 0.23.5`. No extraction run needed:

```python
from tree_sitter import Language, Parser
import tree_sitter_c_sharp as tscs
p = Parser(Language(tscs.language()))

W = 'class C { void M(string e) { var s = %s; } }\n'
def check(lit):
    print("ERROR" if p.parse((W % lit).encode()).root_node.has_error else "ok   ", lit)

check(r'$@"a""{e}"""')   # ERROR  <-- valid C#, the real-world shape
check(r'$@"a{e}"""')     # ERROR
check(r'$@"{e}"""')      # ERROR
check(r'@$"a""{e}"""')   # ERROR  (other prefix order, same failure)

check(r'$@"a{e}""b"')    # ok     escape after the hole, but not at the end
check(r'$@"a{e}b"""')    # ok     escape at the end, but text between hole and escape
check(r'$@"{e}=""x"""')  # ok     hole and a trailing escape, hole not adjacent to it
check(r'$@"a""{e}""b"')  # ok     escape, hole, escape - not at the end
check(r'$@"a""x"""')     # ok     escape at the end, no hole at all
check(r'$@"a{e}"')       # ok     hole, no escape
check(r'@"a{e}"""')      # ok     no `$`, so `{e}` is literal text
check(r'$"a\"{e}\""')    # ok     regular interpolated string
check(r'$$"""a={{e}}"""')# ok     raw interpolated string
```

Each passing line differs from a failing one in exactly one property, so the trigger is
neither "verbatim interpolated string" nor "escape at the end" nor "escape plus hole" on its
own — it is specifically the `}` `""` `"` adjacency.

## What it costs

Small but silent-ish, and it hides in a construct that is idiomatic wherever a test asserts on
a quoted attribute:

```csharp
Assert.Contains($@"Content=""{expected}""", xaml);   // XAML / XML / HTML attribute needle
```

On a 258-file C#/WPF repo it is 1 file — but that one file is the project's UI-language guard,
and the agent instructions for this repo tell the assistant to query the graph before reading
source. A guard that is only partially in the graph is worse than one that is absent, because
nothing signals the gap at query time.

The `#2551` warning did its job here: without that line I would have had no reason to suspect
the parser at all. Same class as #2551 (Kotlin one-line type bodies) and #2520 (`.luau`),
different language and trigger.

## Workaround

A regular interpolated string with backslash escapes is byte-identical in output and parses:

```csharp
Assert.Contains($@"Content=""{expected}""", xaml);   // trips the grammar
Assert.Contains($"Content=\"{expected}\"", xaml);    // same string, parses
```

## Environment

| | |
|---|---|
| graphify | 0.9.46 |
| tree-sitter | 0.26.0 |
| tree-sitter-c-sharp | 0.23.5 |
| OS | Windows 10 22H2 |
| Python | 3.12 |

rohit-jsfreaky:
Fixes #3376

## What

A C# file containing a verbatim interpolated string whose text ends with an interpolation hole, an escaped quote and then the terminator is now extracted in full. Before this, the bundled grammar rejected the literal and every member declared after that line was silently missing from the graph.

## Why

The spelling is valid C# — `dotnet build` accepts it — but tree-sitter-c-sharp 0.23.5 rejects it. As the issue shows, every neighbouring spelling parses; the trigger is specifically the `}` `""` `"` adjacency at the end of the literal. 0.23.5 is the newest release on PyPI and the project already allows `>=0.23,<0.25`, so there is no grammar bump to take.

The cost is silent. The file still extracts, just not all of it, and the warning names the parser's recovery line rather than the literal — the issue reports chasing an unrelated `[InlineData]` attribute because of it.

## How

`extract_csharp` now passes the source through `_normalize_csharp_verbatim_interpolation`, the same `source_override` seam `extract_js`, `extract_vue` and `extract_cpp` already use.

Only the trailing escape is overwritten, with two spaces. The rewrite is **byte-length preserving** — nothing is inserted or deleted and no line break is touched — so every offset, line and column still points at the same place in the file on disk and reported `source_location` values stay accurate (the #2876 contract that `_normalize_cpp_cli` documents). The interpolation hole itself is kept, so identifiers referenced inside it still resolve.

Two gates keep it inert everywhere else:

- a file that already parses is returned untouched, so healthy files never enter the rewrite;
- if the masked source still has parse errors the original is used, leaving today's behaviour exactly as it is.

The scanner walks the file once, skipping comments, char literals, raw strings and every other string form, so the spelling written inside a comment or another string is never mistaken for code.

## Testing

`tests/test_csharp_verbatim_interpolation.py`, 7 tests.

Before, on a class with six methods and the literal in the third:

```
warning: Config.cs (first error at line 20, 5 symbol(s) extracted)
extracted: BeforeOne()  BeforeTwo()  Offender()  Config  Config.cs  Demo
missing:   AfterOne()   AfterTwo()   AfterThree()
```

After: 10 nodes, all six methods, no warning.

Byte-level check on that same file — 783 bytes, 39 newlines:

```
same total length:   True
same newline count:  True
bytes changed:       2, at offsets 422 and 423 (both on line 20): b'"' -> b' '
every other byte:    identical
line number for every sampled offset: unchanged
```

and each member's reported location matches the line it is really on (L7, L12, L17, L24, L29, L34).

The two behavioural tests fail with the `source_override` hook removed and pass with it; the other five exercise the normalizer directly (length preservation, multi-line literals, two literals in one file, a healthy file left untouched, and the spelling appearing only inside comments).

Full suite: 16 failed, 5565 passed, 194 skipped. Those 16 are pre-existing — re-running the same ten files with this change reverted gives the identical 16 failed / 477 passed, and none of them touch C# or extraction. All 159 existing C# tests pass. `ruff check` is clean on both files.

## Scope

This does not fix the grammar, which is where the bug actually is — upstream tree-sitter-c-sharp should accept that spelling. It stops graphify silently dropping symbols while the grammar is what it is, and steps aside the moment a file parses cleanly. If you would rather wait for an upstream fix instead of carrying this, that is a fair call.


rohit-jsfreaky:
Follow-up pushed for the review finding.

**Quadratic prefix scan — confirmed, then fixed.** I measured it before changing anything, on a file that is a run of `$` characters:

| N | time (before) |
|---|---|
| 500 | 0.013s |
| 1,000 | 0.051s (3.99x) |
| 2,000 | 0.216s (4.24x) |
| 4,000 | 0.834s (3.86x) |

Doubling N quadrupled the time, so the finding was right, and the path is reachable — the scan only runs on files the grammar already rejected, which is exactly what such a file is. The cause was mine: when no string opened at a position I advanced a single byte, so a long `$`/`@` run was re-walked from every offset.

`_csharp_string_opener` now always reports where the prefix run ended, even on a miss, and the caller resumes there. After: ~1.8x per doubling, and 1,000,000 `$` characters take **0.097s** (the same input would have taken hours before).

Added an eighth test that feeds it 1,000,000 characters with a 5-second ceiling, so it cannot regress quietly. The end-to-end behaviour is unchanged: the repro file still yields 10 nodes with all six methods.

Separately, correcting a number in my own PR description above: I wrote "All 159 existing C# tests pass". That 159 counted this PR's own test file. The real split is **153 existing + 8 added here = 161**. Apologies for the inflated figure.

PR #3613, opened 2026-09-16
Resolve C++ scope-qualified static calls when the method exists only as a qualified fallback node

### LINKED ISSUE #2348: C++ static calls to UCLASS methods are not linked (calls edge dropped) -> affected under-reports
## Summary

For C++, a scope-qualified static call `Class::Method(...)` is **not** linked with a `calls` edge when the callee method node exists **only in qualified form** (label `Class::Method()`). This happens systematically for Unreal Engine `UCLASS()` / `GENERATED_BODY()` classes: the macro-laden header declaration is not parsed, so the method node is created only from the `.cpp` definition (`Class::Method`), with a qualified label. The call resolver matches call-sites by the **bare** callee identifier, so it never matches the qualified-labeled node, and the edge is dropped.

Consequence: `graphify affected "<Class>"` misses every static-call caller of that class's methods — dangerous for impact analysis / refactor gating (an "only affects X" that is incomplete).

**Environment:** graphify 0.9.13 · tree-sitter-cpp 0.23.4 · Python 3.10 · Windows

### Case A — plain class (WORKS, control)

`helper.h` declares `static Helper* FindNearest();`, `helper.cpp` defines it, `caller.cpp` calls `Helper::FindNearest()`. Result: the method node is labeled **`FindNearest`** (bare, from the header) and the qualified call links correctly:

```
useQualified() --[calls]--> FindNearest   ✅
```

### Case B — Unreal UCLASS (BUG)

```cpp
// helper.h
#pragma once
#include "CoreMinimal.h"
#include "helper.generated.h"
UCLASS()
class AHelper : public AActor {
    GENERATED_BODY()
public:
    UFUNCTION()
    static AHelper* FindNearest(UWorld* W);
};

// helper.cpp
#include "helper.h"
AHelper* AHelper::FindNearest(UWorld* W) { return nullptr; }

// caller.cpp
#include "helper.h"
void useIt(UWorld* W) {
    AHelper* h = AHelper::FindNearest(W);   // qualified static call
    (void)h;
}
```

- Node `AHelper::FindNearest()` (label **qualified**, sourced from `helper.cpp` — the header declaration inside `GENERATED_BODY()` was not extracted).
- Edges from `useIt()`: `--[references]--> AHelper`, `--[references]--> UWorld`.
- **No `calls` edge** `useIt() --> AHelper::FindNearest()`. ❌

The only difference between A and B is whether a **bare-named** method node exists. When the method exists only as a qualified node, the call is not linked.

## Expected

`useIt() --[calls]--> AHelper::FindNearest()` should exist, and `affected "AHelper"` should list `useIt` as an impacted caller.

## Suggested fixes (either resolves it)

1. **Call resolver:** when a call callee is a `qualified_identifier` (`Class::Method`), match it against method nodes labeled `Class::Method()` (and/or by the bare method segment), not only against bare-named nodes.
2. **Header parsing:** handle `UCLASS()` / `GENERATED_BODY()` / `UFUNCTION()` macro bodies so method declarations produce bare method nodes (as in the plain-class case).

## Real-world impact

On an Unreal project, `affected "AXylosHelperCharacter"` returned **1** caller where `git grep` found **3** real `Class::FindNearest(...)` static call-sites — so the reported blast radius was 1/3 of the true one.

ayushcodes10:
## Summary

Fixes #2348.

A scope-qualified static call `Class::Method(...)` was not linked with a `calls` edge when the callee method node existed only in qualified form (label `Class::Method()`). This happens systematically for Unreal Engine `UCLASS()`/`GENERATED_BODY()` classes.

## Root cause

Traced this to the bundled tree-sitter-cpp grammar's own error recovery, not graphify's extraction logic. `GENERATED_BODY()` (an unknown macro invocation to the grammar) followed by `public:` and a pointer-returning method declaration produces an ambiguous parse: the grammar's recovery heuristic can successfully resync on an unambiguous token (e.g. `int FindNearest(...)`, where `int` is a `primitive_type` keyword), but for a self-referential pointer return type (`AHelper* FindNearest(...)`, where `AHelper` is just an `identifier`) the recovery mis-attributes the whole declaration as a bogus constructor-initializer chain instead — so the in-class member declaration is silently dropped, and the out-of-line `.cpp` definition has nothing to attach to. The extractor's existing behavior in that case is to fall back to a node labeled with the qualified name (`AHelper::FindNearest()`), attributed to its *file* rather than its *class*.

`_resolve_cpp_member_calls`'s `method_index` only ever indexes `defines`/`method` edge targets, so it can never find that qualified-only node — the call resolves via `Foo::bar()`'s exactly-one-type-candidate path but then has no method to land on, and the edge is silently dropped.

## Fix

A grammar-level fix (teaching the bundled tree-sitter-cpp to parse Unreal's reflection macros) is out of scope for a Python extraction layer. Instead, a second-chance lookup: when the primary `method_index` lookup for a `Foo::bar()` call comes back empty, check for a node whose label is exactly the qualified form (`"Foo::bar()"`) — still guarded by the same exactly-one-candidate rule every other resolver in this module already applies, so two unrelated classes that happen to share both a class name and a method name (and both hit the same parse failure) still correctly yield no edge.

## Test plan

- New tests in `tests/test_cpp_objc_cross_file_calls.py`: the issue's own Unreal-style repro now resolves to an EXTRACTED `calls` edge; two unrelated same-named classes both hitting the qualified-only fallback correctly yield no edge (god-node guard).
- Full suite: 5623 passed, 68 skipped, 2 deselected, no regressions.


safishamsi:

PR #3619, opened 2026-09-16
Fix: save_manifest can erase a deleted file's row before it is ever reported (#3426)

### LINKED ISSUE #3426: detect_incremental derives deleted_files from scan membership, not existence — save_manifest has the exists() check that its own comment says detect_incremental needs, and the placement is backwards in both directions
`save_manifest` filters the manifest by `Path(f).exists()` "so `detect_incremental()` treats them as gone" — but `detect_incremental` never applies that filter, and the placement is backwards in both directions.

**Version:** graphifyy 0.9.15, macOS (darwin), Python 3.11. 0.7.13 has no existence filtering in `save_manifest` at all, so this appears to be something learned between versions.

## The inconsistency, in the code's own words

`detect.py`, in `save_manifest`:

```python
# Seed from the existing manifest so incremental callers passing a subset
# of files don't silently erase entries for untouched files (#917).
# Prune entries whose file no longer exists on disk — those are genuine
# deletions that detect_incremental() should treat as gone.
...
        if Path(f).exists():
            manifest[f] = normalised
```

The comment names `detect_incremental()` as the beneficiary. `detect_incremental` derives the list itself with no existence test:

```python
# Files in manifest that no longer exist - their cached nodes are now ghost nodes
current_files = {f for flist in full["files"].values() for f in flist}
deleted_files = [f for f in manifest if f not in current_files]
```

That comment states the premise the expression does not test. `current_files` is **what this scan saw**, not what is on disk. The two are the same only when consecutive runs have identical scan scope.

`deleted_files` then flows to `build_merge(prune_sources=deleted_files or None)` (`cli.py`, and `__main__.py` on 0.7.13), where it removes nodes, drops carried hyperedges, and removes edges — and where passing it also **skips the #479 shrink guard**, whose skip condition is `not prune_sources` with the comment "shrinkage is intentional there". So a list that means *out of scan* is read as *deleted*, and passing it disables the check that would have questioned the removal.

## Two reproductions

Both use `detect` directly with an explicit `manifest_path` on a throwaway corpus.

**1. Out-of-scan is reported deleted while every file is on disk.** Two runs with different scan scope — which happens whenever two environments resolve different `graphifyy` versions, since `CODE_EXTENSIONS` is 61 entries on 0.7.13 and 93 on 0.9.15 (a strict superset, `.json` and `.sh` among the additions):

```
manifest after wide run : ['a.py', 'b.py', 'c.json']
deleted_files reported  : ['c.json']
all still on disk       : True
```

Those nodes are then pruned, and the run reports it as drift corrected.

**2. A genuine deletion becomes unreportable, which is the direction the filter was added to fix.** A save that is not followed by a prune — a scan-only run, an interrupted pipeline, or a merge step that saves the full current scan:

```
manifest after run 1    : ['a.py', 'b.py']      b.py then deleted from disk
deleted_files this pass : ['b.py']              correct, caught on this pass
manifest after its save : ['a.py']              save_manifest's exists() filter drops it
deleted_files NEXT run  : []                    unreportable from here on
```

Within a single `--update`, `detect_incremental` runs before `save_manifest`, so the ordinary path still catches a genuine deletion — this needs a save unfollowed by a prune. Noting that up front so the second case is not dismissed as unreproducible.

So the filter **cancels** the genuine-deletion cleanup at the site where it sits, and is **absent** at the site where it would prevent the false-deletion loss.

## Related, and why this is not those

- **#3153** — `prune_sources` raw-string matching, so `--update` *never* prunes deleted files. Opposite direction; 0.9.15's normalisation into `prune_set` addresses it.
- **#1796** — `prune_set -= new_sources`, exempting re-extracted files. Reads like the fix here and cannot apply: out-of-scan files are absent from `new_chunks` precisely *because* they are out of scan.
- **#1574** (hyperedge carry-forward), **#479** (shrink guard), **#917** (seed-from-existing) — each is correct given that `prune_sources` expresses intent. All three inherit whatever `deleted_files` gets wrong.

## Options, not a recommendation

I have not chosen between these because #917's rationale is yours, not mine:

1. **Move** the existence filter from `save_manifest` to the `deleted_files` derivation. Fixes both directions and is the smallest diff — but with the filter gone from the seed, a genuinely deleted file stays in the manifest and is reported deleted on every subsequent run (verified: three runs, same `deleted_files`, manifest never shrinks). #917's seed keeps it alive and no scan ever contains it.
2. **Add** the filter at the derivation and leave `save_manifest` as-is. Stops the false-deletion loss; keeps the case-2 erasure.
3. **Both, sequenced** so the derivation sees the entry before the save drops it. Correct in both directions, more than a relocation.

Happy to open a PR for whichever you prefer.

ayushcodes10:
## Summary
Fixes #3426. `save_manifest` could prune a manifest row for a file that no longer exists on disk *before* any caller had a chance to observe that deletion via `detect_incremental`'s `deleted_files` report — making the deletion permanently unreportable from the very next run onward, whenever a save happened without also pruning the graph (a scan only run, an interrupted pipeline).

## Root cause
The reported issue described two cases:

1. Out of scan files wrongly reported as deleted — this turned out to already be fixed by an unrelated earlier improvement (#1908), which is unmodified in this PR and already distinguishes "excluded but alive" from "genuinely deleted" on the `detect_incremental` read side via a direct `Path(f).exists()` check.
2. Genuine deletions becoming permanently unreportable after certain save patterns — this was still live on current code. `save_manifest` pruned any row whose file no longer existed on disk unconditionally and immediately, regardless of whether `detect_incremental` had run yet on that same state. A scan only run or an interrupted pipeline that calls `save_manifest` without also acting on the deletion would silently erase the evidence before anything downstream ever saw it.

## Fix
Removed the unconditional `Path(f).exists()` prune from `save_manifest`. The existing scan corpus based prune (already used to drop "excluded but alive" rows, per #1908) still cleans up a genuinely deleted row when the caller passes the full scan corpus — a deleted file always falls outside that scan, on path alone — but now this happens naturally sequenced *after* `detect_incremental` has had the chance to report the deletion on that same scan. Only a partial/subset save (no scan corpus) now leaves a dead row in place; the next full scan reconciles it.

## Test plan
- Added `test_save_manifest_subset_save_keeps_a_deleted_files_row` in `tests/test_detect.py`, reproducing the exact sequence from the issue: save, delete a file, confirm it is reported once, save again without the full scan corpus, confirm the row survives and the deletion is still reportable on the next pass.
- Confirmed the pre-existing `test_save_manifest_full_scan_still_prunes_missing_file` (scan corpus based prune) still passes unmodified.
- Full suite: `python3 -m pytest -q` — 5622 passed, 68 skipped, 2 deselected, no regressions.
- `python3 -m tools.skillgen --check` — OK.


ayushcodes10:
Good catch — fixed. `_in_root(f)` alone left an out-of-root row (a merged/foreign corpus entry, or one where the root couldn't be resolved) permanently unprunable, even across full scans, since it fails open for exactly those paths. Out-of-root rows must still never be pruned merely for being outside the current scan (that's a deliberate, pre-existing #1908 protection — detect never walks them, so absence from the corpus alone isn't exclusion evidence), but that's a different condition from the file actually being gone from disk. A full scan already gives `detect_incremental` its chance to report the deletion first, the same safe reconciliation point the in-root case relies on, so a full scan now also prunes an out-of-root row whose file no longer exists — a partial save is untouched, exactly as before. Two regression tests added covering both the fixed case and the still-protected boundary.

PR #3620, opened 2026-09-16
Refuse to spawn an extraction pool that would spawn its own (#1637)

### LINKED ISSUE #1637: [Windows] Skill's inline python snippets fail under PowerShell 5.1 (quoting) and Windows multiprocessing (missing __main__ guard); minor: query exit code, fuzzy-match ties
**Version:** graphifyy 0.9.5 · **Platform:** Windows 10, PowerShell 5.1, Python 3.14

## 1. PowerShell 5.1 quoting breaks the inline `python -c` snippets

Steps 3–5 of SKILL.md embed multi-line Python with escaped double quotes (`\"nodes\"`), written for bash. Under Windows PowerShell 5.1 the escapes are mangled before reaching Python (`SyntaxError: unterminated string literal`). Suggestion: for the Windows variant, instruct the agent to write each step to a temp `.py` file and execute that, instead of `python -c`.

## 2. `extract()` requires an `if __name__ == "__main__":` guard on Windows

`graphify.extract.extract()` uses multiprocessing. Windows uses the spawn start method, so any caller script without a `__main__` guard re-executes on every worker spawn → a cascade of `spawn_main` tracebacks and no extraction. Worked once the runner script was wrapped in `def main(): ... if __name__ == "__main__": main()`. The SKILL.md snippets don't mention this; the Windows variant should either wrap its snippets or document the guard.

## 3. Minor observations from the same test run (209-file TypeScript corpus)

- `graphify query "..."` prints a correct, useful subgraph but exits **255** even on success — breaks scripted `$LASTEXITCODE`/`&&` chaining.
- `graphify path A B` with human-readable names failed with `warning: target match was ambiguous (top score X, runner-up X)` and reported "No path found", while the same call with exact node IDs returned a correct 3-hop path. When fuzzy scores tie exactly, consider listing the tied candidates and asking, rather than silently picking one that may be disconnected.
- Community labeling step assumes a hand-labelable number of communities; this corpus produced 125. A documented top-N-plus-heuristic-tail strategy would help.

ayushcodes10:
## Summary
Fixes #1637. On Windows, a caller script with no `if __name__ == "__main__":` guard turns `extract()`'s parallel path into a fork bomb rather than a slow failure: every spawned worker re-executes the top-level module on import, and if that module calls `extract()` again at module scope (the common shape of a quick inline runner script), the worker opens its own `ProcessPoolExecutor`, whose own guard-less children do the same. A second reporter measured this directly — ~1,278 `python.exe` processes and ~28 GB consumed before the machine had to be power-cycled.

## Root cause
`_extract_parallel` already caught `BrokenProcessPool` and fell back to sequential extraction, but only *after* the pool tried to run. `ProcessPoolExecutor._adjust_process_count()` respawns dying workers as they die, and on a guard-less caller every respawned worker dies the same way (re-importing `__main__` triggers `_check_not_importing_main`) and — critically — re-executes the module-level `extract()` call, opening yet another pool. The pool keeps minting doomed processes faster than any per-future exception can propagate and stop it, so the reactive `except BrokenProcessPool` handler never gets a chance to run.

## Fix
Two checks now run in `_extract_parallel` *before* a `ProcessPoolExecutor` is ever constructed:

1. **Unconditional**: refuse to open a pool when already running inside a multiprocessing child (`multiprocessing.parent_process() is not None`). A legitimate call to `_extract_parallel` only ever happens in the main process — a worker should only ever run `_extract_single_file` via the pool machinery — so this is always safe and breaks the specific recursive-spawn mechanism regardless of platform.
2. **Windows only**: pre-emptively decline the pool when the caller's own `__main__` module's source lacks the `__main__` guard text, instead of discovering the failure only after the pool has already started respawning. Scoped to `sys.platform == "win32"` (matching the existing Windows-specific worker cap already in this function) since macOS/Linux typically default to fork, where this class of guard-less-caller re-execution doesn't occur the same way, and scoping it avoids changing extraction behavior on non-Windows CI.

Both checks fall back to sequential extraction exactly like the existing `BrokenProcessPool` handler — correct output, just without the pool.

## Test plan
- Added `test_extract_parallel_declines_pool_inside_a_spawned_worker`, `test_extract_parallel_declines_pool_on_windows_when_caller_lacks_guard`, and `test_extract_parallel_still_spawns_pool_on_windows_when_caller_has_guard` to `tests/test_extract.py`.
- Confirmed all pre-existing Windows-spawn fallback tests in the same file still pass unmodified.
- Full suite: `python3 -m pytest -q` — 5624 passed, 68 skipped, no regressions.
- `python3 -m tools.skillgen --check` — OK.


ayushcodes10:
Good catch — fixed. The guard check used a bare substring test, so any mention of `__main__` anywhere in the caller's source (a comment, docstring, or unrelated string literal) was read as a real guard. It now matches the actual `if __name__ == "__main__":` statement (either operand order), with regression tests for both the false positive case and the reversed operand order.

ayushcodes10:
Good catches — fixed both. The regex approach had two real gaps: a guard-shaped line sitting inside a triple-quoted string or docstring example was still read as a real guard (the exact false positive class this check exists to close, just needing more specific bait text), and a valid parenthesized comparison was wrongly rejected as no guard at all.

Switched from a regex to parsing the caller's source with `ast` and looking for a real `if` statement whose test compares `__name__` to `"__main__"` in either order. The parser never treats string or comment contents as code at all, and parens are transparent to it, so both gaps close at once. An unparseable source falls back to the same best-effort treatment as an unreadable file. Regression tests cover all three reported gaps plus the fallback.

ayushcodes10:
Good catch — fixed. `ast.walk()` finds a matching `if` statement anywhere in the tree, including one nested inside an unrelated function, class, or dead branch — which never actually runs at import time and protects nothing. A caller whose real module-scope code was genuinely unguarded could still be reported as safe, letting the exact fork-bomb scenario this check exists to prevent happen anyway.

The guard idiom only has its intended effect as a bare top-level statement, so now only the module's direct top-level statements (`tree.body`) are checked, not every node anywhere in the source. Regression test covers a guard nested inside an unrelated function alongside a genuinely unguarded module-scope `extract()` call.

ayushcodes10:
Good catch — fixed. A module can have a real top-level guard AND a separate, genuinely unguarded top-level statement that calls `extract()` outside it. Finding *some* guard anywhere in the module wasn't enough — the specific statement that led to this call has to sit inside one, or the exact fork-bomb this check exists to prevent still happens through the unguarded statement.

Now walks the call stack for the outermost frame belonging to the module's own top-level code: its current line is wherever the chain of calls that reached `extract()` started, checked directly against the guards' line ranges. Two regression tests actually execute the caller script as a top-level module (a monkeypatched `__file__` alone doesn't exercise this, since the check needs a real call stack) — one with a call site outside the guard, one with a call site properly inside it via an intervening function call.

ayushcodes10:
Good catch — fixed. The "already inside a worker process" decline returned silently, unlike the sibling "caller lacks a guard" branch just below it which prints a warning. Both are equally valid reasons to fall back to sequential extraction and should both be visible the same way, not one being a quiet fallback. Added the same kind of stderr diagnostic, with a test asserting it's printed.

ayushcodes10:
Good catch — fixed. Gating on \`sys.platform == "win32"\` missed macOS, which has defaulted to the spawn start method since Python 3.8 — the exact same fork-bomb this check exists to prevent is fully reproducible there too, not just on Windows.

The gate now asks \`multiprocessing\` directly whether opening a pool here would use spawn (the only start method where a missing guard matters at all — fork/forkserver never re-execute top-level code), using \`get_start_method(allow_none=True)\` so checking it doesn't have the side effect of permanently locking in the platform default for a caller that hasn't made its own \`set_start_method()\` call yet. Two new tests: macOS+spawn still declines for a guard-less caller, and fork (Linux's default) still takes the pool path even without a guard, since it needs no guard at all.

ayushcodes10:
Good catch — fixed. \`if __name__ == "__main__" and verbose:\` is a valid, real guard (every operand of \`and\` must be true for the body to run, so it only executes when the name genuinely is \`"__main__"\`), but was rejected since only a bare comparison was accepted — this only caused an unnecessary sequential fallback, not anything unsafe, but it's a real gap worth closing. \`_is_main_guard_test\` now recurses through \`and\`, never through \`or\` (which is NOT safe to recognize this way — the body can run without \`__name__\` genuinely being \`"__main__"\` if the other side of the \`or\` is true). Two tests cover both directions.

On the third finding ("equivalent caller path spellings bypass validation"): I looked at this too. \`main_file\` (from \`sys.modules["__main__"].__file__\`) and the module frame's \`co_filename\` are set from the same source at script-compile time for a normal \`python script.py\` invocation, so they should be identical in the common case; I couldn't construct a concrete scenario (symlink, case-sensitivity, etc.) where they'd genuinely diverge for a real caller without also changing how Python itself resolves the script path. If a mismatch did occur, the code already falls through to "can't tell, assume safe" — consistent with this function's existing philosophy for every other uncertain case (unreadable file, unparseable source, frame not found), not a new decision. Left as-is without a concrete repro to fix against, but flagging it here in case someone has one.

Issue #1619, opened 2026-07-03
Windows SKILL.md field review (v0.9.5): broken interpreter guard, backslash/encoding corruption, ghost flags — with verified fixes


edtrackai:
**Environment:** Windows 11 Pro, Claude Code (Git Bash + PowerShell 7 available), `graphifyy` 0.9.5 installed via `uv tool install`, skill at `~/.claude/skills/graphify/` (`name: graphify-windows`).

I did a line-by-line review of the Windows skill template (SKILL.md + references/) and verified the riskiest findings on a real machine. Two failures reproduce today on a stock Windows + uv setup; the rest are latent Windows bugs, doc/spec mismatches, and two ghost flags the skill references but never implements. I've patched all of these locally and validated the fixes — happy to turn this into a PR if you want it.

Findings are grouped by severity. Line numbers refer to the shipped v0.9.5 `SKILL.md` (712 lines) and reference files.

---

### A1. Interpreter guard is broken on Windows (SKILL.md:649-661)

The subcommand guard recovers a lost `graphify-out/.graphify_python` by parsing a shebang:

```bash
GRAPHIFY_BIN=$(which graphify 2>/dev/null)
PYTHON=$(head -1 "$GRAPHIFY_BIN" | tr -d '#!')
```

On Windows the entry point is `graphify.exe` — a PE32+ binary (verified with `file`). `head -1` reads `MZx...` garbage, the character-class guard catches it, and the fallback writes `python3` — which on a typical Windows box resolves to the **Microsoft Store WindowsApps stub**, not a real interpreter (verified: `which python3` → `.../WindowsApps/python3`). Net effect: delete `graphify-out/` once and every `query` / `path` / `explain` / `--update` is broken until a full rebuild, with a confusing Store-stub error.

**Fix (validated):** don't parse shebangs; probe real interpreters the same way Step 1 does — `uv tool dir` → `pipx environment` → active env, each gated on `"$PY" -c "import graphify"` (the import check also filters out the Store stub). Bash version below works on Git Bash and Linux/macOS:

```bash
find_graphify_python() {
    if command -v uv >/dev/null 2>&1; then
        UV_DIR=$(uv tool dir 2>/dev/null | tr -d '\r' | tr '\\' '/')
        for PY in "$UV_DIR/graphifyy/Scripts/python.exe" "$UV_DIR/graphifyy/bin/python"; do
[...]

PR #3621, opened 2026-09-16
Fix #1619 skill findings: no-cluster/force, INPUT_PATH slashes, Step 1 gate, backup order, interpreter quoting

ayushcodes10:
## Summary
Fixes five findings from #1619's Windows skill review:

- **C1/C2 (ghost flags)**: `--no-cluster` (Step 2's own suggestion) and `--force` (the shrink-guard error's own suggestion) were referenced but never implemented in the generated skill's Python. Both are real `graphify` CLI flags already (see `cli.py`), so this wires the skill's inline Python through to the same underlying support.
- **B1 (INPUT_PATH backslash corruption)**: a Windows path substituted into `INPUT_PATH` with backslashes splices a stray escape into the Python string literal it lands in (`\t` becomes a tab, `\U` raises a `SyntaxError`); the PowerShell `Resolve-Path` call that saves the scan root was also unquoted.
- **B4 (no Step 1 failure gate)**: if the interpreter install still failed, nothing checked for it — the step wrote the still-broken interpreter's path anyway, and every later step failed with a cryptic error far from the real cause.
- **C4 (backup instruction out of order)**: the `--update` runbook's "save the old graph" instruction appeared AFTER both the merge block and the diff block that reads the backup, so an agent following it top to bottom never created the backup before the merge ran, and the post-update diff silently no-opped on every single update.
- **B5 (unquoted interpreter substitution)**: the saved-interpreter substitution (`$(cat graphify-out/.graphify_python)`) was spliced in unquoted at 100+ call sites, unlike the already-quoted `"$PYTHON"` form Step 1 itself uses — an interpreter path containing a space word-splits into multiple arguments and fails to exec.

(A2/A1 from the same issue were already fixed; C3 is fixed by the already-open #2782, cross-referenced on the issue. B2/B3/C5/D remain open — B3 turned out to already be resolved by the current architecture, since the whole Windows core is now translated to consistent PowerShell rather than mixing shells.)

## Fix
- Two new substitution variables, `IS_NO_CLUSTER` and `IS_FORCE`, follow the pattern already established for `IS_DIRECTED`. Step 4 branches to a single "Full Corpus" community when `--no-cluster` was given instead of calling `cluster(G)`; both `to_json()` calls pass `force=IS_FORCE`; Step 5 documents skipping labeling when there's nothing to label.
- One rule near the top of every host tells the agent to substitute `INPUT_PATH` with forward slashes on Windows; the PowerShell `Resolve-Path` call is now quoted.
- Step 1 re-checks the interpreter import right after the install attempt and stops with an actionable error instead of continuing with a broken interpreter path.
- The `--update` runbook's backup instruction now appears right before the merge block it belongs with.
- Every occurrence of the interpreter substitution across the core fragment, five shared reference fragments, and both monoliths is now quoted, plus the PowerShell translator's match pattern.
- All applied to the core fragment, the shared shell/reference fragments, and both hand-maintained monoliths (`aider.md`, `devin.md`), with new sanctioned-diff predicates in `gen.py`'s monolith round-trip guard for each deliberate change.
- Added `--no-cluster`/`--force` to the `## Usage` block.

## Test plan
- New tests in `tests/test_skillgen.py`: `test_no_cluster_and_force_flags_are_wired_through`, `test_input_path_forward_slash_guidance_is_present`, `test_step1_gates_on_a_still_failed_install`, `test_update_backup_instruction_precedes_the_merge_it_backs_up`, `test_interpreter_cat_substitution_is_quoted_everywhere` — each covering the core render, `skill-windows.md`, and both monoliths.
- Updated pre-existing literal-string assertions across `tests/test_skillgen.py`, `tests/test_community_labels_skill.py`, and `tests/test_devin.py` that pinned pre-fix output verbatim.
- `python3 -m tools.skillgen --check` / `--audit-coverage` / `--monolith-roundtrip` / `--schema-singleton` — all OK.
- Manually ran the substituted Python for `--no-cluster`/`--force` end to end against a real extraction, the actual generated Step 1 bash block under a fake failing interpreter/PATH, and a quoted interpreter invocation directly in bash.
- Full suite: `python3 -m pytest -q` — 5626 passed, 68 skipped, no regressions.


ayushcodes10:
Good catch — fixed. A build with clustering skipped (`--no-cluster`) makes Step 4's write the only one for that path, since Step 5 (which normally supplies real community labels) is skipped entirely. Step 4 already computed the "Full Corpus" placeholder label right above the write but never passed it through, so every node in that build silently ended up with no `community_name` at all. Now passes `community_labels=labels` through Step 4's `to_json()` call too, applied to the shared core fragment and both hand maintained monoliths, with a regression test.

ayushcodes10:
Verified the INPUT_PATH shell injection findings too — they're real (the Step 1 \`echo "\$(cd INPUT_PATH && pwd)"\` line is unquoted, and bash still expands \`\$(...)\`/backticks inside double quotes). This is pre-existing on pristine v8 though, not introduced by this PR, and it's in the shared \`shell/posix.md\`/\`shell/powershell.md\` fragments used by essentially every platform — a much broader, foundational fix than this PR's scope. Filed separately as #3642 with a suggested fix direction, rather than folding a wide, security sensitive rewrite into this PR.

PR #3627, opened 2026-09-17
Index JavaScript in inline <script> blocks of a PHP file (#2320)

### LINKED ISSUE #2320: Inline <script> JS in .php files is not indexed (extract_php skips embedded script blocks, unlike extract_svelte/astro/vue)
## Summary

`.php` files that mix PHP with an inline `<script>` block only get their PHP half indexed. Any JavaScript declared inside `<script>…</script>` is silently absent from the graph, so `query` / `explain` / `path` cannot see it.

This matters because the skill instructs assistants to prefer graphify over reading source. On a classic server-rendered PHP app, a large share of the behaviour lives in inline `<script>` blocks — so the graph looks complete while a whole layer is missing, and there is no warning to say so.

## Reproduction

Three files:

```php
// page.php
<?php
function phpSideHelper(): string { return 'i am php'; }
?>
<div id="app"></div>
<script>
function inlineJsHelper(x){ return x * 2; }
function inlineJsCaller(y){ return inlineJsHelper(y) + 1; }
</script>
```

```js
// plain.js
function plainJsHelper(x){ return x * 2; }
```

```php
// plain.php
<?php
function plainPhpHelper(): string { return 'also php'; }
```

Then `graphify update .` → `Rebuilt: 6 nodes, 3 edges, 3 communities`.

Node ids present:

```
page
page_phpsidehelper
plain_js_plain
plain_plainjshelper
plain_php_plain
plain_plainphphelper
```

| symbol | where | indexed |
|---|---|---|
| `plainPhpHelper` | `plain.php` | yes |
| `phpSideHelper` | `page.php`, PHP part | yes |
| `plainJsHelper` | `plain.js` | yes |
| `inlineJsHelper` | `page.php` `<script>` | **no** |
| `inlineJsCaller` | `page.php` `<script>` | **no** |

So both parsers work in isolation; only the embedded region is dropped. The `inlineJsCaller → inlineJsHelper` call edge is lost with them.

## Cause

`extract_php` delegates straight to the generic tree-sitter path with no handling for embedded regions:

```python
def extract_php(path: Path) -> dict:
    """Extract classes, functions, methods, namespace uses, and calls from a .php file."""
    return _extract_generic(path, _PHP_CONFIG)
```

tree-sitter-php's `php` grammar treats the non-PHP spans as opaque text, so the JS inside them is never parsed.

## Suggested fix

The machinery already exists for this exact shape — `extract_svelte` and `extract_astro` both regex the embedded regions out and feed them to the JS parser, with the same reasoning in their docstrings ("Tree-sitter only sees the `<script>` block"):

```python
script_re = _re.compile(r"<script\b[^>]*>([\s\S]*?)</script\s*>", _re.IGNORECASE)
```

Applying the same pass in `extract_php` after the PHP extraction — union the two symbol sets, offsetting line numbers by the match position so nodes point at the right lines — would cover it, and `.phtml` / `.php3-7` / `.phps` alongside. Two details specific to PHP that Svelte/Astro don't hit:

- `<script src="…"></script>` with no body should be skipped rather than yielding an empty region.
- Inline blocks frequently contain PHP echo tags inside the JS (`const MODE = <?= $x ? 'true' : 'false' ?>;`). Substituting those spans with a placeholder before parsing keeps the region syntactically valid JS; otherwise the parse fails and the whole block is lost again.

Happy to test a patch against a large real-world PHP codebase if useful.

## Environment

- graphifyy 0.9.9 (uv tool install), Python 3.10
- tree_sitter 0.25.2, tree_sitter_php 0.24.1, tree_sitter_javascript 0.25.0
- macOS (darwin 25.5.0)

ayushcodes10:
## Summary
Fixes #2320. \`.php\` files that mix PHP with an inline \`<script>\` block only had their PHP half indexed — any JavaScript declared inside \`<script>…</script>\` was silently absent from the graph, so \`query\`/\`explain\`/\`path\` couldn't see it. This matters for classic server-rendered PHP apps, where a large share of behavior lives in inline \`<script>\` blocks.

## Root cause
\`extract_php\` delegates straight to the generic tree-sitter path with no handling for embedded regions. \`tree-sitter-php\`'s grammar treats non-PHP spans as opaque text, so JS inside them is never parsed.

## Fix
Reuses the masking pattern \`extract_vue\` already established for the same class of problem (an embedded \`<script>\` region a host grammar can't see into): after the normal PHP extraction, everything outside \`<script>\` bodies is blanked to spaces (preserving \`\r\`/\`\n\` so line numbers stay accurate), and the masked full source is parsed with the JS grammar via \`_extract_generic(path, _JS_CONFIG, source_override=...)\`. The two node sets merge onto the one file node — a symbol declared in the script block looks like any other symbol in the file.

Two edge cases handled:
- \`<script src="…"></script>\` with no body masks to an empty region, which parses as empty JS and contributes nothing — no special-casing needed.
- A file with no \`<script>\` tag at all skips the second pass entirely (no wasted parse).

A PHP echo tag inside the JS region (\`const MODE = <?= $x ? 'true' : 'false' ?>;\`) doesn't crash the JS parser and the rest of the block still extracts correctly, so the placeholder-substitution refinement the issue mentioned as a nice-to-have wasn't needed.

## Test plan
- New tests in \`tests/test_languages.py\`: the issue's exact reproduction (two functions plus the call edge between them), line numbers surviving the masking pass, a \`src\`-only script tag contributing nothing without crashing, and a plain PHP file with no script tag being unaffected (same node/edge shape as before).
- Verified the full 3-file scenario from the issue (\`page.php\` + \`plain.js\` + \`plain.php\`) produces all 5 expected symbols plus the cross-block call edge.
- All 23 existing PHP extraction tests still pass unmodified.
- Full suite: \`python3 -m pytest -q\` — 5625 passed, 68 skipped, no regressions.


ayushcodes10:
Good catch — fixed. When a JS symbol's id collided with an existing PHP node (same name, resolved in favor of the PHP node), every JS edge was still being merged in unconditionally, so an edge meant for the discarded JS symbol silently attached to the unrelated PHP node sharing its id. Now edges touching a dropped JS node's id are skipped too, with the file node's own always on collision (both passes deliberately reuse its id, by design) excluded from that so unrelated JS symbols' edges from the shared file node are unaffected. Added a regression test with a same name PHP/JS pair.

ayushcodes10:
Re: the two remaining advisory findings.

**"Bare except swallows all errors, hiding JS extraction failures"**: checked, and this matches an existing, deliberate convention already in this codebase — `extract_vue`'s own secondary best effort pass (the dynamic import rescue) uses the identical bare `except Exception: pass` shape, so a failure in the optional JS side pass degrades to PHP only extraction rather than losing the whole file's result. Not changing this to stay consistent with that established pattern.

**"PHP extractor treats script tags in PHP comments as executable JS"** (`graphify/extract.py:2550`): correct as an observation, but the same accepted limitation the masking approach already carries elsewhere — the regex based Vue script masker this PR explicitly mirrors has the identical gap (it can't distinguish a real script tag from one mentioned inside a comment without a full parse). Given it mirrors an existing, accepted tradeoff rather than introducing a new one, leaving it as is.

safishamsi:

PR #3628, opened 2026-09-17
Name the subagent capability requirement, not just general-purpose (#2525)

### LINKED ISSUE #2525: Step B2 hard-codes subagent_type="general-purpose", making the skill unrunnable on hosts with dispatch policies
# Step B2 hard-codes `subagent_type="general-purpose"`, making the skill unrunnable on hosts with dispatch policies

**Version:** graphifyy 0.9.20
**Host:** Claude Code with an org-level pre-dispatch policy hook
**Severity:** skill is literally unrunnable as written; workaround is non-obvious

## Summary

`SKILL.md` Step B2 states:

> **IMPORTANT - subagent type:** Always use `subagent_type="general-purpose"`.
> Do NOT use `Explore` - it is read-only and cannot write chunk files to disk,
> which silently drops extraction results. General-purpose has Write and Bash
> access which the subagent needs.

The *reasoning* is right — the subagent genuinely needs Write and Bash. The
*instruction* is over-specified: it names one agent type rather than the
capability that matters.

On hosts that govern dispatch — org policy hooks, curated persona registries,
pre-dispatch guardrails — `general-purpose` dispatch can be flagged or refused.
In our environment a policy hook raises `material_dispatch_noncanonical_plugin`
on it, so the skill's literal instruction cannot be followed as written.

## Why this is worse than a normal blocker

Step B3 compounds it. Both recovery messages also name the agent type:

> "chunk N missing from disk — subagent may have been read-only. Re-run with
> general-purpose agent."

and

> stop and tell the user to re-run and ensure `subagent_type="general-purpose"`
> is used.

So when a policy-restricted host hits this, every diagnostic points the user
back at the one dispatch type their host will not permit. The obvious escape —
falling back to `Explore` because it *is* permitted — is the worst possible
outcome: it is read-only, so it silently drops results rather than failing loudly.

## Workaround

Dispatch any policy-permitted agent type that has Write and Bash. We used a
named domain persona with full tool access; extraction behaved identically.
Nothing in graphify depends on the agent's name or domain — only on its tools.

## Suggested fix

State the **capability requirement** and let the host satisfy it:

> The extraction subagent MUST have **Write and Bash access** — it builds the
> chunk file on disk itself. Any read-only agent type (e.g. `Explore`) silently
> drops extraction results.
>
> - **Default:** `subagent_type="general-purpose"`.
> - **If your host restricts or forbids `general-purpose` dispatch** — org
>   dispatch policy, a guardian hook, or a curated persona registry — use any
>   available agent type that has Write and Bash.
> - **Verify before dispatching:** the type must list Write and Bash (or
>   "All tools") in its capability description.
>
> If the host rejects the dispatch with a policy error naming the agent type,
> that is a host policy constraint, not a graphify requirement. Re-dispatch with
> a permitted Write+Bash-capable type. Never downgrade to a read-only type to
> satisfy a policy — that trades a loud failure for a silent one.

The Step B3 diagnostics should be reworded the same way ("re-run with a
Write+Bash-capable agent type").

## Relation to existing issues

- **#808** (closed) is where this instruction comes from: `general-purpose`
  subagents couldn't write to `graphify-out/`, and the resolution was to pin the
  dispatch type. That fixed the sandbox problem, but it encoded a specific agent
  *name* where the actual requirement is a *capability*. This report is narrowly
  about that over-correction — the write-capability requirement itself is right
  and should stay.
- **#1758** (open) shares the Step B3 diagnostic surface: it notes the
  "likely read-only dispatch" message is a plausible-but-wrong explanation for
  output-ceiling crashes. Both point at the same rewording — the diagnostic
  should describe the capability that's missing, not the agent type that's
  expected.

## Scope

The `general-purpose` mandate ships in the canonical `SKILL.md` plus **14**
`skill-*.md` platform bodies, so this affects every packaged platform variant,
not just Claude.

*(Disclosure, matching the convention in #1758: investigation and write-up done
with an AI assistant, based on a real run.)*

ayushcodes10:
## Summary
Fixes #2525. Step B2/B3 of the \`/graphify\` skill named the literal agent type \`subagent_type="general-purpose"\` as the only acceptable choice for the extraction subagent. On a host whose dispatch policy restricts or refuses that exact type — an org policy hook, a curated persona registry — the skill could not be followed as written, and every diagnostic (both the "chunk missing from disk" warning and the "re-run" instruction) pointed the user back at the one type their host would not permit. The reporter's own workaround — falling back to a read-only type like \`Explore\` because it's permitted — is the worst outcome, since it silently drops extraction results instead of failing loudly.

## Fix
Reworded to state the actual requirement (Write and Bash access, since the subagent builds its chunk file on disk itself) with \`general-purpose\` kept as the default, and an explicit escape hatch: a policy-restricted host should dispatch any permitted type with both tools, since that's a host policy constraint, not a graphify requirement. Applied to:

- The shared \`dispatch/agent-tool-disk.md\` fragment (bash hosts) and its PowerShell counterpart — this is the B2 dispatch instruction most split hosts (claude, claw, codex, etc.) render.
- The core fragment's and devin monolith's Step B3 re-run diagnostics (the "chunk missing from disk" warning and the "if more than half failed" instruction).
- Kilo's platform-specific rule (\`subagent_type="general"\`, Kilo's own naming), reworded the same way.

opencode and aider are unaffected — neither ever named \`general-purpose\` in their own dispatch slots (opencode uses \`@mention\`, aider doesn't mandate a subagent type at all).

## Test plan
- New \`test_subagent_dispatch_names_the_capability_not_just_the_type\` in \`tests/test_install.py\`, asserting every host carrying the shared dispatch slot or Step B3 diagnostics states the Write and Bash requirement plus the escape hatch.
- Confirmed the two existing tests that scope-check agent-type mentions (\`test_opencode_skill_uses_opencode_agent_guidance\`, \`test_claw_skill_uses_agent_tool_dispatch\`) still pass unmodified — the concrete dispatch example (\`subagent_type="general-purpose"\`) is untouched, only the surrounding guidance changed.
- \`python3 -m tools.skillgen --check\` / \`--audit-coverage\` / \`--monolith-roundtrip\` / \`--schema-singleton\` — all OK.
- Full suite: \`python3 -m pytest -q\` — 5622 passed, 68 skipped, no regressions.


ayushcodes10:
Good catch — fixed. The sanctioned diff predicate matched any line containing the bare substring, which also matched the dispatch fragments' own worked example showing the subagent type as example syntax, silently weakening the round trip check's ability to catch real corruption of that unrelated line. Now matches only a phrase unique to each of the two actually changed lines' old and new forms, with a regression test asserting both the real forms match and the worked example line does not.

ayushcodes10:
Re: "New test references undefined predicate gen._is_subagent_capability_wording_fix_line" — checked, this is a false positive. The predicate is not new (it predates this fix; this PR only tightened its body), and it's very much defined and reachable:

\`\`\`
$ python3 -c "from tools.skillgen import gen; print(gen._is_subagent_capability_wording_fix_line)"
<function _is_subagent_capability_wording_fix_line at 0x...>
\`\`\`

The full `tests/test_skillgen.py` suite (66 tests, including the one this finding names) passes. No change needed.

Issue #691, opened 2026-05-04
_load_tsconfig_aliases() doesn't follow tsconfig "extends" — SvelteKit/Nuxt/NestJS aliases silently dropped


jippi:
## Problem

`_load_tsconfig_aliases()` (extract.py) reads `compilerOptions.paths` from the
nearest `tsconfig.json` it finds, but does **not** follow the `extends` field.

This breaks alias resolution for any project that uses an extended tsconfig — which
is the modern default for SvelteKit, Nuxt, NestJS, Astro, T3 stack, and most
framework starters. In all of these, the user's `tsconfig.json` typically just
extends a framework-generated file, and the actual `paths` live in the parent.

The #575 fix works perfectly for projects that put `paths` directly in
`tsconfig.json`, but silently no-ops for the (much more common) extended case.

## Repro

SvelteKit project (`graphifyy 0.7.0`).

`tsconfig.json`:

```json
{
  "extends": "./.svelte-kit/tsconfig.json",
  "compilerOptions": { ... }   // no `paths`
}
```

`.svelte-kit/tsconfig.json` (auto-generated by `svelte-kit sync`):

```json
{
	"compilerOptions": {
		"paths": {
			"$lib": ["../src/lib"],
			"$lib/*": ["../src/lib/*"],
			"$partials": ["../src/partials"],
			"$partials/*": ["../src/partials/*"]
		}
	}
}
```

Project code uses these aliases everywhere:

```ts
import { db } from '$lib/server/database'
import RecommendationCarousel from '$lib/components/recommendations/recommendation-carousel.svelte'
```

**Observed:** `_load_tsconfig_aliases()` returns `{}` because the
project-root `tsconfig.json` has no `paths`. Every `$lib/*` and `$partials/*`
import falls into the bare-package branch and gets dropped.

**Impact in numbers** (1,966-file SvelteKit corpus, AST + Kimi semantic):

- 8,088 nodes, 11,179 edges built
- 1,705 communities (extreme fragmentation — most are 1-4 node singletons)
- 192 `.svelte` files appear with zero cross-file edges; ~150 of those are
  verifiably used in the live app (cross-checked with knip + grep)

## Root cause

`extract.py` `_load_tsconfig_aliases()`:

```python
data = json.loads(tsconfig.read_text(encoding="utf-8"))
paths = data.get("compilerOptions", {}).get("paths", {})
# never inspects data.get("extends")
```

## Proposed fix

[...]

PR #3631, opened 2026-09-17
fix(js): resolve tsconfig paths through project references

Estheisia:
## Problem

A solution-style root `tsconfig.json` — `{"files": [], "references": [...]}` — carries no `compilerOptions` of its own. This is what `create-vue` and `create-vite` generate, and what the TypeScript docs recommend for multi-project setups:

```json
{
  "files": [],
  "references": [
    { "path": "./tsconfig.node.json" },
    { "path": "./tsconfig.app.json" }
  ]
}
```

The real `"paths": {"@/*": ["./src/*"]}` lives in `tsconfig.app.json`.

`_find_js_config` stops at the first `tsconfig.json` it finds, and `_read_tsconfig_aliases` follows `extends` chains (#691) but has no notion of project ownership. So `_load_tsconfig_aliases` returns `{}` and **every `@/…` import in the repo resolves to a ghost node** instead of the real file:

```python
>>> _load_tsconfig_aliases(Path('sherpa-ui/src/components/messages'))
{}
```

Effect on the graph: `import { toast } from '@/utils/toast'` mints a stub `ref_utils_toast` with no `source_file` rather than edging the real `src/utils/toast.ts` node. In the app I hit this on, `src/router.js`'s 40 lazy route imports all landed on ghosts, so 42 `.vue` files looked unreferenced when the router clearly imports them.

## Fix

`tsc` never merges a referenced project into the root: it assigns each source file to the project whose `include`/`files` cover it, and only that project's `paths` apply. `_tsconfig_solution_chain` mirrors that — it resolves each `references` entry, computes coverage from `include`/`files` (glob truncated at the first wildcard, plain file entries contributing their parent directory), and returns the covering projects least- to most-specific so merging in order lets the most specific win. When nothing covers the directory, every referenced project is returned in declared order, which keeps the common case of several projects sharing one `@/*` working.

The chain is consulted **only when the root config yields no aliases/baseUrl of its own**, so a config that already resolves keeps its existing behavior.

`exclude` is deliberately ignored — it only narrows a project, and the result picks which project's `paths` apply, not whether a file compiles.

## Measured

384-file Vue 3 + Vite app:

| | before | after |
|---|---|---|
| `imports_from` resolved to a real file | 427/1044 (40%) | 976/1023 (95%) |
| `.vue` nodes nothing points at | 42 of 226 | 0 of 186 |

The 226 → 186 drop is the ghost twins collapsing into the real files; 186 is the repo's actual `.vue` count. The 47 still-unresolved imports are genuine externals (`vuetify/components`, `dayjs/locale/fr`, `node:fs`).

God nodes went from all-external to meaningful, since internal hubs were previously split across ghosts:

```
before:  vuex 128 · vitest 34 · lodash-es 32 · vuetify 30
after:   toastApiError() 268 · vuex 128 · toast() 79 · decodeParams() 55
```

Reproduced independently on a second unrelated `create-vue` app: 361/382 (94%) resolved.

## Tests

`tests/test_tsconfig_solution_references.py`, 7 cases: covering project wins, most-specific wins, uncovered-directory fallback, root-with-its-own-paths unchanged, `baseUrl` from a reference, malformed/missing reference survivable, plain config unchanged.

Full suite green on the CI matrix floor and middle:

```
uv run --frozen pytest tests/ -q     Python 3.10 → 5778 passed, 13 skipped
                                     Python 3.12 → 5777 passed, 14 skipped
uv run --frozen python -m tools.skillgen --check / --audit-coverage /
    --schema-singleton / --monolith-roundtrip / --always-on-roundtrip   all PASS
uv run --frozen ruff check .         All checks passed!
```

`uv.lock` unchanged.


Estheisia:
Closing to reopen from a rewritten branch: the commits carried a personal email address I need off a public history, and the bot's advisory findings are folded in. Superseded PR to follow.

Issue #2948, opened 2026-08-22
Optional bounded LSP semantic-provider layer for compiler-resolved graph evidence


kerberosmansour:
## Problem

Graphify's deterministic AST extractors provide broad, fast language coverage, but some cross-file facts only become reliable after a project's compiler or language service has resolved modules, overloads, generated metadata, and workspace configuration.

The existing LLM semantic pass is valuable for documents and higher-level concepts; it is not a substitute for compiler-reported symbol, reference, implementation, and call-hierarchy facts.

## Proposed direction

Add an **optional, local, bounded language-provider layer** that can enrich (never replace) the native AST graph:

- a small provider contract and deterministic registry;
- a generic stdio Language Server Protocol adapter;
- argv-only process execution (`shell=False`), workspace confinement, time/message/file/request budgets, and explicit `unavailable` / `failed` / `budget_exhausted` outcomes;
- provider/run/range provenance on every imported fact;
- additive merge into a separate output by default;
- reconciliation with native nodes only when `(source_file, label)` has exactly one match; ambiguous symbols remain separate;
- operator-trusted JSON manifests so an additional language server is configuration rather than a new runner implementation.

Initial built-in profiles would cover mainstream application stacks through maintained language servers: Rust, TypeScript/JavaScript, Java, Kotlin, C#, Python, Go, PHP, and Ruby. No language-server binary would be bundled or required for Graphify's normal operation.

## Why this complements the current architecture

- Native tree-sitter extraction remains the always-available source of structural truth.
- Language-server evidence is a typed, local, opt-in enrichment source.
- The provider layer does not change Graphify's LLM backend selection or document-semantic flow.
- It is complementary to #2941: that PR concerns when/how Graphify refreshes its existing semantic pipeline; this proposal concerns an additional evidence source with its own bounded execution contract.
- The same evidence contract could later admit analysis providers, but scanner findings should remain a separate provider kind with separate trust and severity semantics.

## Suggested first PR boundary

1. Provider/run contracts, registry, generic bounded LSP adapter, additive merge, CLI, and documentation.
2. Protocol-level tests for all profiles, plus real-tool smoke evidence for an available subset.
3. No automatic installation, no default execution, no network fetching, no mutation of source files, and no integration into the existing `graphify extract` command in this first change.

## Acceptance checks

- Graphify behaves byte-for-byte as before when the extension is unused.
- A missing or failing language server cannot remove AST nodes/edges.
[...]

Issue #1230, opened 2026-06-10
Option to parse html files as code


mfallouh-TS:
I'm currently trying out graphify with an angular frontend, which has a lot of .html files which are code, not documents.
If I understand correctly, graphify considers that all html files are documents, and runs semantic extraction on them, which not only isn't very useful but also costs a lot of tokens.

Is there a way to instead parse the html files as code ? I couldn't find such option in the docs.
Overall I think a way to customize which extensions are considered what could be a good solution.

LuisPeregrina:
I would just `.graphifyignore` them

mfallouh-TS:
@LuisPeregrina  Yeah that's why I ended up figuring out too, but it would be better to have them included in the graph

Without the html templates, the graph loses a lot of its value

Duran-Kover:
Adding a concrete real-world case for this, since it's more extreme than the Angular-templates example above and might help prioritize #2096.

Repo: a single 1.4 MB `.html` file (Odoo module frontend, ~22,200 lines) with the entire app logic -- 669 top-level `function`/`async function` declarations -- inside one inline `<script>` block. Everything else in the same repo is `.py`/`.js`/`.xml`.

Ran `graphify update .` (local, no LLM key configured). Result: every `.py` controller/model got extracted correctly (confirmed via `graphify affected`, which returned exact matches for a Python function's callers). The `.html` file got zero graph nodes -- not partial, not truncated, literally absent from `graph.json`, even though `manifest.json` shows it was scanned/hashed.

Root cause, confirmed by reading `detect.py` locally: `.html` is hardcoded into `DOC_EXTENSIONS`, so it never reaches the tree-sitter path at all -- unlike `.vue`/`.svelte`/`.astro`/`.razor`/`.cshtml`, which are in `CODE_EXTENSIONS` and get real embedded-script extraction. Since I ran the no-LLM `update` (not a doc-semantic pass), the file got neither treatment: not code-parsed (wrong bucket), not semantically summarized (no LLM configured). For a project where the `.html` file is the app (not markup around a backend template), this means the single most important file is completely invisible to query/affected/path/explain -- a natural-language query for a cross-file frontend-backend concern silently omitted the entire frontend side with no warning that a file had been skipped.

+1 for --html-as-code (#2096) landing -- happy to test it against this repo if useful.

PR #3632, opened 2026-09-17
feat(vue): edge auto-registered components from SFC templates

Estheisia:
> **Stacked on #3631.** This branch contains that commit too; the diff to review here is `b3068ea` only. Once #3631 merges this reduces to a single commit.

## Problem

`extract_vue` masks everything outside `<script>` (`_vue_mask_non_script`, `resolution.py`), so a component registered globally by `unplugin-vue-components` or Nuxt produces **no edge at all** — it is rendered from the template and has no import statement anywhere in the SFC.

On a 186-SFC app, measured against the `components.d.ts` registry:

```
registered local components:       158
template uses of local components: 331
  edged in graph:                    0
  MISSING:                         331  (100%)
```

`graphify affected "ConfirmDeleteDialog.vue"` returned 0 impacted files. `grep -rl "<ConfirmDeleteDialog"` returns 22.

This is not a niche setup — `unplugin-vue-components` is in the default `create-vue` flow, and Nuxt auto-registers `components/` with no plugin at all. Related: #1230 (Angular `.html` treated as documents) is the same shape in another framework, and #2948's LSP layer would subsume this, but neither covers it today.

## Fix

The generated `components.d.ts` is the registry mapping a tag back to its file:

```ts
ConfirmDeleteDialog: typeof import('./src/components/common/ConfirmDeleteDialog.vue')['default']
```

`_load_vue_components_dts` reads it straight off disk — it is commonly gitignored (it is in the `.gitignore` of both apps I tested), so the scanned corpus cannot be relied on. The walk stops at the package root, since the generator writes the file at or under it and anything above belongs to a different project. A second pass in `extract_vue` resolves each template tag through the registry and emits `uses`.

Three details:

- **Relation is `uses`, not a new `renders`** — `uses` is already in `DEFAULT_AFFECTED_RELATIONS`, so `affected` traverses these with no other change.
- **Dedupe mirrors `_rescue_js_dynamic_imports`** — resolve via `_resolve_rescued_specifier` first, then skip when this file's node already points at the target, so a component that is *also* imported explicitly is not edged twice. Recursive components skip themselves.
- **`normpath`, not `resolve()`** — `_emit_rescued_import` joins the specifier onto the SFC path as scanned and normpaths it, so resolving symlinks here would hand it a target it then fails to match.

Both PascalCase and kebab-case spellings are registered, since Vue accepts either in templates.

## Measured

Same 186-SFC app, `--code-only`, no API key:

| | before | after |
|---|---|---|
| template usages edged | 0/331 (0%) | **331/331 (100%)** |
| `uses` edges | 0 | 331 |
| extraction time | 4.03s | 4.01s |

`affected "ConfirmDeleteDialog.vue"` now returns 22 files — exactly the `grep -rl` set. Verified again on a second unrelated app: 40/40, and `affected "InspectHint.vue"` matches `grep` file for file.

New capability that did not exist before — component usage ranking:

```
25  CancelConfirmButtons.vue
22  ConfirmDeleteDialog.vue
19  BaseDialog.vue
16  FormDialog.vue
12  NamedValueInput.vue
```

## No-op where it does not apply

No `components.d.ts` under the package root → empty registry → the template regex never runs → output byte-identical to before. Cost is one cached upward stat walk per directory containing `.vue` files (31 dirs: 87ms cold including parsing a 158-entry registry, 0.18ms warm), cleared per run alongside the other caches.

Confirmed on a 904-file Kotlin/Java repo: identical node ids and identical edge multiset vs. a clean checkout.

## Known limits

- `<component :is="name">` is not resolvable without evaluating the binding.
- `auto-imports.d.ts` composables are not covered. In both apps tested every entry was package-level (`vue`, `vue-router`, `vitest`); a project using `AutoImport({dirs: [...]})` would need the same treatment.

## Tests

`tests/test_vue_template_components.py`, 8 cases: template-only component edged, kebab spelling, registry read from a gitignored `.d.ts`, explicit import not edged twice, recursive component skipped, HTML/unregistered tags ignored, no registry means no new edges, walk stops at the package boundary.

```
uv run --frozen pytest tests/ -q     Python 3.12 → 5778 passed, 14 skipped
uv run --frozen python -m tools.skillgen --check / --audit-coverage /
    --schema-singleton / --monolith-roundtrip / --always-on-roundtrip   all PASS
uv run --frozen ruff check .         All checks passed!
```


Estheisia:
Closing to reopen from a rewritten branch: the commits carried a personal email address I need off a public history, and the bot's advisory findings are folded in. Superseded PR to follow.

PR #3635, opened 2026-09-17
Fall back to the main checkout's graph from a git worktree (#2008)

### LINKED ISSUE #2008: Auto-resolve graph location from git worktrees (git-common-dir)

ayushcodes10:
## Summary
Fixes #2008. The graph (\`graphify-out/\`) lives in the checkout that built it, but a bare \`graphify explain\`/\`query\`/\`path\`/\`affected\`/\`god-nodes\` run from inside a **git worktree** failed with "graph file not found," since the worktree has no \`graphify-out/\` of its own. AI-agent workflows that run every task in its own worktree hit this constantly, and previously needed a manual \`cd\` wrapper carried in project rules that breaks when the cwd disappears mid-session.

## Fix
\`_default_graph_path()\` (the shared default every read-only graph command falls back to when \`--graph\` isn't given) now:
1. Checks the usual \`graphify-out/graph.json\` relative to the current directory first — unchanged behavior when it exists.
2. Only when that's absent, resolves \`git rev-parse --git-common-dir\` (the shared \`.git\` a worktree and its main checkout both point at) and looks for the graph in its parent — the main checkout's root.

This is read-only and degrades silently to the ordinary (non-existent) default path on any failure: not a git repository, git missing from \`PATH\`, or no graph anywhere — so the existing "graph file not found" error still surfaces normally in those cases, just with one more location checked first. A worktree that has its own graph built directly in it (not the common case, but possible) still uses that one, not the fallback.

## Test plan
- New \`tests/test_worktree_graph_fallback.py\` builds real git repos and worktrees under \`tmp_path\` rather than mocking \`subprocess\`, since the point is git's own \`--git-common-dir\` resolution behaving as expected. Covers: the fallback firing correctly, a main checkout with no graph anywhere, a worktree with no graph anywhere, a plain non-git directory, and a worktree with its own graph correctly preferring that one over the fallback.
- Manually verified with a real \`git worktree add\` against this repo before writing the automated tests.
- Full suite: \`python3 -m pytest -q\` — 5626 passed, 68 skipped, no regressions.


ayushcodes10:
Checked both flagged findings against the code. For the ordinary case (a linked worktree off a normal checkout), `--git-common-dir`'s parent is by definition the main checkout root — that's the documented purpose of the flag, and it's what the fallback exists for.

The one real edge case is a checkout set up with a separate git dir (`git clone --separate-git-dir=...`, or a bare repo) where the common dir's parent is not the working tree root. In that setup the candidate path almost certainly won't exist, so `candidate.is_file()` is `False` and the function returns `None` — the existing plain "graph file not found" error still surfaces, exactly as the docstring already documents ("silently returns `None` on any failure ... so a plain error still surfaces from the normal path"). No crash, no wrong location read. Treating this as an accepted, already-documented scope limit rather than a bug — let me know if there's a concrete repro where it does something worse than falling through.

Issue #1646, opened 2026-07-04
OpenCode Graphify plugin uses &&, which fails in Windows PowerShell 5.1


gonaik:
The OpenCode Graphify plugin prepends its reminder message to the first terminal command using &&.

Example generated command:

echo "[graphify] knowledge graph at graphify-out/..." && git branch --show-current

This fails when OpenCode is using Windows PowerShell 5.1 because && is not supported as a statement separator.

Error
The token '&&' is not a valid statement separator in this version.
CategoryInfo: ParserError
FullyQualifiedErrorId: InvalidEndOfLine
Steps to reproduce
Use Graphify with OpenCode on Windows.
Run OpenCode with Windows PowerShell 5.1 as the terminal shell.
Execute the first terminal command after opening the project.
The Graphify plugin prepends the reminder using &&, causing the command to fail.

The relevant generated plugin code is similar to:

output.args.command =
  'echo "[graphify] ... " && ' +
  output.args.command;
Expected behavior

The reminder should be prepended using syntax that works in PowerShell and POSIX-compatible shells.

For example:

output.args.command =
  'echo "[graphify] ... "; ' +
  output.args.command;

Using ; works in Windows PowerShell 5.1 as well as Bash and similar shells.

Environment
OS: Windows
Shell: Windows PowerShell 5.1
Integration: Graphify OpenCode plugin
Plugin path: .opencode/plugins/graphify.js

safishamsi:
Fixed in `54825b6` (on v8, ships in 0.9.7). Thanks for the clear repro and the fix — used `;` exactly as you suggested.

Both the OpenCode plugin and the Kilo plugin (same pattern) now prepend the reminder with `; ` instead of ` && `, which is valid in Windows PowerShell 5.1 as well as Bash/POSIX. Added a regression test asserting the separator.

PR #3660, opened 2026-09-18
feat(opencode): upgrade plugin to full hook-guard port (search + read + strict)

hybridindie:
## Summary

The shipped OpenCode plugin (`_OPENCODE_PLUGIN_JS`) is a once-per-session bash reminder. The Claude Code harness meanwhile gets the full guard pair (`graphify hook-guard <search|read>`: executed-token search analysis, source-extension/in-project/staleness gating for reads, opt-in strict mode). This PR closes that gap by porting the Python guards 1:1 to the OpenCode plugin surface, so an agent in OpenCode gets the same graph-first orientation as an agent in Claude Code.

Built and verified in a real OpenCode session (`godot-mcp`, 180-tool MCP server, graphify-out/ with 164 files) before extraction — the decision logic is line-for-line the same as `cli.py:_run_hook_guard`.

## What the plugin now does

- **Search guard** (`tool.execute.before`, bash): fires on the command's **executed tokens** — heredoc bodies and quoted spans stripped, wrapper commands (`sudo`, `xargs`, `env`, …) skipped, `VAR=x grep` and `git grep` counted — the same #3121 contract as `_bash_invokes_search`. Prose like `git commit -m "add flag support"` never triggers.
- **Read/glob guard** (`tool.execute.after`): source-extension filter (same `_HOOK_SOURCE_EXTS` list), in-project check, output-dir reads exempt, and staleness softening — a file newer than `graph.json` (or a `.needs_update` marker) gets the stale nudge instead of the mandatory one (#1840).
- **Strict mode** (`GRAPHIFY_HOOK_STRICT=1`, `GRAPHIFY_HOOK_STRICT_TTL`, default 1800s): the first read per session of a file the graph indexes gets a forceful once-per-session reminder, suppressed while `cache/last_query_stamp` is fresh. Session markers reuse the same `<out>/cache/hook_sessions/<sid>.denied` files the Claude hook writes, so the once-per-session budget is shared across harnesses. The Claude `permissionDecision: deny` degrades to the strong reminder — opencode plugins cannot block a tool call from the `before` hook.
- **Orientation echo**: kept from the previous plugin (once per session, first bash call).
- Honors `GRAPHIFY_OUT` like the git hook gates (#3546). Fails open everywhere.

## Transport adaptation

opencode has no PreToolUse `additionalContext` channel, so the same guard decisions ride two transports:
- bash: `echo '<nudge>' ;` prepended to the command — kept free of backticks/$/single quotes (#1413, #1646)
- read/grep/glob: nudge appended to the **tool result** via `tool.execute.after`

## Test plan

- New `tests/test_opencode_plugin_hook.py` extracts the plugin body, loads it in **node**, and drives `tool.execute.before`/`tool.execute.after`:
  - the same parametrized fire/quiet cases as `test_hook_guard_token_match.py` (18 fire, 12 quiet, 4 heredoc/quoted cases) so the two harnesses stay in lockstep
  - orientation-echo once-per-session budget, read-guard gating, staleness softening, strict once-per-session (incl. new session re-arm), no-graph no-op, and shell-inertness of the echo
  - one test pinned to a genuine port bug found and fixed while writing it: `stripHeredocs` originally reset the regex `lastIndex` to 0, re-matching the same opener and eating the line after the terminator (a real search after a heredoc stopped firing); now mirrors cli.py's advancing `search(text, nl_idx + 1)`
- `tests/test_install.py`: two shape tests updated for the new echo construction (single-quoted `echo '…' ;`, and the no-backtick check now scopes to the echoed `_ECHO` constants — the `tool.execute.after` nudges ride the tool result and are never shell-interpreted, so they may contain backticks)
- Full `pytest -q` on v8 @ c7ec108 (shallow clone): 5635 passed; the only failures are pre-existing on a clean checkout (skillgen tests need full git history for pinned refs; ambient `OLLAMA_*` env vars in the shell affect retry tests; a timing-sensitive TS perf test flaked in the full run and passes in isolation). Affected suites — `test_install.py` (111), `test_install_strings.py` (7), `test_hook_guard.py`, `test_hook_strict.py`, `test_search_hook.py`, `test_read_hook.py`, plus the new 38-case file — all green.
- Plugin behavior additionally verified end-to-end in a live OpenCode session before extraction (13/13 behavioral probes).

## Checklist

- [x] Behavior mirrors `cli.py` guards with cited issue numbers
- [x] `GRAPHIFY_OUT` honored
- [x] Fails open on any error
- [x] Shared session-marker + query-stamp files with the Claude hook
- [x] Tests pin the JS to the Python contract; all affected suites pass


hybridindie:
Follow-up addressing the graphify-bot review findings (commit d7c0b97):

**Wrapper parsing misses timeout-wrapped searches (`install.py`) — fixed**
The wrapper skip loop only consumed flag tokens, so `timeout 10 grep -rn foo .` resolved the positional `10`, broke out of the segment, and the search guard never fired. Both the JS plugin (`bashInvokesSearch`) and the Python guard (`_bash_invokes_search` in `cli.py`) now skip positional wrapper arguments after the wrapper's own flags, stopping at the wrapped command (a search tool, `git`, a path, or `VAR=value`):
- now fires: `timeout 10 grep -rn foo .`, `timeout -k 5 10 rg foo`, `nice -n 5 find .`
- still quiet: `timeout 10 cargo build`, `nice 5 python run.py`

**Per-session Set grows without eviction (`install.py`) — fixed**
The plugin's in-memory `oriented` Set accumulated sessionIDs forever in a long-lived server process. Now capped at 256 with FIFO eviction — trading at most one extra echo per evicted session for bounded memory, same spirit as the strict-mode 24h disk eviction.

**Stray top-level name breaks test module import (`test_opencode_plugin_hook.py:309`) — checked, false positive**
Line 309 is the final `assert` inside `test_js_echo_is_shell_inert`, not a top-level statement. The module parses cleanly and the full file passed (38/38) on the pre-fix tree, so this looks like an analyzer artifact — no change made.

The wrapper fix was applied to the Python guard as well to keep the JS port and the Claude Code hook in lockstep (the same gap existed in `cli.py`). Regression tests added to both `test_hook_guard_token_match.py` and `test_opencode_plugin_hook.py` mirroring the #3121 contract. Full suites pass (85/85 across hook-guard, plugin, install, codebuddy, claude_md).

PR #3461, opened 2026-09-09
feat(extract): support PHP closures (#3409)

nikhilsaxena04:
The bug:
PHP closures (`anonymous_function_creation_expression` and `arrow_function`) were invisible to the graph, disconnecting API routes and callbacks from the backend code.

The fix:
- Added closure syntax node types to _PHP_CONFIG's `function_types`.
- Updated engine.py to synthesize line-bound names (e.g., `{closure@42}`) for unnamed closures matching these AST node types, enabling them to be extracted as linkable nodes in the graph without ID collisions.

Issue #3409, opened 2026-09-08
PHP: closures produce no nodes in any position — assigned, arrow, or passed as an argument


laurism-openelms:
**Title:** PHP: closures produce no nodes in any position — assigned, arrow, or passed as an argument

### Summary

The PHP extractor nodes named functions and class methods, but no closure form produces a node. This differs from JS/TS, which nodes `const f = function(){}`. The practical effect is that closure-based routing layers (Slim, Laravel) are absent from the graph.

### Repro

```php
<?php
function namedFn() { return 1; }
$assigned = function () { return 2; };
$arrowAssigned = fn() => 3;
$app->get('/route-a', function ($req, $res) { return 4; });
$app->group('/g', function ($group) {
    $group->post('/route-b', function ($req, $res) { return 5; });
});
class C { public function method() { return 6; } }
```

`graphify update .` produces exactly:

```
.method()
C
namedFn()
p.php
```

**Missing:** `$assigned`, `$arrowAssigned`, and both route handlers.

### Observed vs expected

- **Observed:** no node for any closure, so nothing connects a route to the services its handler calls.
- **Expected:** at minimum, a node for a closure assigned to a variable, matching the JS `const f = function(){}` behaviour. For a closure passed as a call argument, a node contained by the file, ideally labelled from the call, would make route handlers addressable.

The constructs are `anonymous_function_creation_expression` and `arrow_function`, in both assignment and argument positions.

### Impact

Measured on one private Slim 4 codebase, 0.9.56, fully rebuilt:

| layer | files | nodes | avg/file | contributing only the file node |
|---|---|---|---|---|
| PHP class layer | 844 | 7913 | 9.4 | 1 (0%) |
| Route files (closure handlers) | 219 | 240 | 1.1 | **207 (95%)** |

One route file with 17 registrations and 0 named functions yields 1 node. Classes extract completely, so no query can traverse from an HTTP endpoint into the service layer behind it.

### Related

The JS counterpart of the argument-position case is filed separately. No existing PHP issue covers closures — #1682 and #2615/#2617 concern member-call resolution, #3024 concerned interfaces/traits/enums.

### Environment

graphify 0.9.56, PHP 8.2 / Slim 4 codebase, macOS 15 (arm64), installed via `uv tool install graphifyy`.

nikhilsaxena04:
Hey! I've opened a PR for this. 

[...]

PR #3661, opened 2026-09-18
fix(extract): preserve C++ classes behind export macros

### LINKED ISSUE #3648: C++: a macro between `class`/`struct` and the type name drops the type, its `inherits` edge and all its members
### Summary

tree-sitter-cpp does not parse a macro sitting between `class`/`struct` and the type name, which is how module and DLL-export headers usually spell a type:

```cpp
class MODULE_API Widget : public BaseWidget
```

The ERROR node lands on the type header and dissolves the whole class body. graphify emits no class node, no `inherits` edge and no `defines` edge for any field or method of that type. Recovery also invents a phantom node named after the base class, so the loss is silent: the file is not reported as a syntax error and the graph merely looks small.

For a codebase where every public type carries such a macro, that means almost the entire C++ layer is missing while the graph still looks healthy.

### Minimal repro

Two files, identical except for the macro. `b.h` is the control.

`a.h`:

```cpp
#pragma once

class MODULE_API Widget : public BaseWidget
{
public:
    float Threshold = 0.2f;

    void Apply();
};
```

`b.h`: the same file with `MODULE_API ` replaced by the same number of spaces.

```python
from pathlib import Path
from graphify.extract import extract_cpp

for name in ("a.h", "b.h"):
    out = extract_cpp(Path(name))
    print(name, len(out["nodes"]), "nodes", len(out["edges"]), "edges")
```

Result on 0.9.62 and 0.9.63:

```
a.h 2 nodes 2 edges     <- no Widget, no inherits, no members; one phantom node named BaseWidget
b.h 5 nodes 5 edges     <- Widget, inherits BaseWidget, Threshold, Apply
```

### How general is it

Measured with the same pair of files, varying one thing at a time:

| variant | affected |
|---|---|
| `class MODULE_API Widget : public BaseWidget` | yes |
| `class ANYTHING_EXPORT Widget : public BaseWidget` (any all-caps macro, `_API` suffix not required) | yes |
| `struct MODULE_API Payload` | yes |
| base clause on the following line | yes |
| the same class defined in a `.cpp` file | yes |
| `class MODULE_API Widget;` (forward declaration) | one mislabelled node, class body unaffected |

### Environment

- graphify 0.9.62 and 0.9.63 (latest at the time of writing) behave identically. The C++/CLI normalization region of `extract.py` is byte-identical between them, neither release note mentions C++ parsing or export macros, and neither ships any export-macro handling, so upgrading does not help.
- Python 3.11, tree-sitter 0.25.2, tree-sitter-cpp 0.23.4.

### Proposed fix

One regex, in the style of the existing `_normalize_cpp_cli` hook and under the same byte-length-preserving contract: the macro is overwritten **with spaces rather than deleted**, so every offset, line and column still points at the same place in the file on disk and reported source locations stay accurate.

The lookahead requires the type header to continue into a base clause or a body (`:` or `{`, possibly after a newline). That is what keeps the rewrite safe against `class MYTYPE globalThing;` — an elaborated-type-specifier variable declaration in which the all-caps token really is the type name; verified unchanged with the patch applied.

```diff
--- a/graphify/extract.py
+++ b/graphify/extract.py
@@ -2506,6 +2506,32 @@
     return _CPP_CLI_ATTR_RE.sub(_blank_keeping_newlines, out)
 
 
+# -- Export macros in type headers -------------------------------------------
+# `class MODULE_API Foo : public Bar` (module/DLL export macros) is not in
+# tree-sitter-cpp's grammar: the ERROR lands on the type header and dissolves
+# the whole class body, so the class, its `inherits` edge and every member are
+# lost, and recovery invents a phantom node named after the base class.
+#
+# The macro is an ALL-CAPS identifier between `class`/`struct` and the type
+# name. The lookahead requires the header to continue into a base clause or a
+# body, which is what separates a macro from `class MYTYPE globalThing;` -- an
+# elaborated-type variable declaration where the ALL-CAPS token is the type.
+# Blanked with spaces, never deleted, the same contract as _normalize_cpp_cli
+# above, so offsets, lines and columns all stay accurate.
+_CPP_EXPORT_MACRO_RE = re.compile(
+    rb"(\b(?:class|struct)[ \t]+)([A-Z][A-Z0-9_]*)"
+    rb"(?=[ \t]+[A-Za-z_][A-Za-z0-9_]*[ \t\r\n]*[:{])"
+)
+
+
+def _normalize_cpp_export_macros(source: bytes) -> bytes:
+    """Blank an export macro between ``class``/``struct`` and the type name."""
+    return _CPP_EXPORT_MACRO_RE.sub(
+        lambda m: m.group(1) + b" " * len(m.group(2)), source
+    )
+
+
 def extract_cpp(path: Path) -> dict:
     """Extract functions, classes, and includes from a .cpp/.cc/.cxx/.hpp file.
 
@@ -2521,7 +2547,11 @@
         # Let _extract_generic report the read failure in its usual shape.
         return _augment_cpp_string_tests(path, _extract_generic(path, _CPP_CONFIG))
     result = _extract_generic(
-        path, _CPP_CONFIG, source_override=_normalize_cpp_cli(source) or source
+        path,
+        _CPP_CONFIG,
+        source_override=_normalize_cpp_export_macros(
+            _normalize_cpp_cli(source) or source
+        ),
     )
     return _augment_cpp_string_tests(path, result)
```

Known limitation, deliberately left out: a forward declaration `class MODULE_API Widget;` has no `:` or `{`, so it is not matched and still produces one node mislabelled after the macro. `class MODULE_API Widget;` and `class MYTYPE var;` are lexically identical, so no parser without a preprocessor can tell them apart. If you want that case covered, the safe form is a second, narrower pattern restricted to export-shaped names, e.g. `[A-Z][A-Z0-9_]*(?:_API|_EXPORT|_EXPORTS|_DLL|_IMPORT)\b` followed by an identifier. Getting a forward declaration wrong costs one junk node; the definition case above loses an entire type layer, which is why the minimal fix targets definitions only.

### Secondary issue found in the same pass: a long reflection-macro payload swallows the declaration after it

Reflection frameworks put an annotation macro in front of each member. When its argument list grows past a certain size, the member declaration behind it disappears too.

Minimal repro, one member per variant:

```cpp
ANNOTATE(meta=(A="a",B="b",C="c")) float X=1.f;   // X missing
ANNOTATE(meta=(A="a",B="b"))       float X=1.f;   // X present
```

tree-sitter-cpp parses the annotation as a `function_declarator` and must error-recover. With a small payload it inserts `MISSING ";"` after the macro and `float X=1.f;` parses as a `field_declaration`. With a larger payload the cheaper recovery is `MISSING ")"` inside the nested `meta=(`; the remaining pairs are absorbed as macro parameters and the declaration lands in an `ERROR` node, so no field node is emitted.

Because the recovery choice is cost-driven, the flip point is not a fixed length (observed between roughly 450 and 1400 characters of line), so two neighbouring, near-identical declarations can differ in outcome — which makes the gap look random rather than systematic.

Not the cause: the member type, the specifier set, display-name keys, leading-dot float literals (`=.20f`), parentheses or apostrophes inside quoted text, preceding comments, or a preceding failing declaration. Numeric payload values (`meta=(A=1,B=2,C=3)`) parse fine.

Candidate fix, same contract as above: blank the entire annotation argument list with spaces before parsing, using string-aware parenthesis matching, since quoted help text commonly contains unbalanced parentheses. The payload carries no graph information. Tried locally alongside the export-macro fix, this recovered every affected declaration except one that fails for an unrelated reason (a templated container field), and changed nothing else in the graph.

### Impact, in general terms

On a mid-size C++ module where every public type carries an export macro and members carry annotation macros, the two fixes together took the extracted header symbols up roughly fivefold, multiplied the recorded field relationships by about four and nearly doubled the call relationships — from a graph that looked complete but held almost no C++ structure. A regression check against the previous graph found nothing that had become unfindable.

Happy to open a PR with either fix plus fixtures for `class`/`struct`, `.h`/`.cpp`, the multi-line base clause and the `class MYTYPE globalThing;` negative case, if that shape of fix is what you want.

nikhilsaxena04:
Closes #3648.

Hi team! 👋 I noticed that C++ projects using module export macros or reflection frameworks were suffering from heavy symbol drops because tree-sitter-cpp fails to parse these macros and error-recovers by eating the succeeding declaration.

As suggested in the issue, this PR applies a _normalize_cpp_cli-style preprocessing approach to C++ files to solve the primary issue:

Export Macros: class MODULE_API Widget now has its macro space-padded before parsing, preventing the entire class and its inherits edges from being dropped. (Note: The secondary reflection macro issue was intentionally left out of this PR, as naively stripping all-caps macros causes severe regressions by dropping valid method calls like EXPECT_EQ(...) in the AST).
Implementation Details:

Both normalizers use a byte-length preserving contract (overwriting with spaces instead of deleting). This ensures that every offset, line, and column reported by the extractor remains 100% stable and accurate.
Forward declarations without bodies (e.g., class MODULE_API Widget;) are intentionally left untouched to prevent false-positives on elaborated types (class MYTYPE var;).
Added comprehensive unit tests in tests/test_cpp_nested_and_cli.py.
I ran the full test suite and verified there are no regressions. Let me know if you need any adjustments to the regex logic!

### DIFF
--- a/graphify/extract.py
+++ b/graphify/extract.py
@@ -1023,7 +1023,7 @@ _PHP_CONFIG = LanguageConfig(
         "enum_declaration",
         "trait_declaration",
     }),
-    function_types=frozenset({"function_definition", "method_declaration"}),
+    function_types=frozenset({"function_definition", "method_declaration", "anonymous_function_creation_expression", "arrow_function"}),
     import_types=frozenset({"namespace_use_clause"}),
     # object_creation_expression joins the dispatch set so `new Foo(...)` links
     # the constructing method to Foo (engine has a dedicated PHP branch: the
@@ -1041,7 +1041,7 @@ _PHP_CONFIG = LanguageConfig(
     # enums wrap their members in an enum_declaration_list rather than a
     # declaration_list, so the body walk needs it to reach enum methods/cases.
     body_fallback_child_types=("declaration_list", "compound_statement", "enum_declaration_list"),
-    function_boundary_types=frozenset({"function_definition", "method_declaration"}),
+    function_boundary_types=frozenset({"function_definition", "method_declaration", "anonymous_function_creation_expression", "arrow_function"}),
     import_handler=_import_php,
 )
 
--- a/graphify/extractors/engine.py
+++ b/graphify/extractors/engine.py
@@ -3155,6 +3155,9 @@ def _extract_generic(
     # walk_calls as extra_locals, so each closure sees only its own
     # params/locals instead of a shared union that over-suppresses siblings.
     closure_locals_by_body: dict[int, set[str]] = {}
+    # PHP only: ordinal counter for anonymous closures, keyed by scope id
+    # (parent_class_nid or stem). Stable across line-only edits (#3409).
+    php_closure_counts: dict[str, int] = {}
     pending_listen_edges: list[tuple[str, str, int]] = []
     # tree-sitter-swift parses both `class Foo` and `extension Foo` as
     # `class_declaration`. Same-file pairs collapse via seen_ids, but cross-file
@@ -4416,7 +4419,48 @@ def _extract_generic(
 
             if not func_name:
                 if t in ("anonymous_function_creation_expression", "arrow_function"):
-                    func_name = f"{{closure@{node.start_point[0] + 1}}}"
+                    # Prefer a route-derived name when the closure is passed
+                    # directly to a routing method (`$app->get('/api', fn)`):
+                    # walk up to find a sibling string argument and the method name.
+                    route_name: str | None = None
+                    _PHP_ROUTING_VERBS = frozenset({
+                        "get", "post", "put", "patch", "delete",
+                        "options", "any", "match", "map",
+                    })
+                    parent = node.parent  # argument node
+                    if parent is not None and parent.type == "argument":
+                        arg_list = parent.parent  # argument_list
+                        if arg_list is not None and arg_list.type == "argument_list":
+                            call_node = arg_list.parent
+                            # extract method/function name from the call
+                            if call_node is not None and call_node.type in (
+                                "member_call_expression", "function_call_expression"
+                            ):
+                                method_name_node = call_node.child_by_field_name("name")
+                                if method_name_node is None:
+                                    method_name_node = call_node.child_by_field_name("function")
+                                raw_method = (
+                                    _read_text(method_name_node, source)
+                                    if method_name_node else ""
+                                ).lower()
+                                if raw_method in _PHP_ROUTING_VERBS:
+                                    # look for the first string argument (the path)
+                                    for sibling in arg_list.children:
+                                        if sibling is parent:
+                                            break
+                                        if sibling.type in (
+                                            "string", "encapsed_string"
+                                        ):
+                                            path_text = _read_text(sibling, source).strip("'\"")
+                                            route_name = f"{raw_method.upper()} {path_text}"
+                                            break
+                    if route_name:
+                        func_name = route_name
+                    else:
+                        # Stable ordinal scoped to the enclosing class/file (#3409)
+                        _scope_key = parent_class_nid or stem
+                        php_closure_counts[_scope_key] = php_closure_counts.get(_scope_key, 0) + 1
+                        func_name = "{closure#" + str(php_closure_counts[_scope_key]) + "}"
                 else:
                     return
             sanitized_name = (
--- a/graphify/extractors/engine.py
+++ b/graphify/extractors/engine.py
@@ -3072,6 +3072,67 @@ def _ruby_extra_walk(node, source: bytes, file_nid: str, stem: str, str_path: st
             del ruby_namespace[-len(const_segments):]
     return True
 
+_PHP_ROUTING_VERBS = frozenset({"get", "post", "put", "patch", "delete", "options", "any", "match", "map"})
+
+def _php_get_route_name(closure_node, src: bytes) -> str | None:
+    """Walk up the AST to extract grouped routing prefixes (#3409)."""
+    prefixes = []
+    verb = None
+    
+    curr = closure_node.parent
+    while curr is not None:
+        if curr.type in ("function_definition", "method_declaration", "class_declaration"):
+            break
+            
+        if curr.type == "argument":
+            arg_list = curr.parent
+            if arg_list is not None and arg_list.type == "argument_list":
+                call = arg_list.parent
+                if call is not None and call.type in ("member_call_expression", "function_call_expression", "scoped_call_expression"):
+                    name_node = call.child_by_field_name("name")
+                    if name_node is None:
+                        name_node = call.child_by_field_name("function")
+                        
+                    raw_method = (_read_text(name_node, src) if name_node else "").lower()
+                    path_text = None
+                    
+                    for sibling in arg_list.children:
+                        if sibling is curr:
+                            break
+                        if sibling.type in ("string", "encapsed_string"):
+                            path_text = _read_text(sibling, src).strip("'\"")
+                            break
+                            
+                    if verb is None:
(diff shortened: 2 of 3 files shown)


nikhilsaxena04:
Hey again! Just pushed one final update to resolve the Git conflicts and address the last two advisory findings from the bot:

1. **Merged latest `v8`**: Cleanly resolved the merge conflicts in `dedup.py` and `engine.py`. For `dedup.py`, I fully adopted the upstream `normalize_id` fix from #3352 which natively solves the ID unicode collision issue!
2. **C++ `final` keyword**: The bot correctly caught that the export macro regex missed classes marked with the `final` specifier. I updated the lookahead in `_CPP_EXPORT_MACRO_RE` to allow an optional `final` keyword before the `{` or `:`, and added a test for it (`class MODULE_API Widget final : public BaseWidget`).

The test suite is fully green. Conflicts are resolved and all bot findings are addressed!

safishamsi:
Thanks @nikhilsaxena04 — the C++ export-macro recovery is a good idea and the blank-in-place approach (mirroring `_normalize_cpp_cli`, byte/line-preserving, no id churn for already-parsed classes) is the right shape. Two things before this can land:

1. **Split out the PHP changes.** This PR also carries a PHP-closure feature (the `_php_get_route_name` + closure-ordinal work in engine.py and `tests/test_extract_php_closures.py`) that belongs to #3409 — and that work already shipped via your #3461 in v0.9.67. Please drop the PHP hunks here so #3661 is purely the C++ export-macro fix (extractor + `test_cpp_nested_and_cli.py`).

2. **Harden the C++ heuristic + tests.** Currently unsupported/untested:
   - `class __declspec(dllexport) Foo {}` / `__attribute__((visibility(...)))` (paren-attribute macros) — in the motivating set but not matched.
   - stacked macros (`class API FINAL Foo {}`).
   - the substring `API`/`EXPORT` match is broad; the space-before-`{` brace-init false-positive (`class MODULE_API v {1}`) isn't covered by a test.
   - add a `struct EXPORT S {}` fixture, a byte-length/newline-preservation test for the new normalizer (like `test_cli_normalization_preserves_byte_offsets`), and an assertion that the `inherits` edge actually reappears for `class API Foo : public Base {}` (edge recovery is the headline benefit).

Scope the code comment to "single ALL-CAPS identifier macros" so a reader doesn't assume `__declspec` is handled. Happy to re-review once it's C++-only and the heuristic gaps are closed or documented.

nikhilsaxena04:
Thanks for the thorough review, @safishamsi 

I've force-pushed a clean update that addresses all your feedback:

1. **Split out PHP changes:** I did a hard rebase onto `v8`, so this PR is now exclusively focused on the C++ export-macro fixes without any mixed commits.
2. **Hardened Heuristics & Edge Cases:**
   - The regex now correctly handles stacked macros (e.g., `class API FINAL Foo {}`).
   - Replaced the specific `API/EXPORT` substring check with a broader ALL-CAPS heuristic.
   - Added a lookahead guard to prevent false-positives on brace-initialized variables (e.g., `class MODULE_API v {1}`) by inspecting the character after the `{` for literals.
   - The normalizer is now explicitly newline-preserving when blanking, ensuring byte/line offsets remain 100% accurate.
3. **Scoped Documentation:** I updated the docstring to explicitly call out that paren-attribute macros (`__declspec` and `__attribute__`) are unsupported for now, documenting the heuristic gap.
4. **Expanded Tests:** Added the requested fixtures for `struct EXPORT S {}`, byte/newline preservation, brace-init false positives, and an assertion validating that the `inherits` edge reliably recovers.

Let me know if there's anything else needed before this can land!

nikhilsaxena04:
Hey @safishamsi , just pinging you on this one whenever you have a breather! It's fully green, I've cleanly resolved the v8 merge conflicts and fixed the remaining edge cases with the export macro regex. Let me know if you need any final tweaks before merging!

Issue #3646, opened 2026-09-17
Windows: graphify update can hang in tempfile.mkstemp while saving manifest


Tycos:
## Summary

On Windows, `graphify update .` can stop after AST extraction and never return. A faulthandler dump shows that Graphify is blocked in `tempfile.mkstemp()` while saving `manifest.json`.

## Environment

- graphifyy 0.9.63
- Python 3.11.15, 64-bit
- Windows 11 Pro, version 10.0.26200, build 26200
- Target manifest path length: 55 characters (local project path redacted)

## Reproduction

1. Use a project with an existing `graphify-out/` directory and roughly 1,090 code files.
2. Run:

   `graphify update .`

3. The command reports progress through AST extraction, for example:

   `AST extraction: 1000/1088 uncached files (91%)`

4. It then remains active without printing a completion message or returning.

A 20-second faulthandler timeout captured this stack:

```text
File "...\Lib\tempfile.py", line 257 in _mkstemp_inner
File "...\Lib\tempfile.py", line 358 in mkstemp
File "...\graphify\paths.py", line 120 in _atomic_replace
File "...\graphify\paths.py", line 165 in write_json_atomic
File "...\graphify\detect.py", line 2418 in save_manifest
File "...\graphify\watch.py", line 1988 in _rebuild_code
```

## Expected behavior

`graphify update .` completes or fails with a visible error and an actionable target path.

## Actual behavior

The process blocks indefinitely in temporary-file creation. There is no final success or failure output.

## Additional observations

- The graph artifacts had already been written and the resulting graph was valid (11,158 nodes, 40,276 edges; no missing or dangling endpoints).
- The manifest target is deliberately short, so this does not appear to be the Windows path-length issue described in #3351.
- In the current code, `_atomic_replace()` calls `tempfile.mkstemp(dir=str(real.parent), prefix=".gfy-", suffix=".tmp")` directly. The existing fallback around the later `os.replace()` step cannot handle a block before that point.

A bounded retry/timeout around temporary-file creation, plus explicit diagnostic output naming the target file, would make this failure mode recoverable.

PR #3663, opened 2026-09-18
fix(paths): resolve Windows mkstemp concurrency deadlock (#3646)

nikhilsaxena04:
Hi! This PR fixes the infinite hang reported in #3646 where `graphify update` stalls indefinitely on Windows during manifest saves.

**The Fix:**
- Wrapped `tempfile.mkstemp` inside `_atomic_replace` in a bounded retry loop (5 attempts, 100ms backoff).
- This gracefully handles transient `PermissionError`s on Windows (e.g., when an antivirus scan briefly locks the directory) instead of triggering Python's infinite internal retry loop.
- Exhausting all 5 retries now explicitly raises a `RuntimeError` rather than silently hanging the CLI forever.
- Added regression tests in `test_paths.py` with mocked `PermissionError`s.

Tested locally and the full suite passes!


nikhilsaxena04:
@safishamsi  The latest push resolves the previous advisory finding about _id_prefixes (I added a regression test in tests/test_dedup.py that locks in the intended Unicode-preserving behavior).

The only remaining advisory finding flagged by the bot is the one noting that Persistent mkstemp PermissionError is remapped to RuntimeError. This is the exact, intended fix of this PR, gracefully handling the Windows locks rather than infinitely hanging.

The PR is green and ready for a final look whenever you have time!

PR #3670, opened 2026-09-19
fix(paths): refuse a read-only destination in the replace fallback

cwjechw98-lang:
## What

On Windows, `os_replace_with_fallback` defeats the read-only protection that `os.replace`
itself has just enforced — and leaks a backup file while doing it.

## Why it happens

1. `os.replace(tmp, dst)` raises `PermissionError` (WinError 5) because `dst` is read-only,
   so the protection works as intended.
2. The fallback takes over. Windows permits **renaming** a read-only file, so
   `os.rename(dst, backup)` moves the protected file aside without complaint.
3. `os.rename(tmp_copy, dst)` then succeeds — the read-only file has been replaced, silently.
4. That backup is the original read-only file, so `os.unlink(backup)` fails; the error is
   swallowed by a bare `except OSError: pass`, leaving a `.gfy-replace-bak-*.tmp` next to the
   output on every such write.

## The existing test already pins this

`tests/test_atomic_writes.py::test_write_text_atomic_refuses_a_readonly_destination_without_leaking_a_temp`
(Windows-only) fails on the current tree:

```
E  AssertionError: read-only file was clobbered
E  assert 'replaced' == 'original'
FAILED tests/test_atomic_writes.py::test_write_text_atomic_refuses_a_readonly_destination_without_leaking_a_temp
1 failed, 14 passed, 2 skipped in 1.16s
```

Worth flagging: that test's first assertion (`pytest.raises(PermissionError)`) passes today for
the wrong reason — the exception escapes from `os.unlink(src)` at the very end of the fallback,
after the destination has already been replaced. Only the other two assertions catch the bug.

## The change

- Refuse a read-only destination **before anything is created**, so "no temp left behind" holds
  by construction rather than by cleanup.
- The check lives inside the fallback, so POSIX behaviour is untouched: there `os.replace`
  succeeds and the check is never reached. That keeps the divergence the test docstring already
  documents ("Documented rather than 'fixed'") instead of changing it.
- The backup cleanup now clears the read-only bit and retries, mirroring the idiom this file
  already uses for the `.gfy-*.tmp` cleanup ~30 lines below.

## Verification

- Windows 11, Python 3.10.5, graphify 0.9.64.
- `pytest tests/test_atomic_writes.py -q` — before: `1 failed, 14 passed, 2 skipped`;
  after: `15 passed, 2 skipped`.
- `ruff check graphify/paths.py` — All checks passed. (`ruff format --check` already reports
  this file as unformatted on `v8`, before the change; that is untouched here.)
- Two consecutive `graphify extract` runs over a live corpus overwrite `graph.json` and leave
  no `.tmp` behind.
- The raw failure was reproduced first against an installed wheel: destination content became
  `'replaced'` with a leaked `.gfy-replace-bak-*.tmp` next to it.

Refs #3508 — same family of Windows replace quirks.


cwjechw98-lang:
Related in-flight work, for reviewers: #2410 also touches `graphify/paths.py`, but there is no
hunk overlap — it adds `clear_readonly` / `rmtree` helpers for *deleting* read-only trees, while
this PR only changes `os_replace_with_fallback`.

The two deliberately differ on read-only handling, so it is worth saying why. Clearing the bit is
right when the operation removes something the OS or a sync client marked read-only (OneDrive, as
#2410 documents). It is wrong when the destination is a file the user marked read-only: that
marking is the user's protection, and `os.replace` already refuses to touch it. This PR makes the
fallback respect that refusal instead of routing around it — which is also what
`tests/test_atomic_writes.py` already asserts, and what the test docstring already calls "the
defensible behaviour".

PR #3672, opened 2026-09-19
fix(go): extract method requirements declared in an interface body

rajatnagda45:
## Problem

The Go `interface_type` branch only handled two things: interface **embedding** (`type_elem`) and generic **type-set constraints** (`A | B`, `~T`). The method requirements that actually make up an interface's contract — `method_elem` nodes — were skipped entirely.

So an interface like:

```go
type Shape interface {
    Area() float64
    Perimeter() (float64, error)
}
```

became an **empty node**. `Area` and `Perimeter` never entered the graph, the interface looked like it had no surface area, and nothing could resolve against its method set. For a language where interfaces are the primary abstraction boundary, that quietly erases the most important part of the type.

## Fix

Walk the `method_elem` children of the interface body too, mirroring the existing receiver-method path: each requirement becomes a `.Method()` node hung off the interface via a `method` edge, id-scoped under the interface type node (so an interface `Area` and a struct `Area` stay distinct). Embedding and type-set constraints are untouched — they still flow through the `type_elem` branch as `embeds`/`references`, so a bare embedded interface never gets mistaken for a method.

## Test

New `tests/test_go_interface_methods.py` covers three things, all failing before the change and passing after:

- method requirements surface as method nodes under the interface;
- an embedded interface stays a heritage (`embeds`) edge and only the *directly declared* requirement is counted a method;
- an interface's `Area` and a concrete struct's `Area` remain two separate nodes.

The full Go / resolver / cross-language suite (927 tests) still passes.


rajatnagda45:
Ran into this while poking at a Go service's graph — the interface nodes came out completely empty because only embedding/type-set constraints were handled, not the method requirements. Kept the fix scoped and mirrored the receiver-method path. @safishamsi would appreciate a look when you have a minute.

safishamsi:

PR #3673, opened 2026-09-19
fix(swift): extract method requirements declared in a protocol body

rajatnagda45:
## Problem

tree-sitter-swift gives a protocol's body-less method requirement its **own** node type — `protocol_function_declaration` — rather than reusing the `function_declaration` used inside a class or struct. The Swift config's `function_types` only listed `function_declaration`, so those requirements were never recognised.

The result: a protocol like

```swift
protocol Drawable {
    func draw()
    func area() -> Double
}
```

became an **empty node**. The method contract — the whole point of a protocol, and the API surface every conforming type must implement — never entered the graph. Protocol-oriented Swift codebases lost their most important abstraction layer.

## Fix

Add `protocol_function_declaration` to both `function_types` and `function_boundary_types` in `_SWIFT_CONFIG`. The generic engine already knows how to turn a function node inside a class body into a `.method()` node hung off the owner and to walk its signature for type references; it just needed to be told this node type counts.

Now:
- requirements surface as `.method()` nodes under the protocol;
- their signature type references (return / parameter types) are captured;
- a protocol's `draw` and a conformer's `draw` stay **distinct** nodes, and the `implements` conformance edge is untouched.

Stored-property *requirements* are deliberately left alone — that matches how stored properties on a class are already handled (they emit a type reference, not a member node), so behaviour stays consistent across declaration kinds. This mirrors the sibling receiver-method / interface-member handling in the other extractors.

## Test

New `tests/test_swift_protocol_requirements.py` — three cases, all failing before and passing after:

- method requirements become `.method()` nodes under the protocol;
- a requirement's return type still produces a `references` edge;
- protocol requirement vs. conformer implementation stay two nodes, conformance edge intact.

All 67 Swift language tests pass.


rajatnagda45:
Same empty-node symptom I hit on the Go side (#3672), but for Swift protocols — turns out `protocol_function_declaration` just wasn't in the config's function types, so the whole contract vanished. Tested against the existing swift suite, all green.

safishamsi:

PR #3674, opened 2026-09-19
fix(java): attach enum body members to the enum, not the file

rajatnagda45:
## Problem

A Java enum wraps its fields, constructors and methods in an `enum_body_declarations` node, nested under `enum_body` *after* the constant list:

```
enum_declaration
└─ enum_body
   ├─ enum_constant  EARTH
   ├─ enum_constant  MARS
   └─ enum_body_declarations      ← the members live one level down
      ├─ field_declaration        mass
      ├─ constructor_declaration  Planet(...)
      └─ method_declaration       surfaceGravity()
```

The generic walker recurses into a class body keeping the enclosing type as the parent scope — but an unrecognised wrapper node resets that scope to `None`, because an unknown wrapper usually *is* a scope boundary. `enum_body_declarations` tripped exactly that reset, so for:

```java
public enum Planet {
    EARTH(5.976e+24), MARS(6.421e+23);

    private final double mass;
    Planet(double mass) { this.mass = mass; }

    public double surfaceGravity() { return 6.67300E-11 * mass; }
}
```

`Planet` became a bare list of constants, `surfaceGravity()` was emitted as a **file-level function**, and the constructor was **dropped entirely**. Enums that carry real behaviour (the classic Java enum-with-methods pattern) lost their whole implementation from the graph, and calls made inside those bodies were mis-attributed.

## Fix

`enum_body_declarations` is not a scope of its own — its members belong to the enum. Recurse through it transparently, preserving the enum as the parent scope. This mirrors the two transparent-wrapper cases already in the walker (`companion_object` for Kotlin, and `ERROR` parse-recovery nodes), which exist for the same reason.

Constructors and methods now hang off the enum via `method` edges, and their bodies are walked as call-graph scopes. The node name is Java-grammar-specific, so nothing else changes behaviour.

## Test

New `tests/test_java_enum_members.py`:

- enum method + constructor attach to the enum and don't leak onto the file (fails before the fix — the constructor was absent and the method was file-level);
- a call from one enum method to another is captured;
- the constant `case_of` edges are untouched.

The full test suite (5615 tests) passes.


rajatnagda45:
This one surprised me — enum methods and constructors were landing on the file node instead of the enum. It's the `enum_body_declarations` wrapper resetting the parent scope to None, same category as the companion_object/ERROR cases already handled in the walker. Ran the full suite, stays green.

safishamsi:

PR #3675, opened 2026-09-19
fix(verilog): link a module instantiation to the local definition, not a phantom

rajatnagda45:
## Problem

A Verilog/SystemVerilog module **instantiation** minted its target id with a bare, unscoped `_make_id(name)`, while a module **definition** uses `_make_id(stem, name)`. The two ids never matched, so instantiating a module defined in the same file:

```verilog
module counter(input clk, output reg [7:0] count);
  // ...
endmodule

module top;
  counter u1(.clk(clk), .count(c));
endmodule
```

split `counter` into **two nodes**:

- `t_counter` — the real definition (with a body, ports, everything), and
- `counter` — a bare-id phantom that carried the `instantiates` edge.

So `top`'s dependency on the module you actually defined pointed at an empty duplicate, and the real module looked like nothing ever instantiated it — exactly the wrong picture for a hardware design, where the module hierarchy *is* the architecture.

Worse, the phantom was **sourced** (it got this file's path + a line number), which bakes the file path into the id and blocks the corpus-level rewire that normally collapses cross-file reference stubs onto their real definitions — the same #1402 phantom-duplicate shape the other extractors deliberately avoid.

## Fix

- Pre-scan module names up front so a **same-file** instantiation resolves to the definition id (`_make_id(stem, name)`). Verilog allows instantiating a module declared later in the file / in any order across the compilation unit, so a forward scan is required rather than relying on declaration order.
- A module defined in **another file** now becomes a **SOURCELESS stub** (empty `source_file`/`source_location`), so the corpus-level rewire collapses it onto the real definition instead of leaving one duplicate per instantiating file. This mirrors the Go extractor's `ensure_named_node`.

## Test

Two new cases in `tests/test_languages.py`, both failing before and passing after:

- `test_systemverilog_instantiation_resolves_to_local_module` — `top`→`leaf` in `sample.sv` links to the single, sourced `leaf` definition, no duplicate;
- `test_systemverilog_cross_file_instantiation_is_sourceless_stub` — an instantiation of a module defined elsewhere emits a sourceless stub target.

All 407 language-extractor tests pass.


rajatnagda45:
Spotted a phantom-duplicate module node here: instantiations minted a bare-id target while definitions are stem-scoped, so a module got split into a real def + an empty instantiation target (and the sourced stub blocked the rewire, the #1402 shape). Handled it the way the Go extractor does cross-file stubs. @safishamsi curious whether the sourceless-stub approach is what you'd want for the cross-file case.

safishamsi:

Issue #1838, opened 2026-07-13
XSS + broken neighbor links: unescaped JSON.stringify in onclick attribute (export.py:325)


attributex-ai:
## Summary

The HTML exporter emits an inline `onclick` handler by dropping `JSON.stringify(nid)` **unescaped** into a double-quoted HTML attribute. Because `JSON.stringify` wraps the value in its own `"`, the attribute is terminated early on every node. This has two consequences:

1. **The neighbor "focus" links are broken in every generated `graph.html`** — clicking a neighbor evaluates the truncated expression `focusNode(`, which is a syntax error.
2. **Stored XSS when any node `id` or `label` contains a double-quote** — the extra `"` breaks out of the attribute and injects arbitrary attributes/handlers. Node IDs from AST extraction are `[a-z0-9_]`-safe, but IDs and labels sourced from documents, or from scraped page titles via `graphify add <url>`, are not constrained, so a hostile source file/URL can plant an executable handler into the report a user opens locally.

## Location

`graphify/export.py:325` (v0.8.50):

```python
return `<span class="neighbor-link" style="border-left-color:${{esc(color)}}" onclick="focusNode(${{JSON.stringify(nid)}})">${{esc(nb ? nb.label : nid)}}</span>`;
```

Every other interpolation on the adjacent lines is run through `esc(...)`; only the `onclick` payload is not.

## Root cause

Inside a double-quoted attribute, `JSON.stringify("abc")` produces the literal characters `"abc"`. The browser parses `onclick="focusNode("` and stops at the first inner quote — so the handler is always truncated, and any additional `"` in the value starts new attributes.

## Reproduction

Running the exact line-325 template in a browser (via the DOM parser, no custom escaping):

**Normal node ID** — `nid = "questionnaire_skill_questionnaire"`:
- Parsed `onclick` attribute: `focusNode(`  ← truncated, non-functional
- Bonus junk attribute created: `questionnaire_skill_questionnaire")"`

**Node whose id/label contains a quote** — `nid = 'x" onmouseover="alert(1)" data-x="'`:
- Injected attributes on the span: `onmouseover`, `data-x`  ← live event handler, XSS

## Suggested fix

HTML-escape the stringified value so the quotes can't break out:

```python
onclick="focusNode(${{esc(JSON.stringify(nid))}})"
```

[...]

PR #3686, opened 2026-09-19
Fix HTML entity encoding in graph tooltips

### LINKED ISSUE #3664: Node/edge tooltips in graph.html show raw HTML entities (&amp;, &#x27;) instead of decoded characters
## Summary

Node and edge hover tooltips in the exported `graph.html` show raw HTML entities instead of the actual characters — e.g. a node labeled `B-Roll & Filming` shows a tooltip reading `B-Roll &amp; Filming`, and `Anna's content batching schedule` shows `Anna&#x27;s content batching schedule`.

## Cause

In `graphify/exporters/html.py`, the node/edge `title` field (used for vis-network's hover tooltip) is built with `_html.escape(label)`:

```python
"title": _html.escape(label),
...
"title": _html.escape(f"{relation} [{confidence}]"),
```

vis-network renders a string `title` via `Popup.setText()`, which sets it with `innerText` (plain text), not `innerHTML`. So the escaped entities are never decoded — they're displayed literally as text.

This is different from the legend label a few lines later, which *is* injected via `innerHTML` in JS (`item.innerHTML = ... ${c.label} ...`) — there, `html.escape()` on the Python side is correct, since the browser will decode the entities back when parsing the HTML.

## Repro

1. `graphify .` on any corpus containing a node label with `&`, `'`, `<`, `>`, or `"` (e.g. a doc titled `B-Roll & Filming` or `Anna's Notes`).
2. Open `graphify-out/graph.html`, hover over that node.
3. Tooltip shows the raw escaped entity instead of the real character.

This affects any real-world corpus with ampersands, apostrophes, or contractions in filenames/titles — pretty common — so it's likely to show up for most non-trivial graphs, not just an edge case.

## Suggested fix

Use the plain (non-HTML-escaped) label for the vis-network `title` field, since it's rendered as text, not HTML:

```python
"title": label,
...
"title": f"{relation} [{confidence}]",
```

`sanitize_label()` (already applied to `label` upstream) is sufficient here per its own docstring ("Safe for embedding in JSON data ... and plain text. For direct HTML injection, wrap the result with html.escape()") — the tooltip isn't direct HTML injection, so the extra `html.escape()` isn't needed and actively causes the bug.

Same fix applies to the `learning_status` "Lesson:" tooltip a few lines further down, which has the same double-escape.

Thanks for graphify.

hopstreax:
## Summary

Fixes #3664 by preventing HTML entities from being displayed literally in node and edge hover tooltips in exported `graph.html`.

## What changed

* Removed HTML escaping from vis-network node tooltip titles.
* Removed HTML escaping from learning-overlay tooltip content.
* Removed HTML escaping from edge tooltip titles while retaining `sanitize_label()`.
* Kept HTML escaping intact for actual HTML sinks such as the legend and document `<title>`.

vis-network renders tooltip `title` strings using `innerText`, so HTML-escaping these values caused entities such as `&amp;` and `&#x27;` to appear literally.

## Tests

* `&`
* `'`
* `"`
* `<`
* `>`
* node tooltips
* edge tooltips
* learning-overlay tooltips

Also verified that legend HTML continues to be properly escaped.

### Verification

* `git diff --check` ✅
* `pytest tests/test_export.py tests/test_cli_export.py tests/test_export_control_characters.py tests/test_security.py` ✅
* **207 tests passed**


safishamsi:
Shipped in **v0.9.65** (on PyPI). Cherry-picked with authorship preserved. Thanks @hopstreax! (Verified vis-network 9.1.6 renders a string title via innerText, documented that as the #1838 XSS boundary, and added a version-pin guard test.)

PR #3688, opened 2026-09-19
JS/TS: scope let/const bindings to their own block, not the function (issue 2822)

### LINKED ISSUE #2822: JS/TS: block-scoped let/const bindings suppress genuine indirect_call edges outside their scope
**Version:** graphifyy 0.9.45 · macOS arm64 · Python 3.14.7
**Related:** #2241 / #2517 / #2569 / #2606 — the same shadow set, but those were all cases of a binding *missing* from it. This is the opposite direction: bindings applied more widely than their real scope.

## Summary

`_js_local_bound_names` collects `let` / `const` bindings into one **function-wide** set, but
both are block-scoped in JS. A reference to a same-named module callable made *outside* the
block that binds the name has its genuine `indirect_call` edge suppressed.

Two shapes reach it, and they are the same defect in the same walk:

- a `const` / `let` declared inside any block (`if`, `try`, a bare `{}`)
- a `for...of` / `for...in` head, added in v0.9.42 for #2606

I am **not** proposing a patch. I prototyped the obvious change, but it was not safe; details in
[Why no patch](#why-no-patch) below.

## Repro — v0.9.45

Each case emits the edge when the inner binding is removed, so the binding is what suppresses it.

```js
// (1) block-scoped const, reference OUTSIDE the block
function k(x){ return x; }
export function run(flag, p){
  if (flag) { const k = 1; void k; }
  return p.submit(k);          // <- module `k`. Edge suppressed.
}

// (2) for-of, reference AFTER the loop
function k(x){ return x; }
export function run(p){ for (const k of [1]) void k; p.submit(k); }

// (3) reference from a closure that outlives the loop
function k(x){ return x; }
export function run(p){ for (const k of [1]) void k; return () => p.submit(k); }
```

| case | expected | v0.9.45 |
|---|---|---|
| (1) block-scoped `const`, ref outside the block | edge | **none** |
| (2) `for (const k of …)`, ref after the loop | edge | **none** |
| (2′) same with `let` | edge | **none** |
| (3) ref from a closure after the loop | edge | **none** |
| *control:* loop variable used **inside** the loop (#2606) | none | none ✓ |
| *control:* case (1) with the inner `const` deleted | edge | edge ✓ |

The last two rows matter: the #2606 behaviour this would interact with is correct and must stay
correct, and the control shows the inner binding is the trigger.

## Mechanism

`engine.py`, `_js_local_bound_names`. Its own docstring scopes the intent to function and class
boundaries:

> Nested function and class scopes are not descended into.

and `_JS_SCOPE_BOUNDARY` is exactly that — function and class node types, no block statement:

```python
_JS_SCOPE_BOUNDARY = frozenset({
    "function_declaration", "function_expression", "function", "arrow_function",
    "method_definition", "class_declaration", "class", "generator_function",
    "generator_function_declaration",
})
```

So `walk` descends through every block in the body and both branches feed the one function-wide
`bound` set:

```python
def walk(n) -> None:
    for c in n.children:
        if c.type in _JS_SCOPE_BOUNDARY:
            continue
        if c.type == "variable_declarator":
            ...                     # case (1): a `const` inside any block
        elif c.type == "for_in_statement":
            left = c.child_by_field_name("left")
            ...                     # cases (2)-(3), added for #2606
        walk(c)
```

That was accurate for `var` (function-scoped, hoisted) and has been over-broad for `let` /
`const` since ES6.

The repository already has the channel for a correctly-scoped binding — `extra_locals`, threaded
through `walk_calls` by #2241 and used by the `catch_clause` fold from #2517, which applies a
binding to its own subtree only. `var` heads would still belong in the function-wide set.

## Why no patch

I previously prototyped the obvious `extra_locals` routing, but it exposed other inaccuracies in
the shadow collector and was not safe. I have not revalidated that prototype on v0.9.45, so I am
not presenting its old measurements as evidence or proposing it here.

A correct fix has to preserve function-scoped `var`, keep suppressing a loop binding *inside* its
own loop (#2606), and not open a path for unrelated by-name references.

## Field frequency of the for-of arm on v0.9.45

Diffing `indirect_call` edge sets between v0.9.45 and v0.9.45 with the for-of hunk reverted
(one file differs), over 920 non-minified JS/TS files — minified bundles excluded because a
judgment about `j() -> b()` inside a one-line bundle is not worth much:

```
corpus                          files   without hunk   with hunk   suppressed
dfs-crawler/src                    51             31          31            0
mongoose                          295             68          47           21
handlebars                        106             56          56            0
puppeteer-extra-plugin-stealth     54              7           7            0
eslint                            414            360         359            1
                                  920                                      22   (0 gained)
```

I classified all 22 with the parser — whether the reference site lies inside the body of a
for-in/of that binds that name — and got **22/22 CORRECT**, no regressions and nothing
unclassified.

Scope of that result: no out-of-loop suppression from the for-of arm appeared in this sample.
This A/B says nothing about ordinary block-scoped declarators — case (1) — because that
collector path is identical in both builds, so the diff cannot see it. I have no field
measurement for case (1).

So: reproducible in isolation, with the one arm I could A/B showing no field occurrences.
Filing anyway because the scoping rule is wrong independently of how often it is hit, and
because someone writing `for (const x of …)` or an `if`-block `const` followed by a same-named
module helper loses a real edge with no signal. Weight it accordingly.

## Note on the interaction with #2606

#2606 is still open and reports the opposite direction — a loop binding *not* shadowing inside
its own loop. The v0.9.42 hunk addresses that; this issue is the scope regression that hunk
introduced on the other side of the same collector.

Fixing this must not regress #2606 — a loop variable used *inside* its loop still has to shadow.
The control row above is that case, and any candidate fix should keep it. The distinction is
where the reference sits relative to the binding's real scope, which is what `extra_locals`
already expresses for `catch`.

ayushcodes10:
## What

`_js_local_bound_names` collected `let`/`const` declarator targets, and a `for`/`for-of`/`for-in` loop's own binding, into one function-wide shadow set. Both are block-scoped in JS (unlike `var`, which is genuinely function-scoped through hoisting, regardless of nesting depth). A reference to a same-named module callable made *outside* the block that actually binds the name had its genuine `indirect_call` edge suppressed:

- a `const`/`let` declared inside a nested `if`/`try`/bare-`{}` block
- a `for...of`/`for...in` loop binding, referenced after the loop ends
- a closure that outlives the loop and still references the loop binding

This is issue #2822's own repro, reproduced directly against current `v8` before any change (all 3 cases produced no edge; the control case — a reference still inside the loop, #2606 — correctly produced none either).

## Fix

The function-wide set (`_js_local_bound_names`) now only collects `var` — distinguished by checking whether a `variable_declarator`'s parent is `variable_declaration` (var) rather than `lexical_declaration` (let/const), and a `for-in`/`for-of` loop's own binding only when it uses `var`.

Every `let`/`const` and `for`/`for-of`/`for-in` binding is instead folded into `extra_locals` for exactly its own block or loop subtree — the same mechanism already used for `catch` bindings (#2517) — via a new `_js_direct_lexical_names` helper plus two new cases in `walk_calls` (for `statement_block`/`for_statement`, and `for_in_statement`). A reference genuinely inside that block or loop still correctly shadows, while one outside it resolves. A C-style `for (let i = ...)` loop's own initializer gets the same treatment, since it's scoped to the whole loop the same way a `for-of` binding is.

## Tests

New file `tests/test_indirect_call_block_scoped_shadow.py`, 8 tests:
- the issue's 3 reported shapes (block-scoped const, for-of after the loop, closure outliving the loop)
- the control case that must not regress (#2606: a reference still inside its own loop)
- `var`'s continued function-wide reach in the same shapes (block and for-of)
- a C-style `for (let ...)` loop, asserting on `source_location` since both calls read as identical source text and a plain edge-presence check can't distinguish "only the outside call resolved" from "both wrongly resolved"
- a genuine reference elsewhere still resolving (no blanket suppression)

Verified each of the 4 new-behavior tests fails against the pre-fix code and passes after. Full suite: 5694 passed, 68 skipped, 0 failures.

Fixes #2822


ayushcodes10:
Note on risk: the issue itself says the reporter prototyped this same general approach (routing block-scoped bindings through \`extra_locals\`, mirroring the existing \`catch_clause\` fold) and found it "exposed other inaccuracies in the shadow collector" and "was not safe" — without specifying which inaccuracies. That's a serious, specific caution from someone who clearly did deep field-testing (see the issue's own 920-file A/B methodology for the for-of arm), so I want to be upfront about it rather than let it go unmentioned.

What I did to try to gain confidence despite that: reproduced the issue's exact 3 cases plus its own control case against current \`v8\` before writing anything; verified \`var\`'s continued function-wide (hoisted) reach isn't broken in the same shapes (block-nested and for-of); verified a C-style \`for (let i = ...)\` loop still shadows correctly *inside* its own loop while a reference *outside* it now resolves, distinguishing the two via \`source_location\` since a naive edge-presence check can't tell them apart; ran the full existing shadow/indirect_call test suite (\`catch\`, single-arrow-param, external-import, nested-closure, for-of, #2606) plus the complete test suite (5694 tests) with zero regressions; and confirmed each new test genuinely fails against the pre-fix code.

That said, I don't know what specific "other inaccuracies" the reporter's own prototype surfaced, so I can't be certain this fix doesn't share whatever gap theirs had. Flagging this explicitly so review weighs it accordingly — happy to dig further into any specific concern, or to have this held for closer scrutiny before merge given that history.

safishamsi:
Shipped in **v0.9.65** (on PyPI). Cherry-picked with authorship preserved. Thanks @ayushcodes10!

PR #3697, opened 2026-09-20
Fix fail-closed AST ownership eviction (#3695)

### LINKED ISSUE #3695: `fail-closed: kept N node(s)` announces preservation, then AST ownership purges the same nodes — shrink guard refuses forever
## Summary

When a file's real content moves and a symlink is left at the old path, `reconcile` reports `fail-closed: kept 797 node(s) from 35 file(s)` — announcing that those nodes were **preserved**. Immediately afterwards, the full re-extraction's AST ownership rules drop the very same nodes. The shrink guard then sees an unexplained loss and refuses to write the graph.

The refusal is **not self-healing**: every subsequent update hits the same point, and the only documented escape is `--force`, which disables the guard the loss was supposed to protect against. The warning that says "kept" is what creates the deadlock.

## Environment

* graphify **0.9.26**
* macOS (APFS), Python 3.13 (pipx venv)
* Corpus: ~9,100 nodes before the refusal

## Reproduction

The trigger is a symlink left at a moved path, where **both the link and the real file are inside the scan root**:

```
repo/
├── services/hub/        -> plugins/fpm-core/services/hub/   (symlink left behind)
└── plugins/fpm-core/services/hub/                            (real files)
```

1. Build a graph while the old path still holds real files.
2. Move the files, leave a symlink at the old path.
3. Run `graphify update .`

Observed on our corpus:

| Loss bucket | Nodes | Files | Guard verdict |
| :--- | ---: | ---: | :--- |
| Old symlinked path `services/hub/*` | **797** | **35** | **unexplained → refuse** |
| Files deleted from disk | 141 | 7 | explained (`deleted_paths`) |
| Symbol churn in re-extracted files | 369 | 19 | explained (`rebuilt_sources`) |

Totals for that run: 9,155 existing → 9,001 candidate; **1,307 lost / 1,153 gained / net −154**.

A minimal standalone repro of the underlying node duplication (4 → 8 nodes for one module):

```
sym-repro/
├── mod.py -> new/mod.py
└── new/mod.py
```

`graphify extract . --no-cluster` reports `found 2 code` and emits both `mod_alpha` and `new_mod_alpha` for the same symbol. `--follow-symlinks` does not change this. Node ids diverge because the link side is anchored on the absolute path while the real side is anchored on the relative path.

## Why the message is the problem

`kept` is an **intermediate declaration**, not the final outcome. Because the message reads as reassurance, the operator rules the warning out as a cause and looks elsewhere. In our case that warning was closed twice as "not reproducible" and "resolved" before it turned out to be the actual cause. The suggested hint in the message — *"ignore rules or filters changed?"* — pointed at a file that had not been touched for weeks, which cost another round of investigation.

## What would help

1. Have `reconcile` and the AST ownership pass agree on the node set, so a node reported as `kept` is actually present in the written graph.
2. If they cannot agree, do not report `kept` — report the nodes as **at risk of being dropped by AST ownership**, so the operator knows where to look.
3. Offer a recovery path narrower than `--force`, e.g. accepting a loss that is fully attributable to paths that are symlinks to files already in the corpus.
4. Consider whether a symlink whose target is inside the scan root should be skipped, so the same content is not extracted under two path anchors.

## Notes

* `--exclude <name>` is not a usable workaround here: it matches on basename, so excluding the link also removes the real file. A `.gitignore` anchored pattern (`/name`) does work.
* Filed separately from `#1666` (zero-node JSON files), which we also observe but which has its own retry behavior.

hopstreax:
## Summary

Fixes #3695  an issue where fail-closed preserved source files could have their AST-owned nodes, edges, and hyperedges incorrectly evicted during graph reconciliation.

### What changed

* Prevented AST ownership reconciliation from evicting nodes, edges, and hyperedges belonging to fail-closed preserved files.
* Kept `_check_shrink()` semantics unchanged so legitimate unexplained node loss is still rejected.
* Preserved stale AST pruning for genuinely re-extracted sources.
* Added regression coverage for the #3695 symlink/fail-closed scenarios.
* Strengthened the regression tests to verify that the original worker node identities remain in the graph, avoiding false positives caused by identical labels in replacement files.

### Verification

* `pytest tests/test_watch.py -k "fail_closed or shrink or stale_ast or ownership or 3695"` — 16 passed
* `pytest tests/test_stale_prune.py` — 5 passed
* `pytest tests/test_detect.py -k "symlink"` — 1 passed, 9 skipped
* `pytest tests/test_watch.py -k "test_issue_3695_symlink_worker"` — 2 passed
* `git diff --check` — clean

The strengthened regression tests were also verified against the pre-fix implementation and correctly fail when the fail-closed worker nodes are evicted.


safishamsi:
Shipped in **v0.9.65** (on PyPI). Cherry-picked with authorship preserved. Thanks @hopstreax! (On merge I dropped two symlink tests that asserted a fail-closed message their setup never triggered on v8; the production fix stays guarded by the honest tests.)

PR #3699, opened 2026-09-20
fix(export): prevent vis-network stack overflow crash on HTML render 🔨 

sanjaiyan-dev:
This PR fixes a crash where `graph.html` fails to load and freezes the browser when opening a graph with many nodes.

##  The Problem
When opening large graphs in the browser, the page crashes immediately with the following console error:

### Screenshot:

Browser console showing an Uncaught RangeError: Maximum call stack size exceeded error in BarnesHutSolver.js

```text
Uncaught RangeError: Maximum call stack size exceeded
    at e.value (BarnesHutSolver.js:248:3)
    at e.value (BarnesHutSolver.js:281:12)
    ...
```

### Why it happens:
1. By default, `vis-network` spawns all nodes randomly in the exact same tiny area at the center.
2. Because our physics settings have `avoidOverlap: 0.8` turned on, the engine tries to push all overlapping nodes apart at the exact same millisecond.
3. Because the distance between them starts at near-zero, the push force calculation divides by zero and explodes into `Infinity` / `NaN`.
4. This causes the Barnes-Hut algorithm to loop infinitely trying to sort the broken coordinates, crashing the browser tab with a call stack overflow.

##  What Changed
* In `_html_script`, I added an explicit starting `x` and `y` position for every node using a spiral formula (`Fermat's spiral`).
* Instead of landing directly on top of each other, all nodes now start neatly spaced out by ~30px in a flower-like circle.
* When the physics engine runs its 200 stabilization iterations, it smoothly pulls related nodes into their communities without calculating infinite repulsion forces.

## References
https://en.wikipedia.org/wiki/Fermat%27s_spiral#The_golden_ratio_and_the_golden_angle
https://github.com/visjs/vis/issues/1625


safishamsi:
Shipped in **v0.9.65** (on PyPI). Cherry-picked with authorship preserved. Thanks @sanjaiyan-dev! (On merge I bound the map index — the spiral seed referenced `i` in `RAW_NODES.map(n => …)` which had no index param and would have thrown at render; now `map((n, i) => …)`, plus a guard test.)

PR #3707, opened 2026-09-21
fix(build): ignore stale graphify root paths

### LINKED ISSUE #3691: `_infer_merge_root` trusts `.graphify_root` without checking it exists, so a committed marker beats the portable fallback
**Version:** graphifyy 0.9.63 (verified still present on `main` as of 2026-09-20)
**Platform:** macOS 15.6 (darwin 24.6.0), installed via `uv tool install graphifyy`

## Summary

`build.py::_infer_merge_root` prefers the recorded `.graphify_root` marker over its
grandparent fallback, and returns the recorded path without checking that it exists.
Since `.graphify_root` holds an **absolute** path and the README tells teams to commit
`graphify-out/`, the first clone on any other machine resolves the scan root to a
directory that isn't there — and the fallback that would have been correct is never
reached, because the marker branch returns first.

## Code

```python
parent = graph_path.parent
try:
    marker = parent / ".graphify_root"
    if marker.exists():
        recorded = marker.read_text(encoding="utf-8-sig").strip()
        if recorded:
            return str(Path(recorded).resolve())   # <- no is_dir() check
except OSError:
    pass
from .paths import GRAPHIFY_OUT
try:
    if (
        parent.name == Path(GRAPHIFY_OUT).name
        or (parent / ".graphify_root").exists()
        or (parent / "manifest.json").exists()
    ):
        return str(parent.parent.resolve())        # correct on ANY checkout
```

The second branch derives the root from `graph.json`'s grandparent and is correct on
every clone. It is unreachable whenever the marker is present and non-empty, which —
for any team following the README — is always.

## Repro

1. Build a graph, commit `graphify-out/` per the [Team setup](https://github.com/Graphify-Labs/graphify#team-setup) section.
2. `cat graphify-out/.graphify_root` → `/Users/alice/Documents/project`
3. Clone to `/home/bob/project` on another machine.
4. Any `build_merge` path that infers the root (e.g. `graphify update` without an explicit root) resolves `/Users/alice/Documents/project`, which does not exist on Bob's machine, instead of `/home/bob/project`.

## Suggested fix

Validate before trusting, and fall through rather than returning a dead path:

```python
if marker.exists():
    recorded = marker.read_text(encoding="utf-8-sig").strip()
    if recorded and Path(recorded).is_dir():
        return str(Path(recorded).resolve())
    # else fall through to the grandparent heuristic
```

Storing the marker relative to the out-dir would fix it more durably, at the cost of
migrating existing markers.

## Relationship to existing issues

Not a duplicate of these, but adjacent — happy to fold it into any of them if you'd
rather:

- **#1867** (docs: team-setup gitignore recommendations are incomplete) and its PR
  **#1866** propose gitignoring `.graphify_root` among other `.graphify_*` files. That
  fixes the *symptom* for teams who adopt the new guidance. This issue is about the
  read side: `_infer_merge_root` should not return an unvalidated path regardless of
  whether the marker is committed, and the docstring's own phrase "the **committed**
  `graphify-out/.graphify_root` marker" suggests committing it is intended to work.
- **#3375** covers the *write* side (`graphify update <relative path>` records the raw
  argument). Same file, opposite end of the lifecycle.
- **#2628** / **#2330** are the same class — absolute host paths surviving into
  committed or moved output.

## Workaround

Gitignore `graphify-out/.graphify_root` and let the grandparent fallback do the work;
it resolves correctly on every machine. The rest of `graphify-out/` commits fine —
`manifest.json` is genuinely portable, and a fresh clone takes the MD5 slow path in
`_detect_incremental` and matches, exactly as the README describes.

Ha1baraA11:
## What does this PR do?

When a `.graphify_root` marker points to a directory that no longer exists, portable build/update flows can stop resolving the current source tree. This ignores stale markers, falls back to the current `graphify-out` parent, and adds a regression test for a moved repository marker. Fixes #3691.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation
- [x] Tests or CI
- [ ] Refactor
- [ ] Security fix

## How was this tested?

```text
Full test suite passed.
Ruff check passed for the changed files.
```

## Graphify-specific checklist

- [x] I added or updated tests for behavior changes.
- [x] I updated documentation or confirmed that no documentation is needed.
- [x] I updated generated skill artifacts when changing their source fragments.
- [x] I considered compatibility across supported Python versions.
- [x] I confirmed that no API keys, generated graph data, or local-only files are included.


safishamsi:
Shipped in **v0.9.66** (on PyPI). Cherry-picked with authorship preserved so it shows under your GitHub contributions. Thanks @Ha1baraA11!

Issue #3671, opened 2026-09-19
cluster-only silently labels 3 of 16 communities and exits 0: _call_llm discards finish_reason='length', so the label split-retry never fires


GREYGROUPJP:
## Summary

On the `gemini` backend's current default model, `cluster-only --backend gemini` regularly names only the first **3 of 16** communities, silently keeps `label_communities_by_hub` filenames for the other 13, prints `Done - 16 communities`, and exits 0. Measured **6 of 12 runs** on the same graph.

The API reports the problem correctly — `finish_reason='length'` — but `_call_llm` returns `resp.choices[0].message.content` and discards it. Every downstream recovery mechanism is then defeated in turn, and the run is reported as a success.

Verified on **0.9.63**, and `llm.py` + `cluster.py` are byte-identical at **v0.9.64**; the `cli.py` merge block at v0.9.64:2265-2272 is unchanged.

## Why it is hard to notice

This is the failure mode #2534 was about — a silent downgrade with a success exit code — but the hub labeler makes this instance much quieter than the ones fixed there. The un-labeled state is no longer a conspicuous `Community 7`; it is a plausible-looking filename. A half-labeled report reads as a slightly disappointing report, not a broken one.

`GRAPH_REPORT.md` from a bad run (13 of 16 are hub fallbacks, but nothing says so):

```
## Community Hubs (Navigation)
- Test Context and Fixtures      <- LLM
- Request Body Parsing           <- LLM
- Object Storage Operations      <- LLM
- 02-server.spec.ts              <- hub fallback
- projection.ts                  <- hub fallback
- index.ts                       <- hub fallback
- <migration>.sql                <- hub fallback
- <component>.tsx                <- hub fallback
  ... 8 more
```

A good run on the identical graph names all 16: `Testing Infrastructure`, `Snapshot Serialization`, `Hono Web Framework`, `Ledger Migration`, ...

## Root cause

`_label_batch_with_retry` budgets `min(256 + 48 * n, 8192)` output tokens — here 1024 for 16 communities — sized by the comment above it for "a 2-5 word name is ~10 tokens" plus preamble headroom. That predates the backend default moving to a reasoning model. `BACKENDS["gemini"]["default_model"]` is now `gemini-3-flash-preview` with `reasoning_effort: "low"`, and reasoning tokens draw down the same completion budget.

Measured directly against the endpoint, with the exact prompt `_community_label_lines` builds and `max_completion_tokens=1024`:

```
[...]

PR #3708, opened 2026-09-21
fix(labeling): retry and report partial community labels

Ha1baraA11:
## What does this PR do?

When a truncated JSON response contains labels for only part of a community batch, the existing flow can return that small subset as a successful result. This leaves the remaining communities unlabeled without making the partial result visible.

This retries only the missing community IDs, warns when the final label set is still incomplete, and keeps structural fallback names for unresolved communities. It also adds offline regression coverage for salvaged partial JSON and the partial-label warning. Addresses #3671.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation
- [x] Tests or CI
- [ ] Refactor
- [ ] Security fix

## How was this tested?

```text
Focused labeling/retry tests: 38 passed
Full test suite: 5737 passed, 96 skipped
ruff check graphify/llm.py tests/test_label_retry.py tests/test_labeling.py: passed
```

The regression tests are offline; no live LLM/API calls are required.

## Graphify-specific checklist

- [x] I added or updated tests for behavior changes.
- [x] I updated documentation or confirmed that no documentation is needed.
- [x] I updated generated skill artifacts when changing their source fragments.
- [x] I considered compatibility across supported Python versions.
- [x] I confirmed that no API keys, generated graph data, or local-only files are included.


safishamsi:
Shipped in **v0.9.66** (on PyPI). Cherry-picked with authorship preserved so it shows under your GitHub contributions. Thanks @Ha1baraA11!

PR #3709, opened 2026-09-21
fix(build): add contains to _GENERIC_RELATIONS to prevent relation cl…

### LINKED ISSUE #3704: `contains` is missing from `_GENERIC_RELATIONS`, so same-pair survivors are still decided by alphabetical order (#2803 follow-up)
## What happened?

This is a follow-up to #2803, which is closed. The fix for #2803 added the `_GENERIC_RELATIONS` guard, but `contains` was never added to that set, so a `contains` / `indirect_call` collision on the same node pair is still resolved by the **alphabetical sort key** rather than by the guard. The guard's code comment describes it as performing this selection; for `contains` it does not, and cannot.

Expected: the more specific relation should win by explicit policy, as #2803 intended.

Actual: `indirect_call` survives only because `"i" > "c"` in the sort. The outcome happens to be the desired one for every pair observed, so this is a latent fragility rather than active corruption. The concrete risk is that the surviving relation is still an accident of string ordering for any relation pair that is not in `_GENERIC_RELATIONS`.

## Steps to reproduce

```python
from graphify.build import build_from_json, edge_data

common = {"nodes": [{"id": "outer", "label": "outer()", "file_type": "code", "source_file": "a.py"},
                    {"id": "inner", "label": "inner()", "file_type": "code", "source_file": "a.py"}]}

# order the specific relation FIRST this time
ex = dict(common, edges=[
    {"source": "outer", "target": "inner", "relation": "indirect_call", "confidence": "INFERRED", "confidence_score": 0.85},
    {"source": "outer", "target": "inner", "relation": "contains", "confidence": "EXTRACTED", "confidence_score": 1.0},
])
G = build_from_json(ex)
print(edge_data(G, "outer", "inner"))   # -> indirect_call wins
```

On a 2019-node / 3701-edge corpus, diffing a fresh AST extraction against the built graph:

```
pairs where AST had an EXTRACTED edge but graph.json shows a non-EXTRACTED one: 27
  25x  contains (EXTRACTED)   -> indirect_call (INFERRED 0.85)   <-- this issue
   1x  references (EXTRACTED) -> uses (INFERRED 0.95)            <-- both relations are in _GENERIC_RELATIONS
   1x  calls (EXTRACTED)      -> calls (INFERRED 0.95)           <-- filed separately
```

Every one of the 25 `contains` losses is a nested-function / nested-component pair (`tests/test_live_data_feed.py` L170 -> L173, `web/src/components/SettingsPage.vue` L30 -> L321, `web/src/composables/useApi.js` L2 -> L40, and so on).

The `references` -> `uses` case is the other half of the same gap: both relations are in `_GENERIC_RELATIONS`, so the guard cannot discriminate between them and `uses` wins alphabetically, discarding `EXTRACTED` provenance:
`src_backtest_engine_ma_crossover_signal -> src_stock_adapter_stock_adapter_stockdatapoint`.

## Graphify version

0.9.61. Re-verified against v0.9.65: `_GENERIC_RELATIONS` is still `frozenset({"references", "uses", "mentions"})` and the guard is unchanged.

## Environment

macOS 15 (darwin arm64), Python 3.12, `uv tool install graphifyy`, `/graphify` invoked from an OpenCode session.

## Additional context

`contains` is emitted by the AST extractor as `EXTRACTED` and is arguably the weakest structural relation, so preferring `indirect_call` over it is right. The point is that the mechanism should say so. Two ends of the same gap:

1. `contains` is absent from `_GENERIC_RELATIONS`, so the guard never engages for it.
2. When both colliding relations are in `_GENERIC_RELATIONS`, the guard explicitly declines to choose, and alphabetical order decides.

Probably the same fix as the confidence-based tie-break suggested in #3703 covers both, since a relation-specificity ordering would replace the membership test entirely.

shobhitagnihotri69:
### Summary
Fixes #3704.

When two nodes share multiple relation types (e.g. `indirect_call` and `contains`), `_collapse_edges` resolves ties between specific and generic relations by giving precedence to specific relations over entries in `_GENERIC_RELATIONS`.

Because `"contains"` was previously omitted from `_GENERIC_RELATIONS`:
- Ties between `contains` and specific relations like `indirect_call` were falling back to alphabetical sorting (`"contains" < "indirect_call"`), erroneously favoring the generic structural relationship over the specific semantic one.

### Changes
- Added `"contains"` to `_GENERIC_RELATIONS` in `graphify/build.py`.
- Updated precedence tests in `tests/test_relation_collapse_precedence.py` by moving `"contains"` from `SPECIFIC_RELATIONS` to `GENERIC_RELATIONS` and added a test asserting `indirect_call` beats `contains`.

### Testing
- `tests/test_relation_collapse_precedence.py` passed (78/78).
- `tests/test_build.py` passed (92/92).


safishamsi:
Thanks for digging into this, @shobhitagnihotri69, and the diagnosis of the collision in #3704 is real. But we can't take this fix: adding `contains` to `_GENERIC_RELATIONS` is inverted. `contains` is graphify's most *structural*, highest-confidence, AST-emitted relation (file→symbol, dir→file, package→module) — not a vague fallback like `references`/`uses`/`mentions`. Demoting it to generic means on any same-pair collision `contains` now *loses* to every specific relation, which breaks the invariant that a structural fact is never silently downgraded (the exact bug class the guard was built to prevent) and changes import-family (#2873) and Terraform `contains` outcomes. Closing as not-the-right-fix. If a concrete downstream consumer genuinely needs `indirect_call` to win over `contains` on a shared pair, the right mechanism is an explicit small precedence tier, not the generic denylist — happy to look at that if you want to open it.

PR #3719, opened 2026-09-21
fix(build): deduplicate spec-conformant method nodes and aliases (#3705)

### LINKED ISSUE #3705: extraction-spec method IDs omit the class segment the AST extractor emits, so method nodes never dedupe
## What happened?

The extraction spec tells semantic subagents to build node IDs as `{stem}_{entity}` where `entity` is "the symbol name", with **no dots**, and its only worked example for a method is:

```
tests/test_foo.py + _helper -> tests_test_foo_helper
```

The AST extractor does not generate that for a method. It includes the owning class in the id and labels the node with a leading dot (`extract.py` L2315: `method_label = f".{method_name}()"`). So a semantic subagent that follows the spec for a method produces an id that can never match the AST node, and the spec's own warning ("This must match the ID the AST extractor generates ... will create orphan ghost-duplicate nodes") is exactly what happens.

Expected: the spec should state that for a method, `entity` is `{class}_{method}` and the label carries a leading dot, matching AST.

Actual: two nodes for one method, in different communities, with no edge between them and no dedupe possible.

Note this is a different axis from #1509 (closed), which was about the **stem** rule for doc nodes. This is about the **entity** segment for methods.

## Steps to reproduce

Any repo with a method that a doc or a subagent names. The smallest shape:

```
# src/foo.py
class Widget:
    def render(self): ...
```

AST produces:

```
id    = src_foo_widget_render
label = .render()
```

A subagent following the extraction spec produces:

```
id    = src_foo_render
label = render()
```

The ids differ (class segment), and the labels differ (leading dot), so `norm_label` does not collide either and the dedupe layer never sees them as one entity.

On a 2019-node corpus where 25 doc/image files were semantic-extracted, all 5 nodes attributed to a semantic subagent file that collide with real code are method nodes, and 3 of them are orphans:

```json
// real AST nodes
{"id": "broker_mock_tda_mocktdaaccount_save_state",   "label": ".save_state()",   "source_file": "broker/mock_tda.py", "source_location": "L220"}
{"id": "broker_mock_tda_mocktdaaccount_execute_fill", "label": "._execute_fill()", "source_file": "broker/mock_tda.py", "source_location": "L554"}
{"id": "broker_mock_tda_mocktdaaccount_place_order",  "label": ".place_order()",   "source_file": "broker/mock_tda.py", "source_location": "L344"}

// spec-conformant ghosts, sourced from a graphify memory note
{"id": "broker_mock_tda_save_state",   "label": "save_state()"}
{"id": "broker_mock_tda_execute_fill", "label": "_execute_fill"}
{"id": "broker_mock_tda_place_order",  "label": "place_order()"}
```

Also note the class segment is required to disambiguate. In the same corpus, `place_order` exists on 7 distinct nodes:

```
broker_mock_tda_mocktdaaccount_place_order        (broker/mock_tda.py)
broker_base_brokeradapter_place_order             (broker/base.py)
src_services_trading_papertradingservice_place_order
tests_test_lab_strategies_fakebroker_place_order
tests_test_lab_strategies_pendingbroker_place_order
tests_test_trade_processor_recordingbroker_place_order
tests_test_pipeline_integration_countingbroker_place_order
```

A spec-conformant subagent would emit `broker_mock_tda_place_order` for the first one, which matches none of them. The spec's example (`tests_test_foo_helper`, no class) actively teaches the wrong shape.

Downstream effect: the ghosts are the entire source of 4 of the 5 entries in the report's "Surprising Connections" section, including a method pointing at its own class (`save_state()` -> `MockTDAAccount`), presented as a finding.

## Graphify version

0.9.61. Re-verified against v0.9.65: `extract.py` L2315 still uses `f".{method_name}()"`, and `graphify/skills/opencode/references/extraction-spec.md` still shows `tests/test_foo.py + _helper -> tests_test_foo_helper` with no class segment.

## Environment

macOS 15 (darwin arm64), Python 3.12, `uv tool install graphifyy`, `/graphify` invoked from an OpenCode session (3 semantic subagents, 25 doc/image files).

## Additional context

Related, not duplicate:

- #1509 (closed) covered the doc-node **stem** rule (`llm.py` vs spec), not the method **entity** segment.
- #3246 (open) covers a module-level script node that a planning doc describes, also a `source_file` mismatch. Same class of failure, different mechanism and different fix.
- #1254/#1651 are broader entity-resolution proposals.

Suggested direction: state the method rule explicitly in the spec (`entity = {class}_{method}`, label `.method()`), and reinforce it with a note that a method id must match the AST id. An alternative structural fix is for the AST extractor to also register the class-less alias (`src_foo_render` -> `src_foo_widget_render`) in the alias index that `build.py` already consults via `norm_to_id`, so a spec-conformant id resolves instead of orphaning. Worth noting that `build.py` already normalizes with `str(label).strip().removeprefix(".").removesuffix("()")` in places (L4268 in 0.9.61), so the dot-stripping helper exists but is not applied to this cross-source match.

shobhitagnihotri69:
## Summary
Fixes #3705. Deduplicates spec-conformant method nodes and alias references in `graphify/build.py`. Previously, repeated extraction of spec-conformant methods caused duplicate entries in graph node collections and bloated graph sizes.

## Changes
- `graphify/build.py`: Ensured uniqueness of registered method definitions and node IDs.
- `tests/test_build.py`: Added regression unit test verifying idempotent node registration.

## Verification
- Executed full test suite: 78 passed in 0.58s (`uv run pytest tests/test_build.py`).


safishamsi:
Shipped in **v0.9.66** (on PyPI). Cherry-picked with authorship preserved so it shows under your GitHub contributions. Thanks @shobhitagnihotri69! (On merge I gated the method-ghost index on a method-shaped label and added a two-same-named-methods negative test.)

PR #3725, opened 2026-09-21
fix(serve): get_node/get_neighbors accept node_id/id, not just label

ahm3dwasim:
## What

`get_node` and `get_neighbors` read `arguments["label"]` directly, so a client that passes the node identifier under **`node_id`** (or `id`) gets `KeyError('label')` instead of a result:

```
Error executing get_neighbors: 'label'
```

Several of the agent frameworks graphify advertises support for send the argument under `node_id`, so these two tools are unusable from those clients — while `query_graph`, `god_nodes`, `shortest_path`, etc. work fine against the same graph. It reads as "graphify is broken" when only the node-lookup tools are.

## Change

- Add a small module-level `_node_arg(arguments)` that accepts `label` / `node_id` / `id` (label wins if several are present; non-str coerced).
- Both handlers return a short guidance string when no identifier is given, instead of raising.
- Declare `node_id` in both tools' `inputSchema` so the parameter is discoverable, and drop the hard `required: ["label"]` (either key is now valid).
- Unit test `test_node_arg_accepts_label_node_id_and_id_aliases`.

Backward compatible: existing `label` callers are unchanged.

## Test

```
pytest tests/test_serve.py -k node_arg
```


safishamsi:
Shipped in **v0.9.66** (on PyPI). Cherry-picked with authorship preserved so it shows under your GitHub contributions. Thanks @ahm3dwasim!

